PR #795, opened 2016-08-11
Add improved head tracking code

asarium:
This adds an abstraction layer over the previous TrackIR implementation
and also adds an interface for the FreeTrack protocol.

This has been tested in 64-bit mode.


asarium:
Changes have been rebased onto latest master.

PR #804, opened 2016-08-13
Feature/freelook

z64555:
During the antipodes merge into master, it seems the freelook code I put in years (!) ago got blasted into oblivion, presumably because of a bug it introduced which effectively disabled the target padlock view mode.

This PR includes the original commit as well as the bugfix for it, so we get BOTH freelook AND target padlock view! (So exciting.)

(So it's a feature and a bugfix :+1:)

### DIFF
--- a/code/playerman/playercontrol.cpp
+++ b/code/playerman/playercontrol.cpp
@@ -10,6 +10,7 @@
 
 #include "autopilot/autopilot.h"
 #include "camera/camera.h"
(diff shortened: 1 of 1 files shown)


z64555:
Amended second commit with prettier code

chief1983:
Wasn't goober working on the padlock fix for months?

z64555:
Ah, no, I believe he was working on the look_at sexp. similar in operation, but different.

PR #806, opened 2016-08-13
Converts the flags of ship_weapon to use the new flag system

The-E:


asarium:
The one failed CI test seems to be a Appveyor bug where the build failed to start.

PR #810, opened 2016-08-13
Add GLSL version to debug log

Echelon9:
Adding this to the debug log is probably a good idea to aid end user troubleshooting as more and more OpenGL shading language features are utilized in fs2open.

Not clear why fs2open has not recorded this previously, but here it is.

Uses `glGetString(GL_SHADING_LANGUAGE_VERSION)`.

For example:
```
  OpenGL Vendor    : Intel Open Source Technology Center
  OpenGL Renderer  : Mesa DRI Intel(R) Ironlake Mobile 
  OpenGL Version   : 2.1 Mesa 12.1.0-devel (git-3fb4a9b)
  GLSL Version     : 1.20
```


The-E:
We, uhh, already print the GLSL version to the log. A few lines down though, after the whole max texture units, max texture size thing.

Echelon9:
Ah, yes. We print it conditionally down there, and as I'd been playing around with `-no_glsl` flag, it wasn't appearing the log.

Has been too long since picking up fs2open coding actively. I'll close this PR.

Issue #762, opened 2016-08-07
Incorrect use of flags 


The-E:
I'm currently redoing the work I did in the bitset flags branch, and want to use this issue as a catch-all for things I run across while basically auditing our use of flags in the engine.

First up:
https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/network/multi_ingame.cpp#L1691
I am unsure how to proceed here. There's no "ballistic primary" analogue in the ship flags set; this is always a ship_info flag, and the code here would alter that flag for the entire class. I am going to assume that that's the intended behaviour, but I would like @Goober5000 to confirm this.

The-E:
Next: https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/radar/radarsetup.cpp#L537
Needs to be compared against SF2_STEALTH, PR incoming.

Goober5000:
Ballistic primaries is one of the very first things I did for the SCP, and since I wasn't as familiar with the codebase as I am now, there are things I would have done differently.  (And I would use "ammo-based" instead of "ballistic" since some pedant on HLP pointed out that "ballistic" simply means "moves according to the laws of physics" and is unrelated to the concept of ammo.)

There were three levels of ballistic flags.  At the weapon level, the weapon has the WF_BALLSTIC flag if it is ammo-based.  At the ship class level, the ship_info instance had the SIF_BALLISTIC_PRIMARIES flag if it is capable of equipping ballistic primaries.  This flag isn't actually needed, due to the ship loadout list, and actually adds complexity for no real benefit.  It would be a good candidate for removal to make the code a bit cleaner.

The third level is the SF_BALLISTIC_PRIMARIES flag which was assigned to the ship instance if upon mission load the ship happened to be equipped with ballistic primaries.  This was supposed to be an optimization to avoid iterating through all the weapons on the ship, but it didn't actually optimize much and it introduced unneeded complexity, just like the SIF_ flag.  It also unfortunately led to some bugs where SIF_ was confused for SF_ and vice versa.  It was removed soon after its introduction.

I'm not sure what was being done in the particular code snippet mentioned.  It's possible that this was another SIF_/SF_ confusion that was never caught due to the ingame-join code never actually being used.  It's also possible that I confused this with the ship loadout code and something to do with selecting a ship class.  Regardless, assigning the SIF_ flag here is incorrect behavior.

[...]

PR #812, opened 2016-08-13
remove the SIF_BALLISTIC_PRIMARIES flag

Goober5000:
See also my comments on #762.  The flag is almost entirely redundant, so this PR removes it in all instances where it's used in the ballistic primaries code.  There are one or two places where removing it was not a no-op, so I adjusted the rest of the code accordingly (most notably in the parsing validation when it checks that primary weapon cargo has been assigned).  I also added a few divide-by-zero checks and slightly tweaked the support ship rearm code.  The use of the SIF_BALLISTIC_PRIMARIES flag in the ingame_join code was replaced with a dummy value to avoid bumping the multiplayer packet version number.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -5506,11 +5506,6 @@ void set_primary_weapon_linkage(object *objp)
 	ai_info	*aip;
 	ship_weapon	*swp;
 	weapon_info *wip;
-	
-	int total_ammo;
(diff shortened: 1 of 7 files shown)


Goober5000:
I updated and rebased this to accommodate the big SIF flag refactoring.

Goober5000:
Rebased again.  Come on, guys...

Issue #813, opened 2016-08-14
SDL joystick GUIDs are not unique


asarium:
When I added the joystick GUID code I assumed that the GUID returned by SDL would be unique to that joystick so it could be used for identifying it. However, that is not the case. Apparently the joystick only identifies the type and not the instance of the joystick so if there are two devices of the same type they can not be used properly.

One possible solution would be to also use the device index (the one used with `SDL_JoystickOpen`) and use that as an additional identifier if there are multiple joysticks with the same GUID. However, this index is not constant and can change if the joysticks are plugged into a different USB port. Alternatively, there is a instance ID which is supplied by SDL. I don't know if this changes when devices are plugged in somewhere else.

Since the new input code will support multiple joysticks we need a way to reliably identify those so the mappings stay the same. I need more data in order to determine how this issue should be solved correctly. I'll add some code to FSO that collects this data which may help me solve the problem.

chief1983:
I don't even think the OS considers a joystick the same device when plugged into a new Port.  I think windows still reinstalls the drivers for each new port.  I think if people shuffle their joysticks from port to port then we may have to require them to live with re-assigning them.

asarium:
I read that DirectInput has some sort of mechanism that allows to uniquely identify a device even if it was plugged into a different port. I don't think we have access to the necessary information to properly identify devices across platform so I guess we'll have to live with this limitation for now.

Goober5000:
The immediate issue was addressed, and this seems more like a good thing to be aware of rather than an active issue, so I'll close it.

PR #814, opened 2016-08-14
Refine joystick selection

asarium:
This is related to #813. The new code now also uses the device ID if there are multiple joysticks with the same GUID. That's better than it was before but the device ID is not guaranteed to stay the same if the joystick is plugged into a different port so if that happens the code may break.

### DIFF
--- a/code/io/joy-sdl.cpp
+++ b/code/io/joy-sdl.cpp
@@ -80,19 +80,46 @@ SCP_string getJoystickGUID(SDL_Joystick *stick)
 	joystickGUID.resize(GUID_STR_SIZE - 1);
 
 	// Make sure the GUID is upper case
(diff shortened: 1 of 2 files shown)


code/io/joy-sdl.cpp
```diff
@@ -236,7 +239,7 @@ bool device_event_handler(const SDL_Event &evt)
 			}
 
 			// Add the new device to our list
-			auto added = JoystickPtr(new Joystick(device));
+			auto added = JoystickPtr(new Joystick(joyDeviceEvent.which));
```
z64555: Might be good for debugging if the new joystick info was printed to log, instead of just at startup.

asarium: Done.

Issue #649, opened 2016-07-10
PBR/HDR: Sunshafts wash out the screen


AxemP:
![sunwashout](https://cloud.githubusercontent.com/assets/6625164/16711173/5f2a5e00-4612-11e6-9695-f640bce7b52b.png)

To the left is what it currently looks like if you stare at the sun, the right is what it looked like in 3.7.4. But when sunshafts are disabled, PBR and 3.7.4 situations look nearly identical.

PR #821, opened 2016-08-15
Reorder lightshafts in the post process pipeline. 

SamuelCho:
Fixes the overly bright glare from lightshafts after the PBR and HDR commit. Fixes https://github.com/scp-fs2open/fs2open.github.com/issues/649


MageKing17:
LGTM (what little I can understand of the code, and the actual in-game results).

PR #822, opened 2016-08-15
Correct check for valid parse object flag in alter-ship-flag code.

MageKing17:
`if (parse_obj_flag)` used to be the right check back when it was an integer bitfield, but now it's the index of which bit to set, so we need to use the same `!= NUM_VALUES` check the ship/object flags use in the lines above.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -12955,7 +12955,7 @@ void sexp_alter_ship_flag_helper(object_ship_wing_point_team &oswpt, bool future
 			}
 
 			// see if we have a p_object flag to set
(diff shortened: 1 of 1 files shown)


MageKing17:
It's more than a bit silly that we repeatedly cast between ints and enums... I'm wondering if I should just make everything use the enums at the same time, or fix that later.

The-E:
I've made a PR for that at #824

MageKing17:
Since #824 also includes the fix, I'm closing this PR.

The-E:
It turns out that #824 requires much more comprehensive work, would you mind if we merge this in as a stopgap solution?

The-E:
This has now been superceded by #833

PR #823, opened 2016-08-15
Set targetbox lighting as default.

SamuelCho:
Enables lighting for all models drawn in the targetbox.

### DIFF
--- a/code/hud/hudtargetbox.cpp
+++ b/code/hud/hudtargetbox.cpp
@@ -569,10 +569,6 @@ void HudGaugeTargetBox::renderTargetShip(object *target_objp)
 		render_info.set_object_number(OBJ_INDEX(target_objp));
 
 		switch (Targetbox_wire) {
-			case 0:
-				flags |= MR_NO_LIGHTING;
-
-				break;
 			case 1:
 				if (ship_is_tagged(target_objp))
 					render_info.set_color(*iff_get_color(IFF_COLOR_TAGGED, 1));
@@ -726,10 +722,6 @@ void HudGaugeTargetBox::renderTargetDebris(object *target_objp)
 		model_render_params render_info;
 
 		switch (Targetbox_wire) {
-			case 0:
-				flags |= MR_NO_LIGHTING;
-
-				break;
 			case 1:
 				render_info.set_color(255, 255, 255);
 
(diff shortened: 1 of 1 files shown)


code/hud/hudtargetbox.cpp
```diff
@@ -569,10 +569,6 @@ void HudGaugeTargetBox::renderTargetShip(object *target_objp)
 		render_info.set_object_number(OBJ_INDEX(target_objp));
 
 		switch (Targetbox_wire) {
-			case 0:
```
MageKing17: Why not simply make `Targetbox_wire` default to 2, rather than getting rid of case 0?

asarium:
Since this changes how existing mods look IMO it would be better to toggle this behavior with a mod flag.

SamuelCho:
The issue is that PBR models look like crap when unlit since the lighting model puts a heavy emphasis on specular. There needs to be a specular cue somewhere or else the minimal diffuse just render a black model. I'm not very satisfied with unlit maps as a solution.

asarium:
Then if a mod uses PBR models it can enable the mod option but IMO existing mods should not have their appearance changed.

SamuelCho:
Which mods heavily rely on the unlit look? It frankly looks awful and most modders tended to prefer model lighting in the targetbox to be on but were afraid of performance problems which honestly isn't much of an issue these days.

SamuelCho:
I'll try investigating some alternate solutions for making specular look like not crap in unlit mode in case there's more pushback. This conversation gave me some ideas.

asarium:
I wouldn't worry about the appearance of PBR models. If someone uses them without targetbox lighting then it's an issue with the mod that the engine doesn't need to fix. However, the current models AFAIK look fine and simply changing existing behavior because someone thinks that it "looks awful" is not a good way to handle this.
By adding the mod option modders could choose the behavior and if a PBR model is used then the code could show a warning or print a message that notifies the modder that the target box will probably not look good without lighting.

MageKing17:
A mod option doesn't need to be added; a mod simply needs to include a `-hdg.tbm` containing the line `$Wireframe Targetbox: 2` (see also: [this](http://hard-light.net/wiki/index.php/Hud_gauges.tbl#.24Wireframe_Targetbox:))

SamuelCho:
Well let's be honest, the discussion isn't that this changes behavior but more of "Do modders and players find any value in having a targetbox model be unlit by default". We change rendering behavior all the time but no one raises a stink over those changes it because they find significant value in those changes. I mean, it's not like we're trying really hard to keep per-vertex lighting alive by having a on/off per-pixel lighting switch, ha ha. 

I obviously I didn't put enough thought into making unlit rendering work with specular maps, so I'll stop right there. I have some interesting ideas involving Fresnel with uniform lighting that I can act on.

PR #824, opened 2016-08-15
Cleanup for flag-related sexp helper functions

The-E:
Changes a few of the sexp helper functions to use flags directly instead of casting to and from int


The-E:
This is going to take a bit more doing before it's ready; it seems a refactoring of the multi_sexp infrastructure might be in order.

PR #826, opened 2016-08-15
Add option for enabling AddressSanizizer in gcc

asarium:


chief1983:
If I understand correctly, leak and address shouldn't be used together as ASan includes LSan I think.  Additionally, even if they could, I think they need to be merged into a comma separated string parameter to one instance of the argument.  That behavior would also apply to other sanitizers if we add those in the future, such as thread sanitizer.

PR #829, opened 2016-08-15
Add support for indexing flagset with multiple flags

asarium:
This uses the , operator for combining the flag values.

### DIFF
--- a/code/ship/awacs.cpp
+++ b/code/ship/awacs.cpp
@@ -378,7 +378,7 @@ void team_visibility_update()
 			continue;
 
 		// check if ship is stealthed and friendly-invisible
-		if ((shipp->flags[Ship::Ship_Flags::Stealth]) && (shipp->flags[Ship::Ship_Flags::Friendly_stealth_invis]))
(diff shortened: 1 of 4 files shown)


code/ship/awacs.cpp
```diff
@@ -378,7 +378,7 @@ void team_visibility_update()
 			continue;
 
 		// check if ship is stealthed and friendly-invisible
-		if ((shipp->flags[Ship::Ship_Flags::Stealth]) && (shipp->flags[Ship::Ship_Flags::Friendly_stealth_invis]))
+		if ((shipp->flags[Ship::Ship_Flags::Stealth, Ship::Ship_Flags::Friendly_stealth_invis]))
```
MageKing17: The new line is not equivalent to the old; the new line will evaluate to `true` if either `Ship::Ship_Flags::Stealth` or `Ship::Ship_Flags::Friendly_stealth_invis` are set, while the old code would only evaluate to `true` if both of them were set.

asarium: Oh, right. I changed the logic to be consistent but forgot to update this.

PR #837, opened 2016-08-16
Fix a couple joystick issues.

MageKing17:
First commit makes joysticks work again if commit 8f034b08d4cab39829783fd3fa6aa4dbe1786e03 broke it for you.

Second commit makes force feedback work again... again. Commit f91c681ee1c75f4c22f323062bcd9554400a8073 didn't include any calls to the haptic subsystem; I just put them back.


asarium:
Code looks good, unit tests would be nice but right now we don't have a system in place for mocking config values.

PR #842, opened 2016-08-16
Fix -fsanitize flag setting

asarium:
@chief1983 mentioned that the `-fsanitize` flag is not set correctly. This should fix that.

### DIFF
--- a/cmake/toolchain-gcc.cmake
+++ b/cmake/toolchain-gcc.cmake
@@ -29,6 +29,9 @@ if ("${CMAKE_GENERATOR}" STREQUAL "Ninja")
 	endif()
 endif()
 
+# Start with an empty list
+set(SANITIZE_FLAGS)
+
 if(GCC_ENABLE_ADDRESS_SANITIZER)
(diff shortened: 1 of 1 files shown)


cmake/toolchain-gcc.cmake
```diff
@@ -67,16 +88,8 @@ set(CMAKE_C_FLAGS_DEBUG ${COMPILER_FLAGS_DEBUG})
...
 
-if (GCC_ENABLE_LEAK_CHECK)
-	check_linker_flag("-fsanitize=leak" SUPPORTS_FSANITIZE_LEAK)
-
-	if (SUPPORTS_FSANITIZE_LEAK)
-		set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=leak")
-	endif()
-endif()
-
 if (SUPPORTS_SANITIZE_ADDRESS)
```
MageKing17: Keeping this condition means `GCC_ENABLE_LEAK_CHECK` will fail to do anything on its own.

asarium:
@MageKing17 I fixed the issue you commented on.

PR #844, opened 2016-08-17
OpenGL Core, material draw system, and geometry batching rewrite

SamuelCho:
The first step towards our migration to OpenGL Core. Converted all the shaders to be Core compliant. Added some matrix4 math operations. Manually constructed our view and projection matrices. Moved light handling to a uniform system. Moved vertex and index buffer generation out of opengl_tnl and just made tnl simply responsible for buffer object creation and submission. Moved all draw calls to use vertex buffer objects (even immediate draws). Put safeguards around deprecated function calls. There might be a couple deprecated functions here and there that I missed. 

I also wrote a material system. There's basic material, model_material, particle_material, and distortion_material classes. More types of materials will follow as we write additional types of shaders. These basically act as containers for draw parameters for a particular draw call. The different type of materials correspond to the different shaders we use so the graphics code plucks the draw data it needs and sets up the necessary graphics state and uniforms for a draw. This way we don't need to rely on setting ugly global variables all the time. Eventually we can start being smart about materials and just allocate them once for a predictable draw and constantly reuse them instead of making them over and over for each frame. Combined with vertex_layouts, this will be a very powerful rendering tool. This should let us have the flexibility we need to further extend FSO's graphics for more advance rendering features. 

This paradigm also let's us be able to detach code from the internal graphics backend as it can shifts a lot of the state definition away from gropengl*.cpp code. This will make it lot easier for when we start thinking about supporting Vulkan.

I rewritten the geometry batcher because I hacked in dynamic vertex buffer support into the old grbatch.cpp one. This new one is a lot more flexible and should let us be able to account for different types of draws.

Ultimately, I want to pull us away from the TMAP_ flags which was essentially how we described our draw calls. It's enormously ugly and I'm looking forward to deleting any references to that code as soon as I can.

Next Steps:

- Switch out the GLAD compatibility headers with core headers. Subsequently start ripping out the benign deprecated OpenGL function calls.
- Remove fixed function dependant code.
- Start migrating gropengldraw.cpp functions into OpenGL-agnostic g3_render_ functions.

I haven't tested this on Macs because I haven't been able to get it running since the move to the Cmake build system. If anyone has been able to build FSO on Mac lately, I would greatly appreciate it if they can test this pull request.

---
This change is [Reviewable](https://reviewable.io/reviews/scp-fs2open/fs2open.github.com/844)


asarium:
There is a compilation error when compiling under Linux but that was fixed pretty easily. However, the shaders fail to compile. There are the shaders that Mesa sees along with the error messages: [shader_logs.zip](https://github.com/scp-fs2open/fs2open.github.com/files/422989/shader_logs.zip)
Looks like there is a superfluous semicolon at the end of the file.

Since shaders are now absolutely required, would it make sense to turn a shader compilation failure into a fatal error? It's not like we could fall back to the fixed function renderer anymore (or at least we won't be able when the final changes are merged).

The-E:
One thing I noticed while trying this against the usual War in Heaven cutscenes is that event timing seems to be completely broken for some reason.

SamuelCho:
Shader errors triggering fatal errors is a good idea. As for timing issues, are you sure this isn't related to the microseconds change? I haven't ever touched the game logic in this commit so forgive me if it seems like I'm trying to deflect.

The-E:
Yeah, sorry for causing a panic. it seems that particular issue is on my end only and unrelated to the chenges here (it just showed up here for the first time)

SamuelCho:
That's a relief. Ha ha, thanks for the heart attack, E. :P

MageKing17:
Turned out to be an overflow issue introduced by one of the 64-bit-warning-removal PRs (PR #831), but I'm pretty sure I've fixed it (PR #845).

However, what's this "gropenglextension.h" trails.cpp wants to import, but isn't mentioned anywhere else in the codebase or this PR?

asarium:
After fixing the shader errors I can run FSO but I only get a black screen. The debug output from Mesa shows that that's because this uses some functionality that is deprecated in a core context.

freespace2/SDLGraphicsOperations.cpp
```diff
@@ -69,6 +69,14 @@ std::unique_ptr<os::OpenGLContext> SDLGraphicsOperations::createOpenGLContext(co
 	SDL_GL_SetAttribute(SDL_GL_MULTISAMPLEBUFFERS, (attrs.multi_samples == 0) ? 0 : 1);
 	SDL_GL_SetAttribute(SDL_GL_MULTISAMPLESAMPLES, attrs.multi_samples);
 
+	const int gl_versions[] = { 41, 40, 33, 32, 31, 30, 21, 20 };
```
asarium: Do we actually require an OpenGL 4.1 context? IMO it would be better to stick to one OpenGL version (e.g. 3.3 or 3.2) and if additional functionality is required we obtain that by using the corresponding extension if it exists.

SamuelCho:
I thought the whole point of using GLAD so that we wouldn't have to mess with adding extensions all the time. I'd rather just leave it at 4.1 instead of having to repeatedly have to generate a new header each time I want to implement a new feature. Is that too much to ask?

asarium:
The problem with that is that a driver which does not support OpenGL 4.1 but supports the functionality we need through an extension would not be usable by us.

What's the issue with generating new header if you need new functionality? It's just a small python program and they even have a web interface.

SamuelCho:
The generator is fine and the GLAD files in this commit are ones I generated myself. It still didn't make it less annoying to my workflow having to pick and choose extensions all the time then running the script. It's especially a boner killer if you're doing experimentation or prototyping a feature. 

I mean if you want to tell me to deal with it, I'll deal. I just feel this is unnecessary friction. I was hoping on gating functionality by simple version checking in the future instead of having to juggle extensions. I mean, yes, 3.2/3.3 has most of the features we need. But I feel the case where we would want a 4.0+ feature and want it getting it working on a 3.3 limited driver isn't too terribly important.

MageKing17:
How are you guys getting this to compile? Are you just stripping out the `#include "graphics/gropenglextension.h"` line, or what?

SamuelCho:
Yeah get rid of that line. This was a kind of sloppy merge because certain files were deleted in the main branch and git kept those files since I had changes that needed to be reconciled. Stuff obviously fell between the cracks.

asarium:
@SamuelCho Thanks for clarifying your decision. Just from looking at the code I wasn't sure why you did it this way. If the version checks can be used reliably (and they probably can) then I see no further issue with that.

One thing I noticed is that the Mesa drivers give us a core profile since we request a 4.1 context. This means that some extensions like `EXT_framebuffers` are not being exposed anymore and only the ARB variant is there. Since OpenGL Core includes framebuffer support I would suggest simply removing the extension and using the core functionality.

SamuelCho:
Yeah, I figured that too. I think I did convert the old framebuffer binds away from the EXT extension in my original branch but I didn't carry that over. So I wouldn't be surprised if that, among many other factors, are causing errors in the Mesa context. Again, I wish I could run this on Mac because that will probably give me a more strenuous test. I'll try testing on Intel drivers later. I'm currently proofing the code a bit more.

asarium:
I made the framebuffer changes in my local copy so if you want I could send you a patch with the necessary changes.

SamuelCho:
Sure, send it over.

asarium:
Ok, here are the diffs. One is for the shader changes and one is for the framebuffer changes: [diffs.zip](https://github.com/scp-fs2open/fs2open.github.com/files/424995/diffs.zip)

SamuelCho:
Killed all traces of deprecated API calls. Let's find out if this can run on Mesa.

asarium:
I'm currently compiling your code. You missed an EXT function in `gropengltexture.cpp` on line 1710. Also, please update `extensions.txt` in the glad directory so everyone can generate the same headers.

asarium:
I forgot to disable `-no_glsl` which caused a lot of problems. Once I disabled that those errors were gone but now a shader fails to compile since "error: #extension directive is not allowed in the middle of a shader". That is caused by the FXAA shader since you define a few `out` variables before beginning the shader. Also, there are still a few shaders which have unnecessary semicolons at the end. Here is another ZIP of shaders where the info log wasn't empty: [shader.zip](https://github.com/scp-fs2open/fs2open.github.com/files/426721/shader.zip)

The-E:
With this build, i am getting "OpenGL Error from shader_set_current(): invalid operation" several times per frame.

SamuelCho:
No shader errors in your log?

The-E:
WARNING: 1:37: warning(#62) enable/warn/disable extension is not found. Extension "GL_ARB_texture_buffer_object" is not supported

On pretty much every single one

SamuelCho:
Good to know, thanks.

asarium:
What is the minimum OpenGL version that is required by these changes?

Since the shader errors I am seeing are caused by the compatibility defines needed for older GLSL versions another solution has to be found for that. However, if those versions aren't going to be supported anyways then the compatibility stuff could just be removed and the shaders could use the OpenGL 3 syntax by default.

asarium:
Based on the assumption that the compatibility code isn't needed anymore I removed that code and fixed the shaders. That enabled me to run FSO using the Mesa drivers and HDR enabled (it looks great, great work @SamuelCho!). I can submit these changes to your branch as a PR if you want to pull them in.

asarium:
I also found an error in `shockwave.cpp` on line 402. You used the constant 94 instead of `Shockwave_info[1].bitmap_id` for the bitmap handle which causes an assertion.

SamuelCho:
Yeah, I had those defines to be a fallback so I could easily go between Core and Compatibility. But since version 3.2 is our floor now, we should just get rid of those. 

Go ahead and send the pull request when you're ready and I'll merge it it. Thanks for helping me out. Glad to know it finally works for you.

asarium:
There are still a few issues (e.g. no shockwave distortions) but for the most part everything seems to be working fine. The log file says that shadows are not available but that's probably similar to the framebuffer extension situation.

asarium:
The last few changes haven't fixed all errors. `ls-f.sdr` is still using gl_TexCoord instead of the output variable of the vertex shader. I'll send you another PR with the required changes.

asarium:
I opened another pull request for your branch that removes most extensions and replaces the function calls with core profile operations.

asarium:
Another issue I noticed is that explosion distortions aren't rendered anymore. I tracked the cause down to `batching_render_all` which is never called with a `true` parameter. There is `batching_render_distortions_all` but that is only called in the lab viewer.

asarium:
That last changes fixed the shockwave distortion for me so as far as I can tell everything should be good to go.

One thing that might need some testing is FRED. Since that doesn't use SDL for context creation it only creates a compatibility profile. However, since FRED is Windows only that shouldn't be a problem since those drivers support the Core profile features in the compatibility profile.

code/render/batching.h
```diff
@@ -0,0 +1,129 @@
...
+
+	size_t num_verts() { return Vertices.size();  }
+
+	void clear();
+};
+
+struct primitive_batch_item {
+	batch_info batch_item_info;
+	int offset;
+	int n_verts;
```
asarium: These two fields should probably use `size_t` or `ptrdiff_t` since they are dependent on the pointer size.

code/render/batching.h
```diff
@@ -0,0 +1,129 @@
...
+};
+
+struct primitive_batch_buffer {
+	vertex_layout layout;
+	int buffer_num;
+
+	void* buffer_ptr;
+	int buffer_size;
+
+	int desired_buffer_size;
```
asarium: These two fields should also be a `size_t`.

code/render/batching.cpp
```diff
@@ -0,0 +1,992 @@
...
+{
+	Assert(draw_queue != NULL);
+
+	if ( draw_queue->buffer_size < draw_queue->desired_buffer_size ) {
+		if ( draw_queue->buffer_ptr != NULL ) {
+			vm_free(draw_queue->buffer_ptr);
+		}
+
+		draw_queue->buffer_size = draw_queue->desired_buffer_size;
+		draw_queue->buffer_ptr = vm_malloc(draw_queue->desired_buffer_size);
```
asarium: According to LeakSanitizer, this memory is never freed.

asarium:
I just tested FRED and it seems to work fine so that's not an issue. It might be a good idea to confirm that it's also working with Nvidia but I'd guess that it also works with those drivers.

asarium:
There is another bug I encountered but I don't know if it's related to the Mesa drivers or the rendering engine:
ogl_core_bug

The 3D radar is white instead of transparent. The issue does not appear with the 2D radar.

asarium:
@SamuelCho Since #849 changes how the OpenGL context is created it breaks this PR. I submitted a PR to your branch that should fix that.

asarium:
That last change fixed the radar issue for me. There are still a lot of warnings in the Visual Studio builds that should be fixed before merging this.

Talon1024:
This looks interesting. The description makes it sound like these changes would make physically-based rendering/shading optional on a per-material basis. Am I correct?

SamuelCho:
No, that's not what it does. This just parameterizes rendering state into a data structure. We're just rearranging the plumbing, This offers no new features.

asarium:
@SamuelCho Could you update this to latest master? There have been some CI changes that may require some additional fixes.

asarium:
@SamuelCho Are there any changes left you would like to make before merging this?

SamuelCho:
This should be good to go.

PR #848, opened 2016-08-18
Fix/freelook1

z64555:
This fixes the issues with freelook and viewmodes that the player can control. Namely re-organizes things so that we can better understand what state does what and isolate the state control and view modifications from each other.


z64555:
Just a heads up, there's a few bugs in this PR that need to be worked out. I *had* hoped this would be a simple fix. **grumbles**

z64555:
Ok, looks like I'm finally done with this here.

PR #849, opened 2016-08-18
Add interface for viewport management

asarium:
This separates OpenGL context creation from window creation and also allows to implement multiple viewports in the future. This is a continuation of the previous PR where I added the OpenGL context abstraction.


asarium:
#844 requires some additional changes to the OpenGL context creation interface.

asarium:
Code has been changed, now it should be able to handle what the OpenGL Core changes need.

PR #22, opened 2014-05-05
Add option to dump pilot files in JSON format

niffiwan:
With current format settings, 1000x easier to read than raw files
Downside is that the files are much larger; ~33k-200k
    (size could be reduced by using a less readable format
    and relying on external JSON viewers)
Controlled by cmdline param -json_pilot
    (autoenabled in debug mode)
JSON format files use .plj & .csg as the extensions **NOTE** Commit log has a typo, should be .plj and .csj **END NOTE**
Could be extended in future to use JSON as pilotfile format
    (i.e. read in JSON)

PR #854, opened 2016-08-19
Add option for writing pilot files in JSON format

asarium:
This is a new implementation of the JSON dumping which was already
proposed in #22 but this version tries to allow implementing both
reading and writing JSON files without having to change the existing
code too much. This is done by using an abstract class to hide the
differences between writing the JSON and binary files.

This is currently only capable of writing the files but not reading them
so if the option is enabled all changes to the pilot file are lost. I
will implement reading after this PR has been merged since I would like
to reduce the amount of changes that need to be reviewed.

### DIFF
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -38,9 +38,10 @@ IF (FSO_USE_SPEECH)
 	target_link_libraries(code PUBLIC speech)
 ENDIF(FSO_USE_SPEECH)
 
+TARGET_LINK_LIBRARIES(code PUBLIC jansson)
+
 # Headers for standalone "gui" on UNIX
 IF(UNIX)
-	TARGET_LINK_LIBRARIES(code PUBLIC jansson)
 	TARGET_LINK_LIBRARIES(code PUBLIC ${MONGOOSE_LIBS})
 ENDIF(UNIX)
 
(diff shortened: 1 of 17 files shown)


MageKing17:
> I will implement reading after this PR has been merged since I would like to reduce the amount of changes that need to be reviewed.

This sounds like a terrible idea.

chief1983:
I didn't think reading was ever part of the plan, thought it was just going to be a sanity check for pilot data

asarium:
> This sounds like a terrible idea.

If you are going to make such a statement then you should at least explain why it would be a terrible idea.

MageKing17:
> If you are going to make such a statement then you should at least explain why it would be a terrible idea.

You are encouraging us to merge a PR that would intentionally leave the master branch in a state where it's not actually usable, in the hopes that the code to make it usable will then be *written in the first place*. I'm surprised I have to point out why this is inherently not a good idea.

Like, if you want to split it into multiple commits or even multiple PRs to make it easier to review individual changes, that's one thing, but not even *implementing* reading JSON-based save files until after this PR is merged? That is a **terrible idea**.

asarium:
You do realize that the JSON writing is only enabled when setting a command line option, right?

chief1983:
And even if it always happened, it doesn't break anything.  The files just get written to disk (at least that's how I thought it worked, it isn't replacing the binary writing yet right?)

chief1983:
And even if it always happened, it doesn't break anything.  The files just get written to disk (at least that's how I thought it worked, it isn't replacing the binary writing yet right?)

asarium:
@chief1983 If the option is enabled then no binary file will be written.

MageKing17:
> You do realize that the JSON writing is only enabled when setting a command line option, right?

I did not, because you didn't mention it anywhere in your PR description. In that case, it should be fine.

PR #859, opened 2016-08-19
Converts model flags

The-E:


code/hud/hud.h
```diff
@@ -57,7 +57,7 @@ typedef struct hud_subsys_damage
 {
 	int	str;
 	int	type;
-	char	*name;
+	char* name;
```
asarium: I don't know if it makes sense but this could be changed to a `const char*` if possible.

code/model/modelread.cpp
```diff
@@ -511,8 +512,10 @@ void model_copy_subsystems( int n_subsystems, model_subsystem *d_sp, model_subsy
 		for ( j = 0; j < n_subsystems; j++ ) {
 			dest = &d_sp[j];
 			if ( !subsystem_stricmp( source->subobj_name, dest->subobj_name) ) {
-				dest->flags |= (source->flags & MSS_MODEL_FLAG_MASK);
-				dest->flags2 |= (source->flags2 & MSS_MODEL_FLAG2_MASK);
+                for (auto it = std::begin(carry_flags); it != std::end(carry_flags); ++it) {
```
asarium: This could be a for-each loop.

code/model/modelread.cpp
```diff
@@ -5821,3 +5824,80 @@ void glowpoint_init()
...
+    turret_base_rotation_snd = 0;
+    turret_base_rotation_snd_mult = 0;
+    turret_gun_rotation_snd = 0;
+    turret_gun_rotation_snd_mult = 0;
+
+    alive_snd = 0;
+    dead_snd = 0;
+    rotation_snd = 0;
+
+    engine_wash_pointer = NULL;
```
asarium: Instead of `NULL` you could use `nullptr`.

code/playerman/playercontrol.cpp
```diff
@@ -1600,11 +1600,6 @@ int player_inspect_cap_subsys_cargo(float frametime, char *outstr)
 		return 0;
 	}
 
-	// don't scan cargo on turrets, radar, etc.  only the majors: fighterbay, sensor, engines, weapons, nav, comm
-	if (!valid_cap_subsys_cargo_list(subsys->system_info->subobj_name)) {
```
asarium: Why has this block been removed in the new version?

code/ship/ship.cpp
```diff
@@ -1793,8 +1793,7 @@ ship_info::~ship_info()
 			}
 		}
 
-		vm_free(subsystems);
-		subsystems = NULL;
+        subsystems = NULL;
```
asarium: Previously the memory was freed here. Where is the memory freed now?

code/ship/ship.cpp
```diff
@@ -4403,81 +4401,68 @@ int parse_ship_values(ship_info* sip, const bool is_template, const bool first_t
...
+                }
+                else {
+                    sp->flags = tmp_flags;
+                }
 
-					if ( !flag_found )
-						Warning(LOCATION, "Bogus string in subsystem flags: %s\n", flag_strings[i]);
-				}
+                if (errors.size() > 0) {
+                    for (i = 0; i < errors.size(); ++i) {
```
asarium: This could be a for-each loop.

code/ship/ship.cpp
```diff
@@ -6325,7 +6311,7 @@ void ship_copy_subsystem_fixup(ship_info *sip)
 		// subsystem since previous error checking would have trapped its loading as an error.
 		Assert( it->n_subsystems == sip->n_subsystems );
 
-		source_msp = &it->subsystems[0];
+		source_msp = const_cast<model_subsystem*>(&(it->subsystems[0]));
```
asarium: Why does this need a `const_cast` in the new version?

code/playerman/playercontrol.cpp
The-E: Some time ago, valid_cap_subsys_cargo_list was changed to always return 1. As a result, this block will never be executed; I have removed that function and all calls to it in this PR.

code/ship/ship.cpp
```diff
@@ -13226,15 +13212,15 @@ int ship_get_index_from_subsys(ship_subsys *ssp, int objnum, int error_bypass)
 /**
  * Returns the index number of the ship_subsys parameter
  */
-int ship_get_subsys_index(ship *sp, char *ss_name, int error_bypass)
+int ship_get_subsys_index(ship *sp, SCP_string ss_name, int error_bypass)
```
asarium: `ss_name` should be a `const SCP_string&`.

code/playerman/playercontrol.cpp
asarium: I just saw that function, thanks for the explanation.

PR #861, opened 2016-08-19
Ai profile flags converted

The-E:


code/ai/ai_profiles.cpp
```diff
@@ -28,24 +28,13 @@ static char Default_profile_name[NAME_LENGTH];
 
 
 // utility
-void set_flag(ai_profile_t *profile, char *name, int flag, int type)
+void set_flag(ai_profile_t *profile, char *name, AI::Profile_Flags flag)
```
asarium: While you're at it, could you make `name` a `const char*`?

code/ai/ai_profiles.cpp
```diff
@@ -486,20 +475,14 @@ void parse_ai_profiles_tbl(const char *filename)
...
 				if (optional_string("$allow primary link delay:"))
 				{
 					mprintf(("Warning: \"$allow primary link delay\" flag is deprecated in favor of \"$allow primary link at mission start\"\n"));
 					bool temp;
 					stuff_boolean(&temp);
-					if (temp)
-						profile->flags2 &= ~AIPF2_ALLOW_PRIMARY_LINK_AT_START;
-					else
-						profile->flags2 |= AIPF2_ALLOW_PRIMARY_LINK_AT_START;
+                    profile->flags.set(AI::Profile_Flags::Allow_primary_link_at_start, temp);
```
asarium: I think this needs to be inverted. The previous version disabled the flag then `temp` was true, now it enables it.

asarium:
I looked at all changes and everything, except the lines I commented on, look good.

PR #863, opened 2016-08-20
Add -nograb commandline option

asarium:
This disables all mouse grabbing which is useful for debugging when SDL gets interrupted without giving up control of the mouse. That usually leads to problems since you can't interact with the debugger window anymore.


chief1983:
Didn't the pre-SDL window behave such that if you hit alt-tab or something, it would ungrab the mouse for you?  Would there be a way to duplicate something like that so we could avoid having a command line option added, and not having to remember to enable it when debugging?

asarium:
The problem occurs if our code isn't executed anymore, e.g. if the debugger hits a breakpoint. I haven't seen this issue on Windows but on Linux it happens quite frequently. That means that it's practically impossible to implement a solution that works automatically.

PR #869, opened 2016-08-22
Fix memory leak reported by LeakSanitizer

asarium:


MageKing17:
Using a `unique_ptr` seems kinda overkill when `temp_string` isn't being returned and there aren't any ways for `parse_wi_flags()` to return before `temp_string` is no longer needed; could just stick a `delete` on line 607 instead.

asarium:
`unique_ptr` makes sure that even if an exception is thrown the memory is freed and the parse code can throw exceptions so a `unique_ptr` is a better choice than using `delete`.

MageKing17:
The parse code isn't called during that stretch of code.

asarium:
Right. Nevertheless, `unique_ptr` is almost always the better choice over raw memory management.

PR #870, opened 2016-08-22
Ai info flags conversion

The-E:


code/ai/aicode.cpp
```diff
@@ -13681,7 +13678,7 @@ int aas_1(object *objp, ai_info *aip, vec3d *safe_pos)
 		float outer_rad = ship_get_exp_outer_rad(ship_objp);
 
 		if (vm_vec_dist_quick(&objp->pos, &ship_objp->pos) > outer_rad*1.5f) {
-			aip->ai_flags &= ~AIF_AVOID_SHOCKWAVE_WEAPON;
+			aip->ai_flags.remove(AI::AI_Flags::Avoid_shockwave_ship);
```
The-E: This is (technically!) a behaviour change. Do note however that the original behaviour was wrong: At this point in the code, we're dealing with the question of whether or not to avoid a shockwave from a ship, not a weapon. Some code archeology is needed whether this is a change introduced by the SCP, or whether it was present in the original source release.

I also think the question of whether or not we need two Avoid_shockwave flags is absolutely necessary should be investigated.

code/ai/aicode.cpp
```diff
@@ -6636,14 +6634,15 @@ void attack_set_accel(ai_info *aip, ship_info *sip, float dist_to_enemy, float d
...
 		if ((dist_to_enemy > 200.0f) && (dist_to_enemy < 800.0f)) {
 			if ((dot_from_enemy < 0.9f) || ai_near_full_strength(Pl_objp)) {
 				accelerate_ship(aip, MAX(1.0f - (dist_to_enemy-200.0f)/600.0f, 0.1f));
 				return;
 			}
-		} else
-			aip->ai_flags &= ~AIF_ATTACK_SLOWLY;
+		}
+		else
+			aip->ai_flags.set(AI::AI_Flags::Attack_slowly);
```
asarium: The original code removes the flag but the new code adds it.

code/ai/aicode.cpp
```diff
@@ -13972,36 +13970,37 @@ void ai_frame(int objnum)
...
 	ai_process_mission_orders( objnum, aip );
 
 	//	Avoid a shockwave, if necessary.  If a shockwave and rearming, stop rearming.
-	if (aip->ai_flags & AIF_AVOID_SHOCKWAVE) {
+	if (aip->ai_flags[AI::AI_Flags::Avoid_shockwave_ship, AI::AI_Flags::Avoid_shockwave_weapon]) {
 		if (ai_avoid_shockwave(Pl_objp, aip)) {
-			aip->ai_flags &= ~(AIF_BIG_SHIP_COLLIDE_RECOVER_1 | AIF_BIG_SHIP_COLLIDE_RECOVER_2);
-			if (aip->ai_flags & (AIF_AWAITING_REPAIR | AIF_BEING_REPAIRED))
+			aip->ai_flags.remove(AI::AI_Flags::Big_ship_collide_recover_1);
+			aip->ai_flags.set(AI::AI_Flags::Big_ship_collide_recover_2);
```
asarium: The original code removes both flags but the new code only removes one and sets one.

code/parse/sexp.cpp
```diff
@@ -14668,16 +14649,9 @@ void sexp_ingame_ship_kamikaze(ship *shipp, int kdamage)
...
-		aip->ai_flags |= AIF_KAMIKAZE;
-		aip->kamikaze_damage = kdamage;
-	}
-	else
-	{
-		aip->ai_flags &= ~AIF_KAMIKAZE;
-		aip->kamikaze_damage = 0;
-	}
+	aip->ai_flags.set(AI::AI_Flags::Kamikaze, kdamage > 0);
+	aip->kamikaze_damage = kdamage;
```
asarium: This might not behave like the original code since that assigned 0 if the value is `<= 0` but the new version uses the parameter value directly. If the function is never called with a value `< 0` then that should be enforced by an assertion.

The-E: It's actually enforced by the sexp parsing code, but I'll add an assertion.

PR #872, opened 2016-08-22
Add CMake option for generating debuggable executables

asarium:
This allows to debug executables even if they have been compiled as release builds which can help track down issues that don't occur in debug builds.


MageKing17:
Is configuration limited to Release, or does it work with Fast Debug as well?

asarium:
Since Fast Debug is just a copy of the release configuration these options should also apply to Fast Debug but I haven't tested that.

PR #875, opened 2016-08-23
Fix coverity 1361824

niffiwan:
Avoid potential uint overflow if joystick returns a range greater than
the maximum expected range. I guess this could also be fixed by
limiting raw to rng-1 (?)


asarium:
I think the current solution is fine.

PR #887, opened 2016-08-25
[RFC] Refactor parsing code to get rid of global state and allow unit tests

rtoijala:
Here is something I came up with recently. It is very unfinished and I'm not sure if it is a good idea. Please consider it a request for comments.

The parsing code in `code/parse` and the many places that use it operate mostly on the global variable `Mp`, which makes testing difficult. Considering the multiple recent commits and fixes to parsing, this is not a good state of affairs. The aim of this PR is to get rid of the global parsing state, replacing it with state local to the parsing functions, passed as function parameters.

Examples can be seen in the early phases in this PR. The commits are made to be easily reviewed one at a time, each refractoring one function and its call sites and adding what I hope is appropriate tests to the unit test suite in `test/src/parse/test_parselo.cpp`.

Is this something that should be pursued?
Is the status quo good enough?
Is there a better way than what I have come up with?
Is this a waste of time?


The-E:
I am positively inclined to this, on general principles. Removing global state and replacing it with something a bit more local when possible is never a bad thing in my book.

That being said, I'm kinda thinking that this isn't going far enough. The fact that there's a pointer being passed around through the parsing code makes me think that it might be worthwhile to convert this whole thing into a class that loads the file to be parsed, and then acts in much the same way a C# streamreader does; while not functionally different, it is IMHO a much clearer paradigm to structure the code around.

asarium:
I agree with that. Passing the pointer around is an improvement but there is sill some other global state (e.g. the file text) that could be stored in a class variable.

In order to not convert the whole code base at once which would be a nightmare to review you could reduce the changes necessary to external code by using a global parsing state variable that is used by the existing functions. Then we gradually convert all the remaining parsing code to the class.

rtoijala:
I agree that a class-based system would be more intuitive and clear.

I'm not quite sure what you mean by "using a global parsing state variable that is used by the existing functions" or how that could be integrated with the parser class, unless you mean that the class should be developed completely in parallel, leaving the old parsing code as is until the parser classes are ready?

Such an invasive change without an existing test suite is scary, however.

That being said, I still agree that my approach does not go even nearly far enough.

asarium:
I mean that you build the class that has some basic functionality like loading the file content and keeping track of where in the file the parsing currently is. Then, instead of using the various global variables in parselo.cpp you only have the one global parsing class variable which contains all the state. The existing code would simply use this class instead of the old global state.

Once that has been tested sufficiently, you start adding functionality (checking for strings, parsing values, etc...) and make the existing functions (e.g. `required_string`) use those functions for parsing but still keep the global parsing state variable. That way you would have a testable class that is used by the whole code without requiring any changes to the external code.

As the final step, once the parsing class has been sufficiently tested and validated, you convert the rest of the code to use the new parsing class but instead of using a global variable you use a local instance and since the functionality of the class has been tested already you don't need to worry about that. That would be done piece by piece to make reviewing the changes easier.

MageKing17:
I like the idea of class-based parsing; I've had a couple hare-brained ideas for parsing improvements, but they probably wouldn't work with our current global parsing system. Being able to, say, come back to a set of sections of different files at later times could be useful.

Echelon9:
@rtoijala thanks for preparing this series as small, discrete commits. Makes review much easier.

DahBlount:
Any updates on this?

rtoijala:
I haven't had the time to work on this because of university and another project. I still hope to come back to this one day, but even then the approach I'll take will be the one @asarium described above and not the one taken in this PR. I think this can be closed now and I'll open a new PR when/if I have the time to work more on this.

PR #891, opened 2016-08-26
Use 64 bit build for appveyor

asarium:
Nightly builds are still compiled as 32-bit build though.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,7 +9,7 @@ environment:
       CMAKE_GENERATOR: Visual Studio 12 2013
       PlatformToolset: v120_xp
     - VS_VERSION: 14
-      CMAKE_GENERATOR: Visual Studio 14 2015
+      CMAKE_GENERATOR: Visual Studio 14 2015 Win64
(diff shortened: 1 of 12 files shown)


chief1983:
I know it may extend test times a bit but would having both 64 and 32 bit builds tested for now maybe be a good idea?

asarium:
They are already tested. Visual Studio 2013 compiled 32 bit builds, VS 2015 compiles 64-bit builds.

MageKing17:
The 2013 configuration is still 32-bit; I don't think there's much to be gained from having CI do 32/64 of both 2013 and 2015.

chief1983:
But isn't there a chance that compiler differences could lead to warnings in one that don't occur in the other?  Or are any warnings about bit size likely going to be pretty universal?

asarium:
The most warnings will probably happen on the 64-bit configuration and most warnings that happen with 32-bit also happen with 64-bit. Appveyor build times are already pretty high, I would like to avoid increasing that even more.

PR #892, opened 2016-08-26
Remove global my_errno

rtoijala:
This PR removes the global `my_errno` variable. This simplifies the parsing refractor by removing one global.

`my_errno` was used in `atoi2`, `atof2` and `atol2` to return whether the parsing succeeded (`my_errno = 0`) failed (`my_errno = 1`). This is replaced by returning success as a bool return value and the parsed value via pointer parameter.

Also included as a bonus is removing the executable file permission bits from `sexp.h` and `sexp.cpp`.


asarium:
Code looks good and a quick test showed that the code works fine.

PR #893, opened 2016-08-26
Remove is_minimum_GLSL_version()

asarium:
Since OpenGL Core is required now that function would always return true
so it can be removed.

### DIFF
--- a/code/graphics/gropengl.cpp
+++ b/code/graphics/gropengl.cpp
@@ -464,10 +464,6 @@ void gr_opengl_fog_set(int fog_mode, int r, int g, int b, float fog_near, float 
 
 		return;
 	}
-
-	if ( is_minimum_GLSL_version() && Current_shader != NULL && Current_shader->shader == SDR_TYPE_MODEL ) {
-		return;
-	}
 }
 
 int gr_opengl_set_cull(int cull)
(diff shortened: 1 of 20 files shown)


code/graphics/gropengl.cpp
```diff
@@ -1463,42 +1438,31 @@ static void init_extensions() {
 	GLSL_version = ver;
 
 	// we require a minimum GLSL version
-	if (!is_minimum_GLSL_version()) {
+	if (GLSL_version < MIN_REQUIRED_GLSL_VERSION) {
```
Echelon9: A minor suggestion: you could leave the helper function in, and continue to use it in this `if ()` check. Up to you.

Echelon9:
The name of this PR was a little confusing at first, as on review of the code changes it predominantly removes fixed function rendering code -- rather than the helper function.

I made a small comment within about one suggestion.

code/graphics/gropengl.cpp
asarium: That would be the only location where that function could be used so I'll keep the current version.

asarium:
Initially all I did was remove that function but then I noticed that there were a couple of empty function so I also removed the remaining fixed function code.

code/graphics/gropengl.cpp
Echelon9: Fine by me.

PR #895, opened 2016-08-26
Add OpenGL debug group support

asarium:
This uses GL_KHR_debug for exposing debug group information. This is
very useful for tools like apitrace where it's used for assiciating
OpenGL commands with a specific application function.

The groups are created using a utility macro which is removed in release
builds so it should have no impact on the release performance.

I have added some initial scopes but that obviously doesn't cover everything.
I'll add new scopes whenever I come across a good point for them.


asarium:
One Appveyor build stalled for some reason but the other passed so that should not be a problem.

PR #896, opened 2016-08-27
Refactor shader handling

asarium:
This fixes a few of the bad design descisions of previous shader
iterations. This uses a dedicated class for managing the resources and
uniform locations of a shader program.

Uniforms are also stored in this object which makes a lot more sense
than storing them in the OpenGL state where they had to be reset every
time a new shader was used.

There are still a lot of possible enhancments to make, e.g. by using
uniform blocks and buffers instead of standard uniforms which would make
uniform location handling easier.

Another optimization that I will investigate is reducing the amount or
different compiled shaders by turning some of the current compile-time
defines into bool uniforms.

I also moved the OpenGL source code into a separate folder since that's already
what is being done by the Visual Studio solution.


SamuelCho:
I hope you're not saying my contributions to the shader handling code were part of those "bad design decisions". I'd like to think my additions to the shader code aren't just a result of bad judgement just because I didn't feel the need to refactor that code. I mean, yes, the uniform handling code's tracking only the current shader is a limitation but I felt it really wasn't worth the time to extend it as my goal was getting performance improvements along with implementing deferred lighting and shadows as the state changes between re-binding uniforms between shader switches were largely amortized thanks to the state sorting I implemented. And I worried about the amount of uniform data we were storing and just keeping the current shader seemed to largely ameliorate that worry. 

So I mean, I largely think this is a welcome change. Maybe I just have a big ego or maybe I'm too sensitive, but the language you put forward there bothers me.

asarium:
The uniform handling wasn't ideal but I could recycle most of it in the new code. I don't know who wrote the code but the general shader management code (the code in `gropenglshader.cpp`) is in pretty bad shape which makes it hard to do improvements so I need to clean up that code before I can investigate any possible enhancements (e.g. using uniforms instead of `#defines` for altering shader behavior).

MageKing17:
Does this impact performance (positively or negatively), or mainly just make it easier to keep track of things?

asarium:
The main point was the improvement of how shader resources were handled but there may be a few minor performance improvements since the previous code used a linear search for uniform location lookup and now the code uses a hash map for that.

asarium:
Conflicts have been resolved.

PR #901, opened 2016-08-28
OpenGL state tracking cleanup

SamuelCho:
Did some cleaning. Removed a bunch of fixed function state tracking code that was left behind following the OpenGL Core migration.


SamuelCho:
I should also note that I noticed some strange depth buffering issues. It turns out it was related to how we were tracking zbuffering state. This PR fixes it.

PR #903, opened 2016-08-28
Better Shields

SamuelCho:
At the lovely Spoon's request I looked into some various bugs and rendering limitations of the shield code. The outdated per-triangle method of rendering shields is replaced with a decal method that will be used if the shield detail level is at 5. I also looked into some collision detection bugs involving shields such as shield meshes registering a hit long before the projectile ever reaches the target. There was also a bug involving projectiles phasing through shields which is caused by inaccurate shield mesh normals. That has been fixed by doing a simple preprocessing of shield vertex data.

Oh, and I bumped up the shield impact code to 200 now that these changes allow the code to handle that volume.


code/graphics/gropenglshader.cpp
```diff
@@ -111,10 +111,14 @@ static opengl_shader_type_t GL_shader_types[] = {
...
-		{ opengl_vert_attrib::POSITION, opengl_vert_attrib::TEXCOORD, opengl_vert_attrib::COLOR }, "Passthrough" }
+		{ opengl_vert_attrib::POSITION, opengl_vert_attrib::TEXCOORD, opengl_vert_attrib::COLOR }, "Passthrough" },
+
+	{ SDR_TYPE_SHIELD_DECAL, "shield-impact-v.sdr",	"shield-impact-f.sdr", 0,{ 0, 0, 0 },
+		{ "modelViewMatrix", "projMatrix", "shieldMap", "shieldModelViewMatrix", "shieldProjMatrix", "hitNormal", "srgb", "color" }, 
+		{ opengl_vert_attrib::POSITION, opengl_vert_attrib::NORMAL }, "Shield Decals" }
 };
 
-/**
+/**,
```
asarium: I think this comma was accidentally added.

PR #904, opened 2016-08-29
Remove a crapload of unused rendering code

SamuelCho:
Removed a bunch of old OpenGL and general purpose rendering code we've deprecated following the OpenGL Core migration. Feels good man.


SamuelCho:
Whoops, this needs to be cleaned up a bit more. Standby.

SamuelCho:
Never mind it's all good. I thought there was a problem with FRED but it looks like the problem was introduced earlier than this PR.

Issue #889, opened 2016-08-26
Weirdness in the ship lab when bloom_intensity is 0


The-E:
I haven't tested this in mission yet, but here's a bit of weirdness:
When you are in the lab looking at a model and use the n key to reduce bloom intensity to 0, strange things occur. It looks like whatever is being rendered is getting rendered accumulatively, leading to an eventual white-out where the model used to be.

SamuelCho:
That sounds awesome.

The-E:
It actually kinda is. Still though, not exactly intended behaviour :)

MageKing17:
Screenshot DahBlount took of 0 bloom intensity:
![Screenshot](http://puu.sh/qNzcE/cc033fe871.png)

The-E:
I should add that this behaviour is also present in #844. A fix can therefore wait for (or be done as part of) that PR.

SamuelCho:
It looks like when bloom intensity is zero, the bloom pass function doesn't even run. Looks to be a simple fix but it'll probably happen until after the OpenGL Core PR gets merged.

PR #906, opened 2016-08-29
Fixed bloom intensity weirdness when set it's set to 0.

SamuelCho:
Removed unneeded bloom compositing code in the main post processing shader. Cleaned up the bloom compositing shader. Fixes https://github.com/scp-fs2open/fs2open.github.com/issues/889

### DIFF
--- a/code/graphics/gropenglpostprocessing.cpp
+++ b/code/graphics/gropenglpostprocessing.cpp
@@ -93,10 +93,6 @@ void opengl_post_pass_tonemap()
 
 bool opengl_post_pass_bloom()
 {
-	if (Cmdline_bloom_intensity <= 0) {
-		return false;
-	}
-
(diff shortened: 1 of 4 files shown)


code/graphics/gropenglpostprocessing.cpp
```diff
@@ -408,36 +404,14 @@ void gr_opengl_post_process_end()
 		}
 	}
 
-	// bloom uniforms, but only if we did the bloom
-	if (bloomed) {
```
MageKing17: AppVeyor is complaining because this was the only usage of `bloomed` (set on line 357 above). If `opengl_post_pass_bloom()`'s return value is going to be ignored, maybe it shouldn't return anything?

SamuelCho:
appveyor is a harsh mistress

PR #907, opened 2016-08-29
Small imporvements to debug messages

rtoijala:
Add missing new line to BSP Parse timing.

Log value of HUD_config.rp_dist when it is not valid.
I'm hitting this one with rp_dist=-1, I'll try to debug it later.

### DIFF
--- a/code/model/modelinterp.cpp
+++ b/code/model/modelinterp.cpp
@@ -2356,7 +2356,7 @@ void interp_configure_vertex_buffers(polymodel *pm, int mn)
 
 	int time_elapsed = timer_get_milliseconds() - milliseconds;
 
-	mprintf(("BSP Parse took %d milliseconds.", time_elapsed));
+	nprintf(("Model", "BSP Parse took %d milliseconds.\n", time_elapsed));
 
 	if (total_verts < 1) {
 		return;
--- a/code/pilotfile/plr.cpp
+++ b/code/pilotfile/plr.cpp
@@ -139,6 +139,8 @@ void pilotfile::plr_read_hud()
 
 	HUD_config.rp_flags = cfread_int(cfp);
 	HUD_config.rp_dist = cfread_int(cfp);
+	if (HUD_config.rp_dist < 0 || HUD_config.rp_dist >= RR_MAX_RANGES)
+		Error(LOCATION, "Player file has invalid radar range %d\n", HUD_config.rp_dist);
 
 	// basic colors
 	HUD_config.main_color = cfread_int(cfp);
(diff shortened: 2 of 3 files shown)


code/model/modelinterp.cpp
```diff
@@ -2613,7 +2613,7 @@ void interp_configure_vertex_buffers(polymodel *pm, int mn)
 
 	int time_elapsed = timer_get_milliseconds() - milliseconds;
 
-	mprintf(("BSP Parse took %d milliseconds.", time_elapsed));
+	mprintf(("BSP Parse took %d milliseconds.\n", time_elapsed));
```
The-E: This could very well be hidden behind an nprintf filter. It's information that is not generally useful, and causes quite a bit of clutter if large models are being loaded.

rtoijala: Right, should that be +Model or +ModelAnim, or something totally different?

code/radar/radarsetup.cpp
```diff
@@ -484,6 +484,7 @@ void HudGaugeRadar::drawRange()
 		break;
 
 	default:
+		mprintf(("HUD_config.rp_dist=%d\n", HUD_config.rp_dist));
 		Int3();	// can't happen (get Alan if it does)
```
The-E: This should be turned into an Error, but this is perhaps the wrong place to do it. This value is only set to something unsafe when it is being read from a savefile, any error checking should take place there.

code/model/modelinterp.cpp
The-E: Model is perhaps the most appropriate one.

code/pilotfile/plr.cpp
```diff
@@ -139,6 +139,8 @@ void pilotfile::plr_read_hud()
 
 	HUD_config.rp_flags = cfread_int(cfp);
 	HUD_config.rp_dist = cfread_int(cfp);
+	if (HUD_config.rp_dist < 0 || HUD_config.rp_dist >= RR_MAX_RANGES)
+		Error(LOCATION, "Player file has invalid radar range %d\n", HUD_config.rp_dist);
```
rtoijala: All the other failures in the plr parser are treated as non-fatal, so I'm not sure if this should, too.
Maybe a `Warning`?

The-E: Honestly, I believe they should be. If we can't get intelligible data out of the savefile, then we can't trust that the stuff we were able to read is accurate.

rtoijala: In that case, is this PR fine or is there  still something to change?

The-E: Nope, looks good.

PR #911, opened 2016-08-29
Create data path for log file

rtoijala:
Opening the log file failed when the data path did not exists, e.g.
during the user's first run. Create it before trying to open the log
file.

Also, include `<direct.h>` in windows_stubs/config.h since the equivalent
stubs for Unix are also declared there.


asarium:
At some point we need a central header for defining or including stuff like `_mkdir` but that is out of scope of this PR.

PR #912, opened 2016-08-29
Add support for outputting profiling data as JSON

asarium:
This output is compatible with the Chromium tracing viewer which
provides a very nice visual overview of what FSO does in a frame.

Here is an example of how it looks like: chrometracing

The files are written to a directory called "tracing" where each frame is written into a separate file. Here is an example file that you can use directly in the trace viewer: [frame_354.txt](https://github.com/scp-fs2open/fs2open.github.com/files/442507/frame_354.txt) (You have to change the extension to json).

You can also view these files without Chromium since the [Trace-viewer](https://github.com/catapult-project/catapult/blob/master/tracing/README.md) can also be used as a standalone script.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -238,6 +238,7 @@ Flag exe_params[] =
 	{ "-benchmark_mode",	"Puts the game into benchmark mode",		true,	0,					EASY_DEFAULT,		"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-benchmark_mode", },
 	{ "-noninteractive",	"Disables interactive dialogs",				true,	0,					EASY_DEFAULT,		"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-noninteractive", },
 	{ "-json_pilot",		"Dump pilot files in JSON format",			true,	0,					EASY_DEFAULT,		"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-json_pilot", },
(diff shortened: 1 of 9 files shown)


asarium:
Enabling the option has a rather significant performance impact so I'll see if I can move the file writing into a background thread.

asarium:
I moved the file writing into a background thread but that only improved the situation by a little but I think it's good enough since the option isn't turned on my default.

The-E:
Wouldn't it make sense to cache the log data and write it all out on pause or mission end?

asarium:
My initial approach was to dump it all at mission end but that generated huge files that the trace viewer couldn't handle anymore. I'm currently investigating other solutions but I could cache the jansson structs and write them individually at mission end. That may improve performance.

asarium:
After some profiling I found that the jansson functions are what eat up most of the performance so it's not an IO problem. With that in mind I'll move all jansson function calls to the end of the frame so that it doesn't impact the profiling measurements. If I keep using `std::async` I may even be possible to hide most of the overhead in another thread which would be nice.

asarium:
I replaced all the JSON encoding with a manual string output in a background thread which improves the performance impact. It could be improved by using only a single background thread but I need the boost synchronized bounded queue for implementing that so I just used the quick-and-dirty solution for now. I'll revisit it once we have access to boost.

SamuelCho:
Do we really need boost for this feature? I know a lot of game developers who prefer to keep boost out of their games because of how much bloat it brings into projects. Granted, we're not a AAA developer but this feels like hoping for a sledgehammer to drive a nail.

asarium:
We currently don't need it and the current solution works fine but boost has a synchronized bounded queue which would help with implementing a more efficient solution for moving the trace writing into a different thread.

The-E:
I would agree that importing boost just for this isn't necessary. That said, LGTM.

asarium:
Just to clarify, I wasn't suggesting that Boost should be imported just for this PR but that boost has features that would make the implementation more efficient.

PR #913, opened 2016-08-29
Fix writing binary pilot files.

rtoijala:
startSectionWrite would push unnamed sections on _sectionOffsets, but
endSectionWrite was not removing them, leading to _sectionOffsets
containing tens of unnamed sections when parsing the HUD config.

Pop the unnamed sections in endSectionWrite.

This was causing interesting issues such as a crash when starting a mission because of an invalid radar range as well as all controls being unassigned.


asarium:
Thanks for spotting this. The code looks good so I'll merge this once CI passes.

PR #915, opened 2016-08-30
Fix crash/Error in techmenu with 0 intel entries.

rtoijala:
`tech_render_anim` did not check whether there were any intel/weapon
entries before trying to read from`Current_list`, which means that it
read at `Current_list[-1]`, which is a Bad Thing (TM).

Do not try to render an anim if there is no entry.

### DIFF
--- a/code/menuui/techmenu.cpp
+++ b/code/menuui/techmenu.cpp
@@ -713,6 +713,10 @@ void techroom_anim_render(float frametime)
 	// render common stuff
 	tech_common_render();
 
+	// exit now if there are no entries to show
+	if (Current_list_size == 0 || Cur_entry < 0)
+		return;
+
 	// render the animation
 	if(Current_list[Cur_entry].animation.num_frames > 0)
 	{


asarium:
When does this happen? It's always a good thing to validate parameters but if the code calling this function is using the wrong parameters then it might be better to fix that code and use an Assertion here.

If that is not the case then the code is good to go.

rtoijala:
It happens when overriding `species.tbl` with one that does not contain any intel entries, i.e. an empty file. Also, when `species.tbl` is completely missing (which IMHO should be treated like an empty file). Also, when `species.tbl` contains only entries that are not displayed by default (and none have been unlocked).

tl;dr when there are no entries to show.

`techroom_do_frame` calls `techroom_anim_render` unconditionally when the intel tab is selected. If there are no intel entries, `Current_list_size` is 0 and `Cur_entry` is -1.

asarium:
Ok, that sounds like it's not a bug or data error.

PR #923, opened 2016-09-01
Refactored material handling of shaders

SamuelCho:
This was something I planned on doing before the OpenGL Core migration but I got sidetracked. The way materials decide what shader a draw call was left in an overly complicated state and wasn't very clear. I simplified it so that a material just needs an override of the get_shader_flags() if a shader type of a derived material class needs a flag for its particular case.


asarium:
The Mac Travis CI build fails some times with an obscure error. I restarted the build.

PR #924, opened 2016-09-01
Introduce tests for parsing species.tbl

rtoijala:
This adds unit tests (or maybe these count more as functional tests) for
parsing the intelligence data displayed in the tech room. It is parsed
from `species.tbl` by `code/menuui/techroom.cpp` and stored in the global
array `Intel_info`. The number of entries is saved in the global int
`Integer_info_size`.

I chose `species.tbl` as the first tbl to test since it is perhaps the simplest.

Every test has its own mod dir with a `species.tbl` that it parses by
calling `techroom_intel_init()`. The test fixture resets the intel state
with the new `techroom_intel_reset()` between tests.

These tests assume that the file is preprocessed correctly, so it does
not try to test if e.g. comments are removed correctly, ... Only the
parsing that is unique to `species.tbl` is tested.

The last second to last test (`missing_end_multi_text`), which tests what happens when an entry is missing a $end_multi_text, passes or fails randomly when the tests are run with `bin/unittests --gtest_shuffle`. That means there is something wrong with the parser.
I narrowed it down to `parselo.cpp`, where `stuff_string(..., F_MULTITEXT,...)` calls `copy_text_until(..., "$end_multi_text")`, which calls `stristr`, which does not respect `EOF_CHAR` ending the file. `copy_text_until` then continues to copy past `EOF_CHAR` and into the file that happens to be next in memory.
However, touching `stristr` or `copy_text_until` is not something I would like to do until we have extensive tests for parsing all the tbls.


code/menuui/techmenu.cpp
```diff
@@ -246,6 +246,7 @@ static UI_SLIDER2 Tech_slider;
 // Intelligence master data structs (these get inited @ game startup from species.tbl)
 intel_data Intel_info[MAX_INTEL_ENTRIES];
 int Intel_info_size = 0;
+static int intel_info_init_done = 0;
```
asarium: This should be a `bool` variable.

test/src/menuui/test_intel_parse.cpp
```diff
@@ -0,0 +1,307 @@
...
+};
+static intel_data expected_baz = {
+	"Baz name", // XSTR id 3004
+	"Baz desc", // XSTR id 3005
+	"Baz anim",
+	IIF_IN_TECH_DATABASE | IIF_DEFAULT_IN_TECH_DATABASE
+};
+
+// use SCOPED_TRACE("<test name>") before calling this
+static void test_intel_data_equal(intel_data& i1, intel_data& i2)
```
asarium: The parameters should use `const` since you don't change the values.

test/src/menuui/test_intel_parse.cpp
```diff
@@ -0,0 +1,307 @@
...
+	techroom_intel_init();
+
+	ASSERT_EQ(Intel_info_size, 1);
+
+	test_intel_data_equal(expected_foo, Intel_info[0]);
+}
+
+// The same, but with translation to German.
+TEST_F(IntelParseTest, single_translate) {
+	SCOPED_TRACE("single_translate");
```
asarium: I don't think you need a scoped trace if the message is the same as the unit test name.

test/src/menuui/test_intel_parse.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+	test_intel_data_equal(expected_foo, Intel_info[0]);
+}
+
+// A valid entry followed by one with a missing $Name: line.
+// The second entry should be ignored. Parsing should stop there.
+TEST_F(IntelParseTest, missing_name) {
+	SCOPED_TRACE("missing_name");
+
+	// this might throw, but we don't want to make that contractual
```
asarium: Why is the exception not part of the test?

test/src/menuui/test_intel_parse.cpp
rtoijala: Will change.

code/menuui/techmenu.cpp
rtoijala: Will change.

asarium: Thanks.

test/src/menuui/test_intel_parse.cpp
rtoijala: I considered explicitly testing for it with `EXPECT_THROWS`, but then I thought that throwing here might not ultimately be what we want. Malformed intelligence data should IMHO not be a fatal error, but rather something that can be logged, and maybe a warning displayed.
So basically, I left it out of the test so that it can be changed later. The calling code should not have to care whether the table file was malformed or not, it should just assume everything went fine.

test/src/menuui/test_intel_parse.cpp
asarium: Ah, ok. I misunderstood for what you needed them.

rtoijala:
Force pushed version 2.

- Changed `intel_info_init_done` to bool.
- Made parameters to `test_intel_data_equal` const.
- Added another test `always_techroom_values`.

test/src/menuui/test_intel_parse.cpp
asarium: You should write the test for the code that is used at the moment. If that code may throw exceptions then that should also be tested. If it may not throw exceptions then a thrown exception will make the test fail so you could just remove the try-catch,

Anyway, an exception in a test should never be ignored so you should change the test accordingly.

asarium:
You probably rebased your changes onto an older version of master which pulled in some commits that were already merged. You can try using an interactive rebase and then just discarding the material changes.

rtoijala:
Force pushed version 3.

- Changed try/catch to `EXPECT_ANY_THROW`.

asarium:
I think the unit tests are good enough now. You could improve them by explicitly checking for the right exception type which would probably be an ErrorException since that is thrown when the parser runs into an error but it's not a necessary change.
Tell me if you want to change it, if not then I'll merge it.

rtoijala:
I don't think I'll be changing that, so you can merge this. Thanks.

PR #927, opened 2016-09-02
Add support for ship death particle effect

asarium:
This should allow to create effects similar to my particle scripts but without all the performance issues.

I did some refactoring around the different effect shapes and moved the common functionality into a template class where the shape can be customized with a template parameter. That reduces the amount of virtual function calls since the custom shape is known at compile time.

I also fixed an issue with texture page-in.


asarium:
Don't merge this yet, I found some issues with the shape the particles are generated in.

asarium:
Shape has been fixed, now it should be working properly.

PR #928, opened 2016-09-02
Make shaders use uniforms instead of compile time flags

asarium:
This makes certain shader variant flags use boolean uniforms instead of compile time flags which can improve performance since switching shaders is more expensive than setting a uniform. Here is some performance data I collected:
figure_1

There are some changes but nothing major.


asarium:
I ran some tests on Linux and apparently my changes hurt performance pretty severely:
figure_1

So I guess I can close this now...

PR #929, opened 2016-09-02
Replace texture2D with texture in shaders

asarium:
texture2D has been deprecated for some time and Mac also doesn't support
it anymore. The 1.50 GLSL specification says that texture is the correct
function for texel lookup so I used that.


chief1983:
This runs fine on my Mac, with a bypass for the powermeter crash in place.

SamuelCho:
Strange, I thought I converted all the texture2D calls to texture.

PR #932, opened 2016-09-03
Use aligned storage for the memory header

asarium:
This should ensure that the pointer returned by vm_malloc is suitably
aligned for any type.

@chief1983 Could you test this on mac? This may fix the issues with our custom `new` operator without removing it.


chief1983:
I get a different error now (actually seems closer to the one I added to the other issue).

```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
abort() called
*** error for object 0x7fc6cb8319e0: pointer being freed was not allocated
 

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff973d2f06 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff8ca474ec pthread_kill + 90
2   libsystem_c.dylib             	0x00007fff936bd6df abort + 129
3   libsystem_malloc.dylib        	0x00007fff878b1041 free + 425
4   fs2_open_3_7_5_x64-DEBUG      	0x00000001049c488c _vm_free(void*, char const*, int) + 92
5   fs2_open_3_7_5_x64-DEBUG      	0x00000001049cc3bd operator delete(void*) + 29
6   com.apple.audio.OpenAL        	0x000000010655de81 OALCaptureDevice::~OALCaptureDevice() + 41
7   com.apple.audio.OpenAL        	0x0000000106543d99 alcCaptureCloseDevice + 226
8   fs2_open_3_7_5_x64-DEBUG      	0x00000001050969d2 find_capture_device() + 2850
9   fs2_open_3_7_5_x64-DEBUG      	0x0000000105094857 openal_init_device(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*) + 775
10  fs2_open_3_7_5_x64-DEBUG      	0x00000001050893fd ds_init() + 813
11  fs2_open_3_7_5_x64-DEBUG      	0x000000010509ad9f snd_init() + 95
12  fs2_open_3_7_5_x64-DEBUG      	0x000000010487ed8c game_init() + 748
13  fs2_open_3_7_5_x64-DEBUG      	0x000000010488c3e0 game_main(int, char**) + 160
14  fs2_open_3_7_5_x64-DEBUG      	0x000000010488dfff actual_main(int, char**) + 527
15  fs2_open_3_7_5_x64-DEBUG      	0x000000010488e1c2 main + 34
16  libdyld.dylib                 	0x00007fff877fc5ad start + 1
```

I could maybe try to figure out an ASan build and see if that gives any more info.

asarium:
There was an issue with my existing code. I changed the memory header code which may fix your issue.

chief1983:
Still seeing the same crash as above after updating and building from scratch.

asarium:
I have no idea what else could go wrong. I guess removing newdelete.cpp is the only solution right now.

asarium:
It looks like I finally got this to compile on all platforms. Even though it doesn't fix the Mac issue it's still an improvement since it should improve the alignment of the pointer returned by our `malloc`.

chief1983:
Just a bit ago, before this was merged into master, FreeBSD was getting a SIGKILL and terminating itself.  Now it gets a segmentation fault.

chief1983:
(both stacktraces in gdb were going through newdelete before it happened)

asarium:
Is AddressSanitizer available on FreeBSD? That might give us a clue as to what is going wrong.

chief1983:
It is since clang supported it before gcc, but right now cmake can only enable it for gcc.

asarium:
The CMake code for clang should be pretty much the same.

chief1983:
I thought ASan was supported on FreeBSD since it has clang, but the OS itself doesn't have compatibility with it yet.  But the PR to get it added for clang will allow using it on OS X I believe so at least it's helpful.

asarium:
Apple has a special version of clang so the added option will not affect XCode projects.

chief1983:
Hmm, I guess it would need to enable it a different way then.  Oh well.  It'll work for Linux/clang?  :P

asarium:
I have tested it on Fedora 24 and there it works.

PR #934, opened 2016-09-03
Make sure that the parse strings are c strings

### LINKED ISSUE #933: Mission (and table) text data is not a C-string
I have been investigating why the unit tests fail with a valgrind error which lead me to `stristr` which reads out of the boundaries of the passed string. @rtoijala already described this issue in #924. The same issue is also detected by AddressSanitizer.

I found that `process_raw_file_text` writes `EOF_CHAR` at the end of the parsed strings but not a null character which means that, as far as I can tell, the string in the parse code is not a C-string and may not be used by the various string manipulation functions. Since not using those functions is not really an option the parse code has to be modified so that a null character is appended to the processed text.

asarium:
Fixes #933.

This also uses `size_t` for keeping track of he size of the memory.

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -50,8 +50,8 @@ const char	*token_found;
 static int Parsing_paused = 0;
 
 // text allocation stuff
-void allocate_mission_text(int size);
-static int Mission_text_size = 0;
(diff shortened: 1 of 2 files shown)


Goober5000:
LGTM.

PR #938, opened 2016-09-04
Add shader caching

asarium:
This uses the program binary support included in OpenGL 4.1 to build a
cache of compiled shaders. It does this by computing the MD5 of the
source code of the shaders and using that hash to generate unique file names. That
should make sure that a change to a shader is automatically reflected in
the shader cache.

Currently the shader files are written to a mod specific directory. If that's a problem then I could try figuring out how to write them into the root directory but that would probably involve hacking the cfile system and I would like to avoid that.

Also, in my test runs I found that we use 167 (!!) shaders. It might be a good idea to somehow reduce that.

### DIFF
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -33,6 +33,8 @@ TARGET_LINK_LIBRARIES(code PUBLIC ${SDL_LIBS})
 TARGET_LINK_LIBRARIES(code PUBLIC platform)
 TARGET_LINK_LIBRARIES(code PUBLIC compiler)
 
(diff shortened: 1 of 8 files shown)


SamuelCho:
Commercial games ship with way more shader permutations than that. The most recent game I shipped had like 500 even though it's a game with smaller scope than Freespace. Reducing shader permutations doesn't seem like a priority to me.

asarium:
Interesting. I saw that there were a lot of cached shaders and was a bit surprised that we use that many permutation.

PR #943, opened 2016-09-06
Fix nightly build configuration

asarium:


chief1983:
Let's give this a shot, I'll kick off today's nightly script again.  Better late than never.

PR #946, opened 2016-09-08
Actually initialize Ignore_List.

MageKing17:
`Ignore_List` (the list of flags used to determine if a ship is an invalid target) was originally a statically-initialized bitfield stored in an `int`; the new bitset flags code added a `set_default_ignore_list()` function to provide the default values, but then forgot to call it anywhere. As a result, "hidden-from-sensors" (and the other, less-likely-to-be-noticed flags) was having no effect on the player's ability to target anything. This adds a call to `set_default_ignore_list()` to `ship_init()` so that the original functionality is restored.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -5226,6 +5226,9 @@ void ship_init()
 {
 	if ( !ships_inited )
 	{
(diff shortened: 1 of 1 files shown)


asarium:
LGTM.

PR #947, opened 2016-09-08
Some miscelleaneous cleanup commits.

MageKing17:
None of these should be functional changes, but give them a look over in case I missed something.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -616,6 +616,9 @@ void ai_profile_t::reset()
         delay_bomb_arm_timer[i] = 0;
         chance_to_use_missiles_on_plr[i] = 0;
         player_autoaim_fov[i] = 0;
+    }
+
+    for (int i = 0; i <= MAX_DETAIL_LEVEL; ++i) {
         detail_distance_mult[i] = 0;
     }
 }
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -345,7 +345,7 @@ void parse_ai_profiles_tbl(const char *filename)
 				}
 
 				if (optional_string("$Detail Distance Multiplier:"))
-					parse_float_list(profile->detail_distance_mult, NUM_SKILL_LEVELS);
+					parse_float_list(profile->detail_distance_mult, MAX_DETAIL_LEVEL + 1);
 
 				set_flag(profile, "$big ships can attack beam turrets on untargeted ships:", AI::Profile_Flags::Big_ships_can_attack_beam_turrets_on_untargeted_ships);
 
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -1916,7 +1916,7 @@ void asteroid_parse_tbl()
 	char impact_ani_file[MAX_FILENAME_LEN];
 
 	// How did we get here without having any species defined?
-	Assertion(Species_info.size() > 0,
+	Assertion(!Species_info.empty(),
 		"Cannot parse asteroids/debris if there "
 		"are no species for them to belong to."
 		);
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -608,7 +608,7 @@ static void mkdir_recursive(const char *path) {
     while ((pos = tmp.find_first_of(DIR_SEPARATOR_CHAR, pre)) != std::string::npos) {
         dir = tmp.substr(0, pos++);
         pre = pos;
-        if (dir.size() == 0) continue; // if leading / first time is 0 length
+        if (dir.empty()) continue; // if leading / first time is 0 length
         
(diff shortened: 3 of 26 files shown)


Goober5000:
I reviewed the patch.  All of it LGTM, except the detail distance multiplier.  I would back that out of this PR and make a separate PR just for that -- and not only the parsing, but everywhere else it's used.  It really should be 4, not 5 (as one of the comments says), and there are some bizarre code chunks caused by making it dependent on the skill level.

MageKing17:
> All of it LGTM, except the detail distance multiplier. I would back that out of this PR and make a separate PR just for that -- and not only the parsing, but everywhere else it's used.

I'm unaware of anywhere else in the code `NUM_SKILL_LEVELS` is used for detail distance.

> It really should be 4, not 5 (as one of the comments says)

`MAX_DETAIL_LEVEL` *is* 4, though?

Goober5000:
> I'm unaware of anywhere else in the code `NUM_SKILL_LEVELS` is used for detail distance.

There is no such association.  There is some initialization code in ai_profile_t::reset(), but that's one of the "bizarre code chunks caused by the dependency" that I referred to.  There are enough of them to warrant a separate PR.

(Perhaps I'm not being clear.  I want to fix all the places where `detail_distance_mult` is either set to 5 or associated with `NUM_SKILL_LEVELS` or both.)

> `MAX_DETAIL_LEVEL` is 4, though?

Yes, and `MAX_DETAIL_LEVEL+1` is 5.  I was referring to the size of the detail distance multiplier array.

MageKing17:
Shit, I meant to reply to @Goober5000 and edited his post instead; anyone know a way to revert an edited comment?

MageKing17:
> (Perhaps I'm not being clear.  I want to fix all the places where `detail_distance_mult` is either set to 5 or associated with `NUM_SKILL_LEVELS` or both.)

~~And I'm saying I don't know of any other places like that in the source; this is the only one I noticed.~~

Okay, so there's `ai_profile_t::reset()` which you already mentioned... and that's it. It turns out `detail_distance_mult` is referenced in very few places in the source code; pretty much just when parsed, when reset, the array's declaration in the header, and when it's used in the rendering code. That's it. I'll just add another commit to fix `ai_profile_t::reset()`, but I don't see the need for it to be a separate PR because there's just not that much code.

> Yes, and `MAX_DETAIL_LEVEL+1` is 5.  I was referring to the size of the detail distance multiplier array.

The size of the detail distance multiplier array should not be 4 because there are 5 potential values: one for each valid detail level (which can be 0, 1, 2, 3, or 4), so the maximum detail level is 4 and the total number of values is 5. Hence why `MAX_DETAIL_LEVEL` is 4 and why `MAX_DETAIL_LEVEL + 1` is used as the size of this array.

Goober5000:
If you're putting them all in this PR, then you should also include the chunk in `model_render_determine_depth` where the code uses a case statement for each number between 0 and 4.  That's the other bizarre chunk I was referring to.

And I withdraw my comments about `MAX_DETAIL_LEVEL`.  Due to comparing it with `MAX_SHIP_DETAIL_LEVELS`, I thought we had a case of two differently-sized arrays.  Looking at the original public source, however, makes it clear that `MAX_DETAIL_LEVEL` refers to the highest *index* of the array, whereas `MAX_SHIP_DETAIL_LEVELS` refers to the *length* of the array.  Confusing without looking at it closely.  And then there is `NUM_DEFAULT_DETAIL_LEVELS` which is something completely different, despite having a very similar name.

(If one were to create a separate PR, one might change `MAX_SHIP_DETAIL_LEVELS` to `NUM_SHIP_DETAIL_LEVELS` for clarity, along with the similarly-named `#define` for asteroids.)

MageKing17:
Upon further review, there's a *lot* about `model_render_determine_depth()` and `model_render_determine_detail()` that don't make sense (why do we bother storing the first "$Detail Distance:" value if we're just going to ignore it? Why do we even bother calling `model_render_determine_depth()` if `detail_level_locked` is `>= 0`? ...Oh, because we need the value to compare against `weapon_model_draw_distance`, but if the detail level is locked, shouldn't that be forced on or something?).

While that particular case statement is one of those oddities, it's not actually an instance of being dependent on the *wrong* value (if somebody added a 6th difficulty level, for instance, they wouldn't need to touch that code at all), so I think a separate PR for cleaning up those modelrender.cpp oddities would make a lot of sense (I just don't feel comfortable adding that as a commit on this PR because my understanding of graphics-related code, even that not directly related to the actual rendering itself, is severely lacking).

Goober5000:
Yeah, those are good points.  I'm not familiar enough with the graphics code to clean up those oddities either.  I suppose we can punt that to a future graphics developer.  There's nothing there that's explicitly wrong, just possibly not as organized as it could be.

I'll merge the PR, since the `ai_profile_t::reset()` commit was added and I reviewed the rest of the code earlier.

PR #951, opened 2016-09-08
Fix cursor handling in standalone server

### LINKED ISSUE #941: Standalone will not run
The standalone won't actually run anymore.  No crash, seg faults, etc.  Even when compiled with ASan or run under GDB, there's really no information.  Just that it exited with code 01.  This occurs on Linux (gcc) and FreeBSD (clang), both tested with debug builds.  I had to comment out the newdelete.cpp on FreeBSD as current master as of this posting leads to a segmentation fault otherwise.  This is the same as the proposed fix for the alignment issues on Mac.  Suspect it is related to clang as that is one of the main common items between the two platforms, but that's outside the scope of this issue.

On windows, it crashes with the error in the below comment.

asarium:
There are a few issues with cursors in standalone mode since we don't have video support there. These changes should take care of that and make sure that the standalone server works again.

@chief1983 Please test this to check if it resolves #941.

### DIFF
--- a/code/graphics/2d.cpp
+++ b/code/graphics/2d.cpp
@@ -1073,16 +1073,27 @@ bool gr_init(os::GraphicsOperations* graphicsOps, int d_mode, int d_width, int d
 	bm_init();
 	io::mouse::CursorManager::init();
 
-	// load the web pointer cursor bitmap
-	if (!running_unittests && Web_cursor == NULL) {
-		//if it still hasn't loaded then this usually means that the executable isn't in the same directory as the main fs2 install
-		if ( (Web_cursor = io::mouse::CursorManager::get()->loadCursor("cursorweb", true)) == NULL ) {
-			Error(LOCATION, "\nWeb cursor bitmap not found.  This is most likely due to one of three reasons:\n"
-				"    1) You're running FreeSpace Open from somewhere other than your FreeSpace 2 folder;\n"
-				"    2) You've somehow corrupted your FreeSpace 2 installation, e.g. by modifying or removing the retail VP files;\n"
-				"    3) You haven't installed FreeSpace 2 at all.  (Note that installing FreeSpace Open does NOT remove the need for a FreeSpace 2 installation.)\n"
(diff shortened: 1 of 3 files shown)


chief1983:
I think this broke actually running the game, on the Mac at least.
```
Error: 
Web cursor bitmap not found.  This is most likely due to one of three reasons:
    1) You're running FreeSpace Open from somewhere other than your FreeSpace 2 folder;
    2) You've somehow corrupted your FreeSpace 2 installation, e.g. by modifying or removing the retail VP files;
    3) You haven't installed FreeSpace 2 at all.  (Note that installing FreeSpace Open does NOT remove the need for a FreeSpace 2 installation.)
Number 1 can be fixed by simply moving the FreeSpace Open executable file to the FreeSpace 2 folder.  Numbers 2 and 3 can be fixed by installing or reinstalling FreeSpace 2.

File: 2d.cpp
Line: 1092

  1   fs2_open_3_7_5_x64-DEBUG            0x000000010dd02d71 _ZN2os7dialogs5ErrorEPKc + 1905
  2   fs2_open_3_7_5_x64-DEBUG            0x000000010dd05abf _ZN2os7dialogs5ErrorEPKciS2_z + 2591
  3   fs2_open_3_7_5_x64-DEBUG            0x000000010d8b5341 _Z7gr_initPN2os18GraphicsOperationsEiiii + 2785
  4   fs2_open_3_7_5_x64-DEBUG            0x000000010d75d2e9 _Z9game_initv + 1337
  5   fs2_open_3_7_5_x64-DEBUG            0x000000010d76a2f0 _Z9game_mainiPPc + 160
  6   fs2_open_3_7_5_x64-DEBUG            0x000000010d76bf0a _Z11actual_mainiPPc + 522
  7   fs2_open_3_7_5_x64-DEBUG            0x000000010d76c0d2 main + 34
  8   libdyld.dylib                       0x00007fff8b11c5ad start + 1
  9   ???                                 0x0000000000000001 0x0 + 1
```

asarium:
This should be fixed in #953. I was pretty sure that I tested that code without the standalone flag but apparently I messed something up...

PR #952, opened 2016-09-08
Remove memory tracking code

asarium:
As discussed on IRC, the memory tracking code isn't needed anymore and caused some issues. This also removes the new and delete operators since they aren't needed anymore.


Goober5000:
@asarium This isn't helpful for people who weren't on IRC.  Can you post a chat log or summary?

The-E:
The reason we decided to remove this was that it causes problems without any clear benefit. When memory tracking is something we need to do, the built-in memory profilers in Visual Studio or its equivalents on other OSs can provide much better information than what we can glean from the single memory counter we have.

PR #958, opened 2016-09-10
Fix lua drawImage rendering

asarium:
This now uses the same material properties that `gr_bitmap` already uses and fixes the alpha blending mode for interface materials.

### DIFF
--- a/code/graphics/2d.cpp
+++ b/code/graphics/2d.cpp
@@ -1315,6 +1315,8 @@ void gr_bitmap(int _x, int _y, int resize_mode)
 
 void gr_bitmap_uv(int _x, int _y, int _w, int _h, float _u0, float _v0, float _u1, float _v1, int resize_mode)
 {
+	GR_DEBUG_SCOPE("2D Bitmap UV");
+
 	float x, y, w, h;
 	vertex verts[4];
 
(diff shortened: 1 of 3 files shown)


asarium:
@AxemP found some issues with these changes so it should not be merged yet.

AxemP:
I made a small mod that tests the three common types of drawing images with scripting. drawImage, streamingAnimation and drawMonochromeImage.

It looks like streamingAnimations aren't receiving that alpha parameter either, it always draws as full opacity even with forcing the opacity type to alphablend_filter. That was niffiwan's department so he might have some insight on how it gets drawn...

https://drive.google.com/file/d/0B_q0tkX7VcSaX1BzcnpNcXpMdTA/view?usp=sharing

asarium:
This is very odd. The generic animation renderer uses `GR_ALPHABLEND_NONE` for the APNGs which means that that behavior must have been broken for a long time.

I went ahead and fixed the APNG alpha blend type and also changed the default lua opacity type to the alpha blend filter which should ensure that scripts that rely on the old behavior continue to work.

AxemP:
I've tested the latest changes and they give the expected behavior now. :)

asarium:
Thanks, this should be ready for code review and merging now.

PR #962, opened 2016-09-11
Add script for doing automated performance test

asarium:
This script allows to compare the performance of a branch to master by
running both versions and generating a graph of frametimes for both
versions.

### DIFF
--- a/scripts/performance_check.sh
+++ b/scripts/performance_check.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# This script tests for performance changes between two branches in this repository.
+
+set -ex
+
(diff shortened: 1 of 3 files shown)


scripts/performance_check.sh
```diff
@@ -0,0 +1,46 @@
...
+    cp "$FSO_DIR/profiling.csv" "/tmp/$NAME.csv"
+
+    popd
+}
+
+#git checkout master
+#run_test master "$@"
+
+#git checkout "${BRANCH}"
+#run_test changed "$@"
```
MageKing17: It seems to me that the script is of very little utility as long as these calls are commented out...

asarium: Fixed, stupid testing code...

PR #964, opened 2016-09-11
Add LTTng Tracing

TheMatthew:
This allows fast in production tracing for performance monitoring.
The information is available in Linux, in a the form of binary CTF
trace.

Tracer information is here: www.lttng.org

Signed-off-by: Matthew Khouzam matthew.khouzam@gmail.com

### DIFF
--- a/cmake/platform-unix.cmake
+++ b/cmake/platform-unix.cmake
@@ -3,4 +3,10 @@ INCLUDE(util)
 
 MESSAGE(STATUS "Configuring UNIX specific things and stuff...")
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/finder")
+
+include (FindLTTngUST REQUIRED)
+
+target_link_libraries(platform INTERFACE LTTng::UST)
+
 target_compile_definitions(platform INTERFACE SCP_UNIX USE_OPENAL)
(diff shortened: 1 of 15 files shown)


cmake/platform-unix.cmake
```diff
@@ -3,4 +3,10 @@ INCLUDE(util)
 
 MESSAGE(STATUS "Configuring UNIX specific things and stuff...")
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/finder")
+
+include (FindLTTngUST REQUIRED)
```
asarium: This library is not a platform library and should be linked directly into the code target instead.

code/graphics/2d.cpp
```diff
@@ -2124,6 +2125,7 @@ void gr_flip()
 	// m!m avoid running CHA_ONFRAME when the "Quit mission" popup is shown. See mantis 2446 for reference
 	if (!quit_mission_popup_shown)
 	{
+		tracepoint(fs2open, LUA_On_Frame__begin);
```
asarium: As I said before on IRC, the two profiling systems should be combined since it's a bit cumbersome to maintain two different kinds of tracing systems if they don't share the same external API.

TheMatthew:
Would it be better initially to add LTTng to the project with no tracepoints and then add them in later patches?

PR #966, opened 2016-09-12
Species tbl test improvements

rtoijala:
Since 16d5f88 and c9f9e1b there is no longer the danger of reading old data from `Mission_text`, so the test `missing_end_multi_text` can now be uncommented.

Also, add a unit test to check that intel entry data must be in the correct order for parsing to succeed.


rtoijala:
Seems Travis failed because one build hanged.
Is this my fault and how can it be fixed?

The-E:
Nah, I'll just restart the build.

Issue #880, opened 2016-08-24
MVE cutscenes play at incorrect aspect ratio


Yarn366:
Since the TrueType merge, MVE cutscenes have been squashed (not cropped) to about 2/3 the correct height, with the width still being correct. The video is still in the "correct" position, leaving extra space below. If the video is started from the Tech Room (and maybe before a mission--I haven't checked), garbage appears in that space.

Resolution doesn't make a difference, and neither does enabling the "Disable scale-to-window for movies" option.

Ogg cutscenes still appear to play fine.

Yarn366:
MVE playback is even more broken now: since OpenGL Core was merged, MVE cutscenes don't display at all, although the sound is still present. That is, unless a preload image is present, in which case the opening cutscene _does_ display (still at incorrect aspect ratio); cutscenes played from the Tech Room are still audio only.

Yarn366:
This is now fixed, likely due to commit 5be84cace3a55fb6d85a977d80823f312b186682.

PR #979, opened 2016-09-19
Add improved video player

asarium:
Uses FFmpeg for video decoding and multiple threads for reading and decoding files. Can handle 1080p with no stuttering.

This change requires the boost libraries and some prebuilt libraries. The prebuilt libraries are downloaded as part of the CMake configuration process.

There are currently no Mac builds of FFmpeg since I have no machine where I could compile those. The CI currently uses the homebrew libraries but someone will need to compile the libraries so we can include those in release builds.


cmake/platform-darwin.cmake
```diff
@@ -15,3 +15,5 @@ SET(CMAKE_OSX_DEPLOYMENT_TARGET "10.9")
 SET(CMAKE_SKIP_RPATH FALSE)
 SET(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
 SET(CMAKE_INSTALL_RPATH @loader_path/../Frameworks/)
+
+set(PLATFORM_MAC TRUE CACHE INTERNAL "" FORCE)
```
chief1983: Should have a newline here at the end of the file.

SamuelCho:
Do we really need Ffmpeg to replace the video player? At the cost of adding a huge library like  boost?

chief1983:
I had also voiced concerns earlier about its size, although not particularly because of boost.  If we're adding boost for anything, there are already several other uses for it.  It's the rest of the bulk I had wondered if there wasn't a smaller option for in this case.

asarium:
FFmpeg is for H.264 decoding since that is superior to the Theora codec. As far as I know, the only reason I need boost is for the `sync_bounded_queue` since I don't want to implement that myself. It is very easy to make mistakes in multi threaded code (I tried, it didn't work) and boost has a working implementation. If you have an alternate solution for the bounded queue I would gladly remove the boost dependency.

asarium:
@chief1983 FFmpeg is a proven solution that is being used in a lot of projects. Implementing our own solution for H.264 decoding via another external library will likely lead to errors since we also need to implement a parser for MP4 files.

code/cutscene/ffmpeg/AudioDecoder.h
```diff
@@ -0,0 +1,35 @@
...
+	explicit AudioDecoder(DecoderStatus* status);
+
+	virtual ~AudioDecoder();
+
+	virtual void decodePacket(AVPacket* packet) SCP_OVERRIDE;
+
+	virtual void finishDecoding() SCP_OVERRIDE;
+};
+}
+}
```
chief1983: Need a newline here too

code/cutscene/ffmpeg/VideoDecoder.h
```diff
@@ -0,0 +1,25 @@
...
+	explicit VideoDecoder(DecoderStatus* status);
+
+	virtual ~VideoDecoder();
+
+	virtual void decodePacket(AVPacket* packet) SCP_OVERRIDE;
+
+	virtual void finishDecoding() SCP_OVERRIDE;
+};
+}
+}
```
chief1983: And a newline here.

code/cutscene/ffmpeg/internal.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+	if (audioCodecCtx != nullptr) {
+		avcodec_close(audioCodecCtx);
+#if LIBAVCODEC_VERSION_INT > AV_VERSION_INT(57, 24, 255)
+		avcodec_free_context(&audioCodecCtx);
+#endif
+		audioCodecCtx = nullptr;
+	}
+}
+}
```
chief1983: When braces stack like this, could we use comments indicating what is closed where to increase clarity?

SamuelCho:
Does ffmpeg require multiple threads to be running? Or is extracting good performance out of video decoding contingent on having multithreading?

asarium:
It's for getting better performance. The current ogg player is single threaded and that is barely able to play a 720p cutscene.

SamuelCho:
I see, but why a synchronized bounded queue? Would a lock free queue not work?

asarium:
I need the ability to limit how many frames are waiting in the queue. That is required since otherwise we would just be wasting memory and CPU time with decoding frames we may not even need.

SamuelCho:
Is there any particular reason why we can't use a smaller library like this one? It's meant for single-producer, single-consumer problems which seems to be the case here. Apparently the author used it for audio processing code, very similar to what this is trying to do. And the queue can be bounded or dynamic. https://github.com/cameron314/readerwriterqueue

asarium:
The library looks promising. I'll do some tests to see if it will work with my code and if it does then I should be able to remove the boost dependency. The only reason I added boost in the first place is because I needed a threading library and `std::thread` wasn't available at that point.

asarium:
@SamuelCho The library you linked to doesn't work for this since it doesn't allow to specify an upper limit for the number of elements. It just allows to specify an initial capacity but if more elements are added the queue grows. I'll continue searching for other alternatives but at the moment it looks like boost is the best option.

SamuelCho:
Try_enqueue didn't work?

asarium:
`try_enqueue` immediately returns if the capacity is reached but I need a function that waits until there is room in the queue. The blocking queue that exists only supports that for the removing elements from the queue.

SamuelCho:
What's stopping you from just putting a condition variable that waits if the queue is full? Sorry if this sounds like backseat driving.

asarium:
Then I would just be reimplementing what the boost queue already does and I don't like duplicating work, especially if it's error-prone like multi threading stuff. If there is a solution that can be used without major changes to the code then I'm willing to make the changes but ignoring an existing solution only to avoid including boost is not very efficient.

asarium:
It's not a minor change since I need to make sure that the variable is signaled when the queue isn't full anymore by the function that removes the frames but it also needs to make sure that there are no edge cases where the synchronization or notification may fail. This all requires effort that I am not willing to invest if there is a perfectly fine solution already available.

The-E:
Way I see it, we want this feature. The best option to implement it, according to the expert on the subject, involves using a boost feature. So, we either use boost, or we do not use this feature.

I say we use this. This discussion has gotten borderline ridiculous, and I have seen no good reasons against the use of boost in it.

SamuelCho:
Okay, sorry for pushing back so hard. We kind of have boost allergy in retail game dev land. It's a hard habit to break.

asarium:
@chief1983 and I are currently working on getting Mac libraries for FFmpeg compiled and integrated into our system. After that is done this should be ready to be merged.

asarium:
The libraries @chief1983 has provided seem to be working fine. They are now used by Travis CI but I can't confirm if the generated executable actually runs.

asarium:
From my perspective, these changes are finished and are ready to be merged. Is there anything else that should be changed or discussed?

chief1983:
Got the trimmed-down ffmpeg lib for Mac uploaded, and confirmed that the boost static linking works fine.  If we can get that lib updated, I don't see any issue with this from a mac standpoint.  Ogg videos are still playing.  Anyone have the MVE files to test on Mac?

asarium:
I'm currently recompiling the Linux and Windows libraries with the updated flags. MVE is uses the old code so there shouldn't be any changes. I'll see if I have the MVE cutscenes to test if they still work.

asarium:
So, it turns out that FFmpeg is actually able to decode the old MVE videos so I replaced the old code with the new FFmpeg decoder. This should also fix #880 since the new display code works regardless of which decoder is used.

chief1983:
This makes a perfectly usable Mac build now, with filesizes not terribly exceeding what we've tolerated in the past when we weren't copying in SDL correctly.  Since no one objected then, no one's likely to now, and I don't think can trim down the new boost/ffmpeg dependencies much more unless we want to pick out individual supported formats from ffmpeg.  So, if someone wants to pass this for general code review, it's got my sign-off.

PR #980, opened 2016-09-20
Build both 32- and 64-bit builds for nightlies

asarium:


MageKing17:
My only concern with this is that I'd rather, if problems popped up with running the 64-bit builds, that they got fixed, rather than people just switching to the 32-bit build...

asarium:
The issues would obviously be fixed but if both variants are available then people could check if the issue is specific to 32- or 64-bit.

MageKing17:
> The issues would obviously be fixed but if both variants are available then people could check if the issue is specific to 32- or 64-bit.

Well, they wouldn't get fixed if somebody used the 64-bit build, encountered a problem, tried the 32-bit build, discovered it worked, and then never remembered to report the original problem... but that's obviously just me looking for trouble, so I guess we might as well go for it.

PR #981, opened 2016-09-20
fix/SDL2_framework: Use an unmodified copy of the SDL2 framework

chief1983:
The current SDL2.framework in the Frameworks archive was created with the symlinks being followed and generating duplicated files, significantly wasting space, while also not being the same as the official framework.  This corrects that by overwriting the one in the archive with the official framework.  It was created with the copyfile bit disabled, so no additional dot files, .DS_Store files, etc are in the archive.


chief1983:
Confirmed it works fine, archived size has dropped from 12 to 10 MB, unextracted size has dropped from 42 to 33 MB.

PR #993, opened 2016-10-01
Enable mingw builds & remove boost dep.

ashdnazg:
- Add libs prebuilt with mxe
- Update cmake scripts to use them
- Make MINGW builds use the generic def_files inclusion method
- Some minor fixes to compilation
- Port boost::sync_bounded_queue and remove the dependency

It's highly probable that at the moment the def_files are included twice in the executable (once as a resource and once in the code), I switched to the generic (code) version since I wanted a minimal-changes-working-PR and since I'm not that familiar with resources.

If you don't trust my prebuilt libs (I won't take it personally) you can compile them yourself on a linux environment (I used windows 10's "bash on windows" thing) using http://mxe.cc/ which is amazingly easy to use.

~~This was coded with the pre-ffmpeg&boost version which will make the process quite more complex.~~ Now up to date with the main repo after I've added syncboundedqueue.h which is a port of boost::sync_bounded_queue to std::* line-for-line.

btw, for some arcane reason, it manages to link with openal32.lib. 
As the saying goes, if it works, don't touch it.


code/network/stand_gui.cpp
```diff
@@ -170,16 +170,16 @@ void std_gen_set_text(char *str, int field_num)
 	}
 
 	// otherwise set the text
-	ctrl = GetDlgItem(Multi_gen_dialog, (int)MAKEINTRESOURCE(IDC_FIELD1));
+	ctrl = GetDlgItem(Multi_gen_dialog, (int)(size_t)MAKEINTRESOURCE(IDC_FIELD1));
```
MageKing17: Why is this double-cast necessary?

ashdnazg: 64 bit mingw, doesn't allow you to cast from a 64 bit pointer to a 32 bit integer. (loss of precision in pointers is a severe issue)
So I first have to cast it to 64 bit integer and only then to 32 bit.

asarium: The change is needed, I needed it before when I tried to do the same. However you should use `uintptr_t` since that is an integer type that can hold a pointer. `size_t` isn't the right type for this.

ashdnazg: k, will do

lib/oggvorbis/lib/x64/libvorbis.la
```diff
@@ -0,0 +1,41 @@
+# libvorbis.la - a libtool library file
```
asarium: Why have you added this file? We don't use libtool so this file should not be necessary.

ashdnazg: it seems I've added it by error

ashdnazg:
The issues pointed out are fixed, and I've also made it up to date with master.

Apparently the easiest way to was to remove the boost dependency.
That's because the linker fails if any two projects using boost (in this case boost, ffmpeg and FS2) were compiled with different compiler versions (release, exception model and threading model all have to be the same).
The other option is to force the use specific mingw configurations (what we do in spring) which in my experience is crap.
As long as you guys can live without boost, it's worth it.

The porting of sync_bounded_queue was pretty straightforward, I didn't have to touch any code inside functions, I've only had to delete deprecated code, expand macros and fix some scope issues.

code/libs/ffmpeg/FFmpegContext.cpp
```diff
@@ -100,9 +98,8 @@ std::unique_ptr<FFmpegContext> FFmpegContext::createContext(CFILE* mediaFile) {
 	if (probe_ret < 0) {
 		char errorStr[1024];
 		av_strerror(probe_ret, errorStr, 1024);
-		auto fmt = boost::format("Could not open movie file! Error: %s") % errorStr;
 
-		throw FFmpegException(fmt.str());
+		throw FFmpegException(std::string("Could not open movie file! Error: ") + errorStr);
```
asarium: Please use `SCP_string` instead of `std::string`. It's just a typedef but we prefer to use it to make changing the string type easier.

ashdnazg: cool, I've probably ignored a coding convention doc?

asarium: No, it's not documented anywhere currently.

ashdnazg: _sadface_

ashdnazg:
ty :)

chief1983:
@ashdnazg Looking back at this because I went to upgrade the included SDL libs to SDL 2.0.5 and noticed the new *.a files that were included for the MinGW support.  I was wondering why custom built files were used instead of the official MinGW *.a files from the SDL organization?  I made a simple attempt at replacing the prebuilt ones with the official 2.0.5 MinGW libs, and updated the CMakeLists accordingly.  But I don't have an environment to test this with, nor the insight to know if there was a reason this wasn't done the first time around.  Any light you could shed on that would be helpful.  If we can't use the provided libs, it makes the SDL upgrade process an order of magnitude more complicated (as it was previously download, extract, commit).  Would prefer not to break MinGW support after you got it working but we do need a simple, maintainable process for SDL upgrades.

ashdnazg:
The official libs should work I think, I can't remember if tested though, I can probably try, although a better idea is hacking up an appveyor script for mingw testing .

As some libs (not SDL) are harder to find/build I was using https://github.com/mxe/mxe to build the requirements (ffmpeg or ogg or something). We're using the same environment for spring and it's very nice. 
It's highly probable that I just copied whatever I needed from there instead of looking for official builds online.

I'm also on the discord SCP channel if you need me.

PR #1004, opened 2016-10-03
Cleanup some non-HTL code

niffiwan:
This was probably missed in removal of -nohtl
Also "fixes" Coverity 1363380

### DIFF
--- a/code/model/modeloctant.cpp
+++ b/code/model/modeloctant.cpp
@@ -19,9 +19,6 @@
 #include "model/model.h"
 #include "model/modelsinc.h"
 
-extern void model_allocate_interp_data(int n_verts, int n_norms);
-
-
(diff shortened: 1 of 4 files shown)


code/model/modeloctant.cpp
```diff
@@ -19,7 +19,7 @@
 #include "model/model.h"
 #include "model/modelsinc.h"
 
-extern void model_allocate_interp_data(int n_verts, int n_norms, int n_list_verts = 0);
+extern void model_allocate_interp_data(int n_verts, int n_norms);
```
asarium: This is not the same declaration as in modelread.cpp which has default parameters. In any case, if a function is required more than once then it should be put into a header file instead of using extern declaration.

niffiwan: ok & done

PR #1005, opened 2016-10-03
Use FFmpeg for normal audio loading

asarium:
This is the last part of our code that does audio loading on its own.
After this is merged the WAV/ACM/Ogg code can be removed.


asarium:
I found some audio clipping issues that have been introduced by these changes. I'll fix these before this can be merged.

asarium:
I needed to apply a hack for when the channel number is reduced when converting from two to one channel. This is an asset issue and I added a warning for that but it's still not ideal.

Otherwise, these changes should be mostly finished now.

asarium:
I found the bug that was causing the frequency issues. It should be fixed now which fixed the last issue that I am aware of.

niffiwan:
I'm all for warning about the use of a stereo file for a 3D sound; but wow, there'll be a lot of cleanup required in mods. e.g. blueplanet icarus is logging warning left, right & centre...

niffiwan:
FYI - seems to be logging warnings on retail data.
```
l_hit.wav
m_hit.wav
s_hit_2.wav
Pirate.wav
```

There might be more; I guess it'll depend on which ships/weapons/effects are loaded into any given mission.

asarium:
Damn, mod warnings are ok but retail warnings aren't. I'll have to filter those file names but that should be easy enough to do.

MageKing17:
Is my memory playing tricks on me, or do those filenames overlap with the couple retail sound entries that we ignore spurious extra fields for?

asarium:
I can't find any references in the code to the names @niffiwan mentioned.

niffiwan:
I was just checking that; here's the log output, maybe they're filtered by number/index?

```
$ grep _hit fs2_open.log
Dutifully converting retail sound 36, 'l_hit.wav' to a 3D sound...
Dutifully converting retail sound 37, 'm_hit.wav' to a 3D sound...
WARNING: "Sound 'l_hit.wav' has more than one channel but is used as a 3D sound! 3D sounds may only have one channel." at sound.cpp:339
WARNING: "Sound 'm_hit.wav' has more than one channel but is used as a 3D sound! 3D sounds may only have one channel." at sound.cpp:339
```

asarium:
Weird, I couldn't find those strings but now my IDE find them. I'll just add it to the list of files for which the warning isn't shown.

MageKing17:
I thought that sounded familiar, but it's been a while since I read through that code and trying to search the source on my phone is an exercise in futility.

asarium:
Fix submitted in PR #1013.

PR #1007, opened 2016-10-04
Fix Coverity 1363662

niffiwan:
Missing break in switch


niffiwan:
that was quick \:)  There's six other issues detected by Coverity in (I think) the recent ffmpeg commits that you might want to have a look at, if not I'll try to pick them off

asarium:
Alright, I'll take a look at the issues.

PR #1008, opened 2016-10-04
Move vecmath scripting functions into own file

asarium:
This is the first change that moves some of the scripting functions into their own files. The old monolithic `lua.cpp` file was getting hard to edit since typical IDEs have trouble parsing that much code.

The new system requires some additional code to make sure that the static initializers are run in the right order since the oder the initializers are run in is undefined when there are multiple translation units. Instead, this uses a function with a static variable that gets initialized as soon as it's needed which solves the undefined order issue.


code/globalincs/scp_defines.h
```diff
@@ -6,4 +6,7 @@
 #include "scp_compiler_detection.h"
 #include "platformChecks.h"
 
+#define SCP_TOKEN_CONCAT1(x, y) x ## y
+#define SCP_TOKEN_CONCAT(x, y) SCP_TOKEN_CONCAT1(x, y)
```
MageKing17: Er... what is the purpose of having two macros for this?

ashdnazg: This is the usual way concatenation is implemented using C preprocessor.
IIRC If you only use 1 you can't give macros as arguments

asarium: It's specifically for allowing to use `__LINE__` to be used with that macro. If only the first variant was used then "__LINE__" would appear in the source code as opposed to the line number in the file.

asarium:
Is there anything else I should change? I tested the changes quite extensively and as far as I can tell there are no functional changes and scripts still work as expected.

PR #1015, opened 2016-10-08
Refactor tracing subsystem

asarium:
The new design is centered around the Event Trace JSON format but it
should still be possible to adapt it for other profiling tools. The old
profiling system has been removed.

I'll add a wiki post about how this should be used once it has been merged.

### DIFF
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -27,13 +27,15 @@ static int Timer_inited = 0;
 
 #define MICROSECONDS_PER_SECOND 1000000
 
-#define NANOSECONDS_PER_SECOND 10000000000
+#define NANOSECONDS_PER_SECOND 1000000000
 
 static uint64_t get_performance_counter()
 {
 	Assertion(Timer_inited, "This function can only be used when the timer system is initialized!");
 
-	return SDL_GetPerformanceCounter() - Timer_base_value;
+	auto counter = SDL_GetPerformanceCounter();
+
+	return counter - Timer_base_value;
 }
 
 void timer_close()
@@ -104,7 +106,12 @@ std::uint64_t timer_get_nanoseconds()
 {
 	auto time = get_performance_counter();
 
-	return (time * NANOSECONDS_PER_SECOND) / Timer_perf_counter_freq;
+	if (Timer_perf_counter_freq < NANOSECONDS_PER_SECOND) {
+		// This can cause overflows but I don't know a better solution...
+		return (time * NANOSECONDS_PER_SECOND) / Timer_perf_counter_freq;
+	} else {
+		return time / (Timer_perf_counter_freq / NANOSECONDS_PER_SECOND);
+	}
 }
 
 // 0 means invalid,
--- a/code/mission/missiongoals.cpp
+++ b/code/mission/missiongoals.cpp
@@ -1046,7 +1046,9 @@ void mission_eval_goals()
 			}
 
 			// if we get here, then the timestamp on the event has popped -- we should reevaluate
-			PROFILE("Repeating events", mission_process_event(i));
+
+			TRACE_SCOPE(tracing::RepeatingEvents);
+			mission_process_event(i);
 		}
 	}
 	
@@ -1080,7 +1082,8 @@ void mission_eval_goals()
 			// we will evaluate repeatable events at the top of the file so we can get
 			// the exact interval that the designer asked for.
 			if ( !timestamp_valid( Mission_events[i].timestamp) ){
-				PROFILE("Nonrepeating events", mission_process_event( i ));
+				TRACE_SCOPE(tracing::NonrepeatingEvents);
+				mission_process_event( i );
 			}
 		}
 	}
--- a/code/object/object.cpp
+++ b/code/object/object.cpp
@@ -790,6 +790,8 @@ void obj_player_fire_stuff( object *objp, control_info ci )
 
 void obj_move_call_physics(object *objp, float frametime)
 {
+	TRACE_SCOPE(tracing::Physics);
+
 	int has_fired = -1;	//stop fireing stuff-Bobboau
 
 	//	Do physics for objects with OF_PHYSICS flag set and with some engine strength remaining.
@@ -1112,6 +1114,8 @@ void obj_set_flags( object *obj, const flagset<Object::Object_Flags>& new_flags 
 
 void obj_move_all_pre(object *objp, float frametime)
 {
+	TRACE_SCOPE(tracing::PreMove);
+
 	switch( objp->type )	{
 	case OBJ_WEAPON:
 		if (!physics_paused){
@@ -1176,11 +1180,13 @@ extern fireball Fireballs[];
 
 void obj_move_all_post(object *objp, float frametime)
 {
+	TRACE_SCOPE(tracing::PostMove);
+
 	switch (objp->type)
 	{
 		case OBJ_WEAPON:
 		{
-			profile_auto profile_scope("Weapon post move");
+			TRACE_SCOPE(tracing::WeaponPostMove);
 
 			if ( !physics_paused )
 				weapon_process_post( objp, frametime );
@@ -1227,7 +1233,7 @@ void obj_move_all_post(object *objp, float frametime)
 
 		case OBJ_SHIP:
 		{
-			profile_auto profile_scope("Ship post move");
+			TRACE_SCOPE(tracing::ShipPostMove);
 
 			if ( !physics_paused || (objp==Player_obj) ) {
 				ship_process_post( objp, frametime );
@@ -1274,7 +1280,7 @@ void obj_move_all_post(object *objp, float frametime)
 
 		case OBJ_FIREBALL:
 		{
-			profile_auto profile_scope("Fireball post move");
+			TRACE_SCOPE(tracing::FireballPostMove);
 
 			if ( !physics_paused )
 				fireball_process_post(objp,frametime);
@@ -1319,7 +1325,7 @@ void obj_move_all_post(object *objp, float frametime)
 
 		case OBJ_DEBRIS:
 		{
-			profile_auto profile_scope("Debris post move");
+			TRACE_SCOPE(tracing::DebrisPostMove);
 
 			if ( !physics_paused )
 				debris_process_post(objp, frametime);
@@ -1355,7 +1361,7 @@ void obj_move_all_post(object *objp, float frametime)
 
 		case OBJ_ASTEROID:
 		{
-			profile_auto profile_scope("Asteroid post move");
+			TRACE_SCOPE(tracing::AsteroidPostMove);
 
 			if ( !physics_paused )
 				asteroid_process_post(objp);
@@ -1402,6 +1408,8 @@ MONITOR( NumObjects )
  */
 void obj_move_all(float frametime)
 {
+	TRACE_SCOPE(tracing::MoveObjects);
+
 	object *objp;	
 
 	// Goober5000 - HACK HACK HACK
@@ -1436,7 +1444,7 @@ void obj_move_all(float frametime)
 #endif
 
 		// pre-move
-		PROFILE("Pre Move", obj_move_all_pre(objp, frametime));
+		obj_move_all_pre(objp, frametime);
 
 		// store last pos and orient
 		objp->last_pos = cur_pos;
@@ -1449,12 +1457,12 @@ void obj_move_all(float frametime)
 				multi_oo_interp(objp);
 			} else {
 				// physics
-				PROFILE("Physics", obj_move_call_physics(objp, frametime));
+				obj_move_call_physics(objp, frametime);
 			}
 		}
 
 		// move post
-		PROFILE("Post Move", obj_move_all_post(objp, frametime));
+		obj_move_all_post(objp, frametime);
 
 		// Equipment script processing
 		if (objp->type == OBJ_SHIP) {
@@ -1526,15 +1534,14 @@ void obj_move_all(float frametime)
 	// do pre-collision stuff for beam weapons
 	beam_move_all_pre();
 
-	profile_begin("Collision Detection");
 	if ( Collisions_enabled ) {
+		TRACE_SCOPE(tracing::CollisionDetection);
 		if ( Cmdline_old_collision_sys ) {
 			obj_check_all_collisions();
 		} else {
 			obj_sort_and_collide();
 		}
 	}
-	profile_end("Collision Detection");
 
 	turret_swarm_check_validity();
 
@@ -1577,6 +1584,8 @@ void obj_render(object *obj)
 
 void obj_queue_render(object* obj, draw_list* scene)
 {
+	TRACE_SCOPE(tracing::QueueRender);
+
 	if ( obj->flags[Object::Object_Flags::Should_be_dead] ) return;
 
 	// need to figure out what to do with this hook. 
--- a/code/particle/ParticleManager.cpp
+++ b/code/particle/ParticleManager.cpp
@@ -12,6 +12,7 @@
 
 #include "bmpman/bmpman.h"
 #include "globalincs/systemvars.h"
+#include "tracing/tracing.h"
 
 /**
  * @defgroup particleSystems Particle System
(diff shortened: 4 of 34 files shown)


code/io/timer.cpp
```diff
@@ -104,7 +106,12 @@ std::uint64_t timer_get_nanoseconds()
 {
 	auto time = get_performance_counter();
 
-	return (time * NANOSECONDS_PER_SECOND) / Timer_perf_counter_freq;
+	if (Timer_perf_counter_freq < NANOSECONDS_PER_SECOND) {
```
TheMatthew: It is surprisingly common for Timer_perf_counter_freq to be equal to 10e9, maybe a third patch returning time would be worthwhile.

code/io/timer.cpp
```diff
@@ -104,7 +106,12 @@ std::uint64_t timer_get_nanoseconds()
 {
 	auto time = get_performance_counter();
```
TheMatthew: This is the number of ns since the game started. With 1h you have 3 600 000 000 000, so you need 2^44 bits. you will overflow if you multiply by something greater than 2^20, this is a problem

code/io/timer.cpp
```diff
@@ -104,7 +106,12 @@ std::uint64_t timer_get_nanoseconds()
...
 
-	return (time * NANOSECONDS_PER_SECOND) / Timer_perf_counter_freq;
+	if (Timer_perf_counter_freq < NANOSECONDS_PER_SECOND) {
+		// This can cause overflows but I don't know a better solution...
+		return (time * NANOSECONDS_PER_SECOND) / Timer_perf_counter_freq;
+	} else {
+		return time / (Timer_perf_counter_freq / NANOSECONDS_PER_SECOND);
```
TheMatthew: This will result in a loss of precision, are you OK with that?

Example: imagine if the freq is 1.8 GHz like an XBone, the time returned will be time / (1.8/1.0) -> time /1.0 due to long division.

Also, can dividing by the GCD be useful?

Finally, the divided value can probably be cached instead of recalculated every time.

code/io/timer.cpp
```diff
@@ -104,7 +106,12 @@ std::uint64_t timer_get_nanoseconds()
 {
 	auto time = get_performance_counter();
 
-	return (time * NANOSECONDS_PER_SECOND) / Timer_perf_counter_freq;
+	if (Timer_perf_counter_freq < NANOSECONDS_PER_SECOND) {
+		// This can cause overflows but I don't know a better solution...
+		return (time * NANOSECONDS_PER_SECOND) / Timer_perf_counter_freq;
```
TheMatthew: How about

time * (NS/GCD(NS, tpcf) )/ (tpcf/GCD(NS, tpcf)); both values of the division can be cached as they are consts.

code/tracing/categories.cpp
```diff
@@ -0,0 +1,72 @@
+
```
TheMatthew: To add a tracepoint, one needs to create  a new category here I think.

code/tracing/categories.h
```diff
@@ -0,0 +1,92 @@
+
```
TheMatthew: And then declare it here

code/tracing/categories.h
```diff
@@ -0,0 +1,92 @@
...
+#ifndef _TRACING_CATEGORIES_H
+#define _TRACING_CATEGORIES_H
+#pragma once
+
+namespace tracing {
+class Category {
+	const char* _name;
+	bool _graphics_category;
+ public:
+	Category(const char* name, bool graph) : _name(name), _graphics_category(graph) {
```
TheMatthew: What does "graph" mean?

code/tracing/tracing.cpp
```diff
@@ -103,507 +70,299 @@ static int get_query_object() {
...
-		if (do_gpu_queries) {
-			data.gpu_query = get_query_object();
-			gr_query_value(data.gpu_query, QueryType::Timestamp);
-		} else {
-			data.gpu_query = -1;
-		}
+namespace complete {
 
-		current_frame_data.push_back(data);
+void start(const Category& category, trace_event* evt) {
```
TheMatthew: At this moment, I think that the audio subsystems use a separate thread. Soon SDL and mesa will probably have a threaded OpenGL backend. It may be interesting to store the PID/TID of the event generator to be able to rebuild execution graphs. I think it is being done, but I am not 100% certain.

code/tracing/tracing.cpp
```diff
@@ -103,507 +70,299 @@ static int get_query_object() {
...
+		gpu_trace_event gpu_event;
+		gpu_event.base_evt.category = &category;
+		gpu_event.base_evt.tid = 1;
+		gpu_event.base_evt.pid = GPU_PID;
+		gpu_event.base_evt.type = EventType::Begin;
 
-				break;
-			}
-		}
+		gpu_event.gpu_begin_query = get_gpu_timestamp_query();
```
TheMatthew: GPU times may be different from system times especially with message queues. Questions: does gr_query flush the queues, as it would change the timing info and that would be very impactful. Also, in post-processing are you ok with gpu information being with different time bases?

TheMatthew:
First github review, if my questions are unclear, don't hesitate to ask for clarifications.

code/io/timer.cpp
asarium: It's not necessarily in nanoseconds since the unit depends on what `SDL_GetPerformanceFrequency()` returns but on Linux this is in nanoseconds so there is an issue with possible overflows.

code/io/timer.cpp
asarium: I don't think that the loss of precision is an issue here since we don't actually need such a high precision timer anywhere. Microseconds are more than enough.

I would have to take a look at the math to determine if GCD would be useful here but if you plan to propose a patch for these things then you could include it there.

code/tracing/categories.h
asarium: Yes, that's how it's supposed to work. That should allow to declare the tracepoint ahead of time without using strings for identifying them which should help if another tracing framework needs that.

code/tracing/categories.h
asarium: It's supposed to be an abbreviation for `graphics_category`.

code/io/timer.cpp
TheMatthew: True, I was doing this more to follow the code out loud.

code/tracing/tracing.cpp
asarium: That's exactly what is being done. The PID isn't hugely relevant since we only use one process but if we ever use more than one this should be able to handle that. The TID is correctly recorded and I already tested it with the new multithreaded video player.

code/tracing/tracing.cpp
asarium: The GPU queries are inserted into the command stream at the moment when the `glQuery` is executed but there is no queue in the graphics system to be flushed so I'm not sure what you mean by that.

The problem with different time bases is that the TraceEvent viewer cannot handle them since it thinks all times originate from the same source so I need to synchronize them to avoid crashing the viewer. I hope that the timer in the GPU is ticking at the same rate as the CPU timer but if it isn't then this is the best result we can get without going into clock synchronization and clock drift compensation territory.

code/io/timer.cpp
TheMatthew: So now I have a better idea of what happens: 
In Windows, the value is in 1024 clock cycles. so typically ~2500000 assuming most computers run between 2 and 3 ghz. In Linux, the value on most modern kernels is 1000000000.

code/io/timer.cpp
TheMatthew: This path occurs with Windows It is good that it happens in the first branch.
Also the overflow can be a definite issue on the bulk of users.

code/tracing/categories.h
TheMatthew: Thanks

code/tracing/tracing.cpp
TheMatthew: I would suggest having all the info available as you're doing and compensating after the fact. 

Typically in tracing you find the outliers then write a huge amount of post-mortem analyses. 

According to https://www.hindawi.com/journals/ase/2015/940628/ the clock drift is under 50 ms typically so I would not worry too much.

asarium: The tracing data for one event is directly written to the disk when it's completed so there is no easy way to do a post-run analysis or clock compensation. IMO it's not a big problem to do this compensation this way since it's a non-destructive change to the data. There are a few techniques I could use to improve this but the current approach works well enough.

code/tracing/TraceEventWriter.cpp
```diff
@@ -0,0 +1,111 @@
...
+			return "";
+	}
+}
+
+void writeTime(std::ofstream& out, std::uint64_t time) {
+	// Save stream state
+	auto flags = out.flags();
+	out << std::fixed;
+
+	out << (time / 1000.);
```
TheMatthew: When running on this, we have timestamps ending with 000, this makes 6 bytes extra per event. that will save 5-8% per trace size.

asarium: Also, it's not always 000 since the time is in nanosecond precision.

TheMatthew: I probably poorly explained, the trace has 1234.567000 where the '.' is at the us level, 567 are nanoseconds and 000 is in picoseconds.

code/io/timer.cpp
TheMatthew: I now have a couple of small patches queued behind this one.

TheMatthew:
I am looking at this a bit more, it would be interesting to get some system info at the begining: cpu type, gpu type, ram, version number and build id?

asarium:
CPU and GPU type is a bit hard to determine since it would require platform specific function calls. SDL provides `SDL_GetSystemRAM` so we can use that to determine the available RAM. The version already contains the build ID so that could also be included.

TheMatthew:
It may help debug issues, another patch after though?

asarium:
Yes, that's definitely something that should be done in another pull request.

code/tracing/TraceEventWriter.cpp
z64555: So it would be simple to just modify the float/double format of the out stream to truncate the 0 tail?

asarium: The C++ stream format flags are very weird to work with. I guess `setw` should work here so I'll try that.

code/object/object.cpp
```diff
@@ -1436,7 +1444,7 @@ void obj_move_all(float frametime)
 #endif
 
 		// pre-move
-		PROFILE("Pre Move", obj_move_all_pre(objp, frametime));
+		 obj_move_all_pre(objp, frametime);
```
z64555: Errant space! D:

code/particle/ParticleManager.cpp
```diff
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <memory>
+#include <tracing/categories.h>
```
z64555: Is this needed for this file?

code/tracing/MainFrameTimer.cpp
```diff
@@ -0,0 +1,34 @@
...
+}
+MainFrameTimer::~MainFrameTimer() {
+	_out.close();
+}
+void MainFrameTimer::processEvent(const trace_event* event) {
+	if (event->scope != &MainFrameScope || event->category != &MainFrame) {
+		return;
+	}
+
+	if (event->type == EventType::AsyncBegin) {
```
z64555: Can use a switch block here instead of if-else blocks if `event->type` is the only thing used

code/tracing/ThreadedEventProcessor.h
```diff
@@ -0,0 +1,57 @@
...
+	~ThreadedEventProcessor() {
+		_event_queue.close();
+		_worker_thread.join();
+	}
+
+	void processEvent(const trace_event* event) {
+		try {
+			_event_queue.wait_push_back(std::move(*event));
+		} catch (const sync_queue_is_closed&) {
+			// For some reason the queue is already closed...
```
z64555: Incomplete? Maybe print an output message somewhere

code/tracing/TraceEventWriter.cpp
```diff
@@ -0,0 +1,111 @@
...
+	if (!_first_line) {
+		_out << ",";
+	}
+	_out << "\n{\"tid\": " << event->tid << ",\"ts\":";
+
+	writeTime(_out, event->timestamp);
+
+	_out << ",\"pid\":";
+	if (event->pid == GPU_PID) {
+		_out << "\"GPU\"";
```
z64555: Does this work with multiple GPU's?

code/tracing/TraceEventWriter.cpp
```diff
@@ -0,0 +1,111 @@
...
+		case EventType::End:
+			// Nothing to do here
+			break;
+		case EventType::AsyncBegin:
+		case EventType::AsyncStep:
+		case EventType::AsyncEnd:
+			// Nothing to do here...
+			break;
+		default:
+			Assertion(false, "Invalid enum value!");
```
z64555: Error message could be more descriptive.

code/tracing/categories.h
z64555: A doxy'd comment block in this file, or in the tracing.h file, should be made to outline maintenance of the tracing system and what to do if devs want to add new trace points.

code/tracing/TraceEventWriter.cpp
asarium: The code uses the timestamps provided by OpenGL so it doesn't matter if there are multiple GPUs since that would be hidden behind the OpenGL API.

asarium:
@z64555 I pushed some changes which should address your comments.

code/mission/missiongoals.cpp
```diff
@@ -1080,7 +1082,8 @@ void mission_eval_goals()
 			// we will evaluate repeatable events at the top of the file so we can get
 			// the exact interval that the designer asked for.
 			if ( !timestamp_valid( Mission_events[i].timestamp) ){
-				PROFILE("Nonrepeating events", mission_process_event( i ));
+				TRACE_SCOPE(tracing::NonrepeatingEvents);
+				mission_process_event( i );
```
TheMatthew: Spacing is inconsistent.

code/mission/missiongoals.cpp
```diff
@@ -1046,7 +1046,9 @@ void mission_eval_goals()
 			}
 
 			// if we get here, then the timestamp on the event has popped -- we should reevaluate
-			PROFILE("Repeating events", mission_process_event(i));
+
+			TRACE_SCOPE(tracing::RepeatingEvents);
+			mission_process_event(i);
```
TheMatthew: Spacing is inconsistent.

code/tracing/MainFrameTimer.cpp
```diff
@@ -0,0 +1,37 @@
+//
```
TheMatthew: Probably needs a better copyright header

code/tracing/tracing.cpp
TheMatthew: I viewed traces. the delay is quite acceptable.

code/tracing/MainFrameTimer.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
+void MainFrameTimer::processEvent(const trace_event* event) {
+	if (event->scope != &MainFrameScope || event->category != &MainFrame) {
+		return;
+	}
+
+	switch(event->type) {
+		case EventType::AsyncBegin:
+			_begin_time = event->timestamp;
+		case EventType::AsyncEnd:
```
z64555: Intentional fall-through? This would output a message at the beginning of the event with a message like "2400; 0"

code/mission/missiongoals.cpp
asarium: The spacing is the same as it was before my changes. We generally try to keep the style like it was before when changing code.

code/tracing/MainFrameTimer.cpp
asarium: No, this is definitely not intentional.

code/tracing/MainFrameTimer.cpp
asarium: I generally don't include a copyright header so I'll just remove those comment lines.

asarium:
Thanks for @TheMatthew and @z64555 for reviewing the code. I think this should be ready to be merged now.
Be aware that these changes remove the current on-screen profiling information but IMO that has been superseded by the event trace viewers. The viewers (the one in chrome and [spacecompass](https://github.com/TheMatthew/spacecompass)) make analyzing the data much easier so I don't think the removal of the old profiling output is a big loss.

PR #1017, opened 2016-10-09
Show an error dialog when an exception is thrown in main

asarium:
This will make sure that an exception doesn't just silently crash the
game (if the dialog can be shown at that point). This caused some issues
in the past so I thought it would be a good idea. I also added a check
for a preprocessor define to disable exception catching completely which
should make it possible to use a debugger to debug uncaught exceptions.


chief1983:
This will fall within the silentmode command line flag and go to the terminal when that flag is active right?

asarium:
No necessarily. If there is an exception before the command line is parsed then `Error` will still try to create a dialog window but it's very unlikely that an exception gets thrown before the command line has been parsed.

Issue #986, opened 2016-09-27
FSO crash on start (Linux/Nvidia)


niffiwan:
As of commit efdd4fae68f299a2cfaafd62e2af915efb92b5db FSO logs this error on startup:

```
Error: Unable to initialize display device!

File: gropengl.cpp
Line: 1546
```

Full log: [fs2_open.log.txt](https://github.com/scp-fs2open/fs2open.github.com/files/495423/fs2_open.log.txt)

OS is Mint 17.3 with GTS 450. Prior commits produce this in the log:

```
Initializing OpenGL graphics device at 1600x900 with 32-bit color...
  Initializing SDL video...
  Requested SDL Video values = R: 8, G: 8, B: 8, depth: 24, stencil: 8, double-buffer: 1, FSAA: 4
  Actual SDL Video values    = R: 8, G: 8, B: 8, depth: 24, stencil: 8, double-buffer: 1, FSAA: 4
  OpenGL Vendor    : NVIDIA Corporation
  OpenGL Renderer  : GeForce GTS 450/PCIe/SSE2
  OpenGL Version   : 4.4.0 NVIDIA 340.96

  Using extension "GL_EXT_fog_coord".
  Using extension "GL_ARB_multitexture".
  Using extension "GL_ARB_texture_env_add".
  Using extension "GL_ARB_texture_compression".
  Using extension "GL_EXT_texture_compression_s3tc".
  Using extension "GL_EXT_texture_filter_anisotropic".
  Using extension "GL_ARB_texture_env_combine".
  Using extension "GL_EXT_compiled_vertex_array".
  Using extension "GL_EXT_draw_range_elements".
  Using extension "GL_ARB_texture_mirrored_repeat".
  Using extension "GL_ARB_texture_non_power_of_two".
  Using extension "GL_ARB_vertex_buffer_object".
  Using extension "GL_ARB_pixel_buffer_object".
  Using extension "GL_SGIS_generate_mipmap".
  Using extension "GL_EXT_framebuffer_object".
  Using extension "GL_ARB_texture_rectangle".
  Using extension "GL_EXT_bgra".
  Using extension "GL_ARB_texture_cube_map".
  Using extension "GL_EXT_texture_lod_bias".
  Using extension "GL_ARB_point_sprite".
  Using extension "GL_ARB_shading_language_100".
  Using extension "GL_ARB_shader_objects".
  Using extension "GL_ARB_vertex_shader".
[...]

PR #1018, opened 2016-10-09
Add disable AA info to OGL init failure

niffiwan:
Reported as occuring to several people, so far two on Linux with nVidia
binary blob drivers. Extra info should assist in users fixing the
problem on their own.

(see #986)


MageKing17:
Wouldn't it be simpler to just disable anti-aliasing altogether?

niffiwan:
You mean remove AA from the code entirely, or disable AA if the OGL context creation fails?

MageKing17:
It was my understanding that the non-FXAA anti-aliasing wasn't actually working, and hadn't been for some time (not just on Windows).

niffiwan:
well - that's pretty easy to do; and AFAIK MSAA & FXAA are incompatible so why not \:D  New PR arriving soon.

PR #1019, opened 2016-10-09
Remove MSAA

niffiwan:
Reasons being; it's incompatible with FXAA, it's slow, and it's causing
OGL init to fail on certain systems. Oh, and it probably doesn't work
anyway :)

(#986)

### DIFF
--- a/code/osapi/osapi.h
+++ b/code/osapi/osapi.h
@@ -139,8 +139,6 @@ namespace os
 
 		uint32_t depth_size;
 		uint32_t stencil_size;
-
-		uint32_t multi_samples; //!< The amount of multi-sampling, use 0 for no multi-sampling
 	};
 
 	/**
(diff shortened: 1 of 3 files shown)


code/osapi/osapi.h
```diff
@@ -139,8 +139,6 @@ namespace os
 
 		uint32_t depth_size;
 		uint32_t stencil_size;
-
-		uint32_t multi_samples; //!< The amount of multi-sampling, use 0 for no multi-sampling
```
asarium: IMO you shouldn't remove the option from the viewports API since multi-sampling is still a valid option and may even be enabled again later if we ever find a fix for this bug. You should just set `multi_samples` to `0` and ignore the config value.

niffiwan: ok; I've reverted most of this to implement a minimal disable of MSAA instead

PR #1024, opened 2016-10-10
Improve logic for determining legacy mode

### LINKED ISSUE #858: Default to new config location
For new users (meaning no config whatsoever in either the old or the new location) FSO complains that they should move to the new location (for me: `~/.local/share/HardLightProductions/FreeSpaceOpen/`), but creates all files in the old location (for me: `~/.fs2_open`).

The directory for the new location is created but stays empty.

New users should probably default to the new location, not the legacy location.

Workaround: simply start the game and move the files.
Or: before starting, run 

```
mkdir -p ~/.local/share/HardLightProductions/FreeSpaceOpen
touch ~/.local/share/HardLightProductions/FreeSpaceOpen/fs2_open.ini
```

asarium:
Now it will also check if the old config location exists and if it
doesn't then the new config mode will be used by default. This should
fix #858.

This solution doesn't work too well on windows since the registry system isn't initialized yet when `os_is_legacy_mode` is first called so I just assume that the old config exists. Most Windows users use a launcher which should handle that transition automatically so this shouldn't be a huge issue.

### DIFF
--- a/code/osapi/osapi.cpp
+++ b/code/osapi/osapi.cpp
@@ -19,9 +19,7 @@
 #include "cmdline/cmdline.h"
 #include "osapi/osapi.h"
 
-
-#include <SDL_assert.h>
-
+#include <fstream>
 #include <algorithm>
 
 namespace
@@ -304,6 +302,11 @@ void os_sleep(uint ms)
 #endif
 }
 
+static bool file_exists(const SCP_string& path) {
+	std::ofstream str(path, std::ios::in);
+	return str.good();
+}
+
 bool os_is_legacy_mode()
 {
 	// Make this check a little faster by caching the result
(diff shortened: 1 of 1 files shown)


chief1983:
Will this mean that Windows users only need to use the launcher once, and then can run directly, or will it always be a problem to not use the launcher?

MageKing17:
When has it ever been a problem to run FSO without going through the launcher?

asarium:
It was never a problem to run FSO directly, even after the new config code was merged.

code/osapi/osapi.cpp
```diff
@@ -318,19 +321,31 @@ bool os_is_legacy_mode()
...
+		new_config_exists = file_exists(path_stream.str());
+#ifdef SCP_UNIX
+        path_stream.str("");
+		path_stream << Cfile_user_dir_legacy << DIR_SEPARATOR_CHAR << Osreg_config_file_name;
 
-		if (file == nullptr)
-		{
+		old_config_exists = file_exists(path_stream.str());
+#else
+		// At this point we can't determine if the old config exists so just assume that it does
```
MageKing17: Why can't we check for the existence of the registry key?

asarium: This function is called very early into the game initialization at which point the registry values aren't initialized yet which means that the os_config system isn't able to build the correct registry path.

MageKing17: Surely changing that would be preferable to Windows builds requiring wxLauncher's help to use a sane method configuration storage...

asarium: I will take another look at the code but I was pretty sure that it's not possible without changes to how the config system is initialized. I would like to avoid such changes for this pull request since they would be out of scope.

asarium: I did some debugging and the `os_is_legacy_mode` function is first called from `os_init_cmdline` which is one of the first functions the game calls when it starts. The `os_init` calls could be moved before the command line parsing which will probably work fine. If that is an acceptable solution then I will change the code accordingly.

MageKing17: Well, that obviously won't work because `os_init()` checks various command-line flags. In fact, `os_is_legacy_mode()` and `os_get_config_path()` both check `Cmdline_portable_mode`, which is useless since they both get called before the command line is actually parsed (unless "-parse_cmdline_only" is present). Hrmph... clearly this is a deeper problem that requires some additional thought.

asarium: Yeah, you're right. This might also affect how the portable mode is handled since that flag isn't set yet when `os_is_legacy_mode()` is called first. This might need some more involved solution which also fixes that behavior...

asarium:
@MageKing17 I think it's pretty clear that providing the same, improved behavior on Windows requires a lot more work so how about merging these changes now and fixing the behavior on Windows after the next release? As I said above, Windows isn't affected as much since pretty much everyone uses a launcher.

PR #1025, opened 2016-10-10
Fix reading credits table without #end

asarium:
This just checks if the end of the file has been reached so existing
tables without the #end will still work (I'm looking at you
FSPort...).


MageKing17:
Wait, we have existing tables that result in an infinite loop?

asarium:
Yes, the FSPort credits table misses the ´#end´. It doesn't always result in an infinite loop. On Windows it keeps reading until it finds an ´#end´ somewhere in memory and on Linux it causes a segmentation fault.

Issue #995, opened 2016-10-01
PR #844 breaks PBR rendering


DahBlount:
After doing some checking via git bisect, PBR ships aren't being rendered correctly after the merge of #844.
This was first brought up in the [HTL Leviathan thread](http://www.hard-light.net/forums/index.php?topic=92587.0). It appears to be that ships aren't rendering the PBR related maps and instead ships with PBR textures get treated as only having diffuse, normal, and glow.

niffiwan:
From Coverity 1363381; could this be the root cause? 

```
CID 1363381 (#1 of 1): Arguments in wrong order (SWAPPED_ARGUMENTS)swapped_arguments: The positions of arguments in the call to get_shader_hash do not match the ordering of the parameters:

    frag_content is passed to geom_content
    geom_content is passed to frag
```
https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/graphics/opengl/gropenglshader.cpp#L566

asarium:
No, that's actually my fault but it doesn't cause this issue. I'll fix is ASAP.

asarium:
The new material code does not set the ambient occlusion map but even after fixing that the PBR models are still black.

DahBlount:
Found some interesting things.
-reflect and -shine included:
![](http://puu.sh/rFTTp/916a583131.png)
only -reflect:
![](http://puu.sh/rFTYz/62f5b8aefa.png)
only -shine:
![](http://puu.sh/rFU0Y/a7ef45fa0f.png)

Does this help at all?

PR #1030, opened 2016-10-12
Fixed reflectance maps and ambient/cavity maps not being used

SamuelCho:
A hiccup in the transition to OpenGL Core. Sorry for the inconvenience. Should fix https://github.com/scp-fs2open/fs2open.github.com/issues/995

### DIFF
--- a/code/graphics/material.cpp
+++ b/code/graphics/material.cpp
@@ -575,7 +575,7 @@ uint model_material::get_shader_flags()
 		Shader_flags |= SDR_FLAG_MODEL_GLOW_MAP;
 	}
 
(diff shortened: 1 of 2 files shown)


code/graphics/material.cpp
```diff
@@ -591,6 +591,10 @@ uint model_material::get_shader_flags()
 		Shader_flags |= SDR_FLAG_MODEL_HEIGHT_MAP;
 	}
 
+	if ( get_texture_map(TM_AMBIENT_TYPE) > 0) ) {
```
The-E: One closing brace too many, I'm afraid

PR #1031, opened 2016-10-12
Remove extension from a filename passed to WaveFile::Open

### LINKED ISSUE #1029: String ERANGE errors with persona voice files with long names
This bug was found right in the middle of a long BP mission, so I did my best to recreate the exact circumstances it occurred there.

Please see this mod for a reproducible test case:
 https://drive.google.com/file/d/0B_q0tkX7VcSaY0FUbjliTXFWcjg/view?usp=sharing

Run message.fs2, call in Delta reinforcement wing, and order them to attack the Shivan ship. The game should crash.

The full error you should see is:

`Error: ERANGE: String error. Please Report.
Trying to put into 32 byte buffer:
.ogg`

The filename being played, uef3fleetfattackdesperate.ogg, only has 29 characters with the extension. Looking inside Vis Studio, there's a null terminator replacing the first character of the filename that its trying to process. So the string becomes "\0ef3fleetfattackdesperate.ogg.og"

It seems this only occurs when the voice file is in voice/personas. I had tried putting this in just sounds at first and it didn't crash.

I leave this in anyone's capable hands. :)

(Shout out to Cobra for originally finding this)

asarium:
This fixes #1029. Special thanks to @AxemP for providing a test mod for
reproducing the bug.

There are also some minor formatting changes which I could remove again if required.

### DIFF
--- a/code/sound/ffmpeg/WaveFile.cpp
+++ b/code/sound/ffmpeg/WaveFile.cpp
@@ -16,10 +16,10 @@ AudioProperties getAudioProps(AVStream* stream) {
 
 	int channels;
 #if LIBAVCODEC_VERSION_INT > AV_VERSION_INT(57, 24, 255)
-	channels = stream->codecpar->channels;
-	props.channel_layout = stream->codecpar->channel_layout;
-	props.sample_rate = stream->codecpar->sample_rate;
-	props.format = (AVSampleFormat)stream->codecpar->format;
+    channels = stream->codecpar->channels;
+    props.channel_layout = stream->codecpar->channel_layout;
+    props.sample_rate = stream->codecpar->sample_rate;
+    props.format = (AVSampleFormat)stream->codecpar->format;
 #else
 	channels = stream->codec->channels;
 	props.channel_layout = stream->codec->channel_layout;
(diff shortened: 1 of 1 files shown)


code/sound/ffmpeg/WaveFile.cpp
```diff
@@ -125,16 +124,15 @@ WaveFile::~WaveFile() {
...
-bool WaveFile::Open(const char *pszFilename, bool keep_ext)
-{
+bool WaveFile::Open(const char* pszFilename, bool keep_ext) {
 	using namespace libs::ffmpeg;
 
 	size_t FileSize, FileOffset;
 	char fullpath[MAX_PATH_LEN];
 	char filename[MAX_FILENAME_LEN];
 
 	// NOTE: we assume that the extension has already been stripped off if it was supposed to be!!
```
MageKing17: This comment is no longer accurate.

code/sound/ffmpeg/WaveFile.cpp
```diff
@@ -169,14 +167,21 @@ bool WaveFile::Open(const char *pszFilename, bool keep_ext)
 				throw FFmpegException("Unknown file extension.");
```
MageKing17: This error should probably be clarified (if the extension isn't specified, how can it be unknown?), but may be out of scope for this PR.

code/sound/ffmpeg/WaveFile.cpp
asarium: I'll change this right away.

code/sound/ffmpeg/WaveFile.cpp
asarium: Fixed, not it just says that the file could not be found.

PR #1033, opened 2016-10-12
Add isWeaponUsed scripting function

DahBlount:
Requested by Axem.

Allows a weapon to be paged in via script so if a weapon has special qualities through scripting, say, upgrades given a player state, the weapons can be paged in and rendered correctly.

### DIFF
--- a/code/scripting/lua.cpp
+++ b/code/scripting/lua.cpp
@@ -4638,21 +4638,33 @@ ADE_FUNC(isBeam, l_Weaponclass, NULL, "Return true if the weapon is a beam", "bo
 }
 
 // Checks if a weapon has been paged in (counted as used)
-ADE_FUNC(isWeaponUsed, l_Weaponclass, "[boolean Page_In = false]", "Return true is the weapon is paged in, optionally paging in the weapon.", "boolean", "true if the weapon is paged in, pages in the weapon if otherwise")
+ADE_FUNC(isWeaponUsed, l_Weaponclass, NULL, "Return true is the weapon is paged in, optionally paging in the weapon.", "boolean", "true if the weapon is paged in, pages in the weapon if otherwise")
 {
 	int idx;
-	bool page_check = false;
-	if (!ade_get_args(L, "o|b", l_Weaponclass.Get(&idx), &page_check))
+	if (!ade_get_args(L, "o|b", l_Weaponclass.Get(&idx)))
 		return ADE_RETURN_NIL;
 
 	if (idx < 0 || idx >= Num_weapon_types)
 		return ADE_RETURN_FALSE;
 
-	if (page_check) {
-		weapon_page_in(idx);
-	}
-
 	if (!weapon_used(idx)) {
+		return ADE_RETURN_FALSE;
+	} else {
+		return ADE_RETURN_TRUE;
+	}
+}
+
+// Checks if a weapon has been paged in (counted as used)
+ADE_FUNC(loadWeapon, l_Weaponclass, NULL, "Pages in a weapon. Returns True on success.", "boolean", "True if page in was successful, false otherwise.")
+{
+	int idx;
+	if (!ade_get_args(L, "o|b", l_Weaponclass.Get(&idx)))
+		return ADE_RETURN_NIL;
+
+	if (idx < 0 || idx >= Num_weapon_types)
+		return ADE_RETURN_FALSE;
+
+	if (!weapon_page_in(idx)) {
 		return ADE_RETURN_FALSE;
 	} else {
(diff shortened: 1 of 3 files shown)


asarium:
IMO it would be better to either split the functionality into two functions where one queries the value and one pages the weapon in or it could use a variable with which a script could mark a weapon as used which would also page in the resources of the weapon.

DahBlount:
This was designed to be functionally like the isModelLoaded function for ship handles at the request of Axem.

The issue with the script marking it as used after the fact is that the weapons only get paged in during level paging, not during mission run time. So you have to force the weapons to get paged in and mark them used simultaneously.

asarium:
Ok, then the two function approach seems to be the better solution. Grouping two functions together into one function is a bad design choice since the function does two things but the function name only implies one thing. I think two functions would be better in this situation.

DahBlount:
Alright. Here are the changes.

code/scripting/lua.cpp
```diff
@@ -4637,6 +4637,40 @@ ADE_FUNC(isBeam, l_Weaponclass, NULL, "Return true if the weapon is a beam", "bo
 		return ADE_RETURN_FALSE;
 }
 
+// Checks if a weapon has been paged in (counted as used)
+ADE_FUNC(isWeaponUsed, l_Weaponclass, NULL, "Return true is the weapon is paged in, optionally paging in the weapon.", "boolean", "true if the weapon is paged in, pages in the weapon if otherwise")
```
MageKing17: This description would seem to be wrong.

code/scripting/lua.cpp
```diff
@@ -4637,6 +4637,40 @@ ADE_FUNC(isBeam, l_Weaponclass, NULL, "Return true if the weapon is a beam", "bo
...
+{
+	int idx;
```
MageKing17: Leaving `idx` uninitialized is probably not a good idea (we've had plenty of problems with uninitialized variables in the Lua interface already).

code/scripting/lua.cpp
```diff
@@ -4637,6 +4637,40 @@ ADE_FUNC(isBeam, l_Weaponclass, NULL, "Return true if the weapon is a beam", "bo
...
+		return ADE_RETURN_FALSE;
+
+	if (!weapon_used(idx)) {
+		return ADE_RETURN_FALSE;
+	} else {
+		return ADE_RETURN_TRUE;
+	}
+}
+
+// Checks if a weapon has been paged in (counted as used)
```
MageKing17: This comment is a duplicate of the one for the previous function.

code/scripting/lua.cpp
```diff
@@ -4637,6 +4637,40 @@ ADE_FUNC(isBeam, l_Weaponclass, NULL, "Return true if the weapon is a beam", "bo
...
+ADE_FUNC(loadWeapon, l_Weaponclass, NULL, "Pages in a weapon. Returns True on success.", "boolean", "True if page in was successful, false otherwise.")
+{
+	int idx;
```
MageKing17: See previous comment about uninitialized `idx`.

code/scripting/lua.cpp
```diff
@@ -4637,6 +4637,40 @@ ADE_FUNC(isBeam, l_Weaponclass, NULL, "Return true if the weapon is a beam", "bo
 		return ADE_RETURN_FALSE;
 }
 
+// Checks if a weapon has been paged in (counted as used)
+ADE_FUNC(isWeaponUsed, l_Weaponclass, NULL, "Return true is the weapon is paged in, optionally paging in the weapon.", "boolean", "true if the weapon is paged in, pages in the weapon if otherwise")
+{
+	int idx;
+	if (!ade_get_args(L, "o|b", l_Weaponclass.Get(&idx)))
```
MageKing17: No optional boolean argument anymore.

code/scripting/lua.cpp
```diff
@@ -4637,6 +4637,40 @@ ADE_FUNC(isBeam, l_Weaponclass, NULL, "Return true if the weapon is a beam", "bo
...
+	} else {
+		return ADE_RETURN_TRUE;
+	}
+}
+
+// Checks if a weapon has been paged in (counted as used)
+ADE_FUNC(loadWeapon, l_Weaponclass, NULL, "Pages in a weapon. Returns True on success.", "boolean", "True if page in was successful, false otherwise.")
+{
+	int idx;
+	if (!ade_get_args(L, "o|b", l_Weaponclass.Get(&idx)))
```
MageKing17: This function should never have had an optional boolean argument.

DahBlount:
Applied the suggested changes.

PR #1034, opened 2016-10-13
Fix isWeaponUsed Description

DahBlount:
isWeaponUsed description was wrong.


niffiwan:
FYI, have cancelled & restarted the OSX build as not starting for 2 hrs is suspicious

DahBlount:
Thanks.

PR #1037, opened 2016-10-13
Use prebuilt libraries for SDL2 framework

asarium:
This uses the new prebuilt mechanism for including the SDL2 framework on Mac. I tested the compilation with Travis CI but I can't test if the produced builds still run so I'll need some help with that.


chief1983:
This is working as well as master, that is to say, I still get a crash loading a mission on the 'beep.wav' file.  But the game starts and I can hear sounds, etc.

PR #1042, opened 2016-10-14
Add compatibility code for older FFmpeg versions

asarium:
This adds compatibility with earlier versions of FFmpeg. A part of these changes were provided by @chief1983.


chief1983:
This has compiled fine for me now on FreeBSD, Gentoo, and Ubuntu, all with 2.8.x series FFMpeg.  Was going to test the video playback on FreeBSD but there appears to have been something broken its OpenGL init, so I'll have to look into that before I have a platform that could actually confirm video playback works on a 2.8.x machine.

PR #1043, opened 2016-10-14
Fix TTF rendering on cockpit displays

### LINKED ISSUE #1038: RTT HUD gauges broken with TTF fonts
(duplicated from Mantis #0003185)
Text elements of render-to-texture gauges used with a cockpit model break when used with truetype fonts. The text may display in the wrong place, upside down, backwards, as garbage, or not at all. I have created a demonstration mod that shows off this issue. This is a modification of fightermedic's cockpit mod that replaces font01.vf with a TTF of Bank Gothic. You will see no text on the cockpit gauges when you load any mission (this mod runs off of MediaVPs 2014 and works with retail missions). Removing the fonts.tbl, which defines the truetype font in place of font01.vf, makes everything display correctly.

Download here: https://www.sendspace.com/file/ev2bhb

asarium:
I needed to disable backface culling but I don't think it actually
matters.

Fixes #1038.

### DIFF
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -2994,6 +2994,8 @@ void bm_set_low_mem(int mode) {
 }
 
 bool bm_set_render_target(int handle, int face) {
+	GR_DEBUG_SCOPE("Set render target");
+
 	int n = handle % MAX_BITMAPS;
 
 	if (n >= 0) {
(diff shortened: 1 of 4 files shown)


niffiwan:
To confirm my understanding, face culling has been disabled when rendering text only? And this might cause some imperceptible performance loss due to more faces potentially being rendered? (although not sure how text "objects" would have faces to be culled, they're all 2D, right?)

asarium:
It has been disabled for all NanoVG operations which includes most 2D operations expect bitmap drawing. Since all the triangles should face the screen anyway it wouldn't have done anything in most cases and since most triangles were drawn it also shouldn't cause issues with performance.

PR #1047, opened 2016-10-15
Fix issues with $Max Shield Recharge

niffiwan:
Fix unlimited transfer of energy to shields (energy lost)
Fix thresholds in ai balance shields, rearm timer & call4help
Fix scaling in shield pain flash
Fix variable number of shield segments in shield pain flash
Lastly includes formatting tidy-up in shield pain flash

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -12272,7 +12272,7 @@ void ai_balance_shield(object *objp)
 	float	delta;
 
 	// if we are already at the max shield strength then just bail now
-	if ( Ships[objp->instance].ship_max_shield_strength * Ships[objp->instance].max_shield_recharge <= shield_get_strength(objp) )
+	if ( shield_get_strength(objp) >= shield_get_max_strength(objp) )
 		return;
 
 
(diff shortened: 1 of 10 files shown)


code/ai/aicode.cpp
```diff
@@ -12271,8 +12271,8 @@ void ai_balance_shield(object *objp)
 	float	shield_strength_avg;
 	float	delta;
 
-	// if we are already at the max shield strength for all quads then just bail now
-	if ( Ships[objp->instance].ship_max_shield_strength == shield_get_strength(objp) )
+	// if we are already at the max shield strength then just bail now
+	if ( Ships[objp->instance].ship_max_shield_strength * Ships[objp->instance].max_shield_recharge <= shield_get_strength(objp) )
```
asarium: This will now return from this function every time the shield strength is less than the maximum shield strength, previously it only returned when the shield was at maximum strength. Should this be a `>=`?

asarium: Ah ok. I think I read it backwards which is why I was a bit confused.

niffiwan:
Good idea about the function as it took me quite a while to carefully check all the places so see if the calculation was needed or not.  Note that the intent of the feature is to allow shields to be all set like 100% forward, 0% on all other segments.

PR #1055, opened 2016-10-21
Absolute values

PeterMitsis:
I happened to notice that with the x86 GCC compiler (not tested with other compilers) that using fl_abs() (which evaluates to fabsf) can be used to generate tighter code.

As fabsf() is a function well known to the compiler, GCC is able to optimize it such that the sign bit in a float is unconditionally cleared in a single SSE instruction (and no function call).

Although I would expect that this would generate slightly more efficient code, I have not benchmarked that. I have however done a size check and found that the text section is smaller by about half a page (peanuts in the grand scheme of things).

### DIFF
--- a/code/math/floating.h
+++ b/code/math/floating.h
@@ -100,5 +100,12 @@ inline bool fl_equal(float a, float b)
 // rounds off a floating point number to a multiple of some number
 extern float fl_roundoff(float x, int multiple);
 
+/**
+ * @brief Determines if @a x falls between +/- @a e
+ *
+ * @param x Value to test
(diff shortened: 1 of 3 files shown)


MageKing17:
> The `fl_abs` changes look good but why did you remove `vm_vec_normalize_quick_mag` and `vm_vec_normalized_dir_quick_mag`?

@asarium, the description of commit ff3de3a2b466d68a9cc7abd92e839b0ce668e514 seems to sum it up pretty well to me.

asarium:
Ah, I didn't see that.

PR #1057, opened 2016-10-21
More file types for data/config

X3N0-Life-Form:
Adds .tbl, .tbm, .xml and .csv to the list of supported /config file types.


asarium:
GitHub is having a few issues which breaks the travis and probably also the appveyor builds. I'll restart them when that stuff is working again.

asarium:
I restarted the CI builds since the DNS issues should be solved by now.

asarium:
Have you tested if FSO is able to use the new file extensions properly?

X3N0-Life-Form:
Just wiped a up a quick test script to double checked : All 4 new file types can be read, while the nighlty from oct. 17 can't find any of them.

asarium:
Good, thank you for testing that.

PR #1063, opened 2016-11-04
cmake, gcc: Enable garbage collection

PeterMitsis:
Enable compiler and linker garbage collection to eliminate dead
code from the final image.

Signed-off-by: Peter Mitsis <peter.mitsis@gmail.com>


PeterMitsis:
I don't know about others' systems, but I have found on my own setup (Ubuntu with GCC), enabling garbage collection feature eliminates about 500 kB of unused code/data from the final fs2_open_xxx image.

Issue #965, opened 2016-09-11
gr.hasViewmode() not working correctly


AxemP:
gr.hasViewmode(VM_INTERNAL) should return true when in the cockpit view (and does with 3.7.4) but after around August 18, it no longer does. Instead VM_EXTERNAL_CAMERA_LOCKED is true when we're inside the player's ship. Bit flag problems maybe?

Here's a script that tracks which view modes are active.
http://pastebin.com/17TmpedV

asarium:
This was probably caused by the viewmode changes made by @z64555. Those changes introduced an incompatibility with scripts since `CAMERA_LOCKED` is now also true when in the cockpit.

niffiwan:
As suspected, seems to have been introduced by 9bce334c65fa2ad7aba7ab5655616721ca2ac7fb

z64555:
Oh lovely. Uh. I'll try to get around to looking at lua's implementation. I'd prefer to go with the new convention set by 1e40f71 since in my opinion it's more logical and flexible.

z64555:
I [made a branch](https://github.com/z64555/fs2open.github.com/tree/fix/issue_965) to address this issue.

PR #1065, opened 2016-11-07
Fixes issue #965 by making the VM_INTERNAL view state ignore the camera lock bit

z64555:
This'll get things going, but it's an impartial fix in the sense that existing scripts the use the old VM_EXTERNAL_CAMERA_LOCKED state are effectively outdated.


z64555:
Ok, maybe this comment will link #965

PR #1068, opened 2016-11-12
Check if the used SIMD set is supported by the CPU

asarium:
This will show an error if SDL detectes that the CPU does not support
the SIMD instruction set with which the binary was compiled. This is
currently only used by MSVC since only that actually offers the option
to choose between different SIMD optimizations.

### DIFF
--- a/cmake/toolchain-msvc.cmake
+++ b/cmake/toolchain-msvc.cmake
@@ -100,6 +100,7 @@ set(FSO_INSTRUCTION_SET ${MSVC_SIMD_INSTRUCTIONS})
 
 LIST(FIND POSSIBLE_INSTUCTION_SETS "${MSVC_SIMD_INSTRUCTIONS}" SET_INDEX)
 
+set(SIMD_SET "NONE")
 if (SET_INDEX LESS 0)
 	MESSAGE(STATUS "An invalid instruction set was specified, defaulting to no special compiler options.")
 else()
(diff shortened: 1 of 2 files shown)


MageKing17:
Does this actually work? I thought the executable crashed before it had a chance to execute any of our code, if it's compiled with an unsupported instruction set.

asarium:
I have no idea since I can't test that. My CPU has support for all the instruction sets so it's pretty hard to make it crash. `os_init()` should be executed pretty early but if that's not early enough then the code could be moved to `actual_main` which should hopefully be before any SIMD instructions are used.

The-E:
Tested this out on my Laptop, didn't seem to work correctly. An AVX exe would just quit without throwing up a warning message; I'll have to do some debugging on the machine in order to figure this out.

However, if this doesn't work in the engine, I think it can be made to work more easily in the Launcher.

asarium:
The compiler probably uses AVX instructions before osapi is able to do the availability check. The check could probably be moved to the begin of `actual_main`. Hopefully the compiler doesn't use AVX instructions before that.

asarium:
I moved the code to `actual_main` but I haven't tested it yet. I might need to do some adjustments since SDL is not initialized at that point.

asarium:
@The-E I fixed the updated code and now it should work properly.

The-E:
Nope, doesn't work. It breaks before running the actual init code because there's at least one static initializer (in this case, the Warp_camera initializer in camera.cpp:17) that can call AVX instructions.

asarium:
Well, thanks for testing. I guess this means that these changes can never work without major changes to how we handle static initializers.

PR #1076, opened 2016-11-21
Actually Fixes Issue 965.

z64555:
Previous fixed used a logical NOT (!) instead of a bitwise NOT (~), derp!
In addition, the LE_VM_INTERNAL state now ignores VM_CENTERING, so it won't go false if the camera is returning to center.

Adds scripting support for VM_CAMERA_LOCKED and VM_CENTERING. VM_EXTERNAL_CAMERA_LOCKED is now a soft-state that is equal to (VM_EXTERNAL && VM_CAMERA_LOCKED)

### DIFF
--- a/code/scripting/api/enums.cpp
+++ b/code/scripting/api/enums.cpp
@@ -392,6 +392,12 @@ flag_def_list Enumerations[] = {
 	},
 
 	{
+		"VM_CAMERA_LOCKED",
+		LE_VM_CAMERA_LOCKED,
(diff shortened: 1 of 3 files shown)


z64555:
Addresses Issue #965 
Corrects PR #1065 
Builds upon PR #848

z64555:
Currently rebasing the branch to ontop of current master, and addressing the changes made by @asarium. Should have it done soonish

PR #1082, opened 2016-11-28
timer: Increase precision of timer as well as reducing potential of o…

TheMatthew:
…verflows.

This creates a factor that is used to multiply the timestamp. It will
overflow less, (once every 300 years or so) and not lose as much precision
as the previous implementation.

The performance is slightly improved, there are no more branches and one
floating point multiply.

### DIFF
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -1,11 +1,11 @@
 /*
  * Copyright (C) Volition, Inc. 1999.  All rights reserved.
  *
- * All source code herein is the property of Volition, Inc. You may not sell 
- * or otherwise commercially exploit the source or things you created based on the 
+ * All source code herein is the property of Volition, Inc. You may not sell
+ * or otherwise commercially exploit the source or things you created based on the
  * source.
  *
-*/ 
+*/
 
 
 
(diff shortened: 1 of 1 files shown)


code/io/timer.cpp
```diff
@@ -26,9 +26,10 @@ static int Timer_inited = 0;
 
 
 #define MICROSECONDS_PER_SECOND 1000000
-
 #define NANOSECONDS_PER_SECOND 1000000000
 
+static long double Timer_factor;
```
asarium: This should be renamed to indicate that this factor should be used for converting nanosecond values.

niffiwan:
Appveyor error seems to be:
```
C:\projects\fs2open-github-com\code\io\timer.cpp(110): error C2220: warning treated as error - no 'object' file generated [C:\projects\fs2open-github-com\build\code\code.vcxproj]
C:\projects\fs2open-github-com\code\io\timer.cpp(110): warning C4244: 'return' : conversion from 'long double' to 'uint64_t', possible loss of data [C:\projects\fs2open-github-com\build\code\code.vcxproj]
```

asarium:
Sorry, I didn't notice that you updated your PR.

PR #1083, opened 2016-11-29
Fix Don't Spawn If Shot

DahBlount:
Previously beams would not set the Destroyed by weapon flag, which means spawn weapons shot by beams would spawn regardless of whether "don't spawn if shot" was set on the weapon. This has been fixed according to Axem.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -6260,7 +6260,7 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
 
 	// spawn weapons - note the change from FS 1 multiplayer.
 	if (wip->wi_flags[Weapon::Info_Flags::Spawn]){
-		if (!((wip->wi_flags[Weapon::Info_Flags::Dont_spawn_if_shot]) && (sw_flag == SW_WEAPON_KILL))){			// prevent spawning of children if shot down and the dont spawn if shot flag is set (DahBlount)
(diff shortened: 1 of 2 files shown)


code/weapon/weapons.cpp
```diff
@@ -6260,7 +6260,7 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
 
 	// spawn weapons - note the change from FS 1 multiplayer.
 	if (wip->wi_flags[Weapon::Info_Flags::Spawn]){
-		if (!((wip->wi_flags[Weapon::Info_Flags::Dont_spawn_if_shot]) && (sw_flag == SW_WEAPON_KILL))){			// prevent spawning of children if shot down and the dont spawn if shot flag is set (DahBlount)
+		if ((!wip->wi_flags[Weapon::Info_Flags::Dont_spawn_if_shot]) && (sw_flag == SW_WEAPON_KILL)){			// prevent spawning of children if shot down and the dont spawn if shot flag is set (DahBlount)
```
MageKing17: This conditional originally said "if this wasn't killed by a weapon, or it doesn't have the 'don't spawn if shot' flag, spawn subweapons." Now it says "if this doesn't have the 'don't spawn if shot' flag and was shot down, spawn subweeapons." These two are nowhere near equivalent.

This "fixes" beams spawning subweapons from "don't spawn if shot" weapons by making it impossible for the latter to ever spawn subweapons, which is not so much a fix as taking a tactical nuke to the problem and declaring victory when the specific issue you were having disappears.

PR #1084, opened 2016-12-01
Fiction viewer character conversion

Yarn366:
Unlike most text files that FSO loads, fiction viewer text doesn't undergo the character conversion necessary to support certain non-ASCII characters. This PR aims to fix that.

To avoid duplicating the code that converts the German letter ß to ss, I moved that code to both versions of the maybe_convert_foreign_characters function. This means that that function can now produce strings that are longer than than the originals, so I made the char* version use different strings for input and output.


code/parse/parselo.cpp
```diff
@@ -1665,30 +1666,94 @@ int maybe_convert_foreign_character(int ch)
 }
 
 // Goober5000
-void maybe_convert_foreign_characters(char *line)
+// Yarn - The capacity of out must be at least the value returned by
+// get_converted_string_length(in) (plus one if add_null is true).
+// Returns the number of characters written to out.
+size_t maybe_convert_foreign_characters(char *in, char *out, bool add_null)
```
asarium: 1. Since `in` isn't changed anymore it should be declared `const`.
2. This should either have a separate parameter that specified the length of the buffer or just use an `SCP_string` to avoid buffer overrun errors.

code/parse/parselo.cpp
```diff
@@ -1665,30 +1666,94 @@ int maybe_convert_foreign_character(int ch)
...
+			}
+		}
+	}
+}
 
-	for (SCP_string::iterator ii = line.begin(); ii != line.end(); ++ii)
-		*ii = (char) maybe_convert_foreign_character(*ii);
+// Yarn - Returns what the length of the text will be after it's processed by
+// maybe_convert_foreign_characters, not including the null terminator.
+size_t get_converted_string_length(char *text)
```
asarium: Same here, `text` should be `const char*`.

code/parse/parselo.cpp
Yarn366: Regarding 2: One of the two functions that calls this, process_raw_file_text in parselo.cpp, doesn't know the size of the output buffer, and there doesn't seem to be a way for it to get that size. What do you expect me to do (besides switch to SCP_string, which would probably entail numerous other changes that I'm hoping to avoid with this PR)?

Yarn366: That buffer is passed to `maybe_convert_foreign_characters` as the _input_ buffer. The issue is what's passed as the _output_ buffer, which is (perhaps confusingly) not `outbuf` but rather `mp`, a pointer to some part of `processed_text`, which is also just a pointer.

The-E: To be perfectly honest, I would really like us to use SCP_strings as much as possible in functions like this.

Yarn366: Just using SCP_strings would be better, I agree, but I don't think I have the experience necessary to determine the best way to convert the strings being passed to this function to SCP_string; those same strings get passed to or are generated from char*-only functions, so it looks like the conversion would involve more than just minor changes. If you or someone else can figure it out, then I will change my branch accordingly.

PR #1085, opened 2016-12-05
jettison-cargo sexp

Goober5000:
Upgrades and deprecates the original jettison-cargo-delay sexp by removing the unused delay parameter and adding a new parameter to control the speed of jettisoning.  It further fixes the physics calculations by using the dockpoint normal instead of the vector between the center of the host ship to the center of the cargo.

See also the thread on HLP: http://www.hard-light.net/forums/index.php?topic=92844.0


code/parse/sexp.cpp
```diff
@@ -11981,33 +11982,43 @@ void sexp_cap_waypoint_speed(int n)
...
 /**
  * Causes a ship to jettison its cargo
- * note that the 2nd arg (jettison delay) is not implemented
  */
-void sexp_jettison_cargo(int n)
+void sexp_jettison_cargo(int n, bool jettison_new)
 {
 	char *shipname;
 	int ship_index;
+	int jettison_delay_or_speed;
```
MageKing17: If this needs to be passed as a `float`, wouldn't it make more sense to store it as a `float`?

fred2/sexp_tree.cpp
```diff
@@ -2272,7 +2274,8 @@ int sexp_tree::get_default_value(sexp_list_item *item, char *text_buf, int op, i
 	h = item_handle;
 	index = item_index;
 	type = query_operator_argument_type(op, i);
-	switch (type) {
+	switch (type)
+	{
```
MageKing17: Are all these whitespace changes really a good idea when they'll just make tracking changes more difficult?

code/parse/sexp.cpp
Goober5000: It doesn't *need* to be passed as a float, but using a float preserves more options for setting the speed, as well as being consistent with the rest of the code.  As for why it's an int here, the only numbers that can be used with sexps are ints.

fred2/sexp_tree.cpp
Goober5000: There are only a few whitespace changes compared to the rest of the diff.  I made the changes so that those lines would match the rest of the function.  If you look at the function as a whole, it uses inconsistent formatting before the diff but consistent formatting after.

code/parse/sexp.cpp
Goober5000: But this function is used for `jettison-cargo-delay` as well as the new sexp.  And in that context, the value is an integral number of seconds.  (Note the name of the variable.)  You'd be casting an int to a float and then back to an int again.

MageKing17: Yes, the delay value *isn't* actually used for `jettison-cargo-delay`, which leaves us with exactly 0 instances of this variable actually being used as an `int`, which leads us back to the question of why we're using an `int` to store a value that is only ever used as a `float`.

PR #1086, opened 2016-12-08
Add catch clauses everywhere where parselo is used

asarium:
Before, if any parsing call failed it would throw an exception which
would have been propagated to the exception handler in main() which
terminates the program. This should fix that.

### DIFF
--- a/code/particle/ParticleManager.cpp
+++ b/code/particle/ParticleManager.cpp
@@ -90,22 +90,29 @@ EffectType parseEffectType() {
 void parseCallback(const char* fileName) {
 	using namespace particle;
 
-	read_file_text(fileName, CF_TYPE_TABLES);
+	try {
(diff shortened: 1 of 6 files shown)


code/particle/ParticleManager.cpp
```diff
@@ -90,22 +90,29 @@ EffectType parseEffectType() {
...
-	}
+			ParticleManager::get()->addEffect(constructEffect(name, type));
+		}
 
-	required_string("#End");
+		required_string("#End");
+	}
+	catch (const parse::ParseException& e)
+	{
+		mprintf(("TABLES: Unable to parse '%s'!  Error message = %s.\n", "pixels.tbl", e.what()));
```
asarium: That's a copy-paste error. That should be `filename`...

PR #1090, opened 2016-12-14
Enable extra GCC and Clang warnings again

asarium:
In one of my CMake changes I accidentally disabled the extra warning flags. This should fix that again and also fixes any warnings that were introduced while the option was disabled.


asarium:
The Appveyor failures should be fixed now. Apparently there is a bug in the Visual Studio 2015 compiler where it fails to parse a tripple brace initializer.

PR #1099, opened 2016-12-21
Remove unnecessary __VERSION__ usages in shaders

asarium:
The previous usages were always true since we require a more recent GLSL
version now.

### DIFF
--- a/code/def_files/main-f.sdr
+++ b/code/def_files/main-f.sdr
@@ -217,15 +217,11 @@ void GetLightInfo(int i, out vec3 lightDir, out float attenuation, out float spe
 	lightDir = normalize(lightPosition[i].xyz);
 	attenuation = 1.0;
 	specIntensity = SPEC_INTENSITY_DIRECTIONAL;
-#if __VERSION__ > 120
 	if (lightType[i] != LT_DIRECTIONAL) {
-#else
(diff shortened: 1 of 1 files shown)


code/def_files/main-f.sdr
```diff
@@ -251,12 +247,9 @@ vec3 CalculateLighting(vec3 normal, vec3 diffuseMaterial, vec3 specularMaterial,
    vec3 lightSpecular = vec3(0.0, 0.0, 0.0);
 	#pragma optionNV unroll all
 	for (int i = 0; i < MAX_LIGHTS; ++i) {
-#if __VERSION__ > 120
 		if (i > n_lights)
 			break;
-#endif
-		if(i > 0)
```
MageKing17: Er, why were these two lines removed?

SamuelCho: Yeah, I don't think those lines should have been removed. This basically will apply shadows meant for the scene directional light onto point lights and tube lights in forward lighting passes.

asarium: Damn, I misread the code and thought that this wasn't needed anymore...

chief1983: Don't feel bad, I misread it and almost defended it's removal thinking it would becom unreachable or something.

Issue #1074, opened 2016-11-20
Subtitle images without transparency get drawn incorrectly


AxemP:
Quite a few mods (like Blue Planet) had subtitle images that were just white on black that rendered additively in 3.7.4 and before. Now on the recent nightlies, that black is rendered regardless. See images below and test mod.

This is the current behavior
![screen0827](https://cloud.githubusercontent.com/assets/6625164/20464425/03a6546c-af0d-11e6-9718-ee6e242aee67.jpg)

This is how it was in 3.7.4
![screen0689](https://cloud.githubusercontent.com/assets/6625164/20464430/0aac9fdc-af0d-11e6-8807-3b03bb3596ab.jpg)

Here's a test mod
[subimagetest.zip](https://github.com/scp-fs2open/fs2open.github.com/files/602161/subimagetest.zip)

AxemP:
This appears to be fixed now. :)

AxemP:
Oops, my bad... One thing that seems to be broken with that new commit is apngs now get additive blending even if they have alpha channels.

SamuelCho:
I'm gonna need a test case of this. Especially if this is done through a script cuz there might be a parameter that's not being set in scripting.

AxemP:
[APNGTest.zip](https://github.com/scp-fs2open/fs2open.github.com/files/672037/APNGTest.zip)

Here ye go. Playing around with the gr.CurrentOpacityType didn't appear to affect anything. I am having to set it to ALPHABLEND_NONE to make sure nothing renders additively for my visual novel script. You can uncomment the appropriate line,

AxemP:
Okay, I tried that new change and it appears to work as intended... for animations anyway.

But one last related thing since this seems like its all related together. So I mentioned how I found I needed to start using ALPHABLEND_NONE because otherwise all my drawImage scripting stuff gets rendered additively, I didn't need to do that before PR #1104 and I thought that was just "the way it goes now". But now I find when I do that I can't affect the alpha of images done with gr.drawImage so... maybe not? :\ (I guess judging by the name, nothing has alpha?)

So here's an updated script file that tries to draw a normal image with a gradual alpha fade in with an apng animation doing the same thing. Just throw it in the mod folder from before, its just the script file.
[apng-sct.zip](https://github.com/scp-fs2open/fs2open.github.com/files/673188/apng-sct.zip)

[...]

Issue #961, opened 2016-09-11
Alpha blend filter type behavior has changed


asarium:
As @AxemP and I discovered while fixing #958, the alpha blend filter type was apparently broken for a long time but was fixed by @SamuelCho in the materials system merge.

The default lua opacity type was `GR_ALPHABLEND_NONE` but the bitmap drawing function behaved like normal alpha blending was applied. The rest of the code is probably also affected by this and may need to be reviewed. The bug only affects the `alpha` parameter for `gr_set_bitmap` so most the the code should be fine but if that alpha parameter is used then bitmap rendering may be broken.

SamuelCho:
material_determine_blend_mode already accounts for this. It returns src*alpha + dest*(one minus alpha) even if we say we don't want blending.

asarium:
The original version of `material_set_interface` used `material_determine_blend_mode` but I replaced that with `ALPHA_BLEND_ALPHA_BLEND_ALPHA` since an image without an alpha channel always disabled alpha blending which caused the credits issues.

SamuelCho:
Hmm, I'll have to dig up how opengl_setup_render_states handled the interface flag then to double check.

chief1983:
Did anyone confirm that this was fixed by #1104?

Goober5000:
Pinging @asarium for a fix check...

asarium:
The default opacity type is `GR_ALPHABLEND_FILTER` now so I guess this has been more or less fixed. Since there were no bug reports about broken scripts with 3.8 we can safely assume that this shouldn't matter too much.

PR #1104, opened 2016-12-23
Blending fixes for interface materials.

SamuelCho:
I've realized that we weren't setting the proper color factors when we render additive materials. I reverted the interface material setting function to switch between additive and alpha blending since the core problem has been fixed. This should address https://github.com/scp-fs2open/fs2open.github.com/issues/961 while also fixing https://github.com/scp-fs2open/fs2open.github.com/issues/1074.


asarium:
@AxemP Could you check if this solves #1074 for you?

The Appveyor build stalled for some reason. You should be able to restart the build by rebasing your branch.

PR #1113, opened 2016-12-27
Add previous on-screen profiling readout again

asarium:
This is basically the old code but it has been ported to the new tracing
system.

There is still some weirdness with the percentages but I have no idea
where that comes from since I don't know the code.


SamuelCho:
Thanks for putting this back in. I really appreciate it. Though I'm having problems when trying this out. I keep hitting either the assert in Line 38 or Line 247 in FrameProfiler.cpp. From what I can debug, it looks like certain events get the same timestamps as each other especially if their tracing scopes go into scope not long after each other which causes problems with the sorting function. This causes some non-deterministic weirdness where, as an example, "Post Move" gets sorted after "Ship Post Move".

I posted a screenshot of _bufferedEvents illustrating the problem. I hope this helps.

shippostmove_before_postmove

I'm guessing we might need a counter in FrameProfiler? Perhaps keep bumping that number up for every event and use that to resolve comparisons in the off chance timestamps are equal?

asarium:
That's a common issue with low resolution clocks. The other tracer implementations fix this by discarding events with a zero duration but I'm not sure if that's enough here. I added that code to the FrameProfiler but if that's not enough then I'll have to add a more involved implementation.

asarium:
I think I fixes the assertion issue by applying some creative timestamp offsets. The times are still a bit weird but that's better than causing an assertion.

SamuelCho:
This seems to work in Debug. Though I'm getting a problem where Page Flip is claiming to consume an absurd number of milliseconds but it doesn't seem to cause problems for the other samples. Though running the game and the real-time profiler in Release seems to cause a lock up. I haven't delved into the problem any further though but I wish I had more time to provide some better debugging info. I wonder if this is also a symptom of running in Windows.

asarium:
Alright, I restored some more old behavior and now "Page Flip" appears at the end of the list and has a sensible duration. However, the percentage is still 0.0% for all events and I don't know what's going on there...

asarium:
I can reproduce the hang on Windows. For some reason, the `num_parents` value of some samples is absurdly high which means that the loop that adds the indent characters takes a long time.

chief1983:
Should this have been a debase and not a merge of master?

asarium:
I used the "Resolve conflicts" feature of GitHub. Is something wrong with the merge changes?

chief1983:
Not sure, I just saw the commit was "merge master into feature/screenTracing", but I was under the impression that with PRs we want to keep them rebased onto master, and not just merge master in, to minimize merge commits.  If Resolve Conflicts creates additional merge commits it may be a tool we want to avoid.

asarium:
Usually I do a rebase but I wanted to try that feature. Do we have a policy against merging with master? IMO it's not a big problem since merge commits can be excluded from the history pretty easily.

chief1983:
Excluding is not really the issue, I think it's more the amount of branching and ordering of commits that annoys me personally.  I think it's easiest to work with if things are in as straight a line as possible.

asarium:
I prefer to rebase my branches when there are conflicts but I don't think we should make that a rule for the entire repository since it requires more work from those who don't have a lot of experience with git. Rebasing is a more advanced technique which can cause problems if you don't know what you are doing.

SamuelCho:
I seem to have gotten rid of the lock up if I axe the tracing scopes in obj_move_all_post save for PostMove. I wonder if the switch cases, despite having brackets around them, aren't really considering as having their own scope? Which ends up giving us situations where the destructor of the ship/weapon/etc post move tracing scopes end up getting called before the PostMove destructor.

asarium:
I removed the PostMove tracing scope since it was not needed. I have no idea why that would cause such problems but if it doesn't happen anymore it should be good enough. Do you have any idea why the percentage value is always zero?

SamuelCho:
Sorry I spoke too soon actually. Your recent push worked for a bit but then the game locked up in a different area related to sort colliders and find overlap colliders. The sorter seems to have trouble resolving if timestamps are equal _and_ duration is equal.

Though I have a brute force solution. It's kind of ugly and you might not like it. But I basically just have a counter in the tracing code that just keeps bumping up every time a new event is posted. That way we know for sure which start/end events came before/after each other instead of trying to trying to figure that out based on timestamps. https://github.com/SamuelCho/fs2open.github.com/commit/8c4a8b3d5eb84cc78c983f72843bbfecf2584de6

SamuelCho:
Oh as for percentages, I'm fine with those not working. I'm perfectly fine gutting that feature of the real-time profiler. If you want me to gut it myself, I'll be more than happy to.

asarium:
I added the ID code but now I am hitting the `samples[i].open_profiles == 0` assert again.

asarium:
Ah, good catch. I have merged your changes into my branch. My tests show that the code works perfectly now but you may want to conduct further tests.

SamuelCho:
Sweet, thanks. Yeah it's looking pretty good now. I'll merge it in once the CI has finished. Really thankful this is back in.

Issue #1115, opened 2016-12-29
No Collisions and No Disabled Self Destruct Flags not saving when FRED saves


AxemP:
Click those flags on in the ship editor, save in FRED and they won't be ticked when the mission reloads.

I tried ticking every flag and those were the only two that did that.

asarium:
@AxemP Could you please confirm that the issue is not present in master anymore?

AxemP:
Yeah its fixed now.

PR #1116, opened 2016-12-29
Fix for #1115

The-E:
The "no collide" parse object flag needs to be written as "no_collide" in the flags list, and the "no-disabled-self-destruct" flag needs to be parsed


fred2/missionsave.cpp
```diff
@@ -2997,7 +2997,7 @@ int CFred_mission_save::save_objects() {
 			if (shipp->flags[Ship::Ship_Flags::Scramble_messages])
 				fout(" \"scramble-messages\"");
 			if (!(objp->flags[Object::Object_Flags::Collides]))
-				fout(" \"no-collide\"");
+				fout(" \"no_collide\"");
```
asarium: It it intentional that this uses the underscore but the string in `missionparse.cpp` uses a hyphen?

The-E: It needs to be an underscore everywhere, this should be the case now

PR #1118, opened 2016-12-30
Model Code Cleanup

SamuelCho:
Cleaned up a bunch of  stuff related to draw_list which is now model_draw_list to be a better description of what it does. I also killed the transform class and replaced all uses of transform with matrix4. Killed the transform handling in model_draw_list as well and replaced it with the transform_stack that the OpenGL code uses. Unlit maps are gone since unlit specular doesn't look like ass anymore. Cut out a bunch of unused graphics code I found here and there.


code/math/vecmat.cpp
```diff
@@ -2802,4 +2810,4 @@ void vm_vec_transform(vec3d *dest, vec3d *src, matrix4 *m, bool pos)
 	dest->xyz.x = temp_dest.xyzw.x;
 	dest->xyz.y = temp_dest.xyzw.y;
 	dest->xyz.z = temp_dest.xyzw.z;
-}
+}
```
asarium: This newline should not have been removed.

PR #41, opened 2015-04-27
Ship_info vectorization, ship templates.

MageKing17:
Revamps ship_info with constructors, assignment operators, and a destructor, turns Ship_info into a vector, and reimplements ship templates. See also [Mantis 3083](http://scp.indiegames.us/mantis/view.php?id=3083).

PR #1124, opened 2017-01-01
change excluded_moods to be a bitstring instead of a vector, due to MMessage being a struct

Goober5000:
MMessage, being a struct, cannot have any classes or STL structures in it.  This will lead to memory leaks and/or crashes in FRED, since messages are being reordered, added, and removed.

There are two ways to fix this.  One is to make MMessage a full class, but that would be going down a rabbit hole.  The other is to make the `excluded_moods` field a bitfield.  Since it is unlikely that there will be more than 31 built-in moods, and even more unlikely that there will be more than 31 excluded moods, I have used this technique in the PR.

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -454,15 +454,22 @@ void message_parse(bool importing_from_fsm)
 		msg.mood = 0;
 	}
 
+	msg.excluded_moods = 0;
 	if ( optional_string("$Exclude Mood:")) {
 		SCP_vector<SCP_string> buff;
 		bool found = false;
 
 		stuff_string_list(buff); 
-		for (SCP_vector<SCP_string>::iterator parsed_moods = buff.begin(); parsed_moods != buff.end(); ++parsed_moods) {
+		for (SCP_vector<SCP_string>::iterator parsed_mood = buff.begin(); parsed_mood != buff.end(); ++parsed_mood) {
 			for (SCP_vector<SCP_string>::iterator iter = Builtin_moods.begin(); iter != Builtin_moods.end(); ++iter) {
-				if (!stricmp(iter->c_str(), parsed_moods->c_str())) {
-					msg.excluded_moods.push_back((int)std::distance(Builtin_moods.begin(), iter));
+				if (!stricmp(iter->c_str(), parsed_mood->c_str())) {
+					int exclude_index = static_cast<int>(std::distance(Builtin_moods.begin(), iter));
+					if (exclude_index <= 31) {
+						msg.excluded_moods |= (1 << exclude_index);
+					} else {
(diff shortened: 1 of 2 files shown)


MageKing17:
Er... you do understand that the only difference between a struct and a class is whether the members default to public or private, right? Stylistically, we tend to use structs for things that hold POD and classes for things that don't, but the only actual problem comes from using things like memcopy()/memmove()/memset() on the latter, which we now have static assertions to catch. I certainly don't see any of those being called on MMessage objects, so where is the memory leak, exactly? Because "being pushed onto a vector" is definitely not one.

Goober5000:
Yes, but a class has a few things - e.g. constructors and a destructor - that a struct traditionally does not.  I really should have said plain-old-data type instead of struct.

The reason I opened this PR is because the message vector is resized and recopied every time the OnOk function runs.  Then again, vector reallocation doesn't use memcpy or memmove internally, does it?  I had thought it behaved differently on POD types vs classes, but maybe that doesn't make sense.

MageKing17:
> Yes, but a class has a few things - e.g. constructors and a destructor - that a struct traditionally does not.  I really should have said plain-old-data type instead of struct.

"Traditionally" being the operative word; our distinction between structs and classes is purely a stylistic one. In actuality, if we changed all of our classes to just be structs with constructors and destructors, there would be absolutely no difference in behavior, which is why I was wanting to know where an actual instance of a memory leak involving `MMessage` could be found.

> The reason I opened this PR is because the message vector is resized and recopied every time the OnOk function runs.  Then again, vector reallocation doesn't use memcpy or memmove internally, does it?  I had thought it behaved differently on POD types vs classes, but maybe that doesn't make sense.

No, it doesn't. It uses the default assignment operator internally (or `std::move()` if the type has a move-constructor, which I think, by default, they do?). The default operator works perfectly fine on both POD data and STL containers, since I believe it merely recursively performs assignment on all the individual attributes (except arrays, which can't be assigned wholesale like that; that's the only special-case I'm aware of). In fact, if it weren't for a few customly-allocated pieces of data in `ship_info` (namely, subsystems and `vm_strdup()`-created strings), I wouldn't have needed custom constructors for PR #41, and the default assignment operator would have worked fine for implementing ship templates.

If "being used in a vector while containing STL containers" was your only concern, then it sounds like this PR can just be closed, because vectors are *meant* to be used with other STL containers, even ones contained inside structs.

Goober5000:
> If "being used in a vector while containing STL containers" was your only concern, then it sounds like this PR can just be closed, because vectors are meant to be used with other STL containers, even ones contained inside structs.

Yep.  My concern was prompted by both the vector resize and this bit of code:

```
	for (i=0; i<m_num_messages; i++)
		Messages[i + Num_builtin_messages] = m_messages[i];
```

But if the default assignment operator handles that properly, then there is no issue.

PR #1137, opened 2017-01-06
Add a lua loader that uses the CFile system

asarium:
This allows to use the built-in lua "require" function with our file
system. This will help organize complicated scripts and it will also
allow to use libraries which were designed for this system.

To enable this change, this includes a new mechanism which allows to
load scripts from the default file system so that the files don't need
to be included in every mod. The default files can be overridden by a
mod if a special mod-table option has been set, similar to how external
shaders are handled right now.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -3,6 +3,7 @@ Checks: "-*,\
 modernize-*,\
 -modernize-raw-string-literal,\
 -modernize-use-using,\
+-modernize-make-unique,\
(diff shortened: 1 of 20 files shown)


asarium:
There was a valgrind failure in the Lua code but I have no idea what causes it, especially since it only happens when the code is compiled by Clang. I added a suppression entry for that specific failure to fix the tests because I believe that this is a false positive.

PR #1147, opened 2017-01-18
Reimplement debug window

asarium:
This allows to show the debug output in a second window in real time.
This system could be expanded further to allow more interactivity but
the current status is good enough.

This is a cross-platform reimplementation of a system that was removed when antipodes was merged.

### DIFF
--- a/code/osapi/DebugWindow.cpp
+++ b/code/osapi/DebugWindow.cpp
@@ -1,0 +1,175 @@
+//
+//
+
+#include "DebugWindow.h"
+
+#include "globalincs/alphacolors.h"
+
+namespace {
+uint32_t get_debug_display() {
+	// If there are two or more monitors then we would like to display the window on the second monitor but it shouldn't
+	// be shown on the same monitor that the game is running on
+	// If there is only then then we'll just display it on that one
+	auto numDisplays = SDL_GetNumVideoDisplays();
+	
+	if (numDisplays < 1) {
+		// Some kind of error
+		return 0;
+	}
+	
+	if (numDisplays == 1) {
+		// We only have one display
+		return 0;
+	}
+	auto mainDisplay = os_config_read_uint("Video", "Display", 0);
+	if (mainDisplay == 1) {
+		// Game is on the second monitor => use the primary screen
+		return 0;
+	} else {
+		// Use the secondary screen
+		return 1;
+	}
+}
+}
+
+namespace osapi {
+
+DebugWindow::DebugWindow() {
+	os::ViewPortProperties attrs;
+	uint32_t display = get_debug_display();
+	attrs.display = display;
+
+	SDL_Rect size;
+	SDL_GetDisplayBounds(display, &size);
+	attrs.width = (uint32_t) size.w;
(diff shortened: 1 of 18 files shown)


code/osapi/DebugWindow.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+namespace {
+uint32_t get_debug_display() {
+	// If there are two or more monitors then we would like to display the window on the second monitor but it shouldn't
+	// be shown on the same monitor that the game is running on
+	// If there is only then then we'll just display it on that one
+	auto numDisplays = SDL_GetNumVideoDisplays();
+	
+	if (numDisplays < 1) {
+		// Some kind of error
```
z64555: Not sure what the proper error handler should be. Int3() is the first to come to mind, but I don't think that's the best.

Could put "TODO" in the comment to serve as a marker if we decide to defer putting in the error handler.

code/osapi/DebugWindow.cpp
```diff
@@ -0,0 +1,175 @@
...
+		}
+	}
+
+	// Don't run scripting here
+	gr_flip(false);
+
+	gr_use_viewport(os::getMainViewport());
+}
+
+float DebugWindow::print_line(float bottom_y, const LineInfo& line) {
```
z64555: The debug console code has word wrapping and whitespace handling, if you wanted a reference.

code/osapi/DebugWindow.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+
+#include "globalincs/pstypes.h"
+#include "osapi/osapi.h"
+
+namespace osapi {
+
+class DebugWindow {
```
z64555: Would you please doxy the rest of the class? Each method and the class itself should be documented, and so should any of the non-trivial members.

code/osapi/DebugWindow.cpp
asarium: Well, that code is the error handler. IMO simply defaulting to the primary screen if there was an error is a good solution.

code/osapi/DebugWindow.h
asarium: Yeah, I can do that.

code/osapi/DebugWindow.cpp
asarium: I'll take a look at those.

code/osapi/DebugWindow.h
asarium: Doxygen comments have been added.

code/osapi/DebugWindow.cpp
asarium: I have implemented word wrapping using the standard `split_str`.

z64555: Which library is split_str from? I don't see it in stdlib

MageKing17: It's in parselo.cpp; it's used for wordwrapping all throughout the FSO codebase.

z64555: Ok, split_str looks good to use, and it respects kerning of the font.

PR #1148, opened 2017-01-19
Make gr_bitmap obey clip offsets as it did before the OpenGL Core merge

### LINKED ISSUE #1141: Text cursor no longer appears
The OpenGL Core merge (edecb1ca7320c914be6fe450cc3e7753e8ee957f) broke the text cursor that appears when player is supposed to type something--for example, to enter a name for a new pilot. (The cursor graphic is ~~the third special character in the main font, the same one used for the weapons gauge~~ cursor1.ani.) The bug was introduced either in the first commit of that branch that I could compile (2022f6a792efa38538053a04b9f40c5ef4fc28f1) or earlier in that branch.

Yarn366:
This fixes #1141, where the cursor bitmap that's supposed to appear when the user is entering text is missing.

### DIFF
--- a/code/graphics/2d.cpp
+++ b/code/graphics/2d.cpp
@@ -1270,10 +1270,10 @@ void gr_bitmap(int _x, int _y, int resize_mode)
 	w = i2fl(_w);
 	h = i2fl(_h);
 
-	// I will tidy this up later - RT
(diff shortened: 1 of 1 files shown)


MageKing17:
Well, I can confirm this fixes the cursor, even though I have no idea why it doesn't seem to affect anything else...

PR #1151, opened 2017-01-20
Fixes an assertion due to insufficient amount of hats

The-E:
The logic for joy_down was missing a safety check, which could cause an assertion in Joystick::getHatDownTime if a joystick without a hat (or no joystick at all?) was attached. Was reported by Bryan See here: http://www.hard-light.net/forums/index.php?topic=93045.msg1839250#new

### DIFF
--- a/code/io/joy-sdl.cpp
+++ b/code/io/joy-sdl.cpp
@@ -902,27 +902,7 @@ int joy_down_count(int btn, int reset_count)
 
 int joy_down(int btn)
 {
-	if (btn < 0) return 0;
-
 	auto current = io::joystick::getCurrentJoystick();
 
-	if (current == nullptr)
-	{
-		return 0;
-	}
-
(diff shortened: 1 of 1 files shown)


code/io/joy-sdl.cpp
```diff
@@ -902,27 +902,7 @@ int joy_down_count(int btn, int reset_count)
...
-	}
-	else if (btn >= JOY_NUM_BUTTONS)
-	{
-		// Is hat
-		return current->getHatDownTime(0, hatBtnToEnum(btn), false) > 0;
-	} // Else, is a button
-
-
-	return current->isButtonDown(btn) ? 1 : 0;
+	return current->getButtonDownTime(btn) > 0.0f;
```
asarium: This probably crashes if there is no current joystick since then `current` would be `nullptr`.

The-E: Crap, you're right, pushed the wrong change

PR #1153, opened 2017-01-21
Fix linking error in unit tests

asarium:


MageKing17:
When is `main` already defined?

asarium:
It's from SDL which redefines `main` so that SDL can handle that to hide platform specific entry points.

MageKing17:
How do unit tests ever work without this fix, then?

asarium:
They work on Linux since SDL doesn't redefine `main` there. As far as I can tell the test were broken on Windows by some change in the past and it wasn't caught because, apparently, no one runs the test on Windows...

chief1983:
Makes sense to me, seems like a good idea to get these fixed.

PR #1154, opened 2017-01-22
Fixes two crashes in red-alert missions, cleans up and refactors some of the docking code

z64555:
First fix is when a deleted carried ship (one that has the "red-alert-carry" flag) is docked with another ship in the red-alert mission at start. Before the fix, if a carried ship was deleted, any docked object to the ship would not be properly undocked and would cause an access violation to the NULL location.

Second fix is when a deleted carried ship is on the escort list. Before the fix, it wouldn't be removed from the escort list.

### DIFF
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -781,20 +781,18 @@ void red_alert_store_wingman_status()
 void red_alert_delete_ship(int shipnum, int ship_state)
 {
 	ship* shipp = &Ships[shipnum];
+	int cleanup_mode;	// See ship.h for cleanup mode defines. SHIP_VANISHED etc.
 
-	int cleanup_mode = SHIP_REDALERT;	// See ship.h for cleanup mode defines
 	switch (ship_state) {
 	case RED_ALERT_DESTROYED_SHIP_CLASS:
-		cleanup_mode |= SHIP_DESTROYED;
+		cleanup_mode = SHIP_DESTROYED_REDALERT;
 		break;
 	case RED_ALERT_PLAYER_DEL_SHIP_CLASS:
-		cleanup_mode |= SHIP_DEPARTED;
+		cleanup_mode = SHIP_DEPARTED_REDALERT;
 		break;
 	default:
-#ifdef DEBUG
-		Warning(LOCATION, "Red-alert: Unknown delete state, assuming vanished.");
-#endif
-		cleanup_mode |= SHIP_VANISHED;
+		Assertion(false, "Red-alert: Unknown delete state '%i'\n", ship_state);
+		cleanup_mode = SHIP_VANISHED;
 		break;
 	}
 
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -29,6 +29,8 @@
 #include "missionui/redalert.h"
 #include "mod_table/mod_table.h"
 #include "model/model.h"
+#include "object/deadobjectdock.h"
+#include "object/objectdock.h"
 #include "ship/ship.h"
 #include "sound/audiostr.h"
 #include "sound/fsspeech.h"
@@ -787,6 +789,10 @@ void red_alert_delete_ship(ship *shipp, int ship_state)
 			Error(LOCATION, "Red Alert: asked to delete ship (%s) with invalid ship state (%d)", shipp->ship_name, ship_state);
 		}
 	}
+
+	// If this ship is docked with anybody, undock them all before deleting
+	dock_undock_all(&Objects[shipp->objnum]);
+	dock_dead_undock_all(&Objects[shipp->objnum]);
 
 	ship_add_exited_ship( shipp, Ship::Exit_Flags::Player_deleted );
 	obj_delete(shipp->objnum);
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -781,43 +781,25 @@ void red_alert_store_wingman_status()
 void red_alert_delete_ship(int shipnum, int ship_state)
 {
 	ship* shipp = &Ships[shipnum];
-	if ( (shipp->wing_status_wing_index >= 0) && (shipp->wing_status_wing_pos >= 0) ) {
-		if (ship_state == RED_ALERT_DESTROYED_SHIP_CLASS) {
-		hud_set_wingman_status_dead(shipp->wing_status_wing_index, shipp->wing_status_wing_pos);
-		} else if (ship_state == RED_ALERT_PLAYER_DEL_SHIP_CLASS) {
-			hud_set_wingman_status_none(shipp->wing_status_wing_index, shipp->wing_status_wing_pos);
-		} else {
-			Error(LOCATION, "Red Alert: asked to delete ship (%s) with invalid ship state (%d)", shipp->ship_name, ship_state);
-		}
-	}
 
-	// If this ship is docked with anybody, undock them all before deleting
-	dock_undock_all(&Objects[shipp->objnum]);
-	dock_dead_undock_all(&Objects[shipp->objnum]);
-
-	int cleanup_mode;	// See ship.h for cleanup mode defines (SHIP_VANISHED)
+	int cleanup_mode = SHIP_REDALERT;	// See ship.h for cleanup mode defines
 	switch (ship_state) {
 	case RED_ALERT_DESTROYED_SHIP_CLASS:
-		cleanup_mode = SHIP_DESTROYED;
+		cleanup_mode |= SHIP_DESTROYED;
 		break;
 	case RED_ALERT_PLAYER_DEL_SHIP_CLASS:
-		cleanup_mode = SHIP_DEPARTED;
+		cleanup_mode |= SHIP_DEPARTED;
 		break;
 	default:
-		cleanup_mode = SHIP_VANISHED;
+#ifdef DEBUG
+		Warning(LOCATION, "Red-alert: Unknown delete state, assuming vanished.");
+#endif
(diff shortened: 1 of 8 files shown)


z64555:
These two fixes stem from a test campaign from [Mantis 1185](http://scp.indiegames.us/mantis/view.php?id=1185).

code/missionui/redalert.cpp
```diff
@@ -788,11 +791,32 @@ void red_alert_delete_ship(ship *shipp, int ship_state)
...
+	int cleanup_mode;	// See ship.h for cleanup mode defines (SHIP_VANISHED)
+	switch (ship_state) {
+	case RED_ALERT_DESTROYED_SHIP_CLASS:
+		cleanup_mode = SHIP_DESTROYED;
+		break;
+	case RED_ALERT_PLAYER_DEL_SHIP_CLASS:
+		cleanup_mode = SHIP_DEPARTED;
+		break;
+	default:
+		cleanup_mode = SHIP_VANISHED;
```
asarium: The default case should also have a `break` at the end to make the style consistent.

code/object/object.h
```diff
@@ -304,9 +304,34 @@ void obj_client_post_interpolate();
...
+ *
+ * @returns Nonzero if docked, or
+ * @returns 0 if not docked
+ *
+ * @author Goober5000
+ */
 int object_is_docked(object *objp);
+
+/**
+* @brief Checks if the given object is dead-docked with anyone.
```
asarium: What does "dead-docked" mean? That should be explained in the documentation.

z64555:
I've got this labled as unfinished as I'm not sure if the red-alert delete process is the one to use. In my test with the mantis 1185 campaign it appears to work, but I am unsure if it will cause other issues.

There's also some optimizations I can do here, such as placing the undock-all methods within ship_cleanup

code/missionui/redalert.cpp
z64555: Done.

code/object/object.h
z64555: Added "@details" section on there. It's when the object is dead, and still has objects docked to it. Presumably this is to allow a dying parent object do deal damage only to its direct children instead of the entire dock chain (children of children).

Goober5000:
I'll take a look at this in the next day or two.  Solid work tracking down this problem. :+1:

code/missionui/redalert.cpp
```diff
@@ -788,11 +791,33 @@ void red_alert_delete_ship(ship *shipp, int ship_state)
...
+		break;
+	}
+
+	Objects[shipp->objnum].flags.set(Object::Object_Flags::Should_be_dead);
+	ship_cleanup(shipnum, cleanup_mode);
+	/*
+	ship_add_exited_ship( shipp, exit_flag );
```
asarium: `ship_cleanup` probably calls `ship_add_exited_ship` so this call isn't needed anymore?

Also, once you're done with your changes this code should be removed since commented code blocks are rarely helpful.

code/missionui/redalert.cpp
```diff
@@ -781,43 +781,25 @@ void red_alert_store_wingman_status()
...
+		cleanup_mode |= SHIP_DESTROYED;
 		break;
 	case RED_ALERT_PLAYER_DEL_SHIP_CLASS:
-		cleanup_mode = SHIP_DEPARTED;
+		cleanup_mode |= SHIP_DEPARTED;
 		break;
 	default:
-		cleanup_mode = SHIP_VANISHED;
+#ifdef DEBUG
+		Warning(LOCATION, "Red-alert: Unknown delete state, assuming vanished.");
```
asarium: Warnings are not shown in release so you don't need to wrap this in an `#ifdef`. Also, when could this warning be triggered? If it could only be triggered by a programming error then this should be an `Assertion`.

code/object/deadobjectdock.cpp
```diff
@@ -149,6 +144,16 @@ void dead_dock_remove_instance(object *objp, object *other_objp)
 		// delete it
 		vm_free(ptr);
 	}
+	else
+	{
+		// Raise an error if we're in a release build.
+		// Raise a warning if we're in a debug build, the user may be doing something silly
+#ifndef NDEBUG
+		Error(LOCATION, "Tried to undock an object that isn't docked!\n");
```
asarium: Same as above, this should either always be an error or it should always be a warning. If the warning should appear in release builds then you can use `ReleaseWarning`.

code/object/objectdock.cpp
```diff
@@ -771,6 +766,16 @@ void dock_remove_instance(object *objp, object *other_objp)
 		// delete it
 		vm_free(ptr);
 	}
+	else
+	{
+		// Raise an error if we're in a release build.
+		// Raise a warning if we're in a debug build, the user may be doing something silly
+#ifndef NDEBUG
+		Error(LOCATION, "Tried to undock an object that isn't docked!\n");
```
asarium: Same as above.

code/missionui/redalert.cpp
z64555: Most likely the default case (and the Warning) will only be triggered in a programming error, since this is executed only during red-alert missions. Will make the switch to `Assertion`, as you suggested

Goober5000:
@z64555 I really think the code would be a lot cleaner if you straight-up treated the `cleanup_mode` as an enum.  Look at all the special cases that have to be added when you check whether red-alert is present or not.  Red-alert, destroyed, and departed are really three distinct cases.

In conjunction with this, you can back out the code in `red_alert_delete_ship` that bitwise-ORs the previous mission's status to the red-alert flag.  Because of the previous paragraph, that information does not need to be passed to `ship_cleanup`.

z64555:
Attached is a 2-mission test campaign which tests wingman status as you are booting into a red-alert mission.

3.7.4 Behavior Marks a departed wingman as destroyed, and the destroyed wingman is marked as destroyed (as expected). I think this isn't intuitive, and should be marked the same in between.

Going the enum route seems to be cleaner than fussing around with bit flags, but there'd be two additional modes for red-alert cleanup: CLEAN_DESTROYED_RA and CLEAN_DEPARTED_RA. This would be needed because of the wingman statuses.

https://dl.dropboxusercontent.com/u/97612992/FreeSpace2%20Open/SCP/Test%20Missions/RAT.zip

z64555:
This is about as far as I want to dip into refactoring ship_cleanup(), as it's already outside the original scope of the PR. Pending feedback on the latest commit, I'd like to rebase/squash these down to something cleaner and get this merged in.

code/missionui/redalert.cpp
```diff
@@ -776,23 +778,25 @@ void red_alert_store_wingman_status()
...
+	switch (ship_state) {
+	case RED_ALERT_DESTROYED_SHIP_CLASS:
+		cleanup_mode = SHIP_DESTROYED_REDALERT;
+		break;
+	case RED_ALERT_PLAYER_DEL_SHIP_CLASS:
+		cleanup_mode = SHIP_DEPARTED_REDALERT;
+		break;
+	default:
+		Assertion(false, "Red-alert: Unknown delete state '%i'\n", ship_state);
+		break;
```
MageKing17: We still need to set `cleanup_mode` here, or it'll be uninitialized in a release build if this branch is ever taken.

z64555: Fixed that in amendment

code/missionui/redalert.cpp
```diff
@@ -781,20 +781,18 @@ void red_alert_store_wingman_status()
...
 	case RED_ALERT_PLAYER_DEL_SHIP_CLASS:
-		cleanup_mode |= SHIP_DEPARTED;
+		cleanup_mode = SHIP_DEPARTED_REDALERT;
 		break;
 	default:
-#ifdef DEBUG
-		Warning(LOCATION, "Red-alert: Unknown delete state, assuming vanished.");
-#endif
-		cleanup_mode |= SHIP_VANISHED;
+		Assertion(false, "Red-alert: Unknown delete state '%i', assuming vanished\n", ship_state);
```
MageKing17: "Assuming vanished" doesn't make sense in the `Assertion` message because execution is going to halt there.

z64555: It is very clear I don't know the distinction between our trap macros. :D

z64555:
Rebased and squashed, ready to be merged, pending approval.

code/missionui/redalert.cpp
```diff
@@ -776,23 +778,26 @@ void red_alert_store_wingman_status()
...
+		cleanup_mode = SHIP_DEPARTED_REDALERT;
+		break;
+	default:
+		Assertion(false, "Red-alert: Unknown delete state '%i'\n", ship_state);
+		cleanup_mode = SHIP_VANISHED;
+		break;
 	}
+
+	Objects[shipp->objnum].flags.set(Object::Object_Flags::Should_be_dead);
+	ship_cleanup(shipnum, cleanup_mode);
```
asarium: Does this properly clean up the object as well? The documentation of `ship_cleanup` says that object cleanup is handled elsewhere.

z64555: There's only one place in the codebase where obj_delete is called after ship_cleanup, and that's within the multiplayer code. I'll try to track down obj_delete's callers to see how it's used, and if it is needed in red_alert_delete_ship

z64555: `obj_delete_all_that_should_be_dead()` handles the object cleanup, and is called every frame from `obj_mova_all()`

asarium: Ok, thanks for the clarification.

PR #1156, opened 2017-01-24
Improves FSO's compatibility with the FS2 Demo data

MageKing17:
These commits prevent FSO from crashing when running with the data from the FS2 Demo. Some of these bugs could theoretically be encountered outside the demo, although it's unlikely. This PR makes no attempt to stop demo data from generating warnings; only errors or assertions.

Only commit I'm really iffy on is 559b07da63e52237d5ce66052bc2b07cec8164fa, but I never liked the hi-res interface art check anyway.

### DIFF
--- a/code/graphics/2d.cpp
+++ b/code/graphics/2d.cpp
@@ -995,10 +995,8 @@ bool gr_init(std::unique_ptr<os::GraphicsOperations>&& graphicsOps, int d_mode, 
 
 	if (gr_get_resolution_class(width, height) != GR_640) {
 		// check for hi-res interface files so that we can verify our width/height is correct
-		bool has_sparky_hi = (cf_exists_full("2_ChoosePilot-m.pcx", CF_TYPE_ANY) && cf_exists_full("2_TechShipData-m.pcx", CF_TYPE_ANY));
-
 		// if we don't have it then fall back to 640x480 mode instead
-		if ( !has_sparky_hi ) {
+		if ( !cf_exists_full("2_ChoosePilot-m.pcx", CF_TYPE_ANY)) {
 			if ( (width == 1024) && (height == 768) ) {
 				width = 640;
 				height = 480;
--- a/code/popup/popup.h
+++ b/code/popup/popup.h
@@ -107,4 +107,7 @@ void popup_kill_any_active();
 // change the text inside of the popup 
 void popup_change_text(const char *new_text);
 
+// Used if certain data is missing (e.g. running demo data).
+void popup_game_feature_not_in_demo();
+
 #endif
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -3080,6 +3080,10 @@ int bm_unload(int handle, int clear_render_targets, bool nodebug) {
 	bitmap_entry *be;
 	bitmap *bmp;
 
+	if (handle == -1) {
+		return -1;
+	}
+
 	int n = handle % MAX_BITMAPS;
 
 	Assert((n >= 0) && (n < MAX_BITMAPS));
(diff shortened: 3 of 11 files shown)


code/graphics/2d.cpp
```diff
@@ -995,10 +995,8 @@ bool gr_init(std::unique_ptr<os::GraphicsOperations>&& graphicsOps, int d_mode,
 
 	if (gr_get_resolution_class(width, height) != GR_640) {
 		// check for hi-res interface files so that we can verify our width/height is correct
-		bool has_sparky_hi = (cf_exists_full("2_ChoosePilot-m.pcx", CF_TYPE_ANY) && cf_exists_full("2_TechShipData-m.pcx", CF_TYPE_ANY));
-
 		// if we don't have it then fall back to 640x480 mode instead
-		if ( !has_sparky_hi ) {
+		if ( !cf_exists_full("2_ChoosePilot-m.pcx", CF_TYPE_ANY)) {
```
asarium: The reason that both files were checked could be that the files were distributed in separate VPs (I'm not sure how the High-Res files were distributed for the retail version). If that's the case then this check could pass even though there are files missing for the high resolution interface.

code/popup/popup.h
```diff
@@ -107,4 +107,7 @@ void popup_kill_any_active();
 // change the text inside of the popup 
 void popup_change_text(const char *new_text);
 
+// Used if certain data is missing (e.g. running demo data).
+void game_feature_not_in_demo_popup();
```
asarium: Since this is defined in popup.h/.cpp it should use the `popup` prefix.

code/graphics/2d.cpp
MageKing17: Of course they weren't distributed in separate VPs; if they were, then whichever one wasn't in sparky_hi_fs2.vp should've been removed from the check ages ago. And of course the check can pass even if there are files missing for the high resolution interface; that's true of either version. All you have to do is just stick two files called 2_ChoosePilot-m.pcx and 2_TechShipData-m.pcx into an installation that's missing sparky_hi_fs2.vp and poof, FSO will try to use nonexistant high-res interface art.

Fixing *that* problem is definitely outside the scope of this PR, although I would not mind having each individual screen use either the 640x480 or 1024x768 version of its own interface art depending on whichever was available, rather than it being determined globally.

code/popup/popup.h
MageKing17: I suppose that makes sense.

code/graphics/2d.cpp
asarium: Yeah, I just checked and both files are in the same VP. I have no idea why the code checked for both files but this change should not cause any issues.

Goober5000: The only reason that I can think of for including both checks is if a mod included custom pilot selection screen interface art.  FSPort does this, for instance.  However, even FSPort only changes the visible art, not the mask.  So I too think this commit should be safe.

MageKing17: Even if a mod wanted to include a custom mask, I'd be surprised if PCX were the format of choice.

Goober5000: Mask files need to use PCX because the button indexes correspond to indexes in the PCX palette.  Visible files don't have this requirement so they can use any format.

MageKing17: PCX files aren't the only files with an explicit palette; in 15 years, nobody made the mask code accept PNGs, at least?

Goober5000: That's a good question; I don't actually know.  In all the mods I've looked at that modify the mask file, PCX is the only format I've seen them use.  And only 8-bit PCX, at that.

MageKing17: Well, maybe I'll take a look at it in the future if I decide to try tackling graphics-related code from a different direction (and we should probably stop having this conversation on a closed PR only tangentially related :P ).

Goober5000: Heh.

PR #1158, opened 2017-01-25
Fix null vector normalization in ship select screen

asarium:
`vm_vector_2_matrix` was improperly guarded against bad parameters. In the
case I encountered the `uvec` was the negative of `fvec` which produced a
null vector when the cross product of the two vector was constructed.

These changes should fix that bug by ignoring the `uvec` parameter. Since
`vm_vector_2_matrix` and `vm_vector_2_matrix_norm` are basically the same, I
converted the former into an alias of the latter and only fixed that
code.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -813,50 +813,33 @@ void vm_vector_2_matrix_gen_vectors(matrix *m)
 //returns ptr to matrix
 matrix *vm_vector_2_matrix(matrix *m, const vec3d *fvec, const vec3d *uvec, const vec3d *rvec)
 {
-	vec3d *xvec=&m->vec.rvec;
-	vec3d *yvec=&m->vec.uvec;
-	vec3d *zvec=&m->vec.fvec;
+	vec3d fvec_norm;
+	vm_vec_copy_normalize(&fvec_norm, fvec);
 
-	Assert(fvec != NULL);
-
-	vm_vec_copy_normalize(zvec,fvec);
-
-	if (uvec == NULL) {
-		if (rvec == NULL) {     //just forward vec
-			vm_vector_2_matrix_gen_vectors(m);
-		}
-		else {                      //use right vec
-			vm_vec_copy_normalize(xvec,rvec);
-
-			vm_vec_cross(yvec,zvec,xvec);
-
-			//normalize new perpendicular vector
-			vm_vec_normalize(yvec);
-
-			//now recompute right vector, in case it wasn't entirely perpendiclar
-			vm_vec_cross(xvec,yvec,zvec);
-		}
+	vec3d uvec_norm;
+	if (uvec != nullptr) {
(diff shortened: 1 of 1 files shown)


code/math/vecmat.cpp
```diff
@@ -813,60 +813,45 @@ void vm_vector_2_matrix_gen_vectors(matrix *m)
...
 
 	*zvec = *fvec;
 
 	if (uvec == NULL) {
 		if (rvec == NULL) {     //just forward vec
-			vm_vector_2_matrix_gen_vectors(m);
+			vm_vector_2_matrix_gen_vectors(&temp);
 		}
 		else {                      //use right vec
+			*xvec = *rvec;
```
asarium: I'm not entirely sure about this change but as far as I can tell the `rvec` value was never used in the previous version.

code/math/vecmat.cpp
```diff
@@ -877,14 +862,29 @@ matrix *vm_vector_2_matrix_norm(matrix *m, const vec3d *fvec, const vec3d *uvec,
 		}
 	}
 	else {      //use up vec
+		*yvec = *uvec;
```
asarium: Same as above.

code/math/vecmat.cpp
DahBlount: I don't believe it was. Regardless, if you know the handedness of the coordinate system and you have two orthogonal vectors and know which one is the fvec, rvec, and/or uvec, you can generate the matrix.

MageKing17:
> In the case I encountered the `uvec` was the negative of `fvec` which produced a null vector when the cross product of the two vector was constructed.

Shouldn't something be done about the code that calls it with such vectors?

asarium:
The code which caused the issue in my case was the shadow setup code where the up vector of the eye matrix happend to be the same as the light vector. I don't think there is a universal way of fixing such code.

DahBlount:
The change to vm_vector_2_matrix should be sufficient imo. There's not much we can do about vectors that incidentally happen to be the equal and/or opposite when constructing a matrix except ignore the one that doesn't matter as much. Regardless, this does afford some extra protection within the functions themselves to avoid null vector issues overall.

z64555:
Ok, original code seems to have made the assumption that the input matrix was an Identity matrix, which explains why in vector_2_matrix_norm it was able to compute the matrix while outright ignoring the value of uvec (it just checked if it was nullptr or not).

New changes now respect the uvec, and should solve quite a few issues, and the added safety is quite welcome.

Before we should merge this in, however, the vector_2_matrix functions should have a serious doxy on them. The comments there leave out a lot of important info (like the fact that it assumed the matrix was an identity, and thus was an in/out param)

asarium:
Adding documentation is not in the scope of this changeset since these changes only fix the null vector normalization. If you want to add documentation then that should be done in a separate PR.

z64555:
Very well, will make a new PR with doxy changes shortly.

Issue #1103, opened 2016-12-22
Warnings in external libraries


chief1983:
As of the latest Xcode and macOS SDK, there are now no more warnings in our code, but there are one each in liblua and zlib.

```
CompileC build/lib/lua/FS2_Open.build/Debug/lua51.build/Objects-normal/x86_64/loslib.o lib/lua/loslib.c normal x86_64 c com.apple.compilers.llvm.clang.1_0.compiler
    cd /Users/cliffgordon/scp/fs2_open
    export LANG=en_US.US-ASCII
[...]

PR #1159, opened 2017-01-25
libupgrade/zlib_1_2_11: Update bundled zlib version to 1.2.11.

chief1983:
Fixes a warning in inflate.c on latest clang.  See issue #1103.


chief1983:
Should probably have someone test this with a PNG graphic or texture to ensure nothing is broken but I wouldn't expect anything to have changed.

asarium:
The changelog doesn't contain anything that should affect us so this should be relatively safe to merge. I'll give it a try tomorrow to make sure that PNG display still works.

PR #1160, opened 2017-01-25
libupgrade/libpng_1_6_28: Upgrade libpng to 1.6.28.

chief1983:
Fixes several published security vulnerabilities in most older versions.

> Vulnerability Warning
> 
> Virtually all libpng versions through 1.6.26, 1.5.27, 1.4.19, 1.2.56, and 1.0.66, respectively, have a null-pointer-dereference bug in png_set_text_2() when an image-editing application adds, removes, and re-adds text chunks to a PNG image. (This bug does not affect pure viewers, nor are there any known editors that could trigger it without interactive user input. It has been assigned ID CVE-2016-10087.) The vulnerability is fixed in versions 1.6.27, 1.5.28, 1.4.20, 1.2.57, and 1.0.67, released on 29 December 2016.

Along with the zlib upgrade, this one should  also involve testing that PNG loading still functions as expected.


asarium:
Fortunately, this security issue doesn't affect us since we don't use that functionality. Merging it would still be a good thing though. I'll test it tomorrow.

chief1983:
Yeah I figured it might not affect us, but I planned on trying to get PRs for several of our libraries.  If it doesn't affect us it could wait for after 3.8, but it's a minor version bump so it shouldn't hurt either.

PR #1161, opened 2017-01-26
Fix double rendering of effect ships

### LINKED ISSUE #653: PBR/HDR: 'FS1 Ship Select' ship effect in-mission doesn't display correctly
The select effect in the weapons load out is fine, but in mission it doesn't look like its doing the effect right.

It's hard to see in pictures exactly, but he attached mission will show the problem. Press 1 to trigger the effect on the invisible Fenris. (Rename to .fs2 of course)

[effecttest.txt](https://github.com/scp-fs2open/fs2open.github.com/files/356069/effecttest.txt)

asarium:
This fixes #653.

@AxemP Could you check if this is the desired behavior? In my tests the effects looks good now but just to be sure.

If this works then it's probably among one of the most time-intensive one-line fixes I have done...


AxemP:
Looks like this works as expected now. :)

PR #1162, opened 2017-01-28
Add flag to disable the widely-hated screen shudder effect

The-E:
Created by me because github flakes out on PhantomHoover's end


asarium:
Honestly, this is a gameplay change that should not be handled by a command line flag. It should either be a flag on the weapon, the ship, or a game_settings flag.

z64555:
The jitter only affects the camera angle, not the craft's orientation. Weapon's recoil effects (Weapon::$Recoil Modifier and Ship::$Ship Recoil Modifier) are a separate entity, so a cmdline flag for just the jitter isn't as game-breaking as it would be if it disabled recoil.

This PR skips over FF effects when in engine wash, which may or may not be a desired side-effect.

Also of note, is that the jitter intensity only considers the weapon's physics data, and ignores the craft's physics. It doesn't make sense for a heavier ship, like an Ursa, to jitter the same amount as a lighter ship, like a Perseus.

chief1983:
I still am not sure this is enough to warrant a new command-line flag, because a) we try to avoid that I thought, unless there is good/necessary reason, of which b) can't this already be controlled via a weapons table edit, with no code changes necessary at all?  Like, we're not talking about making a new flag somewhere else, it's something that can already be disabled.  I don't like the lightning effect either because it makes it harder to see what's going on, but that is kind of the point of the game, if I really wanted it off I'd assume I'd be in a minority and edit the mission if I really wanted to.  Or play in debug mode :P

Issues with how the effect behaves are one thing, I'd be for making the shudder more dependent on craft physics too, maybe a bigger shudder if you are linked (if it doesn't do that already).  But I'd still recommend against getting rid of the feature already by adding a command line option to our list.  Just no real justification here I feel.  Opens a floodgate to suggesting command line customization of too many other by design features of the original game.

Goober5000:
Ix-nay on this...

1) Doesn't warrant a command-line flag
2) This is an intrinsic part of the gameplay.  Shudder is also used for things like cutscenes.
3) Feature proposer has not addressed any of the criticism.

Issue #1080, opened 2016-11-25
Model error or code bug?


tomimaki:
In BP AoA mission "The Great Preservers" (bp-15.fs2) I got crash on debug build with this: `ASSERTION: "tswp->current_primary_bank < tswp->num_primary_banks" at aicode.cpp:8756`
It happens during attempt destroy Abalam. Someone can say this is model error, but I don't get crash on nightly 12 August 2016 or earlier. I can't check few later builds bcs of other bugs that were fixed, but on 20 august or later nightly I see this assertion. I suspect bitset flags conversion but can't be sure...

backtrace:
```
Thread 1 "fs2_open_3_7_5_" received signal SIGTRAP, Trace/breakpoint trap.
debug_int3 (file=0x555555f89938 "/home/tomi/fs2open.github.com/code/osapi/dialogs.cpp", line=350) at /home/tomi/fs2open.github.com/code/osapi/osapi.cpp:378
378		gr_activate(1);
(gdb) bt
#0  debug_int3 (file=0x555555f89938 "/home/tomi/fs2open.github.com/code/osapi/dialogs.cpp", line=350) at /home/tomi/fs2open.github.com/code/osapi/osapi.cpp:378
#1  0x0000555555cc07a4 in os::dialogs::Error (text=0x555574efe3b0 "Assert: \"tswp->current_primary_bank < tswp->num_primary_banks\"\nFile: aicode.cpp\nLine: 8756\n\n  /home/tomi/freespace2/fs2_open_3_7_5_x64-DEBUG : os::dialogs::AssertMessage(char const*, char const*, int,"...) at /home/tomi/fs2open.github.com/code/osapi/dialogs.cpp:350
#2  0x0000555555cbf94c in os::dialogs::AssertMessage (text=0x555555f39078 "tswp->current_primary_bank < tswp->num_primary_banks", filename=0x555555f3731e "aicode.cpp", linenum=8756, format=0x0) at /home/tomi/fs2open.github.com/code/osapi/dialogs.cpp:143
#3  0x000055555591a271 in ai_chase () at /home/tomi/fs2open.github.com/code/ai/aicode.cpp:8756
#4  0x0000555555927502 in ai_execute_behavior (aip=0x55555639ffa0 <Ai_info+34400>) at /home/tomi/fs2open.github.com/code/ai/aicode.cpp:13031
[...]

PR #1163, opened 2017-01-29
some riffs on PR #806

Goober5000:
In ship::clear(), @The-E removed

`memset(&weapons, 0, sizeof(ship_weapon));`

but did not replace it with the corresponding

`weapons = ship_weapon();`

This fixes that.  Additionally, this PR removes the need for one of the uses of `ship::clear()` by tweaking `hud_observer_init`.  Originally, the code used memcpy, but if the assignment operator works for `ai_info`, it should be suitable for `ship` too.

Edit:
Link to PR #806

### DIFF
--- a/code/hud/hudobserver.cpp
+++ b/code/hud/hudobserver.cpp
@@ -25,23 +25,9 @@ ai_info Hud_obs_ai;
  */
 void hud_observer_init(ship *shipp, ai_info *aip)
 {
-	// setup the pseduo ship and ai
+	// setup the pseudo ship and ai
+	Hud_obs_ship = *shipp;
 	Hud_obs_ai = *aip;
-	// (we used to do a memcpy here, but that doesn't work any longer, so let's just assign the values we need)
-	Hud_obs_ship.clear();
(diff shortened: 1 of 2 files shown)


tomimaki:
Is this fix for #1080?
I tested this PR and assertion doesn't occur anymore.

z64555:
I'm rather leery about using shallow-copy assignments on classes. Is there anything within `class ship` that would prevent its use?

MageKing17:
Yes; it looks like `warpin_effect` and `warpout_effect` are dynamically-allocated, which makes using the default assignment operator a bad idea here.

Goober5000:
I stand corrected.  I just looked over ai_info and it looks like there are no such constraints for that class.

Goober5000:
K, changes to hudobserver.cpp reverted.  That then just leaves the one line that was missed in #806.

asarium:
Nightstorm has commented on [the forum](http://www.hard-light.net/forums/index.php?topic=93099.msg1840292;boardseen#new) that there is something weird going on with secondary weapons in the newest nightly. Since these changes are the only changes that affect that part of the code it's the probably cause of this new issue.

PR #1166, opened 2017-01-31
Harmony to 1163's malady

z64555:
* Looks like a shallow copy wasn't enough for PR #1163 , so this fix creates ship_weapons::clear() which is called by ship_weapons's constructor and by ship::clear().

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -6425,7 +6425,7 @@ void ship_subsys::clear()
 	awacs_intensity = 0.0f;
 	awacs_radius = 0.0f;
 
-    weapons = ship_weapon();
+    weapons.clear();
 
 	memset(&submodel_info_1, 0, sizeof(submodel_instance_info));
 	memset(&submodel_info_2, 0, sizeof(submodel_instance_info));
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -5821,7 +5821,7 @@ void ship::clear()
 	target_shields_delta = 0.0f;
 	target_weapon_energy_delta = 0.0f;
 
-	weapons = ship_weapon();
+	weapons.clear();
 
 	// ---------- special weapons init that isn't setting things to 0
 	for (i = 0; i < MAX_SHIP_PRIMARY_BANKS; i++)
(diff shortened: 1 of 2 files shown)


asarium:
And how exactly does this fix the issue? As far as I can tell, the new code is functionally the same as the old one since the copy assignment of the new instance to the old field should effectively clear `ship::weapons`.

z64555:
I think I jumped the gun again. 🙄 
I was sure that the issue was with the shallow copy not completely setting all of ship_weapon's members, in particular the arrays. I'll try getting a test build and mission up to see if it'll fix this today.

MageKing17:
Default assignment operators do, in fact, copy arrays. I didn't see anything in the class that shouldn't work with the default assignment operator...

z64555:
Alright, so it doesn't really fix anything... I just did a test mission and was unable to repro [Nightstorm's behavior](http://www.hard-light.net/forums/index.php?topic=93099.msg1840400#msg1840400)

Would this be considered a cleaner code, then? Since creating an empty ship_weapons() just to have it copied over to an existing one doesn't seem like a particularly efficient way of doing things. (of course now I decide to bring this up, instead of mentioning it in #1163)

MageKing17:
It might be slightly more efficient; I'm not sure how the compiler optimizes each case, though, so it might just provide the *appearance* of being more efficient, with the compiler doing the same thing under the hood... on the other hand, you only replaced one of the two instances of `weapons = ship_weapon();` with `weapons.clear();`.

MageKing17:
Erm... you seem to have undone your other changes in the process of changing that line.

z64555:
How the heck did that happen?

MageKing17:
> (Ironically, neither `ship_weapon::clear()` nor `ship_weapon()` is precisely identical to the original memset, due to timestamp values.)

Er... weren't the memsets immediately followed by those timestamps getting set?

MageKing17:
Okay, I figured out how this PR could fix the problem; it's because there's a certain value that `ship_weapon::ship_weapon()` (or `ship_weapon::clear()`, in this PR) doesn't touch at all, and so calling `ship_weapon::clear()` leaves it whatever value it was before while assigning a brand-new instance of `ship_weapon` sets it to an uninitialized value.

I've opened PR #1168 to fix the underlying issue, but this change (resetting `weapons` in-place rather than overwriting it with a fresh instance) would not be incompatible with it.

PR #1170, opened 2017-02-01
Removes the Palette Manager

The-E:
A recent bryan see thread brought up the purpose of the gamepalette file that the engine tries to load on startup. A quick look through the code showed that this palette (and everything that goes with it, such as palman.cpp/.h) is essentially unused and very likely the vestigial remains of a previously existant palettized render mode.

I think it's therefore safe to chuck it all overboard.


z64555:
I noticed the removal of the neb2 code in there, I'm pretty sure they're not used in FS2, but does FSPort use them?

The-E:
They can't. The only way to get to that code is by setting the neb2_render_mode to NEB2_RENDER_POLY, which is only possible through the debug console.

z64555:
Ok. Hm. Isn't the nebula_render() you removed for the NEB2_RENDER_POLY mode?

PR #1171, opened 2017-02-01
Removes the "Object rotated too far" Assertion from aicode

The-E:
Since this particular assertion is singularly useless. Unless someone can come up with a good idea as to why we should keep it, and better yet, a theory as to why this assertion is something that needs to to be checked and what a modder or coder can do to fix the issue, I vote we remove this thing.


z64555:
Paging @Zacam

code/ai/aicode.cpp
```diff
@@ -1238,12 +1238,6 @@ objp->orient = objp_orient_copy; //-V587
 	} else {
 		vm_forward_interpolate(&desired_fvec, &curr_orient, &vel_in, delta_time, delta_bank, &objp->orient, &vel_out, &vel_limit, &acc_limit);
 	}
-	
-#ifndef NDEBUG
-if (!((objp->type == OBJ_WEAPON) && (Weapon_info[Weapons[objp->instance].weapon_info_index].subtype == WP_MISSILE))) {
-	Assertion(!(delta_time < 0.25f && vm_vec_dot(&objp->orient.vec.fvec, &tvec) < 0.1f), "A ship rotated too far. Offending vessel is %s, please investigate.\n", Ships[objp->instance].ship_name);
```
MageKing17: According to AppVeyor, this was the only use of `tvec`.

Goober5000:
These days, we almost always land on the assertion when the situation is excusable, e.g. when a ship is turning too quickly while warping out, to cite an example that happened to me.

But consider the situation when the assertion *would* be useful.  Recall when an obscure physics bug was tracked down because an uninitialized engine subsystem caused the ship to be hurled completely out of the mission area.  Suppose instead that the bug caused the ship to start spinning around like a top.  This assertion would catch that invalid condition just as the mission area bounds check caught the other invalid condition.

An assertion, however obscure, is better than an inexplicable crash, or worse, an unresponsive hang.

The-E:
But the fact that, as you say, this assert is almost always a false positive these days makes me think it shouldn't be an assert. It's not even Warning-type material; at most, this is something that may be printed to the log.

Goober5000:
That's a great point - instead of removing it completely, change it to print to the log.  My original thought was to make the check even fancier (e.g. testing whether the rotation persisted over multiple frames) but logging it is both less complex and more appropriate.

code/ai/aicode.cpp
```diff
@@ -1213,10 +1213,9 @@ void ai_turn_towards_vector(vec3d *dest, object *objp, float frametime, float tu
 	}
 
 	//	Dave Andsager: The non-indented lines here are debug code to help you track down the problem in the physics
-	//	that is causing ships to inexplicably rotate very far.  If you hit the Int3(), set the next statement to be
```
MageKing17: Given that it's an `mprintf` (which doesn't halt execution) and not a `Warning`, this comment may need further clarification.

code/ai/aicode.cpp
```diff
@@ -1213,10 +1213,9 @@ void ai_turn_towards_vector(vec3d *dest, object *objp, float frametime, float tu
...
+	//	that is causing ships to inexplicably rotate very far.  If you hit the Warning, set the next statement to be
```
MageKing17: Given that it's an `mprintf` (which doesn't halt execution) and not a `Warning`, this comment may need further clarification.

The-E:
Closed this, new PR without screwy history is #1208

PR #1173, opened 2017-02-02
Remove tcache_set from rendering API

asarium:
It wasn't used anywhere and has been superseded by the material system.
It would also interfere with a possible texture array usage.

### DIFF
--- a/code/graphics/grstub.cpp
+++ b/code/graphics/grstub.cpp
@@ -48,11 +48,6 @@ int gr_stub_save_screen()
 	return 1;
 }
 
-int gr_stub_tcache_set(int bitmap_id, int bitmap_type, float *u_scale, float *v_scale, int tex_unit = 0)
-{
-	return 0;
-}
-
 int gr_stub_zbuffer_get()
 {
 	return 0;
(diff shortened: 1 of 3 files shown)


code/graphics/grstub.cpp
```diff
@@ -48,11 +48,6 @@ int gr_stub_save_screen()
 	return 1;
 }
 
-int gr_stub_tcache_set(int bitmap_id, int bitmap_type, float *u_scale, float *v_scale, int tex_unit = 0)
```
MageKing17: You've removed `gr_stub_tcache_set()` but not `gr_opengl_tcache_set()`?

asarium: `gr_opengl_tcache_set` is actually still being used by the OpenGL code internally so we need to keep that.

MageKing17: ...But you said...

> It **wasn't used anywhere** and has been superseded by the material system.
> It would also **interfere with a possible texture array usage**.

???

asarium: Nonono, the external tcache API wasn't used anywhere *outside* the rendering code. Obviously, the OpenGL code still needs something to actually use a texture when rendering the game but the code outside of OpenGL doesn't need to know how that's done. That's the great thing about the new material system. All that nasty "low-level" stuff isn't exposed anymore.

MageKing17: Ah, so the problem wasn't with the thing it was doing, but in exposing it to external code. Gotcha.

asarium: Yes, exactly.

PR #1174, opened 2017-02-02
Remove some unneeded OpenGL version checks

asarium:
These are now unneeded since we check for OpenGL 3.2 and GLSL 1.50 support when starting the engine. After that, these conditions were always true which means that they weren't needed anymore.

### DIFF
--- a/code/graphics/opengl/gropengl.cpp
+++ b/code/graphics/opengl/gropengl.cpp
@@ -1641,7 +1641,7 @@ bool gr_opengl_is_capable(gr_capability capability)
 	case CAPABILITY_BATCHED_SUBMODELS:
 		return true;
 	case CAPABILITY_POINT_PARTICLES:
-		return true && !Cmdline_no_geo_sdr_effects;
+		return !Cmdline_no_geo_sdr_effects;
 	case CAPABILITY_TIMESTAMP_QUERY:
 		return GL_version >= 33; // Timestamp queries are available from 3.3 onwards
 	}
--- a/code/graphics/opengl/gropengl.cpp
+++ b/code/graphics/opengl/gropengl.cpp
@@ -1434,12 +1434,6 @@ static void init_extensions() {
 		Error(LOCATION,  "Current GL Shading Langauge Version of %d is less than the required version of %d. Switch video modes or update your drivers.", GLSL_version, MIN_REQUIRED_GLSL_VERSION);
 	}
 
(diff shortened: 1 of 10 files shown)


code/graphics/opengl/gropengl.cpp
```diff
@@ -1646,17 +1631,17 @@ bool gr_opengl_is_capable(gr_capability capability)
...
+		return !Cmdline_no_fbo && !Cmdline_no_deferred_lighting;
 	case CAPABILITY_SHADOWS:
-		return GL_version >= 32;
+		return true;
 	case CAPABILITY_BATCHED_SUBMODELS:
-		return (GLSL_version >= 150);
+		return true;
 	case CAPABILITY_POINT_PARTICLES:
-		return GL_version >= 32 && !Cmdline_no_geo_sdr_effects;
+		return true && !Cmdline_no_geo_sdr_effects;
```
MageKing17: Wouldn't it be a lot simpler to just `return !Cmdline_no_geo_sdr_effects`?

code/graphics/opengl/gropenglstate.cpp
```diff
@@ -334,7 +332,7 @@ GLboolean opengl_state::ClipPlane(GLint num, GLint state)
 
 GLboolean opengl_state::ClipDistance(GLint num, GLint state)
 {
-	Assert( (num >= 0) || (num < (int)(sizeof(clipdistance_Status) / sizeof(GLboolean))) || GL_version >= 30 );
+	Assert( (num >= 0) && (num < (int)(sizeof(clipdistance_Status) / sizeof(GLboolean))) );
```
MageKing17: What is the reason for the logic change here?

code/graphics/opengl/gropengl.cpp
asarium: Yes, it would be a lot simpler...

code/graphics/opengl/gropenglstate.cpp
asarium: I am pretty sure that the logical ORs were wrong here since this condition always evaluated to true (since `GL_version >= 30` was always true). Since this looks like a range check the logical ANDs seem to make a lot more sense.

MageKing17: Ah, yeah; I just saw the structural change from "A or B or C" to "A and B" without actually looking at what it was checking. The old check was obviously wrong.

PR #1178, opened 2017-02-03
Updates the gitignore for VS2017

The-E:


.gitignore
```diff
@@ -281,3 +281,6 @@ documentation/doxygen/*
 
 .idea/*
 projects/*
+/.vs
+/.vs
```
asarium: Why is this line duplicated?

MageKing17:
VS2017 puts crap in the base repo folder instead of the build directory?

.gitignore
The-E: Don't know, will correct

The-E:
Yeah. It's part if this new functionality they introduced where vs comes with cmake built in; apparently, they need to put a couple of json files in there.

PR #1179, opened 2017-02-03
Make sure that team glow colors don't go negativ

### LINKED ISSUE #651: PBR/HDR: Weird team color glow behavior
screen0657

The modded Cain has team color glows that seem to glow through the really long weapon trails being fired by the player.

screen0659

And here the glows seem to go teal when it fires a red beam. It only happens at the end of the beam.

None of this happens if I disable the team glows for the ship in the table.

Here's a test mod that displays the behavior:
https://drive.google.com/file/d/0B_q0tkX7VcSaRmhhVVo4UFJiakE/view?usp=sharing

asarium:
This fixes #651.

@SamuelCho Please take a look at the changes I made. They fix the issue and IMO they make sense but there may be a good reason for making these numbers go negativ.

### DIFF
--- a/code/def_files/main-f.sdr
+++ b/code/def_files/main-f.sdr
@@ -364,8 +364,8 @@ void main()
  #ifdef FLAG_TEAMCOLOR
 	vec4 teamMask = vec4(0.0, 0.0, 0.0, 0.0);
 	teamMask = texture(sMiscmap, texCoord);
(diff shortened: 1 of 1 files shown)


MageKing17:
What is the purpose of the subtraction in the first place?

asarium:
I don't know but the model looks funny if I remove it so it probably makes sense for some reason.

PR #1181, opened 2017-02-04
Add more graphics debug scopes for code called in the main frame

asarium:
These should help when debugging with APITrace or Renderdoc.


asarium:
These changes are included in #1186.

PR #1184, opened 2017-02-04
Search for path types regarless of case

### LINKED ISSUE #1182: CFile can't handle upper case data folders on Linux
The Diaspora development files use an upper case "Data" folder for the main data. When CFile tries to enumerate the files in that directory is searches for a path which uses the lower case "data" but that fails since that directory doesn't exist.

Since we already work around this issue for normal files it would make sense to emulate this for data directories as well.

asarium:
Previously, if a mod used an upper case "Data" folder (which is the case
for the Diaspora development files) the file enumeration would fail on
case sensitive platforms (e.g. Linux) since FSO searched for a path with
a lower case "data". These changes make it so that FSO looks for all
instances of the path type name regeardless of case. Since the real path
was also needed elsewhere I added a new cf_file field which stores the
actual path which can then be used by fopen operations.

This fixes #1182.

I haven't tested this on Windows yet. There shouldn't be any functional changes on that platform but I might have messed something up.

EDIT: I also need to add a unit test for this.

### DIFF
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -30,6 +30,8 @@
 #include <fnmatch.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#include <libgen.h>
+#include <cstdio>
 #endif
 
 #include "cfile/cfile.h"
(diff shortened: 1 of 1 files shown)


chief1983:
In the past, I thought we just asked mods to lowercase filenames like data and others.  Is adding code complexity to avoid that necessary?

asarium:
The problem is that some mods don't do that. Ideally there would be a way to detected this and then issue a warning or an Error message so that it gets caught early in the development process but I don't think that's possible on Windows. If it's possible then that would be the better solution than this.

asarium:
I have found a solution that will display an error on all platforms. These changes are not needed anymore.

PR #1187, opened 2017-02-05
Show error message when a data directory uses the wrong case

### LINKED ISSUE #1182: CFile can't handle upper case data folders on Linux

asarium:
Since Unix is case sensitive this should make sure that case errors are
cought early in mod development. This is an `Error` and not a `Warning`
since a mod with such a folder structure would never work under Unix and
as such it's justified to break mods that currently use this.

This fixes #1182.

### DIFF
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -16,14 +16,13 @@
 #include <sstream>
 #include <algorithm>
 #include <memory>
-#include <locale>
-#include <codecvt>
 
 #ifdef _WIN32
 #include <io.h>
 #include <direct.h>
 #include <windows.h>
 #include <winbase.h>		/* needed for memory mapping of file functions */
+#include <shlwapi.h>
 
 struct dir_handle_deleter {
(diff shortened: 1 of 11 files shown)


niffiwan:
AFAIK directories inside VPs are case insensitive; so technically a mod could develop on Windows with "Data", and still run fine on Unix when it packs everything into a VP.

Still, being a Linux user myself I'd prefer to enforce lower case Data \:)

Probably should advertise this change on the forums prior to committing it as well otherwise it'll break mods and irritate modders.

asarium:
That's what happens with Diaspora. The VPs work fine but the development files completely break the game on linux. I will announce the change on the forums after it has been merged.

asarium:
@niffiwan Here is the forum post: http://www.hard-light.net/forums/index.php?topic=93145.new#new

PR #1192, opened 2017-02-06
Restore original flag parsing behavior

### LINKED ISSUE #1191: No entries shown in weapons database
First reported [on the forums](http://www.hard-light.net/forums/index.php?topic=93140.msg1841055#msg1841055) by Nightstorm.

There are no entries displayed in the weapons database techroom. I can reproduce this bug with retail data so it's not a mod-data issue.

asarium:
The flagset changes misplaced the post-parsing flag processing so that
it was only executed if a special flag was found. That was not the case
previously and this restores that behavior. I also simplified the code a
bit while I was at it.

This fixes #1191.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -576,131 +576,128 @@ void parse_wi_flags(weapon_info *weaponp, flagset<Weapon::Info_Flags> wi_flags)
 
     bool set_nopierce = false;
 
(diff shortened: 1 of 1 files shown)


asarium:
I'm not sure if this actually fixes the issue properly. The `|`  that I added would simply let everything though.

asarium:
@The-E I found the actual issue and fixed it properly now.

PR #1198, opened 2017-02-07
Fix oversight in search path case checking

asarium:
The root directory may differ in case since that is already handled by
the mod support. This simply skips the case test if the current search
path is the root path.


asarium:
This isn't needed anymore since the commit that introduces this will be reverted soon.

PR #1199, opened 2017-02-07
Fix negative array index

asarium:
If there was only one object and that got deleted the code that keeps
track of the highest object index would cause a negativ array index. I
added a check to make sure that the index was positive but that still
means that Highest_object_index will be negativ after that function
returns. I looked at the usages of the value and that value seems to be
fine with that code so I think this should be safe.

### DIFF
--- a/code/object/object.cpp
+++ b/code/object/object.cpp
@@ -452,9 +452,11 @@ void obj_free(int objnum)
 
 	Assert(Num_objects >= 0);
 
-	if (objnum == Highest_object_index)
-		while (Objects[--Highest_object_index].type == OBJ_NONE);
-
+	if (objnum == Highest_object_index) {
+		while (Highest_object_index >= 0 && Objects[Highest_object_index].type == OBJ_NONE) {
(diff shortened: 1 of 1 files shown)


chief1983:
That caveat almost seems by design; any index-tracking variable has to have some default value for when there are no items in the index, and that default is usually a negative number for 0-based index lists.  So I wouldn't expect there to be an issue.

chief1983:
That said, C++ stops evaluating statements once it knows enough to decide the outcome right?  Wouldn't the original short syntax still work, slightly modified then?
```
while (Highest_object_index >= 0 && Objects[--Highest_object_index].type == OBJ_NONE);
```
Should prevent it from getting to the last half of the test I think.

The-E:
I would like to see that code simplified just for readability's sake though. Putting a decrement operator into a conditional like that makes me nervous.

asarium:
I simplified it for exactly that reason. The first time I looked at the loop I though it was an endless loop since I couldn't see any variable changes inside the loop body.

PR #1200, opened 2017-02-07
Stops debug log printing "Creating new directory"

chief1983:
Fix for log printing creating directory lines for directories that already exist.  Numerous "Creating new directory..." lines were appearing in my log, and it seems like checking to see if the directory already exists before creating it would make sense.

I could be wrong, but I assume that this code has already ensured that the path is a folder and not a filename, and that any relevant assert was not done deeper inside the logic that I'm now bypassing.  If I'm wrong on that it can be added here instead with a simple enough addition, but I wanted to keep it as simple as possible.  That check might not also be as cross-platform as just checking for existence so I wanted to avoid that for that reason too.

I have confirmed in testing that both the create statements for existing directories have stopped, and, when I removed one of the directories it constantly attempted to create before, it throws the line in the log only one time now.

### DIFF
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -14,6 +14,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <errno.h>
+#include <sys/stat.h>
 
 #ifdef _WIN32
 #include <io.h>
@@ -23,7 +24,6 @@
 #endif
 
 #ifdef SCP_UNIX
-#include <sys/stat.h>
 #include <glob.h>
 #include <sys/mman.h>
 #endif
@@ -621,6 +621,7 @@ void cf_create_directory( int dir_type )
 	int num_dirs = 0;
 	int dir_tree[CF_MAX_PATH_TYPES];
 	char longname[MAX_PATH_LEN];
+	struct stat statbuf;
 
 	Assertion( CF_TYPE_SPECIFIED(dir_type), "Invalid dir_type passed to cf_create_directory." );
 
@@ -634,13 +635,14 @@ void cf_create_directory( int dir_type )
 
 	} while( current_dir != CF_TYPE_ROOT );
 
-	
 	int i;
 
 	for (i=num_dirs-1; i>=0; i-- )	{
 		cf_create_default_path_string( longname, sizeof(longname)-1, dir_tree[i], NULL );
-		mprintf(( "CFILE: Creating new directory '%s'\n", longname ));
-        mkdir_recursive(longname);
+		if (stat(longname, &statbuf) != 0) {
+			mprintf(( "CFILE: Creating new directory '%s'\n", longname ));
+			mkdir_recursive(longname);
+		}
 	}
 }
 


code/cfile/cfile.cpp
```diff
@@ -634,13 +635,14 @@ void cf_create_directory( int dir_type )
...
 	} while( current_dir != CF_TYPE_ROOT );
 
-	
 	int i;
 
 	for (i=num_dirs-1; i>=0; i-- )	{
 		cf_create_default_path_string( longname, sizeof(longname)-1, dir_tree[i], NULL );
-		mprintf(( "CFILE: Creating new directory '%s'\n", longname ));
-        mkdir_recursive(longname);
+		if( stat( longname, &statbuf ) != 0) {
```
asarium: According to Appveyor, `stat` is not valid on Windows so we need a different solution here. Since you already determine the type of the path you could also check if it's a directory and if it isn't then we could issue a log message or a Warning to notify the user that something is wrong.

chief1983: I think that stat was not present because I did not notice that the stat.h include was in #ifdef for SCP_UNIX.  It was already used in cf_flush_dir but again I missed that it was #ifdef'd out because I was lazy and editing code in Sublime instead of an IDE.  Moving the include to all platforms should work as according to http://stackoverflow.com/questions/18100097/portable-way-to-check-if-directory-exists-windows-linux-c stat should be available on Windows.

And as I mentioned in the description, I didn't want to waste code complexity on checking something that I was pretty sure has already been checked by the time we get to this point.  If it looks likely that's not the case then I can add that in.

asarium: Ah, ok. If `stat` is available everywhere then that should work. I'm pretty sure that the path name isn't validated when the function arrives at the `stat`.

chief1983: So, I have added an Assertion if the stat() returns 0, to make sure that it is a folder.  It is working as it did before.  But, if I create a file in the place of the data folder that it tries to write logs to (MediaVPs_2014/data for instance), it doesn't seem to hit the Assertion, in fact it tries multiple times to call mkdir_recursive anyway.  And no error occurs.  This would actually be the same behavior without any changes in this ticket I think, it would try to make it, and fail silently.  In this case it just can't write a few logs, which it seems not to complain about either.  If it were a players data folder, that would be a bigger issue I imagine, but it doesn't seem there are any checks in the entire create directory code path to ensure this happens successfully.  I think my Assertion isn't reached, because if the data file exists where there should be a folder, stat() does not find it, since the path includes a trailing /.  With the trailing /, I don't think stat will match to a file, and only a folder, so it never even hits the Assertion when the rogue file exists there.

To fix this, I would probably need to do some change to make the path sent to stat() not include a trailing slash.  The effort and complexity added here to prevent something that hasn't been an issue feels out of the scope of this PR, which is simply to stop the log messages from occurring, which it was doing fine without the Assertion I just added.  So I think I'd prefer to just get rid of the Assertion and keep this one simple.  It isn't going to be introducing any issues it seems.  If the directory exists, it skips trying to create it.  If it's a file, it still tries to create it, just like it would have before my change.  Haven't had any issues with that behavior before so I'm not sure it needs to be fixed now.

Goober5000:
LGTM

PR #1207, opened 2017-02-08
Fix minor bug with parallel vector testing

z64555:
Previous version only checked if the two vectors were equal, which only half-way works for normalized vectors.

Two normalized vectors are indeed parallel if they are equal to each other, but they are also parallel if one is the negative scalar of the other (v2 = vm_vec_scale(v1, -1))

New method checks if the crossprod of the two is the zero vector. If using SSE2 optimizations, this should be only 2 ops. I commented in the function definition an alternative method should SSE2 be unavailable. The Best Fix (TM) would be to have a compile-time check for SSE2 optimizations and choose the appropriate version.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -615,14 +615,16 @@ vec3d *vm_vec_cross(vec3d *dest, const vec3d *src0, const vec3d *src1)
 	return dest;
 }
 
-// test if 2 vectors are parallel or not.
 int vm_test_parallel(const vec3d *src0, const vec3d *src1)
 {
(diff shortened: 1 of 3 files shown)


z64555:
I had just thought of something, I'll brb.

MageKing17:
Technically, since `vm_vec_equals()` calls `fl_equal()`, it still uses a threshold... but since that threshold uses `FL_EPSILON`, it's not really all that comparable.

z64555:
Checking for floating point values to be equal to each other has been known to be problematic in the past, so fl_equal was used to allow for minute deltas between the two. FSO is a game, so 100% precision is not needed in most instances.

PR #1208, opened 2017-02-09
Changes the "Object rotated too far" assertion to a log message

The-E:
As per discussion for PR #1171, only without the bad rebase.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -1213,10 +1213,9 @@ void ai_turn_towards_vector(vec3d *dest, object *objp, float frametime, float tu
 	}
 
 	//	Dave Andsager: The non-indented lines here are debug code to help you track down the problem in the physics
-	//	that is causing ships to inexplicably rotate very far.  If you hit the Int3(), set the next statement to be
+	//	that is causing ships to inexplicably rotate very far.  If you see the message below in the log, set the next statement to be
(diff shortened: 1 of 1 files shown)


chief1983:
I know we didn't want to have this keep breaking the game, but an mprintf isn't even a halt, which is what would be required to fix this.  I would have thought a warning, which is only hit in debug builds anyway, would be more useful for fixing the issue if it happens to a dev?  And if it is happening to a modder often, there is probably something to fix there too.

The-E:
That's where I disagree. As far as I can tell, there is no good definition of what "too far" means in context. I am pretty much convinced that whatever error this was designed to catch is no longer fatal, and as a result, I see it pop up more as a false positive than anything else.

What I'm trying to say is, if you're running into a physics issue, you are far more likely to see it pop up in gameplay than you are to see this assert fire; for us coders, the log message is enough of a starting point to work out the error, but for someone who isn't, and who is just hitting this error because its underlying assumptions are faulty, it's a red herring.

chief1983:
That's all I'm confused on, how is the log message enough to aid a coder when the event in question has passed?

The-E:
Because we can set a breakpoint at the point where it's firing after we've confirmed that something is actually going wrong. I don't see how interrupting gameplay for what is most likely to be a false positive is helping things.

PR #1211, opened 2017-02-10
[Feature] Ngon Radar

z64555:
This features adds the Ngon radar gauge, which is a derivative of the standard radar gauge. Instead of the standard's oval plot area, the Ngon radar gauge has a polygonal plot area.

The projection algorithm is the same as the standard's radar, so all the Ngon does is clamp the blip to the edges of the plot area.

I also refactored the standard radar plotting code a bit, there was some clamping in the original code that didn't work right, so blips could be drawn on top of the oval's edges. I'm not counting this as a bugfix, though.

Included is a test .tbm that activates a hexagonal plot area with "points up."

[radar_test-hdg.tbm.txt](https://github.com/scp-fs2open/fs2open.github.com/files/767768/radar_test-hdg.tbm.txt)

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -1227,7 +1227,7 @@ int find_intersection(float* s, const vec3d* p0, const vec3d* p1, const vec3d* v
 	vm_vec_cross(&crossA, v0, v1);
 	vm_vec_cross(&crossB, &v2, v1);
 
-	if (vm_vec_equal(crossA, ZERO_VECTOR)) {
+	if (vm_vec_equal(crossA, vmd_zero_vector)) {
 		// Colinear
 		return -1;
 	}
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -1216,6 +1216,31 @@ float find_nearest_point_on_line(vec3d *nearest_point, const vec3d *p0, const ve
 	return dot/mag;
 }
 
+int find_intersection(float* s, const vec3d* p0, const vec3d* p1, const vec3d* v0, const vec3d* v1)
+{
+	// Vector v2 forms an edge between v0 and v1, thus forming a triangle.
+	// An intersection exists between v0 and v1 if their cross product is parallel with the cross product of v2 and v1
+	// The scalar of v0 can then be found by the ratio between the two cross products
+	vec3d v2, crossA, crossB;
+
+	vm_vec_sub(&v2, p1, p0);
+	vm_vec_cross(&crossA, v0, v1);
+	vm_vec_cross(&crossB, &v2, v1);
+
+	if (vm_vec_equal(crossA, ZERO_VECTOR)) {
+		// Colinear
+		return -1;
+	}
+
+	if (!vm_test_parallel(&crossA, &crossB)) {
+		// The two cross products are not parallel, so no intersection between v0 and v1
+		return -2;
+	}
+
+	*s = vm_vec_mag(&crossB) / vm_vec_mag(&crossA);
+	return 0;
+};
+
 //make sure matrix is orthogonal
 //computes a matrix from one or more vectors. The forward vector is required,
 //with the other two being optional.  If both up & right vectors are passed,
--- a/code/radar/radar.cpp
+++ b/code/radar/radar.cpp
@@ -360,6 +360,27 @@ void HudGaugeRadarStd::pageIn()
 	bm_page_in_aabitmap( Radar_gauge.first_frame, Radar_gauge.num_frames );
 }
 
+void HudGaugeRadarStd::clampBlip(vec3d* blip)
+{
+	float max_radius;
+	float hypotenuse;
+
+	max_radius = i2fl(Radar_radius[0] - 5);
+
+	// Scale blip to radar size
+	vm_vec_scale(blip, i2fl(Radar_radius[0])/2.0f);
+
+	hypotenuse = _hypotf(blip->xyz.x, blip->xyz.y);
+
+	// Clamp to inside of plot area, if needed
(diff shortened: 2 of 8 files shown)


z64555:
I put the unfinisehd tag on because I haven't tested the behavior of the radar icons yet, which probably needs a bit more refactoring of the stardard radar.

code/radar/radar.cpp
```diff
@@ -367,43 +388,25 @@ void HudGaugeRadarStd::plotBlip(blip *b, int *x, int *y)
...
-		max_radius = i2fl(Radar_radius[0] - 5);
-
-		if (hypotenuse >= max_radius)
-		{
-			clipped_x_dist = max_radius * (new_x_dist / hypotenuse);
-			clipped_y_dist = max_radius * (new_y_dist / hypotenuse);
-			new_x_dist = clipped_x_dist;
-			new_y_dist = clipped_y_dist;
-		}
+		new_pos = ZERO_VECTOR;
```
The-E: For some unfathomable reason, this line breaks MSVC 2015. Apparently, you can only assign ZERO_VECTOR to something as part of its declaration (i.e. vec3d x = ZERO_VECTOR; is fine, the code above breaks things).

The easiest solution would be to initialize new_pos to ZERO_VECTOR and invert the zdist check.

code/math/vecmat.cpp
```diff
@@ -1208,6 +1208,31 @@ float find_nearest_point_on_line(vec3d *nearest_point, const vec3d *p0, const ve
...
+	// Vector v2 forms an edge between v0 and v1, thus forming a triangle.
+	// An intersection exists between v0 and v1 if their cross product is parallel with the cross product of v2 and v1
+	// The scalar of v0 can then be found by the ratio between the two cross products
+	vec3d v2, crossA, crossB;
+
+	vm_vec_sub(&v2, p1, p0);
+	vm_vec_cross(&crossA, v0, v1);
+	vm_vec_cross(&crossB, &v2, v1);
+
+	if (vm_vec_equal(crossA, ZERO_VECTOR)) {
```
The-E: Similarly, ZERO_VECTOR isn't going to work in this context.

code/radar/radar.cpp
```diff
@@ -359,6 +359,27 @@ void HudGaugeRadarStd::pageIn()
 	bm_page_in_aabitmap( Radar_gauge.first_frame, Radar_gauge.num_frames );
 }
 
+void HudGaugeRadarStd::clampBlip(vec3d* blip)
```
The-E: "blip" is a struct defined in radarsetup, gcc and clang both think that choosing a different name for this parameter would be good (Or, alternatively, you could pass in the actual blip here?)

code/radar/radar.cpp
```diff
@@ -359,6 +359,27 @@ void HudGaugeRadarStd::pageIn()
...
+{
+	float max_radius;
+	float hypotenuse;
+
+	max_radius = i2fl(Radar_radius[0] - 5);
+
+	// Scale blip to radar size
+	vm_vec_scale(blip, i2fl(Radar_radius[0])/2.0f);
+
+	hypotenuse = _hypotf(blip->xyz.x, blip->xyz.y);
```
The-E: "_hypotf" does not exist on *NIX, the canonical name is just "hypotf"

code/radar/radar.cpp
The-E: It should be noted that VS2017 fixes this issue.

code/radar/radarngon.cpp
```diff
@@ -0,0 +1,58 @@
+#include "radar/radarngon.h"
+
+#include "math/vecmat.h"
+
+#include <math.h>
+
+void HudGaugeRadarNgon::clampBlip(vec3d* blip) {
```
The-E: This blip is still causing issues for the CI bots

code/radar/radarngon.cpp
```diff
@@ -0,0 +1,58 @@
...
+	float hypotenuse;
+
+	hypotenuse = _hypotf(blip->xyz.x, blip->xyz.y);
```
The-E: _hypotf must be changed to hypotf.

code/radar/radarngon.cpp
```diff
@@ -0,0 +1,58 @@
...
+	v->xyz.z = 0.0f;
+
+	vm_vec_sub2(v, p);
+}
+
+// Default to a hexagon
+HudGaugeRadarNgon::HudGaugeRadarNgon()
+	: arclen(PI2 / i2fl(6)), offset(0.0f), r_min(cos(PI / i2fl(6))) {}
+
+HudGaugeRadarNgon::HudGaugeRadarNgon(int num_sides, float offset)
```
The-E: "offset" shadows a member variable

FSCyborg:
z64555 has let me finish up this PR.  Turns out the changes were already on the local branch, just not on the pull request.  The only actual work that I can claim is reversing some white space changes and rebasing it to the current code to make sure there weren't merge conflicts.  

I will test this soon (tm) to make sure it works.

FSCyborg:
Apparently need to redo the cmake config before testing.  I will give this a try this weekend.  

Also need to double check that correct angle units are used, as per z64555's comment on discord.

FSCyborg:
Having trouble compiling, so I haven't been able to test yet.  But I am adding changes that asarium suggested.

FSCyborg:
Fixed the compilation issue (always remember to update your submodules). About to compile and test.  Let's see what happens!!

Goober5000:
@FSCyborg what were the test results?

@The-E can you re-review based on FSCyborg's latest changes?

FSCyborg:
So far tests are showing no gauge differences.

Issue #1205, opened 2017-02-08
Ship lab initial ship position regression


chief1983:
As seen at the beginning of https://youtu.be/QbhOd6WM31g when you load the first model, you are staring at the tail-end of the ship.

I had thought this was a more recent regression, but when I went to confirm, I can see the behavior all the way back to 3.6.16.  In 3.6.14, when you first select a ship, Model Rotation was not disabled.  It seems that disabling that became a default behavior in 3.6.16.  Maybe this isn't the bug I initially suspected it was, but without that setting enabled by default, you are always staring at the tail of the ship, instead of a more appealing angle, such as first seen when you do enable model rotation.  Not sure why it seems model rotation was disabled by default, but it's annoyed me since then to be staring at thrusters when I open a ship up.  Wondering if there would be a quick way to set the initial camera angle, or if we should restore it defaulting to a rotating model.

X3N0-Life-Form:
IIRC, shiplab model rotation was disabled by default because it was disliked by a number of modders, so restoring the old default would likely annoy a number of people.

Goober5000:
There ought to be a way to set the initial angle of the ship without causing it to continuously rotate.

TRBlount:
There is. Would we perhaps want a view similar to the default view you find standard across most 3D modeling apps?
Example: ![](http://puu.sh/tXgap/70648836aa.png)

chief1983:
That is about exactly what it looks like when you first enabled model rotation, before it has started to spin.  That'd be perfect.

TRBlount:
Alright, if that's the case the matrix for it should be somewhere in the code. I'll set about finding it and seeing the best way to apply it.

TRBlount:
Is this a good default orientation?
![](http://puu.sh/tXuWX/90f578a6ff.png)

chief1983:
That's great :)

Goober5000:
Resolved by #1212.

PR #1212, opened 2017-02-10
Change Initial Orientation in the Lab

TRBlount:
Related to #1205.
Instead of being greeted by the rear end of a ship, players will now get this as the default orientation.

The reason for the additional matrix rotation by angles is because matrix transforms are not commutative. That is to say, composing a matrix out of the angles P= -0.6, B = 0.0, and H = 4.0 is NOT the same transform as composing it as P= -0.6, B = 0.0, and H = 0.0 and rotating it around the H axis by 4.0 radians.

### DIFF
--- a/code/lab/lab.cpp
+++ b/code/lab/lab.cpp
@@ -90,7 +90,11 @@ static char Lab_bitmap_filename[MAX_FILENAME_LEN];
 
 static float Lab_viewer_zoom = 1.2f;
 static vec3d Lab_viewer_pos = ZERO_VECTOR;
-static matrix Lab_viewer_orient = IDENTITY_MATRIX;
+/* Lab_viewer_orients default value is a PBH of -0.6, 0.0, 4.0 in radians, with the rotations applied in P, B, H order. 
+ * This aligns the model so it points to the bottom left corner of the screen, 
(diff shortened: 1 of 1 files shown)


code/lab/lab.cpp
```diff
@@ -90,7 +90,7 @@ static char Lab_bitmap_filename[MAX_FILENAME_LEN];
 
 static float Lab_viewer_zoom = 1.2f;
 static vec3d Lab_viewer_pos = ZERO_VECTOR;
-static matrix Lab_viewer_orient = IDENTITY_MATRIX;
+static matrix Lab_viewer_orient = { { { {{{ -0.653643608f, 0.427322835f, 0.624616086f }}}, {{{ 0.0f, 0.825335622f, -0.564642429f }}}, {{{ -0.756802499f, -0.369074941f, -0.539475381f }}} } } };
```
asarium: Could you add a small comment here to document why you chose these values?

TRBlount:
@asarium Done.

asarium:
@The-E Should this be merged now or after the next release? This is more a feature than a bug fix so I am not sure if it's ok to merge it yet.

The-E:
I think it's okay. It improves usability and fixes what is arguably a regression.

PR #1216, opened 2017-02-12
Create main window at 0,0 if its the same size as the display

### LINKED ISSUE #1051: -window command-line argument not working
I tried launching FSO with -window, both using the MSVC 2015 debugger and specifying -window in the command-line using wxLauncher.  In all attempts, FSO launched full-screen.

asarium:
This fixes #1051 by explicity telling SDL to create the window in the
display corner instead of centering the window. This has caused some
problems because the window borders aren't visible anymore in this case
which lead to some confusion.

@Goober5000 Please test if this fixes your issue since I don't have a system where this bug happens.

### DIFF
--- a/freespace2/SDLGraphicsOperations.cpp
+++ b/freespace2/SDLGraphicsOperations.cpp
@@ -163,9 +163,28 @@ std::unique_ptr<os::Viewport> SDLGraphicsOperations::createViewport(const os::Vi
 		windowflags |= SDL_WINDOW_RESIZABLE;
 	}
 
+	SDL_Rect bounds;
+	if (SDL_GetDisplayBounds(props.display, &bounds) != 0) {
+		mprintf(("Failed to get display bounds: %s\n", SDL_GetError()));
+		return nullptr;
+	}
+
+	int x;
+	int y;
+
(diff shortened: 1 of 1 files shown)


chief1983:
Does this affect full-screen (borderless) windowed mode, or does that still work?  I thought as long as this behavior was known, it was actually kind of useful to have a way to have a windowed mode that doesn't look windowed.

asarium:
This doesn't affect the borderless window mode since the result would have been the same anyway. Also, the borderless window mode is working perfectly or are you having issue with it?

chief1983:
I seem to remember something about borderless only being supported oncertain platforms?

asarium:
That was before SDL2. Now it's supported everwhere.

Goober5000:
@asarium I tested this per request.  The window still appears full-screen, without any visible borders or title bar.  However, the log message "SDL: Creating window at 0,0 because window has same size as desktop." does appear which makes diagnosing problems easier.

asarium:
It's unfortunate that the placement is still wrong but, as you said, the log message should help if someone else has this issue in the future.

Goober5000:
True.

PR #1223, opened 2017-02-13
Change get_max_ammo_count_for_bank to return "correct" max missile count

Yarn366:
Most of the game considers a ship bank's max missile count to be the bank's capacity divided by the missile's cargo size, rounded to the nearest integer. So, for example, the GTF Ares' first missile bank, which has a capacity of 90, can hold 23 Rockeyes, which have a cargo size of 4 (90 / 4 = 22.5, rounded to 23).

The `get_max_ammo_count_for_bank` function, however, doesn't do the rounding, so in the scenario above, it thinks the Ares' first missile bank can hold only 22 Rockeyes (90 / 4 = 22.5, truncated to 22). (This function is used by FRED to obtain max missile counts, and is also used by the game to determine how many missiles a support ship can load into a bank during a red-alert mission.)

This PR fixes this discrepancy by adjusting the `get_max_ammo_count_for_bank` function to match the rest of the game.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -16087,7 +16087,7 @@ int get_max_ammo_count_for_bank(int ship_class, int bank, int ammo_type)
 		capacity = (float) Ship_info[ship_class].secondary_bank_ammo_capacity[bank];
 		size = (float) Weapon_info[ammo_type].cargo_size;
 		Assertion(size > 0.0f, "Weapon cargo size for %s must be greater than 0!", Weapon_info[ammo_type].name);
-		return (int) (capacity / size);
+		return fl2ir(capacity / size);
 	}
 }
 


MageKing17:
> This function is used by FRED to obtain max missile counts, and is also used by the game to determine how many missiles a support ship can load into a bank during a red-alert mission.

Wait, it's actually used by FSO somewhere? Well... as silly as it is, maybe that technically counts as a balance change? ...How likely a scenario is that? It *only* triggers in red-alert missions, and only if the ship would have a fractional missile of at least .5? Well... that sounds unlikely enough to probably not be a serious balance change, but maybe other people disagree.

chief1983:
I have to admit, it sounds silly that it ever rounds up.  But I imagine changing that behavior to always truncate would be more apparent to users.

niffiwan:
yeah, especially with large secondaries like bombs. e.g. the default Boanerges in Slaying Ravana would only get 6x Cyclops in banks 2 & 3 instead of 7x.

I think the support-ship-reload during a red alert mission is enough of an edge case that it wouldn't affect balance. I think consistency is more important that that tiny balance change.  

Also; why does the red-alert-support-ship-reload use a different code path to the normal mission support-ship-reload anyway?

chief1983:
Are available support ship munitions a thing?  Maybe that is tracked across red alert missions?

asarium:
@chief1983 I don't think there is. The mechanic just limits the maximum amount of ammunition that can be refilled to the amount available at the start of the mission. It seems like a silly behavior since you can circumvent that by rearming before the next mission begins but that wouldn't be the first silly behavior in the engine...

@Yarn366 The function is actually also used by the scripting API but that shouldn't cause issues as well so I'm going to merge this.

PR #1226, opened 2017-02-13
Fix briefing word wrapping when a script changes the font

asarium:
This happened in BtA where a script changed the font after rendering its
UI. Since the briefing code didn't account for that, the word wrapping
code used the wrong font for calculating how long a line would be which
made the display look wrong.

### DIFF
--- a/code/missionui/missiondebrief.cpp
+++ b/code/missionui/missiondebrief.cpp
@@ -1835,6 +1835,9 @@ void debrief_text_init()
 		}
 	}
 
+	// Make sure that the text wrapping and the rendering code use the same font
+	font::set_font(DEBRIEFING_FONT);
+
 	Num_text_lines = Text_offset = brief_color_text_init("", Debrief_text_wnd_coords[gr_screen.res][2], default_debriefing_color, 0, 0);	// Initialize color stuff -MageKing17
 
 	fsspeech_start_buffer();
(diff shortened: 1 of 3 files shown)


code/missionui/missiondebrief.cpp
```diff
@@ -1835,6 +1835,9 @@ void debrief_text_init()
 		}
 	}
 
+	// MageKing17 - Set the font so that wordwrapping in brief_color_text_init() calculates based on the same font as the debriefing itself.
```
MageKing17: It's a little odd for you to add a comment saying it's from me without moving it from somewhere else.

asarium: I copied that code from the [rendering function](https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/missionui/missiondebrief.cpp#L1931) so technically it it your code.

MageKing17: I'm pretty sure I wasn't the first person to add a font-setting line to the codebase... but regardless, leaving this comment untouched is just being cruel to future `git blame`-ers. :P

asarium: Ok, I'll just use the comment I used for the other instances.

asarium: @MageKing17 I made the changes you requested. Is there anything else that needs to be changed?

MageKing17:
Nope!

code/missionui/missiondebrief.cpp
MageKing17: Nope!

Issue #967, opened 2016-09-12
OS X: Drive root error when run from Finder


chief1983:
![screenshot 2016-09-12 10 18 04](https://cloud.githubusercontent.com/assets/145825/18441316/3dec5e76-78d2-11e6-8e6e-741a0cd6e621.png)

I get this error when double-clicking a recent nightly installed into my FS2 directory.

chief1983:
For those reading emails, the error is:  I get this error when double-clicking a recent nightly installed into my FS2 directory.

SamuelCho:
So initial testing tells me that this has something to do with the g-buffers. Deferred lighting doesn't work on Mac but disabling it should look okay. I'd say just disable deferred lighting as a temporary workaround for now until I can find a solution. Thanks again for unblocking me chief.

asarium:
Is this still an issue?

chief1983:
Not sure, let me try the most recent nightly and report back.

chief1983:
Yup, still get that error.

asarium:
Can you step through `cfile_in_root_dir` in a debugger and determine why `cfile_in_root_dir` fails? Or you could add debug prints to determine what `buf` contains.

chief1983:
If I could figure out how to get debugging to work on a Mac I could but I still haven't had much luck in that department yet, or we'd be a lot closer to fixing that sound crash too :(

asarium:
Unfortunately, I cannot help you with that since I don't have a Mac. However, I can help you if you have questions about the CMake setup.

jg18:
I think I tracked this down.

Apparently this was a known issue and there was an attempt to fix it in actual_main() freespace.cpp. There were a couple issues:

(1) The relevant code block checked for the APPLE_APP preprocessor macro, which AFAICT is not defined at least on the Xcode I'm using (7.2.1). Changing APPLE_APP to __APPLE__ fixes this.

(2) The code uses the function SetCurrentDirectory() which is not in the macOS APIs AFAICT. Switching to the POSIX equivalent chdir() fixes this.

With these changes, FSO now runs when run from the Finder.


The fix brings up a couple of other issues:

- The APPLE_APP macro is also used in os_init_cmdline() in cmdline.cpp. Through a similar fix, I've replaced the GetCurrentDIrectory() function with the POSIX function getcwd().

[...]

PR #1227, opened 2017-02-13
Fix for #967 - Enable APPLE_APP in Mac CMake project

chief1983:
fix/967: Define APPLE_APP to fix some issues introduced on Mac in the CMake conversion (may have other effects we aren't yet aware of).

Specifically fixes being able to double click the app in the finder (issue #967), as opposed to having to launch through the finder.  Before this returned an error, now it will run as expected, with the last saved command line settings.

Also unifies a three-part ifdef into a two part by using our standard DIR_SEPARATOR.

GetCurrentDirectory and SetCurrentDirectory had to be replaced with more standard equivalents as they don't seem to be part of any current Mac framework, if they ever were.

The target_compile_definitions line is paired alongside the EXE_GUI_TYPE set because if the Mac build is ever supported in a non-bundled fashion, those two lines would need to be addressed as a pair.

### DIFF
--- a/cmake/platform-darwin.cmake
+++ b/cmake/platform-darwin.cmake
@@ -1,5 +1,6 @@
 
 MESSAGE(STATUS "Doing Mac OSX specific things...")
+target_compile_definitions(platform INTERFACE APPLE_APP)
 SET(EXE_GUI_TYPE MACOSX_BUNDLE)
 
 FIND_LIBRARY(COCOA_LIBRARY Cocoa)
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -948,26 +948,23 @@ void os_init_cmdline(int argc, char *argv[])
 	reset_cmdline_parms();
 
 	FILE *fp;
-	
+
 	if (!has_cmdline_only_flag(argc, argv)) {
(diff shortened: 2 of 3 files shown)


asarium:
`GetCurrentDirectory`/`SetCurrentDirectory` are usually defined by our Unix compatibility layer but apparently that doesn't happen on Mac for some reason. That doesn't matter though since `chdir` is much more portable.

chief1983:
Strange, since Apple and Windows are the only ones that would need them.  The only instances are within an APPLE_APP ifdef or Fred.  Not to mention I can't find any place in the code base they are defined in any situation.
```
cliffgordon@Cliffs-MacBook-Pro:~/scp/fs2_open [master] $ git grep GetCurrentDirectory
code/cmdline/cmdline.cpp:                       GetCurrentDirectory(MAX_PATH_LEN - 1, data_path);
fred2/fred.cpp: GetCurrentDirectory(MAX_PATH_LEN - 1, Fred_exe_dir);
cliffgordon@Cliffs-MacBook-Pro:~/scp/fs2_open [master] $ git grep SetCurrentDirectory
freespace2/freespace.cpp:       SetCurrentDirectory(path_name);
```

asarium:
Hmm, maybe I removed the other instances as part of my "stubs.cpp must die" changes.

PR #1233, opened 2017-02-15
Cfile tools cmake fix

The-E:
Fixes a wrong reference for the cmake configuration and simplifies the commandline usage instructions.


code/cfilearchiver/cfilearchiver.cpp
```diff
@@ -318,24 +318,6 @@ void print_instructions()
 	printf( "Creates a vp archive out of a FreeSpace data tree.\n\n" );
 	printf( "Usage:     cfilearchiver archive_name src_dir\n");
 	printf( "Example:   cfilearchiver freespace /tmp/freespace/data\n\n");
-	printf( "Directory structure options:\n" );
```
chief1983: What exactly makes these unneessary? It seems like it could be useful information, although it seems a bit outdated?

The-E: I think the only alternative that makes sense is to link to http://www.hard-light.net/wiki/index.php/FS2_Data_Structure
After all, that's a much better explanation of what file is expected where than what can or should be expressed in a commandline output.

chief1983: That works for me.

asarium:
@The-E Could you rebase this to trigger another Travis build? It seems like that build never started.

MageKing17:
No, it was AppVeyor that never started; I just went into the webhook and redelivered that request (it was real easy to spot; apparently the original request timed out, so it was the only warning icon on the page).

PR #1240, opened 2017-02-18
minor readme updates

vazor222:
This is a quick and simple PR to add a note at the beginning of the readme about submodule updates, and fixes a minor typo.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -6,7 +6,7 @@ FreeSpace2 *S*ource *C*ode *P*roject
 
 Building
 --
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -6,6 +6,8 @@ FreeSpace2 *S*ource *C*ode *P*roject
 
 Building
 --
+Before you do anything, make sure you have updated your git submodules, either by running git submodule update --init --recursive or by cloning the repository with the --recursive flag.<br/>
```
asarium: Commands should be put into `code blocks` to separate them from the rest of the text. You can do that by enclosing it in "`" characters.

vazor222: Good point, will update that.

PR #1243, opened 2017-02-18
z64555's Undo system! This is a generic undo system, so you can save …

z64555:
…any single item type (not C arrays) and do undo/redo stuff on them.

This system uses pointers to the data you're saving, so it can't currently save items that are in a dynamic container.

### DIFF
--- a/code/globalincs/undosys.h
+++ b/code/globalincs/undosys.h
@@ -10,6 +10,66 @@
 
 #include <utility>
 
+/**********************************************************
+Usage:
+First, you need an instance of the Undo_system within your module:
+	{
+		Undo_styem Undo_controls;
+	}
+
+From now on, you can save and restore items. (Just be sure to save them before you change them!)
+	{
+		Undo_controls.save(Control_config[z]);             // 1. Save the item!!
+		Control_config[z].bind(cid(CID_KEYBOARD, KEY_UP)); // 2. Make changes to the item!!
+		Undo_controls.undo();                              // 3. Undo's the changes you did in step 2!
+		Undo_controls.redo();                              // 4. Redo's the changes you did in step 2!
+	}
+The undo system can save any type of data, you don't have to have an instance of a system per data type.
+!!NOTE!! C style arrays and strings should be wrapped in a std::array when saving. Simply passing the array/string's head won't save the entire array.
+{
+	Undo_controls.save(std::array<int, JOY_AXIS_ACTIONS>(Axis_map_to));
+}
+
+In several cases, the client code would like to know exactly what was changed. The undo system provides a mechnism for doing this:
+	{
+		Undo_controls.save(Control_config[z], &Control_config[0]);   // Save the item, and a reference to its container. Here, we use the location of the first item in Control_config as our reference
+		Undo_controls.save(Axis_map_to[z], Axis_map_to);             // Saves an axis mapping, here, we use the array's head pointer as our reference (since a int[] is the same as a int*)
+
+		std::pair<const void*, const void*> ref = Undo_controls.undo();    // ::undo() returns a std::pair<const void*, const void*>, the first member referening the item that changed, and the second referencing the item's container (if we provided it)
+
+		// Once we've stored the return value fron ::undo, we can found out what the item is by comparing the second member with containers that we've been saving
+		if (ref.second == &Control_config[0]) {
+			cout << "I'm a button!";
+			Tab = static_cast<Config_item>(ref.first).tab;  // Sets the selected tab within the Controls Config menu. We do a static cast here so we can access the Config_item::tab member
+		} else if (ref.second == Axis_map_to) {
+			cout << "I'm an axis!";
+			Tab = SHIP_TAB;                                 // Sets the selected tab within the Controls config menu. Can't do a cast on this one, since it's a simple C array. But we already know where the axes are kept (on the ship tab)
+		} else {
+			cout << "I don't know what I am!!! D:";         // Usually an error condition. Depending on the client code this can be fatal or just a minor setback
+		}
+	}
+
+Lastly, you can make stacks of undo operations, so that a single op in the system can undo/redo multiple item changes.
+	{
+		Undo_stack stack;  // We'll want to save multiple changes as a single operation. So we have to make a stack.
+
+		for (int i = 0; i < JOY_AXIS_ACTIONS; ++i) {
+			stack.save(Axis_map_to[i], Axis_map_to);		// This example saves a C style array. Does the same thing without a wrapper, but wasteful
+	}
+
+	Undo_controls.save(stack);	// Save the stack as a single item!
+	std::pair<const void*, const void*> ref = Undo_controls.undo();
+	// Undo rolls through the items as they were saved, should your data operate on the same location, this will apply the changes in the correct sequence
+	ref.second == Axis_map_to;
+	ref.first == &Axis_map_to[0];     // ::first references the _last_ item that the undo stack changed
+
+	//Also, performing an undo reverses the stack. So when you do a system undo, the changes are applied in reverse
+	ref = Undo_controls.redo();
+	ref.second == Axis_map_to;
+	ref.first == &Axis_map_to[JOY_AXIS_ACTIONS - 1];      // ::first references the _last_ item that the undo stack changed
+
+***********************************************************/
+
 /*!
  * @brief Base class of Undo_item. Use this when making an undo stack!
  * @sa Undo_item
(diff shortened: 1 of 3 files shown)


MageKing17:
I think a usage example might be helpful.

z64555:
It has application in FRED and in the controls config menu when setting new binds, I'm still working on the controls code overahul, but somebody may be able to implement it in FRED while I'm doing that.

MageKing17:
I meant a code example of how it's meant to be used; I can think of applications for an undo system just fine on my own, thank you very much. :P

z64555:
Ah! Right, of course.

MageKing17:
Why are `undo_stack` and `redo_stack` deques of `Undo_item_base*` pointers instead of deques of `Undo_item_base`s themselves? You might have to change `undo()` and `redo()` to use `std::move()` for shuffling between `undo_stack` and `redo_stack`, if you want to avoid redundant copies being made of things (which the compiler might optimize away anyway), but you'd be able to get rid of all the `new`/`delete` stuff.

z64555:
>Why are undo_stack and redo_stack deques of Undo_item_base* pointers instead of deques of Undo_item_bases themselves?

Because the items that are actually stored are derived from Undo_item_base. If I tried storing them as instances of Undo_item_base, the data will be sliced out. However, if I make an instance of the item elsewhere, and store a reference to them within the stacks, then I can access the item's data.

And yeah, I do plan on using move operations once I figure out how to use them correctly.

MageKing17:
> To further clarify, you cannot push more than one instance of a templated class onto a container. This is because the container is expecting only one type of class, and cannot handle other derived classes or template instances.

Well, that's annoying. Okay then.

OlivierHamel:
z64555 asked for some criticism so here goes:

Summary: I can see a few minor impl issues (style & correctness), and some major design issues.

Impl' issues:
- Operator `new` doesn't return null by default when it fails to allocate enough mem, it throws (see std::nothrow_t); if `new` is overloaded by default in SCP, then it might fail by returning null (depends on how it was written). Either way, it might be better just to ignore the failure case unless you can actually handle it (so an assert is not terribly helpful in this case[\nitpick]).
- You could use C++11 ranged for loops instead of explicit iterators
- There's a lot of manual screwing around to make sure the ptrs are deleted when the stack is popped, etc. Perhaps std::unique_ptr might make this much cleaner? (See design issues.)

Design issues:
-  This is effectively an untyped API. The template member is only there to make it trivial to push datums onto the stack w/ the right size calculations; all type information is lost/discarded because the stacks are heterogeneous.
- Since the stack is heterogenious you basically need to add special tags everywhere just to be able to interpret the data in the stack, which is awkward/alarming (see example provided in the code's comments, it's distressing).
- The code doesn't provide any real amount of functionality above a standard stack w/ all the type-safety removed to allow for heterogenous collections. You have to implement the actual undo-redo logic yourself separately from the stack management.

If you really need heterogenous undo-redo stacks, what you really want is known as the 'command' pattern. TL;DR: You define two operations/functions; one for applying a mutation ('redo') and the other for undoing it. Appling the 'initial' mutation is equiv to doing a 'redo'. In this sense, what you really want is the following:
```using Action = std::function<std::function<Action>()>;
using Stack   = std::stack<Action>;```

Where an `Action` applies some sort of mutation and returns a closure that will reverse whatever it did when invoked (and in turn return a closure that does the same). This models both 'undo' and 'redo'. It hides the details/data allocations required (these would either be lambda-captured, or std::bind-d). It doesn't require finiky management/interpretation of entries on the stack. It succinctly/cleanly handles adding new actions (and their redo/undo-ing) without having to worry about updating all the points where a redo/undo can be applies b/c the application is uniform for all actions. (Just invoke the closure, and stuff the returned `Action` on the redo/undo stack as appropriate.)

**Comment by z64555** [comment]
We talked over IRC about his last bullet point, he had skimmed over the Undo_system class which implements the actual undo/redo operations.

Impl. issues are all good points, and I'll address them soonish.

The design issues are also good points, but unfortunately there's not much I can do with C++ (even C++11) to prevent the data type information from being lost.

When calling undo() or redo(), the client software typically wants to know what changed so that it can tell the user. Depending on how the client software is set up, this can be as simple as comparing the container void* (returned by undo() and redo()) with its known containers, or require access to the changed item. Returning both a pointer to the changed location and a pointer to its container allows the client to do either, albeit in a "weird" way. 

Instead of returning a reference to a parent container, I had considered using an enum. The client code would then pass the enum to save() instead of the parent container address, and use it to determine the data type when doing an undo or redo. This would, however, limit the Undo_system to whatever the client had defined, and would require the client to maintain the enum each time they wanted to add a new type. This would also lose the ability to find the item within the parent container when there are multiple containers with the same type.

I had also considered using a string to identify the type, instead of the enum, but the implementation of that would be more complex and still have the pitfalls as stated above.

File: code/globalincs/undosys.h [hunk 1]
```diff
@@ -0,0 +1,264 @@
...
+		if (data != nullptr) {
+			delete data;
+		}
+	};
+
+	/**
+	 * @brief Restores the saved data
+	 *
+	 * @returns A pair of const void* referencing the affected item and its container (if it was provided)
+	 * @details If successful, the data at dest has been swapped. So if this was an undo item, it's now a redo item
```
asarium: When would the operation ever not be successful? The code always swaps the pointers and returns the pair so it cannot fail.

code/globalincs/undosys.h
```diff
@@ -0,0 +1,264 @@
...
+	 */
+	std::pair<const void*, const void*> restore() {
+		Assert((dest != nullptr) && (data != nullptr));
+
+		T copy = *static_cast<T*>(dest);
```
asarium: You may be able to use `std::swap` here. I'm not sure if that's compatible with the pointers though.

code/globalincs/undosys.h
```diff
@@ -0,0 +1,264 @@
...
+	};
+
+	/*!
+	 * @brief Saves data and its destination as an undo item
+	 *
+	 * @param[in] _data The data to save
+	 * @param[in] _cont The container this data is originally found in
+	 * @details The explicit definition is here to ensure type safety
+	 */
+	void save(T& _data, T* _cont = nullptr) {
```
asarium: `_data` is a reference but you use it like a pointer (e.g. you test for `nullptr`). Is this intentional and if yes then the class should ensure that `T` is a pointer type by using a `static_assert`.

code/globalincs/undosys.h
```diff
@@ -0,0 +1,264 @@
...
+ * @brief Generic Undo/Redo system. Save whatever, restore whatever!
+ *
+ * @details Currently uses a pair of deques. Would ideally use a ring container so that the actual Undo_item instances don't go anywhere.
+ */
+class Undo_system
+{
+public:
+	Undo_system();
+
+	Undo_system(uint _undos);
```
asarium: This parameter should be a `size_t` since it probably controls how big the buffers should be?

code/globalincs/undosys.h
```diff
@@ -0,0 +1,264 @@
...
+
+private:
+	uint max_undos;	//!< Max number of Undo's available
+
+	/*! 
+	 * @brief The undo and redo stacks. These are containers of pointers, because we can't simply store the base class.
+	 *   If we did, then the data of the Undo_item instances would be sliced out.
+	 * @{
+	 */
+	SCP_deque<Undo_item_base*> undo_stack;
```
asarium: You should use a `unique_ptr` if you manage this memory yourself which is, as far as I can tell, the case since you use `delete` on the items in the lists.

code/globalincs/undosys.h
z64555: Should the data or destination references ever get reset to nullptr it will fail, but the comment does suggest the function is able to resume after failure, which it cannot.

code/globalincs/undosys.h
z64555: std:swap should work, as it dereferences the pointers to be able to swap the data between them. using it in next commit

code/globalincs/undosys.h
z64555: The Assert test is something of a carry over from the first draft of this function where _data was a *.  Since its a &, testing against nullptr is moot, and shall be removed

code/globalincs/undosys.h
z64555: Originally intended to use this as a sort of hard limit to how many undo's the system could handle, although that function is handled by a user-configurable soft limit already.  Changed to size_t in next comit.

z64555:
Ok, I tried using unique_ptr's in the Undo_system's undo and redo stacks, but got hung up on how to transfer them between the stacks whenever an undo item is restored and converted into a redo item, and vice versa.

z64555:
Closing due to integration with #2907

PR #1251, opened 2017-02-21
Load the Fire particle animations

### LINKED ISSUE #1245: Impact effects gone
I happened to fire up the Lucifer Dogfight - Ultra Heavy multiplayer mission I found, just to give the standalone a quick test.  While shooting the Lucifer with a Circe/Kayser, combo, I realized something felt absent about what I was seeing.  The bolts seemed to kind of disappear into the hull, the impact sound would play, and then the damage spew would be visible.  It happens quickly, so I wasn't sure I wasn't imagining things.  But I fired up 3.7.4 just for a sanity check anyway, same mission, same layout.  I shot the Lucifer, and I could distinctly see the yellow impact spew, just before the damage spew gets going.  It seems that somehow those are broken in current builds.  If I need to bisect this to find the exact commit I can try to get around to that, but I imagine it's one of the large graphics related commits.  Doubt it has anything to do with the specific target ship, weapon loadout, or multiplayer, but I can try to eliminate all those variables too if no one has any hunches on this.

The-E:
The new particle system didn't load the particleexp01 anim as a generic fire animation. This caused calls to create PARTICLE_FIRE particles to fail silently.

Fixes #1245.

### DIFF
--- a/code/particle/particle.cpp
+++ b/code/particle/particle.cpp
@@ -87,6 +87,11 @@ namespace particle
 	void init()
 	{
 
+		if (Anim_bitmap_id_fire == -1) 
+		{
+			Anim_bitmap_id_fire = bm_load_animation("particleexp01", &Anim_num_frames_fire, nullptr, NULL, 0);
+		}
+
 		// Cough, cough
 		if (Anim_bitmap_id_smoke == -1)
 		{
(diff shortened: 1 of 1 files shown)


code/particle/particle.cpp
```diff
@@ -87,6 +87,11 @@ namespace particle
 	void init()
 	{
 
+		if (Anim_bitmap_id_fire == -1)
```
chief1983: Got some hanging whitespace here

The-E:
Not anymore!

PR #1254, opened 2017-02-22
Use a manually maintained prefix header for cotire

asarium:
Cotire (the CMake module we use for handling precompiled headers) can
auotmaically generate a prefix header by examining the includes of the
project. However, if a system include changes, the prefix header may
change as well which triggers a complete project rebuild. To avoid this
and to add the ability to customize the prefix header, I added a
manually maintaned header which will be used by all platforms and which
will only change if that file is manually changed.


asarium:
I removed the include of `pstypes.h` again since that conflicted with some definitions in the Windows headers. It increases the Unix build times a bit but I think this still gives ~ 30% decrease in compile time on those platforms. Windows is unaffected since cotire was already enabled on that platform.

PR #1260, opened 2017-02-24
Make Skip_packfile_search restrict search to actual files

### LINKED ISSUE #1224: Available missions list in campaign editor is empty
Since the SDL2 merge, the available missions list in the campaign editor has always been empty, even if mission files are present in the missions folder. This makes it impossible to add missions to campaigns with recent versions of FRED!

asarium:
The flag was previously used to restrict searching of mission files to
the root directory of the mod. Since the SDL merge changed what the
default root directory is (now it's the mod folder in the user folder)
this broke the mission editor and it didn't show any missions anymore.

This fixes that by altering how Skip_packfile_search affects the file
search. Now that causes the search to skip files in VPs but it still
includes all other files that are present on the disk.

I also fixed some inconsistent behavior in the various file enumeration
functions since they weren't doing the same thing.

This changes the original behavior slightly since new files aren't found
in this search since the file list is only initialized when CFile is
started so if a new file is added after that, it is not reflected in the
mission list. Since that is pretty rare we can probably ignore that
here.

This fixes #1224.


asarium:
I added a unit test for testing the correct behavior of the flag and also if the flag isn't set.

PR #1263, opened 2017-02-25
Add generic configuration system

asarium:
I would like to add an in-engine options screen at some point but to do
that we first need a system that is actually able to handle loading and
saving of various configuration values in a generic way.

These changes add a config system on top of the existing system that
provide an easy to use API for defining configuration options that are
automatically read and written to the config file at the appropriate
moments.

Since this is clearly a feature addition, this should not be merged before 3.8. Also, since this is supposed to be a generic system that will be used more in the future I would like to know if anyone has an idea on how to make this better.

### DIFF
--- a/code/config/ConfigurationItem.cpp
+++ b/code/config/ConfigurationItem.cpp
@@ -1,0 +1,24 @@
+//
+//
+
+#include "ConfigurationItem.h"
+
+#include "config/ConfigurationManager.h"
+
+namespace {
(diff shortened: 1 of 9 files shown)


asarium:
I wanted to close this for a while now but always forgot it...

The reason I wrote this in the first place was that I wanted an easier way of storing and retrieving options in the config file for use in a possible in-game options menu. I have written such a system for the new high resolution UI project so this is not relevant any more.

PR #1265, opened 2017-02-25
Make table-based texture replacement work

### LINKED ISSUE #1264: Table texture replacement broken in the lab
Tabled texture replacement is currently not being reflected in the ship lab or the techroom.
The attached mod shows this; it replaces the Ulysses' base texture with the terran cockpit one.

[texreplacetest.zip](https://github.com/scp-fs2open/fs2open.github.com/files/801429/texreplacetest.zip)

The-E:
This applies table-based texture replacement settings in the lab, techroom and the weapon/ship select screens.

Fixes #1264

### DIFF
--- a/code/model/modelrender.h
+++ b/code/model/modelrender.h
@@ -75,7 +75,7 @@ class model_render_params
 	int Insignia_bitmap;
 
 	int *Replacement_textures;
-	bool manage_replacement_textures; // This is set when we are rendering a model without an associated ship object;
+	bool Manage_replacement_textures; // This is set when we are rendering a model without an associated ship object;
 									  // in that case, model_render_params is responsible for allocating and destroying
 									  // the Replacement_textures array (this is handled elsewhere otherwise)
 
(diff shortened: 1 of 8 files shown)


code/model/modelrender.h
```diff
@@ -74,6 +75,9 @@ class model_render_params
 	int Insignia_bitmap;
 
 	int *Replacement_textures;
+	bool manage_replacement_textures; // This is set when we are rendering a model without an associated ship object;
```
asarium: Since this class now manages memory on its own it will likely cause either a memory leak or a double free if it is copied so we either need to delete the copy operators or properly implement them.

The-E: As far as I can tell (and the compiler backs me up) there's only one place in the codebase where a model_render_params object was actually copied. I've changed that and disabled the copy operator.

code/model/modelrender.h
```diff
@@ -97,6 +97,8 @@ class model_render_params
 
 	bool Normal_extrude;
 	float Normal_extrude_width;
+
+	model_render_params(const model_render_params& other);
```
asarium: You should use `= delete` and also delete the copy operator to fully disable copying.

PR #1269, opened 2017-02-27
Add a Windows executable manifest to FreeSpace2

asarium:
This sets the dpiAware flag which should improve the compatibility with
High-DPI displays. It also sets a few values for specfying what Windows
versions we support and it also enabled the nicer dialog boxes which
makes the error and warning dialogs look a bit nicer.

### DIFF
--- a/freespace2/resources/win/default.manifest
+++ b/freespace2/resources/win/default.manifest
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3" >>
+    <!-- Enable themes for Windows common controls and dialogs (Windows XP and later) -->
+    <dependency>
+        <dependentAssembly>
+            <assemblyIdentity
+                type="win32"
+                name="Microsoft.Windows.Common-Controls"
+                version="6.0.0.0"
+                processorArchitecture="*"
+                publicKeyToken="6595b64144ccf1df"
+                language="*"
+            />
+        </dependentAssembly>
+    </dependency>
+
+    <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+        <application>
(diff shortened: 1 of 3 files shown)


asarium:
These changes have the possibility of breaking on older Windows versions so we might get some reports about that once a nightly build with this commit is released. I tested it on Windows 10 but that doesn't mean that is works on earlier Windows versions.

freespace2/resources/win/default.manifest
```diff
@@ -0,0 +1,47 @@
...
+            <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
+            <!-- Windows 8.1 -->
+            <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}"/>
+            <!-- Windows 8 -->
+            <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}"/>
+            <!-- Windows 7 -->
+            <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}"/>
+            <!-- Windows Vista -->
+            <supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}"/> 
+        </application>
```
MageKing17: Are we officially dropping Windows XP support, then?

asarium: As far as I know, no. That section of the manifest is not relevant for Windows XP (I looked, there is no GUID for that version). I haven't tested if the builds still work on XP but I haven't found anything that would suggest that this breaks XP compatibility.

MageKing17: I see; in that case, a comment that XP is implied *might* be in order so nobody else gets confused.

asarium: Good point, I'll add that comment.

asarium: Done.

chief1983:
This seems to have broken interface and video scaling on Mac.  In windowed mode it takes a small portion of the screen, in full screen it is stretched to the point that most content is off of the screen.

PR #1270, opened 2017-02-28
Require that the energy bar animation has 4 frames

asarium:
Since the rendering code always accesses all four frames it must be
ensured that they are all present. Currently, there is a bmpman
assertion about an invalid handle if the animation does not have enough
frames. This fixes that by warning about that and then resetting the
bitmap handle to -1 so that the rendering code doesn't try to use the
handle.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -5,7 +5,7 @@
  * or otherwise commercially exploit the source or things you created based on the 
  * source.
  *
-*/ 
+*/
 
 
 
@@ -77,7 +77,7 @@ static int ballistic_hud_index = 0;	// Goober5000
 extern object obj_used_list;		// dummy node in linked list of active objects
 extern char *Cargo_names[];
 
-// shader is used to shade the target box 
+// shader is used to shade the target box
 shader Training_msg_glass;
 
 // the target triangle (that orbits the reticle) dimensions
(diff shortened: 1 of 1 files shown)


code/hud/hudtarget.cpp
```diff
@@ -5533,6 +5533,16 @@ void HudGaugeWeaponEnergy::initBitmaps(char *fname)
 	Energy_bar.first_frame = bm_load_animation(fname, &Energy_bar.num_frames);
 	if ( Energy_bar.first_frame < 0 ) {
 		Warning(LOCATION,"Cannot load hud ani: %s\n", fname);
+	} else {
```
niffiwan: could you confirm that this is the only non-whitespace change in the PR?  This chunk looks fine to me, but I just want to be sure I didn't miss anything.

chief1983: According to git diff -w, that else block is indeed the only non-whitespace change.
For anyone's reference who may be curious, the easiest way to get that info is:
```
# Fetch the pr to a local reference, branch pr1270
git fetch origin pull/1270/HEAD:pr1270
 # Diff (or --stat diff) of the branch to its common ancestor with master, ignoring whitespace (-w)
git diff -w [--stat] `git merge-base origin/master pr1270` pr1270
```

niffiwan: thanks! Although in this case just diff'ing vs the previous commit should be fine?
```
git diff -w HEAD^
```

chief1983: Yeah, that is often the case, but my method will always work, for any PR, even with some pretty messy merges (as long as it doesn't think there are commits in the PR that aren't actually part of the PR).  I'm pretty sure I posted the command line for a GUI would generate a diff like you get on Github, BitBucket, etc.  Or it's the closest I've found to it so far.

PR #1271, opened 2017-03-01
Adjustable fresnel term in lighting model. Model shader cleanup. Deferred lighting shader fix.

SamuelCho:
This pull request should address a lot of the concerns expressed in this thread (http://www.hard-light.net/forums/index.php?topic=93151.0) about the physically-based specular lighting model. Namely, this changes it so that we don't use Fresnel reflectivity when using legacy specular data. This preserves the darker specular colors, preventing legacy specular data from looking really flat at grazing angles when Fresnel reflections become really prominent. 

I also tidied up the model shader a bit, removing a bunch of unused or unnecessary variables.

I also discovered a bug in the deferred lighting shader, where it was using specular light source color as the specular material instead of using the specular material from the g-buffer.

### DIFF
--- a/code/def_files/deferred-f.sdr
+++ b/code/def_files/deferred-f.sdr
@@ -29,9 +29,9 @@ vec3 FresnelSchlick(vec3 specColor, vec3 light, vec3 halfVec)
 {
 	return specColor + (vec3(1.0) - specColor) * pow(1.0 - clamp(dot(light, halfVec), 0.0, 1.0), 5.0);
 }
-vec3 SpecularBlinnPhong(vec3 specColor, vec3 light, vec3 normal, vec3 halfVec, float specPower, float dotNL)
+vec3 SpecularBlinnPhong(vec3 specColor, vec3 light, vec3 normal, vec3 halfVec, float specPower, float fresnel, float dotNL)
 {
-	return FresnelSchlick(max(specColor, vec3(0.04)), light, halfVec) * ((specPower + 2.0) / 8.0 ) * pow(clamp(dot(normal, halfVec), 0.0, 1.0), specPower) * dotNL;
+	return mix(specColor, FresnelSchlick(specColor, light, halfVec), fresnel) * ((specPower + 2.0) / 8.0 ) * pow(clamp(dot(normal, halfVec), 0.0, 1.0), specPower) * dotNL;
 }
 vec4 SpecularLegacy(vec4 specColor, vec3 normal, vec3 halfVec, float specPower)
 {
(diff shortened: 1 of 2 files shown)


asarium:
I tested this with the BP2 Act 3 Intro sequence and IMO it looks better than before but I noticed that the skybox reflections are much stronger than before. I'm not sure if that's intentional but IMO it's a bit too strong now.

chief1983:
@asarium odd, that description reminds me of this ancient bug http://scp.indiegames.us/mantis/view.php?id=1842

Could there have been something similar now this whole time that was just washed out by the glare effect before?  Or is the skybox reflection now visible from all angles?

SamuelCho:
I accidentally discarded specular map alpha if we're not using a reflectance map. That's supposed to attenuate env reflections. My bad.

SamuelCho:
Environment map reflections should be back to normal now.

chief1983:
Was this fix only for when deferred lighting is enabled?  I'm not seeing any discernable improvement on my Mac with the retail asteroids, but deferred lighting is still not working correctly there.

SamuelCho:
The retail asteroids exhibited the same lighting behavior in 3.7.4. I can't do anything about that.

chief1983:
They do have glossy spots but the effect is much more extreme now.  The side of the Orion in the ship lab still behaves like a giant mirror as well.  In 3.7.4, the light source follows the mouse, and you can see the source playing along the side of the hull.  Now, half the side of the ship turns white.  I imagine this is the same as what's happening with the asteroids, whatever the reflection of the light source effect is, it just seems to be at a much higher, broader level than before.  The effect is definitely most noticeable with post processing enabled in the ship lab.  Turning it off and I almost can't tell a difference between 3.7.4, the latest nightlies, and the build I made today.  A build with this change is definitely improved, looking at the Orion in the ship lab, with post processing on.  But the washout effect is still far greater than 3.7.4.  I might try to piece together a video showing different behaviors if that would help (getting good with the QuickTime recorder).

SamuelCho:
Then I guess your problem is with the gamma correction, HDR, and bloom. Either you have it or not. I can't do anything about it at this point.

chief1983:
Just so I understand it then, are these all things that take place in the post processing stage?

SamuelCho:
If you disable post processing, you disable HDR.

Goober5000:
@asarium any comment on chief1983's comment of March 1?

asarium:
The bug I described and that @chief1983 referenced was fixed shortly after so it's unlikely that the bug @chief1983 referenced is related to these changes.

PR #1272, opened 2017-03-02
Added support for renaming controls in controlconfigdefaults.tbl

ln-zookeeper:
Controls can already be renamed via strings.tbl as well, but it is more complicated and for example scripting hooks would still have to refer to the controls by their hardcoded names. This feature allows for controls to be renamed in a more comprehensive manner in TC's.

This seems to be compatible with strings.tbl; as long as the entry has $Has XStr: 1, the string can still be changed in strings.tbl as usual.

It's probably wanted that the strdup()ed strings get freed when quitting, although at a glance I didn't find an obvious place to do it in. Also, this surely conflicts with z64555's rewrites, but a PR was suggested regardless.

### DIFF
--- a/code/controlconfig/controlsconfig.cpp
+++ b/code/controlconfig/controlsconfig.cpp
@@ -999,9 +999,9 @@ void control_config_reset_defaults(int presetnum)
 
 	// Reset keyboard defaults
 	for (i=0; i<CCFG_MAX; i++) {
-		// Note that key_default and joy_default are NOT overwritten here;
-		// they should retain the values of the first preset because
-		// for example the key-pressed SEXP works off the defaults of the first preset
+		// Note that key_default, joy_default and default_text are NOT
+		// overwritten here; they should retain the values of the first preset
+		Control_config[i].text = preset[i].text;
 		Control_config[i].key_id = preset[i].key_default;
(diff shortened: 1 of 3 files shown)


z64555:
Using an SCP_string instead of c-strings would take care of allocation/deallocation without much hassle, but I'm not sure if they'll work in the old array of structures format. The controls overhaul includes changes to the Controls_config list which would allow this, If you need this feature soonish, I may be able to split off those changes for you.

chief1983:
@z64555 we really do need this to move forward so we can clear out the remaining FotG code changes and get back to using nightly builds.  How is the controls overhaul doing?

wookieejedi:
@z64555 any possible update on this feature? : )

z64555:
I'm still pretty backlogged doing physical labor all over the place, when I get home the only thing I want to do is rest. 😫

No current estimates can be made at this time, but you can take a look at my repo if you feel up to the task.

Goober5000:
@z64555 How about now? :)

Goober5000:
This was originally requested by FotG but they no longer need it.  Given this, the complexity of the PR, and the fact that @z64555 hasn't really been available recently, I'm closing this.

Issue #1278, opened 2017-03-04
ADE scripting system is not compatible with C++ classes


asarium:
The scripting system uses `memcpy` for moving the C++ object data into the Lua memory but that doesn't respect any of the restrictions that class may have. Basically, all Lua API types must be trivially copyable but that is currently not the case which can cause bugs and crashes.

The right solution would be to completely refactor how we handle the objects in the scripting API by calling the constructors and destructors instead of using `memcpy` and `__gc`.

asarium:
Oh, I forgot that this exists. This has been fixed for some time now.

PR #1281, opened 2017-03-04
Disallow to use non-trivial types as scripting handles

asarium:
As described in #1278, the current scripting system has issues with
handling non-trivial C++ types. Since fixing that deficiency is too much
work, especially since we are in a feature freeze, I decided to simply
enforce the usage of trivial types for scripting handles. Fortunately,
only two handles used non-trivial types so it was relatively easy to fix
those instances by dynamically allocating those and only storing the
pointers in the lua memory.

This fixes a bug in JAD-XA where the game would crash at the last boss.

Also, I noticed that the HudGauge handle used a copy of the hud gauge
rather than a pointer to the actual instance so I'm pretty sure that any
scripting changes to a custom HUD gauge simply didn't work before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,7 @@ matrix:
                       - llvm-toolchain-precise
                   packages:
                       - clang-3.6
+                      - libc++-dev
                       - *global_deps
         - os: osx
           env: CONFIGURATION="Debug"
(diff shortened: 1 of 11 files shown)


asarium:
I needed to enable libc++ for clang since these changes require some advanced C++11 features which were not available for Clang on Travis CI.

PR #1284, opened 2017-03-05
Light attenuation fixes.

SamuelCho:
After fixing Fresnel reflectivity in the deferred lighting shader, deferred lighting appears a bit less intense (http://www.hard-light.net/forums/index.php?topic=93291.0). That led me to identifying that our light attenuation is incorrect. Attenuation behavior between the forward and deferred lighting shaders should match following this pull request.

I've also fixed spec_tube and spec_point command line factors from affecting diffuse deferred lights while taking the opporunity to remove the defunct specular intensity code from the forward lighting shader.


SamuelCho:
Don't merge this in yet. Linear attenuation is creating some problems of its own.

PR #1288, opened 2017-03-06
Lighting attenuation fixes and deferred tube light stencil test removal

SamuelCho:
Addresses http://www.hard-light.net/forums/index.php?topic=93291.0

I decided to keep the current method of light attenuation in the deferred lighting shader. I also added the correct linear attenuation equation to the forward lighting shader. A lot of the tube light issues seem to be coming from how rendering cylinders with stencil buffer enabled. The stencil buffer isn't letting other tube lights be drawn which makes it seem like tube lights aren't working at all. I decided to disable stencil tests for tube lights as the overlap between the sphere endpoints and the cylinder that comprise the tube light seems negligible (Unless someone complains about it). There should no longer be missing tube lights from busy beam fights now.


asarium:
I'm not sure if I understand how the stencil buffer correctly but couldn't you just use different values for each tube light instead of using the same stencil value everywhere?

As I understand it, the stencil buffer was used to avoid lighting a fragment twice if the cylinder and sphere part of the tube light overlap so we could achieve the same effect by using a different stencil value for each tube light and then set the stencil function accordingly, right?

SamuelCho:
Yes you're correct. When Valathil wrote the original deferred lighting code, it originally wrote into the color g-buffer directly and didn't use additive blending. It used one minus alpha blending which would end up replacing color pixel data entirely. That's why I think he implemented the stencil buffer mask.

But honestly, I feel that the stencil buffer solution of compositing multiple lights into a tube light doesn't work very well now that deferred lighting uses a more correct additive blending approach. The cylinder ends up masking  out the stencil buffer before the spheres have a chance to render anyway. The result looks a lot better if we just let the spheres render out so the lights can at least composite on to each other. Maybe we can find a much more accurate capsuled cylinder area light approach later on but I think the stencil buffer approach won't give us a good result even if it worked as intended.

asarium:
Yes, that makes sense. I did some quick tests and I couldn't see any issues so that change didn't break anything. The code still enables the stencil test, is that still needed? As far as I can tell, the stencil test is never used anyway since it's always set to 0.

SamuelCho:
I intended to do that but I didn't want to do a half-assed job removing the stencil attachment and stencil framebuffer setup code. So I decided to just commit what I had before I went to bed.

asarium:
I meant the code right above, starting at line 2137. IMO, we should keep the stencil attachment in case we need it for something else down the line.

SamuelCho:
The stencil buffer setup code is removed.

PR #1296, opened 2017-03-07
Restore the flag viewers

The-E:
During the bitset flags conversion, the flag viewers in the F3 lab were disabled. This restores their functionality.

### DIFF
--- a/code/lab/lab.cpp
+++ b/code/lab/lab.cpp
@@ -1363,20 +1363,14 @@ void labviewer_set_class_window(int mode)
 
 
 // ------------------------------  Flags Window --------------------------------
-typedef struct lab_flag {
-	Checkbox *cb;
-	int flag;
-	bool second;
-} lab_flag;
-
 template<class T>
-struct lab_flag_new {
+struct lab_flag {
 	Checkbox* cb;
 	T flag;
 };
 
-static SCP_vector<lab_flag_new<Ship::Info_Flags>> Ship_Class_Flags;
-static SCP_vector<lab_flag_new<Weapon::Info_Flags>> Weapon_Class_Flags;
+static SCP_vector<lab_flag<Ship::Info_Flags>> Ship_Class_Flags;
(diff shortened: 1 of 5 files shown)


code/lab/lab.cpp
```diff
@@ -1424,13 +1425,21 @@ void labviewer_populate_flags_window()
 	// clear out anything that already exists
 	labviewer_flags_clear();
     
+	int y = 0;
+
     // ship flags ...
     if (Lab_mode == LAB_MODE_SHIP) {
-        //Needs reimplementation
+		for (int i = 0; i < Num_ship_flags; ++i)
```
MageKing17: Recommend changing `int i = 0` to `auto i = 0` here since we're getting a signed/unsigned mismatch warning.

code/lab/lab.cpp
```diff
@@ -1424,13 +1425,21 @@ void labviewer_populate_flags_window()
...
-        //Needs reimplementation
+		for (int i = 0; i < Num_ship_flags; ++i) 
+		{
+			labviewer_flags_add<Ship::Info_Flags>(nullptr, &y, Ship_flags[i].name, Ship_flags[i].def, Ship_Class_Flags);
+		}
     }
     // weapon flags ...
     else if (Lab_mode == LAB_MODE_WEAPON) {
-        //Needs reimplementation
+		for (int i = 0; i < num_weapon_info_flags; ++i)
```
MageKing17: And here as well.

code/lab/wmcgui.h
```diff
@@ -667,6 +672,22 @@ class Checkbox : public GUIObject
 		SetFlag((int*)in_flag_ptr, in_flag);
 	}
 
+	template <class T>
+	void SetFlag(flagset<T>& in_flag_set, T flag_value, ship_info* si) {
```
MageKing17: VS 2013 doesn't like these overrides for some reason.

PR #1297, opened 2017-03-08
Remove a couple of annoying warnings in ship flag parsing

The-E:
These were pointed out by Spoon on the forums; given that these are just spewing warning dialogues but not applying any corrective measures (which leads me to believe that these Warnings are strictly informative, nothing more), downgrading them to log print should be fine.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -6647,17 +6647,17 @@ int subsys_set(int objnum, int ignore_subsys_info)
 
 		if ((ship_system->system_info->flags[Model::Subsystem_Flags::Turret_salvo]) && (number_of_weapons > 1))
 		{
-			Warning (LOCATION, "\"salvo mode\" flag used with turret which has more than one weapon defined for it\nsubsystem '%s' on ship type '%s'.\nonly single weapon will be used\n", model_system->subobj_name, sinfo->name );
+			mprintf(("\"salvo mode\" flag used with turret which has more than one weapon defined for it\nsubsystem '%s' on ship type '%s'.\nonly single weapon will be used\n", model_system->subobj_name, sinfo->name));
 		}
 
 		if ((ship_system->system_info->flags[Model::Subsystem_Flags::Turret_fixed_fp]) && (number_of_weapons > model_system->turret_num_firing_points))
 		{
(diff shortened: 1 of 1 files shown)


chief1983:
Are we sure these weren't doing more harm than good?  They have been in for almost 8 years, so approaching a decade.  Is a few hours to have discussion on their removal really sufficient?  Is there a more likely chance modders will set their turrets up incorrectly and not notice it now?  Is it possible that the behavior Spoon claims is correct could be achieved in a manner that doesn't result in the warnings?  Considering that Karajorma wrote the warnings and is still an active member I would have really loved to see him weigh in on this before it was merged.  I mean he was even chatting on Discord last night.

The-E:
It boils down to what we want Warnings to mean. To me, a Warning should be issued when the engine encounters data that is a) incorrect and b) correctable at runtime (Similar to the other warnings set in that same code block, or the numerous sanitization tweaks we do when parsing tables).

In this case, these warnings never did any of that. If you read the text of the warnings closely, it becomes clear that they're all something of a "heads up, these flag combinations may result in unwanted behaviour" info blurbs, not the sort of "this is potentially fatal, do not do this" type of stuff we usually use Warning for.

I could be wrong about that, of course -- wouldn't be the first time --, but I really do not think I am. It's pretty clear to me that Spoon's complaint, that these Warnings are useless, is correct: All these things do is drive up the Warnings counter for no good reason.

chief1983:
I can see the lack of any corrective action maybe indicating that these weren't as necessary as we thought.  But I know there's been lots of mods using these flags before, and as stated, the warnings have been around a very long time.  I just wonder if degrading it to a log error is going to lead to people not noticing their turrets aren't working as they expected them to (probably the reason the warning was added in the first place, and the same reason the 3d audio warning was bumped up to a warning, since no one noticed the log messages).  I know that the 3d audio warning is actually something that results in corrective action, but the point was that no one was noticing the log messages.  I have a feeling that this being a warning has helped many modders over the years.

The-E:
In the end though, the things being warned about aren't cases where the resulting behaviour is wrong in any way. It may be _unintended_, but it's not _wrong_ or faulty in any way, and as such, not something we should be using Warnings for.

PR #1298, opened 2017-03-08
Add warning category and suppression system

asarium:
This allows to assign a category string to a specific warning which can
then be used for suppressing that specific warning. This system uses a
new function call so existing warnings need to be upgraded to category
warnings as required. To start, I changed the channel downsampling
warning to such a category warning.

I also added a unit test to check if the system works as intended.


asarium:
This is no longer needed since #1524 has been merged.

PR #1308, opened 2017-03-12
Add colour-configurable tint shader

niffiwan:
FotG needs a tint shader for a feature; basically toggle screen tinting
when a key is pressed as a visual indicator of which mode your ship is
in. Was previously done in a custom shader.

### DIFF
--- a/code/scripting/api/objs/graphics.cpp
+++ b/code/scripting/api/objs/graphics.cpp
@@ -164,7 +164,7 @@ ADE_FUNC(__len, l_Graphics_Posteffects, NULL, "Gets the number or available post
 	return ade_set_args(L, "i", ((int) names.size()) + 1);
 }
 
-ADE_FUNC(setPostEffect, l_Graphics, "string name, [number value=0] [red 0.0 - 1.0] [green 0.0 - 1.0] [blue 0.0 - 1.0]", "Sets the intensity of the specified post processing effect. Optionally set RGB values for post effects that use them", "boolean", "true when successful, false otherwise")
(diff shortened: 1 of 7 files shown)


asarium:
Looks good to me, @SamuelCho any objections?

SamuelCho:
This is fine.

The-E:
I think it would be a good idea to make this effect sexp-controllable.

niffiwan:
Ok, I'll add some optional params to set-post-effect

niffiwan:
ok, I can fix that easier enough. The clang warning-converted-to-error on the other hand...

code/scripting/api/objs/graphics.cpp
```diff
@@ -164,18 +164,23 @@ ADE_FUNC(__len, l_Graphics_Posteffects, NULL, "Gets the number or available post
 	return ade_set_args(L, "i", ((int) names.size()) + 1);
 }
 
-ADE_FUNC(setPostEffect, l_Graphics, "string name, [number value=0]", "Sets the intensity of the specified post processing effect", "boolean", "true when successful, false otherwise")
+ADE_FUNC(setPostEffect, l_Graphics, "string name, [number value=0] [red 0.0 - 1.0] [green 0.0 - 1.0] [blue 0.0 - 1.0]", "Sets the intensity of the specified post processing effect. Optionally set RGB values for post effects that use them", "boolean", "true when successful, false otherwise")
```
asarium: The parameter documentation isn't consistent with the rest of the code. Parameters should be separated by commas and for optional parameters the default value should be specified. The acceptable range should be specified in the function documentation and not in the parameter list.

asarium:
The Lua documentation looks better now, thanks for changing that.

PR #1313, opened 2017-03-13
Fix OpenGL distortion map update

### LINKED ISSUE #1307: Framebuffer Thrusters not Rendering Correctly
Not sure how I missed this, but FB Thrusters no longer look like proper heat distortions.
Instead they look like they slightly bend the thruster and the appearance is only noticeable at a distance.
Near:

Far:

Pic from JAD:

asarium:
The updated values for the distortion map were rendered too close to the
edge of the screen so they were clipped which meant that there was no
distortion data available. This fixes that by moving the rendered
primitives farther away from the edge. I'm not sure why this worked in
3.7.4, maybe the fixed function rendering used some slightly different
matrix.

I also don't know why this wouldn't work consistently across GPU vendors
but maybe the rasterization rules vary a bit across drivers.

This fixes #1307.

### DIFF
--- a/code/graphics/opengl/gropengldraw.cpp
+++ b/code/graphics/opengl/gropengldraw.cpp
@@ -2338,7 +2338,7 @@ void gr_opengl_update_distortion()
 		distortion_verts[i].b = 255;
 		distortion_verts[i].a = 255;
 
-		distortion_verts[i].screen.xyw.x = 0.04f;
+		distortion_verts[i].screen.xyw.x = 1.f;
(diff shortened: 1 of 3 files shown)


TRBlount:
This fixes -fb_thrusters on my system. Is there a way to make the effect more apparent after gamma correction? With post processing on the effect isn't very noticeable past the start of the thruster.

asarium:
@TRBlount I don't know if there is a way but it's definitely not part of this PR.

TRBlount:
Of course, I was just asking if it could be a proposed change for 3.8.2

Issue #969, opened 2016-09-12
OS X graphical issues


chief1983:
[Youtube demo](https://youtu.be/XVFI3ULIibg)

[fs2_open.log](https://github.com/scp-fs2open/fs2open.github.com/files/467944/fs2_open.txt)

There are several things that are really wrong looking on my current setup.  From the log, I should have the OpenGL version high enough, and we are now on the core profile, so I'd hoped things would be better.  Still having some issues though.  @SamuelCho, the ability to run OS X builds now should be greatly improved.

I attempted to run the game with -fb_thrusters as well, but wxLauncher doesn't seem to be acknowledging my changing the settings.  Won't acknowledge a resolution change either.

Also, I ran the video recording with a release build, the log is from a follow up run with a fast debug build.

SamuelCho:
Glad to finally be able to debug graphics on Mac. Is this in master yet or is it on one of your branches?

chief1983:
It was merged over the weekend.  Just verified I have the problem with the thrusters with fb thrusters on as well, manually editing my cmdline config file until wxlauncher 0.12.0 is available.

SamuelCho:
Got it, will try to figure out what's up.

SamuelCho:
So initial testing tells me that this has something to do with the g-buffers. Deferred lighting doesn't work on Mac but disabling it should look okay. I'd say just disable deferred lighting as a temporary workaround for now until I can find a solution. Thanks again for unblocking me chief.

chief1983:
Ok yeah, deferred is still an issue.  However, even with it off, the lighting reflections still seem off.  Might just be a shortcoming of their use with the retail models, but maybe not.  The beams on massive battle two will glow entire sides of the Orion and Hecate.  The subachs cause odd glows on the Shivan capital ships as well.  Whole panels blink purple, etc.  Trying to describe in enough detail that I don't need to bother with uploading another video, can't figure out taking a screenshot of the game in OS X at the moment.

PR #1314, opened 2017-03-17
Reorder fragment shader destination binds to fix Mac deferred lighting.

SamuelCho:
Fixes https://github.com/scp-fs2open/fs2open.github.com/issues/969. glBindFragDataLocation needs to be called before a shader program is linked, not after. Apple likes following the Core spec to the letter.

### DIFF
--- a/code/graphics/opengl/gropenglshader.cpp
+++ b/code/graphics/opengl/gropenglshader.cpp
@@ -582,6 +582,13 @@ int opengl_compile_shader(shader_type sdr, uint flags)
 				glBindAttribLocation(program->getShaderHandle(), i, GL_vertex_attrib_info[i].name.c_str());
 			}
 
+			// bind fragment data locations before we link the shader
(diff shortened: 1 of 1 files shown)


code/graphics/opengl/gropenglshader.cpp
```diff
@@ -582,6 +582,13 @@ int opengl_compile_shader(shader_type sdr, uint flags)
 				glBindAttribLocation(program->getShaderHandle(), i, GL_vertex_attrib_info[i].name.c_str());
 			}
 
+			// bind fragment data locations before we link the shader
+			glBindFragDataLocation(new_shader.program->getShaderHandle(), 0, "fragOut0");
```
asarium: `new_shader.program` isn't valid at this point so this causes a segfault at startup. Instead you must use `program`.

SamuelCho:
I guess I should've committed what I had on my Mac machine instead of duplicating it on my PC...

PR #1318, opened 2017-03-22
Use OpenGL texture arrays for animations

### LINKED ISSUE #1169: Use OpenGL texture arrays for animations
Texture arrays are a special type of OpenGL texture that stores multiple images withing a single texture object. The advantage of this is that a shader can access every single image without having to rebind the texture.

We could use this feature (which is part of the OpenGL Core context since 3.0) to speed up rendering of particles since that would be a great application of this feature. The effect batcher already groups the particles by texture and then draws all particles with the same texture with one draw call. With texture arrays this could be further improved by grouping all particles together that render using the same animation which should drastically reduce the amount of draw calls the particle code has to make.

The particle batching code would need to be changed so that the texture array index is added to the rendered vertices so that the shader can access the right texture.

While we are at it we could also just go ahead and convert all textures to array textures since FSO allows to use animations for a lot of images.

**Possible OpenGL implementation:**
Since we allow animations for pretty much all texture types, all samplers in the shaders would need to be an array sampler. This may not be necessary for single image textures but a texture array with a single image behaves pretty much like a normal texture so that shouldn't cause any issues. For the main shader the required image index would need to be passed to the shader as a uniform but since the individual indices stay the same for a single draw call, there don't need to be any changes to the vertices we pass to the GPU.

**Changes to bmpman:**
To make the usage of array textures easier *bmpman* will handle the conversion between a bitmap handle and the texture array + index.

**Possible issues:**
OpenGL texture arrays require that all images have the same size. Do we allow animations where the individual frames have different dimensions? If that is the case then those need to be handled like standard textures with only a single texture. The code could issue a warning to the modder to let them know that performance could be improved by using the same size for all textures.

asarium:
This converts all usages of bmpman textures to OpenGL texture arrays.
That means that all frames of an animation are stored in a single OpenGL
texture object and can be accessed at the same time in a single draw
call. This should significantly reduce the overhead of drawing lots of
particles since all particles of a single animation can be drawn with a
single draw call.

This fixes #1169.

This is still unfinished and requires additional testing on other OpenGL drivers and platforms.

To simplify the texture setting, I based this on the simplified render interface PR.


SamuelCho:
I feel like we should just get rid of the passthrough shader entirely and have a flag for the default material to allow use of non-array textures.

asarium:
I didn't do that because I wanted to keep texture array handling inside the graphics code since the rest of the code would always use the texture array flag. I guess the flag could always be enabled when the shader would be used by a material but I don't like either solution very much since they both break the graphics API design a bit.

SamuelCho:
Well so far there's nothing egregious about these changes and they all make sense. I'm kind of bristling at the particle shader being seperated into a batch shader to accommodate the vertex array indices but we can probably revise that later down the line if we come up with something better.

asarium:
I don't like that change either but as far as I can tell, there is no other solution except also adding the array index attribute to the default material shader.

One thing I would like to do before this gets merged is to merge the texture coordinate and the array index attribute into one shader attribute. The batching code should be able to handle that and the vertex shader would have to do a little less work. It would also be a lot cleaner.

SamuelCho:
Yeah that's not a bad idea.

asarium:
I converted the array index attribute into a 3d texture coordinate attribute which makes the array index attribute unnecessary. I'm still having issues with environment map reflections. For some reason, the entire object is now rendered as if it always reflects light.

asarium:
Environment map issues have been fixed which resolved the last issue I am currently aware of (I'm sure there are more though).

PR #1319, opened 2017-03-22
Improve handling of environment map redrawing

asarium:
This should make sure that the environment map is only redrawn if
necessary. In my case that should improve frame times by around 1ms for
the normal case when the environment map doesn't need to be redrawn.

This only affects missions which use SEXPs to change the background but since that is more and more common with new releases this should improve framerates a bit.


asarium:
I'm not sure if this should have been merged yet. It has the potential for causing regressions and I would like to avoid that before the 3.8 release.

PR #1321, opened 2017-03-24
Use built-in OpenGL clip plane instead of custom variant

asarium:
The current shaders compute the clip values in the vertex shader and
then discard fragments if those values say that the fragment is clipped.
I don't know why this was done but, as far as I can tell, this behavior
can be fully replaced by using the built-in OpenGL clip-plane feature
which should also reduce the amount of fragment shader invocations. This
won't affect performance too much since we rarely use clip planes but it
seems like the correct thing to do.

### DIFF
--- a/code/def_files/main-f.sdr
+++ b/code/def_files/main-f.sdr
@@ -82,10 +82,6 @@ uniform float neardist;
 uniform float middist;
 uniform float fardist;
 #endif
-#ifdef FLAG_CLIP
-uniform int use_clip_plane;
-in float fragClipDistance;
-#endif
 #ifdef FLAG_NORMAL_ALPHA
 uniform vec2 normalAlphaMinMax;
 #endif
@@ -270,9 +266,6 @@ void main()
 	if(fragNotVisible >= 0.9) { discard; }
 #endif
 #ifdef FLAG_SHADOW_MAP
- #ifdef FLAG_CLIP
-	if(use_clip_plane == 1) { if(fragClipDistance <= 0.0) { discard; } }
- #endif
 	// need depth and depth squared for variance shadow maps
 	fragOut0 = vec4(fragPosition.z, fragPosition.z * fragPosition.z * VARIANCE_SHADOW_SCALE_INV, 0.0, 1.0);
 	return;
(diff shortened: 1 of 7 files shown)


SamuelCho:
It was done because clip plane/distance was crashing AMD drivers when we were still on the Compatibility Profile. I decided that moving to clip distance was a low priority so I postponed moving us to it, concentrating mostly on getting the entire new Core Profile code working.

SamuelCho:
Another interesting thing to do is use clip distance to cull out invisible submodels during batched model rendering so we don't have to do that in the frag shader either.

asarium:
> It was done because clip plane/distance was crashing AMD drivers when we were still on the Compatibility Profile. I decided that moving to clip distance was a low priority so I postponed moving us to it, concentrating mostly on getting the entire new Core Profile code working.

I though that the reason was something like that. I built this on the AMD drivers and it didn't crash so I guess the issue has been fixed now.

> Another interesting thing to do is use clip distance to cull out invisible submodels during batched model rendering so we don't have to do that in the frag shader either.

I saw those `discard` statement but I don't know how the batched rendered works yet so I just left that as it was. I'll take a look at it and maybe integrate it into this PR.

asarium:
I looked though the batch code a bit and I'm not sure if the "not visible" flag is ever being set to true. All I can find is `modelrender.cpp:450` where the not-visible location in the matrix is set to zero.

SamuelCho:
Yeah the model batching code isn't quite clear so my apologies. The transforms are all initialized with the matrix4 identity. The fourth row in the fourth column of each transform  in the buffer represents if a submodel is invisible or not. That gets set to 0.0f whenever model_draw_list::add_submodel_to_batch() from model_render_buffers() is called. Submodels that aren't supposed to draw will implicitly not have model_render_buffers() called for them. Their a1d[15] will still be 1.0f when the transforms are fed to the vertex shader.

But honestly you shouldn't pay attention to any of that. void getModelTransform() in main-v.sdr is where everything gets read. Just look at the "invisible" boolean. 

And really, clipping out invisible submodels doesn't need to happen in this PR. It was just a suggestion or something to muse about.

asarium:
That's weird, if I convert the batched clipping to `gl_ClipDistance` I get a performance *decrease* of ~5%.

Here are the changes, maybe you can spot why this makes performance worse: https://github.com/scp-fs2open/fs2open.github.com/compare/master...asarium:opengl/batchedClipping

Issue #910, opened 2016-08-29
Use Uniform blocks instead of uniform variables


asarium:
I want to discuss the advantages and disadvantages of using uniform blocks instead of standard uniforms for passing data to the shaders. This does not mean that is must happen but instead I want to have an open discussion about if and how it could be implemented.

I had some good experiences with uniform blocks in a testing project where it allowed me to decouple the rendering backend from the code that is using the rendering functions since the rendering backend only gets a uniform buffer handle for the data instead of somehow getting that data down to the rendering functions. It also allows to but data that is often changed together into the same buffer and change it at the same time.
Uniform buffers are also the way shader data is passed to the shaders in Vulkan but that's just a minor point here since we are not in a position to consider adding Vulkan support yet.

The disadvantages I currently see are that it's harder to properly optimize this concept since the drivers are usually pretty good at figuring out when the updated uniform data needs to be sent to the GPU. There is also the issue that the buffer data needs to be properly aligned before the GPU can use it. I have developed a helper class for that but it makes uniform block usage a little bit more complicated. There are probably other issues I haven't encountered yet so I would like your opinion on that.

What do you think? Especially @SamuelCho since you probably already have experience with passing data to the GPU this way.

SamuelCho:
We can just start off by having uniform buffers handle relatively static
data that don't have much variance between draws. We could probably store
shadow view and projection matrices, shadow cascade distances, and all 256
lights in a per-scene uniform buffer. That way you just need a single
glBufferData before a series of draws get sent like I do with the batch
transform texture buffer object. The light data in the uniform buffer can
be indexed using a simple uniform array that's given per-draw.

We can get fancier If you want uniform buffers to handle all uniform data.
Maybe we have two block binding points. One of the block bindings is for
the static draw data described above. The second block binding is for
dynamic draw data. The uniform struct layout of our dynamic draw data
encompasses all possible uniform data inputs of the shader regardless of
what the current #define flags are for the current shader variant. And then
we update a shared dynamic buffer object per-draw using glBufferSubData,
offseting into it like we do for the immediate vertex buffer I implemented
for the OpenGL Core changes. We point glBindBufferRange to that offset
before the draw and we're ready to make the draw call.

I'd just say start us off with the per-scene uniform buffer for now. It'll
be quicker to prototype and test. Then we'll look into per-draw stuff once
we get a good handle on things.

asarium:
[...]

PR #1323, opened 2017-03-26
Use a uniform buffer object for storing deferred lighting data

asarium:
This converts the deferred lighting shader to use uniform buffers for
the dynamic data sent to the GPU while rendering the lighting effects.
This doesn't affect performance but the deferred lighting code is a
nicely isolated piece of rendering code where I could test how the
rendering code could use the uniform buffers.

This is related to #910.


SamuelCho:
Do we really need to have a generic gr_create_buffer call? The reason why gr_create_vertex_buffer and gr_create_index_buffer exist is because vertex submission is definitely something that can be done outside of the graphics code. Uniform creation however, is always something that's done internally in the graphics code. Why allow non-graphics code to be able to create uniform buffers?

asarium:
There is no reason to keep building the uniform buffers locked inside the graphics code. In fact, if we keep it inside the graphics code we will need to add new `gr_` calls for every single uniform block which will all need to replicate some kind of buffering strategy. IMO, it makes much more sense to just expose a small `gr_` API for building uniform buffers and then let the outside code handle building that data and submitting it with a single `gr_` call.

The data format shouldn't be a problem since `std140` is well defined and supported by OpenGL and Vulkan (which is only a theoretical future addition at this point but it still needs to be considered for this).

SamuelCho:
So then that means we'll have to expose setting simple uniforms as well? That seems really messy and inconsistent. At least in an engine like Unreal or Unity, the anybody invoking a shader and creating would just have to define the parameters and the user wouldn't have to worry about if a uniform was transmitted as a uniform buffer or a constant. Buffer creation should be pretty opaque outside of the graphics code.

asarium:
I think we have different expectations about what the "graphics code" and where we should insert the graphics API abstraction layer. I would like to keep that abstraction layer as small as possible to make implementing and maintaining it as easy as possible. Based on that, we could have more "convenience" code which would handle all the common tasks like sending uniforms to GPU memory.

"The user" would simply continue to use the same gr_ API but instead of directly calling into the OpenGL code it would call a common entry point which would then execute the appropriate graphics calls via our abstraction layer. That is what I would like to accomplish with these changes. I don't want to force the rest of the code to handle the uniform buffers but especially the model rendering code needs access to that so I though it would be a good idea to expose that functionality and let the rest of the code use some utility graphics code which would handle the uniform buffers for the boring stuff (e.g. transforms for 2D bitmaps and stuff like that).

While we are having this discussion, what are your future plans for the graphics code? I would like to move towards a lower-level API abstraction upon which we could build more complex functionality without directly calling the actual graphics API.

SamuelCho:
Having a lower level API abstraction is fine but by it's nature, the abstraction will have to be a level below where the gr_ calls are since the actual setting of state will now be abstracted. I kind of want the graphics code to be the place where we do all the heavy graphics lifting, while it interfaces with the a graphics device API abstraction below that. 

The way I've been leaning is towards a paradigm where stuff outside the graphics API just tells it what to draw (Verts, indices, material), then the graphics code figures out how to draw it using resources internal to it, then it calls whatever API state and draw calls it needs. I'm kind of loath to remove logic outside of gr_ calls because there's just so much there now. It might be better to start working on de-abstracting gr_ and create a new lower level abstraction for graphics devices. Projects I worked on have a general class called "GraphicsDevice" that have interfaces for granular state setting, data submission, and draw calls. They then have DirectX, OpenGL, and console graphics interfaces inherited from that that class.

asarium:
I like the idea of a generic graphics device API but, as you said, we currently have a lot of code inside the OpenGL implementation that needs to be moved into the generic graphics code first. From that standpoint, these changes don't make much sense and I now understand what you didn't want this to happen before.

I'll leave this PR open for a little while longer in case anyone else wants to comment.

SamuelCho:
I think these changes are fine so long as we're not seeing any performance degradation from this. Ultimately the gr_create_buffer stuff can go in, I just want to make it clear that something like that may be subject to change in face of larger currents.

asarium:
There are some graphics rendering libraries that provide an interface just like you described. [bgfx](https://github.com/bkaradzic/bgfx) seems to be pretty widely used and has support for a wide array of graphics APIs and platforms. It even has VR support. What do you think about using a library like this? It would involve a lot of work since it uses a custom shader dialect and converting all our code to use that library is a pretty huge task. If we first move our own engine to an interface that is similar to bgfx (or whatever library we are going to use) we could reduce the size of the individual pull requests to make code review and testing easier but it would still be a monumental coding task.

bgfx has a problem with support for more advanced shader techniques since it doesn't support geometry or tesselation shaders which is very unfortunate since our current renderer requires geometry shader support.

SamuelCho:
Yeah I don't know. bgfx is nice but at the same time, as you said, there's no geometry shader support. Plus nanovg is exclusively OpenGL. Plus rewriting all our shaders sounds really awful. 

Honestly, I feel the OpenGL Core code I wrote is going to carry us for quite some time. I'd rather focus on improving our existing code. We can definitely look at bgfx as an inspiration for an interface though, especially to clean up and define the boundaries between rendering and the graphics API.

asarium:
I only found out that bgfx does not have geometry support after I already wrote most of my post and I didn't want to delete that again. I'll use it as an inspiration for future API additions but, as you said, no geometry shaders is a deal breaker for us so using bgfx is sadly not an option for us.

The current materials system is a very good draw call abstraction which means that there is no immediate need for a new abstraction but it's always good to have a plan for the future so that new systems can be designed with that plan in mind.

asarium:
I have improved the code around uniform buffer management a bit and used this branch as its basis. I introduced a manager class which controls all the uniform buffers for a specific uniform block type. If our code needs to submit uniform data to a shader it requests a buffer from that manager and that either creates a new buffer object or reuses a currently unused buffer. Buffers are unused once the graphics operations that use that data have finished.

This is checked with OpenGL fences which are checked once a frame and if the fence has been signaled the we can be sure that the buffer is not in use anymore. This helps reduce the amount of implicit synchronization OpenGL has to do and could also help in the future if we want to use unsynchronized buffer access. This is not directly relevant for us now since all buffer updates currently use `glBufferData` which allocates new storage every time it is called but can be useful in the future.

asarium:
Conflicts have been resolved and changes have been rebased onto latest master.

PR #1328, opened 2017-03-31
Upgrade included libSDL2 to 2.0.5

chief1983:
Testing to see if this breaks any compilation.  Also removed old build system files from the .gitignore, as it was causing it to miss any changes to the *.a files added to support MinGW compilation, and they aren't needed in the CMake world.

With 2.0.5, we may be able to un-kludge some Haptic joystick support.


chief1983:
Reverted 19b5ee97f7b7ab32976d0b3e6589c2466f95f8ca to test if it allows Force Feedback sticks to opened using the correct method in SDL 2.0.5.

code/io/joy_ff-sdl.cpp
```diff
@@ -86,11 +86,8 @@ int joy_ff_init()
 		}
 	}
 #endif
-	
-	// This should be how we do it, but it's currently bugged in SDL so it fails to re-enumerate haptic axes.
-	//haptic = SDL_HapticOpenFromJoystick(joy_get_device());
 
-	haptic = SDL_HapticOpen(0);
+	haptic = SDL_HapticOpenFromJoystick(io::joystick::getCurrentJoystick()->getDevice());
```
MageKing17: A check for `currentJoystick` being `nullptr` should be added, or this line can cause a crash.

Issue #1329, opened 2017-04-01
Game generated shield icon bars are too short


AxemP:
It's been noticed that some ships seem too long to fit in the game generated shield icons. It looks like the bar height is too short.

In hudshield.cpp line 594, the clip height for a generated shield icon is 93. But the bar height on line 710 is 54. That's too short. In game the shields only go down to about 84, so the shields need to be made 9 pixels taller. 63 I guess? (I'd check but I don't have FS set up in VS at the moment)

asarium:
Has this always been broken or was this broken since the last release? I would like to narrow down if this is a bug or a regression.

AxemP:
I think it's maybe always been like this? I always (from when the feature first got added) remember seeing some ships not quite fitting, but it was never a big deal for me to investigate further. Looking at the blame history, there was a change that niffiwan did fixing mantis 2789 back in 2013 and Swifty's original antipodes hud update from 2011. So I don't know which one actually had the incorrect calculations.

niffiwan:
@Axem, is the Ray III from nuWoD a good example of the shield bars being too small?

AxemP:
Yeah I think that's a good example

niffiwan:
OK; what do you think of this; just made the change you recommended:

63
![height_63](https://cloud.githubusercontent.com/assets/3313634/24584163/67c6f138-17a7-11e7-9c5e-e781d6288e22.png)

original (54)
![height_54](https://cloud.githubusercontent.com/assets/3313634/24584164/6fcaa28a-17a7-11e7-974e-c05dd6277a66.png)

AxemP:
Even if bad closeup zoom and closeup pos were the problem, the game shouldn't be letting the icons exceed their bounds like that. In any case, the 63 look good!

asarium:
This has been fixed by #1330.

PR #1330, opened 2017-04-03
Tweak autodraw shield bar positions

niffiwan:
See #1329, this makes the rear shields slightly further away from the
front, so that the model clip area matches the area where the shields
are drawn.

### DIFF
--- a/code/hud/hudshield.cpp
+++ b/code/hud/hudshield.cpp
@@ -590,16 +590,16 @@ void HudGaugeShield::showShields(object *objp, int mode)
 
 		vm_angles_2_matrix(&object_orient, &rot_angles);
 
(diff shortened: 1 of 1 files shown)


code/hud/hudshield.cpp
```diff
@@ -590,14 +590,16 @@ void HudGaugeShield::showShields(object *objp, int mode)
 
 		vm_angles_2_matrix(&object_orient, &rot_angles);
 
-		gr_screen.clip_width = 112;
-		gr_screen.clip_height = 93;
+		static const int cw = 112;
```
asarium: This should only be `const` and the variable name should be more descriptive and upper case since it's a constant.

PR #1331, opened 2017-04-04
Add ability to dynamically add new SEXPs at runtime

### LINKED ISSUE #1320: Add scripting interface for adding custom SEXPs
We already have the capability of executing scripts from SEXPs but it would be even better to allow modders to add actual SEXPs which could be used just like any other engine SEXP. To reduce the amount of work required, the feature could be limited so that new SEXPs could only be added at a specific time when the game is initializing. Since I would like to display the custom SEXPs in FRED as well, that seems like the best choice since. 

I don't have much experience with the inner workings of the SEXP system so I'll need some guidance from more experience SEXP coders to point me into the right direction.

I'll use this issue to document my progress while implementing this feature but it will probably take some time since the SEXP system is based around the idea that the operator list is static.

asarium:
This add a system for adding generic instances of the DynamicSEXP class
to the SEXP system. The only implementation at the moment is the LuaSEXP
which is used by the mn.addSEXP lua API to allow Lua to add new SEXPs.

Since this requires that FRED also runs lua scripts I added a mod table
option to enable scripting in FRED. Since the current scripts are not
built with FRED in mind it would likely break all these scripts if FRED
would be opened with normal scripting enabled.

This fixes #1320.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -256,3 +256,7 @@ CMakeSettings.json
 
 # Weird files created by Linux
 .fuse_hidden*
+
+# This gets ignored by one of the rules above for some reason...
+!/code/scripting/lua/LuaConvert.h
+!/code/debugconsole/
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -508,6 +508,7 @@ char *Cmdline_start_mission = NULL;
 int Cmdline_old_collision_sys = 0;
 int Cmdline_dis_collisions = 0;
 int Cmdline_dis_weapons = 0;
+bool Cmdline_output_sexp_info = false;
 int Cmdline_noparseerrors = 0;
 #ifdef Allow_NoWarn
 int Cmdline_nowarn = 0; // turn warnings off in FRED
@@ -1890,7 +1891,7 @@ bool SetCmdlineParams()
 		Output_scripting_meta = true;
 
 	if (output_sexp_arg.found() ) {
-		output_sexps("sexps.html");
+		Cmdline_output_sexp_info = true;
 	}
 
 	if ( no_pbo_arg.found() )
--- a/code/cmdline/cmdline.h
+++ b/code/cmdline/cmdline.h
@@ -145,6 +145,7 @@ extern char *Cmdline_start_mission;
 extern int Cmdline_old_collision_sys;
 extern int Cmdline_dis_collisions;
 extern int Cmdline_dis_weapons;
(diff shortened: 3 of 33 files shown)


asarium:
I added support for SEXP variables but I needed to make some changes to `sexp_tree.cpp` for that. I removed some code that was, as far as I could tell, not reachable and removed the condition which checked for a SEXP which could change the value of a variable. I'm not sure if that's correct but so far everything seems to be working.

Goober5000:
Which code are you referring to when you said you thought it wasn't reachable?  Is it the popup code found in `add_default_operator` starting at line 2213?

Because that code *is* reachable.  It handles supplying the default arguments for a sexp when you right-click using the popup menu and replace an existing sexp with another sexp.

FRED is absolutely chock full of obscure code that handles special cases, so you should *never* make assumptions about *anything*.  It is regularly the case that you may need to study a piece of code extensively to make any sense out of it.

Don't forget daveb's classic advice:
> "Fred2 is the omega of all giant unwieldy pieces of code. Its big, its horrifying, and it just doesn't care. View it at your own risk." - Dave Baranec

asarium:
Shouldn't `if ( query_operator_argument_type(op, argnum) == OPF_VARIABLE_NAME)` make sure that the `argnum` and operator that is currently being handled is actually a variable parameter? If that is true then one of the following cases must be true since only those operators can have variable name parameters. Well, that was my understanding of the code...

So, if that doesn't work, do you have any recommendation for how this could be implemented? If the original code has to be kept then we need some way of determining of making the same check as the current code but which handles the dynamic SEXPs properly.

Goober5000:
It can certainly be simplified.  You're right in that the `OPF_VARIABLE_NAME` check is redundant to the sexp check.  I went back and looked at the [original implementation](https://pastebin.com/41sKyBWt) in the public source release, and it looks like someone (possibly me) made a mistake when expanding that function to cover more uses than `modify-variable`.  That sexp is special because the second argument's default argument is contingent on the first argument's type, but the current code doesn't handle that correctly.

I'll see about filing a new PR to fix just this one problem.

Goober5000:
Okay, see #1381 for the changes.

asarium:
Thank you for that the pull request. That should ensure that dynamic SEXPs will be handled properly by that code block.

Goober5000:
You're welcome.  Yes, that avoids the need to add a special hook for any dynamic sexps that use variables.

asarium:
I rebased my changes onto master and fixed the conflicts.

PR #1333, opened 2017-04-13
Fix the build on OpenBSD

oldlaptop:
FSO doesn't build from source on OpenBSD anymore. This patch corrects the only significant problem: the iterator in LuaTable.h/cpp known as _L clashes with a #define in OpenBSD's standard headers (ctypes.h to be specific). The new name proposed here is simply taken from the type.

One other problem still prevents building on OpenBSD at this time: an assert in code/scripting/ade.h that calls std::is_trivially_copyable, available in libstdc++ >= 5 (where OpenBSD has 4.9 in ports/packages). OpenBSD on i386/amd64 will probably use clang/libc++ in the future, so this issue will go away anyhow. For the record, the following diff on top of this PR fixes the build on OpenBSD 6.1-current at this time:

```diff
diff --git a/code/scripting/ade.h b/code/scripting/ade.h
index 8478ea3ea..04cf920ef 100644
--- a/code/scripting/ade.h
+++ b/code/scripting/ade.h
@@ -226,7 +226,7 @@ template<class StoreType>
 class ade_obj: public ade_lib_handle {
  public:
        // Make sure that the stored type is compatible with our requirements
-       static_assert(std::is_trivially_copyable<StoreType>::value, "ADE object types must be trivially copyable!");
+       //static_assert(std::is_trivially_copyable<StoreType>::value, "ADE object types must be trivially copyable!");
 
        ade_obj(const char* in_name, const char* in_desc, const ade_lib_handle* in_deriv = NULL) {
                ade_table_entry ate;
```

### DIFF
--- a/cmake/platformChecks.cmake
+++ b/cmake/platformChecks.cmake
@@ -7,6 +7,7 @@ include(CheckLibraryExists)
 include(CheckPrototypeDefinition)
 include(CheckStructHasMember)
 include(CheckSymbolExists)
+include(CheckCXXSymbolExists)
 include(CheckTypeSize)
 
 CHECK_INCLUDE_FILE("execinfo.h" HAVE_EXECINFO_H)
@@ -14,9 +15,18 @@ CHECK_INCLUDE_FILE_CXX("cxxabi.h" HAVE_CXXAPI_H)
 
 CHECK_TYPE_SIZE("max_align_t" MAX_ALIGN_T)
 
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    set(CMAKE_REQUIRED_FLAGS "-std=c++11") # required for g++ <= 5
+endif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+
 set(CMAKE_EXTRA_INCLUDE_FILES "cstddef")
 CHECK_TYPE_SIZE("std::max_align_t" STD_MAX_ALIGN_T LANGUAGE CXX)
+set(CMAKE_EXTRA_INCLUDE_FILES "type_traits")
+CHECK_TYPE_SIZE("std::is_trivially_copyable<int>" STD_IS_TRIVIALLY_COPYABLE LANGUAGE CXX)
 set(CMAKE_EXTRA_INCLUDE_FILES)
+
+set(CMAKE_REQUIRED_FLAGS)
+
 
 CHECK_FUNCTION_EXISTS(strcasecmp HAVE_STRCASECMP)
 CHECK_FUNCTION_EXISTS(strncasecmp HAVE_STRNCASECMP)
--- a/code/scripting/lua/LuaTable.h
+++ b/code/scripting/lua/LuaTable.h
@@ -19,7 +19,7 @@ class LuaTable;
  * be in the exact same state when you call the next method of this class.
  */
 class LuaTableIterator {
-	lua_State* _L = nullptr;
+	lua_State* _luaState = nullptr;
 
 	int _stackTop = 0;
 
--- a/cmake/platformChecks.h.in
+++ b/cmake/platformChecks.h.in
@@ -8,6 +8,8 @@
 #cmakedefine HAVE_MAX_ALIGN_T
 #cmakedefine HAVE_STD_MAX_ALIGN_T
 
+#cmakedefine HAVE_STD_IS_TRIVIALLY_COPYABLE
+
 #cmakedefine01 IS_64BIT
 
 #cmakedefine01 HAVE_STRCASECMP
--- a/code/scripting/ade.h
+++ b/code/scripting/ade.h
@@ -5,6 +5,7 @@
 #define FS2_OPEN_ADE_H
 
 #include "globalincs/pstypes.h"
+#include "platformChecks.h"
 
 extern "C" {
 #include <lauxlib.h>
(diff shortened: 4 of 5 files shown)


code/scripting/lua/LuaTable.h
```diff
@@ -19,7 +19,7 @@ class LuaTable;
  * be in the exact same state when you call the next method of this class.
  */
 class LuaTableIterator {
-	lua_State* _L = nullptr;
+	lua_State* L_State = nullptr;
```
asarium: The rest of the class uses the format `_varname` so this field should do the same. `_luaState` would be an acceptable name.

asarium:
The static assert is needed to ensure that no runtime errors occur but it could be conditionally enabled only if support for `is_trivially_copyable` is detected. This could be done using a CMake platform check. See [platformChecks.cmake](https://github.com/scp-fs2open/fs2open.github.com/blob/master/cmake/platformChecks.cmake) for examples on how that is done.

code/scripting/lua/LuaTable.h
oldlaptop: Ah. I was reluctant to do that originally because leading underscores are theoretically reserved for the standard headers...

asarium: Only global names beginning with an underscore are reserved. For names in non-global namespaces (e.g. a class) the name may begin with an underscore but the next letter must be lower case which also explains why `_L` is an invalid name.

oldlaptop:
I've just added a platform check for is_trivially_copyable, can split off into another PR if desired.

cmake/platformChecks.cmake
```diff
@@ -16,6 +17,9 @@ CHECK_TYPE_SIZE("max_align_t" MAX_ALIGN_T)
 
 set(CMAKE_EXTRA_INCLUDE_FILES "cstddef")
 CHECK_TYPE_SIZE("std::max_align_t" STD_MAX_ALIGN_T LANGUAGE CXX)
+set(CMAKE_EXTRA_INCLUDE_FILES "type_traits")
+CHECK_TYPE_SIZE("std::is_trivially_copyable<int>" STD_IS_TRIVIALLY_COPYABLE LANGUAGE CXX)
+MESSAGE(${HAVE_STD_IS_TRIVIALLY_COPYABLE})
```
asarium: This seems to be working properly, although our Travis GCC config fails this test for some reason although it compiled the previous version properly. That doesn't matter though since Clang works properly.

Once you remove this test message, this is ready to be merged.

oldlaptop: Ah, dammit. Thanks for catching... :)

oldlaptop: My local setup passes the test, by the way (6.3.0 in Debian testing). Unfortunately my machines all have either 'old' gcc (4.9) or new gcc (>6) at the moment; I'll try to investigate behavior on gcc 5 (Debian packages 5.4 at the moment, in testing).

oldlaptop: The reason gcc5 fails the test appears to be quite simple and stupid - from CMakeError.txt:

/usr/include/c++/5/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.

I should be able to push a fix shortly.

cmake/platformChecks.cmake
```diff
@@ -16,6 +17,10 @@ CHECK_TYPE_SIZE("max_align_t" MAX_ALIGN_T)
 
 set(CMAKE_EXTRA_INCLUDE_FILES "cstddef")
 CHECK_TYPE_SIZE("std::max_align_t" STD_MAX_ALIGN_T LANGUAGE CXX)
+set(CMAKE_EXTRA_INCLUDE_FILES "type_traits")
+set(CMAKE_REQUIRED_FLAGS "-std=c++11") # required for g++ <= 5
```
asarium: Since this is only relevant for GCC it should be surrounded by a check for that compiler. `if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")` should work since that is the same string our toolchain detection uses.

Since this applies to all the C++ checks, could you insert that code above the other `CHECK_TYPE_SIZE`? That will make the results more consistent.

oldlaptop: Done.

cmake/platformChecks.cmake
```diff
@@ -7,16 +7,25 @@ include(CheckLibraryExists)
...
 CHECK_INCLUDE_FILE("execinfo.h" HAVE_EXECINFO_H)
 CHECK_INCLUDE_FILE_CXX("cxxabi.h" HAVE_CXXAPI_H)
 
 CHECK_TYPE_SIZE("max_align_t" MAX_ALIGN_T)
 
-set(CMAKE_EXTRA_INCLUDE_FILES "cstddef")
-CHECK_TYPE_SIZE("std::max_align_t" STD_MAX_ALIGN_T LANGUAGE CXX)
-set(CMAKE_EXTRA_INCLUDE_FILES)
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    set(CMAKE_REQUIRED_FLAGS "-std=c++11") # required for g++ <= 5
```
asarium: That's not what I meant. Only this line should be in the if body. The other lines are required on all platforms but `-std=c++11` is only relevant for GCC.

This is how it should look like:
```cmake
if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
    set(CMAKE_REQUIRED_FLAGS "-std=c++11") # required for g++ <= 5
endif()

set(CMAKE_EXTRA_INCLUDE_FILES "cstddef")
CHECK_TYPE_SIZE("std::max_align_t" STD_MAX_ALIGN_T LANGUAGE CXX)
set(CMAKE_EXTRA_INCLUDE_FILES "type_traits")
CHECK_TYPE_SIZE("std::is_trivially_copyable<int>" STD_IS_TRIVIALLY_COPYABLE LANGUAGE CXX)
set(CMAKE_EXTRA_INCLUDE_FILES)

set(CMAKE_REQUIRED_FLAGS)
```

oldlaptop: Ah, sorry.

asarium:
Yes, that looks good now.

PR #1339, opened 2017-04-24
Revert bad fix for Coverity issue

asarium:
fa60c666bcef7f2d560d718e8741bfccb5a85c4d tried to fix an issue reported
by Coverity which was an intentional "issue". This restores the original
behavior which will cause a new coverity issue which can be ignored.

This fixes Mantis 3195.


niffiwan:
should any of the other commits in #746 be reverted as well?

asarium:
The switched argument order only affects the buttons which are only present in the volume sliders. I can revert the other instances as well but I don't think it's needed there.

PR #1345, opened 2017-05-02
Save mission flags using correct format specifier

### LINKED ISSUE #1344: Disable built in messages not working with modded files
I haven't been able to figure out exactly _which_ modded file is causing the issue, but it doesn't happen with retail data.

Test mission: https://pastebin.com/pnreGsNH
Play with JAD2.2

It has the global disable built in messages and disable built in command message options checked in the mission specs. But if you give fighters orders, they will give you those built in messages. But if I were to tick the disable built in messages for the ships through their ship editor screen, they will keep quiet.

I don't know what's causing it. I don't think its the modded messages.tbl because I copied and pasted it into my root /data/tables/ folder and the bug didn't manifest itself then. (But I could be wrong!)

asarium:
The current version simply writes the flagset object as an integer to
the mission file which surely will cause issues.

This fixes #1344.

I have not tested if the mission save is fixed now so until I can confirm that this is still unfinished.


asarium:
@AxemP Confirmed that #1345 was not an engine issue but since the flag value was still written in the wrong way these changes should still be merged.

PR #1349, opened 2017-05-16
Avoid Int3() when finding a rank with an out-of-order points value.

MageKing17:
There's a bit of debug code that checks to make sure all ranks are in ascending point order; if this check fails, FSO calls `Int3()` instead of trying to describe the problem. I've changed it to a Warning that actually explains the problem, so the modder can fix it.

Is a conversion to a Warning the correct response here? It shouldn't be an Assertion instead (it's not a coding mistake, since it results from bad data), but making it an Error seemed like a bad idea (Release builds would go from silently ignoring it to crashing outright). Nonetheless, making it a Warning means that people could continue past it, and certain code expects all point values to be in ascending order. I don't think any of them will actually crash if they aren't, but it could result in things like a forced promotion decreasing how many points the player has, or odd things happening when refusing to accept the results of a mission.

Anyone have any thoughts to add?

### DIFF
--- a/code/stats/scoring.cpp
+++ b/code/stats/scoring.cpp
@@ -105,12 +105,10 @@ void parse_rank_tbl()
 		required_string("#End");
 
 		// be sure that all rank points are in order
-#ifndef NDEBUG
 		for (idx = 0; idx < NUM_RANKS - 1; idx++) {
 			if (Ranks[idx].points >= Ranks[idx + 1].points)
-				Int3();
+				Warning(LOCATION, "Rank #%d (%s) has a higher \"$Points:\" value (%d) than the following rank (%s, %d points). This shouldn't actually crash FSO, but it might result in unexpected or incorrect behavior.\n", idx + 1, Ranks[idx].name, Ranks[idx].points, Ranks[idx+1].name, Ranks[idx+1].points);
 		}
-#endif
 	}
 	catch (const parse::ParseException& e)
 	{


asarium:
I think a warning is fine here. As you said, an Error would also crash a release build and we don't want to do that.

If the warning describes that some weird things may happen if the warning isn't fixed then it should give the modder enough incentive to fix it.

code/stats/scoring.cpp
```diff
@@ -105,12 +105,10 @@ void parse_rank_tbl()
 		required_string("#End");
 
 		// be sure that all rank points are in order
-#ifndef NDEBUG
```
asarium: Since the Warning below will only be seen in debug builds, removing this preprocessor directive isn't needed.

MageKing17: Warnings still have an impact in release builds (if too many occur, a pop-up tells the player how many there were and that the mod may be broken); a Warning should never be hidden behind a debug build check.

asarium: No, the content of `Warning` is entirely empty in [dialogs.cpp](https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/osapi/dialogs.cpp#L448) if we are in release mode.

MageKing17: Okay then, why did you break Release warning-counting in 575091aea4d0f657765b8ed2e066cb307c9f8826 ? :P

asarium: Crap. I was under the impression that I could copy and paste code when I made that change. In that case, disregard my initial comment...

I'll submit a PR for restoring the warning count later today.

code/stats/scoring.cpp
```diff
@@ -105,12 +105,10 @@ void parse_rank_tbl()
...
 		for (idx = 0; idx < NUM_RANKS - 1; idx++) {
 			if (Ranks[idx].points >= Ranks[idx + 1].points)
-				Int3();
+				Warning(LOCATION, "Rank #%d (%s) has a higher \"$Points:\" value (%d) than the following rank (%s, %d points). This shouldn't actually crash FSO, but it might result in unexpected or incorrect behavior.\n", idx + 1, Ranks[idx].name, Ranks[idx].points, Ranks[idx+1].name, Ranks[idx+1].points)
```
The-E: This line needs to be capped by a ;

MageKing17: I have no idea how that happened, given that I modified the `Int3()` call instead of deleting and replacing the entire line... but oh well.

PR #1352, opened 2017-05-18
Use TrackIR DLL with 64 suffix when compiling a 64-bit build

### LINKED ISSUE #974: Investigate what's needed for a 64-bit TrackIR dll
With nightlies now only being 64-bit, users no longer have the ability to test builds with TrackIR without requesting or compiling 32-bit builds.  The DLL released years ago is still just a solution for 32-bit builds.  Need to identify what options we have to go about getting a new 64-bit DLL created.  In a perfect world, @SamuelCho would still have the assets used to create the 32-bit DLL and creating a 64-bit one would be trivial.  If not, we will probably need someone willing to go through the process of validating a new DLL with NaturalPoint again, or at least getting the SDK and building a new DLL.

asarium:
This should make it possible to have both DLL version installed at the
same time without causing conflicts.

Combined with the new DLL, this will finally fix #974.

### DIFF
--- a/code/headtracking/trackirpublic.h
+++ b/code/headtracking/trackirpublic.h
@@ -7,7 +7,11 @@
 
 #include <SDL_syswm.h>
 
+#ifdef _WIN64
+#define TRACKIRBRIDGEDLLNAME "scptrackir64.dll"
+#else
 #define TRACKIRBRIDGEDLLNAME "scptrackir.dll"
+#endif
 
 #define SCP_INITRESULT_SUCCESS 0
 #define SCP_INITRESULT_BADKEY 1


chief1983:
I am confused why this was needed, it looked like two people reported that they had the DLL working without the suffix added.  We don't have suffixing for any of our other pre-built libraries, isn't it odd to do so in this case?

asarium:
I decided that this was a good idea since we load this library dynamically instead of linking to it at compile time. The other prebuilt libraries always have to be distributed alongside the executable but the TrackIR DLL can also be located in the _current working directory_ when FSO tries to load it.

If the engine executable should ever be moved to individual folder for different versions then the other prebuilt libraries would also need to be moved into that folder. Since the TrackIR DLL is an optional component of FSO that is placed into the FS2 root folder we still need to support both DLL versions (32- and 64-bit) at the same time which means that the files need to have different names.

chief1983:
Ok thanks for clearing that up.

chief1983:
Since you managed to have it compiling now, would there be any benefit to a recompile of the 32bit version on a newer compiler as well?  And perhaps putting 32 in its name to match the pattern?

asarium:
Putting 32 into the name would probably break too much to warrant the change but I could recompile the DLL with the same settings as the 64-bit DLL.

chief1983:
I guess it didn't seem that way if adding support for the 64bit one only required this ifdef, figured we could migrate people to a new one with 32 in the name for 3.8 final, and release the pair together as a new bundle.

asarium:
Changing the name of the existing DLL could break current installations for no good reason. If someone only uses the 32-bit FSO builds then the current DLL will continue to work fine so I don't see a good reason for chaning the name of the 32-bit bridge DLL.

The release post should probably include a warning to let people know that the 64-bit builds require a new TrackIR DLL.

PR #1362, opened 2017-05-31
Cleaning up the shockwave code's use of vectors and assertions.

MageKing17:
After someone tripped an assertion in the shockwave code which provided no details, I found some... questionable vector-related code. I changed all the code I could find that looped through the vector with iterator-based code, converted most `Assert`s into `Assertion`s for more informative debugging, and added a couple `Assertion`s to check previously-unchecked assumptions (which theoretically should never be tripped, but we'll see). A test build with the vector fixes resulted in the assertion no longer tripping so I'm fairly sure the improper erasure of the vector was causing problems (leftover data from a previous mission with a model number that no longer made sense because the model was unloaded).

### DIFF
--- a/code/weapon/shockwave.cpp
+++ b/code/weapon/shockwave.cpp
@@ -158,8 +158,8 @@ int shockwave_create(int parent_objnum, vec3d *pos, shockwave_create_info *sci, 
  */
 void shockwave_delete(object *objp)
 {
-	Assert(objp->type == OBJ_SHOCKWAVE);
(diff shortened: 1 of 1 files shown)


MageKing17:
Got the original issue to reproduce and I'm *pretty* sure the problem is related to the improper erasing of the vector elements I cleaned up, so I'm tagging this as a bugfix and will see if I can confirm the problem is fixed whenever AdmiralNelson tries my updated testing build.

MageKing17:
Fixed an improper comparison between strings and updated the description because AdmiralNelson confirmed the problem to be fixed with these changes.

Issue #1366, opened 2017-06-09
Warning dialogs now lead to application abort


Goober5000:
When a Warning dialog pops up and you click the Debug button, a breakpoint is triggered via `SDL_TriggerBreakpoint` in the `debug_int3()` function.  All very well and good so far.  However, shortly afterward is the `abort()` function which terminates the application.

It used to be that you could step through the offending code using the debugger once a breakpoint was triggered, but the `abort()` call means this is no longer possible.

I see this behavior while running MSVC 2015 Community Edition.

PR #1367, opened 2017-06-09
Invisible beam section fix

Goober5000:
While replaying Sesquipedalian's mission "The Fourth Vial", I discovered that all of the beam sections were being removed from the subspace rift beams.  I tracked this down to an unintended side-effect of the `weapon_clean_entries()` function added by Taylor in 2007 (!) in SVN revision 4019.  Invisible beam sections are initialized with no filename, but removed beam sections are also re-initialized with no filename and then culled later on.  This PR adds a check in the cleanup function to whitelist invisible beam sections while still allowing removed beam sections to be culled.

I also added a Warning for beams that (through either improper table entry creation or overzealous beam section removal) end up with 0 sections.  Although issue #1366 makes this of limited utility at present.

Finally, I copied over the expanded Assertion checks that were added to `light_add_point` and `light_add_cone` but not `light_add_point_unique` or `light_add_tube`.  The practical effect of a beam with 0 sections is that the code tries to create lights with negative radius.  Interestingly, release builds seem not to care about this.

### DIFF
--- a/code/lighting/lighting.cpp
+++ b/code/lighting/lighting.cpp
@@ -298,8 +298,12 @@ void light_add_point(const vec3d *pos, float r1, float r2, float intensity, floa
 
 void light_add_point_unique(const vec3d *pos, float r1, float r2, float intensity, float r, float g, float b, int affected_objnum, float spec_r, float spec_g, float spec_b, bool specular)
 {
-	Assert(r1 >0);
-	Assert(r2 >0);
+	Assertion(r1 > 0.0f, "Invalid radius r1 specified for light: %f. Radius must be > 0.0f. Examine stack trace to determine culprit.\n", r1);
+	Assertion(r2 > 0.0f, "Invalid radius r2 specified for light: %f. Radius must be > 0.0f. Examine stack trace to determine culprit.\n", r2);
+
+	if (r1 < 0.0001f || r2 < 0.0001f)
+		return;
+
 	if(!specular){
 		spec_r = r;
 		spec_g = g;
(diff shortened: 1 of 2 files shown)


code/lighting/lighting.cpp
```diff
@@ -298,8 +298,12 @@ void light_add_point(const vec3d *pos, float r1, float r2, float intensity, floa
 
 void light_add_point_unique(const vec3d *pos, float r1, float r2, float intensity, float r, float g, float b, int affected_objnum, float spec_r, float spec_g, float spec_b, bool specular)
 {
-	Assert(r1 >0);
-	Assert(r2 >0);
+	Assertion(r1 > 0.0f, "Invalid radius r1 specified for light: %f. Radius must be > 0.0f. Examine stack trace to determine culprit.\n", r1);
+	Assertion(r2 > 0.0f, "Invalid radius r2 specified for light: %f. Radius must be > 0.0f. Examine stack trace to determine culprit.\n", r2);
+
+	if (r1 < 0.0001f || r2 < 0.0001f)
```
asarium: This condition is not needed and will always be true since we assume that assertions are always true, even in release mode.

code/lighting/lighting.cpp
```diff
@@ -341,8 +345,12 @@ void light_add_point_unique(const vec3d *pos, float r1, float r2, float intensit
 // beams affect every ship except the firing ship
 void light_add_tube(const vec3d *p0, const vec3d *p1, float r1, float r2, float intensity, float r, float g, float b, int affected_objnum, float spec_r, float spec_g, float spec_b, bool specular)
 {
-	Assert(r1 >0);
-	Assert(r2 >0);
+	Assertion(r1 > 0.0f, "Invalid radius r1 specified for light: %f. Radius must be > 0.0f. Examine stack trace to determine culprit.\n", r1);
+	Assertion(r2 > 0.0f, "Invalid radius r2 specified for light: %f. Radius must be > 0.0f. Examine stack trace to determine culprit.\n", r2);
+
+	if (r1 < 0.0001f || r2 < 0.0001f)
```
asarium: Same here.

code/lighting/lighting.cpp
Goober5000: I copied across this condition because it appeared in two out of the four light functions, and it seemed reasonable to have it in all four.  You think it shouldn't even have been added in the first place?

asarium: Yes since it's unnecessary. The `Assertion` ensure that the condition `r1 > 0.0f` holds (and we don't consider the release case where that check doesn't exist since an `Assertion` is there to ensure a correct state).

Goober5000: Eh, makes sense I suppose.  The two original conditions were added in @SamuelCho's enormous commit c6203723, but it looks like it happened in the normal course of writing the function, not because of any observed errors.  (I could be wrong; perhaps he'll comment.)

PR #1368, opened 2017-06-09
Remove abort call from debug_int3

### LINKED ISSUE #1366: Warning dialogs now lead to application abort

asarium:
This fixes #1366.

### DIFF
--- a/code/osapi/osapi.cpp
+++ b/code/osapi/osapi.cpp
@@ -391,11 +391,6 @@ void debug_int3(const char *file, int line)
 #endif
 
 	gr_activate(1);
-	// we have to call os_deinit() before abort() so we make sure that SDL gets
-	// closed out and we don't lose video/input control
-	os_deinit();
-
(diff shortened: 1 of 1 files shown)


Goober5000:
Looks good to me.  Any idea why the abort() was added in the first place?

asarium:
AFAIK, it was there in the Unix/SDL version and then got copied over when the SDL merge happened. It's another case of slightly different behavior across platforms that we only catch now that the two code bases are the same.

PR #1369, opened 2017-06-09
Update qtFRED development files

asarium:
This includes the following changes:

- Connected movement and rotation speed actions with the view physics
- Connected view setting actions with the view settings of the renderer
- Added cursor modes and handling of cursor when it moves over a surface
- Movement speed shortcuts have been rebound to avoid conflicts with camera controls
- Add an application window icon. There is no resource file at the moment so the executable file still has the standard icon.


TRBlount:
An issue exists where rotating too far in any direction causes a crash. Here's a call stack dump: https://pastebin.com/3xspWBQR

asarium:
Ok, should be fixed now.

PR #1370, opened 2017-06-09
Add Briefing Editor Dialog

TRBlount:
This adds a UI for the Briefing Editor.

### DIFF
--- a/qtfred/src/ui/BriefingEditorDialog.cpp
+++ b/qtfred/src/ui/BriefingEditorDialog.cpp
@@ -1,0 +1,34 @@
+#include "BriefingEditorDialog.h"
+#include "ui_briefingeditordialog.h"
+#include <QtWidgets/QMenuBar>
+
+namespace fso {
+namespace fred {
+namespace dialogs {
+
+BriefingEditorDialog::BriefingEditorDialog(QWidget *parent) :
+    QDialog(parent),
+    ui(new Ui::BriefingEditorDialog)
+{
+    ui->setupUi(this);
+
+    QMenuBar *menubar = new QMenuBar();
+    QMenu *teamMenu = new QMenu("Select Team");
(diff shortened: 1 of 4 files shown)


qtfred/src/ui/BriefingEditorDialog.h
```diff
@@ -0,0 +1,22 @@
+#ifndef BRIEFINGEDITORDIALOG_H
+#define BRIEFINGEDITORDIALOG_H
+
+#include <QDialog>
+
+namespace Ui {
+class BriefingEditorDialog;
+}
+
+class BriefingEditorDialog : public QDialog
```
asarium: Dialogs will be put into the `fso::fred::dialogs` namespace so that should also be done here. You can make the UI compiler generate the class with the appropriate namespace by specifying the name of the class with the full namespace (see FredView.ui for an example).

asarium:
There was also a slight error in the CMake setup which broke UI file compilation at some point. If you rebase your changes onto the latest qtfred commit it should be fixed.

TRBlount:
Alright, updated to use the namespaces and rebased.

TRBlount:
Any clue as to what's causing linux compilation to fail?

qtfred/src/ui/BriefingEditorDialog.cpp
```diff
@@ -0,0 +1,34 @@
+#include "BriefingEditorDialog.h"
+#include "ui_briefingeditordialog.h"
```
MageKing17: > ../qtfred/src/ui/BriefingEditorDialog.cpp:2:37: fatal error: ui_briefingeditordialog.h: No such file or directory

asarium:
@TRBlount You are using the wrong case for the header file. It should be `ui_BriefingEditorDialog.h`.

PR #1372, opened 2017-06-10
Add AVX2 to the list of supported MSVC compiler optimizations.

MageKing17:
Also adds AVX2 autodetection so that AVX2-capable processors will generate AVX2 projects by default. To my surprise, there is actually a noticeable performance increase between AVX and AVX2 builds.

With the help of @TRBlount (he runs at a higher resolution than I do), I generated this graph showing the FPS differences (running Icarus as the benchmark):
A graph comparing the FPS of a runthrough of Icarus with an AVX build and an AVX2 build.

The speedup appears to be related to models loading faster the first time they appear.

### DIFF
--- a/cmake/cpufeatures.cpp
+++ b/cmake/cpufeatures.cpp
@@ -27,6 +27,7 @@ int main( int argc, char* argv[] )
 	int SSE     = false;
 	int SSE2    = false;
 	int AVX     = false;
+	int AVX2    = false;
 
 	int info[4];
 	cpuid(info, 0);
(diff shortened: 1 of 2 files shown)


asarium:
I added this to the 3.8.2 milestone since it's a feature and may cause regressions.

MageKing17:
I'm not sure how it could; unless AVX2 is added to the 3.8 release post, it won't make a difference, and the cmake changes don't matter unless you actually generate an AVX2 project.

I agree it is technically a new feature, but I don't see how it could possibly introduce a regression for anyone not actually compiling an AVX2 build.

(Still, by the same reasoning, it's not like it actually matters whether the PR gets merged before or after 3.8 because it won't be added to the release anyway.)

asarium:
It's only a very small chance but if we merge this now it will be included in the source code export which could then generate AVX2 builds. I agree that the actual chance of a regression is small but technically it's still a new feature and we're in a feature freeze.

The-E:
This is definitely something that can wait till 3.8 is out.

PR #1374, opened 2017-06-14
Make l_weapon use IFFs in Team var

TRBlount:
`l_weapon.Team` was using `Num_teams` which only changes value from 1 to 2 in response to whether or not the current game is a multiplayer session. Changing it use `Num_iffs` makes it coherent with the `l_Team` Indexer.


z64555:
Trying to get me more involved, I see. :P

PR #1376, opened 2017-06-15
Adds LUA versioning and copyright info to the scripting.html output

z64555:
* Credits roll now grabs the LUA copyright info from liblua, instead of being hardcoded
* Credits roll now has LUA version info as well

### DIFF
--- a/code/scripting/lua.cpp
+++ b/code/scripting/lua.cpp
@@ -271,6 +271,9 @@ void script_state::OutputLuaMeta(FILE *fp)
 	ade_table_entry *ate;
 	fputs("<dl>\n", fp);
 
+	//***Version info
+	fprintf(fp, "<dd>Version: %s</dd>\n", LUA_RELEASE);
+
 	SCP_vector<ade_table_entry*> table_entries;
 
 	//***TOC: Libraries
--- a/code/menuui/credits.cpp
+++ b/code/menuui/credits.cpp
@@ -28,6 +28,7 @@
 #include "sound/audiostr.h"
 #include "ui/ui.h"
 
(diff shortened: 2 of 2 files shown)


code/scripting/lua.cpp
```diff
@@ -271,6 +271,11 @@ void script_state::OutputLuaMeta(FILE *fp)
 	ade_table_entry *ate;
 	fputs("<dl>\n", fp);
 
+	//***Version and copyright info
+	fprintf(fp, "<dd>%s</dd>\n", LUA_RELEASE);
+	fprintf(fp, "<dd>%s</dd>\n", LUA_COPYRIGHT);
```
asarium: Do we need the Lua copyright in the scripting documentation? If it's included like this it might suggest that the documentation is covered by this copyright which is not the case. Maybe it could be clarified by using something like `<dd>Lua Copyright: %s</dd>`.

z64555: I don't think we do need the copywrite, I originally just wanted to put in the version number, but I was under the impression from @MageKing17 that it might be a good idea to put in the copywrite and authors, too.

z64555:
Oh, didn't know that I could squash from github. Learning something every day. :D

z64555:
Added "Version:" ahead of the version string to hopefully make it more apparent as to what it is.

MageKing17:
Travis failure appears to be the prebuilt binaries failing to download again; I've restarted the failed builds.

asarium:
Since there is no way that this could introduce a regression I'm just going to merge this now since it would be very useful to have in an official release.

PR #1378, opened 2017-06-17
Fix inverted "Has shields" flag in initial status dialog

asarium:
This also fixes a slight behavior change for when multiple selected ships had inconsistent values.

### DIFF
--- a/fred2/initialstatus.cpp
+++ b/fred2/initialstatus.cpp
@@ -30,6 +30,21 @@ void initial_status_mark_dock_leader_helper(object *objp, dock_function_info *in
 void initial_status_unmark_dock_handled_flag(object *objp, dock_function_info *infop);
 void reset_arrival_to_false(int shipnum, bool reset_wing);
 
+/**
+ * @brief Handles setting a flag on a flagset when the value is inconsistent
+ *
+ * This is necessary in case multiple ships with inconsistent object flags have been selected in which case
+ * that flag may not be edited since it would corrupt the value of that flag. This function simplifies handling
+ * that case.
+ */
+template<typename T>
+static void handle_inconsistent_flag(flagset<T>& flags, T flag, int value) {
+	if (value == 1) {
+		flags.set(flag);
+	} else if (value == 0) {
+		flags.remove(flag);
+	}
+}
 
 /////////////////////////////////////////////////////////////////////////////
 // initial_status dialog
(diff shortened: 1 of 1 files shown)


fred2/initialstatus.cpp
```diff
@@ -459,15 +479,17 @@ void initial_status::OnOK()
...
-		Objects[cur_object_index].flags.set(Object::Object_Flags::No_shields, m_has_shields != 0);
-
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Force_shields_on, m_force_shields == 1);
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Ship_locked, m_ship_locked == 1);
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Weapons_locked, m_weapons_locked == 1);
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Primaries_locked, m_primaries_locked == 1);
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Secondaries_locked, m_secondaries_locked == 1);
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Lock_all_turrets_initially, m_turrets_locked == 1);
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Afterburner_locked, m_afterburner_locked == 1);
+		Objects[cur_object_index].flags.set(Object::Object_Flags::No_shields, m_has_shields == 0);
```
MageKing17: As far as I can tell, this line appears to be the only functional change in this entire PR. How exactly is a function that calls either `.set(flag)` or `.remove(flag)` depending on `value` different than the previous behavior of calling `.set(flag, value)`, given that `.remove(flag)` internally just calls `.set(flag, false)`?

asarium: It's important because `value` is not a real boolean value since it can also have the value 2 for when the editor detected an inconsistent flag state when multiple objects are selected. In that case `handle_inconsistent_flag` does nothing which is the same behavior as before the flagset change.

MageKing17: Ah, I see; I misread the `else if` branch as being just an `else`. Now it makes sense.

Goober5000: It looks like you flipped the condition here; `m_has_shields != 0` becomes `m_has_shields == 0`.

fred2/initialstatus.cpp
```diff
@@ -438,17 +453,22 @@ void initial_status::OnOK()
 				if (hflag)
 					MODIFY(objp->hull_strength, (float) m_hull);
 
-				objp->flags.set(Object::Object_Flags::No_shields, m_has_shields == 0);
+				if (m_has_shields == 1)
+					objp->flags.remove(Object::Object_Flags::No_shields);
+				else if (!m_has_shields)
```
Goober5000: More of a nitpick than anything, but I would use `else if (m_has_shields == 0)` to match the `if (m_has_shields == 1)`.

fred2/initialstatus.cpp
MageKing17: @Goober5000 Yes, that would be the fix mentioned in the title of the PR. It was backwards before.

Goober5000: Gak, I wasn't even looking at Object_Flags::No_shields.  Ok then.

fred2/initialstatus.cpp
asarium: That's how the old code did it and I didn't want to change that but I can certainly do that.

Goober5000: It would be good for visual consistency.

asarium:
@Goober5000 I updated the condition.

fred2/initialstatus.cpp
```diff
@@ -459,15 +479,17 @@ void initial_status::OnOK()
...
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Weapons_locked, m_weapons_locked == 1);
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Primaries_locked, m_primaries_locked == 1);
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Secondaries_locked, m_secondaries_locked == 1);
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Lock_all_turrets_initially, m_turrets_locked == 1);
-        Ships[m_ship].flags.set(Ship::Ship_Flags::Afterburner_locked, m_afterburner_locked == 1);
+		Objects[cur_object_index].flags.set(Object::Object_Flags::No_shields, m_has_shields == 0);
+
+		// We need to ensure that we handle the inconsistent "boolean" value correctly. Not strictly needed here but just to be safe...
+		handle_inconsistent_flag(Ships[m_ship].flags, Ship::Ship_Flags::Force_shields_on, m_force_shields);
+		handle_inconsistent_flag(Ships[m_ship].flags, Ship::Ship_Flags::Force_shields_on, m_force_shields);
```
MageKing17: Duplicate line?

asarium: Not anymore :stuck_out_tongue:

fred2/initialstatus.cpp
```diff
@@ -438,17 +453,22 @@ void initial_status::OnOK()
...
-                shipp->flags.set(Ship::Ship_Flags::Force_shields_on, m_force_shields == 1);
-                shipp->flags.set(Ship::Ship_Flags::Ship_locked, m_ship_locked == 1);
-                shipp->flags.set(Ship::Ship_Flags::Weapons_locked, m_weapons_locked == 1);
-                shipp->flags.set(Ship::Ship_Flags::Primaries_locked, m_primaries_locked == 1);
-                shipp->flags.set(Ship::Ship_Flags::Secondaries_locked, m_secondaries_locked == 1);
-                shipp->flags.set(Ship::Ship_Flags::Lock_all_turrets_initially, m_turrets_locked == 1);
-                shipp->flags.set(Ship::Ship_Flags::Afterburner_locked, m_afterburner_locked == 1);
+				// We need to ensure that we handle the inconsistent "boolean" value correctly
+				handle_inconsistent_flag(shipp->flags, Ship::Ship_Flags::Force_shields_on, m_force_shields);
+                handle_inconsistent_flag(shipp->flags, Ship::Ship_Flags::Force_shields_on, m_force_shields);
```
asarium: That is an outdated commit.

TRBlount: I had an in-progress review, so I just approved that.

PR #1381, opened 2017-06-20
Fix handling of arguments for sexps that use variables

Goober5000:
This fixes the handling of default operators in the case of modify-variable, where the behavior had diverged slightly from retail as described in #1331.  (See [the original code](https://pastebin.com/41sKyBWt) which this new PR more closely resembles.)

This PR has two main effects:
1) simplify the checking for sexps that use variable names: as @asarium pointed out, the `OPF_VARIABLE_NAME` is sufficient and there is no need to check sexp identifiers and argument numbers
2) properly check the type of the second argument of `modify-variable`, according to the type of the first argument (which is the actual variable name)

### DIFF
--- a/fred2/sexp_tree.cpp
+++ b/fred2/sexp_tree.cpp
@@ -2187,7 +2187,7 @@ void sexp_list_item::destroy()
 	}
 }
 
-int sexp_tree::add_default_operator(int op, int argnum)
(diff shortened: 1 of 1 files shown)


Goober5000:
I'm of the same opinion.

PR #1388, opened 2017-06-24
fix an old Volition bug in the targeted sexp

Goober5000:
There is an old bug, dating back to FS1 retail (!) where the `targeted` sexp wasn't always reliable.  Sometimes it would fire immediately rather than waiting for the specified time duration.

Just now I figured out why.  There is a special `Players_target`, `Players_targeted_subsys`, and `Players_target_timestamp` set of variables that are updated every frame for the benefit of this sexp.  Unfortunately, the `sexp_targeted` function used `Player_ai->target_objnum` and `Player_ai->targeted_subsys` instead of the former two variables.  This made it possible for the sexp to fire during the single frame when `Player_ai->target_objnum` and `Player_ai->targeted_subsys` held the new value but `Players_target_timestamp` held the old value.

This bug also affected `sexp_node_targeted` so I fixed it in that function too.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -15031,7 +15031,7 @@ int sexp_targeted(int node)
 		}
 
 		if (CDR(CDR(node)) >= 0) {
-			ptr = Player_ai->targeted_subsys;
+			ptr = Players_targeted_subsys;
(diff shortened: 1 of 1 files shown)


Goober5000:
Thanks!

PR #1390, opened 2017-06-25
Ignore leading whitespace for gr.drawString

asarium:
This was reported by @AxemP who wanted to have leading tabs in a text
line. However, `split_str` removed those characters by default since it
ignores leading whitespace. To fix that unexpected behavior of
gr.drawString I added a new paramter to `split_str` which disabled
whitespace removal and enabled that flag for the scripting function.

That should not be a problem for existing scripts since the expected
behavior would be that leading whitespace remains in place and since we
have not received any bug reports about missing white space I'm going to
assume that no one has used the function like this before removed those
characters by default since it ignores leading whitespace. To fix that
unexpected behavior of gr.drawString I added a new paramter to `split_str`
which disabled whitespace removal and enabled that flag for the
scripting function.

That should not be a problem for existing scripts since the expected
behavior would be that leading whitespace remains in place and since we
have not received any bug reports about missing white space I'm going to
assume that no one has used the function like this before.

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -3438,7 +3438,7 @@ char *split_str_once(char *src, int max_pixel_w)
 //	returns:			number of lines src is broken into
 //						-1 is returned when an error occurs
 //
-int split_str(const char *src, int max_pixel_w, int *n_chars, const char **p_str, int max_lines, char ignore_char)
+int split_str(const char *src, int max_pixel_w, int *n_chars, const char **p_str, int max_lines, char ignore_char, bool strip_leading_whitespace)
 {
 	char buffer[SPLIT_STR_BUFFER_SIZE];
 	const char *breakpoint = NULL;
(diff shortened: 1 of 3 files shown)


ngld:
It still strips leading whitespace for the first line because of the while loop before the for loop.

asarium:
Yes, I didn't see that loop. It should be fixed now.

code/parse/parselo.cpp
```diff
@@ -3457,7 +3457,7 @@ int split_str(const char *src, int max_pixel_w, int *n_chars, const char **p_str
 	ignore_until_whitespace = 0;
 
 	// get rid of any leading whitespace
-	while (is_white_space(*src))
+	while (strip_leading_whitespace && is_white_space(*src))
```
z64555: Probably better just to remove this loop entirely, since it's followed by a similar loop for greyspace stripping

code/parse/parselo.cpp
```diff
@@ -3574,7 +3574,7 @@ int split_str(const char *src, int max_pixel_w, SCP_vector<int> &n_chars, SCP_ve
 	memset(buffer, 0, SPLIT_STR_BUFFER_SIZE);
 
 	// get rid of any leading whitespace
-	while (is_white_space(*src))
+	while (strip_leading_whitespace && is_white_space(*src))
```
z64555: Same as above

code/parse/parselo.cpp
asarium: I would like to avoid changing this code too much to avoid regressions.

PR #1392, opened 2017-06-27
Add waypoint editor

asarium:
This adds the first fully functioning editor window to qtFRED. The system uses a Model-View system to separate the GUI view from the underlying data model. Updating the view is done through a signal which is emitted by the model when it changes and the view can change the model using normal function calls.

This is how I would like to build all the editor dialogs so this is the time and place to discuss possible alternatives.

### DIFF
--- a/qtfred/src/ui/FredView.cpp
+++ b/qtfred/src/ui/FredView.cpp
@@ -275,7 +275,29 @@ void FredView::connectActionToViewSetting(QAction* option, bool* destination) {
 }
 
 void FredView::showContextMenu(const QPoint& globalPos) {
-	_viewPopup->exec(globalPos);
+	auto localPos = ui->centralWidget->mapFromGlobal(globalPos);
+
+	auto obj = _viewport->select_object(localPos.x(), localPos.y());
+	if (obj >= 0) {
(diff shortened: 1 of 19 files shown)


qtfred/src/ui/FredView.cpp
```diff
@@ -582,6 +620,10 @@ void FredView::on_actionBriefing_triggered(bool) {
 	auto eventEditor = new dialogs::BriefingEditorDialog(this);
 	eventEditor->show();
 }
+void FredView::on_actionWaypoint_Paths_triggered(bool) {
+	auto editorDialog = new dialogs::WaypointEditorDialog(this, _viewport);
```
TRBlount: Linux and Windows CI builds are failing here. This seems like a perfectly valid line so I'm not sure what the issue is.

TRBlount: Ah wait, the WaypointEditorDialog and related Model files weren't added to the source_groups for the qtfred project.

asarium:
Damn, I had some whitespace and merging issues and those changes apparently didn't make it through my fixes...

Should be fixed now.

PR #1398, opened 2017-07-01
Add object orientation editor

asarium:
This adds another small editor to qtFRED. I also found a minor bug in the old FRED code which caused a negative indexation of an array but apparently that isn't a problem in the usual use cases. I fixed that issue for qtFRED.


Goober5000:
@asarium If the minor bug affects the currently used FRED builds, it should still be fixed.

asarium:
Good idea. I have submitted the FRED2 equivalent of the fix in #1400.

PR #1399, opened 2017-07-03
Use unique_ptr for storing HUD gauge pointers

asarium:
The previous system used naked pointers which probably worked fine but
this should give us some extra memory-leak security.

@MageKing17 `ship_info::clone` is only used by the ship template feature.
Is my assumption correct that this will always be called before the HUD is
parsed? If not then the previous code is invalid and needs to be fixed. Otherwise
this is a simple cleanup PR which does not need to be in 3.8.

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -54,7 +54,7 @@
 #include "weapon/emp.h"
 #include "weapon/weapon.h"
 
-SCP_vector<HudGauge*> default_hud_gauges;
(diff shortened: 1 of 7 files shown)


MageKing17:
> Is my assumption correct that this will always be called before the HUD is parsed?

Uh... probably? I mean, if ship classes weren't parsed before HUDs it would greatly complicate the parsing of ship-specific HUD definitions, but on the other hand I did something silly to allow weapons to reference SSM strikes by name even though weapons are parsed first, so... "assumption should hold until someone does something silly", I guess.

PR #1400, opened 2017-07-03
Fix invalid indexing in object orientation editor

asarium:
This could happen if the user enabled "Point to" and selected "Object"
without having an object selected. This can only happen when there is
only one object in the mission so it's only a minor bug but still
something worth fixing.

### DIFF
--- a/fred2/orienteditor.cpp
+++ b/fred2/orienteditor.cpp
@@ -215,9 +215,10 @@ void orient_editor::update_object(object *ptr)
 		loc.xyz.y = convert(m_location_y);
 		loc.xyz.z = convert(m_location_z);
 		if (((CButton *) GetDlgItem(IDC_POINT_TO_OBJECT))->GetCheck() == 1) {
-			v = Objects[index[m_object_index]].pos;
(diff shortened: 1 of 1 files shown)


Goober5000:
As far as I can tell from a cursory examination, `m_object_index` can only be -1 if the object type is a waypoint, which the enclosing if() block already checks for.  Is there another code path that can produce an error?

asarium:
It could be that `DDX_CBIndex` sets the value to `-1` if there is no item to select. I have verified that this can happen by debugging FRED. If you start FRED, open the orientation editor for the initial object and then press OK then the index will be `-1`.

Goober5000:
Fair enough.  Is there another way to check for a 0,0,0 vector than that questionable if() statement on line 228?

asarium:
It could be done with `vm_vec_cmp(&v, &vmd_null_vector)` but that would be more than a bug fix so this is not the right place to fix that.

Goober5000:
Looks like `IS_VEC_NULL` is the way to fix it.  I would have included it because it's a bugfix, and it's included in the scope of the PR, but I'll open a new one.

PR #1402, opened 2017-07-06
Discard non-main thread events for the frame profiler

asarium:
The old code is not able to handle multi-threaded events so this change
simply filters out all the events that originated from another thread.

This fixes an Assertion caused by an invalid event.

### DIFF
--- a/code/tracing/FrameProfiler.cpp
+++ b/code/tracing/FrameProfiler.cpp
@@ -136,6 +136,17 @@ void FrameProfiler::processEvent(const trace_event* event) {
 		return;
 	}
 
+	if (_mainThreadID == -1) {
+		// We need the ID of the main thread to filter out multi threaded events. We just assume that the first event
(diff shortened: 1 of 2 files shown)


code/tracing/FrameProfiler.cpp
```diff
@@ -136,6 +136,17 @@ void FrameProfiler::processEvent(const trace_event* event) {
 		return;
 	}
 
+	if (_mainThreadID == -1) {
+		// We need the ID of the main thread to filter out multi threaded events. We just assume that the first event
+		// of the main process we see it from the main thread. That should be a safe assumption.
```
MageKing17: Minor typo: "we see it from the main thread" looks like it should be "we see **is** from the main thread".

asarium: Yes, that was a typo. Should be fixed now.

PR #1404, opened 2017-07-07
minor enhancements to the debug sexp

Goober5000:
This does a few things:
1) Adds the ability to specify a plain old string as the debug message, instead of requiring it to be an in-mission message (it looks like Karajorma tried to do this but used the wrong method)
2) Makes the sexp description a little more clear
3) Removes the "(optional)" descriptor for the second argument because it isn't actually optional

Also, I removed an extra period in one of the other sexp descriptions.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -20448,6 +20448,7 @@ void sexp_debug(int node)
 	node = CDR(node); 
 	Assertion (node >= 0, "No message defined in debug SEXP"); 
 	id = CTEXT(node);
+	strcpy_s(temp_buf, id);
 
 	for (i=0; i<Num_messages; i++) {
 		// find the message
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -20448,6 +20448,7 @@ void sexp_debug(int node)
 	node = CDR(node); 
 	Assertion (node >= 0, "No message defined in debug SEXP"); 
 	id = CTEXT(node);
+	strcpy_s(temp_buf, id);
```
asarium: Should `sexp_replace_variable_names_with_values` be called on this buffer if it's not a valid message id? I think that would make sense since the message variant also does this.

asarium:
I added this to the 3.8.2 milestone since this is not a bug fix (as far as I can tell).

code/parse/sexp.cpp
Goober5000: I suppose.  I've modified the sexp function to be a lot more like `sexp_set_death_message`, which it is conceptually close to.

Goober5000:
It's not a clear-cut bugfix, but I would still put it under that category.  I was debugging a FRED mission recently and was getting frustrated that the plain old string wasn't working, despite what the sexp implied.  I had to clutter up my list of messages with every slightly different debug message I wanted to display.

Goober5000:
Cool. I'll wait for someone to merge it or do it myself in the next day or two.

PR #1405, opened 2017-07-07
Fix unloading streaming ANI files.

MageKing17:
Two commits here. The first removes a condition from `message_mission_free_avi()` that didn't seem to serve a useful purpose. The second is the real meat of the PR and so I'll copy its description here:

When streaming an ANI file, first `anim_load()` is called, then `init_anim_instance()`. The latter increments the animation's `instance_count`. If `anim_free()` is called on an animation that still has instances, it refuses to free the animation. `generic_anim_unload()` first called `anim_free()`, then `free_anim_instance()`.

The fact that ANI files would ever unload at all seems to be because some code would unload twice. I can only assume that, combined with the fact that the bug only applies to ANI files, explains how this bug went undetected since November of 2009 (commit b43417882fef18371b6d58c05095d4ec8525c4f4 seems to be when it was added, from a quick glance at git blame).
<br>
This was causing problems when some ANIs didn't unload and slowly accumulated between missions, eventually resulting in all 64 cfile blocks getting used up. ~~For unknown reasons, it doesn't seem to happen to ANIs located in a VP file, even though all the relevant code doesn't seem to care whether or not the file came from a packfile (beyond whether or not it's memory-mapped, but that doesn't change the fact that a cfile block should be consumed, and *stay* consumed until explicitly unloaded by the missionmessage code, either way). This is especially bizarre in that, even without this PR, even having more headanis than the cfile block limit in the same mission (a guaranteed tripping of the "out of cfile blocks" assertion when the ANI files are loose) works perfectly fine if the headanis have been put in a VP file. This nonsensical outcome makes me reluctant to let this PR get merged when we apparently still don't understand the underlying problem; anybody have any idea what's going on there?~~ @ngld discovered why files in VP archives didn't exhibit the same problem, and I've added a third commit to fix the issue.

### DIFF
--- a/code/anim/animplay.h
+++ b/code/anim/animplay.h
@@ -40,9 +40,8 @@ typedef struct {
 
 enum
 {
-	PAGE_FROM_MEM		  = 0,
-	PAGE_FROM_DISK		  = 1,
-	PAGE_FROM_DISK_FORCED = 2
+	PAGE_FROM_DISK		  = 0,
+	PAGE_FROM_MEM		  = 1
 };
 
 extern int Anim_paused;
@@ -59,7 +58,7 @@ int				anim_stop_playing(anim_instance* anim_instance);
 int				anim_show_next_frame(anim_instance *instance, float frametime);
 void				anim_release_all_instances(int screen_id = 0);
 void				anim_release_render_instance(anim_instance* instance);
-anim			  *anim_load(const char *name, int cf_dir_type = CF_TYPE_ANY, int file_mapped = PAGE_FROM_MEM);
+anim			  *anim_load(const char *name, int cf_dir_type = CF_TYPE_ANY, int file_mapped = PAGE_FROM_DISK);
 int				anim_free(anim *ptr);
 int				anim_playing(anim_instance *ai);
 int				anim_write_frames_out(char *filename);
--- a/code/anim/animplay.cpp
+++ b/code/anim/animplay.cpp
@@ -765,7 +765,7 @@ anim *anim_load(const char *real_filename, int cf_dir_type, int file_mapped)
 		}
 
 		// couldn't memory-map file... must be in a packfile, so stream manually
-		if ( file_mapped && !ptr->cfile_ptr ) {
+		if ( file_mapped == PAGE_FROM_MEM && !ptr->cfile_ptr ) {
 			ptr->flags &= ~ANF_MEM_MAPPED;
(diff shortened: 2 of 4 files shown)


MageKing17:
@ngld stepped through `anim_load()` with a debugger and discovered why files in VP archives didn't consume cfile blocks the whole time the animation was loaded. I've added a commit to fix the issue, which seems to have been around since 735f5e199b7063e66e6949701779581476f1acf6 from 2003. Now this PR just needs reviewed.

code/anim/animplay.h
```diff
@@ -59,7 +58,7 @@ int				anim_stop_playing(anim_instance* anim_instance);
 int				anim_show_next_frame(anim_instance *instance, float frametime);
 void				anim_release_all_instances(int screen_id = 0);
 void				anim_release_render_instance(anim_instance* instance);
-anim			  *anim_load(const char *name, int cf_dir_type = CF_TYPE_ANY, int file_mapped = PAGE_FROM_MEM);
+anim			  *anim_load(const char *name, int cf_dir_type = CF_TYPE_ANY, int file_mapped = PAGE_FROM_DISK);
```
asarium: The generic animation system uses the integer literal `0` for the `file_mapped` parameter which now causes a change of behavior. That should be fixed to use the appropriate enum value. The code for this is in [generic.cpp, line 196](https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/graphics/generic.cpp?utf8=%E2%9C%93#L196).

Everything else uses the default parameter value so there should be no change needed.

MageKing17: This isn't a change in behavior because the new default is still `0`.

asarium: I guess that's true but the function should still use the enum constant and this would be a good time to fix that.

MageKing17:
Removing the memory-mapped code entirely is somewhat outside the scope of this PR, but yes, once this PR is merged there's no particular reason to keep it around.

asarium:
Of course, that was just an observation, not a change request.

code/anim/animplay.h
MageKing17: A hypothetical later PR to remove the memory-mapped code would be removing the parameter anyway, so I don't particularly see the point in "fixing" it in this one.

asarium: I guess it doesn't matter too much and since you already showed that my original complaint is not actually relevant I see no further reason to delay this.

PR #1408, opened 2017-07-07
Add Mission Specs Editor

TRBlount:
This adds functioning dialogs for:
* Mission Specs Editor
* Custom Wing Names Editor

The Sound Environment Editor is being delayed as the Sound Environment system does not actually work.

Ready to merge.

### DIFF
--- a/qtfred/src/mission/dialogs/CustomWingNamesDialogModel.cpp
+++ b/qtfred/src/mission/dialogs/CustomWingNamesDialogModel.cpp
@@ -15,14 +15,16 @@ bool CustomWingNamesDialogModel::apply() {
 	int i;
 	if (strcmp(_m_starting[0].c_str(), _m_tvt[0].c_str()))
 	{
-		//MessageBox("The first starting wing and the first team-versus-team wing must have the same wing name.");
+		auto button = _viewport->dialogProvider->showButtonDialog(DialogType::Error, "Custom Wing Error", "The first starting wing and the first team-versus-team wing must have the same wing name.",
+		{ DialogButton::Ok });
 		return false;
 	}
 
 	if (!stricmp(_m_starting[0].c_str(), _m_starting[1].c_str()) || !stricmp(_m_starting[0].c_str(), _m_starting[2].c_str())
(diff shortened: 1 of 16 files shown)


qtfred/src/mission/dialogs/CustomWingNamesDialogModel.cpp
```diff
@@ -0,0 +1,119 @@
...
+CustomWingNamesDialogModel::CustomWingNamesDialogModel(QObject * parent, EditorViewport * viewport) :
+	AbstractDialogModel(parent, viewport) {
+	initializeData();
+}
+
+bool CustomWingNamesDialogModel::apply() {
+	int i;
+	if (strcmp(_m_starting[0].c_str(), _m_tvt[0].c_str()))
+	{
+		//MessageBox("The first starting wing and the first team-versus-team wing must have the same wing name.");
```
asarium: You should be able to display this message with the dialog provider of `_editor`.

qtfred/src/mission/dialogs/MissionSpecDialogModel.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+	return true;
+}
+
+void MissionSpecDialogModel::reject() {
+
+}
+
+void MissionSpecDialogModel::setMissionTitle(SCP_string m_mission_title) {
+	this->_m_mission_title = m_mission_title;
```
asarium: This probably lacks a call to `modelChanged()`.

qtfred/src/mission/dialogs/MissionSpecDialogModel.h
```diff
@@ -0,0 +1,138 @@
...
+	void setAIProfileIndex(int m_ai_profile);
+	int getAIProfileIndex() const;
+
+	void setMissionDescText(SCP_string);
+	SCP_string getMissionDescText();
+
+	void setDesignerNoteText(SCP_string);
+	SCP_string getDesignerNoteText();
+
+	void set_modified();
```
asarium: This should be `private` since it's not something that should be done from the outside.

PR #1414, opened 2017-07-13
Add mod table setting for specifying targetted FSO version

### LINKED ISSUE #1389: Feature idea: Add a mod flag for specifying the targeted FSO version
We have quite a lot of modding flags throughout the FSO code the fix some behavior but which cannot be enabled since it might break retail compatibility.

I possible solution I had in mind could be to introduce a mod table flag that lets a mod specify which FSO version it's designed to be run on. That value could then be used for deciding if a bug fix should be applied or not.

That would also allow to implement a clear deprecation model where obsolete flags are disabled if the mod targets a recent enough version of FSO.

What are your thoughts on this?

asarium:
This adds a small system for letting a mod specify which version of FSO
was designed for. That would allow us to introduce code which would
normally break backwards-compatibility by putting it in an if block
which checks if the mod supports the FSO version which ontroduced the
behavioral change.

There is currently no code which uses the new functions yet so I added a
few tests to verify that the options do what they are designed to do.

This fixes #1389.

### DIFF
--- a/code/globalincs/version.cpp
+++ b/code/globalincs/version.cpp
@@ -5,85 +5,130 @@
  * or otherwise commercially exploit the source or things you created based on the 
  * source.
  *
-*/ 
+*/
 
 #include "globalincs/pstypes.h"
 #include "globalincs/version.h"
 
-namespace gameversion
-{
-	bool check_at_least(int major, int minor, int build, int revision)
-	{
-		if (FS_VERSION_MAJOR < major)
-		{
-			return false;
-		}
-		if (FS_VERSION_MAJOR > major)
-		{
-			// Major is greater than the given version => the rest doesn't matter
-			return true;
-		}
-		// major is now equal to our major version
-		
-		if (FS_VERSION_MINOR < minor)
-		{
-			return false;
-		}
-		if (FS_VERSION_MINOR > minor)
-		{
-			// Minor is greater than the given version => the rest doesn't matter
-			return true;
-		}
-		// minor is now equal to our minor version
-		
-		if (FS_VERSION_BUILD < build)
-		{
-			return false;
-		}
-		if (FS_VERSION_BUILD > build)
-		{
-			// build is greater than the given version => the rest doesn't matter
-			return true;
-		}
-		// build is now equal to our build version
-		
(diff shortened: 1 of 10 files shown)


code/globalincs/version.cpp
```diff
@@ -5,85 +5,122 @@
...
-			return true;
-		}
-		
-		// revision is now equal to our revision version
+#include "parse/parselo.h"
+
+namespace gameversion {
+
+bool check_at_least(const version& v) {
+	return get_executable_version() < v;
```
MageKing17: Either this logic is exactly backwards, or you've implemented "greater-than-or-equal" logic into `operator<`, which is extremely confusing.

asarium: `check_at_least` is actually correct but `operator<` is literally the exact opposite of "less-of". I'll fix that and write a few tests to make sure it works correctly.

asarium: Should be fixed now. I added a bunch of tests to make sure it works correctly.

MageKing17:
Nothing else jumps out at me.

Goober5000:
Although this is categorized as a feature, I think it would make sense to merge it into 3.8.0.  It would help for as many versions of FSO as possible to recognize this flag even if no code specifically checks for it yet.

asarium:
The flag is already being recognized as "`$Minimum version:`". I just added an alias for that option to better fit the functionality of the feature.

I think it would be a bad idea to merge this now. We are already at RC3 which may be the build that becomes the final release. Merging this code which touches some existing functionality may break stuff in unforeseen ways and that should not be done in an RC phase, especially not if there were already 3 RCs.

The-E:
Agreed. This is a 3.8.2 feature.

Goober5000:
Well, I would still suggest a separate PR just for the `$Target Version` alias even if none of the other functionality is included.

MageKing17:
To be honest, I'd rather not add the alias at all than slip it into 3.8.

asarium:
The alias made sense to me because "Minimum version" may mean the same thing but "Targeted Version" somehow seemed to be a more intuitive name, at least for me.

PR #1416, opened 2017-07-17
Add true Unicode support

### LINKED ISSUE #1415: Moving to a pure Unicode text setup
A clarification up front: No, the old tables will of course not be deprecated. I plan on enabling Unicode mode with a mod table switch or enabling it by default if a recent enough [FSO version is targetted](https://github.com/scp-fs2open/fs2open.github.com/pull/1414).

This was prompted by the discussion in the [forum thread about special characters](http://www.hard-light.net/forums/index.php?topic=93685.0) in strings. While I fixed the bug reported in the thread I realized that the current mix of TTF and Bitmap glyphs simply doesn't work. The text rendering functions already support rendering UTF-8 encoded Unicode strings since NanoVG supports it but since we need to support the old tables which don't use UTF-8 we can't use that yet.

So, to fix that we would "simply" need to convert all our table text reading to support UTF-8 encoded strings and then pass that to the text rendering functions. Fortunately, most tables and mission files are already properly encoded since the use ASCII encoding. Only the strings and tstrings tables currently use the special FSO text encoding for special characters so modders would only need to convert those files.

Thoughts?

asarium:
This adds proper Unicode support by allowing the parse system to read
UTF-8 encoded files. The new mode is controlled by a mod option which
allows older mods to continue using the legacy text encoding.

That option changes the behavior of several systems in FSO. It disallows
the usage of Volition fonts since they don't support Unicode encoding
and changes a few usages of special characters to their Unicode
counterparts or generic rendering solutions.

This fixes #1415.

Here are the issues that still need to be fixed:
- [x] Fix `split_str` to iterate over codepoints instead of characters
- [x] Fix the briefing functions which use individual characters for color coding
- [x] Implement automatic reencoding of Latin1 encoded files


code/radar/radar.cpp
```diff
@@ -181,10 +181,10 @@ void HudGaugeRadarStd::drawContactCircle( int x, int y, int rad )
 			if (Missiontime & 8192)
 				return;
 		}
-		renderString( Large_blip_offset_x+x, Large_blip_offset_y+y, Large_blip_string );
+		renderCircle(x, y, 6);
 	} else {
 		// rad = RADAR_BLIP_RADIUS_NORMAL;
-		renderString( Small_blip_offset_x+x, Small_blip_offset_y+y, Small_blip_string );
+		renderCircle(x, y, 4);
```
asarium: I made the choice to always use "real" circles since that it what's in the font files. These should look better than the font glyphs in all circumstances and I don't think there is a mod that uses special radar blip glyphs.

If there is a mod that uses special glyphs we can always change it back but this is, at least in my opinion, always an improvement over the current code so I chose to enable it always.

code/parse/parselo.cpp
```diff
@@ -2194,12 +2176,24 @@ void read_raw_file_text(const char *filename, int mode, char *raw_text)
...
+		if (util::has_bom(raw_text)) {
+			// The encoding has to be UTF-8 here so we know that the BOM is 3 Bytes long
+			// This makes sure that the first byte we read will be the first actual text byte.
+			cfseek(mf, 3, SEEK_SET);
+			file_len -= 3;
+		}
+	} else {
+		if (encoding != util::Encoding::ASCII) {
+			//This is probably fatal, so let's abort right here and now.
+			Error(LOCATION, "%s is in Unicode/UTF format and cannot be read by FreeSpace Open. Please convert it to ASCII/ANSI\n", filename);
```
asarium: Good point.

ngld:
The `brief_render_line` and `brief_text_colorize` in `code/mission/missionbriefcommon.cpp` don't handle multi-byte characters. This doesn't obivously break anything but the text animation on the command briefing screen has a few minor issues (I can provide a video if that helps).

The issues is that a multi-byte character like ü consists of [more than one byte](https://en.wikipedia.org/wiki/%C3%9C#Computing_codes). During the animation FSO cuts the line off which produces an invalid UTF-8 sequence if the cut-off happens between the two bytes for ü. NanoVG handles it gracefully by not displaying the invalid character at the end of the line each time this happens. However, this means the animation for a line containing multiple of these characters is behind a line containing only ASCII characters. Also, on a line like "üüüüüüüüüüüüüüüü" the last character is always supposed to be white which only happens every second frame because of this problem.
All in all, a minor issue which most people probably won't notice. I think the fact that the line length isn't calculated properly (each ü would count as two characters) is the bigger issue.

If any other part of the code assumes that a `char` is a valid character, then that part needs to be checked, too.

Finally, is this supposed to complain about retail tables? It would be nice if FSO could convert those from ISO 8859-1 to UTF-8 where necessary.

asarium:
That bug should be easily fixable by iterating over the codepoints instead of individual characters. I already included a library which does that so it should be pretty easy to implement.

> Finally, is this supposed to complain about retail tables? It would be nice if FSO could convert those from ISO 8859-1 to UTF-8 where necessary.

If you enable the option with retail data then it will complain. That's expected since those files are not UTF-8 encoded and I have no intention of adding an automatic conversion feature since the feature is opt-in. Encoding conversions is not something that I would like to implement. There are better tools for doing that.

MageKing17:
> That's expected since those files are not UTF-8 encoded and I have no intention of adding an automatic conversion feature since the feature is opt-in.

Forcing modders to include re-encoded retail data sounds... sub-optimal.

asarium:
If a modder makes the choice to switch to Unicode mode then they should be able to extract two files from the retail files and run a simple `iconv` command on those files. I did exactly that in less than a minute.

If we start to convert retail data, do we also need to convert MediaVPs data? What about Diaspora? My point is that we should not include code that fixes a choice of a modder. We can provide a Unicode starter mod for making the initial transition easier but I am opposed to "magically" converting retail data. It also not easy to actually detect the old ISO 8859-1 because for most files, it's exactly the same as UTF-8.

ngld:
Which tables did you convert? I had to convert three tables: `strings.tbl`, `tstrings.tbl` and `weapons.tbl`.

To automate this you could use `libiconv` (as you pointed out yourself) and [`uchardet`](https://cgit.freedesktop.org/uchardet/uchardet/) can automatically detect the charset a given text file uses. There's probably an even easier way if we only want to detect ISO 8859-1 files ([Here's the code related code from uchardet](https://cgit.freedesktop.org/uchardet/uchardet/tree/src/nsLatin1Prober.cpp)).
I can understand if you don't want to add new dependencies for this but I don't think we should force modders to deal with something that can be easily solved in code.

Given that retail data is copyright protected, what's the legal status of including re-encoded retail tables? It's probably nothing to worry about since noone will care but I'm curious.

asarium:
`weapons.tbl` contained some non UTF-8 characters which I just deleted. They were in a comment anyway. The other two were `strings.tbl` and `tstrings.tbl`.

I understand your point about converting the files on the fly but if that requires additional libraries I will absolutely not implement that. However, the file you linked to is reasonably short and could be included in FSO (the licence should allow that) and SDL has built-in iconv support so we could use that to convert between Latin1 and UTF-8. I will take a look at adding the conversion code if it's really as simply as it seems to be. In any case, FSO should display a warning that the modder should convert the code to UTF-8 manually.

The MediaVPs include modified FS2 retail missions which is not significantly different from including recorded retail tables. As long as we don't distribute the full retail package it should be fine.

MageKing17:
It's just that I can't think of any other mod table options that make FSO fail to load by enabling them, so if we *can* handle it automatically then (in my opinion) we should.

asarium:
I implemented Latin1 reencoding but retail data still causes a parsing error since the retail tables contain an invalid line. The current parsing code detects this invalid line and fixes it on the fly but if the file is UTF-8 encoded that check will not work anymore since it assumes that the line is Latin-1 encoded. In my opinion, it is not feasible to do the Latin1 -> UTF-8 reencoding after the check since that happens while processing the raw text to its final form and I don't want to mess around in that code since I am not very familiar with it.

I will not add a check for the same invalid line but with UTF-8 encoding. Honestly, I don't think that the engine should be fixing such easily fixable issues in user data if it's not directly in retail data.

asarium:
`split_str` should now be fixed.

asarium:
The briefing text rendering functions should also be fixed now.

asarium:
I created a post with test builds based on these changes on the forums: https://www.hard-light.net/forums/index.php?topic=94159.new#new

asarium:
The changes have been tested with a wide variety of languages and use cases and should be ready for merging now.

PR #1418, opened 2017-07-20
Fix bug in normal map rendering

The-E:
This reapplies the fix from https://github.com/scp-fs2open/fs2open.github.com/commit/8e0304a5, which got overwritten at some point.

### DIFF
--- a/code/def_files/main-f.sdr
+++ b/code/def_files/main-f.sdr
@@ -291,7 +291,7 @@ void main()
    // Normal map - convert from DXT5nm
    vec2 normalSample;
    normal.rg = normalSample = (texture(sNormalmap, texCoord).ag * 2.0) - 1.0;
-   normal.b = 1.0;;
+   normal.b = 1.0;
    normal = fragTangentMatrix * normal;
    float norm = length(normal);
    // prevent breaking of normal maps
--- a/code/def_files/main-f.sdr
+++ b/code/def_files/main-f.sdr
@@ -291,7 +291,7 @@ void main()
    // Normal map - convert from DXT5nm
    vec2 normalSample;
    normal.rg = normalSample = (texture(sNormalmap, texCoord).ag * 2.0) - 1.0;
-   normal.b = 1.0;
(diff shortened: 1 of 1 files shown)


MageKing17:
```Download of prebuilt binaries failed: "SSL connect error"!```

...Are you kidding? Argh. So much for prebuilt download failures being a thing of the past.

File: code/def_files/main-f.sdr [hunk 1]
```diff
@@ -291,7 +291,7 @@ void main()
    // Normal map - convert from DXT5nm
    vec2 normalSample;
    normal.rg = normalSample = (texture(sNormalmap, texCoord).ag * 2.0) - 1.0;
-   normal.b = sqrt(1.0 - dot(normal.rg, normal.rg));
+   normal.b = 1.0;;
```
MageKing17: Is that double semicolon intentional?

asarium:
I have no idea why that suddenly stopped working. We could try disabling the TLS verification but that is not really a good idea, especially if it's used to transfer code that gets executed.

code/def_files/main-f.sdr
The-E: It is not.

code/def_files/main-f.sdr
```diff
@@ -291,7 +291,7 @@ void main()
    // Normal map - convert from DXT5nm
    vec2 normalSample;
    normal.rg = normalSample = (texture(sNormalmap, texCoord).ag * 2.0) - 1.0;
-   normal.b = sqrt(1.0 - dot(normal.rg, normal.rg));
```
MageKing17: After conversation with Kobrar and DahBlount on Discord, the old line was the correct behavior and the fix should probably be just to clamp it to avoid values of 0 (clamping to between, say, 0.0001 and 1.0 should be safe enough).

The-E: I've made that change. @ln-zookeeper, could you test this please?

ln-zookeeper: Yes, the clamping fixes the bug as well.

PR #1419, opened 2017-07-20
Fixed $On Ship Arrive hooks not firing on wing arrivals

ln-zookeeper:
The hook wasn't firing correctly for each ship of an arriving wing, apparently ever since my previous fix in 81d18573.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -2409,7 +2409,8 @@ int parse_create_object_sub(p_object *p_objp)
 		}
 	}
 
-	if (Game_mode & GM_IN_MISSION) {
+	// If the ship is in a wing, this will be done in mission_set_wing_arrival_location() instead
+	if (Game_mode & GM_IN_MISSION && shipp->wingnum == -1) {
 		if (anchor_objnum >= 0)
 			Script_system.SetHookObjects(2, "Ship", &Objects[objnum], "Parent", &Objects[anchor_objnum]);
 		else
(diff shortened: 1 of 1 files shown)


code/mission/missionparse.cpp
```diff
@@ -6068,10 +6071,21 @@ void mission_set_wing_arrival_location( wing *wingp, int num_to_set )
...
+
+			if (anchor_objnum >= 0)
+				Script_system.SetHookObjects(2, "Ship", objp, "Parent", &Objects[anchor_objnum]);
+			else
+				Script_system.SetHookObjects(2, "Ship", objp, "Parent", NULL);
+
+			Script_system.RunCondition(CHA_ONSHIPARRIVE, 0, NULL, objp);
+			Script_system.RemHookVars(2, "Ship", "Parent");
+
+			if (wingp->arrival_location != ARRIVE_FROM_DOCK_BAY) {
```
MageKing17: Shouldn't the removed comment from above have been moved here?

ln-zookeeper: I removed it because it was rather pointless; the code is pretty much as clear and self-documenting as it possibly can be WRT the fact that the warp effect is only created "if in mission and not arriving from docking bay". Plus the "in mission" conditional was moved further up, so that part of the comment would have needed updating anyway.

So, I figured that having a comment+code combo like this would be borderline silly:

```
// create warp effect if not arriving from docking bay
if (wingp->arrival_location != ARRIVE_FROM_DOCK_BAY) {
	shipfx_warpin_start(objp);
}
```

MageKing17: Fair enough.

PR #1420, opened 2017-07-22
Restart spring effect after stopping all FF effects

### LINKED ISSUE #1417: Force feedback occasionally stops working
This has been reported numerous times on the forums (e.g. [here](http://www.hard-light.net/forums/index.php?topic=93709.msg1851201#msg1851201)). This is a regression from the SDL2 merge.

Since there is apparently no SCP coder with a force feedback device we can't reproduce and fix this so for now I'll leave this issue without a milestone and would not categorize it as a blocker for the 3.8 release.

asarium:
`SDL_HapticStopAll` stops all effects, including the spring effect so the
joystick goes limp after one mission (`joy_ff_stop_effects` gets called at
the end of each mission).

This restarts the spring effect which should fix that issue.

This fixes #1417.

### DIFF
--- a/code/io/joy_ff-sdl.cpp
+++ b/code/io/joy_ff-sdl.cpp
@@ -52,6 +52,34 @@ static int joy_ff_has_valid_effects();
 static int joy_ff_effect_playing(haptic_effect_t *eff);
 static void joy_ff_start_effect(haptic_effect_t *eff, const char *name);
 
+static void check_and_print_haptic_feature(uint32_t flags, uint32_t check_flag, const char* description) {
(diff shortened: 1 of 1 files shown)


asarium:
I made some changes that might fix this issue properly and give us some more haptic information in the future:

1. I ported the old code of `joy_ff_stop_effects` to the SDL platform which should match what the old code did a lot better than the current code.
2. I added additional logging to determine the features of the used haptic device.

asarium:
The newest update fixes the bug properly. This was confirmed by Lykurgos88 [in the forum thread](http://www.hard-light.net/forums/index.php?topic=93734.msg1851997#msg1851997).

Issue #470, opened 2015-11-25
Unix standalone isn't handling disconnects properly, leads to server crash


chief1983:
Attempting to write to a closed socket on *nix leads to a SIGPIPE.  Apparently this isn't an issue on Windows, but this causes the server to halt on Linux.

GDB output and backtrace:

```
scpuser@fileserver /usr/share/games/freespace2 $ gdb fs2_open_3.7.3_20150716_DEBUG
GNU gdb (Gentoo 7.9.1 vanilla) 7.9.1
Copyright (C) 2015 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "i686-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://bugs.gentoo.org/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from fs2_open_3.7.3_20150716_DEBUG...done.
(gdb) r -standalone
Starting program: /usr/share/games/freespace2/fs2_open_3.7.3_20150716_DEBUG -standalone
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/libthread_db.so.1".
outwnd not initialized yet...
Standalone running
Future debug output directed to: /home/scpuser/.fs2_open/data/fs2_standalone.log
[New Thread 0xae66cb40 (LWP 30282)]
[New Thread 0xaeff0b40 (LWP 30281)]
[New Thread 0xaf974b40 (LWP 30280)]
[New Thread 0xb02f8b40 (LWP 30279)]
[New Thread 0xb0c7cb40 (LWP 30278)]

Program received signal SIGPIPE, Broken pipe.
0xb7fdc424 in __kernel_vsyscall ()
(gdb) bt
#0  0xb7fdc424 in __kernel_vsyscall ()
#1  0xb77b6e28 in sendto () from /lib/libpthread.so.0
#2  0x081f8695 in FS2NetD_SendData (buffer=0xbfffe590 "\004\036", blen=30) at fs2netd/tcp_socket.cpp:242
[...]

PR #1423, opened 2017-07-25
Fix for #470: Proposal 1

chief1983:
PLEASE DO NOT MERGE WITHOUT A DECISION BETWEEN PROPOSAL 1 OR 2

This is my first attempt at preventing the SIGPIPE standalone crash on all supported *nix platforms.  I recently discovered that jg18's proposal did not work on Solaris based platforms as they don't currently support MSG_NOSIGNAL or SO_NOSIGPIPE.  After a lot of searching I came up with a couple of proposals.  This one is mostly copied from http://www.microhowto.info/howto/ignore_sigpipe_without_affecting_other_threads_in_a_process.html and revolves around using pthread_sigmask to mask the SIGPIPE signal temporarily, and then restore any previous behavior around it.  Other proposals involving sigaction() I decided against because of thread concerns.  It may not be an issue today, but the solutions weren't any simpler so I figured I would just stick with a thread safe example rather than globally adjust SIGPIPE behavior.

### DIFF
--- a/code/fs2netd/tcp_socket.cpp
+++ b/code/fs2netd/tcp_socket.cpp
@@ -281,7 +281,13 @@ int FS2NetD_SendData(char *buffer, int blen)
 
 #if defined(SCP_UNIX) && !defined(SO_NOSIGPIPE) && !defined(MSG_NOSIGNAL) && _POSIX_REALTIME_SIGNALS > 0
 		// Step 3: Accept any pending SIGPIPE using sigtimedwait.
-		struct timespec zerotime = {0};
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
+		// Some older GCC versions seem to incorrectly warn on this initialization.
+		// The pragmas should be removable once GCC 5+ is in use everywhere.
+		// Reference: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61489
+		struct timespec zerotime = {};
+#pragma GCC diagnostic pop
 		if (sigtimedwait(&sigpipe_mask, 0, &zerotime) == -1) {
 			ml_printf("Could not accept possible pending SIGPIPE with sigtimedwait in FS2NetD_SendData");
 			exit(1);
--- a/code/fs2netd/tcp_socket.cpp
+++ b/code/fs2netd/tcp_socket.cpp
@@ -27,6 +27,7 @@
 #include <unistd.h>
 #include <cerrno>
 #include <sys/ioctl.h>
+#include <signal.h>
 #ifdef SCP_SOLARIS
 #include <sys/filio.h>
 #endif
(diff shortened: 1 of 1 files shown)


chief1983:
As currently written, this one exits immediatel after trying to start the standalone on Solaris, so proposal 2 is currently looking like a better starting point at least.

code/fs2netd/tcp_socket.cpp
```diff
@@ -237,9 +250,57 @@ int FS2NetD_SendData(char *buffer, int blen)
...
+
+		// Step 2: Execute the library code which might raise SIGPIPE.
+		auto sent_chars = sendto(mySocket, buffer, blen, flags, (sockaddr*)&FS2NetD_addr, to_len);
+
+#if defined(SCP_UNIX) && !defined(SO_NOSIGPIPE) && !defined(MSG_NOSIGNAL) && _POSIX_REALTIME_SIGNALS > 0
+		// Step 3: Accept any pending SIGPIPE using sigtimedwait.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
+		// Some older GCC versions seem to incorrectly warn on this initialization.
+		// The pragmas should be removable once GCC 5+ is in use everywhere.
```
asarium: At least our CI setup is using GCC >= 5 so this is probably not needed anymore.

chief1983: The platform I noticed doesn't support either of jg18's original solutions, Solaris 11.3, is still at 4.8, so for the time being I would hope to leave it.  Maybe Solaris.next will be 5+ hopefully.  I think OpenIndiana should already be high enough, hope to test that soon too when I get it rebooted into an updated boot environment.

asarium: Ok, I wasn't sure if we still supported GCC 4.

chief1983: Well, it's 4.8.2, which has pretty good C++11 support I think so it's been working so far.  But as it lacks sanitizers and has bugs like this I'm hoping to see the platform move to 5+ soon.
https://gcc.gnu.org/gcc-4.8/cxx0x_status.html
Also, CentOS and openSUSE's latest stable versions are both 4.8.5 and the standalones have been running fine on those platforms so far.

asarium: Hmm, C++11 support seems to be better than I thought with those compilers.

asarium:
@chief1983 Did you fix the issues with this version and do you have an idea what proposal works better? IMO, it would be good to merge one of the proposals before 3.8 since a crashing standalone server is pretty bad.

chief1983:
Well, the standalones haven't crashed yet, but since they're still in the list I don't think the SIGPIPE event has happened yet.  Currently the fix prevents the actual crash, but they still usually disappear from FS2NetD because they fail to restart communications with FS2NetD correctly.  So I'm keeping an eye on them until that changes.  Haven't found a reliable way to trigger the particular hiccup yet.

PR #1424, opened 2017-07-25
Fix for #470: Proposal 2

chief1983:
PLEASE DO NOT MERGE WITHOUT A DECISION BETWEEN PROPOSAL 1 OR 2

This is my second attempt at preventing the SIGPIPE standalone crash on all supported *nix platforms.  I recently discovered that jg18's proposal did not work on Solaris based platforms as they don't currently support MSG_NOSIGNAL or SO_NOSIGPIPE.  After a lot of searching I came up with a couple of proposals.  This one is mostly copied from https://github.com/kroki/XProbes/blob/1447f3d93b6dbf273919af15e59f35cca58fcc23/src/libxprobes.c#L156 and revolves around managing the SIGPIPE restoration differently than the first proposal.  Other proposals involving sigaction() I decided against because of thread concerns.  It may not be an issue today, but the solutions weren't any simpler so I figured I would just stick with a thread safe example rather than globally adjust SIGPIPE behavior.

### DIFF
--- a/code/fs2netd/tcp_socket.cpp
+++ b/code/fs2netd/tcp_socket.cpp
@@ -273,6 +273,7 @@ int FS2NetD_SendData(char *buffer, int blen)
 		is per-process).
 	*/
 	sigset_t pending;
+	sigset_t sigpipe_mask;
 	bool sigpipe_pending;
 	bool sigpipe_unblock;
 	sigemptyset(&pending);
@@ -281,7 +282,7 @@ int FS2NetD_SendData(char *buffer, int blen)
 	if (!sigpipe_pending) {
 		sigset_t blocked;
 		sigemptyset(&blocked);
-		pthread_sigmask(SIG_BLOCK, &config.sigpipe_mask, &blocked);
+		pthread_sigmask(SIG_BLOCK, &sigpipe_mask, &blocked);
 
 		/* Maybe is was blocked already?  */
 		sigpipe_unblock = ! sigismember(&blocked, SIGPIPE);
@@ -298,7 +299,6 @@ int FS2NetD_SendData(char *buffer, int blen)
 		were us who blocked it.
 	*/
 	if (!sigpipe_pending) {
-		sigset_t pending;
 		sigemptyset(&pending);
 		sigpending(&pending);
 		if (sigismember(&pending, SIGPIPE)) {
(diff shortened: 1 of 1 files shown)


chief1983:
Since proposal 1 crashed, I will leave a build based on this one running on a standalone for a while to see if anything odd happens.

asarium:
Is there any update on what solution is the most stable?

chief1983:
Looking into it now.  Noticing some generally odd behavior.  A couple of my servers I can't seem to SSH into right now, or reboot even via their web consoles.  The ones I can, I noticed that they aren't writing any FS2NetD data to the standalone log, even though I could swear I confirmed they connected to FS2NetD the last time I started them.  But they're no longer showing up even though they are still running.

My Solaris 11 box (the main reason I tweaked this solution in the first place) still suffered a SIGPIPE with fix 1.  So, whatever I did, Fix 1 seems to fail at actually preventing the problem at least on that platform.  So I'm leaning towards this one, I will try to see what happened with the unresponsive servers though, and the missing logging.

chief1983:
Ok nevermind that I was looking in the wrong log for the FS2NetD chatter.  Needed to check multi.log.  So, as far as I can tell, Fix 2 has been working pretty good.  Only thing I'm not sure is why I have an Ubuntu and an OpenSUSE box that were both running it and went unresponsive, but I suspect that's an issue with my VPS provider and not the code.  I'm going to run Fix 2 in more locations.  None of the locations I was running it on at my house have had any strange issues, including the fact that it actually seemed to work correctly on OpenIndiana.

asarium:
If you want to do a few more tests then I'll wait for them but if everything works as expected then I will merge this changeset (Proposal 2).

chief1983:
#2 is looking better now.  I was concerned because a couple servers had gone un-responsive.  But one of them was running #1, not #2.  And both actually seem to have been issues with the VPS, either transient or that were cleared up by the support teams.  Still, #1 was a cleaner looking fix and I may want to look into figuring out why it didn't seem to work on the platform it was supposed to fix, but #2 is looking like the way to go right now.  I've got more servers running it this time, over half the ones listed.  If none of them crash in seemingly related ways in the near future I think this is the way to go.

asarium:
Alright. Once you're sure that this works, this can be merged.

chief1983:
Rebased against master, I have seen no sign of issues related to this patch anywhere on my running standalones.

Issue #1425, opened 2017-07-25
old no-collide ship flag not recognized


Goober5000:
It looks like the `no-collide` ship flag has been changed to `no_collide`.  Old missions saved with the `no-collide` flag will have this flag removed when you open them and save them in FRED.

(Since this flag is cleared when the mission is opened in FRED, I am also assuming that old `no-collide` flags will not be recognized if they are played in FSO, but I have not tested that aspect.)

asarium:
Fixed by #1426.

PR #775, opened 2016-08-08
Bitset flags -- take 2

The-E:
This is a redo of the work I did for bitset flags 2 years ago. 

To recap: This introduces a new mechanism for handling bit flags. Instead of using an int (or two, or three) to write bits into, this uses the std::bitset in a wrapper class that accepts an enum class as input. This means that certain classes of errors we've had a few of in the engine, where a coder used the wrong set of flags or checked them against the wrong variable (SF_STEALTH vs SIF_STEALTH, for example) can now be detected at a compiler level rather than not at all.

---
This change is [Reviewable](https://reviewable.io/reviews/scp-fs2open/fs2open.github.com/775)

PR #1426, opened 2017-07-25
for backwards compatibility, check both no-collide and no_collide

Goober5000:
In the bitset flags PR #775, the `no-collide` ship flag was mistakenly changed to `no_collide`.  (This was probably because the equivalent ships.tbl flag was `no_collide`.)  As a result, this flag is not recognized in missions created with older versions of FRED, as noted in #1425.

I added code to check both variants of flag in both ship parsing and mission parsing.  This should not only fix the regression bug but also guard against typos in ships.tbl.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -3036,7 +3036,13 @@ int parse_object(mission *pm, int flag, p_object *p_objp)
         parse_string_flag_list(p_objp->flags, Parse_object_flags, num_parse_object_flags, &unparsed);
         if (!unparsed.empty()) {
             for (size_t k = 0; k < unparsed.size(); ++k) {
-                WarningEx(LOCATION, "Unknown flag in parse object flags: %s", unparsed[k].c_str());
+				// catch typos or deprecations
+				if (!stricmp(unparsed[k].c_str(), "no-collide") || !stricmp(unparsed[k].c_str(), "no_collide")) {
+					p_objp->flags.set(Mission::Parse_Object_Flags::OF_No_collide);
+				}
+				else {
+					WarningEx(LOCATION, "Unknown flag in parse object flags: %s", unparsed[k].c_str());
+				}
             }
         }
     }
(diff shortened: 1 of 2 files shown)


code/mission/missionparse.cpp
```diff
@@ -3036,7 +3036,13 @@ int parse_object(mission *pm, int flag, p_object *p_objp)
         parse_string_flag_list(p_objp->flags, Parse_object_flags, num_parse_object_flags, &unparsed);
         if (!unparsed.empty()) {
             for (size_t k = 0; k < unparsed.size(); ++k) {
-                WarningEx(LOCATION, "Unknown flag in parse object flags: %s", unparsed[k].c_str());
+				// catch typos or deprecations
+				if (!stricmp(unparsed[k].c_str(), "no-collide") || !stricmp(unparsed[k].c_str(), "no_collide")) {
+					p_objp->flags.set(Mission::Parse_Object_Flags::OF_No_collide);
```
asarium: If this spelling of the flag is deprecated, would it make sense to show a `Warning` here?

code/ship/ship.cpp
```diff
@@ -3430,9 +3430,17 @@ int parse_ship_values(ship_info* sip, const bool is_template, const bool first_t
...
 						sip->flags.set(Ship_flags[idx].def);
+
+					break;
 				}
 			}
 
+			// catch typos or deprecations
+			if (!stricmp(ship_strings[i], "no-collide") || !stricmp(ship_strings[i], "no_collide")) {
+				flag_found = true;
+				sip->flags.set(Ship::Info_Flags::No_collide);
```
asarium: Same as above.

code/mission/missionparse.cpp
Goober5000: I didn't consider it worth a Warning because I never actually wanted to deprecate the `no-collide` term.  In fact, I would prefer to put it back to that variant because every other ship flag uses hyphens.

I decided to leave it as the underscore in order to minimize the chance of any typos happening again, but IMO it is the underscore that should be deprecated, not the hyphen.

asarium: Ok. If there is no good reason to deprecate either variant then this should be fine.

PR #1429, opened 2017-08-02
Lower Afterburner force feedback effect strength to previous levels

asarium:
This was suggested by taylor on the forums and should match the previous
values more closely.

### DIFF
--- a/code/io/joy_ff-sdl.cpp
+++ b/code/io/joy_ff-sdl.cpp
@@ -326,7 +326,7 @@ static int joy_ff_create_effects()
 	pAfterburn1.eff.periodic.direction.dir[0] = 0;
 	pAfterburn1.eff.periodic.length = SDL_HAPTIC_INFINITY;
 	pAfterburn1.eff.periodic.period = 20;
-	pAfterburn1.eff.periodic.magnitude = 0x6665;
+	pAfterburn1.eff.periodic.magnitude = 0x3332;
 
 	pAfterburn1.id = SDL_HapticNewEffect(haptic, &pAfterburn1.eff);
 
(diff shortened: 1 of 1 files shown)


code/io/joy_ff-sdl.cpp
```diff
@@ -727,7 +727,7 @@ void joy_ff_afterburn_on()
 		SDL_HapticStopEffect(haptic, pAfterburn2.id);
 
 		pAfterburn2.eff.periodic.length = SDL_HAPTIC_INFINITY;
-		pAfterburn2.eff.periodic.magnitude = 0x3fff;
+		pAfterburn2.eff.periodic.magnitude = 0x1999;
```
MageKing17: Uh... why are the length and magnitude being set again at all?

asarium: The effect is reused for the fly-by effect which uses a different magnitude so whenever the actual afterburner effect should be played the magnitude needs to be reset in case the fly-by effect was played before.

MageKing17: Is there some reason the effects can't be separated?

asarium: It was probably done to reduce the amount of used effects. In any case, I'm not going to poke around in code that I can test and it would be out of scope for this PR anyway.

asarium: The original code uses `INFINITE` for both effects while initializing so it was probably a mistake made while porting the code.

MageKing17:
taylor pointed out another effect with the wrong magnitude: http://www.hard-light.net/forums/index.php?topic=93734.msg1852296#msg1852296

May as well fix it in the same PR.

asarium:
Lykurgos88 confirmed on the forums that these changes fixed the effect strengths so this should be ready to be merged now.

PR #1435, opened 2017-08-10
Add config option for setting the FF strength

asarium:
This adds an `fs2_open.ini` option for specifying the global scale of the effect strengths. This uses the global gain feature of SDL2 so if that is not supported then the config option will have no effect and a release warning will be displayed.


asarium:
The changes have been tested and confirmed to be working in the [forum thread](http://www.hard-light.net/forums/index.php?topic=93734.msg1853309#msg1853309).

PR #922, opened 2016-08-31
Convert the remaining AI flags

The-E:

PR #1436, opened 2017-08-11
fix typo in AI flag conversion

Goober5000:
I noticed that there was one line in #922 where two flags were checked using the comma operator.  This is incorrect as the comma operator evaluates both expressions but only returns the value of the rightmost expression, resulting in only one flag being checked.  This PR fixes the Assert by copying the same condition as another Assert elsewhere in the file.

This PR is actually a no-op because the problematic code is commented out, but one never knows when it might be used again.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -3814,7 +3814,7 @@ float ai_path_0()
 		int path_num;
 		Assert(aip->active_goal >= 0);
 		ai_goal *aigp = &aip->goals[aip->active_goal];
-		Assert(aigp->flags[AI::Goal_Flags::Dockee_index_valid] || aigp->flags[AI::Goal_Flags::Docker_index_valid]);
+		Assert(aigp->flags[AI::Goal_Flags::Dockee_index_valid] && aigp->flags[AI::Goal_Flags::Docker_index_valid]);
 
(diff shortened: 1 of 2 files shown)


Goober5000:
Addendum: Given that in all places the Asserts are found, the two indexes should both be valid, I changed the `||` to `&&`.

asarium:
FYI, the comma operator for bitset flags use a bit of template magic to change that operator to a logical `|` which is noted in [the documentation](https://github.com/scp-fs2open/fs2open.github.com/wiki/Bitset-Flags). Since you changed the places where `||` or `,` was used to `&&` it doesn't make a difference but I wanted to let you know that the `,` operator actually makes sense with flagsets.

Goober5000:
> FYI, the comma operator for bitset flags use a bit of template magic to change that operator to a logical `|` which is noted in [the documentation](https://github.com/scp-fs2open/fs2open.github.com/wiki/Bitset-Flags).

Hmm.  That's interesting, but I'm not sure I like that.  The comma operator has a well-defined function, and changing it like that seems like a recipe for misunderstanding (like what happened here).

asarium:
It was discussed when I added the operator and we decided against using the `|` operator for some reason I can't remember. Anyway, that has nothing to do with these changes and I approved them already so I'm just going to merge these.

PR #1442, opened 2017-08-22
Remove stack trace from warning dialogs

asarium:
It serves no purpose there since it's a message specifically designed to
be seen by modders and most times the stack trace doesn't even contain
useful information since there are no debug symbols present.

### DIFF
--- a/code/osapi/dialogs.cpp
+++ b/code/osapi/dialogs.cpp
@@ -384,10 +384,9 @@ namespace os
 			boxMsgStream << "File: " << filename << "\n";
 			boxMsgStream << "Line: " << line << "\n";
 
-			boxMsgStream << "\n";
-			boxMsgStream << dump_stacktrace();
-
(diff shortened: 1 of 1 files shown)


MageKing17:
Would much rather gate it behind e.g. `#ifdef STACKTRACE` or something, because it's plenty useful for coders or test builds that *do* have .pdb files. Also, doesn't the stacktrace code work on non-Windows systems?

asarium:
My reason for removing it entirely from warnings is that it simply doesn't serve a purpose there. A stacktrace for assertions is nice but warnings are meant to be used in situations where the modder is responsible for them. Most times, the file and line number provide sufficient information for finding where the warning originated from.

MageKing17:
I suppose this can be revisited if there's ever a situation where a stacktrace in a warning dialogue would actually have been useful, rather than just leaving it in "just in case"...

PR #1448, opened 2017-08-25
Add GLM library to project and convert 4D vector math to use it

asarium:
The library is commonly used in OpenGL applications and is specifically
designed to integrate well into such an engine. It supports all features
FSO needs and even more for future usage. It is also much easier to use
than the existing 4D vector math functions.

This is required before submitting my changes for implementing support
for uniform blocks (#910).


lib/glm/CMakeLists.txt
```diff
@@ -6,4 +6,4 @@ set_target_properties(glm PROPERTIES FOLDER "3rdparty")
 
 # Disable warnings in this library
 suppress_warnings(glm)
-target_compile_definitions(glm PUBLIC GLM_FORCE_CXX11 GLM_FORCE_RADIANS)
\ No newline at end of file
+target_compile_definitions(glm PUBLIC GLM_FORCE_RADIANS)
```
chief1983: Did we create this CMakeLists.txt or was it provided by them?  If it was part of their bundle, is some other place to override it outside of their provided code?  If we made it, ignore this :)

asarium: That's a file I wrote. Normally you're supposed to define that before every usage of a GLM header but a global define is a better option here.

SamuelCho:
I'm not going to lie and say I'm unflinchingly sad that this is replacing all the vec4 and matrix4 work I worked on and planned to continue extending. What specifically did you need that you brought in the entire glm library to replace my work?

asarium:
I had this in my repo for quite some time now but when I did the model uniform block stuff I recognized that the current vec4 types are needlessly hard to work with. I even found a comment that says that the code could be replaced in the future:
https://github.com/scp-fs2open/fs2open.github.com/blob/bd62ceeeb0d1b64d8b8a066519e40680e8b8a5f2/code/math/vecmat.cpp#L2538

Also, I'm not replacing your work. I'm just using an existing library to make common tasks easier. There is no reason to write the code to create a standard projection matrix yourself if an external library already has that code and also tested that it works correctly.

SamuelCho:
I didn't leave that comment so that shouldn't be taken as an expression of what I feel or what I felt at the time I wrote the vec4 and matrix4 code. 

And yes, you're clearly replacing parts of my code with glm. I mean, for example, are those deletions in vecmat.cpp not considered a replacement?

So for the record, still sad. Sad sad sad. :(

TRBlount:
I have worries about the SIMD extensibility of GLM. The last time I tried to get GLM to activate SIMD intrinsics (this summer) I had to use several #define values that limites compilation to only SSE2. Additionally, I had to modify code within the library to get several SSE ops to even get used. 
Even when I got them to trigger, the amount of SSE ops used was minimal at best, basically offering no improvements over the old code and generally being very sketchy to setup.

asarium:
@SamuelCho It replaces a few matrix and vector operations that can be found in other places all around the internet so those functions are nothing special (please correct me if I overlooked something there). I don't understand why you would be so attached to code that (at least as far as I can tell) can be easily replaced by an external library.

@TRBlount 1. Why would we need to explicitly write code that uses something newer than SSE2? SSE2 already supports 128-bit SIMD operations which should be enough for our needs which are limited to 4 single precision component vectors. I cannot say anything about how glm handles enabling SIMD operations in practice but at least in their manual it says that it's automatically enabled if support is detected. I'll take a look at how that is handled when I have the time.

TRBlount:
So you're going to drop support for AVX and AVX2 just like that? I'd rather have more options for testing than a library that handles everything for us at a rather significant loss.

MageKing17:
Having manual SSE2 operations should do nothing to stop us from having automatic AVX/AVX2 optimizations, same as now.

asarium:
Exactly, just because GLM only supports SSE2 (since only that makes sense for it) doesn't mean that we can't enable automatic AVX optimizations.

SamuelCho:
@asarium maybe you should just accept that I'm sad that I worked hard on this code, getting everything sorted out, and planned future improvements with it. And now I'm unable to. It just sounds like you're basically saying I have no right to have these feelings and I should just act like some lifeless droid that accepts everything so long as the efficiency criteria is checkmarked. I'm human and I have feelings. Wow.

I'll approve this code review. But please understand that this attitude of stomping over my code, without any remorse or attempt to reach out to me, like what you did with the profiler, is causing me to question my involvement with this project.

asarium:
I'm sorry that I offended you with these changes or my comments. I simply saw some code that could be improved and then implemented the improvements.

I wasn't aware that you were planning future improvements with the vec4 code but if it helps you, I want to let you know that there are no substantial changes required to code which uses the "old" vec4 types and even additions to the actual vector math library should be easy to port to glm since the types are very compatible with each other.

asarium:
I reworked the model uniform code to use the current `vec4` and `matrix4` so this isn't needed anymore. There may be some future application for this so I will leave the branch in my repo in case anyone needs it in the future.

PR #1450, opened 2017-08-26
toss out the campaign savefile if we are restarting a previously finished campaign

Goober5000:
When switching to a previously completed campaign, the game will currently attempt to make further progress due to the existing campaign savefile, which is obviously not possible.  This PR deletes the old campaign savefile (which is retail behavior) and sets the next mission to the first in the campaign.  This way, the player can start the campaign afresh.  Pilot stats are not modified.

This PR fixes Mantis #3174.

### DIFF
--- a/code/menuui/readyroom.cpp
+++ b/code/menuui/readyroom.cpp
@@ -1546,16 +1546,8 @@ void campaign_room_commit()
 		return;
 	}
 
-	if (stricmp(Campaign_file_names[Selected_campaign_index], Campaign.filename)) {  // new campaign selected
-		/* This is all that's stopping us from switching campaigns without restarting - taylor
-		if ((Active_campaign_index >= 0) && campaign_room_reset_campaign(Active_campaign_index)) {
-			gamesnd_play_iface(SND_GENERAL_FAIL);
-			return;
-		}
-
-		mission_campaign_savefile_delete(Campaign_file_names[Selected_campaign_index]);
-		*/
-
+	// new campaign selected?
+	if (stricmp(Campaign_file_names[Selected_campaign_index], Campaign.filename)) {
 		// Goober5000 - reinitialize tech database if needed
 		if ( (Campaign.flags & CF_CUSTOM_TECH_DATABASE) || !stricmp(Campaign.filename, "freespace2") )
 		{
--- a/code/menuui/readyroom.cpp
+++ b/code/menuui/readyroom.cpp
@@ -1563,8 +1563,20 @@ void campaign_room_commit()
 			tech_reset_to_default();
 		}
 
-		mission_campaign_load(Campaign_file_names[Selected_campaign_index]);
-		strcpy_s(Player->current_campaign, Campaign.filename);  // track new campaign for player
+		int load_status = mission_campaign_load(Campaign_file_names[Selected_campaign_index]);
+
+		if (load_status == 0) {
+			strcpy_s(Player->current_campaign, Campaign.filename);  // track new campaign for player
+
+			// sanity check: if we just loaded a savefile, but we have no next mission,
+			// then we are switching back to an old campaign that we previously completed,
+			// and we want to clear it to start afresh
+			if (Campaign.next_mission == -1) {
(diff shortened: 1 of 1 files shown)


asarium:
If I understand the changes and your comments correctly then whenever a user selects a finished campaign it will restart it? If that is true it seems a bit counterintuitive since we already have the "Restart campaign" button. If it's retail behavior then it does make sense to make this change but I would like to make sure I understand the changes before approving the code review.

Goober5000:
You understand correctly.  Here's the quote from the relevant Mantis ticket:

> If you play through a campaign until it's complete, then switch to a new campaign (and click OK), then try to switch back to the completed campaign (and click OK again), the button will play the error beep and FSO does not leave the campaign room.
> 
> Despite this, the asterisk indicator does move to the newly selected campaign, and if you exit the campaign room by pressing Escape, that campaign is now the active one. But when you try to go to the Ready Room, FSO will tell you that the campaign is over.
> 
> To actually be able to play the campaign, you need to go back to the campaign room and click Reset Campaign. The behavior should be that when you switch to a new campaign that has been completed, the campaign should start over from the beginning.

I suppose another way to fix it would be to play the success sound instead of the error beep, and ask the user whether he wants to restart the campaign.  But I think this fix is the way to go, first because it's retail behavior, and second because there's very little reason to switch to a campaign in the completed state.  In fact the only reason I can think of is to view the existing missions and cutscenes, but you can already do that with the Shift-S trick.

The "restart campaign" button is mainly used for going back to the beginning of a campaign that's currently in progress.

asarium:
Ok. I just wanted to confirm that my understanding was correct. If it was retail behavior then these changes should be good to go. The alternative solution (showing a dialog) can be implemented separately.

chief1983:
Not sure this wasn't premature.  If anything, I feel like tossing campaign stats in retail may have been a bug that the current behavior was supposed to fix.  I am pretty sure we have a lot of players who like having long-running pilots with completed campaigns.  I think the retail behavior probably only existed because multiple campaign states could not be tracked by the retail pilot code?  But now we can track multiple campaign states easily within one pilot, so tossing the data doesn't necessarily make sense at all now.  It is counter-intuitive to me at least that simply switching campaigns will eliminate my progress.  New users may not know about the 'cheats' to view all missions, cutscenes, etc.  We do still get a lot of those from GOG and Steam now.  Also, aren't medals tied to your campaign progress as well?  I imagine that would be getting tossed out with this then?  I believe some users would prefer not to lose those either.  It would not have been a big deal in retail because it was known you needed a pilot per campaign to save progress.  Now, that we have multi-campaign support within a pilot, dropping progress on change seems very confusing to me.  If I misunderstood anything in here please let me know, but I think a non-dropping by default solution needs to be implemented for this soon, or perhaps reverted until a long term solution is devised.

Goober5000:
You've got the situation backwards.  This fix *only* tosses the campaign savefile.  It does not modify the pilot stats (as I said in the first post) so the number of kills, weapon shots, points, and so forth are not affected.  Furthermore, without this PR, the only way to replay a campaign would be for the player to hit the "Reset Campaign" button, which tosses the savefile anyway.

Ironically, pilot all-time stats are indeed zeroed out when switching campaigns, and they aren't supposed to be.  My PR #1451 is an attempt to fix this, but I haven't quite got the full solution yet.

chief1983:
Ok, if stats are global to the pilot and not a campaign, that partially helps.  Still, what about medals?  Those are per campaign right?  Or is that per mod/TC?  Also, it still seems intuitive to me that the Reset Campaign button would be _the_ way to reset your campaign, and not that it would 'just happen' when changing campaigns.

Goober5000:
No, all-time medals are global to the pilot just like all-time ship kills are.  There is an `index_list_t` type that handles any table index mismatch.

I guess "intuitive" is a matter of opinion; it seems intuitive to me that switching to a completed campaign would automatically reset it.  Requiring the player to manually reset it is an extra step.

Again, if you can come up with any reason to keep the campaign savefile in the "this campaign is over" state, do share that, because I can't think of one.

PR #1458, opened 2017-08-27
several fixes to time-docked and time-undocked

Goober5000:
The `time-docked` and `time-undocked` sexps will cause FSO to freeze unless you know what you're doing, so this PR fixes some gotcha areas.  Inspired by The E's rant on Discord which, while not correct about `OPF_POSITIVE`, did indicate some areas for improvement.

1) Set the default "number of times docked" argument to 1 when the sexp is created in FRED, same as with the `has-[un]docked-delay` sexps
2) Catch any <= 0 counts when the sexp is called in FRED and set them to 1.  (This was done for other sexps which used `mission_log_get_time_indexed`, but was not done for `time-[un]docked` for some reason.)
3) Add an Assertion to `mission_log_get_time_indexed` as a last resort to guard against invalid counts.

Also, I merged `sexp_time_docked` and `sexp_time_undocked` since the functions were identical except for one value on one line.

### DIFF
--- a/code/mission/missionlog.cpp
+++ b/code/mission/missionlog.cpp
@@ -431,6 +431,7 @@ int mission_log_get_time_indexed( int type, const char *pname, const char *sname
 {
 	int i, found;
 	log_entry *entry;
+	Assertion(count > 0, "The count parameter is %d; it should be greater than 0!", count);
 
 	entry = &log_entries[0];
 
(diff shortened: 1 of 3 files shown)


MageKing17:
AppVeyor failure was a prebuilt binary failure, so I restarted it.

PR #1459, opened 2017-08-27
add string-concatenate-block sexp

Goober5000:
Add `string-concatenate-block` and deprecate `string-concatenate`.  The new sexp is a better design since the destination variable appears first in the argument list and the sexp allows an unlimited number of concatenations (subject to length limits).  The old sexp was cumbersome for doing more than one concatenation at a time.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -20291,7 +20291,7 @@ void sexp_string_concatenate(int n)
 	// check length
 	if (strlen(new_text) >= TOKEN_LENGTH)
 	{
-		Warning(LOCATION, "Concatenated string '%s' has %d characters, but the maximum is %d.  The string will be truncated.", new_text, strlen(new_text), TOKEN_LENGTH - 1);
+		Warning(LOCATION, "Concatenated string '%s' has " SIZE_T_ARG " characters, but the maximum is " SIZE_T_ARG ".  The string will be truncated.", new_text, strlen(new_text), TOKEN_LENGTH - 1);
 		new_text[TOKEN_LENGTH] = 0;
 	}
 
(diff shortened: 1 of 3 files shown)


code/parse/sexp.cpp
```diff
@@ -20298,6 +20300,49 @@ void sexp_string_concatenate(int n)
...
+	while (n >= 0)
+	{
+		new_text.append(CTEXT(n));
+		n = CDR(n);
+	}
+
+	// check length
+	if (new_text.length() >= TOKEN_LENGTH)
+	{
+		Warning(LOCATION, "Concatenated string '%s' is too long and will be truncated.", new_text.c_str());
```
asarium: This warning could include the maximum size of the string. Otherwise the user would need to manually count the characters in the truncated string.

The documentation of this SEXP already contains this information but it would also be useful to see that here.

code/parse/sexp.cpp
```diff
@@ -20289,7 +20291,7 @@ void sexp_string_concatenate(int n)
 	// check length
 	if (strlen(new_text) >= TOKEN_LENGTH)
 	{
-		Warning(LOCATION, "Concatenated string is too long and will be truncated.");
+		Warning(LOCATION, "Concatenated string '%s' is too long and will be truncated.", new_text);
```
MageKing17: Along the same lines as @asarium pointed out below, having this warning display the actual length would be helpful.

Goober5000:
Okay, warning messages edited.

Goober5000:
Good catch, thanks.

MageKing17:
Just for the length; `TOKEN_LENGTH`, being a macro for a numeric literal, causes a warning on Linux (upgraded to an error, hence the Travis failure).

Goober5000:
Well, phooey.  Fixed, hopefully.

PR #1461, opened 2017-08-28
make "same emp time for capships" actually work

Goober5000:
Back when I added the "same emp time for capships" flag in SVN r9540, I neglected to multiply the EMP time by 1000, as the EMP disruption time for fighters is specified in seconds but the time for capships is specified in milliseconds.  So, using this flag would cause capships to suffer EMP disruption for almost no time at all.

This PR fixes that, and also extends the explosion-effect sexp to specify this parameter.  And I also cleaned up the fireball section by calling eval_num(n) just once instead of several times.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -664,7 +664,7 @@ sexp_oper Operators[] = {
 	{ "ship-vaporize",					OP_SHIP_VAPORIZE,						1,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// Goober5000
 	{ "ship-no-vaporize",				OP_SHIP_NO_VAPORIZE,					1,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// Goober5000
 	{ "set-explosion-option",			OP_SET_EXPLOSION_OPTION,				3,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// Goober5000
-	{ "explosion-effect",				OP_EXPLOSION_EFFECT,					11,	13,			SEXP_ACTION_OPERATOR,	},	// Goober5000
+	{ "explosion-effect",				OP_EXPLOSION_EFFECT,					11,	14,			SEXP_ACTION_OPERATOR,	},	// Goober5000
 	{ "warp-effect",					OP_WARP_EFFECT,							12, 12,			SEXP_ACTION_OPERATOR,	},	// Goober5000
 	{ "remove-weapons",					OP_REMOVE_WEAPONS,						0,	1,			SEXP_ACTION_OPERATOR,	},	// Karajorma
 	{ "set-time-compression",			OP_CUTSCENES_SET_TIME_COMPRESSION,		1,	3,			SEXP_ACTION_OPERATOR,	},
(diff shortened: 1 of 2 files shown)


asarium:
Could this alter the behavior of existing missions by altering how long the EMP effect affects turrets? It's probably not an issue but it should be considered before merging this.

Goober5000:
The time is only altered for weapons that have the "same emp time for capships" flag.  And, due to this bug, the flag never worked properly.  If anyone other than me had used this flag, they would have very quickly discovered that it had the opposite of the intended effect.

asarium:
Ok, that's what I already guessed.

code/parse/sexp.cpp
```diff
@@ -10812,6 +10818,11 @@ void sexp_explosion_effect(int n)
 		emp_duration = eval_num(n);
 		n = CDR(n);
 	}
+	if (n != -1)
+	{
+		use_emp_time_for_capship_turrets = is_sexp_true(n);
```
MageKing17: For some reason, `is_sexp_true()` returns an `int`; since `use_emp_time_for_capship_turrets` is a `bool`, this line is generating a performance warning in VS (upgraded to an error, causing the AppVeyor failure). Either `is_sexp_true()` needs to return a `bool` or, if that's not feasible for whatever reason, the call needs to be changed to something like `!!is_sexp_true(n)` to get rid of the warning.

Goober5000: I'll open a new PR for changing `is_sexp_true` to bool, since this happens all over the place.

Is `!!` a new operator?  I'm not getting search results on it, but being special characters, that's not unexpected.

asarium: `!!` is just the double application of `!`. Since "not not true" is still true it doesn't change the boolean value but it will convert any integer into a true/false value.

chief1983: It's just the same operator twice, isn't it?  The suggestion is to cast to bool explicitly by inverting it (does the cast of the int to its inverse Boolean), and then invert that again, to get the Boolean equivalent of the integer.  If I'm reading that correctly.

Is there no (bool) operator in C++ for explicit casts though?  I would think it would complain about using !! to perform that cast implicitly too.

Goober5000: Gotcha.

For the record, there are 324 instances of !! in the codebase, and most if not all of them appear to be the more conventional usage.

chief1983: I got ninja'd.  FWIW, I google'd C++ Boolean cast and found the third link mentioned using !! in the subject.  But I also found a StackOverflow answer that has a compelling argument against not using !! like this, and writing something easier to read and less likely to be mistaken for a typo.

https://stackoverflow.com/questions/1310344/why-use-when-converting-int-to-bool

Basically, just compare it to 0 explicitly to get a Boolean result from the comparison.

`use_emp_time_for_capship_turrets = is_sexp_true(n) != 0;`

Goober5000: Yeah, most of the uses of `is_sexp_true` do it that way.  But this will shortly become moot, as the PR to change to bool is nearly ready to go.

Goober5000:
See PR #1464.  After that's approved I'll rebase this one.

code/parse/sexp.cpp
Goober5000: Ok, the other PR is merged and I've rebased this one.

PR #1464, opened 2017-08-29
change is_sexp_true return type from int to bool

Goober5000:
What it says on the tin.

Any reviewers should pay particular attention to `sexp_or`, `sexp_and`, and `sexp_and_in_sequence`.  I double-checked my changes carefully, but extra eyes are always helpful.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -4372,7 +4372,7 @@ int sexp_or(int n)
 	{
 		if (CAR(n) != -1)
 		{
-			result = result || is_sexp_true(CAR(n));
+			result = is_sexp_true(CAR(n)) || result;
 			if (Sexp_nodes[CAR(n)].value == SEXP_KNOWN_TRUE)
 				return SEXP_KNOWN_TRUE;								// if one of the OR clauses is TRUE, whole clause is true
 			if (Sexp_nodes[CAR(n)].value != SEXP_KNOWN_FALSE)		// if the value is still unknown, they all can't be false
@@ -4380,11 +4380,11 @@ int sexp_or(int n)
 		}
 		// this should never happen, because all arguments which return logical values are operators
 		else
-			result = result || (atoi(CTEXT(n)) != 0);
+			result = (atoi(CTEXT(n)) != 0) || result;
 
 		while (CDR(n) != -1)
 		{
-			result = result || is_sexp_true(CDR(n));
+			result = is_sexp_true(CDR(n)) || result;
 			if ( Sexp_nodes[CDR(n)].value == SEXP_KNOWN_TRUE )
 				return SEXP_KNOWN_TRUE;								// if one of the OR clauses is TRUE, whole clause is true
 			if ( Sexp_nodes[CDR(n)].value != SEXP_KNOWN_FALSE )		// if the value is still unknown, they all can't be false
@@ -4412,7 +4412,7 @@ int sexp_and(int n)
 	{
 		if (CAR(n) != -1)
 		{
-			result = result && is_sexp_true(CAR(n));
+			result = is_sexp_true(CAR(n)) && result;
 			if ( Sexp_nodes[CAR(n)].value == SEXP_KNOWN_FALSE || Sexp_nodes[CAR(n)].value == SEXP_NAN_FOREVER )
 				return SEXP_KNOWN_FALSE;							// if one of the AND clauses is FALSE, whole clause is false
 			if ( Sexp_nodes[CAR(n)].value != SEXP_KNOWN_TRUE )		// if the value is still unknown, they all can't be true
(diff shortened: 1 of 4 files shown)


code/parse/sexp.cpp
```diff
@@ -4365,30 +4365,30 @@ int rand_sexp(int n, bool multiple)
...
 				return SEXP_KNOWN_TRUE;								// if one of the OR clauses is TRUE, whole clause is true
-			if ( Sexp_nodes[CAR(n)].value != SEXP_KNOWN_FALSE )		// if the value is still unknown, they all can't be false
-				all_false = 0;
+			if (Sexp_nodes[CAR(n)].value != SEXP_KNOWN_FALSE)		// if the value is still unknown, they all can't be false
+				all_false = false;
 		}
+		// this should never happen, because all arguments which return logical values are operators
 		else
-			result |= atoi(CTEXT(n));
+			result |= (atoi(CTEXT(n)) != 0);
```
asarium: This should be changed to `result = result || atoi(CTEXT(n)) != 0`. I'm not sure if `|=` works the same way on booleans as it does for ints but it would certainly be cleaner to do it that way.

code/parse/sexp.cpp
```diff
@@ -4417,21 +4416,19 @@ int sexp_and(int n)
 			if ( Sexp_nodes[CAR(n)].value == SEXP_KNOWN_FALSE || Sexp_nodes[CAR(n)].value == SEXP_NAN_FOREVER )
 				return SEXP_KNOWN_FALSE;							// if one of the AND clauses is FALSE, whole clause is false
 			if ( Sexp_nodes[CAR(n)].value != SEXP_KNOWN_TRUE )		// if the value is still unknown, they all can't be true
-				all_true = 0;
+				all_true = false;
 		}
+		// this should never happen, because all arguments which return logical values are operators
 		else
-			result &= atoi(CTEXT(n));
+			result &= (atoi(CTEXT(n)) != 0);
```
asarium: Same as above but with `&&`.

code/parse/sexp.cpp
```diff
@@ -4417,21 +4416,19 @@ int sexp_and(int n)
...
 
 		while (CDR(n) != -1)
 		{
-			int new_result;
-
-			new_result = is_sexp_true(CDR(n));
-			result &= new_result;
+			result &= is_sexp_true(CDR(n));
```
asarium: And again.

code/parse/sexp.cpp
```diff
@@ -4440,30 +4437,30 @@ int sexp_and(int n)
...
-				return SEXP_KNOWN_FALSE;														// if one of the AND clauses is FALSE, whole clause is false
+				return SEXP_KNOWN_FALSE;							// if one of the AND clauses is FALSE, whole clause is false
 			if ( Sexp_nodes[CAR(n)].value != SEXP_KNOWN_TRUE )		// if value is true, mark our all_true variable for later checking
-				all_true = 0;
+				all_true = false;
 		}
+		// this should never happen, because all arguments which return logical values are operators
 		else
-			result &= atoi(CTEXT(n));
+			result &= (atoi(CTEXT(n)) != 0);
```
asarium: Same as above.

Goober5000:
Good catch.  I thought those would act as logical operators; a search indicates this isn't the case.

Goober5000:
Made one important change to the PR.  I realized that C++ will short-circuit logical evaluation whereas the SEXP operators are designed to evaluate all of their logical arguments.

Goober5000:
Anyone else want to take a look at this?  Would be nice to get a third pair of eyes just for sanity's sake.

On that note, how does one restart Appveyor?

asarium:
If you push a commit to your branch then Appveyor will start a new build. Alternatively, if the build failed for some external reason it's possible to resend the GitHub webhook but as far as I can tell, the reason for the failure was a warning in the build so unless you fix that it will just fail again.

MageKing17:
Restarting AppVeyor won't help when you haven't fixed the problem:

```C:\projects\fs2open-github-com\code\parse\sexp.cpp(6727): warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning) [C:\projects\fs2open-github-com\build\code\code.vcxproj]
C:\projects\fs2open-github-com\code\parse\sexp.cpp(17057): warning C4804: '>' : unsafe use of type 'bool' in operation [C:\projects\fs2open-github-com\build\code\code.vcxproj]```

Goober5000:
Ah, sorry, I looked for warnings but didn't see any.  Must have missed those.

MageKing17:
You've changed `subjective` to a `bool`, but the multi code is still expecting an `int`.

Goober5000:
Yeah, I just made a quit change and figured the test build would catch it if there was a problem.  And sure enough there was. :p

PR #1466, opened 2017-08-31
Add new libRocket UI system

asarium:
This new system allows to create a user interface that goes beyond the
existing UI system.

libRocket requires FreeType for TTF support and using the pre-built libraries
does not work for all platforms so I included it in source form which will be
compiled as part of our build.


asarium:
I don't like the idea of including the entirety of the FreeType source so if anyone has a better idea about how that library could be integrated into our build system, please let me know,

PR #1467, opened 2017-08-31
add field for setting the hotspot index of the exit button in the medals screen

Goober5000:
As requested by Admiral Nelson here:
http://www.hard-light.net/forums/index.php?topic=93704.0

Already tested by him, and it works.


asarium:
The Appveyor build didn't start for some reason. I resent the webhook and it seems to be working now.

Goober5000:
AppVeyor was successful; not sure why it's showing as incomplete.  Any ideas?

MageKing17:
No, but I see no reason to hold the PR up when the build succeeded.

PR #1473, opened 2017-09-04
Only use 16 bit-per-pixel for PCX data loading

### LINKED ISSUE #1472: Retail models are rendered completely black after #1318
Well, not only retail models but some older too. Models from Mediavps 2014 are not affected. Tested on nvidia driver 385.41 on windows and intel 17.1.5 on debian.

> f1bf0a411c912f3cac4bef147043dfb563bdb74a is the first bad commit
> commit f1bf0a411c912f3cac4bef147043dfb563bdb74a
> Author: asarium
> Date:   Thu Feb 2 11:09:13 2017 +0100
> 
>     Use OpenGL texture arrays for animations
>     
>     This converts all usages of bmpman textures to OpenGL texture arrays.
>     That means that all frames of an animation are stored in a single OpenGL
>     texture object and can be accessed at the same time in a single draw
>     call. This should significantly reduce the overhead of drawing lots of
>     particles since all particles of a single animation can be drawn with a
>     single draw call.
>     
>     This fixes #1169.

asarium:
The PCX loader requires this or else the data isn't loaded correctly.
This matches what the code did before the texture array merge but it
only does it for PCX textures now since the other formats handle 32 and
24 bits per pixel correctly.

This fixes #1472.

### DIFF
--- a/code/graphics/opengl/gropengltexture.cpp
+++ b/code/graphics/opengl/gropengltexture.cpp
@@ -759,17 +759,27 @@ void opengl_determine_bpp_and_flags(int bitmap_handle, int bitmap_type, ubyte& f
 		case TCACHE_TYPE_CUBEMAP:
 		case TCACHE_TYPE_NORMAL:
 			flags |= BMP_TEX_OTHER;
-			if (bm_has_alpha_channel(bitmap_handle)) {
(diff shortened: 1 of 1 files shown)


asarium:
The Mac build stalled for some reason so I restarted the build.

PR #1477, opened 2017-09-09
Use geometry shader instancing for shadow map rendering

asarium:
This provides ~10% to 20% of an improvement if shadows are enabled. The
required features are not available in our OpenGL target version so I
added a fallback meachanism that will continue to use the standard
instancing method.

With an RX 480 and these changes I can play/watch the BP2 Icarus
cutscene at a consistent 60FPS which wasn't possible before.

Here is the performance difference between current master and these changes:
Performance comparission


asarium:
Here is another performance graph and this time without FPS capping:
No FPS capping

Issue #1041, opened 2016-10-14
Allow to use movie files for other purposes


asarium:
Movie files are an ideal choice for CBanims since those are usually pretty long and could be compressed pretty well by the video codecs. They also typically don't use an alpha channel which is good since typical video codecs also don't support an alpha channel. The obvious choice for adding this functionality is the `generic_anim` system but that relies on bmpman which may cause issues since video files use the YUV encoding which is supported by the video player by directly using OpenGL textures.

A special bitmap type could be introduced that is capable of holding this YUV data. Combined with a special rendering material that encapsulates video rendering it could be used to properly integrate video files into the generic anim system without requiring major changes to the code base (at least I hope...).

Goober5000:
I believe this is now complete.

PR #1484, opened 2017-09-16
Improve video player API

asarium:
This moves all the cutscene specific code out of the movie player code
and into the cutscene code so that the movie player can be reused for
usage in other environments.

This is required for #1041 since an implementation of that feature would
reuse the movie player code of the cutscene code but without all the
main loop stuff.


asarium:
These changes have been merged together with #1523.

PR #1486, opened 2017-09-20
Pack vertex and index data into one buffer

### LINKED ISSUE #1457: Improving performance by packing all Vertex data into one common buffer
We already manage the vertex and index buffers in such a way that all data of a single object is stored in a single buffer object (where possible). This concept could be expanded to all loaded models by packing all that data into the same buffer which should reduce driver overhead and improve performance since the buffers would not need to be rebound every time a new model is drawn.

My plan for implementing this is the following:

- For both vertex and index data, the engine creates separate "heaps" (big buffer objects) that will store the data. This heap would need to be managed by a memory allocator but since the data is relatively static a simple allocation algorithm should suffice.
- Whenever the model load code want to store data on the GPU it requests space from this allocator and instead of normal buffer handle, it receives a buffer handle and an offset into the buffer.
- Since this offset would need to be passed to the GPU via the base-vertex parameter, the stride of the data for each vertex would need to be the same. If that is not the case then a separate heap with that stride would need to be used. Something similar is already being done in the existing model code.

So, to implement this the following parts should be needed (I think):

- [x] A generic heap memory block manager for keeping track of what memory can be used
- [x] Add a relatively generic heap management system to the graphics code so that external code can allocate and deallocate memory easily
- [x] Change the model code so that it uses these new heaps for allocating the model data. This also includes changing the existing vertex offset value so that it matches the offset into the vertex heap.

As far as I can tell, these changes won't actually require changes to the OpenGL code.

This is a rough outline of what I would like to do in this project. I will post updates about this here whenever I manage to actually implement something.

asarium:
This changes the model loading code such that all index and vertex data
is packed in a single GPU buffer for each data type (index and vertex
data are still in separate buffers). This reduces the amount of buffer
rebinding needed when rendering a scene which will reduce the CPU
overhead. Furthermore, this opens up possibilities of future
improvements for which it is beneficial if all data is in the same
buffer (e.g. MultiDrawIndirect rendering).

To manage where to put the data in the GPU buffer I implemented a
custom HeapAllocator which manages where data can be put. It uses a very
primitive first-fit algorithm but that is sufficient for this usage
type. This does add quite a lot of overhead while allocating GPU model
data (I did some tracing and some allocation calls took as long as
130ms).

This fixes #1457.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -216,7 +216,6 @@ Flag exe_params[] =
 	{ "-use_gldrawelements","Don't use glDrawRangeElements",			true,	0,					EASY_DEFAULT,		"Troubleshoot",	"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-use_gldrawelements", },
 	{ "-old_collision",		"Use old collision detection system",		true,	EASY_DEFAULT,		EASY_ALL_ON,		"Troubleshoot",	"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-old_collision", },
 	{ "-gl_finish",			"Fix input lag on some ATI+Linux systems",	true,	0,					EASY_DEFAULT,		"Troubleshoot", "http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-gl_finish", },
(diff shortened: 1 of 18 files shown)


asarium:
I changed the heap resize behavior a bit which should reduce the load time increase a bit.

PR #1493, opened 2017-10-04
Fix several oversights and bugs in the particle effect code

asarium:
The particle sources of the beams were missing multiple properties
(direction and impact normals) and the generic shape effect didn't have
any way of controlling how long the source was active. This fixes that
by introducing an effect timing field similar to the single particle
effect.


asarium:
I made some more changes to how the source orientation is handled to fix another issue Spoon reported.

PR #1495, opened 2017-10-08
Add support for multiple sounds in one game sound entry

asarium:
This adds support for specifying multiple sounds for a single game sound
entry. The engine then selects which sound will be played based on the
strategy provided by the game data. This required a few changes to some
of the game code since previously every game sound had exactly one sound
file attached to it. The most notable change is that it is no longer
possible to determine the exact duration a sound can have so instead the
new code computes the maximum duration a sound can have and uses that
where appropriate.

This also adds support for randomizing the pitch and volume of the sound
to provide a more varied experience even if only a single sound is used.

One possible issue that exists is that the "sequential" strategy only
works globally at the moment so if two sources use the same game sound
concurrently then the two sources will share the sequence of sounds
since there is no sound instance data at the moment. That can be added
in a future change but should not present a huge issue for most
use-cases.

Visual Studio on Appveyor somehow causes deprecation error on _our_
definition of vsprintf. I have no idea why these changes introduced this
warning but I won't debug Visual Studio compiler bugs that I can't
reproduce with my Visual Studio version so I just disabled those
warnings.

I will post some test builds of this to test the new functionality and to make sure that the changes to the sound system didn't break anything even if no sound sets are used.


code/gamesnd/gamesnd.cpp
```diff
@@ -703,22 +734,122 @@ bool required_string_no_create(const char* token, bool no_create)
     return true;
 }
 
+static GameSoundCycleType parse_cycle_type() {
+	if (optional_string("Sequential")) {
+		return GameSoundCycleType::Sequential;
+	} else if (optional_string("Random")) {
+		return GameSoundCycleType::Random;
+	} else {
+		error_display(0, "Failed to parse sound cycle type. Expected 'sequential' or 'random'. Got [%.32s]", next_tokens());
```
MageKing17: I don't think `next_tokens()` consumes those tokens, so I'm pretty sure this will inevitably result in another parsing error, if somebody actually supplies an invalid cycle type.

asarium: Good point. I added a `advance_to_eoln` call which should skip over the bad token in most cases.

asarium:
The feature has been tested and a few bugs have been fixed. This should be ready for a full code review now.

PR #1497, opened 2017-10-22
Refactors swarm_update_direction() a bit...

z64555:
...for readability and consistency.

* Changed stop distance to a stop frame. Frame was chosen instead of distance or time to ensure missiles won't appear to jerk twice in succession.
** This also gives faster missiles a higher chance to hit their target. Having a hard-coded distance would only work for swarm missiles that had a speed similar to vanilla FS2.

### DIFF
--- a/code/weapon/swarm.cpp
+++ b/code/weapon/swarm.cpp
@@ -30,7 +30,7 @@
 
 #define SWARM_TIME_VARIANCE		100		// max time variance when deciding when to change swarm missile course
 
-#define SWARM_DIST_STOP_SWARMING	300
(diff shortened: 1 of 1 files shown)


z64555:
Future work possibilities:
Allow zig-zag change rates to be modder accessable
Change zig-zag change rates to be based on missile speed, instead of hardcoded time.

Notes:
Fast missiles (swarm or not) have a tendency to miss narrow targets, such as the nose of an Amazon Adv. This appears to be due to over-correction by the ai code, which may or may not be an intended behavior.

PR #1500, opened 2017-10-29
Fix directional glow issues

TRBlount:
There were some issues with positioning and size of directional glows as well as jitter during the startup/cooldown sequence.

### DIFF
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -1418,14 +1418,22 @@ void beam_render_muzzle_glow(beam *b)
 		vertex h1[4];
 		vertex *verts[4] = { &h1[0], &h1[1], &h1[2], &h1[3] };
 		vec3d fvec, top1, top2, bottom1, bottom2, sub1, sub2, start, end;
+		int idx;
+		float g_length = bwi->glow_length * pct * rand_val;
 
 		vm_vec_sub(&fvec, &b->last_shot, &b->last_start);
 		vm_vec_normalize_quick(&fvec);
-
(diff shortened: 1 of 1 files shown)


MageKing17:
Are there some before/after images demonstrating the problem/the effect of this PR?

TRBlount:
Right, here's one from before the new commits.
https://www.dropbox.com/s/zz9y4yjt5kmiu1t/Fs2_open_Wings_of_Dawn-EX10-SSE%202017-10-01%2021-56-21-99.webm?dl=0

I had spoon test the changes on the WoD build and he confirmed that it works correctly now, ie. the glows no longer go back into the gun barrel and have a proper aspect ratio.

code/weapon/beam.cpp
```diff
@@ -1418,14 +1418,22 @@ void beam_render_muzzle_glow(beam *b)
...
 
 		vm_vec_sub(&fvec, &b->last_shot, &b->last_start);
 		vm_vec_normalize_quick(&fvec);
-
-		vm_vec_copy_scale(&sub1, &fvec, rad);
-		vm_vec_sub(&start, &b->last_start, &sub1);
-		vm_vec_copy_scale(&sub2, &fvec, bwi->glow_length);
-		vm_vec_add(&end, &start, &sub2);
+		
+		if (bwi->glow_length >= 2.0f*rad) {
```
asarium: This condition could use a comment to explain why the different behaviors are necessary. I assume that it has to do with how directional glows are handled but without looking through the entire code to determine what values `glow_length` and `rad` can take it is not possible to understand this condition.

PR #1501, opened 2017-10-31
Make DRADIS blips use decaying alpha

TRBlount:
DRADIS blips were originally intended to have decaying alpha. After some changes related to materials, they no longer used the alpha and the original functionality was not in line with what the designer wanted.
The blips now decay in alpha based on the time since they were last swept by the DRADIS.

### DIFF
--- a/code/radar/radardradis.cpp
+++ b/code/radar/radardradis.cpp
@@ -132,7 +132,7 @@ void HudGaugeRadarDradis::plotBlip(blip* b, vec3d *pos, float *alpha)
 	
 	b->time_since_update += flFrametime;
 	// If the blip has been pinged by the local x-axis sweep, update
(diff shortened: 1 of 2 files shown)


code/radar/radardradis.cpp
```diff
@@ -129,8 +129,13 @@ void HudGaugeRadarDradis::plotBlip(blip* b, vec3d *pos, float *alpha)
 			fade_multi *= 2.0f;
 		}
 	}
+	
+	b->time_since_update += flFrametime;
+	if (abs(vm_vec_dot(&sweep_normal_x, pos)) < 0.01f) {
```
asarium: This condition would benefit from a comment to explain what it does because I have no idea what this condition checks.

PR #1502, opened 2017-11-03
Allow BMPMAN slot limit to be defined by mod table.

MageKing17:
`bm_init()` now dynamically allocates the `bm_bitmaps` array, and `MAX_BITMAPS` can be changed by the mod table.

After the question was raised as to the possibility of making the BMPMAN limit be a mod table entry (as a way to make Wings of Dawn no longer require custom builds with an increased limit), it was pointed out that the mod table gets parsed before BMPMAN is initialized, which led to me trying out allocating the array at runtime. I've tested it and it *seems* to work.

### DIFF
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -96,7 +96,7 @@ bitmap_entry* bm_bitmaps = nullptr;
 
 // --------------------------------------------------------------------------------------------------------------------
 // Definition of private variables at file scope (static).
-static int bm_inited = 0;
+static bool bm_inited = false;
 static uint Bm_next_signature = 0x1234;
 static int  bm_next_handle = 1;
 static int Bm_low_mem = 0;
@@ -487,7 +487,7 @@ void bm_close() {
 			delete[] bm_bitmaps;
 			bm_bitmaps = nullptr;
 		}
-		bm_inited = 0;
+		bm_inited = false;
 	}
 }
 
@@ -498,7 +498,7 @@ int bm_create(int bpp, int w, int h, void *data, int flags) {
 		Assert((bpp == 16) || (bpp == 24) || (bpp == 32));
 	}
 
-	if (!bm_inited) bm_init();
+	Assertion(bm_inited, "bmpman must be initialized before this function can be called!");
 
 	int n = -1;
 
@@ -949,7 +949,7 @@ void bm_get_palette(int handle, ubyte *pal, char *name) {
 }
 
 uint bm_get_signature(int handle) {
-	if (!bm_inited) bm_init();
+	Assertion(bm_inited, "bmpman must be initialized before this function can be called!");
 
 	int bitmapnum = handle % MAX_BITMAPS;
 	Assertion(bm_bitmaps[bitmapnum].handle == handle, "Invalid bitmap handle %d passed to bm_get_signature().\nThis might be due to an invalid animation somewhere else.\n", handle);		// INVALID BITMAP HANDLE!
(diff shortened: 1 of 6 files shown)


code/bmpman/bmpman.h
```diff
@@ -45,7 +45,9 @@
  *   If anything we could/should reduce MAX_BITMAPS in the future.  Where it's at now should accomidate even the
  *   largest mods.  --  Taylor
  */
-#define MAX_BITMAPS 4750
+#define DEFAULT_MAX_BITMAPS 4750
+
+extern int MAX_BITMAPS;
```
asarium: Since this is basically the size of an array it should be a `size_t`.

code/mod_table/mod_table.cpp
```diff
@@ -37,6 +37,7 @@ bool Beams_use_damage_factors = false;
 float Generic_pain_flash_factor = 1.0f;
 float Shield_pain_flash_factor = 0.0f;
 gameversion::version Targetted_version(2, 0, 0, 0); // Defaults to retail
+int MAX_BITMAPS = DEFAULT_MAX_BITMAPS;
```
asarium: This variable was declared as `exten` in bmpman.h but the actual variable is stored in mod_table.cpp. The extern declaration and variable definition should be in the same header/source pair.

code/mod_table/mod_table.cpp
```diff
@@ -211,6 +212,17 @@ void parse_mod_table(const char *filename)
 				 mprintf(("Game Settings Table: Setting shield pain flash factor to %.2f\n", Shield_pain_flash_factor));
 		}
 
+		if (optional_string("$BMPMAN Slot Limit:")) {
+			int slots;
+			stuff_int(&slots);
+			if (slots < 3500) {
+				Warning(LOCATION, "Invalid BMPMAN slot limit [%d]; must be at least 3500.\n", slots);
```
asarium: This should use `error_display` so that the warning message also includes the file and line number of the currently parsed table.

code/bmpman/bmpman.cpp
```diff
@@ -992,14 +993,16 @@ bool bm_has_alpha_channel(int handle) {
...
-	mprintf(("Size of bitmap extra info = " SIZE_T_ARG " bytes\n", sizeof(bm_extra_info)));
-
 	if (!bm_inited) {
 		bm_inited = 1;
 		atexit(bm_close);
 	}
 
+	bm_bitmaps = new bitmap_entry[MAX_BITMAPS];
+
+	mprintf(("Size of bitmap info = " SIZE_T_ARG " KB\n", sizeof(bm_bitmaps) / 1024));
```
asarium: Just noticed that this was not changed. `sizeof(bm_bitmaps)` is the size of the pointer now and not of the underlying array. You need to use `sizeof(bm_bitmaps[0]) * MAX_BITMAPS` here.

code/mod_table/mod_table.cpp
MageKing17: I wanted it available to everything that previously had access to it (by importing bmpman.h) without forcing them all to include mod_table.h.

code/mod_table/mod_table.cpp
MageKing17: None of the other parse warnings in mod_table.cpp use `error_display()`; I was just keeping consistent.

code/mod_table/mod_table.cpp
asarium: Then maybe the variable should be defined in bmpman.cpp?

code/mod_table/mod_table.cpp
asarium: In my opinion, it is more important to use the right function for the warnings instead of keeping the current style of the file.

code/mod_table/mod_table.cpp
MageKing17: That works.

code/bmpman/bmpman.h
MageKing17: For anyone who didn't follow the discussion on Discord, changing it to `size_t` causes multiple problems (like crashing in certain functions if the handle is negative), so it'll be staying an `int`.

PR #1505, opened 2017-11-09
Add in-game BMPMAN usage display.

MageKing17:
Can be enabled with the "-bmpmanusage" command-line option, toggled via the debug console, or toggled via a debug/cheat key combo (`+B).

I wrote this while I was working on #1502, and it makes an interesting diagnostic tool (although the fact that framerate isn't displayed during menus means, since it uses the same function, you also can't monitor BMPMAN usage during menus, which limits its utility somewhat). ~~I *may* do some experiments with calling `game_show_frametime()` outside of missions, but otherwise~~ this PR is complete.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -1460,6 +1460,7 @@ extern int Training_message_method;
 DCF_BOOL( training_msg_method, Training_message_method )
 DCF_BOOL( show_player_pos, Show_player_pos )
 DCF_BOOL(i_framerate, Interface_framerate )
+DCF_BOOL( show_bmpman_usage, Cmdline_bmpman_usage )
 
 DCF(warp, "Tests warpin effect")
 {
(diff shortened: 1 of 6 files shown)


freespace2/freespace.cpp
```diff
@@ -2047,6 +2048,9 @@ void game_get_framerate()
 	Framecount++;
 }
 
+int bmpman_count_bitmaps();
```
asarium: This function should be declared in the bmpman header instead of here.

freespace2/freespace.cpp
```diff
@@ -2047,6 +2048,9 @@ void game_get_framerate()
...
+extern int MAX_BITMAPS;
```
asarium: This variable is already declared in the bmpman header so that header should be included instead of using an `extern` variable here.

MageKing17: Oh right; I meant this to be a temporary thing just to see if it was working, and then forgot to fix it later. Let's take care of that right now...

MageKing17:
Okay, `bmpman_count_bitmaps()` is now in bmpman.h. I tried, briefly, to make the framerate display work with mainhalls, but nothing displayed, so I probably cocked it up; since it's still possible to check BMPMAN usage during menus with the debug console, I've decided to just leave it be for now.

PR #1509, opened 2017-11-11
Add support for printing command line flags in JSON output

asarium:
This will output the flags data in a better format than the current
binary format. For determining if the JSON output should be used this
uses an optional parameter for -get_flags. The command line parsing does
not cause an error if a parameter is specified for an option that does
not support them. This way, the launcher can just pass "json_v1" to the
option and an old build will just silently ignore it and use the binary
format.

### DIFF
--- a/code/io/joy-sdl.cpp
+++ b/code/io/joy-sdl.cpp
@@ -800,6 +800,42 @@ namespace joystick
 
 		SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
 	}
+
+	SCP_vector<JoystickInformation> getJoystickInformations() {
+		SCP_vector<JoystickInformation> joystickInfo;
+
+		if (SDL_InitSubSystem(SDL_INIT_JOYSTICK) < 0) {
+			return SCP_vector<JoystickInformation>();
+		}
+
+		auto num_joysticks = SDL_NumJoysticks();
+		for (auto i = 0; i < num_joysticks; ++i) {
+			auto joystick = SDL_JoystickOpen(i);
+
+			JoystickInformation info{};
+
+			auto name = SDL_JoystickName(joystick);
+			if (name != nullptr) {
+				info.name = name;
+			}
+			info.guid = getJoystickGUID(joystick);
+
+			info.num_axes = static_cast<uint32_t>(SDL_JoystickNumAxes(joystick));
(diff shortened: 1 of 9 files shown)


MageKing17:
So it outputs to STDOUT instead of flags.lch? That makes a great deal of sense.

I could've sworn there was already functionality for accepting an optional argument for `-get_flags`, though; an integer, defaulting to 0 if not present. Was that just a planned feature, or am I misremembering entirely?

asarium:
I didn't see any code that handled parameters for `-get_flags`. Maybe there was something like this in there at some point but as far as I can tell, there is nothing like that in current master.

asarium:
I added some more information to the JSON output of `-get_flags` which adds information output about connected displays, joysticks, sound devices and available TTS voices. That will make it easier to develop a cross-platform launcher since all the enumeration code is contained within FSO.

It also makes the system future-proof in case the device enumeration changes (e.g. like it did when we switched to SDL2 on all platforms). Newer builds would still output the data in the old format so the launcher would not have to support a new way of enumerating the available devices.

code/io/joy-sdl.cpp
```diff
@@ -800,6 +800,42 @@ namespace joystick
...
+				info.name = name;
+			}
+			info.guid = getJoystickGUID(joystick);
+
+			info.num_axes = static_cast<uint32_t>(SDL_JoystickNumAxes(joystick));
+			info.num_balls = static_cast<uint32_t>(SDL_JoystickNumBalls(joystick));
+			info.num_buttons = static_cast<uint32_t>(SDL_JoystickNumButtons(joystick));
+			info.num_hats = static_cast<uint32_t>(SDL_JoystickNumHats(joystick));
+
+			info.is_haptic = static_cast<bool>(SDL_JoystickIsHaptic(joystick));
```
MageKing17: This line is generating a warning in VS from the `static_cast<bool>`; you need to either use one of the other paradigms for saying "yes, I really want this to become a boolean value" like a double-not, or change `SDL_JoystickIsHaptic()` to return a `bool`.

asarium: How would I change `SDL_JoystickIsHaptic()` to return a `bool`? It's a C API so it doesn't even have that type and it's also an external API so it's absolutely impossible to change that code.

I fixed it by comparing the return value with `SDL_TRUE` which should be the safest way of converting the result into a `bool`.

PR #1510, opened 2017-11-11
Fix motion debris anim loading in mission-set-nebula sexp

TRBlount:


code/starfield/starfield.cpp
```diff
@@ -2490,16 +2490,20 @@ void stars_set_nebula(bool activate)
 		} else {
 			Neb2_render_mode = NEB2_RENDER_HTL;
 		}
-		Debris_vclips = Debris_vclips_nebula;
```
asarium: Why was this (and the similar line below) removed? It would be nice to have some kind of reason just in case we might need it in the future.

MageKing17: They were removed because `Debris_vclips_nebula` and `Debris_vclips_normal` are basically containers for the names of the animations to load, and a convenient place to store the bitmap handle and number of frames. Trying to simply swap between the two like this, when only one set gets loaded at a time, is doomed to failure; the function called below (``stars_load_debris()`) releases the existing set and loads the new set, so does what these lines were *intended* to do, except it does it properly (by unloading the current set, loading the new set, and flagging which set is loaded by setting the `stars_debris_loaded` varaible).

asarium: Thank you for the explanation.

PR #1514, opened 2017-11-17
Window title overhaul.

MageKing17:
FSO now allows the window title to be specified via game_settings.tbl, some never-used function arguments were pruned at the same time, and fred.rc now receives version information (for the About dialogue and the title bar) from a cmake-generated header, so the version numbers get filled in properly. Speaking of FRED's About dialogue, it now displays the copyright symbol properly.

### DIFF
--- a/code/graphics/opengl/gropengl.cpp
+++ b/code/graphics/opengl/gropengl.cpp
@@ -72,6 +72,7 @@ static GLuint GL_screen_pbo = 0;
 float GL_alpha_threshold = 0.0f;
 
 extern const char *Osreg_title;
+extern SCP_string Window_title;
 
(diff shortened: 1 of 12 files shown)


code/graphics/opengl/gropengl.cpp
```diff
@@ -1020,6 +1021,9 @@ int opengl_init_display_device()
 	attrs.height = (uint32_t) gr_screen.max_h;
 
 	attrs.title = Osreg_title;
+	if (Window_title != "") {
```
asarium: This should be `if (!Window_title.empty())`.

MageKing17: Fair enough.

PR #1517, opened 2017-11-20
Fix PCX alpha color correction

### LINKED ISSUE #1516: HUD config looks faded
After
```
72a797f73219f709a4617480bdf53c86c721491f is the first bad commit
commit 72a797f73219f709a4617480bdf53c86c721491f
Author: asarium <asarium@gmail.com>
Date:   Wed Aug 16 15:47:55 2017 +0200

    Move 4 bit color conversion out of graphics code
    
    Currently, the graphics code does some color remapping on 8
    bit-per-pixel textures since those textures use special 4-bit pixel
    values. These changes move that conversion out of the graphics code and
    into the image loading/font bitmap creation code.
    
    The reason this is needed is because this conversion breaks 8-bit
    textures that try to use the full range of colors. I need to do exactly
    that for moving the cutscene renderer to our generic rendering interface
    which I can't do at the moment since the existing color conversion
    corrupts the video data.
    
    I have tested this with retail data and could not find any visual
    changes after applying these changes.

```

Hud config looks faded (?)
screen0046

asarium:
This is a bugfix for an issue introduced in 72a797f73219f709a4617480bdf53c86c721491f which broke 8 bits-per-pixel alpha bitmap reading.

This fixes #1516.

### DIFF
--- a/code/pcxutils/pcxutils.cpp
+++ b/code/pcxutils/pcxutils.cpp
@@ -213,7 +213,7 @@ typedef struct { ubyte b, g, r, a; } COLOR32;
 #endif
 
 //int pcx_read_bitmap_16bpp( char * real_filename, ubyte *org_data, ubyte bpp, int aabitmap, int nondark )
-int pcx_read_bitmap( const char * real_filename, ubyte *org_data, ubyte *pal, int byte_size, int aabitmap, int nondark, int cf_type )
+int pcx_read_bitmap( const char * real_filename, ubyte *org_data, ubyte *pal, int byte_size, int aabitmap, bool mask_bitmap, int cf_type )
 {
 	PCXHeader header;
 	CFILE * PCXfile;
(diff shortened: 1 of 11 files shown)


asarium:
I checked all `bm_lock` calls to make sure the mask bitmaps are loaded with the right flag. There were some issues with that but those should be fixed now.

code/pcxutils/pcxutils.cpp
```diff
@@ -310,7 +310,22 @@ int pcx_read_bitmap( const char * real_filename, ubyte *org_data, ubyte *pal, in
...
+					if (!mask_bitmap) {
+						// 8 bit-per-pixel aa bitmaps are a bit special since they only use values in the range [0, 15] where 15 wraps
+						// around back to 0. Since the rest of the code expects the value to be in the range [0, 255] the pixel value
+						// needs to be adjusted here. By multiplying the value with 17 the original range [0, 15] is mapped to [0, 255]
+						// This only applies to bitmaps that are not used as masks since mask bitmaps use higher values
+						// to indicate their mask area
+						if (data > 15) {
+							pixel_val = 0;
+						} else if (data == 15) {
+							pixel_val = 17;
```
MageKing17: So... 16 or greater wraps around to 0, but 15 is treated like 1?

asarium: Exactly! Welcome to the world of legacy color conversions where the knowledge of why it was done that way has been lost for a long time...

I derived this function from the old translation map which is build by this code: https://github.com/scp-fs2open/fs2open.github.com/blob/fb4169ba9fce882e1a436a4fce67947270aa85e1/code/graphics/opengl/gropengltexture.cpp#L174-L178
Line 178 is the reason for the special case above since the entry for the pixel value 15 is set to the value of 1 which is always `17`. I have no idea why it is done that way, I only replicated it in this code...

MageKing17: Well, okay then... guess it's backwards-compatible.

PR #1521, opened 2017-11-26
Allow GCC builds to use more SIMD instruction sets

PeterMitsis:
When using the GCC toolchain, the cmake build system can now pick the most advanced set of SIMD instructions that both the hardware and the compiler support. The currently coded set of SIMD instruction sets that it searches for is: {sse, sse2, sse3, ssse3, sse4.1, sse4.2, avx, avx2}.

### DIFF
--- a/cmake/cpufeatures.cpp
+++ b/cmake/cpufeatures.cpp
@@ -1,3 +1,13 @@
+#if defined(__GNUC__) && !defined(__x86_64)
+int main(void)
+{
+	/*
+	 * For cases where we are trying to build with GNU
+	 * on a non-x86_64 platform.
+	 */
+	return 0;
+}
+#else
 // Code snippet from http://stackoverflow.com/a/7495023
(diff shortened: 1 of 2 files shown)


MageKing17:
At a glance, it makes sense to me, but I don't know enough about cmake syntax to know if there are any mistakes there, so I'll leave a proper review to someone who does.

asarium:
I appreciate your effort but I don't think this is actually necessary. For GCC our CMake setup already defaults to `-march=native` which should already include support for the best SIMD instructions that the current CPU supports.

Also, this would probably break the existing nightly build since it would try to use whatever instruction set is available on the CI server. That can be easily fixed by setting `GCC_SIMD_OPTIONS` here: https://github.com/scp-fs2open/fs2open.github.com/blob/ea0c4b7b783e8dc660652243c050e79228f3b718/ci/travis/before_script.sh#L29-L31

PR #1525, opened 2017-12-02
Add a few graphics debug scropes to the ship selection

asarium:
I added these while investigating another issue which was fixed by the NanoVG rendering changes but these may still be useful.


asarium:
Since the NanoVG changes fixed the original issue I removed that commit but kept the debug scope changes since there can never be too few debug scopes.

PR #1529, opened 2017-12-04
Replace the special EOF char value with a standard null byte

asarium:
This was an issue since UTF-8 encoded text can contain -128 as part of
an encoding sequence which causes a lot of issues for the parsing code.
I could not find any indication why the original code didn't use null
bytes from the start and I also couldn't find any code that would break
because of this. If anyone knows why this was done like this please let
me know.

### DIFF
--- a/code/parse/parselo.h
+++ b/code/parse/parselo.h
@@ -32,7 +32,6 @@ extern int Token_found_flag;
 
 
 #define	COMMENT_CHAR	(char)';'
-#define	EOF_CHAR			(char)-128
 #define	EOLN				(char)0x0a
 
 #define	F_NAME					1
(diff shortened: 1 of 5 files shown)


asarium:
I found and fixed an issue which was caused by uninitialized data which caused some of the tests to fail randomly. This should now work correctly.

code/parse/parselo.h
```diff
@@ -32,7 +32,6 @@ extern int Token_found_flag;
 
 
 #define	COMMENT_CHAR	(char)';'
-#define	EOF_CHAR			(char)-128
```
MageKing17: Why not just change the define and leave the rest of the code as-is, just in case?

asarium: That would probably break the way the parse system uses `strchr`. For example, in Line 314 in `parselo.cpp` only changing this define would write two null characters into the terminators string. I don't think it would break how the parsing works but it would be a bit inefficient.

PR #1540, opened 2017-12-22
Extends the existing parse pausing system (used by bmpman) to use a

z64555:
bookmark stack.

The stack is First In Last Out, so the parser will resume its most recently saved parsing location before going onto the oldest.

Another mechanism will be needed to ensure double-parsing doesn't occur when resuming from a bookmark, such as an [indexer](https://github.com/z64555/fs2open.github.com/tree/feature/parse_index).

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -31,6 +31,8 @@
 
 #include <utf8.h>
 
+using namespace parse;
+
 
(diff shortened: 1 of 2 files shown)


z64555:
This enhancement should drop-in with the existing system just fine, but should have tests done with multiple pausing/unpausing to ensure intended behavior.

PR #1543, opened 2017-12-23
Convert speech text properly to wide string

asarium:
The previous code could only handle ASCII data correctly but this new
code is also able to handle UTF-8 data correctly by using the Windows
API function `MultiByteToWideChar`. Since the conversion only happens in
Windows-only code this is not a portability issue.

This fixes an issue reported on the forums where non-ASCII text was not
spoken correctly by the TTS system.

### DIFF
--- a/code/sound/speech.cpp
+++ b/code/sound/speech.cpp
@@ -49,6 +49,7 @@
 #pragma warning(pop)
 
 #include "globalincs/pstypes.h"
+#include "utils/unicode.h"
 #include "speech.h"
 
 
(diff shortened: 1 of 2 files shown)


code/sound/speech.cpp
```diff
@@ -106,28 +107,40 @@ bool speech_play(const char *text)
...
-	for(size_t i = 0; i < len; i++) {
-		if(text[i] == '$') {
-			i++;
+	SCP_string work_buffer;
+
+	bool saw_dollar = false;
+	for (auto ch : unicode::codepoint_range(text)) {
+		if (ch == UNICODE_CHAR('$')) {
+			// Skip $ escape sequences which appear in briefing text
+			saw_dollar = false;
```
MageKing17: ...er... surely this should be `saw_dollar = true;`?

MageKing17: also there should probably be a `continue` here as well

asarium: Indeed.

PR #1544, opened 2017-12-24
Consistent beam damage across framerates/times

Flip-Sandwich:
Previous beam damage code was this thing relying on timestamps and "when should I hit next" which was kinda bad and broke beams at both high and low FPS and time compression so I replaced it by having beams apply damage every second, determined by the frametime

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -18405,11 +18405,11 @@ int ship_get_sound(object *objp, GameSoundsIndex id)
 	Assert( objp != NULL );
 	Assert( gamesnd_game_sound_valid(id) );
 
-	// ugh, it's possible that we're an observer at this point
-	if (objp->type == OBJ_OBSERVER)
+	// It's possible that this gets called when an object (in most cases the player) is dead or an observer
+	if (objp->type == OBJ_OBSERVER || objp->type == OBJ_GHOST)
 		return id;
 
-	Assert( objp->type == OBJ_SHIP );
+	Assertion(objp->type == OBJ_SHIP, "Expected a ship, got '%s'.", Object_type_names[objp->type]);
 
 	ship *shipp = &Ships[objp->instance];
 	ship_info *sip = &Ship_info[shipp->ship_info_index];
@@ -18427,7 +18427,11 @@ bool ship_has_sound(object *objp, GameSoundsIndex id)
 	Assert( objp != NULL );
 	Assert( gamesnd_game_sound_valid(id) );
 
-	Assert( objp->type == OBJ_SHIP );
+	// It's possible that this gets called when an object (in most cases the player) is dead or an observer
+	if (objp->type == OBJ_OBSERVER || objp->type == OBJ_GHOST)
+		return false;
+
+	Assertion(objp->type == OBJ_SHIP, "Expected a ship, got '%s'.", Object_type_names[objp->type]);
 
 	ship *shipp = &Ships[objp->instance];
 	ship_info *sip = &Ship_info[shipp->ship_info_index];


chief1983:
The concept sounds solid, but a couple of requests for future PRs - can we omit profanity in this medium please, and have more descriptive commit messages for at least the initial commit in a PR?  Especially one that does not reference an existing issue number.  The former is because this is a semi-professional platform, and many contributors do reference this on job applications, etc, and the latter is because it simply helps with documentation of changes and reading changelogs.  The commit message is also the only part that shows up in our automated nightly posts to the forum where people check to see what's new in the current nightly, so as you can imagine there wouldn't be much to gather from this message as is.

code/weapon/beam.cpp
```diff
@@ -3044,32 +3047,25 @@ void beam_handle_collisions(beam *b)
...
 		r_coll[r_coll_count].c_objnum = target;
 		r_coll[r_coll_count].c_sig = Objects[target].signature;
-		r_coll[r_coll_count].c_stamp = -1;
 		r_coll[r_coll_count].cinfo = b->f_collisions[idx].cinfo;
 		r_coll[r_coll_count].quadrant = -1;
 		r_coll[r_coll_count].is_exit_collision = 0;
 		
+
+		// get the real damage that should be done
+		damage_time_mod = (flFrametime * 1000.0) / i2fl(BEAM_DAMAGE_TIME);
```
MageKing17: You either need to change `1000.0` to `1000.0f`, or wrap the whole right side of this equation in a conversation to `float` to avoid the warning stopping AppVeyor from passing this PR.

Flip-Sandwich:
Gotcha

Flip-Sandwich:
Gotcha

code/weapon/beam.cpp
```diff
@@ -3044,32 +3047,24 @@ void beam_handle_collisions(beam *b)
...
 		r_coll[r_coll_count].c_objnum = target;
 		r_coll[r_coll_count].c_sig = Objects[target].signature;
-		r_coll[r_coll_count].c_stamp = -1;
 		r_coll[r_coll_count].cinfo = b->f_collisions[idx].cinfo;
 		r_coll[r_coll_count].quadrant = -1;
 		r_coll[r_coll_count].is_exit_collision = 0;
 		
+		// get the real damage that should be done
+		damage_time_mod = (flFrametime * 1000.0f) / i2fl(BEAM_DAMAGE_TIME);
+		real_damage = real_damage * damage_time_mod;
```
asarium: This means that the actual damage dealt will decrease with every collision. Is that also what happens in the current code? I cannot see which code in the current version would be responsible for this.

code/weapon/beam.cpp
```diff
@@ -3044,32 +3047,24 @@ void beam_handle_collisions(beam *b)
 		// add to the recent collision list
 		r_coll[r_coll_count].c_objnum = target;
 		r_coll[r_coll_count].c_sig = Objects[target].signature;
-		r_coll[r_coll_count].c_stamp = -1;
 		r_coll[r_coll_count].cinfo = b->f_collisions[idx].cinfo;
 		r_coll[r_coll_count].quadrant = -1;
 		r_coll[r_coll_count].is_exit_collision = 0;
 		
+		// get the real damage that should be done
+		damage_time_mod = (flFrametime * 1000.0f) / i2fl(BEAM_DAMAGE_TIME);
```
asarium: This value is constant throughout all loop iterations (as far as I can tell) and should be computed before the for loop.

code/weapon/beam.cpp
```diff
@@ -3200,7 +3195,7 @@ void beam_handle_collisions(beam *b)
 			}
 			// <-- KOMET_EXT
 		} else {
-			if(draw_effects && do_damage && !physics_paused){
+			if(draw_effects && !physics_paused){
```
asarium: This (and other similar changes below) means that impact effects will now be drawn every frame instead of every `BEAM_DAMAGE_TIME` milliseconds. That might be a performance issue and a behavioral change compared to earlier engine versions.

code/weapon/beam.cpp
```diff
@@ -3294,16 +3289,17 @@ void beam_handle_collisions(beam *b)
 				// hit the ship - again, the innards of this code handle multiplayer cases
 				// maybe vaporize ship.
 				//only apply damage if the collision is not an exit collision.  this prevents twice the damage from being done, although it probably be more realistic since two holes are being punched in the ship instead of one.
+
+
 				if (!b->f_collisions[idx].is_exit_collision)
-					ship_apply_local_damage(&Objects[target], &Objects[b->objnum], &b->f_collisions[idx].cinfo.hit_point_world, beam_get_ship_damage(b, &Objects[target]), b->f_collisions[idx].quadrant);
+					real_damage = beam_get_ship_damage(b, &Objects[target]) * damage_time_mod;
+					ship_apply_local_damage(&Objects[target], &Objects[b->objnum], &b->f_collisions[idx].cinfo.hit_point_world, real_damage, b->f_collisions[idx].quadrant);
```
asarium: You forgot to add braces to this if statement which means that this line will now be executed even for exit collisions which is not the intended behavior.

code/weapon/beam.cpp
```diff
@@ -3294,16 +3289,17 @@ void beam_handle_collisions(beam *b)
...
 				// if this is the first hit on the player ship. whack him
-				if(do_damage)
-				{
-					// Goober5000 - AGH!  BAD BAD BAD BAD BAD BAD BAD BAD bug!  The whack's hit point is in *local*
-					// coordinates, NOT world coordinates!
-					beam_apply_whack(b, &Objects[target], &b->f_collisions[idx].cinfo.hit_point);
-				}
+				// Goober5000 - AGH!  BAD BAD BAD BAD BAD BAD BAD BAD bug!  The whack's hit point is in *local*
+				// coordinates, NOT world coordinates!
+				beam_apply_whack(b, &Objects[target], &b->f_collisions[idx].cinfo.hit_point);
```
asarium: Does `beam_apply_whack` handle frame time scaling correctly? If it assumes that it is only called every `BEAM_DAMAGE_TIME` milliseconds then this will cause some serious issues at high FPS.

Issue #1541, opened 2017-12-22
FSO should only chdir on Mac if launched via Finder (launchd)


chief1983:
In order to make platform behavior a little more consistent, and allow FSO to work with Knossos without introducing any other special behavior, FSO should only chdir to the app's folder when launched via launchd (Finder, Spotlight, open, etc).  Being run by other apps such as launchers or terminal should use the existing working directory and trust that it was set correctly by the environment (necessary for Knossos since the apps live outside of the data directories).

chief1983:
https://stackoverflow.com/questions/25011962/macosx-10-9-detecting-if-an-executable-is-launched-via-launch-services

chief1983:
Should be fixed in the next nightly.

PR #1550, opened 2017-12-30
Only change directory on MacOS when parent process path is /sbin/launchd.

chief1983:
Fix for #1541.  Tested to work when run via wxLauncher, Finder, open, and terminal after entering my FS2 data directory.  Please inspect for string and any necessary const adjustments, as again, this definitely isn't my daily language, and the gist of the code was basically from the internet.


chief1983:
This has been confirmed to fix the issue with Knossos not being able to launch FSO on Macs.

PR #1557, opened 2018-01-03
Use correct attachment for depth/stencil buffer

### LINKED ISSUE #1555: fs2_open.log is flooded by messages OpenGL Debug: Source:OpenGL	Type:Other	ID:65536	Severity:Medium	Message:. Framebuffer object Scene framebuffer is unsupported because the depth and stencil attachments are mismatched.
```
c585314ff12c24456c8937b3ff7c7962e4b42526 is the first bad commit
commit c585314ff12c24456c8937b3ff7c7962e4b42526
Author: asarium <asarium@gmail.com>
Date:   Fri Mar 24 15:48:43 2017 +0100

    Use a uniform buffer object for storing deferred lighting data
    
    This converts the deferred lighting shader to use uniform buffers for
    the dynamic data sent to the GPU while rendering the lighting effects.
    This doesn't affect performance but the deferred lighting code is a
    nicely isolated piece of rendering code where I could test how the
    rendering code could use the uniform buffers.
    
    This is related to #910.

:040000 040000 43dfb3e826c3218d20922224ba0239dad09cf464 45212659ef7deda841bb0cdf
```

Nvidia drivers (384.98 on Debian and 388.91 on Windows)

Log has 24554 lines and 23396 is about Framebuffer unsupported. :P

asarium:
This should be the correct way to set the combined depth/stencil buffer.
This has not been tested yet since I do not have a GPU which causes
these issues.

This fixes #1555.

### DIFF
--- a/code/graphics/opengl/gropengldeferred.cpp
+++ b/code/graphics/opengl/gropengldeferred.cpp
@@ -97,8 +97,7 @@ void gr_opengl_deferred_lighting_finish()
 	opengl_shader_set_current( gr_opengl_maybe_create_shader(SDR_TYPE_DEFERRED_LIGHTING, 0) );
 
 	glFramebufferTexture2D(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, Scene_luminance_texture, 0);
-	glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_RENDERBUFFER, Scene_stencil_buffer);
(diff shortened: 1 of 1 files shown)


The-E:
This fixes #1555 for me.

SamuelCho:
Are you sure you want to attach the stencil buffer in the first place? We don't use stencil tests for deferred lighting anymore. We want the scene depth texture, not the depth stencil render buffer after we're finished with deferred lighting. Soft particles read from the depth texture. If we're not writing to that anymore, did anybody check if soft particles still work with this change?

The-E:
I took a closer look, and these changes have deleterious effects for the godray effect (soft particles seem unaffected)

tomimaki:
[Definitely something is wrong. Yes, that's Ulysses from F3 lab, where I was before loaded mission.](https://i.imgur.com/yeZLbR3.png)

PR #1565, opened 2018-01-05
Add GCC 7 configuration for Travis CI

asarium:
There are a few new warnings in GCC7 so this makes sure that no warning
gets missed by our CI setup. The new configuration should be updated to
the newest GCC version whenever possible.


asarium:
This should be working now. I needed to fix some warnings inside `stb_image.h` so there are some updates in there.

PR #1569, opened 2018-01-10
Upgrade the included SDL libs to 2.0.7.

chief1983:
Stripped out the config headers that I think are only used for reconfiguring, should be everything we need.


asarium:
You need to update the has at the beginning of `lib/prebuilt.cmake` to `0bab5a328b34327cd6d518febc1855e8796c8050` so that the updated mac libraries are also used by these changes.

asarium:
I don't know why but Travis occasionally fails to install the build dependencies. I restarted the builds which should fix the Travis CI status.

PR #1570, opened 2018-01-11
set default volumes on pilot create

niffiwan:
previously newly created pilots inherited the volumes from the last
pilot


niffiwan:
note to self: fix undeclared identifier error

PR #1574, opened 2018-01-12
Add screen space decal rendering system

asarium:
This allows to render screen space decals into the deferred lighting
buffers no matter how complex the underlying geometry is.

Currently there is only limited support for using this in the engine.
Only weapon impact decals are supported at the moment for testing
purposes.


SamuelCho:
This is cool. It looks good to me. Does the decal shader do anything special to readjust the directional lighting? Or will we have to later move directional light to the deferred rendering stage so that directional lights can effect screenspace decals?

asarium:
I'll do the directional lighting conversion in a separate PR then.

PR #1585, opened 2018-01-26
Use mission backgrounds in the F3 lab

The-E:
As the title says, this allows users to look at mission backgrounds in the F3 lab.
This alters the lab camera controls:
Hold LMB + Mouse movement rotates the ship
Hold RMB + Mouse movement rotates the camera
Hold Shift + Hold LMB + Mouse movement up/down moves the camera towards/away from the ship
Things that do not work at the moment:

- [x] 1. When no background is selected, lighting is turned off.
- [x] 2. When a background with a skybox is used, using the material overrides will affect the skybox as well
- [ ] 3. Weapon rendering is disabled
- [x] 4. Animated textures (this includes effects like thrusters) are not advanced, meaning that they're stuck on the first frame of animation
- [x] 5. Several of the render flags do not work correctly
- [ ] 6. Shadows do not seem to work


chief1983:
Is it possible for this to be bound to a different hotkey than the F3 lab, and have both exist side by side for now as this one is developed?

The-E:
Sure?

Not sure why you'd want to have both side-by-side though.

chief1983:
I don't mean rendered side by side I just mean they could both coexist in master instead of having this PR creep along even as it passes the F3 lab feature-wise.

The-E:
Apart from not being able to show weapons, there's already as much feature parity as I feel is necessary.

PR #1586, opened 2018-01-27
Add a workaround for clipping planes on NVIDIA

### LINKED ISSUE #1579: Dynamic Lights flickering if Deferred Lighting and Soft Shadows enabled simultaneously.
First noticed this in the F3 lab. Click on the image for a video (and ignore the people talking).
[](https://www.youtube.com/watch?v=x2Hpi-1cHpI)
Other symptoms include vertical and horizontal lines marking an area of the screen where the mouse light does not affect the model. The light flickering does occur in game with things like beam cannon lights. Already verified that this occurs before GGX got merged in, so it's a little older than that.

To reproduce, simply have both Deferred Lighting and Soft Shadows enabled simultaneously.

asarium:
This will enable a workaround for Nvidia GPUs which will disable the new clipping optimization since it caused issues with those GPUs.

This fixes #1579.

### DIFF
--- a/code/def_files/main-f.sdr
+++ b/code/def_files/main-f.sdr
@@ -117,6 +117,11 @@ in float fragFogDist;
 #ifdef FLAG_ANIMATED
 uniform sampler2D sFramebuffer;
 #endif
+#ifdef WORKAROUND_CLIPPING_PLANES
+ #ifdef FLAG_TRANSFORM
+in float fragNotVisible;
+ #endif
+#endif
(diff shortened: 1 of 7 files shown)


chief1983:
I did confirm that this restores the behavior for my GTX 970 to the Dec 12 nightly, as in it solves both the 5-second flicker and the light/dark plane that passes over the ships in the F3 lab.

tomimaki:
Workaround works for my GT 620M.

PR #1588, opened 2018-02-03
Fix for team-colors not looking correctly

The-E:
Move the SRGB adjustment for diffuse and spec down a bit and add teamcolors to the spec map.


SamuelCho:
I don't think it's a good idea to do gamma correction outside of the gammaSpec conditional because if we do it for legacy -shine specular maps, it's not going to look very good.

The-E:
I reverted the SRGB adjustment lines and redone the logic for the color mixing.

Issue #1587, opened 2018-02-03
Default lighting values not aesthetically pleasing


The-E:
The default lighting values embedded in the exe, which for reference are:

- Directional lights: 1.0
- Tube lights: 1.0
- Point lights: 1.0
- Ambient factor: 128
- Bloom: 75

...are kinda bad. They produce images that are rather flat; in some circumstances, so flat in fact that shadows get swallowed whole by ambient lighting or ship self-illumination. 

Do note that, at the moment, the engine exposes two values (OGL Spec and specular exponent) which do nothing. They should thus not be considered in this discussion, and will be removed in a forthcoming nightly.

So, let's get to making better ones. Reducing bloom is probably the easiest of these; a value of 25 to 30 should produce an effect that is pleasing while not overwhelming.

For reference, here's a playthrough of SM1-09 with the default settings:
https://youtu.be/S42eaMJ50-E

And here's the same mission with -no_emissive_light:
https://youtu.be/_KB0psGhEd0

And again, with -no_emissive_light and the mediavps enabled:
https://youtu.be/IevPmEJxYXA

FrikgFeek:
There's also an interesting way that the lighting interacts with PBR and non-PBR assets. PBR assets will have really dark "main" textures which are the only ones ambient light is applied to(well, it also applies to team colours and added decals). This means -ambient_factor means basically fuck all for PBR models.

Now, they'll also have bright -reflect textures, like non-PBR models with their -shine textures. This is the only part spec_static actually applies to. Now, static lighting does seem to apply to diffuse but it isn't controlled by spec_static. 

What this means in practice is that PBR models will mostly be affected by -spec_static while non-PBR ones will mostly be affected by -ambient_factor.

In my opinion both ambient_factor and spec_static should apply to both diffuse and spec. This way modders(and players) would be able to achieve a somewhat similar lighting look for both PBR and non-PBR models which is what a vast majority of mods currently have.

Additionally, for missions where $Ambient light level: is set to 0 or 7895160(120 on all 3 channels, the default setting) or lower any ambient factor below 70.2 will be identical to 0 as it will result in a "final" ambient value below 0.02. This mission ambient setting is used in all MediaVPs missions and a vast number of mod missions as well. Just a note for tweaking defaults.


I've created an imgur album with pictures of what PBR and non PBR ships look like under different lighting settings to demonstrate my point. 
https://imgur.com/gallery/v8IRA

Goober5000:
[...]

PR #1593, opened 2018-02-05
Remove unused cmdline arguments

The-E:
While discussing the various lighting-related commandline arguments (see #1587), it became clear that two of them (-spec_exp and -ogl_spec) don't actually do anything at the moment. Removing them seems like a good idea.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -308,7 +308,6 @@ int Cmdline_use_last_pilot = 0;
 // Graphics related
 cmdline_parm fov_arg("-fov", "Vertical field-of-view factor", AT_FLOAT);					// Cmdline_fov  -- comand line FOV -Bobboau
 cmdline_parm clip_dist_arg("-clipdist", "Changes the distance from the viewpoint for the near-clipping plane", AT_FLOAT);		// Cmdline_clip_dist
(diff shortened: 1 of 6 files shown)


chief1983:
I guess I am curious about that because if they aren't functional but never intentionally disabled, this would seem to highlight an unintentional effect of one of our graphics upgrades over the last few years.  Perhaps they were intended to remain functional but their implementation in the new pipeline was omitted by mistake?

asarium:
Only @SamuelCho can answer that question conclusively but my guess is that the parameters either became unnecessary of maybe even impossible to implement when we switched to the PBR pipeline.

The-E:
The thing is, and this is something we should also discuss in #1587, what sort of customizability we want the lighting engine to have and how to expose it. I think that, rather than repurposing these old parameters, we should add new ones for whatever functionality we decide to implement. For now, cleaning up the code should take priority.

SamuelCho:
Yeah I've had trouble figuring out how those parameters would translate to the new specular model. Specular power is essentially gloss/inverse roughness. I thought about scaling ogl_spec against 255.0 and using that 0.0-1.0 value for default gloss. But I don't know what spec_exp would correspond to. I could do a bit of code rewinding to figure that out.

MageKing17:
Should we maybe deprecate them to be on the safe side?

PR #1594, opened 2018-02-05
Remove the -missile_lighting commandline parameter

The-E:
This also came up while digging around the lighting-related commandline parameters. In the previosu render pipeline, excluding missiles from lighting made sense; there was a lot of cost associated with it. Now, things are reversed: Excluding models from lighting is more expensive than lighting them, thus this parameter no longer serves any useful purpose.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -7347,7 +7347,7 @@ void weapon_render(object* obj, model_draw_list *scene)
 			uint render_flags = MR_NORMAL|MR_IS_MISSILE|MR_NO_BATCH;
 
 			if (wip->wi_flags[Weapon::Info_Flags::Mr_no_lighting])
-				render_flags &= MR_NO_LIGHTING;
+				render_flags |= MR_NO_LIGHTING;
 
 			if (wip->wi_flags[Weapon::Info_Flags::Transparent]) {
 				render_info.set_alpha(wp->alpha_current);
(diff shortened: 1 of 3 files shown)


chief1983:
This probably needs to be added to deprecated/removed flags list for a release or two so that it does not cause the engine to throw an error when it is hanging around in the launcher.

code/weapon/weapons.cpp
```diff
@@ -7344,10 +7344,10 @@ void weapon_render(object* obj, model_draw_list *scene)
 		{
 			model_render_params render_info;
 
-			uint render_flags = MR_NORMAL|MR_IS_MISSILE|MR_NO_LIGHTING|MR_NO_BATCH;
+			uint render_flags = MR_NORMAL|MR_IS_MISSILE|MR_NO_BATCH;
 
-			if (Cmdline_missile_lighting && !(wip->wi_flags[Weapon::Info_Flags::Mr_no_lighting]))
-				render_flags &= ~MR_NO_LIGHTING;
+			if (wip->wi_flags[Weapon::Info_Flags::Mr_no_lighting])
+				render_flags &= MR_NO_LIGHTING;
```
MageKing17: ...Why are we clearing the render flags here? Shouldn't it be `|=`?

The-E: My only defense is that I hate binary math

PR #1597, opened 2018-02-09
Fix off-by-one error in stuff_int_list

asarium:


MageKing17:
Don't forget the `count + 1` a couple lines down.

asarium:
Should be fixed now.

PR #1599, opened 2018-02-10
Make sure directional lights are always rendered

### LINKED ISSUE #1598: Directional lights may not be rendered under certain conditions
This was found by FrikgFeek:

When a lot of point lights are present, directional lights (i.e. suns) are not getting rendered.
This can be demonstrated using this test mission: https://cdn.discordapp.com/attachments/223511391531827200/411726056169537557/Turn_off_the_sun.fs2
With this edited weapons.tbl: https://cdn.discordapp.com/attachments/223511391531827200/411726079024431114/weapons.tbl

Now, the code is actually supposed to make sure this doesn't happen: Directional lights are always supposed to be rendered.
The problem is in light_add_directional (lighting.cpp, line 201). When the background rendering code tries to add lights, the Lights array is already at capacity when lots of point lights are flying around.

The-E:
This fixes #1598.

### DIFF
--- a/code/graphics/opengl/gropengldeferred.cpp
+++ b/code/graphics/opengl/gropengldeferred.cpp
@@ -73,7 +73,7 @@ void gr_opengl_deferred_lighting_end()
 	glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_FALSE);
 }
 
-extern light Lights[MAX_LIGHTS];
+extern SCP_vector<light> Lights;
 extern int Num_lights;
 extern float static_point_factor;
 extern float static_light_factor;
@@ -113,10 +113,10 @@ void gr_opengl_deferred_lighting_finish()
 	glClearColor(0.0f, 0.0f, 0.0f, 1.0f);
 	glClear(GL_COLOR_BUFFER_BIT);
 
-	light lights_copy[MAX_LIGHTS];
-	memcpy(lights_copy, Lights, MAX_LIGHTS * sizeof(light));
+	SCP_vector<light> lights_copy(Lights);
 
-	std::sort(lights_copy, lights_copy+Num_lights, light_compare_by_type);
+	std::sort(lights_copy.begin(), lights_copy.end(), light_compare_by_type);
+	lights_copy.resize(MAX_LIGHTS);
 
 	using namespace graphics;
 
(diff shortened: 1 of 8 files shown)


code/graphics/opengl/gropengldeferred.cpp
```diff
@@ -127,9 +127,9 @@ void gr_opengl_deferred_lighting_finish()
 	{
 		GR_DEBUG_SCOPE("Build buffer data");
 
-		for (int i = 0; i < Num_lights; ++i) {
-			light* l = &lights_copy[i];
-			if (l->type != LT_CONE && l->type != LT_POINT && l->type != LT_TUBE) {
+		for (auto l : lights_copy) {
```
asarium: The type should be `auto&` to avoid copying the `light' structure.

code/graphics/opengl/gropengldeferred.cpp
```diff
@@ -200,18 +200,17 @@ void gr_opengl_deferred_lighting_finish()
 		GR_DEBUG_SCOPE("Render light geometry");
 
 		size_t element_index = 0;
-		for (int i = 0; i < Num_lights; ++i) {
+		for (auto l : lights_copy) {
```
asarium: Same as above.

code/model/modelrender.cpp
```diff
@@ -582,9 +582,9 @@ void model_draw_list::init()
 {
 	reset();
 
-	for ( int i = 0; i < Num_lights; ++i ) {
-		if ( Lights[i].type == LT_DIRECTIONAL || !Deferred_lighting ) {
-			Scene_light_handler.addLight(&Lights[i]);
+	for ( auto l : Lights) {
```
asarium: This should also use `auto&`.

code/graphics/shadows.cpp
```diff
@@ -360,16 +360,12 @@ matrix shadows_start_render(matrix *eye_orient, vec3d *eye_pos, float fov, float
 	if(Static_light.empty())
 		return vmd_identity_matrix; 
 	
-	light *lp = *(Static_light.begin());
-
-	if ( lp == NULL ) {
-		return vmd_identity_matrix;
-	}
+	auto lp = *(Static_light.begin());
```
asarium: This should use `vector::front` and `auto&`.

code/graphics/shadows.cpp
```diff
@@ -406,12 +402,12 @@ void shadows_render_all(float fov, matrix *eye_orient, vec3d *eye_pos)
...
 	}
 
-	light *lp = *(Static_light.begin());
-
-	if( !Cmdline_shadow_quality || !lp ) {
+	if (!Cmdline_shadow_quality) {
 		return;
 	}
 
+	auto lp = *(Static_light.begin());
```
asarium: Same as above.

PR #1600, opened 2018-02-10
Add persistent buffer mapping support for uniform data

asarium:
This reworks how the engine handles uniform buffers and introduces support for persistently mapped buffers if supported by the OpenGL context.

The rework of the uniform managing code introduces a single large uniform buffer that will be used by all uniform blocks for getting their data. This code follows the usual principle where it allocates three segments which are used in a round-robbin sequence to minimize synchronization with the GPU.

The persistent mapping part simply changes how the data is uploaded to the GPU by replacing a `glBufferSubData` with a `memcpy`.

There are no significant changes in performance since uniform buffers are not used extensively by FSO yet but this should make sure that uniform updates do not become a bottleneck in the future.


code/globalincs/pstypes.h
```diff
@@ -528,12 +528,18 @@ namespace std
...
 	template<typename T, typename U>
-	void *memmove_if_trivial_else_error(T *memmove_dest, U *memmove_src, size_t count)
+	inline void *memmove_if_trivial_else_error(T *memmove_dest, U *memmove_src, size_t count)
 	{
 		static_assert(trivial_check<T>::value, "memmove on non-trivial object T");
 		static_assert(trivial_check<U>::value, "memmove on non-trivial object U");
 		return ptr_memmove(memmove_dest, memmove_src, count);
 	}
+
+	// Not really needed but else clang things ptr_memmove isn't used
```
Goober5000: "thinks" is misspelled

asarium: Fixed.

The-E:
This looks good to me; it probably shouldn't be merged before the next release though.

asarium:
Agreed, this is way too much code to merge in the RC phase. I will merge this after the final has been built.

PR #1605, opened 2018-02-14
Negative damages no longer heal subsystems

Flip-Sandwich:
The +additive armor type no longer heals a subsystem when you shoot a subsystem with a low damage weapon and the weapon has less damage than the damage subtraction


TRBlount:
This needs to be handled differently. The current changes stand to break at least one major mod (JAD).

PR #1608, opened 2018-02-15
Use an enum class to mark light types

The-E:
This also fixes an issue where an uninitialized variable was used as an index

### DIFF
--- a/code/graphics/light.cpp
+++ b/code/graphics/light.cpp
@@ -62,7 +62,7 @@ void FSLight2GLLight(light* FSLight, gr_light* GLLight) {
 	GLLight->Specular.xyzw.z = FSLight->spec_b * FSLight->intensity;
 	GLLight->Specular.xyzw.w = 1.0f;
 
-	GLLight->type = FSLight->type;
+	GLLight->type = static_cast<int>(FSLight->type);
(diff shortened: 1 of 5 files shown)


code/graphics/light.cpp
```diff
@@ -134,11 +134,11 @@ void FSLight2GLLight(light* FSLight, gr_light* GLLight) {
 		break;
 	}
 
-	case LT_CONE:
+	case Light_Type::Cone:
 		break;
 
 	default:
-		Int3();
+		Error(LOCATION, "Unknown light type in FSLight2GLLight. Expected was 0, 1, 2 or 3, we got %i", FSLight->type);
```
asarium: I think `FSLight->type` needs to be cast to an `int` to make this type safe.

code/lighting/lighting.h
```diff
@@ -27,8 +27,15 @@
 #define LT_TUBE			2		// A tube light, like a fluorescent light
 #define LT_CONE			3		// A cone light, like a flood light
 
+enum class Light_Type : int {
```
asarium: The enum values need to be explicitly specified here since you use the `int` value of this type above.

PR #1609, opened 2018-02-16
Fixes beam behaviour with additive/cap/etc damage types

Flip-Sandwich:
Intended armor type behaviour: Subtracting 500 damage from a beam with a tabled damage of 3100 causes the beam to do ~0.85x damage

Actual armor type behaviour: Subtracting 500 damage from a beam with a tabled damage of 3100 causes the beam to do 0 damage on low time accelerations and 1x damage at high time accelerations, as the damage is subtracted every frame and beam damage is recalculated according to frametimes each frame

Hopefully fixed armor type behaviour: Subtracting 500 damage from a beam with a tabled damage of 3100 causes the beam to do ~0.85x damage

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -17724,8 +17724,7 @@ int calculation_type_get(char *str)
 }
 
 //STEP 4: Add the calculation to the switch statement.
-float ArmorType::GetDamage(float damage_applied, int in_damage_type_idx, float diff_dmg_scale)
-{
+float ArmorType::GetDamage(float damage_applied, int in_damage_type_idx, float diff_dmg_scale, int is_beam) {
 	// Nuke: If the weapon has no damage type, just return damage
 	if (in_damage_type_idx < 0) {
 		// multiply by difficulty scaler now, since it is no longer done where this is called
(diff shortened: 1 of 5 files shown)


code/ship/ship.cpp
```diff
@@ -17811,6 +17810,16 @@ float ArmorType::GetDamage(float damage_applied, int in_damage_type_idx, float d
 				constant_val = 0.0f;
 				curr_arg = constant_val;
 			}
+
+			//face: terrible hack to work consistently with beams and additive damages
+			if (is_beam && adtp->Calculations[i] != AT_TYPE_MULTIPLICATIVE
+				&& adtp->Calculations[i] != AT_TYPE_EXPONENTIAL
+				&& adtp->Calculations[i] != AT_TYPE_EXPONENTIAL_BASE
+				&& adtp->Calculations[i] != AT_TYPE_STORE
+				&& adtp->Calculations[i] != AT_TYPE_LOAD) {
```
MageKing17: I think a whitelist of armor types that need the hack makes more sense than a blacklist of the types that don't.

code/ship/ship.cpp
```diff
@@ -17811,6 +17810,16 @@ float ArmorType::GetDamage(float damage_applied, int in_damage_type_idx, float d
...
+				curr_arg = curr_arg * (flFrametime * 1000.0f) / i2fl(170);
```
MageKing17: Beam damage gets calculated based on `BEAM_DAMAGE_TIME`, but here you've hardcoded the magic number of 170ms. This both inconsistent and poor form.

Flip-Sandwich:
Gotcha. Properly done version pushed.

Issue #1615, opened 2018-02-18
More debug_filter weirdness


The-E:
As reported by Axem on discord:

```
Axem: 1: put -s in front of everything except general
Axem: 2: run debug until you hit a game halting error and exit(edited)
Axem: FSO will then resave the debug filter with only +General on game exit
Axem: 3: rerun FSO and the entire debug filter gets reset
Axem: if you do 2 and just exit FSO right away without encountering errors, it will not alter the debug filter
Axem: (which i guess is how i probably checked it before and thought "oh this works now!")
```

I've tested this myself, and can reproduce the error; I am not entirely sure how it can happen though. It seems that, when the error dialog pops up (in my case, a parse error induced by putting gibberish into a -shp.tbm), the OutwndFilter vector gets truncated to its first element (which, in almost all cases, will be "+General"). Since there's no point in the code where that Vector gets cleared, I can't really explain what's happening.

The-E:
Okay, further investigation has revealed the following causal chain:

1. A fatal error occurs, prompting the engine to throw up a messagebox with Exit and Debug options
2. User pushes the "Exit" button
3. This calls exit(1)
4. During exit processing, statically declared vectors (like OutwndFilter and GL_buffer_objects (which will become relevant later)) are cleared
5. gr_opengl_delete_buffer is called, presumably while going through gr_close
6. AssertMessage calls outwnd_print, which sees that no "General" filter has been declared, sets up a new one, and saves the OutwndFilter vector to debug_filter.cfg

PR #1616, opened 2018-02-18
Do not try to delete opengl buffers when we have no idea what they are

The-E:
While working through the atexit table, this function somehow got called, leading to the debug_filter.cfg being altered, as explained in #1615. While this does fix one immediate cause of the behaviour in #1615, it opens up a few more questions: Why gr_opengl_delete_buffer is called at this point, for example.

### DIFF
--- a/code/graphics/opengl/gropengltnl.cpp
+++ b/code/graphics/opengl/gropengltnl.cpp
@@ -293,6 +293,8 @@ void gr_opengl_update_buffer_data_offset(int handle, size_t offset, size_t size,
 
 void gr_opengl_delete_buffer(int handle)
 {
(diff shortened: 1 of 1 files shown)


MageKing17:
It appears it's being called by the deconstructor of `GPUMemoryHeap`.

asarium:
This seems like a problem with the deinitialization order of the memory heap and the OpenGL buffer vector. In your case the buffer vector is probably cleared before the heap pointers are freed so when the heap tries to free its buffer handle the buffer vector is already empty.

I'm not sure how this can be solved properly except by somehow making sure that the OpenGL data structures are cleared after the heap data structures.

PR #1622, opened 2018-02-23
Use deferred shader for applying directional lights

asarium:
This moves all lighting operations into the deferred shading phase of
the rendering pipeline. Previously, directional lights were applied to
all objects while they were being rendered. This was probably done to
avoid doing multiple full screen passes later but caused some problems
since the color information in the G-buffers was not correct for doing
later lighting operations.

This was especially noticeable in dark scenes where point lights would
appear to ignore the color of the rendered objects. This should be fixed
with these changes.

I did some performance testing and to my surprise this actually improves
performance in very demanding scenes. My guess would be that applying
the shadows while rendering the objects caused a lot of overdraw which
is not the case anymore since shadows are only computed for visible
pixels. In simple scenes the opposite happens which leads to decreased
performance in those cases but in my opinion the improvement in
demanding scenes outweighs this performance decrease.

There are still some issues with flickering lights so until I can fix that this
should not be merged.

### DIFF
--- a/code/def_files/data/effects/deferred-clear-f.sdr
+++ b/code/def_files/data/effects/deferred-clear-f.sdr
@@ -2,10 +2,12 @@ out vec4 fragOut0;
 out vec4 fragOut1;
 out vec4 fragOut2;
 out vec4 fragOut3;
+out vec4 fragOut4;
 void main()
 {
 	fragOut0 = vec4(0.0, 0.0, 0.0, 1.0); // color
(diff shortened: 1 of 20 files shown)


asarium:
I fixed the flickering lights issue. The cause was that the deferred lighting code sorted the lights using the standard `std::sort` function which is not stable. This could lead to a mismatch between the light the shadows used while rendering and the first light that was rendered by the deferred renderer.

I fixed this by using `std::stable_sort`.

PR #1625, opened 2018-02-25
Fixes electronics flag behavior with shockwaves

Flip-Sandwich:
The electronics flag only worked if your weapon had a shockwave speed of 0. If it had a shockwave speed > 0 you had to have an AoE.

This fixes that

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -6007,12 +6007,7 @@ void weapon_do_area_effect(object *wobjp, shockwave_create_info *sci, vec3d *pos
 
 	}	// end for
 
-	// if this weapon has the "Electronics" flag set, then disrupt subsystems in sphere
-	if ( (other_obj != NULL) && (wip->wi_flags[Weapon::Info_Flags::Electronics]) ) {
-		if ( other_obj->type == OBJ_SHIP ) {
-			weapon_do_electronics_effect(other_obj, pos, Weapons[wobjp->instance].weapon_info_index);
-		}
-	}
+
 }
 
 //	----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


code/weapon/weapons.cpp
```diff
@@ -5980,7 +5980,7 @@ void weapon_do_area_effect(object *wobjp, shockwave_create_info *sci, vec3d *pos
 		case OBJ_SHIP:
 			// If we're doing an AoE Electronics blast, do the electronics stuff (unless it also has the regular "electronics"
 			// flag and this is the ship the missile directly impacted; then leave it for the regular code below) -MageKing17
-			if ( (wip->wi_flags[Weapon::Info_Flags::Aoe_Electronics]) && !((objp->flags[Object::Object_Flags::Invulnerable]) || ((objp == other_obj) && (wip->wi_flags[Weapon::Info_Flags::Electronics]))) ) {
+			if ( (wip->wi_flags[Weapon::Info_Flags::Aoe_Electronics]) && !((objp->flags[Object::Object_Flags::Invulnerable]) ) ) {
```
MageKing17: Removing this check means that a weapon with both electronics flags and no shockwave will apply electronics effects twice.

code/weapon/weapons.cpp
```diff
@@ -6008,11 +6008,7 @@ void weapon_do_area_effect(object *wobjp, shockwave_create_info *sci, vec3d *pos
 	}	// end for
 
 	// if this weapon has the "Electronics" flag set, then disrupt subsystems in sphere
```
MageKing17: You've removed the code, but not the comment that goes with it.

code/weapon/weapons.cpp
```diff
@@ -6263,6 +6259,13 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
 		emp_apply(&weapon_obj->pos, wip->shockwave.inner_rad, wip->shockwave.outer_rad, wip->emp_intensity, wip->emp_time, (wip->wi_flags[Weapon::Info_Flags::Use_emp_time_for_capship_turrets]) != 0);
 	}	
 
+	// check if this is an electronics weapon
+
+	if (wip->wi_flags[Weapon::Info_Flags::Electronics] && (other_obj != NULL)) {
```
MageKing17: This is kind of a minor thing, but the conditional was probably written the other way around because a null pointer check is faster than a flag check, and you generally want to check the fastest conditions first.

code/weapon/weapons.cpp
```diff
@@ -6263,6 +6257,12 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
 		emp_apply(&weapon_obj->pos, wip->shockwave.inner_rad, wip->shockwave.outer_rad, wip->emp_intensity, wip->emp_time, (wip->wi_flags[Weapon::Info_Flags::Use_emp_time_for_capship_turrets]) != 0);
 	}	
 
+	// if this weapon has the "Electronics" flag set, then disrupt subsystems in sphere
+	if  (other_obj != NULL)) {
```
The-E: There is an extra ')' here that needs removal

Flip-Sandwich:
Try now.

PR #1632, opened 2018-03-02
Add normal map rendering to decals

asarium:
This changes the decal system so that it is able to modify the normal
map of the scene. I tested this with some existing normal maps but I am
not entirely sure if the tangent matrix I construct uses the right
values so some additional testing with normal maps made specifically for
decals would be good.

This also fixed the issues introduced by the changes to how directional
lighting was handled. Now a decal can have a diffuse and a glow map. The
diffuse map is used to change the values of the color buffer and also
supports ambient lighting. The glow map renders into the emissive buffer
and should be used for effects that should not receive lighting.


chief1983:
Didn't we recently get height map shaders working?  I would think use for the damage decals would be even more impressive of a look, if it's as easy to implement.

PR #1633, opened 2018-03-03
Add SRGB adjustment to the team colors

The-E:


code/def_files/data/effects/main-f.sdr
```diff
@@ -293,10 +293,17 @@ void main()
  #ifdef FLAG_TEAMCOLOR
 	vec4 teamMask = vec4(0.0, 0.0, 0.0, 0.0);
 	teamMask = texture(sMiscmap, vec3(texCoord, float(sMiscmapIndex)));
-	vec3 base = max(base_color - vec3(0.5), vec3(0.0));
-	vec3 stripe = max(stripe_color - vec3(0.5), vec3(0.0));
-	baseColor.rgb += (base * teamMask.x) + (stripe * teamMask.y);
+	vec3 team_color = (base_color * teamMask.x) + (stripe_color * teamMask.y);
```
asarium: `base_color` and `stripe_color` are used without changes here but in the previous version `0.5` was subtracted from it. Could you explain why you made this change?

code/def_files/data/effects/main-f.sdr
```diff
@@ -293,10 +293,17 @@ void main()
...
+	vec3 team_color_glow = (base_color * teamMask.b) + (stripe_color * teamMask.a);
+	#ifdef FLAG_HDR
+	team_color = pow(team_color, vec3(SRGB_GAMMA));
+	team_color_glow = pow(team_color_glow, vec3(SRGB_GAMMA)) * 3.0f;
```
asarium: The last factor (`3.0`) should be explained to clarify why it is there. I assume it is for applying the glow effect to team color glows but that is not clear from the code. Also, there is a global glow color intensity value available already. Maybe that should be used here to keep everything consistent.

code/def_files/data/effects/main-f.sdr
```diff
@@ -293,10 +293,17 @@ void main()
...
+	#endif
+
+	baseColor.rgb += team_color;
 	baseColor.rgb = max(baseColor.rgb, vec3(0.0));
+	specColor.rgb += team_color;
+	specColor.rgb = max(specColor.rgb, vec3(0.03f));
```
asarium: There are a few instances in the other changes but GLSL does not require the `f` for specifying a `float` value.

code/def_files/data/effects/main-f.sdr
The-E: It exists because of line 353:
```
 #ifdef FLAG_HDR
	glowColor = pow(glowColor, vec3(SRGB_GAMMA)) * 3.0f;
 #endif
```

Do not ask me why that exists. It just makes sense to me to apply it uniformly here. GLOW_MAP_INTENSITY is a multiplier applied to the result of everything that goes into getting the glowColor value; applying it twice seems wrong.

asarium: I didn't know that glow maps were scaled even more in HDR mode. Thank you for clarifying this.

code/def_files/data/effects/main-f.sdr
The-E: That was actually an oversight. I was experimenting around with it and this change snuck in by accident.

code/def_files/data/effects/main-f.sdr
```diff
@@ -293,10 +293,17 @@ void main()
  #ifdef FLAG_TEAMCOLOR
 	vec4 teamMask = vec4(0.0, 0.0, 0.0, 0.0);
 	teamMask = texture(sMiscmap, vec3(texCoord, float(sMiscmapIndex)));
-	vec3 base = max(base_color - vec3(0.5), vec3(0.0));
-	vec3 stripe = max(stripe_color - vec3(0.5), vec3(0.0));
-	baseColor.rgb += (base * teamMask.x) + (stripe * teamMask.y);
+	vec3 team_color = ((base_color - vec3(0.5)) * teamMask.x) + ((stripe_color - vec3(0.5)) * teamMask.y);
```
asarium: I think it would be better to keep using the original `base` and `stripe` variables to reduce the amount of recompuatation of the adjusted color values. Also the new version is missing the `max` function call from the original version.

The-E: That is intentional. That max() call was responsible for making team colors not show up as strongly as they are intended to; I am not sure why this looked correct on the old renderer but doesn't here.

PR #1635, opened 2018-03-03
Handle emissive light properly in nebulas

asarium:
The emissive part of the lighting was not using the fog values properly
so ships didn't rendered right in nebulas. This fixes that by applying
the standard fog color transform to the emissive part of the scene.


niffiwan:
urgh - now has merge conflicts

asarium:
I have fixed the merge conflicts.

PR #1636, opened 2018-03-03
Don't use discard in main model shader

asarium:
The only remaining occurence of discard was for when blending was
disabled and the pixel of the texture had an alpha value less than 0.95.
That combination cannot happen anymore since all transparent textures
are handled using one of the other blend modes and all other textures
won't have an alpha value less than 0.95.

This has some minor performance benefits in GPU limited scenes.

### DIFF
--- a/code/def_files/data/effects/main-f.sdr
+++ b/code/def_files/data/effects/main-f.sdr
@@ -262,7 +262,6 @@ void main()
    }
  #endif
 	baseColor = texture(sBasemap, vec3(diffuseTexCoord, float(sBasemapIndex)));
-	if ( blend_alpha == 0 && baseColor.a < 0.95 ) discard; // if alpha blending is not on, discard transparent pixels
 	// premultiply alpha if blend_alpha is 1. assume that our blend function is srcColor + (1-Alpha)*destColor.
 	// if blend_alpha is 2, assume blend func is additive and don't modify color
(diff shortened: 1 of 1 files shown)


SamuelCho:
I'm not sure if this is a good idea. Transparencies in Freespace are kind of fucked in that a lot of legacy models don't separate out their opaques from their transparents. We try out best to separate out transparent and opaque draws as best as we can (down to sampling the texture per vert to figure out what triangles are transparent). But the problem is the opaque draws still have transparent triangles. We really don't want those filling up the z-buffer thereby blocking objects from drawing behind them. Even newer models have issues where a modeller/rigger forgot to put glass meshes into a -trans submodel. The Viper VIIE from Diaspora is one such model.

I agree with you that discards should be removed as much as possible. But this is going to make that problem pop up again. I just wonder if there's a way to conditionally not write to the depth buffer. I mean maybe we should check if I did remove transparent triangles from the submodels at load time but I highly doubt it. If you have any other solutions to this, I'm eager to hear it.

asarium:
I thought that the transparent pixel detection already handled separating out the transparent meshes but if that isn't true in all cases then my assumption from above also isn't true.

I can't think of anything that would get rid of that `discard` any other way though so it looks like that, at least for now, this can't be merged.

SamuelCho:
Thinking about it, I wonder if we can conditionally write a really low value to gl_FragDepth if there's a transparency in the opaque pass. That should allow any opaque objects to unconditonally render in front of them.

PR #1638, opened 2018-03-04
Reworks the lighting settings according to the discussion in #1634.

### LINKED ISSUE #1634: Lighting detail settings make no sense
I would like to propose a theory: The way the lighting detail options work right now make absolutely no sense.
As evidence, I present what the option currently does:

lighting 0:

- flat lighting for insignia via light_apply_rgb
- directional lights are enabled

lighting 1: 

- applies static lighting to insignia via light_apply_rgb by manually computing the light influence for a given set of insignia 

lighting 2:

- no difference to 1

lighting 3:

- normal/height/ambient/misc maps enabled
- all weapons cast light 
- debris lightning casts light

lighting 4+:

- explosions cast light

This, I think, could be improved. What we want, I think, is to give the user a tighter control over how many lights are added to a scene; by just switching on all weapon lights at 3 (and there being no explicit difference between detail level 1 and 2), there's room here to add some granularity. A reworked scheme could look like this:

lighting 0:

- render all directional lights
- use all lights in light_apply_rgb
- enable misc maps
- use simplified lighting model in the shaders

lighting 1: 

- enable normal/ambient mapping

lighting 2:

- enable height mapping
- render beam lights

lighting 3:

- render lights from explosions, warp
- use GGX lighting model

lighting 4:

- render lights from weapons
- render lights from glowpoints

Notes:
- Lighting from beams, explosions and warps is among the strongest lighting we currently encounter in the game. As such, it should be enabled way earlier than it is now.
- Normal/Ambient mapping has a dramatic effect on the overall look and, as these things go, is not that performance intensive (and is easily switched off via cmdline if it becomes an issue), so it should happen way sooner
- light_apply_rgb is a bad thing. I want to investigate how much effort it would be to rip the insignia rendering code out of the game and replace it with the decal rendering we got in there now (at which point, they'd be governed by some reworking of the model detail slider, I think)

Questions, comments, concerns?

The-E:
This fixes #1634


code/graphics/opengl/gropenglshader.cpp
```diff
@@ -336,6 +336,10 @@ static SCP_string opengl_shader_get_header(shader_type type_id, int flags, shade
 		sflags << "#define WORKAROUND_CLIPPING_PLANES\n";
 	}
 
+	if (Detail.lighting < 3) {
```
asarium: Do shaders get recompiled if the detail level is changed? If not then this would lead to inconsistent lighting effects until the game is restarted. Maybe a `gr_` function should be added which recompiles all existing shaders.

The-E: That's a good point, a shader recompile definitely needs to happen here.

code/menuui/optionsmenu.cpp
```diff
@@ -1374,12 +1398,21 @@ void options_detail_sliders_update()
...
-	Detail.lighting = Detail_sliders[gr_screen.res][LIGHTING_SLIDER].slider.pos;
+
+	// If the new lighting setting is above 3 and the old one was below or the reverse,
+	// we need to recompile all shaders we have to account for the changed lighting model.
+	
+	if (Detail.lighting != Detail_sliders[gr_screen.res][LIGHTING_SLIDER].slider.pos) {
+		Detail.lighting = Detail_sliders[gr_screen.res][LIGHTING_SLIDER].slider.pos;
+		shader_compile_status = false;
+		shader_compile_started = false;
+		popup_till_condition(recompile_shaders, POPUP_CANCEL, "Recompiling shaders");
```
asarium: This doesn't actually display a popup with changing text but since that would likely require adding a new popup type it's not really in the scope of these changes.

The-E: It does, although indirectly. The shader recompilation callback constructs a status string, which is then applied on-frame via recompile_shaders() calling popup_change_text.
It's a clunky solution, but in the absence of a new popup type specifically for situations like these, the best available.

asarium: `gr_recompile_shaders` doesn't return until all the shaders are compiled so effectively `recompile_shaders` is only called once and the display never updates (at least when I tested the changes it didn't).

PR #1639, opened 2018-03-04
Move gamma correction into common header

asarium:
This moves the constant definition and functions for converting from and
to linear color space into a single file that is included in all other
files.


The-E:
Wouldn't it make the code a bit more readable if you used the `#ifdef FLAG_HDR` stuff in here and turn these functions into nops if that flag isn't present?

asarium:
The flag isn't defined in all the files where this function is used so it can't be used in all cases.

The-E:
Fair enough

PR #1641, opened 2018-03-04
Load unloaded sounds when trying to determine sound length

asarium:
This can happen if a script tries to determine the length of a sound
before actually playing it. This fixes the assertion by loading the
sound when it is necessary.

### DIFF
--- a/code/gamesnd/gamesnd.cpp
+++ b/code/gamesnd/gamesnd.cpp
@@ -1322,7 +1322,10 @@ float gamesnd_get_max_duration(game_snd* gs) {
 	int max_length = 0;
 
 	for (auto& entry : gs->sound_entries) {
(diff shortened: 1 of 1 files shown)


chief1983:
Should this have really even been necessary?  We recently made the 'load mission game sounds during mission load' flag the default behavior, I would have thought any sound to be played in a mission should already be loaded at that point.  Is this because that flag was unaware of sounds that a script might play?  If so, is there a way to find out what sounds scripts would want to play and add that to the existing load-all-sounds code?  Or was this occurring outside of a mission?

asarium:
It occurred when loading a mission but I am not sure if that was before or after the mission sounds were loaded. The code is still needed in any case if the function is used outside a mission so these changes are still necessary.

PR #1652, opened 2018-03-10
Fix for 1623

### LINKED ISSUE #1623: Assertion: Unable to find triggered rotation for ship
This is more of a reminder to myself to fix this up, but here's the deal:
In modelanim.cpp:849, there's an assertion that is supposed to make the code safe. Of course, it doesn't, which is something 2015 me apparently didn't realize; in this case (and in the case of this same assertion used in other places), I should investigate whether we can correct this issue in-place. Right now, a release build hitting this assert would happily move on to do invalid vector accesses, which is obviously something we can't do.

The-E:
Replace this assert with a proper exit path + debug log print.

This fixes #1623

### DIFF
--- a/code/model/modelanim.cpp
+++ b/code/model/modelanim.cpp
@@ -405,8 +405,6 @@ void model_anim_submodel_trigger_rotate(model_subsystem *psub, ship_subsys *ss)
 	Assert( psub != NULL );
 	Assert( ss != NULL );
 	Assert( psub->flags[Model::Subsystem_Flags::Triggered] );
-	if (!psub || !ss || !psub->flags[Model::Subsystem_Flags::Triggered])
-		return;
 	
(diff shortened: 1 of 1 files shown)


code/model/modelanim.cpp
```diff
@@ -405,7 +405,13 @@ void model_anim_submodel_trigger_rotate(model_subsystem *psub, ship_subsys *ss)
 	Assert( psub != NULL );
 	Assert( ss != NULL );
 	Assert( psub->flags[Model::Subsystem_Flags::Triggered] );
-	Assert( ss->triggered_rotation_index >= 0 );
+	if (!psub || !ss || !psub->flags[Model::Subsystem_Flags::Triggered])
```
asarium: This check isn't really needed since it is already covered by the `Assert`s from above.

PR #1657, opened 2018-03-14
Add SEXP tree context menu building

asarium:
This adds displaying the correct context menu for the qtFRED SEXP tree
widget. The actions are not implemented yet.


asarium:
I also ported the code for the individual actions to Qt now. This makes the SEXP tree mostly functional except for variable support.

Issue #1653, opened 2018-03-10
Crash in F3 Lab after leaving mission.


tomimaki:
Load mission then quit, go to F3 Lab and load some model.
```
Thread 1 "fs2_open_3_8_1_" received signal SIGTRAP, Trace/breakpoint trap.
debug_int3 (file=0x555556117ae8 "/home/tomi/fs2open.github.com/code/osapi/dialogs.cpp", line=355) at /home/tomi/fs2open.github.com/code/osapi/osapi.cpp:400
400		gr_activate(1);
(gdb) bt
#0  0x0000555555e20972 in debug_int3(char const*, int) (file=0x555556117ae8 "/home/tomi/fs2open.github.com/code/osapi/dialogs.cpp", line=355)
    at /home/tomi/fs2open.github.com/code/osapi/osapi.cpp:400
#1  0x0000555555e26e67 in os::dialogs::Error(char const*) (text=0x555561d6e500 "Assert: \"model_instance_num < (int)Polygon_model_instances.size()\"\nFile: modelread.cpp\nLine: 3191\n\n  /home/tomi/freespace2/fs2_open_3_8_1_x64-DEBUG : os::dialogs::AssertMessage(char const*, char const"...)
    at /home/tomi/fs2open.github.com/code/osapi/dialogs.cpp:355
#2  0x0000555555e26014 in os::dialogs::AssertMessage(char const*, char const*, int, char const*, ...) (text=0x5555561005e8 "model_instance_num < (int)Polygon_model_instances.size()", filename=0x5555560feb29 "modelread.cpp", linenum=3191, format=0x0) at /home/tomi/fs2open.github.com/code/osapi/dialogs.cpp:148
#3  0x0000555555d47583 in model_get_instance(int) (model_instance_num=1) at /home/tomi/fs2open.github.com/code/model/modelread.cpp:3191
#4  0x0000555555d4c008 in model_clear_submodel_instances(int) (model_instance_num=1) at /home/tomi/fs2open.github.com/code/model/modelread.cpp:4733
#5  0x0000555555f935dd in ship_model_update_instance(object*) (objp=0x555556998090 <Objects+560>) at /home/tomi/fs2open.github.com/code/ship/ship.cpp:12907
#6  0x0000555555c3fbe3 in labviewer_render_model(float) (frametime=0.0184326172) at /home/tomi/fs2open.github.com/code/lab/lab.cpp:392
[...]

PR #1659, opened 2018-03-16
Delete all objects on mission end (Fix for #1653)

The-E:
This adds a step to the post-mission cleanup where we explicitly go through Objects and call obj_delete on everything not already deleted; this obviously needs testing.

Also a few cleanup steps in lab_init to make sure the lab isn't contaminated by leftover data from previously played missions.

### DIFF
--- a/code/object/object.cpp
+++ b/code/object/object.cpp
@@ -518,7 +518,7 @@ int obj_create(ubyte type,int parent_obj,int instance, matrix * orient,
 
 void obj_delete_all() 
 {
-	size_t counter = 0;
+	int counter = 0;
 	for (size_t i = 0; i < MAX_OBJECTS; ++i) 
 	{
 		if (Objects[i].type == OBJ_NONE)
--- a/code/object/object.cpp
+++ b/code/object/object.cpp
@@ -516,6 +516,20 @@ int obj_create(ubyte type,int parent_obj,int instance, matrix * orient,
 	return objnum;
 }
 
+void obj_delete_all() 
+{
+	size_t counter = 0;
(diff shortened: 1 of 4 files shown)


code/object/object.cpp
```diff
@@ -574,7 +588,7 @@ void obj_delete(int objnum)
 	case OBJ_START:
 	case OBJ_WAYPOINT:
 	case OBJ_POINT:
-		Assert(Fred_running);
+		//Assert(Fred_running);
```
asarium: This should be removed or explained in some way.

tomimaki:
Tested this fix by going to missions and lab multiple times and game didn't crash. But there is one strange thing.

screen0127
It looks like lab loads retail starfield even if background is none. Only happens after some mission.

Goober5000:
I could be wrong, but I think one of the reasons objects aren't automatically deleted on mission close is so that they can be reloaded more quickly if you restart the mission.

code/object/object.cpp
The-E: The assert needed to be removed because trying to delete all objects in a mission will, by necessity, hit point objects; an assertion that FRED is running is therefore not needed at that point.

asarium: That makes sense but it should be actually removed and not just commented out.

The-E: It will be in the next commit to this branch

The-E:
@Goober5000 As far as I can tell, calling (for example) ship_delete on a ship doesn't actually unload any of the textures or models associated with that ship; what this does however is delete some of the intermediary data that would have to be recreated on the next mission load anyway (such as model instance data).

tomimaki:
Seems now all looks and works ok.

PR #1667, opened 2018-03-24
Add TeaCI config file and fix MinGW warnings

asarium:
[TeaCI](https://tea-ci.org/) is a CI service which provides MinGW CI builds (in Docker
containers so no "real" Windows machines). This would be useful for
validating the Windows only code with additional GCC warnings. It would
also make sure that the MinGW setup we currently have continues to work.

This also fixes the issues detected by MinGW to make it compile cleanly.


chief1983:
Could the fixes to get it to compile be submitted in separate PR(s) so this could just be the CI implementation changes?  It might even helpful if those included an issue of what specific MinGW issue they addressed.  I feel that might be good documentation to have with each of those changes.

asarium:
Since the CI changes require a clean compilation they would not be very helpful without the fixes. I could separate the fixes into individual commits but that would be a lot of work for pretty much no benefit since there are so few changes in this PR.

If there is anything that is unclear then I would be happy to explain that.

chief1983:
Yeah I was suggesting that the fixes go into a PR that could be merged in first, then this would be a clean compilation.  I wouldn't see why the fixes could not go in ahead of the CI support.

asarium:
I posted a new PR with only the MinGW fixes (#1668).

PR #1669, opened 2018-03-26
Add fiction viewer dialog to qtFRED

jg18:
EDIT: This adds the fiction viewer dialog to qtFRED.

### DIFF
--- a/qtfred/src/ui/dialogs/FictionViewerDialog.cpp
+++ b/qtfred/src/ui/dialogs/FictionViewerDialog.cpp
@@ -40,7 +40,7 @@ FictionViewerDialog::FictionViewerDialog(FredView* parent, EditorViewport* viewp
 	resize(QDialog::sizeHint());
 
 	if (_model->hasMultipleStages()) {
-		auto button = _viewport->dialogProvider->showButtonDialog(DialogType::Information, "Multiple stages detected",
+		viewport->dialogProvider->showButtonDialog(DialogType::Information, "Multiple stages detected",
 			"This mission has multiple fiction viewer stages defined.  Currently, qtFRED will only allow you to edit the first stage.",
 			{ DialogButton::Ok});
 	}
--- a/qtfred/src/ui/dialogs/FictionViewerDialog.cpp
+++ b/qtfred/src/ui/dialogs/FictionViewerDialog.cpp
@@ -1,6 +1,5 @@
 
-// TODO sort out includes
-#include <QtWidgets/QTextEdit>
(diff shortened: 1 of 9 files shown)


asarium:
Here are a few things I noticed:
- The dialog isn't accessible anywhere at the moment. You can do that easily by creating a function called `void on_actionFiction_Viewer_triggered(bool)` in the `FredView` class. That uses the automatic signal/slot matcher from Qt which connects that function to the `triggered` function of the `actionFiction_Viewer` object which is the menu action for opening the fiction viewer dialog.
- The dialog buttons are not connected to the dialog slots which means that it is impossible to close it at the moment.
- I saw that you asked if the instance was created only once or every time. FRED only creates the dialog instances once probably to use less resources but that causes some other coding issues which is why qtFRED create the dialog instances on demand at the moment so you can assume that it will be done like that for the fiction viewer.
- I think you are currently not emitting the `modified` signal, correct? You should do that whenever one of the setters is called with a different value. You can make this easier to implement by using the `modify` template function other dialog models are already using.

Keep up the good work!

qtfred/src/ui/dialogs/FictionViewerDialog.cpp
```diff
@@ -0,0 +1,115 @@
...
+	for (const auto& el : _model->getMusicOptions()) {
+		ui->musicComboBox->addItem(QString::fromStdString(el.name), QVariant(el.id));
+	}
+
+	ui->musicComboBox->setEnabled(ui->musicComboBox->count() > 0);
+}
+void FictionViewerDialog::updateUI() {
+	// We need to block signals here or else updating the combobox would lead to and update of the model
+	// which would call this function again
+	QSignalBlocker blocker(this);
```
asarium: Since the recursive signal blocker is now merged you could use that here to reduce the amount of code.

jg18:
Hmm looks like I messed up something when I tried to rebase from upstream/master. What's the best way to get rid of the commits that don't belong in this branch? Or would squashing into a single commit when merging fix it?

qtfred/ui/FredView.ui
```diff
@@ -20,7 +20,7 @@
      <x>0</x>
      <y>0</y>
      <width>926</width>
-     <height>28</height>
+     <height>21</height>
```
jg18: No idea how the change to that line got there. Should I revert it?

asarium: If you didn't make this change then you should revert that.

Issue #1503, opened 2017-11-03
How to (possibly) fix BMPMAN


asarium:
After the Discord discussion about WoDs usage of bmpman slots and the resulting changes in #1502 I though about how bmpman could be changed to fix the core issue. I don't know a lot about how the internals of bmpman work but I know how it uses the `bm_bitmaps` array for keeping track of what bitmaps are currently in use. If this data structure could be changed to support a variable number of entries then the main issue of bmpman would be solved so I though about how to do that.

I tried to use a hash map for storing the bmpman slots once but that has some issues with how the data is handled and the STL hash maps are also not the fastest solutions out there. My "solution" for this problem would borrow some aspects from a hash map but it would be optimized for how bmpman is currently working. The system would use multiple blocks (where each is similar to the current `bm_bitmaps` array) of fixed size. When bmpman needs to allocate a bitmap info then it will choose a free slot in one of the blocks and then assemble the 32-bit handle (bmpman currently uses an `int` so that cannot be changed to a 64-bit integer easily) by using the lower 16-bits as an index into the block and the upper 16 (15 actually since it is a signed integer) for storing the index of the block.

As far as I can tell, this system should allow bmpman to allocate a new block if the old block does not have enough slots left over which would fix the most pressing issue of bmpman (at least I think that is the most pressing issue).

Thoughts?

TRBlount:
A logical -> physical addressing scheme should be sufficient. This means that we would get 32K blocks each holding 64k slots correct?

asarium:
Yes, that's exactly what I mean. 64k slots is probably too much so this could be lowered to 4096 or something like that since it would use too much memory otherwise.

z64555:
I'm wondering if a double ring system (similar to what beams and others use) would be of use.

Goober5000:
Closing this issue since it was implemented in #1670.

PR #1670, opened 2018-03-29
Add new bmpman slot management system

asarium:
This changes how the bitmap slots are managed by the bmpman system.
Instead of allocating a fixed-size array when the game starts this will
manage multiple "blocks" of bitmap slots. When the engine runs out of
slots in the already allocated blocks it simply creates a new block and
allocates the new slot in that.

This removed the hard limit on how many bitmaps can be loaded at the
same time and should fix issues encountered by mods which use a lot of
bitmap slots (I'm looking at you Spoon).

This implements the idea described in #1503.

I will create a post in the test build forum once the builds have finished compiling. Until this has received more testing it should not be merged.

### DIFF
--- a/code/bmpman/bm_internal.h
+++ b/code/bmpman/bm_internal.h
@@ -23,6 +23,16 @@
 
 #include "bmpman/bmpman.h"
 
+#include <array>
+
+/**
+ * @brief Container class for graphics API specific bitmap data
+ */
+class gr_bitmap_info {
+ public:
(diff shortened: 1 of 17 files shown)


TRBlount:
Preliminary tests from Spoon indicate that the game runs fine and does extend the BMPMAN slot limit once it is hit.

asarium:
Since there are some reports that the changes do not immediately break anything I think this should be ready for merging now.

PR #1678, opened 2018-04-08
Add GUIs for background, shield system, and voice acting editors to qtFRED

jg18:
Add .ui files and stub dialog .h/.cpp files for Background Editor, Shield System Dialog, and Voice Acting Manager to qtFRED.

### DIFF
--- a/qtfred/ui/BackgroundEditor.ui
+++ b/qtfred/ui/BackgroundEditor.ui
@@ -1,0 +1,899 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>fso::fred::dialogs::BackgroundEditor</class>
+ <widget class="QDialog" name="fso::fred::dialogs::BackgroundEditor">
+  <property name="geometry">
+   <rect>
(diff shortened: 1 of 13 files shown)


qtfred/ui/BackgroundEditor.ui
```diff
@@ -0,0 +1,889 @@
...
+          </item>
+         </layout>
+        </item>
+       </layout>
+      </widget>
+     </item>
+    </layout>
+   </item>
+  </layout>
+  <zorder></zorder>
```
asarium: The Qt .ui compiler is complaining about this and the following line.
The warning message is this: `qtfred/ui/BackgroundEditor.ui: Warning: Z-order assignment: '' is not a valid widget.`

jg18: Weird about the zorder lines, not sure how they got there. Added the SHIFT+I shortcut for Background Editor, I think that's what you were referring to? Were there other shortcuts as well? Although does it make sense to add the shortcut when the dialogs currently don't do anything?

PR #1679, opened 2018-04-08
Copy original FRED initialization code

asarium:
This copies the original FRED code for initializing the individual game
systems to qtFRED. This should fix the various issues caused by missing
game system initialization calls.

### DIFF
--- a/qtfred/src/main.cpp
+++ b/qtfred/src/main.cpp
@@ -1,9 +1,6 @@
 #include <unordered_map>
 #include <memory>
-#include <utility>
 #include <functional>
-
-#include <stdlib.h>
 
 #include <QApplication>
 #include <QDir>
@@ -15,13 +12,10 @@
 #include "globalincs/mspdb_callstack.h"
 #endif
 
-#include "globalincs/globals.h" // MAX_SHIPS, NAME_LENGTH
-
 #include "mission/Editor.h"
 #include "mission/management.h"
 
 #include "ui/FredView.h"
-#include "FredApplication.h"
 #include "FredApplication.h"
 
 #include <signal.h>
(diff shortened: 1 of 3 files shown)


qtfred/src/main.cpp
```diff
@@ -119,38 +113,60 @@ int main(int argc, char* argv[]) {
...
+								 { SubSystem::Parse,             app.tr("Initializing parser") },
+								 { SubSystem::TechroomIntel,     app.tr("Initializing techroom intel") },
+								 { SubSystem::Nebulas,           app.tr("Initializing nebulas") },
+								 { SubSystem::Stars,             app.tr("Initializing stars") },
+								 { SubSystem::Ssm,               app.tr("Initializing SSMs") },
+								 { SubSystem::EventMusic,        app.tr("Initializing event music") },
+								 { SubSystem::FictionViewer,     app.tr("Initializing fiction viewer") },
+								 { SubSystem::CommandBriefing,   app.tr("Initializing command briefing") },
+								 { SubSystem::Campaign,          app.tr("Initializing campaign system") },
+								 { SubSystem::NebulaLighting,    app.tr("Initializing nebula lighting") },
```
jg18: Maybe I misread things, but shouldn't this be NebulaLightning?

qtfred/src/mission/management.cpp
```diff
@@ -26,75 +35,220 @@ extern void allocate_mission_text(size_t size);
...
+	Xstr_inited = 1;
+
+#ifndef NDEBUG
 	load_filter_info();
 #endif
 
-	std::setlocale(LC_ALL, "C");
+	//CFREDView *window = CFREDView::GetView();
+	//HWND hwndApp = window->GetSafeHwnd();
+	//os_set_window((uint) hwndApp);
```
jg18: Can you remove this commented out code if we don't need it?

jg18:
I tested these changes with my fiction viewer dialog branch and the fiction music options now appear correctly in the drop down box. Cool stuff. Just a couple small comments.

qtfred/src/main.cpp
asarium: It seems like I misunderstood what that system was for...

qtfred/src/mission/management.cpp
asarium: Fixed.

PR #1681, opened 2018-04-10
Add clang-tidy to Travis CI configuration

asarium:
This will execute [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) when building the code on Travis CI to do some static code analysis. This currently only enables the check for proper `override` usage but I intend to enable more checks in the future.


asarium:
I enabled the performance checks and fixed all detected issues.

PR #1686, opened 2018-04-15
Minor clean up

PeterMitsis:
For your consideration, these patches contain various minor tweaks to help tidy up the code base.

For several files, the 'static' keyword has been applied to various routines.  On my local development machine, applying these patches reduces the size of the resulting image by about 8 kB (Ubuntu 16.04 with gcc version 5.4). Admittedly, this is pretty minor in the larger scheme of things, and mileage with other compilers may vary.

A few routines were discovered to not be in use (for several years if my searches were done correctly) and thus this set includes a patch to remove them.

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -1418,35 +1418,6 @@ void asteroid_level_close()
 DCF_BOOL2(asteroids, Asteroids_enabled, "enables or disables asteroids", "Usage: asteroids [bool]\nTurns asteroid system on/off.  If nothing passed, then toggles it.\n");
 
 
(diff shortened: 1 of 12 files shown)


asarium:
The Windows builds now detected some functions that are unused so you should remove those too.

PeterMitsis:
Hmmm, I would have thought these the unused (static) routines would have been flagged on my linux builds, but unfortunately they are not.  Methinks that bears some investigation. Until then, hopefully there are not too many cases of these unused routines that are being discovered one by one.

asarium:
I've seen that behavior before and I'm not sure why it doesn't happen with all compilers but that is why we have the CI servers.

PR #1701, opened 2018-04-15
is-in-turret-fov sexp

Goober5000:
Returns true if the ship is in a turret's FOV, false otherwise.  Requested by JSRNerdo.  Uses the object_in_turret_fov internal function which is the same one used by the turret code when it checks whether it can fire.  Optionally uses an argument for distance/range.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -508,6 +508,7 @@ SCP_vector<sexp_oper> Operators = {
 	{ "turret-set-secondary-ammo",		OP_TURRET_SET_SECONDARY_AMMO,			4,	4,			SEXP_ACTION_OPERATOR,	},	// DahBlount
 	{ "turret-get-primary-ammo",		OP_TURRET_GET_PRIMARY_AMMO,				3,	3,			SEXP_INTEGER_OPERATOR,	},	// DahBlount
 	{ "turret-get-secondary-ammo",		OP_TURRET_GET_SECONDARY_AMMO,			3,	3,			SEXP_INTEGER_OPERATOR,	},	// DahBlount
+	{ "is-in-turret-fov",				OP_IS_IN_TURRET_FOV,					3,	3,			SEXP_BOOLEAN_OPERATOR,	},	// Goober5000
 	
 	//Models and Textures Sub-Category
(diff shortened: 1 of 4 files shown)


fred2/sexp_tree.cpp
```diff
@@ -4838,6 +4838,7 @@ sexp_list_item *sexp_tree::get_listing_opf_subsystem(int parent_node, int arg_in
 		case OP_IS_AI_CLASS:
 		case OP_MISSILE_LOCKED:
 		case OP_SHIP_SUBSYS_GUARDIAN_THRESHOLD:
+		case OP_IS_IN_TURRET_FOV:
```
asarium: Could you port this change to qtFRED as well?

code/parse/sexp.cpp
```diff
@@ -18278,6 +18280,61 @@ void set_turret_secondary_ammo(ship_subsys *turret, int requested_bank, int requ
...
+	Assertion(target_shipnum >= 0, "Couldn't find target ship '%s' in sexp_is_in_turret_fov!", target_ship_name);
+	Assertion(turret_shipnum >= 0, "Couldn't find turreted ship '%s' in sexp_is_in_turret_fov!", turret_ship_name);
+
+	// ...and their objects
+	target_objp = &Objects[Ships[target_shipnum].objnum];
+	turret_objp = &Objects[Ships[turret_shipnum].objnum];
+
+	// find the turret
+	turret_subsys = ship_get_subsys(&Ships[turret_shipnum], turret_subsys_name);
+	Assertion(turret_subsys != nullptr, "Couldn't find turret subsystem '%s' on ship '%s' in sexp_is_in_turret_fov!", turret_subsys_name, turret_ship_name);
```
asarium: Could it be possible that `turret_subsys_name` is invalid at this point due to a mission designer error?

Maybe if the ship class was changed from the original ship class. If that is possible then it would probably be better to use a `Warning` here or whatever is used in the other SEXPs.

fred2/sexp_tree.cpp
Goober5000: Done.

code/parse/sexp.cpp
Goober5000: I wouldn't know, but that's an interesting contingency.  I'll change it to a `Warning`.

PR #1702, opened 2018-04-15
minor sexp subsystem cleanup

Goober5000:
Synchronize the two subsystem-of-ship checks in sexp.cpp and sexp_tree.cpp.  Mostly this is just adding OP_BEAM_FLOATING_FIRE to sexp_tree.cpp because whoever added that sexp forgot to add it there.


Goober5000:
Okay, done.

asarium:
The Appveyor issue was fixed so updating this to the current master version would have fixed that.

PR #1709, opened 2018-04-20
Finish shield system editor dialog in qtFRED

jg18:


qtfred/src/mission/Editor.cpp
```diff
@@ -3118,5 +3110,71 @@ int Editor::global_error_check_mixed_player_wing(int w) {
 
 	return 0;
 }
+
+bool Editor::compareShieldSysData(const std::vector<int>& teams, const std::vector<int>& types) const {
+	Assert(Shield_sys_teams.size() == teams.size());
+	Assert(Shield_sys_types.size() == types.size());
+	if (Shield_sys_teams.size() != teams.size() || Shield_sys_types.size() != types.size()) {
```
asarium: This check is unnecessary. This only need this `if` statement or the `Assert` not both.

qtfred/src/mission/Editor.cpp
```diff
@@ -3118,5 +3110,71 @@ int Editor::global_error_check_mixed_player_wing(int w) {
...
+
+void Editor::exportShieldSysData(std::vector<int>& teams, std::vector<int>& types) const {
+	teams = Shield_sys_teams;
+	types = Shield_sys_types;
+}
+
+void Editor::importShieldSysData(const std::vector<int>& teams, const std::vector<int>& types) {
+	Assert(Shield_sys_teams.size() == teams.size());
+	Assert(Shield_sys_types.size() == types.size());
+	if (Shield_sys_teams.size() != teams.size() || Shield_sys_types.size() != types.size()) {
```
asarium: Same as above.

asarium:
I just tested the code and it looks pretty good but I noticed that the radio boxes on the right side do not have an initial state like they had in the original FRED.

jg18:
"The dialog looks the same" as in looks correct? Or incorrect?

qtfred/src/mission/Editor.cpp
jg18: I'm not so sure it's unreasonable to have both checks. The asserts will fail noisily in debug mode and the if-check will fail silently in release mode. That said, if you want only one of the two, I'll take the asserts since at least those will fail noisily.

asarium:
I mean that it looks the same as the FRED dialog which means it is correct. I just used different configurations for FRED and qtFRED which caused the difference in appearance.

qtfred/src/mission/Editor.cpp
asarium: Well, the idea behind an assert is that it checks something that should always be true. If you then repeat the check outside of an `Assert` statement it means that the condition may be false in some circumstances which means that there may be a coding error somewhere.

PR #1718, opened 2018-04-28
Fix fog rendering in forward renderer

### LINKED ISSUE #1717: If deferred lighting is disabled, nebula fogging does not work
See attached screenshot.  This can most easily be seen by loading up Mystery of the Trinity and looking in the direction of the debris field.

The command-line is:
C:\GOG Games\Freespace 2\fs2_open_3_8_1_AVX.exe -mod "MediaVPs_2014" -nospec -nomotiondebris
-nonormal -noheight -nolightshafts -no_deferred

Removing -no_deferred fixes the fogging.

screen0110

screen0109

asarium:
There were actually two bugs here that were fixed:
1. The model draw list code disabled the fog rendering when it thought
it was it deferred rendering mode but did not account for explicitly
disabling the deferred renderer.
2. The fog effect was applied to both the emissive color and the base
color simultaneously which caused visual errors when rendering fogged
objects using the deferred renderer.

This fixes #1717.

### DIFF
--- a/code/def_files/data/effects/main-f.sdr
+++ b/code/def_files/data/effects/main-f.sdr
@@ -381,8 +381,10 @@ void main()
  #ifdef FLAG_DIFFUSE_MAP
 	if(blend_alpha == 1) finalFogColor *= baseColor.a;
  #endif
-	baseColor.rgb = mix(baseColor.rgb, finalFogColor, vertIn.fogDist);
-	emissiveColor.rgb = mix(emissiveColor.rgb, finalFogColor, vertIn.fogDist);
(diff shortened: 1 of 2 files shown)


Goober5000:
I tried out the changes and they do indeed fix the fogging.

Not sure I can contribute to a code review though, as I'm not familiar with the graphics stuff.

asarium:
Well, it's not like we have a lot of other graphics coders with the free time required for doing code reviews on graphics related code.

Goober5000:
I'll ping @SCP_dev in Discord.  If nobody responds after a few days I'll approve the review as based on the results.

PR #1721, opened 2018-05-02
Fix scrambling unicode messages

### LINKED ISSUE #1720: scrambled messages deforms unicode strings
Either by sexp or by EMP, if the player has scrambled messages it will deform the unicode string. In retail, the deformed characters will just not draw, but in scripts it will bring the game to a halt.

Test Mod by Novachen below:
[ScrambledMessage.zip](https://github.com/scp-fs2open/fs2open.github.com/files/1959235/ScrambledMessage.zip)

asarium:
The code did not account for unicode escape sequences when inserting
scrambling characters which distorted the text and fequently caused
invalid UTF-8 sequences to appear in the text. This fixes that by
using the text iterators already present in the game code. I needed to
rewrite the logic of this block a bit since the previous logic could not
be converted to the way we need to handle unicode sequences at the
moment.

This fixes #1720.

### DIFF
--- a/code/globalincs/pstypes.h
+++ b/code/globalincs/pstypes.h
@@ -261,14 +261,14 @@ extern int Global_error_count;
 // Disabling this functionality is dangerous, crazy values can run rampent unchecked and the longer its disabled
 // the more likely you are to have problems getting it working again.
 #if defined(NDEBUG)
-#	define Assert(expr) do { ASSUME(expr); } while (0)
+#	define Assert(expr) do { ASSUME(expr); } while (false)
 #else
 #	define Assert(expr) do {\
 		if (!(expr)) {\
 			os::dialogs::AssertMessage(#expr,__FILE__,__LINE__);\
 		}\
 		ASSUME( expr );\
-	} while (0)
+	} while (false)
 #endif
 /*******************NEVER COMMENT Assert ************************************************/
 
--- a/code/globalincs/toolchain/clang.h
+++ b/code/globalincs/toolchain/clang.h
@@ -32,7 +32,7 @@
 #define ASSUME(x)
 
 #if defined(NDEBUG)
-#	define Assertion(expr, msg, ...)  do { } while (0)
+#	define Assertion(expr, msg, ...)  do { } while (false)
 #else
 /*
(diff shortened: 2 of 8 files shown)


Goober5000:
I'm not exactly familiar with the Unicode iterators, but this looks straightforward enough.  I haven't tested it though.

You mentioned on #1720 that messages with voice files are harder to scramble using the Unicode API, but it looks like this PR addresses both cases.  Is there anything that remains unaddressed?

asarium:
No, this PR fixes the entire issue.

Goober5000:
Has @AxemP tested this patch?  I'm willing to approve it if he's satisfied.  Otherwise I looked it over again and it all looked in order.

One requested change though.  It looks like clang is not only reformatting blocks but also function definitions:
```
void message_maybe_distort_text(char* text, int shipnum) {
```
Can that be fixed so that the opening brace is on the following line?

AxemP:
I haven't tested the patch and this was actually more for Novachen's localization projects. Might want to get in contact with them then to see what builds they need

asarium:
@Goober5000 I fixed the formatting issue. FYI, GitHub has a review feature where you can comment on a specific line. That makes it easier to find the right location to fix the issue.

Goober5000:
@asarium Right, yes, I forgot I could do that.

I'll ping Novachen to test this patch.

NovachenFS2:
I will look into that.

Only to be sure and if i understand this correctly. There is no other behavior between scrambled messages with and without voice files? Because that would be an important information for a mission that cover all scenarios in a translation i can think of.

asarium:
The code does some special things if the message has a voice file since the scrambled parts of the text should be the same as the scrambled part of the sound file. However, there is no functional difference between both cases that would be relevant for a mission designer.

Do you know how to compile your own builds? If not I can start some test builds with these changes so you don't have to figure out how to use Visual Studio.

NovachenFS2:
Yes, i know how to compile builds :).

I will do that after my work in the evening. So in 4-5 hours. (it is 3 PM here right now)

Goober5000:
I'll merge this.  @NovachenFS2, you can open a separate issue for FRED2 saving missions in UTF-8 and someone will take a look.

PR #1723, opened 2018-05-03
Enable legacy config mode if legacy config is newer

asarium:
There were some reports that using a newer launcher together with an
outdated launcher was causing some issues since the new launcher would
write the new config file which would cause FSO to ignore the changes of
the old launcher since the new config file already exists.

This fixes that by adding a check for when the config locations were
last modified. On Linux and Mac it checks when the old config locations
were modified last while on Windows it checks the registry to determine
if there was a change after the new config file was created.

### DIFF
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -38,9 +38,6 @@
 
 char Cfile_root_dir[CFILE_ROOT_DIRECTORY_LEN] = "";
 char Cfile_user_dir[CFILE_ROOT_DIRECTORY_LEN] = "";
-#ifdef SCP_UNIX
-char Cfile_user_dir_legacy[CFILE_ROOT_DIRECTORY_LEN] = "";
-#endif
 
 // During cfile_init, verify that Pathtypes[n].index == n for each item
 // Each path must have a valid parent that can be tracable all the way back to the root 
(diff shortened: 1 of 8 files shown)


ngld:
Sounds like a very good idea! However, checking the cmdline file / registry key might be more reliable than checking the configuration.
Knossos only modifies `fs2_open.ini` if the user changed FSO settings while the `cmdline_fso.cfg` is written before each launch. Apparently wxLauncher always writes both files. Not sure about the other launchers but I'd guess they also would modify the cmdline before launching FSO.

asarium:
If Knossos only writes the configuration when it the FSO settings are changed from within Knossos then how would it handle changes done by wxLauncher?

I guess the legacy mode check could be made more robust by also including the modification time of `cmdline_fso.cfg` but I also don't want to over engineer this problem.

ngld:
The settings screen parses the `fs2_open.ini` and once the user clicks on Save the `fs2_open.ini` is updated with the new values. Knossos never uses these values (or saves them) outside of the settings screen. Changes made through wxLauncher would just appear in Knossos as if they'd been done there.

asarium:
I guess that makes sense. I added a check for the modification time of `cmdline_fso.cfg` which chooses whatever file was modified more recently in case one of the files is not changed by the launcher.

PR #1726, opened 2018-05-07
make the set-bit and is-bit-set operators fail gracefully

Goober5000:
These operators used to warn and return NAN if they were given a bit index out of range.  For FREDding convenience, out-of-range bit indexes are now a no-op.  This is particularly helpful in the situation where a FREDder needs to evaluate an extended bit array, one 32-bit chunk at a time, e.g.

0-31 bits: offset 0
32-63 bits: offset -32
64-95 bits: offset -64
96-127 bits: offset -96

Because the sexp system always evaluates every argument, three out of four branches were always evaluated with invalid bit indexes (negative or too large) even if the sexp logic only used one branch.  The new method silently bypasses this behavior.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -4262,8 +4262,9 @@ int sexp_set_bit(int node, bool set_it)
 
 	if (bit_index < 0 || bit_index > 31)
 	{
-		Warning(LOCATION, "Bit index %d out of range!  Must be between 0 and 31.", bit_index);
-		return SEXP_NAN;
+		// out of range, so just return the original value
+		// (we used to warn, but it's more convenient to FREDders to fail gracefully)
+		return val;
 	}
 
 	if (set_it)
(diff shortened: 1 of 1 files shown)


asarium:
Wouldn't this break existing missions that somehow depend on the NaN return value?

Goober5000:
Those missions would need to have a situation where both of the following conditions were true:
a) A bit index was being used that was sometimes in range and sometimes out of range
b) The proper operation of the sexp depended on *delayed evaluation* of the out-of-range value.

There are very few cases where you would even need condition b).  The distance sexp is the most frequently used example of such a sexp, and even there, delayed evaluation is only used for unusual specific scenarios, such as a ship coming close to something, moving farther away, and then coming close again.  I can't think of any reason a FREDder would need a bit index to do this.

Also, I think we can safely say that no FREDder has used this scenario because otherwise they would have complained about the `Warning` message popping up every frame (albeit only in debug builds).

In any case, this does not change the truth value of the sexp.  SEXP_NAN is treated as false except for the delayed evaluation side-effect.  This just changes it to an immediate false.

asarium:
Ok, thanks for the clarification.

Issue #1730, opened 2018-05-19
Some Localization Requests


NovachenFS2:
Because Mantis is not used these days anymore as it seems, i think, that this is the place to post some Requests for some localization features i need or atleast want to discuss.

As you may know i am constantly working on german translations of Freespace Mods.
For a new generation of updates, i added some new translation features where i also encountered some problems so far.

1. - Regardless that briefing labels are supporting XSTRs now, some entries are automatically translated by the code.
In the lcl_translate_brief_icon_name_gr() segment (as well for the other languages) of the localize.cpp there are some examples which briefing icons are translated automatically, if the corresponding string occurs. 
Fortunately the game only refers to the strings, that are appearing in the game, so to speak it is using only the already translated strings as a reference (if available). In the german language i actually have only problems with every briefing icon, that is labeled "Transporter" because it is nearly the same word as the english "Transport" what have to be translated by the rules of the code. If i use "Transporter" in tstrings as a translation, the game actually make it a "Transporterer" in the briefing with an additional -ER, what is totally wrong anyway.
This would not be not a big issue, if i remember to simply translate all "Transport/s" into "Transport" to let the game add the missing -ER by itself. The problem i have is, that this function drops all other characters besides the "Transport" part. If i have a label called "Shivan Transport" for example and translate it to "Shivanischer Transport/er" only the "Transporter/er" will appear. It is dropping the complete part before or after that term. I can only bypass this, if i use a lowercase t at the beginning. But i think, it would be much better to change the code, that these translation rules comes only in effect if the term is totally identical. So that the game still auto translates labels that are called "Transport/s"... but not if they are called "Shivan Transport/s" for example... or already translated terms like "Transporter".

2. - Because i am already in the localize.cpp. Due to the lcl_translate_ship_name_gr() and lcl_translate_targetbox_name_gr() some ship names are translated automatically without breaking the mission. For the creation of a mission that is supporting several languages at once, it would be helpful if there would be a feature to create an alternative name for some of the objects where the original name is used as reference. The ships and weapons.tbl uses the $Alt Name string for that, which make it possible to create translateable weapons (that Alt Name is used in the weapons gauge for example) and ship classes.
Unfortunately $Alt Name in missions is used for an alternative name for ship classes. But actually because the original game do this with Instructors and Support ships i am asking if it would not be helpful if you can use a new entry to translate objects or wings if it is needed. Well, at least there would be even a potential for new SEXPs where you can change the ship name without any scripts.

[...]

PR #1736, opened 2018-06-02
Implement various translation features requested by Novachen

asarium:
This fixes or implements a few things requested in #1730.

New features:
- Alternative name for medals that is only used for displaying their name. This allows them to be translated without causing issues with the medal SEXPs.
- A game_settings option for disabling all currently built-in translations. Since the parts of the engine where this was used have been updated with proper translation support, the built-in translations are not needed anymore. They can be disabled with a game_settings option so that backwards compatibility is not broken.

### DIFF
--- a/code/stats/medals.cpp
+++ b/code/stats/medals.cpp
@@ -224,6 +224,13 @@ medal_stuff& medal_stuff::operator=(const medal_stuff &m)
 
 	return *this;
 }
+const char* medal_stuff::get_display_string() {
+	if (!alt_name.empty()) {
+		return alt_name.c_str();
+	} else {
+		return name;
+	}
+}
 
 void parse_medal_tbl()
 {
@@ -310,6 +317,10 @@ void parse_medal_tbl()
 			// is this rank?  if so, save it
 			if (!stricmp(temp_medal.name, "Rank"))
 				Rank_medal_index = Num_medals;
+
+			if (optional_string("$Alt Name:")) {
+				stuff_string(temp_medal.alt_name, F_NAME);
+			}
 
 			required_string("$Bitmap:");
 			stuff_string(temp_medal.bitmap, F_NAME, MAX_FILENAME_LEN);
@@ -600,7 +611,7 @@ void medal_main_init(player *pl, int mode)
 		Medals_window.set_mask_bmap(bitmap_buf);
 }
 
-void blit_label(char *label, int num)
+void blit_label(const char *label, int num)
 {
 	int x, y, sw;
 	char text[256];
@@ -736,7 +747,7 @@ int medal_main_do()
 
 		default:
 			if (Player_score->medal_counts[region] > 0) {
-				blit_label(Medals[region].name, Player_score->medal_counts[region]);
+				blit_label(Medals[region].get_display_string(), Player_score->medal_counts[region]);
 			}
 			break;
 	} // end switch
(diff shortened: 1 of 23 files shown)


code/stats/medals.cpp
```diff
@@ -609,7 +620,8 @@ void blit_label(char *label, int num)
 
 	// translate medal names before displaying
 	// can't translate in table cuz the names are used in comparisons
-	if (Lcl_gr) {
+	// Medals no have alternate display names so this code can be disabled in the mod table
```
Goober5000: This comment has a typo

code/stats/medals.h
```diff
@@ -46,6 +47,8 @@ class medal_stuff
 	medal_stuff(const medal_stuff &m);
 	medal_stuff& operator=(const medal_stuff &m);
 
+	const char* get_display_string();
```
Goober5000: Medals have `get_display_string` while weapons have `get_display_name`.  This may not matter but I thought I should mention it.

Goober5000:
I notice this also fixes the weapon name in targetbox bug that someone mentioned.  (I can't find it on the issue list so maybe it was never reported as an issue.)

NovachenFS2:
> Add convencience function for retrieving weapon alt name

Even i know only one situation where this name appears and this is in the Targeting Box for Bombs. But now the translated name is visible there. So it seems to be working :)

> Add alternative display name for medals 

This one actually did not work. Regardless of $Alt Name with XSTR in the medals.tbl, the name appear untranslated in the Mission Debriefing aswell in the Medal Case.

> Add option for disabling built-in translations

This looked good.

> Disable hard-coded "cluster bomb" hack 

The translated name of the Cluster Bomb appears in the weapon gauge now, so this one is working also.

NovachenFS2:
The Alternate Name appears in the Medal Case, now. But in the debriefing of a mission the $Name is still displayed.

asarium:
Slight oversight, should be fixed now.

Goober5000:
If @NovachenFS2 if happy with these latest changes, I'm happy to merge the PR.

NovachenFS2:
> Add convencience function for retrieving weapon alt name 

I actually saw, that the Ballistic Weapon Gauge is displaying the $Name also, instead the $Alt Name. I think this still fits to this PR, too? Otherwise i make a new issue for that.

If there is a fix for the wrong medal name in the debriefing (do not see an updated file here?) then i did not test it yet.

asarium:
I fixed the `$Alt Name` usage for the ballistic weapon gauge.

The fix for the debriefing medal name should be in this PR. I added the changes to the original medals commit so that may be why you haven't seen a new commit for that.

NovachenFS2:
Ah okay.

The medals are working now as i had that in mind.
The $Alt Name is displayed in the Ballistic Gauge as well, which is also good.

But this time i had an oversight by myself 😊 . In the Targetbox the $Name of the Bomb is still displayed and not the $Alt Name.

asarium:
Ok, I'll take a look at it tomorrow.

asarium:
I went through all occurrences of `name` and replaced them with the new function where ever the name was shown to the user. That should finally cover all places in the code where the alternate name should be shown.

NovachenFS2:
Yes, these changes are working and the $Alt Name where shown everywhere. So everything is working now. Thank you, that is a massive help for translations. :)

asarium:
@Goober5000 FYI, I made some additional changes since the last code review.

Goober5000:
Looks good and Novachen is happy.  Merging!

PR #1737, opened 2018-06-02
Change Warning about missing 640x480 interface art to a log statement

### LINKED ISSUE #1601: Load Screen 640 warning should be removed from FRED
The warning message "Warning: Mission: found an entry for $Load Screen 1024: but not a corresponding entry for $Load Screen 640:!" is functionally obsolete in 2018 as it is extremely unlikely that anyone will play FSO in such a low resolution.  Please suppress this warning.

Goober5000:
This turns the Warning about missing 640 interface art into a log print.  This applies to all instances of 640/1024 interface art that can be specified in FRED, e.g. loading screens, cbrief and briefing backgrounds, and ship/weapon selection backgrounds.

This fixes #1601.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -424,11 +424,11 @@ void parse_custom_bitmap(const char *expected_string_640, const char *expected_s
 	// error testing
 	if (Fred_running && (found640) && !(found1024))
 	{
-		Warning(LOCATION, "Mission: found an entry for %s but not a corresponding entry for %s!", expected_string_640, expected_string_1024);
+		nprintf(("General", "Mission: found an entry for %s but not a corresponding entry for %s!", expected_string_640, expected_string_1024));
 	}
 	if (Fred_running && !(found640) && (found1024))
 	{
-		Warning(LOCATION, "Mission: found an entry for %s but not a corresponding entry for %s!", expected_string_1024, expected_string_640);
+		nprintf(("General", "Mission: found an entry for %s but not a corresponding entry for %s!", expected_string_1024, expected_string_640));
 	}
 }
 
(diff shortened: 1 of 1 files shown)


code/mission/missionparse.cpp
```diff
@@ -424,11 +424,11 @@ void parse_custom_bitmap(const char *expected_string_640, const char *expected_s
 	// error testing
 	if (Fred_running && (found640) && !(found1024))
 	{
-		Warning(LOCATION, "Mission: found an entry for %s but not a corresponding entry for %s!", expected_string_640, expected_string_1024);
+		nprintf(("General", "Mission: found an entry for %s but not a corresponding entry for %s!", expected_string_640, expected_string_1024));
```
asarium: Isn't `nprintf(("General"` the same as `mprintf((`? If that is the case then maybe this should use `mprintf` instead.

Goober5000: I don't actually know.  There are plenty of `nprintf(("General"` statements throughout the codebase which have to do with parsing so that's why I used it.

Goober5000: There is an `nprintf(("Parse"` but it's only used by `display_parse_diagnostics()`.  What do you think of changing this to `Parse` and then opening a separate PR to change all applicable `mprintf` statements?

asarium: If it's a parse specific log message then "Parse" would be good too.

Goober5000: It's not as cut-and-dried as I thought, so I left the other `mprintf`s alone.  I don't think it matters whether this one is `nprintf(("General"` or `mprintf((`, but I changed it.

PR #1743, opened 2018-06-05
Respect trailing white space in tstrings.tbl strings

### LINKED ISSUE #1739: Small Localized Variable Problems
Because of the Discussion in #1730 i tested a bit with the Translated Variables and found some small problems during this.

In the modify-variable-xstr SEXP the default text is being used in the english language setting, even if the text in the tstrings.tbl under #default is different. This is different to the behavior the rest of the game has, because it prioritize the tstrings.tbl over every text in the tables and missions, most likely to be able to fix spelling errors for example without to change the mission and table files. So the default text from the SEXP should only be used if the ID is invalid or not given at all.

If a string have to be end with a space, because this string will be later concaternate with an Integer where the space is the seperator between them, the Space does not seem to be carried over.
In the translated version the concaternate Variable appears together without a seperator between.

[LocalVAR.zip](https://github.com/scp-fs2open/fs2open.github.com/files/2068764/LocalVAR.zip)
This is a small mod for that, which is based of a scenario from the first BtA: Mefistofele mission, where i encountered this problem.
In this mod, the text in #default of the tlc.tbm is different to the text in the modify-variable-xstr SEXP of the test mission.
And there is a german translated version of the variables as well.

Interestingly, one of the two concaternates variables shows the right behavior here even in the translated version. No idea why this is, because the second entry does not work as it should.

In the testing LocalVAR mission, you have to scan the Poseidon to show its Communication and Jump Drive Energy (which is static here). So there are two Variables, which were connected from a string and number variable.

asarium:
The current code drops trailing white space which can cause issues if
the translated string is used in concatenations. The retail tstrings.tbl
has some entries with white space at the end but since those strings are
not used for concatenations (or else this issue would have been found
earlier) this change should not cause issues there.

This fixes #1739.

### DIFF
--- a/code/localization/localize.cpp
+++ b/code/localization/localize.cpp
@@ -278,7 +278,6 @@ void parse_stringstbl_common(const char *filename, const bool external)
 			if (external) {
 				ignore_white_space();
 				get_string(buf, sizeof(buf));
-				drop_trailing_white_space(buf);
 			} else {
 				stuff_string(buf, F_RAW, sizeof(buf));
 			}


Goober5000:
I kinda don't like this idea.  Dropping trailing whitespace is pretty typical parsing behavior.

Why not add leading whitespace if a string is designed to be concatenated, or else add whitespace in the concatenation function itself?

asarium:
There are multiple problems here which all point towards the solution of not dropping trailing white space:
1. The string in `tstrings.tbl` is enclosed in `"` characters so it would be obvious to a table editor that a string has trailing whitespace. This does not drop invisible, unwanted trailing whitespace. Instead it removes characters that are explicitly added in the tstrings table.
2. The default string is used without removing the trailing white space so this means that there is inconsistent behavior when using a translated string as opposed to using the default language.

Goober5000:
I don't think 2) matters, but 1) definitely makes a difference.  If it's within the quotes, it makes sense not to trim it.

PR #1746, opened 2018-06-05
Add ship display name feature

asarium:
This adds a mission file option for ships which allows specifying a
different string which should be used for displaying the name of the
ship to the user. This name will not be used in SEXPs or other ship
lookup functions so it can be freely translated and two ships can also
have the same display name.

The new feature can be used by adding `$Display name: <string>` after `$Name`
for a ship in a mission file. I think I added proper support for FRED
(and qtFRED) so that this new value is properly saved but it would be
great if someone could test that.

This implements one of the features requested in #1730.

Display name support is still missing from the events screen since
the same data is also used by various SEXPs where the actual ship name
is used.

### DIFF
--- a/fred2/missionsave.cpp
+++ b/fred2/missionsave.cpp
@@ -2672,6 +2672,14 @@ int CFred_mission_save::save_objects() {
 		parse_comments(z ? 2 : 1);
 		fout(" %s\t\t;! Object #%d\n", shipp->ship_name, i);
 
+		// Display name
+		if (Format_fs2_open != FSO_FORMAT_RETAIL && shipp->has_display_name()) {
+			// The display name is only written if there was one at the start to avoid introducing inconsistencies
+			fout("\n$Display name:");
+			fout_ext(" ", "%s", shipp->display_name.c_str());
+			fout("\n");
+		}
+
 		required_string_fred("$Class:");
 		parse_comments(0);
 		fout(" %s", Ship_info[shipp->ship_info_index].name);
--- a/code/hud/hudescort.cpp
+++ b/code/hud/hudescort.cpp
@@ -372,7 +372,7 @@ void HudGaugeEscort::renderIcon(int x, int y, int index)
 	}
 
 	// print out ship name
-	strcpy_s(buf, sp->ship_name);
+	strcpy_s(buf, sp->get_display_string());
 	font::force_fit_string(buf, 255, ship_name_max_width);
 	
     end_string_at_first_hash_symbol(buf);
--- a/code/hud/hudmessage.cpp
+++ b/code/hud/hudmessage.cpp
@@ -558,7 +558,7 @@ void HUD_ship_sent_printf(int sh, const char *format, ...)
 	tmp[sizeof(tmp)-1] = '\0';
 	size_t len;
 
-	snprintf(tmp, sizeof(tmp)-1, NOX("%s: "), Ships[sh].ship_name);
+	snprintf(tmp, sizeof(tmp)-1, NOX("%s: "), Ships[sh].get_display_string());
 	len = strlen(tmp);
 
 	va_start(args, format);
--- a/code/hud/hudobserver.cpp
+++ b/code/hud/hudobserver.cpp
@@ -30,6 +30,7 @@ void hud_observer_init(ship *shipp, ai_info *aip)
 	// (we used to do a memcpy here, but that doesn't work any longer, so let's just assign the values we need)
 	Hud_obs_ship.clear();
 	strcpy_s(Hud_obs_ship.ship_name, shipp->ship_name);
+	Hud_obs_ship.display_name = shipp->display_name;
 	Hud_obs_ship.team = shipp->team;
 	Hud_obs_ship.ai_index = shipp->ai_index;
 	Hud_obs_ship.flags = shipp->flags;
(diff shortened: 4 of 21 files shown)


fred2/missionsave.cpp
```diff
@@ -2672,6 +2672,14 @@ int CFred_mission_save::save_objects() {
 		parse_comments(z ? 2 : 1);
 		fout(" %s\t\t;! Object #%d\n", shipp->ship_name, i);
 
+		// Display name
+		if (save_format != MissionFormat::RETAIL && shipp->has_display_name()) {
```
ngld: `save_format` and `MissionFormat` aren't declared here. They only exist in qtFRED.

NovachenFS2:
Tested this in two missions so far, so i does not have more than a first impression of it and i think that i should test this with some more missions, to get better results from more different situations.

The Good:
In general, this one seems to work quite well already. The display name is appearing during the mission aswell in the log. Even in this state it is more than i thought i would get a week ago ^^.

The Ugly:
Only a oversight i saw so far. In the AI Goals display above the Target box the normal Name is still visible (for example: Dock with: )

The Bad:
Nothing so far that has to do with this PR exactly ;)
The lack of a display name for wings, because so far Display Names can only be used in 1-Wave wings properly or if you want to get rid of the number a ship gets, even the normal wing name is still visible in the log of course.

asarium:
I'll take a look at the missing parts. Unfortunately wings are completely different objects to the engine so those will not be affected by these changes. I'll take a look at adding support for them once this is finished and merged,

NovachenFS2:
Had some more scenarios in the mean time. They looked good so far. 
What i actually did not test so far, is if this one is also working with scripts like the Load/Ship Save. Maybe i had to create a small test campaign with this.

But well, one question, it is possible to expand this feature to Jump Node Names aswell?

asarium:
The checkpoint script will likely not work with this new feature since the display name is a separate attribute of the ship objects which needs to be handled separately. The scripting API has support for this attribute so it will be possible to adapt scripts after this is merged.

I still haven't found a good solution for handling the AI directives since the target name is stored as a string in the AI data structure similar to how the message log works.

I will take a look at Jump Nodes after this is merged.

asarium:
@NovachenFS2 I think I managed to add display name support for the AI directives. Could you test if that works for you now?

Also, I think the function FSO uses to force strings to fit on screen is not compatible with Unicode yet so you might encounter some issues there.

NovachenFS2:
About the script thing, you are right, of course. Even the only problem i encountered with the save/load script is only in situations when an object with a display name is replacing an object with another display name. In this case the initial display name of the object is still used. But for most checkpoint systems this would still even work how it is. But well i have to ask AdmiralMS and Axem (because his message script still uses the object name for the sender) for an update.


But well, this is about the latest changes here.
I never watched the AI directives display before actually, but created a testing mission for that to look at that now.
This seems to work, too. But the only case where the normal ship name appears is if i order to destroy a subsystem. Then the normal object name is used.
Also this does not seem to support Alternative Subsystems, too, because the normal subsystem name is also used. And it show names like Turret06 instead of the actual name of the turret like Missile lnchr.

> Also, I think the function FSO uses to force strings to fit on screen is not compatible with Unicode yet so you might encounter some issues there.
What do you mean?
Well, display names with unicode characters were not a problem here. Even ships with a very long name and without any space that contain unicode characters were not a problem, except that some characters were not drawn, because the name was too long, of course, which is a good thing ^^.

asarium:
The target ship name for the "destroy subsystem" order was an oversight which should be fixed in the latest changes.
The alternative name of the subsystem was not touched by this commit so I'll not fix that here but I'll take a look at it later.

> What do you mean?
Well, display names with unicode characters were not a problem here. Even ships with a very long name and without any space that contain unicode characters were not a problem, except that some characters were not drawn, because the name was too long, of course, which is a good thing ^^.

It's just something I noticed while doing my changes. The `force_fit` function works on a per-byte basis instead of per-character which will likely cause issues if it decides to insert the break inside a UTF-8 encoding sequence. You probably don't see it because the string rendering function ignores invalid characters but it's something I will need to fix at some point but no in this PR.
(Also, you need to leave one line between a quote and the rest of the text to have it display correctly.)

NovachenFS2:
> The alternative name of the subsystem was not touched by this commit so I'll not fix that here but I'll take a look at it later.

Yes of course. But because i saw that, i thought it would be a good idea to mention it right away :).

> (Also, you need to leave one line between a quote and the rest of the text to have it display correctly.)

Yes, i already fixed that as you wrote your answer :).

asarium:
Ok, I think that was the last issue with these changes. I'll see if I can find someone to review them to get it merged as soon as our CI is finished.

Karajorma:
The name on the log screen won't change either, right? That was the main reason I never bothered to try to implement something like this long ago. It just seemed like to much hassle to maintain two logs, one for display and one for use with SEXPs.

asarium:
What log screen do you reference? The mission log screen ("ship arrived", "ship destroyed", ...) is covered by these changes.

NovachenFS2:
Seems that one file was updated in the mean time.
But any news about the merging of this feature in general? Because i run this in own builds for quite some time now and encountered no further errors. Would be nice to see this in Nightlies finally, so that the mentioned Wing name or atleast Jump Node name extension of it can be created :).

PR #1747, opened 2018-06-06
Converted all Int3() calls in sexp.cpp into Warnings, Errors and Asserts

ngld:
I also added a new `Error` in `sexp_transfer_cargo` to make it consistent with `sexp_exchange_cargo`.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -11911,7 +11911,7 @@ void sexp_transfer_cargo(int n)
 	// we must be sure that these two objects are indeed docked
 	if (!dock_check_find_direct_docked_object(&Objects[Ships[shipnum1].objnum], &Objects[Ships[shipnum2].objnum]))
 	{
-		Error(LOCATION, "Tried to transfer cargo between %s and %s although they aren't docked!", Ships[shipnum1].ship_name, Ships[shipnum2].ship_name);
+		Warning(LOCATION, "Tried to transfer cargo between %s and %s although they aren't docked!", Ships[shipnum1].ship_name, Ships[shipnum2].ship_name);
 		return;
 	}
 
@@ -11964,7 +11964,7 @@ void sexp_exchange_cargo(int n)
 	// we must be sure that these two objects are indeed docked
 	if (!dock_check_find_direct_docked_object(&Objects[Ships[shipnum1].objnum], &Objects[Ships[shipnum2].objnum]))
 	{
-		Error(LOCATION, "Tried to transfer cargo between %s and %s although they aren't docked!", Ships[shipnum1].ship_name, Ships[shipnum2].ship_name);
+		Warning(LOCATION, "Tried to exchange cargo between %s and %s although they aren't docked!", Ships[shipnum1].ship_name, Ships[shipnum2].ship_name);
 		return;
 	}
 
@@ -13845,7 +13845,7 @@ int sexp_event_status( int n, int want_true )
 	int i, result;
 
 	name = CTEXT(n);
-	Assertion(name != NULL, "CTEXT returned NULL for node %d!", n);
+	Assertion(name != nullptr, "CTEXT returned NULL for node %d!", n);
 
 	for (i = 0; i < Num_mission_events; i++ ) {
 		// look for the event name, check it's status.  If formula is gone, we know the state won't ever change.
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -11916,6 +11911,7 @@ void sexp_transfer_cargo(int n)
 	// we must be sure that these two objects are indeed docked
 	if (!dock_check_find_direct_docked_object(&Objects[Ships[shipnum1].objnum], &Objects[Ships[shipnum2].objnum]))
 	{
+		Error(LOCATION, "Tried to transfer cargo between %s and %s although they aren't docked!", Ships[shipnum1].ship_name, Ships[shipnum2].ship_name);
```
Goober5000: I suggest a `Warning` here rather than an `Error`. It's not a catastrophic problem.

code/parse/sexp.cpp
```diff
@@ -11968,7 +11964,7 @@ void sexp_exchange_cargo(int n)
 	// we must be sure that these two objects are indeed docked
 	if (!dock_check_find_direct_docked_object(&Objects[Ships[shipnum1].objnum], &Objects[Ships[shipnum2].objnum]))
 	{
-		Int3();			// you are trying to transfer cargo between two ships not docked
+		Error(LOCATION, "Tried to transfer cargo between %s and %s although they aren't docked!", Ships[shipnum1].ship_name, Ships[shipnum2].ship_name);
```
Goober5000: Here too. Also, the message should say "exchange" rather than "transfer". (This sexp swaps the cargo; the other sexp moves it from one to the other, possibly overwriting the cargo in the destination.)

code/parse/sexp.cpp
```diff
@@ -13849,11 +13845,7 @@ int sexp_event_status( int n, int want_true )
 	int i, result;
 
 	name = CTEXT(n);
-
-	if (name == NULL) {
-		Int3();
-		return SEXP_FALSE;
-	}
+	Assertion(name != NULL, "CTEXT returned NULL for node %d!", n);
```
Goober5000: Suggest `nullptr` rather than `NULL`.

code/parse/sexp.cpp
```diff
@@ -13891,11 +13883,7 @@ int sexp_event_delay_status( int n, int want_true, bool use_msecs = false)
 	bool use_as_directive = false;
 
 	name = CTEXT(n);
-
-	if (name == NULL) {
-		Int3();
-		return SEXP_FALSE;
-	}
+	Assertion(name != NULL, "CTEXT returned NULL for node %d!", n);
```
Goober5000: Ditto.

code/parse/sexp.cpp
```diff
@@ -13957,11 +13945,7 @@ int sexp_event_incomplete(int n)
 	int i;
 
 	name = CTEXT(n);
-
-	if (name == NULL) {
-		Int3();
-		return SEXP_FALSE;
-	}
+	Assertion(name != NULL, "CTEXT returned NULL for node %d!", n);
```
Goober5000: Ditto.

code/parse/sexp.cpp
```diff
@@ -21007,11 +20992,7 @@ void sexp_set_camera_position(int n)
 void multi_sexp_set_camera_position()
 {
 	camera *cam = sexp_get_set_camera();
-	
-	if(cam == NULL) {
-		Int3();
-		return;
-	}
+	Assert(cam != NULL);
```
Goober5000: And here.

code/parse/sexp.cpp
```diff
@@ -23586,7 +23567,7 @@ int eval_sexp(int cur_node, int referenced_node)
 
 			case OP_DO_FOR_VALID_ARGUMENTS:
 				// do-for-valid-arguments should only ever be called within eval_when()
-				Int3();
+				Assertion(false, "do-for-valid-arguments was encountered in eval_sexp()!");
```
Goober5000: This should be changed to a `Warning`. It is admittedly a violation of the programming contract, but at the FREDder level, not the game engine level. A FREDder could conceivably make this mistake, and a `Warning` is more FREDder-friendly.

asarium:
We really need a macro for unreachable code that does not use something like `Assert(false)`.

Goober5000:
Good stuff.

PR #1748, opened 2018-06-06
Add macro for marking unreachable code

asarium:
This can be used for mark parts of the code which should be unreachable
in normal execution. In debug mode, if such a statement is reached an
Error dialog is shown which aborts the execution. In release mode the
macro is replaced with a compiler specific expression that lets the
compiler know that this part is unreachable which allows it to optimize
the code better.

### DIFF
--- a/code/object/objectdock.cpp
+++ b/code/object/objectdock.cpp
@@ -437,7 +437,7 @@ void dock_move_docked_objects(object *objp)
 			fastest_objp = objp;
 	}
 	
-	// start a tree with that object as the parent... do NOT use the überfunction for this,
+	// start a tree with that object as the parent... do NOT use the Ã¼berfunction for this,
 	// because we must use a tree for the parent ancestry to work correctly
 
 	// we don't need a bit array because OF_DOCKED_ALREADY_HANDLED takes care of it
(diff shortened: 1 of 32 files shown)


code/object/objectdock.cpp
```diff
@@ -437,7 +437,7 @@ void dock_move_docked_objects(object *objp)
 			fastest_objp = objp;
 	}
 	
-	// start a tree with that object as the parent... do NOT use the �berfunction for this,
+	// start a tree with that object as the parent... do NOT use the überfunction for this,
```
Goober5000: This line shows as modified, but I don't see what changed.

code/object/objectdock.cpp
```diff
@@ -632,7 +632,7 @@ void dock_find_max_speed_helper(object *objp, dock_function_info *infop)
 	}
 }
 // ---------------------------------------------------------------------------------------------------------------
-// end of �ber code block ----------------------------------------------------------------------------------------
+// end of über code block ----------------------------------------------------------------------------------------
```
Goober5000: Same here

code/object/objectdock.cpp
asarium: I'm not entirely sure what happened there but on my initial commit I saw a change from an "ü" to an invalid character. When I tried to revert my changes nothing happened so I retyped the character in a text editor which fixed the displayed character but still showed some changes.

My guess it that the character was using an encoding different from UTF-8 which explains why it appeared as an invalid character.

Goober5000: I thought it might have something to do with the ü but I assumed that if the line was fixed to be identical, the line wouldn't show up as changed in a diff.

Ah well, not a blocker.  LGTM.

Goober5000: On second thought, I betcha the original ü was extended ASCII whereas the new ü is proper Unicode.  That would do it.

PR #1749, opened 2018-06-06
Add support for external subtitle files

asarium:
This allows to use external subtile files (currently only SRT is
configured) for providing subtitles for movie files. This should be more
universal since it also supports container formats with no support for
subtitle streams.

To use this code the srt file needs to be named
`<movie name>-<language code>.srt` where the language code is the file
extension name specified in the localization table. For english the
`-...` part can be omitted.

### DIFF
--- a/code/cutscene/ffmpeg/FFMPEGDecoder.cpp
+++ b/code/cutscene/ffmpeg/FFMPEGDecoder.cpp
@@ -11,6 +11,7 @@
 #include "cutscene/ffmpeg/VideoDecoder.h"
 #include "cutscene/ffmpeg/SubtitleDecoder.h"
 
+#include "FFMPEGDecoder.h"
 #include "localization/localize.h"
 
 using namespace libs::ffmpeg;
@@ -39,6 +40,8 @@ const char* CHECKED_EXTENSIONS[] = {
 	"mve"
 };
 
+const char* CHECKED_SUBT_EXTENSIONS[] = {"srt"};
+
 double getFrameRate(AVStream* stream, AVCodecContext* codecCtx) {
 	auto fps = av_q2d(av_stream_get_r_frame_rate(stream));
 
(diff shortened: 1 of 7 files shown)


asarium:
This code has been tested and confirmed to be working: https://www.hard-light.net/forums/index.php?topic=94802.msg1868971#msg1868971

code/cutscene/ffmpeg/FFMPEGDecoder.cpp
```diff
@@ -55,48 +58,48 @@ double getFrameRate(AVStream* stream, AVCodecContext* codecCtx) {
 	return fps;
 }
 
-CodecContextParameters getCodecParameters(AVStream* stream) {
+CodecContextParameters getCodecParameters(AVStream *stream) {
```
Goober5000: I don't know if it's clang or something else, but this PR has the opening braces of *functions* on the previous line rather than aligned on the following line.

code/cutscene/ffmpeg/FFMPEGDecoder.cpp
```diff
@@ -55,48 +58,48 @@ double getFrameRate(AVStream* stream, AVCodecContext* codecCtx) {
...
-    if (!stricmp(langauge_name, "spa")) {
-        return "Spanish";
-    }
-
-    // Print to log so that we can find the actual value more easily later
-    mprintf(("FFmpeg log: Found unknown language value '%s'!\n", langauge_name));
-
-    // Default to english for everything else
-    return "English";
+SCP_string normalizeLanguage(const char *langauge_name) {
```
Goober5000: Same here.

code/cutscene/ffmpeg/FFMPEGDecoder.cpp
```diff
@@ -165,7 +169,7 @@ std::unique_ptr<DecoderStatus> initializeStatus(std::unique_ptr<InputStream>& st
 	if (audioStream < 0) {
 		if (audioStream == AVERROR_STREAM_NOT_FOUND) {
 			mprintf(("FFmpeg: No audio stream found in file!\n"));
-		} else if (videoStream == AVERROR_DECODER_NOT_FOUND) {
+		} else if (audioStream == AVERROR_DECODER_NOT_FOUND) {
```
Goober5000: It looks like this is a bugfix.  Is there a corresponding Mantis or GitHub issue, or HLP post, or is this just something you noticed?

code/cutscene/ffmpeg/FFMPEGDecoder.cpp
```diff
@@ -312,6 +339,27 @@ std::unique_ptr<InputStream> openInputStream(const SCP_string& name) {
 
 	return nullptr;
 }
+
+std::unique_ptr<InputStream> openSubtitleStream(const SCP_string &name) {
```
Goober5000: Here too with the function brace.

code/cutscene/ffmpeg/FFMPEGDecoder.cpp
```diff
@@ -342,11 +392,12 @@ bool FFMPEGDecoder::initialize(const SCP_string& fileName) {
 
 	// We're done, now just put the pointer into this
 	std::swap(m_input, input);
+	std::swap(m_subtitleInput, subt);
 	std::swap(m_status, status);
 	return true;
 }
 
-MovieProperties FFMPEGDecoder::getProperties() {
+MovieProperties FFMPEGDecoder::getProperties() const {
```
Goober5000: And here.

code/cutscene/ffmpeg/FFMPEGDecoder.cpp
```diff
@@ -454,6 +515,55 @@ void FFMPEGDecoder::close() {
 		// This will delete the InputStream pointer and free all data
 		m_input = nullptr;
 	}
+
+	if (m_subtitleInput) {
+		m_subtitleInput = nullptr;
+	}
+}
+void FFMPEGDecoder::runSubtitleDecoder(SubtitleDecoder *decoder) {
```
Goober5000: And here.

code/cutscene/ffmpeg/SubtitleDecoder.cpp
```diff
@@ -31,7 +31,25 @@ void SubtitleDecoder::decodePacket(AVPacket* packet) {
 	}
 }
 void SubtitleDecoder::finishDecoding() {
```
Goober5000: And here.

code/cutscene/ffmpeg/FFMPEGDecoder.cpp
asarium: It's a simple copy&paste error I noticed while editing the code. It's located inside a code block that is only used for error handling and it doesn't even change the end result so there is no reason to fix this in a separate PR.

asarium:
I used clang-format for these changes so the config file seems to be misconfigured. I'll fix that and reformat my changes.

asarium:
@Goober5000 I fixed the clang-format config and reformatted my changes. You had one comment on a line not changed by this PR so I left that without changes.

code/cutscene/ffmpeg/FFMPEGDecoder.cpp
Goober5000: I didn't mean to imply that this should be a separate PR, just wondering if it addressed an existing ticket.

Goober5000:
Fair enough.

PR #1752, opened 2018-06-07
Only upload mipmap levels that are actually needed

asarium:
Previously, the game uploaded all mipmap levels found in the image files
and then selected which ones it will use by using `GL_TEXTURE_BASE_LEVEL`.
This unnecessarily used additional GPU memory which is especially bad if
the player already needs to use a lower detail level. This fixes that by
only uploading the necessary mipmap levels.


SamuelCho:
This is not a good idea. Mips are still used in hardware when sampling texures in the shader depending on the triangle size. Mip maps reduce unnecessary cache misses since a pixel in the far off distance will most likely hit the same texel as it's neighboring fragments. Sure, this will save memory but it's really unusual to have a game not use hardware mip mapping.

asarium:
Uhh, that is not what this does (unless I did a major mistake).

All this does is making sure that only the mipmap levels that are actually going to be used are uploaded. Normally, the game uses all mipmaps in which case these changes will make no difference. In case the user has a lower texture detail selected, the engine currently uploads all mipmaps and then sets `GL_TEXTURE_BASE_LEVEL` to a value > 0 for selecting the needed base mipmap. That unnecessarily uses additional texture memory which can be improved by only allocating and uploading the mipmap levels that are actually going to be used by the engine.

PR #1755, opened 2018-06-13
Write full log path and add legacy mode reason to log file

asarium:
This should make it a bit easier to determine if FSO is using the new
config file or the old one from only a log file.

This also adds a log line which states the reason for why the legacy or
new config location was chosen which will make it easier to debug issues
with that code.

### DIFF
--- a/code/osapi/osapi.cpp
+++ b/code/osapi/osapi.cpp
@@ -349,6 +349,8 @@ static time_t get_file_modification_time(const SCP_string& path) {
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


ngld:
Looks great! This should help solve issues when FSO picks the wrong folder (admittedly, I've only seen it once but that's enough).

tomimaki:
Should be "legacy" in Osapi_legcy_mode_reason?

asarium:
Thanks, fixed.

PR #1757, opened 2018-06-21
Improve handling of the shader cache

asarium:
This adds an identification of the OpenGL implementation to the shader
content hash to avoid issues when using the same FSO install with
multiple graphics drivers.

Also adds support for querying the supported binary formats and
gracefully handling missing support for the format of a stored shader.
This can be useful when downgrading the version of a graphics driver
since the older version may not support the binary format yet.

### DIFF
--- a/code/graphics/opengl/gropenglshader.cpp
+++ b/code/graphics/opengl/gropenglshader.cpp
@@ -506,6 +506,9 @@ static SCP_string get_shader_hash(const SCP_vector<SCP_string>& vert,
 		md5.update(reinterpret_cast<const char*>(&i), sizeof(i));
 	}
 
+	md5.update(GL_implementation_id.data(),
+	           (MD5::size_type)GL_implementation_id.size() * sizeof(SCP_string::value_type));
+
 	md5.finalize();
 
 	return md5.hexdigest();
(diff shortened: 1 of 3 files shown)


code/graphics/opengl/gropenglshader.cpp
```diff
@@ -564,6 +567,20 @@ static bool load_cached_shader_binary(opengl::ShaderProgram* program, const SCP_
 	auto binary_format = (GLenum) format;
 	json_decref(metadata_root);
 
+	bool supported = false;
+	for (auto supported_fmt : GL_binary_formats) {
+		if (supported_fmt == binary_format) {
```
MageKing17: This comparison is apparently between a signed and an unsigned value; `GL_binary_formats` is a vector of `GLint`s while `binary_format` seems to be a `GLenum`; shouldn't those two types match? If nothing else, this comparison needs to not generate a warning for CI to pass.

asarium: `GL_binary_formats` is initialized from an OpenGL function which only returns `GLint` so that's the reason why it's not a vector of `GLenum`.

I added a cast to silence the warning.

PR #1764, opened 2018-06-27
Make sure that sound pitch range is initialized properly

asarium:
If a sound did not receive a valid sound name then the parsing function
would exit early and not set the pitch range properly. This would cause
a divide by zero if the duration of the sound was queried.

This moves the initialization of the pitch range to the start of the
parsing functions which makes sure that the range is initialized even if
the function exits early.


MageKing17:
This was causing mods that explicitly defined sound 200 (`BALLISTIC_START_LOAD`) as Empty to have support ships unable to actually start loading ballistic ammo, so they'd be stuck rearming (but not actually rearming) forever (because the Empty sound effect had a pitch of `0`, so the duration was `0/0`, which when converted to an `int` was negative, creating a timestamp that could never elapse).

PR #1767, opened 2018-06-28
Add AI profile flag to allow primary weapons to consider range.

MageKing17:
By enabling the "$Use Actual Primary Range:" profile flag (which can also be set on a per-AI-class basis), the code to determine if the AI should fire its primary weapons will not only perform its current "range based on speed of weapon and target" calculation, but it will also compute the weapon's actual range and not fire if the target is outside of it.

### DIFF
--- a/code/ai/ai_flags.h
+++ b/code/ai/ai_flags.h
@@ -110,6 +110,7 @@ namespace AI {
         Smart_subsystem_targeting_for_turrets,
         Strict_turred_tagged_only_targeting,
         Turrets_ignore_target_radius,
+        Use_actual_primary_range,
         Use_additive_weapon_velocity,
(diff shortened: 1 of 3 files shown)


asarium:
The new behaviour seems to make much more sense than the old so maybe it could be enabled by default in newer FSO versions (by using the target FSO version)? I don't know if that is supported by the AI profile parsing system though so this is just a suggestion in case that is easy to implement.

MageKing17:
It is easy to implement, but there are a number of such flags that I want to be enabled by default when 3.8.2 is targeted, so I'm planning on making a separate PR that makes those all happen.

asarium:
Yes, that makes sense.

PR #1769, opened 2018-06-28
Add $Failure Rate:

TRBlount:
This adds a probability based weapon failure system. By default, it simply cancels firing the weapon, otherwise it can substitute a weapon specified by the modder.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -11239,7 +11239,6 @@ int ship_fire_primary(object * obj, int stream_weapons, int force)
 								if (obj == Player_obj) {
 									ship_maybe_play_primary_fail_sound();
 								}
-								//has_fired = true; // Just because it failed, doesn't mean it wouldn't have fired
 								continue;
 							}
 
@@ -11961,6 +11960,14 @@ int ship_fire_secondary( object *obj, int allow_swarm )
 			// create the weapon -- for multiplayer, the net_signature is assigned inside
 			// of weapon_create
 			weapon_num = weapon_create( &firing_pos, &firing_orient, weapon_idx, OBJ_INDEX(obj), -1, aip->current_target_is_locked);
+
+			if (weapon_num == -1) {
+				// Weapon most likely failed to fire
+				if (obj == Player_obj) {
+					ship_maybe_play_secondary_fail_sound(wip);
+				}
+				continue;
+			}
 
 			if (weapon_num >= 0) {
 				weapon_idx = Weapons[Objects[weapon_num].instance].weapon_info_index;
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -10898,7 +10898,6 @@ int ship_fire_primary(object * obj, int stream_weapons, int force)
 			ship_stop_fire_primary_bank(obj, bank_to_fire);
 			continue;
 		}		
-		
 
 		if ( pm->n_guns > 0 ) {
 			int num_slots = pm->gun_banks[bank_to_fire].num_slots;
@@ -11234,6 +11233,16 @@ int ship_fire_primary(object * obj, int stream_weapons, int force)
 							// of weapon_create							
 							weapon_objnum = weapon_create( &firing_pos, &firing_orient, weapon_idx, OBJ_INDEX(obj), new_group_id,
 								0, 0, swp->primary_bank_fof_cooldown[bank_to_fire] );
+
+							if (weapon_objnum == -1) {
+								// Weapon most likely failed to fire
+								if (obj == Player_obj) {
+									ship_maybe_play_primary_fail_sound();
+								}
+								//has_fired = true; // Just because it failed, doesn't mean it wouldn't have fired
+								continue;
+							}
+
 							winfo_p = &Weapon_info[Weapons[Objects[weapon_objnum].instance].weapon_info_index];
 							has_fired = true;
 
--- a/code/weapon/weapon.h
+++ b/code/weapon/weapon.h
@@ -448,6 +448,11 @@ typedef struct weapon_info {
 	int				targeting_priorities[32];
 	int				num_targeting_priorities;
 
+	// Optional weapon failures
+	float failure_rate;
+	char failure_sub_name[NAME_LENGTH];
+	int failure_sub;
+
 	// the optional pattern of weapons that this weapon will fire
 	size_t			num_substitution_patterns;
 	int				weapon_substitution_pattern[MAX_SUBSTITUTION_PATTERNS]; //weapon_indexes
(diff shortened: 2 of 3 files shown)


code/weapon/weapons.cpp
```diff
@@ -5106,6 +5139,21 @@ int weapon_create( vec3d * pos, matrix * porient, int weapon_type, int parent_ob
 		}
 	}
 
+	// Let's setup a fast failure check with a uniform distribution.
+	if (wip->failure_rate > 0.0f) {
+		std::random_device rd;
+		std::mt19937 f(rd());
+		std::uniform_real_distribution<float> gen(0.0f, 1.0f);
```
asarium: You should use `UniformFloatRange` here since that does exactly what you need without all the additional code.

code/weapon/weapon.h
```diff
@@ -448,6 +448,11 @@ typedef struct weapon_info {
 	int				targeting_priorities[32];
 	int				num_targeting_priorities;
 
+	// Optional weapon failures
+	float failure_rate;
+	char failure_sub_name[NAME_LENGTH];
```
asarium: This should be an `SCP_string` unless there is a very good reason to use a raw `char` array.

code/weapon/weapons.cpp
```diff
@@ -3380,6 +3390,29 @@ void weapon_generate_indexes_for_substitution() {
 
 			memset(wip->weapon_substitution_pattern_names, 0, sizeof(char) * MAX_SUBSTITUTION_PATTERNS * NAME_LENGTH);
 		}
+
+		if (wip->failure_rate > 0.0f) {
+			if (stricmp("none", wip->failure_sub_name) != 0) {
```
asarium: You could use `VALID_FNAME` here to ensure consistent behavior with other parts of the code.

code/weapon/weapons.cpp
```diff
@@ -7770,6 +7818,9 @@ void weapon_info::reset()
 	this->thruster_glow_factor = 1.0f;
 	this->target_lead_scaler = 0.0f;
 
+	this->failure_rate = 0.0f;
```
asarium: This block does not initialize `failure_sub_name` which is probably needed here. Even if it is not strictly needed it is better to explicitly reset all fields.

code/weapon/weapons.cpp
```diff
@@ -3380,6 +3390,29 @@ void weapon_generate_indexes_for_substitution() {
...
+					Warning(LOCATION, "Weapon '%s' requests substitution with '%s' which does not seem to exist",
+						wip->name, wip->failure_sub_name);
+					wip->failure_rate = 0.0f;
+				}
+
+				if (Weapon_info[wip->failure_sub].subtype != wip->subtype) {
+					// Check to make sure secondaries can't be launched by primaries and vice versa
+					Warning(LOCATION, "Weapon '%s' requests substitution with '%s' which is of a different subtype.",
+						wip->name, wip->failure_sub_name);
+					wip->num_substitution_patterns = 0;
```
asarium: Why do you reset `num_substitution_patterns` here? At first glance this has nothing to do with the failure rate feature.

code/weapon/weapons.cpp
```diff
@@ -3380,6 +3390,29 @@ void weapon_generate_indexes_for_substitution() {
...
+					wip->failure_sub = -1;
+					wip->failure_rate = 0.0f;
+				}
+			}
+
+			memset(wip->failure_sub_name, 0, sizeof(char) * NAME_LENGTH);
```
asarium: This should use `sizeof(wip->failure_sub_name)` instead of `sizeof(char) * NAME_LENGTH` in case the array size is changed later.

code/ship/ship.cpp
```diff
@@ -11234,6 +11233,16 @@ int ship_fire_primary(object * obj, int stream_weapons, int force)
 							// of weapon_create							
 							weapon_objnum = weapon_create( &firing_pos, &firing_orient, weapon_idx, OBJ_INDEX(obj), new_group_id,
 								0, 0, swp->primary_bank_fof_cooldown[bank_to_fire] );
+
+							if (weapon_objnum == -1) {
```
asarium: This only handles the case when a primary weapon is fired but not secondaries. Does this feature only support primary weapons? I do not see anything in the new parsing code that would enforce this.

code/ship/ship.cpp
```diff
@@ -11234,6 +11233,16 @@ int ship_fire_primary(object * obj, int stream_weapons, int force)
...
+								// Weapon most likely failed to fire
+								if (obj == Player_obj) {
+									ship_maybe_play_primary_fail_sound();
+								}
+								//has_fired = true; // Just because it failed, doesn't mean it wouldn't have fired
```
asarium: Why is this line commented out?

TRBlount: I'm still waiting on Axem to get back to me on whether a weapon failure should trigger the `On_Weapon_Fired` scripting hook.

code/ship/ship.cpp
TRBlount: This was an oversight. Spoon specified this specifically for primary weapons so I overlooked secondary weapons. I will fix that later today.

code/weapon/weapons.cpp
MageKing17: This line probably shouldn't have been copied here.

code/weapon/weapons.cpp
MageKing17: Or, if `failure_sub_name` becomes an `SCP_string`, it shouldn't use `memset()` at all.

code/ship/ship.cpp
asarium: Ok. In that case you should probably add the "unfinished" label so that it is visible that these changes are not final yet.

TRBlount:
Alright, I went through all the feedback and fixed the code in response.

code/weapon/weapons.cpp
asarium: It's in `utils/RandomRange.h`.

code/weapon/weapons.cpp
```diff
@@ -3380,6 +3388,28 @@ void weapon_generate_indexes_for_substitution() {
 
 			memset(wip->weapon_substitution_pattern_names, 0, sizeof(char) * MAX_SUBSTITUTION_PATTERNS * NAME_LENGTH);
 		}
+
+		if (wip->failure_rate > 0.0f) {
+			if (VALID_FNAME(wip->failure_sub_name)) {
+				wip->failure_sub = weapon_info_lookup(wip->failure_sub_name.c_str());
+
+				if (wip->failure_sub == -1) { // invalid sub weapon
+					Warning(LOCATION, "Weapon '%s' requests substitution with '%s' which does not seem to exist",
```
asarium: This will probably cause a format string warning since you use an `SCP_string` as a format argument.

TRBlount:
All checks have passed. LGTM?

PR #1770, opened 2018-06-29
Add mod table options for specifying lighting preset

asarium:
This will make it possible to specify the lighting parameters from a mod
which will then override the values specified on the command line. This
is meant to be used by mods which have some specific requirements on how
the lighting should look like.

One possible issue here is that this might make changes to the lighting system
break backwards compatibility since the lighting options can now be set in
a table. I don't think that this is a big issue but I wanted to mention it here.


asarium:
I don't think this is needed anymore. The new options system will allow modders to specify overrides for specific options which would include support for the lighting preset.

Goober5000:
Closing, since @asarium said this isn't needed anymore.

PR #1772, opened 2018-06-30
CMake: Unify how source files are added to CMake

asarium:
This unifies how our CMake system handles adding source files to the
project so that all subprojects use the same mechanism. This also makes
it easier to add a new source folder to the code project since you don't
need to editor three places at the same time anymore.

I still need to check if the folders are correct in Visual Studio.


asarium:
Visual Studio seems to be working fine with these changes.

PR #1774, opened 2018-06-30
Remove "not enough points in path" warning

The-E:
Change this warning to a log print. This is only an issue when it's an issue, and most of the time it's not

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -10719,8 +10719,8 @@ void ai_dock()
 			char *goal_ship_class_name = goal_sip->name;
 			char *goal_dock_path_name = model_get(goal_sip->model_num)->paths[aip->mp_index].name;
 
-			Warning(LOCATION, "Ship class %s has only %i points on dock path \"%s\".  Recommended minimum number of points is 4.  "\
-				"Docking along that path will look strange.  You may wish to edit the model.", goal_ship_class_name, aip->path_length, goal_dock_path_name);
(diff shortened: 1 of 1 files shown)


Goober5000:
Why was this approved?  This has been an issue on every mod where I've seen the warning appear.  (I remember that a large number of fighters and bombers had this issue when I worked on the Inferno Classic rerelease a few months ago.)  If the path has too few points, the docking ship will start slewing too early, causing docking to be unnecessarily delayed, or will clip through the ship it is docking to.

The purpose of warnings is to alert mod designers to fix incorrect data in their mod, and bad docking paths are certainly worth fixing.  Warnings are only shown during debug builds, so end-users playing release builds will not see them.

asarium:
Warnings are typically used for situations where the engine has detected that something is wrong with the mod data and corrected it. In this case no action is taken by the engine which makes it unusual to use a warning here. The warning is still there in the debug log so a mod author is still being notified that there *might* be an issue.

Goober5000:
Warnings are used regardless of whether the engine has corrected the problem or not.  Sometimes the correction must be done manually by the mod designer, and this is one such case.

The engine still takes action in that it tries to accommodate the shortened path.  But as I said, this often leads to unnaturally slewing or too-close clipping.  The first case can interfere with event timing in the mission; the second case leads to blatant visual errors.

asarium:
Again, it *can* lead to visual error but it doesn't have to. If the path happens to look ok then this is a false positive that would annoy modders needlessly.

PR #1776, opened 2018-07-01
Downgrade null vec3d warning to a log print

The-E:
This is definitely open for discussion, but here's my reasoning:

1. This warning is something a modder cannot do anything about
2. Even for coders, it is hard to impossible to actually figure out the cause of most of these
3. Therefore, interrupting gameplay during debugging for this is not warranted, in my opinion.


Karajorma:
Seems like a sensible change. I've had the game stop cause of this warning quite a few times and there's really not much you can do about it as a modder.

PR #1780, opened 2018-07-06
Retry prebuilt library download if it fails

asarium:
This has been a problem for nighty and release builds which should be
fixed by this.

### DIFF
--- a/lib/prebuilt.cmake
+++ b/lib/prebuilt.cmake
@@ -35,17 +35,30 @@ function(get_prebuilt_path OUT_VAR)
     endif()
     set(DOWNLOAD_URL "https://github.com/scp-fs2open/scp-prebuilt/releases/download/${TAG_NAME}/${FILENAME}")
     set(DOWNLOAD_FILE "${CURRENT_ROOT}/${FILENAME}")
-    
-    message(STATUS "Downloading prebuilt libraries from \"${DOWNLOAD_URL}\"")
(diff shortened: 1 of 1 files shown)


lib/prebuilt.cmake
```diff
@@ -35,17 +35,30 @@ function(get_prebuilt_path OUT_VAR)
...
-    file(DOWNLOAD "${DOWNLOAD_URL}" "${DOWNLOAD_FILE}" SHOW_PROGRESS TLS_VERIFY ON STATUS DOWNLOAD_STATUS_LIST)
-    
-    list(GET DOWNLOAD_STATUS_LIST 0 DOWNLOAD_STATUS)
-    list(GET DOWNLOAD_STATUS_LIST 1 DOWNLOAD_ERROR)
+
+    set(MAX_RETRIES 5)
+    foreach(i RANGE 1 ${MAX_RETRIES})
+        if (NOT (i EQUAL 1))
+            message(STATUS "Retry after 5 seconds (attempt #${i}) ...")
+            execute_process(COMMAND "${CMAKE_COMMAND}" -E sleep "5")
```
MageKing17: An optional (slight) improvement might be to double the wait time with every retry, so as to potentially give the server more of a chance to, idk, recover or whatever? Might just be a complication for no benefit, though.

asarium: I though about it but mathematical operations are a pain to use in CMake so I did not implement it here. I don't think it would be a big issue anyway since the few instances where this is needed, the second try should work regardless of when it is executed.

PR #1784, opened 2018-07-07
Make parse_weapon_targeting_priorities() work when weapon missing.

MageKing17:
Previously, if `parse_weapon_targeting_priorities()` couldn't find a weapon matching the name, it wouldn't parse the "+Target Priority:" line, so it would cause a parsing error when the parser failed to find "$Name:" or "#End" there. Now, the "match not found" codepath still parses it so that no mismatch occurs. Also, I reordered the function so that the parsing wasn't happening in the middle of a loop through the list of weapon types, because it made me cringe.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -18364,7 +18364,7 @@ void parse_weapon_targeting_priorities()
 			}
 		}
 		if(k == MAX_WEAPON_TYPES) {
-			Warning(LOCATION, "Unrecognized weapon '%s' found when setting weapon targeting priorities.\n", tempname);
+			error_display(0, "Unrecognized weapon '%s' found when setting weapon targeting priorities.\n", tempname);
 			if (optional_string("+Target Priority:")) {		// consume the data to avoid parsing errors
 				SCP_vector<SCP_string> dummy;
 				stuff_string_list(dummy);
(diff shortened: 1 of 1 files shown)


code/ship/ship.cpp
```diff
@@ -18359,39 +18359,45 @@ void parse_weapon_targeting_priorities()
 		for(k = 0; k < MAX_WEAPON_TYPES ; k++) {
 			if ( !stricmp(Weapon_info[k].name, tempname) ) {
 				// found weapon, yay!
-				// reset the list
+				// no need to keep searching for more
+				break;
+			}
+		}
+		if(k == MAX_WEAPON_TYPES) {
+			Warning(LOCATION, "Unrecognized weapon '%s' found when setting weapon targeting priorities.\n", tempname);
```
asarium: This should use `error_display` to make finding the error easier.

code/ship/ship.cpp
```diff
@@ -18359,39 +18359,45 @@ void parse_weapon_targeting_priorities()
...
+						if ( !stricmp(Ai_tp_list[j].name, tgt_priorities[i].c_str()))  {
+							wip->targeting_priorities[i] = j;
+							wip->num_targeting_priorities++;
+							break;
 						}
-						if(j == num_groups)
-							Warning(LOCATION, "Unrecognized string '%s' found when setting weapon targeting priorities.\n", tgt_priorities[i].c_str());
 					}
+					if(j == num_groups)
+						Warning(LOCATION, "Unrecognized string '%s' found when setting weapon targeting priorities.\n", tgt_priorities[i].c_str());
```
asarium: Same here.

MageKing17:
Sorry about the delay; I made the changes almost immediately but then forgot to actually commit and push them.

PR #1792, opened 2018-07-12
Shockwaves caused by weapons respect big ship flags while calculating damage

fjelliott:
Shockwaves caused by weapons will no longer deal lethal damage if the weapon that caused the shockwave isn't flagged as being able to deal lethal damage to the ship in question. This is accomplished in the same way it is accomplished in weapon_get_damage_scale from weapons.cpp; if the ship's hull drops below a certain threshold (75% for supercap ships, 10% for all other big damage ships), the maybe_shockwave_damage_adjust function sets the damage that should be applied to zero. There is also some damage scaling that goes on that mirrors the damage scaling done in weapon_get_damage_scale.

For instance this means it's no longer possible to kill a Sathanas with helios torpedoes. It is also impossible to kill smaller ships with the big damage flag (such as corvettes) with smaller missiles that happen to cause shockwaves but aren't bombs (such as the EMP Adv.). Before this was supposed to be impossible, but in reality just took a really long time.

These changes do not affect shockwaves that aren't caused by weapons (such as ship explosions). So kamikaze ships that cause shockwaves and meson bomb explosions will still work as they did before.

### DIFF
--- a/code/ship/shiphit.cpp
+++ b/code/ship/shiphit.cpp
@@ -1842,6 +1842,8 @@ static int maybe_shockwave_damage_adjust(object *ship_objp, object *other_obj, f
 {
 	ship_subsys *subsys;
 	ship *shipp;
+	weapon_info *wip;
+	ship_info *sip;
 	float dist, nearest_dist = FLT_MAX;
 	vec3d g_subobj_pos;
 	float max_damage;
@@ -1860,11 +1862,12 @@ static int maybe_shockwave_damage_adjust(object *ship_objp, object *other_obj, f
 		return 0;
 	}
 
-	if (!(Ship_info[Ships[ship_objp->instance].ship_info_index].is_huge_ship())) {
+	shipp = &Ships[ship_objp->instance];
+	sip = &Ship_info[shipp->ship_info_index];
+
+	if (!sip->is_huge_ship()) {
 		return 0;
 	}
-
-	shipp = &Ships[ship_objp->instance];
 
 	// find closest subsystem distance to shockwave origin
 	for (subsys=GET_FIRST(&shipp->subsys_list); subsys != END_OF_LIST(&shipp->subsys_list); subsys = GET_NEXT(subsys) ) {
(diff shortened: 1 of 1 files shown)


code/ship/shiphit.cpp
```diff
@@ -1882,6 +1885,39 @@ static int maybe_shockwave_damage_adjust(object *ship_objp, object *other_obj, f
...
+
+		// First handle Supercap ships.
+		if ((sip->flags[Ship::Info_Flags::Supercap]) && !(wip->wi_flags[Weapon::Info_Flags::Supercap])) {
+			if (hull_pct <= 0.75f) {
+				*damage = 0.0f;
+				return 1;
+			} else {
+				// If hull isn't below 3/4, then allow damage to be applied just like in weapon_get_damage_scale.
+				// Magic constant is SUPERCAP_DAMAGE_SCALE from weapons.cpp.
+				max_damage *= 0.25f;
```
MageKing17: This should be the aforementioned constant rather than a magic number with the same value.

MageKing17:
Whether or not the mod table option should be enabled by default or not is a question that can be decided later, in consultation with the modding community; there'll be a later PR to make various options be auto-enabled for mods targeting 3.8.2.

code/ship/shiphit.cpp
fjelliott: Where should I define said constant? It's defined in weapons.cpp and not in the header so I had to copy it. Should it be in shiphit.cpp or in shiphit.h?

MageKing17: obviously, you move it to the header so it can be reused

fjelliott: Should I add a prefix to it since the macro will then be global? Like, should I change it to WEAPON_H_SUPERCAP_DAMAGE_SCALE  ?

MageKing17: It shouldn't cause any ambiguity, so no.

fjelliott:
I added the changes MageKing17 requested. SUPERCAP_DAMAGE_SCALE is now defined in weapon.h instead of weapons.cpp and I reference the constant in maybe_shockwave_damage_adjust in shiphit.cpp.

Issue #1738, opened 2018-06-03
Collection of Strings that should, but are not translateable yet


NovachenFS2:
During the Test of #1736, i encountered a string, that still can not be translated.

So i looked through the code and collected some strings that are used in the game, but are not translateable, because they are missing a XSTR in general for the strings.tbl or have an ID of -1.
Of course this collection does not includes strings, which were commented out.
For a complete translateable game i think it would be helpful, that these strings could be translated as well due to the strings.tbl.

barracks.cpp - Line 436
`STRCPY1(Stat_labels[Num_stat_lines], "Current Score:");`

This text is displayed in the barracks screen in the Pilot info.

hudtargetbox.cpp - Line 1037
`sprintf(outstr, NOX("impact: %.1f sec"), dist/speed);`

This text is displayed if you target a bomb. It shows how long it takes to reach its target.

aigoals.cpp - Lines 127 to 128
`	case AI_GOAL_FLY_TO_SHIP:
		return XSTR( "rendezvous with ", -1);`
This one is used in the target box as well. When a AI ship uses a Autopilot this text appears. Or at least that is, what i know about it. I never encoutered it myself, because i never played WCS, which is the most prominent mod for this feature ^^.

hud/hud.cpp - Line 2898
`sprintf(outstr, "Waiting...");`

This one appears after the Support ship has docked with the player ship, but not have started the rearming procedure.

keycontrol.cpp - Line 1070
`HUD_sourced_printf(HUD_SOURCE_HIDDEN, XSTR("Issuing rearm request for %s", -1), Ships[obj_to_rearm->instance].ship_name);`

If i remember it correctly, that text is part of a SCP exclusive cheat, where you can force the Support ship to rearm/repair a specific ship. So it is a System message that appear if you use this cheat.

keycontrol.cpp - Line 1075
`HUD_sourced_printf(HUD_SOURCE_HIDDEN, XSTR("Cannot issue rearm request for %s", -1), Ships[obj_to_rearm->instance].ship_name);`

The same, only that this is the message that appears, if it does not work for the targeted object.

multiui.cpp - Lines 4976 to 4979
`								XSTR("A time limit is being used in a co-op game.\r\n"
									"  Select \'Cancel\' to go back to the mission select screen.\r\n"
									"  Select \'Yes\' to continue with this time limit.\r\n"
									"  Select \'No\' to continue without this time limit.", -1));`

This message appears, if you try to start a multiplayer coop mission with a time limit.

[...]

PR #1795, opened 2018-07-14
Strings.tbl translation adjustments

NovachenFS2:
In the code there are some strings from features added by the SCP that uses two different IDs for the same string. This PR cleans them up, so that only one ID is used.

The IDs that were occupied by duplicates aswell IDs that were already unused before are now replaced by previously untranslated strings, that are now translateable through the strings.tbl.

This contain all requests i made in #1738. except from the bindable controls.

### DIFF
--- a/code/menuui/barracks.cpp
+++ b/code/menuui/barracks.cpp
@@ -433,7 +433,7 @@ void barracks_init_stats(scoring_struct *stats)
 	Num_stat_lines++;
 
 	Assert(Num_stat_lines < Max_stat_lines);
-	STRCPY1(Stat_labels[Num_stat_lines], "Current Score:");
+	STRCPY1(Stat_labels[Num_stat_lines], XSTR( "Current Score:", 1583));
 	sprintf(Stats[Num_stat_lines], "%d", stats->score);
 	Num_stat_lines++;
 
--- a/code/menuui/barracks.cpp
+++ b/code/menuui/barracks.cpp
@@ -433,7 +433,7 @@ void barracks_init_stats(scoring_struct *stats)
 	Num_stat_lines++;
 
 	Assert(Num_stat_lines < Max_stat_lines);
-	STRCPY1(Stat_labels[Num_stat_lines], XSTR( "Current Score:", 1583));
+	strcpy_s(Stat_labels[Num_stat_lines], XSTR( "Current Score:", 1583));
 	sprintf(Stats[Num_stat_lines], "%d", stats->score);
 	Num_stat_lines++;
 
--- a/code/network/multiui.cpp
+++ b/code/network/multiui.cpp
@@ -4976,7 +4976,7 @@ void multi_create_accept_hit()
 								XSTR("A time limit is being used in a co-op game.\r\n"
 									"  Select \'Cancel\' to go back to the mission select screen.\r\n"
 									"  Select \'Yes\' to continue with this time limit.\r\n"
-									"  Select \'No\' to continue without this time limit.", -1));
+									"  Select \'No\' to continue without this time limit.", 1620));
 
 			if (popup_choice == 0) {
 				return;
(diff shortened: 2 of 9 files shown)


The-E:
Seems like we need to rewrite the XSTR macro to make the compilers happy

code/network/multiui.cpp
```diff
@@ -4976,7 +4976,7 @@ void multi_create_accept_hit()
 								XSTR("A time limit is being used in a co-op game.\r\n"
 									"  Select \'Cancel\' to go back to the mission select screen.\r\n"
 									"  Select \'Yes\' to continue with this time limit.\r\n"
-									"  Select \'No\' to continue without this time limit.", -1));
+									"  Select \'No\' to continue without this time limit.", 1620));
```
asarium: You have reused the id `1620` from another XSTR but that would likely not work in another language with retail since FSO would then use the wrong translated string. You need to make sure that all IDs you are using are new to avoid breaking retail or other translation mods.

code/menuui/barracks.cpp
```diff
@@ -433,7 +433,7 @@ void barracks_init_stats(scoring_struct *stats)
 	Num_stat_lines++;
 
 	Assert(Num_stat_lines < Max_stat_lines);
-	STRCPY1(Stat_labels[Num_stat_lines], "Current Score:");
+	STRCPY1(Stat_labels[Num_stat_lines], XSTR( "Current Score:", 1583));
```
asarium: You can just replace `STRCPY1` with `strcpy_s` here to fix the CI failure.

code/network/multiui.cpp
NovachenFS2: I am sorry. What you mean exactly? What have this change to do with retail? Is this ID already used in the Retail code aswell? 😮  I thought Retail is using IDs only up to 1589, because that is the last entry, that is used in the retail strings.tbl for all languages. Then i will go through the retail code aswell to check that.

code/menuui/barracks.cpp
NovachenFS2: Well, i have to admit, that this has irritated me a bit, because STRCPY1 is used for all other XSTRs in that area well and it only complains about this one. Will change that, of course.

code/network/multiui.cpp
asarium: The problem I see here is that the string `"You have selected TCP/IP for multiplayer ..."` from `multi.cpp` is assigned to the XSTR ID 1620 at the moment. That might not be a valid ID for retail but a mod may have used it for translating this new message.

You assigned this message the new ID 1604 and then  used 1620 for the message `"A time limit is being used in a co-op ..."`. So now if a mod has translated what used to be ID 1620 based on the old string (`"You have selected TCP/IP for multiplayer ..."`), then the translated string would be used here where you now use the ID 1620.

My point is that you should not change the ID of existing `XSTR` calls (except if they refer to a duplicate entry then it's obviously OK) to avoid breaking the translation of old mods.

code/menuui/barracks.cpp
asarium: The check that showed the error message only runs on changed lines so that is the reason why it didn't show any errors on any of the other `STRCPY1` calls.

code/network/multiui.cpp
NovachenFS2: That is the point, this 1620 was a duplicate. Line 338 in the mainhallmenu.cpp is using the same string with ID 1604.

And yes, i thought about the compatibility to other translation mods. But actually i was not able to find any, that is using more than the Retail 1589. Actually i am very certain, that there is any translation mod, that is using the additional strings by the SCP.

asarium: If there are no mods that translate this ID (and you are right, there are practically no translation mods) then reassigning this ID wouldn't matter.

If we actually find a mod that translates this then either that mod could be fixed or the ID could be changed back to make it work again.

PR #1805, opened 2018-07-21
Add scripting API for accessing parsing system

asarium:
This allows scripts to use the same parsing functions which are also
used by the engine for parsing tables.

### DIFF
--- a/code/scripting/api/libs/parse.cpp
+++ b/code/scripting/api/libs/parse.cpp
@@ -60,7 +60,7 @@ ADE_FUNC(displayMessage, l_Parsing, "string message, boolean error = false",
          "message will be displayed as an error.",
          "boolean", "true if the operation was successful, false otherwise")
 {
-	if (Mission_text == nullptr) {
+	if (Parse_text == nullptr) {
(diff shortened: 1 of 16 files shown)


code/scripting/api/libs/parse.cpp
```diff
@@ -0,0 +1,205 @@
...
+namespace api {
+
+ADE_LIB(l_Parsing, "Parsing", "parse", "Engine parsing library");
+
+ADE_FUNC(readFileText, l_Parsing, "string file, string directory = <any>",
+         "Reads the text of the given file into the parsing system. If a directory is given then the file is read from "
+         "that location.",
+         "boolean", "true if the operation was successful, false otherwise")
+{
+	const char* file;
```
The-E: It may be necessary to call pause_parse/unpause_parse here if the parser already has a file loaded.

asarium: I am pretty sure that there is no case where a scripting hook is called while another file is being parsed since the first hook (Game Init) is called after all data has been parsed.

Also, `pause_parse` would require us to manage the text data ourself instead of letting the parse system handle that.

Issue #1802, opened 2018-07-18
Inserting a semicolon into FRED's designer notes invalidates the mission


plieblang:
If there is a semicolon in the mission notes and no following newline, FRED is unable to load the mission.  From what I can tell, this is because `$End Notes:` is parsed as part of a comment and thus is ignored, making the mission file invalid.
To reproduce, add a line containing a semicolon to the designer notes section and make sure there is no following line (this includes blank lines).

asarium:
Thanks to @plieblang, this should be fixed by #1803.

PR #1808, opened 2018-07-23
Improve parsing of mission info text fields

plieblang:
Currently, quotation marks play havoc when entered into mission info text fields like title, author name, etc.  An odd number of them renders the mission invalid (and FRED does not give a helpful error, so it can be confusing what the problem is) and an even number causes the game and FRED to ignore everything after the first mark.  Because of this, I don't think there's any problem is just stripping them, which is what this PR does.
Also, I improved handling of #1802 (the designer notes semicolon issue) by writing a newline only if necessary.  This prevents a bunch of useless newlines from accumulating every time the mission is saved.
Finally, the git diff for `management.h` is a mess because of the clang formatter, but I only added the prototypes for `strip_quotation_marks` and `pad_with_newline` at lines 68 and 69.

### DIFF
--- a/qtfred/src/mission/Editor.cpp
+++ b/qtfred/src/mission/Editor.cpp
@@ -3190,7 +3190,7 @@ void Editor::strip_quotation_marks(SCP_string& str) {
 	}
 }
 
-void Editor::pad_with_newline(SCP_string& str, int max_size) {
+void Editor::pad_with_newline(SCP_string& str, SCP_string::size_type max_size) {
 	if (str.back() != '\n' && str.size() < max_size) {
 		str += "\n";
 	}
--- a/qtfred/src/mission/Editor.cpp
+++ b/qtfred/src/mission/Editor.cpp
@@ -1,46 +1,47 @@
 #include "Editor.h"
 
 #include <array>
-#include <vector>
-#include <stdexcept>
 #include <clocale>
+#include <stdexcept>
+#include <vector>
 
 #include <SDL.h>
-#include <sound/audiostr.h>
-#include <parse/parselo.h>
-#include <missionui/fictionviewer.h>
-#include <mission/missiongoals.h>
-#include <asteroid/asteroid.h>
-#include <jumpnode/jumpnode.h>
-#include <util.h>
-#include <mission/missionmessage.h>
-#include <gamesnd/eventmusic.h>
-#include <starfield/nebula.h>
-#include <object/objectdock.h>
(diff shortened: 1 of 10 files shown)


qtfred/src/mission/Editor.cpp
```diff
@@ -3183,6 +3183,19 @@ void Editor::normalizeShieldSysData() {
...
 
+void Editor::strip_quotation_marks(SCP_string& str) {
+	SCP_string::size_type idx = 0;
+	while ((idx = str.find('\"', idx)) != SCP_string::npos) {
+		str.erase(idx, 1);
+	}
+}
+
+void Editor::pad_with_newline(SCP_string& str, SCP_string::size_type max_size) {
+	if (str.back() != '\n' && str.size() < max_size) {
```
asarium: You handled an empty string in the FRED implementation but here you use `str.back()` without checking the size first which might cause issues if the string is empty.

qtfred/src/mission/Editor.cpp
```diff
@@ -3183,6 +3183,19 @@ void Editor::normalizeShieldSysData() {
...
 }
 
+void Editor::strip_quotation_marks(SCP_string& str) {
+	SCP_string::size_type idx = 0;
+	while ((idx = str.find('\"', idx)) != SCP_string::npos) {
+		str.erase(idx, 1);
+	}
+}
+
+void Editor::pad_with_newline(SCP_string& str, SCP_string::size_type max_size) {
```
asarium: While `SCP_string::size_type` is obviously correct here it would be simpler to simply use `size_t` to make the code more readable since the size_type will almost always be size_t (and if it isn't there are going to be a lot of other issues).

qtfred/src/mission/dialogs/CustomWingNamesDialogModel.cpp
```diff
@@ -13,6 +13,18 @@ CustomWingNamesDialogModel::CustomWingNamesDialogModel(QObject * parent, EditorV
 
 bool CustomWingNamesDialogModel::apply() {
 	int i;
+	for (i = 0; i < MAX_STARTING_WINGS; i++) {
```
asarium: clang-tidy will probably tell you that this (and the loops below) could be converted to for-each loops.

plieblang:
For convenience's sake, the full overall diff is at
https://github.com/scp-fs2open/fs2open.github.com/compare/master...plieblang:MissionInfoParsing

PR #1825, opened 2018-08-11
qtfred asteroid editor

niffiwan:
As it says on the tin; a mostly-direct port of the fred code to qtfred


qtfred/src/mission/dialogs/AsteroidEditorDialogModel.cpp
```diff
@@ -0,0 +1,458 @@
...
+		case _O_MAX_Y: return _max_y;
+		case _O_MAX_Z: return _max_z;
+		case _I_MIN_X: return _inner_min_x;
+		case _I_MIN_Y: return _inner_min_y;
+		case _I_MIN_Z: return _inner_min_z;
+		case _I_MAX_X: return _inner_max_x;
+		case _I_MAX_Y: return _inner_max_y;
+		case _I_MAX_Z: return _inner_max_z;
+		default:
+			Error(LOCATION, "Get a coder! Unknown enum value found! %i", type);
```
asarium: You could use the `UNREACHABLE` macro here instead.

PR #1831, opened 2018-08-17
make sure getMissionFilename never has a .fs2 extension

Goober5000:
Wookieejedi and Axem noticed that mn.getMissionFilename() sometimes returns the filename with the .fs2 extension, and sometimes not depending if you are in a campaign or tech room.  This PR ensures that the extension will always be included.

Presumably we do actually want to make them consistent, due to the principle of least astonishment, but do we want to standardize on having the extension, or not having the extension?  Pinging @AxemP for his input.

### DIFF
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -878,7 +878,15 @@ ADE_FUNC(createWeapon, l_Mission, "[weaponclass Class=WeaponClass[1], orientatio
 
 ADE_FUNC(getMissionFilename, l_Mission, NULL, "Gets mission filename", "string", "Mission filename, or empty string if game is not in a mission")
 {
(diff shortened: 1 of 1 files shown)


AxemP:
My opinion is that the mission's file name should **not** have the file extension returned. Most other areas of FreeSpace are, more or less, file extension agnostic. Sound files work if you've got ogg or wav, textures work with dds/tga/pcx/jpg. The engine should be taking care that it's got the right file extension and let the coders and modders just deal with only a filename they are expecting.

My script's work around right now is just to look if the filename ends in `.fs2` and toss that away. So I'm just going to be undoing exactly what this PR does all the time now. :P

Goober5000:
Makes sense.  Good rationale.  I flipped it around. :)

asarium:
Tea-CI had some internal issues with one of configurations but that had nothing to do with these changes so that failure can be ignored.

Issue #1826, opened 2018-08-14
Add $Ignore Mission File


MjnMixael:
Enhancement to add the ability to ignore single mission files like we can ignore campaign files in the game_settings.tbl. This gives mods the ability to make sure FSO removes unplayable single missions from the mod's Single Missions screen that may be in dependencies (such as the GOTY missions in some retail installations).

Issue #1666, opened 2018-03-23
Feature request: Ignorable multiplayer missions


Goober5000:
We already have `#Ignored Campaign File Names` in game_settings.tbl to prevent the campaign room from showing campaigns that are incompatible with the current mod.  It would be nice to have a similar feature to prevent the multiplayer room from listing incompatible multiplayer missions.

This would be useful for FSPort, and probably for other mods.

Goober5000:
I have a bit of time so I'll try to implement this.  I doubt it'll be too complicated.

Goober5000:
Implemented via PR #1832.

PR #1832, opened 2018-08-17
add game_settings feature to ignore single missions

Goober5000:
This works just like `#Ignored Campaign File Names`, just with missions.

```
#Ignored Mission File Names
$Mission File Name: 1.fs2
$Mission File Name: 2.fs2
```
etc.

This will only ignore standalone missions, not missions that are included in campaigns.  It works for both single-player and multiplayer.  I have tested it in both the tech room and the multiplayer "create game" screen.  Based on code searches, these are the only two places in the code where standalone missions are listed.

This PR addresses #1666 and its duplicate #1826.

### DIFF
--- a/code/mission/missionload.cpp
+++ b/code/mission/missionload.cpp
@@ -29,6 +29,8 @@ extern mission The_mission;  // need to send this info to the briefing
 extern int shifted_ascii_table[];
 extern int ascii_table[];
 
+SCP_vector<SCP_string> Ignored_missions;
+
 // -----------------------------------------------
 // For recording most recent missions played
 // -----------------------------------------------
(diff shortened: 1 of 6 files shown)


code/mission/missionload.cpp
```diff
@@ -76,6 +78,20 @@ void ml_update_recent_missions(char *filename)
 	Assert(Num_recent_missions <= MAX_RECENT_MISSIONS);
 }
 
+bool mission_is_ignored(const char *filename)
+{
+	SCP_string filename_no_ext = filename;
+	drop_extension(filename_no_ext);
+
+	for (auto &ii: Ignored_missions) {
+		if (ii == filename_no_ext) {
```
asarium: I think it would make more sense if this was a case insensitive comparison since the FSO file system is case insensitive as well.

Goober5000:
That makes sense, but I'm more surprised by the fact the warning caused the whole thing to fail.  Suppose I like being verbose and didn't want to use ranged-for?

code/mission/missionload.cpp
Goober5000: Ok.

asarium:
You can suppress all clang-tidy warnings by adding `// NOLINT` to the end of the line where the warning occurs. However, it is almost always a good idea to fix the clang-tidy warnings since things like range based for loops improve readability and performance.

PR #1841, opened 2018-08-28
add per-ship-class scanning range feature

niffiwan:
Requested for FotG - allow dedicated scanning ship options
e.g. Y-wing Longprobe & TIE/rc

### DIFF
--- a/code/hud/hudtarget.h
+++ b/code/hud/hudtarget.h
@@ -26,6 +26,7 @@ struct weapon_info;
 #define MATCH_SPEED_THRESHOLD				0.1f		// minimum speed target must be moving for match speed to apply
 #define CARGO_RADIUS_DELTA					100		// distance added to radius required for cargo scanning
 #define CAPITAL_CARGO_RADIUS_DELTA		250		// distance added to radius required for cargo scanning
+static const int CARGO_RADIUS_REAL_DELTA = 50;
 #define CARGO_REVEAL_MIN_DIST				150		// minimum distance for reveal cargo (used if radius+CARGO_RADIUS_DELTA < CARGO_REVEAL_MIN_DIST)
 #define CAP_CARGO_REVEAL_MIN_DIST		300		// minimum distance for reveal cargo (used if radius+CARGO_RADIUS_DELTA < CARGO_REVEAL_MIN_DIST)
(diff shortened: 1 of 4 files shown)


code/hud/hudtarget.h
```diff
@@ -26,6 +26,7 @@ struct weapon_info;
 #define MATCH_SPEED_THRESHOLD				0.1f		// minimum speed target must be moving for match speed to apply
 #define CARGO_RADIUS_DELTA					100		// distance added to radius required for cargo scanning
 #define CAPITAL_CARGO_RADIUS_DELTA		250		// distance added to radius required for cargo scanning
+static const int CARGO_RADIUS_REAL_DELTA = 50;
```
asarium: The `static` here is not necessary since `const` already implies it. Also, maybe a comment describing what this is useful for would be nice here.

code/ship/ship.cpp
```diff
@@ -1700,6 +1704,8 @@ ship_info::ship_info()
 	score = 0;
 
 	scan_time = 2000;
+	scan_range_normal = 150.0f;
+	scan_range_capital = 300.0f;
```
asarium: Is it correct that these values are the constants defined by `CARGO_REVEAL_MIN_DIST` and `CAP_CARGO_REVEAL_MIN_DIST`? If that is true, maybe it would be better to use those constants here instead to make sure that changing the constants actually has an effect.

niffiwan: Yes - I was originally intending to remove the \#define constants since they are no longer required. But I changed my mind after finding commented out code using the constants that I didn't want to change. I could go back & remove the constants as I originally intended, and move the comments with the original comments to this location.

Sorry, I'm not understanding your second sentence. I tested the changes on two missions in MVP 3.8.1 to ensure that the original values would correctly be used if nothing was defined for a ship class.

asarium: My second sentence was meant to show why the constants should be used (but that is pretty obvious). It doesn't have to do anything with actual gameplay changes. I checked the values as well so there is no problem here.

I'm fine with either solution (removing the constants or using the constants here). If they are only used in code that is commented out then I would suggest simply removing the constants.

niffiwan: actually - not important, I'll make the suggested changes & push shortly

Goober5000:
Looks good to me too.

PR #1844, opened 2018-08-29
HDR color space conversion for spec maps is not applied correctly

### LINKED ISSUE #1843: Specular maps aren't being applied correctly.
There seems to be an issue with how spec maps are applied. With certain assets there seems to be some crossover with other maps or some problem which causes very ugly purple and green artifacting.

This is especially noticeable when directly facing lights because of how gloss is applied.

Some examples with MediaVPs assets:

https://cdn.discordapp.com/attachments/480764424035106826/484436541205250059/screen0171.png

high -spec_static produces very obvious artifacting

https://cdn.discordapp.com/attachments/480764424035106826/484437257781248002/screen0006.png

The Arcadia looks discoloured with the spec map applied compared to no spec. You can also see the artifacting even though it's not as exaggerated.

https://cdn.discordapp.com/attachments/480764424035106826/484437303033462784/screen0007.png

The-E:
HDR color space conversion for spec maps only happened in one code path, not all of them. This fixes #1843

### DIFF
--- a/code/def_files/data/effects/main-f.sdr
+++ b/code/def_files/data/effects/main-f.sdr
@@ -280,11 +280,11 @@ void main()
 	}
 	if(gammaSpec) {
 		specColor.rgb = max(specColor.rgb, vec3(0.03f)); // hardcoded minimum specular value. read John Hable's blog post titled 'Everything Is Shiny'. 
- #ifdef FLAG_HDR
-		specColor.rgb = srgb_to_linear(specColor.rgb);
- #endif
 		fresnelFactor = 1.0;
 	}
+	#ifdef FLAG_HDR
+		specColor.rgb = srgb_to_linear(specColor.rgb);
+	#endif
 #endif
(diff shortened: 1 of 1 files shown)


code/def_files/data/effects/main-f.sdr
```diff
@@ -280,11 +280,11 @@ void main()
 	}
 	if(gammaSpec) {
 		specColor.rgb = max(specColor.rgb, vec3(0.03f)); // hardcoded minimum specular value. read John Hable's blog post titled 'Everything Is Shiny'. 
- #ifdef FLAG_HDR
-		specColor.rgb = srgb_to_linear(specColor.rgb);
- #endif
 		fresnelFactor = 1.0;
 	}
+	#ifdef FLAG_HDR
+		specColor.rgb = srgb_to_linear(specColor.rgb);
```
asarium: The indentation is not correct here. It should be one level below since this code is not in the if body anymore.

PR #1850, opened 2018-09-04
Handle DIRECTIVE_WING_ZERO in is-destroyed SEXPs correctly

### LINKED ISSUE #1722: Directive counter occasionally goes negative
The counter next to a HUD directive sometimes goes into the negatives like -997 or -996.  According to @Flip-Sandwich, this happens when all ships in the first wave of a wing are destroyed.
directive

asarium:
The special value DIRECTIVE_WING_ZERO (defined to be -999) is used to
signify to the directives system that an is-destroyed SEXP was found
which specified an empty wing. However, if a second is-destroyed-delay
SEXP is present in the same directive then the count of the parameters
is added to -999 which makes means that the directives system then
displays it on the HUD as a negative number.

This fixes that by resetting the counter to 0 if a second SEXP
encounters the special value which fixes the issue in the supplied test
mission. There are probably other edge cases when using multiple
is-destroyed SEXPs in a directive so I added a warning to the SEXP help
which lets the mission designer know that it is preferable to specify
multiple parameters instead of using multiple is-destroyed SEXPs.

This fixes #1722.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -5010,6 +5010,20 @@ int sexp_is_destroyed(int n, fix *latest_time)
 			if ( latest_time && (time > *latest_time) )
 				*latest_time = time;
 		} else {
+			// If a previous SEXP already had an empty wing then this code would expose the internal value as the
+			// directive count. Instead, we reset the count to zero here to make sure that the wing or ship count is
+			// correct.
+			if (Directive_count == DIRECTIVE_WING_ZERO) {
+#ifndef NDEBUG
+				static bool wing_zero_warning_shown = false;
+				if (!wing_zero_warning_shown) {
+					mprintf(("SEXP: is-destroyed-delay was used multiple times in a directive event! This might have "
+					         "unintended effects and should be replaced by a single use of is-destroyed-delay.\n"));
+					wing_zero_warning_shown = true;
+				}
+#endif
+				Directive_count = 0;
+			}
 			// ship or wing isn't destroyed -- add to directive count
 			if ( (wing_index = wing_name_lookup( name, 1 )) >= 0 ) {
 				Directive_count += Wings[wing_index].current_count;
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -5010,6 +5010,12 @@ int sexp_is_destroyed(int n, fix *latest_time)
 			if ( latest_time && (time > *latest_time) )
 				*latest_time = time;
 		} else {
+			// If a previous SEXP already had an empty wing then this code would expose the internal value as the
+			// directive count. Instead, we reset the count to zero here to make sure that the wing or ship count is
+			// correct.
+			if (Directive_count == DIRECTIVE_WING_ZERO) {
```
Goober5000: If this problem really warrants a warning in the SEXP help, then should there be a debug log statement here too?

code/parse/sexp.cpp
```diff
@@ -30858,7 +30865,9 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"\t1:\tName of the event in the mission."},
 
 	{ OP_IS_DESTROYED_DELAY, "Is destroyed delay (Boolean operator)\r\n"
-		"\tBecomes true <delay> seconds after all specified ships have been destroyed.\r\n\r\n"
+		"\tBecomes true <delay> seconds after all specified ships have been destroyed.\r\n"
+		"\tWARNING: If this is used multiple times in a directive SEXP unexpected results may be observed. "
+		"Specify multiple parameters instead.\r\n"
```
Goober5000: This is a little confusing at first readthrough.  I suggest "WARNING: If multiple is-destroyed-delay SEXPs are used in a directive event, unexpected results may be observed.  Instead, use a single is-destroyed-delay SEXP with multiple parameters."

asarium: Thanks, I changed it to your text.

code/parse/sexp.cpp
Goober5000: Oh right, good thinking.  Yes, we use boolean-guarded log prints elsewhere, so let's use one here too.

(How about resetting the boolean each time a mission is restarted, perhaps in the sexp initialization?)

asarium: I added the debug log print with the static boolean guard. Resetting the value when a mission is loaded would require some ugly extern references to such a variable which I would like to avoid so I kept it simple.

Goober5000: Okay fair enough.

PR #1865, opened 2018-09-19
Remodularize fireball.tbl

Goober5000:
This is a minor overhaul of fireball.tbl, with improvements in both sexps and modular tables.

Modular tables now work as one would expect -- the default fireball.tbl defines values and -fbl.tbm can override them or add new ones.  Previously new fireballs could not be added, and the old way of reconciling the modular table was hard to understand and even harder to adapt for new options.

Each fireball entry can now have a unique ID, which is generated if not specified by the table.  This can be used by the `explosion-effect` and `warp-effect` SEXPs to specify the fireball, rather than the old way of specifying the index.  (The old way still works.)  The unique ID is also another way of overriding a tbl entry via tbm.

This PR also, incidentally, fixes a couple of minor UI problems when referencing tbl indexes for `OPF_GAME_SND` and `OPF_WEAPON_BANK_NUMBER`, since the new `OPF_FIREBALL` needs the same functionality.

### DIFF
--- a/code/fireball/fireballs.cpp
+++ b/code/fireball/fireballs.cpp
@@ -107,42 +107,45 @@ void fireball_play_warphole_close_sound(fireball *fb)
 	snd_play_3d(gamesnd_get_game_sound(sound_index), &fireball_objp->pos, &Eye_position, fireball_objp->radius); // play warp sound effect
 }
 
-static void fireball_generate_unique_id(char *unique_id, int idx)
+static void fireball_generate_unique_id(char *unique_id, int buffer_len, int fireball_index)
 {
-	Assert((idx >= 0) && (idx < MAX_FIREBALL_TYPES));
+	Assert((fireball_index >= 0) && (fireball_index < MAX_FIREBALL_TYPES));
 
-	switch (idx)
+	switch (fireball_index)
 	{
 		// use sensible names for the fireball.tbl default entries
 		case FIREBALL_EXPLOSION_MEDIUM:
-			strcpy(unique_id, "Medium Explosion");
+			strncpy(unique_id, "Medium Explosion", buffer_len);
 			break;
 
 		case FIREBALL_WARP:
-			strcpy(unique_id, "Warp Effect");
+			strncpy(unique_id, "Warp Effect", buffer_len);
 			break;
 
 		case FIREBALL_KNOSSOS:
-			strcpy(unique_id, "Knossos Effect");
+			strncpy(unique_id, "Knossos Effect", buffer_len);
 			break;
 
 		case FIREBALL_ASTEROID:
-			strcpy(unique_id, "Asteroid Explosion");
+			strncpy(unique_id, "Asteroid Explosion", buffer_len);
 			break;
 
 		case FIREBALL_EXPLOSION_LARGE1:
-			strcpy(unique_id, "Large Explosion 1");
+			strncpy(unique_id, "Large Explosion 1", buffer_len);
 			break;
 
 		case FIREBALL_EXPLOSION_LARGE2:
-			strcpy(unique_id, "Large Explosion 2");
+			strncpy(unique_id, "Large Explosion 2", buffer_len);
 			break;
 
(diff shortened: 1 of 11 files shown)


Goober5000:
What's different between qtfred and fred2 that an extern reference to `Fireball_info` will resolve in one but not the other?  Is code.lib not linked against qtfred, or something?

And that reminds me... in the fred2 `management.cpp`, I added a call to `fireball_parse_tbl` so that `Fireball_info` could be populated.  How do I do the same thing in qtfred?

code/fireball/fireballs.cpp
```diff
@@ -114,12 +107,50 @@ void fireball_play_warphole_close_sound(fireball *fb)
 	snd_play_3d(gamesnd_get_game_sound(sound_index), &fireball_objp->pos, &Eye_position, fireball_objp->radius); // play warp sound effect
 }
 
+static void fireball_generate_unique_id(char *unique_id, int idx)
```
asarium: This function is unsafe and could cause a buffer-overflow. This should either have a parameter specifying the length of the `char` array or it should use a `SCP_string`.

code/fireball/fireballs.cpp
```diff
@@ -153,71 +184,128 @@ static void fireball_set_default_color(int idx)
...
+			{
+				fi = &Fireball_info[existing_idx];
+			}
+			// we are creating a new entry, so set some defaults
+			else
+			{
+				// make sure we don't exceed the max
+				if (Num_fireball_types >= MAX_FIREBALL_TYPES)
+				{
+					Warning(LOCATION, "Too many fireball entries!  Max is %d", MAX_FIREBALL_TYPES);
```
asarium: This should use `error_display`.

fred2/management.cpp
```diff
@@ -417,6 +417,11 @@ bool fred_init(std::unique_ptr<os::GraphicsOperations>&& graphicsOps)
 	parse_init();
 	techroom_intel_init();
 
+	// get fireball IDs for sexpression usage
+	// (we don't need to init the entire system via fireball_init, we just need the information)
+	extern void fireball_parse_tbl();
```
asarium: Maybe this function could be added to the fireball header to avoid the extern reference here?

asarium:
qtFRED has pretty much exactly the same function as FRED2. It is located in `qtfred/src/mission/management.cpp` and is called `initialize`.

Goober5000:
> 
> 
> qtFRED has pretty much exactly the same function as FRED2. It is located in `qtfred/src/mission/management.cpp` and is called `initialize`.

That much I figured out.  But I'm not familiar with this `listener(Subsystem::...)` procedure.  I'll add it to the next commit and see what you think.

asarium:
The listener is only relevant for the bigger subsystems. The smaller subsystems don't take long enough for the user to see the notification so it doesn't matter if the listener is called with that subsystem or not.

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -1132,27 +1132,66 @@ int sexp_tree::get_default_value(sexp_list_item* item, char* text_buf, int op, i
...
+				fireball_index = FIREBALL_MEDIUM_EXPLOSION;
+			}
+			else if (Operators[op].value == OP_WARP_EFFECT)
+			{
+				fireball_index = FIREBALL_WARP;
+			}
+
+			if (fireball_index >= 0)
+			{
+				extern fireball_info Fireball_info[MAX_FIREBALL_TYPES];
```
asarium: This here causes the linker issue. Since you declare the extern reference inside this function which belongs to a namespace the compiler thinks that the extern reference also belongs in that namespace.

Instead, you should just declare the array in the header. That will solve the linking issue.

Goober5000: Okay, that makes sense.  I was trying to hide the array from places where it didn't need to be visible, but putting it in the header is both easier and what the project usually does.

PR #1866, opened 2018-09-20
Custom warp glows, balls, and models

Goober5000:
This allows different fireball.tbl entries to have different auxiliary warp effects.  There is actually a warpglow02 bitmap specified in the retail VPs, so the Knossos entry can now be edited to use that bitmap if desired.  (This is not done in the new code for compatibility reasons; the MVPs currently do not specify a replacement warpglow02.)  In the same way, the Knossos entry (or any new entries) can have different "warp balls" and warp models.

This also fixes a small bug in #1865, where `$Unique ID:` was inadvertently not allowed because it is parsed before `$Name:`.

### DIFF
--- a/code/fireball/fireballs.cpp
+++ b/code/fireball/fireballs.cpp
@@ -416,7 +416,7 @@ void fireball_load_data()
 
 		if (strlen(fd->warp_model) > 0 && cf_exists_full(fd->warp_model, CF_TYPE_MODELS)) {
 			mprintf(("Loading warp model '%s'\n", fd->warp_model));
-			fd->warp_model_id = model_load(fd->warp_model, 0, NULL, 0);
+			fd->warp_model_id = model_load(fd->warp_model, 0, nullptr, 0);
 		} else {
 			fd->warp_model_id = -1;
 		}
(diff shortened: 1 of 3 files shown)


code/fireball/fireballs.cpp
```diff
@@ -999,6 +1031,8 @@ float fireball_wormhole_intensity(fireball *fb)
 	return rad;
 } 
 
+extern void warpin_queue_render(model_draw_list *scene, object *obj, matrix *orient, vec3d *pos, int texture_bitmap_num, float radius, float life_percent, float max_radius, bool warp_3d, int warp_glow_bitmap, int warp_ball_bitmap, int warp_model_id);
```
asarium: Why did you declare this function here and not in the header?

Goober5000: It was previously in the header; I removed the header declaration and added this.  I was thinking that header files are ideally suited for common functions used in many places, and specifically intended as interfaces.  Since this function was only used in warpineffect.cpp in one location, there was no harm in the extern.

And yes, this isn't consistent across the project, but it was the same thinking as extern'ing `fireball_parse_tbl` in the other PR.

asarium: I'm generally not in favor for `extern` declarations since they are prone for errors and they make the code less modular but it doesn't really matter here.

PR #1868, opened 2018-09-20
Allow longer anim filenames with APNG and ANI

TRBlount:
Previously, all files were subject to the same limit of 22 chars, excluding the file type+delimiter, null terminator, and frame number (_0000). Since ANIs and APNGs don't have a frame number associated with them as they are technically a single file, the check needs to be modified to allow them to have 27 or more chars excluding the filetype, delimiter, and null terminator.

### DIFF
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -1560,15 +1560,22 @@ int bm_load_animation(const char *real_filename, int *nframes, int *fps, int *ke
 	// EFF/APNG need 5 extra characters for each frame filename too, which just happens to be the same length as the frame designator needed otherwise
 	// MAX_FILENAME_LEN-10 == 5 character frame designator plus '.' plus 3 letter ext plus NULL terminator
 	// we only check for -5 here since the filename should already have the extension on it, and it must have passed the previous check
-	if (strlen(filename) > MAX_FILENAME_LEN - 5) {
+	/*if (strlen(filename) > MAX_FILENAME_LEN - 5) {
 		Warning(LOCATION, "Passed filename, '%s', is too long to support an extension and frames!!\n\nMaximum length for an ANI/EFF/APNG, minus the extension, is %i characters.\n", filename, MAX_FILENAME_LEN - 10);
 		if (img_cfp != nullptr)
 			cfclose(img_cfp);
 		return -1;
-	}
+	}*/
 
 	// it's an effect file, any readable image type with eff being txt
 	if (type == BM_TYPE_EFF) {
+		// Move filename length check to this section to account for differences between EFF, ANI, and APNG
+		if (strlen(filename) > MAX_FILENAME_LEN - 5) {
+			Warning(LOCATION, "Passed filename, '%s', is too long to support an extension and frames!!\n\nMaximum length for an EFF, minus the extension, is %i characters.\n", filename, MAX_FILENAME_LEN - 10);
+			if (img_cfp != nullptr)
+				cfclose(img_cfp);
+			return -1;
+		}
 		if (!bm_load_and_parse_eff(filename, dir_type, &anim_frames, &anim_fps, &key, &eff_type)) {
 			mprintf(("BMPMAN: Error reading EFF\n"));
 			if (img_cfp != nullptr)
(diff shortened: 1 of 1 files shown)


chief1983:
Since it should theoretically be possible to replace any animation type with the same one of another type using the same base filename, shouldn't the length restriction be the same across the board?

TRBlount:
No actually. EFFs are required to have frame number identifiers in the filename, whereas APNG and ANI are singular files. The main issue here is that APNG can be streamed, but in some instances all frames are loaded (like EFF), which is inconsistent behavior. The only file type that is pretty much always(?) streamed is ANI. This is not as trivial as you or I could hope.

chief1983:
What I mean is, I want to avoid a situation where someone puts in an asset using an ANI or an APNG (do apngs have a 4 letter extension?  Because  the comments around this part of the code say it assumed a 3 letter extension previously) and uses a longer filename as would be allowed by this change.  This is done in a place that an EFF could have also been used but they didn't use one (those do exist right?).  Then later, someone wants to replace the APNG/ANI with an EFF simply by putting an asset with the same filename in the filesystem or a higher priority VP file.  Except with this change, that would now be impossible because the EFF would have a filename that is too long.  Am I misunderstanding something here?

Also, the comments in the code mention APNGs having frame numbers in the filenames, I didn't think that was the case either.  Maybe that comment could be updated too?

TRBlount:
APNGs use the .png extension actually. The original plan was to use .apng as the extension but the code would need to be changed a lot to allow that. Spoon was complaining about this earlier because he thought his filenames for messages.tbl were within the char limit, but the EFF case is being applied to ANIs and APNGs when they don't really need to be. Another thing I saw is that AVIs added via messages.tbl are loaded completely into memory, but when it comes time to play the message, it looks like they are streamed, meaning memory is being wasted (at least in the case of ANIs and APNGs). Can you confirm that? Unless I'm misunderstanding the purpose of streaming anims.

chief1983:
But someone may still want to replace an ANI with an EFF, right?  I don't know enough to confirm whether AVIs are streamed or not.  Mostly just concerned with the fact that as long as we support using EFFs in the same place where an ANI or APNG could have been used, the possibility exists that someone might want to drop an EFF in place instead.  As long as we support EFFs it shouldn't be possible to break the engine by trying to load an EFF with a filename that is too long, or not being able to use an EFF because a modder used a filename that was too long for an EFF to be used.

Just curious, is there anywhere an APNG can be used currently, where an EFF would be more performant or more advantageous in any way?

I feel like you think this shouldn't be an issue as long as it's not plausible someone would ever want to do something, but I feel that if it's possible to do it, then we should not break it.  This phrase here, "the EFF case is being applied to ANIs and APNGs when they don't really need to be", as I understand it the entire point of applying the check to all of them is to ensure that any type could always be swapped out for any other type without needing a filename change, regardless of how likely it is that someone would do that swap.  Because unless I missed something this is a code path that is common to loading these file types no matter what situation they're used in, right?  Or is this just for HUD animations that no one has ever used EFF for or ever would?

Goober5000:
Bump...

ngld:
The PR still has the same buffer overflow issues as the last time this was discussed.
The passed filename is checked against `MAX_FILENAME_LEN` but [line 1769](https://github.com/TRBlount/fs2open.github.com/blob/fix/APNG_fname_len/code/bmpman/bmpman.cpp#L1769) puts `%s_%04d` in `entry->filename` which has a [size of `MAX_FILENAME_LEN`](https://github.com/TRBlount/fs2open.github.com/blob/fix/APNG_fname_len/code/bmpman/bm_internal.h#L67)

You need to increase the size of the `filename` field to at least `MAX_FILENAME_LEN + 6` (underscore, 4 digits, underscore and terminating zero).

Also, yes, line 1769 is run for APNGs since [`anim_frames` is set to the number frames in the apng](https://github.com/TRBlount/fs2open.github.com/blob/fix/APNG_fname_len/code/bmpman/bmpman.cpp#L1651).

asarium:
I don't think creating a V2 of the VP format is a worthwhile undertaking since it would require changing the existing VP tools for even getting support for the new format. Instead, we could consider adding support for simple zip files which can also contain uncompressed data that can then be streamed directly from the disk without any compression overhead.

That would also improve the tool support since those "VP" files could be opened and created using standard tools such as 7-zip.

Goober5000:
We shouldn't increase `MAX_FILENAME_LEN`.  And I agree that creating V2 of the VP format isn't practical.

It sounds like this is too complex an issue to be fixed without making the code a lot more convoluted in way or another.  Closing the PR.

PR #1872, opened 2018-09-23
Disable libRocket unless unicode mode is active

asarium:
Since libRocket requires unicode mode this doesn't change anything but
it fixes an error when a mod uses TrueType fonts without unicode mode.

### DIFF
--- a/code/scpui/rocket_ui.cpp
+++ b/code/scpui/rocket_ui.cpp
@@ -12,6 +12,7 @@
 
 #include "scpui/rocket_ui.h"
 #include "cfile/cfile.h"
+#include "mod_table/mod_table.h"
 #include "osapi/osapi.h"
 #include "scpui/RocketFileInterface.h"
 #include "scpui/RocketRenderingInterface.h"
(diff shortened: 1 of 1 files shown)


code/scpui/rocket_ui.cpp
```diff
@@ -543,13 +551,25 @@ void initialize()
 	os::events::addEventListener(SDL_KEYDOWN, os::events::DEFAULT_LISTENER_WEIGHT - 10, key_event_handler);
 
 	os::events::addEventListener(SDL_TEXTINPUT, os::events::DEFAULT_LISTENER_WEIGHT - 10, text_input_handler);
+
+	rocket_initialized = true;
 }
 void setOffset(float x, float y) { rendering_interface->setRenderOffset({x, y}); }
 
-void shutdown_scripting() { Rocket::Core::Lua::Interpreter::Shutdown(false); }
+void shutdown_scripting()
```
Goober5000: Should `rocket_initialized = false;` be added in this function as well?

asarium: No since that only disables the Lua part of libRocket and if `rocket_initialized` is set to `false` here then the actual `shutdown` call will return early without actually shutting down libRocket.

Goober5000: Gotcha.

Issue #1536, opened 2017-12-18
play-dead AI goal's behavior is different than expected


Baezon:
So currently, some AI goals says they only go up to 89 (and can always be overwritten by the player) and others say they go up to 200 (and cannot be overwritten by the player if it's 100+), and so far as I've been able to test, this is accurate. However, there's no real reason for certain goals to be always overwritable, and moreover, this doesn't seem to have always been the case as some campaigns rely on play-dead being non-overwritable, despite it being, at least now, always overwritable.

Flip-Sandwich:
Are there goals that are always overwritable regardless of priority? Since I think all AI goals can go up to 200 unless there's a few that have dumb hard caps at 89 so they can't which ones are those I'll take a look

Baezon:
Play-dead, off the top of my head, there was at least one other, though, I can't quite check right now. Based on my testing all the ones whose sexp blurbs said they could only go up to 89, were those always overwritable ones.

Flip-Sandwich:
garble garble this doesn't make sense to me, nothing in the code i can see right now has any different priorities for play-dead and any other order i'll keep looking

Baezon:
Can you confirm through testing that is indeed always overwritable, at least for you? I've not actually seen any confirmation from others about this, and I'm certainly not willing to rule out it's something specific to my system, but I've been able to reproduce it in a variety of contexts.

Flip-Sandwich:
Yeah play-dead is always overwritable even if you set it to 200 via initial orders or 200 via SEXP and I have no idea what's causing it

Baezon:
Here's a test mission for anyone interested. He gets a play-dead order with a priority of 200 one second in, but you can still order him around.
[test.txt](https://github.com/scp-fs2open/fs2open.github.com/files/2131075/test.txt)

MageKing17:
Okay, so, I ran the test mission through a debugger and finally figured out what's going on... the ship loses the `play-dead` goal the frame after it's given it. Why? Because apparently three goal types call `ai_clear_ship_goals( aip );` (for reasons I'm not really clear on): `AI_GOAL_STAY_STILL`, `AI_GOAL_PLAY_DEAD`, and `AI_GOAL_FORM_ON_WING`. That means they wipe *themselves* from the goal list, so any other goal at any priority will override them. I'm not sure what the consequences of changing that behavior would be.

Goober5000:
Hmm.  Well, I have two ideas on how to address that:

[...]

PR #1880, opened 2018-09-27
add ai-play-dead-persistent

Goober5000:
This adds a new AI goal which is equivalent to play-dead except that it "sticks".  It doesn't clear the goal list
so it will continue playing dead until the goal is removed or another goal with a higher priority is added.

This addresses issue #1536.

I've tested this with the squad orders, and it stays persistent as expected.  HOWEVER, the form-on-wing order overrides even a play-dead order with a higher priority.  This is because the game specifically sets a flag so that form-on-wing will override any other goal.  ~~Still thinking on how to resolve this.~~  See below.


Goober5000:
I'm currently thinking I should modify `ai_goal_priority_compare` so that the override bit does not override a play-dead goal.

Goober5000:
Thanks.  I tested it pretty thoroughly; one of the issues the testing uncovered was the problem with the "override" flag.  Currently it works to my satisfaction.  We'll see what the requestees think.

PR #1881, opened 2018-09-27
Add proper class support to Lua API

asarium:
The current situation is that the scripting API depends on a lot of undefined behavior for making it work. For example, the C++ objects were moved into Lua objects by using `memcpy` which caused issues if the C++ object was not trivially copyable.

The reason for this was that the `ade_set_args`/`ade_get_args` functions did not have access to the C++ type information since it used the C-style varargs for passing a variable number of parameters to Lua. Since C++11 has "proper" support for variadic arguments with full type information, that can be used for improving the C++/Lua API so that non-trivial objects can also be used.

The way this works is that when a C++ object is passed to Lua via `ade_obj.Set` it is copied or moved into an internal variable. The returned object is typed so when it is passed to `ade_set_args`, the Lua API still has access to the type information of the contained object. Then `lua_newuserdata` is used for allocating raw storage for the object. This storage is used for constructing a new instance of the underlying C++ object by using placement-new. Again, the object previously passed to the Lua argument API is moved or copied into the memory provided by Lua at which point the C++ object is entirely inside the Lua state without the usage of undefined behavior. This also allows to use move-only types with the Lua API.

Getting arguments is much easier than setting them so I won't go into detail how that works here.

### DIFF
--- a/code/scripting/api/objs/model.cpp
+++ b/code/scripting/api/objs/model.cpp
@@ -277,16 +277,16 @@ ADE_FUNC(__len, l_ModelTextures, NULL, "Number of textures on model", "number", 
 ADE_INDEXER(l_ModelTextures, "texture", "number Index/string TextureName", "texture", "Model textures, or invalid modeltextures handle if model handle is invalid")
 {
 	modeltextures_h *mth = NULL;
-	int new_tex = -1;
+	texture_h* new_tex   = nullptr;
 	const char* s        = nullptr;
 
-	if (!ade_get_args(L, "os|o", l_ModelTextures.GetPtr(&mth), &s, l_Texture.Get(&new_tex)))
-		return ade_set_error(L, "o", l_Texture.Set(-1));
+	if (!ade_get_args(L, "os|o", l_ModelTextures.GetPtr(&mth), &s, l_Texture.GetPtr(&new_tex)))
+		return ade_set_error(L, "o", l_Texture.Set(texture_h()));
 
 	polymodel *pm = mth->Get();
 
 	if (!mth->IsValid() || s == NULL || pm == NULL)
-		return ade_set_error(L, "o", l_Texture.Set(-1));
+		return ade_set_error(L, "o", l_Texture.Set(texture_h()));
 
 	texture_info *tinfo = NULL;
 	texture_map *tmap = NULL;
@@ -297,7 +297,7 @@ ADE_INDEXER(l_ModelTextures, "texture", "number Index/string TextureName", "text
 		int idx = atoi(s) - 1;	//Lua->FS2
 
 		if (idx < 0 || idx >= num_textures)
-			return ade_set_error(L, "o", l_Texture.Set(-1));
+			return ade_set_error(L, "o", l_Texture.Set(texture_h()));
 
 		tmap = &pm->maps[idx / TM_NUM_TYPES];
 		tinfo = &tmap->textures[idx % TM_NUM_TYPES];
@@ -316,13 +316,13 @@ ADE_INDEXER(l_ModelTextures, "texture", "number Index/string TextureName", "text
 	}
 
 	if(tinfo == NULL)
-		return ade_set_error(L, "o", l_Texture.Set(-1));
+		return ade_set_error(L, "o", l_Texture.Set(texture_h()));
 
 	if (ADE_SETTING_VAR) {
-		tinfo->SetTexture(new_tex);
+		tinfo->SetTexture(new_tex->handle);
 	}
 
-	return ade_set_args(L, "o", l_Texture.Set(tinfo->GetTexture()));
+	return ade_set_args(L, "o", l_Texture.Set(texture_h(tinfo->GetTexture())));
 }
 
 ADE_FUNC(isValid, l_ModelTextures, NULL, "Detects whether handle is valid", "boolean", "true if valid, false if handle is invalid, nil if a syntax/type error occurs")
--- a/code/scripting/api/objs/particle.cpp
+++ b/code/scripting/api/objs/particle.cpp
@@ -22,32 +22,13 @@ bool particle_h::isValid() {
 
 
 //**********HANDLE: Particle
-ADE_OBJ(l_Particle, particle_h*, "particle", "Handle to a particle");
-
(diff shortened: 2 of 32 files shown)


code/scripting/api/objs/model.cpp
```diff
@@ -277,16 +277,16 @@ ADE_FUNC(__len, l_ModelTextures, NULL, "Number of textures on model", "number",
 ADE_INDEXER(l_ModelTextures, "texture", "number Index/string TextureName", "texture", "Model textures, or invalid modeltextures handle if model handle is invalid")
 {
 	modeltextures_h *mth = NULL;
-	int new_tex = -1;
+	texture_h* new_tex   = nullptr;
```
niffiwan: @asarium Could you check this one, `new_tex` is optional so it looks that this could trigger access of a nullptr [below]?(https://github.com/scp-fs2open/fs2open.github.com/pull/1881/files#diff-1869a2e128315108bfa3546b77c94486R322)

code/scripting/api/objs/particle.cpp
```diff
@@ -196,7 +177,7 @@ ADE_VIRTVAR(AttachedObject, l_Particle, "object", "The object this particle is a
 {
 	particle_h *ph = NULL;
 	object_h *newObj = nullptr;
```
niffiwan: Ditto here, with access [here](https://github.com/scp-fs2open/fs2open.github.com/pull/1881/files#diff-90bae49ed40ec7081a0c26a84cfb74b1R191); although this wasn't changed so I might be wrong?

code/scripting/api/objs/ship.cpp
```diff
@@ -56,12 +56,12 @@ ADE_INDEXER(l_ShipTextures, "number Index/string TextureFilename", "Array of shi
 {
 	object_h *oh;
 	const char* s;
-	int tdx=-1;
-	if (!ade_get_args(L, "os|o", l_ShipTextures.GetPtr(&oh), &s, l_Texture.Get(&tdx)))
-		return ade_set_error(L, "o", l_Texture.Set(-1));
+	texture_h* tdx = nullptr;
```
niffiwan: Maybe another one (leads to [here](https://github.com/scp-fs2open/fs2open.github.com/pull/1881/files#diff-cf534f763ab30f11e573c3e1c9daadeeR114)?)

code/scripting/api/objs/texturemap.cpp
```diff
@@ -69,55 +69,55 @@ ADE_OBJ(l_TextureMap, texture_map_h, "material", "Texture map, including diffuse
 ADE_VIRTVAR(BaseMap, l_TextureMap, "texture", "Base texture", "texture", "Base texture, or invalid texture handle if material handle is invalid")
 {
 	texture_map_h *tmh = NULL;
-	int new_tex = -1;
-	if(!ade_get_args(L, "o|o", l_TextureMap.GetPtr(&tmh), l_Texture.Get(&new_tex)))
-		return ade_set_error(L, "o", l_Texture.Set(-1));
+	texture_h* new_tex = nullptr;
```
niffiwan: Another one? https://github.com/scp-fs2open/fs2open.github.com/pull/1881/files#diff-abda744db0327a7ed4780cd91bf54953R80

code/scripting/api/objs/texturemap.cpp
```diff
@@ -69,55 +69,55 @@ ADE_OBJ(l_TextureMap, texture_map_h, "material", "Texture map, including diffuse
...
+	return ade_set_args(L, "o", l_Texture.Set(texture_h(tmap->textures[TM_BASE_TYPE].GetTexture())));
 }
 
 ADE_VIRTVAR(GlowMap, l_TextureMap, "texture", "Glow texture", "texture", "Glow texture, or invalid texture handle if material handle is invalid")
 {
 	texture_map_h *tmh = NULL;
-	int new_tex = -1;
-	if(!ade_get_args(L, "o|o", l_TextureMap.GetPtr(&tmh), l_Texture.Get(&new_tex)))
-		return ade_set_error(L, "o", l_Texture.Set(-1));
+	texture_h* new_tex = nullptr;
```
niffiwan: And again? https://github.com/scp-fs2open/fs2open.github.com/pull/1881/files#diff-abda744db0327a7ed4780cd91bf54953R80

code/scripting/api/objs/texturemap.cpp
```diff
@@ -69,55 +69,55 @@ ADE_OBJ(l_TextureMap, texture_map_h, "material", "Texture map, including diffuse
...
+	return ade_set_args(L, "o", l_Texture.Set(texture_h(tmap->textures[TM_GLOW_TYPE].GetTexture())));
 }
 
 ADE_VIRTVAR(SpecularMap, l_TextureMap, "texture", "Specular texture", "texture", "Texture handle, or invalid texture handle if material handle is invalid")
 {
 	texture_map_h *tmh = NULL;
-	int new_tex = -1;
-	if(!ade_get_args(L, "o|o", l_TextureMap.GetPtr(&tmh), l_Texture.Get(&new_tex)))
-		return ade_set_error(L, "o", l_Texture.Set(-1));
+	texture_h* new_tex = nullptr;
```
niffiwan: And again? 
https://github.com/scp-fs2open/fs2open.github.com/pull/1881/files#diff-abda744db0327a7ed4780cd91bf54953R116

code/scripting/api/objs/model.cpp
asarium: The usage shouldn't be reachable since it will only be called if that parameter is set but it's better to check that to avoid any problems.

I'll go through your comments and I'll submit a pull requests with the fixes.

code/scripting/api/objs/particle.cpp
asarium: Same as above.

code/scripting/api/objs/ship.cpp
asarium: Another case of setting a variable.

code/scripting/api/objs/model.cpp
asarium: I submitted the changes here: #1938.

PR #1884, opened 2018-09-29
Fix set-ets-values sexp documentation bug

niffiwan:
Found by Asteroth - thanks!


Goober5000:
That looks good.  I should have mentioned this before, but could you to the same thing to `get-ets-values`?

niffiwan:
lol & done :grin:

Goober5000:
Looks good. :D

PR #1885, opened 2018-09-30
Only play UI sounds if the event is triggered on the element where the property is set.

ngld:
Currently if an element consists of multiple parts, the hover sound will be played even while moving between these parts.
For example if you move between the text label and icon on the pilot select screen, you'll hear the hover sound even though you never leave the button.

This PR fixes this by only acting on events that are triggered on the element where the sound property is set and ignoring any other events that propagate from children.


asarium:
`test/libRocket` is not intended to be a development branch. Please base this on the current `master` version.

ngld:
Done.

ngld:
Sorry, completely forgot about this. Builds should be running now.

PR #1887, opened 2018-09-30
Make player warpout settings customizable

TRBlount:
This converts the player warpout defines into variables that can be set from the mod_settings.tbl.
This was done at the request of Spoon and Axem because the exponential thrust curve screws over a fair chunk of ships from WoD, making them take an egregious amount of time to warp out.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -15,6 +15,7 @@
 #include "mod_table/mod_table.h"
 #include "parse/parselo.h"
 #include "sound/sound.h"
+#include "playerman/player.h"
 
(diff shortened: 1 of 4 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -15,6 +15,7 @@
 #include "mod_table/mod_table.h"
 #include "parse/parselo.h"
 #include "sound/sound.h"
+#include "playerman\player.h"
```
asarium: This needs to be a `/`.

TRBlount: Fixed.

Issue #1889, opened 2018-10-05
Fix Missing Warpin Sounds for Hyperspace Effect using code from other WE effects


wookieejedi:
I found a bug that the hyperspace warpin effect does not play warpin sounds set in ships.tbl. So I looked at "int WE_Hyperspace::warpStart()" code line 4322 in shipfx.cpp, and I found that it is missing a section that actually plays any of the warpin sounds. The other WE (WarpEffects) have these lines specified below in their warpStart(), but not WE_Hyperspace.

wookieejedi:
I can even do this fix myself and attach a .dff file if that would be preferable.

PR #1890, opened 2018-10-06
Fix missing Warpin sounds from Hyperspace Warpin Effect 

wookieejedi:
I found a bug that the hyperspace warpin effect does not play warpin sounds set in ships.tbl. So I looked at "int WE_Hyperspace::warpStart()" code line 4322 in shipfx.cpp, and I found that it is missing a section that actually plays any of the warpin sounds. The other WE (WarpEffects) have these lines specified below in their warpStart(), but not WE_Hyperspace. The sections WE_Hyperspace::warpStart() are missing are added with this PR. This addresses Issue #1889.

### DIFF
--- a/code/ship/shipfx.cpp
+++ b/code/ship/shipfx.cpp
@@ -4317,6 +4317,11 @@ WE_Hyperspace::WE_Hyperspace(object *n_objp, int n_direction)
 	p_object *p_objp = mission_parse_get_parse_object(shipp->ship_name);
 	if (p_objp != NULL)
 		initial_velocity = (float) p_objp->initial_velocity * sip->max_speed / 100.0f;
+	
+	//*****Sound
+	snd_range_factor = 1.0f;
+	snd_start = snd_end = sound_handle::invalid();
+	snd_start_gs = snd_end_gs = NULL;	
 }
 
 int WE_Hyperspace::warpStart()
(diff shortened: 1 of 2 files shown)


wookieejedi:
Will fix issues that caused build failures.

asarium:
It appears like there are a few members missing in the `WE_Hyperspace` class. Also, instead of `NULL` you should be using `nullptr`. Travis CI will tell you the same thing but it's harder to understand those messages if you are not used to them.

wookieejedi:
> It appears like there are a few members missing in the `WE_Hyperspace` class. Also, instead of `NULL` you should be using `nullptr`. Travis CI will tell you the same thing but it's harder to understand those messages if you are not used to them.

Ah thanks, I will change 'NULL' to 'nullptr'. I copied the 'NULL' from the WarpEffect BSG section, so is it all right that the BG section uses 'NULL' instead of 'nullptr'? I will add the members to WE_Hyperspace, thanks for the clarification!

asarium:
The `nullptr` check only runs on changed code so if you had changed something in the BSG code then it would have shown the errors in those files as well.

The reason it's done like this is to make the transition to `nullptr` gradual since replacing `NULL` in the entire code base at once would cause a lot of merge issues. This way the change only needs to be made when the code is actually changed.

wookieejedi:
> The `nullptr` check only runs on changed code so if you had changed something in the BSG code then it would have shown the errors in those files as well.
> 
> The reason it's done like this is to make the transition to `nullptr` gradual since replacing `NULL` in the entire code base at once would cause a lot of merge issues. This way the change only needs to be made when the code is actually changed.

Ah I see, thank you for your quick explanation! So to add the missing members to WE_Hyperspace I would need to edit another file besides just shipfx.cpp?

asarium:
It looks like the class is defined in `shipfx.h`.

wookieejedi:
Sounds good, I will look there and make any necessary changes. Thanks again, I appreciate the help!

wookieejedi:
Okay I think that should do it.

PR #1891, opened 2018-10-06
Switch communication messages to use SCP_string

Artyrm:
That way we get rid of "Message too long" if we use Unicode. Also it is more healthy approach to use strings.

### DIFF
--- a/code/hud/hudsquadmsg.cpp
+++ b/code/hud/hudsquadmsg.cpp
@@ -105,7 +105,7 @@ int Menu_pgdn_coords[GR_NUM_RESOLUTIONS][2] = {
 typedef struct mmode_item {
 	int	instance;					// instance in Ships/Wings array of this menu item
 	int	active;						// active items are in bold text -- inactive items greyed out
-	char	text[NAME_LENGTH];		// text to display on the menu
+	SCP_string	text;		// text to display on the menu
 } mmode_item;
 
 #define MAX_MENU_ITEMS		50				// max number of items in the menu
@@ -154,7 +154,7 @@ int keys_used[] = {	KEY_1, KEY_2, KEY_3, KEY_4, KEY_5, KEY_6, KEY_7, KEY_8, KEY_
 #define TYPE_REPAIR_REARM_ABORT_ITEM	5
 
 
-char Comm_order_types[NUM_COMM_ORDER_TYPES][NAME_LENGTH];
+SCP_string  Comm_order_types[NUM_COMM_ORDER_TYPES];
 comm_order Comm_orders[NUM_COMM_ORDER_ITEMS];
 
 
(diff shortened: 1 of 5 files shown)


Artyrm:
tea says that Debug 64 build failed with:
make[2]: execvp: /bin/sh: Permission denied
make[2]: *** [lib/CMakeFiles/RocketControlsLua.dir/build.make:263: lib/CMakeFiles/RocketControlsLua.dir/libRocket/Source/Controls/Lua/ElementFormControlInput.cpp.obj] Error 127

But my changes has nothing to do with that?

asarium:
Tea CI is a bit unreliable some times. Unless it's a GCC warning or error you can ignore that.

Artyrm:
Travis got this error:
`../qtfred/src/ui/widgets/sexp_tree.cpp:4131:36: error: no matching function for call to ‘fso::fred::sexp_list_item::add_data(SCP_string&)’
   head.add_data(Comm_orders[i].name);
                                    ^
../qtfred/src/ui/widgets/sexp_tree.cpp:873:6: note: candidate: void fso::fred::sexp_list_item::add_data(const char*, int)
 void sexp_list_item::add_data(const char* str, int t) {
      ^
../qtfred/src/ui/widgets/sexp_tree.cpp:873:6: note:   no known conversion for argument 1 from ‘SCP_string {aka std::basic_string<char>}’ to ‘const char*’
ninja: build stopped: cannot make progress due to previous errors.`

I'll try to figure it out.

Goober5000:
@Artyrm can you provide more details about the "Message too long" error?  When and how does it appear?  Does it appear any time Unicode is used, or only certain languages in Unicode?

If the error appears when strings are translated, it may be necessary to change `temp_comm_order` to also use `SCP_string`.

Artyrm:
It appears then I use Russian strings for HUDSquadMsg.cpp section of strings.tbl, that seemingly not very long and definitely not more than 32 chars.



If you're about temp_comm_order_types it seems not necessary since it is not of constant size:
`	const char *temp_comm_order_types[] =
	{
		XSTR("Ships", 293),`

Goober5000:
Ah, you are right about `temp_comm_order_types`.

Okay this looks good.  Thanks for posting the error dialog message.

Artyrm:
Thank you for participation.

Goober5000:
Well, I was going to merge this, but AppVeyor failed.  @asarium or @chief1983, any ideas?

asarium:
This is the error that is causing the failures:
```
C:\projects\fs2open-github-com\fred2\ignoreordersdlg.cpp(119): error C2664: 'void CWnd::SetWindowTextA(LPCTSTR)': cannot convert argument 1 from 'SCP_string' to 'LPCTSTR' [C:\projects\fs2open-github-com\build\fred2
[...]
C:\projects\fs2open-github-com\fred2\sexp_tree.cpp(5665): error C2664: 'void sexp_list_item::add_data(const char *,int)': cannot convert argument 1 from 'SCP_string' to 'const char *' [C:\projects\fs2open-github-com\build
```

Artyrm:
Here what is strange. I have "FSO_BUILD_FRED2" set in CMake, but it is not created in my build path.

asarium:
You probably need to install MFC in Visual Studio since that is required for building FRED.

Artyrm:
I'll try that. But shouldn't I get warnings/errors like "FRED couldn't build"?

asarium:
No because if CMake doesn't find MFC it disabled FRED building automatically. You should see something like that in your CMake configure output.

Artyrm:
I got this:
`Building FRED2: ON
Building FSO tools: ON
Building qtFRED: OFF
Fatal warnings: OFF
Configuring done`

asarium:
It should be before that. `Building FRED2: ON` simply means that the option is enabled in CMake but not if the requirements for FRED are actually met. Please install MFC (it's in the Visual Studio installer) and if that doesn't fix it (might require a reboot and clearing the CMake cache) then post the configure output here.

Artyrm:
Built with MFC active, judging by those outputs:
```
G:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.11.25503\include\algorithm(911): warning C4244: '=': conversion from 'int' to 'char', possible loss of data (compiling source file G:\Games\Freespace\Source\fs2open.github.com.git\code\scripting\lua.cpp)
2>G:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.11.25503\include\algorithm(924): note: see reference to function temp
```
CMake also don't show that FRED is not building. But in fact it is not :(

Artyrm:
Ah, that's not MFC. I'll try to check somehow whether it is active. But if CMake shows all ok there should be build errors already in VS?

asarium:
CMake will **silently** disable FRED building if MFC is not found. If you don't know if MFC is installed or not then please post a screenshot of the C++ section of the Visual Studio installer here so I can see what you have installed.

Artyrm:
>silently disable FRED building if MFC is not found.
No, I saw this message in CMake output. 

I rebuild the solution, an finally I can see errors on FRED building!

PR #1892, opened 2018-10-06
Add new options system

asarium:
This is a system for storing arbitrary data in the config file and then
easily retrieving it later on. This has full Lua support for exposing
the available options in a custom UI and changing the values of the
options. This ported some of the existing command line options over to
the new system which is opt-in for the moment by using a mod_table flag.

This depends on #1881.


code/camera/camera.cpp
```diff
@@ -24,6 +25,23 @@ SCP_vector<camera*> Cameras;
 camid Current_camera;
 camid Main_camera;
 
+static SCP_string fov_display(float val)
+{
+	auto degrees = static_cast<float>(val * (180.0f / M_PI));
```
Goober5000: Isn't there a radians_to_degrees function somewhere in the codebase?

code/cmdline/cmdline.cpp
```diff
@@ -1685,18 +1676,30 @@ bool SetCmdlineParams()
...
 		Cmdline_window = 0; /* Make sure no-one sets both */
 	}
 
 	if(res_arg.found()){
 		Cmdline_res = res_arg.str();
+
+		int width = 0;
+		int height = 0;
+
+		if ( sscanf(Cmdline_res, "%dx%d", &width, &height) == 2 ) {
```
Goober5000: I suggest logging a warning if this fails to parse.

code/fireball/warpineffect.cpp
```diff
@@ -137,7 +137,7 @@ void warpin_queue_render(model_draw_list *scene, object *obj, matrix *orient, ve
 		vm_vec_scale_add2( &vecs[3], &orient->vec.rvec, -radius );
 		vm_vec_scale_add2( &vecs[3], &orient->vec.fvec, Grid_depth );
 
-		//	vm_vec_scale_add( &vecs[4], �er, &orient->vec.fvec, -Grid_depth );
+		//	vm_vec_scale_add( &vecs[4], �er, &orient->vec.fvec, -Grid_depth );
```
Goober5000: Some sort of character encoding problem here.

code/mod_table/mod_table.cpp
```diff
@@ -414,6 +415,16 @@ void parse_mod_table(const char *filename)
 			stuff_float(&Minimum_player_warpout_time);
 		}
 
+		if (optional_string("$Enable in-game options:")) {
+			stuff_boolean(&Using_in_game_options);
+
+			if (Enable_external_default_scripts) {
+				mprintf(("Game Settings Table: Using im-game options system.\n"));
```
Goober5000: Spelling error

code/mod_table/mod_table.cpp
```diff
@@ -414,6 +415,16 @@ void parse_mod_table(const char *filename)
...
+			} else {
+				mprintf(("Game Settings Table: Not using im-game options system.\n"));
```
Goober5000: Here too

code/options/Option.cpp
```diff
@@ -0,0 +1,138 @@
+//
+//
+
+#include "Option.h"
```
Goober5000: I recommend using lowercase file names.  That was one of the first changes made to the codebase back in 2002.

code/scripting/api/objs/enums.h
```diff
@@ -10,85 +10,88 @@
...
+const int32_t LE_VM_TOPDOWN                  = 67;
+const int32_t LE_VM_FREECAMERA               = 68;
+const int32_t LE_VM_CENTERING                = 76;
+const int32_t LE_MESSAGE_PRIORITY_LOW        = 71;
+const int32_t LE_MESSAGE_PRIORITY_NORMAL     = 72;
+const int32_t LE_MESSAGE_PRIORITY_HIGH       = 73;
+const int32_t LE_OPTION_TYPE_SELECTION       = 77;
+const int32_t LE_OPTION_TYPE_RANGE           = 78;
+
+const int ENUM_NEXT_INDEX = 79; // <<<<<<<<<<<<<<<<<<<<<<
```
Goober5000: This should be kept because new values are not necessarily added to the end of the list.

It would be nice to have some sort of compile-time check for duplicate values but I'm not sure how that could be done.

code/scripting/api/objs/player.cpp
```diff
@@ -266,16 +266,11 @@ ADE_FUNC(loadCampaignSavefile, l_Player, "string campaign = <current>", "Loads t
 scoring_stats_h::scoring_stats_h() = default;
 scoring_stats_h::scoring_stats_h(const scoring_struct& stats)
 {
-	_score = new scoring_struct();
```
Goober5000: I would prefer the scoring changes to be in a separate PR.  What is the rationale?

code/mod_table/mod_table.cpp
```diff
@@ -414,6 +415,16 @@ void parse_mod_table(const char *filename)
 			stuff_float(&Minimum_player_warpout_time);
 		}
 
+		if (optional_string("$Enable in-game options:")) {
+			stuff_boolean(&Using_in_game_options);
+
+			if (Enable_external_default_scripts) {
```
MageKing17: Copy-paste error, looks like.

code/scripting/api/objs/enums.h
asarium: I'm not sure what you are referring to here since nothing was removed here.
However, I just saw that I accidentally reused the value `77` so I'll need to change that.

code/options/Option.cpp
asarium: I named the file after the class that is defined within it (although `OptionBase` is also in there so I might move that to a separate file) and that starts with an upper case letter. I see no reason to use a different casing for the file and the class name.

code/scripting/api/objs/player.cpp
asarium: I think I introduced this suboptimal version in an earlier, related pull request and wanted to improve it here. Since it's already here and there is no real advantage in moving it to another PR I would like to keep this in here.

asarium:
@Goober5000 @MageKing17 I pushed some changes to solve the issues you found.

code/scripting/api/objs/enums.h
Goober5000: I must have misread the change; I thought `ENUM_NEXT_INDEX` was completely removed in the commit.  What you have looks good.

code/options/Option.cpp
Goober5000: It's to make life easier for the Linux folks.  All file names were changed from mixed-case to lowercase back in 2002, and they have been that way ever since.  See for example:
https://www.hard-light.net/forums/index.php?topic=8991.0
https://www.hard-light.net/forums/index.php?topic=23459.msg466635#msg466635

asarium: All the reasons that are described in those threads is about people accidentally checking in changes that break compilation on Linux because they can't test their changes on that platform.

We do not have that problem anymore since we have our CI setup since every pull request is validated before being merged.

Goober5000: All right, fair enough.

chief1983: Even if we can catch that issue now, after having gone through the trouble of standardizing the codebase on lowercase filenames, doesn't keeping that make more sense than just deciding we don't need to anymore?  Personally I hate mixed case projects, in terminal on macs and Linux they tend to show capital filenames first, then lowercase, because they use case sensitive sorting.  It causes me to sometimes take longer to find a file in the list than it should have, not to mention having to remember the case as well when typing out filenames, etc.  I feel like the work that was should not be abandoned just because one particular problem of mixed-case files may no longer be an issue.  Surely we have other classes where the class is capitalized and the containing file is not?

asarium: I'm sure the standardization was necessary when it was done due to the lacking tooling infrastructure but I do not think the same reasons still apply. Using lower case file names simply because that makes it harder to introduce compile breaks on Linux is not a good reason for using lower case names.

I prefer having file names that reflect the name of the class that is being defined in them which is not possible to do completely if file names always have to be lower case.

chief1983: I didn't mention anything about breaking compilation.

Goober5000:
I'm going to make an executive decision and merge this.  The letter-case dispute, while relevant, isn't worth holding up the feature.  It is easy enough to make another PR to change the letter case down the road if we have to.

Issue #1857, opened 2018-09-15
LUA Hook for 'Mission outcome accepted'


wookieejedi:
The game updates campaign persistent variables on the condition that "Mission outcome accepted". Unfortunately On "Mission outcome accepted" is not a game state that can be called with the lua API.

Thus, I am proposing adding a lua hook that runs when the player hits the "Accept" button after a mission.

PR #1893, opened 2018-10-07
Adds Accept Mission Outcome as new Conditional Hook

wookieejedi:
Adds Accept Mission Outcome Event to Game Events, so LUA scripts can be called when a campaign mission is accepted. This addresses Issue #1857.


wookieejedi:
Not sure why the tea build failed with '-- Check size of unsigned int - failed' on only the x64 debug version...

wookieejedi:
Ah you're right, it wasn't actually being called. So I should define a hook in scripting.h and then add it in scripting.cpp?

asarium:
Yes, you can use the existing hooks as a template to see how that is done.

wookieejedi:
Sounds good thanks. I figured I would want to use a hook template similar to CHA_MISSIONEND, but I can't find any reference to it other then in the initial Script_actions[] list, though I will keep looking. : )

asarium:
It's used in `game_level_close` in `freespace.cpp`.

wookieejedi:
I've added the hook, CHA_CMISSIONACCEPT to the appropriate lists in scripting.ccp and .h. Just had a quick question though, would it be more appropriate to use these lines at the end of the function game_level_close() in freespace.cpp?

"
	if ( Game_mode & GM_CAMPAIGN_MODE ){
		mission_num = Campaign.current_mission;
		if ( Campaign.next_mission != mission_num ) {
			Script_system.RunCondition(CHA_CMISSIONACCEPT);
			} 
	}
"

Or 

Should I simply add the line <Script_system.RunCondition(CHA_CMISSIONACCEPT);> in the appropriate place in missioncampaign.cpp and then just include the scripting(.ccp and .h) in top include list for missioncampaign.cpp?

wookieejedi:
Looks like that should do it, as long as that is an appropriate place to add the hook run condition.

PR #1894, opened 2018-10-07
More friendly assert if used font misses a glyph.

Artyrm:
In Debug mode one can have following non-informative message. It can happen if UTF-08 font misses some glyph that is used by the game. I added more informative message and a log print.

### DIFF
--- a/code/graphics/paths/NanoVGRenderer.cpp
+++ b/code/graphics/paths/NanoVGRenderer.cpp
@@ -21,6 +21,22 @@
 #include "NanoVGRenderer.h"
 #include "tracing/tracing.h"
 
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// That is a wrapper function for log prints to be availiable for nanovg components. For now it is stb_truetype.h
+// Planted by ksotar with blessing from asarium
+void nvgOldCPrintf(const char *message, ...) {
+	nprintf(("General", message));
+}
+
+#ifdef __cplusplus
+}
+#endif
+
+
 namespace {
 using namespace graphics::paths;
 
--- a/code/graphics/paths/nanovg/stb_truetype.h
+++ b/code/graphics/paths/nanovg/stb_truetype.h
@@ -1,11 +1,12 @@
-// stb_truetype.h - v1.11 - public domain
-// authored from 2009-2015 by Sean Barrett / RAD Game Tools
+// stb_truetype.h - v1.19 - public domain
+// authored from 2009-2016 by Sean Barrett / RAD Game Tools
 //
 //   This library processes TrueType files:
 //        parse files
 //        extract glyph metrics
 //        extract glyph shapes
 //        render glyphs to one-channel bitmaps with antialiasing (box filter)
+//        render glyphs to one-channel SDF bitmaps (signed-distance field/function)
 //
 //   Todo:
 //        non-MS cmaps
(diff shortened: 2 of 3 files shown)


Artyrm:
Woops, in Fast Debug it prints into log like this:
`WARNING: Probably font character is missing. Unicode decimal index is 41478.`
This code is not true.

Artyrm:
"More informative" assert is like this:

Artyrm:
Travis:

```
../code/graphics/paths/nanovg/stb_truetype.h: In function ‘stbtt__run_charstring’:
../code/graphics/paths/nanovg/stb_truetype.h:2121:7: error: this statement may fall through [-Werror=implicit-fallthrough=]
    if (!has_subrs) {
       ^
```

I believe it is due to the new version of stb_truetype.h

asarium:
It looks like integrating the new stb_truetype version into this isn't going as smoothly as I thought. Can you do your changes without the stb_truetype update? That should make it easier to merge this.

code/graphics/paths/NanoVGRenderer.cpp
```diff
@@ -21,6 +21,22 @@
 #include "NanoVGRenderer.h"
 #include "tracing/tracing.h"
 
+
+#ifdef __cplusplus
```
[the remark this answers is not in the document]
Artyrm: Totally agree.

Artyrm:
Well, I think this fallthrough comment was just illformed for gcc to account it, and now it should do.
If not, then I should do reverse-commit?

Artyrm:
Trying //-fallthrough that is suitable for -Wimplicit-fallthrough=4 according to this:
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough

If we use =5 no comments will help in this case.

Artyrm:
Looks like all is fine now

asarium:
I'll take a closer look when I have some free time (probably this weekend).

code/graphics/paths/NanoVGRenderer.cpp
```diff
@@ -21,6 +21,22 @@
...
+// That is a wrapper function for log prints to be availiable for nanovg components. For now it is stb_truetype.h
+// Planted by ksotar with blessing from asarium
+extern "C" {
+void nvgOldCPrintf(SCP_FORMAT_STRING const char *message, ...) {
+	SCP_string buf;
+	va_list args;
+	va_start(args, message);
+	vsprintf(buf, message, args);
+	va_end(args);
+	#ifdef outwnd_printf2
```
asarium: I'm assuming you are using this to check if we are in debug mode. In FSO this is usually done with `#ifndef NDEBUG`. Also, you can put the entire body of this function in that preprocessor block to save some memory in release mode.

code/graphics/paths/nanovg/stb_truetype.h
```diff
@@ -968,640 +1254,1433 @@ typedef int stbtt__test_oversample_pow2[(STBTT_MAX_OVERSAMPLE & (STBTT_MAX_OVERS
...
+			--entrySelector;
+		}
+		search += 2;
+
+		{
+			stbtt_uint16 offset, start;
+			stbtt_uint16 item = (stbtt_uint16)((search - endCount) >> 1);
+
+			if (unicode_codepoint > ttUSHORT(data + endCount + 2 * item)) {
+				nvgOldCPrintf("WARNING: Probably font character is missing. Unicode decimal index is %d.\n", unicode_codepoint);
```
asarium: The warning has a few grammar issues. It should probably be `WARNING: A font character is probably missing. Unicode decimal index is %d.`.

code/graphics/paths/nanovg/stb_truetype.h
```diff
@@ -968,640 +1254,1433 @@ typedef int stbtt__test_oversample_pow2[(STBTT_MAX_OVERSAMPLE & (STBTT_MAX_OVERS
...
+			}
+			STBTT_assert((unicode_codepoint <= ttUSHORT(data + endCount + 2 * item)) &&
+				"                                        \
+				Looks like the font file you're using misses a character. Look into log for 'WARNING: Probably font' to know its code.");
```
asarium: `... Search the log for 'WARNING: A font character is probably missing.' in order to find out the missing unicode codepoint.`

PR #1895, opened 2018-10-10
Barracks stats adaptation for longer Unicode strings.

Artyrm:
Originally there is many unused space for stats listing even in 1024 resolution.

This change widens used space, so more wide languages can fit there too. Also it uses 2x space to store those strings. Result is like that: 

Original also appears in more centrist view:

### DIFF
--- a/code/menuui/barracks.cpp
+++ b/code/menuui/barracks.cpp
@@ -37,7 +37,7 @@
 
 // stats defines
 //#define NUM_STAT_LINES (21 + MAX_SHIP_CLASSES)	// Goober5000
-#define STAT_COLUMN1_W 40
+#define STAT_COLUMN1_W 40*2							// as we might use Unicode //ksotar
 #define STAT_COLUMN2_W 10
 
 static int Stat_column1_w[GR_NUM_RESOLUTIONS] =
@@ -76,7 +76,7 @@ static int Barracks_stats_coords[GR_NUM_RESOLUTIONS][4] = {
 		32, 212, 240, 250
 	},
 	{ // GR_1024
-		42, 351, 240, 400
+		42, 351, 350, 400
 	}
 };
 
@@ -85,7 +85,7 @@ static int Barracks_stats2_coords[GR_NUM_RESOLUTIONS][3] = {
 		276, 212, 81		// X2, , W2
 	},
 	{ // GR_1024
-		286, 351, 81		// X2, , W2
+		396, 351, 81		// X2, , W2
 	}
 };
 


asarium:
I'm not sure if it's a good idea to change how the retail UI looks. Maybe you could change the UI so the center point is adjusted dynamically depending on the length of the longest string?

Artyrm:
Well, I could give it a try.

Artyrm:
Made it adaptable, and original stays as it was.

code/menuui/barracks.cpp
```diff
@@ -80,6 +81,8 @@ static int Barracks_stats_coords[GR_NUM_RESOLUTIONS][4] = {
 	}
 };
 
+uint16_t Calculated_width;		// For Unicode languages
```
asarium: 1. Please move this variable (and `Calculated_X2`) below the declaration of `Stats` since the variables are more related to those variables.
2. Please make them `static`.
3. The variable is also not very descriptive for what it does. Maybe `Stats_max_width` and `Stats_x2` would be better along with a clarifying comment for what they are used.

code/menuui/barracks.cpp
```diff
@@ -478,13 +485,28 @@ void barracks_init_stats(scoring_struct *stats)
...
 	sprintf(Stats[Num_stat_lines], "%d", score_from_kills);
 	Num_stat_lines++;
 
-	for (i=0; i<Num_stat_lines; i++) {
-		font::force_fit_string(Stat_labels[i], Stat_column1_w[gr_screen.res], Barracks_stats_coords[gr_screen.res][BARRACKS_W_COORD]);
-		font::force_fit_string(Stats[i], Stat_column2_w[gr_screen.res], Barracks_stats2_coords[gr_screen.res][BARRACKS_W_COORD]);
+	int w;
+	Calculated_width = 0;
+	Calculated_X2	 = 0;
+	if ((Unicode_text_mode) && (gr_screen.res != GR_640)) {
```
asarium: The extra parenthesis aren't needed and don't improve the readability of the code. Please remove them.

code/menuui/barracks.cpp
```diff
@@ -478,13 +485,28 @@ void barracks_init_stats(scoring_struct *stats)
...
+		for (i = 0; i < Num_stat_lines; i++) {					// Find the longest string in labels to base list appearance on that
+			gr_get_string_size(&w, nullptr, Stat_labels[i]);
+			if (Calculated_width < w) Calculated_width = (uint16_t) w;
```
asarium: Single line `if`s can be hard to read, please place the body in a separate line.

Issue #1867, opened 2018-09-20
Docking Bug with Warp-In


wookieejedi:
It appears that a docked ship appears both visually and on the radar about a second before the ship it is docked to appears, but this only happens if the large ship is using a warp-in effect. If the the box "No Warp Effect" is checked in FRED then both ships arrive exactly at the same time as expected. Thus, it seems the warp-in animation is delaying the actual arrival of the large ship, but not the smaller ship it is docked with. I have attached a sample mission which shows this. 

This lag effect is especially apparent on ships that have $Warpin Type: Hyperspace, because
 in that case the docked ship appears, disappears, then appears again as it warps in. 

[_Dock_Test.txt](https://github.com/scp-fs2open/fs2open.github.com/files/2402732/_Dock_Test.txt)

wookieejedi:
From section of shipfx_warpin_start() in shipfx.cpp: 

        // docked ships who are not dock leaders don't use the warp effect code
	// (the dock leader takes care of the whole group)
	if (object_is_docked(objp) && !(shipp->flags[Ship::Ship_Flags::Dock_leader]))
	{
		return;
	}

So it looks like the dock leader is not being set correctly, or the dock leader is not properly taking care of the whole group? Mainly documenting this so I remember.

Goober5000:
> So it looks like the dock leader is not being set correctly, or the dock leader is not properly taking care of the whole group?

The second of these.  If the dock leader wasn't being set correctly, a lot more things would be going wrong.

wookieejedi:
Ah good to know. Still strange that even with only one ship the bug still occurs...

Goober5000:
Can you test previous builds and see if this is a recent bug or has been around for a while?

wookieejedi:
Sure, will let you know what I find!

Goober5000:
By visually, do you mean you can see it, or just that you can target it?

I think this is just due to things being visible in stage 1 when they should only be visible in stage 2.

wookieejedi:
I can visually see it and target it. The ship/dockees will show up for about two seconds, then disappear as stage two warpin commences. I think you are right.

Goober5000:
Now that #1932 has been merged, can you test this on the next nightly build?

wookieejedi:
Indeed! I tested it on the most recent nightly (Dec 8th) and unfortunately the bug still exists, so it must be related to something else?

Goober5000:
It must be. :-/

[...]

PR #1896, opened 2018-10-10
Fixes missing "On Ship Arrive" with Docked Wings

wookieejedi:
Fixes bug by adding the conditional hook call "On Ship Arrive" to docked wings that arrive. Previously, ships in a docked wing that would arrive would not trigger the "On Ship Arrive" hook variable.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -2489,6 +2489,24 @@ void parse_bring_in_docked_wing(p_object *p_objp, int wingnum, int shipnum)
 	// make sure we haven't created too many ships
 	Assert(wingp->current_count <= MAX_SHIPS_PER_WING);
 
+	//fixes bug where docked wings are not called with the 'On Ship Arrive' hook variable --wookieejedi
+	int num_ships = wingp->current_count;
+	int anchor_objnum = -1;
+	anchor_objnum = mission_set_arrival_location(wingp->arrival_anchor, wingp->arrival_location, wingp->arrival_distance, OBJ_INDEX(leader_objp), wingp->arrival_path_mask, &pos, &orient);
+	if (num_ships > 0 && Game_mode & GM_IN_MISSION) {
+		for ( index = wingp->current_count - num_ships; index < wingp->current_count; index ++ ) {
+			object *objp = &Objects[Ships[wingp->ship_index[index]].objnum];
+
+			if (anchor_objnum >= 0)
+				Script_system.SetHookObjects(2, "Ship", objp, "Parent", &Objects[anchor_objnum]);
+			else
+				Script_system.SetHookObjects(2, "Ship", objp, "Parent", NULL);
+
+			Script_system.RunCondition(CHA_ONSHIPARRIVE, objp);
+			Script_system.RemHookVars(2, "Ship", "Parent");
+		}
+	}
+	
 	// at this point the wing has arrived, so handle the stuff for this particular ship
 
 	// set up wingman status index
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -1837,6 +1837,7 @@ int parse_create_object_sub(p_object *p_objp)
 {
 	int	i, j, k, objnum, shipnum;
 	int anchor_objnum = -1;
+	bool brought_in_docked_wing = false;
 	ai_info *aip;
 	ship_subsys *ptr;
 	ship *shipp;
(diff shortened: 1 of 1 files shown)


code/mission/missionparse.cpp
```diff
@@ -2489,6 +2489,24 @@ void parse_bring_in_docked_wing(p_object *p_objp, int wingnum, int shipnum)
 	// make sure we haven't created too many ships
 	Assert(wingp->current_count <= MAX_SHIPS_PER_WING);
 
+	//fixes bug where docked wings are not called with the 'On Ship Arrive' hook variable --wookieejedi
+	int num_ships = wingp->current_count;
+	int anchor_objnum = -1;
+	anchor_objnum = mission_set_arrival_location(wingp->arrival_anchor, wingp->arrival_location, wingp->arrival_distance, OBJ_INDEX(leader_objp), wingp->arrival_path_mask, &pos, &orient);
```
asarium: Does `mission_set_arrival_location` change something in the mission? I'm worried that this might cause changes in behaviour.

Also, you can initialize `anchor_objnum` directly from the return value of `mission_set_arrival_location` and you can move the initialization inside the `if` below.

wookieejedi: Ah you're right I think it does. I won't use that function then. In that function, anchor_objnum = Ships[shipnum].objnum, so perhaps I should use that instead of using the whole function.

Also, should I change NULL to nullptr in this case?

asarium: Yes, you always need to fix those things now.

code/mission/missionparse.cpp
```diff
@@ -2489,6 +2489,26 @@ void parse_bring_in_docked_wing(p_object *p_objp, int wingnum, int shipnum)
 	// make sure we haven't created too many ships
 	Assert(wingp->current_count <= MAX_SHIPS_PER_WING);
 
+	//fixes bug where docked wings are not called with the 'On Ship Arrive' hook variable --wookieejedi
```
Goober5000: I see several problems with this PR.

1) `num_ships` will always be >0 here.
2) A wing which arrives because all its ships are docked will have no arrival anchor, so the check is unnecessary.
3) Your code loops through all the ships in the wing, which is wrong here because this function is already called for *each* ship in the wing, *as* it is created.
4) This function is the wrong place for the scripting hook -- normally, the hook is executed at the very bottom of `parse_create_object_sub`.  This is because the FREDder may need to access wing attributes that are assigned later in the function.  Instead of adding a hook to this function, you need to move this code to the same place as the other hook.

wookieejedi: Perhaps something like this at the end of `parse_create_object_sub`?

```
	// The ship is in a wing, but is docked. Adding this block ensures that docked ships actually execute "On Ship Arrive". --wookieejedi
	if (Game_mode & GM_IN_MISSION && shipp->wingnum != -1 && object_is_docked(pobjp)) {
		Script_system.SetHookObjects(2, "Ship", &Objects[objnum], "Parent", nullptr);
		Script_system.RunCondition(CHA_ONSHIPARRIVE, &Objects[objnum]);
		Script_system.RemHookVars(2, "Ship", "Parent");
	}	
```

wookieejedi:
Hopefully that should take care of it. I’m wondering if that might be a potential spot to address the warpin bug with docked hyperspace ships from issue #1867.

code/mission/missionparse.cpp
Goober5000: It's not possible to add the hook to `mission_set_wing_arrival_location()` because of the inside-out way that docked wings are initialized, so the hook will need to be added to the bottom of `parse_create_object_sub()`.

Your proposed code is partially correct, but it doesn't handle the correct set of scenarios.  There are some very specific conditions for initializing a docked ship in this way, and they are all handled in the check near the beginning of the function.

In fact, now that I think about it, the best solution is probably to create a new boolean variable `brought_in_docked_wing` and add this to the if() block at the end of `parse_create_object_sub()`.

Goober5000:
Well, #1867 doesn't deal with docked wings, does it?  That would mean the root cause is something else.

wookieejedi:
True,  #1867 deals with all dock ships, not just wings. 

Also, I incorporated the revisions for adding a bool value, is that what you had in mind? Though I am not entirely sure why the Travis build failed due to `No output has been received in the last 10m0s` on only a single build job...

Goober5000:
Your latest changes look good.  The scripting hook should now work properly, but you should test it just to be sure.

One more change I recommend is to remove the `== true` check.  It's redundant because boolean variables can be tested directly.

Sometimes Travis can hang or stop responding.  I'm not sure how to restart it manually, but it will restart automatically if you push another commit to this PR.

wookieejedi:
Thanks for that tip. Just built/tested and it works as expected, now docked ships in a wing do register with `$On Ship Arrive`.

Goober5000:
Good work.  Approved, and I'll merge once the tests are complete.

I assume the scripting hook is in two different places because the hook needs to know the anchor object for the arrival.  For wings, this is set in `mission_set_wing_arrival_location`.

wookieejedi:
Great, thanks! Ah, that makes sense.

PR #1897, opened 2018-10-12
Fixes Bug with Docked Ships and Warpin

wookieejedi:
Fixes issue #1867. This ensures that a docked ship will not arrive instantly if the ship it is docked to is using a warpin effect. The parent conditions of this 'if' block accounts for both docked ships not in wings and docked ships in wings. I have test this with retail and it doesn't cause any issues.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -2458,7 +2458,8 @@ int parse_create_object_sub(p_object *p_objp)
 		if (object_is_docked(p_objp) && !(p_objp->flags[Mission::Parse_Object_Flags::SF_Dock_leader])) {
 			object* objp;
 			objp = p_objp->created_object;
-			shipfx_warpin_start(objp);		
+			shipfx_warpin_start(objp);	
+		}			
 	}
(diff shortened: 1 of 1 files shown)


wookieejedi:
Not sure why the tea build failed so quickly...? I just built and test on my end and it works.

Goober5000:
Uh, that is wrong.  Ships that are docked and not the dock leader are not supposed to do any arrival code on their own.  They are completely dependent on the dock leader.  The dock leader handles the arrival, the warp, and all necessary actions.

wookieejedi:
> Uh, that is wrong. Ships that are docked and not the dock leader are not supposed to do any arrival code on their own. They are completely dependent on the dock leader. The dock leader handles the arrival, the warp, and all necessary actions.

Ah, this fixed worked so well I thought this fix might be too easy... Guess I am back to finding where the dock leader is supposed to assign the warpin info for the dockees.

Goober5000:
Yep.  Sorry.  Good detective work though.  Keep looking!

PR #1898, opened 2018-10-13
If an icon can't be loaded, display an error instead of asserting.

ngld:
Also, if we can't open an existing file, we should at least display a warning.

### DIFF
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -1892,6 +1892,7 @@ int bm_load_sub_slow(const char *real_filename, const int num_ext, const char **
 	}
 
 	// umm, that's not good...
+	Warning(LOCATION, "Could not open file %s!", res.full_name.c_str());
 	return -1;
 }
 
--- a/code/mission/missionbriefcommon.cpp
+++ b/code/mission/missionbriefcommon.cpp
@@ -609,7 +609,8 @@ void brief_preload_icon_anim(brief_icon *bi)
 	// force read of data from disk, so we don't glitch on initial playback
 	if ( ga->first_frame == -1 ) {
 		ga->first_frame = bm_load_animation(ga->filename, &ga->num_frames);
-		Assert(ga->first_frame >= 0);
+		if ( ga->first_frame < 0 )
+			Warning(LOCATION, "Failed to load icon %s!", ga->filename);
 	}
 }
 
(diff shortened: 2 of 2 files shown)


code/bmpman/bmpman.cpp
```diff
@@ -1892,6 +1892,7 @@ int bm_load_sub_slow(const char *real_filename, const int num_ext, const char **
 	}
 
 	// umm, that's not good...
+	Warning(LOCATION, "Could not open file %s!", res.full_name.c_str());
```
Goober5000: Instead of displaying a Warning, I recommend outputting a message to the log.  I have a feeling that this would cause a lot of annoying Warnings for many mods.

code/mission/missionbriefcommon.cpp
```diff
@@ -609,7 +609,8 @@ void brief_preload_icon_anim(brief_icon *bi)
 	// force read of data from disk, so we don't glitch on initial playback
 	if ( ga->first_frame == -1 ) {
 		ga->first_frame = bm_load_animation(ga->filename, &ga->num_frames);
-		Assert(ga->first_frame >= 0);
+		if (ga->first_frame < 0)
+			Error(LOCATION, "Failed to load icon %s!", ga->filename);
```
Goober5000: Good idea, but maybe *this* would be appropriate as a Warning instead of an Error.  Errors generally mean that the error is unrecoverable.  I speculate that the briefing animation code could be modified to accommodate an invalid icon without too much trouble.

code/bmpman/bmpman.cpp
ngld: This only happens if a file exists but can't be opened. I think this should never happen. Especially since the file would be ignored in this case (a return value of `-1` bascially means the file doesn't exist).

code/mission/missionbriefcommon.cpp
ngld: I wanted to keep the semantic of the previous assert with a better error message.
The briefing animation currently checks for `-1` in `first_frame` and just `Int3`s in that case. I guess that could be changed but I'm not sure if that's worth the effort since then you'll just get a briefing screen with a lot of missing icons.

I just noticed that there are a lot of similar `Assert`s in this file. I probably should change them as well.

code/bmpman/bmpman.cpp
Goober5000: Ok, I stand corrected.

Goober5000:
Did you push your latest changes?  GitHub shows only one commit.

ngld:
I replaced my previous commit with a new one (which is why it's listed below the other comments).

Goober5000:
Gotcha.  Just reviewed it, LGTM.

asarium:
Tea CI is becoming more and more unreliable. We might need to disable that if it goes on...

Issue #1842, opened 2018-08-28
Hud is missing after you were an observer


Artyrm:
You go to multiplayer mission
You die. You go observer. You got your hud off. Mission ends. You go to another. And your hud is still off.

Workaround: go to F2-options and reset you HUD there.

Goober5000:
Closing as fixed.

PR #1899, opened 2018-10-13
Fix missing HUD bug

Artyrm:
It appears that HUD_backup mechanics was implemented (to save players HUD before turning everything off for Observer). But restore was never used. That addresses #1842 , and some requests at HLP forums.

### DIFF
--- a/code/network/multi_endgame.cpp
+++ b/code/network/multi_endgame.cpp
@@ -22,7 +22,8 @@
 #include "network/multiui.h"
 #include "network/multiutil.h"
 #include "network/multi_pmsg.h"
-#include "network/multi_observer.h"
+#include "hud/hudconfig.h"
+//#include "network/multi_observer.h"
 #include "fs2netd/fs2netd_client.h"
 
 
@@ -323,8 +324,8 @@ void multi_endgame_cleanup()
 {
 	int idx;
 
-	if (Player_obj->type == OBJ_OBSERVER) {		// Do not forget restore players HUD
-		multi_obs_restore_hud();
+	if (HUD_config.is_observer) {		// Do not forget restore players HUD
+		hud_config_restore();
 	}
 
 	send_leave_game_packet();			
--- a/code/network/multi_endgame.cpp
+++ b/code/network/multi_endgame.cpp
@@ -22,6 +22,7 @@
 #include "network/multiui.h"
 #include "network/multiutil.h"
 #include "network/multi_pmsg.h"
+#include "network/multi_observer.h"
 #include "fs2netd/fs2netd_client.h"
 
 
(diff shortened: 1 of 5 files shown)


code/network/multi_endgame.cpp
```diff
@@ -322,6 +323,10 @@ void multi_endgame_cleanup()
 {
 	int idx;
 
+	if (HUD_config.is_observer) {		// Do not forget restore players HUD
+		hud_config_restore();
```
asarium: I think it would be better to use `hud_config_as_player` here since the opposing function `hud_config_as_observer` was used for creating the backup.

freespace2/freespace.cpp
```diff
@@ -5760,6 +5760,9 @@ void game_enter_state( int old_state, int new_state )
 		case GS_STATE_DEBRIEF:
 			game_stop_looped_sounds();
 			mission_goal_fail_incomplete();				// fail all incomplete goals before entering debriefing
+			if (HUD_config.is_observer) {				// restore player's HUD if it was observer
+				hud_config_restore();
```
asarium: Same here.

code/network/multi_endgame.cpp
Artyrm: I have an argument on this:
It was my initial train of thoughts to use `hud_config_as_player`. But it only do `hud_config_restore()` so it;s name could be misleading - since I can imagine situation in which factual _observer hud_ will be restored.

asarium: The problem I see here is that the original `hud_config_backup` call was made through `hud_config_as_observer` so if you call `hud_config_restore` directly instead of `hud_config_as_player` you effectively rely on the internal implementation of `hud_config_as_observer` which is never a good idea.

Yes, all `hud_config_as_player` does is call a single function but if the operation is extended in the future then the places which call `hud_config_restore` now may be missed or would need to be updated.

Artyrm:
Tested situation
5) You're a client and connection to server suddenly is lost.

I think, this concludes my tests.

code/network/multi_endgame.cpp
Artyrm: What shell we do about `hud_config_as_player` is plain misleading then?

asarium: Why is it misleading? It's the opposite to `hud_config_as_observer` so it does what the name says...

Artyrm: As I said earlier: what it actually do not necessarily sets hud config as player. It could set it as observer same way.

asarium: Then maybe `hud_config_as_player` should check `HUD_config.is_observer` before calling `hud_config_restore`? Then it would only restore the HUD when the observer mode was active before.

Artyrm: Yes! How come I couldn't figure it myself before :) It is all because of that smeared state context.
Will do.

asarium: Welcome to the world of game development where every bit of game state is scattered over three different files...

Artyrm: While it could be like that. How I miss those perfectly atomic states.

observer

Artyrm:
Before that:

I have tested this with the following variants:

You quit via Shift+ESC.
You quit via ESC (yes, that's a different flow of events).
You end mission via Alt+J being a server;
Server end mission while you're a client.

PR #1902, opened 2018-10-17
Play simulated speech if mission is launched while briefing is paused

### LINKED ISSUE #1900: Voice Synth is Muted Ingame If Mission is Committed While Briefing is Paused
Steps to duplicate:

1. Configure voice synth to play during both briefing and in flight 
2. Load a campaign with no voice acting
3. During mission briefing, hit the pause button.  Voice synth should stop midsentence.
4. Commit the mission

Expectation: Voice synth should resume in flight
Result: Voice synth remains muted.

Exiting the mission and starting the briefing with Pause already enabled will allow voice synth to resume for the first phase of the briefing.  Launching the mission in this state will allow voice synth to continue.

Can we separate voice synth instances/sessions between briefing and mission, such that voice synth will always be enabled in flight, unless disabled in the launcher?

It's probably worth mentioning that this has been an issue for years.  This wasn't caused by any recent updates.

plieblang:
Fixes #1900

### DIFF
--- a/code/missionui/missionshipchoice.cpp
+++ b/code/missionui/missionshipchoice.cpp
@@ -1932,7 +1932,7 @@ void commit_pressed()
 
 	// plieblang - resume simulated speech if necessary
 	if (!Player->auto_advance) {
-		fsspeech_pause(Player->auto_advance);
+		fsspeech_pause(false);
 	}
 
 	// save the player loadout
--- a/code/missionui/missionshipchoice.cpp
+++ b/code/missionui/missionshipchoice.cpp
@@ -16,7 +16,6 @@
 #include "cfile/cfile.h"
 #include "cmdline/cmdline.h"
 #include "freespace.h"
-#include "sound/speech.h"
 #include "gamehelp/contexthelp.h"
 #include "gamesequence/gamesequence.h"
 #include "gamesnd/gamesnd.h"
@@ -48,6 +47,7 @@
 #include "popup/popup.h"
 #include "render/3d.h"
 #include "ship/ship.h"
+#include "sound/fsspeech.h"
 #include "species_defs/species_defs.h"
 #include "weapon/weapon.h"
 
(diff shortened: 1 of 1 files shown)


code/missionui/missionshipchoice.cpp
```diff
@@ -1929,6 +1930,11 @@ void commit_pressed()
 	if (!(The_mission.flags[Mission::Mission_Flags::No_briefing]))
 		gamesnd_play_iface(InterfaceSounds::COMMIT_PRESSED);
 
+	// plieblang - resume simulated speech if necessary
+	if (speech_init && !Player->auto_advance) {
```
asarium:
```suggestion
	if (speech_init() && !Player->auto_advance) {
```
`speech_init` probably needs to be a function call.

plieblang: I'm not sure how that compiled locally.
Anyway, it looks like `speech_init` creates a new voice instance whereas we only want to resume the current voice, so I've changed it in the new commit.

code/missionui/missionshipchoice.cpp
```diff
@@ -1929,6 +1930,11 @@ void commit_pressed()
 	if (!(The_mission.flags[Mission::Mission_Flags::No_briefing]))
 		gamesnd_play_iface(InterfaceSounds::COMMIT_PRESSED);
 
+	// plieblang - resume simulated speech if necessary
+	if (!Player->auto_advance) {
+		fsspeech_pause(Player->auto_advance);
```
asarium: Since `Player->auto_advance` is already known to be `false` here could you pass `false` explicitly as the parameter?

```suggestion
		fsspeech_pause(false);
```

asarium:
Fortunately, Travis CI now has Windows support which includes MinGW (see #1909) so we don't need Tea CI any more.

PR #1903, opened 2018-10-18
Update displayed POF name in the F3 lab when new ship is selected

plieblang:
Currently, the POF filename shown in the bottom right is correct for the first model chosen but does not update when any other ship is selected.


plieblang:
Yes it does.

PR #1905, opened 2018-10-18
Add '$On Ship Depart' Conditional Hook

wookieejedi:
Previously there were no hooks that could be used to tell if a ship departed via docking bay, because the only departure related hook, "On Warp Out", only triggered when a ship departed via hyperspace. This PR adds a conditional hook that executes when a ship departs via any departure method (ie docking bay). It is called with "$On Ship Depart".

### DIFF
--- a/code/scripting/scripting.cpp
+++ b/code/scripting/scripting.cpp
@@ -86,6 +86,7 @@ flag_def_list Script_actions[] =
 	{ "On Simulation",			CHA_SIMULATION,		0 },
 	{ "On Load Screen",			CHA_LOADSCREEN,		0 },
 	{ "On Campaign Mission Accept", 	CHA_CMISSIONACCEPT,	0 },
+    { "On Ship Depart",			CHA_ONSHIPDEPART,	0 },
 };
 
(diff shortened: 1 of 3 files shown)


wookieejedi:
That's a good idea. So the scripter could call what departure method was used...I will see what I can do!

Goober5000:
Wish I had seen this two days ago...

1) It's not a good idea to separate a function and its associated `_helper` function, since they are conceptually very closely related.
2) The script hook is in the wrong place.  It should go in `ship_actually_depart_helper`, or perhaps in `ship_cleanup`.  This is because the first thing `ship_actually_depart` does is call the helper for all docked ships in the group.  In this PR as written, the script hook will not execute for any docked ships.

wookieejedi:
1. Noted, and good to know for future reference. 

2. I will move it to `ship_actually_depart_helper` if that sounds good.

Goober5000:
Sounds good. :+1:

wookieejedi:
> Sounds good. 👍

Is there a way to get the departure `method` from *objp?

wookieejedi:
I ended up adding it to `ship_cleanup` because that way I could still access the departure `method` for the hook variable.

Issue #1710, opened 2018-04-20
Ship lab improvements


plieblang:
Currently, clicking on a ship entry in the F3 lab only opens a menu where the user must click on "Model" to actually view the ship.  Clicking on the entry should display the ship directly.
Scrolling through the entries with the arrow keys would be a big improvement when trying to find a ship quickly and when trying to find tabling errors.

plieblang:
Both of these have been merged.

PR #1906, opened 2018-10-19
Show ship directly when clicking on it in the F3 lab

plieblang:
Addresses part of #1710 
It looks like the current behavior is a leftover from the old lab, where each entry had options for viewing ship LODs.  Now it's just a nuisance.

### DIFF
--- a/code/lab/lab.cpp
+++ b/code/lab/lab.cpp
@@ -1282,7 +1282,7 @@ void labviewer_make_desc_window(Button * /*caller*/)
 void labviewer_make_ship_window(Button * /*caller*/)
 {
 	GUIObject *cbp;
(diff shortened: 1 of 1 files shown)


code/lab/lab.cpp
```diff
@@ -1282,7 +1282,7 @@ void labviewer_make_desc_window(Button * /*caller*/)
 void labviewer_make_ship_window(Button * /*caller*/)
 {
 	GUIObject *cbp;
-	TreeItem *ctip, *stip;
+	TreeItem /**ctip, */*stip;
```
asarium: The commented out part of this line should be removed if it isn't used any more.

PR #1907, opened 2018-10-19
ship-invisible doesn't turn off autotargeting

### LINKED ISSUE #1855: ship-invisible SEXP turns off Auto Target
When the ship-invisible SEXP is applied to a ship auto target is switched off.  The effects of this SEXP should be confined to the ship to which it has been applied and not to the player's HUD settings.  See [https://youtu.be/zx4FbNco4zQ](url) for a video. Per m|m "hud_cease_targeting unsets the flag for auto targetting." and "I have no idea why that SEXP doesn't just use the simpler set_target_objnum..."

plieblang:
Fixes #1855 
From the comment, `hud-cease-targeting` turning off autotargeting may be intentional.  For safety's sake, I created a new function for the SEXPs, but since `hud-cease-targeting` is only used otherwise as a default case when rendering the target box and when the `STOP-TARGETING-SHIP` key is pressed, changing it might be preferable.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -4383,6 +4383,13 @@ void hud_cease_subsystem_targeting(int print_message)
 	hud_lock_reset();
 }
 
+//like hud_cease_targeting but doesn't turn off auto-targeting
+void hud_cease_targeting_ship() {
+	set_target_objnum(Player_ai, -1);
+	hud_cease_subsystem_targeting(0);
+	//don't need to call hud_lock_reset because hud_cease_subsystem_targeting already does
+}
+
 // hud_cease_targeting() will cease all targeting (main target and subsystem)
 //
 void hud_cease_targeting()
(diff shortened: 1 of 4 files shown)


Goober5000:
My understanding is that `hud_cease_targeting` was originally coded for the "turn off targeting" command.  If the player explicitly turns off his targeting system, he's not going to be doing any auto-targeting either.  But then this function was picked up and reused for the `ship-invisible` case which isn't quite the same thing.

Instead of writing a new function, I recommend adding a new boolean parameter `deliberate` which controls the printing of the message and the deactivating of auto-targeting.  (While you're at it, change the subsystem `print_message` parameter to boolean too.)

Goober5000:
Looks good, mostly.  However, I recommend defaulting `deliberate` to false, and only specifying true in keycontrol.cpp.

Goober5000:
Good stuff.  That will follow the principle of least astonishment for any future use of that function.

PR #1908, opened 2018-10-20
Add scripting file handle auto-closing and error message 

asarium:
This adds some auto-closing code to the scripting API of the CFile
system. If a file has to be auto-closed then an error will be displayed
to let the the script writer know that something is wrong.

This should not break existing scripts since causing file handle leaks
should never happen and if it does, this will make it easier to find
these issues.

This depends on the changes of #1881.

### DIFF
--- a/code/cfile/cfile.h
+++ b/code/cfile/cfile.h
@@ -13,10 +13,11 @@
 #define __CFILE_H__
 
 
-#include <ctime>
 #include "globalincs/pstypes.h"
 
+#include <ctime>
 #include <stdexcept>
+#include <memory>
 
 
 #define CF_EOF (-1)
@@ -461,5 +462,18 @@ namespace cfile
 	};
 
 }
+
+// This allows to use std::unique_ptr with CFILE
+namespace std {
+template <>
+struct default_delete<CFILE> {
+	void operator()(CFILE* ptr)
+	{
+		if (cf_is_valid(ptr)) {
+			cfclose(ptr);
+		}
+	}
+};
+} // namespace std
 
 #endif	/* __CFILE_H__ */
--- a/code/scripting/api/objs/file.cpp
+++ b/code/scripting/api/objs/file.cpp
@@ -6,16 +6,30 @@ extern int cfread_lua_number(double *buf, CFILE *cfile);
 namespace scripting {
 namespace api {
 
+cfile_h::cfile_h(CFILE* cfp) : _cfp(cfp) {}
+cfile_h::~cfile_h()
+{
+	if (_cfp && cf_is_valid(_cfp.get())) {
+		Error(LOCATION, "Lua file handle has been left open!\n\nSorry, I can't say anything more than that...");
+	}
+}
+bool cfile_h::isValid() const { return _cfp != nullptr; }
+CFILE* cfile_h::get() const { return _cfp.get(); }
+void cfile_h::close()
+{
+	_cfp = nullptr; // Automatically closes the handle
+}
+
 //**********HANDLE: File
-ADE_OBJ(l_File, CFILE*, "file", "File handle");
+ADE_OBJ(l_File, cfile_h, "file", "File handle");
 
 ADE_FUNC(isValid, l_File, NULL, "Detects whether handle is valid", "boolean", "true if valid, false if handle is invalid, nil if a syntax/type error occurs")
 {
-	CFILE *cfp = NULL;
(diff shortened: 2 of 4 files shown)


code/cfile/cfile.h
```diff
@@ -462,4 +463,17 @@ namespace cfile
 
 }
 
+// This allows to use std::unique_ptr with CFILE
+namespace std {
+template <>
+struct default_delete<CFILE> {
```
CasparKielwein: What do you think about adding a deleter to the unique_ptr as its second template argument?
That makes the intend clearer as the fact that the unique pointer closes the file and doesn't free memory is visible at points of use and not hidden in this template specialization.

code/scripting/api/objs/file.cpp
```diff
@@ -6,63 +6,72 @@ extern int cfread_lua_number(double *buf, CFILE *cfile);
 namespace scripting {
 namespace api {
 
+cfile_h::cfile_h(CFILE* cfp) : _cfp(cfp) {}
```
CasparKielwein: Is implicit conversion intended here? If not I'd declare the constructor explicit.

code/cfile/cfile.h
asarium: That would require using the correct template types for all `std::unique_ptr<CFILE>` usages which could be a problem and would also not be easily detectable since the code would compile but it would fail once the `unique_ptr` was deleted.

With this, it is always valid to use `std::unique_ptr<CFILE>`.

CasparKielwein: I personally wouldn't expect a `unique_ptr` to do more than delete the memory resource when I don't see a custom deleter.

I'm not sure if your deleter wouldn't leak the CFILE, if someone for whatever reasons allocates one on the heap and puts it into the `unique_ptr`.

Maybe the safest solution is a `CfileHandle` struct, which does the clean-up in its destructor and is move only, then memory management and possible CFILEs on the heap are separated from file clean-up.

asarium: `CFILE` is an incomplete type so no one could allocate one on the heap without hacking into the internals of the CFile system.

Also, this code doesn't break the assumption that `unique_ptr` simply deallocates memory (which isn't even universally true) because it actually does "deallocate" memory by passing the pointer to `cfclose`.

CasparKielwein: Isn't CFILE defined in the cfile.h, public header? Maybe its definition could be moved to the .cpp

If it is an incomplete type the generic default deleter and thus unique pointer would not compile.

I'm not against the changeset in general. I just think putting the behaviour into a custom deleter makes intent clearer, but I admit that in the end its more a matter of personal preference.

asarium: Huh, I must have misremembered that somehow.

Anyway, a valid `CFILE` pointer can only be created by the cfile system so using `cfclose` is the correct way for deleting that so I think using the default deleter way is the correct solution here.

code/scripting/api/objs/file.cpp
```diff
@@ -6,63 +6,72 @@ extern int cfread_lua_number(double *buf, CFILE *cfile);
...
+cfile_h::~cfile_h()
+{
+	if (_cfp && cf_is_valid(_cfp.get())) {
+		Error(LOCATION, "Lua file handle has been left open!\n\nSorry, I can't say anything more than that...");
```
Goober5000: Is it possible to say what file name the handle refers to?  If not, could that functionality be added easily?

asarium: No, unfortunately that is not possible without doing some rather extensive changes to the CFILE system. I wanted to do that for the CFILE leak detection I added some time ago but that information is not stored anywhere at the moment.

Goober5000:
With that one issue clarified, I'm happy to approve this for merging, assuming the previous conversation with @CasparKielwein can be resolved.

There is, of course, the requisite clang-tidy complaint: "warning: move constructors should be marked noexcept [misc-noexcept-move-constructor]".

CasparKielwein:
Our Discussion doesn't block merging.
It comes down to matters of preference.

asarium:
@Goober5000 I think that warning is a false-positive since the constructor is in fact marked as `noexcept`. I marked it with a `NOLINT`.

Goober5000:
@CasparKielwein I understand that merging is not blocked, but I would rather not prematurely merge the PR if the discussion is not finished.  I don't like to assume discussions are finished unless they are marked as resolved.

PR #1911, opened 2018-10-21
Add option for disabling spark spawning for subsystem

asarium:
This adds the "no sparks" flag for subsystems which disables the
generation of spark sources for that subsystem when it is hit.

This was requested by Axem.

### DIFF
--- a/code/model/model_flags.h
+++ b/code/model/model_flags.h
@@ -47,7 +47,8 @@ namespace Model {
         Autorepair_if_disabled,		// Allows the subsystem to repair itself even if disabled (MageKing17)
         No_autorepair_if_disabled,	// Inversion of the previous; disallows this particular subsystem if the ship-wide flag is set (MageKing17)
         Share_fire_direction,		// (DahBlount) Whenever the turret fires, make all firing points fire in the same direction.
-		
+        No_sparks,          // Subsystem does not generate sparks if hit - m!m
+
         NUM_VALUES
 	};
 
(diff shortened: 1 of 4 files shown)


asarium:
After a short discussion with @AxemP I renamed the option to "no damage spew" to make it more in line with other options already present in the engine.

code/model/model_flags.h
```diff
@@ -47,7 +47,9 @@ namespace Model {
         Autorepair_if_disabled,		// Allows the subsystem to repair itself even if disabled (MageKing17)
         No_autorepair_if_disabled,	// Inversion of the previous; disallows this particular subsystem if the ship-wide flag is set (MageKing17)
         Share_fire_direction,		// (DahBlount) Whenever the turret fires, make all firing points fire in the same direction.
-		
+        No_sparks,          // Subsystem does not generate sparks if hit - m!m
+        No_debris_spawn,    // Don't spawn the small debris on impact - m!m
```
Goober5000: I suggest renaming the second flag to `No_impact_debris`, both to match the text and also to be a little clearer.  (I initially thought this had something to do with live debris.)

code/ship/shiphit.cpp
```diff
@@ -514,7 +514,11 @@ float do_subobj_hit_stuff(object *ship_objp, object *other_obj, vec3d *hitpos, i
 #endif
 
 	if (!global_damage) {
-		create_subsys_debris(ship_objp, hitpos);
+		auto submodel = ship_get_subsys_for_submodel(ship_p, submodel_num);
```
Goober5000: Why not use `subsys` for the variable name instead of `submodel`?  (`subsys` is used later in the diff.)

code/model/model_flags.h
asarium: Thanks, that is a much better name for that.

code/ship/shiphit.cpp
asarium: Thanks, that name was clearly wrong.

Also, I noticed that the condition after this was also wrong so I fixed that as well.

PR #1913, opened 2018-10-24
calculate physics whack from docked center-of-mass, not docked center

Goober5000:
The physical center of a group of docked objects isn't necessarily the physical center-of-mass.


Goober5000:
I don't think that will be a big problem either.  Most of the time, when a ship is being whacked with something, it's a beam, and that probably means it will be destroyed shortly.  I made a similar improvement to the jettison-cargo physics whack prior to 3.8, and nobody complained.

The bigger issue is actually the fact that the code doesn't recalculate the moment of inertia for the entire docked assembly.  That may be why freighters with cargo tend to spin crazily.  But this has been the case ever since I added the new docking code in 2004-ish, so although it's a problem, it's a low-priority one.

Goober5000:
Having said that, calculating the correct MOI may be easier than I thought.  From Wikipedia:

>The moment of inertia of a rigid composite system is the sum of the moments of inertia of its component subsystems (all taken about the same axis).

PR #1914, opened 2018-10-25
only recompute collision pairs if we moved a colliding object

Goober5000:
The code already only recomputed collisions if we moved a ship or a wing, but this further optimizes things so that only ships or wings that can collide will cause a recomputation.


Goober5000:
Rebase and fix conflicts introduced by #1912

PR #1917, opened 2018-10-28
 Add arrow key scrolling to the lab

plieblang:
Uses the up/down keys.
Addresses the second part of #1710.

### DIFF
--- a/code/lab/lab.cpp
+++ b/code/lab/lab.cpp
@@ -70,6 +70,9 @@ static Window *Lab_background_window = NULL;
 static Text *Lab_description_text = NULL;
 static TreeItem **Lab_species_nodes = NULL;
 
+static Tree *ship_tree = nullptr;
+static Tree *weap_tree = nullptr;
+
 //holds the beginning and ending indices of each specie/type of ship/weapon
 static std::pair<TreeItem*, TreeItem*> *ship_list_endpoints = nullptr;
 static std::pair<TreeItem*, TreeItem*> *weap_list_endpoints = nullptr;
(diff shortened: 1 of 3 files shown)


code/lab/lab.cpp
```diff
@@ -1323,15 +1333,20 @@ void labviewer_make_ship_window(Button * /*caller*/)
 		Lab_species_nodes = NULL;
 	}
 
+	ship_list_endpoints = new std::pair<TreeItem*, TreeItem*>[Species_info.size()];
```
asarium: This should use a vector instead. There is no need to use dynamic allocation here.

code/lab/lab.cpp
```diff
@@ -1516,17 +1543,20 @@ void labviewer_make_weap_window(Button*  /*caller*/)
 	x += cbp->GetWidth() + 10;
 	cbp = Lab_class_toolbar->AddChild(new Button("Class Variables", x, 0, labviewer_make_variables_window));
 
+	weap_list_endpoints = new std::pair<TreeItem*, TreeItem*>[Num_weapon_types];
```
asarium: Same as above.

asarium:
Please describe what your changes to in more detail. The current description isn't very useful.

asarium:
Also, this crashes when pressing up or down when no ship is selected.

plieblang:
Functionally, this allows for scrolling through the list of ships and weapons with the up/down keys.  When at the beginning or end of a list of ships or weapons for a certain species (in the case of ships) or type (for weapons), nothing will happen.
Code-wise, this works by tracking the currently displayed item, and then when the up or down key is pressed, getting the previous or next item in the complete list, checking to make sure it's of the same species/type and then displaying it.

code/lab/wmcgui.h
```diff
@@ -542,6 +542,7 @@ class Tree : public GUIObject
 	void ClearItems();
 
 	TreeItem* GetSelectedItem(){return SelectedItem;}
+	void SetSelectedItem(TreeItem *item) { SelectedItem = item; }
```
asarium: This function changes the selected item but does not fire the listener function of that item. I would suggest that you move the relevant code from `Tree::DoMouseUp` into this function and then use this function in `Tree::DoMouseUp` to keep the functionality in one place.

Also, since the function will be quite big then it would be better to move the body into `wmcgui.cpp`.

asarium:
Looks good! Thank you for making the requested changes.

PR #1920, opened 2018-11-01
Make the lab camera better

The-E:
This changes the model rotation method from a "trackball" to a "turntable" model. 
By default, rotations are entirely disabled in the lab. By pressing "R", you can cycle through the rotation modes (Disabled -> Yaw only -> Pitch only -> Yaw and Pitch -> Disabled), the "S" key allows you to set the rotation speed (Fast -> Slow -> Slowest -> Fast).

NOTE: This only affects the model (i.e. rotations while holding down the left mouse button), not the background/camera (i.e. rotations while holding the right mouse button).

### DIFF
--- a/code/lab/lab.cpp
+++ b/code/lab/lab.cpp
@@ -113,6 +113,7 @@ static float Lab_manual_rotation_speed_divisor = 100.f;
 #define LAB_ROTATION_MODE_YAW 1
 #define LAB_ROTATION_MODE_PITCH 2
 #define LAB_ROTATION_MODE_BOTH 3
+#define LAB_ROTATION_MODE_ROLL 4
 
 // Trackball_mode:
 //   1  ==  rotate	(left-mouse)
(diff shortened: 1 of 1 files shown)


asarium:
With these changes, I cannot rotate the ship with the left mouse button any more. Is this intentional?

The-E:
Sort of?
The rotation control starts off in the "Disabled" state, you need to press R to cycle into a mode where you can actually spin things.

asarium:
I think it's counter-intuitive to have the action of the left mouse button disabled by default. Maybe the camera rotation could be controlled with the left mouse button and the ship rotation with the right?

code/lab/lab.cpp
```diff
@@ -97,77 +94,87 @@ static char Lab_weaponmodel_filename[MAX_SHIP_WEAPONS][MAX_FILENAME_LEN];
...
 static int anim_timer_start = 0;
 
-static float Lab_thrust_len = 1.0f;
+static float Lab_thrust_len      = 1.0f;
 static bool Lab_thrust_afterburn = false;
 
+static int Lab_rotation_mode                   = 0;
+static float Lab_manual_rotation_speed_divisor = 100.f;
+
+#define LAB_ROTATION_MODE_YAW 0
```
asarium: These defines should be an `enum class` or a simple `enum` if necessary.

code/lab/lab.cpp
```diff
@@ -97,77 +94,87 @@ static char Lab_weaponmodel_filename[MAX_SHIP_WEAPONS][MAX_FILENAME_LEN];
...
 SCP_string Lab_team_color = "<none>";
 
 camid Lab_cam;
 float lab_cam_distance = 100.0f;
-float lab_cam_phi = 1.24f; // Values chosen to approximate the initial rotation used previously
-float lab_cam_theta = 2.25f;
-bool Lab_render_wireframe = false;
-bool Lab_render_without_light = false;
+float lab_cam_phi      = 0;
+// 1.24f; // Values chosen to approximate the initial rotation used previously
```
asarium: These comments are outdated and not necessary any more.

asarium:
@The-E Are there any changes left that you would like to add here?

The-E:
Not at this time.

PR #1926, opened 2018-11-08
Set initial slider values to the command line parameters and add functionality to reset view

plieblang:
Currently, the rendering options sliders in the lab for ambient, static and bloom are initially set to 0 by default, even though the actual values are what is passed in on the command line.  This PR changes that so that the sliders accurately reflect the true values.
This also adds the ability to reset a ship to the default position, rotation and lighting when 'v' is pressed.
These changes should improve the experience of trying out different lighting settings.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,11 +9,11 @@ environment:
     - BuildID: 0
       CMAKE_GENERATOR: Visual Studio 14 2015
       PlatformToolset: v140_xp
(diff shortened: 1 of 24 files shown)


plieblang:
Hmm, it looks like I did something wrong with Git.

plieblang:
I'll create a new PR with a better Git history.

PR #1928, opened 2018-11-11
Fix access violation in ship_get_global_turret_gun_info

TRBlount:
When creating the Share Fire Direction flag, I overlooked the value used to access the `Ships[]` array, using an objnum instead of the correct `object.instance` variable. This should be fixed now.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1290,7 +1290,7 @@ void ship_get_global_turret_gun_info(object *objp, ship_subsys *ssp, vec3d *gpos
 		model_instance_find_world_point(&tmp_pos, &avg, Ships[objp->instance].model_instance_num, tp->turret_gun_sobj, &objp->orient, &objp->pos);
 
 		if (targetp == nullptr) {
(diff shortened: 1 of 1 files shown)


code/ai/aiturret.cpp
```diff
@@ -1290,7 +1290,7 @@ void ship_get_global_turret_gun_info(object *objp, ship_subsys *ssp, vec3d *gpos
 		model_instance_find_world_point(&tmp_pos, &avg, Ships[objp->instance].model_instance_num, tp->turret_gun_sobj, &objp->orient, &objp->pos);
 
 		if (targetp == nullptr) {
-            Assertion(ssp->turret_enemy_objnum >= 0, "The turret enemy object number %d for %s on ship number %d is invalid.", ssp->turret_enemy_objnum, ssp->sub_name, ssp->parent_objnum);
+            Assertion(ssp->turret_enemy_objnum >= 0, "The turret enemy object number %d for %s on ship number %d is invalid.", ssp->turret_enemy_objnum, ssp->sub_name, Objects[ssp->parent_objnum].instance);
```
Goober5000: I recommend putting the ship name in the Assertion, rather than the ship number.

PR #1932, opened 2018-11-20
Fix warp clipping

### LINKED ISSUE #1874: warp-in clipping is not rendered correctly
This bug can be observed as far back as 3.6.10.  I wanted to test earlier builds but scp.indiegames.us is currently offline.

When the blue warp opening appears, the ship does not emerge smoothly out of the warp effect.  Instead, maybe the first 10% of the front of the ship appears suddenly, then the rest of the ship gradually follows.

In retail, the entire ship emerges gradually from the warp effect, there is no 10% head start.

Possibly related to #1867.

Goober5000:
This is a rewrite of the code which calculates positions and distances for warpin and warpout.  Instead of assuming the object's length is twice its radius, it calculates the size using the model minimums and maximums.  Instead of using the object's center, it calculates the geometric center using the same method.  In this way all model offsets are properly canceled out and the warp portals are displayed at the proper coordinates.  The various hackish adjustments for ships like the Colossus have been removed.

Note: ship warpin/warpout distances using the new code are not precisely identical to the old code, but they are pretty close.

This fixes #1874.  This also fixes a potential bug where a Knossos could suddenly jump position when a HUGE ship comes through the portal but not through the exact center.

### DIFF
--- a/code/math/vecmat.h
+++ b/code/math/vecmat.h
@@ -69,7 +69,7 @@ extern matrix vmd_identity_matrix;
 #define vm_angvec_make(v,_p,_b,_h) (((v)->p=(_p), (v)->b=(_b), (v)->h=(_h)), (v))
 
 //negate a vector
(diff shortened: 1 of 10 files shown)


niffiwan:
the relevant CI's have passed, I'm going to ignore the branch one same as #1942

PR #1933, opened 2018-11-21
multi capability for destroy-instantly and destroy-subsys-instantly

Goober5000:
Previously, these two SEXPs could only be used in single-player; now they can be used in multiplayer.  (This now allows Admiral MS's ship save/load script to work in multiplayer too.)

Inspired by Parias's patch included in the [FSPort Co-Op Mod Beta](https://www.hard-light.net/forums/index.php?topic=88385.0) though with some modifications.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -11750,69 +11750,106 @@ void sexp_set_subsystem_strength(int n)
 }
 
 // destroys a subsystem without explosions
-void sexp_destroy_subsys_instantly (int n)
+void sexp_destroy_subsys_instantly(int n)
 {
 	char *subsystem;
-	int	shipnum, index, generic_type;
+	int	shipnum, subsys_index, generic_type;
 	ship *shipp;
-	ship_subsys *ss = NULL;
-	
-	// if MULTIPLAYER bail
-	if (Game_mode & GM_MULTIPLAYER) {
-		return;
-	}
+	ship_subsys *ss;
 
 	shipnum = ship_name_lookup(CTEXT(n));
 	// if no ship, then return immediately.
-	if ( shipnum == -1 ){
+	if (shipnum < 0) {
 		return;
 	}
 
 	shipp = &Ships[shipnum];
 	n = CDR(n);
-	
-	//Process subsystems
-	while(n != -1)
+
+	if (MULTIPLAYER_MASTER)
+	{
+		Current_sexp_network_packet.start_callback();
+		Current_sexp_network_packet.send_ship(shipp);
+	}
+
+	for ( ; n >= 0; n = CDR(n))
 	{
 		subsystem = CTEXT(n);
-		if ( !stricmp( subsystem, SEXP_HULL_STRING) || !stricmp( subsystem, SEXP_SIM_HULL_STRING)){
-			n = CDR(n);
+
+		// use destroy-instantly if we want to do this
+		if (!stricmp(subsystem, SEXP_HULL_STRING) || !stricmp(subsystem, SEXP_SIM_HULL_STRING))
 			continue;
-		}
+
 		// deal with generic subsystems
 		generic_type = get_generic_subsys(subsystem);
-		if (generic_type) {
-			for (ss = GET_FIRST(&shipp->subsys_list); ss != END_OF_LIST(&shipp->subsys_list); ss = GET_NEXT(ss)) {
-				if (generic_type == ss->system_info->type) {
+		if (generic_type != SUBSYSTEM_NONE)
+		{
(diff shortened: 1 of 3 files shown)


code/parse/sexp.cpp
```diff
@@ -11737,69 +11737,106 @@ void sexp_set_subsystem_strength(int n)
...
-				nprintf(("Warning", "Couldn't find subsystem %s on ship %s for destroy-subsys-instantly\n", subsystem, shipp->ship_name));
-				n = CDR(n);
-				continue;
-			}
-			if(ss == NULL) {
-				nprintf(("Warning", "Nonexistent subsystem for index %d on ship %s for destroy-subsys-instantly\n", index, shipp->ship_name));
+			if (ss == nullptr)
+			{
+				nprintf(("Warning", "Nonexistent subsystem '%s' on ship %s for destroy-subsys-instantly\n", subsystem, shipp->ship_name));
 				n = CDR(n);
```
asarium: Is this still correct since `n = CDR(n)` is already done by the `for` loop?

code/ship/ship.cpp
```diff
@@ -13396,6 +13395,29 @@ int ship_get_subsys_index(ship *sp, const char* ss_name, int error_bypass)
...
+	ss = GET_FIRST(&shipp->subsys_list);
+	while (ss != END_OF_LIST(&shipp->subsys_list)) {
+		if (ss == subsys)
+			return count;
+		count++;
+		ss = GET_NEXT(ss);
+	}
+
+	if (!error_bypass)
+		Int3();
```
asarium: `Int3()` is more or less deprecated since it doesn't show an error message. This should be replaced by an `Assertion` with a proper error message.

code/ship/ship.h
```diff
@@ -1507,6 +1507,7 @@ extern void ship_get_eye( vec3d *eye_pos, matrix *eye_orient, object *obj, bool
 extern ship_subsys *ship_get_indexed_subsys( ship *sp, int index, vec3d *attacker_pos = NULL );	// returns index'th subsystem of this ship
 extern int ship_get_index_from_subsys(ship_subsys *ssp, int objnum, int error_bypass = 0);
 extern int ship_get_subsys_index(ship *sp, const char* ss_name, int error_bypass = 0);		// returns numerical index in linked list of subsystems
+extern int ship_get_subsys_index(ship *shipp, ship_subsys *subsys, int error_bypass = 0);
```
asarium: Since `error_bypass` is used as a boolean variable it should be a `bool`.

code/parse/sexp.cpp
```diff
@@ -14750,9 +14782,32 @@ void sexp_destroy_instantly(int n)
...
-			if (ship_obj_p == Player_obj)
-				continue;
-			ship_destroy_instantly(ship_obj_p,ship_num);
+			if (ship_obj_p != Player_obj) {
+				ship_destroy_instantly(ship_obj_p, ship_num);
+
+				// multiplayer callback
+				Current_sexp_network_packet.start_callback();
+				Current_sexp_network_packet.send_ship(ship_num);
+				Current_sexp_network_packet.end_callback();
```
asarium: Just to make sure I understand this correctly, if `Current_sexp_network_packet.start_callback()` is called multiple times then it basically sends multiple SEXP execution packets?

The reason I'm asking this is to make sure that this also works when specifying multiple ships.

code/parse/sexp.cpp
Goober5000: Good call.  The original patch was a while loop, and I changed it, but I missed this.

code/ship/ship.cpp
Goober5000: I was mirroring the other functions that performed similar tasks.  In my view all of them should have the Int3() removed since that was probably only needed during development. :man_shrugging: 

The best thing IMO would be to make another PR to remove those Int3()s.  I'll do that.

code/ship/ship.h
Goober5000: See previous comment re matching the other functions.

code/parse/sexp.cpp
Goober5000: It does.  But you're right, I should go back and modify that a bit.

code/ship/ship.cpp
asarium: The other functions could be changed (in another PR) but that doesn't mean that new code should use old functions just to keep the style of existing functions.

Goober5000:
See also #1937

code/parse/sexp.cpp
Goober5000: Should be fixed in the new commits.

Goober5000:
All feedback addressed in the new commits.

PR #1934, opened 2018-11-22
Set initial slider values to the command line parameters and add functionality to reset view

plieblang:
Redo of #1926 because I messed up the Git history there.
Currently, the rendering options sliders in the lab for ambient, static and bloom are initially set to 0 by default, even though the actual values are what is passed in on the command line. This PR changes that so that the sliders accurately reflect the true values.
This also adds the ability to reset a ship to the default position, rotation and lighting when 'v' is pressed.
These changes should improve the experience of trying out different lighting settings.

### DIFF
--- a/code/lab/lab.cpp
+++ b/code/lab/lab.cpp
@@ -173,7 +173,7 @@ void rotate_view(int dx, int dy)
 void reset_view()
 {
 	//reset position
-	Lab_model_pos = ZERO_VECTOR;
+	Lab_model_pos = vmd_zero_vector;
 	Lab_model_orient = vmd_identity_matrix;
 	//reset lighting
 	//if the rendering options window is close, these sliders are null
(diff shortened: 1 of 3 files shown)


code/lab/wmcgui.cpp
```diff
@@ -2701,11 +2701,13 @@ float Slider::GetSliderPos(int x)
 
 void Slider::UpdateSlider(int x)
```
asarium: I dislike how this function currently does two things at the same time which required you to add the `x >= 0` workaround.

Could you change it so that this function would be passed a `float` value which sets the `SliderScale` directly without calling `GetSliderPos`?

The existing `UpdateSlider(x)` invocations could be changed to `UpdateSlider(GetSliderPos(x))` to preserve the previous behaviour. This would lead to a much cleaner API and would generally be much easier to understand.

plieblang:
Besides the requested changes, I also changed the help text to include the reset view key in this commit.

code/lab/lab.cpp
```diff
@@ -160,6 +170,25 @@ void rotate_view(int dx, int dy)
 	CLAMP(lab_cam_phi, 0.01f, PI - 0.01f);
 }
 
+void reset_view()
+{
+	//reset position
+	Lab_model_pos = ZERO_VECTOR;
```
asarium: MSVC 2015 doesn't like this line since it has some kind of bug with nested initializer lists. Try using `vmd_zero_vector` instead.

PR #1939, opened 2018-11-28
objectsort: Move nebula detection to common area

TheMatthew:
Run nebula test once per frame rather than once
per object.

Use a common function to avoid drifting errors.


TheMatthew:
thanks!

PR #1942, opened 2018-11-30
Add On Weapon Created Hook

TRBlount:
This hook is distinct from `$On Weapon Fired:` in that it triggers each time `weapon_create` is called by the engine. Additionally, the weapon object itself is included as a hookvar `Weapon`.

### DIFF
--- a/code/scripting/scripting.cpp
+++ b/code/scripting/scripting.cpp
@@ -317,7 +317,7 @@ bool ConditionedHook::ConditionsValid(int action, object *objp, int more_data)
 					} else if(objp == NULL || objp->type != OBJ_SHIP) {
 						return false;
 					} else if (action == CHA_ONWEAPONCREATED) {
-						if (objp == NULL || objp->type != OBJ_WEAPON)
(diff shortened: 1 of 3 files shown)


AxemP:
I have done some basic tests with this and it works as expected.

code/scripting/scripting.cpp
```diff
@@ -315,6 +316,9 @@ bool ConditionedHook::ConditionsValid(int action, object *objp, int more_data)
 							return false;
 					} else if(objp == NULL || objp->type != OBJ_SHIP) {
 						return false;
+					} else if (action == CHA_ONWEAPONCREATED) {
+						if (objp == NULL || objp->type != OBJ_WEAPON)
```
niffiwan: Travis wants this to be nullptr; not NULL

niffiwan:
how do you restart `continuous-integration/appveyor/branch` anyway? :)

asarium:
I have no idea where that one comes from so I'll just ignore it. I think AppVeyor gets confused sometimes when building branches and pull requests...

PR #1943, opened 2018-12-03
Add a cache map for shaders in order to accelerate shader lookup

TheMatthew:
Shaders lookup in opengl_get_shader_idx performance reduces from
O(n) shader to O(1).

### DIFF
--- a/code/graphics/opengl/gropenglshader.cpp
+++ b/code/graphics/opengl/gropenglshader.cpp
@@ -338,7 +338,7 @@ void opengl_delete_shader(int sdr_handle)
 	Assert(sdr_handle >= 0);
 	Assert(sdr_handle < (int)GL_shader.size());
 	opengl_shader_t &victim = GL_shader[sdr_handle];
-	GL_shader_map.erase(std::make_pair(victim.shader, victim.flags));
+	GL_shader_map.erase(shader_descriptor_t(victim.shader, victim.flags));
 
 	GL_shader[sdr_handle].program.reset();
 	
--- a/code/graphics/opengl/gropenglshader.cpp
+++ b/code/graphics/opengl/gropenglshader.cpp
@@ -863,7 +863,7 @@ int opengl_compile_shader(shader_type sdr, uint flags)
 		sdr_index = (int)GL_shader.size();
 		GL_shader.push_back(std::move(new_shader));
 	}
-	GL_shader_map[std::make_pair(new_shader_shader, new_shader_flags)] = sdr_index;
+	GL_shader_map[shader_descriptor_t(new_shader_shader, new_shader_flags)] = sdr_index;
 	return sdr_index;
 }
 
--- a/code/graphics/opengl/gropenglshader.cpp
+++ b/code/graphics/opengl/gropenglshader.cpp
@@ -854,7 +854,7 @@ int opengl_compile_shader(shader_type sdr, uint flags)
 	}
 
 	int new_shader_shader = new_shader.shader;
-	unsigned int new_shader_flags = new_shader.flags;
+	uint32_t new_shader_flags = new_shader.flags;
 	// then insert it at an empty slot or at the end
 	if ( empty_idx >= 0 ) {
 		GL_shader[empty_idx] = std::move(new_shader);
--- a/code/graphics/opengl/gropenglshader.cpp
+++ b/code/graphics/opengl/gropenglshader.cpp
@@ -31,6 +31,18 @@
 
 SCP_vector<opengl_shader_t> GL_shader;
 
+typedef std::pair<int, uint32_t> shader_descriptor_t;
+
+struct key_hasher
+{
+    size_t operator()(const shader_descriptor_t &obj) const
+    {
+        return obj.first ^ obj.second;
+    }
+};
+
+SCP_unordered_map<shader_descriptor_t, size_t, key_hasher> GL_shader_map;
+
 GLuint Framebuffer_fallback_texture_id = 0;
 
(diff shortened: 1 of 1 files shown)


code/graphics/opengl/gropenglshader.cpp
```diff
@@ -296,12 +316,11 @@ void opengl_shader_set_current(int handle)
 
 size_t opengl_get_shader_idx(shader_type shader_t, unsigned int flags) 
 {
-	for (size_t idx = 0; idx < GL_shader.size(); idx++) {
-		if (GL_shader[idx].shader == shader_t && GL_shader[idx].flags == flags) {
-			return idx;
-		}
+	auto t = shader_descriptor_t(shader_t, flags);
```
asarium: You could avoid a copy operation here and improve the variable name. `t` isn't very descriptive...

```suggestion
	shader_descriptor_t descriptor(shader_t, flags);
```

code/graphics/opengl/gropenglshader.cpp
```diff
@@ -840,6 +862,8 @@ int opengl_compile_shader(shader_type sdr, uint flags)
 		}
 	}
 
+	int new_shader_shader = new_shader.shader;
+	unsigned int new_shader_flags = new_shader.flags;
```
asarium: Please use `uint32_t` here to make the code more explicit:
```suggestion
	uint32_t new_shader_flags = new_shader.flags;
```

code/graphics/opengl/gropenglshader.cpp
TheMatthew: Thanks for the review/feedback.

I will do this tonight.

Also, I tested locally, and everything looks fine. I'm afraid that the performance gain may be from lost shaders though that I didn't notice. I'm not saying my code is bad, but this is the most in-depth patch I've done so far, so please excercise caution while reviewing. :)

code/graphics/opengl/gropenglshader.cpp
TheMatthew: Sure, should shader.flags be updated too?

TheMatthew:
Thanks for the review.

code/graphics/opengl/gropenglshader.cpp
asarium: Sure, if you want to but it's not strictly necessary since you didn't touch that code.

code/graphics/opengl/gropenglshader.cpp
asarium: I would have told you if I found something wrong with the changes :stuck_out_tongue:

I looked though the code and your changes seem to make sense. I didn't compile and run the code though...

TheMatthew: So... you can't have a bad variable name or a copy if you don't have a variable. ;)

code/graphics/opengl/gropenglshader.cpp
```diff
@@ -840,6 +853,8 @@ int opengl_compile_shader(shader_type sdr, uint flags)
 		}
 	}
 
+	int new_shader_shader = new_shader.shader;
```
TheMatthew: Should this be int32_t?

code/graphics/opengl/gropenglshader.cpp
```diff
@@ -31,6 +31,18 @@
 
 SCP_vector<opengl_shader_t> GL_shader;
 
+typedef std::pair<int, uint32_t> shader_descriptor_t;
```
TheMatthew: Should this be int32_t too?

TheMatthew:
I am re-examining the performance impact, it is probably lower than my initial measurements, I found that my hard disk may have been a slowdown in the first round.

TheMatthew:
Tested with 8 runs before and after, discarding the outlyer to avoid caching effects.

1- the load times fluctuate enormously. (20%)
2- the actual gains are noticeable in the trace. On average there is a gain of 15% on that scope.

Performance gain, top is after, bottom is before.

code/graphics/opengl/gropenglshader.cpp
asarium: If you want you can use fixed size types wherever you want but it's not required unless the type name is very long (such as `unsigned int` or `unsigned long long`).

TheMatthew: won't then, I better understand the coding convention now.

code/graphics/opengl/gropenglshader.cpp
```diff
@@ -327,6 +337,8 @@ void opengl_delete_shader(int sdr_handle)
 {
 	Assert(sdr_handle >= 0);
 	Assert(sdr_handle < (int)GL_shader.size());
+	opengl_shader_t &victim = GL_shader[sdr_handle];
+	GL_shader_map.erase(std::make_pair(victim.shader, victim.flags));
```
asarium: I think you should use the `shader_descriptor_t` here:
```suggestion
	GL_shader_map.erase(shader_descriptor_t(victim.shader, victim.flags));
```

code/graphics/opengl/gropenglshader.cpp
```diff
@@ -848,7 +863,7 @@ int opengl_compile_shader(shader_type sdr, uint flags)
 		sdr_index = (int)GL_shader.size();
 		GL_shader.push_back(std::move(new_shader));
 	}
-
+	GL_shader_map[std::make_pair(new_shader_shader, new_shader_flags)] = sdr_index;
```
asarium: Same here.

asarium:
Looks good!

PR #1945, opened 2018-12-06
Fix move constructor of streaminganim_h

### LINKED ISSUE #1944: Assert: "block_index >= 0 && block_index < (int) bm_blocks.size()"
Something in @asarium 's commit (e901e44) on Nov 18 broke the visual novel script used by JAD and WoD. Release will crash to desktop without error, fast debug will give
```Assert: "block_index >= 0 && block_index < (int) bm_blocks.size()"
File: bmpman.cpp
Line: 135
Bitmap handle 1082402341 has an invalid block number 16516!
```

And a full debug build attached to Visual Studio will... work completely fine! I didn't look to see which commit it was exactly, but I'm hoping m!m will know more!

To reproduce the issue, play any VN mission in JAD (most _a or _c missions)

asarium:
The move constructor was written wrongly which caused issues when moving
the value into the Lua memory. This fixes that issue.

This fixes #1944.

### DIFF
--- a/code/scripting/api/objs/streaminganim.cpp
+++ b/code/scripting/api/objs/streaminganim.cpp
@@ -20,12 +20,16 @@ streaminganim_h::~streaminganim_h() {
 	generic_anim_unload(&ga);
 }
 streaminganim_h::streaminganim_h(streaminganim_h&& other) noexcept {
-	*this = std::move(other);
+	// Copy the other data over to us
+	ga = other.ga;
+
+	// Reset the other instance so that we own the only instance
+	generic_anim_init(&other.ga, nullptr);
 }
 streaminganim_h& streaminganim_h::operator=(streaminganim_h&& other) noexcept {
 	generic_anim_unload(&ga);
 
-	memcpy(&ga, &other.ga, sizeof(ga));
+	ga = other.ga;
 
 	// Reset the other instance so that we own the only instance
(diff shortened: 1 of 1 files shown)


asarium:
@Goober5000 `memcpy` is fine for the POD type but an assignment is a bit more expressive here so I changed it.

The `generic_anim_init` call is needed for making sure that only `this` "owns" the bitmap data formerly held by the other instance. If that call wasn't there then the destructor would try to unload the bitmap data even though it is still in use by the new instance.

Goober5000:
Expressivity is nice but what I meant was that if this is a move constructor, do we actually need a full-on copy?  `std::move` seems like it should be the right approach.

The init explanation helps clarify things.  With that in mind, could the function be rewritten as a `std::move` followed by a `generic_anim_init`?

AxemP:
Can confirm this fixes the issue!

asarium:
@Goober5000 `std::move` does nothing for POD type, it still copies the entire structure. Using `std::move` would even be counter-intuitive here since it would suggest that the structure actually could be moved even though it doesn't do anything for it.

niffiwan:
once again ignore the failed appveyor/branch test

asarium:
@niffiwan I think I fixed the appveyor status now. There was a wrong check listed in the protected branches list so GitHub was always waiting for that to be reported. That shouldn't happen any more.

PR #1946, opened 2018-12-07
Make particle rendering parallel-processed

TheMatthew:
Use OpenMP to parallelize particle render all. As particles are
not Z-sorted, this should accellerate scenes with lots of flack
on multi-core systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,13 @@ IF(MSVC_USE_RUNTIME_DLL)
 	INCLUDE(InstallRequiredSystemLibraries)
 ENDIF(MSVC_USE_RUNTIME_DLL)
 
+find_package(OpenMP)
+if (OPENMP_FOUND)
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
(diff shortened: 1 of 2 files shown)


TheMatthew:
Performance gains in bp2-000 are 3%.

In synthetic benchmark, gains can reach 20% on quad core cpu.

FS2 usage went from 25% to 43%.

Attached is test.

TheMatthew:
[flakattack.zip](https://github.com/scp-fs2open/fs2open.github.com/files/2659078/flakattack.zip)

TheMatthew:
This is much less efficient than if the code were written in tbb for instance but it is also 7 lines long. :)

SamuelCho:
I don't think this is going to work the way you intended it to work. You parallelized those loops but haven't done anything to eliminate the race conditions inherent in the particle rendering/batching code. The batching code writes to a vector and a map. Those two data structures aren't thread safe. 

Multithreading in general is something we've been very cautious about in FSO. If we ever do implement any sort of multithreading, it should be something only more senior programmers should be handling as they have a better understanding of the architectural internals of the game engine.

TheMatthew:
Hi Samuel, 

Yeah, I found that race while re-reading the code, a mutex there fixed that. I understand the caution and tbh, I am very patient when it comes to these things. I am having fun w/ the code base at this point and wanting to share my work... Merge it or don't it's fine. I think I do have a precedent though of listening to others.

 Please consider this pr to not be a pure pr but a proof of concept that may eventually make the engine speedier. :)

PR #1948, opened 2018-12-13
Don't rely on Assert to check for a valid AI class in SEXP code.

MageKing17:
Bad data can cause the name search to fail, and if `change-ai-class` was called with an invalid AI class, it could cause later invalid array accesses (using an index of `-1`).


asarium:
I'm not sure but I think three approvals should be barely enough to merge this...

PR #1949, opened 2018-12-15
simple optimization for obj_sort_and_collide

Goober5000:
This PR is based on some Discord discussion.  Apparently this is a known optimization but nobody has bothered to submit a PR yet. :p

From comments by @TRBlount:

> Turns out
if you make the vectors in that function static
that function goes from 1.5ms
to 0.2ms

> by making the vectors static, you no longer have to worry about reallocating unless you add even MORE objects to the scene

> the speedup is basically taking advantage of the fact that the data pointer inside the vector isn't being moved around as often

### DIFF
--- a/code/object/objcollide.cpp
+++ b/code/object/objcollide.cpp
@@ -1043,8 +1043,8 @@ void obj_sort_and_collide()
 	if ( !(Game_detail_flags & DETAIL_FLAG_COLLISION) )
 		return;
 
-	SCP_vector<int> sort_list_y;
-	SCP_vector<int> sort_list_z;
+	static SCP_vector<int> sort_list_y;
+	static SCP_vector<int> sort_list_z;
 
 	sort_list_y.clear();
 	{
--- a/code/object/objcollide.cpp
+++ b/code/object/objcollide.cpp
@@ -1035,6 +1035,10 @@ void obj_collide_retime_cached_pairs(int checkdly)
 	}
 }
 
+// used only in obj_sort_and_collide()
+static SCP_vector<int> sort_list_y;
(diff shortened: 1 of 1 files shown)


code/object/objcollide.cpp
```diff
@@ -1043,8 +1043,8 @@ void obj_sort_and_collide()
 	if ( !(Game_detail_flags & DETAIL_FLAG_COLLISION) )
 		return;
 
-	SCP_vector<int> sort_list_y;
-	SCP_vector<int> sort_list_z;
+	static SCP_vector<int> sort_list_y;
+	static SCP_vector<int> sort_list_z;
```
asarium: These should be actual global variables to make sure that the "cost" of the static variables is clearly visible.

Goober5000: The "cost" doesn't need to be visible.  The cost is plainly apparent from the `static` keyword (and anyway, it is hardly a cost since it speeds up the function).

A much better reason for not making them global variables is that they are only used in the one function, so they should remain entirely within the scope of the one function.

asarium: The "cost" is the maintainability cost since it makes it harder to see where in a module there is global data.

Also, there definitely is a constant cost of static variables for every function call since the code needs to check if the variable has already been initialized and that needs to happen with an atomic instruction so the performance isn't the best. If this is moved to a "normal" static, global variable then there are no hidden costs.

MageKing17: FWIW I agree with @asarium here; the performance difference is small, but not nonexistent, and since it's per call the savings add up. If we can gain a few nanoseconds per function call "for free", why not?

PR #1952, opened 2018-12-21
remove unused functions A to C

CasparKielwein:
I removed a lot of unused functions and some variables in directories code/ai to code/cutscene.
It is part of an effort to be able to switch unused function and variable warnings back on.

This PR does not contain any functional change.
It is one commit per directory. I can squash them if you wish.


chief1983:
I don't have time to go through the entire list right now, but just in case there were any functions that we didn't want to remove because we may want them later, they can also be flagged with a preprocessor define that tells the warning check to ignore a particular function.  But I agree removing is preferred if we no longer have a use for them.

CasparKielwein:
I asked on discord and it seemed to me, that the main problem is the project being overwhelmed by the sheer amount of dead code.

If you are unsure about a particular file, I can take it out of the PR to have it move along with the rest.

My personal take on "we might need it later" is that if someone needs it and knows it was there, he can restore it from git, and if he doesn't know it is there, he isn't going to find it and probably reimplement it instead.

chief1983:
That's a completely fine take on it, and again I'm all for removing dead code.  The only place I knew of off the top of my head we had functions we might want later was in vector math I think, which I don't believe I see in your changes so to my knowledge there's nothing in there we'd be wanting to keep around.  However the only problem with relying on the git history for getting removed code is knowing it existed at some point in the first place and how far back you'd have to go to get it.  That becomes tribal knowledge that any new developer would not be aware of, and is why I would at least question whether something looks like it is likely to be used, or someone may end up rewriting it from scratch without realizing it existed before.

I did just scroll through all the changes, and I have to admit I'm kind of shocked at some of the functions that exist but aren't apparently in use anymore.  At least some I can imagine were simply obsoleted via refactors and got left around, but some of them just seem weird to exist and be unused, such as the camera ones.  That's post-Volition stuff as far as I know.

CasparKielwein:
I've created an Issue: #1955 for centralized discussion of the several related PRs.

code/camera/camera.cpp
```diff
@@ -145,7 +145,7 @@ void camera::set_object_target(object *objp, int n_object_target_submodel)
  * Custom function receives the already-modified current position value.
  * It should be replaced or added to as the custom function modifier sees fit.
  */
-void camera::set_custom_position_function(void (*n_func_custom_position)(camera *cam, vec3d *camera_pos))
```
asarium: For function pointer types it would be better to leave the parameter names in place to document what the parameters mean.

PR #1953, opened 2018-12-21
AI profile flag for support to not equip new primary

plieblang:
Implements [Mantis 3198](http://scp.indiegames.us/mantis/view.php?id=3198).
This does pretty much what the title says:  currently if a primary bank is unequipped, rearming will result the player having a new weapon.  Because this is intentional from retail, this PR adds an AI flag to disable that behavior.

### DIFF
--- a/code/ai/ai_flags.h
+++ b/code/ai/ai_flags.h
@@ -110,7 +110,7 @@ namespace AI {
         Smart_shield_management,
         Smart_subsystem_targeting_for_turrets,
         Strict_turred_tagged_only_targeting,
-		Support_dont_add_primaries,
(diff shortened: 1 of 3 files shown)


code/ai/ai_flags.h
```diff
@@ -110,6 +110,7 @@ namespace AI {
         Smart_shield_management,
         Smart_subsystem_targeting_for_turrets,
         Strict_turred_tagged_only_targeting,
+		Support_dont_add_primaries,
```
asarium: A small comment explaining what this flag would be useful here. Maybe with the attached mantis issue number if someone wants to look up the reason for having this feature.

Issue #1955, opened 2018-12-22
Remove unused Functions


CasparKielwein:
FSO contains a large number of unused functions and variable.

Over time the amount of unused code has grown so large, that compiler warnings needed to be disabled.
This situation is obviously unsustainable.

I've gone through the codebase and removed a lot of obviously unused code and assembled a few PRs.
My workflow has been:
* Run clang tidy to check for unused functions.
* For each function it warns, grep in entire codebase
* If there are no results except definition and/or implementation -> remove function.
* The same for some obvious global variables and defines.
* Ignore scripting related code as the lua bindings make checking for unused functions non-trivial
* ignore library code like lib/ ,vector math, 3dmath.

Since large changes like these are usually hard to review, I suggest the following workflow:
Reviewers have a look and if they are unsure about a particular function or file, I remove it from the PR in an effort to have the noncontroversial changes merged as fast as possible.
I will then later make another PR with the more controversial changes to discuss them.

Goober5000:
`cmeasure_select_next` was an important part of WMC's countermeasure overhaul - the ability to select multiple types of countermeasures.  Unfortunately WMC hasn't been around for ages.

CasparKielwein:
We could tag the last commit before it was removed, in case someone wants to pick up where he left off.

After #1958 is closed, there is one warning left:
`qtfred/src/mission/FredRenderer.cpp:156:5: warning: ‘int {anonymous}::get_subsys_bounding_rect(object*, ship_subsys*, int*, int*, int*, int*)’ defined but not used`

I haven't touched qtfred.

PR #1956, opened 2018-12-22
remove unused Functions D to H

CasparKielwein:
Removes unused code in directories starting with Letters D to H, as described in #1955 .

Pretty much analogue to #1952


code/debugconsole/console.h
```diff
@@ -119,27 +119,6 @@ class debug_command;
...
 	}
 
- /**
-  * @def Shortcut for single-variable setters/monitors
-  *
-  * @param [in] function_name
-  * @param [in] float_variable
-  * @param [in] short_help
-  */
- #define DCF_FLOAT(function_name, float_variable, short_help)	\
```
asarium: This should not be removed since it might be useful later.

code/debugconsole/console.h
```diff
@@ -164,27 +143,6 @@ class debug_command;
...
 	}
 
- /**
-  * @def Shortcut for single-variable setters/monitors
-  *
-  * @param [in] function_name
-  * @param [in] int_variable
-  * @param [in] short_help
-  */
- #define DCF_INT(function_name, int_variable, short_help)	\
```
asarium: This should not be removed since it might be useful later.

code/debugconsole/consoleparse.cpp
```diff
@@ -456,105 +450,6 @@ bool dc_maybe_stuff_string_white(SCP_string &str)
 	}
 }
 
-void dc_required_string(char *pstr)
```
asarium: These are also library functions so they should not be removed.

code/debugconsole/console.h
CasparKielwein: done

code/debugconsole/consoleparse.cpp
CasparKielwein: done

Goober5000:
Aside from what @asarium mentioned, I have no objection.

It's interesting from a historical standpoint to see `hud_target_closest_locked_missile`.

PR #1957, opened 2018-12-22
remove unused functions I to O

CasparKielwein:
Removes unused code in directories starting with Letters I to O, as described in #1955 .

Pretty much analogue to #1952


code/io/joy-sdl.cpp
```diff
@@ -352,25 +352,13 @@ namespace joystick
 		return *this;
 	}
 
-	bool Joystick::isAttached() const
```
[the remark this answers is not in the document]
CasparKielwein: okely dokely

code/mission/missionbriefcommon.cpp
```diff
@@ -1819,50 +1818,6 @@ void brief_set_new_stage(vec3d *pos, matrix *orient, int time, int stage_num)
...
-		return TRUE;
-	
-	return FALSE;
-}
-
-/**
- * Interpolate between matrices.
- * elapsed_time/total_time gives percentage of interpolation between cur and goal.
- */
-void interpolate_matrix(matrix *result, matrix *goal, matrix *start, float elapsed_time, float total_time)
```
Goober5000: Keep this function.  Move it to vecmat.

code/model/modelread.cpp
```diff
@@ -4968,14 +4843,6 @@ void model_do_children_look_at(polymodel * pm, int mn)
 	}
 }
 
-void model_do_look_at(int pn)
```
Goober5000: Keep this function.

code/mission/missionbriefcommon.cpp
CasparKielwein: Are you sure you want to keep it?
The implementation doesn't strike me as remarkable and the interface has problems.

Also it would be pretty trivial to reimplement if necessary.

Goober5000: I'm not super attached to it, but it strikes me as very useful for physics, AI, and rotation code.  A good library function, in other words.  How is the interface problematic?

CasparKielwein: Having two floats as input parameters instead of one float which is checked to be in [0, 1] is error prone.  
When goal and start are equal, nothing happens instead of the result being set to either one.

This seems a bit strange for a function which is effectively C = f * (A+B).

It probably made sense for the special purpose it was originally written for.

Any way: You call if you want to keep it, I wouldn't add it to vecmat in this PR though.

Goober5000: On second thought, if it's in missionbriefcommon.cpp, it probably really is irrelevant.  Go ahead and remove it.

code/model/modelread.cpp
CasparKielwein: done

PR #1960, opened 2018-12-26
Add Lua state destruction event

### LINKED ISSUE #1954: Read Access Violation on Exit (Lua related)
In a mod that uses scripts (like JAD), on quitting in release there sometimes might be a crash on exit. But in fastdebug or full debug, there's a straight CTD or read access violation just on clicking the exit button.

This is the error message
```
Exception thrown: read access violation.
**o** was 0xFFFFFFFFFFFFFFFF. occurred
```

And here is the call stack
```
>	fs2_open_3_8_1_x64_AVX-DEBUG.exe!lua_rawgeti(lua_State * L, int idx, int n) Line 572	C
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!luaL_unref(lua_State * L, int t, int ref) Line 508	C
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!luacpp::UniqueLuaReference::removeReference() Line 54	C++
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!luacpp::UniqueLuaReference::~UniqueLuaReference() Line 48	C++
 	[External Code]	
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!luacpp::LuaFunction::~LuaFunction() Line 52	C++
 	[External Code]	
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!sexp::dynamic_sexp_shutdown() Line 125	C++
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!sexp_shutdown() Line 1129	C++
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!game_shutdown() Line 6926	C++
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!game_main(int argc, char * * argv) Line 6813	C++
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!actual_main(int argc, char * * argv) Line 7987	C++
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!SDL_main(int argc, char * * argv) Line 8033	C++
 	fs2_open_3_8_1_x64_AVX-DEBUG.exe!main_getcmdline(...) Line 161	C
 	[External Code]	
```

@asarium I hope you've got some killer insights into this. :)

asarium:
This can be used by the various subsystems for cleaning up Lua
references before the Lua state is destroyed. This avoids having
dangling references to the Lua state which would cause segfaults later
on.

The first subsystem to use this is the dynamic SEXP system which had
dangling references due to keeping Lua references to the SEXP functions.

This fixes #1954.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -47,6 +47,7 @@ Cpp11BracedListStyle: true
 DerivePointerAlignment: false
 DisableFormat:   false
 ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
+IncludeBlocks: Regroup
 IncludeCategories:
   - Regex:           '^(<.*\.(h|hh|hpp)>)$' # Includes of external libraries
     Priority:        4
--- a/code/parse/sexp/sexp_lookup.cpp
+++ b/code/parse/sexp/sexp_lookup.cpp
@@ -1,9 +1,10 @@
 
 #include "parse/sexp/sexp_lookup.h"
-#include "parse/sexp/LuaSEXP.h"
 
-#include "parse/sexp.h"
 #include "parse/parselo.h"
+#include "parse/sexp.h"
+#include "parse/sexp/LuaSEXP.h"
+#include "scripting/scripting.h"
 
 #include <memory>
 
@@ -82,7 +83,21 @@ void parse_sexp_table(const char* filename) {
 	}
 }
 
+void free_lua_sexps(lua_State* /*L*/)
+{
+	// Remove all Lua sexps from our list so that there are no dangling pointers on the lua state
+	for (auto iter = operator_const_mapping.begin(); iter != operator_const_mapping.end();) {
+		auto lua_sexp = dynamic_cast<LuaSEXP*>(iter->second.get());
+		if (lua_sexp == nullptr) {
+			++iter;
+			continue;
+		}
+
+		iter = operator_const_mapping.erase(iter);
+	}
 }
+
+} // namespace
 
 namespace sexp {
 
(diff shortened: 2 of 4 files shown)


.clang-format
```diff
@@ -47,6 +47,7 @@ Cpp11BracedListStyle: true
 DerivePointerAlignment: false
 DisableFormat:   false
 ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
+IncludeBlocks: Regroup
```
Goober5000: Did you intend to commit clang changes?

code/parse/sexp/sexp_lookup.cpp
```diff
@@ -82,8 +83,22 @@ void parse_sexp_table(const char* filename) {
 	}
 }
 
+void free_lua_sexps(lua_State* /*L*/)
+{
+	// Remove all Lua sexps from our list so that there are no dangling pointers on the lua state
+	for (auto iter = operator_const_mapping.begin(); iter != operator_const_mapping.end();) {
```
Goober5000: I would have made this a `while` loop, but that's nitpicking.  I had to do a double-take and reread it because an iterator increment followed by a continue inside a `for` loop is usually a bug.  (I can see that it's not a bug in this case.)

.clang-format
asarium: Yes, this should reflect how the existing includes are handled a bit more closely.

code/parse/sexp/sexp_lookup.cpp
asarium: Unfortunately, there is no `remove_if` for maps so there isn't a better way for expressing this yet.

PR #1958, opened 2018-12-22
remove unused functions O to R

CasparKielwein:
Removes unused code in directories starting with Letters O to R, as described in #1955 .

Pretty much analogue to #1952

PR #1963, opened 2018-12-31
enable Warnings for unused functions

### LINKED ISSUE #1955: Remove unused Functions

CasparKielwein:
Enable compiler warnings for unused functions in gcc and clang toolchain files.

fixes #1955

### DIFF
--- a/cmake/toolchain-apple-clang.cmake
+++ b/cmake/toolchain-apple-clang.cmake
@@ -9,8 +9,6 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -funroll-loops -fsigned-char -Wno-
 # Omit "conversion from string literal to 'char *' is deprecated" warnings.
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-write-strings")
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-function")
-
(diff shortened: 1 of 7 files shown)


asarium:
I think you need to rebase this to get the changes from #1958 into this.

CasparKielwein:
Yeah, I'll rebase it, and there is one unused function in qtfred left, which needs to be removed or marked with a pragma.

asarium:
The `FredRenderer` code in qtFRED was mostly copied from the FRED2 code. If you want to fix the warning you can just delete both (I have no idea why there are two...) definitions.

Issue #1835, opened 2018-08-19
Enhancement to control warp type in FRED


MjnMixael:
Currently, if we want a ship to use "hyperspace" or other warp type in place of the default warp behavior, we have to create a duplicate ship in the ships.tbl just to change that one setting. I propose we enhance this so that we can choose the warp type to use in FRED either via checkbox in the ships editor or via sexp or both. This way any ship can use any warp type and it becomes a mission design tool instead of, essentially, a ship specific feature.

Goober5000:
You could make the same argument for any parameter that's controlled in ships.tbl.  I think this would be more trouble than it's worth, since changing any ship class attribute is a matter of copy+paste+edit.

I speak from experience, since I was the one that converted the "stealth" flag to also be a ship flag.  Also, FSPort and Scroll both use ship copies for certain cases.  Scroll has a GVD Hatshepsut#FastWarpDecel that's only used in one place in one mission.

PR #1964, opened 2019-01-06
create WarpDirection enum

Goober5000:
I'm doing a bit of refactoring of the warp code for issue #1835 and I think it would be good to get this enum out of the way as a separate PR first.

### DIFF
--- a/code/ship/shipfx.h
+++ b/code/ship/shipfx.h
@@ -122,7 +122,7 @@ void shipfx_engine_wash_level_init();
 // pause engine wash sounds
 void shipfx_stop_engine_wash_sound();
 
-enum class WarpDirection { WD_NONE, WD_WARP_IN, WD_WARP_OUT };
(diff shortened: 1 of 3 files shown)


code/ship/shipfx.h
```diff
@@ -122,25 +122,24 @@ void shipfx_engine_wash_level_init();
 // pause engine wash sounds
 void shipfx_stop_engine_wash_sound();
 
-#define WD_NONE		0
-#define WD_WARP_IN	1
-#define WD_WARP_OUT	2
-float shipfx_calculate_warp_time(object *objp, ship_info *sip, int warp_dir, float half_length, float warping_dist);
+enum class WarpDirection { WD_NONE, WD_WARP_IN, WD_WARP_OUT };
```
asarium: Since the enum class has its own scope the WD_ prefix isn't needed anymore and could be removed to make this easier to read.

Goober5000: Mmm, true.

PR #1966, opened 2019-01-08
per-ship warp type

Goober5000:
This is a new feature that allows overriding warp parameters on a per-ship basis in addition to a per-class basis.  It implements the request from issue #1835.

Instead of bloating the structs with yet more fields, this refactors them into the WarpParams object so that ships, ship classes, and parse objects can simply store an index into a vector.  Ships, etc. that have identical parameters will re-use indexes, which simplifies things and saves memory.

This is also partially dependent on PR #1964, so I will rebase onto master after that has been approved.


code/ship/ship.cpp
```diff
@@ -2281,6 +2204,136 @@ static int parse_and_add_briefing_icon_info()
...
+	sprintf(str, "%s animation:", prefix);
+	if (optional_string(str))
+	{
+		stuff_string(params.anim, F_NAME, MAX_FILENAME_LEN);
+	}
+
+	if (direction == WarpDirection::WARP_OUT)
+	{
+		sprintf(str, "$Player warpout speed:");
+		if (optional_string(str))
```
MageKing17: This reads awkward as hell; would just using plain text for this one option really be that bad?

fred2/fred.rc
```diff
@@ -938,15 +938,15 @@ BEGIN
     GROUPBOX        "Default Display",IDC_DISPLAY,107,7,136,56
 END
 
-IDD_SHIP_EDITOR DIALOGEX 0, 0, 321, 422
+IDD_SHIP_EDITOR DIALOGEX 0, 0, 322, 478
 STYLE DS_SETFONT | DS_MODALFRAME | DS_3DLOOK | DS_CENTER | WS_POPUP | WS_CAPTION | WS_SYSMENU
 EXSTYLE WS_EX_NOPARENTNOTIFY | WS_EX_CLIENTEDGE | WS_EX_CONTEXTHELP
-CAPTION "Edit Ship"
+CAPTION " "
```
MageKing17: Was this intentional?

code/fireball/fireballs.h
```diff
@@ -106,7 +106,7 @@ int fireball_info_lookup(const char *unique_id);
 // reversed is for warp_in/out effects
```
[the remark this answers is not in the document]
Goober5000: Ok.

code/ship/ship.cpp
Goober5000: It's a little silly, but I did that for consistency.  Not just in that one line, but more importantly for the `Warning` a few lines later.  That is the same for each field (except for `accel_exp` which allows 0) and requires the `str` parameter.

fred2/fred.rc
Goober5000: Nope.  I have no idea how that happened.

PR #1967, opened 2019-01-08
Avoid using Asserts for data security in modelread properties-reading code.

MageKing17:
In a Release build, there was absolutely no code making sure that the text "$glow_texture=" was actually present in the properties field of a glow point bank, beyond a simple `strlen()` check. Even in Debug builds, there was no guarantee that the text was at the start of the properties field; it was just assumed (without even an assertion verifying it). Now, the assertions have been removed and the appropriate conditions are checked in both Release and Debug.

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -1956,7 +1956,7 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
 					{
 						auto base_length = strlen("$glow_texture=");
 						char *glow_texture_start = strstr(props, "$glow_texture=");
-						if ( (length > base_length) && (glow_texture_start != nullptr) && (strlen(glow_texture_start + base_length) > 0) ) {
+						if ( (glow_texture_start != nullptr) && (strlen(glow_texture_start + base_length) > 0) ) {
 							char *glow_texture_name = glow_texture_start + base_length;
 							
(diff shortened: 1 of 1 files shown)


code/model/modelread.cpp
```diff
@@ -1953,44 +1955,44 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
...
 					{
 						auto base_length = strlen("$glow_texture=");
-						Assert(strstr( (const char *)&props, "$glow_texture=") != NULL);
-						Assert(length > base_length);
-						char *glow_texture_name = props + base_length;
-						
-						if (glow_texture_name[0] == '$')
-							glow_texture_name++;
+						char *glow_texture_start = strstr(props, "$glow_texture=");
+						if ( (length > base_length) && (glow_texture_start != nullptr) && (strlen(glow_texture_start + base_length) > 0) ) {
```
Goober5000: I think the `(length > base_length)` check is redundant, given the other conditions.

MageKing17: It is redundant; I just carried it over from the old code because I wasn't thinking straight.

MageKing17:
While I was removing the redundant `(length > base_length)` check, I realized the code in the `ID_FUEL` case below was making the same mistake, so I fixed that, too.

Goober5000:
Good stuff. :+1:

PR #1984, opened 2019-01-30
fix some flag information

Goober5000:
Somewhere along the line, "sentry gun" was replaced with "escape pod" in the inverted is_flyable check.  This made the game think an escape pod wasn't flyable and therefore things like waypoint orders were not displayed.  This PR fixes that.

Additionally, I fixed a potential bug in weapons.cpp where the HUGE flag should really check against HUGE ships.  And there's a no-op formatting change in hudtargetbox.cpp.

### DIFF
--- a/code/ship/ship.h
+++ b/code/ship/ship.h
@@ -1214,8 +1214,8 @@ public:
     inline bool is_small_ship() const { return flags[Ship::Info_Flags::Fighter, Ship::Info_Flags::Bomber, Ship::Info_Flags::Support, Ship::Info_Flags::Escapepod]; }
     inline bool is_big_ship() const { return flags[Ship::Info_Flags::Cruiser, Ship::Info_Flags::Freighter, Ship::Info_Flags::Transport, Ship::Info_Flags::Corvette, Ship::Info_Flags::Gas_miner, Ship::Info_Flags::Awacs]; }
     inline bool is_huge_ship() const  { return flags[Ship::Info_Flags::Capital, Ship::Info_Flags::Supercap, Ship::Info_Flags::Drydock, Ship::Info_Flags::Knossos_device]; }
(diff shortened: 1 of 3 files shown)


code/ship/ship.h
```diff
@@ -1214,8 +1214,8 @@ class ship_info
     inline bool is_small_ship() const { return flags[Ship::Info_Flags::Fighter, Ship::Info_Flags::Bomber, Ship::Info_Flags::Support, Ship::Info_Flags::Escapepod]; }
     inline bool is_big_ship() const { return flags[Ship::Info_Flags::Cruiser, Ship::Info_Flags::Freighter, Ship::Info_Flags::Transport, Ship::Info_Flags::Corvette, Ship::Info_Flags::Gas_miner, Ship::Info_Flags::Awacs]; }
     inline bool is_huge_ship() const  { return flags[Ship::Info_Flags::Capital, Ship::Info_Flags::Supercap, Ship::Info_Flags::Drydock, Ship::Info_Flags::Knossos_device]; }
-    inline bool is_flyable() const { return !(flags[Ship::Info_Flags::Cargo, Ship::Info_Flags::Navbuoy, Ship::Info_Flags::Escapepod]); }
-    inline bool is_harmless() const { return !is_flyable(); }
+    inline bool is_flyable() const { return !(flags[Ship::Info_Flags::Cargo, Ship::Info_Flags::Navbuoy, Ship::Info_Flags::Sentrygun]); }
+    inline bool is_harmless() const { return flags[Ship::Info_Flags::Cargo, Ship::Info_Flags::Navbuoy, Ship::Info_Flags::Escapepod]; }	// note: code which previously used this now uses several flags defined in objecttypes.tbl
```
MageKing17: Since this is no longer used, it should just be removed.

Goober5000: I think if you're going to document it you might as well have it in code form, but ok.

PR #1985, opened 2019-02-03
Add 'On Waypoints Done' Hook

wookieejedi:
Adds 'On Waypoints Done' hook variable for scripting system.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -58,6 +58,8 @@
 #include "physics/physics.h"
 #include "playerman/player.h"
 #include "render/3d.h"
+#include "scripting/api/objs/wing.h"
+#include "scripting/scripting.h"
 #include "ship/afterburner.h"
 #include "ship/awacs.h"
 #include "ship/ship.h"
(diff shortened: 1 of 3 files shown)


code/ai/aicode.cpp
```diff
@@ -4586,6 +4588,11 @@ void ai_waypoints()
 					ai_mission_wing_goal_complete( Ships[Pl_objp->instance].wingnum, &(aip->goals[aip->active_goal]) );
 					mission_log_add_entry( LOG_WAYPOINTS_DONE, Wings[Ships[Pl_objp->instance].wingnum].name, aip->wp_list->get_name(), -1 );
 				}
+				// adds scripting hook for 'On Waypoints Done' --wookieejedi
+				Script_system.SetHookObject("Ship", &Objects[Ships[Pl_objp->instance].objnum]);
+				Script_system.SetHookVar("Wing", 'o', scripting::api::l_Wing.Set(Ships[Pl_objp->instance].wingnum));
```
MageKing17: No other call to set a hook variable or variables directly references the scripting API, and *should* not if we don't want to break the fundamental assumption that Lua isn't the only potential scripting language that could be used by FSO.

asarium: 1. It doesn't break that assumption since the handle types are mostly free of anything Lua specific and even most scripting API functions could be used by other scripting languages if the necessary functions would be implemented for them.
1. That assumption has already been broken in other places which also made it much easier to write new functionality since you don't need to consider a hypothetical scripting language as a potential target.

PR #1988, opened 2019-02-16
make is-docked work with a single argument

niffiwan:
Currently is-docked always returns true if only a single argument is
supplied. This is probably as designed (except for not erroring on a
single arg) however it's quite easy to update to work for a single
argument and this case is useful for FotG, and might be useful for
others.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -22692,9 +22692,13 @@ int sexp_is_docked(int node)
 
 		current_shipp = &Ships[shipnum];
 
-		// if we're currently handling the host ship, this is all we need to do for this loop iteration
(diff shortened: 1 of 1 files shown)


niffiwan:
Here's the test mission I used: [isdocked.txt](https://github.com/scp-fs2open/fs2open.github.com/files/2871665/isdocked.txt)

niffiwan:
thanks, great suggestion to simplify this :)

Goober5000:
The change in code is good, but the new comment muddles the situation IMO. I have two suggestions:

1) Change the comment to something like "if host_shipp is NULL, we are on the first loop iteration"
2) Add a comment to the new if() condition like "if the host isn't docked to anything, no need to check each ship individually"

niffiwan:
fair enough - done! :)

PR #1990, opened 2019-02-21
Compile curl ourselves on macOS and cache the result

ngld:
We need to build our own `curl` binary since Homebrew doesn't support `--with-libssh2` anymore.
I've also added caching to both the Linux and macOS code to avoid rebuilding curl for each nightly.
Not sure if the caching actually works but according to [the docs](https://docs.travis-ci.com/user/caching/) it should work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ cache:
   directories:
     - $HOME/cmake-3.8
     - $HOME/.ccache
+    - $HOME/curl_cache
 addons:
   apt:
     packages: &global_deps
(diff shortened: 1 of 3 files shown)


ngld:
Quick note before anyone merges this:
*Don't* merge the second commit. It's supposed to test the upload through Travis CI (and should never end up in `master`) but that didn't work. Looks like Travis just skips the `deploy` section for PRs. Will try again tomorrow.

asarium:
If you enable Travis on your own repository you should be able to use the test mode if you push something to a branch. The encrypted data will still not work since you don't have to correct keys but you can check everything up to the upload part.

chief1983:
Would switching to an alternative method than curl make more sense here?  It seemed like there were other options we discussed on Discord that would have been simpler than manually compiling another library, but maybe they had their own flaws.

asarium:
`scp` may be an option but we would need to set up a SSH key on Travis and the server for that to work without entering a password.

chief1983:
Works for me then, Travis is probably going away anyway :P

PR #1991, opened 2019-02-23
Fix issues with docked ships warping

niffiwan:
There's a number of issues fixed here:
* Radar blips of docked ships were the wrong colour in stage 1
* You could target docked ships in stage 1
* Docked ships thrusters were visible in stage 1
* Hyperspace warptype ship could initially appear at their
warp-destination, then spring back to the correct location to complete
the rest of their warpin (because a docked ship had higher speed than
the dock leader)

There may also be a number of other subtle bugs with docked warpin;
e.g. I suspect (but haven't tested) that AI turrets could fire on
docked ships during warpin
These should also be fixed

Lastly, fixes issue #1867

### DIFF
--- a/code/object/objectdock.cpp
+++ b/code/object/objectdock.cpp
@@ -711,22 +711,26 @@ void dock_find_max_speed_helper(object *objp, dock_function_info *infop)
 
 void object_set_arriving_stage1_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
 {
-	Ships[objp->instance].flags.set(Ship::Ship_Flags::Arriving_stage_1_ndl);
+	if (! Ships[objp->instance].flags[Ship::Ship_Flags::Dock_leader])
+		Ships[objp->instance].flags.set(Ship::Ship_Flags::Arriving_stage_1_dock_follower);
 }
 
 void object_remove_arriving_stage1_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
 {
-	Ships[objp->instance].flags.remove(Ship::Ship_Flags::Arriving_stage_1_ndl);
+	if (! Ships[objp->instance].flags[Ship::Ship_Flags::Dock_leader])
+		Ships[objp->instance].flags.remove(Ship::Ship_Flags::Arriving_stage_1_dock_follower);
 }
 
 void object_set_arriving_stage2_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
 {
-	Ships[objp->instance].flags.set(Ship::Ship_Flags::Arriving_stage_2_ndl);
+	if (! Ships[objp->instance].flags[Ship::Ship_Flags::Dock_leader])
+		Ships[objp->instance].flags.set(Ship::Ship_Flags::Arriving_stage_2_dock_follower);
 }
 
 void object_remove_arriving_stage2_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
 {
-	Ships[objp->instance].flags.remove(Ship::Ship_Flags::Arriving_stage_2_ndl);
+	if (! Ships[objp->instance].flags[Ship::Ship_Flags::Dock_leader])
+		Ships[objp->instance].flags.remove(Ship::Ship_Flags::Arriving_stage_2_dock_follower);
 }
 
 // ---------------------------------------------------------------------------------------------------------------
--- a/code/object/objectdock.cpp
+++ b/code/object/objectdock.cpp
@@ -708,6 +708,27 @@ void dock_find_max_speed_helper(object *objp, dock_function_info *infop)
 		infop->maintained_variables.objp_value = objp;
 	}
 }
+
+void object_set_arriving_stage1_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
+{
+	Ships[objp->instance].flags.set(Ship::Ship_Flags::Arriving_stage_1_ndl);
+}
+
+void object_remove_arriving_stage1_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
+{
+	Ships[objp->instance].flags.remove(Ship::Ship_Flags::Arriving_stage_1_ndl);
+}
+
+void object_set_arriving_stage2_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
+{
+	Ships[objp->instance].flags.set(Ship::Ship_Flags::Arriving_stage_2_ndl);
+}
+
+void object_remove_arriving_stage2_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
+{
+	Ships[objp->instance].flags.remove(Ship::Ship_Flags::Arriving_stage_2_ndl);
+}
+
 // ---------------------------------------------------------------------------------------------------------------
 // end of über code block ----------------------------------------------------------------------------------------
 
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -18990,20 +18990,20 @@ ship_subsys* ship_get_subsys_for_submodel(ship* shipp, int submodel)
 /**
  * Is the requested type of ship arriving in the requested stage?
  *
- * Dock leaders & followers need to be handled separately such that
- * leaders handle the warp process, but followers need to behave as though
- * warping with respect to targeting, drawing thrusters, radar, AI
- * targeting, etc
+ * Dock leaders (plus single ships) & dock followers need to be handled
+ * separately such that leaders+singles handle the warp process, but
+ * followers need to behave as though they are warping with respect to
+ * HUD targeting, drawing thrusters, radar, AI targeting, etc
  *
  * @param stage Check one of stage1, stage2 or both
- * @param dock_leader_only only the dock leader, or leader/followers/non-docked?
+ * @param dock_leader_or_single_only check dock-leaders+singles, or dock-followers?
  *
  * @return is the ship arriving, bool
  */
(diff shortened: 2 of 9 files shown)


code/ship/ship.cpp
```diff
@@ -6718,7 +6718,8 @@ static void ship_find_warping_ship_helper(object *objp, dock_function_info *info
 		return;
 
 	// am I arriving or departing by warp?
-	if (Ships[objp->instance].is_arriving() || Ships[objp->instance].flags[Ship_Flags::Depart_warp])
+	// ignore _ndl flags as they are now on all docked-warping ships; we need the dock leader!
+	if (Ships[objp->instance].flags[Ship_Flags::Arriving_stage_1, Ship_Flags::Arriving_stage_2, Ship_Flags::Depart_warp])
```
MageKing17: Couldn't the same thing be done by adding an argument to `is_arriving()`, rather than by hardcoding this list of flags every time it's needed?

niffiwan: fair call; I'll see about updating that on the weekend. How do you feel about a separate inline function rather than an tri-state argument to `is_arriving()`? e.g. `is_arriving_dock_leader()` & `is_arriving_stage_1_all_docked()`

MageKing17: As long as we're not using a hardcoded flag list in multiple places, I don't particularly mind implementation.

Goober5000:
I will take a look at this in the next day or two.

wookieejedi:
Glad to see this one finally got a solution!

Goober5000:
Stand by, still planning to look at this >.>

niffiwan:
something something standing by :)

code/object/objectdock.cpp
```diff
@@ -708,6 +708,27 @@ void dock_find_max_speed_helper(object *objp, dock_function_info *infop)
 		infop->maintained_variables.objp_value = objp;
 	}
 }
+
+void object_set_arriving_stage1_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
+{
+	Ships[objp->instance].flags.set(Ship::Ship_Flags::Arriving_stage_1_ndl);
```
Goober5000: This will set the flag on all ships in the docked group.  Do you want to assign it to all ships *except* the leader?  If so, add a check for the ship with the leader flag.

code/object/objectdock.cpp
```diff
@@ -708,6 +708,27 @@ void dock_find_max_speed_helper(object *objp, dock_function_info *infop)
...
+}
+
+void object_remove_arriving_stage1_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
+{
+	Ships[objp->instance].flags.remove(Ship::Ship_Flags::Arriving_stage_1_ndl);
+}
+
+void object_set_arriving_stage2_ndl_flag_helper(object *objp, dock_function_info * /*infop*/ )
+{
+	Ships[objp->instance].flags.set(Ship::Ship_Flags::Arriving_stage_2_ndl);
```
Goober5000: Same as for stage 1, above

Goober5000:
Apologies for the delay, and thanks for accommodating it. :)  Overall, an excellent solution, but I have several comments.

code/ship/ship_flags.h
```diff
@@ -57,7 +57,9 @@ namespace Ship {
 		Depart_warp,				// ship is departing via warp-out
 		Depart_dockbay,				// ship is departing via docking bay
 		Arriving_stage_1,			// ship is arriving. In other words, doing warp in effect, stage 1
+		Arriving_stage_1_ndl,		// "Arriving but Not the Dock Leader" (ndl) these guys need some warp stuff done but not all
```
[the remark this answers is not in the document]
niffiwan: sure

code/ship/ship_flags.h
```diff
@@ -57,7 +57,9 @@ namespace Ship {
...
 		Arriving_stage_2,			// ship is arriving. In other words, doing warp in effect, stage 2             
+		Arriving_stage_2_ndl,		// "Arriving but Not the Dock Leader" (ndl) these guys need some warp stuff done but not all
```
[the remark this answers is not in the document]
niffiwan: sure

code/object/objectdock.cpp
niffiwan: I considered excluding the leader and thought it doesn't actually matter. However as you've raised above it's probably _clearer_ to the coder if it's excluded; and that's worthwhile so I'll change it.

code/object/objectdock.cpp
niffiwan: ditto for stage 1

code/ship/ship.h
```diff
@@ -714,7 +714,9 @@ class ship
 	void clear();
 
     //Helper functions
-    inline bool is_arriving() { return flags[Ship::Ship_Flags::Arriving_stage_1, Ship::Ship_Flags::Arriving_stage_2]; }
+    inline bool is_arriving() { return flags[Ship::Ship_Flags::Arriving_stage_1, Ship::Ship_Flags::Arriving_stage_1_ndl, Ship::Ship_Flags::Arriving_stage_2, Ship::Ship_Flags::Arriving_stage_2_ndl]; }
+    inline bool is_arriving_dock_leader() { return flags[Ship::Ship_Flags::Arriving_stage_1, Ship::Ship_Flags::Arriving_stage_2]; }
+    inline bool is_arriving_stage_1_all_docked() { return flags[Ship::Ship_Flags::Arriving_stage_1, Ship::Ship_Flags::Arriving_stage_1_ndl]; }
```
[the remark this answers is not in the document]
niffiwan: I'm happy to rename them, however I just want to confirm that I understand!

`is_arriving_actively()` means two flags; both stage 1 & 2 but not the followers
`is_arriving_dock_follower()` means stage one only; but everyone
`is_arriving()` means four flags; everyone in both stages

thanks for the review!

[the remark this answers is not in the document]
Goober5000: Uh... rats, I didn't even consider the distinction between stage 1 and 2.

I think it really is necessary to make this a function with parameters rather than an inline function.  One parameter should be stage1/stage2/both, and one parameter should be dock-leader-or-solo/dock-follower/dock-agnostic.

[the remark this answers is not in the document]
niffiwan: done, could you please let me know what you think?

code/ship/ship.cpp
```diff
@@ -18985,3 +18986,48 @@ ship_subsys* ship_get_subsys_for_submodel(ship* shipp, int submodel)
...
+ * leaders handle the warp process, but followers need to behave as though
+ * warping with respect to targeting, drawing thrusters, radar, AI
+ * targeting, etc
+ *
+ * @param stage Check one of stage1, stage2 or both
+ * @param dock_leader_only only the dock leader, or leader/followers/non-docked?
+ *
+ * @return is the ship arriving, bool
+ */
+bool ship::is_arriving(ship::warpstage stage, bool dock_leader_only)
```
Goober5000: This looks a lot better; I just didn't realize it would be so complicated!

code/ship/ship.cpp
```diff
@@ -18985,3 +18986,48 @@ ship_subsys* ship_get_subsys_for_submodel(ship* shipp, int submodel)
...
+{
+	if (stage == ship::warpstage::BOTH) {
+		if (dock_leader_only == false) {
```
Goober5000: If this is going to be a boolean, it should check the variable directly (with ! if necessary) not with a ==.

code/ship/ship.cpp
Goober5000: There remains one small area of confusion.  I thought the division would be dock leaders and single arriving ships on the one side, and dock followers on the other.  But with your boolean variable, it's dock leaders on one side, and everything else on the other.  That doesn't match the logic of the code, unless I'm missing something.

Goober5000: It's also worth considering making the second parameter an enum, even if there are only two options.

niffiwan: aw dang; you're correct, I got myself confused! I'll rename the param, switch the calling statements if required & update the doco. 

Regarding enum vs bool; as it's only a binary affair (and I can't see how it'd be extended to a 3rd state) I'd prefer to keep the bool

Goober5000: Okey dokey.

code/ship/ship.cpp
```diff
@@ -18990,36 +18990,36 @@ ship_subsys* ship_get_subsys_for_submodel(ship* shipp, int submodel)
...
+ * HUD targeting, drawing thrusters, radar, AI targeting, etc
  *
  * @param stage Check one of stage1, stage2 or both
- * @param dock_leader_only only the dock leader, or leader/followers/non-docked?
+ * @param dock_leader_or_single_only check dock-leaders+singles, or dock-followers?
  *
  * @return is the ship arriving, bool
  */
-bool ship::is_arriving(ship::warpstage stage, bool dock_leader_only)
+bool ship::is_arriving(ship::warpstage stage, bool dock_leader_or_single_only)
```
Goober5000: Can you remove the word "only" from the parameter?

Good commit, and good change in the comments, but I'm surprised that the only changes necessary were within the function itself.  Do all the function calls use the boolean correctly?

niffiwan: sure, I'll remove `_only`

Yes, all the function calls use the boolean correctly, I've rechecked them and rerun my test mission to confirm.

Goober5000: Excellent.   Then I think we can finally call this done!

PR #1993, opened 2019-02-24
detail boxes can ignore default distance scaling

niffiwan:
Add a submodel property that overrides the hard-coded distance
multipliers applied to the distances at which details boxes & spheres
are rendered. The distance multipliers are set by the "model detail"
setting in the options screen. Note that this is different to the LOD
distance multipliers which are not hard-coded.

This allows some more creative uses of detail boxes & spheres, e.g. to
stop models subobjects from being rendered when they're out of player
view. Past testing showed a reasonable performance improvement with
complicated models.

This is a submodel property rather than global, or a ship flag, to allow
flexibility, i.e. a single model could choose this option only for the
submodels that want to use detail boxes/spheres in this fashion, while
other detail boxes/spheres use the conventional approach.

Code & ideas originally from zookeeper, I've just packaged it up for
the PR.

### DIFF
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -346,6 +346,7 @@ public:
 	bool	use_render_box_offset;		// whether an offset has been defined; needed because one can't tell just by looking at render_box_offset
 	int		use_render_sphere;		// 0==do nothing, 1==only render this object if you are inside the sphere, -1==only if you're outside
 	bool 	use_render_sphere_offset;// whether an offset has been defined; needed because one can't tell just by looking at render_sphere_offset
+	bool    do_not_scale_detail_distances{false}; // if set should not scale boxes or spheres based on 'model detail' settings
 	bool	gun_rotation;			// for animated weapon models
 	bool	no_collisions;			// for $no_collisions property - kazan
 	bool	nocollide_this_only;	//SUSHI: Like no_collisions, but not recursive. For the "replacement" collision model scheme.
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -1513,6 +1513,11 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
 					} else {
 						pm->submodel[n].render_box_max = pm->submodel[n].max;
 					}
+
+					if ( (p = strstr(props, "$do_not_scale_distances")) != nullptr ) {
+						p += 23;
+						pm->submodel[n].do_not_scale_detail_distances = true;
+					}
 				}
 
 				if ( (p = strstr(props, "$detail_sphere:")) != NULL ) {
@@ -1540,6 +1545,11 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
 						pm->submodel[n].use_render_sphere_offset = true;
 					} else {
 						pm->submodel[n].render_sphere_offset = vmd_zero_vector;
+					}
+
+					if ( (p = strstr(props, "$do_not_scale_distances")) != nullptr ) {
+						p += 23;
+						pm->submodel[n].do_not_scale_detail_distances = true;
 					}
 				}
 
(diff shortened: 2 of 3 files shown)


code/model/model.h
```diff
@@ -346,6 +346,7 @@ class bsp_info
 	bool	use_render_box_offset;		// whether an offset has been defined; needed because one can't tell just by looking at render_box_offset
 	int		use_render_sphere;		// 0==do nothing, 1==only render this object if you are inside the sphere, -1==only if you're outside
 	bool 	use_render_sphere_offset;// whether an offset has been defined; needed because one can't tell just by looking at render_sphere_offset
+	bool    lock_detail_object_distances{false}; // if set should not scale boxes or spheres based on 'model detail' settings
```
Goober5000: I would prefer a different term than "lock"; it is not clear without reading the comment.  My first thought was "do_not_scale" but if I understand correctly they *can* scale, just not according to the hard-coded values.

Goober5000: Also, how does this PR allow subobjects to not be rendered when they are out of view?

niffiwan: Your 1st thought was correct, the new subobject property `$lock_distances` disables scaling of detail box distances - the same distances are used no matter what `model detail` is set to in options.

The code already allows subobjects to be rendered when they are out of view. The high level approach is to set `$detail_sphere` to `-1` (i.e. only render _outside_ the sphere) then use offsets to put the sphere on the opposite side of the ship to the subobject. This all works in current code; EXCEPT when a player sets `model detail` to max (i.e. 5) which multiplies all `$detail_sphere` and `$detail_box` distances by 1.2 - which causes the subobjects to not be drawn on the same side of the ship, instead of only on the other side of the ship. 

So - this PR allows an individual subobject to disable scaling-by-hardcoded-magic-numbers, so the subobjects draw _exactly_ when they are wanted no matter what the player detail settings are.

Anyway, let me know if all that makes sense before I look at possibly changing the variable names.

Goober5000: All of that makes sense, except for why multiplying `$detail_sphere` by 1.2 causes it not to work.  Shouldn't that just make it even more negative?

niffiwan: The value of `$detail_sphere` isn't being multiplied, rather `$radius` is multiplied by 1.2; which makes the area where the detail_sphere larger; which can result in the sphere (or box) moving to the other side of the ship. And the subobjects don't display when you want them to - on the FotG ISD model it results in really obvious black textures showing when you get close.

For reference; here's an example of a current suboject properties:
```
$detail_sphere: -1
$radius:4000
$offset:0,-4200,200
$lock_distances
```

Yes, we could change the radius such that offset is larger than 4800 (4000*1.2), but that's causing the subobject to display in more situations, the lower you set your graphics options i.e. counter-intuitive to the player

(yes; this way of using `$detail_sphere` is reasonably counter-intuitive as well :))

Goober5000: Okay, fair enough.

I'll defer to whatever you think is the more intuitive property name and flag name -- if it's lock_distances instead of do_not_scale_distances, we'll go with that.

wookieejedi:
Any update on this one?

niffiwan:
I've come to agree with Goober's suggested names, I'm just putting it off due to me not wanting to change the 200+ subsystems in FotG ships where I've already used the original name... :p

wookieejedi:
Definitely understandable!

niffiwan:
laziness begone - I have updated the code & models enough to test it works. Once merged I'll start fixing up all the other pofs for FotG

wookieejedi:
Excellent news!

niffiwan:
@Goober5000 when do you think you'll have time to have a look at this one?

wookieejedi:
Don't mind me, just poking this PR.

Goober5000:
Sorry, I didn't notice the ping among all my other pings. :)  Currently catching up on stuff.

New changes look great. :+1: Approving and merging.

PR #1998, opened 2019-03-03
make sure to set all hook variables when the script hook is called

### LINKED ISSUE #1996: Could not find index 'Weapon' in type 'HookVariables' when using $On Weapon Collide
This problem comes up pretty erratically, and I can't uncover the exact root cause. 99% of the time, everything works as expected, but 1% of the time, it suddenly raises an odd error.

`Could not find index 'Weapon' in type 'HookVariables'`

Where the trace stack points to a script with these at the indicated lines

```
$On Weapon Collision:
[
if hv.Self:getBreedName() == "Ship" and hv.Weapon.Parent:isValid() and (hv.Weapon.Team ~= hv.Self.Team) then
	local parent = hv.Weapon.Parent
```

Weapon should be a perfectly valid hook variable here.

Going through a debug build and waiting for it to trip the error again gives me this call stack

https://pastebin.com/yaB8gZac

Reproducing this is hard because, again, I can't narrow down what is actually causing this. But m!m and I think its probably a regression because this script was working fine before (with JAD2.21 and 2.22) and no drastic changes were made to the script for about 2 years now.

Goober5000:
In this block of code, the conditions were checked with a set of four hook variables but then subsequently run with only two.  This adds all four previously used variables.  More importantly, it adds the hook variables with expected and user-friendly names, since 'self' and 'object' are not necessarily intuitive.

This should fix #1996.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -2462,11 +2462,9 @@ int parse_create_object_sub(p_object *p_objp)
 	// If the ship is in a wing, this will be done in mission_set_wing_arrival_location() instead
 	// If the ship is in a wing, but the wing is docked then addition of bool brought_in_docked_wing accounts for that status --wookieejedi
 	if (Game_mode & GM_IN_MISSION && ((shipp->wingnum == -1) || (brought_in_docked_wing))) {
-		if (anchor_objnum >= 0)
-			Script_system.SetHookObjects(2, "Ship", &Objects[objnum], "Parent", &Objects[anchor_objnum]);
-		else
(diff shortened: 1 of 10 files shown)


MageKing17:
The same paradigm is used in collidedebrisship.cpp, collidedebrisweapon.cpp, and beam.cpp; if it's a problem in ship/weapon collisions, then it's a problem everywhere.

Goober5000:
I replicated the standard 4-variable script hook pattern for every collision check in the code.  Also, given the fact that this does appear to fix the problem (as I just noticed in Discord), I went and ensured that setting and removing of hook variables only happens in pairs.  (With any luck this may fix some of the obscure memory leaks that have popped up now and then.)

PR #1999, opened 2019-03-03
Unify Cfile blocks and CFILE handles

asarium:
The only thing a CFILE contained was a version field which isn't used
anymore so that could be removed which basically meant that CFILE was
just an indirection to a Cfile_block which I then just renamed to CFILE.
The CFILE type is an opaque handle now since the contents are not
relevant to a user of Cfile.

### DIFF
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -101,8 +101,7 @@ static int Cfile_stack_pos = 0;
 
 static char Cfile_stack[CFILE_STACK_MAX][CFILE_ROOT_DIRECTORY_LEN];
 
-Cfile_block Cfile_block_list[MAX_CFILE_BLOCKS];
-static CFILE Cfile_list[MAX_CFILE_BLOCKS];
+std::array<CFILE, MAX_CFILE_BLOCKS> Cfile_block_list;
 
 static const char *Cfile_cdrom_dir = NULL;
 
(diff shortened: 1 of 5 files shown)


code/cfile/cfile.cpp
```diff
@@ -1015,20 +1000,16 @@ static CFILE *cf_open_packed_cfblock(const char* source, int line, FILE *fp, int
...
-		cfp->version = 0;
-		cfbp->data = NULL;
-		cfbp->fp = fp;
-		cfbp->mem_mapped = false;
-		cfbp->dir_type = type;
-		cfbp->max_read_len = 0;
-
-		cfbp->source_file = source;
-		cfbp->line_num = line;
+		cfp = &Cfile_block_list[cfile_block_index];
```
MageKing17: Wouldn't it be cleaner to just declare and define it on one line?

code/cfile/cfile.cpp
```diff
@@ -1061,43 +1042,39 @@ static CFILE *cf_open_mapped_fill_cfblock(const char* source, int line, FILE *fp
...
-		Cfile_block *cfbp;
-		cfbp = &Cfile_block_list[cfile_block_index];
-
-		cfp = &Cfile_list[cfile_block_index];
-		cfp->id = cfile_block_index;
-		cfp->version = 0;
-		cfbp->max_read_len = 0;
-		cfbp->fp = NULL;
-		cfbp->mem_mapped = true;
+		cfp = &Cfile_block_list[cfile_block_index];
```
MageKing17: Ditto.

code/cfile/cfile.cpp
```diff
@@ -1114,30 +1091,26 @@ static CFILE *cf_open_memory_fill_cfblock(const char* source, int line, const vo
 	}
 	else {
 		CFILE *cfp;
-		Cfile_block *cfbp;
-		cfbp = &Cfile_block_list[cfile_block_index];
+		cfp = &Cfile_block_list[cfile_block_index];
```
MageKing17: And again.

code/cfile/cfile.cpp
asarium: Indeed it would. I made a lot of bulk changes and must have missed this.

PR #2002, opened 2019-03-04
spelling fix

Goober5000:
now that all troys have been removed from all subsystems, this word can be changed


MageKing17:
at last, we have achieved victory against the troys

Issue #1995, opened 2019-02-28
Possible regression: explosion-effect sexp number out of range


ariel-:
As documented [here](https://www.hard-light.net/forums/index.php?topic=95491.0)
@niffiwan 

Steps to reproduce:
- Get latest Blue Planet: Complete and FSO nightly from Knossos
- Execute with following cmdline:
`fs2_open_3_8_1_20190205_caea0f5_x64_SSE2-FASTDBG.exe -window -mod blueplanetcomplete-1.0.1,MVPS-3.7.2 -start_mission bp2-23.fs2 -parse_cmdline_only`

Reference log: https://fsnebula.org/log/5c77691dcb0d3350a87569c1

This behaviour was not present in 3.8.0-3 release version

Goober5000:
This looks like it is related to the fireball.tbl changes I added recently.  Will take a look.

Goober5000:
I tried several times to run Blue Planet Complete in a debug build (within the Visual Studio debugger) and kept getting Out Of Memory errors.

However, from a visual inspection of the log and the mod, this is almost certainly a mod bug that was uncovered by the stricter error checking in the new fireballs.tbl.  Note that there have only ever been six default fireballs in the table, numbered 0 through 5; an index of 6 is out of range.  Here are the definitions from 3.8.0...

```
#define FIREBALL_EXPLOSION_MEDIUM	0		// Used for the 4 little explosions before a ship explodes
#define FIREBALL_WARP				1		// Used for the warp in / warp out effect
#define FIREBALL_KNOSSOS			2		// Used for the KNOSSOS warp in / warp out effect
#define FIREBALL_ASTEROID			3
#define FIREBALL_EXPLOSION_LARGE1	4		// Used for the big explosion when a ship breaks into pieces
#define FIREBALL_EXPLOSION_LARGE2	5		// Used for the big explosion when a ship breaks into pieces
```

There are no fireballs.tbl or *-fbl.tbm tables in bpc-core.vp, and mv_effects-fbl.tbm just modifies the existing entries without adding any new ones.  Thus there could never have been a fireball with an index of 6, even in 3.8.0.

The only change in behavior here is that in the new code, this error is caught when the mission is parsed rather than silently failing at the time the sexp is triggered.

MageKing17:
Why is this a fatal error rather than a warning, though? There's a really obvious recovery path for FSO to take: namely, just keep running without spawning a nonexistent fireball.

[...]

PR #2003, opened 2019-03-05
cancel mission load when a syntax error is found

Goober5000:
Previously, when a sexp syntax error was encountered, the engine would display an Error dialog which would usually result in aborting the entire game session.  This PR changes things so that a Warning dialog is displayed and the engine stops loading the mission.

This addresses one of the complaints in #1995.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -374,7 +374,7 @@ extern fix game_get_overall_frametime();	// for texture animation
 
 // local prototypes
 void parse_player_info2(mission *pm);
-int post_process_mission();
+bool post_process_mission();
 int allocate_subsys_status();
 void parse_common_object_data(p_object	*objp);
 void parse_asteroid_fields(mission *pm);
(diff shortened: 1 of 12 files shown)


code/mission/missionparse.cpp
```diff
@@ -5840,21 +5843,18 @@ void post_process_mission()
...
 
-				if (!Fred_running) {
-					nprintf(("Error", "%s", error_msg.c_str()));
-					Error(LOCATION, "%s", error_msg.c_str());
-				} else {
-					nprintf(("Warning", "%s", error_msg.c_str()));
-					Warning(LOCATION, "%s", error_msg.c_str());
-				}
+				// syntax errors are unrecoverable, so abort
+				return -3;
```
asarium: This should use an `enum class` instead of a magic number or possibly a `bool` if there are only two possible return values.

Goober5000: It's magic numbers all the way down though.  I pushed another commit but it makes this PR much more complicated.

Goober5000:
bump

PR #2004, opened 2019-03-07
Add `Chase View Only` Option for HUD Gauges

wookieejedi:
Currently, the player is unable to see a RTT hud gauge if in chase view. This adds an option in the `hud_gauges.tbl` that tells the game whether to draw a gauge based on if chase view is the active view. So for example, a modder using RTT mapped gauges can now make a HUD that displays on the interior of the ship (ie RTT with the cockpit model) and a HUD that displays when the player switches to chase view. If merged I will update the wiki for usage in `hud_gauges.tbl`.

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -247,7 +247,7 @@ static int Damage_flash_timer;
 HudGauge::HudGauge():
 base_w(0), base_h(0), gauge_config(-1), font_num(font::FONT1), lock_color(false), sexp_lock_color(false), reticle_follow(false),
 active(false), off_by_default(false), sexp_override(false), pop_up(false), disabled_views(0), custom_gauge(false),
-texture_target(-1), canvas_w(-1), canvas_h(-1), target_w(-1), target_h(-1)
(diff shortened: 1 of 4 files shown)


wookieejedi:
Hmm the code successfully build and ran the tests correctly on my local machine. I'm not entirely sure why Travis failed with the ordering of the variables in hud.h?

ngld:
GCC is complaining about the fact that you set `only_render_in_chase_view` after `target_h` in the constructor although it's declared before `target_h` in the header. Move `only_render_in_chase_view` to the correct position in the constructor (line 250, 275, 311 in `hud.cpp`) and it should compile fine.

wookieejedi:
> GCC is complaining about the fact that you set `only_render_in_chase_view` after `target_h` in the constructor although it's declared before `target_h` in the header. Move `only_render_in_chase_view` to the correct position in the constructor (line 250, 275, 311 in `hud.cpp`) and it should compile fine.

Ah, that makes sense. Thanks for the quick and super helpful reply!

PR #2006, opened 2019-03-14
Save Prompt for the Cancel button

Strygon:
Adjusted eventeditor.cpp to have a "Do you want to save first" prompt when clicking "Cancel" to prevent loss of progress through misclicking.

### DIFF
--- a/fred2/eventeditor.cpp
+++ b/fred2/eventeditor.cpp
@@ -796,8 +796,22 @@ void event_editor::OnCancel()
 // this is called the clicking the ID_CANCEL button
 void event_editor::On_Cancel()
 {
+	int z;
+
 	audiostream_close_file(m_wave_id, 0);
 	m_wave_id = -1;
+
+	if (query_modified()) {
+		z = MessageBox("Do you want to keep your changes?", "Close", MB_ICONQUESTION | MB_YESNOCANCEL);
+		if (z == IDCANCEL) {
+			return;
+		}
+
+		if (z == IDYES) {
+			OnOk();
+			return;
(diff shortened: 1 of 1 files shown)


The-E:
I think this PR is a wee bit problematic.

In a modal dialog (which the event editor is), "OK" and "CANCEL" have distinct meanings. "OK" means "exit the dialog accepting all changes", "CANCEL" means "exit the dialog discarding all changes". 
In a dialog such as this one, the proper way to handle unsaved changes is to ask "There are unsaved changes, do you really want to Cancel?" and confirming the user's intent, instead of asking "Do you wish to save?" and altering it.

fred2/eventeditor.cpp
```diff
@@ -796,9 +796,23 @@ void event_editor::OnCancel()
 // this is called the clicking the ID_CANCEL button
 void event_editor::On_Cancel()
 {
+	int z;
+
 	audiostream_close_file(m_wave_id, 0);
 	m_wave_id = -1;
 
+	if (query_modified()) {
+		z = MessageBox("Do you want to keep your changes?", "Close", MB_ICONQUESTION | MB_YESNOCANCEL);
```
asarium: 1. `z` is not a very descriptive variable name. Maybe call it something like `choice`?
2. The scope of variables should be as small as possible so you should declare the variable when you need it first and not before. I know that most of FRED is written with most variables declared at the beginning of the function but that code is old and uses an outdated coding style.
3. Maybe use a `switch`-`case` here? I think that would make the code a little easier to read.

PR #2008, opened 2019-03-22
New translateable strings for the strings.tbl

NovachenFS2:
This commit add Ids for existing strings, that were not translateable before. Now they can be translated through the strings.tbl.
The string limit of FSO was bumped up to 1641 for this.

The following is translateable now:
- The 'or' seperator in the Keypress Directive gauge, if one function is bound to two keys.
- The Supernova warning.
- The error message that is displayed before the pilot selection screen, if something is wrong with the loaded modification.

### DIFF
--- a/code/controlconfig/controlsconfigcommon.cpp
+++ b/code/controlconfig/controlsconfigcommon.cpp
@@ -504,7 +504,9 @@ char *translate_key(char *key)
 	// both key and joystick button are mapped to this control
 	if ((key_code >= 0 ) && (joy_code >= 0) ) {
 		strcpy_s(text, key_text);
-		strcat_s(text, XSTR( " or ", 1638));
+		strcat_s(text, " ");
+		strcat_s(text, XSTR("or", 1638));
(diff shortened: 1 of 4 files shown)


code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -504,7 +504,7 @@ char *translate_key(char *key)
 	// both key and joystick button are mapped to this control
 	if ((key_code >= 0 ) && (joy_code >= 0) ) {
 		strcpy_s(text, key_text);
-		strcat_s(text, " or ");
+		strcat_s(text, XSTR( " or ", 1638));
```
asarium: I think the white space should not be part of the translatable string since that could lead to some easy mistakes:
```suggestion
		strcat_s(text, " ");
		strcat_s(text, XSTR("or", 1638));
		strcat_s(text, " ");
```

NovachenFS2: That make sense, yes.

PR #2009, opened 2019-03-28
Add `drawImage` to Scripted Gauge

wookieejedi:
Adds an additional draw function, `drawImage` to the new Scripted Gauge API. It testing it is most useful when using with a dynamic texture, ie per-frame changes can be drawn to the dynamic texture, then the texture is drawn to the scripted HUD gauge. In other words, a scripter can draw whatever they want to a dynamic texture then simply just use one line (`:drawImage`) to have it all be updated on on the scripted HUD gauge. 

Putting this here just for my reference as well. 

```
$On Gameplay Start: [
		if dy_texture == nil then 
			dy_texture = gr.createTexture(250, 250, TEXTURE_DYNAMIC) --canvas_size, set by gauge type
		end
		
		local scriptied_gname = "Custom Gauge" --or whatever the name of the scripted gauge is in hud_gauges.tbl. 
		if hu.getHUDGaugeHandle(scriptied_gname) and CustomWingGauge.RTT.canvas ~= nil then 
			hu.getHUDGaugeHandle(scriptied_gname).RenderFunction = function(dy_texture)
				gauge:drawImage(dy_texture)
			end
		end
]

$On HUD Draw: [
	if hu.HUDDrawn and mn.getMissionTime() > 0.1 then 				
		--draw gauge 
		if dy_texture ~= nil then 
			if dy_texture:isValid() then
				gr.setTarget(dy_texture)
				gr.clearScreen()
				--do whatever draw functions, ie drawString, drawCircle, etc 
				gr.setTarget()
			end
		end
	end
]
```

### DIFF
--- a/code/scripting/api/objs/hudgauge.cpp
+++ b/code/scripting/api/objs/hudgauge.cpp
@@ -3,6 +3,8 @@
 
 #include "hudgauge.h"
 #include "hud/hudscripting.h"
+#include "texture.h"
+#include "texturemap.h"
 
 namespace scripting {
 namespace api {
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/hudgauge.cpp
```diff
@@ -3,6 +3,8 @@
 
 #include "hudgauge.h"
 #include "hud/hudscripting.h"
+#include "texture.h"
+#include "texturemap.h"
```
asarium: I don't think this include is needed.

asarium:
FYI, drawing you content to a texture and then drawing that to a HUD gauge is terrible for performance since that sets up an entire render target just for doing some simple 2D rendering. It would be better to port the relevant functions to the HUD gauge system and then using those directly.

Also, if you draw your texture in the HUD draw hook then your rendering will always be one frame behind since the currently rendered pixels will only be shown in the next frame when the texture is used.

wookieejedi:
Ah gothca. PMed you Discord to not clutter this discussion. Thanks!

niffiwan:
I don't think you need to worry about cluttering this conversation, in fact I'd prefer to see all the details related to a PR in one place, easier to get context!  :) Any chance of including the Discord PM convo here?

asarium:
The summary of that conversation is basically that additional render targets are bad for performance and should not be used if possible.

PR #42, opened 2015-04-27
*-argument sexp short-circuiting

MageKing17:
Allows *-argument SEXPs to short-circuit by allowing the *-of SEXPs to short-circuit with SEXP_KNOWN_FALSE. See Mantis 3136.

PR #2016, opened 2019-04-18
Three sexps and some sexp improvements

Goober5000:
Several changes here:

* Made the "magic number" sexp return values even more unlikely, in case FREDders want to use numbers around -37000
* Optimize `eval_sexp` by not running the positive/negative number reconciliation code if the negative number is a magic number
* ~~Always check `SEXP_KNOWN_TRUE` when checking `SEXP_TRUE`, and likewise for false. This may not be needed, but better safe than sorry.~~
* Add `is-nan` which checks for not-a-number values
* Add `nan-to-number` which casts NaNs to 0
* Add `functional-if-then-else` which is like the ?: ternary operator

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -6334,7 +6334,7 @@ int sexp_directive_value(int n)
 		replace_current_value = eval_sexp(n);
 	}
 
-	if (replace_current_value == SEXP_FALSE) { // note: any SEXP_KNOWN_FALSE result will return SEXP_FALSE
+	if (replace_current_value == SEXP_FALSE || replace_current_value == SEXP_KNOWN_FALSE) {
 		Directive_count += directive_value;
 	}
 	else {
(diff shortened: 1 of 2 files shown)


code/parse/sexp.cpp
```diff
@@ -22970,11 +22970,11 @@ bool is_sexp_true(int cur_node, int referenced_node)
 {
 	int result = eval_sexp(cur_node, referenced_node);
 
-	// any SEXP_KNOWN_TRUE result will return SEXP_TRUE from eval_sexp, but let's be defensive
+	// any* SEXP_KNOWN_TRUE result will return SEXP_TRUE from eval_sexp, but let's be defensive
+	// *for almost all return paths... there is one path where it has not yet been proven that it won't
```
MageKing17: ...Well, what is that code path, and why not just guarantee it instead of doing a partial revert of PR #42 "just in case"?

Goober5000: Oh dear, it looks like PR #42 may need more scrutinizing if we're adding KNOWN_ values back in. :-/

I made these changes because of lines 25573 et seq:
```
		if ( Sexp_nodes[cur_node].value == SEXP_NAN ) {	// if we had a nan, but now don't, reset the value
			Sexp_nodes[cur_node].value = SEXP_UNKNOWN;
			return sexp_val;
		}
```

...Having gone back and looked at those lines, though, I see that the cases where `sexp_val` is KNOWN_ or FOREVER_ are *already* checked in the several if() blocks proceeding that one.  So it turns out that the leaky code path is, in fact, not leaky at all.  I'll revert those modifications.

Goober5000:
Bump

Issue #2015, opened 2019-04-16
Bug: `detail_sphere` and `detail_box` don't work correctly with `show ship` flag


wookieejedi:
Using the `show ship` flag seems in conjunction with `detail_sphere` or `detail_box` does not work as expected. In short, `detail_sphere:-1` does the opposite of what is expected when using `show ship` flag and internal view. Detailed tests are listed below:

Objective: Make a `$detail_sphere:-1` in the `detail0` submodel of a pof with a radius of 100. 
Expected Behavior:
The model should only render if the player's view or distance is outside the detail sphere. 
Observed Behavior:
Using `show ship` and internal view results in the model still rendering (incorrectly) even though the view is within the detail sphere radius and `$detail_sphere:-1`. 

Objective: Make a `$detail_sphere:1` in the `detail0` submodel of a pof with a radius of 100. 
Expected Behavior:
The model should only render if the player's view or distance is inside the detail sphere. 
Observed Behavior:
Using `show ship` and internal view results in the model not rendering (incorrectly) even though the view is within the detail sphere radius and `$detail_sphere:1`. 

Also, everything works as expected in external view, so it it leads me to think it's an bug with `show ship` or the internal view camera distance. 

I have attached a mission and modified pof to showcase this likely bug. I would be happy to provide any additional tests or answer any questions. 

[data.zip](https://github.com/scp-fs2open/fs2open.github.com/files/3085115/data.zip)

wookieejedi:
So I did some digging and the issue seems to be due to inconsistent `viewpos` values. The function that determines if the detail_sphere or box is rendered is `bool model_render_check_detail_box(vec3d *view_pos, polymodel *pm, int submodel_num, uint flags)` and that function uses `view_pos` which when printing the values each frame it appears to be very inconsistent when in internal view. This results in the rendering not occurring as expected. Here is my tests:
```
Internal View:
viewpos is 997.783203, -10.073664, -1278.403320 
render_detail_sphere? is -1 
viewpos is -22.668497, 6.508561, 0.132012 
render_detail_sphere? is 1 
viewpos is 997.783020, -10.073786, -1278.403564 
render_detail_sphere? is -1 
viewpos is -22.845505, 6.363339, -0.187189 
render_detail_sphere? is 1 
viewpos is 997.782898, -10.073902, -1278.403687 
render_detail_sphere? is -1 
viewpos is -23.061804, 6.264188, -0.377727 
render_detail_sphere? is 1 
[...]

PR #2017, opened 2019-04-20
Fix detailbox showship bug

wookieejedi:
This PR fixes issue #2015, which is a bug where using `show ship` caused `detail_box` calculations to incorrectly use world coordinates for sub-models on the player's ship, rather then the correct local coordinates.

I figured out that the `show_ship rendering` couldn't be changed without causing issues...so I just added an extra optional argument from `show_ship` that gets passed to the `detail_box` check, and that argument tells `detail_box` to do it's distance checks from local coordinates instead of the world coordinates the `show_ship` was using.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -6789,10 +6789,27 @@ void ship_render_cockpit(object *objp)
 
 void ship_render_show_ship_cockpit(object *objp)
 {
+	if (objp->type != OBJ_SHIP || objp->instance < 0)
+		return;
+
 	vec3d cockpit_eye_pos;
 	matrix dummy;
 	ship_get_eye(&cockpit_eye_pos, &dummy, objp, true, true); //Get cockpit eye position
-	
+
+	// Get the eye position for the ship in local coordinates
+	// Will be used for detail_box calculations, which use local coordinates
+	ship* shipp    = &Ships[objp->instance];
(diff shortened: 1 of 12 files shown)


code/ship/ship.cpp
```diff
@@ -6804,8 +6821,8 @@ void ship_render_show_ship_cockpit(object *objp)
 
 	render_info.set_object_number(OBJ_INDEX(objp));
 	render_info.set_detail_level_lock(0);
-
-	model_render_immediate(&render_info, Ship_info[Ships[objp->instance].ship_info_index].model_num, &objp->orient, &vmd_zero_vector); // Render ship model with fixed detail level 0 so its not switching LOD when moving away from origin
+	// Render ship model with fixed detail level 0 so its not switching LOD when moving away from origin
+	model_render_immediate(&render_info, sip->model_num, &objp->orient, &vmd_zero_vector, 3, true, pos_for_detail_box_check, true);
```
niffiwan: It would be better to use MODEL_RENDER_ALL here instead of the magic number '3'

wookieejedi: Good point.

wookieejedi:
Hmm I ran `$ git rebase -i HEAD~4`, converted the `pick` to `squash` then pushed to origin, but somehow it make more commits? Sorry about that!

PR #2018, opened 2019-04-21
Fix `detail_box` calculations when using `show ship`

wookieejedi:
This PR fixes issue #2015, which is a bug where using `show ship` caused detail_box calculations to incorrectly use world coordinates for submodels on the player's ship, rather then the correct local coordinates.

This PR replaces the one that got mangled by re-basing issues. I apologize for the inconvenience.

### DIFF
--- a/code/model/modelrender.cpp
+++ b/code/model/modelrender.cpp
@@ -2539,13 +2539,13 @@ void model_render_debug(int model_num, matrix *orient, vec3d * pos, uint flags, 
 	gr_zbuffer_set(save_gr_zbuffering_mode);
 }
 
-void model_render_immediate(model_render_params *render_info, int model_num, matrix *orient, vec3d * pos, int render, bool sort)
+void model_render_immediate(model_render_params* render_info, int model_num, matrix* orient, vec3d* pos, int render, bool sort, vec3d show_ship_pos, bool is_using_showship)
 {
 	model_draw_list model_list;
 
 	model_list.init();
 
-	model_render_queue(render_info, &model_list, model_num, orient, pos);
+	model_render_queue(render_info, &model_list, model_num, orient, pos, show_ship_pos, is_using_showship);
 
 	model_list.init_render(sort);
 
@@ -2581,7 +2581,7 @@ void model_render_immediate(model_render_params *render_info, int model_num, mat
 	}
 }
 
-void model_render_queue(model_render_params *interp, model_draw_list *scene, int model_num, matrix *orient, vec3d *pos)
+void model_render_queue(model_render_params* interp, model_draw_list* scene, int model_num, matrix* orient, vec3d* pos, vec3d show_ship_pos, bool is_using_showship)
 {
 	int i;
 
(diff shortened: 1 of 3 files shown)


code/model/modelrender.cpp
```diff
@@ -2539,13 +2539,13 @@ void model_render_debug(int model_num, matrix *orient, vec3d * pos, uint flags,
 	gr_zbuffer_set(save_gr_zbuffering_mode);
 }
 
-void model_render_immediate(model_render_params *render_info, int model_num, matrix *orient, vec3d * pos, int render, bool sort)
+void model_render_immediate(model_render_params* render_info, int model_num, matrix* orient, vec3d* pos, int render, bool sort, vec3d show_ship_pos, bool is_using_showship)
```
asarium: The vector should either be `const` reference or even a pointer that defaults to `nullptr` which would also remove the need for the additional `bool` parameter.

code/model/modelrender.cpp
```diff
@@ -2813,6 +2813,9 @@ void model_render_queue(model_render_params *interp, model_draw_list *scene, int
 	//*************************** draw the hull of the ship *********************************************
 	vec3d view_pos = scene->get_view_position();
 
+	if (is_using_showship) {
```
asarium: The control flow can be improved by only calling `scene->get_view_position()` in the else case since we have unnecessary work here if `is_using_showship` is `true`.

code/model/modelrender.h
```diff
@@ -296,8 +296,8 @@ class model_draw_list
 	void reset();
 };
 
-void model_render_immediate(model_render_params *render_info, int model_num, matrix *orient, vec3d * pos, int render = MODEL_RENDER_ALL, bool sort = true);
-void model_render_queue(model_render_params *render_info, model_draw_list* scene, int model_num, matrix *orient, vec3d *pos);
+void model_render_immediate(model_render_params* render_info, int model_num, matrix* orient, vec3d* pos, int render = MODEL_RENDER_ALL, bool sort = true, vec3d show_ship_pos = vmd_zero_vector, bool is_using_showship = false);
```
asarium: This is my main problem with these changes. It adds additional parameters to a very common function that are only needed for one specific use case. In my opinion, it would be better to fix the underlying issue that the view position is not set correctly instead of taking this shortcut.

code/ship/ship.cpp
```diff
@@ -6789,10 +6789,27 @@ void ship_render_cockpit(object *objp)
...
 	ship_get_eye(&cockpit_eye_pos, &dummy, objp, true, true); //Get cockpit eye position
 	
+	// Get the eye position for the ship in local coordinates
+	// Will be used for detail_box calculations, which use local coordinates
+	ship* shipp    = &Ships[objp->instance];
+	ship_info* sip = &Ship_info[shipp->ship_info_index];
+	polymodel* pm  = model_get(sip->model_num);
+	vec3d pos_for_detail_box_check;
+	// Check if there is a valid eye position
+	if (!(shipp->current_viewpoint < 0) && !(pm->n_view_positions == 0) && !(shipp->current_viewpoint > pm->n_view_positions)) {
```
asarium: All these negations should be resolved to make the code easier to understand.

PR #2019, opened 2019-04-22
Cleaner fix for show-ship bug with detail_box calculations

wookieejedi:
This provides a cleaner fix for the bug where using show ship caused detail_box calculations to incorrectly work on the player's ship. 

This PR also reverts the changes made by PR #2018 (that's why there are so many deletions).

Sorry for the repeated PR's and big thank you to m!m for his assistance and patience.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -6789,6 +6789,9 @@ void ship_render_cockpit(object *objp)
 
 void ship_render_show_ship_cockpit(object *objp)
 {
+	if (objp->type != OBJ_SHIP || objp->instance < 0)
+		return;
+
 	vec3d cockpit_eye_pos;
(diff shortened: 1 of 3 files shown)


code/ship/ship.cpp
```diff
@@ -6789,27 +6789,10 @@ void ship_render_cockpit(object *objp)
 
 void ship_render_show_ship_cockpit(object *objp)
 {
-	if (objp->type != OBJ_SHIP || objp->instance < 0)
```
Goober5000: Since this function must operate on a ship, you should keep this if() condition.

wookieejedi: Good point. I added it in with the previous PR so I will add it back in this one, too.

PR #2020, opened 2019-04-22
Add deprecation feature for the scripting API

asarium:
This allows to mark certain functions or virtvars to be declared as
"deprecated" starting with a specific engine version. What this means is
that a special message will be added to the documentation stating since
when the function is deprecated and can contain an optional message. The
actual enforcement of the deprecation depends on the target version of
the mod so that we can stay backwards compatible.

If the mod supports the version since when the function is deprecated,
it will not be added to the Lua state which means that it doesn't exist
for the calling script. If the targetted version is before the stated
version, the function will still be added to the Lua state.

As the first function to use this feature, I have deprecated
createParticle in the testing library and moved it to the graphics
library with a new name reflecting that it creates a persistent
particle. I also added the corresponding function for adding a
non-persistent particle.


code/scripting/ade_api.h
```diff
@@ -276,7 +303,31 @@ class ade_virtvar : public ade_lib_handle {
  */
 #define ADE_VIRTVAR(name, parent, args, desc, ret_type, ret_desc)                                                      \
 	static int parent##_##name##_f(lua_State* L);                                                                      \
-	::scripting::ade_virtvar parent##_##name(#name, parent##_##name##_f, parent, args, desc, ret_type, ret_desc);      \
+	::scripting::ade_virtvar parent##_##name(#name, parent##_##name##_f, parent, args, desc, ret_type, ret_desc,       \
+	                                         ::gameversion::version(), nullptr);                                       \
+	static int parent##_##name##_f(lua_State* L)
+
+/**
+ * @brief Declare an API variable
```
MageKing17: Should probably change this to say "Declare a **deprecated** API variable".

code/scripting/api/libs/graphics.cpp
```diff
@@ -1588,5 +1588,140 @@ ADE_FUNC(openMovie, l_Graphics, "string name, boolean looping = false",
 	return ade_set_args(L, "o", l_MoviePlayer.Set(movie_player_h(std::move(player))));
 }
 
+ADE_FUNC(createPersistentParticle, l_Testing,
```
MageKing17: I believe you meant `l_Graphics` here.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1588,5 +1588,140 @@ ADE_FUNC(openMovie, l_Graphics, "string name, boolean looping = false",
...
+
+	particle::WeakParticlePtr p = particle::createPersistent(&pi);
+
+	if (!p.expired())
+		return ade_set_args(L, "o", l_Particle.Set(particle_h(p)));
+	else
+		return ADE_RETURN_NIL;
+}
+
+ADE_FUNC(createParticle, l_Testing,
```
MageKing17: And here.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1588,5 +1588,140 @@ ADE_FUNC(openMovie, l_Graphics, "string name, boolean looping = false",
...
+
+	if (objh != NULL && objh->IsValid()) {
+		pi.attached_objnum = OBJ_INDEX(objh->objp);
+		pi.attached_sig    = objh->objp->signature;
+	}
+
+	particle::WeakParticlePtr p = particle::createPersistent(&pi);
+
+	if (!p.expired())
+		return ade_set_args(L, "o", l_Particle.Set(particle_h(p)));
```
MageKing17: Forgot to `#include "scripting/api/objs/particle.h"` in order to use `l_Particle` and `particle_h`.

code/scripting/ade.cpp
```diff
@@ -626,10 +634,21 @@ void ade_table_entry::OutputMeta(FILE *fp)
...
+					            DeprecationVersion.major, DeprecationVersion.minor, DeprecationVersion.build,
+					            DeprecationMessage);
+				    } else {
+					    fprintf(fp, "<dd><b>Deprecated starting with version %d.%d.%d.</b></dd>\n",
+					            DeprecationVersion.major, DeprecationVersion.minor, DeprecationVersion.build);
+				    }
+			    }
+
+			    //***Result: ReturnDescription
+			    if (ReturnDescription != NULL) {
```
MageKing17: I'm not sure you meant to change these lines (they now have mixed tab/space indentation), but if you did, this `NULL` should become a `nullptr` to avoid clang-tidy getting mad.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1588,5 +1588,140 @@ ADE_FUNC(openMovie, l_Graphics, "string name, boolean looping = false",
...
+         "Reverse reverse animation, if one is specified"
+         "Attached object specifies object that Position will be (and always be) relative to.",
+         "particle", "Handle to the created particle")
+{
+	particle::particle_info pi;
+	pi.type            = particle::PARTICLE_DEBUG;
+	pi.optional_data   = -1;
+	pi.attached_objnum = -1;
+	pi.attached_sig    = -1;
+	pi.reverse         = 0;
```
MageKing17: Speaking of clang-tidy, this should be set to `false` instead of `0`.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1588,5 +1588,140 @@ ADE_FUNC(openMovie, l_Graphics, "string name, boolean looping = false",
...
+
+	// Need to consume tracer_length parameter but it isn't used anymore
+	float temp;
+
+	enum_h* type       = NULL;
```
MageKing17: This should be `nullptr`.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1588,5 +1588,140 @@ ADE_FUNC(openMovie, l_Graphics, "string name, boolean looping = false",
...
+	bool rev           = false;
+	object_h* objh     = NULL;
```
MageKing17: Same here.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1588,5 +1588,140 @@ ADE_FUNC(openMovie, l_Graphics, "string name, boolean looping = false",
...
+	texture_h* texture = nullptr;
+	if (!ade_get_args(L, "ooffo|fboo", l_Vector.Get(&pi.pos), l_Vector.Get(&pi.vel), &pi.lifetime, &pi.rad,
+	                  l_Enum.GetPtr(&type), &temp, &rev, l_Texture.GetPtr(&texture), l_Object.GetPtr(&objh)))
+		return ADE_RETURN_NIL;
+
+	if (type != NULL) {
```
MageKing17: And here.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1588,5 +1588,140 @@ ADE_FUNC(openMovie, l_Graphics, "string name, boolean looping = false",
...
+				pi.type          = particle::PARTICLE_BITMAP;
+			}
+			break;
+		}
+	}
+
+	if (rev)
+		pi.reverse = 0;
+
+	if (objh != NULL && objh->IsValid()) {
```
MageKing17: And here.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1588,5 +1588,140 @@ ADE_FUNC(openMovie, l_Graphics, "string name, boolean looping = false",
...
+			} else {
+				pi.optional_data = texture->handle;
+				pi.type          = particle::PARTICLE_BITMAP;
+			}
+			break;
+		}
+	}
+
+	if (rev)
+		pi.reverse = 0;
```
MageKing17: This should also be `false`.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1588,5 +1588,140 @@ ADE_FUNC(openMovie, l_Graphics, "string name, boolean looping = false",
...
+         "Reverse reverse animation, if one is specified"
+         "Attached object specifies object that Position will be (and always be) relative to.",
+         "boolean", "true if particle was created, false otherwise")
+{
+	particle::particle_info pi;
+	pi.type            = particle::PARTICLE_DEBUG;
+	pi.optional_data   = -1;
+	pi.attached_objnum = -1;
+	pi.attached_sig    = -1;
+	pi.reverse         = 0;
```
MageKing17: Repeat previous comments about `false` and `nullptr` for this block.

code/scripting/ade_api.h
asarium: Fixed and also fixed the various clang-tidy issues.

PR #2024, opened 2019-05-01
Refactoring of objcollision

CasparKielwein:
First part of my quest to get the collision code up to scratch.
The endgoal is to use a spatial data structure like an oct-tree. But to do that, I need to understand and clean the code.
To limit the scope of the PRs, here is one with some refactoring of it.

It contains general cleanup like replacing NULL with nullptr and declaring variables at point of use.
A bit of unused and commented out code is removed to make reading it easier.
functions used only in objcollide.cpp are now in an anonymous namespace.

The largest part is the removal of some code duplication in objcollide.cpp
the functions obj_add_pair and obj_collide_pair share a large duplicate switch case statement, which has been extracted to a static function.


code/object/objcollide.cpp
```diff
@@ -98,137 +98,184 @@ int reject_due_collision_groups(object *A, object *B)
...
+ * This function was extracted from duplicate code in obj_add_pair and obj_collide_pair
+ */
+static bool check_collision_type(object *A, object *B, CollisionChecker& check_collision, int& swapped)
+{
+    Assert(A != nullptr);
+    Assert(B != nullptr);
+    Assert( A->type < 127 );
+    Assert( B->type < 127 );
 
-	case COLLISION_OF(OBJ_BEAM, OBJ_DEBRIS):
```
CasparKielwein: The original switch case in obj_add_pair was not symmetric for beam weapons.
(It covered COLLISION_OF(OBJ_BEAM, x) but not COLLISION_OF(x, OBJ_BEAM) )

Does anyone know if that is intended or even necessary?
It seems buggy to me.

The-E: It was probably intentional. 
For every pair (x, y) an equivalent pair (y, x) will exist; technically you only need to check one combination (you'll note that for those duplicate cases, a "swapped" flag is set, which will swap the A and B object references around before actually resolving the collision). 
One possible performance optimization could be to try and cull those duplicate pairs; however, I do not know if the overhead of performing that cull is larger than rejecting those pairs during the resolution phase, which is what the code is currently doing.

CasparKielwein: Maybe it is by construction, that beam objects will always be at the end of the array or something like that, and thus collisions will only ever be in that order.

But why is do the symetric cases exist for other combinations like WEAPON + x?
Does anyone know why Swifty added them in his (mostly) duplicate code?
His commit message doesn't really tell anything.

CasparKielwein: It seems that obj_add_pair is actually dead code.
It is only called in obj_add_pairs in object.cpp, which is never called.

So instead of unifying the function, I'll just delete one.

CasparKielwein:
ping @The-E ,any other change required?

code/object/objcollide.cpp
```diff
@@ -1021,466 +625,430 @@ void obj_reset_colliders()
...
+		TRACE_SCOPE(tracing::SortColliders);
+		obj_quicksort_colliders(&sort_list_y, 0, (int)(sort_list_y.size() - 1), 1);
 	}
+	obj_find_overlap_colliders(sort_list_z, sort_list_y, 1, false);
 
-	obj_pair new_pair;	
-
-	new_pair.a = A;
-	new_pair.b = B;
-	new_pair.check_collision = check_collision;
```
niffiwan: Could you please help me understand why it's OK to remove this assignment?

CasparKielwein: They have been moved to lines 963ff.
When I extracted local functions to the anonymous namespace, I moved them before their point of use.

The github diff viewer fails us here.

niffiwan: I can see these three lines there
```
    new_pair.a = A;
    new_pair.b = B;
    new_pair.next_check_time = collision_info->next_check_time;
```
but I can't see 
```
	new_pair.check_collision = check_collision;
```

(btw - I noticed this when using the github diff viewer that ignores whitespace)

CasparKielwein: Yeah, sorry that is confusing (I had to backtrack as well)
That field was unused and has been removed in this commit as well.

see objcollide.h line 57

I will add that to the commit message.

CasparKielwein:
Added additional commit message and rebased onto master.

PR #2026, opened 2019-05-02
various sexp formatting and cleanup changes

Goober5000:
While scouring sexp.cpp for all the NaN changes (to be in a future PR), I made a number of miscellaneous changes not directly related to the NaN fixes.  I thought it would be good to make a separate PR so the NaN PR is not unduly cluttered.

By the way, this should fix the `query-orders` sexp.  That sexp was using `eval_sexp` when it should have used `eval_num`.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -4147,12 +4147,12 @@ int div_sexps(int n)
 
 		while (CDR(n) != -1) {
 			int div = eval_sexp(CDR(n));
+			n = CDR(n);
 			if (div == 0) {
 				Warning(LOCATION, "Division by zero in sexp. Please check all uses of the / operator for possible causes.\n");
 				continue;
 			} 
 			sum /= div;
-			n = CDR(n);
 		}
 	}
 
@@ -4172,12 +4172,12 @@ int mod_sexps(int n)
 
 		while (CDR(n) != -1) {
 			int div = eval_sexp(CDR(n));
+			n = CDR(n);
 			if (div == 0) {
 				Warning(LOCATION, "Modulo by zero in sexp. Please check all uses of the % operator for possible causes.\n");
 				continue;
 			}
 			sum = sum % div;
-			n = CDR(n);
 		}
 	}
 
@@ -5501,14 +5501,14 @@ int sexp_has_docked_or_undocked(int n, int op_num)
 	}
 	else
 	{
-		if ( mission_log_get_time_indexed(op_num == OP_HAS_DOCKED ? LOG_SHIP_DOCKED : LOG_SHIP_UNDOCKED, docker, dockee, count, NULL) )
+		if ( mission_log_get_time_indexed(op_num == OP_HAS_DOCKED ? LOG_SHIP_DOCKED : LOG_SHIP_UNDOCKED, docker, dockee, count, nullptr) )
 			return SEXP_KNOWN_TRUE;
 	}
 
-	if ( mission_log_get_time(LOG_SHIP_DESTROYED, docker, NULL, NULL) || mission_log_get_time(LOG_SHIP_DESTROYED, dockee, NULL, NULL) )
+	if ( mission_log_get_time(LOG_SHIP_DESTROYED, docker, nullptr, nullptr) || mission_log_get_time(LOG_SHIP_DESTROYED, dockee, nullptr, nullptr) )
 		return SEXP_KNOWN_FALSE;
 
-	if ( mission_log_get_time(LOG_SELF_DESTRUCTED, docker, NULL, NULL) || mission_log_get_time(LOG_SELF_DESTRUCTED, dockee, NULL, NULL) )
+	if ( mission_log_get_time(LOG_SELF_DESTRUCTED, docker, nullptr, nullptr) || mission_log_get_time(LOG_SELF_DESTRUCTED, dockee, nullptr, nullptr) )
 		return SEXP_KNOWN_FALSE;
 
 	return SEXP_FALSE;
@@ -6677,7 +6677,7 @@ int sexp_distance_subsystem(int n)
 	return SEXP_NAN;
 }
 
-bool sexp_helper_is_within_box(std::array<float, 6> box_vals, vec3d *pos)
+bool sexp_helper_is_within_box(const std::array<float, 6> &box_vals, vec3d *pos)
 {
 	int i;
 	for(i = 0; i < 3; i++)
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -15619,9 +15555,9 @@ int sexp_path_flown()
 	return SEXP_FALSE;
 }
 
-void sexp_send_training_message(int node)
+void sexp_send_training_message(const int node)
```
MageKing17: Why is this the only time `const` was added?

code/parse/sexp.cpp
```diff
@@ -19389,7 +19262,10 @@ void unset_nav_carry_status(int node)
 			{
 				if (Ships[i].objnum != -1 && !stricmp(Ships[i].ship_name, name))
 				{
-                    Ships[i].flags.remove(Ship::Ship_Flags::Navpoint_carry);
+					if (set_it)
+	                    Ships[i].flags.set(Ship::Ship_Flags::Navpoint_carry);
+					else
+						Ships[i].flags.remove(Ship::Ship_Flags::Navpoint_carry);
```
MageKing17: This could be simplified to include the boolean in the `set()` call: `Ships[i].flags.set(Ship::Ship_Flags::Navpoint_carry, set_it)`

code/parse/sexp.cpp
```diff
@@ -19437,17 +19288,20 @@ void unset_nav_needslink(int node)
...
 			if (Ships[i].objnum != -1 && !stricmp(Ships[i].ship_name, name))
 			{
-                Ships[i].flags.remove(Ship::Ship_Flags::Navpoint_needslink);
+				if (set_it)
+				{
+					Ships[i].flags.remove(Ship::Ship_Flags::Navpoint_carry);
+					Ships[i].flags.set(Ship::Ship_Flags::Navpoint_needslink);
+				}
+				else
+					Ships[i].flags.remove(Ship::Ship_Flags::Navpoint_needslink);
```
MageKing17: Same here.

code/parse/sexp.cpp
```diff
@@ -4138,19 +4139,20 @@ int div_sexps(int n)
...
-			sum = eval_sexp(Sexp_nodes[n].first);
+		if (CAR(n) != -1) {
+			sum = eval_sexp(CAR(n));
+		}
 		else
 			sum = atoi(CTEXT(n));
 
-		while (Sexp_nodes[n].rest != -1) {
-			int div = eval_sexp(Sexp_nodes[n].rest);
-			n = Sexp_nodes[n].rest;
```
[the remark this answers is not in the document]
Goober5000: Good catch.  Fixed.

code/parse/sexp.cpp
```diff
@@ -10087,19 +10098,24 @@ void sexp_hud_set_frame(int n)
 void sexp_hud_set_color(int n)
 {
 	char* gaugename = CTEXT(n);
-	ubyte red = (ubyte) eval_num(CDR(n));
-	ubyte green = (ubyte) eval_num(CDR(CDR(n)));
-	ubyte blue = (ubyte) eval_num(CDR(CDR(CDR(n))));
+	n = CDR(n);
+
+	std::array<int, 3> rgb;
```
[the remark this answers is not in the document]
Goober5000: No advantage in this current PR.  But the next PR will introduce `eval_array` to perform `eval_num` on a number of nodes in sequence.  I moved some of the code into this PR to make the next PR cleaner.

code/parse/sexp.cpp
```diff
@@ -14493,11 +14427,9 @@ void sexp_ships_bomb_targetable(int n, bool targetable)
 void sexp_ship_guardian_threshold(int node)
 {
 	char *ship_name;
-	int ship_num, threshold;
-	int n = -1;
+	int ship_num, threshold, n = -1;
```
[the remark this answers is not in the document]
Goober5000: An oversight.  I've changed this one to initialize to `node` also.

code/parse/sexp.cpp
Goober5000: I don't actually remember. :man_shrugging: I've removed it.

code/parse/sexp.cpp
Goober5000: Aha.  Thanks.

code/parse/sexp.cpp
```diff
@@ -20920,18 +20735,23 @@ void sexp_set_training_context_fly_path(int node)
...
 void sexp_set_training_context_speed(int node)
 {
+	int min = eval_num(node);
+	int max = eval_num(CDR(node));
+
 	Training_context |= TRAINING_CONTEXT_SPEED;
-	Training_context_speed_min = eval_num(node);
-	Training_context_speed_max = eval_num(CDR(node));
+	Training_context_speed_min = min;
+	Training_context_speed_max = max;
```
CasparKielwein: Is this change really useful?
The additional variables min and max, don't really add any information to people reading the code.

code/parse/sexp.cpp
```diff
@@ -19319,54 +19231,14 @@ void sexp_primitive_sensors_set_range(int n)
...
-		
-		// move to next ship/wing in list
-		n = CDR(n);
-	}
-}
-
 //text: unset-nav-carry
 //args: 1+, Ship/Wing name
-void unset_nav_carry_status(int node)
+void set_unset_nav_carry_status(int node, bool set_it)
```
CasparKielwein: what do you think about adding a comment which explains the parameter set_it.
Its not really clear in the signature.

code/parse/sexp.cpp
Goober5000: The upcoming PR will use them.  This change makes that PR cleaner.

code/parse/sexp.cpp
Goober5000: The parameter is self-explanatory.  That sort of function signature is quite clear, and used in numerous places in the code.

Goober5000:
All complaints resolved.  Pinging @MageKing17 or @CasparKielwein

code/parse/sexp.cpp
CasparKielwein: Think of people like me, who do not have the years of experience in the codebase you have :)

Goober5000:
I guess I need @MageKing17 to approve it since he was the guy who did the review.

code/parse/sexp.cpp
Goober5000: Sure.  But what you do is a) look at the function definition, b) look at similar function definitions, and c) look at the body of the function.

Any one of those three can explain what's happening, but in this case, a) is sufficient.  The name of the function indicates that something is being set or unset, and the boolean flag (which can only take two values) indicates whether the thing is being set.

PR #2030, opened 2019-05-05
Allow death max rotation to be set per ship

niffiwan:
This allows modders to restrict the amount of rotation in a death roll.
It's fairly simple to fit in with the existing code which has a lot of
hard coded constants. The intended use case is to restrict pitch &
heading changes so there's less chance of rapid forward rolls or flat
spins, especially with faster-than-FS2 ships.

I'd almost rather rewrite the entire code to dynamically use ships mass
& velocity except for retail compatibility issues.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -833,6 +833,9 @@ void ship_info::clone(const ship_info& other)
 	death_roll_r_mult = other.death_roll_r_mult;
 	death_fx_r_mult = other.death_fx_r_mult;
 	death_roll_time_mult = other.death_roll_time_mult;
+	death_roll_xrotation_cap = other.death_roll_xrotation_cap;
+	death_roll_yrotation_cap = other.death_roll_yrotation_cap;
+	death_roll_zrotation_cap = other.death_roll_zrotation_cap;
 	death_roll_base_time = other.death_roll_base_time;
 	death_fx_count = other.death_fx_count;
 	shockwave_count = other.shockwave_count;
@@ -1136,6 +1139,9 @@ void ship_info::move(ship_info&& other)
 	death_roll_r_mult = other.death_roll_r_mult;
 	death_fx_r_mult = other.death_fx_r_mult;
 	death_roll_time_mult = other.death_roll_time_mult;
+	death_roll_xrotation_cap = other.death_roll_xrotation_cap;
+	death_roll_yrotation_cap = other.death_roll_yrotation_cap;
+	death_roll_zrotation_cap = other.death_roll_zrotation_cap;
 	death_roll_base_time = other.death_roll_base_time;
 	death_fx_count = other.death_fx_count;
 	shockwave_count = other.shockwave_count;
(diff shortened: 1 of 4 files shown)


code/ship/shiphit.cpp
```diff
@@ -1356,7 +1356,7 @@ static void player_died_start(object *killer_objp)
 
 //#define	DEATHROLL_TIME						3000			//	generic deathroll is 3 seconds (3 * 1000 milliseconds) - Moved to ships.tbl
 #define	MIN_PLAYER_DEATHROLL_TIME		1000			// at least one second deathroll for a player
-#define	DEATHROLL_ROTVEL_CAP				6.3f			// maximum added deathroll rotvel in rad/sec (about 1 rev / sec)
+//#define	DEATHROLL_ROTVEL_CAP				6.3f			// maximum added deathroll rotvel in rad/sec (about 1 rev / sec) - moved to shiphit.h & converted to const float
```
CasparKielwein: I'd prefer to remove the code instead of commenting it out.
Just add a note in your commit message that you moved DEATHROLL_ROTVEL_CAP etc.
People can look it up there and possibly restore it if required from git.

code/ship/shiphit.h
```diff
@@ -22,6 +22,8 @@ class object;
 
 #define MISS_SHIELDS		-1
 
+const float DEATHROLL_ROTVEL_CAP = 6.3f;
```
CasparKielwein: Maybe add a comment explaining why this number was chosen.
I know that the old code didn't have one either, but since you changed it. you probably know the reason at the moment and that would help people like me, who don't know that part of the code in the future.

I think we have the compiler support for constexpr float DEATHROLL_ROTVEL_CAP = 6.3f;

code/ship/shiphit.cpp
niffiwan: will do

code/ship/shiphit.h
niffiwan: I'll move the existing comment and expand it. As for why 6.3, aka 1 revolution per sec? I think it's as simple as "it looked cool in retail FS2" :)  aka an artistic choice

niffiwan:
thanks for the review; changes incoming

code/ship/ship.cpp
```diff
@@ -1452,6 +1458,9 @@ ship_info::ship_info()
 	death_roll_r_mult = 1.0f;
 	death_fx_r_mult = 1.0f;
 	death_roll_time_mult = 1.0f;
+	death_roll_xrotation_cap = 0.75f*DEATHROLL_ROTVEL_CAP;
+	death_roll_yrotation_cap = 0.75f*DEATHROLL_ROTVEL_CAP;
```
The-E: If this constant is only ever used after being multiplied by 0.75, why is that multiplication not part of that constant's value? I.e. Why is it 6.3 and not 4.725?

niffiwan: Unfortunately there's [one other place](https://github.com/scp-fs2open/fs2open.github.com/pull/2030/files#diff-d0fc37f76f2dd8a965a41def714ac5bbL1485) where the constant is used as is, i.e. without being multiplied by 0.75.

The-E: Okay, fair enough

PR #2038, opened 2019-05-09
Add new sexp: `add-remove-hotkey`

wookieejedi:
This new sexp allows a FREDer to add or remove a hotkey for a ship or wing. The sexp takes 2 or more arguments, the integer of the hotkey to add or remove (F5=0, F6=1, ... F12=7) and the names of the ship(s) and/or wing(s).


code/parse/sexp.cpp
```diff
@@ -19804,6 +19805,36 @@ void multi_sexp_set_respawns()
...
+		if (oswpt.type == OSWPT_TYPE_SHIP) {
+			objnum = oswpt.shipp->objnum;
+			if (objnum != -1)
+				hud_target_hotkey_add_remove(setnum, &Objects[objnum], HOTKEY_MISSION_FILE_ADDED);
+		}
+		if (oswpt.type == OSWPT_TYPE_WING) {
+			for (int i = 0; i < oswpt.wingp->current_count; i++) {
+				objnum = Ships[oswpt.wingp->ship_index[i]].objnum;
+				if (objnum != -1)
+					hud_target_hotkey_add_remove(setnum, &Objects[objnum], HOTKEY_MISSION_FILE_ADDED);
```
MageKing17: Using `HOTKEY_MISSION_FILE_ADDED` (both here and line 19826) means that, if there are any user-added ships in the hotkey set, this SEXP will be ignored; is that desired behavior here?

code/parse/sexp.cpp
```diff
@@ -32844,6 +32889,12 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"\tRest: The player start ship to operate on.\r\n"
 		"\tOnly really useful for multiplayer."},
 
+	{ OP_SET_HOTKEY, "set-hotkey\r\n"
+		"\tSets hotkey for the specified ship(s) or wing(s)\r\n"
```
MageKing17: It should probably mention that, if the ship(s) are already in the hotkey set, they'll be removed.

code/parse/sexp.cpp
wookieejedi: ATM that is the desired behavior. I was debating on whether to use that or `HOTKEY_USER_ADDED`., and I am very open to thoughts about it, because it's really a debate about who to favor, the mission designer or the player. My thinking was that if the mission designer specifies a hotkey and tells the player to press that key, then they want to ensure that the player didn't bind that key already, because then it might be confusing. In contrast though, the player might be annoyed that their hotkey just went away. Perhaps I should add a bool argument that is something along the lines of `override`, so if true then it uses `HOTKEY_MISSION_FILE_ADDED` and if false it uses `HOTKEY_USER_ADDED`?

code/parse/sexp.cpp
wookieejedi: Ah good point, though in my tests the running the test again did not remove the current hotkeys set by the player or the mission. To actually clear the hotkeys I would have to use the function `hud_target_hotkey_clear()`

Also, in my tests running the sexp with the same arguments 4 seconds apart did not remove the ships in the hotkey set. So for example set-hotkey 1 for Beta did not remove Beta wing from hotkey 1 on the second evaluation of the sexp.

code/parse/sexp.cpp
wookieejedi: Oh wait.. I see what you mean. So my reasoning is still valid, but the to get the sexp to override player hotkeys (ie already existing ones) I would used `HOTKEY_USER_ADDED`.

code/parse/sexp.cpp
wookieejedi: Ah, update. So if I use `HOTKEY_USER_ADDED` then it does remove them, but if I use `HOTKEY_MISSION_FILE_ADDED` the sexp is ignored since they are already part of that hotkey set. I will work on the logic of this a submit a revision.

wookieejedi:
Alright, sorry for all the comments. So I re-read the forum post about what functionally was desired, and settled on using `HOTKEY_USER_ADDED` because it will actually set the hotkey, but also not override the player's present hotkeys, and it will allow the mission designer to set and un-set hotkeys using the same sexp. Does that sound good?

Goober5000:
That sounds good to me, but I recommend also posting that question in the thread, to get input from JSRNerdo and MjnMixael.

code/parse/sexp.cpp
```diff
@@ -19804,6 +19805,36 @@ void multi_sexp_set_respawns()
...
+
+	// first, look for ship name -- if found, then set hotkey to ship
+	// if not a ship, look for wing name, then set hotkey to each ship in wing
+	// note, hud_target_hotkey_add_remove() checks if the object has arrived or is destroyed or departed
+	for (; n != -1; n = CDR(n)) {
+		object_ship_wing_point_team oswpt;
+		sexp_get_object_ship_wing_point_team(&oswpt, CTEXT(n));
+		if (oswpt.type == OSWPT_TYPE_SHIP) {
+			objnum = oswpt.shipp->objnum;
+			if (objnum != -1)
```
Goober5000: Strictly speaking, this and the following check for `(objnum != -1)` aren't required.  If oswpt.type is SHIP or WING, then the function has already determined that the ships or wings are present in-mission.  And if they are present in-mission, then they have valid object numbers.

code/parse/sexp.cpp
```diff
@@ -19804,6 +19805,36 @@ void multi_sexp_set_respawns()
...
+				hud_target_hotkey_add_remove(setnum, &Objects[objnum], HOTKEY_MISSION_FILE_ADDED);
+		}
+		if (oswpt.type == OSWPT_TYPE_WING) {
```
Goober5000: You could make this an `else if`.  Minor nitpick.

code/parse/sexp.cpp
MageKing17: You could also add a new `#define` like, say, `HOTKEY_SEXP_ADDED` or something, and give it the exact behavior you want.

code/parse/sexp.cpp
wookieejedi: Easy adjustment, though!

code/parse/sexp.cpp
wookieejedi: Good point, I'll take those out.

wookieejedi:
Okay, I Incorporated all feedback, including Goober's the specify the rename to `add-remove-hotkey` and mjn.mixael and JSRNerdo both agreed on the forum the updated functionality would work well for them.

Goober5000:
Two additional bits of feedback...

1) The function name should also be changed from `set` to `add_remove`.

2) I considered for a while and I think you should slightly change how the sexp works.  Add an optional boolean as the first argument - true for add, and false for remove.  The reason is that the mission designer will not necessarily know whether the hotkey has already been added.  In most cases, he will want the sexp to ensure the hotkey is assigned, or ensure the hotkey is removed.  Without the boolean, the behavior of the sexp will be unpredictable.

wookieejedi:
New commit has bool argument incorporated.

code/parse/sexp.cpp
```diff
@@ -19804,6 +19805,44 @@ void multi_sexp_set_respawns()
...
+ * set a hotkey for one or more ships/wings
+ */
+void sexp_add_remove_hotkey(int node)
+{
+	int objnum, setnum, n = node;
+	bool is_adding; // True for add, False for remove
+	is_adding = is_sexp_true(n);
+	n = CDR(n);
+	setnum = eval_num(n);
+	n = CDR(n);
```
Goober5000: It would be nice to have a blank line between the parameter setup and the sexp logic, to make things easier to read.  Small stylistic nitpick.

code/parse/sexp.cpp
```diff
@@ -19804,6 +19805,44 @@ void multi_sexp_set_respawns()
...
+/**
+ * set a hotkey for one or more ships/wings
+ */
+void sexp_add_remove_hotkey(int node)
+{
+	int objnum, setnum, n = node;
+	bool is_adding; // True for add, False for remove
+	is_adding = is_sexp_true(n);
+	n = CDR(n);
+	setnum = eval_num(n);
```
Goober5000: You should test to see whether setnum is in the range of allowable hotkeys.  Also I think the maximum hotkey value is reserved for something special.

code/parse/sexp.cpp
```diff
@@ -19804,6 +19805,44 @@ void multi_sexp_set_respawns()
...
+	// if not a ship, look for wing name, then set hotkey to each ship in wing
+	// note, hud_target_hotkey_add_remove() checks if the object has arrived or is destroyed or departed
+	for (; n != -1; n = CDR(n)) {
+		object_ship_wing_point_team oswpt;
+		sexp_get_object_ship_wing_point_team(&oswpt, CTEXT(n));
+		if (oswpt.type == OSWPT_TYPE_SHIP) {
+			objnum = oswpt.shipp->objnum;
+			// check if the ship already has this hot-key and if sexp is adding or removing it
+			if (((oswpt.shipp->hotkey == setnum) && !is_adding) || 
+				(!(oswpt.shipp->hotkey == setnum) && is_adding) ) {
```
Goober5000: Instead of `!(a == b)` use `a != b`.

code/parse/sexp.cpp
```diff
@@ -19804,6 +19805,44 @@ void multi_sexp_set_respawns()
...
+				hud_target_hotkey_add_remove(setnum, &Objects[objnum], HOTKEY_USER_ADDED);
+			}
+		}
+		else if (oswpt.type == OSWPT_TYPE_WING) {
+			for (int i = 0; i < oswpt.wingp->current_count; i++) {
+				objnum = Ships[oswpt.wingp->ship_index[i]].objnum;
+				// check if the ship already has this hot-key and if sexp is adding or removing it
+				if (((oswpt.shipp->hotkey == setnum) && !is_adding) ||
+				    (!(oswpt.shipp->hotkey == setnum) && is_adding)) {
```
Goober5000: Ditto.

wookieejedi:
Thanks, incorporated all of those comments.

PR #2039, opened 2019-05-09
Unit tests for vecmat

The-E:
It occurs to me that one of the fundamental features of this, or any, game engine is its math library.
It also occurs to me that ours isn't tested, at all; This should be rectified. This branch represents a first step towards that goal; Contributions from people with more than my basic math skills are very welcome.

### DIFF
--- a/test/src/math/test_vecmat.cpp
+++ b/test/src/math/test_vecmat.cpp
@@ -467,7 +467,7 @@ TEST_F(VecmatTest, test_vm_vec_copy_normalize)
 		static_randvec_unnormalized(rand32(), &v1);
 		vBackup.xyz = v1.xyz;
 
-		auto magnitude      = vm_vec_copy_normalize(&v2, &v1);
+		auto magnitude = vm_vec_copy_normalize(&v2, &v1);
 		
 		ASSERT_FLOAT_EQ(v1.xyz.x, vBackup.xyz.x);
 		ASSERT_FLOAT_EQ(v1.xyz.y, vBackup.xyz.y);
--- a/test/src/math/test_vecmat.cpp
+++ b/test/src/math/test_vecmat.cpp
@@ -1,35 +1,353 @@
-#include <gtest/gtest.h>
-
+#include "math/staticrand.h"
+#include "math/vecmat.h"
 #include "util/FSTestFixture.h"
 
-#include "math/vecmat.h"
-#include "math/staticrand.h"
+#include <gtest/gtest.h>
 
 class VecmatTest : public test::FSTestFixture {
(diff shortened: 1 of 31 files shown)


test/src/math/test_vecmat.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+TEST_F(VecmatTest, test_vm_vec_add2)
+{
+	for (size_t loop = 0; loop < 1000; ++loop) {
+		vec3d v1, v2, v1backup;
+
+		static_randvec(rand32(), &v1);
+		static_randvec(rand32(), &v2);
+
+		vm_vec_copy_scale(&v1backup, &v1, 1.0f);
```
MageKing17: Here (and in the function below) you make a backup by using `vm_vec_copy_scale()` with a scaling factor of `1.0f`, but the other functions (starting when testing `vm_vec_scale()` below), you assign one vector's `.xyz` property to another instead; I think that being consistent should be of primary importance.

test/src/math/test_vecmat.cpp
```diff
@@ -0,0 +1,353 @@
...
+	void TearDown() override { test::FSTestFixture::TearDown(); }
+};
+
+TEST_F(VecmatTest, test_vm_vec_add)
+{
+	for (size_t loop = 0; loop < 1000; ++loop) {
+		vec3d v1, v2, v3;
+
+		static_randvec(rand32(), &v1);
+		static_randvec(rand32(), &v2);
```
CasparKielwein: The premium solution here would be to use rapidcheck (or another quickcheck clone)
https://github.com/emil-e/rapidcheck

Not necessary, but really cool

Goober5000:
I'll ask our resident math guru CP5670 to look at this.

test/src/math/test_vecmat.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+		ASSERT_FLOAT_EQ(vAvg.xyz.x, accX);
+		ASSERT_FLOAT_EQ(vAvg.xyz.y, accY);
+		ASSERT_FLOAT_EQ(vAvg.xyz.z, accZ);
+	}
+}
+
+TEST_F(VecmatTest, test_vm_vec_avg_n)
+{
+	for (size_t loop = 0; loop < 1000; ++loop) {
```
asarium: All the tests for the random functionality could be improved by using a value parameterized test since that would avoid having the `for` loop in all the test cases.

JohnAFernandez:
Do we still need this?  Does Appveyor do these types of checks now?

Issue #2046, opened 2019-05-16
Destroyed subsystems revert to undestroyed Model when ship dies


wookieejedi:
Destroyed subsystems revert back to the un-destroyed model when a ship dies. Here is an attachment image along with an attachment mission file. In the mission the panel subsystem is destroyed, then the ship is destroyed a second later. As the ship dies, the panel changes from it's destroyed model to un-destroyed model. 
![screen0316](https://user-images.githubusercontent.com/14077810/57824290-9c0e5580-7768-11e9-9814-0303fe4fe514.png)
![screen0317](https://user-images.githubusercontent.com/14077810/57824292-9f094600-7768-11e9-89d5-df0d51c0e225.png)

[sub_test.txt](https://github.com/scp-fs2open/fs2open.github.com/files/3185270/sub_test.txt)
(Just change the .txt fo .fs2)

Goober5000:
Can you check if this happened in retail too?

wookieejedi:
Noting here too, the bug is not present in retail.

Goober5000:
:thinking: 

Can you try testing some old builds?  Look in the official release threads:
http://www.hard-light.net/forums/index.php?topic=92181.0

A link to the previous release will be somewhere in the first post, in tiny print.

wookieejedi:
Sure! So from my first pass of tests, the bug is present in 3.7.4 but is not present in 3.7.2. Will examine the nightlies between those to see if I can tell which one changed it.

wookieejedi:
Hmm so it turns out most of the links to the old nightly builds are dead, so I am unable to test individual nightly builds. Using that time-frame (between April 23, 2015 and July 4, 2016) I searched mantis and found a possibly related ticket here: http://scp.indiegames.us/mantis/view.php?id=3164

And I did the same search in the closed PR section during that time range. I didn't find anything that stood out right away, but #88 (https://github.com/scp-fs2open/fs2open.github.com/pull/88) and #460 (https://github.com/scp-fs2open/fs2open.github.com/pull/460/files) seem like possible options? I'm not too sure what to do at this point.

Goober5000:
That's excellent detective work.  At first glance the two github issues probably aren't related, but the Mantis issue probably is.

The next step is to use `git bisect` to narrow down the precise change that introduced the bug.  Hopefully it will be a simple commit rather than a large commit like the new rendering code.

I can show you how to use `git bisect` on Discord.

[...]

PR #2047, opened 2019-05-19
Fix #206: destroyed subsystems reverting to un-destroyed model

### LINKED ISSUE #206: Order of precedence in obj_render_all()
Compiler warning:
```
object/objectsort.cpp: In function ‘void obj_render_all(void (*)(object*), bool*)’:
object/objectsort.cpp:289:32: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
    if( (obj->type == OBJ_SHIP) && Ships[obj->instance].shader_effect_active || obj->type == OBJ_FIREBALL )
```

wookieejedi:
This PR fixes issue #2046, which was a bug that destroyed subsystems reverted to their undestroyed models when a ship died. I tracked the bug to the commit 64ad950, which introduced the bug by removing the `ship_model_start(obj);` line in `ship_render()` (ship.cpp).

The reason removing this line caused the visual bug was because `model_set_instance()` was not called. Calling `model_set_instance()` properly set the `pm->blown_off values` (not just the `pmi->blown_off` values like `ship_model_update_instance()` does). The reason the `pm` values also need to be updated is because they are used by `shipfx_large_blowup_queue_render()` (in other words that function does not appear to use the `pmi` values that are updated in `ship_model_update_instance()`).

Thus, I simply added lines setting the `pm->blown_off` status to `ship_model_update_instance()`. Also in commit 3eb47c1 the line `model_clear_instance(sip->model_num);` was added to `ship_render()`. The function `model_clear_instance(sip->model_num);` sets the `pm->blown_off values` only based on `pm->submodel[i].is_damaged`. Thus, the `model_clear_instance(sip->model_num);` line should be moved to lower in the function, otherwise the `pm->blown_off values` get reset to default before `shipfx_large_blowup_queue_render()`.

Also moving the `model_clear_instance(sip->model_num);` to lower in the function meant that the lighting arc lines should also be moved, that way lighting arcs stay visible. 

Testing the new lines with multiple scenarios in game (quick restarts, sexp and manual destruction of subsystems, etc) confirmed the fix worked.

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -4742,8 +4742,10 @@ void model_update_instance(int model_instance_num, int sub_model_num, submodel_i
 	// Set the "blown out" flags
 	if ( flags[Ship::Subsystem_Flags::No_disappear] ) {
 		smi->blown_off = false;
+		sm->blown_off = 0;
 	} else {
 		smi->blown_off = sii->blown_off ? true : false;
+		sm->blown_off = sii->blown_off;
 	}
 
 	if ( smi->blown_off && !(flags[Ship::Subsystem_Flags::No_replace]) )	{
@@ -4751,12 +4753,15 @@ void model_update_instance(int model_instance_num, int sub_model_num, submodel_i
 			pmi->submodel[sm->my_replacement].blown_off = false;
 			pmi->submodel[sm->my_replacement].angs = sii->angs;
 			pmi->submodel[sm->my_replacement].prev_angs = sii->prev_angs;
+			pm->submodel[sm->my_replacement].blown_off = 0;
+			pm->submodel[sm->my_replacement].angs = sii->angs;
 		}
 	} else {
 		// If submodel isn't yet blown off and has a -destroyed replacement model, we prevent
 		// the replacement model from being drawn by marking it as having been blown off
 		if ( sm->my_replacement > -1 && sm->my_replacement != sub_model_num)	{
 			pmi->submodel[sm->my_replacement].blown_off = true;
+			pm->submodel[sm->my_replacement].blown_off = 1;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Goober5000:
Nice work!

This will need to be reviewed carefully.  I don't know how many coders are familiar with the submodel code.  @SamuelCho are you possibly available to check?

wookieejedi:
If it helps, all of the new code I added to `model_update_instance()` was from the `model_set_instance()` function.

wookieejedi:
Just your neighborhood friendly bump!

Goober5000:
I may have to take a look at this myself.  But next week is going to be super busy.

wookieejedi:
Thanks, no worries.

code/model/modelread.cpp
```diff
@@ -4742,21 +4742,26 @@ void model_update_instance(int model_instance_num, int sub_model_num, submodel_i
 	// Set the "blown out" flags
 	if ( flags[Ship::Subsystem_Flags::No_disappear] ) {
 		smi->blown_off = false;
+		sm->blown_off = 0;
```
asarium: My understanding of the model instance code is pretty limited but I think one of the main points was to avoid changing the actual `polymodel` data for a specific model so this is probably not correct.

wookieejedi: That seems logical. When I traced this bug back to its first occurrence, it appeared when the model instance code was added, so it appears that this functionality never got ported to using the new system. For my fix I simply added to `model_update_instance()` which was from the `model_set_instance()` function. I would be happy to have a fresh set of eyes on coming up with a better fix, especially since this is a visually glaring bug when ships explode and have all their un-destroyed models pop back into view.

wookieejedi:
Huge thanks to @SamuelCho for fixing this issue the correct way vai #2056. I will close this PR.

PR #2048, opened 2019-05-23
Add per ship class options for when AI sends "Help Me" message

wookieejedi:
This PR allows modders to set, on a per ship class basis, the fields for what hull and shield integrity the AI wing-men send the "Help" message at. One good use case is when ships have low hit points, thus now the modder can set AI can call for help sooner then at simply 30% hull.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -3582,7 +3582,7 @@ static int parse_ship_values(ship_info* sip, const bool is_template, const bool 
 			sip->ask_help_shield_percent = help_shield_val;
 		} else {
 			error_display(0,"Ask Help Shield Percent for ship class %s is %f. This value is not within range of 0-1.0."
-			              "Assuming default value of %i.", sip->name, help_shield_val, DEFAULT_ASK_HELP_SHIELD_PERCENT);
+			              "Assuming default value of %f.", sip->name, help_shield_val, DEFAULT_ASK_HELP_SHIELD_PERCENT);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


code/ship/ship.cpp
```diff
@@ -757,6 +757,11 @@ static int lightningtype_match(char *p)
 #define SHIP_MULTITEXT_LENGTH 4096
 #define DEFAULT_DELTA_BANK_CONST	0.5f
 
+#define ASK_HELP_SHIELD_PERCENT 0.1 // percent shields at which ship will ask for help
+#define ASK_HELP_HULL_PERCENT 0.3   // percent hull at which ship will ask for help
+#define AWACS_HELP_HULL_HI 0.75     // percent hull at which ship will ask for help
+#define AWACS_HELP_HULL_LOW 0.25    // percent hull at which ship will ask for help
```
asarium: The names of these macros could be improved to make it clear that these are the default values. Prefixing them with `DEFAULT_` could work.

Also, since you are already touching this, could you replace this with `const` values to make the code a little more modern.

code/ship/ship.cpp
```diff
@@ -3561,6 +3575,12 @@ static int parse_ship_values(ship_info* sip, const bool is_template, const bool
 	if(optional_string("$Scan range Capital:"))
 		stuff_float(&sip->scan_range_capital);
 
+	if (optional_string("$Ask Help Shield Percent:"))
+		stuff_float(&sip->ask_help_shield_percent);
+
+	if (optional_string("$Ask Help Hull Percent:"))
+		stuff_float(&sip->ask_help_hull_percent);
```
asarium: Both these cases need some error handling to make sure invalid values do not break the code.

code/ship/ship.cpp
wookieejedi: Will do!

code/ship/ship.cpp
wookieejedi: Will do.

wookieejedi:
Hmm it builds on my end, not entirely sure what's causing the build failures here...

asarium:
The error means that you are trying to assign a `double` literal to a `float` constant. In C++ if you write `1.0` that is actually a `double` literal. You want to use `1.0f` to tell the compiler that you would like to have a `float`.

wookieejedi:
Ah that makes sense, thanks!

PR #2050, opened 2019-06-07
Add system for reading flags from stdin

asarium:
This adds support for reading flags from the standard input stream
instead of from the command line. This allows more complex formats to be
used such as JSON which is implemented in this first version. This can
be used for implementing much more sophisticated ways of expressing
complex data structures such as the mod command line.


Goober5000:
Well I'm not nearly familiar enough with the CFile system to give this a good review.  Has it been thoroughly tested?

asarium:
I have tested the "old-style" system pretty thoroughly but the new style is hard to test without also having a launcher that supports it.

Goober5000 (dismissed):

Goober5000:
Closing since this hasn't been updated in four years and would likely require a rewrite.

PR #2052, opened 2019-06-10
Add scripting API support for model paths

asarium:
This adds a scripting API for querying the paths of a model.


MageKing17:
Why do you seem to have partially renamed/duplicated `CHA_ONWEAPONCREATED` to `CHA_ONWEAPONCREATE`?

asarium:
Probably a bad merge. I'll remove that.

PR #2053, opened 2019-06-21
Adds Table Option to Modify Arc Colors

wookieejedi:
This PR adds a table entry that allows modders to specify the RGBI values for the damage and emp arc colors (which were previosuly hardcoded).

This option adds the table entry to `Game_settings.tbl` under `#Graphic Settings` with the lines:
```
$EMP Arc Color:
	+Primary Color Option 1: rgb
	+Primary Color Option 2: rgb
	+Secondary Color Option 1: rgb

$Damage Arc Color:
	+Primary Color Option 1: rgb
	+Primary Color Option 2: rgb
	+Secondary Color Option 1: rgb
```

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -48,6 +48,24 @@ SCP_string Window_icon_path;
 bool Disable_built_in_translations;
 bool Weapon_shockwaves_respect_huge;
 bool Using_in_game_options;
+ubyte Arc_color_damage_p1_r = 64;
+ubyte Arc_color_damage_p1_g = 64;
+ubyte Arc_color_damage_p1_b = 225;
+ubyte Arc_color_damage_p2_r = 128;
+ubyte Arc_color_damage_p2_g = 128;
+ubyte Arc_color_damage_p2_b = 255;
+ubyte Arc_color_damage_s1_r = 200;
+ubyte Arc_color_damage_s1_g = 200;
+ubyte Arc_color_damage_s1_b = 255;
+ubyte Arc_color_emp_p1_r    = 64;
+ubyte Arc_color_emp_p1_g    = 64;
+ubyte Arc_color_emp_p1_b    = 5;
+ubyte Arc_color_emp_p2_r    = 128;
+ubyte Arc_color_emp_p2_g    = 128;
+ubyte Arc_color_emp_p2_b    = 10;
+ubyte Arc_color_emp_s1_r    = 255;
+ubyte Arc_color_emp_s1_g    = 255;
+ubyte Arc_color_emp_s1_b    = 10;
 
 void parse_mod_table(const char *filename)
 {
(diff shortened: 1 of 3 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -48,6 +48,24 @@ SCP_string Window_icon_path;
...
+ubyte Arc_color_damage_s1_b = 255;
+ubyte Arc_color_emp_p1_r    = 64;
+ubyte Arc_color_emp_p1_g    = 64;
+ubyte Arc_color_emp_p1_b    = 5;
+ubyte Arc_color_emp_p2_r    = 128;
+ubyte Arc_color_emp_p2_g    = 128;
+ubyte Arc_color_emp_p2_b    = 10;
+ubyte Arc_color_emp_s1_r    = 255;
+ubyte Arc_color_emp_s1_g    = 255;
+ubyte Arc_color_emp_s1_b    = 10;
```
asarium: The default values should be set in `mod_table_reset` to make the system easier to test.

code/mod_table/mod_table.cpp
```diff
@@ -48,6 +48,24 @@ SCP_string Window_icon_path;
 bool Disable_built_in_translations;
 bool Weapon_shockwaves_respect_huge;
 bool Using_in_game_options;
+ubyte Arc_color_damage_p1_r = 64;
```
asarium: I think it would be better to use a `std::tuple` for all these rgb value triples. That would make this variable declaration much shorter and keep the relevant values together in one variable.

code/mod_table/mod_table.cpp
```diff
@@ -267,6 +285,42 @@ void parse_mod_table(const char *filename)
 			mprintf(("Game Settings Table: $BMPMAN Slot Limit is deprecated and should be removed. It is not needed anymore.\n"));
 		}
 
+		if (optional_string("$EMP Arc Color:")) {
+			if (optional_string("+Primary Color Option 1:")) {
+				stuff_ubyte(&Arc_color_damage_p1_r);
+				stuff_ubyte(&Arc_color_damage_p1_g);
+				stuff_ubyte(&Arc_color_damage_p1_b);
```
asarium: I think it would be better to use `stuff_int_list` here since that already has an established parser implementation for parsing number lists. You will have to do some additional validation though to make sure that exactly 3 values are specified and that those values are in the correct range.

code/mod_table/mod_table.cpp
wookieejedi: Will do!

code/mod_table/mod_table.cpp
wookieejedi: Changed as requested.

code/mod_table/mod_table.cpp
wookieejedi: Couldn't figure out a efficient way of using `stuff_int_list` since the tuples have ubytes for values, and DahBlount stated using three values and `stuff_ubyte` would probably work best so I am sticking with that option for now. I did add in the error checks though, good idea!

code/mod_table/mod_table.cpp
```diff
@@ -267,6 +273,84 @@ void parse_mod_table(const char *filename)
 			mprintf(("Game Settings Table: $BMPMAN Slot Limit is deprecated and should be removed. It is not needed anymore.\n"));
 		}
 
+		if (optional_string("$EMP Arc Color:")) {
+			if (optional_string("+Primary Color Option 1:")) {
+				ubyte r, g, b;
+				stuff_ubyte(&r);
+				stuff_ubyte(&g);
+				stuff_ubyte(&b);
+				if ( (r >= 0 && r <= 255) && (g >= 0 && g <= 255) && (b >= 0 && b <= 255) ) {
```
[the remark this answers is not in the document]
wookieejedi: Oh I see. Good point, I will incorporate those changes.

wookieejedi:
Hmm is there a logical code related reason why travis keeps failing?

wookieejedi:
Hopefully good to go now.

PR #2054, opened 2019-06-24
properly indicate that player ships have departed at the end of a mission

Goober5000:
Evaluating `has-departed-delay "Alpha 1"` during a debriefing stage yields incorrect results in current FSO builds.  This is because the sexp short-circuits and evaluates to false if the ship hasn't arrived yet.  Unfortunately, the arrival check was flawed.

To handle respawns in multiplayer, player ships are not removed from the arrival queue.  A naive test for a yet-to-arrive ship by checking the arrival queue will therefore return false positives for player ships.  The solution is to not only check the list but also check whether the ship was in fact created.

I added a new function `mission_check_ship_yet_to_arrive` and substituted it for `mission_parse_get_arrival_ship` where appropriate.  I also checked all uses of `mission_parse_get_arrival_ship` that were *not* substituted to make sure that they were being used as they should.

Finally I removed the `ship_query_state` function because after implementing this fix, it was only used in one sexp and that instance was better handled by other functions.

This fixes the issue reported in Discord where Into the Lion's Den did not display any briefing stages.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -7362,6 +7362,7 @@ void ship_cleanup(int shipnum, int cleanup_mode)
 
 	ship *shipp = &Ships[shipnum];
 	object *objp = &Objects[shipp->objnum];
+	char *jumpnode_name = nullptr;
 
 	// add the information to the exited ship list
 	switch (cleanup_mode) {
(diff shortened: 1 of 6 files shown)


Drflash55:
Oh nice. Guess I should restore the original mission of The Lion's Den so that I can play it as intended.

code/ship/ship.cpp
```diff
@@ -7362,6 +7362,7 @@ void ship_cleanup(int shipnum, int cleanup_mode)
 
 	ship *shipp = &Ships[shipnum];
 	object *objp = &Objects[shipp->objnum];
+	char *jumpnode_name = nullptr;
```
asarium: Maybe this declaration could be moved a bit closer to the usages.

Goober5000: I considered that, but I put it near the beginning of the function because the other declarations were there too.  Plus having it there keeps the code flow in the middle of the function clean.

PR #2055, opened 2019-06-25
Bug fix: adds missing conditional when using weapon icons and 3D weapon selection

wookieejedi:
Previously if the 3D weapon flag was used it would always create a weapon icon and override any pre-existing icons the modder would specify. This PR fixes that by adding an extra set of conditionals, which specifies two things.
1) If a weapon has an icon file name then that icon should be used and not overwritten by an image of the pof.
2) If the weapon does not have an `$Anim:` but does have a model then that model should be used if the 3D_weapon flag is enabled.

### DIFF
--- a/code/missionui/missionweaponchoice.cpp
+++ b/code/missionui/missionweaponchoice.cpp
@@ -1276,7 +1276,7 @@ void wl_load_icons(int weapon_class)
 
 	multi_send_anti_timeout_ping();
 
-	if ( first_frame == -1 && icon->model_index == -1 || (Cmdline_weapon_choice_3d && !strlen(wip->anim_filename)) )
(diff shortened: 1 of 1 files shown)


wookieejedi:
Ah my bad. The second point check fixes the case of using custom icons but also wanting the weapon selection to use a 3D model when the 3D weapon select flag is enabled. If the second check was not added then the modder's icon will show properly (thanks to the first check fix) but there will be no 3D weapon selection effect. Thus the combined checks ensure that modders can show their custom icons and also see the weapon selection in 3D. If that is still a bit murky I would be happy to block out a more detailed explanation.

Goober5000:
No, that clears it up.  Your additional details also helped me realize that the first check affects the second (since `first_frame` is assigned a value).  Thanks. :+1:

PR #2056, opened 2019-06-26
Make sure various model rendering/collision code reference submodel instances instead of the prototype data

SamuelCho:
We should not be referencing bsp_info for submodel data specific to live game objects. When I migrated the rendering and collision code to the polymodel instance system, these stragglers fell between the cracks.

This should fix https://github.com/scp-fs2open/fs2open.github.com/issues/2046

### DIFF
--- a/code/model/modelcollide.cpp
+++ b/code/model/modelcollide.cpp
@@ -1325,8 +1325,14 @@ int model_collide(mc_info *mc_info_obj)
 		// Check all the the highest detail model polygons and subobjects for intersections
 
 		// Don't check it or its children if it is destroyed
-		if (!Mc_pm->submodel[Mc_pm->detail[0]].blown_off)	{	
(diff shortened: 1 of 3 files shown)


wookieejedi:
@SamuelCho Unfortunately it appears the blown-off bug still occurs even with this fix incorporated. Might it have to do with the placement of the `model_clear_instance(sip->model_num)` line in `ship.cpp`?

SamuelCho:
That's a good lead. Upon a cursory glance, it looks like the `render_info` in `shipfx_queue_render_ship_halves_and_debris()` we send into `model_queue_render()` doesn't have an object number associated with it. 

Maybe see what happens if you add `render_info.set_object_number(shipp->objnum);` after `render_info.set_replacement_textures(shipp->ship_replacement_textures);` on line 1664?

wookieejedi:
Ah that did the trick, thanks so much! I will submit a PR.

PR #2058, opened 2019-07-03
Coverity 1093382

FSCyborg:
Coverity spotted this dead code.  Looks like the original coder forgot that they had checked for !p already or was a case of mindless copy-pasting code..

Because
pcode = strchr(p,' ')+1;
keeps pcode from being null, it looks like it is safe to remove.

### DIFF
--- a/code/inetfile/chttpget.cpp
+++ b/code/inetfile/chttpget.cpp
@@ -231,7 +231,7 @@ void ChttpGet::WorkerThread()
 			return;
 
 		}
-		pcode++
+		pcode++;
 		pcode[3] = '\0';
 		irsp = atoi(pcode);
 
--- a/code/inetfile/chttpget.cpp
+++ b/code/inetfile/chttpget.cpp
@@ -223,7 +223,15 @@ void ChttpGet::WorkerThread()
 	if(strnicmp("HTTP/",p,5)==0)
 	{
 		char *pcode;
-		pcode = strchr(p,' ')+1;
+		pcode = strchr(p,' ');
+		if(pcode == null)
+		{
+			m_State = HTTP_STATE_UNKNOWN_ERROR;	
+			fclose(LOCALFILE);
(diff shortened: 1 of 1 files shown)


Goober5000:
It's testing for a potential error.  The proper thing to do is fix the test, not remove it.

Looks like all that's necessary is to first check for `nullptr`, then do the +1 after the `if` block.

FSCyborg:
Ok, just to help me understand, why is the !p check insufficient in line 222?  Do we always have to check a pointer after creating it, even if the data it is pointing to has already been checked?

Really just trying to learn here.

Goober5000:
Because the `!p` check is not the same as the `!pcode` check.  The second check is searching for any spaces (the `' '` character) within the `p` string.  If there are no spaces found, `strchr` will return `nullptr`.

FSCyborg:
Better?  (DahBlount also suggested changing `!pcode` to `pcode == null`)

code/inetfile/chttpget.cpp
```diff
@@ -223,14 +223,15 @@ void ChttpGet::WorkerThread()
 	if(strnicmp("HTTP/",p,5)==0)
 	{
 		char *pcode;
-		pcode = strchr(p,' ')+1;
-		if(!pcode)
+		pcode = strchr(p,' ');
+		if(pcode == null)
```
asarium: FYI, `null` is not valid in C++. The correct keyword is `nullptr`.

FSCyborg: Thank you

Goober5000:
Looks good now.  But you need a semicolon after `pcode++`.

FSCyborg:
Thanks for your patience. Will fix that.

Goober5000:
Hmm, there are still some build errors that need to be resolved before merging.  Looks like those may be the fault of Travis though.

Issue #1987, opened 2019-02-10
Thrusters rendering in F3 Lab


Nyctaeus:
In current version of F3 lab, thrusters for ships render for only first ship I select. If I change selected ship, no thruster will render for newly selected ship and any further ones,

https://i.imgur.com/suGs3WX.jpg

Mithi83:
Debugging this issue reveals the following behavior:

1. Loading of a new ship is done in `labviewer_change_ship`, which first sets `Lab_selected_index` to the new ship index and then calls labviewer_change_ship_lod, which later calls `labviewer_change_model` with a `sel_index` of the new `ship_index`.
2. In `labviewer_change_model` we always have the situation, that `sel_index == Lab_selected_index` and thus `change_model` is false for all but the first call to `labviewer_change_model`. I'm not sure why it even shows another model at all when switching to another ship.
3. Looking at the code in labviewer_change_model I would guess that the assignment of `Lab_selected_index = sel_index;` is the intended place where the update of `Lab_selected_index` should happen. Thus I commented out the assignment of `Lab_selected_index` in `labviewer_change_ship`. For me this seems to fix the problem, but I'm not sure if it has any side effects.

Maybe someone familiar with this code can take it from here.

On a side note: I can't imagine that passing ship_index as lod to labviewer_change_model is correct. There might be another problem there with unknown consequences.

z64555:
@Nyctaeus Did the fix work?

FSCyborg:
I just tried it, and it worked using the July 29th build.

PR #2060, opened 2019-07-07
fix bug #1987: thruster rendering in lab

Mithi83:
This fixes bug #1987. A more detailed analysis and reasoning is given there.

### DIFF
--- a/code/lab/lab.cpp
+++ b/code/lab/lab.cpp
@@ -1592,7 +1592,6 @@ void labviewer_change_ship_lod(Tree* caller)
 
 void labviewer_change_ship(Tree* caller)
 {
-	Lab_selected_index       = caller->GetSelectedItem()->GetData();
 	Lab_last_selected_object = caller->GetSelectedItem();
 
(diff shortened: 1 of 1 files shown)


FSCyborg:
Was this tested to ensure that it works?

Mithi83:
For me it works, but since I implemented the fix maybe someone else should take a look that nothing else is broken. Then again this "only" touches the lab and does not interfere with the actual gameplay.

Steps to reproduce the bug in case you want to test it yourself:
1. Go to the Lab (press F3 in the Main Hall)
2. In the menu open the "Ship" and "Render Options" windows.
3. In the "Render Options", activate "Show Thrusters"
4. In "Ship" select a ship type, I used "GTC Fenris". Use the mouse to rotate the ship so that you can see the thrusters. This always works, with or without my fix.
5.  In "Ship" select a different ship type, I used "GTM Hippocrates". Then take a look at the thrusters again. Without the fix, the thusters don't show, with the fix they show.

Issue #1791, opened 2018-07-11
'impossible' crash, laser_flag assert in aiturret.cpp:1222, possible memory management issue.


EatThePath:
[assert_fs2_open_log.txt](https://github.com/scp-fs2open/fs2open.github.com/files/2185889/assert_fs2_open_log.txt)
[assert_fs2_open_clipboard.txt](https://github.com/scp-fs2open/fs2open.github.com/files/2185890/assert_fs2_open_clipboard.txt)

I am experiencing an assert crash in very narrow circumstances. With a certain mod pack combination, on a certain mission, played a certain way, on debug builds, I get an assert error box with contents that show up in the fs2_open.log, and then two more error popups that don't appear in the log. I've attached both the log, and a paste of just the error messages.

In a discussion about this issue on Discord I've been informed that it should not be possible for this assert to be reached, and that thus nothing in the mod pack I'm using could logically be causing it. Having no lead whatsoever that makes sense, I'm just going to dump all the information I have on you about what is going on and hope something useful is contained within. I understand Axem encountered the issue before as well in different circumstances, but don't know any details there.

I'm working on a modpack that modifies Blue Planet. The mod adds weapons, ships, sounds, textures, and models, and also modifies many of the same from Blue Planet. The mission file in question is an exact copy of a blue planet mission file except that the player's default and available weapons/fighters, though the ships and effects in play in the mission are all modified themselves.

The mission is BP-04, "Journey of a Thousand Miles". At mission start there is a capital ship on team Unknown(purple) some distance away three alpha wing fighters, and there's also a jump node. The mission is otherwise empty. All three fighters are affected by turret-protect-ship "beam" and turret-protect-ship "laser". After some time the capship gets hit with change-iff "Hostile" and the player is tasked to shoot bits of it, and then shivans start showing up in the mission to fight everyone.

The moment when the ship switches to hostile is when the crash occurs. It only occurs if the player is close to the ship, if they keep their distance until after the switch nothing bad happens.

The crash only occurs with the addon mod active, I have not been able to make it happen in regular Blue Planet with the same play that reliably causes it in the addon. Problems also only manifest in debug builds, in non-debug no issues occur when I carry out the same test procedure.

I have observed this in 3.8.1.20180710_2dc2c06, 3.8.1.20180711_ccc727a, and Multilock-3.8.1-20180708. Probably a few earlier builds when I was intially tripping over it, but I didn't record those enough.

That's all I've got for now.

[...]

PR #2061, opened 2019-07-11
 Fix? Issue #1791

FSCyborg:
Requesting testing!  Looks like the simplest fix for this issue could be to add the checks for the flags to both branches of the involved if statement in aiturret.cpp .

#1791


FSCyborg:
Sorry about the messy commit history.  Will try to avoid that in the future.

chief1983:
You can clean up the commit history and force push over the branch to fix that if you want.  I'd recommend it even. Don't like reverts in a commit history.  Easiest way would be to cherry pick the last commit hash into a new branch based on master.

FSCyborg:
Github is pretty much an obstacle course for newer coders. :|

I'll try to manage that.

Goober5000:
Looks reasonable, but you should make sure that the commented `nprintf` remains next to the line it describes.

PR #2062, opened 2019-07-13
Aiturret Cleanup Take 2

FSCyborg:
While probably not a full fix to #1791 , this should clean up the portion of the code that looks the most suspicious, by making the checks for weapon flags for the whole block instead of just part.  Also attempts to make it more readable and one or two more checks.  Two places I am not removing what appears to be nonsense because I am trying not to change AI behavior.

Hopefully, all these changes are just changing the checking order and adding checks, and should not affect the actual AI.  I apologize in advance for any mistakes, since this ended up being much more involved than I thought.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1179,7 +1179,7 @@ int find_turret_enemy(ship_subsys *turret_subsys, int objnum, vec3d *tpos, vec3d
 			if (Ship_types[sip->class_type].flags[Ship::Type_Info_Flags::Turret_tgt_ship_tgt]) {
 
 				// Turrets and ships should not be targeting the wrong team.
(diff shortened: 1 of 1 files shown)


FSCyborg:
I am certain I found the bug and the most recent revision fixes it.  I recommend keeping the other changes, since without them, "$Turrets prioritize ship target:" will override beam protect, etc.

But I am, of course open to making changes or reverting.  And if I am honest, I think we probably should be moving these checks to valid_turret_enemy(), and just adding a call to it in all the pathways. (there are three)

Even though I know where the bug is, I do not know what combination of IFF status/timing/flags in particular causes it to raise its head, so I cannot *verify* that it fixes the issue.  But I have looked all over this code about a dozen times and can find no other possible cause.

FSCyborg:
Ok.  Doesn't look like I can edit the repository the pull request is coming from, so one more try here.

PR #2063, opened 2019-07-13
Allow positioning sexps to work on ships yet to arrive

Goober5000:
This adds `set-object-position` functionality for ships and wingmen still on the arrival list.  Requested by JSRNerdo.

EDIT: Also added `set-object-orientation`, `set-object-facing`, and `set-object-facing-object`.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -7096,6 +7096,12 @@ void sexp_set_object_position(int n)
 			break;
 		}
 
+		case OSWPT_TYPE_PARSE_OBJECT:
+		{
+			oswpt.p_objp->pos = target_vec;
+			break;
+		}
+
 		case OSWPT_TYPE_WAYPOINT:
 		{
 			oswpt.objp->pos = target_vec;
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -7134,6 +7140,39 @@ void sexp_set_object_position(int n)
...
 			break;
 		}
+
+		case OSWPT_TYPE_WING_NOT_PRESENT:
+		{
+			// search the arrival list for the wing leader and move him first
+			bool found = false;
+			for (p_object *p_objp = GET_FIRST(&Ship_arrival_list); p_objp != END_OF_LIST(&Ship_arrival_list); p_objp = GET_NEXT(p_objp))
+			{
+				if (p_objp->wingnum == WING_INDEX(oswpt.wingp) && p_objp->pos_in_wing == 0)
```
FSCyborg: You're looping through the same objects twice in this section.  Would it save a little cpu time to store the other wingmen you could theoretically find in the first loop and start the second loop where the first loop left off? (not sure if this second part is possbile)

Most instances, the wing would probably exist, no?

This is my first instinct coming from the molasses that is VBA macros.

Goober5000: In FreeSpace missions, objects are not guaranteed to be listed in wing order.  You might end up with Delta 2, 3, 4, and 1, in which case once you found 1 you'd have to go back to the beginning for the other wingmen.

The best alternative option is to store the wing indexes in an array, and then just loop through the new array in the second block.  I'm not sure it's worth switching to this method though.  You're weighing creating a new array (which involves a negligible amount of extra space) with looping through the arrival list (which involves a negligible amount of extra time).

FSCyborg: It probably wouldn't effect much unless someone tried to combine this with every-time or decided it was a great idea to put on repeat count with no interval delay.

Goober5000:
It's been tested, so I know it works.

PR #2065, opened 2019-07-19
Coverity 1093370 

FSCyborg:
Copy paste error, I can think of no reason to compare Max y and min z


FSCyborg:
This is not what I wanted to do.  Please hold.

PR #2066, opened 2019-07-19
Coverity 1093370

FSCyborg:
Copy paste error, I can think of no reason to compare Max y and Min z


FSCyborg:
Again, not what I was trying to do.

PR #2068, opened 2019-07-20
Fix#1791 ChangeIFF & Laser_Flag

FSCyborg:
Extremely rare bug caused by use of multiple protect from weapon flags. One of the Ai-Turret functions was setting only one of the flags because "else if" was typed instead of "if".

Possible improvements to related files to come in *other* PR's.


FSCyborg:
Edits, in other PR's, that is.

PR #2069, opened 2019-07-20
Coverity 1093370

FSCyborg:
Copy paste error, I can think of no reason to compare Max y and Min z, and the code afterward is using it as just a factor in multiplication.  My guess is that it would just cause unpredictable results, especially at coordinates with high y or z values.


Goober5000:
Interesting.  How long has the code been like that?

FSCyborg:
Just checked, and it has been like this since retail.

Goober5000:
Huh.  Well, we'll see what happens.

PR #2070, opened 2019-07-20
Coverity 1093387

FSCyborg:
The assert already checks for the file being Null, so why check again?

In case anyone decides to remove the assert, let's just comment the check out.  (But in my humble opinion, it's a big deal if we're trying to open a jpeg that doesn't exist.)

Both checks were introduced during a bugfix early in the SCP.


Goober5000:
In this *specific case*, you can remove both the check and the Assert, since because of the code flow the variable is guaranteed not to be NULL at this point.

However, *in general*, it's a good idea to have both an Assert and an if() block if the condition has a reasonable chance of occurring.  This is because Asserts are only checked in Debug builds.

FSCyborg:
Ah.  Well let's fix this, then.

PR #2071, opened 2019-07-21
Fix#2067

FSCyborg:
Original behavior in Fred was accidentally removed when upgrading the FSO mission parser to keep it from crashing FSO from a bad sexp.  

Returning false is only necessary in FSO since in FRED, we can fix the error once we know about it.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -5852,8 +5852,10 @@ bool post_process_mission()
 				sprintf(error_msg, "%s.\n\nIn sexpression: %s\n(Error appears to be: %s)", sexp_error_message(result), sexp_str.c_str(), Sexp_nodes[bad_node].text);
 				Warning(LOCATION, "%s", error_msg.c_str());
 
-				// syntax errors are unrecoverable, so abort
-				return false;
+				// syntax errors are recoverable in Fred but not FS
+				if (!Fred_running) {
+					return false;
+				}
 			}
 		}
 	}


FSCyborg:
It's this one: https://github.com/scp-fs2open/fs2open.github.com/commit/47e9c01ce6737f39e66ec47efda44a5a08865a6c

Goober5000:
Okay, good -- there are no loose ends.  Could you change the comment to "syntax errors are unrecoverable in FS, so abort" ?

FSCyborg:
Sure, no problem

Goober5000:
That's not precisely the sentence I recommended, and it looks more confusing.  How about this?

`syntax errors are recoverable in FRED but not in FS`

Goober5000:
It doesn't look like your latest commit changed anything, unless GitHub is not showing me the update.  Currently, the comment I see is

`// syntax errors are unrecoverable in FS but not in Fred, so abort`

which is still the more confusing comment.

FSCyborg:
Sorry, not sure what happened.

FSCyborg:
Please hold off until I can squash the commit.

Goober5000:
Ok.

The comment looks good, but you still need to remove the "so abort" part.

FSCyborg:
I read what you wrote too quickly.

Goober5000:
Looks good now.  Awaiting squash.

Goober5000:
@FSCyborg Are you still planning to squash the commits?

FSCyborg:
Yes, I could only do it from a computer, which I did not have while I was out of town.

PR #2073, opened 2019-07-29
Fix#1916

FSCyborg:
"Cloning" failure was actually being caused by the pilot verification function resetting all pilot info and then not loading controls from the file to replace them.

The Control_Config array would all be set to -1 and then would get written into the file the next time the player was saved, making it seem that the controls were not being copied.

The original and current code would reset controls to defaults when a pilot was created.  And it would just let the old controls remain in case of a clone.  The above problem would occur just as you hit accept on the bottom of the baracks screen.


FSCyborg:
Whoops, wrong issue number.

PR #2074, opened 2019-07-29
Fix#1974

FSCyborg:
"Cloning" failure was actually being caused by the pilot verification function resetting all pilot info and then not loading controls from the file to replace them.

The Control_Config array would all be set to -1 and then would get written into the file the next time the player was saved, making it seem that the controls were not being copied.

The original and current code would reset controls to defaults when a pilot was created.  And it would just let the old controls remain in case of a clone.  The above problem would occur just as you hit accept on the bottom of the baracks screen.

### DIFF
--- a/code/pilotfile/plr.cpp
+++ b/code/pilotfile/plr.cpp
@@ -682,7 +682,7 @@ void pilotfile::plr_write_settings()
 	handler->endSectionWrite();
 }
 
-void pilotfile::plr_reset_data()
+void pilotfile::plr_reset_data(bool reset_all)
 {
 	// internals
 	m_have_flags = false;
@@ -691,7 +691,11 @@ void pilotfile::plr_reset_data()
 	m_data_invalid = false;
 
 	// set all the entries in the control config arrays to -1 (undefined)
-	control_config_clear();
+	// Cyborg17 - clear_all is false when we are just trying to verify the file, but true when we load the whole thing.
+	// Otherwise, we can end up getting rid of a pilot's controls by accident.
+	if (reset_all) {
+		control_config_clear();
+	}
 
 	// init stats
 	p->stats.init();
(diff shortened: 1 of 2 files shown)


FSCyborg:
I tested this on my own machine, checking that cloning and creating from barracks and player select menu would both work.

Goober5000:
Since `plr_reset_data` takes a boolean argument, Travis wants you to use boolean literals.  So change `1` to `true` and `0` to `false`.

code/pilotfile/plr.cpp
```diff
@@ -691,7 +691,9 @@ void pilotfile::plr_reset_data()
 	m_data_invalid = false;
 
 	// set all the entries in the control config arrays to -1 (undefined)
-	control_config_clear();
+	// Cyborg17 - clear_all is false when we are just trying to verify the file, but true when we load the whole thing.
+	// Otherwise, we can end up getting rid of a pilot's controls by accident.
+	if ( reset_all ) control_config_clear();
```
Goober5000: For readability purposes, put `control_config_clear` on the following line.

code/pilotfile/plr.cpp
```diff
@@ -797,7 +799,8 @@ bool pilotfile::load_player(const char* callsign, player* _p, bool force_binary)
 
 	mprintf(("PLR => Loading '%s' with version %d...\n", filename.c_str(), version));
 
-	plr_reset_data();
+	//1 resets everything, 0 sets up file verify.
```
Goober5000: Since the function takes a boolean argument, rather than an integer, you should use true and false instead of 1 and 0.

FSCyborg: Got it, wasn't sure what was acceptable.  Fixes on the way.

Issue #1916, opened 2018-10-26
Turret Targeting Control Frequently Does Not Target New Turret


Goatman217:
"Target Next Live Turret" and "Target Previous Live Turret" keys will break aspect lock whenever tapped, but will often not change which turret is targeted.  There seems to be an emphasis on which turrets are either closest to the ship, or closest to the crosshairs, which inconsistently trumps the "next-in-list" functionality.  Can we make it so that if the target is not changed, aspect lock is not broken, or force a different turret to be targeted?  Ideally non-random.

To replicate:
1. Play a mission with a ship that has a large number of turrets over a large area, such as a Typhon or a Demon
2. Use "Target Next Turret" repeatedly, while flying around and changing orientation to observe differences in behavior

I haven't been able to discern much of a consistent pattern, which is why I think ship orientation or camera orientation plays a part.  Perhaps this is discernible in the code?

MageKing17:
"Target Next Live Turret" will target the closest/in-crosshair turret if either the current targeted subsystem is not a turret, or it's been at least 1 second since the last time the button was pressed (it will only target the turret you're pointing at if your reticle is within 3 degrees of one). A check could probably be added so that if the turret selected is the same turret already targeted, the subsystem targeting code be skipped so that the aspect lock doesn't reset.

FSCyborg:
This one next.  Doesn't this happen for all targeting controls?  After taking a look to see what's going on, maybe we can keep all targeting from resetting aspect lock for the same target or subsystem.

FSCyborg:
That pull request was put on the wrong branch, apologies.

FSCyborg:
This is by far the easiest change I have made in the code.  There only needed to be one change because there is only one place where changing the subsystem target resets lock.  I did a quick test and I'm submitting a pull request now.

Goober5000:
Fixed via PR #2075

PR #2075, opened 2019-07-30
"Fix" 1916

FSCyborg:
Adds a check to see if the subsystem is the same before resetting aspect lock.

More of an enhancement to expected behavior than a bugfix. 

#1916


FSCyborg:
Forgot to mention that I have already tested this, and it happens to be a one line fix because there is only one place that the code resets target lock when changing the target subsystem. (As far as I can tell)

Goober5000:
:+1:

Issue #1848, opened 2018-09-02
$Rearm Rate: does not seem to work for ballistic primaries


Kingspoon:
Primary ballistic weapons just all reload at the exact same rate (a very fast rate)
And in addition to that, no rearming sound is played when primary ballistic weapons are being reloaded by the supply ship.

FrikgFeek:
From my own testing the ballistic rearming sound plays as long as there's an entry for it in sounds.tbl, $rearm rate seems to have changed behaviour recently. From testing it seems to control how many 100s of rounds are loaded per second. Changing the Vulcan's $rearm rate from 20 to 0.5 results in 100 rounds being loaded every 2 seconds or so.

I'm fairly certain this is not intentional.

Example here:

https://www.youtube.com/watch?v=fShIGeD7m9c

Kingspoon:
That is... _Interesting_ behavior. And that explains why in every test case I tried they all seemed to reload at the same rate. Which they _kind_ of do, except its the same amount, at varying rates. 
For primary weapons with low ammo counts this means it doesn't matter what rearm rate you set them as, they will be refilled instantly.

So as it turns out the wiki entry for $rearm rate is just blatantly false:
"Defines the rate the weapons are loaded to the ship
    Syntax: Float, missiles loaded per second"
You don't define how many missiles are loaded per second. The amount that is loaded is always 4 for missiles and 100 for primaries. What you define is how fast those set amounts are loaded.

This is not really ideal behavior. And means pretty much all of my weapons.tbl $rearm rate entries are wrong because of false wiki information and counterintuitive behavior.

As for the sound effect not playing. I see/hear it playing on your video. And I do have the sounds defined in sounds.tbl, but play on my end, they do not.

Kingspoon:
Aaaand I figured out why the sounds weren't playing. I defined the entries in the new sound code way, which doesn't work for some reason. It only works if they're defined in the old retail method. Probably because its hardcoded nonsense. Either way, that's a seperate bug I guess.

Goober5000:
If the ballistic rearm sounds are not defined, the code is supposed to play the standard rearm sounds.

FSCyborg:
Some mods probably depend on this type of behavior, since it looks like it has been around since 2002 or 2003, but we can probably add another option to specify how much ballistic ammo is loaded for a given weapon at a time, with the default being 100.  I *think* I can handle it.

Is "$Rearm Ballistic Batch:" a reasonable name?

[...]

PR #2076, opened 2019-08-05
For Issue #1848 - Add option to change default rearm values

FSCyborg:
Although not technically a bug, several mods have been misled by the description of the "$Rearm Rate:" weapons table setting, and specifying an initial ammo value that is less than 100 for ballistic primaries would force the ballistic ammo to be loaded all at once, with "$Rearm Rate:" just specifying how long that one load would take.

This allows modders to use more realistic behavior for things like anti-turret ballistic primaries, where initial ammo could be in the thirties instead of the thousands, so that any number above zero can be added per load.  

Paired with this pull request, I am going to update the wiki entry for "$Rearm Rate:" to reflect what it actually does (tells the engine how many times per second to add munitions)

#1848

### DIFF
--- a/code/weapon/weapon.h
+++ b/code/weapon/weapon.h
@@ -293,6 +293,7 @@ typedef struct weapon_info {
 	float turn_time;
 	float	cargo_size;							// cargo space taken up by individual weapon (missiles only)
 	float rearm_rate;							// rate per second at which secondary weapons are loaded during rearming
+	int		reloaded_per_batch;				    // number of munitions rearmed per batch
 	float	weapon_range;						// max range weapon can be effectively fired.  (May be less than life * speed)
 
     // spawn weapons
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -215,6 +215,10 @@ int		Weapon_impact_timer;			// timer, initialized at start of each mission
 // range. Check the comment in weapon_set_tracking_info() for more details
 #define LOCKED_HOMING_EXTENDED_LIFE_FACTOR			1.2f
 
+// default number of missiles or bullets rearmed per load sound during rearm
+#define REARM_NUM_MISSILES_PER_BATCH 4              
+#define REARM_NUM_BALLISTIC_PRIMARIES_PER_BATCH 100 
+
 extern int compute_num_homing_objects(object *target_objp);
 
 extern void fs2netd_add_table_validation(const char *tblname);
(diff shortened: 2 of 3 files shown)


code/weapon/weapons.cpp
```diff
@@ -1480,7 +1480,33 @@ int parse_weapon(int subtype, bool replace, const char *filename)
...
+			if (subtype == WP_MISSILE) {
+				wip->reloaded_per_batch = REARM_NUM_MISSILES_PER_BATCH;
+				Warning(LOCATION, "Rearm batch amount of 0 or less in weapon %s, setting to the default 4 for missiles.", wip->name);
+			} else {
+				wip->reloaded_per_batch = REARM_NUM_BALLISTIC_PRIMARIES_PER_BATCH;
+				Warning(LOCATION, "Rearm batch amount of 0 or less in weapon %s, setting to the default 100 for ballistic primaries.", wip->name);
+			}
+		}
+	//This setting requires a default if not specified, otherwise, rearming will break. - Cyborg17
+	} else {
```
MageKing17: This should be `} else if (first_time) {`, or modular tables would have to redefine it every time just to avoid it getting reset to default. Similarly, the above "invalid batch amount" branch will overwrite a (valid) previously-defined rearm ammo increment. I'd suggest, instead, setting the default value to -1 in `weapon_info::reset()` and then do a post-parse "cleanup" of turning that -1 into `REARM_NUM_MISSILES_PER_BATCH` or `REARM_NUM_BALLISTIC_PRIMARIES_PER_BATCH` instead.

FSCyborg: I think this is what you were suggesting.

code/weapon/weapons.cpp
```diff
@@ -1480,7 +1480,30 @@ int parse_weapon(int subtype, bool replace, const char *filename)
...
+			wip->reloaded_per_batch = reloaded_each_batch;
+		} else if (wip->reloaded_per_batch != -1) {
+			Warning(LOCATION, "Rearm increment amount of 0 or less in weapon %s, keeping at previously defined value.",
+			        wip->name);
+		} else {
+			Warning(LOCATION, "Rearm increment amount of 0 or less in weapon %s, setting to the default, 4 for missiles or 100 for ballistic primaries.", wip->name);
+		}
+	}
+	// This setting requires a default if not specified, otherwise, rearming will break. - Cyborg17
+	if (wip->reloaded_per_batch == -1) {
```
MageKing17: Change this to `else if (first_time)` instead.

MageKing17: Nevermind, then you'd have to duplicate code to set the defaults if an invalid value was specified.

code/weapon/weapon.h
```diff
@@ -209,6 +209,9 @@ typedef struct spawn_weapon_info
 
 #define MAX_SPAWN_TYPES_PER_WEAPON 5
 
+#define REARM_NUM_MISSILES_PER_BATCH 4              // default number of missiles rearmed per load sound
+#define REARM_NUM_BALLISTIC_PRIMARIES_PER_BATCH 100 // default number of bullets rearmed in per load sound
```
asarium: These values are not needed in the header and could be moved to the source file.

FSCyborg: Totally missed that you requested this.  Is this a general principle I should follow?  If not needed by other source files, keep it out of the header?

asarium: It's generally a good idea to keep the header as clean as possible. You will not always see that principle followed in the FSO code base but it's a good thing to do where possible.

PR #2077, opened 2019-08-08
Fix #1433

### LINKED ISSUE #1433: Pausing and Unpausing the Game in Subspace stops the Subspace Sound
Recreate by playing a mission that takes place in subspace and pause and unpause the game.

FSCyborg:
For Issue #1433 Susbace ambience does not restart from escape menu. According to Mageking, this is caused by the escape menu strangely circumventing state changes.

Instead of changing how the escape menu handles state changes and risking regression, we are going to enable some code in Release versions that :V: left in debug builds from testing in freespace.cpp. This will let FSO just do a quick "Is susbpace happening" check every frame.


Goober5000:
This is the wrong approach.  See my comments on #1433.

PR #2078, opened 2019-08-09
Fix 1433 Take two

FSCyborg:
Have FSO check for subspace and start the loop up again when subspace is being used.

### DIFF
--- a/code/io/keycontrol.cpp
+++ b/code/io/keycontrol.cpp
@@ -1422,6 +1422,10 @@ void process_player_ship_keys(int k)
 	}
 }
 
+//Cyborg17 - Linking this function to restart sound if needed. This is to avoid having to
+//check for subspace every frame.
+extern void game_start_subspace_ambient_sound();
+
 /**
  * Handler for when player hits 'ESC' during the game
  */
@@ -1448,6 +1452,10 @@ void game_do_end_mission_popup()
 			break;
 
 		case 2:
+			// Cyborg17 - best place to check for this *one* looping sound.
+			if (Game_subspace_effect) {
+				game_start_subspace_ambient_sound();
+			}
 			gameseq_post_event(GS_EVENT_ENTER_GAME);
 			break;
 


FSCyborg:
Only 90% sure I tested the subspace sound correctly since I am having trouble making release builds.

FSCyborg:
Confirmed just now, when not an insomnia zombie, that this does fix the error.

Goober5000:
Commenting to link to #1433

code/io/keycontrol.cpp
```diff
@@ -60,6 +60,7 @@
 #include "cmdline/cmdline.h"
 #include "object/objectshield.h"
 
+
```
Goober5000: Looks like an unnecessary whitespace change was introduced here.

code/io/keycontrol.cpp
```diff
@@ -1448,6 +1453,10 @@ void game_do_end_mission_popup()
 			break;
 
 		case 2:
+			// Cyborg17 - best place to check for this *one* looping sound.
```
Goober5000: As I mentioned in #1433, this should be near the bottom of the function.  As a matter of fact, since you've confirmed the problem affects more than one sound, you should make a `game_restart_looped_sounds` function and add this code to that function.  Then, when you do the engine wash fix, you can put that in the same place.

FSCyborg: I don't see the reasoning for starting the subspace sound when the mission is being ended.  It should only be restarted when the mission is going to be resumed, although the code will stop it again later, it's better for us not to have to depend on that or make the engine more inefficient in that way.

Unless there's something else I am not understanding?

Goober5000: I agree, if the game is being ended then there isn't really a need to restart the sound.  However, the game isn't ended *immediately* -- it waits for the state transition.

What it boils down to is that `game_start_time()` is at the bottom, so the sound restart should also be at the bottom.  It makes the function symmetrical with the time and sound functions up top.

FSCyborg: From a human perspective, it does end immediately.  The game takes what, one, two frames to change states?  This is an imperceptible amount of time, and instead of allowing that transition to happen naturally, we are going to _call attention to it_ by restarting the sound?

For the user, the mission has ended as soon as they press the button, and it is most natural for them to assume that the game is spending all of its time tearing everything down.  In fact, if the sound restarts, _it will feel unpolished_ because then the player is going to start thinking about the code themselves, asking "why did the sound come back?"  and breaking immersion..

Symmetry should not be a distinct objective in code writing.  Literally every other concern should come first.  This function is not hard to read.  The way that this patch is now even makes it easier to read because now it's pretty obvious which choice restarts the game.

If you can't have it any other way, maybe you should just submit your own pr.

Goober5000: Fair enough.

Goober5000:
This was *not* fixed, apparently... the code was put under the wrong case statement. My fix for #2099 includes a fix for this.

PR #2082, opened 2019-08-19
add workarounds for template parameter packs

### LINKED ISSUE #2081: Bad memory allocation causing instant CTD on load with certain sexps
With nightlies on and after July 29 (with commit #2035 ) Surrender Belisarius will crash almost instantly on load. The amount of details I can give are limited but here's what I got...

Callstack: https://pastebin.com/DF3AhxYr

And when looking around in whats stored in memory, I saw some references around the Iota 2 Docking event. I don't know which sexp exactly is causing it though.

And this ONLY happens with release builds. Any fast debug or debug will not cause a crash.

So far this is confirmed on 2 other windows systems. Unknown if it crashes on Mac or Linux.

Goober5000:
This workaround is supposed to fix #2081.  However, when I tested it just now, it crashed anyway. :man_shrugging:

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -4031,6 +4031,7 @@ int eval_nums(int &n, bool &is_nan, bool &is_nan_forever, T &arg)
 	}
 }
 
+#ifndef _MSC_VER
 /**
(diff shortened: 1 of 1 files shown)


Goober5000:
Another fix is coming shortly.

PR #2084, opened 2019-08-20
Engine Wash Optimizations  *testing of effects not complete, in draft status*

FSCyborg:
A few optimizations to the Engine Wash function I found while bug hunting.

The engine was calculating the same value for tan and cos of ewp->angle.  These values never change, so setting them on table parse is trivial and far more optimal.  I think this function is supposed to be called 4 times a second for every ship, so I believe the optimization is worthwhile.

Secondly, on slower computers, engine wash was likely occasionally firing extra times because a check used less than instead of less than or equal to, forcing a weird situation in the timestamp function, making the function do the whole calculation again the next frame.

Third, pre-multiply a value that gets used more than once later on, so we don't square it every time we use it.

Last, we were calling the damage function for no reason when we weren't going to apply damage, anyway.

Very open to suggestions/corrections on this, since I am really just trying to learn with this PR, and I know that the performance gains are not going to be noticeable.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -652,6 +652,8 @@ static void engine_wash_info_init(engine_wash_info *ewi)
 {
 	ewi->name[0] = '\0';
 	ewi->angle = PI / 10.0f;
+	ewi->cos_result  = 1.0f;
+	ewi->tan_result  = 1.0f;
 	ewi->radius_mult = 1.0f;
 	ewi->length = 500.0f;
 	ewi->intensity = 1.0f;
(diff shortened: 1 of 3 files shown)


MageKing17:
Have you profiled it to see if there's a measurable performance increase (or, at the very least, no measurable performance loss)?

FSCyborg:
I just tested to make sure that it didn't change engine wash behavior.  It behaves as it should.

I'll do some benchmarking to make sure it doesn't harm performance.

FSCyborg:
Can probably test this weekend.

FSCyborg:
One month later: I will probably test today.

FSCyborg:
So, testing did not go well for this. I ran both builds and the "optimizations" did end up making it run a tad bit slower. ¯\_(ツ)_/¯

I don't understand why when leaving it the way it is guarantees another function call to a pretty involved function.....  Maybe it is the way AVX2 works, since I'm allowing that when I compile.  Or it could be that the compiler already sees and does the optimizations but has to do things a longer way when I explicitly tell it to take the shorter path.

Lesson learned I guess.

PR #2087, opened 2019-08-24
Write Minidumps when a crash occurs

asarium:
This will write Mini dumps in the event that the executable causes a
crash. This will allow us to take a closer look at any crashes without
having to reproduce them. This will also upload the PDBs of the builds
to our servers so we can use those for resolving symbols in the crash
dumps.

### DIFF
--- a/cmake/toolchain-msvc.cmake
+++ b/cmake/toolchain-msvc.cmake
@@ -51,7 +51,8 @@ set(CMAKE_STATIC_LINKER_FLAGS "")
 # Release
 set(CMAKE_C_FLAGS_RELEASE "/GL /W2 /Gy- /Ox /Ot /Ob2 /fp:precise /GF /Oy /Oi /Zi /W3")
 set(CMAKE_CXX_FLAGS_RELEASE "/GL /W2 /Gy- /Ox /Ot /Ob2 /fp:precise /GF /Oy /Oi /Zi /W3")
-set(CMAKE_EXE_LINKER_FLAGS_RELEASE "/OPT:REF /LTCG /INCREMENTAL:NO")
+# /DEBUG:FULL so that we get PDBs which is needed for debugging crashdumps
+set(CMAKE_EXE_LINKER_FLAGS_RELEASE "/OPT:REF /LTCG /INCREMENTAL:NO /DEBUG:FULL")
 set(CMAKE_STATIC_LINKER_FLAGS_RELEASE "/LTCG")
 
 globally_enable_extra_compiler_warnings()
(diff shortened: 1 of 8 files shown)


cmake/toolchain-msvc.cmake
```diff
@@ -51,7 +51,8 @@ set(CMAKE_STATIC_LINKER_FLAGS "")
 # Release
 set(CMAKE_C_FLAGS_RELEASE "/GL /W2 /Gy- /Ox /Ot /Ob2 /fp:precise /GF /Oy /Oi /Zi /W3")
 set(CMAKE_CXX_FLAGS_RELEASE "/GL /W2 /Gy- /Ox /Ot /Ob2 /fp:precise /GF /Oy /Oi /Zi /W3")
-set(CMAKE_EXE_LINKER_FLAGS_RELEASE "/OPT:REF /LTCG /INCREMENTAL:NO")
+# /DEBUG:FULL so that we get PDBs which is needed for debugging crashdumps
+set(CMAKE_EXE_LINKER_FLAGS_RELEASE "/OPT:REF /LTCG /INCREMENTAL:NO /DEBUG:FULL")
```
MageKing17: Does it really make sense to have this without enabling `MSVC_RELEASE_DEBUGGING`?

asarium: It is needed in order to get PDBs during normal compilation. As far as I can tell, it doesn't change the code generation options and does not increase the file size of the resulting exe.

MageKing17: But if you don't have `MSVC_RELEASE_DEBUGGING` enabled in the first place, what Release build will ever generate a PDB? Or is the `CMAKE_EXE_LINKER_FLAGS_RELEASE`, in spite of its name, not actually limited to Release builds?

asarium: I'm not sure what you mean here. `CMAKE_EXE_LINKER_FLAGS_RELEASE` sets the flags used for linking an executable. In this PR I added the `/DEBUG:FULL` option to that which is the option responsible for generating PDBs. `MSVC_RELEASE_DEBUGGING` previously enabled that but now it only sets a special optimization option.

MageKing17: `CMAKE_EXE_LINKER_FLAGS_RELEASE` is only used when linking Release builds (and, via `CMAKE_EXE_LINKER_FLAGS_FASTDEBUG` which is copied from it, FastDebug builds), which will not generate a PDB unless `MSVC_RELEASE_DEBUGGING` is enabled. Hence my original question: what is the point of the `/DEBUG:FULL` flag if no PDB will be generated?

MageKing17: Wait, nevermind; I just remembered that the `/Zo` flag is only needed to get sensible variables in the debugger, not for a stacktrace.

PR #2096, opened 2019-09-16
add Smart_secondary_evade_chance

tcrayford:
This adds a new AI profile flag that adds a new chance to evade
secondaries, that is *not* entirely based on countermeasure availability.

Basing secondary evades off of CM count (done without this flag) means
the AI is really bad at evading secondaries in a bunch of situations.
Specifically, the AI never tries to evade when more than 8 countermeasures
are left.

One very notable example is that they're terrible at evading trebuchets
- without this flag, most trebuchet engagements (assuming the AI has
more than 8 countermeasures, so doesn't evade), are totally successful.

I've run through many repeated (I repeat each test 10 times) tests with
this flag, using an Ares and BP2-Captain as the AI profiles.

Without the flag set, or with it set to 0, the AI cannot dodge
trebuchets (0/10 times).

With the flag set to below 0.4 or so, the AI cannot dodge trebuchets
very well at all either (0/10 times).

Once the flag gets up to 0.5, the chance of dodging trebs increases
(5/10 times).

Once the flag gets up to 0.7, the chance of dodging trebs increases so
much that the AI nearly always dodges them (9/10 times are a dodge).

With a more agile ship, the evasion chance increases.


code/ai/aicode.cpp
```diff
@@ -12290,9 +12294,12 @@ void ai_maybe_evade_locked_missile(object *objp, ai_info *aip)
 				case AIM_CHASE:
 					//	Don't always go into evade weapon mode.  Usually, a countermeasure gets launched.
 					// If low on countermeasures, more likely to try to evade.  If 8+, never evade due to low cmeasures.
+                    // tcrayford: add smart secondary evade - when enabled, the AI will use the evasion value as a chance
+					// to evade secondaries instead of never doing it with high countermeasures.
 					if (((((Missiontime >> 18) ^ OBJ_INDEX(objp)) & 3) == 0) || 
 						(objp->phys_info.speed < 40.0f) ||
-						(frand() < 1.0f - (float) shipp->cmeasure_count/8.0f)) {
+						(frand() > (1.0f - aip->ai_smart_secondary_evade_chance)) ||
+						(frand() < (1.0f - (float) (shipp->cmeasure_count/8.0f)))) {
```
MageKing17: Is it intended that a ship with both "$Smart Secondary Evade Chance:" *and* fewer than 8 countermeasures gets two chances to decide to evade?

tcrayford: ah, good catch. I'll fix it.

tcrayford:
@MageKing17 I thought pretty hard about how to solve that situation. What I ended up going with was making a `secondary_evade_chance` variable that's the max of this new flag and the countermeasure flag. Then there's only one check against it, so only one chance.

Goober5000:
As a matter of realism, the AI should try *harder* to evade if it has a low number of countermeasures.  This mirrors what a human would do in that situation.  If CMs are plentiful, the pilot can assume the CMs have a good chance of intercepting the missile.  If there are no CMs left, the pilot is in big trouble unless he can evade.

tcrayford:
@Goober5000 I think this addresses that? The AI will continue to try and evade harder with a low number of CMS, but when it's high this chance comes into it as well.

Goober5000:
Unless I misunderstood, your previous comment says the opposite:

> This would mean that once an AI with smart secondary evade chance goes down to 8 countermeasures, its chance to evade could drop a lot until it's launched more CMs.

tcrayford:
@Goober5000 

> This would mean that once an AI with smart secondary evade chance goes down to 8 countermeasures, its chance to evade could drop a lot until it's launched more CMs.

was in regards to an alternative implementation I tried, but disregarded (because of the issue you point out).

tcrayford:
I need to rerun my tests on the new code folks, mind not merging this until I say it's ready? Should be able to do them all tomorrow.

tcrayford:
Ok, I'm comfortable with these changes now. I ran through the same set of test scenarios, but also did runs with `set-num-countermeasures` set to various numbers between 10 and 0, and the code behaves as expected (both with the new flag and without it).

tcrayford:
@Goober5000 @MageKing17 after careful looking at the AI behaviour in testing, I found another bug with long range weapons.

If a weapon is fired from over 1500m away, (https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/ai/aicode.cpp#L1115, only called when missiles are fired), then the AI will never update `danger_weapon_objnum`. What this leads to is it switching (once per frame!) between `SM_WEAPON_EVADE` and the previous submode, in here: https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/ai/aicode.cpp#L4762

If the value of "smart secondary evade chance" is set low (or under the old code, the countermeasures are high enough), then this leads to the AI "wobbling" when a long range weapon gets within range, and always failing to dodge.

So I've added a small commit to this PR to make it so the AI sets `danger_weapon_objnum` when evading a missile, if "smart secondary evade chance" is enabled at all.

I thought alternatively about doing that via a different flag, but I realized this was part of the "make the AI better at evading secondaries" change and adding lots more flags is annoying.

Goober5000:
Interesting, and great catch!  However I disagree that this should be part of the same flag.  Put it in a new PR, as part of the series of "fix bug" ai profiles flags.

tcrayford:
@Goober5000 ok, that's fine by me :) Is that your usual stance on this kind of thing?

Goober5000:
It really depends on the situation.  In this case, there is an important bug affecting balance that modders will want to fix or at least take into account.  Creating a separate flag makes it more explicit than merging it with another feature.

tcrayford:
@Goober5000 @MageKing17 I've removed the change around setting `danger_weapon_objnum` from this PR now. I'd like to do the new flag in another one when I have more time and availability.

code/ai/aicode.cpp
```diff
@@ -14430,6 +14438,8 @@ void init_aip_from_class_and_profile(ai_info *aip, ai_class *aicp, ai_profile_t
 		profile->chance_to_use_missiles_on_plr[Game_skill_level] : aicp->ai_chance_to_use_missiles_on_plr[Game_skill_level];
 	aip->ai_max_aim_update_delay = (aicp->ai_max_aim_update_delay[Game_skill_level] == FLT_MIN) ? 
 		profile->max_aim_update_delay[Game_skill_level] : aicp->ai_max_aim_update_delay[Game_skill_level];
+	aip->ai_smart_secondary_evade_chance = (aicp->ai_smart_secondary_evade_chance[Game_skill_level] == FLT_MIN) ? 
+		profile->max_aim_update_delay[Game_skill_level] : aicp->ai_smart_secondary_evade_chance[Game_skill_level];
```
Goober5000: Looks like a copy+paste typo here.  Shouldn't the field after the question mark be a secondary_evade_chance?

Goober5000:
Ok, but at least start the `danger_weapon_objnum` PR so that we have an active record of it.

wookieejedi:
This can probably be closed since #3177 was merged. I'll leave it up to a senior member.

Goober5000:
I'll defer to @Baezon but I think so.

Baezon:
Given the significant overlap (subset, really) with #3177, this would have to undergo some significant changes to justify as a separate feature (which I definitely wouldn't be opposed to). 

Plus how long its been since there's been any activity, I would argue for its closure, but I dunno if closing for inactivity is something we normally do. I can always be reopened later if tcrayford wants.

Goober5000:
It depends on the ticket.  But we have closed other tickets for significant overlap so it's appropriate to close this one too.

PR #2098, opened 2019-09-23
add seconds-to-goal sexp

tcrayford:
This adds `seconds-to-goal`, as requested by `0rph3u5` on the HLP discord:

request:

> wait a minute, there is no function to get the estimate for a ship reaching its waypoints?
(I mean I get that for simple waypoints I can just do a distance/speed and update that every time I change the speed or each time a waypoint is passed; or just go in the mission and copy what's on the HUD - but really the engine already does that for the HUD and there is no sexp to get the value from that display?)

I've slightly refactored the `ship_return_time_to_goal` so that there's a function to grab the seconds now.

I've tested this:

1. I setup a simple mission with a ship and a waypoint and an initial order to go there, and the time in the hud shows up the same as a normal build
2. I setup the same mission with this sexp:

```
$Formula: ( when 
   ( true ) 
   ( training-msg "debug" ) 
   ( modify-variable @debug[0] (seconds-to-waypoint "Support 1") ) 
)
+Name: Event name
+Repeat Count: -1
+Trigger Count: 276447231
+Interval: 0
```

This message:

```
$Name: debug
$Team: -1
$MessageNew:  XSTR("$debug", -1)
$end_multi_text
```

and the variable `debug` - and the training message shows the timer decreasing in seconds as the ship gets closer.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -758,6 +758,7 @@ SCP_vector<sexp_oper> Operators = {
 	{ "special-check",					OP_SPECIAL_CHECK,						1,	1,			SEXP_ACTION_OPERATOR,	},
 	{ "set-training-context-fly-path",	OP_SET_TRAINING_CONTEXT_FLY_PATH,		2,	2,			SEXP_ACTION_OPERATOR,	},
 	{ "set-training-context-speed",		OP_SET_TRAINING_CONTEXT_SPEED,			2,	2,			SEXP_ACTION_OPERATOR,	},
+	{ "seconds-to-goal",				OP_SECONDS_TO_GOAL,	   	         		1,	1,			SEXP_INTEGER_OPERATOR,	}, // tcrayford
 };
 
 sexp_ai_goal_link Sexp_ai_goal_links[] = {
@@ -16457,6 +16458,12 @@ int sexp_query_orders (int n)
 	return hud_query_order_issued(order_to, order, target, timestamp, order_from, special);
 }
 
+int sexp_seconds_to_waypoint (int n)
+{
+    ship* shipp = sexp_get_ship_from_node(n);
+	return ship_return_seconds_to_goal(shipp);
+}
+
 // Karajorma
 void sexp_reset_orders (int  /*n*/)
 {
@@ -25359,6 +25366,11 @@ int eval_sexp(int cur_node, int referenced_node)
 				sexp_val = sexp_query_orders(node);
 				break;
 
+			case OP_SECONDS_TO_GOAL:
+				sexp_val = sexp_seconds_to_waypoint(node);
+				break;
+
+
 			// Karajorma
 			case OP_RESET_ORDERS:
 				sexp_reset_orders(node);
@@ -26898,6 +26910,7 @@ int query_operator_return_type(int op)
 		case OP_GET_COLGROUP_ID:
 		case OP_FUNCTIONAL_IF_THEN_ELSE:
 	    case OP_GET_HOTKEY:
+        case OP_SECONDS_TO_GOAL:
 			return OPR_NUMBER;
 
 		case OP_ABS:
@@ -27701,6 +27714,10 @@ int query_operator_argument_type(int op, int argnum)
 				return OPF_SUBSYSTEM;
 			else
 				return OPF_SHIP_WING;
+
+		case OP_SECONDS_TO_GOAL:
+			if (argnum == 0)
+				return OPF_SHIP;
 
 		case OP_WAS_DESTROYED_BY_DELAY:
 			if (argnum == 0)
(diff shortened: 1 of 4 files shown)


tcrayford:
Right now I have a fucked up fred/fs2 situation that means I can't easily run fred on a local build (I dev on a mac, and only use fred via a VM), so I haven't checked the format of the help text. I'll try and remediate that.

code/parse/sexp.cpp
```diff
@@ -758,6 +758,7 @@ SCP_vector<sexp_oper> Operators = {
 	{ "special-check",					OP_SPECIAL_CHECK,						1,	1,			SEXP_ACTION_OPERATOR,	},
 	{ "set-training-context-fly-path",	OP_SET_TRAINING_CONTEXT_FLY_PATH,		2,	2,			SEXP_ACTION_OPERATOR,	},
 	{ "set-training-context-speed",		OP_SET_TRAINING_CONTEXT_SPEED,			2,	2,			SEXP_ACTION_OPERATOR,	},
+	{ "seconds-to-goal",				OP_SECONDS_TO_GOAL,	   	         		1,	1,			SEXP_INTEGER_OPERATOR,	}, // tcrayford
```
Goober5000: Change the sexp name to `time-to-goal` and put it in the TIME category.

code/parse/sexp.cpp
```diff
@@ -16457,6 +16458,12 @@ int sexp_query_orders (int n)
 	return hud_query_order_issued(order_to, order, target, timestamp, order_from, special);
 }
 
+int sexp_seconds_to_waypoint (int n)
```
Goober5000: Rename the function to `sexp_time_to_goal`.

code/parse/sexp.cpp
```diff
@@ -33016,7 +33033,10 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"\tBecomes true when a waypoint is hit that is before the last one hit, which "
 		"indicates they have flown a waypoint twice.\r\n\r\n"
 		"Returns a boolean value.  Takes no arguments." },
-
+	{ OP_SECONDS_TO_GOAL, "Seconds-to-goal (Number training operator)\r\n"
```
Goober5000: Sexp help should be rewritten so that it's time instead of training and goal instead of waypoint.  Otherwise it looks good.

code/parse/sexp.h
```diff
@@ -820,6 +820,7 @@ class waypoint_list;
 #define OP_QUERY_ORDERS						(0x0014 | OP_CATEGORY_TRAINING) // Karajorma
 #define OP_NODE_TARGETED					(0x0015 | OP_CATEGORY_TRAINING) // FUBAR
 #define OP_IGNORE_KEY						(0x0016 | OP_CATEGORY_TRAINING) // Karajorma
+#define OP_SECONDS_TO_GOAL					(0x0017 | OP_CATEGORY_TRAINING) // tcrayford
```
Goober5000: `OP_CATEGORY_TIME`, not training.

code/ship/ship.cpp
```diff
@@ -15079,9 +15079,32 @@ SCP_string ship_return_orders(ship* sp)
...
+	if ( time >= 0 ) {
+		minutes = time/60;
+		seconds = time%60;
+		if ( minutes > 99 ) {
+			minutes = 99;
+			seconds = 99;
+		}
+		sprintf(outbuf, NOX("%02d:%02d"), minutes, seconds);
+	} else if ( time == -2) {
+        // -2 is used as a signal of "there is no valid goal right now"
```
Goober5000: This is all well and good, but shouldn't something be done here?  What sort of contract do the callers expect?

Hint: I looked at where the function is used, and there are two use cases for which you return -2.  They are supposed to be handled differently.

LightblueMobius:
At this point I guess you could turn this into a **msecs-to-goal** SEXP in order to ensure high degrees of precision, or add an extra argument (true/false, default would be false) turning seconds into milliseconds.

Goober5000:
At this point, if @tcrayford doesn't come back, it would be nice for another coder to pick up where he left off.

code/parse/sexp.cpp
z64555: "Sexp help should be rewritten so that it's time instead of training"

Er, so like "Time-to-goal (Number time operator)" or what?

code/parse/sexp.h
z64555: Should it also be `OP_NONCAMPAIGN_FLAG`?

code/ship/ship.cpp
z64555: What do you recommend here?
The first -2 return case is when the computed distance is <1.0, presumably in meters.  So this may be assumed to be a short-circuit and should probably return a time estimate of 0s.

The second -2 return case is when the goal is not a waypoint or dock point, which could either return 0s or Unknown.

code/parse/sexp.cpp
z64555: Update, I found the group of time operators and have made the  sexp help to "Time-to-goal (Time operator)" or similar.  See the commit in the new PR for details

z64555:
I've revived this feature in PR #3064. All further discussion and review should please defer to it.

PR #2105, opened 2019-10-12
miscellaneous XSTR cleanup

### LINKED ISSUE #1285: How to handle translated strings in SEXPs?
[Mantis 3192](http://scp.indiegames.us/mantis/view.php?id=3192) reports an issue with how `grant-medal` doesn't work with translated medal names. As far as I can tell, this is caused by the inability of the SEXP system to translate the English medal name in the SEXP to the correct, translated version of the string.

How could this be fixed? I don't know enough about the SEXP system to fix this so maybe someone else has an idea.

Goober5000:
~~The grant-medal and was-medal-granted sexps will now translate medal names, in the same way that tech-add-intel-xstr translates intel entries.  There is also some minor cleanup and the specifying of correct default -1 numbers for the xstr-capable sexps.~~

~~This fixes #1285.~~

With Novachen's comment, this PR drops the modifications to `grant-medal` and `was-medal-granted`, and just does some cleanup with XSTR and CTEXT.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -15328,7 +15328,7 @@ void sexp_ship_guardian_threshold(int node)
 		// check to see if ship destroyed or departed.  In either case, do nothing.
 		ship_name = CTEXT(n);
 
-		if ( mission_log_get_time(LOG_SHIP_DEPARTED, ship_name, NULL, NULL) || mission_log_get_time(LOG_SHIP_DESTROYED, ship_name, NULL, NULL) || mission_log_get_time(LOG_SELF_DESTRUCTED, ship_name, NULL, NULL) ) {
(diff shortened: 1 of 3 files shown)


NovachenFS2:
Mh, is this actually needed? I mean, Medal names can be translated via $Alt Name: in the medals.tbl.

Goober5000:
In that case, #1285 should have been closed.  I had the impression that #1736 did not fix the medal issue.  I will close that now.

However I will modify this PR because there are a few miscellaneous fixes that should be added to the code.

code/parse/sexp.cpp
```diff
@@ -30111,8 +30082,9 @@ char *CTEXT(int n)
 	char *current_argument; 
 
 	Assertion(n >= 0 && n < Num_sexp_nodes, "Passed an out-of-range node index (%d) to CTEXT!", n);
-	if ( n < 0 ) {
-		return NULL;
+	if ( n < 0 || n >= Num_sexp_nodes ) {
+		// we should make CTEXT return const char*, but that's for another PR
+		return const_cast<char*>("!INVALID!");
```
asarium: The returned string should be more specific (e.g. `INVALID CTEXT NODE INDEX SPECIFIED`) so that we immediately see where this comes from if we ever have issues with it.

Goober5000: Ok. Done.

Issue #1138, opened 2017-01-06
FreeBSD warning in stubs.cpp


chief1983:
```
[ 90%] Building CXX object code/CMakeFiles/code.dir/windows_stub/stubs.cpp.o
/usr/home/chief1983/fs2_open_git/code/windows_stub/stubs.cpp:80:20: warning: comparison of integers of different signs: 'int' and 'unsigned long' [-Wsign-compare]
        for (int i = 1; i < numstrings; i++)
                        ~ ^ ~~~~~~~~~~
1 warning generated.
```

Not sure how it got past other compilers, unless they don't end up with 'unsigned long' from the auto type numstrings is defined with.

asarium:
On Linux the signature is `int backtrace (void **, int)` so there is no signed/unsigned conflict on the CI systems. This could be fixed in a cross-platform way by always using `size_t` but that may cause issues on Linux where the backtrace functions use `int` instead of `size_t`.

chief1983:
Is Mac behaving more like Linux than FreeBSD in this case?  Because since it is using clang and is a BSD derivative I would have expected it to be more like FreeBSD.  But I got no such warning on my latest Mac builds.

asarium:
According to the [Apple Documentation](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/backtrace.3.html) it uses the `int` version of the functions.

Goober5000:
A little farther down in the same file, we have this:
```
	for (auto i = 0; i < numstrings; ++i) {
```

Can the offending line be fixed by using `auto`?

As an aside, I don't know if this is a CMake thing, but `stubs.cpp` is not part of the Visual Studio project, even though `config.h` in the same folder is.

asarium:
`auto` wouldn't work since `i` would still be an `int` in that case. However, something like `decltype(numstrings)` should work since then it would automatically use the right type.

Goober5000:
Sounds good.  I opened a new PR.

And, incidentally, I figured out that `stubs.cpp` is not included because it's not a Windows file.

Goober5000:
Fixed in #2107

PR #2107, opened 2019-10-12
fix FreeBSD warning in stubs.cpp

Goober5000:
Use m!m's suggestion to fix issue #1138

### DIFF
--- a/code/windows_stub/stubs.cpp
+++ b/code/windows_stub/stubs.cpp
@@ -77,6 +77,8 @@ SCP_string dump_stacktrace()
 	
 	// iterate over the returned symbol lines. skip the first, it is the
 	// address of this function.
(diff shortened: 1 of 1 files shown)


code/windows_stub/stubs.cpp
```diff
@@ -77,7 +77,7 @@ SCP_string dump_stacktrace()
 	
 	// iterate over the returned symbol lines. skip the first, it is the
 	// address of this function.
-	for (int i = 1; i < numstrings; i++)
+	for (decltype(numstrings) i = 1; i < numstrings; i++)
```
asarium: Maybe a comment could be added to explain why we need that `decltype` stuff? Otherwise someone could remove this again in the future.

Goober5000: Makes sense.

PR #2109, opened 2019-10-15
Mantis 1826 - Sending Packet Updates for Campaigns

JohnAFernandez:
This allows FSO to send campaign descriptions during Multi, and the description of the first mission, like it does when it is in mission mode.  Also allows us to send the whole description via some quick recursion.

Lots of feedback welcome.  This is probably my most extensive contribution so far.  Also needs testing, but I thought I would also open it up for you guys to add initial reactions/corrections if you have them.

### DIFF
--- a/code/network/multi.h
+++ b/code/network/multi.h
@@ -89,6 +89,8 @@ class player;
 #define MAX_GAMENAME_LEN					32				// maximum length in characters of a game name
 #define DESCRIPT_LENGTH						512			// maximum length of a mission description (as specified by Fred)
 #define MAX_PASSWD_LEN						16				// maximum length of the password for a netgame
+#define MAX_PACKET_DESC_LEN		MAX_PACKET_SIZE - 12	// the maximum description length that can be sent in a packet
+#define MAX_TOTAL_DESC_LEN		2*(MAX_PACKET_DESC_LEN) // the maximum that the desciption packet will be able to send period. 
 
 // low level networking defines
 #define IP_ADDRESS_LENGTH					4				// length of the address field for an IP address
@@ -366,21 +368,6 @@ class player;
 
 #define MAX_SHIPS_PER_PACKET		64			// Number of ships in a STATS_MISSION_KILLS or STATS_ALLTIME_KILLS packet
 
-// campaign and mission description packet post-header type info
-#define DESCRIPT_MESSAGE_REQUEST	(1 << 0)
-#define DESCRIPT_PACKET_CAMPAIGN_MESSAGE_1	(1 << 1)
-#define DESCRIPT_PACKET_CAMPAIGN_MESSAGE_2	(1 << 2)
-#define DESCRIPT_PACKET_MISSION_MESSAGE_1	(1 << 3)
-#define DESCRIPT_PACKET_MISSION_MESSAGE_2	(1 << 4)
-#define DESCRIPT_PACKET_NO_ADDENDUM			(1 << 5)
-#define DESCRIPT_PACKET_FINAL_MESSAGE		(1 << 6)
-
-#define MULTI_MAX_DESC_LEN			2*(MAX_PACKET_SIZE - 10)
-
-const char PRE_CAMPAIGN_DESC[23] =	"Campaign Description:\n";
-const char PRE_MISSION_DESC[28]  =	"First Mission:\n";
-const char DOUBLE_NEW_LINE[3]    =  "\n\n";
-
 // ----------------------------------------------------------------------------------------
 
 
@@ -497,8 +484,8 @@ typedef struct netgame_info {
 	char		title[NAME_LENGTH+1];			// title of the mission (as appears in the mission file)
 	char		campaign_name[NAME_LENGTH+1];	// current campaign name	
 	char		passwd[MAX_PASSWD_LEN+1];		// password for the game
-	char		mission_desc[MULTI_MAX_DESC_LEN];      // Cyborg17 - Description of the current mission.
-	char		campaign_desc[MULTI_MAX_DESC_LEN];     // Cyborg17 - Description of the current campaign.
+	char		mission_desc[MAX_TOTAL_DESC_LEN];      // Cyborg17 - Description of the current mission.
+	char		campaign_desc[MAX_TOTAL_DESC_LEN];     // Cyborg17 - Description of the current campaign.
 	int		version_info;						// version info for this game.
 	int		type_flags;							// see NG_TYPE_* defines
 	int		mode;									// see NG_MODE_* defines
--- a/code/network/multi.h
+++ b/code/network/multi.h
@@ -257,7 +257,7 @@ class player;
 #define NETPLAYER_UPDATE			0xE4		// a player update packet
 #define OBJECT_UPDATE				0xE6		// an object update packet from server to all clients
 #define MISSION_LOG_ENTRY			0xE7		// ad an item into the mission log
-#define UPDATE_DESCRIPT				0xE8		// update the netgame description
+#define UPDATE_DESCRIPT				0xE8		// update the netgame description -- Cyborg17, Now updated for multiple description packets
 #define COUNTDOWN						0xE9		// countdown timer for starting a game (pretty benign)
 #define DEBRIEF_INFO					0xEA		// end of mission debriefing information
 #define EMP_EFFECT					0xEB		// EMP effect (mission disk only)
@@ -365,6 +365,21 @@ class player;
 #define STATS_ALLTIME_KILLS			5			// alltime kills, for one player
 
 #define MAX_SHIPS_PER_PACKET		64			// Number of ships in a STATS_MISSION_KILLS or STATS_ALLTIME_KILLS packet
+
+// campaign and mission description packet post-header type info
+#define DESCRIPT_MESSAGE_REQUEST	(1 << 0)
+#define DESCRIPT_PACKET_CAMPAIGN_MESSAGE_1	(1 << 1)
+#define DESCRIPT_PACKET_CAMPAIGN_MESSAGE_2	(1 << 2)
+#define DESCRIPT_PACKET_MISSION_MESSAGE_1	(1 << 3)
+#define DESCRIPT_PACKET_MISSION_MESSAGE_2	(1 << 4)
+#define DESCRIPT_PACKET_NO_ADDENDUM			(1 << 5)
+#define DESCRIPT_PACKET_FINAL_MESSAGE		(1 << 6)
+
+#define MULTI_MAX_DESC_LEN			2*(MAX_PACKET_SIZE - 10)
+
+const char PRE_CAMPAIGN_DESC[23] =	"Campaign Description:\n";
+const char PRE_MISSION_DESC[28]  =	"First Mission:\n";
+const char DOUBLE_NEW_LINE[3]    =  "\n\n";
 
 // ----------------------------------------------------------------------------------------
 
@@ -482,6 +497,8 @@ typedef struct netgame_info {
 	char		title[NAME_LENGTH+1];			// title of the mission (as appears in the mission file)
 	char		campaign_name[NAME_LENGTH+1];	// current campaign name	
 	char		passwd[MAX_PASSWD_LEN+1];		// password for the game
+	char		mission_desc[MULTI_MAX_DESC_LEN];      // Cyborg17 - Description of the current mission.
+	char		campaign_desc[MULTI_MAX_DESC_LEN];     // Cyborg17 - Description of the current campaign.
 	int		version_info;						// version info for this game.
 	int		type_flags;							// see NG_TYPE_* defines
 	int		mode;									// see NG_MODE_* defines
--- a/code/network/multimsgs.cpp
+++ b/code/network/multimsgs.cpp
@@ -73,6 +73,11 @@
 #include "network/multi_sexp.h"
 #include "parse/parselo.h"
 
+// Labels for Mission description
+const char* PRE_CAMPAIGN_DESC =	"Campaign Description:\n";
+const char* PRE_MISSION_DESC  =	"First Mission:\n";
+const char* DOUBLE_NEW_LINE   =  "\n\n";
+
 // #define _MULTI_SUPER_WACKY_COMPRESSION
 
 #ifdef _MULTI_SUPER_WACKY_COMPRESSION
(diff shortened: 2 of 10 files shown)


Goober5000:
Here's a link to Mantis: http://scp.indiegames.us/mantis/view.php?id=1826

FSCyborg:
It's building on MSVC, but I'm still trying to clear out a few warnings.  Probably will ask for help testing tomorrow night.

FSCyborg:
The last push I did should handle the last MSVC warning.  Now waiting on any appveyor and travis warnings to see if there is any thing else I need to fix.  Still have not yet tested.

FSCyborg:
Just continuing to clean up the code.  Current testing makes me thing I should have put a pre-iterator instead of a post-iterator.

notimaginative:
There looks to be a memory leak in here. In mission_campaign_get_info() "first_mission" is malloc'd but it's never free'd after it's used.

FSCyborg:
That's the type of stuff that we amateurs really appreciate you guys looking out for. 👍

notimaginative:
No problem :+1: 

One other bug I noticed as well... When sending just a request, `packet_type` is set but not added to the packet. This will break on the other end since it's going to try and read that byte.

Overall though, this is ultimately a breaking change. I mean it's close, _real_ close, but it breaks compatibility with older builds.

First off, `packet_type` is a flag. As-is it should work fine but the potential to break things later is there and might not be noticed. With the existing processing code it checks if this byte is `1`, specifically. So if `packet_type` ever has two or more values set it's going to break older clients.

Second is that if that byte is not `1` then it sets or appends the message. With older clients sending to newer clients -- where the byte is `0` -- this will append the message. The text buffer is never set though and you should never strcat() when the source doesn't contain a string. It might test okay with a debug build since the text buffer is probably zero'd but you can't depend on that behavior. If the default behavior in that function was to set the text, not append it, then you could avoid that issue altogether. But when newer clients send to older clients the text will always be set to the last packet to arrive thereby causing the description to be wrong if it requires more than one packet.

Third is that UDP makes no guarantee about packet order. So the code is sending multiple packets intended to be in a particular order but the client might not receive them in that order. So in this respect the code is, by design, a hard to reproduce bug. :sweat_smile: 

My suggestion would probably be to leave `process_netgame_descript_packet()` unchanged from what's already there. Modify `send_netgame_descript_packet()` to work like before (using `code` as the first byte, etc.) when sending the first packet, then have the additional descriptions be sent as a brand new packet type (e.g. `UPDATE_DESCRIPT_EXT` with `send_netgame_descript_ext_packet()` / `process_netgame_descript_ext_packet()`). The overall logic can stay the same, just process the additional packets as a new type. That should maintain full compatibility between older and newer builds while newer builds still get the extra functionality.

As for the packet ordering issue, ummm... you might have to get creative :grin:

FSCyborg:
Welp. This is a learning experience, after all, so I really should expect to fix this bug for a fourth time in a row.  Aren't we going to run out of packet types soon?

notimaginative:
Like I said, you're really close with it. I didn't notice the problems right away either. It's actually pretty easy to break the old multi code in ways that aren't readily obvious. I've been burned by it a few times.

I'm not sure how many possible packet types are left. Probably around 20-25. I think that some of them might be deprecated though, left over from FreeSpace 1. And I didn't use any for PXO since it has it's own packet setup.

FSCyborg:
I was trying to work with the existing functions and functionality as much as possible to avoid having to use a new packet type.  I am glad you caught it before I was up for three days straight not understanding why there was a bug.  (in fact, in testing, my brother and I probably already encountered it.

Maybe there's a way to..... tell if a client is using old netcode as apposed to new netcode during the join request and treat them differently from then on?  (Although I would guess that that would probably break the join request packet code)

One better idea I had while reading your message is a packet type that designates the order or existence of other packets so that FSO can perhaps start dealing with dropped/out of order packets.  And we can start storing them in an SCP_list or struct or class?

At the very least we could possibly see how regular packet dropping and out of order packets are?

notimaginative:
You could always just bump the multi version, but that's the nuclear option. I looked at a way to just make it work as-is but that sort of thinking can get really messy. It's best to just avoid workarounds. Adding a new packet type (just need the one) is the cleanest and least hack-y way to get this feature working in a fully compatible way.

As far as the dropped/out-of-order packet thing goes: stay as far away from that rabbit hole as you can. Seriously, that's just not a battle you are going to win. It's simply the nature of UDP and it's working as designed. Outside of a rewrite to TCP, which would break absolutely everything, there just isn't anything you can do to fix that. You can probably find stuff via google about object update packets with regards to dropping and order but that too would be a breaking change. It's also a very difficult thing to test since it's entirely network dependent. Although, I think that the multilag debug code will drop packets if you want to experiment with that.

FSCyborg:
Haven't forgotten about this, but I'll have to re-write it when I get a chance.  Probably after 19.0 is out since it is such a minor thing.

notimaginative:
I've been ripping up `psnet2` in preparation for the IPv6 port and I've gotten a better look at `psnet_rel_send()`. That function sends "reliable" UDP packets. And by that I mean it will resend a packet until the client/server confirms it has it, and it will try to preserve order as well. It's not 100% but it's about as close as you can get. While not a simple thing, if you modify you code to use this functionality (look at `multi_io_send_reliable()` usage) then it should solve your ordering problems.

notimaginative:
I'm not sure how much work you've done to make it compatible yet, but if you hold this until the IPv6 port then it will be safe to make it a breaking change.

JohnAFernandez:
I probably did way too much for this very minor feature, but it's done.

code/network/multimsgs.cpp
```diff
@@ -2161,79 +2161,218 @@ void process_netgame_update_packet( ubyte *data, header *hinfo )
...
+				second_packet_needed = true;
+				desc_len = MAX_PACKET_DESC_LEN;
+				strcpy_s(message, PRE_CAMPAIGN_DESC);
+				strncat(message, temp, MAX_PACKET_DESC_LEN - 23);
+				packet_type |= DESCRIPT_PACKET_CAMPAIGN_MESSAGE_1;
+
+			} else if (addendum) {
+				// if we now recursing
+				backspace(temp, MAX_PACKET_DESC_LEN - 24); // Subtract because we had the campaign desc intro text.
+				desc_len -= (MAX_PACKET_DESC_LEN - 24);
```
notimaginative, line 2216: What's with all of the magic numbers? 23, 24, 27, 28? If they are needed then reference them with some #define's so that future coders won't need to figure out what they're for.

code/network/multimsgs.cpp
```diff
@@ -2161,79 +2161,218 @@ void process_netgame_update_packet( ubyte *data, header *hinfo )
...
+			else if (!addendum && desc_len > MAX_PACKET_DESC_LEN - 27) {
+
+				// if we are going to need to recurse....
+				second_packet_needed = true;
+				desc_len = MAX_PACKET_DESC_LEN;
+
+				// check if we need to have a mission description marker
+				if (message_count > 0) {
+					strcpy_s(message, PRE_MISSION_DESC);
+					strncat(message, temp, MAX_PACKET_DESC_LEN - 28);
```
notimaginative, line 2262: Use `strcat_s()` here.

code/network/multimsgs.cpp
```diff
@@ -2161,79 +2161,218 @@ void process_netgame_update_packet( ubyte *data, header *hinfo )
...
-	
-	Assert(addr != NULL);
-	if(addr != NULL){
+
+		// add bitwise data and string lengths.
+		ADD_DATA(packet_type);
+
+		ADD_DATA(Multi_sent_descript_packet_current_index);
+		// place it into the packet and send
+		mprintf(("packet_type: %d message: %s\n", packet_type, message));
```
notimaginative, line 2306: Debug message should be removed

code/network/multiui.cpp
```diff
@@ -193,6 +200,48 @@ void multi_common_add_text(const char *str,int auto_scroll)
...
+		strcpy_s(Multi_received_mission_description2, str_in);
+	}
+	else if (code & DESCRIPT_PACKET_MISSION_MESSAGE_1) {
+		memset(Multi_received_mission_description3, 0, MAX_PACKET_SIZE);
+		strcpy_s(Multi_received_mission_description3, str_in);
+	}
+	else if (code & DESCRIPT_PACKET_MISSION_MESSAGE_2) {
+		memset(Multi_received_mission_description4, 0, MAX_PACKET_SIZE);
+		strcpy_s(Multi_received_mission_description4, str_in);
+	}
```
notimaginative, line 237: No need to `memset()` here, the `strcpy_s()` will null terminate the string

code/network/multiui.cpp
```diff
@@ -4776,22 +4843,27 @@ void multi_create_list_select_item(int n)
...
 				ng->max_players = mission_parse_get_multi_mission_info( ng->mission_name );				
-				
+
 				Assert(ng->max_players > 0);
 				strcpy_s(ng->title,The_mission.name);								
 
-				// set the information area text
-				multi_common_set_text(The_mission.mission_desc);
+				// set the information area text, and update the ng struct
+				strcpy(ng->mission_desc, The_mission.mission_desc);
```
notimaginative, line 4861: Use `strcpy_s()` here. Also, the size of `The_mission.mission_desc` is larger than the size of `ng->mission_desc` so it will be set to an empty string on overflow.

code/network/multiui.cpp
```diff
@@ -4825,19 +4900,42 @@ void multi_create_list_select_item(int n)
...
 
 				// set the information area text
-				// multi_common_set_text(ng->title);
-				if (campaign_desc != NULL)
-				{
-					multi_common_set_text(campaign_desc);
+				//	Cyborg17 - Now that we can have both descriptions, markers in the text are helpful.
+				//  But that means we have a lot of concatenation...
+				char full_text[MULTI_COMMON_MAX_TEXT];				
+				memset(full_text, 0, MULTI_COMMON_MAX_TEXT);
```
notimaginative, line 4906: `memset()` isn't needed here

code/network/multiui.cpp
```diff
@@ -4825,19 +4900,42 @@ void multi_create_list_select_item(int n)
...
+
+				// We can let the player know that the descption is empty by keeping it empty.
+				strcpy(full_text, PRE_CAMPAIGN_DESC);
+				if (campaign_desc != nullptr) {
+					strcpy(ng->campaign_desc, campaign_desc);
```
notimaginative, line 4911: Use `strcpy_s()` here

code/network/multiui.cpp
```diff
@@ -4825,19 +4900,42 @@ void multi_create_list_select_item(int n)
...
+					strcat_s(full_text, campaign_desc);
+				}	
+
+				// if we successfully got the mission description.
+				if (!get_mission_info(first_mission, mp, true)) {
+					 if (strlen(mp->mission_desc) > 0) {
+						strcat(full_text, DOUBLE_NEW_LINE);
+						strcat(full_text, PRE_MISSION_DESC);
```
notimaginative, line 4919: Use `strcat_s()` here

code/network/psnet2.h
```diff
@@ -32,6 +32,7 @@
 // kazan - I think this should raise the ships limit across the network
 //#define MAX_PACKET_SIZE 4096
 #define MAX_PACKET_SIZE		512
+#define MAX_PACKET_DESC_LEN MAX_PACKET_SIZE - 12
```
notimaginative, line 35: Just move this to `multimsgs.cpp` since that's the only place it's used

code/network/multi.h
```diff
@@ -366,6 +366,21 @@ class player;
...
 #define MAX_SHIPS_PER_PACKET		64			// Number of ships in a STATS_MISSION_KILLS or STATS_ALLTIME_KILLS packet
 
+// campaign and mission description packet post-header type info
+#define DESCRIPT_MESSAGE_REQUEST	(1 << 0)
+#define DESCRIPT_PACKET_CAMPAIGN_MESSAGE_1	(1 << 1)
+#define DESCRIPT_PACKET_CAMPAIGN_MESSAGE_2	(1 << 2)
+#define DESCRIPT_PACKET_MISSION_MESSAGE_1	(1 << 3)
+#define DESCRIPT_PACKET_MISSION_MESSAGE_2	(1 << 4)
+#define DESCRIPT_PACKET_NO_ADDENDUM			(1 << 5)
+#define DESCRIPT_PACKET_FINAL_MESSAGE		(1 << 6)
```
notimaginative, line 376: This should just be moved to `multimsgs.cpp` since it's not used anywhere else.

code/network/multi.h
```diff
@@ -366,6 +366,16 @@ class player;
 
 #define MAX_SHIPS_PER_PACKET		64			// Number of ships in a STATS_MISSION_KILLS or STATS_ALLTIME_KILLS packet
 
+// campaign and mission description packet post-header type info
+#define DESCRIPT_MESSAGE_REQUEST	(1 << 0)
+#define DESCRIPT_PACKET_MESSAGE_1	(1 << 1)
+
+#define MULTI_MAX_DESC_LEN			2*(MAX_PACKET_SIZE - 10)
```
notimaginative, line 378: Okay, I see the difference now, but it's just confusing. `MULTI_MAX_DESC_LEN` vs `MULTI_PACKET_DESC_LEN`, plus more magic numbers. There should be a less confusing way to handle that.

code/network/multi.h
```diff
@@ -366,6 +366,21 @@ class player;
...
+#define DESCRIPT_PACKET_MISSION_MESSAGE_1	(1 << 3)
+#define DESCRIPT_PACKET_MISSION_MESSAGE_2	(1 << 4)
+#define DESCRIPT_PACKET_NO_ADDENDUM			(1 << 5)
+#define DESCRIPT_PACKET_FINAL_MESSAGE		(1 << 6)
+
+#define MULTI_MAX_DESC_LEN			2*(MAX_PACKET_SIZE - 10)
+
+const char PRE_CAMPAIGN_DESC[23] =	"Campaign Description:\n";
+const char PRE_MISSION_DESC[28]  =	"First Mission:\n";
+const char DOUBLE_NEW_LINE[3]    =  "\n\n";
```
notimaginative, line 382: These should just be pointers and also not be defined in `multi.h`. Maybe set them in `multimsgs.cpp` and extern when you need in `multiui.cpp`

code/network/multi.h
JohnAFernandez: I hope you meant a better name would be sufficient.

The one number I could not figure out from the original code was that they cut off 10 bytes for the packet.

code/network/psnet2.h
JohnAFernandez: This one was actually referenced in a few places, so it seemed to be best to move this into the strings section in multi.h.

code/network/multimsgs.cpp
JohnAFernandez: They were just the length of the constant strings, I changed it to strlen()

JohnAFernandez:
A better version  of this is going to be incorporated into the multi packet fixes PR.

PR #2111, opened 2019-10-17
Move filename sort in cf_get_file_list() to before packfile search.

MageKing17:
The intent behind this sort (added in #2094) was to make the order of files returned by the filesystem consistent across platforms; however, the sort occurred not only after the filesystem search, but after the packfile search, which changed the order files load in on all filesystems instead of making it consistent. This moves the sort to be after the filesystem search but before the packfile search, hopefully making load order consistent once more.


Goober5000:
Yeah that should work better.

Goober5000:
While we're sorting, can I request a new PR that rewrites `cf_sort_filenames` to use `std::stable_sort`?

PR #2113, opened 2019-10-17
Give PNG warnings their own category, so they're hidden by default.

MageKing17:
This probably isn't the kind of information that should appear in the average debug log, since it's unlikely to be of use in standard troubleshooting.


Goober5000:
What sort of other `nprintf` categories are there?  Is there an existing one that might fit?

MageKing17:
I did a quick search through the codebase for an appropriate category, but the closest I found were "Warning" (which is enabled by default) and "OpenGL Debug" (in that it's also used to hide irrelevant debug messages generated by a third-party library). Since categories are free, I didn't see a reason not to just make a new one.

Goober5000:
Ok.

PR #2115, opened 2019-10-19
allow multiple simultaneous sexp-controlled audio streams

Goober5000:
Enhances play-sound-from-file, etc. to allow more than one audio stream to be played at the same time, provided all stream are referenced by audio handles contained in variables.


Goober5000:
Draft PR.  Needs testing.

Goober5000:
Tested; ready for review.

PR #2119, opened 2019-10-21
Clear object events when Lua state is shut down

asarium:
Since the Lua state is destroyed before all the object events, this was
causing issues. This should fix that by clearing the events when the lua
state is destroyed.

### DIFF
--- a/code/object/object.cpp
+++ b/code/object/object.cpp
@@ -322,6 +322,14 @@ float get_shield_pct(object *objp)
 	return shield_get_strength(objp) / total_strength;
 }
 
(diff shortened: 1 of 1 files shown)


Goober5000:
"destroy" is misspelled in "on_script_state_destory".

Can you add those two `clear()` calls to `object::~object()` as well?

asarium:
I'll fix the typo.

However, it is not necessary to add the clear calls to the destructor since that will be done automatically by the compiler.

Goober5000:
Ok.  Then I guess `objsnd_num.clear();` doesn't need to be there either.

PR #2121, opened 2019-10-22
Fix-2120

### LINKED ISSUE #2120: Player Select Menu has mode selection issues
If the player clicks on the mode that is currently selected in the player start screen, then the bottom text, which says the mode, disappears.

FSCyborg:
Move the bottom text reset to inside the if so that it isn't kept empty by accident.

Let me know if I should move the comment around at all, or if I need to make any other changes.

Designed to fix #2120

### DIFF
--- a/code/menuui/playermenu.cpp
+++ b/code/menuui/playermenu.cpp
@@ -663,13 +663,14 @@ void player_select_button_pressed(int n)
 		break;
 
 	case SINGLE_BUTTON:
-		player_select_set_bottom_text("");
-
 		Player_select_autoaccept = 0;
 		// switch to single player mode
 		if (Player_select_mode != PLAYER_SELECT_MODE_SINGLE) {
 			// play a little sound
 			gamesnd_play_iface(InterfaceSounds::USER_SELECT);
+
+			//  Only do this when changing modes, keeps bottom text from being empty by accident
+			player_select_set_bottom_text("");
 
 			player_select_set_bottom_text(XSTR( "Single Player Mode", 376));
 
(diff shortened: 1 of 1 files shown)


FSCyborg:
Just ran the updated build, and this fixes the issue.

code/menuui/playermenu.cpp
```diff
@@ -663,14 +663,17 @@ void player_select_button_pressed(int n)
 		break;
 
 	case SINGLE_BUTTON:
-		player_select_set_bottom_text("");
-
+
```
asarium: This empty line should be removed.

code/menuui/playermenu.cpp
```diff
@@ -681,7 +684,6 @@ void player_select_button_pressed(int n)
 		break;
 
 	case MULTI_BUTTON:
-		player_select_set_bottom_text("");
```
asarium: Another empty line.

code/menuui/playermenu.cpp
```diff
@@ -663,14 +663,17 @@ void player_select_button_pressed(int n)
...
-
+		
 		Player_select_autoaccept = 0;
 		// switch to single player mode
 		if (Player_select_mode != PLAYER_SELECT_MODE_SINGLE) {
 			// play a little sound
 			gamesnd_play_iface(InterfaceSounds::USER_SELECT);
 
+			// moving this here, from outside the if, keeps bottom text from being empty by accident,
+			// when the player accidentally presses the already-lit mode
```
asarium: A tip for writing comments, write them to be stateless to make them easier to understand without looking at the file history. I would write it something like this:
```suggestion
      // Only do this when changing modes, keeps bottom text from being empty by accident,
			// when the player accidentally presses the already-lit mode
```

code/menuui/playermenu.cpp
```diff
@@ -694,6 +696,10 @@ void player_select_button_pressed(int n)
...
```
asarium: See above.

FSCyborg:
I should have just fixed everything you addressed.

PR #2125, opened 2019-10-25
create matrixes from vectors

Goober5000:
This should add scripting functionality equivalent to `vm_vector_2_matrix`.  I'm not really a scripting guru so this was made by copying and adapting other functions as appropriate.

### DIFF
--- a/code/scripting/api/libs/base.cpp
+++ b/code/scripting/api/libs/base.cpp
@@ -67,6 +67,17 @@ ADE_FUNC(createOrientation, l_Base, "[p/r1c1, b/r1c2, h/r1c3, r2c1, r2c2, r2c3, 
 	return ade_set_error(L, "o", l_Matrix.Set(matrix_h()));
 }
 
+ADE_FUNC(vectorToOrientation, l_Base, "fvec[, uvec, rvec]", "Creates an orientation object from 1, 2, or 3 vectors", "orientation", "Orientation object")
+{
+	vec3d *fvec = nullptr;
+	vec3d *uvec = nullptr;
+	vec3d *rvec = nullptr;
+
+	ade_get_args(L, "o|oo", l_Vector.GetPtr(&fvec), l_Vector.GetPtr(&uvec), l_Vector.GetPtr(&rvec));
+
+	return ade_set_args(L, "o", l_Matrix.Set(matrix_h(fvec, uvec, rvec)));
+}
+
 ADE_FUNC(createVector, l_Base, "[x, y, z]", "Creates a vector object", "vector", "Vector object")
 {
 	vec3d v3 = vmd_zero_vector;
--- a/code/scripting/api/libs/base.cpp
+++ b/code/scripting/api/libs/base.cpp
@@ -67,17 +67,6 @@ ADE_FUNC(createOrientation, l_Base, "[p/r1c1, b/r1c2, h/r1c3, r2c1, r2c2, r2c3, 
 	return ade_set_error(L, "o", l_Matrix.Set(matrix_h()));
 }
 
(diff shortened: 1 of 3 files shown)


code/scripting/api/libs/base.cpp
```diff
@@ -67,6 +67,17 @@ ADE_FUNC(createOrientation, l_Base, "[p/r1c1, b/r1c2, h/r1c3, r2c1, r2c2, r2c3,
 	return ade_set_error(L, "o", l_Matrix.Set(matrix_h()));
 }
 
+ADE_FUNC(vectorToOrientation, l_Base, "fvec[, uvec, rvec]", "Creates an orientation object from 1, 2, or 3 vectors", "orientation", "Orientation object")
```
asarium: There already is a scripting function for turning a vector into a matrix:
https://github.com/scp-fs2open/fs2open.github.com/blob/f34e76bd5510b5dd9437b129addcd3f7e20d8eef/code/scripting/api/objs/vecmath.cpp#L443-L461

That does not support specifying the u and r vectors though so maybe that functionality could be added to that function instead.

Goober5000: I noticed that.  However this is a member function of the vector itself.  I decided it was clearer to specify the f, u, and r vectors all in the parameter list, which called for either a utility function or a new matrix constructor.  The latter made more sense.

asarium: I do not agree here. Duplicating functionality is always prone to rotting over time. We already have a function that does a subset of this so that should be extended instead of creating a new function that has a slightly different functionality.

Goober5000: Except the two functions serve different purposes.  The existing function is for getting the matrix associated with a vector, and this PR is for creating matrices from scratch.

Also, consider the function invocations.  The new function is `create(f, u, r)` whereas extending the existing one would be `f:get(u, r)`.  That's 1) not very intuitive, and 2) does not match existing examples in the code, making things more difficult for script writers.

asarium: Uhh, no. Both functions would do the exact same thing. They even use the same function under the hood. If a script user wants to create a matrix from scratch they can use `createOrientation`.

Why would `f:get(u, r)` not be intuitive? it fits the general scripting API design much more closely than trying to create it as a free function in the base library. In my opinion, the scripting API also does not need to match the way the code does it under the hood.

Goober5000: See above: "I decided it was clearer to specify the f, u, and r vectors all in the parameter list."  You have one vector calling the function and the other two vectors as parameters.  From the perspective of the matrix, why is that one vector distinct from the other two?

And for a scripter looking for examples from the code, the f, u, and v vectors appear in that order in the parameter list.  With the member function, suddenly the u vector appears first and the f is somewhere else.  That is surprising and likely to impair readability.

asarium: The first vector is distinct because it is the only parameter that is required to be there. That is implicitly guaranteed when using the `f:get` syntax but is something that must be documented when using the `create(f, ...)` syntax.

We can discuss for a long time which version is better and probably not arrive at a satisfactory solution. However, since a function already exists that does the same thing as your proposed new function, it does not make sense to add this new function unless the old one would be severely flawed in which case it should be deprecated. Unfortunately, that is not the case here so I don't see a good reason to introduce the new function.

Goober5000: You know the scripting system more than I do, so I'm inclined to defer to you if you won't budge.  My concern is readability, clarity, and ease of use.  For those reasons I think `create(f, ...)` is more suitable than `f:get`.

asarium: If this was a completely new function then I would probably be fine with it but I am strongly against duplicating API functionality which is why I am so adamant about this. I'm sorry if this might be perceived as harsh but I just want to keep the scripting API clean.

Goober5000: Very well.

PR #2128, opened 2019-10-30
fix an oversight with dont-clamp-max-velocity

Goober5000:
There was an extra check specifically for reverse velocity that I missed.


z64555:
LGTM

PR #2129, opened 2019-11-01
Update Travis to GCC 9 and fix warnings

asarium:
Only one warning type this time, the implicit copy operator for types
with a copy constructor has been deprecated so we need to delcare them
ourself.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -7100,13 +7100,13 @@ void mission_eval_arrivals()
 				if (rship >= 0)
 				{
 					int j;
-					char message_name[NAME_LENGTH + 10];
+					SCP_string message_name;
 					sprintf(message_name, "%s Arrived", wingp->name);
 
 					// see if this wing has an arrival message associated with it
 					for (j = 0; j < MAX_BUILTIN_MESSAGE_TYPES; j++)
 					{
-						if (!stricmp(message_name, Builtin_messages[j].name))
+						if (!stricmp(message_name.c_str(), Builtin_messages[j].name))
 						{
 							message_send_builtin_to_player(j, &Ships[rship], MESSAGE_PRIORITY_LOW, MESSAGE_TIME_SOON, 0, 0, -1, -1);
 							break;
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -2104,9 +2104,9 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		}
 		// now check miss factors for each IFF
 		for(iff=0; iff<Num_iffs; iff++) {
-			char miss_factor_string[NAME_LENGTH + 15];
+			SCP_string miss_factor_string;
 			sprintf(miss_factor_string, "+%s Miss Factor:", Iff_info[iff].iff_name);
-			if(optional_string(miss_factor_string)) {
+			if(optional_string(miss_factor_string.c_str())) {
 				// this Miss Factor applies only to the specified IFF
 				for(idx=0; idx<NUM_SKILL_LEVELS; idx++) {
 					if(!stuff_float_optional(&wip->b_info.beam_iff_miss_factor[iff][idx])) {
--- a/qtfred/src/mission/dialogs/MissionGoalsDialogModel.cpp
+++ b/qtfred/src/mission/dialogs/MissionGoalsDialogModel.cpp
@@ -11,50 +11,43 @@ namespace dialogs {
 MissionGoalsDialogModel::MissionGoalsDialogModel(QObject* parent, fso::fred::EditorViewport* viewport) :
 	AbstractDialogModel(parent, viewport) {
 }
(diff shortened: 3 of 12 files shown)


code/mission/missionparse.cpp
```diff
@@ -7100,13 +7100,13 @@ void mission_eval_arrivals()
 				if (rship >= 0)
 				{
 					int j;
-					char message_name[NAME_LENGTH + 10];
+					SCP_string message_name;
 					sprintf(message_name, "%s Arrived", wingp->name);
```
Goober5000: Was this file intended to be part of this PR?  And incidentally, does sprintf support SCP_string?

code/weapon/weapons.cpp
```diff
@@ -2104,9 +2104,9 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		}
 		// now check miss factors for each IFF
 		for(iff=0; iff<Num_iffs; iff++) {
-			char miss_factor_string[NAME_LENGTH + 15];
+			SCP_string miss_factor_string;
 			sprintf(miss_factor_string, "+%s Miss Factor:", Iff_info[iff].iff_name);
```
Goober5000: Same here

code/mission/missionparse.cpp
asarium: This is intentional. GCC 9 detects more possible buffer overruns when using `sprintf` so I also needed to fix that. `sprintf` support for `SCP_string` is part of FSO and was implemented some time ago.

qtfred/src/mission/dialogs/MissionGoalsDialogModel.cpp
```diff
@@ -11,50 +11,43 @@ namespace dialogs {
...
 	AbstractDialogModel(parent, viewport) {
 }
-bool MissionGoalsDialogModel::apply() {
-	char buf[256], names[2][MAX_GOALS][NAME_LENGTH];
-	int i, count;
-
-	for (i=0; i<Num_goals; i++)
+bool MissionGoalsDialogModel::apply()
+{
+	for (int32_t i = 0; i < Num_goals; i++)
```
Goober5000: Why are you using `int32_t` instead of `int` or `auto` or `size_t`?

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -3945,10 +3945,10 @@ sexp_list_item* sexp_tree::get_listing_opf_ship_wing_shiponteam_point() {
 	sexp_list_item head;
 
 	for (i = 0; i < Num_iffs; i++) {
-		char tmp[NAME_LENGTH + 7];
+		SCP_string tmp;
 		sprintf(tmp, "<any %s>", Iff_info[i].iff_name);
-		strlwr(tmp);
-		head.add_data_dup(tmp);
+		std::transform(begin(tmp), end(tmp), begin(tmp), [](char c) { return (char)::toupper(c); });
```
Goober5000: This transformation uses `toupper`, but the previous one used `strlwr`.

qtfred/src/mission/dialogs/MissionGoalsDialogModel.cpp
asarium: I am trying to use fixed width integers wherever possible to make the size consistent.

qtfred/src/ui/widgets/sexp_tree.cpp
asarium: Indeed, will fix.

qtfred/src/mission/dialogs/MissionGoalsDialogModel.cpp
Goober5000: But this is a `for` loop. (As are the other instances.) It's not a struct or a memory map. Why should that matter?

asarium: To be honest, that is just the coding standard of my work so I used it out of habit. However, I think it is a worthwhile endeavour to use fixed with types wherever possible so I would prefer to leave it like it is.

qtfred/src/ui/widgets/sexp_tree.cpp
asarium: Fixed now.

qtfred/src/mission/dialogs/MissionGoalsDialogModel.cpp
Goober5000: I wouldn't normally care, but for two reasons:

1) You've been proactive about using `auto` in many other places
2) There have only ever been a maximum of 5 goals.  A three-bit data type would be sufficient.  (The dialog has room for 10 so you could go to a whopping four bits.)  Even a `short` would be suitable.  But you're using a 32-bit type.

asarium: I guess I could use auto instead although that is even longer than `int`...

PR #2131, opened 2019-11-02
Address Coverity 1297186,187,188, 190

FSCyborg:
These Coverity issues are all caused by the same code being copy and pasted in three different places.

There does not seem to be actual danger of the negative array access, but the changes that "fix" it are also slight optimizations (placing a few variable assignments outside loops so that it only has to call the function once).

In short, I am basically just moving code around to avoid duplicate checks that coverity is suspicious of, and in other places checking for the negative array index explicitly.

### DIFF
--- a/code/network/multi_voice.cpp
+++ b/code/network/multi_voice.cpp
@@ -1082,12 +1082,14 @@ void multi_voice_player_send_stream()
 	msg_mode = (ubyte)Multi_voice_send_mode;
 	// get the specific target if we're in MSG_TARGET mode
 	target_index = -1;
+	ushort target_signature = 0;  // Cyborg17 - 0 is the invalid value for net_signature
 	if(msg_mode == MULTI_MSG_TARGET){
 		if(Player_ai->target_objnum != -1){
 			target_index = multi_find_player_by_object(&Objects[Player_ai->target_objnum]);
 			if(target_index == -1){
 				return;
 			}
+			target_signature = Objects[Net_players[target_index].m_player->objnum].net_signature;
 		} else {
 			return;
 		}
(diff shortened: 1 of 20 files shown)


FSCyborg:
Please hold while I fix the tiny thing I did wrong.

code/network/multi_voice.cpp
```diff
@@ -1774,12 +1777,30 @@ void multi_voice_client_send_pending()
 
 	// stream all buffered up packets
 	str = &Multi_voice_stream[0];
+
+	// get the current messaging mode
+	msg_mode = (ubyte)Multi_voice_send_mode;
+
+	// get the specific target if we're in MSG_TARGET mode
+	target_index = -1;
+	target_signature = 0; // Cyborg17 - 0 is the invalid value for net_signature
```
asarium: Same as before.

code/network/multi_voice.cpp
```diff
@@ -1824,9 +1832,10 @@ void multi_voice_client_send_pending()
 
 		// add the routing data and any necessary targeting information
 		ADD_DATA(msg_mode);
-		if(msg_mode == MULTI_MSG_TARGET){
-			Assert(Game_mode & GM_IN_MISSION);
-			ADD_USHORT(Objects[Net_players[target_index].m_player->objnum].net_signature);
+
+		// Cyborg17 - add the target signature only if it's been set, which only happens in MSG_TARGET mode
+		if (target_signature){
```
asarium: Same as before.

FSCyborg:
> Can you change them from `while(0)` to `while(false)`?

I went ahead and looked for all instances in the code that used while(0) and switched them over so that those won't give us a problem in the future either.  I put them in a separate commit so that it can be reverted separately if needed, or switched to another PR fairly easily, if requested.

If you guys would like me to separate those changes, or limit them to the macros involved, please let me know.

Goober5000:
I think that's fine.  I would have put them in a different PR, but what actually matters is the fact that they're in a separate commit.  If one ever needs to track down a bug using `git bisect`, having different changes in different commits is ideal.

That being said, you should roll back the changes in `lib/libpng/pngdebug.h` and `lib/libsdl/include/SDL_assert.h`.  Those are third-party libraries so they should be left in their original state when possible.  Since all you changed were two comments, rolling them back should do no harm.

FSCyborg:
Should now be fixed.

PR #2132, opened 2019-11-03
Coverity1370825

FSCyborg:
Covertity is complaining about us initializing members of the verts array while the first vert has members that are uninitialized.

So this is just a quick change to us excplicitly initializing r, b, g, and a of verts 1-4, from the r, b, g, and a that we already set in vert 0.

### DIFF
--- a/code/fireball/warpineffect.cpp
+++ b/code/fireball/warpineffect.cpp
@@ -49,11 +49,9 @@ void warpin_queue_render(model_draw_list *scene, object *obj, matrix *orient, ve
 
 	vm_vec_scale_add( &center, pos, &orient->vec.fvec, -(max_radius/2.5f)/3.0f );
 
-	verts[0].r = 255;
-	verts[0].g = 255;
-	verts[0].b = 255;
-	verts[0].a = 255;
-
+	// Cyborg17, Initialize the *whole* struct, set some starting values and share with the rest of the array.
+	verts[0] = {};
+	verts[0].r = verts[0].g = verts[0].b = verts[0].a = 255;
 	verts[1] = verts[2] = verts[3] = verts[4] = verts[0];
 
 	if (warp_glow_bitmap >= 0) {


FSCyborg:
Finally a pull request where I don't need to make any corrections!

asarium:
:+1:

Goober5000:
Actually, the best solution is probably to memset the entire `verts[0]` struct to 0 before doing the rgb assignment.  Then `// NOLINT` won't be needed.

FSCyborg:
> Actually, the best solution is probably to memset the entire `verts[0]` struct to 0 before doing the rgb assignment. Then `// NOLINT` won't be needed.

Glad I looked again before starting on the changes....

FSCyborg:
I will be able to clean up the commit history once I get home, please hold on merging.

chief1983:
I thought memset was actually fairly high in terms of code risk.

Goober5000:
`memset` is a perfectly cromulent solution.  It's only a problem if used on a class, in which case it would wipe the function table.  Since `vertex` is a regular C-style struct with no class functions, there is no danger.

Moreover, we now have compile-time code checks which flag the use of `memset`, `memcpy`, etc. on classes.

chief1983:
It is one case in which it is dangerous, yes, but there are others.  https://www.viva64.com/en/b/0360/

Not saying to avoid it but just that it might always be worth considering for a moment if it is the right/safe solution in a given case, beyond just whether classes are in play.

FSCyborg:
This looks salient: https://stackoverflow.com/questions/1998752/memset-or-value-initialization-to-zero-out-a-struct

Maybe I can just do 
```
vertex verts[5];
verts[0] = {};
verts[1] = verts[2] = verts[3] = verts[4] = verts[0]; ?
```

Goober5000:
> It is one case in which it is dangerous, yes, but there are others. https://www.viva64.com/en/b/0360/

That's an interesting article, but nearly all of its examples are based on misusing the `sizeof` operator.  The only exceptions are switching the arguments (which could happen with any function) and zeroizing sensitive information (which is the fault of the optimizer).  So it's a cautionary tale on `sizeof`, not `memset`.

The problems with `sizeof` can be easily avoided by using this pattern:

`number_of_elements * sizeof(type)`

Using `sizeof` on pointers, instance variables, or passed arrays is the risky approach and is what should be avoided.


> Maybe I can just do this?

That should work too.  Mind you, the rgba assignment would come between the second and third lines of your code snippet.

FSCyborg:
Not sure what you mean. It's on the third line.

Goober5000:
You must have edited it after I started composing my post.  Originally there were just three lines, without the rgba assignment.

FSCyborg:
And now that this tiny, tiny problem has consumed so much time, let's hope the builds pass.

Issue #1682, opened 2018-04-13
dumb rotate isn't affecting jump nodes


AxemP:
We used to be able to pull off some neat things with dumb rotate, like rotating jump nodes, or subtly moving clouds in a skybox. This appears to not work anymore. By my very rough code reading skills in modelread.cpp, it seems it's only allowed for ships now? Can this be widened to all models once again?

Please see sample mod. The everything (except the player) has a dumb rotate part to it. So part of the skybox should move, the jumpnode should move. The candy bar already rotates, but that's there to show that dumb rotate isn't completely broken. Just kinda.

[dumbrotate.zip](https://github.com/scp-fs2open/fs2open.github.com/files/1905402/dumbrotate.zip)

Goober5000:
This could have been caused by my changes in #512, but that PR received extensive review and non-ship rotating models were specifically mentioned.

Goober5000:
From the #scp chat, it looks like @TRBlount is looking at it.  My assumption is that any model which has a $dumb_rotate (or $dumb_rotate_time) property will have the intrinsic-rotate flag automatically assigned, but I won't have time to look at this until tomorrow.

Goober5000:
My assumption turned out to be correct.  Here's the relevant code excerpt from modelread.cpp...

```
				// note, this should come BEFORE do_new_subsystem() for proper error handling (to avoid both rotating and dumb-rotating submodel)
				int idx = prop_string(props, &p, "$dumb_rotate_time", "$dumb_rotate_rate", "$dumb_rotate");
				if (idx >= 0) {
					pm->submodel[n].movement_type = MOVEMENT_TYPE_INTRINSIC_ROTATE;
					pm->flags |= PM_FLAG_HAS_INTRINSIC_ROTATE;
```

So the bug must be caused by something else, e.g. the model instance creation that someone mentioned on Discord.

Do we know when this bug was first introduced?

Goober5000:
@AxemP can you see if you can narrow down the timeframe when this bug first appeared?

FSCyborg:
So, I just took a good look at everything, and the reason why jump nodes do not use intrinsic rotate is because jump nodes are not added to the polymodel instance list.  I just finished a prototype of the a patch adding that functionality, and I need to test it this weekend.  (It required a few functions be added to the jumpnode class) 

I am still trying to figure out why skyboxes are not moving.  I don't like how roundabout the check is, but I can't yet see how it could be broken.

[...]

PR #2136, opened 2019-11-10
Fix Dumb_Rotate Regression

FSCyborg:
Allows jump nodes to use instrinsic/dumb rotation again by exposing the jump node class to the model instancing code and adding a check for it in modelread.  For #1682

### DIFF
--- a/code/jumpnode/jumpnode.cpp
+++ b/code/jumpnode/jumpnode.cpp
@@ -32,7 +32,8 @@ CJumpNode::CJumpNode() : m_radius(0.0f), m_modelnum(-1), m_objnum(-1), m_flags(0
 /**
  * Constructor for CJumpNode class, with world position argument
  */
-CJumpNode::CJumpNode(vec3d *position) : m_radius(0.0f), m_modelnum(-1), m_objnum(-1), m_flags(0)
+CJumpNode::CJumpNode(vec3d* position)
+    : m_radius(0.0f), m_modelnum(-1), m_objnum(-1), m_flags(0)
 {	
 	Assert(position != NULL);
 	
@@ -43,10 +44,18 @@ CJumpNode::CJumpNode(vec3d *position) : m_radius(0.0f), m_modelnum(-1), m_objnum
 	
 	// Set m_modelnum and m_radius
 	m_modelnum = model_load(NOX("subspacenode.pof"), 0, NULL, 0);
-	if (m_modelnum == -1)
+	if (m_modelnum == -1) {
 		Warning(LOCATION, "Could not load default model for %s", m_name);
-	else
+	} else {
 		m_radius = model_get_radius(m_modelnum);
+
+		// set up animation in case of instrinsic_rotate
+		polymodel* pm = model_get(m_modelnum);
+
+		if (pm->flags & PM_FLAG_HAS_INTRINSIC_ROTATE) {
+			m_polymodel_instance_num = model_create_instance(false, m_modelnum);
+		}
+	}
 	
     m_pos.xyz.x = position->xyz.x;
     m_pos.xyz.y = position->xyz.y;
@@ -59,11 +68,12 @@ CJumpNode::CJumpNode(vec3d *position) : m_radius(0.0f), m_modelnum(-1), m_objnum
 }
 
 CJumpNode::CJumpNode(CJumpNode&& other) SCP_NOEXCEPT
-	: m_radius(other.m_radius), m_modelnum(other.m_modelnum), m_objnum(other.m_objnum), m_flags(other.m_flags)
+	: m_radius(other.m_radius), m_modelnum(other.m_modelnum), m_objnum(other.m_objnum), m_polymodel_instance_num(other.m_polymodel_instance_num), m_flags(other.m_flags)
 {
 	other.m_radius = 0.0f;
 	other.m_modelnum = -1;
 	other.m_objnum = -1;
+	other.m_polymodel_instance_num = -1;
 	other.m_flags = 0;
 
 	m_display_color = other.m_display_color;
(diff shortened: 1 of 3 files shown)


code/model/modelrender.cpp
```diff
@@ -2628,12 +2629,16 @@ void model_render_queue(model_render_params* interp, model_draw_list* scene, int
...
-			if (objp->type == OBJ_ASTEROID)
+			if (objp->type == OBJ_ASTEROID) {
 				pmi = model_get_instance(Asteroids[objp->instance].model_instance_num);
-			else if (objp->type == OBJ_WEAPON)
+			} else if (objp->type == OBJ_WEAPON) {
 				pmi = model_get_instance(Weapons[objp->instance].model_instance_num);
-			else
+			} else if (objp->type == OBJ_JUMP_NODE) {
+				CJumpNode *jnp = jumpnode_get_by_objnum(objnum);
+				pmi = model_get_instance(jnp->GetPolymodelInstanceNum());
```
asarium: An assert to make sure `jnp` is initialized here would be good:
```suggestion
				Assertion(jnp != nullptr, "Could not find jump node with object number %d!", objnum);
				pmi = model_get_instance(jnp->GetPolymodelInstanceNum());
```

Also, the `*` of a pointer type should be aligned to its type and not the variable.

code/jumpnode/jumpnode.cpp
```diff
@@ -392,6 +405,25 @@ CJumpNode *jumpnode_get_by_name(const char* name)
...
+ *
+ * @param objnum to search for
+ * @return Jump node object pointer
+ */
+CJumpNode *jumpnode_get_by_objnum(int objnum)
+{
+	Assert(objnum > -1);
+	SCP_list<CJumpNode>::iterator jnp;
+
+	for (jnp = Jump_nodes.begin(); jnp != Jump_nodes.end(); ++jnp) {
```
asarium: You should be able to use a for-each loop here.

FSCyborg: Do I use `Jump_nodes.size()` to indicate the range?

asarium: You should be able to do something like this:
```c++
for (const CJumpNode& jnp : Jump_nodes)
```

FSCyborg: It didn't like const, but it should be this type now. I should compile and test one more time to make sure it works.

FSCyborg:
Please let me compile and test one more time in case it's ready to be merged, thanks!

Goober5000:
Clang doesn't like something:

```
Running clang-tidy on changed files

/home/travis/build/scp-fs2open/fs2open.github.com/build/../code/jumpnode/jumpnode.h:39:6: warning: use default member initializer for 'm_polymodel_instance_num' [modernize-use-default-member-init]

        int m_polymodel_instance_num;   // polymodel instance number, used for rotations 
```

And I want to take a close look at this as well.  It would be great to get Swifty (@SamuelCho) if he's still around.

asarium:
The clang warning just means that it would like you to initialize the member in the header and not in the source file.

FSCyborg:
I've gotten this warning before and put it into the header file before, but why isn't it warning me about all the other members that are initialized the same way?

code/jumpnode/jumpnode.cpp
```diff
@@ -56,14 +57,18 @@ CJumpNode::CJumpNode(vec3d *position) : m_radius(0.0f), m_modelnum(-1), m_objnum
     flagset<Object::Object_Flags> default_flags;
     default_flags.set(Object::Object_Flags::Renders);
     m_objnum = obj_create(OBJ_JUMP_NODE, -1, -1, NULL, &m_pos, m_radius, default_flags);
+
+	// And finally set up animation info for possible instrinsic_rotate
+	m_polymodel_instance_num = model_create_instance(false, m_modelnum);
```
Goober5000: The model has been loaded at this point, so you can check the model flags to see whether `PM_FLAG_HAS_INTRINSIC_ROTATE` is set before creating the instance.

code/jumpnode/jumpnode.h
```diff
@@ -36,6 +36,7 @@ class CJumpNode
 
     int	m_modelnum;
     int	m_objnum;						// objnum of this jump node
+	int m_polymodel_instance_num {-1}; // polymodel instance number, used for rotations
```
Goober5000: In order to keep things consistent, I recommend adding default initializers to the rest of the class fields too.  Then you can remove the initializers from the `CJumpNode()` and `CJumpNode(vec3d* position)` constructors.

FSCyborg: Good point.

FSCyborg: Changes made.  If you want them squashed instead of a separate commit that's fine.  I just figured it would help anyone tracing the history to have a separate commit here when figuring out why the changes were made.

code/jumpnode/jumpnode.cpp
FSCyborg: I was able to test, and it still works.  👍

FSCyborg:
It still works.  Let's see if the tests like and it.  How long should we wait for SamuelCho?

code/jumpnode/jumpnode.h
Goober5000: Good. :+1: No need to squash; I don't think squashing is required in most cases.

Goober5000:
He isn't active these days so I wouldn't depend on him except in extreme circumstances, like for PR #2056.  Before I reviewed this, I was worried that it would require some esoteric knowledge about the model instance system that only he could provide; but the fix actually turned out to be quite simple.  So I'm comfortable approving this without his input.

PR #2142, opened 2019-11-12
Auto port forwarding

notimaginative:
This has been discussed before but I thought I might actually code something up and test it out. If someone else has done more research or working on this type of stuff previously then please provide your input.

After looking at the various options I decided to implement just NAT-PMP. In part because it's much cleaner and easier to do compared to UPnP. And if I was going to implement just one then NAT-PMP seems like the better option. PCP, the large-scale successor to small-scale NAT-PMP, should also let us get through carrier-grade NAT, which is another reason I went with NAT-PMP (although it's all gateway dependent). The only reason I can see to implement UPnP as well just for compatibility with routers that don't support NAT-PMP or just don't have it enabled.

Regardless of whether it's NAT-PMP or UPnP it still requires router/gateway support, plus actually being enabled. So there still might be something required on the user's side. Everything after that is automatic however, with firewall and DNAT rules being added and removed as needed.

Further reading for the curious:
[NAT-PMP](https://en.wikipedia.org/wiki/NAT_Port_Mapping_Protocol)
[PCP](https://en.wikipedia.org/wiki/Port_Control_Protocol)
[UPnP-IGD](https://en.wikipedia.org/wiki/Internet_Gateway_Device_Protocol)

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -203,6 +203,7 @@ Flag exe_params[] =
 	{ "-multilog",			"",											false,	0,					EASY_DEFAULT,		"Multiplayer",	"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-multilog", },
 	{ "-clientdamage",		"",											false,	0,					EASY_DEFAULT,		"Multiplayer",	"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-clientdamage", },
 	{ "-mpnoreturn",		"Disable flight deck option",				true,	0,					EASY_DEFAULT,		"Multiplayer",	"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-mpnoreturn", },
+	{ "-gateway_ip",		"Set gateway IP address",					false,	0,					EASY_DEFAULT,		"Multiplayer",	"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-gateway_ip", },
 
 	{ "-no_set_gamma",		"Disable setting of gamma",					true,	0,					EASY_DEFAULT,		"Troubleshoot",	"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-no_set_gamma", },
 	{ "-nomovies",			"Disable video playback",					true,	0,					EASY_DEFAULT,		"Troubleshoot",	"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-nomovies", },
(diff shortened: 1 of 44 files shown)


notimaginative:
Refactored the code into a mostly done state. It is now fully async, only does port forwarding when needed, and has a proper renewal process.

Should also be able to add UPnP support at a future time without too much trouble.

notimaginative:
I happy with this now. I would like to put out a test build before saying it's ready though. I'll put something out this weekend and see how it goes. Not really concerned about the code, but what it takes for people to be able to use it (router config stuff).

And after more research I'm not too thrilled about UPnP support. A lot of caveats do deal with there. I was also only able to find one implementation that was fully async and even that was "proof of concept". [miniupnpc-async](https://github.com/miniupnp/miniupnp/tree/master/miniupnpc-async)

code/cmdline/cmdline.cpp
```diff
@@ -405,11 +406,13 @@ cmdline_parm almission_arg("-almission", "Autoload multiplayer mission", AT_STRI
 cmdline_parm ingamejoin_arg("-ingame_join", NULL, AT_NONE);	// Cmdline_ingamejoin
 cmdline_parm mpnoreturn_arg("-mpnoreturn", NULL, AT_NONE);	// Cmdline_mpnoreturn  -- Removes 'Return to Flight Deck' in respawn dialog -C
 cmdline_parm objupd_arg("-cap_object_update", "Multiplayer object update cap (0-3)", AT_INT);
+cmdline_parm gateway_ip_arg("-gateway_ip", "Set gateway IP address", AT_STRING);
 
 char *Cmdline_almission = NULL;	//DTP for autoload multi mission.
 int Cmdline_ingamejoin = 0;
 int Cmdline_mpnoreturn = 0;
 int Cmdline_objupd = 3;		// client object updates on LAN by default
+char *Cmdline_gateway_ip = nullptr;
```
asarium: Please use `SCP_string` for new options since that is easier to use with C++.

code/network/multi_portfwd.cpp
```diff
@@ -0,0 +1,300 @@
...
+static bool PF_natpmp_close();
+static void PF_natpmp_do();
+
+
+void multi_port_forward_init()
+{
+	bool rval;
+	bool renew = false;
+
+	if ( timestamp_elapsed(PF_renewal_timestamp) ) {
```
asarium: The `timestamp_` functions are only usable while within a mission. Is that intentional? If so then the timestamp should at least be reset in `multi_port_forward_close` to make sure that the renewal is done properly. If that is not intentional then you could use the `timer_` functions instead to get a system timestamp.

lib/libnatpmp/LICENSE
```diff
@@ -0,0 +1,26 @@
+Copyright (c) 2007-2011, Thomas BERNARD
```
asarium: This notice should be added to the other libraries in `credits.cpp`.

code/network/multi_portfwd.cpp
notimaginative: Thanks! 17 years working in this code and I'm still messing that up.

code/cmdline/cmdline.cpp
notimaginative: This requires including pstypes.h in cmdline.h. Is that wanted?

notimaginative: Just realized that would duplicate memory too. Might need to rework cmdline code before making the move to `SCP_string`

Goober5000: I wouldn't insist on `SCP_string` here.

Goober5000:
@notimaginative Did you put out a test build like you planned?  And if so, what were the results?

notimaginative:
@Goober5000 I did. No feedback so far. Not sure if that's a good thing or a bad thing.

Goober5000:
@notimaginative Well I was ready to go ahead and merge this, but it would be nice to have some positive test results first.

notimaginative:
@Goober5000 Hold off on this one a bit, until after RC2 at least. I'm experimenting with switching to a different lib that would get us both NAT-PMP and the newer PCP support.

Goober5000:
Ok.

code/cmdline/cmdline.cpp
notimaginative: Going to assume this is no longer an issue. In context it's a pointer to an allocated string in memory and not itself a string.

Goober5000:
@notimaginative Do you still want to hold this off until after RC2?

notimaginative:
@Goober5000 I have no further changes for it so the code is done. My only concern is that there has been only one confirmed test report. So I'll leave it up to you as to whether it goes in or not.

Goober5000:
I'll merge it.  My guess is that there's a better chance players will test it if it's in RC2.

PR #2143, opened 2019-11-13
Fixes 3D Weapon Selection Bug

wookieejedi:
Fixes 3D weapon selection not working when an animation is already specified.


Goober5000:
This is a follow-up to PR #2055.

PR #2146, opened 2019-11-16
Makes slight adjustments to the lighting defaults.

The-E:
This commit disables emissive lighting on objects by default (as the new HDR process makes it unnecessary) and switches the -no-emissive commandline parameter to -emissive; Setting that parameter will restore the old behaviour. The -no-emissive parameter is retained to avoid user-facing error messages.
In addition, this commit sets the default bloom value to 25.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -163,7 +163,7 @@ Flag exe_params[] =
 	{ "-nomotiondebris",	"Disable motion debris",					true,	EASY_ALL_ON,		EASY_DEFAULT,		"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-nomotiondebris",},
 	{ "-noscalevid",		"Disable scale-to-window for movies",		true,	0,					EASY_DEFAULT,		"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-noscalevid", },
 	{ "-nonormal",			"Disable normal maps",						true,	EASY_DEFAULT_MEM,	EASY_MEM_ALL_ON,	"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-normal" },
-	{ "-no_emissive_light",	"Disable emissive light from ships",		true,	0,					EASY_DEFAULT,		"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-no_emissive_light" },
+	{ "-emissive_light",	"Enable emissive light from ships",			true,	0,					EASY_DEFAULT,		"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-no_emissive_light" },
 	{ "-noheight",			"Disable height/parallax maps",				true,	EASY_DEFAULT_MEM,	EASY_MEM_ALL_ON,	"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-height" },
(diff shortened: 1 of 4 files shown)


Goober5000:
This PR is in reference to #1587

code/cmdline/cmdline.cpp
```diff
@@ -28,42 +29,36 @@
 #include "starfield/starfield.h"
 
 #ifdef _WIN32
-#include <io.h>
 #include <direct.h>
+#include <io.h>
```
asarium: As much as I appreciate proper formatting, changing more than 100 lines for a (I guess) 2 line change is a bit excessive.

The-E: It's your clang-format settings!

asarium: I know but please only format the lines that you changed or otherwise tracking down what actually changed is a big hassle.

tomimaki:
> In addition, this commit sets the default bloom value to 25.

Hm, I don't see this change.

The-E:
It seems that change got lost while removing the extraneous formatting changes. This has been fixed.

code/lab/lab.cpp
```diff
@@ -407,14 +407,14 @@ void labviewer_render_model(float frametime)
...
-		auto lab_emissive_light_save  = Cmdline_no_emissive;
+		auto lab_emissive_light_save  = Cmdline_emissive;
 
 		if (Lab_selected_mission == "None") {
 			Lab_render_without_light = true;
 			Motion_debris_enabled    = true;
 		}
 
-		Cmdline_no_emissive = !Lab_emissive_light_override;
+		Cmdline_emissive = !Lab_emissive_light_override;
```
MageKing17: Shouldn't this line remove the `!`, just like line 150?

The-E: You are entirely correct.

code/cmdline/cmdline.cpp
```diff
@@ -163,7 +163,7 @@ Flag exe_params[] =
 	{ "-nomotiondebris",	"Disable motion debris",					true,	EASY_ALL_ON,		EASY_DEFAULT,		"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-nomotiondebris",},
 	{ "-noscalevid",		"Disable scale-to-window for movies",		true,	0,					EASY_DEFAULT,		"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-noscalevid", },
 	{ "-nonormal",			"Disable normal maps",						true,	EASY_DEFAULT_MEM,	EASY_MEM_ALL_ON,	"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-normal" },
-	{ "-no_emissive_light",	"Disable emissive light from ships",		true,	0,					EASY_DEFAULT,		"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-no_emissive_light" },
+	{ "-emissive_light",	"Enable emissive light from ships",			true,	0,					EASY_DEFAULT,		"Graphics",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-no_emissive_light" },
```
Goober5000: Please update the wiki link as well.

The-E: I've updated the link, but it seems I can't log in to the wiki at the moment to change the text; Either the password recovery mail isn't being sent or its being sent to an old address I no longer have access to.

PR #2147, opened 2019-11-17
Add Lua access to not yet arrived ships

asarium:
Requested by @AxemP.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -13,7 +13,7 @@ misc-*,\
 -misc-misplaced-widening-cast,\
 -misc-static-assert,\
 -misc-unused-parameters,\
--misc-non-private-member-variables-in-classes,
+-misc-non-private-member-variables-in-classes,\
 "
 WarningsAsErrors: ''
 HeaderFilterRegex: 'code/.*$|freespace2/.*$|qtfred/.*$|test/src/.*$|build/.*$'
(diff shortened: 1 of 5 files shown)


.clang-tidy
```diff
@@ -4,15 +4,16 @@ modernize-*,\
 -modernize-raw-string-literal,\
 -modernize-use-using,\
 -modernize-make-unique,\
--modernize-use-trailing-return-type\
```
Goober5000: Is `.clang-tidy` supposed to be included in your changeset?

code/scripting/api/objs/parse_object.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+	return ade_set_args(L, "o", l_Matrix.Set(matrix_h(&poh->getObject()->orient)));
+}
+
+ADE_VIRTVAR(ShipClass, l_ParseObject, "shipclass", "The ship class of the object.", "shipclass", "The ship class")
+{
+	parse_object_h* poh = nullptr;
+	int newClass        = -1;
+	if (!ade_get_args(L, "o|o", l_ParseObject.GetPtr(&poh), l_Shipclass.Get(&newClass)))
+		return ade_set_error(L, "o", l_Matrix.Set(matrix_h()));
```
Goober5000: Copy-paste error?  I don't think matrices should be used here and on the next few lines.

PR #2148, opened 2019-11-17
avoid full stream reset when looping music

### LINKED ISSUE #2097: Looping music/sounds have a small pause when looping
When using the `play-sound-from-file` sexp, there's always been this very small pause or gap when the music is supposed to loop back. It'd be super if this could be made so much more seamless and less terrible!

Here's a sample mission of the afterburner loop sound being used in the sexp which shows the problem quite well: 

[soundlooptest.zip](https://github.com/scp-fs2open/fs2open.github.com/files/3640423/soundlooptest.zip)

notimaginative:
fixes #2097


AxemP:
Yes, I certainly do

PR #2149, opened 2019-11-18
Patch n gon radar

FSCyborg:
Axem just reported that his custom hud's radar gauge was broken from when we merged the ngon radar.  Taking a look at the code, it looks like what went wrong is that moving gauge_load_common() down the function for modder's convenience broke gauge_load_common()'s assumed parsing order, forcing the retail gauge to rear its head.

Now, I moved it back up to where it was before ngon radar was merged, and Axem reported it fixed.

This just means that for the radar case, "Ngon Sides:" and "Ngon Offest:" have to come before "Origin:" and "Offset:" in the table.


Goober5000:
This addresses #2150

Issue #2151, opened 2019-11-18
ARRWALKTHEPLANK Crash


FSCyborg:
It's been reported by Rhymes that this cheat crashes on release builds and does nothing on Debug Builds.  Have not had a chance to look at the code for a cause.

FSCyborg:
PR has been merged.  This should be closed! :D

PR #2152, opened 2019-11-18
Bring Pirates of the FSO Back by

FSCyborg:
Preventing a nullptr deep within shipfx_warpin_start()

### DIFF
--- a/code/ship/shipfx.cpp
+++ b/code/ship/shipfx.cpp
@@ -556,6 +556,7 @@ void shipfx_warpin_start( object *objp )
 		return;
 	}
 
+	Assertion(shipp->warpin_effect != nullptr, "shipfx_warpin_start() was fed a ship with an uninitialized warpin_effect.");
 	shipp->warpin_effect->warpStart();
 
 	Script_system.RunCondition(CHA_WARPIN, objp);
@@ -707,6 +708,7 @@ void shipfx_warpout_start( object *objp )
 		return;
 	}
 
+	Assertion(shipp->warpout_effect != nullptr, "shipfx_warpout_start() was fed a ship with an uninitialized warpout_effect.");
 	shipp->warpout_effect->warpStart();
 
 	Script_system.RunCondition(CHA_WARPOUT, objp);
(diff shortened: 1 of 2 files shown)


Goober5000:
Link to issue #2151.

FSCyborg:
Fantastic point.  Will do.

FSCyborg:
I'm not 100% sure if this is the correct way to check for an uninitialized pointer, because I'm not sure it is the same as nullptr.

code/ship/shipfx.cpp
```diff
@@ -707,6 +708,7 @@ void shipfx_warpout_start( object *objp )
 		return;
 	}
 
+	Assertion(shipp->warpin_effect != nullptr,"shipfx_warpout_start() was fed a ship with an uninitialized warpin_effect.");
```
Goober5000: You need to check for `warpout_effect` here, not `warpin_effect`.  Also, as an extremely minor nitpick, put a space between the comma and the quote.

Goober5000:
An uninitialized pointer is not the same as nullptr, but the pointers should not be uninitialized here.  The `ship_clear()` function sets them to nullptr.

code/ship/shipfx.cpp
FSCyborg: Thanks!  Should be good now.

code/ship/shipfx.cpp
```diff
@@ -707,6 +708,7 @@ void shipfx_warpout_start( object *objp )
 		return;
 	}
 
+	Assertion(shipp->warpout_effect != nullptr, "shipfx_warpout_start() was fed a ship with an uninitialized warpin_effect.");
```
Goober5000: It still says warpin_effect at the end of the message.

Goober5000:
It's generally not possible to check for an uninitialized variable.  But debug builds will set them to a known, invalid, but "magic number" value such as `0xcccccccc` to help diagnose problems in the case of a crash.  There's no such help in release builds though.

FSCyborg:
Ok.  Just so you know, any commit I make between 4 and 7 on a weekday I am probably more focused on keeping my son away from the keyboard while I type, than what I am typing....

FSCyborg:
And now??

Issue #1992, opened 2019-02-24
ships with hyperspace warp type are briefly in the wrong radar position at start of their warp


niffiwan:
When a hyperspace warp-type ship warps in, it briefly appears at the wrong radar location. It should start  waaaay off in the distance; however the radar blip briefly appears at its destination before reappearing waaaaay off in the distance.

This is reasonably hard to notice (the radar blip may only be in the wrong place for a single frame?); thus is a minor issue. 

Test mission & modular table attached below.
[bbb-shp.tbm.txt](https://github.com/scp-fs2open/fs2open.github.com/files/2897402/bbb-shp.tbm.txt)
[docktest-hyp.fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/2897403/docktest-hyp.fs2.txt)

Short vid showing the issue:
https://www.youtube.com/watch?v=cjqWHVSUC6M

Goober5000:
This is also fixed by #1991, right?

niffiwan:
unfortunately not; it's a separate issue that I didn't want to hold up the #1991 fixes for

FSCyborg:
I'm building a possible fix now.

FSCyborg:
PR has been merged! This can he closed! :D

PR #2153, opened 2019-11-19
Fix#1992

FSCyborg:
Send the ship to the actual initial position once the engine knows where the ship should end up.

For issue #1992.  Already tested.


FSCyborg:
I realize I should probably explain a little more. The position of the ship during hyperspace warp is calculated by doing some operations to scale_factor, including a power operation and multiplication that all simplifies to -1*scale_factor for the initial frame.  So you can just place -scale_factor in the last argument of vector_scale_add to push it back to where it was supposed to be.

wookieejedi:
Looks good to me, also.

PR #2154, opened 2019-11-20
Keep set-secondary-weapon from using nonsense

FSCyborg:
Set-secondary-weapon and set-secondary-ammo have been up to now been allowed to exceed tabled cargo sizes, as long as you didn't put a value for requested_ammo in the sexp.

The new limits would be kept from the last secondary weapon in the slot, definitely not ideal behavior.

*Have not yet completely tested this fix!*  I only know the changes to one of the functions works for sure.  The changes to set_secondary_ammo need to be tested as well.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -17200,17 +17200,15 @@ void set_primary_ammo (int ship_index, int requested_bank, int requested_ammo, i
 	// Set the number of weapons
 	shipp->weapons.primary_bank_ammo[requested_bank] = requested_ammo ;
 
-	// If a rearm limit has been specified set it.
-	if (rearm_limit >= 0)
-	{
-		// Don't allow more weapons than the bank can actually hold.
-		if (rearm_limit > maximum_allowed) 
-		{
-			rearm_limit = maximum_allowed;
-		}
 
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -17358,17 +17358,13 @@ void set_secondary_ammo (int ship_index, int requested_bank, int requested_ammo,
 	// Set the number of weapons
 	shipp->weapons.secondary_bank_ammo[requested_bank] = requested_ammo ;
 
-	// If a rearm limit has been specified set it.
-	if (rearm_limit >= 0)
+	// Don't allow more weapons than the bank can actually hold. -- Cyborg17 - no matter what.
+	if (rearm_limit > maximum_allowed)
```
MageKing17: Shouldn't this be `if (rearm_limit < 0 || rearm_limit > maximum_allowed)`?

FSCyborg: Absolutely.

I'm now working on matching changes to set_primary_ammo().

FSCyborg:
I only just rebased the code, there are no actual changes since you two reviewed it.

FSCyborg:
As far as my testing can tell, good to go!

PR #2158, opened 2019-11-25
add $Never Viewable flag for cutscenes

Goober5000:
There are now three states for cutscenes: always viewable, where they will appear no matter what; never viewable, where they will never appear; and viewable, which can be modified during the campaign.  Currently cutscenes will go from not-viewable to viewable, but organizing the flags this way will allow a future feature to make cutscenes not-viewable, if that is desired.

There's also a small edit to csv_convert to guard against array overflows.  Not directly related to the feature but it's in the same place and affects the same code.

### DIFF
--- a/code/cutscene/cutscenes.cpp
+++ b/code/cutscene/cutscenes.cpp
@@ -450,7 +450,7 @@ void cutscenes_screen_init()
 	Cutscene_list.clear();
 
 	int u = 0;
-	for (SCP_vector<cutscene_info>::iterator cut = Cutscenes.begin(); cut != Cutscenes.end(); ++cut, ++u)
+	for (auto cut = Cutscenes.begin(); cut != Cutscenes.end(); ++cut, ++u)
 	{
 		int flags = (*cut).flags;
 
--- a/code/cutscene/cutscenes.cpp
+++ b/code/cutscene/cutscenes.cpp
@@ -452,7 +452,7 @@ void cutscenes_screen_init()
 	int u = 0;
 	for (auto cut = Cutscenes.begin(); cut != Cutscenes.end(); ++cut, ++u)
 	{
-		if ((cut->flags[Cutscene::Cutscene_Flags::Viewable] || cut->flags[Cutscene::Cutscene_Flags::Always_viewable]) && !cut->flags[Cutscene::Cutscene_Flags::Never_viewable])
+		if (cut->flags[Cutscene::Cutscene_Flags::Viewable, Cutscene::Cutscene_Flags::Always_viewable] && !cut->flags[Cutscene::Cutscene_Flags::Never_viewable])
 		{
 			Cutscene_list.push_back(u);
 		}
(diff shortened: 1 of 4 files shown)


code/cutscene/cutscenes.h
```diff
@@ -13,12 +13,16 @@
 #include "globalincs/globals.h"
 #include "globalincs/pstypes.h"
 
+#define CF_VIEWABLE			(1<<0)
+#define CF_ALWAYS_VIEWABLE	(1<<1)
+#define CF_NEVER_VIEWABLE	(1<<2)
```
MageKing17: Even though the number of flags is limited, I think it would be better to use a `flagset` just for type safety, although I don't feel strongly enough about it to request changes.

Goober5000: Noted. I would like to dig into flagset functionality in the future, but I stuck with good old C bitfields for this PR because I'm pretty swamped these days.

asarium: I understand not wanting to convert this to a flagset but please use an `enum` to get rid of the preprocessor defines.

code/cutscene/cutscenes.h
```diff
@@ -13,12 +13,16 @@
...
+
 typedef struct cutscene_info
 {
 	char filename[MAX_FILENAME_LEN];
 	char name[NAME_LENGTH];
 	char* description;
-	bool viewable;
+	int flags;
```
asarium: Please use a `uint` or even better a `uint32_t` for bitfields.

code/cutscene/cutscenes.h
Goober5000: Why `enum`?  Do you mean `const`?  An enum is the least suitable option because enums are for distinct values and flags can overlap.

code/cutscene/cutscenes.h
Goober5000: Ok.

code/cutscene/cutscenes.h
asarium: Enums are routinely used for bitflags since they are very similar to macro constants without all the namespacing issues.

Goober5000: This was getting super annoying, so I just changed everything to a flagset.

code/cutscene/cutscenes.cpp
```diff
@@ -441,9 +450,9 @@ void cutscenes_screen_init()
 	Cutscene_list.clear();
 
 	int u = 0;
-	for (SCP_vector<cutscene_info>::iterator cut = Cutscenes.begin(); cut != Cutscenes.end(); ++cut, u++)
+	for (auto cut = Cutscenes.begin(); cut != Cutscenes.end(); ++cut, ++u)
 	{
-		if ((*cut).viewable)
+		if ((cut->flags[Cutscene::Cutscene_Flags::Viewable] || cut->flags[Cutscene::Cutscene_Flags::Always_viewable]) && !cut->flags[Cutscene::Cutscene_Flags::Never_viewable])
```
MageKing17: This can be slightly compactified by replacing the `||` with a combined flag check:
```suggestion
		if (cut->flags[Cutscene::Cutscene_Flags::Viewable, Cutscene::Cutscene_Flags::Always_viewable] && !cut->flags[Cutscene::Cutscene_Flags::Never_viewable])
```

Issue #2166, opened 2019-11-29
Setting ship orientation via scripting sometimes doesn't set properly


wookieejedi:
There is an issue where setting  a ship's orientation via scripting sometimes doesn't set it properly...though this only happens with certain orientations.

I made a test mission and test script which shows it (reproducible). I set a ship's orientation to itself via scripting, but that somehow made the orientation change. To view, run the mission in debug mode and you can see the ship jump in mission and the debug output say that the orientation changed even though it shouldn't have. 

The ship has no orders or velocity, and this bug only happens with certain starting orientations.

This is the relevant section for those who don't want to download files:

```
local ship = mn.Ships["Test Ship"]

print("Checking 1: Test Ship P, B, H = " .. math.deg(ship.Orientation.p) ..", ".. math.deg(ship.Orientation.b) ..", ".. math.deg(ship.Orientation.h) .."...\n")
-->Checking 1: Test Ship P, B, H = 18.886552224974, -2.3717680348047, -22.134541109294...

ship.Orientation = ba.createOrientation(ship.Orientation.p, ship.Orientation.b, ship.Orientation.h)

print("Checking 2: Test Ship P, B, H = " .. math.deg(ship.Orientation.p) ..", ".. math.deg(ship.Orientation.b) ..", ".. math.deg(ship.Orientation.h) .."...\n")
	-->Checking 2: Test Ship P, B, H = 21.738086194015, -2.3717680348047, -22.134541109294...
	
--those two checks should be the same, but they are not equal regarding pitch (for some reason?)
--the ship is not moving and has no orders
--this only happens with certain starting orientations
```

[Orientation Test_fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/3906046/Orientation.Test_fs2.txt)

[ort_test-sct_tbm.txt](https://github.com/scp-fs2open/fs2open.github.com/files/3906047/ort_test-sct_tbm.txt)

wookieejedi:
Also the issue is not with `ba.createOrientation()`, rather the issue is setting a ship's orientation equal to another orientation. In other words,  `ship.Orientation = new_orientation` can cause issues. Though `ship.Orientation = ship.Orientation` works fine.

wookieejedi:
This is related to `angles_2_matrix ` not being completely correct all of the time with angles that is returns from a 3x3 matrix.

FSCyborg:
So, just to clarify, the actual issue was another function which is fixed in 2167?

[...]

PR #2167, opened 2019-11-29
Fixes the extraction of pitch

wookieejedi:
It turns out that `vm_extract_angles_matrix` would return a pitch value that was occasionally not the actual pitch from the 3x3 matrix. The pitch line in the code uses a modified `cosp` as the denominator to calculate the pitch angle. I researched what the calculation should be and found a more accurate way than using the modified `cosp` is to use `( fvec.xyz.x^2 + fvec.xyz.z^2 )^0.5` I tested using the new calculation thoroughly and in each test it returned the proper pitch value from the 3x3 orientation matrix.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -1015,7 +1015,13 @@ angles *vm_extract_angles_matrix(angles *a, const matrix *m)
 	else											//cosine is larger, so use it
 		cosp = m->vec.fvec.xyz.z*cosh;
 
-	a->p = atan2_safe(-m->vec.fvec.xyz.y, cosp);
+	//using the fvec_xz_distance extracts the correct pitch from the matrix --wookieejedi
+	//previously cosp was used as the denominator, but this resulted in some incorrect pitch extractions
+	float fvec_xz_distance;
+
+	fvec_xz_distance = pow((pow((m->vec.fvec.xyz.x), 2.0f) + pow((m->vec.fvec.xyz.z), 2.0f)), 0.5f);
+
+	a->p = atan2_safe(-m->vec.fvec.xyz.y, fvec_xz_distance);
 
 	if (cosp == 0.0f)	//the cosine of pitch is zero.  we're pitched straight up. say no bank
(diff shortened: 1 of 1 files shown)


wookieejedi:
This fixes #2166. I should also note that this shouldn't affect balance in anything since the most the pitch was ever off by was 3 degrees. Also ship orientations seem to rely mainly on the 3x3 matrix, not the p, b, or h values.

code/math/vecmat.cpp
```diff
@@ -1015,7 +1015,13 @@ angles *vm_extract_angles_matrix(angles *a, const matrix *m)
 	else											//cosine is larger, so use it
 		cosp = m->vec.fvec.xyz.z*cosh;
 
-	a->p = atan2_safe(-m->vec.fvec.xyz.y, cosp);
+	//using the fvec_xz_distance extracts the correct pitch from the matrix --wookieejedi
+	//previously cosp was used as the denominator, but this resulted in some incorrect pitch extractions
+	float fvec_xz_distance;
+
+	fvec_xz_distance = pow((pow((m->vec.fvec.xyz.x), 2.0f) + pow((m->vec.fvec.xyz.z), 2.0f)), 0.5f);
```
Goober5000: Try to avoid the `pow` function if at all possible, since it is slow.  This is especially important in the vector math code which is used everywhere, all the time.

First, replace pow(n, 2.0) with n*n.  Then see if the formula can be refactored to use the distance-squared.  If not, use fl_sqrt because that is more efficient than pow(n, 0.5).

wookieejedi: Thanks, unfortunately with everything I tried I was unable to refactor to remove the sqrt.

wookieejedi:
For reference: just as a test, I put a print statement in `vm_extract_angles_matrix` and ran some missions and did not see it popup at all. Thus, it appears `vm_extract_angles_matrix` is not used on a per frame basis like many other matrix functions. Mainly it's only used in some LUA and background related things.

PR #2168, opened 2019-11-30
sanitize parsed multi.cfg settings for PXO

notimaginative:
This will use default settings related to PXO if one of the following conditions is met:
- `multi.cfg` is missing or a required field is missing
- the settings in `multi.cfg` are from retail FS2
- the settings in `multi.cfg` are from FS2NetD

Fresh installs and mods configured for FS2NetD should now work without any extra steps.


notimaginative:
Just to note that this does not make changes specific to standalones. That's whole other thing and will have to be dealt with in a different way.

PR #2169, opened 2019-11-30
fix PXO banners (and inetgetfile stuff)

notimaginative:
This will get the inetgetfile, cftp, and chttpget code working again so that it can be used for PXO banners ("ads" on the chat screen). It does not make the banners clickable as that might lead to issues and requires further discussion.

### DIFF
--- a/code/inetfile/inetgetfile.h
+++ b/code/inetfile/inetgetfile.h
@@ -29,7 +29,7 @@
 class InetGetFile
 {
 public:
-	InetGetFile(char *URL, char *filename, int cf_type = 1 /* CF_TYPE_ROOT */);
+	InetGetFile(const char *URL, const char *filename, int cf_type = 1 /* CF_TYPE_ROOT */);
 	~InetGetFile();
 	bool IsFileReceived();
 	bool IsFileError();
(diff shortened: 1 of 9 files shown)


notimaginative:
A url specified in `multi.cfg` with `+banner_url` would contain a file named `pxobanners.cfg`. The first line of this file is a url which contains the root location of banner images. This url can be HTTP or FTP, but can *not* be HTTPS (for now). The rest of the lines in the file are in pairs, the first specifying the image filename and the second specifying a url to use when the image is clicked on. Any number of these line pairs can be specified and one will be chosen at random for display on the UI.

Example:
```
http://pxo.nottheeye.com/files/banners/
freespace1.pcx
http://www.dsvolition.com/games/
freespace2.pcx
http://www.dsvolition.com/games/
```

code/inetfile/inetgetfile.h
```diff
@@ -29,7 +29,7 @@
 class InetGetFile
 {
 public:
-	InetGetFile(char *URL,char *localfile);
+	InetGetFile(char *URL, char *filename, int cf_type = 1 /* CF_TYPE_ROOT */);
```
asarium: If possible, please use `const char*`.

PR #2171, opened 2019-12-01
Add Documentation Update instructions and moves Building info from Readme to wiki

Kiloku:
Since the info on how to build the project was fragmented, it made more sense to consolidate it in a single place. I chose the wiki because it already had the most complete info of the two.  

This PR also fixes the formatting on a few files.


Goober5000:
I'll go ahead and merge this anyway.

PR #2172, opened 2019-12-01
Fix FRED not working

wookieejedi:
The `On Goals Cleared` scripting hook resulted in FRED not working, thus this PR adds a check for a valid scripting state to `SetHookObjects` and `RunCondition`. Now FRED will run properly. Fixes PR #2159.


wookieejedi:
`RemHookVars` already had `if(LuaState != NULL)` at the start of the function.

PR #2174, opened 2019-12-02
better sanity-checking for repeat_count, and also check trigger_count

Goober5000:
Sanity checking for repeat counts shouldn't crash the game, so change from Error to Warning and use a sensible default.  Also copy the check (with minor modification) to also check trigger counts.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -4890,8 +4890,9 @@ void parse_event(mission * /*pm*/)
 		}
 
 		// sanity check on the trigger count variable
-		if ( event->trigger_count <= 0 ){
-			Warning(LOCATION, "Trigger count for mission event %s is <= 0.\nMust be >= 1!  Setting to 1.", event->name );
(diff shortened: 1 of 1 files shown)


code/mission/missionparse.cpp
```diff
@@ -4869,19 +4869,32 @@ void parse_event(mission * /*pm*/)
...
 		}
-		event->flags |= MEF_USING_TRIGGER_COUNT; 
-	} 
-	else {
+
+		// sanity check on the trigger count variable
+		if ( event->trigger_count <= 0 ){
+			Warning(LOCATION, "Trigger count for mission event %s is <= 0.\nMust be >= 1!  Setting to 1.", event->name );
+			event->trigger_count = 1;
+		}
```
MageKing17: This is wrong; a trigger count of -1 means "infinite", just like a repeat count of -1.

Goober5000: Ok.

PR #2178, opened 2019-12-05
Fix Free look turning in the wrong directions

FSCyborg:
Free look will now move in the same directions as ship controls, whether pitch or yaw, and whether using keyboard or mouse/joystick.

Also, remove an inaccurate comment, since freelook has been enabled.


FSCyborg:
Just re-tested to be sure, and we are good!

wookieejedi:
I'll give this a test, too.

wookieejedi:
This indeed works as expected, where the yaw movements of the mouse or joystick move the ship and camera in the same lateral direction and the pitch movements of the mouse or joystick move the ship and camera in the same vertical direction.

PR #2179, opened 2019-12-05
Fix some CI issues

asarium:
First one is to allow MinGW failures since that is unreliable at the
moment. Second one is a fix for the regex used by clang-tidy so that it
doesn't try to analyze CMake files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -173,6 +173,12 @@ matrix:
       env: CONFIGURATION="Release" CMAKE_TOOLSET=v141_xp CMAKE_GENERATOR="Visual Studio 15 2017"
     - os: windows
       env: CONFIGURATION="Release" CMAKE_TOOLSET="" CMAKE_GENERATOR="Ninja"
+  # There is a GCC compiler bug with MinGW that causes compilation failures on Windows at the moment
+  allow_failures:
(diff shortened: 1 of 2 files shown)


Goober5000:
Have you investigated adding `-fno-asynchronous-unwind-tables` to the MinGW section?
https://stackoverflow.com/questions/43152633/invalid-register-for-seh-savexmm-in-cygwin

asarium:
I looked at it briefly but decided that 1) MinGW isn't that important and 2) I would rather fix this quickly than having to spend another few days trying to fix this.

Goober5000:
Understood, but it looks like a one-line fix, provided you know which line to put the fix on.

asarium:
Ok, I guess that means CI will be blocked for a few more days then...

asarium:
@Goober5000 I tried the flag from the stackoverflow link but that result in obscure linker errors. I have reverted this pull request to the original fix which just ignores failures for MinGW.

Goober5000:
Ok sounds good. :+1: 

I figured a one-line fix was worth trying, but if this is going to be significantly more complicated (which looks to be the case) then I definitely agree with your reasoning.

PR #2185, opened 2019-12-09
Standardize nonstandard "+nocreate" behavior in objecttypes.tbl

MageKing17:
If you modified an existing ship type without using `+nocreate`, it would both modify it and add it to the vector again, leading to (at the very least) wasted memory. I copied over some more standard behavior from other parsing functions to hopefully avoid problems with modular objecttypes.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -4738,27 +4738,46 @@ static engine_wash_info *get_engine_wash_pointer(char *engine_wash_name)
 	return NULL;
 }
 
-static void parse_ship_type()
+static void parse_ship_type(const char *filename, const bool replace)
 {
 	char name_buf[NAME_LENGTH];
-	bool nocreate = false;
-	ship_type_info stp_buf, *stp = NULL;
+	bool create_if_not_found = true;
+	ship_type_info *stp = NULL;
 
 	required_string("$Name:");
 	stuff_string(name_buf, F_NAME, NAME_LENGTH);
 
 	if(optional_string("+nocreate")) {
-		nocreate = true;
+		if(!replace) {
+			Warning(LOCATION, "+nocreate flag used for ship type '%s' in non-modular table", name_buf);
+		}
+		create_if_not_found = false;
 	}
 
 	int idx = ship_type_name_lookup(name_buf);
 	if (idx >= 0)
 	{
+		if (!replace)
+		{
+			Warning(LOCATION, "ship type '%s' already exists in %s; ship type names must be unique.", name_buf, filename);
+			if ( !skip_to_start_of_string_either("$Name:", "#End")) {
+				Int3();
+			}
+			return;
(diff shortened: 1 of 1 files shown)


code/ship/ship.cpp
```diff
@@ -4738,27 +4738,46 @@ static engine_wash_info *get_engine_wash_pointer(char *engine_wash_name)
...
 }
 
-static void parse_ship_type()
+static void parse_ship_type(const char *filename, const bool replace)
 {
 	char name_buf[NAME_LENGTH];
-	bool nocreate = false;
-	ship_type_info stp_buf, *stp = NULL;
+	bool create_if_not_found = true;
+	ship_type_info *stp = NULL;
```
Goober5000: CI is going to insist you change `NULL` to `nullptr`, by the way...

code/ship/ship.cpp
```diff
@@ -4738,27 +4738,46 @@ static engine_wash_info *get_engine_wash_pointer(char *engine_wash_name)
...
 	}
 
 	int idx = ship_type_name_lookup(name_buf);
 	if (idx >= 0)
 	{
+		if (!replace)
+		{
+			Warning(LOCATION, "ship type '%s' already exists in %s; ship type names must be unique.", name_buf, filename);
+			if ( !skip_to_start_of_string_either("$Name:", "#End")) {
+				Int3();
```
Goober5000: `Int3()` considered harmful.  Can you print a debug message here instead?

code/ship/ship.cpp
```diff
@@ -4738,27 +4738,46 @@ static engine_wash_info *get_engine_wash_pointer(char *engine_wash_name)
...
+		}
 		stp = &Ship_types[idx];
 	}
 	else
 	{
-		stp = &stp_buf;
+		if (!create_if_not_found && replace)
+		{
+			if ( !skip_to_start_of_string_either("$Name:", "#End")) {
+				Int3();
```
Goober5000: Ditto.

code/ship/ship.cpp
MageKing17: It's copied from identical blocks elsewhere and I didn't want to change how every XMT table was parsed (nor change just this one instance because standardization is the point).

Goober5000: Fair enough.  Release versions of FSO are going to CTD in this situation, but if it's like that for every XMT, I'm not going to single out this PR.

MageKing17: Release builds don't crash at `Int3()`, and if they do that's extremely odd behavior because when `NDEBUG` is defined, this is the definition of `Int3()`:
```cpp
#define Int3() do { } while (false)
```

Goober5000: o rly.  I thought I remember people complaining about CTD's that were caused by `Int3()`?

Issue #2183, opened 2019-12-08
set-object-speed-x/y/z is Mostly Non-functional, or Malfunctional


Baezon:
I started taking detailed notes on each of their parameters and the resulting behavior, but I realized that this was probably a waste of time when a coder could look more clearly on what was actually happening. Regardless, I will post what I have gathered until this point

Relative Z (set-object-speed-z with the 3rd argument set to true) does nothing.
Relative Y seems to work
Absolute Z and Y (3rd set to false) do some very funky things that are very difficult to describe, but are clearly doing some wrong math on the ships orientation; sometimes you get nothing, sometimes get just a tiny movement, sometimes it looks normal all depending on the ship's orientation.

Didn't test X. All done on 19 RC1 with MVPs 4

wookieejedi:
Just to check, were you testing these by running them every frame? These sexps will only work once per frame.

Baezon:
I was running them once and seeing the effect they had, I am aware they run only once per frame.

wookieejedi:
Excellent. Also, the code shows that set-object-speed-z will set it for a frame but then the speed will reset to whatever the throttled speed of the ship is, though this is intended behavior. I'd be happy to look at a test mission if you want, too. :)

Baezon:
My test mission really just applies set obect speed. My point about setting relative z is that it does nothing, not that it goes away quickly. It could never with any orientation get it to move the ship by any amount in any direction for any amount of time. This of course leaving aside speeds in wrongs directions and wrong magnitudes.

wookieejedi:
Yes I confirm all of these odd behaviors. Recommend with anyone following along to use ship-maneuver instead.

wookieejedi:
Once #2186 is merged this issue can be closed.

PR #2186, opened 2019-12-09
Deprecates `set-object-speed` in favor of `ship-maneuver`

wookieejedi:
Per discussion on Discord and and Github (#2183), `set-object-speed`  does not work properly and the fix would involve making it act the same way as `ship-maneuver`. From discussions with Goober (who originally made it), we have decided to begin the process of deprecating the sexp. I have talked with folks who have used it thus I am not removing it from FRED with this PR but I am open to discussion on that point.

Sorry about the typos in the actual commit comment below, forgot to change that one before committing.


wookieejedi:
> Please use the "hidden in FRED" technique that other deprecated sexps use. Look at `OP_HITS_LEFT_SUBSYSTEM` for example.

Just updated, thanks for the example.

Issue #1629, opened 2018-02-27
String Mismatch in controlconfiguredefaults.tbl


wookieejedi:
Comments in strings.tbl state 'The control config action text goes between 507 and 619.' There are six controls that spill out of this range and overlap with pre-existing string.tbl indexes. These are VIEW_TOPDOWN, VIEW_TRACK_TARGET, AUTO_PILOT_TOGGLE, NAV_CYCLE, TOGGLE_GLIDING, and CYCLE_PRIMARY_WEAPON_SEQUENCE. Thus, when Has Xstr: 1 in controlconfigdefaults.tbl that control will call a string that already has string assigned. For example, VIEW_TOPDOWN uses string index 620, which should be reserved for that control but is already reserved for the string "Alpha channel textures". This results in the controls menu showing "Alpha channel textures" as the string. 

EDIT: Controls indexes are reserved from 507 to 619, and 620-627 have GrD3D strings. There doesn't appear to be any un-allocated section after that. The default table ends at index 1569.

wookieejedi:
This can be closed now that #2187 is approved.

PR #2187, opened 2019-12-10
Method to fix controls translation issue

wookieejedi:
This PR adds functionality to `Has Xstr:` in `controlconfiguredefaults.tbl`. Previously this value was used as a boolean, 0 or 1, to specify if the string index in `strings.tbl` should be used. This PR treats that field as an integer, and thus allows modders to specify which string to use for the control. If the value is 0 (default) then the string is not used, if the value is 1 then the default string is used and if the value is greater then 1 the game will use that specified index in `strings.tbl`.

This allows modders to properly translate all control names.  Specifically, six controls overlapped with pre-existing `stringa.tbl` indexes and thus could not be translated properly. These are `VIEW_TOPDOWN, VIEW_TRACK_TARGET, AUTO_PILOT_TOGGLE, NAV_CYCLE, TOGGLE_GLIDING, and CYCLE_PRIMARY_WEAPON_SEQUENCE`.

This PR will allow modders to now properly have full control name translations. It also provides a fix for issue #1629.

Also I tested this and confirmed it works as expected.


Goober5000:
There are a whole bunch of compile errors:
```
../code/controlconfig/controlsconfigcommon.cpp:193:1: error: could not convert ‘{8242, -1, COMPUTER_TAB, true, "Toggle Auto Speed Matching", CC_TYPE_TRIGGER, -1, -1, 0, false, false}’ from ‘<brace-enclosed initializer list>’ to ‘config_item’
```
etc.  Could be a compiler idiosyncrasy.

wookieejedi:
> There are a whole bunch of compile errors:
> 
> ```
> ../code/controlconfig/controlsconfigcommon.cpp:193:1: error: could not convert ‘{8242, -1, COMPUTER_TAB, true, "Toggle Auto Speed Matching", CC_TYPE_TRIGGER, -1, -1, 0, false, false}’ from ‘<brace-enclosed initializer list>’ to ‘config_item’
> ```
> 
> etc. Could be a compiler idiosyncrasy.

Yeah I was trying to read through the issues. When I complied and testing on my machine everything worked as expected with no errors. Hmmm

code/controlconfig/controlsconfig.h
```diff
@@ -62,7 +62,7 @@ typedef struct config_item {
 	short key_default;		//!< default key bound to action
 	short joy_default;		//!< default joystick button bound to action
 	char tab;				//!< what tab (category) it belongs in
-	bool hasXSTR;			//!< whether we should translate this with an XSTR
+	int hasXSTR = 0;		//!< what string index we should use to translate this with an XSTR
```
Goober5000: Try this.  Take the zero assignment out of the struct and put it right before the `if (optional_string("$Has XStr:")) {` parsed line.

wookieejedi: Will do, thanks!

Goober5000:
Oh!  No, sorry, I know exactly what the problem is.  The initializer arguments are not the same type as the struct.

What you should do is first remove the zero assignment from the parsing section (because otherwise it will overwrite the built-in settings).  Then look through the control config array and change every `false` to `0`.

code/controlconfig/controlsconfig.h
asarium: Also, please either update the name of the field so that it doesn't imply a boolean value anymore or use two fields, one for determining whether the item has an XSTR and a second one for specifying the index.

code/controlconfig/controlsconfig.cpp
```diff
@@ -516,7 +516,9 @@ void control_config_list_prepare()
 	Num_cc_lines = y = z = 0;
 	while (z < CCFG_MAX) {
 		if (Control_config[z].tab == Tab && !Control_config[z].disabled) {
-			if (Control_config[z].hasXSTR) {
+			if (Control_config[z].hasXSTR > 1) {
+				Cc_lines[Num_cc_lines].label = XSTR(Control_config[z].text, Control_config[z].hasXSTR);
```
asarium: I don't like using `XSTR` with a non constant parameter set since it makes it impossible to automatically generate `strings.tbl`. Not sure how to fix that though since I don't think that the text comes from a table.

wookieejedi: Just changed the name. I see what you mean about not being able to automatically generate strings.tbl. Yeah I couldn't find many entries in `strings.tbl` that were actually listed in the code base.

wookieejedi: If this would be a critical issue I could make a new PR which just assigns the controls that don't have a unique strings.tbl index to have a unique index at 1570 or above?

asarium: No, this is not critical. We need to fix this at some point anyway but that out of scope for this.

wookieejedi: Gotcha. At the very least this PR allows for all the controls to be translatable via `strings.tbl` now.

wookieejedi:
Thanks for all you help!

Issue #1775, opened 2018-06-30
FRED fails to launch FSO if the exe files aren't in the working directory


ngld:
Usually (i.e. before Knossos) both FSO and FRED would be in the FS2 root directory which would also be the working directory. However, due to the way Knossos works, this is no longer the case. The root directory still is the current working directory but the exe files are stored elsewhere.

The relevant function for FRED2 is [`CFREDView::OnRunFreeSpace`](https://github.com/scp-fs2open/fs2open.github.com/blob/e8ef727a199f00b44b0fe186e498bf5bbb823897/fred2/fredview.cpp#L3768) which apparently hasn't been ported over to qtFRED, yet.

The function is called when a user clicks on File > Run FreeSpace in FRED. It takes the name of the current executable and replaces `fred2_open` with `fs2_open` and tries to launch the exe with the resulting name. To make this work with Knossos, FRED would have to identify the folder containing its exe file is and use that to build the correct path to FSO's exe.

wookieejedi:
This can be closed now that #2192 is merged.

PR #2192, opened 2019-12-13
Open FSO from FRED Using FRED's Directory

FSCyborg:
According to the CreateProcess() documentation, passing NULL in the eighth argument starts a process in the same directory as the original process, which is what we're trying to do.

Please note this is a Windows only solution, and cannot be ported to QTFred.

Addresses #1775, already tested on Release and Debug.

### DIFF
--- a/fred2/fredview.cpp
+++ b/fred2/fredview.cpp
@@ -3777,12 +3777,12 @@ void CFREDView::OnRunFreeSpace()
 		return;
 
 	si.cb = sizeof(si);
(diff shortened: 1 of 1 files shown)


wookieejedi:
FYI use nullptr :)

FSCyborg:
It's true. I'll make the changes. If I'm going to make those changes, I might as well do it for all the parameters...

FSCyborg:
There you go!

PR #2194, opened 2019-12-14
Allows the custom controls to work with `key-pressed` sexp

wookieejedi:
This expands the functionality of the cusotm controls to work with rebindings when using the sexp `key-pressed`. A key default value must be specified in `controlsconfigcommon.cpp` for the sexp `key-pressed` to track key binding changes. This PR adds that functionality by assigning default keys that were not in use (alt shift 1-5).


Goober5000:
May I suggest, in addition to your custom controls, adding the digits 5-9 and 0.  This will allow key-pressed to work with the entire set of single-digit numbers (currently only 1-4 are supported).

wookieejedi:
> May I suggest, in addition to your custom controls, adding the digits 5-9 and 0. This will allow key-pressed to work with the entire set of single-digit numbers (currently only 1-4 are supported).

Good idea. For 0 and 5-9 they should probably have a string name, so should they also be listed as custom controls 0  and 5-9 or something else in the Control_config array?

Goober5000:
Uh, probably just "Key 5" etc.  Keep them distinct from the other custom controls you added.

This does have the potential flaw that any of the numbers could be re-bound, but missions have been using keys 1-4 for many years and I can recall only one person who encountered an issue.

wookieejedi:
Note that PR#2197 will fix the issue when using of mismatching sexp tree for `key-pressed` when using a custom `Controlconfigdefaults.tbl`

MageKing17:
By increasing the number of keys above 127, this PR technically breaks multi compatibility by changing the value of `NUM_BUTTON_FIELDS`.

wookieejedi:
Ah interesting. Technically the custom controls and  keys 5-9 do the same thing, so should I just delete the custom controls and submit a PR?

PR #2196, opened 2019-12-14
Mantis#2129 Address Rare Multi Crash

JohnAFernandez:
Should fix rare multi crash where reserving net signatures for spawned munitions would cause an ushort overflow, triggering an assert in multi_set_network_signature.

Link to mantis issue: http://scp.indiegames.us/mantis/view.php?id=2129

@notimaginative if you could take a quick look to make sure I'm understanding the situation correctly.

In draft because this is not yet tested, because I do not have a multi testing solution yet.

### DIFF
--- a/code/network/multiutil.cpp
+++ b/code/network/multiutil.cpp
@@ -197,7 +197,9 @@ ushort multi_get_next_network_signature( int what_kind )
 // and is used mainly for firing weapons.  what_kind tells us permanent or non-permanent signature
 void multi_set_network_signature( ushort signature, int what_kind )
 {
-	Assert( signature != 0 );
+	Assertion(signature != 0, "Invalid net signature of 0 requested in multi_set_network_signature().");
+	Assertion(what_kind > 0, "Invalid net signature type of value %d requested in multi_set_network_signature().", what_kind);  // get Allender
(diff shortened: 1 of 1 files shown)


notimaginative:
I think it would be better to fix this in `multi_set_network_signature()` since that will work everywhere and no one should have to deal with it in the future. You should be able to get away with just removing the `Assert()` and doing `NPERM_SIG_MIN + signature`, taking advantage of the wrap-around. Just be sure it stays within the bounds of `NPERM_SIG_MIN` and `NPERM_SIG_MAX`.

FSCyborg:
All right, let's try that instead.

notimaginative:
Looks good to me. May want to add an `Assert()` checking that `NPERM_SIG_MIN + signature` is less than `NPERM_SIG_MAX` (might have to static_cast<uint> something, not sure). It shouldn't ever happen, but it technically can, so best to check for it so we'll know if the near-impossible happens in the future.

JohnAFernandez:
Just finished testing.  Weapons and objects, appear to work as well as before the changes.

wookieejedi:
Looks like everything is good to go then?

JohnAFernandez:
Everything is ready

Issue #1066, opened 2016-11-07
Backward Compatibility and Deprecation of VM_EXTERNAL_CAMERA_LOCKED in scripts


z64555:
PR #1065 Addresses issue #965, however it does not deprecate VM_EXTERNAL_CAMERA_LOCKED that may be present in older scripts.

The new flag, VM_CAMERA_LOCKED, takes the same bit place of VM_EXTERNAL_CAMERA_LOCKED, but is not quite the same behavior.

The old behavior of VM_EXTERNAL_CAMERA_LOCKED is the same as the new behavior of VM_CAMERA_LOCKED when VM_EXTERNAL is also true.

MageKing17:
Is it not possible to have both `VM_EXTERNAL_CAMERA_LOCKED` and `VM_CAMERA_LOCKED`, with the former becoming basically an alias for `VM_CAMERA_LOCKED && VM_EXTERNAL`?

z64555:
That's what I was thinking for getting things backwards compatible, but the `VM_EXTERNAL_CAMERA_LOCKED` flag should still be marked as deprecated when running a debug build. It's not as flexible as having the camera lock state separate from the external view state.

PR #2198, opened 2019-12-15
Add deprecation warning for VM_EXTERNAL_CAMERA_LOCKED

FSCyborg:
Previously deprecated enum for lua function hasViewmode() didn't have a warning.  This fixes that. (And only does it once per instance, test confirmed)

Also, I am adding a check to omit it while writing scripting.html so that modders don't accidentally pick start using it.

It still has the correct behavior from the last time it was edited, tested via a script written by @wookieejedi which checks each flag per frame.  The camera modes are enabled via keyboard controls.  Pressing number pad . turns on the external camera, and pressing number pad enter locks that camera.  The script showed that the check for VM_EXTERNAL_CAMERA_LOCKED returned true only when both VM_EXTERNAL and VM_CAMERA_LOCKED are true.

Only step left for us to is change wiki documentation and do a quick check for documentation elsewhere.

For  #1066 .

### DIFF
--- a/code/scripting/lua.cpp
+++ b/code/scripting/lua.cpp
@@ -363,11 +363,17 @@ void script_state::OutputLuaMeta(FILE *fp)
 	//***Enumerations
 	fprintf(fp, "<dt id=\"Enumerations\"><h2>Enumerations</h2></dt>");
 	for (uint32_t i = 0; i < Num_enumerations; i++) {
-		//WMC - This is in case we ever want to add descriptions to enums.
-		//fprintf(fp, "<dd><dl><dt><b>%s</b></dt><dd>%s</dd></dl></dd>", Enumerations[i].name, Enumerations[i].desc);
 
-		//WMC - Otherwise, just use this.
-		fprintf(fp, "<dd><b>%s</b></dd>", Enumerations[i].name);
+		// Cyborg17 -- Omit the deprecated flag
+		if (!stricmp(Enumerations[i].name,
+		             "VM_EXTERNAL_CAMERA_LOCKED") /* || !stricmp(Enumerations[i], "ADD DEPRECATED ENUM HERE"*/) {
+			continue;
+		}		
+		// WMC - This is in case we ever want to add descriptions to enums.
+		// fprintf(fp, "<dd><dl><dt><b>%s</b></dt><dd>%s</dd></dl></dd>", Enumerations[i].name, Enumerations[i].desc);
+  
+		// WMC - For now, print to the file without the description.
+		fprintf(fp, "<dd><b>%s</b></dd>", Enumerations[i].name);		
 	}
 	fputs("</dl>\n", fp);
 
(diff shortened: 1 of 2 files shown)


FSCyborg:
And let me fix that....

code/scripting/lua.cpp
```diff
@@ -363,11 +363,13 @@ void script_state::OutputLuaMeta(FILE *fp)
...
 	fprintf(fp, "<dt id=\"Enumerations\"><h2>Enumerations</h2></dt>");
 	for (uint32_t i = 0; i < Num_enumerations; i++) {
-		//WMC - This is in case we ever want to add descriptions to enums.
-		//fprintf(fp, "<dd><dl><dt><b>%s</b></dt><dd>%s</dd></dl></dd>", Enumerations[i].name, Enumerations[i].desc);
-
-		//WMC - Otherwise, just use this.
+		// WMC - This is in case we ever want to add descriptions to enums.
+		// fprintf(fp, "<dd><dl><dt><b>%s</b></dt><dd>%s</dd></dl></dd>", Enumerations[i].name, Enumerations[i].desc);
+  
+		if (stricmp(Enumerations[i].name, "VM_EXTERNAL_CAMERA_LOCKED")) { 		// Cyborg17 -- Omit the deprecated flag
```
z64555: Since it may be possible for more flags to be deprecated, I feel this bit here should be something more future proof, such as 
`if ((Enumerations[i].name == "VM_EXTERNAL_CAMERA_LOCKED"))
  // || (Enumerations[i].name == "Some_Deprecated_Flag))
{
  continue;
}`

FSCyborg: The equality operator doesn't play nice with the linux builds on github.  But using the continue statement is a good idea, so I'll separate it out.

z64555: Note that the example is psuedocode, you should use whatever style of string comparison the rest of the file uses.

FSCyborg:
By looking at the failing build info, I think there was an error in downloading packages, so I am going to try to force it to try again.

code/scripting/lua.cpp
```diff
@@ -363,11 +363,15 @@ void script_state::OutputLuaMeta(FILE *fp)
 	//***Enumerations
 	fprintf(fp, "<dt id=\"Enumerations\"><h2>Enumerations</h2></dt>");
 	for (uint32_t i = 0; i < Num_enumerations; i++) {
-		//WMC - This is in case we ever want to add descriptions to enums.
-		//fprintf(fp, "<dd><dl><dt><b>%s</b></dt><dd>%s</dd></dl></dd>", Enumerations[i].name, Enumerations[i].desc);
-
-		//WMC - Otherwise, just use this.
-		fprintf(fp, "<dd><b>%s</b></dd>", Enumerations[i].name);
+		// WMC - This is in case we ever want to add descriptions to enums.
```
Goober5000: WMC's descriptions comment should appear immediately prior to his other comment, because that is what "otherwise" refers to.

Be careful when you make code changes that you don't introduce errors in comments.  They can't be caught in the same way as errors in code.

FSCyborg: Can we change his second comment to 		
`// WMC - For now, print to the file without the description.`

Goober5000: The two comments are two sides of the same coin.  They are conceptually related and should be linked.  They *were* linked until you put the code in the middle and separated them.

Just move your code to the top of the `for` loop.

PR #2200, opened 2019-12-17
POF 2118, BSP Alignment and SLC2 support

Shivansps:
Converted the SLDC shield collision system to SLC2 (char for int for alignment), added conversion of SLDC to SLC2 and BSP_Data Defpoints alignment on model load.

### DIFF
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -1271,4 +1271,8 @@ void glowpoint_init();
 SCP_vector<glow_point_bank_override>::iterator get_glowpoint_bank_override_by_name(const char* name);
 extern SCP_vector<glow_point_bank_override> glowpoint_bank_overrides;
 
+// pof version < 2118 conversion - ShivanSpS
+uint align_bsp_data(ubyte* bsp_in, ubyte* bsp_out, uint bsp_size);
+uint convert_sldc_to_slc2(ubyte* sldc, ubyte* slc2, uint tree_size);
+
 #endif // _MODEL_H
--- a/code/model/modelcollide.cpp
+++ b/code/model/modelcollide.cpp
@@ -982,17 +982,17 @@ bool mc_check_sldc(int offset)
 	// not used
 	//int *size_p = (int *)(Mc_pm->shield_collision_tree+offset+4);
 	// split and polygons
-	vec3d *minbox_p = (vec3d*)(Mc_pm->shield_collision_tree+offset+8);
-	vec3d *maxbox_p = (vec3d*)(Mc_pm->shield_collision_tree+offset+20);
+	auto *minbox_p = (vec3d*)(Mc_pm->shield_collision_tree+offset+8);
+	auto *maxbox_p = (vec3d*)(Mc_pm->shield_collision_tree+offset+20);
 
 	// split
-	unsigned int *front_offset_p = (unsigned int*)(Mc_pm->shield_collision_tree+offset+32);
-	unsigned int *back_offset_p = (unsigned int*)(Mc_pm->shield_collision_tree+offset+36);
+	auto *front_offset_p = (unsigned int*)(Mc_pm->shield_collision_tree+offset+32);
+	auto *back_offset_p = (unsigned int*)(Mc_pm->shield_collision_tree+offset+36);
 
 	// polygons
-	unsigned int *num_polygons_p = (unsigned int*)(Mc_pm->shield_collision_tree+offset+32);
+	auto *num_polygons_p = (unsigned int*)(Mc_pm->shield_collision_tree+offset+32);
 
-	unsigned int *shld_polys = (unsigned int*)(Mc_pm->shield_collision_tree+offset+36);
+	auto *shld_polys = (unsigned int*)(Mc_pm->shield_collision_tree+offset+36);
 
 
 
--- a/code/model/modelcollide.cpp
+++ b/code/model/modelcollide.cpp
@@ -973,24 +973,26 @@ bool mc_shield_check_common(shield_tri	*tri)
 
 bool mc_check_sldc(int offset)
 {
+	//ShivanSpS - Changed the type char for a type int (Now SLC2)
 	if (offset > Mc_pm->sldc_size-5) //no way is this big enough
 		return false;
-	char *type_p = (char *)(Mc_pm->shield_collision_tree+offset);
+
+	int *type_p = (int *)(Mc_pm->shield_collision_tree+offset);
 	
 	// not used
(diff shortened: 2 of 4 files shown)


notimaginative:
align_bsp_data() isn't called from anywhere. Is part of the changes missing from this?

Shivansps:
Im missing one commit... Strange.

Obtener Outlook para Android<https://aka.ms/ghei36>

________________________________

Shivansps:
Ready, not sure what happened, either i forgot to push it or it never went trought, github has been working very bad for me today.

Shivansps:
im going to do one more commit later today to change remaining "unsigned int" and "int" for "uint". So everything is up to the rest of FSO code.

Goober5000:
You've got a few code style issues that Travis is complaining about.

Also, to get rid of those "unknown type" warnings, you can add `struct matrix`, `class ubyte`, and `struct vec3d` at the top of the file near `class polymodel`.

Shivansps:
Strange that is even possible, i trought that was a error with travis since i havent added any new types.

Goober5000:
You added new function definitions in modelsinc.h.  That's where Travis is complaining about matrix, ubyte, and vec3d.

Shivansps:
I know, but ubyte was already used by other functions that were already declared there, im not adding a type that were not previusly there, and im not using vec3d or matrix in the new declarations. That comes of others functions that were already there. So travis was already failling for modelsinc.h before these changes.

And travis is right, modelsinc.h is not incluiding the header were ubyte, vec3d and matrix are declared, this is easily fixed by re-declare them at modelsinc.h with a #ifndef i could add that if you wish.

Goober5000:
You're right that ubyte and vec3d were already in existing modelsinc.h declarations.  The compiler should have complained about those, too.  That's very strange.

I don't know why those other declarations worked in the past, but they should not be expected to work because those types are neither defined nor included in that file.  So you should add the struct/class forward declarations at the top of the file, in the same fashion as polymodel.

Shivansps:
mmm Goober it cant be fixed, try yourself add this to modelsinc.h and try to compile.

`//Copied from pstypes.h
#ifndef vec3d
typedef struct vec3d {
	union {
		struct {
			float x, y, z;
		} xyz;
		float a1d[3];
	};
};
#endif

//Copied from pstypes.h
#ifndef ubyte
typedef std::uint8_t ubyte;
#endif

//Copied from pstypes.h
#ifndef matrix
typedef struct matrix {
	union {
		struct {
			vec3d rvec, uvec, fvec;
		} vec;
		float a2d[3][3];
		float a1d[9];
	};
};
#endif`

its going to start complaining about not being declared all over the place in the code. Travis is also using the header wrong as this warning is also present
Running clang-tidy on changed files
/home/travis/build/scp-fs2open/fs2open.github.com/code/model/modelsinc.h:17:2: error: This should only be used internally by the model library.  See John if you think you need to include this elsewhere. [clang-diagnostic-error]

Is clear that warning should not happen, so travis is using this header in the wrong way.

Shivansps:
The only possible solution i can think off is moving the problematic declarations out of modelsinc.h into the .cpp

to modelinterp.cpp
void model_calc_bound_box(vec3d* box, vec3d* big_mn, vec3d* big_mx);

to modeloctant.cpp
extern vec3d** Interp_verts;

to modelread.cpp
extern int model_interp(matrix* orient, ubyte* data, polymodel* pm);

// endian swapping stuff - kaz
void swap_sldc_data(ubyte* buffer);

// pof version < 2118 conversion - ShivanSpS
uint align_bsp_data(ubyte* bsp_in, ubyte* bsp_out, uint bsp_size);
uint convert_sldc_to_slc2(ubyte* sldc, ubyte* slc2, uint tree_size);

Shivansps:
I got an idea let me see if this works, it would not do any real change to the code but it may make travis accept it.

Goober5000:
Did you, in fact, try to add the struct/class forward declarations at the top of the file like I recommended?

You should not be redefining basic types in multiple places in the code.  And, in fact, the `align_bsp_data` and `convert_sldc_to_slc2` functions should not be in modelsinc.h at all.  Move them to model.h.

Shivansps:
My bad i missunderstood. Ok done, i still think is going to fail due to this.

#ifndef MODEL_LIB 
#error This should only be used internally by the model library.  See John if you think you need to include this elsewhere.
#endif

Well see, in the cases that the build fail on travis is is show, it should not.

Shivansps:
I had to comment that forced error on not model_lib defined, there is no other way. It should pass now.
But keep in mind that i chaged nothing that could trigger any of this, it was like this when i found it. haha.

Shivansps:
Well... that was just bad luck there.

Goober5000:
Uncomment the `#error` line in modelsinc.  If it interferes with compilation, there must be some deeper reason.

Shivansps:
Yes, travis is incluiding modelsinc.h in those two specific cases somewhere where MODEL_LIB is not defined, but since i cant reproduce in MSVC or on Linux GCC figuring that out it will be trial and error.
I can run travis directly on my code folder whiout doing commits? i never used it.

The only lead that i have is that modelsinc.h has been used to place functions declarations for stuff added to FSO, thats the reason why i added my function declarations there, swap_bsp_data added by kaz is there for example. Maybe modelsinc.h was included somewhere it should not to in the past.

Shivansps:
After taking a closer look all modelsinc.h includes are from cpps that has the "#define MODEL_LIB" set, so the code OK, i think this is just a issue with travis. I found this here: https://github.com/ainfosec/ci_helloworld

> This tells CMake to record all of the compilation instructions used to compile your project including flags and definitions. Clang Tidy will use this information to statically analyze your project the same way it was compiled. The advantage to this approach is a significant improvement in accuracy. The main disadvantage to this approach is Clang Tidy is not good at statically analyzing files that do not show up in this compilation database (such as header files). For this reason, if you want to analyze a header, it has to be included by a source file that is included in the compilation database.

modelsinc.h should not be analysed alone, it should be analysed only when included to other cpp, if you run it alone like travis clang tidy is doing you will always run into the missing model_lib definition. This is also why it complained about the missing types as well, as the files that are incluiding it has a reference to pstypes.h somewhere.

You are probably run into the same issue if you modify for example cfilearchive.h.

So tell me what you want me to do, i can uncomment the #error but is going to fail or i can leave it as it is.

Goober5000:
Un-comment the error.  We will review the rest of the PR, and we can still approve/merge it even if Travis fails.

PR #2201, opened 2019-12-17
Make Dinky Shockwaves not dinky by default.

### LINKED ISSUE #774: Dinky shockwaves aren't retail-compatible
Okay so I was looking into something completely different and I discovered that in 2005 (commit 78fab0c331836f883bc56d2d6b46f8e34f8cd0d5), wmcoolmon took a hardcoded division by 4 and turned it into a customizeable "dinky shockwave"/weapon arming system.

Or, at least, that's what he almost certainly _thought_ he was doing.

As it happens, I cannot find any way for the code responsible for the aforementioned division to have ever actually executed; even going back to the original Volition sources, `weapon_hit()` cannot, as far as I can tell, be called with a weapon for the other object. As a result, `SW_WEAPON_KILL` can never be set and the original division-by-4 that was the inspiration for the default stats of dinky shockwaves... never actually happened.

The new logic used by dinky shockwaves, however, doesn't check whether `weapon_hit()` got called with another weapon; it just checks `WF_DESTROYED_BY_WEAPON`. So now all weapons with shockwaves are, by default, exploding for 1/4th normal damage when shot down, and have been for the past 11 years.

To make matters worse, there was _another_ piece of unused code that was meant to multiply the damage back up by 4 when that shockwave hit a ship; presumably, it was expected to do the same thing under the new dinky shockwave system, except since it didn't do anything before, it still didn't do anything after. So if we fixed either end of this thing (either stop making weapons spawn dinky shockwaves without explicitly allowing for their presence, or make it so that dinky shockwaves still do full damage to ships), it would change the balance of any mission involving shooting down bombs balanced over the past 11 years, as suddenly shot-down bombs would deal 4 times as much damage as expected. Unfortunately, leaving it as-is means we're breaking retail balance.

I am honestly at a loss as to what should be done about this.

MageKing17:
Adds a new mod table option "$Default Dinky Shockwaves:" that toggles the behavior. See issue #774 for more information.

Needless to say, fixes #774.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -48,7 +48,7 @@ SCP_string Window_icon_path;
 bool Disable_built_in_translations;
 bool Weapon_shockwaves_respect_huge;
 bool Using_in_game_options;
-bool Dinky_shockwaves_by_default;
+float Dinky_shockwave_default_multiplier;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_damage_p1;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_damage_p2;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_damage_s1;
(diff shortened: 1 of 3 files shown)


MageKing17:
An excellent suggestion.

PR #2203, opened 2019-12-18
avoid ending mission on death unless we should

notimaginative:
When dead, particularly in multi, the mission is closed out even if we aren't done with it. Various menus are still accessible but once you exit the menu it returns to processing the game frame as normal, except everything got closed out which triggers a multitude of asserts and other problems. This should fix Mantis 2377 and anything related.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -5464,36 +5464,39 @@ void game_leave_state( int old_state, int new_state )
 			break;
 
 		case GS_STATE_DEATH_DIED:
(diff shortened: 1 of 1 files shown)


notimaginative:
For the record, retail appears to have this same issue. I'm not sure why it doesn't go crazy there too. However every issue that I noticed stemming from the level close out is with FSO added code.

Goober5000:
Link to Mantis:
http://scp.indiegames.us/mantis/view.php?id=2377

Issue #2199, opened 2019-12-16
void multi_ts_get_team_and_slot() Depends on Behavior that is not guaranteed


JohnAFernandez:
This function depends on the mission creator not using special characters in wing names.  If any mission creator were to do this in a multi mission, a myriad of things would probably break including respawning, and they would have no clue why, and we would probably spend endless time debugging.

`*slot_index = (ship_name[strlen(ship_name)-1] - '1');`  Get's the last character of the ship's name and then subtracts one, oblivious to the fact that we can add other characters to the end of the string using "#".  (Yes, I know this would be a very rare occurrence, but I'd rather change this now instead of trying to figure it out when I get an extremely odd assert)

Not sure if this is related to the bugs that I was looking at when I started looking at this function, but basically we just need a more robust way to get the ship's `wss` slot.

Goober5000:
Is this function ever called with the display name?  From a cursory search, all occurrences are called with the ship's actual name.  The actual name is the full string, including '#' and everything.

MageKing17:
To the best of my knowledge, if a wing has "#" in the name, the ships still get the number added to the end (after the hash), so while hackish, this shouldn't actually be a problem.

FSCyborg:
I looked in FRED before posting and this is the behavior I saw there. Hopefully it is corrected on mission save or at some other point, and I had nothing to worry about.
![bug2](https://user-images.githubusercontent.com/12529946/71055595-40ec7900-2124-11ea-8658-97904f2b8564.png)

In any case, I am currently looking for more robust ways of checking instead of relying on the hack, since there is another bug in the function, and I want to try to guarantee that it works every time, regardless.

This function, I believe, is also the cause (for another reason) of Mantis#2757:
http://scp.indiegames.us/mantis/view.php?id=2757

I lucked out and found a mission where that bug appears every time: All Alone.  For some reason in that mission the ship names are not Alpha 1, Alpha 2 in-mission, and I have already tested the first version of a fix as a host.  We'll see about the rest.  But that's all for now.

FSCyborg:
Every instance I saw was the actual ship name.

MageKing17:
Upon reflection, I think the "ship number goes after the hash" was considered a bug and fixed, with the behavior only being retained if the wing name ended with a hash, which e.g. The Blade Itself uses. So this definitely does need fixing.

FSCyborg:
[...]

PR #2204, opened 2019-12-19
Make Respawning independent from Ship Names

JohnAFernandez:
Changes the way `multi_ts_get_team_and_slot()` searches for the correct slot in order to keep return values from being -1 whenever the current method would run into trouble, like when "#" is used in a wing name, or when `Ship[]` is cleared after a ship's death (thank you, Taylor!!!).  This iteration of the function pulls the wing name directly from a pointer to the `Ships[]` array and also gets more info from a pointer to the 'Wings[]' array.

Tested as a host in a coop match, but needs three more quick things for testing,
1. Need to check to make sure it works for clients as well as hosts (more than likely already does, but let's be sure)
2. Need to check that this works in TVT
3. Need to check that correct loadout is being given to the respawned ship

### DIFF
--- a/code/network/multiteamselect.cpp
+++ b/code/network/multiteamselect.cpp
@@ -36,6 +36,7 @@
 #include "object/object.h"
 #include "parse/parselo.h"
 #include "mission/missionparse.h"
+#include "ship/ship.h"
 
 
 // ------------------------------------------------------------------------------------------------------
@@ -419,7 +420,7 @@ void multi_ts_init_objnums();
 void multi_ts_init_flags();
 
 // get the proper team and slot index for the given ship name
-void multi_ts_get_team_and_slot(char *ship_name,int *team_index,int *slot_index, bool mantis2757switch);
+void multi_ts_get_team_and_slot(ship *shipp,int *team_index,int *slot_index);
 
 // handle an available ship scroll down button press
 void multi_ts_avail_scroll_down();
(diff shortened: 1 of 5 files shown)


FSCyborg:
Should fix http://scp.indiegames.us/mantis/view.php?id=2757 , as well as deal with very minor issue #2199 .

FSCyborg:
Current status: need to figure out what's wrong on these windows builds (my setup can easily compile).    

I took a deep dive to make sure we could trust `shipp->wing_status_wing_pos`. The whole story:
`multi_respawn_common_stuff()` calls `parse_create_object()` which calls `parse_create_object_sub()` which calls `ship_create()`.
`ship_create()` resets all ship values, but *after* it returns, `parse_create_object_sub()`
 restores the value from the object parse struct (created during mission parse).

All other cases are clearer cut.  Testing will further verify this.

CasparKielwein:
> Current status: need to figure out what's wrong on these windows builds (my setup on windows can easily compile).

I hit the same problem in lab.cpp in #2205 and changed the empty calls to clear.
its in 8b09c91

CasparKielwein:
If you rebase your branch onto master, the build problem should be fixed now.

code/network/multiteamselect.cpp
```diff
@@ -1693,49 +1696,63 @@ bool multi_ts_validate_ship(char *shipname, char *wingname)
...
+			if ( !strcmp(wingp->name, TVT_wing_names[idx] )) {				
 				*team_index = idx;
-				*slot_index = (ship_name[strlen(ship_name)-1] - '1');
-
+				*slot_index = (shipp->wing_status_wing_pos);
+				
 				// just Assert(), if this is wrong then we're pretty much screwed either way
-				Assert( (*slot_index >= 0) && (*slot_index < MAX_WSS_SLOTS) );
+				Assertion(*slot_index >= 0, "Within multi_ts_get_team_and_slot() slot index is negative, despite finding a wing. There "
+					"is likely some code error, please report.There is likely some code error, please report.");
```
MageKing17: Duplicated sentence.

JohnAFernandez:
This has a better fix somewhere else now.

PR #2205, opened 2019-12-20
Refactore: cleanup freespace.cpp

CasparKielwein:
The main file, freesapce.cpp is really hard to undestand, so I tried to clean it a bit.

Removes a lot of dead code in the main file freespace2/freespace.h and .cpp.
Touches some other files, where the code was closely linked to the main file.
This should not change any functionality. 

As before with these fairly large cleanups. I will remove anything where a reviewer has objections, to get the rest moved along and maybe create a new PR with the controversial changes for review.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -4112,10 +4112,6 @@ void game_frame(bool paused)
 fix Last_time = 0;						// The absolute time of game at end of last frame (beginning of this frame)
 fix Last_delta_time = 0;				// While game is paused, this keeps track of how much elapsed in the frame before paused.
 static int timer_paused=0;
-#if defined(TIMER_TEST) && !defined(NDEBUG)
-static int stop_count,start_count;
-static int time_stopped,time_started;
-#endif
 int saved_timestamp_ticker = -1;
 
 void game_reset_time()
@@ -4139,23 +4135,13 @@ void game_stop_time()
 		Last_delta_time = time - Last_time;		
 
 		if (Last_delta_time < 0) {
-			#if defined(TIMER_TEST) && !defined(NDEBUG)
-			Int3();		//get Matt!!!!
-			#endif
 			Last_delta_time = 0;
 		}
-		#if defined(TIMER_TEST) && !defined(NDEBUG)
-		time_stopped = time;
-		#endif
 
 		// Stop the timer_tick stuff...
 		saved_timestamp_ticker = timestamp();
 	}
 	timer_paused++;
-
-	#if defined(TIMER_TEST) && !defined(NDEBUG)
-	stop_count++;
-	#endif
 }
 
 void game_start_time()
@@ -4165,19 +4151,11 @@ void game_start_time()
 	if (timer_paused==0) {
 		fix time;
 		time = timer_get_fixed_seconds();
-		#if defined(TIMER_TEST) && !defined(NDEBUG)
-		if (Last_time < 0)
-			Int3();		//get Matt!!!!
-		}
-		#endif
-		// Take current time, and set it backwards to account for time	
+		// Take current time, and set it backwards to account for time
 		// that the frame already executed, so that timer_get_fixed_seconds() - Last_time
 		// will be correct when it goes to calculate the frametime next
 		// frame.
 		Last_time = time - Last_delta_time;		
-		#if defined(TIMER_TEST) && !defined(NDEBUG)
-		time_started = time;
-		#endif
 
 		// Restore the timer_tick stuff...
 		// Normally, you should never access 'timestamp_ticker', consider this a low-level routine
@@ -4185,10 +4163,6 @@ void game_start_time()
 		timestamp_set_value(saved_timestamp_ticker);
 		saved_timestamp_ticker = -1;
 	}
-
-	#if defined(TIMER_TEST) && !defined(NDEBUG)
-	start_count++;
-	#endif
 }
 
 void lock_time_compression(bool is_locked)
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -2037,7 +2037,6 @@ void game_show_framerate()
 	}
 #endif
 
-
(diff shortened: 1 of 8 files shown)


freespace2/freespace.cpp
```diff
@@ -2771,11 +2652,11 @@ void do_timing_test(float frame_time)
...
 		Test_begin = 0;
 
-		for ( i = 0; i < NUM_MIXED_SOUNDS; i++ )
-			snds[i] = sound_handle::invalid();
+		for (auto & snd : snds)
+			snd = sound_handle::invalid();
 
 		// start looping digital sounds
-		for ( i = 0; i < NUM_MIXED_SOUNDS; i++ )
+		for (int i = 0; i < NUM_MIXED_SOUNDS; i++ )
```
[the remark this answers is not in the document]
CasparKielwein: the last loop uses the index to access gamesnd_id(i) so it can't be a range based loop.

It should be possible to simply remove the ```snd= sound_handle::invalid();``` loop as the last loop overwrites all values anyways.

freespace2/freespace.cpp
```diff
@@ -2752,16 +2634,15 @@ void do_timing_test(float frame_time)
...
 
 	if ( test_running ) {
 		framecount++;
 		test_time += frame_time;
 		if ( framecount >= NUM_FRAMES_TEST ) {
 			test_running = 0;
 			nprintf(("General", "%d frames took %.3f seconds\n", NUM_FRAMES_TEST, test_time));
-			for ( i = 0; i < NUM_MIXED_SOUNDS; i++ )
-				snd_stop(snds[i]);
+			for (auto snd : snds)
```
MageKing17: Why is this not using `auto &`?

CasparKielwein: sound_handle is a strong type wrapper around an int, so pass by value would generally be preferable, in the rare cases, where it makes a difference.
But I see that in this context it is irrelevant in case of generated code, and not apparent to a reader.

I'll make it auto& if you prefer that.

(I personally like to use value instead of reference where possible, but will happily follow conventions on the project.)

freespace2/freespace.cpp
```diff
@@ -2722,23 +2607,20 @@ extern void render_shields();
 void player_repair_frame(float frametime)
 {
 	if(MULTIPLAYER_MASTER){
-		int idx;
-		for(idx=0;idx<MAX_PLAYERS;idx++){
-			net_player *np;
+		for(auto & idx : Net_players){
+			net_player *np = &idx;
```
Goober5000: Why not use `np` as the loop variable directly?

freespace2/freespace.cpp
```diff
@@ -7329,26 +6942,19 @@ DCF_BOOL( time_models, Time_models )
...
 {
-	
-
 	if ( !Time_models ) return;
 
 	mprintf(( "Timing models!\n" ));
 
-	int i;
+	ubyte model_used[MAX_POLYGON_MODELS] = {0};
+	int model_id[MAX_POLYGON_MODELS] = {0};
```
Goober5000: `model_id` doesn't need to be initialized here, but if it is, it should be -1, not 0

freespace2/freespace.cpp
```diff
@@ -7376,28 +6981,26 @@ void Do_model_timings_test()
 }
 #endif
 
-// Call this function when you want to inform the player that a feature is disabled in this build
+/// Call this function when you want to inform the player that a feature is disabled in this build
```
Goober5000: Why the extra slash?

freespace2/freespace.cpp
```diff
@@ -7376,28 +6981,26 @@ void Do_model_timings_test()
...
 void game_feature_disabled_popup()
 {
 	popup(PF_USE_AFFIRMATIVE_ICON|PF_BODY_BIG, 1, POPUP_OK, XSTR( "Sorry, the requested feature is currently disabled in this build", 1621));
 }
 
-// format the specified time (fixed point) into a nice string
+/// format the specified time (fixed point) into a nice string
```
Goober5000: Ditto

freespace2/freespace.cpp
MageKing17: I don't have a preference; I was asking for clarification.

freespace2/freespace.cpp
```diff
@@ -2722,23 +2607,20 @@ extern void render_shields();
...
+		for(auto & idx : Net_players){
+			net_player *np = &idx;
 
-			np = &Net_players[idx];
-
-			if(MULTI_CONNECTED(Net_players[idx]) && (Net_player != NULL) && (Net_player->player_id != Net_players[idx].player_id) && (Net_players[idx].m_player != NULL) && (Net_players[idx].m_player->objnum >= 0) && (Net_players[idx].m_player->objnum < MAX_OBJECTS)){
+			if(MULTI_CONNECTED(idx) && (Net_player != nullptr) && (Net_player->player_id != idx.player_id) && (idx.m_player != nullptr) && (idx.m_player->objnum >= 0) && (idx.m_player->objnum < MAX_OBJECTS)){
 
 				// don't rearm/repair if the player is dead or dying/departing
 				if ( !NETPLAYER_IS_DEAD(np) && !(Ships[Objects[np->m_player->objnum].instance].is_dying_or_departing()) ) {
```
MageKing17:
```suggestion
				if ( !NETPLAYER_IS_DEAD(&idx) && !(Ships[Objects[idx.m_player->objnum].instance].is_dying_or_departing()) ) {
```
And then either get rid of `np` or rename `idx` to `np`.

freespace2/freespace.cpp
CasparKielwein: three / is the doxygen short form for 
`/**
* @brief blablabla
*/`

freespace2/freespace.cpp
MageKing17: You still have the now-unused `np` variable.

CasparKielwein: yes, I'll fix it together with goobers remarks, when the kid is in bed.

Issue #1441, opened 2017-08-22
Merge the multilock code and required changes related to other PRs


TRBlount:
If I remember correctly, the only objection @SamuelCho had to merging multilock was related to netcode. The suggested compromise is to display a warning or error when attempting to use the code with multiplayer and include a note in the wiki. I've been keeping a copy of the branch around up to date for Spoon's Wings of Dawn build, and thus far I've only had to make one change related to #647 in regards to seeker heads being ineffective.
If we merged #647 and then multilock, Wings of Dawn would no longer need a custom build and the feature would be available to more modders.

MageKing17:
> Wings of Dawn would no longer need a custom build

...Except for the increased BMPMAN slot limit, you mean?

TRBlount:
Ah, I forgot about that. Yes, there is also that issue, which should probably be addressed at some point. We really shouldn't encourage the use of custom builds at all.

TRBlount:
BMPMan slot limit should no longer be an issue right? In addition, with the new insights from ngld with respect to the netcode, the old issues may become moot as fixes start rolling it.

TRBlount:
In order to reach the Release 19.1 milestone, our current plan is to merge Multilock and then make the multiplayer related fixes for it a milestone for Release 20. The reasoning for this is there are already multiple features that are not supported by multiplayer that are quite important for users in single-player, so adding another one wouldn't affect the current state of the game.

PR #73, opened 2015-05-05
Fix Mantis 3140.

MageKing17:
Changes the "virtual frame" used for aspect locks to be 640x480 so as to correct different lock behavior due to both different aspect ratios and different resolutions. Applies a scaling factor to the drawing of the lock indicator on the HUD so that it traverses roughly the same portion of the screen regardless of resolution.

PR #2207, opened 2019-12-22
Merge Multilock

TRBlount:
The long awaited multilock feature is finally here! The documentation will be added to the wiki ASAP but one final review for this should be done before merging. This will handle the comments from #1441.

### DIFF
--- a/code/hud/hudlock.cpp
+++ b/code/hud/hudlock.cpp
@@ -91,6 +91,7 @@ char Lockspin_fname[NUM_HUD_RETICLE_STYLES][GR_NUM_RESOLUTIONS][MAX_FILENAME_LEN
 };
 
 void hud_lock_determine_lock_point(vec3d *lock_world_pos_out);
+void hud_lock_determine_lock_point(lock_info *current_lock, weapon_info *wip);
 
 // hud_init_missile_lock() is called at the beginning of a mission
 //
@@ -168,7 +169,7 @@ void HudGaugeLock::initialize()
 // lock_point_pos should be the world coordinates of the target being locked. Assuming all the 
 // necessary locking calculations are done for this frame, this function will compute 
 // where the indicator should be relative to the player's viewpoint and will render accordingly.
-void HudGaugeLock::render(float frametime)
+void HudGaugeLock::renderOld(float frametime)
 {
 	int			target_objnum, sx, sy;
 	object		*targetp;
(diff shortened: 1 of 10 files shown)


code/hud/hudlock.cpp
```diff
@@ -267,6 +268,88 @@ void HudGaugeLock::render(float frametime)
...
+		if ( current_lock->locked ) {
+			sx = fl2i(lock_point.screen.xyw.x); 
+			sy = fl2i(lock_point.screen.xyw.y);
+			gr_unsize_screen_pos(&sx, &sy);
+
+			// show the rotating triangles if target is locked
+			renderLockTrianglesNew(sx, sy, current_lock->locked_timestamp);
+		} else {
+			sx = fl2i(lock_point.screen.xyw.x) - (current_lock->current_target_sx - current_lock->indicator_x); 
+			sy = fl2i(lock_point.screen.xyw.y) - (current_lock->current_target_sy - current_lock->indicator_y);
```
MageKing17: This is missing the code changes from #73.

code/hud/hudlock.cpp
```diff
@@ -629,76 +743,987 @@ void hud_do_lock_indicator(float frametime)
...
+	}
+
+	vm_vec_sub(&vec_to_lock_pos,&current_lock->world_pos,&Player_obj->pos);
+	vm_vec_rotate(&lock_local_pos,&vec_to_lock_pos,&Player_obj->orient);
+
+	if ( lock_local_pos.xyz.z > 0.0f ) {
+		// Get the location of our target in the "virtual frame" where the locking computation will be done
+		float w = 1.0f / lock_local_pos.xyz.z;
+		float sx = ((gr_screen.clip_center_x*2.0f) + (lock_local_pos.xyz.x*(gr_screen.clip_center_x*2.0f)*w))*0.5f;
+		float sy = ((gr_screen.clip_center_y*2.0f) - (lock_local_pos.xyz.y*(gr_screen.clip_center_y*2.0f)*w))*0.5f;
```
MageKing17: And here, too.

code/ship/ship.cpp
```diff
@@ -19088,3 +19123,162 @@ bool ship::is_arriving(ship::warpstage stage, bool dock_leader_or_single)
...
+}
+
+void ship_queue_missile_locks(ship *shipp, weapon_info *wip)
+{
+	size_t i;
+
+	shipp->missile_locks_firing.clear();
+
+	// queue up valid missile locks
+	for ( i = 0; i < shipp->missile_locks.size(); ++i ) {
```
CasparKielwein: this loop is essentially a copy_if (https://en.cppreference.com/w/cpp/algorithm/copy)
I suggest using copy_if with a back_inserter for the target container missile_locks

code/ship/ship.cpp
```diff
@@ -19088,3 +19123,162 @@ bool ship::is_arriving(ship::warpstage stage, bool dock_leader_or_single)
...
+			shipp->missile_locks_firing.push_back(shipp->missile_locks[i]);
+		}
+	}
+}
+
+bool ship_lock_present(ship *shipp)
+{
+	size_t i;
+
+	for ( i = 0; i < shipp->missile_locks.size(); ++i ) {
```
CasparKielwein: this loop could be replaced with a call to ```any_of(std::begin(ship->missile_locks), std::end(ship->missile_locks), [](auto it) {return it->locked;});```

(possible syntax errors, since I didn't test it)

code/ship/ship.cpp
```diff
@@ -19088,3 +19123,162 @@ bool ship::is_arriving(ship::warpstage stage, bool dock_leader_or_single)
...
+
+	return false;
+}
+
+bool ship_start_secondary_fire(object* objp)
+{
+	ship *shipp;
+	ship_info *sip;
+	ship_weapon *swp;
+	weapon_info *wip;
```
CasparKielwein: I personally strongly prefer declaring local variables where they are first initialized.
(applies to multiple parts of the code)

The original FS code, did use the old C style pattern, since they hat to support ancient C compilers, we fortunately do not.

code/ship/ship.h
```diff
@@ -1405,6 +1439,9 @@ extern bool in_autoaim_fov(ship *shipp, int bank_to_fire, object *obj);
 extern int ship_stop_fire_primary(object * obj);
 extern int ship_fire_primary(object * objp, int stream_weapons, int force = 0);
 extern int ship_fire_secondary(object * objp, int allow_swarm = 0 );
+extern bool ship_start_secondary_fire(object* objp);
+extern bool ship_stop_secondary_fire(object* objp);
+
```
CasparKielwein: the ```extern``` here and in the original code is actually not necessary

code/weapon/weapon.h
```diff
@@ -38,6 +38,15 @@ extern int Num_weapon_subtypes;
...
 #define	WRT_POF		2
 
+// constants for weapon lock acquire methods
+#define WLOCK_PIXEL		0
+#define WLOCK_TIMER		1
+
+// constants for weapon lock restrictions
+#define LR_CURRENT_TARGET				0		// Only lock current target and subsystem
+#define LR_CURRENT_TARGET_SUBSYS		1		// 
+#define LR_ANY_TARGETS					2
```
CasparKielwein: You could use constexpr here instead of define as all compilers we use, support that.

code/weapon/weapons.cpp
```diff
@@ -7840,3 +7920,40 @@ const char* weapon_info::get_display_string() {
 bool weapon_info::has_alternate_name() {
 	return alt_name[0] != '\0';
 }
+
+// Given a weapon, figure out how many independent locks we can have with it.
+int weapon_get_max_missile_seekers(weapon_info *wip)
+{
+	int max_target_locks;
```
CasparKielwein: you could return the results immediately instead of storing them in the temporary variable first.

code/weapon/weapons.cpp
```diff
@@ -7840,3 +7920,40 @@ const char* weapon_info::get_display_string() {
...
+}
+
+bool weapon_can_lock_on_ship_type(weapon_info *wip, int ship_type)
+{
+	if ( wip->ship_type_restrict.size() == 0 ) {
+		// no restrictions since this list wasn't even populated
+		return true; 
+	}
+
+	for ( size_t i = 0; i < wip->ship_type_restrict.size(); ++i ) {
```
CasparKielwein: this loop cold also be replaced with a find or any_of algorithm

code/ship/ship.h
```diff
@@ -283,6 +284,36 @@ typedef struct cockpit_display_info {
 	int size[2];
 } cockpit_display_info;
 
+// structure to keep track of ship locks
+typedef struct lock_info {
```
JohnAFernandez: Love the struct which organizes everything well.  But the multi issue we were afraid of caused by us switching to this struct: the aspect-seekers fired by clients in multiplayer do not home in on their targets because we no longer set.... `Player_ai->current_target_is_locked` .   Without that, the server does not know to set the missiles as locked on target.

This is my suggested work around until we are ready to bump the multi version with all the new multi stuff we are cooking up.  Here's network/multi_obj.cpp, line 283, where the client figures out whether it should tell the server that it's locked on target.
```
if ( Player_ai->current_target_is_locked ){
		out_flags |= OOC_TARGET_LOCKED;
	}
	if ( Player_ai->ai_flags[AI::AI_Flags::Seek_lock] ){	
		out_flags |= OOC_TARGET_SEEK_LOCK;
	}
```
Just before this section, I would add a loop through the players lock_info struct to check for any target that is `locked`  and then sets the `OOC_TARGET_LOCKED` bit if it finds it.

Bonus points if you can do the same for OOC_TARGET_SEEK_LOCK.  I don't know the side effects of not setting this "properly", but it could cause weird AI side effects.

A full fix, as I mentioned before, would be creating a new packet type for multilock info, but that has to wait.

code/ship/ship.cpp
TRBlount: Like so?

```cpp
        // queue up valid missile locks
	for (i = 0; i < shipp->missile_locks.size(); ++i) {
		if (shipp->missile_locks[i].locked) {
			shipp->missile_locks_firing.push_back(shipp->missile_locks[i]);
		}
	}
	// Using copy_if
	std::copy_if(shipp->missile_locks.begin(), shipp->missile_locks.end(),
				 std::back_inserter(shipp->missile_locks_firing), [](lock_info lock) { return lock.locked; });```

File: code/ship/ship.cpp (hunk 4 above)
Comment (TRBlount) [reply code/ship/ship.cpp]: I believe this should work:
```cpp
return any_of(shipp->missile_lock.begin(), shipp->missile_lock.end(), 
           [](lock_info lock){ return lock.locked; });
```

CasparKielwein: looks good, just a small typo. ´´missile_locks´´ with an s

code/weapon/weapon.h
TRBlount: I'd like further discussion on any particular reason to change that here instead of passing over many of our defines in one go.

Goober5000: Yeah, I don't have a problem with deferring that and then making a separate PR to change things to `constexpr`.

code/ship/ship.h
TRBlount: As suggested on Discord, current proposal is to merge multilock so current master can better serve as a baseline for the imminent multiplayer version bump. Any objections?

TRBlount:
This should be good to merge. All comments have been resolved and CIs are passing.

Issue #1210, opened 2017-02-10
Undefined Behavior Sanitizer errors


chief1983:
Ran the standalone server on a Gentoo box after compiling with the Undefined Behavior sanitizer enabled, and I ran across a few errors on startup.
```
scpuser@fileserver /usr/share/games/freespace2 $ ./fs2_open_3_7_5_20170210_pipefix_san-DEBUG -standalone
/usr/share/games/fs2source/fs2_open_git/code/math/staticrand.cpp:29:27: runtime error: left shift of 1046654190 by 15 places cannot be represented in type 'int'
/usr/share/games/fs2source/fs2_open_git/code/math/staticrand.cpp:29:44: runtime error: signed integer overflow: 732102656 + 1656759491 cannot be represented in type 'int'
/usr/lib/gcc/i686-pc-linux-gnu/4.9.3/include/g++-v4/bits/move.h:175:19: runtime error: load of value 190, which is not a valid value for type 'bool'
```
GCC 4.9.3 was the compiler.

On a GCC 5.4.0 box, I saw those plus another:
```
/home/cliff/fs2_open/code/math/staticrand.cpp:29:27: runtime error: left shift of 1921383941 by 15 places cannot be represented in type 'int'
/home/cliff/fs2_open/code/math/staticrand.cpp:29:44: runtime error: signed integer overflow: 1951268864 + 1968368751 cannot be represented in type 'int [32]'
/usr/include/c++/5/bits/move.h:185:19: runtime error: load of value 190, which is not a valid value for type 'bool'
/home/cliff/fs2_open/code/network/psnet2.cpp:1389:33: runtime error: load of misaligned address 0x0c73986a for type 'unsigned int', which requires 4 byte alignment
0x0c73986a: note: pointer points here
 31 70  4d 44 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00
              ^ 
```
The last one occurred after a client started a standalone game.

chief1983:
I imagine we can't help the one in the system code, but I am trying to see if there is a GCC update for the machine that might remove that one.

chief1983:
There is an update, 4.9.4 is stable on Gentoo now.  But it's only on 37/206 packages that needed updating, so I can probably give that one another test on Monday :P

chief1983:
Updated description since I saw another error on a standalone.

chief1983:
[...]

PR #2208, opened 2019-12-22
make internal variables in static_rand unsigned.

CasparKielwein:
several of the bitwise manipulations on signed integers are technically undefined behaviour.
I made the internal variables unsigned and the results are then converted to signed ints as before.
This should not actually change any observable results.

Did some cleanup as well, while I was at it.

Fixes the first part of  #1210

### DIFF
--- a/code/math/staticrand.cpp
+++ b/code/math/staticrand.cpp
@@ -7,26 +7,21 @@
  *
 */ 
 
-
-
 #include "math/staticrand.h"
 #include "math/vecmat.h"
 
-
-int Semirand_inited = 0;
-int Semirand[SEMIRAND_MAX];
+bool Semirand_inited = false;
+unsigned int Semirand[SEMIRAND_MAX];
 
 /**
  * @brief Initialize Semirand array. Doesn't have to be called.
  */
 void init_semirand()
 {
-	int	i;
+	Semirand_inited = true;
 
-	Semirand_inited = 1;
-
-	for (i=0; i<SEMIRAND_MAX; i++)
-		Semirand[i] = (myrand() << 15) + myrand();
+	for (auto & number : Semirand)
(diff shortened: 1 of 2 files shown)


code/math/staticrand.cpp
```diff
@@ -7,26 +7,21 @@
...
 
-
-
 #include "math/staticrand.h"
 #include "math/vecmat.h"
 
-
-int Semirand_inited = 0;
+bool Semirand_inited = false;
 int Semirand[SEMIRAND_MAX];
```
MageKing17: If this is only used internally, in code sections that now constantly convert it to unsigned, why not just make it an array of unsigned ints?

CasparKielwein: true,
in my first try, I converted the whole interface to unsigned.
When I reverted that, I missed this bit.

code/math/staticrand.cpp
```diff
@@ -37,19 +32,20 @@ void init_semirand()
...
-	c = (num >> (2 * SEMIRAND_MAX_LOG)) & (SEMIRAND_MAX - 1);
+	const unsigned int num_unsigned = num;
+	auto a = num_unsigned & (SEMIRAND_MAX - 1);
+	auto b = (num_unsigned >> SEMIRAND_MAX_LOG) & (SEMIRAND_MAX - 1);
+	auto c = (num_unsigned >> (2 * SEMIRAND_MAX_LOG)) & (SEMIRAND_MAX - 1);
 
-	return Semirand[a] ^ Semirand[b] ^ Semirand[c];
+	return Semirand[a]
+			^ Semirand[b]
+			^ Semirand[c];
```
MageKing17: These lines are now spread out for no reason.

CasparKielwein: narf, I'm to used to having an auto-formatter on save.

PR #2209, opened 2019-12-23
Fixes accidental break with multi due to max key numbers

wookieejedi:
This PR reduces the number of controls back down to 127 so as to keep multi compatibility.


wookieejedi:
Just updated thanks!

PR #2210, opened 2019-12-24
Fix overriding the default values in control table for `disabled` and `continuous_ongoing`

wookieejedi:
Previously, `control_config_common_init()` started with a loop that set the `disabled` and `continuous_ongoing` values to false for all controls in the config array. This overrides the default values set in the array. For example, if a control is set to be disabled by default then this loop overrides that incorrectly.

These lines were added in the following commits:

https://github.com/scp-fs2open/fs2open.github.com/commit/45da46eddd9d9249bf1178ac18164e4ab43570e1#diff-45a0618f2fa10c25b123d4a771f98228

and

https://github.com/scp-fs2open/fs2open.github.com/commit/61d9ce895be392c048437e985cf583ee2fc6161a#diff-45a0618f2fa10c25b123d4a771f98228

The goal of adding these lines was to initiate the values, but these values were already set in the struct table and those lines simply override the values. No other fields in the config array are set like this and simply use the values set in the array.

In tests I have removed these lines and printed what the default values are and the print statements print what is expected from the config array. (For example, `mprintf(("Check_control_disabled_for %s and value is %s...\n", Control_config[i].text, Control_config[i].disabled ? "true" : "false"));` printed what was in the array correctly). 

This PR removes those unneeded lines and thus allows the fields for `disabled` and `continuous_ongoing` to be respected as the default. Of course table changes can be used to alter these, but that's only if a controlconfigdefaults.tbl is present.

### DIFF
--- a/code/controlconfig/controlsconfigcommon.cpp
+++ b/code/controlconfig/controlsconfigcommon.cpp
@@ -590,12 +590,8 @@ void control_config_common_load_overrides();
 // initialize common control config stuff - call at game startup after localization has been initialized
 void control_config_common_init()
 {
-	for (int i=0; i<CCFG_MAX; i++) {
-		Control_config[i].disabled = false;
-		Control_config[i].continuous_ongoing = false;
-	}
-
 	control_config_common_load_overrides();
+
 	if(Lcl_gr){
 		Scan_code_text = Scan_code_text_german;
 		Joy_button_text = Joy_button_text_german;


z64555:
What I find puzzling is that the control_config_common_load_overrides() should be loading the default `disabled` status right after the initial sweep by the common_init().

wookieejedi:
Makes sense. Just so I can understand better, why can the field of `.text` safely use the hard-coded string value in the control config struct while other fields like `.disabled` can not use it safely?

Goober5000:
@z64555 I don't understand your comment?  How would Control_config change over the course of the mission?

Actually a clarification would be helpful.  Is Control_config supposed to be a reference that doesn't change over the course of a game, like `ship_info`, or is it something that can be changed arbitrarily with flags and configurations, like `ship`?

z64555:
@Goober5000. Control_config contains both variable members and constant members.

Variable members are the keybindings, disabled state, and continous_ongoing state.

Most other members are constant

Usage of the terms "variable" and "constant"  correlate to members that may change during a mission (either by the user or a script) and members that can not (name of the control, the control mode type)

wookieejedi:
Thanks for the updated explanation. Given that the code actually does the intended behavior, then that makes it properly up to the modder on what controls to disable or enable via the `controlconfigdefaults.tbl`. Given that, I can close the PR if others are satisfied with the discussion.

z64555:
Doing another look-through of the controls code revealed that the disabled state is actually constant.  I may have been thinking of a feature that made this variable  that never was merged into the main repo.

Complete changes as requested previously,  Keep the for loop but omit line 594

wookieejedi:
Thanks. For some reason my repo had some critical issues and I was unable to commit to this PR. I have closed it and opened a new PR #2241.

PR #2211, opened 2019-12-26
Multiutil.cpp cleanup#1

JohnAFernandez:
Catch-all cleanup of multiutil.cpp. I did my best on these changes, but I am not attached to them, so please make any necessary requests for further changes.  No functionality at all should be changed.

I tried to split the commits into three categories: 

Category 1) the removal of 1 unused function with functionality that exists elsewhere, code that has been commented out since the source code was released, and 2 globals that were there to be stored by the engine and incremented and nothing else. (Their discovery during bug hunting -- and my general mental exhaustion from the same -- inspired this PR)

Category 2) changing NULL to nullptr

Category 3) Changing int3()'s into assertion()'s, error()'s, or verify()'s

### DIFF
--- a/code/network/multiutil.cpp
+++ b/code/network/multiutil.cpp
@@ -725,13 +725,16 @@ void multi_assign_player_ship( int net_player_num, object *objp,int ship_class )
 // depends on the parameter ship_class.  Note that if ship_class is invalid, the ship default_player_ship
 // is used.  Returns 1 on success, 0 otherwise
 
-int multi_create_player( int net_player_num, player *pl, const char* name, net_addr* addr, int ship_class, short id)
+void multi_create_player( int net_player_num, player *pl, const char* name, net_addr* addr, int ship_class, short id)
 {
 	int player_ship_class = ship_class;
 	int current_player_count;
 
-	Assert ( net_player_num < MAX_PLAYERS );				// probably shoudln't be able to even get into this routine if no room	
-	
+	Assertion ( net_player_num < MAX_PLAYERS, "FSO is attempting to create too many players in multi. This error should be prevented in the UI, please report!" );				// probably shoudln't be able to even get into this routine if no room	
+	Assertion(addr != nullptr, "A nullptr net address was passed to multi_create_player(). This is a code error, please report.");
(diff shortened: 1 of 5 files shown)


code/network/multiutil.cpp
```diff
@@ -842,7 +843,7 @@ int multi_create_player( int net_player_num, player *pl, const char* name, net_a
 void multi_make_player_ai( object *pobj )
 {
 
-	Assert ( pobj != NULL );
+	Assert(pobj != NULL);
```
CasparKielwein: you missed a nullptr here

code/network/multiutil.cpp
```diff
@@ -2945,14 +2928,13 @@ void multi_get_mission_checksum(const char *filename)
 
 char multi_unit_to_char(float unit)
 {
+	Verify((unit > 1.0f) && (unit < -1.0f), "Newly implemented function multi_unit_to_char() was passed a value %f, that it cannot handle.", unit);
```
CasparKielwein: here and below:
the Verify macro only supports a single argument.
You could use the old hack/trick of "anding" the check and the message.
The format is not possibly unfortunately. Asserts and their variants like Verify usually only support string literals.

```Verify((unit > 1.0f) && (unit < -1.0f) && "Newly implemented function multi_unit_to_char() was passed a value, that it cannot handle.");```

JohnAFernandez: Ah. I probably should avoid using a hack in a cleanup.  I'll just switch to Assert.  Whoever wants to implement this function can do it using debug builds.

code/network/multiutil.cpp
JohnAFernandez: I think I have this replaced with an assertion but I forgot to commit the changes.

JohnAFernandez:
Just as a note, it looks like in a few places I'm just removing int3()'s, but there was no way to reach an error value in the first place for those.  A lot of this file seems to have been a WIP when the game was released, so they were probably planning on adding some.  Most of the problems for the observer and player create functions would come from bad arguments being passed. There's also no packet being sent or received, so Assertions work really well here.

JohnAFernandez:
One multi file down, 73 to go. >.>

PR #2213, opened 2019-12-27
Minor cleanup to Verify and loading bar

Goober5000:
Some small changes:
a) Fix admonishment about commenting Assert and fix spelling
b) rename VerifyEx to Verification to be similar to Assertion
c) remove old pre-2005 MSVC code
d) update `COUNT_ESTIMATE` to provide a more accurate loading bar, per discussion in Discord

### DIFF
--- a/code/globalincs/pstypes.h
+++ b/code/globalincs/pstypes.h
@@ -240,11 +240,11 @@ extern int Global_error_count;
 // or,
 // Error( LOCATION, "Error opening %s", filename );
 
-/*******************NEVER UNCOMMENT Assert ************************************************/
-// Please never uncomment the functionality of Assert in debug
+/*******************NEVER COMMENT Assert ************************************************/
+// Please never comment the functionality of Assert in debug
 // The code, as with all development like this is littered with Asserts which are designed to throw
 // up an error message if variables are out of range.
-// Disabling this functionality is dangerous, crazy values can run rampent unchecked and the longer its disabled
+// Disabling this functionality is dangerous, crazy values can run rampant unchecked, and the longer it's disabled
 // the more likely you are to have problems getting it working again.
 #if defined(NDEBUG)
 #	define Assert(expr) do { ASSUME(expr); } while (false)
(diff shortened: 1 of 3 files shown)


MageKing17:
> d) update `COUNT_ESTIMATE` to provide a more accurate loading bar, per discussion in Discord

You clearly remember that conversation differently than I do; the only opinion I saw expressed was that the idea of `COUNT_ESTIMATE` itself is fucked up, not that the number is too low (it works pretty well, for retail). If anything, I'd think adding some sort of field to the mission file that overrides `COUNT_ESTIMATE` would make far more sense than just bumping the `#define`.

Goober5000:
I was referring to Discord only inasmuch as it alerted me to the issue, not to any potential course of action.  The purpose of that change isn't to authoritatively fix the loading bar, it's to bring the estimate up to date.

MageKing17:
Up to date with what, though? It's only a useful estimate if it's the approximate value a majority of missions will have; if it's just "this is the highest number I personally saw", it's equally as arbitrary as the current number.

Goober5000:
See the comment.  According to the Volition coders, "it seems to always be about the same", so it should indeed be approximate for a majority of missions.

The-E:
The original V comment only made sense in the context of FS2 though, as the retail missions aren't all that complex (and, generally speaking, fall into a relatively narrow band of complexity in terms of assets used). I would like to know where this "719" number is now coming from; Which mods did you survey in order to get it?

The larger point being that it makes little sense to me to bump this number; IMHO, if this is an area that we want to improve, we should improve it by making the load bar actually accurate by incrementing it for each ship or weapon model loaded or pre-mission scripting hook executed.

Goober5000:
Good grief, this is bikeshedding.  As I said, this is just to bring the estimate more in line with the current state of the code, it's not supposed to be an authoritative fix.  If anyone wants to submit a PR with a more comprehensive rewrite of the loading bar, he is welcome to do so.  That is outside the scope of this particular PR.

I looked at one mission log, loading a retail mission on MVPs 4.0, to get the 719 number.  I just checked a log from Scroll and the number there is 781.  The difference between 425 and 719 is almost 70%, while the difference between 719 and 781 is 8.6%.  The point is that there is a dramatic mismatch between the retail estimate and the current state of the code, while there is a much smaller mismatch between different mods which run on the current state of the code.

Given @CasparKielwein 's approval, I'm merging this.  Again, if anyone wishes to do a more comprehensive upgrade of the loading bar, by all means please do so.

PR #2214, opened 2019-12-28
Multi packet fixes

notimaginative:
Work in progress to fix issues with various multi packets:
- space saving changes
- resolve possible cross-platform issues
- try to avoid packet breakage with engine changes

This should track every item fixed or changed until it's ready to merge.


notimaginative:
Known issues/changes remaining:
- [x] `net_addr` struct - fixed in IPv6 port, just noting it here
- [ ] `beam_fired_packet()` - `beam_info` struct and new fields (check comments in function)
- [ ] `send_sexp_packet()` - should use `ADD_DATA()` macro
- [ ] IEEE-754 packing/unpacking for floats (see also RFC 4506)

wookieejedi:
As an FYI @JohnAFernandez is also working on multi fixes that have been submitted in a PR yet, but you both probably already know that(?).

notimaginative:
Yep, we're keeping track of each others work :)

code/network/multimsgs.cpp
```diff
@@ -7676,74 +7768,111 @@ void process_NEW_countermeasure_fired_packet(ubyte *data, header *hinfo)
...
-	ADD_USHORT(shooter->net_signature);
-	ADD_DATA(subsys_index);
+
+	ADD_USHORT(fire_info->shooter->net_signature);
 	ADD_USHORT(target_sig);
 	ADD_SHORT(u_beam_info);
-	ADD_DATA(b_info);  // FIXME: This is still wrong, we shouldn't be sending an entire struct over the wire - taylor
-//	ADD_DATA(bfi_flags);	// this breaks the protocol but is here in case we decided to do that in the future - taylor
-//	ADD_DATA(target_pos);	// ditto - Goober5000
+	ADD_FLOAT(fire_info->accuracy);
```
Baezon, line 7829: Accuracy is only used to generate `shot_aim` so it shouldn't be necessary to send.

code/network/multimsgs.cpp
```diff
@@ -7676,74 +7768,111 @@ void process_NEW_countermeasure_fired_packet(ubyte *data, header *hinfo)
...
+		ADD_FLOAT(override->dir_a.xyz.y);
+		ADD_FLOAT(override->dir_a.xyz.z);
+		// dir_b
+		ADD_FLOAT(override->dir_b.xyz.x);
+		ADD_FLOAT(override->dir_b.xyz.y);
+		ADD_FLOAT(override->dir_b.xyz.z);
+
+		ADD_FLOAT(override->delta_ang);
+
+		ADD_DATA(override->shot_count);
```
Baezon, line 7852: As mentioned in the PR, (after the PR) shot_count won't need to be sent either.

JohnAFernandez: You'll find that counting on tabled information on packet receipt is not as reliable as it would seem.  While the risk here is much lower than in the object update packet, relying on the number of shield quadrants was causing multi crashes for years.  I couldn't rely on the subsystem information either.  

This is probably safe, just be cautious about your assumptions in multi, because they can cause mysterious crashes.

JohnAFernandez: Also, with IPv6 packet sizes, the shot_count would be only 1/1000th of the packet.

Baezon: I see, well I'll defer to your guys' judgment on this, but my chief concern was reducing the number of things that need to be sent. I am really quite worried if table data can't be relied upon, how on earth can that be the case?

JohnAFernandez: FSO is a minefield and things are just not all completely initialized exactly how you expect -- or at least that's what I've gathered.  Even though a ship is created, its subsystems may not be completely created or maybe it's just not put into the linked subsystem list yet.  

Or maybe there's a difference between the two computers because someone made some manual changes.  Yeah, their stats won't be saved on a PXO game, but no one is going to care, and then the beams you see on the client and server will differ.  

I honestly don't know enough about beams to say anything about this situation, I'm just saying I made this mistake early on.

JohnAFernandez: I should explain.  The server runs the simulation, and so the more information we can get from the server about how everything runs, the better.

Baezon: Fair enough, I'd like to see what taylor says though.

code/network/multimsgs.cpp
notimaginative: Gotcha, I'll remove that.

code/network/multimsgs.cpp
notimaginative: Yeah, the shot count still needs to be added to the packet. With your changes it will only be used for making sure we get exactly the amount of data out of the packet as was put in though.

It's just not safe to depend on the client and server having the exact same data. For instance, if `MAX_BEAM_SHOTS` was changed then it could break multi if we didn't have the safety of the shot count in the packet.

notimaginative:
After the turret and flak packets are fixed I think I'll be done with this. I'm sure that there are other issues still, but it might be best to save those for a future round of fixes.

notimaginative:
@Baezon When you get some time could you give the beam_fired_packet() changes another look?

It should have everything we've talked about but I just want to make sure. I squashed it all down so you'll only have to check that single commit:
https://github.com/scp-fs2open/fs2open.github.com/pull/2214/commits/bf5acd46925afab1c3c3593099fd84a516aeed44

notimaginative:
No further changes planned at this time. There are still improvements to be made but most are likely much larger than the original intended scope of this PR.

Baezon:
It still can't handle 'floating' beams (sexp/spawned/ssm) since it will early return if it can't find a shooter or turret in those cases, but I'm not sure if you were looking on supporting that just yet with this PR?

notimaginative:
Well crap, I totally missed that. I dealt with it on the receiving end but forgot about the check on the sending side. I did intend to support floating beams so I'll get that fixed.

Is there anything else you notice that's off?
And does this include all of the changes required for your other PRs?

Baezon:
Yep, it otherwise looks good! I noticed you were handling `starting_pos` so it indeed seemed a bit strange.

Baezon (approved):
LGTM, at least as far as the beam stuff goes. Obviously, we'll want a few more eyeballs than just mine on this PR though.

code/network/multi_sexp.cpp
```diff
@@ -9,8 +9,8 @@
 #include "network/multi.h"
 #include "network/multiutil.h"
 
-#define CALLBACK_TERMINATOR	255
-int TEMP_DATA_SIZE = -1;
+static const std::uint8_t CALLBACK_TERMINATOR = 255;
+static const short TEMP_DATA_SIZE = -1;
```
z64555, line 13: If this needs to be 2 bytes across all systems, then use an int16_t. "short" "int" and "long" are still subjective to the architecture.

notimaginative: There is a safety check in the ADD_* macros to make sure the size is correct, so that shouldn't be a problem. The code won't even compile if the type size doesn't match what's expected.

Still, I would normally use a properly sized data type for this and I'm not sure why I didn't.

code/network/multimsgs.h
```diff
@@ -28,32 +28,39 @@ struct header;
...
+#define ADD_INT(d) do { static_assert(sizeof(d) == sizeof(std::int32_t), "Size of int is not right!"); Assert((static_cast<size_t>(packet_size) + sizeof(d)) < MAX_PACKET_SIZE); int swap = INTEL_INT(d); memcpy(data+packet_size, &swap, sizeof(d) ); packet_size += sizeof(d); } while (false)
+#define ADD_UINT(d) do { static_assert(sizeof(d) == sizeof(std::uint32_t), "Size of unsigned int is not right!"); Assert((static_cast<size_t>(packet_size) + sizeof(d)) < MAX_PACKET_SIZE); uint swap = INTEL_INT(d); memcpy(data+packet_size, &swap, sizeof(d) ); packet_size += sizeof(d); } while (false)
+#define ADD_ULONG(d) do { static_assert(sizeof(d) == sizeof(std::uint64_t), "Size of unsigned long is not right!"); Assert((static_cast<size_t>(packet_size) + sizeof(d)) < MAX_PACKET_SIZE); std::uint64_t swap = INTEL_LONG(d); memcpy(data+packet_size, &swap, sizeof(d) ); packet_size += sizeof(d); } while (false)
```
z64555, line 45: These macros have been using standard types for the buffer to swap the data into correct endianess, but this one is using a uint64_t?  May want to double check this.

Also remember that standard types may turn out to be smaller than their standards if on an architecture that has an int width smaller than 32bits.  Since FSO isn't targeted for 16bit architectures or below, I think we can let this slide, but it is something to be aware of.

code/network/multimsgs.h
```diff
@@ -28,32 +28,39 @@ struct header;
...
+#define ADD_FLOAT(d) do { Assert((static_cast<size_t>(packet_size) + sizeof(d)) < MAX_PACKET_SIZE); float swap = INTEL_FLOAT(&d); memcpy(data+packet_size, &swap, sizeof(d) ); packet_size += sizeof(d); } while (false)
+#define ADD_STRING(s) do { Assert((static_cast<size_t>(packet_size) + strlen(s) + sizeof(uint8_t)) < MAX_PACKET_SIZE); Assert(strlen(s) <= UINT8_MAX); uint8_t len = static_cast<uint8_t>(strlen(s)); ADD_DATA(len); memcpy(data+packet_size, s, len); packet_size += len; } while (false)
+#define ADD_STRING_16(s) do { Assert((static_cast<size_t>(packet_size) + strlen(s) + sizeof(uint16_t)) < MAX_PACKET_SIZE); uint16_t len = static_cast<uint16_t>(strlen(s)); ADD_USHORT(len); memcpy(data+packet_size, s, static_cast<size_t>(len)); packet_size += len; } while (false)
+#define ADD_STRING_32(s) do { Assert((static_cast<size_t>(packet_size) + strlen(s) + sizeof(int32_t)) < MAX_PACKET_SIZE); int32_t len = static_cast<int32_t>(strlen(s)); ADD_INT(len); memcpy(data+packet_size, s, static_cast<size_t>(len)); packet_size += len; } while (false)
```
z64555, line 49: ADD_STRING_16 uses an uint16_t for len, but ADD_STRING_32 uses an int16_t?

code/network/multi_campaign.cpp
```diff
@@ -40,13 +40,15 @@
...
 #define MC_CODE_START								3			// start campaign information
+#define MC_CODE_MISSION_NAMES						4			// mission names
 
 // packet code for ingame joining
 #define MC_JIP_INITIAL_PACKET						1			// initial data 
 #define MC_JIP_GE_STATUS							2			// goal event status for all missions
 #define MC_JIP_GOAL_NAMES							3			// goal names
 #define MC_JIP_EVENT_NAMES							4			// event names
-#define MC_JIP_END_DATA								5			// this is the end of the data
+#define MC_JIP_MISSION_NAMES						5			// mission names
```
z64555, line 50: Can these be enum's or enum classes?

code/network/multi_campaign.cpp
```diff
@@ -55,6 +57,9 @@
 // flags indicating the "accept" status of all players involved in a campaign
 int Multi_campaign_accept_flags[MAX_PLAYERS];
 
+// safety checks
+static_assert(MAX_CAMPAIGN_MISSIONS <= UINT8_MAX, "MAX_CAMPAIGN_MISSIONS is set too high and breaks multi!");
+static_assert(MAX_GOALS <= UINT8_MAX, "MAX_GOALS is set too high and breaks multi!");
```
z64555, line 62: Nice use of static_assert

code/network/multimsgs.cpp
```diff
@@ -2173,76 +2195,98 @@ void process_netgame_update_packet( ubyte *data, header *hinfo )
...
-	ubyte data[MAX_PACKET_SIZE],val;
-	int desc_len;
+	ubyte data[MAX_PACKET_SIZE];
 	int packet_size = 0;
 
-	// build the header
-	BUILD_HEADER(UPDATE_DESCRIPT);
+	// Get this out of the way, because we don't want to waste our time if the addr is bad.
+	Assert(addr != nullptr);
+	if (addr == nullptr) {
```
z64555, line 2206: Double check the use of Assert, I forget if it allows resumption or not.

code/network/multimsgs.cpp
```diff
@@ -2173,76 +2195,98 @@ void process_netgame_update_packet( ubyte *data, header *hinfo )
...
-			ADD_STRING(The_mission.mission_desc);
-		}
-	} 
-	
-	Assert(addr != NULL);
-	if(addr != NULL){
 		psnet_send(addr, data, packet_size);
 	}
+	// otherwise, break up description string if necessary and send to client
+	else {
```
z64555, line 2221: Somewhat odd style for `else`

code/network/multimsgs.cpp
```diff
@@ -2173,76 +2195,98 @@ void process_netgame_update_packet( ubyte *data, header *hinfo )
...
+		}
 
 		// send an update to this guy
 		send_netgame_descript_packet(&addr, 1);
-	} else {	
-		memset(mission_desc,0,MISSION_DESC_LENGTH+2);		
-		GET_STRING(mission_desc);
+	}
+	// otherwise 'code' will be an index indicating the message number - 1, 2, ...
+	else {
```
z64555, line 2271: Somewhat odd style for `else`. Check for style consistency in rest of file.

code/network/multimsgs.cpp
```diff
@@ -4047,8 +4082,9 @@ void send_netplayer_slot_packet()
 			ADD_DATA(stop);
 			ADD_SHORT(Net_players[idx].player_id);  
 			ADD_USHORT(Objects[Net_players[idx].m_player->objnum].net_signature);
-			ADD_INT(Net_players[idx].p_info.ship_class);
-			ADD_INT(Net_players[idx].p_info.ship_index);			
+			ADD_SHORT(static_cast<short>(Net_players[idx].p_info.ship_class));
+			val = static_cast<char>(Net_players[idx].p_info.ship_index);
+			ADD_DATA(val);
```
z64555, line 4087: Seems odd that index is smaller than ship class

code/network/multimsgs.cpp
```diff
@@ -5632,8 +5684,9 @@ void send_player_settings_packet(net_player *p)
 			// break the p_info structure by member, so we don't overwrite any absolute pointers
 			// ADD_DATA(Net_players[idx].p_info);
 			ADD_INT(Net_players[idx].p_info.team);
-			ADD_INT(Net_players[idx].p_info.ship_index);
-			ADD_INT(Net_players[idx].p_info.ship_class);
+			ADD_SHORT(static_cast<short>(Net_players[idx].p_info.ship_index));
+			val = static_cast<char>(Net_players[idx].p_info.ship_class);
```
z64555, line 5688: Hold it, this is backwards from before.  ship_index was a char while ship_class was a short.

code/network/multimsgs.h
notimaginative: That was an existing macro and I'm not sure who added it. I just added an extra safety check to them for the packet size.

code/network/multimsgs.h
notimaginative: ADD_STRING_32 is only there for compatibility with the original ADD_STRING, which used an signed int. It's only used in a single packet which does the initial game query and must remain retail compatible.

code/network/multimsgs.cpp
notimaginative: I've always done commenting that way. Picked up the habit from the freespace2 code base :)

code/network/multi_campaign.cpp
notimaginative: Probably, but I didn't want to change it just to change it.

code/network/multimsgs.cpp
notimaginative: I'm not sure. This is a leftover from the original code and should probably be removed/changed though. I'll add it to the list.

code/network/multimsgs.cpp
notimaginative: The index is for loadout and is < 12. The class is the Ship_info index. It confused me too.

code/network/multimsgs.cpp
notimaginative: Oof, good catch. That is indeed backwards.

z64555 (changes requested):
Address issue with ship_index and ship_class being the wrong types.

notimaginative:
lol. Fixed it three days ago and forgot to push the changes 🤦

z64555 (approved):

Issue #2215, opened 2019-12-30
Enhancement: Target nearest enemy starfighter


countbuggula:
Please consider adding a key binding to target the nearest enemy starfighter.  Currently there are only options to target nearest attacking craft and nearest enemy craft.  In missions with large amounts of capital ships, this would make finding enemy starfighters (especially those attacking other craft) much easier to find.

Note, this is a feature that has been included in every X-Wing game since the original '93 release.

Goober5000:
Closing per comments on other ticket

z64555:
Scripted controls is a feature we're working on, and would like mod devs to try experimenting with them.  As the controls code is upgraded over time (hopefully within the next few months) the scripted controls support will be expanded and made more robust.

PR #2216, opened 2019-12-30
Adds Target Closest Hostile Fighter

wookieejedi:
In reference to issue #2215, this PR adds a new control that allows the player to target the closest hostile fighter.

I am also open to discussion on whether it would better to incorporate this target control as a script using the new custom controls rather than a PR.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -1554,6 +1554,67 @@ void hud_target_missile(object *source_obj, int next_flag)
 	}
 }
 
+void hud_target_hostile_fighter(object* source_obj, int next_flag)
+{
+	object *A, *target_objp;
+	ai_info* aip;
+	int target_found = 0;
+
+	if (source_obj->type != OBJ_SHIP)
+		return;
+
+	Assert(Ships[source_obj->instance].ai_index != -1);
+	aip = &Ai_info[Ships[source_obj->instance].ai_index];
+
+	ship_obj *startShip, *so;
+
+	if ((aip->target_objnum != -1) && (Objects[aip->target_objnum].type == OBJ_SHIP) &&
+		(Ship_info[Ships[Objects[aip->target_objnum].instance].ship_info_index].flags[Ship::Info_Flags::Fighter])) {
+		int index = Ships[Objects[aip->target_objnum].instance].ship_list_index;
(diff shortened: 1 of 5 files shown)


code/hud/hudtarget.cpp
```diff
@@ -1554,6 +1554,67 @@ void hud_target_missile(object *source_obj, int next_flag)
...
+			continue;
+		}
+
+		// check if ignore
+		if (should_be_ignored(&Ships[A->instance])) {
+			continue;
+		}
+
+		// found a good one
+		target_found = 1;
```
asarium: If you do the `snd_play` somewhere here then you don't need this variable or the `if` further below.

code/hud/hudtarget.cpp
```diff
@@ -1554,6 +1554,67 @@ void hud_target_missile(object *source_obj, int next_flag)
 	}
 }
 
+void hud_target_hostile_fighter(object* source_obj, int next_flag)
+{
+	object *A, *target_objp;
+	ai_info* aip;
+	int target_found = 0;
```
asarium: Just an FYI, if you use variables that contain a boolean variable please use `bool` instead of `int`. A lot of code here is pretty old when there was no `bool` yet so that is why you might see such usage in some cases.

code/hud/hudtarget.cpp
```diff
@@ -1554,6 +1554,67 @@ void hud_target_missile(object *source_obj, int next_flag)
...
+
+		// found a good one
+		target_found = 1;
+		set_target_objnum(aip, OBJ_INDEX(A));
+		hud_shield_hit_reset(A);
+		break;
+	}
+
+	if (target_found) {
+		snd_play(gamesnd_get_game_sound(GameSounds::TARGET_FAIL), 0.0f);
```
asarium: Why `TARGET_FAIL`? That seems more like something for when no target is found...

wookieejedi:
> Have you tried looking into a scripting based solution for this? I would like to avoid adding new controls if the same functionality could also be provided via a script.

Yeah, I'm pretty sure that I could get this same functionality via scripting, and thinking about it more that would also be useful for modders to have more control over which types of small craft can be targeted in the controls. Given your thoughts as well I think I'll go ahead and close this PR. Thanks for providing detailed comments on the PR too, I took notes and will remember them moving forward!

PR #2223, opened 2020-01-05
Fix Setting of Initial Velocity

wookieejedi:
It turns out that the code that set initial ship velocities did not account for local ship rotation changes. This bug is only noticeable when ships are very high speeds, high initial velocities, and have high drift values thus it was never recognized before. This PR fixes that by ensuring that the velocity vectors are set appropriately. As discussed with Goober on Discord, this fixed behavior is indistinguishable when using FS ships and most mods, thus the behavior will be on by default.


JohnAFernandez:
Fantastic that we don't need a flag!  :)

PR #2231, opened 2020-01-07
Better Timing For Interpolation Calc

JohnAFernandez:
I'll just leave this here....

Before: https://youtu.be/dsdS6d1q93A?t=25

After: https://youtu.be/P8sdYKtmjO8?t=36

Does this by refactoring how the timing is calculated.  It also removes the Next Arrival Time Array, which does not actually do a good job of calculating it, anyway.  We can now rely on the other oo_arrive_time arrays exclusively by subtracting from current time.

There are further improvements not included in this pull request which require a multiplayer version bump, which is forthcoming.

### DIFF
--- a/code/network/multi_obj.cpp
+++ b/code/network/multi_obj.cpp
@@ -40,13 +40,12 @@
 float oo_arrive_time[MAX_SHIPS][5];				// the last 5 arrival times for each ship
 int oo_arrive_time_count[MAX_SHIPS];			// size of the arrival queue
 float oo_arrive_time_avg_diff[MAX_SHIPS];		// the average time between arrivals
-float oo_arrive_time_next[MAX_SHIPS];			// how many seconds have gone by. should be equal to oo_arrive_time_avg_diff[] the next time we get an update
 
 // interp stuff
 int oo_interp_count[MAX_SHIPS];
 vec3d oo_interp_points[MAX_SHIPS][2];
-bez_spline oo_interp_splines[MAX_SHIPS][2];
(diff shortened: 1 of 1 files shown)


JohnAFernandez:
That /should/ fix Fred compilation.

code/network/multi_obj.cpp
```diff
@@ -1873,48 +1870,55 @@ void multi_oo_interp(object *objp)
 		return;
 	}	
 
-	// increment his approx "next" time
-	oo_arrive_time_next[objp->instance] += flFrametime;
-
-	// do stream weapon firing for this ship
+	// Now that we know we have a vaild ship, do stream weapon firing for this ship before we do anything else that makes us abort
```
Goober5000: "valid" is misspelled

code/network/multi_obj.cpp
```diff
@@ -1873,48 +1870,55 @@ void multi_oo_interp(object *objp)
...
 		physics_sim(&objp->pos, &objp->orient, &objp->phys_info, flFrametime);
 		return;
+	// in case they are slightly late, we'll just use the last good point.
+	} else if (t > 1.0f) {
+		t = 1.0f;
 	}	
 
-	// otherwise, blend the two curves together to get the new point
+	// Cyborg17 - we are no longer blending the two curves.  I'm not sure *how*, but they were somehow making the interpolation look 
+	// less eratic when the timing was messed up in the first place.  Now we take the *good* curve at its word.
```
Goober5000: "erratic" is misspelled

JohnAFernandez:
Comments should be satisfactorily changed.

Goober5000:
Did you push the change?

JohnAFernandez:
Whoops ... nope

PR #2233, opened 2020-01-09
Proposal to cleanup debug log statement regarding turret fire

wookieejedi:
In `aiturret.cpp` there is a `mprintf`, `mprintf(("nothing ok to fire\n"));`. This print statement greatly pollutes the debug log in missions with turreted ships. I'm not sure what this message helps debug, thus I propose commenting it out.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -2731,7 +2731,7 @@ void ai_fire_from_turret(ship *shipp, ship_subsys *ss, int parent_objnum)
 
 		if(!something_was_ok_to_fire)
 		{
-			mprintf(("nothing ok to fire\n"));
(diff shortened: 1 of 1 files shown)


Goober5000:
You might as well remove that line, comment and all.  (And then remove the blank line.)

But before doing that, I recommend running a git blame on the file to see who added the mprintf in the first place, and for what feature.  That might help you figure out what purpose it was supposed to serve.

wookieejedi:
Good idea. I tracked the addition to here: 
https://github.com/scp-fs2open/fs2open.github.com/commit/c91e5a8bd66746e9fb2b91bfee40514ee4e7538f#diff-5fc056e2156a54b5cb98031a7e58b9e9

Back in Apr 19, 2005, titled fix for the slow firing turret problem. Seems to be added there to just make sure target acquisition was working as expected.

PR #2234, opened 2020-01-11
Replace Warning with error_display in parselo

asarium:
error_display shows the line where the error was detected. Warning
itself is pretty useless for parsing related errors so I replaced all
Warnings (where appropriate) with an equivalent error_display.

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -449,9 +449,7 @@ int required_string(const char *pstr)
 	}
 
 	if (count == RS_MAX_TRIES) {
-		nprintf(("Error", "Error: Unable to find required token [%s]\n", pstr));
-		Warning(LOCATION, "Error: Unable to find required token [%s]\n", pstr);
-        throw parse::ParseException("Required string not found");
+		throw parse::ParseException("Required string not found");
 	}
 
 	Mp += strlen(pstr);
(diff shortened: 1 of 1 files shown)


code/parse/parselo.cpp
```diff
@@ -450,8 +450,8 @@ int required_string(const char *pstr)
 
 	if (count == RS_MAX_TRIES) {
 		nprintf(("Error", "Error: Unable to find required token [%s]\n", pstr));
```
MageKing17: `error_display()` already does this, so this line is redundant and can be removed.

code/parse/parselo.cpp
```diff
@@ -667,7 +667,6 @@ int required_string_either(const char *str1, const char *str2)
 	}
 
 	nprintf(("Error", "Error: Unable to find either required token [%s] or [%s]\n", str1, str2));
-	Warning(LOCATION, "Error: Unable to find either required token [%s] or [%s]\n", str1, str2);
```
MageKing17: You removed a warning without adding an equivalent `error_display()`; why?

asarium: There is an `error_display` a few lines above this so this seems to be redundant.

code/parse/parselo.cpp
asarium: Fixed.

code/parse/parselo.cpp
MageKing17: The exact same statement is true of `required_string()` above.

asarium: Indeed, fixed that now.

MageKing17: Now you've just got the inconsistency of removing the `nprintf()` call in one but not the other.

asarium: Not anymore.

PR #2236, opened 2020-01-12
Adds Scripting Support for the Ship Flag `gun convergence`

wookieejedi:
This PR will allow scripters to change the ship flag `gun convergence`. When enabled this flag uses the normals set in the model to fire from, not the ship's forward orientation. 

For reference:
https://wiki.hard-light.net/index.php/Ships.tbl#.22gun_convergence.22


MageKing17:
Whoa, hey, hold on, why was this merged so quickly? This is the first time we've ever allowed modifications of `ship_info` data in the middle of a mission. Why are we suddenly throwing away all the reasons it's a bad idea?

wookieejedi:
Ah I was unaware there was a history and previous agreement that this was a bad idea. Given that, I'd be happy to submit a PR that removes this.

PR #2243, opened 2020-01-19
Allows Modders to Actually Use Engine Wash Intensity

wookieejedi:
Previously, Engine Wash Intensity didn't actually do anything in the code. This adds a game table setting, `$Use Engine Wash Intensity:`, that let's modders specify to the game to actually take the intensity value into account when calculating engine wash effects.

I was going to use a boolean check at each time `wash_intensity` was used, but since the `ship_intensity` value is cumulative that means that each `wash_intensity` needs to be applied to it's respective thruster before being added to the cumulative total.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -577,4 +577,5 @@ void mod_table_reset() {
 	Arc_color_emp_p1 = std::make_tuple(static_cast<ubyte>(64), static_cast<ubyte>(64), static_cast<ubyte>(5));
 	Arc_color_emp_p2 = std::make_tuple(static_cast<ubyte>(128), static_cast<ubyte>(128), static_cast<ubyte>(10));
 	Arc_color_emp_s1 = std::make_tuple(static_cast<ubyte>(255), static_cast<ubyte>(255), static_cast<ubyte>(10));
+	Use_engine_wash_intensity = false;
 }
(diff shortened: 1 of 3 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -55,6 +55,7 @@ std::tuple<ubyte, ubyte, ubyte> Arc_color_damage_s1;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_emp_p1;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_emp_p2;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_emp_s1;
+bool Use_engine_wash_intensity;
```
asarium: This is missing a default value in `mod_table_reset`.

code/ship/shipfx.cpp
```diff
@@ -2526,7 +2527,7 @@ void engine_wash_ship_process(ship *shipp)
 	float dist_sqr, inset_depth, dot_to_ship, max_ship_intensity;
 	polymodel *pm;
 
-	float max_wash_dist, half_angle, radius_mult;
+	float max_wash_dist, half_angle, radius_mult, wash_intensity;
```
asarium: If new variables are introduced, please place them as near to their usage as possible.

code/mod_table/mod_table.cpp
wookieejedi: Thanks, added.

code/ship/shipfx.cpp
wookieejedi: Will do!

wookieejedi:
Sounds good!

Goober5000:
Ok, now that 19.0 is out, we can merge this as soon as @asarium approves.

PR #2244, opened 2020-01-20
CleanUp Unused Function

JohnAFernandez:
This function does not have any contents and should be removed to make the code more readable.

(I'm trying to keep misc changes made in another branch)

It looks like it was here to update a timestamp that might have been planned to be attached to some packets, but no packets use such a timestamp, and such a timestamp could actually be counterproductive, since pings vary.  We rely on sequence numbers within packets instead.


JohnAFernandez:
Added another unused function.

PR #2247, opened 2020-01-26
Use uniform blocks for matrices

asarium:
This is a generic way to get matrix data to shaders without using
(old-school) uniforms. This is another step towards more compatibility
with how new APIs (Vulkan) do things to increase the odds for a Vulkan
port at some point.


asarium:
Unfortunately GitHub Actions now has the same issue as we have on Travis CI where it fails with some obscure error message so I had to disable that here...

PR #2249, opened 2020-01-26
Fix weapon names not filtering # in HUD gauge

wookieejedi:
Weapon names with a `#` should not display text after that `#` in the HUD. For the most part this works, but there were a few cases this didn't (such as primaries using the weapon name). This PR fixes two spots that fell between the cracks. Note that line 6109 was added in that section/order because that is where it appeared in that order on line 6026.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -6013,6 +6013,7 @@ void HudGaugeWeapons::render(float  /*frametime*/)
 		if(Weapon_info[sw->primary_bank_weapons[0]].hud_image_index != -1) {
 			renderBitmap(Weapon_info[sw->primary_bank_weapons[i]].hud_image_index, position[0] + Weapon_pname_offset_x, name_y);
 		} else {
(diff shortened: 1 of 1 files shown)


code/hud/hudtarget.cpp
```diff
@@ -6105,6 +6106,7 @@ void HudGaugeWeapons::render(float  /*frametime*/)
 
 		// print out the ammo right justified
 		sprintf(ammo_str, "%d", ammo);
+		end_string_at_first_hash_symbol(ammo_str);
```
niffiwan: Won't ammo/ammo_str just be an integer, so there shouldn't be any hashes in the string?

wookieejedi: Good catch, just removed it. On 6026 it's a char not int :)

wookieejedi:
Affirmative, thanks.

niffiwan:
git action on linux/release/gcc-9 seems jammed, I can't see a way to restart it and it's not required so merging anyway :)

Issue #2250, opened 2020-01-27
targetbox RTT doesn't scale models correctly in all cases


niffiwan:
There seems to be a model scaling issue with the HUD target box when the following are true:

1) RTT is used
2) The ratio of "canvas size" to "display size" is not the same in both directions.

After looking closely at the images & tracing through code I'm reasonably sure it's an issue with the models, and **not** with drawing the targeting brackets (i.e. similar symptom but not the same cause as #1303)

Here's some images to demonstrate the issue; the 1st image is using RTT, the second is a standard gauge. 

![rtt-model-scaling-nok](https://user-images.githubusercontent.com/3313634/73153202-a6b62280-411e-11ea-99f2-68f224c580f6.png)
![rtt-model-scaling-ok](https://user-images.githubusercontent.com/3313634/73153203-a74eb900-411e-11ea-86f2-37ff887db293.png)

I'll create a repro mod & upload it shortly.

niffiwan:
closed via #2251

PR #2251, opened 2020-01-27
Fixes RTT Target Box Scaling

wookieejedi:
It seems that when RTT for cockpits was added that the scaling of ships in the target box fell through the cracks. Properly specifying the `clip_aspect` fixes the issue. This PR fixes issue #2250.

In detail, the `HudGaugeTargetBox::renderTargetSetup` function scales the drawn ship model by `gr_screen.clip_aspect`. When using RTT gauges the clip_aspect changes from the default, so that changed aspect ration needs to be used instead of the default `gr_screen.clip_aspect`.

### DIFF
--- a/code/hud/hudtargetbox.cpp
+++ b/code/hud/hudtargetbox.cpp
@@ -519,7 +519,7 @@ void HudGaugeTargetBox::renderTargetSetup(vec3d *camera_eye, matrix *camera_orie
 	// account for gauge RTT with cockpit here --wookieejedi
 	float clip_aspect;
 	if (gr_screen.rendering_to_texture != -1) {
-		clip_aspect = (clip_width / clip_height);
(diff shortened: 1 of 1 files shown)


code/hud/hudtargetbox.cpp
```diff
@@ -519,7 +519,7 @@ void HudGaugeTargetBox::renderTargetSetup(vec3d *camera_eye, matrix *camera_orie
 	// account for gauge RTT with cockpit here --wookieejedi
 	float clip_aspect;
 	if (gr_screen.rendering_to_texture != -1) {
-		clip_aspect = (clip_width / clip_height);
+		clip_aspect = ( i2fl(clip_width) / i2fl(clip_height) );
```
asarium: FYI, no spaces needed around the expression. Also, the parenthesis are also not really needed.

PR #2252, opened 2020-01-28
Fix build on FreeBSD

pkubaj:


chief1983:
Interesting, will have to see how this works on my BSD box.

PR #2253, opened 2020-01-28
Update Version to 20.1.0

asarium:


chief1983:
Do we still want to keep nightly/master builds ending in a .1?

asarium:
Doesn't really make a difference to me. I'll set the minor version though since we probably will use the .build part for the revision at some point.

chief1983:
Well I didn't think they should have the exact same version as the final release at least?  Seems like that would cause problems.  But I thought the minor was going to be increased for each subsequent release in a given year?  So 20.1.0 would be the next stable release?

Goober5000:
We should set the version to 20.0.0, publish a release (to fix the builds not appearing on Knossos), then set the version to 20.1.x for the nightlies.

Alternatively, I could set the version to 20.0.0 locally on the nightlybuild machine, run the script for a new 20.0.0 release, and then this PR could set the version to 20.1.x.

cmake/version.cmake
```diff
@@ -6,9 +6,9 @@ if (EXISTS "${PROJECT_SOURCE_DIR}/version_override.cmake")
     include("${PROJECT_SOURCE_DIR}/version_override.cmake")
 endif()
 
-set_if_not_defined(FSO_VERSION_MAJOR 3)
-set_if_not_defined(FSO_VERSION_MINOR 8)
-set_if_not_defined(FSO_VERSION_BUILD 1)
+set_if_not_defined(FSO_VERSION_MAJOR 20)
```
Goober5000: @MageKing17 pointed out, and I agree, that this should be 19.x.y until 20.0 is released.  I suggest setting this to 19.1.0.

chief1983: I remember this having been the original plan but I also proposed that we could continue with 20.1 for nightlies if we wanted to make the first release 20.2.0.  And at some point we would move the date from the revision into the build number for nightlies.  So going forward we could even just change the major version as soon as a new year rolls around regardless of the current nightly version, for instance 20.3 could roll over to 21.1, etc.  Releases would then almost always be y.2, y.4, etc and we'd never use the y.0 for anything, to keep evens as stable and odds as nightlies.

Goober5000: Think about it this way: until the 20-RC1, version 20 is not yet valid.  Therefore the nightly builds and other builds should be 19.x.  The first release for 2020 should be 20.0, not 20.2.

Release numbers should signify milestones.  "Posting a release" is a milestone, but simply having the calendar flip from one year to the next is not.  Ideally the major number will match the year, but the major number should not change solely because the year changed.

chief1983: That doesn't make much sense.  The RC will always just be the complete version string then -RC, so it doesn't matter what that version string is as long as we account for that.  The nightlies can still be 20.1.2020mmdd, and the first RC would be 20.2.0-RC1.  The first stable would just be 20.2.0.  20.1.2020mmdd < 20.2.0-RC1 < 20.2.0.  Still works if we don't stick with the first release having to be 20.0.  I suggested the same thing you are saying the other day in Discord the other day but there was pushback it isn't necessary and I admit that there is no real reason the first release has to be 20.0.  It just would have been nice to have this cemented with 19.x and we could have moved to 19 for the first nightlies insead of 3.8.1, but there's still no reason we can't change this again.  I have to admit I feel like being able to freely change the nightlies when we want to the new year, at the start of the year, seems nicer than forcing us to keep the previous version around for them _just_ so the first stable release each year is a yy.0 release.

MageKing17: I don't see why we should confuse people by having 20.2 be the first release in 2020.

chief1983:
I agree with that as well, but I think the benefit of not keeping the old version around for nightlies might be more beneficial, and also apps do tend to skip versions arbitrarily from time to time.  And at least this would be a predictable pattern.

Goober5000:
19.x is not the old version, it's the *current* version.  It's more recent than 19.0 while not being considered a "stable" release.  The next stable release would be 19.2, or 20.0 if we decided to switch to the current year.

chief1983:
We would definitely make the first release for this year 20.something.  The only question is should we bump the nightlies to 19.1 and have the next release be 20.0, like we did keeping 3.8.1 even though the next release wasn't going to be 3.anything, or do away with relying on the last release having anything to do with the next nightlies and move to 20.something.  We either then do 20.0 for nightlies and 20.1 for the stable, flipping the even/odd meaning, or just go to 20.1 for the nightlies 20.2 for the release to keep evens as stable.

Goober5000:
We should bump the nightlies to 19.1 and have the next release be 20.0.  Keeping 3.8.1 subsequent to release 3.8 was the right thing to do, so the same should be done here.

And we should keep the even/odd meaning now that the convention has been established, otherwise we'll just introduce a lot of confusion.

wookieejedi:
Sounds like Goober's last comment is the consensus?

chief1983:
Still feel like MageKing and him are the only ones who are still preferring it that way.  I'm actually not really caring much either way but there was definitely pushback against that in Discord, that I don't think is settled yet.

Goober5000:
Can you quote or summarize the Discord pushback?

chief1983:
I've pretty much already done that here.  Sounded like more wanting it to be 20.something now and keep evens for stable.  That would mean 20.1 for nightlies now probably and 20.2 for first stable of this year.

Goober5000:
If there isn't more to the Discord pushback than what you described, I don't think it's at all sufficient to overcome the reasons laid out.

The nightlies should be 19.1, and the next release should be 20.0.  The major version number does not get bumped until a major release.

chief1983:
You haven't really laid out reasons that I can see other than it was based on previous discussion.  But I don't see why nightly versions should be arbitrarily chained to the previous release when they have nothing to do with it as they are building to the next release.  It was the thing that annoyed me most about the last cycle, that 3.8.x had to stick around until the stable release.  Not forcing ourselves to stick with .0 as the first release of the year can solve that problem.  Why does the nightly version have to relate to the previous stable release in your opinion?

Goober5000:
Nightly builds are incremental updates to stable builds.  That's why they share the same major version number.  When the major version number increases, it signifies a major update, not an incremental update.  "Building to the next release" only occurs during a release candidate cycle.

chief1983:
We committed major changes immediately after the stable release.  That happens almost every stable release.  The nightlies almost immediately start to resemble the next release more than the old one.  And at some point between the two the nightlies are definitely closer to the next release, yet they carry the association with the previous version up until the stable release is final.  This makes no sense to me, as it seems that each nightly is growing closer to the next release, having the version indicate that would make more sense to me.

MageKing17:
Nightly builds are ultimately not meant to be used by the average player; in an ideal world, stable releases would be often enough that mods would never have to release with a requirement for a nightly build (let alone a custom build). And from the perspective of a user who only uses stable builds, it makes no sense to go from 19.0 to 20.2 unless the minor version number is going to just never reset each year, which again ruins the point of having a major version number at all.

chief1983:
The minor would reset, just back to .2 instead of .0 with nightlies being .1.  Of course we could flip the even/odd thing and make 20.1 the stable and 20.0 the nightlies...but no one seemed keen on that.

MageKing17:
Yeah, that would be the part that makes no sense.

PR #2264, opened 2020-02-10
Fix build on big-endian architectures

pkubaj:
swap_bsp_data() and swap_sldc_data() use those.

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -5341,7 +5341,7 @@ void swap_bsp_sortnorms( polymodel * pm, ubyte * p )
 }
 #endif // BIG_ENDIAN
 
-void swap_bsp_data( polymodel *  /*pm*/, void * /*model_ptr*/ )
+void swap_bsp_data( polymodel * pm, void * model_ptr )
 {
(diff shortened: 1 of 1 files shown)


pkubaj:
Is `-Wunused-parameter` really required?

It seems to break one of checks.

chief1983:
I am wondering how related this is to the SLDC stuff that was investigated for ARM platform?  https://www.hard-light.net/forums/index.php?topic=96055.msg1890756#msg1890756

chief1983:
Also seems like the checks won't pass as-is since it is unused on other platforms.  I think there is a comment syntax to tell the compiler to ignored the unused params on those lines, that or can the params themselves be behind an ifdef as well?

CasparKielwein:
> Also seems like the checks won't pass as-is since it is unused on other platforms. I think there is a comment syntax to tell the compiler to ignored the unused params on those lines, that or can the params themselves be behind an ifdef as well?

Ont / the idiomatic way to mark an unused paramter is to "cast" it to void.
`(void)pm;`
`(void)model_ptr;`
and
`(void)buffer;`

Should do the trick.

CasparKielwein:
looks good to me

PR #2271, opened 2020-02-16
Actually fix shader compilation error

### LINKED ISSUE #2266: Librocket related shader issue
This bug shows up on the first 20.0 nightly. Does not appear in 19.0 final.

> Failed to load apng: (file single-player-o.png) animation duration <= 0.0f, bad data?
> Compiling new shader:
> 	libRocket UI
>    Loading built-in default shader for: rocketui-v.sdr
>    Loading built-in default shader for: rocketui-f.sdr
> Fragment shader failed to compile:
> 0(19) : error C7011: implicit cast from "int" to "bool"
> 
> OpenGL Debug: Source:OpenGL	Type:Error	ID:1282	Severity:High	Message:GL_INVALID_OPERATION error generated. Handle does not refer to the expected type of object (GL_PROGRAM_OBJECT_ARB)
> 
> Error: A shader failed to compile! Check the debug log for more information.
> File: gropenglshader.cpp
> Line: 748

I'm guessing it's related to this commit.. #2133 But I don't know for sure.

asarium:
And actually compile the shader this time to check if the changes
actually fixed the error.

This fixes #2266. Really! Pinky swear!

### DIFF
--- a/code/def_files/data/effects/rocketui-v.sdr
+++ b/code/def_files/data/effects/rocketui-v.sdr
@@ -10,7 +10,7 @@ layout (std140) uniform genericData {
 	mat4 modelViewMatrix;
 	mat4 projMatrix;
 
-	int textured;
+	bool textured;
 	int baseMapIndex;
 };
 


tomimaki:
I wonder if this line should be also changed.
https://github.com/scp-fs2open/fs2open.github.com/blob/90a97021f526deb513e87a24fdbd15f8858bee1c/code/graphics/util/uniform_structs.h#L269

asarium:
No that is intentional since `bool`s do not have a specified size. Those struct must have a very specific layout which can only be guaranteed by using basic types like `int`s.

tomimaki:
Ah, ok. Layman here. :p
Thanks for answer.

Issue #2270, opened 2020-02-16
Feature Request: Play 2nd arbitrary audio file at specific briefing stage


MjnMixael:
I'd like the ability to play a 2nd audio file at a specific briefing stage. The use-case here is to play a repeating "alarm" sound that will continue playing over/with the music for the entire rest of the briefing. That includes ship/weapon select as well. I know I could add the alarm sound to the briefing voice file, but that gets cut off as soon as the player switches to a new game_state whereas the music doesn't.

So at briefing stage N, play alarm.wav and keep playing it until the player returns to the mainhall menu or starts the mission.

Goober5000:
I think the most flexible way to handle this would be via scripting.  I can add a script hook that fires when a briefing stage appears, and the script could take care of playing the additional audio.

In the most recent update to Scroll, I use something very much like this, where the script selects a particular briefing track based on previous campaign progress.  That script triggers when the game state enters the briefing, but it would be very simple to change it to trigger when the briefing enters a particular stage.

MjnMixael:
So a script would have access to the specific briefing stage? I assume scripts can already play audio on top of audio and would be able to cut it off at the appropriate times (mission start, mission exit)?

Sounds good to me then.

Goober5000:
Exactly.  It's a fairly straightforward PR; I should have it by the end of the day.

Goober5000:
PR has been merged.  The new scripting hook, along with some modifications to `briefing_music-sct.tbm` in Scroll, should do the trick.  Note that `$On Briefing Stage:` will fire for both command briefings and briefings, but it can be restricted by using e.g. `$State: GS_STATE_BRIEFING`.

PR #2272, opened 2020-02-17
add scripting hook for briefing stage

Goober5000:
This is for feature request #2270.  It introduces a new hook that should fire for both command briefings and briefings.


Goober5000:
Here is the test table.
```
#Conditional Hooks
$Application: FS2_Open

$State: GS_STATE_BRIEFING
$On Briefing Stage:
[
    ba.warning("old stage = " .. tostring(hv.OldStage) .. "; new stage = " .. tostring(hv.NewStage))
]

#End
```

Goober5000:
Should be good to go now.  As far as I can determine, the parser didn't like the fact that I indented the debug command.

PR #2284, opened 2020-02-29
Make FFmpeg an optional requirement

asarium:
This does some refactoring to make it possible to cimpile and run FSO
without needing to link to FFmpeg. Since that eliminates support for
cutscenes or sound this is mostly useful for running a standalone server
which has no need for this anyway.

This reduces the image size of the standalone Docker image to only 20MB
and also resolves any licensing issues that might have existed.


chief1983:
Wish we had done this a long time ago haha, setting up ffmpeg on *nix distros is one of the hardest parts.

PR #2285, opened 2020-02-29
Fix standalone signal handling

asarium:
The standalone did not initialize SDL and so did not receive the quit
signal when it was sent SIGTERM or SIGINT. This fixes that by making the
standalone server just use the normal os initialization code.

Also includes some other code cleanup and some more logging to make it
more apparent in the log what is currently happening at standalone
startup.


chief1983:
Wait, I've had trouble killing standalones with ctrl+c forever, are you saying you fixed that?  Meant to file a bug ages ago but I run in gdb normally so ctrl+c still works there, just not when running them normally.

asarium:
Yep, I also couldn't kill my docker containers with Ctrl+C. With these changes I now can do that.

PR #2286, opened 2020-02-29
Fred2 version small fix

### LINKED ISSUE #2281: FRED2_Open improperly reporting version number?
It seems like recent versions of FRED2_Open are improperly identifying themselves as 3.8.1:
obraz

Goober5000:
I originally wanted to embed the version string in fred.rc, but after combing the documentation and stackoverflow for over an hour, I'm no closer to solving it that way.  Instead, I coded a quick hack to modify the title as part of window initialization.

This fixes #2281.

### DIFF
--- a/fred2/fred.rc
+++ b/fred2/fred.rc
@@ -2798,7 +2798,7 @@ END
 
 STRINGTABLE
 BEGIN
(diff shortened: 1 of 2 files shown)


MageKing17:
The "proper" method of doing this is to `#include` a header that's automatically generated.

Like, say, fred_rc.h.in, which is *supposed* to provide this functionality since I introduced it in 2017 (commit b1046b1c1c5d07db2e3dff9abd35820bccda8fbf), but apparently @Karajorma broke it in 2018 when he committed Eternal Variables (commit 9ac97f24223a7446d4364acfdb451c90697d6577).

PR #2294, opened 2020-03-06
Turn Weapon_info into a vector

Goober5000:
This converts the Weapon_info array into a vector.  This is a prerequisite for the `+remove` feature as it applies to weapons.  (Ship_info is already a vector.)

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1812,7 +1812,7 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 	parent_aip = &Ai_info[Ships[Objects[parent_objnum].instance].ai_index];
 	parent_ship = &Ships[Objects[parent_objnum].instance];
 	wip = get_turret_weapon_wip(&turret->weapons, weapon_num);
-	int turret_weapon_class = (wip - Weapon_info.data());
+	int turret_weapon_class = std::distance(Weapon_info.data(), wip);
 
 #ifndef NDEBUG
 	// moved here from check_ok_to_fire
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1812,7 +1812,7 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 	parent_aip = &Ai_info[Ships[Objects[parent_objnum].instance].ai_index];
 	parent_ship = &Ships[Objects[parent_objnum].instance];
 	wip = get_turret_weapon_wip(&turret->weapons, weapon_num);
-	int turret_weapon_class = static_cast<int>(std::distance(Weapon_info.data(), wip));
+	int turret_weapon_class = weapon_info_get_index(wip);
 
 #ifndef NDEBUG
 	// moved here from check_ok_to_fire
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1812,7 +1812,7 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 	parent_aip = &Ai_info[Ships[Objects[parent_objnum].instance].ai_index];
 	parent_ship = &Ships[Objects[parent_objnum].instance];
 	wip = get_turret_weapon_wip(&turret->weapons, weapon_num);
-	int turret_weapon_class = WEAPON_INFO_INDEX(wip);
+	int turret_weapon_class = (wip - Weapon_info.data());
 
 #ifndef NDEBUG
 	// moved here from check_ok_to_fire
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1812,7 +1812,7 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 	parent_aip = &Ai_info[Ships[Objects[parent_objnum].instance].ai_index];
 	parent_ship = &Ships[Objects[parent_objnum].instance];
 	wip = get_turret_weapon_wip(&turret->weapons, weapon_num);
-	int turret_weapon_class = std::distance(Weapon_info.data(), wip);
+	int turret_weapon_class = static_cast<int>(std::distance(Weapon_info.data(), wip));
 
 #ifndef NDEBUG
 	// moved here from check_ok_to_fire
--- a/code/hud/hudlock.cpp
+++ b/code/hud/hudlock.cpp
@@ -277,7 +277,7 @@ void hud_lock_reset(float lock_time_scale)
     
 	if ((swp->current_secondary_bank >= 0) && (swp->secondary_bank_weapons[swp->current_secondary_bank] >= 0)) {
 		Assert(swp->current_secondary_bank < MAX_SHIP_SECONDARY_BANKS);
-		Assert(swp->secondary_bank_weapons[swp->current_secondary_bank] < MAX_WEAPON_TYPES);
+		Assert(swp->secondary_bank_weapons[swp->current_secondary_bank] < static_cast<int>(Weapon_info.size()));
 		wip = &Weapon_info[swp->secondary_bank_weapons[swp->current_secondary_bank]];
 		Player->lock_time_to_target = i2fl(wip->min_lock_time*lock_time_scale);
 	} else {
--- a/code/mission/missioncampaign.cpp
+++ b/code/mission/missioncampaign.cpp
@@ -370,20 +370,10 @@ void mission_campaign_get_sw_info()
 {
     int i, count, ship_list[MAX_SHIP_CLASSES], weapon_list[MAX_WEAPON_TYPES];
 
-    // set allowable ships to the SIF_PLAYER_SHIPs
-    memset(Campaign.ships_allowed, 0, sizeof(Campaign.ships_allowed));
-    for (auto it = Ship_info.cbegin(); it != Ship_info.cend(); ++it) {
-        if (it->flags[Ship::Info_Flags::Player_ship])
-            Campaign.ships_allowed[std::distance(Ship_info.cbegin(), it)] = 1;
-    }
-
-    for (i = 0; i < MAX_WEAPON_TYPES; i++)
-        Campaign.weapons_allowed[i] = 1;
+	memset(Campaign.ships_allowed, 0, sizeof(Campaign.ships_allowed));
+	memset(Campaign.weapons_allowed, 0, sizeof(Campaign.weapons_allowed));
 
     if (optional_string("+Starting Ships:")) {
-        for (i = 0; i < static_cast<int>(Ship_info.size()); i++)
-            Campaign.ships_allowed[i] = 0;
-
         count = stuff_int_list(ship_list, MAX_SHIP_CLASSES, SHIP_INFO_TYPE);
 
         // now set the array elements stating which ships we are allowed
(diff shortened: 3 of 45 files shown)


code/mission/missioncampaign.cpp
```diff
@@ -370,39 +370,42 @@ void mission_campaign_get_sw_info()
...
+				Campaign.weapons_allowed[weapon_list[i]] = 1;
+		}
     }
+	else {
+		// set allowable weapons to the player-allowed ones
+		for (auto it = Weapon_info.cbegin(); it != Weapon_info.cend(); ++it) {
+			if (it->wi_flags[Weapon::Info_Flags::Player_allowed])
+				Campaign.weapons_allowed[std::distance(Weapon_info.cbegin(), it)] = 1;
+		}
+	}
```
MageKing17: Is this a behavior change?

code/weapon/swarm.cpp
```diff
@@ -505,7 +505,7 @@ void turret_swarm_set_up_info(int parent_objnum, ship_subsys *turret, weapon_inf
 	int bank_fired = swp->current_secondary_bank;
 
 	// initialize tsi
-	tsi->weapon_class = WEAPON_INFO_INDEX(wip);
+	tsi->weapon_class = static_cast<int>(wip - Weapon_info.data());
```
MageKing17: See previous comment about changing the macro.

code/ai/aiturret.cpp
```diff
@@ -1812,7 +1812,7 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum,
 	parent_aip = &Ai_info[Ships[Objects[parent_objnum].instance].ai_index];
 	parent_ship = &Ships[Objects[parent_objnum].instance];
 	wip = get_turret_weapon_wip(&turret->weapons, weapon_num);
-	int turret_weapon_class = WEAPON_INFO_INDEX(wip);
+	int turret_weapon_class = (wip - Weapon_info.data());
```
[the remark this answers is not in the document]
Goober5000: As I recall, that sort of pointer arithmetic is frowned upon, especially when one isn't dealing with arrays.  I wanted to remove it, and I successfully did in two places, but wasn't able to remove it in the other two places without additional refactoring.  So I just inlined it.

So, two questions: 1) is that sort of macro discouraged?  b) should the two uses be further refactored to remove the need for it?

code/ship/ship.cpp
```diff
@@ -1865,7 +1865,10 @@ static int parse_ship(const char *filename, bool replace)
 
 	rtn = parse_ship_values(sip, false, first_time, replace);
 
-	return rtn;	//0 for success
+	if (rtn == 0)	//0 for success
+		return ship_id;
+	else
+		return rtn;
```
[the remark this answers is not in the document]
Goober5000: I did this to make it work the same way as `parse_weapon`.

code/mission/missioncampaign.cpp
Goober5000: Not in practical effect.  The previous version allowed all weapons in the entire table, but the player would have only been able to use player-allowed weapons.  This new version allows all player-allowed weapons.

code/weapon/swarm.cpp
Goober5000: Ditto my other response.

code/ship/ship.cpp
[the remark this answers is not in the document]
MageKing17: But if the return value isn't used anywhere, why is that behavior change desirable or in-scope for this PR?

code/mission/missioncampaign.cpp
MageKing17: Since `weapons_allowed` gets written to the .CSG file, I'm pretty sure this *will* have an effect on in-development mods, although how much of one is another matter entirely. Since restricting the player's weapons to player-allowed ones happens elsewhere, though, I'm unsure of the need for this change.

code/ai/aiturret.cpp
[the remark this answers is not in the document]
MageKing17: I don't see the point in inlining a macro that does something discouraged; if anything, it just makes it harder to remove, because you have to find those instances by hand instead of tracking down where the macro is used.

As for that kind of pointer arithmetic being frowned upon... that's because you're encouraged to use `std::distance()` instead, to the best of my knowledge (at least I've seen people recommending `std::distance(some_vector.data(), pointer_to_elem)`, which may not be the same thing as the behavior actually being encouraged).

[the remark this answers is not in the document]
Goober5000: `std::distance()` is exactly what I would use if I had iterators, but these aren't iterators.

I'll take another look into refactoring this.  If that proves impractical, I'll just restore the macro.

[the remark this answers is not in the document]
MageKing17: A pointer is an iterator (by definition).

code/mission/missioncampaign.cpp
Goober5000: I doubt very much that any mods will be affected by this.  Even though both `+Starting Ships:` and `+Starting Weapons:` are optional when parsed, they are always written by the campaign editor save routine.  I'm certain that they are only optional because they were coded in the middle of FS1's development.  As such, this is cleanup to make the ship and weapon sections match.

code/ship/ship.cpp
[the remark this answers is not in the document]
Goober5000: I wanted to improve the consistency between parsing ships and weapons.  Most of the consistency updates (such as they are) will be in the +remove PR, but I made the change here because I was thinking about it at the time.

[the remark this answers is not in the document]
Goober5000: And no, the index isn't actually used, but if we ever needed the functionality in the future, it's a small step to add it now whereas it might be trickier to figure it out later.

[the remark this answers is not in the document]
MageKing17: Okay, but weapon parsing returns the index because countermeasure parsing *uses* the index; what, exactly, would you use the parsed ship index *for*?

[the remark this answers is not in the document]
Goober5000: That's what I mean by "if we ever needed the functionality in the future".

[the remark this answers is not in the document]
MageKing17: If we ever needed the functionality in the future, we could add it in the future (and quite simply); in terms of consistency, I'd much rather change weapon parsing to *not* need to return the index than superfluously add it to ship parsing.

code/ai/aiturret.cpp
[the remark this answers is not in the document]
Goober5000: Somehow I missed the second paragraph of your previous post.  It looks like `std::distance(Weapon_info.data(), wip)` does indeed work.

code/weapon/weapon.h
```diff
@@ -226,25 +226,30 @@ enum InFlightSoundType
 
 #define MAX_SUBSTITUTION_PATTERNS	10
 
-typedef struct weapon_info {
+class weapon_info
+{
+public:
```
asarium: Why did you turn the `struct` into a `class` if you make all members public anyway? It would be enough to just remove the typedef to make this a bit more modern.

code/weapon/weapon.h
```diff
@@ -538,7 +540,6 @@ extern int Num_player_weapon_precedence;				// Number of weapon types in Player_
 extern int Player_weapon_precedence[MAX_WEAPON_TYPES];	// Array of weapon types, precedence list for player weapon selection
 
 #define WEAPON_INDEX(wp)			(int)(wp-Weapons)
-#define WEAPON_INFO_INDEX(wip)		(int)(wip-Weapon_info)
```
asarium: I think it would be better to keep this (and convert it to a function) so that the weird logic of computing the offset is contained in that function instead of being scattered around in the code.

code/hud/hudlock.cpp
```diff
@@ -277,7 +277,7 @@ void hud_lock_reset(float lock_time_scale)
     
 	if ((swp->current_secondary_bank >= 0) && (swp->secondary_bank_weapons[swp->current_secondary_bank] >= 0)) {
 		Assert(swp->current_secondary_bank < MAX_SHIP_SECONDARY_BANKS);
-		Assert(swp->secondary_bank_weapons[swp->current_secondary_bank] < MAX_WEAPON_TYPES);
+		Assert(swp->secondary_bank_weapons[swp->current_secondary_bank] < static_cast<int>(Weapon_info.size()));
```
asarium: I don't like having this `static_cast<int>(Weapon_info.size())` used all over the code since it adds a lot of clutter. Maybe this could be replaced with a single function that is then called where appropriate?

code/ship/ship.cpp
```diff
@@ -12583,6 +12560,10 @@ int wing_lookup(const char *name)
  */
 static int ship_info_lookup_sub(const char *token)
 {
+	// bogus
+	if (token == nullptr)
```
asarium: If this is not a valid input then maybe it could be covered by an Assertion instead?

code/weapon/weapons.cpp
```diff
@@ -7534,6 +7512,33 @@ void validate_SSM_entries()
 	}
 }
 
+int weapon_get_random_usable_weapon()
```
asarium: Maybe call this `weapon_get_random_player_usable_weapon` to make it clearer what exactly this returns.

code/weapon/weapons.cpp
```diff
@@ -7534,6 +7512,33 @@ void validate_SSM_entries()
...
+{
+	int rand_wep;
+	int idx = 0, weapon_list[MAX_WEAPON_TYPES];
```
asarium: Maybe use a vector for `weapon_list` to reduce memory consumption and have compatibility for a future dynamic weapons limit.

code/weapon/weapon.h
Goober5000: Primarily because I think of classes as having member functions, and structs as not having member functions.  Since weapon_info has these, I thought this would make the definition clearer.

code/weapon/weapon.h
Goober5000: Ok.

code/hud/hudlock.cpp
Goober5000: I don't like it either, but we already have `static_cast<int>(Ship_info.size())` all over the code.  That's why I did this.  Of course, we could replace *both* uses...

code/ship/ship.cpp
Goober5000: I did this to match `weapon_info_lookup`.

code/weapon/weapons.cpp
Goober5000: Ok.

code/weapon/weapons.cpp
Goober5000: Ok.

code/weapon/weapon.h
asarium: For me at least, just having member functions does not mean that is has to be a struct. For me a struct should be used if the type contains a lot of public data members which is the case here.

code/hud/hudlock.cpp
asarium: We are going to need that more and more should we ever make the vector truly dynamic so I think it would be good to have a single function that returns the size as an int. Since this is about the weapons I would see doing the same for ships out of scope of this PR.

Goober5000: Why not approve this PR the way it is, and then I'll make a new PR specifically for removing the static_cast on both ships and weapons.

code/weapon/weapon.h
Goober5000: Ok, but we already have `player`, `object`, `ship`, `ship_subsys`, and more - all of which are classes with all-public members.

code/hud/hudlock.cpp
asarium: Works for me.

code/weapon/weapon.h
asarium: The point I am trying to make here is that you added new clutter to the code that is not necessary. Just because other types exist that also do this weird thing does not mean that it is ok to do so here.

Goober5000: I disagree that it is clutter; as I said earlier I thought this would make the definition clearer.  It also makes it consistent with the other places.  But in the interest of getting this approved I'll roll it back.

PR #2298, opened 2020-03-07
Add capability to have a pure lua script

asarium:
This adds the capability to have a Lua script that does not need to use
our table syntax at all but instead defines the hooks by calling API
functions. This also makes it possible to add hook functions at runtime
if the script requires that.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -14,6 +14,8 @@ misc-*,\
 -misc-static-assert,\
 -misc-unused-parameters,\
 -misc-non-private-member-variables-in-classes,\
+readability-*,\
+-readability-implicit-bool-conversion,\
 "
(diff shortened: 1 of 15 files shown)


asarium:
How did you arrive at that conclusion? This is a strict addition of features. Table based scripts will be unaffected.

MageKing17:
The removal of the ADE Debug code.

asarium:
That code has been superseded by the code using the Lua internat stack trace long ago. It was not used or useful anymore.

chief1983:
If they don't need to be fixed shouldn't we add the tidy-ignore comment to them at least?

Goober5000:
Nah.  I just looked at them and they are trivial.

PR #2299, opened 2020-03-07
As it turns out, the soundtrack reset issue is a bit more complicated…

### LINKED ISSUE #2296: In-Mission Music No Longer Plays
I found an oopsie. I can no longer get music to play in mission even in retail using build 20200306.

I suspect #2293 is the cause.

The-E:
… than I initially thought. The force_soundtrack check I previously removed was apparently vital; without it, calling event_music_level_init() (which in actuality is used to start soundtrack playback!) will result in the music being stopped, as the current soundtrack is set to -1.

This PR refactors the function's name to something more indicative of what it actually does, and removes a call made to it from within freespace.cpp which, due to changes in how the soundtrack system works, is now redundant and doesn't actually do anything anymore.

Closes #2296

### DIFF
--- a/code/gamesnd/eventmusic.cpp
+++ b/code/gamesnd/eventmusic.cpp
@@ -532,7 +532,7 @@ void event_music_do_frame()
 }
 
 // -------------------------------------------------------------------------------------------------
-// event_music_level_init() 
+// event_music_level_start() 
 //
 // Called at the start of a mission (level).  Sets up the pattern data, and kicks off the
 // first track to play().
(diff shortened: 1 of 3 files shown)


MageKing17 (approved):

Kerbrer:
In-mission music still seems to be borked. Every mission start after the first one, including restarts, requires me to go to the in-mission F2 menu to get the music to start playing.

wookieejedi:
Just noticed this as well, and was about to submit a bug report but found this one open.

PR #2301, opened 2020-03-14
add functions to remove the clutter of static casts for ship and weapon info vectors

Goober5000:
This is a follow-up to #2294.


Goober5000:
I'm open to changing the function name, or even the case.  Would `Ship_info_size()` and `Weapon_info_size()` be better?

asarium:
I think the current function name is good.

PR #2302, opened 2020-03-15
Xmt remove feature

Goober5000:
This adds the `+remove` tag to both ships and weapons, allowing modular tables to remove a previously parsed entry.  There's also a bit of cleanup and some Int3() removal.

~~Marking as draft because I still need to add the "blacklist" feature to prevent later-parsed TBMs from re-adding the removed entries.~~  Blacklist added.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -1780,8 +1780,8 @@ static void parse_ship(const char *filename, bool replace)
 	bool remove_ship = false;
 
 	required_string("$Name:");
-	stuff_string(fname, F_NAME, SHIP_MULTITEXT_LENGTH);
-	diag_printf("Ship name -- %s\n", buf);
+	stuff_string(fname, F_NAME, NAME_LENGTH);
+	diag_printf("Ship name -- %s\n", fname);
 
 	if (optional_string("+nocreate")) {
 		if (!replace) {
@@ -1818,7 +1818,7 @@ static void parse_ship(const char *filename, bool replace)
 		if (!skip_to_start_of_string_either("$Name:", "#End")) {
 			error_display(1, "Missing [#End] or [$Name] after ship class %s", fname);
 		}
-		return -1;
+		return;
 	}
 	
 	// an entry for this ship exists
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -5337,6 +5337,8 @@ void ship_init()
 			strcpy_s(default_player_ship, "");
 
 			//Parse main TBL first
+			Removed_ships.clear();
+			Ship_info.clear();
 			parse_shiptbl("ships.tbl");
 
 			//Then other ones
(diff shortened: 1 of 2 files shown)


code/ship/ship.cpp
```diff
@@ -1768,77 +1768,112 @@ ship_info::~ship_info()
...
 			Warning(LOCATION, "+nocreate flag used for ship in non-modular table");
 		}
 		create_if_not_found = false;
 	}
 
+	// check first if this is on the remove blacklist
+	auto it = std::find(Removed_ships.begin(), Removed_ships.end(), fname);
+	if (it != Removed_ships.end()) {
+		remove_ship = true;
+	}
```
CasparKielwein: you could simplify this with `remove_ship = std::any_of(Removed_ships.begin(), Removed_ships.end(), [](const auto& it) { return it == fname;});`

code/weapon/weapons.cpp
```diff
@@ -785,48 +790,83 @@ int parse_weapon(int subtype, bool replace, const char *filename)
...
+	//Check if weapon exists already
 	int w_id = weapon_info_lookup(fname);
 
-	if(w_id != -1)
+	// maybe remove it
+	if (remove_weapon)
+	{
+		if (w_id >= 0) {
+			mprintf(("Removing previously parsed weapon '%s'\n", fname));
+			Weapon_info.erase(Weapon_info.begin() + w_id);
```
CasparKielwein: That line scares me :)
maybe add an Assert, that `w_id < weapon_info.size()`

code/ship/ship.cpp
```diff
@@ -1768,77 +1768,112 @@ ship_info::~ship_info()
 	free_strings();
 }
 
+static SCP_vector<SCP_string> Removed_ships;
```
CasparKielwein: Removed_ships is only every pushed into, and never removed from.
On first sight, that looks like a memory problem.
Maybe add a comment, to explain how the size of the vector is limited (if it is).

code/ship/ship.cpp
Goober5000: I wouldn't say that's "simpler". :)  I prefer to keep it the way it is to be consistent with the rest of the code here.

code/weapon/weapons.cpp
Goober5000: That isn't needed because a) `weapon_info_lookup()` will never return a number that high, b) that is already enforced by range checks inside `std::vector.erase()`.

code/ship/ship.cpp
Goober5000: I'll add a comment.

Goober5000: Or something else that I think makes more sense.

code/weapon/weapons.cpp
CasparKielwein: uh, i didn't know about the check in erase.
thats nice

CasparKielwein:
Looks good to me.
I can't say anything about the actual logic though.

PR #2309, opened 2020-03-28
Make sure aip->net_signature always defaults to -1 for no target.

JohnAFernandez:
Although not used very often, ai_info.target_signature should be set to -1 whenever no target is specified, just like everywhere else in the code to make sure that the code is as predictable as possible.  It references the shipnum (afaict). Having it set to zero instead could potentially cause unwanted and unpredictable behavior, albeit in a small percentage of cases.

This seems to be a small oversight present since almost the beginning of the scripting system.

(Found while trying to figure out why the heck would the Object update packet send this ai info at all??)


Goober5000:
Whoa, this should not have been merged.  There are a lot more places in the code where 0 is used for `net_signature`.

PR #2315, opened 2020-04-06
Force subsystem targetting reticle to be square, even on non-square aspect ratios.

fjelliott:
This is a simple hack to fix an annoying problem on multi-monitor setups. As it turns out the subsystem targeting reticle is coded so it has the same width-to-height ratio as the screen the game is being played on. While on screens with a 16:9 or 4:3 aspect ratio this is fine because the reticle looks more or less square, on ultra-wide monitors or multi-monitor setups this is incredibly annoying, as shown in the following image where I'm targetting the engines on the GTCv Escher in one of the missions of the Derelict campaign while playing on my triple-monitor setup:
image

The fix resolves this problem, forcing the reticle to always be square by choosing the smallest of the reticle's width/height and assigning it to height/width. This gives far more desirable results:
image

Also, my changes work as intended when you get a lot closer to the target as well:
image

I don't think there's any need to disable this functionality for retail FS2, since I believe that having square subsystem reticles was always the intention, and it's very unlikely this would affect gameplay otherwise.

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -3368,6 +3368,9 @@ int subobj_find_2d_bound(float radius ,matrix * /*orient*/, vec3d * pos,int *x1,
 	t = (height*Canv_h2)/pnt.world.xyz.z;
 	h = t*Matrix_scale.xyz.y;
 
+	// Use the smaller of the width and height (forces the resulting targetting reticle to be square on non-square aspect ratios).
+	w = h = (w < h) ? w : h;
+
 	if (x1) *x1 = fl2i(pnt.screen.xyw.x - w);
 	if (y1) *y1 = fl2i(pnt.screen.xyw.y - h);
 


chief1983:
Any chance you confirmed that on smaller resolutions (basically as low as the engine can still go) this has no adverse effects?

fjelliott:
> Any chance you confirmed that on smaller resolutions (basically as low as the engine can still go) this has no adverse effects?

Is there a command line flag that can force FS2 to play in a really small resolution?

chief1983:
Shouldn't need to go lower than 1024x768, which I would assume is in the dropdown or whatever.

z64555:
try "-res 1024x768" or whatever smol reso you want to try

fjelliott:
It seems to be fine at smaller resolutions:
image

z64555:
For sake of argument, I think the original intent was to make the box 4:3, but that's a moot point at any rate.

This PR fixes the issue its designed for, which is to prevent excessive scaling on one axis due to "exotic" resolutions.  However I feel like there could be more done to the subsystem targeting box to make it more functional.

Most notably, the box should represent an area the player should aim within to hit it with primaries.  If I remember correctly, simple subsystem hitboxes are spheroids around a point defined in the model file.  Since the HUD uses rectangles, there's a bit of area around the corners where aiming will not score a hit on the subsystem.

I recommend investigating the hitbox and its targeting a bit further, if you feel up to it, and maybe make it more complete. :)

code/model/modelread.cpp
```diff
@@ -3368,6 +3368,9 @@ int subobj_find_2d_bound(float radius ,matrix * /*orient*/, vec3d * pos,int *x1,
 	t = (height*Canv_h2)/pnt.world.xyz.z;
 	h = t*Matrix_scale.xyz.y;
 
+	// Use the smaller of the width and height (forces the resulting targetting reticle to be square on non-square aspect ratios).
+	w = h = (w < h) ? w : h;
```
MageKing17, line 3372: Why not just use `MIN()`?

fjelliott: I didn't know you had a MIN() macro.

fjelliott:
> For sake of argument, I think the original intent was to make the box 4:3, but that's a moot point at any rate.
> 
> This PR fixes the issue its designed for, which is to prevent excessive scaling on one axis due to "exotic" resolutions. However I feel like there could be more done to the subsystem targeting box to make it more functional.
> 
> Most notably, the box should represent an area the player should aim within to hit it with primaries. If I remember correctly, simple subsystem hitboxes are spheroids around a point defined in the model file. Since the HUD uses rectangles, there's a bit of area around the corners where aiming will not score a hit on the subsystem.
> 
> I recommend investigating the hitbox and its targeting a bit further, if you feel up to it, and maybe make it more complete. :)

Well, the lead indicator is where you're supposed to shoot and that changes when you target a subsystem, so I think the subsystem targeting boxes are more to just give the player the general idea of where they are "under the hull". I do agree that making the subsystem box more accurately describe where the subsystem is would be a good thing, but it would involve considerably more work (picking a special point on the boundary of the subsystem radius, then projecting it onto the camera plane, then taking the distance from that to the subsystem's center point to get the width and height of the boundary box). However, generating the special point required for the operation would require the camera plane's normal vector, and I don't know where that is.

chief1983:
You also run into the problem that is fixed here, namely that getting too close would put the brackets so wide you can't see them.

chief1983:
That said, these brackets have always worked this way.  I'm a little curious if this change should only be available via hudgauges.tbl and not a change to default behavior, unless no one else cares about this changing default (retail) behavior.

z64555:
> Well, the lead indicator is where you're supposed to shoot and that changes when you target a subsystem, so I think the subsystem targeting boxes are more to just give the player the general idea of where they are "under the hull". I do agree that making the subsystem box more accurately describe where the subsystem is would be a good thing, but it would involve considerably more work (picking a special point on the boundary of the subsystem radius, then projecting it onto the camera plane, then taking the distance from that to the subsystem's center point to get the width and height of the boundary box). However, generating the special point required for the operation would require the camera plane's normal vector, and I don't know where that is.

eh? just make the width/height smaller so that the circle described by the given radius circumscribes the box.  This bit of code here already does the heavy lifting you mentioned, doesn't it?

fjelliott:
> eh? just make the width/height smaller so that the circle described by the given radius circumscribes the box. This bit of code here already does the heavy lifting you mentioned, doesn't it?

It tries to calculate the width and height of the bounding box in screen coordinates, but obviously doesn't do it correctly (otherwise this hack wouldn't be necessary). I would have to work out the math on this from scratch.

Also, I'm 99% certain that the failed build test is appveyor being stupid. Can someone override it? Or make it retry the test?

z64555:
> It tries to calculate the width and height of the bounding box in screen coordinates, but obviously doesn't do it correctly (otherwise this hack wouldn't be necessary). I would have to work out the math on this from scratch.

Its a conical projection using a simple calculation of ratios.  The calculation worked because at the time it was assumed the monitor being drawn on was a simple flat rectangle in front of the player's face, nothing fancy like multiple monitors or a curved screen, although it does some weirdness with the usage of the Matrix_scale vector.

z64555:
@chief1983 I don't this PR would significantly altar the style of the HUD, so I'm not expecting any objections to pop up to the change.

PR #2320, opened 2020-04-10
fix usage of system version of librocket

sandsmark:


lib/CMakeLists.txt
```diff
@@ -33,7 +33,16 @@ endif()
...
-include(libRocket.cmake)
+find_package(libRocket QUIET)
+if (${libRocket_FOUND})
+    message(STATUS "Using system installed version of libRocket")
+else()
+    message(STATUS "libRocket not found on system, using bundled")
+    IF(NOT IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/../cmake/external/rpavlik-cmake-modules/launcher-templates")
+        message(FATAL_ERROR "External submodules could not be found. Please make sure you have updated your submodules.")
+    endif()
+    include(libRocket.cmake)
```
asarium, line 44: The way we usually do this stuff is so have the `find_package` stuff confined to the cmake file that defines the target so this should be moved to libRocket.cmake.

code/CMakeLists.txt
```diff
@@ -41,7 +41,12 @@ target_link_libraries(code PUBLIC md5)
 
 target_link_libraries(code PUBLIC discord-rpc)
 
-target_link_libraries(code PUBLIC libRocket)
+find_package(libRocket QUIET)
```
asarium, line 44: To keep the "implementation" details concentrated to the site where the libRocket target is defined we usually define such libraries as interface targets with the same name as the one built from source. Then you could keep the same code here and still have it use the system version if it was found.

lib/CMakeLists.txt
```diff
@@ -33,7 +33,16 @@ endif()
 
 add_subdirectory(discord)
 
-include(libRocket.cmake)
+find_package(libRocket QUIET)
+if (${libRocket_FOUND})
```
asarium, line 37: There should be an option to explicitly specify which behaviour to use to make it consistent for CI builds. Since we require a specialized version of libRocket it would make sense to have the default be "build from source" and have a toggle with which the usage of the system libraries could be forced.

sandsmark:
> However, since someone could install that specialized version of libRocket separate I would be ok with this change if you add an option with which the usage of the system version could be enabled.

Since the "upstream" librocket seems pretty dead (their site that their repo points to has even been taken over), I kind of assumed your fork was the new "canonical" one. That's why I created https://aur.archlinux.org/packages/librocket-asarium-git/ (the patches from there are also in an open PR against your librocket repo, and opened a PR against the librocket/librocket repo too just in case they want to merge it).

lib/CMakeLists.txt
sandsmark: Didn't do this yet because it would generate a lot of noise in the diff here because of the indentation change (unless I'm missing a clever way to move it), but if the last version now is okay I can just move it over before a final version.

asarium: You don't need to indent everything if you just `return()` after your new code. Then the existing code does not need to be in an `else()` clause.

lib/CMakeLists.txt
```diff
@@ -33,7 +33,19 @@ endif()
...
+if (USE_BUNDLED_LIBROCKET)
+    IF(NOT IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/libRocket/Build/cmake")
+        message(FATAL_ERROR "libRocket submodule could not be found. Please make sure you have updated your submodules, or set USE_BUNDLED_LIBROCKET to use system version.")
+    endif()
+
+    include(libRocket.cmake)
+else(USE_BUNDLED_LIBROCKET)
+    find_package(libRocket REQUIRED)
+    add_library(libRocket INTERFACE)
+    target_link_libraries(libRocket INTERFACE "${libRocket_LIBRARIES}")
```
asarium, line 47: Not sure how that package is done but does it also have an include directory? Most CMake packages also have that so I maybe that also needs to be added here via `target_include_directories`.

Goober5000:
@sandsmark Any chance of taking another look at this?

JohnAFernandez:
@sandsmark feel free to reopen if you want to give this another try.  But we're closing for now.

PR #2322, opened 2020-04-12
Add extended scripting hook system

asarium:
This moves towards defining the hooks in the code module that executes
them. Also provides the API necessary for better documentation of hooks.
This only covers a few hooks as a test and demonstration of the new API.
If no issues are found with this it can be extended in the future.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,6 @@
 ---
 Language:        Cpp
+Standard: Cpp11
 # BasedOnStyle:  LLVM
 AccessModifierOffset: -2
 AlignAfterOpenBracket: Align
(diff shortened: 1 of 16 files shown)


.clang-format
```diff
@@ -18,8 +20,8 @@ AlwaysBreakAfterDefinitionReturnType: None
 AlwaysBreakAfterReturnType: None
 AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: Yes
-BinPackArguments: true
-BinPackParameters: true
+BinPackArguments: false
+BinPackParameters: false
```
Goober5000, line 24: Why are you including .clang-format changes along with your PR?

In any case I would not make the highlighted change. There are good reasons for using different argument lists in different situations.

asarium: I am including these changes since I am using clang-format for development and was not satisfied with our current configuration.
Since there currently is no enforcement of these formatting rules, everyone is still free to format their code as they like so these changes are not forcing anyone to format their code in a specific way.
The reason I made these changes is that I find a long argument list much easier to read if every argument is on it's own line.

Goober5000: The thing is, it is possible to have long argument lists with long variable/type names, or alternatively long argument lists with many arguments.  The first works well with each parameter on its own line, the second does not.

Issue #2314, opened 2020-04-05
FRED: Blank arguments in SEXPs will cause them to be removed during save, potentially causing broken events


AxemP:
So make an Event

Make the condition "is-destroyed-delay"

Go to the Delay argument and make it blank. Just delete the number and don't put anything in.

If you try and ask FRED to detect errors, it will give none. So now save the mission, and FRED will eventually bring up the error (it might take a save or two)

```
Error in mission event "Event name": Argument count is illegal
In sexpression: ( when 
   ( is-destroyed-delay "GTF Ulysses 0" ) 
   ( do-nothing ) 
)

(Error appears to be: is-destroyed-delay)
```

Not sure what the best preventative course of action is. Detect missing mandatory arguments and give them a default value? Make FRED bring this up before the save will eliminate the empty arguments?

Goober5000:
The argument isn't actually missing; it's just 0-length.  Of course, when the mission is re-parsed, FRED doesn't realize it's 0-length and thinks it's missing.

I think I have a solution.

Goober5000:
Fixed in #2323.

PR #2323, opened 2020-04-12
sanity check missing numeric sexp arguments

Goober5000:
In the SEXP parser, zero-length strings are valid, as are zero-length operators.  But zero-length numbers are interpreted as missing arguments, since their only syntactic separators are spaces and the parser skips over whitespace.  And in any case, zero-length numbers are not semantically valid anyway.  This PR adds a sanity check to intercept all zero-length numbers, as well as malformed numbers, and convert them to 0.

This fixes issue #2314.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -3887,7 +3887,7 @@ void stuff_sexp_text_string(SCP_string &dest, int node, int mode)
 	}
 	// not a variable
 	else {
-		char *ctext_string = CTEXT(node);
+		const char *ctext_string = CTEXT(node);
 
 		// strings are enclosed in quotes
 		if (Sexp_nodes[node].subtype == SEXP_ATOM_STRING) {
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -3831,7 +3831,7 @@ int num_block_variables()
 }
 
 /**
- * Stuff SEXP text string
+ * Stuff this particular SEXP node (just the node, not the tree) into a string representation
  */
 void stuff_sexp_text_string(SCP_string &dest, int node, int mode)
 {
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -3865,8 +3865,9 @@ void stuff_sexp_text_string(SCP_string &dest, int node, int mode)
 				Assert(mode == SEXP_SAVE_MODE);
 				sprintf(dest, "@%s[%s] ", Sexp_nodes[node].text, Sexp_variables[sexp_variables_index].text);
 			}
-		} else {
-			// string
+		}
+		// string
+		else {
```
asarium, line 3870: I don't like commenting on formatting but this style (else on the next line) does not fit with the code above so that should be made consistent.

code/parse/sexp.cpp
```diff
@@ -3883,12 +3884,26 @@ void stuff_sexp_text_string(SCP_string &dest, int node, int mode)
...
-		} else {
-			sprintf(dest, "%s ", CTEXT(node));
+			sprintf(dest, "\"%s\" ", ctext_string);
+		}
+		// numbers and operators are printed as-is
+		else {
+			// do some sanity checking based on Github issue #2314
+			if (Sexp_nodes[node].subtype == SEXP_ATOM_NUMBER) {
+				// if (for whatever reason) we have an empty string or an invalid number, print 0
+				if (!*ctext_string || !can_construe_as_integer(ctext_string)) {
```
asarium, line 3901: I would suggest something like this:
```suggestion
				if (*ctext_string == '\0' || !can_construe_as_integer(ctext_string)) {
```
With that it is easier to understand what this code is doing.

code/object/waypoint.cpp
```diff
@@ -497,7 +497,6 @@ void waypoint_remove(waypoint *wpt)
 					break;
 				}
 			}
-			Assert(ii != Waypoint_lists.end());
```
asarium, line 500: It appears that this (and the fredview changes) have slipped in from #2324.

code/parse/sexp.cpp
Goober5000: I wanted to make the minimal format change to clarify the comment, but...

code/parse/sexp.cpp
Goober5000: Ok

code/object/waypoint.cpp
Goober5000: I saw that.  I rebased before I committed and I thought that fixed it, but I guess not.  I'll rebase to master once 2324 is merged.

Goober5000:
Ok, branch rebased.

PR #2324, opened 2020-04-13
Fix Mantis 3171

Goober5000:
See Mantis 3171:
http://scp.indiegames.us/mantis/view.php?id=3171

This PR properly resets waypoint lists as-needed so that multiple lists can be copied.

This also removes two Asserts that never worked and are not needed anyway.

### DIFF
--- a/fred2/fredview.cpp
+++ b/fred2/fredview.cpp
@@ -526,9 +526,11 @@ int drag_objects()
 		Dup_drag = 0;
 
 	if (Dup_drag == 1) {
-		dup_object(NULL);  // reset waypoint list
 		cobj = Duped_wing = -1;
 		flag = 0;
+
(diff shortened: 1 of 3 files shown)


fred2/fredview.cpp
```diff
@@ -526,9 +526,11 @@ int drag_objects()
 		Dup_drag = 0;
 
 	if (Dup_drag == 1) {
-		dup_object(NULL);  // reset waypoint list
 		cobj = Duped_wing = -1;
 		flag = 0;
+
+		int duping_waypoint_list = -1;
```
asarium, line 532: Please also replicate these changes for qtFRED in `EditorViewport::drag_objects`.

Goober5000: Ok.

PR #2325, opened 2020-04-13
prevent player leaving the loadout screen with incomplete weapons

Goober5000:
This is the best way to address Mantis 2715.
http://scp.indiegames.us/mantis/view.php?id=2715

I took a look at allowing empty banks in the course of working on PR #2318, and it was completely intractable.  But in order to prevent certain "funnies" such as violation of bank restrictions, it's a good idea to ensure all banks are populated.

This PR adds a new XSTR for a popup message, and incidentally adds another XSTR to a message that should have had one already but didn't.

### DIFF
--- a/code/missionui/missionshipchoice.cpp
+++ b/code/missionui/missionshipchoice.cpp
@@ -1860,7 +1860,7 @@ bool is_a_weapon_slot_empty()
 		// a ship must exist in this slot
 		if (Wss_slots[slot].ship_class >= 0)
 		{
-			for (int bank = 0; bank < MAX_SHIP_WEAPONS; bank++)
+			for (int bank = 0; bank < MAX_SHIP_WEAPONS; bank++)		// NOLINT(modernize-loop-convert)
 			{
 				// is there a weapon here?
 				if (Wss_slots[slot].wep_count[bank] <= 0)
--- a/code/missionui/missionshipchoice.cpp
+++ b/code/missionui/missionshipchoice.cpp
@@ -1853,6 +1853,25 @@ bool is_weapon_carried(int weapon_index)
 	return false;
 }
 
+bool is_a_weapon_slot_empty()
+{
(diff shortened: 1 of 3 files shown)


Goober5000:
That clang-tidy suggestion a) is wrong, and b) would hinder clarity of the function, especially when compared to its adjacent counterpart, so I am leaving the loop as-is.

code/missionui/missionshipchoice.cpp
```diff
@@ -1853,6 +1853,25 @@ bool is_weapon_carried(int weapon_index)
...
 }
 
+bool is_a_weapon_slot_empty()
+{
+	for (int slot = 0; slot < MAX_WING_BLOCKS*MAX_WING_SLOTS; slot++)
+	{
+		// a ship must exist in this slot
+		if (Wss_slots[slot].ship_class >= 0)
+		{
+			for (int bank = 0; bank < MAX_SHIP_WEAPONS; bank++)
```
asarium, line 1863: If the suggested change is not desired then please suppress it with `// NOLINT(modernize-loop-convert)`.

code/missionui/missionweaponchoice.cpp
```diff
@@ -4097,7 +4097,7 @@ void wl_apply_current_loadout_to_all_ships_in_current_wing()
 	// display error messages
 	if (error_flag)
 	{
-		SCP_string full_error_message = "The following errors were encountered:\n";
+		SCP_string full_error_message = XSTR("The following errors were encountered:\n", 1641);
```
asarium, line 4100: Please also increment `XSTR_SIZE` to match the new string.

code/missionui/missionshipchoice.cpp
Goober5000: Ok.

code/missionui/missionweaponchoice.cpp
Goober5000: I forgot about that, thanks.

PR #2326, opened 2020-04-13
Cleaned up code that renders subsystem target boxes. (Follow-up to #2315)

fjelliott:
A follow-up to #2315. I Verifyed that the math of the old calculations is valid, except it shouldn't calculate both the width and height of the bounding box separately (it should only calculate height and use that for both). Then I cleaned up that function a little bit, reducing the number of local variables and the number of floating-point operations.

I have tested these changes on regular resolutions (1024 x 768):
image
image

As well as exotic resolutions (5760 x 1080):
image
image

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -3340,12 +3340,15 @@ int submodel_find_2d_bound_min(int model_num,int submodel, matrix *orient, vec3d
  *
  * Note that x1,y1,x2,y2 aren't clipped to 2D screen coordinates.
  *
+ * Calculates the focal length of the camera, and uses the law of similar
+ * triangles to project the subsystem's radius to the screen.
+ *
  * @return zero if x1,y1,x2,y2 are valid
  * @return 2 for point offscreen
  */
 int subobj_find_2d_bound(float radius ,matrix * /*orient*/, vec3d * pos,int *x1, int *y1, int *x2, int *y2 )
 {
-	float w,h;
+	float w,h,focal_length;
 	vertex pnt;
 
 	g3_rotate_vertex(&pnt,pos);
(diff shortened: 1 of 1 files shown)


code/model/modelread.cpp
```diff
@@ -3362,14 +3359,7 @@ int subobj_find_2d_bound(float radius ,matrix * /*orient*/, vec3d * pos,int *x1,
...
 
-	t = (width * Canv_w2)/pnt.world.xyz.z;
-	w = t*Matrix_scale.xyz.x;
-
-	t = (height*Canv_h2)/pnt.world.xyz.z;
-	h = t*Matrix_scale.xyz.y;
-
-	// Use the smaller of the width and height (forces the resulting targetting reticle to be square on non-square aspect ratios).
-	w = h = (w < h) ? w : h;
+	w = h = radius * Canv_h2 * Matrix_scale.xyz.y / pnt.world.xyz.z;
```
z64555, line 3362: I recommend breaking this up into a few more lines to show that we're operating on the height component first, and then copying over to the width.  Also recommend making the Matrix_scale a seperate operation from the projection calculation to make it clearer as to whats going on.

code/model/modelread.cpp
```diff
@@ -3345,12 +3345,9 @@ int submodel_find_2d_bound_min(int model_num,int submodel, matrix *orient, vec3d
  */
 int subobj_find_2d_bound(float radius ,matrix * /*orient*/, vec3d * pos,int *x1, int *y1, int *x2, int *y2 )
```
z64555, line 3346: Please add a comment in this function's doxygen description (@details) to state how the function is working.  Something along the lines of "Uses a simple planar projection of the subsystem's radius to draw a targeting box around it" should suffice

code/model/modelread.cpp
fjelliott: The matrix scale is part of the projection. However, I broke it up a bit to aid in clarity. Any good code optimizing compiler should be able to reduce it to the same number of instructions anyway.

code/model/modelread.cpp
fjelliott: The function operates by calculating the focal length of the camera, then using the law of similar triangles to project the radius of the subsystem to the screen. I added a comment to this affect in the function documentation as per your request.

z64555:
👍

PR #2332, opened 2020-04-14
add an ai-profiles flag to govern the Mantis 3147 fix

Goober5000:
See Mantis 3147:
http://scp.indiegames.us/mantis/view.php?id=3147

I was awaiting a response, and then a response never happened. :)  So I implemented the flag myself.  The new flag should enable the fixed behavior when active, and default to retail behavior when inactive.

The original changes were committed in SVN revision 11284.  Here is the diff from that revision:

```
Index: aicode.cpp
===================================================================
--- aicode.cpp	(revision 11283)
+++ aicode.cpp	(revision 11284)
@@ -5644,7 +5644,7 @@
 	float	dot;
 	vec3d	v2t;
 
-	if (!( vm_vec_mag_quick(&G_predicted_pos) < AICODE_SMALL_MAGNITUDE )) {
+	if ( !(vm_vec_mag_quick(&G_predicted_pos) < AICODE_SMALL_MAGNITUDE ) && !(aip->submode == SM_AVOID) ) {
 		if ( vm_vec_cmp(&G_predicted_pos, &G_fire_pos) ) {
 			vm_vec_normalized_dir(&v2t, &G_predicted_pos, &G_fire_pos);
 			dot = vm_vec_dot(&v2t, &objp->orient.vec.fvec);
@@ -8238,8 +8238,12 @@
 
 	//	If seeking lock, try to point directly at ship, else predict position so lasers can hit it.
 	//	If just acquired target, or target is not in reasonable cone, don't refine believed enemy position.
-	if ((real_dot_to_enemy < 0.25f) || (aip->target_time < 1.0f) || (aip->ai_flags & AIF_SEEK_LOCK)) {
+	if ((real_dot_to_enemy < 0.25f) || (aip->target_time < 1.0f)) {
 		predicted_enemy_pos = enemy_pos;
+	} else if (aip->ai_flags & AIF_SEEK_LOCK) {
+		set_predicted_enemy_pos(&predicted_enemy_pos, Pl_objp, &aip->last_aim_enemy_pos, &aip->last_aim_enemy_vel, aip);	// Set G_fire_pos
+		predicted_enemy_pos = enemy_pos;
+		G_predicted_pos = predicted_enemy_pos;
 	} else {
 		//	Set predicted_enemy_pos.
 		//	See if attacking a subsystem.
```

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -627,8 +627,8 @@ void ai_profile_t::reset()
         detail_distance_mult[i] = 0;
     }
 
-	// via Github #2332, enable bugfixes if we are targeting > 19.0
-	if (Targetted_version > gameversion::version(19)) {
+	// via Github #2332, enable bugfixes if we are targeting version 20 and up
+	if (mod_supports_version(20, 0, 0)) {
 		flags.set(AI::Profile_Flags::Huge_turret_weapons_ignore_bombs);
 		flags.set(AI::Profile_Flags::Fix_linked_primary_bug);
 		flags.set(AI::Profile_Flags::Prevent_targeting_bombs_beyond_range);
@@ -641,7 +641,7 @@ void ai_profile_t::reset()
 		flags.set(AI::Profile_Flags::Use_actual_primary_range);
 	}
 	// this flag has been enabled ever since 3.7.2
-	if (Targetted_version >= gameversion::version(3, 7, 2)) {
+	if (mod_supports_version(3, 7, 2)) {
 		flags.set(AI::Profile_Flags::Fix_ramming_stationary_targets_bug);
 	}
 }
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -7,11 +7,11 @@
 
 
 
-#include "globalincs/pstypes.h"
-#include "def_files/def_files.h"
 #include "ai/ai_profiles.h"
+#include "def_files/def_files.h"
 #include "globalincs/pstypes.h"
 #include "localization/localize.h"
+#include "mod_table/mod_table.h"
 #include "parse/parselo.h"
 #include "ship/ship.h"
 #include "weapon/weapon.h"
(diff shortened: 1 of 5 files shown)


The-E:
My main question here is, does this actually address anything? Are there actual missions or campaigns out there that are unplayable or less playable because of this bugfix? In other words, while it is certainly possible to construct scenarios where the outcome of a mission hinges on the behaviour here, it is imho easier and cleaner to address this at the mission design stage rather than trying to do a code change. After all, this change has been live for 5 years, and I can't remember any sudden drops or increases in mission difficulty being reported during that time that could be attributed to this that weren't fixable by mission designers.

Goober5000:
I'll repost what I said on Discord:
> It is very, very difficult to pinpoint balance changes to exact causes.  There are some Scroll missions that were originally written in 2002-2004 that I had to rebalance before release because they played differently [the reason was never determined].  In the absence of anything conclusive, my policy is to align the code to retail behavior whenever a discrepancy is discovered.
>
> As for why it took five years to get back to this, well, Mantis these days is out of sight, out of mind
>
> Part of me wants to leave the code as-is for the exact reasons niffiwan and The E described, but I don't like leaving the possibility that this may unknowingly affect some mod's gameplay

There are several reasons this should be merged:
* We have a policy of matching retail behavior
* Yes, the fix has been live for 5 years, but the original behavior was live for 17 years
* Every other identified AI fix in the codebase is governed by an ai-profiles flag; this is the only known exception

To directly answer your question, I'm not aware of a specific mission harmed by the fix, but IMO that's the wrong question to ask.

asarium:
Could we use out "Target engine version" feature to have a sane default value based on the engine version the mod supports?

Goober5000:
That's an interesting idea.  How would you propose to do that -- have the flag activated for targets from 2015-2020, but deactivated for 2020+?

asarium:
Since this appears to be a bug fix flag I would say it should be active by default at least from the current version forward. Since this has already been like this for some time the minimum version to activate it could be backdated as well (e.g. 3.8).

Goober5000:
Ok, suppose I were to add this code inside `ai_profile_t::reset()`:

```
	// this flag has been enabled ever since 3.7.2
	if (Targetted_version >= gameversion::version(3, 7, 2)) {
		flags.set(AI::Profile_Flags::Fix_ramming_stationary_targets_bug);
	}
```

And then I'd add code to enable the other bugfix flags for a targeted version above 19.0.

wookieejedi:
Does Goober's last suggestion work for you @asarium?

Goober5000:
I have added the aforementioned flag-enabling code.

code/ai/ai_profiles.cpp
```diff
@@ -624,4 +626,22 @@ void ai_profile_t::reset()
     for (int i = 0; i <= MAX_DETAIL_LEVEL; ++i) {
         detail_distance_mult[i] = 0;
     }
+
+	// via Github #2332, enable bugfixes if we are targeting > 19.0
+	if (Targetted_version > gameversion::version(19)) {
```
asarium, line 631: Maybe you could use `mod_supports_version` here. It's pretty much the same except for doing a `>=` so you would use version 20 here to make it use the old behaviour for 19.0.

Goober5000: The thing is, that would mean test builds and release candidates wouldn't have the benefit of using those flags.  I could fudge it with 19.0.1, but that seems less elegant.

Goober5000: That being said, in this particular instance we moved very quickly to 20.x.  Also, it's `mod_supports_version`, not the *actual* version as used by version-specific commenting.

PR #2341, opened 2020-04-27
Adds Scaling for Target Box Models Separate from Techroom

wookieejedi:
Currently `$Closeup_Pos:` and `$Closeup_Zoom:` control the camera zoom and position values for the Techroom, briefing, and HUD Target Box. Unfortunately, with very large long ships this results in an issue, to fit the view of an entire big ship in the HUD Target Box the zoom and position must be set a certain way which results in the ship looking very far away when viewed in the Techroom. This PR adds two new fields, `$Closeup_pos_targetbox:` and `$Closeup_Zoom_targetbox:`, that control camera position and zoom but only for the HUD Target Box. Now modders can set values for both and have the ships look properly zoomed in both the Techroom and HUD Target Box. The new HUD Target Box values are optional, and if not specified they default to the values for `$Closeup_Pos:` and `$Closeup_Zoom:` . I have tested and everything works as expected.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -3679,9 +3679,9 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 	{
 		stuff_vec3d(&sip->closeup_pos_targetbox);
 	} 
-	else 
+	else if (first_time)
 	{ 
-		vm_vec_copy_scale(&sip->closeup_pos_targetbox, &sip->closeup_pos, 1.0f);
+		sip->closeup_pos_targetbox = sip->closeup_pos;
 	}
 
 	if (optional_string("$Closeup_zoom_targetbox:")) {
@@ -3692,7 +3692,7 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 			sip->closeup_zoom_targetbox = 0.5f;
 		}
 	}
-	else
+	else if (first_time)
 	{
 		sip->closeup_zoom_targetbox = sip->closeup_zoom;
 	}
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -945,6 +945,9 @@ void ship_info::clone(const ship_info& other)
 	closeup_pos = other.closeup_pos;
 	closeup_zoom = other.closeup_zoom;
 
+	closeup_pos_targetbox = other.closeup_pos_targetbox;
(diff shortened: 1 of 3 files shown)


code/ship/ship.cpp
```diff
@@ -3665,6 +3674,28 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
...
 		}
 	}
+
+	if(optional_string("$Closeup_pos_targetbox:"))
+	{
+		stuff_vec3d(&sip->closeup_pos_targetbox);
+	} 
+	else 
+	{ 
+		vm_vec_copy_scale(&sip->closeup_pos_targetbox, &sip->closeup_pos, 1.0f);
```
Goober5000, line 3684: Why not copy the vector using assignment, instead of copying and scaling by 1.0?

code/ship/ship.cpp
```diff
@@ -3665,6 +3674,28 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 			sip->closeup_zoom = 0.5f;
 		}
 	}
+
+	if(optional_string("$Closeup_pos_targetbox:"))
+	{
+		stuff_vec3d(&sip->closeup_pos_targetbox);
+	} 
+	else
```
Goober5000, line 3682: You need to use `else if (first_time)` here.  Otherwise, if you have a ships.tbl and a modular table, and that modular table doesn't have a `$Closeup_pos_targetbox:` entry, it will always be reset to the default.  Take a look at `$Closeup_pos:` above.

code/ship/ship.cpp
```diff
@@ -3665,6 +3674,28 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
...
+
+	if (optional_string("$Closeup_zoom_targetbox:")) {
+		stuff_float(&sip->closeup_zoom_targetbox);
+
+		if (sip->closeup_zoom_targetbox <= 0.0f) {
+			mprintf(("Warning!  Ship '%s' has a $Closeup_zoom_targetbox value that is less than or equal to 0 (%f). Setting to default value.\n", sip->name, sip->closeup_zoom_targetbox));
+			sip->closeup_zoom_targetbox = 0.5f;
+		}
+	}
+	else
```
Goober5000, line 3695: Same deal with `else if` here.

PR #2343, opened 2020-04-28
implement XSTR check as described in Mantis 2942

Goober5000:
Certain mods provide their own tstrings.tbl, but this runs into trouble when you localize data meant for one mod using a tstrings.tbl meant for another mod.  See the Mantis ticket:
http://scp.indiegames.us/mantis/view.php?id=2942
and also the HLP post:
http://www.hard-light.net/forums/index.php?topic=85041.msg1716600#msg1716600

This PR adds an optional flag that will check the untranslated string against the "default language" translated string.  The two strings should be identical.  If they aren't, the code assumes there is a tstrings.tbl mismatch.

I modified the mission listing code to enable this check when localizing mission info (titles and descriptions), and to avoid adding the mission to the list if there is a mismatch.  Fortunately, I think mission listing is the only place where this bug is likely to happen.

Note also that the localization code never actually used to translate strings if the current language was the default language.  This PR changes this to always use the tabled version of the string, which is probably the behavior people expect.

### DIFF
--- a/code/localization/localize.cpp
+++ b/code/localization/localize.cpp
@@ -815,7 +815,7 @@ bool lcl_ext_localize_sub(const SCP_string &in, SCP_string &text_str, SCP_string
 // Goober5000 - wrapper for lcl_ext_localize_sub; used because lcl_replace_stuff has to
 // be called *after* the translation is done, and the original function returned in so
 // many places that it would be messy to call lcl_replace_stuff everywhere
-// Addendum: Now, of course, it provides a handy way to encapsulate another check
+// Addendum: Now, of course, it provides a handy way to encapsulate the unexpected tstring check.
 void lcl_ext_localize(const char *in, char *out, size_t max_len, int *id)
 {
 	// buffer for the untranslated string inside the XSTR tag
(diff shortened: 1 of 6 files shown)


code/localization/localize.cpp
```diff
@@ -802,15 +808,43 @@ void lcl_ext_localize_sub(const SCP_string &in, SCP_string &out, int *id)
...
 		*id = str_id;
 	}
+
+	return true;
 }
 
 // Goober5000 - wrapper for lcl_ext_localize_sub; used because lcl_replace_stuff has to
 // be called *after* the translation is done, and the original function returned in so
 // many places that it would be messy to call lcl_replace_stuff everywhere
+// Addendum: Now, of course, it provides a handy way to encapsulate another check
```
asarium, line 818: Please document which check you mean.

code/menuui/readyroom.cpp
```diff
@@ -999,6 +1008,7 @@ void sim_room_init()
 	int i;
 	sim_room_buttons *b;
 	char wild_card[256];
+	bool dummy_buffer;
```
asarium, line 1011: Can you move this to the `#ifdef` block where this is needed? Then it will not appear in release builds.

Goober5000: Good point.

code/localization/localize.cpp
Goober5000: Ok.

PR #2346, opened 2020-04-28
Set_primary_weapon_linkage cleanup as requested by Goober5000

EatThePath:
Originally requested in https://github.com/scp-fs2open/fs2open.github.com/pull/2328#discussion_r415157575, but subsequent progress on that PR may mean the branch may not touch aicode.cpp at all, so it seemed preferable to pare that cleanup out to a separate PR.


Goober5000:
Good.

Copying from my other comment:
> The point of this change is to move the `MAX_WEAPONS * 0.75f` check to the beginning of the function, where it belongs (and was originally in the public code release), and to decouple it from the flag check.

Issue #1851, opened 2018-09-07
game_settings.tbl proposal: $Disable Dual Secondary Fire


WoolieWool:
This proposal would involve a new flag to game_settings.tbl that would disable the dual secondary linking feature in retail FreeSpace 2, to enable mod authors both to break with established FS convention in yet another way, and to balance missiles around the assumption that they can only be fired singly rather than in pairs (fewer missiles, higher damage).

NovachenFS2:
I actually think, that such feature would make more sense as a weapons.tbl flag... because there is one for Primary Weapons already?

EatThePath:
I could definitely use this. It's a good knob to let mods be more distinct from one another or make missiles more distinct within a mod, and will be even moreso when multi-lock is mainline

FSCyborg:
Officially claiming this one. Just realized my last enhancement is not in because I missed that asarium requested more changes... >.>

MageKing17:
Implemented in #2347.

PR #2347, opened 2020-04-28
Disable doublefire take 2

EatThePath:
Implements the requested features in #1851

Doublefiring can be turned off for AIs and/or players separately through AI profiles flags. Went with that instead of gamesettings largely because that seems consistent with other similar behavior options, and I can conceivably see someone wanting to vary it between missions.

It can also be turned off on a per-weapon basis through a new weapon flag.

Originally attempted in PR #2328, but the feedback there led to enough course corrections that I felt a clean slate was for the better.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -11951,8 +11951,11 @@ int ship_fire_secondary( object *obj, int allow_swarm )
 			goto done_secondary;
 		}
 
-		int start_slot, end_slot;
-
+		// Handle the optional disabling of dual fire
+		// dual fire/doublefire can be disabled for individual weapons, for players, or for AIs
+		// if any of these apply to the current weapon, unset the dual fire flag on the ship 
+		// then proceed as normal.
+		// This is only handled at firing time so dualfire isn't lost when cycling through weapons
 		if (shipp->flags[Ship_Flags::Secondary_dual_fire] &&
 			( wip->wi_flags[Weapon::Info_Flags::No_doublefire] || 
 			( The_mission.ai_profile->flags[AI::Profile_Flags::Disable_ai_secondary_doublefire] && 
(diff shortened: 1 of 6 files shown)


Goober5000:
I haven't forgotten about this.  I'll take a look soon.

code/ship/ship.cpp
```diff
@@ -11952,6 +11953,16 @@ int ship_fire_secondary( object *obj, int allow_swarm )
 
 		int start_slot, end_slot;
```
Goober5000, line 11954: Put these either at the top of the block (where `num_slots` is) or below your new addition (because the next block is the code it was next to originally, and it should still be associated with it).

code/ship/ship.cpp
```diff
@@ -11952,6 +11953,16 @@ int ship_fire_secondary( object *obj, int allow_swarm )
...
 
+		if (shipp->flags[Ship_Flags::Secondary_dual_fire] &&
+			( wip->wi_flags[Weapon::Info_Flags::No_doublefire] || 
+			( The_mission.ai_profile->flags[AI::Profile_Flags::Disable_ai_secondary_doublefire] && 
+				shipp->objnum != OBJ_INDEX(Player_obj) ) ||
+			( The_mission.ai_profile->flags[AI::Profile_Flags::Disable_player_secondary_doublefire] &&
+				shipp->objnum == OBJ_INDEX(Player_obj) ))
+			) {
+			shipp->flags.remove(Ship_Flags::Secondary_dual_fire);
+		}
```
Goober5000, line 11964: Add a comment here to describe the no-dual-secondary feature, since this is really the meat of it.

code/ship/ship.cpp
EatThePath: If moving down, ought num_slots declaration and assignment be moved down below my addition as well? It's not used until that section.

Goober5000: If you want to, but `num_slots` is not part of the scope of this PR.  What I wanted to point out was that, in adding the new if() block, you separated two pieces of code that were intended to be associated together.  Be careful about that.

EatThePath: Reasonable, I'll keep my changes local then.

EatThePath:
I redid some testing after this last round of changes and missiles behave as expected in every case I can think to test.

PR #2348, opened 2020-04-29
bring save format up to date (Mantis 2867)

Goober5000:
See: http://scp.indiegames.us/mantis/view.php?id=2867

FSO-specific features should now all be properly marked off with checks for retail format.

This PR also simplifies the `Format_fs2_open`, `Format_fs2_retail`, and `Format_fs1_retail` flags into the single `Mission_save_format` variable.


JohnAFernandez:
You've got some whitespace changes that snuck in here.

Goober5000:
Those were intentional.  I thought I'd fix the braces at the same time I did the other changes.  I didn't change all braces, just the ones enclosing functions.

PR #2349, opened 2020-04-29
allow empty weapons on locked ships

Goober5000:
As noticed in Discord, in loop1-2 some of your wingmen have missing weapons, but as the ships are locked it's impossible for the player to modify them.  These ships should not be part of the check.

### DIFF
--- a/code/missionui/missionbrief.cpp
+++ b/code/missionui/missionbrief.cpp
@@ -1992,7 +1992,7 @@ int brief_only_allow_briefing()
 		return 1;
 	}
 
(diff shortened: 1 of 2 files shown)


z64555:
LGTM, but I'm not familiar enough with the flag combinators to really approve that style change.

It may be good for a debug printout/warning about the empty slots.

Goober5000:
There is already a warning about empty slots; that's what the "At least one ship has an empty weapon bank." popup is for.  If you mean a warning about locked ships, that's something the FREDder has to explicitly set.

PR #2361, opened 2020-05-04
fred2 variable ui tweak

Goober5000:
The two variable types are mutually exclusive, so they should be radio buttons, not check boxes.


chief1983:
But is one or the other required?  Should one then be checked by default?  If neither is required, how can one clear it after one is checked?

chief1983:
Also I'm not really familiar with how these inputs work in FRED (I only know these controls from HTML really), but what ties them together now so that checking one would uncheck the other?  Doesn't a radio button usually save to the same variable, but checkboxes are each tied to their own variable?

Goober5000:
I believe radio buttons can be grouped so that MFC takes care of making sure only one is selected, but that isn't needed here because the behavior is done through the code.

That being said, it is necessary for the checkboxes to be *deselected* if desired, and that won't be possible if they are changed to radio buttons.  So I'm withdrawing this PR.

PR #2367, opened 2020-05-05
fix some code rot

Goober5000:
I was comparing the retail and FSO versions of `ai_fire_from_turret()` in hopes of resolving Mantis 804:
http://scp.indiegames.us/mantis/view.php?id=804

Function excerpts here:
[fso.txt](https://github.com/scp-fs2open/fs2open.github.com/files/4578819/fso.txt)
[retail.txt](https://github.com/scp-fs2open/fs2open.github.com/files/4578820/retail.txt)

There has been a lot of code drift, but on a preliminary inspection the code appears to be correct, with the notable exception of this little bit.  In SVN revision 5399 KeldorKatarn submitted a "fix" for the fire-down-normals flag that made turrets always fire directly at the enemy, instead of in the direction the turrets were facing.  This PR puts it back the way it's supposed to be.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -2580,7 +2580,9 @@ void ai_fire_from_turret(ship *shipp, ship_subsys *ss, int parent_objnum)
 				ss->turret_next_fire_pos = ffp_pos;
 			}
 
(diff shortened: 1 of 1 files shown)


Goober5000:
This was the result of at least four levels of code drift, but only one level actually introduced the bug.  This may also explain some of the balance issues I noticed in Scroll.  And it also points to the need for a more careful examination of the differences in this function between SCP in retail, so I would welcome additional eyes on the problem.

PR #2369, opened 2020-05-05
Adds `On Pain Flash` Scripting Hook

wookieejedi:
This PR adds a scripting hook which triggers when the pain flash is shown. It has one variable that lists what type of flash was shown, shield with `Pain_Type == 0` and hull with `Pain_Type == 1` .  I have tested and it works as expected. This new hook is very useful for triggering events whenever the player is hurt and can ultimately be used script custom pain flash colors and on a per conditional basis.


wookieejedi:
In my tests this format worked, though not sure or not if I need to keep the `CHA_PAINFLASH` or not?

code/ship/shiphit.cpp
```diff
@@ -72,6 +73,12 @@ vec3d	Original_vec_to_deader;
 
 static bool global_damage = false;
 
+const std::shared_ptr<scripting::OverridableHook> OnPainFlashHook = scripting::OverridableHook::Factory(
+	"On Pain Flash", "Called when a pain flash is displayed.",
+	{ 		
+		{"Pain_Type", "number", "The type of pain flash displayed: shield = 0 and hull = 1."},
+	}, CHA_PAINFLASH);
```
asarium, line 80: Maybe the on frame hook wasn't ideal here. You don't need to add a new `CHA_` entry. The new system takes care of that for you.

PR #2375, opened 2020-05-07
Make the lab's "show damage lightning" flag work again.

The-E:
Note that it may take a few frames for lightning to disappear when unchecking that checkbox.

### DIFF
--- a/code/lab/lab.cpp
+++ b/code/lab/lab.cpp
@@ -445,7 +445,7 @@ void labviewer_render_model(float frametime)
 				obj->hull_strength = 1.0f;
 			}
 			else {
-				obj->hull_strength = 1000000.0f;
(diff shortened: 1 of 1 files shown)


code/lab/lab.cpp
```diff
@@ -440,6 +440,13 @@ void labviewer_render_model(float frametime)
 			ship_model_update_instance(obj);
 
 			Ships[obj->instance].team_name = Lab_team_color;
+
+			if (Lab_viewer_flags & LAB_FLAG_LIGHTNING_ARCS) {
+				obj->hull_strength = 1.0f;
+			}
+			else {
+				obj->hull_strength = 1000000.0f;
```
Goober5000, line 448: Why not use the ship class max strength here?

The-E: Because I forgot to actually make that commit

asarium (approved):

PR #2376, opened 2020-05-07
Make explosions happen in the lab

The-E:
This adds a function to the lab that lets the user observe explosions; a new top-level submenu called "Actions" has been added, with "Kill Ship" currently being the only action in it.
Selecting a different ship (or the same ship) reloads the model.


TRBlount (approved):
Looks good. LGTM

Issue #1904, opened 2018-10-18
Ships not following orders to depart to bay if flag "no-subspace-drive" is set.


wookieejedi:
If I set the flag "no-subspace-drive" and the departure to docking bay then I can't tell the ship to depart (via docking bay). The ship will still follow it's departure cues to depart via the docking bay but the player cannot tell it to depart to the docking bay because the order is grayed out. Example mission is attached. 

[NoDockTest.txt](https://github.com/scp-fs2open/fs2open.github.com/files/2493100/NoDockTest.txt)

Artyrm:
I'll leave it here for now. That roughly how I understand the intended State Machine. Maybe I'll have some time to implement it the right way.

![depart](https://user-images.githubusercontent.com/21180686/47518158-d1b62480-d892-11e8-94f6-6dcc51aaa3fd.png)

wookieejedi:
This seems to be fixed now, so I'm going to close this.

PR #2377, opened 2020-05-08
Fix warp flag and code issues

wookieejedi:
This PR fixes makes the warp out checking code consistent and fixes two bugs. 

Previously, the code that checked for warp out flags did not check all of the same flags for each check. For example, one section checked the flags `never-warp` and `break-warp` but not `no-subspace-drive` (and vice versa). This lead to edge cases that could cause two bugs, one of which results in the AI becoming stationary and not being able to accept any orders again. 

The two bugs that resulted from this were:

1) The player cannot order departure via bays if the target ship has no subspace drive but the mothership is present.
2) When a ship has the break or never warp sexp applied it can be ordered to depart but just ends up becoming stationary in a dead end AI loop within `mission_do_departure()`. Thus the ship will not depart to it's bay if it was assigned to and it will not break out of the loop to do other things if it was supposed to warp out. Also, once in this loop, the ship cannot be broken out, even with player or high priority orders. Apparently I noted this bug in #1904. 

This PR fixes those bugs and makes the warp out checks consistent across the code base. It also updates the scripting function's accordingly. I have tested this across all permutations I can think of and everything works as expected.


code/ai/aigoals.cpp
```diff
@@ -1413,21 +1413,18 @@ int ai_mission_goal_achievable( int objnum, ai_goal *aigp )
 	{
 		ship *shipp = &Ships[objp->instance];
 
-		// always valid if has subspace drive
-		if (!(shipp->flags[Ship::Ship_Flags::No_subspace_drive]))
+		// always valid if has working subspace drive and not disabled
+		if ( ship_can_warp_full_check(shipp) )
 			return AI_GOAL_ACHIEVABLE;
 
 		// if no subspace drive, only valid if our mothership is present
```
Goober5000, line 1420: tweak this comment because the if() check immediately above it now checks additional things

wookieejedi:
Thanks for the valuable and detailed feedback, I have incorporated all the revisions and, just for reference posted the discussion from discord here:

Good point, I looked through everything and the new checks do make `Trying_unsuccessfully_to_warp`  obsolete. 

Here's why:
`ship_can_warp_full_check` takes into account navigation and engine status now, and that occurs before any loops that assign `Trying_unsuccessfully_to_warp` . 

My reasoning for adding navigation and engines into `ship_can_warp_full_check` was the following:
1) `cannot_warp_flags` originally included `Ship::Ship_Flags::Disabled`, which is already an engine check. Also the final warp depart check in the old `ship_can_use_warp_drive` checked for navigation.
2) I figured if a player brings up the comms menu it should only show ships that can actually depart, not ones that can be told to depart but not actually depart.

Ultimately this makes the warp decision as definitively one or the other.

wookieejedi:
Alrighty that should resolve the fork conflict.

code/ai/aigoals.cpp
Goober5000: You tweaked the wrong comment...

code/ship/ship.cpp
```diff
@@ -14673,6 +14670,55 @@ int ship_navigation_ok_to_warp(ship *sp)
...
+		departure_path_mask = wingp->departure_path_mask;
+	} else {
+		departure_location = sp->departure_location;
+		departure_anchor = sp->departure_anchor;
+		departure_path_mask = sp->departure_path_mask;
+	}
+	
+	if ( departure_location == DEPART_AT_DOCK_BAY )
+	{
+		int anchor_shipnum = ship_name_lookup(Parse_names[departure_anchor]);
```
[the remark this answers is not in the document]
Goober5000: The `Assertion` needs to go *here*, i.e. immediately before this line.  This is because it is checking the array access.  It is okay for the `departure_anchor` to be out-of-range in other places because when it isn't used it is -1.

wookieejedi:
Thanks, just incorporated.

code/ship/ship.cpp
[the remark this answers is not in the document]
Goober5000: Good. :)

code/scripting/api/objs/ship.cpp
```diff
@@ -1470,7 +1470,7 @@ ADE_FUNC(warpOut, l_Ship, NULL, "Warps ship out", "boolean", "True if successful
 	return ADE_RETURN_TRUE;
 }
 
-ADE_FUNC(canWarp, l_Ship, NULL, "Checks whether ship has a working subspace drive and is allowed to use it", "boolean", "True if successful, or nil if ship handle is invalid")
+ADE_FUNC(canWarp, l_Ship, nullptr, "Checks whether ship has a working subspace drive, is allowed to use it, and is not disabled or is limited by navigation.", "boolean", "True if successful, or nil if ship handle is invalid")
```
Goober5000, line 1473:
```suggestion
ADE_FUNC(canWarp, l_Ship, nullptr, "Checks whether ship has a working subspace drive, is allowed to use it, and is not disabled or limited by subsystem strength.", "boolean", "True if successful, or nil if ship handle is invalid")
```

code/ai/aigoals.cpp
```diff
@@ -1413,21 +1413,18 @@ int ai_mission_goal_achievable( int objnum, ai_goal *aigp )
 	{
 		ship *shipp = &Ships[objp->instance];
 
-		// always valid if has subspace drive
-		if (!(shipp->flags[Ship::Ship_Flags::No_subspace_drive]))
+		// always valid if has working subspace drive, not disabled, and not limited by navigation
```
Goober5000, line 1416:
```suggestion
		// always valid if has working subspace drive, not disabled, and not limited by subsystem strength
```

code/ai/aigoals.cpp
```diff
@@ -1413,21 +1413,18 @@ int ai_mission_goal_achievable( int objnum, ai_goal *aigp )
...
-		// if no subspace drive, only valid if our mothership is present
-
-		// check that we have a mothership and that we can depart to it
-		if (shipp->departure_location == DEPART_AT_DOCK_BAY)
-		{
-			int anchor_shipnum = ship_name_lookup(Parse_names[shipp->departure_anchor]);
-			if (anchor_shipnum >= 0 && ship_useful_for_departure(anchor_shipnum, shipp->departure_path_mask))
-				return AI_GOAL_ACHIEVABLE;
+		// if no working subspace drive or limited by engines or navigation, only valid if our mothership is present
+		// function accounts for ship's wing and returns true if ship has bay departure and mothership present, false otherwise
```
Goober5000, line 1421:
```suggestion
		// if we can't warp, we can only depart if the ship (or its wing) departs to a fighter bay and the mothership is present
```

wookieejedi:
> I tweaked a few comments for clarity. I think this is ready to go now. :)

All of those look good, thanks!

PR #2388, opened 2020-05-13
Use CMAKE_*_COMPILER_LAUNCHER to run ccache

h3xx:
Setting the compiler launcher to "ccache" is [the recommended way of enabling ccache for the build](https://cmake.org/cmake/help/latest/prop_tgt/LANG_COMPILER_LAUNCHER.html). It's been supported since CMake 3.4 (released on 2015-11-12). On previous versions of CMake, it will just ignore this setting.

Using this technique allows the user to replace the launcher with any number of combinations of solutions such as `-DCMAKE_CXX_COMPILER_LAUNCHER=ccache:distcc` which will run `ccache distcc g++ [...]` when compiling `.cpp`.

If cmake is run with it defined, it causes an error when ccache tries to run:

    ccache: error: Recursive invocation (the name of the ccache binary must be "ccache")

This was because the compiler was getting invoked as `ccache ccache [COMPILER]`

Reproduced by running cmake:

    mkdir BUILD && cd BUILD &&
    cmake \
        -DCMAKE_C_COMPILER_LAUNCHER=ccache \
        -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
        ..
    make

This results in the following error:

    [  0%] Building CXX object lib/CMakeFiles/RocketCore.dir/libRocket/Source/Core/BaseXMLParser.cpp.o
    ccache: error: Recursive invocation (the name of the ccache binary must be "ccache")
    make[2]: *** [lib/CMakeFiles/RocketCore.dir/build.make:83: lib/CMakeFiles/RocketCore.dir/libRocket/Source/Core/BaseXMLParser.cpp.o] Error 1
    make[2]: Leaving directory '/usr/share/media/src/git/fs2_open/_build'
    make[1]: *** [CMakeFiles/Makefile2:651: lib/CMakeFiles/RocketCore.dir/all] Error 2
    make[1]: Leaving directory '/usr/share/media/src/git/fs2_open/_build'
    make: *** [Makefile:172: all] Error 2
    make: Leaving directory '/usr/share/media/src/git/fs2_open/_build'


Goober5000:
The CI build error is caused by this:
```
../code/menuui/techmenu.cpp: In function ‘void techroom_do_frame(float)’:
../code/menuui/techmenu.cpp:1352:4: error: this statement may fall through [-Werror=implicit-fallthrough=]
 1352 |    if ( !(Player->flags & PLAYER_FLAGS_IS_MULTI) ) {
      |    ^~
../code/menuui/techmenu.cpp:1358:3: note: here
 1358 |   case KEY_CTRLED | KEY_UP:
      |   ^~~~
```

Here's the code:
```
		case KEY_CTRLED | KEY_DOWN:
			if ( !(Player->flags & PLAYER_FLAGS_IS_MULTI) ) {
				techroom_button_pressed(SIMULATOR_TAB);
				break;
			}
			// fall through

		case KEY_CTRLED | KEY_UP:
```

I found this tip which seems relevant:
https://stackoverflow.com/questions/45129741/gcc-7-wimplicit-fallthrough-warnings-and-portable-way-to-clear-them
> Note that suppressing the warning with marker comments only works if the compiler actually sees the comment. If the preprocessor runs separately, it needs to be instructed to preserve comments, as with the `-C` option of GCC. For example, to avoid spurious warnings with ccache, you need to specify the `-C` flag when compiling, or, with recent versions of ccache, use the `keep_comments_cpp` option.

h3xx:
I'm not sure what could be causing that error; I didn't touch that file at all. I didn't touch any C or C++ files.

Goober5000:
It's not that file that's the problem; in fact that particular snippet of code has been unchanged since the original source code release.  It's the fact that the `// fall through` comment, which should suppress the warning, isn't suppressing it.  This implies that the comment is being lost somehow.

The link that I found indicates that ccache may have something to do with the comments being dropped, so that's why I posted it.

asarium:
Yes, this might be related to ccache. There is an option to keep comments so that might be useful here.

@h3xx Try adding `- CCACHE_COMMENTS=` to the `env.global` section in `.travis.yml` and see if that fixes the issue.

h3xx:
I'm kind of doubting that this simple fix will do anything. I thought about it last night, and if it does, then the CI process never used ccache for the `g++-9` target, otherwise it would have failed before now.

Goober5000:
I'm almost certain that the CI process rebuilds all builds from scratch for every PR.

asarium:
It appears that Travis has ccache installed by default and now your fixes actually make ccache support work properly. I have purged the caches on Travis and restarted the two builds. Let's see if that fixes the issue.

h3xx:
The build failed, this would be expected if what @asarium says is true; that it never properly used ccache.

What doesn't make sense, however, is how I can build this on my system using ccache 3.7.9 (`keep_cpp_comments=false` (default) ), and GCC 9.3.0.

... The ccache comment-stripper in newer versions of ccache omit stripping the `// fall through` comments.
... Or GCC 9.0.0 - 9.2.9 checks for implicit fallthru, and 9.3.0 doesn't (unlikely since compilers have been getting pickier in recent years)

Goober5000:
I wouldn't be surprised if the compiler changed in the past week or two.  Issue #2392 is failing with a "magic number" warning, while #2325, which added that line, passed without emitting that warning.

h3xx:
Perhaps it's a CI caching issue. As far as you know, do e.g. the GCC-7 and the GCC-9 jobs share the same CI cache?

If they do, it's probably wrong. The solution could be that a matter of just adding the `CACHE_NAME` public environment variable to each compiler type, e.g. `CACHE_NAME=${TRAVIS_OS_NAME}-${CC}-gcc9`.

[More documentation on the `CACHE_NAME` and caching wrt Travis CI here.](https://docs.travis-ci.com/user/caching/#caches-and-build-matrices)

Goober5000:
I don't know, myself.  @asarium might.

h3xx:
Added updates to CI YAML. :crossed_fingers:

asarium:
The [ccache manual](https://ccache.dev/manual/latest.html#_common_hashed_information) claims that the used compiled is included in the cache hit check so I would guess that it support different compilers.

One thing I can think of is that it simply uses an outdated version of ccache that does not handle comments correctly yet. However, if we cannot find a proper solution for this then the issue should just be fixed properly by using out `FALLTHROUGH` macro which is used in other places as well.

h3xx:
I *will* figure this out, even if it means taking apart the whole CI build system piece by piece until I figure out exactly what the issue is... But not this week. I have an interview to work at Facebook this week, and need all my energy focused towards studying and putting my best self forward.

The storm will hopefully be abated by this weekend when I can devote some time to this.

Goober5000:
@h3xx Any chance of taking another look at this?

Goober5000:
My guess is that recent changes to our CI system have fixed this.  In particular the fall-through warning which triggered on an unrelated part of the code has been fixed.

@h3xx can you rebase this PR onto latest master and push it again?

h3xx:
Rebased. I only included the one commit because the second one ("CI: Fix build failure when using ccache+GCC-9") was intended to work around a ccache caching issue in CI to support of the change; if CI is fixed, the workaround won't be necessary.

asarium (approved):

PR #2389, opened 2020-05-13
make Goal_target_names dynamic

Goober5000:
If ever there were more than 100 goal targets in a mission, the old version of ai_get_goal_target_name would start corrupting memory.  Rather than making a hacky workaround, this just makes it dynamic.

There is a bit of cascading const-correctness due to the requirement to reference `.c_str()`, but fortunately not too much.

~~Note: this PR will not compile cleanly for FRED until #2387 is merged.~~


code/ai/ai.cpp
```diff
@@ -82,22 +81,23 @@ void set_wingnum(int objnum, int wingnum)
...
-	int i;
+	Assertion(name != nullptr && index != nullptr, "Arguments cannot be null!");
 
-	for (i=0; i < Total_goal_target_names; i++)
-		if (!stricmp(name, Goal_target_names[i])) {
+	for (int i = 0; i < static_cast<int>(Goal_target_names.size()); i++) {
+		if (!stricmp(name, Goal_target_names[i].c_str())) {
 			*index = i;
-			return Goal_target_names[i];
+			return Goal_target_names[i].c_str();
```
asarium, line 91: As far as I can tell, the returned value will be stored for an extended time which will lead to issues if the vector is resized. In that case the string data might be moved to a different pointer and the old one will become invalid.

A quick fix for this would be to allocate the `SCP_string`s dynamically then they themselves wouldn't be reallocated.

Goober5000: Oh, good point.

Goober5000: I used `vm_strdup` because it fit right in with the rest of the API.

PR #2390, opened 2020-05-14
clean up handling of translated vs. untranslated English

Goober5000:
Since the original release of FS2, internal strings have been translated to their English entries in strings.tbl but external strings have not been translated to their English entries in tstrings.tbl.  It appears that string translation was universal until it was disabled via a hack, probably due to FRED idiosyncrasies (see comment in `fred_init()`), and then not re-enabled for release.  PR #2343 re-enabled it, but this turned out to have undesirable side-effects since many mods used incorrect XSTR indexes for external strings.

This PR restores the previous behavior and also cleanly separates localization into translated and untranslated modes.  FRED will always run in untranslated mode.  FSO will run in translated mode for all languages except English, where it will run in untranslated mode UNLESS the new option `$Use tabled strings for the default language` is enabled in game_settings.tbl.

### DIFF
--- a/code/cutscene/ffmpeg/FFMPEGDecoder.cpp
+++ b/code/cutscene/ffmpeg/FFMPEGDecoder.cpp
@@ -286,7 +286,7 @@ std::unique_ptr<DecoderStatus> initializeStatus(std::unique_ptr<InputStream>& st
 
 	if (status->subtitleStream == nullptr) {
 		// We don't have an external subtitle stream or loading from that file has failed
-		auto& current_language = Lcl_languages[Lcl_current_lang];
+		auto& current_language = Lcl_languages[Lcl_current_lang == LCL_UNTRANSLATED ? LCL_DEFAULT : Lcl_current_lang];
 		for (uint32_t i = 0; i < ctx->nb_streams; ++i) {
 			auto test_stream = ctx->streams[i];
 
@@ -428,7 +428,7 @@ std::unique_ptr<InputStream> openInputStream(const SCP_string& name) {
 
 std::unique_ptr<InputStream> openSubtitleStream(const SCP_string& name)
 {
-	auto& current_language = Lcl_languages[Lcl_current_lang];
+	auto& current_language = Lcl_languages[Lcl_current_lang == LCL_UNTRANSLATED ? LCL_DEFAULT : Lcl_current_lang];
 
 	for (auto ext : CHECKED_SUBT_EXTENSIONS) {
 		SCP_string fileName;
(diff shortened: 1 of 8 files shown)


MageKing17:
Why did this PR disable translating from strings.tbl by default?

Goober5000:
> This PR ... cleanly separates localization into translated and untranslated modes.

The previous code had a weird superposition of states where internal strings were translated but external strings were not.  As I noted in the first post, I am pretty sure that this weird superposition was introduced as a temporary situation during development and then not fixed before release.

In order to properly fix the translation code, it was necessary to treat internal and external strings the same way.  The default for both is untranslated, but this can be changed by the game_settings flag.

MageKing17:
> In order to properly fix the translation code, it was necessary to treat internal and external strings the same way. The default for both is untranslated, but this can be changed by the game_settings flag.

There was absolutely not reason to change the default behavior of the internal `XSTR()` function, and this is a regression from retail behavior.

Goober5000:
I just told you the reason.  It was necessary to fix the superposition of states.  This was just one of the many things that had to be rewritten in the translation code in order to fix translation support for other languages.

And regarding the change from retail behavior, there is a different PR out there somewhere that syncs all of the external `XSTR()` strings between what's in the code and what's in strings.tbl.

MageKing17:
> I just told you the reason. It was necessary to fix the superposition of states. This was just one of the many things that had to be rewritten in the translation code in order to fix translation support for other languages.

Nothing about that code would break if the `Lcl_current_lang != LCL_UNTRANSLATED` check wasn't in `XSTR()`. It was not a necessary change, and it breaks backwards compatibility.

Goober5000:
I don't remember the details, but there were important reasons I did things this way.  This code went through many revisions and testing cycles.  And there was discussion happening in parallel with m!m and Novachen.

Anyway, let's move to the actual point.  It might be possible to force the code back to a superposition of states but I am certain it will make the code more hackish, harder to maintain, and maybe even slower.  It's also been a while since I looked at the code.  What is the situation that brought this PR back to the forefront?

MageKing17:
> What is the situation that brought this PR back to the forefront?

[Any existing mods](https://www.hard-light.net/forums/index.php?topic=97346.0) that modified built-in strings by changing strings.tbl abruptly stopped working. It's all well and good to say they should all enable the game_settings.tbl setting, but now you're making every such mod make an update when *you changed retail behavior*.

Goober5000:
Point taken.  I'll see what I can do.

PR #2392, opened 2020-05-15
modify weapons loadout check

Goober5000:
Based on feedback, this PR now allows gaps in weapon loadouts, but if any gaps exist they must come at the bottom.


Goober5000:
Disregard the CI error; the XSTR id is not a magic number.

PR #2396, opened 2020-05-16
revert the clang-tidy changes from #2298

Goober5000:
These are far too pedantic, and were not related to the feature in #2298 anyway.

This will fix the CI complaints in #2392 and #2394, as well as the CI complaint which delayed merging of #2298 itself.


Goober5000:
Understood.  Sorry about that then.

PR #2397, opened 2020-05-17
Add Lua API for promises and async coroutines

asarium:
This adds a Lua API for representing asynchronous operations and running
coroutines based on those operations comfortably. This is very similar
to features other languages have introduced such as C# or JavaScript. In
fact, the whole "promise" thing is derived from how JavaScript
represents these operations.

This is the initial and very basic system for supporting async
coroutines in Lua. There is no native support for running these
operations at specific points (they run whenever the awaited promise
resolved) or making sure they do not leak through different game states.


wookieejedi:
Forked, built and tested and it works as expected, thanks for adding this!

PR #2399, opened 2020-05-18
Fix 'Guard Attack This' Behavior with AI Flag

wookieejedi:
This PR allows the setting `Guards Attack This` in `objecttype.tbl` to actually work.

Background: there are two instances in the code where AI that is patrolling/guarding a ship decides to go attack guardees attacker. One case is if the guardee is hit, and the other case if there are just hostile ships present (ie it scans the mission every so often).
The `Guards Attack This` check was only applied to the scan mission case, and not the guardee hit case. Thus, if the guardee was hit by a ship that had the `Guards Attack This` setting off, the guard would still attack since that check was never being performed for that guard case. By default all object types have `Guards Attack This` set to `true.` But if a modder didn't want fighters guarding a ship to go off and try to kill a large capital ship then they may think setting this value would be a good way to do that. Unfortunately, this wouldn't work, for the reasons outlined above.

Change: this PR adds the `Guards Attack This` check into the guardee is hit case, essentially making the behavior work as expected by a modder. Given this is a AI behavior change I have added an optional AI flag to toggle this fix on. I have tested this and it works as expected.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -472,8 +472,6 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$use actual primary range:", AI::Profile_Flags::Use_actual_primary_range);
 
-				set_flag(profile, "$fix Guard Attack This bug:", AI::Profile_Flags::Fix_guard_attack_this_bug);
-
 				profile->bay_arrive_speed_mult = 1.0f;
 				profile->bay_depart_speed_mult = 1.0f;
 				if (optional_string("$bay arrive speed multiplier:")) {
@@ -647,7 +645,6 @@ void ai_profile_t::reset()
 		flags.set(AI::Profile_Flags::Fix_ai_path_order_bug);
 		flags.set(AI::Profile_Flags::Aspect_invulnerability_fix);
 		flags.set(AI::Profile_Flags::Use_actual_primary_range);
-		flags.set(AI::Profile_Flags::Fix_guard_attack_this_bug);
 	}
 	// this flag has been enabled ever since 3.7.2
 	if (mod_supports_version(3, 7, 2)) {
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -647,6 +647,7 @@ void ai_profile_t::reset()
 		flags.set(AI::Profile_Flags::Fix_ai_path_order_bug);
 		flags.set(AI::Profile_Flags::Aspect_invulnerability_fix);
 		flags.set(AI::Profile_Flags::Use_actual_primary_range);
+		flags.set(AI::Profile_Flags::Fix_guard_attack_this_bug);
 	}
 	// this flag has been enabled ever since 3.7.2
 	if (mod_supports_version(3, 7, 2)) {
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -472,6 +472,8 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$use actual primary range:", AI::Profile_Flags::Use_actual_primary_range);
 
+				set_flag(profile, "$fix Guard Attack This bug:", AI::Profile_Flags::Fix_guard_attack_this_bug);
+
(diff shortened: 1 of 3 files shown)


code/ai/ai_profiles.cpp
```diff
@@ -472,6 +472,8 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$use actual primary range:", AI::Profile_Flags::Use_actual_primary_range);
 
+				set_flag(profile, "$fix Guard Attack This bug:", AI::Profile_Flags::Fix_guard_attack_this_bug);
```
asarium, line 475: Since this is a sensible fix this should be enabled by default for newer mods using the supported version feature. Maybe you could add a version of `set_flag` that accepts a default value that you then initialize with the return value of `mod_supports_version()`.

wookieejedi: Good idea, looks like there is a nice section that Goober made specifically for bug fixes, so I'll add it to that.

wookieejedi:
Hmm Travis has been building for about a day...is there a way to restart it?

asarium:
I restarted the build. Let's see if that fixes the issue.

wookieejedi:
@Goober5000 Good question. Yes "Guard Attack This" is in the default table of `objecttypes.tbl` in the code base, and the one place it was actually used prior to this PR was line 9708 of `aicode.cpp`. Also, in that default table every object type had "Guard Attack This" set to YES. I checked the git history of that default table file, and the earliest history still had "Guard Attack This". I then checked mantis and couldn't find a change listed there either. Looking on the wiki for `objecttypes.tbl` the entire `objecttypes.tbl` was exposed to tabling with 3.6.10 it seems:
(https://wiki.hard-light.net/index.php/Objecttypes.tbl#.2BGuards_Attack_this:)

So to my best estimation it has been a default feature for all object types since retail.

Goober5000:
Well, not *every* type has "Guard Attack This" set to YES.  As implied by the comment for line 9708, cargo containers don't have it.  Can you go through objecttypes.tbl and pick out the classes that don't have it?

code/ai/aicode.cpp
```diff
@@ -9613,7 +9613,18 @@ void maybe_update_guard_object(object *hit_objp, object *hitter_objp)
 
 			if ((aip->mode == AIM_GUARD) || (aip->active_goal == AI_ACTIVE_GOAL_DYNAMIC)) {
 				if (aip->guard_objnum == OBJ_INDEX(hit_objp)) {
-					guard_object_was_hit(objp, hitter_objp);
```
Goober5000, line 9616: You should think about whether the fix flag should be checked immediately after `if ((aip->mode == AIM_GUARD) || (aip->active_goal == AI_ACTIVE_GOAL_DYNAMIC)) {`.  As the code is now, you treat ships differently depending on whether they are part of a wing.

wookieejedi:
> Well, not _every_ type has "Guard Attack This" set to YES. As implied by the comment for line 9708, cargo containers don't have it. Can you go through objecttypes.tbl and pick out the classes that don't have it?

Ah good point, the three that do not have it are `Navbuoy`, `Escape Pod`, `Cargo`. It makes more sense now that the single previous check has always worked for those types, since they usually are not firing at the guardee (and hence not triggering the object hit section of the code).

code/ai/aicode.cpp
wookieejedi: Good point, I updated the ordering.

Goober5000:
> Ah good point, the three that do not have it are `Navbuoy`, `Escape Pod`, `Cargo`. It makes more sense now that the single previous check has always worked for those types, since they usually are not firing at the guardee (and hence not triggering the object hit section of the code).

Yeah, those were the classes marked as "harmless" in the original public code release.

Ok, so after this detective work, I propose just removing the AI profiles flag.  Given that those three ship types will never fire on anything (at least in the retail objecttypes), there is no difference, from a retail perspective, from attacking *all* ship types and attacking the *guards-attack-this* ship types.

wookieejedi:
I'm good with that, what about you @asarium?

wookieejedi:
Alrighty went ahead an just removed the flag.

PR #2400, opened 2020-05-18
Enable sensible clang readability checks

asarium:
This enables the clang readability checks that I found to be sensible. I
disabled a few that would have caused a lot of unnecessary warnings in
our code.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -14,6 +14,17 @@ misc-*,\
 -misc-static-assert,\
 -misc-unused-parameters,\
 -misc-non-private-member-variables-in-classes,\
+readability-*,\
+-readability-function-size,\
+-readability-identifier-naming,\
+-readability-implicit-bool-conversion,\
+-readability-inconsistent-declaration-parameter-name,\
+-readability-isolate-declaration,\
+-readability-magic-numbers,\
+-readability-named-parameter,\
+-readability-qualified-auto,\
+-readability-uppercase-literal-suffix,\
+-readability-else-after-return,\
 "
 WarningsAsErrors: ''
 HeaderFilterRegex: 'code/.*$|freespace2/.*$|qtfred/.*$|test/src/.*$|build/.*$'
@@ -23,5 +34,7 @@ CheckOptions:
     value:        'assert,Assert,Assertion'
   - key:          'modernize-use-emplace.ContainersWithPushBack'
     value:        'std::vector;std::deque;std::list;SCP_vector;SCP_deque;SCP_list'
+  - key:          'readability-braces-around-statements.ShortStatementLines'
+    value:        '4' # Avoid flagging simple if (...) return false; statements
 ...
 


.clang-tidy
```diff
@@ -23,5 +33,7 @@ CheckOptions:
     value:        'assert,Assert,Assertion'
   - key:          'modernize-use-emplace.ContainersWithPushBack'
     value:        'std::vector;std::deque;std::list;SCP_vector;SCP_deque;SCP_list'
+  - key:          'readability-braces-around-statements.ShortStatementLines'
+    value:        '2' # Avoid flagging simple if (...) return false; statements
```
Goober5000, line 37: This option should not be included.  It affected many lines in my PRs, and if I'm not mistaken it affected a few lines in your PR too.  I specifically reverted this in #2396.

.clang-tidy
```diff
@@ -14,6 +14,16 @@ misc-*,\
...
+readability-*,\
+-readability-function-size,\
+-readability-identifier-naming,\
+-readability-implicit-bool-conversion,\
+-readability-inconsistent-declaration-parameter-name,\
+-readability-isolate-declaration,\
+-readability-magic-numbers,\
+-readability-named-parameter,\
+-readability-qualified-auto,\
+-readability-uppercase-literal-suffix,\
```
Goober5000, line 26: I looked through each one of the `readability-*` checks.  Most of them I agree with, with two exceptions:

1) The `readability-else-after-return` check should be removed, as there are many places in the codebase where this happens.
2) The `readability-redundant-member-init` check should be removed as well.  IIRC there are a few classes with many members that are initialized, and the initialization list matches the list of fields.  Explicitly calling the default constructor visually indicates that the coder did not forget about initializing that field.

.clang-tidy
asarium: Maybe the value is too low. I will increase it to 4. That should make sure that only the bad cases are flagged by this.

.clang-tidy
asarium: 1. I can see that. I'll add it to the ignore list.
2. I don't agree on this. Calling the default constructor explicitly just adds unnecessary clutter and makes it harder to see actual issues in the code.

Goober5000: I don't feel as strongly about 2) as the other points, but consider these two examples from the code.  Wouldn't they be flagged by the `readability-redundant-member-init` check?

```
struct DiyFp {
    DiyFp() : f(), e() {}
```
```
	main_hall_region()
		: mask(0), key(0), description(), action(0), lua_action()
	{}
```

asarium: Yes, both cases would be flagged. In my opinion, that is correct since fixing it would remove code and non-existing code can not cause bugs.

Goober5000: Ok, I'm willing to see what happens.

PR #2402, opened 2020-05-19
SEXP data caching

Goober5000:
This is a first-level pass at optimizing SEXP execution.  In brief, rather than each SEXP node parsing and looking up each argument every frame, this PR saves the value in the SEXP node after it is determined.  Most SEXP data types (e.g. wings, IFFs, messages) will not change their reference for the duration of a mission.

At this point, ships, numeric literals, and variables are all cached.  Since ships can change status during the mission, this PR also introduces a ship registry that maintains the ship status and references to pointers.  This registry is applicable outside the SEXP system; see the example in the AI goal code.

To keep changes manageable, I limited the ship-related changes to mostly those SEXPs which used `sexp_get_ship_from_node`.  After this PR is merged I intend to make a followup PR dedicated specifically to refactoring the rest of the ship SEXPs, and additional followup PRs for other data types as needed.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -7479,7 +7479,6 @@ void ship_cleanup(int shipnum, int cleanup_mode)
 	// Goober5000 - handle ship registry
 	auto entry = &Ship_registry[Ship_registry_map[shipp->ship_name]];
 	entry->status = ShipStatus::EXITED;
-	entry->pobjp = nullptr;
 	entry->objp = nullptr;
 	entry->shipp = nullptr;
 	entry->cleanup_mode = cleanup_mode;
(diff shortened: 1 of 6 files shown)


code/ship/ship.cpp
```diff
@@ -7458,6 +7476,13 @@ void ship_cleanup(int shipnum, int cleanup_mode)
 	object *objp = &Objects[shipp->objnum];
 	char *jumpnode_name = nullptr;
 
+	// Goober5000 - handle ship registry
+	auto entry = &Ship_registry[Ship_registry_map[shipp->ship_name]];
```
asarium, line 7480: An assert to check if the ship actually exists before this would be nice to catch potential issues more easily.

Goober5000: :+1:

PR #2404, opened 2020-05-19
Fixes gr.drawOffscreenIndicator

wookieejedi:
Previously the scripting call `gr.drawOffscreenIndicator` did not properly draw the indicator and/or return coordinates. After some bug hunting it turns out it is because there gauge was not being set to active. This PR adds that line and restores the intended functionality. Tested and works as expected.

### DIFF
--- a/code/scripting/api/libs/graphics.cpp
+++ b/code/scripting/api/libs/graphics.cpp
@@ -1027,30 +1027,36 @@ ADE_FUNC(drawOffscreenIndicator, l_Graphics, "object Object, [boolean draw=true,
 		if (default_hud_gauges[j]->getObjectType() == HUD_OBJECT_OFFSCREEN) {
 			HudGaugeOffscreen *offscreengauge = static_cast<HudGaugeOffscreen*>(default_hud_gauges[j].get());
 
-			offscreengauge->preprocess();
-			offscreengauge->onFrame(flFrametime);
-
-			// needs to be turned on so it can pass the canRender() condition --wookieejedi
-			offscreengauge->updateActive(true);
-
-			if ( !offscreengauge->canRender() ) {
-				break;
-			}
-
-			offscreengauge->resetClip();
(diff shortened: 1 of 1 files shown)


asarium:
Does the active status properly reset after this is called? It looks like this would alter the state of one of the HUD gauges and could lead to weird HUD behaviour.

wookieejedi:
@asarium Good question, and currently the script does not restore the original status (but I can add that in). In reality I think the best solution would be to move the `!offscreengauge->canRender()` check inside the `draw` bool check. Ultimately, the documentation states that the function will return the coordinates regardless, and that's really what I'm after. As it stands right now the function will only return the coordinates if the gauge is on. 

I'll just committed an updated fix that I think makes more sense. Also I tested this commit with rendering on and off and it works as expected.

wookieejedi:
Did another test with disabling the default gauge before hand and it properly does not show the gauge but does return the coordinates as expected.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1027,27 +1027,36 @@ ADE_FUNC(drawOffscreenIndicator, l_Graphics, "object Object, [boolean draw=true,
...
+					if (!setcolor)
+						hud_set_iff_color(targetp, 1);
 
-				if (!setcolor)
-					hud_set_iff_color(targetp, 1);
+					offscreengauge->renderOffscreenIndicator(&outpoint, dir, distance, tri_separation, true);
 
-				offscreengauge->renderOffscreenIndicator(&outpoint, dir, distance, tri_separation, true);
+					// now that the gauge is rendered restore the original status
+					offscreengauge->updateActive(original_status);
```
asarium, line 1058: I think this should be done in every case and not just in the `if`.

wookieejedi: Makes sense, I've moved it out of that specific block.

Issue #2407, opened 2020-05-21
Feature request: Let -weaponspew cover all weapons


EatThePath:
Limiting it to just player-usable weapons makes plenty of sense for certain uses, but it'd be useful to me at least if I had the option of making it cover all weapons. Maybe -weaponspewall?

If nothing else it'd help in keeping tabs on exactly how many weapon slots a mod stack is using.

Goober5000:
Only because this is a fairly minor change...

EatThePath:
resolved by #2409 and #2467

PR #2409, opened 2020-05-22
expand -weaponspew to all non-beam weapons

Goober5000:
In response to request #2407, this expands -weaponspew to optionally list all non-beam weapons.

### DIFF
--- a/code/cmdline/cmdline.h
+++ b/code/cmdline/cmdline.h
@@ -150,7 +150,7 @@ extern bool Cmdline_debug_window;
 extern bool Cmdline_graphics_debug_output;
 extern bool Cmdline_log_to_stdout;
 
-enum WeaponSpewType { NONE = 0, STANDARD, ALL };
+enum class WeaponSpewType { NONE = 0, STANDARD, ALL };
 extern WeaponSpewType Cmdline_spew_weapon_stats;
 
 
--- a/code/cmdline/cmdline.h
+++ b/code/cmdline/cmdline.h
@@ -149,7 +149,9 @@ extern bool Cmdline_show_video_info;
 extern bool Cmdline_debug_window;
 extern bool Cmdline_graphics_debug_output;
 extern bool Cmdline_log_to_stdout;
-extern bool Cmdline_spew_weapon_stats;
+
+enum WeaponSpewType { NONE = 0, STANDARD, ALL };
+extern WeaponSpewType Cmdline_spew_weapon_stats;
 
 
 #endif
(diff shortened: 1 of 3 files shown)


EatThePath:
I very much appreciate accommodating my request. But I feel compelled to point out that omitting beam weapons not only invalidates the one concrete use case I gave, keeping track of the total number of weapons used, but also would make it skip at least one player-equippable weapon I'm aware of in an established mod. That being the vishnan gravity spike in blue planet. Granted, that weapon's dps looks like it's currently being calculated incorrectly... still seems worth bringing up, though.

Goober5000:
I'm pretty pressed for time these days.  Incorporating beam weapons would take a bit more work because beam damage and DPS is calculated differently.  I don't currently have time to investigate this.

Maybe someone else would be willing to take a look.  @wookieejedi is my first suggestion.

FrikgFeek:
Beam DPS is somewhat more intuitive these days than it was a few year before the change to "smooth out" the damage. So now it's $Damage*5.5*(+Life) per pulse. I tested with various durations under 0.182s(which is roughly 1/5.5s) and it scales pretty closely to that formula unlike 3.8.0 where it will do the initial "tick" of damage(so $Damage once) even with durations under 0.182s.

So from there DPS is $Damage*5.5*(+Life)/$Fire Wait. 
Warmup and Warmdown DO NOT factor into DPS, the LRED is a retail example where on insane with full firerate the beam will warm up, fire, then immediately start warming up for the next shot as $Fire Wait is 10 and +Life is 7. So $Fire Wait counts from warmup to warmup and will skip warm down if necessary.

EatThePath:
No need to bring @wookieejedi into it, I've got most of a PR ready of my own now that'll expand this to handle both beams and burst weapons, once this one is in.

Goober5000:
@FrikgFeek Is the damage now calculated at *exactly* 5.5 times whatever?  Because the true value is 5.88 and change.  See `BEAM_DAMAGE_TIME` in beam.h; beam damage was applied every 170 milliseconds which is ~5.88235... times per second.

code/cmdline/cmdline.h
```diff
@@ -149,7 +149,9 @@ extern bool Cmdline_show_video_info;
 extern bool Cmdline_debug_window;
 extern bool Cmdline_graphics_debug_output;
 extern bool Cmdline_log_to_stdout;
-extern bool Cmdline_spew_weapon_stats;
+
+enum WeaponSpewType { NONE = 0, STANDARD, ALL };
```
asarium, line 153: Please use `enum class`

Goober5000: Ok.

PR #2411, opened 2020-05-23
fix various issues with PXO stats

notimaginative:
Leaving as draft until it's confirmed to work 100%


notimaginative:
Confirmed to work so it should be ready to commit.

PR #2416, opened 2020-05-26
Fix a bug in the debris deletion code

The-E:
Fix a bug in the debris deletion code introduced by the lab explosion feature commit. This also makes the debris code use the new flag system to mark debris pieces as nondeleteable instead of relying on a magic lifeleft value.

### DIFF
--- a/code/debris/debris.cpp
+++ b/code/debris/debris.cpp
@@ -95,7 +95,7 @@ void debris_init()
 	// Reset everything between levels
 	Num_debris_pieces = 0;
 	for (i=0; i<MAX_DEBRIS_PIECES; i++ )	{
-		Debris[i].flags = 0;
+		Debris[i].flags.reset();
 		Debris[i].sound_delay = 0;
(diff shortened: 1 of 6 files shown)


code/debris/debris.cpp
```diff
@@ -171,11 +169,11 @@ void debris_delete( object * obj )
 	db = &Debris[num];
 
 	Assert( Num_debris_pieces >= 0 );
-	if ( db->is_hull && (db->flags & DEBRIS_EXPIRE) ) {
-		debris_clear_expired_flag(db);
+	if ( db->is_hull && (!db->flags[Debris_Flags::DoNotExpire]) ) {
```
Goober5000, line 172: You check the DoNotExpire flag both here and inside the `debris_remove_from_hull_list` function.  Want to remove this check?

Goober5000 (commented):
Just one comment...

Goober5000 (approved):

PR #2424, opened 2020-05-30
Fix a crash in the Lab when viewing bombs

### LINKED ISSUE #2422: Some Weapons CTD from the F3 Lab
Starting in the May-10 nightly, some weapons cause FSO to CTD without any other warning. This happens with retail data. Weapons that cause the crash include Stiletto II, Infyrno, Cyclops, and others. It is not immediately apparent to me what is different about these weapons that might cause a CTD while other weapons don't.

The-E:
This fixes #2422 

This also adds a few cleanup functions to the lab to make sure that debris, explosions and particles are deleted when switching ship models.


Goober5000 (approved):
Aha.  `GM_LAB` is a good solution.

PR #2434, opened 2020-06-02
Temporary fix for thrusters rendering through ship geometry

TRBlount:
Fixes an issue with some ships having thrusters render through their hull.

### DIFF
--- a/code/model/modelrender.cpp
+++ b/code/model/modelrender.cpp
@@ -2094,7 +2094,7 @@ void model_queue_render_thrusters(model_render_params *interp, polymodel *pm, in
 		for (j = 0; j < bank->num_points; j++) {
 			Assert( bank->points != NULL );
 
-			float d, D;
+			float d;
 			vec3d tempv;
 			glow_point *gpt = &bank->points[j];
 			vec3d loc_offset = gpt->pnt;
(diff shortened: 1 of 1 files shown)


code/model/modelrender.cpp
```diff
@@ -2213,7 +2213,7 @@ void model_queue_render_thrusters(model_render_params *interp, polymodel *pm, in
 			if (thruster_info.tertiary_glow_bitmap >= 0) {
 				p.screen.xyw.w -= w;
 				p.r = p.g = p.b = p.a = (ubyte)(255.0f * fog_int);
-				batching_add_volume_bitmap_rotated(thruster_info.tertiary_glow_bitmap, &p, magnitude * 4, w * 0.6f * thruster_info.tertiary_glow_rad_factor, fog_int, -(D > 0) ? D : -D);
+				batching_add_volume_bitmap_rotated(thruster_info.tertiary_glow_bitmap, &p, magnitude * 4, w * 0.6f * thruster_info.tertiary_glow_rad_factor, fog_int, -w*0.5f);
```
The-E, line 2216: Apparently, the missing D is causing problems for the CI....

TRBlount: Yeah, I believe `D` becomes unused, I'll just remove it.

PR #2437, opened 2020-06-04
another stage of the sexp refactor

Goober5000:
This PR uses `eval_ship` and `eval_wing` to optimize `object_ship_wing_point_team` and `sexp_deal_with_ship_flags`.  The API changed a bit, but to ensure compatibility, there are fields and functions to support the old behavior.  These are temporary and will be removed when all changes have been merged via PRs.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -5340,7 +5340,7 @@ struct object_ship_wing_point_team
 	{
 		ship_entry = ship_registry_get(sp->ship_name);
 		shipp = ship_entry->shipp;
-		p_objp = ship_entry->pobjp;
+		p_objp = ship_entry->p_objp;
 	}
 
 	object_ship_wing_point_team(p_object *pobjp)
@@ -5348,7 +5348,7 @@ struct object_ship_wing_point_team
 	{
 		ship_entry = ship_registry_get(pobjp->name);
 		shipp = ship_entry->shipp;
-		p_objp = ship_entry->pobjp;
+		p_objp = ship_entry->p_objp;
 	}
 
 	object_ship_wing_point_team(ship_obj *so)
@@ -5458,7 +5458,7 @@ void eval_object_ship_wing_point_team(object_ship_wing_point_team *oswpt, int no
 		oswpt->object_name = ship_entry->name;
 		oswpt->objp = ship_entry->objp;
 		oswpt->shipp = ship_entry->shipp;
-		oswpt->p_objp = ship_entry->pobjp;
+		oswpt->p_objp = ship_entry->p_objp;
 
 		switch (ship_entry->status)
 		{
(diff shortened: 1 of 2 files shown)


code/parse/sexp.cpp
```diff
@@ -14486,18 +14531,17 @@ bool sexp_check_flag_arrays(char *flag_name, Object::Object_Flags &object_flag,
...
 	AI::AI_Flags ai_flag = AI::AI_Flags::NUM_VALUES;
 
 	auto ship_entry = eval_ship(node);
 	if (!ship_entry || ship_entry->status == ShipStatus::EXITED) {
-		return SEXP_KNOWN_FALSE;
+		return SEXP_NAN_FOREVER;
 	}
 	if (ship_entry->status == ShipStatus::NOT_YET_PRESENT) {
-		return SEXP_CANT_EVAL;
+		return SEXP_NAN;
```
asarium, line 14544: Are these intentional? Seems odd to return NaN from a SEXP that returns a boolean value.

code/parse/sexp.cpp
```diff
@@ -14649,58 +14683,70 @@ void multi_sexp_alter_ship_flag()
...
-	else {
-		while (Current_sexp_network_packet.get_int(type)) {
-			oswpt.clear();
-			oswpt.type = type; 
+	else
+	{
+		int type;
+		while (Current_sexp_network_packet.get_int(type))
+		{
+			object_ship_wing_point_team *oswptp = nullptr;
```
asarium, line 14695: Please use `unique_ptr` here to ensure that the memory is always deallocated. If I see that correctly that should work here.

code/parse/sexp.cpp
Goober5000: It's intentional.  In fact I had to make this change in a few places in the SEXP refactor (including commits that aren't yet in PRs).  I checked the documentation and I also went back to the original public code release to make sure I had the right idea.

SEXP_NAN represents an invalid value, or no value, because it's not possible to determine what the number is supposed to be, or what the boolean value is supposed to be.  SEXP_CANT_EVAL means that the event system defers evaluating the event until some future time.  CANT_EVAL is used to prevent directives from immediately appearing, and it is almost exclusively used for "objective" type SEXPs like `has-destroyed-delay`.

code/parse/sexp.cpp
Goober5000: How should that work, exactly?  The compiler didn't like it until I used `unique_ptr` in each place, like so...

```
std::unique_ptr<object_ship_wing_point_team> oswptp = nullptr;
...
oswptp = std::unique_ptr<object_ship_wing_point_team>(new object_ship_wing_point_team(shipp));
...
oswptp = std::unique_ptr<object_ship_wing_point_team>(new object_ship_wing_point_team(p_objp));
...
oswptp = std::unique_ptr<object_ship_wing_point_team>(new object_ship_wing_point_team(wingp));
...
oswptp = std::unique_ptr<object_ship_wing_point_team>(new object_ship_wing_point_team());
```

That's a lot more verbose.  And even with all that, the compiler didn't like `delete oswptp;` at the end of the block.

Goober5000: After a bit of digging and a helpful link...
https://stackoverflow.com/questions/16061407/should-i-assign-or-reset-a-unique-ptr
...I think this is a bit better:

```
std::unique_ptr<object_ship_wing_point_team> oswptp;
...
oswptp.reset(new object_ship_wing_point_team(shipp));
...
oswptp.reset(new object_ship_wing_point_team(p_objp));
...
oswptp.reset(new object_ship_wing_point_team(wingp));
...
oswptp.reset(new object_ship_wing_point_team());
```

That's less verbose but still chokes on `delete`.  Although it appears that `delete` is no longer needed.

code/parse/sexp.cpp
asarium: Thanks for the explanaition.

code/parse/sexp.cpp
asarium: Yes, that is how I meant that to work. `delete` is not necessary since the `unique_ptr` will do that for you once it goes out of scope which also covers cases like exceptions.
I think it is a good idea to use smart pointers wherever you can to avoid having to worry about the finer details of control flow.

PR #2438, opened 2020-06-04
upgrade get_player_from_ship_node

Goober5000:
This PR upgrades the function, and the sexps that use it, to take advantage of the new sexp caching.  It also makes `script_eval_multi` more robust in the face of players who might be respawning.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -940,9 +940,6 @@ int get_effect_from_name(const char* name);
 #define NO_OPERATOR_INDEX_DEFINED		-2
 #define NOT_A_SEXP_OPERATOR				-1
 
-// Karajorma - some useful helper methods
-player * get_player_from_ship_node(int node, bool test_respawns = false);
-
 // hud-display-gauge magic values
 #define SEXP_HUD_GAUGE_WARPOUT "warpout"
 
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -16016,24 +15997,11 @@ void sexp_destroy_instantly(int n)
...
-
-			// if it's the player don't destroy
-			if (ship_obj_p != Player_obj)
-			{
-				ship_destroy_instantly(ship_obj_p);
-			}
-		}
-	}
+	while (Current_sexp_network_packet.get_ship(shipp))
+		ship_destroy_instantly(&Objects[shipp->objnum]);
```
asarium, line 16004: This drops the check for the player ship. Not exactly sure what object that would be in multiplayer but may be relevant here.

code/parse/sexp.cpp
```diff
@@ -23339,95 +23169,88 @@ int sexp_script_eval(int node, int return_type, bool concat_args = false)
...
 	}
 	// go through all the ships that were sent and see if any of them match this client.
 	else {
-		while (Current_sexp_network_packet.get_ship(sindex)) {
-			Assertion(sindex >= 0, "Illegal value for the ship index sent in multi_sexp_script_eval_multi()! Ship %d does not exist!", sindex); 
-			if (Player->objnum == Ships[sindex].objnum) {
-				success = Script_system.EvalString(s, s);
+		while (Current_sexp_network_packet.get_string(ship_name)) {
+			auto ship_entry = ship_registry_get(ship_name);
+			Assertion(ship_entry, "Illegal value for the ship index sent in multi_sexp_script_eval_multi()! Ship %s does not exist!", ship_name);
```
asarium, line 23242: Minor thing: It says `ship index` here but we are looking at a ship name so that could be adjusted here for consistency.

Goober5000: Good point.

code/parse/sexp.cpp
Goober5000: That check would actually be redundant here.  In multiplayer mode, it's not *the* player, it's *any* player.  But I've added some comments for clarification.

Issue #2423, opened 2020-05-30
FR: Minimum angle options


EatThePath:
For both $FOF: and $Spawn Angle: it would be very useful to be able to set a *minimum* angle to spawn at as well as a max. This is mainly useful for more stylized visuals, but not exclusively. It appears that there's already a version of the random vector in cone function that takes a min angle, which ought to help

PR #2441, opened 2020-06-05
Settable minimum spawn angles.

EatThePath:
Implements half of #2423, specifically the child weapon spawn angle part. I know at least one person besides me wants this, so I figure it's better to get it in asap rather than wait for me to finish deciding how much I want to match the existing more complex $fof options.

### DIFF
--- a/code/math/staticrand.cpp
+++ b/code/math/staticrand.cpp
@@ -141,15 +141,14 @@ void static_rand_cone(int num, vec3d *out, const vec3d* const in, float max_angl
 
 //generates a random vector in a cone, with a min amd max angle. 
 //Clone of vm_vec_random_cone overload of the same function, adapted to use static_randf_range
-void static_rand_cone(int num, vec3d* out, const vec3d* const in, float min_angle, float max_angle, const matrix* const orient) {
+void static_rand_cone(int num, vec3d* out, const vec3d* const in, float min_angle, float max_angle, const matrix* const orient)
+{
 	vec3d temp;
 	const matrix* rot;
 	matrix m;
 
 	if (max_angle < min_angle) {
-		auto tmp = min_angle;
-		min_angle = max_angle;
-		max_angle = tmp;
+		std::swap(min_angle, max_angle);
 	}
 
 	// get an orientation matrix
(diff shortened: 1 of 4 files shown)


JohnAFernandez:
Just a note, this is a slight break for multi, but just in missile spawn angles.  A server without this code may have different hits detected, but just for spawned weapons it would minor enough to not be a problem, IMHO, since everyone playing would want to have the same mod version.  They could possibly use an old standalone that doesn't have the code, but if you're using a standalone that doesn't have the mod installed, you're basically asking for something to break.

EatThePath:
To MP compatibility: these changes *shouldn't* affect any games not using the min angles feature, as everything else still goes through the old static_rand_cone() which I've not touched the logic of. And trying to join a modded game using those features with a client not having these features *should* be impossible.

Incidentally, while testing I noticed that static_rand_cone() is still using the old bad cone generating logic that creates a square footprint instead of a circular one. I have not fixed that in this PR, for those compatibility reasons.

code/math/staticrand.cpp
```diff
@@ -139,6 +139,38 @@ void static_rand_cone(int num, vec3d *out, vec3d *in, float max_angle, matrix *o
...
 
+//generates a random vector in a cone, with a min amd max angle. 
+//Clone of vm_vec_random_cone overload of the same function, adapted to use static_randf_range
+void static_rand_cone(int num, vec3d* out, const vec3d* const in, float min_angle, float max_angle, const matrix* const orient) {
+	vec3d temp;
+	const matrix* rot;
+	matrix m;
+
+	if (max_angle < min_angle) {
+		auto tmp = min_angle;
```
Goober5000, line 150: I think there's a SWAP macro or template function somewhere in the codebase that you could use instead.

code/math/staticrand.cpp
```diff
@@ -139,6 +139,38 @@ void static_rand_cone(int num, vec3d *out, vec3d *in, float max_angle, matrix *o
 	vm_rot_point_around_line(out, &t2, fl_radians(static_randf_range(num+2,-max_angle, max_angle)), &vmd_zero_vector, &rot->vec.uvec);
 }
 
+//generates a random vector in a cone, with a min amd max angle. 
+//Clone of vm_vec_random_cone overload of the same function, adapted to use static_randf_range
+void static_rand_cone(int num, vec3d* out, const vec3d* const in, float min_angle, float max_angle, const matrix* const orient) {
```
Goober5000, line 144: For functions, please put the brace on the following line.

code/math/staticrand.cpp
```diff
@@ -139,6 +139,38 @@ void static_rand_cone(int num, vec3d *out, vec3d *in, float max_angle, matrix *o
...
+		rot = orient;
+	}
+	else {
+		vm_vector_2_matrix(&m, in, nullptr, nullptr);
+		rot = &m;
+	}
+
+	// Get properly distributed spherical coordinates (DahBlount)
+	// This might not seem intuitive, but the min_angle is the angle that will have a larger z coordinate
+	float z = static_randf_range(num,cosf(fl_radians(max_angle)), cosf(fl_radians(min_angle))); // Take a 2-sphere slice
```
Goober5000, line 166: There are spaces missing after the first function parameter on this and the following line. /pedantic

EatThePath: I believe this and the above should be resolved now

PR #2442, opened 2020-06-05
update .clang-tidy based on Discord comments

Goober5000:
This removes `readability-avoid-const-params-in-decls` from the list of things clang checks.  Function declarations and headers should be the same, so it makes no sense to warn about `const` in one but not the other.

This also alphabetizes the exceptions for easier reading.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,20 +1,22 @@
 ---
 Checks: "-*,\
 modernize-*,\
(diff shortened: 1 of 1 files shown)


Goober5000:
All required checks in the Travis CI section passed.  Not sure why it isn't showing green, but this isn't the first PR Travis has stalled on, so I'm merging it.

PR #2443, opened 2020-06-05
refine ShipStatus::EXITED state

Goober5000:
There is a slight discontinuity where ships that are destroyed have their mission log entries added a bit before `ship_cleanup` is called.  This addresses that.


Goober5000:
With all checks passing and Travis still stalling, I'm merging this.

PR #2447, opened 2020-06-06
Fix a lab issue and add more explosions

The-E:
This adds the ability to make individual subsystems explode and fixes an issue with font rendering if a non-standard font is used as the main UI font.


asarium (approved):

Goober5000 (approved):
More explosions are always good.

Issue #2440, opened 2020-06-05
Multi: Available player orders are different based on whether the game is peer-to-peer or standalone hosted


JohnAFernandez:
What it says on the tin, standalone servers for some reason often do not let you give ai orders.  As a minor bug I have not investigated the cause, I'm posting on here because I don't want to forget.

notimaginative:
Perhaps related to http://scp.indiegames.us/mantis/view.php?id=1815?

JohnAFernandez:
Well, it turns out that it was a different cause than I thought.  It's actually the Host settings.  Standalone defaults to Team Leader/Wing Leader, and it was *that* option which was buggy because it would disallow communication if you were a wing leader instead of enabling it!  Peer to peer defaults to another setting that I cannot remember.

See the pull request for that broken line.

JohnAFernandez:
This got fixed!

PR #2450, opened 2020-06-07
Allow Team Leaders to Send Messages

JohnAFernandez:
Gotta love your retail bugs!

Function was returning 0 instead of 1 when the team captains were messaging AI.  

Yes we could replace this with a fancier check than, "does the ship name end in 1???"  But we should come up with that system and fix all the places it relies on that behavior at once.

Edit: forgot to mention, this does not affect multi compatibility at all.

Edit2: Realized != matches the function better and keeps everyone else from messaging, so that's what we should go with instead.

### DIFF
--- a/code/network/multiteamselect.cpp
+++ b/code/network/multiteamselect.cpp
@@ -418,9 +418,6 @@ void multi_ts_init_objnums();
 // assign the correct flags to the correct slots
 void multi_ts_init_flags();
 
-// get the proper team and slot index for the given ship name
-void multi_ts_get_team_and_slot(char *ship_name,int *team_index,int *slot_index, bool mantis2757switch);
-
 // handle an available ship scroll down button press
 void multi_ts_avail_scroll_down();
 
@@ -872,7 +869,7 @@ void multi_ts_assign_players_all()
 		// find a valid player ship - ignoring the ship which was assigned to the host
 		if((objp->flags[Object::Object_Flags::Player_ship]) && stricmp(Ships[objp->instance].ship_name,name_lookup) != 0){
 			// determine what team and slot this ship is				
-			multi_ts_get_team_and_slot(Ships[objp->instance].ship_name,&team_index,&slot_index, true);
+			multi_ts_get_team_and_slot(Ships[objp->instance].ship_name,&team_index,&slot_index);
 			Assert((team_index != -1) && (slot_index != -1));
 
 			// in a team vs. team situation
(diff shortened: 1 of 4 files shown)


Goober5000:
> Yes we could replace this with a fancier check than, "does the ship name end in 1???" But we should come up with that system and fix all the places it relies on that behavior at once.

The wing's `special_ship` field refers to its wing leader.  You could check the following:

```
Wings[sp->wingnum].ship_index[Wings[sp->wingnum].special_ship] == Objects[p->m_player->objnum].instance
```

I'd put that in its own function, and add checks that `sp->wingnum` is valid.  Is the "ship name end in 1" test used in other locations?

JohnAFernandez:
That looks good. I'll look for some more places where this is true.

Also, forgot to add the issue #2440 .

Goober5000:
I guess you could add two functions in ship.cpp, `ship_find_index_in_wing` which could be done by iterating through the `ship_index` array, and `ship_is_wing_leader`.  That would allow you to solve both this issue and the other one.

Note that `special_ship` was inconsistently used in the original public release.  At one point I went through the entire codebase and replaced references to `ship_index[0]` with `ship_index[special_ship]`.  So theoretically a ship other than the first one could be wing leader, but to my knowledge nobody has ever used that capability.

JohnAFernandez:
This will get worked on, as soon as I track down the last object update bug and clean up those two pull requests.

JohnAFernandez:
Still need to test to see if it 100% fixed buggy behavior, but it should be ready for review now.

JohnAFernandez:
Tested!  Works as expected.

code/network/multiteamselect.cpp
```diff
@@ -1673,69 +1670,35 @@ void multi_ts_init_objnums()
...
+void multi_ts_get_team_and_slot(char* ship_name, int* team_index, int* slot_index)
 {
-	int idx; 
 
 	// set the return values to default values
 	*team_index = -1;
 	*slot_index = -1;
 
+	// set up the ship registry pointer
+	const ship_registry_entry* ship_regp = ship_registry_get(ship_name);
```
Goober5000, line 1682: Kinda surprised that clang didn't insist on `auto` here, but that's fine with me.

code/network/multiutil.cpp
```diff
@@ -1861,19 +1860,24 @@ int multi_can_message(net_player *p)
...
 		// if the player has an invalid object #
 		if(p->m_player->objnum < 0){
 			return 0;
 		}
 
 		// check to see if he's a wingleader
-		sp = &Ships[Objects[p->m_player->objnum].instance];
-		if (sp->ship_name[strlen(sp->ship_name)-1] == '1') 
+		const ship_registry_entry* ship_regp = ship_registry_get(Ships[p->m_player->objnum].ship_name);
+		if (ship_regp == nullptr){
```
Goober5000, line 1871: This could be an Assertion because the entry should never be null here.

code/network/multiteamselect.cpp
```diff
@@ -1673,69 +1670,35 @@ void multi_ts_init_objnums()
...
-
-				// just Assert(), if this is wrong then we're pretty much screwed either way
-				Assert( (*slot_index >= 0) && (*slot_index < MAX_WSS_SLOTS) );
-			}
-		}
-	} 
-	// if we're _not_ in team vs. team mode
+		// get the slot within the wing, since there's only one wing each in team vs. team.
+		*slot_index = ship_regp->p_objp->pos_in_wing;
+	} // if we're _not_ in team vs. team mode
```
Goober5000, line 1696: I'd prefer the comment to be aligned with the `else` block, as it was before, but this is a nitpick.

code/network/multiutil.cpp
```diff
@@ -1910,19 +1913,24 @@ int multi_can_end_mission(net_player *p)
...
```
Goober5000, line 1924: Ditto on the Assertion

JohnAFernandez: Good to go?

code/network/multiutil.cpp
Goober5000: I meant that the Assertion could take the place of the if() block, because the pointer should realistically *never* be null.  If it's ever null, there is a serious bug somewhere.  Thus it's not something that you need to normally check for.

code/network/multiutil.cpp
```diff
@@ -1861,19 +1860,27 @@ int multi_can_message(net_player *p)
...
+	{
 		// if the player has an invalid object #
 		if(p->m_player->objnum < 0){
 			return 0;
 		}
 
 		// check to see if he's a wingleader
-		sp = &Ships[Objects[p->m_player->objnum].instance];
-		if (sp->ship_name[strlen(sp->ship_name)-1] == '1') 
+		const ship_registry_entry* ship_regp = ship_registry_get(Ships[p->m_player->objnum].ship_name);
```
Goober5000, line 1870: One thing I missed on my first go-round... `objnum` is an index into `Objects`, not into `Ships`.  You will need to get the object's `instance` field, as in the deleted code: `Ships[Objects[p->m_player->objnum].instance]`

code/network/multiutil.cpp
Goober5000: See above comment about the if() not being needed.

code/network/multiutil.cpp
```diff
@@ -1910,19 +1916,27 @@ int multi_can_end_mission(net_player *p)
...
```
Goober5000, line 1926: Also here.

code/network/multiutil.cpp
JohnAFernandez: Yeah, I'm not sure how I missed that, either.  Anyway, that should be fixed now.  Let me know if it needs more changes.

PR #2452, opened 2020-06-08
fix the way SEXPs check ship distances

Goober5000:
Retail has a hud_find_target_distance function that returns the distance between the player position and a bounding box.  This is useful for the HUD because it shows the distance to the perimeter of an object like a jump node or a large ship.  However, this function was re-used for the `distance` SEXP, resulting in the awkward situation where the distance is measured from one ship's center to another ship's bounding box.  This can lead to unexpected results and it also means that the SEXP returns a different distance when the arguments are swapped.

This PR deprecates the original SEXP and introduces the new SEXPs `center-distance` and `bbox-distance` which use consistent methods for both ships.  It also renames the `distance-ship-subystem` SEXP to `center-distance-subsystem` and adds a new `bbox-distance-subsystem` SEXP.


z64555:
Hm, the naming of the new sexps seems to be a bit of an issue.  I'd think `distance-to-center` and `distance-to-bbox` would be better understood as well as grouped together with an alphabetical sorting.  Should see what others think, I guess.

Goober5000:
Well, I like the original names, but I can change them.  New commit pushed.

code/model/model.h
```diff
@@ -1224,7 +1224,10 @@ void model_set_thrust(int model_num = -1, mst_info *mst = NULL);
 //   distance from eye_pos to closest_point.  0 means eye_pos is 
 //   on or inside the bounding box.
 //   Also fills in outpnt with the actual closest point.
-float model_find_closest_point(vec3d *outpnt, int model_num, int submodel_num, matrix *orient, vec3d * pos, vec3d *eye_pos);
+float model_find_closest_point(vec3d *outpnt, int model_num, int submodel_num, const matrix *orient, const vec3d *pos, const vec3d *eye_pos);
+
+// Like the above, but finds the closest two points to each other.
+float model_find_closest_points(vec3d *outpnt1, int model_num1, int submodel_num1, const matrix *orient1, const vec3d *pos1, vec3d *outpnt2, int model_num2, int submodel_num2, const matrix *orient2, const vec3d *pos2);
```
asarium, line 1230: Differing two functions by only an `s` could lead to readability issues. Maybe use something like `model_find_closest_point_on_each`?

Goober5000: I'm of the opposite opinion: the similar names make it clear that the functions are closely related.  I don't think it's likely that the functions will be confused since one of them has twice as many arguments.

asarium: Well, I don't care enough about that to not approve :stuck_out_tongue:

PR #2459, opened 2020-06-11
Allow weapons in the lab

The-E:
This reenables rendering of external weapon models in the lab.

This also adds actions to change a ship's loadout and fire guns. Note that weapons that have a firewait or require lock-on will not work here.
Also note that the lab will no longer open instantly, since we need to load all the weapon data now.

### DIFF
--- a/code/lab/lab.cpp
+++ b/code/lab/lab.cpp
@@ -160,6 +160,9 @@ bool Lab_emissive_light_override = (Cmdline_emissive != 0);
 
 fix Lab_Save_Missiontime;
 
+int Lab_fire_primaries = 0;
+int Lab_fire_secondaries = 0;
+
 // functions
 void labviewer_change_ship_lod(Tree* caller);
 void labviewer_change_ship(Tree* caller);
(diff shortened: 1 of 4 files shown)


Goober5000:
> Also note that the lab will no longer open instantly

I liked having the lab load instantly.  How big is the delay?  Would it be possible to lazily-load the weapons information only when the "show external weapons" box is checked?

The-E:
> I liked having the lab load instantly. How big is the delay? Would it be possible to lazily-load the weapons information only when the "show external weapons" box is checked?

The delay is "however long it takes to load every weapon tabled for the current mod". It's sadly unavoidable; not just for the external weapons but also for all the other weapons stuff in there. I can defer loading those things until either the "show external weapons" flag is ticked or the actions window is opened; I'm gonna try and see if I can do that in a reasonable way (the main issue being that at present, there is no way for me to do this stuff in a non-blocking manner so throwing up a little notice that things are being loaded is more tricky than it appears).

Personally, I'd rather have the load times all at the beginning; that, to me, is less disruptive than having essentially random UI actions incur a sudden wait.

Goober5000:
Putting all the load times at the beginning makes sense, except that a) there will be many times when people will use the lab but not use the weapon features, and b) there is no lab loading screen.

I like your idea about deferring loading until the flag is checked or the window is opened.  I don't think a small blocking delay will be a problem; that already happens when browsing ships in the tech room (or in the lab, for that matter) and people don't seem to mind.

Issue #2458, opened 2020-06-11
Mainhall PA Sounds Need to STAHP


MjnMixael:
2 bugs, possibly related.

1st, when the game is minimzed while you're in the mainhall the music will pause. PA sounds will continue to be selected and played endlessly in the background.

2nd, if you have a longer PA sound and switch to another campaign that has a different mainhall the first PA sound will continue to play until the end of the loop. Instead it should be stopped immediately.

Goober5000:
Same bug.  Testing a fix now.

Goober5000:
Ok, not quite the same bug.  Definitely related though.

Goober5000:
Fixed in #2464.

PR #2464, opened 2020-06-13
fix a couple things with mainhall intercom sounds

Goober5000:
As noticed in issue #2458, intercom sounds did not stop when pausing the game or loading a mission.  I fixed this by augmenting `main_hall_stop_ambient()` to also stop the intercom.

Additionally, it turns out that the intercom delays were not being calculated correctly due to a very bizarre random number formula.  I overloaded `rand32()` to add a min/max variant and used this instead.


Goober5000:
Going to try to turn off door and misc sounds as well.

PR #2474, opened 2020-06-14
Add Large Radius Option as AI Flag for Following Subsystem Path Points

wookieejedi:
Previously, the default distance for determining if a ship had finished a point in a model subsystem path was only 5. This worked fine for ships with low damp or no Newtonian physics. However, it did not work well with ships that use higher damps or Newtonian physics, because those ships continually overshoot that 5 meter distance and get stuck flying in a large circle trying to reach that point. This fix allows modders to set an AI flag to specify that the path distance check uses a larger value of 775 for points that are not the last point in the path. The value of 775 was chosen because it seemed like the average value that PCS2 made for path points when they were autogenerated. If desired I can add a game settings option that allows modders to set this value.

I did not use the radius in the POF model for the paths because the game will create a new path if none exist, thus a more global method is needed to ensure the distance is large. Also, this method was much more efficient for modders, because they do not have to set every single radius in their model. 

Also note that `aip->path_goal_dist` is only used for following points in a path for a subsystem. It is not used for bay arrivals or departures, waypoints, or docking.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -472,17 +472,17 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$use actual primary range:", AI::Profile_Flags::Use_actual_primary_range);
 
-				if (optional_string("$constant radii for subsystem path points:")) {
+				if (optional_string("$override radius for subsystem path points:")) {
 					int path_radii;
 					stuff_int(&path_radii);
-					if (path_radii > 0) {
+					if (path_radii >= 1) {
 						profile->subsystem_path_radii = path_radii;
 					} else {
-						mprintf(("Warning: \"$constant radii for subsystem path points:\" should be greater than 0 (read %i). Value will not be used. ", path_radii));
+						mprintf(("Warning: \"$override radius for subsystem path points:\" should be greater than 1 (read %i). Value will not be used. ", path_radii));
 					}
 				}
 
-				set_flag(profile, "$use model path point radii for subsystem path navigation:", AI::Profile_Flags::Use_subsystem_path_point_radii);
+				set_flag(profile, "$use POF radius for subsystem path points:", AI::Profile_Flags::Use_subsystem_path_point_radii);
 
 				profile->bay_arrive_speed_mult = 1.0f;
 				profile->bay_depart_speed_mult = 1.0f;
@@ -584,6 +584,7 @@ void ai_profile_t::reset()
     flags.reset();
 
     ai_path_mode = 0;
+	subsystem_path_radii = 0;
     bay_arrive_speed_mult = 0;
     bay_depart_speed_mult = 0;
 
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -475,7 +475,7 @@ void parse_ai_profiles_tbl(const char *filename)
 				if (optional_string("$constant radii for subsystem path points:")) {
 					int path_radii;
 					stuff_int(&path_radii);
-					if (path_radii) {
+					if (path_radii > 0) {
 						profile->subsystem_path_radii = path_radii;
 					} else {
 						mprintf(("Warning: \"$constant radii for subsystem path points:\" should be greater than 0 (read %i). Value will not be used. ", path_radii));
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -8,6 +8,7 @@
 
 
 #include "ai/ai_profiles.h"
+#include "ai/aibig.h"
 #include "def_files/def_files.h"
 #include "globalincs/pstypes.h"
 #include "localization/localize.h"
@@ -475,10 +476,10 @@ void parse_ai_profiles_tbl(const char *filename)
 				if (optional_string("$override radius for subsystem path points:")) {
 					int path_radii;
 					stuff_int(&path_radii);
-					if (path_radii >= 1) {
+					if (path_radii >= Minimum_subsystem_path_pt_dist) {
 						profile->subsystem_path_radii = path_radii;
 					} else {
-						mprintf(("Warning: \"$override radius for subsystem path points:\" should be >= 1 (read %i). Value will not be used. ", path_radii));
+						mprintf(("Warning: \"$override radius for subsystem path points:\" should be >= %i (read %i). Value will not be used. ", Minimum_subsystem_path_pt_dist, path_radii));
 					}
 				}
 
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -472,6 +472,8 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$use actual primary range:", AI::Profile_Flags::Use_actual_primary_range);
 
+				set_flag(profile, "$use model path radius:", AI::Profile_Flags::Use_large_path_radius);
+
 				profile->bay_arrive_speed_mult = 1.0f;
 				profile->bay_depart_speed_mult = 1.0f;
 				if (optional_string("$bay arrive speed multiplier:")) {
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -472,7 +472,7 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$use actual primary range:", AI::Profile_Flags::Use_actual_primary_range);
 
-				set_flag(profile, "$use model path radius:", AI::Profile_Flags::Use_large_path_radius);
+				set_flag(profile, "$use large path radius:", AI::Profile_Flags::Use_large_path_radius);
 
 				profile->bay_arrive_speed_mult = 1.0f;
 				profile->bay_depart_speed_mult = 1.0f;
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -472,7 +472,7 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$use actual primary range:", AI::Profile_Flags::Use_actual_primary_range);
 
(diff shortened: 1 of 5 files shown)


wookieejedi:
Double checked everything and all working as expected. I should note there is one difference from the request. If the path radii are supposed to be used but the point is not valid then the game will use the specified radii in `ai_profiles` if one is set.

code/ai/ai_profiles.cpp
```diff
@@ -472,6 +472,18 @@ void parse_ai_profiles_tbl(const char *filename)
...
 				set_flag(profile, "$use actual primary range:", AI::Profile_Flags::Use_actual_primary_range);
 
+				if (optional_string("$constant radii for subsystem path points:")) {
+					int path_radii;
+					stuff_int(&path_radii);
+					if (path_radii) {
+						profile->subsystem_path_radii = path_radii;
+					} else {
+						mprintf(("Warning: \"$constant radii for subsystem path points:\" should be greater than 0 (read %i). Value will not be used. ", path_radii));
+					}
```
Goober5000, line 482: With this if() block, any nonzero radius, even if negative, will be used.  Did you mean to check `if (path_radii > 0)`?

code/ai/aibig.cpp
```diff
@@ -457,13 +456,46 @@ int ai_big_maybe_follow_subsys_path(int do_dot_check)
...
 				path_done=1;
 			}
+			// if not done then set path distances based on specified options
+			else {
+				bool found_point = false;
+				// 1. use radii of path points if specified
+				if (aip->ai_profile_flags[AI::Profile_Flags::Use_subsystem_path_point_radii]) {
+					// need a valid path and point in model to use pof specified radius
+					int path_num = aip->targeted_subsys->system_info->path_num;
+					if ((0 <= path_num) && (path_num <= pm_t->n_paths)) {
```
Goober5000, line 477: Should this be `< n_paths`, not `<=`?

code/ai/aibig.cpp
```diff
@@ -457,13 +456,46 @@ int ai_big_maybe_follow_subsys_path(int do_dot_check)
...
+						// need a valid place in the path
+						if ((0 <= aip->path_cur) && (aip->path_cur <= aip->path_length)) {
```
Goober5000, line 479: Ditto for `path_length`

code/ai/aibig.cpp
```diff
@@ -457,13 +456,46 @@ int ai_big_maybe_follow_subsys_path(int do_dot_check)
...
+							float path_point_radius = pm_t->paths[path_num].verts[aip->path_cur].radius;
+							int goal_dist = fl2i(path_point_radius);
+							// only set if > 0
+							if (goal_dist) {
```
Goober5000, line 483: Again this will test any nonzero value, not any positive value

code/ai/ai_profiles.cpp
wookieejedi: Correct, it should be > 0. Thanks!

code/ai/aibig.cpp
wookieejedi: Ah correct, it should be since indexing starts at 0 not 1.

wookieejedi:
Thanks for the review, just updated.

code/ai/ai_profiles.cpp
```diff
@@ -472,6 +472,18 @@ void parse_ai_profiles_tbl(const char *filename)
...
+					int path_radii;
+					stuff_int(&path_radii);
+					if (path_radii > 0) {
+						profile->subsystem_path_radii = path_radii;
+					} else {
+						mprintf(("Warning: \"$constant radii for subsystem path points:\" should be greater than 0 (read %i). Value will not be used. ", path_radii));
+					}
+				}
+
+				set_flag(profile, "$use model path point radii for subsystem path navigation:", AI::Profile_Flags::Use_subsystem_path_point_radii);
```
Goober5000, line 485: I would suggest being consistent with the option names here: you have "subsystem path points" vs. "subsystem path navigation".

Also "constant radii" is IMO not very intuitive.  And "radii" is technically correct but it's not clear (to the lazy reader) why you're using the plural.

I suggest "override radius for model path points" for the first option and "use POF radius for model path points" for the second option.  (And they should be documented on the wiki page.)

The options *are* applicable for both docking paths and subsystem paths, right?

code/ai/aibig.cpp
```diff
@@ -457,13 +456,46 @@ int ai_big_maybe_follow_subsys_path(int do_dot_check)
...
+			// distance setting --wookieejedi
 			if ( aip->path_cur >= (aip->path_start+aip->path_length-1) ) {
+				// register that path is done and reset goal distance to default
+				aip->path_goal_dist = default_dist;
 				path_done=1;
 			}
+			// if not done then set path distances based on specified options
+			else {
+				bool found_point = false;
+				// 1. use radii of path points if specified
```
Goober5000, line 473: Good documentation and clear logic here for the 1, 2, 3.  Nice work! :)

code/ai/aibig.cpp
```diff
@@ -457,13 +456,46 @@ int ai_big_maybe_follow_subsys_path(int do_dot_check)
...
+				if (aip->ai_profile_flags[AI::Profile_Flags::Use_subsystem_path_point_radii]) {
+					// need a valid path and point in model to use pof specified radius
+					int path_num = aip->targeted_subsys->system_info->path_num;
+					if ((0 <= path_num) && (path_num < pm_t->n_paths)) {
+						// need a valid place in the path
+						if ((0 <= aip->path_cur) && (aip->path_cur < aip->path_length)) {
+							float path_point_radius = pm_t->paths[path_num].verts[aip->path_cur].radius;
+							int goal_dist = fl2i(path_point_radius);
+							// only set if > 0
```
Goober5000, line 482: After some thought, I suggest instead only setting it if the distance is >=1.  The retail POFs, especially, have very tiny radii in some cases.  If 5 is bad, .1 will be much worse.

code/ai/aibig.cpp
```diff
@@ -457,13 +456,46 @@ int ai_big_maybe_follow_subsys_path(int do_dot_check)
...
+							if (goal_dist > 0) {
+								aip->path_goal_dist = goal_dist;
+								found_point = true;
+							}
+						}
+					}
+				}
+				// 2. use radii value in ai_profiles table if specified and no valid point found
+				if (!found_point && The_mission.ai_profile->subsystem_path_radii) {
```
Goober5000, line 491: `subsystem_path_radii` is never initialized to a default value, so if there is no ai_profiles option it will be a random number.

code/ai/aibig.cpp
```diff
@@ -457,13 +456,46 @@ int ai_big_maybe_follow_subsys_path(int do_dot_check)
 		if ( aip->path_length > 0 ) {
 			int path_done=0;
 			int in_view=0;
+			int default_dist = 5;
```
Goober5000, line 459: Since `default_dist` never actually changes, put it in the header file and make it a `#define` or a `const`.

code/ai/ai_profiles.cpp
wookieejedi: ATM these options are only applicable for use in the function `ai_big_maybe_follow_subsys_path` which is only called to attack subsystems (as mentioned in the initial post of this comment :) ). Docking did not have that problem of ships not being able to follow/complete paths because the AI slows down considerably when docking and damp values seem not to be used (similar to bay paths). 

Good suggestions for the name changes.

code/ai/aibig.cpp
wookieejedi: If it's a const int I thought that this line would then not work?
`aip->path_goal_dist = Default_subsystem_path_pt_dist;`

code/ai/aibig.cpp
wookieejedi: Good idea.

code/ai/aibig.cpp
wookieejedi: Thanks!

code/ai/aibig.cpp
wookieejedi: I reckon setting a default value in`ai_profiles.h` like  `int subsystem_path_radii = -1;` is the best way to fix that?

code/ai/ai_profiles.cpp
Goober5000: Yeah, my suggestion of "model path points" was with the assumption that they would be used for all paths.  So "subsystem path points" is fine.

code/ai/aibig.cpp
Goober5000: Not in the header file.  Look in `ai_profile_t::reset()`.

code/ai/aibig.cpp
Goober5000: I see nothing wrong with that.  Why do you think it wouldn't work?

If you mean `const` vs. `non-const`, you're working with values here, not pointers.  Since the value is copied, the original variable won't be affected if `aip->path_goal_dist` subsequently changes.

code/ai/aibig.cpp
```diff
@@ -461,9 +460,41 @@ int ai_big_maybe_follow_subsys_path(int do_dot_check)
...
+				// 1. use radii of path points if specified
+				if (aip->ai_profile_flags[AI::Profile_Flags::Use_subsystem_path_point_radii]) {
+					// need a valid path and point in model to use pof specified radius
+					int path_num = aip->targeted_subsys->system_info->path_num;
+					if ((0 <= path_num) && (path_num < pm_t->n_paths)) {
+						// need a valid place in the path
+						if ((0 <= aip->path_cur) && (aip->path_cur < aip->path_length)) {
+							float path_point_radius = pm_t->paths[path_num].verts[aip->path_cur].radius;
+							int goal_dist = fl2i(path_point_radius);
+							// only set if > 1
```
Goober5000, line 481: Very minor nitpick, but the comment and code don't agree.

wookieejedi: Good catch.

PR #2480, opened 2020-06-16
Ammo sexps

Goober5000:
upgrade and merge the various ammo sexps

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -17144,7 +17144,7 @@ void sexp_set_ammo_sub(ship_weapon *swp, int requested_bank, int requested_ammo,
 
 	// Can only set one bank at a time. Check that someone hasn't asked for the contents of all 
 	// the banks or a non-existant bank
-	if ((requested_bank > num_banks) || requested_bank < 0)
+	if ((requested_bank >= num_banks) || requested_bank < 0)
 		return;
 
 	if (primary)
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -17038,511 +17038,280 @@ int sexp_shield_quad_low(int node)
 	return SEXP_FALSE;	
 }
 
-// Goober5000
-int sexp_primary_ammo_pct(int node)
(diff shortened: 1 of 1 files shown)


asarium:
I think I'll need a bit more context to know what exactly this does before reviewing it.

Goober5000:
It does two main things:

1) Upgrades the sexp to use `eval_ship()` on the ship nodes instead of the old ship lookup procedure, similar to the other recent sexp upgrade PRs
2) Combines the primary/secondary, pilot/turret, count/percent variants of the get/set ammo sexps to simplify the functions and remove a lot of duplicate code.

I admit there are a lot of added/removed lines here, and it may be easier to just read the new versions of the functions rather than combing through the diff.

code/parse/sexp.cpp
```diff
@@ -17038,511 +17038,280 @@ int sexp_shield_quad_low(int node)
...
-	shipp = &Ships[ship_index];
-	
 	// Can only set one bank at a time. Check that someone hasn't asked for the contents of all 
 	// the banks or a non-existant bank
-	if ((requested_bank > shipp->weapons.num_primary_banks) || requested_bank < 0)
-	{
-		return ;
-	}
+	if ((requested_bank > num_banks) || requested_bank < 0)
+		return;
```
asarium, line 17148: The old version didn't have this but this case would indicate that there is something wrong in the mission, right? If so, then it would be nice for the mission designer to have a warning message here to explain why nothing is happening.

Same applies to the ballistics check below.

Goober5000: That actually points to a legit issue; it should be `>= num_banks`. :)

Unfortunately a warning message isn't really possible here from a conceptual standpoint.  Getting ammo is more subtle and a warning would be more appropriate there; but the most important edge case is also a valid use of that sexp.  A bank number on one ship can be the number of banks on another ship, and there's no way to tell what the mission designer intended.  So putting a warning on `get` won't work, and putting a warning only on `set` would result in a false sense of security on `get`.

asarium: Uhh, yes. I totally noticed that!

Yeah, those are good reasons for not including a warning. Was just a suggestion anyway.

PR #2481, opened 2020-06-17
Fix new version of `disarmed-delay`

wookieejedi:
Boolean needs to be false not true since `disarmed-delay` is not checking disabled status. Tested fix and confirms it works.


wookieejedi:
Noting that this was changing a single bool value and the build failures seem to related to the known recent issues of build checks failing.

PR #2489, opened 2020-06-20
Add Expanded Option of Smart Primary Selection

wookieejedi:
With gradual code changes with adding armor values, it seems that primary selection was never updated to take those into account (nor did it include checks to the subsystem or shockwaves). This makes AI not really choose the best weapon for a situation if armor tables are used heavily to determine balance. This PR adds an updated logic in `ai_select_primary_weapon` along with an `ai_profiles` flag to turn it on. I'd also be open to allowing the threshold values used in the function as settable for the modder in `ai_profiles`.

### DIFF
--- a/code/ai/ai_flags.h
+++ b/code/ai/ai_flags.h
@@ -115,6 +115,7 @@ namespace AI {
         Shockwaves_damage_small_ship_subsystems,
         Smart_afterburner_management,
         Smart_primary_weapon_selection,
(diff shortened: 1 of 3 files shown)


wookieejedi:
Lots of commits after initial approval along with not high demand for this PR leads me to closing this. I'll save the relevant code and might reopen with updated PR in future if necessary.

PR #2494, opened 2020-06-21
Expose 'Damp' to Scripting

wookieejedi:
Many physics values are exposed to scripting, including rotational damp. This PR exposes the linear damp to scripting. Note, in the code base linear damp is called `side_slip_time_const`.


Goober5000:
I thought the derivative of velocity was acceleration?

wookieejedi:
You're right, I updated the info string.

PR #2495, opened 2020-06-22
fix a few things with subsystem sexp operations

Goober5000:
1) Something wookieejedi noticed: destroying all turrets didn't trigger the DISARMED mission log event.  This is because a code ordering decision intended to accommodate the destroyed check caused the hitpoints to not be correct when they were checked within `do_subobj_destroyed_stuff`.  Instead I changed the blow-up check to use the same code for set-strength as for sabotage-strength.  This handles destruction properly and also tracks hitpoints properly.
2) Aggregate points were being recalculated for every subsystem operation where they only needed to be recalculated at the end of the sexp, so I moved that.


wookieejedi:
I'll also thank topace for spotting this and passing it along to me. : )

PR #2506, opened 2020-06-23
upgrade sexps affecting rotating submodels

Goober5000:


Goober5000:
All checks passed; dunno why it's not completing.

PR #2513, opened 2020-06-25
giant refactor of sexps to use caching for ships and wings

Goober5000:
This upgrades the remaining sexps and removes the temporary API stuff.  It was originally the branch containing all the PRs that have been merged over the past few weeks; it should be pruned down to bite-size now.


wookieejedi:
I'd imagine that @JohnAFernandez and @notimaginative would be interested in this PR since it changes more multi related functions.

Goober5000:
The changes aren't consequential as far as multi is concerned -- they just make the argument passing more robust and efficient.

PR #2517, opened 2020-06-26
Rewrite the lab

The-E:
The main goal of this rewrite is to increase maintainability of the Lab by breaking the monolithic lab.cpp into several smaller subsections.

The basic structure here is as follows:
A central LabManager instance (LMGR) is responsible for opening and closing the Lab screen as well as handling all the onFrame business that needs handling. It keeps track of the global state, and notifies all subdialogs if it changes so that they can update themselves accordingly.


code/lab/labv2.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "gamesequence/gamesequence.h"
+
+LabManager* LMGR;
+
+void lab_init() {
+	gr_set_clear_color(0, 0, 0);
+
+	LMGR = new LabManager();
+
+	Game_mode |= GM_LAB;
```
asarium, line 13: Is this reset anywhere?

asarium (changes requested):
To be honest, this is quite a lot of code to review and I only barely looked at the new lab code. However, since there was a test build with this (at least I thought I saw one on the forums) and these are pretty self-contained it should be ok to merge this and see if any issues arise later.

Marking as "Request changes" since there is one question to clarify but otherwise should be ok.

code/lab/labv2.cpp
The-E: Yes. When the user leaves the Lab, LabManager::close is called which removes the GM_LAB flag again.

asarium: Ah, ok. A bit confusing to do the "set" part here and the "unset" in a different location but nothing critical.

asarium (approved):

PR #2519, opened 2020-06-27
Add system for adding SEXPs externally

asarium:
This adds a "clear-debris" SEXP with a new system for declaring SEXPs
outside of sexp.cpp. This makes it easier to keep code modules self
contained since they do not need to export previously private
functionality just for having missiong being able to interface with it.

This is mostly a proof-of-concept which only supports basic parameter
types (booleans and integers).

As for the reasoning why this could just be added to sexp.cpp: It's already 1MB big. Let's not add more...

### DIFF
--- a/code/parse/sexp/EngineSEXP.h
+++ b/code/parse/sexp/EngineSEXP.h
@@ -1,0 +1,198 @@
+#pragma once
+
+#include "globalincs/pstypes.h"
+
+#include "DynamicSEXP.h"
+#include "SEXPParameterExtractor.h"
+
+namespace sexp {
+
+class EngineSEXP;
+
+struct dummy_return {
+};
+
+using EngineSexpAction = std::function<int(SEXPParameterExtractor* extractor)>;
+
+class EngineSEXPFactory {
+	std::unique_ptr<EngineSEXP> m_sexp;
+
+	SCP_string _helpText;
+
+	int _category = -1;
+
+	int _subcategory = -1;
+	SCP_string _subcategoryName;
+
+	int _returnType = -1;
+
+	EngineSexpAction _action;
+
+	struct argument {
+		int type = -1;
+		SCP_string help_text;
+
+		bool optional_marker = false;
+		bool varargs_marker = false;
+	};
+
+	SCP_vector<argument> _arguments;
+
+	class ArgumentListBuilder {
+		EngineSEXPFactory* _parent;
+
+	  public:
+		ArgumentListBuilder(EngineSEXPFactory* parent);
+
+		/**
+		 * @brief Adds a parameter of the specified type with help text.
+		 * @param type The type of this parameter
+		 * @param help_text A helpful text for this parameter
+		 * @return The arglist builder
+		 */
+		ArgumentListBuilder& arg(int type, SCP_string help_text);
+		/**
+		 * @brief Specifies that all following arguments are optional
+		 * @return The arglist builder
+		 */
+		ArgumentListBuilder& beginOptional();
+		/**
+		 * @brief Begins the part of the parameter list which can be repeated
+		 * @return The arglist builder
+		 */
+		ArgumentListBuilder& beginVarargs();
+
+		/**
+		 * @brief Finishes the argument list
+		 * @return The parent engine SEXP factory
+		 */
+		EngineSEXPFactory& finishArguments();
+	};
+
+	friend ArgumentListBuilder;
+
+  public:
+	EngineSEXPFactory(std::unique_ptr<EngineSEXP> sexp);
+	virtual ~EngineSEXPFactory();
+
+	EngineSEXPFactory(EngineSEXPFactory&&) noexcept = default;
+	EngineSEXPFactory& operator=(EngineSEXPFactory&&) noexcept = default;
+
+	/**
+	 * @brief Specifies the help text of the SEXP
+	 * @param text The help text
+	 * @return The factory
+	 */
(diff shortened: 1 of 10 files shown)


code/debris/debris.cpp
```diff
@@ -79,6 +79,42 @@ static void debris_start_death_roll(object *debris_obj, debris *debris_p)
...
+				debris_start_death_roll(&Objects[debris_entry.objnum], &debris_entry);
+			}
+		} else {
+			if (nonHullDebris) {
+				debris_start_death_roll(&Objects[debris_entry.objnum], &debris_entry);
+			}
+		}
+	}
+
+	return 0;
```
Goober5000, line 101: 0 is equivalent to `SEXP_FALSE`, but all "void-ish" sexp operators return `SEXP_TRUE`, or 1.

asarium: This value gets returned directly to the SEXP system so that is very relevant. Changed it to `SEXP_TRUE`, thanks for the hint.

code/debris/debris.cpp
```diff
@@ -79,6 +79,42 @@ static void debris_start_death_roll(object *debris_obj, debris *debris_p)
...
+		if (!(debris_entry.flags[Debris_Flags::Used]))
+			continue;
+
+		if (debris_entry.is_hull) {
+			if (hullDebris) {
+				debris_start_death_roll(&Objects[debris_entry.objnum], &debris_entry);
+			}
+		} else {
+			if (nonHullDebris) {
+				debris_start_death_roll(&Objects[debris_entry.objnum], &debris_entry);
```
Goober5000, line 96: If this sexp is supposed to behave just like the `clear-weapons` sexp, then don't call `debris_start_death_roll`.  Instead just set the `Should_be_dead` object flag.

code/parse/sexp/EngineSEXP.h
```diff
@@ -0,0 +1,198 @@
...
+	  public:
+		ArgumentListBuilder(EngineSEXPFactory* parent);
+
+		/**
+		 * @brief Adds a parameter of the specified type with help text.
+		 * @param type The type of this parameter
+		 * @param help_text A helpful text for this parameter
+		 * @return The arglist builder
+		 */
+		ArgumentListBuilder& arg(int type, SCP_string help_text);
```
Goober5000, line 53: Should this be `const SCP_string &`?

code/parse/sexp/EngineSEXP.h
```diff
@@ -0,0 +1,198 @@
...
+
+	EngineSEXPFactory(EngineSEXPFactory&&) noexcept = default;
+	EngineSEXPFactory& operator=(EngineSEXPFactory&&) noexcept = default;
+
+	/**
+	 * @brief Specifies the help text of the SEXP
+	 * @param text The help text
+	 * @return The factory
+	 */
+	EngineSEXPFactory& helpText(SCP_string text);
```
Goober5000, line 86: And here

code/parse/sexp/EngineSEXP.h
```diff
@@ -0,0 +1,198 @@
...
+	 * @param name The name of the SEXP
+	 * @return The engine SEXP factory
+	 */
+	static EngineSEXPFactory create(const SCP_string& name);
+
+  private:
+	void setCategory(int category);
+	void setSubcategory(int subcategory);
+	void setSubcategoryName(SCP_string subcategory);
+	void setHelpText(SCP_string helpText);
```
Goober5000, line 176: And here

code/parse/sexp/EngineSEXP.h
```diff
@@ -0,0 +1,198 @@
...
+	void setReturnType(int returnType);
+	void initArguments(int minArgs, int maxArgs, SCP_vector<int> argTypes, SCP_vector<int> varargsTypes);
```
Goober5000, line 178: And here (`const SCP_vector<int> &`)

code/debris/debris.cpp
asarium: `debris_start_death_roll` has the advantage that it takes care of sending the relevant multi packet so I chose that for "automatic" multi support.

code/parse/sexp/EngineSEXP.h
asarium: No, that is intentional since the string is `std::move`d inside and this avoid a copy operation.

code/parse/sexp/EngineSEXP.h
asarium: Same as above.

code/parse/sexp/EngineSEXP.h
asarium: Same as above.

code/parse/sexp/EngineSEXP.h
asarium: Same as above.

code/debris/debris.cpp
Goober5000: Yeah, but that function also creates a fireball and plays an explosion sound, two things you don't want to happen when you're just deleting all debris.

If I were writing this sexp I would just overload the functions that implement `remove-weapons`; the process is identical except for the `weapon_info_index` check.  Come to think of it, why not change this PR into a pure external sexp system, and I'll create a separate PR to add the `clear-debris` sexp?

asarium: I'm fine with not having that sexp and removed it from the PR.

Goober5000: Ok.

PR #2522, opened 2020-06-28
small cleanup in  sexp.cpp

CasparKielwein:
I looked for possibilities to reduce the size of the parse/sexp.cpp file.
I did not fine any obvious code duplication, so here is a small PR with changes to reduce the amount of warnings from clang-tidy.

Best reviewed on a per commit basis.
The last commit contains only white-space changes which replaces all occurrences of 4 white-space with a tab, to make the indentation consistent within the file.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -17142,6 +17142,20 @@ void sexp_set_countermeasures(int node)
 	Current_sexp_network_packet.end_callback();
 }
 
+void multi_sexp_set_countermeasures()
+{
+	int num_cmeasures = 0;
+	ship *shipp;
+
+	Current_sexp_network_packet.get_ship(shipp);
+	if (shipp == nullptr) {
+		return;
+	}
+	if (Current_sexp_network_packet.get_int(num_cmeasures)) {
+		shipp->cmeasure_count = num_cmeasures;
+	}
+}
+
 // KeldorKatarn - Locks or unlocks the afterburner on the requested ship
 void sexp_deal_with_afterburner_lock (int node, bool lock)
 {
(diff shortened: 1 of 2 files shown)


code/parse/sexp.cpp
```diff
@@ -17145,20 +17142,6 @@ void sexp_set_countermeasures(int node)
...
-
-	Current_sexp_network_packet.get_ship(shipp);
-	if (shipp == NULL) {
-		return;
-	}
-	if (Current_sexp_network_packet.get_int(num_cmeasures)) {
-		shipp->cmeasure_count = num_cmeasures;
-	}
-}
-
```
Goober5000, line 17161: This function should not be deleted.  Instead, `multi_sexp_set_countermeasures` should be added amidst the rest of the multi sexp calls.

code/parse/sexp.cpp
```diff
@@ -21817,14 +21799,6 @@ void sexp_reset_time_compression()
 	 Current_sexp_network_packet.end_callback();
 }
 
-void multi_sexp_reset_time_compression() 
-{
-	set_time_compression(1);
-	lock_time_compression(false);
-}
```
Goober5000, line 21824: This should also be added to the multi sexp list.

code/parse/sexp.cpp
CasparKielwein: can you give me a hint what i need to do to add it to the appropriate list?

Goober5000: I thought it wouldn't be hard to figure out, but sure.  Look in `multi_sexp_eval` where all the other multi_sexp_* functions are called.  Add the appropriate `case` statements and function calls to the big `switch` block.

CasparKielwein:
Failure of the appveyor build seems unrelated.
It is caused by one run of the compiler running out of memory.

Goober5000:
Ok, I restarted it.

PR #2544, opened 2020-07-04
Add multi support to external SEXPs

asarium:
This adds (mostly) automatic multi support to the external SEXP system.
This will record the parameter values that were evaluated for the master
and automatically send them to the client via the standard SEXP packet
system.

For SEXPs where the behavior does not differ between master and client
there needs to be no special handling of the multi case in the SEXP
action code.

FYI, this moves `set-motion-debris-override` to the external engine SEXP and makes it compatible with multi with this new system.


code/parse/sexp.cpp
```diff
@@ -24452,11 +24451,6 @@ int eval_sexp(int cur_node, int referenced_node)
 				sexp_supernova_stop(node);
 				break;
 
-			case OP_SET_MOTION_DEBRIS:
-				sexp_val = SEXP_TRUE;
-				sexp_set_motion_debris(node);
```
Goober5000, line 24457: It looks like you left `sexp_set_motion_debris` in the code even though you deleted its invocation.

code/starfield/starfield.cpp
```diff
@@ -181,6 +182,24 @@ int Num_debris_nebula = 0;
...
+
+const auto __UNUSED SetMotionDebrisSexp =
+	sexp::EngineSEXP::create("set-motion-debris-override")
+		.helpText("Controls whether or not motion debris should be active.\r\n"
+				  "\tThis overrides any choice made by the user through the -nomotiondebris commandline flag.")
+		.category(OP_CATEGORY_CHANGE2)
+		.subcategory(CHANGE_SUBCATEGORY_CUTSCENES)
+		.returnType(OPR_NULL)
+		.beginArgList()
+		.arg(OPF_BOOL, "Sets the new state of the motion debris")
```
Goober5000, line 198: Add the boolean information (true to disable, false to enable) to the help text, since that's counterintuitive.

code/parse/sexp.h
```diff
@@ -734,7 +734,6 @@ class waypoint_list;
 #define OP_GET_ETS_VALUE					(0x0022 | OP_CATEGORY_CHANGE2 | OP_NONCAMPAIGN_FLAG)	// niffiwan
 #define OP_SET_ETS_VALUES					(0x0023 | OP_CATEGORY_CHANGE2 | OP_NONCAMPAIGN_FLAG)	// niffiwan
 #define OP_CALL_SSM_STRIKE					(0x0024 | OP_CATEGORY_CHANGE2 | OP_NONCAMPAIGN_FLAG) // X3N0-Life-Form
-#define OP_SET_MOTION_DEBRIS				(0x0025 | OP_CATEGORY_CHANGE2 | OP_NONCAMPAIGN_FLAG)    // The E
```
Goober5000, line 737: I recommend renumbering the subsequent sexps, or moving the `OP_CHANGE_BACKGROUND` definition, so that we don't leave any gaps in the operator numbering sequence.

code/parse/sexp/sexp_lookup.cpp
```diff
@@ -173,6 +173,14 @@ void dynamic_sexp_init()
 	auto& global = globals();
 	global.initialized = true;
 
+	// These sexps were added in a non determinate order (across different platforms and compilers). To ensure a
+	// consistent order for the same build across different platforms, we make the order consistent by sorting. This
+	// should make it possible to use dynamic SEXP in multi contexts.
```
Goober5000, line 178: If the server and client are running different builds, and a sexp on one computer is missing on another computer, then the subsequent sexps will have different operator numbers.  This will prevent different builds from running any multiplayer missions even if neither one uses the actual sexp that is missing.

Goober5000 (commented):
Interesting approach.  I commented on some things I noticed.

The major potential problem here is that a lot of multi sexps don't need to process the same data on the client as on the server.  Look at `destroy_subsys_instantly` for example: the server sexp analyzes two different types of arguments and does a lot of argument checking, but the multi portion just receives a list of subsystem indexes.  In a lot of cases, the server can filter out invalid calls (e.g. missing ships or out-of-range numbers) before the client even gets any packet information.  Can I assume that, in these situations, the "disableMulti" flag will be set and the code can use the regular SEXP multi packet system?

asarium:
Yes, if a SEXP chooses to do all the multi stuff itself then `disableMulti` is what is needed. In that case the SEXP action is still called on the client so it can all be be handled in the same function.

code/starfield/starfield.cpp
asarium: Fixed.

code/parse/sexp.h
asarium: Good point, fixed. The dynamic SEXP system is also not really set up to handle holes in the operators so this is also a technical necessity.

code/parse/sexp/sexp_lookup.cpp
asarium: Indeed, that is an issue I had not considered before.
What do you think about sending the string name of the operator instead of the number? Then it could be resolved back to the operator number on the client without running into inconsistency issues.
The next problem with that would be that this should only be done for the dynamic SEXPs though and it would still break the multi protocol with old builds but that shouldn't be too big of a problem.

JohnAFernandez: You could always make it part of the impending multi bump as well.

Goober5000: Sending the sexp name would solve the problem, but I would rather not do that because it would roll back one of the major optimizations of the sexp system.  I can think of two good solutions, one simple and one more complex:

1) At mission load, the server sends the number of sexp operators.  If this number is different on any client, mission load fails.  (We can be virtually certain that two builds with the same *number* of operators will have the same *set* of operators.)

2) At mission load, create a hash map of operator numbers and send that to all clients.

I think I favor option #1.  Most of the time there will not be any difference in the number of operators, so a mapping would increase overhead (and it would also be a pain to code).  It would be a good compromise in that it allows players to use different builds while making sure they are still compatible.

wookieejedi:
Any update on this one?

asarium:
No, haven't touched this in a while.

PR #2555, opened 2020-07-08
Fix physics whack for docked ships

### LINKED ISSUE #2525: Physics 'Whack' From Weapons Applying to Docked Parent Ships
Whack from weapons such as morning stars, shockwaves and beams will toss about small ships like an Isis. However, even when docked with a very large ship, like an Orion, both ships will whack with the same vigor if the Isis is whacked, resulting in the Orion wildly smashing any unfortunate ship in the vicinity. It's uh, pretty wack.

Baezon:
Fixes #2525 . Basically this will, after calculating the center of mass of the docked ships and their aggregate MOI, 'whack' each individual ship with corrected rotational and linear velocities so that all ships in the aggregate move along with what the aggregate would rotate and move with. Also replaces all outside callings of physics_apply_whack with ship_apply_whack, so this checking can be done. 

In general, physics_apply_whack shouldn't be called in lieu of ship_apply_whack or you'll get the previous issues. Also physics_apply_whack is now properly passed a world position, instead of a local position as it was doing before, fixing some weird whack behavior you could get and appeared to be retail!

Of note: this does NOT alter shockwaves or collisions, only weapons fire! They will continue to exhibit the bad behavior, and fixing them is on our to-do list.

### DIFF
--- a/code/object/objectdock.cpp
+++ b/code/object/objectdock.cpp
@@ -29,7 +29,8 @@ void dock_calc_max_cross_sectional_radius_squared_perpendicular_to_line_helper(o
 void dock_calc_max_semilatus_rectum_squared_parallel_to_directrix_helper(object *objp, dock_function_info *infop);
 void dock_find_max_speed_helper(object *objp, dock_function_info *infop);
 void dock_find_max_fspeed_helper(object *objp, dock_function_info *infop);
-
+void dock_calc_inv_total_moi_helper(object* objp, dock_function_info* infop);
+void dock_whack_all_docked_objects_helper(object* objp, dock_function_info* infop);
 
 // management prototypes
 
@@ -186,7 +187,7 @@ void dock_calc_docked_extents(vec3d *mins, vec3d *maxs, object *objp)
 	dock_evaluate_all_docked_objects(objp, &dfi, dock_calc_docked_mins_maxs_helper);
 }
 
-void dock_calc_docked_center_of_mass(vec3d *dest, object *objp)
+float dock_calc_docked_center_of_mass(vec3d *dest, object *objp)
 {
 	Assert(dest != NULL);
 	Assert(objp != NULL);
@@ -199,7 +200,9 @@ void dock_calc_docked_center_of_mass(vec3d *dest, object *objp)
 	dock_evaluate_all_docked_objects(objp, &dfi, dock_calc_docked_center_of_mass_helper);
 
 	// overall center of mass = weighted sum of centers of mass divided by total mass
-	vm_vec_scale(dest, (1.0f / dfi.maintained_variables.float_value));
+	float total_mass = dfi.maintained_variables.float_value;
+	vm_vec_scale(dest, (1.0f / total_mass));
+	return total_mass;
 }
 
 float dock_calc_total_docked_mass(object *objp)
@@ -337,6 +340,65 @@ float dock_calc_docked_speed(object *objp)
 	return dfi.maintained_variables.float_value;
 }
 
+void dock_calc_total_moi(matrix* dest, object* objp, vec3d *center)
+{
+	Assert(dest != NULL);
+	Assert(objp != NULL);
+
+	matrix accum, local_moi;
+	vm_mat_zero(&accum);
+
+	dock_function_info dfi;
+	dfi.parameter_variables.vecp_value = center;
+	dfi.maintained_variables.matrix_value = dest;
+
+	dock_evaluate_all_docked_objects(objp, &dfi, dock_calc_inv_total_moi_helper);
+}
+
+void dock_whack_all_docked_objects(vec3d* force, vec3d* world_hit_pos, object* objp)
+{
+	//	Detect null vector.
+	if (whack_below_limit(force))
+		return;
+
+	vec3d world_center_of_mass;
+
+	// calc world hit pos of the hit ship
+	vm_vec_add2(world_hit_pos, &objp->pos);
+
+	// calc overall world center-of-mass of all ships
+	float total_mass = dock_calc_docked_center_of_mass(&world_center_of_mass, objp);
+
+	// the new hitpos is the vector from world center-of-mass to world hitpos
+	vm_vec_sub2(world_hit_pos, &world_center_of_mass);
+
+	matrix moi, local_moi, inv_moi;
+
+	// calculate the effective inverse MOI for the docked composite object about its center of mass
+	dock_calc_total_moi(&moi, objp, &world_center_of_mass);
+
+	vm_inverse_matrix(&inv_moi, &moi);
+
+	vec3d torque;
+	vm_vec_cross(&torque, world_hit_pos, force);
+
+	vec3d delta_rotvel;
+	vm_vec_rotate(&delta_rotvel, &torque, &inv_moi);
+
+	vec3d delta_vel = *force * (1.0f / total_mass);
+	// up to this point is all we would need for a single object but since its a composite object
+	// the individual objects will need to take this data and calculate how it would affect 
+	// them given their orientations and offsets
+
+	dock_function_info dfi;
+	dfi.parameter_variables.float_value = vm_vec_mag(force);
+	dfi.parameter_variables.vecp_value = &delta_rotvel;
+	dfi.parameter_variables.vecp_value2 = &delta_vel;
+	dfi.maintained_variables.vecp_value = &world_center_of_mass;
+
+	//pass on the data to the docked objects so they can whack themselves accordingly
+	dock_evaluate_all_docked_objects(objp, &dfi, dock_whack_all_docked_objects_helper);
+}
 
 // functions to deal with all docked ships anywhere
 // ---------------------------------------------------------------------------------------------------------------
@@ -733,6 +795,50 @@ void object_remove_arriving_stage2_ndl_flag_helper(object *objp, dock_function_i
 		Ships[objp->instance].flags.remove(Ship::Ship_Flags::Arriving_stage_2_dock_follower);
 }
 
+void dock_calc_inv_total_moi_helper(object* objp, dock_function_info* infop)
+{
+	matrix local_moi, unorient, world_moi;
+	vm_inverse_matrix(&local_moi, &objp->phys_info.I_body_inv);
+	vm_copy_transpose(&unorient, &objp->orient);
+	vm_matrix_x_matrix(&world_moi, &local_moi, &unorient);
+	float mass = objp->phys_info.mass;
+	vec3d pos = objp->pos;
+	pos -= *infop->parameter_variables.vecp_value;
+	// moment of inertia for a point mass: I_xx = m(y^2+z^2), I_xy = -mxy, etc
+	world_moi.a2d[0][0] += mass * (pos.xyz.y * pos.xyz.y + pos.xyz.z * pos.xyz.z);
+	world_moi.a2d[1][1] += mass * (pos.xyz.x * pos.xyz.x + pos.xyz.z * pos.xyz.z);
+	world_moi.a2d[2][2] += mass * (pos.xyz.x * pos.xyz.x + pos.xyz.y * pos.xyz.y);
+	world_moi.a2d[0][1] -= mass * pos.xyz.x * pos.xyz.y;
+	world_moi.a2d[1][0] -= mass * pos.xyz.x * pos.xyz.y;
+	world_moi.a2d[0][2] -= mass * pos.xyz.x * pos.xyz.z;
+	world_moi.a2d[2][0] -= mass * pos.xyz.x * pos.xyz.z;
+	world_moi.a2d[1][2] -= mass * pos.xyz.y * pos.xyz.z;
+	world_moi.a2d[2][1] -= mass * pos.xyz.y * pos.xyz.z;
+
+	*infop->maintained_variables.matrix_value += world_moi;
+}
+
+void dock_whack_all_docked_objects_helper(object* objp, dock_function_info* infop)
+{
+	float orig_impulse = infop->parameter_variables.float_value;
+	vec3d* world_delta_rotvel = infop->parameter_variables.vecp_value;
+	vec3d* main_delta_vel = infop->parameter_variables.vecp_value2;
+	vec3d* world_center_of_mass = infop->maintained_variables.vecp_value;
+	vec3d delta_rotvel;
+
+	// compute your rotvel
+	vm_vec_rotate(&delta_rotvel, world_delta_rotvel, &objp->orient);
+	
+	// compute your linear vel based on the composite rotvel + offset and composite linear vel
+	vec3d delta_vel;
+	vec3d rel_pos;
+	vm_vec_sub(&rel_pos, &objp->pos, world_center_of_mass);
+	vm_vec_cross(&delta_vel, &rel_pos, world_delta_rotvel);
+	vm_vec_add2(&delta_vel, main_delta_vel);
+	
+	// whack it
+	physics_apply_whack_direct(orig_impulse, &objp->phys_info, &delta_rotvel, &delta_vel, &objp->orient);
+}
 // ---------------------------------------------------------------------------------------------------------------
 // end of über code block ----------------------------------------------------------------------------------------
 
--- a/code/object/objectdock.cpp
+++ b/code/object/objectdock.cpp
@@ -30,7 +30,6 @@ void dock_calc_max_semilatus_rectum_squared_parallel_to_directrix_helper(object 
 void dock_find_max_speed_helper(object *objp, dock_function_info *infop);
 void dock_find_max_fspeed_helper(object *objp, dock_function_info *infop);
 void dock_calc_total_moi_helper(object* objp, dock_function_info* infop);
-void dock_whack_docked_object_helper(object* objp, dock_function_info* infop);
 
 // management prototypes
 
@@ -377,9 +376,9 @@ object* dock_find_dock_root(object *objp)
 	return fastest_objp;
 }
 
-void dock_whack_docked_object(vec3d* force, vec3d* world_hit_pos, object* objp)
+void dock_whack_docked_object(vec3d* force, const vec3d* rel_world_hit_pos, object* objp)
 {
-	Assertion((objp != nullptr) && (force != nullptr) && (world_hit_pos != nullptr),
+	Assertion((objp != nullptr) && (force != nullptr) && (rel_world_hit_pos != nullptr),
 		"dock_whack_docked_object invalid argument(s)");
 
 	//	Detect null vector.
@@ -387,15 +386,16 @@ void dock_whack_docked_object(vec3d* force, vec3d* world_hit_pos, object* objp)
 		return;
 
 	vec3d world_center_of_mass;
+	vec3d world_hit_pos;
 
 	// calc world hit pos of the hit ship
-	vm_vec_add2(world_hit_pos, &objp->pos);
+	vm_vec_add(&world_hit_pos, rel_world_hit_pos, &objp->pos);
 
 	// calc overall world center-of-mass of all ships
 	float total_mass = dock_calc_docked_center_of_mass(&world_center_of_mass, objp);
 
 	// the new hitpos is the vector from world center-of-mass to world hitpos
-	vm_vec_sub2(world_hit_pos, &world_center_of_mass);
+	vm_vec_sub2(&world_hit_pos, &world_center_of_mass);
 
 	matrix moi, inv_moi;
 
@@ -406,7 +406,7 @@ void dock_whack_docked_object(vec3d* force, vec3d* world_hit_pos, object* objp)
 
 	// calculate the torque about the center of mass in world coords
 	vec3d torque;
-	vm_vec_cross(&torque, world_hit_pos, force);
+	vm_vec_cross(&torque, &world_hit_pos, force);
 
 	// calculate the change in rotvel caused by the whack in world coords
 	vec3d delta_rotvel;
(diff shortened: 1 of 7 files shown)


code/object/objectdock.cpp
```diff
@@ -337,6 +340,65 @@ float dock_calc_docked_speed(object *objp)
 	return dfi.maintained_variables.float_value;
 }
 
+void dock_calc_total_moi(matrix* dest, object* objp, vec3d *center)
+{
+	Assert(dest != NULL);
```
JohnAFernandez, line 345: nullptr!!

code/object/objectdock.cpp
```diff
@@ -337,6 +340,65 @@ float dock_calc_docked_speed(object *objp)
...
+	Assert(objp != NULL);
+
+	matrix accum, local_moi;
```
JohnAFernandez, line 348: Build checks are saying that local_moi is unreferenced.

code/object/objectdock.cpp
```diff
@@ -337,6 +340,65 @@ float dock_calc_docked_speed(object *objp)
...
+	// calc world hit pos of the hit ship
+	vm_vec_add2(world_hit_pos, &objp->pos);
+
+	// calc overall world center-of-mass of all ships
+	float total_mass = dock_calc_docked_center_of_mass(&world_center_of_mass, objp);
+
+	// the new hitpos is the vector from world center-of-mass to world hitpos
+	vm_vec_sub2(world_hit_pos, &world_center_of_mass);
+
+	matrix moi, local_moi, inv_moi;
```
JohnAFernandez, line 375: Same here.

code/object/objectdock.cpp
Baezon: Using NULL is common in this file, should I change all of them?

JohnAFernandez: My way to do it is to write new ones as nullptr, but leave old ones for a cleanup or refactor PR.

code/math/vecmat.cpp
```diff
@@ -2571,6 +2571,56 @@ void vm_vec_boxscale(vec2d *vec, float  /*scale*/)
...
 
+// adds one matrix to another.
+// dest can equal source
+// dest += src
+void vm_matrix_add2(matrix* dest, const matrix* src)
+{
+	dest->vec.fvec += src->vec.fvec;
+	dest->vec.rvec += src->vec.rvec;
+	dest->vec.uvec += src->vec.uvec;
+}
```
Goober5000, line 2582: I would strongly advise making a separate PR for *just* the vecmat changes, and then rebasing this PR after that one is in the codebase.  The reason is I think it would be good to include a few more things here...

* `vm_matrix_add` for two source arguments, and a `vm_matrix_sub`
* `vm_matrix_sub2` equivalent to `vm_matrix_add2`
* some test suite tests to verify that the matrix math is, in fact, correct

code/math/vecmat.cpp
```diff
@@ -2571,6 +2571,56 @@ void vm_vec_boxscale(vec2d *vec, float  /*scale*/)
...
+	inv.a2d[1][0] = m->a2d[1][2] * m->a2d[2][0] - m->a2d[1][0] * m->a2d[2][2];
+	inv.a2d[1][1] = -m->a2d[0][2] * m->a2d[2][0] + m->a2d[0][0] * m->a2d[2][2];
+	inv.a2d[1][2] = m->a2d[0][2] * m->a2d[1][0] - m->a2d[0][0] * m->a2d[1][2];
+	inv.a2d[2][0] = -m->a2d[1][1] * m->a2d[2][0] + m->a2d[1][0] * m->a2d[2][1];
+	inv.a2d[2][1] = m->a2d[0][1] * m->a2d[2][0] - m->a2d[0][0] * m->a2d[2][1];
+	inv.a2d[2][2] = -m->a2d[0][1] * m->a2d[1][0] + m->a2d[0][0] * m->a2d[1][1];
+
+	float det = m->a2d[0][0] * inv.a2d[0][0] + m->a2d[0][1] * inv.a2d[1][0] + m->a2d[0][2] * inv.a2d[2][0];
+	if (det == 0) {
+		dest = nullptr;
```
Goober5000, line 2609: Setting `dest = nullptr` is not going to have any non-local effects.  Instead, the destination matrix should be set to some trivial matrix analogous to the trivial vector in `vm_vec_normalize`.  (I notice that the matrix4 version of this function has the same bug.)

code/math/vecmat.cpp
```diff
@@ -2571,6 +2571,56 @@ void vm_vec_boxscale(vec2d *vec, float  /*scale*/)
...
+		return false;
+	}
+
+	det = 1.0f / det;
+
+	for (int i = 0; i < 3; i++) {
```
Goober5000, line 2615: It would be more efficient to use `a1d` and loop from 0 to <9.

code/math/vecmat.h
```diff
@@ -36,6 +36,10 @@
 //Note: NO RETURN VALUE
 #define vm_vec_zero(v) (v)->xyz.x=(v)->xyz.y=(v)->xyz.z=0.0f
 
+//macro to set a matrix to zero.
+//Note: NO RETURN VALUE
+#define vm_mat_zero(m) (vm_vec_zero(&(m)->vec.rvec), vm_vec_zero(&(m)->vec.uvec), vm_vec_zero(&(m)->vec.fvec))
```
Goober5000, line 41: Get rid of the ampersand here; it should have the same calling syntax as the other macro (`vm_vec_zero(&m)` similar to `vm_vec_zero(&vec)`)

code/math/vecmat.h
```diff
@@ -477,7 +481,14 @@ void vm_vec_dist_squared_to_line(const vec3d *p, const vec3d *l0, const vec3d *l
...
 
-bool vm_inverse_matrix4(const matrix4 *m, matrix4 *invOut);
+// adds one matrix to another.
+// dest can equal source
+// dest += src
+void vm_matrix_add2(matrix* dest, const matrix* src);
+
+bool vm_inverse_matrix(matrix* dest, const matrix* m);
+
+bool vm_inverse_matrix4(const matrix4* m, matrix4* invOut);
```
Goober5000, line 491: If you are going to make a separate vecmat PR (which I recommend), then it would be good to swap the arguments of `vm_inverse_matrix4` and fix the null matrix bug I mentioned.

code/object/objectdock.cpp
```diff
@@ -199,13 +200,13 @@ void dock_calc_docked_center_of_mass(vec3d *dest, object *objp)
...
-	vm_vec_scale(dest, (1.0f / dfi.maintained_variables.float_value));
+	float total_mass = dfi.maintained_variables.float_value;
+	vm_vec_scale(dest, (1.0f / total_mass));
+	return total_mass;
 }
 
 float dock_calc_total_docked_mass(object *objp)
 {
-	Assert(objp != NULL);
-
```
Goober5000, line 208: Why was this removed?

code/physics/physics.cpp
```diff
@@ -725,26 +725,34 @@ void physics_read_flying_controls( matrix * orient, physics_info * pi, control_i
...
 
 
+#define WHACK_LIMIT 0.001f
+#define ROTVEL_WHACK_CONST 0.12
+
+//	-----------------------------------------------------------------------------------------------------------
+// Returns true if this impulse is below the limit and should be ignored.
+bool whack_below_limit(vec3d* impulse)
+{
+	return (fl_abs(impulse->xyz.x) <= WHACK_LIMIT) && (fl_abs(impulse->xyz.y) <= WHACK_LIMIT) &&
```
Goober5000, line 735: Pedantically, this should be <, not <= to be strictly *below* a limit.

code/object/objectdock.cpp
```diff
@@ -733,6 +797,53 @@ void object_remove_arriving_stage2_ndl_flag_helper(object *objp, dock_function_i
 		Ships[objp->instance].flags.remove(Ship::Ship_Flags::Arriving_stage_2_dock_follower);
 }
 
+void dock_calc_inv_total_moi_helper(object* objp, dock_function_info* infop)
```
Goober5000, line 800: Now wait, is this calculating the total MOI or the total inverse MOI?

code/object/objectdock.cpp
```diff
@@ -337,6 +338,68 @@ float dock_calc_docked_speed(object *objp)
...
+	vm_mat_zero(&accum);
+
+	dock_function_info dfi;
+	dfi.parameter_variables.vecp_value = center;
+	dfi.maintained_variables.matrix_value = dest;
+
+	dock_evaluate_all_docked_objects(objp, &dfi, dock_calc_inv_total_moi_helper);
+}
+
+void dock_whack_all_docked_objects(vec3d* force, vec3d* world_hit_pos, object* objp)
```
Goober5000, line 356: Ack, no no no.  You don't whack all the ships, you whack *one* ship using the aggregate mass and aggregate MOI.

code/object/objectdock.cpp
Baezon: The function doesn't do anything that requires it to not be null in the body itself, but it calls `dock_evaluate_all_docked_objects` which does assert on `objp` being null. Not sure what the convention is here.

Goober5000: I suppose that's reasonable.

code/math/vecmat.h
Baezon: vm_vec_zero takes a pointer to a vector so vm_mat_zero needs to send it pointers instead of the vectors themselves.

Baezon:
Whacks will now go only to the 'dock root', the root of the tree that is also used by docking movement, with ships lower down the tree being updated to match it.

code/object/objectdock.cpp
```diff
@@ -733,6 +831,30 @@ void object_remove_arriving_stage2_ndl_flag_helper(object *objp, dock_function_i
 		Ships[objp->instance].flags.remove(Ship::Ship_Flags::Arriving_stage_2_dock_follower);
 }
 
+void dock_calc_total_moi_helper(object* objp, dock_function_info* infop)
+{
+	matrix local_moi, unorient, temp, world_moi;
+	vm_inverse_matrix(&local_moi, &objp->phys_info.I_body_inv);
+	vm_copy_transpose(&unorient, &objp->orient);
+	vm_matrix_x_matrix(&temp, &objp->orient , &local_moi);
+	vm_matrix_x_matrix(&world_moi, &temp, &unorient);
```
Goober5000, line 840: it would be a good idea to add comments and write out what you're doing here, using mathematical formulas, for the benefit of future programmers who look at the code

code/object/objectdock.cpp
```diff
@@ -733,6 +831,30 @@ void object_remove_arriving_stage2_ndl_flag_helper(object *objp, dock_function_i
...
+	// moment of inertia for a point mass: I_xx = m(y^2+z^2), I_xy = -mxy, etc
+	world_moi.a2d[0][0] += mass * (pos.xyz.y * pos.xyz.y + pos.xyz.z * pos.xyz.z);
+	world_moi.a2d[1][1] += mass * (pos.xyz.x * pos.xyz.x + pos.xyz.z * pos.xyz.z);
+	world_moi.a2d[2][2] += mass * (pos.xyz.x * pos.xyz.x + pos.xyz.y * pos.xyz.y);
+	world_moi.a2d[0][1] -= mass * pos.xyz.x * pos.xyz.y;
+	world_moi.a2d[1][0] -= mass * pos.xyz.x * pos.xyz.y;
+	world_moi.a2d[0][2] -= mass * pos.xyz.x * pos.xyz.z;
+	world_moi.a2d[2][0] -= mass * pos.xyz.x * pos.xyz.z;
+	world_moi.a2d[1][2] -= mass * pos.xyz.y * pos.xyz.z;
+	world_moi.a2d[2][1] -= mass * pos.xyz.y * pos.xyz.z;
```
Goober5000, line 853: Would it be appropriate to put this in vecmat?

code/physics/physics.cpp
```diff
@@ -725,26 +725,34 @@ void physics_read_flying_controls( matrix * orient, physics_info * pi, control_i
 }
 
 
+#define WHACK_LIMIT 0.001f
+#define ROTVEL_WHACK_CONST 0.12f
+
+//	-----------------------------------------------------------------------------------------------------------
+// Returns true if this impulse is below the limit and should be ignored.
+bool whack_below_limit(vec3d* impulse)
```
Goober5000, line 733: Make the impulse parameter `const`

code/physics/physics.cpp
```diff
@@ -755,33 +763,48 @@ void physics_apply_whack(vec3d *impulse, vec3d *pos, physics_info *pi, matrix *o
...
+	// Goober5000 - pi->mass should probably be just mass, as specified in the header
+	vec3d delta_vel = *impulse * (1.0f / mass);
+
+	physics_apply_whack_direct(vm_vec_mag(impulse), pi, &delta_rotvel, &delta_vel, orient);
+}
+
+
+// This function applies the calculated delta rotational and linear velocities usually called by ^^^^ apply_whack, 
+// but also called directly by dock_whack_docked_object in objectdock.cpp which has to do some extra legwork
+// and doesn't set this function up using physics_apply_whack
```
Goober5000, line 777: What's the difference between "direct" and non-direct?  What does "direct" mean in this case?

code/ship/shiphit.cpp
```diff
@@ -1783,46 +1783,27 @@ extern int Homing_hits, Homing_misses;
...
 	if (objp == Player_obj) {
 		nprintf(("Sandeep", "Playing stupid joystick effect\n"));
 		vec3d test;
 		vm_vec_unrotate(&test, force, &objp->orient);
 
 		game_whack_apply( -test.xyz.x, -test.xyz.y );
 	}
+	
+	vec3d world_hit_pos;
+	vm_vec_unrotate(&world_hit_pos, hit_pos, &objp->orient);
```
Goober5000, line 1797: Again it looks like this is an incomplete formula.  To go from local to global, you do an unrotate followed by an add.  To go from global to local, you do a subtract followed by a rotate.  These operations must always be paired to obtain the complete transformation.

code/object/objectdock.cpp
```diff
@@ -337,6 +340,106 @@ float dock_calc_docked_speed(object *objp)
...
+		"dock_whack_docked_object invalid argument(s)");
+
+	//	Detect null vector.
+	if (whack_below_limit(force))
+		return;
+
+	vec3d world_center_of_mass;
+
+	// calc world hit pos of the hit ship
+	vm_vec_add2(world_hit_pos, &objp->pos);
```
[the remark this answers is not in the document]
Baezon: The `world_hit_pos` parameter is the world coords hit position relative to the ship, and adding it to the ship's position is getting the absolute position of it. I can change it to something like `rel_world_hit_pos`, and keep it `const` and modify a new local variable for clarity.

code/physics/physics.cpp
Baezon: Naming of this was a little bit of an issue. "Direct" is actually applying the velocities calculated by `physics_apply_whack' and possibly also `dock_whack_docked_object`, those two are where the actual physical calculations are done. I'd be open to a better naming scheme. `physics_calculate_whack` and `physics_apply_whack`?

code/object/objectdock.cpp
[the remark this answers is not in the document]
Goober5000: > world coords hit position relative to the ship

No, *world coordinates* are not relative to anything.  They are absolute with their origin at the mission's 0,0,0 point.

The game uses world coordinates from 0,0,0 and local coordinates relative to the ship, taking into account the ship's position and orientation.  There are no other types of coordinates except for temporary variables in the middle of a calculation.

code/physics/physics.cpp
Goober5000: Yes, that sounds better.

code/physics/physics.cpp
```diff
@@ -730,7 +730,7 @@ void physics_read_flying_controls( matrix * orient, physics_info * pi, control_i
 
 //	-----------------------------------------------------------------------------------------------------------
 // Returns true if this impulse is below the limit and should be ignored.
-bool whack_below_limit(vec3d* impulse)
+bool whack_below_limit(vec3d* const impulse)
```
Goober5000, line 733: The `const` should go first.

code/object/objectdock.cpp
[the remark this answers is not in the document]
Baezon: The hit pos here, are the coordinates with the ship at the origin, but with the world's orientation (that is, taking into account the ship's position, but _not_ its orientation). I don't know whether you want to call that 'local' or 'world', but its used very frequently in the physics code, and physics_apply_whack, even before these changes, requires that its 'pos' argument be supplied in this reference frame.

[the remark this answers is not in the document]
Goober5000: That's neither local nor world; that's a weird hybrid.  Can you cite an example in the physics code?

code/object/objectdock.cpp
```diff
@@ -733,6 +831,38 @@ void object_remove_arriving_stage2_ndl_flag_helper(object *objp, dock_function_i
...
+
+	// We calculate the inverse of the orientation matrix (which is also the transpose)
+	vm_copy_transpose(&unorient, &objp->orient);
+
+	// We calculate the world space MOI using (world MOI) = O^-1 * (local MOI) * O
+	// where O is the orientation matrix (which translates between local space and world space).
+	// Note that this is reverse of the usual convention because FS stores orientation matrices
+	// transposed relative to the linear algebra convention.
+	vm_matrix_x_matrix(&temp, &objp->orient, &local_moi);
+	vm_matrix_x_matrix(&world_moi, &temp, &unorient);
```
Goober5000, line 852: Do you mean that the *formula* is the reverse of the usual convention, or the *code* is the reverse of the formula?  It looks like the latter, but it would be good to clarify in the comments.

code/physics/physics.cpp
```diff
@@ -725,26 +725,34 @@ void physics_read_flying_controls( matrix * orient, physics_info * pi, control_i
...
+}
+
 // ----------------------------------------------------------------------------
 // physics_apply_whack applies an instaneous whack on an object changing
 // both the objects velocity and the rotational velocity based on the impulse
 // being applied.
 //
 //	input:	impulse		=>		impulse vector ( force*time = impulse = change in momentum (mv) )
-//				pos			=>		vector from center of mass to location of where the force acts
+//				pos			=>		vector from center of mass to location of where the force acts (in world coords)
```
Goober5000, line 745: Put the parenthetical after location:
```
location (in world coords)
```
This is the least ambiguous.

code/physics/physics.cpp
```diff
@@ -755,33 +763,48 @@ void physics_apply_whack(vec3d *impulse, vec3d *pos, physics_info *pi, matrix *o
...
+	vm_vec_add2(&pi->vel, delta_vel);
 	if (!(pi->flags & PF_USE_VEL) && (vm_vec_mag_squared(&pi->vel) > MAX_SHIP_SPEED*MAX_SHIP_SPEED)) {
 		// Get DaveA
 		nprintf(("Physics", "speed reset in physics_apply_whack [speed: %f]\n", vm_vec_mag(&pi->vel)));
 		vm_vec_normalize(&pi->vel);
 		vm_vec_scale(&pi->vel, (float)RESET_SHIP_SPEED);
 	}
 	vm_vec_rotate( &pi->prev_ramp_vel, &pi->vel, orient );		// set so velocity will ramp starting from current speed
-																					// ramped velocity is now affected by collision
+																// ramped velocity is now affected by collision
```
Goober5000, line 807: Can you un-indent the comment so it's not way over to the right?

code/physics/physics.cpp
```diff
@@ -725,26 +725,34 @@ void physics_read_flying_controls( matrix * orient, physics_info * pi, control_i
...
-//				pos			=>		vector from center of mass to location of where the force acts
+//				pos			=>		vector from center of mass to location of where the force acts (in world coords)
 //				pi				=>		pointer to phys_info struct of object getting whacked
 //				orient		=>		orientation matrix (needed to set rotational impulse in body coords)
 //				mass			=>		mass of the object (may be different from pi.mass if docked)
 //
-#define WHACK_LIMIT	0.001f
-#define ROTVEL_WHACK_CONST 0.12
-void physics_apply_whack(vec3d *impulse, vec3d *pos, physics_info *pi, matrix *orient, float mass)
+void physics_calculate_whack(vec3d *impulse, vec3d *pos, physics_info *pi, matrix *orient, float mass, matrix *inv_moi)
```
Goober5000, line 750: Rename this to `physics_calculate_and_apply_whack`.  I thought the functions would be entirely separate, but `calculate` feeds into `apply`.

code/object/objectdock.cpp
```diff
@@ -337,6 +339,107 @@ float dock_calc_docked_speed(object *objp)
...
+	fastest_objp = dfi.maintained_variables.objp_value;
+
+	// if we have no max speed, just use the given one
+	if (fastest_objp == nullptr)
+		fastest_objp = objp;
+
+	return fastest_objp;
+}
+
+void dock_calculate_whack_docked_object(vec3d* force, const vec3d* rel_world_hit_pos, object* objp)
```
Goober5000, line 379: Rename this to `dock_calculate_and_apply_whack_docked_object`, per comment on `physics_calculate_whack`

code/object/objectdock.cpp
```diff
@@ -337,6 +339,107 @@ float dock_calc_docked_speed(object *objp)
...
+	float total_mass = dock_calc_docked_center_of_mass(&world_center_of_mass, objp);
+
+	// the new hitpos is the vector from world center-of-mass to world hitpos
+	vm_vec_sub2(&world_hit_pos, &world_center_of_mass);
+
+	matrix moi, inv_moi;
+
+	// calculate the effective inverse MOI for the docked composite object about its center of mass
+	dock_calc_total_moi(&moi, objp, &world_center_of_mass);
+
```
Goober5000, line 404: Remove the blank line here.  It's two halves of one operation.

code/object/objectdock.cpp
```diff
@@ -337,6 +339,107 @@ float dock_calc_docked_speed(object *objp)
...
+	vec3d world_hit_pos;
+
+	// calc world hit pos of the hit ship
+	vm_vec_add(&world_hit_pos, rel_world_hit_pos, &objp->pos);
+
+	// calc overall world center-of-mass of all ships
+	float total_mass = dock_calc_docked_center_of_mass(&world_center_of_mass, objp);
+
+	// the new hitpos is the vector from world center-of-mass to world hitpos
+	vm_vec_sub2(&world_hit_pos, &world_center_of_mass);
```
Goober5000, line 398: Call this new result `world_hit_pos_diff` or even just `hit_pos` because it's not absolutely world coordinates.

code/object/objectdock.cpp
```diff
@@ -337,6 +339,107 @@ float dock_calc_docked_speed(object *objp)
...
+
+	vm_inverse_matrix(&inv_moi, &moi);
+
+	// calculate the torque about the center of mass in world coords
+	vec3d torque;
+	vm_vec_cross(&torque, &world_hit_pos, force);
+
+	// calculate the change in rotvel caused by the whack in world coords
+	vec3d delta_rotvel;
+	vm_vec_rotate(&delta_rotvel, &torque, &inv_moi);
```
Goober5000, line 413: Compare this to the other code:
```
vm_vec_cross(&torque, pos, impulse);
vm_vec_rotate ( &local_torque, &torque, orient );
```
Impulse is not the same as force, and orient is not the same as inverse MOI.  Is one of the formulas in error, or should `torque` be called something else?

Baezon: All of the names here should be related to impulse: `force` should be `impulse` and `torque` should be `delta_angular_momentum` or `angular_impulse`. `delta_rotvel` is named correctly. `inv_moi` is actually inverse MOI here. (I believe the original code was in error in naming `torque` that instead of something like `angular_impulse`, but the multiplication by `orient` in the original code is unrelated to the multiplication by `inv_moi` in the new code.)

code/object/objectdock.cpp
Goober5000: No, `rel_world_hit_pos` should keep its original name.  I meant that `world_hit_pos` should be renamed.  That's the variable holding the result of the `vm_vec_sub2` operation on this line.

code/object/objectdock.cpp
Goober5000: Okay.

Baezon:
It's beautifully boring. Everything behaving in a physically accurate way is terribly dull, and I already miss the exciting days when you could get obliterated by a typhon because you accidentally shot the isis docked to it.

JohnAFernandez:
Great fix @Baezon! Way to stick it out!!

PR #2556, opened 2020-07-08
Addition of matrix3 functions in vecmat

Baezon:
Added inverse, add, add2, sub, sub2 and a ZERO_MATRIX.


code/math/vecmat.h
```diff
@@ -36,6 +36,11 @@
 //Note: NO RETURN VALUE
 #define vm_vec_zero(v) (v)->xyz.x=(v)->xyz.y=(v)->xyz.z=0.0f
 
+//macro to set a vector to zero.  we could do this with an in-line assembly
+//macro, but it's probably better to let the compiler optimize it.
+//Note: NO RETURN VALUE
+#define vm_mat_zero(m) (vm_vec_zero(&(m)->vec.rvec), vm_vec_zero(&(m)->vec.uvec), vm_vec_zero(&(m)->vec.fvec))
```
Goober5000, line 42: Please remove the ampersand.  Like the vector macro, this macro will be called on pointers.

Also, given that you're calling functions here, not performing assignment, you'll need to use a do/while block.

Goober5000: Whoops, per Discord, ampersand is necessary.

PR #2564, opened 2020-07-12
Prevent cargo containers from recovering from a collision

Baezon:
Allowing not big or huge ships to recover from collisions with large ships by attempting to fly away, also allows cargo containers, navbuoys, and sentry guns to attempt it. This keeps it to small ships: fighters, bombers, support ships and escape pods.


Goober5000:
Took me a minute to parse that first sentence.  Basically, first half implies second half.  Good fix.

PR #2567, opened 2020-07-12
Relative error for mag and dist_quick, and seeded vecmat tests

Baezon:
I calculated mag_quick's exact relative error (used also by dist_quick) and am using that instead of an absolute value. Also uses a seed for the random tests to avoid one in a thousand compiles failing for someone.

### DIFF
--- a/test/src/math/test_vecmat.cpp
+++ b/test/src/math/test_vecmat.cpp
@@ -35,7 +35,7 @@ public:
 	VecmatTest() : test::FSTestFixture() { pushModDir("vecmat"); }
 
 protected:
-	void SetUp() override { test::FSTestFixture::SetUp(); }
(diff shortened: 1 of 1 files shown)


test/src/math/test_vecmat.cpp
```diff
@@ -35,7 +35,7 @@ class VecmatTest : public test::FSTestFixture {
 	VecmatTest() : test::FSTestFixture() { pushModDir("vecmat"); }
 
 protected:
-	void SetUp() override { test::FSTestFixture::SetUp(); }
+	void SetUp() override { test::FSTestFixture::SetUp(); srand(1); }
```
asarium, line 38: Please format this so that the two statements are on different lines. The clang-format config should be mostly correct so you can also use that to format this line if you want.

PR #2569, opened 2020-07-13
some UTF-8 improvements

Goober5000:
1) Write all non-UTF-8 strings to JSON, either by converting or truncating
2) Add `coerce_to_utf8` function
3) Tweak `read_raw_file_text` because the function returns length on success, not 0

Fixes Mantis ticket:
http://scp.indiegames.us/mantis/view.php?id=3209

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -2233,7 +2233,9 @@ void read_raw_file_text(const char *filename, int mode, char *raw_text)
 					auto err = SDL_iconv(iconv, &in_str, &in_size, &out_str, &out_size);
 					SDL_iconv_close(iconv);
 
-					if (err == 0) {
+					// SDL returns the number of processed character on success;
+					// error codes are (size_t)-1 through -4
+					if (err < (size_t)-100) {
 						break;
 					} else if (err == SDL_ICONV_E2BIG) {
 						// buffer is not big enough, try again with a bigger buffer. Use a rather conservative size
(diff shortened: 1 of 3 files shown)


code/parse/parselo.cpp
```diff
@@ -2258,6 +2260,74 @@ void read_raw_file_text(const char *filename, int mode, char *raw_text)
...
+		return;
+	}
+
+	bool isLatin1 = util::guessLatin1Encoding(str, len);
+
+	// we can convert it
+	if (isLatin1)
+	{
+		size_t newlen = len;
+		auto newstr = new char[newlen];
```
asarium, line 2283: Please use `std::unique_ptr<char[]>` for a bit more safety.

code/pilotfile/JSONFileHandler.cpp
```diff
@@ -4,6 +4,11 @@
 #include "libs/jansson.h"
 #include "parse/parselo.h"
 
+// it's kinda tricky to #include "utf.h" here, so we forward-declare this, but note that it's a C function
+extern "C" {
+	extern int utf8_check_string(const char *string, size_t length);
```
asarium, line 9: What are the issues with including that header? We should avoid doing library declarations on our own if at all possible. The contents of this function also seem to be more or less the same as `utf8::find_invalid` so if including the header doesn't work maybe you could use that function instead?

Goober5000: Just that it's a very convoluted path.  But I'll try that alternative.

code/parse/parselo.cpp
Goober5000: Ok.

PR #2570, opened 2020-07-13
make _quick quicker

Goober5000:
The functions derived from `vm_vec_mag_quick` run *slower* than `vm_vec_mag`, according to @Baezon who did some testing.  So this PR makes all calls to the _quick functions use their more accurate counterparts.


Baezon:
Don't forget to remove the tests for those functions as well.

Goober5000:
Ok, done.

PR #2571, opened 2020-07-14
Prevent indrectly docked ships from colliding

Baezon:
When the collision code is precluding things from colliding due being docked, it only checks for directly docked objects. `dock_check_find_docked_object` functions identically in this role, but also checks for indirectly docked objects.


Goober5000:
Looks good.  One last change: rename `ships_are-docking` to something that more accurately describes what is being checked -- i.e. the whole docked stack or docked group as opposed to the direct one-on-one ships.  It's really only suitable for collision checks; it's not a general-purpose function even though at first glance it appears to be.

PR #2572, opened 2020-07-14
Randomized weapon shockwave orientations

Baezon:
Currently 3D shockwaves from weapons will always have their orientations fixed at 0,0,0 unless explicitly tabled otherwise. There's really no reason for them not to be simply randomized if the modder didn't set a specific orientation, and it looks quite jarring once you notice it.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -6349,9 +6349,9 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
 	{
 		if(sci->speed > 0.0f) {
 			if (!sci->rot_defined) {
-				sci->rot_angles.p = frand_range(0.0f, 1.99f * PI);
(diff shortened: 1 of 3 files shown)


code/weapon/weapons.cpp
```diff
@@ -6346,6 +6348,11 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
 	if (sci->inner_rad != 0.0f || sci->outer_rad != 0.0f)
 	{
 		if(sci->speed > 0.0f) {
+			if (!sci->rot_defined) {
+				sci->rot_angles.p = frand_range(0.0f, 1.99f * PI);
+				sci->rot_angles.b = frand_range(0.0f, 1.99f * PI);
+				sci->rot_angles.h = frand_range(0.0f, 1.99f * PI);
```
Goober5000, line 6354: Please use the `PI2` contant.  (The range will safely exclude the maximum number.)

PR #2577, opened 2020-07-16
Allow pasting clipboard contents into input boxes

notimaginative:
By special request of Cyborg. Pressing CTRL+v in an active input box will append any text from the system clipboard into the control.

### DIFF
--- a/code/ui/inputbox.cpp
+++ b/code/ui/inputbox.cpp
@@ -373,6 +373,19 @@ void UI_INPUTBOX::process(int focus)
 
 			default:
 				if (!locked) {
+					// allow pasting from system clipboard
+					if (key == (KEY_CTRLED | KEY_V)) {
+						if ( SDL_HasClipboardText() ) {
+							char *cliptext = SDL_GetClipboardText();
+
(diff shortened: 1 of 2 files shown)


code/ui/inputbox.cpp
```diff
@@ -477,3 +490,33 @@ void UI_INPUTBOX::set_text(const char *in)
...
+	// current size
+	size_t textlen = strlen(text);
+
+	if (textlen == static_cast<size_t>(length)) {
+		return;
+	}
+
+	// strcat_s() will zero the string if it's too long, which isn't what we want
+	// SDL_strlcat() is safe while still preserving the dest string on error
+	SDL_strlcat(text, in, length+1);
```
Goober5000, line 509: In the case of overflow, will this paste as much of the text as will fit, or will it just not paste anything?

notimaginative: It will paste as much as will fit. I thought about making it an error and leaving the dest string unchanged, but there doesn't seem to be a good way to report that condition to the user.

PR #2580, opened 2020-07-18
Fix sidebars retaining old rendering from mission quick exit

JohnAFernandez:
And also a minor typo

The issue in other words is that stuff rendered on the last frame of the mission will be stuck on the sidebars for a long time.

Just tested and this works on release and debug. Open to moving the fix somewhere, just let me know if there's a more proper place.


Goober5000:
I like that typo.

Do you need to do clear/flip twice?

The-E:
The double flip/clear is necessary to clear both the front and back buffers; since both are going to be full of mission things, both need to be cleared. This is something we also do when, for example, we transition into the movie player.

PR #2583, opened 2020-07-21
add scripting function to create matrix from vectors

Goober5000:
What it says on the tin.  Previously there was no access to the `vm_vector_2_matrix` API.  One or more of the vectors may be nil or missing.


code/scripting/api/libs/base.cpp
```diff
@@ -75,6 +75,25 @@ ADE_FUNC(createOrientation,
 	return ade_set_error(L, "o", l_Matrix.Set(matrix_h()));
 }
 
+ADE_FUNC(createOrientationFromVectors, l_Base, "[vector fvec, vector uvec, vector rvec]",
+	"Given 0 to 3 arguments, creates an orientation object from vectors.",
```
asarium, line 79: It's probably a good idea to document that passing no parameters will result in the identity matrix. Not strictly necessary but could reduce confusion for people who don't read the code.

PR #2587, opened 2020-07-21
Additional scripting enhancements

Goober5000:
Several new scripting functions and features related to a script I'm working on.  All new scripting functions have been tested and are working.

This PR is feature-complete, but I want to ask about the `YAxisInverted` function before moving it out of draft status.  What I would like to do is make an `AxisInverted[n]` virtual variable (or indexer?) that accepts any joystick axis and allows testing or setting each axis.  What's the best way to do that?  Or would it be better to add separate `XAxisInverted` and `ZAxisInverted` virtual variables instead?

### DIFF
--- a/code/scripting/api/libs/base.cpp
+++ b/code/scripting/api/libs/base.cpp
@@ -106,6 +106,75 @@ ADE_FUNC(createVector, l_Base, "[number x, number y, number z]", "Creates a vect
 	return ade_set_args(L, "o", l_Vector.Set(v3));
 }
 
+ADE_FUNC(createSurfaceNormal,
+	l_Base,
+	"vector point1, vector point2, vector point3",
+	"Determines the surface normal of the plane defined by three points.  Returns a normalized vector.",
+	"vector",
+	"The surface normal, or NIL if a handle is invalid")
+{
+	vec3d *p0 = nullptr, *p1 = nullptr, *p2 = nullptr;
+	if (!ade_get_args(L, "ooo", l_Vector.GetPtr(&p0), l_Vector.GetPtr(&p1), l_Vector.GetPtr(&p2)))
+		return ADE_RETURN_NIL;
+
+	vec3d dest;
+	vm_vec_normal(&dest, p0, p1, p2);
+	return ade_set_args(L, "o", l_Vector.Set(dest));
+}
+
+ADE_FUNC(findIntersection,
+	l_Base,
+	"vector line1-point1, vector line1-point2, vector line2-point1, vector line2-point2",
+	"Determines the point at which two lines intersect.  (The lines are assumed to extend infinitely in both directions; the intersection will not necessarily be between the points.)",
+	ade_type_info({ "vector", "number" }),
+	"Returns two arguments.  The first is the point of intersection, if it exists and is unique (otherwise it will be NIL).  The second is the find_intersection return value: 0 for a unique intersection, -1 if the lines are colinear, and -2 if the lines do not intersect.")
+{
+	vec3d *p0 = nullptr, *p0_end = nullptr, *p1 = nullptr, *p1_end = nullptr;
+	if (!ade_get_args(L, "oooo", l_Vector.GetPtr(&p0), l_Vector.GetPtr(&p0_end), l_Vector.GetPtr(&p1), l_Vector.GetPtr(&p1_end)))
+		return ADE_RETURN_NIL;
+
+	// note: we must translate from this API's two-points method to the code's API of a reference point and a direction vector
+	vec3d v0, v1;
+	vm_vec_sub(&v0, p0_end, p0);
+	vm_vec_sub(&v1, p1_end, p1);
+
+	float scalar;
+	int retval = find_intersection(&scalar, p0, p1, &v0, &v1);
+
+	if (retval == 0)
+	{
+		// per comments:
+		// If you want the coords of the intersection, scale v0 by s, then add p0.
+		vm_vec_scale(&v0, scalar);
+		vm_vec_add2(&v0, p0);
+
+		return ade_set_args(L, "oi", l_Vector.Set(v0), retval);
+	}
+	else
+		return ade_set_args(L, "*i", retval);
+}
+
+ADE_FUNC(findPointOnLineNearestSkewLine,
+	l_Base,
+	"vector line1-point1, vector line1-point2, vector line2-point1, vector line2-point2",
+	"Determines the point on line 1 closest to line 2 when the lines are skew (non-intersecting in 3D space).  (The lines are assumed to extend infinitely in both directions; the point will not necessarily be between the other points.)",
+	"vector",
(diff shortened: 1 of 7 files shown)


asarium:
You can do that by creating a proxy `ADE_OBJ` (where you can even keep using the same C++ class) that only has an `ADE_INDEXER` with which you can then get the member being accessed and do your stuff with that. You don't even need a `__len` function in that case since you would not want to iterate over the object anyway.

You then return an instance of this class from a virtual variable `AxisInverted` and you should be good to go.

code/scripting/api/libs/base.cpp
```diff
@@ -106,6 +106,82 @@ ADE_FUNC(createVector, l_Base, "[number x, number y, number z]", "Creates a vect
 	return ade_set_args(L, "o", l_Vector.Set(v3));
 }
 
+ADE_FUNC(createSurfaceNormal,
+	l_Base,
+	"[vector point1, vector point2, vector point3]",
```
asarium, line 111: The optional specifier here does not match the one passed to `ade_get_args`. Here every parameter is optional but below you specify that everything is required.

code/scripting/api/libs/base.cpp
```diff
@@ -106,6 +106,82 @@ ADE_FUNC(createVector, l_Base, "[number x, number y, number z]", "Creates a vect
...
+		return ADE_RETURN_NIL;
+
+	vec3d dest;
+	vm_vec_normal(&dest, p0, p1, p2);
+	return ade_set_args(L, "o", l_Vector.Set(dest));
+}
+
+ADE_FUNC(findIntersection,
+	l_Base,
+	"[vector line1-point1, vector line1-point2, vector line2-point1, vector line2-point2]",
```
asarium, line 127: Same here.

code/scripting/api/libs/base.cpp
```diff
@@ -106,6 +106,82 @@ ADE_FUNC(createVector, l_Base, "[number x, number y, number z]", "Creates a vect
...
+#ifndef NDEBUG
+	if (retval == 0)
+		mprintf(("intersection found\n"));
+	else if (retval == -1)
+		mprintf(("lines are colinear\n"));
+	else if (retval == -2)
+		mprintf(("no intersection\n"));
+	else
+		mprintf(("unknown intersection retval\n"));
+#endif
```
asarium, line 152: This looks like debug code that is not needed anymore.

code/scripting/api/libs/base.cpp
```diff
@@ -106,6 +106,82 @@ ADE_FUNC(createVector, l_Base, "[number x, number y, number z]", "Creates a vect
...
+	// If you want the coords of the intersection, scale v0 by s, then add p0.
+	vm_vec_scale(&v0, scalar);
+	vm_vec_add2(&v0, p0);
+
+	return ade_set_args(L, "o", l_Vector.Set(v0));
+}
+
+ADE_FUNC(findPointOnLineNearestSkewLine,
+	l_Base,
+	"[vector line1-point1, vector line1-point2, vector line2-point1, vector line2-point2]",
```
asarium, line 166: Optional specifier is also inconsistent here.

code/scripting/api/objs/sexpvar.cpp
```diff
@@ -9,6 +9,24 @@ namespace api {
...
+	if (!ade_get_args(L, "o|s", l_SEXPVariable.GetPtr(&svh), &s))
+		return ade_set_error(L, "s", "");
+
+	if (!svh->IsValid())
+		return ade_set_error(L, "s", "");
+
+	sexp_variable *sv = &Sexp_variables[svh->idx];
+
+	if (ADE_SETTING_VAR && s != nullptr) {
+		Warning(LOCATION, "Reassigning SEXP variable names is not supported.  (Tried to rename %s to %s.)", sv->variable_name, s);
```
asarium, line 25: Please use `LuaError` here since that gives the user a Lua stack trace.

code/scripting/api/objs/vecmath.cpp
```diff
@@ -559,5 +582,38 @@ ADE_FUNC(getNormalized,
...
+		return ADE_RETURN_NIL;
+
+	vec3d dest;
+	vm_vec_projection_onto_plane(&dest, src, normal);
+	return ade_set_args(L, "o", l_Vector.Set(dest));
+}
+
+ADE_FUNC(findNearestPointOnLine,
+	l_Vector,
+	"[vector point1, vector point2]",
```
asarium, line 603: Also inconsistent optional specifier.

code/scripting/api/libs/base.cpp
Goober5000: Oh, oops.

code/scripting/api/libs/base.cpp
Goober5000: It is debug code, but I didn't add it for myself.  I figured it would help script authors who used this function.

code/scripting/api/libs/base.cpp
Goober5000: K

code/scripting/api/libs/base.cpp
Goober5000: K

code/scripting/api/objs/vecmath.cpp
Goober5000: K

code/scripting/api/objs/sexpvar.cpp
Goober5000: Ok.

Goober5000:
After sleeping on it, I realized that separate `XAxisInverted` and `ZAxisInverted` variables would be more intuitive, so that's what I went with.

code/scripting/api/libs/base.cpp
asarium: I don't think this is particularly useful and adding random debug log outputs rarely helps solving issues. If the return value is relevant for the script then you could also return that as a second return value to let the script decide what to do with that information.

Goober5000: Good point.  Do scripters always need to catch all the return values?  I.e. if foo() returns two values, do I always need to call `a, b = foo()` or can I just call `a = foo()` and ignore the second one?

asarium: No, it is valid to just discard the second return value in which case `a` will have just the vector as its value.

Goober5000: How do I return a NIL vector as the first argument?  `l_Vector.Set(nullptr)` isn't valid, and `ADE_RETURN_NIL` is only for one argument.

asarium: You specify `*` as the type: `ade_set_args(L, "*i", 20);` or something like that. You also don't need to specify a value for that.

Goober5000: Cool, thanks.  New change committed.

PR #2590, opened 2020-07-22
Make ship explosion blasts use a local vector for whacks

Baezon:
Instant ship explosions provide a world position for whacks which result in exaggerated rotational velocities, `ship_apply_whack` is expecting a local position.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -7783,7 +7783,9 @@ static void ship_blow_up_area_apply_blast( object *exp_objp)
 				vm_vec_sub( &vec_ship_to_impact, &objp->pos, &exp_objp->pos );
 				vm_vec_copy_normalize( &force, &vec_ship_to_impact );
 				vm_vec_scale( &force, blast );
-				ship_apply_whack( &force, &vec_ship_to_impact, objp );
+				vec3d local_blast_vec;
+				vm_vec_rotate(&local_blast_vec, &vec_ship_to_impact, &objp->orient);
+				ship_apply_whack( &force, &local_blast_vec, objp );
 				break;
 			case OBJ_ASTEROID:
 				asteroid_hit(objp, NULL, NULL, damage);


Baezon:
I'm not sure if there's been significant changes here, or if retail was always confused on this point, but before my changes to whacks normal weapon applied whacks provided a local orientation to ship_apply_whack, but physics_apply_whack expected a world orientation (note the vm_vec_rotate early on, in both the old and new versions) and simply passed it on, which resulted incorrect behavior. Per your comment I presumed that you meant giving a local orientation to ship_apply_whack was the normal calling convention and I had ship_apply_whack turn the local it was provided into world for physics.

I'm happy to use whatever convention, but the callers need to be consistent about it.

Goober5000:
> I'm happy to use whatever convention, but the callers need to be consistent about it.

Agreed, and that is my objective as well.  I think we should standardize on whatever the original public code release used, because it's easier to compare that way.  There have been countless times I've reviewed the original code release to decipher something a FSO coder did, and sometimes it even reveals a bug.

Unfortunately at the time I wrote that comment I did not fully grasp the weird hybrid convention Volition used.  The two principal conventions are:
a) world orientation, world coordinates, 
b) orientation relative to ship, coordinates relative to ship

It wasn't until I talked to you that I realized the physics whack code used:
c) world orientation, coordinates relative to ship

And the beam whack code uses that as well.  When I wrote my comment, I thought the convention was b), but it was really c).  The previous code was a).  So I fixed the bug (or so I thought), but I fixed it wrongly.  I changed it from a) to b) when I should have changed it from a) to c).

Hopefully you understood all that.  What we should do is put that a) b) c) list in the comments to the physics whack functions, make it clear that they are using c), and then properly fix the beam whack to also use c).

Baezon:
I see, so to be clear, the convention from the callers should be a world orientation, and ship position? (As well as a removal of the unrotate done by `ship_apply_whack`). This is the form the physics eventually needs, although several other

PR #2591, opened 2020-07-23
Add homing primaries

Baezon:
This is a much leaner PR than one might expect, most of the code needed for homing primaries has already been processed the entire time, the main obstacle was simply all `WP_LASER`s being given `PF_CONST_VEL`, so physics assumed their velocity could never change (this includes turning).

Aside from that just a few changes that exclude primaries from normal missile speed adjustment, unless the modder could make the change themselves. Collision optimizations are connected to `PF_CONST_VEL` so those will continue to function normally. And I know @JohnAFernandez was concerned about the multi considerations, but by all accounts, if it can handle homing missiles, it should be able to handle homing primaries.

Aspect seeking primaries remain a sticking point, the locking code has lots of assumptions tied to secondary weapons, and how even having both aspect seeking primaries _and_ aspect seeking secondaries would even work (two unrelated lock indicators? different lock indicators?) is not a simple issue. For now, parsing will only allow heat seeking primaries.


Baezon:
On second thought, not homing when out of range should be a separate modder defined "view distance" for heat seekers that I'll implement later.

JohnAFernandez:
Let me run through this with rollback enabled and see if it works....  Do you have a table I could use?

Baezon:
Yep
[test-wep.txt](https://github.com/scp-fs2open/fs2open.github.com/files/4974711/test-wep.txt)

code/weapon/weapons.cpp
```diff
@@ -4395,7 +4408,8 @@ void weapon_home(object *obj, int num, float frame_time)
             (old_dot < wip->fov) &&
             (wp->lifeleft > 0.01f) &&
             (wp->homing_object != &obj_used_list) &&
-            (wp->homing_object->type == OBJ_SHIP))
+            (wp->homing_object->type == OBJ_SHIP) && 
+            (wp->missile_list_index != -1))				// Asteroth - that is, homing primaries don't do this
```
Goober5000, line 4412: I think it would be clearer to check for `WP_LASER` (or `WP_MISSILE`) here because that's conceptually closer to how you are making the distinction.

code/weapon/weapons.cpp
```diff
@@ -4429,7 +4443,8 @@ void weapon_home(object *obj, int num, float frame_time)
 
 		//	Control speed based on dot product to goal.  If close to straight ahead, move
 		//	at max speed, else move slower based on how far from ahead.
-		if (old_dot < 0.90f) {
+		//	Asteroth - but not for homing primaries
+		if (old_dot < 0.90f && wp->missile_list_index != -1) {
```
Goober5000, line 4447: Here too.

Goober5000:
@JohnAFernandez Did you run the test?

Baezon:
Haha, well I suppose I hadn't considered alternative viewpoints, but I'm totally of the opinion that the distinction between primaries and secondaries be eventually made totally arbitrary, at least as far as the modder is concerned.

Baezon:
For reference, this PR is in cold storage until the multi considerations can be worked out.

JohnAFernandez:
So, just as a note, I did run this in a multiplayer setting, and the shots on a low ping look identical on server and client, but I have yet to try it in a high ping situation.  Will probably need a tester in another country for that.

JohnAFernandez (approved):
So we have decided to punt multi support until later.  I am working on a patch to better sync homing weapons, but this doesn't need to wait because rollback works well enough with this, at least for most weapons.

We would just add some caveats to this feature's documentation to explain to modders that multi support is not yet perfect.

PR #2593, opened 2020-07-24
Change MP Initial Player Wing Ship Behavior to Match SP

JohnAFernandez:
The way that this function was structured before would 1. Change the AI mode of player ships (very minor) 2. Keep AI in player wings from doing anything because they were trying to form on the wing of the standalone ship placeholder. 

To clarify, this doesn't mean that they will now start out fighting without order. Instead, they should start out only forming on the wing of their team leader. 

Need to do some more testing to make sure that this is 100% fixed, since Alpha 4 was acting strange during the last test, but otherwise, the initial test looked good.  

Mainly just need to test Team vs. Team on standalone.  This is apparently impossible on one computer, as I just tried and both clients hit an assert.

### DIFF
--- a/code/ai/aigoals.cpp
+++ b/code/ai/aigoals.cpp
@@ -25,6 +25,8 @@
 #include "ship/ship.h"
 #include "weapon/weapon.h"
 
+// Just a reference to this struct being for looking up a ship's team in TVT
+struct ship_registry_entry;
 
 // all ai goals dealt with in this code are goals that are specified through
 // sexpressions in the mission file.  They are either specified as part of a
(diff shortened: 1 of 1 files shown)


JohnAFernandez:
And here's the related Mantis issue. http://scp.indiegames.us/mantis/view.php?id=1815

Hopefully a complete fix for the issue.

code/ai/aigoals.cpp
```diff
@@ -134,28 +136,41 @@ const char *Ai_goal_text(int goal)
...
+	for (j = 0; j < wingp->current_count; j++) {
+		ai_info* aip;
 
-		Assert( wingp->ship_index[j] != -1 );						// get Allender
+		Assert(wingp->ship_index[j] != -1);						// get Allender
 
 		aip = &Ai_info[Ships[wingp->ship_index[j]].ai_index];
 		// don't process Player_ship
-		if ( aip == Player_ai )
+		if (aip == Player_ai || (MULTIPLAYER_MASTER && Objects[Ships[aip->shipnum].objnum].flags[Object::Object_Flags::Player_ship])) {
```
Goober5000, line 153: Does the `MULTIPLAYER_MASTER` check need to be here?  A player ship shouldn't be processed regardless of what type of build is running.

code/ai/aigoals.cpp
```diff
@@ -134,28 +136,41 @@ const char *Ai_goal_text(int goal)
...
+			if (MULTIPLAYER_MASTER && !(Netgame.type_flags & NG_TYPE_DOGFIGHT)) {
+
+				if (Netgame.type_flags & NG_TYPE_TEAM) {
+					const ship_registry_entry* ship_regp = ship_registry_get(Ships[wingp->ship_index[j]].ship_name);
+					ai_add_ship_goal_player(AIG_TYPE_PLAYER_SHIP, AI_GOAL_FORM_ON_WING, -1, Ships[Wings[TVT_wings[ship_regp->p_objp->team]].special_ship].ship_name, aip);
+				}
+				else {
+					ai_add_ship_goal_player(AIG_TYPE_PLAYER_SHIP, AI_GOAL_FORM_ON_WING, -1, Ships[Wings[Starting_wings[0]].special_ship].ship_name, aip);
+				}
+			}	// need to add a form on my wing goal here.  Ships are always forming on the player's wing.
```
Goober5000, line 170: This comment got misplaced.  Put it above the "it is sufficient" line because it describes what the entire if() block is doing.

code/ai/aigoals.cpp
```diff
@@ -134,28 +136,41 @@ const char *Ai_goal_text(int goal)
...
+			// Need to have a more specific target in multi, or they may end up trying to target standalone placeholder.
+			// So let's have them target their team leader.  In dogfight, all player slot ai die, so just exclude.
+			if (MULTIPLAYER_MASTER && !(Netgame.type_flags & NG_TYPE_DOGFIGHT)) {
+
+				if (Netgame.type_flags & NG_TYPE_TEAM) {
+					const ship_registry_entry* ship_regp = ship_registry_get(Ships[wingp->ship_index[j]].ship_name);
+					ai_add_ship_goal_player(AIG_TYPE_PLAYER_SHIP, AI_GOAL_FORM_ON_WING, -1, Ships[Wings[TVT_wings[ship_regp->p_objp->team]].special_ship].ship_name, aip);
+				}
+				else {
+					ai_add_ship_goal_player(AIG_TYPE_PLAYER_SHIP, AI_GOAL_FORM_ON_WING, -1, Ships[Wings[Starting_wings[0]].special_ship].ship_name, aip);
```
Goober5000, line 168: The `special_ship` field refers to the position in the wing, not the ship's shipnum.  But regardless, I am about to remove the `special_ship` field, because it has a fundamental design flaw.  Instead, have the ships form on the wing's ship_index[0]:
```
Ships[Wings[TVT_wings[ship_regp->p_objp->team]].ship_index[0]].ship_name

Ships[Wings[Starting_wings[0]].ship_index[0]].ship_name
```

Also, as an edge case, if for some reason the wing leader is AI-controlled, it will be given an order to form on itself.  That will need to be checked.

JohnAFernandez: Will do

code/ai/aigoals.cpp
JohnAFernandez: The check for multiplayer guarantees two things: one, that the game is in multiplayer so SP will not be affected, and two, that only the server will make these changes.  Since the server takes care of the all of the ai (it gets sent via object update packets), doing otherwise could cause server and client to go out of sync, albeit probably not noticeably.

Goober5000: If only the server should make these changes, then shouldn't you instead put
``
if (!MULTIPLAYER_MASTER)
    return;
```
at the very top of the function?

Comment (JohnAFernandez) [reply code/ai/aigoals.cpp:153]: Hmmmmm.... Yes, that makes sense.  Perhaps in the future someone can go through and further optimize the ai code for clients for similar functions.

File: code/ai/aigoals.cpp (hunk 3 above)
Comment (Goober5000) [reply code/ai/aigoals.cpp:168]: As it turns out, `special_ship` might be fixable, so do this:
```
int wingnum = TVT_wings[ship_regp->p_objp->team];
...Ships[Wings[tvt_wingnum].ship_index[Wings[tvt_wingnum].special_ship]].ship_name...

int wingnum = Starting_wings[0];
...Ships[Wings[wingnum].ship_index[Wings[wingnum].special_ship]].ship_name...
```

Comment (JohnAFernandez) [reply code/ai/aigoals.cpp]: All right. Thanks for putting that together!

File: code/ai/aigoals.cpp [hunk 4]
```diff
@@ -150,25 +156,26 @@ void ai_maybe_add_form_goal(wing* wingp)
 
 		aip = &Ai_info[Ships[wingp->ship_index[j]].ai_index];
 		// don't process Player_ship
-		if (aip == Player_ai || (MULTIPLAYER_MASTER && Objects[Ships[aip->shipnum].objnum].flags[Object::Object_Flags::Player_ship])) {
+		if (aip == Player_ai || ((Game_mode & GM_MULTIPLAYER) && Objects[Ships[aip->shipnum].objnum].flags[Object::Object_Flags::Player_ship])) {
```
Goober5000, line 159: I don't even think you need the multiplayer check here now.  In single-player, only the player ship is the player ship.

JohnAFernandez: So just `if (Objects[Ships[aip->shipnum].objnum].flags[Object::Object_Flags::Player_ship])` to check if it is a player ship?

Goober5000: Right.

JohnAFernandez:
Ok, that should take of the feedback. Once we have had the ability to test this in a team v. team on standalone setting to make sure the behavior is correct, I will switch this to ready for review.

JohnAFernandez:
Team v Team has been tested and works fine.

PR #2594, opened 2020-07-24
Keeps players From Accessing Multi Screens When No Connection Available.

JohnAFernandez:
Leaving the ability to go onto the multi screens causes undesired behavior and asserts. (like trying to validate the tables when there is no connection in the first place.)
  
This also only sets the game state to multi once we're sure that we're going to enter a multi game.

Already tested

### DIFF
--- a/code/menuui/mainhallmenu.cpp
+++ b/code/menuui/mainhallmenu.cpp
@@ -361,6 +361,9 @@ void main_hall_do_multi_ready()
 		return;
 	}
 
+	// Switch to multi here once we're sure that a multi state will be set
+	Game_mode = GM_MULTIPLAYER;
+
 	// go to parallax online
 	if (Multi_options_g.pxo == 1) {
 		Assertion(Multi_options_g.protocol == NET_TCP, "Protocol should always be TCP with PXO!");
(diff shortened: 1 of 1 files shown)


code/menuui/mainhallmenu.cpp
```diff
@@ -882,12 +885,9 @@ void main_hall_do(float frametime)
 				case MULTIPLAYER_REGION:
 					// Make sure we are in multi mode.
 					Player->flags |= PLAYER_FLAGS_IS_MULTI;
```
Goober5000, line 887: Should this flag also be moved?

Alternatively, is this flag properly cleared if there is an error?

notimaginative: Pretty sure this flag is set multiple times before you even get here. It doesn't affect multi state in general, just indicates that the current pilot was last used in multi. Because pilots aren't separated into single/multi any longer this flag is used to say whether we are in single or multi mode (for player select, barracks, etc.)

code/menuui/mainhallmenu.cpp
```diff
@@ -361,6 +361,9 @@ void main_hall_do_multi_ready()
 		return;
 	}
 
+	// Switch to multi here once we're sure that a multi state will be set
+	Game_mode = GM_MULTIPLAYER;
```
notimaginative, line 365: I believe that `Game_mode` is already set to multiplayer by the time you get to the main hall so I'm curious if this really does anything. Maybe there is another issue somewhere too?

JohnAFernandez: I thought about removing or changing this here, but I think the purpose is not just to set it to `GM_Multiplayer` but to also clear out anything that might be accidentally left over before starting a new game.  I'd have to look through it again to be sure, but this could also set the flag correctly if reached in a mainhall that has both single and multiplayer choices, but the pilot was in single player mode previously.

code/menuui/mainhallmenu.cpp
JohnAFernandez: It's also good to have this here in case you start a multi game from a mainhall when the pilot in in SP mode.

JohnAFernandez:
Any other comments or issues?

notimaginative:
Looks good to me

PR #2599, opened 2020-07-25
Fix Non-Standalone Servers Failing to Switch Between PXO and Non-PXO

JohnAFernandez:
Simple fix, already tested.

Adds a function to check if PXO is in use rather than just initialized, so that a very important flag can be properly set.

### DIFF
--- a/code/network/multi_fstracker.cpp
+++ b/code/network/multi_fstracker.cpp
@@ -519,7 +519,7 @@ void multi_fs_tracker_send_game_request()
 // if the API has successfully been initialized and is running
 int multi_fs_tracker_inited()
 {
-	return Multi_fs_tracker_inited;
+	return (Multi_fs_tracker_inited && Multi_options_g.pxo);
 }
 
 // update our settings on the tracker regarding the current netgame stuff


code/network/multi_fstracker.cpp
```diff
@@ -522,6 +522,22 @@ int multi_fs_tracker_inited()
...
+{
+	// if on the standalone, we're either PXO or not, so just check if it's been initalized.
+	if (Game_mode & GM_STANDALONE_SERVER) {
+		return multi_fs_tracker_inited();
+	}
+
+	// on regular servers, the flag is set in the options menu. 
+	if (Multi_options_g.pxo) {
+		return multi_fs_tracker_inited();
+	}
```
Goober5000, line 536: Uh, both of these if() blocks call the same function.

notimaginative: And `Multi_options_g.pxo` is used on both standalone and regular servers, so it's extra redundant.

Probably better to just modify `multi_fs_tracker_inited()` like so:
`return Multi_options_g.pxo && Multi_fs_tracker_inited;`

It would do the same thing and not require an extra function.

code/network/multi_endgame.cpp
```diff
@@ -308,6 +308,11 @@ int multi_quit_game(int prompt, int notify_code, int err_code, int wsa_error)
 		}
 	}		
 
+	// Log out of PXO, if needed
+	if (!(Game_mode & GM_STANDALONE_SERVER) && (Net_player->flags & NETINFO_FLAG_MT_CONNECTED)) {
+		multi_fs_tracker_logout();
+	}
```
notimaginative, line 314: This really shouldn't be needed. Once a game has fully quit it will move to `multi_endgame_cleanup()` where the game will be logged out of the tracker. If there is a condition whereby that doesn't happen, and this change is required, then it's a bug.

notimaginative:
Looks good to me

PR #2600, opened 2020-07-26
add several more scripting functions

Goober5000:
add `ba.rand32()` and `ba.rand32f()`, and `gr.clear()`
enhance `streaminganim:process()`
fix `streaminganim:timeLeft()`

### DIFF
--- a/code/scripting/api/libs/graphics.cpp
+++ b/code/scripting/api/libs/graphics.cpp
@@ -250,6 +250,12 @@ ADE_VIRTVAR(CurrentRenderTarget, l_Graphics, "texture", "Current rendering targe
 	}
 }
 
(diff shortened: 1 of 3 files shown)


code/scripting/api/libs/graphics.cpp
```diff
@@ -250,6 +250,14 @@ ADE_VIRTVAR(CurrentRenderTarget, l_Graphics, "texture", "Current rendering targe
 	}
 }
 
+ADE_FUNC(clear, l_Graphics, nullptr, "Calls gr_clear().  This is sometimes called between setClip and resetClip", nullptr, nullptr)
```
asarium, line 253: That is not a very useful help string. Better would be something like "Fills the entire screen with the currently active color. Useful if you want to have a fresh start for drawing things."-

Goober5000: The documentation I added was the extent of my knowledge. :)  I'll add your suggestion.

PR #2603, opened 2020-07-27
more complete scripting support for debris

Goober5000:
Adds a hook for debris creation, and adds `DoNotExpire` and `LifeLeft` virtual variables for controlling debris removal.

Also add some ship status flags and functions.


code/scripting/scripting.cpp
```diff
@@ -104,6 +104,7 @@ static USED_VARIABLE SCP_vector<std::shared_ptr<BuiltinHook>> Script_actions
 	std::make_shared<BuiltinHook>("On Subsystem Destroyed",	CHA_ONSUBSYSDEATH ),
 	std::make_shared<BuiltinHook>("On Goals Cleared",		CHA_ONGOALSCLEARED ),
 	std::make_shared<BuiltinHook>("On Briefing Stage",		CHA_ONBRIEFSTAGE ),
+	std::make_shared<BuiltinHook>("On Debris Created",		CHA_ONDEBRISCREATED),
```
asarium, line 107: There is a new system for defining hooks that allows to specify documentation about it. See for example `OnMessageReceivedHook` for an example.

I really should add a comment here explaining that...

Goober5000: Ok, changed.

PR #2604, opened 2020-07-27
Supply world coordinates in whack function calls

### LINKED ISSUE #1763: When fired, a weapon with the "apply recoil" flag causes the ship to spin wildly when running a debug build
This does not happen on a release build and is true for nightlies >= 12/21/2017.

Baezon:
This changes all calls to `ship_apply_whack` so that a world position is sent, making them all, most importantly, consistent (this was not true before). All of these paths have been tested to work properly except for the script functions, the scripting implementation is still dark voodoo magic to me, so extra care to ensure I have implemented it properly would be appreciated.

Edit: Also unintentionally fixes #1763 , recoil had always supplied a world position, causing the unusually strong recoil.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -7800,11 +7800,12 @@ static void ship_blow_up_area_apply_blast( object *exp_objp)
 			switch ( objp->type ) {
 			case OBJ_SHIP:
 				ship_apply_global_damage( objp, exp_objp, &exp_objp->pos, damage );
-				vec3d force, vec_ship_to_impact;
+				vec3d force, vec_ship_to_impact, world_hit_pos;
 				vm_vec_sub( &vec_ship_to_impact, &objp->pos, &exp_objp->pos );
 				vm_vec_copy_normalize( &force, &vec_ship_to_impact );
 				vm_vec_scale( &force, blast );
-				ship_apply_whack( &force, &vec_ship_to_impact, objp );
+				vm_vec_add(&world_hit_pos, &vec_ship_to_impact, &objp->pos);
+				ship_apply_whack( &force, &world_hit_pos, objp );
 				break;
 			case OBJ_ASTEROID:
 				asteroid_hit(objp, NULL, NULL, damage);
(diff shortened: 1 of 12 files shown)


code/weapon/weapons.cpp
```diff
@@ -6029,7 +6030,7 @@ void weapon_area_apply_blast(vec3d * /*force_apply_pos*/, object *ship_objp, vec
 			joy_ff_play_vector_effect(&vec_blast_to_ship, blast * 2.0f);
 		}
 	} else {
-		ship_apply_whack( &force, &vec_ship_to_impact, ship_objp);
+		ship_apply_whack( &force, &world_hit_pos, ship_objp);
```
Goober5000, line 6033: You already have `blast_pos` in world coordinates, so you can remove `world_hit_pos` and its arithmetic.

code/ship/ship.cpp
```diff
@@ -16977,6 +16980,7 @@ void object_jettison_cargo(object *objp, object *cargo_objp, float jettison_spee
 
 		// set for relative separation speed (see also do_dying_undock_physics)
 		pos = docker_p0_norm;
+		vm_vec_add2(&pos, &cargo_objp->pos);
```
Goober5000, line 16983: I noticed one of my own bugs here.  The whack is supposed to come from the dockpoint global coordinates, so you should actually add `&objp->pos`.

Baezon: Surely you mean the dockpoint's position, instead of the ship's position? Although after having looked for a bit, it's not clear to me how to get that...

Goober5000: You're right.  I guess what needs to be done is modify `find_adjusted_dockpoint_normal` (which is only used in this one place) to get the dockpoint itself instead of the normal.

If you need help on rewriting it into `find_adjusted_dockpoint`, I can take a look at that later.

Baezon: It looks like `find_adjusted_dockpoint_info` returns the dock points position, although it gives two positions which are very close together and simply averaged in `dock_orient_and_approach`. It seems like this might be I'm looking for?

Goober5000: I based `find_adjusted_dockpoint_normal` on `find_adjusted_dockpoint_info`.  That's why the comment says "// pruned version of below function".

 That being said, computers are faster than they were 15 years ago, and the jettison code will run only once per jettison instance, so you could delete `find_adjusted_dockpoint_normal` and just use the other function.

Baezon: Oh, I see, that's the only place that function is used. Sounds good then.

PR #2605, opened 2020-07-27
Multilock Patch

JohnAFernandez:
Multiplayer Compatibility Portion Of Multilock.

Please get the reviews in now so we can get this merged.

*The reason it's in draft is to be extra careful and I would able to test after addressing feedback but before it's merged.*

### DIFF
--- a/code/network/multi_obj.cpp
+++ b/code/network/multi_obj.cpp
@@ -273,13 +273,8 @@ int multi_oo_pack_client_data(ubyte *data, ship* shipp)
 	char t_subsys, l_subsys;
 	int packet_size = 0;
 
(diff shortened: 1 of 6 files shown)


JohnAFernandez:
If I get another pair of eyes, even if it's just for readability or to do general checkinhg would be good

Goober5000:
I looked at it yesterday.  I didn't see anything wrong with it but I'm not too familiar with the multi code.

JohnAFernandez:
Thanks to everyone who looked it over!

wookieejedi:
Looks like this is good to merge now?

TRBlount:
Agreed.

PR #2608, opened 2020-07-30
IPv6 support

notimaginative:
Adds IPv6 support and related changes. Leaving as draft until some final testing of the newest fixes are done.

NOTE: Drops support for Windows XP!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,7 +103,7 @@ matrix:
       env: CONFIGURATION="Debug"
 
     - os: windows
-      env: CONFIGURATION="Debug" CMAKE_TOOLSET=v141_xp CMAKE_GENERATOR="Visual Studio 15 2017"
+      env: CONFIGURATION="Debug" CMAKE_TOOLSET=v141 CMAKE_GENERATOR="Visual Studio 15 2017"
     - os: windows
       env: CONFIGURATION="Debug" CMAKE_TOOLSET="" CMAKE_GENERATOR="Ninja"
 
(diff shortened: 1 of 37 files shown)


JohnAFernandez:
Just one more test case to make sure this works.  IPv6 connecting with IPv6.  Dahblount and I had a partial success, but we hit another bug that I'm tracking down.

JohnAFernandez:
No issues in testing.  The only thing Keeping this from merging is a review and the build issues on Travis and Appveyor

asarium:
I'm going to push some changes to hopefully fix the CI failures by switching to the non-XP toolset.

asarium:
The Mac build fails with this error:
```
/Users/travis/build/scp-fs2open/fs2open.github.com/code/network/psnet2.cpp:2349:53: error: use of undeclared identifier 'IPV6_ADD_MEMBERSHIP'
                if ( setsockopt(Psnet_mcast_socket, IPPROTO_IPV6, IPV6_ADD_MEMBERSHIP, reinterpret_cast<const char *>(&mRequest), sizeof(mRequest)) ) {
```

notimaginative:
Ah, okay. It's a BSD thing, so I just need to tweak the `#ifdef` that fixes it for FreeBSD.

asarium (approved):
To be honest, I do not know enough about this code to effectively review it but I could not find anything that stood out significantly.

Once this is merged we should probably make an announcement on the forums that we are officially dropping XP support (finally).

PR #2612, opened 2020-07-31
Patch a Crash Point

JohnAFernandez:
Campaign.type is not properly set for MP clients and cannot be relied upon.  This may be a temporary fix until I can research better how to set Campaign.type on clients correctly, if it is possible at all.


JohnAFernandez:
This *should* now have the proper fix.  Once it's clear that multi does not crash with this, I will switch this into ready for review.

JohnAFernandez:
Ghost and Alpha 1 seem to have been able to get through multiple missions with this fix applied.  Setting to ready for review.

Goober5000:
LGTM

PR #2613, opened 2020-07-31
Add 'Process AI before Physics' flag

Baezon:
As is, and far as I've been able to test with a significant chunk of retail, and a few WiH and JAD missions, enabling this flag shouldn't result in any changed behavior. This will lay the groundwork for a better implementation of `ai_turn_towards_vector()` which doesn't result in strong frametime-dependance.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -57,6 +57,7 @@ std::tuple<ubyte, ubyte, ubyte> Arc_color_emp_p1;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_emp_p2;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_emp_s1;
 bool Use_engine_wash_intensity;
+bool Ai_before_physics;
 
 void parse_mod_table(const char *filename)
 {
@@ -104,6 +105,10 @@ void parse_mod_table(const char *filename)
 			stuff_boolean(&Use_tabled_strings_for_default_language);
 
 			mprintf(("Game settings table: Use tabled strings (translations) for the default language: %s\n", Use_tabled_strings_for_default_language ? "yes" : "no"));
+		}
+
+		if (optional_string("$Process AI before physics:")) {
+			stuff_boolean(&Ai_before_physics);
 		}
 
 		optional_string("#CAMPAIGN SETTINGS");
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -595,4 +595,5 @@ void mod_table_reset()
 	Arc_color_emp_p2 = std::make_tuple(static_cast<ubyte>(128), static_cast<ubyte>(128), static_cast<ubyte>(10));
 	Arc_color_emp_s1 = std::make_tuple(static_cast<ubyte>(255), static_cast<ubyte>(255), static_cast<ubyte>(10));
 	Use_engine_wash_intensity = false;
+	Ai_before_physics = false;
 }
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -8856,10 +8856,67 @@ static void lethality_decay(ai_info *aip)
 #endif
 }
 
-void ship_process_pre(object *objp, float frametime)
+void ship_process_pre(object *obj, float frametime)
 {
-	if ( (objp == NULL) || !frametime )
+	if ( (obj == NULL) || !frametime )
 		return;
+
+	int	num;
+	ship* shipp;
+	ship_info* sip;
+
+	if (obj->type != OBJ_SHIP) {
+		nprintf(("Network", "Ignoring non-ship object in ship_process_pre()\n"));
+		return;
+	}
+
+	num = obj->instance;
+	Assert(num >= 0 && num < MAX_SHIPS);
+	Assert(obj->type == OBJ_SHIP);
+	Assert(Ships[num].objnum == OBJ_INDEX(obj));
+
+	shipp = &Ships[num];
+
+	//rotate player subobjects since its processed by the ai functions
+	// AL 2-19-98: Fire turret for player if it exists
+	//WMC - changed this to call process_subobjects
+	if ((obj->flags[Object::Object_Flags::Player_ship]) && !Player_use_ai)
+	{
+		ai_info* aip = &Ai_info[Ships[obj->instance].ai_index];
+		if (aip->ai_flags[AI::AI_Flags::Being_repaired, AI::AI_Flags::Awaiting_repair])
+		{
+			if (aip->support_ship_objnum >= 0)
+			{
+				if (vm_vec_dist_quick(&obj->pos, &Objects[aip->support_ship_objnum].pos) < (obj->radius + Objects[aip->support_ship_objnum].radius) * 1.25f)
+					return;
+			}
+		}
+		if (!shipp->flags[Ship::Ship_Flags::Rotators_locked])
+			process_subobjects(OBJ_INDEX(obj));
+	}
+
+	if (obj == Player_obj) {
+		ship_check_player_distance();
+	}
+
+	// update ship lethality
+	if (Ships[num].ai_index >= 0) {
+		if (!physics_paused && !ai_paused) {
+			lethality_decay(&Ai_info[Ships[num].ai_index]);
+		}
+	}
+
+	// if the ship is an observer ship don't need to do AI
+	if (obj->type == OBJ_OBSERVER) {
+		return;
+	}
+
+	// Goober5000 - player may want to use AI
+	if ((Ships[num].ai_index >= 0) && (!(obj->flags[Object::Object_Flags::Player_ship]) || Player_use_ai)) {
+		if (!physics_paused && !ai_paused) {
+			ai_process(obj, Ships[num].ai_index, frametime);
+		}
+	}
 }
 
 MONITOR( NumShips )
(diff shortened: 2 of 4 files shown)


code/ship/ship.cpp
```diff
@@ -8856,10 +8856,67 @@ static void lethality_decay(ai_info *aip)
...
-	if ( (objp == NULL) || !frametime )
+	// If Ai_before_physics is false everything following is evaluated in ship_process_post()
+	if ( (obj == nullptr) || !frametime || !Ai_before_physics)
 		return;
+
+	int	num;
+	ship* shipp;
+
+	if (obj->type != OBJ_SHIP) {
+		nprintf(("Network", "Ignoring non-ship object in ship_process_pre()\n"));
```
JohnAFernandez, line 8869: There is probably a better category to put this under.  Network is for printing extra multiplayer statements.

Look around, there may be an AI or physics category?

code/ship/ship.cpp
```diff
@@ -8856,10 +8856,67 @@ static void lethality_decay(ai_info *aip)
...
+		return;
+	}
+
+	num = obj->instance;
+	Assert(num >= 0 && num < MAX_SHIPS);
+	Assert(obj->type == OBJ_SHIP);
```
JohnAFernandez, line 8875: Assert(obj->type == OBJ_SHIP); cannot be reached here.  So it isn't needed.

code/ship/ship.cpp
```diff
@@ -9091,24 +9152,24 @@ void ship_process_post(object * obj, float frametime)
...
-		if ( (Ships[num].ai_index >= 0) && (!(obj->flags[Object::Object_Flags::Player_ship]) || Player_use_ai) ){
-			if (!physics_paused && !ai_paused){
-				ai_process( obj, Ships[num].ai_index, frametime );
+		if ((Ships[num].ai_index >= 0) && (!(obj->flags[Object::Object_Flags::Player_ship]) || Player_use_ai)) {
+			if (!physics_paused && !ai_paused) {
+				ai_process(obj, Ships[num].ai_index, frametime);
 			}
 		}
-	}			
+	}
```
JohnAFernandez, line 9172: If you have the time to clean them up, they help keep blame history cleaner.

code/ship/ship.cpp
```diff
@@ -9091,24 +9152,24 @@ void ship_process_post(object * obj, float frametime)
...
 			return;
 		}
 
 		// Goober5000 - player may want to use AI
-		if ( (Ships[num].ai_index >= 0) && (!(obj->flags[Object::Object_Flags::Player_ship]) || Player_use_ai) ){
-			if (!physics_paused && !ai_paused){
-				ai_process( obj, Ships[num].ai_index, frametime );
+		if ((Ships[num].ai_index >= 0) && (!(obj->flags[Object::Object_Flags::Player_ship]) || Player_use_ai)) {
+			if (!physics_paused && !ai_paused) {
+				ai_process(obj, Ships[num].ai_index, frametime);
```
JohnAFernandez, line 9169: More icky white space changes.

code/ship/ship.cpp
```diff
@@ -9091,24 +9152,24 @@ void ship_process_post(object * obj, float frametime)
 		}
 
 		// update ship lethality
-		if ( Ships[num].ai_index >= 0 ){
-			if (!physics_paused && !ai_paused){
+		if (Ships[num].ai_index >= 0) {
+			if (!physics_paused && !ai_paused) {
```
JohnAFernandez, line 9156: Icky white space changes.

code/ship/ship.cpp
```diff
@@ -8856,10 +8856,67 @@ static void lethality_decay(ai_info *aip)
...
+	if (obj->type == OBJ_OBSERVER) {
+		return;
+	}
+
+	// Goober5000 - player may want to use AI
+	if ((Ships[num].ai_index >= 0) && (!(obj->flags[Object::Object_Flags::Player_ship]) || Player_use_ai)) {
+		if (!physics_paused && !ai_paused) {
+			ai_process(obj, Ships[num].ai_index, frametime);
+		}
+	}
```
JohnAFernandez, line 8919: Am I right in thinking that this is a copy of code from elsewhere?  If they have the same exact code, it may be helpful to put both instances of them in a helper function so that changes to one do not have to be duplicated somewhere else.  Unless there are plans for them to diverge at some point.

But maybe someone with more experience can weigh in on that.

code/ship/ship.cpp
Baezon: That makes sense for the -pre() version, since it's literally doing nothing but AI, but would about the -post() one? Is this something you would change for that as well?

JohnAFernandez: So after looking up the categories, imho "General" is probably the best choice here, because it doesn't look like there's anything more specific for physics or object types.

code/mod_table/mod_table.cpp
```diff
@@ -106,6 +107,10 @@ void parse_mod_table(const char *filename)
 			mprintf(("Game settings table: Use tabled strings (translations) for the default language: %s\n", Use_tabled_strings_for_default_language ? "yes" : "no"));
 		}
 
+		if (optional_string("$Process AI before physics:")) {
+			stuff_boolean(&Ai_before_physics);
+		}
```
JohnAFernandez, line 112: On a second look, I can't find where you set this bool's default value. `Use_engine_wash_intensity` is set in `mod_table_reset` in `mod_table.cpp`, for instance.

Goober5000:
Why is this a `game_settings.tbl` flag instead of an `ai_profiles.tbl` flag?

Baezon:
I had initially considered an ai_profiles flag, but those flags are attached to the mission itself, (which might not be an actual issue) and mostly involve 'in the moment' decision making or statistical changes, all of which make it seem like a strange fit for a structural change in the evaluation of AI relative to everything else. I'm definitely not sure exactly of the considerations one would make between the two, but it sort of felt like the odd man out amount the other ai flags. The ai flags always struck me more as "how do you want your ai to behave?" and while slight behavioural changes would occur (later in another pr to take advantage of this flag) the major change is to fix a bug (the framerate problem). But I really would like to hear more opinions on this.

wookieejedi:
I would say that makes sense to me, though I'm just one vote :)

Goober5000:
> I had initially considered an ai_profiles flag, but those flags are attached to the mission itself, (which might not be an actual issue) and mostly involve 'in the moment' decision making or statistical changes, all of which make it seem like a strange fit for a structural change in the evaluation of AI relative to everything else. I'm definitely not sure exactly of the considerations one would make between the two, but it sort of felt like the odd man out amount the other ai flags. The ai flags always struck me more as "how do you want your ai to behave?" and while slight behavioural changes would occur (later in another pr to take advantage of this flag) the major change is to fix a bug (the framerate problem). But I really would like to hear more opinions on this.

My main concern is that mods may have missions balanced with the AI the old way, and then would like to make new missions with the AI balanced the new way.  AI profiles can be changed on a per-mission basis, but game_settings can't.

But there is a case that fundamental changes to game logic belong in game_settings.tbl; see also `$Loop SEXPs Then Arguments:` and `$Use Alternate Chaining Behavior:`.

Baezon:
The explicit intention of this PR is to avoid absolutely any effect on the behavior of the game, balancing-wise or otherwise. Even the later PR, when removing the frametime-dependance, balancing should be absolutely identical. The only reason to have this be in a separate flag at all, is to avoid possibly unintentional bugs being spread into absolutely everything.

The only real behavioral change connected to this would eventually be a flag to use $rotdamp properly, which _would_ strongly alter retail behavior, and should totally be an ai_profiles flag.

code/ship/ship.cpp
```diff
@@ -8856,10 +8856,76 @@ static void lethality_decay(ai_info *aip)
...
+	ship* shipp = &Ships[num];
+
+	if (obj->type != OBJ_SHIP) {
+		nprintf(("AI", "Ignoring non-ship object in ship_process_pre()\n"));
 		return;
+	}
+
+	if ((!(shipp->is_arriving()) || (Ai_info[shipp->ai_index].mode == AIM_BAY_EMERGE)
+		|| ((Warp_params[shipp->warpin_params_index].warp_type == WT_IN_PLACE_ANIM) && (shipp->flags[Ship_Flags::Arriving_stage_2])))
+		&& !(shipp->flags[Ship_Flags::Depart_warp]))
```
Goober5000, line 8925: These conditions don't appear to be part of the original code

code/ship/ship.cpp
```diff
@@ -8856,10 +8856,76 @@ static void lethality_decay(ai_info *aip)
 #endif
 }
 
-void ship_process_pre(object *objp, float frametime)
+// moved out of ship_process_post() so it can be called from either -post() or -pre() depending on Ai_before_physics
+void ship_evaluate_ai(object* obj, float frametime) {
+
+	int num = obj->instance;
```
Goober5000, line 8862: It would be a good idea to add an Assert that `objp->type` is `OBJ_SHIP`

code/weapon/weapons.cpp
```diff
@@ -4470,10 +4470,46 @@ void weapon_home(object *obj, int num, float frame_time)
 	}
 }
 
+
+// moved out of weapon_process_post() so it can be called from either -post() or -pre() depending on Ai_before_physics
+void weapon_update_missiles(object* obj, float  frame_time) {
+
+	weapon* wp = &Weapons[obj->instance];
```
Goober5000, line 4477: It would be a good idea to add an Assert that `objp->type` is `OBJ_WEAPON`

code/ship/ship.cpp
Baezon: They are, you can find them at line 9071 on the current master (9112 on this PR). They're just kind of spaced somewhat away from the AI stuff because it uses the same conditions for some other non-AI stuff.

Goober5000: Ah, ok, thanks.

wookieejedi:
Good to merge?

Goober5000:
Je pense que oui

PR #2618, opened 2020-08-03
Make sure gr_aa_preset_last_frame is initialized correctly

The-E:
Fixes a bug when both -aa and -fxaa or -smaa commandline parameters were specified.


Goober5000:
It's simple enough; I'll take your word for it.

Goober5000 (approved):

PR #2622, opened 2020-08-04
Allow for second-order lead indicator prediction

Baezon:
Adds a flag that will allow the lead indicator to predict changes in predicted position parabolically based on current rotvel. Allows for better prediction for turning targets but still limited by high time-to-target or fast turn speeds (a parabola can only approximate a circle so far). Might add circular prediction for high rotvel targets at some point but would be more complicated.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -595,7 +595,7 @@ void ai_profile_t::reset()
 	subsystem_path_radii = 0;
     bay_arrive_speed_mult = 0;
     bay_depart_speed_mult = 0;
-	second_order_lead_prediction = 0;
+	second_order_lead_predict_factor = 0;
 
     for (int i = 0; i < NUM_SKILL_LEVELS; ++i) {
         max_incoming_asteroids[i] = 0;
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -511,6 +511,13 @@ void parse_ai_profiles_tbl(const char *filename)
 					stuff_boolean(&temp);
                     profile->flags.set(AI::Profile_Flags::Allow_primary_link_at_start, !temp);
 				}
+				if (optional_string("$lead indicator second-order prediction factor:")) {
+					stuff_float(&profile->second_order_lead_predict_factor);
+					if (profile->second_order_lead_predict_factor > 1 || profile->second_order_lead_predict_factor < 0) {
+						mprintf(("Warning: \"$lead indicator second-order prediction factor\" must be 0 - 1, resetting to 0.\"\n"));
+						profile->second_order_lead_predict_factor = 0;
+					}
+				}
 
 
 				// if we've been through once already and are at the same place, force a move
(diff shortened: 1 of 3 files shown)


z64555:
Hold up.  How well does this account for gliding targets, and targets that are side trusting during a turn?

Baezon:
Velocity is luckily in world coords, so it is not affected by the actual forward heading of the ship in question, so it doesn't have any issues with gliding or sliding. Since the eventual newly predicted position of the ship is even more dependent on the current state of the ship, small changes _do_ have an exaggerated effect, causing the the lead indicator to be jitterier in question, and mitigating this is indeed part of the intention of the scalar.

z64555:
Alright.  Looking at it again it assumes the rotvel has an effect on its vel, which as stated before is not the case during glide, but should work with normal mode/atmospheric modes of flight models, and with slower turning craft.

I'm conflicted about including this feature as a hardcode, because it has its flaws for gun sights, autoaim, and AI aiming.  I would reccomend running a test build with the changes in place to see how well the player and AI do vs. Standard in both player vs. AI and AI vs. Ai

wookieejedi:
I'll also note some things I discovered when reviewing this. `polish_predicted_target_pos` is actually only used twice in the whole code. Once for predicting where a shot will land because of auto aim, and the other to draw the lead indicator. When it is called to draw the lead indicator it does not affect any shot placement only the hud gauge rendering position. Thus the only section that `polish_predicted_target_pos` is called where it could affect the angle of shots is line 11107 in `ship.cpp` and it is only checked if auto aim is enabled for the specific ship. Ultimately even if this flag behavior was set by a mod there would be not gameplay difference unless auto aim was also enabled.

Baezon:
This is indeed not used by the AI it seems, I've updated the comment, since I thought it was, but should that be changed? It seems the AI is doing mostly things redundant with this function, and they could easily be updated to use this as well. Perhaps another flag?

wookieejedi:
IIRC the AI does some extra things with the ai table values for `accuracy`, so it would be interesting to see where/how a new change like this would go into that logic flow. 

I'd vote that this stays as a separate flag just for auto-aim/lead indicator for the player. Many folks play with auto aim on and would appreciate the more accurate version, but they don't want the AI to get any more precise since that would break balance in previous missions. The AI related flag would be more useful for in progress missions or mods that have that enabled during mission development.

Baezon:
> 
> 
> Alright. Looking at it again it assumes the rotvel has an effect on its vel, which as stated before is not the case during glide, but should work with normal mode/atmospheric modes of flight models, and with slower turning craft. Even though the phys_info is in the world frame, the way enemy_acc is calculated results in the same issue raised before.
> 
> I'm conflicted about including this feature as a hardcode, because it has its flaws for gun sights, autoaim, and AI aiming. I would reccomend running a test build with the changes in place to see how well the player and AI do vs. Standard in both player vs. AI and AI vs. Ai

Ah, I missed your response, but yes that is a good point, I did not consider that gliding would not result in changes to velocity based on rotvel, but I think this is ok. This is another good reason to not necessarily use 1 for your factor, as this effect would result in over-weighting effect of rotvel, but considering the status quo under-weights this effect, striking somewhere in between would be best. Given that a first-order approximation has done so well for so long, I think these flaws are ok in exchange for some  gain, as long as the flaws are at least noted. The modder should then be able to weight whether or not this kind of thing would be helpful for their situation, and by how much if so.

A better version of this code would use the change in velocity from last frame to this frame, which would also capture linear acceleration as well as being ambivalent towards gliding, but I wanted to avoid adding any extra fields to the state.

wookieejedi:
Sounds like a good option for mods to have. I know Fate of the Galaxy will be using this feature, too.

@z64555 Sound good to you? If so I'll merge it.

PR #2631, opened 2020-08-05
Make redalert missions work better in multi

### LINKED ISSUE #2560: Multi + Red Alert : Dialog text is invisible
When a mission is in red alert mode, dialog text is invisible, but you can still hear the audio.

normal
screen0015

in red alert
screen0017

notimaginative:
Still needs a little more testing to verify it works properly on standalone, host, and clients.

Should address the following issues:
 - support sync state before entering mission
 - return to briefing for sync and stats saving
 - handle campaign progression properly
 - initialize net_player state properly (fixes #2560)

### DIFF
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -28,6 +28,7 @@
 #include "missionui/missionweaponchoice.h"
 #include "missionui/redalert.h"
 #include "network/multiteamselect.h"
+#include "network/multi_endgame.h"
 #include "mod_table/mod_table.h"
 #include "model/model.h"
 #include "object/deadobjectdock.h"
@@ -409,7 +410,12 @@ void red_alert_do_frame(float frametime)
 	switch (k) {
 		case KEY_ESC:
 //			gameseq_post_event(GS_EVENT_ENTER_GAME);
-			gameseq_post_event(GS_EVENT_MAIN_MENU);
+			if (Game_mode & GM_MULTIPLAYER) {
+				gamesnd_play_iface(InterfaceSounds::USER_SELECT);
+				multi_quit_game(PROMPT_ALL);
+			} else {
+				gameseq_post_event(GS_EVENT_MAIN_MENU);
+			}
 			break;
 	}	// end switch
 
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -27,6 +27,7 @@
 #include "missionui/missionscreencommon.h"
 #include "missionui/missionweaponchoice.h"
 #include "missionui/redalert.h"
+#include "network/multiteamselect.h"
 #include "mod_table/mod_table.h"
 #include "model/model.h"
 #include "object/deadobjectdock.h"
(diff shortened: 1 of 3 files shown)


Goober5000:
Code-wise this LGTM, but I think we need @JohnAFernandez to validate with additional testing.

JohnAFernandez:
Has not been tested yet, but I just added it to the testing branch, so I'll rebuild and send it to the testers now.

notimaginative:
This is confirmed to be broken on the client, so a few tweaks are still needed

Goober5000 (approved):
Looks good to me.

JohnAFernandez:
Issues were actually a sexp bug that's been fixed as of this morning.  I'm going to run through it one more time right now.  

And I found a bug.  Really glad I went through it!!  Looks like we forgot to disable clients pressing commit on the red-alert briefing screen.  It's allowing the client to go to the mission without the server.

JohnAFernandez (changes requested):
We have to add logic to the red alert brief screen so that clients are barred from entering the gameplay state without the server.

notimaginative:
I thought I fixed that. Either way it shouldn't be possible for a client to enter the game without the server. At the very least they may be able to get to the post-briefing sync screen, but the host is required to start the mission. Unless that was broken and I didn't notice it.

notimaginative:
Just tested it again, and unless I'm missing something, everything is working as intended. Clients should wait at the post-brief sync screen until the host gets there and hits the launch button. Clients cannot proceed directly into a mission.

I did have changes which were scrapped that would have made the client wait at the red alert brief until the host has clicked continue. But unlike the regular mission briefing there is no feedback for this, so the client just has to keep clicking continue until it works. The chatbox also doesn't look very good on the red alert UI so I avoided adding that.

Having the client just click through to the post-brief sync screen seemed to be the cleaner option. And it has worked without issue in my testing so far. Is there a test case I'm missing that allows to client to proceed past that sync screen?

JohnAFernandez:
What I meant is that I was able to click through the red alert brief before the server was in the mission, and it crashed the server.

notimaginative:
The clients should just wait at the post-brief sync screen until the host arrives to start the mission. I've tested a client that clicked through while the the host stayed at the red alert briefing for 5 minutes without issue. Was this on a standalone? That's one scenario I didn't thoroughly check yet. I'll do some more testing tonight and see if I can replicate a crash. What is the crash btw?

The chatbox is easy to add with very minor code changes, it's just that it covers part of the red alert UI since that UI wasn't designed with multi in mind. It's not absolutely horrible, but it's not pretty either. But since there isn't anything to do on that screen but read some text and click one button I didn't see much use in having it there. The post-brief sync UI does have a chatbot so I figured that would be enough.

JohnAFernandez:
I'll run through it again soon and see if there was a particular set of circumstances that was causing it. It was not on standalone.  I was playing Shivansps's red alert test campaign.

It was a straight crash to desktop.

JohnAFernandez (approved):

JohnAFernandez:
I can no longer reproduce the bug that was holding these up.  Either something changed, or it is a rare edge case bug that will be hard to track down, but which shouldn't hold us up from merging.

PR #2632, opened 2020-08-05
Restored retail-like handling of speed during freeflight time.

EatThePath:
During the free-flight period the coast speed of a missile was being calculated based on the free-flight time, rather than based on the final speed. This returns the logic to what it was in retail, as besides the enclosing condition there isn't any evident reason for the duration to be a factor here.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -4178,9 +4178,9 @@ void weapon_home(object *obj, int num, float frame_time)
 		// since free_flight_time can now be 0, guard against that
 		else if (wip->free_flight_time > 0.0f) {
 			if (obj->phys_info.speed > max_speed) {
-				obj->phys_info.speed -= frame_time * (2 / wip->free_flight_time);
-			} else if ((obj->phys_info.speed < max_speed / (2 / wip->free_flight_time)) && (wip->wi_flags[Weapon::Info_Flags::Homing_heat])) {
-				obj->phys_info.speed = max_speed / (2 / wip->free_flight_time);
+				obj->phys_info.speed -= frame_time * 4;
+			} else if ((obj->phys_info.speed < max_speed / 4) && (wip->wi_flags[Weapon::Info_Flags::Homing_heat])) {
+				obj->phys_info.speed = max_speed / 4;
 			}
 		}
 		// no free_flight_time, so immediately set desired speed


code/weapon/weapons.cpp
```diff
@@ -4178,9 +4178,9 @@ void weapon_home(object *obj, int num, float frame_time)
 		// since free_flight_time can now be 0, guard against that
 		else if (wip->free_flight_time > 0.0f) {
 			if (obj->phys_info.speed > max_speed) {
-				obj->phys_info.speed -= frame_time * (2 / wip->free_flight_time);
-			} else if ((obj->phys_info.speed < max_speed / (2 / wip->free_flight_time)) && (wip->wi_flags[Weapon::Info_Flags::Homing_heat])) {
-				obj->phys_info.speed = max_speed / (2 / wip->free_flight_time);
+				obj->phys_info.speed -= frame_time * 4;
```
Baezon, line 4181: Looks like this one should also be `max_speed / 4`

EatThePath: To be frank I'd do it that way if starting from scratch, but this is the equation used for this case in the oldest code in the repo, and I figured restoring whatever behavior that produces would be preferred. I think this case should really only come up if you've got a fighter significantly faster than the missiles you're launching, anyway.

Old code link: https://github.com/scp-fs2open/fs2open.github.com/blob/663b3471b9e14c31e7102af8b3f76c1aa465c932/code/weapon/weapons.cpp#L1887

Baezon: I don't think reverting to the very oldest situation is necessarily the best answer here, but I may defer that choice to others. Regardless, this will have a negligible effect on the speed of the missile. Frametime is likely to be in the ballpark of 0.04, and multiplied by 4 will do practically nothing to the missiles current speed (which since it should have been just fired at this point so max_speed / 4), in ball park of 20-100, even over dozens of frames.

EatThePath: looking at it that way, this would mean that a missile somehow fired over it's max speed would slow down at a rate of 4 m/s per second. That doesn't strike me as very useful, but it also doesn't strike me as dangerous either.

Goober5000: Here is the original retail code:
```
	//	If not 1/2 second gone by, don't home yet.
	if ((hobjp == &obj_used_list) || ( f2fl(Missiontime - wp->creation_time) < 0.25f )) {
		//	If this is a heat seeking homing missile and 1/2 second has elapsed since firing
		//	and we don't have a target (else we wouldn't be inside the IF), find a new target.
		if (wip->wi_flags & WIF_HOMING_HEAT)
			if ( f2fl(Missiontime - wp->creation_time) > 0.5f )
				find_homing_object(obj, num);

		if (obj->phys_info.speed > wip->max_speed) {
			obj->phys_info.speed -= frame_time * 4;
			vm_vec_copy_scale( &obj->phys_info.desired_vel, &obj->orient.fvec, obj->phys_info.speed);
		} else if ((obj->phys_info.speed < wip->max_speed/4) && (wip->wi_flags & WIF_HOMING_HEAT)) {
			obj->phys_info.speed = wip->max_speed/4;
			vm_vec_copy_scale( &obj->phys_info.desired_vel, &obj->orient.fvec, obj->phys_info.speed);
		}

/*	Removed code that makes bombs drop for a bit.  They looked odd and it was confusing.  People wondered where their weapons went.
		//	Make bombs drop down for first second of life.
		if (wip->wi_flags & WIF_BOMB) {
			if (wip->lifetime - wp->lifeleft < 0.5f) {
				float	time_scale = wip->lifetime - wp->lifeleft;
				vm_vec_scale_add2(&obj->phys_info.desired_vel, &obj->orient.uvec, (time_scale - 0.5f) * max(10.0f, obj->phys_info.speed/2.0f));
			}
		}
*/
		return;
	}
```

EatThePath: Discussion on discord indicates that restoring retail should be the priority and then any deviations to that can be added behind flags of one sort or another, which seems suited to separate PRs if it is done. 

Are there any other issues here?

Goober5000: LGTM.

PR #2637, opened 2020-08-08
Enable dllexport for Lua API

asarium:
This allows external libraries to load the Lua API from our executable
even if it is linked statically. This is useful for a Lua debugger which
needs to interact with the Lua API.

### DIFF
--- a/lib/lua/luaconf.h
+++ b/lib/lua/luaconf.h
@@ -161,7 +161,19 @@
 
 #else
 
+#ifdef WIN32
+// Export even when building statically to allow external libraries to load the Lua API symbols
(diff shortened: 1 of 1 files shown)


Goober5000:
Oh interesting.  Also with this quote from Discord...

> the debugger tries to load the Lua API from the FSO process but that is compiled with a static Lua library so normally those functions are not exported

That would explain why the Lua debugger I was running through Visual Studio said that symbols weren't loaded for the Lua script.  It still traced through functions fine, though. ¯\_(ツ)_/¯

asarium:
Might also be an issue with your Lua debugger since this fix was only for one specific type of debugger.

PR #2640, opened 2020-08-09
Improve vm_orthogonalize_matrix safety

Baezon:
And also add a unit test for it. Currently, orthogonalize is not guaranteed to actually return an orthogonal matrix if two of the vectors are the same, so the logic is tightened up to ensure orthogonal vectors while retaining as much of the original matrix as possible. This _appears_ to obsolete fix_matrix as well, but it seems like it's used by FRED to have different vector preferences.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -1150,6 +1150,8 @@ void find_point_on_line_nearest_skew_line(vec3d *dest, const vec3d *p1, const ve
 	vm_vec_scale_add(dest, p1, d1, numerator / denominator);
 }
 
+
+// normalizes only if above a small threshold, returns if normalized or not
 bool vm_maybe_normalize(vec3d* dst, const vec3d* src) {
 	float mag = vm_vec_mag(src);
 	if (mag < SMALL_NUM) return false;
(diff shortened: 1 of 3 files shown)


code/math/vecmat.cpp
```diff
@@ -1150,47 +1150,66 @@ void find_point_on_line_nearest_skew_line(vec3d *dest, const vec3d *p1, const ve
...
-	if (umag <= 0.0f) {  // no up vector to use..
-		if (rmag <= 0.0f) {  // no right vector either, so make something up
-			if (!m->vec.fvec.xyz.x && !m->vec.fvec.xyz.z && m->vec.fvec.xyz.y)  // vertical vector
-				vm_vec_make(&m->vec.uvec, 0.0f, 0.0f, 1.0f);
-			else
-				vm_vec_make(&m->vec.uvec, 0.0f, 1.0f, 0.0f);
+// normalizes only if above a small threshold, returns if normalized or not
+bool vm_maybe_normalize(vec3d* dst, const vec3d* src) {
+	float mag = vm_vec_mag(src);
+	if (mag < SMALL_NUM) return false;
```
TRBlount, line 1157: This should probably be `if ( abs(1.0f-mag) < std::numeric_limits<float>::epsilon() ) return false;`

Goober5000: `SMALL_NUM` is used in several places in vecmat.cpp; I think that's acceptable.

Goober5000: And incidentally, neither `abs` nor `1-mag` is the right thing to do here.

PR #2644, opened 2020-08-10
Add scripting API for custom execution context

asarium:
This allows a script to create its own execution context so that it can
control when a coroutine is valid to play based on custom data.

Also adds a function which causes a coroutine to end in an error.


code/scripting/api/objs/enums.cpp
```diff
@@ -485,21 +485,12 @@ flag_def_list Enumerations[] = {
...
-		"AUDIOSTREAM_VOICE",
-		LE_ASF_VOICE,
-		0
-	},
+	{"AUDIOSTREAM_EVENTMUSIC", LE_ASF_EVENTMUSIC, 0},
+	{"AUDIOSTREAM_MENUMUSIC", LE_ASF_MENUMUSIC, 0},
+	{"AUDIOSTREAM_VOICE", LE_ASF_VOICE, 0},
+	{"CONTEXT_VALID", LE_CONTEXT_VALID, 0},
+	{"CONTEXT_SUSPENDED", LE_CONTEXT_SUSPENDED, 0},
+	{"CONTEXT_INVALID", LE_CONTEXT_INVALID, 0},
```
Goober5000, line 493: Why did this get reformatted?  (The rest of the list stayed the same.)

asarium: I think our clang-format rules changed at some point which took this with it. Since I like the compact form a bit more I reformatted the entire list and also added the clang-format disable comments around it so that this doesn't get reformatted again.

PR #2645, opened 2020-08-10
Add more scripting APIs for the command briefing

asarium:
This covers the various APIs needed for getting the last parts of the
new command briefing UI to work.

### DIFF
--- a/code/missionui/missionscreencommon.cpp
+++ b/code/missionui/missionscreencommon.cpp
@@ -313,27 +313,37 @@ void set_active_ui(UI_WINDOW *ui_window)
 	Active_ui_window = ui_window;
 }
 
-void common_music_init(int score_index)
+SCP_string common_music_get_filename(int score_index)
 {
-	if ( Cmdline_freespace_no_music ) {
-		return;
+	if (Cmdline_freespace_no_music) {
+		return SCP_string();
 	}
 
-	if ( score_index >= NUM_SCORES ) {
-		Int3();
-		return;
-	}
+	Assertion(score_index >= 0 && score_index < NUM_SCORES, "Invalid score index %d.", score_index);
 
-	if ( Mission_music[score_index] < 0 ) {
-		if ( Num_music_files > 0 ) {
+	if (Mission_music[score_index] < 0) {
+		if (Num_music_files > 0) {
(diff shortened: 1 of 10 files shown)


code/missionui/missionscreencommon.cpp
```diff
@@ -313,27 +313,40 @@ void set_active_ui(UI_WINDOW *ui_window)
...
 {
-	if ( Cmdline_freespace_no_music ) {
-		return;
+	if (Cmdline_freespace_no_music) {
+		return SCP_string();
 	}
 
-	if ( score_index >= NUM_SCORES ) {
+	if (score_index >= NUM_SCORES) {
 		Int3();
```
Goober5000, line 323: I'd use this opportunity to get rid of the Int3() here.

code/scripting/api/libs/ui.cpp
```diff
@@ -390,5 +392,10 @@ ADE_FUNC(getBriefing,
 	return ade_set_args(L, "o", l_CmdBrief.Set(Cmd_briefs[0]));
 }
 
+ADE_FUNC(getBriefingMusicName, l_UserInterface_CmdBrief, nullptr, "Gets the file name of the music file to play for the briefing.", "string", "The file name or empty if no music")
+{
+	return ade_set_args(L, "s", common_music_get_filename(SCORE_BRIEFING).c_str());
+}
+
```
Goober5000, line 399: What about making this a virtvar, or adding a corresponding setBriefingMusic?  I'm assuming there's an opportunity for scripts to run after a mission is loaded but before the briefing starts.

code/missionui/missionscreencommon.cpp
asarium: Done, replaced with Assertion.

code/scripting/api/libs/ui.cpp
asarium: This API is specifically meant for the libRocket API and as such I try to keep is as minimal as possible. It is also not guaranteed to be backwards compatible since I would like to replace this with an actual Model-View architecture at some point.

Goober5000: Ok.

PR #2646, opened 2020-08-10
Banish the Black Void

Baezon:
Since the whacking for docked ships code now needs to invert ships inv_moi's where never before this was needed, the numbers produced could end up much too small for floating points and get inverted into infinity and it all goes bad from there. This turns inverse_matrix's calculations into doubles to handle the entire range of floats (the determinant ends up with cubes of the original inv_moi, limiting its accuracy and range significantly when constrained to a float). Any ship with moi values smaller than 1.4323644 × 10^-13 were affected.

This also provides special cases to handle zero (or any other non-invertable) MOIs properly as well, if it is attempted to invert.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -2274,14 +2274,14 @@ void vm_estimate_next_orientation(const matrix *last_orient, const matrix *curre
 }
 
 //	Return true if all elements of *vec are legal, that is, not NaN or infinity.
-int is_valid_vec(const vec3d *vec)
+bool is_valid_vec(const vec3d *vec)
 {
 	return !std::isnan(vec->xyz.x) && !std::isnan(vec->xyz.y) && !std::isnan(vec->xyz.z)
 		&& !std::isinf(vec->xyz.x) && !std::isinf(vec->xyz.y) && !std::isinf(vec->xyz.z);
 }
 
 //	Return true if all elements of *m are legal, that is, not a NAN.
-int is_valid_matrix(const matrix *m)
+bool is_valid_matrix(const matrix *m)
 {
 	return is_valid_vec(&m->vec.fvec) && is_valid_vec(&m->vec.uvec) && is_valid_vec(&m->vec.rvec);
 }
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -2509,20 +2509,21 @@ void vm_matrix_sub2(matrix* dest, const matrix* src)
 */
 bool vm_inverse_matrix(matrix* dest, const matrix* m)
 {
-	matrix inv;	// create a temp matrix so we can avoid getting a determinant that is 0
+	// Use doubles here because this is used for ship inv_mois and we could be dealing with extremely small numbers
+	double inv[3][3];	// create a temp matrix so we can avoid getting a determinant that is 0
 
 	// Use a2d so it's easier for people to read
-	inv.a2d[0][0] = -m->a2d[1][2] * m->a2d[2][1] + m->a2d[1][1] * m->a2d[2][2];
-	inv.a2d[0][1] = m->a2d[0][2] * m->a2d[2][1] - m->a2d[0][1] * m->a2d[2][2];
-	inv.a2d[0][2] = -m->a2d[0][2] * m->a2d[1][1] + m->a2d[0][1] * m->a2d[1][2];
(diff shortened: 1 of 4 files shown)


code/math/vecmat.cpp
```diff
@@ -2509,29 +2509,32 @@ void vm_matrix_sub2(matrix* dest, const matrix* src)
...
-	inv.a2d[0][2] = -m->a2d[0][2] * m->a2d[1][1] + m->a2d[0][1] * m->a2d[1][2];
-	inv.a2d[1][0] = m->a2d[1][2] * m->a2d[2][0] - m->a2d[1][0] * m->a2d[2][2];
-	inv.a2d[1][1] = -m->a2d[0][2] * m->a2d[2][0] + m->a2d[0][0] * m->a2d[2][2];
-	inv.a2d[1][2] = m->a2d[0][2] * m->a2d[1][0] - m->a2d[0][0] * m->a2d[1][2];
-	inv.a2d[2][0] = -m->a2d[1][1] * m->a2d[2][0] + m->a2d[1][0] * m->a2d[2][1];
-	inv.a2d[2][1] = m->a2d[0][1] * m->a2d[2][0] - m->a2d[0][0] * m->a2d[2][1];
-	inv.a2d[2][2] = -m->a2d[0][1] * m->a2d[1][0] + m->a2d[0][0] * m->a2d[1][1];
-
-	float det = m->a2d[0][0] * inv.a2d[0][0] + m->a2d[0][1] * inv.a2d[1][0] + m->a2d[0][2] * inv.a2d[2][0];
+	inv[0][0] = -(double)m->a2d[1][2] * m->a2d[2][1] + (double)m->a2d[1][1] * m->a2d[2][2];
```
Goober5000, line 2516: I'd recommend putting `double` on all four operands here.  The compiler probably promotes everything to double before doing the math, but putting double everywhere makes it explicit for people reading the code.

code/object/objectdock.cpp
```diff
@@ -836,7 +844,11 @@ void dock_calc_total_moi_helper(object* objp, dock_function_info* infop)
 	// So for each part:
 
 	// We invert the inverse MOI to get an MOI in the local frame
-	vm_inverse_matrix(&local_moi, &objp->phys_info.I_body_inv);
+	if (!vm_inverse_matrix(&local_moi, &objp->phys_info.I_body_inv)) {
+		// This is done on purpose to indicate a zero inv_moi
+		infop->maintained_variables.matrix_value->a1d[0] = NAN;
+		return;
+	}
```
Goober5000, line 851: As an alternative, you could set `maintained_variables.bool_value`.  Not saying you should, just be aware of it if you hadn't considered it.

Returning NAN makes mathematical sense but is riskier from an implementation point of view.  However your additional comment takes care of that.

code/math/vecmat.cpp
```diff
@@ -2509,29 +2509,32 @@ void vm_matrix_sub2(matrix* dest, const matrix* src)
 */
 bool vm_inverse_matrix(matrix* dest, const matrix* m)
 {
-	matrix inv;	// create a temp matrix so we can avoid getting a determinant that is 0
+	// Use doubles here because this is used for ship inv_mois and we could be dealing with extremely small numbers
+	double inv[3][3];	// create a temp matrix so we can avoid getting a determinant that is 0
```
Goober5000, line 2513: You could do this:
```
union {
    double a2d[3][3];
    double a1d[9];
} inv;
```
That would allow you to keep the a2d and a1d references from before.

The main reason to do this is to keep the single loop instead of the nested loop.  This would be faster, but I don't think this is speed-critical code (unless you're inverting matrixes every frame somewhere) so it's optional.

code/object/objectdock.cpp
Baezon: I think I'll keep the NaN return from the helper function, since that's NaN's purpose after all, but I'll change the dock_calc_total_moi to instead catch the NaN and return a bool, to make it a bit more straightforward, if anyone else ends up needing it.

code/math/vecmat.cpp
Baezon: It might be slightly faster, but I think it's easier to understand if the code itself mimics the actual structure of the matrix.

code/object/objectdock.cpp
Goober5000: Okay, but keep the comment that says the `dest` matrix may be NAN if the function returns false.

PR #2649, opened 2020-08-11
Debris enhancements

Goober5000:
1) Make Debris dynamic, and use a soft limit at 64 pieces (the old MAX) where the game will try to remove old pieces
2) Add scripting functions to tell whether the debris is a tiny generic/vaporized piece, rather than a modeled piece of ship
3) Allow per-class specification of death roll sounds and subsystem explosion sounds
4) Fix an off-by-one bug in the Debris[] Lua indexer

### DIFF
--- a/code/scripting/api/objs/debris.cpp
+++ b/code/scripting/api/objs/debris.cpp
@@ -8,9 +8,6 @@
 #include "globalincs/linklist.h"
 #include "ship/ship.h"
 
-extern int Num_hull_pieces;
-extern debris Hull_debris_list;
-
 namespace scripting {
 namespace api {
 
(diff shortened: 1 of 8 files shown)


code/scripting/api/objs/debris.cpp
```diff
@@ -11,6 +11,9 @@
 extern int Num_hull_pieces;
 extern debris Hull_debris_list;
 
+extern int Debris_model;
+extern int Debris_vaporize_model;
```
asarium, line 15: Please move the code that needs this into `debris.cpp` to avoid having extern variables.

Goober5000: Is externing really a big problem for scripting?  There are two more extern'd variables right above it.

Goober5000: I removed all four extern'd variables.

PR #2652, opened 2020-08-14
Update DDS Utilities to DX10 Standard

### LINKED ISSUE #2633: Proposal: Update DDSUtils to DX10/11
Our current implementation of DDS is using the headers defined in the standard for DX7. We could support additional block compression techniques and texture arrays if we were to update the current implementation to DX10/11.

TRBlount:
This replaces old DDS headers with the new headers that are used with DX10+ compliant programs. In addition to providing some future proofing in terms of structures like texture arrays and volume textures stored in a single DDS, this also allows the use of newer block compression techniques. Currently, BC7 block compression is enabled and appears to be working. To give an example comparing BC7 to previously allowed block compression techniques, BC7 is the top image and BC1/DXT1 is the bottom image, both images are run through a difference filter with the source image:

comparison

This will close #2633.


code/graphics/2d.cpp
```diff
@@ -1597,6 +1597,14 @@ bool gr_init(std::unique_ptr<os::GraphicsOperations>&& graphicsOps, int d_mode,
 	mprintf(("  Persistent buffer mapping: %s\n",
 	         gr_is_capable(CAPABILITY_PERSISTENT_BUFFER_MAPPING) ? "Enabled" : "Disabled"));
 
+	mprintf(("Checking mod required rendering features...\n"));
+	for (int ext : Required_render_ext) {
+		if (!gr_is_capable(static_cast<gr_capability>(ext))) {
+			Error(LOCATION, "  Feature %d required by mod not supported by system.\n", ext);
```
asarium, line 1603: Since this is a user facing error message it would be good to have a textual representation of the feature that is missing. Also, why are there two leading spaces and a `\n` at the end?

code/mod_table/mod_table.cpp
```diff
@@ -58,6 +58,11 @@ std::tuple<ubyte, ubyte, ubyte> Arc_color_emp_p2;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_emp_s1;
 bool Use_engine_wash_intensity;
 bool Ai_before_physics;
+SCP_vector<int> Required_render_ext;
```
asarium, line 61: Why not use the enum type here? Same a few lines below. Since `CAPABILITY_BPTC` is already valid, that type should also be available.

code/mod_table/mod_table.cpp
```diff
@@ -58,6 +58,11 @@ std::tuple<ubyte, ubyte, ubyte> Arc_color_emp_p2;
...
+
+SCP_vector<std::pair<SCP_string, int>> req_render_ext_pairs = {
+	std::make_pair("BPTC", static_cast<int>(CAPABILITY_BPTC))
```
asarium, line 64: Maybe call this "BPTC Texture Compression" to make it more explicit what it means.

code/graphics/2d.cpp
TRBlount: The two leading spaces is pretty standard formatting across the `gr_init` calls to mprintf (look a few lines higher for example).

code/mod_table/mod_table.cpp
TRBlount: MSVC couldn't seem to find `gr_capability` despite the fact that Intellisense could detect it and didn't throw any errors, hence why I changed it to an int.

code/graphics/2d.cpp
asarium: Yes, but this isn't an `mprintf` call.

code/mod_table/mod_table.cpp
asarium: What was the error in this case?

TRBlount: I don't remember precisely but it resulted in an undefined symbol error towards the end of compilation.

code/graphics/2d.cpp
TRBlount: Alright, I'll change it. Also, where would be the proper place to put a helper function?

asarium: Probably here in 2d.cpp. The enum is also defined here so that makes sense to me.

code/mod_table/mod_table.cpp
TRBlount:
```
>C:\Users\tyler\Documents\GitHub\fs2open.github.com\code\mod_table/mod_table.h(52,19): error C2065: 'gr_capability': undeclared identifier (compiling source file C:\Users\tyler\Documents\GitHub\fs2open.github.com\code\scripting\api\objs\weaponclass.cpp)
>C:\Users\tyler\Documents\GitHub\fs2open.github.com\code\mod_table/mod_table.h(52,8): error C2923: 'SCP_vector': 'gr_capability' is not a valid template type argument for parameter 'T' (compiling source file C:\Users\tyler\Documents\GitHub\fs2open.github.com\code\scripting\api\objs\weaponclass.cpp)
>C:\Users\tyler\Documents\GitHub\fs2open.github.com\code\mod_table/mod_table.h(52,34): error C2062: type 'unknown-type' unexpected (compiling source file C:\Users\tyler\Documents\GitHub\fs2open.github.com\code\scripting\api\objs\weaponclass.cpp)
```

PR #2663, opened 2020-08-17
make Intel stuff dynamic

Goober5000:
1) Make Intel_info a vector instead of a fixed-size array
2) Make intel descriptions SCP_string so they can be any length

### DIFF
--- a/code/menuui/techmenu.cpp
+++ b/code/menuui/techmenu.cpp
@@ -215,9 +215,9 @@ static int Techroom_ship_modelnum;
 static float Techroom_ship_rot;
 static UI_BUTTON List_buttons[LIST_BUTTONS_MAX];  // buttons for each line of text in list
 
-static int Ships_loaded = 0;
-static int Weapons_loaded = 0;
-static int Intel_loaded = 0;
+static bool Ships_loaded = false;
+static bool Weapons_loaded = false;
+static bool Intel_loaded = false;
 
 int Techroom_overlay_id;
 
@@ -225,7 +225,7 @@ int Techroom_overlay_id;
 typedef struct {
 	int	index;		// index into the master table that its in (ie Ship_info[])
 	const char* name;			// ptr to name string
-	char* desc;			// ptr to description string
+	const char* desc;			// ptr to description string
 	char tech_anim_filename[MAX_FILENAME_LEN];	//duh
 	generic_anim animation;	// animation info
 	int	bitmap;		// bitmap handle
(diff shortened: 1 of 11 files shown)


code/menuui/techmenu.cpp
```diff
@@ -896,35 +895,43 @@ void techroom_change_tab(int num)
 		case INTEL_DATA_TAB:
 
 			// load intel if necessary
-			if ( Intel_loaded == 0 ) {
-				// now populate the entry structs
+			if ( !Intel_loaded ) {
+				if (Intel_list == nullptr) {
+					Intel_list = new tech_list_entry[Intel_info.size()];
+
+					if (Intel_list == nullptr)
```
asarium, line 902: This will never be true since a failure to allocate the data with throw a `std::bad_alloc` exception. You could ass `std::nothrow` to the `new` but I don't really see the point in handling that since we always assume to have enough memory.

code/menuui/techmenu.cpp
```diff
@@ -896,35 +895,43 @@ void techroom_change_tab(int num)
 		case INTEL_DATA_TAB:
 
 			// load intel if necessary
-			if ( Intel_loaded == 0 ) {
-				// now populate the entry structs
+			if ( !Intel_loaded ) {
+				if (Intel_list == nullptr) {
+					Intel_list = new tech_list_entry[Intel_info.size()];
```
asarium, line 900: Any particular reason to use dynamic memory here instead of a vector? I looked at the relevant code and nothing jumped out at me to suggest that we need that dynamic memory here.

code/scripting/api/objs/intelentry.cpp
```diff
@@ -65,16 +65,14 @@ ADE_VIRTVAR(Description, l_Intelentry, "string", "Intel entry description", "str
...
 
 	if(ADE_SETTING_VAR) {
 		if(s != nullptr) {
-			auto len = sizeof(Intel_info[idx].desc);
-			strncpy(Intel_info[idx].desc, s, len);
-			Intel_info[idx].desc[len - 1] = 0;
+			Intel_info[idx].desc = s;
 		} else {
-			strcpy_s(Intel_info[idx].desc, "");
+			Intel_info[idx].desc = "";
```
asarium, line 75: You could also use `.clear()` here.

code/menuui/techmenu.cpp
Goober5000: Good point.  I'll change this.

code/menuui/techmenu.cpp
Goober5000: Two reasons:
1) The ship and weapons lists are handled the same way
2) `Current_list` points to whatever list is being handled in the tech room.  That is of type `tech_list_entry *` so the intel list needs to match it

code/scripting/api/objs/intelentry.cpp
Goober5000: Ok.

Goober5000: Good to know.  I like the empty string though because it's explicit.

code/menuui/techmenu.cpp
asarium: You can still get a pointer to the contents of a vector but this works as well.

asarium (approved):

Goober5000:
I'll have to rebase this once #2666 is merged.

PR #2665, opened 2020-08-18
Don't trigger ship_start/stop_secondary_fire assert on empty bank

TRBlount:
Overlooked this detail from the multilock PR. Empty banks are allowed to exist and should be treated as "unable to fire".

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -6131,9 +6131,9 @@ void ship_weapon::clear()
     num_secondary_banks = 0;
     num_tertiary_banks = 0;
 
-    current_primary_bank = -1;
-    current_secondary_bank = -1;
-    current_tertiary_bank = -1;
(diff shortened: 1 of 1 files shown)


code/ship/ship.cpp
```diff
@@ -19532,7 +19536,12 @@ bool ship_stop_secondary_fire(object* objp)
 		return false;
 	}
 
-	Assert( (swp->secondary_bank_weapons[bank] >= 0) && (swp->secondary_bank_weapons[bank] < weapon_info_size()) );
+	// This is most likely an empty bank or something went wrong
+	// This is most likely an empty bank or something went wrong
```
Goober5000, line 19540: You duplicated a line here.

TRBlount:
I agree. I may end up modifying the equivalent check in `ship_fire_secondary` as well since it is identical.

PR #2668, opened 2020-08-18
Overhaul forward and matrix_interpolate

Baezon:
The old functions were poorly named (this isn't really interpolation), a mess of repeated code, a tangled forest of ifs and elses, and numerous small errors. The new versions accomplish the same goals while keeping things compartmentalized, simpler, and of course, well-commented. 

Also adds the "No directional bias when turning" flag, the explanation is a bit complicated, but basically turning at maximum on all axes to reach your target angular position will not necessarily result in an initial path directly towards your target. This doesn't matter for stationary targets, but for anything moving that initial phase where you aren't going straight towards your target can end up harming your time-on-target, and create strange-looking paths. The flag will ensure the path is always straight. I suspect this may be the cause of missile's poorer accuracy on off-bore targets, but I haven't tested specifically for it.


Baezon:
After more statistical and numerical analysis than I'd like to admit, I'd like to present my finding is to the board. After some adjustments to improve the results, with randomized starts, goals, and max velocities and accelerations, the vector functions agree very well. The average percentage difference in their times-to-arrival is 0.120% and the standard deviation is 0.64%. The matrix functions do not agree quite as well with an average of 1.037% and a standard deviation of 1.466%.

A significant factor there is that the retail matrix function is allowed to 'cheat' and zeros out its velocity when it's very close (while the new version must bring its velocity down very low in addition to its speed). While removing this restriction for the new version can bring the average down to 0.181% and the std dev to 0.85% (which I would prefer not to do), I think its important to note that these discrepancies are at the very *end* of their paths and so I think less harmful in general.

A full log of the results is attached. I'll leave to you whether you think this is an acceptable margin.
[vector.txt](https://github.com/scp-fs2open/fs2open.github.com/files/5114853/vector.txt)

PR #2673, opened 2020-08-21
Fix AI banking when in formation

Baezon:
The AI code makes not one, not two, but three unnecessary and damaging checks for formation and attempting to handle bank. I've commented each location for the problems for each one causes, and how often it wouldn't even do what they were attempting. 

The desired behavior is obvious (as it is mentioned in the code multiple times), while a non-wing leader is flying in formation, it should attempt to match bank with the leader and also move to their assigned relative position. The retail code has the ship turn towards its assigned position, and then later in the function turn to match the leader's orientation. This is wrong in several ways that I've commented, but `ai_turn_towards_vector` comes with an optional `rvec` argument for exactly this reason!! I've added a function that will calculate the matching rvec to their wing leader and will pass that along to `turn_towards_vector` so it can accomplish all its goals in one call.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -11768,7 +11768,7 @@ int ai_formation()
 	}
 
 	if (dist_to_goal > 500.0f) {
-		turn_towards_point(Pl_objp, &goal_point, NULL, 0.0f, leader_orient);
+		turn_towards_point(Pl_objp, &goal_point, nullptr, 0.0f, leader_orient);
 		if (ab_allowed && !(Pl_objp->phys_info.flags & PF_AFTERBURNER_ON) && (dot_to_goal > 0.2f)) {
 			float percent_left = 100.0f * shipp->afterburner_fuel / sip->afterburner_fuel_capacity;
 			if (percent_left > 30.0f) {
@@ -11779,7 +11779,7 @@ int ai_formation()
 		accelerate_ship(aip, 1.0f);
 	} else if (dist_to_goal > 200.0f) {
 		if (dot_to_goal > -0.5f) {
-			turn_towards_point(Pl_objp, &goal_point, NULL, 0.0f, leader_orient);
+			turn_towards_point(Pl_objp, &goal_point, nullptr, 0.0f, leader_orient);
 			float range_speed = shipp->current_max_speed - leader_speed;
 			if (range_speed > 0.0f) {
 				set_accel_for_target_speed(Pl_objp, leader_speed + range_speed * (dist_to_goal+100.0f)/500.0f);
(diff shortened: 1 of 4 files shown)


code/ai/aicode.cpp
```diff
@@ -11942,22 +11951,6 @@ int ai_formation()
...
-
-		vm_vec_copy_scale(&angular_accel, &Pl_objp->phys_info.max_rotvel, 0.2f);
-		vm_matrix_interpolate(&leader_objp->orient, &Pl_objp->orient, &Pl_objp->phys_info.rotvel, flFrametime, &new_orient, &w_out, &Pl_objp->phys_info.max_rotvel, &angular_accel, 1);
-
-		Pl_objp->orient = new_orient;
-		Pl_objp->phys_info.rotvel = w_out;
-	} else {
-		Pl_objp->phys_info.rotvel.xyz.z = 0.0f;
-	}
-
```
Baezon, line 11960: This call to matrix_interpolate will move the ship! After it has already been moved earlier in the function by turn_towards_point, if you're keeping track thats _three orientation changes happened per frame, when it should only be one. Also bizarrely just asserts its bank is 0, if it happens to agree with its leader's uvec.

code/ai/aicode.cpp
```diff
@@ -14219,15 +14212,7 @@ void ai_process( object * obj, int ai_index, float frametime )
...
 		ai_control_info_check(aip, sip, &obj->phys_info);
-		vec3d copy_desired_rotvel = obj->phys_info.rotvel;
 		physics_read_flying_controls( &obj->orient, &obj->phys_info, &AI_ci, frametime);
-		// if obj is in formation and not flight leader, don't update rotvel
-		if (aip->ai_flags[AI::AI_Flags::Formation_object, AI::AI_Flags::Formation_wing]) {
-			if (&Objects[aip->goal_objnum] != obj) {
-				obj->phys_info.desired_rotvel = copy_desired_rotvel;
-				obj->phys_info.rotvel = copy_desired_rotvel;
-			}
-		}
```
Baezon, line 14230: Why, at the very end it feels the need to copy its rotvel, even though `read_flying_controls` does nothing to rotvel, or why it _also_ copies that rotvel into desired_rotvel even though the rest of the AI code doesn't use desired_rotvel and now allows it to turn much faster than normal, I couldn't tell you. This check doesn't even catch exclude all wingleaders because often AI wingleaders won't even have their goal_objnum set. And probably most importantly completely overrides the maneuver overrides!

code/ai/aicode.cpp
```diff
@@ -1507,24 +1507,24 @@ void get_tangent_point(vec3d *tan1, vec3d *p0, vec3d *centerp, vec3d *p1, float
...
 {
 	ai_info	*aip;
 	aip = &Ai_info[Ships[Pl_objp->instance].ai_index];
-	
-	// check if in formation and if not leader, don't change rotvel.z (bank to match leader elsewhere)
-	if (aip->ai_flags[AI::AI_Flags::Formation_object, AI::AI_Flags::Formation_wing]) {
-		if (&Objects[aip->goal_objnum] != objp) {
-			float rotvel_z = objp->phys_info.rotvel.xyz.z;
-			ai_turn_towards_vector(point, objp, Ship_info[Ships[objp->instance].ship_info_index].srotation_time, slide_vec, nullptr, bank_override, 0);
-			objp->phys_info.rotvel.xyz.z = rotvel_z;
```
Baezon, line 1522: The least damaging of the unnecessary checks, but is now not necessary since this function can handle banking just fine.

PR #2674, opened 2020-08-21
refactor display_name for ships, ship classes, and weapon classes

Goober5000:
Several things going on here:
1) refactor `get_display_string()` to `get_display_name()` to be more precise
2) likewise, refactor `alt_name` to `display_name`
3) automatically generate display names when parsing the ship, ship class, and weapon class for the first time
4) remove the on-the-fly truncation of names with hash symbols

Basically this subsumes the `end_string_at_first_hash_symbol` logic into the display name.  Rather than truncate the name every time it's displayed, truncate it once when it's parsed.  This takes advantage of the previous display name infrastructure, since display names are a superset of the case of truncating the hash symbol.

This also has the benefit of now allowing hash symbols within display names.  When updating Destiny of Peace, I noticed that "Lockheed Shuttle #4" would no longer display properly; and it was not previously possible to fix this even with display names.

I could pull out the `get_display_string()` and `alt_name` renaming into a separate PR if desired.


code/scripting/api/objs/ship.cpp
```diff
@@ -1717,13 +1717,13 @@ ADE_FUNC(getAltClassName, l_Ship, NULL, "Gets the alternate class name of the sh
 	if (shipp->alt_type_index < 0)
 		return ade_set_args(L, "s", "");
 
-	char temp[NAME_LENGTH];
+	char temp_altname[NAME_LENGTH];
 
-	*temp = 0;
-	mission_parse_lookup_alt_index(shipp->alt_type_index, temp);
+	*temp_altname = 0;
+	strcpy(temp_altname, mission_parse_lookup_alt_index(shipp->alt_type_index));
```
asarium, line 1723:
```suggestion
	strcpy_s(temp_altname, mission_parse_lookup_alt_index(shipp->alt_type_index));
```
No need to initialize the char array since `strcpy` will do that anyway.

code/scripting/api/objs/ship.cpp
```diff
@@ -1693,7 +1693,7 @@ ADE_FUNC(getCallsign, l_Ship, NULL, "Gets the callsign of the ship in the curren
 	char temp_callsign[NAME_LENGTH];
 
 	*temp_callsign = 0;
-	mission_parse_lookup_callsign_index(shipp->callsign_index, temp_callsign);
+	strcpy(temp_callsign, mission_parse_lookup_callsign_index(shipp->callsign_index));
```
asarium, line 1696:
```suggestion
	strcpy_s(temp_callsign, mission_parse_lookup_callsign_index(shipp->callsign_index));
```

code/ship/ship.cpp
```diff
@@ -1902,6 +1916,12 @@ static void parse_ship(const char *filename, bool replace)
 		first_time = true;
 
 		strcpy_s(sip->name, fname);
+
+		// if this name has a hash, create a default display name
+		if (get_pointer_to_first_hash_symbol(sip->name)) {
```
asarium, line 1921: If I read this right, this could overwrite a previous display name for a `+nocreate` ship entry. Maybe only do this for `create_if_not_found`?

code/weapon/weapons.cpp
```diff
@@ -872,10 +872,24 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		first_time = true;
 
 		strcpy_s(wip->name, fname);
+
+		// if this name has a hash, create a default display name
+		if (get_pointer_to_first_hash_symbol(wip->name)) {
```
asarium, line 877: Same as for ships probably applies here as well.

asarium (changes requested):

code/scripting/api/objs/ship.cpp
Goober5000: Good point

code/ship/ship.cpp
Goober5000: Oh, true.  I'll fix that

Goober5000: This should not overwrite any previous display name.  Look at the enclosing `if()` block; this is only run for brand new entries.

code/weapon/weapons.cpp
Goober5000: See above.

code/ship/ship.cpp
asarium: Oh, yes. I misread that.

asarium (approved):

PR #2680, opened 2020-08-25
Fix usage of float instead of int epsilon

Baezon:
int epsilon is 0, not a very useful threshold.


Goober5000:
This is a follow-up to PR #2640.

PR #2686, opened 2020-08-27
Add "$respect ship axial turnrate differences:" flag (and fix a bug with 2673)

Baezon:
Since the overwhelming majority of calls to `ai_turn_towards_vector` use the ship's `srotation_time`, the function is refactored to not need a turn_time passed in at all, and it will presume srotation is being used. The exceptions to this now use a 'turnrate_mod` to change the default. For instance, formation flying calls to the function pass in (1,1,0.2) to have slower than normal banking when turning for that purpose, which restores the similar retail behavior I overlooked in PR #2673 (EDIT: I have since changed formation flying to use a flag instead of this system since it only changed acceleration, but this is the idea).

This allows the default presumption to be changed to the ship's tabled values for each axis, instead of srotation, with the new flag. I will note this in the documentation as well, but the AI wil _NOT_ take this into account when turning. They will always turn in a direct (ish) path to their target, even if they could bank and align a faster axis towards their goal in the new system. Accounting for this is theoretically possible, but would vastly complicate the current implementation.

### DIFF
--- a/code/ai/ai.h
+++ b/code/ai/ai.h
@@ -566,7 +566,7 @@ extern void ai_do_default_behavior(object *obj);
 extern void ai_start_waypoints(object *objp, waypoint_list *wp_list, int wp_flags);
 extern void ai_ship_hit(object *objp_ship, object *hit_objp, vec3d *hit_normal);
 extern void ai_ship_destroy(int shipnum);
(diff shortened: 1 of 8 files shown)


Goober5000:
I guess it looks good to me to, but I'll repeat the same question as before: does this change retail behavior?

Baezon:
The flag obviously will, but without it, no. The only real changes otherwise are, in the most cases, where srotation gets involved. Before everyone was sending it in as an argument, but now turn_towards_vector presumes it by default, and any multipliers they might have done to srotation are sent in if necessary. The SEXP call to the function needs a bit more set up in the new system, but I got working just as it was previously as well.

Goober5000:
Sounds good.

PR #2687, opened 2020-08-27
various collision group fixes

Goober5000:
Collision groups as previously coded were unusable due to Lua 5.1's lack of support for bitwise operators.  This fixes that deficiency in two ways:
1) Objects now have the functions `addToCollisionGroup` and `removeFromCollisionGroup`
2) An `evaluateNumericSEXP` function has been added for returning SEXP results as numbers (so that scripts can use the bitwise SEXP operators)

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -25590,7 +25590,7 @@ int get_sexp_main()
 	return start_node;
 }
 
-int run_sexp(const char* sexpression)
+int run_sexp(const char* sexpression, bool run_eval_num, bool *is_nan_or_nan_forever)
 {
 	char* oldMp = Mp;
 	int n, i, sexp_val = UNINITIALIZED;
@@ -25608,9 +25608,19 @@ int run_sexp(const char* sexpression)
 	n = get_sexp_main();
 	if (n != -1)
 	{
-		sexp_val = eval_sexp(n);
+		if (run_eval_num)
+		{
+			bool is_nan, is_nan_forever;
+			sexp_val = eval_num(n, is_nan, is_nan_forever);
+			if (is_nan_or_nan_forever != nullptr)
+				*is_nan_or_nan_forever = (is_nan || is_nan_forever);
+		}
+		else
+			sexp_val = eval_sexp(n);
+
 		free_sexp2(n);
 	}
+
 	Mp = oldMp;
 
 	return sexp_val;
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -83,7 +83,7 @@ ADE_FUNC(getObjectFromSignature, l_Mission, "number Signature", "Gets a handle o
 	return ade_set_object_with_breed(L, objnum);
 }
 
-ADE_FUNC(evaluateSEXP, l_Mission, "string", "Runs the defined SEXP script", "boolean", "if the operation was successful")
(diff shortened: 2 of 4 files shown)


code/scripting/api/libs/mission.cpp
```diff
@@ -99,7 +99,35 @@ ADE_FUNC(evaluateSEXP, l_Mission, "string", "Runs the defined SEXP script", "boo
 		return ADE_RETURN_FALSE;
 }
 
-ADE_FUNC(runSEXP, l_Mission, "string", "Runs the defined SEXP script", "boolean", "if the operation was successful")
+ADE_FUNC(evalSEXP, l_Mission, "string", "Runs the defined SEXP script, and returns the result as a boolean", "boolean", "true if the SEXP returned SEXP_TRUE; false if the SEXP returned anything else (even a number)")
```
asarium, line 102: This seems to be an exact copy of the function above. Is that intentional?

code/scripting/api/libs/mission.cpp
```diff
@@ -99,7 +99,35 @@ ADE_FUNC(evaluateSEXP, l_Mission, "string", "Runs the defined SEXP script", "boo
...
+
+ADE_FUNC(evaluateNumericSEXP, l_Mission, "string", "Runs the defined SEXP script, and returns the result as a number", "number", "the value of the SEXP result")
+{
+	const char* s;
+	int r_val;
+
+	if (!ade_get_args(L, "s", &s))
+		return ade_set_args(L, "i", 0);
+
+	r_val = run_sexp(s);
```
asarium, line 126: How would this handle the special NaN return values of the SEXP system? It may be worthwhile to consider those cases and return the float NaN in that case. I think I wrote code for that in the LuaSEXP system actually...

asarium (changes requested):

code/scripting/api/libs/mission.cpp
Goober5000: Oh, that's a good point.  I'll add that.

code/scripting/api/libs/mission.cpp
Goober5000: Yes.  I realized that in most cases the `( when ( true ) [sexp string] )` format isn't necessary, so `evaluateSEXP` is perfectly fine.  But I also wanted a shorter version of the name.

asarium: I think it is not a good idea to add the same function twice just with a different name (unless there is deprecation involved). However, if you dislike the name you can also just rename it locally:
```lua
local evalSexp = mn.evaluateSEXP

evalSexp("( when ( true ) ( do-nothing ) )")
```

Goober5000: Meh, ok

code/scripting/api/libs/mission.cpp
Goober5000: Ok, see what you think.

code/parse/sexp.cpp
```diff
@@ -25590,7 +25590,7 @@ int get_sexp_main()
 	return start_node;
 }
 
-int run_sexp(const char* sexpression)
+int run_sexp(const char* sexpression, bool run_eval_num, bool *check_for_nan)
```
asarium, line 25593: `check_for_nan` is a bit misleading here since at least for me that implies that it changes the behavior of the function. Maybe:
```suggestion
int run_sexp(const char* sexpression, bool run_eval_num, bool *is_nan_out)
```

Goober5000: Mmm, I changed it to `is_nan_or_nan_forever`.

asarium (approved):

code/parse/sexp.cpp
Goober5000: And changed it for real this time.

PR #2691, opened 2020-08-27
SLC2, BSP alignment and POF version 2118 support

Shivansps:
Changed SLDC system to SLC2, this replaced the char at the beginning of the node for an int in order for the tree to be aligned.
Added BSP_Defpoints data alignment and SLDC to SLC2 conversion for pof version < 2118.

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -1717,25 +1717,20 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
 				{
 					pm->submodel[n].bsp_data_size = cfread_int(fp);
 					if (pm->submodel[n].bsp_data_size > 0) {
-						mprintf(("BSP_Data is being aligned.\n"));
-						ubyte *bsp_in, *bsp_out;
+						//mprintf(("BSP_Data is being aligned.\n"));
 
-						bsp_in = (ubyte*)vm_malloc(pm->submodel[n].bsp_data_size);
-						bsp_out = (ubyte*)vm_malloc(pm->submodel[n].bsp_data_size * 2);
+						std::unique_ptr<ubyte[]> bsp_in (new ubyte[pm->submodel[n].bsp_data_size]);
+						std::unique_ptr<ubyte[]> bsp_out (new ubyte[pm->submodel[n].bsp_data_size*2]);
 
-						cfread(bsp_in, 1, pm->submodel[n].bsp_data_size, fp);
+						cfread(bsp_in.get(), 1, pm->submodel[n].bsp_data_size, fp);
 
 						//mprintf(("BSP_Data was %d bytes in size\n", pm->submodel[n].bsp_data_size));
-						pm->submodel[n].bsp_data_size = align_bsp_data(bsp_in, bsp_out, pm->submodel[n].bsp_data_size);
+						pm->submodel[n].bsp_data_size = align_bsp_data(bsp_in.get(), bsp_out.get(), pm->submodel[n].bsp_data_size);
 						//mprintf(("BSP_Data now is %d bytes in size\n", pm->submodel[n].bsp_data_size));
 
 						pm->submodel[n].bsp_data = (ubyte*)vm_malloc(pm->submodel[n].bsp_data_size);
-						memcpy(pm->submodel[n].bsp_data, bsp_out, pm->submodel[n].bsp_data_size);
-
+						memcpy(pm->submodel[n].bsp_data, bsp_out.get(), pm->submodel[n].bsp_data_size);
 						swap_bsp_data(pm, pm->submodel[n].bsp_data);
-
-						vm_free(bsp_in);
-						vm_free(bsp_out);
 					}
 					else {
 						pm->submodel[n].bsp_data = nullptr;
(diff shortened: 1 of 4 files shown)


Shivansps:
Re-uploaded #2200

JohnAFernandez:
So that maybe we can finally get this done, it looks like travis just doesn't like this: ```
`/home/travis/build/scp-fs2open/fs2open.github.com/code/model/modelsinc.h:20:2: error: This should only be used internally by the model library.  See John if you think you need to include this elsewhere. [clang-diagnostic-error]
#error This should only be used internally by the model library.  See John if you think you need to include this elsewhere.`
```

Meaning that we probably just need to enable that to be used elsewhere, if I'm reading that correctly.

**Comment by Shivansps** [comment]
Yeah the problem with that is that error makes no sence, that should only happen if, lets say, i edit dds.cpp and i include modelsync.h, etc

I added no includes, and all the code i added is inside the "modellib", the problem to me is modelsync.h its either being included before the modellib definition for some weird reason, or it is wrongly included somewhere and any change to it triggers the error.

Ill look into it a little more today to see if i can discover anything.

**Comment by asarium** [comment]
The issue here is that clang-tidy tries to only parse that single header so the checked macro is not defined. You can remove that `#error` to fix the issue.

**Comment by Shivansps** [comment]
Goober told me not to comment the error.

**Comment by Goober5000** [comment]
If the error only occurs for this specific PR because of clang, I think we can ignore it and merge the PR anyway.  It will not affect subsequent builds.

This means we can just go ahead and do a proper code review.

File: code/model/modelread.cpp [hunk 1]
```diff
@@ -1700,13 +1700,46 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
...
+					}
+					else {
+						pm->submodel[n].bsp_data = nullptr;
+					}
+				}
+				else
+				{
+					pm->submodel[n].bsp_data_size = cfread_int(fp);
+					if (pm->submodel[n].bsp_data_size > 0) {
+						mprintf(("BSP_Data is being aligned.\n"));
```
asarium, line 1720: This will log print will only spam the log file and is not really useful for mod debugging.

code/model/modelread.cpp
```diff
@@ -1738,14 +1771,37 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
 			}
 
 			case ID_SLDC: // kazan - Shield Collision tree
-				{
+				{   //ShivanSpS - if pof version is 2118 or higher ignore SLDC, otherwise convert it to slc2.
+					if (pm->version < 2118) {
+						mprintf(("SLDC data is being converted to SLC2.\n"));
```
asarium, line 1776: Same here, please remove or convert to `nprintf`.

code/model/modelread.cpp
```diff
@@ -1738,14 +1771,37 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
...
+						pm->sldc_size = cfread_int(fp);
+						ubyte* sldc_tree, * slc2_tree;
+						sldc_tree = (ubyte*)vm_malloc(pm->sldc_size);
+						slc2_tree = (ubyte*)vm_malloc(pm->sldc_size * 2);
```
asarium, line 1780: Please use `std::unique_ptr<ubyte[]>` for automatic deallocation.

code/model/modelread.cpp
```diff
@@ -1738,14 +1771,37 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
...
+						//mprintf(("SLDC Shield Collision Tree was %d bytes in size\n", pm->sldc_size));
+						pm->sldc_size = convert_sldc_to_slc2(sldc_tree, slc2_tree, pm->sldc_size);
+						//mprintf(("SLC2 Shield Collision Tree is %d bytes in size\n", pm->sldc_size));
+						pm->shield_collision_tree = (ubyte*)vm_malloc(pm->sldc_size); //sldc_size is slc2 size, reused variable
+						memcpy(pm->shield_collision_tree, slc2_tree, pm->sldc_size);
+						swap_sldc_data(pm->shield_collision_tree);
+						vm_free(sldc_tree);
+						vm_free(slc2_tree);
+					}
+				}
```
asarium, line 1791: Maybe throw a warning in the else case since the presence of this chunk is not expected or valid in that case.

code/model/modelread.cpp
```diff
@@ -5431,14 +5487,16 @@ void swap_bsp_data( polymodel * pm, void * model_ptr )
 
 void swap_sldc_data(ubyte * buffer)
 {
+//ShivanSpS - Changed type char for a type int for SLC2
 #if BYTE_ORDER == BIG_ENDIAN
```
asarium, line 5491: Wait, we still support Big endian? Do we actually need that for ARM? I thought that could also handle little endian. If that is the case, maybe just remove this.

code/model/modelsinc.h
```diff
@@ -12,6 +12,9 @@
 
 
 class polymodel;
+struct vec3d;
+struct matrix;
+typedef std::uint8_t ubyte;
```
asarium, line 17: We already have this typedef in `pstypes.h` to please only use that instead of repeating it here.

asarium (changes requested):
How extensively has this been testing with existing model data? This has the potential of breaking quite a lot of stuff so maybe create a test build and have some people do some smoke testing on existing mods to check if this works as intended.

asarium:
@Goober5000 Simply ignoring static code analysis is not a good way of using our CI results.
If the header cannot be compiled on its own, then it is, in my opinion, not a good header and the `#error` should be removed.

Shivansps:
As for testing, i did that a year when i wrote this, i created two missions, one that loaded every ship on FS2 with MediaVPS, and another one to test the shield collision was still working properly, i recreated them because i lost them.
Ive uploaded them along with a test build https://www.dropbox.com/s/0utr2rsgscsb916/RP%202691%20Test.rar?dl=0

code/model/modelread.cpp
Shivansps: Ok ill do that

code/model/modelread.cpp
Shivansps: ok

code/model/modelread.cpp
Shivansps: Well, the function of swaping SLDC data is already in the code, so im forced to adjust it to the new node format, otherwise the game will crash on a big endian system.

code/model/modelsinc.h
Shivansps: This is also part of the problem with this specific header, whiout that you are going to get errors about missing these types in CI. For the record these types were already in use by the header, as i only added a definition.

code/model/modelread.cpp
asarium: Yes but the point is if the big endian case is even still necessary since I think all of our supported systems are little endian. That is not a question to answer here though so there is nothing to change here. It was just something I noticed.

code/model/modelread.cpp
Shivansps: The SLDC chunk needs to still be present on pof version 2118 or higher in order to provide backguard compatibility with older versions of FSO. It would just create a lot of unnecessary warnings.

code/model/modelread.cpp
Shivansps: Ok im new to smart pointers, but you can do that when "cfread(sldc_tree, 1, pm->sldc_size, fp);" expects a void* on sldc_tree?

Goober5000:
> @Goober5000 Simply ignoring static code analysis is not a good way of using our CI results.

This isn't ignoring it.  It's carefully considering it and knowing when to follow it (which is most of the time) and when it is giving a false positive.

What if we changed it to `#pragma message`?

code/model/modelread.cpp
asarium: Ah, didn't know this was how that was supposed to work but that makes sense.

code/model/modelread.cpp
asarium: You can still get the "raw" pointer via `.get()` if necessary.

asarium:
Yes, I think a pragma would be the better solution here.

code/model/modelsinc.h
Shivansps: Using pragma message this header now works, but it will fail again if i remove those forward declarations.

JohnAFernandez:
Hey this got merged by accident when we merged the compressor!

PR #2692, opened 2020-08-28
Allow for arbitrary wingmate relative positions

Baezon:
Adds the "$Custom wingmate relative positions:" field to species_defs.tbl. Pretty self-explanatory, modders can now specify whatever positions they want for wingmates to occupy while formation flying. I'll be sure to note in the documentation that these are unscaled by size, and provide the retail positions for reference.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -11429,10 +11429,11 @@ void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wing_inde
 	vec3d	wing_delta, rotated_wing_delta;
 	float		wing_spread_size;
 	species_info* species = &Species_info[Ship_info[Ships[leader_objp->instance].ship_info_index].species];
-	bool is_custom_pos = species->custom_wing_positions > wing_index;
+	int formation_index = Wings[Ships[leader_objp->instance].wingnum].formation;
 
-	if (wing_index != 0 && is_custom_pos) {
-		wing_delta = species->wing_positions[wing_index - 1]; //  custom desired location in leader's reference frame
+
+	if (wing_index != 0 && formation_index != -1) {
+		wing_delta = species->formations[formation_index][wing_index - 1]; //  custom desired location in leader's reference frame
 	}
 	else {
 		get_wing_delta(&wing_delta, wing_index);		//	default desired location in leader's reference frame
@@ -11442,12 +11443,13 @@ void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wing_inde
 
 	// apply debug modifications to spread size and also y scale (2x default) unless it's a custom position
 	if (leader_objp->flags[Object::Object_Flags::Player_ship]) {
-		if(!is_custom_pos)
+		if(formation_index == -1)
 			wing_delta.xyz.y *= Wing_y_scale;
 		wing_spread_size *= Wing_scale;
 	}
 
 	vm_vec_scale(&wing_delta, wing_spread_size * (1.0f + leader_objp->phys_info.speed/70.0f));
+
 
 	vm_vec_unrotate(&rotated_wing_delta, &wing_delta, &leader_objp->orient);	//	Rotate into leader's reference.
 
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -11428,14 +11428,22 @@ void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wing_inde
 {
 	vec3d	wing_delta, rotated_wing_delta;
 	float		wing_spread_size;
+	species_info* species = &Species_info[Ship_info[Ships[leader_objp->instance].ship_info_index].species];
+	bool is_custom_pos = species->custom_wing_positions > wing_index;
 
-	get_wing_delta(&wing_delta, wing_index);		//	Desired location in leader's reference frame
+	if (wing_index != 0 && is_custom_pos) {
+		wing_delta = species->wing_positions[wing_index - 1]; //  custom desired location in leader's reference frame
+	}
+	else {
+		get_wing_delta(&wing_delta, wing_index);		//	default desired location in leader's reference frame
+	}
 
 	wing_spread_size = MAX(50.0f, 3.0f * get_wing_largest_radius(leader_objp, formation_object_flag) + 15.0f);
 
-	// for player obj (1) move ships up 20% (2) scale formation up 20%
+	// apply debug modifications to spread size and also y scale (2x default) unless it's a custom position
 	if (leader_objp->flags[Object::Object_Flags::Player_ship]) {
-		wing_delta.xyz.y *= Wing_y_scale;
+		if(!is_custom_pos)
+			wing_delta.xyz.y *= Wing_y_scale;
 		wing_spread_size *= Wing_scale;
 	}
 
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -11431,14 +11431,13 @@ void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wing_inde
 	species_info* species = &Species_info[Ship_info[Ships[leader_objp->instance].ship_info_index].species];
 	int formation_index = Wings[Ships[leader_objp->instance].wingnum].formation;
 
-
-	if (wing_index != 0 && formation_index != -1) {
+	if (wing_index == 0)
+		wing_delta = vmd_zero_vector;
+	else if ( formation_index != -1) 
 		wing_delta = species->formations[formation_index][wing_index - 1]; //  custom desired location in leader's reference frame
-	}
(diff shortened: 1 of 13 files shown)


code/species_defs/species_defs.cpp
```diff
@@ -328,6 +328,30 @@ void parse_species_tbl(const char *filename)
...
+
+				while (check_for_string("(")) {
+
+					if (wing_index >= MAX_SHIPS_PER_WING)
+					{
+						Warning(LOCATION, "Species %s : Custom wing positions cannot exceed %d.  Ignoring the rest...", species_name, MAX_SHIPS_PER_WING);
+						break;
+					}
+
+					if (stuff_float_list(species->wing_positions[wing_index-1].a1d, 3) != 3) {
```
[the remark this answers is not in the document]
Baezon: Because I didn't look hard enough <.< >.>

code/ai/aicode.cpp
```diff
@@ -11428,19 +11428,29 @@ void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wing_inde
...
+	int formation_index = Wings[Ships[leader_objp->instance].wingnum].formation;
 
-	get_wing_delta(&wing_delta, wing_index);		//	Desired location in leader's reference frame
+
+	if (wing_index != 0 && formation_index != -1) {
+		wing_delta = species->formations[formation_index][wing_index - 1]; //  custom desired location in leader's reference frame
+	}
+	else {
+		get_wing_delta(&wing_delta, wing_index);		//	default desired location in leader's reference frame
+	}
```
Goober5000, line 11440: May I suggest this:
```
if (wing_index == 0)
  wing_delta = vmd_zero_vector;
else if (formation_index != -1)
  [...]
else
  get_wing_delta(...);
```

code/mission/missionparse.cpp
```diff
@@ -4574,6 +4574,14 @@ void parse_wing(mission *pm)
 	// wing could be created if it needs to be
 	wingp->wave_delay_timestamp = timestamp(0);
 
+	// Use a custom formation if specified
+	if (optional_string("+Formation Index:")) {
+		stuff_int(&(wingp->formation));
+	}
+	else { // else use the retail formation
+		wingp->formation = -1;
```
Goober5000, line 4582: This should default to the species default formation (which itself defaults to -1).

code/species_defs/species_defs.cpp
```diff
@@ -328,6 +328,31 @@ void parse_species_tbl(const char *filename)
...
 				stuff_string(species->cmeasure_name, F_NAME, NAME_LENGTH);
 
+			if (optional_string("$Custom wingmate formations:")) {
+
+				while (check_for_string("(")) {
+					species->formations.emplace_back();
+					if (stuff_vec3d_list(species->formations.back().data(), MAX_SHIPS_PER_WING - 1) != MAX_SHIPS_PER_WING - 1) {
+						Warning(LOCATION, "A custom formation for species %s did not have 5 positions. Ignoring.", species->species_name);
+						species->formations.pop_back();
+					}
```
Goober5000, line 338: I wouldn't write directly to `.data()` here.  Instead, make a temporary `SCP_vector<vec3d> vec_list`, stuff that, then push_back the temp list into the formation list or lists.  See how `parse_waypoint_list` does it.

code/species_defs/species_defs.cpp
```diff
@@ -328,6 +328,31 @@ void parse_species_tbl(const char *filename)
...
+				else {
+					stuff_int(&species->default_formation);
+
+					if (species->default_formation >= species->formations.size() || species->default_formation <= -2) {
+						Warning(LOCATION, "Invalid default formation index for species %s. Ignoring.", species->species_name);
+						species->default_formation = -1;
+					}
+				}
+			}
+
```
Goober5000, line 355: There needs to be an `else if (!no_create)` to set `default_formation` to -1.  Alternatively (and this is the better method) set it in the `species_info()` default constructor.

code/mission/missionparse.cpp
Goober5000: Also, add a corresponding block in `missionsave.cpp` (for both FRED and qtFRED) to save the formation index.  Look at how `wing_squad_filename` is saved for an example.

code/species_defs/species_defs.cpp
Baezon: Are there any particular advantages to doing it that way?

Goober5000: It's more of a stylistic or code-safety practice.  The `.data()` method returns a pointer to the raw underlying data of the vector, and one should avoid working with the internal structures of other classes in most cases.

My first reaction was that the compiler shouldn't even have allowed this, but upon further examination in *this particular case* it looks safe.  However it's one of those "code smells".

Baezon: Fair enough.

code/species_defs/species_defs.cpp
```diff
@@ -328,6 +328,34 @@ void parse_species_tbl(const char *filename)
...
+			if (optional_string("$Custom wingmate formations:")) {
+
+				while (check_for_string("(")) {
+					SCP_vector<vec3d> current_formation;
+
+					if (stuff_vec3d_list(current_formation) != MAX_SHIPS_PER_WING - 1) {
+						Warning(LOCATION, "A custom formation for species %s did not have 5 positions. Ignoring.", species->species_name);
+					} else {
+						species->formations.emplace_back();
+						std::copy_n(current_formation.begin(), MAX_SHIPS_PER_WING-1, species->formations.back().begin());
```
Goober5000, line 340: You can use `push_back(std::move(current_formation))` here.  This will avoid the copying, as it simply makes the new vector point to the old vector's data.

code/mission/missionparse.cpp
```diff
@@ -4574,6 +4574,19 @@ void parse_wing(mission *pm)
 	// wing could be created if it needs to be
 	wingp->wave_delay_timestamp = timestamp(0);
 
+	// Use a custom formation if specified
+	species_info species = Species_info[Ship_info[Ships[wingp->ship_index[0]].ship_info_index].species];
+	if (optional_string("+Formation Index:")) {
+		stuff_int(&(wingp->formation));
+		if (wingp->formation <= -2 || wingp->formation >= species.formations.size()) {
```
Goober5000, line 4581: You need to cast `.size()` to int here as well.

code/species_defs/species_defs.cpp
Baezon: Well, it isn't actually a vector of vectors, but a vector of arrays so I don't think you can do this without a copy.

Goober5000: Oh, you are correct. :thonk:

By the way, that 5 in the `Warning` should also be the SHIPS-1 number.  Use SIZE_T_ARG, e.g.
```
error_display(0, "String too long.  Length = " SIZE_T_ARG ".  Max is %i.\n", len, max);
```

code/species_defs/species_defs.cpp
```diff
@@ -328,6 +328,34 @@ void parse_species_tbl(const char *filename)
 			if (optional_string("$Countermeasure type:"))
 				stuff_string(species->cmeasure_name, F_NAME, NAME_LENGTH);
 
+			if (optional_string("$Custom wingmate formations:")) {
+
+				while (check_for_string("(")) {
+					SCP_vector<vec3d> current_formation;
+
+					if (stuff_vec3d_list(current_formation) != MAX_SHIPS_PER_WING - 1) {
+						Warning(LOCATION, "A custom formation for species %s did not have %d positions. Ignoring.", species->species_name, MAX_SHIPS_PER_WING-1);
```
Goober5000, line 337:
```suggestion
					if (stuff_vec3d_list(current_formation) != MAX_SHIPS_PER_WING - 1) {
						Warning(LOCATION, "A custom formation for species %s did not have " SIZE_T_ARG " positions. Ignoring.", species->species_name, MAX_SHIPS_PER_WING-1);
```

Baezon: `MAX_SHIPS_PER_WING-1` here isn't a `size_t`, it is not specified (it is just `6-1`) so I think it gets type `int`. If you say it has type `size_t` then bad things happen unless `MAX_SHIPS_PER_WING-1` is also cast to type `size_t` (or `MAX_SHIPS_PER_WING` is defined to have a `size_t` cast in it). Is that what you mean?

Goober5000: Hmm.  I was thinking "future proofing" in case we ever change `MAX_SHIPS_PER_WING` to `const size_t`, but maybe that's too clever by half.  You can either cast the argument to `size_t` or just leave the whole thing as-is.

Goober5000:
Looks like the next step is for me to add FRED support.  I'll try to get to that tomorrow.

PR #2700, opened 2020-08-30
add some vector scripting functions

Goober5000:
createRandomVector
perturb (for vectors)
perturb (for matrixes)
randomInCircle
randomInSphere

### DIFF
--- a/code/scripting/api/objs/vecmath.cpp
+++ b/code/scripting/api/objs/vecmath.cpp
@@ -623,6 +623,90 @@ ADE_FUNC(findNearestPointOnLine,
 	return ade_set_args(L, "of", l_Vector.Set(dest), f);
 }
 
+ADE_FUNC(perturb,
+	l_Vector,
+	"number angle1, [number angle2]",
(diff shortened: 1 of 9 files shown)


code/scripting/api/objs/vecmath.cpp
```diff
@@ -623,6 +623,90 @@ ADE_FUNC(findNearestPointOnLine,
 	return ade_set_args(L, "of", l_Vector.Set(dest), f);
 }
 
+ADE_FUNC(perturb,
+	l_Vector,
+	"number angle1, [number angle2]",
+	"Create a new normalized vector, randomly perturbed around a given (normalized) vector.  Calls `vm_vec_random_cone`; angles are in radians.  If only one angle is specified, it is the max angle.  If both are specified, the first is the minimum and the second is the maximum.",
```
asarium, line 629: The reference to `vm_vec_random_cone` here is not really useful for the scripting documentation since it will mean nothing to a script writer and a coder can see exactly what is called from the actual code below.

Same for the functions below.

asarium (changes requested):

code/scripting/api/objs/vecmath.cpp
Goober5000: Ok, modified

asarium (approved):

PR #2701, opened 2020-08-31
Fix Two Multi Crash Points

JohnAFernandez:
This fixes two points at which multi can crash. One is during respawn. The net_signature needs to be set to zero on Ghost Object deletion to facilitate respawn properly. (FSO can get confused between two objects with the same net_signature if it tries to look them up via net_signature, so set the dead one to 0 and the new one from the parse object will be set correctly)

The other fix is during dogfight missions and has to do with errant rearm requests (which are invalid because the player is a traitor in dogfight mode).
It is actually a fix for [Mantis 2175](http://scp.indiegames.us/mantis/view.php?id=2175
) 

From my comments on that ticket. "Checking explicitly for a rearm or repair request before canceling it for traitors is the fix. "

*These changes have been part of all builds that I have sent out during the entire multi testing period.*  I just got good enough at git to separate them from unrelated waypoint changes that require a multi bump.

This also has some variable renames in multi_util.cpp which make a few functions much more readable.  Those can be dropped easily on request.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -11991,12 +11991,13 @@ void ai_do_repair_frame(object *objp, ai_info *aip, float frametime)
 {
 	static bool rearm_eta_found=false;
 
-	if (Ships[objp->instance].team == Iff_traitor) {
-		ai_abort_rearm_request(objp);
-		return;
-	}
 
 	if (aip->ai_flags[AI::AI_Flags::Being_repaired, AI::AI_Flags::Awaiting_repair]) {
+		if (Ships[objp->instance].team == Iff_traitor) {
+			ai_abort_rearm_request(objp);
+			return;
+		}
+
 		int	support_objnum;
 		
 		ai_info	*repair_aip;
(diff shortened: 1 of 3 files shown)


code/ai/aicode.cpp
```diff
@@ -11983,7 +11983,7 @@ void ai_do_repair_frame(object *objp, ai_info *aip, float frametime)
 {
 	static bool rearm_eta_found=false;
 
-	if (Ships[objp->instance].team == Iff_traitor) {
+	if ( (Ships[objp->instance].team == Iff_traitor) && ((aip->ai_flags[AI::AI_Flags::Awaiting_repair]) || (aip->ai_flags[AI::AI_Flags::Being_repaired])) ) {
```
Goober5000, line 11986: Why not move the traitor check and make it the first statement inside the next `if()` block?

That would also allow the `else` at the end of the function to properly execute.

JohnAFernandez: I am honestly not sure any more why I did it exactly this way.  Let me take a look.

JohnAFernandez: Ok.  Yes, that's a good idea.

PR #2702, opened 2020-08-31
prevent duplicate entries in the Intel room

Goober5000:


test/test_data/menuui/intel_parse/always_techroom_values/data/tables/species.tbl
```diff
@@ -26,12 +26,11 @@ XSTR(
...
 $Entry:
-$Name: XSTR("Baz name",3004)
-$Anim: Baz anim
+$Name: XSTR("Qux name",-1)
+$Anim: Qux anim
 $AlwaysInTechRoom: -43
 $Description:
 XSTR(
-"Baz desc", 3005)
+"Qux desc", -1)
```
JohnAFernandez, line 35: Ok.  Question, why did we switch these indices to -1 if we have this comment next to the declaration: 	`"Qux desc", // XSTR id 3005` ?

Goober5000: Because "Baz desc" has an entry in the XSTR table, but "Qux desc" does not.  But that's a good catch; that means the comments at the top of the file are wrong.  I'll fix that.

wookieejedi (approved):

PR #2703, opened 2020-08-31
handle audio pause/unpause for all of the F2 screens

Goober5000:
This is a follow-up to #2100.  Don't pause/unpause the audio if we are moving from/to the Control Config or HUD Config screens.  We could hit F2 and move around a bit before resuming the game, so we don't want the audio to resume until we're ready.


JohnAFernandez (approved):
LGTM

It helps that I've made these types of edits before.

PR #2705, opened 2020-09-01
Fix 2526 and Add Two Game_setting.tbl flags

### LINKED ISSUE #2526: Shockwave damage is applied unevenly against large object groups
Shockwaves keep track of the objects they already damaged in order to make sure that they only apply their damage effects once against each target. In cases where very large shockwaves (like the Colossus explosion) hit large groups of objects, the 64 slots shockwaves have to keep track of their victims can get filled; once this happens, debug builds crash and release builds start acting weird.
From shockwave.cpp, line 300 onwards, in shockwave_move:
```		// only apply damage to a ship once from a shockwave
		for ( i = 0; i < sw->num_objs_hit; i++ ) {
			if ( objp->signature == sw->obj_sig_hitlist[i] ){
				break;
			}
		}

		if ( i < sw->num_objs_hit ){
			continue;
		}

		if ( weapon_area_calc_damage(objp, &sw->pos, sw->inner_radius, sw->outer_radius, sw->blast, sw->damage, &blast, &damage, sw->radius) == -1 ){
			continue;
		}

		// okay, we have damage applied, record the object signature so we don't repeatedly apply damage
		Assert(sw->num_objs_hit < SW_MAX_OBJS_HIT);
		if ( sw->num_objs_hit >= SW_MAX_OBJS_HIT) {
			sw->num_objs_hit--;
		}
```

It seems to me that this code would cause the last object added to the list of hit objects to be continuously changed as objects are being processed, leading to those objects taking a lot more damage than intended.

JohnAFernandez:
Fixes #2526  by making the array a vector and also adds a Shockwaves always damage bombs game_setting.tbl setting and a Shockwaves_consistently_apply_damage game_settings.tbl setting.  The second one keeps bombs and asteroids from being hit more than once by shockwaves, like ships do!

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -50,6 +50,8 @@ bool Disable_built_in_translations;
 bool Weapon_shockwaves_respect_huge;
 bool Using_in_game_options;
 float Dinky_shockwave_default_multiplier;
+bool Shockwaves_always_damage_bombs;
+bool Shockwaves_damage_all_obj_types_once;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_damage_p1;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_damage_p2;
 std::tuple<ubyte, ubyte, ubyte> Arc_color_damage_s1;
@@ -538,6 +540,14 @@ void parse_mod_table(const char *filename)
 			}
 		}
 
+		if (optional_string("$Shockwaves Always Damage Bombs:")) {
+			stuff_boolean(&Shockwaves_always_damage_bombs);
+		}
+
+		if (optional_string("$Shockwaves Damage All Object Types Once:")) {
+			stuff_boolean(&Shockwaves_damage_all_obj_types_once);
+		}
+
 		if (optional_string("$Use Engine Wash Intensity:")) {
 			stuff_boolean(&Use_engine_wash_intensity);
 		}
@@ -611,6 +621,8 @@ void mod_table_reset()
 	Weapon_shockwaves_respect_huge = false;
 	Using_in_game_options = false;
 	Dinky_shockwave_default_multiplier = 1.0f;
+	Shockwaves_always_damage_bombs = false;
+	Shockwaves_damage_all_obj_types_once = false;
 	Arc_color_damage_p1 = std::make_tuple(static_cast<ubyte>(64), static_cast<ubyte>(64), static_cast<ubyte>(225));
 	Arc_color_damage_p2 = std::make_tuple(static_cast<ubyte>(128), static_cast<ubyte>(128), static_cast<ubyte>(255));
 	Arc_color_damage_s1 = std::make_tuple(static_cast<ubyte>(200), static_cast<ubyte>(200), static_cast<ubyte>(255));
--- a/code/mod_table/mod_table.h
+++ b/code/mod_table/mod_table.h
@@ -42,6 +42,8 @@ extern bool Disable_built_in_translations;
 extern bool Weapon_shockwaves_respect_huge;
 extern bool Using_in_game_options;
 extern float Dinky_shockwave_default_multiplier;
+extern bool Shockwaves_always_damage_bombs;
+extern bool Shockwaves_damage_all_obj_types_once;
 extern std::tuple<ubyte, ubyte, ubyte> Arc_color_damage_p1;
 extern std::tuple<ubyte, ubyte, ubyte> Arc_color_damage_p2;
 extern std::tuple<ubyte, ubyte, ubyte> Arc_color_damage_s1;
--- a/code/weapon/shockwave.cpp
+++ b/code/weapon/shockwave.cpp
@@ -128,7 +128,7 @@ int shockwave_create(int parent_objnum, vec3d* pos, shockwave_create_info* sci, 
 	sw->blast = sci->blast;
 	sw->radius = 1.0f;
 	sw->pos = *pos;
-	sw->num_objs_hit = 0;
+	sw->obj_sig_hitlist.clear();
 	sw->shockwave_info_index = info_index;		// only one type for now... type could be passed is as a parameter
 	sw->current_bitmap = -1;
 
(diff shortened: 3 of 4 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -538,6 +540,14 @@ void parse_mod_table(const char *filename)
 			}
 		}
 
+		if (optional_string("$Shockwaves Always Damage Bombs:")) {
+			stuff_boolean(&Shockwaves_always_damage_bombs);
+		}
+
+		if (optional_string("$Shockwaves Damage Non-ships Once:")) {
```
Goober5000, line 547: I would call this `$Shockwaves Damage Bombs and Asteroids Once:` otherwise modders might wonder if ships are damaged more than once.

It's hard to think of a good table string here.  The bool variable name is more accurate but not very explanatory.

code/weapon/shockwave.cpp
```diff
@@ -128,7 +128,6 @@ int shockwave_create(int parent_objnum, vec3d* pos, shockwave_create_info* sci,
 	sw->blast = sci->blast;
 	sw->radius = 1.0f;
 	sw->pos = *pos;
-	sw->num_objs_hit = 0;
```
Goober5000, line 131: Shouldn't this be replaced with `sw->obj_sig_hitlist.clear()`?

code/weapon/shockwave.cpp
```diff
@@ -297,32 +295,26 @@ void shockwave_move(object *shockwave_objp, float frametime)
...
 		}
 
-		// only apply damage to a ship once from a shockwave
-		for ( i = 0; i < sw->num_objs_hit; i++ ) {
-			if ( objp->signature == sw->obj_sig_hitlist[i] ){
-				break;
+		// only apply damage to a on object once from a shockwave if it's on the list
+		for (auto comparison : sw->obj_sig_hitlist) {
+			if ( (objp->signature == comparison.first) && (objp->type == comparison.second) ){
+				continue;
```
Goober5000, line 301: This is only going to continue the comparison loop.  You want to set some flag here, break out of this loop, then continue the outer loop.  (That was the purpose of the `if ( i < sw->num_objs_hit )` in the original code.)

code/weapon/shockwave.cpp
```diff
@@ -297,32 +295,26 @@ void shockwave_move(object *shockwave_objp, float frametime)
...
 
-		if ( i < sw->num_objs_hit ){
-			continue;
-		}
-
 		if ( weapon_area_calc_damage(objp, &sw->pos, sw->inner_radius, sw->outer_radius, sw->blast, sw->damage, &blast, &damage, sw->radius) == -1 ){
 			continue;
 		}
 
-		// okay, we have damage applied, record the object signature so we don't repeatedly apply damage
```
Goober5000, line 315: Keep this comment, even if the code following it is removed.

code/weapon/shockwave.cpp
```diff
@@ -297,32 +295,26 @@ void shockwave_move(object *shockwave_objp, float frametime)
...
-		Assert(sw->num_objs_hit < SW_MAX_OBJS_HIT);
-		if ( sw->num_objs_hit >= SW_MAX_OBJS_HIT) {
-			sw->num_objs_hit--;
-		}
-
 		weapon_info* wip = NULL;
+		
+		// only add non-ships to the list if the Game_settings flag is set
+		if (objp->type == OBJ_SHIP || Shockwaves_consistently_apply_damage){
+		sw->obj_sig_hitlist.push_back({objp->signature, objp->type});
```
Goober5000, line 313: Use `emplace_back` here and remove the curly braces around the pair.  Also, it looks like the statement needs to be indented.

code/mod_table/mod_table.cpp
JohnAFernandez: How about Shockwaves_damage_all_obj_types_once ?

Goober5000: There you go, that sounds good. :+1:

code/mod_table/mod_table.cpp
```diff
@@ -538,6 +540,14 @@ void parse_mod_table(const char *filename)
 			}
 		}
 
+		if (optional_string("$Shockwaves Always Damage Bombs:")) {
+			stuff_boolean(&Shockwaves_always_damage_bombs);
+		}
+
+		if (optional_string("$Shockwaves Damage Bombs and Asteroids Once:")) {
```
Goober5000, line 547: Change the table string to `$Shockwaves Damage All Object Types Once:` too.

code/weapon/shockwave.cpp
```diff
@@ -297,32 +296,34 @@ void shockwave_move(object *shockwave_objp, float frametime)
 			}
 		}
 
-		// only apply damage to a ship once from a shockwave
-		for ( i = 0; i < sw->num_objs_hit; i++ ) {
-			if ( objp->signature == sw->obj_sig_hitlist[i] ){
+		bool found_in_list = false;
+
+		// only apply damage to an object once from a shockwave
+		for (auto comparison : sw->obj_sig_hitlist) {
```
Goober5000, line 302: I think you want `auto &` here.

code/mod_table/mod_table.h
```diff
@@ -42,6 +42,8 @@ extern bool Disable_built_in_translations;
 extern bool Weapon_shockwaves_respect_huge;
 extern bool Using_in_game_options;
 extern float Dinky_shockwave_default_multiplier;
+extern bool Shockwaves_always_damage_bombs;
+extern bool Shockwaves_consistently_apply_damage;
```
Goober5000, line 46: This variable still needs to be renamed.

code/mod_table/mod_table.cpp
```diff
@@ -538,6 +540,14 @@ void parse_mod_table(const char *filename)
 			}
 		}
 
+		if (optional_string("$Shockwaves Always Damage Bombs:")) {
+			stuff_boolean(&Shockwaves_always_damage_bombs);
+		}
+
+		if (optional_string("$Shockwaves Damage All Object Types Once")) {
```
Goober5000, line 547: So close.  Missed the colon at the end. :)

JohnAFernandez: Not again!!!

code/weapon/shockwave.cpp
```diff
@@ -297,32 +296,34 @@ void shockwave_move(object *shockwave_objp, float frametime)
...
-		if ( sw->num_objs_hit >= SW_MAX_OBJS_HIT) {
-			sw->num_objs_hit--;
-		}
-
 		weapon_info* wip = NULL;
+		
+		// okay, we have damage applied, record the object signature so we don't repeatedly apply damage 
+		// but only add non-ships to the list if the Game_settings flag is set
+		if (objp->type == OBJ_SHIP || Shockwaves_damage_all_obj_types_once){
+		sw->obj_sig_hitlist.emplace_back(objp->signature, objp->type);
```
Goober5000, line 322: Does this need to be indented or is the indentation correct?

JohnAFernandez: Blarg

Issue #2650, opened 2020-08-13
Fonts.tbl limitations 


JohnAFernandez:
Because the first 3 fonts in fonts.tbl are reserved for specific uses in the engine, and only the first 4 fonts are enabled to display special characters on behalf of true type fonts (see `LCL_MAX_FONTS`), if all three of the first entries are true type fonts, it is easy to run into a situation where there isn't enough room to provide special characters via .vf fonts for all three true type fonts.

Pseudo example:

Entry 1: truetype1
+Special Character Font: Vf1
Entry 2: truetype2
+Special Character Font: Vf2
Entry 3: truetype3
+Special Character Font: Vf3

Entry 4: Vf1
Entry 5: Vf2
Entry 6: Vf3

The first three true types point to the other three volition-type fonts.  Vf3 hits the assert (at least on standalone):
`Assertion((font_num >= 0) && (font_num < LCL_MAX_FONTS), "Passed an invalid font index");`
in `localize.cpp:524`

The simplest solution might be to bump the LCL_MAX_FONTS limit. But I am extremely unfamiliar with this part of the code and I don't know what kind of side-effect this could cause, or even what type of side effects I should be on the lookout for.

JohnAFernandez:
This should now be taken care of.

PR #2706, opened 2020-09-01
Change special_char_indexes to a vector

JohnAFernandez:
Without this change certain font tables can hit the meaningless LCL_MAX_FONTS limit, as fotg did.

LCL_MAX_FONTS is kept around just to facilitate parsing of the hard-coded table.

I am not an expert by an means in this part of the code.  I ran both retail and fotg, which uses a font table with more than 5 entries and it seems to work.

Related to #2650

### DIFF
--- a/code/graphics/software/font.cpp
+++ b/code/graphics/software/font.cpp
@@ -249,8 +249,22 @@ namespace
 					  fontName.c_str(), MAX_SPECIAL_CHAR_IDX);
 			}
 
-			for (auto i = 0; i < (int)Lcl_languages.size(); ++i) {
-				Lcl_languages[i].special_char_indexes[font_id] = (ubyte)default_special_char_index;
+		}
+
+		// add the index specified to all the languages 
+		for (auto & Lcl_language : Lcl_languages) {
+
+			// if there wasn't already something specified for this font.
+			if (font_id >= (int)Lcl_language.special_char_indexes.size()) {
+				// if a lot of indexes were missing, add some defaults to fill in the gaps until ...
+				while (font_id > (int)Lcl_language.special_char_indexes.size()) {
+					Lcl_language.special_char_indexes.push_back(0);
+				}
+				// we're at just the right place in the vector to add the index we just read from the table.
+				Lcl_language.special_char_indexes.push_back((ubyte)default_special_char_index);
+			} // if there was something already specified that's being replaced.
+			else {
+				Lcl_language.special_char_indexes[font_id] = (ubyte)default_special_char_index;
 			}
 		}
 
--- a/code/localization/localize.cpp
+++ b/code/localization/localize.cpp
@@ -190,20 +190,22 @@ void parse_stringstbl_quick(const char *filename)
 
 				if (!Unicode_text_mode) {
 					required_string("+Special Character Index:");
-					stuff_ubyte(&language.special_char_indexes[0]);
-					for (i = 1; i < LCL_MAX_FONTS; ++i) {
+					ubyte temp;
+					stuff_ubyte(&temp);
+					language.special_char_indexes.push_back(temp);
+					for (i = 1; i < LCL_MINIMUM_FONTS; ++i) {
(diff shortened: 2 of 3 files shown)


Goober5000:
How is `LCL_MAX_FONTS` used to facilitate parsing of the hard-coded table?  I searched the codebase and the only place I see it used is with `special_char_indexes`.

JohnAFernandez:
Oh, that's *not* hard-coded....  I'm still using it as a bound in, `void parse_stringstbl_quick(const char *filename)` because it matches retail behavior.  My assumption is that beyond those 5 fonts, you have to use unicode mode?
`
This pr was honestly a little bit of a shot in the dark.

Goober5000:
You can probably remove it entirely.  Originally there were only three fonts, and I bumped it to five for the purposes of using one or two extra fonts in the fiction viewer.  That was way before TTF support and all the additional uses for fonts we have now.

code/graphics/software/font.cpp
```diff
@@ -250,7 +250,10 @@ namespace
 			}
 
 			for (auto i = 0; i < (int)Lcl_languages.size(); ++i) {
-				Lcl_languages[i].special_char_indexes[font_id] = (ubyte)default_special_char_index;
+				while (font_id > (int)Lcl_languages[i].special_char_indexes.size()) {
+					Lcl_languages[i].special_char_indexes.push_back(0);
+				}
+				Lcl_languages[i].special_char_indexes.push_back((ubyte)default_special_char_index);
```
Goober5000, line 256: Is the `push_back` actually correct here?  It seems to be that you still need to access the `[font_id]` entry.

code/localization/localize.cpp
```diff
@@ -190,20 +190,22 @@ void parse_stringstbl_quick(const char *filename)
 
 				if (!Unicode_text_mode) {
 					required_string("+Special Character Index:");
-					stuff_ubyte(&language.special_char_indexes[0]);
+					ubyte temp;
+					stuff_ubyte(&temp);
+					language.special_char_indexes.push_back(temp);
 					for (i = 1; i < LCL_MAX_FONTS; ++i) {
```
Goober5000, line 196: You can remove the `LCL_MAX_FONTS` limit entirely, per my other comment.

code/localization/localize.cpp
```diff
@@ -190,20 +190,22 @@ void parse_stringstbl_quick(const char *filename)
...
-							language.special_char_indexes[i] = 0;
+							language.special_char_indexes.push_back(0);
 						}
 					}
 				} else {
 					// Set all indices to valid values
-					for (i = 0; i < LCL_MAX_FONTS; ++i) {
-						language.special_char_indexes[i] = 0;
+					for (i = 1; i < LCL_MAX_FONTS; ++i) {
+						language.special_char_indexes.push_back(0);
```
Goober5000, line 208: Same question re `push_back` vs. `[]` index.

code/graphics/software/font.cpp
JohnAFernandez: I think here I was working under the assumption that the font specified would be a new font, but if was already specified, I guess we do need to access through [font_id].  I have made that change in the last commit.

code/localization/localize.cpp
JohnAFernandez: Here the languages are being defined for the first time, and the vectors are being built, so attempting to access via `[] index` would result in a crash.

JohnAFernandez:
Signed vs. unsigned. Every. Single. Time.

code/graphics/software/font.cpp
Goober5000: Yeah, I'm referring to the second `push_back` here, not the first one.

JohnAFernandez: Right, in the event that that item in the vector doesn't exist yet, it creates any default entries that need to exist in between and then the second `push_back()` places it at the font_id index.

Goober5000 (approved):

PR #2707, opened 2020-09-02
fix main hall indexing in modular tables

Goober5000:
This is a follow-up to #2675 which fixes the main hall naming for halls which have custom names and are not the first resolution.


JohnAFernandez (approved):
LGTM, simple fixes are the best

PR #2712, opened 2020-09-04
add multi support for red-alert sexp

Goober5000:
This makes the red-alert SEXP execute properly in multiplayer.

Note, this is just the SEXP and a bit of cleanup.  Multi-knowledgeable people will need to evaluate `red_alert_maybe_move_to_next_mission` to make sure that function will work properly in multiplayer.

### DIFF
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -1051,20 +1051,23 @@ void red_alert_start_mission()
 //		return;
 //	}
 
-	// check player health here.  
+	// check player health here, but only in single-player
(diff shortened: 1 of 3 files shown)


Goober5000:
And #2631 takes care of all the state transitioning code that I was wondering about.

JohnAFernandez:
I can no longer reproduce the bug that was holding these up.  Either something changed, or it is a rare edge case bug that will be hard to track down.

JohnAFernandez (approved):
This seems to be good to go.

PR #2714, opened 2020-09-05
Fix a bug with form-on-wing on a object outside the wing

Baezon:
`leader_objp` can't be relied on to provide the formation since it could be some random object the wing was ordered to form up on. Now the wingnum is passed in so the formation is based on whoever is doing the forming up. Also update the new logic for the autopilot version of `get_absolute_wing_pos`.

### DIFF
--- a/code/ai/ai.h
+++ b/code/ai/ai.h
@@ -580,7 +580,7 @@ extern void ai_update_danger_weapon(int objnum, int weapon_objnum);
 
 // called externally from MissionParse.cpp to position ships in wings upon arrival into the
 // mission.
-extern void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wing_index, int wingnum, bool formation_object_flag, bool autopilot = false);
+extern void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wingnum, int wing_index, bool formation_object_flag, bool autopilot = false);
 
 //	Interface from goals code to AI.  Set ship to guard.  *objp guards *other_objp
 extern void ai_set_guard_object(object *objp, object *other_objp);
--- a/code/ai/ai.h
+++ b/code/ai/ai.h
@@ -580,8 +580,8 @@ extern void ai_update_danger_weapon(int objnum, int weapon_objnum);
 
 // called externally from MissionParse.cpp to position ships in wings upon arrival into the
 // mission.
-extern void get_absolute_wing_pos( vec3d *result_pos, object *leader_objp, int wing_index, int formation_object_flag);
-extern void get_absolute_wing_pos_autopilot( vec3d *result_pos, object *leader_objp, int wing_index, int formation_object_flag);
+extern void get_absolute_wing_pos( vec3d *result_pos, object *leader_objp, int wing_index, int formation_object_flag, int wingnum);
+extern void get_absolute_wing_pos_autopilot( vec3d *result_pos, object *leader_objp, int wing_index, int formation_object_flag, int wingnum);
 
 //	Interface from goals code to AI.  Set ship to guard.  *objp guards *other_objp
 extern void ai_set_guard_object(object *objp, object *other_objp);
--- a/code/ai/ai.h
+++ b/code/ai/ai.h
@@ -580,8 +580,7 @@ extern void ai_update_danger_weapon(int objnum, int weapon_objnum);
 
 // called externally from MissionParse.cpp to position ships in wings upon arrival into the
 // mission.
-extern void get_absolute_wing_pos( vec3d *result_pos, object *leader_objp, int wing_index, int formation_object_flag, int wingnum);
-extern void get_absolute_wing_pos_autopilot( vec3d *result_pos, object *leader_objp, int wing_index, int formation_object_flag, int wingnum);
+extern void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wing_index, int wingnum, bool formation_object_flag, bool autopilot = false);
 
 //	Interface from goals code to AI.  Set ship to guard.  *objp guards *other_objp
 extern void ai_set_guard_object(object *objp, object *other_objp);
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -4376,7 +4376,7 @@ void ai_fly_to_target_position(vec3d* target_pos, bool* pl_done_p=NULL, bool* pl
 
 				if (wing_leader != Pl_objp) {
 					// not wing leader.. get my position relative to wing leader
-					get_absolute_wing_pos(&goal_point, wing_leader, wing_index, shipp->wingnum, aip->ai_flags[AI::AI_Flags::Formation_object], true);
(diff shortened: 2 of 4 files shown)


Goober5000 (changes requested):

code/ai/ai.h
```diff
@@ -580,8 +580,8 @@ extern void ai_update_danger_weapon(int objnum, int weapon_objnum);
 
 // called externally from MissionParse.cpp to position ships in wings upon arrival into the
 // mission.
-extern void get_absolute_wing_pos( vec3d *result_pos, object *leader_objp, int wing_index, int formation_object_flag);
-extern void get_absolute_wing_pos_autopilot( vec3d *result_pos, object *leader_objp, int wing_index, int formation_object_flag);
+extern void get_absolute_wing_pos( vec3d *result_pos, object *leader_objp, int wing_index, int formation_object_flag, int wingnum);
+extern void get_absolute_wing_pos_autopilot( vec3d *result_pos, object *leader_objp, int wing_index, int formation_object_flag, int wingnum);
```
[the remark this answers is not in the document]
Goober5000: I meant wingnum followed by wing_index.  That's the way most functions dealing with wings list their arguments.

code/ai/aicode.cpp
```diff
@@ -11443,49 +11443,34 @@ DCF(wing_scale, "Adjusts the wing formation scale. (Default is 1.0f)")
...
 {
 	vec3d	wing_delta, rotated_wing_delta;
 	float		wing_spread_size;
-	int formation_index = Wings[Ships[leader_objp->instance].wingnum].formation;
+	int formation_index = Wings[wingnum].formation;
 
 	if (wing_index == 0)
 		wing_delta = vmd_zero_vector;
-	else if ( formation_index != -1) 
+	else if ( formation_index != -1 && wing_index < 6)
```
Goober5000, line 11457: Use `MAX_SHIPS_PER_WING` here.

code/ai/ai.h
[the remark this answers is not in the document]
Baezon: Ah, I did not know there was a convention.

code/ai/aicode.cpp
```diff
@@ -11523,7 +11508,7 @@ void render_wing_phantoms(object *objp)
 
 	for (i=0; i<32; i++)
 		if (Debug_render_wing_phantoms & (1 << i)) {
-			get_absolute_wing_pos(&goal_point, objp, i, 0);
+			get_absolute_wing_pos(&goal_point, objp, i, shipp->wingnum, false);
```
Goober5000, line 11511: Since `autopilot` defaults to false, you don't need to specify it here.  You can if you want, of course.

code/mission/missionparse.cpp
```diff
@@ -6202,7 +6202,7 @@ void mission_set_wing_arrival_location( wing *wingp, int num_to_set )
 
 				// change the position of the next ships in the wing.  Use the cool function in AiCode.cpp which
 				// Mike K wrote to give new positions to the wing members.
-				get_absolute_wing_pos( &objp->pos, leader_objp, wing_index++, 0);
+				get_absolute_wing_pos( &objp->pos, leader_objp, wing_index++, Ships[objp->instance].wingnum, false);
```
Goober5000, line 6205: Same here with the optional `false`.

code/ai/ai.h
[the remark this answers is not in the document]
Goober5000: Ya, look at `hud_set_wingman_status_departed` for example.

code/ai/aicode.cpp
Baezon: That (and the one below) is actually the `formation_object` flag, this was the '0' previously but I changed to a bool.

Goober5000: Oh, whoops. >.>

code/ai/aicode.cpp
```diff
@@ -4376,7 +4376,7 @@ void ai_fly_to_target_position(vec3d* target_pos, bool* pl_done_p=NULL, bool* pl
 
 				if (wing_leader != Pl_objp) {
 					// not wing leader.. get my position relative to wing leader
-					get_absolute_wing_pos_autopilot(&goal_point, wing_leader, wing_index, aip->ai_flags[AI::AI_Flags::Formation_object]);
+					get_absolute_wing_pos(&goal_point, wing_leader, shipp->wingnum, wing_index, aip->ai_flags[AI::AI_Flags::Formation_object], true);
```
Goober5000, line 4379: Although `shipp->wingnum` is the same as `aip->wing`, use `aip->wing` as the argument to this function because it's consistent with the pattern throughout the rest of this block.

code/ai/aicode.cpp
```diff
@@ -11523,7 +11508,7 @@ void render_wing_phantoms(object *objp)
 
 	for (i=0; i<32; i++)
 		if (Debug_render_wing_phantoms & (1 << i)) {
-			get_absolute_wing_pos(&goal_point, objp, i, 0);
+			get_absolute_wing_pos(&goal_point, objp, shipp->wingnum, i, false);
```
Goober5000, line 11511: Just use plain old `wingnum` instead of `shipp->wingnum` here because that's what's being referenced earlier in the function

code/ai/aicode.cpp
```diff
@@ -11751,7 +11736,7 @@ int ai_formation()
 	leader_speed = leader_objp->phys_info.speed;
 	vec3d leader_vec = leader_objp->phys_info.vel;
 
-	get_absolute_wing_pos(&goal_point, leader_objp, wing_index, aip->ai_flags[AI::AI_Flags::Formation_object]);
+	get_absolute_wing_pos(&goal_point, leader_objp, Ships[Pl_objp->instance].wingnum, wing_index, aip->ai_flags[AI::AI_Flags::Formation_object]);
```
Goober5000, line 11739: I would also use `aip->wing` here

Baezon: heh way ahead of ya ;)

code/mission/missionparse.cpp
```diff
@@ -6202,7 +6202,7 @@ void mission_set_wing_arrival_location( wing *wingp, int num_to_set )
 
 				// change the position of the next ships in the wing.  Use the cool function in AiCode.cpp which
 				// Mike K wrote to give new positions to the wing members.
-				get_absolute_wing_pos( &objp->pos, leader_objp, wing_index++, 0);
+				get_absolute_wing_pos( &objp->pos, leader_objp, Ships[objp->instance].wingnum, wing_index++, false);
```
Goober5000, line 6205: Use `WING_INDEX(wingp)`

Baezon: Ah, I was looking for that kind of functionality.

Goober5000 (changes requested):
A few more things, none of which are strictly necessary but they make the code more maintainable.

Also, take a look at `ai_formation_object_get_slotnum`.  Per the comment, this should be improved by caching the slot number in the `ai_info` struct.  After this PR is done, can you make another PR to handle that?

Baezon:
Seems reasonable.

Goober5000 (approved):

Issue #2364, opened 2020-05-04
Wrong number of subexplosions spawned for ships during deathroll while using time compression (using ScreenCam Script)


MitoPL:
If you are in the mission space and using any sort of time compression, the amount of subexplosions for any ship during its deathroll isn't the same as when not using any time compression (x1). Especially visible when using the ScreenCam Script with its x0.001 compression.
To recreate: load up any mission and have ScreenCam Script in your game files, reach a moment when some larger ship goes down to 0% hull integrity, and then engage ScreenCam's x0.001 time compression and wait a bit. If you disable the compression, you'll be greeted with a massive cloud of explosions around the ship.

It was mentioned on HLP's Discord that it looks as if the subexplosion spawning is bound to the framerate, not absolute timescale. I don't know, just reporting.

JohnAFernandez:
I have a possible fix for this.  About to be tested.

JohnAFernandez:
It did not end up being a timing issue.  Switching to a higher resolution timestamp did nothing.

So my best guess for this is it has something to do with the model code.  It turns out, the more polies a model has, the more this happens.

JohnAFernandez:
Long story short, there was another script that was spawning fireballs with incorrect timing.

PR #2716, opened 2020-09-06
Fix fireball spawn for small time compression factors

JohnAFernandez:
*Needs more testing to ensure that debris, ship halves and Knossos all explode like in retail but do not spawn too many explosions in screencam script, but feel free to review, as afaict, this is the final version and _should_ work*

Switches time calculation to flFrametime to fix too many explosions showing up,

### DIFF
--- a/code/debris/debris.cpp
+++ b/code/debris/debris.cpp
@@ -264,9 +264,10 @@ void debris_process_post(object * obj, float frame_time)
 		return;			// If arc_frequency <= 0, this piece has no arcs on it
 	}
 
-	if ( !timestamp_elapsed(db->fire_timeout) && timestamp_elapsed(db->next_fireball))	{		
+	if ( !timestamp_elapsed(db->fire_timeout) && (db->next_fireball < flFrametime))	{		
 
-		db->next_fireball = timestamp_rand(db->arc_frequency,db->arc_frequency*2 );
+		// use timestamp frequency to convert to float version of time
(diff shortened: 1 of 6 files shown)


JohnAFernandez:
For issue #2364

asarium (changes requested):
Can you go into detail what the original issue was?
The timestamp system exists for a very good reason since it does not suffer from floating point inaccuracies so introducing that again with these changes would be a step backwards.

JohnAFernandez:
Basically, when the screencam script is used, it reduces time compression to .001 or some crazy factor.  When that happens, timestamp starts becoming inaccurate, I assume because it lacks the resolution to properly keep time at that time compression, so many explosions will spawn instead of the amount without time compression.

asarium:
Yes, I think the time stamp system can only be accurate to something like 1/65536th of a second so at some point it will not be able to increment the timer properly and will instead increment is by 1/65536th every frame. However, I do not see how that could cause the described issues if the timestamps are reset properly after they get fired once.

Please look a bit more into this to determine if there is a fundamental timer issue here of if the relevant code maybe simply misused the timestamp system.

JohnAFernandez:
Ok.  I'll take a deeper look.

JohnAFernandez:
Ok.  It looks like you were right.  I tried writing a quick copy of the timestamp system but which uses microseconds instead of seconds and the increased resolution does not fix the problem.  It does seem like there is something wrong with the logic itself.  I will continue to investigate.

JohnAFernandez:
Turns out, afaict, it's not timestamp related at all.

PR #2719, opened 2020-09-08
Store and reuse formation object slotnum

Baezon:
Every frame every ship flying in formation object mode loops through all objects in order to find its slotnum to determine its rank and file position in the formation. This instead stores the number in `ai_info` and only recalculates when an order to enter or leave formation object mode is processed or a ship that was flying in the mode dies (and also does so for all ships flying off a given ship in one pass).

### DIFF
--- a/code/ai/ai.h
+++ b/code/ai/ai.h
@@ -487,6 +487,8 @@ typedef struct ai_info {
 	flagset<AI::Maneuver_Override_Flags>	ai_override_flags;			// flags for marking ai overrides from sexp or lua systems
 	control_info	ai_override_ci;		// ai override control info
 	int		ai_override_timestamp;		// mark for when to end the current override
+
+	int form_obj_slotnum;               // for flying in formation object mode, the position in the formation
 } ai_info;
 
 // Goober5000
@@ -662,5 +664,7 @@ void ai_update_aim(ai_info *aip);
 
 //SUSHI: Random evasive sidethrust
 void do_random_sidethrust(ai_info *aip, ship_info *sip);
+
+void ai_formation_object_recalculate_slotnums(int form_objnum, int dying_objnum = -1);
 
 #endif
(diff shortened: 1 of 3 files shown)


code/ai/aicode.cpp
```diff
@@ -15210,6 +15213,10 @@ void ai_ship_destroy(int shipnum)
 		if (aip->hitter_objnum == objnum)
 			aip->hitter_objnum = -1;
 	}
+
+	if (dead_aip->ai_flags[AI::AI_Flags::Formation_object])
+		ai_formation_object_recalculate_slotnums(dead_aip->goal_objnum, shipnum);
```
Goober5000, line 15218: You're passing `shipnum` here, but the function needs `objnum`.

code/ai/aicode.cpp
```diff
@@ -3379,33 +3379,36 @@ void ai_form_on_wing(object *objp, object *goal_objp)
...
 
 /**
- * Given an object and an object on whose wing to form, return slot to use.
- *
- * @todo This function is called per object in formation per frame.  Should store slot in ::ai_info struct.
+ * Redistributes slotnums to all ships flying off a given formation object
+ * In the case this is needed because a ship is dying, its objnum is needed to explicitly exclude it
  */
-int ai_formation_object_get_slotnum(int objnum, object *objp)
+void ai_formation_object_recalculate_slotnums(int form_objnum, int dying_objnum)
```
Goober5000, line 3394: Change `dying_objnum` to `exiting_objnum`, since the ship could also be departing or vanishing.

Goober5000 (changes requested):
Looks good except for a couple things.

code/ai/ai.h
```diff
@@ -663,4 +665,6 @@ void ai_update_aim(ai_info *aip);
 //SUSHI: Random evasive sidethrust
 void do_random_sidethrust(ai_info *aip, ship_info *sip);
 
+void ai_formation_object_recalculate_slotnums(int form_objnum, int dying_objnum = -1);
```
Goober5000, line 668: Missed a spot. :)

Goober5000 (changes requested):
Also, initialize `form_obj_slotnum` in `init_ai_object`.

Goober5000 (approved):

PR #2722, opened 2020-09-09
Revert "Change special_char_indexes to a vector"

Goober5000:
Reverts scp-fs2open/fs2open.github.com#2706

Will fix #2721

PR #2725, opened 2020-09-09
Minor Refactor of Special Character Index Handling

JohnAFernandez:
Now tested on Retail, Admiral Nelson's 4K mod and FotG . Special characters all appear normally.

1. Stores the filename from the table Parser into the font struct so we can lookup fonts via their filename.
2. Allow "+Special Character Font:" to specify the first entry with the same filename's special character index for their own special character index, provide a default if not, or allow the modder to specify them manually, but do not override already recorded values if the default is picked in order to protect retail behavior.
3. Vectorize Special Character Indexes, so that we don't have a table limit that will randomly crash FSO based on what table you happened to write! Woohoo! But provide a minimum to again protect retail behavior.

### DIFF
--- a/code/graphics/software/FSFont.cpp
+++ b/code/graphics/software/FSFont.cpp
@@ -25,6 +25,11 @@ namespace font
 		this->name = newName;
 	}
 
+	void FSFont::setFilename(const SCP_string& newName) 
+	{
+		this->filename = newName;
+	}
+
 	float FSFont::getBottomOffset() const
 	{
 		return this->offsetBottom;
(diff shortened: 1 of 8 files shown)


JohnAFernandez:
tl:dr version, it was too complicated to just remove the limit without also allowing the table parser to use a previous Special Character Index.

Goober5000:
Follow-up to #2722 and #2706

Goober5000:
Shrug of approval

Goober5000 (approved):

PR #2727, opened 2020-09-09
Remove all int3s in Freespace.cpp, ai.cpp, and aibig.cpp

JohnAFernandez:


freespace2/freespace.cpp
```diff
@@ -3629,17 +3628,13 @@ void game_maybe_do_dead_popup(float frametime)
 				break;
 
 			// this should only happen during a red alert mission
-			case 3:				
+			case 3:			
+				Assertion(The_mission.flags[Mission::Mission_Flags::Red_alert], "FSO tried to reach a Red_alert state without being in a Red_alert mission.");
```
Goober5000, line 3632: This is not reaching a red-alert state; this is choosing the previous-mission option of the pop-up dialog while in a red-alert mission.

Goober5000 (changes requested):
One message needs correcting

freespace2/freespace.cpp
```diff
@@ -3629,17 +3628,13 @@ void game_maybe_do_dead_popup(float frametime)
 				break;
 
 			// this should only happen during a red alert mission
-			case 3:				
+			case 3:			
+				Assertion(The_mission.flags[Mission::Mission_Flags::Red_alert], "FSO tried to reach replay the previous mission without being in a Red_alert mission.");
```
Goober5000, line 3632: Tried to reach replay?

Goober5000 (approved):

PR #2730, opened 2020-09-10
Fix declation of waypoint function so it can be used

JohnAFernandez:
Somehow this function declaration had the wrong arguments, and the function it declares was unreachable.  So this fixes the arguments so we can reach the function. (Especially because I want to use this in multi)


JohnAFernandez:
To clarify, there wasn't an overload, the declaration currently has no definition.

Goober5000 (approved):

PR #2734, opened 2020-09-11
Call send_ship before destroying it to avoid a crash.

### LINKED ISSUE #2732: FS1 COOP UP Doomsday crashes on 20.1.X builds
This mission is working fine on 19.0.0 builds but crashes on every 20.1.X builds.

I tried to pinpoint the reason but i couldt find the reason, all i can tell is: Is not related to red alert, removing red alert does nothing, and this only happens on multi setting, if i set the mission back to single player it works.

[coopsm2-10a.zip](https://github.com/scp-fs2open/fs2open.github.com/files/5204988/coopsm2-10a.zip)

JohnAFernandez:
Fixes #2732

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -15866,11 +15866,11 @@ void sexp_destroy_instantly(int n)
 		// if it's the player don't destroy
 		if (!get_player_from_ship_node(n))
 		{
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Aha!

PR #2735, opened 2020-09-11
Improve beam collision optimization

Baezon:
The `beam_get_cone_dot` function used for early collision culling will claim that non-slashers could possibly collide in a "very small" cone of 50 degrees. This is obviously absurdly conservative and, more problematically, I could also adversarially beat it and get erroneous early outs. This led me down a rabbit hole of profiling every step of the beam collision process and attempting multiple approaches to improve performance and most importantly, be actually correct in all cases. 

From this profiling I've learned that memory access is almost certainly a greater barrier to speed now as opposed to math, and the various functions calls and basic math trying their damndest to skirt around ever using a square root are no longer really relevant on modern architecture. Replacing the 'dot test' (checking the angle of the beam to the possible collider's position) and 'dist test' (check if the beam origin is within possible collider's radius) with a basic cylinder-sphere collision function takes about as much time (~700 nanoseconds), and improves the early culling percentage from ~65% to ~95%, with an overall 'per collision check' average from ~3500 ns to ~2000 ns. (tested with 6 beam-armed warships and a dozen fighters in the attached test mission)
[beamtest.txt](https://github.com/scp-fs2open/fs2open.github.com/files/5207541/beamtest.txt)

The actual `model_collide` function will use `fvi_segment_sphere` regardless, which is largely similar to the added `fvi_cylinder_sphere_may_collide` but excludes some unnecessary sphere intersection points and doesn't need to do all the set up of the collision info struct ahead of time (excluding the early out, reaching and concluding `fvi_segment_sphere` has already cost ~1400 ns).

Similar optimizations could probably be made all over the collision code, but... one step at a time.

Almost unrelatedly, but this was the original change that sent me down the rabbit hole, beam width never changes during its lifetime and the beam stores `beam_width` anyway (but only in the case its manually defined), so there's no reason for a function to recalculate it everytime its needed; calculate once at creation and then use that.

### DIFF
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -268,6 +268,29 @@ void beam_level_close()
 	list_init( &Beam_used_list );
 }
 
+// get the width of the widest section of the beam
+float beam_get_widest(beam* b)
+{
+	int idx;
+	float widest = -1.0f;
+
+	// sanity
+	Assert(b->weapon_info_index >= 0);
+	if (b->weapon_info_index < 0) {
+		return -1.0f;
+	}
+
+	// lookup
+	for (idx = 0; idx < Weapon_info[b->weapon_info_index].b_info.beam_num_sections; idx++) {
+		if (Weapon_info[b->weapon_info_index].b_info.sections[idx].width > widest) {
+			widest = Weapon_info[b->weapon_info_index].b_info.sections[idx].width;
+		}
+	}
+
+	// return	
+	return widest;
+}
+
 // fire a beam, returns nonzero on success. the innards of the code handle all the rest, foo
 int beam_fire(beam_fire_info *fire_info)
 {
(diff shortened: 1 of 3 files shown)


code/weapon/beam.cpp
```diff
@@ -392,6 +384,21 @@ int beam_fire(beam_fire_info *fire_info)
...
+	else {
+		// lookup
+		float widest = -1.0f;
+		for (int idx = 0; idx < Weapon_info[new_item->weapon_info_index].b_info.beam_num_sections; idx++) {
+			if (Weapon_info[new_item->weapon_info_index].b_info.sections[idx].width > widest) {
+				widest = Weapon_info[new_item->weapon_info_index].b_info.sections[idx].width;
+			}
+		}
+		new_item->beam_width = widest;
+	}
```
Goober5000, line 401: You may want to keep this in a function even if it's only used once.  It would help organize the code.

Goober5000 (commented):
Impressive work, pilot. :+1:

Goober5000 (approved):
LGTM

PR #2738, opened 2020-09-12
fix getPath() and add getName()

Goober5000:
The getPath() function would crash if the id was -1.  Also, there was no function for getting the file name.

### DIFF
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -109,14 +109,14 @@ static const char *Cfile_cdrom_dir = NULL;
 // Function prototypes for internally-called functions
 //
 static int cfget_cfile_block();
-static CFILE *cf_open_fill_cfblock(const char* source, int line, FILE * fp, int type);
-static CFILE *cf_open_packed_cfblock(const char* source, int line, FILE *fp, int type, size_t offset, size_t size);
-static CFILE *cf_open_memory_fill_cfblock(const char* source, int line, const void* data, size_t size, int dir_type);
+static CFILE *cf_open_fill_cfblock(const char* source, int line, const char* original_filename, FILE * fp, int type);
+static CFILE *cf_open_packed_cfblock(const char* source, int line, const char* original_filename, FILE *fp, int type, size_t offset, size_t size);
+static CFILE *cf_open_memory_fill_cfblock(const char* source, int line, const char* original_filename, const void* data, size_t size, int dir_type);
 
 #if defined _WIN32
-static CFILE *cf_open_mapped_fill_cfblock(const char* source, int line, HANDLE hFile, int type);
+static CFILE *cf_open_mapped_fill_cfblock(const char* source, int line, const char* original_filename, HANDLE hFile, int type);
 #elif defined SCP_UNIX
-static CFILE *cf_open_mapped_fill_cfblock(const char* source, int line, FILE *fp, int type);
+static CFILE *cf_open_mapped_fill_cfblock(const char* source, int line, const char* original_filename, FILE *fp, int type);
 #endif
 
 static void cf_chksum_long_init();
@@ -732,7 +732,7 @@ CFILE* _cfopen(const char* source, int line, const char* file_path, const char* 
 
 		FILE *fp = fopen(longname, happy_mode);
 		if (fp)	{
-			return cf_open_fill_cfblock(source, line, fp, dir_type);
+			return cf_open_fill_cfblock(source, line, file_path, fp, dir_type);
  		}
 		return NULL;
 	} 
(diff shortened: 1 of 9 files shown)


JohnAFernandez (approved):
LGTM

Always nice to see a good sanity check added.

Goober5000:
It turns out that getting the filename isn't as simple as I thought, so this will be a bit more work...

Goober5000:
Ok, that should do it.  Asking for Taylor's input since he's worked with cfile before.

notimaginative:
The code has changed enough that I'll need to take a little time to go over it. I'll try and go through it tonight.

code/cfile/cfile.cpp
```diff
@@ -732,7 +732,7 @@ CFILE* _cfopen(const char* source, int line, const char* file_path, const char*
 
 		FILE *fp = fopen(longname, happy_mode);
 		if (fp)	{
-			return cf_open_fill_cfblock(source, line, fp, dir_type);
+			return cf_open_fill_cfblock(source, line, file_path, fp, dir_type);
```
notimaginative, line 735: `file_path` can be a full path here rather than just the file name. There is an earlier check to see if it's a full path so you can just grab the filename then if possible, otherwise you would have to split it yourself.

code/cfile/cfile.cpp
```diff
@@ -969,6 +969,7 @@ static CFILE *cf_open_fill_cfblock(const char* source, int line, FILE *fp, int t
 		cfp->dir_type = type;
 		cfp->max_read_len = 0;
 
+		cfp->original_filename = original_filename;
```
notimaginative, line 972: `original_filename` points to a temporary location or a location that will be overwritten. Need to `vm_strdup()` this in all of the `cf_open_*()` functions and free it on `cfclose()` to avoid memory errors and/or invalid values when checking the value later.

notimaginative (changes requested):
Couple of issues but otherwise looks okay to me

notimaginative:
Not sure if it would be a cleaner solution or not, but you could probably just modify `CFileLocation` to also store the file name. If a file is indexed by the cfile system it has it's name stored in `cf_file->name_ext` and it not then you can just split it from the path manually. That way you've got the name sent from `cf_find_file_location()` and could just pass the `CFileLocation` struct itself to `cfopen_special()`.

All of the work would then stay inside of cfile rather than having to deal with it with every call to `cfopen_special()`

code/cfile/cfile.cpp
asarium: It would also be possible to have `original_filename` be an `SCP_string` since `CFILE` does not need to be trivially copyable.

Goober5000:
> Not sure if it would be a cleaner solution or not, but you could probably just modify `CFileLocation` to also store the file name

Correct me if I'm wrong but it looks like `CFileLocation` isn't referenced by the `CFILE` struct anywhere.  And there is no reference to `CFileLocation` in `cf_open_fill_cfblock` and related functions.

code/cfile/cfile.cpp
Goober5000: Oh good catch.

code/cfile/cfile.cpp
Goober5000: Ok.

notimaginative:
It's the return value of `cf_find_file_location*()` and values from it are what's passed to `cfopen_special()`. So you could add the filename to it and pass the whole struct to `cfopen_special()` instead of the individual parts. Otherwise your changes could remain basically the same. But rather than having to keep track of the original filename from the calling function it would all be in cfile code.

Instead of this:
`CFILE *test = cfopen_special(real_filename, res.full_name.c_str(), "rb", res.size, res.offset, res.data_ptr, dir_type);`

you would have this:
`CFILE *test = cfopen_special(res, "rb", dir_type);`

And `res` would have a filename field to reference for `original_filename` which you pass to `cf_open_*()`.

Goober5000:
> It's the return value of `cf_find_file_location*()` and values from it are what's passed to `cfopen_special()`. So you could add the filename to it and pass the whole struct to `cfopen_special()` instead of the individual parts.

Hmm, ok.  Will take a look at this.

Goober5000:
Ok I think I'd rather stick with the existing solution.  I'd have to piggyback the filename on the existing CFileLocation struct (modifying the struct as you said) and I don't think it's a good idea to do that, given that the struct is itself a return value from a different function.  Additionally I'm not super familiar with the cfile code in the first place, so I'd rather minimize my own changes to it.

notimaginative:
I meant that `cf_find_file_location()` has all of the information you need and can return the correct filename for you along with the rest of the info. At that point you can just send the that return value (the struct) to `cfopen_special()`.

If you like I can make up a quick patch to demonstrate what I mean and you could incorporate that in with your changes.

notimaginative (approved):

Goober5000:
> If you like I can make up a quick patch to demonstrate what I mean and you could incorporate that in with your changes.

Yes please do.

One of the obstacles is that not all the code paths went through `_cfopen_special`.  See `cf_open_mapped_fill_cfblock` for example; although there is a `res` variable in scope, that function would have to be modified to take it as a parameter.

notimaginative:
Not thoroughly tested, but this is what I was talking about. Then for your `cf_open_*()` calls you can just pass `res.name_ext.c_str()` in for `original_filename`. No need to make `cf_open_*()` take `CFileLocation` itself.

[cf_find_things.txt](https://github.com/scp-fs2open/fs2open.github.com/files/5215344/cf_find_things.txt)

Goober5000:
Oh, that looks sensible.  Will apply and update.

Goober5000:
Ok, patch applied, with some additional places I noticed that weren't in the patch.  Probably good if this gets another review.

code/cfile/cfile.cpp
```diff
@@ -1863,3 +1868,44 @@ int cfile_get_path_type(const SCP_string& dir)
...
+	separators = "/\\:";
+#endif
+
+	// find the last separator (if any) in the path sequence
+	ch = path_str - 1;
+	while ((ch = strpbrk(ch + 1, separators)) != nullptr) {
+		last_separator = ch;
+	}
+
+	return last_separator;
```
notimaginative, line 1889: You may could just do this instead (note the 2 r's, for reverse):

`return strrchr(ch, DIR_SEPARATOR_CHAR);`

which will return a ptr to the last directory separator, or NULL. No need to specify the separators then either

code/cfile/cfile.cpp
```diff
@@ -1863,3 +1868,44 @@ int cfile_get_path_type(const SCP_string& dir)
...
```
notimaginative, line 1910: Same thing here

notimaginative (changes requested):

code/cfile/cfile.cpp
Goober5000: I could, except that the cfile code explicitly specifies more than one possible separator char.

notimaginative: On windows it's looking for something like `c:\` or `c:/` to indicate a full path. In all other places in the code we specifically look for DIR_SEPARATOR_*, so I don't think changing this would break anything or else we should have already noticed.

May want to give it a test though. I don't have a windows box handy to confirm that it would be okay.

Goober5000: Ok, seems to work.

notimaginative (approved):

Issue #487, opened 2015-12-03
Missile turn rate is framerate dependent


asarium:
Higher framerates cause missles to turn slower (or low FPS causes missiles to turn faster). This can also be seen when increasing time compression, higher time compression causes the missiles to turn much sharper than when no time compression is used.

I guess that this is caused by using a wrong method of computing the next orientation and position of the missile that isn't stable for lower framerates.
This bug probably also affects ships but it isn't as noticeable because they don't turn as fast as missiles.

chief1983:
So is this like a direct relationship?  Like 30fps vs 60fps vs 120fps still look different, or do you have to get to the really low end of the spectrum to see it?

asarium:
As far as I can tell it's a direct relationship. You can even notice it when the framerate suddently gets lower or higher because the turn rate changes.

MageKing17:
Turn rate getting faster as framerate decreases makes me instinctively expect that frametime is being used wrong in the calculations somewhere. Is this a retail problem, or was it introduced by FSO?

asarium:
I'll try to reproduce it with retail.

chief1983:
Anyone ever reproduce this on a much older build?

JohnAFernandez:
So, I think I know what's happening here, although I think someone else should also take a look.

I took a look through the physics code to see if missiles were treated any differently than normal objects, and as far as I can tell, they're not.  Their rotational damp is set to zero, but that doesn't change the root cause of the bug in the end.  

Since ships have a damp factor, their rotation accelerates and the effect is not *as* pronounced.  You would see it only notice it for a ship if it is already rotating at or its maximum rotational velocity and you are actively looking for a ship to move weirdly.

Missiles, by virtue of being the example of rotational velocity we keep the most track of (in order to avoid) and having this issue be even more pronounced because of no damp factor, are just the first example we've discovered.  In the end, the real problem is fundamental to our physics engine. 

Here's the real problem: we simulate velocity for the whole frame at once, and then we simulate rotation for the whole frame at once.  This works for short frame times because it reasonably approximates real world physics, but once frame times become long, this is no longer true. So after moving a long distance in a long frame, it will be in that position and rotate a lot.  The rotational velocity is directly multiplied by frametime.

[...]

PR #2740, opened 2020-09-13
Remove turning framerate dependence for AI

Baezon:
Puts issue #487 into a well-deserved grave. I have a lot to say about this PR, even just for posterity reasons so I'm chunking it up into sections.

THE PROBLEM

First, an overview of the problem. In retail for a given AI controlled entity on a given frame, physics happens first, but nothing interesting happens here because the AI's influence hasn't been introduced yet, so we'll just sort of gloss over this first physics step. Later in the frame the AI is run (or `weapon_home` for missiles), it does its decision procedure and decides it wants to turn toward something. In order to do this, it runs `angular_move` which the AI feeds with its angular velocity and acceleration limits, and current angular velocity, and then proceeds to do the basic math -- for each axis it slows down or speeds up as necessary to arrive at its goal orientation with 0 rotvel as quickly as possible. Importantly, this function is capable of sub-frame maneuvers, if it _can_ accelerate, decelerate and arrive all within the frame it will do so, unlike velocity and position, where velocity is simply set, and then it waits until next frame to see how close it is and maybe speeds up or slows down.

An obvious but important note is that if the ship can't accelerate much, but is moving fast (perhaps it was collided with) the ship will attempt to slow down but otherwise report that it will mostly have turned due to the already existing rotvel.

Once finished the function returns with a new orientation and rotvel, _both of which it sets directly to the ship_, physically turning the ship. On the first frame, all of it's turning came from this AI function.

Next frame: physics comes first. Importantly none of those calculations changed `desired_rotvel`, which is always left at 0, and this is what `physics_sim_rot` works on. It will move the objects current rotvel towards its `desired_rotvel` exponentially, based on its the objects $rotdamp, more rotdamp = less movement towards desired_rotvel. When concluded this function _also_ updates the ship's rotvel and orientation, and since the `desired_rotvel` is always 0, this is always attempting to slow down the AI. If the AI is run again this frame `angular_move` will work on this new orientation, providing a _second_ updated orientation and overwriting the previous rotvel, resulting in (up to) double the apparent angular motion, but half the rotvel it should be given that motion.

THE SYMPTOMS

For a ships and missiles, the apparent results will diverge quite a bit based on that rotdamp. For a very slow capital ship with a rotdamp of 2, the physics doesn't end up slowing it down much at all, but _will_ make the orientation change, in effect doubling the movement that the AI is doing, since it's making basically the same orientation update when it goes as well. The result is that most capitals have a doubled turnrate or an actual rotation time of half of what is claimed in the table.

For a zippy fighter with a low rotdamp of 0.35, the physics reducing it's speed is quite effective. It's exponential so the 'physics speed damping' gets more effective the farther it is from 0, so the ship will sluggishly accelerate (`angular_move` is polynomial so the acceleration is constant so early on it will win) until it eventually reaches a point where the two balance out, and its reached its maximum effective rotvel. For a rotdamp less than about 0.5, this will always be slower than it's actual tabled rotation time, the lower the rotdamp the worse its maximum effective rotvel will be.

Missiles always have a rotdamp of 0, so the physics will always take it to 0 rotvel every frame. The only turning it gets is what it is able to accelerate later in the frame; none of that accumulated velocity will carry over into the next frame. 

For ships although they aren't turning at their tabled rates and low rotdamp makes them rotationally accelerate _slower_ rather than faster, ships aren't really affected by framerate. Both the physics and AI's effects are scaled in proportion to the framerate, so effect is largely a constant one.

However the physics effect on missiles _isn't_ scaled by frametime, it always goes to 0 whether it was turning at 0.01 or 100, but its acceleration from AI has no such help. Luckily for the missile, AI happens later so it an still get some turning done, but the turning is directly dependent on how long the frame is, the more time it has the more it can accelerate, before physics puts it back to 0, which is why missiles have such framerate dependent turnrate. Ships also suffer but in a constant way so it's not so bad.

THE SOLUTION

`angular_move` put forth a lot of work to calculate what the ship's orientation and rotvel should be and `physics_sim_rot` is relatively clueless and simply messing things up. So if a flag can be tripped then `physics_sim_rot` can basically skip what it's doing and take `angular_move` at its word. This can only be done if the AI happens first in the frame however, which is part of the changes (more specifically this change was made alone in PR #2613 ). I was initially very hesitant to make this change, but after extensive testing I've yet to find any AI breakages due to this. 

So now with the 'Framerate_independent_turning' flag, the AI will set the `ai_desired_orient` and `desired_rotvel` (for this purpose `desired_rotvel` is treated differently by physics; I decided to reuse an otherwise unused variable in the interest of space, but I'd be open to the argument that a new value would be preferable in the interest of clarity) values to the output from `angular_move` which `physics_sim_rot` then applies, so at least actual orientation and rotvel changes are still happening in `physics_sim`, even if they aren't being calculated there.

The old, table-accurate values being produced by `angular_move` aren't actually what was observed however, (as explained in 'THE SYMPTOMS') so a new function is used to take in the values, including rotdamp, and produce what was actually being observed, so retail behavior is preserved when it comes to ships. For missiles, "objectively true" turntimes aren't so clear, since it has always been framerate dependent. For now, it produces observed behavior at 60fps. There were discussions to expose this value in some way, including as game_settings option, but it's an imperfect solution since that is for the most part, determined by the mod, not the player. Framerate is player-specific, not mod-specific, so ideally it should be exposed to them, but I'm not sure what the best approach is here. Input would be appreciated.

THE IMPLICATIONS

This has the unsavory result of various physical aspects being duplicated, in physics and also fed to `angular_move`, (such as the immobile flag, and shockwaves) since they're both attempting to "simulate physics". This duplication is a fundamental aspect of this 'sub-frame' 'hands-on' approach to turning the AI, I don't know why Volition did AI turning this way, I haven't seen what turning looks like if its done the 'normal' way, but I suspect they did it for a reason, and I'm merely trying to properly implement what appeared to be their intentions (based on what little I've been able to piece together it appears they were aware of some of these issues, but unaware of why they were happening).

In this new scenario rotational maneuver overrides **will not function** while the AI is attempting to rotationally maneuver itself. Before it appears it would do a weird sort of 'halfsies' mix of both. I think this is okay because I'm pretty sure ship-maneuver recommends disabling the AI anyway during maneuvers.

CHANGES FROM #2613 'Ai before physics'

I've renamed the flag to "$AI use framerate independent turning" from "$Process AI before physics:" because I think it's a lot clearer to users as to what the flag actually does, although it is granted that it may no longer be quite as clear to coders why this framerate turning thing needs to evaluate AI at a different time, but I'm open to suggestions on this. I realize that naming flags is a bit of a faux pas, but it was introduced very recently and I'm pretty sure no one was using it.

Also this removes `cscrew_process_post` from the purview of the flag, since it really is a form of post-processing and needs to be in `obj_move_post`

Phew. Now tell me this isn't the biggest PR description for FSO.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -1138,16 +1138,52 @@ void ai_update_danger_weapon(int attacked_objnum, int weapon_objnum)
 	}
 }
 
+// Asteroth - figures out if we're turning in a frametime-independant way or not, and if so
+// ensures retail behavior is maintained (if the modder wants to).
+// vel and acc_limit should already be sent in with retail values
+void ai_get_vel_acc_limit_for_turning(vec3d* vel_limit, vec3d* acc_limit, float rotdamp, bool weapon) {
+
+	if (!Framerate_independent_turning)
+		return;
+	
+	if (weapon) {
+		// Missiles accelerated instantly technically, but this should do.
+		vm_vec_copy_scale(acc_limit, vel_limit, 1000.f);
+		// Approximately what you'd get at 60fps
+		*vel_limit *= 0.128f;
+		return;
+	}
+	else { // If we're a ship instead things get a bit hairier
+
+		for (int i = 0; i < 3; i++) {
+			float& vel = vel_limit->a1d[i];
+			float& acc = acc_limit->a1d[i];
+
+			float v1 = 2 * acc * rotdamp;
+			float v2 = 2 * vel;
+			if (v1 <= v2) {
+				vel = v1;
+			}
+			else {
+				vel = v2;
+				if (v1 == 0)
+					acc = 1000.f;
+				else
+					acc = acc * (2 - v2 / v1);
+			}
+		}
+	}
+}
+
 // If rvec != NULL, use it to match bank by calling vm_angular_move_matrix.
 // (rvec defaults to NULL)
 // optionally specified turnrate_mod contains multipliers for each component of the turnrate (RATE, so 0.5 = slower)
-void ai_turn_towards_vector(vec3d *dest, object *objp, vec3d *slide_vec, vec3d *rel_pos, float bank_override, int flags, vec3d *rvec, vec3d* turnrate_mod)
+void ai_turn_towards_vector(vec3d* dest, object* objp, vec3d* slide_vec, vec3d* rel_pos, float bank_override, int flags, vec3d* rvec, vec3d* turnrate_mod)
 {
 	matrix	curr_orient;
 	vec3d	vel_in, vel_out, desired_fvec, src;
 	float		delta_time;
 	physics_info	*pip;
-	vec3d	vel_limit, acc_limit;
 	float		delta_bank;
 
 	Assertion(objp->type == OBJ_SHIP || objp->type == OBJ_WEAPON, "ai_turn_towards_vector called on a non-ship non-weapon object!");
@@ -1170,6 +1206,7 @@ void ai_turn_towards_vector(vec3d *dest, object *objp, vec3d *slide_vec, vec3d *
 	curr_orient = objp->orient;
 	delta_time = flFrametime;
 
+	vec3d vel_limit, acc_limit;
 	vm_vec_zero(&vel_limit);
 
 	// get the turn rate if we have Use_axial_turnrate_differences
@@ -1218,6 +1255,9 @@ void ai_turn_towards_vector(vec3d *dest, object *objp, vec3d *slide_vec, vec3d *
 	acc_limit = vel_limit;
 	if (objp->type == OBJ_WEAPON)
 		vm_vec_scale(&acc_limit, 8.0f);
+
+	//maybe change vel and acc based on framerate_independant_turning
+	ai_get_vel_acc_limit_for_turning(&vel_limit, &acc_limit, pip->rotdamp, objp->type == OBJ_WEAPON);
 
 	// for formation flying
 	if ((flags & AITTV_SLOW_BANK_ACCEL)) {
@@ -1273,16 +1313,26 @@ vel_in = vel_in_copy;	//	HERE //-V587
 objp->orient = objp_orient_copy; //-V587
 #endif
 	
+	matrix out_orient;
+
 	if (rvec != NULL) {
-		matrix	out_orient, goal_orient;
+		matrix	goal_orient;
 
 		vm_vector_2_matrix(&goal_orient, &desired_fvec, NULL, rvec);
 		vm_angular_move_matrix(&goal_orient, &curr_orient, &vel_in, delta_time,
 			&out_orient, &vel_out, &vel_limit, &acc_limit, The_mission.ai_profile->flags[AI::Profile_Flags::No_turning_directional_bias]);
-		objp->orient = out_orient;
 	} else {
 		vm_angular_move_forward_vec(&desired_fvec, &curr_orient, &vel_in, delta_time, delta_bank,
-			&objp->orient, &vel_out, &vel_limit, &acc_limit, The_mission.ai_profile->flags[AI::Profile_Flags::No_turning_directional_bias]);
+			&out_orient, &vel_out, &vel_limit, &acc_limit, The_mission.ai_profile->flags[AI::Profile_Flags::No_turning_directional_bias]);
+	}
+	
+	if (Framerate_independent_turning) {
+		pip->ai_desired_orient = out_orient;
+		pip->desired_rotvel = vel_out;
+	}
+	else {
+		objp->orient = out_orient;
+		pip->rotvel = vel_out;
 	}
 
 	#ifndef NDEBUG
@@ -1291,7 +1341,7 @@ objp->orient = objp_orient_copy; //-V587
 			mprintf(("A ship rotated too far. Offending vessel is %s, please investigate.\n", Ships[objp->instance].ship_name));
 	}
 	#endif
-	pip->rotvel = vel_out;
+
 }
 
 //	Set aip->target_objnum to objnum
@@ -9103,7 +9153,7 @@ float dock_orient_and_approach(object *docker_objp, int docker_index, object *do
 	ship_info	*sip0, *sip1;
 	polymodel	*pm0, *pm1;
 	ai_info		*aip;
-	matrix		dom, nm;
+	matrix		dom, out_orient;
 	vec3d docker_p0, docker_p1, docker_p0_norm;
 	vec3d dockee_p0, dockee_p1, dockee_p0_norm;
 	vec3d docker_point, dockee_point;
@@ -9213,10 +9263,19 @@ float dock_orient_and_approach(object *docker_objp, int docker_index, object *do
 		if (sip0->flags[Ship::Info_Flags::Support])
 			vm_vec_scale(&acc_limit, 2.0f);
 
+		ai_get_vel_acc_limit_for_turning(&vel_limit, &acc_limit, docker_objp->phys_info.rotdamp, false);
+
 		// true at end of line prevent overshoot
-		vm_angular_move_matrix(&dom, &docker_objp->orient, &omega_in, flFrametime, &nm, &omega_out, &vel_limit, &acc_limit, false, true);
-		docker_objp->phys_info.rotvel = omega_out;
-		docker_objp->orient = nm;
+		vm_angular_move_matrix(&dom, &docker_objp->orient, &omega_in, flFrametime, &out_orient, &omega_out, &vel_limit, &acc_limit, false, true);
+
+		if (Framerate_independent_turning) {
+			docker_objp->phys_info.desired_rotvel = omega_out;
+			docker_objp->phys_info.ai_desired_orient = out_orient;
+		}
+		else {
+			docker_objp->phys_info.rotvel = omega_out;
+			docker_objp->orient = out_orient;
+		}
 
 		//	Translate towards goal and note distance to goal.
 		goal_point = &Path_points[aip->path_cur].pos;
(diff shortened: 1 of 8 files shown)


JohnAFernandez:
So, I'm reading this correctly the new behavior is completely hidden behind a flag.

Baezon:
Yes it is all behind a flag. I *do* think eventually, once enough trust can be put in this, it should become completely standard and not under a flag, but that's just my opinion.

And I'm not sure what you mean by 'mathematically verify'. I have an external mathematical model of the entire situation, which agrees very strongly with the observed behavior and has very good agreement with the implementation of the solution. I think, better than 'mathematical' verification, it can be experimentally verified. Futzing with the time compression has a dramatic effect on missiles normally, but now it... doesn't, they just act like everything else.

JohnAFernandez:
I meant something akin to unit testing.  But comparing the behavior to an external model satisfies any anxiety I had with it before.

Goober5000:
**OOOF**

Well, my hat is off to you for figuring all of this out.

code/weapon/weapons.cpp
```diff
@@ -4588,10 +4588,6 @@ void weapon_update_missiles(object* obj, float  frame_time) {
 		if (wp->swarm_index >= 0) {
 			swarm_update_direction(obj);
 		}
-
-		if (wp->cscrew_index >= 0) {
-			cscrew_process_post(obj);
```
JohnAFernandez, line 4593: How does is moving this function call out of `weapon_home` possible without affecting retail behavior and why not move `swarm_update_direction` as well?

code/physics/physics.cpp
```diff
@@ -162,47 +162,65 @@ void physics_sim_rot(matrix * orient, physics_info * pi, float sim_time )
...
 		rotdamp = pi->rotdamp;
 	}
 
-	// Do rotational physics with given damping
-	apply_physics( rotdamp, pi->desired_rotvel.xyz.x, pi->rotvel.xyz.x, sim_time, &new_vel.xyz.x, NULL );
-	apply_physics( rotdamp, pi->desired_rotvel.xyz.y, pi->rotvel.xyz.y, sim_time, &new_vel.xyz.y, NULL );
-	apply_physics( rotdamp, pi->desired_rotvel.xyz.z, pi->rotvel.xyz.z, sim_time, &new_vel.xyz.z, NULL );
-
-	Assert(is_valid_vec(&new_vel));
+	bool frit_ai_wants_to_move = Framerate_independent_turning && !IS_MAT_NULL(&pi->ai_desired_orient);
```
JohnAFernandez, line 183: What does "frit" in this variable stand for?  It's a long variable name, but I want to make sure that it's readable for everyone who needs to read this function.

code/ai/aicode.cpp
```diff
@@ -1257,41 +1304,29 @@ void ai_turn_towards_vector(vec3d *dest, object *objp, vec3d *slide_vec, vec3d *
 			delta_bank = -delta_bank;
 	}
 
-	//	Dave Andsager: The non-indented lines here are debug code to help you track down the problem in the physics
-	//	that is causing ships to inexplicably rotate very far.  If you see the message below in the log, set the next statement to be
-	//	the one marked "HERE".  (Do this clicking the cursor there, then right clicking.  Choose the right option.)
-	//	This will allow you to rerun vm_angular_move_forward_vec() with the values that caused the error.
-#ifndef NDEBUG
```
JohnAFernandez, line 1264: Yes, double checking everything, sorry.  This was not an issue during your work, right?

code/ai/aicode.cpp
```diff
@@ -1257,41 +1304,29 @@ void ai_turn_towards_vector(vec3d *dest, object *objp, vec3d *slide_vec, vec3d *
...
 		vm_angular_move_matrix(&goal_orient, &curr_orient, &vel_in, delta_time,
 			&out_orient, &vel_out, &vel_limit, &acc_limit, The_mission.ai_profile->flags[AI::Profile_Flags::No_turning_directional_bias]);
-		objp->orient = out_orient;
 	} else {
 		vm_angular_move_forward_vec(&desired_fvec, &curr_orient, &vel_in, delta_time, delta_bank,
-			&objp->orient, &vel_out, &vel_limit, &acc_limit, The_mission.ai_profile->flags[AI::Profile_Flags::No_turning_directional_bias]);
+			&out_orient, &vel_out, &vel_limit, &acc_limit, The_mission.ai_profile->flags[AI::Profile_Flags::No_turning_directional_bias]);
 	}
-
-	#ifndef NDEBUG
```
JohnAFernandez, line 1288: And am I right in thinking this is also part of that same debug code they had earlier?

code/ai/aicode.cpp
```diff
@@ -9210,10 +9245,20 @@ float dock_orient_and_approach(object *docker_objp, int docker_index, object *do
...
+		if(Framerate_independent_turning)
+			ai_compensate_for_retail_turning(&vel_limit, &acc_limit, docker_objp->phys_info.rotdamp, false);
+
 		// true at end of line prevent overshoot
-		vm_angular_move_matrix(&dom, &docker_objp->orient, &omega_in, flFrametime, &nm, &omega_out, &vel_limit, &acc_limit, false, true);
-		docker_objp->phys_info.rotvel = omega_out;
-		docker_objp->orient = nm;
+		vm_angular_move_matrix(&dom, &docker_objp->orient, &omega_in, flFrametime, &out_orient, &omega_out, &vel_limit, &acc_limit, false, true);
+
+		if (Framerate_independent_turning) {
```
JohnAFernandez, line 9254: You had an opportunity, too, to test docker behavior?

code/object/object.cpp
```diff
@@ -926,15 +927,13 @@ void obj_move_call_physics(object *objp, float frametime)
 			// then reset the flag and don't move the object.
             if (MULTIPLAYER_MASTER && (objp->flags[Object::Object_Flags::Just_updated])) {
 				objp->flags.remove(Object::Object_Flags::Just_updated);
-				goto obj_maybe_fire;
```
JohnAFernandez, line 929: Love that you were able to get rid of a `goto` statement.  Those are just plain confusing.

code/physics/physics.cpp
```diff
@@ -150,10 +150,10 @@ void physics_set_viewer( physics_info * p, int dir )
 
 void physics_sim_rot(matrix * orient, physics_info * pi, float sim_time )
 {
-	angles	tangles;
+	angles	tangles = { 0,0,0 };
 	vec3d	new_vel;
 	matrix	tmp;
-	float		shock_amplitude;
+	float		shock_amplitude = 0.0f;
```
JohnAFernandez, line 156: Since you are adjusting these lines, anyway, might as well just move the declaration and initialization over to `line 165` to match more modern coding practices of declaring as close to the variable use as possible.

code/physics/physics.cpp
```diff
@@ -150,10 +150,10 @@ void physics_set_viewer( physics_info * p, int dir )
 
 void physics_sim_rot(matrix * orient, physics_info * pi, float sim_time )
 {
-	angles	tangles;
+	angles	tangles = { 0,0,0 };
```
JohnAFernandez, line 153: We should have `vmd_zero_angles` for this. Recently added.

JohnAFernandez (changes requested):
I know you've done your homework, but a few pretty minor things stuck out to me.

code/weapon/weapons.cpp
Baezon: It's not terribly clear, but as I mentioned `cscrew_process_post` really is a form of post-processing, the missiles have already used `ai_turn_towards_vector` like regular missiles in `weapon_home` and the function basically just puts them in a 'fake' position to make it look like its corkscrewing when it really isn't. In order for corkscrew missiles to function properly they _must_ be in `obj_move_post`. This function is not related to AI or 'proper' turning at all. It was an oversight by me in #2613 to put them in `weapon_update_missiles`.

Unlike corkscrew missiles, swarm missiles will not use the `ai_turn_towards_vector` in `weapon_home`, `swarm_update_direction` uses it in a different way , and is _not_ a form of post-processing.

code/physics/physics.cpp
Baezon: `Framerate_independent_turning `. I was hoping that would be more obvious, but I wanted to make sure in some way explicitly that the flag is bundled into the boolean, _as well_ as the ai 'wanting to move'. I'm not sure what a better name would be without making it cumbersomely long.

JohnAFernandez: I think it's a fairly good name, maybe just add a comment then.

code/ai/aicode.cpp
Baezon: Ah! I had meant to comment on this change. I have never seen the mprintf later in the function that this refers to, and I'm almost certain this was used for debugging something during development. I don't mind leaving it in just in case, but after working so much on this function it really becomes a readability eyesore for me.

JohnAFernandez: I honestly don't think it's a big deal.  I don't like these old debug statements or things that tell you to "ask Allender" and things.  But someone above my pay grade may say something.  No one has taken a deeper look at this than you, so I imagine you would have found the problem if it still existed.

code/ai/aicode.cpp
Baezon: I didn't test it explicitly, although I've rearmed myself quite a few times during my testing. I'll do some more explicit tests to be sure.

Baezon: Yep, everything seems to be in order.

JohnAFernandez (approved):
Ok.  My changes are done.  Good job on everything.  I think we should wait one one other reviewer to at least take a look and give it a thumbs up.

Goober5000:
I'll take a look.  Not today though.

Goober5000:
Well, I'm pretty much unable to follow the code itself.  I just have a few questions:

1) Have you sanity-checked this by flying a variety of missions from a variety of mods, to see if you notice any difference?

2) I would ask about doing another numerical analysis to prove that the new code is close to the old code, but if the effect is dependent on frame rate then there *is* no standard, unless we assume a FPS of, say, 60.  Is the new turn behavior "close enough" to the old behavior?

3) Can you link this github ticket somewhere in the code so that other coders can read your dissertation if they need to?

4) Provided AI is disabled as you said, do the ship-maneuver sexps have the same effect they used to?

5) Some of the maneuver flags (e.g. turn without banking) have been copied as ships.tbl ship class flags, in which case they would be used by AI ships operating under AI control, even if no ship maneuver is taking place.  Do the flags themselves still work?

Baezon:
I'll be able to respond to your other points soon, but I'm a little disconcerted you weren't able to follow the code changes. While tracking it all down and laying it out was a bit complex it should be fairly straightforward once its clear what the problem is. Was there any part in particular that gave you trouble?

Goober5000:
It's more that there is no intuitive connection between the code and the high-level concept.  But I'm not really bothered by that, as I'm not too familiar with the physics code and I'm also only seeing the diffs, not the whole file in context.

wookieejedi:
I'll chime in and add that I've been testing this branch/flag enabled with Fate of the Galaxy and haven't ran into any noticeable changes in how ships move.

Baezon:
1. I have thoroughly played retail with it, but not much modded. I'll try some of the more 'out there' ones to be sure.
2. As far as ships go the numbers beings produced are spot on as far as my predictions go, and it feels identical. Missiles, its hard to say how much is just the placebo effect, but the effect *is* very small if anything. Tornados don't seem *quite* as dead on as I remember them, but I might just be used to playing on framerates less than 60, and disabling the flag seems to have the same behavior for around 60fps.
3. That's a good idea, there's a few places where I'll do that.
4. Yes, while playing dead, all the maneuvers work properly.
5. The only flag for turning, `Dont_bank_when_turning`, is implemented into `ai_turn_towards_vector` so there are no issues taking it into account.

Baezon:
JAD:XA seems perfectly functioning, as well as BP:WiH (The Blade Itself, as well. Plenty of missiles there :) ). And apparently FotG seems good as well. Everything seems to be working as intended.

Goober5000:
That's a reasonable selection of a variety of different mods. :+1:  Coupled with reports from @wookieejedi and @JohnAFernandez I'd say this PR is good to go.

PR #2743, opened 2020-09-14
don't break the parser in Unicode mode

Goober5000:
If a mod is running in Unicode mode, ignore Special Character Index if the table happens to supply it.


JohnAFernandez (approved):
LGTM

I should have added this previously.

MageKing17 (approved):

PR #2744, opened 2020-09-15
More cleanup and Int3() refactors

JohnAFernandez:
So `ai_attack_object()` had a parameter that was intended for a future or canceled feature by :V:.  This removes the parameter without altering the logic.  -- Not extremely attached to this change

Also cleans up a bunch of `Int3()`'s in `aicode.cpp`


code/ai/aicode.cpp
```diff
@@ -1893,8 +1894,14 @@ float get_wing_lowest_av_ab_speed(object *objp)
...
+	Assertion(((*ignore_objnum <= UNUSED_OBJNUM) || (*ignore_objnum >= 0)), "Unexpected value for ignore_objnum %d. This is a coder error, please report.", *ignore_objnum); 
+	if ((*ignore_objnum > UNUSED_OBJNUM) && (*ignore_objnum < 0)) {
+		return 0;
+	}
+
 	// Not ignoring anything.
-	if (*ignore_objnum == UNUSED_OBJNUM)
+	if (*ignore_objnum <= UNUSED_OBJNUM) // includes
 	{
 		return 0;
```
MageKing17, line 1906: It seems to me that it would be simpler to just have one conditional here:
```cpp
if (*ignore_objnum < 0)
{
	return 0;
}

File: code/ai/aicode.cpp [hunk 2]
```diff
@@ -2316,32 +2317,33 @@ void force_avoid_player_check(object *objp, ai_info *aip)
...
 	int temp;
-	ai_info	*aip;
+	ai_info* aip;
 
-	Assert(attacker != NULL);
+	Assert(attacker != nullptr);
 	Assert(attacker->instance != -1);
 	Assert(Ships[attacker->instance].ai_index != -1);
+	//	Bogus!  Who tried to get me to attack myself!
+	Assertion(attacker != attacked, "%s was told to attack itself! This may be an error in the mission file.  If that checks out, please report to a coder!", Ships[attacker->instance].ship_name);
```
MageKing17, line 2329: If an error in the mission file can cause this, then it shouldn't be an `Assertion()`; bad data should never trip an assertion. Given that recovery is easy from here (it recovers quite easily by just returning), it should be a `Warning()` instead.

MageKing17 (changes requested):
Just a couple of minor issues.

code/ai/aicode.cpp
```diff
@@ -235,7 +235,7 @@ const char *Skill_level_names(int level, int translate)
 			str = XSTR("Insane", 473);
 			break;
 		default:	
-			Int3();
+			Error(LOCATION, "An invalid difficulty level of %d was passed to Skill_level_names().", level);
```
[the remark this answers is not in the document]
JohnAFernandez: Ok.  I did not know about that.  I will probably use that a lot more now.

code/ai/aicode.cpp
```diff
@@ -1893,13 +1894,19 @@ float get_wing_lowest_av_ab_speed(object *objp)
  */
 int is_ignore_object_sub(int *ignore_objnum, int *ignore_signature, int objnum)
 {
+	// Should never happen.  I thought I removed this behavior! -- MK, 5/17/98
+	Assertion(((*ignore_objnum <= UNUSED_OBJNUM) || (*ignore_objnum >= 0)), "Unexpected value for ignore_objnum %d. This is a coder error, please report.", *ignore_objnum); 
+	if ((*ignore_objnum > UNUSED_OBJNUM) && (*ignore_objnum < 0)) {
+		return 0;
+	}
+
```
MageKing17, line 1902: This conditional is redundant with the one immediately below.

MageKing17 (changes requested):

MageKing17 (approved):
Nothing looks wrong to me.

JohnAFernandez:
Thanks!

PR #2747, opened 2020-09-16
Add Scripting Hook `On Countermeasure Fire`

wookieejedi:
This PR adds a scripting hook which triggers when a ship fires a countermeasure. It has two variables, the ship firing the countermeasure and the number of countermeasures remaining. I have tested and it works as expected. This new hook complements the preexistsing `On Secondary Fire` and `On Primary Fire` hooks.


asarium (approved):

Goober5000 (changes requested):
Clarification: is this the number of countermeasures *after* firing this one?  Make that clear in the documentation.

Also, add a hook parameter for the countermeasure object itself.

wookieejedi:
Thanks, good idea. Just updated.

Goober5000 (approved):

PR #2748, opened 2020-09-17
Fix an ai check

JohnAFernandez:
Keep multi from crashing from an out of bound check when trying to set up a formation for a respawning ship.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -11457,8 +11457,10 @@ void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wingnum, 
 
 	if (wing_index == 0)
 		wing_delta = vmd_zero_vector;
-	else if ( formation_index != -1 && wing_index < MAX_SHIPS_PER_WING) 
+	else if (!Wing_formations.empty() && (formation_index >= 0) && (wing_index < MAX_SHIPS_PER_WING)) {
+		Assertion((int)Wing_formations.size() > formation_index, "%s wing's formation_index %d is higher than the Wing_formations size.  Something went very wrong, go tell a coder!", Wings[wingnum].name, formation_index);
 		wing_delta = Wing_formations[formation_index].positions[wing_index - 1]; //  custom desired location in leader's reference frame
+	}
 	else 
 		get_wing_delta(&wing_delta, wing_index);		//	default desired location in leader's reference frame
 	


Goober5000 (approved):

Goober5000:
What exactly was the value of `formation_index` here?

JohnAFernandez:
I can't remember, the problem was actually that `Wing_formations.size()` was zero.

Goober5000 (changes requested):
Oh.  Then this is actually the wrong fix.  The correct fix is to figure out why `Wing_formations` is zero here.

code/ai/aicode.cpp
```diff
@@ -11457,7 +11457,7 @@ void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wingnum,
 
 	if (wing_index == 0)
 		wing_delta = vmd_zero_vector;
-	else if ( formation_index != -1 && wing_index < MAX_SHIPS_PER_WING) 
+	else if ((formation_index >= 0) && (formation_index < (int)Wing_formations.size()) && (wing_index < MAX_SHIPS_PER_WING))
```
Goober5000, line 11460: else if (!Wing_formations.empty() && (formation_index >= 0) && (wing_index < MAX_SHIPS_PER_WING))

Goober5000: The github suggestions feature isn't working, so use that line instead.

Turns out it's okay for `Wing_formations` to be empty, but it's better to explicitly check for that.

JohnAFernandez: If we switch to that, would there be a reason to add an assert to check `formation_index < Wing_formations.size()` ?

Goober5000: Yeah, an Assert is reasonable for that.  It shouldn't ever happen, but it's good to safeguard against it.

code/ai/aicode.cpp
```diff
@@ -11457,8 +11457,10 @@ void get_absolute_wing_pos(vec3d *result_pos, object *leader_objp, int wingnum,
 
 	if (wing_index == 0)
 		wing_delta = vmd_zero_vector;
-	else if ( formation_index != -1 && wing_index < MAX_SHIPS_PER_WING) 
+	else if (!Wing_formations.empty() && (formation_index >= 0) && (wing_index < MAX_SHIPS_PER_WING)) {
+		Assertion(Wing_formations.size() < formation_index, "%s wing's formation_index %d is higher than the Wing_formations size.  Something went very wrong, go tell a coder!", Wings[wingnum].name, formation_index);
```
Goober5000, line 11461: Er, the inequality check is backwards.  Also CI is complaining about signed vs. unsigned.
```suggestion
		Assertion(formation_index < (int)Wing_formations.size(), "%s wing's formation_index %d is higher than the Wing_formations size.  Something went very wrong, go tell a coder!", Wings[wingnum].name, formation_index);
```

JohnAFernandez: Should be fixed now.

About half the time I'm able to code I'm literally falling asleep while doing it.

Goober5000 (approved):

PR #2749, opened 2020-09-17
Fix special Char crash

JohnAFernandez:
When I wrote the original change to a vector instead of an array, I was working on the assumption that this section of code was parsed after other code that had already made sure that the special_char_indexes vector was at least three entries.  Further testing has revealed this is not always true, and this fix keeps it from crashing.


Goober5000 (approved):
Looks good. :+1:

PR #2752, opened 2020-09-20
Object update packet upgrade

JohnAFernandez:
All right, the second big change.  This basically reworks nearly every part of the object update packet.  I may not have touched the support ship section, but that may be it.

**Things that need to be finished before merging:**
~~1. Need better handling for AI because their movement is literally in another physics system as discovered by @Baezon .~~ Decided that this isn't strictly needed.  The "issue" is that there is minor forward and back jittering of only AI controlled craft. But it's so small that accuracy is not affected.  I will just spend the time writing the lua packet, instead.
~~2. Bug on standalones.  Crash occurs if waypoints are in use.  Cause not yet discovered.~~
~~3. Need to abort adding information to the packet and fix flags once the packet is filled.~~ 

**This upgrade:**
1. Keeps the object update packet aligned, even if the ship is dying or has not yet arrived on the client.
2. Properly handles model point shields (yes, that's what was causing multi to crash the most!)
3. Reworks positional interpolation timing and adds rotational interpolation.
 -- note timing information is now sent with every object update packet.
4. Saves a lot of bandwidth by only sending data that is changing. (no more sending all subsystems once per second!)
5. Better handles out of order packets (but can still use their information if recent enough!)
6. Builds the basis for Rollback by beginning to record the positions and orientations and velocities of ships for the last 30 frames.
7. Adds subsystem rotation -- Yeah, you saw that right, subsystems will now spin in multi.
8. Waypoint movement is now enabled via sexp, and time to arrival of waypoints is updated on the target info gauge.

**Things to keep in mind:**
I was pretty inexperienced when I started this, even as I made progress and fixed bugs, I reworked the structs that keep track of this a lot, trying to make them as understandable as possible.  But I'm open to suggestions, even though a lot of work has already been done.

**Here's an edited version of what I sent to Taylor to help explain in more detail:**

So the object update upgrade separates the Hull and Shields section, and adds a timestamp section as a header to the whole packet, not each individual ship's update. It also upgrades the rest of the packet sections.  `Seq_num` now an int, is also moved to that header and has been changed from an arbitrary number to the actual frame on the other computer as tracked by the Oo_info struct.  This allows FSO to figure out more certainly (but still approximately) how much time has elapsed on another computer, improving interpolation and adding a basis for the rollback feature.

All of these sections now check to see if the information has been sent already before sending it, and the receiving side will filter out that section's information if a newer packet already provided information from that section by storing that same sequence number when it uses new information.

The Hull section is basically the same, the shields section is basically the same, except that this section is skipped if attempting to bash the info would cause a crash (the ship is already dying, departing or possibly just arriving).  These crashes were introduced with the model point shields feature that made the number of shield sections dynamic.

Didn't change the afterburner flag.

Subsystem section has a big change. It was taking up a tremendous amount of space sending all of the subsystems, especially for larger ships, as it sent them once a second.  It keeps track of them in the Oo_info struct. Again, this section has to be skipped if the ship is arriving departing or dying. It is put through a compression function which does list every subsystem with a "does/does not" have updated info flag, but each subsystem that does not have updated info only takes up one bit.  Also if no subsystems have updated info, it will not send any subsystem info.

The big, big changes are with the position and orientation.  It now just sends everything whenever the ship is moving. Pos, velocity, desired velocity, orientation, rotational velocity, desired rotational velocity.

_Desired rotational velocity is *not* sent for ai._

All of the information is compressed in some way. If they were compressed before, their compression was more than likely adjusted. The biggest bandwidth saver here is sending angles for the orientation instead of the whole matrix (or at least part of it).  The angles are just derived from the orientation matrix before and the converted back to a matrix after.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -2763,7 +2763,7 @@ void vm_interpolate_angles_quick(angles *dest0, angles *src0, angles *src1, floa
 	  // pitch
 	  // if start and end are basically the same, assume we can basically jump to the end.
 	if ( (fabs(arc_measures.p) < 0.00001f) ) {
-			arc_measures.p = 0.0f;
+		arc_measures.p = 0.0f;
 	} // Test for positive difference
 	else if (arc_measures.p > 0.0f)  {
 
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -2744,3 +2744,94 @@ void vm_match_bank(vec3d* out_rvec, const vec3d* goal_fvec, const matrix* match_
 	vm_vec_rotate(&temp, &match_orient->vec.rvec, &source_frame);
 	vm_vec_unrotate(out_rvec, &temp, &dest_frame);
 }
+
+// Cyborg17 - Rotational interpolation between two angle structs in radians.  Assumes that the rotation direction is the smaller arc difference.
+// src0 is the starting angle struct, src1 is the ending angle struct, interp_perc must be a float between 0.0f and 1.0f inclusive.
+// rot_vel is only used to determine the rotation direction. This functions assumes a <= 2PI rotation in any axis.  
+// You will get inaccurate results otherwise.
+void vm_interpolate_angles_quick(angles *dest0, angles *src0, angles *src1, float interp_perc) {
+	
+	Assertion((interp_perc >= 0.0f) && (interp_perc <= 1.0f), "Interpolation percentage, %f, sent to vm_interpolate_angles is invalid. The valid range is [0,1], go find a coder!", interp_perc);
+
+
+	angles arc_measures;
+
+	arc_measures.p = src1->p - src0->p;	
+	arc_measures.h = src1->h - src0->h;	
+	arc_measures.b = src1->b - src0->b;
+
+	  // pitch
+	  // if start and end are basically the same, assume we can basically jump to the end.
+	if ( (fabs(arc_measures.p) < 0.00001f) ) {
+			arc_measures.p = 0.0f;
+	} // Test for positive difference
+	else if (arc_measures.p > 0.0f)  {
+
+		// do we actually need to go in the other direction
+		if (arc_measures.p > (PI*1.5f)) {
+			arc_measures.p = PI2 - arc_measures.p;
+
+		}
+
+	} // Negative difference then.
+	else {
+
+		// do we actually need to go in the other direction
+		if (arc_measures.p < -PI_2) {
+			arc_measures.p = -PI2 - arc_measures.p;
+
+		}
+	}
+
+	// heading
+	// if start and end are basically the same, assume we can basically jump to the end.
+	if ( (fabs(arc_measures.h) < 0.00001f) ) {
+		arc_measures.h = 0.0f;
+	} // Test for positive difference
+	else if (arc_measures.h > 0.0f)  {
+
+		// do we actually need to go in the other direction
+		if (arc_measures.h > (PI*1.5f)) {
+			arc_measures.h = PI2 - arc_measures.h;
+
+		}
+
+	} // Negative difference then.
+	else {
+
+		// do we actually need to go in the other direction
+		if (arc_measures.h < -PI_2) {
+			arc_measures.h = -PI2 - arc_measures.h;
+
+		}
+	}
+
+	// bank
+	// if start and end are basically the same, assume we can basically jump to the end.
+	if ( (fabs(arc_measures.b) < 0.00001f) ) {
+		arc_measures.b = 0.0f;
+	} // Test for positive difference
+	else if (arc_measures.b > 0.0f)  {
+
+		// do we actually need to go in the other direction
+		if (arc_measures.b > (PI*1.5f)) {
+			arc_measures.b = PI2 - arc_measures.b;
+
+		}
+
+	} // Negative difference then.
+	else {
+
+		// do we actually need to go in the other direction
+		if (arc_measures.b < -PI_2) {
+			arc_measures.b = -PI2 - arc_measures.b;
+
+		}
+	}
+
+	// Now just multiply the difference in angles by the given percentage, and then subtract it from the destination angles.
+	// If arc_measures is 0.0f, then we are basically bashing to the ending orientation without worrying about the inbetween.
+	dest0->p = src0->p + (arc_measures.p * interp_perc);
+	dest0->h = src0->h + (arc_measures.h * interp_perc);
+	dest0->b = src0->b + (arc_measures.b * interp_perc);
+}
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -2753,7 +2753,6 @@ void vm_interpolate_angles_quick(angles *dest0, angles *src0, angles *src1, floa
 	
 	Assertion((interp_perc >= 0.0f) && (interp_perc <= 1.0f), "Interpolation percentage, %f, sent to vm_interpolate_angles is invalid. The valid range is [0,1], go find a coder!", interp_perc);
 
-
 	angles arc_measures;
 
 	arc_measures.p = src1->p - src0->p;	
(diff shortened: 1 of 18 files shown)


JohnAFernandez:
Travis is showing a false positive, because where it is asking for a range based loop, one is not possible.  However, I suspect that is one of the places that will likely have changes requested.

Goober5000:
You should be able to use `// NOLINT` on the offending line(s) to disable the CI checks.

code/math/vecmat.cpp
```diff
@@ -2744,3 +2744,94 @@ void vm_match_bank(vec3d* out_rvec, const vec3d* goal_fvec, const matrix* match_
...
+	angles arc_measures;
+
+	arc_measures.p = src1->p - src0->p;	
+	arc_measures.h = src1->h - src0->h;	
+	arc_measures.b = src1->b - src0->b;
+
+	  // pitch
+	  // if start and end are basically the same, assume we can basically jump to the end.
+	if ( (fabs(arc_measures.p) < 0.00001f) ) {
+			arc_measures.p = 0.0f;
```
notimaginative, line 2766: Too much indentation here

code/math/vecmat.cpp
```diff
@@ -2744,3 +2744,94 @@ void vm_match_bank(vec3d* out_rvec, const vec3d* goal_fvec, const matrix* match_
...
+
+	} // Negative difference then.
+	else {
+
+		// do we actually need to go in the other direction
+		if (arc_measures.p < -PI_2) {
+			arc_measures.p = -PI2 - arc_measures.p;
+
+		}
+	}
```
notimaginative, line 2784: This is an unnecessary amount if-else's. You really only need to do the > PI check and then the < PI check

code/math/vecmat.cpp
```diff
@@ -2744,3 +2744,94 @@ void vm_match_bank(vec3d* out_rvec, const vec3d* goal_fvec, const matrix* match_
...
+
+	} // Negative difference then.
+	else {
+
+		// do we actually need to go in the other direction
+		if (arc_measures.h < -PI_2) {
+			arc_measures.h = -PI2 - arc_measures.h;
+
+		}
+	}
```
notimaginative, line 2807: Same here

code/math/vecmat.cpp
```diff
@@ -2744,3 +2744,94 @@ void vm_match_bank(vec3d* out_rvec, const vec3d* goal_fvec, const matrix* match_
...
+
+	} // Negative difference then.
+	else {
+
+		// do we actually need to go in the other direction
+		if (arc_measures.b < -PI_2) {
+			arc_measures.b = -PI2 - arc_measures.b;
+
+		}
+	}
```
notimaginative, line 2830: And here

code/network/multi.cpp
```diff
@@ -1228,10 +1229,10 @@ void multi_do_frame()
 				}
 			} else if ( !(Player_ship->is_departing() ) ){
 				// if the rate limiting system says its ok
-				if(multi_oo_cirate_can_send()){
+//				if(multi_oo_cirate_can_send()){
 					// use the new method
-					multi_oo_send_control_info();
-				}				
+					multi_oo_send_control_info(); // Cyborg17 - Don't limit clients updating themselves.
+//				}
```
notimaginative, line 1235: Is this really necessary? It basically makes object update rate limits irrelevant. And I'm a bit worried about what that will do to bandwidth on slower connections

code/network/multi_obj.h
```diff
@@ -32,19 +35,80 @@ struct net_player;
 #define OOC_AFTERBURNER_ON			(1<<5)
 // two spots now left for more OOC flags
 
+// what we are setting the subsystem "index" to if there was no subsystem.
+#define OOC_INDEX_NULLPTR_SUBSYSEM 255
```
notimaginative, line 39: If the subsystem count ever gets bumped this could be a problem. May want to look for a different solution here (use short type to allow for -1?)

code/network/multi_obj.h
```diff
@@ -32,19 +35,80 @@ struct net_player;
...
+
+// Cyborg17, Server will be tracking only the last second of frames
+#define MAX_FRAMES_RECORDED		30
+#define PRIMARY_PACKET_CUTOFF			2000
+
 // and two special values to help with multilock
 #define OOC_INDEX_NULLPTR_SUBSYSEM 255
```
notimaginative, line 46: Unless my eyes are messed up this is just the same #define as above. Surprised a compiler warning didn't get triggered here

freespace2/freespace.cpp
```diff
@@ -947,6 +947,7 @@ void game_level_init()
 	NavSystem_Init();				// zero out the nav system
 
 	ai_level_init();				//	Call this before ship_init() because it reads ai.tbl.
+	multi_init_oo_and_ship_tracker();	// Inits/resets multiplayer ship tracking system.  Has to be done before creating any ships.
```
notimaginative, line 950: There needs to be a multi check here. Might be best to put it in the function as I believe it's called from somewhere else too

code/network/multiutil.cpp
```diff
@@ -210,13 +223,16 @@ void multi_set_network_signature( ushort signature, int what_kind )
 	} else if ( what_kind == MULTI_SIG_ASTEROID ) {
 		Assert( (signature >= ASTEROID_SIG_MIN) && (signature <= ASTEROID_SIG_MAX) );
 		Next_asteroid_signature = signature;
+	} else if ( what_kind == MULTI_SIG_WAYPOINT ) {
+		Assert( (signature >= WAYPOINT_SIG_MIN) && (signature <= WAYPOINT_SIG_MAX) );
```
notimaginative, line 227: The signature isn't set here. Is that intentional?

code/network/multiutil.cpp
```diff
@@ -3771,81 +3591,398 @@ int multi_pack_unpack_rotvel( int write, ubyte *data, matrix * /*orient*/, vec3d
 	}
 }
 
-// Packs/unpacks desired rotvel
-// Returns number of bytes read or written.
-int multi_pack_unpack_desired_rotvel( int write, ubyte *data, matrix * /*orient*/, vec3d * /*pos*/, physics_info *pi, ship_info *sip)
+ubyte multi_pack_unpack_desired_vel_and_desired_rotvel( int write, bool full_physics, ubyte *data, physics_info *pi, vec3d* local_desired_vel)
```
notimaginative, line 3594: Another ubyte function that shouldn't be

code/network/multiutil.cpp
```diff
@@ -3771,81 +3591,398 @@ int multi_pack_unpack_rotvel( int write, ubyte *data, matrix * /*orient*/, vec3d
...
+				else {
+					a = SUBSYSTEM_PACKER_TRUE;
+					bitbuffer_put( &buf, (uint)a,1);
+				}
+			}
+
+			// reset skipping rotation flag
+			done = false;
+		}
+		return (ubyte)bitbuffer_write_flush(&buf);
```
notimaginative, line 3899: Shouldn't cast the return value here

code/network/multiutil.cpp
```diff
@@ -3771,81 +3591,398 @@ int multi_pack_unpack_rotvel( int write, ubyte *data, matrix * /*orient*/, vec3d
...
+				} 
+
+				flags->push_back(0);
+				current_subsystem_index++;
+			}
+
+			a = bitbuffer_get_unsigned(&buf, 1);
+		} while (a == SUBSYSTEM_PACKER_TRUE);
+		
+		return (ubyte)bitbuffer_read_flush(&buf);
```
notimaginative, line 3985: Shouldn't cast this either

notimaginative (changes requested):

code/network/multi_obj.h
JohnAFernandez: One came in from the multilock changes, and one accidentally from this patch since it should be already defined.

code/network/multi_obj.h
JohnAFernandez: It was bumped to a ushort, but I forgot to make the change to this constant to allow for that.  I can just use the ushort max?

notimaginative: If you used a short then you should be able to assign it without having to check for this special value. And SHRT_MAX is still a bunch of subsystems.

It works either way though, so whichever you think works better in context should be fine.

code/network/multi_obj.h
notimaginative: I figured that would be the case. And a reminder to update the multilock packets to use short/ushort too.

code/math/vecmat.cpp
JohnAFernandez: Ok, I'll give that a try and run a test.  I was having trouble with this function previously, but I don't think I tried that.

code/network/multi.cpp
JohnAFernandez: It makes it irrelevant for clients, yes.  I'll run some tests without it and check that it doesn't cause any slowdowns.

freespace2/freespace.cpp
JohnAFernandez: Yes, it's called once for the standalone code path and once for the regular code path.

code/network/multiutil.cpp
JohnAFernandez: That is a big error I made!  Looks like that allows my test case of *one* waypoint to work but would fail in other cases.

code/network/multi_obj.h
JohnAFernandez: I mean, we have no idea what crazy things modders will do in the future, so future-proofing as much as possible is my goal.  Some of the things that Oddgrim and others are doing make me worried about how to fully support models of the future in multi.

JohnAFernandez:
Ok, first round of feedback should be dealt with

code/network/multi_obj.h
Goober5000: What about doing this:
```
#define OOC_INDEX_NULLPTR_SUBSYSEM static_cast<ushort>(-1);
```
Casting -1 to an unsigned type automatically makes it the highest value of that type, I think.  If the type is ever changed to unsigned int, that method should still work.

code/math/vecmat.cpp
JohnAFernandez: Ok. I forgot to make these changes and then test them.

code/network/multi.cpp
notimaginative: It's not just the clients though, it greatly increases the bandwidth requirements for the server/standalone too. I'd feel a lot better about this with some hard data to back it up: 1) that it's actually necessary for the feature to work properly, and 2) that the bandwidth requirements for both server and client client are reasonable on a 120 FPS client with an upstream rate of no more than 512kbps.

JohnAFernandez:
After the next test, I'll put this in ready for review.

code/network/multi.cpp
JohnAFernandez: Testing after restoring, made only a slight decrease in interpolation accuracy.  I'll remove the change.

notimaginative (approved):

notimaginative:
Nothing else stands out to me. I can't verify the math, but otherwise it looks okay

PR #2753, opened 2020-09-21
Fix memory error which broke reliable sockets

notimaginative:
This was apparently broken the entire time in the IPv6 code and we all missed it.


JohnAFernandez:
I think this just highlights how insidious multi bugs can be.  Testers played through multiple missions with seemingly no issues, but I guess they were getting lucky and it wasn't affecting any of the more obvious packets.

JohnAFernandez (approved):
I understand this one well enough to approve it.

Good thing it wasn't any more complex!

PR #2754, opened 2020-09-21
Fix #2724

### LINKED ISSUE #2724: Weapons can fire without lock
Download the following mission.  Press 1 to make the Orion hostile.  Fire a Cyclops at it -- note that you will need to acquire lock.  Then wait for the countdown timer to finish, turn the ship so that aspect lock is lost, and fire another Cyclops.  You'll see that it can fire without a lock.  Then wait for the countdown timer one more time and fire a third Cyclops.  This will require a lock again.

[hangartest.txt](https://github.com/scp-fs2open/fs2open.github.com/files/5192752/hangartest.txt)

TRBlount:
This fixes #2724


JohnAFernandez (approved):
This actually makes perfect sense.  Getting rid of the lock applies to all missiles, not just multilock missiles.

PR #2755, opened 2020-09-21
Fix multi crash when ship has less than two primary weapons

notimaginative:
If a ship has less than two primary weapons the client will set the primary weapon type to 255 instead of -1 due to the data type.

A proper fix for this bug is already included in the multi_packet_fixes PR which allows for more than 255 weapon types for primaries.


notimaginative:
@JohnAFernandez See if this does anything to fix your red alert crash in FSPort

Goober5000 (approved):

PR #2759, opened 2020-09-22
Fix multi screen timeouts

notimaginative:
Attempt to fix any states that cause multi to time out or that are poorly handled in multi.

### DIFF
--- a/code/cutscene/movie.cpp
+++ b/code/cutscene/movie.cpp
@@ -10,6 +10,7 @@
 
 
 #ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #endif
 
@@ -27,6 +28,7 @@
 #include "io/timer.h"
 #include "io/key.h"
 #include "mod_table/mod_table.h"
+#include "network/multi.h"
 
 extern int Game_mode;
 extern int Is_standalone;
@@ -215,6 +217,12 @@ void movie_display_loop(Player* player, PlaybackState* state) {
 		}
 
 		processEvents();
+
+		if ( (Game_mode & GM_MULTIPLAYER) && Net_player && (Net_player->flags & NETINFO_FLAG_DO_NETWORKING) ) {
+			// consider movie playback a "paused" state and process network
+			// packets accordingly
+			multi_pause_do_frame();
+		}
 
 		if (passed < sleepTime) {
 			auto sleep = sleepTime - passed;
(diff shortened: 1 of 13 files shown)


notimaginative:
Work in progress. More to come as we find and fix issues...

code/cutscene/movie.cpp
```diff
@@ -10,6 +10,7 @@
 
 
 #ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
```
JohnAFernandez, line 13: What's this for? I didn't see a reference to it in the changes.

code/popup/popupdead.cpp
```diff
@@ -445,65 +445,67 @@ int popupdead_do_frame(float  /*frametime*/)
...
+
+				case POPUPDEAD_RESPAWN_ONLY:
+				case POPUPDEAD_RESPAWN_QUIT:
+					Popupdead_choice = POPUPDEAD_DO_RESPAWN;
+					if ( (Popupdead_multi_type == POPUPDEAD_RESPAWN_QUIT) && (choice == 1) )
+						Popupdead_choice = POPUPDEAD_DO_MAIN_HALL;
+					break;
+
+				default:
+					Int3();
```
JohnAFernandez, line 476: While we're here, we should change this to an `UNREACHABLE()`

code/popup/popupdead.cpp
```diff
@@ -445,65 +445,67 @@ int popupdead_do_frame(float  /*frametime*/)
...
```
JohnAFernandez, line 506: UNREACHABLE()

JohnAFernandez (changes requested):
One question and two suggested changes.  Unreachable can make it a lot easier for users to know when to report something to SCP.

JohnAFernandez:
Also, this does not require a multi bump?

code/cutscene/movie.cpp
notimaginative: That prevents `windows.h` from including the winsock headers, which conflict with including them in psnet by way of `multi.h`.

Or as I like to say: "I hate Windows..." ;)

code/popup/popupdead.cpp
notimaginative: Ah, wasn't aware of that macro. I'll get that added.

notimaginative:
No multi bump required here. The state changes are discarded without error on older builds and none of the other changes affect compatibility.

JohnAFernandez (commented):
Looks even better than before!  I saw the extra safety checks.  Instead of approving, as discussed on discord, I'll await the last timing changes.

JohnAFernandez (approved):
Shivansps just reported *no* issues with this build.  Time to merge.

PR #2760, opened 2020-09-23
polish some SEXPs

### LINKED ISSUE #2758: Adjust play-sound-from-file and close-sound-from-file to specify that they need numeric variables
At the moment the two SEXPs do not specify what type of variable they need but the code needs a numeric variable and shows a warning if that is not the case.

To improve usability, that requirement should be added to the SEXP description.

Goober5000:
Update SEXP help regarding variable types, and also add them to the SEXP syntax checker.

Fixes #2758.


JohnAFernandez (approved):
LGTM

PR #2761, opened 2020-09-23
Improve Homeworld warp effect

Baezon:
This PR might be a bit confusing if you aren't familiar with the games, but several changes were made:

Stages were added (or more accurately, unused stages were actually used) at the very beginning and end of the effect, for when the "sweeper" is a line and gaining/shrinking in width before/after gaining/shrinking in height.

The first and final two stages when the sweeper is changing size were adjusted to move more smoothly to follow the canonical motion of the effect, as well as timings adjusted.

Allow the sweeper to follow the position of the ship at all stages.

And add the sweeper-type warp to the exclusions from requiring a specific speed before the warp may commence (if a speed was not manually specified). Canonically, Homeworld ships were shown warping regardless of their speed.

EDIT:
Attached is a before and after video of the effect.
[HWwarp.zip](https://github.com/scp-fs2open/fs2open.github.com/files/5265657/HWwarp.zip)

~~Thrusters render on the wrong side of the sweeper when warping out, but this is the second time I've encountered a sign error regarding the warp clip plane, so I'm going to go through and check every place where it is used and have it be a separate PR.~~
EDITEDIT: The solution was simple (in general the warp's orient always follow the ship's orient), turned out to be not related to other similar issues, and doesn't break anything else.

### DIFF
--- a/code/ship/shipfx.cpp
+++ b/code/ship/shipfx.cpp
@@ -3928,6 +3928,7 @@ int WE_BSG::getWarpOrientation(matrix* output)
 }
 
 //********************-----CLASS: WE_Homeworld-----********************//
+#define HOMEWORLD_SWEEPER_LINE_THICKNESS 0.002f // How tall the initial and final hyperspace "lines" are, factor of height
 WE_Homeworld::WE_Homeworld(object *n_objp, WarpDirection n_direction)
 	:WarpEffect(n_objp, n_direction)
 {
@@ -3940,16 +3941,16 @@ WE_Homeworld::WE_Homeworld(object *n_objp, WarpDirection n_direction)
 
 	//Stage duration presets
 	stage_duration[0] = 0;
-	stage_duration[1] = 1000;
-	stage_duration[2] = 0;
+	stage_duration[1] = 1200;
+	stage_duration[2] = 1600;
 	stage_duration[3] = -1;
-	stage_duration[4] = 0;
-	stage_duration[5] = 1000;
+	stage_duration[4] = 1600;
+	stage_duration[5] = 1200;
 
 	//Configure stage duration 3
 	stage_duration[3] = params->time - (stage_duration[1] + stage_duration[2] + stage_duration[4] + stage_duration[5]);
 	if (stage_duration[3] <= 0)
-		stage_duration[3] = 3000;
+		stage_duration[3] = 2400; // set for a 8 second total time
 
 	//Anim
 	anim = bm_load_either(params->anim, &anim_nframes, &anim_fps, nullptr, true);
@@ -4028,8 +4029,8 @@ int WE_Homeworld::warpStart()
 	if(direction == WarpDirection::WARP_OUT)
 		vm_vec_negate(&fvec);
 
-	width = width_full;
-	height = 0.0f;
+	width = 0.f;
+	height = height_full * HOMEWORLD_SWEEPER_LINE_THICKNESS;
 
 	if(direction == WarpDirection::WARP_IN)
 	{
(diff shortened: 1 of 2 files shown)


JohnAFernandez:
I think this warrants a before and after video.

Baezon:
Updated.

JohnAFernandez:
Code is easy enough to follow, but I'm not sure whether I should insist on 0.0f instead of 0.f, which is how we mark it everywhere else I've seen in FSO.

Baezon:
Oh, sorry, that's something I had gotten used to. Sometimes I overlook these style/formatting habits of mine.

JohnAFernandez (approved):
An easy to read change, with video proof of an improvement.  Let's get this guy merged.

code/ship/shipfx.cpp
```diff
@@ -3928,6 +3928,7 @@ int WE_BSG::getWarpOrientation(matrix* output)
 }
 
 //********************-----CLASS: WE_Homeworld-----********************//
+#define HOMEWORLD_SWEEPER_LINE_THICKNESS 0.002f // How tall the initial and final hyperspace "lines" are, factor of height
```
asarium, line 3931: Please use a `const float`.

asarium (changes requested):
I am not aware of any mods using this so any possible change in behaviour should not be an issue.

chief1983:
> I am not aware of any mods using this so any possible change in behaviour should not be an issue.

I know of at least The Apocalypse Project that would have previously used it.  I believe there was a separate mod at one point putting HW2 ships in FSO that might have used it too.

chief1983:
I believe there was a public release at one point but by now it may be classified as a retired project on HLP.

asarium:
I don't think this possibly tiny compatibility issue is a problem. It can be revisited when it actually causes problems.

chief1983:
Yeah I don't think it will either, was more trying to see if one of these mods ever release anything public so it could be tested at all.  But I can't find anything.  TAP was the HW2 mod though, I forget what the other HW1 mod was called, or if it even had a name.  I think it was just The Original Homeworld Mod or something.

X3N0-Life-Form:
The other Homeworld project was Journey to Hiigara : https://www.hard-light.net/forums/index.php?board=197.0

Both mods are dead and buried, and neither of them released any campaign, as far as I know.

Baezon:
I was actually not aware of these projects (I had started while the forums were still down)! There was also Homeworld: Pilot Academy, which is fairly old, and also didn't release a campaign (although it doesn't appear he intended to). I am currently 'remastering' it, hence my interest in this area.

Also are `const` variables always preferable to `#define`s? Sorry, I'm still in the 'try to assimilate with nearby code' phase.

asarium:
`const`s and `#define`s are mostly the same but defines can cause weird issues and do not respect scopes so if possible they should be avoided, especially for constants. Constants also give you more type safety which is always nice.

chief1983:
I double checked and I did see Journey to Hiigara earlier, but there was also a project for HW1 contained solely in one forum thread, named "The Original Homeworld Mod".   So that's at least 4 projects that might have made use of this, yet with no public releases I can find.  So yeah, nothing to worry about with this change apparently.

Goober5000:
Why does the engine glow turn off while the square is moving?  It then reappears when the ship model has disappeared.

asarium (approved):

Baezon:
The orientation of warp effect was done inconsistently with the retail one. I have since fixed it.

JohnAFernandez:
Looks like it's time to merge.  Well done!

PR #2762, opened 2020-09-23
Refactor embedded files CMake code

asarium:
I have been thinking about some shader related techniques for the Vulkan
renderer and for that we will need to add embedded files in a different
environment than they are right now since the actual shader code will
not be needed in the engine.

For this reason, I extracted the "default file" mechanism to be more
generic so that all that is necessary to include a file is to just call
a custom CMake command and the function keeps track of what needs to be
done.

Also, since the "generic" code has worked pretty well so far I see no
reason to keep around two systems which just increases the amount of
code we need to maintain. So, I just removed the windows exclusive
implementation. Now all platforms use the same "generic" way of
embedding files.


Goober5000 (approved):
Shrug of approval.

Are we certain, then, that the generic method will work properly in Windows builds?

asarium:
There are tests that cover the default file system and those appear to have passed.

PR #2763, opened 2020-09-23
Expose Weapons SS Damage Threshold for Turret Inaccuracy

wookieejedi:
This PR adds a table entry that allows modders to specify the damage threshold for when turrets become less accurate due to weapon subsystem damage. A Game table setting is easier to set once and tune, thus this PR allows that efficient exposure to the modder and does not require a large change or complex table entry.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -62,6 +62,7 @@ bool Use_engine_wash_intensity;
 bool Ai_before_physics;
 bool Swarmers_lead_targets;
 SCP_vector<gr_capability> Required_render_ext;
+float Weapon_SS_Threshold_Turret_Inaccuracy;
 
 SCP_vector<std::pair<SCP_string, gr_capability>> req_render_ext_pairs = {
 	std::make_pair("BPTC Texture Compression", CAPABILITY_BPTC)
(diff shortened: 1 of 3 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -556,6 +557,16 @@ void parse_mod_table(const char *filename)
 			stuff_boolean(&Swarmers_lead_targets);
 		}
 
+		if (optional_string("$Damage Threshold for Weapons Subsystems to Trigger Turret Inaccuracy:")) {
+			float weapon_ss_threshold;
+			stuff_float(&weapon_ss_threshold);
+			if ( (weapon_ss_threshold >= 0.0f) && (weapon_ss_threshold <= 1.0f) ) {
+				Weapon_SS_Threshold_Turret_Inaccuracy = weapon_ss_threshold;
+			} else {
+				mprintf(("Game Settings Table: '$Damage Threshold for Weapons Subsystems to Trigger Turret Inaccuracy:' value of %.2f is not between 0 and 1. Using default value of 0.70.\n", weapon_ss_threshold));
```
JohnAFernandez, line 566: I don't know if others will agree with me, but I would upgrade this to a warning because this might notify someone that they did a typo and the table is not doing what they want it to do.

wookieejedi: That's a good point. I'll note that when I first started doing these kinds of PRs I noticed the precedent in the code for `mprintf` statements to be used in cases where values check a range, whereas popups would occur if there was a critical parsing error or other substantial syntax error. Since noticing that in the code I've followed that method, but if other folks want a warning here I'd be happy to add one!

JohnAFernandez: If that's what the surrounding code is doing, I guess we'll stick to that method for now.  Those that use these really specific options probably already know how to look for their mistakes.  I have plans for table parsing, which, as a side benefit, will hopefully make looking for these mistakes much easier, anyway.

JohnAFernandez (approved):
Isn't it great when PR's are simple?  LGTM

PR #2764, opened 2020-09-24
make the scripted turret firing code a bit more robust

Goober5000:
The scripted code previously didn't check to see if a weapon bank was valid before firing it, which could have led to memory problems and starting March 15 would actually crash the game.  This adds a few safety checks.

### DIFF
--- a/code/scripting/api/objs/subsystem.cpp
+++ b/code/scripting/api/objs/subsystem.cpp
@@ -593,6 +593,12 @@ ADE_FUNC(fireWeapon, l_Subsystem, "[number TurretWeaponIndex = 1, number FlakRan
 	if (!sso->isSubsystemValid())
 		return ADE_RETURN_NIL;
 
+	// no place to fire a weapon; this may not actually be a turret
+	if (sso->ss->system_info->turret_num_firing_points <= 0)
+	{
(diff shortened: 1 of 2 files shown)


code/scripting/api/objs/subsystem.cpp
```diff
@@ -607,8 +613,6 @@ ADE_FUNC(fireWeapon, l_Subsystem, "[number TurretWeaponIndex = 1, number FlakRan
 
 	bool rtn = turret_fire_weapon(wnum, sso->ss, OBJ_INDEX(sso->objp), &gpos, &gvec, NULL, flak_range);
 
-	sso->ss->turret_next_fire_pos++;
```
JohnAFernandez, line 610: What purpose does removing this line serve?

Goober5000: I thought it was redundant, but upon closer inspection it actually isn't.  I'll put it back.

JohnAFernandez (approved):
LGTM -- ER um, didn't realize this was owned by Asarium. :V

Goober5000:
It was automatic based on the single script file affected, which is not even the meat of the PR.  So you're fine. :D

PR #2765, opened 2020-09-24
Suppress muzzle flashes from the player (and add a flag to re-enable it)

Baezon:
I went into this PR expecting to implement this, only to find this was always intended! Simply using `Viewer_mode` to determine if the player is in the cockpit is no longer enough as `VM_CAMERA_LOCKED` is true from the get-go. It was modified in PR #848 , but I'm not sure if that behavior (camera_locked being on from the start) is actually a bug or not.

Regardless I've handled the logic slightly differently to allow muzzle flashes to go deep enough to create light, but not the flash animation itself. The idea here is to quietly sweep under the rug the obnoxious bitmaps that would've been flashing in the players face, rather than implying the muzzle flash never existed in the first place.

Also extends the logic to beams, so you can now use BGreens for testing purposes without the entire screen turning white.

EDIT: Also adds a flag to force player muzzle flashes regardless, per wookieejedi's suggestion.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -63,6 +63,7 @@ bool Framerate_independent_turning; // an in-depth explanation how this flag is 
 bool Swarmers_lead_targets;
 SCP_vector<gr_capability> Required_render_ext;
 float Weapon_SS_Threshold_Turret_Inaccuracy;
+bool Render_player_mflash;
 
 SCP_vector<std::pair<SCP_string, gr_capability>> req_render_ext_pairs = {
 	std::make_pair("BPTC Texture Compression", CAPABILITY_BPTC)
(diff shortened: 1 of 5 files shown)


wookieejedi:
Good idea, though I wanted to ask if you also considered putting this as a flag that as on by default? Some mods (such as FotG and likely others) have purposely smaller muzzle flashes the the animations look good with the cockpit and removing the animation playing detracts that effect a bit, thus an option where the modder could turn the old behavior back on might be useful. This is just my input though, and if other folks say that's not a good enough reason to add a flag then that's understandable.

Baezon:
Fair enough, that seems reasonable.

wookieejedi:
Thanks, LGTM but I'll wait to merge until others have had a chance to take a look.

wookieejedi (approved):

JohnAFernandez:
Having a flag is the way to go, especially if a major mod is trying to maintain that behavior.

PR #2766, opened 2020-09-24
Put beam->finished_growing into beam->flags

Baezon:
Bit more efficient this way.


Goober5000 (approved):
I'm not sure if there's an efficiency difference execution-wise, but it helps keep the struct tidy and the code organized. :+1:

PR #2770, opened 2020-09-26
Improve shockwave handling for docked ships

Baezon:
This scales down the blast force based on the ship's contribution to its total docked mass, if docked. This will prevent crazy scenarios like small ships being docked to big ships swinging around the big ship like crazy if the small one gets hit by a shockwave, and generally makes ships more resistant to shockwaves if they are docked/the more/bigger ships are docked like one would expect. 

As the comment says this is still not technically generating the correct rotations, since the shockwave will apply a rotational force based on the dimensions of ship's bbox and its orientation to the shockwave; an aggregate bbox would have to be specially calculated and handled for this purpose. While I plan on doing that eventually, this should prevent any egregious or damaging behavior that you could get otherwise in the meantime.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -32,6 +32,7 @@
 #include "network/multimsgs.h"
 #include "network/multiutil.h"
 #include "object/objcollide.h"
+#include "object/objectdock.h"
 #include "scripting/scripting.h"
 #include "particle/particle.h"
 #include "playerman/player.h"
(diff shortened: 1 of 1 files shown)


JohnAFernandez:
Interesting that this is all it takes to fix.  Would a video showing the changes be possible?

Baezon:
[shockwavetest.zip](https://github.com/scp-fs2open/fs2open.github.com/files/5297211/shockwavetest.zip)

Also it looks like travis timed out trying to get a prebuilt library, so a restart would be appreciated.

Goober5000 (approved):

PR #2775, opened 2020-10-02
Implement Beam Flags and some adjustments to burst handling

Baezon:
Mostly wanted to be sure I was implementing a flagset properly, I'm still a bit new to to the fancier features of C++, but it does work so I think that's a good sign. For now, only implements "burst shares random targets" flag, but I will be making much more significant use of it soon.

This means that the random targets chosen by direct-fire and slash beams will still be random but seeded according to their originating burst. For this purpose a "burst seed" is used on `ship_weapon` which is recalculated on the start of a new burst, which could be used later for any other 'burst-seeded' random values for weapons.

Also removes a duplication of burst counter handling in the fighter beam code that caused erroneous burst lengths, but maintains the one difference from the main, proper burst handling: excluding fighter beams from ai fire rate scaling.


notimaginative:
Just to confirm, this is only used to fill in the `dir_a` and `dir_b` fields of the `beam_info` struct, correct? If so then this shouldn't affect multi, but I want to be sure in case I need to make any changes there.

Baezon:
That's right, beam_info is unaffected, it'll just choose those values differently sometimes.

notimaginative:
The seed was added to the `beam_fire_info` struct, which I just got through reworking multi packets to support properly. That's why I was asking. But it doesn't look like it's something we'll need to keep track of for multi.

Goober5000 (approved):
Looks good, I think.

PR #2781, opened 2020-10-03
Use a stack for scripting hook variables

asarium:
Previously, if a script called a function that called another scripting
hook, some hook variables might be removed. This was the case since
after executing the hook the code would remove the set hook variables
again. If these variables had the same name as previously set variables,
they would overwrite and corrupt the state the original script expected.

This was unexpected behavior for script writers so this puts values on a
stack and only adds or removes from the stack instead of overwriting
values. This should make the hook variable state consistent for scripts
even if scripting hooks are called indirectly.

### DIFF
--- a/code/scripting/api/libs/hookvars.cpp
+++ b/code/scripting/api/libs/hookvars.cpp
@@ -29,8 +29,13 @@ ADE_INDEXER(l_HookVar,
 	if (iter == hookVars.end()) {
 		return ADE_RETURN_NIL;
 	}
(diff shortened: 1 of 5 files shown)


Goober5000 (approved):
I didn't see any issues, but then I think you're the only active coder who fully understands this system.

I assume the unit tests are robust enough to catch any potential issues?

asarium:
Yes, the unit tests should cover the expected use cases.

PR #2783, opened 2020-10-05
Flag to allow ships with no shields to manage ETS

wookieejedi:
Using the `No_shields` flag for ships is common usage in multiple mods, and as it turns out ships with that flag set will not have the AI manage ETS values due to an early return in the `ai_manage_ets` function. This PR adds an AI flag to allow ships with no shields to manage ETS regarding weapons and engines. This flag was tested and works as expected.

### DIFF
--- a/code/hud/hudets.cpp
+++ b/code/hud/hudets.cpp
@@ -225,7 +225,7 @@ void ai_manage_ets(object* obj)
 		decrease_recharge_rate(obj, WEAPONS);
 	}
 
-	if (obj->flags[Object::Object_Flags::No_shields]) {
+	if (!(obj->flags[Object::Object_Flags::No_shields])) {
 		float shield_left_percent = get_shield_pct(obj);
 		// maximum level check for shields
 		if (shield_left_percent == 1.0f) {
@@ -248,7 +248,7 @@ void ai_manage_ets(object* obj)
 	}
 
 	// emergency check for ships with shields
-	if (obj->flags[Object::Object_Flags::No_shields]) {
+	if (!(obj->flags[Object::Object_Flags::No_shields])) {
 		float shield_left_percent = get_shield_pct(obj);
 		if ( shield_left_percent < SHIELDS_EMERG_LEVEL_PERCENT ) {
 			if (ship_p->target_shields_delta == 0.0f)
--- a/code/hud/hudets.cpp
+++ b/code/hud/hudets.cpp
@@ -202,6 +202,7 @@ void ai_manage_ets(object* obj)
 {
 	ship* ship_p = &Ships[obj->instance];
 	ship_info* ship_info_p = &Ship_info[ship_p->ship_info_index];
+	ai_info	*aip = &Ai_info[Ships[Pl_objp->instance].ai_index];
 
 	if ( ship_info_p->power_output == 0 )
 		return;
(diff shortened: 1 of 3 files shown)


code/hud/hudets.cpp
```diff
@@ -202,37 +202,44 @@ void ai_manage_ets(object* obj)
...
-	if (!(obj->flags[Object::Object_Flags::No_shields]) && (shield_left_percent == 1.0f)) {
-		decrease_recharge_rate(obj, SHIELDS);
-	}
-
-	// minimum check
-
-	if (!(obj->flags[Object::Object_Flags::No_shields]) && (shield_left_percent < SHIELDS_MIN_LEVEL_PERCENT)) {
-		if ( weapon_left_percent > WEAPONS_MIN_LEVEL_PERCENT )
-			increase_recharge_rate(obj, SHIELDS);
+	bool has_shields = !ship_p->ship_max_shield_strength && !(obj->flags[Object::Object_Flags::No_shields]);
```
Goober5000, line 228: Is this condition correct?  You're checking to see whether the ship has 0% maximum shields and also doesn't have the no-shields flag.

code/hud/hudets.cpp
```diff
@@ -241,16 +248,16 @@ void ai_manage_ets(object* obj)
 		increase_recharge_rate(obj, ENGINES);
 	}
 
-	// emergency check
-	if (!(obj->flags[Object::Object_Flags::No_shields])) {
+	// emergency check for ships with shields
+	if (has_shields) {
```
JohnAFernandez, line 252: Whether or not the other place has_shields is kept, we may want to keep the original check here.  These tiny differences could cause AI differences that are difficult to detect later.  The original writer could have extended the check to include `!ship_p->ship_max_shield_strength`, but they didn't, and we just have to stick to retail as close as possible.

code/hud/hudets.cpp
```diff
@@ -276,20 +282,21 @@ void ai_manage_ets(object* obj)
 void set_default_recharge_rates(object* obj)
 {
 	int ship_properties;
-
+	
 	ship* ship_p = &Ships[obj->instance];
 	ship_info* ship_info_p = &Ship_info[ship_p->ship_info_index];
-
+	mprintf(("settig up ETS for ship %s ...\n", ship_p->ship_name));
```
Goober5000, line 288: Is this statement going to spam the debug log?

code/hud/hudets.cpp
```diff
@@ -276,20 +282,21 @@ void ai_manage_ets(object* obj)
...
 
 	ship_properties = 0;	
 	if (ship_has_energy_weapons(ship_p))
 		ship_properties |= HAS_WEAPONS;
 	
-	if (!(obj->flags[Object::Object_Flags::No_shields]))
+	if (!(obj->flags[Object::Object_Flags::No_shields])) {
 		ship_properties |= HAS_SHIELDS;
-
+		mprintf(("following ship has sheilds %s ...\n", ship_p->ship_name));
```
Goober5000, line 298: Also here

code/hud/hudets.cpp
```diff
@@ -317,6 +324,7 @@ void set_default_recharge_rates(object* obj)
 			ship_p->shield_recharge_index = ZERO_INDEX;
 			ship_p->weapon_recharge_index = ONE_HALF_INDEX;
 			ship_p->engine_recharge_index = ONE_HALF_INDEX;
+			mprintf(("settig shield to 0 for ship %s ...\n", ship_p->ship_name));
```
Goober5000, line 327: And here

Goober5000 (changes requested):

JohnAFernandez (changes requested):

code/hud/hudets.cpp
wookieejedi: Good point, I'll stick to exactly what the conditions were since the check in the early return looks at `!ship_p->ship_max_shield_strength`

code/hud/hudets.cpp
wookieejedi: Correct, though with Cyborg's comments this has changed.

code/hud/hudets.cpp
wookieejedi: Indeed, thought I had removed that but forgot to push it!

JohnAFernandez (approved):

code/hud/hudets.cpp
```diff
@@ -241,16 +247,16 @@ void ai_manage_ets(object* obj)
 		increase_recharge_rate(obj, ENGINES);
 	}
 
-	// emergency check
-	if (!(obj->flags[Object::Object_Flags::No_shields])) {
+	// emergency check for ships with shields
+	if (obj->flags[Object::Object_Flags::No_shields]) {
```
Goober5000, line 251: You flipped the boolean condition here.

code/hud/hudets.cpp
```diff
@@ -202,37 +202,43 @@ void ai_manage_ets(object* obj)
...
-	if (!(obj->flags[Object::Object_Flags::No_shields]) && (shield_left_percent == 1.0f)) {
-		decrease_recharge_rate(obj, SHIELDS);
-	}
-
-	// minimum check
-
-	if (!(obj->flags[Object::Object_Flags::No_shields]) && (shield_left_percent < SHIELDS_MIN_LEVEL_PERCENT)) {
-		if ( weapon_left_percent > WEAPONS_MIN_LEVEL_PERCENT )
-			increase_recharge_rate(obj, SHIELDS);
+	if (obj->flags[Object::Object_Flags::No_shields]) {
```
Goober5000, line 228: Here too.

code/hud/hudets.cpp
wookieejedi: Whoops, thanks for the catch!

code/hud/hudets.cpp
wookieejedi: Thanks!

Goober5000 (approved):
Looks good. :+1:

wookieejedi:
Excellent, thanks!

PR #2784, opened 2020-10-06
Message size fix

Goober5000:
1) Fix underline length calculation in message scrollback
2) Use SCP_string for message source and message text (which allows ships with long display names to work without crashing)
3) Upgrade assorted functions to work with SCP_string

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -233,7 +233,7 @@ const auto OnMessageReceivedHook = scripting::Hook::Factory(
 	});
 
 // forward declarations
-void message_maybe_distort_text(char *text, int shipnum, bool for_death_scream);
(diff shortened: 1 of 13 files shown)


code/mission/missionmessage.cpp
```diff
@@ -2276,8 +2274,7 @@ void message_maybe_distort_text(char *text, int shipnum, bool for_death_scream)
 
 			++i;
 		}
-		Assertion(result_str.size() <= buffer_size, "Buffer after scrambling message is bigger than before!");
-		strcpy(text, result_str.c_str());
+		text = result_str;
```
asarium, line 2277: You could use `std::move` for a bit of efficiency:
```suggestion
		text = std::move(result_str);
```

asarium (approved):

PR #2788, opened 2020-10-09
Multiplayer Resimulation for Client Fired Linear Weapons (Rollback)

JohnAFernandez:
This creates a packet that sends the relative position and orientation info for a primary or dumbfire missile shot.  The client creates this packet when firing those two types of weapons and sends it to the server.  The server goes back to the frame that the client fired the weapon and quickly re-simulates all the frames in-between.

This includes collision detection, restoring positions, orientations and velocities, but not subsystem orientations.

Now that the object update packet is merged, note that this is mainly to ensure that high ping clients can fire primary shots with the same accuracy on server and client.

### DIFF
--- a/code/network/multi_obj.cpp
+++ b/code/network/multi_obj.cpp
@@ -22,6 +22,7 @@
 #include "network/multi_rate.h"
 #include "network/multi.h"
 #include "object/object.h"
+#include "object/objcollide.h"		// for multi rollback collisions
 #include "object/objectshield.h"
 #include "ship/ship.h"
 #include "playerman/player.h"
@@ -48,6 +49,7 @@ struct oo_ship_position_records {
 	vec3d positions[MAX_FRAMES_RECORDED];							// The recorded ship positions, cur_frame_index is the index.
 	matrix orientations[MAX_FRAMES_RECORDED];						// The recorded ship orientations, cur_frame_index is the index. 
 	vec3d velocities[MAX_FRAMES_RECORDED];							// The recorded ship velocities (required for additive velocity shots and auto aim), cur_frame_index is the index.
+	vec3d rotational_velocities[MAX_FRAMES_RECORDED];				// The recorded ship rotational velocities (required for auto aim if certain ), cur_frame_index is the index.
 };
 
 // keeps track of what has been sent to each player, helps cut down on bandwidth, allowing only new information to be sent instead of old.
@@ -132,6 +134,7 @@ struct oo_rollback_restore_record {
 	vec3d position;			// position to restore to
 	matrix orientation;		// orientation to restore to
 	vec3d velocity;			// velocity to restore to
+	vec3d rotational_velocity; // rotational velocity to restore to
 };
 
 // Keeps track of how many shots that need to be fired in rollback mode.
@@ -159,8 +162,6 @@ struct oo_general_info {
 	// how many times that is reset, up to MAX_SERVER_TRACKER_SMALL_WRAPS, we multiply these two to get the seq_num
 	// sent to oo_packet recipients. larger_wrap_count allows us to figure out if we are in an "odd" larger wrap
 	int number_of_frames;									// how many frames have we gone through, total.
-	short larger_wrap_count;								// how many of the larger wrap, used to set an OO_packet flag.
-	ushort wrap_count;										// how many times have we wrapped the cur_frame_index.
 	ubyte cur_frame_index;									// the current frame index (to access the recorded info)
 
 	int timestamps[MAX_FRAMES_RECORDED];					// The timestamp for the recorded frame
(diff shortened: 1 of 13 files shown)


JohnAFernandez:
If you started a review, I simplified one of the functions, by sending an int instead of a ubyte in the new packet type, so you may want to take another look.

code/network/multimsgs.cpp
```diff
@@ -7611,6 +7580,220 @@ void process_NEW_primary_fired_packet(ubyte *data, header *hinfo)
...
+	}
+
+	// Add the shooting object.
+	ADD_USHORT(objp->net_signature);
+	ADD_DATA(flags);
+	ADD_USHORT(ref_objp->net_signature);
+
+	// We need the time elpased, so send the last frame we got from the server and how much time has happened since then.
+	int last_received_frame = multi_client_lookup_frame_idx();
+	auto time_elapsed = (ubyte)(timestamp() - multi_client_lookup_frame_timestamp());
```
notimaginative, line 7635: What happens if this wraps? Given sufficient lag it would be possible so if this would cause a problem then I'd suggest making this a ushort.

code/object/objcollide.cpp
```diff
@@ -1023,7 +1023,7 @@ void obj_find_overlap_colliders(SCP_vector<int> &overlap_list_out, SCP_vector<in
 static SCP_vector<int> sort_list_y;
 static SCP_vector<int> sort_list_z;
 
-void obj_sort_and_collide()
+void obj_sort_and_collide(SCP_vector<int> Collision_list)
```
notimaginative, line 1026: This should be `&Collision_list`. Otherwise you're just copying the passed list to this one and sorting the copy. The list you intended to sort remains unchanged.

JohnAFernandez: Well that's inefficient!

JohnAFernandez:
I'll get started on these changes because it turns out the fotg bug had nothing to do with this pr whatsoever.  Autoaim in multi needs to be fixed.

code/network/multimsgs.cpp
JohnAFernandez: Changes made....

code/network/multi_obj.cpp
```diff
@@ -600,6 +580,220 @@ int multi_ship_record_find_time_after_frame(int starting_frame, int ending_frame
...
+	object* wobjp = &Objects[wep_objnum];
+
+	// check for valid pointer
+	if (wobjp == nullptr){
+		mprintf(("Nullptr when trying to add weapons to the weapon rollback tracker.\n"));
+		return;
+	}
+	
+	// add it to the list of weapons we'll need to add to the simulation.
+	Oo_info.rollback_wobjp_created_this_frame.push_back(wobjp);
```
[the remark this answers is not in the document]
JohnAFernandez: Changes made.  Need to double check that I didn't break anything with these changes.

code/network/multi_obj.cpp
```diff
@@ -600,6 +580,220 @@ int multi_ship_record_find_time_after_frame(int starting_frame, int ending_frame
...
+// Returns whether weapons currently being created should be part of the rollback simulation.
+bool multi_ship_record_get_rollback_wep_mode() 
+{
+	return Oo_info.rollback_mode;
+}
+
+// Adds an object pointer to the list of weapons that needs to be simulated as part of rollback.
+void multi_ship_record_add_rollback_wep(int wep_objnum) 
+{
+	object* wobjp = &Objects[wep_objnum];
```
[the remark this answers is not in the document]
JohnAFernandez: Changes made.

code/object/objcollide.cpp
JohnAFernandez: Changes made...

JohnAFernandez:
Now to retest....

JohnAFernandez:
Testing went fine!  I made a couple of edits that are different from before @notimaginative

code/object/objcollide.cpp
```diff
@@ -1023,7 +1023,7 @@ void obj_find_overlap_colliders(SCP_vector<int> &overlap_list_out, SCP_vector<in
 static SCP_vector<int> sort_list_y;
 static SCP_vector<int> sort_list_z;
 
-void obj_sort_and_collide()
+void obj_sort_and_collide(SCP_vector<int>* Collision_list)
```
notimaginative, line 1026: You might need a `nullptr` check on this. It really shouldn't be needed the way it's used, but you never know. I'll leave it up to you.

But if you do add it I might suggest making the default be `nullptr` rather than pointing to `Collision_sort_list`, then if it's `nullptr` assign `Collision_sort_list` at that time. That makes it a little easier to understand what's going to happen without having to look at another part of the code. That's more of a personal preference/style though.

notimaginative (approved):
LGTM, as far as I can tell. Maybe someone else will give it a look and notice something that I didn't.

code/object/objcollide.cpp
JohnAFernandez: I'm all for more readability, so I'll make that change.

JohnAFernandez:
I'll give it a day before I merge.  Speak now or forever hold your peace!

PR #2789, opened 2020-10-10
Add 'AI respect tabled turn time and rotdamp' flag

Baezon:
In combination with framerate independent turning, allows modders to actually get the numbers they're listing on the table, rather than attempting to compensate, and also has more logical handling of rotdamp so it will actually effect AI ships more-or-less as it does players (rather than reversed as it is in retail, where more rotdamp meant faster acceleration). I used the '+' on the flag here to hopefully try and indicate more that it is explicitly dependent on the framerate independent turning flag.


wookieejedi (approved):
Nice to see this one finally get fully integrated.

PR #2790, opened 2020-10-10
Fix banking constant not applying to AI

Baezon:
Pretty simple fix, `ai_turn_towards_vector` never used `delta_bank_const` so 'Banking Constant' would be ignored for them in most cases. ~~This is default 1, so this won't affect retail.~~ The default is actually 0.5 as mentioned below.


Goober5000 (approved):

wookieejedi:
Just for archival sake I'll note that default banking constant is 0.5, `Defines a factor for how much roll is added during a yaw. Set as 1 for full roll and 0 is no roll. Default is 0.5`. Given discussion with Baezon it seems like retail still won't be affected though. [https://wiki.hard-light.net/index.php/Ships.tbl#.24Banking_Constant:](url)

Baezon:
I've doubled the existing fudge factor of 100, this way, for the default value of 0.5, the resulting behavior is unaffected.

PR #2791, opened 2020-10-11
Add 'Turning Acceleration Time' to missiles

Baezon:
Allows missiles to accelerate to their maximum turning speed, adds an interesting extra dynamic to their flight, and also makes their path look a bit more realistic, but does require `Framerate_independent_turning`

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -1333,6 +1333,14 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 				stuff_float(&wip->turn_time);
 			}
 
+			if (optional_string("+Turning Acceleration Time:")) {
+				stuff_float(&wip->turn_accel_time);
+				if (!Framerate_independent_turning) {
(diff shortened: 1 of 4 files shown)


wookieejedi (approved):

code/weapon/weapons.cpp
```diff
@@ -1333,6 +1333,14 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 				stuff_float(&wip->turn_time);
 			}
 
+			if (optional_string("+Turning Acceleration Time:")) {
+				stuff_float(&wip->turn_accel_time);
+				if (!Framerate_independent_turning) {
+					Warning(LOCATION, "Turning Acceleration Time for weapon %s requires \"AI use framerate independent turning\" in game_settings.tbl ", wip->name);
```
JohnAFernandez, line 1339: Since this would be true for any entry where they attempt to use `+Turning Acceleration Time:`, and they only need to set `Framerate_independent_turning` once to fix it for all the entries that use this new flag, we should probably use a static bool to only show this warning once.

JohnAFernandez (changes requested):

JohnAFernandez (approved):

PR #2793, opened 2020-10-11
Support vararg specifiers in Lua documentation

asarium:
This was the last thing that was missing for complete support of all
function parameter declarations.

To ensure parameter lists stay valid a unit test was added that will
validate the argument lists and cause the test to fail if they cannot be
parsed.


Goober5000 (approved):
Shrug of approval.

PR #2794, opened 2020-10-11
Fix to Non Shielded Ships Manage ETS

wookieejedi:
This one line fix properly specifies which ship to use for ETS managment. The recent merge of `Non-shielded ships manage ETS` had a object reference error, so this rectifies that. Thanks to Axem for finding this!


JohnAFernandez (approved):
Happens to everyone

PR #2795, opened 2020-10-12
Allow for 'laser-style' particles

Baezon:
Adds the 'Length' field to particles in particles.tbl, which, when specified, will render the particle like a laser, with the head and tail radius as the 'Size', and the length... as the 'Length'. Importantly also orienting the bitmap along its velocity vector. I'm still acclimating to the fancier C++ structures, so some oversight would be appreciated.

### DIFF
--- a/code/particle/particle.cpp
+++ b/code/particle/particle.cpp
@@ -440,7 +440,7 @@ namespace particle
 
 			Assert( cur_frame < part->nframes );
 
-			if (part->length) {
+			if (part->length != 0.0f) {
 				vec3d p0 = part->pos;
 
 				vec3d p1;
--- a/code/particle/particle.cpp
+++ b/code/particle/particle.cpp
@@ -140,6 +140,7 @@ namespace particle
 		part->reverse = info->reverse;
 		part->particle_index = (int) Persistent_particles.size();
 		part->looping = false;
+		part->length = info->length;
 
 		switch (info->type)
 		{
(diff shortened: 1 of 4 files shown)


asarium:
Do you want to support particles with negative lengths as well or just positive length? At the moment, the code supports negative values as well.

Baezon:
Yeah it looks like normally lasers with length less than 0.0001 aren't rendered, but `batching_add_laser` doesn't seem to have any issues with it, it renders them as one would expect (backwards). Unless it actually causes problems I see no reason to exclude it.

code/particle/particle.cpp
```diff
@@ -439,7 +440,20 @@ namespace particle
 
 			Assert( cur_frame < part->nframes );
 
-			batching_add_volume_bitmap(framenum + cur_frame, &pos, part->particle_index % 8, part->radius, alpha);
+			if (part->length) {
```
asarium, line 443: Please make it specific that 0 is the magic value here.
```suggestion
			if (part->length == 0.0f) {
```

code/particle/util/ParticleProperties.cpp
```diff
@@ -6,7 +6,7 @@
 
 namespace particle {
 namespace util {
-ParticleProperties::ParticleProperties() : m_radius(0.0f, 1.0f), m_lifetime(0.0f, 1.0f) {
+ParticleProperties::ParticleProperties() : m_radius(0.0f, 1.0f), m_lifetime(0.0f, 1.0f), m_length (0.0f, 1.0f){
```
asarium, line 9: Since exiting code should be unaffected, the random range should only contain the single, "magic" zero value.
```.suggestion
ParticleProperties::ParticleProperties() : m_radius(0.0f, 1.0f), m_lifetime(0.0f, 1.0f), m_length (0.0f) {
```

asarium (changes requested):

wookieejedi (approved):
Makes sense to me, nice addition.

asarium (approved):

PR #2800, opened 2020-10-14
Add first basic rendering with Vulkan

asarium:
This is basically the tutorial from vulkan-tutorial.com adapted to using
VulkanHpp and with some refactorings that extract some locigally
self-contained code.

Currently only draws a simple colored triangle.


JohnAFernandez (approved):
I'll take the plunge and approve this.

PR #2802, opened 2020-10-15
Add -no_screenshake flag to optionally disable screen shaking effects…

PhantomHoover:
… for afterburner/cannon/shockwaves

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -383,7 +383,7 @@ cmdline_parm weapon_choice_3d_arg("-weapon_choice_3d", NULL, AT_NONE);	// Cmdlin
 cmdline_parm use_warp_flash("-warp_flash", NULL, AT_NONE);	// Cmdline_warp_flash
 cmdline_parm allow_autpilot_interrupt("-no_ap_interrupt", NULL, AT_NONE);
 cmdline_parm stretch_menu("-stretch_menu", NULL, AT_NONE);	// Cmdline_stretch_menu
-cmdline_parm no_screenshake("-no_screenshake", NULL, AT_NONE); // Cmdline_no_screenshake
+cmdline_parm no_screenshake("-no_screenshake", nullptr, AT_NONE); // Cmdline_no_screenshake
 
 int Cmdline_3dwarp = 0;
 int Cmdline_ship_choice_3d = 0;
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -188,6 +188,7 @@ Flag exe_params[] =
 	{ "-3dwarp",			"Enable 3D warp",							true,	0,									EASY_DEFAULT,					"Gameplay",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-3dwarp", },
 	{ "-warp_flash",		"Enable flash upon warp",					true,	0,									EASY_DEFAULT,					"Gameplay",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-warp_flash", },
 	{ "-no_ap_interrupt",	"Disable interrupting autopilot",			true,	0,									EASY_DEFAULT,					"Gameplay",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-no_ap_interrupt", },
+	{ "-no_screenshake",	"Disable screen shaking",					true,	0,									EASY_DEFAULT,					"Gameplay",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-no_screenshake", },
 
 	{ "-nosound",			"Disable all sound",						false,	0,									EASY_DEFAULT,					"Audio",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-nosound", },
 	{ "-nomusic",			"Disable music",							false,	0,									EASY_DEFAULT,					"Audio",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-nomusic", },
@@ -382,6 +383,7 @@ cmdline_parm weapon_choice_3d_arg("-weapon_choice_3d", NULL, AT_NONE);	// Cmdlin
 cmdline_parm use_warp_flash("-warp_flash", NULL, AT_NONE);	// Cmdline_warp_flash
 cmdline_parm allow_autpilot_interrupt("-no_ap_interrupt", NULL, AT_NONE);
 cmdline_parm stretch_menu("-stretch_menu", NULL, AT_NONE);	// Cmdline_stretch_menu
+cmdline_parm no_screenshake("-no_screenshake", NULL, AT_NONE); // Cmdline_no_screenshake
 
 int Cmdline_3dwarp = 0;
 int Cmdline_ship_choice_3d = 0;
@@ -389,6 +391,7 @@ int Cmdline_weapon_choice_3d = 0;
 int Cmdline_warp_flash = 0;
 int Cmdline_autopilot_interruptable = 1;
 int Cmdline_stretch_menu = 0;
+int Cmdline_no_screenshake = 0;
 
 // Audio related
 cmdline_parm voice_recognition_arg("-voicer", NULL, AT_NONE);	// Cmdline_voice_recognition
@@ -1840,6 +1843,10 @@ bool SetCmdlineParams()
 		Cmdline_autopilot_interruptable = 0;
 	}
 
+	if ( no_screenshake.found() ) {
+		Cmdline_no_screenshake = 1;
+	}
+	
 	if ( stretch_menu.found() )	{
 		Cmdline_stretch_menu = 1;
 	}
--- a/code/cmdline/cmdline.h
+++ b/code/cmdline/cmdline.h
@@ -90,6 +90,7 @@ extern int Cmdline_weapon_choice_3d;
 extern int Cmdline_warp_flash;
 extern int Cmdline_autopilot_interruptable;
 extern int Cmdline_stretch_menu;
+extern int Cmdline_no_screenshake;
 
 // Audio related
 extern int Cmdline_voice_recognition;
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -2854,8 +2854,8 @@ void apply_view_shake(matrix *eye_orient)
 	tangles.h = 0.0f;
 	tangles.b = 0.0f;
 
-	// do shakes that only affect the HUD (unless disabled by cmdline)
-	if (Viewer_obj == Player_obj && !Cmdline_no_screenshake) {
+	// do shakes that only affect the HUD (unless disabled by cmdline in singleplayer)
+	if (Viewer_obj == Player_obj && (!Cmdline_no_screenshake || Game_mode & GM_MULTIPLAYER)) {
 		physics_info *pi = &Player_obj->phys_info;
 
 		// Make eye shake due to afterburner
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -2854,8 +2854,8 @@ void apply_view_shake(matrix *eye_orient)
 	tangles.h = 0.0f;
 	tangles.b = 0.0f;
 
-	// do shakes that only affect the HUD
-	if (Viewer_obj == Player_obj) {
+	// do shakes that only affect the HUD (unless disabled by cmdline)
+	if (Viewer_obj == Player_obj && !Cmdline_no_screenshake) {
 		physics_info *pi = &Player_obj->phys_info;
 
 		// Make eye shake due to afterburner


JohnAFernandez (commented):
Might be good to have a game_settings.tbl option, too, since then we can have mods define that if it is their preference.

chief1983:
So, this was actually requested once before, and I'm still kind of against it being a command line flag.  It'd be a pretty significant impact on multiplayer to not have to deal with this effect compared to other players.  I'm all for mods being able to turn off or adjust it, and if you want an optional package that could installed with FSO to turn it off, fine.  But as a command line flag, I'm just not a fan.

chief1983:
#1162

Baezon:
Given the strong effect it has I think this pretty reasunable to expose this, multi might possibly override it, or even if not its an advantage everyone is freely able to have. 

Importantly however this does **NOT** affect shockwaves (see `physics_sim_rot` in physics.cpp). Shockwaves physically jostle the ship, not just the camera.

chief1983:
This is actually almost identical to the previously submitted PR from almost four years ago.  I'm not sure what new light there is to justify it as-is now compared to then.  I'm sorry PH doesn't like the effect, but we don't really want to have command line options for every nitpick out there.  I'd still rather see more command line options removed than new ones added, personally.  Again, I'm not against this being something in a game settings tbm, then it could be automatically invalidated on multiplayer if it's in use through existing table validation.  But hacking around command line options and  multiplayer is not something we should be getting into when there are better options.

chief1983 (dismissed):
This set of changes was rejected by three devs previously, and I don't think those opinions have changed.  Would prefer to see implemented as a game_settings option.  And please review the other comments in that original PR to ensure that the toggle itself is affecting the correct situations.

Baezon:
I personally don't find any of the previous arguments convincing. FSO provides plenty of avenues for players to customize their aesthetic experience, and this is definitely an aesthetic/cosmetic issue, it does not affect behavior at all, so I don't see the issues with providing the option. PH is far from alone on this (youll notice the original PR says "widely hated" for a reason). I find it quite tiresome to switch to chase view every time I need to use the maxim, so it would be appreciated if I didn't have to do that.

EatThePath:
the idea of making screenshake a gamesettings table option seems to be missing the point of this being a personal user preference issue, unless someone who finds screenshake unpleasant is expected to just slap that change on top of every mod they ever want to play.

Personally this seems like something that ought to live in the ingame graphics options rather than a command line flag, but that's not really an option for FSO I suppose.

chief1983:
It's not just purely aesthetic though.  It's not just how something looks in the tech room, or how wide the main hall is.  It's a feature of the game itself that all users have had to deal with forever, and turning it off would obviously make things in the game easier, or you wouldn't be asking for it in the first place apparently.  The maxim being obnoxious with the screen shaking is kind of the point, disabling the screen shake for it would basically be a balance buff, and only for those who turned it off.  If it makes the Maxim more difficult to use, that's by design, not just an aesthetic.  This is why a command line is the wrong place for this kind of option.  It's a feature that everyone should be on the same page with, either the way the mod wanted it to be experienced, or the same everywhere in a multiplayer environment, or enabled similar to 3d radar/cockpits, via a table setting/mod, not a command line flag.

I'm also all for upgrading the effect to behave better on modern tech.  Back in the 640x480 or 800x600 30fps era, maybe it wasn't as rough as on 120fps at 4k now.  Similar to how the motion debris really aged with time, although I don't think anyone could argue that turning off the motion debris somehow provides a competitive advantage or affects balance.  If anything, that made it harder to orient yourself at time, not easier.

EatThePath:
I guess to me the balance issues created feel relatively insignificant. Since it doesn't actually move your aim as far as I know, the effect seems to me to make using a weapon unpleasant more than difficult. Which aren't entirely unrelated, perhaps, but there is a difference. For example if an airhorn went off next to my ear every time I fired a missile, I wouldn't much care for using that missile, but it wouldn't actually be any hard to use. Freespace screenshake feels more like that than anything else to me.

Baezon:
I don't turn it off because it makes things easier I turn it off because it makes things not visually obnoxious. There are many more difficult features which I am happy to leave as is because I have nothing against the challenge. Shudder isn't a challenge, its a visual effect that gives me a headache. Given how easy it is to disable, and how strongly some people feel about it, it seems bizarre to argue against player control of the look and feel of their game.

Mods have default flags for a reason, if the modded has a preference theyll be free to set it.

wookieejedi:
Would it be possible for the flag to only work on single-player? So if the mode is multi then there is shaking, if the mode is single-player then the shaking follows the flag.

PhantomHoover:
The concern about adding more command-line flags is legitimate, but it's a wider problem that's outside the scope of this pull request. FSO does have a real problem with configuration piling up in the command line because there's no functional alternative, but the solution is to make that alternative, not tie the hands of users.

I don't really accept any of the gameplay-based arguments. Firstly, most players learn sooner or later that they can switch to 3rd person and entirely circumvent screenshake. Secondly, and more deeply, I believe that FSO should first and foremost serve the interests of the user running it on their own computer. I bothered to actually submit this change because at least two people immediately said that if this flag existed they'd use it, and I just don't think it's the place of SCP coders to tell players that they're not allowed to play their own game in certain ways.

There is something to be said for this perhaps providing an unfair advantage in multi. Honestly, though, since 3rd person works just as well in multi, I think letting people turn off 1st person screenshake is objectively a non-issue in balance terms. If there's strong feeling on this front I can look into restricting the flag's effect to single player.

Adding mod control of this to gamesettings.tbl seems like a good extension to make but the logical flow in the code would be a bit more complex, as the cmdline flag must override gamesettings.tbl if and only if it's present, and should be able to force screenshake on as well. Frankly I'd prefer to sort out the debate over whether the feature has a right to exist out first.

PhantomHoover:
If FSO is in a place where tables are acceptably configurable for a player to use them to override settings per their own preferences, without using VP tools, then I'm happy with that. But as far as I know it isn't. game_settings.tbl is not modular and it is very much designed to be a monolithic configuration set by the current mod, and FSO's only real user configuration interface remains the command-line flags.

MageKing17:
> game_settings.tbl is not modular

Er... it is, though? The modular equivalent is the "-mod.tbm" suffix.

That being said, while I have no strong opinions about screen shudder, I also don't buy the argument that turning it off has a major impact on game balance. If you want to make that argument, you could argue that HUD clutter affects game balance and that turning off HUD gauges shouldn't be allowed in multiplayer... which is ridiculous.

X3N0-Life-Form:
Honestly, I do think it has an effect on balance, as it makes it more difficult to adjust your aim vs moving targets, which I think was the point of the effect. I personally believe that the screen not shaking in chase view is a :V: oversight rather than a feature.

PhantomHoover:
I don't disagree that it's a balance issue: there's a reason it's only used on the Maxim, after all. But this is also a low-key accessibility issue: the reason screenshake is so widely hated isn't that it's throwing off your aim on Maxim shots, it's that it makes the game physically unpleasant to look at (especially when you pump afterburners, which is an intermediate technique that players all learn at some point). That is something that needs to be opt-out, and it's something that _ordinary users_ need to be allowed to opt out of, through the standard configuration interface. FSO's user configuration interface is currently a mess of command line flags; there is ongoing work to replace this with a proper in-game interface, but for now I need to implement this feature with the existing tool.

Chief is also right that this effect is a lot harsher at 120Hz and modern resolutions than it was in 1999, and could really benefit from a modernisation to make it less harsh. But any effort that tries to fix that will likely run into this exact same project politics wall: there will be pushback to any patch that changes retail screenshake logic by default, there will be pushback to any further command-line clutter if it's switched on there, and so it'll be pushed into table settings where only modders get to decide when players can play the game without it hurting their eyes. I want this question resolved here, now, on this change which presents a minimum viable solution to the screenshake problem.

The-E:
Count me on the side of people not considering the removal of screen shake as much of a balance impact.

I think the benefits for usability/accessibility outweigh the fact that people can aim their maxims more good.

notimaginative:
The screen shake was a conscious decision as part of the engine design. It's not a bug to be fixed. Because you don't like it isn't a valid reason to disable it. Regardless, this should absolutely not be a cmdline option. We've already got too many of those that shouldn't exist.

As a game_settings option it might make sense. Then it stays a design decision rather than a random preference. However I'm not aware of any mod asking for such a thing.

As a feature of accessibility it makes for a good argument. But that argument was made after the fact, as an excuse to allow the change. If this idea were reintroduced as part of series of accessibility changes then it would be worth taking another look at. Assuming that those changes are tbl/tbm based such that they could be distributed as a special accessibility VP. This would allow those who need it to easily get the changes and would give modders something to build upon and customize to better suit their mod.

PhantomHoover:
> The screen shake was a conscious decision as part of the engine design. It's not a bug to be fixed. Because you don't like it isn't a valid reason to disable it.

Of course it's a valid reason for me to disable it, it's my game and I'm playing it. If there was an acceptable user-facing configuration panel other than the command line I'd have used that, but it doesn't exist. Command-line proliferation is messy but it's not a giant problem given that all the launchers have interfaces to present them in a manageable format.

The purpose of FreeSpace Open is to act as a platform for modders to offer their work for players to enjoy. If players are consistently hating screen-shake then it's absolutely appropriate for FSO to provide them a means to disable it, just as it provides difficulty, cheats and audio volume settings. For this reason I am opposed in the strongest terms to the entire suggestion in your last paragraph.

The-E:
I don't really see what the big deal is here. Yes, this allows users to disable a feature Volition put in for balance or aesthetic reasons ... but honestly? We allow people to change literally every single thing about the game's balance through tables, we even have our original cheat codes in there which were definitely never factored into any balancing and yet still exist.

We have users who say that this is impacting their ability to play and enjoy these games. Simple as that. We can fix that issue through one commandline flag; a user choosing to enable that feature does not destroy the game balance for a user who doesn't.

notimaginative:
> Of course it's a valid reason for me to disable it, it's my game and I'm playing it. If there was an acceptable user-facing configuration panel other than the command line I'd have used that, but it doesn't exist. Command-line proliferation is messy but it's not a giant problem given that all the launchers have interfaces to present them in a manageable format.

Your argument is the _exact_ reason this should not go in. If you want to use it then by all means do so, you've already got it coded. But once it goes into the code base it becomes everyone else's problem. Someone else has to maintain it, keep it working, deal with any issues that arise from having it there or removing it in the future. I often make changes to do things I like better or that make my life easier, but I would never try to make the other devs deal with those preferences.

My issue is that you are implementing something in entirely the wrong way and for entirely the wrong reasons. Do it the right way, and not just for your personal preference, and it you'll likely get a more favorable response from the other SCP devs.

And I never mentioned anything about balance issues. Considering the multitude of other options that could more dramatically affect balance, I really don't see the argument there.

Baezon:
Why do you act like making changes for one's personal preference is some kind of sin; "entirely the wrong reasons" "not just for your personal preference". Since when users did become the enemy here? Why are they being barred from customizing their experience? There is absolutely no reason to lock this under modders only, and a user's visual preferences aren't some destructive damaging influence as you seem to suggest. In what way is this "the wrong way"?

I'd be more than happy to deal with any repercussions of the flag, under the assumption that PH is unwilling or unable.

TRBlount:
So completely ignoring balancing related arguments (FreeSpace was never known for being balanced anyhow, hence FSBlue existing), it sounds like the problem here is mostly that this is a cmdline option and not a table option. The problem with plopping it into a non-modular table is that it is effectively useless in terms of usability. Obviously we overuse cmdline options a ton but there are many user-specific options there already (graphics and personal preference related). If it can't go in as a cmdline option because of bloat and it can't go into a table because that would basically defeat the point of adding it in the first place then ***where should it go***?

I see no issue with adding a feature like this since it's more aesthetic than anything (personally I dislike how it's implemented but that's a discussion for another time). Should this wait until we get an in-game options menu (where many things in `cmdline.cpp` could also be moved)? That is highly reliant on the work of Mjn and m!m/Asarium, both of which are busy with other things to my knowledge. I'm willing to give this more time and even put it to a vote but I have yet to see a good argument against having a feature ***like*** this. If your problem is it being in the cmdline then find an alternative that isn't a non-modular table and vice-versa. Sitting around arguing isn't going to solve this any faster, so I'd like to hear back from the parties involved before 0700 UTC-7 October 18.

notimaginative:
Users aren't the enemy and neither is customizing one's experience. The specific reason for the change, affecting the specific method of the change, is anti-user in general. The _idea_ of the change isn't bad and I think would be fine. But it's implemented in a individual-centric manner. It's another silly cmdline option in a sea of cmdline options which you need to study a wiki to figure out. It doesn't have to be a mod specific thing or something difficult for the user, but it could be far more useful than what's proposed here. Having worked on FSO since 2003 I can tell exactly where this will end up because I've seen it before, and that's likely why this change was denied the last time it was proposed.

But I'm just offering an opinion here based on my 17 year history with the project. I have no real say in the matter and I'm certainly not blocking anything. It's not something that I'll have to deal with so I'm not going to waste my time and energy fighting over it.

TRBlount:
An idea that just occurred to me: `user_settings.json`
This would be a modular file that could theoretically be modified by either an in-game options menu or dialog in Knossos. It would serve a similar purpose to the old INIs FSO used back in the day but seem to have been fallen out of use due to Knossos handling everything with JSON.

EatThePath:
> It's another silly cmdline option in a sea of cmdline options which you need to study a wiki to figure out. 

I just want to say that with a launcher like Knossos, command line flags like this become a lot more discoverable and akin to actual options. That interface isn't as good or as obvious as it could be, but when I came back to FSO after a long time not playing it and installed it through Knossos, I enabled a ton of options that way without ever checking the wiki. It's a clunky, un-ideal way to do it, but it seems the best way for users that is available and doesn't require waiting an indeterminate amount of time for other new features.

Baezon:
If it would make it easier to make these sorts of changes, I'm totally in favor of a sort of 'user settings' file of some kind.

MageKing17:
> The problem with plopping it into a non-modular table

~~I know this is really not the point here, but why do people keep calling game_settings.tbl non- modular?~~

I don't buy the maintainability argument either; this is an incredibly straightforward change. On the off chance it actually causes problems later that (for some reason) nobody is willing to fix, it can be removed again *very* easily. It's hardly `-nohtl`, interlaced through almost every piece of graphics code, where I accidentally caused a moderate graphics bug removing it.

I also don't understand the "wrong way to do it" argument; if keeping

MageKing17:
my bad; phone misclicked

PhantomHoover:
> But it's implemented in a individual-centric manner. It's another silly cmdline option in a sea of cmdline options which you need to study a wiki to figure out.

As EatThePath noted, there is effectively an entire ad-hoc protocol in FSO to allow launchers like Knossos to query it for command line arguments and present a user-friendly list.

MitoPL:
From a player's perspective, I think that feature is a good idea. The initial reason for this absolutely and primarily is a matter of accesibility (not only for people who experience severe physical discomforts but also for people who just find the effect confusing and annoying). Also in retail you can turn off this effect with the chase view, as mentioned earlier. If you find yourself really wanting fully retail like experience, you can just not enable such a flag, it's your choice.

As for putting that option into game_settings.tbl: in all honesty, how different would usage of it be from just removing the  shudder effect from the aforementioned weapon in weapons.tbl or tweaking $Ship Recoil Modifier in ships.tbl?

Also this flag doesn't need to exist indefinitely, it might be just a workaround solution for until the current shudder effect is modernised and/or until a new sort of game configuration method is created - and even then, it'd be a very welcome option to be able to choose.

MageKing17:
> or tweaking $Ship Recoil Modifier in ships.tbl?

Recoil and shudder are completely unrelated.

MageKing17 (approved):
I can see no reason not to approve this PR.

chief1983:
Ok, so whether it can genuinely affect balance or not is beside the point.  I am still, like many other devs against the idea of adding new command line options in general, without them being really really necessary.  But that's not my only issue here.

This 'aesthetic option' as it has been described could not possibly have been more intentional by the devs.  Balance or no, there is an obviously intended _penalty_ for spamming afterburner.  Or for firing the Maxim.  Having an in game option to simply turn off this penalty just seems like such an incredibly strange thing to clutter the game with to me.  I get some people don't _like_ the effect.  THAT'S THE POINT.  It wasn't just an accident that it isn't more tolerable.  It is by design.  Turning this off is, to me, more akin to enabling a cheat code than just simply changing something aesthetically about the game.  It's not just disabling 'view bob' in an FPS.

But enough people disagree with me and are even volunteering to maintain this that I'll withdraw most of my change requests on this.  However I would still like to see this flag blocked in multiplayer.  Whether some people think this can affect balance or not, I would rather not even have to see that discussion be had among anyone who does feel like getting into multiplayer once the current efforts on that side are completed.

chief1983 (changes requested):
Please disable the effect of this flag in multiplayer so we can avoid having to have a discussion on whether this actually affects competitive balance.

chief1983 (approved):

PR #2804, opened 2020-10-16
Two fixes to initial missile physics set-up

Baezon:
Both noticed by @MageKing17 . Homing missiles normally inherit the forward component of their parent object's velocity (plus their intrinsic initial speed), which is all well and good if the missile is pointing forward, but quite nonsensical in any other direction. They should inherit the parent object's velocity *in the direction of missile firing*. 

Additionally, since inheriting this speed is default, `Use_additive_weapon_velocity` should not be double dipping on their speed!


MageKing17:
This can't affect retail balance, right? I mean, I don't think any retail ships have off-angle missile launchers, and any retail ships with missile turrets won't be moving at a speed fast enough for the effect to be noticeable, I'd think...

MageKing17 (approved):
It seems straightforward enough.

PR #2806, opened 2020-10-16
Fix #2767

### LINKED ISSUE #2767: Only one swarm or corkscrew missile will track target
Equip Tornados or Hornets, lock onto a target and fire. Only one of the missiles will track the target with the rest just going off into the distance.

Confirmed to be the result of fixing #2724.

TRBlount:
This will close #2767.

This restores the old check but adds additional conditions to ensure that even non-multilock missiles have their locks cleared upon firing.


JohnAFernandez (changes requested):
Could we just add a quick comment to it to make sure others know why those conditional has to be so complicated?

Otherwise, looks good.

JohnAFernandez (approved):
Thanks for the quick edit!  Let's get this merged

PR #2808, opened 2020-10-17
stop needs more space so as not to crash during really busy scenes

JohnAFernandez:
Puts a bit more space in the object update packet for the "Stop".  It wasn't properly accounted for previously and would eventually crash in busy missions.


JohnAFernandez:
I should mention, this is safe to merge without a multi bump, because this section of the code is how it determines how many ships to send in one packet.

asarium (approved):

PR #2809, opened 2020-10-17
Change graphics interface to function objects

asarium:
This will make it easier to have a graphics interface without relying on
global objects since the object can be kept alive in the function
object. No other functional changes since the existing syntax is also
valid for function objects.

### DIFF
--- a/code/graphics/2d.h
+++ b/code/graphics/2d.h
@@ -641,207 +641,260 @@ enum class BufferUsageHint { Static, Dynamic, Streaming, PersistentMapping };
 typedef void* gr_sync;
 
 typedef struct screen {
-	uint	signature;			// changes when mode or palette or width or height changes
-	int	max_w, max_h;		// Width and height
-	int max_w_unscaled, max_h_unscaled;
-	int max_w_unscaled_zoomed, max_h_unscaled_zoomed;
-	int center_w, center_h;	// Width and height of center monitor
-	int center_offset_x, center_offset_y;
-	int	save_max_w, save_max_h;		// Width and height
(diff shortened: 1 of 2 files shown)


code/graphics/2d.h
```diff
@@ -1039,112 +1092,120 @@ inline void gr_post_process_restore_zbuffer() {
 
 __inline void gr_render_primitives(material* material_info, primitive_type prim_type, vertex_layout* layout, int vert_offset, int n_verts, int buffer_handle = -1, size_t buffer_offset = 0)
 {
-	(*gr_screen.gf_render_primitives)(material_info, prim_type, layout, vert_offset, n_verts, buffer_handle, buffer_offset);
+	gr_screen
+		.gf_render_primitives(material_info, prim_type, layout, vert_offset, n_verts, buffer_handle, buffer_offset);
```
MageKing17, line 1096: why line break?

asarium: Fixed.

wookieejedi:
Ah this will be very useful in cleaning up the global namespace regarding drawing scripts.

Goober5000:
I have read that `std::function` carries a lot more overhead than a function pointer.  Is this going to increase the memory requirements or slow down the execution speed of the graphics functions?

asarium:
I did some research before this and it appears that for optimized code a `std::function` uses around 5 instructions for a call as opposed to 3 for a "normal" function pointer.
Given the improvements in maintainability and extensibility, I think that is a fair trade-off.

Goober5000:
Ok.

Goober5000 (approved):

PR #2810, opened 2020-10-18
Improved collision avoidance for fighters

Baezon:
Tired of enemies afterburning directly into the broad side of an Orion? Not anymore! 

Adds the "$better combat collision avoidance for fightercraft:" flag. While fighting, and only while fighting, (that is, `AIM_CHASE`) small ships will project their velocity outward to predict collisions and avoid colliding with big or huge ships that it finds, before executing any of its submodes. This is sort of collision detection is usually only done while in an attack submode, and only along the me-to-my-target vector, making it only useful to circumvent intervening obstacles. For this reason (and probably also for now irrelevant performance reasons), the timings used by the function are relatively long, so a `time_scale` parameter was added so it could be used much more frequently by the general combat collision avoidance. 1.5 seconds, the default time to recheck if no collision was found, is more than enough time to go from 'a comfortable distance from the destroyer' to a collision, so the time scale has been set for it to check ten times more frequently.

Additionally, since this system is not perfect and collisions will still occasionally occur, ships in combat will now spend far less time recovering from it. Rather than up to 45 seconds (!!!) and usually making them utterly useless and a sitting duck, these ships will only spend two seconds recovering, instead of leisurely waltzing to the ship's bounding box.

One small change that is *not* covered by the flag, and will affect retail is that the recovery system was slightly altered. Normally after a collision the ship will project a vector 100m away from the collision to go to for phase 1, and then the closest bbox point for phase 2 (spending at most 15 and 30 seconds on each phase). I've modified it slightly so that phase 1 generates a relative direction to travel in, rather than a global position, and since it can't actually reach this point now the time was reduced to 5 seconds (an average case for phase 1). For high velocity collisions its entirely possible the ship can end up several hundred meters from its original position by the time it regains control, so it may end up spending phase 1 uselessly traveling basically back to the original collision point rather than actually moving away from the collision.

This was done because it works very well in tandem with the other changes, but is hard to reconcile with the retail handling while also being very minor and unlikely to break anything.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -7056,22 +7056,25 @@ void mabs_pick_goal_point(object *objp, object *big_objp, vec3d *collision_point
 /**
  * Return true if a large ship is being ignored.
  */
-int maybe_avoid_big_ship(object *objp, object *ignore_objp, ai_info *aip, vec3d *goal_point, float delta_time)
+int maybe_avoid_big_ship(object *objp, object *ignore_objp, ai_info *aip, vec3d *goal_point, float delta_time, float time_scale = 1.f)
 {
 	if (timestamp_elapsed(aip->avoid_check_timestamp)) {
 		float		distance;
 		vec3d	collision_point;
 		int		ship_num;
+		int next_check_time;
 		if ((ship_num = will_collide_with_big_ship_all(Pl_objp, ignore_objp, goal_point, &collision_point, &distance, delta_time)) != -1) {
 			aip->ai_flags.set(AI::AI_Flags::Avoiding_big_ship);
 			mabs_pick_goal_point(objp, &Objects[ship_num], &collision_point, &aip->avoid_goal_point);
 			float dist = vm_vec_dist_quick(&aip->avoid_goal_point, &objp->pos);
-			aip->avoid_check_timestamp = timestamp(2000 + MIN(1000, (int) (dist * 2.0f)));	//	Delay until check again is based on distance to avoid point.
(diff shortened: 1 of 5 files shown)


wookieejedi:
I’ll also note that I’ve been testing the branch for the past two days and turning on the updated collision even makes for immersion improved since the AI is behaving a lot more competently.

wookieejedi (approved):
Definitely want to give some other folks time to look at it, but through code evaluation and personal testing everything worked well.

code/ai/aicode.cpp
```diff
@@ -13892,29 +13912,41 @@ int maybe_big_ship_collide_recover_frame(object *objp, ai_info *aip)
 {
 	float	dot, dist;
 	vec3d	v2g;
+
+	bool better_collision_avoid_and_fighting = The_mission.ai_profile->flags[AI::Profile_Flags::Better_collision_avoidance] && aip->mode == AIM_CHASE;
+
+	// if this guy's in battle he doesn't have the time to spend up to 35 seconds recovering!
+	int phase1_time = better_collision_avoid_and_fighting ? 2 : 5;
```
JohnAFernandez, line 13919: Might be good to expose this value at some point and make it settable via the ai table.  Maybe within the scope of this PR?

Baezon: I'm generally averse to adding profile options if the code can handle it properly in general for all cases (obviously exempting the very flag I'm adding), which is the intention of the values chosen here as well as with `collision_avoidance_aggression`. The only use case I can think of for exposing it would be to choose unusual values to fit unusual ships, which this *should* be able to handle, even as is. I am more than willing to re-examine this code to improve and generalize it further later, but once things go into the tables they rarely come out.

PR #2814, opened 2020-10-19
Fix missing background color in paletted HUD gauges

### LINKED ISSUE #2811: Bug: Hud has lost background color
Hello;

Sometime between 3.8.0 Final and 19.0.0 the code responsible for drawing the HUD gauges transparency has changed, resulting in the loss of colored background for all gauges.

If I had to guess I think this only affects paletted .ANI gauges (e.g. all retail gauges), as they define their inner section with the color blue. This is probably being mapped to blank color rather than semi transparent dark color causing this issue.

I think this hurts readability of the HUD especially around nebulas and busy space backdrops, and of course looks pretty different from retail.

This affects the hud graphics from both FS1 and FS2 retail vps.

The effects are much more visible if you view these images on a dark background than on a white webpage:

3.8.0:
fso_hud_3_8_0

19.0.0:
fso_hud_19_0_0

bangstk:
Fixes #2811

Code was moved around between 3.8.0 and 19.0.0 which broke the colored backgrounds of paletted HUD gauges.

Palette index 15 is used by the HUD gauge graphics to indicate background area of HUD gauges. When the code to remap palette indexes to alpha values was moved to packunpack.cpp in Commit 72a797f, palette index 15 was mistakenly mapped to alpha value 0 (nothing) rather than the alpha that palette index 1 would get (which is what retail/3.8.0 code does) causing all HUD gauges to be missing their dark transparent backgrounds.

Also: the code maps palette indexes [0, 15] to alpha values [0, 255]. But due to an oversight (palette index 15 not being part of this gradient and being mapped to a lower value) alpha value 255 is never used so the range of alpha values for HUD gauges ended up being [0, 238], never reaching full white/fully opaque. Conversion math is now tweaked to make the alpha range [0, 252] which is much closer to [0, 255] while still not needing to do multiplication with floats

### DIFF
--- a/code/anim/packunpack.cpp
+++ b/code/anim/packunpack.cpp
@@ -151,14 +151,21 @@ int unpack_pixel(anim_instance *ai, ubyte *data, ubyte pix, int aabitmap, int bp
 			bit_16 = (ushort)pix;
 			break;
 		case 8:
-			// 8 bit-per-pixel aa bitmaps are a bit special since they only use value in the range [0, 15] where 15 wraps
-			// around back to 0. Since the rest of the code expects the value to be in the range [0, 255] the pixel value
-			// needs to be adjusted here. By muliplying the value with 17 the original range [0, 15] is mapped to [0, 255]
-			if (pix >= 15) {
-				bit_8 = 0;
-			} else {
-				bit_8 = (ubyte)(pix * 17);
+			// 8 bit-per-pixel aa bitmaps are a bit special since they only use a palette index value in the range [0, 15]. These 
+			// palette indexes must be remapped to alpha values between [0, 255] which is what graphics code expects. Palette 
+			// range [0, 14] is a gradient from black to white, and palette index 15 is a special color which indicates the background
+			// area of a HUD gauge. Retail code uses the final alpha value for index 1 for this special index to give gauges a dark
+			// transparent background.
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
Good catch and detective work. I'll give some time for other folks who might want to look before merging, too.

JohnAFernandez:
@bangstk Just want to see before we merge, can we get a before and after screenshot?

bangstk:
I can post some later tonight, but with the fix it looks like the 3.8.0 pic from the issue the original request links to. Note the green background of the radar that is missing in the 19.0.0 pic

bangstk:
Ok, here's some full screenshots.

Old broken 19.0.0+ HUD (as of Nightly 20201014):
screen0066

HUD as of this pull request:
screen0065

Expected results:
-Colored backgrounds back in all gauges. Note in particular the radar, target view, or weapon list gauges. Seen as a shade of dark green in the fixed versions (and in retail game/FSOpen on 3.8.0 and older) and flat black in the 19.0.0 bugged version.
-Slightly brighter HUD elements, due to using more of the full [0,255] range of alpha values instead of having the top of the range cut off. This makes the output match the graphics inside the .ani files closer just like PNG HUD graphics would, though it is a fairly minor difference overall.

PR #2815, opened 2020-10-19
Fix some inconsistencies with remote detonation

### LINKED ISSUE #2812: FR: Enable full "Remote Detonate"-functionality on "child" weapons
I've been making proof of concept for my fancier weapon ideas today and I noticed three peculiar behaviors when comes to the combination of the "Remote Detonate"-flag and the "child"-flag.

1 - If the parent weapon has the "Remote Detonate"-flag and the child has the "Remote Detonate"-flag, triggering the detonation of the parent will also rigger the detonation of the child.

2a - If the parent weapon has the "Remote Detonate"-flag and the child has the "Remote Detonate"-flag, letting the parent fly to end of its lifetime, will mark the child with the brackets associated with the "Remote Detonate"-flag but weapon cannot be detonate remotely by the player.

2b - If the partent weapon doesn't not have the "Remote Detonate"-flag but the child has it, the child will have the associated brackets but cannot be remotely detonated by the player.

In order to realize area denial weapons, such a player deployed and detonated mines, and enhance the player control over the weapons with multiple stages, I'd like to ask for a way to enable to the remote detonation of child weapons and an ability to seperate their remote detonation from that of the parent.

ps.
Scenario 1 and 2a can be observed in this video: 
https://youtu.be/sUKyCgOzlEs
the first and third missile fired fall under case 2a
the second missile falls under case 1

Baezon:
Fixes #2812 , and makes remote detonation more robust and flexible in general. The retail handling is pretty crude and does a bunch of futzing about with the missile's objnum and signature, presumably to avoid having to loop through all the missiles in the mission... except that it needs to do this anyway because of dual fire. Now it handles it entirely like dual fire, just loop through all the missiles, if they're mine, a remote detonator, and weren't spawned this frame (preventing premature detonation of child remote detonators being produced by this very loop), then detonate it.

Also puts brackets around all of your remote detonators, retail *did* explicitly comment that only one bracket was on purpose which make sense for dual fire, but when considering things like child spawned remote detonators, it becomes clear that bracketing all of them is important to maintain vital player feedback.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -6343,8 +6343,7 @@ void ship_weapon::clear()
     detonate_weapon_time = 0;
     ai_class = 0;
 
-    last_fired_weapon_index = -1;
-    last_fired_weapon_signature = -1;
+	remote_detonaters_active = 0;
 }
 
 ship_weapon::ship_weapon() {
@@ -6969,8 +6968,7 @@ static int subsys_set(int objnum, int ignore_subsys_info)
 			}
 		}
 
-		ship_system->weapons.last_fired_weapon_index = -1;
-		ship_system->weapons.last_fired_weapon_signature = -1;
+		ship_system->weapons.remote_detonaters_active = 0;
 		ship_system->weapons.detonate_weapon_time = -1;
 		ship_system->weapons.ai_class = sinfo->ai_class;  // assume ai class of ship for turret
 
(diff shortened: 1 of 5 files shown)


code/ship/ship.cpp
```diff
@@ -11897,92 +11895,35 @@ void ship_process_targeting_lasers()
...
+			for ( mo = GET_FIRST(&Missile_obj_list); mo != END_OF_LIST(&Missile_obj_list); mo = GET_NEXT(mo) ) {
+				object	*mobjp;
+				Assert(mo->objnum >= 0 && mo->objnum < MAX_OBJECTS);
+				mobjp = &Objects[mo->objnum];
+				if (mobjp->parent_sig == obj->parent_sig && Weapon_info[Weapons[mobjp->instance].weapon_info_index].wi_flags[Weapon::Info_Flags::Remote]) {
+					// dont detonate if this guy just spawned i.e. we just spawned him in a previous iteration of this loop
+					if (Missiontime - Weapons[mobjp->instance].creation_time > fl2f(0.01)) {
+						weapon_detonate(mobjp);
+						swp->remote_detonaters_active--;
 					}
```
Goober5000, line 11918: Why not check the `detonate_weapon_time` field here?

Even you leave the check as-is, I would still change it to 0.1 or 0.2 seconds.  A 120 FPS game could trigger twice within 0.01 seconds.

Baezon: As the comment says, this isn't catching possible retriggers from the player, that's caught above this with `detonate_weapon_time`. This is for not detonating weapons which have been created by this very loop in this frame because `weapon_detonate` is possibly going to create more `missile_objects`. Really, it should probably 0.

Goober5000: Ah, I missed the `if (timestamp_elapsed(swp->detonate_weapon_time))` check at the top.

Ok, I guess this looks gold.

Goober5000 (approved):

Issue #2412, opened 2020-05-24
Proposal: Slash beams be free


EatThePath:
I have a variety of outlandish ideas for things I want to do with slash beams that are currently impossible. The first step in all of them is allowing slash beams that sweep along arbitrary random planes rather using the octant sweeping of traditional slash beams.

This should be something I can implement myself, and looking at the code it seems like besides parsing and storing new table values all the changes may be contained within beam_get_binfo in beam.cpp. That said it seems prudent to put my thoughts and plans up for general review before going off half-cocked. I’d appreciate any direction or recommendations people may have

Obviously overwriting the existing behavior is a non-starter for compatibility reasons. Routes evident to me are adding a beam type 5, a "free slash" flag, or switching to the free slashing if any of the controlling values are set. My preferred option is type 5 beams.

The basic logic I envision is as follows:

* Determine the angular width of the slash. This will be the angular width of the target's diameter from the beam origin's point of view, multiplied by the miss factor(and accuracy?) of the beam.
* Pick a point for the center of the slash. Either a point on the target picked by the same logic used by type 0 beams, or the center of the target. I'll call this the aim point.
* Pick a random plane including both the beam origin and the aim point.
* Pick slash points on that plane satisfying the angular width.

Further, I can imagine a number of useful table options to tune this behavior. 
* Aim style: Center or type0. Obviously a default needs to be picked, but it may be useful to let a modder decide which of these two options are preferable to them.
* Min and max aim point jitter: Add a random offset to the aim point, either to make it look more 'natural' or just make it less effective. Possible creative use with high values. Defined as a factor of target angular width, just like the slash. Possibly should be configurable based on difficulty like accuracy?
* min and max sweep offset: Allow the sweep to be distributed unevenly, so that instead of always sweeping through aim + x to aim - x degrees the beam can sweep between aim+x+y to aim-x+y. Defined in multiples of sweep half-angle. Default is none, but extreme values could create interesting effects, like creating a cage of beams around a target without actually hitting it. 
* Sweep offset bias: possible values of none, start, or end. Together with the above, lets you if sweep is always offset towards or away from the start of the sweep
* Start and end pause: Let the beam spend time holding the begining or ending posiion before slashing. This seems the most ambitious and most niche. A skim of the code suggests it'd need changes in places none of the rest of this need to touch, so probably best cut and implemented in a seperate pass if at all. Listed for completeness anyway.

EatThePath:
[...]

Issue #2313, opened 2020-03-30
Feature request: Target-independant beam slash.


EatThePath:
I would like to be able to specify a fixed angle to sweep through for slashbeams, overriding the default dependence on target radius. This would among other things allow for spawned slashbeams without targets to work like slash beams should rather than being indistinguishable from regular beams.

MageKing17:
And the slash angle would just be random, then?

EatThePath:
@MageKing17 as in the orientation of the plane it sweeps through? I suppose so. I assumed it already was, honestly.

MageKing17:
It's semi-random; it uses a randomly-chosen "octant pair" from the model, chosen from a list of "good octant pairs". It's not a terribly complicated method for choosing coordinates for the beam to sweep through (since it's supposed to be fast), but it *does* require a model to be targeted, so we need some other way of generating points to slash between; "pick a random angle" is the most obvious answer, but I wanted to confirm that's what was wanted.

asarium:
Maybe just specify two point it will slash through. If I remember the internal system correctly that is how slashing beams work already.

MageKing17:
...Right, we already implicitly passed that stage; we're talking about a *way* to specify two points when spawning subweapons.

JohnAFernandez:
Fulfilled, woohoo!

Issue #2304, opened 2020-03-19
Feature request: Target-independant beam deviation.


EatThePath:
As I understand the current mechanics, beams are designed to have the same chance to miss targets of all sizes, by way of their firing angle being multiplied by the radius of their target. While this has utility, it would be very nice to have the option to give beams a cone of fire in absolute terms, allowing for beams that can miss small targets without their field of fire ballooning wildly against big targets.

JohnAFernandez:
Fulfilled!  Woohoo!

PR #2817, opened 2020-10-21
Add Type 5 beams

Baezon:
These beams are a highly customizable, flexible version of... every other beam, and then some. They work on the basic framework of slash beams (with lots of parameters to define the start and end points of the slash) plus the new ability to rotate over its lifetime around another parameterized axis. Works with every form of beam firing, fighter-fired, turret-fired, missile-spawned, and (~once I test to make sure it works properly~ it does :) ) subspace strike beams.

~~Still needs a bit of clean up and commenting~~, plus any bug squashing that I might find while assembling a cool showreel to show off some of the things you can do with these.

~~Explicitly depends on #2775 so if someone could review it, I would appreciate it.~~

Implements #2304 #2313 and most of #2412


code/weapon/beam.h
```diff
@@ -45,9 +43,10 @@ struct vec3d;
 // uses to define beam behavior ahead of time - needed for multiplayer
 typedef struct beam_info {
 	vec3d			dir_a, dir_b;						// direction vectors for beams	
-	float				delta_ang;							// angle between dir_a and dir_b
-	ubyte				shot_count;							// # of shots	
-	float				shot_aim[MAX_BEAM_SHOTS];		// accuracy. this is a constant multiple of radius. anything < 1.0 will guarantee a hit	
+	float			delta_ang;							// angle between dir_a and dir_b
+	ubyte			shot_count;							// # of shots	
+	float			shot_aim[MAX_BEAM_SHOTS];		// accuracy. this is a constant multiple of radius. anything < 1.0 will guarantee a hit
+	vec3d           rot_axis;
```
notimaginative, line 49: This breaks multi. The `beam_info` struct is sent in it's entirety so any change to it will break compatibility. That is fixed in the multi_packet_fixes PR though. If you can do this in another way, even if it's temporary, then it can be added back safely as part of multi_packet_fixes.

code/weapon/beam.h
```diff
@@ -73,12 +72,16 @@ typedef struct beam_fire_info {
 	int point;									// for fighters, which point on the bank it is from
 	int bfi_flags;
 	char team;									// for floating beams, determines which team the beam is on
+	int burst_seed;								// used for sharing random targets if part of the same burst
+	float current_per_shot_rot;                         // for type 5 beams
+	float per_burst_rotation;                         // for type 5 beams
+	int burst_index;
```
notimaginative, line 78: Does this need to be accounted for in multi? I assume that `burst_seed` doesn't (same as other PR), but `burst_index` and `per_burst_rotation` look like they might. It doesn't break anything in this case but I'd need to account for them in the packet fixes PR in order for type F beams to work properly in multi.

And `current_per_shot_rot` doesn't appear to be used anywhere.

notimaginative (changes requested):

code/weapon/beam.h
Baezon: There's very little that can be done to avoid sending rot_axis, for the same reasons that dir_a and b need to be sent. The best I can afford is that several other things here in b_info don't actually need to be need to be sent, `delta_ang` at the very least.

In the end, I don't really mind waiting, as I suspected this would require some sort of multi adjustment, although I didn't quite know the nature of it.

code/weapon/beam.h
Baezon: I think you might be right, on second thought I think I was misunderstanding how `beam_info` is actually used. And yes that bit is vestigal, thank you for pointing that out.

notimaginative: `beam_fire_info` isn't properly accounted for in multi right now but I'm working to fix that. As things stand now the less general a beam is the less likely it is to produce the same result in multi as it does in single-player. We've tested that pretty much ever current beam functionality is working in multi now (with the fixes) so I just want to continue that trend.

code/weapon/beam.h
notimaginative: Well if you don't mind waiting a bit then that solves the problem. I can go ahead and add `rot_axis` as part of the packet fixes and you can just make use it. Then everything should "just work".

code/weapon/beam.cpp
```diff
@@ -2152,6 +2224,229 @@ void beam_get_binfo(beam *b, float accuracy, int num_shots)
...
+			if (bwi->bpi.end_pos != POS_CENTER)
+				static_rand_cone(rand32(), &pos1, &center, 0.f, 180.f);
+
+			if (bwi->bpi.no_translate || bwi->bpi.end_pos == POS_SAME_RANDOM)
+				pos2 = pos1;
+
+			pos1 *= 300.f;
+			pos2 *= 300.f;
+			vec3d move_forward;
+			vm_vec_make(&move_forward, 0.f, 0.f, 800.f);
```
EatThePath, line 2321: > // if we have no target let's act as though we're shooting at something with a 300m radius 800m away

Due to the way the beams are configured I think it will lead to a lot less user confusion down the like if the distance and radius values of the 'virtual target' in these cases are equal

EatThePath (changes requested):
We discussed this privately after I did a lot of testing of the branch, but I'm sticking it in here so it doesn't slip either of our memories before the branch is merged

notimaginative (approved):

Baezon:
FYI, the multi parts have been approved but a review for the rest of it is still needed.

code/weapon/weapon.h
```diff
@@ -176,6 +178,42 @@ typedef struct beam_weapon_section_info {
...
+	Type5BeamPos end_pos;               // same as above but but with an extra 'same random as start' option
+	vec3d start_pos_offset;                      // position simply added to start pos (possibly manipulated by the bools below)
+	vec3d end_pos_offset;                        // position simply added to end pos (possibly manipulated by the bools below)
+	vec3d start_pos_rand;                        // same as above but a randomly chosen between defined value for each axis and its negative
+	vec3d end_pos_rand;
+	bool target_orient_positions;                // if true, offsets are oriented relative to the target, else the shooter's pov
+	bool target_scale_positions;                 // if true, offsets are scaled by target radius, else its a fixed span from the shooters pov
+	                                             // regardless of distance
+	float continuous_rot;                        // radians per sec rotation over beam lifetime
+	Type5BeamRotAxis continuous_rot_axis;   // axis to do ^^ on
```
JohnAFernandez, line 209: Pedantic request, please change

Baezon: Err, change what? Be more specific with the comment, you mean?

JohnAFernandez (changes requested):

JohnAFernandez:
Honestly, there's so much vector math, it's hard to make sure there isn't anything wrong.

code/weapon/beam.h
```diff
@@ -26,14 +26,12 @@ struct beam_weapon_info;
 struct vec3d;
 
 // beam types
-// REMINDER : if you change the behavior of any of these beam types, make sure to update their "cones" of possible
-// movement inside of the function beam_get_cone_dot(...) in beam.cpp  Otherwise it could cause collisions to not
-// function properly!!!!!!
```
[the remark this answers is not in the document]
Baezon: Yes, that function was removed awhile ago during some cleanup.

code/ai/aiturret.cpp
```diff
@@ -1785,8 +1785,17 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum,
...
 
-		// set next fire timestamp for the turret
+		// grab and set some burst data before turret_set_next_fire_timestamp wipes it
 		int old_burst_seed = swp->burst_seed[weapon_num];
+		int old_burst_counter = swp->burst_counter[weapon_num];
+		// only used by type 5 beams
+		if (turret->weapons.burst_counter[weapon_num] == 0) {
+			swp->per_burst_rot += wip->b_info.t5info.per_burst_rot;
+			if (swp->per_burst_rot > PI2)
+				swp->per_burst_rot -= PI2;
```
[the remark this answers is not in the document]
Baezon: Decided to keep this as is, but limit the input to -+360 degrees (and also make it wrap from below as a result)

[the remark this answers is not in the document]
JohnAFernandez: Perfect.

JohnAFernandez (approved):

PR #2819, opened 2020-10-21
Address issues that can arise if psnet fails to init properly

notimaginative:


JohnAFernandez (approved):
I honestly don't fully understand the changes, but I trust you do, and we'll see soon enough if anything is borked.  I do like the change, too, on freespace.cpp:6416 though. If I'm understanding that right, it should make that particular situation less mysterious, since the code usually hits an assert later.

notimaginative:
This came about after diagnosing some weird issues with one of the chief's standalones. Technically it's overkill, with 3 different fixes for the same basic issue, but I didn't want anyone to have to deal with it later.

JohnAFernandez:
All right, let's see what happens

PR #2822, opened 2020-10-22
Fix a bug with hud brackets during target deathroll

Baezon:
For some reason the entire block that draws brackets, not just for your target, but for message senders, asteroids, remote detonated missiles, and the rest, is gated by your target not being in deathroll, so while it is, you get no brackets on anything. This fixes it so it only excludes your target's brackets.

### DIFF
--- a/code/hud/hudbrackets.cpp
+++ b/code/hud/hudbrackets.cpp
@@ -395,99 +395,100 @@ void HudGaugeBrackets::renderObjectBrackets(object *targetp, color *clr, int w_c
 	int bound_rc;
 	SCP_list<CJumpNode>::iterator jnp;
 
-	if ( Player->target_is_dying <= 0 ) {
-		int modelnum;
+	if (Player->target_is_dying > 0 && targetp == &Objects[Player_ai->target_objnum])
+		return;
+	
+	int modelnum;
 
-		switch ( targetp->type ) {
-		case OBJ_SHIP:
-			modelnum = Ship_info[Ships[targetp->instance].ship_info_index].model_num;
+	switch ( targetp->type ) {
+	case OBJ_SHIP:
+		modelnum = Ship_info[Ships[targetp->instance].ship_info_index].model_num;
+		bound_rc = model_find_2d_bound_min( modelnum, &targetp->orient, &targetp->pos,&x1,&y1,&x2,&y2 );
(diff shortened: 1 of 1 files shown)


code/hud/hudbrackets.cpp
```diff
@@ -395,99 +395,100 @@ void HudGaugeBrackets::renderObjectBrackets(object *targetp, color *clr, int w_c
...
-			if (modelnum != -1)
-				bound_rc = model_find_2d_bound_min( modelnum, &targetp->orient, &targetp->pos,&x1,&y1,&x2,&y2 );
-			else {
-				vertex vtx;
-				g3_rotate_vertex(&vtx,&targetp->pos);
-				g3_project_vertex(&vtx);
-				x1 = x2 = (int) vtx.screen.xyw.x;
-				y1 = y2 = (int) vtx.screen.xyw.y;
-			}
+	if (Player->target_is_dying > 0 && targetp == &Objects[Player_ai->target_objnum])
```
JohnAFernandez, line 398: So this is the logic change, right?  It looks like the order that the code is executed has changed?  Not sure what's going on.

Baezon: Yeah its kind of confusing look at it in github. That is correct, the initial similar if is replaced with this which does an early return, the rest of the function was un-indented since it no longer in an if statement.

Baezon: I guess it probably would've been cleaner if just added `|| targetp == &Objects[Player_ai->target_objnum]` into the old `if` ...

code/hud/hudbrackets.cpp
```diff
@@ -395,7 +395,7 @@ void HudGaugeBrackets::renderObjectBrackets(object *targetp, color *clr, int w_c
 	int bound_rc;
 	SCP_list<CJumpNode>::iterator jnp;
 
-	if ( Player->target_is_dying <= 0 ) {
+	if ( Player->target_is_dying <= 0 || targetp != &Objects[Player_ai->target_objnum]) {
```
JohnAFernandez, line 398: Is there some guarantee that Player_ai->target_objnum won't be -1 or otherwise invalid?

JohnAFernandez (approved):
LGTM

PR #2823, opened 2020-10-22
Allow for continuously spawning weapons

Baezon:
Adds `$Spawn Interval` for weapons, doing their spawn event every time that amount of time passes, plus `$Spawn Chance`  for some randomization and `$Spawn Effect` since it can look a bit weird just spawning them naked.

~~This... probably breaks multi. As with type 5 beams, I'm more than willing to wait, but I'm not really sure what it's doing or what its implications for multi would be. I've commented the multi-relevant sections.~~ Multi issues worked out, due to assistance from Cyborg and taylor.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -5888,7 +5888,6 @@ int weapon_create( vec3d * pos, matrix * porient, int weapon_type, int parent_ob
  */
 void spawn_child_weapons(object *objp, int spawn_index_override)
 {
-	int	i, j;
 	int	child_id;
 	int	parent_num;
 	ushort starting_sig;
@@ -5945,7 +5944,7 @@ void spawn_child_weapons(object *objp, int spawn_index_override)
 	int num_spawn_weapons = detonate_spawn ? wip->num_spawn_weapons_defined : 2 ;
 	int start_index = detonate_spawn ? 0 : spawn_index_override;
 
-	for (i = start_index; i < num_spawn_weapons; i++)
+	for (int i = start_index; i < num_spawn_weapons; i++)
 	{
 		// don't spawn continuous spawning weapons on detonation
 		if (detonate_spawn && wip->spawn_info[i].spawn_interval > 0.f)
@@ -5964,7 +5963,7 @@ void spawn_child_weapons(object *objp, int spawn_index_override)
 			spawn_count = wip->spawn_info[i].spawn_count;
 		}
 
-		for (j = 0; j < spawn_count; j++)
+		for (int j = 0; j < spawn_count; j++)
 		{
 			int		weapon_objnum;
 			vec3d	tvec, pos;
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -601,7 +601,7 @@ void parse_wi_flags(weapon_info *weaponp, flagset<Weapon::Info_Flags> preset_wi_
                     name_length = num_start - temp_string.get() - skip_length;
                 }
 
-                weaponp->total_children_spawned += weaponp->spawn_info[weaponp->num_spawn_weapons_defined].spawn_count;
+                weaponp->total_detonation_children_spawned += weaponp->spawn_info[weaponp->num_spawn_weapons_defined].spawn_count;
 
                 Spawn_names[Num_spawn_types] = vm_strndup(&flag_text[skip_length], name_length);
                 Num_spawn_types++;
@@ -2195,6 +2195,9 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		if (num_spawn_intervals_defined < MAX_SPAWN_TYPES_PER_WEAPON)
 		{
 			wip->spawn_info[num_spawn_intervals_defined].spawn_interval = dum_float;
+			// continuous spawn weapons dont count towards total_detonation_children_spawned
+			if (dum_float > 0)
+				wip->total_detonation_children_spawned -= wip->spawn_info[num_spawn_intervals_defined].spawn_count;
 			num_spawn_intervals_defined++;
 		}
 	}
@@ -5650,7 +5653,7 @@ int weapon_create( vec3d * pos, matrix * porient, int weapon_type, int parent_ob
 			// for weapons that respawn, add the number of respawnable weapons to the net signature pool
 			// to reserve N signatures for the spawned weapons
 			if ( wip->wi_flags[Weapon::Info_Flags::Spawn] ){
-                multi_set_network_signature( (ushort)(Objects[objnum].net_signature + wip->total_children_spawned), MULTI_SIG_NON_PERMANENT );
+                multi_set_network_signature( (ushort)(Objects[objnum].net_signature + wip->total_detonation_children_spawned), MULTI_SIG_NON_PERMANENT );
 			} 
 		} else {
 			Objects[objnum].net_signature = multi_assign_network_signature( MULTI_SIG_NON_PERMANENT );
@@ -8048,7 +8051,7 @@ void weapon_info::reset()
 	this->WeaponMinRange = 0.0f;
 
 	this->num_spawn_weapons_defined = 0;
-	this->total_children_spawned = 0;
+	this->total_detonation_children_spawned = 0;
 	for (i = 0; i < MAX_SPAWN_TYPES_PER_WEAPON; i++)
 	{
 		this->spawn_info[i].spawn_type = -1;
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -601,7 +601,7 @@ void parse_wi_flags(weapon_info *weaponp, flagset<Weapon::Info_Flags> preset_wi_
                     name_length = num_start - temp_string.get() - skip_length;
                 }
 
-                weaponp->total_detonation_children_spawned += weaponp->spawn_info[weaponp->num_spawn_weapons_defined].spawn_count;
+                weaponp->maximum_children_spawned += weaponp->spawn_info[weaponp->num_spawn_weapons_defined].spawn_count;
 
                 Spawn_names[Num_spawn_types] = vm_strndup(&flag_text[skip_length], name_length);
                 Num_spawn_types++;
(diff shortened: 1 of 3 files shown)


code/weapon/weapons.cpp
```diff
@@ -5579,7 +5651,7 @@ int weapon_create( vec3d * pos, matrix * porient, int weapon_type, int parent_ob
 			// to reserve N signatures for the spawned weapons
 			if ( wip->wi_flags[Weapon::Info_Flags::Spawn] ){
                 multi_set_network_signature( (ushort)(Objects[objnum].net_signature + wip->total_children_spawned), MULTI_SIG_NON_PERMANENT );
```
Baezon, line 5653: On weapon creation it 'reserves' net signatures for its children.

code/weapon/weapons.cpp
```diff
@@ -5854,20 +5937,32 @@ void spawn_child_weapons(object *objp)
 	parent_shipp = &Ships[Objects[objp->parent].instance];
 	starting_sig = 0;
 
-	if ( Game_mode & GM_MULTIPLAYER ) {		
+	if ( Game_mode & GM_MULTIPLAYER ) {	 	
 		// get the next network signature and save it.  Set the next usable network signature to be
 		// the passed in objects signature + 1.  We "reserved" N of these slots when we created objp
 		// for it's spawned children.
 		starting_sig = multi_get_next_network_signature( MULTI_SIG_NON_PERMANENT );
 		multi_set_network_signature( objp->net_signature, MULTI_SIG_NON_PERMANENT );
```
Baezon, line 5945: At the start of child spawning, it does... something with the signatures.

code/weapon/weapons.cpp
```diff
@@ -5949,9 +6044,9 @@ void spawn_child_weapons(object *objp)
 	}
 
 	// in multiplayer, reset the next network signature to the one that was saved.
-	if ( Game_mode & GM_MULTIPLAYER ){
+	if ( Game_mode & GM_MULTIPLAYER ){ 
 		multi_set_network_signature( starting_sig, MULTI_SIG_NON_PERMANENT );
-	}
```
Baezon, line 5954: And at the end resets it to how it was before I guess.

code/weapon/weapons.cpp
JohnAFernandez: Yes, this would be the problem.  You would have to find a way for it to calculate how many children it is going to spawn.

Baezon: It's variable, there is no guarantee for any specific number, and that's fundamental to this mechanic.

Baezon:
After some discussion with @JohnAFernandez I've updated the `total_children_spawned` to `maximum_children_spawned`. This doesn't take into account `spawn_chance` (it just assumes it'll always spawn) or premature destruction of the spawner so this could be a significant overestimate, but I've been assured this shouldn't be an issue. I'm still not sure how exactly this system assigns the signatures, and if that's still valid.

notimaginative:
Overestimating is fine. Underestimating is bad bad bad. The net sigs will just keep wrapping around their allowable range so you really shouldn't have to worry about using them up. The problem is if you end up with multiple objects in a scene with the same sig. That's when bad things happen.

I would advise to at least put a `Warning()` during parse if the spawn count passes some soft limit. A spawn count of 100 isn't a big deal, until there are 20 ships firing that weapon at the same time.

code/weapon/weapons.cpp
```diff
@@ -2186,6 +2186,45 @@ int parse_weapon(int subtype, bool replace, const char *filename)
...
+
+	while (optional_string("$Spawn Interval:"))
+	{
+		stuff_float(&dum_float);
+
+		if (num_spawn_intervals_defined < MAX_SPAWN_TYPES_PER_WEAPON)
+		{
+			wip->spawn_info[num_spawn_intervals_defined].spawn_interval = dum_float;
+			if (dum_float > 0)
+				wip->maximum_children_spawned += (int)(wip->spawn_info[num_spawn_intervals_defined].spawn_count / wip->lifetime);
```
notimaginative, line 2199: You should probably put a safety cap on `dum_float` to avoid really small values.

code/weapon/weapons.cpp
```diff
@@ -4913,6 +4952,28 @@ void weapon_process_post(object * obj, float frame_time)
 
 	wip = &Weapon_info[wp->weapon_info_index];
 
+	if (wip->wi_flags[Weapon::Info_Flags::Spawn]) {
+		for (int i = 0; i < wip->num_spawn_weapons_defined; i++) {
+			if (wip->spawn_info[i].spawn_interval > 0.f) {
+				// continuously advance next_spawn_time and spawn weapons until the time goes into the future
+				// incase we've passed multiple spawn events in the frame
+				while (timestamp_elapsed(wp->next_spawn_time[i])) {
```
notimaginative, line 4960: This scares me a bit. Handled properly it may be okay, but generally you never want `timestamp_elapsed()` in the expression of a `while()` loop. It would be much safer to use a `for()` loop with `wip->maximum_children_spawned` and then check the timestamp in that. That way you've got a guaranteed breakout point.

Even if your specific use case works okay, it sets an example that someone else may copy badly without realizing the dangers of `timestamp_elapsed()`.

notimaginative (changes requested):
Just the one main issue. Otherwise it LGTM. Nothing to really worry about multi wise with the max count issue addressed.

code/weapon/weapons.cpp
Baezon: I wasn't aware of the issues myself, is the worry that `timestamp_elapsed` will take too long or never return false? I think I'll put a 0.1 minimum interval which should be plenty fast even in the most extreme scenarios, so even in this case only very few spawns should be necessary within a frame.

Baezon:
I do worry about warning for max spawn count on parse, thinking ahead I can imagine that this could very likely be used for an absurd BFG type projectile, or others who have expressed using it like a mine or turret: long-lasting and constantly shooting projectiles. I could imagine spawn counts as high as 1000 in the most extreme but feasible scenarios. I think implicitly modders would be adverse to to filling the screen too much and would *probably* compensate by not having these weapons used often, so it seems over-aggressive to warn on parse in the worry that they'll be used by 20 ships at once. Then again, I'm not sure an over-abundance of these can be easily detected and curtailed during runtime either.

EatThePath:
> I could imagine spawn counts as high as 1000 in the most extreme but feasible scenarios. I think implicitly modders would be adverse to to filling the screen too much and would _probably_ compensate by not having these weapons used often, so it seems over-aggressive to warn on parse in the worry that they'll be used by 20 ships at once

Can confirm, I can easily see myself putting enough offending weapons in a mod to make users see the many warnings message on release builds.

code/weapon/weapons.cpp
notimaginative: `timestamp_elapsed()` isn't really dependable when used like that. It's not slow but there can be conditions where it might result in an endless loop. Your specific usage isn't too bad, but it still worries me.

@Goober5000 Can you take a look at this and tell me if I'm overreacting or not?

notimaginative:
I agree and can easily see modders making impressive use of high numbers. But modders have notoriously done really bad things too. I've written several posts about that over the years on the HLP forums. That's why I'm saying to use a soft limit and a Warning. Make it known, but don't prevent it. Some good documentation on the subject *might* be enough however.

I mentioned that you don't have to worry about net signatures, but there are just 32,767 available for all weapons in a scene. It is very possible that a modder can break multi support if they aren't careful. For some mods that wouldn't be an issue, but for those where it is I can already imagine the bug reports on this and there wouldn't be anything we can do about it.

Baezon:
I'm just not sure how to draw the distinction between a mod that is interested in maintaining multi integrity and one that isn't. If the modder isn't I can already imagine how annoyed they'll be when faced with the warning every time they start their game because they always use debug (as they should). I guess I've thought of warnings as 'something you should almost certainly fix regardless of what you're doing'. A zero MOI for a model on model read is printed, but not warned, because this isn't a problem if the model is a missile (which isn't obvious at that point yet), however merely printing is probably much too quiet if the modder *is* interested in multi.

What about on a multi mission load?

notimaginative:
A multi-only warning sounds good. Like you said, it's hard to say what too much actually is in this case, so the warning message should be more informational rather than a complaint.

But if you can think of a better way to get the message across then we can just do away with the warning. All I really want is for this to not be a surprise to anyone.

EatThePath:
>  If the modder isn't I can already imagine how annoyed they'll be when faced with the warning every time they start their game because they always use debug (as they should).

You can surpress those popups with a command line flag, and I often do, but my bigger worry is that if a mod has a lot of such weapons then every time it's launched with a release build the player will see a message that there's a lot of warnings in the mod. That's something that's going to irk modders who are doing it quite deliberately and relatively responsibly, I think.

Baezon:
Ok, added a multi warning if it finds a weapon with more than 300 (is that a good number?) spawns, and I'm pretty sure `Game_mode & GM_MULTIPLAYER` is valid on mission load. Combined with a minimum spawn interval of 0.1 seconds and only 3 attempts at spawning sub-frame spawns, it should be good on too many spawns or spawning forever.

code/weapon/weapons.cpp
```diff
@@ -225,6 +225,8 @@ const float HOMING_DEFAULT_FREE_FLIGHT_FACTOR = 0.25f;
 #define REARM_NUM_MISSILES_PER_BATCH 4              
 #define REARM_NUM_BALLISTIC_PRIMARIES_PER_BATCH 100 
 
+const float minimum_spawn_interval = 0.1f;
```
notimaginative, line 228: Please use `static` on this, and also capitalize it. Makes it easier to recognize as a global when capitalized.

notimaginative (changes requested):

notimaginative:
Thanks for making those changes! Just the one new thing and it should be good to go.

I'm not sure if 300 is too high or not high enough, but it's a good start. It can be tweaked later as needed. Assuming all of this gets documented in the wiki it's probably a good idea to make sure the multi issue gets mentioned.

notimaginative (approved):

code/weapon/weapons.cpp
Goober5000: As long as the timestamp is valid, I can't think of a situation where it would cause an endless loop.  Can you be more specific?

notimaginative: Mostly worried about wrapping. It wouldn't be an endless loop, but it would cause a delay until it catches up. That's generally a multi issue though, since it's really only standalone servers that run long enough to wrap the timestamp. If something bad and unrelated happens (memory overwrite, etc.) and the timestamp turns invalid then it might be endless. I'm probably just being overly paranoid though.

@Baezon Your changes here are okay but if you prefer the while() loop instead I guess that would be fine. It's definitely cleaner looking. If it ends up causing a problem later it can be dealt with then.

Baezon: Eh, I might just get rid of it entirely. I was thinking ahead for very small intervals but with a minimum of 0.1 seconds, having to do this more than once is very unlikely for normal gameplay, and you'll have bigger issues if your framerate is 10 or less.

Baezon:
Ok, should be ready. Made some tweaks to delay to make it make a bit more sense, and removed the subframe spawns because they aren't really necessary anymore.

PR #2830, opened 2020-10-24
Avoid resetting player_has_lock unless they've actually lost their lock

PhantomHoover:


TRBlount (approved):
This makes sense. Fixes an oversight I had made.

PR #2835, opened 2020-10-26
Better excluding of beam-weapon collisions with the same parent

Baezon:
`pair->b == b->objp` in `beam_collide_missile` (that is, `colliding_missile == beam->objp` ) used to cull collisions will never be true, `beam->objp` will always be a ship, but I've replaced it with a similar check that weapon-weapon collisions do to ensure they never collide if fired from the same parent. Beams don't have `parent_sig` so it has some looking up to do first, essentially preventing a ship's beams from shooting down its own missiles. This is an absurdly unlikely situation normally, but becomes significantly more likely with continuous spawn.

Also renames the abysmally named beam variables in the collision functions so you don't have horribly confusing looking lines like `pair->b == b->objp` (`pair->b` is the other collider, `b` is the beam).

And also removes the bizarre exclusion of beams from never being collision checked if the function returns 1. It's hard to say if this was retail since collision was completely overhauled at some point, but I couldn't find any indications that it was. I had assumed this was done because it didn't return 1 properly or had some way of invalidating a previous claim of 'never being able to collide'. But the only time these functions return 1 is when this parenting relation excludes them or it can't get a model to collide against. Both of which are perfectly valid and permanent situations.

### DIFF
--- a/code/object/objcollide.cpp
+++ b/code/object/objcollide.cpp
@@ -889,7 +889,7 @@ void obj_collide_pair(object *A, object *B)
         collision_info->next_check_time = timestamp(0);
     }
 
-    if ( valid &&  A->type != OBJ_BEAM ) {
+    if ( valid ) {
         // if this signature is valid, make the necessary checks to see if we need to collide check
         if ( collision_info->next_check_time == -1 ) {
             return;
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -2382,7 +2382,7 @@ void beam_jitter_aim(beam *b, float aim)
 // collide a beam with a ship, returns 1 if we can ignore all future collisions between the 2 objects
 int beam_collide_ship(obj_pair *pair)
 {
-	beam *b;
+	beam * a_beam;
 	object *weapon_objp;
 	object *ship_objp;
 	ship *shipp;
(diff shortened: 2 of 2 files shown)


code/object/objcollide.cpp
```diff
@@ -889,7 +889,7 @@ void obj_collide_pair(object *A, object *B)
         collision_info->next_check_time = timestamp(0);
     }
 
-    if ( valid &&  A->type != OBJ_BEAM ) {
```
JohnAFernandez, line 892: What's going on here that it's a good idea to remove this check?

Baezon: (my last paragraph in the description)

JohnAFernandez: Were you waiting for me to post this comment?

Baezon: Pure happenstance I assure you.

JohnAFernandez: Ok.  The reason for the confusion is that I don't see what you're talking about in the third paragraph in or around this line.

Baezon: This is gating the early out for general collision checking if the `next_check_time` is -1. This is done to indicate that the pair will never collide and collision can move on to the next pair immediately. However, even if a previous collision returned 1, to indicate `next_check_time` should be set to -1 and never checked again, if a beam is involved (beams are always `A`) its ignores this early out and attempts collision anyway. As mentioned I see no good reason for this to be the case. It's probably not much of a time cost, but still bizarre and as far as I can tell, unnecessary.

JohnAFernandez: Can't remember what happened after your last reply.  Might have fallen asleep or had to take care of my kid.

JohnAFernandez (approved):

PR #2836, opened 2020-10-26
Add 'Choose Random Bitmap' for particle effects

wookieejedi:
This PR adds the ability to specify a list of bitmaps for a particle effect to choose from. The selected bitmap will be randomly chosen from the list when the individual particle is created. This feature allows modders to make much more varied particle effects much more efficiently. I've tested this with a variety of single and composite effects with both lists and single filenames and everything works as expected.  

Thanks to m!m for all his help with the implementation and preliminary early review!

### DIFF
--- a/code/particle/ParticleManager.cpp
+++ b/code/particle/ParticleManager.cpp
@@ -362,7 +362,7 @@ SCP_vector<int> parseAnimationList(bool critical) {
 	
 	SCP_vector<int> handles;
 
-	for (SCP_string name: bitmap_strings) {
+	for (auto name: bitmap_strings) {
 		auto handle = bm_load_animation(name.c_str());
 		if (handle >= 0) {
 			handles.push_back(handle);
--- a/code/particle/ParticleManager.cpp
+++ b/code/particle/ParticleManager.cpp
@@ -343,28 +343,26 @@ bool required_string_if_new(const char* token, bool no_create) {
 	return true;
 }
 
-int parseAnimation(bool critical) {
-	SCP_string name;
-	stuff_string(name, F_FILESPEC);
-
-	auto handle = bm_load_animation(name.c_str());
-
-	if (handle < 0) {
-		int level = critical ? 1 : 0;
-		error_display(level, "Failed to load effect %s!", name.c_str());
-	}
-
-	return handle;
-}
-
 SCP_vector<int> parseAnimationList(bool critical) {
 
 	SCP_vector<SCP_string> bitmap_strings;
-	int num_bitmaps = stuff_string_list(bitmap_strings);
+	
+	// check to see if we are parsing a single value or list
+	ignore_white_space();
+	if (*Mp == '(') {
+		// list of names case
+		stuff_string_list(bitmap_strings);
+	}
+	else {
+		// single name case
+		SCP_string name;
+		stuff_string(name, F_FILESPEC);
+		bitmap_strings.push_back(name);
+	}
+	
 	SCP_vector<int> handles;
-	size_t tbl_idx;
 
-	for (auto i = 0; i < num_bitmaps; i++) {
+	for (auto i = 0; i < bitmap_strings.size(); i++) {
 		SCP_string name = bitmap_strings[i];
 		auto handle = bm_load_animation(name.c_str());
 		if (handle < 0) {
--- a/code/particle/ParticleManager.cpp
+++ b/code/particle/ParticleManager.cpp
@@ -356,5 +356,26 @@ int parseAnimation(bool critical) {
 
 	return handle;
 }
+
+SCP_vector<int> parseAnimationList(bool critical) {
+
+	SCP_vector<SCP_string> bitmap_strings;
+	int num_bitmaps = stuff_string_list(bitmap_strings);
+	SCP_vector<int> handles;
+	size_t tbl_idx;
+
+	for (auto i = 0; i < num_bitmaps; i++) {
+		SCP_string name = bitmap_strings[i];
+		auto handle = bm_load_animation(name.c_str());
+		if (handle < 0) {
+			int level = critical ? 1 : 0;
+			error_display(level, "Failed to load effect %s!", name.c_str());
+		}
+		handles[i] = handle;
+	}
+
+	return handles;
+}
+
 }
 }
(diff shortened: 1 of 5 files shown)


code/particle/ParticleManager.cpp
```diff
@@ -343,18 +343,37 @@ bool required_string_if_new(const char* token, bool no_create) {
...
+	SCP_vector<int> handles;
+
+	for (size_t i = 0; i < bitmap_strings.size(); i++) {
+		SCP_string name = bitmap_strings[i];
+		auto handle = bm_load_animation(name.c_str());
+		if (handle < 0) {
+			int level = critical ? 1 : 0;
+			error_display(level, "Failed to load effect %s!", name.c_str());
+		}
+		handles.push_back(handle);
```
asarium, line 372: The bitmap handle should only be added when it is valid to avoid issues with accidentally using invalid handles.

code/particle/ParticleManager.h
```diff
@@ -160,15 +160,16 @@ ParticleEffectHandle parseEffectElement(EffectType forcedType = EffectType::Inva
...
- * @brief Parses an animation
+ * @brief Parses an animation or list of animations.
  *
- * Parses and animation and returns the handle. If critical is @c true then a failure to load the animation will cause
+ * Parses an animation or list of animations and returns the handle(s). If critical is @c true then a failure to load the animation will cause
  * an error. Otherwise it will be cause a warning.
  *
  * @param critical @c true if a failure is critical
- * @return The animation handle
+ * @return The vecotr of animation handles
```
asarium, line 169:
```suggestion
 * @return The vector of animation handles
```

code/particle/effects/SingleParticleEffect.cpp
```diff
@@ -58,7 +58,9 @@ SingleParticleEffect* SingleParticleEffect::createInstance(int effectID, float m
 	Assertion(minSize >= 0.0f, "Minimum size may not be less than zero, got %f!", minSize);
 
 	auto effectPtr = new SingleParticleEffect("");
-	effectPtr->m_particleProperties.m_bitmap = effectID;
+	SCP_vector<int> bitmap_list;
+	bitmap_list.push_back(effectID);
+	effectPtr->m_particleProperties.m_bitmap_list = bitmap_list;
```
asarium, line 63: You can also just use `push_back` on the bitmap list instead:
```suggestion
	effectPtr->m_particleProperties.m_bitmap_list.push_back(effectID);
```

code/particle/util/ParticleProperties.cpp
```diff
@@ -34,8 +34,28 @@ void ParticleProperties::parse(bool nocreate) {
...
+	if (m_bitmap_list.empty())
+		return -1;
+
+	int bitmap_index;
+	int num_bitmaps = (int)m_bitmap_list.size();
+
+	if (num_bitmaps == 1) {
+		bitmap_index = m_bitmap_list[0];
+	} else if (num_bitmaps > 1) {
+		bitmap_index = m_bitmap_list[rand() % num_bitmaps];
```
asarium, line 49: `rand()` has numerous issues which is why `RandomRange` exists. There you could add a new alias that has a `size_t` as the value type.
That type can then be used in addition to `m_bitmap_list` where you initialize that to return value in the range from 0 to `m_bitmap_list.size - 1`.
Then you could also simplify this code a bit since the random range takes care of most of your cases here.

code/particle/util/ParticleProperties.cpp
```diff
@@ -34,8 +34,28 @@ void ParticleProperties::parse(bool nocreate) {
 	}
 }
 
+int ParticleProperties::chooseBitmap()
+{
+	// failsafe check, though we should not have been able to get to this point
+	if (m_bitmap_list.empty())
```
asarium, line 40: Since returning `-1` here will cause issues later down the line (since bitmap particles can not be rendered without a texture) it would be better to just use an `Assertion` here to check that the list is not empty.

code/particle/util/ParticleProperties.cpp
```diff
@@ -63,9 +83,14 @@ WeakParticlePtr ParticleProperties::createPersistentParticle(particle_info& info
 }
 
 void ParticleProperties::pageIn() {
-	if (m_bitmap >= 0) {
-		bm_page_in_aabitmap(m_bitmap, -1);
+	
+	size_t num_bitmaps = m_bitmap_list.size();
```
asarium, line 87: Could also use a range based for loop here since that automatically handles the empty case properly.

code/particle/util/ParticleProperties.h
```diff
@@ -29,6 +30,11 @@ class ParticleProperties {
 	 */
 	void parse(bool nocreate);
 
+	/**
+	 * @brief Choose particle from bitmap list
+	 */
+	int chooseBitmap();
```
asarium, line 36: This is an internal function so it should be made `private`.

asarium (changes requested):

wookieejedi:
Thanks for all comments @asarium, feedback has been incorporated!

code/particle/util/ParticleProperties.cpp
```diff
@@ -37,21 +37,11 @@ void ParticleProperties::parse(bool nocreate) {
...
-	if (num_bitmaps == 1) {
-		bitmap_index = m_bitmap_list[0];
-	} else if (num_bitmaps > 1) {
-		bitmap_index = m_bitmap_list[rand() % num_bitmaps];		
-	} else {
-		bitmap_index = -1;
-	}
+	Assertion(!m_bitmap_list.empty(), "No bitmaps found!");
+
+	size_t index = ::util::UniformRange<size_t>(0, m_bitmap_list.size() - 1).next();
```
asarium, line 42: Creating a `RandomRange` is a potentially resource intensive operation so it would be better to only create this once when the bitmap list has been loaded and then just use that instance here.

code/particle/util/ParticleProperties.cpp
```diff
@@ -37,21 +37,11 @@ void ParticleProperties::parse(bool nocreate) {
...
 
-	return bitmap_index;
+	return m_bitmap_list[(int)index];
```
asarium: Index operations can use `size_t` as their argument (it is actually the more correct type for that operation) so there is no need to have the int cast here.

asarium (changes requested):

asarium (approved):

PR #2838, opened 2020-10-27
Add "resets" burst flag

Baezon:
When enabled, let's a burst reset back to 0 if the full fire wait has elapsed, even if it was 'in the middle' of a burst.


wookieejedi (approved):
Very useful.

PR #2839, opened 2020-10-27
redo the locking animations for multilock

### LINKED ISSUE #2826: Bug: Multilock broke FS1 missile lock indicator animation
This one is just another HUD cosmetic issue. The FS1 missile lock indicator (4 spinning triangles) are supposed to spin indefinitely when locked on, while in current nightlies it gets stuck on the last frame and sits there looking like this while flashing:

fs1_lock_broke

When multilock was added, it switched the lock animation code to a new function renderLockTrianglesNew in hudlock.cpp.

There is a variable "loop_locked_anim" that is set by whether the fs1 or fs2 hud is used. I think the new "time elapsed" code in renderLockTrianglesNew is causing this if statement to never be true:
`if(Lock_anim.time_elapsed < Lock_anim.total_time)`

Therefore it never conditionally calls hud_anim_render() with the loop parameter set to 1 so the animation stops on completion instead of looping even when "loop_locked_anim" is true

Also: in retail FS1 this indicator never blinks on and off either, this could be something also controlled by "loop_locked_anim" if there is interest in making it even more accurate to FS1

Goober5000:
The multilock patch did not handle `Lock_anim` in the best way, and handled `Lock_gauge` in a very bizarre way, despite the fact that they are two simple hud animations.  This changes the `lock_info` struct to track the correct `time_elapsed` value for both, and changes the locking code to use the values in the correct way.

Additionally, this properly loops the lock triangles on the FS1 lock indicator.  Strangely, even prior to the multilock merge it appears that the triangles wouldn't have animated more than a quarter turn.

Finally, this prevents the FS1 lock indicator from blinking, since in retail FS1 it didn't blink.

Tested with FSPort and no-mods.  Fixes #2826.

### DIFF
--- a/code/hud/hudlock.cpp
+++ b/code/hud/hudlock.cpp
@@ -143,6 +143,11 @@ void HudGaugeLock::initLoopLockedAnim(bool loop)
 	loop_locked_anim = loop;
 }
 
+void HudGaugeLock::initBlinkLockedAnim(bool blink)
+{
+	blink_locked_anim = blink;
+}
+
 void HudGaugeLock::initBitmaps(char *lock_gauge_fname, char *lock_anim_fname)
 {
 	hud_anim_init(&Lock_gauge, 0, 0, lock_gauge_fname);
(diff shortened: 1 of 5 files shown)


JohnAFernandez (approved):
Afaict, it does look like a better system.

PR #2840, opened 2020-10-27
Cache bitmap lookup buffers to avoid wasteful copying

PhantomHoover:


code/bmpman/bmpman.cpp
```diff
@@ -188,6 +194,13 @@ float bitmap_lookup::get_channel_alpha(float u, float v)
 	return i2fl(Bitmap_data[(y*Width + x)*Num_channels + 3]) / 255.0f;
 }
 
+void clear_bm_lookup_cache() {
+	for(auto iter = bm_lookup_cache.begin(); iter != bm_lookup_cache.end(); iter++) {
+		free(iter->second);
+	}
```
JohnAFernandez, line 200: Looks like travis wants you to use modern loops. 

```
for(auto iter = bm_lookup_cache.begin(); iter != bm_lookup_cache.end(); iter++) {

        ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

           (auto & iter : bm_lookup_cache)
```

JohnAFernandez (approved):

code/bmpman/bmpman.cpp
TRBlount: Looks like this is fixed in the current commit?

TRBlount (approved):
Excellent work.

PR #2843, opened 2020-10-29
Fix some issues with warp collision

### LINKED ISSUE #2359: Colliding with unrendered ship during warp in
It's possible to get collide with a ship jumping into a mission even when solidly behind it's entry portal. This comes up rarely, but is very unwelcome when it does.

Baezon:
There was a lot of duplication with checks against the warp clip plane depending on if the ship was arriving or departing so I've modified the default and homeworld `getWarpOrientation` to invert when warping out as one might expect (those two are the only ones that need their orientation like this) (`getWarpOrientation` is only used for this purpose; culling stuff on the other side) and made a function to deal with handling what side a point is on, so all the cases can be handled consistently.

There were some sign errors and ordering issues with ship-weapon collisions that were fixed, and checks against the warp plane were added for several types of collisions.

This *only prevents* where otherwise it *would* have collided, if it *would not* have collided before, it will *still not* now. I figured this was the safer option, even as unlikely as this is to actually break anything. 

Closes #2359

EDIT: Also, collision detection tries not to bother too often so it calculates when the next time two ships could collide based on their speed and distance. Huge ships get a special exemption during warping since they can go way faster than normal during this time, but big ships should be added to the check as well, sine they can too.

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -1149,7 +1149,7 @@ int asteroid_check_collision(object *pasteroid, object *other_obj, vec3d *hitpos
 		vec3d actual_world_hit_pos = mc.hit_point_world + heavy_obj->pos;
 		if ((shipp->is_arriving()) && (shipp->warpin_effect))
 			warp_effect = shipp->warpin_effect;
-		else if ((shipp->flags[Ship::Ship_Flags::Depart_warp]))
+		else if ((shipp->flags[Ship::Ship_Flags::Depart_warp]) && (shipp->warpout_effect))
 			warp_effect = shipp->warpout_effect;
 
 		if (warp_effect != nullptr && point_is_clipped_by_warp(&actual_world_hit_pos, warp_effect))
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -36,6 +36,7 @@
 #include "particle/particle.h"
 #include "render/3d.h"
 #include "ship/ship.h"
+#include "ship/shipfx.h"
 #include "ship/shiphit.h"
 #include "species_defs/species_defs.h"
 #include "stats/scoring.h"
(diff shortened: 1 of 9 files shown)


code/asteroid/asteroid.cpp
```diff
@@ -1137,6 +1138,23 @@ int asteroid_check_collision(object *pasteroid, object *other_obj, vec3d *hitpos
 
 		}
 	}
+
+	// check if the hit point is beyond the clip plane if the ship is warping.
+	if (mc_ret_val) {
+		WarpEffect* warp_effect = nullptr;
+		ship* shipp = &Ships[pship_obj->instance];
+
+		// this is extremely confusing but mc.hit_point_world isn't actually in world coords
```
JohnAFernandez, line 1147: Hey, if the other variable was that poorly named then we should change it now while we're thinking about it.

code/asteroid/asteroid.cpp
```diff
@@ -1137,6 +1138,23 @@ int asteroid_check_collision(object *pasteroid, object *other_obj, vec3d *hitpos
...
+		// everything above was calculated relative to the heavy's position
+		vec3d actual_world_hit_pos = mc.hit_point_world + heavy_obj->pos;
+		if ((shipp->is_arriving()) && (shipp->warpin_effect))
```
JohnAFernandez, line 1150: Checking against nullptr or zero or both? Even if it's done similarly around this part of the code, probably better to explicitly check, if for no other reason than for readability.  You've probably picked up by now that that's one of my big things.

Also the compiler will probably optimize it for us.

code/object/collideshipweapon.cpp
```diff
@@ -354,13 +354,37 @@ static int ship_weapon_check_collision(object *ship_objp, object *weapon_objp, f
 		hull_collision = model_collide(&mc_hull);
 	}
 
+	// check if the hit point is beyond the clip plane when warping out.
+	if (hull_collision || shield_collision) {
+		WarpEffect* warp_effect = nullptr;
+		bool hull_no_collide, shield_no_collide;
```
JohnAFernandez, line 360: These can be declared and initialized closer to where they are used.

code/ship/shipfx.cpp
```diff
@@ -1230,33 +1230,16 @@ void shipfx_emit_spark( int n, int sn )
 		vm_vec_add2(&outpnt,&obj->pos);
 	}
 
-	//WMC - I have two options here:
```
JohnAFernandez, line 1233: Removing these comments is the right move, IMO, but others might see it differently.

JohnAFernandez (changes requested):
Since no one more senior seems to have time, I just have a few nitpicks.  Where you declare and use bools, your code could be a little more concise since the functions are returning book values anyway.l, but I'm not going to say that has to be done.

code/ship/shipfx.cpp
Baezon: phreak implemented the very thing WMC was stressing over later, so it really doesn't make much sense to leave the comment here as it's referring to an older version of the warpeffect system.

code/asteroid/asteroid.cpp
Baezon: Well see quite often it *is* in fact the world hit point, so it's not an inaccurate name *most of the time*. It's definitely unideal, and I looked at refactored the above collision to have it actually be a world position but it quickly became clear it would need some not insignificant changes, testing, and just got out of scope.

JohnAFernandez (approved):
LGTM!  Looks like one of the checks is snagged, let me see where that went wrong

JohnAFernandez:
Looks like it's the trail thing again.  it might be fixed if you rebase against master.

PR #2847, opened 2020-10-30
Add '$Collision Radius Override' to weapons

Baezon:
Useful if a laser's bitmap doesn't visually map well to the radius the modder thinks it should have, or otherwise wants a weapon to have a particular radius. Ship-weapon collisions still only check against a ray so this won't affect them, which I may attempt to address at some point, but the specific requester only need it for weapon-weapon collisions so this should be fine for now.

### DIFF
--- a/code/weapon/weapon.h
+++ b/code/weapon/weapon.h
@@ -285,6 +285,7 @@ struct weapon_info
 	color	laser_color_1;						// for cycling between glow colors
 	color	laser_color_2;						// for cycling between glow colors
 	float	laser_head_radius, laser_tail_radius;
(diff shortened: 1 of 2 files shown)


JohnAFernandez (changes requested):
Ok.  Not a bad idea.

But please match the float format in the rest of FSO, if for no other reason, just to not confuse newbies who come in.  C++ will give them enough other things to be confused about already.

Baezon:
Oh, sorry. I didn't realize it was an issue.

JohnAFernandez (approved):

PR #2851, opened 2020-10-31
Don't warn for "no radius doubling" if the weapon isn't a bomb

### LINKED ISSUE #2850: "No Radius Doubling" on non-bomb interceptable missiles causes warning
As title. The warning also says it's ignoring the flag, which is undesirable if still true.

Baezon:
Because the weapon could still be interceptable and simply not a bomb, and fixed and moved the check into the proper place. And also renamed `Hard_target_bomb` to `No_radius_doubling` because it was a confusing name and... well why wasn't it called that from the get go they shouldn't have different names anyway

Closes #2850

### DIFF
--- a/code/object/collideweaponweapon.cpp
+++ b/code/object/collideweaponweapon.cpp
@@ -56,7 +56,7 @@ int collide_weapon_weapon( obj_pair * pair )
 	B_radius = B->radius;
 
 	if (wipA->weapon_hitpoints > 0) {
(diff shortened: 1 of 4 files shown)


JohnAFernandez (approved):
The sensible changes are everywhere!

EatThePath (approved):
Appologies if this is me overstepping, but the changes are pretty straight-forward and look good to me, for what it's worth.

PR #2853, opened 2020-10-31
Allow polish_position() to use any player's position in multi

JohnAFernandez:
9 year old bug. This will now switches from Player_obj to the shooting object if the shooting object is a player object.  The edited function now uses Player_obj as the default parameter to maintain the same balance for those mods that use this feature.


JohnAFernandez:
To be clearer, the bug is that autoaim currently only works at random times for clients in multi.  But it only doesn't work on the server.  This fix is tested, it now autoaims correctly on server and client.

Baezon (approved):

PR #2858, opened 2020-11-02
Add beam parameter validation

Baezon:
Adds a `beam_has_valid_params` function that beams use to validate their vital inputs. This was being handled pretty much ok already, but having an actual exhaustive list of the ways beams can be fired and what sorts of data they're actually supposed to have is A. less confusing and B. useful for multiplayer, which was the main impetus for this change. The changes to beam packet sending/receiving are here mostly just not to break anything in general, as @notimaginative will have a much better version of these soon. 

~~Also finishes removing `delta_ang` . This was only ever used in some dubious collision checking for beams that was fixed in #2735 , and `BEAM_T` which returns the 0-1 value of the beam's lifetime where it uselessly multiplied and divided by `delta_ang`.~~ holding off on this for now, per comments below

### DIFF
--- a/code/weapon/beam.h
+++ b/code/weapon/beam.h
@@ -99,7 +99,6 @@ typedef struct fighter_beam_fire_info {
 	int warmdown_stamp;
 	float life_left;	
 	float life_total;
-	int  fire_method;
 } fighter_beam_fire_info;
 
 // max # of collisions we'll allow per frame
--- a/code/weapon/beam.h
+++ b/code/weapon/beam.h
@@ -55,6 +55,14 @@ typedef struct beam_info {
 #define BFIF_TARGETING_COORDS	(1<<2)
 #define BFIF_FLOATING_BEAM		(1<<3)
 
+// to ensure validity of fire_info, the related fields MUST be provided
+#define BFM_TURRET_FIRED         0   // objp, subsys, target
+#define BFM_TURRET_FORCE_FIRED   1   // objp, subsys, target OR target_pos
+#define BFM_FIGHTER_FIRED        2   // objp, subsys
+#define BFM_SPAWNED              3   // starting pos, target OR target_pos
+#define BFM_SEXP_FLOATING_FIRED  4   // starting pos, target OR target_pos
+#define BFM_SUBSPACE_STRIKE      5   // starting pos, target
+
 // pass to beam fire 
 typedef struct beam_fire_info {
 	int				beam_info_index;				// weapon info index 
@@ -73,6 +81,7 @@ typedef struct beam_fire_info {
 	int point;									// for fighters, which point on the bank it is from
 	int bfi_flags;
 	char team;									// for floating beams, determines which team the beam is on
+	int  fire_method;
 } beam_fire_info;
 
 typedef struct fighter_beam_fire_info {
@@ -90,6 +99,7 @@ typedef struct fighter_beam_fire_info {
 	int warmdown_stamp;
 	float life_left;	
 	float life_total;
+	int  fire_method;
 } fighter_beam_fire_info;
 
 // max # of collisions we'll allow per frame
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1878,6 +1878,7 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 			else
 				fire_info.target_subsys = NULL;
 			fire_info.turret = turret;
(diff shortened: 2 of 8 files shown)


JohnAFernandez:
I don't see anything wrong with it, but @notimaginative is going to have to sign off on this since he is working on the multi packet fixes.

notimaginative:
Please avoid making changes to `send_beam_fired_packet()` and the `beam_info` struct. Those aren't really necessary for this at the moment and I can make those changes as part of the other PR, so we only have to break multi once.

You can modify `process_beam_fired_packet()` to handle `fire_method` so it will still work without breaking anything. Only `BFM_TURRET_FIRED`, `BFM_TURRET_FORCE_FIRED`, and `BFM_FIGHTER_FIRED` need to be accounted for since the other beam types aren't supported just yet. And those three should be easily determined based on the information that `process_beam_fired_packet()` has available. Then `fire_method` can be dealt with properly as part of the packet fixes PR as it would support all of the beam types.

code/weapon/beam.h
```diff
@@ -133,6 +141,7 @@ typedef struct beam {
 	int		framecount;				// how many frames the beam has been active
 	int		flags;					// see BF_* defines 	
 	float		current_width_factor;		// applied to width due to shrink or grow
+	ubyte     shot_count;              // # of shots
```
notimaginative, line 144: Could this be moved to `beam_fire_info` instead? As part of the packet fixes PR I just send that struct to `send_beam_fired_packet()`, so if it's part of that it would be easier to use in that case.

Additionally, just make this an `int`. Data size is much less important than alignment in this case.

notimaginative (changes requested):

code/weapon/beam.h
Baezon: Per the discussion on the multi packet fixes PR I was planning on simply reverting back to it being on `beam_info`. Did you want it somewhere else instead? It does have to persist with the beam in some way, or else look it up from `weapon_info` which seemed like it might be a bad idea.

Baezon:
I'm not sure I entirely follow. If I don't change `send_beam_fired_packet()` but do change `process_beam_fired_packet()` then current multi would certainly be broken, would this not be "breaking multi twice"? Is the goal here to leave `multimsgs` as untouched as possible? Preserve the integrity of of current multi (what little there is...)? I you want `beam_info` left alone, are you saying that you intend to remove `delta_ang` as part of the packet fixes? I'm happy to handle this however you want

code/weapon/beam.h
notimaginative: Having it back in `beam_info` is fine.

notimaginative:
Sorry for not being clear there. I just meant that in `process_beam_fired_packet()` you can manually set `fire_method` to `BFM_TURRET_FIRED`, or if it is a fighter beam set it that. The other beam types and flags aren't supported in the current code so it doesn't really matter about the other values for `fire_method`. So you're not changing the packet at all, just setting `fire_method` to a reasonably sane value.

That way the multi code is still fully compatible. A bit of a hack, but not more so that what's there now. Then it can be properly addressed in the packet fixes PR since it will support all of the beam types. And of course your changes here could go ahead and go in rather than waiting for the other PR to be ready.

Baezon:
Ok, I think I understand. What about `delta_ang`?

notimaginative:
Already have `delta_ang` removed in the packet fixes PR. The PR might not be fully up to date with my local changes though.

Baezon:
Ok, I reverted the changes to `shot_count`, put back `delta_ang`, and hardcoded `process_beam_fired_packet()` to only do turret and fighter beams.

code/weapon/beam.h
```diff
@@ -90,6 +99,7 @@ typedef struct fighter_beam_fire_info {
 	int warmdown_stamp;
 	float life_left;	
 	float life_total;
+	int  fire_method;
```
notimaginative, line 102: This is never used, correct? I don't see it set or checked against anywhere.

Actually, I'm not sure why this struct even still exists.

notimaginative (approved):

notimaginative:
Just the one little thing, but otherwise it looks good to me.

code/weapon/beam.h
Baezon: It's used for `beam_fire_targeting()` which is a weird very niche case for beams. It handles safety well enough on its own so yeah, this isn't really necessary. It could probably be refactored as a special case of regular `beam_fire`, but one thing at a time.

PR #2861, opened 2020-11-02
Fix a bug with collision_radius_override

Baezon:
It was never initialized, and so could be corrupting all weapons' radius.


EatThePath (approved):
Simple change, looks good, and testing shows it resolves the issue that prompted it's discovery without evident side effects

asarium (approved):

PR #2862, opened 2020-11-03
Don't perform automatic hash truncation before template-copying.

MageKing17:
In commit 3b2af702f764b7f7a7bd5fc35eef6586a6e02cb5, the hash truncation was changed so that instead of being part of the HUD, it's stored as the class's display name. However, this action was performed before `+Use Template:` was parsed, which would then overwrite the display name and clear the `Has_display_name` flag. I've moved it so that it now happens after the copy operation (and added a check for an existing display name, in case the template defines one). I use a new boolean variable to indicate that default display name creation needs to happen rather than duplicate that code.


JohnAFernandez (approved):
Looks good.  Simple fixes are always nice.

PR #2865, opened 2020-11-03
Some visual fixes to local ssms

### LINKED ISSUE #2792: Particles and Tails are still generated for weapons with "local ssm"-flag
I observed weapons with the "local ssm"-flag (fired from an AI bomber) to still generate a tail and particles (via $PSpew) on their original trajectory between the missile warping out and back in.

EDIT: Builds used were 20201099 and 20201011

Baezon:
No particle spew during stage 3, don't update the trail for corkscrew missiles in stage 3, kill the old trail, and create a new one when entering/exiting warp, and make sure to use the model's radius for the warp effect, in case the object''s radius was overwritten. 

Closes #2792

### DIFF
--- a/code/weapon/corkscrew.cpp
+++ b/code/weapon/corkscrew.cpp
@@ -246,7 +246,7 @@ void cscrew_process_post(object *objp)
 	vm_vec_sub(&ci->cen_p, &cen, &objp->pos);
 
 	// do trail stuff here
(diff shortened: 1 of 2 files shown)


JohnAFernandez (approved):
Ok.  Looks straight forward enough.  And it's all cosmetic, so we'll know somewhat quickly if something went wrong.

PR #2866, opened 2020-11-05
Add CYCLE_PRIMARY_WEAPON_SEQUENCE to Multi controls sent to server.

JohnAFernandez:
Bug was discovered in N95's.  Couldn't switch between single bolt and double bolt modes.  Turns out this control was mistakenly added to the non-essential controls list.


JohnAFernandez:
This does fix the issue and does not require a multi version bump.  However, requesting a review from Taylor since I'm not as familiar with this packet.

notimaginative:
I'm not totally sure, but I think that you'll need to also modify `multi_server_update_player_weapons()` to handle what the control changed. It should be called in `button_function_critical()`. Check `CYCLE_NEXT_PRIMARY` as an example.

That wouldn't be a breaking change either. It should just make sure that the server matches the client for weapon state.

JohnAFernandez:
I'll try it out with that changed as well.  It looked fine without it, but who knows what might be wrong behind the scenes.

JohnAFernandez:
It looks like the variable that Dynamic Primary linking affects is `primary_bank_slot_count` but that isn't included in `s_info`, which is what `multi_server_update_player_weapons` updates.  The primary banks did seems to work correctly without calling `multi_server_update_player_weapons`, so right now it doesn't seem like I need to add it to this PR.  But if you can think of something else that will need to be updated from this being updated, I'll put it into `s_info` and figure out how to look up that value from there.

notimaginative (approved):
I didn't look at `multi_server_update_player_weapons` that closely so if you don't think it's needed then that's good enough for me.

JohnAFernandez:
All right!  We can always take another look later if something seems off.

PR #2868, opened 2020-11-06
Controls0 - Vectorize Controls_config

z64555:
First PR of the overhaul.  This changes the Controls_config array into an SCP_vector to allow scripting to add more controls as they wish, enabling mod devs to create new controls with scripting and not needing hardcode support in the future.  (Custom controls are still a ways off, however)

Notice: Due to the nature of the controls code, Some features had to be disabled to get a compilation.  These will be address with the next PRs as soon as possible.

* Disabled the Undo system of the controls config menu.  You won't be able to undo bindings outside of doing a total reset via Escape.
* Disabled the Config presets.  You won't be able to cycle through or create new presets with this build.
* Restricted controlsconfigdefaults.tbl to use the first override set.  This is because the table used the config presets.  For now, the overrides are put directly into Control_config items as the defaults, thereby overriding the default hardcoded bindings for duration of the selected mod.

Vanilla behavior should not be affected in any way, this is primarily a refactor PR, but focus on the particular thigns:
[  ] Bind mode behaves as it should
[  ] Search mode behaves as it should
[  ] Loading controlconfigdefaults with retail data
[  ] Loading controlconfigdefaults with mod data
[  ] Undo shouldn't work (Intentionally disabled)
[  ] Presets shouldn't work (Intentionally disabled)

### DIFF
--- a/code/pilotfile/plr_convert.cpp
+++ b/code/pilotfile/plr_convert.cpp
@@ -126,7 +126,7 @@ plr_data::~plr_data()
 void pilotfile_convert::plr_import_controls()
 {
 	int idx;
-	config_item con;
+	CCI con;
 
 	unsigned char num_controls = cfread_ubyte(cfp);
 
@@ -134,8 +134,8 @@ void pilotfile_convert::plr_import_controls()
 		return;
 	}
 
-	// it may be less than 118, but it shouldn't be more than 118
-	if (num_controls > 118) {
+	// TODO: Currently only checks for hardcoded control bindings. Scripted controls will be sliced out in the pilot file
+	if (num_controls > CCFG_MAX) {
 		throw "Data check failure in controls!";
 	}
 
--- a/code/pilotfile/plr_convert.cpp
+++ b/code/pilotfile/plr_convert.cpp
@@ -134,8 +134,9 @@ void pilotfile_convert::plr_import_controls()
 		return;
 	}
 
-	// TODO: Currently only checks for hardcoded control bindings. Scripted controls will be sliced out in the pilot file
-	if (num_controls > CCFG_MAX) {
+	// it may be less than 118, but it shouldn't be more than 118
+	// Don't touch this magic number! This is for old playerfiles. See banner at top of this file.
+	if (num_controls > 118) {
 		throw "Data check failure in controls!";
 	}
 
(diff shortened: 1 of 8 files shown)


z64555:
Labeled as Unfinished as I haven't tested the behavior in retail yet. Will remove once it passes a runtime test.

z64555:
Testing the new PR Draft feature.  Seems to cancel most of the CI except for travis and appveyor.  Not sure if those two will still try building with further commits.

code/pilotfile/plr_convert.cpp
```diff
@@ -126,16 +126,16 @@ plr_data::~plr_data()
...
 	unsigned char num_controls = cfread_ubyte(cfp);
 
 	if ( !num_controls ) {
 		return;
 	}
 
-	// it may be less than 118, but it shouldn't be more than 118
-	if (num_controls > 118) {
+	// TODO: Currently only checks for hardcoded control bindings. Scripted controls will be sliced out in the pilot file
+	if (num_controls > CCFG_MAX) {
```
notimaginative, line 138: Can't do that. The magic number there is set to match the number of controls that were used in the old plr file format. Changing it will corrupt the data on conversion.

I'm pretty sure that I put a warning in there somewhere about not changing the magic numbers.

notimaginative (changes requested):

notimaginative:
This doesn't actually change CCFG_MAX, correct? That affects multi so I just wanted to be sure. We'll soon have it fixed where that doesn't break anything in the future though.

z64555:
Correct, it doesn't add any more controls to the IoActionId enum, so multi is not affected.

code/pilotfile/plr_convert.cpp
z64555: Reverted.  Also added a comment to say "can't touch this"

z64555:
Tested in retail, verified operation and disablement of the Undo.

Have not tested controlsconfigdefaults.tbl behavior. Have not tested with mod data. Tested the reset to defaults, which should do the preset cycling, but haven't tested the preset cycling disablement.

z64555:
Letsee, there is a pilotfile test for the CI which will probably need updating

z64555:
Closing this PR and merging it with #2907

PR #2870, opened 2020-11-07
Use armor/damage types for beam-weapon damage

Baezon:
Only useful with "Beams damage weapons", but an important step.

### DIFF
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -3333,6 +3333,12 @@ void beam_handle_collisions(beam *b)
 
 							float damage = real_damage * attenuation;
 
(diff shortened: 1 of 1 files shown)


JohnAFernandez (commented):
It looks simple enough to approve.  Just double checking that you tested it to make sure it follows expected behavior.  We're starting to get heavy on pull requests again.

Baezon:
A bit awkward to test, beams are treated as ray instead of a cylinder for weapon collision, but seems to be working properly

JohnAFernandez:
We should merge this right after stable.

Goober5000 (approved):

PR #2871, opened 2020-11-07
AI Edge Case Fix

JohnAFernandez:
This is for an edge case crash.  A small minority of AI values was causing crashes on clients because the target being nullptr was not being handled correctly with my upgraded object update code.  Because this only changes the logic that handles incoming information, there is no need for a multi bump.

Still needs testing, so leaving it in draft status for now.


JohnAFernandez:
Reports of this crash at least, have not popped up.  Now to handle the 8 other bugs.

wookieejedi (approved):
Good catch.

JohnAFernandez:
It was all Talor.

PR #2873, opened 2020-11-08
Add healing weapons

Baezon:
Weapons that heal instead of damage their target :)

The main functions, `ship_do_healing` and `do_subobj_heal_stuff` are fundamentally the same to their counterparts `ship_do_damage` and `do_subobj_do_hit_stuff`, but I found that the differences were just too significant and numerous to handle as special cases of those functions. At least this way, it is less likely to break any existing behavior.

Healing works generally the same way as damage, using factors and armor/damage types, but there are several quirks as well, such as the healing *always* carrying, as opposed to damage, which could be intercepted by the shield, and then by subsystems, before damaging the hull. Healing applies in full to the shield, subsystems and hull, this way a damaged fighter with its shields up won't block your attempts to heal its hull.


code/ship/shiphit.cpp
```diff
@@ -2393,6 +2608,125 @@ static void ship_do_damage(object *ship_objp, object *other_obj, vec3d *hitpos,
...
+		if (shipp->armor_type_idx != -1)
+			healing = Armor_types[shipp->armor_type_idx].GetDamage(healing, dmg_type_idx, 1.0f, other_obj_is_beam);
+		
+		if (wip->wi_flags[Weapon::Info_Flags::Puncture])
+			healing /= 4;
+
+		healing *= wip->armor_factor;
+
+		// multiplayer clients don't do healing
+		if (!MULTIPLAYER_CLIENT) {
```
JohnAFernandez, line 2696: Multiplayer clients shouldn't do *anything* healing related: shields, hull, or subsystem. So clients should probably abort much sooner than this.

JohnAFernandez (changes requested):
Mostly small stuff that I see.  But like I said.  We could probably use a second iteration at some point that allows you to heal hull, shield or subsystems in a specific order.  I think there's a way to do it without too many changes.

code/ship/shiphit.cpp
Baezon: This is mostly copied from the structure of the damaging flow, should the damage *also* abort much sooner?

JohnAFernandez: Servers handle all actual damage calculation. Maybe in the damage codepath they already do abort sooner, and that was just someone thinking they are doing something when they are not.

JohnAFernandez (approved):

PR #2876, opened 2020-11-09
Fix 'getFrametime' scripting function

Baezon:
The documention stating of its optional parameter `[boolean adjustForTimeCompression]` is inverse of what the function actually does, setting this to true will give you the real frametime, instead. So per @asarium's suggestion, I've deprecated the old function in favor of two separate functions for each kind of frametime. I'm still not super happy with this approach, but I'm not sure how else to handle it.

IF YOU HAVE ANY SCRIPTS USING THIS YOU MAY WANT TO DOUBLE CHECK ITS IMPLEMENTATION.
getFrametime(true) -> Real frametime.
getFrametime(false or omitted) -> time compressed frametime.

EDIT: This is the crux of #2364

### DIFF
--- a/code/scripting/api/libs/base.cpp
+++ b/code/scripting/api/libs/base.cpp
@@ -229,17 +229,22 @@ ADE_FUNC(getFrametimeOverall, l_Base, NULL, "The overall frame time in seconds s
 	return ade_set_args(L, "x", game_get_overall_frametime());
 }
 
-ADE_FUNC(getTimeCompressedFrametime, l_Base, NULL, "Gets how long this frame is calculated to take. Use it to for animations, physics, etc to make incremental changes. Increased or decreased based on current time compression", "number", "Frame time (seconds)")
+ADE_FUNC(getTimeCompressedFrametime, l_Base, nullptr, "Gets how long this frame is calculated to take. Use it to for animations, physics, etc to make incremental changes. Increased or decreased based on current time compression", "number", "Frame time (seconds)")
 {
 	return ade_set_args(L, "f", flFrametime);
 }
 
-ADE_FUNC(getRealFrametime, l_Base, NULL, "Gets how long this frame is calculated to take in real time. Not affected by time compression.", "number", "Frame time (seconds)")
+ADE_FUNC(getRealFrametime, l_Base, nullptr, "Gets how long this frame is calculated to take in real time. Not affected by time compression.", "number", "Frame time (seconds)")
 {
 	return ade_set_args(L, "f", flRealframetime);
 }
 
-ADE_FUNC(getFrametime, l_Base, "[boolean adjustForTimeCompression]", "Gets how long this frame is calculated to take. Use it to for animations, physics, etc to make incremental changes. DEPRECATED, due a mistake in the documentaion.", "number", "Frame time (seconds)")
+ADE_FUNC_DEPRECATED(getFrametime, l_Base, 
+	"[boolean adjustForTimeCompression]", 
+	"Gets how long this frame is calculated to take. Use it to for animations, physics, etc to make incremental changes.", 
+	"number", "Frame time (seconds)", 
+	gameversion::version(20, 2, 0, 0),
+	"Please use getRealFrametime() or getTimeCompressedFrametime() instead.")
 {
 	bool b=false;
 	ade_get_args(L, "|b", &b);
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/base.cpp
```diff
@@ -229,7 +229,17 @@ ADE_FUNC(getFrametimeOverall, l_Base, NULL, "The overall frame time in seconds s
...
+{
+	return ade_set_args(L, "f", flFrametime);
+}
+
+ADE_FUNC(getRealFrametime, l_Base, NULL, "Gets how long this frame is calculated to take in real time. Not affected by time compression.", "number", "Frame time (seconds)")
+{
+	return ade_set_args(L, "f", flRealframetime);
+}
+
+ADE_FUNC(getFrametime, l_Base, "[boolean adjustForTimeCompression]", "Gets how long this frame is calculated to take. Use it to for animations, physics, etc to make incremental changes. DEPRECATED, due a mistake in the documentaion.", "number", "Frame time (seconds)")
```
JohnAFernandez, line 242: You should add a warning that only shows once per game session if this one is used to inform users and modders that this script needs to be updated.

JohnAFernandez (dismissed):

code/scripting/api/libs/base.cpp
Baezon: Note that, it is entirely possible to be using this function correctly (the fault is really just with the documentation), and that until the MVPs is updated this will be warning everyone who uses or depends on it. Given this I really fear that a warning is over-aggressive.

JohnAFernandez: Then we need to poke the MVPs people aggressively to fix this ASAP instead.  Let's see what other people have to say about this PR.

Goober5000:
If the problem is only with the documentation, then why not change the documentation (and maybe the name of the function parameter) and leave the function as-is?

code/scripting/api/libs/base.cpp
asarium: This should use `ADE_FUNC_DEPRECATED` which will handle the deprecation behavior automatically. The deprecation version should be the next release so `20.2`.

That will mean that there will be a message in the API doc specifying that the function is deprecated and the function will also not be available when the mod reports a supported engine version equal or higher than the deprecation version.

asarium (changes requested):

asarium:
@Goober5000 That would also be an option if it is made clear that a change was made in case someone comes across a script using the old behavior.

code/scripting/api/libs/base.cpp
Baezon: Ah, I was wondering how to handle deprecation.

code/scripting/api/libs/base.cpp
```diff
@@ -229,7 +229,22 @@ ADE_FUNC(getFrametimeOverall, l_Base, NULL, "The overall frame time in seconds s
 	return ade_set_args(L, "x", game_get_overall_frametime());
 }
 
-ADE_FUNC(getFrametime, l_Base, "[boolean adjustForTimeCompression]", "Gets how long this frame is calculated to take. Use it to for animations, physics, etc to make incremental changes.", "number", "Frame time (seconds)")
+ADE_FUNC(getTimeCompressedFrametime, l_Base, nullptr, "Gets how long this frame is calculated to take. Use it to for animations, physics, etc to make incremental changes. Increased or decreased based on current time compression", "number", "Frame time (seconds)")
```
asarium, line 232: A better name would be `getMissionFrametime` since this returns how much in-mission time has elapsed.

code/scripting/api/libs/base.cpp
```diff
@@ -229,7 +229,22 @@ ADE_FUNC(getFrametimeOverall, l_Base, NULL, "The overall frame time in seconds s
...
+{
+	return ade_set_args(L, "f", flRealframetime);
+}
+
+ADE_FUNC_DEPRECATED(getFrametime, l_Base, 
+	"[boolean adjustForTimeCompression]", 
+	"Gets how long this frame is calculated to take. Use it to for animations, physics, etc to make incremental changes.", 
+	"number", "Frame time (seconds)", 
+	gameversion::version(20, 2, 0, 0),
+	"Please use getRealFrametime() or getTimeCompressedFrametime() instead.")
```
asarium, line 247: Would be good to explain _why_ this function is deprecated. For example "The parameter of this function is flipped from what the naming suggests (passing true returns non-adjusted time). Please use either getMissionFrametime() or getRealFrametime()."

asarium (changes requested):

asarium (approved):

PR #2878, opened 2020-11-09
Fix pilot file data loss

notimaginative:
This addresses two pilot file issues:

1. Total data loss if a pilot has any medals in multi stats
2. Partial data loss in active pilot when entering the barracks


JohnAFernandez (approved):
Ok.  It helped to break that down by commit.  Looks good!

PR #2881, opened 2020-11-11
Two fixes to weapon states

Baezon:
When the weapon state is changed for the purposes of creating particle sources, while switching to the new one it checks if the old one was `WeaponState::INVALID` and if so attempts to create a `WeaponState::NORMAL` particle source. Given that the normal, not-invalid case creates a new source for the new state directly afterwards anyway, this is redundant and causes NORMAL weapon states (that is, particles spawned by primaries) to have double the particle sources.

Additionally an invalid `wp->homing_object` is `&obj_used_list` and not `nullptr`, this was causing all dumbfired missiles to not create any particle sources at all.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -4876,25 +4876,7 @@ static void weapon_set_state(weapon_info* wip, weapon* wp, WeaponState state)
 		return;
 	}
 
-	auto current_state = wp->weapon_state;
-
 	wp->weapon_state = state;
-
-	if (current_state == WeaponState::INVALID)
-	{
-		// First weapon state, create the in-flight effect
-		auto map_entry = wip->state_effects.find(WeaponState::NORMAL);
-
-		if (map_entry != wip->state_effects.end())
-		{
-			auto source = particle::ParticleManager::get()->createSource(map_entry->second);
-
-			source.moveToObject(&Objects[wp->objnum], &vmd_zero_vector);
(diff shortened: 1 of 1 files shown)


code/weapon/weapons.cpp
```diff
@@ -4929,7 +4911,7 @@ static void weapon_update_state(weapon* wp)
 			infree_flight = true;
 		}
 		else if (lifetime >= fl2f(wip->free_flight_time) &&
-			(lifetime - Frametime) <= fl2f(wip->free_flight_time) && wp->homing_object != nullptr)
+			(lifetime - Frametime) <= fl2f(wip->free_flight_time) && wp->homing_object != &obj_used_list)
```
JohnAFernandez, line 4914: We'll need a comment about the &obj_used_list to make sure it's clear to others in the future.  I myself would feel very confused if I saw that.

JohnAFernandez (changes requested):

code/weapon/weapons.cpp
Baezon: It's still weird to me, but its something about linked lists. It is used all over the code though, so it seems strange to single out these spots, should I be commenting all of them?

JohnAFernandez: *looks around* Just here, I think.

JohnAFernandez (approved):

JohnAFernandez:
Sorry I left this languishing for a little bit.

PR #2883, opened 2020-11-11
Add trail 'spread'

Baezon:
Causes trail points to drift away from each other making a very nice visual effect :)

### DIFF
--- a/code/weapon/trails.cpp
+++ b/code/weapon/trails.cpp
@@ -304,6 +304,8 @@ void trail_render( trail * trailp )
 	g3_render_primitives_colored_textured(&material_def, Trail_v_list, nv, PRIM_TYPE_TRISTRIP, false);
 }
 
+// Adds a new segment to trailp at pos, in order to do spread an orient is provided 
+// so it can drift away orthogonally from the direction of the trail
 void trail_add_segment( trail *trailp, vec3d *pos , const matrix* orient)
 {
 	int next = trailp->tail;
@@ -324,7 +326,7 @@ void trail_add_segment( trail *trailp, vec3d *pos , const matrix* orient)
 	if (orient != nullptr && trailp->info.spread > 0.0f) {
 		vm_vec_random_in_circle(&trailp->vel[next], &vmd_zero_vector, orient, trailp->info.spread, false);
 	} else 
-		vm_vec_make(&trailp->vel[next], 0, 0, 0);
+		vm_vec_zero(&trailp->vel[next]);
 }		
 
 void trail_set_segment( trail *trailp, vec3d *pos )
--- a/code/weapon/trails.cpp
+++ b/code/weapon/trails.cpp
@@ -368,7 +368,7 @@ void trail_move_all(float frametime)
 					num_alive_segments++;	// Record how many still alive.
 				}
 
-				trailp->pos[n] += trailp->vel[n] * frametime;
+				trailp->pos[n] += trailp->vel[n] * frametime; 
 
 			} while ( n != trailp->head );
 		}		
--- a/code/weapon/trails.cpp
+++ b/code/weapon/trails.cpp
@@ -304,7 +304,7 @@ void trail_render( trail * trailp )
 	g3_render_primitives_colored_textured(&material_def, Trail_v_list, nv, PRIM_TYPE_TRISTRIP, false);
 }
 
-void trail_add_segment( trail *trailp, vec3d *pos )
+void trail_add_segment( trail *trailp, vec3d *pos , const matrix* orient)
 {
 	int next = trailp->tail;
 	trailp->tail++;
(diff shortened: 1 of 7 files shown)


code/weapon/trails.h
```diff
@@ -63,7 +66,7 @@ void trail_render_all();
 
 // Returns -1 if failed
 trail *trail_create(trail_info *info);
-void trail_add_segment( trail *trailp, vec3d *pos );
+void trail_add_segment( trail *trailp, vec3d *pos , const matrix* orient = nullptr );
```
JohnAFernandez, line 69: The only thing I see is that this function now has a default parameter, so I'm assuming there were other places this is called that are not part of this pr?

JohnAFernandez (commented):
It's wonderful when

code/weapon/trails.h
Baezon: Every instance of `trail_add_segment()` can, was updated to use, and tested with trail spread, so currently all of them send in orient, so it _could_ be made mandatory, but I thought this was a nicer approach in general/for the future since it isn't actually necessary, it would just be a normal trail.

JohnAFernandez: I think it's better to have it mandatory now, and then coders in the future could make it optional, when they need it to be. Optional values are nice, but your indirectly communicating to future coders that there already *is* an instance of this function that uses the default value somewhere and they're just not good at looking for it.

The videos do look really nice, btw.

JohnAFernandez (approved):

Goober5000 (changes requested):
Elaboration needed

code/weapon/trails.h
Goober5000: @JohnAFernandez I read your comment several times and I'm still not sure what you mean.  If the parameter *can* be optional, and is *designed* to be optional, then why not make it optional?  Punting the "optional" implementation to the future is just going to make it get lost.

The optional form of the function is perfectly reasonable.  Who uses a function without looking at the function prototype?

JohnAFernandez: But the optional argument is never used.

Baezon: I'm ambivalent either way. It is not required in order to function like a basic trail. Every caller is capable of using it as a 'spreading' as opposed to basic trail, and so all callers send in the orient (although it may not be used). I'm happy to go with whatever you guys think is a better approach.

JohnAFernandez: Let me put it this way:

A rookie coder is looking at this function and is trying to figure out where it's used to get a better handle on what it's used for.  Maybe they have MSVC and intellisense but they're not sure how well they can trust it, since they are brand new.  They search for the function using multiple methods and yet cannot find anywhere using the default argument. 

"Why would they have a default argument, if it is never used?  I must be doing something wrong."

It's just clearer code to not have the default argument, and it makes it more accessible to newer coders, who we shouldn't be discouraging.

Goober5000: I think a coder is just as likely to blame the *previous* coder as he is to blame himself.  Even so, this isn't an API problem but a documentation problem.

If there's a good reason to make the argument optional, then make it optional.  If there's a good reason to always require the argument, then always require the argument.

But I think the root problem here can be solved with some documentation.  @Baezon, add some comments explaining what the new argument does, what values are allowed, and why a coder may want to choose one over the other.

JohnAFernandez: To be clear, I meant that the argument just needs to not be default.  It is a good idea to add more comments.

Baezon: I do think, for any possible future usage of this I can think of, some forward vector to spread away from orthogonally should be available, so mandatory should be OK. With some documentation explaining this is the purpose of the orient, of course.

Goober5000: That sounds good to me. And the documentation is a good start, but please expand a bit on what "spread" is (a link or reference is fine), describe what the orient indicates, and describe how the trail changes with an orient vs. a nullptr.

Baezon: I mean, the function leads them directly to `trail->info.spread` if they want the full explanation, and I'm not sure how much more there is to say about how the orient is used.

code/weapon/trails.cpp
```diff
@@ -304,7 +304,9 @@ void trail_render( trail * trailp )
 	g3_render_primitives_colored_textured(&material_def, Trail_v_list, nv, PRIM_TYPE_TRISTRIP, false);
 }
 
-void trail_add_segment( trail *trailp, vec3d *pos )
+// Adds a new segment to trailp at pos, in order to do spread an orient is provided 
+// so it can drift away orthogonally from the direction of the trail
+void trail_add_segment( trail *trailp, vec3d *pos , const matrix* orient)
```
Goober5000, line 309:
```suggestion
// Adds a new segment to trailp at pos.
// In order for trailp's `spread` field to have any effect, it must be nonzero and the orient must be non-null.
// The orient is relative to [what?]
// [It can drift away orthogonally from the direction of the trail] <-- what is "it"?
void trail_add_segment( trail *trailp, vec3d *pos , const matrix* orient)
```

code/weapon/trails.h
Goober5000: The comment is confusing.  I tried to make a suggestion but it really wasn't clear what you were trying to say.  The terms need to be made explicit and the explanation needs to provide more detail.

code/weapon/trails.cpp
Baezon: Ok, I think I see what you mean.

Goober5000 (approved):
Much better :+1:

JohnAFernandez:
Re-running the checks, let's see what happens.

PR #2884, opened 2020-11-11
Controls1 - Refactor Presets

z64555:
Parent PR: #2868 

This re-enables control presets and adds a bunch of features.

* Hardcoded defaults are now a preset.
* Mod devs may override the hardcoded defaults with the controlconfigdefaults.tbl by naming a #ControlConfigOverride section as "default"
* Mod devs may add as many presets as they like with #ControlConfigPreset sections.  These have the same syntax and structure as #ControlConfigOverride but only do bindings.
* Added ability to change the name of controls as they show up in the menu
* Added possibility of resumption, should a $Bind Name: not be found within the controls.

[  ] Bind mode behaves as it should
[  ] Search mode behaves as it should
[  ] Browse mode behaves as it should
[  ] Reset to defaults should behave as it should
[  ] Repeat hitting of "Reset" button should cycle through all presets
[  ] controlconfigdefaults.tbl is able to override the default preset
[  ] controlconfigdefaults.tbl is able to make new presets with the #ControlConfigOverride section
[  ] controlconfigdefaults.tbl is able to make new presets with the #ControlConfigPreset section
[  ] controlconfigdefaults.tbl is able to relabel controls
[  ] relabled controls should retain their bindings from PLR and CSG data.
[  ] Invalid entries in controlconfigdefaults.tbl should have some degree of resumption

### DIFF
--- a/code/controlconfig/controlsconfig.cpp
+++ b/code/controlconfig/controlsconfig.cpp
@@ -824,22 +824,17 @@ int control_config_do_reset()
 	return 0;
 }
 
-/**
- * @brief Resets all controls to values within the currently selected preset
- */
 void control_config_reset_defaults()
 {
-	int i;
-
 	// Reset keyboard defaults
 	const CC_preset &preset = Control_config_presets[Defaults_cycle_pos];
-	for (auto i = 0; i < Control_config.size(); ++i) {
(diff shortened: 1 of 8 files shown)


z64555:
Closing this PR and merging it with #2907

PR #2885, opened 2020-11-12
Expose ability to be targeted by 'huge' and 'small only' weapons

Baezon:
These flags before were treated as whether the target belonged to one of the hardcoded classes, and now can be controlled with "+Targeted by 'Huge' weapons:" and "+Ignored by 'small only' weapons:" objecttypes.tbl fields. I might've preferred "Targeted by..." for both of them, but this way they are both default FALSE, to better match how the old logic would treat them. You must opt into being 'targeted by huge' and 'ignored by small' (Huge meant ONLY [a hardcoded big class], and small only meant NOT [a hardcoded big class]). I've commented a few places to explain this difference in logic, but basically Huge includes only its valid targets, while small only excludes only invalid targets.

I've removed the warning about combining 'small only' and 'huge' flags on a weapon, since now you can technically have overlap if you want, but I'm not super sure about this, because in _most_ cases it is still probably not what people want.


Goober5000:
> I've removed the warning about combining 'small only' and 'huge' flags on a weapon, since now you can technically have overlap if you want, but I'm not super sure about this, because in most cases it is still probably not what people want.

The warning should be kept, or it should be rewritten to explain exactly what the behavior would be.

Baezon:
Ok, I've merged the two flags, into the somewhat cumbersomely long "Targeted by huge/Ignored by small only". This way these two always go together, the far more straightforward scenario, so the old warning can be kept without confusion or edge cases. Also added this as part of the ai fighter weapon selection decision procedure that i missed the first time

wookieejedi (approved):
Looks good.

Issue #2882, opened 2020-11-11
Rotating Animation no longer working properly.


UndyingNephalim:
Triggered rotation animations no longer work properly. They begin their animation and almost immediately revert to their default position.

z64555:
Whats the earliest nightly that the issue arises?

JohnAFernandez:
It's confirmed, the PR fixed the issue.

PR #2888, opened 2020-11-14
Fix a bug with model animation

Baezon:
Even after some investigation, it was pretty difficult to for me to try and pry apart the model animation code and properly diagnose this. I stumbled on this merely by walking back changes in #2195 , which was confirmed to be the cause of #2882 . I've checked out other animations, mostly those from BP ships, and all of them continue to function normally after this change, and if anyone can think of any other cases, I'd be happy to test those as well.

The relevant code seems to be returning the total time of an animation, and in the problem case `psub->triggers[i].end + psub->triggers[i].start` was 9000, 9 seconds total, rather than 0. The other cases, such as the BP ships also go through this code without problem with or without this change so I'm not really sure what the root cause of the issue was, but given it critically broke a soon-to-be-released campaign, I figured making it go away promptly while not breaking anything else was the best way to go.

### DIFF
--- a/code/model/modelanim.cpp
+++ b/code/model/modelanim.cpp
@@ -810,7 +810,7 @@ int model_anim_get_time_type(ship_subsys *pss, AnimationTriggerType animation_ty
 			} else {
 				// if it isn't moving then it's trivial.
 				// no currently playing animation
-				ani_time = 0;
+				ani_time = psub->triggers[i].end + psub->triggers[i].start;
 			}
 
 			if (ret < ani_time)


JohnAFernandez:
It may be that the best thing we can do to make sure this is the fix is to check whether it breaks animations that were already working, like those in BP.  I'm not saying that *you* have to do it.  You've been doing a lot lately.  Maybe we can wait until things calm down a little to give it a test.

Baezon:
"I've checked out other animations, mostly those from BP ships, and all of them continue to function normally after this change"

Goober5000 (approved):
Ah, yes, this change makes sense.

Good strategy for this PR.  I'm pretty sure this is the proper fix, but regardless, you analyzed the problem in exactly the right way.

PR #2890, opened 2020-11-15
Close out mission when returning to multi screens

notimaginative:
This should fix any number of weird multi issues when exiting a mission without actually having gone into full game play state.


JohnAFernandez (approved):
Glad you found it.  This is a very straightforward fix, too.  Honestly, I knew it had to be in the state code, but it was driving me crazy not being able to figure it out.

PR #2892, opened 2020-11-15
Allow Fred to Manually Adjust Squadron Wings When Some Are Not Present

JohnAFernandez:
In order to avoid a multi bug that would be a lot of work to fix properly (reworking the entire way that multi handles loadouts), this adjusts the Fred saving function to, at first, ignore unused wings in the Squadron Wings array.  It then sets those missing names at the end of the list so that the slots can be set correctly during mission initialization.

### DIFF
--- a/fred2/missionsave.cpp
+++ b/fred2/missionsave.cpp
@@ -2438,24 +2438,35 @@ int CFred_mission_save::save_mission_info()
 			fout(")");
 		}
 
-		// We are now always writing squadron wings to avoid a multi bug where, if the mission designer skips one of the squadron wings
+		// We are now always writing multi squadron wings to avoid a multi bug where, if the mission designer skips one of the squadron wings
 		// it would disappear from the mission
-		fout("\n$Squadron wing names: ( ");
+		if (The_mission.game_type & MISSION_TYPE_MULTI) {
+			fout("\n$Squadron wing names: ( ");
 
-		SCP_vector<int> undefined_wings;
-		for (i = 0; i < MAX_SQUADRON_WINGS; i++) {
-			if (is_wing_defined(Squadron_wing_names[i])) {
-				fout("\"%s\" ", Squadron_wing_names[i]);
-			} else {
-				undefined_wings.push_back(i);
+			SCP_vector<int> undefined_wings;
+			for (i = 0; i < MAX_SQUADRON_WINGS; i++) {
+				if (is_wing_defined(Squadron_wing_names[i])) {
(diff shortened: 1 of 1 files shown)


fred2/missionsave.cpp
```diff
@@ -4348,3 +4370,15 @@ int CFred_mission_save::save_wings()
 	Assert(count == Num_wings);
 	return err;
 }
+
+// quick helper function to check if a wing is currently defined.
+bool is_wing_defined(char* wing_name)
```
Goober5000, line 4375: You may want to make this `const char *`.

Additionally, if you move the function above `CFred_mission_save::autosave_mission_file`, you won't need the forward declaration.

fred2/missionsave.cpp
```diff
@@ -2435,15 +2438,34 @@ int CFred_mission_save::save_mission_info()
 			fout(")");
 		}
 
-		// squadron wings
-		if (strcmp(Squadron_wing_names[0], "Alpha") || strcmp(Squadron_wing_names[1], "Beta") || strcmp(Squadron_wing_names[2], "Gamma") || strcmp(Squadron_wing_names[3], "Delta") || strcmp(Squadron_wing_names[4], "Epsilon")) {
+		// We are now always writing multi squadron wings to avoid a multi bug where, if the mission designer skips one of the squadron wings
+		// the wing after would disappear from the mission
+		if (The_mission.game_type & MISSION_TYPE_MULTI) {
```
Goober5000, line 2443: Add a condition here to check that the mission save format is not retail.  If someone ever wanted to make a retail-compatible multi mission, the "output only if they are not the default" behavior should be used, otherwise the retail engine would choke on the unexpected data.

Goober5000 (changes requested):
Looks all right, except for one needed change and one(-ish) suggested change.

Please also make the corresponding changes to `qtfred\src\mission\missionsave.cpp`.

Goober5000:
I just had an idea.  What if, instead of modifying the *save* code, the *load* code was modified?  In other words, load the wings as specified in the mission file, and then load any undefined wings.  This change would be implemented at the mission file parse level, so it wouldn't need to mess with the design of multi loadouts.

JohnAFernandez:
So, I actually tried that for a couple hours yesterday, but after taking a fresh look just now, I think I see a way to do it without to much issue.

JohnAFernandez:
So, actually going to just add a warning to mod makers when saving a mission that there is a mistake that is better manually fixed instead of we programmers guessing what order you want the squadron wings to be in.

PR #2895, opened 2020-11-15
Replace multicast code with mDNS/DNS-SD

notimaginative:
This replaces the original multicast code with mDNS/DNS-SD and should allow for LAN discovery with minimal to zero effort on the part of users.

### DIFF
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -52,6 +52,8 @@ endif()
 
 target_link_libraries(code PUBLIC parsers)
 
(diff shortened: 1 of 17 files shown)


JohnAFernandez (approved):
While I would love to have more sanity testing on another setup, the testing you've done may be all that we can get before the stable.  And this is definitely more stable than the regression of leaving it broken. 

I did look over everything, and like the other parts of the back end code, it only vaguely makes sense to me.  Good thing most of it is a library we're using and not custom code.

Issue #2889, opened 2020-11-14
shield_integrity[i] is unused


JohnAFernandez:
The ship::shield_integrity[] field is initialized but never used.  

Shield integrity is tracked by the quadrants instead.  It is also malloc'ed, so it should really be removed to prevent any chance of a memory leak in the future.

JohnAFernandez:
Well that was easy

PR #2898, opened 2020-11-17
Remove the unused shield_integrity field

JohnAFernandez:
This variable is never used and could cause a memory leak if anything about it was accidentally altered.  Also, ship creation will end up creating an entry for every tri in the shield mesh,meaning that we're being pretty inefficient by keeping it in.


Baezon (approved):
Assuming this is in fact all references to it, LGTM

JohnAFernandez:
> 
> 
> Assuming this is in fact all references to it, LGTM

I checked it by building after making the removals.

JohnAFernandez:
#2889

PR #2899, opened 2020-11-17
Prevent loadout issues in multi due to iff changes

notimaginative:
Fixes a bug where the loadout will be different when respawning or with in-game joining if the iff of the ship/wing has changed since the mission started.


JohnAFernandez (approved):
Yeah, I really do think this is the best fix for the job.  For anyone on the outside looking in, it turns out some of the items in the parse_object struct are changed after the object is created. When I first refactored this function, I had assumed that the team field would never change.  So much for that.

PR #2900, opened 2020-11-18
Play directive completion sounds for clients

JohnAFernandez:
This allows multiplayer clients to hear the directive completion sound at the right time and to also see when a directive has been failed.

### DIFF
--- a/code/mission/missiongoals.cpp
+++ b/code/mission/missiongoals.cpp
@@ -888,6 +888,11 @@ void mission_event_unset_directive_special(int event)
 	Mission_directive_special_timestamp = timestamp(-1);
 }
 
+void mission_event_set_completion_sound_timestamp()
+{
+	Mission_directive_sound_timestamp = timestamp(DIRECTIVE_SOUND_DELAY);
+}
+
 // function which evaluates and processes the given event
 void mission_process_event( int event )
 {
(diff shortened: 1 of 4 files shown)


Goober5000:
It looks as it the code you added is duplicating some stuff that is already run on the client side - setting `satisfied_time` for example.

Since you already call `send_event_update_packet(event);`, can the client function that receives that packet play the directive sound?

JohnAFernandez:
> It looks as if the code you added is duplicating some stuff that is already run on the client side -- setting `satisfied_time` for example.
> 
> Since you already call `send_event_update_packet(event);`, can the client function that receives that packet play the directive sound?

The original bug was that they weren't playing it.  So I found where it was playing on the server, and that part of the function was not playing on the client.  But let new look again at `satisfied_time` to see if there's another place where it just wasn't following the correct code path.

JohnAFernandez:
Ok. I see what you're saying now.  Yes, that would be a lot more optimal.  Let's see if I can write something cleaner.

Goober5000 (approved):
That looks like it makes much more sense.

Goober5000:
I think @notimaginative should take a look at it just as a general multiplayer sanity check.

notimaginative (approved):
LGTM

PR #2901, opened 2020-11-18
Make Anti-Rubberbanding less aggressive to keep the client from showing nonsense

JohnAFernandez:
Basically, if the anti rubber-banding code is too aggressive, ships (especially ships without afterburners) will be kept from moving forward at all.  This reduces the anti rubber-banding to half its original strength.

As discussed with @Baezon

### DIFF
--- a/code/network/multi_obj.cpp
+++ b/code/network/multi_obj.cpp
@@ -3292,15 +3292,17 @@ void multi_oo_interp(object* objp)
 
 	local_rubberband_correction = vmd_zero_vector;
 
+	constexpr float anti_rubberbanding_factor = 0.5f;
+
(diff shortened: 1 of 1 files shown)


Baezon (approved):
I might wonder in the future if there is a better way to either expose or dynamically determine this magic number, but in the interest of decent results quickly, this should probably be fine.

JohnAFernandez:
And just to clarify, @Shivansps who has been our main tester, seems very satisfied with the changes.

The next step (if one is taken) is to make AI rotation look less stiff somehow.  Since they don't have rotational velocity, their current rotation looks like a slide show, and there *might* be some way for us to somehow interpolate for AI rotation better.

Goober5000 (approved):

PR #2903, opened 2020-11-18
Slightly alter when a timestamp warning is written to the log.

JohnAFernandez:
This was always writing a warning about negative timestamps to the log.  But negative timestamp differentials are supposed to happen on the first two frames because there is no actual information to compare to.


Goober5000:
If negative timestamp differentials are supposed to happen on the first two frames, then shouldn't the warning only be displayed for >= 2?

JohnAFernandez:
> 
> 
> If negative timestamp differentials are supposed to happen on the first two frames, then shouldn't the warning only be displayed for >= 2?

Yes, whoops

MageKing17 (approved):

PR #2904, opened 2020-11-18
Fix a bug with shockwave damage parsing

Baezon:
`$Damage` only updates the shockwave's damage the first time through, future .tbms won't affect it. With this, changes to `$Damage` also affect the shockwave damage unless it was specifically set.


Goober5000 (approved):

Goober5000:
Looks like this can be merged even though the Windows check failed.

PR #2905, opened 2020-11-18
adjust weapon calculations based on Discord analysis

Goober5000:
General Battuta's summary:

> BLAST BUT NO SHOCK (SHOCK SPEED 0): Double damage to hull, regular damage to subsystems
> BLAST AND SHOCK: Double damage to hull, double damage to subsystems


JohnAFernandez (approved):
I didn't review this earlier, because I thought you were changing the behavior, but now I realized you're just changing the weapon info spew code.

PR #2906, opened 2020-11-18
fix some sort of build problem in the continuous integration

Goober5000:


JohnAFernandez (approved):

notimaginative:
`pow()` takes two doubles as arguments and returns a double, except for the weird overload thing that Windows does. It's a platform issue because the code is wrong. This just isn't the right way to fix the problem.

You need to either cast the arguments to double/float, or use `powf()` instead, which I don't believe has those stupid Microsoft workarounds.

Goober5000:
I'll keep `powf` in mind for the future.  I don't think these fixes are strictly *wrong* though.  According to my searches, `pow` is, even in 2020, slower than multiplication for integer powers.

PR #2907, opened 2020-11-19
Controls 0, 1, and 2 - Vectorize Controls_config, Refactor Control Presets, and Use Undosys

z64555:
This PR combines branches `feature/controls`, `feature/controls1`, and `feature.controls2`. This was done because`feature/controls` breaks two features which are fixed in `feature/controls1` and `feature/controls2`, and the three should not be merged in individually.

<hr>

PR #2868 Controls0 - `feature/controls`

This changes the Controls_config array into an SCP_vector to allow scripting to add more controls as they wish, enabling mod devs to create new controls with scripting and not needing hardcode support in the future.  (Custom controls are still a ways off, however)

- [x] Bind mode behaves as it should
- [x] Search mode behaves as it should
- [x] Browse mode behaves as it should
- [x] Loading controlconfigdefaults with retail data behaves as it should
- [x] Loading controlconfigdefaults with mod data behaves as it should

<hr>

PR #2884 Controls 1 - `feature/controls1`

This re-factors control presets and adds a bunch of features.

* Hardcoded defaults are now a preset.
* Mod devs may override the hardcoded defaults with the controlconfigdefaults.tbl by naming a `#ControlConfigOverride` section as "default"
* Mod devs may add as many presets as they like with `#ControlConfigPreset` sections.  These have the same syntax and structure as `#ControlConfigOverride` but only add bindings, and start off as a blank slate.
* Added ability to change the name of controls as they show up in the menu using `$Text:`
* Added possibility of resumption, should a `$Bind Name:` not be found within the controls.

- [x] Reset to defaults should behave as it should
- [x] Repeat hitting of "Reset" button should cycle through all presets
- [x] controlconfigdefaults.tbl is able to override the default preset
- [x] controlconfigdefaults.tbl is able to make new presets with the #ControlConfigOverride section
- [x] controlconfigdefaults.tbl is able to make new presets with the #ControlConfigPreset section
- [x] controlconfigdefaults.tbl is able to relabel controls

<hr>

PR #2907 Controls 2 - `feature/controls2`

Introduces and improves the undo system from PR #1243 and uses it for the controls config menu to re-enable the undo feature.
- [x] Should be able to undo single buttons
- [x] Should be able to undo single axis bindings
- [x] Should be able to undo preset cycling
- [x] Should be able to undo multiple button/control changes (such as an entire wipe of the controls)
- [x] Should be able to do all vanilla functions normally

### DIFF
--- a/code/controlconfig/controlsconfig.cpp
+++ b/code/controlconfig/controlsconfig.cpp
@@ -821,7 +821,7 @@ int control_config_do_reset()
 	// first, determine how many bindings need to be changed
 	for (size_t e = 0; e < Control_config.size(); ++e) {
 		auto item = Control_config[e];
-		auto default_item = Control_config_presets[Defaults_cycle_pos].bindings[i];
+		auto default_item = Control_config_presets[Defaults_cycle_pos].bindings[e];
 
 		if (item.disabled) {
 			// skip
--- a/code/controlconfig/controlsconfig.cpp
+++ b/code/controlconfig/controlsconfig.cpp
@@ -824,22 +824,17 @@ int control_config_do_reset()
 	return 0;
 }
 
-/**
- * @brief Resets all controls to values within the currently selected preset
- */
 void control_config_reset_defaults()
 {
-	int i;
-
 	// Reset keyboard defaults
 	const CC_preset &preset = Control_config_presets[Defaults_cycle_pos];
-	for (auto i = 0; i < Control_config.size(); ++i) {
+	for (size_t i = 0; i < Control_config.size(); ++i) {
 		Control_config[i].key_id = preset.bindings[i].first.btn;
 		Control_config[i].joy_id = preset.bindings[i].second.btn;
 	}
 
 	// Reset joy defaults.  No presets for joysticks currently
-	for (i=0; i<NUM_JOY_AXIS_ACTIONS; i++) {
+	for (size_t i = 0; i < NUM_JOY_AXIS_ACTIONS; i++) {
 		Axis_map_to[i] = control_config_axis_default(i);
 		Invert_axis[i] = Invert_axis_defaults[i];
 	}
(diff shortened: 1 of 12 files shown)


z64555:
Noticed minor bug with drift.  Not sure if due to controller or something else.

z64555:
Ok, I seem to have resolved all compile/build errors, so now that's all thats left is the automated tests.

I'm not sure why this is failing, as this group does minimal changes to the player files.

code/controlconfig/controlsconfig.cpp
```diff
@@ -171,14 +173,18 @@ static int Background_bitmap;
...
 static UI_WINDOW Ui_window;
-static unsigned int Defaults_cycle_pos; // the controls preset that was last selected
+static unsigned int Defaults_cycle_pos = 0; // the controls preset that was last selected
 
 int Control_config_overlay_id;
 
-static struct {
+struct conflict {
 	int key;  // index of other control in conflict with this one
 	int joy;  // index of other control in conflict with this one
```
asarium, line 182: You can initialize the members directly as well:

```suggestion
	int key = -1;  // index of other control in conflict with this one
	int joy = -1;  // index of other control in conflict with this one
```
Then you do not need to explicitly declare the default constructor.

asarium (approved):
To be honest, this is too big to effectively review. I'm going to approve it but we should seriously think about including it in the upcoming release since there is a relatively large chance for a regression to happen.

JohnAFernandez:
I'm with waiting.  Not really enough time to test and fix.  But merging right after stable gives us more than enough time to test and allows users to start taking advantage of these new features.

code/controlconfig/controlsconfig.cpp
z64555: I think one of the CI's complained when I did that, most likely clang tidy.

Goober5000 (changes requested):
Agreed, we should not merge this until after the stable release.  I'm putting this in the Request Changes state so that it's not merged accidentally.

Goober5000 (approved):

z64555:
there's a number of issues that arose in testing this, and one of them is a freeze when pressing Clear All.

z64555:
Latest commit should be green lit.  Ran through all checkboxes here.  We'll see how Controls3 will end up after this.

PR #2909, opened 2020-11-19
Add ships.tbl flags to tune most aspects of ETS behaviour

### LINKED ISSUE #2852: ETS transfer rates are not customizable
As a vintage 90s space sim, FS2 features an energy transfer system that allows the user to shunt energy from shields to weapons and back again. The transfer can trade 1:1 from shields to weapons but 2:1 from weapons to shields; these values are hardcoded and shouldn't be.

A table option should be added to allow this ratio to be customized; I'll leave it to community feedback and the implementor (implementress?) to define where this option is best placed.

PhantomHoover:
Fixes #2852

### DIFF
--- a/code/ship/ship.h
+++ b/code/ship/ship.h
@@ -1129,6 +1129,14 @@ public:
 	float		max_shield_regen_per_second;	// Goober5000 - max percent/100 of shield energy regenerated per second
 	float		max_weapon_regen_per_second;	// Goober5000 - max percent/100 of weapon energy regenerated per second
 
+	// Fields for tuning the ETS' direct shield<->weapon transfer feature
+	float		shield_weap_amount;				// fraction of shield capacity to transfer
+	float		shield_weap_efficiency;			// efficiency multiplier for output into weapons capacitor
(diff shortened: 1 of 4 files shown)


code/ship/ship.h
```diff
@@ -1129,6 +1129,13 @@ class ship_info
 	float		max_shield_regen_per_second;	// Goober5000 - max percent/100 of shield energy regenerated per second
 	float		max_weapon_regen_per_second;	// Goober5000 - max percent/100 of weapon energy regenerated per second
 
+	float		shield_weap_amount;
+	float		shield_weap_efficiency;
+	float		shield_weap_speed;
+	float		weap_shield_amount;
+	float		weap_shield_efficiency;
+	float		weap_shield_speed;
```
Baezon, line 1137: Could we get some comments here? Nothing complicated, just so its clear this is referring to the ets system, and what these values do, like the speeds are default 0.04, 0.04 of what?

Baezon (changes requested):

Baezon (approved):
LGTM

PR #2911, opened 2020-11-21
Only save hotkeys if the player has actually added to the group, otherwise leave mission defaults in place

PhantomHoover:
This very very slightly alters existing behaviour, in that hotkeys the player has *removed* will not persist the way they used to, but it fixes the far greater outstanding problem of mission-specified hotkeys, a very widely used tool nowadays, being dropped very frequently between replays.

Edit: after revision the fix will now reflect the old (weird) behaviour on any group the player has actually added to in the first 15 seconds of the mission, but otherwise will use the mission defaults.

### DIFF
--- a/code/mission/missionhotkey.cpp
+++ b/code/mission/missionhotkey.cpp
@@ -403,6 +403,7 @@ void mission_hotkey_maybe_save_sets()
 	int i;
 	htarget_list	*hitem, *plist;
 	HK_save_info *hkp;
+	bool found_player_hotkey;
 
 	if ( !timestamp_elapsed(Mission_hotkey_save_timestamp) ) {
 		mission_hotkey_maybe_restore();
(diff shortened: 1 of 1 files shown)


Goober5000:
> in that hotkeys the player has *removed* will not persist the way they used to

Does this mean that if the hotkey assigns F5 to wing A, and the player removes it from wing A and assigns it to wing B, then if the player restarts the mission the hotkey will apply to both A and B?

PhantomHoover:
Yes. The existing hotkey saving logic is already extremely unintuitive and essentially not fit for purpose: it saves hotkey assignments 15 seconds into the mission and then ignores other sources of hotkey binds on restart, and it is unable to distinguish between assignments deliberately removed by the user and those for ships which were not present in the mission. The change from retail behaviour is so marginal that I don't think it justifies holding this fix up.

z64555 (approved):

PR #2914, opened 2020-11-21
Add 'spout' type particle effect

Baezon:
Instead of randomly making particles in a sphere or cone, this effect makes them in the direction of a randomly walking vector contained within a sphere or cone, the 'spout'. For @asarium's note, or anyone else interested in the structure of the particle effect system, this adds the `m_effectData` pointer to particle sources, to be used by whatever the effect in question wants. Sources did not before have effect-specific data, but its fundamentally required for this type of effect in order to have per-source memory regarding the current position of the spout. And I do think adding this opens up many good doors in the future as far as what can be achieved by particle effects.

### DIFF
--- a/code/particle/ParticleSource.h
+++ b/code/particle/ParticleSource.h
@@ -263,8 +263,8 @@ class SourceTiming {
  * @brief A particle source
  *
  * A particle source contains the information about where and for how long particles are created. A particle effect uses
- * this information to create new particles. A particle source has not effect-specific information which means that an
- * effect can only use the information contained in this object.
+ * this information to create new particles. m_effectData is only effect-specific information that a 
+ * particle source has.
  *
  * @ingroup particleSystems
  */
--- a/code/particle/ParticleSource.h
+++ b/code/particle/ParticleSource.h
@@ -282,6 +282,9 @@ class ParticleSource {
 
 	void initializeThrusterOffset(weapon* wp, weapon_info* wip);
  public:
+	 
+	std::shared_ptr<void> m_effectData; //!< a pointer to a data record specific to the effect type
+
 	ParticleSource();
 
 	inline const ParticleEffect* getEffect() const { return m_effect; }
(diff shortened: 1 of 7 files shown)


code/particle/ParticleSource.h
```diff
@@ -282,6 +282,9 @@ class ParticleSource {
 
 	void initializeThrusterOffset(weapon* wp, weapon_info* wip);
  public:
+	 
+	std::shared_ptr<void> m_effectData; //!< a pointer to a data record specific to the effect type
```
asarium, line 286: I'm sorry but this exactly what I did not want to do with this system since this now puts particle source data in random locations in memory. Instead, you could use a union (although that is a hassle to work with in C++) or a variant (via https://github.com/mpark/variant) to store effect specific data directly in the source.

asarium (changes requested):

code/particle/ParticleSource.h
Baezon: I think a union should be ok for now, since we can use the `particleEffect` as the descriminant. And if locality is the goal, I assume that a vector (of spouts, what 'spout' types use `m_effectData` for) is probably out too? Using an array and a maximum number of spouts per source isn't the worst thing in the world, but obviously it would be nice to keep the unboundedness.

wookieejedi: Just out of curiosity any update on a consensus @asarium?

Baezon: I'll need to change it to a union and with an array for the spouts, I think. I've just left this one hanging for awhile.

asarium: Another option might be to just use the actions system for this. You could have actions that create single effects that would then act like the "spouts" this wants to implement.

Baezon:
I don't think this can really be elegantly done within the constraints of the particle system, or if it can I'm not skilled enough to do it.

PR #2915, opened 2020-11-21
Warn about and correct invalid squadron wings in multi

JohnAFernandez:
So, this should actually *fix* invalid squadron names in multi.  Already tested as working for the test case.  Cannot seem to get the warning to show up, though...

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -4757,12 +4757,14 @@ void parse_wings(mission* pm)
 
 	bool found = false;
 
+	static_assert(MAX_TVT_WINGS_PER_TEAM == 1, "Unless you also update the section of code below or redo the loadout code, for TvT, there should be just one player wing, otherwise, wings may start disappearing in game.");
+
 	// now see if we found the missing wing.  We're looking for a wing that is there after a wing that is not.
 	// for now TvT mission do not have enough player wings to be affected by this bug.
 	if (pm->game_type & (MISSION_TYPE_MULTI | MISSION_TYPE_MULTI_COOP)) {
 		do {
 			found = false;
-			// we only search up to the MAX_STARTING_WINGS because non-starting wings should not be in starting wing indeces (0-2)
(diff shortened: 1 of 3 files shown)


JohnAFernandez:
Barring CI checks, this should be now be ready for review.

Tested:
Alpha, Gamma and no Beta -- Even with player switching to slot in Gamma.
Alpha, and Epilson

code/mission/missionparse.cpp
```diff
@@ -4729,15 +4740,44 @@ void parse_wing(mission *pm)
...
 	{
 		Assert(Num_wings < MAX_WINGS);
 		parse_wing(pm);
 		Num_wings++;
 	}
+
+	bool found = false;
+
+	// now see if we found the missing wing.  We're looking for a wing that is there after a wing that is not.
+	// for now TvT mission do not have enough player wings to be affected by this bug.
```
Goober5000, line 4761: I would add an Assertion here to check that the number of TvT wings is under the bug threshold.  In the event that someone bumps TvT wings in the future, he may not think to check here.

code/mission/missionparse.cpp
```diff
@@ -4729,15 +4740,44 @@ void parse_wing(mission *pm)
...
+	if (pm->game_type & (MISSION_TYPE_MULTI | MISSION_TYPE_MULTI_COOP)) {
+		do {
+			found = false;
+			// we only search up to the MAX_STARTING_WINGS because non-starting wings should not be in starting wing indeces (0-2)
```
Goober5000, line 4765: It's spelled "indices" or "indexes".

Goober5000 (changes requested):
Some minor things.

code/mission/missionparse.cpp
JohnAFernandez: Smart. Thanks for the review!

JohnAFernandez: See what you think of that assert.

Goober5000: Looks good!

Goober5000 (approved):

PR #2916, opened 2020-11-22
Add 'Dont_override_old_maneuvers' flag to maneuvers

Baezon:
Maneuvers normally clear any existing maneuvers before taking effect, with this flag it will not do that. Though it does refresh the duration of the old maneuver, which is unideal, this allows for independent sexp/scripting control over lateral and rotational maneuvers at the same time, which is incredibly useful.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -14227,8 +14227,28 @@ static void ai_control_info_check(ai_info *aip, ship_info *sip, physics_info *pi
 	if (aip->ai_override_flags.none_set())
 		return;
 
-	if (!aip->ai_override_flags[AI::Maneuver_Override_Flags::Never_expire] && timestamp_elapsed(aip->ai_override_timestamp))
+	bool no_lat = false;
+	bool no_rot = false;
+
+	if (!aip->ai_override_flags[AI::Maneuver_Override_Flags::Lateral_never_expire] && timestamp_elapsed(aip->ai_override_lat_timestamp))
 	{
+		aip->ai_override_flags.remove(AI::Maneuver_Override_Flags::Sideways);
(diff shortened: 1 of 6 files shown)


Goober5000 (approved):

asarium (approved):

Goober5000:
> Though it does refresh the duration of the old maneuver

You may want to separate the timestamps for lateral and rotational maneuvers, which would solve this problem.

Baezon:
Ah that's true, that would be a fairly effective but lightweight way to do it, although it would be more awkward to describe in the sexp doc...

Baezon:
The code is a bit more cumbersome this way, but it would definitely be more convenient for the modder.

code/ai/aicode.cpp
```diff
@@ -14227,8 +14227,28 @@ static void ai_control_info_check(ai_info *aip, ship_info *sip, physics_info *pi
 	if (aip->ai_override_flags.none_set())
 		return;
 
-	if (!aip->ai_override_flags[AI::Maneuver_Override_Flags::Never_expire] && timestamp_elapsed(aip->ai_override_timestamp))
+	bool no_lat = false;
+	bool no_rot = false;
+
+	if ((!aip->ai_override_flags[AI::Maneuver_Override_Flags::Lateral_never_expire] && timestamp_elapsed(aip->ai_override_lat_timestamp)))
```
Goober5000, line 14233: Any reason you have an extra set of parentheses here?

Goober5000 (approved):
New code looks good.  There's one comment that's more of a nitpick than anything.

code/ai/aicode.cpp
Baezon: ah whoops, used to be part of a bigger expression

PR #2917, opened 2020-11-22
Mission flag to disable the EMP scrambling effect on messages

PhantomHoover:


Goober5000 (changes requested):
If you don't want to scramble comms when the EMP is active, then the mission flag check should go in the `hud_communications_state` function, in hudtarget.cpp:4603.  What you've coded here is essentially "never scramble comms under any circumstances".

Also, this PR is missing not only the FRED code but also to load the mission flag using the mission parse code.

PhantomHoover:
I'm aware that the necessary FRED code is missing but I am basically unable to add it; the Windows UI changes needed require a dev who has the necessary tools.

I have directly tested As Lightning Fall by manually changing the flag in the mission file, and distortion appeared or disappeared as expected. I'm pretty confident this works properly on the game client side.

Goober5000:
Closing in favor of #6141.

PR #2918, opened 2020-11-22
Fix multiplayer stat save

JohnAFernandez:
When pressing escape, it was previously possible to attempt saving the stats multiple times, but this will keep it to just one attempt.

### DIFF
--- a/code/network/multiui.cpp
+++ b/code/network/multiui.cpp
@@ -718,6 +718,7 @@ int Multi_join_should_send = -1;
 // master tracker details
 int Multi_join_frame_count;						// keep a count of frames displayed
 int Multi_join_mt_tried_verify;					// already tried verifying the pilot with the tracker
+bool Multi_already_tried_stats_submit;
 
 // data stuff for auto joining a game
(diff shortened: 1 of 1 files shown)


notimaginative:
Not sure that I like this. The point of that flag was to indicate that stats were stored, not that they *might* have been stored.

I may have misunderstood the problem of at which point the ESC key is hit. If it's on the debriefing screen then I think that's a different issue. If it's actually during stats save this probably needs a fix in `multi_fs_tracker_store_stats()` instead.

notimaginative (changes requested):
I'd rather the netgame flag remain as-is so that it's still relevant.

But it should be possible to just add a boolean local to multiui which accomplishes the same result that this change does. Then use that boolean instead of the `if ( !(Netgame.flags & NG_FLAG_STORED_MT_STATS) )` check in both `multi_debrief_accept_hit()` and `multi_debrief_esc_hit()`.

Otherwise I haven't found a better way to address this than what you're attempting .

JohnAFernandez:
All right, how about this?

notimaginative (approved):

notimaginative:
LGTM, thanks!

PR #2922, opened 2020-11-24
Refactor hud escort list to be less crazy

### LINKED ISSUE #1674: Feature request: Multiplayer Keep Players in Escort List after respawn
Right now if you add a player in your escort list it will remain there until it dies, but no one wants that, if i add my friend on the escort list it is because i want to keep an eye on him all the time and find him fast, having player removed after they die in multiplayer kinda kill the purpose of adding players to escort list in multi.

Also displaying his nick as well as his wing name on Escort list will be good.

notimaginative:
This should make the escort list code a little easier to deal with and be a bit more optimized.

 - avoid rebuilding the list constantly
 - add and remove ships more efficiently
 - better handling of player ships in multi (fixes #1674)

### DIFF
--- a/code/hud/hudescort.cpp
+++ b/code/hud/hudescort.cpp
@@ -486,9 +486,7 @@ static bool escort_compare(const escort_info &escort1, const escort_info &escort
 			return false;
 		}
 
-		if (Net_players[n1].m_player->stats.m_kill_count_ok > Net_players[n2].m_player->stats.m_kill_count_ok) {
-			return true;
-		}
+		return (Net_players[n1].m_player->stats.m_kill_count_ok >= Net_players[n2].m_player->stats.m_kill_count_ok);
 	} else {
 		// in multi, players go to the top
 		if (Game_mode & GM_MULTIPLAYER) {
--- a/code/hud/hudescort.cpp
+++ b/code/hud/hudescort.cpp
@@ -718,6 +718,24 @@ void hud_escort_view_toggle()
 // try to add a ship to the escort list, if slot available
 void hud_add_ship_to_escort(int objnum, int supress_feedback)
 {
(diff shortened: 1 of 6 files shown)


notimaginative:
This needs a lot of testing to make sure there isn't some big regression. Leaving as draft until then.

code/hud/hudescort.cpp
```diff
@@ -391,99 +462,77 @@ void hud_escort_init()
...
-			} else {
-				ret = 1;
-			}
+		if (Net_players[n1].m_player->stats.m_kill_count_ok > Net_players[n2].m_player->stats.m_kill_count_ok) {
+			return true;
 		}
 	} else {
-		diff = escort2->priority - escort1->priority;
+		// in multi, players go to the top
+		if (Game_mode & GM_MULTIPLAYER) {
```
JohnAFernandez, line 493: Maybe we should limit this behavior to Dogfight mode?  If there is a high number of respawns and a very important objective, like a mission critical ship, I know which one I would rather have on the escort list.

notimaginative: Dogfight mode is already handled differently and I tried to leave that special handling alone. This specific change should only affect player ships that are manually added, so my thinking was that those ships were particularly important.

It's kind of iffy though and I'm still not sure if high priority mission objectives should come first or not. Needs testing and further discussion.

wookieejedi:
Tested with many different situations, including manually adding and removing entries, adding and removing hotkeys, setting escort values in the ship editor and via sexp. Everything works as expected.

wookieejedi (approved):
Okay tested on a coop multi mission with my laptop and desktop setup. Removing and adding names worked, clearing escort list worked, and escort list also worked as expected with mission assigned escort ships, including players. With that and the extensive single-player testing I'll give this an approval.

PR #2923, opened 2020-11-24
Controls3 - Primary/Secondary Bindings

z64555:
Depends on #2907 
Migrates key_id and joy_id to CC_binds of Primary and Secondary, in preparation for multi-controller support.

Refactors and adds functionality to the retail control config menu:
* Bound keys and buttons are now in two separate columns, to reflect the move to Primary and Secondary bindings.
* Expanded upon the retail ability to select the bindings, you can now select the Primary or Secondary binding by clicking on it or using the left or right keys to select them on the selected line.
* You can now select a binding by clicking anywhere over its draw area, you are no longer required to click exactly on the visible text of the binding to select it
* Improved the color highlighting for selection, mouseover, and conflict states
* Tweaked the 640x480 and 1024x768 positioning so the new columns are readible.  640x480 is undoubtedly cramped, but usable.

- [x] Bind mode behaves as it should
- [x] Search mode behaves as it should
- [x] Browse mode behaves as it should
- [x] Undo system behaves as it should
- [x] Presets system behaves as it shoud
- [x] Can select a primary or secondary item (with arrow keys and mouse), and directly clear it
- [x] Can select a primary or secondary item (with arrow keys and mouse), and directly bind it
- [x] Can select the action itself, and clear both bindings
- [x] Search mode UI should highlight the primary or secondary binding
- [x] All bindings fire as they should in-game
- [x] Custom controls behave as they should


z64555:
Merge conflicts were detected, reviewers please be diligent in potential errors that may have slipped past due to the merge.

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -4106,10 +4106,11 @@ sexp_list_item* sexp_tree::get_listing_opf_ship_type() {
 sexp_list_item* sexp_tree::get_listing_opf_keypress() {
 	int i;
 	sexp_list_item head;
+	auto& Default_config = Control_config_presets[0].bindings;
 
 	for (i = 0; i < CCFG_MAX; i++) {
-		if (Control_config[i].key_default > 0 && !Control_config[i].disabled) {
-			head.add_data_dup(textify_scancode(Control_config[i].key_default));
+		if ((Default_config[i].get_btn(CID_KEYBOARD) != -1) && !Control_config[i].disabled) {
+			head.add_data_dup(textify_scancode(Default_config[i].get_btn(CID_KEYBOARD)));
```
Goober5000, line 4113: The code in qtFred is somewhat different than the code in Fred.

z64555: Hopefully addressed.

z64555:
CLI automation cleared, need completion of checkbox list.

z64555:
added a checkbox item due to issues reported with Between the Ashes.  The scripted custom controls were not activating as they should have with the changes done by Controls2 due to the change in how default bindings and how scripts, sexps, check for key activation.

Controls3 should address this since it strips out the old default bindings in the CCI and installs the new primary/secondary scheme.

wookieejedi:
Everything works in the checklist except for one thing. When I in the `Ship` controls tab and attempt the bind a key that already has a binding, then I am unable to click `Clear Conflict`. This inability to clear conflict bug only occurs on the `Ship` tab.

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1396,7 +1393,7 @@ void LoadEnumsIntoActionMap() {
 	ADD_ENUM_TO_ACTION_MAP(CUSTOM_CONTROL_5)
 #undef ADD_ENUM_TO_ACTION_MAP
 
-	assert(mActionToVal.size() == CCFG_MAX);
+	Assert(mActionToVal.size() == CCFG_MAX);
```
JohnAFernandez, line 1396: Can we do an assertion here instead?  Helps users get the message that these are errors that need to be reported.

JohnAFernandez (changes requested):
While it's true this is impossible to properly review, I was able to find a few things that do need to be changed.  We should not be adding any int3()'s.

code/controlconfig/controlsconfigcommon.cpp
z64555: Most of the Asserts in the controls code is on hardcode data and not userdata, so if an error occurs its likely due to an oversight done by the coder working on the area.  I've upgraded a number of Assertions in *common to make it more descriptive.

JohnAFernandez (approved):

Goober5000:
This is really too big to effectively review, but I trust wookieejedi's testing and Cyborg's comments.  I'm willing to merge this if there are no additional comments in the next day or two.

PR #2924, opened 2020-11-24
Death roll for big ships and a death roll multiplier

Baezon:
Two simple additions, a flags that makes big|huge ships do a deathroll (running AI on them despite being dying clobbers their `desired_rotvel` normally), and a multiplier for the intensity of the death roll rotation.


JohnAFernandez (approved):
Looks good! Nice addition

PR #2925, opened 2020-11-24
Separate ship splitting explosions from propagating explosions

Baezon:
Another small death-related addition: whether a ship splits in two at the end of its death roll can now be set separately from a propagating explosion.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -3284,7 +3284,7 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 		stuff_boolean(&sip->explosion_splits_ship);
 	}
 	else if (first_time) {
-		sip->explosion_splits_ship = sip->explosion_propagates;
+		sip->explosion_splits_ship = sip->explosion_propagates == 1;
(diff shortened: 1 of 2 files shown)


code/ship/ship.cpp
```diff
@@ -8336,8 +8346,8 @@ static void ship_dying_frame(object *objp, int ship_num)
 				}
 			}
 
-			// If this is a large ship with a propagating explosion, set it to blow up.
-			if ( ship_get_exp_propagates(shipp) )	{
```
JohnAFernandez, line 8340: Wait.  So what happens to this function, and where food the code now use propagating explosions?

Baezon: It's still there (it just returns `sip->explosion_propagates`), and it's still used for the main thing everyone expects to be; enabling those smaller explosions before the big one.

JohnAFernandez: All right, the changes seem straight forward enough.  If you're sure it works, let's get it merged.

JohnAFernandez (approved):

PR #2927, opened 2020-11-25
Reduce docking 'jitter'

Baezon:
`find_adjusted_dockpoint_info()` used for various docking procedures to determine the situation of a dockpoint, notably `DOA_DOCK_STAY` in order to stay rigidly docked to an object, normally returns `docker_p0` `docker_p1` and `docker_p0_norm`, *global* positions (and global directional vector). From these the 'true' position of the dockpoint and its orientation is calculated; its uvec for instance is the p1->p0 directional vector. However since these are global positions being used to create a normalized vector, they can be quite sensitive to floating point inaccuracy when the distances are great (only 20km or so for Mystery of the Trinity can occasionally cause this to be visible).

Since every caller of `find_adjusted_dockpoint_info()` uses these points to make the position and orientation from them, it is numerically safer for the function to do it itself, while they're still local and accurate, and then explicitly exports the orient (rather than implicitly through the points and normal).

This makes for no unusual amounts of jittering or oscillation up to 100km, even beyond I did not observe indefinite oscillation, although everything starts to get a bit jittery at that point.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -9055,7 +9055,7 @@ float dock_move_towards_point(object *objp, vec3d *start, vec3d *finish, float s
 //	Set the orientation in the global reference frame for an object to attain
 //	to dock with another object.  Resultant global matrix returned in dom.
 // Revised by Goober5000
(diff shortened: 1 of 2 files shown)


Goober5000 (approved):
Great idea.

PR #2929, opened 2020-11-26
Expose the power output in the ship class to sexps and scripting

The-E:


asarium (approved):

code/parse/sexp.cpp
```diff
@@ -16870,6 +16871,20 @@ int sexp_gse_recharge_pct(int node, int op_num)
...
+* Get a ship's power output
+*/
+int sexp_get_power_output(int node) 
+{
+	node = CDR(node);
+	auto ship_entry = eval_ship(node);
+
+	if (ship_entry == nullptr)
+		return SEXP_NAN_FOREVER;
+	return (int)(std::lround(Ship_info[ship_entry->shipp->ship_info_index].power_output));
```
Goober5000, line 16884: If `ship_entry->status` is `NOT_YET_PRESENT` or `EXITED`, then `ship_entry->shipp` will be a `nullptr`.

Goober5000 (changes requested):

code/parse/sexp.cpp
Goober5000: Your first commit was correct.  If `ship_entry` is null, then the ship is completely unknown to the mission, so `SEXP_NAN_FOREVER` is the appropriate return value.

If `ship_entry` is non-null, then it is a valid ship which may be either present, exited, or not-yet-arrived.  Of those three, the `shipp` field is only valid when the status is `PRESENT`.

The-E: Thanks, let this be a lesson (if any lesson were needed) that coding after dark is bad....

Goober5000 (approved):

PR #2938, opened 2020-11-28
Add "first-of" SEXP that behaves like "any-of" with a cap.

MageKing17:
It takes a number as a maximum number of arguments that can be evaluated in a single call, but otherwise behaves very similarly to `any-of`; alternatively, if the number is 1, it's like an `in-sequence` that also skips arguments that don't satisfy the condition instead of just invalid ones.


Goober5000 (approved):
Looks good.

PR #2941, opened 2020-11-29
Fix a small copy paste error in weapon parsing

Baezon:


JohnAFernandez (approved):
As future modders would be confused by the fact that their table entry is not working, this should be fixed.  Any modders negatively affected by this change should fix their entries.

MageKing17 (approved):

PR #2942, opened 2020-11-29
Don't short circuit eval_sexp on nodes which contain the special argument.

PhantomHoover:
This implements (a more efficient version of) the solution I outlined to the problem in this thread: https://www.hard-light.net/forums/index.php?topic=97146

Most (all?) sexps that touch the .cache element of a sexp node are careful to make sure that it isn't dependent on the special argument before doing so. This simply extends those checks to the short-circuit values in .value. Performance impact should be minimal: it's just checking a flag. Testing consisted of playing Her Finest Hour a lot and confirming it didn't break.


code/parse/sexp.cpp
```diff
@@ -23109,13 +23110,13 @@ int eval_sexp(int cur_node, int referenced_node)
 	}
 
 	// now do a quick return whether or not we log, per the comment above about trapping known sexpressions
-	if (Sexp_nodes[cur_node].value == SEXP_KNOWN_TRUE) {
+	if (Sexp_nodes[cur_node].value == SEXP_KNOWN_TRUE && !(Sexp_nodes[cur_node].flags & SNF_SPECIAL_ARG_IN_NODE)) {
```
JohnAFernandez, line 23113: I'm still the rookie but, we could nest these three conditionals inside a check for !(Sexp_nodes[cur_node].flags & SNF_SPECIAL_ARG, since they all check the same thing.

PhantomHoover: Good shout, I really didn't like how ungainly those conditionals were. PR updated.

code/parse/sexp.cpp
```diff
@@ -23094,29 +23094,32 @@ int eval_sexp(int cur_node, int referenced_node)
...
+	// we can't 'know' its value so we skip this behaviour.
 
 	// we want to log event values for KNOWN_X or FOREVER_X before returning
-	if (Log_event && ((Sexp_nodes[cur_node].value == SEXP_KNOWN_TRUE) || (Sexp_nodes[cur_node].value == SEXP_KNOWN_FALSE) || (Sexp_nodes[cur_node].value == SEXP_NAN_FOREVER))) {
-		// if this is a node that has been assigned the value by short-circuiting,
-		// it might not be the operator that returned the value
-		int op_index = get_operator_index(cur_node);
-		if (op_index < 0)
-			op_index = get_operator_index(CAR(cur_node));
+	if (!(Sexp_nodes[cur_node].flags & SNF_SPECIAL_ARG_IN_NODE)) {
```
MageKing17, line 23101: Why are you only checking this one node instead of using `special_argument_appears_in_sexp_tree()`, as I thought you were planning based on the Discord discussion? Surely, if short-circuiting `( is-destroyed-delay "<argument>" 0 )` is bad, then short-circuiting `( or ( is-destroyed-delay "<argument>" 0 ) ( is-departed-delay "<argument>" 0 ) )` is also bad.

Goober5000 (changes requested):
There appears to be some sort of confusion on how `when-argument` is used.  The existing behavior is the correct behavior. If `when-argument` is in an event by itself, with no repeats or trigger counts, then it should only execute *once*.  It should not continue to execute for all of its arguments.

In other words, the behavior described in this post...
https://www.hard-light.net/forums/index.php?topic=97146.msg1905733#msg1905733
...is correct, and is the way `when-argument` has behaved since it was originally implemented.  If it is desired to have the event *continue* to evaluate its sexps even after the event has been satisfied, then `every-time` should be used.

MageKing17:
> If `when-argument` is in an event by itself, with no repeats or trigger counts, then it should only execute _once_.

Nobody has ever, at any point, said that there were no repeats or trigger counts.

code/parse/sexp.cpp
PhantomHoover: Ack, you're right. I think I meant for SNF_SPECIAL_ARG_IN_TREE... but at this stage it'd be better just to go back to using special_argument_appears_in_sexp_tree().

Goober5000 (approved):

PhantomHoover:
It makes absolutely no sense to short-circuit a sexp node based on its result when evaluated with a different special argument. Goober, your example of 'intended' behaviour appears to have nothing at all to do with the semantics of this change.

Goober5000:
Yeah the confusion was on my end.  On further investigation this does appear to be a loose end that needs patching.

MageKing17 (approved):

Goober5000 (changes requested):
Wait, I want to put this on hold again.  The new code will skip the short-circuiting if the special-arg appears in *any node in the sexp tree*, which is not what we want.

Goober5000:
For this specific point:

> It makes absolutely no sense to short-circuit a sexp node based on its result when evaluated with a different special argument. 

It does make sense if the event is supposed to complete after a certain condition.  If the event is satisfied, it doesn't matter how many other arguments remain.

Now, the sexp nodes containing the arguments are flushed upon re-evaluation in `test_argument_nodes_for_condition`.  So, to the extent there is short-circuiting, it shouldn't apply to the argument nodes anyway.

Something else must be going on below the surface, or the interaction must be more complex than it appears.

Goober5000:
Ok, I believe the root cause is the short-circuiting within `eval_when` itself.  At the bottom of the function, `eval_when` will return `SEXP_NODE_FALSE` if it is short-circuiting.  However, as indicated, this isn't appropriate for `when`s that are being evaluated for multiple arguments.

I think all that needs to be done is to put the short-circuiting logic inside a test for `if (Sexp_current_argument_nesting_level > 0)` like so:
```
if (Sexp_current_argument_nesting_level > 0)
{
	// thanks to MageKing17 for noticing that we need to short-circuit on the correct node!
	int short_circuit_node = (arg_handler >= 0) ? arg_handler : cond;

	if (Sexp_nodes[short_circuit_node].value == SEXP_KNOWN_FALSE || Sexp_nodes[short_circuit_node].value == SEXP_NAN_FOREVER)
		return SEXP_KNOWN_FALSE;  // no need to waste time on this anymore
}
```

Goober5000 (approved):
Okay, after further analysis, with discussion on Discord, my concerns have been resolved.

PR #2946, opened 2020-12-01
Scripting Bugfix: previous HUD set/get gauge color referred to config instead of ingame gauge

SelfGamer:
Previously, the Scripting API functions setHUDGaugeColor and getHUDGaugeColor referred to HUD_config, which stores the configuration of the gauges by the user. However, modifying the configuration this way does not modify the gauge color in mission.
After this change, it does by accessing not the configuration but rather the currently loaded gauges.
In addition, a new virtvar is added to the HUD environment to allow scripts to query the number of existing gauges (so that this does not need to be hardcoded and subsequently be in danger of failing with an update of the gauges)

### DIFF
--- a/code/scripting/api/libs/hud.cpp
+++ b/code/scripting/api/libs/hud.cpp
@@ -53,6 +53,24 @@ ADE_VIRTVAR(HUDDisabledExceptMessages, l_HUD, "boolean", "Specifies if only the 
 		return ADE_RETURN_FALSE;
 }
 
+ADE_VIRTVAR(HUDDefaultGaugeCount, l_HUD, "number", "Specifies the amount of HUD gauges defined by FSO", "number", "The number of FSO HUD gauges")
+{
+	int amount = default_hud_gauges.size();
+
+	if (!ade_get_args(L, "*|i", &amount))
+		return ADE_RETURN_NIL;
+
+	if (ADE_SETTING_VAR)
+	{
+		hud_disable_except_messages(amount);
+	}
+
+	if (hud_disabled_except_messages())
+		return ADE_RETURN_TRUE;
+	else
+		return ADE_RETURN_FALSE;
+}
+
 ADE_FUNC(setHUDGaugeColor, l_HUD,
          "number gaugeIndex, [number red, number green, number blue, number alpha]",
          "Color used to draw the gauge", "boolean", "If the operation was successful")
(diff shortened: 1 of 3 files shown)


code/scripting/api/libs/hud.cpp
```diff
@@ -53,6 +53,24 @@ ADE_VIRTVAR(HUDDisabledExceptMessages, l_HUD, "boolean", "Specifies if only the
 		return ADE_RETURN_FALSE;
 }
 
+ADE_VIRTVAR(HUDDefaultGaugeCount, l_HUD, "number", "Specifies the amount of HUD gauges defined by FSO", "number", "The number of FSO HUD gauges")
```
asarium, line 56: I do not know what this is supposed to do. Could you elaborate on the use case for this?

code/scripting/api/libs/hud.cpp
```diff
@@ -61,15 +79,15 @@ ADE_FUNC(setHUDGaugeColor, l_HUD,
...
+	int a = 255;
 
 	if(!ade_get_args(L, "i|iiii", &idx, &r, &g, &b, &a))
 		return ADE_RETURN_FALSE;
 
-	if ((idx < 0) || (idx >= NUM_HUD_GAUGES))
+	if ((idx < 0) || (idx >= default_hud_gauges.size()))
 		return ADE_RETURN_FALSE;
 
-	gr_init_alphacolor(&HUD_config.clr[idx], r, g, b, a);
```
asarium, line 72: These changes (in both functions) do not cover the use case of saving the updated values to the pilot file which is currently done for the `HUD_config` object.

To support both use cases you could introduce new functions with your changed code to support this different use case.

asarium (changes requested):

code/scripting/api/libs/hud.cpp
SelfGamer: FSO defines a number of gauges, hardcoded, that are then rendered. This vector is, at the moment, expected to have 42 gauges, the HUD Config has 39 (I am not sure where the difference comes from, but hudparse.cpp defines 42 and hudgauges.h defines 39. I assume that hudparse has some gauges that were added over the years).
This virtvar returns the number of actually loaded gauges, so that in case a script needs to iterate over all existing gauges, it can do this with this variable, even if there may be further additions to the gauges in the future (as apparently, there have been). If this is guaranteed never to happen, this virtvar is unnecessary (then it should be noted in the docs that gauges range from index 0 to 41).

code/scripting/api/libs/hud.cpp
SelfGamer: Moved the new functionality to a separate function. Will still modify the old documentation as it is currently misleading

code/scripting/api/libs/hud.cpp
asarium: Ok, then this is not correct thus far.

1. You need to return the amount like this below: `return ade_set_args(L, "i", amount);` There should be other functions that use something similar so you can use that for reference.
2. Since this is effectively a "read-only" variable you can omit the `ADE_SETTING_VAR` part since there is no meaningful implementation for that anyway.

SelfGamer: Alright, changed it.

asarium (approved):

PR #2947, opened 2020-12-01
don't flush the tree when evaluating the *-argument conditionals

Goober5000:
This is a follow-up to PR #2942.  Given that the short-circuiting is now explicitly skipped for node trees containing the special argument, it should no longer be necessary to flush the conditional branch of the sexp tree.  (The .cache field is already skipped for special-arg nodes.)

The sexp tree is still explicitly flushed for `every-time` and `every-time-argument`.


MageKing17 (approved):
As a bonus, the conditional node should now short-circuit properly if it *doesn't* rely on `"<argument>"`.

PR #2950, opened 2020-12-03
Allow normal weapons to use 'fire down normals'

Baezon:
With `fire_down_normals` `ship_get_global_turret_info()` properly sets up `gvec` to be actually pointing in the direction the turret is, which is used by the turret firing code prior to this to determine things like being in fov, but when it comes to actually firing it doesn't bother and just fires straight at its target. This is fine normally, but given we have a specific flag to force firing down `gvec` its quite unintuitive that this won't be the case for normal weapons (it is only used by beams AFAICT).

Also a little readability change right next to it because it was bothering me. I dunno about you guys but if a big function call might need several different arguments, I prefer setting up the arguments beforehand, and calling the function once rather than a bunch of conditional calls to the function.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -2718,11 +2718,16 @@ void ai_fire_from_turret(ship *shipp, ship_subsys *ss, int parent_objnum)
 				something_was_ok_to_fire = true;
 				Num_turrets_fired++;
 
+				vec3d shoot_vector = v2e;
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Looks good.  I would make `shoot_vector` a pointer, but that's a nitpick.

I have an upcoming PR to clean up aiturret.cpp and fix a few bugs, so if you're semi-familiar with the turret code, I'd like you to take a look when I post it.

Issue #2953, opened 2020-12-03
Scripting FR: Keep Collision Overrides consistent for Ship-Ship and other collisions


SelfGamer:
After testing and looking through the code, I notice that even with an Override for a Ship-Ship collision set, the physics will still get applied (only the damage + UI effects get overridden). This is inconsistent with how Ship-Weapon / Ship-Debris / Ship-Asteroid collision overrides work, as those will also disable relevant physics.
Is changing Ship-Ship overrides to work like the other overrides is the correct approach? I would prepare a PR then.
If not, what is the intended way of nullifying Ship-Ship collisions from lua?

Goober5000:
Your PR looks like the correct thing to do.

SelfGamer:
PR #2953 merged, issue fixed

PR #2954, opened 2020-12-03
Modified Ship-Ship Collision Overrides to fit other Ship-X collision Overrides

SelfGamer:
Refers to Issue #2953 
All I've done to fix this, is move some code. Namely, the function ship_ship_check_collision (which is supposed to _check_ if there is a collision) contained some code to already handle and process the collision (push objects around and notify AI). This code, I've moved after the override checks of the lua API. I did not modify this code (apart from the renaming of variables that were passed down into the function and the subsequent direct accessing as opposed to dereferencing the passed pointers) and copied it as a whole block.
Note, that the other functions for detection of those collisions (debris_check_collision, asteroid_check_collision, and so on) do exactly that, and only do detection and not physics handling, which is then processed after checking the override, in a comparable manner as to how ship ship now works.

### DIFF
--- a/code/object/collideshipship.cpp
+++ b/code/object/collideshipship.cpp
@@ -1167,7 +1167,12 @@ int collide_ship_ship( obj_pair * pair )
 
 			if(!a_override && !b_override)
 			{
-				ship* light_shipp = &Ships[ship_ship_hit_info.heavy->instance];
+				//
+				// Start of a codeblock that was originally taken from ship_ship_check_collision
+				// Moved here to properly handle ship-ship collision overrides and not process their physics when overridden by lua
+				//
+
+				ship *light_shipp = &Ships[ship_ship_hit_info.heavy->instance];
 				ship *heavy_shipp = &Ships[ship_ship_hit_info.heavy->instance];
 
 				object* heavy_obj = ship_ship_hit_info.heavy;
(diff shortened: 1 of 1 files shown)


Baezon:
Generally looks good to me, I might want to see someone more qualified review whether this was indeed intentional, but otherwise ok if you've tested to make sure normal and overridden collisions work as they should now.

SelfGamer:
> Generally looks good to me, I might want to see someone more qualified review whether this was indeed intentional, but otherwise ok if you've tested to make sure normal and overridden collisions work as they should now.

Yeah, would probably be good to have someone familiar with this code check as well.
In my test scenario, it worked properly, but I only tested tens of collisions not hundreds.

Baezon:
Ah well its not about the number of collisions but the weird obscure edge cases we didn't think of when it comes to bugs. What you've already done should be fine.

Goober5000 (changes requested):
Looks good.  This was almost certainly an idiosyncratic implementation choice by Volition that didn't harm anything because of the original code flow.  When the scripting coder added the hook, it probably didn't occur to him to move the "valid hit occurred" code.

The only thing I would suggest is to put a comment above and below the block of code, saying that it was moved from its original location in `ship_ship_check_collision`.

SelfGamer:
> Looks good. This was almost certainly an idiosyncratic implementation choice by Volition that didn't harm anything because of the original code flow. When the scripting coder added the hook, it probably didn't occur to him to move the "valid hit occurred" code.
> 
> The only thing I would suggest is to put a comment above and below the block of code, saying that it was moved from its original location in `ship_ship_check_collision`.

Alright, marked the codeblock. Thanks for checking.

Goober5000 (approved):
:+1:

PR #2955, opened 2020-12-04
Turn framerate-independent turning on by default

PhantomHoover:
Asteroth's framerate-independent turning feature put a huge amount of work into remaining backwards-compatible with retail behaviour at 60FPS, and at this stage I've been playing with it turned on for everything for months and I've yet to find an identifiable bug due to it in any mod. I think it's definitely time to switch it on by default (leaving the table flags in place so that mods still have the option to disable it). The only thing it needs now is more thorough testing for full compatibility, and the best way to get that is to have it incorporated into the release candidate builds.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -654,7 +654,7 @@ void mod_table_reset()
 	Arc_color_emp_p2 = std::make_tuple(static_cast<ubyte>(128), static_cast<ubyte>(128), static_cast<ubyte>(10));
 	Arc_color_emp_s1 = std::make_tuple(static_cast<ubyte>(255), static_cast<ubyte>(255), static_cast<ubyte>(10));
 	Use_engine_wash_intensity = false;
-	Framerate_independent_turning = false;
+	Framerate_independent_turning = true;
 	Ai_respect_tabled_turntime_rotdamp = false;
 	Swarmers_lead_targets = false;
 	Required_render_ext.clear();


Baezon:
I think ill recuse myself from an explicit approval for obvious reasons, but I am, of course, in favor of this. It would bring some good stability/reliability to mission behavior/outcomes.

PhantomHoover:
Yeah, in case anyone's unfamiliar with the issue this fixes: at 120FPS, which FSO will default to on (increasingly common) high-refresh monitors, all missiles will turn *much* slower than at 60FPS, which is generally accepted to be the de facto standard that retail and subsequent mods are designed for. This isn't quite game-breaking but it obviously has very serious impacts on balance across the board, and FSO's quality and consistency on modern setups will benefit hugely from this change.

JohnAFernandez:
I am in favor.  Initially I was hesitant, but it's true.  This should be on by default

Goober5000:
I think this is reasonable.  Rule #1 of the SCP is to not change the default retail behavior, but in this case, there is *no* standard retail behavior.  It's a similar situation as the locking animation taking different times to lock on resolutions other than 640x480.  To the extent there was a sensible default, 640x480 was it.  Likewise, with this flag, to the extent there was a default, 60 FPS is an appropriate standard.

JohnAFernandez (approved):
Everyone has had a chance to weigh in.  So, looks like it's time.

PR #2957, opened 2020-12-04
Adds Fireball as a new Object Subclass to scripting

SelfGamer:
This allows creation and script-side management of warp-effects and explosions.
Adds:
fireballclass
fireball : object

tb.FireballClasses[]
mn.Fireballs[]

mn.createWarpeffect()
mn.createExplosion()

Create Warpeffect itself is implemented closely resembling Goober's warp-effect sexp.

Also, slightly corrects a simple documentation mistake for tb.WeaponClasses


code/scripting/api/libs/mission.cpp
```diff
@@ -955,6 +989,99 @@ ADE_FUNC(createWeapon,
...
+	int flags = warp3d ? FBF_WARP_VIA_SEXP | FBF_WARP_3D : FBF_WARP_VIA_SEXP;
+
+	duration = duration < 4.0f ? 4.0f : duration;
+
+	// sanity check, if these were specified
+	if (duration < opentime + closetime)
+	{
+		//Both warp opening and warp closing must occur within the duration of the warp effect.
+		opentime = closetime = duration / 2.0f;
+	}
```
asarium, line 1028: For these kind of validation checks it would be good to cause a `LuaError` when the necessary condition is not met to allow the script writer to see that there is something wrong and where the error is coming from.

asarium (changes requested):
Looks good! Just needs some minor usability enhancements in one function.

code/scripting/api/libs/mission.cpp
SelfGamer: Okay. Was thinking about this but wasn't sure if this was "worth" changing the behaviour of the function in comparison to it's sexp counterpart (that just does this).

asarium: Even for SEXPs it is, in my opinion, a good idea to issue a warning about these things. Silently changing the values is likely to cause some frustration for the writer. For scripts it is even more useful since `LuaError` also prints a stacktrace of from where exactly the wrong function call came from.

code/scripting/api/libs/mission.cpp
```diff
@@ -955,6 +989,99 @@ ADE_FUNC(createWeapon,
...
+	int flags = warp3d ? FBF_WARP_VIA_SEXP | FBF_WARP_3D : FBF_WARP_VIA_SEXP;
+
+	duration = duration < 4.0f ? 4.0f : duration;
+
+	// sanity check, if these were specified
+	if (duration < opentime + closetime)
+	{
+		//Both warp opening and warp closing must occur within the duration of the warp effect.
+		LuaError(L, "The duration of the warp effect must be higher than the sum of the opening and close durations");
+	}
```
asarium, line 1028: The duration check should also cause an error when that duration check fails. The same goes for the `IS_VEC_NULL_SQ_SAFE` check a bit below this.

Also, `LuaError` can still return normally if the user ignored the error so your previous error correction could would still be good to have here.

asarium (changes requested):

code/scripting/api/libs/mission.cpp
SelfGamer: Ah, thanks for the clarification

asarium (approved):

code/scripting/api/objs/fireballclass.cpp
```diff
@@ -0,0 +1,126 @@
...
+		return ade_set_error(L, "s", "");
+
+	if(ADE_SETTING_VAR && s != nullptr) {
+		auto len = sizeof(Fireball_info[idx].unique_id);
+		strncpy(Fireball_info[idx].unique_id, s, len);
+		Fireball_info[idx].unique_id[len - 1] = 0;
+	}
+
+	return ade_set_args(L, "s", Fireball_info[idx].unique_id);
+}
```
Goober5000, line 60: The fireball unique ID is not supposed to be modifiable.  It's supplied by fireballs.tbl or it's generated when the table is parsed.  Therefore, attempting to set the unique ID should cause an error.

Also, this virtual variable should be called "UniqueID", not "Name".  In fireballs.tbl, "Name" is the file name of the fireball animation.

code/scripting/api/objs/fireballclass.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+ADE_VIRTVAR(FireballFilename, l_Fireballclass, NULL, "Fireball class animation filename (LOD 0)", "string", "Filename, or empty string if handle is invalid")
```
Goober5000, line 62: "FireballFilename" is redundant because this is the fireball class.  Just call it "Filename".

code/scripting/api/objs/fireballclass.cpp
```diff
@@ -0,0 +1,126 @@
...
+	if(!ade_get_args(L, "o", l_Fireballclass.Get(&idx)))
+		return ADE_RETURN_NIL;
+
+	if(idx < 0 || idx >= Num_fireball_types)
+		return ADE_RETURN_FALSE;
+
+	return ADE_RETURN_TRUE;
+}
+
+ADE_FUNC(getFireballClassIndex, l_Fireballclass, NULL, "Gets the index value of the fireball class", "number", "index value of the fireball class")
```
Goober5000, line 114: This isn't intuitive because it's not clear what the "index value" is.  It looks like it's the number in the list of fireball types.  Maybe call it `getTableIndex`?  Or `getNumericIndex`?  Or `getIndexInList`?

Goober5000 (changes requested):
A few things.

SelfGamer:
Fair points. Changed

Goober5000 (approved):
Good to go. :+1:

Issue #2258, opened 2020-02-02
no ffb on version 19.0 work fine in 3.8 


luca2125:
hi,

I have updated to version 19.0 but the force feedback not work.

Back to version 3.8 the ffb work fine.

Can you check please ?

z64555:
Hey luca, are you still around?  Do you still have this issue?

The only force feedback controller I have is a busted up Xbox controller, and it needs some updates to the FFB code before it'll let FSO use it.

luca2125:
Hi z64555,

We have already discussed on discord "marco000" about:

https://discord.com/channels/@me/771053661882810380

are you working about FFB code ?

z64555:
I wanted to try to look into this, but the only force feedback controller I have is an Xbox controller.  FSO's looking for support of certain waveforms and hasn't been updated to use SDL_HAPTIC_LEFTRIGHT which is what Xinput controllers use.

So, as said before, in order for me to even test behavior I'd have to update the FFB code, maybe even overhaul it.

Most of the discussion in discord was over the GUID mismatch between knossos, which I should try to address in the controls code overhaul that's in review.

luca2125:
When I have connected only one device with force feedback (Sidewinder Force Feedback 2) the force feedback work. 

The problem is that sometimes (not always) after a update of windows or change the USB device the force feedback not work.

I can tell you that when work, work fine.

A bit detail but not very important, this game don't allow to set the force level (spring effects , overvall effects).

However the default setting is good for my device and for my preference I don't known for other users like to change ...

z64555:
Right, so. The intended workaround for this until forcefeedback can be properly addressed is to make use of the multi-joy feature that's upcoming.  You'd assign your haptic/force feedback joystick as joy-0, and all other joysticks/virtual joysticks as joy-1, joy-2, or joy-3 and bind them accordingly.

z64555:
Delayed until 21.4, when Controls4/5 gets merged.

luca2125:
It is a very good news if the force feedeback will be fixed in future, the time is not important.

z64555:
Delayed again until 21.6 due to issues with getting Controls5 merged.  This was a stressful summer...

luca2125:
thank you I will test when will be done.

z64555:
@luca2125 Please test with the latest RC: [Release 22.0 RC2](https://github.com/scp-fs2open/fs2open.github.com/releases/tag/release_22_0_0_RC2)

[...]

PR #2944, opened 2020-11-29
Controls4 - Multijoy

z64555:
The long awaited Multi joy feature. Multiple joysticks may be claimed by FSO from within the fs2_open.ini using the new keys:

```
Joy0ID=
Joy0GUID=
Joy1ID=
Joy1GUID=
Joy2ID=
Joy2GUID=
Joy3ID=
Joy3GUID=
```
Should "CurrentJoystick" be found, it'll assume only one joystick, so you'll have to manually edit fs2_open.ini to get this feature to work until the launchers are updated.

Joysticks show up in game as "Joy-0" with the respective bindings.

This provides significant changes over #2923 as well 
* Axis actions merged into the Control_config vector
* ADE function for inverting by action, instead of by axis
* Separate Mouse entirely from Joystick bindings. Shows up as "Mouse Button 1" etc.
* Removes saving bindings to .CSG, retaining only .PLR bindings.
  * This will be addressed with the ability to save to preset files in the future.

TODO: strings.tbl additions for axis labels, mouse and joystick button labels, controller labels

Issue #623, opened 2016-06-06
FR: Allow Multiple Gamepads/Joysticks


mikeloeven:
Recently i got a new HOTAS stick and was dying to try it out with FS2 Open however it seems that there is a major flaw in the game that prevents you from mapping controls from more than one Gamepad type device. Considering that a lot of HOTAS sticks are two separate devices the game needs to be updated to allow it to map inputs from more than one Gamepad device.

niffiwan:
This is a known issue. SDL2.0 is planned to be merged post 3.7.4, this will migrate the Windows platform to SDL and provide the framework that'll allow FSO to recognise multiple controllers.  More work will probably be required after that point to actually get them integrated/working.

mikeloeven:
I didn't see any other issues when i searched is there another bug tracker associated with the project

niffiwan:
yeah; we have a [mantis instance] (http://scp.indiegames.us/mantis/my_view_page.php). Not sure if that particular issue has been reported in there or not.

MageKing17:
I'm unaware of a corresponding Mantis issue (and I just did some fruitless searching for one); it's more of a well-known engine limitation (dating back to retail) than an issue with FSO. That being said, it *is* a known issue, and it's expected to be taken care of some time post-3.7.4.

mikeloeven:
Ahh i see though back when i actually played FS2 for the first time my joystick was serial and hotas devices were several thousand a pop XD

JohnAFernandez:
Having to push this to 21.4 as this is a feature.

z64555:
Deferring to 21.6, due to Controls5 deferment.

z64555:
Controls5 is in pending smoothing out the rest of corners.  Expect in 22.0

PR #2958, opened 2020-12-05
Controls 4 and 5 - Multi-joy and Preset Files

z64555:
Depends on #2923 and #2214, Optionally #2988 
Addresses #623 
This PR combines branches `controls4` and `controls5`, due to `controls4` breaking binding saves in CSG

<hr>

PR #2944 Controls4 - `controls4`
The long awaited Multi joy feature. Multiple joysticks may be claimed by FSO from within the fs2_open.ini using the new keys:

```
Joy0ID=
Joy0GUID=
Joy1ID=
Joy1GUID=
Joy2ID=
Joy2GUID=
Joy3ID=
Joy3GUID=
```

~~Should "CurrentJoystick" be found, it'll assume only one joystick, so you'll have to manually edit fs2_open.ini to get this feature to work until the launchers are updated.~~
The new Joy keys will be used, if found, using CurrentJoystick if Joy0 keys are not found.

Joysticks show up in game as "Joy-0" with the respective bindings.

This provides significant changes over #2923 as well 
* Axis actions merged into the Control_config vector
* ADE function for inverting by action, instead of by axis
* Separates Mouse entirely from Joystick bindings. Shows up as "Mouse Button 1" etc.
* Removes saving bindings to .CSG, retaining only .PLR bindings.

TODO: strings.tbl additions for axis labels, mouse and joystick button labels, controller labels

- [x] Bind mode should respond to all joysticks
- [x] Search mode should respond to all joysticks
- [x] Browse mode should behave as it should
- [x] Axis actions should have primary and secondary bindings available
- [x] Axes bound to the same axis action should both be able to control said axis action
- [x] Lua scripting should be able to invert both axes bound to it, effectively inverting the action
- [x] Mouse should be able to be bound and used independently of Joy0

<hr>

PR #2958 Controls5 - `controls5`
This PR removes bindings entirely from the PLR and CSG files, solely relying on preset .json's to contain the bindings.

- Old PLR files should generate a preset .json should it differ from existing. Its name is the <player name>.json and is kept in the /preset/subdir
- Old CSG files should generate a preset .json should its bindings differ from existing.  Its name is <player name>.<campaign name>.json and is kept in the /presets/ subdir
- Fixes a hidden bug in the os_registry API which would cause an access violation should the os_registry not be initialized before attempting to read data.  Instead, the default value is returned.

- [x] Conversion of PLR2 to PLR3 + preset bindings
- [x] Conversion of CSG5 to CSG6 + preset bindings
- [x] PLR should select its generated preset upon successful conversion.
- [x] CSG should select its generated preset upon successful conversion.
- [x] Switching between campaigns should retain the preset assigned to them
- [x] Switching between campaign and multiplayer should retain the preset assigned to them.
- [x] No unintended changes to other functionality
- [x] use `-get_flags json_v1` on the commandline to verify joystick information (GUID, etc.) is correct.
-- See #3328

<hr>
Misc. Additions:

* Adds the os::dialogs::Information() dialog to display information using SDL.  Does not currently have localization support.  Use sparingly in instances where you need to inform the player of something important, but does not require breaking into the debugger nor halting unit tests.

* Adds new XSTR indices for joystick and mouse axis/button labels.

default str, XSTR id
"X Axis", 1647
"Y Axis", 1648
"Z Axis", 1649
"rX Axis", 1650
"rY Axis", 1651
"rZ Axis", 1652

"X1 Button", 1653
"X2 Button", 1654
"Wheel Up", 1655
"Wheel Down", 1656
"Wheel Left", 1657
"Wheel Right", 1658

* Adds new XSTR indices for the configuration labels
"Preset: %s", 1659
"Preset: custom", 1660
"Next Preset", 1661


MageKing17:
>     * Old CSG files should generate a preset .json should its bindings differ from existing.  Its name is exactly the campaign name and is kept in the /presets/ subdir

This doesn't seem to be accurate (and shouldn't be); instead, it uses the filename of the .csg minus the extension (which it should, because otherwise you'd run into name collisions between different pilots that played the same campaign).

z64555:
> 
> 
> > ```
> > * Old CSG files should generate a preset .json should its bindings differ from existing.  Its name is exactly the campaign name and is kept in the /presets/ subdir
> > ```
> 
> This doesn't seem to be accurate (and shouldn't be); instead, it uses the filename of the .csg minus the extension (which it should, because otherwise you'd run into name collisions between different pilots that played the same campaign).

Eh, the comment is misleading I take it.  It's supposed to take the player name, plus the campaign name, minus the .csg.

MageKing17:
> Eh, the comment is misleading I take it. It's supposed to take the player name, plus the campaign name, minus the .csg.

Yes it should, which means "Its name is exactly the campaign name" (from the first post) is wrong.

z64555:
The format of the JSON object returned by -joy_info is as follows:
{
    "joysticks": [
        {
            "name": "3Dconnexion KMJ Emulator",
            "guid": "03000000EFBE00006D04000000000000",
            "id": 0,
            "device": 0
        }
    ]
}

code/controlconfig/controlsconfig.cpp
```diff
@@ -2341,56 +2524,137 @@ int check_control(int id, int key)
...
+	// Read raw sticks.
+	for (short j = CID_JOY0; j < CID_JOY_MAX; ++j) {
+		joystick_read_raw_axis(j, JOY_NUM_AXES, axe[j]);
 	}
-	if (Invert_axis[4]) {
-		*tr = -(*tr);
+
+	// Read raw mouse, stuff in axes_values[0]
+	if (Use_mouse_to_fly) {
+		mouse_get_delta(&axe[MOUSE_ID][JOY_X_AXIS], &axe[MOUSE_ID][JOY_Y_AXIS], &axe[MOUSE_ID][JOY_Z_AXIS]);
```
MageKing17, line 2640: It's mouse code, so it should be using the mouse axes (even if they are the same values).
```suggestion
		mouse_get_delta(&axe[MOUSE_ID][MOUSE_X_AXIS], &axe[MOUSE_ID][MOUSE_Y_AXIS], &axe[MOUSE_ID][MOUSE_Z_AXIS]);
```

code/controlconfig/controlsconfig.cpp
```diff
@@ -409,136 +523,163 @@ int joy_get_scaled_reading(int raw)
...
+	}
 
-	if ( (axis == -1) && Use_mouse_to_fly ) {
+	if ( (j == CID_NONE) && Use_mouse_to_fly ) {
+		// Nothing found amongst the joysticks. Check the mouse.
 		mouse_get_delta( &dx, &dy, &dz );
 
 		if ( (dx > fudge) || (dx < -fudge) ) {
+			j = CID_MOUSE;
 			axis = 0;
```
MageKing17, line 576:
```suggestion
			axis = MOUSE_X_AXIS;
```

code/controlconfig/controlsconfig.cpp
```diff
@@ -409,136 +523,163 @@ int joy_get_scaled_reading(int raw)
...
 		} else if ( (dy > fudge) || (dy < -fudge) ) {
+			j = CID_MOUSE;
 			axis = 1;
```
MageKing17, line 579:
```suggestion
			axis = MOUSE_Y_AXIS;
```

code/controlconfig/controlsconfig.cpp
```diff
@@ -409,136 +523,163 @@ int joy_get_scaled_reading(int raw)
...
 		} else if ( (dz > fudge) || (dz < -fudge) ) {
+			j = CID_MOUSE;
 			axis = 2;
```
MageKing17, line 582:
```suggestion
			axis = MOUSE_Z_AXIS;
```

MageKing17 (dismissed):
Just something I noticed, although I'll come back and do a full review later.

code/controlconfig/controlsconfig.cpp
z64555: At one point I made the mouse axes enums, but then later decided it may be better to have a generic set that was just "X_AXIS, Y_AXIS" etc.  I think I left it as the existing JOY_X_AXIS due to time constraints

z64555: Ah, found the aliases. looks like I just didn't put them to use.

z64555:
There's an issue with clang-tidy and default initialization. The [modernize-use-default-member-init] option flags a warning on controlsconfig.cpp:195 and :196 telling us to use the modernized initialization of `int value(default_value)` but MSVS 2017 flags this as an error.

z64555:
There's an issue with clang-tidy thinking PresetFileHandler::ensureNotExists() can be a static member.  This is because the Assertions used in it evaluate to an Assume() within Release build configuration, but of course it cannot be a static member due to its Debug build evaluation.

z64555:
The [https://github.com/scp-fs2open/fs2open.github.com/pull/2958/commits/3e1e0997e6a9a1774012da4a9e6be561a94c2d63](-joy_info commit) may partially resolve issue #2258.  Complete resolution requires a change to the launchers (namely knossos) to make use of the new -joy_info command line option.

ngld:
> Use `-joy_info` to query FSO and SDL to get joystick information of all connected joysticks as a JSON object.

Sorry for only mentioning it this late but that seems redundant since the JSON flag info for launchers already includes the joystick list:
https://github.com/scp-fs2open/fs2open.github.com/blob/f67860a502d8bd00bfa79f0b65088c8dea4e95c0/code/cmdline/cmdline.cpp#L1411-L1430

z64555:
> 
> 
> > Use `-joy_info` to query FSO and SDL to get joystick information of all connected joysticks as a JSON object.
> 
> Sorry for only mentioning it this late but that seems redundant since the JSON flag info for launchers already includes the joystick list:

Ok. so... that begs the question "Why is that giving the wrong GUID when `-joy_info` gives the correct GUID? Is it possible that something's not being initialized all the way?

I'll look through the codepath to see how much it differs from `-joy_info` which I have proven works on my machine.

ngld:
> Ok. so... that begs the question "Why is that giving the wrong GUID when `-joy_info` gives the correct GUID? Is it possible that something's not being initialized all the way?

I just ran Knossos against the most recent nightly (after patching it to work around the issues caused by #2988; this is going to make Knossos dev annoying for all affected builds; can) and the result was that `-get_flags json_v1` reports the same joystick UUID that `-joy_info` from your `controls5` branch reports which also matches the value Knossos writes to `CurrentJoystickGUID` in `fs2_open.ini`.
As far as I can tell, there is no issue to fix here.

WindowsTerminal_lHns2BAGEa

z64555:
> > Ok. so... that begs the question "Why is that giving the wrong GUID when `-joy_info` gives the correct GUID? Is it possible that something's not being initialized all the way?
> 
> I just ran Knossos against the most recent nightly (after patching it to work around the issues caused by #2988; this is going to make Knossos dev annoying for all affected builds) and the result was that `-get_flags json_v1` reports the same joystick UUID that `-joy_info` from your `controls5` branch reports which also matches the value Knossos writes to `CurrentJoystickGUID` in `fs2_open.ini`.
> As far as I can tell, there is no issue to fix here.

As discussed on Discord, the GUID reported by `-get_flags` is inconsistent whereas `-joy_info` appears to be more reliable.  It's thought to be a race condition somewhere with SDL since `-joy_info`'s method of obtaining the GUID is a full init of the joystick subsystem, whereas `-get_flags` takes a more lean and direct approach.

Also discussed on Discord, it was agreed that we'd test the effect of incorporating `-joy_info`'s method of getting the GUID into `-get_flags`, so sometime within the next few days I'll push a commit out for it.

z64555:
Ok, CI issues addressed.  I've got a commit that assimilates `-joy_info` into `-get_flags`, need to test the build and then I'll have the PR open for review.

z64555:
Pushing this back to 21.4, as there still appears to be problems from Controls3 that wasn't discovered.

JohnAFernandez (commented):
I marked what I noticed, which isn't much.

code/controlconfig/controlsconfig.cpp
```diff
@@ -1144,22 +1071,41 @@ void control_config_toggle_modifier(int bit)
...
+		// first
+		item.first.invert_toggle();
+		break;
+	case selItem::Secondary:
+		// second
+		item.second.invert_toggle();
+		break;
+	default:
+		// unhandled
+		mprintf(("Unhandled selItem in control_config_toggle_invert(): %i", static_cast<int>(Selected_item)));
```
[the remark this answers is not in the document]
z64555: Yes, it should be an UNREACHABLE.  I may have made this bit of code before I found out about the UNREACHABLE macro, or I may not have made it into a habit of using it.

z64555:
So as PhantomHoover noted, the axis labels in the control config options menu still say "Joystick/Mouse X axis."  This is due to the fact that the strings are XSTR'd and are still using their lines in strings.tbl.  I can't programmatically resolve this other than force usage of another XSTR index that doesn't exist in retail data, but that also means that localization support from retail and other mods that depend on strings.tbl for their translations will have to update their strings.tbl.

Goober5000:
A new set of XSTR indices sounds like the right approach.  You could provide defaults if those aren't found.

z64555:
> 
> 
> A new set of XSTR indices sounds like the right approach. You could provide defaults if those aren't found.

Yeah that's... probably the only thing I can do from the code side of things.  The hardcoded text is the "default" string using the XSTR method at any rate.  the strings.tbl more or less only supports German and French translations.  Most recent commit uses -1 for the indices but it should be trivial to assign them to new indices.  Will try to do that tomorrow.

EatThePath:
So here's a weird bug. If you have the mouse X and Y axis bound, and then you select the bind by clicking on the *binding name* specifically, the left hand column, then use the invert toggle, you get the axis names replaced with the left and right mouse buttons. But only in name, the mouse still seems to work properly. It also doesn't happen if you click on the mouse axis name instead of the turning axis name.

z64555:
> So here's a weird bug. If you have the mouse X and Y axis bound, and then you select the bind by clicking on the _binding name_ specifically, the left hand column, then use the invert toggle, you get the axis names replaced with the left and right mouse buttons. But only in name, the mouse still seems to work properly. It also doesn't happen if you click on the mouse axis name instead of the turning axis name.

Should be addressed with commit 1acf95a


> Renaming the 'defaults' to 'cycle presets' and making the text for the preset name bigger would be good for user experience but I assume that's not really practical.

Renamed `defaults` to `next preset` to try to minimize the string size in commit bc150e7

As for making the preset name label in the upper left of the screen bigger, I don't have any clean way of arbitrarily setting the font size in the retail menu UI, its either the size it currently is, or 2x as big (and the bigger size doesn't fit nicely).

z64555:
Ok, looks like clang tidy is hanging up on two issues:
> /__w/fs2open.github.com/fs2open.github.com/code/controlconfig/controlsconfigcommon.cpp:2327:10: warning: redundant boolean literal in conditional return statement [readability-simplify-boolean-expr]

...exactly how is that redundant?

> ../code/osapi/dialogs.h:140:20: warning: using decl 'Information' is unused [misc-unused-using-decls]

but it is used at `pilotfile/plr.cpp:616`

ngld:
> ...exactly how is that redundant?

Instead of
```cpp
if (x) {
  return false;
}
return true;
```
it wants you to write the following (simpler but equivalent) solution:
```cpp
return !x;
```

> but it is used at `pilotfile/plr.cpp:616`

The CI script runs `clang-tidy` once for each changed file which means that it only checked the header by itself and didn't consider any other file that might include the header. I think it's safe to ignore this warning unless you want to copy that `using` clause into each file that includes `dialogs.h`.

z64555:
> > ...exactly how is that redundant?
> 
> Instead of
> 
> ```c++
> if (x) {
>   return false;
> }
> return true;
> ```
> 
> it wants you to write the following (simpler but equivalent) solution:
> 
> ```c++
> return !x;
> ```

Ok, so its complaining about the last conditional because instead of using a conditional and two return statements (as is the normal style when using if statements) clang-tidy wants me instead to use a single return statement.  While that does make it theoretically simpler by having only one exit path (at this point), I feel like clang-tidy is complaining about a single tree that's in a forest.  If anything at all, I believe using the single return statement would obfuscate what's going on here.


> > but it is used at `pilotfile/plr.cpp:616`
> 
> The CI script runs `clang-tidy` once for each changed file which means that it only checked the header by itself and didn't consider any other file that might include the header. I think it's safe to ignore this warning unless you want to copy that `using` clause into each file that includes `dialogs.h`.


Well, that makes sense, I guess.  I'm not too certain what to do, since I was going by what was done for the other dialogs to try to keep things consistent with the API.

Goober5000:
Here's a script to test the X- and Y-axis inversion.  Tested on current master.  Load any mission and press X for the X-axis and Y for the Y-axis.  The settings will revert once you exit the mission.
[invert_axis-sct.tbm.txt](https://github.com/scp-fs2open/fs2open.github.com/files/6928499/invert_axis-sct.tbm.txt)

z64555:
Pushing back for next RC.

ngld:
Can you add a new entry (i.e. `MultiJoy`) to this list?
https://github.com/scp-fs2open/fs2open.github.com/blob/9ed7d971f2c8a348f273cce31e5912df53a350ea/code/cmdline/cmdline.cpp#L1316-L1323

That way, I can tell which builds support multiple joysticks and display this in Knossos to help players figure out why only some mods work with multiple joysticks.

z64555:
Latest CI failure is due to static_assert from #3739 as a reminder to revert forward compatibility feature for SEXP containers.

JohnAFernandez (approved):
Approving because we need to send it to the community for an open beta.

PR #2959, opened 2020-12-05
some fixes to aiturret code

Goober5000:
I did a careful review of all the functions in aiturret.cpp.  This cleans up the code a bit to more closely follow the original code flow from the public code release (for ease of comparison) and fixes several bugs.

1) fixed other weapons on a turret being handicapped by spawning weapon fire rate

2) fixed beam-free flag on a turret with multiple weapons

3) fixed turret accuracy penalty never being applied

4) fixed weapon timestamp being overridden

5) fixed player sound not being played, if specified, with swarm turret weapons

6) fixed flak not jittering if it had the fire-down-normals flag

This finally resolves Mantis 804.
http://scp.indiegames.us/mantis/view.php?id=804

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -2587,16 +2587,13 @@ void ai_fire_from_turret(ship *shipp, ship_subsys *ss, int parent_objnum)
 			ship_get_global_turret_gun_info(&Objects[parent_objnum], ss, &gpos, &gvec, use_angles, &predicted_enemy_pos);
 
 			// Fire in the direction the turret is facing, not right at the target regardless of turret dir.
+			// [Yet this retail comment precedes the calculation of vector-to-enemy...]
 			vm_vec_sub(&v2e, &predicted_enemy_pos, &gpos);
 			dist_to_enemy = vm_vec_normalize(&v2e);
 			float dot = vm_vec_dot(&v2e, &gvec);
 
-			// if the weapon is a flak gun, add some jitter to its aim so it fires in a "cone" 
-			// to make a cool visual effect and make them less lethal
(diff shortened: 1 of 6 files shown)


code/ai/aiturret.cpp
```diff
@@ -2380,105 +2361,119 @@ void ai_fire_from_turret(ship *shipp, ship_subsys *ss, int parent_objnum)
...
+		// Similar check for small weapons
+		if ( wip->wi_flags[Weapon::Info_Flags::Small_only] ) {
+			if ( lep->type != OBJ_SHIP) {
+				tentative_return = true;
+				continue;
+			}
+			if ( Ship_info[Ships[lep->instance].ship_info_index].is_big_or_huge() ) {
+				tentative_return = true;
+				continue;
+			}
```
Baezon, line 2401: This changes the logic of `small_only` targeting. In the original, `small_only` weapons _can_ be fired at non-ships, basically the inverse of huge (must be a ship AND big|huge).

Baezon (changes requested):
Looks pretty good aside from one small issue. I didn't quite realize how inconsistent `Use_multiple_guns` would have been before. And I forgot about the flak jittering aim with the `fire_down_normals` change, sorry!

code/ai/aiturret.cpp
Goober5000: Ah, good catch.  New commit pushed.

Goober5000: And no worries about missing the `fire_down_normals` check.  We all miss something occasionally!

Baezon (approved):

PR #2962, opened 2020-12-05
Add flag for AI to only fire Turrets / Secondaries when having clear Line of Sight

SelfGamer:
This added the Weapon Flag "require exact los", which will prevent AI form firing it from a turret or as a secondary weapon if there is no clear line of sight from the AI's ship (center) to it's targeted ship (center)
In addition, this added the AI Profile Flag "$firing requires exact los:", which has the same effect, but applied to all turrets and AI controlled secondaries regardless of weapon or weapon flags

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -6124,7 +6124,7 @@ int check_ok_to_fire(int objnum, int target_objnum, weapon_info *wip)
 
 		if (The_mission.ai_profile->flags[AI::Profile_Flags::Require_exact_los] || wip->wi_flags[Weapon::Info_Flags::Require_exact_los]) {
 			//Check if th AI has Line of Sight and is allowed to fire
-			if (!check_los(objnum, target_objnum, 1.0f)) {
+			if (!check_los(objnum, target_objnum, 10.0f)) {
 				return 0;
 			}
 		}
@@ -6204,11 +6204,12 @@ bool check_los(int objnum, int target_objnum, float threshold) {
 
 		//Early Out Model too small
 
-		float radius = model_get(model_num)->rad;
+		float radius = objp->radius;
 		if (radius < threshold)
 			continue;
 
-		vec3d a, b, c;
+		//Alternate "is in cylinder relevant for LoS" check
+		/*vec3d a, b, c;
 		vm_vec_sub(&a, &start, &objp->pos);
 		vm_vec_sub(&b, &end, &start);
 		vm_vec_cross(&c, &b, &a);
@@ -6223,6 +6224,29 @@ bool check_los(int objnum, int target_objnum, float threshold) {
 		//Early out Model too far from LoS
 		if (distLoSSquared > radius || -maxTdelta > t || maxTdelta + 1 < t)
 			continue;
+		*/
+
+		//Asteroth's implementation
+		// if objp is inside start or end then we have to check the model
+		if (vm_vec_dist(&start, &objp->pos) > objp->radius && vm_vec_dist(&end, &objp->pos) > objp->radius) {
+			// now check that objp is in between start and end
+			vec3d start2end = end - start;
+			vec3d end2start = start - end;
+			vec3d start2objp = objp->pos - start;
+			vec3d end2objp = objp->pos - end;
+
+			// if objp and end are in opposite directions from start, then early out
+			if (vm_vec_dot(&start2end, &start2objp) < 0.0f)
+				continue;
+
+			// if objp and start are in opposite directions from end, then early out
+			if (vm_vec_dot(&end2start, &end2objp) < 0.0f)
+				continue;
+
+			// finally check if objp is too close to the path
+			if (vm_vec_mag(&start2objp) > (vm_vec_mag(&start2end) + objp->radius))
+				continue; // adding objp->radius is somewhat of an overestimate but thats ok
+		}
 
 		mc_info hull_check;
 		mc_info_init(&hull_check);
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -6124,7 +6124,7 @@ int check_ok_to_fire(int objnum, int target_objnum, weapon_info *wip)
 
 		if (The_mission.ai_profile->flags[AI::Profile_Flags::Require_exact_los] || wip->wi_flags[Weapon::Info_Flags::Require_exact_los]) {
 			//Check if th AI has Line of Sight and is allowed to fire
-			if (!check_los(objnum, target_objnum)) {
+			if (!check_los(objnum, target_objnum, 1.0f)) {
 				return 0;
 			}
 		}
@@ -6170,8 +6170,8 @@ int check_ok_to_fire(int objnum, int target_objnum, weapon_info *wip)
 
 //	--------------------------------------------------------------------------
 //  Returns true if *aip has a line of sight to its current target.
-//	This is a computationally expensive operation, so use sparingly
-bool check_los(int objnum, int target_objnum) {
+//	threshold defines the minimum radius for an object to be considered relevant for LoS
+bool check_los(int objnum, int target_objnum, float threshold) {
 	vec3d& end = Objects[target_objnum].pos;
 	vec3d& start = Objects[objnum].pos;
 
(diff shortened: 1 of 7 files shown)


code/ai/aicode.cpp
```diff
@@ -6159,6 +6168,57 @@ int check_ok_to_fire(int objnum, int target_objnum, weapon_info *wip)
...
 }
 
+//	--------------------------------------------------------------------------
+//  Returns true if *aip has a line of sight to its current target.
+//	This is a computationally expensive operation, so use sparingly
+int check_los(int objnum, int target_objnum) {
+	vec3d& end = Objects[target_objnum].pos;
+	vec3d& start = Objects[objnum].pos;
+
+	for (int i = 0; i < Highest_object_index; i++) {
```
Goober5000, line 6178: Rather than iterate through all object numbers, it would be better to iterate on `obj_used_list`.

code/ai/aicode.cpp
```diff
@@ -6159,6 +6168,57 @@ int check_ok_to_fire(int objnum, int target_objnum, weapon_info *wip)
 	return 1;
 }
 
+//	--------------------------------------------------------------------------
+//  Returns true if *aip has a line of sight to its current target.
+//	This is a computationally expensive operation, so use sparingly
+int check_los(int objnum, int target_objnum) {
```
Goober5000, line 6174: Since we are now using C++ conventions, please change the return type of this function to `bool`.

Goober5000 (changes requested):
Good feature idea.  Just a couple comments from me.

code/ai/aicode.cpp
JohnAFernandez: To add on, you can mostly ignore the style and conventions of the existing code.  It's not worth any of our coder's time to switch every NULL to nullptr, for instance, but every time we add code, we use nullptr.  Any thing that you can think of to make your code more modern, do it.

SelfGamer:
Alright, modified the code.
Thanks for the feedback!

code/ai/aicode.cpp
```diff
@@ -6159,6 +6168,57 @@ int check_ok_to_fire(int objnum, int target_objnum, weapon_info *wip)
...
+		hull_check.p0 = &start;
+		hull_check.p1 = &end;
+		hull_check.flags = MC_CHECK_MODEL;
+
+		if (model_collide(&hull_check)) {
+			return 0;
+		}
+	}
+
+	return 1;
```
Baezon: Firstly, since we have the benefit of C++, this can return Boolean instead of an int.

As you say, a model checking collision isn't exactly cheap so we want to avoid it where possible using simpler data. For instance if neither us nor our target is inside our possibly-line-of-sight-blocking object and it is farther away than our target plus its radius we know it couldn't possibly occlude us. It could also be in the wrong direction as our target, there's a few of these so I might directly code suggest them a little bit later.

Furthermore its important to consider what the intended use of this function really is. You probably don't want to hold your fire if a fighter or fast moving peice of debris happens to be in the way, its likely that our shot won't even hit them by time it gets to them because they'll be long gone by then, so I think it might be preferable to check for small or fast-moving objects and just immediately disregard them saving a lot of computation. Or you could have an optional parameter for it if you really want strict LoS.

code/ai/aicode.cpp
Baezon: Iterating through objects is generally done through ´obj_used_list´ (searching for that will 1quickly bring up how its used in this way). I don't know if its better that way, but its the more commonly used method.

Baezon (changes requested):

code/ai/aicode.cpp
SelfGamer: > distance less than its radius
I did think about that, but as model_collide() has a pretty quick exit with a Bounding Sphere check, I assumed that would be redundant here
> if a fighter or fast moving peice of debris
it would be an option, I will be adding minRadius as a parameter
> maybe_avoid_big_ship´ could then use this function instead
I can see what I can do

code/ai/aicode.cpp
SelfGamer: Goober also suggested this, so okay. I changed it

Goober5000 (approved):
Changes look good to me.  I'll leave it to @Baezon on how best to handle the nuances he mentioned.

code/ai/aicode.cpp
Baezon: You should check out #2735 . Once I started profiling these things the results were quite surprising. Doing this checking here instead of just letting ´segment_sphere´ handle it in ´model_collide´ is a pretty miniscule time saving but this early outing will be happening _a lot_, it will be the overwhelmingly common situation and suddenly that time saving becomes not so miniscule, because simply setting up the model collision takes a significant portion of that time.

That said, I think only 3 simple checks are really necessary here, whether you or the target are inside the objects radius (no avoiding the model collision then) checking that the object is between you and the target, and that its close enough to the path to occlude it. And I can help you with those in a few hours if you're not sure how to implement them.

SelfGamer: Okay, added some earlier outs for small radius (for now I assume things with a radius smaller than 1 are irrelevant for los when firing weapons), as well as not relevant for los due to distance from the relevant los as well as being behind target or ship.
Regarding usage in maybe_avoid_big_ship, I'd like to make that a separate PR, as that needs in itself a lot more testing as you'd need to make sure you're not breaking anything that relies on this AI behaviour

SelfGamer: And while I didn't expect there to be that much difference, your profiling results speak for themselves.

code/ai/aicode.cpp
```diff
@@ -6159,6 +6168,81 @@ int check_ok_to_fire(int objnum, int target_objnum, weapon_info *wip)
...
+		float distToTargetRecip = 1.0f / vm_vec_dist_squared(&start, &end);
+		float distLoSSquared = vm_vec_mag_squared(&c) * distToTargetRecip;
+		float t = -vm_vec_dot(&a, &b) * distToTargetRecip;
+
+		radius *= radius;
+		float maxTdelta = sqrtf(radius * distToTargetRecip);
+		
+		//Early out Model too far from LoS
+		if (distLoSSquared > radius || -maxTdelta > t || maxTdelta + 1 < t)
+			continue;
```
Baezon, line 6225: I'll be honest, I'm not really able to follow what exactly you're doing here, this is more or less what I had in mind:

```suggestion
    // if objp is inside start or end then we have to check the model
		if (vm_vec_dist(&start, &objp->pos) > objp->radius && vm_vec_dist(&end, &objp->pos) > objp->radius){
		    // now check that objp is in between start and end
		    vec3d start2end = end - start;
		    vec3d end2start = start - end;
		    vec3d start2objp = objp->pos - start;
		    vec3d end2objp = objp->pos - end;
		    
		    // if objp and end are in opposite directions from start, then early out
		    if(vm_vec_dot(&start2end, &start2objp) < 0.0f)
		        continue;
		         
		    // if objp and start are in opposite directions from end, then early out
		    if(vm_vec_dot(&end2start, &end2objp) < 0.0f)
		        continue; 
		        
		    // finally check if objp is too close to the path
		    if (vm_vec_mag(&start2objp) >  (vm_vec_mag(&start2end) + objp->radius))
		        continue; // adding objp->radius is somewhat of an overestimate but thats ok
		}
```

I hope I'm not being too 'micro-managey'

Baezon (changes requested):
Aside from the below, 1 meter is a very small threshold, you have your use case so if that's really what you want then ok, and the API now exists so its not a big deal to use different numbers for different contexts, but such a threshold won't really cut out much of anything. Only the very smallest of fighter debris will have a radius of 1 meter, you may as well be checking everything. If that's what you want, well then no problem, really.

code/ai/aicode.cpp
```diff
@@ -6159,6 +6168,81 @@ int check_ok_to_fire(int objnum, int target_objnum, weapon_info *wip)
...
+		else if (type == OBJ_SHIP) {
+			model_num = Ship_info[Ships[objp->instance].ship_info_index].model_num;
+			model_instance_num = Ships[objp->instance].model_instance_num;
+		}
+		else
+			continue;
+
+		//Early Out Model too small
+
+		float radius = model_get(model_num)->rad;
```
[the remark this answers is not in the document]
SelfGamer: ohh right. Good point.
And true, that is way too small, after checking sizes in fred again. Will be increasing that to 10

code/ai/aicode.cpp
SelfGamer: This was essentially an implementation of https://mathworld.wolfram.com/Point-LineDistance3-Dimensional.html
And it then checks the distance from the Line of Sight to be checked, and if it is between both points (or within it's radius from the edges), creating a probably quite fast "is point in cylinder"-check. 
But if you want, I can use your code as well, resulting in a more readable source.
As for which is faster / better, I don't really know.

Baezon: Ah ok, I think a simpler, more easily-unstandable-at-a-glance approach would be better. We don't need an exact solution, just something that can quickly weed out 95% of cases, speed for both is probably pretty similar

Goober5000: The functions `vm_vec_dist_to_line` and `vm_vec_dist_squared_to_line` can be found in vecmat.cpp.  It is better to use existing functions unless there is a clear advantage in writing a new one.

SelfGamer: I originally didn't use the functions as I could reuse lots of the intermediates for the second check, but as I've replaced the code anyway with Asteroth's suggestion, I guess it's fine anyways.
But yeah, maybe I should've used the vecmat implementation

Baezon (approved):
Alright, looks good!

PR #2965, opened 2020-12-07
Fix an assert with 0 initial beam width

Baezon:
Actually 0 width causes problems so I've capped it at a minimum 0.1 just like shrinking does. I think that's kind of high, and I'm also not a huge fan of doing this silently but this is exactly how shrink handles it so... dunno.


Goober5000 (approved):
Ok I guess.

Issue #2960, opened 2020-12-05
Fade-ins misbehaving after mission restart?


MitoPL:
If you restart a cutscene mission that uses black fade-ins, they will become white fade-ins instead. This behaviour also seems to persist over subsequent runs.

My way of reproducing: launch Blue Planet Complete, launch the Icarus cutscene (via "continue campaign" option), when the mission ends, use "Restart mission" button. It seems to happen pretty consistently. Tested on 2020-10-01 nightly build, but I recall seeing this problem possibly as far back as a year ago, also in multiple cutscenes and missions.

z64555:
Confirmed, looking into it

z64555:
Ultimately the issue arose because of weird behavior using legacy functionality of the ```fade-in``` and ```fade-out``` sexps,   in addition to the Fade_out color not being initialized on mission start.

The ```fade-in``` sexp had a feature which would use the last color as set by ```fade-out``` However, ```fade-out``` did not save its color if it was black.  Thus, should any ```fade-out``` effect use any color other than the default black, the ```fade-in``` color reuse feature would malfunction.

The fade colors were only be initialized once on program start, so it would be possible for a color ```fade-out``` at the end of one mission and then be used for a ```fade-in``` effect on the next mission.  However this behavior was not explicitly documented anywhere, and in my opinion is bad design anyway.

Goober5000:
Fixed by #2967

PR #2967, opened 2020-12-07
Save fade-out color, even if its black.

z64555:
Fixes issue #2960 

The existing logic for black fade-ins is overridden to use whatever color the previous fade-out did, and at some point during the end of the cutscene, the fade_out colors were being bashed to white.

edit:  As insurance, init the fade_out color to black at mission start with init_sexp

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -891,6 +891,11 @@ ship_subsys *Players_targeted_subsys;
 int Players_target_timestamp;
 int Players_mlocked_timestamp;
 
+// for sexp_fade
+static int Fade_out_r = 0;
+static int Fade_out_g = 0;
+static int Fade_out_b = 0;
+
 // for play-music - Goober5000
 SCP_vector<int>	Sexp_music_handles;		// All handles used by all invocations of play-sound-from-file.  The default handle is in index 0.
 
@@ -1198,6 +1203,11 @@ void init_sexp()
 	Assert(Locked_sexp_true != -1);
 	Sexp_nodes[Locked_sexp_true].type = SEXP_ATOM;  // fix bypassing value
 	Sexp_nodes[Locked_sexp_true].value = SEXP_KNOWN_TRUE;
+
+	// init fade_out colors
+	Fade_out_r = 0;
+	Fade_out_g = 0;
+	Fade_out_b = 0;
 }
 
 void sexp_shutdown() {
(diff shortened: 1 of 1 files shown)


Goober5000 (changes requested):
Keep that Mantis 2944 comment; it's important to know that information.

Instead of removing the RGB comparisons entirely, try this.  Store the array count from the `eval_array` call several lines up, and do this:
```
// Mantis #2944, if we're fading out to some specific color, save that color
if (!fade_in && rgb_count > 0)
```

z64555:
ehhhh. honestly I think this section could be refactored a bit

Goober5000:
> ehhhh. honestly I think this section could be refactored a bit

I don't think so; I think the sexp is written just about as well as it can be.

However, the `Fade_out_[rgb]` variables aren't reset upon mission load.  Try resetting them to -1 in `init_sexp`.

z64555:
Ok, so lets break down whats happening in sexp_fade

ints R, G, and B are read in from the sexp, given a value of -1 should the passed values are invalid or nonexistant.

Should any of the R, G, or B values be negative, (or implicitly outside of the range of 0~255), legacy fade color block kicks in.  This block will select it to be one of four colors: White if R =1, Red if R = 2, and Black for any other value of R.

If a fadeout had occurred beforehand, instead of using Black it will use the fadout color.


> Instead of removing the RGB comparisons entirely, try this. Store the array count from the `eval_array` call several lines up, and do this:
> 
> ```
> // Mantis #2944, if we're fading out to some specific color, save that color
> if (!fade_in && rgb_count > 0)
> ```

So, having the rgb_count check for the second if block is moot, because by this time R, G, B are all set.

The refactor would get rid of this line entirely:
```
if (fade_in && (Fade_out_r >= 0) && (Fade_out_g >= 0) && (Fade_out_b >= 0))
```
and instead that else statement will just set the RBG to the Fade_out RGB, and finally have the Fade_out RGB initialized to black in the first place.

Baezon:
"Maybe" fixes? That doesn't exactly inspire confidence, were you able to replicate the issue to see if it was fixed?

z64555:
> 
> 
> "Maybe" fixes? That doesn't exactly inspire confidence, were you able to replicate the issue to see if it was fixed?

The "maybe" comment was from me not sure if I was running all of the same settings done with release 20.0 when I was running it with the test build, as I'm not familiar with how to do testing with mod data through the debugger.

Goober5000:
> So, having the rgb_count check for the second if block is moot, because by this time R, G, B are all set.

No, that's why I said to store the count from `eval_array`.  Like so:
```
int rgb_count = eval_array<int>(temp_rgb, n, is_nan, is_nan_forever, [](int num) {
```
This way `rgb_count` will be > 0 if any color arguments were supplied.

MageKing17 (approved):

Goober5000 (approved):
Yup, I think this makes sense.

z64555:
Confirmed as fixed on my side, using WiH part 2 campaign opening cutscene

PR #2969, opened 2020-12-08
Fix fred control crash

### LINKED ISSUE #2283: Controls combo allowing you to accidentaly exceed INT_MAX coordinates in FRED2_Open
I've recently experimented a bit with FRED controls and it turns out there's a combo of settings that can easily lead you to moving your camera beyond acceptable coordinates. In order to do it, you need to:
- Use the 'Select and Move' mode,
- Use the 'Rotate around object' mode,
- Grab an object (the camera will snap on to the selected one so doesn't matter what you grab) and then rapidly move your mouse around while holding the object.

The error is 100% reproduceable, ending with a crash - but it actually happened to me to also experience a BSOD while trying to cause this.
I think the problem here is solveable by locking the location of the selected object when in 'Rotate around object' mode, but I think there are a couple other solutions possible. Discuss?

JohnAFernandez:
Yes, it really was that simple to fix #2283 .

Currently attempting to install qt in order to test the fix on qtFRED.

### DIFF
--- a/fred2/fredview.cpp
+++ b/fred2/fredview.cpp
@@ -604,7 +604,8 @@ int drag_objects()
 	}
 	*/
 
-	if (!query_valid_object())
+	// Do not move ships that we are currently centered around (Lookat_mode). The vector math will start going haywire and return NAN
+	if (!query_valid_object() || Lookat_mode)
 		return -1;
 
 	if ((Dup_drag == 1) && (Briefing_dialog))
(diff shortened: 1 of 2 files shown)


Goober5000 (approved):
[Petrarch] Excellent work. [/Petrarch]

I don't think it's necessary to build qtFRED to verify the fix, but if you want to do so, it certainly can't hurt.

JohnAFernandez:
I'm going to need to, anyway, if I want to do any work in qtFRED.  But we can always fix it after the fact if something is somehow wrong, especially since qtFRED is not actively used yet.

Issue #1821, opened 2018-08-01
Disabling hud by shift + o prevents from gaining target lock for aspect secondaries or locks permanently until target explodes.


tomimaki:
Retail data.
There is some sort cheat. If you acquire lock, disable hud, then turn back from target, launch e.g. cyclops, swap fast to next sbank and shot without targeting, and notice how secondaries still follow target. ;)
Of course this works until you enable hud again.

z64555:
Need repro of this for confirmation

FSCyborg:
I was able to reproduce on a November nightly in Slaying Ravana.  I'm going to look at the relevant code to see if I'm up to the challenge.

FSCyborg:
So anyone is able to jump in here, but my idea on a proper fix would be to perform missile locking as normal when the hud is disabled.  But if this ends up being too much, perhaps just disabling all missile locks until the hud is re-enabled.

Goober5000:
Sound reasonable.

z64555:
Confirmed.  While there's no solid lock for the missiles (that is, they won't hit the target) they will still seek to the target instead of the expected behavior of flying straight like a dumbfire.

This doesn't exactly have a huge impact on gameplay, as you can't hit anything.  Although I haven't confirmed the effect  super-agile missiles.

tomimaki:
Fixed.

PR #2970, opened 2020-12-08
Make sure lock indicators are done even if hud is disabled

JohnAFernandez:
Someone apparently already did the work of separating the render code from the lock tracking code, now we just make sure that this function always runs even if the hud is disabled. (it just makes sense for it to)  

I tried this out and it works as expected with hud disabled: turning away from your target breaks lock, turning towards your target allows you to acquire lock, and lock is maintained if it was already present and you are facing the target.

For issue #1821 

EDIT: Btw, the locking sounds still run with the HUD disabled, should I add logic that disables it if the hud is disabled?

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -1613,6 +1613,8 @@ void hud_render_preprocess(float frametime)
 	}
 
 	if ( hud_disabled() ) {
+		// if the hud is disabled, we still need to make sure that the indicators are properly handled
+		hud_do_lock_indicators(flFrametime);
 		return;
 	}
 


Goober5000:
> EDIT: Btw, the locking sounds still run with the HUD disabled, should I add logic that disables it if the hud is disabled?

Hmm, good question.

I say leave the sounds on.  It provides good audio feedback to the player and he may want to know when he achieves target lock even if he can't see it.  In the case where there's an in-mission cutscene and sounds are not desired, the FREDder can simply disable sound effects.

Goober5000 (approved):

PR #2971, opened 2020-12-08
Model preliminary edits

Goober5000:
Several housekeeping changes in preparation for the submodel rotation upgrade.  These are in a separate PR so as to not clutter up the upcoming main PR.  All changes should be no-ops.


Goober5000:
I'd like the cleanup to be merged as soon as feasible.  The feature itself can wait until after the next release.

JohnAFernandez (approved):
These are all straightforward changes. LGTM

PR #2972, opened 2020-12-08
When Fullscreen, dialogs have no parent now. Should fix #2949

### LINKED ISSUE #2949: Warning/Error dialogs not clickable if FSO starts in full screen
The issue identified in #1886, and fixed in #2139, seems to be back.  Was SDL upgraded again?

SelfGamer:
By setting the parent window of dialogs to NULL instead of the FSO context, this should make them refocusable, even through the fullscreen context being blocked by the dialog itself.
Should fix #2949

### DIFF
--- a/code/osapi/dialogs.cpp
+++ b/code/osapi/dialogs.cpp
@@ -230,7 +230,7 @@ namespace os
 			boxData.flags = SDL_MESSAGEBOX_ERROR;
 			boxData.message = boxText.c_str();
 			boxData.title = "Error!";
-			boxData.window = os::getSDLMainWindow();
+			boxData.window = !(Cmdline_fullscreen_window || Cmdline_window) ? NULL : os::getSDLMainWindow();
 
 			gr_activate(0);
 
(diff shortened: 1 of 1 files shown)


z64555 (changes requested):
I think its better to just set the parent window to null, regardless of FSO's window mode.  This way its consistent behavior and guarantees the message box will be focusable.

SelfGamer:
I'll quickly test that behaviour

MageKing17:
> I think its better to just set the parent window to null, regardless of FSO's window mode.

Seconded.

code/osapi/dialogs.cpp
```diff
@@ -230,7 +230,7 @@ namespace os
 			boxData.flags = SDL_MESSAGEBOX_ERROR;
 			boxData.message = boxText.c_str();
 			boxData.title = "Error!";
-			boxData.window = os::getSDLMainWindow();
+			boxData.window = !(Cmdline_fullscreen_window || Cmdline_window) ? NULL : os::getSDLMainWindow();
```
asarium, line 233: Since you duplicated this code for every invocation, please move it to a common function

asarium (changes requested):

asarium:
Setting the parent to null would cause issues for people who do not use the default screen. Also, even if the Windows behaviour is ok with those changes, other platforms may not.

I think it is a good change to not have a parent when we are in fullscreen but the original behaviour should be kept otherwise.

SelfGamer:
I agree with @asarium here, as setting it to NULL always results in the window being focusable with the dialog in the background when in windowed mode. Which is not good.

asarium (approved):

MageKing17 (approved):

z64555 (approved):

PR #2979, opened 2020-12-13
Fix line-of-sight checking for multilock

Baezon:
It would only lock on _out_ of sight subsystems. Whoops!


EatThePath (approved):
I built and tested this branch and it does what it says it'll do.

JohnAFernandez (approved):

asarium (approved):

PR #2981, opened 2020-12-14
Add Keyboard-Layout Sensitive Control Bindings & InputBoxes

### LINKED ISSUE #2225: International Keyboard Layouts
Hi all,
fso doesn't take international keyboard layout into account, it keeps using qwerty layout even though I'm in azerty (mac).

Steps to reproduce:
1° from the system pick any other keyboard layout than the qwerty pc one
2° go to create a player, type anything, it will still be qwerty

configuration
fso version : 3.8.1 (compiled from github), 3.8.0 (from knossow), 3.8.0 debug (from knossois)
os: 10.14.6
language and keyboard set to French

Djamé

SelfGamer:
Depends on #2958 as it's based on @z64555 's new Control Code 
Should fix #2225 .
An additional control-preset to controlconfigdefaults.tbl utility will be added later as discussed in the relevant issue.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -241,6 +241,7 @@ Flag exe_params[] =
 	{ "-output_sexps",		"Output SEXPs to sexps.html",				true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-output_sexps", },
 	{ "-output_scripting",	"Output scripting to scripting.html",		true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-output_scripting", },
 	{ "-output_script_json",	"Output scripting doc to scripting.json",	true,	0,								EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-output_script_json", },
+	{ "-controlconfig_tbl",	"Save control presets to table",			true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-controlconfig_tbl", },
 	{ "-save_render_target",	"Save render targets to file",			true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-save_render_target", },
 	{ "-verify_vps",		"Spew VP CRCs to vp_crcs.txt",				true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-verify_vps", },
 	{ "-reparse_mainhall",	"Reparse mainhall.tbl when loading halls",	false,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-reparse_mainhall", },
@@ -527,6 +528,7 @@ cmdline_parm get_flags_arg(GET_FLAGS_STRING, "Output the launcher flags file", A
 cmdline_parm output_sexp_arg("-output_sexps", NULL, AT_NONE); //WMC - outputs all SEXPs to sexps.html
 cmdline_parm output_scripting_arg("-output_scripting", NULL, AT_NONE);	//WMC
 cmdline_parm output_script_json_arg("-output_script_json", nullptr, AT_NONE);	// m!m
+cmdline_parm generate_controlconfig_arg("-controlconfig_tbl", nullptr, AT_NONE);	
 
 // Deprecated flags - CommanderDJ
 cmdline_parm deprecated_spec_arg("-spec", "Deprecated", AT_NONE);
@@ -1952,6 +1954,9 @@ bool SetCmdlineParams()
 	if (output_sexp_arg.found() ) {
 		Cmdline_output_sexp_info = true;
 	}
+
+	if (generate_controlconfig_arg.found())
+		Generate_controlconfig_table = true;
 
 	if ( no_pbo_arg.found() )
 	{
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -195,6 +195,7 @@ Flag exe_params[] =
 	{ "-no_enhanced_sound",	"Disable enhanced sound",					false,	0,									EASY_DEFAULT,					"Audio",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-no_enhanced_sound", },
 
 	{ "-portable_mode",		"Store config in portable location",		false,	0,									EASY_DEFAULT,					"Launcher",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-portable_mode", },
+	{ "-joy_info",			"Outputs SDL joystick info",				true,	0,									EASY_DEFAULT,					"Launcher",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-joy_info",},
 
 	{ "-standalone",		"Run as standalone server",					false,	0,									EASY_DEFAULT,					"Multiplayer",	"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-standalone", },
 	{ "-startgame",			"Skip mainhall and start hosting",			false,	0,									EASY_DEFAULT,					"Multiplayer",	"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-startgame", },
@@ -419,6 +420,7 @@ char *Cmdline_gateway_ip = nullptr;
 
 // Launcher related options
 cmdline_parm portable_mode("-portable_mode", NULL, AT_NONE);
+cmdline_parm joy_info("-joy_info", "Outputs SDL joystick info", AT_NONE);
 
 bool Cmdline_portable_mode = false;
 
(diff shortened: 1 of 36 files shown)


code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -771,7 +773,7 @@ const char *textify_scancode(int code)
 		}
 	}
 
-	strcat_s(text, Scan_code_text[keycode]);
+	strcat_s(text, SDL_GetKeyName(SDL_GetKeyFromScancode(FS2toSDL[keycode])));
```
z64555, line 776: std::map::operator[] will return an empty value if the given key is not already in the map, in addition to adding the new key and empty value pair to the map.  You must either ensure `keycode` is already mapped or use ::at()

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -416,7 +416,8 @@ SCP_unordered_map<SCP_string, IoActionId> old_text = {
 	{"Custom Control 5",                        CUSTOM_CONTROL_5},
 };
 
-const char *Scan_code_text_german[] = {
+//Scan code texts are no longer needed as SDL handles the names
```
z64555, line 419: If these char arrays are going to be eliminated, might as well delete the whole definition instead of commenting outl.  They'll still be in the git history if we ever need to find them again.

code/io/key.cpp
```diff
@@ -237,6 +238,10 @@ void FillSDLArray ()
 	SDLtoFS2[SDL_SCANCODE_PRINTSCREEN] = KEY_PRINT_SCRN;
 	SDLtoFS2[SDL_SCANCODE_PAUSE] = KEY_PAUSE;
 	//	SDLtoFS2[SDL_SCANCODE_BREAK] = KEY_BREAK;
+
+	for (auto& lookup : SDLtoFS2) {
+		FS2toSDL[lookup.second] = lookup.first;
```
z64555, line 243: operator[]

code/io/key.cpp
```diff
@@ -277,6 +282,37 @@ int key_to_ascii(int keycode )
...
+int key_to_ascii_modern(int keycode)
+{
+	bool shifted;
+
+	if (!key_inited) return 255;
+
+	shifted = keycode & KEY_SHIFTED;
+	keycode &= 0xFF;
+
+	const char* keys = SDL_GetKeyName(SDL_GetKeyFromScancode(FS2toSDL[keycode]));
```
z64555: operator[]

z64555 (changes requested):

code/controlconfig/controlsconfigcommon.cpp
SelfGamer: considering that we need to actually handle the case that keycode not being in FS2toSDL should return SDL_SCANCODE_UNKNOWN, I've replaced this with a find and a check if this returns map::end instead of ::at which would throw an exception.
But the insertion behaviour of [] would certainly be undesirable

code/io/key.cpp
SelfGamer: This function doesn't exist anymore. It was a first implementation of text input, but was replaced fully by the event in inputbox.cpp, and was thus removed in one of the later commits. But I do agree on the []

code/io/key.cpp
```diff
@@ -54,7 +54,8 @@ SDL_mutex* key_lock;
 								// when typing in your pilots callsign.  This global flag is checked before execution
 								// is stopped.
 
-SCP_map<int, int> SDLtoFS2;
+SCP_map<SDL_Scancode, int> SDLtoFS2;
+SCP_map<int, SDL_Scancode> FS2toSDL; // Convenience map to speed up FS2->SDL lookups needed for displaying localized keycodes in config
```
z64555, line 58: Consider using only the string -> int map. To do reverse map lookups you can perform a seach on the ::second member of the key-value pair.

We can do this because both the key and the values are both unique to each other, and int searches are far simpler to check for equality than it is to check strings.

code/ui/inputbox.cpp
```diff
@@ -80,6 +80,30 @@ void UI_INPUTBOX::init_cursor()
 
 void UI_INPUTBOX::create(UI_WINDOW *wnd, int _x, int _y, int _w, int _text_len, const char *_text, int _flags, int pixel_lim, color *clr)
 {
+	textListener = os::events::addEventListener(SDL_TEXTINPUT, 10000, [this](const SDL_Event& event) {
+		bool focus = my_wnd->selected_gadget == this;
```
z64555, line 84: I'm generally not a fan of lambdas used in this fashion, I prefer instead making this a member function or a local helper function.  Will defer to others opinions, however.

code/ui/inputbox.cpp
```diff
@@ -326,11 +354,11 @@ void UI_INPUTBOX::process(int focus)
 
 	if (focus) {
 		key = my_wnd->keypress;
-		switch (key) {
+			switch (key) {
 			case 0:
 				break;
 
-			//case KEY_LEFT:
+				//case KEY_LEFT:
```
z64555, line 361: er... unnecessary tab?

code/ui/inputbox.cpp
```diff
@@ -354,12 +382,13 @@ void UI_INPUTBOX::process(int focus)
...
-				if (flags & UI_INPUTBOX_FLAG_ESC_CLR){
+				if (flags & UI_INPUTBOX_FLAG_ESC_CLR) {
 					if (position > 0) {
 						set_text("");
-						key_used = 1;						
+						key_used = 1;
 
-					} else {
+					}
+					else {
```
z64555, line 391: Please retain the existing style of `} else {` within this file.

code/ui/inputbox.cpp
```diff
@@ -389,72 +418,72 @@ void UI_INPUTBOX::process(int focus)
...
+					gr_get_string_size(&_w, nullptr, passwd_text);
 
-									if (flags & UI_INPUTBOX_FLAG_PASSWD) {
-										passwd_text[position] = 0;
-									}
-								}
-							}
-						}
+				}
+				else {
```
z64555, line 469: Please use existing `} else {` style in this file

z64555 (changes requested):

code/io/key.cpp
SelfGamer: SDL_Scancode is an Enum, so techincally just an int compare.
I did this to avoid using extra search time when searching the map by value, but if you want to keep it neater and not clutter code and memory with this reverse map, I can change it.

code/ui/inputbox.cpp
SelfGamer: uhh, sure. I personally think lambdas are quite neat, especially with capturing this, but I can make it a member function (which would need a bind though). Local helper is annoying IMO, since we do need to refer to this fairly often

code/io/key.cpp
z64555: Yes please.  I recommend making a function with the reverse map's name to do the job.

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1399,14 +1392,14 @@ int control_config_common_write_tbl(bool overwrite = false) {
 			buf_str = "NONE";
 		}
 
-		cfputs(("$Bind Name: " + item.text + "\n").c_str(), cfile);
+		puts(("$Bind Name: " + item.text + "\n").c_str(), cfile);
```
z64555, line 1395: Use our cfile versions of functions such as `puts`, as they use their own allocator and can handle other SCP containers.  Don't directly use c std stuff for file reading and writing.

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1351,25 +1353,16 @@ void control_config_common_read_tbl() {
...
-		return 1;
-	}
-
-	CFILE* cfile = cfopen("controlconfigdefaults.tbl", "w", CFILE_NORMAL, CF_TYPE_TABLES);
-	if (cfile == nullptr) {
-		// Could not open. Bail.
-		return 1;
-	}
+template<class FILETYPE>
+int control_config_common_write_tbl_segment(FILETYPE* cfile, int preset, int (* puts)(const char *, FILETYPE*) ) {
```
z64555, line 1357: Double check that the puts callback fits the same convention as the other cfile handlers.

code/cmdline/cmdline.cpp
```diff
@@ -241,6 +241,7 @@ Flag exe_params[] =
 	{ "-output_sexps",		"Output SEXPs to sexps.html",				true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-output_sexps", },
 	{ "-output_scripting",	"Output scripting to scripting.html",		true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-output_scripting", },
 	{ "-output_script_json",	"Output scripting doc to scripting.json",	true,	0,								EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-output_script_json", },
+	{ "-controlconfig_tbl",	"Save control presets to table",			true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-controlconfig_tbl", },
```
z64555, line 244: I'm actually kind of against using a commandline option for this feature, since its basically a one-shot thing.  Using the `save_ccd()` DCF instead, modders can adjust presets on a whim (as discussed on discord just recently).  I recommend expanding the DCF instead.

z64555 (changes requested):

SelfGamer:
Closing due to modified branching strategy.
This'll go into z64555/controls6

PR #2983, opened 2020-12-15
hoist model_get and model_get_instance out of a bunch of functions

Goober5000:
More housekeeping changes to the model code, similar to PR #2971, prior to the submodel rotation upgrade.  This reduces the number of model lookups when iterating over submodels and subsystems, which makes collision and rendering code ever so slightly more efficient.  But it's mostly useful for type safety and organization.

This is now ready to review.

### DIFF
--- a/code/network/multi_obj.cpp
+++ b/code/network/multi_obj.cpp
@@ -1312,38 +1312,39 @@ int multi_oo_pack_data(net_player *pl, object *objp, ushort oo_flags, ubyte *dat
 
 			// retrieve the submodel for rotation info.
 			if (subsystem->system_info->flags[Model::Subsystem_Flags::Rotates, Model::Subsystem_Flags::Dum_rotates]) {
+				angles *angs_1 = &subsystem->submodel_info_1.angs;
+				angles *angs_2 = &subsystem->submodel_info_2.angs;
 
 				// here we're checking to see if the subsystems rotated enough to send.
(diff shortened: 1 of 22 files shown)


code/network/multi_obj.cpp
```diff
@@ -1312,38 +1312,39 @@ int multi_oo_pack_data(net_player *pl, object *objp, ushort oo_flags, ubyte *dat
 
 			// retrieve the submodel for rotation info.
 			if (subsystem->system_info->flags[Model::Subsystem_Flags::Rotates, Model::Subsystem_Flags::Dum_rotates]) {
+				angles *angs_1 = &subsystem->submodel_info_1.angs;
+				angles *angs_2 = &subsystem->submodel_info_2.angs;
 
 				// here we're checking to see if the subsystems rotated enough to send.
-				if (subsystem->submodel_info_1.angs.b != Oo_info.player_frame_info[pl->player_id].last_sent[objp->net_signature].subsystem_1b[i]) {
+				if (angs_1 != nullptr && angs_1->b != Oo_info.player_frame_info[pl->player_id].last_sent[objp->net_signature].subsystem_1b[i]) {
```
asarium, line 1319: The `!=` checks here are not needed since we will always have a valid pointer here.

asarium (approved):

code/network/multi_obj.cpp
Goober5000: In this version of the code, yes, but in the upcoming version of the code these might be null.

asarium: Ah, alright then.

PR #2988, opened 2020-12-16
In windows, attach to the console we're being evoked from and reopen STDIO streams.

### LINKED ISSUE #2966: Output to stdout broken on Windows
When trying to pass any of the help options to FSO on the commandline ("-h", "-help", "--help", or  "-?") Nothing happens.

The following text should be displayed on the console:

> FreeSpace 2 Open, version <FSO_VERSION_FULL>
> Website: http://scp.indiegames.us
> Mantis (bug reporting): http://scp.indiegames.us/mantis/
> Usage: fs2_open [options]

but instead only the prompt is display.  Running through the MSVS debugger shows that it correctly detects the help options and passes through just fine, so there is an issue somewhere with printf.

Occurs on latest builds as well as 19.0.  Issue occurs on Windows, with symptoms showing on both command prompt and powershell.  Appears to not show up on linux.

Suspected cause is something is redirecting the stdout to somewhere else during the time that cmdline.cpp code executes. Perhaps to an outwnd?

This is not specific to the help options, but also applies to anything that uses stdout to output data to console or otherwise.  Some windows dev environments don't have this issue.

z64555:
Fixes #2966 

This needs some avid testing from other folks to ensure there's no side effects.  I'm not sure how SDL handles its error and debug outputs, so its possible this change may affect them.

Based on: [https://stackoverflow.com/questions/34539135/how-do-i-print-to-the-console-while-an-sdl-2-program-is-running](url)

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -7535,6 +7535,15 @@ int main(int argc, char *argv[])
 	::CoInitialize(nullptr);
 
 	SCP_mspdbcs_Initialise();
+
+	// If we're being evoked from a console, attach the STDIO streams to it and reopen the streams
+	// This is needed because Windows assumes SUBSYSTEM:WINDOWS programs won't need console IO.  Additionally, SDL
+	// seems to close or otherwise grab the streams for somthing else.
+	if (AttachConsole(ATTACH_PARENT_PROCESS)) {
+		freopen("CONIN$", "r", stdin);
+		freopen("CONOUT$", "w", stdout);
+		freopen("CONOUT$", "w", stderr);
+	}
 #else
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):

Goober5000:
I'm not sure why the CI is choking, but try rebasing onto current master.

ngld:
This breaks Knossos if it's launched in debug mode and attached to a console since FSO will directly write to the console and not to the pipe from Knossos.
Can we please revert this change? It breaks basically everything around console I/O as soon as anything in the process tree is running in a console. Hell, even the intended use case doesn't really work:

cmd_4MqOFowE68

PR #2997, opened 2020-12-20
Add Scripting API isNebula and isSubspace

SelfGamer:
This adds a functionality so that scripts can check whether the current mission is a nebula mission and if it is a subspace mission.

### DIFF
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1271,24 +1271,12 @@ ADE_FUNC(isInCampaign, l_Mission, NULL, "Get whether or not the current mission 
 
 ADE_FUNC(isNebula, l_Mission, nullptr, "Get whether or not the current mission being played is set in a nebula", "boolean", "true if in nebula, false if not")
 {
-	bool b = false;
-
-	if (The_mission.flags[Mission::Mission_Flags::Fullneb]) {
-		b = true;
-	}
-
-	return ade_set_args(L, "b", b);
+	return ade_set_args(L, "b", The_mission.flags[Mission::Mission_Flags::Fullneb]);
 }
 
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/mission.cpp
```diff
@@ -1269,6 +1269,28 @@ ADE_FUNC(isInCampaign, l_Mission, NULL, "Get whether or not the current mission
...
 
+ADE_FUNC(isNebula, l_Mission, nullptr, "Get whether or not the current mission being played is set in a nebula", "boolean", "true if in nebula, false if not")
+{
+	bool b = false;
+
+	if (The_mission.flags[Mission::Mission_Flags::Fullneb]) {
+		b = true;
+	}
+
+	return ade_set_args(L, "b", b);
```
asarium, line 1280: Would be easier to just write it like this:
```suggestion
	return ade_set_args(L, "b", The_mission.flags[Mission::Mission_Flags::Fullneb]);
```

code/scripting/api/libs/mission.cpp
```diff
@@ -1269,6 +1269,28 @@ ADE_FUNC(isInCampaign, l_Mission, NULL, "Get whether or not the current mission
...
+
+ADE_FUNC(isSubspace, l_Mission, nullptr, "Get whether or not the current mission being played is set in subspace", "boolean", "true if in subspace, false if not")
+{
+	bool b = false;
+
+	if (The_mission.flags[Mission::Mission_Flags::Subspace]) {
+		b = true;
+	}
+
+	return ade_set_args(L, "b", b);
```
asarium, line 1291: Same here.

asarium (changes requested):

code/scripting/api/libs/mission.cpp
SelfGamer: Yeah, only did it like this cause isInCampaign did it too.

asarium (approved):

PR #3000, opened 2020-12-20
another model to model_instance refactor

Goober5000:
A small follow-up to #2992.  This removes arc information from bsp_info and moves it to submodel_instance.  It also adds a model instance to each debris chunk to keep track of it.


wookieejedi (approved):
LGTM

Issue #2952, opened 2020-12-03
Multilock lock indicators spinning too long


EatThePath:
Currently with multilock the target lock indicators all keep spinning until all are locked, when at some point in the past they stopped spinning once they were an actual lock and only the unlocked ones kept spinning.

At this point I'm not entirely certain what the ideal or correct behavior is, to be honest.

Goober5000:
Can we get a lightweight mod to reproduce this?  I looked at a video posted on Discord but all of the indicators locked simultaneously, so it wasn't a good demonstration.

I also took a quick look through the code but couldn't see any obvious bug.

EatThePath:
@Goober5000 I discussed the issue with @TRBlount some on discord after posting it, and I intend to bisect it later this week. When I do I'll make a clear-as-possible test case and post it here.

EatThePath:
Bisection result.

0727: Multi-lock was newly merged, and lock indicators were broken in that they stop spinning after the first lock
1028: Broken as in original merge
1029: Lock indicators spin forever
1210: Still spinning 

It looks like if the indicators ever worked as intended it was sometime before the merge. #2839 appears to be where this change happened, but it was broken in a different way before that, not actually working.

Repro file has a weapon table and a mission. Start the mission and you should immediately see locking begin.
[lockbug.zip](https://github.com/scp-fs2open/fs2open.github.com/files/5674706/lockbug.zip)

JohnAFernandez:
@EatThePath Thanks for the bug mod and bisection, I'm going to take a look now.

JohnAFernandez:
Should be fixed via #3002

PR #3002, opened 2020-12-20
Set correct time stamp for already-locked indicators.

JohnAFernandez:
This should handle #2952 by setting a consistent timestamp for locked indicators instead of accidentally inheriting them from indicators that are still locking.  Tested with normal and multilock missiles.

Took entirely too long to track down.


Goober5000 (approved):
Looks good. Short and sweet.

PR #3003, opened 2020-12-21
Expose disposing FBO's to scripting

SelfGamer:
As discussed on Discord with @asarium, there should be a way to detach FBO's from created textures when done rendering.
After a look through the Graphics code, I think implementing FBO's as an object type for scripting would be hard.
However, it seemed like it was already intended at some point to have the option to detach an FBO while the bitmap stays intact.
To expose this (and slightly fix it, as it seems to have never been used), I did the following:

1. Expose ``void bm_free_data(bitmap_slot* n, bool release = false);`` in a header file
2. Made the release bool from that function relevant for actually clearing the bitmap_slot, as opposed to only pass it to the graphics API (which would, if set to false, not clear the texture slot)
3. Modified ``opengl_kill_render_target``, as clearing the FSO-FBO object would also clear the rendered texture (although I don't fully understand why). However, the OpenGL FBO is still always cleared.
4. Wrapped the ``bm_free_data(n, false);`` in a scripting function that can be called on a texture. If called on a non-render texture, the checks for detaching FBO's will trigger and nothing will happen, so this should be safe.

### DIFF
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -207,11 +207,6 @@ void clear_bm_lookup_cache() {
 static void bm_convert_format(bitmap *bmp, ushort flags);
 
 /**
- * Frees a bitmap's data if it can
- */
-static void bm_free_data(bitmap_slot* n, bool release = false);
-
-/**
  * A special version of bm_free_data() that can be safely used in gr_*_texture
  * to save system memory once textures have been transfered to API memory
  * it doesn't restore the slot to a pristine state, it only releases the data
@@ -622,6 +617,10 @@ void bm_free_data(bitmap_slot* bs, bool release)
 	bmp = &be->bm;
 
 	gr_bm_free_data(bs, release);
+
+	//If we only unattach the RenderTarget but not the bitmap, abort here
+	if (!release)
+		return;
 
 	// If there isn't a bitmap in this structure, don't
 	// do anything but clear out the bitmap info
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -207,6 +207,11 @@ void clear_bm_lookup_cache() {
 static void bm_convert_format(bitmap *bmp, ushort flags);
 
 /**
+ * Frees a bitmap's data if it can
+ */
+static void bm_free_data(bitmap_slot* n, bool release = false);
+
+/**
  * A special version of bm_free_data() that can be safely used in gr_*_texture
  * to save system memory once textures have been transfered to API memory
  * it doesn't restore the slot to a pristine state, it only releases the data
@@ -617,10 +622,6 @@ void bm_free_data(bitmap_slot* bs, bool release)
 	bmp = &be->bm;
 
 	gr_bm_free_data(bs, release);
-
-	//If we only unattach the RenderTarget but not the bitmap, abort here
-	if (!release)
-		return;
 
 	// If there isn't a bitmap in this structure, don't
 	// do anything but clear out the bitmap info
@@ -2851,6 +2852,38 @@ int bm_release(int handle, int clear_render_targets) {
 	}
 
 	return 1;
+}
+
+bool bm_release_rendertarget(int handle) {
+	Assert(handle >= 0);
+
+	bitmap_entry* be = bm_get_entry(handle);
+
+	if (be->type == BM_TYPE_NONE) {
+		return false;	// Already been released?
+	}
+
+	Assertion(be->handle == handle, "Invalid bitmap handle number %d (expected %d) for %s passed to bm_release_rendertarget()\n", be->handle, handle, be->filename);
+
+	if (!((be->type == BM_TYPE_RENDER_TARGET_STATIC) || (be->type == BM_TYPE_RENDER_TARGET_DYNAMIC))) {
+		nprintf(("BmpMan", "Tried to release a render target of a non-rendered bitmap!\n"));
+		return false;
+	}
+
+	if (bm_is_anim(be)) {
+		int i, first = be->info.ani.first_frame;
+		int total = bm_get_entry(first)->info.ani.num_frames;
+
+		// Go through all frames and release each ones rendertarget 
+		for (i = 0; i < total; i++) {
+			gr_bm_free_data(bm_get_slot(first + i), false);
+		}
+	}
+	else {
+		gr_bm_free_data(bm_get_slot(handle), false);		// clears flags, bbp, data, etc
+	}
+
+	return true;
 }
 
 int bm_reload(int bitmap_handle, const char* filename) {
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -2880,7 +2880,7 @@ bool bm_release_rendertarget(int handle) {
 		}
 	}
 	else {
-		gr_bm_free_data(bm_get_slot(handle), false);		// clears flags, bbp, data, etc
+		gr_bm_free_data(bm_get_slot(handle), false);
 	}
 
 	return true;
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -2864,24 +2864,14 @@ bool bm_release_rendertarget(int handle) {
 	}
 
 	Assertion(be->handle == handle, "Invalid bitmap handle number %d (expected %d) for %s passed to bm_release_rendertarget()\n", be->handle, handle, be->filename);
+	Assertion(!bm_is_anim(be), "Cannot release a render target of an animation (bitmap handle number %d for %s)!\n", be->handle, be->filename);
 
 	if (!((be->type == BM_TYPE_RENDER_TARGET_STATIC) || (be->type == BM_TYPE_RENDER_TARGET_DYNAMIC))) {
 		nprintf(("BmpMan", "Tried to release a render target of a non-rendered bitmap!\n"));
 		return false;
 	}
 
-	if (bm_is_anim(be)) {
-		int i, first = be->info.ani.first_frame;
-		int total = bm_get_entry(first)->info.ani.num_frames;
-
-		// Go through all frames and release each ones rendertarget 
-		for (i = 0; i < total; i++) {
-			gr_bm_free_data(bm_get_slot(first + i), false);
-		}
-	}
-	else {
-		gr_bm_free_data(bm_get_slot(handle), false);
-	}
+	gr_bm_free_data(bm_get_slot(handle), false);
 
 	return true;
 }
--- a/code/graphics/opengl/gropengltexture.cpp
+++ b/code/graphics/opengl/gropengltexture.cpp
@@ -1601,7 +1601,7 @@ int opengl_check_framebuffer()
 	return 0;
 }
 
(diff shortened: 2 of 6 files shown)


asarium (changes requested):
I'm sorry but this does not fix the issue you are facing. All you need is an API for scaling a monochrome image and the whole FBO thing is just a workaround.

This also exposes an internal function that can have very unintended consequences when used incorrectly so making these changes is potentially dangerous.

SelfGamer:
While scaling the monochrome images is certainly part of the issue, I would also like to be able to generally render on images.
Currently, in terms of the radar-script this would be used in, other uses that need iconbaking are we intend to load border and infill separately and then combine them based on parameters, as well as automatically create icons for "Wings of Craft" that will be multiple icons next to each other.
While the monochrome scaling is certainly a useful and needed feature, as long as we have an API for drawing to images in LUA that will crash unless we clean up after it, we should be able to do exactly that. Otherwise what's the point of drawing to textures, if we can't reliably use it.

In regards to exposing an internal function:
Would it be better to write a bm_release-esque handler that wraps this function internally?

asarium:
Regarding the use cases you mentioned, most of those can be done by simply drawing the icons in real time so you don't need to go through prebaking them. For performance, the different is probably not significant anyway unless you can avoid dozens of draw calls per icon with the prebaking code.

For getting this merged it will at least need error checking and a dedicated `bmpman` API for freeing the render target of a RTT handle.
Also, `bm_get_slot` is an internal API that should not be used by outside code. Not sure why that is even available in that function...

SelfGamer:
> For getting this merged it will at least need error checking and a dedicated bmpman API for freeing the render target of a RTT handle.

Alright, I'll try to implement that. I feel like, even if I don't end up using it, this should be a thing to facilitate use of gr.createTexture without fear of crashing the engine due to too many Framebuffers without having to fully unload textures with :unload (which defeats the purpose of drawing on them).

>Also, bm_get_slot is an internal API that should not be used by outside code. Not sure why that is even available in that function...

That's cause I made it visible. Will change that back for the next try.

code/bmpman/bmpman.cpp
```diff
@@ -2854,6 +2854,38 @@ int bm_release(int handle, int clear_render_targets) {
...
+	}
+
+	Assertion(be->handle == handle, "Invalid bitmap handle number %d (expected %d) for %s passed to bm_release_rendertarget()\n", be->handle, handle, be->filename);
+
+	if (!((be->type == BM_TYPE_RENDER_TARGET_STATIC) || (be->type == BM_TYPE_RENDER_TARGET_DYNAMIC))) {
+		nprintf(("BmpMan", "Tried to release a render target of a non-rendered bitmap!\n"));
+		return false;
+	}
+
+	if (bm_is_anim(be)) {
```
asarium, line 2873: Render targets are never animations so this is not needed. You could even turn it into an assertion so that we catch this when we get into some weird state.

code/graphics/opengl/gropengltexture.cpp
```diff
@@ -1627,7 +1627,8 @@ void opengl_kill_render_target(bitmap_slot* slot)
 		fbo->renderbuffer_id = 0;
 	}
 
-	opengl_free_fbo_slot(fbo->fbo_id);
+	if(remove_slot)
```
asarium, line 1630: What exactly is this supposed to do? At this point the OpenGL framebuffer object is already destroyed so the only thing this controls is whether we free the slot in our array.

code/scripting/api/objs/texture.cpp
```diff
@@ -117,6 +117,21 @@ ADE_FUNC(unload, l_Texture, NULL, "Unloads a texture from memory", NULL, NULL)
 	return ADE_RETURN_NIL;
 }
 
+ADE_FUNC(unloadRenderTarget, l_Texture, nullptr, "Unloads a texture's render target. Call this when done drawing to a texture, as it frees up resources.", nullptr, nullptr)
```
asarium, line 120: "unload" is not the right term here since we do not delete any data. Maybe `detachRenderTarget` or simply `destroyRenderTarget`?

Also maybe put into the description that this is needed for cases where a render target is only drawn to once but rendered many times.

code/scripting/api/objs/texture.cpp
```diff
@@ -117,6 +117,21 @@ ADE_FUNC(unload, l_Texture, NULL, "Unloads a texture from memory", NULL, NULL)
...
+{
+	texture_h* th;
+
+	if (!ade_get_args(L, "o", l_Texture.GetPtr(&th)))
+		return ADE_RETURN_NIL;
+
+	if (!th->isValid())
+		return ADE_RETURN_NIL;
+
+	bm_release_rendertarget(th->handle);
```
asarium, line 130: Please verify before this that `bm_is_render_target` is true and cause a `LuaError` otherwise.

asarium (changes requested):

code/bmpman/bmpman.cpp
SelfGamer: Okay. I wasn't sure about that, but thanks for confirming.
Assert or return 0; with debug message?

asarium: Assert is appropriate here since a failure of that condition indicates a programming error.

code/graphics/opengl/gropengltexture.cpp
SelfGamer: I am not 100% sure why this is, but removing the object from our array will result in the Image not being renderable anymore (will just be transparent), If you leave this object, it's fine.

asarium: Not sure why this would cause rendering to fail. However, since not calling this function effectively produces a memory leak it will need to be fixed before this is merged.

code/scripting/api/objs/texture.cpp
```diff
@@ -117,6 +117,24 @@ ADE_FUNC(unload, l_Texture, NULL, "Unloads a texture from memory", NULL, NULL)
...
+	texture_h* th;
+
+	if (!ade_get_args(L, "o", l_Texture.GetPtr(&th)))
+		return ADE_RETURN_NIL;
+
+	if (!th->isValid())
+		return ADE_RETURN_NIL;
+
+	if(!bm_is_render_target(th->handle))
+		LuaError(L, "Tried to destroy a render target of a non-renderable texture!");
```
asarium, line 131: `LuaError` can still return normally so you also need to `return` here to ensure that `bm_release_rendertarget` is not called.

code/graphics/opengl/gropengltexture.cpp
```diff
@@ -1554,8 +1554,13 @@ static void opengl_free_fbo_slot(int id) {
 
 	Assertion(fbo != nullptr, "Invalid id passed to opengl_free_fbo_slot!");
 
+	//We need to reset the FBO but keep it's ID, otherwise they'll get reassigned with ID 0
+	int fbo_id = fbo->fbo_id;
+
 	// Reset this slot using the default constructor
 	*fbo = fbo_t();
+
+	fbo->fbo_id = fbo_id;
```
asarium, line 1563: This is still a memory leak since that FBO slot will never be reused if the id is kept as a non-zero value.

asarium (changes requested):

code/graphics/opengl/gropengltexture.cpp
asarium: Ok, that fixed the memory leak but can you go into detail a bit why this is necessary? It's not good to basically leave dangling references around.

SelfGamer: So, the issue is that the code makes two assumptions:
A released FBO Object is immediately good to go (hence the return &target; once it finds a free buffer), and that new ID's that are assigned are always increased by one.
There are three options to solve this:
1. Don't save old, freed FBO pointer, and never assign another FBO object with the ID this object once had (which now would be free)
2. Keep the old, freed FBO pointer, create a fresh fbo object in it (that was what the code did before), and keep the ID this pointer has (that's what I changed, otherwise all kept fbo's would have ID 0, which is bad).
3. Keep the old, freed FBO pointer, create a fresh fbo object in it, and once it is reused, assign it a free ID. I didn't do this, as it'd either mean iterating through FBO Objects until we find an ID that is free (and given this is techincally an unordered vector, that's expensive), or keeping track of freed ID's.

SelfGamer: TL;DR, assigning the new object this ID is less of a "dangling reference" and more of a: We reset the FBO Object that has this ID.
The proper way would be to save FBO objects as a map with their ID as the key, but that'd be a major overhaul of the whole system

asarium: You described what is going on but now _why_ your changes are the correct ones. What does keeping the id change in the runtime behavior that fixes rendering of detached render targets?

SelfGamer: Ah, my bad.
The next rendertarget being created after an detachment get's assigned the old FBO Object.
Without the change, it'll always have the ID 0 (which is assumed to then be a non-existing FBO, see check that prints "Tried to get an OpenGL FBO that didn't exist!" when ID = 0). And that means, that setTarget(texture) will fail, as the FBO doesn't have a legit ID
With the change, the FBO Object has a legit and unique ID, so that setting that texture as a target actually works.
So it's less that the texture can't be rendered without this change, but rather that it can't be rendered _to_, and is thus always black or transparent (either of which AABitmap then displays as transparent, making it seem like it had never rendered)

asarium: Ahh, now I see. Thanks for the explanation and the detective work here!
However, with that in mind then it would be more correct to change `opengl_get_free_fbo` so that a new FBO ID is created (using `next_fbo_id`) instead of leaving it at `-1` which is just completely wrong.

I guess so far noone bothered to clean up their render targets so this bug was never discovered...

SelfGamer: If you always assign a new ID, then you will end up with ID's of freed up FBO's being unused, rather than reused like this. Would that be preferable to my current solution?

asarium: At least in my opinion, that is the preferred solution since it simplifies the cleanup code and keeps the FBO structs in a "default" state when they are not in use.

code/graphics/opengl/gropengltexture.cpp
```diff
@@ -1534,16 +1534,21 @@ static fbo_t* opengl_get_fbo(int id) {
...
-	last.fbo_id = next_fbo_id;
+	if (fbo == nullptr) {
+		RenderTarget.push_back(fbo_t());
+		fbo = &RenderTarget.back();
+	}
+	
+	fbo->fbo_id = next_fbo_id;
 	++next_fbo_id;
 
 	return &RenderTarget.back();
```
asarium, line 1554: You need to return `fbo` here.

asarium (changes requested):

code/graphics/opengl/gropengltexture.cpp
SelfGamer: oops

asarium (approved):

PR #3004, opened 2020-12-21
Add AA bitmap support for standard gr.drawImage

asarium:
Since that function already has full support for UV scaling it makes
sense to unify the behavior in one function instead of having a separate
monochrome function that does weird things.

Also deprecates that function since all its use cases are now covered by
drawImage.


Goober5000:
The Inferno team will be pleased to see this.  I'll ask them to test it.

Goober5000 (approved):
Approving for RC1

PR #3008, opened 2020-12-22
additional null checks

Goober5000:
Found a null pointer CTD in modelanim that slipped through initial testing.  Also added some defense-in-depth null checks to scripting.

### DIFF
--- a/code/model/modelanim.cpp
+++ b/code/model/modelanim.cpp
@@ -868,8 +868,10 @@ void model_anim_set_initial_states(ship *shipp)
 			if (psub->triggers[i].type == AnimationTriggerType::Initial) {
 				if (psub->type == SUBSYSTEM_TURRET) {
 					// special case for turrets
-					pss->submodel_instance_2->angs.p = psub->triggers[i].angle.xyz.x;
-					pss->submodel_instance_1->angs.h = psub->triggers[i].angle.xyz.y;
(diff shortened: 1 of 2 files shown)


code/model/modelanim.cpp
```diff
@@ -868,7 +868,8 @@ void model_anim_set_initial_states(ship *shipp)
 			if (psub->triggers[i].type == AnimationTriggerType::Initial) {
 				if (psub->type == SUBSYSTEM_TURRET) {
 					// special case for turrets
-					pss->submodel_instance_2->angs.p = psub->triggers[i].angle.xyz.x;
+					if (pss->submodel_instance_2 != nullptr)
+						pss->submodel_instance_2->angs.p = psub->triggers[i].angle.xyz.x;
 					pss->submodel_instance_1->angs.h = psub->triggers[i].angle.xyz.y;
```
asarium, line 873: I see you added null checks for the `_1` variant below so is that also necessary here?

asarium (approved):

code/model/modelanim.cpp
Goober5000: In the scripting case, a script might call the functions on any old subsystem, whether it's a turret or not.  In this case, I would assume upstream checks should check this before we reach this point.  However, that's an assumption, so I think a null check does make sense.

PR #3010, opened 2020-12-22
Add scripting API to access briefing color tags

asarium:


asarium:
I found another issue with the new UI code so I'll just include that fix here as well.

code/scripting/api/libs/ui.cpp
```diff
@@ -386,6 +390,46 @@ ADE_LIB_DERIV(l_UserInterface_CmdBrief,
 	"API for the new UI system. This should not be used by other code and may be removed in the future!",
 	l_UserInterface);
 
+ADE_VIRTVAR(ColorTags,
+	l_UserInterface_CmdBrief,
+	nullptr,
+	"The available tagged colors",
+	ade_type_map("string", "color"),
+	"A apping from tag string to color value")
```
Goober5000, line 398: I assume this should be "A mapping"

code/scripting/api/objs/color.cpp
```diff
@@ -0,0 +1,73 @@
...
+			CLAMP(newVal, 0.0f, 255.0f);
+		}
+
+		clr->*member = static_cast<ubyte>(newVal);
+	}
+
+	return ade_set_args(L, "i", clr->*member);
+}
+
+ADE_VIRTVAR(Red,
```
Goober5000, line 31: I suggest changing this to `RedValue` or something similar; and the same with Green and Blue.  A scripter might think that `color.Red` is a factory method that returns a Color object set to pure red.

Goober5000 (changes requested):
Looks good overall but I had a few comments.

code/scripting/api/libs/ui.cpp
asarium: Correct!

code/scripting/api/objs/color.cpp
asarium: In Lua, the `color` "class" is not available to scripts and `color.Red` is not a valid expression so this should not be an issue.

Goober5000: Well it's not my preference, but I'll approve it.

Goober5000 (approved):

PR #3011, opened 2020-12-22
Prevent almost immediate crash and Warn User about corrupt/unreachable SDL.dll

JohnAFernandez:
This may not be the correct approach for this, but the code here crashes without a window open or debug log created if SDL cannot get the preferred path.  I started encountering this crash without any noticeable cause except for the fact that my builds were being created in a sub-directory of documents (which I've been doing for 18 months without issue).  

I went with the option of having a pretty verbose warning message, telling the user FSO will attempt to use the current directory, but will probably crash without a warning message.

This avoids the immediate crash, and warns the user, but it cannot prevent the other things that will occur, like a possible crash when trying to load OpenAL.  

I don't mind switching to an Error instead of a warning if requested.

### DIFF
--- a/code/osapi/osapi.cpp
+++ b/code/osapi/osapi.cpp
@@ -42,8 +42,20 @@ namespace
 		// Lazily initialize the preferences path
 		if (!preferencesPath) {
 		    preferencesPath = SDL_GetPrefPath(ORGANIZATION_NAME, APPLICATION_NAME);
-		    if (!preferencesPath) {
-			    mprintf(("Failed to get preferences path from SDL: %s\n", SDL_GetError()));
+			
+			// this section will at least tell the user if something is seriously wrong instead of just crashing without a message or debug log.
+			// It may crash later, especially when trying to load sound. But let's let it *try* to run in the current directory at least.
+		    if (preferencesPath == nullptr) {
+				static bool sdl_is_borked_warning = false;
+				if (!sdl_is_borked_warning) {
+					ReleaseWarning(LOCATION, "%s\n\nSDL and Windows are unable to get the preferred path for the reason above. "
+						"Installing FSO, its executables and DLLs in another non-protected folder may fix the issue.\n\n"
+						"You may experience issues if you continue playing, and FSO may crash. Please report this error if it persists.\n\n"
+						"Report at www.hard-light.net or the hard-light discord.", SDL_GetError());
+					sdl_is_borked_warning = true;
+				}
+				// No preferences path, try current directory.  
+				return "." DIR_SEPARATOR_STR;
 		    }
 #ifdef WIN32
 		    try {


asarium:
What prompted this change? More `null` checks are always nice but I am a bit concerned about the error message. If the SDL DLL is not available then Windows should just refuse to start the executable outright.
It is possible for the SDL API to return a `null` value but that only happens if the underlying Windows API failed. I think it would be better to just have a general "Determining the preferences path failed" error since this issue has not happened often so far.

JohnAFernandez:
I believe it has happened before to general users, and I think it was being misdiagnosed as a knossos bug, since there is a similar knossos bug, the still-present joystick selection bug.  Spart_n ran into something very similar, and the only thing that ended up "fixing" it was erasing everything Freespace related from his computer.    

The real problem with this portion of code is that when it crashes here, it is before the debug log is started, and so we should check for null and tell the user *something* so that someone who usually uses the debug log to figure out what is happening will be completely unable to help.

I wrote the message with my best guess as to what could be causing it and the best ways to address it, because short, vague error messages are not usually that helpful to end users.  Perhaps a pared down message along the lines of. "SDL and Windows are unable to get the preferred path for an unknown reason. Installing the executable and DLL in another location may fix the issue. Please report this error if it persists."  

I'm open to other versions, as I'm trying to make the message accessible enough so that someone like EatThePath or MjnMixael could at least start the process of debugging if a coder is not available, instead of blindly guessing that something is wrong with knossos.

JohnAFernandez:
Oh, the change in freespace.cpp was unrelated and should not be there.

code/osapi/osapi.cpp
```diff
@@ -42,8 +42,21 @@ namespace
...
 		    preferencesPath = SDL_GetPrefPath(ORGANIZATION_NAME, APPLICATION_NAME);
-		    if (!preferencesPath) {
-			    mprintf(("Failed to get preferences path from SDL: %s\n", SDL_GetError()));
+			
+			// this section will at least tell the user if something is seriously wrong instead of just crashing without a message or debug log.
+			// It may crash later, especially when trying to load sound. But let's let it *try* to run in the current directory at least.
+		    if (preferencesPath == nullptr) {
+				static bool sdl_is_borked_warning = false;
+				if (!sdl_is_borked_warning) {
+					Warning(LOCATION, "%s\n\nFSO is having issues using SDL.dll, or it was not included. The build and dll may be in a protected folder,"
```
asarium, line 51: This should be a `ReleaseWarning` so that it will show up for Release build.

freespace2/freespace.cpp
```diff
@@ -6485,8 +6485,10 @@ int game_main(int argc, char *argv[])
 	// maybe spew pof stuff
 	if (Cmdline_spew_pof_info) {
 		game_spew_pof_info();
-		game_shutdown();
-		return 0;
+		if (!Cmdline_verify_vps) {
+			game_shutdown();
+			return 0;
+		}
```
asarium, line 6491: This appears to be unrelated.

asarium (changes requested):

code/osapi/osapi.cpp
JohnAFernandez: Changed, and I tried to make the message descriptive but also not try to do everything under the sun..

asarium (approved):

PR #3012, opened 2020-12-22
additional null checks

Goober5000:
Some subsystem code was being run for submodels when the subsystem didn't necessarily have a submodel attached.


JohnAFernandez (approved):
+1 for nullptr safety

The-E (approved):

PR #3015, opened 2020-12-23
move gun_submodel_rotation from model to instance

Goober5000:
Since instance-specific fields have been moved to polymodel_instance and submodel_instance, the gun_submodel_rotation field should also be moved.  This is also reworked a bit to keeping track of the rotation via the normal submodel instance angles.

This removes an unnecessary look_at field as well.


wookieejedi (approved):
LGTM

PR #3016, opened 2020-12-23
Use parse items

JohnAFernandez:
The starting commits for using parse items. But in a branch instead of in a regular pull request because 1) this is a lot of work and it's better to get feedback sooner rather than later 2) it would be nice if other coders could contribute if/when they can.

### DIFF
--- a/code/parse/parse_manager.h
+++ b/code/parse/parse_manager.h
@@ -1,0 +1,77 @@
+#include "globalincs/systemvars.h"
+#include "globalincs/pstypes.h"
+
+constexpr int NO_ENFORCED_COUNT = -1;
+constexpr int INVALID_VERSION = -1;
+
+enum class Parse_Input_Types
+{
+	USAGE_NOTE,
+	STRING,
+	XSTR,
+	INT,
+	INT_LIST,
+	LONG,
+	FLOAT,
+	FLOAT_LIST,
+	DOUBLE,
+	VEC2,
+	VEC3,
+	BOOL,
+	FLAGS,
+	MARKER,
+	MODEL_FILENAME,
+	TEXTURE_FILENAME,
+	SOUND_FILENAME
+};
+
+class parse_item {
+	friend class parse_item_builder;
+
+private:
+	// basic description 
+	SCP_string	_table_string   = "FIX ME!";
+	SCP_string	_description    = "Someone didn't initialize me correctly. Go get a coder.";
+
+	Parse_Input_Types	_type   = Parse_Input_Types::USAGE_NOTE;
+	// rules for this element
+	bool		_required		= false;
+	int			_enforced_count = NO_ENFORCED_COUNT;		// if this is a list that has a specific number of entries, mark it with this variable
+	int			_major_version_number_implemented = INVALID_VERSION;
+	int			_minor_version_number_implemented = INVALID_VERSION;
+	int			_revision_number_implemented	  = INVALID_VERSION;
+	int			_major_version_number_deprecated  = INVALID_VERSION;
+	int			_minor_version_number_deprecated  = INVALID_VERSION;
+	int			_revision_number_deprecated		  = INVALID_VERSION;
+	SCP_string	_deprecation_message			  = "";
+
+	void reset_fields();
+
+public:
+	parse_item() = default;
+	
+};
+
+class parse_item_builder {
+	parse_item _product;
+public:
+	parse_item_builder();
+
+	parse_item_builder new_parse_item(SCP_string name);
+	parse_item_builder set_description(SCP_string description);
+	parse_item_builder set_type(Parse_Input_Types type);
+	parse_item_builder set_required(bool required);
+	parse_item_builder set_enforced_count(int enforced_count);
+	parse_item_builder set_implementation_version(int major_implementation_version, int minor_implementation_version, int revision_implementation_number);
+	parse_item_builder set_deprecation_version(int major_deprecation_version, int minor_deprecation_version, int revision_deprecation_number);
+	parse_item_builder set_deprecation_message(SCP_string deprecation_message);
+	
+	parse_item* finished_product();
+};
+
+class table_manager {
+protected:
+	SCP_vector<parse_item> _all_items;
+public:
+	void dump_to_file();
+};
--- a/code/parse/parse_manager.h
+++ b/code/parse/parse_manager.h
@@ -1,3 +1,9 @@
+#pragma
+
+#ifndef parse_manager_h
+#define parse_manager_h
+
+
 #include "globalincs/systemvars.h"
 #include "globalincs/pstypes.h"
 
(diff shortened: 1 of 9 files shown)


code/parse/parse_manager.h
```diff
@@ -0,0 +1,60 @@
+#include "globalincs/systemvars.h"
+#include "globalincs/pstypes.h"
+
+enum class Parse_Input_Types : std::uint8_t
```
asarium, line 4: I don't see a reason to have a type for this enum since it is not used for serialization anyway.

code/parse/parse_manager.h
```diff
@@ -0,0 +1,60 @@
...
+	Parse_Input_Types	_type;
+	// rules for this element
+	bool		_required;
+	int			_enforced_count;		// if this is a list that has a specific number of entries, mark it with this variable
+	int			_major_version_number_implemented;
+	int			_minor_version_number_implemented;
+	int			_revision_number_implemented;
+	int			_major_version_number_deprecated;
+	int			_minor_version_number_deprecated;
+	int			_revision_number_deprecated;
```
asarium, line 40: Please initialize these with default values.

code/parse/parse_manager.h
```diff
@@ -0,0 +1,60 @@
...
+	SCP_string	_deprecation_message;
+
+public:
+	parse_item() = default;
+	// our factory functions
+	parse_item(SCP_string* table_string);
```
asarium, line 46: Single parameter constructors need to be marked as `explicit`.
Also, parameters should be passed by `const &` for non-value types (also below).

code/parse/parse_manager.h
```diff
@@ -0,0 +1,60 @@
...
+	parse_item set_description(SCP_string* description);
+	parse_item set_type(Parse_Input_Types type);
+	parse_item set_required(bool required);
+	parse_item set_enforced_count(int enforced_count);
+	parse_item set_implementation_version(int major_implementation_version, int minor_implementation_version, int revision_implementation_number);
+	parse_item set_deprecation_version(int major_deprecation_version, int minor_deprecation_version, int revision_deprecation_number);
+	parse_item set_deprecation_message(SCP_string* deprecation_message);
```
asarium, line 53: If this is supposed to be a factory class, then you probably also want to return by reference here (and above).

code/parse/tables/ai_class_info.cpp
```diff
@@ -0,0 +1,108 @@
+#include "ai_class_info.h"
```
asarium, line 1: I would prefer to keep the table definition close to the code that also parses the table so this would be better to have somewhere in the AI code.

code/parse/parse_manager.h
```diff
@@ -0,0 +1,60 @@
...
+	
+};
+
+
+struct weapon_info_parse_items {
```
asarium, line 58: Left over?

asarium (changes requested):

code/parse/parse_manager.h
JohnAFernandez: I think this is handled yet.  Please tell me if I misunderstood.

code/parse/parse_manager.h
JohnAFernandez: It was.

code/parse/parse_manager.h
JohnAFernandez: Default values should now be set in the struct instead of the constructor.

code/parse/parse_manager.h
JohnAFernandez: I think I marked the correct line as explicit, but you'll have to forgive me because I'm still a little fuzzy in this area.  Passing `const &` looks like...

`parse_item* set_enforced_count(const int* enforced_count)` on the declaration?

code/parse/tables/ai_class_info.cpp
JohnAFernandez: It should be moved now.

asarium (changes requested):
Please fix compilation. I will take a look once the current errors have been resolved.

code/parse/parse_manager.h
```diff
@@ -0,0 +1,77 @@
...
+	parse_item_builder();
+
+	parse_item_builder new_parse_item(SCP_string name);
+	parse_item_builder set_description(SCP_string description);
+	parse_item_builder set_type(Parse_Input_Types type);
+	parse_item_builder set_required(bool required);
+	parse_item_builder set_enforced_count(int enforced_count);
+	parse_item_builder set_implementation_version(int major_implementation_version, int minor_implementation_version, int revision_implementation_number);
+	parse_item_builder set_deprecation_version(int major_deprecation_version, int minor_deprecation_version, int revision_deprecation_number);
+	parse_item_builder set_deprecation_message(SCP_string deprecation_message);
```
asarium, line 67: All these functions need to return a reference instead to avoid copying the factory for every call.

code/parse/parse_manager.h
```diff
@@ -0,0 +1,77 @@
...
+	
+	parse_item* finished_product();
```
asarium, line 69: You should return the item by value instead and use `std::move` when returning it since pointers to internal values are a bit dangerous if used incorrectly.

code/parse/parse_manager.h
```diff
@@ -0,0 +1,77 @@
...
+};
+
+class table_manager {
```
asarium, line 72: There is no need to use inheritance for the manager. Instead, what is could do is managing the parse item vector and providing a function for creating a parse item factory.

asarium (changes requested):

code/parse/parse_manager.h
JohnAFernandez: I don't understand what you're trying to say.

In short, my reason for using inheritance is to keep from having to have duplicate functions and to keep the items private/protected within the struct.  There are probably going to be at least three main functions that these manager will oversee, with fine detail being provided in each parse item.  

Currently working on the rest.

JohnAFernandez: Just realized, I was also trying to have the table info be added in the constructor.

code/parse/parse_manager.h
JohnAFernandez: Ok.  That should be done.

code/parse/parse_manager.h
JohnAFernandez: Let me know if I put `std::move` in the wrong place.

code/parse/parse_manager.h
asarium: My reasoning probably boils down to "Composition over inheritance". If you want to share functionality between several use-cases, then extract that functionality into an external class and call that from the places where it is needed. Inheritance is actually very rarely _really_ useful. Most times, composition is what you want and makes code much easier to understand.

JohnAFernandez: All right, well, changes are made.  If it passes again, I hope you can take another look before the weekend is over.

PR #3017, opened 2020-12-23
Fix clang tidy commit range

asarium:
By only looking at the HEADs of the respective branches, we included
irrelevant changes in the diff passed to clang-tidy-diff so it was
possible to generate irrelevant warnings in a PR.


asarium:
Also noticed that the names for the Nightly and Release workflows was not correct so I also fixed that here.

JohnAFernandez (approved):

PR #3019, opened 2020-12-24
Changed require exact los to track from the exact firing point instead of center

SelfGamer:
As this feature is very new, it's probably fine to change it instead of adding a new flag.
This change make the actual firing point of the turret / secondary bank matter and not just raycast center to center, which should help for large capital ships using this flag.
While it seems to work for me, I'd like some help with testing. Pinging @EatThePath, as he requested this change for a specific mission it seems.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -5775,6 +5775,15 @@ int ai_fire_primary_weapon(object *objp)
 		aip->primary_select_timestamp = timestamp(5 * 1000);	//	Maybe change primary weapon five seconds from now.
 	}
 
+	//We can only check LoS if we have a target defined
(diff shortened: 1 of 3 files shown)


SelfGamer:
Made to draft to investigate if we can make beams respect this flag.
General change seems to have worked

SelfGamer:
Seemed to work fine, with primaries now also respecting LoS

EatThePath (commented):
I don't feel equipped to comment on the code quality itself, but I will say that I've tested this flag+branch on a handful of decently crowded missions, and as far as I can see it is stable and causes no noticeable behaviour changes outside of what it is supposed to.

Baezon (approved):
LGTM

PR #3021, opened 2020-12-25
some refactor and cleanup

Goober5000:
1) modify and prune some turret function definitions
2) clean up KeldorKatarn's submodel orientation parsing

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -462,6 +462,56 @@ void get_user_prop_value(char *buf, char *value)
 	*p1 = c;
 }
 
+// routine to parse out a vec3d from a user property field of an object
+bool get_user_vec3d_value(char *buf, vec3d *value, bool require_brackets)
+{
+	float f1, f2, f3;
+	char closing_bracket = '\0';
+	bool success = false;
+
+	pause_parse();
+	Mp = buf;
+
+	do {
+		// skip white space and equal sign
+		while (isspace(*Mp) || (*Mp == '='))
+			Mp++;
+
+		if (require_brackets)
+		{
+			// look for vector bracket
+			if (*Mp == '{')
+				closing_bracket = '}';
+			else if (*Mp == '[')
+				closing_bracket = ']';
+			else
+				break;
+		}
+
+		// get comma-separated floats
+		if (stuff_float_optional(&f1) != 2)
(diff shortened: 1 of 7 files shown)


code/model/modelread.cpp
```diff
@@ -462,6 +462,56 @@ void get_user_prop_value(char *buf, char *value)
...
+		{
+			ignore_white_space();
+
+			if (*Mp != closing_bracket)
+				break;
+		}
+
+		value->xyz = { f1, f2, f3 };
+		success = true;
+	} while (false);
```
asarium, line 509: Why the do-while?

code/model/modelread.cpp
```diff
@@ -3804,29 +3775,34 @@ void model_make_turret_matrix(polymodel *pm, polymodel_instance *pmi, model_subs
...
 
-	bsp_info * gun = &pm->submodel[turret->turret_gun_sobj];
-	bsp_info * base = &pm->submodel[turret->subobj_num];
+	// This should not happen
+	if ( turret->subobj_num == turret->turret_gun_sobj ) {
+		return 0;
+	}
+
+	auto base_sm = &pm->submodel[turret->subobj_num];
+//	auto gun_sm = &pm->submodel[turret->turret_gun_sobj];
```
asarium, line 3789: Not relevant anymore?

asarium (approved):
Nothing major so approving already.

Not sure if this should be merged while we are in a RC phase but we are still early in that so could be ok. I'll leave that up to you.

code/model/modelread.cpp
Goober5000: It's so that I could use `break`.  I initially just used a bunch of `return`s, but the problem is that you always need to call `unpause_parse`.

code/model/modelread.cpp
Goober5000: Relevant in the future.  I decided to add it because it's a very similar line, and all I need to do is uncomment it.  I would have left it uncommented but that triggers an unused variable warning.

code/model/modelread.cpp
asarium: Ohh, I see. Yes, that is a good reason to do this. Or use a bunch of `goto`s :wink:

PR #3025, opened 2020-12-26
Prevent Fire_turret from using null values

JohnAFernandez:
What is says on the tin.

### DIFF
--- a/code/network/multimsgs.cpp
+++ b/code/network/multimsgs.cpp
@@ -3302,21 +3302,21 @@ void send_turret_fired_packet( int ship_objnum, int subsys_index, int weapon_obj
 	ADD_SHORT( static_cast<short>(Weapons[objp->instance].weapon_info_index) );
 	ADD_SHORT( static_cast<short>(subsys_index) );
 
-	// since the two following values can accidentally dereference nullptrs, send 0 when null
-	constexpr float zero_value = 0.0f;
+	// where we might accidentally dereference nullptrs, send 0 instead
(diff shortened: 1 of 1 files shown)


notimaginative (approved):

MageKing17 (approved):

notimaginative (changes requested):
Sorry, I'm stupid. The flack_fired packet does this same thing so it will need the changes as well. I meant to mention that earlier and completely forgot.

JohnAFernandez:
Just glad you remembered and we don't have to do this twice.  I'll get that added fairly soon.

code/network/multimsgs.cpp
```diff
@@ -77,6 +77,9 @@
 
 // #define _MULTI_SUPER_WACKY_COMPRESSION
 
+// where we might accidentally dereference nullptrs, send 0 instead
+constexpr float ZERO_VALUE = 0.0f;
```
notimaginative, line 81: I know it's used in more than one function, but I'm not sure that's worth making it a global.

If it is a global though it needs to be `static`. And you might rename it to `ZERO_VALUE_float`, or something, to indicate it's type in case someone else needs an `int` version.

notimaginative (changes requested):

code/network/multimsgs.cpp
JohnAFernandez: I'll just move it back into the functions.

notimaginative (approved):

Issue #3027, opened 2020-12-27
Deprecation of drawMonochromeImage in lua script API causing CTDs, massive backwards compatibilty break


EatThePath:
Function calls to drawMonochromeImage get the error

> Lua Error: Deprecated function 'drawMonochromeImage' has been called that is not available in the targetted engine version. Check the documentation for a possible replacement.

on debug builds and a silent CTD on release builds. Axmessage, for one, is broken by this, and I'm sure many other scripts

I haven't bisected but I have done a personal build with simply taking the deprecated mark off the still-present code for the function restores things, so I am fairly confident that this happened with #3004, and as a scripting I'd probably be pinging @asarium either way.

EatThePath:
It turns out this breakage isn't going to be quite so prevalent felt as I thought, as it looks like most people who use axmessage in their mod configure it in a way that avoids the offending function call. Still a pretty big deal imo despite that.

EatThePath:
Digging result: the mod I encountered the crash on was targeting 20.1 in it's game_settings.tbl, and the other ones I tested on weren't. That deprecation is set up to only complain on 20.1 targeting or later. Shifting the mod's setting makes everything work again

So I guess the root issue is that lua deprecation messages are errors rather than warnings.

asarium:
This is working as designed. The deprecation messages are errors because the function simply does not exist when you specify a higher targetted engine version.

The actual issue here is that I specified a version that was already being used for nightlies so there were mods around which specified support for that version. I submitted a PR to fix the deprecation version and change it to 20.1.

PR #3028, opened 2020-12-27
Restore deprecated drawMonochromeImage

EatThePath:
this is a quick fix to #3027, which simply undoes the marking of the function as deprecated. This is a bandaid, I assume it would be preferable to have the function still deprecated with a log message/warning rather than an error/CTD, but given the severity of the issue caused I thought a quick fix may be appropriate.


asarium (changes requested):
This is fixed properly by #3035.

PR #3038, opened 2020-12-28
Fix CTD on death while support is trying to dock

SelfGamer:
This is an issue that only occurs on release builds. I suspect debug builds silently filter bad ID's earlier when evaluating collisions.
The issue is not directly the object (presumably the player) being null, rather it's id being invalid. This filters this, but obviously doesn't really solve the underlying issue, of the player's ship being target for docking even after dying.
For more context, the stack was:

ship_ship_check_collision
->check_for_docking_collision (the aip2->mode == AIM_DOCK branch)
-> dock_check_find_docked_object
bad read on (now) line 119, the function call to dock_evaluate_all_docked_objects

### DIFF
--- a/code/object/objectdock.cpp
+++ b/code/object/objectdock.cpp
@@ -105,8 +105,8 @@ bool dock_check_find_direct_docked_object(object *objp, object *other_objp)
 
 bool dock_check_find_docked_object(object *objp, object *other_objp)
 {
-	Assert(objp != NULL);
-	Assert(other_objp != NULL);
+	Assert(objp != nullptr && objp->signature > 0);
+	Assert(other_objp != nullptr && other_objp->signature > 0);
 
 	if (!(objp != nullptr && objp->signature > 0))
 		return false;
--- a/code/object/objectdock.cpp
+++ b/code/object/objectdock.cpp
@@ -105,8 +105,11 @@ bool dock_check_find_direct_docked_object(object *objp, object *other_objp)
 
 bool dock_check_find_docked_object(object *objp, object *other_objp)
 {
-	Assert(objp != nullptr && objp->signature > 0);
-	Assert(other_objp != nullptr && other_objp->signature > 0);
+	Assert(objp != nullptr);
+	Assert(objp->signature > 0);
+	Assert(other_objp != nullptr);
+	Assert(other_objp->signature > 0);
+
 
 	if (!(objp != nullptr && objp->signature > 0))
 		return false;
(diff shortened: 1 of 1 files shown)


The-E (approved):

code/object/objectdock.cpp
```diff
@@ -108,6 +108,11 @@ bool dock_check_find_docked_object(object *objp, object *other_objp)
 	Assert(objp != NULL);
 	Assert(other_objp != NULL);
 
+	if (!(objp != nullptr && objp->signature > 0))
```
asarium, line 111: Please also add `Assertion`s for the added condition here to catch this in debug builds as well.

asarium (changes requested):

code/object/objectdock.cpp
```diff
@@ -105,8 +105,13 @@ bool dock_check_find_direct_docked_object(object *objp, object *other_objp)
 
 bool dock_check_find_docked_object(object *objp, object *other_objp)
 {
-	Assert(objp != NULL);
-	Assert(other_objp != NULL);
+	Assert(objp != nullptr && objp->signature > 0);
+	Assert(other_objp != nullptr && other_objp->signature > 0);
```
asarium, line 109: FYI, it is better to have separate assertions so that we know exactly what part of the condition failed.

asarium (approved):

code/object/objectdock.cpp
```diff
@@ -105,8 +105,16 @@ bool dock_check_find_direct_docked_object(object *objp, object *other_objp)
...
 {
-	Assert(objp != NULL);
-	Assert(other_objp != NULL);
+	Assert(objp != nullptr);
+	Assert(objp->signature > 0);
+	Assert(other_objp != nullptr);
+	Assert(other_objp->signature > 0);
+
+
+	if (!(objp != nullptr && objp->signature > 0))
```
JohnAFernandez, line 114: I'm pretty sure this check is slightly wrong.  Valid signatures start *at* 0 because they store the index for whatever array that object is a part of, ships, weapons, etc. So this should be >= or > -1.

Net_signatures are "invalid" at 0, but that's another field and only matters during multi.

Welcome to the weirdness of FSO.

JohnAFernandez (changes requested):

code/object/objectdock.cpp
asarium: Signatures start at 1. 0 is invalid:
```c++
int Object_next_signature = 1;	//0 is bogus, start at 1
```

The `instance` starts at 0 and is an index into the relevant subtype array.

SelfGamer: This is adapted from the scripting API that has the object-is-valid function defined as 
``return (objp != NULL && objp->signature == sig && sig > 0);``
So if that is wrong, the scripting API needs a change too.

JohnAFernandez: Wth

PR #3056, opened 2020-12-30
Allow submodels to rotate around an arbitrary axis

Goober5000:
This is a rather substantial rewrite/refactor of the submodel rotation code.  Submodel rotation is now tracked via an axis-angle representation, which is used to calculate a matrix to represent the submodel orientation in the model instance.  The axis can be any unit vector, not limited to the standard X, Y, and Z vectors.

The submodel orientation is now used for position calculation in almost every place the submodel angles were previously used.  In conjunction with the simplification of KeldorKatarn's code, this saves a significant number of calculations for every movable submodel in every frame.  Hopefully this should lead to a measurable speedup in both rendering and collision detection.

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -1045,13 +1045,13 @@ int asteroid_check_collision(object *pasteroid, object *other_obj, vec3d *hitpos
 					smi->collision_checked = false;
 
 					// set angles for last frame (need to set to prev to get p0)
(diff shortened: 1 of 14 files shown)


Baezon (approved):
Pretty involved obviously, but if it tests well, then it looks alright to me. Let's leave it for a bit longer in case anyone else wants to chip in, but definitely don't want to leave this hanging forever.

PR #3058, opened 2020-12-30
weaponspew fixes

### LINKED ISSUE #3044: Weaponspew damage calculations don't account for puncture or a default beam behaviour
Puncture reduces hull damage by 75%, and beams ignore tabled damage factors unless otherwise specified in gamesettings.tbl. Weaponspew calculations don't reflect either.

Goober5000:
Add calculations for the `puncture` flag, and handle beam damage factors appropriately.

Fixes #3044.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -8417,12 +8417,14 @@ void weapon_spew_stats(WeaponSpewType type)
 
 			// doubled damage is handled strangely...
 			float hull_multiplier = 1.0f;
+			float shield_multiplier = 1.0f;
 			float subsys_multiplier = 1.0f;
 
 			// area effect?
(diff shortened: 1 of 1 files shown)


asarium (approved):

JohnAFernandez (commented):
All these mprintf's are probably not necessary anymore.  It's probably out of the scope of this PR, but we should probably get rid of them.

Goober5000:
All those mprintf's are intentional.  The purpose of -weaponspew is to print stats to the debug log.

JohnAFernandez:
Should have realized that.  Really sticking my foot in my mouth lately.

Goober5000:
No worries. :)  I've made plenty of mistakes like that myself.

PR #3060, opened 2020-12-31
Range aware secondary selection

EatThePath:
Makes AIs only select secondary weapons that are in range of the AI's target. This is in addition to the normal selection filtering that AIs do to try to pick appropriate missiles. Controlled by a AIProfiles flag, which is set up for later expansion to multiple modes; I wanted to implement a preference for longer or shorter ranges when multiple weapons are in range, I'm deffering that to later rather than letting this languish on my drives any more.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -58,6 +58,26 @@ int ai_path_type_match(char *p)
 	return -1;
 }
 
+
+const char* AI_secondary_range_aware_select_modes[] = {
+	"retail",
+	"aware",
+};
+
+int Num_ai_secondary_range_aware_select_modes = sizeof(AI_secondary_range_aware_select_modes) / sizeof(char*);
+
+int ai_secondary_range_select_mode_match(char* p)
+{
+	int i;
+	for (i = 0; i < Num_ai_secondary_range_aware_select_modes; i++)
+	{
+		if (!stricmp(AI_secondary_range_aware_select_modes[i], p))
+			return i;
+	}
+
+	return -1;
+}
+
 void parse_ai_profiles_tbl(const char *filename)
 {
 	int i;
@@ -546,6 +566,20 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$fixed ship-weapon collisions:", AI::Profile_Flags::Fixed_ship_weapon_collision);
 
+				//Intention is to expand this feature to include a preference for close or long range weapons
+				//hence using something besides a simple flag.
+				profile->ai_range_aware_secondary_select_mode = AI_RANGE_AWARE_SEC_SEL_MODE_RETAIL;
+				if (optional_string("$AI secondary range awareness:"))
+				{
+					stuff_string(buf, F_NAME, NAME_LENGTH);
+					int j = ai_secondary_range_select_mode_match(buf);
+					if (j >= 0) {
+						profile->ai_range_aware_secondary_select_mode = j;
+					}
+					else {
+						Warning(LOCATION, "Invalid ai secondary range awareness mode '%s' specified", buf);
+					}
+				}
 				// if we've been through once already and are at the same place, force a move
 				if (saved_Mp && (saved_Mp == Mp))
 				{
@@ -622,6 +656,7 @@ void ai_profile_t::reset()
     bay_arrive_speed_mult = 0;
     bay_depart_speed_mult = 0;
 	second_order_lead_predict_factor = 0;
+	ai_range_aware_secondary_select_mode = AI_RANGE_AWARE_SEC_SEL_MODE_RETAIL;
 
     for (int i = 0; i < NUM_SKILL_LEVELS; ++i) {
         max_incoming_asteroids[i] = 0;
--- a/code/ai/ai_profiles.h
+++ b/code/ai/ai_profiles.h
@@ -18,8 +18,8 @@
 #define	AI_PATH_MODE_NORMAL 0
 #define	AI_PATH_MODE_ALT1	1
 
-#define	AI_RANGE_SEC_SEL_MODE_RETAIL 0
-#define	AI_RANGE_SEC_SEL_MODE_AWARE 1
+#define	AI_RANGE_AWARE_SEC_SEL_MODE_RETAIL 0
+#define	AI_RANGE_AWARE_SEC_SEL_MODE_AWARE 1
 	
 #define MAX_AI_PROFILES	5
 
@@ -112,7 +112,7 @@ public:
 	float second_order_lead_predict_factor;
 	
 	//Controls if the AI is dumb enough to keep trying to use out-of-range secondaries. 
-	int ai_range_secondary_select_mode;
+	int ai_range_aware_secondary_select_mode;
 
     void reset();
 };
--- a/code/ai/ai_profiles.h
+++ b/code/ai/ai_profiles.h
@@ -17,6 +17,9 @@
 // AI Path types
 #define	AI_PATH_MODE_NORMAL 0
 #define	AI_PATH_MODE_ALT1	1
+
+#define	AI_RANGE_SEC_SEL_MODE_RETAIL 0
+#define	AI_RANGE_SEC_SEL_MODE_AWARE 1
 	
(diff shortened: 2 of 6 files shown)


JohnAFernandez (commented):
Pointers look fine, just make sure that none of those numbers used as indices are -1 (which may have already been done earlier in the codepath, I don't know) or we'll have rare and mysterious crashes later.

EatThePath:
> 
> 
> Pointers look fine, just make sure that none of those numbers used as indices are -1 (which may have already been done earlier in the codepath, I don't know) or we'll have rare and mysterious crashes later.

can a valid ship be assumed to have a valid AI index, or does that need to be safeguarded? I'd think it'd have to have an AI, and a target, to be firing a missile at all through this function

Goober5000:
> can a valid ship be assumed to have a valid AI index, or does that need to be safeguarded?

Almost certainly yes.  It wouldn't hurt to have an Assert, but I don't think it needs more than that.

code/ship/ship.cpp
```diff
@@ -13159,15 +13160,59 @@ int get_available_secondary_weapons(object *objp, int *outlist, int *outbanklist
...
+	
+	float target_range, weapon_range_max, weapon_range_min;
+	target_range = 0.0f;
 
+	if (The_mission.ai_profile->ai_range_secondary_select_mode!= AI_RANGE_SEC_SEL_MODE_RETAIL) {
+		vec3d our_position = objp->pos;
+		vec3d target_position;
+		object *target = &Objects[Ai_info[shipp->ai_index].target_objnum];
+		if (target->type == OBJ_SHIP) {
+			if (aip->targeted_subsys != NULL) {
```
Goober5000, line 13177: change `NULL` to `nullptr`

code/ai/ai_profiles.cpp
```diff
@@ -58,6 +58,26 @@ int ai_path_type_match(char *p)
 	return -1;
 }
 
+
+const char* AI_secondary_range_select_modes[] = {
+	"retail",
+	"aware",
```
Goober5000, line 64: Change this to `range_aware`

code/ai/ai_profiles.h
```diff
@@ -17,6 +17,9 @@
 // AI Path types
 #define	AI_PATH_MODE_NORMAL 0
 #define	AI_PATH_MODE_ALT1	1
+
+#define	AI_RANGE_SEC_SEL_MODE_RETAIL 0
+#define	AI_RANGE_SEC_SEL_MODE_AWARE 1
```
Goober5000, line 22: Change this to `AI_RANGE_SEC_SEL_MODE_RANGE_AWARE`

code/ship/ship.cpp
```diff
@@ -13159,15 +13160,59 @@ int get_available_secondary_weapons(object *objp, int *outlist, int *outbanklist
...
 	for (i=0; i<shipp->weapons.num_secondary_banks; i++)
 		if (shipp->weapons.secondary_bank_ammo[i]) {
-			outbanklist[count] = i;
-			outlist[count++] = shipp->weapons.secondary_bank_weapons[i];
+			if (The_mission.ai_profile->ai_range_secondary_select_mode != AI_RANGE_SEC_SEL_MODE_RETAIL) {
+				wepp = &Weapon_info[shipp->weapons.secondary_bank_weapons[i]];
+				weapon_range_min = wepp->WeaponMinRange;
+				weapon_range_max = wepp->weapon_range;
+				//If weapon range is not set in the weapon info, derive it
+				if (weapon_range_max >= 999999999.0f) {
```
Goober5000, line 13199: I think it would be better to pick a smaller number as the threshold.

Goober5000 (commented):
Some minor comments

code/ship/ship.cpp
EatThePath: If memory serves, this number was chosen because it is the default if no range is set by the table. I don't want to ignore a modder's tabled value if they have one, even if it's an absurd one.

Goober5000: Two reasons:

1) The original number was probably chosen arbitrarily.  There may be good reasons to override the range if it's very large but below the threshold.
2) The initialized number is "magic".  It isn't defined anywhere.  Suppose someone decides to change the initialized value but forgets to change it here too?

If there's a sensible threshold above which to derive the range, then do that.  Otherwise, change `999999999.0f` to a #define and use the #define in both places.

EatThePath: I'll make the changeover of the magic number to a define, but unless someone can give me a concrete and compelling situation where it'd be beneficial to do so I can't see my way to overriding excessively long but user set ranges.

Goober5000: I can't think of such a situation offhand, but I can't be sure that no situation exists.

EatThePath: I'm pretty opposed to implementing any sort of override on principal, as it's my general stance that the engine should let the modder do things that seem absurd on the surface unless they are blatantly invalid and destabilizing. And wthout a specific case I can't very well pick a reasonable override to implement anyway, so I'm going to leave it as it is.

Goober5000: Fair enough.

code/ai/ai_profiles.cpp
Goober5000: You marked this as resolved, but you didn't write any comments, nor do I see that the string has been changed.

Goober5000: Anyway, the reason for my suggestion was future-proofing.  If anyone adds more options in the future, "aware" might become ambiguous, or less clear.  It was for the same reason that I suggested changing `AI_RANGE_SEC_SEL_MODE_AWARE` to `AI_RANGE_SEC_SEL_MODE_RANGE_AWARE`.

code/ai/ai_profiles.cpp
```diff
@@ -699,7 +734,4 @@ void ai_profile_t::reset()
 	if (mod_supports_version(3, 7, 2)) {
 		flags.set(AI::Profile_Flags::Fix_ramming_stationary_targets_bug);
 	}
-	if (mod_supports_version(21, 4, 0)) {
-		flags.set(AI::Profile_Flags::Fixed_ship_weapon_collision);
-	}
```
Goober5000, line 704: Rebase onto master, or remove this from your changeset, so that this item from a different PR isn't reverted.

Goober5000 (changes requested):

code/ai/ai_profiles.cpp
EatThePath: I could have sworn it was originally AI_secondary_range_select_modes and I assumed you wanted more specificty in the early part of the name. I see no evidence that I actually made that change, but that misunderstanding is still why I made the changes I did. I hit resolve because I thought I had implemented that name change everywhere it applied, and had just somehow obscured it with git fumbling.

"aware" was intended to be future proof, representing the very basic level of range consideration vs the retail lack of any awareness. I guess arguably that could be a bit confusing in some cases, but I don't see how "range aware" is more clear and specific as a mode string for a setting named "range awareness" than "aware" by itself is.

code/ai/ai_profiles.cpp
EatThePath: this is actually debris from a bad rebase onto master that I overlooked. Will fix.

code/ai/ai_profiles.cpp
Goober5000: Ok.  I don't have a strong opinion on what the word should be, but I wanted to make sure future expansion possibilities were considered.  Also, I thought it was odd that the #define was changed but this wasn't.  (And if you aren't planning to change this string, you may want to consider changing the #define back.)

EatThePath: Reviewing the name changes I did make now suggests I was considerably more tired than I realized when I made them. A push with more consistent names and a fix to the other issue is coming once my local build finishes, which takes five or ten minutes at least.

code/ai/ai_profiles.cpp
EatThePath: This change has now been excised

Goober5000 (approved):
Ok, I think this is good.

PR #3061, opened 2020-12-31
Add dynamic action programming system

asarium:
This system adds the code necessary for letting content creators specify
complex events by specifying a simple, program-like sequence of
operations that will be executed by the engine when a certain event is
triggered.


Goober5000:
I've started reviewing this.  Sorry for the tardiness.

I don't really have any code comments, but I do have some design questions.  I haven't reviewed the whole thing yet though.
1) Can multiple actions be executed in parallel?  (This may already be true for all practical purposes, if an Action starts something "asynchronously" and immediately returns.)
2) Can actions or programs be specified in one place - let's say `actions.tbl` - and then referenced elsewhere?  This could allow us to define a FiringProgram consisting of a particle effect and a sound, and then specify FiringProgram by name for each turret on a ship.  That would remove the need to write out the program explicitly for every single turret.
3) Can programs be written independently of submodels?

asarium:
1. Depends on what you mean by "actions". The actions in a single program are executed sequentially but if you specify two programs for an action then the actions from those two programs will run "in parallel".
2. I have several changes building on this that will allow to build custom actions by compositing existing ones. This is essentially your proposed sub-program feature.
3. See 2.
4. The basic system allows not to have sub-objects attached to a program so yes. It technically also allows to have programs without an object but none of the actions support that mode yet.

code/model/modelread.cpp
```diff
@@ -5544,4 +5544,10 @@ void model_subsystem::reset()
 
     turret_max_bomb_ownage = 0;
     turret_max_target_ownage = 0;
+
+	beam_warmdown_program = actions::ProgramSet();
```
Goober5000, line 5548: I guess it will be necessary to add a hook like this wherever we want to add program functionality?  I'm just worried about `beam_warmup_program` and `beam_firing_program` and any number of other programs being added that would increase overhead.

code/weapon/weapons.cpp
```diff
@@ -5764,10 +5777,8 @@ int weapon_create( vec3d * pos, matrix * porient, int weapon_type, int parent_ob
 
 		objp->radius = model_get_radius(wip->model_num);
 
-		// if we intrinsic-rotate, make sure we have a model instance
-		if (model_get(wip->model_num)->flags & PM_FLAG_HAS_INTRINSIC_ROTATE) {
-			wp->model_instance_num = model_create_instance(false, wip->model_num);
-		}
+		// Always create an instance in case we need them
+		wp->model_instance_num = model_create_instance(false, wip->model_num);
```
Goober5000, line 5781: I don't like this; the vast majority of weapons are not going to need a model instance.  Can this be created conditionally?

Goober5000 (commented):
Overall, this looks very solid and a great addition to the code.

What about making `ProgramLocals` an interface, and making the current implementation merely one instance?  (Or make a simple superclass with only `waitTimestamp`.)  I can imagine a few programs that don't need a subobject, position, and orientation; and it's possible that someone in the future will need to add additional fields.

Also see the code comments.

code/weapon/weapons.cpp
asarium: The issue is that the action code only deals with model instances since that is the way to go now. Making this conditional would require a lot of special handling for something that (as far I know) only requires a bit more memory allocations.

code/model/modelread.cpp
asarium: That is correct. Every single potential program entry would need a separate program set member.

What overhead are you referring to and how would you improve it? I do not see either thing being a problem since unused program hooks will just be an empty vector.

asarium:
> What about making `ProgramLocals` an interface, and making the current implementation merely one instance? (Or make a simple superclass with only `waitTimestamp`.) I can imagine a few programs that don't need a subobject, position, and orientation; and it's possible that someone in the future will need to add additional fields

Making the locals an abstract class would require dynamic allocation which was something I wanted to avoid as much as possible. Also, the way local variables are handled will be changed with future changes I have already written so this will not be the final state.

code/model/modelread.cpp
Goober5000: I'm speaking very generally: overhead in execution time and overhead in memory usage.  The new hooks will require a certain amount of storage space and will need to be checked every frame.  Both of these will apply to every subsystem on every ship in the mission.

I don't actually know how consequential the overhead would be.  That being said, I have a hunch that nowadays a new MVP release will add incremental overhead that far exceeds the incremental overhead from this sort of thing.

code/weapon/weapons.cpp
Goober5000: I don't think it would require too much special handling.  All that's necessary is to check whether the instance is -1, something that's done in the relevant places (such as the rendering code) already.  Also, I just looked through the PR and there are only two places that need model instances, neither of which are going to be used for weapons.  If you revert these few lines, the rest of the PR will work with no further changes.

Yes, it just requires a bit of extra memory, but there are a *lot* of weapons used in every mission.  It adds up.

Goober5000:
> Making the locals an abstract class would require dynamic allocation which was something I wanted to avoid as much as possible. Also, the way local variables are handled will be changed with future changes I have already written so this will not be the final state.

Ok.

code/model/modelread.cpp
asarium: Memory overhead would only involve the actions parsed at startup which would be constant after that so there will be no significant allocations at runtime because of this system. If a program is triggered, new program locals are allocated for that but the memory required for that is pretty minimal. If no program is defined (as would be the case for every current mod) then only an empty vector would be allocated which takes up pretty much no memory.

For CPU overhead, if no program is defined then there is no additional overhead "per-frame" so existing mods would not cause any significant difference. For a running program, _once it is triggered_, we check every frame if things need to happen but that is only limited to the runtime of the program. If a program is triggered only as a response to an event happening then there is no overhead for that either.

code/weapon/weapons.cpp
asarium: Model instances are required for weapons if the action moves between subobjects of a weapon.

Something I could do relatively easily would be to check if we have a program and only create the instance in that case. However, that seems like a weird design to me since it would require changes for every program that is added which I would like to avoid to keep the code as generic as possible.

Would something along those lines be acceptable?

code/model/modelread.cpp
Goober5000: Well, that all sounds very good then.

code/weapon/weapons.cpp
Goober5000: My thinking is - how many programs are going to need to operate on the weapon's subobjects rather than the weapon as a hole?  The only uses I can think of are very esoteric, and I would rather cross that bridge when we come to it.

However, your suggestion is an acceptable alternative.

Goober5000 (approved):

Goober5000:
Good feature.  I'll let @asarium do the actual merging so that he can view the latest comments.

PR #3064, opened 2020-12-31
Feature/time to goal

z64555:
Revives PR #2098 

Adds a sexp to retrieve the estimated time of arrival of a ship to its ultimate destination in a waypoint/docking waypoint list.


z64555:
One thing that I didn't address was the desire/need to use milliseconds for the sexp version or seconds.  What is the recommended design regarding that?

code/parse/sexp.cpp
```diff
@@ -26817,6 +26830,10 @@ int query_operator_argument_type(int op, int argnum)
 			else
 				return OPF_SHIP_WING;
 
+		case OP_SECONDS_TO_GOAL:
```
z64555: Flagged by CI.  This cases a fall-through case when argnum != 0.  Briefly looking at the other cases doesn't give me a clear idea of whats supposed to happen here, obviously something needs to be returned as the else statement, but I don't know what that should be at this time.

code/parse/sexp.cpp
```diff
@@ -26817,6 +26835,10 @@ int query_operator_argument_type(int op, int argnum)
 			else
 				return OPF_SHIP_WING;
 
+		case OP_TIME_TO_GOAL:
+			if (argnum == 0)
```
z64555, line 26839: Flagged by CI. This causes a fall-through case when argnum != 0. Briefly looking at the other cases doesn't give me a clear idea of whats supposed to happen here, obviously something needs to be returned as the else statement, but I don't know what that should be at this time.

code/parse/sexp.h
```diff
@@ -304,6 +304,7 @@ class waypoint_list;
 #define	OP_MISSION_TIME_MSECS				(0x0007 | OP_CATEGORY_TIME | OP_NONCAMPAIGN_FLAG)	// Goober5000
 #define	OP_TIME_DOCKED						(0x0008 | OP_CATEGORY_TIME | OP_NONCAMPAIGN_FLAG)
 #define	OP_TIME_UNDOCKED					(0x0009 | OP_CATEGORY_TIME | OP_NONCAMPAIGN_FLAG)
+#define OP_TIME_TO_GOAL						(0x000a | OP_CATEGORY_TIME) // tcrayford
```
Goober5000, line 307: As you noticed on the other PR, yes, this needs `OP_NONCAMPAIGN_FLAG`.

code/parse/sexp.cpp
Goober5000: You're returning the parameter type for each parameter.  But this sexp only has *one* parameter, so there is no need to specifically check for the first one.  Just remove the if() here and always return `OPF_SHIP`.

code/parse/sexp.cpp
```diff
@@ -16804,6 +16805,17 @@ int sexp_query_orders (int n)
 	return hud_query_order_issued(order_to, order, target, timestamp, order_from, special);
 }
 
+int sexp_time_to_goal(int n)
+{
+	auto ship_entry = eval_ship(n);
+	if (!ship_entry || !ship_entry->shipp) {
```
Goober5000, line 16811: It's a little tricky to figure out what this particular sexp should return.  I think this makes the most sense:
- if `ship_entry` is null, return `SEXP_NAN`
- if the ship status is NOT_YET_ARRIVED, return `SEXP_CANT_EVAL`
- if it's EXITED, return `SEXP_NAN_FOREVER`.

As for the special cases:
- if the distance is < 1, the sexp should return 0.  It assumes the goal fulfillment is imminent.
- if there is no valid goal, the sexp should return `SEXP_CANT_EVAL`.  This will put directives on pause until the goal is valid and active.

code/ship/ship.cpp
```diff
@@ -15619,9 +15619,32 @@ SCP_string ship_return_orders(ship* sp)
...
+	if ( time >= 0 ) {
+		minutes = time/60;
+		seconds = time%60;
+		if ( minutes > 99 ) {
+			minutes = 99;
+			seconds = 99;
+		}
+		sprintf(outbuf, NOX("%02d:%02d"), minutes, seconds);
+	} else if ( time == -2) {
+        // -2 is used as a signal of "there is no valid goal right now"
```
Goober5000, line 15634: This comment about -2 isn't necessary here, I don't think.  More importantly, this `else if` block should be removed because a message needs to go into `outbuf` even in this situation.

Goober5000 (changes requested):
Ok, I checked all the comments on this PR and the old PR and I think I've covered every situation.

code/parse/sexp.cpp
z64555: As raised on discord, there was a case where `ship_return_seconds_to_goal()` could return -1 to mean "is not moving or can not move" which should also evaluated to `SEXP_CANT_EVAL`

Goober5000 (changes requested):
Looks good.  I noticed one more thing though.

Goober5000 (approved):

z64555:
Ok, that should at least do it from the code side of things.  Just need a run through a test mission to confirm operation then should be all set.

JohnAFernandez:
I think this fell under the radar while you were working on controls code.  Since it's a feature, I'm pushing it back to 21.4.

z64555:
Yeah, uh.  I fumbled the progress of this one by not knowing how to make a good test mission for it.  So if anybody is more capable of doing so, I ask that they make one. 🥺

Goober5000:
Here's a mission (rename .txt to .fs2):
[time-to-goal.txt](https://github.com/scp-fs2open/fs2open.github.com/files/6687878/time-to-goal.txt)

You can press 1 to display the time remaining.  The freighter has a dock goal and a waypoint goal.  It works perfectly for each, so I'm approving it.

PR #3067, opened 2021-01-01
Add full type coverage for return types and variables

asarium:
This introduces type parsing for all types used within the scripting
documentation. This improves the generated JSON output and also
validates new code via the unit test.


Goober5000 (approved):
A good idea, and I don't see any problems with the code.  (I just did a basic-level review though.)

PR #3072, opened 2021-01-02
Make sure that scripting changes the starting ammo as well as the weapon

JohnAFernandez:
Not tying this to 21.0 in case it somehow ends up being really complicated.  I don't want to hold up the process. This just makes sure that the rearmed ammo is changed when the scripting function in called.

EDIT: FYI, this is a pretty minor bug.


Goober5000 (approved):
I mean if #3059 is in the next release, this should be too.

Look good.

PR #3073, opened 2021-01-02
Fred comments

### LINKED ISSUE #2931: FR: Event comments
Request from Battuta, relayed due to account issue
> I'm doing some very complicated events and I just can't keep track of all the nested Whens and Wherefores.
> 
> It would be amazing to be able to enter short comments that you could see in the event window. Just lines of text that wouldn't be parsed by the mission as SEXPs. For example, like so
> 
> -WHEN
> --true
> --do nothing
> The above event is a placeholder, use it later.
> Below is the control event for Colossus turret ROF.
> -WHEN
> --is-ship-colossus(Colossus)
> --set-turret-rof(99999999)
>
> This would be HUGE for my short term progress.

Goober5000:
Since @GeneralBattuta begged and @naomimyselfandi made a scripting workaround, and since I've sometimes felt the need of this feature myself, I took a stab at making FRED comments.

These aren't the typical source-code style comments that are embedded in mission or table files; these are regular strings that are stored as mission data.  You might think of them as annotations.  Right-click on any node in the event tree and click on Edit Comment to add one.  The tree node will then show a different icon and the comment will be visible as a tooltip when you mouse over the node text.

This resolves #2931.


Goober5000:
Although this is a feature, I'd like it considered for 21.0 since it's a longstanding request and it mostly affects FRED, not FSO.

JohnAFernandez:
This can wait until after 21.0.... A brand new feature like this should be given time for testing before merging into what is a stable branch.

Goober5000:
Unfortunately, since qtFRED loading was disabled, the clang-tidy checks aren't running...
```
Run asarium/clang-tidy-action@v1
/usr/bin/docker exec  5e16f442e96ec8403c1bc20db6406931872234acdaa90d11c4aca3dfa4250426 sh -c "cat /etc/*release | grep ^ID"
Error: clang-tidy: 'mission/Editor.h' file not found (clang-diagnostic-error)
Error: clang-tidy: 'QtCore/QObject' file not found (clang-diagnostic-error)
Error: clang-tidy: 'mission/Editor.h' file not found (clang-diagnostic-error)
Error: clang-tidy: 'mission/Editor.h' file not found (clang-diagnostic-error)
Error: Found 4 clang-tidy issues
```

Baezon (approved):
As I said I'm not very familiar with FRED, but nothing terrible jumps out at me. Let's give it a little more time, just in case anyone else might want to say something.

PR #3076, opened 2021-01-02
preprocess alt names before mission start

Goober5000:
For compatibility purposes some mods want alt names to truncate at the hash symbol.  But we can't actually do that at mission load,
since we have to save them again in FRED.  So this PR processes the names just before the mission starts.

To further complicate things, some mods actually want the hash to be displayed.  So this PR introduces the double-hash as an escape sequence for a single hash.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -1314,6 +1314,8 @@ void game_post_level_init()
 
 	freespace_mission_load_stuff();
 
+	mission_process_alt_types();
+
 	// m!m Make hv.Player available in "On Mission Start" hook
 	if(Player_obj)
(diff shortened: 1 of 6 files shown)


freespace2/freespace.cpp
```diff
@@ -1272,6 +1272,8 @@ void freespace_mission_load_stuff()
 	}
 }
 
+extern void mission_process_alt_types();
```
asarium, line 1275: Please define this in the relevant header.

asarium (changes requested):

freespace2/freespace.cpp
Goober5000: Done.

asarium (approved):

PR #3079, opened 2021-01-03
convert float parsing to use strtof

Goober5000:
One small detail that was overlooked in PR #3021... KeldorKatarn's old code used `strtod` while the new code used `stuff_float`.  This wouldn't be a problem except that a couple MVP models use annoyingly formatted floats like 8.9e-08.  So I decided to convert `stuff_float` to use `strtof` and do a few other minor improvements at the same time.

### DIFF
--- a/code/gamesnd/gamesnd.cpp
+++ b/code/gamesnd/gamesnd.cpp
@@ -663,11 +663,9 @@ void parse_gamesnd_old(game_snd* gs)
 		{
 			int temp_min, temp_max;
 
-			ignore_gray_space();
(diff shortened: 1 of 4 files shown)


Goober5000:
This should be considered for 21.0 since model properties for a few MVP ships won't parse under the current code.

asarium:
I'm sorry but changing a core parsing function is not something I would consider to be necessary in the RC phase. Can you maybe fix or revert the specific code that was broken and once 21.0 is done we can merge this?

Goober5000:
Ok.  I'll make an interim PR.

asarium (approved):

PR #3080, opened 2021-01-03
Vectorize tech lists

JohnAFernandez:
This is a safer way of doing a dynamic size for tech entries that fixes a crash involving only the release executables.  As reported on discord.

It replaces a manually created array (if that is the correct name) with a few vectors.

### DIFF
--- a/code/menuui/techmenu.cpp
+++ b/code/menuui/techmenu.cpp
@@ -5,7 +5,7 @@
  * or otherwise commercially exploit the source or things you created based on the 
  * source.
  *
-*/ k to the state before "git rebase", run "git rebase --abort".
+*/
 
 
 
@@ -1208,20 +1208,21 @@ void techroom_lists_reset()
 			generic_anim_unload(&list_item.animation);
 		}
 
-	// now that we're sure all the bitmaps are released, clear the vectors.
-	Weapon_list.clear();
-	Weapons_loaded = false;
+		// now that we're sure all the bitmaps are released, clear the vectors.
+		Weapon_list.clear();
+		Weapons_loaded = false;
 
-	Intel_list.clear();
-	Intel_loaded = false;
+		Intel_list.clear();
+		Intel_loaded = false;
 
-	// clear the current list of items
-	Current_list.clear();
+		// clear the current list of items
+		Current_list.clear();
 
-	// free all models in use before clearing the ship list.
-	model_free_all();
-	Ship_list.clear();
-	Ships_loaded = false;
+		// free all models in use before clearing the ship list.
+		model_free_all();
+		Ship_list.clear();
+		Ships_loaded = false;
+	}
 }
 
 void techroom_close()
--- a/code/menuui/techmenu.cpp
+++ b/code/menuui/techmenu.cpp
@@ -1193,29 +1193,14 @@ void techroom_lists_reset()
 
 	// This can be cleared immediately because there are no anims or bitmaps associated.
 	Ship_list.clear();
-
 	Ships_loaded = false;
 
-	for (auto & list_item : Weapon_list) {
-		if (list_item.animation.num_frames != 0) {
-			generic_anim_unload(&list_item.animation);
-		}
+	// now that we're sure all the bitmaps are released, clear the vectors.
+	Weapon_list.clear();
+	Weapons_loaded = false;
 
-		// now that we're sure all the bitmaps are released, clear the vectors.
-		Weapon_list.clear();
-		Weapons_loaded = false;
-
-		Intel_list.clear();
-		Intel_loaded = false;
-
-		// clear the current list of items
-		Current_list->clear();
-
-		// free all models in use before clearing the ship list.
-		model_free_all();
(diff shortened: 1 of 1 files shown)


code/menuui/techmenu.cpp
```diff
@@ -5,7 +5,8 @@
  * or otherwise commercially exploit the source or things you created based on the 
  * source.
  *
-*/ 
+*/ k to the state before "git rebase", run "git rebase --abort".
+
```
notimaginative, line 9: ???

code/menuui/techmenu.cpp
```diff
@@ -235,14 +236,10 @@ typedef struct {
...
-static tech_list_entry *Weapon_list = nullptr;
-static int Weapon_list_size = 0;
-static tech_list_entry *Intel_list = nullptr;
-static int Intel_list_size = 0;
-static tech_list_entry *Current_list;								// points to currently valid display list
-static int Current_list_size = 0;
+SCP_vector<tech_list_entry> Ship_list;
+SCP_vector<tech_list_entry> Weapon_list;
+SCP_vector<tech_list_entry> Intel_list;
+SCP_vector<tech_list_entry> Current_list;								// A copy of the currently valid display list
```
notimaginative, line 242: These should be `static`. If something is defined globally, but not needed in other files, it should always be declared `static`.

notimaginative, line 242: `Current_list` should just be a ptr to the active list. Unless I'm missing something, there really isn't a reason to duplicate that info.

Additionally, bitmaps and animations are loaded into `Current_list` instead of the originals, and never released. Another reason it should just point to the appropriate list.

notimaginative (changes requested):

code/menuui/techmenu.cpp
JohnAFernandez: I think a few errant lines got in there.  I was in a rush, sorry

code/menuui/techmenu.cpp
Goober5000: Agreed, `Current_list` should be a pointer.

JohnAFernandez: All right, I'll try to remember this for the future.  Sorry us newer c++ programmers have a lot that we're not yet used to.

JohnAFernandez: Alright, made the change, built it and it runs.

Goober5000: Well, that's why we have code reviews.  Not every person knows about every code technique.

notimaginative: Most people don't seem to do it. Older C guys or embedded coders are the main users of those optimizations I think. I learned it from the far more experienced coders that worked on the initial icculus.org freespace2 port. Now it's just one of those easy things that I try to pass on to others :)

JohnAFernandez:
Ok, it's not telling me which nullptrs it wants, so I'll just change them all in this file.

notimaginative:
It says there's three on line 741 and three on line 750

code/menuui/techmenu.cpp
```diff
@@ -274,64 +271,63 @@ void techroom_init_desc(const char *src, int w)
...
+			intel_entry.bitmap = -1;
 		}
 	}
 }
 
 void techroom_select_new_entry()
 {
-	Assert(Current_list != NULL);
-	if (Current_list == NULL || Current_list_size <= 0) {
+	if (Current_list->empty() || Current_list->at(0).index == -1) {
```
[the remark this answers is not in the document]
JohnAFernandez: You're right, it isn't because that value is never set to -1, even during initialization.  I was trying to provide an early out to avoid the assert if the vector just had a default entry in it, but it should never be -1, anyway.

Now about the operator.  When you have a pointer to a vector, you can not use the [ ] operator to reference the items within the vector it is pointing to, but at() basically does the same thing.  I found that out while designing the rollback system.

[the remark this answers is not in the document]
Goober5000: Oh right, this is a pointer now.  That makes sense.

code/menuui/techmenu.cpp
```diff
@@ -1223,64 +1183,39 @@ void techroom_init()
...
-		for (i = 0; i < Weapon_list_size; i++) {
-			if (Weapon_list[i].animation.num_frames != 0) {
-				generic_anim_unload(&Weapon_list[i].animation);
-			}
-
-			if (Weapon_list[i].bitmap >= 0) {
-				bm_release(Weapon_list[i].bitmap);
-				Weapon_list[i].bitmap = -1;
-			}
+	for (auto & list_item : Weapon_list) {
```
notimaginative, line 1199: This block is wrong and shouldn't be here. The animations have already been unloaded, and the rest of the contents in this loop should be *outside* of it. `model_free_all()` and clearing of the ships list is done twice here.

notimaginative (changes requested):

notimaginative (approved):

notimaginative:
LGTM

PR #3083, opened 2021-01-04
Improve comment on vm_vec_delta_ang

Baezon:
The usage of this third vector is prone to cause some confusion given its vague and brief description. If you think of the rotation of a turret's base (a situation Goober has been in recently) where `v0` is its forward and `v1` is its desired angular position, feeding `uvec` with the turret's own up vector allows you to differentiate between a left or right turn.


Goober5000 (approved):
Looks good.

PR #3086, opened 2021-01-06
fix out of range error when AI targeting weapons

EatThePath:
assert on debugs, weapon_can_lock_on_ship() trying to index into the ships array with invalid index because it was called for ships targeting weapons.

Lock-on behavior when targeting weapons would be nice, but seems like it's not an intended feature at this time so just avoiding this logic rather than making new logic seemed to me to be the preferable path especially for the RC phase.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -9065,7 +9065,7 @@ void ai_chase()
 								}
 							}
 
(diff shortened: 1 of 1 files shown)


Baezon:
The fix for this one goes a bit deeper into `weapon_can_lock_on_ship()` itself, and I happened to have been simultaneously working on a fix for this.

PR #3087, opened 2021-01-06
Safety for lock restriction handling

Baezon:
The previous method only handled ships, even though the target could be a non-ship. This handles all targets properly now, and I also changed its name because "can lock on ship" is pretty vague so I thought it better to be explicit that this is about lock restrictions

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -4093,7 +4093,7 @@ void find_homing_object(object *weapon_objp, int num)
 				continue; 
 
 			homing_object_team = obj_team(objp);
-			bool can_lock_on_ship = objp->type != OBJ_SHIP || weapon_can_lock_on_ship(wip, objp->instance);
+			bool can_lock_on_ship = objp->type != OBJ_SHIP || weapon_allowed_lock_restriction_object(wip, objp);
 			bool can_attack = weapon_has_iff_restrictions(wip) || iff_x_attacks_y(wp->team, homing_object_team);
 			if (can_lock_on_ship && can_attack)
 			{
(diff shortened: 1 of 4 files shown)


code/weapon/weapons.cpp
```diff
@@ -4093,7 +4093,7 @@ void find_homing_object(object *weapon_objp, int num)
 				continue; 
 
 			homing_object_team = obj_team(objp);
-			bool can_lock_on_ship = objp->type != OBJ_SHIP || weapon_can_lock_on_ship(wip, objp->instance);
+			bool can_lock_on_ship = objp->type != OBJ_SHIP || weapon_allowed_lock_restriction_object(wip, objp);
```
asarium, line 4096: Since `OBJ_SHIP` is handled in the function, the check here is not needed anymore.

asarium (changes requested):
To be honest, I prefer `weapon_can_lock_on_ship` (or rather `weapon_can_lock_on_object`). Since it accurately describes what the function does. You current name is not very helpful to me as someone without extensive knowledge in this code.

Also, this looks like an additional feature or refactoring to me. Is that correct?
If it isn't a bug fix then merging this should be delayed until 21.0 is finished.

Baezon:
This is in fact a bugfix, using a generic `instance` to index into Ships regardless of type is obviously very bad and will cause crashes. This add no new functionality or behavior.

Whether you can lock onto a ship is a complicated question, perhaps I've been spending too much time in the multilock code, where soon various related functions to this one will be split off for generic use. These ask questions like _is a secondary fired from here in range to this target?_ _Is this target in the lock fov, or otherwise an unsuitable target (wrong team, stealthed etc)?_ and of course _does this weapon have lock restrictions which would disallow this target?_ Those all need to be separate functions (eventually) and all together answer _can this weapon lock on this ship?_ 

So to me, the lock restrictions is merely a small part of the question, and leaving it as "can lock on ship" is vague and most certainly not exhaustive (just because you answered true doesn't actually mean a lock may be valid for the myriad of previously mentioned reasons)! 

You're right that it's not really clear what's it doing on the tin, and I'd be open to a better name, but "can_lock_on_ship" is too simple and misleading in my opinion. The function is doing a very specific thing (only checking the lock restrictions), and it should try to reflect that in some way.

asarium:
Thanks for the explanation, that definitely clears things up for me. I added this to the 21.0 milestone so we can merge this once everything is ready.

Also thanks for the function name explanation. That helps me understand what this function does but the name is still not the best in my opinion. Maybe something like `weapon_target_object_satisfies_lock_restriction`?

Baezon:
A bit on the long side, but better verbose and clear than short and confusing.

EatThePath:
> A bit on the long side, but better verbose and clear than short and confusing.

weapon_validate_lock_conditions or something like that perhaps?

Baezon:
Well see I don't want to say something generic like "conditions" because it isn't really checking all the lock conditions, its checking the 'lock restrictions' a specific thing for weapons.

asarium (approved):

JohnAFernandez (approved):
Looks good to me too, new names communicate the adjusted functionality clearly.

PR #3093, opened 2021-01-08
Fix a crash with scripted targeting brackets

### LINKED ISSUE #3091: gr.drawTargetingBrackets in lua API crashes when called on some valid weapons
If a weapon is a secondary and uses a laser graphic instead of a model, gr.drawTargetingBrackets called on that weapon will crash the game. These weapons are otherwise valid and produce no warnings or errors, and if bombs they can even be targetted safely by players, though they display nothing in their target preview box.

Perhaps worse, it crashes in the same way if called on a laser primary, too. That's an unconventional use case for the function, but not an inconceivable one. 

This crash appears to be because drawTargetingBrackets, at 901-905 of graphics.cpp, only checks if a weapon is a secondary, which in itself is the wrong thing to check, and only in an assert, not any kind of escape condition. It then calls model_find_2d_bound_min which has no model validity checks itself and starts trying to get and use a pointer to model number -1.

Putting a safety escape condition in drawTargetingBrackets seems easy enough, but also like the wrong solution. Drawing boxes around lasers seems both desirable and doable, just not trivially doable from where I'm sitting.

Test condition: Get release version 2 of fso://open/etp_bugtest, or grab 
[boxcrash.zip](https://github.com/scp-fs2open/fs2open.github.com/files/5784376/boxcrash.zip), and play the included mission. Pull the trigger or wait a few seconds. Boom.

Baezon:
Implements a catch for a -1 `modelnum` that the normal hud bracketing code uses, but the scripted version didn't. Still not great quality of life here, it would be nice if this actually looked at the laser's radii or length or something, but it works without breaking anything and the scripters can always mess with the padding if they need to.

Closes #3091

### DIFF
--- a/code/scripting/api/libs/graphics.cpp
+++ b/code/scripting/api/libs/graphics.cpp
@@ -899,9 +899,16 @@ ADE_FUNC(drawTargetingBrackets, l_Graphics, "object Object, [boolean draw=true, 
 			}
 			break;
 		case OBJ_WEAPON:
-			Assert(Weapon_info[Weapons[targetp->instance].weapon_info_index].subtype == WP_MISSILE);
 			modelnum = Weapon_info[Weapons[targetp->instance].weapon_info_index].model_num;
-			bound_rc = model_find_2d_bound_min( modelnum, &targetp->orient, &targetp->pos,&x1,&y1,&x2,&y2 );
+			if (modelnum != -1)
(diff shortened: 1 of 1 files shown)


asarium (approved):
The relevant code should probably be extracted from `HudGaugeBrackets` and then reused here to avoid these issues with duplicated code but that is not in the scope of this fix...

Issue #360, opened 2015-09-27
AI Shield Management in TvT Multiplayer questionable


z64555:
Stumbled on this while going through nixprime's shield management. Currently, AI have a difficulty factor applied to how frequently they manage their shields. This factor is only applied if the Player's team is hostile to the attacked ship's team.

function in question is ai_manage_shield() in aicode.cpp:12329.

My concern is how ai_manage_shield behaves in a Team vs. Team multiplayer match, whether all teams (with AI opponents) use the same shield management frequency or not.

JohnAFernandez:
Yeah, it really looks like there's no way that it balances during TvT, let me start looking at ways to correct this.

PR #3135, opened 2021-01-12
Add TvT logic for shield recharge

JohnAFernandez:
To start handling #360

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -12517,11 +12517,42 @@ void ai_manage_shield(object *objp, ai_info *aip)
 
 		//	Scale time until next manage shield based on Skill_level.
 		//	Ships on player's team are treated as if Skill_level is average.
-		if (iff_x_attacks_y(Player_ship->team, Ships[objp->instance].team))
+		if (MULTI_TEAM)
 		{
-			delay = aip->ai_shield_manage_delay;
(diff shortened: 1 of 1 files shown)


Goober5000:
This bit of code is going to get wonky because the typical way the game determines whether a ship is an "enemy" falls over in TvT.

The IFF code wasn't in retail; it was a FSO feature that I added several years ago.  Most IFF checks for hostile ships, such as the targeting control and the enemy locator, don't actually look for ships that aren't on the player's team.  They look for ships that the player's team attacks.

So for shield recharge, any team that the player attacks will have one bit of logic, and any team that the player's team does *not* attack will have the other bit of logic.  This means that Unknown will charge using the same rules as Friendly, if the player is Friendly and we are using the retail teams.

Now in TvT situations you end up with a scenario where *every* team is both on a team that *a* player attacks, and also on a team that *a* player doesn't attack.  The system of equations has no solution.

For TvT missions I would therefore suggest that every player team uses the "player" recharge and every team that has no players on it uses the non-player recharge.  I would also recommend moving this logic into a separate function defined in iff_defs because there are probably going to be multiple TvT situations that need to be handled similarly.

JohnAFernandez:
Well that's good to know....

JohnAFernandez:
This is not the right approach for this.

PR #3139, opened 2021-01-13
More safety checks for the HUD_config struct

### LINKED ISSUE #3136: Crash from unknown radar range
I crashed when saving hud color settings during a mission. I got an error about "unknown radar range of -1" in radarsetup.cpp line 479. I haven't reproduced yet.

When I restarted FSO (in debug mode), I got a warning about the unknown radar range of -1, and got a warning that radar range was set to default. Brain is dumb and I don't recall if I crashed or not. Debug log [here](https://fsnebula.org/log/5ffe14729ec55980ffa8d59b).

Attaching a minidump from the initial crash and pilot files from before the start in debug: [radarrangecrash.zip](https://github.com/scp-fs2open/fs2open.github.com/files/5804960/radarrangecrash.zip).

z64555:
* Adds more safety checks for PLR and CSG reading for the HUD_config structure.
* Adds a #define for the HC_color array size
* Doxy's the HUD_config struct

Looking into causes of #3136 revealed the HUD_config struct is somewhat haphazard in initialization.  It makes the assumption that the struct will be properly initialized by a PLR or CSG read, or by creation of a new player.  No where else is the singleton struct init'd, making it a potential issue for data races should anything try to operate on non-init'd data.

@JohnAFernandez Is still looking into the cause of the data corruption due to a bad exit of the HUD Config menu.  This PR doesn't completely fix #3136 but it should at least prevent crashes due to bad data.

### DIFF
--- a/code/pilotfile/csg.cpp
+++ b/code/pilotfile/csg.cpp
@@ -966,6 +966,7 @@ void pilotfile::csg_write_redalert()
 void pilotfile::csg_read_hud()
 {
 	int idx;
+	int strikes = 0;
 
 	// flags
 	HUD_config.show_flags = cfread_int(cfp);
@@ -986,10 +987,19 @@ void pilotfile::csg_read_hud()
 
 	// basic colors
 	HUD_config.main_color = cfread_int(cfp);
-	HUD_color_alpha = cfread_int(cfp);
+	if (HUD_config.main_color < 0 || HUD_config.main_color > HUD_COLOR_SIZE) {
+		Warning(LOCATION, "Campaign file has invalid main color selection %i, setting to default.\n", HUD_config.main_color);
+		HUD_config.main_color = HUD_COLOR_GREEN;
+	}
 
-	if (HUD_color_alpha < HUD_COLOR_ALPHA_USER_MIN) {
+	HUD_color_alpha = cfread_int(cfp);
+	if (HUD_color_alpha < HUD_COLOR_ALPHA_USER_MIN || HUD_color_alpha > HUD_COLOR_ALPHA_USER_MAX) {
+		Warning(LOCATION, "Campaign file has invalid alpha color %i, setting to default.\n", HUD_color_alpha);
 		HUD_color_alpha = HUD_COLOR_ALPHA_DEFAULT;
+	}
+
+	if (strikes == 3) {
+		Warning(LOCATION, "Campaign file has too many hud config errors, and is likely corrupted. Please verify and save your settings in the hud config menu.");
 	}
 
 	hud_config_record_color(HUD_config.main_color);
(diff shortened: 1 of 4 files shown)


code/pilotfile/csg.cpp
```diff
@@ -986,12 +987,21 @@ void pilotfile::csg_read_hud()
...
 
-	if (HUD_color_alpha < HUD_COLOR_ALPHA_USER_MIN) {
+	HUD_color_alpha = cfread_int(cfp);
+	if (HUD_color_alpha < HUD_COLOR_ALPHA_USER_MIN || HUD_color_alpha > HUD_COLOR_ALPHA_USER_MAX) {
+		Warning(LOCATION, "Campaign file has invalid alpha color %i, setting to default.\n", HUD_color_alpha);
 		HUD_color_alpha = HUD_COLOR_ALPHA_DEFAULT;
 	}
 
+	if (strikes == 3) {
+		Warning(LOCATION, "Campaign file has too many hud config errors, and is likely corrupted. Please verify and save your settings in the hud config menu.");
```
asarium, line 1002: Maybe this could be a `ReleaseWarning`? That seems to be something that would be good to know even in release builds.

code/pilotfile/csg.cpp
```diff
@@ -966,6 +966,7 @@ void pilotfile::csg_write_redalert()
 void pilotfile::csg_read_hud()
 {
 	int idx;
+	int strikes = 0;
```
asarium, line 969: This is not used in this function but should probably be incremented with every error correction (like in the other function).

code/pilotfile/plr.cpp
```diff
@@ -187,13 +188,26 @@ void pilotfile::plr_read_hud()
...
-	if (HUD_color_alpha < HUD_COLOR_ALPHA_USER_MIN) {
+	HUD_color_alpha = handler->readInt("color_alpha");
+	if (HUD_color_alpha < HUD_COLOR_ALPHA_USER_MIN || HUD_color_alpha > HUD_COLOR_ALPHA_USER_MAX) {
+		Warning(LOCATION, "Player file has invalid alpha color %i, setting to default.\n", HUD_color_alpha);
 		HUD_color_alpha = HUD_COLOR_ALPHA_DEFAULT;
+		strikes++;
+	}
+
+	if (strikes == 3) {
+		Warning(LOCATION, "Player file has too many hud config errors, and is likely corrupted. Please verify and save your settings in the hud config menu.");
```
asarium, line 210: Same as above, could be a `ReleaseWarning`.

asarium (changes requested):

Goober5000 (changes requested):

Goober5000 (approved):

asarium (approved):

PR #3142, opened 2021-01-14
Switch the correct ammo variable.

JohnAFernandez:
The previous pr was supposed to fix what ammo capacity the ship was reloaded to, but instead it was changing how much ammo was in the magazine.


Goober5000 (approved):
`>.>`

PR #3144, opened 2021-01-14
reimplement look_at

Goober5000:
This PR re-implements the look_at feature, which has been broken for many years.  A look-at rotation is a type of intrinsic_rotation, just like dumb-rotations. The same code handles both; the only difference is whether submodel_rotate or submodel_look_at is called to apply the actual movement.

The previous look_at parsing code has been optimized by performing the name-to-submodel resolution when the model is actually parsed, instead of when the first look-at rotation is called.  This avoids the need to store a name in the submodel struct and also avoids cluttering the look_at function.

This PR is essentially bookkeeping, except for the submodel_look_at method.  This method takes advantage of the submodel rotation upgrade and calculates angles in almost the same way as the new turret rotation code.  It's much more elegant, much less hackish, and works for all detail levels.

~~This PR depends on #3056 and will remain in draft status until that is merged.~~

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -126,7 +126,7 @@ int Num_dock_type_names = sizeof(Dock_type_names) / sizeof(flag_def_list);
 SCP_vector<glow_point_bank_override> glowpoint_bank_overrides;
 
 
-// Goober5000 - reimplementation of Bobboau's $dumb_rotation feature in a way that works with the rest of the model instance system
+// Goober5000 - reimplementation of Bobboau's $dumb_rotation and $look_at features in a way that works with the rest of the model instance system
 // note: since these data types are only ever used in this file, they don't need to be in model.h
 
 class intrinsic_rotation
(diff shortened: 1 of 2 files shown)


code/model/modelread.cpp
```diff
@@ -3548,115 +3617,61 @@ void submodel_stepped_rotate(model_subsystem *psub, submodel_instance *smi)
...
-	float dot=vm_vec_dot(&l,&mp);
-	if (dot>=0.0f) {
-		*a = asinf(c.a1d[axis]);
-	} else {
-		*a = PI-asinf(c.a1d[axis]);
+		// ensure the offset is in the proper range (see submodel_canonicalize)
+		while (sm->look_at_offset > PI2)
+			sm->look_at_offset -= PI2;
+		while (sm->look_at_offset < 0.0f)
+			sm->look_at_offset += PI2;
```
Baezon, line 3664: Why not `modf()`?

Baezon (approved):
LGTM, only one very minor comment, that really isn't an issue or anything.

code/model/modelread.cpp
Goober5000: I don't think that would actually work.

PR #3145, opened 2021-01-15
Stop including tube lights between anti-fighter beam pulses

EatThePath:
Fixing tube lighting revealed a bug were the light remained on between the pulses of a anti-fighter beam cycle. This adds a check in  `beam_move_all_post()` to avoid that.

also adds a comment to `beam_type_d_get_status()` about the returns of that function, because I used that here and that comment would have saved me some time had it been there already.

### DIFF
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -168,6 +168,8 @@ void beam_type_c_move(beam *b);
 
 // type D functions
 void beam_type_d_move(beam *b);
+// stuffs the index of the current pulse in shot_index
+// stuffs 0 in fire_wait if the beam is active, 1 if it is between pulses
 void beam_type_d_get_status(beam *b, int *shot_index, int *fire_wait);
 
 // type e functions
@@ -1037,6 +1039,7 @@ void beam_move_all_post()
 	beam *next_one;	
 	int bf_status;	
 	beam_weapon_info *bwi;
+	int type_d_index, type_d_wait = 0;
 
 	// traverse through all active beams
 	moveup = GET_FIRST(&Beam_used_list);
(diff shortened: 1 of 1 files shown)


code/weapon/beam.cpp
```diff
@@ -1037,6 +1039,7 @@ void beam_move_all_post()
 	beam *next_one;	
 	int bf_status;	
 	beam_weapon_info *bwi;
+	int type_d_index, type_d_wait = 0;
```
asarium, line 1042: Variables should be defined as close as possible to the usage.

code/weapon/beam.cpp
```diff
@@ -1136,8 +1139,16 @@ void beam_move_all_post()
...
-		if(moveup->objp != NULL)
-			beam_add_light(moveup, OBJ_INDEX(moveup->objp), 1, NULL);
+		if (moveup->objp != NULL) {
+			//assume we are in the waiting phases of a type D beam
+			type_d_wait = 0;
+			//test that assumption
+			if (moveup->type == BEAM_TYPE_D)
+				beam_type_d_get_status(moveup, &type_d_index, &type_d_wait);
+			//if it remains true, create a tube light.
+			if (type_d_wait == 0)
```
asarium, line 1149: This code is a bit hard to understand since it mixes terminology (e.g. type_d is used even for branches where the beam is not a Type D.

At least for me, it would be easier to understand if the two branches (Type D and non-D) were handled completely separately.

asarium (changes requested):
I do not see this as a major bug so this can wait until after we have the final release in a few days.

EatThePath:
> 
> 
> I do not see this as a major bug so this can wait until after we have the final release in a few days.

Personally I find it pretty unsightly in places and would hate for the fix of tube lights to be marred by it. https://www.youtube.com/watch?v=Y_TY3Cne6XE shows it very prominently early with the Ravana's beam fire. But if that is not persuasive I will contest no further.

asarium:
Merging this now would mean that we would need an RC5 and if we keep doing that for non-breaking, cosmetic fixes we will never get a release done.

Unless there is a regression or a game breaking bug of some kind, no changes should be merged until the 21.0 final is done.

code/weapon/beam.cpp
EatThePath: would it be clearing if it stayed in the current logic but the variable was renamed "between_type_d_pulses" or the like? To me it would then make plenty of sense that you aren't between pulses in, but if that's still confusing to others it should be an easy structural change

code/weapon/beam.cpp
EatThePath: I was under a mistaken impression that there were reasons to avoid declaring variables inside a loop. That impression has been dispelled, I'll move them

code/weapon/beam.cpp
asarium: My point here is that the code seems to handle different kinds of beam types but you put in a Type D specific check in the middle. For me while reading the code, it is much clearer if you would write something like `if (<type_d> ) <check if between pulses and add light> else <add light>`.

That would make the code flow apparent immediately without having to go back to `type_d_wait` to see that its default value is `0`.

asarium (approved):

PR #3149, opened 2021-01-17
fix a rare crash in the tech room

Goober5000:
The tech room loop checked that all the lines could fit in the available screen real estate, but it didn't check to see that the number of lines was less than the array capacity.  Normally this never happened, but certain mods which use smaller fonts will overflow the limit.  This adds a proper bounds check and also increases the bounds so that more of the smaller lines can be used.

### DIFF
--- a/code/menuui/techmenu.cpp
+++ b/code/menuui/techmenu.cpp
@@ -43,7 +43,7 @@
 
 #define NUM_BUTTONS	16
 #define NUM_TABS		3
-#define LIST_BUTTONS_MAX	42
+#define LIST_BUTTONS_MAX	50
 
 #define SHIPS_DATA_MODE		(1<<0)
 #define WEAPONS_DATA_MODE	(1<<1)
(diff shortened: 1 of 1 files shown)


code/menuui/techmenu.cpp
```diff
@@ -43,7 +43,7 @@
 
 #define NUM_BUTTONS	16
 #define NUM_TABS		3
-#define LIST_BUTTONS_MAX	42
+#define LIST_BUTTONS_MAX	50
```
JohnAFernandez, line 46: If there isn't a reason to keep the number low, perhaps we shouldn't stop at 50? A larger number might be more future proof.

Goober5000: The most future-proof way would be to change the whole thing to a vector.  42 was sufficient for more than 20 years until a few mods started using smaller fonts.  50 will be sufficient unless the UI completely changes in such a way that the screen real estate vastly expands.  If that ever occurs, changing to a vector will most likely be part of it anyway.

JohnAFernandez (approved):

asarium:
Was merging this really necessary? Now we either have to do RC5 or revert this commit.

Goober5000:
See my comment on #3150.  I think it was appropriate to merge and I also don't think it requires an RC5.

PR #3150, opened 2021-01-17
Revert "fix a rare crash in the tech room"

asarium:
Reverts scp-fs2open/fs2open.github.com#3149

Since the original PR was just for a rare crash we do not need to do that before the final release. This will bring the master branch back to the state of RC4.

### DIFF
--- a/code/menuui/techmenu.cpp
+++ b/code/menuui/techmenu.cpp
@@ -43,7 +43,7 @@
 
 #define NUM_BUTTONS	16
 #define NUM_TABS		3
-#define LIST_BUTTONS_MAX	50
+#define LIST_BUTTONS_MAX	42
 
 #define SHIPS_DATA_MODE		(1<<0)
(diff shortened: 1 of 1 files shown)


Baezon (approved):

Goober5000 (approved):
I understand the restriction against merging changes after RC4, but this is a bona-fide crash-to-desktop.  Aren't those appropriate to fix?

I don't know *how* rare this is.  Right now it only affects Inferno that I know of, but it would presumably affect any mod with a lot of ships and a small font.

I also don't think this merits RC5.  It's a very clear and straightforward crash with a very clear and straightforward fix.

That being said, I am in favor of the revert if it means 21.0 will be out the door faster.

asarium:
My point is that even the most innocent changes might have something that breaks everything. The point of a release candidate is to have a state which will be release exactly as it is without any additional changes. Otherwise we just throw out untested code and declare it "stable" even though it has not even be tested.

PR #3151, opened 2021-01-17
Convert some Lua hooks to the new style to add documentation

asarium:
This takes a few of the existing hooks and converts them to the new
style of hook definition which can include more extensive documentation
and enumeration of supported hook variables.

### DIFF
--- a/code/object/collideshipship.cpp
+++ b/code/object/collideshipship.cpp
@@ -23,8 +23,7 @@
 #include "object/object.h"
 #include "object/objectdock.h"
 #include "object/objectshield.h"
-#include "scripting/scripting.h"
-#include "scripting/api/objs/vecmath.h"
+#include "scripting/global_hooks.h"
 #include "playerman/player.h"
 #include "render/3d.h"			// needed for View_position, which is used when playing 3d sound
 #include "ship/ship.h"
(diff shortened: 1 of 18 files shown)


code/object/collideshipship.cpp
```diff
@@ -1357,18 +1362,21 @@ int collide_ship_ship( obj_pair * pair )
...
+				scripting::hooks::OnShipCollision->run(scripting::hook_param_list(scripting::hook_param("Self", 'o', A),
+														   scripting::hook_param("Object", 'o', B),
+														   scripting::hook_param("Ship", 'o', A),
+														   scripting::hook_param("ShipB", 'o', B),
+														   scripting::hook_param("Hitpos", 'o', world_hit_pos)),
+					A);
 			}
 			if((b_override && !a_override) || (!b_override && !a_override))
 			{
-				// Yes, this should be reversed.
```
Goober5000, line 1367: I would leave the comment in the code, or even make it more explicit that the swapping of A and B is intentional.

Goober5000 (changes requested):
Added a note about a comment.

Also, if you're going to be messing with the `On Death` hook anyway, I suggest deprecating that hook and making new individual `On Ship Death` and `On Asteroid Death` hooks.

code/object/collideshipship.cpp
asarium: Good point. Added the comment again.

asarium:
Also a good point about the deprecation. However, that will require some more work to integrate it into the existing deprecation system so that is out of scope for this PR.

Goober5000 (approved):
LGTM

TRBlount:
Is this still in Draft status?

z64555:
Deferring to 21.6 due to lack of feedback.

TRBlount:
If this is still in Draft status, then it will be delayed until 22.2

z64555:
Closing upon request due to lack of progress.  Please resubmit when you get back to it.

PR #3153, opened 2021-01-17
Improve graphics API to reduce return values

asarium:
This is an effort to clean up the graphics API a bit and reduce return
values that are not really needed.

The reason behind this is to reduce the amount of synchronization points
necessary when implementing a threaded rendering backend. The easy way
to do threading is to just dispatch graphics calls to a separate thread.
However, if there are return values involved, there needs to be
synchronizations which defeats the point of multi-threading.


Goober5000:
After consultation with @asarium, closing this PR.

PR #3154, opened 2021-01-17
fix a rare crash in the tech room

Goober5000:
The tech room loop checked that all the lines could fit in the available screen real estate, but it didn't check to see that the number of lines was less than the array capacity.  Normally this never happened, but certain mods which use smaller fonts will overflow the limit.  This adds a proper bounds check and also increases the bounds so that more of the smaller lines can be used.

This is a redo of #3149 to be merged after the 21.0 release.

### DIFF
--- a/code/menuui/techmenu.cpp
+++ b/code/menuui/techmenu.cpp
@@ -43,7 +43,7 @@
 
 #define NUM_BUTTONS	16
 #define NUM_TABS		3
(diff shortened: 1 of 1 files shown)


Goober5000:
Per @TRBlount this is now mergeable for RC5.

asarium (approved):

PR #3159, opened 2021-01-21
Fix a bug with previous_bank and ship class swapping

Baezon:
Previous bank also needs to be reset when swapping or this will cause crashes with scripts using it as a conditional.

As much I would love to PR all my features, I do think this should be in stable as JAD currently will very reliably crash on current builds, and I suspect this is also the cause of another similar bug reported in WoD recently.


JohnAFernandez:
It said it was failing at python installation, so I reran the checks.

Goober5000 (approved):

Goober5000:
Per @TRBlount this is now mergeable for RC5.

asarium (approved):

PR #3164, opened 2021-01-23
Limit instances of Client Collisions to Avoid Instant Death

JohnAFernandez:
Often in Multi, a client colliding with another ship, especially a capship, will result in instant death.  By limiting the maximum frequency of damage to client ships from ship-ship collisions to once every .2 seconds, instant death can be avoided in most cases.

Before
https://youtu.be/bUfe_dEUSeE

After
https://youtu.be/tFzUu3WlXxc

### DIFF
--- a/code/network/multi_observer.cpp
+++ b/code/network/multi_observer.cpp
@@ -60,6 +60,7 @@ void multi_obs_create_player(int player_num,char *name,net_addr *addr,player *pl
 	Net_players[player_num].reliable_socket = PSNET_INVALID_SOCKET;
 	Net_players[player_num].s_info.kick_timestamp = -1;
 	Net_players[player_num].s_info.voice_token_timestamp = -1;
+	Net_players[player_num].s_info.player_collision_timestamp = -1;
 	Net_players[player_num].s_info.tracker_security_last = -1;
 	Net_players[player_num].s_info.target_objnum = -1;
 	Net_players[player_num].s_info.accum_buttons = 0;
--- a/code/network/multi_observer.cpp
+++ b/code/network/multi_observer.cpp
@@ -60,7 +60,7 @@ void multi_obs_create_player(int player_num,char *name,net_addr *addr,player *pl
 	Net_players[player_num].reliable_socket = PSNET_INVALID_SOCKET;
 	Net_players[player_num].s_info.kick_timestamp = -1;
 	Net_players[player_num].s_info.voice_token_timestamp = -1;
-	Net_players[player_num].s_info.player_collision_timestamp = -1;
+	Net_players[player_num].s_info.player_collision_timestamp = timestamp(0);
 	Net_players[player_num].s_info.tracker_security_last = -1;
 	Net_players[player_num].s_info.target_objnum = -1;
 	Net_players[player_num].s_info.accum_buttons = 0;
--- a/code/network/multiui.cpp
+++ b/code/network/multiui.cpp
@@ -2799,6 +2799,7 @@ void multi_sg_init_gamenet()
 	// assign my player struct and other data	
 	Net_player->flags |= (NETINFO_FLAG_CONNECTED | NETINFO_FLAG_DO_NETWORKING);
 	Net_player->s_info.voice_token_timestamp = -1;	
+	Net_player->s_info.player_collision_timestamp = -1;
 
 	// if we're supposed to flush our cache directory, do so now
 	if(Net_player->p_info.options.flags & MLO_FLAG_FLUSH_CACHE){
--- a/code/network/multiui.cpp
+++ b/code/network/multiui.cpp
@@ -2799,7 +2799,7 @@ void multi_sg_init_gamenet()
 	// assign my player struct and other data	
 	Net_player->flags |= (NETINFO_FLAG_CONNECTED | NETINFO_FLAG_DO_NETWORKING);
 	Net_player->s_info.voice_token_timestamp = -1;	
-	Net_player->s_info.player_collision_timestamp = -1;
+	Net_player->s_info.player_collision_timestamp = timestamp(0);
 
 	// if we're supposed to flush our cache directory, do so now
 	if(Net_player->p_info.options.flags & MLO_FLAG_FLUSH_CACHE){
--- a/code/network/multiutil.cpp
+++ b/code/network/multiutil.cpp
@@ -800,7 +800,7 @@ void multi_create_player( int net_player_num, player *pl, const char* name, net_
 	Net_players[net_player_num].reliable_socket = PSNET_INVALID_SOCKET;
 	Net_players[net_player_num].s_info.kick_timestamp = -1;
 	Net_players[net_player_num].s_info.voice_token_timestamp = -1;
-	Net_players[net_player_num].s_info.player_collision_timestamp = -1;
+	Net_players[net_player_num].s_info.player_collision_timestamp = timestamp(0);
 	Net_players[net_player_num].s_info.tracker_security_last = -1;
 	Net_players[net_player_num].s_info.target_objnum = -1;
 	Net_players[net_player_num].s_info.accum_buttons = 0;
--- a/code/network/multiutil.cpp
+++ b/code/network/multiutil.cpp
@@ -800,6 +800,7 @@ void multi_create_player( int net_player_num, player *pl, const char* name, net_
 	Net_players[net_player_num].reliable_socket = PSNET_INVALID_SOCKET;
 	Net_players[net_player_num].s_info.kick_timestamp = -1;
 	Net_players[net_player_num].s_info.voice_token_timestamp = -1;
+	Net_players[net_player_num].s_info.player_collision_timestamp = -1;
 	Net_players[net_player_num].s_info.tracker_security_last = -1;
 	Net_players[net_player_num].s_info.target_objnum = -1;
 	Net_players[net_player_num].s_info.accum_buttons = 0;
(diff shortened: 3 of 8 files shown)


JohnAFernandez:
Just realized that I left out the damage adjustment, so more testing is required.  Probably going to build a calmer test mission too.

JohnAFernandez:
Accidentally pressed close when I should have hit comment.

code/network/multi_observer.cpp
```diff
@@ -60,6 +60,7 @@ void multi_obs_create_player(int player_num,char *name,net_addr *addr,player *pl
 	Net_players[player_num].reliable_socket = PSNET_INVALID_SOCKET;
 	Net_players[player_num].s_info.kick_timestamp = -1;
 	Net_players[player_num].s_info.voice_token_timestamp = -1;
+	Net_players[player_num].s_info.player_collision_timestamp = -1;
```
Goober5000, line 63: This should be initialized to `timestamp(-1)`

(and almost certainly `kick_timestamp` and `voice_token_timestamp` as well)

code/network/multiui.cpp
```diff
@@ -2799,6 +2799,7 @@ void multi_sg_init_gamenet()
 	// assign my player struct and other data	
 	Net_player->flags |= (NETINFO_FLAG_CONNECTED | NETINFO_FLAG_DO_NETWORKING);
 	Net_player->s_info.voice_token_timestamp = -1;	
+	Net_player->s_info.player_collision_timestamp = -1;
```
Goober5000, line 2802: Ditto

code/network/multiutil.cpp
```diff
@@ -800,6 +800,7 @@ void multi_create_player( int net_player_num, player *pl, const char* name, net_
 	Net_players[net_player_num].reliable_socket = PSNET_INVALID_SOCKET;
 	Net_players[net_player_num].s_info.kick_timestamp = -1;
 	Net_players[net_player_num].s_info.voice_token_timestamp = -1;
+	Net_players[net_player_num].s_info.player_collision_timestamp = -1;
```
Goober5000, line 803: Ditto

code/object/collideshipship.cpp
```diff
@@ -1328,7 +1329,37 @@ int collide_ship_ship( obj_pair * pair )
...
+										if (!timestamp_elapsed(current_player.s_info.player_collision_timestamp)) {
+											damage = 0.0f;
+										} else if (!(&Ships[LightOne->instance] != Player_ship)) {
+											current_player.s_info.player_collision_timestamp = timestamp(PLAYER_COLLISION_TIMESTAMP);
+										}
+									}
+								}
+							}
+						} else {
+							damage *= (float) (Game_skill_level*Game_skill_level+1)/(NUM_SKILL_LEVELS+1);
```
Goober5000, line 1361: Shouldn't the damage multiplier happen on `MULTIPLAYER_MASTER` as well, when damage is not set to 0?

Goober5000 (changes requested):
Noticed some things

code/object/collideshipship.cpp
JohnAFernandez: Yes, that's why I switched it back into draft.  I made that change and ran another test.  Still ran pretty well.  Going to run one last one in release instead of debug, despite the fact that that shouldn't effect anything.

JohnAFernandez:
Ran some more tests with the damage factor restored and now it causes damage indistinguishable from singleplayer.

code/network/multi_observer.cpp
JohnAFernandez: Changes made

JohnAFernandez:
New tests:

Before fix:
https://youtu.be/bVBXyAsFupE

After fix:
https://youtu.be/YqTOc4DERuc

code/network/multi_observer.cpp
JohnAFernandez: I switched it back to `timestamp(0)` because there's no need to set it to an invalid timestamp and check for invalid timestamp every time there's a collision.  It seems like timestamp(0) was set up for this exact situation.

code/object/collideshipship.cpp
```diff
@@ -1328,7 +1329,48 @@ int collide_ship_ship( obj_pair * pair )
...
-						damage *= (float) (Game_skill_level*Game_skill_level+1)/(NUM_SKILL_LEVELS+1);
+
+						// Cyborg17 - Pretty hackish, but it's our best option, limit the amount of times a collision can
+						// happen to multiplayer clients, because otherwise the server can kill clients far too quickly.
+						// So here it goes, first only do this on the master (has an intrinsic multiplayer check) 
+						if (MULTIPLAYER_MASTER) {
+							// iterate through each player
+							for (net_player & current_player : Net_players) {
+								// check that this player's ship is valid, and that it's not the server ship.
+								if ((current_player.m_player != nullptr) && (&current_player != Netgame.server) && (current_player.m_player->objnum > 0) && current_player.m_player->objnum < MAX_OBJECTS) {
```
notimaginative, line 1340: `&current_player != Netgame.server` is the same as `!(current_player.flags & NETINFO_FLAG_AM_MASTER)`, which is how that check is done most everywhere else in the code. It's probably best to just stick with the flag check so that it's easier to find relevant code later on.

notimaginative (changes requested):

code/object/collideshipship.cpp
```diff
@@ -1328,7 +1329,48 @@ int collide_ship_ship( obj_pair * pair )
...
+										// check to see if it has been long enough since the last collision, if not negate the damage
+										if (!timestamp_elapsed(current_player.s_info.player_collision_timestamp)) {
+											damage = 0.0f;
+										} else {
+											// make the usual adjustments
+											damage *= (float) (Game_skill_level*Game_skill_level+1)/(NUM_SKILL_LEVELS+1);
+											// if everything is good to go, set the timestamp for the next collision
+											current_player.s_info.player_collision_timestamp = timestamp(PLAYER_COLLISION_TIMESTAMP);
+										}
+									} else if (HeavyOne == &Objects[current_player.m_player->objnum]) {
```
Baezon, line 1355: This check isn't mirrored above. Is that intentional?

code/object/collideshipship.cpp
```diff
@@ -1328,7 +1329,48 @@ int collide_ship_ship( obj_pair * pair )
...
+										// check to see if it has been long enough since the last collision
+										if (!timestamp_elapsed(current_player.s_info.player_collision_timestamp)) {
+											damage = 0.0f;
+										} else if (!(&Ships[LightOne->instance] != Player_ship)) {
+											// make the usual adjustments
+											damage *= (float) (Game_skill_level*Game_skill_level+1)/(NUM_SKILL_LEVELS+1);
+											// if everything is good to go, set the timestamp for the next collision
+											current_player.s_info.player_collision_timestamp = timestamp(PLAYER_COLLISION_TIMESTAMP);
+										}
+									}
```
Baezon, line 1368: The two big blocks above could be handled in just one by having a single object* set to either LightOne or HeavyOne depending on if its current_player, and then just acting on that one object.

code/object/collideshipship.cpp
```diff
@@ -1328,7 +1329,48 @@ int collide_ship_ship( obj_pair * pair )
...
+											current_player.s_info.player_collision_timestamp = timestamp(PLAYER_COLLISION_TIMESTAMP);
+										}
+									} else if (HeavyOne == &Objects[current_player.m_player->objnum]) {
+										// set this as not an interpolated ship
+										multi_oo_set_client_simulation_mode(HeavyOne->net_signature);
+
+										// check to see if it has been long enough since the last collision
+										if (!timestamp_elapsed(current_player.s_info.player_collision_timestamp)) {
+											damage = 0.0f;
+										} else if (!(&Ships[LightOne->instance] != Player_ship)) {
```
Baezon, line 1362: This check isn't mirrored above. Is that intentional?

code/object/collideshipship.cpp
JohnAFernandez: Won't feel confident about that fix until I have my IDE back....

code/object/collideshipship.cpp
JohnAFernandez: Ok this should be fixed now.

code/object/collideshipship.cpp
JohnAFernandez: Ok, I was probably doing way more edits than you were expecting, but I think this does everything.  I'll need to run it through one more test to be sure.

code/object/collideshipship.cpp
```diff
@@ -1328,7 +1329,48 @@ int collide_ship_ship( obj_pair * pair )
...
+										// check to see if it has been long enough since the last collision, if not negate the damage
+										if (!timestamp_elapsed(current_player.s_info.player_collision_timestamp)) {
+											damage = 0.0f;
+										} else {
+											// make the usual adjustments
+											damage *= (float) (Game_skill_level*Game_skill_level+1)/(NUM_SKILL_LEVELS+1);
+											// if everything is good to go, set the timestamp for the next collision
+											current_player.s_info.player_collision_timestamp = timestamp(PLAYER_COLLISION_TIMESTAMP);
+										}
+									} else if (HeavyOne == &Objects[current_player.m_player->objnum]) {
```
Goober5000, line 1355: It looks like you can consolidate these mostly-similar blocks with the following:
```
if (LightOne == &Objects[current_player.m_player->objnum] || HeavyOne == &Objects[current_player.m_player->objnum]) {
	// set this as not an interpolated ship
	multi_oo_set_client_simulation_mode(Objects[current_player.m_player->objnum].net_signature);
```

See also @Baezon's review question.

Goober5000 (changes requested):

Baezon (approved):
LGTM

Goober5000 (approved):
Looks good.

notimaginative:
I see the change now. Looks good.

Issue #3161, opened 2021-01-22
A plan to revise tube lights.


EatThePath:
This is essentially a follow-up to #3146, but for various reasons it seems better to start a new one and close that. 

Tube lights have an ugly cut-off in certain circumstances. I want to fix them, and @TRBlount  asked me to detail my plan, so here we are.

To recap the current engine situation: When a beam is being fired, two `light` objects are created, a point light at the muzzle and a tube light running the length of the beam. The tube is primarily defined by a radius and two points. 

![image](https://user-images.githubusercontent.com/4766365/105557131-14187700-5cd1-11eb-9ea1-2d8b06257c32.png)

When the rendering functions handle the light, the tube light is translated into a cylindrical lighting volume and two spherical ones

![image](https://user-images.githubusercontent.com/4766365/105557217-44601580-5cd1-11eb-9f01-cd8b59db3604.png)

Lighting attenuation within the cylinder is done based on distance to the nearest point on the vector running along it's center. Since deferred lighting shades fragments if they visually overlap a light's volume in screen space, rather than if they are inside the volume in world-space, if the viewpoint is close enough to the center of the cylinder the full length of the cylinder's smooth falloff can often be rendered. If however we have a viewpoint like the example, then the sharp edge of the cylinder can be visible in the right lighting conditions.

![image](https://user-images.githubusercontent.com/4766365/105557077-fe0ab680-5cd0-11eb-9949-94290f1c3835.png)

My plan is to move the cylinder back along it's central vector by 1 radius and extend it by 2. Cylinder light volumes are a simple cylinder mesh one unit long with one face at the origin, so these operations should be quite simple. The shader has access to the radius, and when building the beam vector on which to find the nearest point it can repeat those operations in reverse. The end cap point volumes can be eliminated as redundant, but are left in the next illustration for demonstration purposes.

![image](https://user-images.githubusercontent.com/4766365/105557250-580b7c00-5cd1-11eb-9de1-41143530b48a.png)
 
The current attenuation math can already be seen to produce the right results if allowed to act. This should always allow that when appropriate, instead of just in select circumstances. It is essentially turning the light volume into a bounding cylinder for a whole capsule, rather than one component element of that capsule. 

[...]

PR #3167, opened 2021-01-23
Cylinder only tube lights

### LINKED ISSUE #3146: Graphical issue: Tube lights cut off abruptly at one end.
Now that tube lights are actually doing something besides acting like sphere lights at their origin some flaws in their implementation are coming to, ahem, light.

The beam tube light cuts off sharply at one end, and only one end, the 'far' end of the light, which is the firing point of a beam. Light extends as you would like past the impact point. The end cap light makes this flaw less visible but does not make it entirely invisible, under the right conditions it can stick out badly. 

screen0864
It is visible here(Everything that isn't a tube light in this scene has been terminated for clarity) both at the very back of the Orion where the shivan beams cut off, and on the ear of the Hecate for the AAA beam.  From this I can also be pretty sure it's not a flat circular cut-off perpendicular to the end of the beam, at least to my eyes. (Edit: Or looking at the shivan beams again, maybe it is, and there's something screwy about the scaling of the very flickery lighting of the pulse beam that's throwing me off. It's possible, there's a lot of data transformation steps along the way giving chances to cross wires.) 

From grappling with the code in `deferred-f.sdr` that handles the light attenuation for tubes (line 55-78) I have a hunch that the nearest-point calculation goes screwy past the end of the beam, but I haven't been able to wring obvious flaws out of it yet.(Ed: The longer I think about this the more I think that's unlikely to be an issue, the proceedure there seems straightforward and sound enough.)

EatThePath:
Implements #3161 / fixes #3146. Extends tube light cylinder volumes to cover the end cap sphere volumes and removes those spheres. Before and after test videos available at https://www.youtube.com/watch?v=G5pXm7Jv8Us and https://www.youtube.com/watch?v=gC84qUBXJq8, but leaving in draft till I get some more general play time on the modified build to make sure no weird edge cases crop up

### DIFF
--- a/code/def_files/data/effects/deferred-f.sdr
+++ b/code/def_files/data/effects/deferred-f.sdr
@@ -65,10 +65,17 @@ void GetLightInfo(vec3 position, out vec3 lightDir, out float attenuation)
 		if (lightType == LT_TUBE) {  // Tube light
 			float beamLength = length(beamVec);
 			vec3 beamDir = beamVec / beamLength;
+			//The actual 'lighting element' is shorter than the light volume cylinder
+			//To compensate the light is moved forward along the beam one radius and the length shortened
+			//this allows room for clean falloff of the light past the ends of beams.
+			vec3 adjustedLightPos = lightPosition - (beamDir * lightRadius);
+			beamLength = beamLength - (lightRadius * 2.0);
+			//adjustments having been made, lightdir needs recalculating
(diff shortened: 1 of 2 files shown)


EatThePath:
I've done less playtesting of this than I wanted and don't forsee doing much more any time soon, but that was mostly desired out of an abundance of caution, so it's either submit this for real now or let it languish for possibly months. I choose the former.

The-E (approved):

PR #3172, opened 2021-01-26
Fix special char hex codes for utf8

### LINKED ISSUE #3163: UTF-8: Crash if Training Messages contain Prompts with Special Character Keys
If Unicode Mode is active, the game crashes if a training message with a key prompt appear where the prompted function was set on a key that has a special character in it.

Examples for such keys are ä, ö, ç or ñ which are present on some keyboards.

I am playing with "$Disable built-in translations: Yes"

Interestingly, the game still calls Backspace "Rücktaste" in the german language setting nevertheless. So it is still translated.
But as the special character keys are not displayed in the Config Screen and in the Hotkey Tips in the directives monitor at all right now, this is maybe only some kind of UTF-8 oversight regarding the Training Message Gauge :)

[fs2_open.log](https://github.com/scp-fs2open/fs2open.github.com/files/5860816/fs2_open.log)

BMagnu:
Should fix #3163 until the proper change with the controls code comes through.
In Unicode mode, these are now valid and display correctly, in non-unicode mode, it just displays as a blank / broken char, just like before.
I have no clue what encoding these were before, so I had to guess which characters were actually needed there, but I am reasonably sure to have picked the correct ones, save for polish, especially Arrow Up, which is now called "Kursor Góra". I am fairly sure it's what it was called before as well, but I cannot confirm that this spelling is actually correct.
Shouldn't matter too much anyways, since this will get deleted with controls6 anyways

### DIFF
--- a/code/controlconfig/controlsconfigcommon.cpp
+++ b/code/controlconfig/controlsconfigcommon.cpp
@@ -219,7 +219,7 @@ config_item Control_config[CCFG_MAX + 1] = {
 	{                           -1,                 -1, -1,           0, "",                                       CC_TYPE_TRIGGER,    -1, -1, 0, false, false }
 };
 
-const char* Scan_code_text_german[] = {
+const char* Scan_code_text_german_u[] = {
 	"",				"Esc",				"1",				"2",				"3",				"4",				"5",				"6",
 	"7",				"8",				"9",				"0",				"Akzent '",				"Eszett",				"R\xc3\xbc""cktaste",		"Tab", // NOLINT
 	"Q",				"W",				"E",				"R",				"T",				"Z",				"U",				"I",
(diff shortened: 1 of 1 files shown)


asarium (changes requested):
Did you test this without unicode mode enabled? At least to me, this looks like it will break non-unicode mode since then we do not support rendering UTF-8 encoded text.

BMagnu:
> Did you test this without unicode mode enabled? At least to me, this looks like it will break non-unicode mode since then we do not support rendering UTF-8 encoded text.

Yes. It is not perfect though, as it displays potentially one bad character, as seen here:
image
image

However, missing characters like this are also visible for all non ASCII chars anways

BMagnu:
It may, worst case, cause two broken chars instead of one (however, one should always be filtered out by most text functions killing chars above 127), but should never crash. And since the old code also never properly shows special localized chars, it shouldn't make anything worse.

asarium:
I'm sorry but if this breaks the existing rendering without unicode mode then this is not a viable fix in this state.

BMagnu:
Huh, seems like the rendering of the old code worked in mission. I always assumed it breaks as well, just like in the menu.
I can extend the fix to select the table based on status of the unicode flag

asarium (approved):

PR #3175, opened 2021-01-27
Non-linear Trail decay and Texture stretch

Baezon:
Adds a field for an exponent to apply to the 0 - 1 alpha progress for non-linear decay.

Also allows texture to stretch by a specified amount rather than 1 full texture per trail segment. Switching `i` for `n` was necessary because `i` is measured from the tip to the tail, and thus the points move along with the trail, `n` are the points that are stable in space which is why they are needed. Before it was 1 texture per segment anyway so it didn't matter which it was.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -3878,6 +3878,10 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 		if (optional_string("+Alpha Decay Exponent:")) {
 			trails_warning = false;
 			stuff_float(&sip->afterburner_trail_alpha_decay_exponent);
+			if (sip->afterburner_trail_alpha_decay_exponent < 0.0f) {
+				Warning(LOCATION, "Trail Alpha Decay Exponent of ship %s cannot be negative. Reseting to 1.\n", sip->name);
+				sip->afterburner_trail_alpha_decay_exponent = 1.0f;
+			}
 		}
 			
 		if ( optional_string("+Life:") ) {
@@ -4297,8 +4301,13 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 		required_string("+End Alpha:");
 		stuff_float(&ci->a_end);
 
-		if (optional_string("+Alpha Decay Exponent:"))
+		if (optional_string("+Alpha Decay Exponent:")) {
 			stuff_float(&ci->a_decay_exponent);
+			if (ci->a_decay_exponent < 0.0f) {
+				Warning(LOCATION, "Trail Alpha Decay Exponent of ship %s cannot be negative. Reseting to 1.\n", sip->name);
+				ci->a_decay_exponent = 1.0f;
+			}
+		}
 
 		required_string("+Max Life:");
 		stuff_float(&ci->max_life);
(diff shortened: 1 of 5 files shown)


code/ship/ship.cpp
```diff
@@ -4286,6 +4314,14 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 		generic_bitmap_init(&ci->texture, name_tmp);
 		generic_bitmap_load(&ci->texture);
 
+		if (optional_string("+Bitmap Stretch:")) {
+			stuff_float(&ci->texture_stretch);
+			if (ci->texture_stretch == 0.0f) {
+				Warning(LOCATION, "Trail bitmap stretch of ship %s cannot be 0.  Setting to 1.\n", sip->name);
+				ci->texture_stretch = 1.0f;
+			}
```
Goober5000, line 4322: Ditto the negative question.

code/weapon/weapons.cpp
```diff
@@ -1797,6 +1797,9 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		if ( optional_string("+End Alpha:") )
 			stuff_float(&ti->a_end);
 
+		if (optional_string("+Alpha Decay Exponent:"))
+			stuff_float(&ti->a_decay_exponent);
```
Goober5000, line 1801: Ditto the sanity-check or range-check.

code/weapon/weapons.cpp
```diff
@@ -1810,6 +1813,14 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 			generic_bitmap_init(&ti->texture, fname);
 		}
 
+		if (optional_string("+Bitmap Stretch:")) {
+			stuff_float(&ti->texture_stretch);
+			if (ti->texture_stretch == 0.0f) {
+				Warning(LOCATION, "Trail bitmap stretch of weapon %s cannot be 0.  Setting to 1.\n", wip->name);
+				ti->texture_stretch = 1.0f;
+			}
```
Goober5000, line 1821: Ditto the negative question.

code/ship/ship.cpp
```diff
@@ -3844,6 +3850,15 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
...
 			stuff_string(sip->afterburner_trail.filename, F_NAME, MAX_FILENAME_LEN);
 		}
+
+		if (optional_string("+Bitmap Stretch:")) {
+			trails_warning = false;
+			stuff_float(&sip->afterburner_trail_tex_stretch);
+			if (sip->afterburner_trail_tex_stretch == 0.0f) {
+				Warning(LOCATION, "Trail bitmap stretch of ship %s cannot be 0.  Setting to 1.\n", sip->name);
+				sip->afterburner_trail_tex_stretch = 1.0f;
+			}
```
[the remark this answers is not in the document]
Baezon: It will simply flip the bitmap, a little odd, but nothing usual or particularly misleading, so I don't think there's any reason to exclude it.

code/ship/ship.cpp
```diff
@@ -4272,6 +4297,9 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 		required_string("+End Alpha:");
 		stuff_float(&ci->a_end);
 
+		if (optional_string("+Alpha Decay Exponent:"))
+			stuff_float(&ci->a_decay_exponent);
+
```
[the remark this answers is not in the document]
Baezon: Negative will cap all the values to maximum, no value would be damaging or cause any errors, but negatives could be confusing so yeah that probably shouldnt be allowed

Goober5000 (approved):
It's spelled "Resetting", but that's a nitpick.  Approving.

PR #3176, opened 2021-01-27
Angle-based damage multiplier

Baezon:
Adds fields to weapons to multiply damage based on angle of incidence.


Goober5000 (approved):
Great idea.  Just one comment... why is `+Max` listed before `+Min`?

Baezon:
Well, kinda arbitrary innit? Is it more intuitive to have min first?

Baezon:
Ideally I would make the order not matter, but there's no easy way to do that and not end up confusing (like parsing min, then max, then min again).

PR #3177, opened 2021-01-27
Improved missile avoidance

### LINKED ISSUE #2951: AI reacts to trebs too late
Making the issue because two other blokes are stubborn about it.

Gist of it is, that the AI doesn't react to certain missiles, most notably the trebuchet, at a satisfactory time/distance to where they're able to dodge or launch a counter measure.

@Baezon has mentioned over discord that he wants to look into this.

Baezon:
What it says on the tin, no more AI eating trebs!

Actual changes are a variety of things: only drop countermeasures if the missile is within cm range, always dodge even if you have countermeasures, don't dodge too early, afterburn based on time to impact, not distance, only consider evading missiles actually pointing at you, and better evasion maneuvers (the old method was very strange and not particularly effective).

While they are much better overall there are still reasonable limitations, difficulty affects whether the AI can afterburn, and there are many reasonable situations where the AI simply can't react in time; this doesn't turn AI into superhumans.

Fixes #2951, also sorry @tcrayford but this kinda obviates #2096 as well.


code/ai/ai_profiles.cpp
```diff
@@ -529,6 +529,8 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$better combat collision avoidance for fightercraft:", AI::Profile_Flags::Better_collision_avoidance);
 
+				set_flag(profile, "improved missile avoidance for fightercraft:", AI::Profile_Flags::Improved_missile_avoidance);
```
Goober5000, line 532: This should probably have a dollar sign at the beginning of the line.

Goober5000 (changes requested):
Looks reasonable I think.  One nitpick.

wookieejedi:
Looking forward to testing this one.

Goober5000 (approved):

PR #3178, opened 2021-01-28
Generic debris expansion and configuration

Baezon:
Greatly expands the configuration of the usage of generic debris, allows for arbitrary generic debris models per ship, and allows for big ships to use them as well.

You will see some checks for a -1 `model_num` replaced with a check for the `hull_flag`, **these were equivalent before, but are now no longer.** -1 `model_num` is now a request for the retail debris01.pof, and `is_hull` is whether the debris is collideable custom debris retrieved from the original model, generic debris is the opposite of 'hull' debris.

Since the system is more general now, there are more checks of the generic debris' radius to determine things like cull distance and lifetime. Note that the debris01.pof hardcoded generic debris models have a radius of roughly 1 meter on average and the methods for determining cull distance/lifetime were chosen to maintain roughly their original values at 1 meter, so the existing generic debris will be largely unaffected.

Note also that **I have disabled 'no lighting' for generic debris.** I think we can afford this now and any decently sized generic debris looks like shit without lighting, but I would be open to discussion on this; perhaps a radius check.

Finally a bit of clean up, the very similar functions `create_vaporize_debris()` and `create_subsys_debris()` were rolled into one `create_generic_debris()`, moved to debris.cpp, and values were chosen to roughly maintain their previous behavior.


code/debris/debris.cpp
```diff
@@ -430,7 +430,7 @@ object *debris_create(object *source_obj, int model_num, int submodel_num, vec3d
 		if (vaporize) {
 			dist /= 2.0f;
 		}
-		if ( dist > 200.0f ) {
+		if ( dist > source_obj->radius * 200.0f ) {
```
Goober5000, line 433: Is this right?  Shouldn't it be addition?

code/debris/debris.h
```diff
@@ -45,7 +45,7 @@ typedef struct debris {
 	int		model_instance_num;		// What model instance this uses - needed for arcs
 	int		submodel_num;			// What submodel this uses
 	int		next_fireball;			// When to start a fireball
-	int		is_hull;				// indicates a large hull chunk of debris
+	int		is_hull;				// indicates whether this is a collideable, destructable peice of debris from the model, or just a generic debris fragment
```
Goober5000, line 48: It's spelled "piece"

code/ship/shipfx.cpp
```diff
@@ -1956,6 +1956,22 @@ static void maybe_fireball_wipe(clip_ship* half_ship, sound_handle* handle_array
 				particle::emit( &pe, particle::PARTICLE_SMOKE2, 0, range );
 			}
 
+			if (sip->generic_debris_model_num >= 0) {
+				// spawn a bunch of debris peices, first determine the cross sectional average position to be the force explosion center
```
Goober5000, line 1960: "piece"

code/model/modelinterp.cpp
```diff
@@ -1287,6 +1287,95 @@ void submodel_get_two_random_points_better(int model_num, int submodel_num, vec3
 	}
 }
 
+void submodel_get_cross_sectional_avg_pos(int model_num, int submodel_num, float z_slice_pos, vec3d* pos)
+{
+	polymodel* pm = model_get(model_num);
+
+	if (pm != NULL) {
```
Goober5000, line 1294: CI is going to complain about nullptr

code/model/modelinterp.cpp
```diff
@@ -1287,6 +1287,95 @@ void submodel_get_two_random_points_better(int model_num, int submodel_num, vec3
...
+
+		*pos /= accum_scale_factor;
+	}
+}
+
+void submodel_get_cross_sectional_random_pos(int model_num, int submodel_num, float z_slice_pos, vec3d* pos)
+{
+	polymodel* pm = model_get(model_num);
+
+	if (pm != NULL) {
```
Goober5000, line 1337: Here too

Goober5000 (commented):
Misc comments

code/debris/debris.cpp
Baezon: Ah! That was a typo, I should have 20.0f there, not 200. 10m fighter -> 1m debris, 20.0f x 10m = 200m. so it would roughly follow the precedent set by the original hardcoded value.

code/debris/debris.h
Baezon: I don't think I will ever learn how to spell it properly.

code/debris/debris.cpp
```diff
@@ -475,7 +498,7 @@ object *debris_create(object *source_obj, int model_num, int submodel_num, vec3d
 				db->lifeleft = -1.0f;		// large hull pieces stay around forever
 			}
 		} else {
-			db->lifeleft = (myrand() * RAND_MAX_1f) * 2.0f + 0.1f;
+			db->lifeleft = (frand() * 2.0f + 0.1f) * sqrtf(radius);
```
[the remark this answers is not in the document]
Baezon: This means that larger debris will stick around longer, because otherwise it would only be 0.1-2 seconds, which is far too quick for large or medium sized debris. I found the square root of the radius to give the nicest results for sticking around appropriately long for its size, but still going away relatively quickly, and also maintained the 10m fighter -> 1m debris default values (i.e. it boils back down to the original expression for retail inputs)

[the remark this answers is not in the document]
JohnAFernandez: That is a good idea.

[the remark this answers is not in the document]
JohnAFernandez: I looked at everything else and I'm ready to approve it once you have that comment in.

JohnAFernandez (approved):

PR #3179, opened 2021-01-28
Don't search for empty sound names

Baezon:
If a sound entry is left completely blank, literally just
```
	+WarmupSound:
```
It will match the first index with an empty sound definition! This is bad.

### DIFF
--- a/code/gamesnd/gamesnd.cpp
+++ b/code/gamesnd/gamesnd.cpp
@@ -303,6 +303,10 @@ gamesnd_id gamesnd_get_by_name(const char* name)
 {
 	Assert( Snds.size() <= INT_MAX );
 
+	// empty name is not valid!
+	if (name == nullptr || name[0] == '\0')
+		return gamesnd_id(-1);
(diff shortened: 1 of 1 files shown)


Goober5000:
Wouldn't empty sounds be called `empty` or `none` instead of being completely blank?

Also, would it be better to do a quick out on the blank name, rather than using the blank name to look through a bunch of entries that are going to be skipped anyway?

Baezon:
When 'empty' is found it takes that to mean the filename should be literally empty i.e. the empty string. And its true I suppose, an empty string probably shouldn't get this far to start searching since with this it won't find anything anyway.

Goober5000:
If you check at the top of the function whether `name` is a blank string, that should be sufficient I think.

Goober5000 (changes requested):
Good, but make sure `name` isn't null before you dereference `name[0]`.  Null names also return -1.

Goober5000 (approved):

PR #3180, opened 2021-01-28
Split beam width for collision purposes and lighting purposes

Baezon:
This way `$Collision Radius Override:` can also be used for beam width if `+BeamWidth:` is not specified. For compatibility purposes `+BeamWidth:` is used for *both* if specified, the widest section is the fallback for both.

### DIFF
--- a/code/weapon/beam.h
+++ b/code/weapon/beam.h
@@ -184,7 +184,8 @@ typedef struct beam {
 	int Beam_muzzle_stamp;
 	int firingpoint;
 
-	float		beam_width;
+	float		beam_collide_width;
+	float		beam_light_width;
 } beam;
 
 extern std::array<beam, MAX_BEAMS> Beams;				// all beams
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -433,7 +433,7 @@ int beam_fire(beam_fire_info *fire_info)
 		vm_vec_zero(&new_item->target_pos2);
 	}
 
-	for (int i = 0; i < MAX_BEAM_SECTIONS; i++)
+	for (int i : new_item->beam_section_frame)
 		new_item->beam_section_frame[i] = 0.0f;
 
 	// beam collision and light width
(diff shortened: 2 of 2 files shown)


code/weapon/beam.h
```diff
@@ -184,7 +184,8 @@ typedef struct beam {
 	int Beam_muzzle_stamp;
 	int firingpoint;
 
-	float		beam_width;
+	float		beam_collide_width;
+	float		beam_light_width;
```
Goober5000, line 188: I suggest calling this `beam_visible_width` to be unambiguous.  Since I recently worked in the collision code I first thought of light vs. heavy.

Goober5000 (changes requested):
One naming suggestion, otherwise this looks good.

code/weapon/beam.h
Baezon: Well these numbers really only are used by collision or lighting, I think 'visible' might be confusing since whatever it actually looks like is still determined by the individual widths of the sections.

EatThePath: `beam_light_radius` would be relatively consistent with how that value is refereed to where it will ultimately be used.

EatThePath: Actually scratch that, because you're setting the beam width that lighting is based off of, not the actual light radius. Which actually is a bit not my favorite path forward, but I'll make a separate comment on that.

Goober5000: It still refers to the width of the complete beam, right, not the sections?  I think that's fine, because if you're interested in the sub-widths you would use something like `beam_section_width`.

Baezon: Well, see the whole point of this is trying to clear up the ambiguity of "width of the complete beam". With this PR `beam_light_width` is *only* used for sizing the light cast by the beam, nothing else. This is *sort of* what one may mean by "visible beam", but I personally would take that to mean the width of the individual sections, or more broadly, whichever is the widest. That is relevant to light and collision only insofar as it is the last fallback.

Goober5000: Okay.  I'll defer to your understanding then.

Goober5000 (approved):

Issue #2393, opened 2020-05-15
FR: AI missile lock time override


EatThePath:
This request is mainly pertinent to when multilock features are included. Under mutli-lock, certain setups perform very differently for players and AIs. Say you have a bomb that can lock a single target four times. The effect of that is that the player can fire one bomb relatively quickly, or keep holding course and lock all four bombs after waiting four times as long.

However, since the AI doesn't use the multi-lock features itself, what happens is they wait the base lock time, then fire a volley of four missiles. The wild disparity in behavior between the AI and player when using the same missile can get quite awkward. Thus, I would like a way to define a wait time before firing for AIs only that overrides the normal behavior of waiting out the normal base lock time.

z64555:
I would prefer the AI be smart enough to determine if it wants to fire one bomb or all 4 in the volley, rather than had something hardcoded that would essentially have them *only* fire the full volley.

I'd say do a little research into how the AI acquires a lock on their target and what all determines when they're able to fire, as they should closely emulate as to whats capable by a human pilot.

Baezon:
Although this is technically a separate feature request, I'll close this for now since the original reason it was proposed was fulfilled. Can always be re-opened later if this feature is wanted for another reason.

PR #3183, opened 2021-01-28
Add AI support for multilock

### LINKED ISSUE #2629: FR: Make the AI respect multilock mechanics and restrictions.
At present the AI treats multilock missiles as if they were single-target corkscrew or swarm missiles without any of the mutlilock mechanics applying to them. This can seriously distort their behavior and poses a major balance challenge to modes. 

If implemented this would obviate my original reasons for request #2393

Baezon:
Closes #2629 and #2393

There some discussion on how to handle the decision procedure for AI using multilock
```
//  This function returns if the ship could theoretically have fully locked all available targets
// given the current time it's been locking on its primary target
//  If so, it cheats a little bit and assumes that's what it has done, and fills 
// the ai's missile_locks_firing vector, and to be fired at those targets
```
But I think this is a good compromise for speed, reasonable effectiveness, and unnoticeable cheating. Also pinging @TRBlount because I've re-organized multilock in general by separating out common multilock functions into weapons.cpp


JohnAFernandez:
What bug does this fix?

Baezon:
As #2629 mentions, the AI will currently ignore not only the benefits but also the limitations imposed by the multilock system.

code/ai/aicode.cpp
```diff
@@ -7888,11 +8030,13 @@ void update_aspect_lock_information(ai_info *aip, vec3d *vec_to_enemy, float dis
 			(tshpp == NULL ||
 			ship_get_closest_subsys_in_sight(tshpp, SUBSYSTEM_ENGINE, &aiobjp->pos))))) {
 				aip->aspect_locked_time += flFrametime;
-				if (aip->aspect_locked_time >= wip->min_lock_time) {
-					aip->aspect_locked_time = wip->min_lock_time;
```
Goober5000, line 7892: Hmm, why remove this line?

code/hud/hudlock.cpp
```diff
@@ -807,36 +776,16 @@ void hud_lock_acquire_uncaged_subsystem(weapon_info *wip, lock_info *lock, float
...
-			vm_vec_normalize(&vec_to_target);
-			ss_dot = vm_vec_dot(&Player_obj->orient.vec.fvec, &vec_to_target);
+		for (ship_subsys* ss = GET_FIRST(&sp->subsys_list); ss != END_OF_LIST(&sp->subsys_list); ss = GET_NEXT(ss) ) {
 
-			if ( ss_dot < wip->lock_fov ) {
+			if (!weapon_multilock_can_lock_on_subsys(Player_obj, lock->obj, ss, wip, &ss_dot))
 				continue;
-			}
-
-			if ( !ship_subsystem_in_sight(lock->obj, ss, &Eye_position, &ss_pos) ) {
```
Goober5000, line 837: I wonder how much of an impact changing from Eye_position to the object's position will have on the targeting code.  In most cases it shouldn't matter, but there *are* ships where the eye position is noticeably offset.  Can you change the function to accept the eye position?

code/weapon/weapons.cpp
```diff
@@ -8812,3 +8812,91 @@ bool weapon_has_iff_restrictions(weapon_info* wip)
...
+	extern int Nebula_sec_range;
+	// reduce firing range in nebula
+	if ((The_mission.flags[Mission::Mission_Flags::Fullneb]) && Nebula_sec_range) {
+		weapon_range *= 0.8f;
+	}
+
+	return dist_to_target <= weapon_range;
+}
+
+bool weapon_multilock_can_lock_on_subsys(object* shooter, object* target, ship_subsys* target_subsys, weapon_info* wip, float* out_dot) {
```
Goober5000, line 8840: Remove "multilock" from the function name since it should be applicable to single-lock missiles too.

code/weapon/weapons.cpp
```diff
@@ -8812,3 +8812,91 @@ bool weapon_has_iff_restrictions(weapon_info* wip)
...
+		return false;
+
+	vec3d gsubpos;
+	vm_vec_unrotate(&gsubpos, &target_subsys->system_info->pnt, &target->orient);
+	vm_vec_add2(&gsubpos, &target->pos);
+
+	return ship_subsystem_in_sight(target, target_subsys, &shooter->pos, &gsubpos) == 1;
+}
+
+bool weapon_multilock_can_lock_on_target(object* shooter, object* target_objp, weapon_info* wip, float* out_dot) {
```
Goober5000, line 8868: Remove "multilock" here too.

code/weapon/weapons.cpp
```diff
@@ -8812,3 +8812,91 @@ bool weapon_has_iff_restrictions(weapon_info* wip)
...
+
+	vec3d vec_to_target;
+	vm_vec_normalized_dir(&vec_to_target, &target_objp->pos, &shooter->pos);
+	float dot = vm_vec_dot(&shooter->orient.vec.fvec, &vec_to_target);
+
+	if (out_dot != nullptr)
+		*out_dot = dot;
+
+	if (dot < wip->lock_fov)
+		return false;
```
Goober5000, line 8899: This if() block isn't in the original code.  Or rather, it *is*, but not in the same place.  Should it be taken out?

Goober5000 (changes requested):
Noticed a few things.

code/ai/aicode.cpp
Baezon: Normally, `aspect_locked_time` is capped at `min_lock_time`, but for multilock, it needs to go above so it can get 2x 3x, etc for double-locking, triple-locking etc.

The only other time this is relevant is when lock is lost, where it would've normally started counting down from `min_lock_time`, since it was capped there, so now the lines below were added to reset it back to `min_lock_time` if it was above.

code/hud/hudlock.cpp
Baezon: Ah, I hadn't even noticed that change, yeah that should be possible.

code/weapon/weapons.cpp
Baezon: Ah, good catch, thank you. This function is just checking the general ship properties to determine if it would be a suitable target, comparing dot needs to be done later which depends on if its a big ship (where dot is compared instead to the specific subsystems) or not. Dot is already properly checked later so yeah it doesn't need to be here, and I'll clarify the comment on this fuction.

code/weapon/weapons.cpp
Baezon: While all of this is indeed checked for single-target missiles, these functions *are* only used by multilock, which I think is important information to convey. I guess I would say that the "multilock" here is less about what its actually doing and instead about who uses these functions.

Goober5000: Oh hm.  Yes, that is important information to convey, but I thought these functions were used by single-target also.

Baezon: They still do all these things but they are a lot more spread around, its doing a bunch of different stuff in between all these checks. These functions are basically optimized to only the bare minimum for checking the validity of a target (since its being done en masse). If you still think the name should be changed, fair enough, but that's the situation.

Goober5000: Gotcha.  Then keep the current name, but add a comment to document that this function really is only used by multilock.

Goober5000 (approved):

Goober5000:
I'll call this ready to go.

PR #3187, opened 2021-01-29
Give swarm missile handling a well-deserved makeover

Baezon:
It was getting a bit old and crusty in swarm.cpp

Swarm missiles were doing *a lot* of bending over backwards and complicated overhead, fucking around with these 'paths' for the swarm missile (even going so far as to store state on the ship that fired it?? `next_swarm_path`) the whole purpose of which is to ensure that the swarm missiles choose different direction to 'zig' in, so they don't all follow the same perfect path which would look weird. What they did was massively overengineered for something that can be easily solved with the right RNG function.

`golden_ratio_rand()`, as the comment describes, has all the benefits of rng while avoiding runs and clumps, subsequent calls to the function will evenly spread themselves around the 0 - 1 range, and is perfectly suited for this purpose. Getting rid of these dumb paths was most of the fat cutting here.

This is a comparison, and I'll give you dollar if you can tell the difference.

https://user-images.githubusercontent.com/17305613/106226914-484ed480-61b6-11eb-8322-b36ed07052fd.mp4

https://user-images.githubusercontent.com/17305613/106226923-4d138880-61b6-11eb-89c8-35cc2c9ad22a.mp4

(mp4 support now woot!)

Additionally, weapons now have a unique pointer to their `swarm_info`, instead of an index into the arbitrarily sized `Swarm_missiles` array allowing for indefinite numbers of them simultaneously. Only a single element was plucked from the array each time anyway so this should have a negligible effect on speed.

Turret swarm handling is even worse in terms of ugly overhead tbh, but one step at a time.

### DIFF
--- a/code/math/floating.cpp
+++ b/code/math/floating.cpp
@@ -11,6 +11,7 @@
 
 #include "io/timer.h"
 #include "math/floating.h"
+#include "math/staticrand.h"
 
 
 /**
@@ -77,7 +78,10 @@ static float accum_golden_ratio_rand_seed = 0.0f;
 // generates a quasirandom, low discrepancy number sequence
 // acts very much like a random number generator, but has the property of being very well distributed
 // use if you need quasirandom numbers, but don't want ugly 'runs' or 'clumps' of similar numbers 
-float golden_ratio_rand() {
+float golden_ratio_rand(int seed) {
+	if (seed >= 0) 
+		accum_golden_ratio_rand_seed = static_randf(seed);
+
 	accum_golden_ratio_rand_seed += GOLDEN_RATIO;
 	if (accum_golden_ratio_rand_seed >= 1.0f)
 		accum_golden_ratio_rand_seed -= 1.0f;
--- a/code/math/floating.cpp
+++ b/code/math/floating.cpp
@@ -71,3 +71,15 @@ int rand_chance(float frametime, float chance)
 
 	return frand() < (frametime * (chance + 1.0f));
 }
+
+static float accum_golden_ratio_rand_seed = 0.0f;
+
+// generates a quasirandom, low discrepancy number sequence
+// acts very much like a random number generator, but has the property of being very well distributed
+// use if you need quasirandom numbers, but don't want ugly 'runs' or 'clumps' of similar numbers 
+float golden_ratio_rand() {
+	accum_golden_ratio_rand_seed += GOLDEN_RATIO;
+	if (accum_golden_ratio_rand_seed >= 1.0f)
+		accum_golden_ratio_rand_seed -= 1.0f;
+	return accum_golden_ratio_rand_seed;
+}
--- a/code/math/floating.cpp
+++ b/code/math/floating.cpp
@@ -78,10 +78,7 @@ static float accum_golden_ratio_rand_seed = 0.0f;
 // generates a quasirandom, low discrepancy number sequence
 // acts very much like a random number generator, but has the property of being very well distributed
 // use if you need quasirandom numbers, but don't want ugly 'runs' or 'clumps' of similar numbers 
-float golden_ratio_rand(int seed) {
-	if (seed >= 0) 
-		accum_golden_ratio_rand_seed = static_randf(seed);
-
+float golden_ratio_rand() {
 	accum_golden_ratio_rand_seed += GOLDEN_RATIO;
 	if (accum_golden_ratio_rand_seed >= 1.0f)
 		accum_golden_ratio_rand_seed -= 1.0f;
--- a/code/math/floating.h
+++ b/code/math/floating.h
@@ -115,6 +115,6 @@ extern float fl_roundoff(float x, int multiple);
 
 const float GOLDEN_RATIO = 0.618033989f;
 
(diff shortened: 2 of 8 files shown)


code/weapon/swarm.cpp
```diff
@@ -195,65 +155,34 @@ void swarm_update_direction(object *objp)
...
 	pi = &Objects[wp->objnum].phys_info;
-	swarmp = &Swarm_missiles[wp->swarm_index];
 
-	// check if homing is lost.. if it is then get a new path to move swarm missile along
+	// check if homing is lost.. if it is then set the target point dead ahead
 	if ( swarmp->homing_objnum != -1 && hobjp == &obj_used_list ) {
 		swarmp->change_timestamp = 1;
-		swarmp->path_num = -1;
 		swarmp->homing_objnum = -1;
+		swarm_set_defaut_target_pos(objp, swarmp);
```
JohnAFernandez, line 166: Just double checking, did you check if the behavior matches if the target is lost, too?

code/weapon/swarm.cpp
```diff
@@ -128,65 +117,36 @@ void swarm_maybe_fire_missile(int shipnum)
...
-	if ( i >= MAX_SWARM_MISSILES ) {
-		nprintf(("Warning","No more swarm missiles are available\n"));
-		return -1;
-	}
-
-	swarmp->flags = 0;
 	swarmp->change_timestamp = 1;
-	swarmp->path_num = -1;
 	swarmp->homing_objnum = -1;
+	swarmp->zig_direction = golden_ratio_rand() * PI2;
```
JohnAFernandez, line 134: This is a big ask, but is there a way to create a static version of this rand function for multi syncing?  Anything based on the net signature will really help the missiles to take the same path.  If anyone decides later on to create a swarm bomb that the multiplayer clients have to shoot down, having the static function will make it possible rather than completely impossible.

JohnAFernandez (commented):
This is really good work!  I've got a big ask, though.  

Also, I can't really comment on unique pointers because I have yet to use them.

code/math/floating.h
```diff
@@ -113,4 +113,8 @@ extern float fl_roundoff(float x, int multiple);
  */
 #define IS_NEAR_ZERO(x, e) (fl_abs(x) < (float)(e))
 
+const float GOLDEN_RATIO = 0.618033989f;
```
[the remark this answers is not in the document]
Baezon: I honestly think it's ok as is, this is part of why the golden ratio is so great, the golden ratio -1 has like literally the all properties of the "regular" golden ratio. So much so, that I really think that making the distinction basically doesn't matter. (Except for here where I would like to wrap less often, the only reason that I subtracted one)

code/weapon/swarm.cpp
Baezon: Yes, even as largely irrelevant I might think it to be, this was done specifically the maintain behavior on lost target.

code/math/floating.cpp
```diff
@@ -71,3 +71,15 @@ int rand_chance(float frametime, float chance)
 
 	return frand() < (frametime * (chance + 1.0f));
 }
+
+static float accum_golden_ratio_rand_seed = 0.0f;
```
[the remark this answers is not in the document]
Baezon: Yeah it could, although I'm fairly sure no one would notice.

code/weapon/swarm.cpp
Baezon: Yeah, that shouldn't be a problem.

JohnAFernandez: Thank you!

code/weapon/swarm.cpp
```diff
@@ -128,65 +117,36 @@ void swarm_maybe_fire_missile(int shipnum)
 	}
 }
 
+// if it can't home on something, just set target dead ahead
+void swarm_set_defaut_target_pos(object* objp, swarm_info* swarmp) {
```
Goober5000, line 121: "default" is misspelled

code/weapon/swarm.cpp
```diff
@@ -35,11 +35,8 @@
 #define TURRET_SWARM_VALIDITY_CHECKTIME	5000	// number of ms between checks on turret_swam_info checks
 
 #define SWARM_USED						(1<<0)
```
Goober5000, line 37: It looks like `SWARM_USED` can also be removed, if I'm reading things right.

Goober5000 (changes requested):
This is a great idea, and that golden ratio trick is pretty clever.

code/weapon/swarm.cpp
Baezon: It is still used by turret swarm handling unfortunately.

Goober5000 (approved):
Sorry, I apparently dropped the ball on getting back to this.  It's a great PR, the before and after videos look indistinguishable, and the comments have been addressed.  Approving.

JohnAFernandez:
@Baezon and I also decided that the multi issues are very minor and not worth holding this up. (just no easy way for swarm missiles to have the same zigzags, since seeding seems to ruin the effect). I think it's time.

PR #3191, opened 2021-02-01
Additional FRED features

Goober5000:
This implements a few requested FRED features...

1) The Event dialog is now resizable
2) Shortcut keys (Ctrl+X, Ctrl+C, Ctrl+V, Ctrl+P) work for event nodes, albeit with an error beep due do a somewhat hackish implementation
3) FREDders can now annotate their events with background colors, similar to the existing annotation with comments.

The main window shortcut keys have been slightly cleaned up as well.  In particular "Rotate around object" is now Alt+R (to match with rotate R) rather than Ctrl+V.

It turns out that neither event node indentation nor "inline" comments are possible, at least with MFC.  I also blame MFC for providing at least three ways to potentially solve the cut/copy/paste beep problem, none of which appear to work.


JohnAFernandez (approved):
I honestly could not find anything wrong.  It is pretty annoying that this would also have an error sound, but I'm hoping we'll have better luck with that in qtFRED.

Goober5000:
The main problem is that there is an "official" way to do cut/copy/paste.  MFC is beeping because that official way is not being used.  Unfortunately, despite hours of trying, I could not make the "official" way work, and I had to resort to a hack.  I would welcome anyone who would like to explore further.

Issue #1062, opened 2016-11-01
Proposal: fire-turret sexp


WoolieWool:
This sexp would work exactly the same as fire-beam, but would be used to fire a normal turret instead. This would allow precise control of heavy non-beam turrets like railguns, mass drivers, torpedo launchers, etc.

asarium:
As far as I know, the reason this doesn't exist yet is that before firing the turret it needs to be aimed at the target which may take some time so it's not easy to implement with SEXPs. However, I added the required functionality to the scripting API some time ago so you could write a script that aims and fires the turret and then call that script from a SEXP.

WoolieWool:
But don't beam turrets on multiparts like the Orion's frontal beam need to be aimed too, and regular turrets that are not on multiparts likewise wouldn't need to be aimed?

asarium:
I don't know the specifics of the code but AFAIK the fire-beam SEXP just ignores the multi-part nature of the turret and fires the beam directly.

wookieejedi:
To update this discussion, there is a script which does this, but it doesn't quite work as expected for single part turrets. My guess is because if the ship is not in the fov for the single part turret then it will forget the target, but then because turret.Target = ship was called it turns off auto targeting. As opposed to a multi part that will try and move and search for it

RhymesHLP:
Reiterating the need for this because this is a significant hurdle for balancing capship-heavy missions.

Goober5000:
I actually tried coding a `fire-turret` sexp *years* ago.  It turns out it was more complicated than it looked.  That doesn't mean it's impossible, just that it involved more moving parts than I wanted to deal with at the time.

wookieejedi:
Indeed... In thinking about/reviewing the current the code I think the best way to make an easier PR and still get a good amount of use would be a `set-unset-turret-target` sexp. This would allow the FREDer to get pretty close to the behavior they might want (ie telling which ships to fire at what) but would be much easier to code and hence more likely to actually get done.

asarium:
In my opinion, if something is feasible to be implemented with scripting then it should be implemented that way and exposed to FRED via a scripting SEXP. That way the SCP does not have to deal with keeping backwards compatibility for two APIs (SEXPs and Lua) and the scripting API is possibly extended to more use cases which could be useful in the future.

Baezon:
[...]

PR #3192, opened 2021-02-03
 Add turret-set-forced-target SEXP

Baezon:
...and sister SEXPs `turret-set-forced-subsys-target` and `turret-clear-forced-target`

No more turret-pretend-to-fire silliness! I've put forth my thoughts already on #1062 since there seemed to be some resistance to this idea, but just to reiterate my main point 

"Given the clearly very large demand for this in sexp form and the significantly higher barrier to entry and overhead involved with scripting I think a sexp implementation is sorely needed. This very situation has resulted in FREDders using hacky workarounds like 'turret-pretend-to-fire' a clearly undesirable result!"

I do think the scripting version shouldn't change, it's more 'forgetful' of its assigned target, but sometimes that's good especially if you're controlling it with scripts anyway and want a 'lighter' touch. The more brute-force common scenario however is FREDders who want the damn AI to do what they tell it to do. This gives them that tool.

Edit: I would also argue this would warrant closure of #1062


wookieejedi (approved):
Looks straightforward to me. Only optional nitpick is adding an extra tab of whitespace in the additions to sexp.cpp.

The-E (approved):

PR #3195, opened 2021-02-03
Fix speed differential for retail cap ships

wookieejedi:
As astutely determined by @Baezon `the speed differential for retail cap ships is caused by 'ships with no shields can manage ETS', being part of the bugfixes with target version`. This PR adds two flags per Discord suggestion, one for all ships and one for only fighters/bombers.


Baezon (approved):

code/ai/ai_profiles.cpp
```diff
@@ -525,7 +525,9 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$respect ship axial turnrate differences:", AI::Profile_Flags::Use_axial_turnrate_differences);
 
-				set_flag(profile, "$ships with no shields can manage ETS:", AI::Profile_Flags::nonshielded_ships_can_manage_ets);
+				set_flag(profile, "$ships with no shields can manage ETS:", AI::Profile_Flags::all_nonshielded_ships_can_manage_ets);
```
Goober5000, line 528: I would make the text explicit, such as `$any ship with no shields can manage ETS:`.  Keep the old string too, for compatibility, but add a deprecation warning.

code/ai/ai_profiles.cpp
```diff
@@ -525,7 +525,9 @@ void parse_ai_profiles_tbl(const char *filename)
...
+
+				set_flag(profile, "fighters/bombers with no shields can manage ETS:", AI::Profile_Flags::fightercraft_nonshielded_ships_can_manage_ets);
```
Goober5000, line 530: Add a `$`.

code/hud/hudets.cpp
```diff
@@ -212,10 +212,14 @@ void ai_manage_ets(object* obj)
...
+	if (!ship_info_p->max_speed || !ship_info_p->max_weapon_reserve) {
+		return;
+	}
+
 	// also check if the ship has no shields and if the AI is allowed to manage weapons and engines --wookieejedi
-	if ( !ship_info_p->max_speed || !ship_info_p->max_weapon_reserve || 
-		(!ship_p->ship_max_shield_strength && !aip->ai_profile_flags[AI::Profile_Flags::nonshielded_ships_can_manage_ets]) ) {
-			return;
+	if (!ship_p->ship_max_shield_strength && !(aip->ai_profile_flags[AI::Profile_Flags::all_nonshielded_ships_can_manage_ets] || 
+		(ship_info_p->is_fighter_bomber() && aip->ai_profile_flags[AI::Profile_Flags::fightercraft_nonshielded_ships_can_manage_ets])) ) {
```
Baezon, line 221: Actually, the logic here is a bit confusing, could we get some more parentheses to make it more clear?

Baezon (changes requested):

Goober5000 (changes requested):
Couple things, otherwise this looks good.

code/ai/ai_profiles.cpp
wookieejedi: Good idea, committed!

Goober5000 (approved):

code/hud/hudets.cpp
wookieejedi: Definitely

Baezon (approved):

PR #3197, opened 2021-02-04
Expose Additional Countermeasure Options to Modders

wookieejedi:
This PR provides modders with additional options to alter countermeasure values. Previously, countermeasures had a fixed, hardcoded fire wait time of 333 milliseconds, regardless of what `$Fire Wait:` was specified. Also, the AI would wait a fixed delay between firing of failed countermeasure launches and successful countermeasure launches. For successful launches the AI would double the fire wait. Also, Friendly ships would only use a fixed constant value for countermeasure firing chance, and not the value based on their class specific AI profile.

This PR exposes those values with the following options:

- Add main AI table flag: `$friendly ships use AI profile countermeasure chance:`. This will allow Friendly ships to use their AI class profile countermeasure firing chance instead of the default fixed value from main AI table.

- Add value in CM section of weapons table: Countermeasures use fire wait time `+Use Fire Wait`. This is a boolean that allows the countermeasure’s fire wait to actually use what is specified in the `$Fire Wait:` section of that weapon.

- Add value in CM section weapons table: `+Fail Delay Equals Fire Wait`. The delay between failed countermeasure launches is set to use the fire wait value.

- Add value in CM section weapons table: `+Successful Launch Delay Multiplier`. This allows modders to set the value that is multiplied to fire wait for successful countermeasure launches.

Everything is tested and works as expected.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -2344,12 +2344,24 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 			wip->cmeasure_firewait = (int) (wip->fire_wait*1000.0f);
 		}
 
-		if (optional_string("+Fail Delay Equals Fire Wait:")) {
-			stuff_boolean(&wip->cmeasure_faildelay_equals_firewait);
+		if (optional_string("+Failure Launch Delay Multiplier for AI:")) {
+			int delay_multiplier;
+			stuff_int(&delay_multiplier);
+			if (delay_multiplier > 0) {
+				wip->cmeasure_failure_delay_multiplier_ai = delay_multiplier;
+			} else {
+				mprintf(("Warning: \"+Failure Launch Delay Multiplier for AI:\" should be >= 0 (read %i). Value will not be used. ", delay_multiplier));
+			}
 		}
 
-		if (optional_string("+Successful Launch Delay Multiplier:")) {
-			stuff_int(&wip->cmeasure_sucess_delay_multiplier);
+		if (optional_string("+Successful Launch Delay Multiplier for AI:")) {
+			int delay_multiplier;
+			stuff_int(&delay_multiplier);
+			if (delay_multiplier > 0) {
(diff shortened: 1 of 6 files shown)


Baezon (changes requested):
Remember that the fails and successes messing with the countermeasure firewait is an AI only thing, you'll probably want to make that quite clear.

Also a minor point but it seems a bit backwards to assume a fail = firewait while a success gets the multiplier. It would seem that the success is the more 'normal' situation where you'd want it to normally be firewait while the unusual fails get some modification of the fire wait.

wookieejedi:
Good idea, I've incorporated those suggestions and added some error checks, too.

code/weapon/weapons.cpp
```diff
@@ -2344,12 +2344,24 @@ int parse_weapon(int subtype, bool replace, const char *filename)
...
 
-		if (optional_string("+Fail Delay Equals Fire Wait:")) {
-			stuff_boolean(&wip->cmeasure_faildelay_equals_firewait);
+		if (optional_string("+Failure Launch Delay Multiplier for AI:")) {
+			int delay_multiplier;
+			stuff_int(&delay_multiplier);
+			if (delay_multiplier > 0) {
+				wip->cmeasure_failure_delay_multiplier_ai = delay_multiplier;
+			} else {
+				mprintf(("Warning: \"+Failure Launch Delay Multiplier for AI:\" should be >= 0 (read %i). Value will not be used. ", delay_multiplier));
```
Baezon, line 2353: Should be a legit `Warning()`, this will merely silently print to the debug log, look around you'll see how others do it in this area. Ditto below.

Baezon (changes requested):
Looks good, aside from the one thing.

code/weapon/weapons.cpp
wookieejedi: Good idea, just updated.

Baezon (approved):

PR #3198, opened 2021-02-05
Add 'landmark' logging

Baezon:
When looking at a debug log for more than just the particular error/warning, and attempting to narrow down a particular time frame for perhaps when an error may have occurred, there are very few landmarks to help get an idea of where in a mission something happens just through the log. Shader compilations, ani warnings, assorted minor errors, frames "too long" all are pretty useless to determine what the heck is going on in the mission at the time, even if you know what all these are.

This adds what I like to call 'landmark' logging. Merely prints when big|huge ships arrive, depart, or are destroyed providing 'landmarks' for orienting oneself temporally. Shouldn't be too spammy either, even in the worst scenarios several dozen prints will hardly put a dent in it.

For the non-arriving situation, some existing code was repurposed that was being filtered out, being `nprintf`'s.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -2457,23 +2457,20 @@ int parse_create_object_sub(p_object *p_objp)
 
 	// If the ship is in a wing, this will be done in mission_set_wing_arrival_location() instead
 	// If the ship is in a wing, but the wing is docked then addition of bool brought_in_docked_wing accounts for that status --wookieejedi
-	if (Game_mode & GM_IN_MISSION) {
+	if (Game_mode & GM_IN_MISSION && ((shipp->wingnum == -1) || (brought_in_docked_wing))) {
+		object *anchor_objp = (anchor_objnum >= 0) ? &Objects[anchor_objnum] : nullptr;
 
-		if (Ship_info[shipp->ship_info_index].is_big_or_huge()) {
-			float mission_time = f2fl(Missiontime);
-			int minutes = (int)(mission_time / 60);
-			int seconds = (int)mission_time % 60;
+		Script_system.SetHookObjects(2, "Ship", &Objects[objnum], "Parent", anchor_objp);
+		Script_system.RunCondition(CHA_ONSHIPARRIVE, &Objects[objnum]);
+		Script_system.RemHookVars({"Ship", "Parent"});
+	}
 
-			mprintf(("%s arrived at %02d:%02d\n", shipp->ship_name, minutes, seconds));
-		}
+	if (Game_mode & GM_IN_MISSION && Ship_info[shipp->ship_info_index].is_big_or_huge()) {
+		float mission_time = f2fl(Missiontime);
+		int minutes = (int)(mission_time / 60);
+		int seconds = (int)mission_time % 60;
 
-		if ((shipp->wingnum == -1) || (brought_in_docked_wing)) {
-			object* anchor_objp = (anchor_objnum >= 0) ? &Objects[anchor_objnum] : nullptr;
-
-			Script_system.SetHookObjects(2, "Ship", &Objects[objnum], "Parent", anchor_objp);
-			Script_system.RunCondition(CHA_ONSHIPARRIVE, &Objects[objnum]);
-			Script_system.RemHookVars({"Ship", "Parent"});
-		}
+		mprintf(("%s arrived at %02d:%02d\n", shipp->ship_name, minutes, seconds));
 	}
 
 	return objnum;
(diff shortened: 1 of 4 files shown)


wookieejedi:
This is an excellent idea which I think many folks will find very useful.

code/mission/missionparse.cpp
```diff
@@ -2457,12 +2457,23 @@ int parse_create_object_sub(p_object *p_objp)
...
-		Script_system.SetHookObjects(2, "Ship", &Objects[objnum], "Parent", anchor_objp);
-		Script_system.RunCondition(CHA_ONSHIPARRIVE, &Objects[objnum]);
-		Script_system.RemHookVars({"Ship", "Parent"});
+		if ((shipp->wingnum == -1) || (brought_in_docked_wing)) {
+			object* anchor_objp = (anchor_objnum >= 0) ? &Objects[anchor_objnum] : nullptr;
+
+			Script_system.SetHookObjects(2, "Ship", &Objects[objnum], "Parent", anchor_objp);
+			Script_system.RunCondition(CHA_ONSHIPARRIVE, &Objects[objnum]);
+			Script_system.RemHookVars({"Ship", "Parent"});
+		}
```
Goober5000, line 2476: This isn't the right way to do it.  First, that is one conceptually complete condition; separating it up like you did doesn't really make sense.  Second, there are multiple places where a ship can arrive and this is just one place.

Put the condition back to the way it was.  If you want to print your log message along with the script hook, this makes sense, but you will need to do it all four places the script hook appears.  Alternatively, you could put your mprintf statements inside `mission_log_add_entry`.

Goober5000 (changes requested):

code/mission/missionparse.cpp
Baezon: I considered putting it in `mission_log_add_entry` but it is very general purpose needing only the name of the entity, I kind of didn't want to hurt its 'elegance' by having to do complicated switching over its type to extract the name after the fact, all so I could use the Ship to see if it was big|huge.

Baezon:
I've added the printing to ship-creation, but I think leaving it out of wing arrivals is fine, this is most helpful for just important vessels, not *every* ship.

code/mission/missionparse.cpp
```diff
@@ -2465,6 +2465,14 @@ int parse_create_object_sub(p_object *p_objp)
...
 	}
 
+	if (Game_mode & GM_IN_MISSION && Ship_info[shipp->ship_info_index].is_big_or_huge()) {
+		float mission_time = f2fl(Missiontime);
+		int minutes = (int)(mission_time / 60);
+		int seconds = (int)mission_time % 60;
+
+		mprintf(("%s arrived at %02d:%02d\n", shipp->ship_name, minutes, seconds));
+	}
+
```
Goober5000, line 2475: This correctly belongs inside the previous `if` block, just don't split the condition while you're doing it.

Goober5000 (changes requested):
Ok about not adding this to wing ship arrivals.  But it still is not quite in the right place

code/mission/missionparse.cpp
Goober5000: Perhaps, but `mission_log_add_entry` does contain code that already gets the ship object.

code/mission/missionparse.cpp
Baezon: I'm not sure I understand, the wingnum and dock wing stuff isn't relevant for the logging, so it should not be under that condition. If I only put it under the in-mission condition, then it would necessarily involve splitting the in-mission condition from the wingum and dock wing conditions.

Goober5000: The wingnum and dock wing stuff is very relevant.  Consider that the script hook must fire once and only once for every ship that arrives.  There is one hook here that handles single ships, and one hook in `mission_set_wing_arrival_location` that handles ships in wings.  So if you want your logging statement to fire once and only once for every ship, you should use exactly the same logic.

(Since you don't care about ships in wings, technically the best thing to do is put your block inside the other `if` block, below the script condition, and add `!brought_in_docked_wing` to your block's condition.)

In your very first draft, the logging statement would have fired for every non-wing ship, but it also would have fired for *some* wing ships - not all of them - and it would not have been obvious why some ships were being logged but others weren't.

In the current draft, the most appropriate condition is
```
if (Game_mode & GM_IN_MISSION && Ship_info[shipp->ship_info_index].is_big_or_huge() && (shipp->wingnum == -1)) {
```
but since it's so similar, you might as well enclose it in the previous block with the one modification.

Baezon: Ok, I think I see, I wasn't aware of this wing stuff. So within the same if block and only conditional on big|huge right?

Goober5000: Yes, so just:
```
if (Ship_info[shipp->ship_info_index].is_big_or_huge() && !brought_in_docked_wing)
```

Goober5000 (approved):
That looks good.

JohnAFernandez (approved):

PR #3199, opened 2021-02-06
add useful function

Goober5000:
Another function ported from vecmat to vecmath.

### DIFF
--- a/code/scripting/api/objs/vecmath.cpp
+++ b/code/scripting/api/objs/vecmath.cpp
@@ -591,6 +591,22 @@ ADE_FUNC(getNormalized,
 	return ade_set_args(L, "o", l_Vector.Set(v3));
 }
 
+ADE_FUNC(projectParallel,
+	l_Vector,
+	"vector unitVector",
+	"Returns a projection of the vector along a unit vector.  The unit vector MUST be normalized.",
+	"vector",
+	"The projected vector, or NIL if a handle is invalid")
+{
+	vec3d *src, *unit;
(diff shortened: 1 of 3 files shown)


code/scripting/api/objs/vecmath.cpp
```diff
@@ -591,6 +591,22 @@ ADE_FUNC(getNormalized,
 	return ade_set_args(L, "o", l_Vector.Set(v3));
 }
 
+ADE_FUNC(projectParallel,
+	l_Vector,
+	"vector unitVector",
+	"Returns a projection of the vector along a unit vector.  The unit vector MUST be normalized.",
```
asarium, line 597: Since that precondition is covered by an assert, could it also be checked here?

asarium (changes requested):

code/scripting/api/objs/vecmath.cpp
Goober5000: You mean, copy the Assert that's in `vm_vec_projection_parallel` and put it in `projectParallel` too?

I shouldn't think that's necessary.  The Lua function is merely a wrapper around the FSO function, and the FSO function handles the Assert.  The `projectOntoPlane` function is almost identical, wraps a different function in the same way, and also doesn't contain the Assert that's inside the wrapped function.

asarium: I meant that more along "check the condition normally and cause a `LuaError` if it fails". Asserts are next to useless for scripts since they do not contain the Lua callstack.

Goober5000: Oh, I see.  That does make sense.

asarium (approved):

PR #3203, opened 2021-02-10
Actually fix radar range turning to -1 after hud gauge config

JohnAFernandez:
Problem was caused by negative array index write in this section of code.  It looked like this was impossible because either I have a specific gauge selected or all of them selected right?  Not unless I start moving the sliders around *before* I select a gauge!!


JohnAFernandez:
Should fix #3136

The-E (approved):

PR #3204, opened 2021-02-10
Overhaul briefing icon parsing

Baezon:
Adds modular table support for icons.tbl, with xxx-ico.tbm. icons.tbl can use either the 'old style' or 'new style' (explicit $species and $icon type) definitions, but .tbms must be done in the new style.

Additionally adds '$Borrows Briefing Icons from:' to species, to simply copy another species' icons. Unfortunately, because this is done *after* modular table parsing, this will override modular tables and thus any species icons that were coped from another cannot be modified through modular tables. This is unideal, but if this copying is done before modular tables, not all icons may even be defined yet, and also changes made to a 'lender' species icons will not propagate to its associated 'borrower' species (although this may be desirable).

A more complex form of parsing would need to be done to resolve this in general, so I decided to pick the option I thought was more generally useful.

### DIFF
--- a/code/mission/missionbriefcommon.cpp
+++ b/code/mission/missionbriefcommon.cpp
@@ -303,7 +303,7 @@ int add_briefing_icons() {
 
 	Briefing_icon_info.push_back(bii);
 
-	return Briefing_icon_info.size() - 1;
+	return (int)Briefing_icon_info.size() - 1;
 }
 
 void brief_icon_parse_cleanup() {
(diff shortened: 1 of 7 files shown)


JohnAFernandez (approved):
I can't see anything wrong.  A very welcome change indeed!  Just make sure it parses tables the same as before the change, and I think we should be good to go.  Maybe mark the lambda functions with a quick comment just in case a newer coder comes across it, but I'll leave that up to you.

Baezon:
> 
> 
>Just make sure it parses tables the same as before the change, and I think we should be good to go. 

Missed this the first time around, just wanted to clarify that this is what I meant by 'old style', like the retail table. It will be used exactly the same way, new mods built on retail never have to touch icons.tbl if they dont want to.

JohnAFernandez:
Ok.  Everyone has had a chance to look at it.  Let's see what happens... *pushes button*

PR #3205, opened 2021-02-11
If the first #ControlConfigOverride isn't named, use "default".

MageKing17:
This causes it to replace the first preset in the list, which more closely approximates the old behavior.


Goober5000 (approved):

z64555 (approved):
Looks a bit rough, but it should work.

PR #3206, opened 2021-02-12
restore retail behavior of XSTR

Goober5000:
This is a follow-up to #2390.  Retail used a weird hybrid approach where internal strings were translated but external strings were not.  This restores that behavior by adding another language, LCL_RETAIL_HYBRID, which is the default.  The game can still be run in completely untranslated or completely translated mode.

All three modes have been tested and translate strings, or not, as expected.  The `$Use tabled strings for the default language:` flag has also been tested and still works.

### DIFF
--- a/code/cutscene/ffmpeg/FFMPEGDecoder.cpp
+++ b/code/cutscene/ffmpeg/FFMPEGDecoder.cpp
@@ -286,7 +286,7 @@ std::unique_ptr<DecoderStatus> initializeStatus(std::unique_ptr<InputStream>& st
 
 	if (status->subtitleStream == nullptr) {
 		// We don't have an external subtitle stream or loading from that file has failed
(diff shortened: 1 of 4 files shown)


JohnAFernandez:
So, I'm not sure what we're trying to solve here.  Wouldn't the translation of external strings be an enhancement that simply improves on retail behavior, or is it causing problems somewhere else?

Goober5000:
Simply put, in retail, internal strings were translated and external strings were not.  This restores that behavior by default.  Other possible behaviors are translating both, and not translating both.  For any language aside from English, the game will always translate both internal and external strings.

The main reason for this PR (as well as many previous language PRs) is that mods are designed with the assumption that XSTR is handled in a particular way.  At one point, briefly, all strings were always translated, but it was quickly discovered that mods which reused random XSTR indexes would have their strings replaced in unpredictable ways.

MageKing17 (approved):
I still think using "special" language codes is adding complexity for little benefit, but at least this restores the retail behavior.

PR #3207, opened 2021-02-12
Fix for waypoint completion with FIT

Baezon:
To determine when a waypoint is complete it compares the distance to the goal versus the distance it will travel this frame (plus some padding). The latter is estimated as the difference between `last_pos` and `pos`, due to the AI reordering with FIT (framerate independent turnrate) however those points will straddle any SEXP `set-object-position`s where before it would not, causing it to erroneously satisfy this condition if the distance moved was at least half the distance to the goal. 

Speed x frametime gives a more consistent estimate of the frame's travel in all scenarios.

Shoutout to Phantom Hoover for hunting this down!

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -4604,13 +4604,14 @@ void ai_fly_to_target_position(vec3d* target_pos, bool* pl_done_p=NULL, bool* pl
 		}
 	}
 
-	if ( (dist_to_goal < MIN_DIST_TO_WAYPOINT_GOAL) || (vm_vec_dist_quick(&Pl_objp->last_pos, &Pl_objp->pos) > 0.1f) ) {
+	float dist_to_cover_this_frame = (Pl_objp->phys_info.speed * flFrametime);
+	if ( (dist_to_goal < MIN_DIST_TO_WAYPOINT_GOAL) || dist_to_cover_this_frame > 0.1f ) {
 		vec3d	nearest_point;
 		float		r;
 
 		r = find_nearest_point_on_line(&nearest_point, &Pl_objp->last_pos, &Pl_objp->pos, target_pos);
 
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
Looks good.  Three questions:
1) Is `phys_info.speed` the correct speed to use?
2) Does this frametime calculation work a) with 1x mission time, b) with accelerated or decelerated (non-1x) mission time, c) if the game is paused and then resumed?
3) Are there any other places in the codebase where this calculation needs to be similarly fixed?

Baezon:
1. Yes, this only cares about distance, so just the straight up speed of the ship is all thats needed.
2. It will work pretty much as it did before, `vm_vec_dist_quick(&Pl_objp->last_pos, &Pl_objp->pos)` implicitly has frametime bundled in it already, this merely decouples the calculation from any futzing with its position the FREDder might be doing.
3. There was only one other location in that uses the `last_pos`-`pos` distance, and it is only one early check, where other later checks dont rely on it.

Goober5000:
1) Ok.
2) Well, I'm wondering particularly if this frametime calculation has the same sort of distinction that requires `getRealFrametime` and `getMissionFrametime` in scripting.
3) And this other location isn't affected if a mission uses `set-object-position`?

Baezon:
2. The connection here is to the ships' position, since that changes with the mission's frametime, the new verison should be too.
3. This the location if you're interested https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/ai/aicode.cpp#L3988 The goal there is merely to catch when its traveling at a "non-tiny" speed. It *does* draw a line between `last_pos` and `pos` later and if that line crosses close enough to the path point it counts as having reached that point, so i guess if you set the position of a ship in the direction of, but way beyond its path point, that would count as having reached it (erroneously?). Although this is an extremely unlikely situation. That check really could just be for some arbitrary small speed instead, but I figured I would leave it since it wasn't harmful as far as i could tell.

Goober5000 (approved):
Ok.

PR #3208, opened 2021-02-12
Properly fix static_rand()'s range and account for this in functions that were misusing it

PhantomHoover:


code/math/staticrand.cpp
```diff
@@ -20,8 +20,9 @@ void init_semirand()
 {
 	Semirand_inited = true;
 
+	// Originally this made a 30-bit rand by sticking two 15-bit rands from myrand() together. Instead we trim rand32() down to size.
 	for (auto & number : Semirand)
-		number = (static_cast<unsigned>(myrand()) << 15u) + myrand();
+		number = (unsigned) (rand32() & 0x3fffffff);
```
Goober5000, line 25: I think you should use STATIC_RAND_MAX here instead of the number itself.

Goober5000 (changes requested):

code/math/staticrand.cpp
PhantomHoover: Good point; I hadn't defined the constant when I wrote that. Fixed.

The-E (approved):

Goober5000 (approved):

PR #3214, opened 2021-02-14
fix event editor closing on key press

Goober5000:
Restores the MFC functions that were taken out in #3073.  These functions appeared redundant, but they were actually necessary to intercept the MFC handling of Enter and Escape.  In addition to fixing that behavior, this PR also adds some clarifying comments and renames some functions to prevent confusion.


JohnAFernandez (commented):
In noticed that escape closed it earlier.  So what happens when you would press ESC and enter with these changes?

Goober5000:
With these changes, neither Escape nor Enter will cause the dialog to close.  But if you are editing a node or an event or a message, Enter/Escape will confirm or cancel the edit respectively.

JohnAFernandez (approved):

Issue #2945, opened 2020-11-30
The fireball limit is not that high and local subspace missiles behave badly when it is reached


EatThePath:
Ran into this recently. Subspace portals are fireballs, only 200 fireballs are allowed at any time(MAX_FIREBALLS defined on line 32 of fireball\fireballs.cpp), and when that limit is reached it just stops creating new ones when asked. 

So, what happens if a local subspace missile(LSSM) is fired under these conditions? It just flies straight ahead until it times out, and then dies. Now if a LSSM tries to re-enter the mission space when there are no fireballs available it handles the situation much more gracefully, but the exiting behavior is pretty rough and also the most likely to be seen by the player.

You may ask, who is going to have more than 200 SSMs fired at once? To which I answer: this guy right here. But it's not a burning need, so I'd consider this a fairly low priority thing.

JohnAFernandez:
This is fixed, wth

PR #3224, opened 2021-02-17
Remove the fireball limit

JohnAFernandez:
Vectorizes the fireball limit.  Thanks to @asarium for some great suggestions.  In draft until we get a proper stress test from @EatThePath 

For #2945

### DIFF
--- a/code/fireball/fireballs.cpp
+++ b/code/fireball/fireballs.cpp
@@ -1068,13 +1068,13 @@ void fireball_render(object* obj, model_draw_list *scene)
 	}
 }
 
+// Because fireballs are only added and removed in two places, and Unused_fireball_indices is always updated in those places to contain unused indices,
+// this very simple code will give you the correct count of currently existing fireballs in use. 
 int fireball_get_count()
 {
-	int count = static_cast<int>(Fireballs.size() - Unused_fireball_indices.size());
+	int count = static_cast<int>(Fireballs.size()) - static_cast<int>(Unused_fireball_indices.size());
 
-	if (count < 0) {
-		count = 0;
-	}
-
+	Assert (count >= 0);
+	
 	return count;
 }
--- a/code/fireball/fireballs.cpp
+++ b/code/fireball/fireballs.cpp
@@ -31,13 +31,15 @@ int Knossos_warp_ani_used;
 
 #define MAX_WARP_LOD	0
 
-fireball Fireballs[MAX_FIREBALLS];
+constexpr int INTITIAL_FIREBALL_CONTAINTER_SIZE = 256;
+
+SCP_vector<fireball> Fireballs;
+SCP_vector<int> Unused_fireball_indices;
 
 fireball_info Fireball_info[MAX_FIREBALL_TYPES];
 
 int fireball_used[MAX_FIREBALL_TYPES];
 
-int Num_fireballs = 0;
 int Num_fireball_types = 0;
 
 bool fireballs_inited = false;
@@ -435,8 +437,6 @@ void fireball_load_data()
 // This will get called at the start of each level.
 void fireball_init()
 {
-	int i;
-
 	if ( !fireballs_inited ) {
 		// Do all the processing that happens only once
 		fireball_parse_tbl();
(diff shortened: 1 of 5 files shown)


wookieejedi:
Yessss

code/fireball/fireballs.cpp
```diff
@@ -31,13 +31,15 @@ int Knossos_warp_ani_used;
 
 #define MAX_WARP_LOD	0
 
-fireball Fireballs[MAX_FIREBALLS];
+constexpr int INTITIAL_FIREBALL_SIZE = 256;
+
+SCP_vector<fireball> Fireball_vec(INTITIAL_FIREBALL_SIZE);
+SCP_vector<int> Unused_fireball_indices(INTITIAL_FIREBALL_SIZE);
```
asarium, line 37: Careful with the vector constructor call. It _adds_ that many elements and does not only reserve as much. Since you already have the relevant code to reserve the required size below, you can just use the default constructor here.

code/fireball/fireballs.cpp
```diff
@@ -467,27 +467,25 @@ void fireball_delete( object * obj )
 	fireball	*fb;
 
 	num = obj->instance;
-	fb = &Fireballs[num];
+	// some temporary asserts until we're sure the new system works fine.
```
asarium, line 470: Why temporary? More consistency checks are always good :wink:

code/fireball/fireballs.cpp
```diff
@@ -863,51 +837,58 @@ int fireball_create(vec3d *pos, int fireball_type, int render_type, int parent_o
...
 			warp_lifetime = 4.0f;
-		fb->total_time = warp_lifetime;	// in seconds
+		new_fireball.total_time = warp_lifetime;	// in seconds
 	} else {
-		fb->total_time = i2fl(fl->num_frames) / fl->fps;	// in seconds
+		new_fireball.total_time = i2fl(fl->num_frames) / fl->fps;	// in seconds
 	}
-	
+
+	if (n == static_cast<int>(Fireball_vec.size())) {
```
asarium, line 885: If you keep the fireball a pointer and just point that to the correct element around line 787 then you can avoid having that very specific check here.

code/fireball/fireballs.h
```diff
@@ -95,8 +95,7 @@ typedef struct fireball {
 } fireball;
 // end move
 
-extern fireball Fireballs[MAX_FIREBALLS];
-extern int Num_fireballs;
+extern SCP_vector<fireball> Fireball_vec;
```
asarium, line 98: The name should stay as `Fireballs` for consistency with the rest of the code. That `_vec` suffix isn't really necessary.

code/scripting/api/libs/mission.cpp
```diff
@@ -705,26 +705,32 @@ ADE_INDEXER(l_Mission_Fireballs, "number Index", "Gets handle to a fireball obje
...
 }
 ADE_FUNC(__len, l_Mission_Fireballs, NULL, "Number of fireball objects in mission. Note that this is only accurate for one frame.", "number", "Number of fireball objects in mission")
 {
-	return ade_set_args(L, "i", Num_fireballs);
+	int count = 0;
+	for (auto& current_fireball : Fireball_vec) {
+		if (current_fireball.objnum >= 0) {
+			count++;
+		}
+	}
```
asarium, line 732: Since this relies on the internal implementation of the fireballs, it should be moved to within the fireball module and only called here via a function.

asarium (changes requested):

code/fireball/fireballs.cpp
JohnAFernandez: Sounds good

code/scripting/api/libs/mission.cpp
JohnAFernandez: OK this all sounds easy to do. Thanks for the review!

JohnAFernandez: Take another look if you can!

code/fireball/fireballs.cpp
```diff
@@ -1093,3 +1067,14 @@ void fireball_render(object* obj, model_draw_list *scene)
 			Int3();
 	}
 }
+
+int fireball_get_count()
+{
+	int count = static_cast<int>(Fireballs.size() - Unused_fireball_indices.size());
+
+	if (count < 0) {
```
asarium, line 1075: This is a surprisingly simple way of doing this. Maybe add a comment in this function to explain why this works to make understanding this a bit easier.

Also, be careful with your casting range checks. The way this currently works means that this will not be hit in the cases you want. Since both `size` calls return a `size_t`, you will get an underflow if `Unused_fireball_indices` is larger than `Fireballs`. If you do the casts individually then the result will be as you expect it.

Also, is this a case you would expect? I can't think of something where this could happen without a programming error in which case an assertion would be sufficient.

asarium (changes requested):

code/fireball/fireballs.cpp
JohnAFernandez: I am not expecting it, I just got paranoid. So you're right, I should make it an assert.

asarium (approved):

PR #3233, opened 2021-02-20
Store standalone image on GitHub

asarium:
Docker Hub is going into a weird direction with their rate limits and
the GitHub Container Registry is better in that regard.

Old images will stay in Docker Hub for the time being though.


Goober5000 (approved):
Shrug of approval

Issue #3230, opened 2021-02-19
FR: Expose simulated hull to scripting API


naomimyselfandi:
Exactly what it says on the tin. We can read and write to the ship's actual hull, but interactions with the simulated hull have to go through `runSEXP`.

Admittedly, the use case for this is really narrow - you need to be either writing a script that interacts with a mission that uses sim-hits, or you need to be writing a generic utility script - but the latter case actually did come up in BTA.

Goober5000:
Easily done.  PR forthcoming momentarily.

wookieejedi:
#3234 was merged, so closing this.

PR #3234, opened 2021-02-20
add SimHitpointsLeft

Goober5000:
Implements #3230

### DIFF
--- a/code/scripting/api/objs/object.cpp
+++ b/code/scripting/api/objs/object.cpp
@@ -186,7 +186,7 @@ ADE_VIRTVAR(Physics, l_Object, "physics", "Physics data used to move ship betwee
 
 ADE_VIRTVAR(HitpointsLeft, l_Object, "number", "Hitpoints an object has left", "number", "Hitpoints left, or 0 if handle is invalid")
 {
-	object_h *objh = NULL;
+	object_h *objh = nullptr;
 	float f = -1.0f;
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/object.cpp
```diff
@@ -202,6 +202,24 @@ ADE_VIRTVAR(HitpointsLeft, l_Object, "number", "Hitpoints an object has left", "
...
+	float f = -1.0f;
+	if (!ade_get_args(L, "o|f", l_Object.GetPtr(&objh), &f))
+		return ade_set_error(L, "f", 0.0f);
+
+	if (!objh->IsValid())
+		return ade_set_error(L, "f", 0.0f);
+
+	//Set sim hull strength.
+	if (ADE_SETTING_VAR) {
+		objh->objp->sim_hull_strength = f;
```
asarium, line 217: Is a negative value valid? If not then a range check and error might be a good idea here.

asarium (approved):

code/scripting/api/objs/object.cpp
Goober5000: Good call.

Goober5000: Looks like negative values are, in fact, valid.  For regular hull, they'll just cause the ship to explode immediately.  For sim hull, the value is range-checked when the sexp is called.

PR #3238, opened 2021-02-21
Add vanish Lua function to some objects

BMagnu:
This allows scripting to delete certain objects, akin to ship-vanish in sexp.

### DIFF
--- a/code/scripting/api/objs/fireball.cpp
+++ b/code/scripting/api/objs/fireball.cpp
@@ -122,23 +122,21 @@ ADE_FUNC(isWarp, l_Fireball, NULL, "Checks if the fireball is a warp effect.", "
 	return ADE_RETURN_FALSE;
 }
 
-ADE_FUNC(vanish, l_Fireball, nullptr, "Vanishes this fireball from the mission.", nullptr, "Returns nothing.")
+ADE_FUNC(vanish, l_Fireball, nullptr, "Vanishes this fireball from the mission.", "boolean", "True if the deletion was successful, false otherwise.")
 {
 
 	object_h* oh = nullptr;
 
 	if (!ade_get_args(L, "o", l_Fireball.GetPtr(&oh)))
-		return ADE_RETURN_NIL;
+		return ade_set_error(L, "b", false);
 
 	if (!oh->IsValid())
-		return ADE_RETURN_NIL;
-
-	fireball* fb = &Fireballs[oh->objp->instance];
+		return ade_set_error(L, "b", false);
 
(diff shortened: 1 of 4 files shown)


code/scripting/api/objs/ship.cpp
```diff
@@ -1853,6 +1853,31 @@ ADE_FUNC(getDisplayString, l_Ship, nullptr, "Returns the string which should be
...
+	if (!objh->IsValid())
+		return ADE_RETURN_NIL;
+
+	shipp = &Ships[objh->objp->instance];
+
+	// if MULTIPLAYER bail
+	if (Game_mode & GM_MULTIPLAYER)
+		return ADE_RETURN_NIL;
+
+	auto ship_entry = &Ship_registry[Ship_registry_map[shipp->ship_name]];
```
asarium, line 1874: `objh->objp->instance` will also give you the ship number so you can avoid going through the registry map to get that.

code/scripting/api/objs/fireball.cpp
```diff
@@ -122,6 +122,25 @@ ADE_FUNC(isWarp, l_Fireball, NULL, "Checks if the fireball is a warp effect.", "
...
+	if (!ade_get_args(L, "o", l_Fireball.GetPtr(&oh)))
+		return ADE_RETURN_NIL;
+
+	if (!oh->IsValid())
+		return ADE_RETURN_NIL;
+
+	fireball* fb = &Fireballs[oh->objp->instance];
+
+	//Should be sufficient for Fireballs, as the fireball internal functions also call this, for example to free up a fireball if the limit is reached
+	obj_delete(fb->objnum);
```
asarium, line 139: You can also use `OBJ_INDEX` on `oh->objp` to get the object number without going through the fireball array.

asarium (changes requested):
In general, you could change the return types to a boolean and only return `true` when the call actually succeeded.

code/scripting/api/objs/ship.cpp
BMagnu: Is the Ship_registry index guaranteed to be equal to the Ship's instance number in the Ships array?
Was not sure of this since other ship functions to this as well as opposed to using the the instance number when accessing ShipStatus.

asarium: Yes, I was mistaken here.

However, you do not need to go through the registry at all here. If you have a valid object handle (which is the case at this point) then you know that the ship is present and you can call `ship_actually_depart` unconditionally.

BMagnu: Ah okay. That explains why the sexp needs to do that check but the API does not.
Thanks for clarifying

asarium (approved):

PR #3239, opened 2021-02-21
Update Linux build image and make consistent

asarium:
For some reason the last update of the PR pipeline did not update the
other workflows as well so we had some minor inconsistencies.

This should hopefully fix the nightly build issues for real this time.


JohnAFernandez (approved):
Sounds it needs to go in quick

PR #3240, opened 2021-02-21
openddl-parser library

bryansee:
Lately, I've decided to implement the OpenDDL support via the kimkulling/openddl-parser library, which is simple and fast.

As you know, OpenDDL is a general-purpose, human-readable, and strongly-typed data language for information exchange. It was originally developed as the basis for the Open Game Engine Exchange (OpenGEX) format, which I intend to implement once this is stabilised.

https://openddl.org/


JohnAFernandez:
To quote @The-E:

- OpenGEX is licensed under GPL v3. We can not use that code under any circumstances.
- Even if we could use that code, OpenGEX itself isn't suitable for ingame use, as models made in it are unbelievably huge, parsing them is slow, and the work required to make them usable by the rest of the engine ... is prohibitive and better done in a bespoke asset conversion tool

The-E:
While openddl, by itself, does not suffer from licensing issues that would prohibit its use in FSO, there is no point in adding it without an actual use case.

@bryansee Do not pursue this issue further. No PR related to this project of yours will ever be accepted for inclusion in FSO.

PR #3243, opened 2021-02-23
Decouple asteroids from species

### LINKED ISSUE #3181: FR: Decouple Asteroid.tbl from species_defs.tbl
It's minor but super annoying to deal with species when you have to mess with adding dummy entries in other tables just to add new species. So here's my idea on solving that for asteroid.tbl.

I'm assuming that FSO reads the asteroid.tbl and creates some kind of indexed list that is referenced by the species based on the species index. So with that assumption in mind...

I was thinking something along the lines of giving each debris (it's 3 per species) a +set. So you have something like...

> $Name:				Terran Debris 1
> +Set: Terran Debris
> $Other data

Each set needs to have 3 entries or we error out.

Then in Species_Defs you add $Debris Set.

> $Species_Name: Terran
> $Other Data
> $Debris Set: Terran Debris

If no $Debris Set: is listed then the species gets the debris set matching the species index. If there is no matching index and no debris set then we error out.

Baezon:
The actual connections between asteroids and species are extremely shallow, with retail data they are mostly merely coincidental, so complaining to the high heavens that you didn't give it 3 + 3 * species asteroids and bringing the program to a screeching halt is completely absurd.

Modders can now define however many 'special' asteroids types as they like, or none, the number of species has no relevance or bearing on them (3 rocky type asteroids is still required). The few areas where it was relevant (hud targeting display name and spawn weighting) have been exposed to modders, and unfortunately we'll just have to deal with very strange defaults in those cases.

Closes #3181


Goober5000:
Is it still possible for each species to have its own "debris" asteroid field though?  This was the rationale for debris in FS2; you could have a field of Terran debris chunks or Vasudan debris chunks or Shivan debris chunks.

In general, I agree that having the default asteroid set plus any number of "special" asteroid sets is a much better design.

Baezon:
Of course, you can make the models look like whatever you want and have whatever name you want to display when targeted, this is about as deep as the connection to species goes.

code/asteroid/asteroid.cpp
```diff
@@ -50,9 +49,9 @@ asteroid_obj	Asteroid_objs[MAX_ASTEROID_OBJS];	// array used to store asteroid o
 asteroid_obj	Asteroid_obj_list;						// head of linked list of asteroid_obj structs
 
 // used for randomly generating debris type when there are multiple sizes.
-#define SMALL_DEBRIS_WEIGHT	8
-#define MEDIUM_DEBRIS_WEIGHT	4
-#define LARGE_DEBRIS_WEIGHT	1
+#define SMALL_DEBRIS_WEIGHT	8.0f
+#define MEDIUM_DEBRIS_WEIGHT	4.0f
+#define LARGE_DEBRIS_WEIGHT	1.0f
```
Goober5000, line 54: You could turn these into `const float`s if you want.

code/asteroid/asteroid.cpp
```diff
@@ -599,7 +573,7 @@ void asteroid_create_all()
 		} else {
 			Assert(num_debris_types > 0);
 
-			int rand_choice = rand() % max_weighted_range;
```
Goober5000, line 602: Is the changed random generator still going to produce the same results?

code/asteroid/asteroid.cpp
```diff
@@ -1849,20 +1823,48 @@ static void asteroid_parse_section(asteroid_info *asip)
...
 	}
+
+	if (optional_string("$Display Name:")) {
+		stuff_string(asip->display_name, F_NAME);
+	} else {
+		// take the first word of the name and append "debris"
+		SCP_string current_name = asip->name;
+		size_t split = current_name.find(' ');
+		asip->display_name = current_name.substr(0, split);
+		asip->display_name += " debris";
```
Goober5000, line 1834: Why not append " debris" to the end of the string?

Goober5000 (commented):
Looks good; just a few comments.

code/asteroid/asteroid.cpp
Baezon: I did? I'm not sure what you mean

Goober5000: In this code, you split the string at the first space.  A more accurate question would be, why are you splitting the string before appending?

Baezon: Oh, I'm stupid, I don't think I realized "debris" was the second word of the of ship asteroid names anyway. Still, the first two words is a bit more awkward of a default. Maybe strip numbers?

Goober5000: What would a typical `asip->name` look like if you didn't modify it?

Baezon: 'Terran Debris 1' 'Vasudan Debris 2' etc

Believe me, I would *really* prefer to just use the names as is. Maybe an explicit check for the retail ones? At least so everyone else isn't saddled with this weird default or forced to use $Display Name

Goober5000: Oh, I see, it's actually specified in asteroids.tbl.

Why not use a `has_display_name` field like ship and weapon classes?

Baezon: I'm not sure how to go about it though, because it would have to work without any table changes for the retail ship asteroids, since the HUD was hardcoded to display them as "[species] debris", regardless of their actual name.

Goober5000: Ah.  Then, yeah, do an explicit check, but I would do it this way: If the name is any species, followed by "debris", followed by a number, then the name is changed to "[species] debris".  Otherwise it is left alone.

Baezon: Should it be for *any* "[species] debris #" or just the specific retail names? I could hardcode those retail names in the code to be checked against. I'm likely going to need to do that anyway because I'm going to do another PR to check for the retail asteroids to suppress the inconsistent LOD number warnings they generate.

Goober5000: It should be any species.  Remember that there are plenty of mods out there which defined new species, and some of them may have also defined species-specific debris.

Baezon: Fair enough.

Goober5000: > What if it detects this style of naming and then uses the old index-based species "guess" for the species to fill in the new name

That sounds like the most sensible solution.

code/asteroid/asteroid.cpp
Baezon: Yes, it was merely changed to dealing with floats instead of ints for ease of use by modders, and should have the same behavior.

code/asteroid/asteroid.cpp
```diff
@@ -1849,20 +1823,77 @@ static void asteroid_parse_section(asteroid_info *asip)
...
+
+	if (split == std::string::npos)
+		return;
+
+	SCP_string remaining_name = name.substr(split + 1, name.length());
+	SCP_string species_name = name.substr(0, split);
+
+	int idx = species_info_lookup(species_name.c_str());
+	if (idx < 0)
+		return;
```
Goober5000, line 1865: It would be better to iterate through the list of names and check whether the asteroid name is prefixed by the species name.  Some species' names are two words, so splitting by space wouldn't work.

code/asteroid/asteroid.cpp
```diff
@@ -1849,20 +1823,77 @@ static void asteroid_parse_section(asteroid_info *asip)
...
+
+	if (num.empty() || std::find_if(num.begin(),
+		num.end(), [](unsigned char c) { return !std::isdigit(c); }) != num.end())
+		return;
+
+	// make sure this asteroid would correspond the 'species section' of the old style retail asteroids
+	if (Asteroid_info.size() < NUM_DEBRIS_SIZES + Species_info.size() * NUM_DEBRIS_SIZES && Asteroid_info.size() >= NUM_DEBRIS_SIZES) {
+		idx = (int)(Asteroid_info.size()) / NUM_DEBRIS_SIZES - 1;
+		strcpy_s(asip->name, Species_info[idx].species_name);
+		strcat(asip->name, " debris");
```
Goober5000, line 1886: To be really future-proof, strcat a space, then strcat `XSTR("debris", 348)`

Goober5000 (changes requested):

code/asteroid/asteroid.cpp
Baezon: Good point.

code/asteroid/asteroid.cpp
Baezon: Ah, ok, I wasn't sure how to handle translation like that.

code/asteroid/asteroid.cpp
```diff
@@ -1852,17 +1852,21 @@ static void asteroid_parse_section(asteroid_info *asip)
...
+	size_t split = std::string::npos;
+
+	for (species_info species : Species_info) {
+		split = name.find(species.species_name);
+
+		if (split != std::string::npos) {
+			split = strlen(species.species_name);
+			break;
+		}
+	}
```
Goober5000, line 1864: This won't split correctly if the species name appears in the middle (not the beginning) of the string.

Goober5000 (changes requested):

code/asteroid/asteroid.cpp
Baezon: If split is 0, then right? instead of npos

Goober5000: That would work.  You could also use `compare` with a length parameter, or `strncmp` on the `c_str()`.

code/asteroid/asteroid.cpp
```diff
@@ -1849,20 +1823,80 @@ static void asteroid_parse_section(asteroid_info *asip)
...
+
+	SCP_string remaining_name = name.substr(split + 1, name.length());
+
+	split = remaining_name.find(' ');
+
+	if (split == std::string::npos)
+		return;
+
+	SCP_string debris = remaining_name.substr(0, split);
+	SCP_string num = remaining_name.substr(split + 1, name.length());
```
Goober5000, line 1875: Should this be `remaining_name.length()`?

code/asteroid/asteroid.cpp
```diff
@@ -1849,20 +1823,80 @@ static void asteroid_parse_section(asteroid_info *asip)
...
+
+	split = remaining_name.find(' ');
+
+	if (split == std::string::npos)
+		return;
+
+	SCP_string debris = remaining_name.substr(0, split);
+	SCP_string num = remaining_name.substr(split + 1, remaining_name.length());
+
+	if (debris.compare("Debris") != 0)
```
Goober5000, line 1877: Last comment, I think: Use `stricmp` and `c_str()`.  We want to not be case-sensitive here.

Goober5000 (approved):
Looks good!

Now, have you traced through the code and checked that it works as expected on the retail entries?

Baezon:
Yep, works perfectly on retail, but retail is easy to test, 20 years worth of mods however...

Will probably work in 99% of cases where people were just copy pasting retail entries or replacing species names.

Issue #3225, opened 2021-02-17
Scripting / Sound Bug: audio_stream.play does not use preset volume


the-maddin:
The function audio_stream:play, when invoked with no additional arguments, is documented as using its stream type's volume.

So, for a valid audio_stream `as = ad.openAudioStream("L2-2_SN_02a.wav", AUDIOSTREAM_VOICE)`,
the calls `as:play(ad.MasterVoiceVolume)` and `as:play()` should have the same effect (namely Snipes yelling DIVE).
Instead, no sound plays on the second call (as would be expected from `as:play(0.0)`).

A quick glance at the implementation shows the argument actually _is_ passed to `ausdiostream_play` (from audiostr.h),
so the bug appears to be in the sound code, not in scripting.

the-maddin:
Resolved by merging fix, also updated documentation.

PR #3244, opened 2021-02-24
Set default volume on audiostream creation

the-maddin:
Addressing my issue #3225, I'm setting sane values found in other
parts of the sound system at stream creation, to be used or overwritten later.

Also fixed a potential leak of stream slots and deduplicated some code.

### DIFF
--- a/code/sound/audiostr.cpp
+++ b/code/sound/audiostr.cpp
@@ -959,8 +959,6 @@ static int audiostream_use_next_free( int type )
             return -1;
     }
 
-    Audio_streams[i].Set_Default_Volume(type == ASF_VOICE ? Master_voice_volume * aav_voice_volume : Master_sound_volume);
-
     return i;
 }
 
--- a/code/sound/audiostr.cpp
+++ b/code/sound/audiostr.cpp
@@ -18,6 +18,7 @@
 #include "sound/IAudioFile.h"
 #include "sound/sound.h"
 #include "sound/openal.h"
+#include "gamesnd/eventmusic.h"
 
 #ifdef WITH_FFMPEG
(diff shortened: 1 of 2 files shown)


the-maddin:
Also updated scripting documentation, because that's what `audiostream_play()` _actually_ does.

Goober5000:
Hmm.  Who is familiar with the audio code?  @ln-zookeeper or @z64555 or @notimaginative perhaps?

code/sound/audiostr.cpp
```diff
@@ -922,6 +923,47 @@ void audiostream_close()
...
+            break;
+        case ASF_VOICE: // As in: sound.cpp:590
+            Audio_streams[i].Set_Default_Volume(Master_voice_volume * aav_voice_volume);
+            break;
+        default:
+            Audio_streams[i].status = ASF_FREE;
+            return -1;
+    }
+
+    Audio_streams[i].Set_Default_Volume(type == ASF_VOICE ? Master_voice_volume * aav_voice_volume : Master_sound_volume);
```
asarium, line 962: This will overwrite the value set previously. Is that intended?

asarium (changes requested):
Please also make sure to use the proper whitespace configuration (tabs over spaces). We have an editorconfig file which is supported by most editors (either natively or via a plugin).

code/sound/audiostr.cpp
the-maddin: That was an oversight, left in from before I realized there are more than the two defaults exposed via LUA. Correcting.

notimaginative:
LGTM, other than the whitespace issue that asarium pointed out.

asarium (approved):

PR #3245, opened 2021-02-24
Bugfix for disabled fighters interupting their own support calls

Baezon:
When a fighter is disabled it calls for support and also gives itself an "attack forever" order (presumably so it doesn't try any other complicated behaviors until fixed). That order calls `ai_attack_object()` which will also interrupt any pending calls for support! Preventing and interrupting being fixed indefinitely. This just disables the abort order if it is disabled, when you're disabled getting fixed should be your top priority!

Also renames the function to make it clear this is specifically about support, and not just docking (and also not a 'maybe').


JohnAFernandez:
I love the idea of these changes.  We should have a brief discussion to make sure no one can think of why this shouldn't be done.  This would prevent those couple instances, where you see a fighter just sitting there at the end of the mission, and the reaction is "stupid AI"...

I also feel like this will actually make balance steadier and more predictable, since the random chance of a fighter being disabled does not immediately disqualify it from fighting.

But what does everyone else think?

Goober5000:
Hmm.  I've been noodling over whether or not this needs a flag, ~~but I think no flag is necessary~~ or maybe not, see below.

One can use a similar argument as standardizing on 60 FPS for weapons.  In the 60 FPS case, there was *no* standard, so any FPS was just as good as any other FPS, and 60 FPS is both a) common, and b) makes sense as a benchmark.  In the disabled-support case, the common and benchmark cases are both to assume that *no* AI fighter is disabled in the course of the mission.  Therefore if a fighter is disabled, it can be treated the same way as "undefined behavior" in compiled C code: the compiler can do whatever it wants.  So allowing the fighter to call support successfully is a valid course of action.

However, I thought of one potential pitfall.  Suppose a mission has an objective to capture an enemy fighter.  Once the fighter is disabled, it's trapped and can't escape from the capture situation.  However, if it can call for support, then the capture situation might be thwarted in a way the FREDder didn't anticipate.  This would not happen for Enter the Dragon, since Shivans don't have support ships, but could conceivably happen in a mission where a Terran or Vasudan fighter is captured.

Goober5000:
On further investigation, unless the mod modifies iff_defs.tbl, only the Friendly IFF can call for support.  So the capture situation does not apply.

Baezon:
I think in practice this is going to be an extraordinary edge case, especially for enemies, and it's clearly buggy behavior so I would really prefer not have it as a flag. If the modder needs to, they can always disable support for a particular side on a mission..

Goober5000:
Agreed.  Given that support is by default only called by Friendlies, I'll approve this.

Goober5000 (approved):

JohnAFernandez:
I think everyone has had enough time to look this over.

PR #3246, opened 2021-02-24
add missing #include

Goober5000:
On some setups the hierarchy of #include files doesn't catch an `<iterator>`, so #include it explicitly.

(This fixes the "`std::back_inserter` not found" error on MSVC 2015.)

### DIFF
--- a/code/globalincs/vmallocator.h
+++ b/code/globalincs/vmallocator.h
@@ -2,6 +2,7 @@
 #define _VMALLOCATOR_H_INCLUDED_
 
 #include <deque>
+#include <iterator>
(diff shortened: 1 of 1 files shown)


asarium (approved):
Maybe also add it to the precompiled header.

Goober5000:
> Maybe also add it to the precompiled header.

Where do I find that?

asarium:
It's at `code/prefix_header.h`.

Goober5000:
Hmm, it's already there.  Maybe MSVC 2015 wasn't processing it for some reason.

PR #3251, opened 2021-02-26
Add a CMake platform check for proper non-copyable type support

asarium:
Visual Studio has issues in earlier versions where it does not properly
support returning a type that has a deleted copy operator/constructor.

This works around that by checking for that error with CMake and
applying a workaround if that check fails.


Goober5000 (approved):

z64555:
Related to #3254

Goober5000:
Yeah this new strategy is good.  Not quite as clever as the original (see also the "Yo dawg" comment on #3242) but it does  the job.

PR #3253, opened 2021-02-27
Feature/sexp containers

Karajorma:
SEXP Containers feature. 

This code allows the FREDder to create maps and lists which can be treated similarly to SEXP variables. Integration into FRED is done but the code for writing these containers into the player or campaign files hasn't been written so they are not persistent yet. 

The feature works via CTEXT and does mean that you shouldn't call CTEXT twice for the same node within a call to a SEXP as the result may have changed (that entry might have been removed or an entry added in front of it). This is not a major problem as few SEXPs that did this probably should have been buffering the result rather than taking a computationally expensive trip through CTEXT() anyway. 

There may be an issue preventing containers being altered during briefings and debriefings. I remember simply turning the ability to alter containers off except during the mission itself to prevent the parsing code altering the contents if it thought they were broken. This should be fairly trivial to fix but I never got round to it. 

This thread explains how the feature was supposed to work from a FREDders point of view.
https://www.hard-light.net/forums/index.php?topic=88613.0

And this thread explains strict typing, which is a feature I wanted to add around the same time and would be very useful for containers. 
https://www.hard-light.net/forums/index.php?topic=88533.msg1766351#msg1766351


JohnAFernandez:
Closing this to clear some of the clutter.  Please see the other sexp container prs.

PR #3255, opened 2021-02-27
Properly initialize trail texture stretch

Baezon:
"I really should stop forgetting to initialize things" -Me, like yesterday

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -4331,7 +4331,7 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 				ci->texture_stretch = 1.0f;
 			}
 		}
-		else
+		else if (first_time)
 			ci->texture_stretch = 1.0f;
 
 		if (optional_string("+Faded Out Sections:") ) {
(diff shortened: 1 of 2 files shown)


code/ship/ship.cpp
```diff
@@ -4331,6 +4331,8 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 				ci->texture_stretch = 1.0f;
 			}
 		}
+		else
+			ci->texture_stretch = 1.0f;
```
Goober5000, line 4335: This is not quite correct.  This branch will always be hit when parsing a modular table that doesn't have the `+Bitmap Stretch:` tag, even if the tag was in a previously parsed table.

Changing this to `else if (first_time)` should be sufficient.

Goober5000 (changes requested):
There is an easily-made mistake here.

code/ship/ship.cpp
Baezon: Ah, right, right, thank you

Goober5000 (approved):

JohnAFernandez (approved):

PR #3256, opened 2021-02-28
add switch and functional-switch operators

Goober5000:
Like the C equivalent, these SEXPs allow one action to be taken, or one value to be returned, based on a numeric expression.


Baezon (approved):
Very useful.

Issue #3165, opened 2021-01-23
FR: scripting: Make "Simulated Speech Overrides" accessible through tables (tb) library


the-maddin:
For my tts script, it would be nice to have a list with all the message sound names that get overridden by the in-engine TTS system, so I can mirror its behavior more closely. These consist of some hardcoded filenames and those defined in messages.tbl under #Simulated Speech Override.

I propose, for that purpose, to add an array tb.SimulatedSpeechOverrides to the tables library.

Where this started on Discord: https://discord.com/channels/223511295431933953/223511391531827200/802528546878652417

PR #3260, opened 2021-02-28
Add scripting API for accessing registered Simulated Speech Overrides

BMagnu:
Implements #3165.
I decided to rename the list in question to uppercase to make it conform to other global variable names.

### DIFF
--- a/code/scripting/api/libs/tables.cpp
+++ b/code/scripting/api/libs/tables.cpp
@@ -13,6 +13,7 @@
 #include "weapon/weapon.h"
 #include "menuui/techmenu.h"
 #include "fireball/fireballs.h"
+#include "mission/missionmessage.h"
 
 
 extern bool Ships_inited;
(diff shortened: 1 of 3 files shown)


code/scripting/api/libs/tables.cpp
```diff
@@ -218,6 +219,27 @@ ADE_FUNC(__len, l_Tables_FireballClasses, NULL, "Number of fireball classes", "n
...
+
+ADE_LIB_DERIV(l_Tables_SimulatedSpeechOverrides, "SimulatedSpeechOverrides", NULL, NULL, l_Tables);
+
+ADE_INDEXER(l_Tables_SimulatedSpeechOverrides, "number Index", nullptr, "string", "Truncated filenames of simulated speech overrides or empty string if index is out of range.")
+{
+	int idx = -1;
+	if (!ade_get_args(L, "*i", &idx)) {
+		return ade_set_error(L, "s", "");
+	}
+	if (idx > 0 && (unsigned int) idx <= Generic_message_filenames.size()) {
```
asarium, line 231: You need to use `size_t` here to get the correct type for either 32- or 64-bit:
```suggestion
	if (idx > 0 && (size_t) idx <= Generic_message_filenames.size()) {
```

asarium (changes requested):

asarium (approved):

Issue #3158, opened 2021-01-20
FR: Allow LUA to change active Campaign


BMagnu:
Right now, lua can load an existing campaign file for a pilot. However, this only works if such a csg file exists, which is only the case if the campaign has been selected once before. I would like to have this extended as so lua can change the current active campaign of a pilot to any currently available campaign, even if this pilot has never loaded said campaign before.

MjnMixael:
+1

BMagnu:
Implemented by #3261

PR #3261, opened 2021-02-28
Lua add campaign switch function

BMagnu:
Implements #3158 
Used code similar to the commit button functionality of the campaign selection dialog


Goober5000 (approved):
Looks ok, I think.

PR #3262, opened 2021-03-01
allow two SEXPs to be used in multiplayer that previously couldn't be

Goober5000:
1) The ship-vanish SEXP (and script function) was already multiplayer-ready
2) The is-ship-visible SEXP was enhanced to add a second argument, which defaults to the player's ship in single-player mode

And this adds a few useful Asserts to the caching functions as well.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -242,7 +242,7 @@ SCP_vector<sexp_oper> Operators = {
 	//Ship Status Sub-Category
 	{ "is-in-mission",					OP_IS_IN_MISSION,						1,	INT_MAX,	SEXP_BOOLEAN_OPERATOR,	},	// Goober5000
 	{ "is-docked",						OP_IS_DOCKED,							1,	INT_MAX,	SEXP_BOOLEAN_OPERATOR,	},	// Goober5000
-	{ "is-ship-visible",				OP_IS_SHIP_VISIBLE,						1,	1,			SEXP_BOOLEAN_OPERATOR,	},
+	{ "is-ship-visible",				OP_IS_SHIP_VISIBLE,						1,	2,			SEXP_BOOLEAN_OPERATOR,	},
 	{ "is-ship-stealthy",				OP_IS_SHIP_STEALTHY,					1,	1,			SEXP_BOOLEAN_OPERATOR,	},
 	{ "is-friendly-stealth-visible",	OP_IS_FRIENDLY_STEALTH_VISIBLE,			1,	1,			SEXP_BOOLEAN_OPERATOR,	},
 	{ "is-iff",							OP_IS_IFF,								2,	INT_MAX,	SEXP_BOOLEAN_OPERATOR,	},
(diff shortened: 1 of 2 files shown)


JohnAFernandez (commented):
Double checking, you were able to set up a multiplayer test environment to check that it is multiplayer compatible?

Goober5000:
> Double checking, you were able to set up a multiplayer test environment to check that it is multiplayer compatible?

No, I haven't gotten around to setting that up and it will probably be a while before I do.

JohnAFernandez:
Ok.  It ends up being easier than it seems.  I think I'm going to write a quick guide for how to set up testing in multi.

Goober5000:
Not sure why the CI is flagging an error, because these apparently have nothing to do with the code...
```
D:\a\fs2open.github.com\fs2open.github.com\code\network\multimsgs.cpp(7630): error C2220: the following warning is treated as an error [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\network\multimsgs.cpp(7630): warning C4701: potentially uninitialized local variable 'missile_lifetime' used [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\network\multimsgs.cpp(7631): warning C4701: potentially uninitialized local variable 'missile_pos' used [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\network\multimsgs.cpp(7633): warning C4701: potentially uninitialized local variable 'launch_speed' used [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\network\multimsgs.cpp(7632): warning C4701: potentially uninitialized local variable 'orient_in' used [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
```

code/parse/sexp.cpp
```diff
@@ -6789,32 +6797,40 @@ int sexp_hits_left(int node, bool sim_hull)
...
+
+	// the second argument is the viewer
+	if (n >= 0)
+	{
+		auto viewer_entry = eval_ship(n);
+		if (viewer_entry)
+			viewer_shipp = viewer_entry->shipp;
+	}
+	// if the second argument is not supplied, and we're single-player, default to the player
+	else if (!(Game_mode & GM_MULTIPLAYER))
```
asarium, line 6819: This will silently fail if we are in multiplayer and no second argument is supplied. Is that intentional? Otherwise a warning might be useful to detect this.

asarium (approved):
Not sure if you want to do anything with my comment. Code is good to go but a usability improvement could be done.

I think you can safely ignore that CI failure. That appears to come from #3270. No idea why GitHub thinks it belongs to this PR. We can just force the merge.

code/parse/sexp.cpp
Goober5000: Good idea.  I'll make a change.

PR #3263, opened 2021-03-01
Fix strings.tbl in unicode mode

### LINKED ISSUE #2994: Bug: Starting FSO in German or French with Unicode Mode Enabled crashes
This is due to an issue in the strings.tbl (an FS2 retail table).
A discussion in the discord came to the conclusion, that this should be handled in-engine.

Detailed issue for German:
line 4224 of the strings.tbl has one ``."`` too much at the end of the Line, resulting in a parsing error.

Detailed issue for French:
Somehow, the Engine gets parsing errors on every comment in strings.tbl after line 5390. Remvoing all comments from the file lets the game start properly

BMagnu:
This will fix #2994.
It consist of two parts, one two helper functions to convert strings of a known encoding into the encoding currently used by FSO. Even if we wanted to hardcode the alterative unicode match strings as opposed as determining them at runtime, we _will_ need these functions anyway for the controls6, as there the local-specific keynames by SDL are always in UTF-8 and need to be converted, which is why I would advocate for adding these functions anyways.
Part two is the actual matching. This is... kinda ugly here. Especially with the offset where the changes need to be made changing depending on the game being in unicode mode or not. At least it works, and will now start both in normal mode and unicode mode when not set to english. Though I am open to suggestions, in case you don't like this solution.

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -2327,6 +2327,146 @@ void coerce_to_utf8(SCP_string &buffer, const char *str)
 	Warning(LOCATION, "Truncating non-UTF-8 string '%s' to '%s'!\n", str, buffer.c_str());
 }
 
+//Converts a string in ISO 8859-1 to the currently used encoding
+//Only legal for strings with correct ISO 8859-1 encoding
+SCP_string iso8859_1_to_current(const char* str)
+{
+	SCP_string buffer;
+	//We are not in unicode mode, so just return as is, the string is correct
+	if (!Unicode_text_mode) {
+		buffer.assign(str);
+		return buffer;
+	}
+
+	//If not, convert to utf-8
+	auto len = strlen(str);
+
+	// Validate the UTF-8 encoding
+	bool valid = true;
+
+	for (size_t i = 0; i < len; i++) {
+		if (str[i] > 0x79 || str[i] < 0) {
+			valid = false;
+			break;
+		}
+	}
+
+	if (valid)
+	{
+		// turns out this is valid UTF-8
+		buffer.assign(str);
+		return buffer;
+	}
+
+	//Skip checks. This function is only allowed for known iso8859-1 encoded strings
+
+	size_t newlen = len;
+	std::unique_ptr<char[]> newstr(new char[newlen]);
+
+	do {
+		auto in_str = str;
+		auto in_size = len;
+		auto out_str = newstr.get();
+		auto out_size = newlen;
+
+		auto iconv = SDL_iconv_open("UTF-8", "ISO-8859-1");
+		auto err = SDL_iconv(iconv, &in_str, &in_size, &out_str, &out_size);
+		SDL_iconv_close(iconv);
+
+		// SDL returns the number of processed character on success;
+		// error codes are (size_t)-1 through -4
+		if (err < (size_t)-100)
+		{
+			// successful re-encoding
+			buffer.assign(newstr.get(), newlen - out_size);
+			return buffer;
+		}
+		else if (err == SDL_ICONV_E2BIG)
+		{
+			// buffer is not big enough, try again with a bigger buffer. Use a rather conservative size
+			// increment since the additional size required is probably pretty small
+			newlen += 10;
+			newstr.reset(new char[newlen]);
+		}
+		else
+		{
+			break;
+		}
+	} while (true);
+
+	return "";
+}
+
+//Converts a string in UTF-8 to the currently used encoding
+//Only legal for strings with correct UTF-8 encoding
+SCP_string utf8_to_current(const char* str)
+{
+	SCP_string buffer;
+	//We are in unicode mode, so just return as is, the string is correct
+	if (Unicode_text_mode) {
+		buffer.assign(str);
+		return buffer;
+	}
+
+	//If not, convert to iso8859-1
+	auto len = strlen(str);
+
+	// Validate the UTF-8 encoding
+	bool valid = true;
+
+	for (size_t i = 0; i < len; i++) {
+		if (str[i] > 0x79 || str[i] < 0) {
+			valid = false;
+			break;
+		}
+	}
+
+	if (valid)
+	{
+		// turns out this is valid ISO8859-1
+		buffer.assign(str);
+		return buffer;
+	}
+
+	//Skip checks. This function is only allowed for known utf-8 encoded strings
+
+	size_t newlen = len;
+	std::unique_ptr<char[]> newstr(new char[newlen]);
+
+	do {
+		auto in_str = str;
+		auto in_size = len;
+		auto out_str = newstr.get();
+		auto out_size = newlen;
+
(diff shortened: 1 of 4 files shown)


code/parse/parselo.cpp
```diff
@@ -2327,11 +2327,156 @@ void coerce_to_utf8(SCP_string &buffer, const char *str)
...
+
+	// Validate the UTF-8 encoding
+	bool valid = true;
+
+	for (size_t i = 0; i < len; i++) {
+		if (str[i] > 0x79 || str[i] < 0) {
+			valid = false;
+			break;
+		}
+	}
```
Goober5000, line 2352: This would be useful to put in its own function.

However, isn't it possible to get the same result with `utf8::find_invalid`?

code/parse/parselo.cpp
```diff
@@ -2327,11 +2327,156 @@ void coerce_to_utf8(SCP_string &buffer, const char *str)
...
+			break;
+		}
+	} while (true);
+
+	return "";
+}
+
+//Converts a string in UTF-8 to the currently used encoding
+//Only legal for strings with correct UTF-8 encoding
+SCP_string utf8_to_current(const char* str)
```
Goober5000, line 2402: There is a lot of duplicate code between this function and the above function.  It would be better to combine them and use function parameters to customize how the strings are handled.

code/parse/parselo.cpp
```diff
@@ -2327,11 +2327,156 @@ void coerce_to_utf8(SCP_string &buffer, const char *str)
 	Warning(LOCATION, "Truncating non-UTF-8 string '%s' to '%s'!\n", str, buffer.c_str());
 }
 
+//Converts a string in ISO 8859-1 to the currently used encoding
+//Only legal for strings with correct ISO 8859-1 encoding
+SCP_string iso8859_1_to_current(const char* str)
```
Goober5000, line 2332: This looks very similar to `coerce_to_utf8`.  Why can't you use that function?

Goober5000 (commented):
This looks useful, but it also looks like half of this PR duplicates code that already exists (and then duplicates the duplicate).

code/parse/parselo.cpp
BMagnu: utf8::find_invalid is not the same as this does here. What find_invalid does, it figure out if the sequence is a correct utf8 string. What this does is check, if the sequence only contains chars between 0 and 0x79. This is because in that special case, a utf8 string and an iso8859-1 string are equal and no change needs to be done. If other chars exist, they can be valid utf8, but not valid iso8859-1. Thus this different check is necessary. But I will change it to it's own function.
As a sidenote: A valid iso8859-1 string that does not fulfill the above check is not valid utf8 (I think), so in that direction the utf-8 check would work, but this more stringent check is definitely needed for the utf8 to iso8859-1 conversion

code/parse/parselo.cpp
BMagnu: Coerce_to_utf8 has two distinct differences:
first, it _guesses_ whether the source encoding is latin1 based on heuristics, and only if it thinks this is the case will it actually convert. As opposed to that, the iso8859_1_to_current function, will assume it gets iso8859-1 encoding and work from that, even if the heuristic may fail.
Secondly, and more importantly, the function converts it's input to _the current_ encoding. Which might be utf-8, but it might also be iso8859-1, as opposed to the coerce function which will always output utf-8.
But code reuse could probably be reduced by combining both to_current functions to a single conversion function that operates with more parameters, and then adapting the coerce function to perform it's checks and then call upon said conversion function.

Goober5000: > But code reuse could probably be reduced by combining both to_current functions to a single conversion function that operates with more parameters, and then adapting the coerce function to perform it's checks and then call upon said conversion function.

Yes, that sound like the best approach.

code/parse/parselo.h
```diff
@@ -256,6 +260,8 @@ extern void read_file_text_from_default(const default_file& file, char *processe
 extern void read_raw_file_text(const char *filename, int mode = CF_TYPE_ANY, char *raw_text = NULL);
 extern void process_raw_file_text(char *processed_text = NULL, char *raw_text = NULL);
 extern void coerce_to_utf8(SCP_string &buffer, const char *src);
+extern const char* get_encoding_string(int encoding);
+extern void convert_encoding(SCP_string& buffer, const char* src, int encoding_src, int encoding_dest = ENCODING_CURRENT);
```
asarium, line 264: These functions are not parse specific and should be moved to `utils/unicode`.

code/parse/parselo.h
```diff
@@ -66,6 +66,10 @@ extern int Token_found_flag;
 #define SEXP_SAVE_MODE				1
 #define SEXP_ERROR_CHECK_MODE		2
 
+#define ENCODING_CURRENT	0
+#define ENCODING_UTF8		1
+#define ENCODING_ISO8859_1	2
```
asarium, line 71: Please use an `enum class` here.

code/parse/parselo.cpp
```diff
@@ -2285,53 +2285,112 @@ void coerce_to_utf8(SCP_string &buffer, const char *str)
...
+
+	size_t newlen = len;
+	std::unique_ptr<char[]> newstr(new char[newlen]);
+
+	do {
+		auto in_str = src;
+		auto in_size = len;
+		auto out_str = newstr.get();
+		auto out_size = newlen;
+
```
asarium, line 2351: 1 empty line is enough :wink:

asarium (changes requested):

asarium (approved):

code/utils/unicode.cpp
```diff
@@ -156,4 +156,87 @@ size_t encoded_size(codepoint_t cp) {
...
+
+	// Validate if no change needs to be done
+	bool valid = true;
+
+	for (size_t i = 0; i < len; i++) {
+		if (src[i] > 0x79 || src[i] < 0) {
+			valid = false;
+			break;
+		}
+	}
```
Goober5000, line 199: This should still be in a separate function.

Goober5000 (changes requested):
Looks better except for one task that remains.

Goober5000 (commented):
Can the code on lines 2232 et seq. be replaced by the new conversion functions?

BMagnu:
> Can the code on lines 2232 et seq. be replaced by the new conversion functions?

I think it _could_, but we'd lose the warning about truncated / failed conversion that is specific about a conversion failure with these files. Which is, I think, the reason that this function wasn't coerce_to_utf8 before as well but rather it's own code

Goober5000:
> Which is, I think, the reason that this function wasn't coerce_to_utf8 before as well but rather it's own code

But the `coerce_to_utf8` function was written *after* that code was written.

Goober5000 (approved):
This looks good :+1:

JohnAFernandez:
If m!m and Goober have approved it, it's probably ready to go.

PR #3264, opened 2021-03-01
Increase max nebula poofs to 32

Baezon:
As part of a collab with @Goober5000 this is just the engine side of the changes, FRED will need to be updated at some point to allow setting of more than the first 6 without manually editing the mission file. The currently used methods for poof assignment uses a bitfield stored in an `int`, so we have 26 bits lying around that can be used without any significant data structure changes. 

Also comes with support for 0 (or not max) poofs, previously not possible, and a less confusing random poof picker (the assertion added should always be true because `neb2_render_poofs()` is the only thing that calls `neb2_get_bitmap()`.

### DIFF
--- a/code/nebula/neb.cpp
+++ b/code/nebula/neb.cpp
@@ -964,10 +964,6 @@ void neb2_render_poofs()
 	if (!(The_mission.flags[Mission::Mission_Flags::Fullneb])) {
 		return;
 	}
-	
-	// no poofs, bail
-	if (!Neb2_poof_flags)
-		return;
 
 	if (Neb2_regen) {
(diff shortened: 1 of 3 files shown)


Goober5000 (approved):
This looks good. :+1:

Baezon:
Lightning doesn't work properly with 0 poofs. Putting it into draft until I can figure it out.

Baezon:
OK, works now, no need to bail with no poofs, still gotta set up the cubes and a -1 bitmap poof is ok the rendering just skips it.

Goober5000 (approved):
New commit looks reasonable.

PR #3266, opened 2021-03-01
Make nebula poof detail settings partially dynamic

### LINKED ISSUE #3250: Nebula poof rendering needs improvement
As documented by Durandal [here](https://www.youtube.com/watch?v=4boRXk2DWY4&t=2s) and [here](https://www.youtube.com/watch?v=gpxrsFG0-CM&feature=youtu.be), nebula poof rendering currently causes some bad issues with visual consistency, as poofs larger than retail will be faded out very quickly as they move across the scene.

The underlying reason are the nebula detail default settings. These set very low (by modern standards) values for when to start fading things out (see neb.cpp:169 and following).

These default values need to be adjusted. Ideally, they'd be made dynamic, taking as reference the actual size of the currently loaded poofs and the current screen aspect ratio.

Similarly, the default value for `break_alpha` can also be substantially lowered to further improve visual consistency.

see also https://github.com/scp-fs2open/fs2open.github.com/issues/2689

The-E:
After loading the nebula poofs, resize the break_x and break_y values so they are equal to the largest poof bitmap in use.

Also reduce the break alpha value by 90% for all detail levels.

This closes #3250

### DIFF
--- a/code/io/keycontrol.cpp
+++ b/code/io/keycontrol.cpp
@@ -692,15 +692,6 @@ void process_debug_keys(int k)
 		case KEY_DEBUGGED + KEY_H:
 			hud_target_toggle_hidden_from_sensors();
 			break;
-
-		case KEY_DEBUGGED + KEY_F: 
-			extern int wacky_scheme;
-			if(wacky_scheme == 3){
-				wacky_scheme = 0;
-			} else {
-				wacky_scheme++;
-			}
-			break;
 		
 		case KEY_DEBUGGED + KEY_ALTED + KEY_F:
 			Framerate_delay += 10;
--- a/code/nebula/neb.cpp
+++ b/code/nebula/neb.cpp
@@ -426,7 +426,7 @@ void neb2_post_level_init()
 	for (auto poof : Neb2_poofs) {
 		int bm_height;
 		bm_get_info(poof, nullptr, &bm_height);
-		height = MAX(height, static_cast<float>(bm_height));
+		height = MAX(height, static_cast<float>(bm_height * 2));
 	}
 
 	for (auto& detail : Neb2_detail) {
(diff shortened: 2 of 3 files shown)


Baezon:
Since `break_x` and `break_y` are dynamic now, can we delete them from `Neb2_detail`? To avoid the confusion of these defaults values not being used. I notice you write them back to the details, but they could just be globals of this .cpp right?

code/nebula/neb.cpp
```diff
@@ -419,6 +419,18 @@ void neb2_post_level_init()
 		}
 	}
 
+	float height = Neb2_detail[MAX_DETAIL_LEVEL-1].break_y;
+	for (auto poof : Neb2_poofs) {
+		int bm_height;
+		bm_get_info(poof, nullptr, &bm_height);
+		height = MAX(height, static_cast<float>(bm_height * 2));
```
Baezon, line 426: I'm not sure I understand the connection to the bitmap's size? Since regardless of the resolution of the image it'll be displayed at whatever size the poof should be anyway. Should we not be basing break_y/x on the *screen's* size? Since 'the number of pixels offscreen' is going to count a lot less nowadays than it used to.

Baezon (changes requested):

code/nebula/neb.cpp
The-E: That's a good point.

code/io/keycontrol.cpp
```diff
@@ -692,15 +692,6 @@ void process_debug_keys(int k)
...
 			break;
-
-		case KEY_DEBUGGED + KEY_F: 
-			extern int wacky_scheme;
-			if(wacky_scheme == 3){
-				wacky_scheme = 0;
-			} else {
-				wacky_scheme++;
-			}
-			break;
```
Baezon, line 703: Whoops, did git get confused?

Baezon (approved):

code/io/keycontrol.cpp
The-E: Yeah, whatever it used to do has been lost to time. I stumbled across it when I removed the console command to alter the break_* values, and when it became clear that it did nothing except get cycled by a cheat.....

Issue #3102, opened 2021-01-09
FR: Replace-Texture SEXP


MjnMixael:
What it says on the tin. A sexp that can be used to replace a specific texture just like can be done via tables or the ships dialog in FRED.

Tnadz:
I second this initiative.  We've seen a few instances where a texture swap can make a mission incredibly dynamic.  Would love it if a SEXP master was able to work something out with this.

wookieejedi:
Awesome work with @BMagnu's addition. Closing.

PR #3268, opened 2021-03-02
Add replace-texture SEXP

BMagnu:
Adds a SEXP to replace a texture of a specific ship with another texture.
Should implement #3102.


code/parse/sexp.cpp
```diff
@@ -34771,6 +34862,14 @@ SCP_vector<sexp_help_struct> Sexp_help = {
...
 		"\t3:\tThe XSTR index. If set to -1 then the default value will be used\r\n"
 	},
+
+	{ OP_REPLACE_TEXTURE, "replace-texture\r\n"
+		"\tChanges a texture of a ship to a different texture, similar to the FRED texture replace.\r\n"
+		"Takes 3 arguments...\r\n"
+		"\t1: Name of the Ship.\r\n"
+		"\t2: Name of the texture to be replaced.\r\n"
+		"\t3: Name of the texture to be changed to.\r\n"
+	},
```
Goober5000, line 34872: To enhance usability I suggest changing the arguments as follows:
```
		"Takes 3 or more arguments...\r\n"
		"\t1: Name of the texture to be replaced.\r\n"
		"\t2: Name of the texture to be changed to.\r\n"
		"\tRest: Name of the ship or wing.\r\n"
```
This would be more useful if the FREDder had a number of ships or wings of identical models that all required the same replacement.

Goober5000 (changes requested):

code/parse/sexp.cpp
```diff
@@ -28959,6 +29066,14 @@ int query_operator_argument_type(int op, int argnum)
 		case OP_SET_MOTION_DEBRIS:
 			return OPF_BOOL;
 
+		case OP_REPLACE_TEXTURE:
+			if (argnum == 0)
+				return OPF_STRING;
+			if (argnum == 1)
+				return OPF_STRING;
```
Goober5000, line 29073: You could check `argnum == 0 || argnum == 1` but that's a stylistic choice.

Goober5000 (changes requested):
Looks good aside from some comments.  I assume you've tested this in-game?

BMagnu:
Yes, I have tested the replace in game, both with not-yet-arrived an present ships and wings.

Goober5000 (approved):
Good to go. :+1:

Issue #3267, opened 2021-03-02
Multiple annotations in the same event break mission parsing


naomimyselfandi:
With the new FRED annotations feature, it's possible to save a mission that won't parse correctly. With really bad luck, it's possible to save a mission in a state that FRED can't even open it.

There are two reliable ways to reproduce this bug:

1. Create an event (a generic `( when ( true ) ( do-nothing ) )` will suffice). Annotate one of the operators, then annotate the event name.
2. Create an event and annotate three of its nodes.

----

This parses correctly:

```
$Formula: ( when ( false ) ( do-nothing ) )
+Name: Continue campaign
+Repeat Count: 1
+Interval: 1
$Annotations Start
+Comment: TODO: This comes true when the player passes the mission.
$end_multi_text
+Background Color: 255, 0, 0
$Annotations End
```

So does this:

```
$Formula: ( when ( false ) ( do-nothing ) )
+Name: Continue campaign
+Repeat Count: 1
+Interval: 1
$Annotations Start
+Comment: TODO: This comes true when the player passes the mission.
$end_multi_text
+Background Color: 255, 0, 0
+Path: 0, 0
$Annotations End
```

This, however, does not (pay close attention to the placement of the `+Path:`), and it results from case 1 up above.

```
$Formula: ( when ( false ) ( do-nothing ) )
+Name: Continue campaign
+Repeat Count: 1
+Interval: 1
$Annotations Start
+Comment: TODO: This comes true when the player passes the mission.
$end_multi_text
+Path: 0, 0
+Background Color: 255, 0, 0
$Annotations End
```

Here's the error:

```
Error: C:\Users\PC\AppData\Local\Knossos\library\FS2\BtA-2.0.0\bta_core\data\missions\0_██████Template.fs2(line 617):
Error: Missing required token: [$Annotations End]. Found [Background Color: 255, 0, 0] instead.

File: parselo.cpp
Line: 309

ntdll.dll! NtWaitForSingleObject + 20 bytes
KERNELBASE.dll! WaitForSingleObjectEx + 147 bytes
fred2_open_21_1_0_20210301_66b7b011c_x64_SSE2.exe! <no symbol>
```

This is on FSO 21.1.0-20210301; I've only tested it in FRED.

Also, the error message really does say `Background Color:`, not `+Background Color:`.

Goober5000:
This is a very good and detailed report, thank you.  I'll take a look.

Goober5000:
[...]

Issue #3265, opened 2021-03-01
Assert: "count < max_floats"


naomimyselfandi:
From @MjnMixael:

> BtA 1.5.8 gets an this assert on FSO builds from Feb 1 2021 or newer. Suspect a change in #3079 causes the issue. Unclear what data FSO is tripping over here, but the mod worked before this change.

The assert is in parselo.cpp, line 3179.

Goober5000:
I'll take a look.

naomimyselfandi:
@Goober5000 Mjn eventually tracked it down to a typo in one of our .tbl files - `0..5` instead of `0.5`.

Goober5000:
Oh ok, thanks for the update.  I'll call this Not-A-Bug then.

Goober5000:
On second thought, I'll keep it open to see if I can report the error in a clearer way.

naomimyselfandi:
Well, philosophically, I don't think asserts should get tripped by a data issue. Practically, it meant we didn't get a filename or line number. (BTA is about as large as you'd expect, so if Mjn hadn't been familiar with #3079, I don't know how we'd have _ever_ found it.)

Goober5000:
Agreed, asserts should only be used to catch programming mistakes.  This should be caught with `error_display` like the rest of the parsing validation.

Goober5000:
#3269 will prevent this error, but I will make another PR to actually fix the error reporting here.

naomimyselfandi:
> #3269 will prevent this error (by catching and reporting a different error), but I will make another PR to actually fix the error reporting here.

Just a heads-up that this was from a .tbl file (I think it was species_defs.tbl), not a FRED annotation. It sounds like your fix will address both - I just wanted to make sure that was clear because the title of your new issue mentions annotations in particular.

Goober5000:
Yeah, I loaded up the error and and tested it directly from the mod - it was caused by hud_guages.tbl.  The two issues are both related to number parsing, which is why the new PR mentions both issues.

Goober5000:
Fixed by #3269 and #3271.

PR #3269, opened 2021-03-02
Various annotation fixes

Goober5000:
Make some improvements to the annotation code and the parsing code:

1) A single plus sign or single minus sign is not a number, and this is what was causing the parser to treat the + on the following line as another number in the Path list.  This is the major fix for #3267.
2) If there is an unexpected character after a number, the number should be considered a bad token, versus just reading as much of the number as we can and considering the result valid.  This is a partial fix for #3265.
3) If the annotation contains unmodified information for a certain value (such as background color), don't render the unmodified values.
4) There are also a few minor fixes here, such as checking for all three possible annotation strings and slightly cleaning up the reporting of parse errors.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -5082,7 +5082,7 @@ void parse_event(mission * /*pm*/)
 	if (optional_string("$Annotations Start")) {
 		// annotations are only used in FRED
 		if (Fred_running) {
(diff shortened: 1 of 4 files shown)


JohnAFernandez:
I was going to approve this as a fairly important bugfix, but I honestly do not understand the changes.

Baezon (approved):

PR #3270, opened 2021-03-03
Enable multi for two sexps

Goober5000:
This is #3262, but for the multi branch.


JohnAFernandez (approved):

JohnAFernandez:
This basically allows us to do testing on his changes in the most up to date multi environment.

The use of "uninitialized" variables is a false positive from the multi macros.

PR #3271, opened 2021-03-03
rewrite stuffing of lists

Goober5000:
Consolidate all the functions which parse lists of data.  They're almost the same except for the data itself, so use templates.  As for the templated function which handles the list, boost the robustness by adding additional checking and error messages.

This completes the fix for #3265.

~~In draft form until I finish testing.~~  Should be ready now, but it would be better if #3269 were merged first because there are some changes that very slightly overlap.

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -2634,67 +2634,16 @@ int stuff_int_optional(int *i)
 	return stuff_int(i, true);
 }
 
-int stuff_int_or_variable (int &i, bool positive_value = false);
-int stuff_int_or_variable (int *ilp, int count, bool positive_value = false);
-
-// Stuffs an int value or the value of a number variable. Returns the index of the variable or NOT_SET_BY_SEXP_VARIABLE.
-int stuff_int_or_variable (int &i, bool positive_value)
-{
-	int index = NOT_SET_BY_SEXP_VARIABLE;
-
-	if (*Mp == '@')
-	{
-		Mp++;
-		int value = -1;
-		char str[128];
-		stuff_string(str, F_NAME, sizeof(str));
-
-		index = get_index_sexp_variable_name(str);
-
-		if (index > -1 && index < MAX_SEXP_VARIABLES)
-		{
-			if (Sexp_variables[index].type & SEXP_VARIABLE_NUMBER)
-			{
-				value = atoi(Sexp_variables[index].text);
-			}
-			else
-			{
-				Error(LOCATION, "Invalid variable type \"%s\" found in mission. Variable must be a number variable!", str);
-			}
-		}
-		else
-		{
-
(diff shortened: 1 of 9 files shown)


code/parse/parselo.cpp
```diff
@@ -2763,161 +2709,132 @@ void stuff_boolean(bool *b, bool a_to_eol)
...
-	int num_strings = stuff_string_list(slp, defs_size);
-	int i, j;
+	SCP_vector<SCP_string> slp;
+	stuff_string_list(slp);
 
-	for(i = 0; i < num_strings; i++)
+	for (auto &str : slp)
 	{
-		for(j = 0; j < defs_size; j++)
+		for (size_t j = 0; j < defs_size; j++)
```
ngld, line 2783: Small nitpick: Wouldn't it make sense to rename `j` to `i` here since the previous `i` variable is gone?

code/parse/parselo.h
```diff
@@ -144,26 +144,26 @@ extern int stuff_long(long *l, bool optional = false);
...
-    char(*slp)[NAME_LENGTH] = (char(*)[32])new char[n_defs*NAME_LENGTH];
-    int num_strings = stuff_string_list(slp, (int)n_defs);
+	SCP_vector<SCP_string> slp;
+    stuff_string_list(slp);
 
-    for (auto i = 0; i < num_strings; i++)
+	for (auto &item : slp)
     {
         bool string_parsed = false;
         for (size_t j = 0; j < n_defs; j++)
```
ngld, line 164: Same as above

code/parse/parselo.cpp
```diff
@@ -2763,161 +2709,132 @@ void stuff_boolean(bool *b, bool a_to_eol)
...
-		//Assert ( *Mp == '\"' );					// should always be enclosed in quotes
 
-		if (count < max_strings) {
-			get_string( slp[count++] );
-		} else {
-			char trash[NAME_LENGTH];
-			get_string( trash );
-		}
-		ignore_white_space();
+		strcpy(slp[i], list[i].c_str());
```
ngld, line 2834: Why use `strcpy` instead of `strcpy_s`?

ngld (approved):
Looks good to me. I only found a few nitpicks. Feel free to ignore the i/j changes since they're really minor.

code/parse/parselo.h
Goober5000: Same as above

code/parse/parselo.cpp
Goober5000: Oops :eyes:  I'll do that

code/parse/parselo.cpp
ngld: That's fair.

PR #3272, opened 2021-03-03
Translateable Asteroid Impact Time

NovachenFS2:
The english "Impact:" is translateable for targetable bombs, but not for Asteroids in escort missions through asteroid fields. This PR fixes that. Both "Impact:" entries are now translateable through the same string in strings.tbl.


JohnAFernandez (approved):
Glad to see this was easy to find and fix. :)

PR #3273, opened 2021-03-03
Translateable Asteroid Name in Asteroid Fields

NovachenFS2:
Never noticed this, as Asteroid translates to Asteroid in German 😆.
However after i switched to another language i found out, that the English name is used in the target monitor aswell in the target brackets.
This applies to the generated Asteroids you can target in Asteroid Fields, which is widely used in any kind of Escort Missions through them.
So this is actually a follow up to #3272.

This PR change the Asteroid's name to a translateable one by the strings.tbl.
I decided to use the already established ID 431, as that one is used for the displayed name after you clicked on the Asteroid Field icon image in mission briefings.
I can not think of a scenario, where this name should be any different.

### DIFF
--- a/code/hud/hudtargetbox.cpp
+++ b/code/hud/hudtargetbox.cpp
@@ -1160,7 +1160,7 @@ void HudGaugeTargetBox::renderTargetAsteroid(object *target_objp)
 		case ASTEROID_TYPE_SMALL:
 		case ASTEROID_TYPE_MEDIUM:
 		case ASTEROID_TYPE_LARGE:
-			strcpy_s(hud_name, NOX("asteroid"));
+			strcpy_s(hud_name, XSTR("Asteroid", 431));
 			break;
 
 		default:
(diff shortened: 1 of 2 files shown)


code/hud/hudtargetbox.cpp
```diff
@@ -1160,7 +1160,7 @@ void HudGaugeTargetBox::renderTargetAsteroid(object *target_objp)
 		case ASTEROID_TYPE_SMALL:
 		case ASTEROID_TYPE_MEDIUM:
 		case ASTEROID_TYPE_LARGE:
-			strcpy_s(hud_name, NOX("asteroid"));
+			strcpy_s(hud_name, XSTR("Asteroid", 431));
```
Goober5000, line 1163: In English, the strings.tbl entry has a lowercase 'a'.  The NOX entry also has a lowercase 'a', so these match and the string should be changed to lowercase.

I was going to say that the hudbrackets.cpp change does not match and we may have a problem, *except* that the "debris" entry 348 is also lowercase, not uppercase as shown in the file.  So change both "asteroid" and "debris" to lowercase.

Goober5000 (changes requested):

Goober5000 (approved):
Looks good.

PR #3274, opened 2021-03-04
Add shooter damaging weapons

Baezon:
Weapons, beams and their descendants can't normally damage the ship that fired them. Now they can, if the modder wants them to.

The change to `PLAYER_KILLED_SELF` appears to have been a minor oversight that prevents the non-specific player self kill (`PLAYER_FLAGS_KILLED_SELF_UNKNOWN `) from actually giving a "you rekt urself" death message, since it checks `PLAYER_KILLED_SELF` first.


code/object/object_flags.h
```diff
@@ -27,6 +27,7 @@ namespace Object {
 		Marked,					// Object is marked (Fred).  Can be reused in FreeSpace for anything that won't be used by Fred.
 		Temp_marked,			// Temporarily marked (Fred).
 		Hidden,					// Object is hidden (not shown) and can't be manipulated
+		Collides_with_parent,	// Asteroth
```
JohnAFernandez, line 30: My only issue with this is that this needs a comment that it should only be set for weapons.   My initial reading of the changes to the collision code made me think that there was a mistake because a ship with the flag would have all of its weapons potentially collide with it.

JohnAFernandez (changes requested):

code/object/object_flags.h
Baezon: This was my fear as well, but I never observed it during my testing ¯\_(ツ)_/¯

Perhaps with different/more elaborate fire point set ups this could happen, but the sad fact is that if it happens, it would be very difficult to elegantly solve in the code, and by far the simplest solution would be to simply slightly move the firepoints away.

JohnAFernandez (approved):

PR #3275, opened 2021-03-06
addendum to u_offset

Goober5000:
This improves PR #3257 in two ways:
1) ~~If a beam has a `u_offset_local`, it will no longer step on the global `u_offset` every time it's fired~~
2) The frametime is updated after the beam fires, not before (so that the first frame will always be at offset zero)

### DIFF
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -1276,10 +1276,10 @@ void beam_render(beam *b, float u_offset)
 
 	bwi = &Weapon_info[b->weapon_info_index].b_info;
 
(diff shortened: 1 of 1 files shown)


JohnAFernandez (approved):

Baezon:
I don't think `u_offset_local` actually steps on the global `u_offset`, since its sends in a copy as in an argument to `beam_render()`?

Goober5000:
> I don't think `u_offset_local` actually steps on the global `u_offset`, since its sends in a copy in an argument to `beam_render()`?

Oh, you're right. :thonk:  Good catch.

PR #3276, opened 2021-03-06
add small fix from SEXP containers PR

Goober5000:
While skimming the differencs in @jg18's branch of PR #3253...
https://github.com/jg18/fs2open.github.com/tree/feature/sexp-containers-revised

...I noticed a TODO that is almost certainly a genuine bugfix.  As such, it deserves its own PR.  The line in question was last modified in 2012 by @Karajorma, and the bugfix should test for the special-arg in the entire actions tree rather than just the first action.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -9491,7 +9491,7 @@ int eval_when(int n, int when_op_num)
 
 		// if the mod.tbl setting is in effect we want to each evaluate all the SEXPs for 
 		// each argument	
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
Nice catch.

Goober5000:
Well, I *think* this is a catch, but I'd like @Karajorma's opinion (since he wrote it) or @MageKing17's opinion (since he noticed a similar bug in the past) before merging it.

Goober5000:
Confirmed by @Karajorma as an actual bug fix.  Merging.

PR #3277, opened 2021-03-06
Turrets on movable submodels

Goober5000:
The code always assumed that turrets were part of a fixed reference frame with respect to the main model.  This updates the relevant turret calculation code (mainly FOV and turret normals) so that turrets now use the appropriate submodel functions.  As a result, turrets now work on movable submodels.

Note that most of the `turret_matrix` code has been ripped out because it assumes a fixed reference frame.  This isn't a problem because most of the code can be easily converted to use `turret_norm` via the submodel functions.  The only exception is the `Turret_alt_math` flag, whose code is rather inscrutable.  For those cases, I just regenerated a new turret matrix on the fly.  In the future, someone who is more familiar with that code should rewrite it to use `turret_norm`.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -159,8 +159,15 @@ bool is_object_radius_in_turret_fov(object *objp, ship_subsys *ss, vec3d *tvec, 
 		}
 
 		if (tp->flags[Model::Subsystem_Flags::Turret_alt_math]) {
+			// Since we no longer maintain world_to_turret_matrix, regenerate it here.
+			vec3d turret_norm;
+			matrix turret_matrix, world_to_turret_matrix;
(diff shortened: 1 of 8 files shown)


JohnAFernandez (approved):
You made this look easy.  I know it took a lot of background knowledge to get this to work, but it's such a clean patch.  I read through everything and the vector math makes sense.  So I'll approve, and I guess we'll wait for m!m or Asteroth to give it a second look.

Goober5000:
Thanks!  The heavy lifting was done pre-21.0.  Because of the rewritten code, this patch turned out to be pretty straightforward, and future patches (including submodel translation, which I'm working on now) should also be pretty straightforward.

Baezon (approved):
Indeed surprisingly simple! Looks good to me, when I have the time ill try to take a look at the alt math thing.

PR #3278, opened 2021-03-08
Switch to exponential fog and other nebula visibility improvements

Baezon:
After thinking about this for awhile, I really do think that straight switching to exponential fog over linear fog is the best solution for pretty much everyone, not least of which because it would be a crime to not get to see this in retail.

A comparison is here if you want to see the exact visual nature of the changes.:
https://imgur.com/a/DjooD9w
Those scenes are considered *equivalent* for the same fog near and far values! So a mission which looked like those linear pictures will from now on look like those exponential ones.

Additionally, pretty much everything I could think of now has the ability to be nebula fogged with some optional flags (because some mods rely on some of these things always being visible).

Some things that didn't work out or I couldn't figure out: the main fog equation says its " the same formula that is also used by the main shader" which is true, in `main-v.sdr` but in trying to update it to use the new system it crashed during shader compilation in some obscure place way outside of FSO code somehow, and in the only places I could find that this was used, 3D shockwaves and modeled thrusters, it would look like crap because fully fogged shows up as a distinctly brighter shade of the nebula BG bleeding through as an obvious silhouette (this is currently the case for modeled thrusters, but not for shockwaves because they have fog disabled currently), so I'm probably not going to bother because those are pretty niche situations anyway.


Goober5000 (commented):
The screenshots and video are very impressive, and I agree that this effect looks much better than the original fogging method, but there is one obstacle.  Many missions set the appearance of a certain ship to be triggered by a certain mission event, and if the fogging distance is altered, the timing could also get out of sync.  Examples are the appearance of the Trinity in Mystery of the Trinity; of the Knossos in Straight, No Chaser; of the Sathanas and Volition Bravos in Through the Looking Glass (Derelict); and there are probably others.

Kanangus:
> 
> 
> The screenshots and video are very impressive, and I agree that this effect looks much better than the original fogging method, but there is one obstacle. Many missions set the appearance of a certain ship to be triggered by a certain mission event, and if the fogging distance is altered, the timing could also get out of sync. Examples are the appearance of the Trinity in Mystery of the Trinity; of the Knossos in Straight, No Chaser; of the Sathanas and Volition Bravos in Through the Looking Glass (Derelict); and there are probably others.
> 
> So my question is: does this new effect change the timing of ship appearances?

Could we implement a toggle for legacy & exponential render modes per mission file on fog rendering?

Baezon:
I didn't notice it myself in mystery of the trinity, I didn't even know it spawned in. The behavior of the mission itself will be unchanged, but I suppose it is possible that if the ship is spawned in it may be noticeable, but this would only happen for unusually close spawn distances, certainly well within sensor range.

Switching between both modes is possible but I would really like to avoid it if possible, the only possible downsides and very niche and within that very minor as well, and are far outweighed by the benefits imo. And of course making it opt in would make it completely unusable for retail and very unlikely to be used for older nebula-heavy mods.

Goober5000:
None of the ships I mentioned use arrival cues or spawn tricks.  They're just placed so that if the player is matching speed with his wingmen, they will come out of the fog at precisely the right moment in the dialogue.

Baezon:
Ah ok, I understand, in which case yes it will affect the distance at which you can first see things, if only vaguely at that distance. Due to the differences in these falloff curves it is unavoidable to see farther in the long-range (though not with much detail) without obliterating the visibility of the mid-ranges.

Interestingly, in mystery of the trinity and straight no chaser I have *never* been able to see the trinity or knossos when it was called out and I suspect that if I crunch the numbers they are in fact still completely invisible when they do. In mystery of the trinity with the new fog it is only vaguely discernable when called out, at least in the times I tested it.

Baezon:
Ok, I'm noticing the current fogging is way more intense than it used to be. I have no idea what caused it or when, but apparently original fs2 fog is like 2 or 3 times farther than current FSO.

Baezon:
Ok fog figured out, doesn't really affect this PR.

Anyway, I think the distance ships become visible still fits very nicely with the original, if message timings are to be believed, it's just less "becomes visible" now more than it is "becomes discernable".

Here's Mystery of the Trinity and Through the Looking Glass, as soon as the relevant messages are triggered:

screen0143

screen0144

In both cases I didn't notice them until they were called out, but I *could* see them beforehand a few seconds if I knew exactly where to look and knew it was coming.

Goober5000:
That's good enough for me.  Happy to approve.

Goober5000 (approved):

PR #3284, opened 2021-03-11
Fix probability distribution of vm_vec_random_in_circle/sphere

Baezon:
Randomly getting a point on a circle/sphere edge and scaling inward by `frand()` is *not* a uniform distribution, as the comments say, this will bias the distribution towards the center, so introduce a factor to combat this. The only thing that should be noticeably affected that isn't just visual is beam aim 'jitter', since this is retail an optional parameter is added to to maintain this 'bias' (and also for trails because they look better that way)

As an aside, randomizing a vector's coordinates and normalizing is *also* not a uniform distribution over a spherical surface (ironically that method was introduced in the same PR that introduced the proper method for random_cone, which has been used for circle and sphere now).

screen0138

screen0139

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -2261,7 +2261,7 @@ void vm_vec_random_in_circle(vec3d *out, const vec3d *in, const matrix *orient, 
 	vec3d temp;
 	float scalar = frand();
 
-	// sqrt because scaling inward increases the probability desnity by the square of its distance towards the center
+	// sqrt because scaling inward increases the probability density by the square of its proximity towards the center
 	if (!bias_towards_center)
 		scalar = sqrtf(scalar);
 
@@ -2274,7 +2274,7 @@ void vm_vec_random_in_circle(vec3d *out, const vec3d *in, const matrix *orient, 
 
 // given a start vector and a radius, generate a point in a spherical volume
 // if on_surface is true, the point will be on the surface of the sphere
-void vm_vec_random_in_sphere(vec3d *out, const vec3d *in, float radius, bool on_surface)
+void vm_vec_random_in_sphere(vec3d *out, const vec3d *in, float radius, bool on_surface, bool bias_towards_center)
 {
 	vec3d temp;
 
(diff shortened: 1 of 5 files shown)


code/math/vecmat.cpp
```diff
@@ -2256,15 +2256,20 @@ void vm_vec_random_cone(vec3d *out, const vec3d *in, float min_angle, float max_
...
 // if on_edge is true, the point will be on the edge of the circle
-void vm_vec_random_in_circle(vec3d *out, const vec3d *in, const matrix *orient, float radius, bool on_edge)
+void vm_vec_random_in_circle(vec3d *out, const vec3d *in, const matrix *orient, float radius, bool on_edge, bool bias_towards_center)
 {
 	vec3d temp;
+	float scalar = frand();
 
-	// point somewhere in the plane
-	vm_vec_scale_add(&temp, in, &orient->vec.rvec, on_edge ? radius : frand_range(0.0f, radius));
+	// sqrt because scaling inward increases the probability desnity by the square of its distance towards the center
```
Goober5000, line 2264: density

code/math/vecmat.cpp
```diff
@@ -2273,13 +2278,12 @@ void vm_vec_random_in_sphere(vec3d *out, const vec3d *in, float radius, bool on_
...
-	util::UniformFloatRange coords(-1.0f,1.0f);
-	vm_vec_make( &temp, coords.next(), coords.next(), coords.next() );
-	vm_vec_normalize(&temp);
+	float z = util::UniformFloatRange(-1, 1).next(); // Take a 2-sphere slice
+	float phi = util::UniformFloatRange(0.0f, PI2).next();
+	vm_vec_make(&temp, sqrtf(1.0f - z * z) * cosf(phi), sqrtf(1.0f - z * z) * sinf(phi), z); // Using the z-vec as the starting point
 
-	// We then add the scaled result to the initial position to get the final position
-	vm_vec_scale_add(out, in, &temp, on_surface ? radius : util::UniformFloatRange(0.0f, radius).next());
+	// cube root because scaling inward increases the probability desnity by the cube of its distance towards the center
```
Goober5000, line 2285: density

code/math/vecmat.cpp
```diff
@@ -2273,13 +2278,12 @@ void vm_vec_random_in_sphere(vec3d *out, const vec3d *in, float radius, bool on_
...
+	vm_vec_scale_add(out, in, &temp, on_surface ? radius : powf(util::UniformFloatRange(0.0f, 1.0f).next(), 0.333f) * radius);
```
Goober5000, line 2286: It would be nice to have a `bias_towards_center` parameter just like the companion function

code/weapon/beam.cpp
```diff
@@ -2437,7 +2437,7 @@ void beam_jitter_aim(beam *b, float aim)
 	vm_vector_2_matrix(&m, &forward, NULL, NULL);
 
 	// get a vector on the circle - this should appear to be pretty random
```
Goober5000, line 2439: Edit the comment to clarify it's random with a bias.  I know the `true` parameter should suffice but people might read just the comment and not look at that.

Goober5000 (changes requested):
A few thoughts

Goober5000 (changes requested):
Looks good.  One last request: Could you enhance the `randomInCircle` and `randomInSphere` scripting functions to use the new optional parameter?

Baezon:
Hmm, I suppose it should default to true for better backwards compatability?

Goober5000:
> Hmm, I suppose it should default to true for better backwards compatability?

Yup.

Goober5000:
> Hmm, I suppose it should default to true for better backwards compatability?

Defaulting to false is fine, and I think it is better for future proofing.  Just make sure the scripting functions use the compatible default and you should be okay.

Baezon:
As in, both the internal and scripting function should default to false (but allow for true if you want)?

Goober5000:
> As in, both the internal and scripting function should default to false (but allow for true if you want)?

I was thinking that the internal function should default to no biasing, but the scripting function should default to biasing to remain compatible with previous scripts.

Goober5000 (approved):
Very nice. :+1:

PR #3291, opened 2021-03-14
Peoperly initialize alpha decay exponent for contrails

Baezon:
Basically the same problem as #3255. I've run out of excuses, but I ran through weapons, afterburner trails and contrails with and without the new fields and everything seems to be set properly.


Goober5000 (approved):
Aha.

PR #3296, opened 2021-03-14
LZ4 Compressed files Support

Shivansps:
There is some old commits on this branch that i reverted before starting working on the compressed files.

Thread for context: https://www.hard-light.net/forums/index.php?topic=97436.0

This is to add compressed file support so the files that are inside the vps or in the data folder can be compressed to use less disk space and with HDDs can even improve loading times.


code/cfile/cfilecompression.cpp
```diff
@@ -0,0 +1,219 @@
...
+	auto fSize = fread(&cf->size, 4, 1, cf->fp);
+	auto fBsize = fread(&cf->compression_info.block_size,4, 1, cf->fp);
+
+	#if !defined(NDEBUG)
+	if (cf->compression_info.numOffsets <= 0 || cf->size <= 0 || cf->compression_info.block_size <= 0 || fNumoffsets < 1 || fSize < 1 || fBsize < 1)
+	{
+		//mprintf(("\n(CI) File: %s . Something went wrong reading the setup data, file is possibly in the wrong format.\n", cf->original_filename.c_str(), cf->compression_info.numOffsets, cf->compression_info.block_size));
+		Assertion(cf->compression_info.numOffsets > 0, "Invalid number of offsets, compressed file is possibly in the wrong format.");
+		Assertion(cf->size > 4, "Invalid filesize, compressed file is possibly in the wrong format.");
+		Assertion(cf->compression_info.block_size > 16, "Invalid block size, compressed file is possibly in the wrong format.");
```
ngld, line 139: This looks like `Assertion()`s for `fNumoffsets`, `fSize` and `fBsize` are missing.

code/cfile/cfilecompression.cpp
```diff
@@ -0,0 +1,219 @@
...
+}
+
+void lz41_load_offsets(CFILE* cf)
+{
+	int maxBlocks = (int)cf->size + 8 / cf->compression_info.block_size;
+	cf->compression_info.offsets = (int*)malloc(maxBlocks);
+	int block;
+	int* offsetsPtr = cf->compression_info.offsets;
+
+	fso_fseek(cf, (-4 * (cf->compression_info.numOffsets + 1)) - (sizeof(int)*2), SEEK_END);
```
ngld, line 155: Where does `-4` come from? Shouldn't it be `-sizeof(int)`?

code/cfile/cfilecompression.cpp
```diff
@@ -0,0 +1,219 @@
...
+			int cmpBytes = cf->compression_info.offsets[currentBlock + 1] - cf->compression_info.offsets[currentBlock];
+			auto bytes_read = fread(cmpBuf, cmpBytes, 1, cf->fp);
+			Assertion(bytes_read == 1, "Error reading from compressed file.");
+
+			const int decBytes = LZ4_decompress_safe_continue(lz4StreamDecode, cmpBuf, cf->compression_info.decoderBuffer, cmpBytes, cf->compression_info.block_size);
+			if (decBytes <= 0)
+				return (size_t)LZ41_DECOMPRESSION_ERROR;
+
+			/* Write out the part of the data we care about */
+			size_t blockLength = ((length) < ((decBytes - offset)) ? (length) : ((decBytes - offset)));
```
ngld, line 199: You can strip a few parenthesis here to make the expression easier to read:
`(length < (decBytes - offset) ? length : (decBytes - offset));`

code/cfile/cfilecompression.cpp
```diff
@@ -0,0 +1,219 @@
...
+		{
+			char* cmpBuf = (char*)malloc(LZ4_compressBound(cf->compression_info.block_size));
+			/* The difference in offsets is the size of the block */
+			int cmpBytes = cf->compression_info.offsets[currentBlock + 1] - cf->compression_info.offsets[currentBlock];
+			auto bytes_read = fread(cmpBuf, cmpBytes, 1, cf->fp);
+			Assertion(bytes_read == 1, "Error reading from compressed file.");
+
+			const int decBytes = LZ4_decompress_safe_continue(lz4StreamDecode, cmpBuf, cf->compression_info.decoderBuffer, cmpBytes, cf->compression_info.block_size);
+			if (decBytes <= 0)
+				return (size_t)LZ41_DECOMPRESSION_ERROR;
```
ngld, line 196: You're leaking `cmpBuf` here.

code/cfile/cfilecompression.cpp
```diff
@@ -0,0 +1,219 @@
...
+		fso_fseek(cf, cf->compression_info.offsets[currentBlock], SEEK_SET);
+	
+	offset = offset % cf->compression_info.block_size;
+
+	/* Start decoding */
+	for (; currentBlock < endBlock; ++currentBlock)
+	{
+		if (cf->compression_info.lastDecBlockPos != cf->compression_info.offsets[currentBlock])
+		{
+			char* cmpBuf = (char*)malloc(LZ4_compressBound(cf->compression_info.block_size));
```
ngld, line 188: This looks like the size of `cmpBuf` stays the same for all blocks. If you allocate `cmpBuf` outside the `for()` loop, you can save some overhead when multiple blocks are read.

lib/lz4/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+	IF(FSO_COPY_EXE AND BUILD_SHARED_LIBS)
+		INSTALL(
+			TARGETS lz4
+			RUNTIME DESTINATION ${BINARY_DESTINATION}
+			LIBRARY DESTINATION ${BINARY_DESTINATION}
+			)
+	ENDIF(FSO_COPY_EXE AND BUILD_SHARED_LIBS)
+
+	SET(LZ4_LIBS LZ4 CACHE INTERNAL "lz4 library" FORCE)
```
ngld, line 46: You set `${LZ4_LIBS}` here but don't actually use it anywhere. Either change `TARGET_LINK_LIBRARIES(code PUBLIC lz4)` in `code/CMakeLists.txt` to use this variable or remove this line.

lib/lz4/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+ELSE(FSO_BUILD_INCLUDED_LIBS OR NOT LZ4_FOUND)
+	INCLUDE(util)
+
+	ADD_IMPORTED_LIB(lz4 ${LZ4_INCLUDE_DIR} ${LZ4_LIBRARIES})
+
+	SET(LZ4_LIBS lz4 CACHE INTERNAL "LZ4 library" FORCE)
```
ngld, line 52: dito

code/cfile/cfilearchive.h
```diff
@@ -23,6 +22,17 @@
...
 
+struct COMPRESSION_INFO {
+	int header = 0;
+	size_t compressed_size = 0;
+	int block_size = 0;
+	int numOffsets = 0;
+	int* offsets = nullptr;
+	char* decoderBuffer = nullptr;
+	int lastDecBlockPos = 0;
+	int lastDecBytes = 0;
```
ngld, line 33: You should probably decide on a naming scheme and stick with it. Please don't mix `underscore_names` and `camelCaseNames`. To stay consistent with existing code, I'd prefer underscores.

code/cfile/cfilecompression.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+	int maxBlocks = (int)cf->size + 8 / cf->compression_info.block_size;
+	cf->compression_info.offsets = (int*)malloc(maxBlocks);
+	int block;
+	int* offsetsPtr = cf->compression_info.offsets;
+
+	fso_fseek(cf, (-4 * (cf->compression_info.numOffsets + 1)) - (sizeof(int)*2), SEEK_END);
+	for (block = 0; block <= cf->compression_info.numOffsets; ++block)
+	{
+		auto bytes_read = fread(offsetsPtr++, sizeof(int), 1, cf->fp);
```
ngld, line 158: You should probably check `cf->compression_info.numOffsets * sizeof(int) <= maxBlocks` with an `Assertion()` to prevent out-of-bounds writes here.

ngld (changes requested):
This looks pretty good so far. Most of my comments are nitpicks related to stylistic issues. However, there are a few problems with memory handling that I'd like to see addressed.

code/cfile/cfilearchive.cpp
```diff
@@ -54,18 +54,42 @@ void cf_init_lowlevel_read_code( CFILE * cfile, size_t lib_offset, size_t size,
...
+	cfseek(cfile, 0, SEEK_SET);
+	if (comp_check_header(header) == COMP_HEADER_MATCH)
+		comp_create_ci(cfile, header);
+}
+
+//This function is called to cleanup compression info when the cfile handle is reused.
+void cf_clear_compression_info(CFILE* cfile)
+{
+	free(cfile->compression_info.offsets);
+	free(cfile->compression_info.decoderBuffer);
```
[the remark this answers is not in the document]
Shivansps: Ok, added a if to check the header, if is not zero then it means the pointers arent nullptr, if the header is zero then it does nothing as nothing needs to be done.

code/cfile/cfilearchive.cpp
```diff
@@ -264,6 +301,4 @@ int cfread_lua_number(double *buf, CFILE *cfile)
 	#endif
 
 	return items_read;
-
-}
-
+}
```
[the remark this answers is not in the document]
Shivansps: this is likely a accident ill add it back

code/cfile/cfilearchive.cpp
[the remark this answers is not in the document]
ngld: What happens if opening a file fails after the header is detected? Does FSO always quit in such a situation (i.e. due to an `Assertion()`) or is it possible that a CFile block is re-used that has a header but doesn't have an `offsets` and/or `decoderBuffer` pointer?

code/cfile/cfilecompression.cpp
Shivansps: yeah ill make it more clear

code/cfile/cfilecompression.cpp
Shivansps: ill add it, but thats is going to trigger an assertion at cfread anyway.

code/cfile/cfilecompression.cpp
Shivansps: i did that and some other changes to the random access function, i simplified the if.

code/cfile/cfilearchive.cpp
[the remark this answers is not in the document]
Shivansps: Is not possible to have nullprts along with a header id, after the header is identified you either load the file in the right way or you will trigger some assertion along the way. The two assertions at lz41_load_offsets() makes it almost impossible to pass that point with bad data, and if for some reason you do, the decoder function is going to fail and trigger a file decompression error assertion at cfread,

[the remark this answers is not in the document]
ngld: Fair enough I guess. I don't like that all of this relies on the fact that the functions will always be called in the right order since the implementation and inner workings of CFile might change in the future and someone might accidentally break this. However, it seems pretty unlikely for now.

ngld (approved):
This looks great! Thanks for accomodating my requests and the simplified `lz41_stream_random_access`. The new version is much easier to read.

notimaginative:
There really needs to be an extension change on this at the least. Compressed VPs aren't going to be supported by older builds so there needs to be a way to support both compressed and uncompressed versions of the VPs in the same location where multiple build versions could be used (like with MediaVPs, etc.)

Shivansps:
This is not tied to the VPs, it just handles files that are tagged as compressed by the file header, the file could be inside a VP or just dropped into the data folder, so, there is no need to compress everything and you can mix compressed and non compressed files inside a vp. There should be no issues there.

ngld:
Old builds won't be able to parse VPs with compressed files but having both compressed and uncompressed versions of the VPs side-by-side kind of defeats the point.
Just like any other engine feature, mods that use this feature will require builds that support it. The MediaVPs in particular won't be able to use this for past releases. It should be fine to use in new MediaVPs releases once this feature is supported in the next FSO stable.

notimaginative:
Okay. Maybe I'm missing something or just have a fundamental misunderstanding about it. My previous work on a similar feature is probably clouding my vision on this. I just can't shake the clusterfuck vibe this thing keeps giving me.

Shivansps:
The fundamentals of how this works is:
At compression:
-Add "LZ41" as file header, store the compressed blocks, then the offset list, number of offsets, filesize and block size.
So, for example, if you open a compressed "fighter01.pof" with a text editor, you are going to see "LZ41" at the beginning.
-Then you store these files into the vp normally, OR you can choose not to use a VP and just put it into the data folder.

FSO:
-When cfopen is called on a file, it reads the first 4 bytes to check if the header is present, if it is present it marks the file as compressed and does the extra work needed.

The one thing that i was thinking about doing is to add ".lz4" to the file extensión, so the file end up as "fighter01.pof.lz4" and then strip the ".lz4" when the files are detected. But at the end i decided nothing was actually gained from this, because that stops older fso versions picking up the compressed files but whiout those files the mod is not going to work anyway so there is no point and it was a bunch of unnecessary extra work so i dropped the idea.

If thats desired it could be added at any point since that changes nothing in this pr and that work is done in a diferent part of the code as well.

notimaginative:
Yeah, I get how it works. I just don't understand *why* it's like that.

There is no way to tell a regular VP apart from a compressed VP, outside of scanning through all of the files to see if one is compressed or not. There is nothing to prevent an old build from attempting to use a compressed VP, where the compressed data would be unusable. Outside of manually naming the VP with an indicator that its compressed there is no way to really know if a VP actually is compressed. VP compatibility is preserved at the expense of... compatibility. And efficiency... And file size...

Compressing loose files has the same problem without a rename. File extensions are used to determine file type, but the file type is different when compressed.

This just seems like one of the least efficient, least compatible, least user-friendly ways to implement compressed VPs that I can think of. That's why I have to assume that I'm missing something. I just don't get it.

Baezon:
I don't see how this is different from any other new feature though. A table with a new feature and thus is completely unreadable by an old build is completely indistinguishable from a compatible table aside from manually scanning each and every entry. 

I just don't see how this would be a practical problem any more than it already is, the simple fact is that users attempting to run new data on old builds is an extremely rare situation.

notimaginative:
* shrugs *

Sorry, I just can't see anything but a bad idea here. I'll let you guys sort it out and hope for the best.

ngld:
> This just seems like one of the least efficient, least compatible, least user-friendly ways to implement compressed VPs that I can think of. That's why I have to assume that I'm missing something. I just don't get it.

I don't see how it could be more compatible or more user-friendly. The only downsides I can see is that it's not particularly obvious that a VP contains compressed files, however, it would be obvious to the user since they have to use a tool to create the compressed files in the first place. Using a flag in the VP format instead of the `LZ41` header would save a few bytes per file but I think the difference is negligible.

Regarding the compatibility: I don't think it makes a difference whether an old build wouldn't see compressed VPs due to a new file extension and complain about missing files that exist or if it complains about corrupted / unreadable files because it can't parse the compressed files.

ngld:
Also, regarding the file extension: Trying to use tables that work fine on FSO 21.0 on an old 3.6.12 build is very likely to fail and yet they have the same file extension. I don't see how this applies to tables but doesn't apply to VPs. Technically, the VP format didn't even change. Instead, all virtual files can be wrapped in a compression container (like `.gz`). This container format doesn't change the file extension, true, but that would be a pain to implement (since now you'd have to change all references to the compressed files as well or change how CFile looks up files).

The only issue I can see is that some text file or LUA file in a mod coincidentally starts with `LZ41` which would trigger the decompression logic and produce corrupted data since the file wasn't compressed to begin with. It seems very unlikely to that this is the case, though.

Shivansps:
The compressor can add "compressed" to the vp name if it contains compressed files since vp names changes nothing.

What is needed to keep in mind here is that you just cant make a ".CVP" and make that all files inside it to be compressed, that is very inefficient, some of the files FSO uses are very small, like .fc2 or .eff that will always go up in size, other formats, like .png are almost always going to go up in size, the compressor needs to detect those cases and only store files that are smaller than the original, so you need to mix both compressed and uncompressed in the same container there is no other way.

This forces you to have a header that identifies what files are compressed, it does not really matter if is stored on the file or on a index, you need to store that data somewhere. The header also helps you for, in the future, to add other compression formats or a update to the LZ4 compression system whiout breaking compatibility with the older files.

Of the rest of the information that is added to the file, the offset list is needed for fast random access, the number of offsets/blocks you will need it to parse that data, the original filesize is needed is needed so FSO filesystem can actually use that file, and the blocksize is there so all optimizations are done on compressor side, you can remove that and stick with fixed block size, what is a bad idea because you can fine tune the compressor to use a optimal block size taking in considerations the file size and type.

So in the end you are forced to have all that extra data, it does not really matter where it is stored, the fact that it is embedded in the file allows for standalone files support along with vp support whiout making changes to the vp structure or parsing.

The older builds arent going to work either way, making it so you can avoid older builds to pick up the compressed files are going to cause the mod to complain about missing files, and in this way you get parse errors/assertions. It does not going to work either way.

JohnAFernandez:
All right, we decided we're good to go on merging this before the release cycle. Great work, @Shivansps !

PR #3297, opened 2021-03-14
Fix ammo safety regressions.

MageKing17:
When various SEXP functions related to ammo were consolidated back in June (in commit 54409ec92887d2e6db92450a2ccf78d8fc90ff2d, PR #2480), certain safety catches meant to avoid having too much ammo or the wrong ammo capacity after changing weapons were lost. This commit adds code that should be functionally equivalent.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -17322,7 +17322,7 @@ void sexp_set_ammo_sub(ship_weapon *swp, int requested_bank, int requested_ammo,
 	}
 	else if (ammo[requested_bank] > maximum_allowed)
 	{
-		// Make sure the current ammo doesn't exceed the maximum
+		// Make sure the current ammo doesn't exceed the maximum (this can otherwise happen with the set-*-weapon SEXPs)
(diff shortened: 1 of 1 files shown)


JohnAFernandez (approved):
Looks good.  But us having to fix this over and over really shows one of the weaknesses of FSO: not using setter functions which would always be able to do these safety checks by default.

Goober5000 (changes requested):
Interesting.  Yes, I see now how the `set-*-weapon` sexps would have unexpected results.

I would like comments added to the `else` blocks clarifying that those code paths are needed when using `set-*-weapon`, because it took me a bit of scrutinizing to realize that.  Also, remove or reword the "No matter what" comment since its context is lost.  Otherwise this looks good.

MageKing17:
> Also, remove or reword the "No matter what" comment since its context is lost.

I mean, the context seems pretty clear, given that there's exactly one other comment in the block, but sure, I'll clarify it.

Goober5000 (approved):
That looks good.

PR #3300, opened 2021-03-15
Fix assert with combining flak and non-flak using substitute

### LINKED ISSUE #2415: Assert when using $Substitute to mix flak and non-flak
Weapon A does not have the flak flag, but uses $Substitute: to fire weapon B part of the time. Weapon B has flak. 

This appears to behave just fine on non-debug builds, but on debug builds trips an assert that appears to be there to verify flak effects are only being asked for from flak turrets. 

https://fsnebula.org/log/5ecad881cb0d3321adc98627

Log is running on a recent build of the multilock test branch, because that's what I had a full debug of already built, but I've verified I hit the assert in current nightly fast debugs. 

I'll see about building an isolated test case for this.

Baezon:
This should be a perfectly fine thing to do, and only complains because flak muzzle flashes have a pointless wrapper function for the normal muzzle flash function that exists solely to complain about weapons being not flak. 

Closes #2415

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -11892,11 +11892,6 @@ int ship_fire_primary(object * obj, int stream_weapons, int force, bool rollback
 								{
 									flak_set_range(&Objects[weapon_objnum], flak_range - winfo_p->untargeted_flak_range_penalty);
 								}
-
-								if ((winfo_p->muzzle_flash>=0) && (((shipp==Player_ship) && (vm_vec_mag(&Player_obj->phys_info.vel)>=45)) || (shipp!=Player_ship)))
-								{
-									flak_muzzle_flash(&firing_pos,&obj->orient.vec.fvec, &obj->phys_info, swp->primary_bank_weapons[bank_to_fire]);
-								}
 							}
 							// create the muzzle flash effect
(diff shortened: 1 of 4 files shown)


code/ship/ship.cpp
```diff
@@ -11892,11 +11892,6 @@ int ship_fire_primary(object * obj, int stream_weapons, int force, bool rollback
 								{
 									flak_set_range(&Objects[weapon_objnum], flak_range - winfo_p->untargeted_flak_range_penalty);
 								}
-
-								if ((winfo_p->muzzle_flash>=0) && (((shipp==Player_ship) && (vm_vec_mag(&Player_obj->phys_info.vel)>=45)) || (shipp!=Player_ship)))
```
JohnAFernandez, line 11896: Doesn't part of this check keep the player from having their view obstructed by a muzzle flash if they are moving fairly slowly?  Doesn't that need to be preserved somewhere else?

Baezon: Yes, this was an already awkward way to handle that, but normal muzzle flash logic knows how to deal with that, as it depends on several factors, like whether 'show ship model' or '$render player muzzle flashes" is used.

JohnAFernandez: Ok.  Sounds good then.

JohnAFernandez (approved):

PR #3301, opened 2021-03-16
Implement substitution for beams

Baezon:
Also make explicit no mixing of beams/non-beams substitution, that's just going to be painful to attempt.


JohnAFernandez (approved):
AFAICT, it looks good the way it is.

PR #3304, opened 2021-03-17
Add pierce ship weapon flag

BMagnu:
Adds "pierce ship" as a weapon info flag, which will not remove a weapon on collision with an object, and rather allow it to penetrate and continue flying until it's lifetime is 0, detonated by a player, or shot down.
Regarding possible particle spew upon leaving a pierced ship:
It seems insufficient to just flip the weapon direction for the collision checks to deal with backface collisions as soon as the weapon entered a ship. Larger changes would be necessary to properly check for backface collisions which are, for now, out of scope.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -182,6 +182,7 @@ flag_def_list_new<Weapon::Info_Flags> Weapon_Info_Flags[] = {
 	{ "no impact spew",					Weapon::Info_Flags::No_impact_spew,						true, false },
 	{ "require exact los",				Weapon::Info_Flags::Require_exact_los,					true, false },
 	{ "can damage shooter",				Weapon::Info_Flags::Can_damage_shooter,					true, false },
+	{ "pierce ships",					Weapon::Info_Flags::Pierce_ships,						true, false },
 };
 
 const size_t num_weapon_info_flags = sizeof(Weapon_Info_Flags) / sizeof(flag_def_list_new<Weapon::Info_Flags>);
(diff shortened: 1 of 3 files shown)


Baezon (changes requested):
Looks good, this should cover all cases pretty well I think. I'd like to add some of those QoL things like exit collisions, but those can be added later, if its possible. Only thing I would ask is that children spawning I think should probably be moved too. Detonate spawns I think it's reasonable to only spawn if it *truly* detonates.

BMagnu:
I'd probably like to make that configurable then.
Since I can well imagine interesting "on hit" spawns (which were so far identical to detonate).
What'd be the preferred method of adding this type of config to the table? Change the flag into a proper entry with an optional sideentry about the spawns? Or adding another flag?

Baezon (approved):
Eh, I'm not convinced this really needs to be configurable, its an extraordinarily niche scenario anyway so unless its specifically needed I think it would just introduce pointless complication. 

But I think you make a good point that the normal situation should be triggering all the 'on hit' effects, including spawning.

Baezon (changes requested):
On second thought (this is my last flip flop I promise) it should be its own entry in the table, as there are several sub fields id like to add at some point but aren't necessary right away. So like '$pierces ship: yes' or whatever.

code/weapon/weapons.cpp
```diff
@@ -6859,15 +6866,17 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
 		}
 	}
 
-	// spawn weapons - note the change from FS 1 multiplayer.
-	if (wip->wi_flags[Weapon::Info_Flags::Spawn]){
-		if (!((wip->wi_flags[Weapon::Info_Flags::Dont_spawn_if_shot]) && (Weapons[num].weapon_flags[Weapon::Weapon_Flags::Destroyed_by_weapon]))){			// prevent spawning of children if shot down and the dont spawn if shot flag is set (DahBlount)
-			spawn_child_weapons(weapon_obj);
+	if (!wip->pierce_objects || wip->spawn_childs_on_pierce || other_obj) {
```
Baezon, line 6869: Is that right? Should be `!other_obj` right?

code/weapon/weapons.cpp
```diff
@@ -1755,6 +1754,14 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		}
 	}
 
+	if (optional_string("$Pierce Objects:")) {
+		stuff_boolean(&wip->pierce_objects);
+
+		if (optional_string("+Spawn Childs On Hit:")) {
+			stuff_boolean(&wip->spawn_childs_on_pierce);
```
Baezon, line 1761: Pedantic, but should be `spawn_children_on_pierce`. And also the table entry should specify "on pierce" just to be that much clearer.

Baezon (changes requested):

code/weapon/weapons.cpp
BMagnu: absolutely right, that'd have been a bad oversight

Baezon (approved):

PR #3305, opened 2021-03-17
Adding "Start missions in Chase View by default" flag in game_settings.tbl

### LINKED ISSUE #3152: FR: set third person "chase" camera as default
Possible flag to set in game_settings.tbl. Could be useful for arcadey TCs like Starfox: Event Horizon and Solaris. The option to switch between cockpit "first person" and chase "third person" views is still there but the game defaults to the chase view.

Strygon:
If the mod has "$Player starts in Third Person View by Default: TRUE" in their game_settings.tbl or any such .tbm, missions will automatically put the player in the third person chase view. This was requested as a FR by Darius on January 17th.

This will NOT stop the player from going back into first person by pressing the key for it.

closes #3152

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -62,7 +62,7 @@ bool Use_engine_wash_intensity;
 bool Framerate_independent_turning; // an in-depth explanation how this flag is supposed to work can be found in #2740 PR description
 bool Ai_respect_tabled_turntime_rotdamp;
 bool Swarmers_lead_targets;
-bool Third_Person_is_Default;
+bool Chase_view_default;
 SCP_vector<gr_capability> Required_render_ext;
 float Weapon_SS_Threshold_Turret_Inaccuracy;
 bool Render_player_mflash;
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -659,7 +659,7 @@ void mod_table_reset()
 	Use_engine_wash_intensity = false;
 	Framerate_independent_turning = true;
 	Ai_respect_tabled_turntime_rotdamp = false;
-	Third_Person_is_Default = false;
+	Chase_view_default = false;
 	Swarmers_lead_targets = false;
 	Required_render_ext.clear();
 	Weapon_SS_Threshold_Turret_Inaccuracy = 0.7f; // Defaults to retail value of 0.7 --wookieejedi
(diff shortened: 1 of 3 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -62,6 +62,7 @@ bool Use_engine_wash_intensity;
 bool Framerate_independent_turning; // an in-depth explanation how this flag is supposed to work can be found in #2740 PR description
 bool Ai_respect_tabled_turntime_rotdamp;
 bool Swarmers_lead_targets;
+bool Third_Person_is_Default;
```
BMagnu, line 65: This should be `Chase_view_default `as well.

code/mod_table/mod_table.cpp
```diff
@@ -654,6 +659,7 @@ void mod_table_reset()
 	Use_engine_wash_intensity = false;
 	Framerate_independent_turning = true;
 	Ai_respect_tabled_turntime_rotdamp = false;
+	Third_Person_is_Default = false;
```
BMagnu, line 662: This should also be `Chase_view_default`

BMagnu (changes requested):
Apart from the few bad bools it seems good

BMagnu (approved):
LGTM

Baezon (approved):

Goober5000 (changes requested):
This should be a missions flag, not a game_settings flag.

code/playerman/playercontrol.cpp
```diff
@@ -1379,7 +1379,15 @@ void player_level_init()
 	Viewer_external_info.preferred_distance = 0.0f;
 	Viewer_external_info.current_distance = 0.0f;
 
-	Viewer_mode = 0;
+	
+	if (Chase_view_default == true)
```
Goober5000, line 1383: This is a redundant boolean comparison, which some compilers will complain about.

Goober5000 (changes requested):

Baezon:
I don't see why it should be a mission flag, all the current examples of it are as a stylistic choice for the entire mod. While it's not inconceivable that someone might want it per mission, its clear the vastly more common scenario is for the entire mod.

Goober5000:
It really should be a per-mission flag.  If we want to use game_settings to make the mission flag default to true for every mission, that's fine, but this setting should be available on a per-mission basis.  I can already think of one use case and I'm sure there are a few I haven't thought of.

Goober5000 (approved):
Per discussion with @Baezon I'll approve this, and I'll add the mission flag in a separate PR.

Issue #3299, opened 2021-03-15
Turret fov and orientation issues on retail aeolus model


Baezon:
Some retail models (I noticed it on the Aeolus) turret's seem to have incorrectly determined fov, causing them to only fire/track their targets if they are in the direction of their barrels. Additionally the Aeolus's rear top turret seems to have an inverted base orientation (points in the opposite direction, preventing the gun from being able to pitch down to shoot targets properly), and its gun submodels appear to always aim like 30 degress *below* their target, but I couldn't replicate those two on any other models, so I dunno if its specific to the Aeolus or what.

Goober5000:
I think the root cause is that `ship_get_global_turret_info` is calculating the wrong `gvec`.  This should be the turret normal, but instead it looks like it's returning the direction that the gun barrel is pointing.

Goober5000:
Or maybe it's the `turret_norm` itself that's bad.  This is strange...

Baezon:
That's the big weird part, which I really don't know how to consistently square: the turret normal isn't facing the same direction as the turret submodel's up vector. I'm not sure to what degree that's a fundamental problem or just something that's odd.

Goober5000:
I think this is actually not a bug.  To the extent that there were problems with the turret FOV, they were fixed in the other PRs.  @Baezon do you concur?

Baezon:
I would agree, for posterity: the gun angle and base angle issues existed for far longer than the recent changes, the gun angle problem being based on the model file itself. While its possible a solution or workaround could be found, it is not behavior affecting and specific to the aeolus and so not critical.

PR #3309, opened 2021-03-19
Update Turret_alt_math for moving subsystems

Baezon:
And renames it to what it *actually* means: a restricted fov, with limited elevation or base fov. Uses `model_instance_find_world_dir` for the original turret model's frame of reference vectors to produce a properly rotated turret matrix which the limited fov's can then be calculated from.

Additionally, most things that need the turret's normal or any direction based on the turret but *not* based on the turret's current 'angles', should not get `model_instance_find_world_dir` using the turret's subobject, but instead its parent. An optional parameter was added to `model_instance_find_world_dir` for this purpose so callers don't have to faff about with the polymodel.


Goober5000 (commented):
My brain is too fried to review this properly right now, but I have one question.  Instead of adding a `parent` parameter to that function, why not make sure every invocation that needs the parent actually calls it with the parent submodel?

Baezon:
Most callers don't have the `pm` handy (which is what you need to find the parent) and so let the overload of `model_instance_find_world_dir` do the lookup itself with the `model_instance_num` which cuts down on those repetitive lookups from all the callers. Its definitely kind of awkward, so if be happy to handle this however you'd think is best.

Goober5000:
I guess that is fine.  But rename the parameter to something more explicit like `use_submodel_parent`.

Goober5000 (commented):
Ok there is one difference here which might be a problem... or it might not.

Originally, `turret_matrix` was simply a representation of `turret_norm` as a matrix.  This matrix was fixed to the reference frame of the main model.  Upward-pointing turrets were one handedness and downward-pointing turrets were another handedness.

This is not the same concept as the turret frame-of-reference, which is the fvec/uvec defined in the POF.  In this sense, the `turret_norm` is always pointing "directly up" from its perspective, but its perspective orientation is defined by the fvec/uvec.

In this PR, you've taken a calculation based on the original sense of the turret matrix and rewritten it to use the fvec/uvec method.  This is fine -- and it's what this section of code needed -- but I want to make sure you understand the nuances of the different methods.

Baezon:
Iirc `world_to_turret_matrix` was *not* equivalent to `turret_matrix` because, as you mentioned it was just the normal turned into a matrix. The restricted fov case really needs that separate fvec and uvec as defined in the pof in order for its restrictions to make sense, as they are from that point of view.

It should be noted that, at least for restricted Fov turrets, the turret normal *must* be the same as its uvec, or it violates several implicit assertions. We've touched on this briefly before in DMs , but its clearer to me now how this should work. If they aren't restricted, its less clear to me and probably not as harmful but I'm still not really sure how its supposed to work.

wookieejedi:
Tested with restricted FOV turrets from FotG and it fixes things and works as expected.

Goober5000 (approved):
Looks good.  This might have something to do with #3299 as well.  We'll see.

PR #3310, opened 2021-03-19
Fix Fov_edge_checks

Baezon:
This flag is useless, it attempts to hold on to targets that have gone out of fov by getting the gun as close as possible and firing in that direction, but that's not a great idea since that point could very well not be aiming at the ship at all, and furthermore the turret is constantly re-targeting either new subsystems on it's target or new targets in its fov, which this edge checks system doesn't touch anyway, so that work will get clobbered in moments after losing fov in the 'standard' way. 

Rather rolling a new function that isn't called in the right places, this flag will have another branch in `object_in_turret_fov`, the real function that determines whether to acquire a new target, and that's the function that is too conservative with regards to the target's size and proximity, while the normal method *does* take them into account, it will still often ignore ships which have portions which are comfortably in fov. The new method simply checks each of the bbox points (with a small fov penalty, don't want to be too greedy if it can only see a tiny portion). 

The changes look confusing, but what happened was that `is_object_radius_in_turret_fov` was removed and an extra branch was added to `object_in_turret_fov`.

Edit: also fixes the crashes that will happen if you attempt to use this flag due to a minor oversight in #3277, the fix is pretty trivial but I figured I would kill two birds with one stone.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -98,124 +98,51 @@ typedef struct eval_enemy_obj_struct {
  *
  * @return 1 if objp is in fov of the specified turret.  Otherwise return 0.
  */
-int object_in_turret_fov(object *objp, ship_subsys *ss, vec3d *tvec, vec3d *tpos, float dist)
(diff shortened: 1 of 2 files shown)


Goober5000 (approved):
Ok, this looks good. :+1:

PR #3314, opened 2021-03-21
Make sure objects and ships get cleared if leaving mission sync

JohnAFernandez:
This properly clears ships if you are leaving the sync state for any other state besides the gameplay state.

Previously, if the player left this state and then started another game, the old ships would still be present and could causes nan crashes.


notimaginative (approved):
LGTM

Issue #3241, opened 2021-02-22
FR: Enable $default keys$ in debriefings, recommendations


naomimyselfandi:
There are some controls that are _very_ useful in very specific cases but aren't particularly well-known, like "Target my target's target". It'd be really nice to be able to mention these keys in debriefings and _especially_ in recommendations:

> [SPOILER REDACTED]. Make sure to prioritize threats appropriately. You can press $G$ to target a ship attacking your target, or $J$ to target your target's target. Both of these controls can be put to good use here.

MjnMixael:
+1

PR #3315, opened 2021-03-21
$default keys$ in briefings, debriefings and recommendations

bryansee:
#3241 documents some controls that are very useful in very specific cases but aren't well-known, like "Target my target's target". This implements $default keys$ in not just debriefings and recommendations, but also briefings.

There may be work to follow up, but this is the only good start for me, as I'm still learning.


asarium (changes requested):
Ok, this code is tangentially related to the request in the issue.

However, this simply makes invalid color tags show up in the text which is 1. a change in behavior and 2. on its own not related to the feature request.

I see no point in this PR so you either need to add all the required changes here or at least explain what you are doing.

The-E:
Nope, this is not it.

Bryan: *Learn to read*

PR #3316, opened 2021-03-22
Add -pilot cmdline flag

### LINKED ISSUE #2869: FR: -pilot command line flag
When repeatedly relaunching a mod to test changes, the fewer clicks the better. Launching a mission from command line already gets a lot, which is good, but picking the pilot that way too would remove half the remaining ones at worst, and all of them at best.

Alternately one could have an option to always pick the last pilot again at launch and make the player click on the pilot room if they wanted to switch.

I haven't looked at the code but I would imagine one of these would be within my own coding reach if nobody else jumps on it or decides it'd be a bad feature, but if so it'll not be right away, so on the issue stack it goes.

the-maddin:
Fixes #2869 .

A Pilot can now be set via the Dev cmdline argument `-pilot`, which requires a parameter `"<pilotname>"`.
This currently implies single-player mode.
Using this flag sets `ui.PilotSelect.isAutoselect()` to true (as does `-benchmark_mode`).
Also clarify error message when an invalid pilot name or file is specified.


code/cmdline/cmdline.cpp
```diff
@@ -246,7 +246,8 @@ Flag exe_params[] =
 	{ "-noninteractive",	"Disables interactive dialogs",				true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-noninteractive", },
 	{ "-no_unfocused_pause","Don't pause if the window isn't focused",	true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-no_unfocused_pause", },
 	{ "-benchmark_mode",	"Puts the game into benchmark mode",		true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-benchmark_mode", },
-	{ "-profile_frame_time","Profile frame time",				true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-profile_frame_time", },
+	{ "-pilot",	"Selects specified pilot",								true,	0,									EASY_DEFAULT,					"Dev Tool",		"http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-pilot", },
```
asarium, line 249: This should not be added here since this will export the flag to the launchers which do not support supplying values to command line parameters.

code/cmdline/cmdline.cpp
```diff
@@ -510,6 +512,7 @@ int Cmdline_reparse_mainhall = 0;
 bool Cmdline_profile_write_file = false;
 bool Cmdline_no_unfocus_pause = false;
 bool Cmdline_benchmark_mode = false;
+char *Cmdline_pilot = nullptr;
```
asarium, line 515: Please use `SCP_string` to avoid issues with memory management.

code/scripting/api/libs/ui.cpp
```diff
@@ -220,7 +220,7 @@ ADE_FUNC(isAutoselect, l_UserInterface_PilotSelect, nullptr,
          "Determines if the pilot selection screen should automatically select the default user.", "boolean",
          "true if autoselect is enabled, false otherwise")
 {
-	return ade_set_args(L, "b", Cmdline_benchmark_mode);
```
asarium, line 223: Please also add an `ADE_VIRTVAR` that returns the command line value so that a script can replicate the behavior of the "normal" menu.

asarium (changes requested):

code/cmdline/cmdline.cpp
the-maddin: Are you certain about this? Other flags in this file use `char*` without exception, `SCP_string` is not used in any exported flag variable.

asarium: Huh, I thought that was changed at some point. I guess char is fine then but it should still be `const char*`.

the-maddin: It's not `const` anywhere else either, but I agree, it should be.

asarium (approved):

PR #3319, opened 2021-03-24
Allow for setting of nebula fog color explicitly

Baezon:
The color of the nebula fog/background is implicitly determined from the .pcx texture file provided for the background, as an average of its palette. This is kind of weird and inconvenient so this allows the engine to read from the mission file a color directly as an alternative to the texture, as well as a sexp to change it on the fly. This does not have FRED support yet, hopefully @Goober5000 or anyone else familiar with FRED may be able to help, but it should hopefully be as simple as allowing for a little color picker or something in the nebula section.


Goober5000 (changes requested):
Looks good.  FRED support shouldn't be difficult to add.  A few things...

1) There's neb2 parsing code in lab_renderer.cpp.  It looks very similar; it would be nice to consolidate them somehow.  At minimum, the new parsing needs to be added there too.
2) Your nemesis, initialization, bit you again.  As far as I can tell, `Neb2_texture_name` is only initialized when the program starts, not at every mission load -- except for the `strcpy_s` that you removed.

Baezon:
2. So it should be wiped out at mission close right? The `strlen` neb.cpp does on it made me figure that `Neb2_texture_name` should be empty if it's not a nebula or instead is using `Neb2Color`

Goober5000:
> 2. So it should be wiped out at mission close right? The `strlen` neb.cpp does on it made me figure that `Neb2_texture_name` should be empty if it's not a nebula or instead is using `Neb2Color`

It is expected to be empty, yes, but it won't be empty if you load a mission with the file name and then subsequently load a mission with the color.

If you add a `strcpy_s(Neb2_texture_name, "");` where previously you had a `strcpy_s(Neb2_texture_name, "Eraseme3");`, it should parse correctly in all situations.

Goober5000 (approved):

PR #3320, opened 2021-03-24
Add `no impact debris` for whole ship

wookieejedi:
Previously, the `no impact debris` flag was only utilized for subsystems. This PR adds that flag for the whole ship. It also will disable debris spew for subsystems when the ship flag is enabled.

### DIFF
--- a/code/ship/shiphit.cpp
+++ b/code/ship/shiphit.cpp
@@ -517,8 +517,8 @@ float do_subobj_hit_stuff(object *ship_objp, object *other_obj, vec3d *hitpos, i
 	if (!global_damage) {
 		auto subsys = ship_get_subsys_for_submodel(ship_p, submodel_num);
 
-		if (subsys == nullptr || !subsys->system_info->flags[Model::Subsystem_Flags::No_impact_debris] ||
-			!Ship_info[ship_p->ship_info_index].flags[Ship::Info_Flags::No_impact_debris]) {
+		if (subsys == nullptr || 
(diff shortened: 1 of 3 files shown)


code/ship/shiphit.cpp
```diff
@@ -1556,8 +1557,9 @@ static void ship_vaporize(ship *shipp)
 	ship_objp = &Objects[shipp->objnum];
 	ship_info* sip = &Ship_info[shipp->ship_info_index];
 
-	// create debris shards
-	create_generic_debris(ship_objp, &ship_objp->pos, (float)sip->generic_debris_spew_num, sip->generic_debris_spew_num * 2.0f, 1.4f, true);
+	// create debris shards if allowed
+	if (!sip->flags[Ship::Info_Flags::No_impact_debris])
+		create_generic_debris(ship_objp, &ship_objp->pos, (float)sip->generic_debris_spew_num, sip->generic_debris_spew_num * 2.0f, 1.4f, true);
```
Baezon, line 1562: This isn't quite the same situation as an "impact" per se, this is when the ship is destroyed by a large beam weapon. While it's not crazy to have a flag to control that, it seems like this is maybe not what you intended? Something similar happens when a small ship is destroyed as well in `shipfx_blow_up_model()`. If you simply don't want to the ship to produce generic debris on death setting +Generic Debris Spew Num: to 0 *should* suffice.

wookieejedi: Ah indeed, this addition should not be used here!

Baezon (approved):

PR #3323, opened 2021-03-24
Regression: Save Mouse control binds

### LINKED ISSUE #3322: User-bound mouse buttons are cleared or ignored
In recent nightlies, I effectively cannot bind any mouse button to seemingly any control. Default bindings for joy-0 work perfectly for the corresponding mouse buttons.

When I bind mouse buttons to anything, then accept,, and return to the controls screen from the first options screen where difficulty can be set, all mouse buttons stay bound.
If I instead bind mouse buttons and exit to the mainhall, then return to the controls screen, mouse buttons are no longer bound. The right column shows "None" bound even if I had replaced an existing joystick binding. 
If I bind mouse buttons mid-mission, the mouse buttons still don't do anything, but returning to the controls screen still shows that they are bound. The default joy-0 button bindings work for the mouse, but I don't know how to bind them to arbitrary keys.

BMagnu:
Should fix #3322 until controls4/5 rolls around.
Still, @z64555 should have a look at this as well.

### DIFF
--- a/code/pilotfile/csg.cpp
+++ b/code/pilotfile/csg.cpp
@@ -1221,6 +1221,8 @@ void pilotfile::csg_read_controls()
 		if (idx < CCFG_MAX) {
 			Control_config[idx].take(CC_bind(CID_KEYBOARD, id1), 0);
 			Control_config[idx].take(CC_bind(CID_JOY0, id2), 1);
+			if(id3 != -1) // This means the mouse was bound.
+				Control_config[idx].take(CC_bind(CID_MOUSE, id3), 1);
 		}
 	}
 }
@@ -1236,8 +1238,7 @@ void pilotfile::csg_write_controls()
 	for (idx = 0; idx < CCFG_MAX; idx++) {
 		cfwrite_short(Control_config[idx].get_btn(CID_KEYBOARD), cfp);
 		cfwrite_short(Control_config[idx].get_btn(CID_JOY0), cfp);
-		// placeholder? for future mouse_id?
-		cfwrite_short(-1, cfp);
+		cfwrite_short(Control_config[idx].get_btn(CID_MOUSE), cfp);
 	}
 
 	endSection();
--- a/code/pilotfile/plr.cpp
+++ b/code/pilotfile/plr.cpp
@@ -566,18 +566,20 @@ void pilotfile::plr_write_stats_multi()
 
 void pilotfile::plr_read_controls()
 {
-	short id1, id2;
+	short id1, id2, id3;
 	int axi, inv;
 
 	auto list_size = handler->startArrayRead("controls", true);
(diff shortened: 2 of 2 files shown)


MageKing17 (changes requested):
I have no idea if this is the "correct" temporary solution, but I *do* know that it's definitely incomplete if it only applies to CSG files and ignores PLR files.

z64555:
This fix shouldn't have any affect, Controls3 still has the mouse doubled up with the joystick, Controls5 is when they're split.

@BMagnu what are you trying to fix?

BMagnu:
> This fix shouldn't have any affect, Controls3 still has the mouse doubled up with the joystick, Controls5 is when they're split.
> 
> @BMagnu what are you trying to fix?

So, what happens here is that if a control gets bound to the mouse, it obviously has CID_MOUSE (-1) as its CID. When it get's saved to the CSG file, it gets the bindings for CID_KEYBOARD (-2) and CID_JOY0 (0), which means mouse (button) bindings just get thrown away and will not be saved.

With this change, it will save the bindings to CID_MOUSE as well, and then apply them again on load, if they are not -1 (which they will be if the control has no mouse button bound).

At least in my tests, this actually fixes the behaviour of #3322 as with this change, it is possible to assign mouse buttons again, which it wasn't before.

BMagnu:
> I have no idea if this is the "correct" temporary solution, but I _do_ know that it's definitely incomplete if it only applies to CSG files and ignores PLR files.

Good catch. Missed this since I only followed the code path of the controls code, which only seemed to read / save csg's and not plr's.

z64555:
That may fix behavior in the controls config menu (being able to bind and retain mouse bindings), but what about in-game?

BMagnu:
> That may fix behavior in the controls config menu (being able to bind and retain mouse bindings), but what about in-mission, does the control activate with the mouse bindings?

Indeed, that also doesn't work.
Seems like a bigger issue then.
Shall I leave you to it @z64555, or should I try to dig through the code?

BMagnu:
Though, presumably a quick fix would be to add some code to every place where .get_btn(CID_JOY0) is, to add a check similar to the one for writing the CSG here, that checks if the mouse binding is not -1, use that instead of the CID_JOY0 binding

z64555:
Yeah let me whip something up, I'll have it posted tomorrow.

Since the mouse is supposed to still be doubled up with the joystick, its completely moot to try to save the mouse binding or even be able to bind the mouse.  The fix will remove the instances of CID_MOUSE and replace them with CID_JOY0, and will be tested to ensure operation in-mission.

As an aside, Controls3 got as it is from me trying to piecemeal the overhaul project without prior knowledge of what practices and methods to use... so I feel responsible for breakages like this.

PR #3324, opened 2021-03-25
Fix #3322

### LINKED ISSUE #3322: User-bound mouse buttons are cleared or ignored

z64555:
When in bind mode, identify mouse as CID_JOY0 instead of CID_MOUSE.

Identifying the mouse as CID_MOUSE while in bind mode allowed users to bind controls to the mouse specifically.  However, this binding is neither saved nor respected in-mission.

Fixes #3322

### DIFF
--- a/code/controlconfig/controlsconfig.cpp
+++ b/code/controlconfig/controlsconfig.cpp
@@ -1934,7 +1934,7 @@ void control_config_do_frame(float frametime)
 					for (i=0; i<MOUSE_NUM_BUTTONS; i++) {
 						if (mouse_down(1 << i)) {
 							Assert(!(z & JOY_AXIS));
-							control_config_bind_btn(z, CC_bind(CID_MOUSE, static_cast<short>(i)), 1);
+							control_config_bind_btn(z, CC_bind(CID_JOY0, static_cast<short>(i)), 1);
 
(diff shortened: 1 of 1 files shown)


JohnAFernandez (approved):
This is really all there is to it?  Looks like a copy paste error maybe?

z64555:
> 
> 
> This is really all there is to it? Looks like a copy paste error maybe?

That's really all there is to it.  As I might have mentioned, I originally wanted to try splitting off the mouse from the joystick as soon as Controls3, but circumstances had me instead strip out that feature and defer it to the multijoy branch (Controls4).  I had gotten most of everything reverted but this one line change somehow got missed.

PR #3326, opened 2021-03-25
POF 2118 and SLC2 support

Shivansps:
Changed SLDC system to SLC2, this replaced the char at the beginning of the node for an int in order for the tree to be aligned.
Added BSP_Defpoints data alignment and SLDC to SLC2 conversion for pof version < 2118.

Re-upload after #2691 got merged and reverted by accident.

### DIFF
--- a/code/model/modelcollide.cpp
+++ b/code/model/modelcollide.cpp
@@ -973,44 +973,46 @@ bool mc_shield_check_common(shield_tri	*tri)
 
 bool mc_check_sldc(int offset)
 {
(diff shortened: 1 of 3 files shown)


JohnAFernandez (approved):
This has already been reviewed in the past.  Also, the models have been tested and ShivanSPS has  mentioned that the fixed models will be backwards compatible.  This should be merged.

wookieejedi:
Heavily tested with FotG and everything works as expected.

PR #3327, opened 2021-03-25
fix over alloc to store lz4 offsets

Shivansps:
This is a small bug that unfortunately i only managed to find today, since it does not affects functionality. Old code was reserving way too much memory to store the offset list. so i removed the max blocks calculation and just used the number of offsets*sizeof(int).

To compasate the lost Assertion at release builds, i moved the assertion checking the number of offsets that is read from file out of the NDEBUG.

### DIFF
--- a/code/cfile/cfilecompression.cpp
+++ b/code/cfile/cfilecompression.cpp
@@ -121,8 +121,8 @@ void lz41_create_ci(CFILE* cf, int header)
 	auto fSize = fread(&cf->size, sizeof(int), 1, cf->fp);
 	auto fBsize = fread(&cf->compression_info.block_size, sizeof(int), 1, cf->fp);
 
-	#if !defined(NDEBUG)
 	Assertion(cf->compression_info.num_offsets > 0, "Invalid number of offsets, compressed file is possibly in the wrong format or corrupted.");
+	#if !defined(NDEBUG)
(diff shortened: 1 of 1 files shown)


code/cfile/cfilecompression.cpp
```diff
@@ -121,8 +121,8 @@ void lz41_create_ci(CFILE* cf, int header)
 	auto fSize = fread(&cf->size, sizeof(int), 1, cf->fp);
 	auto fBsize = fread(&cf->compression_info.block_size, sizeof(int), 1, cf->fp);
 
-	#if !defined(NDEBUG)
 	Assertion(cf->compression_info.num_offsets > 0, "Invalid number of offsets, compressed file is possibly in the wrong format or corrupted.");
```
JohnAFernandez, line 124: In order for this not to show up on release builds, you'll have to switch to the Verification macro, which works on both release and debug.

JohnAFernandez (changes requested):

code/cfile/cfilecompression.cpp
Shivansps: Yeah i was wrong about Assertions usage, there is no need to run that on release, thats no more important than anything else on CFILE. Can we leave it as it is?

JohnAFernandez: Absolutely

JohnAFernandez (approved):

PR #3328, opened 2021-03-25
Add FSO_KEEP_STDOUT env var to skip the AttachConsole() call

### LINKED ISSUE #3317: FSO 21.0 Stable won't share or release console output to launchers
Caused by #2988, certain launchers that use the console for debugging, such as Knossos, are unable to do so because FSO won't let it.

@ngld discovered the issue while trying to upgrade Knossos.

Proposed solutions include:
* Building FSO with the SUBSYSTEM: set to CONSOLE.  This has the downside of showing a console at awkward times, even when its not needed
* Have FSO grab console on activation of specific flags
* Have FSO grab console only when it needs it, then release when its done.

ngld:
Fixes #3317

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -7545,13 +7545,24 @@ int main(int argc, char *argv[])
 
 	SCP_mspdbcs_Initialise();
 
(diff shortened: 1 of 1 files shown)


TRBlount (approved):
Looks good.

PR #3329, opened 2021-03-26
Split frame profile trace into two onscreen columns if needed

PhantomHoover:
Especially on 1080p screens, the frametime profile trace often runs off the bottom of the screen, making it useless for performance diagnostics. Splitting it is a simple workaround and should work fine for a debug feature.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -2030,8 +2030,21 @@ void game_show_framerate()
 		gr_set_color_fast(&HUD_color_debug);
 
 		if (Cmdline_frame_profile) {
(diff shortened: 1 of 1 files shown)


Baezon (approved):

Baezon:
Since this is very minor and just aesthetic I'm going to merge this tonight unless there are any objections.

Goober5000:
I'm in favor of merging it, provided it does the job.  Can someone post a screenshot?

PhantomHoover:
screen0016
screen0017

These two should show that it works in both single and double column mode, and also that it accounts properly for dynamic changes to the display font.

Goober5000:
Oh very nice.

I would like to see the column headers duplicated in the second column, but I'm not going to stand in the way of the merge for that.

PR #3331, opened 2021-03-27
Check for these states

JohnAFernandez:
Because GS_STATE_MULTI_MISSION_SYNC is actually shared for a few different multi screens, we do actually need to check if we are going to the PXO or Multi Join screens.


JohnAFernandez:
This fixes no briefing showing up and a crash on multi mission start.

notimaginative (approved):
Looks good, and I actually tested it this time :)

PR #3332, opened 2021-03-27
Replace rand() calls  with a Random class

jg18:
The class uses `std::mt19937` internally, but we could switch to any RNG that supports the functions that the engines in the C++ standard library header <random> support.

The design is tentative and up for discussion.

There are a number of comments with `DISCUSSME` with things I'd like to talk about.

I'm not sure what to do with the static_rand stuff, so would like to talk about that, too.

I found a few bugs and potential bugs while working on this, but since some could affect gameplay, I'm not sure whether fixing them is the right thing to do.


jg18:
Hmm, no idea how I explicitly requested a review from m!m (did I push the wrong button?) but of course I'm looking for feedback. 🙂

jg18:
I'll need to look into the CI failures. Unfortunately I don't have a Linux setup at the moment.

Goober5000:
> Hmm, no idea how I explicitly requested a review from m!m (did I push the wrong button?) but of course I'm looking for feedback. slightly_smiling_face

Sometimes a request will automatically occur if a coder is listed as "code owner" of a particular file or folder.  In this case it was the scripting code.

> I'll need to look into the CI failures. Unfortunately I don't have a Linux setup at the moment.

They're just time conversion warnings that are treated as errors.  Nothing platform-specific.

```
D:\a\fs2open.github.com\fs2open.github.com\code\render\3ddraw.cpp(1193,39): error C2220: the following warning is treated as an error [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\render\3ddraw.cpp(1193,39): warning C4244: '=': conversion from 'int' to 'ubyte', possible loss of data [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\render\3ddraw.cpp(1198,39): warning C4244: '=': conversion from 'int' to 'ubyte', possible loss of data [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\render\3ddraw.cpp(1203,39): warning C4244: '=': conversion from 'int' to 'ubyte', possible loss of data [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\render\3ddraw.cpp(1323,39): warning C4244: '=': conversion from 'int' to 'ubyte', possible loss of data [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\render\3ddraw.cpp(1328,39): warning C4244: '=': conversion from 'int' to 'ubyte', possible loss of data [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\render\3ddraw.cpp(1333,39): warning C4244: '=': conversion from 'int' to 'ubyte', possible loss of data [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
 D:\a\fs2open.github.com\fs2open.github.com\code\weapon\weapons.cpp(8194,55): error C2220: the following warning is treated as an error [D:\a\fs2open.github.com\fs2open.github.com\build\code\code.vcxproj]
D:\a\fs2open.github.com\fs2open.github.com\code\weapon\weapons.cpp(8194,55): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data [D:\a\fs2open.github.com\fs2open.github.com
```

code/starfield/starfield.cpp
```diff
@@ -869,19 +870,21 @@ void stars_post_level_init()
...
-			red= (ubyte)(myrand() % 63 +192);		//192-255
-			green= (ubyte)(myrand() % 63 +192);		//192-255
-			blue= (ubyte)(myrand() % 63 +192);		//192-255
-			alpha = (ubyte)(myrand () % 192 + 24);	//24-216
+			// DISCUSSME: the max value for these ranges is 254 and not 255. Do we fix it?
+			red= (ubyte)(util::Random::next(63) +192);		//192-255
+			green= (ubyte)(util::Random::next(63) +192);		//192-255
+			blue= (ubyte)(util::Random::next(63) +192);		//192-255
+			// DISCUSSME: despite the comment, the range is actually 24-215. Do we fix it?
+			alpha = (ubyte)(util::Random::next(192) + 24);	//24-216
```
PhantomHoover, line 887: Despite my general position that we should preserve existing behaviour over apparent intention, this is a) a tiny change in purely (?) cosmetic code, which is b) mostly obsolete in FSO today. Of course, given the latter, you wonder if it's worth changing it at all. I don't see why you don't just refactor these into two-argument range calls to `next()`, though.

code/utils/Random.cpp
```diff
@@ -0,0 +1,71 @@
...
+	return next() % modulus;
+}
+
+int Random::next(int low, int high)
+{
+	const int range = high - low + 1;
+	Assert(range > 0);
+
+	return low + (SCP_rng.next() % range);
+}
```
PhantomHoover, line 65: This could use a comment to make it explicit that this is an inclusive range that can return both `low` and `high`, particularly as the definition immediately above is _exclusive_ at the upper bound and cannot return `modulus`.

lib/lua/lmathlib.c
```diff
@@ -181,6 +181,7 @@ static int math_max (lua_State *L) {
 static int math_random (lua_State *L) {
   /* the `%' avoids the (rare) case of r==1, and is needed also because on
      some systems (SunOS!) `rand()' may return a value larger than RAND_MAX */
+    // DISCUSSME: what do we do here?
   lua_Number r = (lua_Number)(rand()%RAND_MAX) / (lua_Number)RAND_MAX;
```
PhantomHoover, line 185: Replace with a call to `rand32f()`, maybe?

lib/lua/lmathlib.c
```diff
@@ -207,6 +208,7 @@ static int math_random (lua_State *L) {
 
 
 static int math_randomseed (lua_State *L) {
+    // DISCUSSME: what do we do here?
   srand(luaL_checkint(L, 1));
```
PhantomHoover, line 212: Probably the same thing that sexps that manipulate the RNG seed do.

PhantomHoover (commented):
I've weighed in on most of the DISCUSSME items. As noted on Discord, I'm not keen on the verbosity of having a full `util::Random::` namespace qualification everywhere unless it's truly necessary.

PhantomHoover:
As for `static_rand`, its purpose is more analogous to a hash function than a random number generator. It should work absolutely fine with this change, so I'd say leave it alone for now. Probably forever, honestly: the way it works is stupid but it's called all over the engine and it seems to work fine on all platforms.

code/scripting/api/libs/base.cpp
```diff
@@ -61,26 +62,27 @@ ADE_FUNC(rand32,
...
+	"Calls FSO's Random::next() function and transforms the result to a float.  If called with no arguments, returns a random float from [0.0, 1.0).  If called with one argument, returns a float from [0.0, max).",
 	"number",
 	"A random float")
 {
 	float _max;
 	int numargs = ade_get_args(L, "|f", &_max);
 
-	float result = static_cast<float>(rand32()) / static_cast<float>(0x7fffffff);
+	// DISCUSSME: is division at the level of 2**31 ok for floats?
+	float result = static_cast<float>(::util::Random::next()) / static_cast<float>(::util::Random::MAX_VALUE);
```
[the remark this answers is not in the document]
jg18: yup and my experiments with `float` have shown that `float` is precise enough that it can handle `1.0/(2**31-1)` (to put it in Python). I'll use `RAND_MAX_1f` instead of division, since that's what's done elsewhere.

lib/libpcp/src/pcp_event_handler.c
```diff
@@ -61,6 +61,7 @@
 
 #define MIN(a, b) (a<b?a:b)
 #define MAX(a, b) (a>b?a:b)
+// DISCUSSME: what should we do here?
```
PhantomHoover, line 64: This is a 3rd-party library we've vendored into the FSO source tree and it appears to only be used to support port forwarding in the multiplayer code. None of this should be changed at all.

lib/lua/lmathlib.c
PhantomHoover: Maybe leave as-is and translate the existing logic to use `next()` and `MAX_VALUE` actually. The cast to `lua_Number` before the division here is actually significant — `lua_Number` is a typedef for `double` and so the calculation is being performed with greater precision.

code/starfield/starfield.cpp
JohnAFernandez: I'm on the side of matching the code to the comments here.

lib/lua/lmathlib.c
asarium: This is external code. It should not be changed.

code/debris/debris.cpp
```diff
@@ -460,15 +460,15 @@ object *debris_create(object *source_obj, int model_num, int submodel_num, vec3d
...
 		}
 		else if (vaporize) {
 			db->model_num = Debris_vaporize_model;
-			db->submodel_num = (myrand() >> 4) % Debris_num_submodels;
+			db->submodel_num = (util::Random::next() >> 4) % Debris_num_submodels;
 		}
 		else {
 			db->model_num = Debris_model;
-			db->submodel_num = (myrand() >> 4) % Debris_num_submodels;
+			db->submodel_num = (util::Random::next() >> 4) % Debris_num_submodels;
```
asarium, line 471: These three cases look suspiciously like they depend on the old maximum value. These could be changed to make use of the range limiting built into `Random::next()`.

code/globalincs/pstypes.h
```diff
@@ -17,9 +17,11 @@
 #include "windows_stub/config.h"
 #include "globalincs/scp_defines.h"
 #include "globalincs/toolchain.h"
+#include "utils/Random.h"
 #include "utils/strings.h"
 
 #include <cstdio>    // For NULL, etc
+// DISCUSSME: with rand() gone, should we reconsider how much we need to include stdlib.h/cstdlib?
```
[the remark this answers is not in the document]
asarium: All these headers are pulled in via the precompiled header anyway so they are included everywhere.

lib/libpcp/src/pcp_event_handler.c
JohnAFernandez: Only @notimaginative would know

code/network/multiui.cpp
```diff
@@ -2734,7 +2734,8 @@ void multi_sg_init_gamenet()
 	
 	Net_player->tracker_player_id = Multi_tracker_id;
 
-	Multi_sg_netgame->security = (rand() % 32766) + 1;			// get some random security number	
+	// DISCUSSME: do we want security to remain within such a narrow range? Out of scope?
+	Multi_sg_netgame->security = util::Random::next(32766) + 1;			// get some random security number
```
asarium, line 2738: Since this is a straight port of the old functionality this is fine as it is. We could increase the range in a separate PR which would also make that change more independent.

code/ship/shiphit.cpp
```diff
@@ -1048,7 +1048,8 @@ void ship_hit_sparks_no_rotate(object *ship_objp, vec3d *hitpos)
 
 	int n = ship_p->num_hits;
 	if (n >= MAX_SHIP_HITS)	{
-		n = rand() % MAX_SHIP_HITS;
+		// DISCUSSME: was the minimum really intended to be 0?
+		n = util::Random::next(MAX_SHIP_HITS);
```
asarium, line 1052: Since 0 is explicitly handled below I think this is fine.

code/utils/Random.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <cstdint>
+
+#include "globalincs/pstypes.h"
+
+namespace util {
+namespace {
+template <typename RngType>
+class RandomImpl {
+public:
+void seed(const unsigned int val) {
```
asarium, line 15: Please indent class member functions by one.

code/utils/RandomCCompat.h
```diff
@@ -0,0 +1,21 @@
+#ifndef SCP_RANDOM_C_COMPAT_H
+#define SCP_RANDOM_C_COMPAT_H
+
+// provides a C compatibility layer to the Random class
+
+#ifdef __cplusplus
+extern "C" {
```
asarium, line 7: Why the `extern "C"`. All our code is C++ anyway so this seems to be unnecessary.

lib/libpcp/src/pcp_event_handler.c
notimaginative: Third party stuff should definitely *not* be changed if at all possible. It just makes it more difficult to merge in upstream changes later on.

And the PCP stuff is covered by multiple RFCs so rng usage here can be assumed to be following the spec.

test/src/math/test_vecmat.cpp
```diff
@@ -38,6 +39,7 @@ class VecmatTest : public test::FSTestFixture {
 	void SetUp() override 
 	{ 
 		test::FSTestFixture::SetUp(); 
+		// DISCUSSME: what do we do here?
```
asarium, line 42: Tests should probably have a constant seed to make them predictable. If that value doesn't work for the new system it could be changed though.

asarium (changes requested):

lib/libpcp/src/pcp_utils.h
```diff
@@ -219,8 +219,10 @@ static inline void createNonce(struct pcp_nonce *nonce_field)
     int i;
     for (i = 2; i >= 0; --i)
 #ifdef WIN32
+		// DISCUSSME: what should we do here?
         nonce_field->n[i]=htonl (rand());
 #else  //WIN32
+		// DISCUSSME: what should we do here?
         nonce_field->n[i]=htonl(random());
 #endif //WIN32
```
notimaginative, line 227: Nothing. Do nothing.

code/utils/Random.cpp
```diff
@@ -0,0 +1,76 @@
...
+int Random::next(int modulus)
+{
+	Assert(modulus > 0);
+
+	return next() % modulus;
+}
+
+int Random::next(int low, int high)
+{
+	const int range = high - low + 1;
```
TRBlount, line 66: This `const` is pointless since the variable is created and destroyed across function calls.

code/debris/debris.cpp
PhantomHoover: It's well-known C folklore that `rand()` has historically been a lot less random in its low-order bits than it should be, so I assume the bitshift here is simply to chop them out before clamping into a range with modulo. I agree this should be replaced with `util::Random::next(Debris_num_submodels)`.

code/utils/Random.cpp
asarium: No it's not. Marking variables as `const`is very useful for ensuring that later changes don't accidentally change previously constant values.

code/radar/radar.cpp
```diff
@@ -89,9 +90,10 @@ void HudGaugeRadarStd::blipDrawFlicker(blip *b, int x, int y)
 		return;
 	}
 
-	if ( rand() & 1 ) {
-		xdiff = -2 + rand()%4;
-		ydiff = -2 + rand()%4;
+	if (util::Random::next() & 1 ) {
```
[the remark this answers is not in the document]
MageKing17: Wouldn't this be equivalent to `util::Random::next(1)`, not `util::Random::next(2)`?

code/model/modelrender.cpp
```diff
@@ -850,7 +850,8 @@ void model_render_add_lightning( model_draw_list *scene, model_render_params* in
 		switch ( smi->arc_type[i] ) {
 			// "normal", FreeSpace 1 style arcs
 		case MARC_TYPE_NORMAL:
-			if ( (rand()>>4) & 1 )	{
+			// DISCUSSME: given >> 4, did this line assume RAND_MAX was 0x7ff?
+			if ( (util::Random::next()>>4) & 1 )	{
```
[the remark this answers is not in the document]
MageKing17: This should be `util::Random::next(2)`.

lib/libpcp/src/pcp_utils.h
jg18: So I shouldn't integrate our third-party libraries (specifically, libpcp and liblua) into this PR? Should they continue to use `rand()`/`srand()`, then?

code/utils/RandomCCompat.h
jg18: The reason I added the C compatibility layer was to allow us to change the `rand()`/`srand()` calls in the `lib` subfolder (the relevant libraries are written in C) to use the new `Random` class.

But if the team thinks I should leave the files in `lib` alone (that seems to be @notimaginative 's opinion), I'll scrap the C compatibility layer. What do you think?

I don't think libRocket can be integrated, anyway, since we use it via git submodule.

asarium: I agree with @notimaginative, stuff in lib should not be changed so this can be removed.

test/src/math/test_vecmat.cpp
jg18: I updated `RandomImpl::seed()` to follow `srand()` semantics, so that if you pass 1 to `Random::seed()`, the RNG will use its default seed, which for `std::mt19937` at least is part of the C++ standard and thus should be the same across platforms.

https://www.cplusplus.com/reference/random/mersenne_twister_engine/seed/

code/network/multiui.cpp
jg18: Yup agreed that the behavior here shouldn't change in this PR. I can open an issue to discuss whether we should expand the range of valid values for `Multi_sg_netgame->security `.

code/debris/debris.cpp
jg18: I agree with @PhantomHoover  's explanation as the most likely one, in which case I also agree that we can get rid of these right shifts. They appear in a number of places in the code.

For all I know, MT may also be less random in its low order bits, but IMO even if so, the extra bit twiddling isn't worth it.

lib/libpcp/src/pcp_event_handler.c
jg18: @notimaginative sure, makes sense. I read this comment after reading your "Do nothing" comment (reading these out of order) so I'll scrap the C compatibility layer.

lib/libpcp/src/pcp_utils.h
jg18: Never mind, I just read your other comment about not messing with third-party libraries, so I'll leave those untouched. I figured it was worth asking, though.

code/utils/Random.cpp
jg18: Agreed with @asarium about the general principle, but I'm not gonna make a fuss over this one. If the team (or a majority) wants the `const` removed, I'll drop it.

lib/lua/lmathlib.c
jg18: As per @notimaginative 's comments, I'll be removing all changes I made in the `lib` subfolder.

lib/libpcp/src/pcp_utils.h
notimaginative: Sorry, I should have been more descriptive here. Made the assumption that you would have read my earlier comment first for the reason and just meant to tag this change something to be ignored.

code/particle/particle.cpp
```diff
@@ -551,7 +552,8 @@ namespace particle
 		n2 = (pe->num_high * percent) / 100;
 
 		// How many to emit?
-		n = (rand() % (n2 - n1 + 1)) + n1;
+		// DISCUSSME: does this rewrite look correct?
+		n = ::util::Random::next(n1, n2);
```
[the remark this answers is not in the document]
jg18: True and, more importantly, it's conceptually what the code was trying to express.

code/globalincs/pstypes.h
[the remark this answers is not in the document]
jg18: Thanks @asarium; that's good to know. In that case, I'll get rid of any removals of including stdlib.

code/model/modelrender.cpp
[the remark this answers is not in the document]
jg18: Bitwise `&` likely uses fewer CPU instructions than `%`, but maybe compiler optimizers are smart enough to know that if `n` is a power of 2 then `x % n == x & (n-1)`.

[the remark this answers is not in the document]
jg18: The code has so many instances of
```
if (some_flavor_of_rand() & 1)
```
that I'm considering adding a new function
```
bool Random::flip_coin()
```
that just returns `next() & 1`.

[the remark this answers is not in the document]
jg18: Or perhaps `flip_coin()` should return `next() < MAX_VALUE/2`; that would be fairer and also easier to read.

I'm open to some name other than `flip_coin()` but I don't have a better idea at the moment.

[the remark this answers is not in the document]
MageKing17: Actually, having just checked gcc 10.2 in compiler explorer, `& 1` is a lot more efficient than `% 2`, with the former being a single assembly instruction while the latter is five.

lib/libpcp/src/pcp_utils.h
jg18: It's cool, and agreed that tinkering with third-party library code is generally a bad idea, for the reasons you mentioned.

In this particular case, I wasn't sure if the game would run properly if Lua (no idea what libpcp even does) was still using `rand()`/`srand()`.

code/radar/radar.cpp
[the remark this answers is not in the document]
jg18: I changed this and similar places to use a new function `Random::flip_coin()` which is conceptually what this code is trying to do.

code/model/modelrender.cpp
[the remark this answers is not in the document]
jg18: Yeah `%` is pretty expensive, as arithmetic operations go.

The new `Random::flip_coin()` that I now use here is implemented using `<=` which I would think also uses just one instruction.

code/ship/shiphit.cpp
jg18: Ok, good. I was just on the lookout for potentially odd-looking code

code/ai/aicode.cpp
```diff
@@ -9756,7 +9757,8 @@ void remove_farthest_attacker(int objnum)
 				aip->ignore_objnum = aip->target_objnum;
 				aip->ignore_signature = Objects[aip->target_objnum].signature;
 				aip->ai_flags.set(AI::AI_Flags::Temporary_ignore);
-				aip->ignore_expire_timestamp = timestamp(((myrand() % 10) + 20) * 1000);	//	OK to attack again in 20 to 24 seconds.
+				// DISCUSSME: should we fix the code to match the comment? Or the other way around?
```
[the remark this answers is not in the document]
jg18: Ok, fair point. I did switch to using the two-arg version of `Random::next()` where possible, perhaps overreaching, although it does make the code easier to read IMO.

code/starfield/starfield.cpp
```diff
@@ -1750,22 +1751,24 @@ void stars_draw_debris()
...
+			d->pos.xyz.y = f2fl(Random::next() - Random::HALF_MAX_VALUE);
+			d->pos.xyz.z = f2fl(Random::next() - Random::HALF_MAX_VALUE);
 
 			vm_vec_normalize(&d->pos);
 
 			vm_vec_scale(&d->pos, MAX_DIST);
 			vm_vec_add2(&d->pos, &Eye_position );
 			d->active = 1;
+			// DISCUSSME: is the following line correct? What was the author trying to do?
 			d->vclip = i % MAX_DEBRIS_VCLIPS;	//rand()
```
jg18, line 1764: I wonder if the author intended what would now be expressed as
```
d->vclip = Random::next(MAX_DEBRIS_VCLIPS);
```
but I take it I should leave this line alone, including leaving the existing comment as is?

jg18: This is the last remaining `DISCUSSME` btw.

code/network/multiui.cpp
jg18: Opened Issue #3336 to discuss this question.

code/starfield/starfield.cpp
asarium: I would say leave it as it is. We can change is later if needed but this is working now so no need to change anything around this.

asarium (approved):
LGTM

asarium:
Since we want to do a release "soon" this should probably not be merged until that is out to avoid last minute regressions.

code/globalincs/pstypes.h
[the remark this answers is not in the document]
PhantomHoover: OK, so let's remove redundant inclusions when we know they aren't necessary.

PR #3333, opened 2021-03-27
log missing XSTR strings only once

Goober5000:
The "No XSTR entry in strings.tbl for index %d" message was previously printed to the log every time the string was accessed, which for commonly-used strings will clutter up the log.  This makes the message only appear once per missing index.


wookieejedi (approved):
Good idea.

PR #3334, opened 2021-03-27
Clarify naming of vm_vec_rotate and vm_vec_unrotate

Baezon:
This can wait until after the stable, if desired, but it's just a small renaming change.

While working on something unrelated I realized that I would need to "`vm_vec_rotate`" a vector through a matrix when the situation was not at all in fact rotation, and indeed this situation situations happens several times throughout the code, so I think it's less confusing to have the more mathematically general names `vm_matrix_x_vector` and `vm_vector_x_matrix`, so the image of "rotation" can be avoided when necessary and rotate and unrotate can be macros of these functions when "rotation" *should* be evoked.

Now, which you want to call which is down to convention but it needs to be consistent with regards to associativity, but the convention used matches `vm_matrix_x_matrix`. i.e. (MatrixA x MatrixB) x VectorA (matrix_x_matrix followed by rotate) = MatrixA x (MatrixB x VectorA) (two rotates).

The already existing operator* overload (which had yet to be used) for matrix * vector, would *not* have followed this convention which is why it was changed.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -835,7 +835,7 @@ matrix *vm_vector_2_matrix_norm(matrix *m, const vec3d *fvec, const vec3d *uvec,
 //
 // Goober5000: FYI, the result of rotating a normalized vector through a rotation matrix will
 // also be a normalized vector.  It took me awhile to verify online that this was true. ;)
-vec3d *vm_vec_rotate(vec3d *dest, const vec3d *src, const matrix *m)
+vec3d *vm_matrix_x_vector(vec3d *dest, const matrix *m, const vec3d* src)
 {
 	Assert(dest != src);
 
@@ -861,7 +861,7 @@ vec3d *vm_vec_rotate(vec3d *dest, const vec3d *src, const matrix *m)
 //
 // Goober5000: FYI, the result of rotating a normalized vector through a rotation matrix will
 // also be a normalized vector.  It took me awhile to verify online that this was true. ;)
-vec3d *vm_vec_unrotate(vec3d *dest, const vec3d *src, const matrix *m)
+vec3d *vm_vector_x_matrix(vec3d *dest, const vec3d *src, const matrix *m)
 {
 	Assert(dest != src);
 
@@ -1141,7 +1141,7 @@ float find_nearest_point_on_line(vec3d *nearest_point, const vec3d *p0, const ve
 
 	vm_vec_outer_product(&mat, &norm);
 
-	vm_vec_rotate(&projected_point, &xlated_int_pnt, &mat);
+	vm_matrix_x_vector(&projected_point, &mat, &xlated_int_pnt);
 	vm_vec_add(nearest_point, &projected_point, p0);
 
 	dot = vm_vec_dot(&norm, &projected_point);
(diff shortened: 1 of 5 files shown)


Goober5000:
Hmm.  I'm skeptical of this, and I would prefer that the names be kept as they are, but you make a good point, and the addition of the #define's mitigates my skepticism somewhat.

A bigger problem is that if you are going to interpret the function in a general way, you really need to scrutinize the math to make sure it's actually correct.  As I recall, there was a big SCP discussion a few years ago which concluded that rotate and unrotate are actually *swapped*.  Rotate should have been called unrotate and vice-versa.  We considered renaming all the functions but concluded that a) it would have been a huge job, b) it would have led to confusion when using the functions in new situations, and c) it would have led to massive confusion when comparing old code against new.

So if there is a standard `matrix_x_vector` formula, make sure the function actually matches the formula and isn't getting the elements backwards, or switching row-major/column-major format, or transposing the matrix, or whatever.

Baezon:
What operation you want to call a 'rotation' or an....  "unrotation" is a matter of convention, the math itself is just a bunch of multiplications and could be interpreted either way. If you transpose all the matrices from external data and swap the code for rotate and unrotate literally everything would be the same.

So really its not actually a matter of correctness imo but rather about being consistent about our terminology. This makes it more explicit what we mathematically consider a "rotation" or an "unrotation", and makes it consistent in associativity with matrix x matrix.

PhantomHoover:
For what it's worth, I would definitely call `vm_vec_rotate`/`vm_matrix_x_vector` a 'forward' rotation based on the naming of the `rvec`, `uvec` and `fvec` components. It sends the x-axis to `rvec`, the y-axis to `uvec` and the z-axis to `fvec`, and the result of `vm_matrix_x_vector(dest, A, x)` is the vector Ax. It's all consistent within the engine's own conventions and standard mathematical notation, as far as I can tell. `vm_vec_unrotate(dest, x, A)` = `vm_vector_x_matrix(dest, x, A)` is calculating ((x^T)A)^T = (A^T)x; if A is a rotation matrix then A^T = A^-1, which is why it works as an inverse. I'm not convinced `vm_vector_x_matrix` is actually a better name for this operation: it's not calculating xA, which isn't even a well-defined operation. Maybe something like `vm_transpose_x_vector` would be clearer.

PhantomHoover:
I typed up a whole reply here, got confused, deleted it, but I think I understand now. FSO matrices are stored in row-major order for some reason, but vectors are still column vectors, so `vm_vec_rotate(dest, x, A)` = `vm_matrix_x_vector(dest, A, x)` is calculating Ax. `vm_vec_unrotate(dest, x, A)` is actually calculating (A^T)x, which happens to be an inverse for rotation matrices as for them, A^T = A^-1. I don't think `vm_vector_x_matrix` is a good name for this operation as xA is not even a well-defined matrix multiplication, so I think something like `vm_transpose_x_vector` would be clearer.

JohnAFernandez:
It sounds like this may be the best the way it is, since everyone more or less understands the way it currently works and we should just add some comments to clarify what rotate and unrotate actually do?

Baezon:
Well, I feel less strongly about this now when I first made the PR, but PH has since come up with an alternative implementation of this which could be preferable.

PhantomHoover:
A mere 2.5 months after doing the work, I have actually submitted my alternative implementation of this work for review at #3470.

Baezon:
Obselete and redundant with #3470

PR #3337, opened 2021-03-28
Fix a rare but eventual crash in the interpolation code.

JohnAFernandez:
I don't know when this was introduced but the bug is extremely obvious once someone hit the crash.  It should just be using the timestamp that the packet was received, not trying to put that timestamp into a vector whose index is server frame numbers.


notimaginative (approved):
LGTM, and fixed the crash too.

PR #3339, opened 2021-03-28
Make sure the Netgame state is set to prevent a crash

JohnAFernandez:
If the client tries to respawn a player when the server is trying to end the game it will lead to a crash on the client side.  This sets the client's game to the correct state when an end game packet is received and adds extra checks before respawning.

I am hoping for a little more testing before this is merged.

### DIFF
--- a/code/network/multi_respawn.cpp
+++ b/code/network/multi_respawn.cpp
@@ -644,9 +644,14 @@ void multi_respawn_process_packet(ubyte *data, header *hinfo)
 		p_object *pobjp;
 
 		GET_USHORT( net_sig );
-		pobjp = mission_parse_get_arrival_ship( net_sig );
-		Assert( pobjp != NULL );
-		multi_respawn_ai( pobjp );
+
+		// only attempt to respawn if we are in the mission
+		if (!(Net_player->flags & NETINFO_FLAG_WARPING_OUT)) {
+			pobjp = mission_parse_get_arrival_ship(net_sig);
+			Assert(pobjp != NULL);
+			multi_respawn_ai(pobjp);
+		}
+
 		break;		
 
 	case RESPAWN_BROADCAST:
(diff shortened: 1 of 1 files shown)


code/network/multi_respawn.cpp
```diff
@@ -644,9 +644,14 @@ void multi_respawn_process_packet(ubyte *data, header *hinfo)
...
 		GET_USHORT( net_sig );
-		pobjp = mission_parse_get_arrival_ship( net_sig );
-		Assert( pobjp != NULL );
-		multi_respawn_ai( pobjp );
+
+		// only attempt to respawn if we are in the mission
+		if (Netgame.game_state == NETGAME_STATE_IN_MISSION) {
+		pobjp = mission_parse_get_arrival_ship(net_sig);
+		Assert(pobjp != NULL);
+		multi_respawn_ai(pobjp);
```
notimaginative, line 652: This needs to be indented.

code/network/multimsgs.cpp
```diff
@@ -4062,6 +4062,7 @@ void process_endgame_packet(ubyte * /*data*/, header *hinfo)
 
 		// make sure we're not already in the debrief state
 		if((gameseq_get_state() != GS_STATE_DEBRIEF) && (gameseq_get_state() != GS_STATE_MULTI_DOGFIGHT_DEBRIEF)){
+			Netgame.game_state = NETGAME_STATE_ENDGAME;
```
notimaginative, line 4065: Only the server should update `Netgame.game_state` so this is might cause issues since there is a small window in which it could be reset. I'm sure there are probably a few other places in the code where it get set like this, but it's best to try and avoid doing it.

I believe that `multi_warpout_all_players()` already sets a `Net_player` flag to indicate warpout so it might be safer to check that instead.

notimaginative (changes requested):

code/network/multimsgs.cpp
JohnAFernandez: Ok, changed, but I do still want to do a testing session before enabling it to be merged.

notimaginative (approved):
Looks good to me. Approved pending your tests to verify it's working the way you want.

PR #3340, opened 2021-03-28
make num-valid-arguments work for the for-* sexps

Goober5000:
This adds a hook into the for-* handlers to return the number of valid arguments.  Fixes a bug reported on Discord.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -10001,7 +10001,7 @@ int eval_in_sequence(int arg_handler_node, int condition_node)
 }
 
 // Goober5000
-int eval_for_counter(int arg_handler_node, int condition_node)
+int eval_for_counter(int arg_handler_node, int condition_node, bool just_count = false)
 {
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -10249,14 +10258,34 @@ int sexp_num_valid_arguments( int n )
 
 	arg_handler = get_handler_for_x_of_operator(n);
 
+	// thanks to woutersmits for finding this loophole
```
JohnAFernandez, line 10261: Which loophole!  My only issue with the PR is we don't know why they need special handling by just looking at the comments.

JohnAFernandez (changes requested):

code/parse/sexp.cpp
Goober5000: Merely that using the SEXP outside of an -argument operator will cause a crash due to the node being -1.  I can make the comment more explicit.

JohnAFernandez (approved):

PR #3344, opened 2021-03-30
Fix regression with huge/small only targeting

Baezon:
The `+Targeted by 'Huge' weapons and Ignored by 'small only' weapons:` flag took over the the job of the simple `is_big_or huge()` hardcoded class flag when it comes to targeting for those weapons, and suitable defaults were added to the built-in objecttypes.tbl. Unfortunately, since this defaults to false otherwise, if a mod had their own objecttypes.tbl and used the same name as a hardcoded big class they before would've benefited from this hardcoded behavior, but now must include this new flag in order to maintain targeting behavior, breaking backwards compatability.

This checks to see if it is a .tbl and the same name as a hardcoded big|huge class and if so, sets the matching default. I'm unhappy with this solution, because its requires more references to the hardcoded classes, and because no other ship type flag does this, even though it would have certainly been a problem for them when they were implemented. But it's too large a change to leave as is, and I'm not sure how else to handle it.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -5260,7 +5260,7 @@ static void parse_ship_type(const char *filename, const bool replace)
 		stricmp(stp->name, "corvette") == 0 || stricmp(stp->name, "gas miner") == 0 || stricmp(stp->name, "awacs") == 0)
 		big_ship = true;
 
-	if (stricmp(stp->name, "capital") == 0 || stricmp(stp->name, "supercap") == 0 || stricmp(stp->name, "knossos device") == 0 || stricmp(stp->name, "drydock") == 0)
+	if (stricmp(stp->name, "capital") == 0 || stricmp(stp->name, "super cap") == 0 || stricmp(stp->name, "knossos device") == 0 || stricmp(stp->name, "drydock") == 0)
 		huge_ship = true;
 
 	//Okay, now we should have the values to parse
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -5207,6 +5207,7 @@ static void parse_ship_type(const char *filename, const bool replace)
 		create_if_not_found = false;
 	}
 
+	bool first_time;
 	int idx = ship_type_name_lookup(name_buf);
 	if (idx >= 0)
 	{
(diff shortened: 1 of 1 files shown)


code/ship/ship.cpp
```diff
@@ -5251,6 +5254,15 @@ static void parse_ship_type(const char *filename, const bool replace)
 		Warning(LOCATION, "Bad ship type name in %s\n\nUsed ship type is redirected to another ship type.\nReplace \"%s\" with \"%s\"\nin %s to fix this.\n", filename, stp->name, ship_type, filename);
 	}
 
+	bool big_ship = false;
+	bool huge_ship = false;
+	if (stricmp(stp->name, "cruiser") == 0 || stricmp(stp->name, "freighter") == 0 || stricmp(stp->name, "transport") == 0 ||
+		stricmp(stp->name, "corvette") == 0 || stricmp(stp->name, "gas miner") == 0 || stricmp(stp->name, "awacs") == 0)
+		big_ship = true;
+
+	if (stricmp(stp->name, "capital") == 0 || stricmp(stp->name, "supercap") == 0 || stricmp(stp->name, "knossos device") == 0 || stricmp(stp->name, "drydock") == 0)
```
Goober5000, line 5263: The `stp->name` field will be "knossos", not "knossos device".  Some of the strings differ between ship classes and ship types, so there is an if() block that standardizes them a few lines up.  You might want to use the `ship_type` string instead of `stp->name`.

Goober5000 (changes requested):
Looks good except for an idiosyncratic bit.

code/ship/ship.cpp
Baezon: I'm not sure I follow, thats the entire reason why this was put after the if block, so it could be guaranteed all types that got to this point had standardized names.

Goober5000: Oh.  O.o  I guess I had it confused with what happens inside the `ship_type_name_lookup` function.  I see that this is used slightly differently.

That being said, a change is still needed here.  Replace "supercap" with "super cap".

Baezon: Oh man that's super confusing, I got "supercap" from the ship class flags. The hardcoded ship-info flag is "Supercap", the class flag to enable it is "suprecap", but the name of the builtin ship type is "Super cap" and the table will complain if you tried to define "Supercap" because it should be "Super cap".

Goober5000 (approved):
That looks good now.

PR #3346, opened 2021-03-31
Add a safety check to the $On Frame: Hook

### LINKED ISSUE #3007: Accessing the player's target every frame in lua script causes Lua Error when restarting after death
As the title says, I have a script that does things every frame based on the player's current target. When the player dies, I have safety checks in place that detects the player's target isn't valid on the last frame of the game state, so that's all good. But if the player hits "Quick Start Mission", on the first frame of the next game state `hv.Player` reports as a valid object of breed Ship, but trying to touch `hv.Player.Target` in any way, even to check if it exists, gets a `Lua Error: Could not find index 'Target' in type 'object'`. 

This happens if you quick start after dying. It does not if you restart without dying, or if you return to the mainhall after dying and then restart the mission.

I'm really not sure how to safety-check around this, and CTD on quickstart is a pretty painful flaw in this script. 

This seems like something @asarium might have insight into.

I built a repro script table. Just drop it in, die, and quickstart.
[restart-crash-sct.zip](https://github.com/scp-fs2open/fs2open.github.com/files/5727813/restart-crash-sct.zip)

JohnAFernandez:
So #3007 is a larger bug than just trying to access the player's target.  It happens when using a combination of the use of `$State: GS_STATE_GAME_PLAY` and `$On frame:` 

The real problem is that the game state is not changed from `GS_STATE_GAME_PLAY` during the mission load when restarting the mission.  So the script thinks it's in the mission, tries to access the target and then crashes.  But this won't be the only bug.  Any script that tries to access data on Frame will run into possible problems on mission restart.

This change checks that if we are currently in `GS_STATE_GAME_PLAY` that `Game_Mode` is set to `GM_IN_MISSION`.

It might be a slightly complicated way to do it, but it may be the only way possible without making a new game state for mission reload and dealing with all the headaches that would come with that.

Closes #3007

### DIFF
--- a/code/graphics/2d.cpp
+++ b/code/graphics/2d.cpp
@@ -41,6 +41,7 @@
 #include "scripting/scripting.h"
 #include "tracing/tracing.h"
 #include "utils/boost/hash_combine.h"
+#include "gamesequence/gamesequence.h"
 
 #ifdef WITH_OPENGL
 #include "graphics/opengl/gropengl.h"
@@ -2547,7 +2548,8 @@ void gr_flip(bool execute_scripting)
 	executor::OnFrameExecutor->process();
 
 	// m!m avoid running CHA_ONFRAME when the "Quit mission" popup is shown. See mantis 2446 for reference
-	if (execute_scripting && !popup_active()) {
+	// Cyborg - A similar bug will occur when a mission is restarted so check for that, too.
+	if (execute_scripting && !popup_active() && GameState_Stack_Valid() && !((gameseq_get_state() == GS_STATE_GAME_PLAY) && !(Game_mode & GM_IN_MISSION))) {
 		TRACE_SCOPE(tracing::LuaOnFrame);
 
 		// WMC - Do conditional hooks. Yippee!


Goober5000 (commented):
Good sleuthing.  But is the `gameseq_get_depth()` check necessary?  I think it will always pass in this case, and so should be taken out.  No other state checks in the codebase use it.

JohnAFernandez:
The check is necessary because the game state is -1 at the start of FSO and it triggers an assert in `gameseq_get_state()`.

Goober5000:
In that case what about using `GameState_Stack_Valid()`?

Goober5000 (changes requested):
Or even better - change `gameseq_get_state()` to return `GS_STATE_INVALID` if the depth is -1.

JohnAFernandez:
> 
> 
> Or even better - change `gameseq_get_state()` to return `GS_STATE_INVALID` if the depth is -1.

That's risky and could make it harder to determine if there are game state issues in the future.  It's also out of the scope of this PR.  I did switch to `GameState_Stack_Valid()`.  Easier to read and better compartmentalizes the code.

Goober5000 (approved):

Goober5000:
I'll open a different PR for `GS_STATE_INVALID`.

PR #3347, opened 2021-03-31
Make fireballs default to 1 LOD

Baezon:
The wiki claims this is the case, but it isn't, but it should be. Fireballs without at least 1 LOD will not function, even if the rest of their data is good.

### DIFF
--- a/code/fireball/fireballs.cpp
+++ b/code/fireball/fireballs.cpp
@@ -287,11 +287,13 @@ static void parse_fireball_tbl(const char *table_filename)
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


JohnAFernandez (commented):
Looks like this makes table parsing less onerous, but is the table modular?

Baezon:
Whoops. Duh.

JohnAFernandez (approved):

Goober5000 (approved):
Good call.  This isn't initialized anywhere else that I can see.

PR #3348, opened 2021-03-31
Strip hashes from explicitly defined display names in tables.

TRBlount:
This was an oversight from the last changing relating to alt names. This will apply the same hash stripping rules to display/alt names as are applied to normal ship names when generating a display name automatically.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -2654,6 +2654,8 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 	if (optional_string("$Alt name:") || optional_string("$Display Name:"))
 	{
 		stuff_string(sip->display_name, F_NAME, NAME_LENGTH);
+		end_string_at_first_hash_symbol(sip->display_name, true);
+		consolidate_double_characters(sip->display_name, '#');
 		sip->flags.set(Ship::Info_Flags::Has_display_name);
 	}
 
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -895,7 +895,8 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		// if this name has a hash, create a default display name
 		if (get_pointer_to_first_hash_symbol(wip->name)) {
 			strcpy_s(wip->display_name, wip->name);
-			end_string_at_first_hash_symbol(wip->display_name);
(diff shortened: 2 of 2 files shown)


JohnAFernandez (approved):
Knock another one down.

Goober5000 (changes requested):
Wait, what?  This isn't an oversight at all; this is how display names with hashes are supposed to be specified.

TRBlount:
The issue is that people are confused by the terminology and expect `$Display Name:` to strip as well. I mentioned this as well, but some discussion made it seem like not stripping hashes from `$Display Name:` was incorrect behavior based on previous PRs. How should we handle this then?

MageKing17:
It's a backwards-incompatibility; mods were made with the assumption that the display name would be hash-stripped, and now they aren't. We don't have to worry about breaking retail in this instance, but it's really not that much effort to make the behavior both consistent with people's expectations *and* with how mods were made for years.

Goober5000:
I caught up on the conversation.  Given the amount of confusion this has caused, I agree with returning to hash truncation even on display names, and adding the double-hash escape sequence.  2018 wasn't *that* long ago, but it's long enough, and Release 19 would have had the truncation behavior.

Goober5000 (changes requested):
Additional steps required:

1) Truncate hashes on weapon classes as well
2) Add the double-hash escape sequence for ship class display names
3) Add the double-hash escape sequence for weapon class display names

Goober5000:
Once #3353 is merged, the `consolidate_double_characters` function can be reused for this PR.

Goober5000:
And it's merged.

TRBlount:
> Additional steps required:
> 
> 1. Truncate hashes on weapon class display names as well
> 2. Add the double-hash escape sequence for ship class display names
> 3. Add the double-hash escape sequence for weapon class display names

@Goober5000 This should be handled in the latest version of the PR.

Goober5000 (approved):
Looks good.

Goober5000:
> Also, we might wish to extend `consolidate_double_characters()` to handle multiple characters at once (ie. make the second parameter `char*` instead of `char`).

When would we need to do that?

TRBlount:
Not sure yet tbqh. It was just an idea I had since functions like that often have the ability to handle multiple types of characters. (See: Boost libs string parsing functions)

PR #3350, opened 2021-04-01
Some fixes to nebula fog set up

Baezon:
The mission-specific fog near/far/density values need to set before being used, but this is only done by the `neb2_get_fog_visibility(object *obj)` function so the other overload (that only takes a position) may not have this information when necessary. Although this is unlikely (as it is common for both functions to be called several times a frame), it should really be part of `neb2_post_level_init`. 

Pulling that out was basically the only real difference between the two so it can just be removed. Also puts a flag check on trail fading since it is new.


Goober5000 (approved):
Looks good to me too

PR #3351, opened 2021-04-01
Add profiling trace for scripting hooks

PhantomHoover:
This will separate out Lua scripting hooks from the rest of our performance traces, allowing us to separate time spent in engine code and in modder-supplied scripts.

### DIFF
--- a/code/tracing/categories.cpp
+++ b/code/tracing/categories.cpp
@@ -13,6 +13,7 @@ bool Category::usesGPUCounter() const {
 }
 
 Category LuaOnFrame("LUA On Frame", true);
+Category LuaHooks("LUA hooks", true);
 
 Category DrawSceneTexture("Draw scene texture", true);
 Category UpdateDistortion("Update distortion", true);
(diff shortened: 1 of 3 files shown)


code/tracing/categories.cpp
```diff
@@ -13,6 +13,7 @@ bool Category::usesGPUCounter() const {
 }
 
 Category LuaOnFrame("LUA On Frame", true);
+Category LuaHooks("LUA hooks", false);
```
asarium, line 16: Since Lua can trigger graphics intesive stuff you might want to turn this into a GPU trace as well.

asarium (approved):
One optional suggestion, good to go as it is.

code/tracing/categories.cpp
PhantomHoover: Honestly I don't really know how our graphics tracing works and if this would be a reasonable thing to do, so I'd rather leave it like this unless there's a clear benefit.

asarium: I think it would be reasonable to do that. The additional overhead is relatively minor and it might be useful at some point so it would be better to have it than not have it.

PhantomHoover: OK, I switched it on but on your head be it!!

PR #3354, opened 2021-04-02
actually use GS_STATE_INVALID

Goober5000:
Make `gameseq_get_state` return GS_STATE_INVALID if the stack is invalid.

With this change, the check for stack validity can be removed in `gr_flip`.

Follow-up to #3346.


wookieejedi (approved):
Looks good to me.

PR #3357, opened 2021-04-03
Only close common_select when we should

notimaginative:
Address various issues that cropped up when the various pre-briefing states were fixed for multi. This should fix the various places where `common_select_close()` was getting called when it shouldn't, or not getting called at all.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -4954,6 +4954,40 @@ void game_process_event( int current_state, int event )
 	}
 }
 
+static bool going_to_briefing_state(int new_state)
+{
+	switch (new_state) {
+		case GS_STATE_TEAM_SELECT:
+		case GS_STATE_SHIP_SELECT:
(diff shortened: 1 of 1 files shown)


notimaginative:
This should fix the briefing music problem as well as various related issues due to `common_select_close()` actually doing something now in more places.

JohnAFernandez (approved):
So I had been going back and forth on this PR because it is such a big change to the gamestate code.  But this is really how it should have been handled in the first place. I talked to Taylor about it, and he tested it thoroughly.  It also helps to note that this pr specifically adjusts some of the gamestate edits during the 21.0 multiplayer bump and should help to handle any of the bugs we might encounter with those shared SP/MP gamestates, while also cleaning up this section of the code.

While it's been a while since this was opened, someone else may want to look at this, too, even though I couldn't find anything wrong with it.

Goober5000 (approved):
I had looked at it myself, but wanted to defer to the multi people.  LGTM.

PR #3360, opened 2021-04-05
Remove string comparisons and switch to vector in script conditions

EatThePath:
String comparisons in lua hook evaluation take up a significant amount of runtime in FSO. This PR shifts those comparisons to hook creation instead of evaluation. Currently only covers the three most frequently evaluated hooks, which in profiling cleans up the vast majority of evaluation's footprint.

~~In draft because it's still pretty messy and slapdash in places, but I wanted to let people review the approach at the least.~~
Now prepared for final review.

### DIFF
--- a/code/scripting/scripting.cpp
+++ b/code/scripting/scripting.cpp
@@ -255,41 +255,61 @@ DCF(script, "Evaluates a line of scripting")
 extern char Game_current_mission_filename[];
 bool ConditionedHook::AddCondition(script_condition *sc)
 {
+	//Since string comparisons are expensive and these hooks have to be checked very frequently
+	//where possible whatever string comparison is done here and the outcome stored for later
+	//nature of value stored depends on condition type.
 	switch (sc->condition_type)
 	{
-	case CHC_STATE:
-		sc->condition_reference_index = gameseq_get_state_idx(sc->condition_string.c_str());
+	case CHC_SHIPCLASS:
+		sc->condition_cached_value = ship_info_lookup(sc->condition_string.c_str());
 		break;
-	case CHC_VERSION:
-		char buf[32];
-		sc->condition_reference_index = 0;
-		sprintf(buf, "%i.%i.%i", FS_VERSION_MAJOR, FS_VERSION_MINOR, FS_VERSION_BUILD);
-		if (stricmp(buf, sc->condition_string.c_str()) == 0)
-		{
-			sc->condition_reference_index = 1;
-		} 
-		else if(FS_VERSION_BUILD == 0) //In case some people are lazy and say "3.7" instead of "3.7.0" or something
-		{
-			sprintf(buf, "%i.%i", FS_VERSION_MAJOR, FS_VERSION_MINOR);
-			if (stricmp(buf, sc->condition_string.c_str()) == 0) {
-				sc->condition_reference_index = 1;
+	case CHC_STATE:
+		sc->condition_cached_value = gameseq_get_state_idx(sc->condition_string.c_str());
+		break;
+	case CHC_WEAPONCLASS:
+		sc->condition_cached_value = weapon_info_lookup(sc->condition_string.c_str());
+		break;
+	case CHC_OBJECTTYPE:
+		for (int i = 0; i < MAX_OBJECT_TYPES; i++) {
+			if (stricmp(Object_type_names[i], sc->condition_string.c_str()) == 0) {
+				sc->condition_cached_value = i;
+				break;
 			}
 		}
 		break;
-		
+	case CHC_VERSION:
+	{
+		char buf[32];
+		sc->condition_cached_value = 0;
+		sprintf(buf, "%i.%i.%i", FS_VERSION_MAJOR, FS_VERSION_MINOR, FS_VERSION_BUILD);
+		if (stricmp(buf, sc->condition_string.c_str()) == 0)
+		{
+			sc->condition_cached_value = 1;
+		}
+		else if (FS_VERSION_BUILD == 0) //In case some people are lazy and say "3.7" instead of "3.7.0" or something
+		{
+			sprintf(buf, "%i.%i", FS_VERSION_MAJOR, FS_VERSION_MINOR);
+			if (stricmp(buf, sc->condition_string.c_str()) == 0) {
+				sc->condition_cached_value = 1;
+			}
+		}
+		break;
+	}
 	case CHC_APPLICATION:
-		sc->condition_reference_index = 1;
+	{
+		sc->condition_cached_value = 1;
 		if (Fred_running)
 		{
 			if (stricmp("FRED2_Open", sc->condition_string.c_str()) != 0 && stricmp("FRED2Open", sc->condition_string.c_str()) != 0 && stricmp("FRED 2", sc->condition_string.c_str()) != 0 && stricmp("FRED", sc->condition_string.c_str()) != 0)
-				sc->condition_reference_index = 0;
+				sc->condition_cached_value = 0;
 		}
 		else
 		{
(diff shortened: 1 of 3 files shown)


Goober5000:
:+1:

code/scripting/scripting.cpp
```diff
@@ -510,42 +540,29 @@ bool ConditionedHook::ConditionsValid(int action, object *objp, int more_data)
...
 					int action_index = more_data;
 
-					if (action_index <= 0 || stricmp(scp->condition_string.c_str(), Control_config[action_index].text.c_str()) != 0)
+					if (action_index <= 0 || stricmp(scp.condition_string.c_str(), Control_config[action_index].text.c_str()) != 0)
 						return false;
 					break;
 				}
 			case CHC_VERSION:
 				{
 					// Goober5000: I'm going to assume scripting doesn't care about SVN revision
```
Goober5000, line 549: Remove the obsolete comment lines to prevent comment rot.

code/scripting/scripting.cpp
```diff
@@ -255,16 +255,47 @@ DCF(script, "Evaluates a line of scripting")
 extern char Game_current_mission_filename[];
 bool ConditionedHook::AddCondition(script_condition *sc)
 {
-	for(int i = 0; i < MAX_HOOK_CONDITIONS; i++)
+	switch (sc->condition_type)
 	{
-		if(Conditions[i].condition_type == CHC_NONE)
+	case CHC_STATE:
+		sc->condition_reference_index = gameseq_get_state_idx(sc->condition_string.c_str());
```
Goober5000, line 261: Is there a risk that the state index could collide with the explicitly-assigned 0 or 1 index?

EatThePath: the state index should only ever get set in this one function on initial add. I can probably add some commenting in the declaration to make that explicit.
If there are conditions where that approach doesn't work, then either that paradigm will need to be reconsidered or they can keep on working the old way. The string comparisons aren't ideal, but these three conditions made up somewhere north of 95% of the hook evaluation time in my test cases, so any of the rest remaining as they are isn't a huge problem.

code/scripting/scripting.cpp
```diff
@@ -255,16 +255,47 @@ DCF(script, "Evaluates a line of scripting")
...
+		break;
+	case CHC_VERSION:
+		char buf[32];
```
asarium, line 264: Careful with variable declarations in a switch-case. Please put the case into its own block.

code/scripting/scripting.cpp
```diff
@@ -279,20 +310,19 @@ bool ConditionedHook::AddAction(script_action *sa)
 
 bool ConditionedHook::ConditionsValid(int action, object *objp, int more_data)
 {
-	uint i;
 
 	//Return false if any conditions are not met
-	script_condition *scp;
+	//script_condition *scp;
```
asarium, line 315: Please remove commented out code

code/scripting/scripting.cpp
```diff
@@ -510,42 +540,29 @@ bool ConditionedHook::ConditionsValid(int action, object *objp, int more_data)
...
 					break;
 				}
 			case CHC_VERSION:
 				{
 					// Goober5000: I'm going to assume scripting doesn't care about SVN revision
-					char buf[32];
-					sprintf(buf, "%i.%i.%i", FS_VERSION_MAJOR, FS_VERSION_MINOR, FS_VERSION_BUILD);
-					if(stricmp(buf, scp->condition_string.c_str()) != 0)
+					
+					//if(stricmp(Script_engine_version.c_str(), scp.condition_string.c_str()) != 0)
```
asarium, line 551: This should also be removed.

code/scripting/scripting.cpp
```diff
@@ -510,42 +540,29 @@ bool ConditionedHook::ConditionsValid(int action, object *objp, int more_data)
...
-							return false;
-						}
+						return false;
 					}
 					break;
 				}
 			case CHC_APPLICATION:
 				{
-					if(Fred_running)
+					if(scp.condition_reference_index == 0)
```
asarium, line 563: Please also add a comment here what the value of `condition_reference_index` means in this case.

code/scripting/scripting.h
```diff
@@ -134,6 +134,7 @@ struct script_condition
 {
 	ConditionalType condition_type = CHC_NONE;
 	SCP_string condition_string;
+	int condition_reference_index;
```
asarium, line 137: Should be explicitly initialized to a sane value. `0`? `-1`? Not really sure what "sane" means here.

asarium (changes requested):
To be honest, this whole condition code is quite bad but this is a good measure to improve performance without a massive refactoring.

Some small points but then this should be good to go.

asarium:
There is quite a bit of potential for regressions with this so maybe we might want to wait with this until after the final release of 21.2.

EatThePath:
> 
> 
> There is quite a bit of potential for regressions with this so maybe we might want to wait with this until after the final release of 21.2.

Agreed. The performance issues this addresses are not new and for most cases they're not titanic either, so trying to cram it into a release at the last minute wouldn't make sense to me.

Also, I'll mention that I left this in draft because I was curious to have some eyes on the overall logic, but knew there were a number of sloppy things I needed to correct before a detailed review. The dead comments in particular. If I do that again I'll be more clear about the intention to save you good folks some time.

code/scripting/scripting.h
EatThePath: The idea was to use -1 as a 'not set' signifier since this is typically going to be an item in an array or enum. That was decided when I hadn't realized it'd be viable to just evaluate on add and was planning to have them evaluated the first time the condition was checked, though.

code/scripting/scripting.cpp
Goober5000: It's not the *setting* of the value that I'm worried about, but the value itself.  0 corresponds to `GS_STATE_INVALID`, and 1 to `GS_STATE_MAIN_MENU`.

EatThePath: I'm not sure I'm understanding your concern here so apologies if'm backing the train of thought too far back here.

I think the reference_index name isn't ideal now, and I need to think of a better name, but I named it when I assumed the highest impact was going to be storing indices for ship_info, weapon_info and so on instead of looking up those every call. STATE more or less follows that model, storing the enum value the state to check for. But not every condition type works the same way, and that approach isn't really applicable for APPLICATION and VERSION. For them and them alone it essentially stands in for a boolean, as those values aren't going to change after load so no repeated evaluation at all is really needed.

If none of that is relevant to what you're getting at, I'm afraid I don't understand what you're getting at.

code/scripting/scripting.h
asarium: It still needs to be initialized to _something_ otherwise we have unintialized data here and that is never a good thing.

EatThePath: Got it, I misread you initially. Despite the example right above that it slipped my mind that initializing in the struct declaration like that was an option, so I did it locally at what I believe is the only place script_conditions are created, but it'd be better to move it and I will.

code/scripting/scripting.cpp
Goober5000: Oh I see what you mean.  I thought that this index could be interpreted as a state in those cases, in which case it risked being assigned the wrong state.

Perhaps a clearer way to do this is to make a `union` of the integer `state_index`, a boolean `condition_value`, and whatever other indexes/values are needed.  Then when you initialize it you would have to `memset` the entire union to 0.

EatThePath:
In case anyone re-reviews before I finish fixing it: I realize now changed some stuff that shouldn't be changed in other conditions, in the misguided interest of cleaner code. Will try to squash things together to eliminate some of the noise in the commit history once it's fixed.

EatThePath:
to be clear post the last comment, the last two pushes took care of that issue and I believe I've sufficiently addressed all the requested changes.

asarium (approved):

Goober5000 (approved):
Looks ok to me.

PR #3362, opened 2021-04-06
Remove unecessary ship_get_subsys in model_rotate_gun

Baezon:
According to some profiling done by @EatThePath this `ship_get_subsys` in `model_rotate_gun` is a surprisingly large expense. Among its users `model_rotate_gun` is definitely the odd man out, all the others are dealing with external user-provided strings, while the callers of `model_rotate_gun` in all cases already have the `ship_subsys` handy anyway, so the function is modified to take it as its argument instead of the `model_subsystem`.


Goober5000 (approved):
Huh, good call.

PR #3363, opened 2021-04-08
Fix event parsing by setting flags to zero before loading.

### LINKED ISSUE #3355: Interval & Delay use Msecs flag gets added erroneously.
I don't see a rhyme or reason for this. It could be related to a certain build or it could just be totally random. In my case, I tracked down the flag being added around Jan 26, but I could not tell you what FSO build I was using at the time.

The missions got this flag added when I opened the missions in FRED to add a couple new events, iirc. The way the flag is added does not seem to follow a pattern. See this screenshot of an SVN diff. Two events that are nearly identical where one had the flag added and another did not.
image

The flag is not added to all events. Just some. This is, imo, a really big issue. I cannot trust FRED to not break past missions in the background if I happen to open those missions to change or fix anything else.

JohnAFernandez:
We need to set bit flags to zero before stuffing.

fixes #3355

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -5065,6 +5065,9 @@ void parse_event(mission * /*pm*/)
 		}
 	}
 
+	// Need to set this to zero so that we don't accidentally reuse old data.
+	event->flags = 0;
+
 	if (optional_string("+Event Flags:")) {
 		parse_string_flag_list(&event->flags, Mission_event_flags, Num_mission_event_flags);
 	}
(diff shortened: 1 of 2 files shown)


Goober5000 (changes requested):
There are a few places in the code where this would overwrite existing bits that need to be kept.  Instead, add a comment to this function explaining this potential pitfall, and set the event flags to 0 around missionparse.cpp line 5068, just before the if() block.

Goober5000:
Excellent detective work btw.

Goober5000 (approved):
Looks good.

I don't think `event->flags = 0;` needs to specifically be called out -- `event->team = -1;` is a few lines up and serves a similar function, and isn't called out -- but that's my opinion.

PR #3369, opened 2021-04-10
Properly cull generic debris

### LINKED ISSUE #3368: absurd amounts of debris objects
I haven't dissected or bisected this and I don't think I'm likely to soon. Sorry for the lack of detail, but it does seem like a potentially serious issue and I'd rather file it than forget about it.

I was running the Warmachine version of BP:AoA's Universal Truth when I hit #3361. @JohnAFernandez suggested I bump up object limits and see how high my peak object usage gets, so I did. After tripling the object limit I hit that assert again. According to my stat tracking, which I'm rather fairly confident in, at that time there were 9862 debris objects being generated. It surged into the thousands several times during the missions, passing the standard engine object limits more than once. 

I don't know exactly what events line up with those surges. Beam volleys maybe? Whatever it is, there seems like there has to be something behaving badly here.

Baezon:
Generic debris used to be culled only based on a factor of the source ship's size, but I was only considering explosion debris which would be presumably sized appropriately to the ship in question. Regular impact debris is still always small so failing to cull it on a destroyer adds up extremely quickly.

This checks first for the builtin debris which is always small with the original 200m check, and then later does an actual debris size proportional cull once its radius is known. Doing two culls based on distance is kind awkward but i figured doing the first quick cull was better since that was the vastly more common scenario.

Closes #3368

### DIFF
--- a/code/debris/debris.cpp
+++ b/code/debris/debris.cpp
@@ -424,18 +424,6 @@ object *debris_create(object *source_obj, int model_num, int submodel_num, vec3d
 	sip = &Ship_info[shipp->ship_info_index];
 	vaporize = (shipp->flags[Ship::Ship_Flags::Vaporize]);
 
-	// cull the builtin generic debris if its too far away
-	if ( model_num == -1 )	{
-		// Make vaporize debris seen from farther away
-		float dist = vm_vec_dist_quick( pos, &Eye_position );
-		if (vaporize) {
-			dist /= 2.0f;
-		}
-		if ( dist > 200.0f ) {
-			return NULL;
-		}
-	}
-
 	// try to maintain our soft limit
 	if (hull_flag && (Num_hull_pieces >= SOFT_LIMIT_DEBRIS_PIECES)) {
 		// cause oldest hull debris chunk to blow up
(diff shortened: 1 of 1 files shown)


code/debris/debris.cpp
```diff
@@ -480,6 +481,15 @@ object *debris_create(object *source_obj, int model_num, int submodel_num, vec3d
 
 	float radius = submodel_get_radius(db->model_num, db->submodel_num);
 
+	// if its generic, maybe cull it if its too small and far
+	if (!hull_flag) {
+		// Make vaporize debris seen from farther away
+		float dist = vm_vec_dist_quick(pos, &Eye_position);
+		if (dist > radius * 200.0f) {
+			return NULL;
```
JohnAFernandez, line 489: nullptr at least for new lines.  If you want to update the rest of this function with nullptr to match, I don't think anyone else will complain.

Everything else looks good.

JohnAFernandez (changes requested):

JohnAFernandez (approved):

JohnAFernandez:
Fixes #3368

code/debris/debris.cpp
```diff
@@ -480,6 +481,15 @@ object *debris_create(object *source_obj, int model_num, int submodel_num, vec3d
 
 	float radius = submodel_get_radius(db->model_num, db->submodel_num);
 
+	// if its generic, maybe cull it if its too small and far
+	if (!hull_flag) {
+		// Make vaporize debris seen from farther away
```
MageKing17, line 486: This comment makes no sense without a `Ship::Ship_Flags::Vaporize` check.

MageKing17 (changes requested):
One minor comment issue (seems like a copy&paste mistake), but otherwise this seems fine.

code/debris/debris.cpp
Baezon: Oh whoops, yes, I think I meant to remove the comment as well but this would cull vaporize debris anyway if they passed the check above so it should just ask about vaporizing again here..

Baezon:
Blech, whatever I just made it the one check, the duplication was really bothering me. Ok so even the builtin debris will need to do some model lookups before it rejects due to distance, big whoop.

MageKing17 (approved):

PR #3371, opened 2021-04-11
Add guardrails to object cleanup during object creation

### LINKED ISSUE #3361: Assert creating weapon while out of objects
I hit an assert on weapons.cpp:5926, which apparently trips if the weapon can't get a valid object number, which in turn appears to happen only if there are no object slots available. 

Filing on the general principle that assertions should never actually be hit.

JohnAFernandez:
This does three things.  1) It changes an INT3 and an Assertion that can be triggered if there are not enough object slots to debug log statements. 2) It removes some of the problematic logic that was causing crashes when trying to clear weapons when short on slots.  Previously, if none of the "safe" weapons could be removed, then it would resort to the oldest weapon, which could be colliding!  (this crash started my 4 hour search two nights ago)  3) It allows object creation to fail on purpose for certain types of objects like fireballs and asteroids instead of forcing weapons to be removed for them, if there are only 10 object slots remaining.

These changes are no longer urgent due to @Baezon's changes to how debris is being culled.  But they are still important changes until if/when objects would be made dynamic to stabilize future builds if another bug like that is introduced.

Closes #3361 

PS: I first encountered the removing the colliding weapon bug during BP's Icarus, previous to Asteroth's changes, and hit it very early in the second scene the first time running it. This prevented the collision bug from reappearing after trying Icarus 6 more times.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -5690,13 +5690,9 @@ int weapon_create( vec3d * pos, matrix * porient, int weapon_type, int parent_ob
 	num_deleted = 0;
 	if (Num_weapons >= MAX_WEAPONS-5) {
 
-		//No, do remove for AI ships -- MK, 3/12/98  // don't need to try and delete weapons for ai ships
-		//if ( !(Objects[parent_objnum].flags[Object::Object_Flags::Player_ship]) )
-		//	return -1;
-
 		num_deleted = collide_remove_weapons();
 
-		nprintf(("WARNING", "Deleted %d weapons because of lack of slots\n", num_deleted));
+		mprintf(("Deleted %d weapons because of lack of slots.\n", num_deleted));
 		if (num_deleted == 0){
 			return -1;
 		}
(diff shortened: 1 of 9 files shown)


Goober5000:
Great idea.  This looks good to me, but I'd like @Baezon to take a look too.

code/weapon/weapons.cpp
```diff
@@ -5766,8 +5762,20 @@ int weapon_create( vec3d * pos, matrix * porient, int weapon_type, int parent_ob
...
-	objnum = obj_create( OBJ_WEAPON, parent_objnum, n, orient, pos, 2.0f, default_flags);
-	Assert(objnum >= 0);
+	// mark this object creation as essential, if it is created by a player.  
+	// You don't want players mysteriously wondering why they aren't firing.
+	if (parent_objp->flags[Object::Object_Flags::Player_ship]) {
+		objnum = obj_create( OBJ_WEAPON, parent_objnum, n, orient, pos, 2.0f, default_flags);
+	}
+	else {
+		objnum = obj_create( OBJ_WEAPON, parent_objnum, n, orient, pos, 2.0f, default_flags, false);
+	}
```
Baezon, line 5772: Just a style suggestion, bit more succinct
```suggestion
	// mark this object creation as essential, if it is created by a player.  
	// You don't want players mysteriously wondering why they aren't firing.
		objnum = obj_create( OBJ_WEAPON, parent_objnum, n, orient, pos, 2.0f, default_flags, parent_objp->flags[Object::Object_Flags::Player_ship]);
```

Baezon (commented):
I am loathe to actually remove the 'remove old weapons' code, because I don't think it's a terrible idea in general, just that the situation had a bad priority. But it's hard to guarantee this situation won't happen again in some other way if we still have it. Weapons can still create other weapons on collision...

Neither option is great, since we're just trading new objects to maintain these old ones, but I'm not going to demand it stay.

code/weapon/weapons.cpp
JohnAFernandez: That is elegant.

Baezon (approved):

PR #3372, opened 2021-04-11
Allow Fred to Edit and Allow Command briefs in FRED

JohnAFernandez:
They are currently allowed in the game, but not being allowed in Fred was an oversight of mine.

Tested and it works as expected.  A TVT mission will also load if it has a command brief defined, even though TVT will not show the command brief in-game.

As it's already allowed in-game, this should probably get into stable.

### DIFF
--- a/fred2/fredview.cpp
+++ b/fred2/fredview.cpp
@@ -4369,7 +4369,7 @@ void CFREDView::OnUpdateDisableUndo(CCmdUI* pCmdUI)
 
 void CFREDView::OnUpdateCmdBrief(CCmdUI* pCmdUI) 
 {
-	pCmdUI->Enable(!(The_mission.game_type & MISSION_TYPE_MULTI));
+	pCmdUI->Enable(true);
 }
 
 int get_visible_sub_system_count(ship *shipp)
--- a/fred2/missionsave.cpp
+++ b/fred2/missionsave.cpp
@@ -1371,9 +1371,6 @@ int CFred_mission_save::save_cmd_brief()
 	required_string_fred("#Command Briefing");
 	parse_comments(2);
 
-	if (The_mission.game_type & MISSION_TYPE_MULTI)
-		return err;  // no command briefings allowed in multiplayer missions.
-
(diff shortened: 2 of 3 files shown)


Goober5000 (changes requested):
Make the same change to QtFRED's missionsave.cpp.  Otherwise this looks good.

And I think the title of this PR might have a typo. ;)

JohnAFernandez:
> Make the same change to QtFRED's missionsave.cpp. Otherwise this looks good.
> 
> And I think the title of this PR might have a typo. ;)

I keep forgetting that qtFred has a different mission save file.

JohnAFernandez:
Should be good now.

notimaginative:
Does this need a safety check for when a retail compatible mission is being worked on? Cmd briefings still aren't going to work in that case, or with earlier versions of FSO for that matter...

Goober5000 (approved):

Goober5000:
This is just the FRED side.  I don't think it will hurt anything for the cbrief to be present in the file of a retail mission.  The retail engine won't show it in a TVT mission, of course, but I don't think it will barf when parsing it.

notimaginative:
I believe it will still load on the server side though. It's mainly the client side of it that we fixed. So it wouldn't be ignored in other words, it just creates a server/client sync issue where the client doesn't know what the server is doing.

JohnAFernandez:
I tried it on TVT and it ran fine by ignoring it.

JohnAFernandez:
I just have to ask, why would someone use FSO's FRED/qtFRED to add a command brief to a retail mission, and then play on retail's broken multiplayer instead of on FSO?

Unless it is very easy to do or I am misunderstanding, I don't see why we should be trying to support this situation.

notimaginative:
Not being a FREDer I'm not really sure either way. I just wanted to make sure it wasn't an issue to be concerned about.

Goober5000:
The main motivation for adding a retail-compatible save option to FRED is so that retail FS wouldn't choke on any FSO-specific formatting.  There is less attention toward FSO features that would cause *logic* errors rather than syntax errors.  For instance, FSO sexps will save in the retail format, but retail FS obviously wouldn't be able to run them.  I would put this cbrief idiosyncrasy in the same category.

notimaginative:
Gotcha. Sounds good then.

PR #3373, opened 2021-04-11
Make an invalid variable name throw an error.

MageKing17:
There were a few assertions in the SEXP code that would trip if an invalid variable name were specified in a SEXP, but nothing that actually stopped a Release build from performing an out-of-bounds array access if someone made a typo in notepad. Since, as far as I could tell, everything should go through `get_sexp_text_for_variable()` first, I changed its `Assert()` into an `Error()` that should hopefully ensure that we avoid ever doing anything with `Sexp_variables[-1]`.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -3391,7 +3391,9 @@ void get_sexp_text_for_variable(char *text, const char *token)
 	if ( !Fred_running ) {
 		// freespace - get index into Sexp_variables array
 		sexp_var_index = get_index_sexp_variable_name(text);
-		Assert(sexp_var_index != -1);
+		if (sexp_var_index == -1) {
+			Error(LOCATION, "Invalid variable name [%s]!", text);
+		}
 		sprintf(text, "%d", sexp_var_index);
 	}
 }


code/parse/sexp.cpp
```diff
@@ -3391,7 +3391,9 @@ void get_sexp_text_for_variable(char *text, const char *token)
 	if ( !Fred_running ) {
 		// freespace - get index into Sexp_variables array
 		sexp_var_index = get_index_sexp_variable_name(text);
-		Assert(sexp_var_index != -1);
+		if (sexp_var_index == -1) {
+			Error(LOCATION, "Invalid variable name [%s]!", text)
```
JohnAFernandez, line 3395: Imho, we should to tell them that there is something wrong with the mission file that can be fixed in FRED, or less competent users are going to come to us and think that we need to solve their problem.

MageKing17: The problem may not be with the mission file; a script can also cause this error, via e.g. `mn.runSEXP()`. I suppose it could also technically occur in a campaign file, even though they can't define variables in the first place, simply because the SEXP parser doesn't care if any variables exist or not, or whether we're in a campaign file or not, when attempting to parse a variable reference.

MageKing17:
Forgot the semicolon, ~~like a chump~~.

Goober5000 (approved):

PR #3379, opened 2021-04-12
address issues with error handling in chat_api

notimaginative:
- detect server connection error messages before blindly
   moving to user login (crash fix)
 - detect read errors and fail accordingly (crash fix)
 - allow server disconnect to run even if we aren't connected (cleanup & reset)

### DIFF
--- a/code/network/chat_api.cpp
+++ b/code/network/chat_api.cpp
@@ -10,6 +10,7 @@
 #include "globalincs/pstypes.h"
 #include "network/chat_api.h"
 #include "network/psnet2.h"
+#include "network/multi_log.h"
 
 #ifdef SCP_UNIX
 #include <sys/time.h>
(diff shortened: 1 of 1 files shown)


notimaginative:
@JohnAFernandez This should be all of the needed fixes but I'd recommend giving it a little extra testing to be sure since I've only tested it locally. I'm resetting the chat connection limit to 2 so it will be easier for you to reproduce the crash.

JohnAFernandez:
Looks like I'm still getting the crash.  Same garbage string is appearing in the chat buffer.   `ÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌ`

JohnAFernandez (approved):

PR #3380, opened 2021-04-13
add new popup type to handle mission/table validation

notimaginative:
A new type of conditional popup which can be created, run multiple test conditions, then close. This provides an alternative to `popup_till_condition()` which simply creates a popup, runs a single test condition, then closes.

Fixes the dreaded popup sounds during mission and table validation in multi.

### DIFF
--- a/code/network/multiutil.cpp
+++ b/code/network/multiutil.cpp
@@ -3074,6 +3074,10 @@ void multi_update_valid_missions()
 	// now poll for all unknown missions
 	was_cancelled = false;
 
+	if ( !Is_standalone ) {
+		popup_conditional_create(0, XSTR("&Cancel", 667), XSTR("Validating missions ...", -1));
+	}
+
 	for (idx = 0; idx < Multi_create_mission_list.size(); idx++) {
 		if (Multi_create_mission_list[idx].valid_status != MVALID_STATUS_UNKNOWN) {
 			continue;
@@ -3087,6 +3091,10 @@ void multi_update_valid_missions()
 		}
 
 		Multi_create_mission_list[idx].valid_status = (char)rval;
+	}
+
+	if ( !Is_standalone) {
+		popup_conditional_close();
 	}
 
 	// if the operation was cancelled, don't write anything new
--- a/code/network/multiutil.cpp
+++ b/code/network/multiutil.cpp
@@ -3074,7 +3074,7 @@ void multi_update_valid_missions()
 	// now poll for all unknown missions
 	was_cancelled = false;
 
-	if ( !Is_standalone ) {
+	if ( !(Game_mode & GM_STANDALONE_SERVER) ) {
 		popup_conditional_create(0, XSTR("&Cancel", 667), XSTR("Validating missions ...", -1));
 	}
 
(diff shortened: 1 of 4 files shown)


notimaginative:
This really needs to be tested better, particularly on standalone and mission stats saving. Submitting it now but it should probably *not* be part of the current release.

code/popup/popup.cpp
```diff
@@ -1173,3 +1174,139 @@ void popup_game_feature_not_in_demo()
...
+
+// the same as popup_till_condition(), but split into separate parts such that
+// it can run multiple conditions with a single popup
+bool popup_conditional_create(int flags, ...)
+{
+	char *format, *s;
+	va_list args;
+
+	if (Popup_is_active) {
+		Int3();
```
Goober5000, line 1186: We should use something other than Int3() in new code.

code/popup/popup.cpp
```diff
@@ -1173,3 +1174,139 @@ void popup_game_feature_not_in_demo()
...
+		popup_conditional_close();
+		return false;
+	}
+
+	return true;
+}
+
+int popup_conditional_do(int (*condition)(), const char *text)
+{
+	int choice = -1, done = 0;
```
Goober5000, line 1228: `done` ought to be a boolean

Goober5000 (changes requested):
Looks good aside from some minor things.

I would also suggest making a version with `bool (*condition)()` so that the API can be migrated to `bool` at some point.

code/popup/popup.cpp
notimaginative: Made a mental note of that, then forgot about it. 🤦
Fixed now.

code/popup/popup.cpp
notimaginative: `done` fixed.

notimaginative:
I'm not sure that a `bool (*condition)()` version would work out well. You generally need at least 3 states returned from the test fuction: success, failure, and waiting. With a boolean version you would only know that it was waiting or done, not the result of the test.

Goober5000:
> 
> 
> I'm not sure that a `bool (*condition)()` version would work out well. You generally need at least 3 states returned from the test fuction: success, failure, and waiting. With a boolean version you would only know that it was waiting or done, not the result of the test.

Good point.  Then we may want to thing about an enum for the future, but that's really beyond the scope of this particular PR.

Goober5000 (approved):

code/network/multiutil.cpp
```diff
@@ -3089,6 +3093,10 @@ void multi_update_valid_missions()
 		Multi_create_mission_list[idx].valid_status = (char)rval;
 	}
 
+	if ( !Is_standalone) {
```
JohnAFernandez, line 3096: At some point we should standardize which way to check if the instance is a standalone.  `Game_mode & GM_STANDALONE_SERVER`  I have seen in more places.

notimaginative: I did it that way at first and got a compiler error. Let me go back and check it again since I agree that we should standardize on the `Game_mode` check except in places that we can't/shouldn't include `freespace.h`.

PR #3384, opened 2021-04-14
Turn all asinf and acosf into safe versions

Baezon:
Goober added `vm_vec_delta_ang_norm_safe` to clamp the input of an `acosf` because floating point inaccuracy can sometimes produce numbers slightly greater than 1 or less than 1. After I ran into the exact same problem in `vm_vec_interp_constant` I think it's clear this is not specific to that particular usage, and instead all the `asinf`s and `acosf`s should be sanitized when used.

This also removes the occasional inconsistent bounds checking a few of their callers already did, so it can be made consistent for all usages.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -586,33 +586,11 @@ float vm_vec_delta_ang_norm(const vec3d *v0, const vec3d *v1, const vec3d *uvec)
 	float a;
 	vec3d t;
 
-	a = acosf(vm_vec_dot(v0,v1));
+	a = acosf_safe(vm_vec_dot(v0,v1));
 
 	if (uvec) {
 		vm_vec_cross(&t,v0,v1);
 		if ( vm_vec_dot(&t,uvec) < 0.0 )	{
-			a = -a;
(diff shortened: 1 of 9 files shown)


code/math/vecmat.cpp
```diff
@@ -977,21 +955,11 @@ angles *vm_extract_angles_matrix_alternate(angles *a, const matrix *m)
...
-		a->p = PI_2;	// pi/2
-	} else {
-		a->p = asinf(sp);
-	}
+	a->p = asinf_safe(-m->vec.fvec.xyz.y);
 
 	// Check for the Gimbal lock case, giving a slight tolerance
 	// for numerical imprecision
-	if (fabs(sp) > 0.9999f) {
+	if (fabs(a->p) > 0.9999f) {
```
Goober5000, line 962: In the original code, this was the argument *to* asinf, not the result of it.

Goober5000 (changes requested):
Great idea.  Noticed one thing though.

Goober5000 (approved):

PR #3385, opened 2021-04-15
Fix variable replacement in text in briefings and debriefings

jg18:
As reported by @Karajorma in https://www.hard-light.net/forums/index.php?topic=97412.msg1909252#msg1909252 text replacement for variables in briefings/debriefings doesn't take into account the effects of evaluating SEXPs used to determine whether to include the  briefing/debriefing stage.

This PR moves text replacement for variables so that it's applied on a per-stage basis and only if the stage is selected for inclusion (i.e., if its SEXP formula evaluates to true).

**Gotchas**:
1. Text replacement in multi assumes that variables and variable operations (e.g., modify) are replicated on all clients. The same will apply to SEXP containers, once those are added.
2. Debriefing SEXPs are unavoidably evaluated server-side multiple times: once for *every* team in `debrief_multi_server_stuff()`, and then, if the server is not a standalone, *again* in `debrief_set_stages_and_multi_stuff()`. Therefore, **debriefing SEXPs in multi missions *should not* have side effects.**
3. Text replacement in debriefings, however, *can* have side effects in multi missions, because replacement occurs only once in both server and clients.

### DIFF
--- a/code/missionui/missiondebrief.cpp
+++ b/code/missionui/missiondebrief.cpp
@@ -1890,8 +1890,6 @@ static void debrief_init_music()
 
 void debrief_init()
 {
-	int i;
-
 	Assert(!Debrief_inited);
 //	Campaign.loop_enabled = 0;
 	Campaign.loop_mission = CAMPAIGN_LOOP_MISSION_UNINITIALIZED;
--- a/code/missionui/missiondebrief.cpp
+++ b/code/missionui/missiondebrief.cpp
@@ -699,6 +699,9 @@ void debrief_set_multi_clients( int stage_count, int active_stages[] )
 	// set the pointers to the debriefings for this client
 	for (i = 0; i < stage_count; i++) {
 		Debrief_stages[Num_debrief_stages++] = &Debriefing->stages[active_stages[i]];
+		// in multi, debriefing text replacement must occur client-side
+		// update most recently added stage only, in case replacement has side effects
+		debrief_replace_stage_text(*Debrief_stages[Num_debrief_stages - 1]);
 	}
 
 	Debrief_multi_stages_loaded = 1;
--- a/code/missionui/missiondebrief.cpp
+++ b/code/missionui/missiondebrief.cpp
@@ -460,6 +460,7 @@ int Debrief_award_text_num_lines = 0;
 // prototypes, you know you love 'em
 void debrief_add_award_text(const char *str);
 void debrief_award_text_clear();
+void debrief_replace_stage_text(debrief_stage &stage);
 
 
 
@@ -720,6 +721,8 @@ void debrief_multi_server_stuff()
 		stages[i] = stage_active[i];
 		for (j=0; j<debriefp->num_stages; j++) {
 			if ( eval_sexp(debriefp->stages[j].formula) ) {
+				// DISCUSSME: how to handle repeatedly evaluating SEXPs that have side effects?
+				// DISCUSSME: at what point should we replace text?
 				stage_active[i][num_stages] = j;
 				num_stages++;
 			}
@@ -780,7 +783,9 @@ int debrief_set_stages_and_multi_stuff()
 	}
 
 	for (i=0; i<debriefp->num_stages; i++) {
+		// DISCUSSME: is SEXP_TRUE the only acceptable value? What about SEXP_KNOWN_TRUE?
 		if (eval_sexp(debriefp->stages[i].formula) == 1) {
+			debrief_replace_stage_text(debriefp->stages[i]);
 			Debrief_stages[Num_debrief_stages++] = &debriefp->stages[i];
 		}
 	}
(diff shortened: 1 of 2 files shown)


code/missionui/missiondebrief.cpp
```diff
@@ -720,6 +721,8 @@ void debrief_multi_server_stuff()
 		stages[i] = stage_active[i];
 		for (j=0; j<debriefp->num_stages; j++) {
 			if ( eval_sexp(debriefp->stages[j].formula) ) {
+				// DISCUSSME: how to handle repeatedly evaluating SEXPs that have side effects?
```
MageKing17, line 724: IMO, we shouldn't do anything special; it's up to the mission designer to make sure their SEXP won't break by being evaluated for each team.

code/missionui/missiondebrief.cpp
```diff
@@ -780,7 +783,9 @@ int debrief_set_stages_and_multi_stuff()
 	}
 
 	for (i=0; i<debriefp->num_stages; i++) {
+		// DISCUSSME: is SEXP_TRUE the only acceptable value? What about SEXP_KNOWN_TRUE?
```
MageKing17, line 786: `eval_sexp()` will never return `SEXP_KNOWN_TRUE`, so this condition covers both (although it should maybe use `SEXP_TRUE` instead of a magic number 1).

MageKing17 (commented):
A couple of observations.

code/missionui/missiondebrief.cpp
jg18: Ok, but since in TVT missions, a stage's SEXP is evaluated multiple times, at what point do we perform text replacement on the stage? The first time the stage's SEXP evaluates to true?

code/missionui/missiondebrief.cpp
jg18: ok. so `eval_sexp()` will always return either `SEXP_TRUE` or `SEXP_FALSE`?

I think the other calls to `eval_sexp()` that this PR deals with consider `true` to be any non-zero value (i.e., anything other than `SEXP_FALSE` ). But I guess I should leave the code logic untouched.

So yeah, I guess the only change I'll make is replace `1` with 1SEXP_TRUE`, as you suggested.

code/missionui/missiondebrief.cpp
Karajorma: For a TvT it seems sensible not to try to second guess what the FREDder might intend. A FREDder might want different teams to have different values displayed. For instance, a "You need $points-required to win" message should be different for team 1 and 2.

jg18: @Karajorma And that's exactly the issue I was concerned about. I don't think there's a way to do this.

The only debrief info the server sends to clients is which debriefing stages to display (by index); see `send_debrief_info()` in multimsgs.cpp. The stage indices for each team are selected in this function we're talking about.

In fact, AFAICT there is no way for the server to send updated debriefing text to clients, so I'm not sure text replacement can even work in multi.

Are mission variables and variable operations (e.g., modify) replicated on the clients? If so, then we *might* have a chance of making text replacement work client-side, but I'm not sure there's a way to perform the sort of context-sensitive text replacement with the side effects you expect, or at least, not without a major rewrite of the `DEBRIEF_INFO` packet definition. But that'd be way out of scope.

jg18: I pushed the best possible solution for multi I can think of. Please see the list of gotchas in the PR's OP.

jg18: On second thought, I wonder why  I haven't seen this issue with briefings in multi. Maybe I overlooked something in the code....

Goober5000: Won't the debriefing stage SEXPs be evaluated on the client?  For TvT missions, each client will belong to a particular team, so the stage SEXP evaluation should necessarily take place on the client, not the server.  And therefore, each client will have its own copy of the text, which *should* have client-specific text replacement.

This is all educated guesswork though.  It will be necessary to run some TvT checks to see if this is a correct summary of stage handling.

Karajorma: SEXPs are never evaluated on the client. Only ever on the server. Text replacement however will be done on the client side. 

Anything that is labelled as a network variable should automatically be sent to the clients. The problem then is that the server will evaluate the event for both teams and then either send out two packets or send one containing the last value. This means you're going to need two variables for the TvT example I gave above. Nothing that can be done about that.

jg18: Ah. so variables marked as "Network" are replicated on the clients.

And yeah, per your TvT example, surely you can't have a single variable `points` that has different values for team 1 and team 2. How would the server maintain an authoritative value for `points` in that case?

So, to sum up the **multi-only** gotchas:
1. Text replacement works only for network variables.
2. SEXPs in debriefings should never have side effects.
3. Text replacement in debriefings, however, can have side effects.

Are these ok? Any other concerns about the PR?

Note that, as a consequence, unless we support "Network Containers", containers will not be usable for text replacement in multi.

Goober5000: No other concerns from me.  Just make sure the "gotchas" are documented somewhere that modders will see them.

Goober5000 (approved):
Looks good.

PR #3386, opened 2021-04-16
add firewall status

notimaginative:
Gets status of the firewall connectivity test from the PXO server. Logs the status to `multi.log` and also shows it to the user via the in-game chatbox. This works for normal game creation and for standalone servers to verify that games are visible to the internet.


notimaginative:
Just noticed a bug with debug builds which *don't* have this feature hitting an assert. Not an issue with this PR specifically but an overzealous safety check in the existing code which is triggered by this feature being implemented on the PXO server side. It will likely be disabled server side until a fix can be implemented.

notimaginative:
A working fix for the assert has now been implemented. A new packet type (`GNT_GAMEUPDATE_STATUS`) was added which will set a flag in the tracker that the client requests a connectivity status report when a game server is added. This method requires the least amount of client code changes while preserving maximum compatibility with older builds and retail.

code/network/gtrack.cpp
```diff
@@ -446,6 +463,17 @@ void IdleGameTracker()
 				// send it to the PXO screen				
 				multi_pxo_channel_count_update(channel,num_servers);
 				break;
+
+			case GNT_GAME_PROBE_STATUS:
+				int flags;
+				int next_try;
+
+				memcpy(&flags, inpacket.data, sizeof(int));
+				memcpy(&next_try, inpacket.data+sizeof(int), sizeof(int));
```
JohnAFernandez, line 472: No need to make these variable sizes explicit?  You would know more than me, I just want to make sure.

code/network/multiui.cpp
```diff
@@ -3782,6 +3778,11 @@ void multi_create_game_do()
 
 		// don't bother setting netgame state if ont the server
 		if(Net_player->flags & NETINFO_FLAG_AM_MASTER){
+			// tell PXO about this game
+			if ( multi_fs_tracker_inited() ) {
+				multi_fs_tracker_login_freespace();
```
JohnAFernandez, line 3783: I'm guessing moving this is about having the code be more organized.

code/network/multi_fstracker.cpp
```diff
@@ -1360,6 +1366,56 @@ void multi_fs_tracker_report_stats_results()
...
+			int hours = next_try / (60*60);
+			SDL_snprintf(t_str, SDL_arraysize(t_str), "%d hours...>", hours);
+		}
+
+		str += t_str;
+	} else if (flags & PROBE_FLAGS::SUCCESS) {
+		str = "<PXO firewall probe was a success!>";
+	} else {
+		// getting here shouldn't happen, but it's not technically fatal
+		return;
```
JohnAFernandez, line 1410: Maybe a one-time mprintf() or multilog line with the value that caused it to reach here would be helpful to catch hard to  find bugs?

JohnAFernandez (commented):
Nothing that I can see is wrong.  Just looking for clarification, and one suggestion before I approve.  I'll try to build and test now.

code/network/multiui.cpp
JohnAFernandez: Oh wait, I just saw that it's to fix a bug.

code/network/gtrack.cpp
notimaginative: The PXO code is weird because of it being badly written in the first place requiring extra care needed in porting and compatibility handling. I avoided making some changes so that it would be easier to track against the unmodified retail code. But I also added various safety checks to make sure that various things stay compatible across platforms/compilers. Because of those two things I tend to do stuff like this where I typically wouldn't.

So it looks bad at a glance, but I don't just assume that the data size is always the same and have multiple safety checks to help make sure everything is okay.

code/network/multi_fstracker.cpp
notimaginative: There could technically be a case for that to happen which is why I handled it. But it's also self-correcting so there really isn't a need to be notified of it.

JohnAFernandez (approved):
Tested and bonafide!

PR #3387, opened 2021-04-17
Fix ai update danger weapon and Add guardrails for other checks

JohnAFernandez:
The AI code seems to assume that bugs can never happen and that it only has to check to see if an index is -1.  The trouble is, with memory bugs, this is not always the case, and we need to add some asserts to prevent error message-less edge case crashes if they start occurring. 

The first commit in this PR gracefully handles a case in multi where the objnum is nonsense.   The others just add those safety checks to throw an assert so that we don't do a negative array access by accident.

Bug reported here: https://discord.com/channels/223511295431933953/423504150437953537/832038232623022110

The minidumps used to find the break point are here: https://discord.com/channels/223511295431933953/423504150437953537/832062427226963982

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -2349,8 +2349,9 @@ void ai_set_goal_abort_support_call(object *objp, ai_info *aip)
 		object	*repair_obj;
 
 		if (aip->support_ship_objnum == -1) {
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Looks good aside from some typos.

PR #3394, opened 2021-04-20
Keep Fred from freezing

### LINKED ISSUE #3155: Fred hangs after being idle for long periods
The fix for mantis 3006 may have partially fixed this, but we just got a slew of new reports on discord on the Fred channel, including from Darius.

JohnAFernandez:
Could be because of variable overflows or from `thistime` somehow being much smaller than `lasttime`.

Hard to find the root cause, but manually setting the memory to bad values will no longer cause the freezing with this patch since we are explicitly checking for a negative value that is far too negative. 

Closes #3155


JohnAFernandez:
Fixes  #3155

Goober5000 (approved):

PR #3395, opened 2021-04-20
in perform-actions, evaluate the boolean last

Goober5000:
As noticed by @naomimyselfandi, it can be useful to defer evaluation of the boolean until after all the actions have been performed, in case the FREDder wants the value to depend on the actions.

This introduces a small incompatibility, but I don't know of any mod that actually depends on the boolean being evaluated first.  In the overwhelming majority of uses, the evaluation order is agnostic.


Goober5000:
I plan to replace this PR with a more comprehensive solution.

PR #3397, opened 2021-04-20
handle weapon "None" properly

### LINKED ISSUE #3393: Assert in FRED selecting 'none' for a secondary weapon
Specifically when the last table defined weapon prior to the secondary weapons has a cargo size of 0, tripping the `size > 0.0f` assert in `get_max_ammo_count_for_bank`. `WeaponEditorDlg::OnSelchangeMissile#()` uses `m_missile1 + First_secondary_index - 1` for the weapon_info argument for `get_max_ammo_count_for_bank`, so when 'none' is selected (m_missile1 = 0), the weapon_info immediately previous to the `First_secondary_index` is used.

Pretty minor, since this has no otherwise ill effects I can tell, and requires a fairly unlikely table setup.

Goober5000:
If the secondary weapon selection is None, the ammo count will be based on an incorrect weapon.  So guard against that.

Fixes #3393.

### DIFF
--- a/fred2/weaponeditordlg.cpp
+++ b/fred2/weaponeditordlg.cpp
@@ -585,10 +585,14 @@ void WeaponEditorDlg::OnSelchangeMissile1()
 {
 	UpdateData(TRUE);
 	UpdateData(TRUE);
-	if (m_cur_item == 0) {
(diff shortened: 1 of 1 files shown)


Baezon (approved):
LGTM, like I said in the issue this is fairly minor so this can probably be deferred until after the stable.

PR #3398, opened 2021-04-24
Karajorma's SEXP containers, revised

jg18:
Continuation of @Karajorma 's work on SEXP containers from PR #3253.

Remaining things to sort out (would appreciate help as part of the review):
1. SEXP syntax checking (what's required there?)
2. Pasting in FRED's events editor
3. The `Ctext_strings` hack (IMHO) in sexp.cpp. Maybe we can store a resolved name returned by `CTEXT()` as a `char[TOKEN_LENGTH]` field in the `sexp_cached_data` struct?


jg18 (commented):
To make the review more tractable, I'll start things off by asking for feedback on some things that I know should be addressed. This is just a start; I'll add more of these later.

code/parse/sexp.cpp
```diff
@@ -3786,6 +3876,164 @@ int get_sexp()
...
+		log_printf(LOGFILE_EVENT_LOG, "SEXP Container name %s is longer than limit %u",
+			name.c_str(),
+			sexp_container::NAME_MAX_LENGTH);
+		valid = false;
+	}
+
+	required_string("$Data Type:");
+	stuff_string(temp_type_string, F_NAME);
+	if (!strcmp(temp_type_string.c_str(), "Number")) {
+		type |= SEXP_CONTAINER_NUMBER_DATA;
```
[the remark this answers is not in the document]
Karajorma: That's what the error code above is supposed to check for.

code/parse/sexp.cpp
```diff
@@ -23298,6 +23567,706 @@ void sexp_set_motion_debris(int node)
...
+		node = CDR(node);
+	}
+
+	return SEXP_TRUE;
+}
+
+/**
+* Check if a SEXP map container has a key whose data matches a specific string.
+* If a third argument (string variable) is provided, store the data's associated key, if a match is found.
+* Returns -1 if the element is not in the container
```
jg18, line 23684: Oops, will need to remove this line of the comment.

code/parse/sexp.cpp
```diff
@@ -23298,6 +23567,706 @@ void sexp_set_motion_debris(int node)
...
+	const SCP_string possible_data = CTEXT(node);
+
+	for (const auto &kv_pair : Sexp_containers[index].map_data) {
+		if (possible_data == kv_pair.second) {
+			// check for optional variable to store the key
+			node = CDR(node);
+			if (node != -1) {
+				const int var_index = sexp_get_variable_index(node);
+				// DISCUSSME: should we do an explicit if-check instead?
+				Assert(var_index >= 0);
```
jg18, line 23713: As the comment above says, should we not assert and instead issue a warning if the variable doesn't exist?

code/parse/sexp.cpp
```diff
@@ -23298,6 +23567,706 @@ void sexp_set_motion_debris(int node)
...
+				// TODO: type checking of string key/var vs. number key/var?
```
jg18, line 23714: Should the check that the variable is the same type  as the map's keys (number or string) be done in `check_sexp_syntax()`? Can we assume that, if we get to this point, the types have already been checked?

code/parse/sexp.cpp
```diff
@@ -23298,6 +23567,706 @@ void sexp_set_motion_debris(int node)
...
+		Sexp_containers[list_index].list_data.emplace_back(kv_pair.first);
+	}
+}
+
+int get_sexp_container_index_generic(const char *name,
+	const std::function<bool(const char*, const char*)>& match_func)
+{
+	const char initial_lower = SCP_tolower(name[0]);
+
+	if (Container_indices_by_initial.count(initial_lower) == 0) {
```
jg18, line 24000: The usefulness of `Container_indices_by_initial` is debatable, although it can make verifying that a container name is *invalid* fast. If the name to check starts with a letter that no actual container's name starts with, then we can return -1 right away (i.e., in the line right below this one).

code/parse/sexp.cpp
```diff
@@ -23298,6 +23567,706 @@ void sexp_set_motion_debris(int node)
...
+		}
+
+		num_chars_to_replace += strlen(Container_modifiers[modifier_index].name);
+
+		// for the "At" modifier, we need to include the chars used by the index
+		if (modifier_index == SNF_CONTAINER_AT_INDEX) {
+			num_chars_to_replace += number_length;
+		}
+
+		// TODO: verify that the char we're about to skip below is a '&'
```
jg18, line 24169: How important is it to check that the char we skip is the final `&`?

And what should we do if it isn't an `&` (i.e., the replacement is malformed)?

code/parse/sexp.cpp
```diff
@@ -29803,6 +30945,222 @@ int check_text_for_variable_name(const char *text)
...
+			default:
+				Error(LOCATION, "Unknown modifier type found in container"); 
+				break;
+		}
+	}
+
+	return false;
+}
+
+int container_push_return_string(const SCP_string &result)
```
jg18, line 31098: This part of the PR is probably the one I'm the most concerned about.

When containers are used with `CTEXT()`, we store the resulting strings in a global array with 🤞🏻 that the string won't be overwritten throughout the time that we need it.

Maybe the resulting string should be written into the `Sexp_node`'s cached data instead?

jg18 (commented):
Here's a second round of review comments. Still more to do.

code/parse/sexp.cpp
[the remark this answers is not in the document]
jg18: I'll need to take a closer look at that, but I thought that was validating only values that are used in SEXPs?

What about data that's loaded into the container during mission parsing and then retrieved via text replacement in, say, messages?

Or maybe type correctness really matters only when used in SEXPs, anyway.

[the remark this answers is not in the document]
Karajorma: It only matters for SEXPs. The text replacement code should output both types of container.

code/mission/missionparse.cpp
```diff
@@ -5785,6 +5785,84 @@ void parse_variables()
...
+		return;
+	}
+
+	// first update this mission's containers from campaign-persistent containers
+	for (const auto &current_pc : Campaign.persistent_containers) {
+		int index = get_sexp_container_index(current_pc.container_name.c_str());
+		if (index != -1) {
+			Assert(index >= 0 && index < (int)Sexp_containers.size());
+			auto &container = Sexp_containers[index];
+			// FIXME TODO: whatto do if theres a type mismatch between the two containers?
```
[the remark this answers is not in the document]
Karajorma: If there's a mismatch I think we definitely have to use the one in the mission file rather than the persistent one since eternal containers could come from a completely unrelated campaign.

The game should give a warning but nothing beyond that.

code/parse/sexp.cpp
Karajorma: Yeah, that one probably should be a warning rather than an assertion.

code/parse/sexp.cpp
Karajorma: I agree with the concern, but what could be overwriting it?

code/parse/sexp.cpp
```diff
@@ -4085,6 +4339,12 @@ int build_sexp_string(SCP_string &accumulator, int cur_node, int level, int mode
 			stuff_sexp_text_string(buf, node, mode);
 			accumulator += buf;
 
+		} else if (Sexp_nodes[node].type & SEXP_FLAG_CONTAINER) {
```
jg18, line 4342: I haven't dug into the `build_sexp_string()` and `build_extended_sexp_string()` functions and thus have no idea whether this fragment and the similar one below it are correct.

code/parse/sexp.cpp
```diff
@@ -23298,6 +23567,706 @@ void sexp_set_motion_debris(int node)
...
+
+	int data_index = 0;
+	for (const SCP_string &data : Sexp_containers[index].list_data) {
+		if (possible_data == data) {
+			return data_index;
+		}
+		++data_index;
+	}
+
+	return -1;
```
jg18, line 23641: Should we issue a Warning() if we reach this line? Doesn't it indicate that the FREDder messed up? Or not necessarily?

jg18: My reasoning is that if the FREDder just wants to know whether the list has some data, they would use the `list-has-data` SEXP. With `list-data-index`, the assumption is that the data is in the list somewhere and the FREDder just wants to know where in the list it is.

code/parse/sexp.cpp
```diff
@@ -23298,6 +23567,706 @@ void sexp_set_motion_debris(int node)
...
+**/
+bool sexp_replace_container_refs_with_values(SCP_string &text)
+{
+	bool replaced_anything = false;
+	size_t lookHere = 0;
+	int num_replacements = 0;
+
+	while (lookHere < text.length()) {
+		// look for the meta-character
+		size_t foundHere = text.find('&', lookHere);
```
jg18, line 24191: Since a FREDder might reasonably want to use "&" in player-facing text, maybe we should pick a different meta-character for containers. Defining it in exactly one spot, like a `static constexpr char` field on `sexp_container` struct would be nice, too.

How about "^" as the containers meta-character? I can't image FREDders would want to use it otherwise, and it's near the "$" key that FREDders are used to using for variable name replacement.

code/parse/sexp.cpp
```diff
@@ -23298,6 +23567,706 @@ void sexp_set_motion_debris(int node)
...
+/**
+* Replaces container references in a text string with their values
+**/
+bool sexp_replace_container_refs_with_values(SCP_string &text)
+{
+	bool replaced_anything = false;
+	size_t lookHere = 0;
+	int num_replacements = 0;
+
+	while (lookHere < text.length()) {
```
jg18, line 24189: This code was originally modeled off of `sexp_replace_variable_names_with_values()`, but since container replacement is more complicated (especially given multidimensionality) and also since `do`-`while` loops are hard to understand IMO, I switched to this version.

code/parse/sexp.cpp
```diff
@@ -29803,6 +30945,222 @@ int check_text_for_variable_name(const char *text)
...
+
+			case SNF_CONTAINER_AT_INDEX:
+				node = CDR(node);
+				if (node == -1) {
+					return false;
+				}
+				bool is_nan, is_nan_forever;
+				data_index = eval_num(node, is_nan, is_nan_forever); 
+				if (is_nan || is_nan_forever) {
+					return false;
```
jg18, line 31082: Should we issue a Warning() here that `Sexp_nodes[node].text` was supposed to represent a number but didn't? Or should that have been caught earlier, in SEXP syntax checking?

code/parse/sexp.cpp
```diff
@@ -29817,6 +31175,9 @@ const char *CTEXT(int n)
 
 	// Goober5000 - MWAHAHAHAHAHAHAHA!  Thank you, Volition programmers!  Without
 	// the CTEXT wrapper, when-argument would probably be infeasibly difficult to code.
+	//
+	// Karajorma - Due to the addition of the container code, coders shouldn't call CTEXT twice from the same function for the same sexp node.
```
jg18, line 31179: Why shouldn't `CTEXT()` be called twice on the same node? Is it because of text replacement that, in the case of lists, can modify the list?

jg18: IMO, ideally `CTEXT()` should be an idempotent operation (i.e., calling it multiple times should have the same effect as calling it once).

If we cache in a `sexp_node`'s `sexp_cached_data` field the result of calling `CTEXT()` on it, shouldn't that ensure that `CTEXT()` remains idempotent?

code/parse/sexp.h
```diff
@@ -925,6 +943,11 @@ const char *CTEXT(int n);
 // flags for sexpressions -- masked onto the end of the type field
 #define SEXP_FLAG_PERSISTENT				(1<<31)		// should this sexp node be persistant across missions
 #define SEXP_FLAG_VARIABLE					(1<<30)
+#define SEXP_FLAG_LIST_CONTAINER			(1<<29)
```
jg18, line 946: Not sure that separate flags for list container vs. map container are needed. What are these sorts of flags for, anyway?

code/parse/sexp.h
```diff
@@ -1264,6 +1384,15 @@ bool generate_special_explosion_block_variables();
 int num_block_variables();
 bool has_special_explosion_block_index(ship *shipp, int *index);
 
+// sexp_containers
+// DISCUSSME: rename to sexp_get_container_index()?
+int get_sexp_container_index(const char* name);
```
jg18, line 1389: I guess the name of this function should be `sexp_get_sexp_container_index()`? I think it was originally something like that, but I didn't understand why, so I got rid of it.

Ditto that it should presumably be `sexp_update_sexo_containers()` below.

code/pilotfile/pilotfile_convert.h
```diff
@@ -163,6 +163,8 @@ struct plr_data {
 
 	// variables
 	SCP_vector<sexp_variable> variables;
+
+	SCP_vector<sexp_container> containers;
```
jg18, line 167: Maybe this field isn't needed? Isn't this file just for converting from the old pilot file format?

jg18: Err, maybe I was confused by the file name `pilotfile_convert.h` that this code is for pilot file conversion?

jg18 (commented):
Third PR pre-review pass. Finally got through the SEXP code, phew! Will continue later.

I'll reply to the existing comments when I get a chance. Trying to power on through the pre-review first.

code/parse/sexp.cpp
Karajorma: I figured that they might wish to try either method. If they don't want the index but just want to know the data is still in there somewhere they'd use the list-has-data. If they actually want the index they'd use list-data-index and then test if the index is -1. That way they'd save having to search the container twice.

code/parse/sexp.cpp
Karajorma: I'm fine with those changes.

code/parse/sexp.cpp
Karajorma: Returning NAN is perfectly fine for SEXPs. NAN_FOREVER is fine in every-time SEXPs. Neither actually imply the FREDder has screwed up.

code/parse/sexp.cpp
Karajorma: I probably should have made that comment clearer. I actually did mean to have a discussion about it at some point.

CTEXT has never been idempotent. It's just that the containers stuff reveals that. 

For the most part this isn't really an issue. The comment relates to a bad habit a few coders had of writing code for sexps like this 

something  CTEXT(n) 
other operations
something CTEXT(n)

instead of simply buffering the results of the first call to CTEXT. For the most part this was fine but it was banking on variables not changing (which to be fair, doesn't happen with any of the SEXPs I'm aware of). 

The big problem comes if you use the remove functions for a container. The second call to CTEXT(n) will return a different value since the original value in the container has now been removed. The solution is to not be lazy and just buffer the results of the first call to CTEXT since you shouldn't be calling such a computationally expensive function a second time anyway. 

The comment should warn people to not call CTEXT twice for the same node within a function as the contents of that node may have changed. 



As for the sexp_cached_data, I don't remember that existing when I wrote the container code. Now that I've looked at it, I don't think we could use it for that. In fact, it is actually the source of a bug in the container code since the game checks for variables or arguments on the grounds they might change but NOT containers. 

The if (!(Sexp_nodes[node].type & SEXP_FLAG_VARIABLE) && !(Sexp_nodes[node].flags & SNF_SPECIAL_ARG_IN_NODE)) style checks should probably be replaced with a function that checks if the node is a variable, container or argument.

code/parse/sexp.h
Karajorma: I seem to remember that the code used them to actually tell the difference between the two types. I can't remember if I later changed that.

code/parse/sexp.h
Karajorma: I've always tended to use the convention that functions that start with sexp are actual SEXPs. That one seems to be mostly respected in the code, but there are a few exceptions.

code/pilotfile/pilotfile_convert.h
Karajorma: Could be. I probably just copied the variable vector and then never realised it wasn't needed since old files wouldn't have containers.

fred2/editcontainerdlg.cpp
```diff
@@ -0,0 +1,842 @@
...
+	buf[ret] = '\0';
+	return !strcmp(buf, test_str);
+}
+
+void CEditContainerDlg::OnListerSelectionChange()
+{
+	const int index = m_container_data_lister.GetCurSel();
+
+	if (index == LB_ERR) {
+		// TODO: clear key and data text  edit boxes?
```
jg18, line 325: Hmm, should get rid of this line. Deleting user data is rarely a good thing.

fred2/missionsave.cpp
```diff
@@ -1635,6 +1635,128 @@ int CFred_mission_save::save_common_object_data(object *objp, ship *shipp)
...
+	if (Sexp_containers.empty()) {
+		fso_comment_pop(true);
+		return 0;
+	}
+
+	required_string_fred("#Sexp_containers");
+	parse_comments(2);
+
+	bool list_found = false;
+	bool map_found = false;
```
jg18, line 1653: These local vars are required because lists appear in the mission file before maps. I guess I could mention that in a comment.

fred2/missionsave.cpp
```diff
@@ -1635,6 +1635,128 @@ int CFred_mission_save::save_common_object_data(object *objp, ship *shipp)
...
+		if (Sexp_containers[i].is_list()) {
+			list_found = true;
+		} else if (Sexp_containers[i].is_map()) {
+			map_found = true;
+		}
+	}
+
+	if (list_found) {
+		required_string_fred("$Lists");
+		parse_comments(2);
```
jg18, line 1666: What does this function do? I can't tell if it's being used correctly.

jg18 (commented):
Fourth PR pre-review pass. I think all that's left to pre-review is FRED's SEXP tree code changes.

As mentioned, I'll go through replies and continue the discussion after done with the pre-review.

jg18 (commented):
Started the pre-review of the FRED SEXP tree code changes but will need to finish later. This is among the parts of the original PR that I understand the least.

code/parse/sexp.h
```diff
@@ -925,6 +943,11 @@ const char *CTEXT(int n);
 // flags for sexpressions -- masked onto the end of the type field
 #define SEXP_FLAG_PERSISTENT				(1<<31)		// should this sexp node be persistant across missions
 #define SEXP_FLAG_VARIABLE					(1<<30)
+#define SEXP_FLAG_LIST_CONTAINER			(1<<29)
+#define SEXP_FLAG_MAP_CONTAINER				(1<<28)
+#define SEXP_FLAG_CONTAINER_MODIFIER		(1<<27)
```
Goober5000, line 948: Do note that a few lines down, bits 29 through 27 are already in use by `SEXP_VARIABLE_SAVE_ON_MISSION_PROGRESS` etc.  You'll need to use bits 26 through 24.

This is not documented too well, but the sexp node type, the sexp variable type, and these extra flags are all stuffed into the sexp type field.  The problem is, different flags are used in different fields.  Some of them are used together and some aren't.

fred2/sexp_tree.h
```diff
@@ -27,13 +27,15 @@
 #define SEXPT_UNKNOWN	0x0002
 
 #define SEXPT_VALID		0x1000
-#define SEXPT_TYPE_MASK	0x00ff
+#define SEXPT_TYPE_MASK	0x0fff
```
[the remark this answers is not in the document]
Goober5000: It's needed because `SEXPT_CONTAINER` and `SEXPT_MODIFIER` occupy an additional "column" in the hexadecimal number.

This is going to make things more confusing because the `sexp_node` struct has the type divided up between the first two bytes and the last two bytes, whereas `tree_node` now has the type divided up between the first one byte and the last three bytes.  This is such a fundamental change that I would recommend pulling it out into its own separate PR, in case of unexpected side effects.

Goober5000 (changes requested):
Two comments, and one strongly recommended separate PR.

fred2/sexp_tree.h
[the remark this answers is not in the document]
jg18: Ok, good to know. These definitions will go into what'll become the first containers PR of a series.

I'm confused, though -- aren't `0x00ff` and `0x0fff` just two bytes?

code/parse/sexp.h
jg18: Ah, so the `SEXP_VARIABLE_*` persistence flags are actually SEXP flags?

Also, I don't think we really need separate SEXP flags for list and map containers, especially given how few open slots are available for new SEXP flags.

If the code needs to know the type of a container associated with a `sexp_node`, it can use
```
int index = get_sexp_container_index(Sexp_nodes[node].text);
Assert(index != -1);
auto &container = Sexp_containers[index];
```
and then call `container.is_map()` or `container.is_list()`.

jg18: On second thought, since we already have `SEXP_ATOM_CONTAINER`, `SEXP_FLAG_CONTAINER` would just be redundant, so I'll get rid of that, too. We'll keep `SEXP_FLAG_CONTAINER_MODIFIER`, though.

fred2/sexp_tree.h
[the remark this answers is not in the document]
Goober5000: Er yeah, they're just two bytes.  So change "byte" to "nibble" in what I wrote above.

jg18:
All of the functionality in this PR is covered by the Parts 1-8 PRs, so there's no need to keep this one open.

The comments may be useful for reference.

The PR may also be useful as a gold standard, since the player-facing behavior from builds off master should eventually match builds off this PR.

PR #3400, opened 2021-04-24
Fix APNG Animation Framecounter

### LINKED ISSUE #3399: FSO repeats just one item from looped `+Misc anim group:` if animations are stored in animated png format.
This is a issue that has resurfaced during my attempts to apng-ify mjn's mainhalls. Here, a real world example:

From table\aquitaine-hall.tbm:
```
+Num Misc Animations:	14
+Misc anim:		2_aqu-m1a
+Misc anim:		2_aqu-m1b
+Misc anim:		2_aqu-m1c
+Misc anim:		2_aqu-m2a
+Misc anim:		2_aqu-m2b
+Misc anim:		2_aqu-m2c
+Misc anim:		2_aqu-m3a
+Misc anim:		2_aqu-m3b
+Misc anim:		2_aqu-m3c
+Misc anim:		2_aqu-m4
+Misc anim:		2_aqu-m5
+Misc anim:		2_aqu-m6a
+Misc anim:		2_aqu-m7a
+Misc anim:		2_aqu-m8
+Misc anim group:	1
+Misc anim group:	1
+Misc anim group:	1
+Misc anim group:	2
+Misc anim group:	2
+Misc anim group:	2
+Misc anim group:	3
+Misc anim group:	3
+Misc anim group:	3
+Misc anim group:	-1
+Misc anim group:	-1
+Misc anim group:	-1
+Misc anim group:	-1
+Misc anim group:	-1
+Misc anim delay:	-1 0 0
+Misc anim delay:	-1 0 0
+Misc anim delay:	-1 0 0
+Misc anim delay:	-1 9000 30000
+Misc anim delay:	-1 9000 30000
+Misc anim delay:	-1 9000 30000
+Misc anim delay:	-1 5000 9000
+Misc anim delay:	-1 5000 9000
+Misc anim delay:	-1 5000 9000
+Misc anim delay:	-1 0 0
+Misc anim delay:	-1 0 0
+Misc anim delay:	-1 0 0
+Misc anim delay:	-1 0 0
+Misc anim delay:	-1 0 0
```

There are three groups. Group 1 `/2_aqu-m1[abc]/` is the animation group for the lift and the man at the top of the lift. Group 2 `/2_aqu-m2[abc]/` is the animation group for the windows above the ready room doors and campaign room screen. Group 3 `/2_aqu-m3[abc]/` is the animation group for the large window on the left below the barracks and multiplayer doors and the fighter bay beyond it.

When the mainhall loads, one animation appears to be randomly selected from each group and is played after the associated delay. It then immediately repeats. No other animation from the group is played. There is no apparent delay between the animation reaching it's end and being repeated. Which animation gets played seems to change each time the mainhall is loaded.

This can actually be seen here: https://youtu.be/iKdcxRuE_X8?t=18 Watch the windows in the top right. `2_aqu-m2c` plays twice in a row with no delay and starts to play a third time before i open the campaign screen.

This is a test version of the mod containing just the Aquitane Mainhall and all animations converted to apng (including the full -hall.tbm from above: https://annapuma.scrapironcity.net/downloads/mjnmh-apng-1.0.0a.rar

BMagnu:
This fixes #3399.
The problem is, that due to the exception for playing direction in APNG's made for first frames of backwards-playing doors ([code/graphics/generic.cpp#L493](https://github.com/BMagnu/fs2open.github.com/blob/79f5eda0d62343b823544a6cc5b13dd023708717/code/graphics/generic.cpp#L493)), the ``ga->current_frame`` got desynchronized from the actual ``framecount ga->png.anim->current_frame``.
This also allowed it to sometimes sink to -1 (which caused the sound crash, as that calculated a volume of over 100% with that), as well as caused it to never reach the end in a way that the animation system would count it as finished to start a new item from the same group.
Due to the early exit in the actual render function (that doesn't render the frame when the current frame and the previous frame are the same), changing the ``ga->current_frame`` _before_ the ``generic_render_png_stream(ga)`` didn't work.
I know this solution is not the prettiest, so any better ideas are very welcome. The proper solution would probably be to completely remove the separate apng framecounter, but that's probably a lot of work.

### DIFF
--- a/code/graphics/generic.cpp
+++ b/code/graphics/generic.cpp
@@ -676,6 +676,12 @@ void generic_anim_render_variable_frame_delay(generic_anim* ga, float frametime,
 		// I'm not even sure that the existing ani/eff code would allow non-streaming generic anims
 		generic_render_png_stream(ga);
 
+		//generic_render_png_stream can, in some cases, like displaying the first image of the start
+		//of an ANPG that plays backwards, decide to advance the stream in a different direction than
+		//the current_frame counter in/decrementation of this method. Said change of the frame counter
+		//is still necessary for correct handling of not rendering the same frame twice, hence why
+		//the current_frame counter needs to be synchronized to the actual apng frame counter
+		//after the the apng stream rendering is complete
 		ga->current_frame = ga->png.anim->current_frame;
 
 		gr_set_bitmap(ga->bitmap_id, GR_ALPHABLEND_FILTER, GR_BITBLT_MODE_NORMAL, alpha);
--- a/code/graphics/generic.cpp
+++ b/code/graphics/generic.cpp
@@ -675,6 +675,9 @@ void generic_anim_render_variable_frame_delay(generic_anim* ga, float frametime,
 		// note: generic anims are not currently ever non-streaming in FSO
 		// I'm not even sure that the existing ani/eff code would allow non-streaming generic anims
 		generic_render_png_stream(ga);
+
+		ga->current_frame = ga->png.anim->current_frame;
+
 		gr_set_bitmap(ga->bitmap_id, GR_ALPHABLEND_FILTER, GR_BITBLT_MODE_NORMAL, alpha);
 	}
 }


JadedDragoon:
"Due to the early exit in the actual render function (that doesn't render the frame when the current frame and the previous frame are the same), changing the ga->current_frame before the generic_render_png_stream(ga) didn't work."

It's worth noting identical sequential frames shouldn't happen with apngs, as `apngasm` combines such frames into a single frame and increase that frame's duration instead. E.G. Ten identical frames lasting .25 seconds each ends up as one frame lasting 2.5 seconds. So every frame in an animated png should be significantly different from the one before or after it. Perhaps this could provide a way to simplify the fix for this issue? Just throwing that out there.

BMagnu:
What happens here is that if the number of the current and the previous frame is identical, it doesn't even try to render the frame.
And due to shenanigans with how the first frame of doors is implemented, that happens with the normal animation of doors if you just try to change the current_frame counter in the same way that the apng frame counter counts.

code/graphics/generic.cpp
```diff
@@ -675,6 +675,9 @@ void generic_anim_render_variable_frame_delay(generic_anim* ga, float frametime,
 		// note: generic anims are not currently ever non-streaming in FSO
 		// I'm not even sure that the existing ani/eff code would allow non-streaming generic anims
 		generic_render_png_stream(ga);
```
JohnAFernandez, line 677: Needs a comment to explain why.

JohnAFernandez (changes requested):

code/graphics/generic.cpp
BMagnu: fair point

JohnAFernandez (approved):

JohnAFernandez:
Yeah, I know you're thinking it's an ugly patch, but to me, what you're really doing is making sure that the variable tracking what frame is being used is tracking the correct frame.  And if the function which tracks the animation current frame is now updated correctly, and now has the comment to explain the workaround, it's actually a good solution to apng being added after-the-fact and not following the rest of the engine's expected behavior.

But maybe someone else with more anim experience can chime in.

MageKing17:
It seems to me that if `generic_render_png_stream()` is going to be changing `ga->png.anim->current_frame` out of sync with `ga->current_frame`, that perhaps this is indicative of a deeper underlying bug; after all, the way the system is written is supposed to ensure that it only ever advances by one frame at a time, and all the surrounding code is supposed to guarantee that state of affairs (after all, if `ga->current_frame` hasn't changed, `generic_render_png_stream()` doesn't even do anything!). So if `ga->current_frame` and `ga->png.anim->current_frame` aren't both incrementing by 1 at the same time... what exactly *is* the code doing?

BMagnu:
> It seems to me that if `generic_render_png_stream()` is going to be changing `ga->png.anim->current_frame` out of sync with `ga->current_frame`, that perhaps this is indicative of a deeper underlying bug; after all, the way the system is written is supposed to ensure that it only ever advances by one frame at a time, and all the surrounding code is supposed to guarantee that state of affairs (after all, if `ga->current_frame` hasn't changed, `generic_render_png_stream()` doesn't even do anything!). So if `ga->current_frame` and `ga->png.anim->current_frame` aren't both incrementing by 1 at the same time... what exactly _is_ the code doing?

Just to follow up on this:
The outer ``current_frame`` will always stupidly roll back one frame, as it will only ever go it's actual direction.
What I think then happens is that during the first frame of each backwards APNG being rendered, ``current_frame`` will not change due to the outer conditions for it and stay 0, but ``generic_render_png_stream()`` contains a separate path for backwards streams whose last frame is -1 (i.e. never played) to ensure that at least the first frame is played once before stopping the backwards animation at it's first frame. Since that just now played the inner APNG in the forward direction, the APNG frame counter actually advances here, and is 1 while ``current_frame`` is 0. The next frame, this means that ``ga->png.anim->current_frame > 0`` is indeed true, resulting in ``current_frame`` being decremented to -1. And while the rest of the code works, that now means that there are not only times where current_frame is negative, it also not equals to the actual frame counter in the APNG anymore. So the safe way is to just always synchronize them, since there may be more such code paths that throw them out of sync that I may have missed.

MageKing17:
Actually, the correct solution is to change the conditionals so that the fact that `ga->png.anim->current_frame` is 1 ahead of the actual frame is accounted for, but then you start getting into other bugs related to assumptions in both the mainhall code *and* the generic animation code that can cause the animation to advance past the last frame, triggering an exception. Indeed, this should be happening with this "fix": in addition to animations being unable to be played more than once (because this synchronization will cause them to immediately pause after trying to unpause), trying to play the same misc animation more than twice will cause the APNG code to advance past the end of the animation. This PR needs to be reverted and a more comprehensive solution deferred to 21.4.

PR #3406, opened 2021-04-25
Avoid negative lookup when parsing loadouts.

### LINKED ISSUE #3405: BUG: invalid weapons in weapon pool lead to crash on mission load.
Hit an exception when loading a mission with a bad entry in it's loadout list, despite the fact that there's already error handling for that case. It looks like after checking if the index is valid, it still always goes on to check if the weapon at that index is player allowed, which for an invalid index is bad naturally. (parselo.cpp:3000). The mission I hit this on has been like this for a while, so I think this is new, but I don't have the will to bisect it or build an isolated repro right now.

```
>	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!stuff_loadout_list::__l2::<lambda>(loadout_row * buf) Line 3000	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!stuff_token_list<loadout_row,bool <lambda>(loadout_row *)>(std::vector<loadout_row,std::allocator<loadout_row>> & list, stuff_loadout_list::__l2::bool <lambda>(loadout_row *) stuff_one_token, const char * type_as_string) Line 2759	C++
 	[Inline Frame] fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!stuff_loadout_list(std::vector<loadout_row,std::allocator<loadout_row>> & lookup_type, int) Line 2947	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!parse_player_info2(mission * pm) Line 927	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!parse_player_info(mission * pm) Line 835	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!parse_mission(mission * flags, int) Line 5840	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!parse_main(const char * mission_name, int) Line 6277	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!mission_load(const char * filename_ext) Line 132	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!game_start_mission() Line 1360	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!game_enter_state(int old_state, int new_state) Line 5540	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!gameseq_set_state(int new_state, int override) Line 284	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!game_process_event(int current_state, int event) Line 4612	C++
 	[Inline Frame] fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!gameseq_process_events() Line 376	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!game_main(int argc, char * * argv) Line 6565	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!SDL_main(int argc, char * * argv) Line 7654	C++
 	fs2_open_21_1_0_x64_AVX2-FASTDBG.exe!main_getcmdline(...) Line 177	C
```

MageKing17:
#3271 changed the logic of `stuff_token_list()` in such a way that the supplied index being used to access an array was no longer skipped if it was invalid; naturally, this means that this warning was now being inevitably followed up by an out-of-bounds array access. Adding a `!skip_this_entry` check should hopefully fix the problem while still allowing the loadout count parsing code below it to run, which was presumably the reason the `continue` statement was replaced with the `skip_this_entry` variable in the first place. Fixes #3405.

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -2992,16 +2992,18 @@ void stuff_loadout_list(SCP_vector<loadout_row> &list, int lookup_type)
 			Warning(LOCATION, "Weapon '%s' has more than 300 possible spawned weapons over its lifetime! This can cause issues for Multiplayer.", Weapon_info[buf->index].name);
 		}
 
-		// similarly, complain if this is a valid ship or weapon class that the player can't use
-		if ((lookup_type == MISSION_LOADOUT_SHIP_LIST) && (!(Ship_info[buf->index].flags[Ship::Info_Flags::Player_ship])) ) {
-			error_display(0, "Ship type \"%s\" found in loadout of mission file. This class is not marked as a player ship...skipping", str.c_str());
-			skip_this_entry = true;
-		}
-		else if ((lookup_type == MISSION_LOADOUT_WEAPON_LIST) && (!(Weapon_info[buf->index].wi_flags[Weapon::Info_Flags::Player_allowed])) ) {
-			nprintf(("Warning",  "Warning: Weapon type %s found in loadout of mission file. This class is not marked as a player allowed weapon...skipping\n", str.c_str()));
-			if ( !Is_standalone )
-				error_display(0, "Weapon type \"%s\" found in loadout of mission file. This class is not marked as a player allowed weapon...skipping", str.c_str());
-			skip_this_entry = true;
+		if (!skip_this_entry) {
+			// similarly, complain if this is a valid ship or weapon class that the player can't use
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Good catch.

PR #3409, opened 2021-04-30
add missing newlines to several mprintf statements

### LINKED ISSUE #3407: Few log statements placed in one line in fs2_open.log
`Control item defined without a valid tab. Disabling: Show ObjectivesControl item defined without a valid tab. Disabling: Show Nav MapDutifully converting retail sound 36, 'l_hit.wav' to a 3D sound...`

21.2.0.RC3
It took me few moments to realize that there are 3 messages in 1 line.
It should look like this:
```
Control item defined without a valid tab. Disabling: Show Objectives
Control item defined without a valid tab. Disabling: Show Nav Map
Dutifully converting retail sound 36, 'l_hit.wav' to a 3D sound...
```

Goober5000:
Fixes #3407, and I used a regex search to find what I hope are all other examples.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -481,7 +481,7 @@ void parse_ai_profiles_tbl(const char *filename)
 					if (path_radii >= Minimum_subsystem_path_pt_dist) {
 						profile->subsystem_path_radii = path_radii;
 					} else {
-						mprintf(("Warning: \"$override radius for subsystem path points:\" should be >= %i (read %i). Value will not be used. ", Minimum_subsystem_path_pt_dist, path_radii));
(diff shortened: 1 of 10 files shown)


z64555:
Does the doxy of `mprintf` mention that it needs a `\n` at the end of the line?  This may be a good opportunity to add it if not.

Otherwise, LGTM 👍

z64555 (approved):

Goober5000:
> Does the doxy of `mprintf` mention that it needs a `\n` at the end of the line? This may be a good opportunity to add it if not.

There isn't any:
```
#define mprintf(args) do { if (LoggingEnabled) { outwnd_printf2 args; } } while (false)
#define nprintf(args) do { if (LoggingEnabled) { outwnd_printf args; } } while (false)
```

PR #3418, opened 2021-05-07
Qtfred runtime fixes

the-maddin:
Fixes blurry background image / stars and related Qt warning on startup, as well as a crash on quit, caused by deallocating a glsync too late.

### DIFF
--- a/qtfred/src/main.cpp
+++ b/qtfred/src/main.cpp
@@ -183,7 +183,7 @@ int main(int argc, char* argv[]) {
 	game_busy_callback(game_busy_callback);
 
 	// Show all top level windows that are our window
-	FredView* fredview;
+	FredView* fredview(nullptr);
 	for (auto& window : qApp->topLevelWidgets()) {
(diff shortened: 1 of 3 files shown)


JohnAFernandez (changes requested):
Looks like you should initialize the Fredview pointer to nullptr to avoid uninitialized variable warnings.

JohnAFernandez (approved):

JohnAFernandez:
I'll give it a day in case anyone else wants to look at it.

qtfred/src/main.cpp
```diff
@@ -180,12 +182,14 @@ int main(int argc, char* argv[]) {
 	// Use this to keep the app responsive
 	game_busy_callback(game_busy_callback);
 
+	// Show all top level windows that are our window
```
ngld, line 185: You should fix this comment. It used to be true but this PR changes the following loop to only show the *first* top level window. It shouldn't cause any noticable changes, though, since we only ever have one `FredView`.

PR #3421, opened 2021-05-08
fix add-data for variable names

### LINKED ISSUE #3298: FRED: add data for variable paramaters behaving differently from replace data for same, causing error
If you have a SEXP with an *optional* variable parameter and you use the event editor right click 'add data' option to populate the variable name, it adds only the variable name as a string rather than a proper variable reference. This causes an warning on save, and presumably breaks in game too. If you then use 'replace variable' it works properly, of course. It seems like since replace knows what to do with variable parameters, add could and should too.

Goober5000:
This properly handles variable additions by calling the replace-variable code on the new variable node immediately after the variable is added.  It's a bolted-on hack but it's the simplest way to fix the issue.  A better fix would be to completely redesign the way the sexp tree handles variables (and everything, really) but that would be a major undertaking.

Fixes #3298.

### DIFF
--- a/fred2/sexp_tree.cpp
+++ b/fred2/sexp_tree.cpp
@@ -985,11 +985,19 @@ void sexp_tree::right_clicked(int mode)
 						menu.EnableMenuItem(Operators[ptr->op].value, MF_ENABLED);
 
 					} else {
+						UINT flags = MF_STRING | MF_ENABLED;
+
+						if (!((data_idx + 3) % 30)) {
+							flags |= MF_MENUBARBREAK;
+						}
+
 						// add data
-						if ( (data_idx + 3) % 30) {
-							add_data_menu->AppendMenu(MF_STRING | MF_ENABLED, ID_ADD_MENU + data_idx, ptr->text.c_str());
+						if (type == OPF_VARIABLE_NAME) {
+							char buf[128];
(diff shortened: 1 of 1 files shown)


wookieejedi:
Awesome, this will be really nice to have fixed.

fred2/sexp_tree.cpp
```diff
@@ -1785,8 +1794,29 @@ BOOL sexp_tree::OnCommand(WPARAM wParam, LPARAM lParam)
...
+			auto saved_item_index = item_index;
+
+			if (Sexp_variables[var_idx].type & SEXP_VARIABLE_NUMBER) {
+				type = SEXPT_NUMBER;
+			}
+			else if (Sexp_variables[var_idx].type & SEXP_VARIABLE_STRING) {
+				type = SEXPT_STRING;
+			}
+			else {
+				Int3();	// unknown type
```
JohnAFernandez, line 1812: Need to use the UNREACHABLE() macro here.

JohnAFernandez (changes requested):
Just the one thing.

fred2/sexp_tree.cpp
Goober5000: Fixed

JohnAFernandez (approved):

PR #3430, opened 2021-05-15
Add fancy lua error message additions

BMagnu:
Adds capability to add custom text to lua errors, allowing to inform a user / programmer what went wrong instead of just showing a trace (or possibly printing variable contents as well).
Requested by @EatThePath in discord.

### DIFF
--- a/code/osapi/dialogs.cpp
+++ b/code/osapi/dialogs.cpp
@@ -5,6 +5,7 @@
 #include "cmdline/cmdline.h"
 #include "graphics/2d.h"
 #include "scripting/ade.h"
+#include "scripting/scripting.h"
 
 #include <SDL_messagebox.h>
 #include <SDL_clipboard.h>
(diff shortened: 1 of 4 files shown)


JohnAFernandez (approved):
This looks like a pretty simple change, but could use another look by someone who is more familiar with scripting to ensure we didn't fall into any random lesser-known pitfalls.

asarium (changes requested):
Sorry but no. We already have too many magic global variables.

I also fail to see the usability of such a feature `LuaError`s are supposed to be used for script errors and usually should not be visible to end users.

Goober5000:
I concur with @asarium.  We already have support for logging debug statements in scripts, which should serve the same purpose.

EatThePath:
@asarium @Goober5000 my request that led to this was a fairly off the cuff one that, on reflection, can probably be managed with Lua's error handling faculties if they work fully in FSO, and I can't imagine they wouldn't. My use case is essentially to put log-flooding debug prints in released scripts without actually flooding the logs so that when something croaks users might have a more informative error than the raw lua errors. And yeah, ideally end users would never see such a thing, but I'm not so confident in my own scripting ability as to think they never will.

Using full on error handling everywhere looks like enough of a hassle that I doubt I'll do it, but as the option (probably) exists I suppose I can't complain too hard. Apologies to @BMagnu for accidentally sending them on a snipe hunt, I suppose

Goober5000:
It's all good.  Sometimes it takes a bit of prototyping to realize another approach is better.

PR #3431, opened 2021-05-16
Fix AI targeting itself during EMP

### LINKED ISSUE #3366: Ship told to attack itself. I was told to report it.
Got this while playing Shell Game in BtA's first demo. It happened during the sunspot portion of the mission, so there's EMP bursts every couple of seconds. The hostile ships (War wing) also have cluster missiles.

`Warning: War 1#b was told to attack itself! This may be an error in the mission file.  If that checks out, please report to a coder!
aicode.cpp
Line: 2383`

I double checked the mission file as instructed. The only ai orders given to War 1#b are to attack Mako#3.

JohnAFernandez:
To keep the AI code from going crazy, we should keep the AI from targeting itself during/after EMP.  The AI will attempt to do this if they are set to traitor.

Closes #3366

### DIFF
--- a/code/weapon/emp.cpp
+++ b/code/weapon/emp.cpp
@@ -363,7 +363,7 @@ void emp_process_ship(ship *shipp)
 		int ship_lookup = ship_get_random_team_ship(iff_get_attackee_mask(shipp->team));
 
 		// if we got a valid ship object to target
-		if((ship_lookup >= 0) && (Ships[ship_lookup].objnum >= 0) && !(Objects[Ships[ship_lookup].objnum].flags[Object::Object_Flags::Protected])){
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
This is good, but as a matter of code style I think it should be the second condition, after `(ship_lookup >= 0)`.  Intuitively, checking that the ship isn't itself would be the second thing to look for, right after checking for a valid index.

JohnAFernandez:
> 
> 
> This is good, but as a matter of code style I think it should be the second condition, after `(ship_lookup >= 0)`. Intuitively, checking that the ship isn't itself would be the second thing to look for, right after checking for a valid index.

I disagree.  Because the program will follow the same order of checks as it does in the code, and this is the rarest case of it being rejected, this is the most efficient way to add the check.

PR #3433, opened 2021-05-17
Add nullptr check for obj_create priority

Baezon:
Check for nullptr first, otherwise weapons without parents will crash on creation.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -5770,7 +5770,7 @@ int weapon_create( vec3d * pos, matrix * porient, int weapon_type, int parent_ob
 
 	// mark this object creation as essential, if it is created by a player.  
 	// You don't want players mysteriously wondering why they aren't firing.
-		objnum = obj_create( OBJ_WEAPON, parent_objnum, n, orient, pos, 2.0f, default_flags, parent_objp->flags[Object::Object_Flags::Player_ship]);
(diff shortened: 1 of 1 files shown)


code/weapon/weapons.cpp
```diff
@@ -5770,7 +5770,7 @@ int weapon_create( vec3d * pos, matrix * porient, int weapon_type, int parent_ob
 
 	// mark this object creation as essential, if it is created by a player.  
 	// You don't want players mysteriously wondering why they aren't firing.
-		objnum = obj_create( OBJ_WEAPON, parent_objnum, n, orient, pos, 2.0f, default_flags, parent_objp->flags[Object::Object_Flags::Player_ship]);
+	objnum = obj_create( OBJ_WEAPON, parent_objnum, n, orient, pos, 2.0f, default_flags, parent_objp != nullptr && parent_objp->flags[Object::Object_Flags::Player_ship]);
```
JohnAFernandez, line 5773: I know this is a higher priority fix, but can you add a parentheses for readability

JohnAFernandez (changes requested):

code/weapon/weapons.cpp
Baezon: something like `(parent_objp != nullptr) && parent_objp->flags[Object::Object_Flags::Player_ship]`?

JohnAFernandez: `(parent_objp != nullptr && parent_objp->flags[Object::Object_Flags::Player_ship])`

Helps the eye realize that the last argument is that expression.

JohnAFernandez (approved):

PR #3435, opened 2021-05-17
Expose Shadow Cascade Distances

wookieejedi:
The shadow cascade distances were tuned/added by swifty with the following logic:
`// these cascade distances are a result of some arbitrary tuning to give a good balance of quality and banding.`

Mods, especially TCs, may have very different sized ships compared to FS, and they may desire more precise control over shadow cascade distances.

This PR exposes those values with settings in the `game_settings.tbl`.

PR is tested and works as expected.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -401,10 +401,10 @@ void parse_mod_table(const char *filename)
 		if (optional_string("$Shadow Cascade Distances:")) {
 			float dis[4];
 			stuff_float_list(dis, 4);
-			if ((dis[0] >= 0) && (dis[1] >= 0) && (dis[2] >= 0) && (dis[3] >= 0)) {
-				Shadow_distances = std::make_tuple(static_cast<float>(dis[0]), static_cast<float>(dis[1]), static_cast<float>(dis[2]), static_cast<float>(dis[3]));
+			if ((dis[0] >= 0) && (dis[1] > dis[0]) && (dis[2] > dis[1]) && (dis[3] > dis[2])) {
+				Shadow_distances = std::make_tuple((dis[0]), (dis[1]), (dis[2]), (dis[3]));
 			} else {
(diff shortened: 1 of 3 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -397,6 +398,16 @@ void parse_mod_table(const char *filename)
 			stuff_boolean(&Neb_affects_fireballs);
 		}
 
+		if (optional_string("$Shadow Cascade Distances:")) {
+			float dis[4];
+			stuff_float_list(dis, 4);
+			if ((dis[0] >= 0) && (dis[1] >= 0) && (dis[2] >= 0) && (dis[3] >= 0)) {
+				Shadow_distances = std::make_tuple(static_cast<float>(dis[0]), static_cast<float>(dis[1]), static_cast<float>(dis[2]), static_cast<float>(dis[3]));
```
asarium, line 405: Is the cast to float necessary here? The original array is already a float so I would guess not.

code/mod_table/mod_table.cpp
```diff
@@ -397,6 +398,16 @@ void parse_mod_table(const char *filename)
 			stuff_boolean(&Neb_affects_fireballs);
 		}
 
+		if (optional_string("$Shadow Cascade Distances:")) {
+			float dis[4];
+			stuff_float_list(dis, 4);
+			if ((dis[0] >= 0) && (dis[1] >= 0) && (dis[2] >= 0) && (dis[3] >= 0)) {
```
asarium, line 404: The shadow algorithm probably cannot handle non-monotonically increasing values so this check should also check that the individual values are greater than the previous one:
```suggestion
			if ((dis[0] >= 0) && (dis[1] > dis[0]) && (dis[2] > dis[1]) && (dis[3] > dis[2])) {
```

asarium (changes requested):

code/mod_table/mod_table.cpp
wookieejedi: Very true, and thanks for the update suggestion!

asarium (approved):

PR #3436, opened 2021-05-19
Simplistic Ship Transparency

BMagnu:
This PR introduces the ability for models to be rendered with a transparency multiplier, settable through a SEXP, as requested by @Baezon.
One thing to note is that this PR does _not_ introduce polygon-level sorting or a proper OIT solution. This means, that setting the transparency like this, there are dangers of the model looking off (especially with high opacity or high self-occlusion). Fixing this would require a proper OIT solution, as polygon level depth sorting is not feasible for FSO. I've started trying out some ideas for this, but after discussion with a few people, it's become clear that this requires a lot more effort and thought, as presumably a new deferred rendering pass will be needed (which is also questionable performance-wise). Hence, a solution for generic models, even those not made for transparency, will need some effort by someone with more time than me at the moment.
This method however is just fine as long as the modder pays attention to which model this type of transparency is applied to.


Baezon (commented):
I can't comment much on the implementation itself, being a bit out of my wheelhouse, but it works pretty much as I expected and was prepared for. Even if this won't necessarily look good on arbitrary models at arbitrary levels, it still provides a useful tool that a modder can work with to get good results for at least some use cases.

PR #3437, opened 2021-05-19
Self-Shadow on 3D-Cockpits

### LINKED ISSUE #2023: Self-shadowing for cockpits
Swifty's differed lighting (self-shadows) rendering enhancements have not been implemented for cockpits.  Initially (circa 2014) he stated "[maybe later](https://www.hard-light.net/forums/index.php?topic=87121.msg1746370#msg1746370)" for that functionality.  To my knowledge no work has been done yet along these lines.  At this point, true shadows (instead of shading, which currently works for cockpits) is one of the few things remaining that holds the SCP back from competing visually with other more modern graphics engines.  As 3d models are added to more and more mods (including FS and FS2), this would add immensely to player immersion.

Thoughts on possible implementation from DahBlount:
It may actually be as simple as rendering a shadow map for the cockpit itself tbh
since there's only one cockpit model at a time, it would work out I'd think
you could probably get away with a 512x512 map for a single cockpit and not have to worry about it
it would essentially be:
shadow pass
normal scene pass
cockpit shadow pass
cockpit pass

BMagnu:
This finishes @wookieejedi's work on cockpit self-shadow. With this, when shadows are enabled, cockpits will now self-shade automatically. Note, that due to the implementation (and different shadow parameters), cockpits will not get shaded by any other objects (including the rest of their parent craft). Still looks quite good in most situations, and like this there is little performance cost (unlike if we had to re-render the whole scene's shadowmap).
Closes #2023

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -7371,7 +7371,7 @@ void ship_render_cockpit(object *objp)
 
 	vm_vec_unrotate(&pos, &sip->cockpit_offset, &eye_ori);
 
-	bool Shadow_override_backup = Shadow_override;
+	bool shadow_override_backup = Shadow_override;
 
 	//Deal with the model
 	model_clear_instance(sip->cockpit_model_num);
@@ -7414,7 +7414,7 @@ void ship_render_cockpit(object *objp)
 
 	//Restore the Shadow_override
 	if (Shadow_quality != ShadowQuality::Disabled)
-		Shadow_override = Shadow_override_backup;
+		Shadow_override = shadow_override_backup;
 }
 
 void ship_render_show_ship_cockpit(object *objp)
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -7371,13 +7371,15 @@ void ship_render_cockpit(object *objp)
 
 	vm_vec_unrotate(&pos, &sip->cockpit_offset, &eye_ori);
 
+	bool Shadow_override_backup = Shadow_override;
+
+	//Deal with the model
+	model_clear_instance(sip->cockpit_model_num);
 	if (Shadow_quality != ShadowQuality::Disabled) {
 		gr_reset_clip();
 		Shadow_override = false;
 
-		shadows_start_render(&Eye_matrix, &leaning_position, Proj_fov, gr_screen.clip_aspect, 0.25f, 0.75f, 1.5f, 3.0f);
-
-		model_clear_instance(sip->cockpit_model_num);
+		shadows_start_render(&Eye_matrix, &leaning_position, Proj_fov, gr_screen.clip_aspect, std::get<0>(Shadow_distances_cockpit), std::get<1>(Shadow_distances_cockpit), std::get<2>(Shadow_distances_cockpit), std::get<3>(Shadow_distances_cockpit));
 
 		model_render_params shadow_render_info;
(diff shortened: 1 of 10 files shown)


code/ship/ship.cpp
```diff
@@ -7393,6 +7412,10 @@ void ship_render_cockpit(object *objp)
 	gr_end_proj_matrix();
 
 	hud_save_restore_camera_data(0);
+
+	//Restore the Shadow_override
+	if (Shadow_quality != ShadowQuality::Disabled)
+		Shadow_override = true;
```
asarium, line 7418: This might have external effects. If you want to restore the previous state then you need to store that in a local variable before updating it in the block above.

code/ship/ship.cpp
```diff
@@ -7369,7 +7370,25 @@ void ship_render_cockpit(object *objp)
 	vec3d pos = vmd_zero_vector;
 
 	vm_vec_unrotate(&pos, &sip->cockpit_offset, &eye_ori);
-	gr_set_proj_matrix(Proj_fov, gr_screen.clip_aspect, 0.02f, 10.0f*pm->rad);
+
+	if (Shadow_quality != ShadowQuality::Disabled) {
+		gr_reset_clip();
+		Shadow_override = false;
+
+		shadows_start_render(&Eye_matrix, &leaning_position, Proj_fov, gr_screen.clip_aspect, 0.25f, 0.75f, 1.5f, 3.0f);
```
asarium, line 7378: Since we have the config option to specify these distances now it would be good to also have one for these cascade distances.

code/ship/ship.cpp
```diff
@@ -7369,7 +7370,25 @@ void ship_render_cockpit(object *objp)
...
+
+		model_clear_instance(sip->cockpit_model_num);
```
asarium, line 7380: This is only needed once so you can move the call from above before this if block.

BMagnu:
Converting to draft to quickly add configurable cockpit shadow distances, like in #3435

asarium (changes requested):
Some minor changes could be done here.

Regarding the implementation, this should be mostly fine for traditional GPUs if the driver can figure out that we have no dependencies to the previous render commands. I am not sure how exactly that works but there is `glInvalidateFramebuffer` which might be useful for indicating this.

However, my guess would be that this can cause problems with tiled renderers if that dependency (or lack thereof) is not expressed properly. Since we do not officially support platforms with that rendering scheme it should be fine.
Maybe something to consider in the future though.

wookieejedi:
Very immense thank you for figuring this one out!

wookieejedi:
Will also address #2023.

code/ship/ship.cpp
```diff
@@ -7369,11 +7370,28 @@ void ship_render_cockpit(object *objp)
 	vec3d pos = vmd_zero_vector;
 
 	vm_vec_unrotate(&pos, &sip->cockpit_offset, &eye_ori);
-	gr_set_proj_matrix(Proj_fov, gr_screen.clip_aspect, 0.02f, 10.0f*pm->rad);
-	gr_set_view_matrix(&leaning_position, &Eye_matrix);
+
+	bool Shadow_override_backup = Shadow_override;
```
asarium, line 7374: Variables should be lower case.

asarium (changes requested):
Mostly good to go now.

asarium (approved):

PR #3439, opened 2021-05-21
Fix erroneous line break generation for non-ASCII string and high resolutions

### LINKED ISSUE #3438: FR: Max briefing text size could be increased
I have been messing with the Japanese Translation on NTP, but the translation always CTDs during certain command briefing stages due to
Assert: "n_chars[i] < MAX_BRIEF_LINE_LEN"
missionbriefcommon.cpp
Line: 1701

Based on the variable name I am seeing, I am assuming that the translation text has tripped the maximum text length for briefings. If this limit could be extended, that would be appreciated.

BMagnu:
FSO's function to automatically linebreak text was previously only looking at the line lengths in pixels, and copying that into fixed length buffers. Hence, if you had characters using a lot of bytes per width (like in the NTP's Japanese translation with Kanjis) or a high resolution resulting in more characters per line being displayable (this is conjecture only, I haven't actually tested that this was an error before, but I am fairly certain it was), many functions using this linebreak functions could trip asserts, or worse, overrun their buffers causing CTD's. Thus, this also fixes #3438.
For the briefing parts specifically I also increased the buffer size, as this is a relatively safe thing to do, and otherwise lines could be broken earlier than the screens end, which is ugly (but not a CTD anymore).
I severely hope to not have missed any uses of this function, that would also need a specific maximum buffer length set.

### DIFF
--- a/code/menuui/optionsmenumulti.cpp
+++ b/code/menuui/optionsmenumulti.cpp
@@ -656,7 +656,7 @@ void options_multi_notify_process()
 
 	// otherwise display the string
 	line_height = gr_get_font_height() + 1;
-	line_count = split_str(Om_notify_string, 600, n_chars, p_str, 3);	
+	line_count = split_str(Om_notify_string, 600, n_chars, p_str, 3, 255);
 	y_start = OM_NOTIFY_Y;
 	gr_set_color_fast(&Color_bright);
 	for(idx=0;idx<line_count;idx++){
--- a/code/menuui/optionsmenumulti.cpp
+++ b/code/menuui/optionsmenumulti.cpp
@@ -58,6 +58,8 @@ int Om_mask_0		  = -1;
 int Om_background_1 = -1;
 int Om_mask_1       = -1;
 
+#define OM_NOTIFICATION_LINE_LEN 255
+
 // screen modes
 #define OM_MODE_NONE									-1		// no mode (unintialized)
 #define OM_MODE_GENERAL								0		// general tab
@@ -637,7 +639,7 @@ void options_multi_notify_process()
 	int w;
 	const char *p_str[3];
 	int n_chars[3];
-	char line[255];
+	char line[OM_NOTIFICATION_LINE_LEN];
 	int line_count;
 	int y_start;
 	int idx;
@@ -656,11 +658,11 @@ void options_multi_notify_process()
 
 	// otherwise display the string
 	line_height = gr_get_font_height() + 1;
-	line_count = split_str(Om_notify_string, 600, n_chars, p_str, 3, 255);
+	line_count = split_str(Om_notify_string, 600, n_chars, p_str, 3, OM_NOTIFICATION_LINE_LEN);
 	y_start = OM_NOTIFY_Y;
 	gr_set_color_fast(&Color_bright);
 	for(idx=0;idx<line_count;idx++){
-		memset(line, 0, 255);
+		memset(line, 0, OM_NOTIFICATION_LINE_LEN);
 		strncpy(line, p_str[idx], n_chars[idx]);
 				
 		gr_get_string_size(&w,NULL,line);
--- a/code/mission/missionbriefcommon.cpp
+++ b/code/mission/missionbriefcommon.cpp
@@ -1681,7 +1681,7 @@ int brief_color_text_init(const char* src, int w, const char default_color, int 
 	}
 
 	Assert(src != NULL);
-	n_lines = split_str(src, w, n_chars, p_str, BRIEF_META_CHAR);
+	n_lines = split_str(src, w, n_chars, p_str, BRIEF_META_CHAR, true, MAX_BRIEF_LINE_LEN);
 	Assert(n_lines >= 0);
 
 	//for compatability reasons truncate text from everything except the fiction viewer
--- a/code/mission/missionbriefcommon.cpp
+++ b/code/mission/missionbriefcommon.cpp
@@ -1681,7 +1681,7 @@ int brief_color_text_init(const char* src, int w, const char default_color, int 
 	}
 
 	Assert(src != NULL);
-	n_lines = split_str(src, w, n_chars, p_str, BRIEF_META_CHAR, true, MAX_BRIEF_LINE_LEN);
+	n_lines = split_str(src, w, n_chars, p_str, MAX_BRIEF_LINE_LEN, BRIEF_META_CHAR, true);
 	Assert(n_lines >= 0);
 
 	//for compatability reasons truncate text from everything except the fiction viewer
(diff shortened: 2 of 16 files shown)


code/network/multiui.cpp
```diff
@@ -234,13 +234,11 @@ void multi_common_split_text()
 	int	n_chars[MAX_BRIEF_LINES];
 	const char	*p_str[MAX_BRIEF_LINES];
 
-	n_lines = split_str(Multi_common_all_text, Multi_common_text_coords[gr_screen.res][2], n_chars, p_str, MULTI_COMMON_TEXT_MAX_LINES, MULTI_COMMON_TEXT_META_CHAR);
+	n_lines = split_str(Multi_common_all_text, Multi_common_text_coords[gr_screen.res][2], n_chars, p_str, MULTI_COMMON_TEXT_MAX_LINES, MULTI_COMMON_TEXT_MAX_LINE_LENGTH, MULTI_COMMON_TEXT_META_CHAR);
 	Assert(n_lines != -1);
 
 	for ( i = 0; i < n_lines; i++ ) {
-		//The E -- This check is unnecessary, and will break when fonts that aren't bank gothic are used
```
BMagnu, line 241: Just as a note as to why I removed this comment and readded the Assertion:
The assert here does _not_ check for string width in pixels or anything. 
It just checks that the string, in bytes, fits the buffer. Which is an essential Assertion, and that this is missing in other places where this string split was used is the reason that it didn't assert but CTD.
While the new function should always avoid this case, it's still a sensible assert, should that function ever be reworked

code/menuui/optionsmenumulti.cpp
```diff
@@ -656,7 +656,7 @@ void options_multi_notify_process()
 
 	// otherwise display the string
 	line_height = gr_get_font_height() + 1;
-	line_count = split_str(Om_notify_string, 600, n_chars, p_str, 3);	
+	line_count = split_str(Om_notify_string, 600, n_chars, p_str, 3, 255);
```
Goober5000, line 659: The 255 should be a `#define` or `const`, not a hardcoded number.

code/network/multi_pxo.cpp
```diff
@@ -3248,7 +3248,7 @@ void multi_pxo_chat_process_incoming(const char *txt,int mode)
 	}
 
 	// split the text up into as many lines as necessary
-	n_lines = split_str(msg_total, Multi_pxo_chat_coords[gr_screen.res][2] - 5, n_chars, p_str, 3);
+	n_lines = split_str(msg_total, Multi_pxo_chat_coords[gr_screen.res][2] - 5, n_chars, p_str, 3, 512);
```
Goober5000, line 3251: Same with the 512.

Goober5000 (changes requested):

code/mission/missionbriefcommon.cpp
```diff
@@ -1681,7 +1681,7 @@ int brief_color_text_init(const char* src, int w, const char default_color, int
 	}
 
 	Assert(src != NULL);
-	n_lines = split_str(src, w, n_chars, p_str, BRIEF_META_CHAR);
+	n_lines = split_str(src, w, n_chars, p_str, MAX_BRIEF_LINE_LEN, BRIEF_META_CHAR, true);
```
Goober5000, line 1684: I don't think leading whitespace should be stripped.  Some briefings may use whitespace-aligned text.

code/network/multiui.cpp
Goober5000: Good reasoning, but you should test this PR on non-BankGothic fonts to make sure you're not missing anything.

code/parse/parselo.cpp
```diff
@@ -3419,7 +3419,7 @@ int split_str(const char *src, int max_pixel_w, int *n_chars, const char **p_str
 		buffer[buf_index] = 0;  // null terminate it
 
 		gr_get_string_size(&sw, NULL, buffer);
-		if (sw >= max_pixel_w) {
+		if (sw >= max_pixel_w || buf_index >= max_line_length) {
```
Goober5000, line 3422: Should this check for length-1?  You have to account for not only the number of bytes but also the null terminator.

code/parse/parselo.cpp
```diff
@@ -3542,7 +3542,7 @@ int split_str(const char *src, int max_pixel_w, SCP_vector<int> &n_chars, SCP_ve
...
```
Goober5000, line 3545: Ditto

Goober5000 (commented):
A few more comments

code/mission/missionbriefcommon.cpp
BMagnu: This, as this true is the default argument, would be a change to the current behaviour.
I agree with your reasoning, but should we really change established behaviour for this?

code/parse/parselo.cpp
BMagnu: Hence I check >= and not >.
If, with the char already added, the length of the string is equal to the buffer size (and thus would fill the buffer exactly with no \0), it will be rolled back until the last separator (or by one).
Should work like this

code/network/multiui.cpp
BMagnu: Tbh, testing this very line is hard for me, as it's a part of multi and I have no clue how to actually get into a mission with a specific mod to test this assertion.
However, I would advocate for that being unnecessary anyways:
definition of the array we srncpy into:
``char Multi_common_text[MULTI_COMMON_TEXT_MAX_LINES][MULTI_COMMON_TEXT_MAX_LINE_LENGTH];``
definition of the assert
``Assert(n_chars[i] < MULTI_COMMON_TEXT_MAX_LINE_LENGTH); ``
strncpy in question:
``strncpy(Multi_common_text[i], p_str[i], n_chars[i]);``

Even if, somehow, that Assert had anything to do with linesize, if it would be triggered here (and be commented out), it would _always_ cause a CTD, unless that string to be copied contains magical \0's. And an assert (with broken fonts) is always better than a CTD (with broken fonts).

Goober5000: Ok

code/mission/missionbriefcommon.cpp
Goober5000: I didn't realize `true` was default.  In that case, don't change established behavior, but also it's probably better not to add the optional argument.

code/parse/parselo.cpp
Goober5000: Ok

Goober5000 (approved):
Looks good

code/mission/missionbriefcommon.cpp
BMagnu: ok, cleaned that up

Goober5000 (changes requested):
Just make the one change to remove the default `true`

Goober5000 (approved):

PR #3440, opened 2021-05-21
Fix floating point error on subsys hits during ship class changing

Baezon:
Although this change my seem trivial on the surface, this prevents 'damaging' operations being repeatedly applied to `max_hits` directly. Without this the `current_hits` of the subsystem can end up slightly higher than its max hits, and trip https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/ship/ship.cpp#L10448 this assert when changing ship class.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -2317,7 +2317,7 @@ int parse_create_object_sub(p_object *p_objp)
 		{
 			ptr->max_hits = ptr->system_info->max_subsys_strength * (shipp->ship_max_hull_strength / sip->max_hull_strength);
 
-			float new_hits = ptr->max_hits * (100.0f - sssp->percent) / 100.f;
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Approved, but are you sure this is the guilty line, and not some other line?

Baezon:
Yes, I specifically went through everything which wrote to `current_hits` because I knew that without taking damage there was no reason that `current_hits` should differ from `max_hits`. I forgot to mention, but this is specifically relevant in the 100% full health subsys situation, where those two really should be the same.

JohnAFernandez:
Maybe we should use a CLAMP instead and add a comment?  This would ensure desired behavior, even in other weird cases.

Baezon:
Assuming you mean here (as opposed to change_ship_type), it is possible to have a clamp which would certainly be more explicit about this, but I'm not sure its necessary to really call that much attention to this, as it is relatively minor. The assert simply drew my attention to it, as there's normally no reason why it should be above max hits.

PR #3441, opened 2021-05-21
Add support for ballistic beams

Baezon:
What it says on the tin. Surprisingly finicky to get the ammo to decrement properly. It shouldn't really matter too much but I semi-arbitrarily chose 4 per second as the ammo consumption rate for beams. Could be exposed but its kinda weird, ammo consumption is normally never specified because its always implicitly 1 per shot, but just like energy its nicer to for beams to have a continuous consumption.

Also moves energy handling into the common beam handling instead of just type c ones.


Goober5000:
I would either make the ammo configurable or set it to 1 per shot.  Someone might decide to make a "conventional" primary weapon that uses the beam code for its effects.  In that case there would need to be 1 ammo consumed per beam.

And actually, to amend that slightly, it should be 1 ammo per firing point, not per shot.

code/weapon/beam.cpp
```diff
@@ -1177,6 +1167,39 @@ void beam_move_all_post()
 			}
 		}
 
+		// deal with ammo/energy for fighter beams
+		bool multi_ai = MULTIPLAYER_CLIENT && (moveup->objp != Player_obj);
+		bool cheating_player = Weapon_energy_cheat && (moveup->objp == Player_obj);
+		if (moveup->flags & BF_IS_FIGHTER_BEAM && !multi_ai && !cheating_player) {
```
JohnAFernandez, line 1173: Just add a line here to help coders see the start of the if block, and I think we're good.

JohnAFernandez (changes requested):
I agree with Goober, but you didn't set out to create it like that.  Let's create a feature request with that behavior, once this is merged.

I only see one other thing.

Baezon:
Well, he's kinda right, but also adding ammo consumption is gonna be awkward, I've already starting doing it and it's like, should a function which returns if a bank is out of ammo return if the bank has more than 0 ammo or more than the ammo consumption of the weapon currently loaded into the bank? It certainly would increase the scope of the PR quite a bit.

JohnAFernandez:
Right, let's leave it for another PR, making sure there's a feature request for it so we don't forget.  Just make the other change, and we should merge this.

Baezon:
Ok, so I've brought it back to the regular 1 ammo per beam, just like regular weapons. Might want to leave this 'till tomorrow though, as we've already merged quite a few things today.

JohnAFernandez (approved):

Baezon:
By 'tomorrow' I of course meant a month.

PR #3443, opened 2021-05-21
Optimize trail rendering

Baezon:
Moves trails to the batching system instead of those `g3_render_immediate` functions. In my tests with a bazillion trails I found about ~75% reduction in total time spent on trail drawing stuff. Also cleans up a bunch of trail rendering code in general.

Removes the `TrailDraw` tracing scope because before this was used for the final rendering call, which was a a significant time sink itself, but was replaced with `batching_add_quad` which has a negligible cost (since its not actually rendering anything at that time), so all the time spent is all the related overhead with making the call which is captured by `RenderTrails`.


code/weapon/trails.cpp
```diff
@@ -190,129 +143,129 @@ void trail_render( trail * trailp )
...
-		w = trailp->val[n] * w_size + ti->w_start;
+		float w = trailp->val[n] * w_size + ti->w_start;
 
 		float fade = trailp->val[n];
 		
 		if (trailp->info.a_decay_exponent != 1.0f)
 			fade = powf(trailp->val[n], trailp->info.a_decay_exponent);
 
+		ubyte current_alpha = 0;
 		if (init_fade_out != 1.0f) {
```
BMagnu, line 177: This if is not too nice, being a float equals comparison. Yes, we just set it to the same literal a few lines above so it probably will actually trigger despite being a float equals check, but that still doesn't make it pretty.
As far as I can tell, it should be easily doable to move the condition of the if that sets ``init_fade_out`` to be the condition of this if, and then only ever even declare ``init_fade_out`` _within_ this if's block.
Alternatively, this if could just be removed. It's not doing anything semantically, and is just there to keep two multiplications from happening. Though I am not even convinced that a branch is really cheaper than two float multiplications anyway.

BMagnu (changes requested):
Seems mostly solid, apart from that one if.
Also, it needs this merge conflict resolved, which seems to be due to the ``&& Neb_affects_weapons`` condition added in 6ec1f45

BMagnu (approved):
LGTM.

PR #3444, opened 2021-05-22
Optimize motion debris rendering

Baezon:
"Asteroth, who the heck cares about motion debris rendering" 

Well, you might be surprised to learn, at least on my machine in testing, occasionally motion debris rendering could take as long as _a millisecond_. By switching it to the batching system, it significantly reduces the render time.

As always, a bunch of other improvements and cleanup related to drawing the motion debris. It now fades in from a distance, the max distance increased as a result, and the total number increased as a result of _that_. The changes to the size aren't "real" changes, for reasons that are unclear to me, the switch of rendering methods significantly increased their apparent size, so I adjusted the size constants to compensate.

### DIFF
--- a/code/starfield/starfield.cpp
+++ b/code/starfield/starfield.cpp
@@ -27,6 +27,7 @@
 #include "options/Option.h"
 #include "parse/parselo.h"
 #include "render/3d.h"
+#include "render/batching.h"
 #include "starfield/nebula.h"
 #include "starfield/starfield.h"
 #include "starfield/supernova.h"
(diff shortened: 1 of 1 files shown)


The-E (approved):

code/starfield/starfield.cpp
```diff
@@ -1791,23 +1770,18 @@ void stars_draw_debris()
...
-		vm_vec_sub( &d->last_pos, &d->pos, &Eye_position );
+			// scale alpha from 0 at max range to full at 60% range
+			alpha *= (vdist - MAX_DIST_RANGE) / -(MAX_DIST_RANGE * 0.6f);
 
-		vdist = vm_vec_mag_quick(&d->last_pos);
+			g3_transfer_vertex(&pnt, &mdebris.pos);
 
-		if ( (vdist < MIN_DIST_RANGE) || (vdist > MAX_DIST_RANGE) )
-			d->active = 0;
+			batching_add_bitmap(Debris_vclips[mdebris.vclip].bm + frame, &pnt, 0, mdebris.size, alpha);
```
asarium, line 1779: I do not see a call to actually execute the batches you add here in this function and if I remember correctly that is what actually makes the GPU do stuff in that system. While that does not cause issues with the rendering of the debris it might change your measurements since no graphics operations actually take place within this scope but the cost is added somewhere else.

This is still a worthwhile change but might not be the x30 performance increase you are seeing in your measurements.

Baezon: On second thought, you are right, that number I remembered was specifically this loop itself because I wanted the gauge the effect of increasing the total number of debris (its been a little bit since I wrote this code). But in general I am aware that the batch renders themselves should also be counted, which did come out favorably although I can't recall the exact numbers produced.

PR #3446, opened 2021-05-25
Add volume effects

### LINKED ISSUE #2935: Particle Effects FR: Ability to spawn particles within a spheroid region rather than from a single point
The single biggest limitation of the particle system right now is the fact that particles can only be spawned from one point. Therefore, I'd like an option to change the point to a region.

The region should have a radius (or diameter, whatever), as well as a length parameter – so that it can be compressed down into a disc, or extended into an ovoid. The length axis would be aligned with the direction of the source object. The particles, then, would spawn at random points within this region.

The ability to select a regular distribution or a normal distribution (with particles spawning more often near the region’s center) would be really nice, but is less important than having this functionality at all.

Baezon:
Adds the 'Volume' effect type, which spawns particles in a spherical volume. Options for biasing the particles towards the center or the edge, or 'stretching' it into an ellipsoid are provided as well.

Closes #2935

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -2863,3 +2863,14 @@ std::ostream& operator<<(std::ostream& os, const vec3d& vec)
 	os << "vec3d<" << vec.xyz.x << ", " << vec.xyz.y << ", " << vec.xyz.z << ">";
 	return os;
 }
+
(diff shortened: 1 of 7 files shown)


wookieejedi:
Ah this would be very, very useful. Eager to hear what @asarium thinks.

asarium (approved):
Seems sensible. I can't judge the math but I'm sure you worked that out already.

PR #3448, opened 2021-05-26
Fix a bug where shields could occasionally be pierced

Baezon:
...despite full shields in that quadrant. If the ship moves towards the weapon in the frame, the next collision check for the weapon may start it *inside* the shield where last frame it may have ended just *outside* the shield. By adding in the ship's velocity we essentially take into account the apparent motion of the weapon from the ship's own moving reference frame. This was tested and works with regular and highly adversarial examples, such as very slow moving weapons which are normally very difficult to actually trigger collisions against for this reason.

The bigger issue, however, is backwards compatibility. I think almost everyone has noticed this happen at some point or another, and although I haven't specifically tested for it, I highly suspect this is the reason Seraphim and Nephilim bombers can occasionally be destroyed in a single volley of trebuchets, despite the stats on the paper not allowing for this. This could be behind a flag (possibly enabled by `$Target version`) but I am loathe to do this because of how simple, clearly a bug, and fundamental this is to the collision system.

### DIFF
--- a/code/ai/ai_flags.h
+++ b/code/ai/ai_flags.h
@@ -137,6 +137,7 @@ namespace AI {
 		Improved_missile_avoidance,
 		Friendlies_use_countermeasure_firechance,
 		Improved_subsystem_attack_pathing,
+		Fixed_ship_weapon_collision,
 
 		NUM_VALUES
 	};
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -544,6 +544,8 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$improved subsystem attack pathing:", AI::Profile_Flags::Improved_subsystem_attack_pathing);
 
+				set_flag(profile, "$fixed ship-weapon collisions:", AI::Profile_Flags::Fixed_ship_weapon_collision);
+
 				// if we've been through once already and are at the same place, force a move
 				if (saved_Mp && (saved_Mp == Mp))
 				{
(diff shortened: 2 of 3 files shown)


JohnAFernandez:
It is a definite bug.  It's just a balance affecting bug...  So even though it was never intended, it may be better to put it behind a `$Target Version`, as loathsome as I agree it is.

Goober5000:
I concur.  We have put less obvious bugs behind flags.

I would add it to ai_profiles and then set the flag to be automatically activated for 21.4+, similar to several other ai_profiles flags.

Baezon:
I figured as much, although I'd prefer to avoid putting this in ai_profiles since it is not AI related, in lieu of something like game_settings similar to `$Fixed Turret Collisions`

Goober5000:
AI_Profiles probably should have had a different name from the beginning.  But the main difference between ai_profiles and game_settings is that ai_profiles can be changed on a per-mission basis, while game_settings applies mod-wide.  Therefore, a lot of flags affecting balance have been added to ai_profiles.  This allows missions created before the flag to use the old setting, while missions created after the flag can use the new setting.  (Scroll uses this a lot because of its long development history.)

Baezon:
Fair enough, I thought 'fixed turret collisions' was under target version, but apparently not. And ai_profiles already other not really AI-related stuff anyway.

`fixed ship-weapon collisions` isn't really a good name, but I wasn't sure what else to call it (this effect is not specific to shields).

JohnAFernandez (approved):
I agree that this is the best name that I can think of, too.

Goober5000 (approved):
LGTM too.

PR #3449, opened 2021-05-28
karajorma's SEXP containers, Part 1: Definition, parsing, and saving

jg18:
First in a series of PRs to add SEXP containers, based on PR #3398 .

This PR adds
- definition (`struct sexp_container`) and basic bit flags
- parsing containers from mission (`.fs2`) files
- saving containers to mission files

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -960,7 +960,6 @@ void multi_sexp_modify_variable();
 SCP_vector<SCP_string> *Current_event_log_buffer;
 SCP_vector<SCP_string> *Current_event_log_variable_buffer;
 SCP_vector<SCP_string> *Current_event_log_argument_buffer;
-SCP_vector<SCP_string> *Current_event_log_container_buffer;
 
 // Goober5000 - arg_item class stuff, borrowed from sexp_list_item class stuff -------------
 void arg_item::add_data(char *str, int n)
@@ -23492,8 +23491,7 @@ void add_to_event_log_buffer(int op_num, int result)
 {
 	Assertion ((Current_event_log_buffer != nullptr) &&
 				(Current_event_log_variable_buffer != nullptr)&& 
-				(Current_event_log_argument_buffer != nullptr) &&
-				(Current_event_log_container_buffer != nullptr), "Attempting to write to a non-existent log buffer");
+				(Current_event_log_argument_buffer != nullptr), "Attempting to write to a non-existent log buffer");
 
 	if (op_num == -1) {
 		nprintf(("SEXP", "ERROR: add_to_event_log_buffer() function called with op_num of %i; this should not happen. Contact a coder.\n", op_num));
@@ -23536,14 +23534,6 @@ void add_to_event_log_buffer(int op_num, int result)
 			tmp.append(Current_event_log_variable_buffer->back());
 			Current_event_log_variable_buffer->pop_back();
 			tmp.append("]");
-		}
-	}
-
-	if (!Current_event_log_container_buffer->empty()) {
-		tmp.append("\nContainers:\n");
-		while (!Current_event_log_container_buffer->empty()) {
-			tmp.append(Current_event_log_container_buffer->back().c_str()); 
-			Current_event_log_container_buffer->pop_back();
 		}
 	}
 
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -14,6 +14,10 @@
 //	It uses a very baggy format, allocating 16 characters per token, regardless
 //	of how many are used.
 
+#include <algorithm>
+#include <functional>
+#include <iterator>
+
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
@@ -847,6 +851,12 @@ sexp_node *Sexp_nodes = nullptr;
 sexp_variable Sexp_variables[MAX_SEXP_VARIABLES];
 sexp_variable Block_variables[MAX_SEXP_VARIABLES];			// used for compatibility with retail. 
 
+SCP_vector<sexp_container> Sexp_containers;
+// non-extern vars
+// we can't use full name (SCP_string) as the key, because get_sexp_container_index() takes const char*
+// so every call to get_sexp_container_index() would construct a string before lookup (ugh)
+static SCP_unordered_map<char, SCP_vector<int>> Container_indices_by_initial;
+
 int Num_special_expl_blocks;
 
 SCP_vector<int> Current_sexp_operator;
@@ -896,6 +906,7 @@ void build_extended_sexp_string(SCP_string &accumulator, int cur_node, int level
 void update_sexp_references(const char *old_name, const char *new_name, int format, int node);
 int sexp_determine_team(const char *subj);
 void init_sexp_vars();
+void init_sexp_containers();
 
 // for handling variables
 void add_block_variable(const char *text, const char *var_name, int type, int index);
@@ -959,6 +970,7 @@ void multi_sexp_modify_variable();
 SCP_vector<SCP_string> *Current_event_log_buffer;
 SCP_vector<SCP_string> *Current_event_log_variable_buffer;
 SCP_vector<SCP_string> *Current_event_log_argument_buffer;
+SCP_vector<SCP_string> *Current_event_log_container_buffer;
 
 // Goober5000 - arg_item class stuff, borrowed from sexp_list_item class stuff -------------
 void arg_item::add_data(char *str, int n)
@@ -1158,6 +1170,7 @@ void init_sexp()
 
 	sexp_nodes_init();
 	init_sexp_vars();
+	init_sexp_containers();
 	Locked_sexp_false = Locked_sexp_true = -1;
 
 	Locked_sexp_false = alloc_sexp("false", SEXP_LIST, SEXP_ATOM_OPERATOR, -1, -1);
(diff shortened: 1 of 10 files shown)


code/parse/sexp.h
```diff
@@ -1142,6 +1183,59 @@ typedef struct sexp_variable {
 	char	variable_name[TOKEN_LENGTH];
 } sexp_variable;
 
+struct sexp_container
+{
+	// meta-character for containers in text replacement, etc.
+	static constexpr char DELIM = '&';
```
jg18, line 1189: I still think we should consider using a different metacharacter like `'^'`.

A FREDder might reasonably want to use `'&'` in player-facing text.

jg18: BTW, in future containers PRs, I'm planning on replacing uses of the literal `'&'` with `sexp_container::DELIM`, which I added in this PR (it's not in PR #3398 ). Thoughts?

Karajorma: I'm fine with that.

Goober5000: Will the metacharacter's use here *prevent* its use in player-facing text?  As I recall, the $ can be used perfectly normally in most text; it only acquires special behavior when it is used as a color tag or placed before a variable.  If neither of these is true, the $ is displayed as expected.

If the & can be implemented the same way, then I think & is a better choice than ^.

code/parse/sexp.cpp
```diff
@@ -3758,6 +3771,164 @@ int get_sexp()
...
+	required_string("$Name:");
+	stuff_string(name, F_NAME);
+
+	if (name.empty()) {
+		Warning(LOCATION, "SEXP Container with empty name found");
+		log_printf(LOGFILE_EVENT_LOG, "SEXP Container with empty name found");
+		valid = false;
+	} else if (name.length() > sexp_container::NAME_MAX_LENGTH) {
+		Warning(LOCATION,
+			"SEXP Container name %s is longer than limit %u",
```
Goober5000, line 3800: Use `SIZE_T_ARG` instead of %u, here and elsewhere.  This is for cross-platform compatibility reasons.

code/parse/sexp.h
```diff
@@ -1142,6 +1181,59 @@ typedef struct sexp_variable {
...
+	}
+
+	int size() const
+	{
+		if (is_list()) {
+			return (int)list_data.size();
+		} else if (is_map()) {
+			return (int)map_data.size();
+		} else {
+			Assert(false);
```
Goober5000, line 1231: Use `UNREACHABLE` here.

Goober5000 (changes requested):
Added some comments.

Considering how long sexp.cpp is already, I would recommending putting the container stuff in a separate file.  M!m has mentioned in the past that sexp.cpp really needs to be broken up into smaller chunks, as it's too long for the text editor he uses.

code/parse/sexp.h
jg18: @Goober5000 Yes, if the text following a `&` is not a container name, then ,the text replacement code will skip over the `&`.

But why unnecessarily spend the computational overhead on checking whether the text following a `&` is a valid container name when we could just as easily use `^` as the meta-char instead?

code/parse/sexp.cpp
```diff
@@ -847,6 +851,12 @@ sexp_node *Sexp_nodes = nullptr;
 sexp_variable Sexp_variables[MAX_SEXP_VARIABLES];
 sexp_variable Block_variables[MAX_SEXP_VARIABLES];			// used for compatibility with retail. 
 
+SCP_vector<sexp_container> Sexp_containers;
+// non-extern vars
+// we can't use full name (SCP_string) as the key, because get_sexp_container_index() takes const char*
+// so every call to get_sexp_container_index() would construct a string before lookup (ugh)
+static SCP_unordered_map<char, SCP_vector<int>> Container_indices_by_initial;
```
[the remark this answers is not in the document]
jg18: Yup, I thought about using `const char *` as the key type but then realized that may not work as expected.

The Stack Overflow article doesn't seem relevant AFAICT, because it's about `std::map`, but we're using `std::unordered_map`, so the comparison after hashing is `==` and not `<`.

I also thought about using `SCP_string` as the key type but gave that a 👎🏻 because we'd then pay the cost of implicitly constructing an `SCP_string` object (because we mostly work with C-style strings) *every time* we need to check whether some (part of a) char buffer is a container name.

The benefit of using `Container_indices_by_initial`, compared to not using it:
1. It reduces the number of container names we have to check,, unless *every* container name in the mission starts with the same letter
2. It makes determining that a char sequence is *not* a container name potentially faster, in the case where the first letter of the char sequence isn't the first initial of any container name in the mission

Granted, whether using a lookup cache is really necessary/beneficial in the first place is a bit questionable. AFAIK, looking up a variable's index (or finding that no such variable exists) just does a linear scan through the variables vector, and that doesn't seem to be a perf bottleneck.

code/parse/sexp.cpp
jg18: Thanks for pointing out this issue.

Since C++11, `"%zu"` is supported for `size_t`, meaning it's portable across any compiler we'd want to use, and IMO it's easier to read, so I'd rather use that.

See the big table at https://en.cppreference.com/w/cpp/io/c/fprintf

That said, if SCP guidelines require using `SIZE_T_ARG` here, I'll switch to that.

jg18:
I give a big 👍🏻 to not adding more code to (and eventually breaking up) `sexp.cpp`, even if it didn't cause text editor issues, but all the more so if it does.

I'll create `code/parse/sexp_container.h` and `code/parse/sexp_container.cpp` unless there's a better option.

I'll also move any new header-related code to `sexp_container.h` , except for code related to new SEXPs or `Sexp_nodes`, since that'll obviously have to go in `sexp.h`.

As a nice bonus, that change should make my work on containers go faster. Any change to `sexp.h` requires re-compiling practically the entire codebase.

code/parse/sexp.h
Goober5000: The computational overhead is going to be spent regardless.  Both & and ^ are valid "player-facing" characters.  Here's an example from BTA:

> Cluster weapon. Each mortar carries three explosive submunitions. A full salvo will blanket an area approximately 1 km^3 large.

Using & instead of ^ is simply my personal preference.  But regardless of the character chosen, we can't assume that all uses of that character are necessarily containers.

code/parse/sexp.cpp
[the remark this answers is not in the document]
Goober5000: The distinction between `==` and `<` isn't relevant here.  The relevant point is that, without a custom comparison functor, the *pointer itself* is going to be evaluated, which is not going to work in cases where different pointers point to the same string.

And yes, I understand that the map is using `char`, not `char*`.  That's the reason for my EDIT in the comment above.

Anyway, I think `Container_indices_by_initial` is a case of premature optimization.  Yes, constructing a `SCP_string` object is not ideal, but I don't think this is performance-critical code.  You saw yourself that the variable lookup isn't a big problem.  The clarity and ease-of-use of using `SCP_string` outweigh the minor disadvantages.

code/parse/sexp.cpp
Goober5000: @asarium would be able to explain in more detail, but yes, this has been SCP policy for a while now.  And in the FSO codebase, `SIZE_T_ARG` is replaced by `"%zu"` on most platforms.

Goober5000:
> EDIT: Do `extern` functions in the new files require a special prefix like `sexp_container_` or something like that?

Maybe?  It depends on the function.  Whether `sexp_` or `sexp_container_`, use whatever prefix you think makes the most sense for that function.  We can change it if necessary during review.

code/parse/sexp.h
jg18: Okay and yes, the text replacement code handles meta-char instances that are not followed by a container name. If there's some case where the new code doesn't handle it properly, we can discuss during the review for that PR, when it comes up.

We can stick with `&` as the meta-char.

code/parse/sexp.cpp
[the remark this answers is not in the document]
jg18: Unfortunately, using `SCP_string` isn't quite straightforward, either, because it requires defining a custom hash function and equality comparison to ensure it's case-insensitive. Kara and I agreed in the containers thread that container names should be treated case-insensitively.

While that's certainly doable, it's honestly more work than I'd like to do for something that, like `Container_indices_by_initial` itself (as you rightly pointed out), is a case of premature optimization.

I'll just do a simple linear scan through `Sexp_containers` to find a container.

In the unlikely event that this approach can be demonstrated to be a bottleneck, we can look into optimizing it.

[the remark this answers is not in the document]
Goober5000: Well I had to figure out case-insensitivity already for the ship registry.  The hard work is already done; you just need to add the hash and comparator to the map declaration.  Here's how the ship registry map looks:

`SCP_unordered_map<SCP_string, int, SCP_string_lcase_hash, SCP_string_lcase_equal_to> Ship_registry_map;`

code/parse/sexp.h
jg18: Done. Glad there's a straightforward solution to this issue, because it pops up all over this place with this feature.

As a note, if we ever add a new container type (unlikely but possible), this block will become reachable if we forget to handle that new type in some spot, but then the code would need to be fixed, anyway.

code/parse/sexp.cpp
jg18: Actually, I'd forgotten that `sexp_container::NAME_MAX_LENGTH` is of type `int` and not `size_t`, so I'm now using `%d` to format it.

code/parse/sexp.cpp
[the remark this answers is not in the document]
jg18: Cool, nice work.

Well, that mostly solves it. The only remaining issue is `get_sexp_container_index_special()`, where the checking is by prefix, rather than match, so I'll just use a linear scan for that search. But that's only used for text replacement in messages/briefings/etc., so not used often.

jg18:
All comments I'm aware of should now have been addressed.

I also switched retrieval functions' return type from an index into `Sexp_containers` to a `sexp_container *`.

Advantages:
1. We can stop `Assert()`ing that the returned index is less than `Sexp_containers.size()`. This'd show up a good bit in later PRs otherwise.
2. A validity check by `!= nullptr` is nice and intuitive IMO.
3. We can stop having to use `Sexp_containers[index]` (very common in later PRs).
4. We now receive a properly typed pointer to work with, as opposed to a plain `int`.

Disadvantage is that indices into global containers are widely used elsewhere in the codebase. But with new code, why not use more modern approaches? Hence also the switch to typed `enum class` for container type info and list modifiers.

Goober5000:
This all looks good now, and I concur with your recently posted points.  One remaining question: does any of the missionsave code need to be ported to the QtFRED missionsave equivalent?

jg18:
> does any of the missionsave code need to be ported to the QtFRED missionsave equivalent?

Well, if we don't port it, then if a mission with containers is opened and saved in QtFRED, the container data will be lost. Is that a blocker for the PR?

What state is QtFRED in these days? Is it considered usable?

I'd consider adding full QtFRED support for containers, probably not until all of the functionality found in PR #3398 is merged.

jg18:
Note that even if we port the missionsave code to QtFRED, compatibility will break once we add container support to FRED2's `tree_nodes`.

jg18:
Just removed the event log stuff for containers, because it's not needed until `CTEXT()` support is added, and just realized that support was incomplete besides.

Goober5000:
I would port the missionsave stuff to QtFRED.  Don't worry about the `tree_node` stuff.  This matches other recent PRs, such as event annotations.  They will save and load in QtFRED, but visual/graphic stuff isn't supported yet.

Goober5000:
As for what state QtFRED is in, I'm not sure, but I think it's around 75%-ish done.

Goober5000 (approved):
Ok, I think this is good to go.

jg18:
Looks like I inadvertently made `ContainerType` an `enum` instead of `enum class`. Fixed now.

PR #3450, opened 2021-05-29
Minor beam sound fix

Baezon:
Beam 'launch' sounds only play if their loop sound is provided as well. There's no real reason for this to be the case, they are logically disconnected, all those sounds that are valid should be played when appropriate, irrespective of the status of other sounds.


wookieejedi (approved):
Makes sense, good point.

PR #3452, opened 2021-05-29
Add weapon ambient sound

Baezon:
Very confusingly $InFlightSnd is _not_ a 3d sound that the weapon emits, like EngineSnd for ships. It is a HUD sound that constantly plays while the player has one of these weapons in flight. This adds 'AmbientSnd' for weapons, as a counterpart to EngineSnd, any instance of this weapon passively emits this 3d sound as it moves.

### DIFF
--- a/code/weapon/weapon.h
+++ b/code/weapon/weapon.h
@@ -387,10 +387,11 @@ struct weapon_info
 	gamesnd_id	impact_snd;
 	gamesnd_id disarmed_impact_snd;
 	gamesnd_id	flyby_snd;							//	whizz-by sound, transmitted through weapon's portable atmosphere.
+	gamesnd_id	ambient_snd;
 	
-	gamesnd_id hud_tracking_snd; // Sound played when this weapon tracks a target
-	gamesnd_id hud_locked_snd; // Sound played when this weapon locked onto a target
-	gamesnd_id hud_in_flight_snd; // Sound played while the weapon is in flight
+	gamesnd_id hud_tracking_snd; // Sound played when the player is tracking a target with this weapon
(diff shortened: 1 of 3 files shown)


code/weapon/weapon.h
```diff
@@ -387,6 +387,7 @@ struct weapon_info
 	gamesnd_id	impact_snd;
 	gamesnd_id disarmed_impact_snd;
 	gamesnd_id	flyby_snd;							//	whizz-by sound, transmitted through weapon's portable atmosphere.
```
JohnAFernandez, line 389: We should fix this comment.  Otherwise, hurray for simple PRs!

JohnAFernandez (changes requested):

code/weapon/weapon.h
Baezon: Fix what? It's completely correct, aside from the obvious tongue-in-cheek joke

JohnAFernandez: Oh sorry, misread your PR description then.  What do you mean by: "plays while the player has one of these weapons in flight"?  It plays if the player fires it until that weapon dies?

Baezon: Yes, if the player fires it, it just plays until the weapon dies. It's a little clearer in the code where it is called `hud_in_flight_snd` (and grouped with the more obviously HUD sounds `hud_tracking_snd` and `hud_locked_snd`), but I plan on updating the wiki to clear up this possible confusion.

JohnAFernandez: Ok, that's what I'm saying should be added to that comment: that it's for player weapons only.

JohnAFernandez (approved):

PR #3454, opened 2021-05-29
Provide better error messages for invalid vec3ds in POF files

ngld:
Before:
```
Error: D:\Knossos\FS2\final-gen-1.2.0\core\data\missions\demo1.fs2(line 1):
Error: Expected float, found [ -0.972221$special=subsystem].

File: parselo.cpp
Line: 312
```

After:
```
Error: submodel turret10-base on INF_Selket.pof(line 1):
Error: Expected float, found [ -0.972221$special=subsystem].

File: parselo.cpp
Line: 312
```

This change also adds some logic to treat a `$` as a line break if necessary. That was added to fix the above error in already released mods since this error became visible only very recently and is present in an Inferno model released in 2015.

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -463,7 +463,7 @@ void get_user_prop_value(char *buf, char *value)
 }
 
 // routine to parse out a vec3d from a user property field of an object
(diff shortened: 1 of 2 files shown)


Goober5000 (approved):
Good changes. :+1:

PR #3457, opened 2021-05-30
Add weapon optimum range for AI fighters

Baezon:
Allows modders to set an 'optimum range' for a weapon which the AI will attempt to maintain while attacking. Sniper enemies will hangout from afar and shotgunners can get right up in your face.

Also the camel case instead of snake case on weapon min range bothered me.


code/ai/aibig.cpp
```diff
@@ -777,11 +777,23 @@ void ai_big_chase_attack(ai_info *aip, ship_info *sip, vec3d *enemy_pos, float d
 		
 		ai_turn_towards_vector(enemy_pos, Pl_objp, nullptr, rel_pos, 0.0f, 0, rvec);
 
+		// set an optimal range if applicable
+		float optimal_range = 0.0f;
+		ship_weapon* weapons = &Ships[Pl_objp->instance].weapons;
+		weapon_info* wip = nullptr;
+		if (weapons->num_primary_banks >= 1 && weapons->current_primary_bank >= 0) {
```
JohnAFernandez, line 784: We need a line in between `weapon_info* wip = nullptr;` and `if (weapons->num_primary_banks >= 1 && weapons->current_primary_bank >= 0) {` and at the bottom before the next if.  It helps when reading code to be able to identify blocks at a glance.

code/ai/aicode.cpp
```diff
@@ -6998,10 +6998,24 @@ void attack_set_accel(ai_info *aip, ship_info *sip, float dist_to_enemy, float d
 		return;
 	}
 
+	// assume an optimal range of 200m
```
JohnAFernandez, line 7001: I think we should add some detail here that it's from retail behavior.

code/ai/aicode.cpp
```diff
@@ -6998,10 +6998,24 @@ void attack_set_accel(ai_info *aip, ship_info *sip, float dist_to_enemy, float d
...
+	float optimal_range = 200.0f;
```
JohnAFernandez, line 7002: We should use a `constexpr` or `#define` to avoid magic numbers and so that this can be referenced elsewhere.

JohnAFernandez (changes requested):
Just some very minimal stuff.

JohnAFernandez (approved):
Thanks! Sorry for not seeing the changes you made!

JohnAFernandez:
If someone else was going to say anything about this pr, they would have done it already.

PR #3460, opened 2021-05-30
Fix heatseeker max homing limit for non-players

Baezon:
Like the comment itself says, `max_allowed_player_homers` should only be used for the player, so y'know... might be good to actually _check_ for that. >.>

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -4266,7 +4266,7 @@ void find_homing_object(object *weapon_objp, int num)
 					//	If this is a player object, make sure there aren't already too many homers.
 					//	Only in single player.  In multiplayer, we don't want to restrict it in dogfight on team vs. team.
 					//	For co-op, it's probably also OK.
-					if (!( Game_mode & GM_MULTIPLAYER )) {
+					if (!( Game_mode & GM_MULTIPLAYER ) && objp == Player_obj) {
 						int	num_homers = compute_num_homing_objects(objp);
 						if (The_mission.ai_profile->max_allowed_player_homers[Game_skill_level] < num_homers)
 							continue;


wookieejedi:
This is a good idea, but has possible side effects. IIRC the max attackers variable is also used with all ships even though the name suggests it is only for the player. Just from personal experience I know having these values used by the AI is actually very useful and affects mission balance. Of course adding another flag to specify this is just for the player is cumbersome, so I'm not sure what the best solution might be...

Baezon:
Uggggghhhh. There's another perfectly clearly named thing, that is in fact something completely different. Obviously, the fact that `$Max Player Attackers:` applies to all ships is extremely regrettable, and counter-intuitive, to say the least. _However_ `$Max Missiles Locked on Player:` *is* properly checked for the most part. When fired, missiles will properly discard their target if they exceed that value, and *only* for players. My change is only for the relatively niche scenario of a heatseeker needing to reacquire targets, where it would normally not check for that.

We *really* should be fixing the code to make the behavior *more* clear, rather than doubling down on the obfuscation. To me this a pure and simple bugfix, and very unlikely to actually affect balance. And if you ever need to limit the number of homers on non-players, I'd be more than happy to implement that as a separate value.

wookieejedi:
Very good points on all accounts.

wookieejedi (approved):

JohnAFernandez:
> Uggggghhhh. There's another perfectly clearly named thing, that is in fact something completely different. Obviously, the fact that `$Max Player Attackers:` applies to all ships is extremely regrettable, and counter-intuitive, to say the least. _However_ `$Max Missiles Locked on Player:` _is_ properly checked for the most part. When fired, missiles will properly discard their target if they exceed that value, and _only_ for players. My change is only for the relatively niche scenario of a heatseeker needing to reacquire targets, where it would normally not check for that.
> 
> We _really_ should be fixing the code to make the behavior _more_ clear, rather than doubling down on the obfuscation. To me this a pure and simple bugfix, and very unlikely to actually affect balance. And if you ever need to limit the number of homers on non-players, I'd be more than happy to implement that as a separate value.

I agree, heat seeker reacquire is a very rare event in any case.

PR #3461, opened 2021-05-31
Add random homing target acquisition

Baezon:
Instead of always grabbing the closest, heatseekers can also be specified to grab a random target from those available. This is only for the case where the missile must grab a target on its own, such as being an "untargeted heat seeker" or having lost its target.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -1576,6 +1576,18 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		}
 	}
 
+	if (optional_string("$Homing Auto-Target Method:"))
+	{
+		char	temp[NAME_LENGTH];
+		stuff_string(temp, F_NAME, NAME_LENGTH);
+		if (!stricmp(temp, NOX("CLOSEST")))
+			wip->auto_target_method = HomingAcquisitionType::CLOSEST;
+		else if (!stricmp(temp, NOX("RANDOM")))
+			wip->auto_target_method = HomingAcquisitionType::RANDOM;
+		else
+			Warning(LOCATION, "Homing weapon %s has an unrecognized Homing Auto-Target Method %s", wip->name, temp);
+	}
+
 	// swarm missiles
 	int s_count;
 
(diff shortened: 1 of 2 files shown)


code/weapon/weapons.cpp
```diff
@@ -4188,32 +4200,25 @@ void detonate_nearby_missiles(object* killer_objp, object* missile_objp)
...
-    ship_info   *sip;
-	float       best_dist;
-    int         homing_object_team;
-    float       dist;
-    float       dot;
-    vec3d       vec_to_object;
     ship_subsys *target_engines = NULL;
 
-	wp = &Weapons[num];
+	weapon* wp = &Weapons[num];
```
Goober5000, line 4205: You could use `auto` here and on the other declarations.  This is by no means required, I just wanted to make sure you were aware of it.

code/weapon/weapons.cpp
```diff
@@ -4283,28 +4288,44 @@ void find_homing_object(object *weapon_objp, int num)
...
+		// pick a random target from the valid ones
+		object* target = prospective_targets[Random::next((int)prospective_targets.size())];
+
+		wp->homing_object = target;
+		wp->target_sig = target->signature;
+		wp->homing_subsys = nullptr;
+
+		if (wip->wi_flags[Weapon::Info_Flags::Homing_javelin] && target->type == OBJ_SHIP) {
+			wp->homing_subsys = ship_get_closest_subsys_in_sight(&Ships[target->instance], SUBSYSTEM_ENGINE, &weapon_objp->pos);
+		}
```
Goober5000, line 4326: Isn't `target_engines` expected to be set at this point?  Why not use that?

Goober5000 (commented):
Looks good, but I have some comments.

code/weapon/weapons.cpp
Baezon: I am, yeah. In general I tend to prefer explicit types unless its some obnoxiously long C++ bullcrap.

code/weapon/weapons.cpp
Baezon: `target_engines` is going to be reset every time it scans a new ship, and for the closest method it'll stuff it into `homing_subsys` to maybe be overwritten later by a better candidate. But since I can only get retrieve a proper target at the end for random, I either have to carry along each target's engines in addition to their pointers, or just recalculate it.

Although now that you mention it, this would make javelins possibly grab bad targets, for which they can't even target the engines, which normally would've been ignored. Random javelins should probably only consider targets with visible engines...

Goober5000 (approved):
Ok, LGTM

PR #3462, opened 2021-05-31
karajorma's SEXP containers, Part 2: Persistence

jg18:
Build on PR #3449 by adding SEXP container persistence.

This PR adds support for
- saving containers to/loading containers from `.plr` and `.csg` files
- replacing a mission's containers with persistent ones and vice versa
- "Red Alert" containers


code/mission/missioncampaign.cpp
```diff
@@ -1079,10 +1083,59 @@ void mission_campaign_store_variables(int persistence_type, bool store_red_alert
...
+	}
+
+	for (const auto &container : get_all_sexp_containers()) {
+		if (!container.is_eternal()) {
+			if (any(container.type & persistence_type)) {
+				// see if we already have a container with this name
+				auto cpc_it = std::find_if(Campaign.persistent_containers.begin(),
+					Campaign.persistent_containers.end(),
+					[container](const sexp_container& cpc) {
+						return !stricmp(container.container_name.c_str(), cpc.container_name.c_str());
```
jg18, line 1105: Should this search also take container type comparison into account, in addition to name comparison? Or was that type checking already done in `parse_sexp_containers()`?

code/mission/missioncampaign.cpp
```diff
@@ -1079,10 +1083,59 @@ void mission_campaign_store_variables(int persistence_type, bool store_red_alert
...
+					Campaign.persistent_containers.emplace_back(container);
+				}
+			}
+		} else if (any(persistence_type & ContainerType::SAVE_ON_MISSION_PROGRESS) &&
+				   any(container.type & persistence_type) && container.is_eternal()) {
+			// we might need to save some eternal player-persistent containers
+			auto ppc_it = std::find_if(Player->containers.begin(),
+				Player->containers.end(),
+				[container](const sexp_container& ppc) {
+					return !stricmp(container.container_name.c_str(), ppc.container_name.c_str());
```
jg18, line 1121: Same question as above, about whether container comparison here should include container type (see `sexp_container::type_matches()`) and not just name.

code/mission/missioncampaign.cpp
```diff
@@ -1863,6 +1918,42 @@ void mission_campaign_save_on_close_variables()
...
+			continue;
+		}
+
+		// deal with eternals
+		if (container.is_eternal()) {
+			// check if container already exists and update it
+			auto ppc_it = std::find_if(Player->containers.begin(),
+				Player->containers.end(),
+				[container](const sexp_container& ppc) {
+					return !stricmp(container.container_name.c_str(), ppc.container_name.c_str());
```
jg18, line 1941: Same question a third time, about whether container comparison here should take container type into account and not just name.

code/mission/missionparse.cpp
```diff
@@ -5801,6 +5801,87 @@ void parse_sexp_containers()
...
+			auto &container = *p_container;
+
+			// if this is an eternal container that shares the same name as a non-eternal, warn but do nothing
+			if (container.is_eternal()) {
+				error_display(0,
+					"SEXP container %s is marked eternal but has the same name as another persistent container. One of "
+					"these should be renamed to avoid confusion",
+					container.container_name.c_str());
+			} else if (container.is_persistent()) {
+				if (container.type_matches(current_pc)) {
```
jg18, line 5826: See `sexp_container.cpp` for definition of `type_matches()`, but it's comparing only the non-persistence-related container type flags.

My reasoning for doing type matching in the first place was that replacing (i.e., updating) persistent containers shouldn't replace, for example:
- a map container with a list container
- a container using string data with a container using numeric data
- a map using string keys with a map using numeric keys

If/when strict typing is added, we should take that into account, too.

But @Karajorma what do you think should be the expected behavior here?

I don't think replacing persistent variables checks variable types, but type differences matter a lot more with containers.

code/mission/missionparse.cpp
```diff
@@ -5801,6 +5801,87 @@ void parse_sexp_containers()
...
+		auto *p_container = get_sexp_container(player_container.container_name.c_str());
+		if (p_container != nullptr) {
+			auto &container = *p_container;
+
+			if (container.is_persistent()) {
+				if (player_container.is_eternal() && !container.is_eternal()) {
+					// use the mission's non-eternal container over the player-persistent eternal container
+					continue;
+				} else {
+					if (container.type_matches(player_container)) {
```
jg18, line 5861: Same idea as above about using `type_matches()` here before replacing, but expected behavior is up for discussion.

jg18 (commented):
Brought up a few points for discussion.

JohnAFernandez:
Also, the original code patch written by Karajoma is #3253  for reference.

PR #3463, opened 2021-05-31
karajorma's SEXP containers, Part 3: "Add/Modify Container" FRED dialog

jg18:
Build on top of PR #3449 by adding FRED dialogs for SEXP containers, based on code from PR #3398.

This PR adds support for visually managing containers via the new "Add/Modify Container" dialog that is reachable from the right-click menu in FRED's Events Editor.


JohnAFernandez:
The original code written by Karajoma is here for reference: #3253

Goober5000 (approved):
LGTM I think.  I merely gave this a once-over, and didn't scrutinize it too closely, but I don't think adding a new dialog carries much risk.  The only area I have concern about is the mask #define, but we talked about that previously.

fred2/sexp_tree.cpp
```diff
@@ -6478,3 +6487,24 @@ int sexp_tree::get_loadout_variable_count(int var_index)
...
+
+int sexp_tree::get_container_usage_count(const char* container_name) const
+{
+	SCP_string container_name_lower = container_name;
+	SCP_tolower(container_name_lower);
+
+	int count = 0;
+
+	for (uint idx = 0; idx < tree_nodes.size(); idx++) {
+		// DISCUSSME: should nodes not marked SEXPT_VALID also be counted?
```
Goober5000, line 6499: In regards to this item, no, nodes without SEXPT_VALID should not be counted.

jg18:
As an side, if all of the PR's changes are in the `fred2` folder, then does the CI even need to run the Linux build jobs?

jg18:
Just passed a quick sanity check. I started FRED, added/edited a container, saved, quit and re-opened FRED, and checked that the container appeared as expected.

Issue #3459, opened 2021-05-30
New Random class trips assertions on bad user data


naomimyselfandi:
The new `Random.cpp` has an assertion to ensure it's getting a valid range:

```cpp
int Random::next(int modulus)
{
	Assert(modulus > 0);
	return SCP_rng.next() % modulus;
}

int Random::next(int low, int high)
{
	const int range = high - low + 1;
	Assert(range > 0);
	return low + (SCP_rng.next() % range);
}
```

However, these functions can be called directly by SEXP operators such as `rand` and `rand-multiple`, so the assertions can be misleadingly tripped by user data.

On a semi-related note, it'd be really nice if `rand` and `rand-multiple` would issue warnings when their arguments are backwards. If the warnings include the arguments themselves, then it'll be so much easier to track them down.

Goober5000:
This used to be silently converted to a zero-range, not asserted...
```
    // get diff - don't allow negative or zero
    diff = high - low;
    if (diff < 0)
        diff = 0;

    return (low + rand32() % (diff + 1));
```

Goober5000:
@jg18 do you want to handle this one?

jg18:
@Goober5000  Yeah I can take this up.

@naomimyselfandi Thanks for reporting this.

What should be the correct behavior in this situation, other than no Assertion failure? Should the function just return `0`?

jg18:
Or, probably a better solution, FSO/FRED code that calls these functions should validate argument values first, or at least validate in cases that handle user data. The caller code should be defensive against invalid input.

Speaking of which:

>On a semi-related note, it'd be really nice if `rand` and `rand-multiple` would issue warnings when their arguments are backwards. If the warnings include the arguments themselves, then it'll be so much easier to track them down.

@naomimyselfandi  That sounds useful to me. Could you file a separate issue with your request? Thanks again for your help.

naomimyselfandi:
> @naomimyselfandi That sounds useful to me. Could you file a separate issue with your request? Thanks again for your help.

I'm sorry, I didn't see this until now. Would you still like me to create this? (Feel free to shoot me a message on Discord; I'll see it faster that way!)

jg18:
No problem, and no need to file another issue (changed my mind). PR #3464  with the fix, which should be merged soon, will include the extra warning you requested. Thanks for including it.

jg18:
PR #3464 with the fix has been merged, so I am closing this issue as fixed.

[...]

PR #3464, opened 2021-05-31
Validate user-provided values passed to Random::next()

jg18:
Validate values supplied to `rand` and `rand-multiple` SEXPs and also to `rand32()` scripting function before calling `Random::next()`, and issue a warning if the provided values are invalid.

Fixes issue #3459.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -5096,7 +5096,8 @@ int rand_sexp(int node, bool multiple)
 
 	if (low > high) {
 		Warning(LOCATION, "rand%s was passed an invalid range (%d ... %d)!", multiple ? "-multiple" : "", low, high);
-		return SEXP_KNOWN_FALSE;
+		// preserve old behavior from before Random class was introduced
+		return low;
 	}
 
 	// get the random number
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -5094,6 +5094,11 @@ int rand_sexp(int node, bool multiple)
 	if (is_nan_forever)
 		return SEXP_NAN_FOREVER;
 
+	if (low > high) {
+		Warning(LOCATION, "rand%s was passed an invalid range (%d ... %d)!", multiple ? "-multiple" : "", low, high);
+		return SEXP_KNOWN_FALSE;
+	}
+
 	// get the random number
 	rand_num = rand_internal(low, high, seed);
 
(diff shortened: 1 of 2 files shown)


JohnAFernandez (approved):

code/parse/sexp.cpp
```diff
@@ -5094,6 +5094,11 @@ int rand_sexp(int node, bool multiple)
 	if (is_nan_forever)
 		return SEXP_NAN_FOREVER;
 
+	if (low > high) {
+		Warning(LOCATION, "rand%s was passed an invalid range (%d ... %d)!", multiple ? "-multiple" : "", low, high);
+		return SEXP_KNOWN_FALSE;
```
Goober5000, line 5099: This should return `low`, both for compatibility reasons (it's what used to be returned) and because caller sexps expect a number, not a boolean.

code/scripting/api/libs/base.cpp
```diff
@@ -62,12 +62,23 @@ ADE_FUNC(rand32,
...
-		result = Random::next(a, b);
-	else if (numargs == 1)
-		result = Random::next(a);
-	else
+	if (numargs == 2) {
+		if (a <= b) {
+			result = Random::next(a, b);
+		} else {
+			Warning(LOCATION, "rand32() script function was passed an invalid range (%d ... %d)!", a, b);
+			result = 0;
```
Goober5000, line 70: This should also return `low` to match the sexp.

Goober5000 (changes requested):

Goober5000 (approved):

jg18:
I'll wait to get @asarium's thoughts before I merge this.

code/scripting/api/libs/base.cpp
```diff
@@ -62,12 +62,23 @@ ADE_FUNC(rand32,
...
-	if (numargs == 2)
-		result = Random::next(a, b);
-	else if (numargs == 1)
-		result = Random::next(a);
-	else
+	if (numargs == 2) {
+		if (a <= b) {
+			result = Random::next(a, b);
+		} else {
+			Warning(LOCATION, "rand32() script function was passed an invalid range (%d ... %d)!", a, b);
```
asarium, line 69: This and the other should be a `LuaError` since that will give the script user a stack trace in their script.

asarium (changes requested):

asarium (approved):

PR #3465, opened 2021-06-01
Fix calling split_str() with a maximum line length of 0.

MageKing17:
Commit 6c2ebab01c2ab51c1fd1876370667d6f4400be7c (#3439) added a `max_line_length` argument to `split_str()`, but one call had the extra argument added in the wrong spot, and so a previously-pointless-but-harmless `ignored_char` of 0 became a `max_line_length` of 0, resulting (ultimately) in an exception in the unicode code when the iterator tried to decrement at the beginning of a ship's description. This gets rid of the argument entirely, rather than reordering them, because ignoring null characters doesn't really do anything anyway.

I've also added `Assertion`s to check `max_line_length` that should prevent this specific problem from occurring again in the future.

### DIFF
--- a/code/missionui/missionshipchoice.cpp
+++ b/code/missionui/missionshipchoice.cpp
@@ -1222,7 +1222,7 @@ void ship_select_blit_ship_info()
 	if(Ship_select_ship_info_text[0] != '\0'){
 		// split the string into multiple lines
 		// MageKing17: Changed to use the widths determined by Yarn here: http://scp.indiegames.us/mantis/view.php?id=3144#c16516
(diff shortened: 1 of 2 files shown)


BMagnu (approved):
Good catch.
The Assertions there are also a very good idea to make the function-signature change that #3439 introduces more safe.

BMagnu:
Just FYI, by getting rid of the argument instead of moving it, you change the separator from 0 to -1 (as that's the default), but that shouldn't make a difference in practice

MageKing17:
> Just FYI, by getting rid of the argument instead of moving it, you change the separator from 0 to -1 (as that's the default), but that shouldn't make a difference in practice

`cp == ignore_char` should never be true whether `ignore_char` is 0 or -1, hence why I called it "pointless-but-harmless".

wookieejedi (approved):

PR #3466, opened 2021-06-01
Add "no collide" for weapons

Baezon:
While a silly idea on the surface, modders can (and already do) use weapons for a wide variety of utilitarian purposes, far beyond just things you throw at other things to make them blow up. From triggering visual effects, audio effects, spawning other weapons, or any other thing might want them for, actual collisions might just be getting in the way of what they need.


wookieejedi (approved):
This is a good idea and brings many useful ideas to mind.

PR #3467, opened 2021-06-01
Fix Ai_aims_from_center flag

Baezon:
God _dang_ it, Sushi! Test your shit!! A single test with a seraphim or nephilim would tell you immediately that no, just because you put the flag in `set_predicted_enemy_pos` which, yes, is *one* of the places where it arbitrarily grabs the first firepoint does *not* mean the ship will start shooting from its center! Look at what the effect your code is. It's using that data to calculate *collision time* (from the fire point or center), which going to be negligibly affected by whether its counting from the firepoint or center anyway. At no point in that function does it actually *point the ship at something*. That's handled by `ai_turn_towards_vector`'s `rel_pos` parameter, so *that's* what you should be trying to affect.

I don't expect people to know all that, I know that because I've worked with that code before. But I do expect you to _test your code and see if your thing is even working at all_. And before anyone asks, yes, I checked the blame, all the elements which caused it to not work now also existed at the time it was implemented.

/rant

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -7554,7 +7554,7 @@ void ai_chase_attack(ai_info *aip, ship_info *sip, vec3d *predicted_enemy_pos, f
 	vec3d	new_pos;
 
 	start_bank = Ships[aip->shipnum].weapons.current_primary_bank;
-	if (po->n_guns && start_bank != -1 ) {
+	if (po->n_guns && start_bank != -1 && !(The_mission.ai_profile->flags[AI::Profile_Flags::Ai_aims_from_ship_center])) {
 		rel_pos = &po->gun_banks[start_bank].pnt[0];
 	} else
 		rel_pos = NULL;


JohnAFernandez:
So ... uh ... how does this fix it?

Baezon:
"That's handled by `ai_turn_towards_vector`'s `rel_pos` parameter, so that's what you should be trying to affect."

`rel_pos` is just that, a relative position to the "aiming object", instead of the normal way of calling `ai_turn_towards_vector` "just point me at this thing", adding in `rel_pos` is saying "aim me, but aim me such that my `rel_pos` is pointing at this thing". And its used precisely for the purpose of aligning an (arbitrary) firepoint at their target.

JohnAFernandez:
Ok, I see now.  But what happened to the original place where this flag was inserted?  I only see where we're adding where it's supposed to go.

Baezon:
That would be here https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/ai/aicode.cpp#L6759

JohnAFernandez (approved):
Ok, that answers my questions.  It would be kind of ridiculous to put this behind another flag, so I'll approve.

PR #3468, opened 2021-06-01
Fix a bug with applyWhack scripting function

Baezon:
Like other similar functions, the `offset` should be initialized to nullptr and then checked to see if the user filled it. By initializing with a reference to `vmd_zero_vector`, if the user doesn't fill it, that vector add can be modifying the zero vector! Yikes!

### DIFF
--- a/code/scripting/api/objs/physics_info.cpp
+++ b/code/scripting/api/objs/physics_info.cpp
@@ -464,13 +464,16 @@ ADE_FUNC(applyWhack, l_Physics, "vector Impulse, [ vector Position]", "Applies a
 	object_h objh;
 	physics_info_h *pih;
 	vec3d *impulse;
-	vec3d *offset = &vmd_zero_vector;
+	vec3d *offset = nullptr;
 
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/physics_info.cpp
```diff
@@ -464,13 +464,16 @@ ADE_FUNC(applyWhack, l_Physics, "vector Impulse, [ vector Position]", "Applies a
 	object_h objh;
 	physics_info_h *pih;
 	vec3d *impulse;
-	vec3d *offset = &vmd_zero_vector;
+	vec3d *offset = nullptr;
```
JohnAFernandez, line 467: Fantastic catch!! Did you happen to do a search to look for other instances of this?

JohnAFernandez: I think I'll do another search myself.  In the end, it might be safer to just make it a vector, instead of a pointer.  Others probably copied the safe code to make this buggy code.  I know one of my first iterations of rollback had some incredibly weird bugs from this very same cause.

JohnAFernandez: Ok, I looked through pretty much every instance of it, and you're right that this is the only case.  Approving.

JohnAFernandez (approved):

Issue #3143, opened 2021-01-14
Weapons/Ships FR: Optional rotation information for weapon external models.


EatThePath:
As it says on the tin. Would need extra info defined in the ship pof on a per firepoint basis.
Clarifying edit:
I am requesting the ability to change the static orientation of weapon submodels for specific firepoints, so that you can have missile rack models mounted on the top, bottom, and side of a ship, for instance. An example of planned use: https://p3d.in/cLXd4

Goober5000:
This is very vague.  What rotation information are you proposing, and how would it be used?

EatThePath:
Essentially I want to be able to associate a uvec with a firepoint, to be used along with the standard vector being the fvec, so that external gun models can be mounted sideways, upside down, etc. My understanding is that there is no way to do this presently.

Could potentially be a table thing instead of a pof thing I suppose, and it'd be potentially nice to have a per firepoint flip flag too, while I'm at it.

Aeesar has talked about his own preferences for solving this same problem, which are very different and as I recall consist of having submodels in the pof for each weapon type and firepoint. I can see the merit in that system too, but even if we had it I'd advocate for augmenting the current external model system with this.

Goober5000:
That would require removing some of the optimizations for that feature.  They currently use a lightweight rotation system that doesn't have some of the infrastructure of full-fledged rotating submodels.

EatThePath:
Would a lack of those optimizations be likely to cause notable problems or loss of performance?

Goober5000:
There would certainly be some degree of performance impact, but I couldn't estimate how much without profiling the code.  There would also be a non-trivial amount of code rewriting necessary.  In order to turn the lightweight system into the heavyweight system, there would need to be additional code infrastructure built.

In other words, it's not just optimization in terms of execution time, it's also optimization in terms of return on investment: incremental gain in feature vs. incremental cost in lines of code.

EatThePath:
In case it's needed, https://p3d.in/cLXd4 provides some context for my request. No matter how you slice them, neither the primary or secondary models are going to work as designed without some rotation capability.

Goober5000:
It's a neat-looking model, but how would the weapons rotate?  They all appear to be fixed in place.

EatThePath:
[...]

PR #3469, opened 2021-06-02
Add external weapon model angle offset

Goober5000:
@EatThePath pointed out that external weapon models are always rendered in the same orientation as the parent ship, and many models would benefit from being rotated to different orientations.  This adds an angle offset as a property of gun and missile points.  It requires a bump in the POF version, which consequently requires the extra information to be added via a hex editor, at least until PCS2 is upgraded.

Gun and missile points are also cleaned up a bit, mostly by being dynamically allocated so that they don't waste so much memory (about 4kb for every model in the game).

Implements #3143.  ~~This is in draft status until tested.~~

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -7214,7 +7214,7 @@ void HudGaugeHardpoints::render(float  /*frametime*/)
 					if (num_secondaries_rendered >= sp->weapons.secondary_bank_ammo[i])
 						break;
 
-					if(sp->secondary_point_reload_pct[i][k] <= 0.0)
+					if(sp->secondary_point_reload_pct.get(i, k) <= 0.0)
 						continue;
 
 					model_render_params weapon_render_info;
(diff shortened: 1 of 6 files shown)


Goober5000:
Tested, and a minor bug fixed.

BMagnu (approved):
The ``// NOLINT`` comments aren't particularly elegant, but given that we always check for nullptr before deletion, it's fine here IMO. 
Eventually I'd like for the external model to also rotate according to the points normal, but since that doesn't require pof changes it can be done in a different PR later down the line.

Goober5000:
> Eventually I'd like for the external model to also rotate according to the points normal

Can you clarify what you mean by this?

BMagnu:
There's a ship flag called gun convergence, and if that is enabled the ships weapons will fire along the normals set in the pof file as opposed to straight ahead.
And if missiles or guns fire, say, straight up instead of forwards, the external model should be rotated accordingly.
But yeah, since that is already part of the pof, it'd only need integration into the rotation code, similar to this feature

Goober5000:
Ok, I see what you mean.

PR #3470, opened 2021-06-03
Reimplement matrix3 multiplication based on a common overloaded operator

PhantomHoover:
This is the product of a general effort on my part to understand and clean up the code and documentation surrounding 3D vectors, especially when it comes to multiplication. The main consequence is that `vm_vec_rotate`, `vm_vec_unrotate` and `vm_matrix_x_matrix` are now wrappers around cleaner `operator*` instances — a great deal of manual indexing and arithmetic has been replaced with structured calls to `vm_vec_dot`. The potential for using this as a hook for future vectorisation has been discussed but is still speculative.

I really wanted to clean up matrix4 in the same way but it turns out to be a weird mess that relies on behaviour inconsistent with the 3x3 matrix ops. For now I'm leaving it alone, rather than trying to standardise it in its quasi-broken state.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -808,44 +808,22 @@ matrix *vm_vector_2_matrix_norm(matrix *m, const vec3d *fvec, const vec3d *uvec,
 }
 
 
-//rotates a vector through a matrix. returns ptr to dest vector
-//dest CANNOT equal source
-//
-// Goober5000: FYI, the result of rotating a normalized vector through a rotation matrix will
-// also be a normalized vector.  It took me awhile to verify online that this was true. ;)
+// rotates a vector through a matrix, writes to *dest and returns the pointer
+// if m is a rotation matrix it will preserve the length of *src, so normalised vectors will remain normalised
 vec3d *vm_vec_rotate(vec3d *dest, const vec3d *src, const matrix *m)
 {
-	Assert(dest != src);
-
-	dest->xyz.x = (src->xyz.x*m->vec.rvec.xyz.x)+(src->xyz.y*m->vec.rvec.xyz.y)+(src->xyz.z*m->vec.rvec.xyz.z);
-	dest->xyz.y = (src->xyz.x*m->vec.uvec.xyz.x)+(src->xyz.y*m->vec.uvec.xyz.y)+(src->xyz.z*m->vec.uvec.xyz.z);
-	dest->xyz.z = (src->xyz.x*m->vec.fvec.xyz.x)+(src->xyz.y*m->vec.fvec.xyz.y)+(src->xyz.z*m->vec.fvec.xyz.z);
+	*dest = (*m) * (*src);
 
 	return dest;
 }
 
-//rotates a vector through the transpose of the given matrix. 
-//returns ptr to dest vector
-//dest CANNOT equal source
-// This is a faster replacement for this common code sequence:
-//    vm_copy_transpose(&tempm,src_matrix);
-//    vm_vec_rotate(dst_vec,src_vect,&tempm);
-// Replace with:
-//    vm_vec_unrotate(dst_vec,src_vect, src_matrix)
-//
-// THIS DOES NOT ACTUALLY TRANSPOSE THE SOURCE MATRIX!!! So if
-// you need it transposed later on, you should use the 
-// vm_vec_transpose() / vm_vec_rotate() technique.
-//
-// Goober5000: FYI, the result of rotating a normalized vector through a rotation matrix will
-// also be a normalized vector.  It took me awhile to verify online that this was true. ;)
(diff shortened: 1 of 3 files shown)


Baezon (approved):
Yep looks good to me. Let's give it a bit in case anyone else has any input.

code/math/vecmat.cpp
```diff
@@ -808,44 +808,21 @@ matrix *vm_vector_2_matrix_norm(matrix *m, const vec3d *fvec, const vec3d *uvec,
...
-//    vm_vec_rotate(dst_vec,src_vect,&tempm);
-// Replace with:
-//    vm_vec_unrotate(dst_vec,src_vect, src_matrix)
-//
-// THIS DOES NOT ACTUALLY TRANSPOSE THE SOURCE MATRIX!!! So if
-// you need it transposed later on, you should use the 
-// vm_vec_transpose() / vm_vec_rotate() technique.
-//
-// Goober5000: FYI, the result of rotating a normalized vector through a rotation matrix will
-// also be a normalized vector.  It took me awhile to verify online that this was true. ;)
```
JohnAFernandez, line 841: A lot of the comments you added are helpful, but I don't see why we should remove all of these comments.

At the very least we should keep the last portion because not everyone is as well versed in Linear Algebra.

JohnAFernandez (changes requested):

code/math/vecmat.cpp
```diff
@@ -808,44 +808,21 @@ matrix *vm_vector_2_matrix_norm(matrix *m, const vec3d *fvec, const vec3d *uvec,
...
 
-//rotates a vector through a matrix. returns ptr to dest vector
-//dest CANNOT equal source
-//
-// Goober5000: FYI, the result of rotating a normalized vector through a rotation matrix will
-// also be a normalized vector.  It took me awhile to verify online that this was true. ;)
+// rotates a vector through a matrix, writes to *dest and returns the pointer
 vec3d *vm_vec_rotate(vec3d *dest, const vec3d *src, const matrix *m)
 {
-	Assert(dest != src);
```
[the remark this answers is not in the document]
Baezon: I'm not sure I follow, the assert was removed because the operation no longer requires it to be the case, its totally safe to have them be the same.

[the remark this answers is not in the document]
JohnAFernandez: Ok, I see the copies.  Did we profile these to make sure we don't have a performance hit by having to copy them?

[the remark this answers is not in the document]
PhantomHoover: I haven't profiled it beyond a basic "play the game and say yep, seems about the same" smoke test, and I'm honestly not sure how I'd go about it offhand, but I do doubt that it matters. The old approach of writing everything in place to a caller-supplied pointer is from the 90s when CPUs performed in very different ways to what they do now, and returning smallish structs by value was a lot rarer.

code/math/vecmat.cpp
Goober5000: Agreed, these comments should all be kept.  This is all important insight that coders may need to know.

code/math/vecmat.h
```diff
@@ -640,25 +640,13 @@ inline matrix& operator-=(matrix& left, const matrix& right)
...
-	vec3d out;
-	vm_vec_unrotate(&out, &right, &left);
-	return out;
-}
-
-inline matrix operator*(const matrix& left, const matrix& right) {
-	matrix out;
-	vm_matrix_x_matrix(&out, &left, &right);
-	return out;
-}
```
Goober5000, line 661: Given that the old inline operator used the wrong multiplication order, are there any leftover instances of multiplication that need their order fixed?

code/math/vecmat.cpp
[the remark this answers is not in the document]
JohnAFernandez: It would be better to get some hard numbers.  It's such an ubiquitous and essential part of FSO, that we should make sure it's not slowed down by this change.

code/math/vecmat.cpp
PhantomHoover: I think I just took it for granted that rotations preserve length, but I can expand this comment in a bit more explanatory detail.

code/math/vecmat.h
PhantomHoover: These operators were defined but literally never used, something I've just confirmed by deleting them from master and compiling without so much as a warning.

JohnAFernandez:
Tests ran well!  No slowdowns detected.  Please fix the comments!

code/math/vecmat.cpp
PhantomHoover: Comments extended, hope this is resolved.

JohnAFernandez (approved):
I'll take it.

PR #3472, opened 2021-06-03
deprecate script-eval

Goober5000:
The script-eval SEXP evaluates all arguments as individual script commands, which is not ideal because most FREDders expect the arguments to be concatenated.  Since the script-eval-block SEXP handles its arguments this way, it makes sense to hide the script-eval SEXP in FRED to prevent confusion.


wookieejedi (approved):
This is a good idea.

PR #3473, opened 2021-06-04
A few small fixes to SEXP container definition

jg18:


code/parse/sexp_container.h
```diff
@@ -28,18 +28,30 @@ enum class ContainerType : int {
...
 	using ContainerTypeInt = std::underlying_type<ContainerType>::type;
 	return static_cast<ContainerType>(static_cast<ContainerTypeInt>(lhs) |
 		static_cast<ContainerTypeInt>(rhs));
 }
 
-inline void operator|=(ContainerType lhs, ContainerType rhs) { lhs = lhs | rhs; }
+inline void operator|=(ContainerType &lhs, ContainerType rhs)
+{
+	lhs = lhs | rhs;
+}
```
Goober5000, line 54: I'm pretty sure compound assignment operators need to return a reference to themselves.  The return type should be `ContainerType&` instead of `void`, and the function should return `lhs`.

Goober5000 (changes requested):
Mostly looks good, but I have one comment.  Also why are these operators inline instead of members of the class?

code/parse/sexp_container.h
jg18: FWIW based on the fact that the code changes for PR #3463 compile, it seems like a `void` return type could work. But even if that's true, following conventions is a good thing, and maybe being able to chain `|=` will come in handy, anyway.

jg18: So [cppreference says](https://en.cppreference.com/w/cpp/language/operators) says
>Besides the restrictions above, the language puts no other constraints on what the overloaded operators do, or on the return type (it does not participate in overload resolution), but in general, overloaded operators are expected to behave as similar as possible to the built-in operators

And further down, it says bitwise operators are expected (not not required) to follow the pattern of arithmetic operators, and the arithmetic operators like `+=`, `-=`, etc. all return a reference.

But even more importantly, that page also says
>User-defined classes and enumerations that implement the requirements of [BitmaskType](https://en.cppreference.com/w/cpp/named_req/BitmaskType) are required to overload the bitwise arithmetic operators `operator&`, `operator|`, `operator^`, `operator~`, `operator&=`, `operator|=`, and `operator^=`

And once I do all of those overloads, we can get rid of the super-ugly `remove_container_type()` and then use the natural syntax like
```
container.type &= ~ContainerType::STRING_DATA;
```
or even
```
container.type &= ~ContainerType::STRING_DATA |= ContainerType::N_DATA;
```
And from the [BitMaskType](https://en.cppreference.com/w/cpp/named_req/BitmaskType) page , it sounds like the fact that, for example, `~ContainerType::STRING_DATA` is not a meaningful container type is irrelevant.

So looks like I have some more work to do.

jg18:
Thanks for your careful review.

> Also why are these operators inline instead of members of the class?
Despite the `class`  keyword, an `enum class` is just a strongly typed enumeration and not a class, meaning it can't have member functions. If you try it, you'll get a compiler error telling you that.

Also, from [this cppreference page](https://en.cppreference.com/w/cpp/language/enum) (see (1) under the heading "Scoped enumerations") it sounds like inheriting from `int` is redundant for an `enum class` (they inherit from `int` by default), so I could get rid of that inheritance.

code/parse/sexp_container.h
Goober5000: That's a good bit of research - which I missed when checking the compound assignment return type yesterday - and should certainly enhance the API usability.

code/parse/sexp_container.h
```diff
@@ -28,29 +28,44 @@ enum class ContainerType : int {
 	SAVE_TO_PLAYER_FILE = 0x40000000,
 };
 
+using ContainerTypeInt = std::underlying_type<ContainerType>::type;
+
 inline constexpr bool operator&(ContainerType lhs, ContainerType rhs)
```
Goober5000, line 33: Shouldn't this return `ContainerType` instead of `bool`?

jg18: Strictly speaking, yes, it should.

However, if we return `ContainerType` here, intuitive expressions like `container.type & ContainerType::STRING_KEYS` then fail to compile, and there is no way to define implicit type conversion from an `enum class` to `bool` or to the underlying type `int` (which could then be implicitly converted to `bool` by the usual method)

And luckily, with the way *we* use `&` with `ContainerType` in practice, we expect a "yes"/"no" answer, hence our using it in conditional expressions throughout.

That said, if there's a way to return `ContainerType` form this function **and** get the above expression to compile without hacks, I'm open to considering it.

jg18: FWIW, if for some reason, we actually need to get
```
ContainerType ct3 = ct1 & ct2;
```
we can still get it:
```
ContainerType ct3 = ct1;
ct3 &= ct2;
```

Goober5000: > intuitive expressions like `container.type & ContainerType::STRING_KEYS` then fail to compile

How do you mean?  What is the full expression? because that is just part of an expression that doesn't use the return type at all.

> if for some reason, we actually do need

I think we would need that expression often enough that the workaround would be cumbersome and awkward.

And more significantly, the `operator|` *does* have the expected return type.  It violates the principle of least astonishment to have `&` and `|` behave differently.

jg18: >How do you mean? What is the full expression?

Take a look at the `sexp_container` member functions `is_list()` and `is_map()` in `sexp_container.h` for examples of using `&` with `ContainerType` values to return a `bool` value, You can find examples of using `&` with `ContainerType` values as `if` conditions in the code changes to `missionsave.cpp` in PR #3449.

>I think we would need that expression often enough that the workaround would be cumbersome and awkward.

Offhand, I can't think of any instance of using `&` with a `sexp_container`'s `type` field (the original untyped `int` version) in the "full feature" PR #3398 where the intent is to use the result as a container type. Let me know if you find one, though.

>And more significantly, the `operator|` does have the expected return type. It violates the principle of least astonishment to have `&` and `|` behave differently.

Oh I agree that the current behavior with `operator&` returning `bool` is something of a convenient hack. It's a workaround for being unable to define implicit type conversion from an `enum class` to `bool`.

As such, I'll go looking for the least messy solution possible where `operator&`'s return type is changed to `ContainerType`  without excessively awkward rewordings of the usages of `&` I cited above.

FWIW the reason the "full feature" PR #3398 doesn't have this issue is that container type wasn't strongly typed/`enum class` then.

jg18: I attempted to fulfill your request.

I am, frankly, unhappy with the result, which IMO is now more verbose, more error-prone (e.g., correctly using `==` vs. `!= with `ContainerType::NONE`), and considerably harder to read, but I don't have a better idea of how to provide the behavior you asked for while keeping the code working properly.

Goober5000: This is better from a type safety perspective, and I don't think the readability suffers too much.  But you can improve it further by defining an `any` function and a `none` function, and then evaluate e.g. `any(ct & ContainerType::SOME_VALUE)`.  See this link for more info:
https://stackoverflow.com/questions/9874960/how-can-i-use-an-enum-class-in-a-boolean-context

jg18: Ah, so that is how the type conversion from `enum class` to `bool` is performed. I saw that SO page earlier but didn't understand before what it was saying.

I concede that the resulting code does indeed look better and, as you said, now uses proper type conversion.

Goober5000 (approved):
Much better. :+1: I think this is good to go now.

PR #3475, opened 2021-06-05
Remove superfluous warning about missing "child" flag.

MageKing17:
If a weapon has "inherit parent target" but not "child", a Warning will be generated. However, there is no dependency between the two flags; the weapon will function as the modder presumably expects. The Warning even seems to indicate this, by saying "No changes in behavior will occur" (as opposed to other similar warnings, which generally indicate that a flag will be ignored). Since the situation being warned about does not actually cause any issues, the Warning is entirely superfluous. As such, I've removed it outright.

Reported by @MatthTheGeek on Discord.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -692,11 +692,6 @@ void parse_wi_flags(weapon_info *weaponp, flagset<Weapon::Info_Flags> preset_wi_
         Warning(LOCATION, "\"smart spawn\" flag used without \"spawn\" flag in %s\n", weaponp->name);
     }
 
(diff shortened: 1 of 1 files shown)


JohnAFernandez:
Maybe what the warning is trying to say is that weapons already inherit the ship's target, and so the flag will effectively do nothing?

Baezon (approved):
The default behavior is no inheritance, you have opt into it with the flag. But retail was very separate with what were supposed to be child vs non-child weapons, at least organizationally in the tables, and I believe this largely superfluous "child" flag in general is just a reflection of that.

Behaviorally it doesn't matter much what is and isn't a "child", spawn just spawns whatever you want it to, so I would agree that this isn't necessary to warn about. I've hit it before and its nothing but an annoyance to the modder to throw this flag in there just to silence the warning.

JohnAFernandez (approved):
Thanks for the background @Baezon

PR #3476, opened 2021-06-06
add scripting functions to control sounds assigned to objects

Goober5000:
Scripters can now add and remove persistent sounds from objects, such as engine sounds or other ambient sounds.


code/scripting/api/objs/object.cpp
```diff
@@ -514,5 +517,76 @@ ADE_FUNC(addPostMoveHook, l_Object, "function(object object) => void callback",
 	return ADE_RETURN_NIL;
 }
 
+ADE_FUNC(assignSound, l_Object, "soundentry GameSnd, [vector Offset=nil, number Flags=0, subsystem Subsys=nil]",
```
asarium, line 520: If you want to have the flags as a parameter then you should also expose the relevant values as Lua enumerations. Otherwise a script writer would have to dig through the code to determine what flags were valid here.

asarium (changes requested):

code/scripting/api/objs/object.cpp
Goober5000: Is that as simple as adding the values to the `flag_def_list Enumerations[]` list in enums.cpp?

More importantly, will Lua allow the user to add two of them together and pass the sum to the function?

asarium: Yes, adding it to the enumerations should work just fine. It would be better to use the bitwise operation API since adding two enums is not guaranteed to give the correct result and is easy to accidentally misuse.

Goober5000: That doesn't look like quite the right approach.  I can map a new value `LE_OS_DS3D` to `OS_DS3D` easily enough, but if the new values are integers from 89 upward, then it won't be possible to use bitwise operations.  The scripter needs the ability to specify multiple flags if necessary.

asarium: You can choose arbitrary values for the enumerations. The index stuff is only relevant if you want to ensure that all values are distinct so there is no technical reason not to use bit flags with enums.

Goober5000: Okay, hopefully I did this right.

asarium (approved):

PR #3477, opened 2021-06-06
Fix collision culling for accelerating weapons... again

Baezon:
The collision system will predict how far into the future to defer its next collision check which does this based on the weapon's speed. If the weapon is accelerating however these predictions will be much too far in the future, as the prediction was made with its previous slower velocity in mind. While it's possible to have a secondary quadratic solution for the weapon's minimum speed in addition to its maximum speed it's much simpler to just exclude accelerating weapons from this and treat them more conservatively like missiles.

Note that once the weapon finishes accelerating it gets `CONST_VEL` and can benefit from this calculation again.


JohnAFernandez (approved):
This is much more clear-cut than the other one.

The-E (approved):

Issue #3356, opened 2021-04-02
Bug(?): Muzzleflashes are dimmer than they should be


Kestrellius:
When muzzleflashes appear in-game, they are often darker and more transparent than their animations suggest they should be. Even pure FFFFFF white will not be completely opaque. This problem becomes more or less severe under different conditions. It's not entirely clear what the conditions are, but I think it might have something to do with the muzzleflash clipping into the ship's geometry.

[A test case, with pure white bitmaps.](https://youtu.be/AUiBrd5EPXk)

Note that it's possible to see where the flashes are overlapping with each other and the laser bolts, despite the fact that they should all be pure opaque white. Also, contrast the lasers -- which are pure white, functioning correctly -- with the dimmer muzzleflashes.

[Another example, from FOTG, with actual art.](https://youtu.be/h_-BLyNVVgE)

Again, this muzzleflash bitmap is, in its central regions, pure FFFFFF white. It should be totally opaque at the center.

Observe the way the flashes become brighter or dimmer as the angle of viewing changes.

The inability to reliably create bright muzzleflashes is a problem, and I would like it if they were rendered as bright as their bitmaps.

Kestrellius:
The animations used in the test cases:
[muzzletesteffects1.zip](https://github.com/scp-fs2open/fs2open.github.com/files/6251704/muzzletesteffects1.zip)
[muzzletesteffects2.zip](https://github.com/scp-fs2open/fs2open.github.com/files/6251705/muzzletesteffects2.zip)

Baezon:
Gonna un-assign myself from this. As mentioned in #3479, beyond what I've done, I'm out of my depth.

wookieejedi:
#3479 fixed the issue for all intents and purposes, muzzle flashes look nice and bright now.

Kestrellius:
Yeah, I'm in agreement. This has more or less fully remedied the concerns I had. Though as Asteroth has said, I guess it's still probably worth looking into the remaining minor dimness for "why is this happening and will it cause any other problems" reasons.

MjnMixael:
Closing based on the conversation above.

PR #3479, opened 2021-06-08
Improve particle alpha handling for small particles

Baezon:
Particles start to fade out based on proximity at a hardcoded 30 meters, which can look very odd for particles smaller than 30 meters, so take the min of the two before starting to fade. 

This improves, but does not completely solve #3356. For reasons I'm unable to determine those pure white effects will still be noticeably shy of actually pure white even when the call to `batching_add_volume_bitmap` is passed 100% alpha (and that value survives all the way to the vertices being directly pushed to the batch), regardless of the view distance. Beyond that point, my confidence in being able to debug it drops rapidly.

### DIFF
--- a/code/particle/particle.cpp
+++ b/code/particle/particle.cpp
@@ -47,8 +47,8 @@ namespace
 		float dist;
 		float alpha = 0.99999f;
 
-		const float inner_radius = 30.0f;
-		const float magic_num = 2.75f;
+		const float inner_radius = MIN(30.0f, rad);
+		const float magic_num = MIN(2.75f, rad / 10.0f);
 
 		// determine what alpha to draw this bitmap with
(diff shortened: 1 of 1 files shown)


EatThePath:
> This improves, but does not completely solve #3356. For reasons I'm unable to determine those pure white effects will still be noticeably shy of actually pure white even when the call to batching_add_volume_bitmap is passed 100% alpha (and that value survives all the way to the vertices being directly pushed to the batch), regardless of the view distance. Beyond that point, my confidence in being able to debug it drops rapidly.

Just checking, is this true regardless of post-processing being on or off?

JohnAFernandez (approved):

Baezon:
> Just checking, is this true regardless of post-processing being on or off?
Looks to be unaffected by the flag.

PR #3480, opened 2021-06-08
Add No_shield_damage_from_ship_collisions flag

Baezon:
For some dumb reason, ship-ship collisions will try to count the damage as a shield hit for the heavier one. This adds a flag to make it hull damage for all involved parties.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -580,6 +580,9 @@ void parse_ai_profiles_tbl(const char *filename)
 						Warning(LOCATION, "Invalid ai secondary range awareness mode '%s' specified", buf);
 					}
 				}
+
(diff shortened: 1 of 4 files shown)


code/ai/ai_profiles.cpp
```diff
@@ -580,6 +580,9 @@ void parse_ai_profiles_tbl(const char *filename)
 						Warning(LOCATION, "Invalid ai secondary range awareness mode '%s' specified", buf);
 					}
 				}
+
+				set_flag(profile, "$all ships take hull damage in collisions:", AI::Profile_Flags::All_ships_take_hull_damage_from_collisions);
```
JohnAFernandez, line 584: Would "$no shield damage from collisions:" be a better?

Baezon: That might be a bit better, yeah.

JohnAFernandez (approved):

PR #3483, opened 2021-06-09
Improve damage type handling when dealing damage

### LINKED ISSUE #3482: Various sources of damage erroneously recieve the collision damage type
If a ship has a collision damage type specified, then instant blasts from it exploding and `~ + k` kills from the player if they are flying that ship will receive this damage type, despite not being due to a collision. `ship_do_damage` should probably have some more complicated way of determining that a collision happened instead of just polling the type of the `other_obj` (ditto for blasts to make sure they get the shockwave damage type)

Baezon:
I wouldn't have bothered so soon after posting the issue, but I kept running into it so I guess I gotta do something about it.

When dealing damage, damage callers should pass in the appropriate damage type, rather than having `ship_do_damage` make wild guesses based only on the type of the damaging object which cut down on those obnoxious chains of if elses in there. The changes to collidedebrisship.cpp look daunting but I had to do some varaible renaming to avoid awkward stuff like carrying around `pship` and `shipp` or something.

The main behavioral changes of note here are that instant blasts from ship explosions and weapons now deal their associated shockwave damage type. That might sound worrying for weapons, but remember that the shockwave damage type for a weapon is by default the same as the weapon. This will only make it use the shockwave damage type if it was explicitly provided, where before it would've silently ignored it (or used the collision damage type for ship explosions).

Awkward scenarios like cheat codes and even AI self-destruct attempts being possibly averted by the ship's collision damage type are also fixed, and inherit no damage type.

Fixes #3482

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -14116,7 +14116,7 @@ void ai_maybe_self_destruct(object *objp, ai_info *aip)
 		}
 
 		if (timestamp_elapsed(aip->self_destruct_timestamp)) {
-			ship_apply_local_damage( objp, objp, &objp->pos, objp->hull_strength*flFrametime + 1.0f, MISS_SHIELDS);
+			ship_apply_local_damage( objp, objp, &objp->pos, objp->hull_strength*flFrametime + 1.0f, -1, MISS_SHIELDS);
 		}
 	}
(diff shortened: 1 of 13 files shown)


Goober5000 (approved):
LGTM

PR #3485, opened 2021-06-10
support opening of web urls on PXO screens

notimaginative:
(copied out of icculus.org version)

### DIFF
--- a/code/network/multi_pxo.cpp
+++ b/code/network/multi_pxo.cpp
@@ -11,6 +11,8 @@
 
 #ifdef _WIN32
 #include <winsock2.h>
+#else
+#include <sys/wait.h>
 #endif
 
 #include "network/multi_pxo.h"
@@ -230,7 +232,7 @@ void multi_pxo_handle_disconnect();
 void multi_pxo_strip_space(char *string1,char *string2);
 
 // fire up the given URL
-void multi_pxo_url(char *url);
+void multi_pxo_url(const char *url);
 
 // unload the palette
 void multi_pxo_unload_palette();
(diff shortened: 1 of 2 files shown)


code/network/multi_pxo.cpp
```diff
@@ -1585,34 +1587,71 @@ void multi_pxo_strip_space(char *string1,char *string2)
...
-		}					
+#if defined(_WIN32) || defined(__APPLE__)
+	const char *open_cmd = "open";
+#else
+	const char *open_cmd = "xdg-open";
+#endif
+
+	char s_url[256];
+
+	// make sure it's a valid web uri
```
JohnAFernandez, line 1600: Typo uri instead of url?

code/network/multi_pxo.cpp
```diff
@@ -1585,34 +1587,71 @@ void multi_pxo_strip_space(char *string1,char *string2)
...
+
+		exit(rv);
+	} else {
+		waitpid(mpid, &statval, 0);
+
+		if ( WIFEXITED(statval) ) {
+			if (WEXITSTATUS(statval) == 0) {
+				return 0;
+			} else {
+				return -1;
```
JohnAFernandez, line 1634: Just as a note, I cannot tell at all what is happening with this section after the #else

JohnAFernandez (changes requested):

code/network/multi_pxo.cpp
notimaginative: The idea is to make sure we are passing a valid URI scheme to launch a web browser but not check that it's actually a valid URL that the browser can do something with. A basic security check so we aren't executing random commands but passing the bigger validation and security checks off to the browser.

So in context "uri" should be more appropriate than "url".

code/network/multi_pxo.cpp
notimaginative: For unix systems you need to fork off a child process in which to execute an external command. Depending on the return value of `fork()/vfork()` it indicates an error, whether it's the child process, or whether it's the parent process. If it's the child process then it execs the external command and exits. If it's the parent then it waits for the child to exit and checks the exit status to see if there was an error. So the highlighted block is the parent checking the return value of the child and returning the appropriate response.

JohnAFernandez: All right.  I'll give others a chance to look at it, since I don't know if what you're describing is being done correctly or not.  If it got into the icculus.org version, it's probably fine.

JohnAFernandez (approved):

JohnAFernandez:
Let me run a test to make sure it works on windows

JohnAFernandez:
Test worked great!  Only thing is, is there a way to add a cool down to keep it from opening a bunch of extra copies if the user gets impatient and clicks on a banner a bunch of times?

notimaginative:
Ah, good point. I'll add a short timeout to help prevent that.

notimaginative:
Timeout added. I also looked into a visual indicator that it was doing something, but that was either super messy or seriously beyond what this PR should be.

notimaginative:
@JohnAFernandez Want to take another look before it goes in?

JohnAFernandez (approved):

PR #3487, opened 2021-06-10
center the persistent sound by default

Goober5000:
The assumed location of a ship's "EngineSnd" is at its rearmost extent, which is fine for ships with engines, but not suitable for stationary objects such as space stations and comm nodes.  If the ship doesn't have an "engine" subsystem, place the persistent sound in the center.


Baezon (approved):
Sounds reasonable to me.

PR #3489, opened 2021-06-11
add some more sophisticated POF version handling

Goober5000:
This adds additional version definitions so that features can be added which support both the aligned and non-aligned POF formats.


wookieejedi (approved):
LGTM

PR #3490, opened 2021-06-12
SEXP containers, Part 4: Text replacement

jg18:
Support replacing container references in text anywhere variable replacement is supported (except in `modify-variable` SEXP) including in
- Messages
- Command briefings
- Briefings
- Debriefings
- Debug SEXP
- Scripting
- Subtitles

The syntax for list containers is
```
&ContainerName&Modifier&
```
where `Modifier` is one of
- `Get_First`
- `Get_Last`
- `Remove_First`
- `Remove_Last`
- `Get_Random`
- `Remove_Random`
- `At#` (where `#` is a zero-based index into the list)

The syntax for map containers is
```
&ContainerName&MapKey&
```

Map keys are case-sensitive, but container names and list modifiers aren't.

Based on PR #3398.

Also replaces `list_modifier::get_modifier()` with the simpler name `get_list_modifier()` (didn't seem worth creating a separate PR just for that).


jg18:
The PR currently doesn't support container text replacement in values assigned to variables using the `modify-variable` SEXP. Should it?

If so, should we support using *both* types of replacement in the same SEXP invocation?

Karajorma:
I'm not sure I follow. Replacement for SEXP-variables is already handled by the code.

jg18:
@Karajorma In the overload of `sexp_modify_variable()` that appears around line 30321 in sexp.cpp, the code looks for a `$` in the value to be assigned and if found, performs variable text replacement before assigning the value to the SEXP variable.

Should this function also check for `&` and if found, perform container text replacement?

If so, should it check for `&` only if `$` is not found?

Karajorma:
> 

Yeah, looking at it, I think it probably should. For consistency if nothing else. That said I was unaware that this was even an option so I don't know how many FREDders are aware of it.

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
...
+				key.c_str());
+			return false;
+		}
+
+		replacement_text = iter->second;
+
+		num_chars_to_replace += key.length();
+	} else {
+		// for list containers, check if this matches a list modifier
+		Assert(container.is_list());
```
JohnAFernandez, line 381: Another assert to Assertion would be helpful here.

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
...
+			Warning(LOCATION, "Attempt to replace text using empty list container %s", container.container_name.c_str());
+			return false;
+		}
+
+		const auto modifier = get_list_modifier(text.c_str() + lookHere, true);
+
+		if (modifier == ListModifier::INVALID) {
+			Warning(LOCATION,
+				"sexp_container_replace_refs_with_values() found a list container called %s to replace but the modifer "
+				"is not recognised",
```
JohnAFernandez, line 393: Again, I think I would be mystified if I saw this as a modder, and I wanted to figure out what to do.

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
...
+		case ListModifier::REMOVE_RANDOM:
+			data_index = util::Random::next((int)list_data.size());
+			std::advance(list_it, data_index);
+			replacement_text = *list_it;
+			list_data.erase(list_it);
+			break;
+
+		case ListModifier::AT_INDEX:
+			number_string = text.substr(lookHere + modifier_length);
+			Assert(!number_string.empty());
```
JohnAFernandez, line 438: Another Assert

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
...
+				return false;
+			}
+			replacement_text = *std::next(list_it, data_index);
+
+			// include spaces, in case the FREDder put one between "At" and the index
+			number_length = strspn(number_string.c_str(), " 0123456789");
+			break;
+
+		default:
+			Warning(LOCATION, "sexp_container_replace_refs_with_values() found a container called %s to replace but the modifer is not recognised", container.container_name.c_str());
```
JohnAFernandez, line 454: Would Unreachable be preferable here?

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
...
+bool sexp_container_replace_refs_with_values(char *text, size_t max_len)
+{
+	Assert(text != nullptr);
+
+	SCP_string text_str = text;
+
+	bool replaced_anything = sexp_container_replace_refs_with_values(text_str);
+
+	if (replaced_anything) {
+		Assert(max_len > 0);
```
JohnAFernandez, line 559: Another Assert

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
...
+**/
+bool sexp_container_replace_refs_with_values(SCP_string &text)
+{
+	bool replaced_anything = false;
+	size_t lookHere = 0;
+	int num_replacements = 0;
+
+	while (lookHere < text.length()) {
+		// look for the meta-character
+		size_t foundHere = text.find(sexp_container::DELIM, lookHere);
```
JohnAFernandez, line 484: Ah, I was wrong another camel case.  Again, not a huge deal, and I wouldn't hold back approval for it.

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
...
+
+				// + 2 for the enclosing '&' chars around the container name
+				size_t num_chars_to_replace = p_container->container_name.length() + 2;
+
+				// make sure the enclosing '&' appears after the container name
+				size_t expected_amp_pos = foundHere + num_chars_to_replace - 1;
+				if (expected_amp_pos >= text.length() || text[expected_amp_pos] != sexp_container::DELIM) {
+					// no modifier/key specified - ignore this matching container name, as the replacement is malformed
+					lookHere = foundHere + 1;
+					Warning(LOCATION,
```
JohnAFernandez, line 501: Again, should be an Assertion here, since the mission save function will add that last `&`?

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
...
+					// skip past the text we were unable to replace
+					lookHere = foundHere + num_chars_to_replace;
+					continue;
+				}
+
+				// perform the replacement
+				text.replace(foundHere, num_chars_to_replace, replacement_text);
+				replaced_anything = true;
+
+				++num_replacements;
```
JohnAFernandez, line 524: Another weird thing here, although this might just be me.  For me `++var` is specific to a case where we need to increment before the variable is used in an expression.  Every time I've read it during this review, I've thought, oh, this is specifically using this syntax, let me figure out why -- Oh nevermind, it's just incrementing the variable.  So for me it's a slight code readability issue, but again, it might just be me.

JohnAFernandez (changes requested):
Nothing structurally wrong that I could find.  Mostly, user-facing warnings and errors need to be changed.

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
 	return nullptr;
 }
 
+/**
+* Helper function for sexp_container_replace_refs_with_values(). Given a SEXP Container it works out what modifer was used and what the replacement string should be.
+**/
+bool get_replace_text_for_modifier(const SCP_string &text,
+	sexp_container &container,
+	const size_t lookHere,
```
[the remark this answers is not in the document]
jg18: The reason for camel case here is that this code was adapted from `sexp_replace_variable_names_with_values()`, i.e., text replacement for variables. I wanted to use the same local variable names that it does, to make the parallels between this code and that code easier to see.

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
...
+	SCP_string &replacement_text,
+	size_t &num_chars_to_replace)
+{
+	// for map containers, check if this matches a map key
+	if (container.is_map()) {
+		if (container.map_data.empty()) {
+			Warning(LOCATION, "Attempt to replace text using empty map container %s", container.container_name.c_str());
```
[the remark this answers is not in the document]
jg18: The FREDder can Ctrl+F in Notepad++ for the string "&<MapContainerName>&" where `<MapContainerName>` is the name of the map container that appears in the Warning. They then need to figure out why this map container is empty at the time this text gets replaced.

Since the warning is about text replacement, it's unrelated to SEXPs but rather is about player-facing text in some briefing/message/debriefing/subtitle/etc.

I'm not sure how to word this message to make it clearer to FREDders what they did wrong. I don't think the code has enough context at this point to tell the FREDder which message/briefing/etc. it was. Any ideas?

All that the player can do is report it, so that the FREDder knows their mission has a bug.

BTW, `Warning()` messages get written to the debug log, right?

code/parse/sexp_container.cpp
```diff
@@ -329,6 +337,233 @@ sexp_container *get_sexp_container_special(const SCP_string& text, size_t start_
...
+			return false;
+		}
+
+		const size_t key_ends_here = text.find(sexp_container::DELIM, lookHere);
+		if (key_ends_here == SCP_string::npos) {
+			Warning(LOCATION,
+				"sexp_container_replace_refs_with_values() found a map container called %s to replace but the key is "
+				"missing a final '&'",
```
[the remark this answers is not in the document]
jg18: This warning doesn't indicate a code issue but rather that the FREDder has malformed container text replacement syntax somewhere in their mission's player-facing text (a briefing/debriefing/message/etc.).

The PR description explains the syntax.

The FREDder is missing the final "&" character that tells FSO what part of the text that follows the second "&" is the map key to use.

code/parse/sexp_container.cpp
jg18: See my other comment explaining why camel case is used here.

Goober5000 (changes requested):
Only one comment besides @JohnAFernandez 's -- be careful to distinguish `max_len` from `max_size` due to the null terminator.  My own code makes that mistake too so I'll have to fix it.

EDIT: See #4159

code/parse/sexp_container.cpp
jg18: Yup, you're right. Will fix.

code/parse/sexp_container.cpp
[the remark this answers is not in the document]
JohnAFernandez: Got it.  Thanks for the explanation.

code/parse/sexp_container.cpp
JohnAFernandez: Got it.

code/parse/sexp_container.cpp
JohnAFernandez: Ah.  I did not know that.  Today I learned.

code/parse/sexp_container.cpp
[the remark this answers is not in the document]
JohnAFernandez: I think warning does get to the debug log, but @Goober5000 can you confirm?

code/parse/sexp_container.cpp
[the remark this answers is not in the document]
JohnAFernandez: All right.

code/parse/sexp_container.cpp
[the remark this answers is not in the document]
JohnAFernandez: Now that I fully understand the context, I think you're right that there isn't much we can do to clarify unless we want to spoon-feed them with a "Make sure the container has text before trying to use it for replacements!"

JohnAFernandez (approved):

JohnAFernandez:
My notes have been satisfactorily resolved.

Goober5000 (approved):
Sorry for the delay.  Everything looks good now. 👍

code/parse/sexp_container.cpp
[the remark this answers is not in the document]
Goober5000: Belatedly -- yes, warnings are written to the debug log.

PR #3493, opened 2021-06-13
fix to #3476

Goober5000:
The default sound position should be the zero-vector, not nullptr.  Also, this adds a bit of extra type safety.


JohnAFernandez (approved):
Looks really straightforward with some good secondary changes.

PR #3495, opened 2021-06-14
karajorma's SEXP containers, Part 5: Sexp_nodes

jg18:
Add support in FSO's SEXP tree for containers and container modifiers (list modifiers/map keys).

The PR includes
- Reading/writing container SEXP nodes (string format: `&ContainerName&`)
- `sexp_container_CTEXT()`
- Updated `check_sexp_syntax()` for containers
- Event log for containers


jg18:
On second thought, I'm concerned about the fact that this change currently breaks idempotency of `CTEXT()`. Can we be *absolutely* sure that `CTEXT()` will never get called multiple times for the same SEXP node?

OTOH, if we're just checking SEXP syntax, then the code ensures that container operations with side effects become no-ops, so maybe it's ok?

Karajorma:
Unfortunately, there isn't much that can be done to prevent CTEXT() from being called multiple times for the same node. But the alternative would be to have no way to alter a container once created, basically making them pointless.

Goober5000:
Yeah, CTEXT should be idempotent.  It's part of the expectation of the sexp contract.  And from a practical point of view, there are many, many places where CTEXT is used repeatedly to access the same node.

jg18:
Err, missed the replies. Yeah, I figured we'd need to keep `CTEXT()` idempotent.

One idea is to add a new field to `sexp_cached_data`, say `char ctext_result[TOKEN_LENGTH] = ""`, and then change `CTEXT()` to return that value if it's present. We'd have to make sure to clear the field when resetting the `sexp_node`.

That change would also mean we could get rid of the `Ctext_strings` array that this PR adds to sexp_container.cpp.

While I'm at it, I might as well change the type of `sexp_node::cache` to `std::unique_ptr<sexp_cached_data>`, so that we can stop having to manually manage that memory. Probably best for a separate preliminary PR, though.

Come to think of it, why *are* we dynamically allocating `sexp_cached_data`, anyway? Surely any space savings are more than outweighed by the runtime overhead of dynamic memory allocation.

Goober5000:
> Come to think of it, why _are_ we dynamically allocating `sexp_cached_data`, anyway? Surely any space savings are more than outweighed by the runtime overhead of dynamic memory allocation.

I did it that way because I it made sense to me.  Each node's cache should only ever be allocated at most once per mission, and not every node needs something cached.

jg18:
Ok, not yet tested, but `CTEXT()` with containers should now be idempotent.

code/parse/sexp.cpp
```diff
@@ -1947,6 +1948,29 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
+		} else if (Sexp_nodes[node].subtype == SEXP_ATOM_CONTAINER) {
+			const auto *p_container = get_sexp_container(Sexp_nodes[node].text);
+			Assert(p_container != nullptr); // should have been checked in get_sexp()
+
+			switch (type) {
+				case OPF_CONTAINER_NAME:
+				case OPF_LIST_CONTAINER_NAME:
+				case OPF_MAP_CONTAINER_NAME:
+					// set two values for extra certainty
+					// since SEXP_ATOM_CONTAINER == OPR_AI_GOAL (yay untyped enums)
```
Goober5000, line 1960: It shouldn't matter that these enums overlap.  Where is the other section of code that motivated the "set two values" redundancy here?

code/parse/sexp.cpp
```diff
@@ -3585,6 +3620,33 @@ int get_sexp()
 
 		}
 
+		// Sexp container
+		else if (*Mp == sexp_container::DELIM) {
+			Mp++;
+
+			char container_name[TOKEN_LENGTH];
+			const SCP_string container_delim(1, sexp_container::DELIM);
```
Goober5000, line 3628: If this is just an ad-hoc way to turn a char into a string, then why not add `DELIM_STR` as a sibling to `DELIM`?

code/parse/sexp.cpp
```diff
@@ -3585,6 +3620,33 @@ int get_sexp()
...
+			// bump past closing '&'
+			Mp += 2;
+
+			if (get_sexp_container(container_name) == nullptr) {
+				Error(LOCATION, "Attempt to use unknown container '%s'", container_name);
+				return -1;
+			}
+
+			// advance to the control options, since we'll read them when calling get_sexp() below
+			while (*Mp != '(') {
```
Goober5000, line 3640: If the string is malformed, then the loop could iterate way too far.  Add a sanity error-check within the loop (maybe check for `\n` or `\0`).

code/parse/sexp_container.cpp
```diff
@@ -37,6 +42,22 @@ const SCP_vector<list_modifier>& get_all_list_modifiers()
...
+static bool str_prefix(const char* prefix, const char* str)
+{
+	for (; *prefix; ++prefix, ++str) {
+		if (SCP_tolower(*str) != SCP_tolower(*prefix)) {
+			return false;
+		}
+	}
+
+	return true;
+}
```
Goober5000, line 59: Why not use `strnicmp`?

code/parse/sexp_container.cpp
```diff
@@ -92,14 +124,28 @@ void init_sexp_containers()
...
 		Containers_by_name_map.emplace(container.container_name, &container);
 	}
+
+	if (!renamed_containers.empty()) {
+		for (int i = 0; i < Num_sexp_nodes; i++) {
+			auto &node = Sexp_nodes[i];
+			if (node.type == SEXP_ATOM && node.subtype == SEXP_ATOM_CONTAINER) {
+				const auto new_name_it = renamed_containers.find(node.text);
+				if (new_name_it != renamed_containers.cend()) {
+					strcpy_s(node.text, new_name_it->second.c_str());
```
Goober5000, line 144: Copying something directly into `Sexp_nodes[].text` doesn't smell right.  I think I need to know more about how the design works.

code/parse/sexp.h
```diff
@@ -1111,6 +1119,9 @@ struct sexp_cached_data
 	int numeric_literal = 0;				// i.e. a number
 	int ship_registry_index = -1;			// because ship status is pretty common
 	void *pointer = nullptr;				// could be an IFF, a wing, a goal, or other unchanging reference
+	// jg18 - used to ensure sexp_container_CTEXT() is idempotent
+	char sexp_node_text[TOKEN_LENGTH] = "";
```
Goober5000, line 1123: There is already a `Sexp_nodes[].text` field.  As with the other review comment, I think I need to know more about how the design of containers is supposed to work.

code/parse/sexp_container.cpp
jg18: The reason is that the FREDder can rename containers in FRED's "Add/Modify Container" dialog. Since the text for container nodes is the container name, if the FREDder renames a container, then the SEXP nodes that refer to that container need to have their text changed.

I wonder how FRED/FSO handle the analogous issue of renaming variables. A quick test with FRED shows that when I rename a variable, references to it in the tree have their names updated. Are the variables' names not stored in the SEXP nodes' `text` field?

code/parse/sexp_container.cpp
jg18: Hmm, I guess I didn't know about `strnicmp()`. Will fix. Thanks.

code/parse/sexp.cpp
jg18: Good point, will fix.

code/parse/sexp_container.cpp
jg18: I'll have to check how renaming variables is implemented.

code/parse/sexp.h
jg18: The reason is that a given instance of "Replace Container" should be evaluated only once, then the computed value should be used after that. I'll work on creating an examplw.

With a node that uses "Replace Variable", is the latest value retrieved every time the node is processed with `CTEXT()`? My impression from the code is "yes".

jg18:
@Goober5000  As I mentioned to @Karajorma, I'm thinking of putting the following constraints on container usage in SEXP nodes. What do you think?

- In a SEXP that takes a container name as one of its arguments, that argument's value must be the literal name of a container and can't use Replace Variable or Replace Container.
- Similarly, if a SEXP takes a variable name as one of its arguments, that argument's value must be the literal name of a variable and also can't use Replace Variable/Replace Container. (In other words, FRED doesn't support pointer variables.) I'm pretty sure FRED/FSO already work this way with, e.g., `modify-variable`.
- However, other SEXP arguments can use Replace Variable or Replace Container.
- For Replace Container with list containers, the container modifier must be one of the supported list modifiers (e.g., `Get_First`) and can't use Replace Variable or Replace Container.
- However, for Replace Container with map containers, the container modifier can be any value that matches the container's key type, and the modifier *can* use Replace Variable or a (chained) call to Replace Container.

code/parse/sexp.h
jg18: Err, after noticing that text replacement in a message is applied anew each time the message is sent, maybe we *do* want Replace Container on a node to be evaluated more than once?

The tricky aspect is that some list container modifiers, such as `Remove_First`, have side effects, and we wanted to keep `CTEXT()` idempotent.

One (partial?) solution is to perform the side effect only when containers should be allowed to be modified, such as in-mission (see `are_containers_modifiable()` in sexp_container.cpp).

jg18: Another option is to disallow in "Replace Container" the list modifiers that have side effects (`Remove_First`, `Remove_Last`, `Remove_Random`) and support their functionality through new SEXPs instead. These modifiers could still be supported in text replacement.

But what effect would that change have on the feature's usefulness for FREDders? @Karajorma any thoughts?

Karajorma:
>And from a practical point of view, there are many, many places where CTEXT is used repeatedly to access the same node.

Why?

CTEXT() is a pretty heavy function. Why is anyone repeatedly calling it instead of calling it once and buffering the result?

code/parse/sexp.cpp
jg18: Tried to address this.

code/parse/sexp.h
jg18: I renamed the field to `container_CTEXT_result` and removed the `bool` field.

The problem this solves is that `sexp_container_CTEXT()` dynamically generates its result and thus we require space to store it. In the cases of list modifiers that remove the retrieved value (e.g., `Remove_First`), the original value won't even exist anymore by the time `sexp_container_CTEXT()` returns.

jg18:
From @Karajorma, the proposed new idempotency contract for `CTEXT()` is that it remains idempotent *except* when all of the following are true:
- `CTEXT()` is called on a "Replace Container" node whose container modifier is a list modifier that removes the retrieved item (e.g., `Remove_Last`).
- `are_containers_modifiable()` returns `true` (see sexp_container.cpp).

Is this change problematic?

code/parse/sexp_container.cpp
jg18: Ok, found it. This code is doing the container equivalent of `sexp_tree::modify_sexp_tree_variable()`.

The differences are
- You can rename multiple containers while using the Add/Modify Container dialog, but AFAICT the Modify Variable dialog allows you to rename only one variable at a time.
- With variables, any of the type/name/default value can chance, so the node needs to be fully updated. However, containers don't have default values, and their type info is stored with the container objects in `Sexp_containers`, so the only value in the tree node that needs to be updated is the name.

Let me know if that clarifies things.

code/parse/sexp.h
jg18: Oh wait, now I get your concern. Yeah, I think you're right, we should be updating the text in `tree_nodes` instead and then allowing FRED to update `Sexp_nodes` when it saves its view of the SEXP tree.

If that's the case, then if the FREDder makes changes in the Add/Modify Container dialog and clicks OK om that dialog *but* then presses Cancel in the Events Editor, should none of the container changes be applied to `Sexp_containers`? Does modifying variables work that way?

If so, then the current Add/Modify Container dialog behavior is incorrect, because that dialog updates `Sexp_containers` after you click OK in it, regardless of what you click in the Events Editor. I'll create a separate PR to fix that.

code/parse/sexp_container.cpp
jg18: Oh wait, now I get your concern. Yeah, I think you're right, we should be updating the text in `tree_nodes` instead and then allowing FRED to update `Sexp_nodes` when it saves its view of the SEXP tree.

If that's the case, then if the FREDder makes changes in the Add/Modify Container dialog and clicks OK om that dialog *but* then presses Cancel in the Events Editor, should none of the container changes be applied to `Sexp_containers`? Does modifying variables work that way?

If so, then the current Add/Modify Container dialog behavior is incorrect, because that dialog updates `Sexp_containers` after you click OK in it, regardless of what you click in the Events Editor. I'll create a separate PR to fix that.

jg18: Well... wait a sec. What should happen if the FREDder renames containers but then presses Cancel in the Events Editor? Will affected "Replace Container" nodes in `Sexp_nodes` get their text updated to the new container names?

I'll have to check how this situation is handled for variables.

jg18: Interesting. When I rename a variable in the Modify Variable dialog and click OK, then click Cancel in the Events Editor, the result is an assertion failure when trying to save the mission, specifically at 
```
Assert(sexp_variable_index != -1);
```
in `CTEXT()` in the middle of `check_sexp_syntax()`.

So it looks like this (admittedly improbable) scenario isn't handled correctly for variables. I'd say that qualifies as a bug.

Karajorma: > Interesting. When I rename a variable in the Modify Variable dialog and click OK, then click Cancel in the Events Editor, the result is an assertion failure when trying to save the mission, assuming that `Sexp_nodes` had a "Replace Variable" node with the affected variable prior to opening the Events Editor.
> 
> ```
> Assert(sexp_variable_index != -1);
> ```
> 
> in `CTEXT()` in the middle of `check_sexp_syntax()`.
> 
> So it looks like this (admittedly improbable) scenario isn't handled correctly for variables. I'd say that qualifies as a bug.

I vaguely remember stumbling across that bug with the original container code (which did things the same way) and then fixing it. I didn't think to check if the variable code I'd based it on also had the same bug.

jg18: I filed Issue #3849 for the above FRED crash.

To prevent this same crash for containers, I think we should keep these `Sexp_nodes` updates.

When I add container support to FRED's SEXP tree in the next PR, then I'll also update node text in `tree_nodes` when containers are renamed.

In both cases, the text is the only value that needs to be updated when containers are modified, since all info on types/contents is stored in `Sexp_containers` itself.

code/parse/sexp.cpp
jg18: I removed this code as part of an overhaul of SEXP syntax checking for containers.

TRBlount (approved):
I am going to approve this. Nothing here should break any player/fredder-facing code.

jg18:
Per discussion with @TRBlount on Discord, any follow-up comments on these changes can be addressed in a follow-up PR. Merging this PR now ensures it gets into 22.0.

CC: @Goober5000 @Karajorma

Issue #3367, opened 2021-04-09
Cutscene Flags Don't Work On Librocket Interfaces


MjnMixael:
@asarium 

Cutscenes that should play before Command Briefing stages do not play. This will affect all stages should the HD UI ever be completed any further.

These flags specifically. (You cannot currently add them via FRED.) https://wiki.hard-light.net/index.php/Cutscene_flags

Goober5000:
To document what's been going on behind the scenes, I looked into this a week ago and messaged with @asarium about how it could be solved.  It's basically a two-part fix.  Part 1 is to implement the required scripting function, which is now done.  Part 2 is to modify the UI script to use the new function.

Essentially, since the custom UI is controlled through its own script, the script has to handle everything associated with a state transition.  And that includes, in this case, playing the cutscene.

I'll have to defer to @asarium on how the script should be modified.  I assume the relevant file to change is `ui_system.lua`, but I don't know exactly where the cutscene should be plugged in.

MjnMixael:
I assume this is another case of abandoned librocket UI. I have not yet heard from @asarium on adding this functionality to the necessary lua files.

Goober5000:
I wouldn't call it abandoned.  I assume @asarium is either busy or occupied with RL priorities.

z64555:
Setting to low priority and assigning it to 21.6 so it hopefully doesn't get buried again.

Goober5000:
I'm not familiar with what is necessary for Part 2 of this.

MjnMixael:
Resolved

PR #3496, opened 2021-06-14
add a scripting function for playing cutscenes in the UI

Goober5000:
Implements `common_maybe_play_cutscene` as a scripting function.  Partial fix for #3367.

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -19,6 +19,7 @@
 #include "scpui/rocket_ui.h"
 #include "scripting/api/objs/cmd_brief.h"
 #include "scripting/api/objs/color.h"
+#include "scripting/api/objs/enums.h"
 #include "scripting/api/objs/player.h"
 #include "scripting/lua/LuaTable.h"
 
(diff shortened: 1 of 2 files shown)


code/scripting/api/libs/ui.cpp
```diff
@@ -110,6 +111,19 @@ ADE_FUNC(playElementSound,
...
+	enum_h movie_type;
+	bool restart_music = false;
+	int score_index = 0;
+
+	if (!ade_get_args(L, "*obi", l_Enum.Get(&movie_type), &restart_music, &score_index))
+		return ADE_RETURN_NIL;
+
+	// if an out-of-range enum is passed to this function, it will just be ignored
+	common_maybe_play_cutscene(movie_type.index, restart_music, score_index);
+}
```
asarium, line 125: A `return ADE_RETURN_NIL;` is missing here.

asarium (approved):

code/scripting/api/libs/ui.cpp
Goober5000: Whoops.  Wonder how that compiled successfully locally...

PR #3497, opened 2021-06-15
fix missiles locking during pause

### LINKED ISSUE #3491: Exploitable bug: Lock acquisiton continues durring viewer pause
As the title. If you go into view pause mode, you can lock your missile(s) 'instantly' and then fire after unpausing. 

Secondairly, you can also spread locks on multilock missiles in illegal ways, as switching targets does not drop existing locks unless it is switching to a friendly target. To do this, you take a weapon that locks multiple missiles on one target sequentially, pause after the locking starts, then switch targets. The first lock sequence will continue, and then the next lock will start with your new target as  long as that target would be a typically valid one. In many cases the locks will be dropped as soon as you unpause, but if you cycle back to your original target(or possibly you just need to be on a valid target, testing has not been too extensive) you can keep your illegal locks and fire them successfully.

JohnAFernandez:
Fixes #3491

### DIFF
--- a/code/hud/hudlock.cpp
+++ b/code/hud/hudlock.cpp
@@ -12,6 +12,7 @@
 
 #include "ai/ai.h"
 #include "debugconsole/console.h"
+#include "gamesequence/gamesequence.h"
 #include "gamesnd/gamesnd.h"
 #include "globalincs/linklist.h"
 #include "hud/hudlock.h"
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):

Goober5000 (commented):
This looks good, but I noticed that no other code refers to `GS_STATE_GAME_PAUSED`.  Would it be appropriate to use `physics_paused` or `ai_paused` instead?

JohnAFernandez:
I'll try physics paused, that makes more sense.

JohnAFernandez:
Neither of those works.  It goes to another game state to pause, so it never changes those variables.

Goober5000 (approved):
Ok, I poked around a bit, and this does look like the best way to do it.

PR #3498, opened 2021-06-15
Add Menu and Mission resetting to scripting state.

### LINKED ISSUE #3474: Assert when exiting scripted menu then reloading mission
Happens on debug build only. Tested on current (1.1.5) BP Complete with 21.3.0.20210603_22be402de

My assumption is parts of the missions are not unloaded properly because we exit the mission in an unexpected way.

Steps to reproduce: 
- Load BP1 mission bp-01-1.fs2. 
- From briefing, enter the database with the button on top right. 
- Exit with escape (not with the exit button), this brings you back to the mainhall.
- Load the same mission again.

```
ASSERTION: "(check == this || !check)" at jumpnode.cpp:251
 Jumpnode Jump Node 1 is being renamed to Delta Serpentis, but a jump node with that name already exists in the mission!


Assert: "(check == this || !check)"
File: jumpnode.cpp
Line: 251
Jumpnode Jump Node 1 is being renamed to Delta Serpentis, but a jump node with that name already exists in the mission!
```

JohnAFernandez:
This adds the scripting state to the list of states that checks whether it should reset data or not.

Fixes #3474

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -5439,6 +5439,14 @@ void game_leave_state( int old_state, int new_state )
 			break;
 
 		case GS_STATE_SCRIPTING:
+			// this can happen because scripting can be done in odd places.
+			if ( !going_to_briefing_state(new_state) ) {
+				common_select_close();
+
+				if ( going_to_menu_state(new_state) ) {
+					freespace_stop_mission();
+				}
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Aha.  I think you've solved it.

PR #3499, opened 2021-06-15
Add escape character for < and > in html output

### LINKED ISSUE #3390: Malformed documentation in SEXP HTML engine output, for argument-using SEXPs
All instances of `<argument>` in said documentation are unescaped, or whatever the correct term in HTML is, and so they get read as HTML tags and don't get rendered when the page is opened in a browser. This can make the documentation for SEXPs that deal with arguments somewhat incoherent.

JohnAFernandez:
Fixes #3390


Goober5000 (approved):
Looks good, but as a nitpick, I would recommend matching the brace style of the rest of the function

PR #3500, opened 2021-06-15
Animation Overhaul: Phase 1

BMagnu:
This is the first PR regarding to the eventual full rewrite of the animation system. It introduces the framework, and replaces the simplest type of existing animation, the initial-type animations (as to keep the PR manageably small). It will be followed up by the following PRs eventually:
- Phase 2: Adding required components to match the old system, and replace / remove the old system
- Phase 3: Add parsing for the new animation table that will allow much more flexible animations, and introduce translations
- Phase 4: Extend the system with new features as desired

To explain how this system works and differs from the old one:
First, an animation now is _not_ bound to a subsystem anymore. Instead, animations are defined on a ship class, and executed on a ship instance. Internally, each animation can operate on multiple submodels.
Secondly, animations are not handled by one large function anymore. Instead, an animation is comprised of segments, each of which has to calculate their own delta to the model's state, allowing modularization and extendability by just adding new segments with new features as desired.


Goober5000 (commented):
I reviewed it twice and it looks good to me.  Before explicitly approving it I'd like to download the branch and run a few actual tests.  Also, I'd like the opinions of a few other coders, especially @asarium.

code/model/modelanimation.cpp
```diff
@@ -10,7 +10,9 @@ namespace animation {
 
 	ModelAnimationState ModelAnimation::play(float frametime, ship* ship) {
 
-		switch (m_state) {
+		float& time = m_time[ship];
```
Goober5000: Could you use a different name?  This could cause problems if we add/change a time library, or if we have some other `time` struct or definition.

code/model/modelanimation.cpp
```diff
@@ -6,23 +6,23 @@ extern float flFrametime;
 
 namespace animation {
 
-	std::multimap<ship*, std::shared_ptr<ModelAnimation>> ModelAnimation::s_runningAnimations;
+	std::multimap<polymodel_instance*, std::shared_ptr<ModelAnimation>> ModelAnimation::s_runningAnimations;
```
Goober5000, line 9: I think these maps are going to persist for more than one frame.  In this case, the map indexes should be the integer id (shipnum or objnum, and pmi->id).  The `Polymodel_instance` collection is a vector, and the `Ships` collection may (someday) also be a vector.  Vectors may re-allocate their storage when their size increases, which means that all pointers will be invalidated.

code/lab/manager/lab_manager.h
```diff
@@ -28,16 +28,8 @@ class LabManager {
...
-			obj_delete(CurrentObject);
-		CurrentObject = -1;
-
 		animation::ModelAnimation::clearAnimations();
-
-		for (auto d : Dialogs) {
-			d->close();
-		}
-
+
```
Goober5000: Uh... was it correct to remove this code?

Goober5000 (changes requested):
Reviewed the latest changes.

Goober5000 (approved):
Looks good! :+1:  With @asarium and me reviewing the design, and the initial animation steps tested, and looking through the code several times -- I'm ready to approve it.

Issue #3375, opened 2021-04-11
$Enable In-Game Options Disables Settings


MjnMixael:
This flag, available in Game Settings.tbl seems to disable options like lightshafts and bloom. There does not seem to be options to enable these features in the new Options menu (which does require custom assets... of which my mod does contain). Not sure if that's just unfinished or an oversight. If it's just unfinished, then close this issue and I'll disable this flag for now.

jg18:
@MjnMixael I think I may have a fix. Please see your HLP Forums PMs for a link to a test build, which is based on PR #3501.

Are there any other settings affected by this issue, in addition to lightshafts and bloom?

Thanks!

jg18:
This issue should have been fixed by PR #3501; awaiting confirmation.

z64555:
Did we ever get confirmation of this being fixed or not?

jg18:
@z64555 Not that I'm aware of.

MjnMixael:
This is not fixed. At least not out of the box. If additional work is needed to expose the fix in the librocket Options Menu, I don't have the skill to do that.

However, since the librocket UI seems to be currently abandoned, I'm not sure this is really a priority anymore.

MjnMixael:
I have disabled this flag in BtA. Until In-Game Options becomes more widespread (I don't know of any other mod that uses it), this does not need to be High Priority.

MjnMixael:
Based on my current testing on 22.2, this seems to be fixed.

PR #3501, opened 2021-06-15
Add in-game options menu options for lightshafts and bloom

jg18:
***This is experimental work.***

Intended to address Issue #3375.

### DIFF
--- a/code/graphics/post_processing.cpp
+++ b/code/graphics/post_processing.cpp
@@ -195,6 +195,15 @@ SCP_vector<graphics::post_effect_t>& PostProcessingManager::getPostEffects() { r
 const lightshaft_parameters& PostProcessingManager::getLightshaftParams() const { return m_lightshaftParams; }
 lightshaft_parameters& PostProcessingManager::getLightshaftParams() { return m_lightshaftParams; }
 
+bool PostProcessingManager::bloomShadersOk() const
+{
+	return m_bloomShadersOk;
+}
+
+void PostProcessingManager::setBloomShadersOk(bool ok)
+{
+	m_bloomShadersOk = ok;
+}
 } // namespace graphics
 
 bool gr_lightshafts_enabled()
--- a/code/graphics/post_processing.cpp
+++ b/code/graphics/post_processing.cpp
@@ -2,6 +2,7 @@
 #include "post_processing.h"
 
 #include "graphics/grinternal.h"
+#include "options/Option.h"
 #include "parse/parselo.h"
 #include "ship/ship.h"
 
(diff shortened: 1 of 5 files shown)


code/graphics/post_processing.h
```diff
@@ -56,12 +56,19 @@ class PostProcessingManager {
 	const lightshaft_parameters& getLightshaftParams() const;
 	lightshaft_parameters& getLightshaftParams();
 
+	bool bloomShadersOk() const { return m_bloomShadersOk; }
+	void setBloomShadersOk(bool ok) { m_bloomShadersOk = ok; }
```
asarium, line 60: Function bodies should be in the source file.

asarium (changes requested):

jg18:
Still pending verification from Mjn. I sent him a link to a test build over HLP Forums PM.

Since the latest commit doesn't affect behavior, I won't update the test build.

code/graphics/post_processing.cpp
```diff
@@ -34,6 +35,31 @@ PostEffectUniformType mapUniformNameToType(const SCP_string& uniform_name)
...
+// used by In-Game Options menu
+bool Post_processing_enable_lightshafts = true;
+
+auto LightshaftsOption =
+	options::OptionBuilder<bool>("Graphics.Lightshafts", "Lightshafts", "Enable lightshafts (requires post-processing)")
+		.category("Graphics")
+		.default_val(true)
+		.level(options::ExpertLevel::Advanced)
+		.bind_to(&Post_processing_enable_lightshafts)
+		.importance(60)
```
jg18, line 48: The importance values are total guesses. I presume they define the order of options in the menu?

code/graphics/post_processing.cpp
```diff
@@ -34,6 +35,31 @@ PostEffectUniformType mapUniformNameToType(const SCP_string& uniform_name)
...
+		.finish();
+
+int Post_processing_bloom_intensity = 25; // using default value of Cmdline_bloom_intensity
+
+auto BloomIntensityOption = options::OptionBuilder<int>("Graphics.BloomIntensity",
+	"Bloom intensity",
+	"Set bloom intensity (requires post-processing)")
+	.category("Graphics")
+	.range(0, 200)
```
jg18, line 57: The [FS Wiki says](https://wiki.hard-light.net/index.php/Command-Line_Reference#-bloom_intensity) the range for `-bloom_intensity` is 0-200, but the call site for `gr_bloom_intensity()` in render_options.cpp makes me wonder if it's really 0-100.

jg18 (commented):
A couple things to call out for discussion.

code/graphics/post_processing.cpp
asarium: Yes, that is what they are used for in the existing UI. The actual values do not matter and are only compared between items in the same category.

asarium (approved):

PR #3504, opened 2021-06-15
maintain a copy of the default translations to check against untranslated strings

### LINKED ISSUE #3471: Bug: single mission not visible if title is translated
In non-english version, single missions with a title translated with tstrings.tbl/xxx-tlc.tbm are not visible in the tech room.

This bug does not affect campaign missions.

Goober5000:
Bug #3471 occurs because the game only keeps a translation table for the current language.  Since the mission list needs to compare untranslated strings against English strings, this adds a duplicate table for English.  However, it only keeps strings that are NAME_LENGTH or less, because the comparison only needs to be run against mission names.

Fixes #3471.

### DIFF
--- a/code/localization/localize.cpp
+++ b/code/localization/localize.cpp
@@ -462,8 +462,8 @@ void parse_tstringstbl(const char *filename)
 // initialize the xstr table
 void lcl_xstr_init()
 {
-	for (int i = 0; i < XSTR_SIZE; i++)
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):
LGTM, and good to know that tests with this build solved the issue for Novachen. Always good to have more translations fixes.

PR #3505, opened 2021-06-15
Add Keyboard Shortcut for `Save As` in FRED

wookieejedi:
Finally got around to adding in the `Ctrl+Shift+S` shortcut for Save As in FRED. Tested and works as expected.

EDIT: Thanks @Goober5000 for the typo notice! :D


Goober5000 (approved):
Looks good, and it's the MFC-standard way to do it, too. :+1:

PR #3506, opened 2021-06-15
some bitwise operation enhancements

Goober5000:
Allow AND and OR to take up to 10 arguments.  Deprecate addBit and replace with setBit, and add unsetBit.


wookieejedi:
Makes sense, set bit and unset bit are much more intuitive for FSO scripts, especially since  FRED has set and unset bits too. Given depreciation is involved, I also check Axem's scripts since many mods use those scripts, and BitOps is not used in any of them.

asarium (approved):

Issue #2689, opened 2020-08-27
FR: Nebula rendering enhancements & configurability


Durandal-Thoth:
As per discussion with The_E, Goober5000, & Axem -

- [ ] 1. Ability to fade nebula poof bitmaps in and out when toggled via sexp. Again, you can add or remove the bitmaps, but they immediately appear and disappear which is quite jarring when you're trying to do a natural transition.

- [x] 2. Ability to use an RGB color pick for nebula background instead of the presets. This one is probably less important, but would make everyone's lives easier and give people more freedom in color design for their nebula missions. Incidentally this would also circumvent the need for .pcx files as a color palette for the background.

- [x] 3. Ability to configure transparency, scale, rotation, density, velocity, etc of different nebula poofs individually via nebula.tbl. Right now, nebula textures need to be created a very certain way and match the background color in order to avoid looking too janky when fading in and out at the edges of the screen (which shouldn't even really be necessary). Ability to control these parameters per bitmap would allow a wider variety of art styles to be employed. As an example, [this](https://cdn.discordapp.com/attachments/433363226105741315/748326665670754354/unknown.png) may look great as a still image, but current rendering means these high quality poofs pop in and out of existence in an unrealistic looking fashion.

Baezon:
2. Partially implemented with a SEXP `nebula-change-fog-color` in #3319, actual UI support for a color picker should be a goal for qtFRED at least, though.

3. Implemented in #3507

Goober5000:
1. Implemented in #5612

PR #3507, opened 2021-06-16
Overhaul nebula poof system

Baezon:
The old system used a 5x5x5 cube, each cell of which was occupied by a randomly selected poof from those available. While I would learn during this process that this approach comes with many subtle advantages that were difficult to capture more generally, it handles differing poof types with different sizes and view ranges very poorly, so the system is refactored to instead handle each poof type one at a time randomly generating in a well-distributed spherical range and density specific to each poof type. 

As always, default values were chosen that mimic the old behavior very very closely if the new options are not taken advantage of. The poof types come with a name, so that poof types with the same bitmap but different spawn parameters can be differentiated, by default the name is the same as the bitmap.

Additionally, rendering has been sped up significantly by using the batching system, so taking advantage of the new parameters is very cheap.

Through this process, pretty much all of those `Neb2_detail` values no longer became applicable. As it is, the detail settings will now only scale down the total number of poofs, the main thing that has much of an effect on rendering speed.

Implements the last point of #2689


code/nebula/neb.cpp
```diff
@@ -348,6 +290,26 @@ void neb2_level_init()
...
+	for (size_t i = 0; i < Poof_info.size(); i++) {
+		if (poof_is_used(i)) {
+			Poof_density_sum_square += Poof_info[i].density * Poof_info[i].density;
+			Poof_density_sum += Poof_info[i].density;
+			float dist_threshold = Poof_info[i].view_dist * (UPKEEP_DIST_MULT - 1.0f);
+			if (dist_threshold < Poof_dist_threshold)
+				Poof_dist_threshold = dist_threshold;
+		}
+	}
+	Poof_density_multiplier = Poof_density_sum_square / (Poof_density_sum * Poof_density_sum);
```
BMagnu, line 309: This will divide by 0 if all poofs are disabled. In practice, this results in the value never being read, so no issue.
However, in theory, this is undefined behaviour in C++ and should, just to be safe, be prevented.

code/nebula/neb.cpp
```diff
@@ -689,253 +653,100 @@ float neb2_get_alpha_2shell(float inner_radius, float outer_radius, float magic_
...
+
+			rand_pos.xyz.x *= gen_side_length / 2;
+			rand_pos.xyz.y *= gen_side_length / 2;
+			rand_pos.xyz.z *= gen_side_length / 2;
+
+			pos += rand_pos;
+
+			// we generated poofs in a cube, now keep only those that are within the view sphere, and weren't within the last view sphere
+			// not terribly efficient but very simple
+			if (vm_vec_dist(&eye_pos, &pos) <= (gen_side_length / 2) &&
```
BMagnu, line 724: Could / should these poofs not also be culled by being in front of the camera? You cull these later during rendering if they are behind the camera, but then they wouldn't need to be generated here as well, right?

code/nebula/neb.cpp
```diff
@@ -689,253 +653,100 @@ float neb2_get_alpha_2shell(float inner_radius, float outer_radius, float magic_
...
-}
 
-/*
- * TODO: remove this
-float max_area = 100000000.0f;
-DCF(max_area, "")
-{
-	dc_stuff_float(&max_area);
+	// cull distant poofs
+	if (!Neb2_poofs.empty()) {
```
BMagnu, line 731: This should really happen before the new poofs are generated. Otherwise, this'll just needlessly iterate over more poofs (namely all that were just generated) which are guaranteed to never need to be culled, as their generating condition is the same for them to not be culled here.

code/nebula/neblightning.cpp
```diff
@@ -498,33 +498,15 @@ void nebl_process()
...
-					s1 = 4;
-					s2 = 0;
-				}
-				if((e1 == 2) && (e2 == 2) && (e3 == 2)){
-					e1 = 0;
-					e2 = 4;
-				}
-
-			// sanity
-			} while((s1 == e1) && (s2 == e2) && (s3 == e3));
```
BMagnu, line 524: This was previously preventing any bolts with 0 length.
While the `vm_vec_dist(&start, &Eye_position) > 200.0f` check does prevent no bolts starting or ending very close to the camera, it does not prevent a bolt that is very short. Should such a short-bolt-prevention be added here, or should these potentially very short bolts be called a feature now?

BMagnu (changes requested):
Looks pretty clean overall, just some minor things.
Finally adds lots of customization to the poofs.

code/nebula/neb.cpp
Baezon: The poof upkeep is only triggered by moving the camera, and so in order to do that I would need to trigger on looking around as well. Additionally, this maintains 'stability' to the specific poofs when looking around (i.e. you see the same ones in the same positions), in general I don't value this very much, especially if it's unlikely to be noticeable, but in the situation of very large distant poofs, they can be very obvious visual landmarks and failing to keep stability when looking around could be very noticeable.

code/nebula/neb.cpp
Baezon: Ah, true. (the latter used to not be the case in previous versions)

code/nebula/neb.cpp
BMagnu: Fair point as for looking around.
Then this is absolutely fine.

BMagnu (approved):
LGTM

PR #3508, opened 2021-06-16
add bitwise operations with enums

Goober5000:
This adds `EnumAND` and `EnumOR` which will return new enums based on the specified operations.  These are handy for enums that correspond to bit fields, such as those used for object persistent sounds.

### DIFF
--- a/code/scripting/api/objs/object.cpp
+++ b/code/scripting/api/objs/object.cpp
@@ -541,16 +541,8 @@ ADE_FUNC(assignSound, l_Object, "soundentry GameSnd, [vector Offset=nil, enumera
 	auto subsys = tgsh ? tgsh->ss : nullptr;
 	if (!offset)
 		offset = &vmd_zero_vector;
-	if (enum_flags.IsValid())
-	{
+	if (enum_flags.index >= 0)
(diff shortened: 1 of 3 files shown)


code/scripting/api/objs/object.cpp
```diff
@@ -541,17 +541,9 @@ ADE_FUNC(assignSound, l_Object, "soundentry GameSnd, [vector Offset=nil, enumera
 	auto subsys = tgsh ? tgsh->ss : nullptr;
 	if (!offset)
 		offset = &vmd_zero_vector;
-	if (enum_flags.IsValid())
-	{
+	if (enum_flags.index >= 0)
```
asarium, line 544: I can't think of a way to generate the "0" value for this. Maybe it would be good to also add a `OS_NONE` enum with the value 0 so that can be specified here if not flags are desired.

asarium (approved):
Nothing wrong with the code, just something that might be useful for users of the API.

code/scripting/api/objs/object.cpp
Goober5000: That's a good point.

PR #3510, opened 2021-06-17
Add More Options to Scripted Persistent Sounds

wookieejedi:
This adds two new enums to persistent sounds which can be utilized by scripts.
1. OS_PLAY_ON_PLAYER, which allows persistent sounds to play on the player's ship.
2. OS_LOOPING_DISABLED, which allows the scripter to set the sound to play once and not have to then call `removeSound` later. Very handy for playing an assigned sound once with one line.

Tested and everything works as expected.

### DIFF
--- a/code/object/objectsnd.cpp
+++ b/code/object/objectsnd.cpp
@@ -471,8 +471,9 @@ void obj_snd_do_frame()
 	for ( osp = GET_FIRST(&obj_snd_list); osp !=END_OF_LIST(&obj_snd_list); osp = GET_NEXT(osp) ) {
 		Assert(osp != NULL);
 		objp = &Objects[osp->objnum];
-		if ( Player_obj == objp && observer_obj == Player_obj ) {
+		if ((Player_obj == objp) && (observer_obj == Player_obj) && !(osp->flags & OS_PLAY_ON_PLAYER)) {
 			// we don't play the engine sound if the view is from the player
+			// unless OS_PLAY_ON_PLAYER was set manually
 			continue;
 		}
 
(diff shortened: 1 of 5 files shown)


code/scripting/api/objs/enums.cpp
```diff
@@ -116,6 +116,9 @@ flag_def_list Enumerations[] = {
 	{"OS_SUBSYS_DEAD", OS_SUBSYS_DEAD, 0},
 	{"OS_SUBSYS_DAMAGED", OS_SUBSYS_DAMAGED, 0},
 	{"OS_SUBSYS_ROTATION", OS_SUBSYS_ROTATION, 0},
+	{"OS_PLAY_ON_PLAYER", OS_PLAY_ON_PLAYER, 0},
+	{"OS_LOOPING_DISABLED", OS_LOOPING_DISABLED, 0},
+	{"OS_STARTED_PLAYING", OS_STARTED_PLAYING, 0},
```
Goober5000, line 121: I would remove `OS_STARTED_PLAYING` from this list since it's an internal flag.  Also remove `OS_USED` for the same reason, and remove `OS_DS3D` from both this list and the #define list since it's not used anywhere at all.

code/object/objectsnd.cpp
```diff
@@ -577,7 +578,8 @@ void obj_snd_do_frame()
 
 		go_ahead_flag = TRUE;
 		float max_vol,new_vol;
-		if (!osp->instance.isValid()) {
+		// start playing if sound is invalid and looping or non-looping but has not started
+		if (!osp->instance.isValid() && (!(osp->flags & OS_LOOPING_DISABLED) || !(osp->flags & OS_STARTED_PLAYING))) {
```
Goober5000, line 582: Upon further consideration, this if() block should be augmented a bit.  We want non-looping sounds that have already played once to be removed from the object sound list.  So I suggest something like this:
```
if (!osp->instance.isValid() && (osp->flags & OS_LOOPING_DISABLED) && (osp->flags & OS_STARTED_PLAYING)) {
    list_remove(&obj_snd_list, osp);
else if  (!osp->instance.isValid()) {
```
Not 100% sure that's correct, but I think it's close.

Goober5000 (changes requested):
Looks good, but a couple things to say...

code/object/objectsnd.cpp
Goober5000: This was not only insufficient, it was in the wrong place.  Ended up rewriting this a bit.

Goober5000 (approved):
Changes addressed

wookieejedi:
All of my tests work as expected too, so will merge this at the end of the day if there are no objections.

PR #3514, opened 2021-06-18
Properly handle death due to negative hull repair rate

Baezon:
Negative `$Hull Repair Rate` is explicitly allowed, so I do think it should be supported, even if not commonly used. If the hull strength drops below 0, make sure to actually kill the ship rather let it persist in that weird negative health zombie state.


Goober5000 (approved):
:+1:

PR #3520, opened 2021-06-19
Restore original error handling of stuff_int_list

Baezon:
https://github.com/scp-fs2open/fs2open.github.com/blame/ed2cc47093f2a1b2f8bd89c54784ad5a2de49179/code/parse/parselo.cpp#L3014

Attempts to restore the original error handling of `stuff_int_list` for `WEAPON_LIST_TYPE`, `WEAPON_POOL_TYPE`, and `SHIP_INFO_TYPE`. Where `WEAPON_LIST_TYPE` would merely warn and stuff -1, and the other two would stuff nothing and produce no warning or error. This changes it to warnings in all cases but only stuffing as it did before. Erroring instead of warning is a very harsh backwards compatibility break which people have been running into, and I would like to ease if possible.

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -2901,44 +2901,40 @@ size_t stuff_int_list(int *ilp, size_t max_ints, int lookup_type)
 				case SHIP_TYPE:
 					num = ship_name_lookup(str.c_str());	// returns index of Ship[] entry with name
 					if (num < 0)
-						error_display(1, "Unable to find ship %s in stuff_int_list!", str.c_str());
-
-					*buf = num;
+						error_display(0, "Unable to find ship %s in stuff_int_list!", str.c_str());
 					break;
 
 				case SHIP_INFO_TYPE:
 					num = ship_info_lookup(str.c_str());	// returns index of Ship_info[] entry with name
 					if (num < 0)
 						error_display(0, "Unable to find ship class %s in stuff_int_list!", str.c_str());
-					else
-						*buf = num;
 					break;
 
 				case WEAPON_POOL_TYPE:
 					num = weapon_info_lookup(str.c_str());
 					if (num < 0)
 						error_display(0, "Unable to find weapon class %s in stuff_int_list!", str.c_str());
-					else
-						*buf = num;
 					break;
 
 				case WEAPON_LIST_TYPE:
 					num = weapon_info_lookup(str.c_str());
 					if (num < 0 && !str.empty())
 						error_display(0, "Unable to find weapon class %s in stuff_int_list!", str.c_str());
-
-					*buf = num;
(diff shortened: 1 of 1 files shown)


code/parse/parselo.cpp
```diff
@@ -2902,36 +2902,43 @@ size_t stuff_int_list(int *ilp, size_t max_ints, int lookup_type)
 					num = ship_name_lookup(str.c_str());	// returns index of Ship[] entry with name
 					if (num < 0)
 						error_display(1, "Unable to find ship %s in stuff_int_list!", str.c_str());
```
Goober5000, line 2904: This will still display an Error.  Is that what `SHIP_TYPE` did before?

Goober5000 (commented):
Good change.  Two questions:

1) see other review comment
2) I would like to be consistent in the token handling.  Is it possible to make all the token types either stuff -1 or stuff nothing?

code/parse/parselo.cpp
Baezon: It is, yeah. It would've hard errored in that case before as well.

Baezon:
This function is used so often its hard to say whether making it consistent about what to stuff would be a good idea or not. My main priority was trying to encapsulate whatever it was doing before, which I can only speculate had good reasons for sometimes stuffing nothing and other times stuffing -1. When downgrading the severity from Error to Warning, one would expect some kind of reassurance that it is *actually* possible to proceed.

Of the two, I think stuffing nothing would probably be the safer option.

code/parse/parselo.cpp
Goober5000: Since I'm a fan of consistency, would there be any problem with making this a Warning to match the others?

Baezon: Fine by me.

Goober5000:
I would say that making it consistent *is* a good idea, because this is at the level of detail of the general contract of the function, not at the level of detail of implementation.  In other words, the programmer shouldn't need to remember which particular type handles things one way and which handles things the other way.

The odds are pretty good that there's no rhyme or reason to whether the data was stuffed or not.  All the functions were coded up separately, and there was little code reuse among the different types.  They behaved similarly, but the code showed signs of being written mostly from scratch each time.


> Of the two, I think stuffing nothing would probably be the safer option.

I think I agree.  Data types should be initialized to sensible defaults.  And, in fact, I realized that this is the convention for table files anyway - to stuff nothing because a previous tbm or tbl might have already handled it.

Baezon:
Very good points, this function should probably be as simple and straightforward as possible, and the callers can figure out any special handling they want to do if things didn't turn out the way they expected.

Goober5000 (approved):
:+1:

PR #3521, opened 2021-06-19
Bugfix type 5 shooter scale offset

Baezon:
Shooter scale offsets (that is, not target scale) should mean the offset is centered on the target but scaled to a constant angular span from the point of view of the shooter. Using just 300 meters centered on the target still has a differing angular span depending on its distance, it should directly use the distance, this way the offset scale basis is constant no matter its position.

Also should use `turret_point` like the rest of this function, `last_start` isn't necessarily valid yet.


wookieejedi (approved):
Good catch.

EatThePath:
btw I've done some playtesting with this and it does seem to work as desired without visible side effects.

PR #3526, opened 2021-06-21
Allow non-unique sexp subcategories

BMagnu:
Previously, luasexps could not be placed into subcategories that were not unique.
Not only did this include custom categories, but also built-in ones, so sexps couldn't be placed into categories like "Weapons, Shields, and Engines", as that exists both in status and in change.
This does away with this requirement (tbh, I am not sure why it was there in the first place).
If you specify a built in subcategory but it only exists in a different main category, this will now create a new subcategory in the main category specified, as I would expect from the luasexps

### DIFF
--- a/code/parse/sexp/LuaSEXP.cpp
+++ b/code/parse/sexp/LuaSEXP.cpp
@@ -77,9 +77,9 @@ int get_category(const SCP_string& name) {
 	return -1;
 }
 
-int get_subcategory(const SCP_string& name) {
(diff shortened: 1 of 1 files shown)


Goober5000:
I don't see anything wrong with this, but since @asarium wrote this code I'd like his input.

asarium (approved):
I remember having a good reason for not allowing that but maybe I just didn't understand the code properly. If FRED accepts it then it is good enough for me.

PR #3528, opened 2021-06-21
add is-language

Goober5000:
Checks whether the game is running in the specified language.  Since the fiction viewer doesn't use XSTR-style localization, this SEXP can be used to display different stages for different languages.


JohnAFernandez (approved):
LGTM

PR #3529, opened 2021-06-22
Add shield regen hit delay

Baezon:
Adds an optional delay after hitting before shields will start to recharge.

In order for this to work on players `last_hit_time` needs to be properly updated for them too, and wasn't before, which should also make the behavior of player ships under AI control a bit more consistent.


wookieejedi:
Good idea, this will be very useful.

wookieejedi (approved):
Approving so it can be in before RC phase starts.

PR #3531, opened 2021-06-23
Fix model point shield handling for mini target shields

Baezon:
This causes the little mini target hull/shield indicator just below the reticle to be handled like the regular shield indicator is, only use `Quadrant_xlate` if it has default shields, for model point shields simply use the .ani frames in the direct shield quadrant order. Importantly this will prevent single quadrant shields from indexing out of bounds (because it would use `Quadrant_xlate` to index the 1st shield quadrant for the 0th ani frame)

### DIFF
--- a/code/hud/hudshield.cpp
+++ b/code/hud/hudshield.cpp
@@ -378,7 +378,13 @@ void hud_shield_show_mini(object *objp, int x_force, int y_force, int x_hull_off
 			break;
 		}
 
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):

Baezon:
Blech, there's a few related things that need to be fixed too.

Baezon:
Ok, little confusing that the old hud stuff is still in here so things have to be done twice, but definitely works now on old ships and model point shield ships.

JohnAFernandez (approved):

PR #3535, opened 2021-06-24
Inline matrix multiplication ops for performance improvements

PhantomHoover:
Asteroth suggested I use the unit tests to profile the new matrix ops. Multiplication was slow as hell. I added inlining. The artificial unit test benchmark became fast! The total binary size seems the same. What could possibly go wrong?


JohnAFernandez (approved):
The file sizes being the same means there shouldn't be a downside...

PR #3538, opened 2021-06-26
Be smarter about when we try to run scripting hooks

PhantomHoover:
Don't even bother trying if no scripts are even registered to run on them.

### DIFF
--- a/code/scripting/scripting.cpp
+++ b/code/scripting/scripting.cpp
@@ -163,6 +163,8 @@ void script_parse_table(const char* filename)
 			while (st->ParseCondition(filename));
 			required_string("#End");
 		}
+
+		st->AssayActions();
 	}
 	catch (const parse::ParseException& e)
 	{
@@ -756,6 +758,8 @@ void script_state::Clear()
 	ConditionalHooks.clear();
 	HookVariableValues.clear();
 
+	AssayActions();
+
 	if (LuaState != nullptr) {
 		OnStateDestroy(LuaState);
 
@@ -1137,9 +1141,33 @@ bool script_state::ParseCondition(const char *filename)
 	return true;
 }
 
-void script_state::AddConditionedHook(ConditionedHook hook) { ConditionalHooks.push_back(std::move(hook)); }
+void script_state::AddConditionedHook(ConditionedHook hook) {
+	ConditionalHooks.push_back(std::move(hook));
+	AssayActions();
+}
 
 void script_state::AddGameInitFunction(script_function func) { GameInitFunctions.push_back(std::move(func)); }
+
+// For each possible script_action this maintains an array that records whether any scripts are actually using this action
+// This allows us to avoid significant overhead from checking everything at the potential hook sites, but you must call
+// AssayActions() after modifying ConditionalHooks before returning to normal operation of the scripting system!
+void script_state::AssayActions() {
+	for (bool &i : ActiveActions) {
+		i = false;
+	}
+
+	for (const auto &hook : ConditionalHooks) {
+		for (const auto &action : hook.Actions) {
+			if (action.action_type >= -1 && action.action_type <= CHA_LAST) {
+				ActiveActions[action.action_type] = true;
+			}
+		}
+	}
+}
+
+bool script_state::IsActiveAction(ConditionalActions action_id) {
+	return ActiveActions[action_id];
+}
 
 //*************************CLASS: script_state*************************
 bool script_state::IsOverride(script_hook &hd)
--- a/code/scripting/scripting.h
+++ b/code/scripting/scripting.h
@@ -152,10 +152,9 @@ struct script_action
 
 class ConditionedHook
 {
-private:
+public:
 	SCP_vector<script_action> Actions;
 	SCP_vector<script_condition> Conditions;
-public:
 	bool AddCondition(script_condition *sc);
 	bool AddAction(script_action *sa);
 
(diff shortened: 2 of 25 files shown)


wookieejedi:
Works as expected in tests with script-heavy mods such as FotG. Also, seems like it's 5-8 FPS faster in FotG's internal heavy stress test missions.

PhantomHoover:
"But Hoover", I hear you say, "how can you be sure these checks aren't going to affect existing behaviour!" Good question; I believe these two checks:
https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/scripting/scripting.cpp#L607
https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/scripting/scripting.cpp#L622

mean that `ConditionedHook::IsOverride` and `ConditionedHook::Run` will return before doing anything important if the `action` passed to them doesn't match at least one `ConditionedHook`'s `action_type`. If this is the case the corresponding entry in `ActiveActions` should be set to true and so everything will be executed as before, and otherwise no scripts or other observable behaviour could've run anyway.

JohnAFernandez:
Looking at the changes, it's mostly simple changes, even though it's in a lot of files.

PhantomHoover:
Oh, also, for posterity, here's the averaged framerates of 10 runs each of Icarus with and without this change, on my system:
plot

z64555 (changes requested):
LGTM, could maybe double check the bracket/bracing style if you want to be thorough.

I do have an objection to seeding the RNG with a static number in freespace.cpp:1600, however.  I do agree that there should be an option to use a static or user defined seed for benchmarking, but I don't quite agree to having it as a default option, and this tidbit is logically different from the rest of the PR.

code/scripting/scripting.h
```diff
@@ -152,10 +152,9 @@ struct script_action
 
 class ConditionedHook
 {
-private:
+public:
```
asarium, line 155: Why is this necessary to be public now? I don't see a reason in the code to change that.

code/scripting/scripting.h
```diff
@@ -243,6 +242,10 @@ class script_state
 	void ParseGlobalChunk(ConditionalActions hookType, const char* debug_str=nullptr);
 	bool ParseCondition(const char *filename="<Unknown>");
 	void AddConditionedHook(ConditionedHook hook);
+	// ActiveActions lets code that might run scripting hooks know whether any scripts are even registered for it.
+	// AssayActions is responsible for keeping it up to date.
+	bool ActiveActions[ConditionalActions::CHA_LAST+1];
```
asarium, line 247: The check if there is an active action should be put into a function which can then use this data structure internally. That will make it easier to refactor the internals of the scripting system later if that is necessary.

freespace2/freespace.cpp
```diff
@@ -1596,8 +1596,8 @@ void game_init()
 
 	Game_current_mission_filename[0] = 0;
 
-	// seed the random number generator
-	Random::seed(static_cast<unsigned int>(time(nullptr)));
+	// seed the random number generator, using a static seed if benchmarking to encourage consistency
+	Random::seed(Cmdline_benchmark_mode ? 2335 : static_cast<unsigned int>(time(nullptr)));
```
asarium, line 1600: I agree with @z64555, this should be in its own PR since it has nothing to do with the scripting changes.

asarium (changes requested):
Seems to be a sensible change, just some style things I noticed.

code/scripting/scripting.h
PhantomHoover: Because AssayActions needs to be able to iterate over the Actions vector, but it isn't a ConditionedHook member function. Conditions doesn't strictly need to be private but I felt it was cleaner just to take the honest route and make the whole thing public.

freespace2/freespace.cpp
PhantomHoover: I've removed this. In all honesty it wasn't even producing more consistent benchmarks.

code/scripting/scripting.h
PhantomHoover: I've refactored everything to do it this way.

z64555 (approved):

code/scripting/scripting.h
asarium: Ok, that is a valid point. In that case, please just make it a `struct` and remove the `public` altogether.

MageKing17: The general convention throughout the codebase is that, if it has methods, it's a `class`.

code/scripting/scripting.cpp
```diff
@@ -1137,10 +1141,34 @@ bool script_state::ParseCondition(const char *filename)
...
+	}
+
+	for (const auto &hook : ConditionalHooks) {
+		for (const auto &action : hook.Actions) {
+			if (action.action_type >= -1 && action.action_type <= CHA_LAST) {
+				ActiveActions[action.action_type] = true;
+			}
+		}
+	}
+}
```
Goober5000, line 1166: Why does this function clear the entire array and then re-populate it whenever it's called?  Why not initialize the array when the script system is initialized, and then just toggle an action to true when a hook is added?

EatThePath: seems like this would be fine if this only gets called once, after all the script tables have been loaded? I'm failing right now to imagine  a case where this needs to be done more than once.

PhantomHoover: It also had to be run in AddConditionedHook:

https://github.com/scp-fs2open/fs2open.github.com/pull/3538/files#diff-8763caafc296d3dd73ccf6d72b40529289c3908b57fa9f06da1fc26ac840f547R1146

I also added it to the Clear method on script_system, but that was probably overkill as I think that's one of those dumb cleanup functions we have that only actually gets called when the game is about to exit.

As for why I did it this way, I just found it easier to write a function that would guarantee the active action lookup was accurate and then drop it in between any code that might add new conditional hooks and its return to the scripting system, versus fiddling around with various individual sites that might add hooks. It may look a little inefficient but consider that script_state::RunCondition iterates over every single conditional hook and does even more stuff than this hundreds or thousands of times per frame. AssayActions is probably run a few dozen times on engine startup at most.

code/scripting/scripting.h
asarium: Alright I guess.

asarium (approved):

PR #3540, opened 2021-06-27
Remove Buggy Function and Let Debris End Naturally

JohnAFernandez:
Removes this retail function that was actually perpetually postponing the death of debris at long distances instead of removing them.  Debris lifetime is suitably short for the small, non-hull pieces (only a few seconds), while large pieces are supposed to be permanent, since they can be visible even at longer ranges.   This function would also remove debris based on player position, which is not the correct approach, since the camera can be free from the player ship.

Additionally, there's no real reason to try to remove debris to save on performance, since even in a busy mission like Icarus, once this function is removed, there are at most 60 pieces of debris in the mission.

### DIFF
--- a/code/debris/debris.cpp
+++ b/code/debris/debris.cpp
@@ -197,34 +197,6 @@ void debris_delete( object * obj )
 	db->objnum = -1;
 }
 
-/**
- * If debris piece *db is far away from all players, make it go away very soon.
- * In single player game, delete if MAX_DEBRIS_DIST from player.
- * In multiplayer game, delete if MAX_DEBRIS_DIST from all players.
- */
-void maybe_delete_debris(debris *db)
-{
-	object	*objp;
-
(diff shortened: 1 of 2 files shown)


PhantomHoover:
Yeah, uh, I think I can substantiate the performance improvements from this:

mdplot

wookieejedi (approved):
Good catch.

PhantomHoover:
I would note that after removing this function the following code is redundant and should also be removed:

`MAX_DEBRIS_DIST` (debris.cpp:L55)
`DEBRIS_DISTANCE_CHECK_TIME` (debris.cpp:L56)
`debris.next_distance_check` (debris.h:L53)
`db->next_distance_check = Random::next(2000) + 4*DEBRIS_DISTANCE_CHECK_TIME;` (debris.cpp:L513)

JohnAFernandez:
You're absolutely right.  Please don't merge this until I can remove that.

PR #3552, opened 2021-07-09
Fix a secondary range bug from AI multilock

Baezon:
A bit of fallout from AI multilock, this was erroneously changed from `target_world_pos` (which had just been calculated for this purpose) to `lock_slot->obj->pos`, preventing locks on subsystems if the player is is out of range of the ship's center.


Goober5000 (approved):

Goober5000:
Looks like this happened in #3183

PR #3553, opened 2021-07-11
fix campaign loop in certain cases

Goober5000:
If the loop mission and the next mission are the same, then don't show the campaign loop screen.  This prevents the campaign from getting into a campaign loop state if it shouldn't, and also allows campaign loops to be conditionally accessible.

Tested and works as expected.


wookieejedi (approved):
Good catch.

PR #3554, opened 2021-07-11
FRED turret info

Goober5000:
Add some useful QoL improvements for turrets.
1) Each turret now not only displays its name, but also whatever weapons it is carrying.
2) A new menu option allows displaying info from all turrets at once.

Tested and works as expected.  See screenshot:

all_turrets


wookieejedi (approved):
Changes look good, and in testing it worked well, too. Very handy feature. Since it is an enhancement I would imagine merging will wait until after stable?

Goober5000:
Perhaps.  I'm not 100% sure we're actually in a code freeze ATM.  I'll ask DahBlount.

Goober5000:
Ok, we're not officially in a code freeze yet.  Merging.

PR #3555, opened 2021-07-13
Add "num firepoints for burst shots" flag

Baezon:
This burst flag essentially sets burst shots to the number of firepoints of whatever is firing the weapon at the moment. This is useful for mods like FotG, where multiple turrets all use the same weapon, but have differing numbers of firepoints, where the burst length should correspond to it, and circumvents the cumbersome use of many weapon copies.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1605,8 +1605,9 @@ void turret_set_next_fire_timestamp(int weapon_num, weapon_info *wip, ship_subsy
 {
 	Assert(weapon_num < MAX_SHIP_WEAPONS);
 	float wait = 1000.0f;
+	int burst_shots = (wip->burst_flags[Weapon::Burst_Flags::Num_firepoints_burst_shots] ? turret->system_info->turret_num_firing_points - 1 : wip->burst_shots);
 
-	if (wip->burst_shots > turret->weapons.burst_counter[weapon_num]) {
+	if (burst_shots > turret->weapons.burst_counter[weapon_num]) {
 		wait *= wip->burst_delay;
 		turret->weapons.burst_counter[weapon_num]++;
 	} else {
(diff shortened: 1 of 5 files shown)


wookieejedi (approved):
Very useful feature indeed. Looks good, too.

code/ai/aiturret.cpp
```diff
@@ -1605,8 +1605,9 @@ void turret_set_next_fire_timestamp(int weapon_num, weapon_info *wip, ship_subsy
 {
 	Assert(weapon_num < MAX_SHIP_WEAPONS);
 	float wait = 1000.0f;
+	int burst_shots = (wip->burst_flags[Weapon::Burst_Flags::Num_firepoints_burst_shots] ? turret->system_info->turret_num_firing_points - 1 : wip->burst_shots);
```
Goober5000, line 1608: Why is this the firing points - 1?

code/ship/ship.cpp
```diff
@@ -11496,7 +11500,8 @@ int ship_fire_primary(object * obj, int stream_weapons, int force, bool rollback
 		int old_burst_counter = swp->burst_counter[bank_to_fire];
 		int old_burst_seed = swp->burst_seed[bank_to_fire];
 
-		if (winfo_p->burst_shots > swp->burst_counter[bank_to_fire]) {
+		int burst_shots = (winfo_p->burst_flags[Weapon::Burst_Flags::Num_firepoints_burst_shots] ? num_slots - 1 : winfo_p->burst_shots);
```
Goober5000, line 11503: Here too with slots - 1

code/ai/aiturret.cpp
Baezon: This is to match the behavior of how burst shots is handled. If the number of burst shots is greater than the burst counter (which is incremented on each burst shot and reset when the last shot in the burst is fired) that means this shot is a burst shot, if they're equal, this is the end of a burst and it should use firewait. So 0 burst shots is as normal, every shot is firewait, 1, means a two shot burst, every other shot gets burst delay or firewait, and so on.

Essentially, for the same reason that `$Burst Shots` is decremented by 1 when it is parsed. A way of thinking about it is that the internally used `burst_shots` value is always "how many burst delay shots happen between each firewait shot"

Goober5000: Ah ok.

Does it make sense to add a comment essentially saying "this is intentional; see parsing" or similar?

Baezon: `wip->burst_shots` doesn't mention this, so yeah, I'll add a comment clarifying how that works, and refer these spots to that comment

Goober5000 (approved):

PR #3556, opened 2021-07-13
Add setVolume function for audio_stream objects

Baezon:
Just like `setVolume` for regular `sound` handles

### DIFF
--- a/code/scripting/api/objs/audio_stream.cpp
+++ b/code/scripting/api/objs/audio_stream.cpp
@@ -132,6 +132,21 @@ ADE_FUNC(isPlaying,
 	return ade_set_args(L, "b", audiostream_is_playing(streamHandle) != 0);
 }
 
+ADE_FUNC(setVolume,	l_AudioStream, "number volume", "Sets the volume of the audio stream, 0 - 1", "boolean", "true on success, false otherwise")
+{
+	int streamHandle = -1;
+	float volume = -1.0f;
+	if (!ade_get_args(L, "of", l_AudioStream.Get(&streamHandle), &volume)) {
+		return ADE_RETURN_FALSE;
+	}
+
+	if (streamHandle < 0) {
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/audio_stream.cpp
```diff
@@ -132,6 +132,21 @@ ADE_FUNC(isPlaying,
...
+	if (!ade_get_args(L, "of", l_AudioStream.Get(&streamHandle), &volume)) {
+		return ADE_RETURN_FALSE;
+	}
+
+	if (streamHandle < 0) {
+		return ade_set_args(L, "b", false);
+	}
+	audiostream_set_volume(streamHandle, volume);
+	return ade_set_args(L, "b", true);
+}
```
Goober5000, line 148: Why didn't you use `ADE_RETURN_FALSE` and `ADE_RETURN_TRUE` here?  Especially since you used `ADE_RETURN_FALSE` a few lines up.

Baezon: Good point, I mostly copy-pasted this from another function, but now I'm kind of confused. How is `ADE_RETURN_FALSE` different from `ade_set_args(L, "b", false)`? They're both just returning false right?

Goober5000: They're literally the same thing:

```
#define ADE_RETURN_TRUE				ade_set_args(L, "b", true)

#define ADE_RETURN_FALSE			ade_set_args(L, "b", false)
```

Baezon: OK, they should definitely be the macro then.

Baezon: I'm seeing the explicit version used quite a bit elsewhere in general, I might make a separate PR to switch those to the macro since that's more readable.

Goober5000 (approved):

Issue #3341, opened 2021-03-28
Sync recent FRED changes to QtFRED


Goober5000:
Many of the recent changes to the FRED code affect the UI and don't have analogous implementations in QtFRED.  These include event annotations and new flags on the ship and mission dialogs.  It will be necessary to go through the recent commits to the fred/ directory and make sure all of those commits (both bugfixes and features) have been ported over.

MjnMixael:
This is nearly 4 years old and doesn't specify exactly what changes are being referred to. This ticket is just clutter at this point as it will be more useful to compare FRED to QtFRED dialog by dialog than do code archeology to figure out what this ticket was in reference to. This ticket should be closed.

MjnMixael:
I am once again closing this issue and this time I have receipts. I had Claude do an audit of all commits to the FRED and QtFRED directories since January of 2020, a full year predating the creation of this issue. It churned for almost an hour, btw. The findings are conclusive that there are far too many commits (720+ to FRED alone) to both directories for a human to reasonably go through them all in any sort of timely manner. Additionally the QtFRED code has diverged so greatly that 1:1 comparisons are not straightforward, requiring that I manually checked many things that Claude thought might have been missed. After lots of manual checking and wasting of time, Claude finds no meaningful gaps.

-----

# FRED → QtFRED Sync Audit

## Context

The user requested a complete audit of all commits touching `fred2/` since January 2020, cross-referenced against `qtfred/`, to identify bug fixes or features that landed in FRED2 but were never ported to QtFRED.

**Methodology note**: The local repository is a shallow clone with only ~280 commits going back to February 2025. The full history was audited via the GitHub API (`mcp__github__list_commits` with `path=fred2`, `since=2020-01-01`, 8 pages × 100 commits = 720+ commits examined). Each candidate commit was inspected via `mcp__github__get_commit` — commits touching both `fred2/` AND `qtfred/` counted as already-ported; commits touching only `fred2/` were individually verified in the QtFRED source.

---

## Manually Checked by User

| Commit / Feature | Reason for Exclusion |
|------------------|----------------------|
| Sound environment dropdown style | QtFRED was never broken |
| Adjust Grid Dialog | Migrated to Preferences |
| Hide Marked Objects / Show Hidden Objects | Replaced by Layer system |
| Double Fine / Anti-Aliased Gridlines | Dropped intentionally |
| Various Settings Menu Items | Migrated to Preferences |
[...]

PR #3558, opened 2021-07-16
Qtfred menu update

Goober5000:
There are several updates to the menu in FRED that hadn't yet been ported over to qtFRED.  This PR ports over the changes to the menu and toolbars, viz. a few added menu options, a menu reorganization, some caption changes, and some shortcut key changes.

This partially addresses #3341.


JohnAFernandez (approved):
*shrugs* I don't see anything wrong.  Need to keep making progress on qtFred, so if no one else sees anything, we should get this merged.

PR #3562, opened 2021-07-18
Force portable if unicode

### LINKED ISSUE #3173: Cannot launch with a unicode Windows username
If the application is launched and the user folder contains unicode, they are presented with this error:
image
This is due to poor logic in getPreferencesPath() within [osapi.cpp](https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/osapi/osapi.cpp). It correctly handles SDL_GetPrefPath failing by offering to use the current directory, but does not do the same if the unicode check fails. I do not see any reference to the "portable mode" mentioned in the error.

If "portable mode" exists, it needs to be checked for. Otherwise, it should check if the current directory also contains unicode before failing.

JohnAFernandez:
This allows FSO to assume that it's in portable mode if it detects Unicode characters in the path provided by SDL.  I tested this by creating a local account with Unicode characters.  It not only runs as expected, but also keeps Knossos from getting stuck on the executable when trying to parse the command line.  

Fixes #3173

### DIFF
--- a/code/osapi/osapi.cpp
+++ b/code/osapi/osapi.cpp
@@ -54,7 +54,8 @@ namespace
 						"Report at www.hard-light.net or the hard-light discord.", SDL_GetError());
 					sdl_is_borked_warning = true;
 				}
-				// No preferences path, try current directory.  
+				// No preferences path, try current directory.
+				Cmdline_portable_mode = true;
 				return "." DIR_SEPARATOR_STR;
 		    }
 #ifdef WIN32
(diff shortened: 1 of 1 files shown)


code/osapi/osapi.cpp
```diff
@@ -23,6 +23,8 @@
 #include <sys/types.h>
 #endif
 
+bool Force_Portable_Mode = false;
```
JohnAFernandez, line 26: Btw, not sure if this is the best place to put this because I had to use extern bool and therefore had to put it outside the anonymous namespace.

Goober5000 (commented):
This looks good, but why not set the `Cmdline_portable_mode` flag instead?

JohnAFernandez:
I thought it was better to keep it separate, but that's only based on a guess for best practices.  I can go either way.

Goober5000 (approved):
I would use a single flag.  The reason is that every place where portable mode needs to be handled, the engine now needs to check two flags instead of one.

You already added the `force_portable_warning` to indicate whether the forcing actually took place.  If we need to remember the context we can give that variable a wider scope.

Goober5000 (changes requested):
(hit the wrong option)

JohnAFernandez:
> I would use a single flag. The reason is that every place where portable mode needs to be handled, the engine now needs to check two flags instead of one.
> 
> You already added the `force_portable_warning` to indicate whether the forcing actually took place. If we need to remember the context we can give that variable a wider scope.

Ok.  That shouldn't be hard to change.

JohnAFernandez:
Changed.

code/osapi/osapi.cpp
```diff
@@ -64,15 +65,21 @@ namespace
...
-					          "contains a Unicode character (%s). If possible, choose a different user name or "
-					          "use portable mode.",
-					          preferencesPath, std::string(current, invalid_end).c_str());
+						static bool force_portable_warning = false;
+						if (!force_portable_warning) {
+							utf8::prior(current, preferencesPath);
+							ReleaseWarning(LOCATION,
+								"Determined the preferences path as \"%s\". That path is not supported since it "
+								"contains a Unicode character (%s). Using portable mode. Set -portable_mode in "
+								"the commandline to avoid in the future",
```
Goober5000, line 77: To be unambiguous, say "to avoid this message", or something similar.

Goober5000 (approved):
:+1:

ngld:
I think this will cause FSO to ignore all `fs2_open.ini` settings since Knossos doesn't support portable mode and will still place that file in the user's `AppData` folder which FSO can't access. FSO ignoring those settings is still better than crashing on launch, though.

PR #3564, opened 2021-07-18
add Draw Outlines on Selected Ships option

Goober5000:
This option is helpful because the mesh outlines will often obscure important information like submodel names and turret weapons.


wookieejedi (approved):
Good idea, and LGTM.

JohnAFernandez (approved):
Yeah, despite how many files are changed it is a pretty simple pr.

PR #3565, opened 2021-07-21
fix briefing icons changing to ships unexpectedly

Goober5000:
If the briefing icon doesn't actually show a ship, then don't use the ship icon override.

Also remove a redundant assignment.

### DIFF
--- a/code/mission/missionbriefcommon.cpp
+++ b/code/mission/missionbriefcommon.cpp
@@ -649,8 +649,21 @@ briefing_icon_info *brief_get_icon_info(brief_icon *bi)
 		return NULL;
 	ship_info *sip = &Ship_info[bi->ship_class];
 
+	bool allow_override = true;
+	switch (bi->type)
+	{
+		case ICON_PLANET:
+		case ICON_ASTEROID_FIELD:
+		case ICON_WAYPOINT:
+		case ICON_UNKNOWN:
+		case ICON_UNKNOWN_WING:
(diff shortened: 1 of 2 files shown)


Baezon:
If the icon has a ship class, why would you not use the override? Isn't that its point? What is the unexpected scenario here?

Goober5000:
Because if the icon shows something other than a ship, the mission designer wanted to show that other thing.

All briefing icons have ship classes regardless of whether they actually show ships.  The ship class override is useful for separating the cruiser icon into the Fenris and Aeolus, for example.  But if you have a planet icon with a Fenris as the underlying ship class, then that icon shouldn't be replaced.

See 1:46:06 here: https://www.youtube.com/watch?v=h9Yub-hP6QM

Baezon (approved):
On the one hand, this seems like a poor choice of underlying ship for this icon, given its icon is overridden, I think its less intuitive to silently 'override the override', but on the other, I respect that that wasn't the order of operations (the ship-icon came first and the override later) and this can't really be solved otherwise and there's no real situation where someone would choose an icon for a ship and want it to be anything other than the specific icon they chose, override or not. Unless of course if it were something like "cruiser wing" where they'd want the override wing instead, which comes back to this only applying if the chosen icon was a ship, so fair enough.

Goober5000:
There really isn't any good choice of underlying ship.  FRED always uses one, but there is no "planet" ship in ships.tbl.

PR #3567, opened 2021-07-22
Fix a nullptr bug with change_ship_type

Baezon:
Always ensure that `p_objp` is not null before accessing it, SEXP or script created ships will have no parse object. There is only one other place where `p_objp` is accessed https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/ship/ship.cpp#L10855, but this apparently will only happen in the loadout screen so it should be safe.

As for texture replacements, (as always the diff is confusing but I just wrapped it in p_objp != nullptr) as the comment states, the new positions may be different so we can't import the old ones and if this ship also doesn't have a parse object then it's out of luck and will need to have texture-replace reran on it.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -10880,9 +10880,10 @@ void change_ship_type(int n, int ship_type, int by_sexp)
 		for (i = 0; i < MAX_REPLACEMENT_TEXTURES; i++)
 			sp->ship_replacement_textures[i] = -1;
 
-		if (p_objp != nullptr) {
+		if (p_objp != nullptr) 
+		{
 			// now fill them in according to texture name
-			for (SCP_vector<texture_replace>::iterator tr = p_objp->replacement_textures.begin(); tr != p_objp->replacement_textures.end(); ++tr)
+			for (const auto &tr : p_objp->replacement_textures)
 			{
 				int j;
 				polymodel* pm = model_get(sip->model_num);
@@ -10892,9 +10893,9 @@ void change_ship_type(int n, int ship_type, int by_sexp)
 				{
 					texture_map* tmap = &pm->maps[j];
 
-					int tnum = tmap->FindTexture(tr->old_texture);
+					int tnum = tmap->FindTexture(tr.old_texture);
 					if (tnum > -1)
-						sp->ship_replacement_textures[j * TM_NUM_TYPES + tnum] = tr->new_texture_id;
+						sp->ship_replacement_textures[j * TM_NUM_TYPES + tnum] = tr.new_texture_id;
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


code/ship/ship.cpp
```diff
@@ -10874,26 +10874,28 @@ void change_ship_type(int n, int ship_type, int by_sexp)
...
-		for (SCP_vector<texture_replace>::iterator tr = p_objp->replacement_textures.begin(); tr != p_objp->replacement_textures.end(); ++tr)
-		{
-			int j;
-			polymodel *pm = model_get(sip->model_num);
-
-			// look for textures
-			for (j = 0; j < pm->n_textures; j++)
+		if (p_objp != nullptr) {
+			// now fill them in according to texture name
+			for (auto tr : p_objp->replacement_textures)
```
Goober5000, line 10885: This should be changed to `auto &tr` and the `tr->` uses should be changed to `tr.`.

code/ship/ship.cpp
```diff
@@ -10874,26 +10874,28 @@ void change_ship_type(int n, int ship_type, int by_sexp)
...
 
-		// now fill them in according to texture name
-		for (SCP_vector<texture_replace>::iterator tr = p_objp->replacement_textures.begin(); tr != p_objp->replacement_textures.end(); ++tr)
-		{
-			int j;
-			polymodel *pm = model_get(sip->model_num);
-
-			// look for textures
-			for (j = 0; j < pm->n_textures; j++)
+		if (p_objp != nullptr) {
```
Goober5000, line 10883: Brace should match the enclosing and following brace style, probably.

Goober5000 (commented):
Looks good except for some tweaks.

code/ship/ship.cpp
```diff
@@ -10874,26 +10874,29 @@ void change_ship_type(int n, int ship_type, int by_sexp)
...
-		for (SCP_vector<texture_replace>::iterator tr = p_objp->replacement_textures.begin(); tr != p_objp->replacement_textures.end(); ++tr)
+		if (p_objp != nullptr) 
 		{
-			int j;
-			polymodel *pm = model_get(sip->model_num);
-
-			// look for textures
-			for (j = 0; j < pm->n_textures; j++)
+			// now fill them in according to texture name
+			for (const auto tr : p_objp->replacement_textures)
```
Goober5000, line 10886: The `&` in `const auto &tr` is important.  We want to reference the elements, not copy them.

Goober5000 (changes requested):

code/ship/ship.cpp
Baezon: Oh, I see, I wasn't sure how this worked, its not `const` OR `&`, its a const ref.

Goober5000: Rather, it's a ref or a const ref, but you essentially got it.

Goober5000 (approved):

PR #3568, opened 2021-07-22
Move some flag setting to subsys_set()

Baezon:
Some more bugs from assumptions of ships having parse objects. Beam free by default and turrets that are supposed to start locked should be handled in `subsys_set()` instead of just `parse_create_object_sub()`. I would've liked to remove `Beam_free_all_by_default` handling from `parse_create_object_sub()` entirely, in addition to starting locked, since it should be redundant now, but turrets freed in this way have their initial fire timestamp set to a different value for some reason.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -2300,7 +2300,7 @@ int parse_create_object_sub(p_object *p_objp)
 			}
 		}
 
-		if (shipp->flags[Ship::Ship_Flags::Lock_all_turrets_initially] || ptr->system_info->flags[Model::Subsystem_Flags::Turret_locked])
+		if (shipp->flags[Ship::Ship_Flags::Lock_all_turrets_initially])
 		{
 			// mark all turrets as locked
 			if(ptr->system_info->type == SUBSYSTEM_TURRET)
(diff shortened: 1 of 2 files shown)


Goober5000:
What are the two initial fire timestamp values?

Baezon:
In `subsys_set()` it is initialized to `timestamp((int) frand_range(1.0f, 500.0f))`, but `parse_create_object_sub()`, if applicable, will immediately overwrite it with `timestamp((int) frand_range(50.0f, 4000.0f))`, but only if the beam free by default flag is used, and it doesn't matter if the turret in question is a beam turret.

Goober5000:
Oh.  Whoops, guess I goofed when writing that...

Goober5000:
I would say you can remove the beam-free-all-by-default fire stamp and then consolidate the flags to subsys_set.

Baezon:
I mean, its definitely a small discrepancy, but it does worry me to remove it...

Goober5000:
I need to go back in time and code-review a lot of the commits to CVS and SVN that were done without code review.  The timestamp shouldn't have been added in the first place; that was just a blind copying of the sexp without understanding the ramifications.

Baezon:
Well, alight then. Aside from the timestamp discrepancy, default beam free all can then be handle entirely in `subsys_set()`.

Goober5000 (approved):
Cool, let's see what happens

PR #3569, opened 2021-07-23
Adding basic Vampirism Capabilities to weapons

Strygon:
By giving a weapon the "Vampirism" flag and specifying a "$Vampiric Healing Factor" (taking suggestions for better names) in the tables allows a weapon to heal the shooter by the amount of raw damage dealt by the weapon, multiplied by the healing factor. 

This is currently a fairly basic framework and I'm open to suggestions/ideas on how to further expand this, if desired.

As of now, this does NOT work with beams or shockwaves due to a problem with weapon_info_index registering as -1 and causing crashes. So both types of damage currently do not deal any healing (but they also won't crash the game anymore).

It does however, work with primaries and secondaries.

### DIFF
--- a/code/ship/shiphit.cpp
+++ b/code/ship/shiphit.cpp
@@ -2387,6 +2387,27 @@ static void ship_do_damage(object *ship_objp, object *other_obj, vec3d *hitpos, 
 				apply_diff_scale = false;
 			}
 		}
+			// if weapon is vampiric, slap healing on shooter instead of target
+		if (Weapons[other_obj->instance].weapon_info_index >= 0){
+			weapon_info *wip = &Weapon_info[Weapons[other_obj->instance].weapon_info_index];
+			if (wip->wi_flags[Weapon::Info_Flags::Vampiric]){
+				if (other_obj->parent > 0){
+					object* parent = &Objects[other_obj->parent];
+					if (parent->type == OBJ_SHIP) {
+						if (parent->signature == other_obj->parent_sig){
+							ship* shipparent = &Ships[parent->instance];
+							if (!parent->flags[Object::Object_Flags::Should_be_dead]) {
+								parent->hull_strength += damage * wip->succ_factor;
+								if (parent->hull_strength > shipparent->ship_max_hull_strength){
+									parent->hull_strength = shipparent->ship_max_hull_strength;
+								}
+							}
+						}
+					}
+				}
+			}
+		
+		}
 		// Nuke: this is done incase difficulty scaling is not applied into damage by getDamage() above
 		if (apply_diff_scale) {
 			damage *= difficulty_scale_factor; // Nuke: we can finally stop doing this now
--- a/code/weapon/weapon.h
+++ b/code/weapon/weapon.h
@@ -473,7 +473,10 @@ struct weapon_info
 	// Energy suck effect
 	float weapon_reduce;					// how much energy removed from weapons systems
 	float afterburner_reduce;			// how much energy removed from weapons systems
-
+	
+	// Vampirism Effect -Strygon 7/23/2021
+	float succ_factor;					// Factor by which a vampiric weapon will multiply the healing done to the shooter
+	
 	// tag stuff
 	float	tag_time;						// how long the tag lasts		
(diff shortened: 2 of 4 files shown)


code/weapon/weapon.h
```diff
@@ -473,7 +473,10 @@ struct weapon_info
 	// Energy suck effect
 	float weapon_reduce;					// how much energy removed from weapons systems
 	float afterburner_reduce;			// how much energy removed from weapons systems
-
+	
+	// Vampirism Effect -Strygon 7/23/2021
+	float succ_factor;					// Factor by which a vampiric weapon will multiply the healing done to the shooter
```
Baezon, line 478: Amusing, but you should keep silliness to localized areas of code, this field will show up in autocomplete, and passers-by could get quite confused, so you shouldn't need to track down comments to figure out what this is.

code/ship/shiphit.cpp
```diff
@@ -2386,6 +2386,27 @@ static void ship_do_damage(object *ship_objp, object *other_obj, vec3d *hitpos,
 				damage = Armor_types[shipp->armor_type_idx].GetDamage(damage, damage_type_idx, difficulty_scale_factor, other_obj_is_beam);
 				apply_diff_scale = false;
 			}
+		}
+			// if weapon is vampiric, slap healing on shooter instead of target
+		if (Weapons[other_obj->instance].weapon_info_index >= 0){
```
Baezon, line 2391: This is where you'd use `shiphit_get_damage_weapon(other_obj)` to get the proper `weapon_info_index` before continuing with the rest of the code if you want compatibility with shockwaves and beams.

code/weapon/weapons.cpp
```diff
@@ -2173,6 +2174,11 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 	if( optional_string("$Leech Afterburner:") ){
 		stuff_float(&wip->afterburner_reduce);
 	}
+	
+	// Multiplier for the healing done to the shooter of the weapon.
+	if( optional_string("$Vampiric Healing Factor:") ){
+		stuff_float(&wip->succ_factor);
```
Baezon, line 2180: Costing health to shoot is a cool idea, but I think should be handled separately, and therefore probably best as a separate field entirely. Given that, you should probably make sure that value is not negative. And perhaps also warn if the "vampiric" flag hasn't been used yet they are trying to set this factor? Or the reverse, they used the flag but failed to set this value.

Baezon (changes requested):
Whenever you add a new field to `weapon_info` or `ship_info` or what have you, make sure you initialize the value to something harmless in `weapon_info::reset()` in weapons.cpp.

code/ship/shiphit.cpp
```diff
@@ -2386,6 +2386,27 @@ static void ship_do_damage(object *ship_objp, object *other_obj, vec3d *hitpos,
...
+			weapon_info *wip = &Weapon_info[Weapons[other_obj->instance].weapon_info_index];
+			if (wip->wi_flags[Weapon::Info_Flags::Vampiric]){
+				if (other_obj->parent > 0){
```
Goober5000, line 2394: You can consolidate these two conditions using `&&`.

code/ship/shiphit.cpp
```diff
@@ -2386,6 +2386,27 @@ static void ship_do_damage(object *ship_objp, object *other_obj, vec3d *hitpos,
...
+					object* parent = &Objects[other_obj->parent];
+					if (parent->type == OBJ_SHIP) {
+						if (parent->signature == other_obj->parent_sig){
```
Goober5000, line 2397: Same here.  Also you can add `!parent->flags[Object::Object_Flags::Should_be_dead]` to further consolidate things.

Goober5000 (changes requested):
Some `if()` conditions can be consolidated to cut down on the number of nested braces.

code/weapon/weapon.h
JohnAFernandez: Hint, rename it to something like `vampire_factor` or `vampiric_factor`

PR #3570, opened 2021-07-24
fix WEAPON_LIST_TYPE parsing for -1

Goober5000:
When an empty string is encountered in the weapon list, it's valid, but the -1 should actually be stored, not skipped.

This bug was preventing ship weapons from being explicitly set to None.


Baezon (approved):
This bug was introduced in #3520

PR #3571, opened 2021-07-24
Add some chase view Improvements

Baezon:
Adds a 'Chase View Offset` so that modders can directly specify the 'rest position' of the chase camera for the ship.

Also, the old chase view was very stiff and awkward when it came to 'responding' to the ship's movements, it has since been replaced with a system which takes the ship's acceleration into account and exponentially decays towards the camera's rest position, creating a much more dynamic and responsive feeling to the camera while the ship is maneuvering, making it feel more like a camera man which is trying to keep up with the ship. The 'Rigidity' parameter was added so that modders can define 'how good the camera man is at keeping up', so the camera won't 'swoosh' all over the place while maneuvering too crazily if they don't want.

Acceleration was added to `phys_info` for this purpose, which is a little overengineered for just some camera stuff, but will be a very useful value and I have some more plans in store for it, so I think it's worth it.


wookieejedi:
Ah this will be very useful!

freespace2/freespace.cpp
```diff
@@ -3132,50 +3134,58 @@ camid game_render_frame_setup()
...
-				else
-				{
-					vm_vec_scale_add(&eye_pos, &Viewer_obj->pos, &move_dir, -0.02f * Viewer_obj->radius);
-					vm_vec_scale_add2(&eye_pos, &eyemat.vec.fvec, -2.5f * Viewer_obj->radius - Viewer_chase_info.distance);
-					vm_vec_scale_add2(&eye_pos, &eyemat.vec.uvec, 0.75f * Viewer_obj->radius + 0.35f * Viewer_chase_info.distance);
-					vm_vec_sub(&tmp_dir, &Viewer_obj->pos, &eye_pos);
-					vm_vec_normalize(&tmp_dir);
+				eye_pos = Viewer_obj->pos;
+
+				//get a point 1000m forward of ship
```
wookieejedi, line 3154: Adding a note for updating the comment here so I don't forget.

wookieejedi (changes requested):
Been playing around and looking over this for a few days. Very useful feature, and really helps modders get the right feel for 3rd person in any ship. Only trivial code related aspect was a comment cleanup.

wookieejedi (approved):

code/physics/physics.h
```diff
@@ -100,6 +100,7 @@ typedef struct physics_info {
 
 	matrix ai_desired_orient;   // Asteroth - This is only set to something other than the zero matrix if Framerate_independent_turning is enabled, and 
 								// only by the AI after calls to angular_move. It is read and then zeroed out for the rest of the frame by physics_sim_rot
+	vec3d acceleration;
```
Goober5000, line 103: I would call this `derived_acceleration` or something to make it clear that this is calculated based on the effects of whatever the physics code happens to be doing, *not* instructions for the physics code.

Goober5000 (changes requested):

code/physics/physics.h
Baezon: That's a very good point, but I do prefer the conciseness of `acceleration`. Would a comment that makes this clear suffice? Thinking about it, it follows more or less how `velocity` works, at least for fighters, for better or for worse, in that its really more of a description of the trend of its position, then necessarily a predictor of it (even if it is correlated to it)

Goober5000: If this is consistent with the existing code, then fair enough.  But a comment would be helpful.

Goober5000 (approved):

PR #3572, opened 2021-07-25
Add screenshake and shudder options to game settings tbl

JohnAFernandez:
May not be the best way to implement, but at least the basic work is done, and the PR is submitted as well.


JohnAFernandez:
Just realized that as this is it will disable afterburner and engine wash camera shudder when chase screenshake is enabled, but the player is not in chase view.

Baezon:
The real problem here is that `Viewer_obj` is null at this point, even in personal chase cam, which doesn't make much sense to me. It is specifically nulled out by `game_render_frame_setup()` after it's finished with it. I would see if there's any harm in keeping it around for longer so, yknow, code can actually use it.

PR #3574, opened 2021-07-25
fix Time To: Unknown on the HUD

Goober5000:
PR #3064 inadvertently caused all goals to display Time To: Unknown, even though time was only supplied for docking and waypoints.  This restores the original behavior where the Time To: line simply wasn't displayed.


wookieejedi (approved):
Good catch.

PR #3578, opened 2021-07-29
Anim Rework Phase 1 Fixes

BMagnu:
Fixes two issues identified with the Animation Rework first Phase:
1. Multipart turrets had base and turret angles switched.
2. Initial Type Animations, when called again (due to something like FRED saving and reloading), have already been "completed" and will not start again. This introduces a "forced reset and start" mode for animations, to handle cases like these initial type animations restarting instead of not continuing. Might also be useful for other types of animations in the F3 lab later.

### DIFF
--- a/code/model/modelanimation.cpp
+++ b/code/model/modelanimation.cpp
@@ -103,12 +103,20 @@ namespace animation {
 		}
 	}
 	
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):
Straightforward fix with a quick turnaround.

PR #3579, opened 2021-07-30
Animation Rework Phase 1 Fixes #2

BMagnu:
Adds a missing copy function in the ship_info class missing so far that was occasionally clearing the animations before,
Additionally, convert turret-based AnimationSubmodels into their own child class to properly modularize that code due to all the required exceptions for turret handling everywhere. Also properly sets cur_angle and desired_base/gun_angle for proper behaviour once the turrets target anything.


wookieejedi (approved):
Seems reasonable, and another quick fix. If @Goober5000 gets a chance perhaps he can take a look as well.

PR #3580, opened 2021-08-02
qtFRED Command Breifing

JohnAFernandez:
Adds a fully functioning Command Briefing Dialog to qtFRED

### DIFF
--- a/qtfred/src/mission/dialogs/CommandBriefingDialogModel.cpp
+++ b/qtfred/src/mission/dialogs/CommandBriefingDialogModel.cpp
@@ -1,0 +1,347 @@
+#include "CommandBriefingDialogModel.h"
+#include "missionui/missioncmdbrief.h"
+#include "sound/audiostr.h"
+#include <QMessageBox>
+
+namespace fso {
+namespace fred {
+namespace dialogs {
+
+CommandBriefingDialogModel::CommandBriefingDialogModel(QObject* parent, EditorViewport* viewport) 
+	: AbstractDialogModel(parent, viewport)
+{
+		initializeData();
+}
+
+void CommandBriefingDialogModel::initializeData() 
+{
+	Cur_cmd_brief = Cmd_briefs;  // default to first cmd briefing
+	_wipCommandBrief.num_stages = Cur_cmd_brief->num_stages;
+	strcpy_s(_wipCommandBrief.background[0],Cur_cmd_brief->background[0]);
+	strcpy_s(_wipCommandBrief.background[1],Cur_cmd_brief->background[1]);
+	
+	if (!strlen(_wipCommandBrief.background[0])) {
+		strcpy_s(_wipCommandBrief.background[0], "<default>");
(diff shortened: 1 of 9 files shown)


qtfred/src/ui/dialogs/CommandBriefingDialog.cpp
```diff
@@ -0,0 +1,311 @@
...
+		QFileInfo fileInfo(QFileDialog::getOpenFileName());
+
+		if (fileInfo.fileName().length() == 0) {
+			return;
+		}
+
+		QString filename(fileInfo.fileName());
+
+		if (filename.length() >= CF_MAX_FILENAME_LENGTH) {
+			ReleaseWarning(LOCATION, "No filename in FSO can be longer than 31 characters.");
```
Goober5000, line 108: It would be better to use `SIZE_T_ARG` and provide the #define as an argument, rather than hard-coding it.

qtfred/src/ui/dialogs/CommandBriefingDialog.cpp
```diff
@@ -0,0 +1,311 @@
...
+	{
+		// not yet supported
+	}
+
+	void CommandBriefingDialog::on_actionCopyToOtherTeams_clicked()
+	{
+		// not yet supported.
+	}
+
+	void CommandBriefingDialog::on_actionBrowseAnimation_clicked()
```
Goober5000, line 97: This bit of code is repeated for four functions with only one small difference each time.  It can be separated into its own function and then each onAction can call it.

Goober5000 (commented):
Some comments

qtfred/src/ui/dialogs/CommandBriefingDialog.cpp
JohnAFernandez: Looks like CI complains if I put `SIZE_T_ARG`.  I'm trying int now.

Goober5000: Ah yeah, that isn't a size so `SIZE_T_ARG` wouldn't work.  My bad.

qtfred/src/mission/dialogs/CommandBriefingDialogModel.cpp
```diff
@@ -0,0 +1,347 @@
...
+{
+	_stageNumberUpdateRequired = true;
+	_briefingTextUpdateRequired = true;
+
+	stopSpeech();
+
+	if (_wipCommandBrief.num_stages >= CMD_BRIEF_STAGES_MAX) {
+		_wipCommandBrief.num_stages = CMD_BRIEF_STAGES_MAX;
+		_currentStage = _wipCommandBrief.num_stages - 1;
+		modelChanged(); // signal that the model has changed, in case of inexplicable invaild index.
```
Goober5000, line 149: Missed a misspelling

Goober5000 (approved):

PR #3582, opened 2021-08-02
Animation Rework Phase 1 Fixes #3

BMagnu:
Another round of fixes.
This time, templates caused three problems:
When defining animations for templates, the ship_info_index is yet undefined. For simplicities sake, ship_info information required to find turret model numbers are looked up using ship_info names. This is stable, even for templates.
When copying ship_infos for instantiating templates, animationsegments are not unique anymore. Hence, all the unique pointer actually need to be shared pointers, since the same pointer data is accessed by multiple ship_infos.
To be able to override a ship template's animation data, animations are now all named, with the names being unique. For the old tables, this name is the subobject name. This means that overriding a submodel will then also override the corresponding animation instead of adding to it.

### DIFF
--- a/code/model/modelanimation.cpp
+++ b/code/model/modelanimation.cpp
@@ -169,7 +169,7 @@ namespace animation {
 			cleanRunning();
 	}
 
-	void ModelAnimation::addSubsystemAnimation(std::unique_ptr<ModelAnimationSubmodel> animation) {
+	void ModelAnimation::addSubmodelAnimation(std::shared_ptr<ModelAnimationSubmodel> animation) {
 		m_submodelAnimation.push_back(std::move(animation));
 	}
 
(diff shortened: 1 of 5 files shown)


wookieejedi (changes requested):
Tested and works as expected, even in more involved cases of ships using multiple templates with subsystems added to each successive template. Changes look logical and good. Only thing I noticed was one small comment nitpick. 

Regarding a successive template `overriding a submodel will then also override the corresponding animation instead of adding to it`, very reasonable. I'll note here too that in some templates I change the turret weapon but don't retype the turret animation, and the turret animation is retained (as expected). Only specifying a new animation will override it, which would be expected anyway, so that's also good. 

Overall another solid and quick fix.

wookieejedi (approved):

PR #3584, opened 2021-08-03
FRED: Don't clear poof flags for poof 7 and after

BMagnu:
Previously, on closing the background editing dialog, FRED would clear all poof bits, not just the ones set by the checkboxes.
Since, if someone edited the mission file to add new custom poofs, they probably don't want them reset every time they edit the background.
This changes this, so only the bits that can actually be set again through the checkboxes are cleared. Eventually, and in qtFred, we'd ideally want a dynamic number of checkboxes here.
For now just a quick fix that enables modders to still edit backgrounds when using many custom poofs

### DIFF
--- a/fred2/bgbitmapdlg.cpp
+++ b/fred2/bgbitmapdlg.cpp
@@ -372,7 +372,9 @@ void bg_bitmap_dlg::OnClose()
 		}
 
 		// store poof flags
-		Neb2_poof_flags = 0;
+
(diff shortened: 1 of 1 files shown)


Baezon (approved):
I do think that FRED takes the right approach here in general, but as you said this should be ok until a proper UI can handle it.

BMagnu:
Yeah, usually I'd agree that clearing these values is safer, but not at the cost of always resetting custom poofs if no other way of handling that exists

PR #3586, opened 2021-08-04
ships, wings, etc. can share names with teams

Goober5000:
Partial revert of Mantis 2053, added in 7d4f9f0fb5:
http://scp.indiegames.us/mantis/view.php?id=2053

The patch was speculative and there was never a confirmed bug.  By contrast, there are plenty of existing missions that use Unknown as a ship, wing, or jump node name and work just fine.

### DIFF
--- a/fred2/createwingdlg.cpp
+++ b/fred2/createwingdlg.cpp
@@ -103,13 +103,7 @@ void create_wing_dlg::OnOK()
 		ptr = GET_NEXT(ptr);
 	}
 
-	for (i=0; i<Num_iffs; i++) {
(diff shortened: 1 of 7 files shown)


JohnAFernandez (approved):
Looks like just a ui change for FRED, which power users could have gotten around by manually editing the mission file anyway.  If we start getting reports of the actual issue, we can deal with it.

Goober5000:
Well, the previous behavior did cause problems when editing existing missions in FRED.  I loaded up The Great Hunt (FS2 sm1-06), and opened the Ravana (which is called Unknown) in the ship editor.  That got me into an endless loop where the name kept getting reset to itself and re-triggering the error message.

PR #3588, opened 2021-08-06
Add Bomb-locking secondaries

BMagnu:
This is very useful for making anti-missile secondaries. Also locks onto weapons that are explicitly flagged targetable, even if they aren't bombs.
The `hud_lock_acquire_uncaged_target_weapon` is very closely adapted from the existing acquire uncaged target function.


Baezon:
Its a good idea, but I have a vague dissatisfaction with this design, using +Target lock restriction instead of something like +ship types (+object types?). But I'm not sure how that would shake out in the implementation.

BMagnu:
Hmm, generic object types would be hard, it would probably need its own implementation for each type.
Also, the target lock restrictions themselves get weird:
Current Target makes little sense with weapons as targets, Current Target any Subsystem isn't valid. Only any target really makes sense

PR #3589, opened 2021-08-06
Add ammoless secondaries

BMagnu:
This is essentially the ballistic flag in reverse for secondaries:
With this flag, secondaries no longer consider ammo for shooting, ammo will not be displayed on the hud, rearmed, or anything.


Baezon (commented):
LGTM, an excellent idea and a good addition.

I do wonder if `ship_secondary_bank_has_ammo()` could be used in some areas, or perhaps a slightly different version/macro of it, to cut down on all those checks for `ammo > 0 || NoAmmo`

wookieejedi:
Would it be possible to add this functionality to countermeasures, too?

BMagnu:
As for the macro:
Would certainly be possible, at least for the common case where both ammo and weapon_info are read from a ship_weapon and a secondary bank index.

As for the countermeasures:
I presume so, but since those are handled by a different ammo value, they'd need their checks in different places.
That'd make that a different PR, IMO (other opinions?)

wookieejedi:
Ah, adding a different PR makes sense.

Baezon (approved):

PR #3592, opened 2021-08-07
Handle turret fired type 5 beams without targets properly

### LINKED ISSUE #3590: Bug: Type 5 Beams CTD when used in fire-beam-at-coordinates
Type 5 beams will always crash to desktop when used in fire-beam-at-coordinates.
This applies to T5 beams as well, that do not need any data from the targeted ship (i.e. both start and end relative to center position, neither scale nor orient relative to target).
This should at least be an error, not a CTD, but even better would be to assume a default case (like, orientation is set to the source ship's orientation, hull is assumed to have radius 0, etc), which'd allow for using at least those T5 beams that don't need data from the target to work with this SEXP.

Baezon:
Fixes #3590 

Just a small oversight, even a turret-fired beam is not guaranteed to have a target, if fired using `fire-beam-at-coordinates`. That *should* guarantee that is has coordinates at least, so assert if neither is the case.

### DIFF
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -2368,8 +2368,14 @@ void beam_get_binfo(beam *b, float accuracy, int num_shots, int burst_seed, floa
 		if (b->flags & BF_IS_FIGHTER_BEAM) {
 			orient = b->objp->orient;
 		} else if (b->subsys) {
-			vec3d fvec, uvec;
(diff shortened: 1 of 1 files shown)


BMagnu (changes requested):
Assertion(false, msg) should probably be UNREACHABLE(msg), but apart from that it looks proper, and also works in my test case now.

BMagnu (approved):

PR #3593, opened 2021-08-07
add a version guard for MSVC compiler bug fix

Goober5000:
Per ngld's suggestion, this adds a version check around the disabling of the qfloat16 overloads.

Follow-up to #3557.


wookieejedi (approved):
Good idea.

PR #3594, opened 2021-08-09
Animation Rework Phase 1 Fixes 4 / Performance Issues

BMagnu:
This fixes one critical issue with the animation rework:
The stepAnimation call was done once per frame _per ship_, which is obviously wrong. For initial animations (that don't require stepping), this has no visual impact, but a massive performance issue. The call is moved to the correct place, once per frame, with this PR.
Additionally, since the (due to turrets) very common initial animations don't require stepping at all, they now have special handling to save on unnecessary performance waste. With this, the new animation system should again be not noticeable in performance.


wookieejedi (approved):
Tested and everything works as expected, including getting performance in large missions up to where it should be.

PR #3595, opened 2021-08-09
improve logging of ship creation and cleanup

Goober5000:
It's useful for *all* ship cleanup to be logged, not just big/huge ships.  And by extension, it's also useful for ship creation to be logged.

### DIFF
--- a/code/mission/missionlog.cpp
+++ b/code/mission/missionlog.cpp
@@ -383,8 +383,67 @@ void mission_log_add_entry(LogType type, const char *pname, const char *sname, i
 			nprintf(("missionlog", "new highwater point reached for mission log (%d entries).\n", last_entry));
 		}
 	}
-#endif
 
+	float mission_time = f2fl(Missiontime);
+	int minutes = (int)(mission_time / 60);
+	int seconds = (int)mission_time % 60;
+
+	// record the entry to the debug log too
+	switch (entry->type) {
+		case LOG_SHIP_DESTROYED:
+		case LOG_WING_DESTROYED:
+			nprintf(("missionlog", "MISSION LOG: %s destroyed at %02d:%02d\n", entry->pname, minutes, seconds));
+			break;
+		case LOG_SHIP_ARRIVED:
+		case LOG_WING_ARRIVED:
+			nprintf(("missionlog", "MISSION LOG: %s arrived at %02d:%02d\n", entry->pname, minutes, seconds));
+			break;
+		case LOG_SHIP_DEPARTED:
+		case LOG_WING_DEPARTED:
+			nprintf(("missionlog", "MISSION LOG: %s departed at %02d:%02d\n", entry->pname, minutes, seconds));
+			break;
+		case LOG_SHIP_DOCKED:
+		case LOG_SHIP_UNDOCKED:
+			nprintf(("missionlog", "MISSION LOG: %s %sdocked with %s at %02d:%02d\n", entry->pname, entry->type == LOG_SHIP_UNDOCKED ? "un" : "", entry->sname, minutes, seconds));
+			break;
+		case LOG_SHIP_SUBSYS_DESTROYED:
+			nprintf(("missionlog", "MISSION LOG: %s subsystem %s destroyed at %02d:%02d\n", entry->pname, entry->sname, minutes, seconds));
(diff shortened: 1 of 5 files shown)


BMagnu (approved):

Baezon (changes requested):
I'm not sure this is such a great idea, I did specifically big/huge ships because there are, in general, not a lot of them, and the person sifting through the logs is fairly likely to know when they showed up/were destroyed giving them hopefully some sense of when during a mission a particular log may have been.

Including everything potentially generates as many as *hundreds* of logs, most of which the person sifting through is far less likely to know or care when that particular one came or went, and possibly burying the more important ships.

BMagnu:
Hmm, though having small ships can be useful too in some situations.
Might it be sufficient to add a prefix to the log line for large ships, so that they still stand out / can be easily searched for?

Goober5000:
I thought transports and freighters weren't considered big ships, but it turns out they are, so that takes care of one concern of mine.  Even so, I would cautiously advocate for logging all ships.  You never know when it might be useful.

Generating hundreds of log lines is not a big deal in the context of FSO's log file, because in my previous run the actual beginning of a mission started at line 4,130.  A substantial part of the log file is taken up by compilation of shaders, which I neither care about nor have the expertise to analyze.  But it's easy enough to search for "created at" or "destroyed at" to get to relevant ship information.

I'd certainly be open to suggestions on how the line might be formatted, whether it's by adding a prefix or doing something else.

Baezon:
I mean I'm not really a big fan of all those shader compilations either to be honest..., but my main concern when I added it was merely to give a sense of temporal placement in a given mission, not to actually give information about the ships in question. The kind of information that might be needed is simply too broad and hard to predict for the log to reasonably print. In that sense, adding all ships just adds a lot of noise to that equation, even if you can still find your way around, I'd prefer not to introduce even more bloat than necessary to the logs. 

What exactly is the useful information being provided? It's vaguely useful in the broad sense that almost all gameplay state is possibly useful but we can't print everything. I was mostly just adding it so I have an approximate idea of when in the mission a warning or error was generated. I'm not married to my particular interpretation of this feature, but I'm not convinced of the gain to be had here. Perhaps others could weigh in.

BMagnu:
At least from my perspective, the main use could be that, when modding, your events or scripts don't work because a ship isn't there that should be there.
It can then help a modder find when and how said ship was destroyed/left/etc.
Similarly, maybe it never even showed up, that would be shown from the logs as well.
Though, I agree, input from others would certainly be a very good idea here.

wookieejedi:
In the events editor in FRED there are boxes to check certain conditions when printing to the mission log, would having something like that for the debug log make sense? (Print big ships, all ships, do not print shaders, etc?)

Goober5000:
I don't think setting flags for the debug log would be appropriate here.  That's overengineering a simple task.

My main reason for logging all ships instead of big ships is a) it really is not that much extra logging, especially compared to the log file as a whole; b) the lack of small ship logging could be misleading; c) the particular event that inspired this PR--

This PR was motivated by a debugging effort for the checkpoint script in Colt's upcoming Shepherds campaign.  Colt had been using navbuoys as ship trackers and goal trackers, and the navbuoys weren't showing up as expected.  I wanted an unambiguous indication of whether the buoys had arrived or not (independent of the F4 mission log and in context with the logging done by the saveload script), so I wrote this code.  And navbuoys aren't big ships.

wookieejedi:
Makes sense to me, and given the example case I can see how it would be useful.

JohnAFernandez:
I often turn the debug log into a .csv when doing analysis on some bug, like the recent debris culling bug, or when analyzing bandwidth from multi packets.  The more lines I have to remove, the more arduous the task becomes, especially since each ship often also creates multiple lines.  I know it's not the most used feature -- I've never even used it -- but small ships should be added to a debug filter option.

Baezon:
I would be in favor of having non-big ships be under one of those `nprintf` more verbose logging categories.

Goober5000:
I discovered that some of the debug printing was redundant or in the wrong place or both.  So I made it match the mission log, expanded it to all mission log entry types, and added a debug filter.

Baezon (approved):

PR #3596, opened 2021-08-11
fix icon on FS1 mission import

Goober5000:
The SD Demon doesn't use the support ship icon anymore, and hasn't since FSPort switched from FS2-style icons to FS1-style icons.


wookieejedi (approved):
LGTM.

PR #3597, opened 2021-08-11
Fix for 'anemic' poofs

Baezon:
Follow up to #3507. Due to the confluence of 3 old bugs, the nebula poofs people were getting had significantly more 'oomph' (density, opacity *and* size) than a straightforward analysis of the code would suggest, all of which are compensated for in this PR.

1. Despite literally handing ~0.5 alpha to the renderer, for reasons I can't determine, the actual results were significantly more opaque. Since the new rendering actually displays the proper amount of alpha, I've basically just guess and checked comparisons to adjust; the apparent alpha is actually more like 0.8

2. Unlike all other `Detail` settings, nebula detail has an extra level of indirection for no reason; any changes to nebula detail must update  `neb2_set_detail_level()` rather than just using the value directly. Likely because of this inconsistency, an oversight when retrieving detail settings from the player file means this function is not called, and effectively nebula detail is never updated and always set to the medium initialized value for all players, unless they specifically go to the detail options screen.
All the values chosen for defaults were based on the highest detail, but this means pretty much everyone has been seeing the medium detail, which has smaller and closer poofs, and so the defaults have also been adjusted to those values instead and this pointless indrection has been removed.

3. Due to a math error in g3_render_rect_rotated()` all bitmaps drawn in that way were augmented in size by 41%. I've commented all the functions which used this math to make this behavior more clear and again compensated for this size difference for the new rendering (which renders at the proper size)

Poofs with test bitmaps

Retail:
https://user-images.githubusercontent.com/17305613/128966683-10429d78-ea78-4b28-bcc0-6876dce72241.mp4

#3507 
https://user-images.githubusercontent.com/17305613/128966692-7947e300-7be8-46b4-87f6-819bd6a4545d.mp4

This PR
https://user-images.githubusercontent.com/17305613/128966707-7b1acf29-7241-42f5-853b-56a95cf0d001.mp4

(the changes to the pattern of poofs is intentional, of note here is the density, opacity, and size of poofs)

### DIFF
--- a/code/globalincs/systemvars.cpp
+++ b/code/globalincs/systemvars.cpp
@@ -328,9 +328,6 @@ void detail_level_set(int level)
 	Assert( level < NUM_DEFAULT_DETAIL_LEVELS );
 
 	Detail = Detail_defaults[level];
-
-	// reset nebula stuff
-	neb2_set_detail_level(level);
 }
 
 // Returns the current detail level or -1 if custom.
--- a/code/menuui/optionsmenu.cpp
+++ b/code/menuui/optionsmenu.cpp
@@ -1353,7 +1353,6 @@ void options_detail_sliders_update()
 
 	// modify nebula stuff
 	Detail.nebula_detail = Detail_sliders[gr_screen.res][NEBULA_DETAIL_SLIDER].slider.pos;
-	neb2_set_detail_level(Detail.nebula_detail);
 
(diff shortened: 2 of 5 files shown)


Baezon:
The old density value was total poof cube size at max detail divided by 5 (750 / 5 = 150 distance between each poof). medium detail is 550, so 110.

Size was 150 at max detail, and 125 at medium detail, additionally increased by 41% = 175.

wookieejedi:
Seconded.

BMagnu:
That makes sense then.

PR #3604, opened 2021-08-17
Some fixes to shockwave rotation

### LINKED ISSUE #3603: FR: Randomize rotation of 2d shockwaves

Baezon:
Converting from degrees to radians is *not* 2PI / 180 it's PI /180, so use the macro that has this figured out already. Yes, this will break the rotations of all currently defined shockwave rotations, but at least you can actually rotate into all orientations instead of just half of them now...

Additionally remove a redundant radian conversion when rendering 2D shockwaves, as this effectively prevents any usage of the rotation beyond a few degrees.

Also move `rot_angles` into the shockwave code, it should be modifying the new shockwave, not the `shockwave_create_info`.

Fixes #3603

### DIFF
--- a/code/weapon/shockwave.cpp
+++ b/code/weapon/shockwave.cpp
@@ -136,7 +136,13 @@ int shockwave_create(int parent_objnum, vec3d* pos, shockwave_create_info* sci, 
 	sw->time_elapsed=0.0f;
 	sw->delay_stamp = delay;
 
-	sw->rot_angles = sci->rot_angles;
+	if (!sci->rot_defined) {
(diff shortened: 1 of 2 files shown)


EatThePath:
Pulled and visually verified that it appears to be doing what it's supposed to now

Goober5000:
Impressive... this bug dates back to the `dinky shockwave` commit in 2005: 78fab0c331

As I recall there were a few other bugs from that commit discovered over the past few years.  Too bad we didn't have our review process back then.

Goober5000 (approved):

PR #2634, opened 2020-08-07
Allow multilock missiles to restrict lock based on class or species

Baezon:
Pretty much what it says on the tin. A lock is considered valid if it can match into any of the non-empty restriction lists, or if all are empty. Also warnings for invalid names in any of the lists.

PR #3605, opened 2021-08-17
Add some more missile target restriction checking

Baezon:
Follow-up to #2634. Add target restriction checking to turrets and also AI attacking large ships.


EatThePath:
I can confirm that this at least resolves the test case that prompted it.

JohnAFernandez (approved):
LGTM and the cases that it fixes are part of the expected behavior.

PR #3607, opened 2021-08-19
Reinforcements Dialog

JohnAFernandez:
Creates a working version of the reinforcements Dialog.

Still needs one more bug fixed before merging.

### DIFF
--- a/qtfred/src/mission/dialogs/ReinforcementsEditorDialogModel.cpp
+++ b/qtfred/src/mission/dialogs/ReinforcementsEditorDialogModel.cpp
@@ -1,0 +1,324 @@
+#include "ReinforcementsEditorDialogModel.h"
+#include "ship/ship.h"
+#include <algorithm>
+
+namespace fso {
+namespace fred {
+namespace dialogs {
+
+ReinforcementsDialogModel::ReinforcementsDialogModel(QObject* parent, EditorViewport* viewport)
+	: AbstractDialogModel(parent, viewport)
+{
+	initializeData();
+}
+
+void ReinforcementsDialogModel::initializeData()
+{
+	for (int i = 0; i < Num_reinforcements; i++) {
+		_reinforcementList.emplace_back(Reinforcements[i].name, Reinforcements[i].uses, Reinforcements[i].arrival_delay);
+	}
+
+	// add the wings to the model's internal storage
+	for (auto& currentWing : Wings) {
+
+		if (currentWing.wave_count == 0) {
+			continue;
+		}
+
+		// no players allowed in reinforcements!
+		for (auto& playerTest : currentWing.ship_index) {
+			if (playerTest > -1 && (Objects[Ships[playerTest].objnum].type == OBJ_START)){
+				continue;
+			}
+		}
+
+		bool found = false;
+		for (auto& reinforcement : _reinforcementList) {
+			if (std::get<0>(reinforcement) == currentWing.name) {
+				found = true;
+				break;
+			}
+		}
+
+		if (!found) {
+			_shipWingPool.push_back(currentWing.name);
+		}
+	}
+
+	// add other ships to the remaining storage.
+	for (auto& currentShip : Ships) {
+
+		if (!strlen(currentShip.ship_name) || currentShip.wingnum >= 0 || Objects[currentShip.objnum].type == OBJ_START) {
+			continue;
+		}
+
+		bool found = false;
+
+		for (auto& reinforcement : _reinforcementList) {
+			if (std::get<0>(reinforcement) == currentShip.ship_name) {
+				found = true;
+				break;
+			}
+		}
+
+		if (!found) {
+			_shipWingPool.push_back(currentShip.ship_name);
+		}
+	}
+
+	_selectedReinforcements.clear();
+	_selectedReinforcementIndices.clear();
+	_numberLineEditUpdateRequired = true;
+	_listUpdateRequired = true;
+	modelChanged();
+}
+
+bool ReinforcementsDialogModel::apply() 
+{
+	Num_reinforcements = static_cast<int>(_reinforcementList.size());
+
+	int i = 0;
+
+	// Properly set all reinforcement info.
+	for (auto& modelReinforcement : _reinforcementList) {
+		strcpy_s(Reinforcements[i].name, std::get<0>(modelReinforcement).c_str());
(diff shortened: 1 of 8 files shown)


qtfred/src/mission/dialogs/ReinforcementsEditorDialogModel.cpp
```diff
@@ -0,0 +1,316 @@
...
+			continue;
+		}
+
+		// compiler seems to be optimizing this out despite it being a perfectly valid check.
+		for (auto& playerTest : currentWing.ship_index) {
+			// no players allowed in reinforcements!
+			if (playerTest > -1 && (Objects[Ships[playerTest].objnum].type == OBJ_START)){
+				continue;
+			}
+		}
```
Goober5000, line 33: This `continue` is only continuing the innermost loop.  Since it does nothing else, that's why the compiler optimizes it away.  You would need to either use `goto`, or set a flag, break, check the flag, and *then* `continue`.

qtfred/src/mission/dialogs/ReinforcementsEditorDialogModel.cpp
```diff
@@ -0,0 +1,316 @@
...
+{
+	return _spinboxUpdateRequired;
+}
+
+bool ReinforcementsDialogModel::listUpdateRequired()
+{
+	return _listUpdateRequired;
+}
+
+void ReinforcementsDialogModel::addToReinforcements(SCP_vector<SCP_string> namesIn)
```
Goober5000, line 115: As clang-tidy says, this should be a `const &` parameter

qtfred/src/mission/dialogs/ReinforcementsEditorDialogModel.cpp
```diff
@@ -0,0 +1,316 @@
...
+		_shipWingPool.push_back(std::get<0>(_reinforcementList.back()));
+		_reinforcementList.pop_back();
+	}
+
+	_listUpdateRequired = true;
+	_spinboxUpdateRequired = true;
+	modelChanged();
+}
+
+void ReinforcementsDialogModel::removeFromReinforcements(SCP_vector<SCP_string> namesIn)
```
Goober5000, line 142: Here too

qtfred/src/mission/dialogs/ReinforcementsEditorDialogModel.cpp
```diff
@@ -0,0 +1,316 @@
...
+			return -1;
+		}
+
+	}
+
+	return current;
+}
+
+
+void ReinforcementsDialogModel::selectReinforcement(SCP_vector<SCP_string> namesIn)
```
Goober5000, line 227: And here

Goober5000 (changes requested):

qtfred/src/ui/dialogs/ReinforcementsEditorDialog.cpp
```diff
@@ -0,0 +1,236 @@
+#include "ReinforcementsEditorDialog.h"
+#include "ui_ReinforcementsDialog.h"
+
+#include <globalincs/linklist.h>
+#include <ui/util/SignalBlockers.h>
+#include <QCloseEvent>
+#include <QListWidget>
+#include <qlineedit.h>
+
+constexpr int INTMAX = 2147483647;
```
Goober5000, line 10: Isn't this constant available from somewhere?

JohnAFernandez: I couldn't find one that wasn't part of a Microsoft header, and I didn't want to add a header just for that.

Goober5000: There are many places in the codebase that refer to `INT_MAX`, so it's defined somewhere in a cross-platform way.  Try `#include "globalincs/pstypes.h"`.

Goober5000:
Well, a `const SCP_vector<T> *` works, but why did you use that instead of `const SCP_vector<T> &`?

JohnAFernandez:
I'll come back to this later

JohnAFernandez:
Not ready.  Somehow the wings are not showing up, even though last time I tested they were.

qtfred/src/mission/dialogs/ReinforcementsEditorDialogModel.cpp
```diff
@@ -0,0 +1,324 @@
...
+	for (auto& currentWing : Wings) {
+
+		if (currentWing.wave_count == 0) {
+			continue;
+		}
+
+		// no players allowed in reinforcements!
+		for (auto& playerTest : currentWing.ship_index) {
+			if (playerTest > -1 && (Objects[Ships[playerTest].objnum].type == OBJ_START)){
+				continue;
```
Goober5000, line 31: You're going to run into the problem I mentioned earlier where `continue` only continues the innermost loop.

qtfred/src/mission/dialogs/ReinforcementsEditorDialogModel.cpp
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	int previous = -1, current = -1;
+
+	for (auto& item : _selectedReinforcementIndices) {
+		current = std::get<1>(_reinforcementList.at(item));
+
+		if (previous == -1 || current == previous) {
+			previous = current;
+			continue;
```
Goober5000, line 202: This particular `continue` can be removed as it has no effect.

Goober5000 (changes requested):

qtfred/src/mission/dialogs/ReinforcementsEditorDialogModel.cpp
JohnAFernandez: Ok, test bool added.

qtfred/src/mission/dialogs/ReinforcementsEditorDialogModel.cpp
JohnAFernandez: Done.

Goober5000 (approved):

JohnAFernandez:
I'll try to do one more quick test tonight to make sure things work, and then it's getting merged.

JohnAFernandez:
Ok, merging.  Notice that there is some kind of bug with wing creation where currently a wing may not show up in the reinforcement window until you reload the mission, but that has to be something wrong with the wing creation dialog and not the reinforcements dialog.

PR #3610, opened 2021-08-21
some ai_profiles fixes

Goober5000:
1) Several AI profiles options were getting reset on every iteration through the loop.  They should only be reset when the profile is initially created.
2) Keeping track of AI hits on the player ship was added by Sushi in 9246a2f74d and needs to be guarded by a flag.  This flag has been enabled for mods that set the required FSO version.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -468,7 +468,6 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$firing requires exact los:", AI::Profile_Flags::Require_exact_los);
 
-				profile->ai_path_mode = AI_PATH_MODE_NORMAL;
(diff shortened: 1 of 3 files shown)


Baezon (approved):

code/ai/ai_profiles.cpp
```diff
@@ -568,7 +565,6 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				//Intention is to expand this feature to include a preference for close or long range weapons
 				//hence using something besides a simple flag.
-				profile->ai_range_aware_secondary_select_mode = AI_RANGE_AWARE_SEC_SEL_MODE_RETAIL;
```
EatThePath, line 571: In my defense, I didn't realize this was the one table that looped! 😉 Funny, since I've recently been using this table as a role model and never noticed this goof since...

PR #3614, opened 2021-08-22
Exclude dead subsys from targeting

### LINKED ISSUE #3613: "Current target, any subsystem" multilock mode targets destroyed turrets
This applies both to holding existing locks and creating new ones. Not sure if the same is true of non-turret subsystems. I have not historically used this mode that much, so I suspect this isn't a new flaw, but it doesn't seem like it's likely to be the intended functionality.

BMagnu:
Fixes #3613.
Turns out ``weapon_multilock_can_lock_on_subsys`` just never checked for HP...

### DIFF
--- a/code/hud/hudlock.cpp
+++ b/code/hud/hudlock.cpp
@@ -1485,7 +1485,7 @@ void hud_do_lock_indicators(float frametime)
 			continue;
 		}
 
-		if ( wip->wi_flags[Weapon::Info_Flags::Multilock_ignore_dead_subsys] && lock_slot->subsys != nullptr && lock_slot->subsys->current_hits <= 0.0f) {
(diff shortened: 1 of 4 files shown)


Goober5000 (approved):

Baezon (changes requested):
Hold on now, I'm not sure this is a mistake. Continued in #3613

Baezon (approved):
Although, given I'm not sure this can really be classified as a bug anymore, I'm not sure we should merge this until after the stable.

BMagnu:
Agreed.

EatThePath:
I would like to argue against the not-a-bug decision, but rather than do it here I posted my pamphlet on the matter over in #3613

EatThePath:
In any case I tested this branch, and it does appear to work at keeping you from *acquiring* locks on dead subs, it doesn't do anything about locks you already have on subsystems when they die

BMagnu:
Flipped flag after discussion.
This now changes default behaviour, with the flag to opt out of it and restore original behaviour.
With this, this PR will wait for 21.4 to release

PR #3616, opened 2021-08-22
fixes to look_at and intrinsic rotation

Goober5000:
1) Make sure the `PM_FLAG_HAS_INTRINSIC_ROTATE` flag is only set after submodels pass validation and checking
2) Maintain compatibility with old models that used `$look_at` with the submodel number instead of name

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -1482,7 +1482,6 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
 				// note, this should come BEFORE do_new_subsystem() for proper error handling (to avoid both rotating and look-at submodel)
 				if ((p = strstr(props, "$look_at")) != nullptr) {
 					pm->submodel[n].movement_type = MOVEMENT_TYPE_INTRINSIC_ROTATE;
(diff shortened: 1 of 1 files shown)


code/model/modelread.cpp
```diff
@@ -2592,20 +2590,35 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
...
+					pm->submodel[i].movement_type = MOVEMENT_TYPE_NONE;
 				}
 			}
 		}
 	}
 
+	// And now look through all the submodels and set the model flag if any are intrinsic-rotating
+	for (i = 0; i < pm->n_models; i++) {
+		if (pm->submodel[i].movement_type == MOVEMENT_TYPE_INTRINSIC_ROTATE) {
+			pm->flags |= PM_FLAG_HAS_INTRINSIC_ROTATE;
```
JohnAFernandez, line 2618: Can't we break here? It looks like we only have to find one submodel with the flag.

Goober5000: That's a good idea.

Goober5000: That's a good idea.

JohnAFernandez (approved):

PR #3625, opened 2021-08-28
match wing names and aliases on the status gauge

Goober5000:
If a wing has a hash symbol, it needs to match the squadron status gauge for the equivalent name without the hash.  This could happen for in-mission jumps or cases where a wing leaves and comes back.

Also remove an obsolete function.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -18322,9 +18322,13 @@ int ship_squadron_wing_lookup(const char *wing_name)
 	}
 	else 
 	{
+		// match duplicate wings, such as Epsilon and Epsilon#clone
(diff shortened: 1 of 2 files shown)


code/ship/ship.cpp
```diff
@@ -18322,9 +18322,12 @@ int ship_squadron_wing_lookup(const char *wing_name)
 	}
 	else 
 	{
+		auto ch = get_pointer_to_first_hash_symbol(wing_name);
```
jg18, line 18325: A comment with explanation would be great IMO.

jg18 (approved):
One small non-blocking request but otherwise looks good.

code/ship/ship.cpp
Goober5000: k

PR #3630, opened 2021-08-29
Expose Outer and Inner Radii to the Lua API in the weaponclass type

Kiloku:
This allows scripters to have a little more data to work with. I used it to predict which ships are in range of an Infyrno missile, as a proof of concept for 0rp3u5's [request on HLP](https://www.hard-light.net/forums/index.php?topic=96194.0).


Goober5000 (approved):
Looks good.

PR #3631, opened 2021-08-30
rewrite XSTR to be more robust

### LINKED ISSUE #3627: Escaped quotes in xstr cause freeze in parsing
Tested in a few ship table fields. `\"` inside the string causes the parsing to get stuck on that file but never actually error.

Goober5000:
The old XSTR parsing code was a bit of a mess and did not even properly enforce its format.  These new routines enforce the actual XSTR format and warn when it is malformed, which is helpful for catching table errors as early as possible.  This fixes #3627 and also catches mismatched quotes and other common XSTR errors.  No more wondering why comments are being mysteriously parsed instead of ignored!

### DIFF
--- a/code/localization/localize.cpp
+++ b/code/localization/localize.cpp
@@ -713,6 +713,7 @@ void lcl_fred_replace_stuff(SCP_string &text)
 // XSTR("wheeee", -1)
 // XSTR("whee", 2000)
 // and these should cover all the externalized string cases
+// NOTE: max_len is the maximum string length, not buffer length
 // fills in id if non-NULL. a value of -2 indicates it is not an external string
 // returns true if we were able to extract the XSTR elements (text_str and maybe id are populated)
 bool lcl_ext_localize_sub(const char *in, char *text_str, char *out, size_t max_len, int *id, bool use_default_translation = false)
@@ -731,6 +732,7 @@ bool lcl_ext_localize_sub(const char *in, char *text_str, char *out, size_t max_
 	auto ch = in;
 	bool attempted_xstr = false;
 
+	// this is a pseudo-goto block, not a loop
 	do {
 		// check to see if this is an XSTR() tag
 		if (strnicmp(ch, "XSTR", 4) != 0)
@@ -874,10 +876,14 @@ bool lcl_ext_localize_sub(const char *in, char *text_str, char *out, size_t max_
 		error_display(0, "Malformed XSTR detected:\n\n%s\n", in);
 
 	// copy the entire string (or as much as we can)
+	auto str_len = strlen(xstr_str);
 	strncpy(out, xstr_str, max_len);
+	if (str_len > max_len)
+		out[max_len] = '\0';
+	else
+		out[str_len] = '\0';
 
 	// maybe warn about length
-	auto str_len = strlen(xstr_str);
 	if (str_len > max_len && !Lcl_unexpected_tstring_check)
 		error_display(0, "Token too long: [%s].  Length = " SIZE_T_ARG ".  Max is " SIZE_T_ARG ".\n", xstr_str, str_len, max_len);
 
@@ -895,6 +901,7 @@ bool lcl_ext_localize_sub(const SCP_string &in, SCP_string &text_str, SCP_string
 	auto ch = in.c_str();
 	bool attempted_xstr = false;
 
+	// this is a pseudo-goto block, not a loop
 	do {
 		// check to see if this is an XSTR() tag
 		if (strnicmp(ch, "XSTR", 4) != 0)
--- a/code/localization/localize.cpp
+++ b/code/localization/localize.cpp
@@ -803,18 +803,20 @@ bool lcl_ext_localize_sub(const char *in, char *text_str, char *out, size_t max_
 		//
 
 		// check bounds
-		if (str_end - str_start > PARSE_BUF_SIZE - 1)
+		auto str_length = str_end - str_start;
+		if (str_length > PARSE_BUF_SIZE - 1)
 		{
 			error_display(0, "String cannot fit within XSTR buffer!\n\n%s\n", str_start);
 			break;
 		}
 
 		// now that we know the boundaries of the actual string in the XSTR() tag, copy it
-		strncpy(text_str, str_start, str_end - str_start);
-		text_str[str_end - str_start] = '\0';
+		strncpy(text_str, str_start, str_length);
+		text_str[str_length] = '\0';
 
 		// bounds for id too
-		if (id_end - id_start > PARSE_ID_BUF_SIZE - 1)
+		auto id_length = id_end - id_start;
+		if (id_length > PARSE_ID_BUF_SIZE - 1)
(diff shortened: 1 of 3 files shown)


JadedDragoon:
uh, hmm... any chance of shoe-horning support for string ids in there? As per https://discord.com/channels/223511295431933953/771694956707315752/880933332236501002 ?

Goober5000:
> uh, hmm... any chance of shoe-horning support for string ids in there? As per https://discord.com/channels/223511295431933953/771694956707315752/880933332236501002 ? Yeah, should be it's own FR...

That's not practical.  It would require a significant rewrite of the localization system.

code/localization/localize.cpp
```diff
@@ -724,199 +717,318 @@ void lcl_fred_replace_stuff(SCP_string &text)
...
-	if ( !Xstr_inited || (str_id < 0) || (!use_default_translation && ((Lcl_current_lang == LCL_UNTRANSLATED) || (Lcl_current_lang == LCL_RETAIL_HYBRID))) ) {
-		if ( strlen(text_str) > max_len && !Lcl_unexpected_tstring_check )
-			error_display(0, "Token too long: [%s].  Length = " SIZE_T_ARG ".  Max is " SIZE_T_ARG ".\n", text_str, strlen(text_str), max_len);
+		//
+		// split off the strings and id sections
+		//
 
-		strncpy(out, text_str, max_len);
+		// check bounds
+		if (str_end - str_start > PARSE_BUF_SIZE - 1)
```
jg18, line 806: Consider adding
```
int str_length = (int)(str_end - str_start);
```
above this line and then using `str_length` here and also below.

Goober5000: Good idea.  Done.

code/localization/localize.cpp
```diff
@@ -724,199 +717,321 @@ void lcl_fred_replace_stuff(SCP_string &text)
...
+		//
+		// split off the strings and id sections
+		//
 
-	// set the id #
-	if (id != NULL){
-		*id = str_id;
-	}
+		// now that we know the boundaries of the actual string in the XSTR() tag, copy it
+		text_str.assign(str_start, str_end);
```
jg18, line 970: Shouldn't the second argument be `str_end - str_start`?

code/localization/localize.cpp
```diff
@@ -724,199 +717,321 @@ void lcl_fred_replace_stuff(SCP_string &text)
...
+	auto str_len = strlen(xstr_str);
+	if (str_len > max_len && !Lcl_unexpected_tstring_check)
+		error_display(0, "Token too long: [%s].  Length = " SIZE_T_ARG ".  Max is " SIZE_T_ARG ".\n", xstr_str, str_len, max_len);
 
-	return true;
+	return xstr_valid;
 }
 
 // ditto for SCP_string
 bool lcl_ext_localize_sub(const SCP_string &in, SCP_string &text_str, SCP_string &out, int *id, bool use_default_translation = false)
```
jg18, line 888: It might make the code simpler if this function just called into the version above that uses C-strings, but maybe that isn't practical. I assume we wouldn't want the reverse, i.e. have that version just call into this one, because of the performance hit from creating/destroying temporary `SCP_string` objects.

If we knew what value to pass for `max_len` and could preallocate that amount of space in `text_str`/`out`, would calling the C-string version work? Or maybe a bad idea?

jg18 (changes requested):
One requested change, two optional changes, and some "food for thought" while I continue to review.

code/localization/localize.cpp
```diff
@@ -724,199 +717,321 @@ void lcl_fred_replace_stuff(SCP_string &text)
...
+	// set whatever id we have
+	if (id != nullptr)
+		*id = xstr_id;
+
+	// if we made an attempt but failed, let the modder know
+	if (xstr_id == -2 && attempted_xstr)
+		error_display(0, "Malformed XSTR detected:\n\n%s\n", in);
+
+	// copy the entire string (or as much as we can)
+	strncpy(out, xstr_str, max_len);
```
jg18, line 877: Do we need to add the terminating null char to `out`?

code/localization/localize.cpp
Goober5000: I haven't investigated whether that would work, but this also works.  This treats the function parameters as a start iterator and an end iterator.

code/localization/localize.cpp
```diff
@@ -724,199 +717,321 @@ void lcl_fred_replace_stuff(SCP_string &text)
...
 
 	// NOTE: "Token too long" warnings are disabled when Lcl_unexpected_tstring_check is active,
 	// because in such cases we actually anticipate that the length might be exceeded.
 
-	// default (non-external string) value
-	if (id != NULL) {
-		*id = -2;
-	}
+	// set up return values
+	auto xstr_str = in;
```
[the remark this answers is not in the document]
Goober5000: Yeah, I don't usually do that.

code/localization/localize.cpp
Goober5000: I considered combining the functions, but I want to leave them as they are.  Ideally, the SCP_string version would use proper SCP_string accessor functions, but that can wait for a future upgrade.

code/localization/localize.cpp
Goober5000: We probably do.

code/localization/localize.cpp
```diff
@@ -724,199 +717,321 @@ void lcl_fred_replace_stuff(SCP_string &text)
...
 
-	str_len = strlen(in);
+	auto ch = in;
+	bool attempted_xstr = false;
 
-	// if the string is < 9 chars, it can't be an XSTR("",) tag, so just copy it
-	if (str_len < 9) {
-		if (str_len > max_len && !Lcl_unexpected_tstring_check)
-			error_display(0, "Token too long: [%s].  Length = " SIZE_T_ARG ".  Max is " SIZE_T_ARG ".\n", in, str_len, max_len);
+	do {
```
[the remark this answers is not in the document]
Goober5000: *shrug*

code/localization/localize.cpp
jg18: Okay.

jg18 (approved):
While I'm not an expert on this code, this change looks reasonable.

code/localization/localize.cpp
```diff
@@ -720,203 +713,332 @@ void lcl_fred_replace_stuff(SCP_string &text)
...
+		if (str_length > PARSE_BUF_SIZE - 1)
+		{
+			error_display(0, "String cannot fit within XSTR buffer!\n\n%s\n", str_start);
+			break;
+		}
 
-		if (id != NULL)
-			*id = str_id;
+		// now that we know the boundaries of the actual string in the XSTR() tag, copy it
+		strncpy(text_str, str_start, str_length);
```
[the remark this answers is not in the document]
Goober5000: Yup, this is in the parent function `lcl_ext_localize`.

PR #3632, opened 2021-08-30
Add full team option to change-iff SEXP

BMagnu:
This allows the change-iff SEXP to take in another team to change.
A usecase for this can be an event that makes every neutral craft hostile for example, and this change allows for that without having to list all potentially neutral wings and crafts in the sexp

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -10639,6 +10639,22 @@ void sexp_change_iff_helper(object_ship_wing_point_team oswpt, int new_team)
 
 			break;
 		}
+
+		case OSWPT_TYPE_WHOLE_TEAM:
+		{
+			for (ship_obj* so = GET_FIRST(&Ship_obj_list); so != END_OF_LIST(&Ship_obj_list); so = GET_NEXT(so)) {
+				ship* shipp = &Ships[Objects[so->objnum].instance];
+				if (shipp->team == oswpt.team)
+					sexp_ingame_ship_change_iff(shipp, new_team);
+			}
+
+			for (p_object* p_objp = GET_FIRST(&Ship_arrival_list); p_objp != END_OF_LIST(&Ship_arrival_list); p_objp = GET_NEXT(p_objp))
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -32158,7 +32179,7 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 
 	// Goober5000 - added wing capability
 	{ OP_CHANGE_IFF, "Change IFF (Action operator)\r\n"
-		"\tSets the specified ship(s) or wing(s) to the specified team.\r\n"
+		"\tSets the specified ship(s) or wing(s) or all ships of a team to the specified team.\r\n"
 		"Takes 2 or more arguments...\r\n"
 		"\t1:\tTeam to change to (\"friendly\", \"hostile\" or \"unknown\").\r\n"
 		"\tRest:\tName of ship or wing to change team status of." },
```
Goober5000, line 32185: Update the help text here too

Goober5000:
We already have the `for-team` sexp which can be used with `when-argument`.  Is it really a good idea to wholesale change one team to another?  *Especially* for ships on that team which haven't arrived yet?

BMagnu:
I think there can be use cases for it.
Admittedly, it's more of an edge case, but especially for the case where you want non-arrived ships to also be changed, for-team has shortcomings.

Goober5000:
In what situation would you want a whole team of non-arrived ships to be changed?

BMagnu:
It's a case usually only occurring if you have three or more teams in mission, and one of these teams as a whole changes their allegiance to the other teams.
My specific usecase looks as follows:
There is the player + faction, as well as two other factions. Both of these other factions are hostile to each other, but friendly to the player. The player can then support either faction, which'll flip the other faction to outright hostile to both the player and the supported faction. This obviously should include later arrivals of this team as well.

Goober5000:
Ah, okay.

Goober5000 (approved):

PR #3635, opened 2021-09-01
run Lua garbage collector on mission start

### LINKED ISSUE #2961: Performance hit upon subsequent mission replays
Recently I have noticed, that replaying the Icarus cutscene causes the FPS to drop considerably. Here are my results from three subsequent plays, launching the mission through "continue campaign" and then replaying it through "replay mission" button in the debriefing.

05-12-2020, 20:19:41 fs2_open_20_1_0_20201001_39d77d8_x64_SSE2.exe benchmark completed, 16972 frames rendered in 178.406 s

                     Average framerate  :   95.1 FPS
                     Minimum framerate  :   48.0 FPS
                     Maximum framerate  :  123.9 FPS
                     1% low framerate   :   45.4 FPS
                     0.1% low framerate :   21.3 FPS
05-12-2020, 20:22:49 fs2_open_20_1_0_20201001_39d77d8_x64_SSE2.exe benchmark completed, 14889 frames rendered in 179.469 s

                     Average framerate  :   82.9 FPS
                     Minimum framerate  :   49.3 FPS
                     Maximum framerate  :  120.3 FPS
                     1% low framerate   :   46.3 FPS
                     0.1% low framerate :   25.4 FPS
05-12-2020, 20:25:57 fs2_open_20_1_0_20201001_39d77d8_x64_SSE2.exe benchmark completed, 14232 frames rendered in 176.687 s

                     Average framerate  :   80.5 FPS
                     Minimum framerate  :   48.2 FPS
                     Maximum framerate  :  120.7 FPS
                     1% low framerate   :   44.5 FPS
                     0.1% low framerate :   21.7 FPS

The benchmarks were made with MSI Afterburner, with a version of the software no more than a couple weeks old. As you can see, the FSO version used is 2020-10-01 nightly. Hardware specifications are: Ryzen 7 4800H with 16GB of single channel 3200MHz memory with an RTX 2060 Mobile (I believe it's the refreshed version). There was no throttling of any sort. Also my observations of the performance statistics suggest that there is a considerable CPU bottleneck at play.

Also, important note: the first and third runs happened to experience the issue where the Karuna model stops colliding with beam weapons after destroying a specific turret on it, so I assume that the actual average FPS results would have been ever so slightly higher in these two runs if not for this bug.

This might also suggest that FSO performance is constantly degrading between mission replays, and this performance degradation is being removed if you restart the game. This might be feasible given some vague reports of people restarting FSO after longer game sessions to restore decent FPS levels.

Goober5000:
This might fix #2961.  By explicitly running the garbage collector before a mission starts, this will ensure each mission begins with a clean slate.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -188,6 +188,13 @@
 #include "weapon/shockwave.h"
 #include "weapon/weapon.h"
 
+// for the Lua garbage collector
+extern "C"
+{
+	#include "lgc.h"
+}
+
+
 #include <SDL.h>
 #include <SDL_main.h>
 
@@ -1360,7 +1367,13 @@ bool game_start_mission()
 
 	game_busy( NOX("** starting mission_load() **") );
 	load_mission_load = (uint) time(nullptr);
-	if ( !mission_load(Game_current_mission_filename) ) {
+	bool load_success = mission_load(Game_current_mission_filename);
+	load_mission_load = (uint)(time(nullptr) - load_mission_load);
+
+	// free up memory from parsing the mission
+	stop_parse();
+
+	if ( !load_success ) {
 		if ( !(Game_mode & GM_MULTIPLAYER) ) {
 			popup(PF_BODY_BIG | PF_USE_AFFIRMATIVE_ICON, 1, POPUP_OK, XSTR( "Attempt to load the mission failed", 169));
 			gameseq_post_event(GS_EVENT_MAIN_MENU);
@@ -1376,11 +1389,15 @@ bool game_start_mission()
 
 		return false;
 	}
-	load_mission_load = (uint) (time(nullptr) - load_mission_load);
 
-	// free up memory from parsing the mission
-	extern void stop_parse();
-	stop_parse();
+	// Since we just freed up memory, now is probably a good time to run the Lua garbage collector.
+	// This is also after the preliminary checks are complete but before things start getting paged in.
+	auto L = Script_system.GetLuaSession();
+	if (L != nullptr)
+	{
+		game_busy(NOX("** cleaning up Lua objects **"));
+		luaC_fullgc(L);
+	}
 
 	game_busy( NOX("** starting game_post_level_init() **") );
 	load_post_level_init = (uint) time(nullptr);


Goober5000:
[fs2_open_21_3_0_x64_AVX2.zip](https://github.com/scp-fs2open/fs2open.github.com/files/7087642/fs2_open_21_3_0_x64_AVX2.zip)

Test build attached.

freespace2/freespace.cpp
```diff
@@ -1360,7 +1367,13 @@ bool game_start_mission()
 
 	game_busy( NOX("** starting mission_load() **") );
 	load_mission_load = (uint) time(nullptr);
-	if ( !mission_load(Game_current_mission_filename) ) {
+	bool load_success = mission_load(Game_current_mission_filename);
+	load_mission_load = (uint)(time(nullptr) - load_mission_load);
```
z64555, line 1371: Er, what's the reason for moving this up here?

Goober5000: Because if the mission load is unsuccessful, the variable will not be updated.

It's for the same reason I moved `stop_parse` - if the mission load is unsuccessful, the memory wouldn't be freed until, presumably, the next time a mission needed to be parsed.

JohnAFernandez (commented):
Looks G2G for me, pending testing.

BMagnu:
Not sure if it fixes the performance problems, but I've tested a few scripts, including some that rely on inter-mission data (like the radaricon script), and they seem to be working fine.
Regardless of if this actually fixes the performance, forcing the Lua GC there seems to be a sensible idea anyhow, so it should be merged regardless IMO.

freespace2/freespace.cpp
```diff
@@ -1376,11 +1389,15 @@ bool game_start_mission()
...
-	extern void stop_parse();
-	stop_parse();
+	// Since we just freed up memory, now is probably a good time to run the Lua garbage collector.
+	// This is also after the preliminary checks are complete but before things start getting paged in.
+	auto L = Script_system.GetLuaSession();
+	if (L != nullptr)
+	{
+		game_busy(NOX("** running Lua garbage collector **"));
+		luaC_fullgc(L);
+	}
```
jg18, line 1400: Is it worth adding an `else` clause where we log that we were unable to get the Lua session?

If `L` were `nullptr`, would that mean something's wrong? Or not necessarily?

freespace2/freespace.cpp
```diff
@@ -1376,11 +1389,15 @@ bool game_start_mission()
 
 		return false;
 	}
-	load_mission_load = (uint) (time(nullptr) - load_mission_load);
 
-	// free up memory from parsing the mission
-	extern void stop_parse();
-	stop_parse();
```
jg18, line 1383: Just wondering why this call was removed?

Is `stop_parse()` called elsewhere as well? Or is calling it not strictly required?

freespace2/freespace.cpp
Goober5000: I don't know, TBH.  It might mean that scripting isn't supported, or that the engine was built without scripting support.  It might not actually be *possible* for `L` to be `nullptr` here, but I figured better safe than sorry.

jg18: Oh I absolutely agree that we should defensively check for `nullptr`. I was just wondering whether it'd be worthwhile to log a debug message if `L` *is* `nullptr`, e.g.,
```
Attempted to run Lua GC after parsing the mission but Lua session is null
```
but it's not important.

freespace2/freespace.cpp
jg18: I'm unclear on what `load_mission_load` is even for. I see no uses of it after this line.

freespace2/freespace.cpp
```diff
@@ -1376,11 +1389,15 @@ bool game_start_mission()
...
+	// Since we just freed up memory, now is probably a good time to run the Lua garbage collector.
+	// This is also after the preliminary checks are complete but before things start getting paged in.
+	auto L = Script_system.GetLuaSession();
+	if (L != nullptr)
+	{
+		game_busy(NOX("** running Lua garbage collector **"));
```
jg18, line 1398: A more player-friendly message like
```
"** cleaning up Lua objects **"
```
might be nice, but then again, this message appears only in debug builds, so not that important.

Goober5000: I like that; will do

freespace2/freespace.cpp
Goober5000: I don't think that's necessary.  If `L` is valid, the garbage collector should be run; if not, the gc is irrelevant.

freespace2/freespace.cpp
Goober5000: I think it could be for timing or profiling.  I left it in in case it could be usable in the future.  There are other apparently orphaned variables too; maybe this could be addressed in a future PR.

jg18 (approved):

freespace2/freespace.cpp
jg18: Yes, definitely agree that leaving it be was the right thing to do, for the reasons you mentioned. :+1:

PR #3636, opened 2021-09-01
Fix turrets not resetting sometimes

Baezon:
`ai_fire_from_turret()` has been renamed to something more appropriate `ai_turret_execute_behavior()` to make it clear that it's more general than literally just shooting at things. The function needs to be evaluated regardless of whether there are enemies on the field or the turret has a target because that is where resetting is done. The function is perfectly capable of determining on its own that there are no valid targets and not to do anything other than possibly reset.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -2130,7 +2130,7 @@ int Num_turrets_fired = 0;
  * Given a turret tp and its parent parent_objnum, fire from the turret at its enemy.
  */
 extern int Nebula_sec_range;
-void ai_fire_from_turret(ship *shipp, ship_subsys *ss)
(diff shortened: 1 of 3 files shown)


code/ai/aiturret.cpp
```diff
@@ -2130,7 +2130,7 @@ int Num_turrets_fired = 0;
  * Given a turret tp and its parent parent_objnum, fire from the turret at its enemy.
  */
 extern int Nebula_sec_range;
-void ai_fire_from_turret(ship *shipp, ship_subsys *ss)
+void ai_turret_execute_behavior(ship *shipp, ship_subsys *ss)
```
Goober5000, line 2133: Add a comment with the old function name, just in case someone is comparing FSO vs. retail code.

Otherwise this looks good.

Goober5000 (approved):

PR #3637, opened 2021-09-01
Expose OSWPT to Lua and allow for LuaSEXP argument

BMagnu:
This finally remedies my biggest annoyance with LuaSEXPs: It is now possible to specify OSWPT-types as arguments, allowing for custom sexps to work for much broader situations.
As convenience functions, it's not only possible to access OSWPT data directly, but rather supply a function that should then be applied to every ship (or parse object) this OSWPT applies to, allowing for the creation of easy "do n for all Ships/Wings/Teams" SEXPs with little effort.
The newly added argument types are called the following (they include the oswpt prefix to signify that all of these return an OSWPT object, and just differ in how they can be populated in FRED), and are generally equivalent to their built-in SEXP counterparts:
```
ship+waypoint
ship+wing
ship+wing+team
ship+wing+ship_on_team+waypoint
ship+wing+waypoint
ship+wing+waypoint+none
```

### DIFF
--- a/code/parse/sexp/LuaSEXP.cpp
+++ b/code/parse/sexp/LuaSEXP.cpp
@@ -10,6 +10,7 @@
 #include "parse/sexp.h"
 #include "parse/sexp/sexp_lookup.h"
 #include "scripting/api/objs/message.h"
+#include "scripting/api/objs/oswpt.h"
 #include "scripting/api/objs/sexpvar.h"
 #include "scripting/api/objs/ship.h"
 #include "scripting/api/objs/shipclass.h"
@@ -38,7 +39,14 @@ SCP_unordered_map<SCP_string, int> parameter_type_mapping{{ "boolean",      OPF_
 														  { "wing",         OPF_WING },
 														  { "shipclass",    OPF_SHIP_CLASS_NAME },
 														  { "weaponclass",  OPF_WEAPON_NAME },
-														  { "soundentry",   OPF_GAME_SND }, };
+														  { "soundentry",   OPF_GAME_SND }, 
+														  { "oswpt-ship+waypoint",   OPF_SHIP_POINT },
+														  { "oswpt-ship+wing",   OPF_SHIP_WING },
+														  { "oswpt-ship+wing+team",   OPF_SHIP_WING_WHOLETEAM },
+														  { "oswpt-ship+wing+team+waypoint",   OPF_SHIP_WING_SHIPONTEAM_POINT },
+														  { "oswpt-ship+wing+waypoint",   OPF_SHIP_WING_POINT },
+														  { "oswpt-ship+wing+waypoint+none",   OPF_SHIP_WING_POINT_OR_NONE }, };
+
 std::pair<SCP_string, int> get_parameter_type(const SCP_string& name)
 {
 	SCP_string copy = name;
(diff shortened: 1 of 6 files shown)


JohnAFernandez:
I'd love to review this, but it looks like it could really use someone who knows scripting a little better?

code/scripting/api/objs/oswpt.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include "team.h"
+#include "waypoint.h"
+#include "wing.h"
+#include "ship/ship.h"
+
+namespace scripting {
+namespace api {
+
+//**********HANDLE: Wing
+ADE_OBJ(l_OSWPT, object_ship_wing_point_team, "oswpt", "OSWPT handle");
```
asarium, line 16: `OSWPT` isn't really useful in this context since it does not help in explaining what this handle does. Please add a better description here which explains that this is a combined handle to either of those categories.

asarium (changes requested):

code/parse/sexp/LuaSEXP.cpp
```diff
@@ -38,7 +39,14 @@ SCP_unordered_map<SCP_string, int> parameter_type_mapping{{ "boolean",      OPF_
...
 														  { "shipclass",    OPF_SHIP_CLASS_NAME },
 														  { "weaponclass",  OPF_WEAPON_NAME },
-														  { "soundentry",   OPF_GAME_SND }, };
+														  { "soundentry",   OPF_GAME_SND }, 
+														  { "oswpt-ship+waypoint",   OPF_SHIP_POINT },
+														  { "oswpt-ship+wing",   OPF_SHIP_WING },
+														  { "oswpt-ship+wing+team",   OPF_SHIP_WING_WHOLETEAM },
+														  { "oswpt-ship+wing+team+waypoint",   OPF_SHIP_WING_SHIPONTEAM_POINT },
+														  { "oswpt-ship+wing+waypoint",   OPF_SHIP_WING_POINT },
+														  { "oswpt-ship+wing+waypoint+none",   OPF_SHIP_WING_POINT_OR_NONE }, };
```
Goober5000, line 48: You could remove the `oswpt-` prefix and just declare the scripting type to be the data it accepts

code/scripting/api/objs/oswpt.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+	switch (oswpt.type) {
+	case OSWPT_TYPE_SHIP:
+		return ade_set_args(L, "s", "ship");
+	case OSWPT_TYPE_PARSE_OBJECT:
+		return ade_set_args(L, "s", "parseobject");
+	case OSWPT_TYPE_WING:
+	case OSWPT_TYPE_WING_NOT_PRESENT:
+		return ade_set_args(L, "s", "wing");
+	case OSWPT_TYPE_SHIP_ON_TEAM:
```
Goober5000, line 32: `SHIP_ON_TEAM` should be distinguished separately

code/scripting/api/objs/oswpt.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+	switch (oswpt.type) {
+	case OSWPT_TYPE_SHIP:
+		return ade_set_args(L, "o", l_Ship.Set(object_h(oswpt.objp)));
+	case OSWPT_TYPE_PARSE_OBJECT:
+		return ade_set_args(L, "o", l_ParseObject.Set(parse_object_h(oswpt.ship_entry->p_objp)));
+	case OSWPT_TYPE_WING:
+	case OSWPT_TYPE_WING_NOT_PRESENT:
+		return ade_set_args(L, "o", l_Wing.Set(WING_INDEX(oswpt.wingp)));
+	case OSWPT_TYPE_SHIP_ON_TEAM:
```
Goober5000, line 58: Will `l_Team.Set` work with the ship-on-team values?

BMagnu: Yes, ship-on-team will fill the wingp property of oswpt

code/scripting/api/objs/oswpt.cpp
BMagnu: Since this is about the data-type that the scripter would have to deal with (which can vary for multiple invocations of the sexp), whereas the ships-on-team vs team distiction is always the same for the whole sexp, it makes sense to keep them handled the same here

asarium (approved):

Goober5000 (approved):

Issue #3414, opened 2021-05-05
adjust-audio-volume [effects] doesn't fade beams


MjnMixael:
This sexp currently will not fade the volume of any currently firing beams. After the beam finishes, the next beam will be adjusted properly. I suspect this goes deeper and doesn't actually adjust the volume of any currently playing effects (it does fade music; I did not test voice), but I could be wrong. It's easy to tell that it doesn't fade beams. Harder with lasers and stuff.

Test mission attached. Press 1 to fire a beam and begin a fade out of the audio.
[BeamSoundFade.zip](https://github.com/scp-fs2open/fs2open.github.com/files/6430250/BeamSoundFade.zip)

jg18:
@MjnMixael Thanks for the test mission. So far I've just dug into the code.

Turns out the main beam firing sounds are in a special category of "looping 3D" sounds that the `adjust-audio-volume` SEXP currently doesn't handle.

There are a few other "looping 3D" sounds, but I haven't yet looked at what they are.

(Details for coders: these are the sounds you get from calling `snd_play_3d()` with `looping` set to `1`.)

The beam sounds (and maybe other "looping 3D" sounds) also need to be able to be paused (e.g., when you go to the Options menu) and then unpaused with their volumes from before the pause. It looks like `beam_unpause_sounds()` just sets the beam sound volumes to full, which isn't what we want.

I have some ideas on what to do but will need to think and experiment with the test mission.

jg18:
Ok, t he other "looping 3D" sounds are "object-persistent" sounds (coders: see objectsnd.cpp), which is for things like other ships' engine wash and turret/subsystem (in)operation (except for turret fire).

I guess all of these sounds should also be affected by `adjust-audio-volume`.

Still looking into the best way to implement this.

Btw, one looping (non-3D) sound that *isn't* affected by `adjust-audio-volume` is the player's own engine. That's the only exception I've noticed so far.

Goober5000:
Are you referring to `ship_p.wav`, which is "engine sound (as heard in cockpit)"?  Because EngineSnd isn't played on the player's own ship.

jg18:
@Goober5000 I was referring to `Player_engine_snd_loop` which is set in `update_throttle_sound()` in hud.cpp by calling `snd_play_looping()` with `scriptingUpdateVolume` set to `FALSE`.

I'm not exactly sure what sound that var corresponds to, though.

[...]

PR #3638, opened 2021-09-03
Apply adjust-audio-volume SEXP for effects to looping 3D sounds

jg18:
Looping 3D sounds are defined as sounds played using `snd_play_3d()` with the `looping` argument set to `1`. These include beam fire and object-persistent sounds (see objectsnd.cpp), which include the engine and subsystem operation sounds of other ships.

Intended to fix Issue #3414.

### DIFF
--- a/code/sound/sound.cpp
+++ b/code/sound/sound.cpp
@@ -111,16 +111,6 @@ struct LoopingSoundInfo {
     }
 };
 
-bool does_ls_info_match_sig(const LoopingSoundInfo &ls_info, sound_handle sig)
-{
-	return ls_info.m_dsHandle == sig;
-}
-
-std::function<bool(const LoopingSoundInfo&)> create_ls_info_match_pred(sound_handle sig)
-{
-	return std::bind(does_ls_info_match_sig, std::placeholders::_1, sig);
-}
-
 SCP_list<LoopingSoundInfo> currentlyLoopingSoundInfos;
 SCP_list<LoopingSoundInfo> currentlyLooping3dSoundInfos;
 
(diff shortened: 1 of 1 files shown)


code/sound/sound.cpp
```diff
@@ -737,6 +743,10 @@ sound_handle snd_play_3d(game_snd* gs, vec3d* source_pos, vec3d* listen_pos, flo
 	}
 
 	if (handle.isValid()) {
+		if (looping) {
+			currentlyLooping3dSoundInfos.emplace_back(handle, default_volume, vol_scale);
```
jg18, line 747: These parameters were copied from the existing call in `snd_play()` for `currentlyLoopingSoundInfos`. I'm not sure if these are the correct parameters to use here.

Baezon (commented):
Tested and works.

code/sound/sound.cpp
```diff
@@ -904,6 +919,12 @@ sound_handle snd_play_looping(game_snd* gs, float pan, int /*start_loop*/, int /
 	return handle;
 }
 
+void remove_looping_sound(SCP_list<LoopingSoundInfo> &looping_sounds, sound_handle sig)
+{
+	// the cast to void avoids warnings about unused return value
+	(void)std::remove_if(looping_sounds.begin(), looping_sounds.end(), create_ls_info_match_pred(sig));
```
Goober5000, line 925: Use a lambda here, not a full-up `std::function`.

Goober5000 (changes requested):
see comment

code/sound/sound.cpp
jg18: Good point. The intent was to avoid code duplication, but the result ended up making things more complicated, haha.

Goober5000 (approved):

Issue #3388, opened 2021-04-17
Bug/misfeature report: Handling of missing sounds needs to change.


EatThePath:
It appears that if a sound is absent from the game files the engine will try to load it every time it could occur. This is an expensive operation and can cause performance loss if the sound is called frequently. Consequently, missing sounds should either be upgraded to a debug warning instead of a logprint, or the engine should remember failures and not try to reload bad sounds, or probably both.

Profiling a Solaris mission is showing me about 16% of CPU time is lost to trying to reload a single missing sound file, for context.

notimaginative:
We could probably get away with using a placeholder sound. It's easy to create silence so if a file is missing or fails to load we could log it then load silence in it's place. The engine would load and play the "silence" rather than continuing to look for the missing file.

Goober5000:
From @notimaginative on Discord:
> I didn't comment again on 3388 but I got a little worried about possible side effects when sounds would technically always load successfully. A permanent fail might be the least problematic solution, but I'm not sure what that would entail.

Goober5000:
fixed in #3639

PR #3639, opened 2021-09-03
handle invalid gamesnds

Goober5000:
If an entry in a game sound fails to load, mark the sound as invalid and don't try to load it afterwards.  This is the "permanent fail" fix for #3388.

### DIFF
--- a/code/gamesnd/gamesnd.cpp
+++ b/code/gamesnd/gamesnd.cpp
@@ -544,12 +544,12 @@ void gamesnd_preload_common_sounds()
 		return;
 
 	Assert( Snds.size() <= INT_MAX );
-	for (SCP_vector<game_snd>::iterator gs = Snds.begin(); gs != Snds.end(); ++gs) {
-		if ( gs->preload ) {
-			for (auto& entry : gs->sound_entries) {
+	for (auto& gs: Snds) {
+		if ( gs.preload ) {
+			for (auto& entry : gs.sound_entries) {
 				if ( entry.filename[0] != 0 && strnicmp(entry.filename, NOX("none.wav"), 4) != 0 ) {
 					game_busy( NOX("** preloading common game sounds **") );	// Animate loading cursor... does nothing if loading screen not active.
-					entry.id = snd_load(&entry, gs->flags);
+					entry.id = snd_load(&entry, &gs.flags);
 				}
 			}
 		}
@@ -565,12 +565,12 @@ void gamesnd_load_gameplay_sounds()
 		return;
 
 	Assert( Snds.size() <= INT_MAX );
-	for (SCP_vector<game_snd>::iterator gs = Snds.begin(); gs != Snds.end(); ++gs) {
-		if ( !gs->preload ) { // don't try to load anything that's already preloaded
-			for (auto& entry : gs->sound_entries) {
+	for (auto& gs: Snds) {
+		if ( !gs.preload ) { // don't try to load anything that's already preloaded
+			for (auto& entry : gs.sound_entries) {
 				if (entry.filename[0] != 0 && strnicmp(entry.filename, NOX("none.wav"), 4) != 0) {
 					game_busy(NOX("** preloading gameplay sounds **"));        // Animate loading cursor... does nothing if loading screen not active.
-					entry.id = snd_load(&entry, gs->flags);
+					entry.id = snd_load(&entry, &gs.flags);
 				}
 			}
 		}
@@ -583,8 +583,8 @@ void gamesnd_load_gameplay_sounds()
 void gamesnd_unload_gameplay_sounds()
 {
 	Assert( Snds.size() <= INT_MAX );
-	for (SCP_vector<game_snd>::iterator gs = Snds.begin(); gs != Snds.end(); ++gs) {
-		for (auto& entry : gs->sound_entries) {
+	for (auto& gs: Snds) {
+		for (auto& entry : gs.sound_entries) {
 			if (entry.id.isValid()) {
 				snd_unload(entry.id);
 				entry.id = sound_load_id::invalid();
@@ -602,10 +602,10 @@ void gamesnd_load_interface_sounds()
 		return;
 
 	Assert( Snds_iface.size() < INT_MAX );
-	for (SCP_vector<game_snd>::iterator si = Snds_iface.begin(); si != Snds_iface.end(); ++si) {
-		for (auto& entry : si->sound_entries) {
+	for (auto& gs: Snds) {
+		for (auto& entry : gs.sound_entries) {
 			if ( entry.filename[0] != 0 && strnicmp(entry.filename, NOX("none.wav"), 4) != 0 ) {
-				entry.id = snd_load(&entry, si->flags);
+				entry.id = snd_load(&entry, &gs.flags);
 			}
 		}
 	}
@@ -1349,7 +1349,7 @@ float gamesnd_get_max_duration(game_snd* gs) {
 	for (auto& entry : gs->sound_entries) {
 		if (!entry.id.isValid()) {
 			// Lazily load unloaded sound entries when required
-			entry.id = snd_load(&entry, gs->flags);
+			entry.id = snd_load(&entry, &gs->flags);
 		}
 
 		max_length = std::max(max_length, snd_get_duration(entry.id));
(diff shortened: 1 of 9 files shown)


code/sound/sound.cpp
```diff
@@ -655,20 +655,26 @@ sound_handle snd_play_3d(game_snd* gs, vec3d* source_pos, vec3d* listen_pos, flo
 		Int3();
 		return sound_handle::invalid();
 	}
+	if (gs->flags & GAME_SND_NOT_VALID) {
```
jg18, line 658: This check should be added to `snd_play()` and `snd_play_looping()` as well, correct?

I assume Issue #3388 is not limited to 3D sounds?

code/sound/sound.cpp
```diff
@@ -655,20 +655,26 @@ sound_handle snd_play_3d(game_snd* gs, vec3d* source_pos, vec3d* listen_pos, flo
...
-		MONITOR_INC( Num3DSoundsLoaded, 1 );
+		MONITOR_INC(Num3DSoundsLoaded, 1);
 	} else if (entry->id_sig != Sounds[entry->id.value()].sig) {
 		entry->id = snd_load(entry, gs->flags);
 	}
 
-	if (!entry->id.isValid())
+	if (!entry->id.isValid()) {
+		Warning(LOCATION, "Failed to load one or more sounds for gamesnd %s!", gs->name.c_str());
+		gs->flags |= GAME_SND_NOT_VALID;
```
jg18, line 675: There is similar code for attempting to load the sound in `snd_play()` and `snd_play_looping()`. We should add the `GAME_SND_NOT_VALID` flag there if needed as well, correct?

In fact, the code fragment starting with `auto entry =` a few lines above is nearly identical in all three `snd_play*()` functions, except that a couple of them call `MONITOR_INC()`.

Perhaps this common code fragment could be refactored into a new function, say
```
game_snd_entry* snd_try_load(game_snd* gs, bool 3d, bool looping)`
```
(or use an `enum` to indicate sound type of 2D vs. 3D vs. 2D looping)

That should ensure that we have to set `GAME_SND_NOT_VALID` in only one place in the code.

jg18 (changes requested):
Unless there's something I'm missing, the changes to `snd_play_3d()` should apply to `snd_play()` and `snd_play_looping()` as well.

jg18:
Also,` snd_load()`, which is called in a number of places, should probably check
```
	if (flags & GAME_SND_NOT_VALID)
		return sound_load_id::invalid();
```

jg18:
Another potential solution is to modify `snd_load()` to take `game_snd* gs` instead of `int flags` (pass `nullptr` in existing calls that pass `0`) and then set `GAME_SND_NOT_VALID` if loading the sound fails, assuming `gs` is non-null of course. This would result in fewer code changes. I see only two places in snd_load() where you would need to set it.

code/sound/sound.cpp
Goober5000: Good point.  I don't know if #3388 is limited to 3D sounds or not, but it makes sense that other sounds would behave similarly.

code/sound/sound.cpp
Goober5000: I've added similar code to the other places, and tweaked them a bit.  I don't really feel like refactoring things now though, and one could make the case to keep this PR simple since it's a bugfix.

jg18: Thanks, and fair point about not refactoring here. The code duplication it would eliminate wouldn't be much, anyway.

jg18 (changes requested):
Good progress, thanks.

Another request: `snd_load()`, which is called in a number of places in the codebase, should also check for `flags & GAME_SND_NOT_VALID`.

Goober5000:
That's starting to make the PR more complicated, but ok.

jg18:
Oh, I thought we were sticking with your original solution. The only remaining change I was looking for was adding the check for `flags & GAME_SND_NOT_VALID` in `snd_load()`.

The alternate solution was just an idea that I put out there for discussion. Agreed that it ends up making things more complicated.

I'm ok with going back to your original solution and just adding that check in `snd_load()`. Does that sound good?

Goober5000:
Well, setting the flag in `snd_load` does make a certain kind of sense.  I'm ok with keeping the PR the way it is right now.

code/mission/missionmessage.cpp
```diff
@@ -1008,9 +1008,9 @@ void message_load_wave(int index, const char *filename)
 		return;
 	}
 
-	game_snd_entry tmp_gs;
-	strcpy_s( tmp_gs.filename, filename );
-	Message_waves[index].num = snd_load( &tmp_gs, 0, 0 );
+	game_snd_entry tmp_gse;
```
jg18, line 1011: Nice catch. :+1:

code/sound/sound.cpp
```diff
@@ -300,14 +300,15 @@ static std::unique_ptr<sound::IAudioFile> openAudioFile(const char* fileName)
 // a single instance, and can be played multiple times simultaneously.  Through the magic of
 // DirectSound, only 1 copy of the sound is used.
 //
-// parameters:		gs							=> file of sound to load
+// parameters:		entry							=> entry of sound to load
+// parameters:		flags							=> flags of sound to load
```
jg18, line 304: Not a blocker, but could you update this comment to "pointer to flags of sound to load (can be null)" or similar?

code/sound/sound.cpp
```diff
@@ -564,24 +572,33 @@ sound_handle snd_play(game_snd* gs, float pan, float vol_scale, int priority, bo
 	if (!Sound_enabled)
 		return sound_handle::invalid();
 
-	if (gs == NULL) {
-		Int3();
+	if (!ds_initialized)
```
jg18, line 575: Moving this check up here is also a nice catch. :+1:

code/gamesnd/gamesnd.cpp
```diff
@@ -544,12 +544,12 @@ void gamesnd_preload_common_sounds()
...
-	for (SCP_vector<game_snd>::iterator gs = Snds.begin(); gs != Snds.end(); ++gs) {
-		if ( gs->preload ) {
-			for (auto& entry : gs->sound_entries) {
+	for (auto& gs: Snds) {
+		if ( gs.preload ) {
+			for (auto& entry : gs.sound_entries) {
 				if ( entry.filename[0] != 0 && strnicmp(entry.filename, NOX("none.wav"), 4) != 0 ) {
 					game_busy( NOX("** preloading common game sounds **") );	// Animate loading cursor... does nothing if loading screen not active.
-					entry.id = snd_load(&entry, gs->flags);
+					entry.id = snd_load(&entry, &gs.flags);
```
jg18, line 552: In theory, we could break out of this loop if the `GAME_SND_NOT_VALID` flag was just set on `gs`, but that'd probably be unnecessarily making things more complicated. Trying to be clever is when I tend to get myself in trouble. :wink:

code/sound/sound.cpp
```diff
@@ -354,13 +358,15 @@ sound_load_id snd_load(game_snd_entry* entry, int flags, int /*allow_hardware_lo
 	std::unique_ptr<sound::IAudioFile> audio_file = openAudioFile(entry->filename);
 
 	if (audio_file == nullptr) {
+		if (flags)
+			*flags |= GAME_SND_NOT_VALID;
 		return sound_load_id::invalid();
```
jg18, line 363: Probably out of scope, but should there be an `nprintf()`, like there is below, that creating the audio file object failed?

code/object/objectsnd.cpp
```diff
@@ -483,6 +483,9 @@ void obj_snd_do_frame()
 		}
 
 		gs = gamesnd_get_game_sound(osp->id);
```
jg18, line 485: Would a `nullptr` check with `UNREACHABLE()` be worthwhile?

It's weird that `gs` is retrieved here but not accessed for another ~100 lines. 🤷‍♂️

jg18 (approved):
Looks good. Added a few final thoughts.

code/sound/sound.cpp
Goober5000: Ya

code/sound/sound.cpp
Goober5000: There is, but it's deeper down in `audio_file->Open`.

code/object/objectsnd.cpp
Goober5000: I don't think a `nullptr` check is warranted here, given that the underlying operation should never return null.

I like that `gs` is retrieved and checked here; if it's invalid, a lot of work is skipped.

PR #3641, opened 2021-09-04
fix nebula range in mission-set-nebula

### LINKED ISSUE #3619: Nebula mission vs toggle-nebula SEXP targeting issues
When you specify a mission as a Nebula mission in the Background editor, an AWACs is not required in-mission to target hostiles. It uses some % of the range specified in the background editor.

If you use a sexp to toggle nebula on mid-mission it seems like you must have an AWACs present or you can't target hostiles. I'm aware there's a sexp to set AWACs distance but it's arguments require you to specify a ship first.

Goober5000:
The mission-set-nebula sexp never initialized the nebula range, which is usually done in the background editor.  This PR allows the range to be specified, and sets it to 3000 (the same default as in the background editor) if not specified.

Fixes #3619.

### DIFF
--- a/code/starfield/starfield.cpp
+++ b/code/starfield/starfield.cpp
@@ -2530,6 +2530,14 @@ void stars_set_nebula(bool activate, float range)
 		Neb2_awacs = range;
 		if (Neb2_awacs < 0.1f)
 			Neb2_awacs = 3000.0f;	// this is also the default in the background editor
+
+		// this function is not currently called in FRED, but this would be needed for FRED support
+		if (Fred_running)
+		{
+			Neb2_render_mode = NEB2_RENDER_POF;
+			stars_set_background_model(BACKGROUND_MODEL_FILENAME, Neb2_texture_name);
+			stars_set_background_orientation();
+		}
 	}
 	else
 	{
--- a/code/starfield/starfield.cpp
+++ b/code/starfield/starfield.cpp
@@ -2517,7 +2517,7 @@ const char *stars_get_name_from_instance(int index, bool is_a_sun)
 }
 
 // WMC/Goober5000
-void stars_set_nebula(bool activate)
+void stars_set_nebula(bool activate, float range)
 {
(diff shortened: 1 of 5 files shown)


code/starfield/starfield.cpp
```diff
@@ -2517,27 +2517,27 @@ const char *stars_get_name_from_instance(int index, bool is_a_sun)
...
-			stars_set_background_model(BACKGROUND_MODEL_FILENAME, Neb2_texture_name);
-			stars_set_background_orientation();
-		} else {
-			Neb2_render_mode = NEB2_RENDER_HTL;
-		}
+
+		Neb2_render_mode = NEB2_RENDER_HTL;
+		Neb2_awacs = range;
+		if (Neb2_awacs < 0.1f)
+			Neb2_awacs = 3000.0f;
```
jg18, line 2532: Can you add a comment on where this number comes from?

code/starfield/starfield.cpp
```diff
@@ -2517,27 +2517,27 @@ const char *stars_get_name_from_instance(int index, bool is_a_sun)
...
-void stars_set_nebula(bool activate)
+void stars_set_nebula(bool activate, float range)
 {
     The_mission.flags.set(Mission::Mission_Flags::Fullneb, activate);
 	
     if (activate)
 	{
 		Toggle_text_alpha = TOGGLE_TEXT_NEBULA_ALPHA;
 		HUD_contrast = 1;
-		if(Fred_running) {
```
jg18, line 2528: I'm unclear on why special treatment while FRED is running is no longer needed?

jg18 (approved):
A couple small things but seems reasonable overall.

code/starfield/starfield.cpp
Goober5000: Done

code/starfield/starfield.cpp
Goober5000: Because the function is never called from FRED.  I suppose it doesn't hurt to put this back in, though, if that were ever changed.

PR #3645, opened 2021-09-09
correct shipname return type

Goober5000:
The return type of "shipname" is a string, so if the ship can't be found, return an empty string instead of an invalid ship.


TRBlount (approved):
Good catch. LGTM

PR #3646, opened 2021-09-11
Fix Turrets in Lab

BMagnu:
Fixes a regression introduced by #3636.
Due to the turrets now doing actual submodel processing, they'll try to acquire targets, which CTD's due to the asteroid list being uninitialized (and not just empty).


Baezon (approved):
"The function is perfectly capable of determining on its own that there are no valid targets..." MY ASS

The-E (approved):

PR #3647, opened 2021-09-12
add hide-in-mission-log ship flag

Goober5000:
This flag causes any events involving this ship to not be viewable in the mission log.  They are still logged, though, so sexps like `is-destroyed-delay` will still work.


Baezon (approved):
LGTM

PR #3648, opened 2021-09-13
add some SEXP documentation

### LINKED ISSUE #3612: FR: FRED Documentation for ai-ignore-new
Could information be added to the in-fred documentation for both ai-ignore and ai-ignore-new to explain why ai-ignore-new exists and when you should use it, or something like that? I'm sure there are sexps that could use this too

Goober5000:
Per request, add some more documentation to ai-ignore and ai-ignore-new (and ai-play-dead while we're at it).

Fixes #3612.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -115,10 +115,10 @@
 #undef MessageBox
 #endif
 
-
-
-#define TRUE	1
-#define FALSE	0
+// these are useful for embedding numbers in SEXP help strings
+// see https://stackoverflow.com/questions/5459868/concatenate-int-to-string-using-c-preprocessor
+#define STR_HELPER(x) #x
+#define STR(x) STR_HELPER(x)
 
 
 SCP_vector<sexp_oper> Operators = {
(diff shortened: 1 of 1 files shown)


EatThePath:
These look like they satisfyingly address my request

code/parse/sexp.cpp
```diff
@@ -33154,19 +33154,23 @@ SCP_vector<sexp_help_struct> Sexp_help = {
...
-		"\t2:\tGoal priority (number between 0 and 89)." },
+		"\t1:\tName of target to ignore.\r\n"
+		"\t2:\tGoal priority (number between 0 and 89) - note, this does not imply any ranking of ignored targets." },
 
 	// Goober5000
 	{ OP_AI_IGNORE_NEW, "Ai-ignore-new (Ship goal)\r\n"
-		"\tTells the specified ship to ignore the given ship and not consider it as a valid "
-		"target to attack.\r\n\r\n"
+		"\tTells the specified ship to ignore the given target and not consider it when picking something "
+		"to attack.  Up to MAX_IGNORE_NEW_OBJECTS targets can be ignored at a time.  Targets ignored by "
```
BMagnu, line 33169: MAX_IGNORE_NEW_OBJECTS is going to be arcane magic to any non-coder reading this.
Should we just not expand this here, and show the user the actual number behind this define?

Goober5000: The problem is you can't concatenate an int to a string here.  Or, at least, I thought you couldn't.  I just found a way and will push it momentarily.

BMagnu (approved):

PR #3652, opened 2021-09-14
Anti-glint shader trick to reduce glimmer on distant objects

qazwsxal:
This is based off a GDC talk given by valve in 2015: http://media.steampowered.com/apps/valve/2015/Alex_Vlachos_Advanced_VR_Rendering_GDC2015.pdf (Page 43)
This decreases the glossiness of surfaces if the derivatives of the normal buffer change too quickly in respect to screen space.

Example of tweak in use:
https://www.youtube.com/watch?v=YjyqpeVCY9I
Default settings, no AA.


BMagnu (approved):
Seems reasonable enough.
IMO it can go in like this with no flags or anything attached.

The-E (approved):
Yeah, I think this qualifies as a straight bugfix

qazwsxal:
pog

PR #3656, opened 2021-09-14
Scale distant combat related effects so they stay visible

brihernandez:
Left is with these changes enabled, right is current behavior.

While playing through Freespace 1, I noticed that distant lasers remained visible despite being very far away. The below battle is taking place ~7km away from the player. This far enough that the none of the individual ships can be seen, but it's still obvious a battle is happening due to the visible fire.

In Freespace 2, this does not happen anymore. It's been a pet peeve of mine for a long while. This implementation scales the following to ensure they are always visible in the case of point objects, and visible enough for lines to prevent stair-stepping.
- **Thrusters**: When ships are flying away from you, thrusters are visible as bright points in the distance. When facing you, they are no longer visible. This is exactly the same behavior thrusters currently have, it's only visible from a distance now.
- **Lasers**: Self-explanatory. Lasers are scaled such that if they become smaller than a few pixels, they maintain their size and remain visible and do not flicker from subpixels not rasterizing nicely.
- **Muzzle flashes**: Similar behavior to lasers.
- **Trails**: This applies to *all* trails. This primarily affects missiles, but can also affect ship contrails and thruster trails as are used in the Freespace Upgrade Media VPs. Trails are scaled enough to prevent stair-stepping.
- **Beams**: Similar to trails, they are scaled enough to prevent stair-stepping.

Exactly how much scaling can be adjusted using the `game_settings.tbl` or a `***-mod.tbm`. [Attached are example values that I think are ideal for FS2 retail assets](https://github.com/scp-fs2open/fs2open.github.com/files/7165600/tables.zip). Big enough to prevent flickering and stair stepping of beams, but only just so.

```
#GRAPHICS SETTINGS

$Minimum Pixel Size Thrusters: 1.5
$Minimum Pixel Size Beams: 5.0
$Minimum Pixel Size Muzzle Flashes: 4.0
$Minimum Pixel Size Trails: 3.0
$Minimum Pixel Size Lasers: 2.0

#END
```

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -74,6 +74,11 @@ bool Neb_affects_fireballs;
 std::tuple<float, float, float, float> Shadow_distances;
 std::tuple<float, float, float, float> Shadow_distances_cockpit;
 bool Custom_briefing_icons_always_override_standard_icons;
+float Min_pixel_size_thruster;
+float Min_pixel_size_beam;
+float Min_pizel_size_muzzleflash;
+float Min_pixel_size_trail;
+float Min_pixel_size_laser;
 
 SCP_vector<std::pair<SCP_string, gr_capability>> req_render_ext_pairs = {
 	std::make_pair("BPTC Texture Compression", CAPABILITY_BPTC)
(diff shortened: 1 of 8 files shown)


Baezon (changes requested):
Looks pretty good, I would add quick comments everytime you apply the scaling, just so cursory viewers don't get too confused. Like `// maybe apply size scaling if the camera is very far away`

PR #3660, opened 2021-09-16
LUA: Add iterator over ships in mission

BMagnu:
This makes a loop over all ships much faster in lua, as before, each access to mn.Ships[i] in lua required iterating the Objects array, while the iterator will provide the next element of the ship_objects list in constant time.

### DIFF
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1520,6 +1520,8 @@ ADE_FUNC(getShipList,
 	ship_obj* so = &Ship_obj_list;
 
 	return ade_set_args(L, "u", luacpp::LuaFunction::createFromStdFunction(L, [so](lua_State* LInner, const luacpp::LuaValueList& /*params*/) mutable -> luacpp::LuaValueList {
+		//Since the first element of a list is the next element from the head, and we start this function with the the captured "so" object being the head, this GET_NEXT will return the first element on first call of this lambda.
+		//Similarly, an empty list is defined by the head's next element being itself, hence an empty list will immediately return nil just fine
 		so = GET_NEXT(so);
 
 		if (so == END_OF_LIST(&Ship_obj_list)) {
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1510,6 +1510,26 @@ ADE_FUNC(getArrivalList,
 	                    l_ParseObject.Set(parse_object_h(&Ship_arrival_list)));
 }
 
+ADE_FUNC(getShipList,
+	l_Mission,
+	nullptr,
+	"Get an iterator to the list of ships in this mission",
+	"iterator<ship>",
+	"An iterator across all ships in the mission. Can be used in a for .. in loop")
+{
+	ship_obj* so = &Ship_obj_list;
+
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/mission.cpp
```diff
@@ -1510,6 +1510,26 @@ ADE_FUNC(getArrivalList,
...
+	l_Mission,
+	nullptr,
+	"Get an iterator to the list of ships in this mission",
+	"iterator<ship>",
+	"An iterator across all ships in the mission. Can be used in a for .. in loop")
+{
+	ship_obj* so = &Ship_obj_list;
+
+	return ade_set_args(L, "u", luacpp::LuaFunction::createFromStdFunction(L, [so](lua_State* LInner, const luacpp::LuaValueList& /*params*/) mutable -> luacpp::LuaValueList {
+		so = GET_NEXT(so);
```
Goober5000, line 1523: Add a comment explaining why `GET_FIRST` isn't used.  Initially I thought this was a bug, as all loops start by calling `GET_FIRST`, but then I realized this would work if the iterator iterated before returning the first object.

Also affirmatively indicate that this should work on empty lists, which I think is true for the same reason as above.

Goober5000 (changes requested):

Baezon (commented):
Why does this look so different from the way `getArrivalList()` is defined? Or is it like actually the same and the lua magic is blinding me?

BMagnu:
getArrivalList operates on parse_objects.
These have their next list object pointer stored inside of them. Hence that function can work by finding the next element from the last.
getShipList operates on objects. These don't have a pointer to their ship_object or that list's next object. Hence out function has to take care of remembering where it was itself. And since we need a functor and not just a function for that, it is implemented through createFromStdFunction as opposed to createFromCFunction

code/scripting/api/libs/mission.cpp
```diff
@@ -1510,6 +1510,28 @@ ADE_FUNC(getArrivalList,
...
+	l_Mission,
+	nullptr,
+	"Get an iterator to the list of ships in this mission",
+	"iterator<ship>",
+	"An iterator across all ships in the mission. Can be used in a for .. in loop")
+{
+	ship_obj* so = &Ship_obj_list;
+
+	return ade_set_args(L, "u", luacpp::LuaFunction::createFromStdFunction(L, [so](lua_State* LInner, const luacpp::LuaValueList& /*params*/) mutable -> luacpp::LuaValueList {
+		//Since the first element of a list is the next element from the head, and we start this function with the head as so, this GET_NEXT will return the first element on first call of this lambda.
```
Goober5000, line 1523: The comment is good, but reword "as so".  I know you're referring to `so` the variable, but "so" is an actual word and "as so" can be misinterpreted as a comparison.

Goober5000 (changes requested):
Comment looks good now.  However, I just realized that this iterator can stick around for multiple frames and therefore it needs to account for the possibility the ship will be removed from the list.  If this happens, `GET_NEXT` will return null.

BMagnu:
Hmm, the same problem will happen with the ``getArrivalList`` iterator. I will add to the documentation that the iterators are not stable over multiple frames, and if they encounter a nullptr as the next element, they'll just stop iterating, as a proper recovery would be hard, since we have lost our place in the list.

Goober5000 (approved):

PR #3661, opened 2021-09-17
Fix homing child weapons in niche circumstances

EatThePath:
This line, in code that runs if a seeker's range to target ever increases noticeably, removes the target_num but doesn't touch the values that are primarily used for homing. The only effect it has that I've been able to find is to break the inherit parent target weapon flag, which checks the target number.

### DIFF
--- a/code/ai/ai.h
+++ b/code/ai/ai.h
@@ -405,7 +405,6 @@ typedef struct ai_info {
 
 
 	union {
-	float		lead_scale;							//	Amount to lead current opponent by.
 	float		stay_near_distance;				//	Distance to stay within for AIM_STAY_NEAR mode.
 	};
 
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -14865,7 +14865,6 @@ void init_ai_object(int objnum)
 	aip->danger_weapon_objnum = -1;
 	aip->danger_weapon_signature = -1;
 
-	aip->lead_scale = 0.0f;
 	aip->last_hit_target_time = Missiontime;
 	aip->last_hit_time = Missiontime;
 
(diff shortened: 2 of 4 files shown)


Baezon (approved):

Goober5000 (changes requested):
One thing that *will* change is that it will cause the "If our target is still valid, then update some info." block to be run every frame, since it checks `target_num`.  Will this have any effect?

Baezon:
After some investigation, it appears that both `lead_scale` and `nearest_dist` are used by literally nothing. They are initialized elsewhere and updated by this block, and that's it. >.>

Given that this whole block might just be deleted. This is clearly not a 'lost your target' situation so it really shouldn't be setting `target_num` to -1.

EatThePath:
> 
> 
> After some investigation, it appears that both `lead_scale` and `nearest_dist` are used by literally nothing. They are initialized elsewhere and updated by this block, and that's it. >.>
> 
> Given that this whole block might just be deleted. This is clearly not a 'lost your target' situation so it really shouldn't be setting `target_num` to -1.

I can expand the pr to just clean that all out then, in that case.

Baezon:
Make sure to give it another round of testing as well, to be sure.

EatThePath:
@asarium sorry about that, sloppy keyboarding and sloppy git-fu accidentally made a spurious white-space change to a scripting file, since force-pushed out of the timeline.

EatThePath:
> 
> 
> One thing that _will_ change is that it will cause the "If our target is still valid, then update some info." block to be run every frame, since it checks `target_num`. Will this have any effect?

As Asteroth indicated, the data being updated is data only ever used in this section of the code anyway. As far as I can tell everything in the if block indicated by that comment might as well be a noop as far as the rest of the simulation is concerned, except the line I was deleting to begin with. With my last push, it's all torn out now.

Goober5000 (approved):
Ok.

Goober5000:
Let's not merge this until after 21.4 is out though.

PR #3664, opened 2021-09-19
Fix multiplayer ai crashes

### LINKED ISSUE #3663: Multiplayer AI Client Crashes
On the client AI info is not cleared out every frame, so it is the likely cause for a family of crashes.  The most common is when a player calls the support ship and then dies before their rearm is complete.

JohnAFernandez:
This theoretically fixes AI crashes from clients having invalid AI goals by clearing them once per frame.  The mostly likely of these crashes to show up is when players call the support ship and then die before the support ship is done reloading them.

Closes #3663

### DIFF
--- a/code/ai/aigoals.cpp
+++ b/code/ai/aigoals.cpp
@@ -2039,6 +2039,39 @@ void validate_mission_goals(int objnum, ai_info *aip)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


JohnAFernandez:
Pending testing.

JohnAFernandez:
Runs into other issues.  Will have to try something else before next stable.

PR #3670, opened 2021-09-22
Clean up beam type naming scheme

### LINKED ISSUE #3668: FR: validate beam type is valid beam type at table prase
as title. It's not being done currently.

Baezon:
Gets rid of the whole terribly confusing `BEAM_TYPE_A`, `...TYPE_B`, 0, 1, 2, 3, bullshit and makes everything use clearly named enums for beam types. Tables can still use numbers but will also accept string names for them, and adds a warning for an unrecognized type (which was not done before).

Fixes #3668

### DIFF
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -486,18 +486,18 @@ int beam_fire(beam_fire_info *fire_info)
 		new_item->type = BeamType::TARGETING;
 	}
 
-	// if the targeted subsystem is not NULL, force it to be a type A beam
-	if(new_item->target_subsys != nullptr && new_item->type != BeamType::ANTIFIGHTER && new_item->type != BeamType::OMNI){
+	// if the targeted subsystem is not NULL, force it to be a direct fire beam
+	if(new_item->target_subsys != nullptr && new_item->type != BeamType::TARGETING && new_item->type != BeamType::OMNI){
 		new_item->type = BeamType::DIRECT_FIRE;
 	}
 
-	// type D weapons can only fire at small ships and missiles
+	// antifighter beam weapons can only fire at small ships and missiles
 	if(new_item->type == BeamType::ANTIFIGHTER){
 		// if its a targeted ship, get the target ship
 		if((fire_info->target != NULL) && (fire_info->target->type == OBJ_SHIP) && (fire_info->target->instance >= 0)){		
 			ship *target_ship = &Ships[fire_info->target->instance];
 			
-			// maybe force to be a type A
+			// maybe force to be direct fire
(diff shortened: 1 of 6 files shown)


code/weapon/beam.cpp
```diff
@@ -482,24 +482,24 @@ int beam_fire(beam_fire_info *fire_info)
...
-	if (fire_info->bfi_flags & BFIF_IS_FIGHTER_BEAM && new_item->type != BEAM_TYPE_F) {
-		new_item->type = BEAM_TYPE_C;
+	if (fire_info->bfi_flags & BFIF_IS_FIGHTER_BEAM && new_item->type != BeamType::OMNI) {
+		new_item->type = BeamType::TARGETING;
 	}
 
 	// if the targeted subsystem is not NULL, force it to be a type A beam
-	if(new_item->target_subsys != nullptr && new_item->type != BEAM_TYPE_C && new_item->type != BEAM_TYPE_F){
-		new_item->type = BEAM_TYPE_A;
+	if(new_item->target_subsys != nullptr && new_item->type != BeamType::ANTIFIGHTER && new_item->type != BeamType::OMNI){
```
BMagnu, line 490: Shouldn't this be BeamType::TARGETING?

code/weapon/beam.cpp
```diff
@@ -2105,21 +2105,21 @@ int beam_start_firing(beam *b)
 	// any special stuff for each weapon type
 	switch(b->type){
 	// re-aim type A and D beam weapons here, otherwise they tend to miss
```
BMagnu, line 2107: This comment should be modified accordingly

code/weapon/beam.h
```diff
@@ -25,13 +25,14 @@ struct obj_pair;
...
 struct vec3d;
 
-// beam types
-#define BEAM_TYPE_A					0				// unidirectional beam
-#define BEAM_TYPE_B					1				// "slash" in one direction
-#define BEAM_TYPE_C					2				// targeting lasers (only lasts one frame)
-#define BEAM_TYPE_D					3				// similar to the type A beams, but takes multiple shots and "chases" fighters around
-#define BEAM_TYPE_E					4				// stupid beam. like type A, only it doesn't aim. it just shoots directly out of the turret
-#define BEAM_TYPE_F					5				// SCP type, highly flexible and configurable
+typedef enum class BeamType {
```
BMagnu, line 28: These definitions should really include the old letter (and possibly number) used for the beam. It's obvious which is which to us, but in the future it'll be really hard to understand. Additionally, wording like "similar to the type A beams" should also be reworded now

BMagnu (changes requested):

BMagnu (approved):

PR #3671, opened 2021-09-23
expand some sexp documentation

### LINKED ISSUE #3669: Documentation request: Hits-related SEXPs
These questions really should be answerable by looking at the in-fred SEXP documentation and, as far as I can tell, are not.
```
EatThePath: are guardian thresholds absolute or percentages of base?
Goober5000: percentages
EatThePath: how does that interact special hits? if I double a ship's health in fred, is guardian 50 going to be it's normal base health, or half that?
Asteroth: it probably will take special hits into account
      lemme check...
      yeah it does
      so its just 50% of whatever that specific ship's max hull is
EatThePath: okay... does all this apply to the hits-left sexp too?
DefCynodont: I believe so.
```

Goober5000:
Adds important information to `ship-guardian` and `show-subtitle-image`.  Fixes #3669.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -33286,7 +33286,7 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"\t1+:\tName of ships to deactivate shields on." },
 
 	{ OP_SHIP_GUARDIAN, "ship-guardian\r\n"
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
Good updates on the descriptions.

PR #3676, opened 2021-09-27
Expose nebula visibility factors

Baezon:
Exposes the previously hardcoded factors for visibility falloff of various effects at distance in nebula fog.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -15,6 +15,7 @@
 #include "mission/missionload.h"
 #include "mission/missionmessage.h"
 #include "missionui/fictionviewer.h"
+#include "nebula/neb.h"
 #include "mod_table/mod_table.h"
 #include "parse/parselo.h"
 #include "sound/sound.h"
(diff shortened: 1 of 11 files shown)


BMagnu:
While exposing these values is great, shouldn't they be set in the mission file?
Since different nebulae in a mod could presumably have different characteristics, after all...
These should possibly even be SEXP-modifiable IMO, like other nebula parameters such as poofs.

Baezon:
I guess it isn't clear, but these are global global factors of how 'strongly' an effect is affected by nebula fog, and already take into account the actual density of the fog itself. Ship hulls for instance are theoretically '1.0', this doesn't need to be changed because designers can already modify the density of fog if they want things to be more or less visible. The factors are how much less dense the fog acts like for them, since these should be 'bright' and pierce through the fog to some degree, but are otherwise supposed to be fundamental aspects of the way things look like in your fictional universe (and FSO approximations of such a thing).

I don't think they should have mission specific overrides any more than there should be mission specific overrides for making ship hull more or less glossy, or changing the relative brightness of point lights vs static light.

BMagnu:
Right, in that case it makes sense to only expose them as -mod values.

BMagnu (approved):

code/nebula/neb.h
```diff
@@ -41,16 +41,15 @@ extern float Neb2_awacs;
...
-#define NEB_FOG_VISIBILITY_MULT_TRAIL			1.0f
-#define NEB_FOG_VISIBILITY_MULT_THRUSTER		1.5f
-#define NEB_FOG_VISIBILITY_MULT_WEAPON			1.3f
-#define NEB_FOG_VISIBILITY_MULT_SHIELD			1.2f
-#define NEB_FOG_VISIBILITY_MULT_GLOWPOINT		1.2f
-#define NEB_FOG_VISIBILITY_MULT_BEAM(size)		4.0f + (size / 10)
-#define NEB_FOG_VISIBILITY_MULT_B_MUZZLE(size)  NEB_FOG_VISIBILITY_MULT_BEAM(size)
-#define NEB_FOG_VISIBILITY_MULT_PARTICLE(size)  1.0f + (size / 12)
-#define NEB_FOG_VISIBILITY_MULT_SHOCKWAVE		2.5f
-#define NEB_FOG_VISIBILITY_MULT_FIREBALL(size)	1.2f + (size / 12)
```
Goober5000, line 53: Instead of having one value that combines the offset and divisor, I would rather have two values.  It makes things clearer and is also more flexible.

Goober5000 (changes requested):
Looks good mostly, but one design change request.

code/nebula/neb.h
Baezon: You mean those `(size / 10)` values right? So you would have a constant factor and a separately controllable size factor?

Goober5000: Right, I may have worded it badly.  Visibility A and visibility B, where the factor = A + (size / B)

code/nebula/neb.cpp
```diff
@@ -113,10 +113,23 @@ float Neb2_awacs = -1.0f;
...
+float Neb2_fog_visibility_weapon = 1.3f;
+float Neb2_fog_visibility_shield = 1.2f;
+float Neb2_fog_visibility_glowpoint = 1.2f;
+float Neb2_fog_visibility_beam_const = 4.0f;
+float Neb2_fog_visibility_beam_size_factor = 0.1f;
+float Neb2_fog_visibility_particle_const = 1.0f;
+float Neb2_fog_visibility_particle_size_factor = 0.08f;
+float Neb2_fog_visibility_shockwave = 2.5f;
+float Neb2_fog_visibility_fireball_const = 1.2f;
+float Neb2_fog_visibility_fireball_size_factor = 0.08f;
```
Goober5000, line 130: .08 is not quite equal to 1/12.  You could just set the values to `(1.0f / 12.0f)`.

code/mod_table/mod_table.cpp
```diff
@@ -396,20 +397,68 @@ void parse_mod_table(const char *filename)
...
+
 		if (optional_string("$Beams affected by nebula visibility:")) {
 			stuff_boolean(&Neb_affects_beams);
+
+			if (optional_string("+Constant visibility factor:")) {
+				stuff_float(&Neb2_fog_visibility_beam_const);
+			}
+
+			if (optional_string("+Size visibility factor:")) {
+				stuff_float(&Neb2_fog_visibility_beam_size_factor);
```
Goober5000, line 420: This is acceptable, but I would prefer something like `+Visibility factor slope` and `+Visibility factor intercept` (or `offset`).  The term "constant" doesn't sit right for me here, and the slope-intercept terms have the advantage of being standard math terms that match the common `y = mx + b` formula.

Goober5000 (changes requested):

code/mod_table/mod_table.cpp
Baezon: Maybe, but I would also like people who aren't very good at understanding things mathematically to understand this as well, and I worry "slope" and "intercept" would be too confusing in this context, hopefully as opposed to "constant" (the amount you always get) and "size" (i.e. the amount that size affects it). From the descriptions I'll give it should hopefully be clear either way what its mathematical behavior should be.

Goober5000: How about `+Scaled visibility factor`?

Goober5000 (approved):
Very tardy re-review and approval

PR #3677, opened 2021-09-27
Add 'No evasion' weapons

Baezon:
Adds a flag that causes AI to ignore certain weapon and never attempt to dodge them. Useful if you're really spamming around weapons or otherwise don't want an AI to get distracted by certain kinds of weapons fire.


wookieejedi (approved):
Looks good, simple and effective.

PR #3679, opened 2021-09-29
patch up scripted kills

Goober5000:
When a ship is killed via scripting, the killer should be credited for the kill in the mission log and the correct hull percentage should be used.  Also, the script should work with either the ship or its weapon passed as the killer parameter.


wookieejedi (approved):
LGTM

PR #3683, opened 2021-09-30
Fix warnings concerning ballistic weapons

Baezon:
A ship which has primary bank capacities defined, but has no ballistic weapons should *not* be warned for. There absolutely no harm in providing those capacities anyway, and there many ways that weapons be arbitrarily swapped in, regardless of what its defaults or 'allowed' banks are, and so it is *only safer* to allow the specifications of capacities anyway.

Additionally, `ballistic_possible_for_this_ship` should also check the default weapons in addition to the allowed weapons. Allowed weapons are optional, but you should *definitely* be warned if a ship has default ballistic weapons but no capacities.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -5611,7 +5611,7 @@ DCF_BOOL( show_velocity_dot, ship_show_velocity_dot )
 static bool ballistic_possible_for_this_ship(const ship_info *sip)
 {
 	// has no weapons!
-	if (sip->allowed_bank_restricted_weapons.empty() && sip->num_primary_banks < 1)
+	if (sip->num_primary_banks < 1)
 		return false;
 
 	for (int i = 0; i < MAX_SHIP_PRIMARY_BANKS; i++)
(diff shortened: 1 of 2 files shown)


code/ship/ship.cpp
```diff
@@ -5610,15 +5610,23 @@ DCF_BOOL( show_velocity_dot, ship_show_velocity_dot )
 
 static bool ballistic_possible_for_this_ship(const ship_info *sip)
 {
-	if (sip->allowed_bank_restricted_weapons.empty())
+	// has no weapons!
+	if (sip->allowed_bank_restricted_weapons.empty() && sip->num_primary_banks < 1)
```
Goober5000, line 5614: Since `sip->allowed_bank_restricted_weapons.empty()` is now being tested below, I would venture to say you can take this out.  This will also be more robust if for whatever reason there were allowed weapons on a ship with no primaries.

Goober5000 (changes requested):

Goober5000 (approved):
Looks good.

I'd recommend squashing this PR since it has a change, revert, change, and tweak.

PR #3684, opened 2021-10-01
add OnShipDeathStarted scripting hook

Goober5000:
This occurs at the beginning of the ship death routine, similarly to OnDepartureStarted occurring at the beginning of the ship departure routine.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -410,8 +410,9 @@ MONITOR(NumShipDepartures)
 
 const std::shared_ptr<scripting::Hook> OnDepartureStartedHook = scripting::Hook::Factory(
 	"On Departure Started", "Called when a ship starts the departure process.",
-	{ 		
-		{"Ship", "ship", "The ship that has began the departure process."},
+	{
+		{"Self", "ship", "An alias for Ship."},
+		{"Ship", "ship", "The ship that has begun the departure process."},
 	});
 
 // Goober5000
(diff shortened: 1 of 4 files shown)


BMagnu:
What does this hook gain over the existing OnDeath hook, just by being run a bit earlier in the function?

Goober5000:
It allows just-in-time configuration of certain ship options: persona lament/praise messages, the death message displayed to the player, the vaporize flag, the mission log, etc.  The existing hook at the end of the function happens too late to affect any of that.

BMagnu:
I see, that makes sense.

code/mission/missionparse.cpp
```diff
@@ -7316,10 +7316,10 @@ int mission_do_departure(object *objp, bool goal_is_to_warp)
 
 	mprintf(("Entered mission_do_departure() for %s\n", shipp->ship_name));
 
-	// add scripting hook for 'On Depature Started' --wookieejedi
-	// hook is placed at the begining of this function to allow the scripter to
+	// add scripting hook for 'On Departure Started' --wookieejedi
+	// hook is placed at the beginning of this function to allow the scripter to
 	// actually have access to the ship's departure decisions before they are all executed
-	OnDepartureStartedHook->run(scripting::hook_param_list(scripting::hook_param("Ship", 'o', objp)));
+	OnDepartureStartedHook->run(scripting::hook_param_list(scripting::hook_param("Self", 'o', objp), scripting::hook_param("Ship", 'o', objp)));
```
BMagnu, line 7322: If you add a new parameter here (even though this new "Self" parameter seems identical to the ship parameter), it should be documented in the hook itself.

BMagnu (changes requested):

code/mission/missionparse.cpp
Goober5000: Good point

BMagnu (approved):

PR #3687, opened 2021-10-02
Add getColor to lua gr library, which returns the active color

Kiloku:
This function returns the current active color (as set in `gr.setColor`). 
An example use-case for this is when debugging a script that uses 2D drawing, you can save the active color in a variable, change it to draw some visual aid at a different color, then reset the color to the one you saved.

### DIFF
--- a/code/scripting/api/libs/graphics.cpp
+++ b/code/scripting/api/libs/graphics.cpp
@@ -493,6 +493,16 @@ ADE_FUNC(setColor, l_Graphics, "number Red, number Green, number Blue, [number A
 	return ADE_RETURN_NIL;
 }
 
+ADE_FUNC(getColor, l_Graphics, nullptr, "Gets the active 2D drawing color", "number, number, number, number" , "rgba color which is currently in use for 2D drawing")
+{
+	
+	if(!Gr_inited)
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/graphics.cpp
```diff
@@ -493,6 +493,16 @@ ADE_FUNC(setColor, l_Graphics, "number Red, number Green, number Blue, [number A
 	return ADE_RETURN_NIL;
 }
 
+ADE_FUNC(getColor, l_Graphics, nullptr, "Gets the active 2D drawing color", "number, number, number, number" , "rgba color which is currently in use for 2D drawing")
+{
+
```
Goober5000, line 498: I'd get rid of this newline at the top of the function but otherwise this looks good.

(I'm assuming you tested the functionality here; I haven't tried returning values from functions in this way.)

Goober5000 (approved):

code/scripting/api/libs/graphics.cpp
Kiloku: Yep, I tested the functionality by reproducing the use case I described in the PR. 

Just pushed a commit removing the newline.

PR #3691, opened 2021-10-03
Add velocity inheritance for particle effects

Baezon:
Uses the previously unused `GetVelocity()` function to return, not only the particle or object's velocity, but also a plain vector set by the initial caller of the particle source set up, such as the velocity of a weapon when it detonates. Particle Effects can then inherit a specified amount of this velocity when spawned (default 0).

The velocity of the source is on the `SourceOrigin` which a little awkward perhaps, but I'd be open to a better place for it.


wookieejedi (approved):
Ah this an extremely useful PR and will substantially help with improving the look of particle-based effects. `SourceOrigin` make sense and everything else looks good, but I'll also tag @asarium.

code/particle/effects/GenericShapeEffect.h
```diff
@@ -179,6 +185,10 @@ class GenericShapeEffect : public ParticleEffect {
 			saw_deprecated_effect_location = true;
 		}
 
+		if (optional_string("+Velocity Inherit:")) {
```
asarium, line 188: To me, this name seems a bit weird. It seems to indicate that it is a simple boolean value but instead is a factor on the inherited velocity.
Maybe something like `+Velocity Factor:` would be more appropriate? Also, you could make use of a random range here to allow for some variability if desired by the modder.

Baezon: `Parent velocity factor:` maybe?

PR #3694, opened 2021-10-03
Safety for replace-texture

Baezon:
If the original texture due to be replaced cannot be found, `final_index` is not changed from its original -1 and will write whatever its result from the new texture is into that position for the `ship_replacement_textures` array, corrupting the heap! So, uh, don't do anything if you can't find the old texture.


Goober5000 (approved):
Whoops...

PR #3704, opened 2021-10-08
Allow for arbitrary asteroid targets in missions

Baezon:
Instead of throwing asteroids at the first big ship you find in the field, the FREDder can specify an arbitrary list of ships that are the sole targets for asteroids, which disables the retail behavior. Only the engine side changes, as FRED currently no means to populate the requisite "$Asteroid targets" field in the mission file, which hopefully @Goober5000 can help with.

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -55,8 +55,6 @@ const float LARGE_DEBRIS_WEIGHT = 1.0f;
 
 int	Asteroids_enabled = 1;
 int	Num_asteroids = 0;
-int	Asteroid_throw_objnum = -1;		//	Object index of ship to throw asteroids at.
-int	Next_asteroid_throw;
 
 SCP_vector< asteroid_info > Asteroid_info;
 asteroid			Asteroids[MAX_ASTEROIDS];
@@ -72,6 +70,55 @@ float	Asteroid_icon_closeup_zoom;
 #define	ASTEROID_CHECK_WRAP_TIMESTAMP			2000	// how often an asteroid gets checked for wrapping
 #define	ASTEROID_UPDATE_COLLIDE_TIMESTAMP	2000	// how often asteroid is checked for impending collisions with escort ships
 #define	ASTEROID_MIN_COLLIDE_TIME				24		// time in seconds to check for asteroid colliding
+
+SCP_vector<SCP_string> Asteroid_target_ships;
+bool Default_asteroid_throwing_behavior = true;
+
+typedef struct asteroid_target {
+	int objnum;
+	int signature;
+	int throw_stamp;
+	int incoming_asteroids;
+} asteroid_target;
+
+// if default throwing behavior is enabled, then this should always have at most 1 target or possibly none, determined the retail way
+// if not, then this is whatever number of mission-specified ships (after they arrive, list is sanitized when they exit)
+SCP_vector<asteroid_target> Asteroid_targets;
+
+
+/**
+ * Return number of asteroids expected to collide with a ship.
+ */
+static int count_incident_asteroids(int target_objnum)
+{
+	object* asteroid_objp;
+	int		count;
+
+	count = 0;
+
+	for (asteroid_objp = GET_FIRST(&obj_used_list); asteroid_objp != END_OF_LIST(&obj_used_list); asteroid_objp = GET_NEXT(asteroid_objp)) {
+		if (asteroid_objp->type == OBJ_ASTEROID) {
+			asteroid* asp = &Asteroids[asteroid_objp->instance];
+
+			if (asp->target_objnum == target_objnum) {
+				count++;
+			}
+		}
+	}
+
+	return count;
+}
+
+// add a ship as a new asteroid target
+void asteroid_add_target(object* objp) {
+	asteroid_target new_target;
+	new_target.objnum = OBJ_INDEX(objp);
+	new_target.signature = objp->signature;
+	new_target.throw_stamp = timestamp(Random::next(500, 2000));
+	new_target.incoming_asteroids = count_incident_asteroids(OBJ_INDEX(objp)); // this *should* normally be 0
+
+	Asteroid_targets.push_back(new_target);
+}
 
 /**
  * Force updating of pair stuff for asteroid *objp.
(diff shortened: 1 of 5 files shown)


code/asteroid/asteroid.cpp
```diff
@@ -73,6 +71,55 @@ float	Asteroid_icon_closeup_zoom;
...
+	return count;
+}
+
+// add a ship as a new asteroid target
+void asteroid_add_target(object* objp) {
+	asteroid_target new_target;
+	new_target.objnum = OBJ_INDEX(objp);
+	new_target.signature = objp->signature;
+	new_target.throw_stamp = timestamp(Random::next(500, 2000));
+	new_target.incoming_asteroids = count_incident_asteroids(OBJ_INDEX(objp)); // this *should* normally be 0
```
JohnAFernandez, line 118: If this isn't zero, would it point to any undesired behavior?  I'm guessing probably not.  This would allow a target to be added and removed from the list without miscounting the incomings.

code/asteroid/asteroid.cpp
```diff
@@ -159,42 +206,27 @@ static float asteroid_cap_speed(int asteroid_info_index, float speed)
...
+		(pos->xyz.y > asfieldp->inner_min_bound.xyz.y - delta) && (pos->xyz.y < asfieldp->inner_max_bound.xyz.y + delta) &&
+		(pos->xyz.z > asfieldp->inner_min_bound.xyz.z - delta) && (pos->xyz.z < asfieldp->inner_max_bound.xyz.z + delta);
 }
 
-/**
- * Check if asteroid is within inner bound
- * @return 0 if not inside or no inner bound, 1 if inside inner bound
- */
-static int asteroid_in_inner_bound(asteroid_field *asfieldp, vec3d *pos, float delta) {
+static bool asteroid_ship_is_inside_field(asteroid_field* asfieldp, vec3d* pos, float radius) {
```
JohnAFernandez, line 222: Maybe `asteroid_is_ship_inside_field` is a little better a name?

code/asteroid/asteroid.cpp
```diff
@@ -159,42 +206,27 @@ static float asteroid_cap_speed(int asteroid_info_index, float speed)
...
 
-	if (!asfieldp->has_inner_bound) {
-		return 0;
-	}
+	radius *= 2.0f;
 
-	return (asteroid_in_inner_bound_with_axes(asfieldp, pos, delta) == 7);
+	return pos->xyz.x + radius > Asteroid_field.min_bound.xyz.x && pos->xyz.x - radius < Asteroid_field.max_bound.xyz.x&&
+		   pos->xyz.y + radius > Asteroid_field.min_bound.xyz.y && pos->xyz.y - radius < Asteroid_field.max_bound.xyz.y&&
```
JohnAFernandez, line 227: So far the only thing that I'm pretty sure should be changed. At the end of these lines there really should be a space after and x and y for readability.

code/asteroid/asteroid.cpp
```diff
@@ -733,52 +766,56 @@ static void asteroid_aim_at_target(object *objp, object *asteroid_objp, float de
...
+		if (target_objp->type == OBJ_NONE || target_objp->signature != Asteroid_targets[i].signature) {
+			Asteroid_targets[i] = Asteroid_targets.back();
+			Asteroid_targets.pop_back();
+		} else  // if we needed to cull we should not advance because we just moved a new asteroid target into this spot
+			i++;
+	}
+
+	for (asteroid_target& target : Asteroid_targets) {
+		if (!timestamp_elapsed(target.throw_stamp))
+			return;
```
JohnAFernandez, line 786: So if the first target's timestamp is not elapsed, no asteroid should be thrown this frame at all?

JohnAFernandez (changes requested):
Only one small change needed, and some questions.

code/asteroid/asteroid.cpp
Baezon: It should handle it correctly yeah, although currently as written, these targets can't be removed or added willy-nilly so in practice, at least for now, I would always expect 0.

code/asteroid/asteroid.cpp
Baezon: Whoops! should be `continue`

JohnAFernandez: istockphoto-183826186-1024x1024

JohnAFernandez (approved):
Approval given, but apparently Goober did want to take a look too.

code/mission/missionparse.cpp
```diff
@@ -2445,6 +2445,15 @@ int parse_create_object_sub(p_object *p_objp)
 		}
 	}
 
+	// if this is an asteroid target, add it to the list
+	if (!Asteroid_target_ships.empty()) {
```
Goober5000, line 2449: This `if` is redundant given that you have a loop right after it.

code/mission/missionparse.cpp
```diff
@@ -2445,6 +2445,15 @@ int parse_create_object_sub(p_object *p_objp)
...
+		for (SCP_string& name : Asteroid_target_ships) {
+			if (strcmp(name.c_str(), shipp->ship_name) == 0) {
```
Goober5000, line 2451: I would use `stricmp` since that is typically used elsewhere in parsing situations.

code/mission/missionparse.cpp
```diff
@@ -2445,6 +2445,15 @@ int parse_create_object_sub(p_object *p_objp)
...
+				asteroid_add_target(&Objects[objnum]);
```
Goober5000, line 2452: You can break out of the loop at this point.

code/mission/missionparse.cpp
```diff
@@ -5686,6 +5695,11 @@ void parse_asteroid_fields(mission *pm)
 		} else {
 			Asteroid_field.has_inner_bound = 0;
 		}
+
+		if (optional_string("$Asteroid Targets:")) {
+			stuff_string_list(Asteroid_target_ships);
+			Default_asteroid_throwing_behavior = false;
+		}
```
Goober5000, line 5702: Add the equivalent saving code to missionsave.cpp (both FRED and qtFRED versions) so that if you load and save a mission, the asteroid target data won't be wiped.  This will allow missions to use the feature in a stable way, even without UI support.

code/asteroid/asteroid.cpp
```diff
@@ -733,52 +766,57 @@ static void asteroid_aim_at_target(object *objp, object *asteroid_objp, float de
...
-	}
+	//sanitize the list
+	for (size_t i = 0; i < Asteroid_targets.size();) {
+		object* target_objp = &Objects[Asteroid_targets[i].objnum];
+		if (target_objp->type == OBJ_NONE || target_objp->signature != Asteroid_targets[i].signature) {
+			Asteroid_targets[i] = Asteroid_targets.back();
+			Asteroid_targets.pop_back();
+		} else  // if we needed to cull we should not advance because we just moved a new asteroid target into this spot
+			i++;
+	}
```
Goober5000, line 782: Methinks the list sanitizing should be in its own function, called from `asteroid_frame`.

code/asteroid/asteroid.cpp
```diff
@@ -733,52 +766,57 @@ static void asteroid_aim_at_target(object *objp, object *asteroid_objp, float de
...
-		object *A = &Objects[so->objnum];
-		if (so->objnum == Asteroid_throw_objnum) {
-			int subtype = Random::next(NUM_DEBRIS_POFS);
-			while (Asteroid_field.field_debris_type[subtype] == -1) {
-				subtype = (subtype + 1) % NUM_DEBRIS_POFS;
-			}
-			object *objp = asteroid_create(&Asteroid_field, ASTEROID_TYPE_LARGE, subtype);
-			if (objp != NULL) {
-				asteroid_aim_at_target(A, objp, ASTEROID_MIN_COLLIDE_TIME + frand() * 20.0f);
+		while (Asteroid_field.field_debris_type[subtype] == -1)
```
Goober5000, line 802: Is there a sanity check somewhere to make sure there is at least one asteroid subtype that isn't -1?

Goober5000 (changes requested):
This looks really good.  But I have some comments.

code/asteroid/asteroid.cpp
Baezon: It appears FRED will not let you leave the dialog without at least one selected, but still, better safe than sorry...

code/asteroid/asteroid.cpp
```diff
@@ -597,11 +597,13 @@ void asteroid_create_all()
...
-			while (Asteroid_field.field_debris_type[subtype] == -1) {
-				subtype = (subtype + 1) % NUM_DEBRIS_POFS;
+			for (int i = 0; i < NUM_DEBRIS_POFS; i++) {
+				if (Asteroid_field.field_debris_type[subtype] == -1)
+					subtype = (subtype + 1) % NUM_DEBRIS_POFS;
 			}
 
-			asteroid_create(&Asteroid_field, ASTEROID_TYPE_LARGE, subtype);
+			if (subtype >= 0)
+				asteroid_create(&Asteroid_field, ASTEROID_TYPE_LARGE, subtype);
```
Goober5000, line 606: Add an `else` branch with a warning.  Modders should be notified if they create an asteroid field without valid asteroids.   (And break out of the `max_asteroids` loop.)

code/asteroid/asteroid.cpp
```diff
@@ -798,9 +806,14 @@ static void maybe_throw_asteroid()
 		target.throw_stamp = timestamp(1000 + 1200 * target.incoming_asteroids /(Game_skill_level+1));
 
 		int subtype = Random::next(NUM_DEBRIS_POFS);
```
Goober5000, line 808: Hmm.  If there are fewer than `NUM_DEBRIS_POFS`, this method of choosing an asteroid will bias towards choosing the first one.  For 3 max POFs and only two types, the first type will be chosen 2/3rds of the time.  Can you write a new function to just count the number of valid subtypes and choose a random index from that count?  Use it both here and in the other location.

Goober5000 (changes requested):

code/asteroid/asteroid.cpp
Baezon: I put it earlier as an early return, since it's easy to explicitly catch now.

code/mission/missionparse.cpp
Baezon: I've haven't done FRED mission saving stuff before, and I can't currently compile either so please make sure I did this correctly.

code/asteroid/asteroid.cpp
```diff
@@ -733,52 +780,75 @@ static void asteroid_aim_at_target(object *objp, object *asteroid_objp, float de
...
+		int counter = Random::next(Asteroid_field.num_used_field_debris_types);
+		int subtype = -1;
+		for (int i = 0; i < NUM_DEBRIS_POFS; i++) {
+			if (Asteroid_field.field_debris_type[i] >= 0) {
+				if (counter == 0) {
+					subtype = i;
+					break;
+				}
+				else
+					counter--;
```
Goober5000, line 829: I was thinking that this selection logic could be put in its own function, but since it's only used in two places (remember the "rule of three") this is probably ok.

fred2/missionsave.cpp
```diff
@@ -768,6 +768,21 @@ int CFred_mission_save::save_asteroid_fields()
...
+			} else {
+				fout("\n$Asteroid Targets: (");
+			}
+
+			for (SCP_string& name : Asteroid_target_ships) {				
+				fout(" %s", name.c_str());
+			}
+
+			fout(" )");
+		}
```
Goober5000, line 784: Add a `fso_comment_push(";;FSO 22.0.0;;");` to the top of this block, and change `fout("\n$Asteroid Targets: (");` to use `fout_version`.

Also put a `fso_comment_pop()` at the bottom of this block.

qtfred/src/mission/missionsave.cpp
```diff
@@ -783,6 +783,21 @@ int CFred_mission_save::save_asteroid_fields()
...
+			} else {
+				fout("\n$Asteroid Targets: (");
+			}
+
+			for (SCP_string& name : Asteroid_target_ships) {				
+				fout(" %s", name.c_str());
+			}
+
+			fout(" )");
+		}
```
Goober5000, line 799: Add the same changes as above.

Goober5000 (changes requested):
Looks good, just one optional thing and a saving fix in two places.

Goober5000 (approved):
Looks good

PR #3705, opened 2021-10-08
Add filename-hash-mapped CFile cache

BMagnu:
This utilizes the already existing (but seemingly unused) cfile index to provide a mapping of CF Pathtypes and filenames to cf_file objects to massively speed up file discovery.

Previously, at least on windows, the repeated calls of ``_findfirst(longname, &findstruct);`` were a major slowdown for file loading, which is now taken over by a single map lookup. In subjective tests, this reduces microstutters ingame, for example when particles are loaded in-mission.

### DIFF
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -107,10 +107,15 @@ static const char *Cfile_cdrom_dir = NULL;
 
 CFileLocation::CFileLocation(const cf_file& file) {
 	found = true;
-	if(file.name_ext)
-		name_ext = file.name_ext;
+	name_ext = file.name_ext;
 	if (file.real_name)
 		full_name = file.real_name;
+	else {
+		//In Memory files need their full name generated
+		full_name = Pathtypes[file.pathtype_index].path;
+		full_name += DIR_SEPARATOR_STR;
+		full_name += name_ext;
+	}
 	size = file.size;
 	offset = file.pack_offset;
 	data_ptr = file.data;
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -105,6 +105,17 @@ std::array<CFILE, MAX_CFILE_BLOCKS> Cfile_block_list;
 
 static const char *Cfile_cdrom_dir = NULL;
 
+CFileLocation::CFileLocation(const cf_file& file) {
+	found = true;
+	if(file.name_ext)
+		name_ext = file.name_ext;
+	if (file.real_name)
+		full_name = file.real_name;
+	size = file.size;
+	offset = file.pack_offset;
+	data_ptr = file.data;
+}
+
 //
 // Function prototypes for internally-called functions
 //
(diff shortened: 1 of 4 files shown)


BMagnu:
Right, I can tidy up the code to remove some of the duplication, as well as add protection for file updates.
However I still believe that having a filename-based map would be beneficial over looping over the file index, especially with some mods approaching the maximum file limit.
Additionally, I believe modernizing this section in a followup PR to this, building on top of the map cache introduced, I can get rid of the fixed-size array and thus the limit of files FSO can handle (which is already being met for some mods).
Maybe this could also lead to general tidying of the CFile code.

BMagnu:
After some more detailed reading of the whole CFile code, I've come to the conclusion that the ``cf_find_file_location`` method does not have special protections against changing filesystems that would be endangered by this cache: As long as the file has the same absolute path, which it must have, given the combination of the set filename and searching the folder structure filetype by filetype, the same file will always be loaded, both before and after this PR. This method does not consider the time a file has been modified at all, and shouldn't as well.
This PR does not change the behaviour of the methods that do so, like ``cf_get_file_list``.
The only duplication of behvaiour remaining is the return path at the end of the function, if no localized file version has been found in the index.
Yes, both the index and the filename map contain similar data, but the access by hashed maps is much better than potentially 2^16 string compares to find a file, hence this addition still makes sense.
Eventually this should be condensed into just the map, getting rid of the file index and slow search altogether, fully replacing it with the map, however that is out of scope for now, as this PR mainly intends to create an acceptable performance of the main cfile bottleneck in mission (and that it does sufficiently, IMO).

wookieejedi:
In my tests there was definitely a noticeable improvement, made combat in missions more enjoyable, too.

BMagnu:
Well, if a file gets added and is not found in the named cache, it just falls back to the old slow search.
Only if a file gets deleted this might be a problem, but if that is actually a case that can happen, then the cache-by-index with which localized files or files not found with fopen are loaded at the end of the function will produce exactly the same problem.

> Out of curiosity, have you tried simply removing those _findfirst() blocks, with no other changes, and seeing if there is any performance change? 

It makes it almost worse, subjectively in terms of microstutters. fopen itself is not too quick on windows and becomes the major bottleneck on my system during the lagspikes as per my profiler (in fact, it does so too _with_ the cache, but since it only has to run once with the cache to actually open the file instead of a few times per file lookup, it's much smoother with the cache).

notimaginative:
Multiple optimizations and fixes have been made since this PR was introduced. Some of which specifically target the stuttering during a mission, which seems to now be largely resolved. So does this PR still have a noticeable performance impact if rebased against the current code base?

And to answer your question that I appear to have missed previously, I use both a HDD and SSD. The main game files are on the HDD and my home directory and OS is a SSD. So some file loading slowdowns would be hidden by the SSD when it looks for files in the user path.

wookieejedi:
I'll also add that when I tested this with a mod with multiple dependencies it did not load files in the expected order (hence conversion to draft status).

BMagnu:
> Multiple optimizations and fixes have been made since this PR was introduced. Some of which specifically target the stuttering during a mission, which seems to now be largely resolved. So does this PR still have a noticeable performance impact if rebased against the current code base?
> 
> And to answer your question that I appear to have missed previously, I use both a HDD and SSD. The main game files are on the HDD and my home directory and OS is a SSD. So some file loading slowdowns would be hidden by the SSD when it looks for files in the user path.

Had some time for another round of profiling, and all lags caused by small-file-accesses seem to be gone, at least I can't pin it down with my profiler for now if there still are some.

PR #3710, opened 2021-10-09
Change Build to C++17

### LINKED ISSUE #3709: C++17 Support
By now, we have multiple places in the code that use custom implementation of C++17 features, like variant and optional. This is a valid stopgap for now, but should eventually be replaced with their stdlib counterparts. The increasing use of shared_ptr's also makes it attractive to have make_shared and the likes.
And all in all, it also doesn't really cost us much to allow it: Just drop GCC 5, and add the c++17 flag to the build system.
GCC 10 and clang 9 already fully support C++17.
As for MSVC, you need VS17 for proper C++17 support, and while we currently mandate "at least VS15", our build systems use 19.

What is the general opinion on this? Are there reasons to set a hard limit at VS2015? Do we really need to be able to compiled by a 6 year-old GCC version?

BMagnu:
From the comments in #3709 I take that requiring at least GCC 8, clang 5, or MSVC 2017 is not a problem.
Hence, if there is no further opposition to it, this will close #3709
To do this while keeping our pipeline working, the following steps were necessary:
- Remove GCC 5 from the PR pipeline, since it does not support it. Since the pipeline still contains GCC 10 and clang 9, I don't believe adding a replacement version of GCC is required.
- Setting the compiler for nightlies and releases to GCC 10 instead of GCC 5
- Set the C++ target to cxx17 in the CMakeLists.txt
- Replacing removed functions with their C++17 alternatives (this is only one ``random_shuffle``, replaced by the newer ``shuffle``)


asarium (dismissed):

Goober5000 (approved):
Discussed at length to everyone's satisfaction.  The advantages are clear and persuasive, and the negatives are minimal.  Previous objections have been resolved or adequately mitigated.

Reviewed the PR and everything looks good. 👍

PR #3720, opened 2021-10-12
Remove 'stream' weapon handling

Baezon:
The `stream` weapon flag has *no behavioral differences*. This has been noted for a long time, but the code has still been forced to awkwardly make room for whatever vestigial feature this used to be. This removes all the code that handles the useless code paths made by enforcing `ship_fire_primary` has the correct parameter when firing stream weapons (which is literally all it does, from that point on stream and non-stream weapons are treated identically).


Goober5000:
Ok, I just wanted to make sure that `Trigger_down` still functions correctly for its intended use case.

PR #3729, opened 2021-10-16
Guarantee alphabetical order for 'loose' files loaded on Unix

PhantomHoover:
Modders have been assuming that this ordering is consistent. It's not actually consistent on Windows, either, but it's less inconsistent and nobody has noticed. Someone else will have to do an equivalent to this for the Windows branch of the code.

### DIFF
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -713,12 +713,23 @@ void cf_search_root_path(int root_index)
 
 		if ( dirp ) {
 			struct dirent *dir = nullptr;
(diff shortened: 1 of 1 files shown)


notimaginative:
Wouldn't this be better done by `cf_get_file_list()`? It calls `readdir()` too and that order seems like would be used instead of the indexed list order so this change doesn't seem like it would help all that much.

(and I just noticed that the comment at the top of `cf_get_file_list()` specifically points out the order problem)

Goober5000:
Closing since #3734 has been merged

PR #3734, opened 2021-10-17
clean up some inconsistencies in cfilesystem

notimaginative:
- simplify OS specific code
 - make sure loose files are always sorted the same regardless of OS or filesystem
 - reduce differing behavior in cf_get_file_list*() variants
 - refactor case-sensitive directory handling to work everywhere it's needed

### DIFF
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -171,7 +171,7 @@ struct _file_list_t {
 	size_t size;
 };
 
-static bool sort_file_list(const _file_list_t a, const _file_list_t b)
+static bool sort_file_list(const _file_list_t &a, const _file_list_t &b)
 {
 	return stricmp(a.name.c_str(), b.name.c_str()) < 0;
 }
@@ -303,7 +303,7 @@ static void cf_init_root_pathtypes(cf_root *root)
 			path = parentPathIter->second;
 		}
 
-		if (path.size() && path.back() != DIR_SEPARATOR_CHAR) {
+		if ( !path.empty() && path.back() != DIR_SEPARATOR_CHAR) {
 			path += DIR_SEPARATOR_CHAR;
 		}
 
@@ -1522,7 +1522,7 @@ int cf_get_file_list(SCP_vector<SCP_string>& list, int pathtype, const char* fil
 	}
 
 	if (Get_file_list_child && !verify_file_list_child() ) {
-		Get_file_list_child = NULL;
+		Get_file_list_child = nullptr;
 	}
 
 	cf_create_default_path_string(filespec, sizeof(filespec) - 1, pathtype, (char*)Get_file_list_child, false,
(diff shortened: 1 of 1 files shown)


notimaginative:
This was done rather quickly but was tested as well I could on Linux and Windows. It really deserves a cautious review and some extra testing to double-check that general file behavior is as expected.

jg18:
@notimaginative does this PR include the functionality added by PR #3729? Or does that PR address a separate issue?

CC: @PhantomHoover

notimaginative:
It does. The OS specific code is moved to it's own function so that it can be used anywhere it's needed and the sorting is applied regardless of OS. This way it should give identical results regardless of the platform or function that it's used from.

code/cfile/cfilesystem.cpp
```diff
@@ -54,14 +53,21 @@ enum CfileRootType {
...
 //    searching for pack files on CD-rom tree
 typedef struct cf_root {
-	char	path[CF_MAX_PATHNAME_LENGTH];	// Contains something like c:\projects\freespace or c:\projects\freespace\freespace.vp
+	SCP_string	path;	// Contains something like c:\projects\freespace or c:\projects\freespace\freespace.vp
 	int		roottype;						// CF_ROOTTYPE_PATH  = Path, CF_ROOTTYPE_PACK =Pack file, CF_ROOTTYPE_MEMORY=In memory
 	uint32_t location_flags;
+
+#ifdef SCP_UNIX
+	// map of existing case sensitive paths
+	SCP_unordered_map<int, SCP_string> pathTypeToRealPath;
```
jg18, line 62: Since AIUI `SCP_UNIX` includes macOS and HFS+ is case-insensitive by default, will the `SCP_UNIX`/non-`_WIN32` changes in this PR preserve macOS compatibility?

I know FSO isn't working on macOS at the moment, but IMHO it'd be nice to avoid breaking Mac compatibility if possible.

However, if SCP wants to drop Mac support, IMO that would require a GitHub issue of its own for discussion.

jg18:
@notimaginative
>  The OS specific code is moved to it's own function so that it can be used anywhere it's needed and the sorting is applied regardless of OS. This way it should give identical results regardless of the platform or function that it's used from.

If this PR includes a comprehensive solution to the issue that PR #3729 addresses, should we close that PR (no offense intended @PhantomHoover)?

code/cfile/cfilesystem.cpp
notimaginative: It shouldn't have any real impact on macOS. This code was already there and in-use previously, but was only used when files are initially indexed, so it wasn't applied during the other times that it might be needed. I just made it a part of the general process so that it wouldn't create a different file structure on case-sensitive filesystems (`Data` on read, `data` on write).

notimaginative:
Assuming this PR works as intended and doesn't introduce any side effects, then it's a more robust and uniform fix to that issue and others, so yes.

Goober5000 (approved):
Looks ok to me, though I agree that some more testing is warranted.  But now that we have nightly builds working again, we could just merge it and see if anyone runs into any issues.

the-maddin:
This does not fix the Solaris issue for me (which started the investigation of the bug addressed), namely the script tables are not loaded in inverse alphabetical order, as seen here:

PR #3736, opened 2021-10-17
Removing dead code paths

EatThePath:
gr_set_lighting is only ever called with literals, either `(true,true)` or `(false,false)`. the first argument isn't and can't ever be used, and the second only exists to trigger an immediate early exit from the function without doing anything if false. Removing the arguments and deleting the calls with `(false,false)` improves the readability of the code and shouldn't be able to affect anything. I've done a little bit of mission play to see that it doesn't look like it does have any changes. I'm not currently set up to build fred so if someone else could make sure it behaves that'd be grand.

### DIFF
--- a/code/graphics/light.cpp
+++ b/code/graphics/light.cpp
@@ -294,10 +294,7 @@ void gr_light_init() {
 	gr_lights.reserve(1024);
 }
 
-void gr_set_lighting(bool  /*set*/, bool state) {
-	if (!state) {
-		return;
-	}
+void gr_set_lighting() {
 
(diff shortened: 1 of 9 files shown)


JohnAFernandez:
It does look good to me, but it's always good to test, so I will wait until we can build on Fred and test.

z64555:
Badged members can push a copy of the PR to the `/test/` branch to have a test build made, if needed.  See [the wiki](https://github.com/scp-fs2open/fs2open.github.com/wiki/Automated-Test-Build-compilation) for details.

PR #3737, opened 2021-10-17
Remove light filtering data

EatThePath:
`light_ignore_objnum` and `affected_objnum` support features that, as far as I can determine, can't work in deferred and don't even work in non-deferred lighting. Their continued presence in the code is a distraction.

Submitted in draft at the moment because my ability to actually test these changes has been very limited.

### DIFF
--- a/code/lighting/lighting.cpp
+++ b/code/lighting/lighting.cpp
@@ -221,42 +221,6 @@ void light_add_point(const vec3d *pos, float r1, float r2, float intensity, floa
 	Lights.push_back(l);
 }
 
-void light_add_point_unique(const vec3d *pos, float r1, float r2, float intensity, float r, float g, float b, float spec_r, float spec_g, float spec_b, bool specular)
-{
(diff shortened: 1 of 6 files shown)


EatThePath:
The commit log mess is from me rebasing on current master . I can withdraw this PR and resubmit the changes in a new branch if that's going to make a mess.

EatThePath:
@wookieejedi It's hard to prove a negative, which is the situation here as I don't see how the removed code would have any visible impact on the game. However I've run the game in lighting heavy situations both with and without the change, and with and without -no_deferred, and I can see no difference to my eye.

asarium (approved):
I agree that such a feature is not possible to implement in our current lighting setup so there is no reason to keep this code.

z64555 (changes requested):
Looks good otherwise for a few style/formatting nitpicks

z64555 (approved):

z64555:
Final call for testing before merge.

EatThePath:
I've been working this weekend on a branch incorporating these changes and they're working fine, so I'm pushing the button

PR #3738, opened 2021-10-17
Add nullptr safety for animations

BMagnu:


PhantomHoover:
As discussed on Discord I've tested this and there are other nullptr errors it hasn't covered, and I think we'll be far better off factoring all the code that basically does `Ship_info[shipp->ship_info_index].animations.startAll(...)` into a single function which does a null check on the ship pointer.

Issue #3723, opened 2021-10-13
PLR and CSG bump needed due to sexp containers persistance


z64555:
PR #3462 adds a new section to the .plr and .csg files, but does not bump the version info for either.  In addition to bumping the file versions the read/write controls need another look over to see if they're able to allow forward compatibility for older FSO versions.

If this is not possible, then we'll need to make a fat notice on the next release that the pilotfiles will be incompatible with older versions.

PR #3739, opened 2021-10-18
Bump PLR/CSG versions and preserve compatibility when possible

jg18:
Bump version numbers for .plr and .csg files, but preserve compatibility when there are no persistent containers to save.

Intended to address Issue #3723.

### DIFF
--- a/code/pilotfile/csg.cpp
+++ b/code/pilotfile/csg.cpp
@@ -1728,6 +1728,8 @@ bool pilotfile::save_savefile()
 
 	// header and version
 	cfwrite_int(CSG_FILE_ID, cfp);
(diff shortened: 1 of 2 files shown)


Goober5000:
On second thought, this doesn't "preserve compatibility" as such, because the version number will still be different.  The most robust way to preserve compatibility would be to use the old version number if there are no containers to save.

z64555:
Uh, one more thing.  The controls5 PR has its changes to the CSG to allow players to save and select per-campaign presets.  The compatibility code will thus need some way of coping with this, but I suppose that will be something I'll be responsible for.

z64555:
Ok, whats the intended plan for here? Should this be reverted inside of Controls5 entirely, or only pieces of it?

PR #3741, opened 2021-10-19
fix: cmdbrief not saving current stage text by closing the window

Shivansps:
Command Briefing dialog on Fred2 was not saving the current stage text changes if after editing the text you close the window with the X instead of using the OK.


Goober5000 (commented):
This looks good as far as the PR is concerned, but I think this is expected behavior.  Clicking the X in this situation is equivalent to hitting Cancel.

Goober5000 (approved):

PR #3743, opened 2021-10-20
make gamesnd a little more robust

Goober5000:
a) add a new method `gamesnd_game_sound_exists`
b) add script support for this new method
c) use this method when checking for existence of ballistic rearming sounds
d) prevent "empty" and "none" sounds from triggering warnings

### DIFF
--- a/code/gamesnd/gamesnd.cpp
+++ b/code/gamesnd/gamesnd.cpp
@@ -861,9 +861,10 @@ void parse_gamesnd_new(game_snd* gs, bool no_create)
 	// Default pitch is 1.0. This is set here in case we don't have a valid file name
 	gs->pitch_range = util::UniformFloatRange(1.0f);
 
-	if (!stricmp(name, NOX("empty")))
+	if (!stricmp(name, NOX("empty")) || !stricmp(name, NOX("none")))
 	{
 		entry->filename[0] = 0;
+		gs->flags |= GAME_SND_NOT_VALID;
 		return;
 	}
 
@@ -1344,7 +1345,7 @@ bool gamesnd_interface_sound_valid(interface_snd_id sound) {
 	return sound.isValid() && sound.value() < (int) Snds_iface.size();
 }
 
-bool gamesnd_game_sound_exists(gamesnd_id sound)
+bool gamesnd_game_sound_try_load(gamesnd_id sound)
 {
 	if (!gamesnd_game_sound_valid(sound)) {
 		return false;
--- a/code/gamesnd/gamesnd.cpp
+++ b/code/gamesnd/gamesnd.cpp
@@ -642,10 +642,11 @@ void parse_gamesnd_old(game_snd* gs)
 
 	stuff_string(entry.filename, F_NAME, MAX_FILENAME_LEN, ",");
 
-	if (!stricmp(entry.filename, NOX("empty")))
+	if (!stricmp(entry.filename, NOX("empty")) || !stricmp(entry.filename, NOX("none")))
 	{
 		entry.filename[0] = 0;
-		advance_to_eoln(NULL);
+		advance_to_eoln(nullptr);
+		gs->flags |= GAME_SND_NOT_VALID;
 		return;
 	}
 	Mp++;
@@ -1341,6 +1342,24 @@ bool gamesnd_game_sound_valid(gamesnd_id sound) {
 }
 bool gamesnd_interface_sound_valid(interface_snd_id sound) {
 	return sound.isValid() && sound.value() < (int) Snds_iface.size();
+}
+
+bool gamesnd_game_sound_exists(gamesnd_id sound)
+{
+	if (!gamesnd_game_sound_valid(sound)) {
+		return false;
+	}
+	auto gs = gamesnd_get_game_sound(sound);
+
+	for (auto& entry : gs->sound_entries) {
+		if (!entry.id.isValid()) {
+			// Lazily load unloaded sound entries when required
+			entry.id = snd_load(&entry, &gs->flags);
+		}
+	}
+
+	// check flag
+	return (gs->flags & GAME_SND_NOT_VALID) == 0;
 }
 
 float gamesnd_get_max_duration(game_snd* gs) {
--- a/code/gamesnd/gamesnd.h
+++ b/code/gamesnd/gamesnd.h
@@ -404,7 +404,7 @@ bool gamesnd_game_sound_valid(gamesnd_id sound);
 /**
 * @brief Checks if the given sound handle can be loaded
 */
-bool gamesnd_game_sound_exists(gamesnd_id sound);
+bool gamesnd_game_sound_try_load(gamesnd_id sound);
 
 /**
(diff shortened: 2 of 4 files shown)


code/gamesnd/gamesnd.cpp
```diff
@@ -1343,6 +1344,24 @@ bool gamesnd_interface_sound_valid(interface_snd_id sound) {
...
+	auto gs = gamesnd_get_game_sound(sound);
+
+	for (auto& entry : gs->sound_entries) {
+		if (!entry.id.isValid()) {
+			// Lazily load unloaded sound entries when required
+			entry.id = snd_load(&entry, &gs->flags);
+		}
+	}
+
+	// check flag
```
jg18, line 1361: IMHO the code below is already clear and doesn't need a comment, but I don't feel that strongly about it.

code/scripting/api/objs/sound.cpp
```diff
@@ -124,6 +124,15 @@ ADE_FUNC(isValid, l_SoundEntry, NULL, "Detects whether handle is valid", "boolea
 	return ade_set_args(L, "b", seh->IsValid());
 }
 
+ADE_FUNC(exists, l_SoundEntry, nullptr, "Detects whether handle references a sound that can be loaded", "boolean", "true if exists, false if not, nil if a syntax/type error occurs")
```
jg18, line 127: If we change the function name as requested above, we should change the name here as well.

code/ship/ship.cpp
```diff
@@ -14276,7 +14276,7 @@ int ship_do_rearm_frame( object *objp, float frametime )
 				{
 					// Goober5000
 					gamesnd_id sound_index;
-					if (gamesnd_game_sound_valid(GameSounds::BALLISTIC_START_LOAD))
+					if (gamesnd_game_sound_exists(GameSounds::BALLISTIC_START_LOAD))
```
jg18, line 14279: Is the change here and similar changes below a concern (e.g., potential perf issue) in that `gamesnd_game_sound_valid()` runs fast and has no side effects, whereas the new function doesn't have those guarantees?

I assume it's not a problem but figured I should ask.

jg18 (changes requested):
Some requests, but these are good ideas overall.

code/gamesnd/gamesnd.cpp
```diff
@@ -642,10 +642,11 @@ void parse_gamesnd_old(game_snd* gs)
 
 	stuff_string(entry.filename, F_NAME, MAX_FILENAME_LEN, ",");
 
-	if (!stricmp(entry.filename, NOX("empty")))
+	if (!stricmp(entry.filename, NOX("empty")) || !stricmp(entry.filename, NOX("none")))
```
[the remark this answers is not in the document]
Goober5000: eh, I'd lean toward that being too much scope creep.  However I'll change the other check.

code/gamesnd/gamesnd.cpp
```diff
@@ -642,10 +642,11 @@ void parse_gamesnd_old(game_snd* gs)
...
 	{
 		entry.filename[0] = 0;
-		advance_to_eoln(NULL);
+		advance_to_eoln(nullptr);
+		gs->flags |= GAME_SND_NOT_VALID;
```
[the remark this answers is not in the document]
Goober5000: Added to the other check

code/gamesnd/gamesnd.cpp
```diff
@@ -1343,6 +1344,24 @@ bool gamesnd_interface_sound_valid(interface_snd_id sound) {
...
+{
+	if (!gamesnd_game_sound_valid(sound)) {
+		return false;
+	}
+	auto gs = gamesnd_get_game_sound(sound);
+
+	for (auto& entry : gs->sound_entries) {
+		if (!entry.id.isValid()) {
+			// Lazily load unloaded sound entries when required
+			entry.id = snd_load(&entry, &gs->flags);
```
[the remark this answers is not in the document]
Goober5000: yeah, I followed existing code patterns

code/gamesnd/gamesnd.h
```diff
@@ -401,6 +401,11 @@ game_snd* gamesnd_get_interface_sound(interface_snd_id handle);
  */
 bool gamesnd_game_sound_valid(gamesnd_id sound);
 
+/**
+* @brief Checks if the given sound handle can be loaded
+*/
+bool gamesnd_game_sound_exists(gamesnd_id sound);
```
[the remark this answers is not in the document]
Goober5000: Reasonable.  I'll use `gamesnd_game_sound_try_load`

code/scripting/api/objs/sound.cpp
Goober5000: Done

code/ship/ship.cpp
Goober5000: It would not be a problem in this case because if the sound can be loaded, it will be used.

code/gamesnd/gamesnd.cpp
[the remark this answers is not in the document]
jg18: Ok, that's fair. When I examined all those cases more closely, I saw they weren't as similar as they first looked.

code/gamesnd/gamesnd.cpp
```diff
@@ -860,9 +861,11 @@ void parse_gamesnd_new(game_snd* gs, bool no_create)
 	// Default pitch is 1.0. This is set here in case we don't have a valid file name
 	gs->pitch_range = util::UniformFloatRange(1.0f);
 
-	if (!stricmp(name, NOX("empty")))
+	if (!stricmp(name, NOX("empty")) || !stricmp(name, NOX("none")))
 	{
 		entry->filename[0] = 0;
+		advance_to_eoln(nullptr);
```
jg18, line 867: Hmm, I wonder why this line wasn't there before. Was it just missing?

code/scripting/api/objs/sound.cpp
```diff
@@ -124,6 +124,15 @@ ADE_FUNC(isValid, l_SoundEntry, NULL, "Detects whether handle is valid", "boolea
 	return ade_set_args(L, "b", seh->IsValid());
 }
 
+ADE_FUNC(tryLoad, l_SoundEntry, nullptr, "Detects whether handle references a sound that can be loaded", "boolean", "true if a load succeeded, false if not, nil if a syntax/type error occurs")
```
jg18, line 127: Just curious what the motivation was for adding this function?

jg18 (approved):
Thanks for the updates. Looks good. Just a couple non-blocking questions.

code/scripting/api/objs/sound.cpp
Goober5000: Just to match the `isValid` function above it.  In this use case, `isValid` does not do what a user might think it does.

code/gamesnd/gamesnd.cpp
Goober5000: Hmm, good question.  I don't think it was missing because new gamesnds are parsed differently.  I'll take it out.

PR #3745, opened 2021-10-20
spelling fixes

Goober5000:
"target" has the stress on the first syllable, so the 't' shouldn't be doubled


Goober5000:
As Phantom Hoover said in Discord, he retracted his objection.

JohnAFernandez (approved):
I'm often in favor of changing ints to bools when appropriate.  Makes code easier to read and maintain.

PR #3750, opened 2021-10-25
better syntax checking for the special argument

### LINKED ISSUE #3740: Misplaced <argument> stops event from firing, isn't reported by Shift-H
When an event mentions `<argument>` outside of a `when-argument` or `every-time-argument`, the event won't fire in FSO, and FRED's error checker doesn't report it as a problem.

Goober5000:
Improves syntax checking in two ways...
1) Properly checks each sibling of a special argument, rather than skipping the whole list even if just one is a special argument
2) Throws a syntax error if a special argument is found without an argument operator higher in the sexp tree, which fixes #3740


code/parse/sexp.cpp
```diff
@@ -1968,6 +1963,33 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 			continue; 
 		}
 
+		// if this is the special argument, make sure it is being properly used
+		if (type2 == SEXP_ATOM_STRING && Sexp_nodes[node].flags & SNF_SPECIAL_ARG_IN_NODE) {
```
jg18, line 1967: Just wondering why we're not using `special_argument_appears_in_sexp_list()` here, like we did in the code that's being removed?

code/parse/sexp.cpp
```diff
@@ -1968,6 +1963,33 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
+			bool found = false;
+			z = node;
+			while (true) {
+				z = find_parent_operator(z);
+				if (z < 0) {
```
jg18, line 1972: I take it that `z < 0` means we've gone all the way up the SEXP tree without finding what we're looking for?

code/parse/sexp.cpp
```diff
@@ -1968,6 +1963,33 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
+					break;
+				}
+				if (is_blank_argument_op(get_operator_const(z))) {
+					found = true;
+					break;
+				}
+			}
+
+			// if there is no argument operator higher in the tree, we have a problem
+			if (!found) {
```
jg18, line 1982: Optional: I think starting with the positive case is easier to read. I was initially confused about the code, because I missed the `!` and thought that finding a blank operator was the negative case. It's not important, though, and partly a matter of personal taste.

code/parse/sexp.cpp
```diff
@@ -1938,7 +1933,7 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 			type2 = SEXP_ATOM_STRING;
 
 		} else {
-			Assert(0);
+			UNREACHABLE("SEXP subtype is %d when it should be SEXP_ATOM_LIST, SEXP_ATOM_NUMBER, or SEXP_ATOM_STRING!", Sexp_nodes[node].subtype);
```
jg18, line 1936: Nice update. :+1:

jg18 (approved):
Some non-blocking questions/suggestions but looks good overall, and a nice robustness improvement. I have some experience with SEXP syntax checking thanks to PR #3495.

code/parse/sexp.cpp
Goober5000: yup

code/parse/sexp.cpp
Goober5000: See item 1 in the PR description... `special_argument_appears_in_sexp_list` checks the entire parameter list.  We only want to do argument syntax checking on the argument node.

code/parse/sexp.cpp
Goober5000: I originally had the if() block the other way around, and then I modified it to what you see here.  The comments got a little messy the other way.

code/parse/sexp.cpp
```diff
@@ -1820,11 +1820,6 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 	if (!check_operator_argument_count(count, op))
 		return SEXP_CHECK_BAD_ARG_COUNT;  // incorrect number of arguments
 
-	// Goober5000 - if this is a list of stuff that has the special argument as
-	// an item in the list, assume it's valid
-	if (special_argument_appears_in_sexp_list(op_node))
-		return 0;
-
```
MageKing17, line 1827: Removing this check breaks the dreamscape, Her Finest Hour, and Universal Truth from BP2 (and presumably other missions in other mods). Whenever FRED saves a mission that uses `modify-variable` to set a number variable to `<argument>`, it treats that second node as being `SEXPT_NUMBER` and therefore `SEXP_ATOM_NUMBER`, meaning it gets written to the mission without quotation marks. When it comes time to load the mission, the lack of this check means that it goes on to the `SEXP_ATOM_NUMBER` validation code below, which detects that `<argument>` is not a number and returns `SEXP_CHECK_INVALID_NUM`. This makes FSO refuse to load not only missions that used to work fine before this PR was merged, but also missions made using current FRED2 (and, presumably, qtFRED).

PR #3752, opened 2021-10-25
Add destroy-instantly-with-debris

### LINKED ISSUE #3626: FR: Ability to generate ship debris mid-mission without explosion
The Ships editor in FRED gives you an option to destroy a ship N seconds before mission start which effectively gives you a debris field of that ship's debris.

I'd like to do the same thing for a location the player arrives at later in the mission. My only option right now is to sabotage the hull and wait for the explosion to be done before I move the player to that location or they'll hear all of the explosion noise.

I'm thinking either...

1. Add an optional boolean to destroy-instantly (defaults to false) that would allow the generation of the debris but still not do the death roll or explosion. However I suspect this sexp side-steps the entire death process and that may not be easy to add.
2. Add an optional boolean to self-destruct (defaults to true) that would disable the death roll animation and the ship instantly becomes debris.
3. A new sexp specific for this case.

Baezon:
Just like `destroy-instantly`, with an exemption for debris. Would've liked to make this an optional argument for `destroy-instantly`, but its indefinitely long list of ship arguments makes that impossible unfortunately.

Closes #3626


wookieejedi (approved):
LGTM

PR #3634, opened 2021-08-30
Vectorize IFF Defs for unlimited IFFs

BMagnu:
Caveat: Toggling per-iff ship visibility in FRED only works for the first ten IFF's. Selecting IFF's for ships or in SEXPS works fine.

PR #3760, opened 2021-10-31
Fix Crash On second opening of select box cause by #3634

TaGenius:
The IDC_FILTER_SHIPS_IFF is not properly cleared resulting in the dialog crashing on second opening due to the data being duplicated and and pushing filter_iff out of range.


BMagnu (approved):

JohnAFernandez (approved):
Yup, that would do it.

PR #3761, opened 2021-10-31
polish sexp syntax checking 

### LINKED ISSUE #3758: set-subsystem-strength with <argument> causes a mission to fail to load
On the latest nightly, I just got the following warning:

> Warning: Invalid ship name.
> 
> In sexpression: ( when-argument 
>    ( in-sequence 
>       "Redacted"
>    )
>    ( is-in-mission "&gt;argument&lt;" ) 
>    ( sabotage-subsystem 
>       "&gt;argument&lt;" 
>       "&gt;all engines&lt;" 
>       0 
>    )
>    ( set-subsystem-strength 
>       "&gt;argument&lt;" 
>       "Hull" 
>       ( rand-multiple 10 90 ) 
>    )
> [...]
> (Error appears to be: Hull)
> File: missionparse.cpp
> Line: 6117

The mission then failed to load.

"Hull" is, of course, the subsystem name, not the ship name, and the mission loads correctly on yesterday's nightly.

This is likely related to [commit 14123b1a04a44c9d7604bdeeb14c433b9c13a823](https://github.com/scp-fs2open/fs2open.github.com/commit/14123b1a04a44c9d7604bdeeb14c433b9c13a823).

Goober5000:
Some instances of the ship syntax checker will check other nodes in the sexp tree.  In those cases, make sure to accommodate special-arg nodes.

This also merges syntax checking for goal/event names and docker/dockee points since they are highly similar.

Fixes #3758.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -2147,7 +2147,7 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 			case OPF_SUBSYS_OR_GENERIC:
 			{
 				int shipnum,ship_class;
-				int ship_index;				
+				int ship_node;				
 
 				if (type2 != SEXP_ATOM_STRING){
 					return SEXP_CHECK_TYPE_MISMATCH;
@@ -2181,36 +2181,40 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 					case OP_SHIP_SUBSYS_GUARDIAN_THRESHOLD:
 					case OP_IS_IN_TURRET_FOV:
 					case OP_TURRET_SET_FORCED_TARGET:
-						ship_index = CDR(CDR(op_node));
+						ship_node = CDR(CDR(op_node));
 						break;
 
 					case OP_BEAM_FIRE:
 					case OP_TURRET_SET_FORCED_SUBSYS_TARGET:
 						if(argnum == 1){
-							ship_index = CDR(op_node);
+							ship_node = CDR(op_node);
 						} else {
-							ship_index = CDR(CDR(CDR(op_node)));
+							ship_node = CDR(CDR(CDR(op_node)));
 						}
 						break;
 
 					case OP_BEAM_FLOATING_FIRE:
-						ship_index = CDDDDDR(CDDR(op_node));
+						ship_node = CDDDDDR(CDDR(op_node));
 						break;
 
 					case OP_QUERY_ORDERS:
-						ship_index = CDR(CDR(CDR(CDR(op_node))));
+						ship_node = CDR(CDR(CDR(CDR(op_node))));
 						break;
 	
 					case OP_WEAPON_CREATE:
-						ship_index = CDDDDDR(CDDDDR(op_node));
+						ship_node = CDDDDDR(CDDDDR(op_node));
 						break;
 
 					default:
-						ship_index = CDR(op_node);
+						ship_node = CDR(op_node);
 						break;
 				}
 
-				auto shipname = CTEXT(ship_index);
+				// we can't check special-arg ships
+				if (Sexp_nodes[ship_node].flags & SNF_SPECIAL_ARG_IN_NODE)
+					break;
+
+				auto shipname = CTEXT(ship_node);
 				shipnum = ship_name_lookup(shipname, 1);
 				if (shipnum >= 0)
 				{
@@ -2543,24 +2547,31 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 				break;
 
 			case OPF_AI_GOAL:
+			{
 				if (type2 != OPR_AI_GOAL){
 					return SEXP_CHECK_TYPE_MISMATCH;
 				}
 
-				if (Fred_running) {
-					int ship_num, ship2, w = 0;
+				while (Fred_running) {
+					int ship_num, ship2, wing_num = 0;
 
-					// if it's the "goals" operator, this is part of initial orders, so just assume it's okay
+					// if it's the "goals" operator, this is part of initial orders, so we can't grab the ship from it
 					if (get_operator_const(op_node) == OP_GOALS_ID) {
 						break;
 					}
 
-					ship_num = ship_name_lookup(CTEXT(Sexp_nodes[op_node].rest), 1);	// Goober5000 - include players
+					auto ship_node = Sexp_nodes[op_node].rest;
+
+					// we can't check special-arg ships
+					if (Sexp_nodes[ship_node].flags & SNF_SPECIAL_ARG_IN_NODE)
+						break;
+
+					ship_num = ship_name_lookup(CTEXT(ship_node), 1);	// Goober5000 - include players
 					if (ship_num < 0) {
-						w = wing_name_lookup(CTEXT(Sexp_nodes[op_node].rest));
-						if (w < 0) {
+						wing_num = wing_name_lookup(CTEXT(ship_node));
+						if (wing_num < 0) {
 							if (bad_node){
-								*bad_node = Sexp_nodes[op_node].rest;
+								*bad_node = ship_node;
 							}
 
 							return SEXP_CHECK_INVALID_SHIP;  // should have already been caught earlier, but just in case..
@@ -2577,8 +2588,8 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 						}
 
 					} else {
-						for (i=0; i<Wings[w].wave_count; i++){
-							if (!query_sexp_ai_goal_valid(z, Wings[w].ship_index[i])){
+						for (i=0; i<Wings[wing_num].wave_count; i++){
+							if (!query_sexp_ai_goal_valid(z, Wings[wing_num].ship_index[i])){
 								return SEXP_CHECK_ORDER_NOT_ALLOWED;
 							}
 						}
@@ -2590,6 +2601,9 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 							return SEXP_CHECK_DOCKING_NOT_ALLOWED;
 						}
 					}
+
+					// this is a fake loop that only runs once
+					break;
 				}
 
 				// we should check the syntax of the actual goal!!!!
@@ -2599,6 +2613,7 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 				}
 
 				break;
+			}
 
 			case OPF_SHIP_TYPE:
 				if (type2 != SEXP_ATOM_STRING){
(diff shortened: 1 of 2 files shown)


code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
-					if (i < Campaign.num_missions) {
-						for (t=0; t<Campaign.missions[i].num_goals; t++)
+
+					if (type == OPF_GOAL_NAME) {
+						for (t = 0; t < Campaign.missions[i].num_goals; t++)
 							if (!stricmp(CTEXT(node), Campaign.missions[i].goals[t].name))
 								break;
 
 						if (t == Campaign.missions[i].num_goals)
 							return SEXP_CHECK_INVALID_GOAL_NAME;
```
jg18, line 2743: What about setting `*bad_node` here?

Is that not needed if FRED is running? Even if so, maybe we should defensively set it anyway (if`bad_node` is non-null of course), in case the logic changes in the future such that FSO makes these checks as well?

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
+					if ((Operators[op].value == OP_PREVIOUS_GOAL_TRUE) || (Operators[op].value == OP_PREVIOUS_GOAL_FALSE) || (Operators[op].value == OP_PREVIOUS_GOAL_INCOMPLETE))
 						break;
 
-					for (i=0; i<Num_goals; i++)
+					for (i = 0; i < Num_goals; i++)
 						if (!stricmp(CTEXT(node), Mission_goals[i].name))
 							break;
 
 					if (i == Num_goals)
 						return SEXP_CHECK_INVALID_GOAL_NAME;
```
jg18, line 2762: Same note as above about setting `*bad_node`, assuming the pointer is non-null, of course.

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
-					if (i < Campaign.num_missions) {
-						for (t=0; t<Campaign.missions[i].num_events; t++)
-							if (!stricmp(CTEXT(node), Campaign.missions[i].events[t].name))
-								break;
-
-						if (t == Campaign.missions[i].num_events)
-							return SEXP_CHECK_INVALID_EVENT_NAME;
-					}
-				} else {
+				} else if (type == OPF_EVENT_NAME) {
```
jg18, line 2763: If I'm reading the code right, then we'll get here if either FRED isn't running *or* we're not in campaign mode. Is that the behavior we want?

And if we're not in campaign mode, that implies single-mission mode, correct?

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
 					// MWA -- short circuit evaluation of these things for now.
```
jg18, line 2764: Since we're changing the logic around these lines (and polishing the code besides), could we update this comment from retail to something more relevant/meaningful today?

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
 			case OPF_DOCKER_POINT:
+			case OPF_DOCKEE_POINT:
 				if (type2 != SEXP_ATOM_STRING)
 					return SEXP_CHECK_TYPE_MISMATCH;
 
 				// This makes massive assumptions about the structure of the SEXP using it. If you add any 
 				// new SEXPs that use this OPF, you will probably need to edit this section to accommodate them.
 				if (Fred_running) {
-					int ship_num, model;
+					int ship_num, ship_node = -1, model;
```
jg18, line 2785: Same comment as the one around line 2568 under `case OPF_AI_GOAL`, about using `ship_node` when we're also using the same var name in the outer scope.

code/parse/sexp.cpp
```diff
@@ -33947,9 +33928,9 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 	{ OP_SET_DOCKED, "set-docked\r\n"
 		"\tCauses one ship to become instantly docked to another at the specified docking ports.  Takes 4 arguments...\r\n"
 		"\t1: Docker ship\r\n"
-		"\t1: Docker point\r\n"
-		"\t1: Dockee ship\r\n"
-		"\t1: Dockee point\r\n"
+		"\t2: Docker point\r\n"
+		"\t3: Dockee ship\r\n"
+		"\t4: Dockee point\r\n"
```
jg18, line 33933: Nice catch. :+1:

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
-						ship_num = ship_name_lookup(CTEXT(Sexp_nodes[op_node].rest), 1);
+						if (get_operator_const(op_node) == OP_AI_DOCK)	// ai-dock with dockee
+							ship_node = CDR(z);
+						else if (type == OPF_DOCKER_POINT)
+							ship_node = CDR(z);
+						else if (type == OPF_DOCKEE_POINT)
+							ship_node = CDDDR(z);
 					}
 
+					// can't check special-arg or missing node
```
jg18, line 2809: I'm confused as to why it's ok if the node is missing? Is it because it's an optional arg? If so, maybe worth mentioning that in the new comment for clarity?

jg18: Also, if it *is* an optional arg, can we be sure that it's optional in all SEXP types we're checking here?

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
-						ship_num = ship_name_lookup(Sexp_nodes[z].text, 1);
-					}
-					else {
-						ship_num = ship_name_lookup(CTEXT(Sexp_nodes[op_node].rest), 1);
+						if (get_operator_const(op_node) == OP_AI_DOCK)	// ai-dock with dockee
+							ship_node = CDR(z);
+						else if (type == OPF_DOCKER_POINT)
+							ship_node = CDR(z);
+						else if (type == OPF_DOCKEE_POINT)
+							ship_node = CDDDR(z);
```
jg18, line 2806: How about adding an `else` clause with `UNREACHABLE()`?

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
-					}
-					else {
+					// Look for the node containing the docker/dockee ship. In most cases, we want 
+					// the current SEXP operator, but for ai-dock and the docker, we want its parent.
+					if (get_operator_const(op_node) == OP_AI_DOCK && type == OPF_DOCKER_POINT) {
 						z = find_parent_operator(op_node);
-					
-						// if it's the "goals" operator, this is part of initial orders, so just assume it's okay
+
+						// if it's the "goals" operator, this is part of initial orders, so we can't grab the ship from it
```
jg18, line 2792: Same question as the one above about a similar comment. Just want to check if the answer is the same here.

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
 						if (get_operator_const(z) == OP_GOALS_ID) {
```
jg18, line 2793: How about a check or assertion that `z >= 0`?

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
+						for (t = 0; t < Campaign.missions[i].num_events; t++)
+							if (!stricmp(CTEXT(node), Campaign.missions[i].events[t].name))
+								break;
+
+						if (t == Campaign.missions[i].num_events)
+							return SEXP_CHECK_INVALID_EVENT_NAME;
 					}
-				} else {
+				} else if (type == OPF_GOAL_NAME) {
 					// MWA -- short circuit evaluation of these things for now.
```
jg18, line 2753: Same note as above about updating this comment from retail.

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
+
+					if (type == OPF_GOAL_NAME) {
+						for (t = 0; t < Campaign.missions[i].num_goals; t++)
 							if (!stricmp(CTEXT(node), Campaign.missions[i].goals[t].name))
 								break;
 
 						if (t == Campaign.missions[i].num_goals)
 							return SEXP_CHECK_INVALID_GOAL_NAME;
+					} else if (type == OPF_EVENT_NAME) {
+						for (t = 0; t < Campaign.missions[i].num_events; t++)
```
jg18, line 2745: Since this loop body is multi-line, using braces for this `for`-loop would be nice, but I admit this is partially personal taste.

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
+							if (!stricmp(CTEXT(node), Campaign.missions[i].events[t].name))
+								break;
+
+						if (t == Campaign.missions[i].num_events)
+							return SEXP_CHECK_INVALID_EVENT_NAME;
```
jg18, line 2750: What about setting `*bad_node` here?

code/parse/sexp.cpp
jg18: Also, how about an `else` with `UNREACHABLE()` at the end of this `else-if` block, around line 2774? What would it mean for the condition at this line to be `false`?

jg18 (changes requested):
Assorted requests/questions, but I think these are generally good ideas.

Given the logical complexity of `check_sexp_syntax()`, I might have missed some things, so there may be follow-up questions/requests.

code/parse/sexp.cpp
```diff
@@ -2147,7 +2147,7 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 			case OPF_SUBSYS_OR_GENERIC:
 			{
 				int shipnum,ship_class;
-				int ship_index;				
+				int ship_node;
```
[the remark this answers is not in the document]
Goober5000: Yup, because one's first impression of `ship_index` is that it could be the index into the `Ships` array.

code/parse/sexp.cpp
```diff
@@ -2226,7 +2230,12 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 						if (type == OPF_SUBSYSTEM_OR_NONE)
 							break;
 						else
+						{
+							if (bad_node)
+								*bad_node = ship_node;
```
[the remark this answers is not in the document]
Goober5000: Yup

code/parse/sexp.cpp
```diff
@@ -2543,24 +2552,31 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
 					return SEXP_CHECK_TYPE_MISMATCH;
 				}
 
-				if (Fred_running) {
-					int ship_num, ship2, w = 0;
+				while (Fred_running) {
+					int ship_num, ship2, wing_num = 0;
 
-					// if it's the "goals" operator, this is part of initial orders, so just assume it's okay
+					// if it's the "goals" operator, this is part of initial orders, so we can't grab the ship from it
```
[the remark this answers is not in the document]
Goober5000: We can't get it if this is an initial-goal sexp.  Here's an example of the GTT Comet docking to the Omega when the Comet arrives in mission 4 of FS1:
```
$AI Goals: ( goals ( ai-dock "Omega" "topside docking" "topside docking" 89 ) )
```
In this situation, the name of the docking ship cannot be found from within the sexp.

code/parse/sexp.cpp
```diff
@@ -2543,24 +2552,31 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
 					if (get_operator_const(op_node) == OP_GOALS_ID) {
 						break;
 					}
 
-					ship_num = ship_name_lookup(CTEXT(Sexp_nodes[op_node].rest), 1);	// Goober5000 - include players
+					auto ship_node = Sexp_nodes[op_node].rest;
```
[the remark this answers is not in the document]
Goober5000: The other `ship_node` variable is confined within a separate scope, though.  They are two different instances of the same purpose, like reusing `i` in two different `for` loops.

Checking that `ship_node` >= 0 isn't necessary here because that must have already been true in order for the parser to get to this point.  In any case, it's not done in the similar syntax checking block for subsystems.

code/parse/sexp.cpp
```diff
@@ -2590,6 +2606,9 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 							return SEXP_CHECK_DOCKING_NOT_ALLOWED;
 						}
 					}
+
+					// this is a fake loop that only runs once
+					break;
```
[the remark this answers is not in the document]
Goober5000: Using multiple levels of if() blocks would have been messy.  But I figured out how to reorder the logic to remove the `while`.

code/parse/sexp.cpp
```diff
@@ -2181,36 +2181,40 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
 					default:
-						ship_index = CDR(op_node);
+						ship_node = CDR(op_node);
 						break;
 				}
 
-				auto shipname = CTEXT(ship_index);
+				// we can't check special-arg ships
+				if (Sexp_nodes[ship_node].flags & SNF_SPECIAL_ARG_IN_NODE)
+					break;
```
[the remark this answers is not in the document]
Goober5000: Again this should have been already checked.  I guess I can add an Assert though

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
 
+					if (i >= Campaign.num_missions) {
+						if (bad_node)
+							*bad_node = z;
+
+						return SEXP_CHECK_INVALID_MISSION_NAME;
+					}
+
 					// read the goal/event list from the mission file if both num_goals and num_events
 					// are < 0
```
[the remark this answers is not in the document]
Goober5000: done

code/parse/sexp.cpp
```diff
@@ -2681,159 +2701,120 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
 						if (t == Campaign.missions[i].num_goals)
 							return SEXP_CHECK_INVALID_GOAL_NAME;
+					} else if (type == OPF_EVENT_NAME) {
+						for (t = 0; t < Campaign.missions[i].num_events; t++)
+							if (!stricmp(CTEXT(node), Campaign.missions[i].events[t].name))
+								break;
+
+						if (t == Campaign.missions[i].num_events)
+							return SEXP_CHECK_INVALID_EVENT_NAME;
 					}
```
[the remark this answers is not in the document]
Goober5000: ok

code/parse/sexp.cpp
Goober5000: Setting `bad_node` is only necessary if the bad node is something other than the node being checked.  Elsewhere in the function it assumes the bad node is the current node.

code/parse/sexp.cpp
Goober5000: That's the behavior we want, because if we're not running FRED, or we are but we're not in campaign mode, then we only check the goals/events of the current mission.  And yes, non-campaign mode implies single-mission mode.

Added `UNREACHABLE` with an explanation

code/parse/sexp.cpp
Goober5000: ok

code/parse/sexp.cpp
Goober5000: same response - `ship_node` is not being used in the outer scope

code/parse/sexp.cpp
Goober5000: eh, good points - will change

code/parse/sexp.cpp
Goober5000: same answer

code/parse/sexp.cpp
Goober5000: As before - `bad_node` doesn't need to be set for the current node

code/parse/sexp.cpp
Goober5000: Eh, it's all single-line statements, and this is already somewhat messy code

code/parse/sexp.cpp
Goober5000: done

code/parse/sexp.cpp
```diff
@@ -2681,159 +2700,126 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
+					int ship_num, ship_node = -1, model;
 
-					// Look for the node containing the docker ship as its first argument. For set-docked, we want 
-					// the current SEXP. Otherwise (for ai-dock), we want its parent.
-					if (get_operator_const(op_node) == OP_SET_DOCKED) {
-						z = op_node;
-					}
-					else {
+					// Look for the node containing the docker/dockee ship. In most cases, we want 
+					// the current SEXP operator, but for ai-dock and the docker, we want its parent.
```
jg18, line 2790: Thanks for including this comment. I was confused by the below code change until I read it.

jg18 (approved):
Looks reasonable.

Thanks for the updates and replies and for doing this cleanup. Making non-trivial changes to `check_sexp_syntax()` is always an adventure, to say the least. 🙂

PR #3762, opened 2021-11-02
Ai use lock status for optimum range choice

Baezon:
What it says on the tin, AI won't choose an optimum range based on a weapon which is currently locked.


JohnAFernandez (approved):
Code looks fine.  Probably need to put a message in #SCP_announcements about the behavior change, since this will likely affect balance wherever mods have decided to use locked weapons.

I think it's rare enough to approve, though, without a special flag.  If anyone else thinks differently, feel free to speak up.

Baezon:
If you think so, I suppose. This is a very unlikely set of circumstances and a very 'young' feature so I really doubt anyone will be caught off guard, personally.

Goober5000:
LGTM too

PR #3767, opened 2021-11-07
guard against CTD in sexp variables

Goober5000:
Don't assign strings that are too long for the string variable capacity.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -30316,6 +30316,7 @@ void sexp_modify_variable(const char *text, int index, bool sexp_callback)
 	Assert(index >= 0 && index < MAX_SEXP_VARIABLES);
 	Assert(Sexp_variables[index].type & SEXP_VARIABLE_SET);
 	Assert( !MULTIPLAYER_CLIENT );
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -30337,14 +30337,21 @@ void sexp_modify_variable(const char *text, int index, bool sexp_callback)
 		SCP_string temp_text = text;
 		sexp_replace_variable_names_with_values(temp_text);
 
+		if (temp_text.length() > TOKEN_LENGTH - 1)
```
jg18, line 30340: Optional: Instead of using `TOKEN_LENGTH - 1` repeatedly, consider adding a line before this `if`-block like
```
const size_t maxCopyLen = TOKEN_LENGTH - 1;
```
and then use `maxCopyLen` everywhere instead.

That'll avoid the issue of one of the `- 1`s getting inadvertently dropped in the future, and it improves readability a bit.

jg18 (approved):
Nice catch. Just one suggestion below.

code/parse/sexp.cpp
Goober5000: Ok

Issue #527, opened 2016-01-30
g3_draw_sphere_ez isn't suitable for debugging


Goober5000:
I wanted to use this function to draw spheres for debugging purposes.  It would be nice to visually confirm that I'm getting coordinates to line up properly.  Unfortunately, the spheres aren't being displayed.

I posted the following test code and @DahBlount thinks it should work:

```
bool do_g3 = (G3_count < 1);

if (do_g3)
    g3_start_frame(1);

gr_set_color(128, 128, 128);
g3_draw_sphere_ez(&vmd_zero_vector, 50.0f);

if (do_g3)
    g3_end_frame();
```

(Note: if `g3_start_frame` and `g3_end_frame` are not included, the engine will crash with an Assert since `G3_count` is 0.)

Any ideas?

EDIT: I'll amend the issue.  The function works perfectly if it's added somewhere in the graphics render pipeline (e.g. if the waypoint render functions are uncommented in `obj_queue_render`) but it doesn't work if you just plop it in a random function using this test code.  So I'll amend this issue: how can this function be used for debugging?  Is there other wrapper code that could be used?  It looks like `g3_start_frame` and `g3_end_frame` are not sufficient.

EatThePath:
Seems like the thing to do here would be to do like lights do, have a vector of debug spheres/other that is filled by a function called from regular gameplay, read during the graphics pipeline to build the spheres, the clear that vector to be refilled during the next frame. 

I'm assuming, anyway. I would need to poke more at it.

Goober5000:
@Kiloku Do you have any insight into this issue based on your recent work with drawing lines in scripting?

Goober5000:
Hang on, looks like #3769 satisfies this.

EatThePath:
Small followup comment to that implementation : g3_draw_sphere_ez is kind of visually crap and there's a lot of room for improvement. I think the way debug spheres are implemented leaves room for that improvement if someone wants to come along at any point and replace the current drawing with something else

PR #3769, opened 2021-11-07
Implemented debug sphere queuing and rendering.

EatThePath:
Implements #527. Any code should be able to call debug_spheres::add() with a position and size and get a sphere drawn that frame. 
There's a lot of room for improvement here, because g3_draw_sphere_ez is not great actually, but that improvement can be done transparently later if it's actually needed.


code/graphics/debug_sphere.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "globalincs/pstypes.h"
+struct debug_sphere {
+	vec3d pos;
+	float radius;
+};
+
+namespace debug_spheres {
```
JohnAFernandez, line 9: Did you think about making this a class?  You're using a very modular design here, which is excellent, but it could be even further improved, imo, by making this a class with the sphere vector being a member and the global variable being of the class instead.  The advantage would be that you would automatically protect the sphere vector from accidental outside edits, and you could mark `clear()` as private, while keeping `add()` and `render()` as public functions as well.  Again, this avoids some other chunk of code clearing the vector because a bug got through.

code/graphics/debug_sphere.h
```diff
@@ -0,0 +1,13 @@
...
+	void add(vec3d pos, float radius);
+	void clear();
+	void render();
+}
```
JohnAFernandez, line 13: This probably looks weird to you and it's not a big deal, but I think you can fix them by pressing enter after the ending brace, iirc.

code/graphics/debug_sphere.cpp
```diff
@@ -0,0 +1,26 @@
...
+}
+
+void clear(){
+	Spheres.clear();
+}
+void render(){
+	for(auto s : Spheres){
+		g3_draw_sphere_ez(&(s.pos),s.radius);
+	}
+	clear();
```
JohnAFernandez, line 23: It's an interesting idea to just clear the whole vector every time.  Maybe, if expanded in the future, you could have the option of having a sphere persist.

JohnAFernandez (changes requested):
Needs some small changes, and perhaps one larger change.

code/graphics/debug_sphere.h
JohnAFernandez: It's not showing up here, but I mean the red circle that appears if you go to the files changed tab.

code/graphics/debug_sphere.h
EatThePath: I did think about putting it all into a class and doing things like that. I'm not sure why I went with a namespace, I'd probably just been reminded they exist by another part of the code and was trying to mimic conventions. When in Rome and so forth. I'll happily switch it over, in any case.

JohnAFernandez: You can even do both, but moving it over is the best way to keep anyone from accidentally adding an inappropriate `clear()` call.

code/graphics/debug_sphere.cpp
EatThePath: Clearing every frame mimics how lights are implemented. There's definitely room here for some more logic in the future, yeah.

freespace2/freespace.cpp
```diff
@@ -10,6 +10,7 @@
 
 
 
+#include "graphics/debug_sphere.h"
```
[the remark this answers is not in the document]
EatThePath: I think one of my VScode extensions auto-adds missing includes as you write code, I'll have to keep an eye on that in the future.

code/graphics/debug_sphere.h
EatThePath: I've reorganized everything into a class with Spheres a private static field. If there's any further adjustment it could use just let me know. My lighting data branch uses a similar pattern, so it'll be to good to know if this is solid or needs tweaking.

code/graphics/debug_sphere.cpp
```diff
@@ -0,0 +1,26 @@
+#include "debug_sphere.h"
+#include "globalincs/vmallocator.h"
+#include "render/3d.h"
+
+SCP_vector<debug_sphere> debug_sphere::Spheres;
+
+void debug_sphere::add(vec3d pos, float rad)
```
Goober5000, line 7: use `vec3d &pos` (with or without a `const`) to avoid a copy

code/graphics/debug_sphere.cpp
```diff
@@ -0,0 +1,26 @@
...
+{
+	debug_sphere s;
+	s.pos = pos;
+	s.radius = rad;
+	debug_sphere::Spheres.push_back(s);
```
Goober5000, line 12: you could use `emplace_back` with a constructor to avoid another copy

code/graphics/debug_sphere.cpp
```diff
@@ -0,0 +1,26 @@
...
+}
+
+void debug_sphere::clear()
+{
+	debug_sphere::Spheres.clear();
+}
+
+void debug_sphere::render()
+{
+	for(auto s : Spheres){
```
Goober5000, line 22: use `auto &s` here to avoid a copy on every loop iteration

Goober5000 (changes requested):
a few efficiency recommendations

code/graphics/debug_sphere.cpp
```diff
@@ -0,0 +1,38 @@
...
+SCP_vector<debug_sphere> debug_sphere::Spheres;
+
+debug_sphere::debug_sphere(const vec3d& pos_in, float rad_in)
+{
+	pos = pos_in;
+	radius = rad_in;
+}
+void debug_sphere::add(const vec3d& pos, float rad)
+{
+	debug_sphere::Spheres.emplace_back(debug_sphere(pos, rad));
```
Goober5000, line 16: I *think* you can take `debug_sphere` out of this line and it will be constructed in-place.

Goober5000 (approved):
Looks good.  One optional suggestion.

JohnAFernandez (approved):

code/graphics/debug_sphere.cpp
EatThePath: @Goober5000 I wasn't sure at first which instance `debug_sphere` you meant so I hit them both :)

Tested afterwards and all works as it should

Goober5000: I meant the second one, but you're right, the first one is removable too. 👍

PR #3774, opened 2021-11-13
indicate ambiguous goal or event matches

### LINKED ISSUE #3766: Multiple events with the same name don't cause a FRED warning
Exactly what it says on the tin. Accidentally giving multiple events the same name can lead to really confusing mission bugs and should trigger a warning in FRED.

Goober5000:
Check for multiple matches, not just missing matches.  Also streamline the syntax checking a bit more.  Fixes #3766.


code/parse/sexp.cpp
```diff
@@ -1762,6 +1762,16 @@ int check_operator_argument_count(int count, int op)
 	return 1;
 }
 
+template <typename T>
+int count_named_items(const char *name, int num, const T* item)
```
jg18, line 1766: Since `item` is an array, please rename to `items` or similar.

For `num` I'd prefer `num_items` or `size`.

Also, I think it's more common (and more intuitive) to have the array size argument after the pointer/array itself.

code/parse/sexp.cpp
```diff
@@ -1762,6 +1762,16 @@ int check_operator_argument_count(int count, int op)
...
+{
+	int count = 0;
+	for (int i = 0; i < num; ++i)
+		if (!stricmp(name, item[i].name))
```
jg18, line 1770: Optional: `Assert(item[i].name != nullptr)`

code/parse/sexp.cpp
```diff
@@ -1762,6 +1762,16 @@ int check_operator_argument_count(int count, int op)
 	return 1;
 }
 
+template <typename T>
+int count_named_items(const char *name, int num, const T* item)
+{
```
jg18, line 1767: Optional: Validate inputs via assertions, i.e., that the size > 0 and that both pointer arguments are non-null.

code/parse/sexp.cpp
jg18: Also, the name `count_named_items()` is a bit unintuitive IMO. Maybe something like `count_items_matching_name()`, or maybe `count_matching_items()`/`count_occurrences()`  and rename the first argument to `name_to_match`.

code/parse/sexp.cpp
```diff
@@ -2718,11 +2731,10 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
...
 
 					// look for mission
-					for (i = 0; i < Campaign.num_missions; i++)
-						if (!stricmp(CTEXT(z), Campaign.missions[i].name))
-							break;
+					count = count_named_items(CTEXT(z), Campaign.num_missions, Campaign.missions);
 
-					if (i >= Campaign.num_missions) {
+					// only check for a missing mission -- it's ok if the same mission appears multiple times in the campaign
+					if (count <= 0) {
```
jg18, line 2737: Why `<= 0` as opposed to `== 0`? Additional defensiveness?

code/parse/sexp.cpp
```diff
@@ -2718,11 +2731,10 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 						break;
 
 					// look for mission
-					for (i = 0; i < Campaign.num_missions; i++)
-						if (!stricmp(CTEXT(z), Campaign.missions[i].name))
-							break;
+					count = count_named_items(CTEXT(z), Campaign.num_missions, Campaign.missions);
 
-					if (i >= Campaign.num_missions) {
+					// only check for a missing mission -- it's ok if the same mission appears multiple times in the campaign
```
jg18, line 2736: Thanks for including this comment. I was initially confused until I read it. :+1:

code/parse/sexp.cpp
```diff
@@ -2702,9 +2712,12 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 
 			case OPF_GOAL_NAME:
 			case OPF_EVENT_NAME:
+			{
 				if (type2 != SEXP_ATOM_STRING)
 					return SEXP_CHECK_TYPE_MISMATCH;
 
+				count = 0;
```
jg18, line 2719: How about initializing `count` when declaring it instead?

code/parse/sexp.cpp
```diff
@@ -30138,6 +30137,12 @@ const char *sexp_error_message(int num)
 		case SEXP_CHECK_MISPLACED_SPECIAL_ARGUMENT:
 			return "Found " SEXP_ARGUMENT_STRING " without an argument handler higher in the sexp node tree";
 
+		case SEXP_CHECK_AMBIGUOUS_GOAL_NAME:
```
jg18, line 30140: Instead of "ambiguous", how about "duplicate"? IMO that more precisely captures the error.

"Ambiguous" is a bit, well, ambiguous, hence the need for the parenthetical explanation.

jg18 (changes requested):
A few requests/questions, but overall reasonable and a good idea.

code/parse/sexp.cpp
Goober5000: ok, messed around with this

code/parse/sexp.cpp
Goober5000: The name itself should have been caught before this point, but I added Asserts for the function arguments.

code/parse/sexp.cpp
Goober5000: I guess, but probably not necessary

code/parse/sexp.cpp
Goober5000: Because `count` is declared at the beginning of `check_sexp_syntax` and could be used in more places than this.

code/parse/sexp.cpp
Goober5000: A duplicate name is not itself an error; it only becomes an error if `is-event-true` could refer to multiple events.  It's the ambiguity that's the real problem here.

In line with that reasoning, I think the error message says everything that needs to be said.

Goober5000:
Comments addressed

code/parse/sexp.cpp
jg18: Ok, sure.

jg18 (approved):
Looks good. Thanks for the replies/updates!

PR #3776, opened 2021-11-14
Rotated images

Kiloku:
This would allow scripters to draw rotated images to the screen by passing an optional angle parameter, similar to the way you can draw rotated rectangles.


Kiloku:
This wasn't supposed to include all of those commits. I might have done something wrong in my own local merge

PR #3777, opened 2021-11-14
Allow drawing rotated bitmaps

Kiloku:
This would allow scripters to draw rotated images to the screen by passing an optional angle parameter, similar to the way you can draw rotated rectangles.
Now without git fuckery. Apologies for not checking properly before creating the previous PR.

### DIFF
--- a/code/graphics/render.cpp
+++ b/code/graphics/render.cpp
@@ -1194,12 +1194,6 @@ static void draw_bitmap_list(bitmap_rect_list* list, int n_bm, int resize_mode, 
 	float sw = 0.1f;
 
 
-	float rad = 0;
-
-	if (angle != 0) {
-		rad = fl_radians(angle);
-	}
-
 	for (int i = 0; i < n_bm; i++) {
 		// stuff coords
 		
@@ -1218,8 +1212,8 @@ static void draw_bitmap_list(bitmap_rect_list* list, int n_bm, int resize_mode, 
 		vertex* V = &vert_list[i * 6];
 		
 		if (angle != 0.f) {
-			V->screen.xyw.x = cosf(rad) * (b->x - centerX) - sinf(rad) * (b->y - centerY) + centerX;
-			V->screen.xyw.y = sinf(rad) * (b->x - centerX) + cosf(rad) * (b->y - centerY) + centerY;
+			V->screen.xyw.x = cosf(angle) * (b->x - centerX) - sinf(angle) * (b->y - centerY) + centerX;
+			V->screen.xyw.y = sinf(angle) * (b->x - centerX) + cosf(angle) * (b->y - centerY) + centerY;
 		}
 		else {			
 			V->screen.xyw.x = (float)b->x;
@@ -1234,8 +1228,8 @@ static void draw_bitmap_list(bitmap_rect_list* list, int n_bm, int resize_mode, 
 
 		V++;
 		if (angle != 0.f) {
-			V->screen.xyw.x = cosf(rad) * (b->x + b->w - centerX) - sinf(rad) * (b->y - centerY) + centerX;
-			V->screen.xyw.y = sinf(rad) * (b->x + b->w - centerX) + cosf(rad) * (b->y - centerY) + centerY;
+			V->screen.xyw.x = cosf(angle) * (b->x + b->w - centerX) - sinf(angle) * (b->y - centerY) + centerX;
+			V->screen.xyw.y = sinf(angle) * (b->x + b->w - centerX) + cosf(angle) * (b->y - centerY) + centerY;
 		}
 		else {
 			V->screen.xyw.x = (float)(b->x + b->w);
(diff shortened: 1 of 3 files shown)


Kiloku:
[Example video](https://i.imgur.com/yp6wBaf.mp4)

code/graphics/render.cpp
```diff
@@ -1193,33 +1193,71 @@ static void draw_bitmap_list(bitmap_rect_list* list, int n_bm, int resize_mode,
 	auto vert_list = new vertex[6 * n_bm];
 	float sw = 0.1f;
 
+
+	float rad = 0;
+
+	if (angle != 0) {
+		rad = fl_radians(angle);
+	}
```
Goober5000, line 1201: Lua functions are in radians, so this should be in radians also.

code/scripting/api/libs/graphics.cpp
```diff
@@ -1352,11 +1352,12 @@ ADE_FUNC(loadTexture, l_Graphics, "string Filename, [boolean LoadIfAnimation, bo
...
 	l_Graphics,
 	"string|texture fileNameOrTexture, [number X1=0, number Y1=0, number X2, number Y2, number UVX1 = 0.0, number UVY1 "
-	"= 0.0, number UVX2=1.0, number UVY2=1.0, number alpha=1.0, boolean aaImage = false]",
+	"= 0.0, number UVX2=1.0, number UVY2=1.0, number alpha=1.0, boolean aaImage = false, number angle = 0.0]",
 	"Draws an image file or texture. Any image extension passed will be ignored."
 	"The UV variables specify the UV value for each corner of the image. "
 	"In UV coordinates, (0,0) is the top left of the image; (1,1) is the lower right. If aaImage is true, image needs "
-	"to be monochrome and will be drawn tinted with the currently active color.",
+	"to be monochrome and will be drawn tinted with the currently active color."
+	"The angle variable can be used to rotate the image, by degrees.",
```
Goober5000, line 1360: The angle should be specified as radians, not degrees.  And, looking through the rest of the scripting functions, I see that `drawRectangle` was coded to use degrees also.  That should be switched over to use radians.  It was only added last month so the API impact should be quite low.

code/graphics/render.cpp
```diff
@@ -1193,33 +1193,71 @@ static void draw_bitmap_list(bitmap_rect_list* list, int n_bm, int resize_mode,
...
+		}
+
 		// tri one
 		vertex* V = &vert_list[i * 6];
-		V->screen.xyw.x = (float)b->x;
-		V->screen.xyw.y = (float)b->y;
+		
+		if (angle != 0.f) {
+			V->screen.xyw.x = cosf(rad) * (b->x - centerX) - sinf(rad) * (b->y - centerY) + centerX;
+			V->screen.xyw.y = sinf(rad) * (b->x - centerX) + cosf(rad) * (b->y - centerY) + centerY;
```
Goober5000, line 1222: I just want to double-check that the center point is calculated correctly.  Is the x,y point always going to be the top-left corner of the image, even when the image is rotated?  If you draw the same image with a 0 angle and with a rotated angle, will their top-left corners meet at the same point?

Or is it instead the case that rotated images are drawn with the x,y representing their center point?  In fact, thinking along those lines, it would be handy to have an alternate function that uses the center point and the width/height, in addition to the x1,y1-x2,y2 method.

Goober5000 (changes requested):
Looks good, mostly.  One change that should be made (use radians rather than degrees) and several questions.

code/graphics/render.cpp
Kiloku: Sorry about that. I could swear that degrees were the default. I'll fix it.

code/scripting/api/libs/graphics.cpp
Kiloku: I think I'm the only one who has used that feature at any point 😅 
So yeah, very low impact.

code/graphics/render.cpp
Kiloku: Here's [a screenshot](https://i.imgur.com/TskqmEK.png) of how it looks when a rotated and a non-rotated image are in the same place. The one in the foreground is the non-rotating one. 

Without this calculation, the image would rotate around the top left, which would effectively change its position while it rotates. 

I agree that it'd be nice to have a version that uses width/height, because IMO the x2/y2 is a bit unintuitive.

Kiloku:
I'll work on the requested changes possibly later tonight, or if I'm too busy, I'll wait for the weekend.

code/graphics/render.cpp
Goober5000: I can't really tell how those images are aligned - if they share a corner, or share a center, or share width/height.

I agree that the image shouldn't rotate around its top-left corner, but I'm not sure what it is actually doing.

Kiloku: They share a center

Goober5000 (approved):
Looks good. 👍

PR #3786, opened 2021-11-17
refactor submodel flags and add $rotate_accel

Goober5000:
1) Changes the submodel flags to use a flagset instead of several booleans
2) Adds `rotate_accel` to allow submodel rotation acceleration to be specified

Submodel acceleration can now be instant; dumb-rotating submodels always instantly accelerate (and have always done so)

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -640,7 +640,6 @@ void model_copy_subsystems( int n_subsystems, model_subsystem *d_sp, model_subsy
 				dest->pnt = source->pnt;
 				dest->radius = source->radius;
 				dest->type = source->type;
-				dest->turn_rate = source->turn_rate;
 				dest->turret_gun_sobj = source->turret_gun_sobj;
 
                 strcpy_s(dest->name, source->name);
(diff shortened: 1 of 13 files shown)


code/model/modelread.cpp
```diff
@@ -1549,9 +1576,8 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
 						turn_rate = static_cast<float>(atof(buf));
 					}
 
-					sm->dumb_turn_rate = turn_rate;
-				} else {
-					sm->dumb_turn_rate = 0.0f;
+					sm->default_turn_rate = turn_rate;
+					sm->flags.set(Model::Submodel_flags::Instant_rotate_accel);
```
BMagnu, line 1580: This should also set the ``Dumb_rotates`` flag, no?
Though it was missing before as well...

BMagnu (changes requested):
Looks good, apart from one small thing

code/model/modelread.cpp
Goober5000: Good call.

Goober5000: Turns out it's not even needed (it's redundant in the one place it's used) so I removed it.

BMagnu (approved):

PR #3790, opened 2021-11-20
Change rectangle rotation to use radians

Kiloku:
@Goober5000 pointed out that lua functions should use radians for angles by default. Since the rectangle rotation feature has only been implemented recently (and as far as I know I'm the only one who's used it), the impact is low enough that changing this is feasible.

### DIFF
--- a/code/graphics/render.cpp
+++ b/code/graphics/render.cpp
@@ -1054,7 +1054,7 @@ void gr_rect(int x, int y, int w, int h, int resize_mode, float angle) {
 		float offsetX = x + w / 2.0f;
 		float offsetY = y + h / 2.0f;
 		path->translate(offsetX, offsetY);
-		path->rotate(fl_radians(angle));
+		path->rotate(angle);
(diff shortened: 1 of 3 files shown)


code/graphics/render.cpp
```diff
@@ -1054,7 +1054,7 @@ void gr_rect(int x, int y, int w, int h, int resize_mode, float angle) {
 		float offsetX = x + w / 2.0f;
 		float offsetY = y + h / 2.0f;
 		path->translate(offsetX, offsetY);
-		path->rotate(fl_radians(angle));
+		path->rotate(angle);
```
Goober5000, line 1057: I searched through the codebase for `gr_rect` and it looks like no other functions are affected beyond the scripting ones, which is good.  But you should change the documentation for the `gr_rect` declaration (in render.h).

Goober5000 (changes requested):
One change, otherwise this looks good.

Goober5000 (approved):

PR #3791, opened 2021-11-21
guard against trying to animate non-existent submodels in more places.

EatThePath:
#3732 missed some spots, and currently there are some reliable had crashes as a result. This safeties the ones I'm aware of.

### DIFF
--- a/code/model/modelanimation.cpp
+++ b/code/model/modelanimation.cpp
@@ -35,13 +35,17 @@ namespace animation {
 		
 		if (applyOnly) {
 			for (const auto& animation : m_submodelAnimation) {
-				ModelAnimationData<> base = animation->s_initialData.at({ pmi->id, animation->m_submodel });
-				ModelAnimationData<true>& previousDelta = (*applyBuffer)[animation->m_submodel].second;
-				(*applyBuffer)[animation->m_submodel].first = animation.get();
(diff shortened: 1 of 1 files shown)


code/model/modelanimation.cpp
```diff
@@ -35,13 +35,17 @@ namespace animation {
...
 			for (const auto& animation : m_submodelAnimation) {
-				ModelAnimationData<> base = animation->s_initialData.at({ pmi->id, animation->m_submodel });
-				ModelAnimationData<true>& previousDelta = (*applyBuffer)[animation->m_submodel].second;
-				(*applyBuffer)[animation->m_submodel].first = animation.get();
+				std::pair<int,int> keypair = {pmi->id, animation->m_submodel};
 
-				base.applyDelta(previousDelta);
-				ModelAnimationData<true> delta = animation->play(instanceData.time, instanceData.time, ModelAnimationDirection::FWD, pmi, base, true);
-				previousDelta.applyDelta(delta);
+				if(animation->s_initialData.count(keypair)>0){
```
Baezon, line 40: Pedantic, I know, but please put spaces around the `>` here and below

Baezon (changes requested):

code/model/modelanimation.cpp
EatThePath: Done

Baezon (approved):

PR #3793, opened 2021-11-22
Animation Rework: Phase 3

### LINKED ISSUE #3425: FR: Animate submodel to rotate forwards/backwards, looping
I'd like to be able to setup a subobject to rotate one direction for N seconds and then rotate the opposite direction for N seconds. Then repeat the sequence indefinitely.

Ideally these would also include all the trappings of the Animation code including, set distance, speed, accel/decel.

Maybe this is currently possible, but based on a thorough reading of the [POF specs](https://wiki.hard-light.net/index.php/Subobject_and_subsystem_properties#.24stepped) and the [Animation Code specs](https://wiki.hard-light.net/index.php/Animation_Code) on the Wiki, I conclude that it is not.

BMagnu:
**Features of Phase 3:**
Phase 3 extends the new system of Phase 2 by adding a custom table for animations, and adding more modes than just simple rotation to allow for much more powerful and flexible animations. This now enables arbitrary parallel and sequential combinations of transformations. For details, see [the wiki entry for the new -anim.tbm](https://wiki.hard-light.net/index.php/Animation.tbl)'s.

Secondly, more animation triggers have been added: primary-fired and secondary-fired now allow animation to trigger when a fighters weaponbank was fired, and not only selected. on-spawn provides similar functionality to initial, but will not cause the animation to be interpreted as starting value for a submodel, but rather actually triggering once the ship in question spawns. weapon-launched is similar to on-spawn, but specific for modelled weapon projectiles, of which submodels can now be animated as well. Additionally, the subtype system has been changed, to allow for more intuitive specification of how certain animations trigger.

Finally, animations can now have flags, that change certain specific behaviours. One such flag is the "loop" flag, which allows for an animation to automatically loop once it has completed. This closes #3425.

**Clips in action:**

https://user-images.githubusercontent.com/6238428/142804644-2cbf9ae3-c51f-4825-a640-e583f1efa59d.mp4

Taken from the event-horizon channel on the discord:

https://user-images.githubusercontent.com/6238428/142804688-60e024ca-a5a7-47fe-ade4-1a6337b869dd.mp4

**What is still missing to leave draft:**
~~The translation segment is depending on a PR to implement the submodel translation backend. Without this PR, the segment in question would parse correctly, but just not translate any model.~~
_Skipped, since the Translation Backend won't be ready in time. Translation is implemented, but the parser deactivated for the time being._
~~Furthermore, another round of rebases will be needed once that PR is merged.~~
_Done_


code/model/modelanimation.cpp
```diff
@@ -894,65 +753,112 @@ namespace animation {
...
+	{ModelAnimationTriggerType::PrimaryBank, {"primary-bank", false}},
+	{ModelAnimationTriggerType::SecondaryBank, {"secondary-bank", false}},
+	{ModelAnimationTriggerType::DockBayDoor, {"fighterbay", false}},
+	{ModelAnimationTriggerType::Afterburner, {"afterburner", false}},
+	{ModelAnimationTriggerType::TurretFiring, {"turret-firing", false}},
+	{ModelAnimationTriggerType::Scripted, {"scripted", false}},
+	{ModelAnimationTriggerType::TurretFired, {"turret-fired", true}},
+	{ModelAnimationTriggerType::PrimaryFired,   {"primary-fired", true}},
+	{ModelAnimationTriggerType::SecondaryFired, {"secondary-fired", true}},
+	{ModelAnimationTriggerType::WeaponLaunched, {"weapon-launched", false}} //Atypical case. While no reverse trigger occurs, it is also guaranteed to not trigger more than once per model, hence non-resetting animations are fine here.
```
Goober5000, line 854: Isn't this the same as `on-spawn`, but for weapons rather than ships?

code/model/modelanimation.cpp
```diff
@@ -1011,4 +917,497 @@ namespace animation {
...
+					started |= set.start(pmi, type, dock_name, direction, forced, instant, pause);
+					started |= set.start(pmi, type, "", direction, forced, instant, pause);
+					started |= set.startAll(pmi, type, direction, forced, instant, pause, subtype, true);
+					return started;
+				},
+				[&set, pmi, type, dock_name, subtype]() -> int {
+					int time1 = set.getTime(pmi, type, dock_name);
+					int time2 = set.getTime(pmi, type, "");
+					int time3 = set.getTimeAll(pmi, type, subtype, true);
+					return time1 > time2 ? (time1 > time3 ? time1 : time3) : (time2 > time3 ? time2 : time3);
```
Goober5000, line 949: Why not use `std::max`?

code/model/modelanimation.h
```diff
@@ -145,81 +150,85 @@ namespace animation {
...
+	private:
+		//Polymodel Instance ID -> ModelAnimationData
+		std::map<int, ModelAnimationData<>> m_initialData;
 
-		friend class ModelAnimation;
 	public:
-		//Create a submodel animation based on the name of the submodel
-		ModelAnimationSubmodel(SCP_string submodelName, std::shared_ptr<ModelAnimationSegment> mainSegment);
-
+		ModelAnimationSubmodel(SCP_string submodelName);
```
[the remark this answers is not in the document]
BMagnu: Clang will then complain that since the value is just stored in a member variable (and thus copied anyways), it should use call-by-value and then std::move() in the implementation (as it is currently)

code/model/modelanimation.h
```diff
@@ -145,81 +150,85 @@ namespace animation {
...
 		*/
-		ModelAnimationSubmodelTurret(SCP_string subsystemName, bool findBarrel, SCP_string SIPname, std::shared_ptr<ModelAnimationSegment> mainSegment);
 
-		ModelAnimationSubmodel* copy(const SCP_string& newSIPname) override;
+		void renameSIP(const SCP_string& newSIPname) override;
+		ModelAnimationSubmodel* copy() const override;
+
+		friend class ModelAnimationSet;
+	public:
+		ModelAnimationSubmodelTurret(SCP_string subsystemName, bool findBarrel, SCP_string SIPname);
```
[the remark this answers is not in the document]
BMagnu: Same with above

code/model/modelanimation.h
```diff
@@ -228,68 +237,137 @@ namespace animation {
...
+		static void cleanRunning();
+
+		void initializeSubmodelBuffer(polymodel_instance* pmi, ModelAnimationSubmodelBuffer& applyBuffer) const;
+
+		friend class ModelAnimation;
+		friend class ModelAnimationParseHelper;
+	public:
+		std::map <std::pair<ModelAnimationTriggerType, int>, std::map <SCP_string, std::vector<std::shared_ptr<ModelAnimation>>>> m_animationSet;
+
+		ModelAnimationSet(SCP_string SIPname = "");
```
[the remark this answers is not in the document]
BMagnu: Same with above

code/model/modelanimation.cpp
BMagnu: Yes, it is.
I'll change on-spawn to replace weapon-launched

code/model/modelanimation_segments.cpp
```diff
@@ -53,93 +53,203 @@ namespace animation {
...
+	}
+
+	ModelAnimationParseHelper::Registrar ModelAnimationSegmentSetPHB::reg("$Set Orientation:", &parser);
+	std::shared_ptr<ModelAnimationSegment> ModelAnimationSegmentSetPHB::parser(ModelAnimationParseHelper* data) {
+		angles angle;
+		bool isRelative = true;
+
+		required_string("+Angle:");
+		stuff_angles_deg_phb(&angle);
+		isRelative &= !optional_string("+Absolute");
```
Goober5000, line 219: I think it would be nice to allow the modder to specify either `+Absolute` or `+Relative` here

code/model/modelanimation_segments.cpp
```diff
@@ -387,23 +537,392 @@ namespace animation {
...
+	std::shared_ptr<ModelAnimationSegment> ModelAnimationSegmentRotation::parser(ModelAnimationParseHelper* data) {
+		optional<angles> angle, velocity, acceleration;
+		optional<float> time;
+		bool isAbsolute = false;
+
+		if (optional_string("+Angle:")) {
+			angles parse;
+			stuff_angles_deg_phb(&parse);
+			angle = parse;
+			isAbsolute |= (bool)optional_string("+Absolute");
```
Goober5000, line 558: same here

code/model/modelanimation_segments.cpp
```diff
@@ -387,23 +537,392 @@ namespace animation {
...
+			}
+		}
+		else { //Calc time
+			float& duration = m_duration[pmi->id] = 0.0f;
+
+			vec3d& v = instanceData.m_actualVelocity;
+			const vec3d& d = instanceData.m_actualTarget;
+
+			vec3d actualAccel{ {{ 0,0,0 }} };
+			vec3d accelTime{ {{0,0,0 }} };
```
Goober5000, line 711: the whitespace is different than the previous and subsequent lines

code/model/modelanimation_segments.cpp
```diff
@@ -387,23 +537,392 @@ namespace animation {
...
+
+		if (CoordinateSystem::COORDS_LOCAL_AT_START == m_coordType) {
+			instanceData.m_rotationAtStart = base[m_submodel].first.orientation;
+		}
+	}
+
+	void ModelAnimationSegmentTranslation::calculateAnimation(ModelAnimationSubmodelBuffer& base, float time, int pmi_id) const {
+		const instance_data& instanceData = m_instances.at(pmi_id);
+
+		vec3d currentOffset{ {{ 0,0,0}} };
```
Goober5000, line 769: inconsistent whitespace here too

code/model/modelanimation_segments.h
```diff
@@ -91,22 +116,62 @@ namespace animation {
...
+
+		//PMI ID -> Instance Data
+		std::map<int, instance_data> m_instances;
+
+		std::shared_ptr<ModelAnimationSubmodel> m_submodel;
+
+		optional<vec3d> m_target;
+		optional<vec3d> m_velocity;
+		optional<float> m_time;
+		optional<vec3d> m_acceleration;
```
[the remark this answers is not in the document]
BMagnu: This isn't std optional, but rather my own tiny implementation of an optional in modelanimation.h
But yeah, having an _actual_ std::optional would be preferrable, but not as long as we're stuck with C++11

code/model/modelanimation_segments.cpp
```diff
@@ -53,93 +53,203 @@ namespace animation {
...
+	
+	void ModelAnimationSegmentWait::recalculate(ModelAnimationSubmodelBuffer& /*base*/, polymodel_instance* pmi) {
+		m_duration[pmi->id] = m_time;
 	};
 
+	ModelAnimationParseHelper::Registrar ModelAnimationSegmentWait::reg("$Wait:", &parser);
+	std::shared_ptr<ModelAnimationSegment> ModelAnimationSegmentWait::parser(ModelAnimationParseHelper* /*data*/) {
+		required_string("+Time:");
+		float time = 0.0f;
+		stuff_float(&time);
```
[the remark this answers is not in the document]
BMagnu: Since the whole animation system currently uses floating point time, it's probably not currently useful.
_Should_ precision become an issue here, which I doubt, then it might be wise to retool this to fixed-point values.

Goober5000 (approved):

PR #3796, opened 2021-11-23
Fix Weapon-Loadout-Specific Models

BMagnu:
By specifying a different techroom model, the weapon loadout screen can display a different model than used ingame.
However, when loaded it would also reinitialize the model_subsystems of the ship_info, and thus change the model_num of the subsystems. This causes a bunch of problems down the line. Since the recent change to subsystem rotation handling, a corrupt model_num can just cause straight CTD's on mission start.

This addresses this by not loading the ship_info's subsystems with the techmodel, but a seperate submodel buffer, thus not erroniously overwriting the subsystem data.

### DIFF
--- a/code/missionui/missionweaponchoice.cpp
+++ b/code/missionui/missionweaponchoice.cpp
@@ -416,7 +416,6 @@ UI_XSTR Weapon_select_text[GR_NUM_RESOLUTIONS][WEAPON_SELECT_NUM_TEXT] = {
 	}
 };
 
-
 ///////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


Goober5000 (changes requested):
Instead of using a temporary subsystem buffer, why not just pass `0, nullptr` to `model_load`?  It looks like this is done in several other places in the code already.

BMagnu:
Fair, that works too.

Goober5000 (approved):

JohnAFernandez (approved):

PR #3801, opened 2021-11-28
Lt22

EatThePath:
Adds selectable tonemapping functions to the shaders and a new table to control it.
This PR adds only fixed static control over these options, one 'profile' only. The code is a bit overengineered for that, but I intend to add support for multiple profiles and features to use them in the future, as well as adding control of more parts of the lighting pipeline. Ran out of time before the freeze for all that.

### DIFF
--- a/code/lighting/lighting_profiles.cpp
+++ b/code/lighting/lighting_profiles.cpp
@@ -1,0 +1,158 @@
+#include "def_files/def_files.h"
+#include "globalincs/pstypes.h"
+#include "globalincs/safe_strings.h"
+#include "globalincs/vmallocator.h"
+#include "lighting/lighting.h"
+#include "lighting/lighting_profiles.h"
+#include "osapi/dialogs.h"
+#include "parse/parselo.h"
+
+
+bool optional_parse_into_float(const SCP_string &fieldname, float* valuetarget){
+	if(optional_string(fieldname.c_str())){
+		stuff_float(valuetarget);
+		return true;
+	}
+	return false;
+}
+
+void light_profile::reset(){
+
+    static_light_factor = 1.0f;
+    static_tube_factor = 1.0f;
+    static_point_factor = 1.0f;
+    tonemapper = tnm_Uncharted;
+
+	ppc_values.toe_strength = 0.5f;
+    ppc_values.toe_length = 0.5f;
+    ppc_values.shoulder_strength = 0.0f;
+    ppc_values.shoulder_length = 0.5f;
+    ppc_values.shoulder_angle = 0.1f;
+	exposure = 4.0f;
+}
+
+TonemapperAlgorithm light_profile_manager::name_to_tonemapper(SCP_string &name){
+	SCP_tolower(name);
+	TonemapperAlgorithm r = tnm_Invalid;
+	if(name == "uncharted" || name == "uncharted 2" ){
+		r = tnm_Uncharted;
+	}
+	if(name == "linear"){
+		r = tnm_Invalid;
+	}
+	if(name == "aces"){
+		r = tnm_Invalid;
+	}
+	if(name == "aces approximate"){
+		r = tnm_Invalid;
+	}
+	if(name == "cineon"){
+		r = tnm_Invalid;
+	}
+	if(name == "cineon"){
+		r = tnm_Invalid;
+	}
+	if(name == "reinhard jodie"){
+		r = tnm_Invalid;
+	}
+	if(name == "reinhard extended"){
+		r = tnm_Invalid;
+	}
+	if(name == "PPC"){
+		r = tnm_Invalid;
+	}
+	if(name == "PPC RGB"){
+		r = tnm_Invalid;
+	}
+	return r;
+}
+
+light_profile default_profile;
+void light_profile::load_profiles(){
+	parse_all();
+}
+
+//The logic for grabbing all the parseable files
+
+void light_profile::parse_all(){
+	default_profile.reset();
+	if (cf_exists_full("lighting_profiles.tbl", CF_TYPE_TABLES)){
+		mprintf(("TABLES:Starting parse of lighting profiles.tbl"));
+		light_profile::parse_file("lighting_profiles.tbl");
+	}
+	
+	mprintf(("TBM  =>  Starting parse of lighting profiles ...\n"));
+	parse_modular_table("*-ltp.tbm", light_profile::parse_file);
+
+}
+
+void parse_default_section(const char *filename){
+	bool keep_going = true;
+	SCP_string buffer;
+	TonemapperAlgorithm tn;
+	while(!optional_string("#END DEFAULT PROFILE") && keep_going){
+		keep_going = false;
+		if(optional_string("$Tonemapper:")){
+			stuff_string(buffer,F_NAME);
+			tn = light_profile_manager::name_to_tonemapper(buffer);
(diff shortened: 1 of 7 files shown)


code/lighting/lighting_profiles.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+//Handle an individual file.
+void lighting_profile::parse_file(const char *filename)
+{
+	try
+	{
+		if (filename == nullptr){
+			//Normal parsing uses this:
+			//read_file_text_from_default(defaults_get_file("lighting_profiles.tbl"));
+			//To set the defaults, I intend to set them as virtual table lines before here
```
BMagnu, line 123: Is this still correct?
It seems to me like the defaults are applied by just using the reset.

code/lighting/lighting_profiles.cpp
```diff
@@ -0,0 +1,166 @@
...
+	mprintf(("TBM  =>  Starting parse of lighting profiles ...\n"));
+	parse_modular_table("*-ltp.tbm", lighting_profile::parse_file);
+}
+
+void lighting_profile::parse_default_section()
+{
+	bool keep_going = true;
+	SCP_string buffer;
+	TonemapperAlgorithm tn;
+	while(!optional_string("#END DEFAULT PROFILE") && keep_going){
```
BMagnu, line 96: Why is this a loop?
Since this is specifically the _default_ profile, it seems to me like it should only be possible to have one entry.

code/lighting/lighting_profiles.cpp
```diff
@@ -0,0 +1,166 @@
...
+			if(optional_string("#DEFAULT PROFILE")){
+				keep_going = true;
+				parse_default_section();
+			}
+			//TODO: Handle case when there's no line matched but we haven't hit an #end
+			if(!keep_going)
+			{
+				advance_to_eoln(nullptr);
+				keep_going = true;
+			}
```
BMagnu, line 140: This looks like it could cause problems:
keep_going is never set to false, so when you have a table that has an additional newline or something after the default profile end, you'll never advance_to_eoln and never reach the end of the file.
Seems like keep_going should be set to false at the beginning of the loop.
Though in all honesty, I'd just remove the loop here as well, and make it into a single ``if(optional_string("#DEFAULT PROFILE"))``, eventually possibly followed by a single ``if(optional_string("#SELECTABLE PROFILE")`` or whatever. There's no need for the overarching sections to exist more than once, especially for the default profile section.

code/lighting/lighting_profiles.h
```diff
@@ -0,0 +1,47 @@
...
+	float exposure;
+
+    void reset();
+
+private:
+	static lighting_profile default_profile;
+	static void parse_all();
+	static void parse_file(const char *filename);
+	static void parse_default_section();
+	static void add_default_default();
```
BMagnu, line 46: This function does not exist, the declaration should go too.

code/lighting/lighting_profiles.h
```diff
@@ -0,0 +1,47 @@
...
+	float shoulder_length;
+	float shoulder_angle;
+};
+
+class lighting_profile{
+public:
+	static enum TonemapperAlgorithm name_to_tonemapper(SCP_string &name);
+	static void load_profiles();
+	static TonemapperAlgorithm current_tonemapper();
+	static piecewise_power_curve_values current_piecewise_values();
```
BMagnu, line 31: I'd suggest for this to return a ``const piecewise_power_curve_values &``.

BMagnu (commented):
A few comments on the FSO-side of the code.
As for the shader-side, I don't really feel qualified to judge if the actual tonemapper algorithms are correct. Purely syntactic it seems fine to me (though I am not the biggest fan of the undocumented magic constants and single-letter variables), but I am no expert and would like some more reviews for the shaders.

The-E:
> 

The shaders are fine as they are. While it's a bit off-putting to see single-letter variable names and magic constants, this is somewhat unavoidable - A lot of them are derived from graphics coding papers that provide explanations for the various formulas, and as we all know there is nothing academic mathematicians hate more than proper variable naming 😁 Renaming those variables wouldn't make sense, as it would just make the task of cross-referencing with their reference implementations harder.

code/lighting/lighting_profiles.cpp
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if(name == "reinhard extended"){
+		r = tnm_Invalid;
+	}
+	if(name == "PPC"){
+		r = tnm_Invalid;
+	}
+	if(name == "PPC RGB"){
+		r = tnm_Invalid;
+	}
```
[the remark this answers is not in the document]
EatThePath: Er, yes. Pure oversight after some code reorg, whoops. Will fix.

code/lighting/lighting_profiles.cpp
EatThePath: That's largely leftover from a fancier branch, will make sure to revise.

code/lighting/lighting_profiles.cpp
EatThePath: To the latter, looping allows for the sections to be in any order. That said, given there's only one section currently it can probably go from this version.

EatThePath:
>  While it's a bit off-putting to see single-letter variable names and magic constants, this is somewhat unavoidable - A lot of them are derived from graphics coding papers that provide explanations for the various formulas, and as we all know there is nothing academic mathematicians hate more than proper variable naming 😁 Renaming those variables wouldn't make sense, as it would just make the task of cross-referencing with their reference implementations harder.

Yeah, I really hate things like `float white = ((W * (A * W + C * B) + D * E) / (W * (A * W + B) + D * F)) - E / F;` myself, but it's pretty much what we're stuck with in this domain. 

I'll put some reference URLs in the shader code, at least.

BMagnu:
> I'll put some reference URLs in the shader code, at least.

That's great, if there's a link to a paper or reference code that uses the same names, it's perfectly fine.
Otherwise it'd need some explanation comments IMO, so that it's still possible to understand what is going on there at a later point in time.

code/lighting/lighting_profiles.h
EatThePath: Correct me if I'm wrong, but that would the calling code access to the actual underlying data structure, while the current code creates and returns a copy of the struct's values? I would prefer to return only the values and make sure nothing can accidentally tweak the profile data in the future.

BMagnu: What this does is return a reference to the values of the underlying data structure, preventing an unnecessary copy.
But since you return a ``const`` reference and not just a reference, it cannot be used to modify the underlying data, only allowing read-only access. Since this both prevents modification and unnecessary copying, returning by const & should always be preferable to returning by copy unless you either need the copy every time, or you return something that's allocated on the stack (a variable declared within the function). Since neither is true here, you can and should return a const reference.

EatThePath: Okay, that makes sense then.

EatThePath:
Well I'd convert this to full now if I'd actually remembered to submit it in draft.

Testing revealed a few blips I've cleaned up, and added a bit more documentation. If it passes review now I'm calling this phase of it done and mergable.

BMagnu (approved):
Looks good now.
And with E also saying that the shaders look fine, I am ready to approve it.

PR #3803, opened 2021-11-30
Expose turret re-targeting delay

Baezon:
This the time at which a turret will reacquire a target, maybe finding a new one, maybe re-affirming the current one, etc. If you have very fast missiles and very fast CIWS even 2 seconds can be too long a delay, but potentially comes with performance problems, as the check that it triggers has to search all ships and weapons, and is done per turret.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -581,6 +581,16 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$reset last_hit_target_time for player hits:", AI::Profile_Flags::Reset_last_hit_target_time_for_player_hits);
 
+				if (optional_string("$Turret target recheck time:"))
+				{
+					stuff_float(&profile->turret_target_recheck_time);
+					if (profile->turret_target_recheck_time < 0) {
+						Warning(LOCATION, "Turret target recheck time must be positive.");
+						profile->turret_target_recheck_time = 2000.0f;
+					}
+				}
+
+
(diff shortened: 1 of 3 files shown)


code/ai/ai_profiles.cpp
```diff
@@ -581,6 +581,16 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$reset last_hit_target_time for player hits:", AI::Profile_Flags::Reset_last_hit_target_time_for_player_hits);
 
+				if (optional_string("$Turret target recheck time:"))
+				{
+					stuff_float(&profile->turret_target_recheck_time);
+					if (profile->turret_target_recheck_time < 0) {
```
Goober5000, line 587: Is there a sensible minimum value we want to enforce for the recheck time?

Baezon: Theoretically, being just a timestamp, anything >= 0 is logically valid and I haven't seen any issues in my test cases with 0. But practically, EatThePath using this with 0 delay resulted in frequent stuttering (and not just poor framerate), which is a bit unusual to me as I would've expected merely low framerate, and recommended 15-30ms as a minimum for no stuttering. Ideally though, I would like to not limit this unless there are actual logic errors with it, and leave potential performance concerns to warning labels on the wiki documentation.

Goober5000: Fair enough.

Goober5000 (approved):

PR #3804, opened 2021-12-01
set the nebula mission flag when setting the nebula

Goober5000:
Several other places in the code check the mission flag, so make sure it accurately reflects the state of the background.


Baezon:
Is this not what `stars_set_nebula()` does in its first line?

Goober5000:
Ugh, it does.  Now I need to figure out why my debugging thought the flag wasn't set...

Closing as redundant.

PR #3806, opened 2021-12-03
Fix the Tonemapper strings

BMagnu:
since the content of the name these are matched against is converted to lowercase, the builtin names also need to be lowercase to be able to match.


EatThePath (approved):
oops

wookieejedi (approved):
Good catch!

PR #3811, opened 2021-12-07
Set Default Detail Levels to be Consistent at Ultra Settings

wookieejedi:
By default, new pilots have `Model` and `Nebula` detail only set to 4/5 (all others are on 5/5). Furthermore, clicking `Very High` on the model presets options still does not increase the `Model` or `Nebula` detail, as they remain at 4/5. This occurs with making a new pilot in retail, along with making new pilots in mods with a game settings default detail level of `Ultra`.

Simply changing the default values in the code for `Model` and `Nebula` detail to match the other detail levels is the easiest and cleanest solution IMO. I would be happy to alter this though to use a game settings table flag instead, if that is preferred.

Tested and works as expected.

### DIFF
--- a/code/globalincs/systemvars.cpp
+++ b/code/globalincs/systemvars.cpp
@@ -243,8 +243,8 @@ detail_levels Detail_defaults[NUM_DEFAULT_DETAIL_LEVELS] = {
 	{				// Highest level
 		3,			// setting
 					// ===== Analogs (0-MAX_DETAIL_LEVEL) ====
-		3,			// nebula_detail;				// 0=lowest detail, MAX_DETAIL_LEVEL=highest detail
-		3,			// detail_distance;			// 0=lowest MAX_DETAIL_LEVEL=highest		
+		4,			// nebula_detail;				// 0=lowest detail, MAX_DETAIL_LEVEL=highest detail
+		4,			// detail_distance;			// 0=lowest MAX_DETAIL_LEVEL=highest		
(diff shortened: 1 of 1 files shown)


Goober5000:
This is fine with me, I think, though it would be good for others to chime in.

Can you clarify what you mean about `Ultra` - do you mean that this is a new detail level you added?  Also, is the `Ultra` behavior to set the detail to 5/5, and is it independent of this PR?

wookieejedi:
Good question. I should remedy that initial statement, wording with Ultra was likely incorrect; I mean when the settings are set to `Very High` in either the game settings table (`$Default Detail Level: 3`) or clicking the `Very High` radio button in the detail config screen. In either of those cases all details levels go to 5/5 except model and nebula detail which only go to 4/5. 

I think like it makes most sense to put all slides on the highest detail if `Very High` detail (the highest option) is clicked/set in the game settings table.

Baezon:
I am in favor. These various settings are probably pretty arbitrary, and if they aren't must presumably be based on the performance impacts of the techniques used on the various hardware at the time, and are almost certainly completely irrelevant now with in many cases vastly changed techniques and completely new hardware.

And in any case, as wookieejedi pointed out, it's very unintuitive to the user that a maximum detail setting does not put maximums on all values.

Goober5000 (approved):
Works for me.  With two votes plus the author, I'll approve this.

PR #3817, opened 2021-12-14
improve PXO data validation code

notimaginative:
Adds ability to validate table and mission files in batches to reduce the number of network transfers, substantially decreasing the time needed to validate files.


Goober5000 (approved):
Looks ok to me.

JohnAFernandez:
I don't think my motivation has yet returned to review something of this magnitude. >_>

notimaginative:
It looks more complicated than it is, I think, as this replaces some of the older code. The complicated bits are just working around the limitations of the retail code as much as possible without breaking compatibility. Hence the special (de)serialization of the new packets. The simple option was to just limit it to 10 files max at a time which would fit with the retail way of doing things, but I figured someone would outgrow that soon enough.

PR #3820, opened 2021-12-19
reduce memory required for bsp alignment

notimaginative:
Previous code allocated 4x bsp_data_size for alignment, whether it was
needed or not. Now it should allocate slightly more than 2x and only if
alignment is required.


Goober5000 (commented):
I think this looks ok but I'd also like to get @Shivansps to look at it.

Shivansps:
Looks good to me. i really like that nice little trick to check if the model is aligned or not before assigning extra memory.

Goober5000 (approved):

PR #3837, opened 2021-12-20
improve SEXP error handling a bit

Goober5000:
1) Sync formatting of each place SEXP errors are reported
2) Add error location (event or goal) if possible when mission is loaded

Note that the two FRED error dialogs already indicate the event or goal.


JohnAFernandez (approved):
Seems simple enough

PR #3838, opened 2021-12-21
Removed consts being set from uniforms to fix #3821

### LINKED ISSUE #3821: Recent tonemapper change causes fatal shader error on Linux
Changes from PR #3801 cause the tonemapper shader to fail to compile on Linux which results in a fatal error.

`0:113(27): error: initializer of const variable 'toeStrength' must be a constant expression`

The error is generated for lines 113-117 and appears to be due to `const` not being valid for a variable initialized by a uniform. Simply removing the `const` allows everything to work as normal. My GLSL knowledge is at least 10 years out of date however so there may be a more appropriate fix.

This *may* be a Linux driver bug but I have not confirmed that so far.

```  OpenGL Vendor    : Intel Open Source Technology Center
  OpenGL Renderer  : Mesa DRI Intel(R) HD Graphics 615 (KBL GT2)
  OpenGL Version   : 4.6 (Core Profile) Mesa 20.3.5

  Max texture units: 2 (32)
  Max client texture states: 2 (1)
  Max elements vertices: 3000
  Max elements indices: 3000
  Max texture size: 16384x16384
  Max render buffer size: 16384x16384
  Can use compressed textures: YES
  Texture compression available: YES
  Post-processing enabled: YES
  Using trilinear texture filter.
  OpenGL Shader Version: 4.60
  Max uniform block size: 65536
  Max uniform buffer bindings: 84
  Uniform buffer byte offset alignment: 32
```

EatThePath:
fixes immediate fatal crash described in #3821

### DIFF
--- a/code/def_files/data/effects/tonemapping-f.sdr
+++ b/code/def_files/data/effects/tonemapping-f.sdr
@@ -109,25 +109,25 @@ vec3 linearToneMapping(vec3 hdr_color)
 	hdr_color = linear_to_srgb(hdr_color); // return from linear color space to SRGB color space
 	return hdr_color;
 }
-
-const float toeStrength = toeS; // [0-1]
-const float toeLength = toeL; // [0-1]
-const float shoulderStrength = shoulderS; // in F stops
-const float shoulderLength = shoulderL; // [0-1]
-const float shoulderAngle = shoulderA; // [0-1]
-
+/*
+ float toeStrength = toeS; // [0-1]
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
Noting here for convivence from asarium `All those computations need to be moved into a function (maybe stored in a struct for easier passing around in the shader) to fix this properly.` Approving with agreement that such a rework will be incorporated soon and before stable.

Issue #3644, opened 2021-09-06
Possibly misleading APNG load errors.


EatThePath:
The attached thruster png in blue planet generates a warning in the debug log, I believe every time it is referenced: 
[thruster02-04.zip](https://github.com/scp-fs2open/fs2open.github.com/files/7117873/thruster02-04.zip)
the error is:
`Failed to load apng: (file thruster02-04.png) animation duration <= 0.0f, bad data?`
I also get that error with several recent png beam sections in warmachine, sections which I am fairly confident are actually loaded and displayed properly. I am not certain of what the thrusters are supposed to look like so I'm not sure but I haven't noticed any obvious problems with them.
If they're trying to load as APNG and then falling back when they don't, messages in the log to that effect would be good, because as-is it just looks like the texture is failing to load.

z64555:
Needs a good look at to determine if its a bug or a not-a-bug for next RC.

jg18:
Fixed with PR #3839. Thanks @EatThePath for reporting this issue.

PR #3839, opened 2021-12-21
Clarify logging when loading thruster textures as static images

jg18:
Log when falling back to loading a thruster texture as a static image instead of animated, and log when loading the texture fails entirely.

Fixes issue #3644.

### DIFF
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -1424,7 +1424,7 @@ static int bm_load_image_data(int handle, int bpp, ushort flags, bool nodebug)
 	return 0;
 }
 
-int bm_load_animation(const char *real_filename, int *nframes, int *fps, int *keyframe, float *total_time, bool can_drop_frames, int dir_type) {
+int bm_load_animation(const char *real_filename, int *nframes, int *fps, int *keyframe, float *total_time, bool can_drop_frames, int dir_type, bool rethrow_exceptions) {
 	int	i, n;
 	anim	the_anim;
 	CFILE	*img_cfp = nullptr;
@@ -1594,10 +1594,15 @@ int bm_load_animation(const char *real_filename, int *nframes, int *fps, int *ke
 			img_size = the_apng.imgsize();
 		}
 		catch (const apng::ApngException& e) {
-			mprintf(("Failed to load apng: %s\n", e.what() ));
 			if (img_cfp != nullptr)
 				cfclose(img_cfp);
-			return -1;
+			if (rethrow_exceptions) {
+				throw e;
+			}
+			else {
+				mprintf(("Failed to load apng: %s\n", e.what()));
+				return -1;
+			}
 		}
 	}
 	else {
--- a/code/bmpman/bmpman.h
+++ b/code/bmpman/bmpman.h
@@ -289,7 +289,14 @@ bool bm_release_rendertarget(int handle);
  * @returns The bm number of the first bitmap in the sequence if successful, or
  * @returns A negative value if unsuccessful
  */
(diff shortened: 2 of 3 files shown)


Goober5000 (commented):
These logging statements look good, but is the "Failed to load apng" message still displayed anywhere?

Baezon:
This doesn't seem to affect the "animation duration <= 0.0f, bad data?" message, will it not continue to print that as well? For that matter, should it be logged at all if a png isn't interpretable as animated? In this case that it is static, a normal and extremely common scenario, it's potentially a lot of logging for a completely expected (for the modder) scenario.

jg18:
Yeah, the "failed to load apng" message will still print to the log.

I don't know why the "failed to load apng" message doesn't appear for thruster02-01.png, even though it's also a static image. It must have something to do with the actual bytes inside the two .png files and how the APNG loader is (mis)interpreting them.

@Baezon Fair point that this is a common and expected occurrence and that logging this will clog up the logs.

I think the only way to ensure the "Could not load as animation" message is logged *only* when there was a serious issue with loading as animation is to rethrow the caught exception and then have the caller catch the exception. But rethrowing would require handling all call sites for `bm_load_animation()`.

Modified idea: we could add an extra `bool` argument to bm_load_animation() to rethrow caught exceptions. The callers can then choose whether they want to handle them.

jg18:
If we don't log anything at all when a .png file fails to load as APNG, then we won't know if the .png was just a static PNG file or if it was a malformed APNG file.

Baezon:
Hopefully a malformed APNG would not be able to be a valid static PNG, but if it can (or more accurately, is a common or likely scenario) then it's certainly better to log whatever the result is, than be always silent.

Goober5000:
> Hopefully a malformed APNG would not be able to be a valid static PNG, but if it can (or more accurately, is a common or likely scenario)

It can, and not only that, it's designed to be.

jg18:
Updated PR with the modified idea I proposed yesterday. Now the log line will appear only if `bm_load_animation()` throws an exception. Thoughts?

Baezon (approved):
Works for me. I was afraid that a malformed APNG and a valid PNG might be indistinguishable, so best we can do is be clearer about what went wrong/what's happening.

code/bmpman/bmpman.cpp
```diff
@@ -1594,10 +1594,15 @@ int bm_load_animation(const char *real_filename, int *nframes, int *fps, int *ke
...
 		catch (const apng::ApngException& e) {
-			mprintf(("Failed to load apng: %s\n", e.what() ));
 			if (img_cfp != nullptr)
 				cfclose(img_cfp);
-			return -1;
+			if (rethrow_exceptions) {
+				throw e;
+			}
+			else {
+				mprintf(("Failed to load apng: %s\n", e.what()));
```
Goober5000, line 1603: Add the file name to the log message

Goober5000 (changes requested):

Goober5000:
@jg18  Looks good, just one more request

code/bmpman/bmpman.cpp
jg18: The file name is already included in the exception message. See `_apng_failed()` in pngutils.cpp.

Goober5000: Oh.  Well then, is it also (redundantly) included in the exception messsage on line 4484?

jg18: In ship.cpp? It is. My reasoning for explicitly including it there was that that call site could, in the future, catch exceptions other than `ApngException`, and those future exceptions might not include the file name in the exception message. Thus I figured it's better to tolerate the redundancy in that case.

Goober5000: Gotcha

Goober5000 (approved):

Issue #3840, opened 2021-12-22
Sphere-type particle effects are not quite uniform across the sphere


Baezon:
[test-wep.txt](https://github.com/scp-fs2open/fs2open.github.com/files/7759877/test-wep.txt)
[test-part.txt](https://github.com/scp-fs2open/fs2open.github.com/files/7759879/test-part.txt)

These cause subach bolts to explode into a bazillion particles in all directions, what should be uniformly randomly in all directions but this the result:
(viewed from the interior of the spherical shell of particles)
![screen0012](https://user-images.githubusercontent.com/17305613/147033059-7ff7420b-a32e-423d-a861-3de9f8b94582.png)
(directly behind looks the same)

Gonna call this low priority because this is subtle enough that you really need a lot of particles to even tell this is happening.

jg18:
Fixed with PR #3841.

PR #3841, opened 2021-12-22
Attempt to fix random selection of points inside/on spheres

jg18:
Also compute/convert from spherical coordinates in a way that's easier to understand.

Intended to address Issue #3840.


Baezon (approved):
Looks good to me. Were you able to test and see that it's working?

jg18:
@Baezon havent yet tested, although it would be tough for me to tell if there were still noticeable patterns. Would someone else be willing to give this a go?

Baezon (approved):
A re-approval after some algorithm consternation, and verified through testing.

PR #3843, opened 2021-12-22
update polymodel_instance to reference object

Goober5000:
See title.  All references to `model_create_instance` have been updated to supply an object number or -1.


Baezon (commented):
Is there any particular reason for this change? The pmi doesn't appear to use this objnum.

Goober5000:
This is in preparation for @BMagnu 's multiplayer fixes.  There are several places where he needs to access the object but all he has is the polymodel instance.  This solves the problem easily and also is a sensible piece of information to add to the struct.

Baezon (approved):

PR #3844, opened 2021-12-23
fix multi awacs crash due to code init order

notimaginative:
Fix crash related to PR #3634 changes due to when the awacs info is setup. Awacs is setup/updated on each game frame. However this happens after multi packets are processed that frame, so a client can send related info to the server before the server has had the chance to init awacs causing a crash due to the vector being empty. This should provide a safety catch for when such an out of order issue occurs.


Goober5000 (approved):
Simple, and it doesn't look like anything would be harmed by calling this function eagerly in this case.

PR #3846, opened 2021-12-23
Animation Rework Phase 3 - Fixes for Multi

### LINKED ISSUE #3836: New animation code causes crashes in multi
I completely missed this in the original PR for the phase 2 animations, but when the animation packet is sent it uses `animationId` and `pmi`, both of which are only valid in the local instance. Neither index is necessarily present on a different client or the server and either causes the animation to do nothing or more than likely to just crash the server/client.

The packet should use the net signature of the object instead of the pmi and from that the proper local model instance can be retrieved (`multi_get_network_object()` followed by `object_get_model_instance()`). I'm not sure about the animation id however as I don't know the animation enough.

### LINKED ISSUE #3842: Multiple log statements stacked in one line
I see multiple "Specified deprecated non-initial type animation on subsystem ___ of ship class ____. Consider using *-anim.tbm's instead."  in one line in fs2_open.log.
It's a bit annoying when I don't use word wrapping.
I believe #3793 caused this.

BMagnu:
This PR changes the multi message for the Animations to use netcode signatures instead of PMI ID's.
At the same time, protection against nonexistant animations or objects was added.
Furthermore, animation ID's are now only a function of a unique animation name and a name of whatever class this animation is defined for. This means, that for an animation that has the same name and is used in the same ship or weapon class, the id will be the same as well, making it usably stable for multi.
Closes #3836 
Small logging fixes to close #3842 as well.

### DIFF
--- a/code/model/modelanimation.cpp
+++ b/code/model/modelanimation.cpp
@@ -2,6 +2,7 @@
 #include "model/modelanimation_segments.h"
 #include "network/multi.h"
 #include "network/multimsgs.h"
+#include "parse/encrypt.h"
 #include "ship/ship.h"
 #include "utils/Random.h"
 
@@ -20,7 +21,7 @@ namespace animation {
 	const size_t Num_animation_flags = sizeof(Animation_flags) / sizeof(flag_def_list_new<animation::Animation_Flags>);
 
 	std::map<int, ModelAnimationSet::RunningAnimationList> ModelAnimationSet::s_runningAnimations;
-	std::vector<std::shared_ptr<ModelAnimation>> ModelAnimation::s_animationById;
+	std::map<unsigned int, std::shared_ptr<ModelAnimation>> ModelAnimationSet::s_animationById;
 
 	ModelAnimation::ModelAnimation(bool isInitialType) : m_set(nullptr), m_isInitialType(isInitialType) { }
 
@@ -139,7 +140,7 @@ namespace animation {
 			//We are in multiplayer. Send animation to server to start. Server starts animation online, and sends start request back (which'll have multiOverride == true).
 			//If we _are_ the server, also just start the animation
 
-			object* objp = nullptr;
+			object* objp = pmi->objnum > -1 ? &Objects[pmi->objnum] : nullptr;
 
 			if(objp != nullptr)
 				send_animation_triggered_packet(id, objp, 0, direction, force, instant, pause);
@@ -515,18 +516,31 @@ namespace animation {
 		return *this;
 	}
 
-	void ModelAnimationSet::emplace(const std::shared_ptr<ModelAnimation>& animation, const SCP_string& name, ModelAnimationTriggerType type, int subtype) {
+	void ModelAnimationSet::emplace(const std::shared_ptr<ModelAnimation>& animation, const SCP_string& name, ModelAnimationTriggerType type, int subtype, unsigned int uniqueId) {
 		auto newAnim = std::shared_ptr<ModelAnimation>(new ModelAnimation(*animation));
 		newAnim->m_set = this;
 		newAnim->m_animation = std::shared_ptr<ModelAnimationSegment>(animation->m_animation->copy());
 		newAnim->m_animation->exchangeSubmodelPointers(*this);
+		newAnim->id = uniqueId;
+		ModelAnimationSet::s_animationById[uniqueId] = newAnim;
 		m_animationSet[{type, subtype}][name].push_back(newAnim);
 	}
 
 	void ModelAnimationSet::changeShipName(const SCP_string& name) {
+		unsigned int hash_change = hash_fnv1a(m_SIPname) ^ hash_fnv1a(name);
+
 		m_SIPname = name;
 		for (const auto& submodel : m_submodels) {
 			submodel->renameSIP(name);
+		}
+
+		for (const auto& animationTypes : m_animationSet) {
+			for (const auto& animations : animationTypes.second) {
+				for (const auto& animation : animations.second) {
+					animation->id ^= hash_change;
+					ModelAnimationSet::s_animationById[animation->id] = animation;
+				}
+			}
 		}
 	}
 
@@ -1198,16 +1212,16 @@ namespace animation {
 
 	}
 
-
 	void ModelAnimationParseHelper::parseTables() {
 		s_animationsById.clear();
 		parse_modular_table("*-anim.tbm", parseTableFile, CF_TYPE_TABLES);
 	}
+
+	unsigned int ModelAnimationParseHelper::getUniqueAnimationID(const SCP_string& animName, char uniquePrefix, const SCP_string& parentName) {
+		return hash_fnv1a(animName + uniquePrefix) ^ hash_fnv1a(parentName);
+	}
 	
-	void ModelAnimationParseHelper::parseAnimsetInfo(ModelAnimationSet& set, ship_info* sip) {
-		if(sip != nullptr)
-			set.changeShipName(sip->name);
-		
+	void ModelAnimationParseHelper::parseAnimsetInfo(ModelAnimationSet& set, char uniqueTypePrefix, const SCP_string& uniqueParentName) {
 		SCP_vector<SCP_string> requestedAnimations;
 		stuff_string_list(requestedAnimations);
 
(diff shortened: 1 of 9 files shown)


code/network/multimsgs.cpp
```diff
@@ -8011,14 +8011,16 @@ static constexpr size_t animation_forced_bit = 1 << 1;
...
 	int packet_size;
+	ushort netsig_to_send = special_mode == 0 && parent_object != nullptr ? parent_object->net_signature : (ushort)0;
 	ubyte data[MAX_PACKET_SIZE];
 
 	BUILD_HEADER(ANIMATION_TRIGGERED);
 
 	ADD_INT(animationId);
-	ADD_INT(pmi);
+	ADD_SHORT(netsig_to_send);
+	ADD_SHORT(special_mode); // Currently empty. Reserved to find special pmi's for non-object animations
```
JohnAFernandez, line 8023: With these packet changes we have to bump the multi version again.  I'll probably have more comments later.

code/network/multimsgs.cpp
```diff
@@ -8011,14 +8011,16 @@ static constexpr size_t animation_forced_bit = 1 << 1;
 static constexpr size_t animation_instant_bit = 1 << 2;
 static constexpr size_t animation_pause_bit = 1 << 3;
 
-void send_animation_triggered_packet(int animationId, int pmi, const animation::ModelAnimationDirection& direction, bool force, bool instant, bool pause, const int* /*time*/) {
+void send_animation_triggered_packet(unsigned int animationId, object* parent_object, ushort special_mode, const animation::ModelAnimationDirection& direction, bool force, bool instant, bool pause, const int* /*time*/) {
 	int packet_size;
+	ushort netsig_to_send = special_mode == 0 && parent_object != nullptr ? parent_object->net_signature : (ushort)0;
```
notimaginative, line 8016: I get the idea here, but I worry about it being too confusing or difficult to maintain in the future. It would be cleaner to just initialize it to zero and add an `if()` statement to conditionally assign the net_signature. That would make more sense to others looking at the code a few years down the line I think.

Also, if a non-null `parent_object` together with a non-zero `special_mode` shouldn't happen then it should either warn or error so we'll know to do something about it. And the `if()` statement would provide a good place for that.

code/network/multimsgs.cpp
notimaginative, line 8023: These macros should be changed to the unsigned versions to match the variable type (i.e., `ADD_UINT()`, `ADD_USHORT()`)

code/network/multimsgs.cpp
```diff
@@ -8039,14 +8041,16 @@ void send_animation_triggered_packet(int animationId, int pmi, const animation::
 
 void process_animation_triggered_packet(ubyte* data, header* hinfo) {
 	int offset; // linked;	
-	int animationId, pmi;
+	int animationId;
```
notimaginative, line 8044: Should be made `unsigned` to match other changes.

code/network/multimsgs.cpp
```diff
@@ -8039,14 +8041,16 @@ void send_animation_triggered_packet(int animationId, int pmi, const animation::
...
+	ushort netsig, special_mode;
 	ubyte metadata;
 	int time;
 
 	// read all packet info
 	offset = HEADER_LENGTH;
 	GET_INT(animationId);
-	GET_INT(pmi);
+	GET_SHORT(netsig);
+	GET_SHORT(special_mode);
```
notimaginative, line 8053: And again with the `GET_UINT()` / `GET_USHORT()` to match the variable types.

notimaginative (changes requested):
I'll take a better look later today, but other than the minor things mentioned, this looks good to me. And as Cyborg pointed out we'll need a multi version bump for this too.

code/network/multimsgs.cpp
BMagnu: A non-zero object and a non-zero special_mode _could_ be valid, I can imagine. There's no case for it right now, but there could be special things that are technically objects, but should be synced differently in the future. Hence that possibility should be valid.
If both are zero, however, than this indeed needs an error, or probably really an Assert. It would currently just be discarded in the packet recieve method, but is technically never valid and should not occur.

code/model/modelanimation.cpp
```diff
@@ -141,11 +136,17 @@ namespace animation {
 	void ModelAnimation::start(polymodel_instance* pmi, ModelAnimationDirection direction, bool force, bool instant, bool pause, const float* multiOverrideTime) {
 		instance_data& instanceData = m_instances[pmi->id];
```
notimaginative, line 137: There should really be a nullptr check on `pmi` here.

code/network/multimsgs.cpp
```diff
@@ -8011,14 +8011,20 @@ static constexpr size_t animation_forced_bit = 1 << 1;
...
 
-void send_animation_triggered_packet(int animationId, int pmi, const animation::ModelAnimationDirection& direction, bool force, bool instant, bool pause, const int* /*time*/) {
+void send_animation_triggered_packet(unsigned int animationId, object* parent_object, ushort special_mode, const animation::ModelAnimationDirection& direction, bool force, bool instant, bool pause, const int* /*time*/) {
 	int packet_size;
+	ushort netsig_to_send = 0;
 	ubyte data[MAX_PACKET_SIZE];
 
+	Assertion(special_mode != 0 || parent_object != nullptr, "Tried to synchronize an animation with neither an attached object, nor a valid special object mode.");
+	if (special_mode == 0)
+		netsig_to_send = parent_object->net_signature;
```
notimaginative, line 8021: Don't forget the nullptr check on `parent_object` here since release builds would just crash without warning. It would be best to just `return` if neither option is valid and not send the packet.

code/network/multimsgs.cpp
```diff
@@ -8061,11 +8069,21 @@ void process_animation_triggered_packet(ubyte* data, header* hinfo) {
 
 	float delay = time * 0.001f;
 
-	animation::ModelAnimation::s_animationById[animationId]->start(model_get_instance(pmi), direction, forced, instant, pause, &delay);
+	object* objp = multi_get_network_object(netsig);
+
+	auto animation = animation::ModelAnimationSet::s_animationById.find(animationId);
+	if (animation != animation::ModelAnimationSet::s_animationById.end()) {
+		if (special_mode == 0 && objp != nullptr) {
+			animation->second->start(model_get_instance(object_get_model_instance(objp)), direction, forced, instant, pause, &delay);
```
notimaginative, line 8077: `object_get_model_instance()` needs to be separated here for error checking. A return value `-1` is perfectly valid if a ship is joining or respawning since it may not be set up yet, but that triggers an assert in `model_get_instance()`, or it returns null in release builds which crashes in `start()`.

What you could probably do is add a nullptr check to the top of `object_get_model_instance()` returning -1 (it could use it anyway) and use that in place of the `objp != nullptr` test, checking for the `-1` return value instead.

notimaginative (changes requested):

notimaginative:
Someone may need to double check me on this, but I was testing with some bombers in multi and their turrets never moved on clients. Worked fine on the server. The animation code did appear to be processing something on the clients however.

The turrets didn't look right though, so it might be a general submodel issue rather than anything animation related. Possibly a bug in a some init code or object update.

Goober5000 (approved):
LGTM, though I defer to @notimaginative and @JohnAFernandez on the multi parts.

JohnAFernandez:
Iirc, the turrets would actually fall under object update, which is really choppy because it only transmits the angles of the submodel periodically. Dumb rotate does as well, I believe. It's better than not moving, but a pretty poor system.  I was hoping that this animation rework would allow a better multi solution for these types.

notimaginative:
> Iirc, the turrets would actually fall under object update, which is really choppy because it only transmits the angles of the submodel periodically. Dumb rotate does as well, I believe. It's better than not moving, but a pretty poor system. I was hoping that this animation rework would allow a better multi solution for these types.

That was my first thought, but the turrets don't move at all over the course of the mission. And the gun barrels on the turret (Medusa bomber) aren't visible either.

After some more testing though I confirmed that the animation code *is* working on multi (Medusa wings move on AB), just not for turrets. If I disable the position updates in `process_turret_fired_packet()` then the gun barrels show up again, facing backwards, but no part of the turret ever moves. So maybe there is some conflict with the new animation code and the old multi anim hacks that's breaking things.

BMagnu:
So, the animation code should technically not interfere with turrets.
There are only two things, from the animation code, that have an effect there:
There is an active, running animation that's moving a turret. This will override the turret data. Though things like initial state animations will not run in a way that does this.
Second, an initial turret animation will tell a turret what it's idle position should be.

So, unless you have tabled a non-initial type animation for the turret, it's probably not coming from the animation system. It might be possible that if the initial types are handled wrongly, that the gun barrel points in the wrong direction at start, but that would not prevent it from moving. There have been changes to submodel updates in #3713 that might have changed things in a way that conflicts with multi, but that's outside my area of expertiese. The animation code only calculates what the orientation of a submodel should be (canonical_orient), but leaves the actual movement and application to the submodel code.

notimaginative:
Thanks for the info, that's good to know. That helps confirm what I was wondering since disabling the animation packet didn't appear to change anything with the turret. There must be another issue somewhere so I'll see if I can track that down.

notimaginative (approved):

notimaginative (changes requested):
Oops, one thing that was missed. `object_get_model_instance()` needs to have the `nullptr` added so that it returns -1 if `objp` is null.

BMagnu:
> Oops, one thing that was missed. `object_get_model_instance()` needs to have the `nullptr` added so that it returns -1 if `objp` is null.

Mhm, that can't happen here.
since we return if objp is null and special mode is 0, a special mode of zero a few lines later will guarantee a non-null objp.
But I can add it for good measure, since it probably should have it nonetheless

notimaginative (approved):

PR #3847, opened 2021-12-23
Make secondaries use $energy consumed

Baezon:
It was completely ignored before. Just to be sure, I checked retail and no secondary had any energy consumption listed, nor did any of the high-profile mods I checked, which makes sense, since usually people will use retail weapons as a base.

Additionally, the weapon energy bar will now only flash if the weapon in question is actually starved of energy. Being out of missiles, or ballistics with no ammo will not flash the energy bar (ballistics used to do so even if they were merely out of ammo).

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -431,6 +431,7 @@ typedef struct weapon_flash
 {
 	int flash_duration[MAX_WEAPON_FLASH_LINES];
 	int flash_next[MAX_WEAPON_FLASH_LINES];
+	bool flash_energy[MAX_WEAPON_FLASH_LINES];
 	int is_bright;
 } weapon_flash;
 weapon_flash Weapon_flash_info;
@@ -1001,6 +1002,7 @@ void hud_weapons_init()
 	for ( int i = 0; i < MAX_WEAPON_FLASH_LINES; i++ ) {
 		Weapon_flash_info.flash_duration[i] = 1;
 		Weapon_flash_info.flash_next[i] = 1;
+		Weapon_flash_info.flash_energy[i] = false;
 	}
 
 	// The E: There used to be a number of checks here. They are no longer needed, as the new HUD code handles it on its own.
@@ -4521,7 +4523,7 @@ void HudGaugeTargetTriangle::render(float  /*frametime*/)
 }
 
 // start the weapon line (on the HUD) flashing
-void hud_start_flash_weapon(int index)
+void hud_start_flash_weapon(int index, bool flash_energy)
 {
 	if ( index >= MAX_WEAPON_FLASH_LINES ) {
 		Int3();	// Get Alan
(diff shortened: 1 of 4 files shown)


code/hud/hudtarget.cpp
```diff
@@ -5776,9 +5779,10 @@ void HudGaugeWeaponEnergy::render(float  /*frametime*/)
 			}
 		}
 
-		for ( i = 0; i < sw->num_primary_banks; i++ )
+		// maybe flash the energy bar
+		for ( i = 0; i < sw->num_primary_banks + sw->num_secondary_banks; i++ )
```
Goober5000, line 5783: Is this addition of the extra bank capacity going to mess with the indexing at all?

Also keep in mind that the bank offsets may be different.  If a ship has 1 primary and 2 secondary weapons, it will have the same total number as 2 primary and 1 secondary, but the indexes will be different.

code/ship/ship.cpp
```diff
@@ -11585,13 +11585,13 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
 					points = num_slots;
 				}
 
-				if ( shipp->weapon_energy < points*winfo_p->energy_consumed*flFrametime || 
-					(winfo_p->wi_flags[Weapon::Info_Flags::Ballistic] && shipp->weapons.primary_bank_ammo[bank_to_fire] <= 0))
+				bool no_energy = shipp->weapon_energy < points* winfo_p->energy_consumed* flFrametime;
```
Goober5000, line 11588: The spacing was changed inconsistently here.

code/ship/ship.cpp
```diff
@@ -11669,13 +11669,13 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
 				// the weapon's energy_consumed to 0 and it'll work just fine. - Goober5000
 
 				// fail unless we're forcing (energy based primaries)
-				if ( (shipp->weapon_energy < points*numtimes * winfo_p->energy_consumed)			//was num_slots
-				 && !force ) {
+				bool no_energy = shipp->weapon_energy < points* numtimes* winfo_p->energy_consumed; //was num_slots
```
Goober5000, line 11672: Here too.

code/ship/ship.cpp
Baezon: I'm not sure I follow, what spacing are you referring to exactly?

code/hud/hudtarget.cpp
Baezon: Yes, this is intentional, these *aren't* being used to index into weapons anymore, these are essentially changed into indices for the weapon lines in the weapon gauge itself. Whether it has 1 primary/2 secondary or 2 primary/1 secondary, index one is always going to the be the second line displayed on the hud and is therefore index 1, regardless. It only checked up to the num primary banks because it used to only flash the energy bar if any primary weapon line was flashing. Now it checks all of the currently used weapons, and only cares if that one in particular asked for a flashing energy bar, regardless whether it's a primary or secondary.

Goober5000: Gotcha.

code/ship/ship.cpp
Goober5000: The spacing around the asterisk.  Spaces were added after them, but not before.  I'd be fine with either a space on each side or no space on either side, but not this one-sided spacing.

Baezon: Oh, I see. It must have been automatically when I pasted it back in...

Goober5000 (approved):
LGTM

PR #3850, opened 2021-12-23
Fix weapons with submodel rotation firing

### LINKED ISSUE #3848: Weapon Gattler from Blue Planet stopped firing on build 23 October and later
As title says Gattler doesn't work on 23 October nightly. Only thing that could affect is #3720, which is weird, because weapon doesn't use "stream" flag. Fact is on build 1022 - weapon works, on 1023 - not.

Baezon:
In #3720 there was some code which fired stream weapons for all ships, and then set `has_fired` to 1 (regardless if it even had any stream weapons), thus preventing the code removed in this PR from ever triggering and stopping primary fire. After it was removed  this would trigger every time, thus stopping weapons with submodel rotation rate requirements from ever reaching them. The code can be safely removed as both players and AI handle stopping their own primary fire on their own.

Fixes #3848

### DIFF
--- a/code/object/object.cpp
+++ b/code/object/object.cpp
@@ -829,8 +829,6 @@ void obj_move_call_physics(object *objp, float frametime)
 {
 	TRACE_SCOPE(tracing::Physics);
 
-	int has_fired = -1;	//stop fireing stuff-Bobboau
-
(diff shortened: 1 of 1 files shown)


JohnAFernandez (approved):

Goober5000 (approved):
I've been skeptical of this firing code for years but never really sat down to look at it.  Good research.

PR #3852, opened 2021-12-26
SEXP containers, Part 6: FRED/qtFRED SEXP trees

jg18:
Add support for containers and container modifiers to FRED's and qtFRED's `sexp_tree` classes, including support for the "Replace Container Name"/"Replace Container Data" features.


code/parse/sexp_container.cpp
```diff
@@ -109,14 +124,18 @@ list_modifier::Modifier list_modifier::get_modifier(const char *modifier_name)
 	return ListModifier::INVALID;;
 }
 
-bool list_modifier::match_name(const char *other_name) const
+const char *get_list_modifier_name(const ListModifier modifier)
```
Goober5000, line 127: Should this argument be a reference?

fred2/addmodifycontainerdlg.cpp
```diff
@@ -800,24 +806,19 @@ void CAddModifyContainerDlg::OnBnClickedDeleteContainer()
...
 
-	const int times_used = m_p_sexp_tree->get_container_usage_count(name_to_check.c_str());
+	const int times_used =
+		is_null_container_name(orig_name) ? 0 : m_sexp_tree.get_container_usage_count(orig_name);
 	Assert(times_used >= 0);
 
 	if (times_used > 0) {
 		SCP_string container_name_str = "'" + curr_name + "'";
-		if (!orig_name.empty()) {
+		if (curr_name != orig_name) {
```
Goober5000, line 821: Should this be a case-insensitive check?

Goober5000 (commented):
Partial review for now

jg18:
The only missing part is fixing the known issue with the Add/Modify Container FRED dialog, where pressing Enter/Esc causes changes to be lost without prompting the user.

code/parse/sexp.cpp
```diff
@@ -1776,6 +1776,120 @@ int count_items_with_name(const char *name, const T* item_array, int num_items)
...
+		case OP_CONTAINER_REMOVE_FROM_MAP:
+			Assert(any(con_type & ContainerType::MAP));
+			if (!check_map_container_key_sexp_arg_type(con_type, is_string, is_number)) {
+				return SEXP_CHECK_WRONG_MAP_KEY_TYPE;
+			}
+			break;
+
+		case OP_CONTAINER_ADD_TO_MAP:
+			Assert(any(con_type & ContainerType::MAP));
+			if (argnum % 2 != 0) {
```
Goober5000, line 1829: Shouldn't this be == 0?  Don't keys come before data?  Arg0 - key, Arg1 - data, etc.

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -4900,6 +5277,15 @@ std::unique_ptr<QMenu> sexp_tree::buildContextMenu(QTreeWidgetItem* h) {
 
 	auto replace_variable_menu = popup_menu->addMenu(tr("Replace Variable"));
 
+	popup_menu->addSeparator();
+
+	popup_menu->addSection("Containers");
+
+	auto add_modoify_container_act = popup_menu->addAction(tr("Add/Modify Container"), this, []() {});
```
Goober5000, line 5284: modify is misspelled "modoify"

code/parse/sexp_container.cpp
jg18: `ListModifier` is an alias for the `Modifier` enum class inside `list_modifier` (see sexp_container.h), so passing it by value makes sense to me. What do you think?

fred2/addmodifycontainerdlg.cpp
jg18: Hmm, good question. Since this is purely for a user-facing message, I'd say it depends on what makes for the best UX.

If the FREDder renamed a container from "foobar" to "FooBar" and then attempted to delete "FooBar" before closing the dialog, even though the container is currently in use, I can't decide whether it would be helpful to remind the FREDder that the name was originally "foobar". I don't know whether it would make the fact that container names are case-insensitive clearer or more confusing.

Thoughts?

code/parse/sexp.cpp
jg18: Since `add--to-map`'s first argument is the name of a map container, the SEXP's map keys start at `argnum` 1 and values start at `argnum` 2.

qtfred/src/ui/widgets/sexp_tree.cpp
jg18: Thanks, good catch, will fix.

fred2/addmodifycontainerdlg.cpp
Goober5000: Oh, I see what you mean.  Hmm.

Ok, in this case I think a case-sensitive check is appropriate.  It's a change that the user made, even if it is treated as equivalent.

code/parse/sexp_container.cpp
Goober5000: I missed the fact that this was an enum.  Yes, pass-by-value is fine here.

jg18:
I started adding assertion failure messages, but it'll take a little time to finish, since there are over 100 of them.

jg18:
I finished updating the assertions.

There are a few assertions left of the form
```
Assert(op >= 0);
```
but they predate this PR, so modifying them seems out of scope.

fred2/sexp_tree.cpp
```diff
@@ -7361,9 +7369,15 @@ int sexp_tree::get_container_usage_count(const SCP_string &container_name) const
...
 bool sexp_tree::rename_container_nodes(const SCP_string &old_name, const SCP_string &new_name)
 {
-	Assert(!old_name.empty());
-	Assert(!new_name.empty());
-	Assert(new_name.length() <= sexp_container::NAME_MAX_LENGTH);
+	Assertion(!old_name.empty(),
+		"Attempt to rename container nodes looking for empty name. Please report!");
+	Assertion(!new_name.empty(),
+		"Attempt to rename container nodes with empty name. Please report!");
+	Assertion(new_name.length() <= sexp_container::NAME_MAX_LENGTH,
```
Goober5000, line 7376: Just to check... is `NAME_MAX_LENGTH` a length or a buffer size?

jg18: It's a length (see sexp_container.h).

Goober5000 (approved):

PR #3857, opened 2021-12-27
PPC tonemapper values followup

EatThePath:
Moved the PPC tonemapper intermediate calculations to C-side to clean up shader compatibility with older versions of opengl. Followup to #3838

### DIFF
--- a/code/graphics/util/uniform_structs.h
+++ b/code/graphics/util/uniform_structs.h
@@ -228,12 +228,17 @@ struct passthrough_data {
 struct tonemapping_data {
 	float exposure;
 	int tonemapper;
-	float toeS;
-	float toeL;
-	
-	float shoulderS;
-	float shoulderL;
-	float shoulderA;
+	float x0; //from here on these are for the PPC tonemappers
+	float y0;
+
+	float x1;
+	float toe_B;
+	float toe_lnA;
+	float sh_B;
+
+	float sh_lnA;
+	float sh_offsetX;
+	float sh_offsetY;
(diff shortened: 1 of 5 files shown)


wookieejedi:
Thanks for the quick turnaround on following @asarium's request.

Goober5000:
I took a look at this but I don't think I'm really qualified to review it.

code/graphics/util/uniform_structs.h
```diff
@@ -228,12 +228,17 @@ struct passthrough_data {
...
+
+	float x1;
+	float toe_B;
+	float toe_lnA;
+	float sh_B;
+
+	float sh_lnA;
+	float sh_offsetX;
+	float sh_offsetY;
 	float pad[1];
```
asarium, line 242: Interesting side note, you don't actually need the trailing padding bytes. I got that wrong initially which is why most of these structs have them but the uniform aligner takes care of aligning the individual instances properly. Not something to change here since it was already there.

asarium (approved):

code/graphics/util/uniform_structs.h
EatThePath: Good to know. Would it be welcome if sometime I submit a PR to just remove all the padding that's lingering in the various uniforms?

asarium: Sure, feel free to do that. Just to be clear, this only applies to the padding at the end of structs. The padding within is still needed to comply with std140.

Issue #3147, opened 2021-01-15
Mysterious missing subsystems on the Repulse in ST:R's Forced Hand


Goober5000:
Several people have reported that the Repulse has no subsystems at the beginning of the mission.  Nothing can be targeted, and firing Stilettos at the location of the engine system has no effect.  As a result it is impossible to disable the Repulse.

I have seen this in several videos, but it appears to happen randomly and cannot be reliably reproduced.  Exiting the game and restarting the mission seems to fix it.

The issue can be seen in MatthTheGeek's stream at 23:00...
https://www.twitch.tv/videos/874683238

JohnAFernandez:
One of the questions we need to ask on this one is, has  this happened on any other models. Or even any other missions in ST:R which have had this issue with the orion model.

Goober5000:
As far as I know it only happens on that one mission, and didn't happen earlier than a year or two ago.  If I had to speculate I would say that a somewhat recent FSO change caused the engine to simply run out of subsystem slots after playing a certain number of missions.  This is why exiting and restarting fixes it.

Of course I have no idea if that is really what is happening, or even if it is possible to run out of subsystem slots in the first place.

tomimaki:
It happened even 5 years ago: https://www.hard-light.net/forums/index.php?topic=89403.0

Goober5000:
Possibly fixed in #3858 - we'll need to see if it shows up after this.

notimaginative:
I think this issue might predate when the leak bug was introduced (~3 years ago). But I did give it a test with 50-60 attempts at the mission in various scenarios (after other missions, repeated mission restarts, etc.) and never was able to reproduce this. The subsystem leak is exceptionally small compared to the two other fixes though (colliders and awacs) so maybe a combination of things were/are at work.

Goober5000:
Ok.  I'll assume it's fixed then, and if it shows up later, we can always reopen this.

Goober5000:
The bug is back, as Belisarius's screenshot proves:
<img width="554" height="262" alt="Image" src="https://github.com/user-attachments/assets/cfeb592a-b7e1-441a-8c50-e0dfb483e503" />

But this time, Claude is on the case.

PR #3858, opened 2021-12-27
fix memory leak when reallocating ship subsystems

notimaginative:


JohnAFernandez (approved):
LGTM

Goober5000 (approved):
LGTM as well

JohnAFernandez:
I'm willing to bet this memory issue was a possible cause for the subsystems not appearing in that ST:R mission.

Goober5000:
That's an interesting theory, and quite plausible.

I'll tag #3147 so that the issue is linked.

PR #3861, opened 2021-12-28
SEXP containers, Part 7: Status SEXPs

jg18:
Add the following SEXPs:
- `is-container-empty`
- `get-container-size`
- `list-has-data`
- `list-data-index`
- `map-has-key`
- `map-has-data-item`

## SEXP evaluation cheat sheet

Here's the background needed to understand the PR.

### Notes on the changes in sexp.cpp

The function `query_operator_argument_type()` specifieds the expected types for each SEXP's arguments, using the zero-indexed parameter `argnum`.

The SEXP argument type `OPF_CONTAINER_VALUE` represents either container data or a map container key. We need this special type because data/keys can be either strings *or* numbers, depending on the specific container.

### Notes on the changes in sexp_container.cpp:

The parameter `node` is an index into the `Sexp_nodes` vector and represents the currently evaluated argument for the SEXP being evaluated.

The statement `node = CDR(node);` means "get the next argument for this SEXP".

The expression `node == -1` means that there are no additional arguments for this SEXP and unsurprisingly, `node != -1` means the opposite.

The function call `CTEXT(node)` means "get the textual representation of the current SEXP argument". The reason we can't simply use `Sexp_nodes[node].text` is that the node may use Replace Variable or Replace Container Data, in which case we need to perform the replacement to get the correct text.


code/parse/sexp_container.cpp
```diff
@@ -787,3 +787,261 @@ const char *sexp_container_CTEXT(int node)
...
+
+// status SEXPs
+int sexp_is_container_empty(int node)
+{
+	const char *container_name = CTEXT(node);
+	const auto *p_container = get_sexp_container(container_name);
+
+	if (!p_container) {
+		Warning(LOCATION, "Is-container-empty called on nonexistent container %s.", container_name);
+		return SEXP_FALSE;
```
JohnAFernandez, line 799: I kept going back and forth about whether this invalid data should return true or false here.  My use case for this sexp would be to check if a container which contains a list or queue of items that I need to do *something* to has contents before I try to do things with the items in it.  So I would probably constantly use `!(sexp_is_container_empty)`.  But others could use it to try to tell if something needs to be added to that container, because it's some container that the mission needs to be doing something with constantly.

There doesn't seem to be a good way to "foolproof" this for users during mission testing. I can't think of any way to gracefully handle it not existing since we can only return true or false, and either can be the expected or desired result of a container not existing.

I think there are two ways to improve this, if desired:

A) Change to an error, or
B) Add that ignoring the warning will lead to undesired behavior.

In the future, maybe it would be good to add a SEXP_ERROR, that prevents SEXPs from evaluating in situations like this and then logs the error in the debug log.

code/parse/sexp_container.cpp
```diff
@@ -787,3 +787,261 @@ const char *sexp_container_CTEXT(int node)
...
+}
+
+int sexp_get_container_size(int node)
+{
+	const char *container_name = CTEXT(node);
+	const auto *p_container = get_sexp_container(container_name);
+
+	if (!p_container) {
+		Warning(LOCATION, "Get-container-size called on nonexistent container %s.", container_name);
+		return 0;
```
JohnAFernandez, line 829: In contrast, this can definitely stay as-is, since 0 makes sense in any situation.  The mission designer is going to have to fix it any way.

So far the other instances can stay as-is, too.

code/parse/sexp_container.cpp
```diff
@@ -787,3 +787,261 @@ const char *sexp_container_CTEXT(int node)
...
+		case OP_MAP_HAS_KEY:
+			return sexp_map_has_key(node);
+			break;
+
+		case OP_MAP_HAS_DATA_ITEM:
+			return sexp_map_has_data_item(node);
+			break;
+
+		default:
+			UNREACHABLE("Unknown container status SEXP operator %d", op_num);
```
JohnAFernandez, line 1044: Just needs a please report, imho.

JohnAFernandez (approved):
Nothing blocking.  The biggest thing I can see is in `is-container-empty`.

code/parse/sexp_container.cpp
```diff
@@ -787,3 +787,261 @@ const char *sexp_container_CTEXT(int node)
...
+	if (container.is_map()) {
+		if (container.map_data.empty()) {
+			return SEXP_TRUE;
+		}
+	} else if (container.is_list()) {
+		if (container.list_data.empty()) {
+			return SEXP_TRUE;
+		}
+	} else {
+		Error(LOCATION,
```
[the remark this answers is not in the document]
jg18: Agreed, `UNREACHABLE()` makes more sense here.

If we get to this point, then either there's a bug in the existing code, or we added a new container type and forgot to update this SEXP, also a bug. Will fix.

code/parse/sexp_container.cpp
jg18: Good thoughts!

Arguments that should be a variable name (`OPF_VARIABLE_NAME`) are handled strictly. The value *must* be a name of a known variable: `check_sexp_syntax()` rejects anything else.

However, @Karajorma designed container SEXPs to be much more flexible. An argument that should be a container name (`OPF*CONTAINER_NAME`) *doesn't* have to be a string literal. It can be a special argument or a usage of Replace Variable or even Replace Container Data.

That's why we have to do this runtime check on container name arguments in the first place.

Although you could make a case for `Error()` in this situation, Kara chose to use `Warning()` when he wrote the original code.

I'll double-check if that was intended behavior, and if so, what the reasoning was.

jg18: As for return value here, my vote is still `SEXP_FALSE`. A nonexistent container can't be empty.

We could argue it's vacuously true and thus return `SEXP_TRUE`, but that's definitely not intuitive behavior for non-mathy people,. 😜

code/parse/sexp_container.cpp
jg18: Good catch, I updated the `Assert()` messages but forgot to update the `UNREACHABLE()` ones.

code/parse/sexp_container.cpp
JohnAFernandez: Yup, and the hope is that the mission tester will heed the warning and fix the mission. So it should be fine either way.

Goober5000 (commented):
Looks good.  No additional comments.

code/parse/sexp_container.cpp
jg18: Per discussion with Kara, we've decided to leave the messages as warnings but print them to the mission log as well. I've pushed a commit.

Issue #3184, opened 2021-01-28
FR: FRED Events Editor Feature List


MjnMixael:
A non-exhaustive list of things FREDers want for the Events editor... arguably where we spend 98% of our time when making missions.

- [ ] Inline comments (read: not necessarily mouse over)
- [x] Resizable window (read: BIG for both events, and messages). Ideally the events tree would wrap to multiple columns if the window is large enough.
- [x] Keyboard shortcuts. All of them. Functioning. Ctrl-A, Ctrl-C, Ctrl-V, Ctrl-Z and more.
- [x] Colors to arbitrarily highlight events in arbitrary colors.
- [ ] Some ability to visually organize events. Whether that's indents, arbitrary spaces, event folders.. whatever. Events come in groups and often those groups are kind of like their own code function. OK, that feature works. Let's move it out of the way and do something else.
- [x] The Events window should verify that you actually intended to close the Events window when you hit cancel if you have unsaved events.

the-maddin:
It would be nice to have a events filtering feature: a text field called "Filter events", and when the user enters "Alpha 1", all events containig sexps that have "Alpha 1" as an argument (or contain at any position the entered string) are highlighted, maybe by setting a (customizable?) text color.
Most generally useful would be, in my opinion, a full-text search of events, so the filter string "cargo" would match events using the sexp (is-cargo  ...) as well as those with a vessel "Cargo Container 77" or a variable "cargoes-destroyed-num".

Goober5000:
Converted to checkbox form and checked off three items (resizable FRED, keyboard shortcuts, and colors) implemented in #3191.

Based on my investigation, inline comments are not possible in the MFC tree control.  Tooltip comments or @naomimyselfandi's script nodes are the best that can be done.

I haven't investigated, but my hunch is that visual organization of events are not possible either.  I ruled out indentation during the course of #3191, and I don't think arbitrary spacing is feasible either.  I also don't think event folders are feasible because they need the same kind of specialized UI as inline comments.

The only remaining issue is event window confirmation which ought to be pretty easy to do, so I will mark it as Good First Issue.

Goober5000:
Checked off the last item since it was done in #3624.

Goober5000:
Removed "good first issue" since the low-hanging fruit is all done.  I don't think the remaining requests are tractable (see my comments on Feb 10) so this should be considered for closure.

PR #3864, opened 2021-12-28
Add comment on why event_editor::OnCancel() is blank

jg18:
~~This PR also establishes the semantics that after the Events Editor is closed for any reason, `Event_editor_dlg` is always deleted.~~

~~Note that the behavior being added to `OnCancel()` already exists in `OnClose()`.~~

Clarify why the function `event_editor::OnCancel()` is empty.

### DIFF
--- a/fred2/eventeditor.cpp
+++ b/fred2/eventeditor.cpp
@@ -855,7 +855,9 @@ void event_editor::OnDelete()
 // this is called when you hit the escape key..
 void event_editor::OnCancel()
 {
-	OnButtonCancel();
+	// override MFC default behavior and do nothing
+	// the Esc key is used for certain actions inside the events editor
+	// so presseing Esc shouldn't close the window
 }
 
 // this is called when you click the ID_CANCEL button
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
The current behavior (neither Enter nor Escape closing the dialog) is by design.  Neither key should affect the overall Event Editor dialog, because both keys affect things that are done within the dialog.  See further information on PR #3214.

This PR should be closed.

jg18:
@Goober5000  Thanks for explaining. A comment in `OnCancel()` explaining why the function is empty would be cool, but not worth a PR just for that.

Is Issue #3184's request (quoted below) already done, then? Or did I misunderstand the request?
> The Events window should verify that you actually intended to close the Events window when you hit cancel if you have unsaved events.

I'll close this as soon as that's clarified. Thanks.

Goober5000:
Agreed, a comment in `OnCancel` is a good idea.

Yes, the #3184 request was addressed by #3624.  Both the close button and the cancel button will ask for confirmation.

jg18:
Updated the PR to instead add the comment to `OnCancel()`.

Goober5000 (approved):
That works.  Although "presseing" should be "pressing".

PR #3865, opened 2021-12-29
Fix scripted setting of weapon targets

Baezon:
A weapon fired without a lock has a proper `target_sig` but no `homing_object` and will not home, so when setting the target of a weapon through scripting check for that case as well, to prevent edge cases where a weapon might not allow one to set the target manually through scripting.

All the rest is adding a clarity function for how 'no homing object' is represented, which has always confused me because it looks weird. Some odd pointer shenanigans, but it works.


Goober5000 (approved):
LGTM

JohnAFernandez (approved):

PR #3868, opened 2021-12-29
set up awacs team arrays just the once

notimaginative:
Awacs team arrays only need to be initialized once but were expanded by Iff_info.size() every second of a mission and never cleared.

(Also includes cleanup)


JohnAFernandez (approved):

Goober5000 (approved):
YIkes.  Approved.

PR #3869, opened 2021-12-29
take pathtype extensions into account during file indexing

### LINKED ISSUE #3863: CFile cannot handle more than 65536 files.
Revision 251702511 from 2021-11-22 introduced a bug, that makes FSO unusable if you play with MVPs and mjnhalls.
Debug Build crashes right away.

Release Build is launchable, but the game crashes after i go into the campaign room for example.
2021-11-21 still works, but every build after that causes the error message and crash.

[fs2_open.log](https://github.com/scp-fs2open/fs2open.github.com/files/7785892/fs2_open.log)

I hit the limit because of my 17k cache files. So after i deleted them it worked again. But earlier builds worked even with all the cache files in place. So there seems to be a limit now, that was not present before.

notimaginative:
The cfilesystem cleanup PR (3734) failed to take pathtype extension filtering into account when indexing loose files. This adds back that functionality to prevent extra/unknown files from being indexed.

Fixes #3863

### DIFF
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -691,6 +691,12 @@ void cf_search_root_path(int root_index)
 		cf_get_list_of_files(search_path, files, "*.*");
 
 		for (auto &file : files) {
+			auto ext_idx = file.name.rfind('.');
+
+			if ( ext_idx == SCP_string::npos || !is_ext_in_list(Pathtypes[i].extensions, file.name.substr(ext_idx+1).c_str()) ) {
+				continue;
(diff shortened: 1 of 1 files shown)


JohnAFernandez:
I want to review it, but I have no idea what I'm looking at.

notimaginative:
It's just checking for an extension on the filename and makes sure that extension is a valid for the cfile pathtype. If there is no extension, or the extension isn't in the allowed list, then it's skipped.

Goober5000 (approved):
LGTM

PR #3870, opened 2021-12-30
Exclude certain recent error types from the fatal exit on mission parse.

EatThePath:
Following the changes in #3774 many old missions which have been perfectly functional as far as anyone could tell now error on load. This makes an exception to the parse error logic to allow those missions to continue loading, while still firing the new warnings.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -6150,7 +6150,8 @@ bool post_process_mission()
 				Warning(LOCATION, "%s", error_msg.c_str());
 
 				// syntax errors are recoverable in Fred but not FS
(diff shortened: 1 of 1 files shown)


Goober5000 (changes requested):
1) make a new function in sexp.cpp, either before or after `sexp_error_message` called `sexp_non_fatal_error` which returns a boolean
2) use a case statement which lists the non-fatal errors
3) reference the function in this if() condition
4) add `SEXP_CHECK_INVALID_CUSTOM_HUD_GAUGE` and remove the hack around line 3043 of sexp.cpp

Goober5000:
Closing in favor of #3873

PR #3872, opened 2021-12-30
make cfile root and file blocks dynamic

notimaginative:
Remove limit on number of roots and files supported by cfilesystem.

### DIFF
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -72,46 +72,49 @@ typedef struct cf_root_sort {
 	int				cf_type;
 } cf_root_sort;
 
-#define CF_NUM_ROOTS_PER_BLOCK   32
-#define CF_MAX_ROOT_BLOCKS			256				// Can store 32*256 = 8192 Roots
-#define CF_MAX_ROOTS					(CF_NUM_ROOTS_PER_BLOCK * CF_MAX_ROOT_BLOCKS)
+#define CF_NUM_ROOTS_PER_BLOCK		32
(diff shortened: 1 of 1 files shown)


notimaginative:
Works on my limited data set but needs further testing with very large numbers of roots/files.

code/cfile/cfilesystem.cpp
```diff
@@ -72,62 +72,65 @@ typedef struct cf_root_sort {
...
+#define CF_NUM_FILES_PER_BLOCK		512
+#define CF_MIN_FILE_BLOCKS			16				// minimum number of file blocks to create on resize
 
 typedef struct cf_file_block {
 	cf_file						files[CF_NUM_FILES_PER_BLOCK];
 } cf_file_block;
 
 static uint Num_files = 0;
-static cf_file_block  *File_blocks[CF_MAX_FILE_BLOCKS];
+static SCP_vector<cf_file_block *> File_blocks;
```
asarium, line 109: You could use `unique_ptr`s to avoid the manual memory management here.

asarium (approved):

code/cfile/cfilesystem.cpp
notimaginative: Good point. I don't really use them enough for that to be my goto for stuff like this.

EatThePath:
> Works on my limited data set but needs further testing with very large numbers of roots/files.

@notimaginative how large is usefully large for testing?

notimaginative:
Preferably above 70k, which puts it over the old limit. But anything over 50k would probably be a good enough high mark to verify everything works.

notimaginative:
Tested with over 100k files without issue so this should be as ready as it's going to get.

PR #3873, opened 2021-12-30
make certain SEXP errors nonfatal

Goober5000:
Some SEXP errors, although they indicate real problems, will not prevent the mission from running.  Add these as exceptions.  Implemented per requests from users of Blue Planet and Solaris.

Supersedes #3870.


wookieejedi (approved):
LGTM

PR #3877, opened 2022-01-01
Animation Rework Phase 3, Turret fix

### LINKED ISSUE #3876: Multi-Part Turrets and Wrong Initial Positions
I've been working with TBP and a few of the ships have been displaying some incorrect default positions under current builds.  The odd thing is that in FRED, the lab and the techroom, they display properly.  The ships I've encountered this are Nova#guns and Nova#shadow.  These ships each feature 6 twin-barrel multi-part turrets on the top, 6 on the bottom, and 3 down each side.  The proper position for the dorsal and ventral turrets should be the forward 4 pairs facing level and forward, with the aft 2 pairs facing aft and level.  Unfortunately in-game, one whole side is facing aft, as is the opposite side on the ventral.

BMagnu:
Turrets will now correctly calculate their default position based on POF'd rotation axis. Previously, this could cause problems.
This now also properly supports the case if a turret has been rotated off-axis by an animation, and will set the turret angle for AI and stuff to the closest possible angle on the rotation axis.

Should fix #3876

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -1603,7 +1603,8 @@ float vm_closest_angle_to_matrix(const matrix* mat, const vec3d* rot_axis, float
 	const auto& a = rot_axis->a1d;
 
 	const float w = m[0]+m[4]+m[8];
-	const float y = -2 * ( m[0]*(a[1]*a[1]+a[2]*a[2]) + m[4]*(a[0]*a[0]+a[2]*a[2]) + m[8]*(a[0]*a[0]+a[1]*a[1]) - a[0]*a[1]*(m[1]+m[3]) - a[0]*a[2]*(m[2]+m[6]) - a[1]*a[2]*(m[5]+m[7]));
+	const float y = -2 * ( m[0]*(a[1]*a[1]+a[2]*a[2]) + m[4]*(a[0]*a[0]+a[2]*a[2]) + m[8]*(a[0]*a[0]+a[1]*a[1]) -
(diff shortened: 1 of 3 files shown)


Goober5000 (commented):
Looks fine as far as it goes, but I'll defer the math to someone better suited to review that part.

code/math/vecmat.cpp
```diff
@@ -1585,6 +1586,70 @@ void vm_matrix_to_rot_axis_and_angle(const matrix *m, float *theta, vec3d *rot_a
...
+	//                          a[0]*a[1]*(m[1]+m[3]) - a[0]*a[2]*(m[2]+m[6]) - a[1]*a[2]*(m[5]+m[7])) * sin(1/2 * x)^2
+	//					   + (a[0]*(m[5]-m[7]) + a[1]*(-m[2]+m[6]) + a[2]*(m[1]-m[3])) * sin(x)
+
+	// The factor before the sine squared will be calculated as y, the factor before the sine as z, the summand as w:
+
+	const auto& m = mat->a1d;
+	const auto& a = rot_axis->a1d;
+
+	const float w = m[0]+m[4]+m[8];
+	const float y = -2 * ( m[0]*(a[1]*a[1]+a[2]*a[2]) + m[4]*(a[0]*a[0]+a[2]*a[2]) + m[8]*(a[0]*a[0]+a[1]*a[1]) - a[0]*a[1]*(m[1]+m[3]) - a[0]*a[2]*(m[2]+m[6]) - a[1]*a[2]*(m[5]+m[7]));
```
Baezon, line 1606: I would break up this line into two, to better match the way this line is displayed above.

Baezon (changes requested):
Looks sound to me, just the one minor style comment.

Baezon (approved):

Issue #3854, opened 2021-12-27
Mission time runs slightly faster than clock time


Goober5000:
This can be seen in the happynewyear mod, where various events relating to the end-of-mission song (the "happynewyear" banner, the "twenty-nine" subtitle, and the actual end of the mission) occur sooner than they should.

According to my stopwatch, one in-game second is about .97 of a real second.

~~The discrepancy happened sometime between 19-RC2 and now, because on 19-RC2 all the events occurred at the expected times.~~  Maybe not, see latest comment.

Goober5000:
@notimaginative any ideas here?

MjnMixael:
Just wanting to make sure its known that the last time someone tried to synchronize these two, it affected all existing in-game cutscenes and other timed events in a negative way. I'm not opposed to a fix or if this is exactly the same as what occurred before, but I'd recommend caution.

Goober5000:
Noted.  Do you happen to remember the PR or issue for that last synchronization?

MjnMixael:
I've been searching for 20 minutes and I can't find it. I know for sure that it was committed to a public build and that it was eventually reverted. I spent a lot of time trying to convince whoever did it that it broke timed events (typically things timed to music). I want to say it was a zookeeper commit.. but I can't find anything on github or Mantis. I'm decently sure the issue predated Discord, so whatever IRC chats we had are long gone. I _think_ this all happened before BtA released, so it's been at least five years.

Goober5000:
After further investigation it looks like the environment has something to do with this.  I originally tested this mission on 19-RC2 on Windows 7, and all the timings were correct.  I just now tested this on 19-RC2 on Windows 11, and the timings are wrong.  So the problem must lie either in the system library or the code's interaction with it.

I did verify that the audio file's timing is correct.  The lyrics that occur at 0:44 and 3:20 in the song also occur at 0:44 and 3:20 in real time.

notimaginative:
I not sure that it's a bug, just the way it works. Mission time isn't supposed to be in sync with system time as far as I know. It is based on system time but it's definitely not "real time".

This may be the commit that MjnMixael is referring to: https://github.com/scp-fs2open/fs2open.github.com/commit/e98ff515358c8fdd1a0a621dc865bd01423744ef

And the linked HLP forum post discussing the issue: https://www.hard-light.net/forums/index.php?topic=79070.0

notimaginative:
[...]

PR #3878, opened 2022-01-02
Timer preliminary cleanup

Goober5000:
I've started work on the timer reimplementation and found a few things that need cleaning up, plus one bug (incorrect initialization of an event music timestamp).  To minimize the mess I figured I'd submit them as an independent PR.

Related to #3854


wookieejedi (approved):
Seems straightforward and LGTM

PR #3879, opened 2022-01-02
convert the FSO timestamp system to use SDL time

### LINKED ISSUE #3854: Mission time runs slightly faster than clock time

Goober5000:
Previously, the game used a dead-reckoning system for measuring time, where several time counters were incremented as long as time was running.  This unfortunately has no way to synchronize time in the long term, and therefore in-game time will tend to run faster or slower than real time by a few percent, depending on various factors.

This PR redesigns the timestamp system to use the SDL time counter which will always keep it in sync with real time.  However this necessarily involves a bit of clever handling when the game is paused and unpaused.  It also revealed several minor bugs in several places in the code where timestamps were handled.

~~This PR is currently in draft status until the design can be updated to handle time compression.  In all other respects it should be working.~~  ~~The PR will be rebased after #3878 is merged.~~

Fixes #3854.

EDIT:

This should have several benefits:
1) Music and mission events will now be consistently synchronized across computers
2) Mission message timestamps will now match the mission time counter
3) Timestamps will no longer lose accuracy at high time compressions

### DIFF
--- a/code/gamesnd/eventmusic.cpp
+++ b/code/gamesnd/eventmusic.cpp
@@ -659,7 +659,7 @@ void event_music_first_pattern()
 			audiostream_stop( Patterns[Current_pattern].handle );
 	}
 
-	Pattern_timer_id = 2000;	// start music delay
+	Pattern_timer_id = timestamp(-1);	// don't let this start quite yet; see event_music_set_start_delay()
 	
 	Event_music_begun = FALSE;
 	if ( Event_Music_battle_started == TRUE ) {
@@ -668,6 +668,13 @@ void event_music_first_pattern()
 	else {
 		Current_pattern = SONG_NRML_1;
 	}
+}
+
+// Event music was originally hard-coded to start at 2 seconds, regardless of the timestamp.  Since the default player entry delay is 1 second,
+// this yields a relative start time of 1 second.  Mission start will now use relative rather than absolute delay.
+void event_music_set_start_delay()
+{
+	Pattern_timer_id = timestamp(1000);
 }
 
 // -------------------------------------------------------------------------------------------------
--- a/code/gamesnd/eventmusic.h
+++ b/code/gamesnd/eventmusic.h
@@ -107,6 +107,7 @@ void	event_music_parse_musictbl(const char *filename);
 void	event_music_change_pattern(int new_pattern);
 int	event_music_return_current_pattern();
 void	event_music_first_pattern();
+void event_music_set_start_delay();
 int	event_music_battle_start();
 int	event_music_enemy_arrival();
 int	event_music_friendly_arrival();
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -157,6 +157,21 @@ void timer_unpause_timestamp(bool sudo)
 	}
 }
 
+void timer_adjust(float delta)
+{
+	Assertion(Timer_inited, "Timer should be initialized at this point!");
+	Assertion(Timer_timestamp_offset_from_counter != 0 && Timer_timestamp_paused_at_counter != 0 && Timestamp_microseconds_at_mission_start != 0,
+		"Warranty void if these variables have not been set!");
+
+	auto microseconds = (long double) delta * MICROSECONDS_PER_SECOND;
+	auto timer_delta = (uint64_t) (microseconds / Timer_to_microseconds);
+
+	// adjust all the internal variables so it is as if the timer jumped forward
+	Timestamp_microseconds_at_mission_start -= (uint64_t) microseconds;
+	Timer_timestamp_offset_from_counter -= timer_delta;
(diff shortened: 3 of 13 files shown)


notimaginative:
I'm going to make time tomorrow to give this a good test. Nothing looks particularly out of place, but since the old code was based on frame rate I'm just a little concerned that there may be some unintended consequences in some areas.

Goober5000:
Much appreciated. :+1:

Goober5000:
Rebased so it's a bit cleaner now.  (Also updated a stub function in the test suite.)

notimaginative:
Umm, there's something funky going on here...

I'm getting a very noticeable delay entering a mission. It varies a little bit but is generally between 0.5-3 seconds of lag after hitting "Commit". Typically there is zero lag and I'm in the mission pretty much instantly. Also the High Noon mission just hangs totally. The game isn't in a loop or anything, it's still processing away like normal, but the window is frozen.

Neither of those issues appear on plain master.

Goober5000:
Ok, I'll investigate.

Goober5000:
I figured it out.  Previously, the game made the timer sprint through a quarter-second every frame before the mission started.  I added a function to simulate this with the timer.  First, I'll want to ensure that I used the correct types and type casting when I calculated the timer deltas.  Second, we should consider whether any timestamps are stored before the player enters the mission.

The `timer_adjust` function will be invisible from the perspective of anything that checks elapsed time from the beginning of the mission, whether sexps, mission time, or timestamps.  However, if any timestamps *were saved* during that short period before the player enters, they will no longer be correct.  I think this is fairly unlikely, but it should be mentioned.

Goober5000:
On second thought, the raw value of the timer isn't used anywhere in the timestamp system.  The only thing that is saved is the output of `timestamp()`, which already takes these offsets into account.  So maybe my concern is unfounded.

notimaginative:
Confirmed to fix my two earlier issues. Tried a handful of different missions, plus the worst offenders from before, and everything worked as expected.

However I noticed that mainhall animations do not play. You get the sound effect but that's it. All other animations that I looked for appeared to work as expected though, so as far as I can tell it's just the mainhall ones that are having a problem.

Further testing will have to wait until next week unfortunately. But I'll get to it as soon as I'm able.

Goober5000:
@wookieejedi Quite possible.  I'll look into it.

Goober5000:
Yep - timestamps are used in many other places - credits, main hall, certain UI animations, briefing and debriefing pauses, etc.  I modified things so that the timestamps are running most of the time, just paused in certain special cases.  I also changed the stop/start functions to use a boolean instead of an int, since it's too easy to mis-match the number of times a start or stop is called.

Since the timestamps need to be stopped during mission initialization and yet run during the briefing, I added a new function to reset the timestamps to the beginning - much the reverse of the function that skips ahead during pre-player-entry frames.  This was done in another form in the original code, so it's not too outlandish.

I tested main halls before and after a mission, as well as pausing/unpausing with both the Pause and Esc key, and navigating to the options screens and back.  Everything appears to work as expected.  That again, hopefully, leaves time compression as the only remaining issue.

Goober5000:
Time compression was definitely a puzzle, but I've managed to squash it.  This is now out of draft status.

Goober5000:
Rebased this after #3931

Goober5000:
@notimaginative How is this week looking?

notimaginative:
All of my single-player tests passed just fine. One multi issue however, which I haven't tracked down yet, and I can't say for sure that it's even related to this code.

Otherwise my only real complaint is with wording: "timer" vs "timestamp". My opinion is that the all of the timestamp related functions and variables should not use "timer" in the name unless absolutely necessary. Not doing that is just confusing since `timer_*()` could either be a timer function or a timestamp function and those a very different things. It might even be helpful to split the timestamp stuff into a separate cpp (but same header) given how much new content was added.

notimaginative:
Multi bug is reproducible on current master so it doesn't appear related to this PR.

Goober5000:
Thanks for the testing, and good point about the warning.  The timer and timestamp are related, and in a few places integrated with one another (especially with the static variables), so I think they should stay in the the same file.  But renaming some symbols is a good idea.

Goober5000:
Pushed some changes.

notimaginative:
Looks good. I tracked down the multi bug so let me deal with that real quick and then I'll do a final test to verify everything is good here with the new changes.

notimaginative (approved):
Looks good to me and all of my tests passed.

PR #3880, opened 2022-01-02
Add tonemapping options to lab render settings

EatThePath:
This punches a bunch of holes in the abstraction of light profiles that I will want to revisit when I get light profiles fully featured, but this is too useful for previewing different settings to sit on until then.

### DIFF
--- a/code/lab/dialogs/render_options.cpp
+++ b/code/lab/dialogs/render_options.cpp
@@ -1,5 +1,6 @@
 #include "lab/labv2_internal.h"
 #include "lab/dialogs/render_options.h"
+#include "lighting/lighting_profiles.h"
 
 void set_model_rotation_flag(Checkbox* caller) {
 	auto value = caller->GetChecked();
@@ -153,6 +154,12 @@ void set_bloom(Slider* caller) {
 	getLabManager()->Renderer->setBloomLevel(fl2i(value));
 }
 
+void set_exposure(Slider* caller) {
+	auto value = caller->GetSliderValue();
+
+	getLabManager()->Renderer->setExposureLevel(value);
+}
+
 void change_detail_texture(Tree* caller) {
 	auto detail = (TextureQuality)caller->GetSelectedItem()->GetData();
 
@@ -163,6 +170,12 @@ void change_aa_setting(Tree* caller) {
 	auto setting = (AntiAliasMode)caller->GetSelectedItem()->GetData();
 
 	LabRenderer::setAAMode(setting);
+}
+
+void change_tonemapper_setting(Tree* caller) {
+	auto setting = (TonemapperAlgorithm)caller->GetSelectedItem()->GetData();
+
+	LabRenderer::setTonemapper(setting);
 }
 
 void RenderOptions::open(Button* /*caller*/) {
@@ -257,6 +270,12 @@ void RenderOptions::open(Button* /*caller*/) {
 	dialogWindow->AddChild(bloom_sldr);
 	y += bloom_sldr->GetHeight() + 2;
 
+	auto exposure_sldr= new Slider("Exposure", 0, 80, 0, y + 2, set_exposure, dialogWindow->GetWidth());
+	exposure_sldr->SetSliderValue(lighting_profile::current_exposure());
+	dialogWindow->AddChild(exposure_sldr);
+	y += exposure_sldr->GetHeight() + 2;
+
+
 	// start tree
 	auto cmp = (Tree*)dialogWindow->AddChild(new Tree("Detail Options Tree", 0, y + 2, nullptr, dialogWindow->GetWidth()));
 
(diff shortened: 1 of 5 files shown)


code/lab/dialogs/render_options.cpp
```diff
@@ -280,4 +357,15 @@ void RenderOptions::open(Button* /*caller*/) {
...
+	cmp->AddItem(tonemapper, "Linear", (int)TonemapperAlgorithm::tnm_Linear, false, change_tonemapper_setting );
+	cmp->AddItem(tonemapper, "Uncharted", (int)TonemapperAlgorithm::tnm_Uncharted, false, change_tonemapper_setting );
+	cmp->AddItem(tonemapper, "ACES", (int)TonemapperAlgorithm::tnm_Aces, false, change_tonemapper_setting );
+	cmp->AddItem(tonemapper, "ACES Approximate", (int)TonemapperAlgorithm::tnm_Aces_Approx, false, change_tonemapper_setting );
+	cmp->AddItem(tonemapper, "Cineon", (int)TonemapperAlgorithm::tnm_Cineon, false, change_tonemapper_setting );
+	cmp->AddItem(tonemapper, "Piecewise Power Curve", (int)TonemapperAlgorithm::tnm_PPC, false, change_tonemapper_setting );
+	cmp->AddItem(tonemapper, "Piecewise Power Curve(RGB)", (int)TonemapperAlgorithm::tnm_PPC_RGB, false, change_tonemapper_setting );
+	cmp->AddItem(tonemapper, "Reinhard Extended", (int)TonemapperAlgorithm::tnm_Reinhard_Extended, false, change_tonemapper_setting );
+	cmp->AddItem(tonemapper, "Reinhard Jodie", (int)TonemapperAlgorithm::tnm_Reinhard_Jodie, false, change_tonemapper_setting );
+
```
JohnAFernandez, line 370: A link in a comment to where we can view what these are might be a good idea.

code/lab/dialogs/render_options.cpp
```diff
@@ -257,6 +304,36 @@ void RenderOptions::open(Button* /*caller*/) {
 	dialogWindow->AddChild(bloom_sldr);
 	y += bloom_sldr->GetHeight() + 2;
 
+	auto exposure_sldr= new Slider("Exposure", 0, 8, 0, y + 2, set_exposure, dialogWindow->GetWidth());
+	exposure_sldr->SetSliderValue(lighting_profile::current_exposure());
+	dialogWindow->AddChild(exposure_sldr);
+	y += exposure_sldr->GetHeight() + 2;
+	auto ppcv = lighting_profile::lab_get_ppc();
```
JohnAFernandez, line 311: Readability: There should be a newline separating this one from the block above it, to make it easier to spot since the blocks below it all use `ppcv` to set their slider values.

code/lab/dialogs/render_options.cpp
```diff
@@ -257,6 +304,36 @@ void RenderOptions::open(Button* /*caller*/) {
...
+
+	auto ppcsl_sldr= new Slider("PPC: Shoulder Length", 0, 10, 0, y + 2, set_ppc_sl, dialogWindow->GetWidth());
+	ppcsl_sldr->SetSliderValue(ppcv.shoulder_length);
+	dialogWindow->AddChild(ppcsl_sldr);
+	y += ppcsl_sldr->GetHeight() + 2;
+
+	auto ppcss_sldr= new Slider("PPC: Shoulder Strength", 0, 1, 0, y + 2, set_ppc_ss, dialogWindow->GetWidth());
+	ppcss_sldr->SetSliderValue(ppcv.shoulder_strength);
+	dialogWindow->AddChild(ppcss_sldr);
+	y += ppcss_sldr->GetHeight() + 2;
```
JohnAFernandez, line 336: Readability: And a new line below this one to separate the block below.

JohnAFernandez (approved):
Ok. So, not knowing that much about graphics, it really does still look like the changes are straightforward.  There are two readability things that I saw, and one thing outside the scope of this PR:

The`default_profile` member at some point should probably be removed from the `lighting_profiles` class, because putting it there has some unintended consequences. None of them major, but it would be a better practice to separate it out in another PR when there's some more time.

JohnAFernandez:
A graphics coder, at least giving this a brief once over before merging, is probably a good idea.

code/lab/dialogs/render_options.cpp
EatThePath: I've opted for a note on where more complete documentation is, that seems more feasible for keeping them in sync in the future if changes are made to the main documentation.

EatThePath:
> A graphics coder, at least giving this a brief once over before merging, is probably a good idea.

Honestly I don't think this needs any graphics expertise is really needed to review this, since it only very barely touches any graphics code.

> The `default_profile` member at some point should probably be removed from the `lighting_profiles` class, because putting it there has some unintended consequences. None of them major, but it would be a better practice to separate it out in another PR when there's some more time.

The storage and use of `default_profile` should change considerably once I implement multiple profile support, so I'll try to make sure I check in about it at that time

PR #3884, opened 2022-01-03
Fix Particles `pageIn()` to use actual number of frames

### LINKED ISSUE #3706: Prefetch particles used in Mission
Particles that can occur in a mission are loaded once they are needed the first time. This can introduce microstutters, due to the required access to the disk. While #3705 improves this already, there are still some microstutters remaining due to the actual access to the disk and loading of the file.
To avoid this, possible particles from weapons / ships / whatever should be prefetched before the mission like other resources.

wookieejedi:
Previously the `bm_page_in_aabitmap` call in `ParticleProperties::pageIn()` used -1 as the number of frames. It seemed particles with more than 1 frame still worked correctly, but likely prevent preloading of particles. This PR fixes that to use the actual number of frames. Fixes #3706  

Thanks for @notimaginative for catching this. Would be useful to have review by @asarium even though it's just a few lines.

### DIFF
--- a/code/particle/util/ParticleProperties.cpp
+++ b/code/particle/util/ParticleProperties.cpp
@@ -73,8 +73,11 @@ WeakParticlePtr ParticleProperties::createPersistentParticle(particle_info& info
 
 void ParticleProperties::pageIn() {
 
-	for (int bitmap: m_bitmap_list) {
-		bm_page_in_aabitmap(bitmap, -1);
+	int nframes = 1;
+
+	for (int bitmap : m_bitmap_list) {
+		bm_get_info(bitmap, nullptr, nullptr, nullptr, &nframes);
+		bm_page_in_aabitmap(bitmap, nframes);
 	}
 
 }


asarium (approved):
Oops.
Although, that is a weird API. Maybe it would be worth looking into fixing the API so it always pages in all frames?

wookieejedi:
> Oops. Although, that is a weird API. Maybe it would be worth looking into fixing the API so it always pages in all frames?

Yeah, that's a good point. Fiing the API would be a seperate/larger PR, so I'll plan to merge this PR at least to fix the current bug. In looking over other uses of `bm_page_in_aabitmap` it seems like loading all frames might not be the best solution in some cases. An example is the wingmen dot in the wingmen HUD, it takes a frame number of 2, and so if there were more frames in the animation I'd imagine loading those would just take up un-needed space?

asarium:
If that is required, you could still convert the default behavior to query the number of frames. It's just a suggestion and I didn't look at the code in detail. As you mentioned, this does not concern the changes in this PR though.

wookieejedi:
Ah, that's a good point. I'll look into it in more detail since it could certainly be useful.

PR #3885, opened 2022-01-03
Fix shadow check on lightshafts

qazwsxal:
This appears to be a bug present since the initial lightshafts code was added, while a check is made to see if the sun is occluded, the `||` conditional short circuits, and causes lightshafts to always be rendered:

image

Had a quick discussion with @TRBlount and @EatThePath and we all agreed that the original intention was probably to block lightshaft rendering when the sun was occluded, so `&&` would make more sense.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -722,7 +722,7 @@ void game_sunspot_process(float frametime)
 			for(idx=0; idx<n_lights; idx++)	{
 				bool in_shadow = shipfx_eye_in_shadow(&Eye_position, Viewer_obj, idx);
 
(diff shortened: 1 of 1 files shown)


JohnAFernandez (approved):

JohnAFernandez:
I'll be curious to see what it looks like after the change.

EatThePath:
> I'll be curious to see what it looks like after the change.

I made a quick clip at @qazwsxal's request https://www.youtube.com/watch?v=d94AkQLO5RM

Goober5000:
The original commit that added the `||` logic was a0d1c8445a.

PR #3887, opened 2022-01-03
file loading performance fixes

notimaginative:
Fix a broken cfile optimization and tweak bm_load() so that it can take advantage of existing optimizations when searching for files in a known path type.


wookieejedi:
In qualitatively testing it seems to speed loads up a bit.

JohnAFernandez (approved):
Straightforward

Goober5000 (approved):

PR #3888, opened 2022-01-03
add game_settings flag to not target turrets

Goober5000:
In FS2, the HUD will automatically target the first turret on any BIG or HUGE ship.  In FS1, this did not happen.  This adds a flag to disable the new behavior.

The change is needed because the scan view cone is significantly constrained when any turrets or subsystems are targeted.  This makes Playing Judas quite a bit more difficult than it was designed to be.


wookieejedi (approved):
This would be useful indeed. LGTM

PR #3891, opened 2022-01-04
Handle forgotten 0 case for new vecmat function

BMagnu:
The vecmat function introduced in #3877 can, if axis and target exactly match in a 90 degree angle, run into a zero case in which the math simplifies, but notably also doesn't properly compute with how I've done it, requiring an early exit.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -1608,9 +1608,15 @@ float vm_closest_angle_to_matrix(const matrix* mat, const vec3d* rot_axis, float
 	const float z = (a[0]*(m[5]-m[7]) + a[1]*(-m[2]+m[6]) + a[2]*(m[1]-m[3]));
 
 	// If both y and z are close to 0, then the rotation axis points in the same direction as the matrix, thus any orientation r would be perpendicular to m
-	if(fabs(y) < 0.001f && fabs(z) < 0.001f){
-		*angle = 0.0f;
-		return PI_2;
+	// If y is 0, the rest of the math simplifies, and we always find the angle at pi/2
+	if(fabs(y) < 0.01f) {
+		if (fabs(z) < 0.01f) {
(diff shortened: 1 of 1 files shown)


code/math/vecmat.cpp
```diff
@@ -1608,9 +1608,15 @@ float vm_closest_angle_to_matrix(const matrix* mat, const vec3d* rot_axis, float
 	const float z = (a[0]*(m[5]-m[7]) + a[1]*(-m[2]+m[6]) + a[2]*(m[1]-m[3]));
 
 	// If both y and z are close to 0, then the rotation axis points in the same direction as the matrix, thus any orientation r would be perpendicular to m
-	if(fabs(y) < 0.001f && fabs(z) < 0.001f){
-		*angle = 0.0f;
-		return PI_2;
+	// If y is 0, the rest of the math simplifies, and we always find the angle at pi/2
+	if(fabs(y) < 0.01f) {
+		if (fabs(z) < 0.01f) {
```
Baezon, line 1613: Was it intentional to change these to 0.01 from 0.001? They are getting a bit... large

Baezon (changes requested):

code/math/vecmat.cpp
BMagnu: I was a bit scared of something like a 0.0011 going through the function, and it then being jumbled around, possibly introducing enough error to actually flip to 0, which will cause Inf * 0, and the resulting nans are problems.
Though maybe 0.001 is still fine.
A y close to 0 should always be pretty close to pi/2 anyways, but I can increase it back if you think that's still safe

Baezon: I suppose that's fair, I just wanted to throw it by you first, in case it was a mistake. I'd be fine with either way.

Baezon (approved):

PR #3893, opened 2022-01-04
Fix Particle `PageIn()` (V2)

### LINKED ISSUE #3890: Broken Impact Effect (?) on current build (20220104)
I noticed broken Impact Effect (?) in some weapons on current build i.e. 20220104
screen0805

wookieejedi:
It turns out the particle preloading fix revealed a long standing bug in the particle `pageIn()` function--it should be calling `bm_page_in_texture()` instead of `bm_page_in_aabitmap()`. It's been wrong the entire time and we didn't notice because of the frame bug, which was fixed in #3884. Without this fix, when it's preloaded it locks the image as 8-bit, when it force loads it's 24-bit. 

Tested and works as expected. Fixes #3890. It also has the added benefit of making the code cleaner because the `bm_page_in_texture()` loads the frames if none are specified. 

Massive thanks to taylor for spotting this!

### DIFF
--- a/code/particle/util/ParticleProperties.cpp
+++ b/code/particle/util/ParticleProperties.cpp
@@ -73,11 +73,8 @@ WeakParticlePtr ParticleProperties::createPersistentParticle(particle_info& info
 
 void ParticleProperties::pageIn() {
 
-	int nframes = 1;
-
 	for (int bitmap : m_bitmap_list) {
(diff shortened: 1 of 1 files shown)


notimaginative (approved):
LGTM. And I forgot that `bm_page_in_texture()` will load all frames so good for catching that too!

asarium (approved):
Well, now that is a perfect case where copy/paste went wrong...

PR #3897, opened 2022-01-06
fix first-of to return the correct arguments

### LINKED ISSUE #3896: the first-of sexp works incorrectly
The `first-of` sexp returns its arguments in forward order, as it should, but it selects them from the *end* of the list, not the beginning.  This is because the arguments are returned in reverse order due to being treated as a stack, and the feature didn't account for that.

Thanks to 0rph3u5 for identifying the bug.  A fix will be up shortly.

Goober5000:
Change the threshold logic to return the correct elements from the beginning of the list, not the end.  Fixes #3896.


wookieejedi (approved):
LGTM

Issue #3892, opened 2022-01-04
Crash writing player config file (Controls5)


notimaginative:
`control_config_get_current_preset()` has no safety against returning `Control_config_presets.end()` which can lead to a crash.

It doesn't appear to be crashing in general, only at the next pilot file save after controls have been modified and saved to a different preset (and/or saved to an existing preset??).
 
```(gdb) bt
#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:96
#1  0x0000555556c57ddb in pilot::JSONFileHandler::writeString(char const*, char const*) (this=0x55555b547dd0, name=0x55555712e29b "preset", str=0x14171300000027 <error: Cannot access memory at address 0x14171300000027>) at ../../code/pilotfile/JSONFileHandler.cpp:115
#2  0x00005555569b93e1 in pilotfile::plr_write_controls() (this=0x555557cd3600 <Pilot>) at ../../code/pilotfile/plr.cpp:733
#3  0x00005555569bae04 in pilotfile::save_player(player*) (this=0x555557cd3600 <Pilot>, _p=0x0) at ../../code/pilotfile/plr.cpp:1133
#4  0x00005555566cb3b2 in sim_room_close() () at ../../code/menuui/readyroom.cpp:1189
#5  0x00005555563030b0 in game_leave_state(int, int) (old_state=20, new_state=51) at ../../freespace2/freespace.cpp:5094
#6  0x0000555556432a03 in gameseq_set_state(int, int) (new_state=51, override=0) at ../../code/gamesequence/gamesequence.cpp:280
#7  0x0000555556302061 in game_process_event(int, int) (current_state=20, event=1) at ../../freespace2/freespace.cpp:4634
#8  0x0000555556432ede in gameseq_process_events() () at ../../code/gamesequence/gamesequence.cpp:379
#9  0x0000555556305733 in game_main(int, char**) (argc=3, argv=0x7fffffffde88) at ../../freespace2/freespace.cpp:6595
#10 0x00005555563079d3 in main(int, char**) (argc=3, argv=0x7fffffffde88) at ../../freespace2/freespace.cpp:7657```

notimaginative:
Just to add: this is deleting my active pilot file every time it happens.

wookieejedi:
Can confirm, same thing happens to me

ananace:
[...]

PR #3898, opened 2022-01-06
Fix 3892

z64555:
Should fix issue #3892 

Add catches for edge cases when control_config_get_current_preset() returnes ::end() within csg and plr read/write.

* Add assertsions during csg and plr read/write to make sure the Controls_config is initialized.


notimaginative:
It may not crash (haven't tested), but this is just going to assert every time it happens, which is frequently. So I'm not sure this really solves anything in my case at least.

PR #3899, opened 2022-01-06
add scripting hook to run after object render

Goober5000:


Goober5000:
On second thought, this may not actually be necessary - at least for the use case I coded it for.

Issue #3902, opened 2022-01-07
pre-Controls 4/5 builds error on pilot select screen if last pilot is from newer build


notimaginative:
When initializing the pilot select screen it checks the last used pilot. If that pilot has the new control setup (controls 4/5) then it will crash older builds while attempting to read the controls section. The only way around the crash is to remove or change the "LastPlayer" entry in the ini file.

Though I do not consider this a bug in controls 4/5, but a fundamental issue with the pilot code in general, the new controls code could be altered to handle this better.  Notably, if it used a new controls section ("controls_new", or something) and either discarded the old "controls" section after conversion or just saved it to be written back later for compatibility sake. Old builds would still discard the new controls section, but with that info now in a separate file it isn't that big of a loss.

```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff5d0b537 in __GI_abort () at abort.c:79
#2  0x00000000015ae4f7 in os::dialogs::Error(char const*) (text=0x4dbfb00 "Error: JSON reading requires a value with name 'controls' but there is no such value!\nFile: JSONFileHandler.cpp\nLine: 209\n") at /home/taylor/Working/src/code/fs2open/code/osapi/dialogs.cpp:293
#3  0x00000000015af461 in os::dialogs::Error(char const*, int, char const*, ...) (filename=0x1e68801 "JSONFileHandler.cpp", line=209, format=0x1e68b8d "JSON reading requires a value with name '%s' but there is no such value!") at /home/taylor/Working/src/code/fs2open/code/osapi/dialogs.cpp:285
#4  0x000000000191e4c8 in pilot::JSONFileHandler::ensureExists(char const*) (this=0x3395b00, name=0x1e3705b "controls") at /home/taylor/Working/src/code/fs2open/code/pilotfile/JSONFileHandler.cpp:209
#5  0x000000000191eba9 in pilot::JSONFileHandler::startArrayRead(char const*, bool) (this=0x3395b00, name=0x1e3705b "controls") at /home/taylor/Working/src/code/fs2open/code/pilotfile/JSONFileHandler.cpp:310
#6  0x000000000166aa61 in pilotfile::plr_read_controls() (this=0x29aac18 <Pilot>) at /home/taylor/Working/src/code/fs2open/code/pilotfile/plr.cpp:657
[...]

PR #3905, opened 2022-01-08
Fix #3901: Inverted Controls Don't Save

### LINKED ISSUE #3901: Inverted Controls don't save
With Controls4/5 #2958, it seems that inverted controls no longer save.
When setting them in mission, they are set for the mission only, set elsewhere and they will not be applied at all.

z64555:
Preset identification should only be invert agnostic when reading a CSG<7 file, for all other uses of control_config_get_current_preset(), it should be exact equality with the preset.

Agnostic equals was added by @Goober5000 to address an issue with LUA scripts that attempted to invert axes controls.  In addition, there was a bug where a tester was switching back and forth between pre-Controls5 test builds only to have the conversion process continously trigger, because the CSG files did not save the invert flag and its generated preset was not exactly equal with the preset generated by his player file or current bindings.  PLR files did have the invert flag in them, however.

!!Notice!! I don't have time right now to do a full build test nor a run test, so please test this before merging.

Switching back and forth between pre and post builds may work, but another bug (Issue #3902 ) may cause a crash if you try to do so, should the "controls" key in the PLR.json get deleted (or rather not written) if you have a new pilot created in Controls5.  The workaround to this currently is to have an old pilotfile from before Controls5 be converted, or manually edit in an empty "controls: []" section in the .json.

### DIFF
--- a/code/controlconfig/controlsconfig.cpp
+++ b/code/controlconfig/controlsconfig.cpp
@@ -1584,7 +1584,7 @@ void control_config_close()
 	Undo_controls.clear();
 }
 
-SCP_vector<CC_preset>::iterator control_config_get_current_preset() {
+SCP_vector<CC_preset>::iterator control_config_get_current_preset(bool invert_agnostic) {
 	// Find the matching preset.
 	// We do this instead of relying on Defaults_cycle_pos because the player may end up duplicating a preset
 	auto it = Control_config_presets.begin();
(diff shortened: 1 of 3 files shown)


Goober5000 (approved):
Looks good as far as it goes.  I haven't tested it myself though.

z64555:
Alright, I did a quick test and it does as advertised, which is detect and save inversion. Still need to test:

-[ ] That it doesn't break the scripting inversion
-[ ] That conversions of CSG's from other pilots don't clash when they should have the same, or similar presets
-[ ] Make sure I'm not missing any other edge cases that I can't think of right now

Goober5000:
Tested the scripting inversion on this branch and it works. 👍

PR #3908, opened 2022-01-08
Fix text overflow on multi-related screen

wookieejedi:
Multiplayer screens such as PXO lobby, chat boxes, and multi-game setup screens had hard coded values for the number of lines to display. These would only work correctly with the default .vf font01 from retail. If a mod or user used a TTF or other .vf font than text would spill over. This PR makes these values take into the account the font used.
TTF_Spill

Tested and works as expected with TTF fonts and the retail fonts.  Marking as draft to see what others think.


Goober5000 (approved):
This is a great solution.  Clean, easy to understand, and minimally invasive.

PR #3913, opened 2022-01-10
Add AI to Fix Negative Turret Ammo

wookieejedi:
Previously, turrets with ammo would end up firing one extra shot, resulting in a -1 ammo.

This was because of the check `ammo[bank_to_fire] >= 0`, in `aiturret.cpp`. This PR adds an AI flag to have the turret only fire if the ammo is greater than 0 (`ammo[bank_to_fire] > 0`), as opposed to greater than or equal to 0.

Tested and the flag fix works as expected.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1907,6 +1907,7 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 						if (swp->primary_bank_ammo[bank_to_fire] >= points) {
 							swp->primary_bank_ammo[bank_to_fire] -= points;
 						} else if ((swp->primary_bank_ammo[bank_to_fire] >= 0) && !(The_mission.ai_profile->flags[AI::Profile_Flags::Prevent_negative_turret_ammo])) {
+							// default behavior allowed ammo to be negative
 							swp->primary_bank_ammo[bank_to_fire] -= points;
 						} else if (!(turret->system_info->flags[Model::Subsystem_Flags::Use_multiple_guns]) && (swp->primary_bank_ammo[bank_to_fire] < 0)) {
 							swp->current_primary_bank++;
@@ -1948,6 +1949,7 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 							if (swp->secondary_bank_ammo[bank_to_fire] > 0) {
 								swp->secondary_bank_ammo[bank_to_fire]--;
 							} else if ((swp->secondary_bank_ammo[bank_to_fire] == 0) && !(The_mission.ai_profile->flags[AI::Profile_Flags::Prevent_negative_turret_ammo])) {
+								// default behavior allowed ammo to be negative
 								swp->secondary_bank_ammo[bank_to_fire]--;
 							} else if (!(turret->system_info->flags[Model::Subsystem_Flags::Use_multiple_guns]) && (swp->secondary_bank_ammo[bank_to_fire] < 0)) {
 								swp->current_secondary_bank++;
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1904,7 +1904,9 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 							points = 1;
 						}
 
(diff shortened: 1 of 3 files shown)


code/ai/aiturret.cpp
```diff
@@ -1904,7 +1904,9 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum,
 							points = 1;
 						}
 
-						if (swp->primary_bank_ammo[bank_to_fire] >= 0) {
+						if (swp->primary_bank_ammo[bank_to_fire] > 0) {
```
Baezon, line 1907: In this case, we probably want `points`, since that's what it's going to attempt to subtract, since it might be more than one shot it's doing.

Baezon (changes requested):

code/ai/aiturret.cpp
wookieejedi: Good point, pushed. Also altered the logical operators to match.

Goober5000 (commented):
Does this actually need an ai_profiles flag?  This looks like a clear-cut bug fix.

wookieejedi:
I was thinking about that too, though I'm guessing that it could affect previous mod balance since having this on means turrets get 1 less shot (even if it is correct)?

Goober5000:
I would assume those mods expect turrets to use the correct ammo, so I wouldn't worry about that.

Also, I suspect the impact would be minor.  If anyone complains, the flag could always be added later.

code/ai/aiturret.cpp
```diff
@@ -1904,7 +1904,9 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum,
 							points = 1;
 						}
 
-						if (swp->primary_bank_ammo[bank_to_fire] >= 0) {
+						if (swp->primary_bank_ammo[bank_to_fire] >= points) {
+							swp->primary_bank_ammo[bank_to_fire] -= points;
+						} else if ((swp->primary_bank_ammo[bank_to_fire] >= 0) && !(The_mission.ai_profile->flags[AI::Profile_Flags::Prevent_negative_turret_ammo])) {
```
JohnAFernandez, line 1909: This doesn't seem to make sense, actually. The ammo can still end up being negative. This would fire from each primary point, right? So isn't it still firing too many shots?  Looking for clarification.

wookieejedi: For the check on line 1909? Or on 1907?

JohnAFernandez: 1909

wookieejedi: Correct, 1909 is just the default behavior, which allowed the ammo to go negative. :)

JohnAFernandez: And that's why you don't comment on PRs after a three day drive.

wookieejedi: Perhaps would have made more sense to put that line first (to correspond to where the old check was, and to add a comment about why). Be happy to add that if folks want.

JohnAFernandez:
I would go the other way and add the flag now.

Baezon:
I was also unsure about having this be a flag, my preference is of course a clear cut fix, but balance can be strongly affected even in the most innocuous ways. If it *is* a flag, it should be default on for a recent targeted version.

wookieejedi:
Good point, added target for 22.

wookieejedi:
I'd reckon it seems like the vote is leaning toward keeping it as a flag?

Goober5000:
How many mods use ammo counts for turrets?  This is a very recent feature.

In fact when were ammo counts added?

Goober5000:
Hmm.  I seem to recall, though, that they didn't actually *work* until a fix that was done much more recently.

wookieejedi:
Hmm the only PRs related to that I can see are that #3589 in August broke turret ammo, then #3735 in October fixed it.

Goober5000:
Ah ok, so turret ammo has in fact been working since its original introduction.  In that case, and since the feature is not new, a flag makes sense.

wookieejedi:
Sounds good, I'll keep it as is. Looks like only thing left then is @Baezon 's follow up approval.

Baezon (approved):

Issue #3912, opened 2022-01-10
silent crash to desktop on attempting to display name of mouse button


EatThePath:
first reported on discord by DMK85, recent builds crash during the first FS2 training mission, during the weapons instructions with the amazon. My debugging points to the crash happening when the mission tries to display a keybinding for a mouse key. `CC_bind::textify()` accesses `Mouse_button_text[]` to get the button name, and my debugger shows that array as being populated entirely by null values at that time

```
fs2_open_21_5_0_x64_AVX2-DEBUG.exe!strlen() Line 70 (c:\Program Files\Microsoft VS Code\minkernel\crts\ucrt\src\appcrt\string\amd64\strlen.asm:70)
fs2_open_21_5_0_x64_AVX2-DEBUG.exe!std::_Narrow_char_traits<char,int>::length(const char * const _First) Line 414 (c:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\xstring:414)
fs2_open_21_5_0_x64_AVX2-DEBUG.exe!std::string::assign(const char * const _Ptr) Line 3264 (c:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\xstring:3264)
fs2_open_21_5_0_x64_AVX2-DEBUG.exe!std::string::basic_string<char,std::char_traits<char>,std::allocator<char>>(const char * const _Ptr) Line 2522 (c:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\xstring:2522)
fs2_open_21_5_0_x64_AVX2-DEBUG.exe!CC_bind::textify() Line 2419 (d:\Code\fso2\src\code\controlconfig\controlsconfigcommon.cpp:2419)
fs2_open_21_5_0_x64_AVX2-DEBUG.exe!translate_key(char * key) Line 925 (d:\Code\fso2\src\code\controlconfig\controlsconfigcommon.cpp:925)
fs2_open_21_5_0_x64_AVX2-DEBUG.exe!message_translate_tokens(const char * text) Line 661 (d:\Code\fso2\src\code\mission\missiontraining.cpp:661)
fs2_open_21_5_0_x64_AVX2-DEBUG.exe!message_training_setup(int m, int length, char * special_message) Line 819 (d:\Code\fso2\src\code\mission\missiontraining.cpp:819)
[...]

PR #3914, opened 2022-01-11
Fix issue #3912

z64555:
Move dynamic XSTR stuff out of `controlsconfig.cpp` and move it into `controlsconfigcommon.cpp`.
- Needed because these strings are used in places other than the controls menu.

### DIFF
--- a/code/controlconfig/controlsconfig.cpp
+++ b/code/controlconfig/controlsconfig.cpp
@@ -1516,22 +1516,6 @@ void control_config_init()
 	control_config_conflict_check();
 
 	// setup strings
-	// Old XSTR indices have been put beside their corresponding label in case we use them again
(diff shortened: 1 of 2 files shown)


Goober5000 (approved):

Goober5000:
Looks good

z64555:
clang wants to use a range-based for, but I don't know if it would play nice with `vm_free()` (which is `std::free()`)

Goober5000:
The range-based `for` would be iterating over pointers, so that should still be fine.  Or you could put `//NOLINT` at the end of the lines to stop clang from complaining.

z64555:
Alright, I put in the range-based for loops but... it looks like VS optimized it away.  I set a breakpoint to enter `controls_config_common_close()` but it never triggered, nor did it even enter `game_shutdown()` nor was I able to set the breakpoint on `game_shutdown()` in `freespace.cpp:6602`

PR #3915, opened 2022-01-11
Adds -deadzone parameter to command line

Kiloku:
This allows for finer control of the joystick deadzone, as a number from 0 to 100. 
When this is set, the in-game deadzone slider is disabled.


code/cmdline/cmdline.cpp
```diff
@@ -382,6 +382,7 @@ cmdline_parm use_warp_flash("-warp_flash", NULL, AT_NONE);	// Cmdline_warp_flash
 cmdline_parm allow_autpilot_interrupt("-no_ap_interrupt", NULL, AT_NONE);
 cmdline_parm stretch_menu("-stretch_menu", NULL, AT_NONE);	// Cmdline_stretch_menu
 cmdline_parm no_screenshake("-no_screenshake", nullptr, AT_NONE); // Cmdline_no_screenshake
+cmdline_parm deadzone("-deadzone", "Sets the joystick deadzone. Integer value from 0 to 100. Disables deadzone slider in the in-game Options menu.", AT_INT); //Cmdline_deadzone
```
wookieejedi, line 385: Should this save `Overrides deadzone slicer` instead of `Disables`? `Disables` makes it sound like the buttons on the options screen will not even show up.

code/controlconfig/controlsconfig.cpp
```diff
@@ -482,7 +483,12 @@ int joy_get_scaled_reading(int raw)
 
 	raw -= JOY_AXIS_CENTER;
 
-	dead_zone = (JOY_AXIS_MAX - JOY_AXIS_MIN) * Joy_dead_zone_size / 100;
+	if (Cmdline_deadzone >= 0)	{
+		dead_zone = (JOY_AXIS_MAX - JOY_AXIS_MIN) * Cmdline_deadzone / 200; //Making this div by 200, the max deadzone value in cmdline is 100, which is more intuitive
+	}
+	else{
```
wookieejedi, line 489: Nitpick comment, but have the `} else {` on the same line with a space after the else is more consistent with other code formatting.

code/menuui/optionsmenu.cpp
```diff
@@ -542,6 +542,11 @@ void options_tab_setup(int  /*set_palette*/)
 			Options_sliders[gr_screen.res][i].slider.enable();
 			Options_sliders[gr_screen.res][i].slider.unhide();
 		}		
+		if (Cmdline_deadzone >= 0){
```
wookieejedi, line 545: Nitpick, but a space between `)` and `{` would be good here and in the following two spots.

z64555:
Erm, I'm still tied up with the mouse-fly changes for Controls5, I'll try to give this a review as soon as I'm done.

code/cmdline/cmdline.cpp
z64555: Its disabled in the same way that the difficulty slider is disabled, the background image is there, but the dots are either all black or otherwise won't accept input from the user.

code/cmdline/cmdline.cpp
```diff
@@ -1826,6 +1828,10 @@ bool SetCmdlineParams()
 		Cmdline_no_screenshake = 1;
 	}
 	
+	if (deadzone.found()){
```
z64555, line 1831: Please be consistent with the style.

`if ( deadzon.found() ) {`

code/cmdline/cmdline.cpp
z64555: Recommend stating somewhere in the description that the max deadzone size is 50% of the joystick range.

code/controlconfig/controlsconfig.cpp
z64555: Please be consistent with the style.

`} else {`

z64555: Maybe add comment above or close by explaining what the max deadzone size is, since its not immediately apparent just looking at this line here.

Original code has `Joy_dead_zone_size` to be 5 * the number of dots the slider has (0 to 10), thus has the range of 0 to 50.  Thus, `Joy_dead_zone_size / 100` has the range of 0f to 0.5f

code/menuui/optionsmenu.cpp
z64555: The bracing style in this file is inconsistent, but I agree with wookiejedi here.

code/controlconfig/controlsconfig.cpp
Kiloku: My intent was to allow the user to go beyond the original maximum as well. The way the PR is right now, the user could have the deadzone be 99% of the range, or more (making the joystick unresponsive)

Kiloku:
Sorry it took me so long for such small fixes. Had a busy week

z64555:
Bleh, this got buried somehow over the past 2 weeks.  Changes look good, did you do a test on your machine to check that the cmdline value of 100 means 100% deadzone?

Kiloku:
I did. Also tested with 99 to make sure that the threshold was correct and it worked as expected.

z64555 (approved):

PR #3918, opened 2022-01-12
Clean up Player_orders

BMagnu:
Previously, possible player orders were stored redundantly in three separate, fixed-size arrays: ``Comm_orders``, ``Sexp_comm_orders``, and ``Player_orders``.  This PR does away with these and creates a new ``Player_orders`` vector, allowing for a dynamic amount of player orders. In preparation for eventually changing up player orders in such a way that the old ID's are removed and not based on bitfields, thus allowing for more than 32 different ones, ships now store player order acceptance as sets of the index of the order in the Player_orders vector. 

This is needed to eventually facilitate LuaAI, a feature that allows a scripter to add AI logic based on Lua which can be added both by SEXP, as well as made available as a player order.

~Still needs testing in addition to a review:~ Finished testing, it seems like everything still works as it should.
- [x] Player orders work as expected ingame
- [x] The query-order training sexp works as expected
- [x] Player orders properly parse, show and save in FRED
- [x] Player orders properly parse, show and save in QtFRED

### DIFF
--- a/fred2/ignoreordersdlg.cpp
+++ b/fred2/ignoreordersdlg.cpp
@@ -108,7 +108,7 @@ BOOL ignore_orders_dlg::OnInitDialog()
 		// Need to add more checkboxes.
 		if (m_num_checks_active >= MAX_CHECKBOXES)
 		{
-			Warning("Tried to add more orders than FRED2 has checkboxes for them (can be max %d)!", MAX_CHECKBOXES);
+			Warning(LOCATION, "Tried to add more orders than FRED2 has checkboxes for them (can be max %d)!", MAX_CHECKBOXES);
 			break;
 		}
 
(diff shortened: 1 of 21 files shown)


fred2/ignoreordersdlg.cpp
```diff
@@ -104,22 +103,19 @@ BOOL ignore_orders_dlg::OnInitDialog()
...
-			if (m_num_checks_active >= MAX_CHECKBOXES)
-			{
-				Int3();
-				break;
-			}
-
-			check_boxes[m_num_checks_active].button->SetWindowText(Comm_orders[i].name.c_str());
-			check_boxes[m_num_checks_active].id = Comm_orders[i].item;
-			m_num_checks_active++;
+			Int3();
```
Goober5000, line 111: Display a `Warning` here, since `Int3` will cause crashes in some builds.

Goober5000 (changes requested):
Looks ok, I think.  One comment.

Goober5000 (approved):

PR #3922, opened 2022-01-13
3920 qtFRED save only saves once

### LINKED ISSUE #3920: qtFRED save only saves once
During debugging of some other features, I noticed that changing parts of the mission and hitting Ctrl+S did not write anything to disk except the first time qtFRED saved to that specific file. Save as still works, but hitting save after save as will also not write anything to disk a second time. 
I tried both changing options within a ship, as well as moving it, but that did not change the save behaviour.

TheForce172:
Fix #3920 Actually save on save button rather than only on save as.


Goober5000 (approved):
LGTM

PR #3923, opened 2022-01-14
Standalone fixes/optimizations

notimaginative:
Various fixes and optimizations for standalone servers, plus minor general fixes.

- reduces memory usage for standalones 35-40 MB by not initializing unneeded stuff
- fix crash on standalone reset if level has never loaded
- fix small portfwd memory leak

### DIFF
--- a/code/graphics/2d.cpp
+++ b/code/graphics/2d.cpp
@@ -1605,8 +1605,6 @@ bool gr_init(std::unique_ptr<os::GraphicsOperations>&& graphicsOps, int d_mode, 
 		return false;
 	}
 
-	gr_light_init();
-
 	gr_set_palette_internal(Gr_current_palette_name, NULL, 0);
 
 	bm_init();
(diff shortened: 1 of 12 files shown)


notimaginative:
@asarium Please check that I didn't do anything too weird with the particle stuff. I tested it all pretty well but I'm not too familiar with that newer code. All of the particle changes are in the single commit (second from top) so you should be able to check it pretty easily.

asarium (approved):

asarium:
Seems good to me.

notimaginative:
Hmm, font related changes are causing a problem. All of the font stuff takes 4+ MB and is otherwise useless on standalones, but it's difficult to disable entirely. I tried to cheat a bit by not initializing the path stuff, which saves 2+ MB, but the lack of error handling causes null ptr crashes.

Have to figure out if wasting that RAM is worth a proper fix or not.

notimaginative:
@Goober5000 If you've got some time I'd appreciate a second opinion on this before it gets merged.

Goober5000:
@notimaginative I'll take a look later today.

Goober5000 (approved):
Yup, looks good to me.

PR #3924, opened 2022-01-14
Fix incorrect net signature for wave ships

wookieejedi:
Previously, multiplayer missions with wings with waves would incorrectly set the `net_signature` of ships in the initial wave. The code is supposed to only set the `net_signature` if it was past the initial wave. This is because original ships in the wave have their `net signature` set at mission parse. (Note, the check `wingp->current_wave > 1` is also used in the docking code to check for initial wave or not status).

This PR fixes that and has been confirmed to work via multiplayer tests.


JohnAFernandez:
Bug shout out to fotg for finding it

notimaginative (approved):
LGTM

PR #3925, opened 2022-01-15
Allow AI flag for bay orientation to account for parent `uvec`

wookieejedi:
The previously added flag `Fighterbay_arrivals_use_carrier_orient` (#3698) incorporated the parent ship's `rvec`, which greatly helped align emerging ships from fighter bays. In parent some ships though, emerging ships still showed a `uvec` that did not match the parent ship (such as FotG's Star Destroyer). This PR allows the `Fighterbay_arrivals_use_carrier_orient` flag to take parent `uvec` into account.

Note, that departing the mission into a bay already has code to ensure a `wheels-down` landing.

PR is tested and works as expected/fixes `uvec` issue.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -13264,21 +13264,21 @@ int ai_acquire_emerge_path(object *pl_objp, int parent_objnum, int allowed_path_
 	pl_objp->phys_info.forward_thrust = 0.0f;		// How much the forward thruster is applied.  0-1.
 	pl_objp->pos = pos;
 
-	// use the modder-defined arrival rvec, if applicable
 	vec3d rvec;		vm_vec_zero(&rvec);
+	vec3d uvec;		vm_vec_zero(&uvec);
+
+	// use the modder-defined arrival rvec, if applicable
 	SCP_string pathName(pm->paths[path_index].name);
 	if (parent_sip->pathMetadata.find(pathName) != parent_sip->pathMetadata.end() && !IS_VEC_NULL(&parent_sip->pathMetadata[pathName].arrival_rvec)) {
 		vm_vec_copy_normalize(&rvec, &parent_sip->pathMetadata[pathName].arrival_rvec);
 		vm_vec_unrotate(&rvec, &rvec, &Objects[aip->path_objnum].orient);
 	}
 	else if (The_mission.ai_profile->flags[AI::Profile_Flags::Fighterbay_arrivals_use_carrier_orient]) {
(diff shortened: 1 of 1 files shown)


Baezon:
I'm a little dubious. In the example, the ships are coming out of the right side of the parent ship, and thus do not share an orientation with their parent ship, and really shouldn't be using it, they only happen to share a uvec. Perhaps this should be in the documentation, but this is really only for ships like the Orion, whose launching ships use the carrier's exact same orientation. If they use anything else, they should probably be using `$Path Metadata` and `+arrival rvec:` to explicitly provide the new orientation.

That said, there's nothing really wrong with this change, but the existing method only failed because the intended orientation deviated so much from the carrier's orientation.

code/ai/aicode.cpp
```diff
@@ -13274,7 +13274,14 @@ int ai_acquire_emerge_path(object *pl_objp, int parent_objnum, int allowed_path_
 	else if (The_mission.ai_profile->flags[AI::Profile_Flags::Fighterbay_arrivals_use_carrier_orient]) {
 		rvec = parent_objp->orient.vec.rvec;
 	}
-	vm_vector_2_matrix(&pl_objp->orient, &fvec, nullptr, IS_VEC_NULL(&rvec) ? nullptr : &rvec);
+
+	// use parent ship uvec, if applicable
+	vec3d uvec;		vm_vec_zero(&uvec);
+	if (The_mission.ai_profile->flags[AI::Profile_Flags::Fighterbay_arrivals_use_carrier_orient]) {
+		uvec = parent_objp->orient.vec.uvec;
```
Baezon, line 13281: Let's put all this alongside the `rvec` stuff, assigning both rvec and uvec in the same block above.

Baezon (changes requested):
After hashing it out on discord, we've figured things out. The flag *is* working as expected. But it could work better. Although there are cases where the flag might be used even when the intended arrival orient isn't the same as its carrier's, it should at least work well enough to function as a decent 'lazy' option. Where I think it is more intuitive that the carrier's `uvec` is preferable to attempt to match first, which avoids unintutive scenarios like rear-facing arrivals ending up side down (a consequence of preferring the rvec over the uvec).

code/ai/aicode.cpp
wookieejedi: Thanks for summarizing the discussion. Updated following feedback.

Baezon (approved):

PR #3927, opened 2022-01-15
FFmpeg deprecations fix for FFmpeg >= 5

tgurr:
```
FFmpeg commit e67e02d15672 [1] dropped avcodec.h from avformat.h includes hence
add an include for it to fix build after that commit.

[1]: https://github.com/FFmpeg/FFmpeg/commit/e67e02d15672a87da1b0566e197a1e19dc7e1e33
```


Goober5000 (approved):
LGTM.  Great catch.

PR #3936, opened 2022-01-17
allow dockpoints to be anchored to submodels explicitly

Goober5000:
By anchoring a dockpoint to a submodel, the dockpoint will move when the submodel moves.  Historically this has been done by inspecting the path associated with a dockpoint, but this method will not work if there is no path or no rotation mode specified.  Here is an alternate method that allows the submodel to be specified explicitly through a dockpoint property.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -9480,6 +9480,14 @@ int find_parent_rotating_submodel(polymodel *pm, int dock_index)
 	Assertion(dock_index >= 0 && dock_index < pm->n_docks, "for model %s, dock_index %d must be >= 0 and < %d!", pm->filename, dock_index, pm->n_docks);
 	int path_num, submodel;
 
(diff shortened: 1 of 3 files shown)


wookieejedi (commented):
Overall looks good. Curious what the justification was for changing `model_maybe_adjust_movement_axis` to `maybe_adjust_movement_axis`?

Goober5000:
Just because that function is confined to the one file.  Some other functions don't have the `model_` prefix, and I figured the distinction was that `model_` functions are meant to be part of the public API while the others aren't.  But that isn't followed 100% and I can change it back if you think I should.

wookieejedi:
Ah that makes sense. Yeah, given that the ideal would be to have the prefix for only the public API I'd say it's fine to keep it as is and not have to put it back in the name.

wookieejedi (approved):

PR #3937, opened 2022-01-18
fix missing texture bug on standalone

notimaginative:
The file used by bmp_load()/bm_load_animation() is a hardcoded interface image on standalone servers. If the value of dir_type isn't ALL or INTERFACE then it can't find the image. For ship textures this can cause primary weapon hits to not register on the server.


wookieejedi (approved):
Fantastic find

Goober5000 (approved):
Wow.  Nicely done!

PR #3938, opened 2022-01-18
log whether command briefing is using scroll buttons

### LINKED ISSUE #3550: Command Briefing Scroll Buttons Disabled for TC
When we first added scroll buttons 19 year ago, we disabled them if there was not a custom background specified.  This is probably an incomplete check to maintain retail behavior.  If a Total Conversion simply replaces the original bitmaps but has a long Command Briefing, it will simply disable the scroll buttons without the modder having any idea why.  Better to have a check for retail before disabling.

See https://github.com/scp-fs2open/fs2open.github.com/blob/4ef212157a6c21fa61ff52d319163c9f6eea3446/code/missionui/missioncmdbrief.cpp

line 571

Goober5000:
Add a logging statement to let modders know that the command briefing is deliberately using, or not using, scroll buttons.

Fixes #3550.


jg18 (approved):
LGTM

JohnAFernandez (approved):

PR #3941, opened 2022-01-19
Update links and edit text in FRED's and qtFRED's About dialogs

jg18:


fred2/fred.rc
```diff
@@ -670,7 +670,7 @@ BEGIN
     LTEXT           ABOUT_LINE_1,IDC_STATIC,39,8,184,10,SS_NOPREFIX
     LTEXT           L"Copyright \x00A9 1999 Volition, Inc.  All Rights Reserved",IDC_STATIC,39,17,164,8
     DEFPUSHBUTTON   "OK",IDOK,229,7,48,14,WS_GROUP
-    LTEXT           """Fred2 is the omega of all giant unwieldy pieces of code. Its big, its horrifying, and it just doesn't care. View it at your own risk."" - Dave Baranec",IDC_STATIC,39,62,238,20
+    LTEXT           """FRED2 is the omega of all giant unwieldy pieces of code. It's big, it's horrifying, and it just doesn't care. View it at your own risk."" - Dave Baranec",IDC_STATIC,39,62,238,20
     LTEXT           "Ported to OGL by RT",IDC_STATIC,39,40,68,8
     LTEXT           "Bravely maintained by Goober5000 and WMCoolmon",IDC_STATIC,39,32,181,8
```
jg18, line 675: Given that WMC hasn't been around in years, I was wondering if his name should still be there, but perhaps he contributed enough to merit having his name kept there in perpetuity? :man_shrugging:

Goober5000: He definitely contributed a lot over several years.  I would keep his name there somehow, but the dialog doesn't have to imply that he's currently maintaining it.

Goober5000 (commented):
Looks good, but I'll wait for a response about the About line.

fred2/fred.rc
jg18: One idea: below the "Bravely maintained" line, we could add a line "With major contributions from" and add WMC there.

Who else should be included in that list?

To get a list of potential additional contributors, I collected `git blame` output from the `fred2/` source files and generated a line count. It's obviously a *very* rough approximation, especially since it doesn't account for one person's lines being deleted/overwritten by others, and line counts aren't really a great way to count contribution quantity besides. But it's a starting point for finding names.

I'll let you determine whose work qualifies as "major contributions".

```
  45507 goober5000
   7424 z64555
   2523 karajorma
   2517 the_e
   1284 jg18
    955 taylor
    841 asarium
    582 zacam
    312 wmcoolmon
    244 portej05
    154 mageking17
    126 plieblang
    107 issmneur
     71 bmagnu
     65 kazan
     55 valathil
     55 sushi
     51 chief1983
     41 wanderer
     41 echelon9
     40 niffiwan
     32 yarn366
     29 fscyborg
     22 baezon
     17 wookieejedi
     16 asteroth
     11 phreak
     10 the
     10 johnafernandez
      9 downwash
      5 fubar-bdhr
      4 swifty
      2 turey
      2 trblount
      2 samuelcho
      2 artyrm
      1 shivansps
      1 flaming_sword
```

jg18: For reference, I used these commands in Git Bash to generate the above list:
```
$ cd fred2
$ for i in `ls -1 | egrep "\.(h|cpp)$"`; do git blame $i; done > blame.txt
$ cat blame.txt | cut -d '(' -f 2 | cut -d ' ' -f 1 | tr '[:upper:]' '[:lower:]' | sed 's/fabian/the_e/g' | sed 's/m_m/asarium/g' | sed 's/joshua/jg18/g' | sort | uniq -c | sort -r
```
The non-exhaustive list of substitutions was to improve readability and get more accurate counts, with all of the aforementioned disclaimers about the dubious value of line counts.

Goober5000: Wow, that is an eye-opening list!  And incidentally, RT = RandomTiger; I'm surprised to not see his name on there.  Then again, his contributions came before the FRED code was moved to the `fred2` folder, so maybe the search isn't picking them up.  That could be affecting number from other people as well.

Given that my number is over 6x the closest runner up, maybe this would work

```
Ported to OpenGL by RandomTiger
Maintenance and major contributions by Goober5000
```

Is there a similar dialog for qtFRED?

jg18: I was thinking of the same changes to the RT line. 🙂 

Even though several people have made major contributions to FRED over the years, keeping the dialog short is important IMO, so I went with something simple.

No idea about QtFRED.

jg18: Whoops just noticed the last line gets cut off. Will fix.

jg18: Ok fixed now.

Goober5000: I like your changes, but to tie up loose ends, I searched for other dialogs.  The qtFRED dialog is in `qtfred\ui\AboutDialog.ui`, while the wxFRED dialog (which probably doesn't need any updates) is in `wxfred2\base\wxFRED_base.xrc` and `wxfred2\base\wxFRED_base.cpp`.

jg18: Ok thanks, will take a look at updating QtFRED.

I learned from updating the FRED dialog not to just edit text but make sure the result looks ok. I think I have a copy of Qt Creator lying around somewhere.

Goober5000 (approved):

PR #3943, opened 2022-01-20
TBP warp fix

### LINKED ISSUE #3917: No 3D Departure Warp for TBP
I am unsure if this is a coding problem, or if the mechanics of the feature have changed and TBP is lagging behind.  In TBP, subspace warps (or 'Hyperspace jump points') use a different physical shape than FS does, in the form of an elongated cone.  Furthermore, when a ship arrives in the game area, it uses a blue effect, and an orange one when departing.  In the fireball.tbl, the "Jump in" is used in place of the standard warp, while the "knossos" space is used as the "jump out".  The Jump In effect uses the elongated cone for jumps, but for some reason the Jump Out effect doesn't.  It uses this wide, shallow cone shape for warps.  I have attempted to manually add the warp model line into the fireball table, and I've used the custom warpin/warpout parameters area in departure/arrival cues of a ship and have confirmed that the jump out or knossos effect will only use this default, wide-cone shape for exit warps.  This extends to sexp-generated exit warps as well.

Goober5000:
Set the flag for the Knossos warp effect being used when the effect is part of the normal warpin/warpout without involving the Knossos device.

Change the flag from `int` to `bool` and tweak a few things.

Also make sure the warp model is used if it is explicitly specified in the fireball tbl.

Fixes #3917.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -375,7 +375,7 @@ int Cmdline_rearm_timer = 0;
 int Cmdline_targetinfo = 0;
 
 // Gameplay related
-cmdline_parm use_3dwarp("-3dwarp", NULL, AT_NONE);			// Cmdline_3dwarp
+cmdline_parm use_3dwarp("-3dwarp", NULL, AT_NONE);			// Is now Fireball_use_3d_warp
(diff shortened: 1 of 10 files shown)


wookieejedi (approved):
LGTM

PR #3944, opened 2022-01-20
fix training mission stats bug

### LINKED ISSUE #3916: crash on training mission replay
I'm having trouble finding the clean repro for this, and regrettably I failed to properly capture the stack trace or dump file, but here's what I know.
The bug was found on the branch for #3914.
It happened after I completed the first tutorial mission in the mvps, and then hit replay mission from the briefing
It tripped the `UNREACHABLE("Ship kills of '%s' not found, should have been added by pilotfile::update_stats.", Ship_info[i].name);` in `pilotfile.cpp` line 293, with the ship being the amazon.

notimaginative:
Training missions don't save kills so make sure that we don't try to back out kills by mistake when replaying a training mission.

Fixes #3916

### DIFF
--- a/code/missionui/missiondebrief.cpp
+++ b/code/missionui/missiondebrief.cpp
@@ -2037,13 +2037,13 @@ void debrief_close()
 						scoring_backout_accept(sc);
 
 						if (Net_player == &Net_players[i]) {
-							Pilot.update_stats_backout( sc );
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
Good find!

Goober5000 (approved):
Seconded

PR #3945, opened 2022-01-21
accommodate elapsed time during debugging

Goober5000:
If a breakpoint is hit in the code, the SDL time will keep running even though FSO doesn't realize things are effectively, if not actually, paused.  Detect this situation and adjust the timestamp offset to compensate.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -4284,7 +4284,7 @@ void game_set_frametime(int state)
 		mprintf(("Frame %2i too long!!: frametime = %.3f (%.3f)\n", Framecount, f2fl(Frametime), f2fl(debug_frametime)));
 
 		// If the frame took more than 5 seconds, assume we're tracing through a debugger.  If timestamps are running, correct the elapsed time.
-		if (!timestamp_is_paused() && (Last_frame_timestamp != 0) && (f2fl(Frametime) > 5.0f)) {
+		if (!Cmdline_slow_frames_ok && !timestamp_is_paused() && (Last_frame_timestamp != 0) && (f2fl(Frametime) > 5.0f)) {
 			auto delta_timestamp = timestamp() - Last_frame_timestamp;
 			mprintf(("Adjusting timestamp by %2i milliseconds to compensate\n", delta_timestamp));
 			timestamp_adjust_pause_offset(delta_timestamp);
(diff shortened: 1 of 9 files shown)


wookieejedi (approved):
LGTM

notimaginative:
Looks okay to me, but this is triggering some nightmare scenarios in my brain. Not an issue with this code but with similar instances of the problem. Let me mull it over for a few hours if you don't mind and I'll update if I verify any other situations that could be rolled into this fix.

Goober5000:
@notimaginative Certainly.  Let me know if you think of anything.

freespace2/freespace.cpp
```diff
@@ -4281,6 +4281,13 @@ void game_set_frametime(int state)
 	if (Frametime > MAX_FRAMETIME)	{
 #ifndef NDEBUG
 		mprintf(("Frame %2i too long!!: frametime = %.3f (%.3f)\n", Framecount, f2fl(Frametime), f2fl(debug_frametime)));
+
+		// If the frame took more than 5 seconds, assume we're tracing through a debugger.  If timestamps are running, correct the elapsed time.
+		if (!timestamp_is_paused() && f2fl(Frametime) > 5.0f) {
+			auto delta_timestamp = timestamp() - Last_frame_timestamp;
```
notimaginative, line 4287: One thing I noticed is that `Last_frame_timestamp` isn't initialized when it's declared, so technically it's not set until after it's used here. If there is a long enough startup time such that this is hit on the first call then it could have unpredictable results.

Most people probably won't have to worry about that but profiling tools can slow things down enough to cause a problem.

notimaginative (changes requested):

notimaginative:
My worries were mostly over things like valrgrind, long level loads (particularly on standalone), and certain multi states. The long level loads thing is complicated by the bug that was pointed out in the forums so I'm going to come back to that once there is a fix. I have been unable to find any notable issue with the multi states but I'm going to make a note of it so we can be on the lookout, just in case.

Valgrind and other profiling tools are the bigger problem since they can slow the game down enough to trigger this *very* often and I'm not sure if that's a good thing or a bad thing.

freespace2/freespace.cpp
Goober5000: Good catch.  I initialized it.

Goober5000:
I pushed a fix for `Last_frame_timestamp`.

I'm also not sure whether adjusting the timestamp for profiling is good or bad, but what I could do is add a command-line option that will simply disable the timestamp adjustment.

notimaginative (approved):

PR #3947, opened 2022-01-22
Fix/issue 3902

z64555:
Addresses issues discussed in issue #3902 

* Writes default controls in pre-Controls5 format for the .json to prevent a CTD in non-Controls5 versions when trying to load Controls5 pilot
* Displays a second nag screen with a wall of text to explain that the controls will be defaulted when attempting to use a Controls5 pilot in a non-Controls5 build

### DIFF
--- a/code/pilotfile/plr.cpp
+++ b/code/pilotfile/plr.cpp
@@ -770,6 +770,9 @@ void pilotfile::plr_write_controls()
 
 		} else if (mouse >= 0) {
 			handler->writeInt("axis_map", mouse);
+
+		} else {
+			handler->writeInt("axis_map", -1);
 		}
 		
 		handler->writeInt("invert_axis", item.is_inverted() ? 1 : 0);
--- a/code/pilotfile/plr.cpp
+++ b/code/pilotfile/plr.cpp
@@ -740,7 +740,46 @@ void pilotfile::plr_read_controls()
 void pilotfile::plr_write_controls()
 {
 	handler->startSectionWrite(Section::Controls);
-	
+
+	// Save the default bindings to prevent crash. See github issue 3902
+	auto& Control_config_default = Control_config_presets[0].bindings;
+	handler->startArrayWrite("controls", JOY_AXIS_BEGIN, true);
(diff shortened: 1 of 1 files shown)


code/pilotfile/plr.cpp
```diff
@@ -728,9 +737,51 @@ void pilotfile::plr_read_controls()
 void pilotfile::plr_write_controls()
 {
 	handler->startSectionWrite(Section::Controls);
-	
+
+	// Save the default bindings to prevent crash. See github issue 3902
+	auto& Control_config_default = Control_config_presets[0].bindings;
```
notimaginative, line 742: Couldn't this just be the current preset instead, since it's referenced just below here already? That would allow for a basically consistent control setup between the builds. And it would also make that "technical difficulties" info box pretty much irrelevant.

notimaginative (changes requested):

code/pilotfile/plr.cpp
z64555: No. It cannot be the current preset.  This is because the new binding scheme allows for multiple joysticks and the mouse to be bound to the same control, and the old binding scheme can not achieve that.  When the compatibility code goes to convert the post-Controls5 bindings into pre-Controls5 bindings, it loses the multiple joysticks (and mouse) bindings making it significantly different from the preset it was based on.  Thus, when the compatibility portion of the pilot is re-imported, the plr conversion code sees it as a completely new preset and essentially duplicates the bindings.

notimaginative: Ah, right.

notimaginative:
Things are still broken, but that's not related to the controls changes. It goes back to the lousy error handling in the pilot file code and I'll bring that new issue up on discord. Just made it annoying to test this fix.

But this specific change does appear to work as expected jumping between 21.4 and 21.5/22 builds. No errors are generated and no crashes.

notimaginative (approved):
Works as expected.

wookieejedi:
Per the Discord discussion, sounds like this is good to merge?

PR #3948, opened 2022-01-22
fix arrival delays and other mission time wonkiness

### LINKED ISSUE #3946: Arrival Delays are Not Working Correctly
As first noticed by Anatares, mission time starts ticking while in the briefing, which affects in-mission times. For example, If I set the delay of a ship to be 5 seconds in, and wait around it the briefing for at least that long the ship is arrived when the mission starts. Introduced by #3879.

Goober5000:
The `Timestamp_microseconds_at_mission_start` variable should not, in fact, be adjusted when the timer counter variables are adjusted.

Fixes #3946.


notimaginative (approved):
LGTM and appears to fix the Hope mission. A few other test missions that I didn't even notice had issues are now fixed as well.

PR #3949, opened 2022-01-23
add IP version preference troubleshooting options

notimaginative:
Add cmdline options to help deal with connectivity issues over IPv4/IPv6. This allows the user to set a preference for either IPv4 or IPv6 which will be used for DNS lookups. Both IPv4 and IPv6 still work normally however, allowing personal connection overrides and preventing a broken psnet setup.

Also now allows PXO data validation for mod ident to fall back to a different IP version if the first attempt times out.


wookieejedi (approved):
Good idea. Looks straightforward and clear.

PR #3950, opened 2022-01-23
some small timestamp fixes

Goober5000:
A few assorted fixes for timestamps:

1) The pre-player delay should be fast-forwarded by 1/4 second per frame regardless of time compression
2) If time compression is very slow, there could be a 0-millisecond change in timestamps, so don't adjust this
3) Properly handle a change in soundtrack that occurs during a mission, not just when the mission is parsed


wookieejedi (approved):
Looks good and works in my tests.

PR #3953, opened 2022-01-23
some small fixes

### LINKED ISSUE #3951: set-unscanned won't accept multiple ships
set-unscanned suggests in the sexp help that it can accept any number of ships.

set-unscanned
	Sets the cargo on the specified ship or ship subsystem as unknown or unscanned.  Takes 1 or more arguments...
	1:	Name of a ship
	Rest:	Name of a subsystem on that ship (optional)
	
However, after the first argument it only accepts \<Any Data\> and after the second argument you can no longer click Add Data.

Either the sexp help needs to be clarified or the sexp itself fixed.

Goober5000:
1) Fix set-scanned and set-unscanned to take multiple arguments as designed.
2) Fix documentation on the "perturb" scripting functions.

Fixes #3951


wookieejedi (approved):
Nice small fix.

PR #3954, opened 2022-01-23
Treat Mouse as Joy0 if "turned on" in options menu

z64555:
Old Mouse Option Group:
image

New Mouse Option Group:
options_mouse

This control now, instead of simply doing nothing within the Controls5 version, toggles the mouse inputs between "Mouse" and "Joy-0."

When On, the mouse is treated as Joy0 and thus its axes and buttons are seen as coming from Joy-0.  This includes binding in the config menu as well as during game use, so some confusion may occur when a user is trying to bind their mouse and the button shows up as a Joy-0.  Should there be a physical joystick in addition to the mouse, the joystick inputs should play nice.

When Off, the mouse is treated as the mouse and is independent from Joy0 completely, and axes and buttons are seen as coming from the Mouse controller instead of Joy0.

This PR adds the following two XSTR indices:
1662 - "Mouse Input Mode"
1663 - "Joy-0"

Things to do:

- [x] Investigate whats needed to update the label on the Options menu to something more descriptive
- [x] Update the Information() string for plr conversion to inform the changes.


z64555:
> ../code/controlconfig/controlsconfig.cpp:2696:3: error: this statement may fall through [-Werror=implicit-fallthrough=]
 2696 |   if (Use_mouse_to_fly) {
      |   ^~
../code/controlconfig/controlsconfig.cpp:2704:2: note: here
 2704 |  case CID_JOY1:
      |  ^~~~

Fallthrough is intentional here, but not sure how to //NOLINT it proper

wookieejedi:
@BMagnu IIRC you had a fix for the previous //NOLINT, happen to know the fix here too?

Goober5000 (commented):
Looks ok to me.  However I think I should hold off on explicit approval until the two "things to do" checkboxes are checked.

z64555:
"Update the Information() string for plr conversion to inform the changes."

Since #4005 does the info popup differently, this sub-feature should depend on it.

z64555:
XSTR indices will need to be updated as soon as #3988 goes through, since that PR also adds XSTR's

Goober5000 (approved):

z64555:
XSTR indices updated. Enabled auto-merge.

z64555:
Disabling auto merge because I forgot to add shout-out in the popup info. 🙅

PR #3960, opened 2022-01-25
add scripting support for decals

Goober5000:
This extends the existing decal system so that decals can be created through scripting.

~~Testing reveals there may be some unexpected problems preventing decals from working.  This is in draft status until this is resolved.~~  Now that #4053 has been merged, this PR has been tested and is out of draft status.

### DIFF
--- a/code/decals/decals.h
+++ b/code/decals/decals.h
@@ -8,6 +8,52 @@
 
 namespace decals {
 
+class DecalDefinition {
+	SCP_string _name;
+
+	SCP_string _diffuseFilename;
+	bool _loopDiffuse = true;
+
+	SCP_string _glowFilename;
+	bool _loopGlow = true;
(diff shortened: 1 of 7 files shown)


code/decals/decals.h
```diff
@@ -8,6 +8,96 @@
...
+
+	SCP_string _normalMapFilename;
+	bool _loopNormal = true;
+
+	int _diffuseBitmap = -1;
+	int _glowBitmap = -1;
+	int _normalBitmap = -1;
+
+ public:
+	explicit DecalDefinition(SCP_string name) : _name(std::move(name)) {
```
asarium, line 28: Please move all the function definitions into the source file to avoid compilation overhead.

asarium (changes requested):

PR #3961, opened 2022-01-25
Add view show status for HUD gauges in scripting

wookieejedi:
Many FSO mods use scripted HUD gauges, but previously the scripted HUD gauges were always forced to be on for the player, unlike retail gauges that can be toggled in the HUD Config screen. This PR allows scripters to view the show status of a HUD gauge, which is very useful for adding a check in the script to only show the gauge if a certain gauge is active. This is similar behavior to the getColor function, which allowed scripters to get the HUD Config colors of a given default gauge.

Tested and works as expected.

### DIFF
--- a/code/scripting/api/libs/hud.cpp
+++ b/code/scripting/api/libs/hud.cpp
@@ -60,6 +60,22 @@ ADE_VIRTVAR(HUDDefaultGaugeCount, l_HUD, "number", "Specifies the amount of HUD 
 	return ade_set_args(L, "i", amount);
 }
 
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
Short and sweet.  Looks good.

Before I approve this, have you considered making this a VIRTVAR?  That would allow scripts to not only read the show-status, but set it too.

wookieejedi:
I did play around with it as a VIRVAR, though in evaluating it was much more involved/troublesome because of how the pilot file is used with setting HUD config values. As a result, being able to get the status would cover the most common use cases for custom HUD gauges while also being the most minimally invasive.

Goober5000:
Ok, fair enough.

Goober5000 (approved):

PR #3962, opened 2022-01-25
Inverse Kinematics for Animations

BMagnu:
This PR introduces an animation segment type used to combine multiple submodels into a chain which can be controlled by inverse kinematics.
Allows for each joint to have a specified rotation constraint, as well as to target a specific orientation with the end effector.
Implements the FABRIK algorithm with additions for constraints and end effector rotation.
The hinge constraint _could_ cause certain cases where FABRIK doesn't properly converge, but these'd need to be investigated in the wild when found.

As a side note: Apart from the outstanding translation module by Goober (for which animation support already exists), this is the final thing on my ToDo for animations. So unless someone has more feature requests, the overhaul can be considered basically complete with this PR.

### DIFF
--- a/code/math/ik_solver.cpp
+++ b/code/math/ik_solver.cpp
@@ -22,6 +22,9 @@ void ik_solver_fabrik::solve(const vec3d& targetPos, const matrix* targetOrient)
 	while(iterationCounter++ < m_maxIterations && endEffectorDistance > m_targetDistance && endEffectorDistanceDelta > m_minProgress) {
 		endEffectorDistanceDelta = endEffectorDistance;
 
+		for(auto& node : m_nodes)
+			node.lastPos = node.calculatedPos;
+		
 		m_nodes.back().calculatedPos = targetPos;
 		
 		//Backwards pass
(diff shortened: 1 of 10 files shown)


BMagnu:
Converted to draft to quickly add moveable support

Baezon (approved):
Looks like a pretty good implementation of the FABRIK algorithm to me. FABRIK is mentioned, but should there perhaps be a link to the paper?

code/math/ik_solver.cpp
```diff
@@ -0,0 +1,218 @@
...
+				// Convert new local rotation and global rotation of tha parent's parent to new parent global rotation
+				vm_matrix_x_matrix(&parent.calculatedRot, &localRot, &parentRot);
+				
+				// Find new child position from new parent global rotation, child offset and parent position
+				vm_vec_unrotate(&child.calculatedPos, &child.submodel->offset, &parent.calculatedRot);
+				child.calculatedPos += parent.calculatedPos;
+			}
+		}
+		
+		//After positioning the end effector, it now needs to be oriented correctly. If we have a target orientation use it, otherwise assume it's parents rotation. Then constrain.
```
Goober5000, line 110: it's -> its

Goober5000 (commented):
That's about it

Goober5000 (approved):

PR #3970, opened 2022-01-28
replicate submodels that aren't associated with subsystems

### LINKED ISSUE #3967: Bug: Submodel rotation LOD updating only occurs on subsystems
The submodel update function needs to be applied to all submodels, not just subsystems.
This way, the new animation system can properly work across LODs

Goober5000:
There are a few related improvements here:

1) Add a new function for iterating over the submodel hierarchy
2) Add a counterpart to the `subobj_num` field so that a submodel can refer to its ship_info subsystem
3) Update `ship_model_replicate_submodels` to replicate all movable submodels, even the ones that aren't associated with subsystems

Fixes #3967

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -13531,7 +13531,6 @@ void ship_model_replicate_submodels(object *objp)
 {
 	model_subsystem	*psub;
 	ship		*shipp;
-	ship_info	*sip;
 	ship_subsys	*pss;
 
 	flagset<Ship::Subsystem_Flags> empty;
@@ -13541,7 +13540,6 @@ void ship_model_replicate_submodels(object *objp)
 	Assert(objp->type == OBJ_SHIP);
 
 	shipp = &Ships[objp->instance];
-	sip = &Ship_info[shipp->ship_info_index];
 
 	polymodel_instance *pmi = model_get_instance(shipp->model_instance_num);
 	polymodel *pm = model_get(pmi->model_num);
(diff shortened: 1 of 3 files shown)


code/ship/ship.cpp
```diff
@@ -13552,6 +13559,24 @@ void ship_model_replicate_submodels(object *objp)
...
+			auto sm = &pm->submodel[submodel];
+
+			// skip submodels that belong to subsystems since we already updated them above
+			// (Turrets are tricky because they have two submodels per subsystem, but because the subsystem index is copied
+			// when the model is read, both submodels are correctly skipped here.)
+			if (sm->subsys_num >= 0)
+				return;
+
+			// the only non-subsystem submodels that need to be updated are the ones that can move
+			if (!sm->flags[Model::Submodel_flags::Can_move])
```
BMagnu, line 13575: This assumption is not true anymore.
The $triggered property is no longer required due to the way that animations step over time.
Hence, an animated submodel might not have this flag set.

Goober5000: That flag needs to be set though.  Rotating subsystems set that flag, and so will translating subsystems once that's finished.  Plenty of places in the code use that flag to determine whether they should treat submodels as static or moveable: dockpoints, thrusters, etc.

In fact I just realized the collision code needs to be fixed to use it too; it currently uses an alternate method that won't capture all situations.

Goober5000: Added the flag to the collision code.

BMagnu: #3972 addresses the issue that not all animated subsystems can have Can_move

BMagnu (approved):

PR #3972, opened 2022-01-28
Various Animation Fixes

BMagnu:
1. Some turrets had wrong initial orientations in mission only. This is fixed by removing the errors from ``vm_closest_angle_to_matrix``.
2. Now sets the Can_move flag for all submodels that might be animated by an animation
3. Properly set canonical_prev_orientation for collisions and others
4. Fix moveables flickering when updating

### DIFF
--- a/code/model/modelanimation.cpp
+++ b/code/model/modelanimation.cpp
@@ -321,7 +321,10 @@ namespace animation {
 		if (!submodel.first || !submodel.second) 
 			return;
 
-		submodel.second->flags.set(Model::Submodel_flags::Can_move);
+		if(!submodel.second->flags[Model::Submodel_flags::Can_move]){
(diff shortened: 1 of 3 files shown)


code/model/modelanimation.cpp
```diff
@@ -321,6 +321,8 @@ namespace animation {
 		if (!submodel.first || !submodel.second) 
 			return;
 
+		submodel.second->flags.set(Model::Submodel_flags::Can_move);
```
Goober5000, line 324: I recommend adding a logging statement, if the submodel didn't previously have this flag, indicating that the flag has been added.  Ideally, models should not be modified after being read, but this is a valid and appropriate exception.

Goober5000 (changes requested):

Goober5000 (approved):

PR #3973, opened 2022-01-28
update mdns to v1.4.3

notimaginative:


notimaginative:
This was a breaking API change which requires more code on our side to work now. However it does address multiple stability/security/compliance issues in the lib.

Leaving as a draft pending any further changes and additional testing.

PR #3976, opened 2022-01-29
Improve Animation SEXP usability

BMagnu:
This adds better description to the new SEXPs, adds proper error checking for the entered animation triggers, and shows the FREDer the available triggers / names.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -35933,7 +35933,8 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"Takes 3 or more arguments...\r\n"
 		"\t1: The ship to trigger the animation on.\r\n"
 		"\t2: The trigger type of the animation.\r\n"
-		"\t3: The triggered-by value of the animation. Depends on trigger type. Refer to the wiki on *-anim.tbm's.\r\n"
+		"\t3: The triggered-by value of the animation. Must be the same as in the table for the animation. Leave blank if not specified in the table.\r\n"
+		"\t\tException: fighterbay-type animations must specify the number of the fighter bay path.\r\n"
 		"\t4: If the animation should play forwards (true) or backwards (false). Defaults to true.\r\n"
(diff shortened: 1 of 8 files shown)


code/parse/sexp.cpp
```diff
@@ -35835,7 +35933,8 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"Takes 3 or more arguments...\r\n"
 		"\t1: The ship to trigger the animation on.\r\n"
 		"\t2: The trigger type of the animation.\r\n"
-		"\t3: The triggered-by value of the animation. Depends on trigger type. Refer to the wiki on *-anim.tbm's.\r\n"
+		"\t3: The triggered-by value of the animation. Must be the same as in the table for the animation. Leave blank if not specified in the table.\r\n"
```
Goober5000, line 35936: Does the syntax checking correctly handle the blank string?

Goober5000 (commented):
Looks good, but one question.

BMagnu:
In this case yes, cause if they don't have that specified, they'll have a default subtype. This causes them to get validated by trigger string which is then empty, thus allowing empty animation if such an animation exists.On 30 Jan 2022 1:13 am, Goober5000 ***@***.***> wrote:
@Goober5000 commented on this pull request.

Looks good, but one question.

In code/parse/sexp.cpp:
> @@ -35835,7 +35933,8 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"Takes 3 or more arguments...\r\n"
 		"\t1: The ship to trigger the animation on.\r\n"
 		"\t2: The trigger type of the animation.\r\n"
-		"\t3: The triggered-by value of the animation. Depends on trigger type. Refer to the wiki on *-anim.tbm's.\r\n"
+		"\t3: The triggered-by value of the animation. Must be the same as in the table for the animation. Leave blank if not specified in the table.\r\n"

Does the syntax checking correctly handle the blank string?

Goober5000 (approved):
Ok sounds good.

PR #3978, opened 2022-01-29
handle primary and secondary inversion explicitly and independently

Goober5000:
Treating both bindings together will fail when a player has one binding inverted but the other uninverted.  This was complicated by the dual-binding method mysteriously delegating to the single-binding method in certain cases.  This change removes the dual-binding methods altogether and explicitly handles the bindings individually.  It also adds additional scripting VIRTVARs for all current use cases.

### DIFF
--- a/code/pilotfile/plr.cpp
+++ b/code/pilotfile/plr.cpp
@@ -685,7 +685,8 @@ void pilotfile::plr_read_controls()
 
 			if (idx < NUM_JOY_AXIS_ACTIONS) {
 				Control_config[idx + JOY_AXIS_BEGIN].take(CC_bind(CID_JOY0, static_cast<short>(axi), CCF_AXIS), 0);
-				Control_config[idx + JOY_AXIS_BEGIN].invert(inv != 0);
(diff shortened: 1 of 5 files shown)


code/pilotfile/plr.cpp
```diff
@@ -772,7 +773,7 @@ void pilotfile::plr_write_controls()
 			handler->writeInt("axis_map", -1);
 		}
 		
-		handler->writeInt("invert_axis", item.is_inverted() ? 1 : 0);
+		handler->writeInt("invert_axis", (item.first.is_inverted() || item.second.is_inverted()) ? 1 : 0);
```
Goober5000, line 776: Yes, I changed the logic from `&&` to `||` here, as I think this better handles the case of saving the old data when the player has two different inversions.

z64555:
LGTM

Does this still support independent inversion of primary and secondary bindings or no?

Goober5000:
> Does this still support independent inversion of primary and secondary bindings or no?

Yes.  The only things I changed had to do with the few cases where they were handled as a pair.

z64555 (approved):

PR #3979, opened 2022-01-29
refactoring in preparation for model translation

Goober5000:
Mostly renaming of fields and functions, and moving some modelread sections into functions.  This reorganization will put the code on a better footing for the model translation feature.  No model translation code is included yet, and there should be no change in functionality.

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -53,7 +53,7 @@ flag_def_list model_render_flags[] =
   	 
 int model_render_flags_size = sizeof(model_render_flags)/sizeof(flag_def_list);
 
-#define MAX_SUBMODEL_COLLISION_ANGLE	(PI / 6.0f)		// max 30 degrees per frame
+#define MAX_SUBMODEL_COLLISION_ANGULAR_VELOCITY		(PI / 6.0f)		// max 30 degrees per frame
 
 // info for special polygon lists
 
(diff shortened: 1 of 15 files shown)


code/model/modelread.cpp
```diff
@@ -53,7 +53,7 @@ flag_def_list model_render_flags[] =
   	 
 int model_render_flags_size = sizeof(model_render_flags)/sizeof(flag_def_list);
 
-#define MAX_SUBMODEL_COLLISION_ROT_ANGLE (PI / 6.0f)	// max 30 degrees per frame
+#define MAX_SUBMODEL_COLLISION_ANGLE	(PI / 6.0f)		// max 30 degrees per frame
```
BMagnu, line 56: Why did you rename this as well, since it was specific to rotation, not any generic movement?
Tbh, I'd expect this to be called something like ``MAX_SUBMODEL_COLLISION_ROT_VELOCITY``, as it's specific about rotational velocities

code/model/modelread.cpp
```diff
@@ -1138,15 +1138,245 @@ void model_calc_bound_box( vec3d *box, vec3d *big_mn, vec3d *big_mx)
...
+		movement_axis = &sm->rotation_axis;
+		movement_type = &sm->rotation_type;
+	}
+	else
+	{
+		UNREACHABLE("Not yet implemented");
+		movement_axis_id = nullptr;
+		movement_axis = nullptr;
+		movement_type = nullptr;
+	}
```
BMagnu, line 1313: You do this chunk of code sufficiently often, that I'd say it might be wise to convert it into a function that takes three reference-to-pointers

BMagnu (changes requested):
Seems like it basically just makes renaming clarifications on movement in general vs rotation as a specific form of movement, as well as collating some movement setup functions.
Small comments.

code/model/modelread.cpp
Goober5000: I didn't like the `ROT` abbreviation, but I'm open to other terms.  I like your suggestion, so I'll use that with a small change.

code/model/modelread.cpp
Goober5000: Hmm...

BMagnu (approved):

PR #3981, opened 2022-01-30
Fix Animation edge case for turret lookup with template and display name

BMagnu:
When a ship is initialized from a template _and_ has a display name, previously the ship class name was not properly copied to the animation system for multi-hashing, turret lookup, and so far.
This remedies this _and_ adds robustness and performance by using the new pmi objnum to find the shipclass for turret lookup instead of a string search


Goober5000 (approved):
Good fix. 👍

PR #3987, opened 2022-01-30
Fix Damage Gauge Display

### LINKED ISSUE #3982: Bug: Damage Display Gauge does not display specific subsystem health
The HUD damage gauge iterates through all subsystems, but only shows the aggregate value for that subsystem type (shown below). This is very strange with ships with more than 1 subsystem of a given type (for example 2 engines). It also technically presents the player with incorrect information. 

unknown

wookieejedi:
Previously, the damage display gauge only showed the integrity for subsystem types. This was fine if there were a 1:1 match, with each type only having 1 subsystem. If there were more than one subsystem of a given type though, it would not display accurate information to the player.

Fixed Version
Working

This PR sets the displayed subsystem integrity to that actual specific subsystem, not the aggregate. Tested and works as expected (screenshot attached) and fixes #3982.

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -2056,7 +2056,14 @@ void HudGaugeDamage::render(float  /*frametime*/)
 		psub = pss->system_info;
 		strength = ship_get_subsystem_strength(Player_ship, psub->type);
 		if ( strength < 1 ) {
-			screen_integrity = fl2i(strength*100);
+			// display the actual health of this specific subsystem --wookieejedi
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Looks good.  I noticed that a few lines down, the subsystem flash calculation has the same problem, but fixing that would be considerably more complicated - and a fix is not as clearly necessary as it is for this issue.

wookieejedi:
Yep, I noticed that as well, and had the same thought.

PR #3989, opened 2022-01-31
point camera correctly when using set-camera-facing

### LINKED ISSUE #3889: set-camera-facing-object ignores camera host rotation
[setcamfacingobjecttest.zip](https://github.com/scp-fs2open/fs2open.github.com/files/7804810/setcamfacingobjecttest.zip)

When a camera host is set, set-camera-facing-object always asumes the host rotation is 0 on all axis, and it works as long as it remains as such. But as soon the host rotation is anything else than 0, set-camera-facing-object fails to face the object. It would still points at the object as if the host rotation were still 0.

This is also true for set-camera-facing.

Goober5000:
A few fixes for the camera code...

1) When pointing the camera toward a target, take into account the host's orientation.  Use a game_settings.tbl flag to control this.
2) When using a submodel to host the camera, correctly calculate the submodel orientation.

Fixes #3889

### DIFF
--- a/code/camera/camera.cpp
+++ b/code/camera/camera.cpp
@@ -378,32 +378,32 @@ void camera::get_info(vec3d *position, matrix *orientation)
 		{
 			if(object_target.IsValid())
 			{
-				object *objp = object_target.objp;
-				int model_num = object_get_model(objp);
-				polymodel *pm = nullptr;
-				polymodel_instance *pmi = nullptr;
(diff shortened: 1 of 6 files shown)


Baezon (changes requested):
Looks good to me, I might be misunderstanding the issue, but as I understand it this behavior is more intuitive and should probably target new versions.

Baezon (approved):

PR #3991, opened 2022-01-31
Fix edge case rollback crash

JohnAFernandez:
So if your ping somehow gets up to ~500ms, and you try to shoot a primary shot or dumbfire missile, your shot might trigger an assert, because the code was previously letting in shots that were a little bit too old. So make sure that if the server is about to enter that frame index that the shot would have been mapped to, reject the shot.


JohnAFernandez:
Also, some whitespace changes that make the code a little clearer.

Goober5000 (approved):

PR #3992, opened 2022-01-31
Fix limitations of GetColor script function

wookieejedi:
It is very valuable for custom HUD gauges to be able to show/use the IFF color of any given object. Previous scripting solutions were limited and did not account for traitor status, varying team combinations (i.e. TvT multiplayer), and other cases.

This PR allows scripters to get the IFF color of a given object properly, as it uses the same method that retail HUD gauges use. It also allows HUD scripts to be updated to properly work in the aforementioned cases. For example, this will fix FotG's custom HUD gauges in multiplayer in TvT and Dogfight missions. 

Also, it turns out there was already a function `hud_get_iff_color` defined in `hud.h`. It was never used though, as `hud_set_iff_color` both got the color then set it. I've added the color getting to the `get` function and the setting to the `set` function, that way the script function can easily just call the `hud_get_iff_color` function. If this is undesirable, I can instead copy the entire logic of the function `hud_set_iff_color` into the script section.

Tested and works as expected.

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -3049,7 +3049,19 @@ void hud_set_dim_color()
  * @param objp			Object to test for team color to base on
  * @param is_bright		Default parameter (value 0) which uses bright version of IFF color
  */
-void hud_set_iff_color(object *objp, int is_bright)
+void hud_set_iff_color(object* objp, int is_bright) {
+
+	gr_set_color_fast(hud_get_iff_color(objp, is_bright));
+
+}
+
+/**
+ * @brief Will get the color to the IFF color based on the team
+ *
+ * @param objp			Object to test for team color to base on
+ * @param is_bright		Default parameter (value 0) which uses bright version of IFF color
+ */
+color* hud_get_iff_color(object* objp, int is_bright)
 {
 	color *use_color;
 
@@ -3076,7 +3088,7 @@ void hud_set_iff_color(object *objp, int is_bright)
 		}
 	}
 
-	gr_set_color_fast(use_color);
+	return use_color;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


code/scripting/api/objs/object.cpp
```diff
@@ -594,5 +594,29 @@ ADE_FUNC(removeSound, l_Object, "soundentry GameSnd, [subsystem Subsys=nil]",
...
+	"number, number, number", 
+	"IFF rgb color of the object or nil if object invalid")
+{
+	object_h* objh;
+
+	if (!ade_get_args(L, "o", l_Object.GetPtr(&objh)))
+		return ade_set_error(L, "iii", ADE_RETURN_NIL, ADE_RETURN_NIL, ADE_RETURN_NIL);
+
+	if (!objh->IsValid())
+		return ade_set_error(L, "iii", ADE_RETURN_NIL, ADE_RETURN_NIL, ADE_RETURN_NIL);
```
Goober5000, line 609: Just return `ADE_RETURN_NIL` here.  Lua will apply it to any number of arguments the caller tries to get.  The way you have it here will return a valid triplet of `0, 0, 0`.

Goober5000 (changes requested):

code/scripting/api/objs/object.cpp
wookieejedi: Good to know, thanks! Just updated.

Goober5000 (approved):

code/hud/hud.cpp
```diff
@@ -3049,7 +3049,19 @@ void hud_set_dim_color()
  * @param objp			Object to test for team color to base on
```
JohnAFernandez, line 3049: Does this documentation still make sense?

wookieejedi: I had thought about that, too. The object affects what the team color will be set to, so I left it as is. The alternative I had tried was `Object to test for team color to set`, but that didn't really sound right, as it's not setting a team color for the object.

wookieejedi: Updated with discussion from Discord to clarify comments.

PR #3994, opened 2022-01-31
Coverity 1361855 Always free Vertices and Indices

JohnAFernandez:
AIUI, this should ensure that all the memory is freed. Not yet tested.


JohnAFernandez:
Added the same type of changes to the matching sphere init function.  Now also should fix coverities 1361844 and 1361840.

JohnAFernandez:
Now ready for review since I verified that the game runs and has no problem getting to a mission environment with these changes.

Goober5000 (approved):

PR #4001, opened 2022-02-02
Preserve the table-defined offset for initial misc animation playback

### LINKED ISSUE #4000: Misc anim delay in mainhall.tbl does not work as expected
Reported by MjnMixael:
In mainhall.tbl, Misc anims can be defined with delay parameters. These are expressed as a triplet of integers, the first of which is supposed to set an initial delay for an animation, followed by min and max delay parameters that govern the looping behaviour of an animation. Actual observed behavior seems to be that animations with that first parameter set to something that isn't the default value (-1) basically play immediately when entering the mainhall.

Upon inspection of the relevant code in mainhallmenu.cpp, my hypothesis is this:
1. During parsing, the value as defined in the table is stuffed into the `misc_anim_delay` vector
2. When determining which animation to render, this value is read and treated as a timestamp. 
3. Given that an indeterminate amount of time has passed between reading the mainhall table and actually entering the mainhall, the actual timestamp is likely to be far in advance of the table value, thus the animation is immediately restarted

This should be a relatively simple fix of adjusting the timestamps on entering the mainhall so that they are correctly treated as an offset, not an absolute value.

The-E:
This fixes #4000

### DIFF
--- a/code/menuui/mainhallmenu.cpp
+++ b/code/menuui/mainhallmenu.cpp
@@ -2373,8 +2373,10 @@ void parse_one_main_hall(bool replace, int num_resolutions, int &hall_idx, int &
 			// anim delay
 			required_string("+Misc anim delay:");
 			stuff_int(&m->misc_anim_initial_delay.at(idx));
-			// We do not set the first value in misc_anim_delay, as it is used as a timestamp that gets
-			// updated dynamically when the mainhall initializes.
+
+			// We set the first value here to -1; if the first delay parameter is set to something that isn't
+			// -1, we will replace it with a proper timestamp while loading up the mainhall for presentation.
+			m->misc_anim_delay.at(idx).at(0) = -1;
 			stuff_int(&m->misc_anim_delay.at(idx).at(1));
 			stuff_int(&m->misc_anim_delay.at(idx).at(2));
 		}
--- a/code/menuui/mainhallmenu.cpp
+++ b/code/menuui/mainhallmenu.cpp
@@ -1982,6 +1982,8 @@ void misc_anim_init(main_hall_defines &m, bool first_time, int base_num)
 		m.misc_anim_delay.back().push_back(-1);
 		m.misc_anim_delay.back().push_back(0);
 		m.misc_anim_delay.back().push_back(0);
+
+		// set the default initial delay to -1 
 		m.misc_anim_initial_delay.push_back(-1);
 
 		// misc_anim_paused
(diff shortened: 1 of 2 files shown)


code/menuui/mainhallmenu.cpp
```diff
@@ -1978,6 +1982,7 @@ void misc_anim_init(main_hall_defines &m, bool first_time, int base_num)
 		m.misc_anim_delay.back().push_back(-1);
 		m.misc_anim_delay.back().push_back(0);
 		m.misc_anim_delay.back().push_back(0);
+		m.misc_anim_initial_delay.push_back(-1);
```
Goober5000, line 1985: Add a comment here because "0 for the others" might lead someone to think this should be 0.

code/menuui/mainhallmenu.cpp
```diff
@@ -2365,7 +2370,7 @@ void parse_one_main_hall(bool replace, int num_resolutions, int &hall_idx, int &
 		for (int idx = base_num; idx < m->num_misc_animations; idx++) {
 			// anim delay
 			required_string("+Misc anim delay:");
-			stuff_int(&m->misc_anim_delay.at(idx).at(0));
+			stuff_int(&m->misc_anim_initial_delay.at(idx));
```
Goober5000, line 2373: The original line should be kept somehow.  I suggest restoring the original line and then setting the initial delay equal to the first delay.

Goober5000 (changes requested):
Makes sense, but two comments.

code/menuui/mainhallmenu.cpp
The-E: done.

code/menuui/mainhallmenu.cpp
The-E: This doesn't make much sense to me. The original behaviour was bogus; whatever value is set here is never going to be valid later on. Whatever value is put in here is going to be nuked anyway once we initialize the mainhall for proper use.

Goober5000: It doesn't sit right with me to initialize indexes 1 and 2 but not 0.  It may not be needed with this fix, but if someone comes along and changes things later, setting the value would keep things consistent.

The-E: If we have to set this value to something to preserve symmetry, we should set it to -1 and not something completely invalid.

Goober5000: Setting it to -1 is ok with me.

The-E: Done.

Goober5000 (approved):

PR #4003, opened 2022-02-05
Hud gauge and string upgrades

Goober5000:
Multiple features for more correct management of gauges, and drawing of strings on the screen.

1) upgrade scripting to reference default hud gauges by name, not just by index
2) allow a scripted hud gauge object to represent a default gauge, not just a custom gauge
3) rename hud_get_gauge to hud_get_custom_gauge for clarity
4) upgrade gr.drawString to be able to use a custom sizing mode


wookieejedi (approved):
Tested with the multitude of HUD related scripts that Fate of the Galaxy used and things works as expected. Also examined the code and it looks good to me.

PR #4011, opened 2022-02-09
Fix #4010

### LINKED ISSUE #4010: "not" sexp breaks subtly when used with arguments
Consider the following pseudo-event:
```
When-argument
Any-of
AND
 Destroyed-or-departed-delay
     Argument
 NOT
   Is cargo known delay
    Argument
   Modify variable
````
The intention is to modify the variable for each Argument with unknown cargo. However, `cargo-known-delay` can short-circuit, and when it does, this is inherited by `not`, breaking the intended behaviour.

The-E:
When evaluating the not sexxp, do not short circuit because we might be in an argument evaluation and _known_ values might not actually be known

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -5520,9 +5520,9 @@ int sexp_not(int n)
 		{
 			result = is_sexp_true(CAR(n));
 			if ( Sexp_nodes[CAR(n)].value == SEXP_KNOWN_FALSE || Sexp_nodes[CAR(n)].value == SEXP_NAN_FOREVER )
-				return SEXP_KNOWN_TRUE;												// not KNOWN_FALSE == KNOWN_TRUE;
+				return SEXP_TRUE;												// not KNOWN_FALSE == KNOWN_TRUE;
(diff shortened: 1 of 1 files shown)


Goober5000 (changes requested):
The comments should be updated because they now no longer reflect the code.

And this will require a bit of thought.  It makes sense, but it is also a change to original Volition code in an operator that dates back to FS1.  We'll need to analyze whether `SEXP_KNOWN_` values imply anything beyond the obvious short-circuiting.

Goober5000:
After doing some testing with the source mission this evening, I think the problem is likely to be somewhere else.  I think the `not` operator is a red herring.  Not only that, but the operator is behaving exactly as it is supposed to.

The "known" property is sticky.  It's not known-vs-not-known, it is known-[true-or-false].  The opposite of known-true is not unknown, it is known-false.

Think of it in the context of a ship exiting a mission.  If a ship's cargo was scanned, then after the ship exists, that sexp will always return known-true.  There is no possibility of it ever becoming something other than true, so it's known.  Similarly, the inverse of that is known-false, because there is no possibility of the inverse ever becoming something other than false.

Goober5000:
Closing as this change should not be made.  However I'll open a new PR to improve the comments in that function.

PR #4018, opened 2022-02-12
Fix qtFRED crash when using Edit Data

jg18:
On Windows, there's currently a debug heap assertion failure when trying to use Edit Data.

This change seems to fix it.

I originally thought the issue was with Add Data >> Number/String (hence the branch name) but realized it was actually Edit Data.

I wonder if the issue is `QString`'s implementation of `toStdString()` itself.

Unrelated known issue: when Add Data >> Number is used and the text is left as "number" and the Events Editor is closed with OK, qtFRED doesn't automatically change the value to 0, but if you save/reload the mission, it's set to 0.

### DIFF
--- a/qtfred/src/ui/widgets/sexp_tree.cpp
+++ b/qtfred/src/ui/widgets/sexp_tree.cpp
@@ -5736,7 +5736,7 @@ void sexp_tree::handleItemChange(QTreeWidgetItem* item, int  /*column*/) {
 
 	if (update_node) {
 		modified();
-		strncpy(tree_nodes[node].text, str.toStdString().c_str(), len);
+		strncpy(tree_nodes[node].text, str.toUtf8().constData(), len);
 		tree_nodes[node].text[len] = 0;
 
 		// let's make sure we aren't introducing any invalid characters, per Mantis #2893
--- a/qtfred/src/ui/widgets/sexp_tree.cpp
+++ b/qtfred/src/ui/widgets/sexp_tree.cpp
@@ -5736,7 +5736,8 @@ void sexp_tree::handleItemChange(QTreeWidgetItem* item, int  /*column*/) {
 
 	if (update_node) {
 		modified();
-		strncpy(tree_nodes[node].text, str.toUtf8().constData(), len);
+		auto strBytes = str.toUtf8(); // avoid using dangling ptr
+		strncpy(tree_nodes[node].text, strBytes.constData(), len);
 		tree_nodes[node].text[len] = 0;
 
 		// let's make sure we aren't introducing any invalid characters, per Mantis #2893


jg18:
Could someone verify this fix on Linux?

Goober5000:
This is odd.  What's the actual assertion failure say?

And does it make any difference if you take the string length of the `c_str()`?

jg18:
Still investigating this, but the existing call to `str.toStdString().c_str()` is surely a problem. It's returning a pointer to a `string` object that will get collected sometime after `c_str()` returns, meaning we're using a potentially dangling pointer.

I tried declaring
```
const SCP_string scpStr = str.toStdString();
```
and then using `scpStr.c_str()` in the call to `strncopy()`, but am now getting what I think is the same heap error but now a few lines below that, in I think the call to `lcl_fred_replace_stuff()`, but I can't see how/why.

jg18:
From Visual Studio debugger console:
```
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: wil::ResultException at memory location 0x000000000A43CFD0.
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: wil::ResultException at memory location 0x000000000A43D430.
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
The thread 0x1ea4 has exited with code 0 (0x0).
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: wil::ResultException at memory location 0x000000000A43CFD0.
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: wil::ResultException at memory location 0x000000000A43D430.
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: wil::ResultException at memory location 0x000000000A43CFD0.
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: wil::ResultException at memory location 0x000000000A43D430.
Exception thrown at 0x00007FF9965E4F69 in qtfred_21_5_0_x64_AVX2-DEBUG.exe: Microsoft C++ exception: [rethrow] at memory location 0x0000000000000000.
Debug Assertion Failed!

Program: C:\Games\FreeSpace2\qtfred_21_5_0_x64_AVX2-DEBUG.exe
File: minkernel\crts\ucrt\src\appcrt\heap\debug_heap.cpp
Line: 996

Expression: __acrt_first_block == header

For information on how your program can cause an assertion
failure, see the Visual C++ documentation on asserts.
```

jg18:
However, with this PR's change, I can't repro the heap assertion failure.

jg18:
@Goober5000 I'm not sure what you mean by
> take the string length of the `c_str()`
but I can say from the debugger than the value of `len` passed to `strncpy()` looks correct. I can triple-check this later.

JohnAFernandez (approved):
I think you have the right approach here.  I'm approving.

PR #4026, opened 2022-02-14
a few scripting bugfixes

Goober5000:
When returning objects, be sure to use `ade_set_object_with_breed`, per the documentation on that function.  This ensures that all relevant subclass information can be used (e.g. for ships, the `.Name` virtvar).

Also fix the return value of streaminganim's isValid() function.


wookieejedi (approved):
Good fixes, was just actually thinking about making a PR for this, too.

PR #4029, opened 2022-02-14
Don't nullify axis_out[action] value when using mouse as joy

### LINKED ISSUE #4028: Mouse as Joy Doesn't work if both Joy-0 and Mouse axes bound to same control
image

image

Reported by wookijedi, Should the mouse input mode be set to Joy-0, and the mouse is bound to the same axes control as a joystick, then the inputs are effectively ignored.

z64555:
Fixes #4028


wookieejedi (approved):
Tested and fixes the issue.

Issue #3985, opened 2022-01-30
Assertion relating to animated submodels


MjnMixael:
BtA has encountered this Assert 

Assert: "pm->submodel[submodel_num].movement_type == MOVEMENT_TYPE_ROT || pm->submodel[submodel_num].movement_type == MOVEMENT_TYPE_INTRINSIC"

The model in question has two animated submodels. They are doors that use animation type Rotation on docking-stage-2 to open and animation type Rotation on Docked to close. Looks like the Assert happened as the ship was going down, possibly while in the middle of docking or undocking. Relevant log is attached.

Animation details are here. And I verified that the submodels have no additional properties in the POF.

> $Name: bta_morgens_dock_s2
> $Type: docking-stage-2
> +Triggered By: 2
> $Segment Parallel:
> 	$Rotation:
> 		+Angle: 0,0,90
> 		+Velocity: 0,0,40
> 		+Acceleration: 0,0,5
> 		+Submodel: Door_L-a
> 	$Rotation:
> 		+Angle: 0,0,-90
> 		+Velocity: 0,0,40
> 		+Acceleration: 0,0,5
> 		+Submodel: Door_R-a
> +End Segment
> 
> $Name: bta_morgens_docked
> $Type: docked
> +Triggered By: 2
> $Segment Parallel:
> 	$Rotation:
> 		+Angle: 0,0,-90
> 		+Velocity: 0,0,40
> 		+Acceleration: 0,0,5
> 		+Submodel: Door_L-a
> 	$Rotation:
> 		+Angle: 0,0,90
> 		+Velocity: 0,0,40
> 		+Acceleration: 0,0,5
> 		+Submodel: Door_R-a
> +End Segment

[fs2_open (5).log](https://github.com/scp-fs2open/fs2open.github.com/files/7966423/fs2_open.5.log)

chief1983:
Haha, FotG just found this yesterday I think.  Goober is looking into it, but fixing it will take a bit he said.  We implemented a workaround but not sure that will help your situation at the moment.

> Goober figured out that collisions do not work with nested submodels, but it looks like that will be an involved fix. Fortunately, though, I just flattened out the nested submodels in the X-wing and updated the animations to ensure s-foils still work properly.

MjnMixael:
I'm not sure I fully understand your workaround. The doors here are parented only to detail0. I can't flatten that out any further, I don't think.

chief1983:
[...]

PR #4030, opened 2022-02-15
clean up model point/dir/normal/orient functions

Goober5000:
Consolidated these functions and replaced the `find_submodel_point_*` functions with their closest `world_find_` counterparts.  (One of those functions may not have actually worked correctly.)  Also use the correct coordinate logic when returning normals from collision info in scripting.

Related to #3985 although not yet a fix for it.

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -4101,7 +4101,7 @@ void model_find_submodel_offset(vec3d *outpnt, const polymodel *pm, int submodel
 	}
 }
 
-void model_find_world_point(vec3d *outpnt, vec3d *mpnt, int model_num, int submodel_num, const matrix *objorient, const vec3d *objpos)
+void model_find_world_point(vec3d *outpnt, const vec3d *mpnt, int model_num, int submodel_num, const matrix *objorient, const vec3d *objpos)
 {
 	return model_find_world_point(outpnt, mpnt, model_get(model_num), submodel_num, objorient, objpos);
 }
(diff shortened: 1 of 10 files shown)


code/model/modelread.cpp
```diff
@@ -4181,152 +4191,103 @@ void world_find_model_instance_point(vec3d *out, vec3d *world_pt, const polymode
...
 
-		matrix tnorm = *outorient;
-		vm_matrix_x_matrix(outorient, &tnorm, &pmi->submodel[parent_mn].canonical_orient);
+	// instance up the tree for this point
+	while ( (mn >= 0) && (pm->submodel[mn].parent >= 0) ) {
+		vm_vec_unrotate(&tpnt, &pnt, &pmi->submodel[mn].canonical_orient);
+		vm_vec_add(&pnt, &tpnt, &pm->submodel[mn].offset);
 
-		vm_vec_add2(outpnt, &offset);
+		orient = orient * pmi->submodel[mn].canonical_orient;
```
BMagnu, line 4276: Just to comment on this:
The order for the matrix multiplications is changed here on purpose it seems, as the previous order would incorrectly apply the rotations to the stack. As we're going up the chain, and not down, this seems to change it to the correct order

BMagnu (approved):
Looks to be good to me.
These functions will need some more documentations and renaming to not be confusing and with all the other functions like these with similar names, but as discussed with @Goober5000 this'll happen in a separate followup PR

PR #4031, opened 2022-02-15
add scripting support for submodels

Goober5000:
Allow scripts to reference submodels, submodel instances, and model instances.  Currently just a few properties (like Orientation and Radius) can be accessed, but this can be easily expanded in the future.

Also add relevant functions to calculate real-world positions and orientations of rotated submodels.

### DIFF
--- a/code/scripting/api/objs/mc_info.cpp
+++ b/code/scripting/api/objs/mc_info.cpp
@@ -18,7 +18,7 @@ bool mc_info_h::IsValid() { return valid; }
 //**********HANDLE: Collision info
 ADE_OBJ(l_ColInfo, mc_info_h, "collision_info", "Information about a collision");
 
-ADE_VIRTVAR(Model, l_ColInfo, "model", "The model this collision info is about", "model", "The model")
+ADE_VIRTVAR(Model, l_ColInfo, "model", "The model this collision info is about", "model", "The model, or an invalid model if the handle is not valid")
 {
 	mc_info_h* info;
 	model_h * mh = nullptr;
@@ -44,16 +44,16 @@ ADE_VIRTVAR(Model, l_ColInfo, "model", "The model this collision info is about",
 	return ade_set_args(L, "o", l_Model.Set(model_h(modelNum)));
 }
 
-ADE_FUNC(getCollisionSubmodel, l_ColInfo, nullptr, "The submodel where the collision occurred, if applicable", "submodel", "The submodel, or nil if none")
+ADE_FUNC(getCollisionSubmodel, l_ColInfo, nullptr, "The submodel where the collision occurred, if applicable", "submodel", "The submodel, or nil if none or if the handle is not valid")
 {
 	mc_info_h *info;
 	submodel_h *sh = nullptr;
 
 	if (!ade_get_args(L, "o|o", l_ColInfo.GetPtr(&info), l_Submodel.GetPtr(&sh)))
-		return ade_set_error(L, "o", l_Submodel.Set(submodel_h()));
+		return ADE_RETURN_NIL;
 
 	if (!info->IsValid())
-		return ade_set_error(L, "o", l_Submodel.Set(submodel_h()));
+		return ADE_RETURN_NIL;
 
 	mc_info *collide = info->Get();
 
@@ -85,7 +85,7 @@ ADE_FUNC(getCollisionDistance, l_ColInfo, NULL, "The distance to the closest col
 	}
 }
 
-ADE_FUNC(getCollisionPoint, l_ColInfo, "[boolean local]", "The collision point of this information (local to the object if boolean is set to <i>true</i>)", "vector", "The collision point or nil of none")
+ADE_FUNC(getCollisionPoint, l_ColInfo, "[boolean local]", "The collision point of this information (local to the object if boolean is set to <i>true</i>)", "vector", "The collision point, or nil if none or if the handle is not valid")
 {
 	mc_info_h* info;
 	bool local = false;
@@ -111,7 +111,7 @@ ADE_FUNC(getCollisionPoint, l_ColInfo, "[boolean local]", "The collision point o
 	}
 }
 
-ADE_FUNC(getCollisionNormal, l_ColInfo, "[boolean local]", "The collision normal of this information (local to object if boolean is set to <i>true</i>)", "vector", "The collision normal or nil of none")
+ADE_FUNC(getCollisionNormal, l_ColInfo, "[boolean local]", "The collision normal of this information (local to object if boolean is set to <i>true</i>)", "vector", "The collision normal, or nil if none or if the handle is not valid")
 {
 	mc_info_h* info;
 	bool local = false;
@@ -152,7 +152,7 @@ ADE_FUNC(isValid, l_ColInfo, NULL, "Detects if this handle is valid", "boolean",
 	mc_info_h* info;
 
 	if(!ade_get_args(L, "o", l_ColInfo.GetPtr(&info)))
-		return ADE_RETURN_NIL;
+		return ADE_RETURN_FALSE;
 
 	if (info->IsValid())
 		return ADE_RETURN_TRUE;
--- a/code/scripting/api/objs/mc_info.cpp
+++ b/code/scripting/api/objs/mc_info.cpp
@@ -44,6 +44,25 @@ ADE_VIRTVAR(Model, l_ColInfo, "model", "The model this collision info is about",
 	return ade_set_args(L, "o", l_Model.Set(model_h(modelNum)));
 }
 
+ADE_FUNC(getCollisionSubmodel, l_ColInfo, nullptr, "The submodel where the collision occurred, if applicable", "submodel", "The submodel, or nil if none")
+{
+	mc_info_h *info;
+	submodel_h *sh = nullptr;
+
+	if (!ade_get_args(L, "o|o", l_ColInfo.GetPtr(&info), l_Submodel.GetPtr(&sh)))
+		return ade_set_error(L, "o", l_Submodel.Set(submodel_h()));
+
+	if (!info->IsValid())
+		return ade_set_error(L, "o", l_Submodel.Set(submodel_h()));
+
+	mc_info *collide = info->Get();
+
+	if (collide->hit_submodel < 0)
+		return ADE_RETURN_NIL;
+
+	return ade_set_args(L, "o", l_Submodel.Set(submodel_h(collide->model_num, collide->hit_submodel)));
+}
+
 ADE_FUNC(getCollisionDistance, l_ColInfo, NULL, "The distance to the closest collision point", "number", "distance or -1 on error")
 {
 	mc_info_h* info;
--- a/code/scripting/api/objs/model.cpp
+++ b/code/scripting/api/objs/model.cpp
@@ -66,7 +66,7 @@ bool submodel_h::IsValid()
 }
 
 
-ADE_VIRTVAR(Submodels, l_Model, "modelsubmodels_h", "Model submodels", "modelsubmodels_h", "Model submodels, or an invalid modelsubmodels handle if the model handle is invalid")
+ADE_VIRTVAR(Submodels, l_Model, "Submodels", "Submodels", "Submodels", "Model submodels, or an invalid modelsubmodels handle if the model handle is invalid")
 {
 	model_h *mdl = nullptr;
 	if (!ade_get_args(L, "o", l_Model.GetPtr(&mdl)))
@@ -155,7 +155,7 @@ ADE_VIRTVAR(Dockingbays, l_Model, "dockingbays", "Docking bays handle of model",
 	return ade_set_args(L, "o", l_Dockingbays.Set(dockingbays_h(pm)));
 }
 
-ADE_VIRTVAR(BoundingBoxMax, l_Model, "vector", "Model bounding box maximum", "vector", "Model bounding box, or an empty vector if the handle is invalid")
+ADE_VIRTVAR(BoundingBoxMax, l_Model, "vector", "Model bounding box maximum", "vector", "Model bounding box, or an empty vector if the handle is not valid")
 {
 	model_h *mdl = NULL;
 	vec3d *v = NULL;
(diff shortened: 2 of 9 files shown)


code/scripting/api/objs/model.cpp
```diff
@@ -243,15 +278,186 @@ ADE_VIRTVAR(Radius, l_Model, "number", "Model radius (Used for collision & culli
...
+	submodel_h *smh;
+	if (!ade_get_args(L, "o", l_Submodel.GetPtr(&smh)))
+		return ADE_RETURN_NIL;
+
+	return ade_set_args(L, "b", smh->IsValid());
+}
+
+
+//**********HANDLE: modelsubmodels
+ADE_OBJ(l_ModelSubmodels, modelsubmodels_h, "modelsubmodels_h", "Array of submodels");
```
BMagnu, line 400: This should not be its own handle type.
Arrays of a custom type are otherwise handled using sublibraries in other places, such as the mission library (see [Mission Library](https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/scripting/api/libs/mission.cpp#L160...L185), just that instead of throwing away the first object, capture it as l_Model, as is done in most other cases where we use indexers)

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+{
+	return _pmi;
+}
+bool modelinstance_h::IsValid()
+{
+	return (_pmi != nullptr);
+}
+
+
+ADE_OBJ(l_SubmodelInstance, submodelinstance_h, "submodelinstance", "Submodel instance handle");
```
BMagnu, line 34: as to stay consistent with naming, I'd suggest snake case here for the user-facing type name: submodel_instance

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+#include "model.h"
+#include "modelinstance.h"
+#include "object.h"
+#include "vecmath.h"
+
+namespace scripting {
+namespace api {
+
+ADE_OBJ(l_ModelInstance, modelinstance_h, "modelinstance", "Model instance handle");
```
BMagnu, line 12: Snake case here as well, model_instance

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+	auto pmi = mih->Get();
+	if (!pmi)
+		return ade_set_error(L, "o", l_ModelSubmodelInstances.Set(modelsubmodelinstances_h()));
+
+	if (ADE_SETTING_VAR)
+		LuaError(L, "Attempt to use Incomplete Feature: submodel instance copy");
+
+	return ade_set_args(L, "o", l_ModelSubmodelInstances.Set(modelsubmodelinstances_h(pmi)));
+}
```
BMagnu, line 111: The whole modelsubmodelinstances_h thing should be a sublibrary as well, same as modelsubmodels

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+ADE_FUNC(isValid, l_ModelInstance, nullptr, "True if valid, false or nil if not", "boolean", "Detects whether handle is valid")
+{
+	modelinstance_h *mih;
+	if (!ade_get_args(L, "o", l_ModelInstance.GetPtr(&mih)))
+		return ADE_RETURN_NIL;
+
+	return ade_set_args(L, "b", mih->IsValid());
+}
+
+ADE_VIRTVAR(Orientation, l_SubmodelInstance, "orientation", "Gets or sets the submodel instance orientation (world orientation)", "orientation", "Orientation, or null orientation if handle is invalid")
```
BMagnu, line 122: nil instead of null

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+{
+	submodelinstance_h *smih;
+	matrix_h *mh = nullptr;
+	if (!ade_get_args(L, "o|o", l_SubmodelInstance.GetPtr(&smih), l_Matrix.GetPtr(&mh)))
+		return ade_set_error(L, "o", l_Matrix.Set(matrix_h(&vmd_identity_matrix)));
+
+	if (!smih->IsValid())
+		return ade_set_error(L, "o", l_Matrix.Set(matrix_h(&vmd_identity_matrix)));
```
BMagnu, line 130: if not valid, actually return nil instead of an identity matrix.

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+		vm_closest_angle_to_matrix(&smi->canonical_orient, &sm->rotation_axis, &angle);
+
+		smi->cur_angle = angle;
+		smi->turret_idle_angle = angle;
+	}
+
+	return ade_set_args(L, "o", l_Matrix.Set(matrix_h(&smih->Get()->canonical_orient)));
+}
+
+ADE_FUNC(findWorldPoint, l_SubmodelInstance, "vector", "Calculates the world coordinates of a point in a submodel's frame of reference", "vector", "Point, or null vector if handle is invalid")
```
BMagnu, line 150: nil as well

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+	vec3d pnt, outpnt;
+	if (!ade_get_args(L, "oo", l_SubmodelInstance.GetPtr(&smih), l_Vector.Get(&pnt)))
+		return ade_set_error(L, "o", l_Vector.Set(vmd_zero_vector));
+
+	if (!smih->IsValid())
+		return ade_set_error(L, "o", l_Vector.Set(vmd_zero_vector));
+
+	auto pmi = smih->GetModelInstance();
+	if (pmi->objnum < 0)
+		return ade_set_error(L, "o", l_Vector.Set(vmd_zero_vector));
```
BMagnu, line 162: As well as actually returning nil instead of 0

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+	auto objp = &Objects[pmi->objnum];
+
+	model_instance_find_world_point(&outpnt, &pnt, pmi->id, smih->GetSubmodelIndex(), &objp->orient, &objp->pos);
+	return ade_set_args(L, "o", l_Vector.Set(outpnt));
+}
+
+ADE_FUNC(findWorldDir, l_SubmodelInstance, "vector", "Calculates the world direction of a vector in a submodel's frame of reference", "vector", "Vector, or null vector if handle is invalid")
```
BMagnu, line 169: nil

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+	vec3d dir, outdir;
+	if (!ade_get_args(L, "oo", l_SubmodelInstance.GetPtr(&smih), l_Vector.Get(&dir)))
+		return ade_set_error(L, "o", l_Vector.Set(vmd_zero_vector));
+
+	if (!smih->IsValid())
+		return ade_set_error(L, "o", l_Vector.Set(vmd_zero_vector));
+
+	auto pmi = smih->GetModelInstance();
+	if (pmi->objnum < 0)
+		return ade_set_error(L, "o", l_Vector.Set(vmd_zero_vector));
```
BMagnu, line 181: Return nil as well

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+	auto pmi = smih->GetModelInstance();
+	if (pmi->objnum < 0)
+		return false;
+	auto objp = &Objects[pmi->objnum];
+
+	model_instance_find_world_point_orient(outpnt, outorient, &local_pnt, local_orient->GetMatrix(), smih->GetModel(), pmi, smih->GetSubmodelIndex(), use_object ? &objp->orient : nullptr, use_object ? &objp->pos : nullptr);
+	return true;
+}
+
+ADE_FUNC(findObjectPointAndOrientation, l_SubmodelInstance, "vector, orientation", "Calculates the coordinates and orientation, in an object's frame of reference, of a point and orientation in a submodel's frame of reference", "vector, orientation", "Vector and orientation, or null values if a handle is invalid")
```
BMagnu, line 208: nil

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+{
+	vec3d pnt;
+	matrix orient;
+
+	if (!findObjectPointAndOrientationSub(L, false, &pnt, &orient))
+		return ade_set_error(L, "oo", l_Vector.Set(vmd_zero_vector), l_Matrix.Set(matrix_h(&vmd_zero_matrix)));
```
BMagnu, line 214: return nil as well

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+	return ade_set_args(L, "oo", l_Vector.Set(pnt), l_Matrix.Set(matrix_h(&orient)));
+}
+
+ADE_FUNC(findWorldPointAndOrientation, l_SubmodelInstance, "vector, orientation", "Calculates the world coordinates and orientation of a point and orientation in a submodel's frame of reference", "vector, orientation", "Vector and orientation, or null values if a handle is invalid")
```
BMagnu, line 219: nil

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+{
+	vec3d pnt;
+	matrix orient;
+
+	if (!findObjectPointAndOrientationSub(L, true, &pnt, &orient))
+		return ade_set_error(L, "oo", l_Vector.Set(vmd_zero_vector), l_Matrix.Set(matrix_h(&vmd_zero_matrix)));
```
BMagnu, line 225: return nil

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+	return ade_set_args(L, "oo", l_Vector.Set(pnt), l_Matrix.Set(matrix_h(&orient)));
+}
+
+ADE_FUNC(isValid, l_SubmodelInstance, nullptr, "True if valid, false or nil if not", "boolean", "Detects whether handle is valid")
+{
+	submodelinstance_h *smih;
+	if (!ade_get_args(L, "o", l_SubmodelInstance.GetPtr(&smih)))
+		return ADE_RETURN_NIL;
```
BMagnu, line 234: This should really never return nil, just true or false in all cases

code/scripting/api/objs/object.cpp
```diff
@@ -171,6 +172,25 @@ ADE_VIRTVAR(LastOrientation, l_Object, "orientation", "Object world orientation
 	return ade_set_args(L, "o", l_Matrix.Set(matrix_h(&objh->objp->last_orient)));
 }
 
+ADE_VIRTVAR(ModelInstance, l_Object, "modelinstance", "model instance used by this object", "modelinstance", "Model instance, nil if this object does not have one, or invalid model instance handle if object handle is invalid")
```
BMagnu, line 175: the second "modelinstance" string should be nullptr if the value cannot be assigned. This will correctly display non-assignability in the documentation

BMagnu (changes requested):
Some comments on this.
Generally a good and imporant addition, but some pointers regarding the implementation

code/scripting/api/objs/mc_info.cpp
```diff
@@ -44,6 +44,25 @@ ADE_VIRTVAR(Model, l_ColInfo, "model", "The model this collision info is about",
 	return ade_set_args(L, "o", l_Model.Set(model_h(modelNum)));
 }
 
+ADE_FUNC(getCollisionSubmodel, l_ColInfo, nullptr, "The submodel where the collision occurred, if applicable", "submodel", "The submodel, or nil if none")
+{
+	mc_info_h *info;
+	submodel_h *sh = nullptr;
+
+	if (!ade_get_args(L, "o|o", l_ColInfo.GetPtr(&info), l_Submodel.GetPtr(&sh)))
+		return ade_set_error(L, "o", l_Submodel.Set(submodel_h()));
```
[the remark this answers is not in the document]
Goober5000: The reason I did it this way was that, for better or worse, it appears most of the rest of the code does it that way.  Look at `getObjectFromSignature` or `getSoundentry` or `getStartDirectionInfo`.  In these cases (plus many others; search for "return ade_set_error") these return an invalid object rather than nil.

code/scripting/api/objs/model.cpp
Goober5000: I patterned this after the other arrays in model.cpp, but I'll take a look at the mission library

code/scripting/api/objs/modelinstance.cpp
Goober5000: "null orientation" in this case means the identity matrix

BMagnu: Ah, that should be made clear in the documentation then.

code/scripting/api/objs/model.cpp
BMagnu: IIRC the only place that does it like that are the model textures, and I _really_ don't like these as well.
We should avoid creating more of these, as it introduces additional types that have no use apart from wrapping functions.

code/scripting/api/objs/object.cpp
Goober5000: Ok

code/scripting/api/objs/mc_info.cpp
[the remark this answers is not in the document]
BMagnu: In that case the documentation must be extended to make clear that an invalid request will give an invalid submodel and _not_ a nil for "no submodel found", which it currently implies

code/scripting/api/objs/modelinstance.cpp
Goober5000: ok

BMagnu (approved):

PR #4035, opened 2022-02-16
Expand scripting string functions

### LINKED ISSUE #4034: Add `ba.directiveText()`
Exactly what it says on the tin. `ba.directiveText("Press $Alt-J$")` would evaluate to "Press Alt-J" or "Press Shift-J" or whatever the player's mapped the control to, and so on, just like directive keypress text currently does.

wookieejedi:
Allows scripters to use the functions `message_translate_tokens` and `sexp_replace_variable_names_with_values`. This is very useful with scripted HUD gauges and scripted messages. Also, fixes #4034.

### DIFF
--- a/code/scripting/api/libs/base.cpp
+++ b/code/scripting/api/libs/base.cpp
@@ -484,10 +484,11 @@ ADE_FUNC(XSTR,
 }
 
 ADE_FUNC(translateTokens, 
-	l_Base, "string to translate", 
+	l_Base, 
+	"string to translate", 
 	"Returns a string that translates any default control binding to current binding (same as Directive Text). Default binding must be encapsulated by '$$' to be translated.",
 	"string",
-	"string")
+	"string or nil if invalid")
 {
 	const char* untranslated_str;
 	if (!ade_get_args(L, "s", &untranslated_str)) {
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/base.cpp
```diff
@@ -481,6 +483,41 @@ ADE_FUNC(XSTR,
...
+	if (!ade_get_args(L, "s", &untranslated_str)) {
+		return ADE_RETURN_NIL;
+	}
+
+	SCP_string translated_str = message_translate_tokens(untranslated_str);
+
+	return ade_set_args(L, "s", translated_str.c_str());
+}
+
+ADE_FUNC(replaceVariableValue,
```
Goober5000, line 503: I would call this `translateVariables`, to be consistent with the other function name.  Or maybe a better idea would be to use `replaceTokens` and `replaceVariables`.

Goober5000 (changes requested):

code/scripting/api/libs/base.cpp
wookieejedi: Good idea, updated names.

Goober5000 (approved):

PR #4037, opened 2022-02-19
add some new scan features

### LINKED ISSUE #4027: FR: Improve scan features
It's come to my attention that features for scanning in FSO are fairly limited, as we only have time per ship and range per ship class right now. Two new features that immediately come to mind that I have a use case for are as follows, but given how little attention scanning has apparently received, we may want to see if others have anything else to add.

1. Custom scan times per subsystem - this is almost a must. Right now, all subsystem scan times inherit the time of their parent ship, resulting in some scan times that are inordinately long, particularly on freighters and cruisers.

2. Scan speed multiplier per ship - this one is more of a personal request, but I think it would be very cool to specify a scan multiplier in the sensor subsystem of a player-flown ship which would allow certain ships eg stealth/recon fighters to scan targets more rapidly. This could even be made more modular so that "advanced avionics packages" or whatever could be equipped a la Tenebra which take up a weapon slot to boost scan speed.

Goober5000:
1. Add custom scan times per subsystem
2. Add scanning time and range multipliers for player ships

Resolves #4027

### DIFF
--- a/code/hud/hudtargetbox.cpp
+++ b/code/hud/hudtargetbox.cpp
@@ -693,7 +693,7 @@ void HudGaugeTargetBox::renderTargetShip(object *target_objp)
 	setGaugeColor();
 
 	renderTargetShipInfo(target_objp);
-	maybeRenderCargoScan(target_sip);
+	maybeRenderCargoScan(target_sip, Player_ai->targeted_subsys);
 }
 
 /**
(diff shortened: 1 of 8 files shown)


wookieejedi (commented):
Mage stated that we do not want to change ship class info via scripting, and Asteroth and I have also talked about making sure to limit anything scripts that do that in the future. If this is a critical feature for a mod, adding the value to the ship object (like what was done with auto-aim) could be done.

Baezon:
Since it's what all the others do, I don't think he can really be blamed for doing it too, but I do think we need to have a collective discussion on whether we truly intend to support runtime modification of ship_info or if it should be deprecated across the board, although that is beyond the scope of this PR.

Goober5000:
mmh.  I thought there might be occasions where it could be useful, but it's not a critically required feature.  I've removed the ability to set the values.

Baezon (approved):

z64555:
Er, please don't forget to update the wiki. :)

Goober5000:
I did! :)

PR #4039, opened 2022-02-20
fix an edge case for ships destroyed before mission

Goober5000:
Make sure the ship registry correctly records these ships as having exited.


JohnAFernandez (approved):

code/mission/missionparse.cpp
```diff
@@ -3573,6 +3573,13 @@ void mission_parse_maybe_create_parse_object(p_object *pobjp)
 				// Make sure that the ship is marked as destroyed so the AI doesn't freak out later
 				ship_add_exited_ship(&Ships[objp->instance], Ship::Exit_Flags::Destroyed);
 
+				// Same with the ship registry so that SEXPs don't refer to phantom ships
```
JohnAFernandez, line 3576: I see what you did there.

Issue #4040, opened 2022-02-20
ControlConfigOverride in ccfg.tbl doesn't unbind defaults


z64555:
When using the controlsconfigdefaults.tbl to set the default preset, key bindings can not be intentionally unbound, instead they retain their FS2 defaults.

Observed behavior:
`$Key Default: NONE` does nothing

Expected behavior:
`$Key Default: NONE` should unbind the given control

Goober5000:
Closing as fixed

PR #4042, opened 2022-02-20
Fix issue #4040

z64555:
Set key to -1 and let CCB::take() do the actual binding later on instead of overwriting changes.

Should fix issue #4040


wookieejedi (approved):
Just tested and it fixes the issue. Thanks!

PR #4050, opened 2022-02-23
Feature/texture replacement dialog

TheForce172:
image
New texture replacement dialog. Made with bulk editing in mind. Available map types dynamically changes based on what subtypes the default map has.
image
It also no longer requires the user to type out the subtype names i.e. -misc -shine. that's handled automatically. and they can inherit their name from the main texture if all the submaps are named the same.

### DIFF
--- a/qtfred/src/mission/dialogs/ShipTextureReplacementDialogModel.cpp
+++ b/qtfred/src/mission/dialogs/ShipTextureReplacementDialogModel.cpp
@@ -451,7 +451,9 @@ namespace fso {
 							}
 						}
 						else {
-							error_display(0, "You told us to replace the -misc map with inherited data but didnt change the main texture name. Ignoring -misc map change");
+							SCP_string buf;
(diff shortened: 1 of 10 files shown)


qtfred/src/mission/dialogs/ShipTextureReplacementDialogModel.cpp
```diff
@@ -0,0 +1,642 @@
...
+										{ DialogButton::Ok });
+									if (button == DialogButton::Ok) {
+										return;
+									}
+								}
+
+							}
+						}
+						else {
+							error_display(0, "You told us to replace the -misc map with inherited data but didnt change the main texture name. Ignoring -misc map change");
```
Goober5000, line 454: I would phrase this in a different way.  Suggestion - "Cannot use inherited data without changing the main texture name.  Ignoring -misc map change."

Goober5000 (changes requested):

Goober5000 (approved):

Issue #4048, opened 2022-02-23
memory access violation when loading the Ganymede


Goober5000:
In the `align_bsp_data` function:
```
	do {
		//Read Chunk type and size
		memcpy(&bsp_chunk_type, bsp_in, 4);
		memcpy(&bsp_chunk_size, bsp_in + 4, 4);
```

When loading the Ganymede model from the MediaVPs on a 32-bit build, the last quoted line throws an access violation.  Presumably it is trying to read past the end of the file.  At the time of the crash, `bsp_chunk_type` has the value 0, indicating EOF, but `bsp_chunk_size` cannot be read.

64-bit builds can load the Ganymede without crashing, but this may just be hiding the underlying error.

Goober5000:
Fixed by #4051

PR #4051, opened 2022-02-23
Dont try to read chunk size if chunk type is 0

Shivansps:
Chunk type 0 is EOF and lacks the following int indicating the total chunk size.


Goober5000:
Should fix #4048.  However @notimaginative will be running a few tests first

Goober5000 (approved):
Fixes the crash, and @notimaginative looked at it too, so I'll approve it.

PR #4052, opened 2022-02-27
Repair Automated Forum and Nebula Scripts

z64555:
Salvages #3749 

* Documents the .py scripts with some abominable cross between doxygen and docstrings.
* Minor refactors to how groups are formed for clarity

TODO:
- [x] Inspect the python scripts for potential use in manually triggered workflows
- [x] Double check the nightly release workflow to see where their files are put/fetched
- [x] Re-enable the scripts to automatically fire as a post-process to the Release and Nightly build workflows


z64555:
The scripts use HTTP GET requests with Github REST API to retrieve metadata of the various files.  In particular, it grabs the metadata for the files hosted in the [Releases server](https://github.com/scp-fs2open/fs2open.github.com/releases).

I can probably make a post-release.yaml to run these scripts.  This workflow would then be triggered by the release and nightly workflows and optionally manually triggered should either of them fail for whatever reason.  The scripts should need, at most, only the git tag of the release/nightly they're pushing.

z64555:
Some notes:

1. I haven't tested this, and I'm not sure if I'd be able to test on my fork since this uses forum and Nebula credentials
2. The post-build-release workflow is responsible for doing the forum post and nebula post.  It grabs the release files from the github Release location by using the tag name as its reference, if it can't find the tag then it should fail before attempting to post or upload to nebula
3. The post-build-release has two triggers, one that should be automatically triggered by the build-release workflow and one that can be manually triggered.  The manual trigger does not check to see if there was any failures building or publishing (it can't, really...) so use only after you've verified that all relavent files are there.
4. Theoretically the same workflows and scripts in this PR can be used for Nightly posts, but since we still have the old Nightly trigger in place that would have to be addressed first.

asarium:
You should be able to test this on your fork without affecting the main repository. If you do not set the HLP or Nebula tokens then there is also no risk of accidentally posting a release.

z64555:
Finally managed to get the manual trigger to go through.  All that's left is to test it on the main repo.

TRBlount (approved):
Looks alright, lgtm.

PR #4053, opened 2022-02-28
Fix uniform offset calcuation for decal header

asarium:
This was broken by the move to a single, global uniform buffer where we
started doing suballocation so offset 0 was reading bad data.

Also fixed a crash regarding an unknown locale name on Linux.


code/globalincs/systemvars.cpp
```diff
@@ -469,7 +469,7 @@ void insertion_sort(void *array_base, size_t array_size, size_t element_size, in
 
 // Stuff that can't be included in vmallocator.h
 
-std::locale SCP_default_locale("");
+std::locale SCP_default_locale;
```
Goober5000, line 472: I'm surprised the empty string is causing an error here, since this is apparently what you are supposed to do.  Here are the instructions I followed:
https://stackoverflow.com/questions/46364608/creating-an-stdlocale-with-an-empty-string

Goober5000 (approved):

code/globalincs/systemvars.cpp
asarium: I think my IDE broke something there. From a terminal this code is fine. Reverted it again.

PR #4056, opened 2022-03-03
add capability (plus sexps) to set navpoint color

Goober5000:
This is useful because certain backgrounds (such as a nebula) will make it hard to see the default navpoint color.  Also this means that different navpoints can now have different colors.


wookieejedi (approved):
Quite useful QoL feature, and overall LGTM.

Issue #4057, opened 2022-03-05
pilot conversion warning occurs even with new pilots


Goober5000:
The dialog "This pilot was created with a different version of FreeSpace" appears even if you create a brand new pilot and try to launch with it.

This may have happened because I selected an old pilot before trying to create a new one.  Not 100% sure about that.

z64555:
Please give the referenced PR a look.  It's possible that the player t_plr within Pilot::verify() wasn't getting initialized properly and thus was using old data from a previous call.

Goober5000:
Fixed by #4058

PR #4058, opened 2022-03-05
Clean out player t_plr before use

z64555:
* Directly reference t_plr instead of player* p
Should fix issue #4057

### DIFF
--- a/code/pilotfile/plr.cpp
+++ b/code/pilotfile/plr.cpp
@@ -1215,8 +1215,8 @@ bool pilotfile::verify(const char *fname, int *rank, char *valid_language, int* 
 {
 	player t_plr;
 
-	// Ensure t_plr is cleanly init
-	t_plr.reset();
+	t_plr.reset();	// Ensure t_plr is cleanly init
+	p = &t_plr;		// Set player* so the rest of the file handler can work
 
 	filename = fname;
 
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
Was the `p = &t_plr;` line needed?  Because I fired up this branch and the first thing I got was a nullptr exception due to `this->p` being null here:
```
void pilotfile::plr_read_flags()
{
	// tips?
	p->tips = (int)handler->readUByte("tips");
```

z64555:
*loud grumbling*

Yes, apparently it is needed because of the way that the rest of the file handler works.  I thought I could get away with omitting the assignment since in that function, it appears moot to be using a pointer when you could be referencing the object direct instead.

notimaginative:
Why not just use a simple `p->reset();` and be done with it? Is there some reason that doesn't work?

z64555:
> Why not just use a simple `p->reset();` and be done with it? Is there some reason that doesn't work?

If you wanted to go the minimalist route, yeah, it would work.

I felt the changes to using the local `t_plr` were an opportunity to try to organize/clean up `verify()` since it seems to be a bit sloppy.  I don't particularly like the way it sets the `player* p` as part of its interface, and then use functions to modify its members hiding the fact that `p` can't be omitted.

Goober5000 (approved):
Ok, this new PR shows the dialog when expected and doesn't show it when not expected.

PR #4060, opened 2022-03-06
Fixes the version string in main.py

z64555:
This should now use the correct version string for both rc and release builds, which affects both Nebula publishing and the forum post title

### DIFF
--- a/ci/post/main.py
+++ b/ci/post/main.py
@@ -62,22 +62,41 @@ def _match_version_number(text, regex):
 	return int(match.group(1))
 	
 
-def get_source_version(date_version):
+def get_source_version(date_version: datetime, tag_name: str) -> semantic_version.Version:
 	"""! Retrieves the build's version from `version.cmake`, forms it as a string, and appends the date
 
 	@param[in] `date_version` Date this build is published for release
+	@param[in] 'tag_name'     Tag of this release, used to determine if release, rc, or nightly
 
-	@return String of the form `MAJOR_VERSION.MINOR_VERSION.BUILD_VERSION-date`
+	@return String of the form `MAJOR_VERSION.MINOR_VERSION.BUILD_VERSION-date`, ' MAJOR_VERS
 	"""
-	
+	major = minor = build = version = 0
 	with open(os.path.join("..", "..", "cmake", "version.cmake"), "r") as f:
 		filetext = f.read()
-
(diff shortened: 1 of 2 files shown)


ci/post/main.py
```diff
@@ -62,22 +62,41 @@ def _match_version_number(text, regex):
...
 
-def get_source_version(date_version):
+def get_source_version(date_version: datetime, tag_name: str) -> semantic_version.Version:
 	"""! Retrieves the build's version from `version.cmake`, forms it as a string, and appends the date
 
 	@param[in] `date_version` Date this build is published for release
+	@param[in] 'tag_name'     Tag of this release, used to determine if release, rc, or nightly
 
-	@return String of the form `MAJOR_VERSION.MINOR_VERSION.BUILD_VERSION-date`
+	@return String of the form `MAJOR_VERSION.MINOR_VERSION.BUILD_VERSION-date`, ' MAJOR_VERS
```
Goober5000, line 71: Is `MAJOR_VERS` supposed to be `MAJOR_VERSION`?

ci/post/main.py
```diff
@@ -62,22 +62,41 @@ def _match_version_number(text, regex):
...
 	with open(os.path.join("..", "..", "cmake", "version.cmake"), "r") as f:
 		filetext = f.read()
-
 		major = _match_version_number(filetext, MAJOR_VERSION_PATTERN)
 		minor = _match_version_number(filetext, MINOR_VERSION_PATTERN)
 		build = _match_version_number(filetext, BUILD_VERSION_PATTERN)
 
-		return semantic_version.Version("{}.{}.{}-{}".format(major, minor, build, date_version))
+	if "rc" in tag_name.lower():
+		# Release canidates retain an internal version unstable, but external version is the tagname
```
Goober5000, line 81: canidates -> candidates

also what does "an internal version unstable" mean?

ci/post/main.py
z64555: blah, I got interrupted and forgot that I was mid-sentence.  Its supposed to document the three variations of version string it can produce

ci/post/main.py
z64555: This is supposed to document that release candidates still maintain their unstable/nightly revision scheme, but the .exe name uses the <major>.<minor>.<build>-<RCn> form

Goober5000 (approved):

PR #4061, opened 2022-03-06
fix support ships breaking in in-mission jumps

Goober5000:
If a support ship's repair goal is removed, abort the rearm.  There is extra cleanup that needs to be done for this goal only.

### DIFF
--- a/code/ai/aigoals.cpp
+++ b/code/ai/aigoals.cpp
@@ -266,13 +266,19 @@ void ai_remove_ship_goal( ai_info *aip, int index )
 	// ai goal code look
 	Assert ( index >= 0 );			// must have a valid goal
 
+	if (index == aip->active_goal)
+	{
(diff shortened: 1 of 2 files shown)


code/ai/aigoals.cpp
```diff
@@ -266,14 +266,20 @@ void ai_remove_ship_goal( ai_info *aip, int index )
 	// ai goal code look
 	Assert ( index >= 0 );			// must have a valid goal
 
+	if (index == aip->active_goal)
```
z64555, line 269: Any particular reason to move this up here?

Goober5000: The `goals[index].ai_mode` check needs to be up here, and consequently I thought there was no need to have more than one `(index == aip->active_goal)` check.

z64555: 👍

z64555 (approved):

PR #4069, opened 2022-03-11
Remove new assert in `model_instance_local_to_global_dir`

Baezon:
Consider this approach tentative, but *apparently* model collide info does not necessarily fill `hit_normal` even when hits are registered. This happens in the case of a sphere hitting the 'edge' of a polygon (although I'm not convinced a hit normal cant be generated in this case, but this would require a deeper dive), which, at least in my tests, seems to happen a lot with maras and beams for whatever reason. 

Other parts of the code even explicitly check for edge hits and don't check normals.
https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/object/collideshipship.cpp#L336

And, at least for beams, this hit normal is passed directly to the relevant function.
https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/weapon/beam.cpp#L3689

Apparently before, it would simply operate on a null vector, which for this math is at least safe, it simply remains null. Whether perhaps the function itself or its calls should be changed to something else more reasonable, I'll leave to discussion, but this is the simplest approach.

### DIFF
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -1099,7 +1099,7 @@ typedef struct mc_info {
 	int		hit_bitmap;			// Which texture got hit.  -1 if not a textured poly
 	float		hit_u, hit_v;		// Where on hit_bitmap the ray hit.  Invalid if hit_bitmap < 0
 	int		shield_hit_tri;	// Which triangle on the shield got hit or -1 if none
-	vec3d	hit_normal;			//	Vector normal of polygon of collision.  (This is in submodel RF)
+	vec3d	hit_normal;			//	Vector normal of polygon of collision (This is in submodel RF). CAN BE ZERO, if edge_hit is true
 	bool		edge_hit;			// Set if an edge got hit.  Only valid if MC_CHECK_THICK is set.	
 	ubyte		*f_poly;				// pointer to flat poly where we intersected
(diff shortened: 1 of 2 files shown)


BMagnu:
So, removing the assert from the function (and it's siblings) seems sensible.
There are cases after all, where you might want location-independent vector rotation based on submodel rotations, but don't actually have a normalized vector. The math _should_ be safe and basically always return a vector of the same magnitude as the input. It's the callers job to make sure that is sensible.

That hit_normal can be 0,0,0 though is otherwise concerning as well.
Not sure how much it affects beams, but if that can happen with actual collisions, the physics code won't properly calculate momentum transfer, and the smaller ship will likely tunnel through the larger one as a result.

BMagnu (changes requested):
As noted, likely sensible, but if you remove that assert, please also remove it from its sibling functions

Goober5000:
I wonder if this is a reason for the occasional "Null vector in vm_vec_normalize, trace out and fix!" message.

(Or was a reason.  I no longer see that message in the code.)

Goober5000 (approved):
Looks ok, and the reasoning is sound.  I should also point out that this Assert was the reason for PR #4038

BMagnu (approved):

PR #4071, opened 2022-03-13
fix off-by-one error in text splitting logic

Goober5000:
`gr_get_string_size` takes a length parameter, but `split_str_once` was calling it with the position instead.  This led to incorrect results occasionally.  If the incorrect pixel width had the bad luck to be exactly equal to the remaining space, this would cause an infinite loop.  This fixes the bug and adds a bunch of safeguards to prevent a similar situation in the future.

### DIFF
--- a/code/graphics/software/font.cpp
+++ b/code/graphics/software/font.cpp
@@ -688,7 +688,8 @@ void gr_get_string_size(int *w1, int *h1, const char *text, int len)
 	float w = 0.0f;
 	float h = 0.0f;
 
-	FontManager::getCurrentFont()->getStringSize(text, static_cast<size_t>(len), -1, &w, &h);
+	if (len != 0)
+		FontManager::getCurrentFont()->getStringSize(text, static_cast<size_t>(len), -1, &w, &h);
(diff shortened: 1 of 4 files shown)


code/graphics/software/font.cpp
```diff
@@ -688,7 +688,8 @@ void gr_get_string_size(int *w1, int *h1, const char *text, int len)
 	float w = 0.0f;
 	float h = 0.0f;
 
-	FontManager::getCurrentFont()->getStringSize(text, static_cast<size_t>(len), -1, &w, &h);
+	if (len != 0)
```
z64555, line 691: `getStringSize()` Should have its own safety for when it detects a length <= 0, whats happening there that requires this additional safety?

Goober5000: It's not "required" per se, it's just a small optimization.

jg18 (approved):
A small suggestion but overall a nice catch. :+1:

I also checked other calls to `gr_set_string_size()` that use the extra parameter and they all look OK.

PR #4072, opened 2022-03-13
prevent decals from being used on invalid objects

Goober5000:


code/decals/decals.cpp
```diff
@@ -270,6 +273,8 @@ struct Decal {
 			if (smi->blown_off) {
 				return false;
 			}
+		} else {
+			Assertion(false, "Only ships are currently supported for decals!");
```
jg18, line 277: Shouldn't we return `false` here as well? What will happen in release builds without it?

jg18 (changes requested):

code/decals/decals.cpp
Goober5000: Good point

jg18 (approved):

PR #4074, opened 2022-03-15
some debug console updates

Goober5000:
Tweak the help for the 'bastion' command and change 'dcf_fov' to 'fov'.  Also add a new 'mainhall' command that can be used in the main hall, not just the player select screen.  And for good measure, makes main hall names case-insensitive.

### DIFF
--- a/code/menuui/playermenu.cpp
+++ b/code/menuui/playermenu.cpp
@@ -1328,7 +1328,8 @@ void player_select_cancel_create()
 	Player_select_autoaccept = 0;
 }
 
-DCF(bastion,"Sets the player to be on the bastion (or any other main hall)")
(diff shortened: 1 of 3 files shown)


code/menuui/playermenu.cpp
```diff
@@ -1328,7 +1328,8 @@ void player_select_cancel_create()
 	Player_select_autoaccept = 0;
 }
 
-DCF(bastion,"Sets the player to be on the bastion (or any other main hall)")
+// See also the mainhall command in mainhallmenu.cpp
```
z64555, line 1331: Recommend a "see also" in the help text

z64555 (approved):

z64555:
Went ahead an merged, this is a small QoL change that shouldn't affect players.

PR #4083, opened 2022-03-19
small XSTR qol improvement

Goober5000:
Allow `XSTR` to be used by itself, without implying an `XSTR()` tag.

### DIFF
--- a/code/localization/localize.cpp
+++ b/code/localization/localize.cpp
@@ -744,13 +744,14 @@ bool lcl_ext_localize_sub(const char *in, char *text_str, char *out, size_t max_
 			break;
 		ch += 4;
 
-		attempted_xstr = true;
-
 		// the next non-whitespace char should be a (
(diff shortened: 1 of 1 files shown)


jg18:
Just wondering what prompted the creation of this PR? Did a mod want to use the string literal `"XSTR"` in player-facing text?

Goober5000:
> Just wondering what prompted/motivated this PR? Did a mod want to use the string literal `"XSTR"` in player-facing text?

Yep, specifically a scripted sexp I'm writing for Ertanax:

```
$Parameter:
	+Description: Name of an intel entry
	+Type: string
$Parameter:
	+Description: XSTR ID of the intel entry, or -1 if there is no XSTR for it.
	+Type: number
```

jg18 (approved):
Took a little time to understand what the code is doing and what effect these changes have, but now, it makes sense.

PR #4087, opened 2022-03-21
Add delay into ai_new_maybe_reposition_attack_subsys for performance 

EatThePath:
Profiling found ai_new_maybe_reposition_attack_subsys accounting for around 13% of runtime in crowded missions. This adds a 1 second delay between checking new points, which takes it to 0% of runtime in those same missions. I have not been able to discern a visible change in ship behavior.

### DIFF
--- a/code/ai/ai.h
+++ b/code/ai/ai.h
@@ -267,6 +267,8 @@ extern pnode	*Ppfp;			//	Free pointer in path points.
 // Goober5000 (based on the "you can only remember 7 things in short-term memory" assumption)
 #define MAX_IGNORE_NEW_OBJECTS	7
 
+#define AI_DYNAMIC_PATH_RECHECK_DELAY 1*1000
+
 typedef struct ai_info {
 	flagset<AI::AI_Flags> ai_flags;				//	Special flags for AI behavior.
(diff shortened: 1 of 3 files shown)


Baezon (approved):
Maybe should wait, though. Probably don't want to accidentally introduce bugs this late in the game.

EatThePath:
I have no strong opinion on waiting or not, for my part.

wookieejedi:
When the time comes would it be useful to make this an AI profiles flag like the turret check timer?

EatThePath:
> When the time comes would it be useful to make this an AI profiles flag like the turret check timer?

I'd prefer to reserve that till someone can produce a plausible case where the current delay causes issues, to avoid unnecessary complexity. I think it'd have to be a *very* exotic mod design for it to come up.

Goober5000:
I searched the AI code for similar situations and found that `maybe_recreate_path` also sets a timestamp of 1 second, so there is precedent.

Goober5000 (approved):

EatThePath:
A note to anyone looking this up in the future: The check to see if the target is in sight so the AI can start turning for an actual attack is _outside_ the delay and is still run every frame, which should minimize any appearance of sluggish response.

PR #4088, opened 2022-03-21
Remove unneeded search of whole sexp node array from free_sexp

EatThePath:
Profiling shows that free_sexp contributes to significant runtime consumption in mods with heavy use of sexps in scripts, especially in missions with lots of sexp nodes. This change removes the need for that looping, cutting about half of the runtime from those sexps from scripts. Given some testing in Blighted and Warmachine with no evident breakage of either scripts or conventional mission logic.


wookieejedi (approved):
Nothing breaks in my tests.

PR #4090, opened 2022-03-22
FRED Nebula background UI enhancements

Goober5000:
Two FRED enhancements to the background editor:

1) Support the full number of nebula poofs using a multi-select list
2) Add the ability to edit nebula fog colors, since that previously was not part of the FRED UI

Also add save support for nebula fog colors, since that previously did not save properly.


wookieejedi (approved):
Overall looks good to me. The failsafe for a default nebula color seems like a good RGB color, too.

Goober5000:
The default nebula color is from Volition, so that's good to hear. :)

PR #4092, opened 2022-03-23
Fix edge case wrongly updating moveable

BMagnu:
In certain cases, moveables were not correctly updated and reset.
This simplifies how it's done and relies on the existing animation code, ensuring that the value used is always the canonical value.


Baezon:
Fwiw, this fixes some issues I came across in my test case.

Goober5000 (approved):

PR #4097, opened 2022-03-25
Added error handling to map accesses in contronconfigdefaults parsing

### LINKED ISSUE #4041: ControlsConfigDefaults.tbl has no safeties for errant data
This is due largely to the quirks of std::map::operator[].  As a reminder, when a key is not found in a std::map with operator[] a new key-value pair is created with the default value of the value type (for standard data types such as int, float, etc, the default value is 0), this behavior is to allow easy insertions of new keys using the `map[new_key] = new_value;` syntax;

This does however historacally causes problems when you want to use operator[] to access the keys, and *dont* want a new key-value pair created, which is the case here.

Specifically, `$Key Default:`, `$Category:`, and `$Type:` don't have a safegaurd against bad strings passed to them, and will do a silent acceptance and default the values to 0, which is not desirable.

EatThePath:
Fixes #4041

### DIFF
--- a/code/controlconfig/controlsconfigcommon.cpp
+++ b/code/controlconfig/controlsconfigcommon.cpp
@@ -9,6 +9,7 @@
 
 #include <cstdio>
 #include <cstdarg>
+#include <exception>
 #include <string>
 
 #include "cfile/cfile.h"
@@ -16,12 +17,14 @@
 #include "controlconfig/presets.h"
 #include "debugconsole/console.h"
 #include "def_files/def_files.h"
+#include "globalincs/pstypes.h"
 #include "globalincs/systemvars.h"
 #include "io/joy.h"
 #include "io/key.h"
 #include "io/mouse.h"
 #include "localization/localize.h"
 #include "options/Option.h"
+#include "osapi/dialogs.h"
 #include "parse/parselo.h"
 
 #include <map>
@@ -1607,7 +1610,12 @@ void control_config_common_read_section(int s, bool first_override) {
 				key = -1;
 			} else {
 				stuff_string(szTempBuffer, F_NAME);
-				key = mKeyNameToVal[szTempBuffer];
+				try{
+					key = mKeyNameToVal.at(szTempBuffer);
+				}
+				catch(std::out_of_range&){
+					Warning(LOCATION,"Table tried to set default for a key that doesn't exist: %s",szTempBuffer.c_str());
+				}
 			}
 		}
 
@@ -1640,7 +1648,12 @@ void control_config_common_read_section(int s, bool first_override) {
 			// Config menu options
 			if (optional_string("$Category:")) {
 				stuff_string(szTempBuffer, F_NAME);
-				item->tab = mCCTabNameToVal[szTempBuffer];
+				try{
+					item->tab = mCCTabNameToVal.at(szTempBuffer);
+				}
+				catch(std::out_of_range&){
+					Warning(LOCATION,"Table refered to a key category that doesn't exist: %s",szTempBuffer.c_str());
+				}
 			}
 
 			if (optional_string("$Text:")) {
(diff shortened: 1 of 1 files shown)


code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1640,7 +1648,12 @@ void control_config_common_read_section(int s, bool first_override) {
 			// Config menu options
 			if (optional_string("$Category:")) {
 				stuff_string(szTempBuffer, F_NAME);
-				item->tab = mCCTabNameToVal[szTempBuffer];
+				try{
+					item->tab = mCCTabNameToVal.at(szTempBuffer);
+				}
+				catch(std::out_of_range&){
+					Warning(LOCATION,"Table refered to a key category that doesn't exist: %s",szTempBuffer.c_str());
```
Goober5000, line 1655: "referred" is misspelled

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1655,7 +1668,12 @@ void control_config_common_read_section(int s, bool first_override) {
 
 			if (optional_string("$Type:")) {
 				stuff_string(szTempBuffer, F_NAME);
-				item->type = mCCTypeNameToVal[szTempBuffer];
+				try{
+					item->type = mCCTypeNameToVal.at(szTempBuffer);
+				}
+				catch(std::out_of_range&){
+					Warning(LOCATION,"Table refered to a key type that doesn't exist: %s",szTempBuffer.c_str());
```
Goober5000, line 1675: here too

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1607,7 +1610,12 @@ void control_config_common_read_section(int s, bool first_override) {
 				key = -1;
 			} else {
 				stuff_string(szTempBuffer, F_NAME);
-				key = mKeyNameToVal[szTempBuffer];
+				try{
```
z64555, line 1613: Please use the following brace style, it seems to be the predominant style in this .cpp
```
try {
    // code
} catch {
    // code
}
```

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1607,7 +1610,12 @@ void control_config_common_read_section(int s, bool first_override) {
...
+					key = mKeyNameToVal.at(szTempBuffer);
+				}
+				catch(std::out_of_range&){
+					Warning(LOCATION,"Table tried to set default for a key that doesn't exist: %s",szTempBuffer.c_str());
```
z64555, line 1617: Please reference the .tbl option in the warning message.  This should give mod devs a better clue of whats going wrong.

ex:
"Table tried to set \'$Key Default\' for a key that doesn't exist: %s"

(If you can come up with a better looking style then by all means use it :smile:)

code/controlconfig/controlsconfigcommon.cpp
z64555: Please reference the .tbl option in the warning message.

code/controlconfig/controlsconfigcommon.cpp
z64555: Please reference the .tbl option in the warning message.

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1640,7 +1648,12 @@ void control_config_common_read_section(int s, bool first_override) {
 			// Config menu options
 			if (optional_string("$Category:")) {
 				stuff_string(szTempBuffer, F_NAME);
-				item->tab = mCCTabNameToVal[szTempBuffer];
+				try{
```
z64555, line 1651: Please use the predominant brace style.

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1655,7 +1668,12 @@ void control_config_common_read_section(int s, bool first_override) {
 
 			if (optional_string("$Type:")) {
 				stuff_string(szTempBuffer, F_NAME);
-				item->type = mCCTypeNameToVal[szTempBuffer];
+				try{
```
z64555, line 1671: Please use the predominant brace style.

z64555 (changes requested):

z64555 (approved):

z64555:
Now then... whats the deal with gcc failing?

Goober5000 (changes requested):
Formatting can be improved: a single space between `catch` and the parenthesis, and a single space after each comma.  Also there is an extra single space after the opening quote for Category.

I'm not a fan of the wording - it feels awkward.  You could say "'$Key Default:' entry not found: %s", etc., for each message.  Or use the wording @z64555 suggested.

Goober5000 (approved):
Good enough. 👍 

(missed the spaces after the commas, but I'll allow it)

PR #4100, opened 2022-03-26
Make sure obj_udpate packet

JohnAFernandez:
This allows the object update and rollback packets to use Missiontime instead of timestamp(), since the local timestamp() is not updated when option menus are added.  This should also begin the upgrade to the interpolation code that should get shuddering out of multiplayer missions. 

Needs testing.


Goober5000:
It looks like your branch is either missing some old commits, or reverts them.  I see rollbacks to the timestamp code and the model code in here.

JohnAFernandez:
Not sure what arcane git command caused this.  I may need to submit a new pr.

JohnAFernandez:
Ok.  This is now replaced.

PR #4102, opened 2022-03-26
fix a bunch of FRED dialog problems

Goober5000:
1) Fixes the ship and wing dialog resizing for both sexp help and arrival/departure cues
2) Restores the sexp help in the Mission Goals dialog, which had been broken since the mini-help-box was added


z64555 (approved):
LGTM

PR #4104, opened 2022-03-27
Fix LUA Steering Control Override

wookieejedi:
I've been using `LUA_FULL_CONTROLS` in the updated mouse-as-a-joystick script, and it works well, but there's one last small bug that just was found that I'd like to fix. `LUA_FULL_CONTROLS` affects things that beyond just steering values, which has the unfortunate side-effect of also overriding non-steering controls such as cruise percent. It would be less intrusive to use the `LUA_STEERING_CONTROLS`. Unfortunately, `LUA_STEERING_CONTROLS` just sets everything to 0, which I think is a code issue bug that has been around since the LUA controls were added in 2010 here:
(https://github.com/scp-fs2open/fs2open.github.com/commit/2a69aaa23ba0def3dbfb9f566706e7130a33866f).

This PR fixes usage of `LUA_STEERING_CONTROLS`, and in testing everything works as expected (including fixing the cruise percent issue). Also, I do not think there are any scripts that use `LUA_STEERING_CONTROLS`, but if desired I can make a new version called `LUA_STEERING_CONTROLS_FIXED` or something to that effect.

### DIFF
--- a/code/playerman/playercontrol.cpp
+++ b/code/playerman/playercontrol.cpp
@@ -916,31 +916,45 @@ void read_keyboard_controls( control_info * ci, float frame_time, physics_info *
 	}
 }
 
-void copy_control_info(control_info *dest_ci, control_info *src_ci)
+void copy_control_info(control_info *dest_ci, control_info *src_ci, int control_copy_method)
 {
-	if (dest_ci == NULL)
+	if (dest_ci == nullptr)
 		return;
 
-	if (src_ci == NULL) {
-		dest_ci->pitch = 0.0f;
-		dest_ci->vertical = 0.0f;
-		dest_ci->heading = 0.0f;
-		dest_ci->sideways = 0.0f;
-		dest_ci->bank = 0.0f;
-		dest_ci->forward = 0.0f;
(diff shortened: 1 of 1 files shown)


code/playerman/playercontrol.cpp
```diff
@@ -916,31 +916,45 @@ void read_keyboard_controls( control_info * ci, float frame_time, physics_info *
...
+		if (src_ci == nullptr) {
+			dest_ci->pitch = 0.0f;
+			dest_ci->heading = 0.0f;
+			dest_ci->bank = 0.0f;
+		} else {
+			dest_ci->pitch = src_ci->pitch;
+			dest_ci->heading = src_ci->heading;
+			dest_ci->bank = src_ci->bank;
+		}
+	} else if (control_copy_method & LGC_FULL) {
```
Goober5000, line 935: It looks like `control_copy_method` is a bitfield, which means it would theoretically be possible for *both* LGC_ flags to be set, wouldn't it?  In which case, the LGC_FULL check should come first.

wookieejedi: Good catch, just updated.

Goober5000 (approved):

PR #4110, opened 2022-03-28
Don't ignore user-set LDFLAGS

freijon:
Currently LDFLAGS set by the user are ignored and the LINKER_FLAGS are hard-coded. This PR uses the ENV LDFLAGS if set, otherwise same behavior as before.


cmake/toolchain-gcc.cmake
```diff
@@ -27,6 +27,11 @@ endif()
 set(COMPILER_FLAGS "")
 set(LINKER_FLAGS "")
 
+# Don't ignore user-set LDFLAGS
+if(DEFINED ENV{LDFLAGS})
```
asarium, line 31: Could you also replicate this in [toolchain-clang.cmake](https://github.com/scp-fs2open/fs2open.github.com/blob/master/cmake/toolchain-clang.cmake)?
Then this would cover both compilers.

asarium (changes requested):

asarium (approved):

PR #4111, opened 2022-03-28
Fred object orientation features

Goober5000:
1) Add ability to rotate objects relative to other objects in the Object Editor
2) Add dialog to calculate relative distance and orientation between ships
3) Add flag to point ships while incorporating the ship's uvec


wookieejedi (approved):
FRED side of things looks good, will defer to @Baezon for the angle/rotations code.

Baezon (approved):
LGTM

Issue #4099, opened 2022-03-26
Multi cannot use timestamp() but uses it anyway


JohnAFernandez:
A minor issue.  Timestamp() is unreliable in multi because of a workaround to allow menus to work while the game is unpaused in multi (so that players do not de-sync).  All multi functions that use timestamp() should probably use Missiontime instead.  This would necessitate a multi bump.  The main consequence of this is that players would likely be unable to update their position to the server if the player is in a menu.

Should Missiontime be used or some other solution be found?

z64555:
Gonna need a map of what all uses Missiontime and Timestamp first.

JohnAFernandez:
The most affected things are in multi_obj.cpp .  The object update packet was using the timestamp as a way to track how far ahead or behind an incoming packet is to the local simulation.  There is maybe one other packet that uses it.

Goober5000:
Yeah, Missiontime sounds plausible, but I too would want to see how exactly timestamp is used in the multi code first.

It might be possible to replace some timestamps with `UI_TIMESTAMP` which designed for user-experience things like blinking lights, and is not affected by time compression or pausing.

notimaginative:
`UI_TIMESTAMP` is relative to the system timer, correct? That's what it looks like anyway. And if that's the case then the values are only useful to the local system. For multi it would have to be a time which could be relatively the same across systems, which was previously the timestamp.

And I think that `Missiontime` is directly based on the timestamp now so I'm assuming it's not going to give you different results than using `timestamp()` directly.

Goober can correct me on those two points if I'm off base and don't remember/understand the code correctly.

We may just need to bite the bullet and implement a timer sync internal to multi. Have a value that's sync'd during join and is incremented by the system timer, and perhaps re-sync'd every few minutes.

Goober5000:
Missed the follow up earlier.  `UI_TIMESTAMP` doesn't pause and doesn't use time compression, but aside from that, both it and the regular timestamps use the system timer.

`Missiontime` is, indeed, directly based on the timestamp now.

Theoretically, a timer sync could be implemented by making timestamps dependent on the server when in multiplayer mode.  Either the server could set the value directly, or the server could add or subtract time if the clients got too different.

notimaginative:
[...]

PR #4114, opened 2022-03-29
Fix unreliable timestamps

JohnAFernandez:
This allows the object update and rollback packets to use Missiontime instead of timestamp(), since the local timestamp() is not updated when option menus are accessed. This should also begin the upgrade to the interpolation code that should get shuddering out of multiplayer missions.

Needs testing, requires a multi bump, fixes https://github.com/scp-fs2open/fs2open.github.com/issues/4099 .


JohnAFernandez:
Ok.... Not sure where that extra stuff came from in commit 1.  But I'll remove it ASAP

JohnAFernandez:
Not sure this is the right approach, any longer.

PR #4117, opened 2022-03-29
dynamic IFF list in the ship select editor

Goober5000:
Allow all IFFs to be listed, even when the number exceeds 10.  Follow-up to #4116.  Related to #4090 as it uses the same multi-selection method.


wookieejedi (approved):
Robust fix, and looks good overall.

PR #4118, opened 2022-03-30
add a class encapsulation for Volition's linked list

Goober5000:
This should allow collections which use the Volition linked list (for example `obj_used_list`) to be iterated using range-based `for` loops.

### DIFF
--- a/code/globalincs/linklist.h
+++ b/code/globalincs/linklist.h
@@ -7,7 +7,7 @@
  *
 */ 
 
-
+#include <utility>	// for std::move
 
 #ifndef _LINKLIST_H
 #define _LINKLIST_H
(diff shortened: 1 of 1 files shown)


code/globalincs/linklist.h
```diff
@@ -83,4 +83,87 @@ do {												\
...
 
+template <class T>
+class volition_linked_list
+{
+	T* head;
+
+public:
+	class iterator
+	{
+		T* ptr;
```
asarium, line 94: This can be changed to `const T*` since this class (and the iterator) does not modify the pointer value.

code/globalincs/linklist.h
```diff
@@ -83,4 +83,87 @@ do {												\
...
+			return ptr == rhs.ptr;
+
+		}
+
+		bool operator!=(const iterator& rhs) const
+		{
+			return ptr != rhs.ptr;
+		}
+
+		T&& operator*()
```
asarium, line 142: I would suggest returning a `const T&` here. Moving the pointer value may change the list we are iterating on.

asarium (changes requested):

code/globalincs/linklist.h
Goober5000: updated with a `const_iterator`

code/globalincs/linklist.h
Goober5000: updated with a `const_iterator`

Goober5000:
Even with these latest changes, when I try
```
for (auto ptr: list_range(&obj_used_list))
```
Visual Studio displays this message:
inaccessible

asarium (approved):

Goober5000:
Then again, I could return `ptr` from `operator*` instead of `*ptr`...

Goober5000:
Yep, that was the solution.

Issue #4120, opened 2022-03-30
Distance displayed in the targeting Brackets and the target monitor are not consistent


AdmiralNelsonFSO:
The targeting brackets and the target monitor do not display the same values for distance, as seen in this video: https://youtu.be/3Dah1mMwIEM

Testing shows me that the distance shown in the target monitor changes as I target different subsystems, whereas the distance displayed in the brackets does not change.  The HUD should be consistent; the targeting brackets should also display distance to subsystem, or both should display distance to bounding box or center.

chief1983:
Seems like knowing this allows you to get more information at one time than you could with both of those values being the same.  Perhaps this is by design?

AdmiralNelsonFSO:
But what distance do the targeting brackets actually display?  Is it the same as the distance SEXP? Without documentation, we can't tell the player which distance value to use or what the numbers shown mean.

chief1983:
Does the targeting brackets number change when you first target a subsystem?  If not I would assume it's displaying the overall object distance, I assume the center of the bounding box.

AdmiralNelsonFSO:
They don't AFAICS.  We should at least document which of the different distance SEXPs shows the same value as the targeting brackets.  For instance, in this mission the player is being told to maintain a certain distance from the target.  The correct distance SEXP that corresponds to the targeting brackets needs to be used so that the player can be told to look at the distance value in the brackets and have it be correct.

EatThePath:
Honestly, the status quo seems like neither a bug, confusing, or a problem. The difference seems quite intuitive. The corner target view in the corner doesn't change and frames the whole ship, so the distance it shows is the distance to what it shows, the ship. The camera's focus never changes, so the distance shouldn't change. The target brackets on the other hand don't frame the ship, they frame the current subsystem target. You want them to change because that distance is important, and having it on the brackets means it's right at hand when you need it most.

As a player I would assume directives or objectives are talking about distance to the ship unless they refer to a specific turret or subsystem. Otherwise I'd be able to change the mission flow by changing which subsystem I'm targeting. And as a player I need distance to subsystem for combat reasons and potentially need distance to ship if I ever have an objective about that. It's important in fact that the HUD not be consistent, because both pieces of information are important and there's probably missions out there needing both.

[...]

PR #4121, opened 2022-04-01
make HUD target distance consistent

Goober5000:
Several careful tweaks to improve handling of distance as measured on the HUD.

1) Use `Player_ai->current_target_distance` for rendered HUD gauges which display the distance, because the distance is calculated in `hud_update_frame` (which is called from `game_simulation_frame`) and is used for a couple of these gauges already.  This change makes it all consistent.
1b) Some gauges which depend on distance but do not actually display it (e.g. the lead indicator) still calculate their distance immediately before using it.  I didn't change these because a) the calculation is still correct, even if redundant, and b) there is a small possibility of `hud_update_frame` changing the target after the distance is calculated, so these gauges are guaranteed to have the correct calculation.

2) On discovering that the `evaluate_ship_as_closest_target` specifically switched from bbox-distance to center-distance, I changed all "target closest X" functions to consistently evaluate center-distance.  I believe Volition did this to prioritize small fighters when dogfighting next to large ships.

3) Moved `hud_target_change_check` below the computation of `Player_ai->current_target_distance` because it uses the computed distance to determine the acceleration trend.  This way it will use the distance of the current frame, not the previous one.

4) Minor move of a `Target_static_looping` check into `hud_target_change_check` because it made sense.

Addresses #4120.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -1305,7 +1305,9 @@ ship_obj *advance_ship(ship_obj *so, int next_flag)
 static object* select_next_target_by_distance(const bool targeting_from_closest_to_farthest, const int valid_team_mask, const int attacked_object_number = -1, flagset<Ship::Info_Flags>* target_filters = NULL) {
     object *minimum_object_ptr, *maximum_object_ptr, *nearest_object_ptr;
     minimum_object_ptr = maximum_object_ptr = nearest_object_ptr = NULL;
-    float current_distance = hud_find_target_distance(&Objects[Player_ai->target_objnum], Player_obj);
+    // we need to use the same distance method as evaluate_ship_as_closest_target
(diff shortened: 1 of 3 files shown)


code/hud/hudtarget.cpp
```diff
@@ -1360,7 +1362,9 @@ static object* select_next_target_by_distance(const bool targeting_from_closest_
                 continue;
             }
 
-            new_distance = hud_find_target_distance(prospective_victim_ptr, Player_obj);
+			// we need to use the same distance method as evaluate_ship_as_closest_target
```
wookieejedi, line 1365: this is very nitpicky, but it looks like there is an extra whitespace tab?

Goober5000: It turns out this function indented using spaces whereas I used tabs.  I changed my indentation in this function.

wookieejedi (approved):
Logical changes and overall looks good.

PR #4126, opened 2022-04-02
Fix auto-trigger for post-release build workflow, Checkout passed release tag, Fix version string

z64555:
Auto triggers were failing because the `linux_result` and `windows_result` variables were expected by the .yaml to be bools but were in fact strings.  This fixes that assumption to be the correct type.

Changed checkout of current master HEAD to the passed release tag, thus allowing the script to be run on only the intended release data instead of accidentally including new merges after the release tag.

Fixes the Nebula and forum post version strings to correctly pull version string data from `version_override.cmake` and `version.cmake`, instead of trying to fudge it from the tag_name or incorrectly from the `version.cmake` alone.

Also Adds a `changelog.py` script to create a `change.log` file.  This grabs the titles of merged PR's since a given date passed as an argument of the form of `YYYY-MM-DD`.  This can be manually run but be aware there's a limit to the number of HTTP GET requests you can do from Github (This script doesn't authorize yet)

The `changelog.py` is still a WIP, but it can be manually run for now.  In the future it'll be used to auto-generate the change.log for the forum post and the `main.py` script will insert as much of the log into the post.  Since the forums have a character limit per post, a constant was added and the intended use is to truncate the changelog text once the limit is reached.

### DIFF
--- a/ci/post/changelog.py
+++ b/ci/post/changelog.py
@@ -1,0 +1,156 @@
+# Retrives a log of commit messages and attempts to pretty them into a human-readbile format
+
+import datetime
+import os
+import re
+import requests
+import sys
+
+from datetime import datetime
+from subprocess import check_output
+
+MAX_PAGES = 100  # maximum number of PR pages to go through
+warnings = 0
+
+def get_last_page(response: requests.Response) -> int:
+    """
+    @brief Extract the last page number from the response header
+
+    @param[in] response The response from a GET call
+
+    @returns The page number of the last page, or MAX_PAGES
+    """
+
+    # The link header is a heckn dictionary in string form, and for that matter the
+    # value is in front of the key.  It is of the form '<<url1>>; rel=<key1>, <url2>, <key2>'
+    link = response.headers['link']
+    link = link.split(',')
+    for x in link:
+        y = x.split(';')        # <url>; rel=<key>
+        if "last" not in y[1]:  # look for rel='last'
+            continue
+        # search in the url for `page=`` and grab the int immediately after
+        # the int may be delimited by '&' or '>'
+        str = y[0][y[0].find('page=') : ]   # slice everything out before 'page='
+        str = str.split('&', 1)[0]          # slice out any possible queries after page=<int>
+        last_page = int( re.sub('\D', "", str) )    # strip out just the page number, as integer
+        if last_page > MAX_PAGES:
+            return MAX_PAGES
+        else:
+            return last_page
+    
+
+def main():
+    os.chdir(os.path.dirname(__file__))	# Change working directory to this file's directory
+    
+    start_date = datetime.strptime(sys.argv[1], "%Y-%m-%d")
+
+    print("start_date: {}".format(start_date))
+
+    # Retrive the git log for master, since start date, order by date, in ascending order, use YYYY-MM-DD date format,
+    #   each line is of the form "<author short date> | <message>", only grab commits from master,
+    #   output to `fsofordigest.txt`,
+    
+    # check_output(("git log master", "--after=2021-09-15", "--date-order", "--reverse", "--date=short",
+    # "--pretty=\"%as | %s\"",  "--first-parent",
+    # "--output fsofordigest.txt"), text=True)
+
+    # The commit messages are either of the form "<body> (#<PR_number>)" or "Merge pull request #<PR_number> from <source_repo>"
+    # For our purposes, we want them be all of the form "<body> (#<PR_number>)"
+    # Typically, the second form does not have a usable body, so we'll have to pull the PR title from github
+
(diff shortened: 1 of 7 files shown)


ci/post/changelog.py
```diff
@@ -0,0 +1,158 @@
...
+    #   "number: <int>"     # The PR id number
+    #   "title: <string>"   # Title of the PR
+    #   "body: <string>"    # The body paragraph detailing the PR
+    #   "labels: [label_object]"    # All labels attached to this PR
+
+    changelog = {}
+    url = "https://api.github.com/repos/scp-fs2open/fs2open.github.com/pulls"
+    page = 1
+    last_page = MAX_PAGES
+    while (page <= last_page):
```
ngld, line 80: You could use `for page in range(1, last_page + 1):` instead if you remove the `page = 1` and `page += 1` lines. It's not a huge difference but IMO a bit clearer.

ci/post/main.py
```diff
@@ -145,12 +155,29 @@ def main():
...
+
+	# Populate version and date
 	tag_name = os.environ["RELEASE_TAG"]	##!< commit tag string
 	date = datetime.now()	##!< current date
 	version = get_source_version(date.strftime(DATEFORMAT_VERSION), tag_name)	##!< form full version string
-	success = os.environ["LINUX_RESULT"] == "success" and os.environ["WINDOWS_RESULT"] == "success"	##!< true if both linux and windows builds successful
-
-	# check that tag_name is actually in the git repo and find the previous tag release
+	
+	# check that tag_name is in the
```
ngld, line 179: Why is this comment incomplete?

ci/post/main.py
```diff
@@ -145,12 +155,29 @@ def main():
 		print("ERROR: Invalid release mode %s passed. Expected release or nightly!" % sys.argv[1])
 		sys.exit(1)
 
+	# bail if any build subflows failed
+	success = ((os.environ["LINUX_RESULT"] == "success") and (os.environ["WINDOWS_RESULT"] == "success"))	##!< true if both linux and windows builds successful
+	if not success:
+		print("ERROR: One or more builds failed to publish!")
+		sys.exit(1)
```
ngld, line 162: We used to post nightlies on Nebula and the forum even if one of the builds failed. Is this change intentional?

ci/post/main.py
```diff
@@ -63,42 +65,50 @@ def _match_version_number(text, regex):
...
 	"""
-	major = minor = build = version = 0
-	with open(os.path.join("..", "..", "cmake", "version.cmake"), "r") as f:
+
+	with open(os.path.join("..", "..", "version_override.cmake"), "r") as f:
 		filetext = f.read()
 		major = _match_version_number(filetext, MAJOR_VERSION_PATTERN)
 		minor = _match_version_number(filetext, MINOR_VERSION_PATTERN)
 		build = _match_version_number(filetext, BUILD_VERSION_PATTERN)
+	#	revision = _match_version_number(filetext, REVISION_VERSION_PATTERN)
```
ngld, line 86: The indentation here is off.

ngld (changes requested):
There are a few stylistic nitpicks I could add (mostly regarding unnecessary braces in boolean expressions) but I don't know if you care about that.
Aside from the issues I've commented on, this looks good.

ci/post/changelog.py
```diff
@@ -0,0 +1,158 @@
+# Retrives a log of commit messages and attempts to pretty them into a human-readbile format
+
+import datetime
```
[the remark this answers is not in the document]
z64555: VSCode has a habit of adding in `imports` by itself, so I'm not sure if I did that or if VSCode did. :shrug:

ci/post/main.py
z64555: Indeed, why is that incomplete???  I might've been sidetracked right in the middle of updating it and forgot to save before committing the changes.

ci/post/main.py
z64555: Right now this script only works on the Releases, the Nightlies are still working in their own script on the nightlybuild repo.  I'd have to confer with ya'll on Discord to see what the build failure policy is.

ci/post/main.py
```diff
@@ -22,6 +22,7 @@
 # LINUX_RESULT: bool, if the linux builds were sucessfully uploaded
 # WINDOWS_RESULT: bool, if the windows builds were successfully uploaded
 
+from curses import window
```
ngld, line 25: I guess this is another stray auto-import from VSCode?
You can disable that. I think the option is called [`autoImportCompletions`](https://code.visualstudio.com/docs/python/settings-reference#_python-language-server-settings).

ngld (changes requested):
Looks good! You just have to remove the bogus `curses` import.

ngld (approved):
LGTM

PR #4127, opened 2022-04-02
vectorize Soundtracks and Spooled_music

Goober5000:
This removes the limits on soundtracks and menu music.  Some design tweaks are thrown in as well.


wookieejedi (approved):
Another good update regarding vectorization and is straightforward.

PR #4128, opened 2022-04-02
Switched light shader attenuation math to something that looks more real

EatThePath:
This is a change I've been playing with on my personal lighting experiments builds a lot and think is a big step up. I think it may make sense to have the 22.2 default weapon light sizes or brightness be different under this, as the result otherwise is generally dimmer lighting, but followup PRs will add controls for those defaults to lighting profiles. We could potentially have defaults tied to mod target version then, or something.

### DIFF
--- a/code/def_files/data/effects/deferred-f.sdr
+++ b/code/def_files/data/effects/deferred-f.sdr
@@ -98,7 +98,7 @@ void GetLightInfo(vec3 position, out vec3 lightDir, out float attenuation)
 				}
 			}
 		}
-		attenuation*=attenuation;
+		attenuation *= attenuation;
 		lightDir = normalize(lightDir);
 	}
 }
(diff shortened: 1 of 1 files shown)


wookieejedi:
Tested this myself and this looks a lot better IMO.

code/def_files/data/effects/deferred-f.sdr
```diff
@@ -98,7 +98,7 @@ void GetLightInfo(vec3 position, out vec3 lightDir, out float attenuation)
 				}
 			}
 		}
-
+		attenuation*=attenuation;
```
JohnAFernandez, line 101: Just a style thing.  It's much more readable if it's `attenuation *= attenuation`. It might be helpful to think of how white space, parentheses and such make the code clearer.  You want your code to be readable by any intrepid new coder who comes down the block.

JohnAFernandez (approved):

code/def_files/data/effects/deferred-f.sdr
EatThePath: For the record you don't have to explain the value of style fixes in general to me, I get it, I just suck at paying attention to and spotting it.

TRBlount (approved):

PR #4130, opened 2022-04-02
Render buffering optimization

EatThePath:
Based on reading http://hacksoflife.blogspot.com/2010/02/double-buffering-vbos.html and other similar things. We appear to use that trick in several places but not everywhere. It's possible some of these changes are redundant or superfluous, but extensively benchmarking a decently high load cutscene level showed in the area of 10% improvement of framerate.

### DIFF
--- a/code/graphics/opengl/gropengldeferred.cpp
+++ b/code/graphics/opengl/gropengldeferred.cpp
@@ -433,6 +433,7 @@ void gr_opengl_deferred_light_cylinder_init(int segments) // Generate a VBO of a
 	// make sure we have one
 	if (deferred_light_cylinder_vbo) {
 		glBindBuffer(GL_ARRAY_BUFFER, deferred_light_cylinder_vbo);
+		glBufferData(GL_ARRAY_BUFFER, nVertex * sizeof(float), nullptr, GL_STATIC_DRAW);
 		glBufferData(GL_ARRAY_BUFFER, nVertex * sizeof(float), Vertices, GL_STATIC_DRAW);
 
 		// just in case
@@ -455,6 +456,7 @@ void gr_opengl_deferred_light_cylinder_init(int segments) // Generate a VBO of a
 	// make sure we have one
 	if (deferred_light_cylinder_ibo) {
 		glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, deferred_light_cylinder_ibo);
+		glBufferData(GL_ELEMENT_ARRAY_BUFFER, nIndex * sizeof(ushort), nullptr, GL_STATIC_DRAW);
 		glBufferData(GL_ELEMENT_ARRAY_BUFFER, nIndex * sizeof(ushort), Indices, GL_STATIC_DRAW);
 
 		// just in case
@@ -533,6 +535,7 @@ void gr_opengl_deferred_light_sphere_init(int rings, int segments) // Generate a
 	// make sure we have one
 	if (deferred_light_sphere_vbo) {
 		glBindBuffer(GL_ARRAY_BUFFER, deferred_light_sphere_vbo);
+		glBufferData(GL_ARRAY_BUFFER, nVertex * sizeof(float), nullptr, GL_STATIC_DRAW);
 		glBufferData(GL_ARRAY_BUFFER, nVertex * sizeof(float), Vertices, GL_STATIC_DRAW);
 
 		// just in case
@@ -555,6 +558,7 @@ void gr_opengl_deferred_light_sphere_init(int rings, int segments) // Generate a
 	// make sure we have one
 	if (deferred_light_sphere_ibo) {
 		glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, deferred_light_sphere_ibo);
+		glBufferData(GL_ELEMENT_ARRAY_BUFFER, nIndex * sizeof(ushort), nullptr, GL_STATIC_DRAW);
 		glBufferData(GL_ELEMENT_ARRAY_BUFFER, nIndex * sizeof(ushort), Indices, GL_STATIC_DRAW);
 
 		// just in case
--- a/code/graphics/opengl/gropengltnl.cpp
+++ b/code/graphics/opengl/gropengltnl.cpp
@@ -308,8 +308,9 @@ void* gr_opengl_map_buffer(gr_buffer_handle handle)
 	Assertion(GLAD_GL_ARB_buffer_storage != 0, "Persistent mapping is not available in this OpenGL context!");
 
 	opengl_bind_buffer_object(handle);
+	glBufferData(buffer_obj.type, buffer_obj.size, nullptr, buffer_obj.gl_usage);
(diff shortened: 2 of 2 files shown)


code/graphics/opengl/gropengltnl.cpp
```diff
@@ -308,8 +308,9 @@ void* gr_opengl_map_buffer(gr_buffer_handle handle)
 	Assertion(GLAD_GL_ARB_buffer_storage != 0, "Persistent mapping is not available in this OpenGL context!");
 
 	opengl_bind_buffer_object(handle);
+	glBufferData(buffer_obj.type, buffer_obj.size, NULL,buffer_obj.gl_usage);
```
wookieejedi, line 311: minor formatting, but whitespace missing after null,

EatThePath: Forcepushed fixes.

wookieejedi:
Looks good and works on my tests. Given there is precedence for using this trick and the quite good improvements in performance, I definitely support adding this change. I'd be interested in hearing what @asarium might think, too.

code/graphics/opengl/gropengldeferred.cpp
```diff
@@ -433,6 +433,7 @@ void gr_opengl_deferred_light_cylinder_init(int segments) // Generate a VBO of a
 	// make sure we have one
 	if (deferred_light_cylinder_vbo) {
 		glBindBuffer(GL_ARRAY_BUFFER, deferred_light_cylinder_vbo);
+		glBufferData(GL_ARRAY_BUFFER, nVertex * sizeof(float), nullptr, GL_STATIC_DRAW);
```
asarium, line 436: Can you explain why you added this? The initialized data will be overwritten by the call in the next line so this call is not necessary.

code/graphics/opengl/gropengltnl.cpp
```diff
@@ -308,8 +308,9 @@ void* gr_opengl_map_buffer(gr_buffer_handle handle)
 	Assertion(GLAD_GL_ARB_buffer_storage != 0, "Persistent mapping is not available in this OpenGL context!");
 
 	opengl_bind_buffer_object(handle);
+	glBufferData(buffer_obj.type, buffer_obj.size, nullptr, buffer_obj.gl_usage);
 	return glMapBufferRange(buffer_obj.type, 0, buffer_obj.size,
-	                        GL_MAP_WRITE_BIT | GL_MAP_PERSISTENT_BIT | GL_MAP_FLUSH_EXPLICIT_BIT);
+	                        GL_MAP_WRITE_BIT | GL_MAP_PERSISTENT_BIT | GL_MAP_FLUSH_EXPLICIT_BIT | GL_MAP_INVALIDATE_BUFFER_BIT);
```
asarium, line 313: This call has a completely different use case than what is described in the article to which you linked. Here we are mapping the buffer only once when it has been created and then we reuse the pointer for updating the data.

code/graphics/opengl/gropengltnl.cpp
```diff
@@ -308,8 +308,9 @@ void* gr_opengl_map_buffer(gr_buffer_handle handle)
 	Assertion(GLAD_GL_ARB_buffer_storage != 0, "Persistent mapping is not available in this OpenGL context!");
 
 	opengl_bind_buffer_object(handle);
+	glBufferData(buffer_obj.type, buffer_obj.size, nullptr, buffer_obj.gl_usage);
```
asarium, line 311: This call is wrong. In the case of persistent mapping we need to specify the `GL_MAP_PERSISTENT_BIT` flag via  `glBufferStorage`. Since this has already happened before this call, this is unnecessary anyway.

asarium (changes requested):
I'll be honest, if there are performance improvements due to this then I do not know where they come from. Please do more benchmarking to check exactly where that additional performance is coming from since the changes I see here, I would classify as inconsequential at best.

EatThePath:
> I'll be honest, if there are performance improvements due to this then I do not know where they come from. Please do more benchmarking to check exactly where that additional performance is coming from since the changes I see here, I would classify as inconsequential at best.

@asarium My profiling consisted of running a reasonably constrained cutscene mission, Blue Planet's Icarus, with a lua script to end the mission and restart it after exactly 150 seconds of play. This was set to run for 15 cycles, at which point the script ends the game.  I used release builds identical except for these changes, with care taken to try to have them run in similar computing conditions, and data was collected on release builds with the -profile_write_file flag's output.I overlaid the data and graphed through python scripts that identify the loop point of the mission(there's a single abnormal frametime at the start of each loop to sync off of) and combined the runs for an overall performance picture.

image

The scatter plots are fps (inverse frametime, strictly speaking) from individual frames, with the blue line being a rolling average. Top is the before build and bottom is the after.

I won't pretend I intimately understand the code or system I'm dealing with here, so I did my bet to be certain of the impact before submitting. What would you consider sufficient benchmarking?  Do you need me to use more raw data than the ~37 minutes each collected here? Do you have a particular context you need to see it tested in? Must I benchmark each changed line individually? I can likely do any of these, in time.

asarium:
The first thing to check is what change made the difference so yes, please benchmark with either of the files reverted back to `master` to determine which one of those changes actually causes the performance increase.

Furthermore, none of the changed code should be called often so that is why I am not sure why there is a performance difference. What GPU are you using? Maybe that additional `glBufferData` in `gr_opengl_map_buffer` somehow triggers a better path in the driver somehow.

EatThePath:
> Furthermore, none of the changed code should be called often so that is why I am not sure why there is a performance difference. What GPU are you using? Maybe that additional `glBufferData` in `gr_opengl_map_buffer` somehow triggers a better path in the driver somehow.

GeForce RTX 3060

[DxDiag.txt](https://github.com/scp-fs2open/fs2open.github.com/files/8405156/DxDiag.txt)

asarium:
Ok, then you are definitely using the persistent mapping path here which is the only location where `gr_opengl_map_buffer` is used.

EatThePath:
@asarium I did some profiling runs tonight with some permutations on the changes submitted. Crunching the numbers in any detail will have to be anotgher day, but the preliminary results suggest:
1. The changes in gropengldeferred.cpp produce mildly reduced overall frame-rates when only that set of changes is included.
2. Including either or both of the changed lines in gropengltnl.cpp produces comperable overall improvements in framerate, all three being close enough as to be probably insignificantly different.
3. Including the changes of both files produces results in line with changing just gropengltnl.cpp.

asarium:
Ok, so the improvements are definitely in the buffer mapping code. I checked the OpenGL spec again and (unless there is a bug in our code), the change should produce a `GL_INVALID_OPERATION` error since `glBufferData` is called on a buffer that was initialized with `glBufferStorage` which should make the buffer immutable.

If such an error is not produced then this may be a bug in the Nvidia driver. Do we have any data on what happens either with an AMD card or maybe another driver implementation (e.g. the open source Linux drivers)?

EatThePath:
If I can find any clearer version of this down the line I'll reopen it, but for now I'll take it off the backlog.

PR #4132, opened 2022-04-02
lua ship.checkVisibility mimicing sexp ship-is-visible

EatThePath:
the ship-is-visibile sexp is used heavily from inside lua in the popular radaricons script, and evaluating those sexps is unavoidably a heavy performance cost. This function can replace those sexp calls.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -7038,8 +7038,7 @@ int sexp_hits_left(int node, bool sim_hull)
  */
 int sexp_is_ship_visible(int node)
 {
-	ship *viewer_shipp = nullptr;
-	int n = node, ship_is_visible = 0;
+	int n = node;
 
 	// get the ship
 	auto ship_entry = eval_ship(n);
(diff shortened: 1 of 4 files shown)


code/parse/sexp/LuaShared.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+
+/**
+ * Determine if ship visible on radar
+ * 
+ * @return 0 - not visible
+ * @return 1 - marginally targetable (jiggly on radar)
+ * @return 2 - fully targetable
+ */
+int shared_check_ship_visibility(const ship *viewed, ship *viewer)
```
asarium, line 16: This function should be defined in the ship code since it has no inherent relation to SEXP or Lua as far as I can tell.

code/parse/sexp/LuaShared.cpp
```diff
@@ -0,0 +1,63 @@
...
+			}
+		}
+		else
+			viewer = Player_ship;
+	}
+
+	object * viewed_obj = &Objects[viewer->objnum];
+	int ship_is_visible = 0;
+	// get ship's *radar* visiblity
+	//if (viewer)
```
asarium, line 44: Please check if this is needed or not. If not then please remove the commented code.

asarium (changes requested):
Please also check your formatting again. The formatting of the changes is very inconsistent. Clang-format cal help if you do not want to do that manually.

code/parse/sexp/LuaShared.cpp
EatThePath: That was an unfortunate oversight on my part. The code has been revised to handle this case again.

code/parse/sexp/LuaShared.cpp
EatThePath: Fair enough, I've moved and renamed it, and excised the new files.

EatThePath:
> Please also check your formatting again. The formatting of the changes is very inconsistent. Clang-format cal help if you do not want to do that manually.

I recall in the past I'd run into trouble with the clang-format generating changes I was asked to undo, so I'd assumed it vestigial and kept it turned off. I much prefer having that kind of tool in play though, and I see it can be invoked in narrow enough form it will hopefully avoid trouble, so I will strive to use it as much as I can.

asarium (approved):

asarium:
In a file that is not new, formatting an entire file will cause a lot of problems so there I would suggest selecting the relevant changes and only formatting those with clang-format. As far as I'm aware, we do not have a strict style guide so how to handle formatting is ultimately up to you.

PR #4133, opened 2022-04-02
Controls6: Keyboard Localization and Scripted Contols

### LINKED ISSUE #2225: International Keyboard Layouts

z64555:
* Should fix #2225 
* Adds script bindings to allow proper hooking into control actions based on the control's enumeration name
  * This includes axes, buttons
* Allows scripted controls to override hardcoded functionality
* Re-Adds in `-joy-info` cmdline flag, for slightly quicker Joystick GUID retrieval
* Adds a cmdline flag to print out a template controlconfigdefaults.tbl

- [x] New pilot name text input respects keyboard layout
- [x] New preset save text input respects keyboard layout
- [x] Text input boxes, from the menus, LUA, etc. text input respects keyboard layout
- [x] Keybinds should be correctly reported in the Controls Config menu and in training messages<sup>1</sup>
- [x] Default keybinds: Saved and loaded as scan code (key position)
- [x] Preset keybinds: Saved and loaded as scan code

<sup>1</sup> Issue with displaying unicode in ascii/latin1 mode

<hr>

Test builds (fast debug):
[linux-FastDebug](https://github.com/scp-fs2open/fs2open.github.com/suites/6084939203/artifacts/210766349)
[windows-FastDebug-Win32-SSE2](https://github.com/scp-fs2open/fs2open.github.com/suites/6084939203/artifacts/210766351)
[windows-FastDebug-Win64-SSE2](https://github.com/scp-fs2open/fs2open.github.com/suites/6084939203/artifacts/210766352)

<hr>

SHA-1's before merge with master:
736db1dde24230987368b2189e49a3f3f002f389  98a121d28a4ed8b4467b3049cb44e38891fd58b2  aafb80512a3fc43fac459f2fa163b8386e55693d  fdf4a26f3f5f2babec755259cd59c7611357d15f  8ef87600c049d687d80bcbe0a5f598ede6f66caa

<hr>

Note: Badged members may fill/unfill Checkboxes by clicking on them


z64555:
Pain. 
image

BMagnu:
Also, just to add for the records: It also adds script bindings that allow proper hooking into control actions (so based on the controlsconfigdefault.tbl name, not any sort of localized default button or any of that crap), including for control actions that the engine considers axes, as well as continous buttons and normal trigger buttons. It also now allows for overriding button presses, which allows lua to process button presses but not pass them onto the engine.
Similarly, functionality to generate controlconfigdefault.tbl's from known presets is added.

BMagnu:
I can confirm that pilot and preset text inputs are properly local-sensitive.

Training messages and the controls config are showing localized keybinds as well, though they have trouble displaying some characters when not in unicode mode. This has to do with a deeper running issue with rendering special characters in FSO in general. FSO's VF files do not adhere to any standardized encoding. The previously hardcoded names were using this non-standard encoding. But they were the _only_ thing using it. All other files and text (like strings.tbl _as of retail data_, as well as the new controls display) use latin1 encoding. And in non-unicode mode, rendering text from these files will simply break. What'd be needed to fix this FSO wide, unrelated to the controls code, is a latin1 -> VF character layout conversion function, which is evoked on any string that's intended to be rendered.

As for saving keycode vs scancode: IMO, it only ever makes sense to save scancode or an equivalent keycode on a US-keyboard. Having any code saved that depends on the current locale is just asking for trouble and makes things non-portable.
Some locales might not have certain keys, making default keybinds unusable. Some users might switch between keyboard layouts, breaking their presets along the way, and so on.

z64555:
Re: defaults and presets in controlconfigdefault.tbl
  These are intended to be set by the mod developers, and are not as big of a concern for providing locale support.  We do have a ccfg.tbl generator in this pull request, so we could use the scan codes and thus covers the locale support to an extent.  Just need some way of making the nuances of keyboard locales known to the modder somehow.

Re: presets in the .pst files
  These are intended to be set and shared by the users, but as with the ccfg.tbl, these are not intended to be hand-edited.  We could add another attribute to the JSON to specify the locale if desired, but ultimately using the scan codes here is the Occam's Razor.

z64555:
The remaining tasks need to be performed on keyboards that are not the standard english QWERTY layout.  I'm looking into how to do this on my machine, but if others are already set up then feel free to test.

z64555:
For Windoes 10 users, follow the steps outlined here to add keyboard layouts: (https://www.windowscentral.com/how-change-your-keyboard-layout-windows-10)

Then, disable the Alt+Shift hotkey by the following steps:

1. Right click on Start
2. Left click on Settings
3. Click on Time and Language
4. Click on Language in the left nav menu (if the nav menu is not shown then click on the home button in the upper left)
5. Click on the Keyboard button at the top of the Language page
6. Click on Input Language Hotkeys to open the Text Services and Input Languages dialog
7. Select the "Between input languages" action and click on the "Change Key Sequence" button
8. Select the "Not Assigned" option for both radio button boxes in the popup and click OK
9. Click Apply, and OK to dismiss the dialog

Doing the above will stop the keyboard layout list from grabbing your key inputs should you hit both Alt and Shift.  You should still be able to cycle between keyboard layouts by the WinKey + Space hotkey sequence.

BMagnu:
I was able to confirm that librocket-style interfaces respect localized keyboards, as well as that player presets are saved as if the player has a US keyboard (which is exactly what we want: The presets are exchangeable and every keyboard type will end up with the bindings in the same place regardless of locale). Same with default controls: Specifying Y for example will correctly bind to a Z on a QWERTZ layout.

code/controlconfig/controlsconfig.cpp
```diff
@@ -2601,17 +2608,21 @@ int check_control_used(int id, int key)
...
+	if (is_control_used && Ignored_keys[id] > 0) {
+		Ignored_keys[id]--;
 	}
 
-	if (Control_config[id].continuous_ongoing) {
+	bool control_triggered = is_control_used && !is_ignored;
+
+	//Only call lua in here when it's a continuous button
+	if (Control_config[id].type == CC_TYPE_CONTINUOUS)
+		control_triggered &= !control_run_lua(static_cast<IoActionId>(id), control_triggered ? 1 : 0);
```
JohnAFernandez, line 2623: I'll be doing single comments because I don't know how far I'll get to review before we trigger 22.2 RC phase.

A bitwise operation on a bool makes me nervous even though it's probably fine, and the syntax is confusing.  I had to do a logic table to figure out what the possibilities are.  It turns out that this is equivalent to 

```if ((Control_config[id].type == CC_TYPE_CONTINUOUS) && !control_run_lua(static_cast<IoActionId>(id)), control_triggered ? 1 : 0)) {
		control_triggered = false; 
}```

Because `control_triggered` can only *remain* true if `control_run_lua(...)` returns false, the compiler respects the order that we write our checks, so, aiui, it will not run `control_run_lua(...)` if `Control_config[id].type == CC_TYPE_CONTINUOUS` is false.

File: code/controlconfig/controlsconfig.h [hunk 2]
```diff
@@ -361,6 +372,16 @@ class CC_bind {
...
+	/**
+	 * Takes the given binding.
+	 */
+	void take(CID _cid, short _btn, char _flags = 0);
+
+	/**
+	 * Validates the binding, clearing it if necassary
+	 */
+	void validate();
+
```
JohnAFernandez, line 384: Just a question.  Why was this moved?

code/controlconfig/presets.h
```diff
@@ -216,6 +216,14 @@ bool save_preset_file(CC_preset preset, bool overwrite);
  */
 void load_preset_files();
 
+/**
+ * @brief Checks if the given preset is a duplicate within Control_config_presets vector
+ * @returns tterator to the duplicate if found, or
```
JohnAFernandez, line 221: Typo? `tterator`

code/debugconsole/console.cpp
```diff
@@ -488,6 +488,21 @@ void dc_putc(char c)
...
+			if (c < 32)
+				break;
+
+			if (dc_command_buf.size() < MAX_CLI_LEN) {
+				dc_command_buf.push_back(c);
+			}
+		}
+	}
+
+	return true;
```
JohnAFernandez, line 503: This doesn't seem like it can return false.  Why does it return a bool in the first place?

code/controlconfig/controlsconfig.cpp
z64555: @BMagnu Need confirmation on the logic here.

  check_control_used()'s return value is the state of the control, that is, wether it was activated or not since the last frame.

  control_run_lua()'s return value is "whether the control should be overridden or not" according to its docs, but my questions are "overridden by what?" and "How does this play into control_check_used()?"

code/debugconsole/console.cpp
z64555: This is a handler method that must be of a form that can be accepted by os::events::addEventListener() in osapi, which requires the boolean return value AFAIK.

code/debugconsole/console.cpp
```diff
@@ -488,6 +488,21 @@ void dc_putc(char c)
 	line_str->push_back(c);
 }
 
+bool handle_textInputEvent(const SDL_Event& event) {
+	if (event.text.text[0] != '\0' && event.text.text[0] != '\1') {
+		for (char c : event.text.text) {
+			if (c < 32)
```
z64555, line 494: Hmmm... I'm assuming this check is to omit control/escape characters in the ASCII table, but I'm not entirely sure.

z64555: Yup, all characters less than 32 are special: https://github.com/z64555/fs2open.github.com/pull/4/files#r546547542

code/controlconfig/presets.h
z64555: 👀

code/controlconfig/controlsconfig.h
z64555: Moved for organization, to better group the Getter methods together.

code/controlconfig/controlsconfig.cpp
BMagnu: A control is overridden, if the Lua override hook returns true.
General logic on most scripting overrides in FSO imply, that something being successfully overridden is as if it had never happened at all. In the case of controls: If you return true from an override hook, behave as if the button has not been pressed.
Hence, an overridden control is, at least as far as the rest of FSO is concerned, not used.

JohnAFernandez: So then, the logic is correct.  It's just convoluted. I believe what I put above is an easier way to express the same logic.

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1523,126 +1253,489 @@ size_t find_control_by_text(SCP_string& text) {
...
+		Controls_lua_override_cache.emplace(id, override);
+	}
+
+	return override;
+}
+
+void control_register_hook(IoActionId id, const luacpp::LuaFunction& hook, bool is_override, bool enabledByDefault) {
+	Control_config[id].scriptEnabledByDefault = enabledByDefault;
+
+	LuaHook* hook_entry = &Lua_hooks[static_cast<IoActionId>(id)];
```
JohnAFernandez, line 1372: It looks like id is already type `IoActionId`.

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1523,126 +1253,489 @@ size_t find_control_by_text(SCP_string& text) {
...
+ *
+ * @param[in] s Value of a call to optional_string_either(); 0 = "ControlConfigOverride" 1 = "ControlConfigPreset"
+ * @param[in] first_override Legacy support for unnamed #ControlConfigOverrides.  If this is the first unnamed
+ *   override, then overwrite the default preset, else, save it as an "<unnamed preset>"
+ * 
+ * @details ControlConfigPresets are read in the exact same manner as ControlConfigOverrides, however only the bindings are available for modification.
+ *  There may be only one #Override section, since it is in charge of non-binding members of the Control_config items
+ */
+void control_config_common_read_section(int s, bool first_override) {
+	Assert((s == 0) || (s == 1));
```
JohnAFernandez, line 1660: This should be an assertion that dumps what s is. Otherwise the coder debugging is going to shrug if they can't reproduce.

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1674,40 +1767,61 @@ void control_config_common_read_section(int s, bool first_override) {
...
 
-	if ((s == 0) && (new_preset.name == default_preset.name)) {
-		// If this is an override section, and the preset name is the same as the hardcoded defaults, override the defaults
-		auto& new_bindings = new_preset.bindings;
-		std::copy(new_bindings.begin(), new_bindings.end(), default_bindings.begin());
+	// Override the hardcoded defaults if this is an override section and the preset name is "default"
+	// Error case of preset sections named "default" is handled in the beginning of this function
+	if ((s == 0) && (new_preset.name == "default")) {
+		Control_config_presets[0] = new_preset;
+		mprintf(("[controlconfigdefaults.tbl] Overrode default preset."));
```
JohnAFernandez, line 1800: New line!

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1674,40 +1767,61 @@ void control_config_common_read_section(int s, bool first_override) {
...
-				unique = false;
-				break;
-			}
-		}
+	auto duplicate = preset_find_duplicate(new_preset);
+
+	if (duplicate == Control_config_presets.end()) {
+		// No duplicate, add new preset
+		Control_config_presets.push_back(new_preset);
+		mprintf(("[controlconfigdefaults.tbl] Added Preset '%s'", new_preset.name.c_str()));
```
JohnAFernandez, line 1808: New line!

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1674,40 +1767,61 @@ void control_config_common_read_section(int s, bool first_override) {
...
+		
+		} else {
+			// Presets are not allowed to rename
+			Warning(LOCATION, "Ignoring Preset '%s'. (Duplicate of '%s' and is not an Override)", new_preset.name.c_str(), duplicate->name.c_str());
+			mprintf(("[controlconfigdefaults.tbl] Ignoring Preset '%s'. (Duplicate of '%s' and is not an Override)", new_preset.name.c_str(), duplicate->name.c_str()));
 		}
-	}
+	} // Else, silently ignore the duplicate since it has the same name
+
+	mprintf(("[controlconfigdefaults.tbl] Ignoring duplicate Preset '%s'", new_preset.name.c_str()));
```
JohnAFernandez, line 1824: New line!

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1674,40 +1767,61 @@ void control_config_common_read_section(int s, bool first_override) {
...
+		if (s == 0) {
+			// Rename the duplicate with the new_preset name
+			// The .tbl takes precedence over any player presets
+			duplicate->name = new_preset.name;
+			mprintf(("[controlconfigdefaults.tbl] Renamed duplicate '%s' preset to '%s'", duplicate->name.c_str(), new_preset.name.c_str()));
+		
+		} else {
+			// Presets are not allowed to rename
+			Warning(LOCATION, "Ignoring Preset '%s'. (Duplicate of '%s' and is not an Override)", new_preset.name.c_str(), duplicate->name.c_str());
+			mprintf(("[controlconfigdefaults.tbl] Ignoring Preset '%s'. (Duplicate of '%s' and is not an Override)", new_preset.name.c_str(), duplicate->name.c_str()));
```
JohnAFernandez, line 1820: New line!

code/controlconfig/controlsconfigcommon.cpp
```diff
@@ -1674,40 +1767,61 @@ void control_config_common_read_section(int s, bool first_override) {
...
-		if (unique) {
-			Control_config_presets.push_back(new_preset);
-		} else if (!running_unittests) {
-			Warning(LOCATION, "Preset '%s' found in 'controlconfigdefaults.tbl' is a duplicate of existing preset '%s', ignoring", new_preset.name.c_str(), it->name.c_str());
+	} else if (duplicate->name != new_preset.name) {
+		if (s == 0) {
+			// Rename the duplicate with the new_preset name
+			// The .tbl takes precedence over any player presets
+			duplicate->name = new_preset.name;
+			mprintf(("[controlconfigdefaults.tbl] Renamed duplicate '%s' preset to '%s'", duplicate->name.c_str(), new_preset.name.c_str()));
```
JohnAFernandez, line 1815: New line!

JohnAFernandez (changes requested):
I cannot guarantee that I found everything.  But I looked.

z64555:
Thank you much.

JohnAFernandez (approved):

Issue #4098, opened 2022-03-25
collide_remove_weapons not useful


EatThePath:
I recently hit the weapon cap, and discovered that the safety function `collide_remove_weapons` in that case is not effective. collide_remove_weapons is supposed to go through the weapon collider cache and find any weapons that are safe to remove, but in practice what it does is consume around 150ms and then return zero. 
To show this experimentally I cranked the weapon limit down to 1000, added some code to make sure it doesn't rerun on the same frame after failing once, and gave my test mission a run. During four minutes of gameplay `collide_remove_weapons` ran 280 times, and in 276 instances it removed 0 weapons. So, it's not impossible for it to remove something, but it is exceedingly unlikely. Given the performance it from running it in the conditions where it's relevant(when I caught it in a debugger with normal limits it was checking over 2 million cached collider pairs for validity), in the current state it'd make more sense to just not create weapons in these cases than to make this attempt to scavenge room for them.

JohnAFernandez:
Closing the issues as the associated pull request closed.

PR #4134, opened 2022-04-02
Removed collide_remove_weapons from weapon_create

EatThePath:
Collide_remove_weapons being called in weapon_create to recover weapon slots is actively destructive to player experience and almost never recovers slots.  See #4098 for more on that. Instead, simply don't create a weapon in those cases.
Tested on the same scenario as originally prompted #4098 and encountered no player-visible issues.


JohnAFernandez:
Having also been in this part of the code, because of another bug, I'm in support of this approach. I'm not sure that there's a good way to scavenge slots efficiently. We eventually just need to make the limit dynamic, as with pretty much all the other limits.

Goober5000 (approved):

PR #4144, opened 2022-04-05
Interpolation Overhaul 2

### LINKED ISSUE #3042: Shaking related to interpolation is more prominent on certain setups, server framerates, and host computers.
This points to some sort of bug in the code.  If it can run smoothly on some setups, it should run smoothly with a similar bandwidth and framerate on other setups.

Test A: standalone at 30 fps, host at 60 fps, client at 60 fps.

A: Host and client a little jerky but easily playable

Test B: standalone at 60 fps, host at 60 fps, client at 60 fps
B: no apparent difference

Test C: master/host at 60 fps, client at 60 fps
C: master/host smooth, client pretty jerky

Test D: master/host at 30 fps, client at 60 fps
D: master/host smooth, client smooth

Test E: standalone at 30 fps, host at 30 fps, client at 60 fps
Test F: standalone at 60 fps, host at 30 fps, client at 60 fps
Both E and F were smooth on the client, though E may have been slightly smoother
I also watched the mission time, which did not stay in sync. The inactive window always ran behind the active window, so I would have one active until it was 3-4 seconds ahead, then switch to the other one until it ended up 3-4 seconds ahead. Regardless of the time difference the game remained smooth in both E and F

JohnAFernandez:
Base interpolation timing on Missiontime on remote instance vs. Mission time on the local instance.  

Based on the limited testing that I've done so far, it seems to fix the jerkiness that used to occur when the interpolation code basically guessed the timing on when packets were coming in.  It is also a cleaner and easier to read solution than the last one.

More testing required before merge, but it's in a good enough state to start being reviewed.

Closes #3042

### DIFF
--- a/code/network/multi_interpolate.cpp
+++ b/code/network/multi_interpolate.cpp
@@ -1,0 +1,203 @@
+#include "network/multi_interpolate.h"
+#include "globalincs/pstypes.h"
+#include "freespace.h"
+
+///////////////////////////////////////////
+// interpolation info management functions
+
+// seeks through the packets to find the one that we need, starting from the end, notice we cannot use MULTIPLAYER_CLIENT macro here.  We cannot include multi.h
+void interpolation_manager::reassess_packet_index(vec3d* pos, matrix* ori, physics_info* pip) 
+{
+	auto current_time = Multi_Timing_Info.get_current_time();	
+	int current_index = static_cast<int>(_packets.size()) - 2;
+	int prev_index = static_cast<int>(_packets.size()) - 1;
+
+	// iterate through the packets
+	for (; current_index > -1; current_index--, prev_index--) {
+
+		// did we find where we should interpolate?
+		if ((_packets[current_index].remote_missiontime >= current_time) && (_packets[prev_index].remote_missiontime <= current_time)) {
+			_upcoming_packet_index = current_index;
+			_prev_packet_index = prev_index;
+
+			// probably the "hackiest" thing about this.  If we were just straight simulating, 
+			// and we now need to go back, pretend that the position we were in *was* our old packet
+			// and we are now going towards our new packet's physics.
+			if (_simulation_mode) {
+				replace_packet(prev_index, pos, ori, pip); // TODO, if simulation mode was forced by the collision code, this method regresses a bug where collisions instantly kill
+				_simulation_mode = false;
+			}
+
+			return;
+		}
+	}
+
+	// if we didn't find indexes then we are overwhelmingly likely to have passed the server somehow
+	// and we need to make sure that we just straight simulate these ships
+	_simulation_mode = true;
+}
+
+// the meat and potatoes.  Basically, this figures out if we should interpolate, and then interpolates or sims
+void interpolation_manager::interpolate(vec3d* pos, matrix* ori, physics_info* pip, bool player_ship)
+{
+	// make sure its a valid ship and valid mode
+	Assert(Game_mode & GM_MULTIPLAYER);
+
+	// if we have not received enough packets, simulate then return, 
+	// To optimize, we should not reassess_packet_index with a negative index.  
+	// The index will be made positive by add_packet, once a second packet has been received.
+	if (_upcoming_packet_index < 0 ) {
+		physics_sim(pos, ori, pip, flFrametime);
+
+		// duplicate the rest of the physics engine's calls here to make the simulation more exact.
+		pip->speed = vm_vec_mag(&pip->vel);
+		pip->fspeed = vm_vec_dot(&ori->vec.fvec, &pip->vel);
+
+		return;
+	}
+
+	reassess_packet_index(pos, ori, pip);
+
+	// if we are off the beaten path
+	if(_simulation_mode) {
+		
+		float sim_time = flFrametime;
+
+		// we need to push this ship up to the limit of where we were on the remote instance, if we haven't already.
+		// then we need to adjust our timing since some of the sim time is used up getting to that last packet.
+		if (!_packets_expended && !_packets.empty()) {
+			*pos = _packets.front().position;
+			vm_angles_2_matrix(ori, &_packets.front().angle);
+			pip->vel = _packets.front().velocity;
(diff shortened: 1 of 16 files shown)


code/network/multi_interpolate.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+	// remove them
+	for (auto& culled_item : to_remove) {
+		_frame_status[player_index].erase(culled_item);
+	}
+}
+
+// make sure that we recalculate the current local time
+void multiplayer_timing_info::update_current_time() {
+	_current_time = f2i(static_cast<std::int64_t>(Missiontime) * TIMESTAMP_FREQUENCY) - _start_time + _skipped_time;
```
Goober5000, line 264: This conversion doesn't produce anything, really.  Take a look at `timestamp_get_mission_time()` which is how `Missiontime` is calculated.

Incidentally, can't you just work with `Missiontime` directly, without converting it to anything?

code/network/multi_interpolate.h
```diff
@@ -0,0 +1,208 @@
...
+#include "globalincs/systemvars.h"
+#include "io/timer.h"
+#include "math/vecmat.h"
+
+// A manually written copy of Int max to keep linux CI from complainint.
+constexpr int COMPAT_INT_MAX = 2147483647;
+
+struct physics_info;
+
+constexpr int DEFAULT_FRAME_TIME = 17; // 1000 ms / 60 fps
```
notimaginative, line 13: How dependent on 60 fps is this to work correctly?

Standalone defaults to 30 fps and that can only be changed in the UI. Plus the ultimate goal with the new standalone changes is to make it dynamic, 1 - 120 fps, so it can respond to packets as efficiently as possible.

JohnAFernandez: We really just need some kind of default.  On one of my tests, the client was missing about 20 early frames, and then wasn't missing one until about frame 700.  I can try to have it check if it is on a standalone and adjust?  But dropped packets are not as common a thing as one would expect.

code/network/multi_interpolate.cpp
JohnAFernandez: I see what you're saying now.  f2i takes a fix, so it wasn't keeping the 32 bit word size.  It was to prevent overflow, so maybe it will properly do that now that I just added the bitshift.

The time information from the other machine are coming in as timestamp deltas.  Converting `Missiontime` here allows us to work in the same format as the incoming data.  I could change what it sends, but that would require a multi version bump.

Goober5000: Oh, I misinterpreted your previous code as referring to the timer frequency, not timestamp frequency.  Please rebase onto latest master and use `MILLISECONDS_PER_SECOND` instead.

I see what you mean about converting the time to milliseconds.  Though an extra comment to indicate the units would be helpful.

Most of the other code that works with `Missiontime` converts the deltas to `fix`.  For example:
```
		//	If a very long time since attacked, attack no matter what!
		if (Missiontime - aip->last_attack_time > i2f(6)) {
```
Is that an option?

code/network/multi_interpolate.h
JohnAFernandez: With the change to use a full timestamp and a multi bump now being required, this is no longer an issue.

code/network/multi_interpolate.cpp
JohnAFernandez: I changed everything over to timestamp, because something seemed to be going wrong with the calculation.  Later investigations point to it being something else, but I started using a basic timestamp() call until it's more stable (it's no longer crashing, but we have to make the movements a lot smoother now)

JohnAFernandez:
Just a little bit more testing and refinement before this is ready.

code/network/multi_interpolate.cpp
Goober5000: Ok.

Goober5000 (commented):
Now that `timestamp_since` has been merged, `time_delta_since_ms` can be replaced with that.

Goober5000 (commented):
Can you rebase this?  It looks like there's some old or invalid code here - the `timestamp_elapsed_safe` function is the incorrect version that used `timer_get_milliseconds`.

Goober5000:
The rest of it looks ok, though I'm not really qualified to review the multi parts.

code/object/object.h
```diff
@@ -203,6 +206,8 @@ extern object obj_free_list;
 extern object obj_used_list;
 extern object obj_create_list;
 
+extern ushort Test_subject;
```
notimaginative, line 209: I believe this is some left-over code and should be removed.

code/network/multi_interpolate.h
```diff
@@ -0,0 +1,127 @@
...
+		_prev_packet_index = -1;
+		_simulation_mode = false;
+		_packets_expended = false;
+		_local_skip_forward = 0;
+		_hull_comparison_frame = -1;
+		_shields_comparison_frame = -1;
+		_subsystems_comparison_frame.reserve(64);
+		_ai_comparison_frame = -1;
+		_source_player_index = -1;
+	}
```
notimaginative, line 126: Since this is attached to general `object`'s, no memory should be allocated in this constructor. `Objects[]` is still a static size much of this allocated memory could end up wasted. And this is 100% wasted in single player too. Just have the `reset()` take care of it.

code/network/multi_interpolate.h
```diff
@@ -0,0 +1,127 @@
...
+
+	void set_ai_comparison_frame(int frame) { _ai_comparison_frame = frame; }
+
+	void force_interpolation_mode() { _simulation_mode = true; }
+
+	void reset() 
+	{
+		_packets.clear();
+		_packets.reserve(PACKET_INFO_LIMIT * 2);
+		_upcoming_packet_index = -1;
```
notimaginative, line 96: There should be a multi mode check here so that it will bail if single player. That way single player mode doesn't waste time/memory dealing with this interpolation stuff.

notimaginative (changes requested):

notimaginative (approved):

PR #4151, opened 2022-04-08
Fix cockpit engine sound volume changing unexpectedly

Baezon:
...due to sudden changes in engine energy, which can affect `current_max_speed`. This makes 100% volume at the ship's tabled default max speed.


JohnAFernandez (approved):
I think we've discussed this one to death. Approving

PR #4155, opened 2022-04-09
Insertion sort update

Goober5000:
Turn `insertion_sort` into a templated function that uses assignment, so that it can be used with non-POD types.  Also make the sort do lazy-copying of elements and make a few other tweaks to the function.

### DIFF
--- a/code/globalincs/systemvars.h
+++ b/code/globalincs/systemvars.h
@@ -213,7 +213,39 @@ int current_detail_level();
 
 
 // Goober5000
-void insertion_sort(void *array, size_t array_size, size_t element_size, int (*fncompare)(const void *, const void *));
+// (Taylor says that for optimization purposes malloc/free should be used rather than vm_malloc/vm_free here)
+template <typename T>
+void insertion_sort(T* array_base, size_t array_size, int (*fncompare)(const T*, const T*))
+{
+	size_t i, j;
+	T* current;
 
+	// allocate space for the element being moved
+	current = (T*)malloc(sizeof(T));
+	if (current == nullptr)
+	{
+		UNREACHABLE("Malloc failed!");
+		return;
+	}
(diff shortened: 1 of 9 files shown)


z64555:
Why is `windebug.cpp` being compiled by the linux builds, anyway?

z64555:
Ok, as @BMagnu suggested in discord, we can try `#include <windows.h> //NOLINT`

If that doesn't work, we can also try guarding the .cpp with `#ifdef _WIN32`.

Probably the better solution would be to update the regex to ignore `globalincs/windebug.cpp` in `ci/linux/clang-tidy.sh`

Goober5000:
Ok, regex updated and sort tweaked again.

code/globalincs/systemvars.h
```diff
@@ -213,7 +213,56 @@ int current_detail_level();
 
 
 // Goober5000
-void insertion_sort(void *array, size_t array_size, size_t element_size, int (*fncompare)(const void *, const void *));
-
+// (Taylor says that for optimization purposes malloc/free should be used rather than vm_malloc/vm_free here)
+// NOTE: j *must* be a signed type because j reaches -1 and j+1 must be 0.  So we might as well make them ints
+// and make array_size an int as well.
+template <typename T>
+void insertion_sort(T* array_base, int array_size, int (*fncompare)(const T*, const T*))
```
z64555, line 220: Recommend documenting the parameters, especially the `fncompare`.  I'm assuming it's formed similarly to std::operators, i.e. `bool operator>(left, right)`

Bonus points for doxygen format.

z64555 (approved):

code/globalincs/systemvars.h
Goober5000: k

PR #4156, opened 2022-04-09
upgrade supernova timestamps

Goober5000:
This adds a type-safe `TIMESTAMP`, similar to `UI_TIMESTAMP`.  Eventually all of the FSO codebase will be switched over to this new type.  In the meantime, just the supernova code is switched over.

### DIFF
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -24,9 +24,6 @@ static Uint64 Timer_base_value;
 
 static bool Timer_inited = false;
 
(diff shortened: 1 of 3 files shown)


z64555 (approved):
LGTM.  I'm curious about that `td = 2` hack at the rollover point, however.

Goober5000:
> LGTM. I'm curious about that `td = 2` hack at the rollover point, however.

That's in the original code.  I'm not quite sure what it means either, but I suspect it has something to do with the special values of 0 and 1.

PR #4157, opened 2022-04-10
add game_settings.tbl option to specify subtitle coords in pixels

Goober5000:
What it says on the tin.  This allows more precise positioning of subtitles, and a future PR will add the ability to scale them according to screen resolution.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -88,6 +88,7 @@ float Min_pizel_size_muzzleflash;
 float Min_pixel_size_trail;
 float Min_pixel_size_laser;
 bool Supernova_hits_at_zero;
+bool Show_subtitle_uses_pixels;
 
 void mod_table_set_version_flags();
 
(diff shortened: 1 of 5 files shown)


z64555:
Whats the rationale behind using `game_settings.tbl` for subtitle positioning?

There is a hud configuration file, .hcf, that's currently used to hold gauge colors, but I suppose trying to make the UI for subtitle positioning here for the user is a pain.

code/mod_table/mod_table.cpp
```diff
@@ -297,6 +298,15 @@ void parse_mod_table(const char *filename)
 			}
 		}
 
+		if (optional_string("$Show-subtitle uses pixels:")) {
+			stuff_boolean(&Show_subtitle_uses_pixels);
+			if (Show_subtitle_uses_pixels) {
+				mprintf(("Game Settings Table: Show-subtitle uses pixels\n"));
+			} else {
+				mprintf(("Game Settings Table: Show-subtitle uses percentages\n"));
```
z64555, line 306: Since this is the default value, should this message be printed out even if the `$Show-subtitle uses pixels:` option is not present?

Goober5000:
> Whats the rationale behind using `game_settings.tbl` for subtitle positioning?
> 
> There is a hud configuration file, .hcf, that's currently used to hold gauge colors, but I suppose trying to make the UI for subtitle positioning here for the user is a pain.

This is not at all going to be a full positioning system.  There are only two options - one in this PR to use pixels, and one in the next PR to set a base resolution.  Putting those in `game_settings.tbl` is appropriate IMO.

code/mod_table/mod_table.cpp
Goober5000: I don't think so.  This follows the pattern of other options: they are only logged to the log file if they are explicitly specified somewhere.

z64555 (approved):

PR #4158, opened 2022-04-10
add the ability for subtitles to be scaled correctly to the screen 

Goober5000:
By specifying a base resolution in game_settings.tbl, subtitles can be scaled to the screen in the same way as HUD gauges.  This can be used to prevent subtitles from appearing smaller at larger resolutions.

Depends on and is a follow-up to https://github.com/scp-fs2open/fs2open.github.com/pull/4157.


wookieejedi (approved):
LGTM

PR #4159, opened 2022-04-10
guard against off-by-one errors in replace_one and replace_all

Goober5000:
Also add some test cases for checking correct behavior.

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -3938,6 +3938,8 @@ void consolidate_double_characters(char *src, char ch)
 ptrdiff_t replace_one(char *str, const char *oldstr, const char *newstr, size_t max_len, ptrdiff_t range)
 {
 	Assert(str && oldstr && newstr);
+	if (!str || !oldstr || !newstr)
+		return -3;
 
 	// search
 	char *ch = stristr(str, oldstr);
@@ -3948,13 +3950,13 @@ ptrdiff_t replace_one(char *str, const char *oldstr, const char *newstr, size_t 
 		// not found within bounds?
 		if ((range > 0) && ((ch - str) > range))
 		{
-			return 0;
+			return -1;
 		}
 
 		// determine if replacement will exceed max len
 		if (strlen(str) + strlen(newstr) - strlen(oldstr) > max_len)
 		{
-			return -1;
+			return -2;
 		}
 
 		// allocate temp string to hold extra stuff
@@ -3979,7 +3981,7 @@ ptrdiff_t replace_one(char *str, const char *oldstr, const char *newstr, size_t 
 	// not found
 	else
 	{
-		return 0;
+		return -1;
 	}
 
 	// return pos of replacement
@@ -4003,59 +4005,74 @@ int replace_all(char *str, const char *oldstr, const char *newstr, size_t max_le
 		}
 	}
 
-	return (val < 0) ? -1 : tally;
+	// return the tally, even if it's 0, unless there was an exceptional situation like exceeding max len
+	return (val < -1) ? (int)val : tally;
 }
 
-SCP_string& replace_one(SCP_string& context, const SCP_string& from, const SCP_string& to)
+ptrdiff_t replace_one(SCP_string& context, const SCP_string& from, const SCP_string& to)
 {
 	size_t foundHere;
 	if ((foundHere = context.find(from, 0)) != SCP_string::npos)
 	{
 		context.replace(foundHere, from.length(), to);
+		return foundHere;
 	}
-	return context;
+	else
+		return -1;
 }
 
-SCP_string& replace_one(SCP_string& context, const char* from, const char* to)
+ptrdiff_t replace_one(SCP_string& context, const char* from, const char* to)
 {
 	size_t foundHere;
 	if ((foundHere = context.find(from, 0)) != SCP_string::npos)
 	{
 		context.replace(foundHere, strlen(from), to);
+		return foundHere;
 	}
-	return context;
+	else
+		return -1;
 }
 
 // http://www.cppreference.com/wiki/string/replace
-SCP_string& replace_all(SCP_string& context, const SCP_string& from, const SCP_string& to)
+int replace_all(SCP_string& context, const SCP_string& from, const SCP_string& to)
 {
 	size_t from_len = from.length();
 	size_t to_len = to.length();
 
 	size_t lookHere = 0;
 	size_t foundHere;
+	int tally = 0;
+
 	while ((foundHere = context.find(from, lookHere)) != SCP_string::npos)
 	{
+		tally++;
+
 		context.replace(foundHere, from_len, to);
 		lookHere = foundHere + to_len;
 	}
-	return context;
+
+	return tally;
 }
 
 // http://www.cppreference.com/wiki/string/replace
-SCP_string& replace_all(SCP_string& context, const char* from, const char* to)
+int replace_all(SCP_string& context, const char* from, const char* to)
 {
 	size_t from_len = strlen(from);
 	size_t to_len = strlen(to);
 
 	size_t lookHere = 0;
 	size_t foundHere;
+	int tally = 0;
+
 	while ((foundHere = context.find(from, lookHere)) != SCP_string::npos)
 	{
+		tally++;
+
 		context.replace(foundHere, from_len, to);
 		lookHere = foundHere + to_len;
 	}
-	return context;
+
+	return tally;
 }
 
 // WMC
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -3938,8 +3938,9 @@ void consolidate_double_characters(char *src, char ch)
 // Note that the parameter here is max *length*, not max buffer size.  Leave room for the null-terminator!
 ptrdiff_t replace_one(char *str, const char *oldstr, const char *newstr, size_t max_len, ptrdiff_t range)
 {
-	Assert(str && oldstr && newstr);
-	if (!str || !oldstr || !newstr)
+	Assertion(str && oldstr && newstr, "Arguments must not be null!");
+	Assertion(max_len < SIZE_MAX, "Size must be less than SIZE_MAX because an extra char is added for the null terminator.");
+	if (!str || !oldstr || !newstr || max_len == SIZE_MAX)
 		return -3;
 
 	// search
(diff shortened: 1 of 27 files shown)


fred2/campaigntreewnd.cpp
```diff
@@ -222,8 +222,8 @@ void campaign_tree_wnd::OnCpgnFileSaveAs()
 			return;		
 		}
 
-		string_copy(Campaign.filename, name, MAX_FILENAME_LEN);
-		string_copy(campaign_path, dlg.GetPathName(), 256);
+		string_copy(Campaign.filename, name, MAX_FILENAME_LEN - 1);
+		string_copy(campaign_path, dlg.GetPathName(), 256 - 1);
```
jg18, line 226: How about `sizeof()` instead of 256?

fred2/management.cpp
```diff
@@ -146,15 +146,16 @@ int create_ship(matrix *orient, vec3d *pos, int ship_type);
...
 {
 	if (modify)
 		if (strcmp(src, dest))
 			set_modified();
 
 	auto len = strlen(src);
-	if (len >= max_len)
-		len = max_len - 1;
+	if (len > max_len)
+		len = max_len;
```
jg18, line 158: Should we log the fact that we're clamping `len`?

fred2/orienteditor.cpp
```diff
@@ -369,7 +369,7 @@ float orient_editor::convert(const CString &str)
 	char buf[256];
 	size_t i, j, len;
 
-	string_copy(buf, str, 255);
+	string_copy(buf, str, 256 - 1);
```
jg18, line 372: As above, is using `sizeof()` possible?

fred2/voiceactingmanager.cpp
```diff
@@ -406,7 +406,7 @@ void VoiceActingManager::OnGenerateScript()
 		return;
 
 	CString dlgPathName = dlg.GetPathName( );
-	string_copy(pathname, dlgPathName, 256);
+	string_copy(pathname, dlgPathName, 256 - 1);
```
jg18, line 409: `sizeof()`?

It's especially useful when the declaration of the char buffer is several lines up, as opposed to immediately above.

test/src/parse/test_replace.cpp
```diff
@@ -0,0 +1,165 @@
+
+#include <gtest/gtest.h>
+
+#include <parse/parselo.h>
+
+#include "globalincs/safe_strings.h"
+
+#include "util/FSTestFixture.h"
+
+class ReplaceTest : public test::FSTestFixture {
```
jg18, line 10: Thanks for taking the time to write tests! :+1:

test/src/parse/test_replace.cpp
```diff
@@ -0,0 +1,165 @@
...
+	void TearDown() override {
+		stop_parse();
+
+		test::FSTestFixture::TearDown();
+	}
+};
+
+TEST_F(ReplaceTest, replace_one_cstr)
+{
+	constexpr int len = 10;
```
jg18, line 29: Maybe `LEN` or `MAX_LEN`, since it's a `constexpr` integral value?

test/src/parse/test_replace.cpp
jg18: Optional: it might be good if the test for the `replace_one()` overload that uses C-strings also tested for return value of -3. The tests you have already cover a lot of ground, though.

test/src/parse/test_replace.cpp
```diff
@@ -0,0 +1,165 @@
...
+	ASSERT_STREQ("0123456789", buf);
+}
+
+TEST_F(ReplaceTest, replace_one_scpstr)
+{
+	SCP_string buf = "aabbccddee";
+
+	// search string not found
+	ASSERT_EQ(-1, replace_one(buf, "ff", "XX"));
+	// SCP_string version of replace_one does not have a range argument
```
jg18, line 76: As an side, it's interesting that we haven't had a need for for `replace_foo()` functions that take both `SCP_string`s and position offsets into the string, rather than always handling the entire string. But I guess there's no point in introducing the complexity that comes with that flexibility unless/until we need it.

test/src/parse/test_replace.cpp
```diff
@@ -0,0 +1,165 @@
...
+	ASSERT_STREQ("abxyz45678", buf);
+
+	// successful replacement of entire string
+	ASSERT_EQ(0, replace_one(buf, "abxyz45678", "0123456789", len));
+	ASSERT_STREQ("0123456789", buf);
+}
+
+TEST_F(ReplaceTest, replace_one_scpstr)
+{
+	SCP_string buf = "aabbccddee";
```
jg18, line 72: It'd be nice to use a variable name that makes clear this is a string object and not a char buffer. Ditto for similar code below.

fred2/bgbitmapdlg.cpp
```diff
@@ -458,8 +458,8 @@ void bg_bitmap_dlg::OnClose()
 
     The_mission.flags.set(Mission::Mission_Flags::Subspace, m_subspace != 0);
 
-	string_copy(The_mission.skybox_model, m_skybox_model, NAME_LENGTH, 1);
-	string_copy(The_mission.envmap_name, m_envmap, NAME_LENGTH, 1);
+	string_copy(The_mission.skybox_model, m_skybox_model, NAME_LENGTH - 1, 1);
+	string_copy(The_mission.envmap_name, m_envmap, NAME_LENGTH - 1, 1);
```
jg18, line 462: Generally speaking, how did you find the call sites to update? Have all calls to, e.g., `strncpy()` also been checked?

jg18 (changes requested):
Some requests, including some optional ones, but overall this is a great idea. :+1:

Another good reason for this change is that some `replace_foo()` overloads use `SCP_string`s, which of course don't include buffer size as a user-facing notion.

code/parse/parselo.cpp
```diff
@@ -3978,19 +3981,21 @@ ptrdiff_t replace_one(char *str, const char *oldstr, const char *newstr, size_t
...
 	return (ch - str);
 }
 
 // Goober5000
-ptrdiff_t replace_all(char *str, const char *oldstr, const char *newstr, size_t max_len, ptrdiff_t range)
+// Note that the parameter here is max *length*, not max buffer size.  Leave room for the null-terminator!
+int replace_all(char *str, const char *oldstr, const char *newstr, size_t max_len, ptrdiff_t range)
 {
-	ptrdiff_t val, tally = 0;
+	ptrdiff_t val;
```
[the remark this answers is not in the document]
Goober5000: je pense que non

code/parse/parselo.cpp
```diff
@@ -3934,9 +3934,12 @@ void consolidate_double_characters(char *src, char ch)
 }
 
 // Goober5000
+// Note that the parameter here is max *length*, not max buffer size.  Leave room for the null-terminator!
 ptrdiff_t replace_one(char *str, const char *oldstr, const char *newstr, size_t max_len, ptrdiff_t range)
```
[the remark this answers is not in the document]
Goober5000: Ok

code/parse/parselo.cpp
```diff
@@ -4000,59 +4005,74 @@ ptrdiff_t replace_all(char *str, const char *oldstr, const char *newstr, size_t
...
+
 		context.replace(foundHere, from_len, to);
 		lookHere = foundHere + to_len;
 	}
-	return context;
+
+	return tally;
 }
 
 // http://www.cppreference.com/wiki/string/replace
```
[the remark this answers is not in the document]
Goober5000: Well, a) it's not a built-in function even though it uses them as building blocks, and b) I thought it made sense to credit the source of the algorithm.

code/parse/sexp.cpp
```diff
@@ -31479,7 +31480,10 @@ bool sexp_replace_variable_names_with_values(char *text, int max_len)
 				strncpy(what_to_replace, pos, strlen(Sexp_variables[var_index].variable_name) + 1);
 
 				// replace it
-				pos = text + replace_one(text, what_to_replace, Sexp_variables[var_index].text, max_len);
+				auto diff = replace_one(text, what_to_replace, Sexp_variables[var_index].text, max_len);
```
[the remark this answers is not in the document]
Goober5000: nah

fred2/campaigntreewnd.cpp
Goober5000: Eh, possible, but technically if `char` is ever more than 1 byte it wouldn't be correct

fred2/management.cpp
Goober5000: I don't think so.  Truncation isn't really the sort of thing you need to log in these cases.

fred2/orienteditor.cpp
Goober5000: Possible, but I would decide against it

fred2/voiceactingmanager.cpp
Goober5000: See above

test/src/parse/test_replace.cpp
Goober5000: K

test/src/parse/test_replace.cpp
Goober5000: Yeah

test/src/parse/test_replace.cpp
Goober5000: ok

fred2/bgbitmapdlg.cpp
Goober5000: I don't remember why I checked `string_copy` but I think it was because it was closely related to replacements in FRED.  IIRC I changed the function name, recompiled, checked all the compilation errors, then changed the name back.  I did not check `strncpy`.

jg18 (changes requested):
I'm concerned that if a caller of one of the modified functions tries to use a buffer of size 0, then computing `max_len` by subtracting 1 will underflow to `SIZE_MAX`.

Can you double-check all of the call sites affected by this PR and confirm that can't possibly happen?

Probably best to add to the top of all of the functions that take `max_len` an `Assertion()` that `max_len != SIZE_MAX`.

Goober5000:
> Can you double-check all of the call sites affected by this PR and confirm that can't possibly happen?

Double checked and I fixed the one potential problem area, plus a weird line in `replace_one`.

jg18 (approved):
Looks good. Thanks for the revisions.

PR #4164, opened 2022-04-12
Made a clock to control how often beam impact spews are created

EatThePath:
needs tuning and cleanup still


Goober5000 (commented):
Please use the new type-safe `TIMESTAMP` type introduced in #4156.

EatThePath:
FYI, this is still in draft largely because tests have so far failed to show the expected improvements in performance.

JohnAFernandez:
Hmmmmm, maybe 30 per second is still too much?  If you're running 120 fps that only cuts down the spew by a factor of four, no?  Try 15 per second?

Issue #3629, opened 2021-08-29
F3 lab might not be updating the envmaps?


EatThePath:
This issue is sketch and exists mainly as a way to make sure I check this again later, but the enviromental reflections in the lab seem wrong to me

EatThePath:
evidence mounts, and I ought to be able to do something about it, probably

JohnAFernandez:
Resolving pull request merged.  Don't hesitate to reopen on recurrence.

PR #4165, opened 2022-04-12
Fix for #3629

qazwsxal:
Couple of minor fixes here,

Firstly, envmap_name wasn't zero initialised, so strlen() would return =>1 even if no filename had been found, leading to an erroneous call to bm_load.

Secondly, I added a guard to check that bm_load returned a valid handle, so that environment maps are still generated if precalculated ones are missing.


qazwsxal:
@EatThePath you're assigned to this but it came up while I was fiddling with a new feature. I'd appreciate C&C :)

EatThePath:
I'm not actually familiar with this spot in the code so I'll need more chance to look at context than I have this second, but if someone else wants to do so before I can, by all means. I can say I did build it and it does appear to resolve #3629 without immediately noticeable side effects, and it's a very dramatic improvement.

EatThePath (approved):
looks good to me, now that I've had the chance to look at it's context more.

Issue #4166, opened 2022-04-12
[FR] Preloading of the subspace tunnel


DefCynodont:
(this is written from the experiences of both myself and naomimyselfandi)

When a mission switches from normal space into subspace, the initial load of the subspace tunnel can take long enough that it interferes with the scene transition. Preloading the tunnel seems like it would resolve the issue, but there doesn't seem to be any way to do that. Being able to preload skyboxes would be helpful for a similar reason.

Relatedly, starting a subspace mission also takes several seconds (in my case dozens of seconds) after I click "Commit", during which the subspace ambiance plays but I'm still looking at the briefing (it's fairly trippy). This also means that marking a mission as starting in subspace and immediately cancelling it isn't really a viable option, either.

I don't know how well our machines compare to the average FSO user, but Naomi gets 60 FPS during normal gameplay and I get about 40-ish, we don't have this problem with any assets other than the subspace tunnel.

Goober5000:
Skyboxes are already preloaded - see `sexp_set_skybox_model_preload`.  Looks like the subspace tunnel is never preloaded though.

DefCynodont:
huh, I was unaware there was a way to preload skyboxes, I changed the title.

DefCynodont:
I saw the pull request :)  
I have a small Question: if the mission starts in subspace by default will it still preload? or will it require the use of the mission-set-subspace sexp?

Goober5000:
If the mission starts in subspace by default, it will load the subspace assets before it starts.  This is what it has always done.  The new PR uses the same code - that's the main reason I was able to write it so quickly.

PR #4169, opened 2022-04-13
add preload ability for mission-set-subspace

Goober5000:
Page in the subspace stuff if the mission-set-subspace sexp is used.

Implements #4166.  Note that this will only preload the tunnel for sexps, not scripts.  If using scripts, just add a `(when (true) (mission-set-subspace 0))` as a mission event and it will trigger the preloader.  The preloader only checks for the use of the sexp operator, not whether it actually sets subspace or not.


notimaginative (approved):
LGTM

PR #4175, opened 2022-04-17
Create commandline option for logging data packets

JohnAFernandez:
Creates a mechanism for adding packet types to the debug log.  This is for debugging misaligned packets.  The last "good" packet is usually the culprit.

### DIFF
--- a/code/network/multi.cpp
+++ b/code/network/multi.cpp
@@ -53,6 +53,7 @@
 #include "debugconsole/console.h"
 #include "network/psnet2.h"
 #include "network/multi_mdns.h"
+#include "cmdline/cmdline.h"
 
 // Stupid windows workaround...
 #ifdef MessageBox
@@ -492,6 +493,12 @@ void multi_client_check_server()
 void process_packet_normal(ubyte* data, header *header_info)
 {
 	switch ( data[0] ) {
+
+		// this is for helping to diagnose misaligned packets.  The last sensible packet 
+		// is usually the culprit that needs to be analyzed.
+		if (Cmdline_dump_packet_type) {
+			mprintf(("Packet type of %d received.\n", data[0]));
+		}
 
 		case JOIN:
 			process_join_packet(data, header_info);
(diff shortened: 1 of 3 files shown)


EatThePath (approved):
Looks very straight-forward and good to my eyes.

code/network/multi.cpp
```diff
@@ -493,6 +494,12 @@ void process_packet_normal(ubyte* data, header *header_info)
 {
 	switch ( data[0] ) {
 
+		// this is for helping to diagnose misaligned packets.  The last sensible packet 
+		// is usually the culprit that needs to be analyzed.
+		if (Cmdline_dump_packet_type) {
```
EatThePath, line 499: On second look, shouldn't this if go outside the switch?

EatThePath (changes requested):

code/network/multi.cpp
JohnAFernandez: Yup...... I wasn't rushing at all

EatThePath: The changed version seems a bit extra on white space, but up to you if you want to adjust it imo.

EatThePath (approved):

code/network/multi.cpp
```diff
@@ -491,6 +492,14 @@ void multi_client_check_server()
 
 void process_packet_normal(ubyte* data, header *header_info)
 {
+	// this is for helping to diagnose misaligned packets.  The last sensible packet 
+	// is usually the culprit that needs to be analyzed.
+	if (Cmdline_dump_packet_type) {
+		mprintf(("Packet type of %d received.\n", data[0]));
+	}
+
+
```
Goober5000, line 501: I would remove the extra newlines when you get a chance.

Goober5000 (approved):
Looks good, with a formatting suggestion for the future

notimaginative:
Well, didn't give me time to review :(

I would have gone a little further here and added the total data length being processed and after the switch statement output the number of bytes processed in each packet. You could remove the `\n` from the earlier `mprintf()` and have it all go on the same line rather than have multiple lines in the log for each individual packet being processed.

PR #4176, opened 2022-04-17
Disengage the sound limiters

EatThePath:
Investigation and personal use cases have shown the limit cap of 15 to be unneeded and a hindrance at times, and the 128 channel cap is also at times a bit too low. This removes the first and makes the second dynamic based on system reported capabilities.

### DIFF
--- a/code/sound/ds.cpp
+++ b/code/sound/ds.cpp
@@ -306,12 +306,27 @@ int ds_load_buffer(int *sid, int  /*flags*/, sound::IAudioFile* file)
 }
 
 /**
- * Initialise the ::Channels[] array
+ * Initialise the ::Channels[] array dynamically based on system resources.
  */
 void ds_init_channels()
 {
-	MAX_CHANNELS = Cmdline_no_enhanced_sound ? 32 : 128;
+	// Legacy assumed safe value if for some reason dynamicly setting it does not work
+	MAX_CHANNELS = 128;
 
+	if (Cmdline_no_enhanced_sound)
+		MAX_CHANNELS = 32; // Old sound code limts
+	else {
+		ALCint size;
+		alcGetIntegerv(ds_sound_device, ALC_ATTRIBUTES_SIZE, 1, &size);
+		std::vector<ALCint> attrs(size);
+		alcGetIntegerv(ds_sound_device, ALC_ALL_ATTRIBUTES, size, &attrs[0]);
+		for (size_t i = 0; i < attrs.size(); ++i) {
+			if (attrs[i] == ALC_MONO_SOURCES) {
+				MAX_CHANNELS = attrs[i + 1];
+				mprintf(("  ALC reported %i available sources", attrs[i + 1]));
+			}
+		}
+	}
 	try {
 		Channels = new channel[MAX_CHANNELS];
 	} catch (const std::bad_alloc&) {
--- a/code/sound/ds.cpp
+++ b/code/sound/ds.cpp
@@ -321,9 +321,11 @@ void ds_init_channels()
 		std::vector<ALCint> attrs(size);
 		alcGetIntegerv(ds_sound_device, ALC_ALL_ATTRIBUTES, size, &attrs[0]);
 		for (size_t i = 0; i < attrs.size(); ++i) {
-			if (attrs[i] == ALC_MONO_SOURCES) {
-				MAX_CHANNELS = attrs[i + 1];
-				mprintf(("  ALC reported %i available sources", attrs[i + 1]));
+			//We reserve a portion of the reported channels to avoid colliding with other types of sound
+			//Using the 32 channels of the no_enhanced_sound as a floor.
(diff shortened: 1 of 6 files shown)


EatThePath:
I should add, I've tested this in very sound dense environments and found it both useful and stable as far as I can determine.

code/sound/ds.cpp
```diff
@@ -306,12 +306,27 @@ int ds_load_buffer(int *sid, int  /*flags*/, sound::IAudioFile* file)
...
+
+	if (Cmdline_no_enhanced_sound)
+		MAX_CHANNELS = 32; // Old sound code limts
+	else {
+		ALCint size;
+		alcGetIntegerv(ds_sound_device, ALC_ATTRIBUTES_SIZE, 1, &size);
+		std::vector<ALCint> attrs(size);
+		alcGetIntegerv(ds_sound_device, ALC_ALL_ATTRIBUTES, size, &attrs[0]);
+		for (size_t i = 0; i < attrs.size(); ++i) {
+			if (attrs[i] == ALC_MONO_SOURCES) {
```
notimaginative, line 324: As far as I know (my knowledge might be outdated) this doesn't necessarily report the number of playable sources. Last I checked the spec was unclear about what this actually reports and even then it can vary with the implementation. If OpenAL-Soft is required then that makes things easier, but it should be forced if that's the case. Otherwise you have to do a dynamic check to figure out how many can be played.

notimaginative:
A lot of sources can hinder performance, and that performance hit will vary based on how the sound is played/mixed (2d vs 3d, mono vs stereo) and whether EFX is enabled. And it's all CPU bound with OpenAL Soft. So be careful assuming that it performs well across the board.

Also, the code really isn't written with how OpenAL works in mind. It's the old DirectSound way of doing things made to work with OpenAL with the simplest terms. One major issue is that sources aren't created centrally like code written for OpenAL from the start would require. For instance the streaming audio side of things creates its own sources, so you have to make sure to leave a number of sources free for it to use for music and voice.

Really the code should just be modified such that OpenAL is initialized, creates the number of sources that it can and store the handles in an array, and then when a sound is ready to play the calling code would request an available source as opposed to creating one.

code/sound/ds.cpp
EatThePath: This code was suggested by @TRBlount some time ago and I used it fairly uncritically. Don't we ship with our own openAL and can thus be sure of the implementation? Or is it deeper down than that?

If not, and this code is thus unsafe, my fallback would be to do somthing to the effect of `MAX_CHANNELS = attrs[i + 1]*0.9f;` (plus whatever casting is needed of course) to give a safety factor. The current 128 is a limit I surpass handily when I run tests with limits removed. It doesn't have noticable performance impact for me, it's possible that it will for others but this is a mod that is knowingly high-spec already.

If even that is not acceptable, I'll just drop this section of the changes from the PR, and possibly revisit it later in a separate one.

EatThePath:
> Really the code should just be modified such that OpenAL is initialized, creates the number of sources that it can and store the handles in an array, and then when a sound is ready to play the calling code would request an available source as opposed to creating one.

That kind of refactor seems like a good idea but well beyond the scope of what I'm prepared to undertake at the moment.

notimaginative:
> > Really the code should just be modified such that OpenAL is initialized, creates the number of sources that it can and store the handles in an array, and then when a sound is ready to play the calling code would request an available source as opposed to creating one.
> 
> That kind of refactor seems like a good idea but well beyond the scope of what I'm prepared to undertake at the moment.

Agreed, and I didn't mean to sound like I was suggesting it for this. It's just something we need to look into addressing in the future (for sources and buffers). I was primarily looking to make the point that with the current code we have to be careful about how many sources are being allocated and for what, since we need enough playable sources for at least three different sections of the code that don't know about each other (effects, streaming, capturing).

You're allocating pretty much all of the available sources for effects which means in a busy soundscape voices/music may stop working. I think the 128 limit was, at least in part, taking that into account.

code/sound/ds.cpp
notimaginative: It's definitely wrong with the way the code currently works (see my other comment/clarification) since it probably needs to be 30-40 less than that value for safety. But as long as OpenAL Soft is all we want to work with then the attributes method should work okay I'm guessing. Though it would be nice if, perhaps in another PR, we make that a point with a vendor string check combined with an error or stern warning. Assuming one doesn't already exist (haven't looked in a good number of years)

Since OpenAL Soft mixing is entirely CPU bound, performance is really going to vary based on the individual computer and perhaps the audio backend as well. It might be a good fallback to have a low-medium-high option for people that have sound related performance problems. With "high" being the max/default, "medium" being half that, and "low" being 32. Perhaps it could replace/augment the current cmdline option or something like that.

EatThePath: I've pushed some changes to preserve an amount of headroom.

notimaginative (approved):
Should work fine until proper source and buffer pools are implemented.

PR #4177, opened 2022-04-17
various HUD gauge fixes

### LINKED ISSUE #4171: Scripting regression in hudgauge
Axem's prompt-box script broke on the 20220207 nightly, bisecting a blue planet crash narrowed down to there. As a reproduction download blue planet and play Age of Aquarius, the prompt box is used immediately on the start of the first non-cutscene mission.

@Goober5000 #4003 looks like the most likely cause, but I'm not immediately sure of why. 

The error in question is
```
[Window Title]
Error!

[Content]
LUA ERROR: [string "proBox-sct.tbm - On Game Init"]:87: attempt to index a nil value

------------------------------------------------------------------

stack traceback:
	[C]: ?
	[string "proBox-sct.tbm - On Game Init"]:87: in function 'GetTextFromGauges'
	[string "proBox-sct.tbm - On Game Init"]:64: in function 'UpdateText'
	[string "proBox-sct.tbm - On Game Init"]:54: in function 'MakeList'
	[string "proBox-sct.tbm - On Game Init"]:198: in function 'proBoxOn'
	[string "String: proBoxOn()"]:1: in main chunk
------------------------------------------------------------------

------------------------------------------------------------------


[ This info is in the clipboard so you can paste it somewhere now ]


[Debug] [Continue] [Exit]
```
and the function in question, 
```

function PromptBox:GetTextFromGauges()
	
	local t = {}
	for i, v in pairs(self.Gauges) do
		local gauge = hu.getHUDGaugeHandle(v).Text
			if gauge:gsub("%s+", "") ~= "" and not self.Locked[i] then
				t[i] = ">> " .. gauge
			else
				t[i] = ""
			end
	end
	
	return t

end
``` 
The gauges are there in the files, and 20220205 displays them just fine.

Goober5000:
1) Fix part of the lookup logic in `hud_get_default_gauge_index`
2) Add back part of the lookup that was removed in #4003
3) Use custom checks with the `default_hud_gauges` list
4) Add a clarification comment to `default_hud_gauges`

Fixes #4171

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -57,6 +57,7 @@
 #include "weapon/emp.h"
 #include "weapon/weapon.h"
 
+// This contains not only the retail HUD gauges but also any custom HUD gauges that are not specific to a ship.
 SCP_vector<std::unique_ptr<HudGauge>> default_hud_gauges;
 
 // new values for HUD alpha
@@ -3864,7 +3865,7 @@ HudGauge* hud_get_custom_gauge(const char* name, bool check_all_gauges)
 			}
 		}
 		for (auto &gauge : default_hud_gauges) {
-			if (!stricmp(name, gauge->getCustomGaugeName())) {
+			if (gauge->isCustom() && !stricmp(name, gauge->getCustomGaugeName())) {
 				return gauge.get();
 			}
 		}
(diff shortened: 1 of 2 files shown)


notimaginative:
Should `hud_get_custom_gauge()` really be using an if-else? It looks like it should actually work as a fall-through instead with the all_gauges check first, then custom player ship check, falling back to default guages, then nullptr. I don't really know the code or problem it's fixing, so maybe I'm off base, but that jumped out at me.

Otherwise it LGTM.

Goober5000:
I just went back to the pre-existing code, but yeah, that's a good plan for cleaning up the function a bit.

notimaginative (approved):

PR #4180, opened 2022-04-18
Log data packets2

JohnAFernandez:
Upgrade packet dumping command line option to include the amount of data that has been processed from the packet so far.  Also, upgrade the `nprintf` that dumps unknown packet types to an `mprintf` because even power users very rarely use the debug filter.

### DIFF
--- a/code/network/multi.cpp
+++ b/code/network/multi.cpp
@@ -951,7 +951,7 @@ void process_packet_normal(ubyte* data, header *header_info)
 // process_tracker_packet() as defined in MultiTracker.[h,cpp]
 void multi_process_bigdata(ubyte *data, int len, net_addr *from_addr, int reliable)
 {
-	int type, bytes_processed;
+	int bytes_processed;
 	int player_num;
 	header header_info;
 	ubyte *buf;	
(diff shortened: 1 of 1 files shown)


code/network/multi.cpp
```diff
@@ -932,11 +924,17 @@ void process_packet_normal(ubyte* data, header *header_info)
...
+			mprintf(("Received packet with unknown type %d\n", data[0] ));
 			header_info->bytes_processed = 10000;
 			break;
 
 	} // end switch
+
+	// this is for helping to diagnose misaligned packets.  The last sensible packet 
+	// is usually the culprit that needs to be analyzed.
+	if (Cmdline_dump_packet_type) {
+		mprintf(("Packet type of %d received.  Total amount of data processed from larger packet is %d.\n", data[0], header_info->bytes_processed));
```
notimaginative, line 936: Then just remove the `Packet type of %d received.` part and this line will continue on the earlier one.

notimaginative (changes requested):
Another thing that would be useful is the `len` value from `multi_process_bigdata()` as that will give the raw packet size, which could contain multiple game packets. That way we'd know how much should actually be processed. 

Multiple ways to do that so whichever you think is easiest/cleaner. It should preferably end up as part of the same log line though so that it's more obvious that the game packets being processed might be part of a larger raw packet.

notimaginative (approved):

PR #4186, opened 2022-04-19
background bitmap fix

Goober5000:
Part 1 - Adds the framework for loading and saving backgrounds with correct or incorrect angles.  Backgrounds in FSO will all have fixed angles internally, and newly created backgrounds in FRED will have fixed angles, but existing mission files will have incorrect angles.  The parser will determine whether the mission file contains fixed or unfixed angles and act accordingly.  (Actually fixing the angles will be done in part 2.)

Part 2 by @Baezon - Uses simpler and more correct math for bg bitmap angles, and when old values are parsed, they are translated into correct values


jg18 (changes requested):
As discussed with @Goober5000 over Discord, the expression "fixed angles" could be misinterpreted to mean "pre-determined/unchanging angles", so I've asked for it to be changed to "corrected angles" or similar.

code/parse/sexp.h
```diff
@@ -808,6 +808,8 @@ struct ship_obj;
 #define OP_CONTAINER_REMOVE_FROM_MAP		(0x004b | OP_CATEGORY_CHANGE2 | OP_NONCAMPAIGN_FLAG)	// Karajorma/jg18
 #define OP_CONTAINER_GET_MAP_KEYS			(0x004c | OP_CATEGORY_CHANGE2 | OP_NONCAMPAIGN_FLAG)	// Karajorma/jg18
 #define OP_CLEAR_CONTAINER					(0x004d | OP_CATEGORY_CHANGE2 | OP_NONCAMPAIGN_FLAG)	// Karajorma/jg18
+#define OP_ADD_BACKGROUND_BITMAP2			(0x004e | OP_CATEGORY_CHANGE2 | OP_NONCAMPAIGN_FLAG)	// Goober5000
+#define OP_ADD_SUN_BITMAP2					(0x004f | OP_CATEGORY_CHANGE2 | OP_NONCAMPAIGN_FLAG)	// Goober5000
```
jg18, line 812: I have a slight preference for a SEXP name like `add-sun-bitmap-new` or some other name to indicate that this SEXPs should be used instead of `add-sun-bitmap`, and similar for backgrounds.

A name like `add-sun-bitmap2` could just mean a variant of the original, rather than the replacement, even if the original is marked in the SEXP help text as deprecated.

code/mission/missionparse.cpp
```diff
@@ -5498,6 +5522,9 @@ void parse_one_background(background_t *background)
 		// add it
 		background->bitmaps.push_back(sle);
 	}
+
+	// always set the flag when we're done parsing, because the angles are now stored correctly and we want to save them correctly
+	background->flags.set(Starfield::Background_Flags::Corrected_angles_in_mission_file);
```
jg18, line 5527: This line confuses me. It makes it sound like the corrected angles have been saved to the mission file, when what we've actually done is correct the background's angles that are stored in memory.

jg18 (changes requested):
A few more requests/questions.

code/parse/sexp.h
Goober5000: Changed.  Also changed `add-to-colgroup2` and `remove-from-colgroup2` since they follow the same pattern.

code/parse/sexp.cpp
```diff
@@ -553,8 +553,8 @@ SCP_vector<sexp_oper> Operators = {
 	{ "collide-invisible",				OP_COLLIDE_INVISIBLE,					1,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// Goober5000
 	{ "add-to-collision-group",			OP_ADD_TO_COLGROUP,						2,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// The E
 	{ "remove-from-collision-group",	OP_REMOVE_FROM_COLGROUP,				2,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// The E
-	{ "add-to-collision-group2",		OP_ADD_TO_COLGROUP2,					2,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// Goober5000
-	{ "remove-from-collision-group2",	OP_REMOVE_FROM_COLGROUP2,				2,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// Goober5000
+	{ "add-to-collision-group-new",		OP_ADD_TO_COLGROUP_NEW,					2,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// Goober5000
+	{ "remove-from-collision-group-new",	OP_REMOVE_FROM_COLGROUP_NEW,		2,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// Goober5000
```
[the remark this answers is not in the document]
Goober5000: I think it's close enough to the other request (to change `add-*-bitmap2`) to be in scope.

code/parse/sexp.cpp
```diff
@@ -25306,8 +25318,10 @@ int eval_sexp(int cur_node, int referenced_node)
 				break;
 
 			case OP_ADD_SUN_BITMAP:
+			case OP_ADD_SUN_BITMAP_NEW:
 			case OP_ADD_BACKGROUND_BITMAP:
-				sexp_add_background_bitmap(node, op_num == OP_ADD_SUN_BITMAP);
+			case OP_ADD_BACKGROUND_BITMAP_NEW:
+				sexp_add_background_bitmap(node, op_num == OP_ADD_SUN_BITMAP, op_num == OP_ADD_SUN_BITMAP_NEW || op_num == OP_ADD_BACKGROUND_BITMAP_NEW);
```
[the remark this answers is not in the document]
Goober5000: Good catch!

code/starfield/starfield_flags.h
```diff
@@ -0,0 +1,15 @@
+#ifndef STARFIELD_FLAGS_H
+#define STARFIELD_FLAGS_H
+
+#include "globalincs/flagset.h"
+
+namespace Starfield {
+
+	FLAG_LIST(Background_Flags){
+		Corrected_angles_in_mission_file = 0,	// angles are saved to the mission file correctly; if this flag is not present they are saved using incorrect math
```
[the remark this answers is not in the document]
Goober5000: The angles are *always* stored correctly in memory now.  This flag just contains whether correct angles are saved or incorrect angles are saved.

code/mission/missionparse.cpp
Goober5000: We have corrected the angles that are stored in memory, yes.  Now we need to choose whether to save them correctly or save them incorrectly, so the default is to save them correctly.  But this can be changed in the background editor.

code/starfield/starfield_flags.h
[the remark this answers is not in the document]
jg18: Well, since saving corrected angles is now the default behavior, perhaps it'd be more logical to make the flag represent the exception, e.g., `Save_angles_using_legacy_format`?

Or, failing that, I suggest renaming this flag to something like `Save_angles_as_corrected`.

Or, better still, maybe use this flag only when FRED is running?

The problem with `_in_mission_file` is that it makes it sound like the angles in the mission file on disk are saved in corrected format, which may not necessarily be the case. For example, if we're running FSO and loading an old mission file that doesn't use corrected angles, then applying this flag doesn't make sense to me.

What do you think?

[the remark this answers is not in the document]
Goober5000: Let me try to clarify.  The angles are always correct *in memory*, that is, in the internal game state, so that the correct angles are available for scripts and such.  But they are only saved in the correct format if the flag is enabled.  If the flag is not enabled, they are saved in the incorrect format.

When the mission is loaded, the angles are converted if the flag is not found, and left alone if the flag is found.

The flag must be set up so that it is active if and only if the background is saved to the mission file with correct angles.  This is because older missions with incorrect angles don't have the flag.

code/mission/missionparse.cpp
```diff
@@ -5447,6 +5463,10 @@ void parse_one_background(background_t *background)
 		stuff_float(&sle.ang.b);
 		stuff_float(&sle.ang.h);
 
+		// correct legacy bitmap angles which used incorrect math in older versions
+		if (!background->flags[Starfield::Background_Flags::Corrected_angles_in_mission_file])
+			stars_correct_background_angles(&sle.ang);
```
jg18, line 5468: I wonder if it might be worth writing to the debug log whether we're making this correction. Ditto for the identical code below while parsing data on background bitmaps.

code/parse/sexp.cpp
```diff
@@ -14110,6 +14120,8 @@ void sexp_add_background_bitmap(int n, bool is_sun)
 	eval_angles(&sle.ang, n, is_nan, is_nan_forever);
 	if (is_nan || is_nan_forever)
 		return;
+	if (!uses_correct_angles)
+		stars_correct_background_angles(&sle.ang);
```
jg18, line 14124: Similar to parsing above, I wonder if it might be worth writing to the debug log whether we're making this correction and whether it's for a sun or for a background.

OTOH that's equivalent to logging whether the mission uses the deprecated SEXPs for adding suns/backgrounds, the new ones, or (eep) a mix.

code/scripting/api/libs/mission.cpp
```diff
@@ -1337,6 +1331,9 @@ ADE_FUNC(addBackgroundBitmap,
 	}
 
 	sle.ang     = *orient.GetAngles();
+	if (!uses_correct_angles)
```
jg18, line 1334: Also wondering whether it's worth logging to the debug log which scripting function (deprecated or new) was called. Ditto for the functions for adding sun bitmaps.

Then again, does the fact that the old functions are deprecated result in a debug log message whenever they're called?

code/starfield/starfield.cpp
```diff
@@ -2668,6 +2688,16 @@ void stars_delete_entry_FRED(int index, bool is_a_sun)
 	}
 }
 
+// Goober5000
+void stars_add_blank_background(bool creating_in_fred)
+{
+	Backgrounds.emplace_back();
+
+	// any background that is created will have correct angles, so be sure to set the flag
+	if (creating_in_fred)
```
jg18, line 2697: Why do we need this parameter? Why can't we just check `Fred_running` here?

code/starfield/starfield.cpp
```diff
@@ -2668,6 +2688,16 @@ void stars_delete_entry_FRED(int index, bool is_a_sun)
...
+		Backgrounds.back().flags.set(Starfield::Background_Flags::Corrected_angles_in_mission_file);
```
jg18, line 2698: If the new checkbox in the Background Editor is unchecked, where does the flag get unset? In `background_flags_close()`?

jg18 (approved):
Some non-blocking questions/comments but looks good overall. Thanks for making the requested changes. 👍 

The most important question is the one about `creating_in_fred` in `stars_add_blank_background()`.

code/mission/missionparse.cpp
```diff
@@ -5430,9 +5430,25 @@ void parse_reinforcements(mission *pm)
...
 {
 	// clear here too because this function can be called from more than one place
+	background->flags.reset();
 	background->suns.clear();
 	background->bitmaps.clear();
 
+	// we might have some flags
+	if (optional_string("+Flags:"))
+	{
+		// we'll assume the list will contain no more than 4 distinct tokens
```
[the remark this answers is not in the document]
Goober5000: It will truncate the list to only 4.  But as it won't have the ability to recognize more flags, this won't matter much.

code/mission/missionparse.cpp
```diff
@@ -5430,9 +5430,25 @@ void parse_reinforcements(mission *pm)
...
+		char flag_strings[4][NAME_LENGTH];
```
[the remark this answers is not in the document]
Goober5000: This is a "cross that bridge when you come to it" situation.  I only needed support for one flag, but I wanted the code to be easily expandable to more flags if needed in the future.

code/mission/missionparse.cpp
Goober5000: I don't think that's necessary, but I'll leave the PR open in case anyone else wants to comment.

code/parse/sexp.cpp
Goober5000: Same as above - I don't think a debug statement is necessary.

code/scripting/api/libs/mission.cpp
Goober5000: Not a debug log message per se, but it will emit an error if the scripting function is called by a mod that requires a future version.

code/starfield/starfield.cpp
Goober5000: Because the code must run only when FRED is creating a new background.  It can't run when FRED is running but merely loading the file.

code/starfield/starfield.cpp
Goober5000: Yes.

PR #4190, opened 2022-04-20
More Pof space optimizations

Baezon:
Adds the `SORTNORM2` chunk and merges the `BOUNDBOX` chunk into `TMAP2POLY` as part of version 2300 (more details [here](https://github.com/Baezon/pof-tools/issues/36#issuecomment-1101733254)). Technically, a 'breaking' change, but since it just introduced and there's no pofs in the wild with this version yet it should be ok.

Also removes several dead bsp parsing functions.


TRBlount (approved):
Already looked at this earlier. Should be good.

Issue #3587, opened 2021-08-04
Particles FR: Ability to set spawn chance for particles below 50%


Kestrellius:
At present, there is no way to make a particle effect that will spawn particles less than 50% of the time. This is because the lowest possible range for number of particles (other than just zero) is (0 1).

In various situations -- especially things like beam impacts, where the engine will fire off the particle effect every frame -- it would be useful to have a lower probability than this, e.g. an effect that will only successfully spawn a particle every one hundred times it's activated.

One conceptually simple way to implement this would be to allow the number parameter to parse negative numbers, but clamp negatives to zero once a number has been chosen. That is: say that you want a .25 chance of spawning a particle. You set the number of particles to (-2 1).

Then, the game randomly chooses a number: -2, -1, 0, or 1. But if -2 or -1 is chosen, they get clamped to 0. So the actual distribution is 0, 0, 0, 1. With a little bit of arithmetic, it would be easy enough to use this technique to generate a wide range of probabilities.

A more robust method of implementation would be a simple new parameter: spawn chance. It would default to 100, resulting in the existing behavior, but at, say 70, the effect would have a thirty percent chance to "fail" and not spawn any of its particles. (If the effect has a chance to spawn zero particles, then that would be something that happen after a "success" -- not factored into this calculation.)

(@EatThePath asked to be pinged in this FR.)

talhydras:
Warmachine could take huge advantage of improved particle spawning functionality - we heavily use short duration beams, so having more precise control over spawning of particles would allow us to make more visually appealing impact fx as well as reduce unnecessary particle spawns.

Kestrellius:
Feature has now been implemented by EatThePath.

PR #4191, opened 2022-04-21
Random chance of particle spawning

EatThePath:
Adds a random spawn chance to particle effects. Implements #3587, with testing by requestor. Originally submitted in draft while testing was completed, now ready.

### DIFF
--- a/code/particle/effects/GenericShapeEffect.h
+++ b/code/particle/effects/GenericShapeEffect.h
@@ -171,9 +171,8 @@ class GenericShapeEffect : public ParticleEffect {
 		if (optional_string("+Chance:")) {
 			float chance;
 			stuff_float(&chance);
-			CLAMP(chance, 0.0f, 1.0f);
 			if (chance <= 0.0f) {
-				Error(LOCATION,
+				Warning(LOCATION,
 					"Particle %s tried to set +Chance: %f\nChances below 0 would result in no particles.",
 					m_name.c_str(), chance);
 			} else if (chance > 1.0f) {
--- a/code/particle/effects/GenericShapeEffect.h
+++ b/code/particle/effects/GenericShapeEffect.h
@@ -1,5 +1,7 @@
 #ifndef GENERIC_SHAPE_EFFECT_H
 #define GENERIC_SHAPE_EFFECT_H
+#include "osapi/dialogs.h"
+#include "parse/parselo.h"
 #pragma once
 
 #include "globalincs/pstypes.h"
@@ -32,7 +34,8 @@ class GenericShapeEffect : public ParticleEffect {
 	ConeDirection m_direction = ConeDirection::Incoming;
 	::util::UniformFloatRange m_velocity;
 	::util::UniformUIntRange m_particleNum;
-
+	float m_particleChance;
+	::util::UniformFloatRange m_particleRoll;
 	ParticleEffectHandle m_particleTrail = ParticleEffectHandle::invalid();
 
 	util::EffectTiming m_timing;
@@ -102,6 +105,12 @@ class GenericShapeEffect : public ParticleEffect {
 			matrix dirMatrix;
 			vm_vector_2_matrix(&dirMatrix, &dir, nullptr, nullptr);
 			for (uint i = 0; i < num; ++i) {
+				if (m_particleChance < 1.0f) {
+					auto roll = m_particleRoll.next();
+					if (roll <= 0.0f)
+						continue;
+				}
+
 				matrix velRotation = m_shape.getDisplacementMatrix();
 
 				matrix rotatedVel;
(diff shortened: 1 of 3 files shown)


Baezon:
If I might suggest a simpler API, `m_particleNum` could be changed to a float range instead. All other code would remain largely the same, except in the `for (uint i = 0; i < num; ++i)` loop (which would be using floats instead), when `i` is between 0 and 1, to do a final roll where the remainder is the chance.

This has the benefit of not needing any extra parsing or data, simplifying the code, allows old particle effects to continue working identically, and works fairly intuitively for non-integers, where they get treated as a random chance.

EatThePath:
> If I might suggest a simpler API, `m_particleNum` could be changed to a float range instead. All other code would remain largely the same, except in the `for (uint i = 0; i < num; ++i)` loop (which would be using floats instead), when `i` is between 0 and 1, to do a final roll where the remainder is the chance.

After consideration, I'd rather not. That doesn't seem especially intuitive to a user's perspective to me, at least not more intuitive than having the option to just have a simple %chance definition. Also having both range and chance also opens the door to tilted random distributions, (100),0.99 and (100),0.01 are very different things from (0,100), for example. I do not immediately know what someone might want to do with that, but it feels like something that might get used more than the ability to give the spawn counts fuzzy edges. 

That said, 'fuzzy edges' may also be useful to someone working in the low count ranges. (1.0,1.25) would produce behavior not achievable otherwise. That may be worth implementing at some point, but as supplement, not replacement.

Also, 
> allows old particle effects to continue working identically
My code as it stands shouldn't change the behavior of existing particles without a +Chance: value at all. If it does, that's an oversight I will have to correct.

EatThePath:
@[Kestrellius](https://github.com/Kestrellius) reports successful testing, and also approves of the current implementation, so I'm switching to submitted

code/particle/effects/GenericShapeEffect.h
```diff
@@ -156,6 +165,26 @@ class GenericShapeEffect : public ParticleEffect {
...
 		}
+		if (!nocreate) {
+			m_particleChance = 1.0f;
+		}
+		if (optional_string("+Chance:")) {
+			float chance;
+			stuff_float(&chance);
+			CLAMP(chance, 0.0f, 1.0f);
+			if (chance <= 0.0f) {
+				Error(LOCATION,
```
MageKing17, line 176: You've already clamped `chance`, so it can't be less than zero. Additionally, this doesn't seem like an unrecoverable problem, so I'm not sure why this is an `Error` while the similar `> 1.0f` case (which, again, can't execute because the value has already been clamped) is only a `Warning`.

code/particle/effects/GenericShapeEffect.h
```diff
@@ -156,6 +165,26 @@ class GenericShapeEffect : public ParticleEffect {
...
+					m_name.c_str(), chance);
+			} else if (chance > 1.0f) {
+				Warning(LOCATION,
+					"Particle %s tried to set +Chance: %f\nChances above 1 are ignored, please use +Number: (min,max) "
+					"to spawn multiple particles.", m_name.c_str(), chance);
+				chance = 1.0f;
+			}
+			m_particleChance = chance;
+		}
+		m_particleRoll = ::util::UniformFloatRange(m_particleChance - 1.0f, m_particleChance);
```
MageKing17, line 187: Er... so if `m_particleChance` is `0.5f`, you want to generate random values between `0.5f` and `0.5f`?

code/particle/effects/GenericShapeEffect.h
```diff
@@ -102,6 +105,12 @@ class GenericShapeEffect : public ParticleEffect {
 			matrix dirMatrix;
 			vm_vector_2_matrix(&dirMatrix, &dir, nullptr, nullptr);
 			for (uint i = 0; i < num; ++i) {
+				if (m_particleChance < 1.0f) {
+					auto roll = m_particleRoll.next();
+					if (roll <= 0.0f)
```
MageKing17, line 110: Because I'm pretty sure `0.5f` will never be `<= 0.0f`.

code/particle/effects/VolumeEffect.cpp
```diff
@@ -79,6 +85,31 @@ namespace particle {
...
+					Warning(LOCATION,
+						"Particle %s tried to set +Chance: %f\nChances above 1 are ignored, please use +Number: "
+						"(min,max) to spawn multiple particles.",
+						m_name.c_str(),
+						chance);
+					chance = 1.0f;
+				}
+				m_particleChance = chance;
+			}
+			m_particleRoll = ::util::UniformFloatRange(m_particleChance - 1.0f, m_particleChance);
```
MageKing17, line 111: Same problems with this file.

MageKing17 (changes requested):
A few bits of code that confuse me.

code/particle/effects/GenericShapeEffect.h
EatThePath: If m_particleChance is `0.5f`, then this is `(0.5f-1.0f, 0.5f)`, or `-0.5f` to `0.5f`.

MageKing17: Ah, right, for some reason I was reading it as `1.0f - m_particleChance`; now it makes sense.

code/particle/effects/VolumeEffect.cpp
MageKing17: Okay, not this exact line, but the `CLAMP`/`Error` stuff from above.

code/particle/effects/GenericShapeEffect.h
EatThePath: the point about clamping is true. I must have written that before deciding to do the more elaborate handling and forgotten to remove it.
My thinking with warning vs error was rooted in just a gut sense of how wrong each case was, a >100% chance to spawn is arguably modeled well by 100%, but I have no idea what someone who set for <0% wants. But, it would make sense to treat both as warnings now that you question it.

MageKing17: Well, regardless of what they *wanted*, they've clearly told the engine never to spawn the particle. Which gives a very simple way to recover from this issue: never spawn the particle. Therefore it doesn't really make sense (to me) to halt program execution here.

EatThePath: Fair, I'll adjust.

MageKing17 (approved):

PR #4192, opened 2022-04-21
probably fix #4187

### LINKED ISSUE #4187: Assert upon player death
> Assert: "form_objnum >= 0"
> File: aicode.cpp
> Line: 3466
> Invalid object number in ai_formation_object_recalculate_slotnums!

The player was killed by a large space station explosion that also killed many other nearby objects as well (likely including the player's wingmates).

Goober5000:
My guess is that, in `ai_ship_destroy`, if the dead ship's goal object has already died in the same frame, the `goal_objnum` will be -1 when `ai_formation_object_recalculate_slotnums` is about to be called.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -15916,7 +15916,7 @@ void ai_ship_destroy(int shipnum)
 			aip->hitter_objnum = -1;
 	}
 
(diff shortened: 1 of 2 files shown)


JohnAFernandez (approved):
If it *doesn't* fix it, we are at least adding a sensible check.

Baezon (approved):

PR #4194, opened 2022-04-21
Fix FRED accessibility of new flag `ai-attackable-if-no-collide` V2

wookieejedi:
Turns out the `MAX_OBJECT_FLAG_NAMES` was not increased originally, so this PR increases it to match the new amount now that `ai-attackable-if-no-collide` flag was added.


JohnAFernandez (approved):
Lame magic numbers are lame.  One-liners, however, are awesome.

PR #4198, opened 2022-04-21
LuaAI

BMagnu:
This PR adds the ability to add AI modes backed by scripting, similar to scripted SEXPs.
Additionally, these AI goals can be (in a limited fashion) registered to work as player orders (given through the comms menu).

An example of how this works is attached to the PR [here](https://github.com/scp-fs2open/fs2open.github.com/files/8535404/LuaAI.zip).

There are currently a few limitations and caveats to this.

General limits of the system with no improvement planned:
- Lua AI is limited to one target argument at most
- When a Lua AI goal is registered as a player order, it can only accept no targets or a ship as a target, not any oswpt type

Limits of the system that _could_ be improved or fixed, but with no immediate plans to do so:
- Since player orders themselves are limited to 32, Lua AI can only register about 16 more (possibly less in the future. SEXP-only AI modes, however, are unlimited) 
- Lua AI cannot be used for initially set goals of ships. They must be added by SEXP or by player order

Limits of the system that have soon-ish followup improvements already planned
- FRED saves allowed orders as a bitfield. This makes them susceptible to table loading order. In the future, this will be changed to an order-independent flagset. As that is a fairly large change however, it is deferred into a future PR
- Lua AI will play the Yessir message when recieving a new player order. Making that configurable is halfway implemented, but needs deliberation as to how to expose these messages to the table format
- FSO and FRED currently do not play nice with having more than 10 valid player orders for any objecttype. This limit will be removed in a future PR
- _Edit:_ Currently, scripting goals are always valid. Allowing the user to optionally define a function that tests whether the goal is currently valid for execution is planned for a future PR

Caveats that a user might stumble over if not aware of:
- The AI mode itself will be valid for all object types by default. If registered as a player order however, it needs to be added to the allowed player orders in the ``+Player Orders:	`` field of the objecttypes.tbl of the specific objecttype
- Lua AI will not take care of implausible orders for the scripter. If, for example, the target dies while the AI mode is active, it is the responsibility of the scripter to check for this and abort the AI mode if needed.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -12905,7 +12905,7 @@ void ai_maybe_evade_locked_missile(object *objp, ai_info *aip)
 				case AIM_WARP_OUT:
 					break;
 				default:
-					UNREACHABLE("Unknown AI Mode! Get a coder!");
+					UNREACHABLE("Unknown AI Mode %d! Get a coder!", aip->mode);
 					break;
 				}
 			}
@@ -12966,7 +12966,7 @@ void maybe_evade_dumbfire_weapon(ai_info *aip)
 	case AIM_LUA:
 		return;
 	default:
-		UNREACHABLE("Unknown AI Mode! Get a coder!");
+		UNREACHABLE("Unknown AI Mode %d! Get a coder!", aip->mode);
 		return;
 	}
 
@@ -13035,7 +13035,7 @@ void maybe_evade_dumbfire_weapon(ai_info *aip)
 		case AIM_LUA:
 			break;
 		default:
-			UNREACHABLE("Unknown AI Mode! Get a coder!");
+			UNREACHABLE("Unknown AI Mode %d! Get a coder!", aip->mode);
 		}
 	}
 }
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -15704,6 +15704,8 @@ void ai_ship_hit(object *objp_ship, object *hit_objp, vec3d *hit_normal)
 	case AIM_WARP_OUT:
 		return;
 		break;
+	case AIM_LUA:
+		return;
 	default:
 		Int3();	//	Bogus mode!
 	}
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -12900,11 +12900,12 @@ void ai_maybe_evade_locked_missile(object *objp, ai_info *aip)
 				case AIM_PLAY_DEAD:
 				case AIM_BAY_DEPART:
 				case AIM_SENTRYGUN:
+				case AIM_LUA:
 					break;
 				case AIM_WARP_OUT:
 					break;
 				default:
-					Int3();			//	Hey, what mode is it?
+					UNREACHABLE("Unknown AI Mode! Get a coder!");
 					break;
 				}
 			}
@@ -12965,7 +12966,7 @@ void maybe_evade_dumbfire_weapon(ai_info *aip)
 	case AIM_LUA:
 		return;
 	default:
-		Int3();	//	Bogus mode!
+		UNREACHABLE("Unknown AI Mode! Get a coder!");
 		return;
 	}
 
@@ -13034,7 +13035,7 @@ void maybe_evade_dumbfire_weapon(ai_info *aip)
 		case AIM_LUA:
 			break;
 		default:
-			Int3();	//	Bogus mode!
+			UNREACHABLE("Unknown AI Mode! Get a coder!");
 		}
 	}
 }
@@ -13621,7 +13622,7 @@ void ai_execute_behavior(ai_info *aip)
 		ai_lua(&Ships[Pl_objp->instance]);
 		break;
 	default:
-		Int3();		//	This should never happen -- MK, 5/12/97	
+		UNREACHABLE("Unknown AI Mode! Get a coder!");
 		break;
 	}
 
@@ -15707,7 +15708,7 @@ void ai_ship_hit(object *objp_ship, object *hit_objp, vec3d *hit_normal)
 	case AIM_LUA:
 		return;
 	default:
-		Int3();	//	Bogus mode!
+		UNREACHABLE("Unknown AI Mode! Get a coder!");
 	}
 
 	if (timestamp_elapsed(aip->ok_to_target_timestamp)) {
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -13619,7 +13619,7 @@ void ai_execute_behavior(ai_info *aip)
 		// properly until after 3.6.7 just to avoid delaying release or breaking something - taylor
 		break;
 	case AIM_LUA:
-		ai_lua(&Ships[Pl_objp->instance]);
+		ai_lua(aip);
 		break;
 	default:
 		UNREACHABLE("Unknown AI Mode! Get a coder!");
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -22,6 +22,7 @@
 #include "ai/aibig.h"
 #include "ai/aigoals.h"
 #include "ai/aiinternal.h"
+#include "ai/ailua.h"
 #include "asteroid/asteroid.h"
 #include "autopilot/autopilot.h"
 #include "cmeasure/cmeasure.h"
@@ -133,6 +134,8 @@ const char *Mode_text[MAX_AI_BEHAVIORS] = {
 	"BAY_DEPART",
 	"SENTRYGUN",
 	"WARP_OUT",
+	"FLY_TO_SHIP",
+	"LUA_AI"
 };
 
 //	Submode text is only valid for CHASE mode.
@@ -12959,6 +12962,7 @@ void maybe_evade_dumbfire_weapon(ai_info *aip)
 	case AIM_SENTRYGUN:
 	case AIM_WARP_OUT:
 	case AIM_FLY_TO_SHIP:
+	case AIM_LUA:
 		return;
 	default:
 		Int3();	//	Bogus mode!
@@ -13027,6 +13031,7 @@ void maybe_evade_dumbfire_weapon(ai_info *aip)
 		case AIM_BAY_EMERGE:
 		case AIM_BAY_DEPART:
 		case AIM_SENTRYGUN:
+		case AIM_LUA:
 			break;
 		default:
 			Int3();	//	Bogus mode!
@@ -13611,6 +13616,9 @@ void ai_execute_behavior(ai_info *aip)
 	case AIM_GET_BEHIND:
 		// FIXME: got this from TBP and added it here to skip the Int3() but don't really want to handle it
 		// properly until after 3.6.7 just to avoid delaying release or breaking something - taylor
+		break;
+	case AIM_LUA:
+		ai_lua(&Ships[Pl_objp->instance]);
 		break;
 	default:
 		Int3();		//	This should never happen -- MK, 5/12/97	
--- a/code/ai/aigoals.cpp
+++ b/code/ai/aigoals.cpp
@@ -11,6 +11,7 @@
 
 
 #include "ai/aigoals.h"
+#include "ai/ailua.h"
 #include "globalincs/linklist.h"
 #include "mission/missionlog.h"
 #include "mission/missionparse.h"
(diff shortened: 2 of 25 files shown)


code/ai/aicode.cpp
```diff
@@ -15758,6 +15766,8 @@ void ai_ship_hit(object *objp_ship, object *hit_objp, vec3d *hit_normal)
 	case AIM_WARP_OUT:
 		return;
 		break;
+	case AIM_LUA:
+		return;
 	default:
 		Int3();	//	Bogus mode!
```
JohnAFernandez, line 15772: Since you're in the area, can you replace these INT3's with some calls to UNREACHABLE.

code/ai/aicode.cpp
```diff
@@ -13677,8 +13683,11 @@ void ai_execute_behavior(ai_info *aip)
 		// FIXME: got this from TBP and added it here to skip the Int3() but don't really want to handle it
 		// properly until after 3.6.7 just to avoid delaying release or breaking something - taylor
 		break;
+	case AIM_LUA:
+		ai_lua(&Ships[Pl_objp->instance]);
```
Goober5000, line 13687: If you want to make Lua AI applicable to anything in the OSWPT package, then I don't think `&Ships[Pl_objp->instance]` is appropriate here

code/ai/aigoals.cpp
```diff
@@ -1419,7 +1459,8 @@ int ai_mission_goal_achievable( int objnum, ai_goal *aigp )
 	//  these orders are always achievable.
 	if ( (aigp->ai_mode == AI_GOAL_KEEP_SAFE_DISTANCE)
 		|| (aigp->ai_mode == AI_GOAL_CHASE_ANY) || (aigp->ai_mode == AI_GOAL_STAY_STILL)
-		|| (aigp->ai_mode == AI_GOAL_PLAY_DEAD) || (aigp->ai_mode == AI_GOAL_PLAY_DEAD_PERSISTENT) )
+		|| (aigp->ai_mode == AI_GOAL_PLAY_DEAD) || (aigp->ai_mode == AI_GOAL_PLAY_DEAD_PERSISTENT) 
+		|| (aigp->ai_mode == AI_GOAL_LUA))
```
Goober5000, line 1463: Does it make sense to delegate the achievable check to a scripting function?

code/ai/aigoals.h
```diff
@@ -103,14 +105,14 @@ extern int ai_goal_num(ai_goal *goals);
...
+extern void ai_add_goal_sub_sexp( int sexp, int type, ai_goal *aigp, char *actor_name);
 
 extern int ai_remove_goal_sexp_sub( int sexp, ai_goal* aigp );
 extern void ai_remove_wing_goal_sexp( int sexp, wing *wingp );
 
 // adds goals to ships/sings through player orders
-extern void ai_add_ship_goal_player( int type, int mode, int submode, char *shipname, ai_info *aip );
-extern void ai_add_wing_goal_player( int type, int mode, int submode, char *shipname, int wingnum );
+extern void ai_add_ship_goal_player( int type, int mode, int submode, char *shipname, ai_info *aip, object_ship_wing_point_team lua_target = object_ship_wing_point_team());
+extern void ai_add_wing_goal_player( int type, int mode, int submode, char *shipname, int wingnum, object_ship_wing_point_team lua_target = object_ship_wing_point_team());
```
Goober5000, line 115: I'm not really a fan of copying an `object_ship_wing_point_team` struct everywhere, since it will only be used if in fact there is a Lua AI target.  That said I don't want to rule it out either.  What are the tradeoffs between doing it this way and, say, storing an index into a global vector of OSWPTs?

code/hud/hudsquadmsg.h
```diff
@@ -56,6 +56,8 @@ class object;
 // out of order, but it was this way in the original source
 #define DISABLE_SUBSYSTEM_ITEM		(1<<15)
 
+#define MAX_ENGINE_PLAYER_ORDER		DISABLE_SUBSYSTEM_ITEM
```
Goober5000, line 59: Is this setting up a situation where it will no longer be possible to add any items beyond this one?  If so, then it would be better to count backwards from 31 while the built-in orders count upwards from 15.

freespace2/freespace.cpp
```diff
@@ -1917,6 +1917,9 @@ void game_init()
 
 	animation::ModelAnimationParseHelper::parseTables();
 
+	// Initialize dynamic SEXPs. Must happen before ship init for LuaAI
+	sexp::dynamic_sexp_init();
+
```
Goober5000, line 1922: This looks like a catch-22.  Dynamic sexps need to be initialized after the script system, and the script system needs to be initialized almost last.

Goober5000 (changes requested):
Partial review for now.

code/ai/aicode.cpp
BMagnu: Pl_objp is in fact the ship that will be lua-controlled and not its target, but you're very right that it's not clean to directly pass it here. It can easily be the aip pointer itself.
Changed.

code/ai/aigoals.cpp
BMagnu: It might make sense to allow the modder to define a "is AI valid for current situation" function. Though, I'd certainly make it optional, and defer to always assuming it's achievable otherwise. In that case, it probably should be put into a future PR?

code/ai/aigoals.h
BMagnu: Indeed, that's bad. Changed to const references

code/hud/hudsquadmsg.h
BMagnu: Yes it is such a situation. But we still need this kind of define to warn the modder as to when more orders cannot be added due to collisions with engine orders. What we can get rid of by counting backwards however, is the ``constexpr int max_highest_player_order_type = (1 << 31);``. 
Changed.

freespace2/freespace.cpp
BMagnu: Not true, actually. The dynamic SEXP system, on parse at least, only really registers the available SEXPs and does not interact with the script system at all. Only three things are important order-wise with dynamic SEXPs and Lua:

1. Lua must not _run_ the on-init hooks before dynamic SEPXs are initialized (this happens later in the function anyways, even before this PR)
2. No dynamic SEXP must be executed (defining them is fine) before the Script system is initialized, as it'll be missing the Lua-state to convert the values
3. No dynamic SEXP must be executed before the Script system has run its on-init hooks

So as far as I can tell, it should be fine to init the dynamic SEXPs before the scripting system, provided you don't try to use one of these SEXPs until scripting is initialized

code/ai/aicode.cpp
Goober5000: Ah right

code/ai/aigoals.cpp
Goober5000: I think the first "achievable" goal will always be executed.  So the question here is whether the modder will always want to run Lua code, as opposed to deferring it for some period to run standard behaviors like attacking.  But doing that in a future PR sounds fine.

freespace2/freespace.cpp
Goober5000: There is this line in `dynamic_sexp_init` which is presumably not valid until the script system is initialized:
```
	Script_system.OnStateDestroy.add(free_lua_sexps);
```

code/ai/ailua.h
```diff
@@ -22,6 +22,6 @@ bool ai_lua_add_order(int sexp_op, player_order_lua order);
 bool ai_lua_has_mode(int sexp_op);
 const ai_mode_lua* ai_lua_find_mode(int sexp_op);
 const player_order_lua* ai_lua_find_player_order(int sexp_op);
-void ai_lua(ship* shipp);
+void ai_lua(ai_info* shipp);
```
Goober5000, line 25: Might want to rename `shipp` to `aip`

code/scripting/api/objs/luaaisexp.cpp
```diff
@@ -0,0 +1,75 @@
...
+bool lua_ai_sexp_h::isValid() {
+	return sexp_handle != nullptr;
+}
+
+ADE_OBJ(l_LuaAISEXP, lua_ai_sexp_h, "LuaAISEXP", "Lua AI SEXP handle");
+
+ADE_VIRTVAR(ActionEnter,
+	l_LuaAISEXP,
+	"function(ai_helper helper, oswpt | nil arg) => boolean action",
+	"The action of this AI SEXP to be executed once when the AI recieves this order. Return true if the AI goal is complete.",
```
Goober5000, line 22: "receives"

freespace2/freespace.cpp
BMagnu: I _think_ that that's fine, since it's just adding a listener to when the LuaState object is deallocated, instead of actually requiring access to the LuaState in that moment. And the Script_system object itself already exists, since that has a static lifetime, and only just has no initialized LuaState yet.

Goober5000: Ok.

asarium: I agree with @BMagnu here that this move should be fine. Unfortunately, there is a circular dependency between the two systems since some Lua SEXPs have resources inside the Lua state. The `OnStateDestroy` callback is there to break that dependency so moving the init location should be fine.

However, please also replicate this change in QtFRED to ensure that stays in sync.

code/ai/ailua.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+	if (dynamicSEXP != nullptr && typeid(*dynamicSEXP) == typeid(sexp::LuaAISEXP)) {
+		auto lua_ai_sexp = static_cast<sexp::LuaAISEXP*>(dynamicSEXP);
+		const auto& action = lua_ai_sexp->getActionFrame();
+
+		run_ai_lua_action(action, lua_ai, aip);
+	}
+}
+
+void ai_lua_start(ai_goal* aip, object* objp){
```
Goober5000, line 90: The convention in the code is for `ai_goal` pointers to use `aigp` and `ai_info` pointers to use `aip`.

code/parse/sexp/LuaAISEXP.cpp
```diff
@@ -0,0 +1,150 @@
...
+using namespace luacpp;
+
+namespace sexp {
+static const SCP_unordered_set<int> allowed_oswpt_parameters{ OPF_SHIP, OPF_WING, OPF_SHIP_POINT, OPF_SHIP_WING, OPF_SHIP_WING_WHOLETEAM, OPF_SHIP_WING_SHIPONTEAM_POINT, OPF_SHIP_WING_POINT, OPF_SHIP_WING_POINT_OR_NONE };
+
+LuaAISEXP::LuaAISEXP(const SCP_string& name) : DynamicSEXP(name) {
+}
+
+void LuaAISEXP::initialize() {
};
```
Goober5000, line 17: There are a lot of `^M` characters in this file but not any of the other files.  Were there newline errors when this file was added?

code/parse/sexp/LuaAISEXP.cpp
```diff
@@ -0,0 +1,150 @@
...
+void LuaAISEXP::parseTable() {
+	SCP_stringstream help_text;
+	help_text << _name << "\r\n";
+
+	required_string("$Description:");
+
+	SCP_string description;
+	stuff_string(description, F_NAME);
+
+	help_text << "\t" << description << "\r\n";
```
Goober5000, line 45: If you want to allow multi-line descriptions, it will be necessary to add the `parseCheckEndOfDescription` logic from LuaSEXP.

code/parse/sexp/LuaAISEXP.cpp
```diff
@@ -0,0 +1,150 @@
...
+		help_text << "\t1";
+		help_text << ": " << param_desc << "\r\n";
+
+		required_string("+Type:");
+		SCP_string type_str;
+		stuff_string(type_str, F_NAME);
+
+		auto type = LuaSEXP::get_parameter_type(type_str);
+		if (type.second < 0 || allowed_oswpt_parameters.count(type.second) <= 0) {
+			error_display(0, "Parameter type '%s' is not an valid target type!", type_str.c_str());
```
Goober5000, line 70: should be "a valid"

Goober5000 (changes requested):
Finished going through all the files.

Goober5000 (approved):

PR #4200, opened 2022-04-22
Don't treat a bare <argument> as an error.

MageKing17:
PR #3750 made `check_sexp_syntax()` less permissive of node trees containing `<argument>`, which causes problems given that FRED saves `modify-variable`'s second argument as a `SEXP_ATOM_NUMBER`, as now `check_sexp_syntax()` complains that `<argument>` isn't a valid number. This commit both makes the `SEXP_ATOM_NUMBER` check skip the number validation logic if the `SNF_SPECIAL_ARG_IN_NODE` flag is set, as well as removing the `SEXP_ATOM_STRING` type check from the later `SNF_SPECIAL_ARG_IN_NODE` check (for making sure it's inside an `-argument` SEXP).

This caused problems with no fewer than three missions in BP2: the dreamscape, Her Finest Hour, and UT2. After compiling a build with these changes, I was able to successfully load the dreamscape and the event that was previously generating the error worked correctly.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -1923,32 +1923,35 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 			if (type == OPF_BOOL)  // allow numbers to be used where boolean is required.
 				type2 = OPR_BOOL;
 
-			while (*ptr) {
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Yup, this looks good. 👍

PR #4201, opened 2022-04-23
Delay processing "$Player Weapon Precedence:" until after parsing.

### LINKED ISSUE #4199: This -wep.tbm causes loadout problems mod wide
The symptom is that if you switch ships in Ship Loadout to a ship whose default secondary weapons are not available in the mission's weapon pool then the ship will be set to have no weapons in the secondary bank at all. If players don't notice this, then they can start the mission and have no secondaries. Expected behavior is the secondary banks will be filled with the first weapon that is allowed for the ship.

The attached tbm causes the issue and the included mission file exposes it. Both work on retail. Load the mission and note that your Ulysses is stocked with Rockeyes, which the Hercules can carry. Switch your ship to the Hercules and return to the Weapon Loadout screen to see that your secondary banks are not filled with Rockeyes, but are instead empty.

I suspect this is partially an engine bug and partially bad mod data with the tbm missing some field that FSO expects to have but doesn't report not having.
[loadout_issue.zip](https://github.com/scp-fs2open/fs2open.github.com/files/8543972/loadout_issue.zip)

MageKing17:
The `stuff_int_list()` call converts the weapon names into the appropriate index into the `Weapon_info` array; however, in `weapon_do_post_parse()`, we call `weapon_sort_by_type()`, which reorders the array; this makes these indices potentially point to the wrong weapon. Additionally, there's no protection against someone `+remove`ing one of the weapons in the list in a later TBM.

This commit makes it so we instead store the names into a vector of strings, then convert them into indices after `weapon_sort_by_type()` has already been called. Additionally, `Player_weapon_precedence` is changed from a fixed-size array to a vector, since it's usually not as long as `MAX_WEAPON_TYPES`.

Fixes #4199.

### DIFF
--- a/code/missionui/missionweaponchoice.cpp
+++ b/code/missionui/missionweaponchoice.cpp
@@ -1761,7 +1761,7 @@ void wl_remove_weps_from_pool(int *wep, int *wep_count, int ship_class)
 	
 					if ( (Wl_pool[wi_index] <= 0) || (wep_count[i] == 0) ) {
 						// fresh out of this weapon, pick an alternate pool weapon if we can
-						int wep_pool_index, wep_precedence_index, new_wi_index = -1;
+						int wep_pool_index, new_wi_index = -1;
 						for ( wep_pool_index = 0; wep_pool_index < weapon_info_size(); wep_pool_index++ ) {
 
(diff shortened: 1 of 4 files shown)


wookieejedi (approved):
Good catch with a solid fix.

PR #4203, opened 2022-04-23
Add a ship flag to allow the firing fail sound to play with locked primaries

Kiloku:
This Ship flag (`Fail_sound_locked_primaries`) causes the game to call the firing fail sound (like when you're out of weapon energy) if the player tries to fire their primaries when they're under the effects of the `Primaries_locked` flag. 

Tested and worked fine.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -12454,7 +12454,7 @@ int ship_fire_secondary( object *obj, int allow_swarm, bool rollback_shot )
 	}
 	swp->next_secondary_fire_stamp[bank] = timestamp((int)(t * 1000.0f));
 	swp->last_secondary_fire_stamp[bank] = timestamp();
-	
+
 	// Here is where we check if weapons subsystem is capable of firing the weapon.
 	// do only in single player or if I am the server of a multiplayer game
 	if ( !(Game_mode & GM_MULTIPLAYER) || MULTIPLAYER_MASTER ) {
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -525,7 +525,8 @@ ship_flag_name Ship_flag_names[] = {
 	{ Ship_Flags::No_disabled_self_destruct,	"no-disabled-self-destruct" },
 	{ Ship_Flags::Hide_mission_log,				"hide-in-mission-log" },
 	{ Ship_Flags::No_passive_lightning,			"no-ship-passive-lightning" },
-	{ Ship_Flags::Fail_sound_locked_primary, 	"fail-sound-locked-primary"}
+	{ Ship_Flags::Fail_sound_locked_primary, 	"fail-sound-locked-primary"},
+	{ Ship_Flags::Fail_sound_locked_secondary, 	"fail-sound-locked-secondary"}
 };
 
 static int Laser_energy_out_snd_timer;	// timer so we play out of laser sound effect periodically
@@ -12267,7 +12268,7 @@ int ship_fire_secondary( object *obj, int allow_swarm, bool rollback_shot )
 	}
 
 	// If the secondaries have been locked, bail
-	if (shipp->flags[Ship_Flags::Secondaries_locked])
+	if (shipp->flags[Ship_Flags::Secondaries_locked] && !(obj == Player_obj && shipp->flags[Ship_Flags::Fail_sound_locked_secondary]))
 	{
 		return 0;
 	}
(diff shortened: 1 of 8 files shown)


code/ship/ship.cpp
```diff
@@ -11676,6 +11678,14 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
...
 				}
 
+				// The player is trying to fire primaries which are locked. We've set the Fail_sound_locked_primaries flag, so it should make the fail sound.
+				if  (obj == Player_obj && shipp->flags[Ship_Flags::Primaries_locked] && shipp->flags[Ship_Flags::Fail_sound_locked_primary])
+				{					
+					ship_maybe_do_primary_fail_sound_hud(false);
+					ship_stop_fire_primary_bank(obj, bank_to_fire);
+					continue;
+				}
+
```
MageKing17, line 11688: This seems a bit far down the function to do this; I know this is adjacent to the regular `ship_maybe_do_primary_fail_sound_hud()` call, but that has reasons to be in the nitty-gritty of this function, while this kinda doesn't; it should have the same effect if you called `ship_maybe_do_primary_fail_sound_hud(false)` around line 11302 immediately before returning.
```suggestion
```

code/ship/ship.cpp
```diff
@@ -11298,8 +11299,9 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
 		return 0;
 	}	
 
-	// If the primaries have been locked, bail
-	if (shipp->flags[Ship_Flags::Primaries_locked])
+	// If the primaries have been locked, bail. 
+	// Unless we're dealing with the player and their ship has the flag set to allow fail sounds when firing locked primaries.
+	if (shipp->flags[Ship_Flags::Primaries_locked] && ! (obj == Player_obj && shipp->flags[Ship_Flags::Fail_sound_locked_primary]))
 	{
```
MageKing17, line 11305: For example:
```suggestion
	// If the primaries have been locked, bail.
	if (shipp->flags[Ship_Flags::Primaries_locked]) {
		// If we're dealing with the player and their ship has the flag set to allow fail sounds when firing locked primaries, do that first.
		if (obj == Player_obj && shipp->flags[Ship_Flags::Fail_sound_locked_primary]) {
			ship_maybe_do_primary_primary_fail_sound_hud(false);
		}
```

MageKing17 (changes requested):
One nitpick, but other than that, it looks pretty straightforward.

wookieejedi (changes requested):
If you would like this to be toggle-able via the `alter-ship-flag` sexp you'll also need to add to FRED and QTFRED (#4194 is an example of relevant sections). 

Alternatively, might be useful for adding this feature but for secondary sounds, too.

code/ship/ship.cpp
Kiloku: That's what I tried at first. When I added the code in that spot, the sound played every 50ms regardless of the weapon's fire-rate.  
Adding it afterwards ensures the fire rate has already been handled.

wookieejedi (approved):

MageKing17:
> If you would like this to be toggle-able via the `alter-ship-flag` sexp you'll also need to add to FRED and QTFRED (#4194 is an example of relevant sections).

It already had the needed changes to allow `alter-ship-flag` to use it; commit 01a741074de0ad6ceeb561fd5f9cffe1aa5a1bcb made it so it could be saved to and loaded from the mission file, but has nothing to do with `alter-ship-flag`, and without changing the ship editor for FRED and qtFRED, is somewhat dead code. PR #4194 fixed alter-ship-flag support by actually incrementing `MAX_OBJECT_FLAG_NAMES`, while this PR already (correctly) incremented `MAX_SHIP_FLAG_NAMES`.

code/ship/ship.cpp
MageKing17: In that case, there's still an earlier location you can put it: around line 11527 (on master, anyway) there's a block that starts like this:
```cpp
		// Here is where we check if weapons subsystem is capable of firing the weapon.
		// Note that we can have partial bank firing, if the weapons subsystem is partially
		// functional, which should be cool.  		
		if ( ship_weapon_maybe_fail(shipp) && !force) {
```
This does something similar (play the gun-jammed sound effect and stop the firing action if the weapons subsystem is damaged), and avoids wasting its time calculating auto-aim and whatnot. Additionally, if you leave the code where it is, it only applies to non-beam primaries, while the block mentioned above happens before that conditional split.

code/ship/ship.cpp
```diff
@@ -12454,6 +12466,10 @@ int ship_fire_secondary( object *obj, int allow_swarm, bool rollback_shot )
 				}
 			goto done_secondary;
 		}
+		if (shipp->flags[Ship_Flags::Secondaries_locked] && (obj == Player_obj && shipp->flags[Ship_Flags::Fail_sound_locked_secondary])) {
+			ship_maybe_do_secondary_fail_sound_hud(wip, false);
+			goto done_secondary;
+		}
```
MageKing17, line 12472: It would make more sense to do this above the above "due to weapons system damage" failure, because locked secondaries should take priority.

MageKing17 (changes requested):
One small remaining issue, but it should be a very quick fix.

MageKing17 (approved):

PR #4205, opened 2022-04-23
Move of special parsing functions out of lighting_profiles into new file

EatThePath:
This takes a little bit of tedium reduction I used in lighting profiles parsing and makes it more generally available.
I have done some work on applying this concept to all common table parsing constructs, but I would like to make use of this approach in other things long before I would ever get done doing that, so I'm PRing this as is with the intent of adding more capabilities as I actually need them.

### DIFF
--- a/code/parse/parsehi.h
+++ b/code/parse/parsehi.h
@@ -1,0 +1,7 @@
+#pragma once
+#include "parse/parselo.h"
+
+//parsehi is intended for higher level, frequently used combinations of parselo functions
+//and also for collection of standardized parsing of complex types where appropriate
+
+extern bool parse_optional_float_into(const SCP_string& field_name, float* value_target);
--- a/code/lighting/lighting_profiles.cpp
+++ b/code/lighting/lighting_profiles.cpp
@@ -4,6 +4,7 @@
 #include "def_files/def_files.h"
 #include "lighting/lighting.h"
 #include "lighting/lighting_profiles.h"
+#include "parse/parsehi.h"
 #include "parse/parselo.h"
 
 //TODO: maybe a parsehi.cpp would be in order here?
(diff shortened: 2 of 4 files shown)


code/parse/parsehi.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+#include "parse/parselo.h"
+
+//parsehi is intended for higher level, frequently used combinations of parselo functions
+//and also for collection of standardized parsing of complex types where appropriate
+
+extern bool parse_optional_float_into(const SCP_string& field_name, float* value_target);
```
Goober5000, line 7: This should use a C-string rather than an SCP_string.  It will allow the function to be used while parsing missions or tables if desired.

Goober5000 (changes requested):

code/parse/parsehi.h
EatThePath: Can you provide an example of the sort of usage you're talking about? The current code works perfectly fine when passing string literals in to that argument, which is the only way I see `optional_string` being used in the codebase currently.

Goober5000: I was a little too hasty; I was thinking of `Mp` which doesn't apply here.  However if you change the first argument to `const char *` you remove the unnecessary creation of multiple SCP_strings.

EatThePath: Is that really that high a cost? I'd seen enough functions using this as their means of taking string parameters that I'd gotten the impression it was the way to go for new code, especially since it can take either a c-string or a scp_string without complaint while giving a c-string argument an un-decorated scp_string can't. With this having the aim of higher level functions to streamline code that seemed like the way to go.

MageKing17: Unless the overhead of `std::string` is drastically higher than I thought it was (in which case we should be refactoring all of our parsing code to never use `SCP_string` ever), there's absolutely no reason to change this function.

Goober5000: It's not a really high cost, it's just a principle of good code design.  Since the underlying function uses a C-string and your function essentially passes the parameter through, your function should also use a C-string.  The way it is now, you're round-tripping a C-string literal to a SCP_string and back to a C-string again.

You could argue that the overhead is too small to make a difference, and it probably is, but many cases across the code base add up.  In the extreme case you end up with the situation where Google Chrome made 25,000 string allocations per keystroke: https://groups.google.com/a/chromium.org/g/chromium-dev/c/EUqoIz2iFU4/m/kPZ5ZK0K3gEJ

We do have a lot of functions that use SCP_string, but most of the cases fall into one of two categories.  Either a) SCP_string is used consistently across the API, removing the need for round-tripping; or b) there are parallel implementations, one for C-string and one for SCP_string.

MageKing17: All those parallel implementations are in cases where *the C-string implementation was first* and we kept it for backwards-compatibility reasons. This is a new function, so there's no backwards compatibility to keep. There's just as much "good code design" in never using C-style strings and always using `std::string`, so that argument doesn't hold water either. Either there are unacceptable performance costs, or you're insisting that other coders engage in premature optimization for the sake of your personal preferences, which just sounds like a good way to discourage people from contributing to the SCP at all. Seriously, this function is fine just the way it is.

Goober5000: This function is a wrapper around a pre-existing function (`optional_string`) that *does* use a C-string.  That's the compatibility layer you're looking for.  Aside from that, you're making quite a few logical leaps jumping from premature optimization to personal preferences to discouraging people.

Anyway this is @EatThePath 's PR so I'll wait to see what he says.

EatThePath: While the allocation is of course a non-zero cost and the inner functions use the cstring anyway, I don't find the performance boogieman that compelling as is. 

For some added context: this function (and other future ones like it) is in part intended to be helpful with Parse Items, and was originally conceived of in early versions of in very code, where all the field names *are* stored in SCP strings rather than literals. That's part of my preference, when that takes over parsing I'd prefer to keep things 'wrapped' as long as possible.

It's really a small thing, and I'll change it if that's required to pass review, but for those reasons I'd rather leave as is unless it becomes a demonstrably measurable performance issue when widely deployed.

Goober5000: Ok.  I still prefer it the other way, but I'll withdraw this request and approve the PR.

Goober5000 (approved):

PR #4206, opened 2022-04-23
add some script functions for HUD gauge data

Goober5000:
These are useful for adapting scripts based on whether HUD gauges are scaled and whether they use different coordinates or fonts.


wookieejedi (approved):
These will be quite useful. Everything looks good to me.

PR #4207, opened 2022-04-23
Fix Texture Replace and Team Colors with Ship Changes

wookieejedi:
Tabled `$Texture Replace` did not work on ships where the class is changed in the loadout or with the `change-ship-class` sexp. This was because the replacement textures only were loaded during mission parse, and the `change_ship_type` did not account for loading new textures from ship classes that were not in the initial mission parse. Part of the bug was also that the code that updates texture replacements was after the texture page-in section of `change_ship_type`.

This PR fixes that bug by properly loading the textures in when the ship class changes. It also moves the code to within the `ship_model_change` function (which is only used in the `change_ship_type` function) so that the texture loading happens correctly before texture page in. The updated code also fixes some other edge case bugs to ensure that texture replace happens even if the old ship type did not use texture replacement. 

PR also fixes bug where Team Color was not maintained between loadout changes with ships that all had team colors enabled.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -10094,18 +10094,18 @@ int ship_create(matrix* orient, vec3d* pos, int ship_type, const char* ship_name
  */
 static void ship_model_change(int n, int ship_type)
 {
-	int i;
-	ship_info* sip;
-	ship* sp;
-	polymodel* pm;
-	object* objp;
+	int			i;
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Looks good.

MageKing17 (approved):
Looks good to me.

PR #4208, opened 2022-04-24
Weapon lights

EatThePath:
Configurable lighting data for weapons, both on a per-weapon basis and globally via lighting profiles. 

I will be doing a bit more testing to makes sure last minute changes or history rewriting didn't break anything, but this is now ready to review.

### DIFF
--- a/code/graphics/color.cpp
+++ b/code/graphics/color.cpp
@@ -1,0 +1,227 @@
+#include "graphics/color.h"
+#include "graphics/2d.h"
+#include "globalincs/pstypes.h"
+#include "globalincs/vmallocator.h"
+#include "math/floating.h"
+
+void hdr_color::set_vecf(const SCP_vector<float>* input)
+{
+	size_t l = input->size();
+	Assertion(l >= 3, "attempted to set color values invalid component count");
+	this->red = input->at(0);
+	this->green = input->at(1);
+	this->blue = input->at(2);
+	if (l > 3)
+		this->alpha = input->at(3);
+	if (l > 4)
+		this->intensity = input->at(4);
+}
+
+/*
+ * @brief Fills a vector with the raw float color components of the color, in the order red, green, blue, alpha,
+ * intensity
+ *
+ * @param to_fill: The fector that will be filled with the color compoents
+ */
+void hdr_color::get_v5f(SCP_vector<float>* to_fill) const
+{
+	if (to_fill == nullptr)
+		to_fill = new SCP_vector<float>;
+	to_fill->clear();
+	to_fill->push_back(this->red);
+	to_fill->push_back(this->green);
+	to_fill->push_back(this->blue);
+	to_fill->push_back(this->alpha);
+	to_fill->push_back(this->intensity);
+}
+
+hdr_color::hdr_color()
+{
+	this->red = 1.0f;
+	this->green = 1.0f;
+	this->blue = 1.0f;
+	this->alpha = 1.0f;
+	this->intensity = 1.0f;
+}
+
+hdr_color::hdr_color(float new_r, float new_g, float new_b, float new_a, float new_i)
+{
+	this->red = new_r;
+	this->green = new_g;
+	this->blue = new_b;
+	this->alpha = new_a;
+	this->intensity = new_i;
+}
+
+hdr_color::hdr_color(const hdr_color* const source_color)
+{
+	this->red = source_color->red;
+	this->green = source_color->green;
+	this->blue = source_color->blue;
+	this->alpha = source_color->alpha;
+	this->intensity = source_color->intensity;
+}
+
+/**
+ * @brief Sets RGB values from three 0-255 ints
+ */
+void hdr_color::set_rgb(int new_r, int new_g, int new_b)
+{
+	this->red = i2fl(new_r) / 255.0f;
+	this->green = i2fl(new_g) / 255.0f;
+	this->blue = i2fl(new_b) / 255.0f;
+}
+
+/**
+ * @brief Sets RGBA values from an old style color object
+ */
+void hdr_color::set_rgb(const color* const new_color)
+{
+	this->set_rgb(new_color->red, new_color->green, new_color->blue);
+	this->alpha = i2fl(new_color->alpha);
+}
+
+/**
+ * @brief Sets RGB values from an array of three 0-255 ints
+ */
+void hdr_color::set_rgb(const int* const new_rgb)
+{
+	this->set_rgb(new_rgb[0], new_rgb[1], new_rgb[2]);
+}
+
+
+/**
+ * @brief retreives unmultiplied 0.0f-1.0f color component.
+ */
+float hdr_color::r() const
+{
+	return red;
+}
+
+/**
+ * @brief sets and returns unmultiplied 0.0f-1.0f color component.
+ */
+float hdr_color::r(float in)
+{
+	red = in;
+	return red;
+}
+
+/**
+ * @brief sets and returns unmultiplied 0.0f-1.0f color component from 0-255 int.
+ */
+float hdr_color::r(int in)
+{
+	red = i2fl(in) / 255.0f;
+	return red;
+}
+
+
+/**
+ * @brief retreives unmultiplied 0.0f-1.0f color component.
+ */
+float hdr_color::g() const
+{
+	return green;
+}
+
+/**
+ * @brief sets and returns unmultiplied 0.0f-1.0f color component.
+ */
+float hdr_color::g(float in)
+{
+	green = in;
+	return green;
+}
+
+/**
+ * @brief sets and returns unmultiplied 0.0f-1.0f color component from 0-255 int.
+ */
+float hdr_color::g(int in)
+{
+	green = i2fl(in) / 255.0f;
+	return green;
+}
+
+
+/**
+ * @brief retreives unmultiplied 0.0f-1.0f color component.
+ */
+float hdr_color::b() const
+{
+	return blue;
+}
+
+/**
+ * @brief sets and returns unmultiplied 0.0f-1.0f color component.
+ */
+float hdr_color::b(float in)
+{
+	blue = in;
+	return blue;
+}
+
+/**
+ * @brief sets and returns unmultiplied 0.0f-1.0f color component from 0-255 int.
+ */
+float hdr_color::b(int in)
+{
+	blue = i2fl(in) / 255.0f;
+	return blue;
+}
+
+
+/**
+ * @brief retreives unmultiplied 0.0f-1.0f color component.
+ */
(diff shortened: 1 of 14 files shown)


z64555:
For every 'new' there must be a 'delete' at the end of the lifetime of the created object.  'delete' is usually written in the destructor of the class that called 'new'

Member data that are standard data types char, int, float, etc. and arrays do not need to be explicity cleaned up in the destructor of a class, they will be dealt with automatically.

Member data that are pointers to objects do not need to be managed either, however their referenced data must be cleaned up one way or another.  Objects that are defined in the codebase "normally" (that is, *not* using 'new' or 'malloc') will be cleaned up the moment they go out of execution scope. (Lookup scope for C++)

<hr>
Make use of the keywords 'public', 'protected', and 'private' to control access to methods and member data. Public is accessible to all. Protected is accessible to the class and its children, private is accessible only to itself.  Class friendship is also a thing but is usually not useful.

When prototyping a new class its easy to  just slap down 'public' and get to coding, but this is a bad habit.  You should think at the start of how its members should be interacted with the rest of the codebase.
  Public member data can be both accessed and modified by all sources, which can cause action at a distance bugs due to abuse of public members, but it does ease the workload of needing to supply getter and setter methods to access member data.

Protected members and methods can be accessed and modified by the class and its children/subclasses that inherit it.  This should be the highest level of helper subroutines.

Refrain from defining helper subroutines C-style within the .cpp implementation of the class, as this obfuscates them from maintainers and means they're effectively private static methods.

Private members and methods are data that can *only* be used by that class. If member data are protected or private they must have getter and setter methods for outside code to see or modify them.  Pointers to data on the heap (from 'new' or 'malloc') that are created within the class should be 'protected' or 'private' to gate potential abuse.

wookieejedi:
Great to see this one is ready for review, really looking forward to being able to use these updates in FotG (and I'm sure other mods are too). Given the smaller number of graphics coders, I'm guessing @TRBlount would be most qualified to start this off?

EatThePath:
> Great to see this one is ready for review, really looking forward to being able to use these updates in FotG (and I'm sure other mods are too). Given the smaller number of graphics coders, I'm guessing @TRBlount would be most qualified to start this off?

Possibly yes. Fiddling with weapons and beams brings @Baezon to mind, and the parsing sections makes me think @JohnAFernandez too. Ultimately I don't thinking too much of this is sticky code, but it touches enough things to be a bit of a task to gather up all the context I guess.

code/parse/parsehi.cpp
```diff
@@ -17,3 +19,42 @@ bool parse_optional_float_into(const SCP_string& field_name, float* value_target
...
+ * @return True if a value was parsed, false if not
+ */
+bool parse_optional_color3i_into(const SCP_string& field_name, hdr_color* out_color)
+{
+	if (optional_string(field_name.c_str())) {
+		int components[3] = {255, 255, 255};
+		stuff_int_list(components, 3, RAW_INTEGER_TYPE);
+		if (out_color == nullptr) {
+			out_color = new hdr_color();
+		}
```
Goober5000, line 55: This isn't going to do anything for `out_color` because the calling function won't have the address of the allocated memory.  To make this work you would need a pointer-to-pointer.  However, given the risky memory management here, I would recommend taking this out and just adding an Assertion that the pointer isn't null.

code/weapon/weapon.h
```diff
@@ -251,6 +252,8 @@ typedef struct beam_weapon_info {
 	float range;						// how far it will shoot-Bobboau
 	float damage_threshold;				// point at wich damage will start being atenuated from 0.0 to 1.0
 	float beam_width;					// width of the beam (for certain collision checks)
+	bool beam_light_flicker;			// If beam light is affected by the flickering
+	bool beam_light_as_multiplier;		// If beam light is used directly or multiplies the section width
```
Goober5000, line 256: Why not add these to the flag set?

code/weapon/weapon.h
```diff
@@ -354,6 +357,10 @@ struct weapon_info
 	float	laser_head_radius, laser_tail_radius;
 	float	collision_radius_override;          // overrides the radius for the purposes of collision
 
+	bool 		light_color_set;
```
Goober5000, line 360: Similarly, this could be added to the weapon_info flag set

code/object/object.cpp
```diff
@@ -1210,25 +1214,33 @@ void obj_move_all_post(object *objp, float frametime)
 						cast_light = 0;
 					}
 				}
-
-				if ( cast_light )	{
-					weapon_info * wi = &Weapon_info[Weapons[objp->instance].weapon_info_index];
-
-					if ( wi->render_type == WRT_LASER )	{
+				weapon_info* wi = &Weapon_info[Weapons[objp->instance].weapon_info_index];
```
Goober5000, line 1217: Why did you move this outside the if() block?

code/graphics/color.cpp
```diff
@@ -0,0 +1,227 @@
+#include "graphics/color.h"
+#include "graphics/2d.h"
+#include "globalincs/pstypes.h"
+#include "globalincs/vmallocator.h"
+#include "math/floating.h"
+
+void hdr_color::set_vecf(const SCP_vector<float>* input)
```
Goober5000, line 7: You could make this a const reference (`const SCP_vector<float>&`) if you want

code/graphics/color.cpp
```diff
@@ -0,0 +1,227 @@
...
+		this->intensity = input->at(4);
+}
+
+/*
+ * @brief Fills a vector with the raw float color components of the color, in the order red, green, blue, alpha,
+ * intensity
+ *
+ * @param to_fill: The fector that will be filled with the color compoents
+ */
+void hdr_color::get_v5f(SCP_vector<float>* to_fill) const
```
Goober5000, line 26: You could make this a non-const reference.  In fact, this has the same pointer bug and memory risk as the function in parsehi.

code/lighting/lighting_profiles.cpp
```diff
@@ -8,6 +8,134 @@
...
+		if (parses == 0) {
+			Warning(LOCATION,
+				"Lighting profile value '%s' in file '%s' profile '%s' parsed but set no properties, possible "
+				"malformed table.",
+				valuename,
+				filename,
+				profile_name.c_str());
+		} else if (parses > 5) {
+			Warning(LOCATION,
+				"Lighting profile value '%s' in file '%s' profile '%s' parsed too man no properties, possible "
```
Goober5000, line 121: "too man no"?

code/lighting/lighting_profiles.cpp
```diff
@@ -96,6 +257,38 @@ void lighting_profile::parse_default_section()
 		parsed |= parse_optional_float_into("$PPC Shoulder Length:",&default_profile.ppc_values.shoulder_length);
 		parsed |= parse_optional_float_into("$PPC Shoulder Strength:",&default_profile.ppc_values.shoulder_strength);
 		parsed |= parse_optional_float_into("$PPC Shoulder Angle:",&default_profile.ppc_values.shoulder_angle);
+		parsed |= parse_optional_float_into("$Exposure:",&default_profile.exposure);
+
+		parsed |= lighting_profile_value::parse(filename,"$Missile light brightness:","Default profile",
```
Goober5000, line 262: I would suggest setting a `const char *` variable to "Default profile" and then using that for each line.

Goober5000 (changes requested):
Reviewed a lot of it but skipped the graphics/lighting code.  Noticed a few things.

code/weapon/weapon.h
EatThePath: mainly I forgot flagsets exist. Will correct.

code/object/object.cpp
EatThePath: I believe I was at one intermediate stage using something from wi in the if condition, and forgot to moved it back when that became no-longer true. I'll put it back

code/lighting/lighting_profiles.cpp
EatThePath: I wonder how tired I was when writing that one, woof.

code/weapon/weapon.h
EatThePath: Upon review and discussion, adding these(and the other bool later) to the existing regular flag sets would have undesirable side effects. To my mind creating new flagsets for them would be added complexity for little to no benefit, so I'm inclined to leave them as is unless there's a compelling benefit I'm unaware of.
The other review items have been addressed, though.

Goober5000:
Ok, LGTM now for the parts I looked at.  Still need someone else to review the lighting code.

code/weapon/weapon.h
Baezon: What undesirable side effects would there be to adding it to beam info flags?

EatThePath: It was my understanding that if anyone adds a beam flag in a tbm it will clear any beam flags set upstream, which isn't something I much care for really. Even if I was misinformed or misunderstood and that doesn't apply here, I'd still be curious to know what benefit that system provides over this method.

code/graphics/color.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
```
TRBlount, line 1: Generally, we seem to use `#ifndef _FILE_H_` instead of `#pragma once`, would need feedback from others on how they feel about this.

code/lighting/lighting.cpp
```diff
@@ -148,6 +150,12 @@ static void light_rotate(light * l)
 	}
 }
 
+void light_add_directional(const vec3d* p0, float r1, float r2, const hdr_color* new_color)
+{
+	Assert(new_color!= nullptr);
+	light_add_directional(p0, r1, r2, new_color->i(), new_color->r(), new_color->g(), new_color->b());
```
TRBlount, line 156: Several function calls don't match a visible prototype/definition here.

code/weapon/weapon.h
TRBlount: Is this resolved or awaiting followup?

TRBlount (commented):
Waiting for reviewer feedback and some fixes.

code/weapon/weapon.h
EatThePath: @TRBlount I took it as resolved, personally, but I could be wrong

Goober5000: It's resolved.  The only existing flagset is the weapon flags, which is not appropriate for this.

code/lighting/lighting.cpp
EatThePath: Copy-paste oversights. I think I got them all in my last push

TRBlount (approved):
Looks good now.

Goober5000 (approved):

PR #4209, opened 2022-04-24
Limit all MP client collisions to prevent bugs

JohnAFernandez:
Clients with bad pings can find themselves being thrown to the far corners of the map because multiple collisions can happen before an update can be received from the server, compounding the effects of collisions.  (And on the second collision, an interpolated ship can actually do the opposite of react properly to a collision impulse, especially if a brand new object update packet has just come in, making the subsequent collision even more powerful)

Client players are still affected by collisions on their local instance, but other ships will literally pass through each other until updated by the server (avoids other bugs with stationary ships).  A new timestamp controls the frequency of collisions with the player ship, so that the buggy secondary collisions do not happen.

Still needs to be tested, but ready to be reviewed.

### DIFF
--- a/code/object/collideshipship.cpp
+++ b/code/object/collideshipship.cpp
@@ -805,6 +805,15 @@ void calculate_ship_ship_collision_physics(collision_info_struct *ship_ship_hit_
 		Assert( !vm_is_vec_nan(&direction_light) );
 		vm_vec_scale_add2(&heavy->pos, &direction_light,  0.2f * lighter->phys_info.mass / (heavy->phys_info.mass + lighter->phys_info.mass));
 		vm_vec_scale_add2(&heavy->pos, &ship_ship_hit_info->collision_normal, -0.1f * lighter->phys_info.mass / (heavy->phys_info.mass + lighter->phys_info.mass));
+
+		// while we are in a block that has already checked if we should collide, set the MP client timestamps
+		if (MULTIPLAYER_CLIENT){
+			if (lighter == Player_obj && heavy->type == OBJ_SHIP){
+				Ships[heavy->instance].multi_client_collision_timestamp = _timestamp( calculate_next_multiplayer_client_collision_time(impulse_mag) );
+			} else if (lighter->type == OBJ_SHIP) {
+				Ships[lighter->instance].multi_client_collision_timestamp = _timestamp( calculate_next_multiplayer_client_collision_time(impulse_mag) );
+			}
+		}
 	}
 	
 	//For landings, we want minimal movement on the light ship (just enough to keep the collision detection honest)
@@ -822,14 +831,6 @@ void calculate_ship_ship_collision_physics(collision_info_struct *ship_ship_hit_
 			lighter->phys_info.mass = copy_mass;
 		} else {
 			heavy->phys_info.mass = copy_mass;
-		}
-	}
-
-	if (MULTIPLAYER_CLIENT && should_collide){
-		if (lighter == Player_obj && heavy->type == OBJ_SHIP){
-			Ships[heavy->instance].multi_client_collision_timestamp = _timestamp( calculate_next_multiplayer_client_collision_time(impulse_mag) );
-		} else if (lighter->type == OBJ_SHIP) {
-			Ships[lighter->instance].multi_client_collision_timestamp = _timestamp( calculate_next_multiplayer_client_collision_time(impulse_mag) );
 		}
 	}
 }
--- a/code/object/collideshipship.cpp
+++ b/code/object/collideshipship.cpp
@@ -473,6 +473,33 @@ static bool check_subsystem_landing_allowed(ship_info *heavy_sip, collision_info
 	return false;
 }
 
+/**
+ * Helper function that calculates the next time this ship should be allowed to collide with a client player on multiplayer.
+ */
+int calculate_next_multiplayer_client_collision_time(float impulse_magnitude)
+{
+	// Uses a continuous cubic rational function that allows us to space out collisions based on collision strength.
+	static constexpr float IMPULSE_RANGE_FACTOR = 1.0f / powf(9900, 3.0f); // based on some quick tests from Asteroth about how strong an im
+	static constexpr float MIN_IMPULSE = 100;
+	static constexpr int LONGEST_COLLISION_INTERVAL = 200;
+
+	impulse_magnitude -= MIN_IMPULSE;
+
+	// if barely anything is happening, then do not limit collisions
+	if (impulse_magnitude < 0.0f) { 
+		return 0;
+	}
+
+	// caculate the impulse we have basically as a percentage. (CLAMP macro works best as a separate line)
+	float factor = powf(impulse_magnitude, 3.0f) * IMPULSE_RANGE_FACTOR;
+	CLAMP(factor, 0.0f, 1.0f);
+
+	// then multiply times the longest interval we could want.
+	factor *= LONGEST_COLLISION_INTERVAL;
+
+	return static_cast<int>(factor);
+}
+
 // ------------------------------------------------------------------------------------------------
 //		input:		ship_ship_hit		=>		structure containing ship_ship hit info
 //		(includes)	A, B					=>		objects colliding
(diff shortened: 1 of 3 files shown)


code/object/collideshipship.cpp
```diff
@@ -473,6 +473,33 @@ static bool check_subsystem_landing_allowed(ship_info *heavy_sip, collision_info
...
 
+/**
+ * Helper function that calculates the next time this ship should be allowed to collide with a client player on multiplayer.
+ */
+int calculate_next_multiplayer_client_collision_time(float impulse_magnitude)
+{
+	// Uses a continuous cubic rational function that allows us to space out collisions based on collision strength.
+	static constexpr float IMPULSE_RANGE_FACTOR = 1.0f / powf(9900, 3.0f); // based on some quick tests from Asteroth about how strong an im
+	static constexpr float MIN_IMPULSE = 100;
+	static constexpr int LONGEST_COLLISION_INTERVAL = 200;
```
notimaginative, line 484: These shouldn't be declared `static`. That allocates them on the global stack, where they have no business being.

notimaginative (changes requested):

code/object/collideshipship.cpp
```diff
@@ -473,6 +473,33 @@ static bool check_subsystem_landing_allowed(ship_info *heavy_sip, collision_info
 	return false;
 }
 
+/**
+ * Helper function that calculates the next time this ship should be allowed to collide with a client player on multiplayer.
+ */
+int calculate_next_multiplayer_client_collision_time(float impulse_magnitude)
```
notimaginative, line 479: The function should be declared static, since it's not used outside of this file.

notimaginative (changes requested):

code/object/collideshipship.cpp
JohnAFernandez: Should be fixed.

code/object/collideshipship.cpp
JohnAFernandez: Should be fixed.

notimaginative (approved):
LGTM. I'll obviously defer to you and Asteroth on the correctness of the rest of it though.

JohnAFernandez:
Tests for this went well.  Collisions felt normal (for freespace) after the changes.

code/object/collideshipship.cpp
```diff
@@ -473,6 +473,36 @@ static bool check_subsystem_landing_allowed(ship_info *heavy_sip, collision_info
...
 
+/**
+ * Helper function that calculates the next time this ship should be allowed to collide with a client player on multiplayer.
+ */
+static int calculate_next_multiplayer_client_collision_time(float impulse_magnitude)
+{
+	// Uses a continuous cubic rational function that allows us to space out collisions based on collision strength.
+	// based on some quick tests from Asteroth about how strong impuse is on average.
+	constexpr float IMPULSE_RANGE_FACTOR = 0.000000000001031f; 	// The inverse of the range of regular impulse values, 9900, to the third power
+	constexpr int MIN_IMPULSE = 100;							// The lowest values of collisions that we see.  Ignore if lower than this.
```
TRBlount, line 484: If this is going to be used only with other floats, make this a `float` to save a type conversion.

code/object/collideshipship.cpp
```diff
@@ -473,6 +473,36 @@ static bool check_subsystem_landing_allowed(ship_info *heavy_sip, collision_info
...
+	constexpr int LONGEST_COLLISION_INTERVAL = 175;				// 1/5th of a second is usually enough time to get an update from the server.
```
TRBlount, line 485: ditto

code/object/collideshipship.cpp
```diff
@@ -473,6 +473,36 @@ static bool check_subsystem_landing_allowed(ship_info *heavy_sip, collision_info
...
+	constexpr int MINIMUM_INTERVAL = 25;						// if there is sufficient impulse, don't go below this value.
```
TRBlount, line 486: And again.

code/object/collideshipship.cpp
```diff
@@ -751,17 +792,23 @@ void calculate_ship_ship_collision_physics(collision_info_struct *ship_ship_hit_
 	// We will try not to worry about the left over time in the frame
 	// heavy's position unchanged by collision
 	// light's position is heavy's position plus relative position from heavy
-	vm_vec_add(&lighter->pos, &heavy->pos, &ship_ship_hit_info->light_collision_cm_pos);
+	if (should_collide){
```
TRBlount, line 795: Why are there 3 separate if statements with the same condition? Looks like the first two could be combined at least.

code/object/collideshipship.cpp
```diff
@@ -751,17 +792,23 @@ void calculate_ship_ship_collision_physics(collision_info_struct *ship_ship_hit_
...
 	// Try to move each body back to its position just before collision occured to prevent interpenetration
 	// Move away in direction of light and away in direction of normal
 	vec3d direction_light;	// direction light is moving relative to heavy
 	vm_vec_sub(&direction_light, &ship_ship_hit_info->light_rel_vel, &local_vel_from_submodel);
 	vm_vec_normalize_safe(&direction_light);
 
-	Assert( !vm_is_vec_nan(&direction_light) );
-	vm_vec_scale_add2(&heavy->pos, &direction_light,  0.2f * lighter->phys_info.mass / (heavy->phys_info.mass + lighter->phys_info.mass));
-	vm_vec_scale_add2(&heavy->pos, &ship_ship_hit_info->collision_normal, -0.1f * lighter->phys_info.mass / (heavy->phys_info.mass + lighter->phys_info.mass));
+	if (should_collide){
```
TRBlount, line 805: Please double check if these separate blocks are needed, or if there are missing conditions.

TRBlount (changes requested):

code/object/collideshipship.cpp
```diff
@@ -711,15 +741,26 @@ void calculate_ship_ship_collision_physics(collision_info_struct *ship_ship_hit_
 		impulse_mag = -impulse_mag;
 	}
 
+	// On multi clients should this collision occur? (limiting the frequency of applying the impulse prevents bugs)
+	bool should_collide = (!MULTIPLAYER_CLIENT 
+		|| (lighter == Player_obj && heavy->type == OBJ_SHIP && timestamp_elapsed(Ships[heavy->instance].multi_client_collision_timestamp))
+		|| (heavy == Player_obj && lighter->type == OBJ_SHIP && timestamp_elapsed(Ships[lighter->instance].multi_client_collision_timestamp)));
```
Baezon, line 747: Are these the right checks? Shouldn't the timestamp be checked/set for the player ship, rather than the things they're colliding into?

Baezon (changes requested):

code/object/collideshipship.cpp
JohnAFernandez: Yes, this is the correct way to do it.  I changed the design so that if multiple ships are colliding with the player, it will track the timestamp separately. 

As a verbose example for clarity: I'm about to collide with Foo and Bar?  I'll collide with Foo once and set that timestamp. Now I will wait to collide with Foo until it's time.  While I'm waiting I collide with Bar once, set that timestamp.   Foo's timestamp pops, now I can collide with Foo again but not with Bar until Bar's pops.

code/object/collideshipship.cpp
JohnAFernandez: Should be good now.

code/object/collideshipship.cpp
JohnAFernandez: Should be good now.

Baezon (approved):

code/object/collideshipship.cpp
JohnAFernandez: Combined.  Yeah, there's absolutely no reason to check twice.

code/object/collideshipship.cpp
JohnAFernandez: Combined. Yeah, there's absolutely no reason to check twice for the first two.

code/object/collideshipship.cpp
JohnAFernandez: The things that I cannot combine are the landing blocks, but I was able to nest the last conditional within one of the others.

code/object/collideshipship.cpp
JohnAFernandez: These should all be taken care of.

TRBlount (approved):
Looks to be good now.

PR #4210, opened 2022-04-24
QTFred Loadout Editor

JohnAFernandez:
Finally ready for review.

This includes the new required weapons feature.


MjnMixael (approved):
This looks good. Nothing jumps out to me in the code. I built & ran some tests. Everything works as I expect it should. Well done.

JohnAFernandez:
Thanks!

Ok.  I'm going to merge if for no other reason to avoid merge conflicts.  Since qtFred is still in development, if we need to make changes after this merges, there should be no issues making them.

PR #4211, opened 2022-04-24
Beam light cleanup

EatThePath:
`beam_add_light_small()` and `beam_add_light_large()` duplicate a lot of logic, and in working on #4208 I found this to be an obstacle to making any changes to them and my attempts to work around that are significant enough to deserve individual scrutiny in my mind. I've moved what are to me logically discrete steps of that shared logic into functions, and eliminated some logically unreachable code as well.


Goober5000 (approved):
Good cleanup and reorganization. 👍

PR #4213, opened 2022-04-25
add a script function to flash parts of the target box

Goober5000:
Add the `hu.flashTargetBox()` function.  Also move the `STR` macro to scp_defines.h and rename it `SCP_TOKEN_TO_STR`.


wookieejedi (approved):
LGTM

PR #4215, opened 2022-04-25
Early return instead of assert on 0 or less sized hud brackets

### LINKED ISSUE #4214: Hudbrackets assert
I have run into this four times now in the same mission. I'll potentially provide more debug info tomorrow because it's late.

I suspect this may be due to our script for drawing brackets around escort ships. I can provide the script if it helps. But other than that, the assert seems random. I can't tell any specific repeatable event that is causing it.

> Assert: "width > 0"
> File: hudbrackets.cpp
> Line: 60

Baezon:
This is a relatively minor issue that ought not to bring everything to a screeching halt.

I'm not going to say this will fix (github pls) #4214 just yet. It's possible other engine code is doing something wonky that resulted in bad parameters that warrants a bit more investigation.

### DIFF
--- a/code/hud/hudbrackets.cpp
+++ b/code/hud/hudbrackets.cpp
@@ -57,9 +57,9 @@ void draw_brackets_square(graphics::line_draw_list* draw_list, int x1, int y1, i
 	}
 	
 	width = x2 - x1;
-	Assert( width > 0);
 	height = y2 - y1;
-	Assert( height > 0);
+	if (width < 1 || height < 1)
+		return;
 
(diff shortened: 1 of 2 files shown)


wookieejedi (approved):

z64555:
Recommend doxy the new behavior within `hud/hudbrackets.h` and consider the other `draw_hud_bracket` functions, since they appear to have the same problem.

Baezon:
The other bracket functions do not assert a >0 area, so no, evidently they don't have the same issue. Since they never required that, I'm a bit reluctant to suddenly make them require it.

PR #4216, opened 2022-04-25
Fix RTT Cockpits Gauges with Change Ship Class

wookieejedi:
Previously, using the `change-ship-class` sexp or scripting function did not update RTT cockpit gauges. This PR adds the cockpit code update functions to fix the issue. Tested and works as expected.


JohnAFernandez (approved):
Yet another change ship class bug bites the dust.

PR #4217, opened 2022-04-25
Always record primary bank ammo capacity during ship creation.

MageKing17:
`ship_set_default_weapons()` (which is called by `ship_create()`) was only setting `swp->primary_bank_capacity` for a bank if the default weapon for that bank was ballistic; if the ship was then equipped with an ammo-using primary and the `set-primary-ammo` SEXP was used on it, it would think the bank had a capacity of zero, resulting in 0 ammo.

This commit simply moves the assignment outside of the conditional, so that primary bank capacities are always recorded in the `ship_weapon` class.


JohnAFernandez (approved):
Makes sense and straightforward.

PR #4218, opened 2022-04-25
make variable list parsing a little more robust

### LINKED ISSUE #4170: FRED should check for duplicate variable names and throw an error or warning
I get the assertion:
ASSERTION: "get_index_sexp_variable_name(var_name) == -1" at sexp.cpp:4106

Assert: "get_index_sexp_variable_name(var_name) == -1"
sexp.cpp
Line: 4106

when opening an old mission in FRED debug.  Per EatThePath this assertion is the result of duplicate variable names present in the mission - which is indeed the case.  However, FRED should either not allow the mission to be saved, or throw a warning or error message when this happens.  Currently there is none displayed on save or in the error checker.

Goober5000:
Properly report errors and warnings, and don't create duplicate variables.  Fixes #4170.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -4055,7 +4055,8 @@ int stuff_sexp_variable_list()
 			type = SEXP_VARIABLE_BLOCK;
 		} else {
 			type = SEXP_VARIABLE_UNKNOWN;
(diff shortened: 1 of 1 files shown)


BMagnu (approved):
Seems simple enough

PR #4220, opened 2022-04-26
Add cockpit FoV

### LINKED ISSUE #4219: FR: Allow separate FOV setting for cockpits
Currently, cockpits are rendered using the same FOV as the scene. With large enough FOV values, this can make the readouts hard to read, as they get smaller in proportion to the user's monitor. Conversely, this could also cause some cockpit HUD elements to be off-screen with a small enough FOV. 
Since cockpits are (as far as I understand) rendered separately from the scene and composited onto it, it should be possible to use a different FOV for it. 
I think ideally users should be able to set a global value for cockpit FOV, and SEXPs or scripting functions that affect FOV should get equivalent cockpit-related versions.  

Alternatively, cockpit FOV could be set as a multiplier of regular FOV, so for example if an user sets cockpit FOV to 0.5, their cockpit would use half the FOV used to render the scene. This would have the added benefit of not adding any difficulty to modders and scripters, as well as being backwards compatible. 

Example screenshots from FotG (shared with Wookiejedi's permission): 

Normal FOV
Screenshot from 2022-04-26 10-15-32

High FOV (using 1.3 fov, which is very high but serves the purpose of illustrating the issue)
Screenshot from 2022-04-26 10-16-23

 Low FOV (using 0.3 fov, same caveat as before)
Screenshot from 2022-04-26 10-17-16

BMagnu:
Implements / fixes #4219.
With this PR, SEXP'd FoV has precedence over the -fov_cockpit cmdline parameter, which has precedence over the -fov cmdline parameter which has precedence over the default 0.75 FoV.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -7332,6 +7332,9 @@ void ship_render_cockpit(object *objp)
 	vm_vec_unrotate(&pos, &sip->cockpit_offset, &eye_ori);
 
 	bool shadow_override_backup = Shadow_override;
+	float fov_backup = Proj_fov;
+
+	g3_set_fov(Sexp_fov <= 0.0f ? COCKPIT_ZOOM_DEFAULT : Sexp_fov);
 
 	//Deal with the model
 	model_clear_instance(sip->cockpit_model_num);
@@ -7366,6 +7369,8 @@ void ship_render_cockpit(object *objp)
 	render_info.set_replacement_textures(Player_cockpit_textures);
 
 	model_render_immediate(&render_info, sip->cockpit_model_num, &eye_ori, &pos);
+
+	Proj_fov = fov_backup;
 
 	gr_end_view_matrix();
 	gr_end_proj_matrix();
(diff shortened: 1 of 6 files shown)


code/ship/ship.cpp
```diff
@@ -7332,6 +7332,9 @@ void ship_render_cockpit(object *objp)
 	vm_vec_unrotate(&pos, &sip->cockpit_offset, &eye_ori);
 
 	bool shadow_override_backup = Shadow_override;
+	float fov_backup = Proj_fov;
+
+	g3_set_fov(Sexp_fov <= 0.0f ? COCKPIT_ZOOM_DEFAULT : Sexp_fov);
```
JohnAFernandez, line 7337: Am I reading this right that the sexp would adjust both the gobal and cockpit fov?

JohnAFernandez: Ok.  I guess it can always be expanded with a cockpit fov sexp later, especially if it's requested.

JohnAFernandez (changes requested):
If you could just add documentation to the sexp to mention that it will overwrite cockpit fov, it will get an approval from me.  We'll see if anyone else has any other issues.

JohnAFernandez:
If there are no objections, I think you can probably merge this @BMagnu .

PR #4222, opened 2022-04-27
Model code cleanup #1

TRBlount:
This is some much needed cleanup of the model code related to parsing some of the bsp data structure. There are some updates to macros, swapping out a function for some safe typecasting, and the addition of a new struct to make parsing TMAPPOLY less ugly. Also removed some magic numbers.


Baezon (approved):
Looks pretty straightforward to me

PR #4227, opened 2022-04-28
Fix multiplayer in game joining 

JohnAFernandez:
Again, creating early in draft since my dev environment is not really available.

In game joining has been on a long journey.  Volition added most of the original code back in 1999.  Since then, it has only been worked on sparsely for the past few years. Other multiplayer priorities, like making sure you can hit anything, have always superceded it.

With this, players will be able to join any multiplayer game that is in the game play state.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -6233,46 +6233,44 @@ bool post_process_mission()
 
 	// before doing anything else, we must validate all of the sexpressions that were loaded into the mission.
 	// Loop through the Sexp_nodes array and send the top level functions to the check_sexp_syntax parser
+	// Cyborg -- If you are a ingame joiner, your sexps will be taken care of by the server, and checking will 
+	// actually crash mission load, since a list of objects present are also given to them only after this point.
 
-	for (i = 0; i < Num_sexp_nodes; i++) {
-		if (is_sexp_top_level(i) && (!Fred_running || (i != Sexp_clipboard))) {
-			int result, bad_node, op;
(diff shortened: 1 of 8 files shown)


code/mission/missionparse.cpp
```diff
@@ -6177,40 +6177,44 @@ bool post_process_mission()
 
 	// before doing anything else, we must validate all of the sexpressions that were loaded into the mission.
 	// Loop through the Sexp_nodes array and send the top level functions to the check_sexp_syntax parser
+	// Cyborg -- If you are a ingame joiner, your sexps will be taken care of by the server, and checking will 
+	// actually crash mission load, since a list of objects present are also given to them only after this point.
+
+	if (!(Game_mode & GM_MULTIPLAYER) && (Net_player->flags & NETINFO_FLAG_INGAME_JOIN)) {
```
Goober5000, line 6183: This condition is incorrect (the logical NOT should apply to the conjunction of both flag checks)

JohnAFernandez: Thanks! Fixed

notimaginative:
Had a (very) quick look at the boost version of the bitset stuff and I think it might be better to implement a `to_block()`/`from_block()` for our bit set instead of using the string method. Basically just goes to/from a vector of single bytes: for count of bitset.size() divided by 7, apply set flags to a byte and push to vector. For multi we'd write a single byte for the vector size, then each byte in the vector, giving 7*255 total supported flags per bitset. Long term that should work much better given the space constraints of multi packets.

JohnAFernandez:
Closing since work continued on another branch.

Issue #4228, opened 2022-04-28
FR: Swap Ctrl-V and Ctrl-P in the Events Editor


MjnMixael:
OK, hear me out. Ctrl-V is Paste. We all know it as paste. It makes sense that it's tied to Paste in the Events Editor. For those that don't know, Ctrl-P is print.. er I mean Add Paste. I propose we swap that. Ctrl-V for Add Paste and Ctrl-P for Paste. Here's why.

1. Add Paste is the more common transaction. My evidence is anecdotal, but I definitely add paste way more than just paste. Ctrl-C, Ctrl-P is much more of a stretch than the Ctrl-C, Ctrl-V I use so often in general computing. That's important because the Events Editor is still M&KB required controls. You rarely have both hands on the keyboard.
2. Paste is destructive where Add Paste is not. If you paste over something accidentally, it's gone... with no way to get it back because we don't have Undo in the Events Editor. As an experienced FREDer, I have nightmares about Paste. The easier it is to Add Paste instead of Paste the better.
3. P is for **P**aste not for **A**dd paste in a sane and sensible universe. (This isn't really a third reason, but it's my FR... I can write what I want.)

Goober5000:
Strong no on this one.  Paste is a standardized operation; Add Paste is not.  Intuitively, the standard shortcut should go with the standard operation and the non-standard shortcut with the non-standard operation.

This is also anecdotal, but I use the shortcuts for Paste and Add Paste about the same, and the distinction between the two is clear.  Switching the shortcuts would add extra cognitive load to every use of *both* shortcuts which would take away the advantage of shortcuts in the first place.

Now, one way this *could* work is if customizable key bindings were added to qtFRED.  Then anyone could remap the keys to their heart's content.

MjnMixael:
Keeping destructive behavior in the forefront! Got it. I don't have enough energy to fight you on this. Go ahead and close it. Gotta love ui friction bruh!

MjnMixael:
Actually. One further thought. You're objectively wrong on paste. In normal operations, I have to select something in order to replace it. Think text. Normal paste with no other input is add paste.

Ok that's all the energy I have for this.

Goober5000:
But selecting things is intrinsic to how the event editor works.  There's no way to get a paste context without selecting a node.

MjnMixael:
Yeah whatever man. I don't expect you to actually understand the ins and outs of UI design. Just close this.

[...]

PR #4230, opened 2022-04-29
Clearer Paste function names and saner defaults

qazwsxal:
Fix for #4228

### DIFF
--- a/fred2/fred.rc
+++ b/fred2/fred.rc
@@ -450,7 +450,8 @@ BEGIN
         MENUITEM SEPARATOR
         MENUITEM "Cut\tCtrl+X",                 ID_EDIT_CUT, GRAYED
         MENUITEM "Copy\tCtrl+C",                ID_EDIT_COPY
-        MENUITEM "Paste\tCtrl+V",               ID_EDIT_PASTE, GRAYED
+        MENUITEM "Paste (Below)\tCtrl+V",       ID_EDIT_PASTE, GRAYED
+        MENUITEM "Paste (Overwrite)\tCtrl+P",   ID_EDIT_PASTE_SPECIAL, GRAYED
         MENUITEM SEPARATOR
         POPUP "Add Operator"
         BEGIN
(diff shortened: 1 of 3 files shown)


Goober5000 (changes requested):
As I said in the issue, the Ctrl+V shortcut belongs with the original paste, not the Add Paste.  More importantly, this PR breaks a lot of things.  First, the `ID_EDIT_PASTE` and `ID_EDIT_PASTE_SPECIAL` should keep their original meanings, or at the very least, be named appropriately.  The menu change removes the separation of items that was added to reduce accidental misclicks.  And the change to the shortcut handling breaks the stay-on-node behavior that is needed for repeated keypresses.

qazwsxal:
Happy to re-order menu items back to their originals and have done so in the most recent commit. Unfortunately, `ID_EDIT_PASTE` and `ID_EDIT_PASTE_SPECIAL` are defined by the ATL/MFC library. I believe this boils down to the same semantic issue present in the original FR. The majority opinion is that the expected behaviour of a CTRL+V (the "normal" paste) operation is to add clipboard content without overwriting any existing data.
I am unable to reproduce the "stay on node" behaviour breakage, repeatedly add-pasting with CTRL+V is adding new nodes underneath without moving the selected node. I'm also not sure how this might have been introduced, as the only changes are function calls being swapped over without any editing of the calls themselves.

Goober5000 (commented):
I would suggest "Paste (Add Child)" instead of "Paste (Below)".

The current version of the code looks like it should work properly with regards to repeated Add Paste actions, although I would have just edited the keys instead of moving code around.

The original request dealt with just the shortcut keys, but this PR (although it fixes the separation) nevertheless swaps the destructive and non-destructive menu items.  This breaks muscle memory precedent.

EatThePath:
> The original request dealt with just the shortcut keys, but this PR (although it fixes the separation) nevertheless swaps the destructive and non-destructive menu items. This breaks muscle memory precedent.

For what it's worth I would agree that this is probably undesirable. It's a bit awkward for new user accessibility, but unlike the hotkey this *is* a 20 year old usage pattern that probably should be kept unless doing a truly sweeping UI overhaul.

Goober5000 (approved):
This is acceptable.

EatThePath:
For the record I also am good with this outcome from an interface perspective, though I'm not equipped to review the actual MFC code

PR #4231, opened 2022-04-30
fix up conditions for script hooks

### LINKED ISSUE #3874: Many hook condition combinations do not work
Things like `$Weapon class:` or `$Ship class:` listed before the hook itself. Many combinations are not supported, sometimes this is simple oversights, but others are more fundamental to how the system is handled. I don't really have the time to list and test all the combinations that would seem like they should work but don't, but here's a few examples:

On Weapon Created can't use weapon class conditions.

On Beam Collision's conditions apply to the *other* object in the collision, and so you can't condition on the beam's weapon class.

On Ship Collision, same thing, can't condition using any ship-related condition (ship name/class/type), unless the other object is a ship.

Overall, I've had an extremely poor success rate of attempting to use this system (and usually just do it manually in the script), if it were just one omission I'd be happy to fix it myself, but it's so common that I think it's clear this entire system needs a serious audit, or a complete refactor to actually work more generally.

Goober5000:
For script hooks which use conditions, be more consistent about evaluating both objects involved, as well as evaluating the different types of objects.

Fixes #3874.

### DIFF
--- a/code/scripting/scripting.cpp
+++ b/code/scripting/scripting.cpp
@@ -425,10 +425,20 @@ bool ConditionedHook::ConditionsValid(int action, object *objp1, object *objp2, 
 			{
 				for (auto objp : objp_array)
 				{
-					if (objp != nullptr && (objp->type == OBJ_WEAPON || objp->type == OBJ_BEAM))
+					if (objp == nullptr)
+						continue;
+
+					if (objp->type == OBJ_WEAPON)
 					{
 						// scp.condition_cached_value holds the weapon_info_index of the requested weapon class
 						if (Weapons[objp->instance].weapon_info_index != scp.condition_cached_value)
+							return false;
+						break;
+					}
+					else if (objp->type == OBJ_BEAM)
+					{
+						// scp.condition_cached_value holds the weapon_info_index of the requested weapon class
+						if (Beams[objp->instance].weapon_info_index != scp.condition_cached_value)
 							return false;
 						break;
 					}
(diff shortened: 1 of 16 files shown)


code/scripting/scripting.cpp
```diff
@@ -345,249 +346,284 @@ bool ConditionedHook::ConditionsValid(int action, object *objp, int more_data)
...
-						return false;
-					if(len > 4 && !stricmp(&Mission_filename[len-4], ".fs2"))
-						len -= 4;
-					if(strnicmp(scp.condition_string.c_str(), Mission_filename, len) != 0)
-						return false;
-					break;
+					if (objp != nullptr && (objp->type == OBJ_WEAPON || objp->type == OBJ_BEAM))
+					{
+						// scp.condition_cached_value holds the weapon_info_index of the requested weapon class
+						if (Weapons[objp->instance].weapon_info_index != scp.condition_cached_value)
```
Baezon, line 431: There will need to be a separate check for `OBJ_BEAM` in the `Beams` array since I don't believe those have a weapon object at all (making this an invalid access).

Baezon (changes requested):
Nice, definitely like this design more, with an object array instead. There is the possibility some edge case might not work now where it did before, but I don't believe so, and it shouldn't be hard to fix if so.

code/scripting/scripting.cpp
Goober5000: Good catch.  Updated.

Baezon (approved):

PR #4232, opened 2022-04-30
add some UX guards to the briefing editor icons

### LINKED ISSUE #3860: Briefing Icons have unintuitive behavior for FREDers
Briefing icons can be set to an icon image; this is the retail class. It ranges from Fighter to Juggernaut and includes things like Asteroid, Jump Node, Waypoint, Unknown, Planet, etc. Briefing icons also have a model field where you can select the ship class to be shown. FSO has a feature where you can designate custom icons per ship class in the ships.tbl. If a model has an icon specified in the ships.tbl then that icon will be used instead of the selected icon image from the above field.

However, if you select a model with an icon, then the Icon Image field gets greyed out because it's unused... sort of... If the Icon Image is set to Asteroid or a handful of others while at the same time a model is chosen that has its own icon specified, then the model's icon will be used but the model displayed when the player clicks on it will be the asteroid. It's like Asteroid, Jump Node, et al are overrides to the model but not the icon while at the same time the image field is greyed out so you can't change it anymore. The FREDer has to select a model that doesn't have an icon specified, then change the Icon Image away from any of the overrides, then change the model back to the ship class they want to use.

Personally, I think the easiest solution is simply to not grey out the Icon Image field in FRED. 

I don't have a ready-made mod to show this practice off, but you could pretty easily do it with the currently released BtA mod. It makes extensive use of per-ship-class icons.

Goober5000:
1) Prevent the ship class from being changed if the icon image has a special closeup (like a jump node)
2) If the icon image is changed to a special closeup, automatically choose a ship class that doesn't have a custom icon

Fixes #3860

### DIFF
--- a/code/mission/missionbriefcommon.cpp
+++ b/code/mission/missionbriefcommon.cpp
@@ -643,14 +643,9 @@ void brief_init_colors()
 {
 }
 
-briefing_icon_info *brief_get_icon_info(brief_icon *bi)
+bool brief_special_closeup(int briefing_icon_type)
 {
-	if (bi->ship_class < 0)
-		return NULL;
-	ship_info *sip = &Ship_info[bi->ship_class];
-
-	bool allow_override = true;
-	switch (bi->type)
+	switch (briefing_icon_type)
 	{
 		case ICON_PLANET:
 		case ICON_ASTEROID_FIELD:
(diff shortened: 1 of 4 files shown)


wookieejedi (approved):
LGTM

PR #4233, opened 2022-04-30
Add TBL Viewer

TheForce172:
Adds TBL Viewer to the ship editor

No picture as it looks identical to the old one.

### DIFF
--- a/qtfred/src/ui/dialogs/ShipEditorDialog.cpp
+++ b/qtfred/src/ui/dialogs/ShipEditorDialog.cpp
@@ -183,7 +183,9 @@ void ShipEditorDialog::on_initialOrdersButton_clicked()
 
 void ShipEditorDialog::on_tblInfoButton_clicked()
 {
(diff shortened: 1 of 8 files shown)


qtfred/src/ui/dialogs/ShipEditorDialog.cpp
```diff
@@ -156,6 +156,16 @@ ShipEditorDialog::ShipEditorDialog(FredView* parent, EditorViewport* viewport)
 
 ShipEditorDialog::~ShipEditorDialog() = default;
 
+int ShipEditorDialog::getShipClass()
+{
+	return _model->getShipClass();
+}
+
+bool ShipEditorDialog::getMulti()
```
JohnAFernandez, line 164: Multi has a lot of different meanings: multiplayer, multicontroller.  If you could make this a little more specific for readability that would be good.

Goober5000 (commented):
Concur with @JohnAFernandez , otherwise this looks good

Goober5000 (approved):

PR #4234, opened 2022-04-30
Encapsulate testing for LoS in a function and expose to Lua

BMagnu:
This adds the ability to test for a free line of sight from Lua.

The small change to the scripting internals in this PR extends the functionality of LuaTable. Previously, LuaTables were already interactible from our C++ code, but we could only get lua-internal values like number or booleans. This change allows to also extract userdata like objects or ships from these LuaTables.


code/object/objcollide.cpp
```diff
@@ -1038,3 +1040,107 @@ void obj_sort_and_collide(SCP_vector<int>* Collision_list)
 	}
 	obj_find_overlap_colliders(sort_list_y, sort_list_z, 2, true);
 }
+
+bool test_line_of_sight(vec3d* from, vec3d* to, std::unordered_set<const object*>&& excluded_objects, bool test_for_shields, bool test_for_hull, float* first_intersect_dist, float threshold) {
```
asarium, line 1044: This function is a higher level function unlike the other functions in this file. It would be better to move this to a more object specific place like the object folder and possibly into its own file.

code/scripting/lua/LuaValue.h
```diff
@@ -159,6 +159,21 @@ class LuaValue {
...
+	 */
+	template<typename Type>
+	void getValue(scripting::ade_odata_getter<Type>&& od) const {
+		_reference->pushValue(_luaState);
+
+		if (!convert::popValue<Type>(_luaState, std::forward<scripting::ade_odata_getter<Type>>(od))) {
+			lua_pop(_luaState, 1);
+			throw LuaException("Failed to pop value");
+		}
+	}
```
asarium, line 175: This type is needlessly specific. It should be possible to change this to a generic rvalue reference which will avoid the direct dependency on the ade types:
```suggestion
	template<typename Type>
	void getValue(Type&& value) const {
		_reference->pushValue(_luaState);

		if (!convert::popValue<Type>(_luaState, std::forward<Type>(value))) {
			lua_pop(_luaState, 1);
			throw LuaException("Failed to pop value");
		}
	}
```

asarium (changes requested):

code/object/objcollide.cpp
Goober5000: I'd suggest keeping it in aicode.cpp, even if it's now a separate function.

code/scripting/api/libs/mission.cpp
```diff
@@ -1456,6 +1459,67 @@ ADE_FUNC(isRedAlertMission,
...
+				}
+				catch (const luacpp::LuaException& /*e*/) {
+					// We were likely fed a userdata that was not an object. 
+					// Since we can't actually tell whether that's the case before we try to get the value, and the attempt to get the value is printing a LuaError itself, just eat the exception here and return
+					return ADE_RETURN_FALSE;
+				}
+			}
+			else {
+				//This happens on a non-userdata value, i.e. a number
+				LuaError(L, "Table with objects to be excluded contained non-object values! Aborting...");
```
asarium, line 1499:
```suggestion
				LuaError(L, "Table with objects to be excluded contained non-userdata values! Aborting...");
```

code/scripting/lua/LuaValue.h
```diff
@@ -159,6 +159,21 @@ class LuaValue {
 		}
 	}
 
+	/**
+	 * @brief Gets the value of a userdata type using the passed getter function or throws an exception
```
asarium, line 163: The comment is outdated now.
```suggestion
	 * @brief Same as above but allows passing a reference to a value where to store the lua value
```

asarium (approved):
I only found some comments that are not entirely correct now. Looks good to me otherwise.

PR #4235, opened 2022-04-30
Add Subsystem Flag to Hide Turrets from Ship Loadout Stats

wookieejedi:
The statistics summary window in the ship loadout editor shows various aspects and stats of the selected ship. These can be toggled and customized through the ships table, but one line is actually autogenerated, `+Turrets`. This is unideal for certain ships, especially those which use turrets on ships in non-standard ways (for example having an AI controlled turret which shoots bombs). In these cases, the turrets are not actual turrets, and it is confusing for players seeing that there is a turret on the ship in the loadout editor. (I can give more detailed examples using FotG if needed).

This PR adds a subsystem flag which allows the modders to hide a turret subsystem so it is not counted in the auto generation of the `Turrets` summary line. (As a slight cleanup I also removed cleaned up the whitespace formatting on a nearby, but unrelated line).

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -265,7 +265,8 @@ flag_def_list_new<Model::Subsystem_Flags> Subsystem_flags[] = {
 	{ "don't autorepair if disabled", Model::Subsystem_Flags::No_autorepair_if_disabled,        true, false },
 	{ "share fire direction",       Model::Subsystem_Flags::Share_fire_direction,               true, false },
 	{ "no damage spew",             Model::Subsystem_Flags::No_sparks,                          true, false },
-	{ "no impact debris",           Model::Subsystem_Flags::No_impact_debris,                    true, false },
+	{ "no impact debris",           Model::Subsystem_Flags::No_impact_debris,                   true, false },
+	{ "hide turret from ship summary", Model::Subsystem_Flags::Hide_turrets_in_loadout_stats,   true, false },
 };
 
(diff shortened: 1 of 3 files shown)


JohnAFernandez (approved):
Approved.  How many subsystem flags does that bring us to?

wookieejedi:
> Approved. How many subsystem flags does that bring us to?

Quick count looks like 42?

code/ship/ship.cpp
```diff
@@ -265,7 +265,8 @@ flag_def_list_new<Model::Subsystem_Flags> Subsystem_flags[] = {
 	{ "don't autorepair if disabled", Model::Subsystem_Flags::No_autorepair_if_disabled,        true, false },
 	{ "share fire direction",       Model::Subsystem_Flags::Share_fire_direction,               true, false },
 	{ "no damage spew",             Model::Subsystem_Flags::No_sparks,                          true, false },
-	{ "no impact debris",           Model::Subsystem_Flags::No_impact_debris,                    true, false },
+	{ "no impact debris",           Model::Subsystem_Flags::No_impact_debris,                   true, false },
+	{ "hide turret from ship summary", Model::Subsystem_Flags::Hide_turrets_in_loadout_stats,   true, false },
```
Goober5000, line 269: There's a difference between the quoted name and the flag name.  Did you want to sync those up?

wookieejedi: Good point, just updated. Would be happy to continue to refine if needed.

Goober5000: Looks good, although you missed one spot. ;)

wookieejedi: Opps, forgot to press save for each file before committing.

PR #4241, opened 2022-05-04
Clean up QTfred Save

TheForce172:
Stops QTFred From saving the entire messages.tbl to the mission file

### DIFF
--- a/qtfred/src/mission/Editor.cpp
+++ b/qtfred/src/mission/Editor.cpp
@@ -555,6 +555,8 @@ void Editor::clearMission() {
 	ENVMAP = -1;
 
 	missionLoaded("");
+	*Parse_text = *Parse_text_raw = '\0';
+	Parse_text[1] = Parse_text_raw[1] = 0;
 }
 
 void Editor::initialSetup() {


Goober5000 (commented):
This is already done up around line 477.  Why does it need to be done again?

TheForce172:
Because after that messages.tbl is parsed in messages_init which the leaves the full messages.tbl in Parse_text_raw

Goober5000:
Why doesn't the same thing happen in regular FRED?

TheForce172:
Not sure, regular Fred does something deep in the windows logic that causes it to loops around again, and because the message table has already been loaded it doesn't load the table again so the parse is kept cleared from the lines above. Rather than replicate this strange looping I thought it would be better to simply clear the pares after its used.

Goober5000:
That is a bad code smell.  It would be better to figure out why the looping is happening in the first place, why `Parse_text_raw` affects the mission save, and why ships.tbl being parsed after messages.tbl doesn't affect it.

TheForce172:
Parse_text_raw affect the mission save because it it loaded into raw_ptr which is used for comments in the save code and is expected to be empty. 
Ship_init doesn't cause a problem because it was already run on line 396 so the table was already parsed when its cleared on line 477.
As I said the looping is done somewhere in the MFC logic

Goober5000:
Well that's the first problem, albeit a trivial one ... `ship_init` is called more than once.

But this needs further investigation.  I have a hunch the MFC looping isn't the root cause of the problem.

Goober5000:
Ok, I think I see what's going on.

Goober5000:
I did some debugging this evening and you're right, the parsing is never cleaned up when FRED is initialized.  Presumably those earlier `Parse_text` statements *were* the cleanup, but they were overshadowed by later FRED additions, and the overshadowing was not a problem because of the looping as you mentioned.  (As for the looping, it is because `clear_mission()` is called multiple times in FRED and is not caused by MFC.)

So take a look at #4242.  This calls the proper parse cleanup routines, following the example from `player_select_close()`.

TheForce172:
Closed in favour of #4242

PR #4242, opened 2022-05-07
clean up some parse problems in FRED

Goober5000:
These were first noticed by @TheForce172 in PR #4241.  Instead manually editing the parse buffer, call the parsing functions to properly free the memory used by parsing the tables.  Also remove redundant calls to `ship_init`.


BMagnu (approved):
LGTM

PR #4243, opened 2022-05-07
disable Enter and Escape in the container dialog

### LINKED ISSUE #4240: Enter/Esc keys and Close button don't work as they should in Add/Modify Container FRED dialog
Pressing the Enter key in FRED's Add/Modify Container dialog closes the dialog without saving changes.
Pressing the Esc key cancels without asking whether to save changes.

Proposed behavior:
* Pressing Enter should do nothing, since its meaning in that dialog is ambiguous.
* Pressing Esc should either do nothing or bring up a "Save changes?" dialog with Yes/No/Cancel buttons

Goober5000:
Override some MFC signal paths to prevent Enter and Escape from doing their default behavior.  Also add a "save changes" prompt to the Cancel and Close buttons.

Fixes #4240.


jg18 (approved):
Looks good and passes manual test. Thanks so much for writing this!

The only downside is that the "Save changes?" dialog appears even when the FREDder hasn't made changes, but this is already known to happen with the Events Editor. I suppose I could create an Issue just to track it and put it on the backlog, since it's low priority IMO.

Goober5000:
> Looks good and passes manual test. Thanks so much for writing this!

You're welcome!

> The only downside is that the "Save changes?" dialog appears even when the FREDder hasn't made changes, but this is already known to happen with the Events Editor. I suppose I could create an Issue just to track it and put it on the backlog, since it's low priority IMO.

This is the reason for the following line:
```
if (true /*query_modified()*/) {
```
There is a `query_modified` function in other dialogs, but not in this one.  If someone were to write such a function, that should solve this issue.

PR #4244, opened 2022-05-07
Fix strange warp glow sizing

Baezon:
Due to all the attention on the warp portal recently, I noticed some strange inconsistencies with the 'warp glow', a simple bitmap at the center of the portal, which flickers and changes size based on the portal's lifetime. On the left here is a diagram of the size of this effect over the portal's time with the 'warp_flash' cmd line flag active (C on the right is with the flag inactive).

Untitled-2

As you can see, there are some strange dips in the middle, this PR removes these dips, making the flag active = A, and flag inactive = C.

This is what I believe to be most pleasing and what I interpreted to be meant by "warp flash", but the history of this effect is quite complicated, and mostly involved incorrect combinations of math for the effect and B is in fact the original volition code for this effect. I would be open to no flag = B as well, but either way that interpolation code was clearly never meant to be mixed with warp flash, which only makes sense as being regular-size most of the time with large spikes at the beginning and end.

Here's some demonstrations, with the size of the effect increased overall and noise removed for demonstrative purposes:

Current:

https://user-images.githubusercontent.com/17305613/167243227-800d89ba-f65d-482d-b94e-aa91cffa481c.mp4

A:

https://user-images.githubusercontent.com/17305613/167243260-75e518b8-6146-458e-9f27-77ec5af23d95.mp4

B:

https://user-images.githubusercontent.com/17305613/167243263-0dd26f85-de7d-45d5-8ce3-8542e77a4bf9.mp4

C:

https://user-images.githubusercontent.com/17305613/167243265-1dbd354b-a798-4bea-88b4-348518f2d32a.mp4


BMagnu (commented):
Purely code-wise this looks good, and I'd approve it, if purely for the code.
While the code to result in the current effect looks intentional, I also can't see any proper reason for as to why it was made that way. So, IMO, A when the flag is active and C if it is not is the most sensible way, as done in this PR (even though I'd have made it a -mod or -aip flag, or even better, configurable per warp effect, as opposed to a command line flag, but I guess that discussion comes a few years too late now).

Before approving / merging this though, it should get some more comments from other devs on the topic, as this is a fairly substantial aesthetic choice.

Baezon:
I don't believe the end result of the combination of those linear interpolations in the middle with the flashes at the ends was intentional, each thing separately was done intentionally, but I don't think the result was tested/scrutinized well enough for them to determine they were combining in a clearly buggy way.

Certainly, Gamm39er's interest in this has indicated, that this should probably be more thoroughly exposed at some point.

JohnAFernandez:
It disappearing like that twice is not an acceptable way for it work, imho.  It's basically nonsense.  I also think A and C make a lot of sense as the fixed options.

Goober5000:
I see what you mean about the warp effect.  Well done figuring that out.

I'd prefer to keep B as the flag-less warp effect.  It provides a sort of emphasis on the warp portal, almost as if the portal is glowing brighter after is is fully open.  It also stands out more than it would if the bitmap size closely tracked the portal size.  And it's the original Volition pattern.

What about turning this into a game_settings.tbl flag and providing all three options?

EatThePath:
> While the code to result in the current effect looks intentional, I also can't see any proper reason for as to why it was made that way.

I suspect it may have been in mimicry of a star trek warp flash? Flash, ship or portal shrinks away to nothing, and then another bright flash after everything has vanished. It's not a perfect imitation but I think it's probably intended, given the name of the flag... Option A does make for a similar effect, but the lack of the dissapearing before the flash kind of breaks the resemblance honestly

JohnAFernandez:
So yeah, option A?

Baezon:
If someone wants to throw an approval in, I think we should go with the current version of the PR (flag = A, no flag = C), which involves the least change to the current behavior while fixing the bug. Re-adding B as a game settings option, or any other kind, is trivial should someone want to do it afterward.

PR #4245, opened 2022-05-07
Fix background sun positioning as well

Baezon:
Follow-up bugfix for #4186. Suns are getting their positions and everything stored and retrieved correctly, but the math that positions them was never corrected like bg bitmaps are now. The default sun position must be corrected as well, since that is entirely internal.


wookieejedi (approved):

Goober5000 (approved):
Good catch.

PR #4246, opened 2022-05-07
Fix memory leak and use tmap verts directly

### LINKED ISSUE #4185: Large memory dump occurs when a debug build exits
Steps to reproduce:
1) Compile and run a debug build ~~of FRED~~ in the Visual Studio debugger
2) Save a blank mission
3) Exit
4) The game will print an enormous memory dump to the debug log

It takes a long time for the memory dump to be printed - about a minute.  As far as I can determine, the dump occurs because a memory leak was detected.

EDIT: Turns out this happens in both FRED and FSO.

Baezon:
There were several dynamic arrays `new`'d but not `delete[]`ed, however in this case they were largely unecessary. Instead of copying the `model_tmap_vert_old`s into `model_tmap_vert`s and using those, the `model_tmap_vert_old`s can be used directly. Also do some updating of the easily-missed-because-it-doesn't-get-compiled-on-my-machine swap bsp functions for big endian.

Should fix #4185

### DIFF
--- a/code/model/modelcollide.cpp
+++ b/code/model/modelcollide.cpp
@@ -448,36 +448,30 @@ void model_collide_parse_bsp_tmappoly(bsp_collision_leaf *leaf, SCP_vector<model
 
 	uint i;
 	uint nv;
-	model_tmap_vert *verts;
 
-	nv = uw(p+36);
+	nv = uw(p + TMAP_NVERTS);
 
 	if ( nv > TMAP_MAX_VERTS ) {
 		Int3();
 		return;
 	}
 
-	int tmap_num = w(p+40);
+	int tmap_num = w(p + TMAP_TEXNUM);
 
 	Assert(tmap_num >= 0 && tmap_num < MAX_MODEL_TEXTURES);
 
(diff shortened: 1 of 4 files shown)


JohnAFernandez (commented):
It's definitely a lot cleaner looking now.  I can't pretend that I understand fully what's happening under the hood. But we just have to make sure that we did some basic testing with these changes.  If that was done, I can approve for the sake of getting the fix in.

Baezon:
Yes, I tested basically by just going in game and seeing that ships still look and collide correctly. Getting this kind of model stuff wrong at all tends to be pretty all or nothing, either everything works fine or it has failed catastrophically. I'm not as sure about the memory leaking by explicit testing, but given that Goober narrowed it down to the exact commit which introduced a bunch of `new`s which this removes, I'm 99% sure this is it.

JohnAFernandez (approved):

TRBlount (approved):
These changes are perfectly acceptable. The issue was essentially not calling `delete[]` on `verts`.

PR #4248, opened 2022-05-09
fix Asserts added in #4212

Goober5000:
Freeing a node of -1 should simply return, not Assert.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -1385,13 +1385,13 @@ int free_sexp(int num, int calling_node)
 {
 	int i, rest, count = 0;
 
-	Assert((num >= 0) && (num < Num_sexp_nodes));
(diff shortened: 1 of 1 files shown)


MageKing17:
In what scenario are we currently calling `free_node(-1)`, and why isn't that a problem worth fixing?

Goober5000:
I ran into that Assert with these lines in ship_delete:
```
	free_sexp2(shipp->arrival_cue);
	free_sexp2(shipp->departure_cue);
```
My assumption was that these had already been freed after the ship arrived, but it looks like that may not have been the case.  However, these ships were created with the ship-create sexp.  It would make sense to assign cues when those ships are created.

Goober5000:
Ok, this PR is probably obsolete.  New PR shortly.

PR #4249, opened 2022-05-09
fix assignment of alt-name when using the hash

Goober5000:
Alt-names with hashes that were added by the `change-alt-name` sexp used to have their hashes truncated properly.  This most likely broke when hash truncation was moved from the HUD code to the "display name" code.  This PR truncates hashes in the same way as truncation elsewhere.


wookieejedi (approved):
LGTM

PR #4250, opened 2022-05-09
assign cues when a ship is ship-created

### LINKED ISSUE #4251: Assert in F3 Lab after viewing a ship
Tested on latest nightly. Go to F3 Lab. View a ship. Either then try to view another ship or ESC to leave the lab. You will get this Assert.

Assert: "(num >= 0) && (num < Num_sexp_nodes)"
sexp.cpp
Line: 1427

Goober5000:
Assign locked-* cues to ship-created ships so that their arrival and departure cues are not -1.  Supersedes PR #4248.

Fixes #4251.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -6689,6 +6689,10 @@ static void ship_set(int ship_index, int objnum, int ship_type)
 	shipp->warpin_params_index = sip->warpin_params_index;
 	shipp->warpout_params_index = sip->warpout_params_index;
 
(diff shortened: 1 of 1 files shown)


MageKing17 (approved):
Seems like a good fix.

Goober5000:
I moved this from the `ship-create` sexp into the `ship_set` function itself, since it also comes into play when ships are created through scripts, as well as in the ship lab as shown by #4251.

MageKing17 (approved):

PR #4255, opened 2022-05-10
Fix clear-container SEXP

jg18:
Follow up on PR #3862 to make two changes to the `clear-container` SEXP:
1. Fix SEXP help text
2. Allow for any number of arguments, rather than only one.


jg18:
If you check the "Hide whitespace" box in the Settings under the "Files changed" tab, the PR becomes super-small.

Goober5000 (approved):

JohnAFernandez (approved):
Looks good.  The only thing I could think of is somehow the new loop becoming infinite.  But the way sexps are parsed does not allow that, aiui.

jg18:
@JohnAFernandez The loop will terminate, because the SEXP always has a finite number of arguments.

I just tried out the modified SEXP in FRED.

Although Add Data brings up the right list of suggestions, the newly added data node appears with the Data icon, rather than the Container Name icon. However, I don't think this is an issue, since only FRED cares about the distinction. and using Replace Container Name still works.

A more serious issue is that the Add Data submenu also has Add String enabled, but I have an idea for how to fix that, which will be in a separate PR.

PR #4256, opened 2022-05-11
partial revert of targeting distance ranking change

Goober5000:
Normally, when the player cycles through targets using (for example) the hostile targeting key, the targets will be shown in order of distance.  In PR #4121, this was changed from bounding-box distance to center-distance, which has very little noticeable effect on the cycle except that it now agrees with the AI strategy for measuring distance.

However, when the player targets a hostile for the first time, or when the cycle resets after 1.5 seconds have passed, the first ship targeted will be the closest ship *or* the ship with the closest turret.  And furthermore, when the target cycle continues, the next ship targeted will be the next ship farther away.  There can be a dramatic difference between a turret, which may be very close, and the center of the ship, which may be very far away.  The target cycle will continue with these far-away ships, and this has the effect of leapfrogging any ships that are closer that should be targeted first.

There isn't any easy way to tell that a ship has been targeted because of its turret, and not because of some other targeting key or because the cycle has restarted, but fortunately the bounding-box distance is a reasonable approximation because turrets tend to be on the bounding boxes.  So this commit simply restores the distance functions to what they were before.  The other fixes from #4121 are not related and remain in place.


wookieejedi (approved):
Clear revert, looks good.

PR #4257, opened 2022-05-11
Ensure Replace/Add String and (Add) Paste are disabled for container name nodes

jg18:
Container name nodes should never be arbitrary strings, and pasting over or add-pasting into SEXPs that expect them shouldn't be allowed.


Goober5000 (approved):
LGTM too

PR #4258, opened 2022-05-11
Correct sun position compensation math

Baezon:
Due to extra 'bank' shenanigans background bitmaps have a more complicated conversion, but suns don't have this so their conversion is simpler (and also incorrect if you try to do the banking stuff). One result of this now is that 'banking' a sun bitmap in FRED will have no apparent effect, but given that sun bitmaps could never actually be banked anyway (it only affected it's 2d position before), this isn't really a loss.

### DIFF
--- a/code/starfield/starfield.cpp
+++ b/code/starfield/starfield.cpp
@@ -2425,7 +2425,21 @@ int stars_add_bitmap_entry(starfield_list_entry *sle)
 	return (int)(Starfield_bitmap_instances.size() - 1);
 }
 
-void stars_correct_background_angles(angles *angs_to_correct)
+void stars_correct_background_sun_angles(angles* angs_to_correct)
+{
+	matrix mat;
+	vm_angles_2_matrix(&mat, angs_to_correct);
+	vm_transpose(&mat);
+	vm_extract_angles_matrix(angs_to_correct, &mat);
+}
+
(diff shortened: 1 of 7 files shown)


code/starfield/starfield.cpp
```diff
@@ -2425,7 +2425,24 @@ int stars_add_bitmap_entry(starfield_list_entry *sle)
...
 
-void stars_correct_background_angles(angles *angs_to_correct)
+void stars_correct_background_sun_angles(angles* angs_to_correct)
+{
+	matrix mat;
+	vm_angles_2_matrix(&mat, angs_to_correct);
+	vm_transpose(&mat);
+	vm_extract_angles_matrix(angs_to_correct, &mat);
+	angs_to_correct->p = fmod(angs_to_correct->p + PI2, PI2);
+	angs_to_correct->b = fmod(angs_to_correct->b + PI2, PI2);
```
Goober5000, line 2435: Does this assignment of the bank actually do anything now?

Baezon: The bank value is still there, and of course, this is used in both directions, so it's important to restore the old value too. Only for completeness do I do this, since the weirdness with bank is technically outside the scope of this function. These `fmod`s are also here just to make sure its in the 0 - 2PI range since FRED complains otherwise.

Goober5000 (approved):

PR #4260, opened 2022-05-11
revised disabling of Enter and Escape

Goober5000:
Turns out PR #4243 actually introduced a bug that appeared to implement the feature.  The previous implementation had the effect of intercepting all keypresses and routing them to the main window.  This fixes that bug by using `CDialog` instead of `CWnd`.  It also turns out that the `hWnd` window handle for some reason does not match the expected window, but fortunately, this implementation is only called in the correct window so the restriction is not needed.


JohnAFernandez (approved):

jg18 (approved):
Looks good and also passes manual test.

z64555:
TFW a bug did the intended behavior of the desired feature.

Goober5000:
> TFW a bug did the intended behavior of the desired feature.

Yeah, pretty crazy.

PR #4261, opened 2022-05-11
disable sun bank controls

Goober5000:
Per Asteroth, and following PR #4258, sun banking is no longer meaningful and should be disabled.


JohnAFernandez (approved):

Baezon:
This will gray it out, right? I wonder if that would cause more confusion than simply removing it.

Goober5000:
It will gray it out, yes.

PR #4262, opened 2022-05-12
miscellaneous fixes related to old Coverity fixes

Goober5000:
1) Remove the extra branch in `check_ok_to_fire` that was added in SVN revision 10392, or commit 0d54630e83419f53ede08056bac3b26928d059cb.  If there was an actual Coverity error, this does not fix it at the proper location.  More importantly, this breaks certain turret firing behavior.
2) Fix a couple of bounds checks.
3) Remove an incorrect stale comment.

This fixes the broken turret swarm weapons that were uncovered by the fixes in #2959.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -6347,11 +6347,6 @@ int check_ok_to_fire(int objnum, int target_objnum, weapon_info *wip, int second
 			}
 		}
 	}
-	else
-	{
-		// We have no valid target object, we should not fire at it...
-		return 0;
-	}
 
 	return 1;
 }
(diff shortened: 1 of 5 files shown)


Baezon:
`turret_fire_weapon` uses `turret_enemy_objnum` as a valid index in several places, so I think those really need to be audited as well if we're going to assert that it is valid to fire without a target in general, or at least gate it behind spawn weapons.

Goober5000:
Fair enough.

Goober5000:
@Baezon Done.

Baezon (approved):

PR #4264, opened 2022-05-13
Multi sexp fixes

notimaginative:
Addresses some multi sexp issues:
* try to handle incompatible sexp packets better
* fix operator name issue with debug messages
* enable the new show-subtitle-text options

### DIFF
--- a/code/network/multi_sexp.cpp
+++ b/code/network/multi_sexp.cpp
@@ -7,6 +7,7 @@
 #include "network/multi_sexp.h"
 #include "parse/sexp.h"
 #include "network/multi.h"
+#include "network/multi_log.h"
 #include "network/multiutil.h"
 
 static const std::uint8_t CALLBACK_TERMINATOR = 255;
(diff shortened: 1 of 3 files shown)


notimaginative:
This is a breaking change so it's in draft status until we're ready to merge other breaking multi PRs to avoid so many version bumps.

notimaginative:
@JohnAFernandez It passed all of my tests so I think it's ready to go. We just need to make sure it ends up in the same nightly as one of your changes which includes a version bump.

JohnAFernandez (commented):
Not that many lines changed.  And it looks straight forward.  Because of my current state, I'll take another look tomorrow, just to see if I can catch anything.

Goober5000 (commented):
There are four total instance of
```
// TODO: uncomment when Github ticket #3773 is implemented
```
Did you want to uncomment all of them?

notimaginative:
Yes I did, but obviously did a poor job of looking for them.

Goober5000 (approved):

PR #4265, opened 2022-05-13
fix a standalone crash from disabled uniform buffer manager

notimaginative:


JohnAFernandez (approved):
Simple fix.

Goober5000 (approved):
Yup, simple and straightforward.

PR #4267, opened 2022-05-14
Add cancel-future-waves

naomimyselfandi:
Ever wanted to tell a wing to _stop arriving?_ Now you can.


Baezon (approved):
Fantastic idea.

code/parse/sexp.cpp
```diff
@@ -13220,6 +13221,16 @@ void sexp_self_destruct(int node)
 	}
 }
 
+void sexp_cancel_future_waves(int node)
+{
+	for (int n = node; n != -1; n = CDR(n))	{
+		auto wingp = eval_wing(n);
+		if (wingp) {
+			wingp->num_waves = wingp->current_wave;
```
Goober5000, line 13229: This is an interesting approach.  I was a little bit skeptical so I looked through all places `num_waves` and `current_wave` are used.  I didn't find any explicit errors, although the `ship-type-destroyed` SEXP might become inaccurate depending on whether you consider the future wings not-yet-arrived or never-existed-in-the-first-place.  So this might be fine, but I suggest a different approach.

Typically, when I want a wing to stop arriving, I make the stop-arrive condition part of the wing's arrival cue.  What about doing the same thing here - when `cancel-future-waves` is run, free the ship arrival cues and then set them all to `Locked_sexp_false`?

naomimyselfandi: Like `is-destroyed-delay`, `ship-type-destroyed` simply won't see the canceled future waves. I think that's the most intuitive and most useful behavior, but it's probably worth documenting explicitly, so I'll do that.

(I have a test mission with Alpha (three Ulysses) against Aries (three waves of two Seths) and an event that calls `( cancel-future-waves "Aries" )` once Aries 3 arrives. I added a test for `ship-type-destroyed` just now. `( ship-type-destroyed 50 "fighter" )` becomes true once Alpha shoots down Aries 1-4.)

I chose this approach because it most closely matches how existing code checks for future wings, so it gave me the most confidence that I'm not introducing any broken corner cases. I think it's also the simplest approach.

Goober5000: Ok, that's reasonable, and the help text addition is good too

Goober5000 (approved):

PR #4268, opened 2022-05-14
some wing wave cleanup

Goober5000:
In the process of looking through `num_waves` and `current_wave` for #4267, I found a couple of confusing bits of code.  This clarifies things.


wookieejedi (approved):
Looks good, especially since it's just comment cleanup and 1 formatting line

PR #4271, opened 2022-05-14
fix maybePlayCutscene arguments

Goober5000:
This must have been coded as `ui:maybePlayCutscene` when it should be `ui.maybePlayCutscene`


EatThePath:
If there are scripts out there which already call this with `:` is this change going to break them?

Goober5000:
> If there are scripts out there which already call this with `:` is this change going to break them?

It would, but I don't think there are actually any scripts out there with this function.  This was only added a year ago, for BtA, and BtA is only just now starting to use it (which is how the bug was discovered).  No other mods I've checked use it.

asarium (approved):

PR #4273, opened 2022-05-15
resolve FIXME for AIM_GET_BEHIND

### LINKED ISSUE #4270: FIXME for AIM_GET_BEHIND
There is an old FIXME related to the `AIM_GET_BEHIND` code:

```
// FIXME: got this from TBP and added it here to skip the Int3() but don't really want to handle it
// properly until after 3.6.7 just to avoid delaying release or breaking something - taylor
```

There is also a related @todo:
```
@todo Complete the AIM_GET_BEHIND option
```

Goober5000:
This mode was never actually implemented, even in retail, so there is no FIXME as such.  Tweak the code to reflect this.  Fixes #4270.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -13515,7 +13515,6 @@ void ai_sentrygun()
 
 /**
  * Execute behavior given by aip->mode.
- * @todo Complete the AIM_GET_BEHIND option
  */
 void ai_execute_behavior(ai_info *aip)
 {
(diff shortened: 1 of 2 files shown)


MageKing17:
Surely the FIXME was suggesting that it *should* be implemented, not that it used to be implemented and that we somehow lost it.

Goober5000:
My interpretation of @notimaginative 's comment was that the FIXME indicated further investigation was required.  Now that we know the mode was never implemented, that satisfies the investigation IMO.

code/ai/aicode.cpp
```diff
@@ -13592,6 +13591,14 @@ void ai_execute_behavior(ai_info *aip)
 			aip->mode = AIM_NONE;
 		}
 		break;
+	case AIM_GET_BEHIND:
+		if (En_objp) {
```
JohnAFernandez, line 13595: I recommend an `nprintf` indicating that this mode was triggered, so that if there is a future AI bug which requires a really deep investigation, we can spot it that this mode is being triggered, which might be contributing to it.

JohnAFernandez (approved):
One optional suggestion.

code/ai/aicode.cpp
Goober5000: Hmm.  Ok.

PR #4274, opened 2022-05-15
add Lua BitOp 1.0.2

Goober5000:
Adds the standard Lua extension library for bit operations in Lua 5.1.
http://bitop.luajit.org/

Note that this library is normally called "bit", but it is added here as "luabit" because there is already a FSO scripting library called "bit".

It would be nice if we were able to merge the two libraries so that both sets of functions are available under "bit".  This PR is in draft form until that question can be resolved.


lib/lua/CMakeLists.txt
```diff
@@ -174,6 +174,8 @@ ELSE(FSO_USE_LUAJIT)
 			lzio.h
 
 			lua.natvis
+
+			bitop/bit.c
```
asarium, line 178: This is not a standard library feature and as such must not be included in the lua code since platforms with native Lua packages will not work properly.

asarium (changes requested):
Changing the code of external libraries is a very dangerous operation since you risk loosing those changes if someone does an update of the library. Please move the new code somewhere else, outside the lua library.

Goober5000:
I don't see how the code can be moved outside the lua library given that changes to `lib/lua/linit.c` and `lib/lua/lualib.h` are required to make the code work.

As for working on other platforms... the reason for this extension is because I want to use a script that works on another platform running Lua 5.1 but not under FSO.

Here's where I found it:
http://lua-users.org/wiki/BitwiseOperators
> [[Lua BitOp]](http://bitop.luajit.org/) (5.1) - C library providing bitwise operations on fixed-sized bitfields implemented as Lua numbers. Has consistent semantics across 16, 32 and 64 bit platforms as well as IEEE 754 doubles, int32_t and int64_t lua_Number types. Also defines LuaJIT 2 bit op semantics. Generally regarded as current best-practice for bit ops in Lua. Similar interface to bitlib but newer and intended to supersede it [[2]](http://lua-users.org/lists/lua-l/2008-12/msg00287.html)[[3]](http://lua-users.org/lists/lua-l/2008-12/msg00201.html).

asarium:
> I don't see how the code can be moved outside the lua library given that changes to `lib/lua/linit.c` and `lib/lua/lualib.h` are required to make the code work.

You could replicate the functionality of `linit.c` in our code since the "library open" function can be called from any code, even our script initialization code. Please let me be clear, these changes _will_ break in some use cases so the changes in the Lua library need to be removed for this to move forward.

Goober5000:
Ok, this should work.

asarium (approved):

Issue #4013, opened 2022-02-11
Certain sexps do not cooperate with -argument operators


Goober5000:
A few sexp operators cache their values in the arguments of the sexp itself.  This causes problems when that operator is later re-evaluated for a different argument.  This is the actual root cause of the bug that led to #4010.

This happens in the following operators:

`and-in-sequence`
`is-cargo-known-delay`
`cap-subsys-cargo-known-delay`
`has-been-tagged-delay`

The `has-been-tagged-delay` operator is the best example because it is the least complicated, and also because it is almost identical to its retail version.  The implementation saves `SEXP_KNOWN_TRUE` in the parameter (not operator) nodes of the sexp so that it can retrieve that value if the sexp is evaluated again later.  These are not cleared by `when-argument` because the clearing happens on operator nodes.

I think the fix is as straightforward as removing the caching for the latter three operators, but I wanted to get additional eyes on the issue.

The caching can't be removed from `and-in-sequence` because it is intrinsic to the design of that operator.  But that can be addressed by adding a warning that it can't be used with when-argument.  Fortunately this operator is seldom used anyway.

PR #4275, opened 2022-05-15
Allow SCPUI to run `$On Briefing Stage:` hooks

naomimyselfandi:
BTA added a script that listens for briefing stages. It works fine for regular briefings, but not for command briefings. Mjn realized that SCPUI (which we use for CBs, but not regular briefings) wasn't triggering the `$On Briefing Stage:` hooks. This adds a scripting function to trigger them.

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -476,7 +476,8 @@ ADE_FUNC(runBriefingStageHook, l_UserInterface_CmdBrief, "number oldStage, numbe
 	int oldStage = -1, newStage = -1;
 	if (ade_get_args(L, "ii", &oldStage, &newStage) == 2)
 	{
-		common_fire_stage_script_hook(oldStage, newStage);
+		// Subtract 1 to convert from Lua conventions to C conventions
+		common_fire_stage_script_hook(oldStage -1, newStage -1);
 	}
 	else
 	{
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -471,5 +471,19 @@ ADE_FUNC(getBriefingMusicName, l_UserInterface_CmdBrief, nullptr, "Gets the file
 	return ade_set_args(L, "s", common_music_get_filename(SCORE_BRIEFING).c_str());
 }
 
+ADE_FUNC(runBriefingStageHook, l_UserInterface_CmdBrief, "number oldStage, number newStage", "Run $On Briefing Stage: hooks.", nullptr, nullptr)
+{
+	int oldStage = -1, newStage = -1;
+	if (ade_get_args(L, "ii", &oldStage, &newStage) == 2)
+	{
+		common_fire_stage_script_hook(oldStage, newStage);
+	}
+	else
+	{
+		LuaError(L, "Bad arguments given to ui.runBriefingStageHook!");
+	}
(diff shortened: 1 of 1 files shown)


naomimyselfandi:
@asarium, I'm sorry, it looks like I (somehow) accidentally sent you a review request...?

Goober5000 (changes requested):
The briefing stage hook is not run for command briefings, nor is there currently an `$On Command Briefing Stage` hook.  It would be better to implement that hook.

naomimyselfandi:
> The briefing stage hook is not run for command briefings

It's called in `cmd_brief_new_stage(int)` on line 364 of `missionui/missioncmdbrief.cpp`.

Goober5000:
Wow, you're right.  I ran a couple of searches but I completely missed that.

But that just moves the explanation from obvious to puzzling.  How does SCPUI run the hooks for the regular briefings?

naomimyselfandi:
> But that just moves the explanation from obvious to puzzling. How does SCPUI run the hooks for the regular briefings?

Oh, there's a (disappointingly) simple answer here. SCPUI doesn't support every screen yet, and briefings are (understandably, given how complex they are) one of the ones that aren't. You can observe this in BTA, depending on your screen resolution; the command briefing uses the whole screen, while the briefing is letterboxed.

When it does support briefings, a similar function should be provided.

Goober5000:
Ok, I think I understand now.  Since SCPUI rolls its own interface, it doesn't fire script hooks for cbrief *or* brief, but you only use SCPUI for cbrief.

This type of solution didn't sit right with me - and still doesn't - because it's a sort of brute-force invocation of another part of the code that should be part of the normal flow.  (It would be good to get @asarium 's input if possible.)

There is one more avenue that should be explored.  In `ui_system-sct.lua`, there are several `addHook` instances which appear to link up with the standard state transitions.  Have you tried adding briefing hooks in a similar way?  It's my hunch that this might be what you're looking for.

asarium:
At least in my opinion, this solution is fine for now. Sure, a generic system for invoking Lua hooks from Lua might be nice but would be overkill in this case.

naomimyselfandi:
> but you only use SCPUI for cbrief

SCPUI does not support non-command briefings at this time.

>  It's my hunch that this might be what you're looking for.

boolean addHook(string name, function() -> void hookFunction, [table conditionals (Empty table by default), function() -> boolean override_func (Function returning false by default)])
    Adds a function to be called from the specified game hook
    Returns: true if hook was installed properly, false otherwise

> This type of solution didn't sit right with me - and still doesn't - because it's a sort of brute-force invocation of another part of the code that should be part of the normal flow.

SCPUI is already outside of the normal flow. If you have another suggestion based on something more substantive than a hunch, I'd love to hear it.

Goober5000:
> boolean addHook(string name, function() -> void hookFunction, [table conditionals (Empty table by default), function() -> boolean override_func (Function returning false by default)])
Adds a function to be called from the specified game hook
Returns: true if hook was installed properly, false otherwise

Yeah, that's almost the opposite of what we need here.

> SCPUI is already outside of the normal flow. If you have another suggestion based on something more substantive than a hunch, I'd love to hear it.

Hunches can be useful; they have helped me track down some esoteric issues including #4013.

I'm satisfied with the discussion at this point, especially after getting @asarium 's feedback.  Will approve.

Goober5000 (approved):

PR #4277, opened 2022-05-16
add hu.getTargetDistance()

Goober5000:
Add a scripting function to return the distance as seen on the HUD.

### DIFF
--- a/code/scripting/api/libs/hud.cpp
+++ b/code/scripting/api/libs/hud.cpp
@@ -4,10 +4,13 @@
 #include "hud.h"
 #include "scripting/api/objs/enums.h"
 #include "scripting/api/objs/hudgauge.h"
+#include "scripting/api/objs/object.h"
+#include "scripting/api/objs/vecmath.h"
 
 #include "hud/hud.h"
 #include "hud/hudconfig.h"
 #include "hud/hudtargetbox.h"
+#include "hud/hudtarget.h"
 
 namespace scripting {
 namespace api {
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/hud.cpp
```diff
@@ -237,6 +240,29 @@ ADE_FUNC(flashTargetBox, l_HUD, "enumeration section, [number duration_in_millis
 	return ADE_RETURN_NIL;
 }
 
+ADE_FUNC(getTargetDistance, l_HUD, "object targetee, [vector targeter_pos]", "Returns the distance as displayed on the HUD, that is, the distance from a position to the bounding box of a target.  If targeter_pos is nil, the function will use the player's position.", "number", "The distance, or nil if invalid")
```
wookieejedi, line 243: Thoughts on changing this to `targeter_position` in both cases? Seems like, in general, that full name descriptions are a bit easier to read and more common in the scripting variable names.

Goober5000: Yeah that makes sense.

Goober5000: Well, in the documentation anyway.  In the code, I kept it as-is to be consistent with other usages.

wookieejedi: Ah yeah just the documentation was what I was noting.

Goober5000: Ok.  Got the other one in the documentation too.

wookieejedi (approved):

code/scripting/api/libs/hud.cpp
```diff
@@ -237,6 +240,29 @@ ADE_FUNC(flashTargetBox, l_HUD, "enumeration section, [number duration_in_millis
...
+		if (Player_obj)
+			targeter_pos = &Player_obj->pos;
+		else
+		{
+			Warning(LOCATION, "Player_obj is NULL; cannot assign position!");
+			targeter_pos = &vmd_zero_vector;
+		}
+	}
+
+	auto dist = hud_find_target_distance(targetee->objp, targeter_pos);
```
asarium, line 262: `targetee` may be invalid so in order to avoid crashes, that should be checked before calling this.

asarium (changes requested):

code/scripting/api/libs/hud.cpp
Goober5000: Good point.  Updated.

asarium: You also need to check for an invalid handle since `objp` may be invalid otherwise.

Goober5000: Missed that.  Updated again.

asarium (approved):

PR #4280, opened 2022-05-16
add improvements to aid in multi troubleshooting

notimaginative:
- report when multi.cfg is missing
 - add game version, multi version, cmdline, and PXO status
   to multi.log
 - use non-interactive mode by default on standalone and
   enable it as soon as possible

### DIFF
--- a/code/network/multi_log.cpp
+++ b/code/network/multi_log.cpp
@@ -15,6 +15,10 @@
 #include "parse/generic_log.h"
 #include "cfile/cfile.h"
 #include "parse/parselo.h"
+#include "globalincs/version.h"
+#include "network/multi_options.h"
+#include "network/multi_fstracker.h"
+#include "network/multi.h"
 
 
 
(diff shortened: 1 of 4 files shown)


code/network/multi_log.cpp
```diff
@@ -77,11 +81,36 @@ void multi_log_write_update()
 	ml_printf("Server has been active for %d hours, %d minutes, and %d seconds", hours, mins, seconds);
 }
 
+// write out some info helpful for debugging
+static void multi_log_write_info()
+{
+	extern bool Multi_cfg_missing;
+
+	if (Multi_cfg_missing) {
+		ml_string("**  multi.cfg is missing!  **");
```
JohnAFernandez, line 90: Oh, so `ml_string` does a `\n` by default?

notimaginative: Yep, the base generic logging function puts an EOL on every line by default. So both `ml_printf()` and `ml_string()` get `\n` automatically.

PR #4282, opened 2022-05-17
Replace +Orders Accepted: Bitfield with flag list

BMagnu:
As detailed in the LuaAI PR (#4198), the orders a ship is allowed to follow in mission are currently saved as a bitfield. This bitfield is only accurate if the orders are always within the same order however. And since LuaAI allows adding such orders with modular tables, this is now very brittle and prone to failure and to breaking missions if AI stuff is tweaked by modders.
Hence this PR removes the bitfield and replaces it with a flagset where the parse names of all allowed orders are explicitly listed

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -3347,6 +3347,7 @@ int parse_object(mission *pm, int  /*flag*/, p_object *p_objp)
 
 	// check for the optional "orders accepted" string which contains the orders from the default
     // set that this ship will actually listen to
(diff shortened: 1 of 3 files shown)


Goober5000 (changes requested):
This should be capable of saving the numeric bitfield if the save format is retail - `Mission_save_format == FSO_FORMAT_RETAIL` for FRED and `save_format == MissionFormat::RETAIL` for qtFRED

Goober5000 (approved):

PR #4283, opened 2022-05-18
Fix HUD squad menu PageDown

BMagnu:
In order to facilitate more player orders being added with LuaAI, the comms menu needs to support more than the ten entries available.
Fortunately, a system to effectively scroll in that menu already exists.
Unfortunately, it was broken except for one or two modes.
This fixes this by doing three things:

1. Make sure that ``First_menu_item``, the tracker for the current scroll position, is only reset on mode change (as it was before), and _not_ in any on-frame method
2. Make sure that the key-check method always takes ``First_menu_item`` into consideration for which button is selectable
3. Make sure that the key-check method always takes ``First_menu_item`` into consideration for which result was actually selected instead of relying on the callers to add this offset.

This now makes scrolling in the menu possible and working properly for each menu mode, including the main menu, the order menu, and the ship menu.

### DIFF
--- a/code/hud/hudsquadmsg.cpp
+++ b/code/hud/hudsquadmsg.cpp
@@ -587,11 +587,11 @@ int hud_squadmsg_get_key()
 				return i;
 
 			// play general fail sound if inactive item hit.
-			else if ( (i < Num_menu_items) && !(MsgItems[i].active) )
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
Looks like a good fix/update.

PR #4284, opened 2022-05-18
use SCP_string for constructing multi/event log lines

notimaginative:


JohnAFernandez:
I'm surprised that clang isn't pinging the `NULL`'s in the surrounding lines, but I guess that just makes it easier on us.

MageKing17 (approved):

PR #4286, opened 2022-05-19
add timestamp_since

Goober5000:
Add `timestamp_since()` and modify a few places that use `timestamp_until()`.

### DIFF
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -238,8 +238,10 @@ int timestamp_until(int stamp)
 
 int timestamp_until(TIMESTAMP stamp)
 {
+	Assertion(stamp.isValid() && !stamp.isNever(), "timestamp_until was called with an Invalid or Never timestamp!");
 	if (!stamp.isValid() || stamp.isNever())
 		return INT_MAX;
+
 	if (stamp.isImmediate())
 		return 0;
 
@@ -248,8 +250,10 @@ int timestamp_until(TIMESTAMP stamp)
 
 int ui_timestamp_until(UI_TIMESTAMP stamp)
 {
+	Assertion(stamp.isValid() && !stamp.isNever(), "timestamp_until was called with an Invalid or Never timestamp!");
 	if (!stamp.isValid() || stamp.isNever())
 		return INT_MAX;
+
 	if (stamp.isImmediate())
 		return 0;
 
@@ -263,8 +267,10 @@ int timestamp_since(int stamp)
 
 int timestamp_since(TIMESTAMP stamp)
 {
+	Assertion(stamp.isValid() && !stamp.isNever(), "timestamp_since was called with an Invalid or Never timestamp!");
 	if (!stamp.isValid() || stamp.isNever())
 		return INT_MIN;
+
 	if (stamp.isImmediate())
 		return 0;
 
@@ -273,8 +279,10 @@ int timestamp_since(TIMESTAMP stamp)
 
 int ui_timestamp_since(UI_TIMESTAMP stamp)
 {
+	Assertion(stamp.isValid() && !stamp.isNever(), "timestamp_since was called with an Invalid or Never timestamp!");
 	if (!stamp.isValid() || stamp.isNever())
 		return INT_MIN;
+
 	if (stamp.isImmediate())
 		return 0;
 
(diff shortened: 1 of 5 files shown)


code/io/timer.cpp
```diff
@@ -256,6 +256,31 @@ int ui_timestamp_until(UI_TIMESTAMP stamp)
...
 
+int timestamp_since(int stamp)
+{
+	return timestamp_ms() - stamp;
+}
+
+int timestamp_since(TIMESTAMP stamp)
+{
+	if (!stamp.isValid() || stamp.isNever())
+		return INT_MIN;
```
JohnAFernandez, line 267: I'm too tried to properly think about this, but I don't think silently failing to INT_MIN is the right approach here, it could lead to unpredictable side-effects.  Since it's brand new, maybe we should Assert that the timestamp is valid and not Never.  Anyone using it that is not sure that their timestamp is Valid and not never, can check for it, and implement their own work-around for it.

Goober5000: I did this to be consistent with the existing `timestamp_until`, which returns `INT_MAX` for an Invalid or Never timestamp.  It's a graceful way to devolve to a value that makes sense without throwing an error.  And anyone calling the function can check for isValid or isNever just like usual.

JohnAFernandez: I anticipate it will not be the graceful.  At the very least we need better comments for the functions that explains what the special cases will do for isValid and isNever, because I strongly believe that this is not intuitive behavior, even if it is consistent behavior 

INT_MAX in that case basically means, "will never happen".  Here INT_MIN does not mean "has never happened", it means "happened at the beginning of time".

JohnAFernandez (changes requested):
See my last comment.

code/io/timer.cpp
JohnAFernandez: Wait, no it doesn't.  It means "will never happen" because it's negative, but that's still not what you would want from this function anyway..

Goober5000: Consider the existing case of `timestamp_until`.  If you have an "invalid" or "never" timestamp, the amount of time until it times out is infinite.  This can be represented by `INT_MAX` which is infinite for all practical purposes in FSO.  This will work in timestamp expressions, and it also makes sense in timestamp comparisons or when sorting a list of timestamps.

Applying this to `timestamp_since`, the amount of time *since* an "invalid" or "never" timestamp has timed out is negative, placing it in the future.  Similarly `INT_MIN` is negative infinity for all practical purposes in FSO.

That's the graceful behavior that allows these timestamp states to be consistent with "normal" timestamps.  It's a mathematical convenience in the same way that zero factorial is equal to 1.

JohnAFernandez: Except it's not graceful for `timestamp_since`.  I already understand everything you just explained.  The problem is that the use case of `timestamp_since` is different than the use case for `timestamp_until`.  I can set `timestamp_until` to infinity (mostly) without a problem because it's tricking FSO into waiting forever. (although I don't think this is the correct behavior, honestly, and it should be documented, too) 

If I don't check the output value, it will lead to a crash.  And nothing in the documentation suggests that I cannot trust the output value. I have to read the actual function to know that it is an issue.  And no case of `timestamp_until` really checks the output, so I will probably just use `timestamp_since`.  What's my use case?  figuring out the time differences between server and client.  

What's that, it's crashing every five minutes on the interpolation calculations because we decided to use `INT_MIN` for the fail value, and that causes extremely bad and hard to diagnose physics problems that are compounded by it being a multiplayer use case?

I don't expect to change your mind, but you need to add some comments to the headers, so that no one makes that mistake.

Goober5000: Actually you raise a really good point.  My use case is when you intentionally have valid and invalid timestamps and you need to compare them.  But if a timestamp is accidentally invalid, then the situation becomes a lot more tricky to deal with -- and as you said, this is heavily compounded when it's over multiplayer.  And furthermore, my design is preserving flexibility for future use, which is theoretical, but it's running up against a real-world problem, which is practical.

So let reason be silent when experience gainsays its conclusions.  I'll add some Assertions as well as that documentation.

JohnAFernandez (approved):

code/io/timer.cpp
JohnAFernandez: Thank you.

PR #4289, opened 2022-05-20
briefing icon behavior tweaks per request

Goober5000:
Per #3860, don't disable any of the icon dropdowns, and don't change the ship class when the icon type is a special popup.

### DIFF
--- a/fred2/briefingeditordlg.cpp
+++ b/fred2/briefingeditordlg.cpp
@@ -329,27 +329,6 @@ void briefing_editor_dlg::restore_editor_state()
 	m_cur_icon = icon_saved;
 }
 
-int briefing_editor_dlg::get_first_ship_with_no_custom_icon()
-{
-	if (first_ship_with_no_custom_icon < 0)
-	{
-		for (int i = 0; i < ship_info_size(); ++i)
-		{
-			if (Ship_info[i].bii_index_ship < 0)
-			{
-				first_ship_with_no_custom_icon = i;
-				break;
-			}
-		}
-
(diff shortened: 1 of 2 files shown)


fred2/briefingeditordlg.cpp
```diff
@@ -521,11 +521,6 @@ void briefing_editor_dlg::update_data(int update)
 			}
 			ptr->icons[m_last_icon].type = m_icon_image;
 
-			// if this icon has a special closeup image (like a jump node)
-			// then make sure the ship class doesn't have a special icon
-			if (brief_special_closeup(m_icon_image))
-				m_ship_type = get_first_ship_with_no_custom_icon();
```
jg18, line 527: Do we still need this function, or can we get rid of it?

jg18 (commented):
Looks good overall, just one question.

fred2/briefingeditordlg.cpp
Goober5000: We still need it.  It's still used in FSO, and it could be useful for other purposes in the future.

jg18: Okay about potentially being useful in the future, but AFAICT `get_first_ship_with_no_custom_icon()` a protected function of this dialog class and is called just here.

jg18 (approved):

fred2/briefingeditordlg.cpp
Goober5000: Ah, I thought you meant `brief_special_closeup`.

PR #4290, opened 2022-05-20
Make number of player orders FRED can handle dynamic

BMagnu:
This enables to use more than 10 player orders for a single object type in FRED, which is important when more player orders are added using LuaAI.
QtFRED seems to already be dynamic, so no changes are needed for it.


fred2/ignoreordersdlg.h
```diff
@@ -30,8 +30,8 @@ class ignore_orders_dlg : public CDialog
 	ignore_orders_dlg(CWnd* pParent = NULL);   // standard constructor
 
 	ship *m_shipp;
-	check_box_info check_boxes[MAX_CHECKBOXES];
-	int m_num_checks_active;
+	SCP_vector<size_t> m_orderList;
+	std::unique_ptr<CCheckListBox> m_checklistbox;
```
jg18, line 34: Is it possible to declare `m_checklistbox` as part of the object *without* using dynamic memory allocation? I think that could make the changes a good bit simpler.

Take a look at how `CAddModifyContainerDlg` does this with `m_container_data_lister`.

fred2/ignoreordersdlg.h
```diff
@@ -30,8 +30,8 @@ class ignore_orders_dlg : public CDialog
 	ignore_orders_dlg(CWnd* pParent = NULL);   // standard constructor
 
 	ship *m_shipp;
-	check_box_info check_boxes[MAX_CHECKBOXES];
-	int m_num_checks_active;
+	SCP_vector<size_t> m_orderList;
```
jg18, line 33: Why `size_t`?

I would think that type would be used only to represent a "size" of some kind.

fred2/ignoreordersdlg.h
jg18: Also, a name for this variable that indicates what the check boxes mean, as `m_container_data_lister` does,, would be really nice.

fred2/ignoreordersdlg.cpp
```diff
@@ -101,53 +87,24 @@ BOOL ignore_orders_dlg::OnInitDialog()
...
-
-		// shrink the size of the parent window by the size of the checkbox
-		if ( last_bottom != 0 )
-			window_size.bottom -= (last_bottom - check_size.bottom );
+	for (size_t order_id : default_orders){
+		int checkboxNr = (int) m_orderList.size();
 
-		last_bottom = check_size.bottom;
+		list->AddString(Player_orders[order_id].localized_name.c_str());
+		list->SetCheck(checkboxNr, BST_UNCHECKED);
```
jg18, line 97: Isn't this index out of range?

Also using `Nr` to mean "number" seems unintuitive to me. That convention isn't used elsewhere in the codebase.

fred2/ignoreordersdlg.cpp
```diff
@@ -101,53 +87,24 @@ BOOL ignore_orders_dlg::OnInitDialog()
...
-		if (m_num_checks_active >= MAX_CHECKBOXES)
-		{
-			Warning(LOCATION, "Tried to add more orders than FRED2 has checkboxes for them (can be max %d)!", MAX_CHECKBOXES);
-			break;
-		}
-
-		check_boxes[m_num_checks_active].button->SetWindowText(Player_orders[order_id].localized_name.c_str());
-		check_boxes[m_num_checks_active].id = (int)order_id;
-		m_num_checks_active++;
+	CCheckListBox* list = ((CCheckListBox*)GetDlgItem(IDC_IGNORE_ORDERS_LIST));
```
jg18, line 90: Optional: A more descriptive name than `list`, maybe `ignore_orders_list`, would be nice, but it's not super important.

jg18 (changes requested):
Some requests after a first pass, but I like the idea overall. It's telling that the PR deletes a lot more lines than it adds. :+1:

fred2/ignoreordersdlg.h
BMagnu: Cause its the size of the vector.
Since all commands now represent indices of a vector, a size_t is appropriate. Also, it's already used as command in a fair few other places, so this is to keep consistent

fred2/ignoreordersdlg.cpp
```diff
@@ -101,53 +86,23 @@ BOOL ignore_orders_dlg::OnInitDialog()
...
-		if ( last_bottom != 0 )
-			window_size.bottom -= (last_bottom - check_size.bottom );
+	for (size_t order_id : default_orders){
+		int checkboxNo = (int) m_orderList.size();
 
-		last_bottom = check_size.bottom;
+		ignore_orders_list->AddString(Player_orders[order_id].localized_name.c_str());
+		ignore_orders_list->SetCheck(checkboxNo, BST_UNCHECKED);
 
+		m_orderList.push_back(order_id);
```
jg18, line 97: SCP is generally moving away from the old functions `push_back()`  and `insert()` towards the C++11 functions `emplace_back()` and `emplace()` , unless the old version is needed in some situation. However, it's less important with primitive types like `size_t`.

fred2/ignoreordersdlg.cpp
```diff
@@ -199,31 +154,32 @@ void ignore_orders_dlg::OnOK()
...
 
+	CCheckListBox* ignore_orders_list = ((CCheckListBox*)GetDlgItem(IDC_IGNORE_ORDERS_LIST));
+
 	// clear out the orders, then set the bits according to which check boxes are set
 	if ( m_ship >= 0 ) {
-		for ( i = 0; i < m_num_checks_active; i++ ) {
-			if ( check_boxes[i].button->GetCheck() )
-				orders_accepted.insert(check_boxes[i].id);
+		for ( i = 0; i < (int) m_orderList.size(); i++) {
+			if (ignore_orders_list->GetCheck(i) == BST_CHECKED)
```
jg18, line 162: Wouldn't ` != BST_UNCHECKED` be more correct, like you have above?

jg18 (changes requested):
A few more requests, some optional.

fred2/ignoreordersdlg.cpp
```diff
@@ -62,24 +53,18 @@ END_MESSAGE_MAP()
...
-	check_boxes[2].button = (CButton *)GetDlgItem( IDC_CHECK3 );
-	check_boxes[3].button = (CButton *)GetDlgItem( IDC_CHECK4 );
-	check_boxes[4].button = (CButton *)GetDlgItem( IDC_CHECK5 );
-	check_boxes[5].button = (CButton *)GetDlgItem( IDC_CHECK6 );
-	check_boxes[6].button = (CButton *)GetDlgItem( IDC_CHECK7 );
-	check_boxes[7].button = (CButton *)GetDlgItem( IDC_CHECK8 );
-	check_boxes[8].button = (CButton *)GetDlgItem( IDC_CHECK9 );
-	check_boxes[9].button = (CButton *)GetDlgItem( IDC_CHECK10 );
+
+	//Because Windows, this needs to be manually created in addition to having the resource thingy.
```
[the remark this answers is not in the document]
BMagnu: Yes. It needs to be changed slightly (from "manually created" to something like "manually subclassed"), but overall its still relevant due to some MFC annoyances

fred2/ignoreordersdlg.cpp
```diff
@@ -199,31 +154,32 @@ void ignore_orders_dlg::OnOK()
...
-				orders_accepted.insert(check_boxes[i].id);
+		for ( i = 0; i < (int) m_orderList.size(); i++) {
+			if (ignore_orders_list->GetCheck(i) == BST_CHECKED)
+				orders_accepted.insert(m_orderList[i]);
 		}
 		Ships[m_ship].orders_accepted = orders_accepted;
 	} else {
 		for ( objp = GET_FIRST(&obj_used_list); objp != END_OF_LIST(&obj_used_list); objp = GET_NEXT(objp) ) {
 			if (((objp->type == OBJ_SHIP) || (objp->type == OBJ_START)) && (objp->flags[Object::Object_Flags::Marked])) {
-				Ships[objp->instance].orders_accepted.clear();
```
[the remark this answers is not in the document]
BMagnu: The previous handling was simply incorrect here. It would clear all allowed orders, and then remove orders explicitly disallowed (from the empty set, thus doing nothing) and readding explicitly allowed entries. That leaves the indeterminate-state orders also cleared. This is wrong, however, as we want orders for which the checkboxes are indeterminate to stay as they are in their respective ships (which may or may not be allowed). To do that, we don't clear the orders and leave them as is, then make sure they're up to date by removing disallowed orders and readding allowed orders.

fred2/ignoreordersdlg.cpp
```diff
@@ -232,113 +188,12 @@ void ignore_orders_dlg::OnOK()
...
-	CButton *button;
-
-	button = (CButton *)GetDlgItem(IDC_CHECK10);
-	if (button->GetCheck() == 1)
-		button->SetCheck(0);
-	else
-		button->SetCheck(1);
-}
+	int i = ignore_orders_list->GetCurSel();
+	if (ignore_orders_list->GetCheck(i) == BST_INDETERMINATE)
```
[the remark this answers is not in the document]
BMagnu: Not really handle, as that function is only called when a checkbox changed, which means the user must have selected one. However, Asserting that ``button`` is not LB_ERR might be fine.

fred2/ignoreordersdlg.cpp
BMagnu: In that path it does not matter. It _should_ be impossible for a checkbox to be indeterminate there.

jg18 (approved):
Looks good. Thanks for ~~putting up with~~ err, addressing/replying to my comments. :wink:

PR #4291, opened 2022-05-21
Add copy-container SEXP

jg18:
Per suggestion from @MjnMixael in #fred.


code/parse/sexp_container.cpp
```diff
@@ -1311,6 +1320,68 @@ void sexp_clear_container(int node)
...
+	const char *dest_container_name = CTEXT(node);
+	auto *p_dest_container = get_sexp_container(src_container_name);
+
+	if (!p_dest_container) {
+		report_nonexistent_container("Copy-container", dest_container_name);
+		return;
+	}
+	auto &dest_container = *p_dest_container;
+
+	if (src_container.get_non_persistent_type() != dest_container.get_non_persistent_type()) {
```
JohnAFernandez, line 1348: Seems pretty straight forward.  The only thing is the name of this function `get_non_persistent_type`.  That just means what type of container type it is?

jg18: Yeah, get a container's type (function return type is `enum class ContainerType`) with persistence flags zeroed out.

See the SEXP help text and also the use case described in #fred for why this is done.

jg18: Admittedly `get_non_persistent_Type()` wasn't the best function name. 🤷‍♂️

JohnAFernandez: Go figure

JohnAFernandez (approved):
To my "sexp-amateur" eyes, it looks fine.

jg18:
The latest version passes manual tests by @MjnMixael.

JohnAFernandez (approved):

PR #4293, opened 2022-05-21
fix some memory management problems in the for-* operators

Goober5000:
1) In `for-counter`, when `just_count` is true there is a memory leak, since the code never enters `test_argument_vector_for_condition`, but for the same reason, duplicating the string is not necessary in that case, so just add a blank string
2) Remove unnecessary `vm_strdup` calls from the other for-* operators


code/parse/sexp.cpp
```diff
@@ -10455,14 +10455,21 @@ int eval_for_counter(int arg_handler_node, int condition_node, bool just_count =
...
-		// Note: we do not call vm_free() on the contents of argument_vector, and we don't for the very good
-		// reason that those pointers are then passed along by test_argument_vector_for_condition(), below.
-		// The strings will then be freed by arg_item::expunge() or arg_item::clear_nesting_level(), or even
-		// inside test_argument_vector_for_condition() (if the argument doesn't satisfy the condition). So
-		// under no circumstances try to free these strings inside this function! It will cause a double-free
-		// situation, resulting in a crash at best. -MageKing17
+		// if we're just counting we should just avoid the memory management problem mentioned below,
+		// since we return before calling test_argument_vector_for_condition
+		if (just_count)
+			argument_vector.emplace_back("", -1);
```
jg18, line 10461: Won't this call result in dangling pointers?

Where are the empty strings we're adding here being stored?

code/parse/sexp.cpp
```diff
@@ -10539,7 +10546,7 @@ int eval_for_ship_collection(int arg_handler_node, int condition_node, int op_co
 				}
 
 				if (constraint_index == ship_index)
-					argument_vector.emplace_back(vm_strdup(shipp->ship_name), -1);
+					argument_vector.emplace_back(shipp->ship_name, -1);
```
jg18, line 10549: Why were these `vm_strdup()` calls removed?

Won't these changes result in attempts to free memory we shouldn't be trying to free?

jg18 (changes requested):
I like the idea of using `const char *` when possible but have some concerns.

FWIW I was thinking of a narrower change, something like this:
```
if (just_count) {
    /* [clean up the allocated strings in argument_vector] */

    return static_cast<int>(argument_vector.size());
}
```

This approach would keep the changes tightly localized.

code/parse/sexp.cpp
Goober5000: Strings that are hard-coded do not end up as dangling pointers.  They are presumably kept in persistent area of the code.

code/parse/sexp.cpp
Goober5000: Hmm, yeah, that would be problematic.  Might need to figure out a way around that.

Goober5000:
> FWIW I was thinking of a narrower change

That would work, but I thought it made sense to hit the other areas at the same time.

jg18 (approved):
Looks good. A much better solution than my original idea.

I also like all the extra `const`s and think having the extra `const_cast`s is a worthwhile trade-off. :+1:

PR #4296, opened 2022-05-21
Add for-* SEXPs for containers

jg18:
Add the following SEXPs:
- `for-container-data`
- `for-map-container-keys`


jg18:
The PR is complete, but I'm not sure if it fulfills the expected semantics.

code/parse/sexp.cpp
```diff
@@ -10561,6 +10566,62 @@ int eval_for_ship_collection(int arg_handler_node, int condition_node, int op_co
...
+		&num_false,
+		&num_known_true,
+		&num_known_false);
+
+	// use the sexp_or algorithm
+	if (num_known_true || num_true)
+		return SEXP_TRUE;
+	// TODO: is the check below correct?
+	else if (num_valid_arguments > 0 && num_known_false == num_valid_arguments)
+		return SEXP_KNOWN_FALSE;
```
jg18, line 10620: Is this `else-if` clause correct, in terms of fulfilling expected behavior? Or should it be commented out, like in `eval_for_ship_collection()`?

BMagnu: I am not sure, but I don't think it is correct for the container stuff to return ``SEXP_KNOWN_FALSE``, because (if I understood correctly) it is possible that an event might _add_ a value to a container. Which in turn means that an event that evaluated ``KNOWN_FALSE`` for all valid arguments (and thus would flag here as ``KNOWN_FALSE`` as well) is stopped being evaluated, even with the new value added. To prevent that we must only return ``SEXP_FALSE``, so that it is reevaluated when new items are added.

jg18:
Passes tests by MjnMixael using the attached test mission.
[test_4296.fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/8768183/test_4296.fs2.txt)

code/parse/sexp_container.cpp
```diff
@@ -1425,3 +1425,89 @@ int sexp_container_eval_change_sexp(int op_num, int node)
...
+	SCP_vector<std::pair<const char*, int>> &argument_vector,
+	bool just_count,
+	const char *sexp_name,
+	bool map_keys_only)
+{
+	Assertion(node != -1, "%s wasn't given any ontainers. Please report!", sexp_name);
+
+	// filter out duplicates, but keys/data are case-sensitive
+	// unordered is also ok, because the for-* SEXPs are "any-of"
+	SCP_unordered_set<SCP_string> arguments;
```
BMagnu, line 1439: Why is this a set?
As a user, I'd assume that a duplicate entry in a container will also be evaluated twice. It might lead to unintuitive behaviour if it is silently filtered.
While I personally would like both in-order evaluation and duplicate-including evaluation, there is a case to be made that a user usually doesn't wants this kind of duplicate evaluation (though I could think of a few cases where it would be nice to have). But if you end up wanting to filter the duplicates here, it needs to be clearly documented in the SEXP help text (and while you're on it, lso specifying that it does not adhere to the order of the data in the container would be nice), or some user will scratch their head later on trying to figure this out.

BMagnu (changes requested):

code/parse/sexp_container.cpp
jg18: Hmm, I think I'd misinterpreted the `already-dupped` function argument in `test_argument_vector_for_condition()` to mean it was doing de-duplication of the argument list, but I found out later it was referring to memory management.

I'll get rid of this.

As a bonus, it'll simplify the code some.

Issue #4298, opened 2022-05-22
FR: Support using containers as options for when-argument SEXPs


jg18:
Allow specifying a list container of strings among options to `when-argument`.

jg18:
This feature was completed by the PRs mentioned above.

PR #4299, opened 2022-05-22
Add container name subtype to Sexp_nodes

jg18:
Use special syntax for specifying container names as SEXP arguments, instead of specifying them as ordinary strings.

The new syntax is `&&ContainerName` instead of `"ContainerName"`.

Required for Issue #4298. Otherwise, FSO/FRED can't tell that an option for `when-argument` is a container and not a string literal.

Container names can no longer include spaces.

As bonuses, syntax checking for container names as SEXP arguments is now much stricter,  and `update_sexp_containers()` now updates all renamed containers correctly.

**NOTE**: This introduces ***breaking changes*** for containers, but they can easily be fixed in a text editor.


BMagnu (approved):
While I am pensive about this kind of breaking change, it might be fine in this instance since the containers haven't yet seen a stable.
Otherwise this seems good.

PR #4301, opened 2022-05-23
Fix FRED Add/Replace

### LINKED ISSUE #4295: Paste(add) and paste(overright) FRED menu conditions flipped
In situations where one of these should be available and the other not, the wrong one is greyed out. This is recent behavior, I believe caused by #4230 but I haven't conclusively bisected or reverted.

The relevant code appears to be in sexp_tree.cpp, and is likely to be a matter of just not having flipped the names when the action names were switched. Building fred is giving me trouble so I can't really attack it myself. Care will also have to be taken to make sure the containers code added since is using the right ones, I think.

If anyone is having this issue, it appears to be a viable work-around to use the hotkeys instead of the menu options, they seem to work fine. 

[test_paste.zip](https://github.com/scp-fs2open/fs2open.github.com/files/8748303/test_paste.zip)
To reproduce, take the attached mission, copy the `when` statement from EV2, and then right click on the `do-nothing` statement in EV1. In older builds you will have only the paste-over available and in new builds you will have only the paste-add option.

Also, if we're not stuck with them for some reason, PASTE and PASTE_SPECIAL are really a terrible pair of internal names whichever way they're assigned. IMO the code would be much more readable if it was PASTE_INTO and PASTE_OVER or something similarly explicit.

BMagnu:
Fixes #4295 caused by #4230

### DIFF
--- a/fred2/fred.rc
+++ b/fred2/fred.rc
@@ -454,7 +454,7 @@ BEGIN
         MENUITEM SEPARATOR
         MENUITEM "Cut\tCtrl+X",                 ID_EDIT_CUT, GRAYED
         MENUITEM "Copy\tCtrl+C",                ID_EDIT_COPY
-        MENUITEM "Paste (Overwrite)\tCtrl+Shift+V",   ID_EDIT_PASTE_SPECIAL, GRAYED
+        MENUITEM "Paste (Overwrite)\tCtrl+Shift+V",   ID_EDIT_PASTE, GRAYED
         MENUITEM SEPARATOR
         POPUP "Add Operator"
         BEGIN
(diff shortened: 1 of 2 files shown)


Goober5000 (changes requested):
Since `ID_EDIT_PASTE` and `ID_EDIT_PASTE_SPECIAL` were swapped, the better strategy is to swap them in this function.  Swapping `add_type` and `replace_type` instead is likely to lead to errors or missed examples.  (Of course, the better option is to un-swap the two defined tokens and instead just swap the keyboard shortcuts.)

Goober5000 (approved):

PR #4302, opened 2022-05-23
Crossing the specular rubicon

EatThePath:
as discussed in https://www.hard-light.net/forums/index.php?topic=98195.0 this PR replaces the command line args that scale lighting with whole-light controls, with controls in the lighting profiles to globally tune light types taking over for mods.

Effectively feature complete now but submitted in draft as there's a lot of cleanup and documentation I need to do still.

Includes changes from #4208, will have a history cleanup once that's merged in.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -318,17 +318,12 @@ cmdline_parm clip_dist_arg("-clipdist", "Changes the distance from the viewpoint
 cmdline_parm ambient_power_arg("-ambient", "Multiplies the brightness of all ambient light", AT_FLOAT);
 cmdline_parm light_power_arg("-light", "Multiplies the brightness of all light", AT_FLOAT);
 cmdline_parm emissive_power_arg("-emissive", "Multiplies the brightness of all ambient light", AT_FLOAT);
-
-cmdline_parm spec_static_arg("-spec_static", "Adjusts suns contribution to specular highlights", AT_FLOAT); //TODO: move to deprecated
-cmdline_parm spec_point_arg("-spec_point", "Adjusts laser weapons contribution to specular highlights", AT_FLOAT); //TODO: move to deprecated
-cmdline_parm spec_tube_arg("-spec_tube", "Adjusts beam weapons contribution to specular highlights", AT_FLOAT); //TODO: move to deprecated
-cmdline_parm ambient_factor_arg("-ambient_factor", "Adjusts ambient light applied to all parts of a ship", AT_INT);		// Cmdline_ambient_factor TODO: move to deprecated
+cmdline_parm emissive_arg("-emissive_light", "Enable emissive light from ships", AT_NONE);		// semi-deprecated but still functional
 cmdline_parm env("-noenv", NULL, AT_NONE);								// Cmdline_env
 cmdline_parm glow_arg("-noglow", NULL, AT_NONE); 						// Cmdline_glow  -- use Bobs glow code
 cmdline_parm nomotiondebris_arg("-nomotiondebris", NULL, AT_NONE);		// Cmdline_nomotiondebris  -- Removes those ugly floating rocks -C
 cmdline_parm noscalevid_arg("-noscalevid", NULL, AT_NONE);				// Cmdline_noscalevid  -- disable video scaling that fits to window
 cmdline_parm spec_arg("-nospec", NULL, AT_NONE);			// Cmdline_spec  --
-cmdline_parm emissive_arg("-emissive_light", "Enable emissive light from ships", AT_NONE);		// Cmdline_no_emissive  -- don't use emissive light in OGL //TODO: move to deprecated
 cmdline_parm normal_arg("-nonormal", NULL, AT_NONE);						// Cmdline_normal  -- disable normal mapping
 cmdline_parm height_arg("-noheight", NULL, AT_NONE);						// Cmdline_height  -- enable support for parallax mapping
 cmdline_parm enable_3d_shockwave_arg("-3dshockwave", NULL, AT_NONE);
@@ -350,7 +345,6 @@ cmdline_parm no_deferred_lighting_arg("-no_deferred", NULL, AT_NONE);	// Cmdline
 cmdline_parm anisotropy_level_arg("-anisotropic_filter", NULL, AT_INT);
 
 float Cmdline_clip_dist = Default_min_draw_distance;
-int Cmdline_ambient_factor = 128;
 float Cmdline_ambient_power = 1.0f;
 float Cmdline_emissive_power = 0.0f;
 float Cmdline_light_power = 1.0f;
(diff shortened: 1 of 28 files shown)


EatThePath:
I'm leaving this in draft until weapon lights is merged but it is now feature complete save for qaz's shader work

EatThePath:
Still pending a few final changes and qaz's contributions, but I'm putting it to submitted now to see if it passes checks,

z64555:
Postponing after 22.2

code/cmdline/cmdline.cpp
```diff
@@ -1997,8 +1998,8 @@ bool SetCmdlineParams()
 		Cmdline_no_fbo = 1;
 	}
 
-	if ( emissive_arg.found() ) {
-		Cmdline_emissive = 1;
+	if ( emissive_arg.found() && !emissive_power_arg.found()){
+		Cmdline_emissive_power = 0.30f;
```
BMagnu, line 2002: Every time this condition can be true, it will have already set the value to 0.3 in Line 1848, so this if can just be deleted.

code/lab/dialogs/render_options.cpp
```diff
@@ -289,15 +294,20 @@ void RenderOptions::open(Button* /*caller*/) {
...
+	auto ambient_sldr = new Slider("Ambient Factor", 0.0f, 10.0f, 0, y + 2, set_ambient_factor, dialogWindow->GetWidth());
+	ambient_sldr->SetSliderValue(lighting_profile::lab_get_ambient());
 	dialogWindow->AddChild(ambient_sldr);
 	y += ambient_sldr->GetHeight() + 2;
 
-	auto direct_sldr = new Slider("Direct. Lights", 0.0f, 2.0f, 0, y + 2, set_static_light_factor, dialogWindow->GetWidth());
-	direct_sldr->SetSliderValue(static_light_factor);
-	dialogWindow->AddChild(direct_sldr);
-	y += direct_sldr->GetHeight() + 2;
+	auto emissive_sldr = new Slider("Emissive Amount", 0, 10.0f, 0, y + 2, set_emissive, dialogWindow->GetWidth());
```
BMagnu, line 307: Amount? Is this a factor? Is this absolute brightness? Can this be given a more concise name, or is this a somewhat arbitrary value?

code/lighting/lighting.cpp
```diff
@@ -177,12 +171,11 @@ void light_add_directional(const vec3d *dir, float intensity, float r, float g,
...
-	l.spec_r = spec_r;
-	l.spec_g = spec_g;
-	l.spec_b = spec_b;
 
-	//Direcional lights not yet switched to use lighting profiles multipliers as that's part of a seperate project.
-	l.intensity = intensity;
+	//Direcional lights not yet switched to use lighting profiles multipliers as that's part of a seperate project
+	auto lp = lighting_profile::current();
+	l.intensity = lp->directional_light_brightness.handle(intensity);
+	l.intensity = lp->overall_brightness.handle(l.intensity);
```
BMagnu, line 178: You assign intensity twice here. That seems not really correct, right?
Also, since this now _does_ use some lighting profile stuff, is the comment still accurate?

code/lighting/lighting.cpp
```diff
@@ -209,27 +202,19 @@ void light_add_point(const vec3d *pos, float r1, float r2, float intensity, floa
...
 	l.g = g;
 	l.b = b;
-	l.spec_r = spec_r;
-	l.spec_g = spec_g;
-	l.spec_b = spec_b;
 
 	//configurable global tuning of light qualities
 	auto lp = lighting_profile::current();
 	l.intensity = lp->point_light_brightness.handle(intensity);
+	l.intensity = lp->overall_brightness.handle(l.intensity);
```
BMagnu, line 217: Same here, you just reassign intensity. Either the first assignment can go, or they need to be combined somehow, or the second assignment is bad

code/lighting/lighting.cpp
```diff
@@ -270,13 +249,11 @@ void light_add_tube(const vec3d *p0, const vec3d *p1, float r1, float r2, float
...
 	l.g = g;
 	l.b = b;
-	l.spec_r = spec_r;
-	l.spec_g = spec_g;
-	l.spec_b = spec_b;
 
 	//configurable global tuning of light qualities
 	auto lp = lighting_profile::current();
 	l.intensity = lp->tube_light_brightness.handle(intensity);
+	l.intensity = lp->overall_brightness.handle(l.intensity);
```
BMagnu, line 256: Same here

code/lighting/lighting.cpp
```diff
@@ -488,12 +455,10 @@ void light_add_cone(const vec3d *pos, const vec3d *dir, float angle, float inner
 	l.r = r;
 	l.g = g;
 	l.b = b;
-	l.spec_r = spec_r;
-	l.spec_g = spec_g;
-	l.spec_b = spec_b;
 
 	auto lp = lighting_profile::current();
 	l.intensity = lp->cone_light_brightness.handle(intensity);
+	l.intensity = lp->overall_brightness.handle(l.intensity);
```
BMagnu, line 461: Same here

code/missionui/missionscreencommon.cpp
```diff
@@ -1939,4 +1925,22 @@ void draw_model_rotating(model_render_params *render_info, int model_id, int x1,
...
+	light_reset();
+	auto tempv = vm_vec_new(-1.0f,0.3f,-1.0f);
+	auto tempc = hdr_color(1.0f,0.95f,0.9f, 0.0f, 1.5f);//use 3 for qaz shaders
+	light_add_directional(&tempv,&tempc);
+	tempv.xyz={-0.4f,0.4f,1.1f};
+	tempc = hdr_color(0.788f,0.886f,1.0f,0.0f,1.5f);//use 3 for qaz shaders
+	light_add_directional(&tempv,&tempc);
+	tempv.xyz={0.4f,0.1f,0.4f};
+	tempc = hdr_color(1.0f,1.0f,1.0f,0.0f,0.4f);//use 0.8 for qaz shaders
+	light_add_directional(&tempv,&tempc);
```
BMagnu, line 1942: Eventually, these should be configurable from lighting_profiles I imagine, right?

BMagnu (changes requested):
Some comments and questions

code/lab/dialogs/render_options.cpp
EatThePath: Emissive is a flat amount of white light added to the scene. It's not a factor, in fact it acts more like the horribly named 'ambient factor' currently does. To me it's a pretty arbitrary name, but I'm not very invested in the name either way so if something else is more intuitive to other people then, eh.

code/lighting/lighting.cpp
EatThePath: Regarding the comment, you're correct.
Regarding the assignment, yeah, they could be done all on one line I suppose. 
```cpp
L.I = A.handle(I);
L.I= B.handle(L.I);
```
vs
```cpp
L.I = B.handle(A.handle(I));
```
I think this two line way is more plain to read and half expect the compiler to optimize away any negligible cost incurred, but I'm flexible on that if you disagree.

code/missionui/missionscreencommon.cpp
EatThePath: Yes, that's on my to-do list. Probably broken out by context if desired.

code/lab/dialogs/render_options.cpp
BMagnu: Okay.
In that case, it's fine as is

code/lighting/lighting.cpp
BMagnu: Ah, I see.
IMO, we should really use the lower version, as it's much easier to tell what is actually going on. Or at least use a temporary value to hold the intermediate result from A.handle, so that it's not tripping up people reading the code.
The compiler will, as you say, optimize either to the inline version internally anyways.

BMagnu (approved):
Looks good to me now.

The-E (approved):

PR #4304, opened 2022-05-23
Add LuaAI validity callbacks

BMagnu:
These additional optional callbacks help with more fine grain LuaAI control.
They can allow manual target restriction control for player orders in addition to the builtin target restrictions, as well as allowing for determining AI Mode Achievability (which is used to either defer AI execution for later or toss an AI goal that is unfulfillable).


code/ai/aigoals.h
```diff
@@ -84,6 +84,11 @@ struct ai_goal;
 
 #define MAX_AI_DOCK_NAMES				25
 
+#define AI_GOAL_ACHIEVABLE			1
+#define AI_GOAL_NOT_ACHIEVABLE		2
+#define AI_GOAL_NOT_KNOWN			3
+#define AI_GOAL_SATISFIED			4
```
Goober5000, line 90: As long as these are being moved, can they be turned into `enum class` without too many side effects?

code/ai/ailua.cpp
```diff
@@ -149,7 +137,108 @@ bool ai_lua_is_valid_target(int sexp_op, int target_objnum, ship* self) {
...
+			luaParameters.push_back(luacpp::LuaValue::createValue(action.getLuaState(), scripting::api::l_OSWPT.Set(aigp->lua_ai_target)));
+		}
+
+		auto retVals = action.call(Script_system.GetLuaSession(), luaParameters);
+
+		if (!retVals.empty() && retVals[0].getValueType() == luacpp::ValueType::USERDATA) {
+			scripting::api::enum_h enumData;
+			retVals[0].getValue(scripting::api::l_Enum.Get(&enumData));
+			if (!enumData.IsValid()) {
+				Error(LOCATION, "LuaAI SEXP Availability hook returned an invalid avilability enum.");
```
Goober5000, line 222: "avilability" misspelled

code/ai/ailua.cpp
```diff
@@ -149,7 +137,108 @@ bool ai_lua_is_valid_target(int sexp_op, int target_objnum, ship* self) {
...
+
+			switch (enumData.index) {
+			case scripting::api::LE_LUAAI_ACHIEVABLE:
+				return AI_GOAL_ACHIEVABLE;
+			case scripting::api::LE_LUAAI_NOT_YET_ACHIEVABLE:
+				return AI_GOAL_NOT_KNOWN;
+			case scripting::api::LE_LUAAI_UNACHIEVABLE:
+				return AI_GOAL_NOT_ACHIEVABLE;
+			default:
+				Error(LOCATION, "LuaAI SEXP Availability hook returned an invalid avilability enum.");
```
Goober5000, line 234: here too

code/ai/ailua.cpp
```diff
@@ -149,7 +137,108 @@ bool ai_lua_is_valid_target(int sexp_op, int target_objnum, ship* self) {
...
+				return AI_GOAL_ACHIEVABLE;
+			}
+		}
+		else {
+			Error(LOCATION, "LuaAI SEXP Availability hook did not return any avilability enum!");
```
Goober5000, line 239: and here

code/ai/aigoals.h
BMagnu: Done. The whole area needs a cleanup, replacing int3's with unreachables and asserts instead of if-then-int3's, but that's kinda out of scopefor this I think.

code/ai/ailua.cpp
BMagnu: replaced with the more sensible achievability here anyways

Goober5000 (approved):

PR #4305, opened 2022-05-24
Support using string-based containers in some when-argument SEXPs

jg18:
Support using list containers of strings and map containers with string keys as arguments to the following SEXPs:
- `any-of`
- `every-of`
- `number-of`
- `first-of`

Arguments to these SEXPs can include a mix of strings and containers.

These SEXPs now use the new argument type `OPF_DATA_OR_STR_CONTAINER`.

Arguments retrieved from containers can't be individually invalidated or validated.

Includes some changes that aren't strictly needed until PR #4326, but including them here avoids complications from having to modify those functions in that PR.

Partially addresses Issue #4298.


jg18:
Also in the plans is to allow map containers, in which case their keys will be added to the argument list.

jg18:
Tip: Checking the "Ignore whitespace" box from the gear icon under the "Files changed" tab shrinks the PR to just +135/-10.

fred2/sexp_tree.cpp
```diff
@@ -2869,6 +2870,10 @@ int sexp_tree::get_default_value(sexp_list_item *item, char *text_buf, int op, i
 				item->set_data("<any data>");
 			return 0;
 
+		case OPF_DATA_OR_STR_CONTAINER:
+			item->set_data("<data or string container>");
```
BMagnu, line 2874: I would still call this "any data or string container"

fred2/sexp_tree.cpp
```diff
@@ -3221,6 +3226,10 @@ int sexp_tree::get_default_value(sexp_list_item *item, char *text_buf, int op, i
 			str = "<any data>";
 			break;
 
+		case OPF_DATA_OR_STR_CONTAINER:
+			str = "<data or string container>";
```
BMagnu, line 3230: same here

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -1073,6 +1073,10 @@ int sexp_tree::get_default_value(sexp_list_item* item, char* text_buf, int op, i
 			item->set_data("<any data>");
 		return 0;
 
+	case OPF_DATA_OR_STR_CONTAINER:
+		item->set_data("<data or string container>");
```
BMagnu, line 1077: same here

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -1393,6 +1397,10 @@ int sexp_tree::get_default_value(sexp_list_item* item, char* text_buf, int op, i
 		str = "<any data>";
 		break;
 
+	case OPF_DATA_OR_STR_CONTAINER:
+		str = "<data or string container>";
```
BMagnu, line 1401: same here

code/parse/sexp.cpp
```diff
@@ -10085,43 +10111,48 @@ int test_argument_nodes_for_condition(int n, int condition_node, int *num_true,
...
-			if ( Sexp_nodes[condition_node].value == SEXP_KNOWN_TRUE ||
-					Sexp_nodes[condition_node].value == SEXP_KNOWN_FALSE) {
+			if (Sexp_nodes[condition_node].value == SEXP_KNOWN_TRUE ||
+				Sexp_nodes[condition_node].value == SEXP_KNOWN_FALSE) {
 				val = Sexp_nodes[condition_node].value;
-			} else if ( Sexp_nodes[condition_node].value == SEXP_NAN_FOREVER ) {
+			} else if (Sexp_nodes[condition_node].value == SEXP_NAN_FOREVER) {
 				// In accordance with SEXP_NAN/SEXP_NAN_FOREVER becoming SEXP_FALSE in eval_sexp()
 				val = SEXP_KNOWN_FALSE;
 			}
```
BMagnu, line 10124: This must be within the for loop.
You have to test the validity of the condition for _each_ of the potential arguments, and only apply them to the ``Applicable_arguments_temp`` if that specific argument returns true.
Otherwise, the validity of the first container argument would determine whether or not all container arguments are evaluated, which is evidently incorrect

BMagnu (changes requested):

code/parse/sexp.cpp
jg18: Good catch!

I initially thought it didn't need to be in the loop, since `condition_node` was the only node passed to `eval_sexp()`, but then realized that `<argument>` is going to be different for each call to `eval_sexp()`.

The test mission "Containers as Arguments" that Karajorma posted in #fred on Discord now works correctly again.

BMagnu (approved):
LGTM now

PR #4306, opened 2022-05-24
De-Bitfield-y Player Orders

BMagnu:
This allows for a nearly unlimited number of player orders as opposed to the 32 before.
Previously, orders were sometimes stored and passed as bitfields, and sometimes as indices into the player_orders array.
This PR tidies that up and replaces it to _always_ be an index into the player_orders array. 
As these bitfields were also sent to multi, replacing them with indices is a multi version bump.

As this is a deep rooted change, the following things need to be tested exhaustively before this can be merged:
- [x] Issuing player orders still works
- [x] player orders still have correct behaviour for determining valid targets
- [x] player orders still work with LuaAI
- [x] Using controls-bound shortcuts for player orders still works
- [x] player orders still work in multi
- [x] player orders issue correct hud displays in multi

### DIFF
--- a/code/hud/hudsquadmsg.h
+++ b/code/hud/hudsquadmsg.h
@@ -30,56 +30,51 @@
 
 class object;
 
-// defines for messages that can be sent from the player.  Defined at bitfields so that we can enable
-// and disable messages on a message by message basis
+// defines for messages that can be sent from the player.  Indexes into Player_orders
 
-#define ATTACK_TARGET_ITEM			(1<<0)
-#define DISABLE_TARGET_ITEM			(1<<1)
-#define DISARM_TARGET_ITEM			(1<<2)
-#define PROTECT_TARGET_ITEM			(1<<3)
-#define IGNORE_TARGET_ITEM			(1<<4)
-#define FORMATION_ITEM				(1<<5)
-#define COVER_ME_ITEM				(1<<6)
-#define ENGAGE_ENEMY_ITEM			(1<<7)
-#define CAPTURE_TARGET_ITEM			(1<<8)
+#define ATTACK_TARGET_ITEM			0
+#define DISABLE_TARGET_ITEM			1
+#define DISARM_TARGET_ITEM			2
+#define PROTECT_TARGET_ITEM			3
+#define IGNORE_TARGET_ITEM			4
+#define FORMATION_ITEM				5
+#define COVER_ME_ITEM				6
+#define ENGAGE_ENEMY_ITEM			7
+#define CAPTURE_TARGET_ITEM			8
 
 // the next are for the support ship only
-#define REARM_REPAIR_ME_ITEM		(1<<9)
-#define ABORT_REARM_REPAIR_ITEM		(1<<10)
-#define STAY_NEAR_ME_ITEM			(1<<11)
-#define STAY_NEAR_TARGET_ITEM		(1<<12)
-#define KEEP_SAFE_DIST_ITEM			(1<<13)
+#define REARM_REPAIR_ME_ITEM		9
+#define ABORT_REARM_REPAIR_ITEM		10
+#define STAY_NEAR_ME_ITEM			11
+#define STAY_NEAR_TARGET_ITEM		12
+#define KEEP_SAFE_DIST_ITEM			13
 
 // next item for all ships again -- to try to preserve relative order within the message menu
-#define DEPART_ITEM					(1<<14)
+#define DEPART_ITEM					14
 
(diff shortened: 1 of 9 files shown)


Goober5000:
We'll need to be sure that all multi PRs are merged on the same day.  We're ending up with a lot of them.

fred2/management.cpp
```diff
@@ -584,11 +584,7 @@ int create_ship(matrix *orient, vec3d *pos, int ship_type)
 			if (!(sip->is_small_ship()))
 			{
 				shipp->orders_accepted = ship_get_default_orders_accepted( sip );
-
-				for(size_t i = 0; i < Player_orders.size(); i++) {
-					if (Player_orders[i].id & DEPART_ITEM)
-						shipp->orders_accepted.erase(i);
-				}
+				shipp->orders_accepted.erase(DEPART_ITEM);
```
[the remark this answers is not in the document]
BMagnu: That would be very very hacky, as that (static) enum would have to share a value range with the dynamically alotted lua player orders (which have to be an int / size_t) which are just assigned in increasing order. 
It'd be possible by continually static casting the enum to size_t everywhere it's used as well as explicitly assigning each enum value, but I'm not sure if that makes it better or worse.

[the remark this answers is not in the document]
Goober5000: Ok, I'm fine with leaving this as it is then.

notimaginative:
Where are the bitfields sent in multi? I could just be missing it, but I don't see any breakage here necessitating a multi bump.

BMagnu:
> Where are the bitfields sent in multi? I could just be missing it, but I don't see any breakage here necessitating a multi bump.

The file contains a lot of calls to ``send_player_order_packet``. The calls themselves and that message are untouched, but the value used in its third argument, ``command`` has changed, as it now is an index and not a bitfield. The packet handler then passes that command back to ``hud_squadmsg_send_ship_command`` (or the wing / all fighter alternative). Hence the multi files have not changed. However the data sent, has.

notimaginative:
> > Where are the bitfields sent in multi? I could just be missing it, but I don't see any breakage here necessitating a multi bump.
> 
> The file contains a lot of calls to `send_player_order_packet`. The calls themselves and that message are untouched, but the value used in its third argument, `command` has changed, as it now is an index and not a bitfield. The packet handler then passes that command back to `hud_squadmsg_send_ship_command` (or the wing / all fighter alternative). Hence the multi files have not changed. However the data sent, has.

Ah, got it, I was indeed looking for the wrong thing.

code/hud/hudsquadmsg.h
```diff
@@ -30,56 +30,55 @@
...
-#define IGNORE_TARGET_ITEM			(1<<4)
-#define FORMATION_ITEM				(1<<5)
-#define COVER_ME_ITEM				(1<<6)
-#define ENGAGE_ENEMY_ITEM			(1<<7)
-#define CAPTURE_TARGET_ITEM			(1<<8)
+// defines for messages that can be sent from the player.  Indexes into Player_orders
+
+#define ATTACK_TARGET_ITEM			0
+#define DISABLE_TARGET_ITEM			1
+#define DISARM_TARGET_ITEM			2
```
notimaginative, line 37: Shouldn't this start at `1`? It looks like there is at least one case where a value of `0` is used to imply no command.

notimaginative (changes requested):

code/hud/hudsquadmsg.h
Goober5000: Concur, and good catch.  Previously this was defined as (1<<0) or 1.

BMagnu: good point. Addressed.
Now only multi testing remains (though I am confident that it should work, given the defines applied the changes automatically to send and recieve).
How do we go about multi testing the easiest?

notimaginative: Multi should be fine but I'll pull this into a test build tomorrow and make sure of it. This will be pretty easy for me to test locally so it should be quick to verify whether it works or not.

notimaginative:
Multi appears to be unaffected. All orders issued between host, client, and AI worked just as expected.

notimaginative:
Works as far I as can test. This probably needs to hold off on merging until some other multi changes are ready though, to avoid so many version bumps. The multi sexp PR is ready now but we should check with @JohnAFernandez about the rest.

notimaginative:
Can you remove the multi bump (or I can do it)?
We're going to bump it at once for the 3 breaking PRs today.

PR #4307, opened 2022-05-24
tweak strdup handling for argument vector

Goober5000:
Follow-up to #4293

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -10145,9 +10145,11 @@ int test_argument_nodes_for_condition(int n, int condition_node, int *num_true, 
 	return num_valid_arguments;
 }
 
+enum class STRDUP_STATUS { ALREADY_DUPPED, DUP_NEEDED, DUP_NOT_NEEDED };
+
 // Goober5000
 // NOTE: if you change this function, check to see if the previous function should also be changed!
-int test_argument_vector_for_condition(const SCP_vector<std::pair<const char*, int>> &argument_vector, bool already_dupped, int condition_node, int *num_true, int *num_false, int *num_known_true, int *num_known_false)
+int test_argument_vector_for_condition(const SCP_vector<std::pair<const char*, int>> &argument_vector, STRDUP_STATUS strdup_status, int condition_node, int *num_true, int *num_false, int *num_known_true, int *num_known_false)
 {
 	int val, num_valid_arguments;
 	SCP_vector<std::pair<const char*, int>> Applicable_arguments_temp;
@@ -10204,7 +10206,7 @@ int test_argument_vector_for_condition(const SCP_vector<std::pair<const char*, i
 
 			// if the argument was already dup'd, but not added as an applicable argument,
 			// we need to free it here before we cause a memory leak
-			if ((val == SEXP_FALSE || val == SEXP_KNOWN_FALSE) && already_dupped)
+			if ((val == SEXP_FALSE || val == SEXP_KNOWN_FALSE) && (strdup_status == STRDUP_STATUS::ALREADY_DUPPED))
 				vm_free(const_cast<char*>(argument.first));
 
 			// clear argument, but not list, as we'll need it later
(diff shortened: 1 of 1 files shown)


JohnAFernandez (approved):

code/parse/sexp.cpp
```diff
@@ -10145,9 +10145,11 @@ int test_argument_nodes_for_condition(int n, int condition_node, int *num_true,
 	return num_valid_arguments;
 }
 
+enum class STRDUP_STATUS { ALREADY_DUPPED, DUP_NEEDED, DUP_NOT_NEEDED };
```
z64555, line 10148: Please make a brief doxy explaining what the enum class is for.  The enumerations appear to be aptly named, so I don't believe they require to be individually documented.

I do recommend having one enumeration per line, however.

z64555 (changes requested):

code/parse/sexp.cpp
```diff
@@ -10219,12 +10221,14 @@ int test_argument_vector_for_condition(const SCP_vector<std::pair<const char*, i
 	// Sexp_applicable_argument_list is a stack and we want the first argument in the list to be the first one out
 	while (!Applicable_arguments_temp.empty())
 	{
-		// we need to dup the strings regardless, since we're not using Sexp_nodes[n].text as a string buffer,
-		// but we need to know whether the calling function dup'd them, or whether we should dup them here
-		if (already_dupped)
+		// if we're using a temporary buffer for the string (as opposed to a permanent buffer like shipp->ship_name or Sexp_nodes[n].text)
+		// then we need to dup the strings, but we need to know whether the calling function dup'd them, or whether we should dup them here
+		if (strdup_status == STRDUP_STATUS::ALREADY_DUPPED)
```
z64555, line 10226: This if/else statement could be changed to a switch statement.

code/parse/sexp.cpp
```diff
@@ -10593,7 +10597,7 @@ int eval_for_ship_collection(int arg_handler_node, int condition_node, int op_co
 					if (just_count)
 						num_valid_arguments++;
 					else
-						argument_vector.emplace_back(vm_strdup(shipp->ship_name), -1);
+						argument_vector.emplace_back(shipp->ship_name, -1);
```
z64555, line 10600: vm_strdup is removed here. What effect does that have, exactly?

code/parse/sexp.cpp
```diff
@@ -10705,15 +10709,15 @@ int eval_for_players(int arg_handler_node, int condition_node, bool just_count =
 			if (just_count)
 				num_valid_arguments++;
 			else
-				argument_vector.emplace_back(vm_strdup(Player_ship->ship_name), -1);
+				argument_vector.emplace_back(Player_ship->ship_name, -1);
```
z64555, line 10712: vm_strdup is removed here. What effect does that have, exactly?

code/parse/sexp.cpp
Goober5000: Duping the string is not needed because `ship_name` is a permanent buffer.  The argument to `test_argument_vector_for_condition` is changed to `DUP_NOT_NEEDED` accordingly.

code/parse/sexp.cpp
Goober5000: Same as above

code/parse/sexp.cpp
Goober5000: It could, but I'd rather keep this as it is.  It works well enough, and the switch block would add several additional lines.

code/parse/sexp.cpp
Goober5000: I view this as an extension of the following function definition, so I don't think it requires a multi-line declaration or a doxy.  I've added a doxy anyway, but I left it single-line so that it's visually close to the function.

z64555: Yep, that's what I was suggesting, a one-liner doxy. 👍

z64555: Well... I'll allow it.

z64555 (approved):

PR #4308, opened 2022-05-25
Handle deprecated cmdline flags programmatically.

MageKing17:
Instead of keeping the old variables around (except renamed to include "_deprecated") and manually checking them one at a time with a lot of duplicated code, use existing information to generate the messages automatically.

Got the idea from @EatThePath complaining about the code duplication, and realizing we already had the necessary information to do this automatically. Tested to make sure that the new code not only compiles, but also, given some deprecated flags, produces the same debug log output as the old code.


EatThePath (approved):
Looks pretty good to me

Goober5000 (approved):
Me too

PR #4313, opened 2022-05-26
Allow Weapon To Override Turret Name

TheForce172:
Done following a discussion in #Freespace a week or so ago. Allows a weapon to override a turrets name if it is the first weapon in a turret. Does not override a Subsystems Alt name if specified in ships.tbl

### DIFF
--- a/code/hud/hudtargetbox.cpp
+++ b/code/hud/hudtargetbox.cpp
@@ -1566,16 +1566,30 @@ void get_turret_subsys_name(ship_weapon *swp, char *outstr)
 
 	//WMC - find the first weapon, if there is one
 	if (swp->num_primary_banks || swp->num_secondary_banks) {
-		int weaponid;
+		bool done = false;
 		if (swp->num_primary_banks) {
-			weaponid = swp->primary_bank_weapons[0];
+			for (auto weapon : swp->primary_bank_weapons) {
+				if (weapon >= 0) {
+					if (strlen(Weapon_info[weapon].altSubsysName) != 0) {
+						sprintf(outstr, "%s", Weapon_info[weapon].altSubsysName);
+						done = true;
+						break;
+					}
+				}
+			}
+		} 
+		if ((swp->num_secondary_banks) && !done) {
+			for (auto weapon : swp->secondary_bank_weapons) {
+				if (weapon >= 0) {
+					if (strlen(Weapon_info[weapon].altSubsysName) != 0) {
+						sprintf(outstr, "%s", Weapon_info[weapon].altSubsysName);
+						done = true;
+						break;
+					}
+				}
+			}
 		}
-		else {
-			weaponid = swp->secondary_bank_weapons[0];
-		}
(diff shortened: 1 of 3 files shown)


code/hud/hudtargetbox.cpp
```diff
@@ -1566,43 +1566,51 @@ void get_turret_subsys_name(ship_weapon *swp, char *outstr)
...
-			sprintf(outstr, "%s", XSTR("Beam turret", 1567));
-		} else if (flags[Weapon::Info_Flags::Flak]) {
-			sprintf(outstr, "%s", XSTR("Flak turret", 1566));
+		int weaponid;
+		if (swp->num_primary_banks) {
+			weaponid = swp->primary_bank_weapons[0];
+		}
+		else {
+			weaponid = swp->secondary_bank_weapons[0];
+		}
```
Baezon, line 1575: Perhaps it should search through all the banks instead of just checking the first ones?

Baezon (changes requested):

code/hud/hudtargetbox.cpp
```diff
@@ -1566,43 +1566,65 @@ void get_turret_subsys_name(ship_weapon *swp, char *outstr)
...
-					sprintf(outstr, "%s", XSTR("Turret", 1487));
-				}
-				// the TVWP has some primaries flagged as bombs
-				else if (flags[Weapon::Info_Flags::Bomb])
-				{
-					sprintf(outstr, "%s", XSTR("Missile lnchr", 1569));
+		bool done = false;
+		if (swp->num_primary_banks) {
+			for (auto weaponid : swp->primary_bank_weapons) {
+				if (weaponid >= 0) {
```
Goober5000, line 1572: Why not loop from 0 to `swp->num_primary_banks`?  All weaponids after that point are going to be -1 anyway.

code/hud/hudtargetbox.cpp
```diff
@@ -1566,43 +1566,65 @@ void get_turret_subsys_name(ship_weapon *swp, char *outstr)
...
+					if (strlen(Weapon_info[weaponid].altSubsysName) != 0) {
```
Goober5000, line 1573: I would check that the string is non-empty (checking that the first character != 0) rather than getting the length.

code/hud/hudtargetbox.cpp
```diff
@@ -1566,43 +1566,65 @@ void get_turret_subsys_name(ship_weapon *swp, char *outstr)
...
+		} 
+		if ((swp->num_secondary_banks) && !done) {
+			for (auto weaponid : swp->secondary_bank_weapons) {
+				if (weaponid >= 0) {
+					if (strlen(Weapon_info[weaponid].altSubsysName) != 0) {
+						sprintf(outstr, "%s", Weapon_info[weaponid].altSubsysName);
+						done = true;
+						break;
+					}
 				}
```
Goober5000, line 1589: Same comments here.

Goober5000:
Ack, badly timed...

PR #4314, opened 2022-05-26
Fix rollback primary collisions

JohnAFernandez:
Taylor, Wookiejedi, and I have all been going back and forth heavily testing these fixes.  On stationary targets, we have 100% accuracy.  On moving targets, we have one more test pending, but results are promising so far, even at high pings.  This improvement works best with Interpolation Overhaul 2, but is independent of it.

### DIFF
--- a/code/network/multi_obj.cpp
+++ b/code/network/multi_obj.cpp
@@ -56,7 +56,7 @@ struct oo_ship_position_records {
 
 // keeps track of what has been sent to each player, helps cut down on bandwidth, allowing only new information to be sent instead of old.
 struct oo_info_sent_to_players {	
-	int timestamp;					// The overall timestamp which is used to decide if a new packet should be sent to this player for this ship.
+	TIMESTAMP timestamp;					// The overall timestamp which is used to decide if a new packet should be sent to this player for this ship.
 
 	vec3d position;					// If they are stationary, there's no need to update their position.
 	float hull;						// no need to send hull if hull hasn't changed.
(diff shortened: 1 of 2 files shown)


JohnAFernandez:
Moving this back to draft for a little bit, because I just realized there is one more fix missing for doubled collisions.

code/network/multi_obj.cpp
```diff
@@ -703,19 +731,28 @@ void multi_ship_record_do_rollback()
...
 
+	bool continue_rollback = true;
+
 	do {
 		// move all ships to their recorded positions
 		multi_oo_restore_frame(frame_idx);
 
 		// push weapons forward for the frame (weapons do not get pushed forward for the first frame of their existence)
-		multi_oo_simulate_rollback_shots(frame_idx);
+		continue_rollback = (multi_oo_simulate_rollback_shots(frame_idx));
```
notimaginative, line 741: A bit of a nitpick, but the extra parenthesis should be removed. clang tends to complain about this too so I'm a little surprised that clang-tidy let it through.

notimaginative (changes requested):
Just the one minor thing, but otherwise it looks good to me.

code/network/multi_obj.cpp
```diff
@@ -496,25 +508,30 @@ int multi_ship_record_find_frame(int client_frame, int time_elapsed)
 	for (int i = Oo_info.cur_frame_index - 2; i > -1; i--) {
 
 		// Check to see if the client's timestamp matches the recorded frames.
-		if ((Oo_info.timestamps[i] <= target_timestamp) && (Oo_info.timestamps[i + 1] > target_timestamp)) {
+		if ((Oo_info.timestamps[i] <= target_timestamp) && (Oo_info.timestamps[i + 1] >= target_timestamp)) {
```
Goober5000, line 511: This looks like it could benefit from a timestamp ordering check.  I'll add that to the `timestamp_since` PR.

Goober5000: Or a new PR.  See #4324.

code/network/multi_obj.cpp
JohnAFernandez: My reasoning was always to help indicate that it is a bool assignment.  I'll refrain from now on.

notimaginative (approved):

JohnAFernandez:
Literally waiting on one last safety test before merging

notimaginative (approved):

PR #4319, opened 2022-05-27
Add gr.CurrentResizeMode

Goober5000:
The various graphics drawing functions default to the `GR_RESIZE_NONE` drawing behavior.  When the screen is scaled, it is useful to be able to scale the drawing accordingly.  Rather than add duplicates of all functions, as was done in #4003 for `gr.drawStringResized`, it is useful to add one virtvar to control sizing across the board.

This also fixes a gap in the enumeration sequence, and patches the movie-type enumerations to use the correct indexing.


code/scripting/api/libs/graphics.cpp
```diff
@@ -818,7 +840,7 @@ ADE_FUNC(drawModel, l_Graphics, "model model, vector position, orientation orien
 	matrix *orient = mh->GetMatrix();
 
 	//Clip
-	gr_set_clip(0, 0, gr_screen.max_w, gr_screen.max_h, GR_RESIZE_NONE);
+	gr_set_clip(0, 0, gr_screen.max_w, gr_screen.max_h, GR_RESIZE_NONE);	// don't use lua_ResizeMode here
```
jg18, line 843: Optional: maybe explain why not? Ditto for the other spot where this happens.

jg18 (approved):
Looks good. One optional request.

code/scripting/api/libs/graphics.cpp
Goober5000: Done

PR #4321, opened 2022-05-28
remove safe_strings and convert strncpy to strncpy_s

Goober5000:
Portej05's original comment in safe_strings read as follows:
```
 * VC2005+ define some safe string functions which check buffer sizes before doing anything
 * Unfortunately, GCC and MACOS do not provide these functions, therefore, we must!
 * (if only to reduce the amount of noise the static analysis tools are spitting out)
 * They are part of ISO/IEC TR 24731 and may find their way into the CRTs at some point, at which
 * point these functions must be removed from the engine.
 * While these functions do not add a huge amount of benefit for heap-allocated strings, they
 * can protect against a class of buffer overruns in stack allocated situations.
```
Since the codebase now uses C++11, this can be removed.

Additionally, I have gone through the codebase and replaced most instances of `strncpy` with `strncpy_s` which always adds a null-terminator and includes some extra error checking.  Quite a few of these places did not use the length properly and have been fixed.

Inspired by #4320.


Goober5000:
Converting to draft for now because these functions are not actually supported in GCC, etc.  See the following links:
https://stackoverflow.com/questions/40045973/strcpy-s-not-working-with-gcc
https://sourceforge.net/projects/safeclib/

PR #4324, opened 2022-05-28
timestamp comparisons

Goober5000:
Add methods to compare timestamps, as well as methods to calculate deltas from existing timestamps.

Follow-up to #4286.

### DIFF
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -254,7 +254,7 @@ int timestamp_until(int stamp)
 
 int timestamp_until(TIMESTAMP stamp)
 {
-	Assertion(stamp.isValid() && !stamp.isNever(), "timestamp_until was called with an Invalid or Never timestamp!");
+	Assertion(stamp.isValid() && !stamp.isNever(), "timestamp_until was called with a%s timestamp!", !stamp.isValid() ? "n invalid" : " Never");
 	if (!stamp.isValid() || stamp.isNever())
 		return INT_MAX;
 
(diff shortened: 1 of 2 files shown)


code/io/timer.cpp
```diff
@@ -289,17 +305,111 @@ int ui_timestamp_since(UI_TIMESTAMP stamp)
...
+		return 0;
+	else if (t1.value() > t2.value())
+		return 1;
+	else
+		return -1;
+}
+
+bool timestamp_in_between(TIMESTAMP stamp, TIMESTAMP before, TIMESTAMP after)
+{
+	Assertion(stamp.isValid() && !stamp.isNever(), "timestamp_in_between was called with an Invalid or Never 'stamp' timestamp!");
```
JohnAFernandez, line 380: If there's two conditions on an assert or assertion, we really should output which statement tripped it up.

JohnAFernandez (changes requested):

code/io/timer.cpp
Goober5000: Which statement?  Do you mean whether it was 'Invalid' or 'Never'?

Is that really necessary?  That would double the number of Assertions here.  There are already three for the three separate parameters.

JohnAFernandez: What's wrong with `Assertion(stamp.isValid() && !stamp.isNever(), "timestamp_in_between was called with an %s 'stamp' timestamp!", (stamp.isNever()) ? "isNever" : "invalid");`?

Goober5000: Seems a bit convoluted, but I can add it.

JohnAFernandez (approved):

code/io/timer.cpp
JohnAFernandez: Thanks.  Imho, we make the code clear when we can, but clarity of Assertions and error messages take priority over that.

PR #4325, opened 2022-05-28
removed vestigial bit-depth asserts for aabitmaps.

EatThePath:
Removes all the relevant asserts I could find

### DIFF
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -592,12 +592,9 @@ void bm_convert_format(bitmap *bmp, ushort flags) {
 		return;
 
 	if (Is_standalone) {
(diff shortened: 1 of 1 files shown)


Baezon:
I'm not familiar with this code, but in general, if this *is* vestigial you should explain why, or link to issues/discussion where this was discovered, etc.

EatThePath:
> I'm not familiar with this code, but in general, if this _is_ vestigial you should explain why, or link to issues/discussion where this was discovered, etc.

Entirely right. I had the idea that the name of the branch would have github establish a link to the issue where this was originally discussed. Evidently not. I've edited the description to be more informative.

Goober5000 (approved):
LGTM

PR #4326, opened 2022-05-29
Support using containers with random-multiple-of SEXP

jg18:
Support using list containers of strings and map containers with string keys as arguments to `random-multiple-of`.

Depends on PR #4305.

Also partially addresses Issue #4298.

Would appreciate help with testing.

With my current test mission ([test_random_multiple_of.fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/8792805/test_random_multiple_of.fs2.txt)) it works as expected the first time I press "2", but the second time I press "2", random messages get sent out in what feels like an infinite loop. *However*, when I use a build without my changes and use strings for all arguments instead, I get the same issue, so my mission is likely just buggy (I'm not a FREDder).


jg18:
Tip: Once again, using the "Ignore whitespace" option under the "Files changed" tab shrinks the PR a bit and makes it easier to see what's changed.

Karajorma:
Yeah, the problem was a FREDding issue. Key-reset is notoriously buggy with certain keys. Switching to Key-reset-multiple fixes the mission and it then works perfectly.

code/parse/sexp.cpp
```diff
@@ -10421,11 +10414,122 @@ int eval_random_of(int arg_handler_node, int condition_node, bool multiple)
...
+	if (Sexp_nodes[n].flags & SNF_ARGUMENT_VALID)
+	{
+		// ensure special argument list is empty
+		Sexp_applicable_argument_list.clear_nesting_level();
+
+		const int arg_value_index = (Sexp_nodes[n].subtype == SEXP_ATOM_CONTAINER_NAME)
+										? random_arg_value - cumulative_arg_counts[random_argument - 1] - 1
+										: -1;
+		const int num_args = copy_node_to_replacement_args(n, arg_value_index);
+		SCP_UNUSED(num_args);
```
BMagnu, line 10516: This is kind of weird.
Instead of avoiding the warning with a macro, you should just not store the result.

BMagnu (changes requested):
A minor detail.
While the split of random-of and random-multiple-of is annoying, I understand why we need it, as random-of is fundamentally incompatible with the dynamic nature of the containers.

BMagnu (approved):

PR #4327, opened 2022-05-29
add some extra safeguards to gr.drawString

Goober5000:
These will add insight if the errors reported by RedMadeJoe pop up again.  The root cause is in the script, but this will prevent FSO from choking.

### DIFF
--- a/code/scripting/api/libs/graphics.cpp
+++ b/code/scripting/api/libs/graphics.cpp
@@ -1250,20 +1250,26 @@ static int drawString_sub(lua_State *L, bool use_resize_arg)
 		SCP_vector<int> linelengths;
 		SCP_vector<const char*> linestarts;
 
-		if (y2 >= 0 && y2 < y)
+		if (x2 <= x)
+		{
+			// This would pass a <=0 value to split_str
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/graphics.cpp
```diff
@@ -1250,20 +1250,26 @@ static int drawString_sub(lua_State *L, bool use_resize_arg)
 		SCP_vector<int> linelengths;
 		SCP_vector<const char*> linestarts;
 
-		if (y2 >= 0 && y2 < y)
+		if (x2 <= x)
+		{
+			// This would pass a <=0 value to split_str
+			Warning(LOCATION, "Illegal values passed to gr.drawString: x2 (%d) must be greater than x (%d).", x2, x);
```
JohnAFernandez, line 1256: Won't this throw a warning every time that swapped values are used?  Can we use a `static bool` to warn just once?

Goober5000: Yeah, we could do that.  Done.

wookieejedi (approved):
Extra safeguards look good.

PR #4328, opened 2022-05-29
Remove pointless assertion in mission parsing.

MageKing17:
When parsing escort priority, FSO asserts that the escort flag is set, but this assertion is only valid in missions output by FRED; it's fairly trivial to use notepad to remove the flag without removing escort priority, thereby triggering the assertion. Furthermore, there doesn't actually seem to be a problem if a parse object has an escort priority without the escort flag; the engine won't actually use the priority until/unless the ship is manually added to the escort list, and then it doesn't matter that one is set.


EatThePath (approved):
If missions that would have been caught in this work fine, this seems reasonable and straight-forward to me.

Goober5000 (approved):

PR #4330, opened 2022-05-30
Detect concurrent container modification

jg18:
Detect attempts to call a mutation SEXP on a container while the container is being used as an argument to a `*-argument` SEXP.

Depends on PR #4305.


z64555:
Whats the severity of this PR, and should it be included in 22.2?

jg18:
@z64555 I'd put it at Critical, since the intent is to avoid crashes that could result from FREDders trying to modify a container while they're using it as an argument to a `*-of` SEXP.

As I put it in #scp-private, it's the FSO equivalent of Java's [`ConcurrentModificationException`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ConcurrentModificationException.html).

z64555:
Alright, is this PR ready for review then?

jg18:
Hoping to get it in a reviewable state this weekend.

jg18:
Here are two versions of the same test mission from Karajorma in #fred.

The ["bad" version](https://github.com/scp-fs2open/fs2open.github.com/files/8839051/Bad_Containers_As_Args.fs2.txt) triggers the new warning but the ["good" version](https://github.com/scp-fs2open/fs2open.github.com/files/8839052/Good_Containers_As_Args.fs2.txt) doesn't.

JohnAFernandez (approved):
It looks so simple and elegant.  It's odd that it took quite a few commits to get it to this state.

jg18:
@JohnAFernandez the reason for all those commits is that it's a branch off two other PR branches (can't look up numbers at the moment).

jg18:
The PR's commits are mostly left over from PR #4299 and PR #4305 which are now merged.

They'll all be eliminated in the squash/merge, of course.

jg18:
For the record, this PR was also part of completing Issue #4298.

PR #4331, opened 2022-05-31
Draw 3d lines in space with Lua

Kiloku:
This PR gives 3D line drawing functionality. Lua scripters can use the `draw3dLine` function to draw a line of any width from point A to point B in space, with proper depth checking. 

The lua function supports both an opaque mode and a translucent mode. The opaque mode draws a fully opaque line. The translucent mode blends the line additively into the scene. It does not use the alpha value from lua's graphics api working color (this is due to technical limitations).


code/scripting/api/libs/graphics.cpp
```diff
@@ -795,6 +796,40 @@ ADE_FUNC(drawSphere, l_Graphics, "[number Radius = 1.0, vector Position]", "Draw
...
+		
+		bm_set_render_target(pseudo_beam_line_texture, 0);
+		color temp = gr_screen.current_color; 	//Store our working color
+		color c;
+		gr_init_color(&c, 255, 255, 255);
+		gr_set_color_fast(&c); 					//set white as the working color
+		gr_pixel(0, 0, GR_RESIZE_NONE);			//paint the pixel to the texture
+		gr_set_color_fast(&temp);				//Reset our working color
+		bm_set_render_target(-1, 0);			//Reset our render target
+	}
```
Baezon, line 826: This is some pretty ugly low-level code for something relatively high up the 'API chain'. I feel like these dirty details ought to be contained entirely inside your new `batching_add_beam()` function instead, maybe called something else, since it's kind of taking a life on it's own. Remember, this could be used in several other ways in other places in the code, do you really want all the callers to have to repeat all this stuff?

Baezon (changes requested):

code/scripting/api/libs/graphics.cpp
Kiloku: That's fair. I guess it makes no sense to pass a texture to the new add_beam function if it's always the same texture, right? 

I also agree on naming the function something else. Thanks for the feedback.

Baezon (approved):
LGTM

code/scripting/api/libs/graphics.cpp
```diff
@@ -795,6 +796,28 @@ ADE_FUNC(drawSphere, l_Graphics, "[number Radius = 1.0, vector Position]", "Draw
...
+{
+	vec3d *v1 = &vmd_zero_vector;
+	vec3d *v2 = &vmd_zero_vector;
+	bool translucent = true;
+	float thickness = 1.0f;
+	float thicknessEnd = -1.0f;
+
+	if (!ade_get_args(L, "oo|bff", l_Vector.GetPtr(&v1), l_Vector.GetPtr(&v2), &translucent, &thickness, &thicknessEnd))
+		return ADE_RETURN_NIL;
+
```
Goober5000, line 811: I recommend adding a bounds check for `thickness` here, in case someone specifies a negative value for some reason.

Goober5000 (changes requested):

code/scripting/api/libs/graphics.cpp
Kiloku: Zero works on either end, it just tapers into a single pixel point, which IMO doesn't look very good but might be an effect someone wants. 
image

Doing zero on both values at the same time, however, makes the line invisible. I'll add a check for negative values on either end and another for zero on both ends, then.

Kiloku: I also added similar checks in the `batching_add_line` function just in case it eventually gets used by some other feature added in the future

Goober5000 (approved):

PR #4334, opened 2022-05-31
Add timestamp get delta

JohnAFernandez:
For returning the delta between two timestamps, as a counterpart to timestamp_delta.

Found this functionality to be missing while trying to upgrade rollback to use the new timestamp code.  Hopefully, the only missing piece.


code/io/timer.h
```diff
@@ -138,6 +138,9 @@ UI_TIMESTAMP ui_timestamp(int delta_ms);
 TIMESTAMP timestamp_delta(TIMESTAMP stamp, int delta_ms);
 UI_TIMESTAMP ui_timestamp_delta(UI_TIMESTAMP stamp, int delta_ms);
 
+int timestamp_get_delta(TIMESTAMP before, TIMESTAMP after);
+int ui_timestamp_get_delta(TIMESTAMP before, TIMESTAMP after);
```
Goober5000, line 142: The parameters to the ui_ function need to be the UI_TIMESTAMP type.

(After 22.2 is out I'm going to merge the regular and ui_ functions using templates.  The types need to be distinct, but the functions can be the same.)

code/io/timer.cpp
```diff
@@ -179,6 +179,39 @@ UI_TIMESTAMP ui_timestamp_delta(UI_TIMESTAMP stamp, int delta_ms)
 	return UI_TIMESTAMP(stamp.value() + delta_ms);
 }
 
+// Returns the difference between two timestamps as an int, does not check that the result will be positive.
+// Also only error check if a timestamp is invalid, but when either is immediate or never, it returns 0.
+int timestamp_get_delta(TIMESTAMP before, TIMESTAMP after)
+{
+	Assertion(before.isValid(), "timestamp_get_delta called with an invalid before timestamp%s. This is a coder mistake, please report!", (after.isValid()) ? "" : " and an invalid after timestamp.");
```
Goober5000, line 186: Why add the check for the 'after' timestamp here, given that the next line checks it anyway?

code/io/timer.cpp
```diff
@@ -179,6 +179,39 @@ UI_TIMESTAMP ui_timestamp_delta(UI_TIMESTAMP stamp, int delta_ms)
...
+	Assertion(after.isValid(), "timestamp_get_delta called with an invalid after timestamp. This is a coder mistake, please report!");
+
+
+	if (!before.isValid() || before.isImmediate() || before.isNever())
+		return 0;
+
+	if (!after.isValid() || after.isImmediate() || after.isNever())
+		return 0;
```
Goober5000, line 194: I would say that if 'before' is immediate and 'after' is never, the delta should be INT_MAX, and if reversed it should be INT_MIN.

Goober5000 (changes requested):

code/io/timer.cpp
JohnAFernandez: Because the after timestamp could be invalid as well, and it's helpful to know whether both are invalid instead of just one.

Goober5000: Yeah but literally the very next line will let you know.

JohnAFernandez: Except that literally the next line will not trigger on the same run if the first one does.

Goober5000: Unless you deliberately continue, but yeah, I see your point.

code/io/timer.cpp
```diff
@@ -179,6 +179,59 @@ UI_TIMESTAMP ui_timestamp_delta(UI_TIMESTAMP stamp, int delta_ms)
...
 
+// Returns the difference between two timestamps as an int, does not check that the result will be positive.
+// Also only error check if a timestamp is invalid, but when either is immediate or never, it returns 0.
+int timestamp_get_delta(TIMESTAMP before, TIMESTAMP after)
+{
+	Assertion(before.isValid(), "timestamp_get_delta called with an invalid before timestamp%s. This is a coder mistake, please report!", (after.isValid()) ? "" : " and an invalid after timestamp.");
+	Assertion(after.isValid(), "timestamp_get_delta called with an invalid after timestamp. This is a coder mistake, please report!");
+
+
+	if (!before.isValid() || after.isValid()) {
```
Goober5000, line 190: I think this is supposed to be `!after`.  Ditto for the other function

JohnAFernandez: Whoops.

Goober5000 (approved):

PR #4336, opened 2022-06-01
Add release CI config for Mac

virusman:
Adds release CI config for Mac. Note: this only adds enough to build and upload Mac artifacts to the Github release, but it doesn't change anything in the Python scripts that upload to Nebula and forums. That'd be a separate PR, also I'm not familiar with all that code so far.

### DIFF
--- a/.github/workflows/build-release.yaml
+++ b/.github/workflows/build-release.yaml
@@ -310,11 +310,104 @@ jobs:
             ${{ steps.generate_package.outputs.package_path }}
             ${{ steps.generate_package.outputs.debug_path }}
 
+  build_mac:
+    strategy:
+      matrix:
+        configuration: [FastDebug, Release]
+        compiler: [clang]
+        arch: [x86_64]
+    name: Mac
+    needs: create_release
+    runs-on: macos-latest
+    steps:
+      - name: Cache Qt
+        id: cache-qt-mac
+        uses: actions/cache@v1
+        with:
+          path: ${{ github.workspace }}/../Qt
+          key: ${{ runner.os }}-QtCache-${{ env.QT_VERSION }}
+      - name: Install Qt
+        uses: jurplel/install-qt-action@v2
+        with:
+          version: ${{ env.QT_VERSION }}
+          dir: ${{ github.workspace }}/..
+          cached: ${{ steps.cache-qt-mac.outputs.cache-hit }}
+          setup-python: 'false'
+          aqtversion: ==1.1.3
+
+      - uses: actions/checkout@v1
+        name: Checkout
+        with:
+          submodules: true
+          fetch-depth: 0
+      - name: Configure CMake
+        env:
+          CONFIGURATION: ${{ matrix.configuration }}
+          COMPILER: ${{ matrix.compiler }}
+          ARCHITECTURE: ${{ matrix.arch }}
+          JOB_CMAKE_OPTIONS: ${{ matrix.cmake_options }}
+        run: $GITHUB_WORKSPACE/ci/linux/configure_cmake.sh
+      - name: Compile
+        working-directory: ./build
+        run: LD_LIBRARY_PATH=$Qt5_DIR/lib:$LD_LIBRARY_PATH ninja all
+      - name: Run Tests
+        working-directory: ./build
+        env:
+          CONFIGURATION: ${{ matrix.configuration }}
+          XDG_RUNTIME_DIR: /root
+        run: $GITHUB_WORKSPACE/ci/linux/run_tests.sh
+      - name: Package build result
+        working-directory: ./build
+        run: cd bin && tar -cvzf macos-build.tar *.app
+      - name: Upload build result
+        uses: actions/upload-artifact@v2
+        with:
+          name: mac-${{ matrix.configuration }}
+          path: ${{ github.workspace }}/build/bin/macos-build.tar
+  mac_zip:
+    name: Build Mac distribution zip
+    needs: build_mac
+    runs-on: ubuntu-latest
+    container: ghcr.io/scp-fs2open/sftp_upload:sha-748b19d
+    steps:
+      - uses: actions/checkout@v2
+        name: Checkout
+        with:
+          submodules: true
+          fetch-depth: '0'
+          ref: '${{ github.ref }}'
+      - name: Download Release builds
+        uses: actions/download-artifact@v2
+        with:
+          name: mac-Release
+          path: builds
+      - name: Download FastDebug builds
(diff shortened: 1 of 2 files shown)


z64555:
Postponing this PR for after the 22.2 release cycle.

z64555:
Um, actually.  This RC may be a perfect opportunity to review the return of mac builds.  However its not a top priority at the moment.  The past few days have been plagued by CI issues preventing the builds being published.

Perhaps by RC2 or 3 this will get merged in.

virusman:
@z64555 That'd be great. Let me know if I can help with anything.

virusman:
@z64555 Could this get in for 22.2, or is it deferred to a later release?

z64555:
I've been tied up with IRL events and haven't been able to do much code stuff.  I'm hoping this week I'll be able to get some things done, so please check on me again in a few days.

virusman:
@z64555 Are there any plans to continue with this?

.github/workflows/build-release.yaml
```diff
@@ -310,11 +310,110 @@ jobs:
...
+          cached: ${{ steps.cache-qt-mac.outputs.cache-hit }}
+          setup-python: 'false'
+          aqtversion: ==1.1.3
+
+      - uses: actions/checkout@v1
+        name: Checkout
+        with:
+          submodules: true
+          fetch-depth: 0
+      - name: Set up test version
```
z64555, line 343: Release workflows do not use `version_override.cmake`, using the strings in `version.cmake` as is.  The release procedure updates the version strings before running the workflow, although this could be a candidate for automating from Github instead of needing to make a commit with the version changes.

.github/workflows/build-release.yaml
```diff
@@ -310,11 +310,110 @@ jobs:
             ${{ steps.generate_package.outputs.package_path }}
             ${{ steps.generate_package.outputs.debug_path }}
 
+  build_mac:
+    strategy:
+      matrix:
+        configuration: [FastDebug, Release]
+        compiler: [clang]
```
z64555, line 317: The compiler string must match the options listed in linux/configure_cmake.sh.  There may be another script somewhere that sanitizes "clang" into "clang-9," so it is a good idea to double check.

.github/workflows/build-release.yaml
```diff
@@ -310,11 +310,110 @@ jobs:
...
+          [[ "${{ github.ref }}" =~ ^refs\/heads\/test\/(.*)$ ]]
+          # Override the revision string so that the builds are named correctly
+          echo "set(FSO_VERSION_REVISION_STR ${BASH_REMATCH[1]})" > "version_override.cmake"]
+      - name: Configure CMake
+        env:
+          CONFIGURATION: ${{ matrix.configuration }}
+          COMPILER: ${{ matrix.compiler }}
+          ARCHITECTURE: ${{ matrix.arch }}
+          JOB_CMAKE_OPTIONS: ${{ matrix.cmake_options }}
+        run: $GITHUB_WORKSPACE/ci/linux/configure_cmake.sh
```
z64555, line 355: Double check that the `$RUNNER_OS` envvar is set to "macOS", as required by linux/configure_cmake.sh

.github/workflows/build-release.yaml
```diff
@@ -310,11 +310,110 @@ jobs:
...
+      - name: Compile
+        working-directory: ./build
+        run: LD_LIBRARY_PATH=$Qt5_DIR/lib:$LD_LIBRARY_PATH ninja all
+      - name: Run Tests
+        working-directory: ./build
+        env:
+          CONFIGURATION: ${{ matrix.configuration }}
+          XDG_RUNTIME_DIR: /root
+        run: $GITHUB_WORKSPACE/ci/linux/run_tests.sh
```
z64555, line 364: The `run_tests.sh` currently does not have a check for `[$CONFIGURATION = 'FastDebug']`, please double check that this is intended.

z64555 (changes requested):
Looks good so far, we just need to double check a few things.

.github/workflows/build-release.yaml
virusman: Is there a reason to use the old `macOS` params? The PR checks build without the macOS override, and it seems to work fine.

.github/workflows/build-release.yaml
virusman: Doesn't this apply to Linux builds as well?

.github/workflows/build-release.yaml
virusman: Also `$RUNNER_OS` isn't even set anywhere, I believe it's a leftover from the old build system.

virusman: Ah, no, `RUNNER_OS` is actually set by Github Actions automatically.

virusman: Yeah, this already works in PR builds, it should work just fine in release. Github Actions sets this variable.

.github/workflows/build-release.yaml
virusman: It works fine in PR builds, is overriding `CC` and `CXX` necessary? They default to clang on Mac anyway.

z64555: According to the notes in configure_cmake.sh the `CC` and `CXX` overrides are a workaround for a bug.  I'm not sure if that's still relevant or not for mac.

.github/workflows/build-release.yaml
z64555: I don't think it does, but since I'm not familiar at all with the specifics of Mac building and debugging I thought it would be best to have you double check.  The conditional within `run_tests.sh` seem to add more flags/options before running the unit test app.

.github/workflows/build-release.yaml
virusman: This works fine on PR checks, and this config is no different, so I guess this doesn't need the workaround?

.github/workflows/build-release.yaml
virusman: Removed

.github/workflows/build-release.yaml
virusman: We're not using valgrind on FastDebug or Release builds on Linux, and I don't think we need to do that on Mac either.

virusman:
@z64555 Answered questions and addressed the feedback, this is ready for another review iteration. Thanks!

z64555 (approved):
LGTM.

virusman:
@z64555 Thanks, feel free to merge whenever. I can't merge myself.

PR #4337, opened 2022-06-01
Add arm64 builds for Mac

virusman:
(Stacked PR)

Enable arm64 builds for Mac.


Goober5000:
@virusman Could you rebase this and make whatever other updates are needed?  Once that's done, we'll see if we can get someone to review it.

notimaginative:
This has been superseded by #5853 and the series of commits leading up to it which made arm64 builds possible/usable. Some of that was based on your work @virusman, so thanks!

virusman:
@notimaginative This is great, thanks for doing that!

PR #4339, opened 2022-06-01
log non-psnet packets for debugging

notimaginative:
Previously, debug builds would assert if a packet was not a psnet packet
type. Port scans and general traffic checks could trigger this for any
public server. So we should just log these events instead, taking into
account that some of this traffic could be malicious in nature.


JohnAFernandez (approved):
Everything looks fine.

 11 is a magic number, but I'm not going to make a big deal about it.

PR #4341, opened 2022-06-01
better error checking of campaign loop branches in FRED

Goober5000:
In the FRED campaign editor, error checking for campaign branches did not correctly handle loop branches.  This avoids performing inapplicable error checking for loop branches, adds a new applicable check, and excludes loop branches from the count of regular branches.  Fixes error checking for the FS2 main campaign.


wookieejedi (approved):
Solid quick turnaround.

PR #4342, opened 2022-06-02
Add option to expand warning on unbound keys

### LINKED ISSUE #4340: FR: When a hotkey text replacement string refers to an unbound control, use the bind name instead of "none"
The average user is pretty sunk if a fredder is using an uncommon control that a user doesn't have bound.

wookieejedi:
In Training missions, the game will check if keys used in directives are bound or unbound, and if unbound will display a message to the player to bind the key.

This is limited to only Training missions though, so this PR adds a `game_settings` option to allow modders to use that check in singleplayer mission types too, which is very useful for mods that rely on key presses in non-training missions. Also fixes #4340.

### DIFF
--- a/code/mission/missiontraining.cpp
+++ b/code/mission/missiontraining.cpp
@@ -658,7 +658,8 @@ SCP_string message_translate_tokens(const char *text)
 				ptr = translate_key(temp);  // try and translate key
 				if (ptr != nullptr) {  // was key translated properly?
 					if (!stricmp(ptr, NOX("none")) && (Training_bind_warning != Missiontime)) {
-						if ( The_mission.game_type & MISSION_TYPE_TRAINING ) {
+						// check if a warning message should be displayed if the key is unbound
+						if ( (The_mission.game_type & MISSION_TYPE_TRAINING) || (Always_warn_player_about_unbound_keys & MISSION_TYPE_SINGLE) ) {
 							r = popup(PF_TITLE_BIG | PF_TITLE_RED, 2, XSTR( "&Bind Control", 424), XSTR( "&Abort mission", 425),
 								XSTR( "Warning\nYou have no control bound to the action \"%s\".  You must do so before you can continue with your training.", 426),
 								XSTR(Control_config[Failed_key_index].text.c_str(), CONTROL_CONFIG_XSTR + Failed_key_index));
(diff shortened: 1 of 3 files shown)


code/mission/missiontraining.cpp
```diff
@@ -658,7 +658,8 @@ SCP_string message_translate_tokens(const char *text)
 				ptr = translate_key(temp);  // try and translate key
 				if (ptr != nullptr) {  // was key translated properly?
 					if (!stricmp(ptr, NOX("none")) && (Training_bind_warning != Missiontime)) {
-						if ( The_mission.game_type & MISSION_TYPE_TRAINING ) {
+						// check if a warning message should be displayed if the key is unbound
+						if ( (The_mission.game_type & MISSION_TYPE_TRAINING) || (Always_warn_player_about_unbound_keys & MISSION_TYPE_SINGLE) ) {
 							r = popup(PF_TITLE_BIG | PF_TITLE_RED, 2, XSTR( "&Bind Control", 424), XSTR( "&Abort mission", 425),
 								XSTR( "Warning\nYou have no control bound to the action \"%s\".  You must do so before you can continue with your training.", 426),
```
JohnAFernandez, line 664: Super annoying, I know but if they're in a regular mission, it's probably weird to have it say, "continue with your training."  Is there a way to add a separate `XSTR` for regular single player missions?

JohnAFernandez (changes requested):

code/mission/missiontraining.cpp
wookieejedi: Yeah, that's something I was thinking about as well. There's a few options I was thinking of:
1) Modders can set their own string in tbm of the strings table.
2) Add a new string/XSTR, though I reckon that it should have the base translations added too(?)
3) Change the existing string in the default tables.

JohnAFernandez: Changing the existing string means we invalidate additional translations that have happened since retail, right?

Adding the new string with corresponding base translations seems to be the best option to me, even though it is more work.  But what do others think?

wookieejedi: `Adding the new string with corresponding base translations seems to be the best option to me, even though it is more work.` Yeah, I agree that's probably the best bet.

JohnAFernandez: So, after a quick chat on discord, we decided option 1 is better for now.

JohnAFernandez (approved):

PR #4349, opened 2022-06-04
Multi Bump for 22.2 release

JohnAFernandez:
Instead of having three separate multiplayer version bumps, just merge this the day that the three pr's are merged.  It makes things easier to handle if we need to bisect.  

@notimaginative 's idea.


z64555:
Please reference the "three PR's" here by number.

notimaginative:
The three PRs are #4144, #4264, and #4306.

Goober5000 (approved):

Issue #3812, opened 2021-12-07
FR: A way to prevent aspect lock on a per-ship basis.


MjnMixael:
I could use a way to prevent aspect lock (notably not target lock!) for specific ships. Ideally this would be an alter-ship-flag option. (Also note that no-secondary-lock-on prevents a ship from locking on to all ships, so it's not the same as what I'm looking for.)

Bonus points for an IFF flag that applies this to all ships with that IFF by default (even when they are switched to that IFF mid-mission). But that's just a QoL bonus objective.

PR #4351, opened 2022-06-04
Add new ship flag Aspect_immune

Kiloku:
This is for #3812. 
I tested it by setting and unsetting the flag through `alter-ship-flag` on ships (both AI and the player). As far as I can tell, this works fine, doesn't interfere with the player's heatseekers, and doesn't stop the AI from fighting with their primaries. 

I couldn't test whether this interferes the AI's heatseekers due to qtFRED not having a loadout editor (as I don't have access to a Windows machine). 
It does successfully stop the AI fighter from using their aspect missiles (which are the default secondaries on the Maras which I used as the enemies in the test).


Kiloku:
Converted to draft as I accidentally forgot to merge upstream on my fork

PR #4352, opened 2022-06-04
 Add new ship flag Aspect_immune

Kiloku:
This is for https://github.com/scp-fs2open/fs2open.github.com/issues/3812.
I tested it by setting and unsetting the flag through alter-ship-flag on ships (both AI and the player). As far as I can tell, this works fine, doesn't interfere with the player's heatseekers, and doesn't stop the AI from fighting with their primaries.

I couldn't test whether this interferes the AI's heatseekers due to qtFRED not having a loadout editor (as I don't have access to a Windows machine).
It does successfully stop the AI fighter from using their aspect missiles (which are the default secondaries on the Maras which I used as the enemies in the test).

(Resubmitted PR because I accidentally overwrote other changes due to my fork being outdated)


wookieejedi (commented):
This is a super nitpicky thing, but to match the surrounding comments it would be good to have a space after the `//` and before the comment in the code.

wookieejedi (commented):
Looks overall good to me. Would recommend waiting at least a day or so for other to chime in, too.

JohnAFernandez (approved):
Nothing obviously wrong

PR #4354, opened 2022-06-05
save PXO stats to local pilot file

notimaginative:
Retail saved tracker stats to the local pilot file after read, however the newer pilot file code doesn't work the same way. This adds the ability to properly update the pilot file with multiplayer stats from the tracker during login.

NOTE: This will always overwrite multi stats in the pilot file with the stats from the tracker!

### DIFF
--- a/code/network/multi_fstracker.cpp
+++ b/code/network/multi_fstracker.cpp
@@ -28,6 +28,7 @@
 #include "network/stand_gui.h"
 #include "network/multi_pmsg.h"
 #include "playerman/player.h"
+#include "pilotfile/pilotfile.h"
 
 // -----------------------------------------------------------------------------------
 // FREESPACE MASTER TRACKER DEFINES/VARS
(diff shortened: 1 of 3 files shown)


JohnAFernandez (approved):
I trust that it's properly handled, but how does it handle stats from multiple games/mods?  Are they all tracked separately and then saved separately on the locally?

notimaginative:
> I trust that it's properly handled, but how does it handle stats from multiple games/mods? Are they all tracked separately and then saved separately on the locally?

I initially had a comment about this in there but removed it since it's a bit complicated. It's really kind of a mess and I just went with what I believe is the least crazy way to have it work.

Basically the local stats will just be what the tracker sends back, 100%. So what the tracker sends will completely replace what's in the local multi stats. This will generally match whatever mod is in use, since that's the pilot data that the tracker should be dealing with. But only the current/last used mod will have it's stats in the pilot file. It doesn't support mixing stats for multiple mods or for PXO and non-PXO stats.

For multiple mods there is really no way for the tracker to deal with this. If mod ident works properly then the correct pilot stats will be sent back. If mod ident fails then it sends the default pilot stats back. But there really isn't a way to sanity check this against local pilot data, and ship kills and medals on the tracker don't contain enough info to merge in multiple mods the way pilot files support.

In the case of mixing PXO and non-PXO stats, the tracker will always replace the local stats. But non-PXO games will update local stats, even if they were originally provided by PXO. Switching back to PXO mode will then overwrite those stats with what the tracker provides. We would have to add a separate stats section to the pilot file in order for both PXO stats and non-PXO to be saved. And since that breaks compatibility, I avoided that as an option.

JohnAFernandez:
Yeah, stats really are a mess.

PR #4355, opened 2022-06-05
Convert some Lua hooks to the new style to add documentation

Goober5000:
This takes a few of the existing hooks and converts them to the new style of hook definition which can include more extensive documentation and enumeration of supported hook variables.

This is an updated version of #3151 to work with the latest codebase.

NOTE: Before this can be integrated into the codebase, it needs an implementation for `bool script_state::IsActiveAction(int hookId)` which apparently only has support for old-style script hooks.

### DIFF
--- a/code/scripting/scripting.cpp
+++ b/code/scripting/scripting.cpp
@@ -1197,25 +1197,21 @@ void script_state::AddGameInitFunction(script_function func) { GameInitFunctions
 // This allows us to avoid significant overhead from checking everything at the potential hook sites, but you must call
 // AssayActions() after modifying ConditionalHooks before returning to normal operation of the scripting system!
 void script_state::AssayActions() {
-	for (bool &i : ActiveActions) {
-		i = false;
-	}
+	ActiveActions.clear();
 
 	for (const auto &hook : ConditionalHooks) {
 		for (const auto &action : hook.Actions) {
-			if (action.action_type >= -1 && action.action_type <= CHA_LAST) {
-				ActiveActions[action.action_type] = true;
-			}
+			ActiveActions[action.action_type] = true;
 		}
 	}
 }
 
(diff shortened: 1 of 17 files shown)


code/scripting/scripting.cpp
```diff
@@ -1212,6 +1214,10 @@ bool script_state::IsActiveAction(ConditionalActions action_id) {
 	return ActiveActions[action_id];
 }
 
+bool script_state::IsActiveAction(int hookId) {
+	// TODO
```
asarium, line 1218: Support for this will require making `ActiveActions` dynamic, e.g. via a unorder map. The hook id is always equal or greater than `CHA_LAST` so they will not conflict with the built-in hook ids.

asarium (approved):

code/scripting/scripting.cpp
```diff
@@ -1197,25 +1197,23 @@ void script_state::AddGameInitFunction(script_function func) { GameInitFunctions
 // This allows us to avoid significant overhead from checking everything at the potential hook sites, but you must call
 // AssayActions() after modifying ConditionalHooks before returning to normal operation of the scripting system!
 void script_state::AssayActions() {
-	for (bool &i : ActiveActions) {
-		i = false;
+	for (auto &entry : ActiveActions) {
```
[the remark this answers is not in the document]
Goober5000: Good point

z64555:
Was `bool script_state::IsActiveAction(int hookId)` implemented?

Goober5000:
> Was `bool script_state::IsActiveAction(int hookId)` implemented?

Yes, with modifications for the unordered map.

code/scripting/scripting.cpp
Goober5000: Done

PR #4357, opened 2022-06-05
additional checks for FS1 message conversion

### LINKED ISSUE #4356: FS1 message missing audio in FSPort's First Strike
As reported in Discord:
fsport-first-strike-missing-first-comm-audio

This is caused by a message being converted to Command when no audio existed for it.

Goober5000:
Before setting the `MQF_CONVERT_TO_COMMAND` flag, check to make sure the conversion will succeed.

This should ensure proper behavior for any case where command-conversion is needed.  I've tested it with both a message that should not be converted and a message that should, and both behave as expected.

Follow-up to #4150 and #4078.  Fixes #4356.

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -1046,6 +1046,13 @@ bool message_filename_has_fs1_wingman_prefix(const char *filename)
 		&& filename[2] != '\0');
 }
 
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
LGTM

PR #4359, opened 2022-06-06
Small fix for music initialization

Baezon:
Follow-up for #3879 particularly this commit 0dc3c914d920c11852000ff7a392f2a203e52579. Simply re-adds the various checks that the delay time had to go through before being set. This fixes a bug where music would not play on a mission where a mission without music was played previously. The logic causing that exact symptom gets kind of convoluted, but the basic cause is not checking the same conditions before setting the timestamp.

Unfortunately, I don't think this is the cause of a related bug where all music stops in a long play session. This one "fixes itself" after a single bugged music mission.


Goober5000 (approved):
Weird but ok.👍

PR #4363, opened 2022-06-07
Clean up list modifier code

jg18:
Make these changes to the list modifier code:
1. Get rid of a ton of code duplication.
2. Detect/prevent concurrent container modification (see PR #4330).


Goober5000 (approved):
Looks good 👍

jg18:
For the record, since this PR added detection for concurrent modification with mutating list modifiers, it's also part of Issue #4298.

PR #4370, opened 2022-06-08
Do our best to avoid invalid types in obj_team

JohnAFernandez:
An assert in `obj_team` is triggered very rarely, and we have yet to get a minidump or catch it in a debugger.  So I went through all the instances of `obj_team` and I tried to add checks or asserts that incoming objects were in fact what they should be (mostly ships).

There was one specific instance where already existing assertions should have been run before checking the object's team, on a code path having to do with friendly fire, which I hope is the cause.  Since I had to change that logic around, the indentations change and this PR looks a lot larger than it is.

In draft as it waits for testing, to make sure I didn't go too strict on the assertions.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -14487,6 +14487,10 @@ void ai_frame(int objnum)
 {
 	// Cyborg - can you believe this was added in *2022*??
 	Assertion (Objects[objnum].type == OBJ_SHIP, "Ai_frame was called for a non-ship of type %d. Please report!", Objects[objnum].type);
+	
+	if (Objects[objnum].type != OBJ_SHIP){
+		return;
+	}
 
 	ship		*shipp = &Ships[Objects[objnum].instance];
 	ai_info	*aip = &Ai_info[shipp->ai_index];
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -15107,10 +15107,9 @@ void init_aip_from_class_and_profile(ai_info *aip, ai_class *aicp, ai_profile_t 
 
 void ai_do_default_behavior(object *obj)
 {
-    Assert(obj != nullptr);
-    Assert(obj->instance != -1);
-    Assert(obj->type == OBJ_SHIP);
-    Assert(Ships[obj->instance].ai_index != -1);
+    Assertion(obj != nullptr, "Ai_do_default_behavior() was given a bad information.  Specifically obj is a nullptr. This is a coder error, please report!");
+    Assertion((obj->type == OBJ_SHIP) && (obj->instance > -1) && (obj->instance < MAX_SHIPS), "This function that requires an object to be a ship was passed an invalid ship. Please report\n\nObject type: %d\nInstance #: %d", obj->type, obj->instance);
+    Assertion(Ships[obj->instance].ai_index > 0, "A ship has been found to have an invalid ai_index of %d. This should have been assigned and is coder mistake, please report!", Ships[obj->instance].ai_index);
 
     ai_info	*aip = &Ai_info[Ships[obj->instance].ai_index];
     ship_info* sip = &Ship_info[Ships[obj->instance].ship_info_index];
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -16155,7 +16155,7 @@ void maybe_cheat_fire_synaptic(object *objp)
 		ship	*shipp;
 		int	wing_index, time;
 
-		Assertion(objp->type == OBJ_SHIP, "maybe_cheat_fire_synaptic was called with a non-ship object type of %d. Please report!", objp->type);
+		Assertion(objp->type == OBJ_SHIP, "This ai function requires a ship, but was called with a non-ship object type of %d. Please report!", objp->type);
 
 		shipp = &Ships[objp->instance];
 
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -15549,11 +15549,14 @@ void ai_ship_hit(object *objp_ship, object *hit_objp, vec3d *hit_normal)
 			return;
 		
 		hitter_objnum = hit_objp->parent;
-		Assert((hitter_objnum >= 0) && (hitter_objnum < MAX_OBJECTS));
+		Assertion((hitter_objnum >= 0) && (hitter_objnum < MAX_OBJECTS), "hitter_objnum in this function is an invalid index of %d.  This can cause random behavior, and is a coder mistake.  Please report!", hitter_objnum);
 		objp_hitter = &Objects[hitter_objnum];
 
+		Assertion((objp_hitter->type == OBJ_SHIP) || (objp_hitter->type == OBJ_WEAPON) || (objp_hitter->type == OBJ_ASTEROID) || (objp_hitter->type == OBJ_BEAM) || (objp_hitter->type == OBJ_DEBRIS),
+		"The ai code is passing an invalid object type of %d to a function when trying to decide how to respond to something hitting it.  This is a bug in the code, please report!", objp_hitter->type);
+
 		// lets not check hits by ghosts any further either
-		if(objp_hitter->type != OBJ_SHIP && objp_hitter->type != OBJ_WEAPON && objp_hitter->type != OBJ_ASTEROID && objp_hitter->type != OBJ_BEAM)
+		if(objp_hitter->type != OBJ_SHIP && objp_hitter->type != OBJ_WEAPON && objp_hitter->type != OBJ_ASTEROID && objp_hitter->type != OBJ_BEAM && objp_hitter->type != OBJ_DEBRIS)
 			return;
 		
 		//	Hit by a protected ship, don't attack it.
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -2387,6 +2387,7 @@ void ai_attack_object(object* attacker, object* attacked, int ship_info_index)
 
 	Assert(attacker != nullptr);
 	Assert(attacker->instance >= 0);
+	Assert(attacker->type == OBJ_SHIP);
 	Assert(Ships[attacker->instance].ai_index != -1);
 	//	Bogus!  Who tried to get me to attack myself!
 	if (attacker == attacked) {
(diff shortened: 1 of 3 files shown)


JohnAFernandez:
This got a sanity test, and it seems to be performing normally.  I'm marking as ready for review, but as an AI change, it should probably be given some extra attention during review.

code/ai/aicode.cpp
```diff
@@ -15541,7 +15553,7 @@ void ai_ship_hit(object *objp_ship, object *hit_objp, vec3d *hit_normal)
 		objp_hitter = &Objects[hitter_objnum];
 
 		// lets not check hits by ghosts any further either
-		if(objp_hitter->type == OBJ_GHOST)
+		if(objp_hitter->type != OBJ_SHIP && objp_hitter->type != OBJ_WEAPON && objp_hitter->type != OBJ_ASTEROID && objp_hitter->type != OBJ_BEAM)
```
Goober5000, line 15556: `OBJ_DEBRIS` is also a possibility.  But I would add an assert here as nothing that isn't those object types has any business getting to this point.

code/ai/aicode.cpp
```diff
@@ -10006,7 +10007,7 @@ void guard_object_was_hit(object *guard_objp, object *hitter_objp)
 			aip->submode = SM_ATTACK;
 			aip->submode_start_time = Missiontime;
 			aip->active_goal = AI_ACTIVE_GOAL_DYNAMIC;
-		} else {
```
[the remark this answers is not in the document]
JohnAFernandez: `num_ships_attacking` calls `obj_team`. This seemed like the best place to head it off.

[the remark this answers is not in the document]
Goober5000: But `obj_team` is able to work with objects of other types, and it's possible a ship could be chasing another object, like an asteroid or weapon.

This PR should not be focused *per se* on catching objects that aren't ships, but rather on catching `OBJ_NONE`.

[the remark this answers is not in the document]
JohnAFernandez: Then we're going to be stuck back here in 5 years doing the exact same thing.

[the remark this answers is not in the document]
Goober5000: Looking at the actual code, I see that this `else` belongs with `if (aip->target_objnum == -1)`, not with `else if (aip->previous_mode == AIM_GUARD)` which was my impression from github.  So I'll retract my comment - that change looks fine.

code/ai/aicode.cpp
```diff
@@ -10006,7 +10007,7 @@ void guard_object_was_hit(object *guard_objp, object *hitter_objp)
...
+		} else if (Objects[aip->target_objnum].type == OBJ_SHIP) {
```
jg18, line 10010: Do we need a new `else` clause? What should happen if neither `if`-condition applies?

code/ai/aicode.cpp
```diff
@@ -15201,118 +15206,125 @@ void maybe_process_friendly_hit(object *objp_hitter, object *objp_hit, object *o
...
+	// Cyborg -- I had to switch around the logic here because it was potentially
+	// sending non-ships to obj_team, which would trigger an assert
+	if (objp_hitter == Player_obj){
 
-		// AL 12-4-97: It is possible the Player is a OBJ_GHOST at this point.  If so, bail out.
+		// Cyborg - So check if this is a ship first. Probably true, but you never know, with FSO
+		// (especially in multi)
 		if ( objp_hitter->type != OBJ_SHIP ) {
 			return;
 		}
```
jg18, line 15217: Is it expected that this function may get called on a non-ship object? If Or would that indicate a bug?

code/ai/aicode.cpp
```diff
@@ -15101,8 +15105,9 @@ void init_aip_from_class_and_profile(ai_info *aip, ai_class *aicp, ai_profile_t
 
 void ai_do_default_behavior(object *obj)
 {
-    Assert(obj != NULL);
+    Assert(obj != nullptr);
     Assert(obj->instance != -1);
+    Assert(obj->type == OBJ_SHIP);
```
jg18, line 15110: Perhaps make any new assertions `Assertion()` rather than `Assert()`?

code/ai/aicode.cpp
```diff
@@ -16139,6 +16151,8 @@ void maybe_cheat_fire_synaptic(object *objp)
 		ship	*shipp;
 		int	wing_index, time;
 
+		Assertion(objp->type == OBJ_SHIP, "maybe_cheat_fire_synaptic was called with a non-ship object type of %d. Please report!", objp->type);
```
jg18, line 16154: IMHO it's nice to use assertion failure messages that explain the context in English rather than with function names,  given that it's user-facing text, but it's not essential. Striving for that has been helpful in my own PRs to ensure I understand what the assertion failing would really mean.

Ditto for other assertions this PR adds.

jg18 (commented):
A few questions/thoughts.

I like the idea of trying to avoid invalid calls to `obj_team()`, given the rare non-reproducible crashes you mentioned. I think those go at least as far back as Diaspora R1, if not more.

Tip for reviewers: checking "Hide whitespace" on the "Files changed" tab shrinks the PR massively.

code/ai/aicode.cpp
JohnAFernandez: Ok, those have been added.

code/ai/aicode.cpp
JohnAFernandez: You're right.  We should put an unreachable in an else because it *should* be a previously defended ship.  But if it isn't, it would be the source of our bug, and so it brings us one step closer to finding the source.

code/ai/aicode.cpp
JohnAFernandez: This is in a section where we are only working with the player ship.  AFAIK, In multi, `Player_obj` can be a ship, observer, or none, depending on the player status.  If the `Player_obj` is some other kind of object, then I would expect we would experience problems everywhere.  If the hitter here is the player and they player is not in their ship anymore, the AI shouldn't really act on that information anyway.

code/ai/aicode.cpp
JohnAFernandez: I think I've cleared them up.

code/ai/aicode.cpp
JohnAFernandez: Yes.

code/ai/aiturret.cpp
```diff
@@ -573,6 +573,8 @@ void evaluate_obj_as_target(object *objp, eval_enemy_obj_struct *eeo)
 			}
 		}
 
+		Assertion(Objects[eeo->turret_parent_objnum].type == OBJ_SHIP,"evaluate_obj_as_target was about to call object_is_targetable with a non-ship object.  Please report!");
```
Baezon, line 576: This is impossible, the assertion condition is the same as the if branch this is contained in. If we wanted to make sure `object_is_targetable()` is only run on ships, then that function should probably assert this, although it seems to be able to handle multiple different object types.

code/hud/hudtarget.cpp
```diff
@@ -1530,6 +1530,8 @@ void hud_target_missile(object *source_obj, int next_flag)
 				continue;
 			}
 
+			Assertion(Objects[so->objnum].type == OBJ_SHIP, "hud_target_missile was about to call obj_team with a non-ship obejct with type %d. Please report!", Objects[so->objnum].type);
```
Baezon, line 1533: And here, `continue` if not.

code/ai/aicode.cpp
```diff
@@ -14482,6 +14485,9 @@ void validate_mode_submode(ai_info *aip)
  */
 void ai_frame(int objnum)
 {
+	// Cyborg - can you believe this was added in *2022*??
+	Assertion (Objects[objnum].type == OBJ_SHIP, "Ai_frame was called for a non-ship of type %d. Please report!", Objects[objnum].type);
```
Baezon, line 14489: Return here as well. We're about to index into `Ships` with this thing, if it isn't a ship for whatever reason we need to immediately return.

code/ai/aicode.cpp
```diff
@@ -15101,9 +15107,9 @@ void init_aip_from_class_and_profile(ai_info *aip, ai_class *aicp, ai_profile_t
 
 void ai_do_default_behavior(object *obj)
 {
-    Assert(obj != NULL);
-    Assert(obj->instance != -1);
-    Assert(Ships[obj->instance].ai_index != -1);
+    Assertion(obj != nullptr, "Ai_do_default_behavior() was given a bad information.  Specifically obj is a nullptr. This is a coder error, please report!");
+    Assertion((obj->type == OBJ_SHIP) && (obj->instance > -1) && (obj->instance < MAX_SHIPS), "This function that requires an object to be a ship was passed an invalid ship. Please report\n\nObject type: %d\nInstance #: %d", obj->type, obj->instance);
+    Assertion(Ships[obj->instance].ai_index > 0, "A ship has been found to have an invalid ai_index of %d. This should have been assigned and is coder mistake, please report!", Ships[obj->instance].ai_index);
```
Baezon, line 15112: `>=`? Is 0 ok?

code/ai/aicode.cpp
```diff
@@ -16139,6 +16155,8 @@ void maybe_cheat_fire_synaptic(object *objp)
 		ship	*shipp;
 		int	wing_index, time;
 
+		Assertion(objp->type == OBJ_SHIP, "This ai function requires a ship, but was called with a non-ship object type of %d. Please report!", objp->type);
```
Baezon, line 16158: Same here, return if it's not a ship.

Baezon (changes requested):
Seems good to me. We do want to make sure that behavior is as smooth as possible even in release, so in addition to just asserting we should try to `return`, `continue`, whatever is least detrimental when we continue.

code/ai/aiturret.cpp
JohnAFernandez: You are correct it was redundant.  I think it's supposed to be able to handle more than one type of object, so I'm not going to make any changes there.

code/hud/hudtarget.cpp
JohnAFernandez: This one will already continue at the next if block if it is false.

code/ai/aicode.cpp
JohnAFernandez: 100%.  Making that change now.

code/ai/aicode.cpp
JohnAFernandez: Yes.  Another great catch.

code/ai/aicode.cpp
JohnAFernandez: Ok, should be done.

code/hud/hudtarget.cpp
Baezon: The one starting with `		if (!iff_x_attacks_y(Player_ship->team, obj_team(A))) {`? That will then end up running `obj_team()` on it first, will it not?

JohnAFernandez: Yeah, but on release it returns -1 for invalid objects, so the check `!iff_x_attacks_y(Player_ship->team, obj_team(A)))` where they are at war aiui returns false.

Baezon (approved):

Issue #4350, opened 2022-06-04
Linux build occassionally fails in CI


z64555:
We've been seeing this more recently.  Essentially, one or multiple builds on the Linux target fails.

My investigation led me to a failure by `cp`, a copy command for bash, stating that it could not copy a .so file because it already exists in the destination, which appears to be nonsense.

The line that this call to `cp` happens in is rather long, making commandline length limit highly probable.  The linux length limit of 4096 characters coincidentally has it run this as its last command on the line: ```cp -a /__w/fs2open.github.com/fs2open.github.com/build/lib```

Further investigation revealed that `cp` is being called by `ninja`, and the `build.ninja` file its running on is from `cmake`, but that's about as far as I got.  I suspect there's a configuration not quite right in cmake that's leading to the symptom.

z64555:
IIRC I had stumbled across this issue very recently when messing around with the pi-pico.  The solution with it was to update cmake, so this is a possible solution for this situation as well.

z64555:
The CMake versions in the linux runners *should* be the latest release, so that makes it unlikely that upgrading the cmake version will resolve this.

z64555:
It's possible that there's a configuration with cmake that can limit the commandline options that it generates, but since I'm not familiar with cmake nor how to navigate its documentation I cannot make a reasonable time limit.

Since this issue is mostly an nuisance rather than a progress stopper, I'm demoting the priority and removing it from the 22.2 milestone.

notimaginative:
I haven't seen this issue happen in several years now. I have to assume that we've fixed it with standard project changes or cmake updates.

Closing for now and someone can reopen if this should get another look or it happens again.

PR #4380, opened 2022-06-10
Changelog.py improvements and post-release-build.yaml debugging

z64555:
This adds an `echo` to print out the `github.event_name` of the calling workflow, so that the Auto Trigger can be debugged. (For some reason it hasn't been firing when called by Release.yaml)

Fixes Changelog.py going into an infinite loop due to not breaking out when page = last_page + 1, issue turned out to be due to how `for .. in ..` loops differ from C++'s `for` loops.

Also makes some improvements to the Changelog.py to be more user friendly as well as allow authentication to get around the REST API HTTP requests limits.

### DIFF
--- a/ci/post/changelog.py
+++ b/ci/post/changelog.py
@@ -43,7 +43,7 @@ def get_last_page(response: requests.Response) -> int:
     
 
 def main():
-    parser = argparse.ArgumentParser(description="Retrives a log of commit messages and attempts to pretty them into a human-readbile format")
+    parser = argparse.ArgumentParser(description="Retrieves a log of commit messages and attempts to pretty them into a human-readable format")
     parser.add_argument("start_date",
                         help="Day of the earliest commit to add to the log.")
     parser.add_argument("-u", "--user", nargs=2, metavar=("NAME", "TOKEN"), default=["", ""],
--- a/ci/post/changelog.py
+++ b/ci/post/changelog.py
@@ -86,7 +86,7 @@ def main():
     url = "https://api.github.com/repos/scp-fs2open/fs2open.github.com/pulls"
     page = 1
     last_page = MAX_PAGES
-    while page < last_page + 1:
+    while page <= last_page:
         # GET <url>?base=master&sort=updated&direction=desc&state=closed&page=1
         response = requests.get(url,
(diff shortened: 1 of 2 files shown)


ci/post/changelog.py
```diff
@@ -76,10 +84,13 @@ def main():
 
     changelog = {}
     url = "https://api.github.com/repos/scp-fs2open/fs2open.github.com/pulls"
+    page = 1
     last_page = MAX_PAGES
-    for page in range(1, last_page + 1):
+    while page < last_page + 1:
```
MageKing17, line 89: I think it would be clearer like this:
```suggestion
    while page <= last_page:
```

ci/post/changelog.py
```diff
@@ -42,9 +43,16 @@ def get_last_page(response: requests.Response) -> int:
     
 
 def main():
+    parser = argparse.ArgumentParser(description="Retrives a log of commit messages and attempts to pretty them into a human-readbile format")
```
MageKing17, line 46:
```suggestion
    parser = argparse.ArgumentParser(description="Retrieves a log of commit messages and attempts to pretty them into a human-readable format")
```

MageKing17 (changes requested):
Couple typos in the description, but otherwise changes look good.

MageKing17 (approved):
No idea what this particular CI failure means, but the code seems fine.

z64555:
> No idea what this particular CI failure means, but the code seems fine.

Hmm, I'm not seeing any failures on the PR checks. Where do you see the failures?

MageKing17:
> > No idea what this particular CI failure means, but the code seems fine.
> 
> Hmm, I'm not seeing any failures on the PR checks. Where do you see the failures?

Someone must've re-ran it... it was a strange failure about being unable to create a symlink, I think; hopefully just a fluke.

z64555:
> > Hmm, I'm not seeing any failures on the PR checks. Where do you see the failures?
> 
> Someone must've re-ran it... it was a strange failure about being unable to create a symlink, I think; hopefully just a fluke.

linux gcc-10 has been failing at random but frequent times due to a "file exists" error.  See #4350

Issue #2351, opened 2020-04-29
Feature Request: Custom Cheat


MjnMixael:
Here's a fun request to work on amidst all the bug fixes...

I'd like to be able to define a custom cheat to work in addition to arrrrwalktheplank. Define a custom cheat string and a ship to summon. The rest works the same as it does with the Voltion Bravos.

My end goal is to include the HLP Bravos in FSU with a cheat to summon it. I'm thinking Game_settings.tbl is a good candidate for the definition locations?

PR #4387, opened 2022-06-11
Cheat framework allowing for custom cheats, sexp for cheat used and On Cheat lua hook

Kiloku:
I saw #2351 and went a bit overboard. Blame covid, I got bored on sick leave.

This brings a new table type (cheats.tbl and *-cht.tbm). Adding a ship to spawn is optional, as a modder might want to just use the Sexp or lua hook to check if the player used the cheat.

The table looks like this:
```
#CUSTOM CHEATS

$Cheat:
	+Code: puppyy
	+Message: It followed me home
	+SpawnShip: 
		+Name: Puppy
		+Class: SJ Sathanas
$Cheat:
	+Code: foo
	+Message: bar
	+RequireCheats: YES

#END
```
If a cheat has `RequireCheats` as true, then the player needs to enable cheats first, by typing www.freespace2.com (Now that I think about it, maybe I should add the option of using another string for that too...)
The table parsing checks if the ship class exists, and pops up a warning if it doesn't, but allows the game to continue. 

I had to mess with the original cheat code a bit, the biggest change being that now the Bravos was moved out of that list into the new framework. All original cheat codes still function as usual, as far as I can tell.


Kiloku:
A quick re-read makes me realize that his needs a few cleanups. Bad variable names, commented out code from previous attempts that should just be removed, probably more. I'll change it into a draft for now.

Kiloku:
Requested by MediaVPs to add the HLP Bravos!

code/cheats_table/cheats_table.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "parse/parselo.h"
+
+void cheat_table_init() {
+	customCheats.clear();
+
+	// Add default Spawn Ship Cheat (Volition Bravos).
+	// Won't interfere with Total Conversions, since if the ship doesn't exist, the game will just ignore the cheat.
+	// The other default cheats do not use the Custom Cheat Framework and are still handled in keycontrol.cpp
+	std::unique_ptr<CustomCheat> bravos(new SpawnShipCheat("arrrrwalktheplank", "Walk the plank", false, "Volition Bravos", "Volition Bravos"));
+	customCheats.emplace(SCP_string(bravos->cheatCode), std::move(bravos));
```
BMagnu, line 11: instead of hard-defining the cheat like this here, could I suggest a similar method as with default game_settings.tbl / fonts.tbl / etc,, located in code/def_files/data/tables.
This would allow any mod to override otherwise built-in cheats or just outright remove them altogether.

code/cheats_table/cheats_table.cpp
```diff
@@ -0,0 +1,110 @@
...
+			if (optional_string("+RequireCheats:")) {
+				stuff_boolean(&requireCheats);
+			}
+
+			if (optional_string("+SpawnShip:")) {
+				shipSpawn = true;
+				required_string("+Name:");
+				stuff_string(shipName, F_RAW);
+				required_string("+Class:");
+				stuff_string(shipClass, F_RAW);
```
BMagnu, line 50: These two should probably read an ``F_NAME`` instead of an ``F_RAW``

code/cheats_table/cheats_table.cpp
```diff
@@ -0,0 +1,110 @@
...
+			stuff_string(code, F_RAW);
+
+			//We are limited to the buffer size used for the traditional cheats. We can still use short cheats, though.
+			if (code.length() > CHEAT_BUFFER_LEN) {
+				Warning(LOCATION, "Cheat code %s is too long. It will be cut off to maximum length (%i characters).", code.c_str(), CHEAT_BUFFER_LEN);
+				code = code.substr(0, CHEAT_BUFFER_LEN);
+			}
+
+			required_string("+Message:");
+			stuff_string(msg, F_RAW);
```
BMagnu, line 39: Read an ``F_MULTITEXT`` here, maybe?

code/cheats_table/cheats_table.cpp
```diff
@@ -0,0 +1,110 @@
...
+	memset (check_buffer, 0, buffer_length * sizeof(char));
+
+	// We construct a checking buffer from the original.
+	int idx = 0;
+	for (int i = 0; i < buffer_length; i++) {
+		if (buffer[i] != '\0') {
+			check_buffer[idx] = buffer[i];
+			idx++;
+		}
+	}
```
BMagnu, line 98: So, why do you copy the string here?
If you strstr on buffer directly, it should work.
Also, for type safety's sake, you should pass buffer as a const char* or const char[], since you really don't want to modify it in this method

code/cheats_table/cheats_table.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+	// Loop through our map of custom cheats and check if the code is within the buffer.
+	// We don't just check for equality, as the player may have typed other characters before the cheat.
+	for (auto &ccheat : customCheats) {
+		auto found = std::strstr(check_buffer, ccheat.first.c_str());
+		if (found) {
```
BMagnu, line 104: we try to avoid doing if(ptr) direcly. It'd be better to have an explicit found != nullptr

code/cheats_table/cheats_table.h
```diff
@@ -0,0 +1,124 @@
...
+	public:
+		SCP_string cheatCode;
+		SCP_string cheatMsg;
+		bool requireCheatsEnabled;
+
+		CustomCheat(SCP_string cheat_code, SCP_string cheat_msg, bool require_cheats_enabled) {
+			cheatCode = cheat_code;
+			cheatMsg = cheat_msg;
+			requireCheatsEnabled = require_cheats_enabled;
+		}
```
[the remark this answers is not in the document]
BMagnu: Also, a list initalization would be a lot cleaner here:
```suggestion
		CustomCheat(SCP_string cheat_code, SCP_string cheat_msg, bool require_cheats_enabled) : 
			cheatCode(std::move(cheat_code)),
			cheatMsg(std::move(cheat_msg)),
			requireCheatsEnabled(require_cheats_enabled) { }
```

code/cheats_table/cheats_table.h
```diff
@@ -0,0 +1,124 @@
...
+
+		virtual ~CustomCheat() = default;
+
+		virtual void runCheat() {
+			if (canUseCheat()) {
+				HUD_printf("%s", cheatMsg.c_str());
+				scripting::hooks::OnCheat->run(scripting::hook_param_list(scripting::hook_param("Cheat", 's', cheatCode)));
+				CheatUsed = SCP_string(cheatCode);
```
BMagnu, line 32: this SCP_string() is not needed. Since CheatUsed is not a pointer, it will be automatically copied and does not need this manual object-copy

code/cheats_table/cheats_table.h
```diff
@@ -0,0 +1,124 @@
...
+
+class SpawnShipCheat : public CustomCheat {
+	protected:
+	SCP_string shipClassName;
+	SCP_string shipName;
+	public:
+
+	SpawnShipCheat(SCP_string cheat_code, SCP_string cheat_msg,  bool require_cheats_enabled, SCP_string class_name, SCP_string ship_name) : CustomCheat(cheat_code, cheat_msg, require_cheats_enabled) {
+		shipClassName = class_name;
+		shipName = ship_name;
```
BMagnu, line 49: Same thing with list initialization and std::move as above

code/cheats_table/cheats_table.h
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	void runCheat() override {
```
BMagnu, line 52: This method, and really, all methods of these classes, should be in the cheats_table.cpp and not the .h

code/parse/sexp.cpp
```diff
@@ -37488,6 +37510,12 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"Builtin languages are English, German, French, and Polish, and others can be defined in strings.tbl.\r\n"
 	},
 
+	{ OP_USED_CHEAT, "used-cheat\r\n"
+		"\tReturns true if the give cheat has been used during this mission.  Takes 1 argument.\r\n"
```
BMagnu, line 37514: _And_ there has not been another cheat used since. The SEXP can only test for the last cheat used, and this documentation should account for that

BMagnu (changes requested):

code/cheats_table/cheats_table.cpp
Kiloku: This causes the default parsing to fail (seems that it needs a specific terminator in the table). On the other hand, there's an `F_MESSAGE`, which works fine

BMagnu: ``F_MESSAGE`` is obsolete though. I wonder what the proper course of action is here.

Goober5000: It's obsolete for in-mission messages.  It's still used for a few other things though.  I'd be ok with using it.

Goober5000: You could also use F_NAME or F_LNAME, depending on whether you wanted localization.

code/cheats_table/cheats_table.cpp
Kiloku: This worked. I don't know exactly why I did it that way, but I do remember that I had a hard time and got really confused building this function 😅

BMagnu: My guess is that it has to do with the null chars in the beginning of the cheat code string.
Try this, plus make ``buffer`` into a const char*.
This'll scroll check_buffer to the start of the buffer without needlessly copying it.

```suggestion
	const char* check_buffer = buffer;
	for (int i = 0; i < buffer_length; i++) {
		if (*check_buffer == '\0') {
			check_buffer++;
		}
		else {
			break;
		}
	}
```

Kiloku: Ok, no, it didn't work. I remember now. Just passing the buffer doesn't work because it'll stop at the null terminator if the buffer is not full, so a cheat that's smaller than 17 characters (the max cheat length) won't work until 17 keypresses have been registered. 

I'm thinking we can just initialize the Cheat Buffer to be a string of 17 spaces or some other character.

BMagnu: Either that, or you do this loop which will scroll forward to the first non-null char in the buffer

Kiloku: Now I got it right. Thanks!

code/io/keycontrol.cpp
```diff
@@ -1582,15 +1583,28 @@ void game_process_cheats(int k)
 	
 	cheatCode detectedCheatCode = CHEAT_CODE_NONE;
 
+	extern bool checkForCustomCheats(char converted_buffer[], int buffer_length);
```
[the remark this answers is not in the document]
Kiloku: IIRC I was having trouble with reference loops in the headers at some point, this must be a relic of that. Fixed.

code/cheats_table/cheats_table.cpp
```diff
@@ -0,0 +1,181 @@
...
+		}
+		required_string("#END");
+	}
+	catch (const parse::ParseException& e) {
+		mprintf(("TABLES: Unable to parse '%s'!  Error message = %s.\n", filename, e.what()));
+		return;
+	}
+}
+
+bool checkForCustomCheats(char buffer[], int buffer_length) {
```
BMagnu, line 160: This should be passed a const char*, not a char[]

BMagnu (changes requested):
One tiny thing is left

BMagnu (approved):

PR #4389, opened 2022-06-11
Remove a pointless assert

JohnAFernandez:
There is no point in putting a define in an assert to make sure it's greater than zero.  Just put a comment or use a `static_assert` if you're especially paranoid.


Goober5000 (approved):
Shrug of approval

PR #4391, opened 2022-06-11
fix use of default tech database flags

### LINKED ISSUE #4390: Removing an entry from the Tech Database through lua doesn't work if the flag is set to true in the table
`shipClass.InTechDatabase` (and others like weaponInfo) can be set by lua scripts. It succesfully changes the corresponding flag in the engine (such as `Ship::Info_Flags::In_tech_database`) to the new value. 
However, if the new value is `false` while the original value set in the table is `true`, then the entry will still show in the tech lab. 
This is because the code that fills the techroom list checks both `Ship::Info_Flags::In_tech_database` and `Ship::Info_Flags::Default_in_tech_database` (which is set by the table), adding the entry if *any* of the two flags are true for that ship. 

Since lua can't change `Ship::Info_Flags::Default_in_tech_database`, lua's `shipClass.InTechDatabase` is only useful for adding new entries, but not for removing default entries from the list. 

I believe this is not by design, but I wanted to ask before attempting to change it.

Goober5000:
The "default" tech database flags indicate whether entries are in the tech room by default, that is, when the table is originally parsed and the campaign is started.  Flags may be changed later, so that when the tech database is "reset to default" the flags can be returned to the known state as specified in the tables.

A bug was introduced back in commit c924e3997a (!) based on a misinterpretation of what "default" meant.  This fixes the bug and clarifies the flags.

Fixes #4390.

### DIFF
--- a/code/menuui/techmenu.cpp
+++ b/code/menuui/techmenu.cpp
@@ -760,7 +760,6 @@ void techroom_change_tab(int num)
 	switch (Tab) {
 		case SHIPS_DATA_TAB:
             si_mask.set(multi ? Ship::Info_Flags::In_tech_database_m : Ship::Info_Flags::In_tech_database);
-            si_mask.set(multi ? Ship::Info_Flags::Default_in_tech_database_m : Ship::Info_Flags::Default_in_tech_database);
 			
 			// load ship info if necessary
 			if ( !Ships_loaded ) {
@@ -809,7 +808,6 @@ void techroom_change_tab(int num)
 				Weapon_list.reserve(Weapon_info.size());
 
 				wi_mask.set(multi ? Weapon::Info_Flags::Player_allowed : Weapon::Info_Flags::In_tech_database);
-                wi_mask.set(Weapon::Info_Flags::Default_in_tech_database);
 
 				int i = 0;
 				tech_list_entry temp_entry;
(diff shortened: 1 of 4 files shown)


JohnAFernandez (approved):

notimaginative:
I don't remember the exact details, but I believe that change was actually intended to fix a bug related to something with pilot files (mod switching or related??). I do agree that the fix itself is a bug though. I wish I could remember what issue it was supposed to fix so we'd know to look for it if it pops up again. There's probably a comment thread buried somewhere in the HLP forums about it.

Goober5000:
It could have been related to the fact that the pilot file needs to store a tri-state but the tech db flag is a boolean.  The pilot file code has been rewritten since then, of course, but this new code looks relevant:

```
	// write whether it differs from the default for ships
	for (auto &si : Ship_info) {
		if (si.flags[Ship::Info_Flags::In_tech_database] == si.flags[Ship::Info_Flags::Default_in_tech_database]) {
			state = TechroomState::DEFAULT;
		} else if (si.flags[Ship::Info_Flags::In_tech_database]) {
			state = TechroomState::ADDED;
		} else {
			state = TechroomState::REMOVED;
		}

		cfwrite_ubyte((ubyte) state, cfp);
	}
```

notimaginative:
Quite possible. The initial "fix" was done after the first big pilot file rewrite and most likely isn't something that applies to the current code. I did try to address a lot of the strange issues that popped up before/after the first pilot file rewrite while initially working on the newer code, though I'm not sure how much of that ended up being in the final code.

But this LGTM.

Issue #4335, opened 2022-05-31
Crash in loadout screen when clicking a weapons-locked ship


Baezon:
Weapons locked *only*, that is. At `missionshipchoice.cpp:2126` the case where a slot is only weapons locked is not handled. I tried fixing this, but I only broke/exposed more places that don't handle a ship being only weapons locked, and not ship locked. It also erroneously grays out weapons locked ships.

jg18:
Looks like this part of the code treats weapons-locked and ship-locked wing slots as none of empty/filled/locked, which is incorrect.

Even if a wing slot is weapons-locked, we still need to track whether the slot is empty or filled, and the code currently doesn't do this.

IMHO, slot status should obey these constraints:
1. Every slot should be exactly one of empty, filled, or locked.
2. A locked slot should never be ship-locked, weapons-locked, or the player.
3. A ship-locked slot should always be filled.

Karajorma:
That last bit of logic is correct. I never expected anyone would ever lock a ship's weapons but not the ship itself cause that seemed pretty strange. But I expected there would be plenty of FREDders who wanted to lock a ship but still allow the player to choose the weapons available on the ship.

Baezon:
Then you should have disallowed it when you implemented it! Alas, you built it, and now someone has come with a use case for it lol

jg18 said he'll post an update soon, but we discussed this a bit on the Discord; weapons-locked will attempt to fill with default weapons for the class, taking from the pool as necessary, and using <none> otherwise. I'll leave to you whether this is a good idea, but it *is* what would be most useful for me.

jg18:
After more Discord discussion, we've decided to keep weapons-locked independent of ship-locked, even if it wasn't the intended behavior.

Therefore, we're effectively going with the slot constraints I proposed above, which I've updated slightly for clarity.

To enforce these constraints in the code, I propose adding

```
enum class SlotStatus {
	Empty,
	Filled,
	Ship_Locked,
	Locked
};
```
and then replacing `ss_slot_info`'s `int status` with two fields, `SlotStatus status` and `bool weapons_locked`.

 I'd also add a new internal variable
```
ss_slot_info	*Ss_player_slot = nullptr;
```
and then update `Ss_player_slot` as each mission loads/unloads.

While it'll entail a number of code changes, it'll force us to think through what should happen under which circumstances.

Thoughts @karajorma @Baezon @notimaginative @johnafernandez?

jg18:
[...]

PR #4396, opened 2022-06-13
Clean up how wing slot status is marked

jg18:
Replace `int ss_slot_info status` with a `SlotStatus` enum class and other changes.

Known issue is that drawing the wing block uses incorrect icon coloring for selected icon. Still working on it.

Intended to fix Issue #4335.

Tip: Setting "Hide whitespace" on the "Files changed" tab shrinks the changes significantly.

### DIFF
--- a/code/missionui/missionshipchoice.cpp
+++ b/code/missionui/missionshipchoice.cpp
@@ -2157,7 +2157,7 @@ void pick_from_wing(int wb_num, int ws_num)
 //				hot_slot	=>		index of slot that mouse is over
 //				selected_slot	=>	index of slot that is selected
 //				class_select	=>	all ships of this class are drawn selected (send -1 to not use)
-void draw_wing_block(int wb_num, int hot_slot, int selected_slot, int class_select)
+void draw_wing_block(int wb_num, int hot_slot, int selected_slot, int class_select, bool ship_selection )
 {
 	GR_DEBUG_SCOPE("Wing block");
 
@@ -2274,7 +2274,7 @@ void draw_wing_block(int wb_num, int hot_slot, int selected_slot, int class_sele
 				bitmap_to_draw = Wing_slot_empty_bitmap;
 				break;
 
-			case WING_SLOT_LOCKED:
+			default:
 				if ( icon ) {
 					if(icon->model_index == -1)
 						bitmap_to_draw = icon->icon_bmaps[ICON_FRAME_DISABLED];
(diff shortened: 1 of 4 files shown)


jg18:
Tip: Setting "Hide whitespace" on the "Files changed" tab shrinks the changes significantly.

code/missionui/missionshipchoice.cpp
```diff
@@ -2181,49 +2218,22 @@ void draw_wing_block(int wb_num, int hot_slot, int selected_slot, int class_sele
...
+			bool done = false;
 
-				if ( class_select >= 0 ) {	// only ship select
-					if ( Carried_ss_icon.from_slot == slot_index ) {
-						if ( ss_carried_icon_moved() ) {
-							bitmap_to_draw = Wing_slot_empty_bitmap;
-						} else {
-							bitmap_to_draw = -1;
-						}
-						break;
```
jg18, line 2206: I've replaced this line with
```
done = true;
```

code/missionui/missionshipchoice.cpp
```diff
@@ -2181,49 +2218,22 @@ void draw_wing_block(int wb_num, int hot_slot, int selected_slot, int class_sele
...
+					if ( ss_carried_icon_moved() ) {
+						bitmap_to_draw = Wing_slot_empty_bitmap;
+					} else {
+						bitmap_to_draw = -1;
 					}
+					done = true;
 				}
+			}
 
-				if ( ws->status & WING_SLOT_LOCKED ) {
```
jg18, line 2210: Removing this whole block is surely incorrect and probably related to the visual glitches, but I haven't yet figured out what to do with this.

jg18 (commented):
Here are a couple notes for reviewers.

jg18:
Here's a [test mission](https://github.com/scp-fs2open/fs2open.github.com/files/8887828/issue_4335.fs2.txt) I've been using. I don't get any crashes when this PR is applied, but visual glitches occur on the ship select and weapons select screens.

code/missionui/missionshipchoice.cpp
jg18: Looks like the only case where a slot could be simultaneously locked and something else is in multiplayer.

Since there are already plans to overhaul this code for multi, probably best not to go the route this PR is taking. Oh, well.

jg18:
While this approach ended up looking too complicated/invasive, it paved the way for finding a simpler solution.

PR #4399, opened 2022-06-13
Handle weapons-locked slots that aren't ship-locked in selection screens.

jg18:
Attempt a simpler solution to Issue #4335 than PR #4396.

### DIFF
--- a/code/missionui/missionshipchoice.cpp
+++ b/code/missionui/missionshipchoice.cpp
@@ -2100,7 +2100,7 @@ void pick_from_wing(int wb_num, int ws_num)
 		return;
 	}
 
-	switch ( ws->status ) {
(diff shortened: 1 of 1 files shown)


Baezon:
Tested and everything seems to work correctly. +1 for a simpler fix just for the RC, for now, as well.

jg18:
I'm a bit concerned about the fact that AFAICT, FSO doesn't track whether weapons-locked-only slots are empty/filled, but if things are working for the modder who wants to use this behavior, perhaps it's good enough, especially given that there are plans for a rewrite of this code in the near-ish future.

wookieejedi (approved):
Works in my tests and agree that it's nice and simple.

PR #4401, opened 2022-06-13
add isActive checks for new-style script hooks

### LINKED ISSUE #4398: Not all script hooks are checked for isActive
PR #3538 established checks that run script hooks only when they are active, but not all script hooks are checked.  The new-style hooks need the isActive() check added.

Goober5000:
What it says on the tin.  Follow-up to #3538.  Fixes #4398.


EatThePath (commented):
Looks fairly solid and straightforward. Just for safety's sake though, have mods using some of the affected hooks been run on this branch to make sure they behave?

Goober5000:
Yup, I've run a couple missions from Scroll and other mods that make use of these checks.

EatThePath (approved):
Looks good to me, then

PR #4404, opened 2022-06-14
Fix the standalone ship being added to ship list

JohnAFernandez:
While it may seem counterintuitive, we have to keep the standalone bogus ship from being added to the ship object list so that we don't have to exclude it from all the places it could cause crashes or mysterious AI behaviors.  Both were being observed, and this should fix them.

I literally looked at every single place that the list is used, and most instances are potentially problematic, with other instances being benign, but none of them serving any practical use.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -1798,7 +1798,7 @@ void parse_dock_one_docked_object(p_object *pobjp, p_object *parent_pobjp)
 	ai_dock_with_object(objp, dockpoint, parent_objp, parent_dockpoint, AIDO_DOCK_NOW);
 }
 
-int parse_create_object_sub(p_object *objp);
+int parse_create_object_sub(p_object *objp, bool standalone_ship = false);
 
 // Goober5000
 void parse_create_docked_object_helper(p_object *pobjp, p_dock_function_info *infop)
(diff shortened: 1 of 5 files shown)


code/mission/missionparse.cpp
```diff
@@ -1841,7 +1841,7 @@ int parse_create_object(p_object *pobjp)
 		// if the leader will be destroyed before the mission starts, then *only* create the leader;
 		// don't create the rest of the group (this is what retail did)
 		if (pobjp->destroy_before_mission_time >= 0)
-			return parse_create_object_sub(pobjp);
+			return parse_create_object_sub(pobjp, standalone_ship);
```
Goober5000, line 1844: This one can be removed.  If we don't need to specify the flag in the dock helper because it'll never be docked, then we don't need to specify it here because it'll never be destroyed before mission.

Goober5000 (changes requested):
One comment, otherwise this looks good.

JohnAFernandez:
Yeah, we had a discussion about it but didn't come to a resolution. My thought was that these bugs persisted because the hack is not widely known.  While there is value in keeping a hack hidden from users, if a hack is required, then hiding the hack will result in additional bugs, such as the crash that initiated this bug hunt.  Also, I don't think the standalone ship being a hack should preclude me from passing default arguments down.

PR #4406, opened 2022-06-14
Properly set rotation data from LuaAI

BMagnu:
Previously, rotation data was just written to the AI_ci, which, for rotation, is reset _and_ ignored.
Now fixes this, and properly resembles thrust (as described), regarding tabled damping and turntime values, as well as properly interacting with other AI functions setting the rotation.

### DIFF
--- a/code/scripting/api/objs/ai_helper.cpp
+++ b/code/scripting/api/objs/ai_helper.cpp
@@ -6,6 +6,8 @@
 #include "vecmath.h"
 
 #include "ai/ai.h"
+#include "mission/missionparse.h"
+#include "ship/ship.h"
 
 namespace scripting {
 namespace api {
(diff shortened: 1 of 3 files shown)


code/scripting/api/objs/ai_helper.cpp
```diff
@@ -39,19 +41,65 @@ inline int aici_getset_helper(lua_State* L, float control_info::* value) {
...
+		if (turn_time > 0.0f)
+		{
+			vel_limit.xyz.x = PI2 / turn_time;
+			vel_limit.xyz.y = PI2 / turn_time;
+			vel_limit.xyz.z = PI2 / turn_time;
+		}
+	}
+	vec3d acc_limit = ai_get_acc_limit(&vel_limit, ship.objp);
+
+	float currentThrustVel = ((ship.objp->phys_info.desired_rotvel.a1d[axis] - ship.objp->phys_info.rotvel.a1d[axis]) / AI_frametime) / acc_limit.a1d[axis];
```
Baezon, line 71: Should this not be
`((ship.objp->phys_info.desired_rotvel.a1d[axis] - ship.objp->phys_info.rotvel.a1d[axis]) * AI_frametime) / acc_limit.a1d[axis]`?

Baezon (changes requested):

code/scripting/api/objs/ai_helper.cpp
BMagnu: In that line I don't calculate the new thrust, but rather the thrust as commanded by eventual preceeding AI calls to ``ai_turn_towards_vector``
So, if ``Δv = a * Δt``, then, where the acceleration ``a`` is proportional to our thrust, ``a = Δv/Δt``, no?

Baezon: Yeah, nevermind, that is correct. Although I think "Vel" here is probably the wrong term. I think perhaps "rate" like the doc uses is probably more accurate.

Baezon (approved):

PR #4407, opened 2022-06-15
A comprehensive fix for the `player.target_is_dying` variable.

### LINKED ISSUE #4393: 22.2 RC2: Asteroid brackets disappear or stop getting updated when your target dies
I noticed this issue in the retail mission "Rebels & Renegades". 

When Asteroids started being flinged at the Iceni, their off-screen indicators worked fine, but when I turned towards them they were unbracketed. Targeting *anything* made their brackets show up again.  

I eventually realised that it happened when any non-asteroid ship you have targeted dies. Asteroid brackets that are already on-screen don't disappear, but the distance indicator near it stops being updated. Asteroids that are off-screen won't get bracketed when you turn to them. 
It seems that just turning off targeting sometimes (but not always) also causes this issue, without needing to wait for a target ship to die.

JohnAFernandez:
A comprehensive fix for the `target_is_dying` variable.

Retail expected this variable to be set to -1 when the player target had died, but that assignment was only on the player struct reset.  So this adds an assignment when it's been determined that there is no current valid target.  I also added comments and made sure the logic was consistent on all the checks.  There is a little bit of redundancy on the checks, but honestly that's probably better in c++.  But there is more than one way to fix this, so I'm open if anyone feels strongly that it should be done another way.

Tested and works as expected.
Closes #4393

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -1515,6 +1515,7 @@ void hud_update_frame(float  /*frametime*/)
 	}
 
 	if (Player_ai->target_objnum == -1) {
+		Player->target_is_dying = -1; // according to comments elsewhere in the code, set to -1 when no target
 		hud_target_change_check();
 		return;
 	}
--- a/code/hud/hudbrackets.cpp
+++ b/code/hud/hudbrackets.cpp
@@ -395,7 +395,12 @@ void HudGaugeBrackets::renderObjectBrackets(object *targetp, color *clr, int w_c
 	int bound_rc;
 	SCP_list<CJumpNode>::iterator jnp;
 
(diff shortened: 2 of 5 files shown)


Goober5000 (changes requested):
Copying comments from Discord...

> first, it needs to calculate the distance differently, and second the condition can be linked to the other condition
> first, put this at the top of the function, and use it for the check on line 398 as well as line 478
> ```
> bool targetp_override = (Player_ai->target_objnum < 0 || targetp != &Objects[Player_ai->target_objnum]);
> ```
> second, use this check for line 478
> ```
> if (targetp_override) {
>     distance = hud_find_target_distance(targetp, Player_obj);
> } else {
>     distance = Player_ai->current_target_distance;
> }
> ```

PR #4410, opened 2022-06-15
Fix issues with loading of hud config presets

notimaginative:
This allows for retail hud config presets to once again show up in the list, as they were previously excluded from the file cache. Additionally it fixes newly created presets that were unable to be found without a restart when a mod is specified.


notimaginative:
Before this gets merged it should be noted that the default path location override just includes player locations. It's possible that the cache and multi cache should also be added to this list. However adding the cache might cause issues for any mods that expect the cache folder to be specific to them.

So this PR should probably be amended to either add the cache and multi cache directories to the override list, or a comment should be added to indicate that those paths are specifically excluded.

wookieejedi:
`However adding the cache might cause issues for any mods that expect the cache folder to be specific to them.`
Yeah, being able to save things like json and other files via scripting to the specific mod folder is very important for multiple mods, including BtA and FotG. So I highly support the second listed option `a comment should be added to indicate that those paths are specifically excluded.`

notimaginative:
Added a comment to note that cache directories should not be added to the override. Should be good to go now.

wookieejedi (approved):
With that comment addition it looks good to me.

PR #4411, opened 2022-06-15
New options for lead indicators

Kiloku:
Adds a new game_settings.tbl option: `$Lead indicator behavior`, which can be set to `default`, `multiple` or `average`. 
Also adds a new flag to weapons.tbl: `don't merge lead indicators`

When `$Lead indicator behavior` is set to `default` (or unset), lead indicators act just like in retail, meaning when you link them, only the indicator for the one with the farthest range will be drawn.
When set to `multiple`, a lead indicator is drawn for all primary weapons in range when you have them linked.  
When set to `average`, a single lead indicator is drawn in the average of the position where each of them would usually be drawn.  
Invalid selections will revert to default and give a warning.

When Weapon flag `don't merge lead indicators` is set for a homing secondary, its lead indicator will be drawn as if it was a dumbfire missile.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -3947,40 +3947,6 @@ void HudGaugeLeadIndicator::renderLeadCurrentTarget()
 	if ( swp->num_primary_banks == 0 )
 		return;
 
-	bank_to_fire = hud_get_best_primary_bank(&prange);
-
-	if ( bank_to_fire < 0 )
-		return;
-
-	wip = &Weapon_info[swp->primary_bank_weapons[bank_to_fire]];
-
-	if (pm->n_guns && bank_to_fire != -1 ) {
-		rel_pos = &pm->gun_banks[bank_to_fire].pnt[0];
-	} else {
-		rel_pos = NULL;
-	}
-
-	// source_pos will contain the world coordinate of where to base the lead indicator prediction
-	// from.  Normally, this will be the world pos of the gun turret of the currently selected primary
-	// weapon.
-	source_pos = Player_obj->pos;
-	if (rel_pos != NULL) {
-		vec3d	gun_point;
-		vm_vec_unrotate(&gun_point, rel_pos, &Player_obj->orient);
-		vm_vec_add2(&source_pos, &gun_point);
-	}
-
-	// Determine "accurate" distance to target.
-	// This is the distance from the player ship to:
-	//   (if targeting a subsystem) the distance to the subsystem centre
-	//     (playing it safe, will usually be in range at slightly further away due to subsys radius)
-	//   (otherwise) the closest point on the bounding box of the target
-	if ( Player_ai->targeted_subsys != NULL ) {
-		dist_to_target = vm_vec_dist(&target_pos, &Player_obj->pos);
-	} else {
-		dist_to_target = hud_find_target_distance(targetp, Player_obj);
-	}
-
 	srange = ship_get_secondary_weapon_range(Player_ship);
 
 	if ( (swp->current_secondary_bank >= 0) && (swp->secondary_bank_weapons[swp->current_secondary_bank] >= 0) )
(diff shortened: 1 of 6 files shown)


code/hud/hudtarget.cpp
```diff
@@ -3979,46 +3970,117 @@ void HudGaugeLeadIndicator::renderLeadCurrentTarget()
...
+				else if (Lead_indicator_behavior == leadIndicatorBehavior::AVERAGE) {
+					vm_vec_add2(&averaged_lead_pos, &lead_target_pos);
+					average_instances++;
+				}
+			}
+		}
+		if (Lead_indicator_behavior == leadIndicatorBehavior::AVERAGE) {
+			averaged_lead_pos = averaged_lead_pos/i2fl(average_instances);
+			renderIndicator(frame_offset, targetp, &averaged_lead_pos);
 		}
```
BMagnu, line 4033: What does this do for average_instances == 0?
Should this if not also explicitly check for that, and in that case, not render any indicator?

code/hud/hudtarget.cpp
```diff
@@ -3979,46 +3970,117 @@ void HudGaugeLeadIndicator::renderLeadCurrentTarget()
...
+			} else {
+				rel_pos = NULL;
+			}
+
+			source_pos = Player_obj->pos;
+			if (rel_pos != NULL) {
+				vec3d	gun_point;
+				vm_vec_unrotate(&gun_point, rel_pos, &Player_obj->orient);
+				vm_vec_add2(&source_pos, &gun_point);
+			}
```
BMagnu, line 4011: All of this is just to calculate ``source_pos``, but it's never actually used / read from. This whole block can just go

code/hud/hudtarget.cpp
```diff
@@ -3979,46 +3970,117 @@ void HudGaugeLeadIndicator::renderLeadCurrentTarget()
...
+			if (swp->primary_bank_weapons[bank] < 0)
+				continue;
+
+			wip = &Weapon_info[swp->primary_bank_weapons[bank]];
+
+			if (wip->wi_flags[Weapon::Info_Flags::Nolink] ||  (wip->wi_flags[Weapon::Info_Flags::Ballistic] && swp->primary_bank_ammo[bank] <= 0))
+				continue;
+
+			if (bank < 0)
+				continue;
```
BMagnu, line 3998: If you have the bank >= 0 assert further up, this check here makes little sense. Either remove it, or at least move it far enough up so that bank isn't already used to index into arrays

code/hud/hudtarget.cpp
```diff
@@ -3979,46 +3970,117 @@ void HudGaugeLeadIndicator::renderLeadCurrentTarget()
...
-	}
+		// source_pos will contain the world coordinate of where to base the lead indicator prediction
+		// from.  Normally, this will be the world pos of the gun turret of the currently selected primary
+		// weapon.
+		source_pos = Player_obj->pos;
+		if (rel_pos != NULL) {
+			vec3d	gun_point;
+			vm_vec_unrotate(&gun_point, rel_pos, &Player_obj->orient);
+			vm_vec_add2(&source_pos, &gun_point);
+		}
```
BMagnu, line 4050: Same here, source_pos is just not needed

code/hud/hudtarget.cpp
```diff
@@ -3979,46 +3970,117 @@ void HudGaugeLeadIndicator::renderLeadCurrentTarget()
...
+
+		if (frame_offset >= 0) {
+			hud_calculate_lead_pos(&lead_target_pos, &target_pos, targetp, wip, dist_to_target);
+			renderIndicator(frame_offset, targetp, &lead_target_pos);
+		}
 	}
 
+	if ( bank_to_fire < 0 )
+		return;
+
```
BMagnu, line 4065: Since you moved this check down, can it happen that, in the default mode, an indicator is drawn for bank_to_fire == -1 before this check exists the function here?

BMagnu (changes requested):

code/hud/hudtarget.h
```diff
@@ -69,6 +69,12 @@ extern const char *Strafe_submode_text[];
 
 extern void hud_init_targeting_colors();
 
+enum leadIndicatorBehavior {
```
BMagnu, line 72: This should be an enum class, not just an enum

code/hud/hudtarget.cpp
Kiloku: Apparently `source_pos` _and_ `rel_pos` were never used in this function even before these changes :sweat_smile: 
I'll remove both

code/hud/hudtarget.cpp
Kiloku: I think you're right, but this also makes me realize that the original check (as well as this one) also kills the dumbfire secondary lead indicator when `bank_to_fire == -1`. A player without primaries but with dumbfire secondaries would not see an indicator, it seems. Will fix.

BMagnu: Yes. This _could_ be changed, but would have to be enabled with a game settings table, as it's a rather significant difference to retail

Kiloku: Oh, that's fair

BMagnu (changes requested):
Looks pretty good now. One tiny detail, still.

code/hud/hudtarget.cpp
```diff
@@ -3981,27 +3968,66 @@ void HudGaugeLeadIndicator::renderLeadCurrentTarget()
...
+			hud_calculate_lead_pos(&lead_target_pos, &target_pos, targetp, wip, dist_to_target);
+			renderIndicator(frame_offset, targetp, &lead_target_pos);
+		}
 	}
 
-	if (frame_offset >= 0) {
-		hud_calculate_lead_pos(&lead_target_pos, &target_pos, targetp, wip, dist_to_target);
-		renderIndicator(frame_offset, targetp, &lead_target_pos);
-	}
+	if ( bank_to_fire < 0 )
```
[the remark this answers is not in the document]
BMagnu: or at least an ``else if``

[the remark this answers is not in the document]
Kiloku: Done

BMagnu (approved):

code/hud/hudtarget.cpp
Kiloku: One note that I realized afterwards by reading old code: 
Retail apparently never had dumbfire lead indicators. It was the first post-volition change to this file :laughing: 

But since it's so old, I think it still stands that this is the way to go

PR #4413, opened 2022-06-16
end the campaign properly when a mission is skipped

### LINKED ISSUE #4412: Skipping training causes crash in FSPort
If the last mission in a campaign is a training mission, attempting to skip that mission using "skip training" will cause a crash.  This occurs on FSPort because the last mission is configured as a training mission.

Reported by Daft Mugi on Discord.

Goober5000:
The "skip training" button and the "exit loop" button should both check whether the campaign is actually over before attempting to load the next mission.  The fix is simple - just use the same logic as in `mission_campaign_skip_to_next`.

Fixes #4412.

### DIFF
--- a/code/mission/missioncampaign.cpp
+++ b/code/mission/missioncampaign.cpp
@@ -1822,7 +1822,12 @@ void mission_campaign_exit_loop()
 
 	// set things up for next mission
 	mission_campaign_next_mission();
-	gameseq_post_event(GS_EVENT_START_GAME);
+
+	if ( Campaign.next_mission == -1 || (The_mission.flags[Mission::Mission_Flags::End_to_mainhall]) ) {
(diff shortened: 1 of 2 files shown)


code/mission/missioncampaign.cpp
```diff
@@ -1822,7 +1822,12 @@ void mission_campaign_exit_loop()
 
 	// set things up for next mission
 	mission_campaign_next_mission();
```
jg18, line 1824: Is this call still correctly placed?

jg18 (approved):
Looks good. Just one question.

code/mission/missioncampaign.cpp
Goober5000: It should be.  In any case, there's no harm in calling it even if the campaign has ended.

Issue #4408, opened 2022-06-15
Should ship classes with the same model but different subsystems work or not?


Baezon:
The table syntax perfectly allows modders to define two ship classes with the same model, but different subsystems (particularly different numbers of subsystems), and logically there's no obvious reason why this shouldn't work, where the model file's subsystems are 'available' to the ship class for usage in any combination or number consistent with the model file. Yet, functions like `ship_copy_subsystem_fixup()` seem to assert that classes which share a model must also share subsystems. 

I don't really understand that code very well, so I don't know if that system is just bugged and doesn't need to do that, or this is indeed a fundamental requirement that is completely unenforced or mentioned.

MageKing17:
AIUI, `ship_copy_subsystem_fixup()` is copying *model* subsystems, not *ship* subsystems; it should be pretty safe to assume that two instances of the same model should have the same model subsystems.

Baezon:
I can confirm that it is the `model_subsystem`s in `ship_info->subsystems` that can differ even if they share the same model that is causing this issue.

Baezon:
Ok, well that's not great design, the warning was being suppressed because the `mprintf(("No subsystems found for model \"%s\".\n", model_get(model_num)->filename));` was happening earlier, which flags `ss_warning_shown`. I'm looking into that now, but I think that covers it.

Baezon:
https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/ship/ship.cpp#L4026
>.>
Cool, because I didn't provide a $Closeup_pos: it manually calculated it from the model, by loading the model right then and there with NULL as the subsystem list (because it hasn't even been parsed yet), which triggers the 'No subsystems found...' mprintf, and suppresses the warning. Obviously, this behavior needs some clean up but the core issue is clear, table and model subsystems need to match no matter what.

PR #4414, opened 2022-06-16
split subsystem warning display checks

Goober5000:
Since "No subsystems found" and "Not all subsystems have a record in ships.tbl" are different warnings, they should have different flags.  Follow-up to #4408.


Baezon (approved):
I'm also going to re-jigger $Closeup_pos to not need a `model_load()` at all (and maybe audit any other frivolous `model_load()`s), but that's probably better saved until after the release.

PR #4416, opened 2022-06-16
Two minor lab fixes

### LINKED ISSUE #4415: Two F3 Lab CTDs
1) In either debug or release open the F3 lab and press M. Game will CTD, with this error in debug 
```
Assert: "handle >= 0"
File: bmpman.cpp
Line: 112
Invalid handle -1 passed to bm_get_slot!
```

2) In debug open the F3 lab and press T. Game will CTD with no message (no crash in release).

The-E:
Make sure that changing team colors only happens when there are team colors to change and only dump the envmap to disk if there actually is an envmap to dump.

Fixes #4415

### DIFF
--- a/code/lab/manager/lab_manager.cpp
+++ b/code/lab/manager/lab_manager.cpp
@@ -224,7 +224,9 @@ void LabManager::onFrame(float frametime) {
 				break;
 
 			case KEY_M:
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):
LGTM

PR #4417, opened 2022-06-16
Update `ai_turn_towards_vector` to use a target vel for banking

Baezon:
`bank_override` in `ai_turn_towards_vector` is a target angular position for banking in radians and it will do as much as it can to reach that position in the frame it is called. But all of the calls using the parameter with non-zero use... the ship's speed. Yes, 40, 80 whatever meters per second -> do what you can to bank by 40, 80 whatever radians this frame. In effect, this was only ever 0 banking, or bank as hard as possible. So this changes it so `bank_override` is instead a target velocity (expressed in % of max bank speed) and `vm_angular_move_forward_vec` is updated to instead aim for such a target velocity.

Not a bug fix per se, but this allows @BMagnu to use banking in turnTowardsPoint in LuaAI in a much more intuitive and useful way.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -1195,7 +1195,7 @@ void ai_turn_towards_vector(vec3d* dest, object* objp, vec3d* slide_vec, vec3d* 
 	vec3d	vel_in, vel_out, desired_fvec, src;
 	float		delta_time;
 	physics_info	*pip;
-	float		delta_bank;
+	float		bank;
 
 	Assertion(objp->type == OBJ_SHIP || objp->type == OBJ_WEAPON, "ai_turn_towards_vector called on a non-ship non-weapon object!");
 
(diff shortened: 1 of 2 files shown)


code/ai/aicode.cpp
```diff
@@ -1289,17 +1289,18 @@ void ai_turn_towards_vector(vec3d* dest, object* objp, vec3d* slide_vec, vec3d*
...
-		delta_bank = 0.0f;
+		bank = 0.0f;
 	else if ((bank_override) && (iff_x_attacks_y(Ships[objp->instance].team, Player_ship->team))) {	//	Theoretically, this will only happen for Shivans.
-		delta_bank = bank_override;
+		bank = bank_override;
 	} else {
-		delta_bank = vm_vec_dot(&curr_orient.vec.rvec, &objp->last_orient.vec.rvec);
-		delta_bank = 200.0f * (1.0f - delta_bank) * pip->delta_bank_const;
+		bank = vm_vec_dot(&curr_orient.vec.rvec, &objp->last_orient.vec.rvec);
+		bank = 200.0f * (1.0f - bank) * pip->delta_bank_const;
```
BMagnu, line 1299: In theory, this should be changed to accommodate the change.
In practice, the values are so close to zero both before and after the change that the impact on the actual maneuver is negligible

BMagnu (approved):
One thing just to note for the record, but should be fine as-is

PR #4420, opened 2022-06-17
only display one popup when the player gets to the main hall

### LINKED ISSUE #4418: Player tips displayed after "campaign not found" popup
If the "campaign cannot be found" popup is displayed, the player tips popup can be displayed immediately afterward, after the first popup's resolution has been chosen but before it actually happens.  Not a bug, strictly speaking, but this can be confusing.

Goober5000:
There are three popups of decreasing priority that can appear in the main hall screen: campaign not found, pilot upgraded, and player tip.  Only one of these should actually be displayed -- the first popup will take the player to a different screen, rendering subsequent popups useless; and the second popup is more important than the third popup.  This PR prioritizes the popups accordingly.

Yes, this uses `goto`.  It's a succinct and clear way of implementing the necessary logic.  It works better than both a one-iteration loop and a cascading if() block.

Fixes #4418.

### DIFF
--- a/code/menuui/playermenu.cpp
+++ b/code/menuui/playermenu.cpp
@@ -1459,7 +1459,10 @@ void player_tips_popup()
 	} while(ret > 0);
 }
 
-void player_tips_controls() {
+/**
+ * Displays a popup notification if the pilot was converted from pre-22.0 which might disrupt control settings.  Returns true if the popup was shown.
+ */
+bool player_tips_controls() {
 	if (Player->save_flags & PLAYER_FLAGS_PLR_VER_PRE_CONTROLS5) {
 		// Special case. Since the Controls5 PR is significantly different from retail, users must be informed
 		// of changes regarding their bindings
@@ -1476,7 +1479,11 @@ void player_tips_controls() {
 			Pilot.save_player(Player);
 			Pilot.save_savefile();
 		}
+
+		return true;
 	}
+
+	return false;
 }
 
 SCP_vector<SCP_string> player_select_enumerate_pilots() {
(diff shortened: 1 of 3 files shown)


code/menuui/playermenu.cpp
```diff
@@ -1476,7 +1476,12 @@ void player_tips_controls() {
 			Pilot.save_player(Player);
 			Pilot.save_savefile();
 		}
+
+		// popup was displayed
```
jg18, line 1480: Why not just comment above the function, either here or in the header file,
```
// returns true if the popup was displayed, false otherwise
```

jg18 (commented):
TBH, I am *suuuuper* skeptical that `goto` is warranted here.

What's wrong with adding an extra bool flag around line 1061 like
```
bool popup_shown = false;
```

then set `popup_shown` to `true` on line 1084, and then change lines 1088-1094 to something like this (comments omitted here for brevity):

```
	if (!popup_shown) {
		popup_shown = player_tips_controls();
	}

	if (!popup_shown) {
		player_tips_popup();
	}
```

That said, if you insist on this approach, I'll approve it, albeit with a clothespin clamped to my nose.

Also, to fully understand the danger of `goto`, see https://xkcd.com/292/

Goober5000:
Yeah, that'll work.  It's a sequential if() block, which works better here than a cascading if() block.

code/menuui/playermenu.cpp
Goober5000: Good point

jg18 (approved):
Thanks, looks good now.

PR #4421, opened 2022-06-17
add missing SEXP goals

Goober5000:
The rearm-ship and fly-to-ship goals were never set up to work with the `add-goal` SEXP.  This adds the necessary code to make them both work.

The funny thing is that any ship pursuing a fly-to-ship goal will crash into its target ship.  The goal uses the exact position, not a position some distance away like most other goals.


wookieejedi (approved):
Was going to recommend to adjust the formatting so that columns lined up, but then remembered that the huge list in sexp.cpp is very inconsistent in formatting (and cleaning that would be it's own whitespace-based PR). Overall looks good.

PR #4422, opened 2022-06-17
Followup improvement for the LuaAI AI Control Helper

BMagnu:
Followup to #4406, properly fixing the rotation controls in a way that is both predictable and intuitive.

### DIFF
--- a/code/scripting/api/objs/ai_helper.cpp
+++ b/code/scripting/api/objs/ai_helper.cpp
@@ -41,65 +41,19 @@ inline int aici_getset_helper(lua_State* L, float control_info::* value) {
 	return ade_set_args(L, "f", AI_ci.*value);
 }
 
-inline int pi_rotation_getset_helper(lua_State* L, int axis) {
-	//Use with care! This works only if, as expected, the l_AI_Helper object is the one supplied to the luaai scripts this very frame
-	object_h ship;
-	float f = 0.0f;
-	if (!ade_get_args(L, "o|f", l_AI_Helper.Get(&ship), &f))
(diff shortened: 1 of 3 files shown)


code/scripting/api/objs/ai_helper.cpp
```diff
@@ -128,11 +82,14 @@ ADE_FUNC(turnTowardsPoint,
 	vec3d* target;
 	bool diffTurn = true;
 	vec3d* modifier = nullptr;
-	if (!ade_get_args(L, "oo|bo", l_AI_Helper.Get(&ship), l_Vector.GetPtr(&target), &diffTurn, l_Vector.GetPtr(&modifier))) {
+	float bank = 0.0f;
+
+	int argnum = ade_get_args(L, "oo|bof", l_AI_Helper.Get(&ship), l_Vector.GetPtr(&target), &diffTurn, l_Vector.GetPtr(&modifier), &bank);
+	if (argnum == 0) {
```
Baezon, line 88: Why `== 0`? There's two required arguments, right? Why not simply use the functional call itself?

Baezon (changes requested):

code/scripting/api/objs/ai_helper.cpp
BMagnu: Because 0 is a special case for "error".
All other functions just do !ade_get_args, but I don't since I need to grab the actual arg number to make the last parameter truely optional

Baezon: Ah, right of course.

Baezon (approved):

PR #4423, opened 2022-06-18
fix parsing issues with hud color presets

notimaginative:
The first issue is that if a gauge isn't found it will fail to parse any remaining gauges in the file. The second issue is that when saving the preset the game uses translated strings for the gauge name but when loading the preset it uses untranslated strings.

This should address both issues while keeping the original purpose of the changes as well as not breaking existing presets or requiring any mod changes.

Ref: http://scp.indiegames.us/mantis/view.php?id=3099

### DIFF
--- a/code/hud/hudconfig.cpp
+++ b/code/hud/hudconfig.cpp
@@ -1613,6 +1613,7 @@ void hud_config_color_load(const char *name)
 	int idx;
 	char str[1024];
 	char *fname;
+	ubyte r, g, b, a;
 
 	fname = cf_add_ext(name, ".hcf");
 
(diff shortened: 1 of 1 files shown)


EatThePath (approved):
Looks good to me. One thought:
I recently made some changes intended to move towards more standardized and succinct parsing in tables, which include parsing 3-int colors into a new type for light colors. This can be seen implemented in code/parse/parsehi.cpp and used in weapons/weapons.cpp:1168 or thereabouts.
This fix presents the opportunity to advance that pattern, but the context is different enough that it would need extra implementation and not just using an already-available function, so this is only a suggestion and not a hold-up at all.

notimaginative:
I like the idea, but it looks like adding the proper implementation would require more lines of code than the actual fix does. That takes it a bit beyond the scope of this PR I think.

PR #4424, opened 2022-06-19
fix control chars test that also excluded space key

notimaginative:


JohnAFernandez (approved):
Wow.  All this over that.

Goober5000 (approved):
LOL wow.

Goober5000:
Are there any other tests in this file that might need to be tweaked?  For example, this one:
```
if ((ascii > 0) && (ascii < 255)) {
```
Should those be >= and <= ?

notimaginative:
> Are there any other tests in this file that might need to be tweaked? For example, this one:
> 
> ```
> if ((ascii > 0) && (ascii < 255)) {
> ```
> 
> Should those be >= and <= ?

I think that test is correct. I believe that's actually from retail, just moved to a different spot.

z64555:
:eyes:

PR #4426, opened 2022-06-19
Flush the mouse state after toggling between Use_mouse_to_fly states

### LINKED ISSUE #4173: controls are not flushed when weapons are unlocked
Prior to the controls merge, if the player's primaries and secondaries were locked, any "fire weapon" control presses would be ignored or flushed.  After the controls merge, the control presses are no longer ignored - unlocking the primary and/or secondary weapons will cause a delayed weapons fire.

I tested this with the "primaries-locked" and "secondaries-locked" flags, although it's possible it happens with "afterburners-locked" as well.

z64555:
Fixes #4173 

...Was it really that simple? 😩

### DIFF
--- a/code/scripting/api/libs/controls.cpp
+++ b/code/scripting/api/libs/controls.cpp
@@ -334,6 +334,8 @@ ADE_VIRTVAR(MouseControlStatus, l_Mouse, "boolean", "Gets and sets the retail mo
 		{
 			Use_mouse_to_fly = false;
 		}
+
+		mouse_flush();
 	}
 
 	if (Use_mouse_to_fly)


z64555:
@Goober5000 @Kiloku Please test this PR to verify the fix works on your end.

z64555:
Flushing the mouse state at the toggle clears out the button_pressed members of the mouse.

This is probably a stop-gap measure, but it does grealy narrow down where the phenomenon happens.

Kiloku:
Tested and it works here

Goober5000 (approved):
Tested and works 👍

z64555:
Lua checks mouse_flags to see if a button is down or not, but does not clear the respective button counter when doing so.  So, when the original script rolls around to unlocking the mouse, the controls code checks if a mouse button is currently down *and* if the any of the down counters are non-zero.

PR #4427, opened 2022-06-19
Ai goal refinement

Goober5000:
1) Move ai_goal definitions into aigoal.h
2) Add a proper `ai_goal_reset` function to clear goal fields to known initialized values
3) Add an optional distance to `ai-stay-near-ship`


BMagnu (approved):
Looks generally fine, but should be well-tested with a few missions before it is merged.

Goober5000:
Tested and tweaked.

PR #4428, opened 2022-06-20
more fixes for the skip-training-mission crash

### LINKED ISSUE #4412: Skipping training causes crash in FSPort

Goober5000:
Certain campaign-related functions would run even when the game was not in campaign mode (e.g. when playing in the tech room).  Normally these would be benign, but with a new pilot or with a skipped mission, the current campaign mission would be -1 and a crash would occur.

Should fully fix #4412.

### DIFF
--- a/code/mission/missioncampaign.cpp
+++ b/code/mission/missioncampaign.cpp
@@ -951,6 +951,9 @@ void mission_campaign_store_goals_and_events()
 	int cur, i;
 	cmission *mission_obj;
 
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
Straightforward fix.

PR #4430, opened 2022-06-20
Keep Player Selection after changing modes

JohnAFernandez:
Basically on the player selection screen, it's annoying when the selected player switches between multi and single modes and the selected player changes.  This keeps the player's choice after switching.


code/menuui/playermenu.cpp
```diff
@@ -946,7 +946,7 @@ int player_select_get_last_pilot()
 	return 0;
 }
 
-void player_select_init_player_stuff(int mode)
+void player_select_init_player_stuff(int mode, SCP_string const current_callsign)
```
notimaginative, line 949: Why modify the function call? It seems like a value which could easily be determined in the function itself for a cleaner implementation.

notimaginative (changes requested):

code/menuui/playermenu.cpp
JohnAFernandez: Well, this operation should only be done during half of the calls to this function.  I can make a global bool instead?

JohnAFernandez:
Closing because I realized that this is the wrong approach.  We shouldn't be rebuilding the list when switching modes because it's the same list.

PR #4433, opened 2022-06-22
stop non-printable keys adding text to multi msg buffer

notimaginative:
Arrow, function, Ctrl, Alt ... and any other non-ascii keys would previously add a blank space to the multi msg buffer. This prevents that from happening, making sure that it's an ascii character before appending it to the buffer.

(Retail bug??)


wookieejedi (approved):
Simple enough fix!

PR #4434, opened 2022-06-23
sync Orientation virtvars

Goober5000:
The two `VIRTVAR`s for Orientation in subsystems and submodel instances do the same thing, but did not have the same code - in particular, the subsystem version did not update the subsystem angle and turret angle.  This syncs them up.


wookieejedi (approved):
Straightforward and looks good

PR #4436, opened 2022-06-24
Fixes Team Colors not working with `show-ship`

### LINKED ISSUE #4435: Team color does not show up on `show-ship` from first person view
If the player is in first-person view with a ship that uses `show-ship` flag, then team colors do not appear. If the player switches to third-person then the team colors are properly visible.

wookieejedi:
Updates the `show-ship` function to account for team colors. Also, cleans up the function a bit. Tested and works as expected. Fixes #4435

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -7419,10 +7419,20 @@ void ship_render_show_ship_cockpit(object *objp)
 	model_render_params render_info;
 
 	render_info.set_object_number(OBJ_INDEX(objp));
-	render_info.set_replacement_textures(Ships[objp->instance].ship_replacement_textures);
+
+	// update any replacement and/or team color textures (wookieejedi), then render
+	ship* shipp = &Ships[objp->instance];
+	ship_info* sip = &Ship_info[Ships[objp->instance].ship_info_index];
+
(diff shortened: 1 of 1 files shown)


code/ship/ship.cpp
```diff
@@ -7419,10 +7419,20 @@ void ship_render_show_ship_cockpit(object *objp)
 	model_render_params render_info;
 
 	render_info.set_object_number(OBJ_INDEX(objp));
-	render_info.set_replacement_textures(Ships[objp->instance].ship_replacement_textures);
+
+	// update any replacement and/or team color textures (wookieejedi), then render
+	ship* shipp = &Ships[objp->instance];
+	ship_info* sip = &Ship_info[Ships[objp->instance].ship_info_index];
+
+	render_info.set_replacement_textures(shipp->ship_replacement_textures);
```
JohnAFernandez, line 7427: There's no need to check that the textures are valid before calling? Does `set_replacement_textures` already check?

wookieejedi: Good question. That function does not have that `is_team_color_set` check within it. As such, all other calls of `render_info.set_replacement_textures` had that same `if` check.

JohnAFernandez (approved):

PR #4437, opened 2022-06-25
fix crash when removing form-on-wing

Goober5000:
This goal was missed in `ai_remove_goal_sexp_sub`, so add handling for it.  Also make the error message clearer in case another goal is missed in the future.

~~In draft status until it can be tested by Durandal.~~  Tested and verified.


wookieejedi (approved):
Good to hear the test is verified. LGTM.

PR #4440, opened 2022-06-25
better handling of invalid message sources

Goober5000:
Another bug discovered by Durandal.  The message sender was a variable containing an invalid value, and the game crashed.  This handles it more gracefully.

Tested and verified.


wookieejedi (approved):
Durandal is on a roll. Nice and simple update.

PR #4445, opened 2022-06-26
Ignore Wshadow warnings in stale librocket headers

the-maddin:
GCC 11 raises a warning due to a parameter "word" shadowing a typedef "word". I made a PR correcting
that to the upstream repo some time ago, which is still ignored. This is the last remaining issue with GCC 11.


the-maddin:
There was a random error building unittests in the last PR run. Could somebody restart them?

the-maddin:
... And the same filesystem error again

asarium (approved):

PR #4446, opened 2022-06-26
Fix Seg Fault on Standalones with Safer Check for Secondaries

wookieejedi:
Tests with standalone multiplayer builds revealed that the server would seg fault if a player tried to switch secondaries if the ship did not have secondaries. Tests with standalones showed this fixed the bug. Thanks to @chief1983 for the bug IDing and the fix code.


JohnAFernandez (approved):
This should be the correct fix.

PR #4447, opened 2022-06-26
even more fixes for the skip-training-mission crash

Goober5000:
As noted in #4412, it's possible for `Campaign.current_mission` to be -1 even in campaign mode.  This explictly checks for that.

### DIFF
--- a/code/mission/missioncampaign.cpp
+++ b/code/mission/missioncampaign.cpp
@@ -951,7 +951,7 @@ void mission_campaign_store_goals_and_events()
 	int cur, i;
 	cmission *mission_obj;
 
-	if (!(Game_mode & GM_CAMPAIGN_MODE))
+	if (!(Game_mode & GM_CAMPAIGN_MODE) || (Campaign.current_mission < 0))
 		return;
 
 	cur = Campaign.current_mission;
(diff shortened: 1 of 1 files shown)


code/mission/missioncampaign.cpp
```diff
@@ -951,7 +951,7 @@ void mission_campaign_store_goals_and_events()
 	int cur, i;
 	cmission *mission_obj;
 
-	if (!(Game_mode & GM_CAMPAIGN_MODE))
+	if (!(Game_mode & GM_CAMPAIGN_MODE) || (Campaign.current_mission < 0))
```
JohnAFernandez, line 954: Does this preclude someone using Eternal variables on these types of missions?

Goober5000: Pedantically speaking this has nothing to do with variables, since that would be handled in `mission_campaign_store_variables`, not this function.  Even there, the check doesn't affect any variables being *loaded*.

As for variables being *saved*, this PR changes no functionality; if the game crashed before it should now fail gracefully, and if it worked before it should still work now.

JohnAFernandez: So you're saying that we should ensure that eternal (and player-persistent?) variables can saved after these types of missions in a future PR?

Goober5000: There is no "after these types of missions" because this situation arises when a training mission or a loop mission is skipped *in the briefing*.  This change just makes that situation not crash.

JohnAFernandez: Ah.

JohnAFernandez (approved):

Issue #4432, opened 2022-06-22
Async waiting not working correctly on their first invocation


Baezon:
I have seen many symptoms, waiting an incorrect amount of time, ~triggering multiple times~ (false positive on that one!), triggering not at all, and it's difficult to figure out exactly what's happening, but I have a straightforward test case at least.

```
#Conditional Hooks
$Application: FS2_Open

$On Mission Start: [
    mn.sendPlainMessage("message1!")
    async.run(function()
        async.await(mn.waitAsync(1.0))
        mn.sendPlainMessage("message2!")
    end)
]
#End
```
[test-sct.txt](https://github.com/scp-fs2open/fs2open.github.com/files/8954467/test-sct.txt)

This should send message1 immediately, and message2 after one second, but currently, will only send message1 and message2 is skipped (specifically everything in the function is run until the `waitAsync` afterward, nothing runs). If you restart, the issue seems to fix itself. If you quit and re-enter the issue returns. I tried switching to the newer timestamp system, didn't help.

wookieejedi:
Seems that the bug occurs with `$On Mission Start:` but not with 
```
$State: GS_STATE_GAME_PLAY

$On Gameplay Start: [
]
```

Goober5000:
If that's true, then it must have something to do with when the timing system is initialized.

Goober5000:
It actually has nothing to do with timestamps.  The following check is evaluating to true:
```
if (contextState == executor::IExecutionContext::State::Invalid) {
```
So the promise is aborted and the `timestamp_elapsed` check is never satisfied.

Goober5000:
And that, in turn, is caused by this:
```
IExecutionContext::State GameStateExecutionContext::determineContextState() const
{
	const auto depth = gameseq_get_state_idx(m_contextState);

	if (depth < 0) {
		// Our game state is not present anymore
		return IExecutionContext::State::Invalid;
	}
```
If I had to guess, I'd say that the context is being reset when the game crosses state boundaries (starting in `GS_STATE_START_GAME` and ending in `GS_STATE_GAME_PLAY`).  But I'll defer to @asarium to explain what is going on here.

asarium:
Yes, that is likely to be the issue. The game state execution context stuff is meant to deal with things like going to the options menu and handling cases where the mission is switched with a running coroutine.

[...]

PR #4448, opened 2022-06-27
add some logging to waitAsync

Goober5000:
Log when the context is created, completes, and is aborted.  Adds insight for the situation described in #4432.

### DIFF
--- a/code/osapi/outwnd.cpp
+++ b/code/osapi/outwnd.cpp
@@ -29,7 +29,8 @@ static const char *FILTERS_ENABLED_BY_DEFAULT[] =
 {
 	"error",
 	"warning",
-	"general"
+	"general",
+	"scripting"
 };
 
 struct outwnd_filter_struct {
--- a/code/scripting/api/LuaPromise.cpp
+++ b/code/scripting/api/LuaPromise.cpp
@@ -92,7 +92,7 @@ LuaPromise::LuaPromise(const std::shared_ptr<resolve_context>& resolveContext) :
 {
 	m_state->state = State::Pending;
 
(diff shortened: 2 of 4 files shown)


Baezon (changes requested):
I wonder if an unconditional print for the successful creation and resolution of the waitAsync might be too much. I use a *lot* of waitAsync. Perhaps if it was an `nprintf` for the creation and successful resolution and `mprintf` for an invalid context?

Goober5000:
I don't think so; I think in this situation all three should use mprintf.  If the user sees an aborted context via mprintf, he is going to want to see when it was created, and using nprintf for the creation would impose a barrier.

If the concern is log pollution, I think the focus should be on the lower-hanging fruit such as "entering warpout" and all the shader log messages.

Baezon:
I mean, yes, that is true, but the fact there are things which pollute the log *more* than this might, isn't related. The fact that the situation is bad isn't a reason to make it worse. It being an `nprintf` is a minor imposition in the case where it is the cause of a bug; it being an `mprintf` is a major imposition in all other cases, at least for me. Blighted has waitAsyncs happening as often as every few seconds in common scenarios over an hour or more play session, these prints will blow shader compilations out of the water. What about a default-on category? I would really just like some way to turn it off for myself.

Goober5000:
Yeah, good point.

I could add a "scripting" category to nprintf, but I would like for it to default to "on", and I don't see a good way to do that.

Goober5000:
Ok, take a look at the new commit.

Baezon (approved):

PR #4449, opened 2022-06-27
Prevent a crash by exiting here

JohnAFernandez:
And make an assertion more explicit, just in case.

Fixes a crash for observers with certain ship table setups.

### DIFF
--- a/code/hud/hudreticle.cpp
+++ b/code/hud/hudreticle.cpp
@@ -313,6 +313,8 @@ void HudGaugeReticle::getFirepointStatus() {
 	// allow the firepoint status to be empty when a multiplayer observer
 	// this is not a bug, the observer will simply *not* have any firepoints.
 	if (Objects[Player->objnum].type == OBJ_OBSERVER) {
+		// only clients should be getting here!
+		Assertion((Game_mode & GM_MULTIPLAYER), "Somehow FSO thinks its player object is an observer even though it's not in Multiplayer. Please report!");
 		return; 
 	}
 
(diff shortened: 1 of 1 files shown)


code/hud/hudreticle.cpp
```diff
@@ -309,12 +309,19 @@ ship_info *sip = &Ship_info[Player_ship->ship_info_index];
 }
 
 void HudGaugeReticle::getFirepointStatus() {
+
+	// allow the firepoint status to be empty when a multiplayer observer
+	// this is not a bug, the observer will simply *not* have any firepoints.
+	if (Objects[Player->objnum].type == OBJ_OBSERVER) {
+		return;
```
jg18, line 316: Should we also assert here that `Game_mode & GM_MULTIPLAYER`?

code/hud/hudreticle.cpp
```diff
@@ -309,12 +309,19 @@ ship_info *sip = &Ship_info[Player_ship->ship_info_index];
...
+		return; 
+	}
+
 	//First, get the player ship
 	ship_info* sip;
 	ship* shipp;
 	polymodel* pm;
 
-	Assert(Objects[Player->objnum].type == OBJ_SHIP);
+	Assertion(Objects[Player->objnum].type == OBJ_SHIP, "HudGaugeReticle::getFirepointStatus was passed an invalid object type of %d.", Objects[Player->objnum].type);
```
jg18, line 324: How about adding a "Please report"?

jg18 (commented):
Adding a couple thoughts.

code/hud/hudreticle.cpp
JohnAFernandez: That is a great idea! :D

code/hud/hudreticle.cpp
JohnAFernandez: Yes, of course!

EatThePath (approved):
Looks all good to me

JohnAFernandez:
Thanks @EatThePath

Issue #4309, opened 2022-05-25
Campaign progress bugged when quitting to system on debriefing screen via "Exit Game" in options menu


Typhlomence:
If you complete a campaign mission and are on the debriefing screen, enter the options menu and use the exit game button to quit to the system, when you restart the game, you will move on to the next mission when going to the ready room. However, if you go to the mission simulator, the mission that was just completed is not shown as available to play.

This is erroneous behaviour in two ways:
1. The campaign progress and available missions in the simulator are out of sync.
2. Exiting to the system via other methods on the debriefing screen (such as Alt-F4) do not let you proceed to the next mission in the campaign.

A friend of mine discovered this bug in a version of FSO that is several years old now (specifically, 3.8.0-RC2); however, this still appears to be an issue in **22.0.0**.

Attached are the player files of a pilot that has reached a successful debriefing of "Surrender, Belisarius!" (skipping the training missions) in retail, and then quit to the system via the options menu. After restarting the game, the next mission upon clicking the ready room will be "The Place of Chariots". The pilot stats for the completed "Surrender, Belisarius!" are also present if you look in the barracks. However, "Surrender, Belisarius!" is not available when going to the tech room and selecting the mission simulator; only the first three training missions are present.

[Options Exit Bug Demo Rel.zip](https://github.com/scp-fs2open/fs2open.github.com/files/8770874/Options.Exit.Bug.Demo.Rel.zip)

Typhlomence:
Whoops! I was using RC6 when 22.0.0 was released; thank my Knossos settings for that. It still occurs in the release version though; I've edited the OP to reflect this.

jg18:
This fix is merged and will be included in the next 22.2 Release Candidate (RC). Thanks for reporting this.

PR #4450, opened 2022-06-27
Auto-accept completed campaign mission before exiting game from debriefing screen

jg18:
If the player exits the game from the options menu while at the debriefing screen after completing a mission during a campaign, then auto-accept the mission outcome.

Has not yet been scrutinized/tested.

Should fix Issue #4309.

### DIFF
--- a/code/menuui/optionsmenu.cpp
+++ b/code/menuui/optionsmenu.cpp
@@ -709,7 +709,7 @@ void options_button_pressed(int n)
 			gamesnd_play_iface(InterfaceSounds::USER_SELECT);
 			choice = popup( PF_NO_NETWORKING | PF_BODY_BIG, 2, POPUP_NO, POPUP_YES, XSTR( "Exit Game?", 374));
 			if ( choice == 1 ) {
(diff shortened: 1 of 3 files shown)


Goober5000 (commented):
Looks good to me.  Will hold off on approving until it is tested and out of draft status.

jg18:
With the latest change, a manual test now passes.

code/menuui/optionsmenu.cpp
```diff
@@ -707,8 +708,13 @@ void options_button_pressed(int n)
 		case ABORT_GAME_BUTTON:
 			gamesnd_play_iface(InterfaceSounds::USER_SELECT);
 			choice = popup( PF_NO_NETWORKING | PF_BODY_BIG, 2, POPUP_NO, POPUP_YES, XSTR( "Exit Game?", 374));
-			if ( choice == 1 )
+			if ( choice == 1 ) {
+				if (gameseq_get_state(1) == GS_STATE_DEBRIEF && (Game_mode & GM_CAMPAIGN_MODE)) {
```
jg18, line 712: The only way this call could be problematic is if there are no previous game states, which isn't possible in this case.

Goober5000 (approved):
Good catch. 👍  Approved.

PR #4452, opened 2022-06-28
set up debug_filter.cfg to have default filters that can be turned off

Goober5000:
Small refactoring of the debug filter, needed as a prerequisite for #4448.

### DIFF
--- a/code/osapi/outwnd.cpp
+++ b/code/osapi/outwnd.cpp
@@ -25,11 +25,19 @@
 #include "cfile/cfilesystem.h"
 #include "parse/parselo.h"
 
+static const char *FILTERS_ENABLED_BY_DEFAULT[] =
+{
+	"error",
+	"warning",
+	"general"
+};
+
 struct outwnd_filter_struct {
 	char name[NAME_LENGTH];
 	bool enabled;
 };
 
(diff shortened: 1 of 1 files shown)


code/osapi/outwnd.cpp
```diff
@@ -69,30 +86,18 @@ void load_filter_info()
...
-		new_filter.enabled = true;
-		filter_vector().push_back( new_filter );
-
 		return;
 	}
 
 	Outwnd_no_filter_file = 0;
 
 	while ( fgets(inbuf, NAME_LENGTH+3, fp) ) {
+		bool enabled;
```
notimaginative, line 95: Although I think the `continue` statement below would take care of it, valgrind might throw a warning about this variable not being initialized. It might be nice to just initialize this to `false` and not have to worry about it again.

But valgrind *should* be smart enough to not flag this, so I'll leave it up to you whether or not to make the change.

notimaginative (approved):
One minor thing that you can address, or not, but otherwise it LGTM.

code/osapi/outwnd.cpp
Goober5000: The if() block below looks "neater" with the variable the way it is, so I prefer it as is.  I'll give valgrind the benefit of the doubt, and cross that bridge if we come to it.

PR #4458, opened 2022-06-30
Warn on mismatched turret objects

### LINKED ISSUE #4204: Assert triggered by incorrect turret arm
```
ASSERTION: "vm_vec_is_normalized(plane_normal)" at vecmat.cpp:136
 plane_normal must be normalized!

Assert: "vm_vec_is_normalized(plane_normal)"
File: vecmat.cpp
Line: 136
plane_normal must be normalized!
```

[GTI_Asteria.zip](https://github.com/scp-fs2open/fs2open.github.com/files/8547459/GTI_Asteria.zip)

This was triggered by turret10's arm mistakenly assigned to turret11 in the above pof.

I assume not an urgent issue as poftools immediately notifies the issue, but as far as I understand an assert should not happen even with bad data, so here you go.

Baezon:
Gun objects must be an immediate child of their base object. Also replace the confusing `parent` `physical_parent` naming scheme that is literally only used right here.

Fixes #4204

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -2385,26 +2385,29 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
 				int n_banks = cfread_int(fp);			// Number of turrets
 
 				for ( i = 0; i < n_banks; i++ ) {
-					int parent;							// The parent subobj of the turret (the gun base)
-					int physical_parent;				// The subobj that the firepoints are physically attached to (the gun barrel)
 					int n_slots;						// How many firepoints the turret has
 					model_subsystem *subsystemp;		// The actual turret subsystem
 
-					parent = cfread_int( fp );
-					physical_parent = cfread_int(fp);
+					int base_obj = cfread_int(fp);		// The parent subobj of the turret (the gun base)
+					int gun_obj = cfread_int(fp);       // The subobj that the firepoints are physically attached to (the gun barrel)
+					
+					if (base_obj != gun_obj && pm->submodel[gun_obj].parent != base_obj) {
+						Warning(LOCATION, "Model %s turret %s has a gun submodel that is not an immediate child object of the base", pm->filename, pm->submodel[base_obj].name);
+						gun_obj = base_obj; // fall back to singlepart handling
+					}
 
 					int snum=-1;
 					if ( subsystems ) {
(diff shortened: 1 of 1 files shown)


code/model/modelread.cpp
```diff
@@ -2385,26 +2385,29 @@ int read_model_file(polymodel * pm, const char *filename, int n_subsystems, mode
...
 
-					parent = cfread_int( fp );
-					physical_parent = cfread_int(fp);
+					int base_obj = cfread_int(fp);		// The parent subobj of the turret (the gun base)
+					int gun_obj = cfread_int(fp);       // The subobj that the firepoints are physically attached to (the gun barrel)
+					
+					if (base_obj != gun_obj && pm->submodel[gun_obj].parent != base_obj) {
+						Warning(LOCATION, "Model %s turret %s has a gun submodel that is not an immediate child object of the base", pm->filename, pm->submodel[base_obj].name);
+						gun_obj = base_obj; // fall back to singlepart handling
+					}
```
Goober5000, line 2397: If either `base_obj` or `gun_obj` is -1, this error check will itself cause an error.

Goober5000 (changes requested):
Good refactoring.  One comment.

code/model/modelread.cpp
Baezon: They are indices into the array of submodels; it is certainly not valid. While it's true that that would itself cause an error, frankly, that sort of checking is out of scope of this PR. There are in fact quite a few rather flagrant assumptions being made about the pof file, some of which are much more likely than a submodel index being out of bounds, that I do intend to have FSO verify eventually.

Goober5000: If "it is certainly not valid" does that imply a guaranteed crash?

I'm ok with ignoring most flagrant assumptions, but exchanging an Assert for an array-out-of-bounds crash does not seem like an improvement.

Baezon: The index here was already being implicitly relied upon and indexed many times. All I'm doing is enforcing the gun object's parent, and if this gun object index is out of bounds, then a lot of unrelated stuff is going to break, with or without this warning.

Goober5000: Ok, fair enough.

Goober5000 (approved):

Issue #4226, opened 2022-04-27
set-subsystem-strength string variable


MjnMixael:
set-subsystem-strength does not accept a string variable for the ship name and the mission fails to load if you try it. Seems like this should be a thing you can do?

z64555:
It should be working with string for the ship name, there might be a bug with the cache system.

set-substem-strength takes several parameters: string ship_name, string subsystem_name, int strength_percent, bool repair_submodel (optional), bool repair_ancestor_submodel=true (optional)

Goober5000:
Works as expected for me.  This test case sets the player's sensors to 50% at T=3s, and the transport's sensors to 50% at T=6s.  And it uses a repeating event as well.

[set-subsystem-strength.zip](https://github.com/scp-fs2open/fs2open.github.com/files/8587101/set-subsystem-strength.zip)

MjnMixael:
My event was, admittedly, several conditionals deep but it specifically called out that sexp and variable. There is no doubt that that was the cause. I have since changed the entire sexp to work around it. Guess I'll have to try and recreate it.

MjnMixael:
[Test.zip](https://github.com/scp-fs2open/fs2open.github.com/files/8655526/Test.zip)
This retail compatible mission gives me an error about the string variable on save.

Goober5000:
Took a look at the mission.  This happens because the mission is checking the default value of the string, "na", as a ship name.  In most cases where a variable is used, the syntax checker will just look to see if the variable type matches (number = number, string = string) and that's it.  However, for a few complex types such as subsystems, the syntax checker delves a few layers deeper and these extra layer don't have the simplified variable check.

jg18:
Closing now that the fix is merged.

PR #4460, opened 2022-07-01
Handle syntax checking and caching for nodes with dynamic values

jg18:
Nodes whose values can change are any of special argument, variable, or container data.

We should avoid caching values from such nodes and also avoid attempting "deep inspection" while checking SEXP syntax for things like ship/subsystem names.

We can, however, still do some type checking.

Should fix Issue #4226.


jg18:
Since this PR also fixes the issue of (incorrectly) caching nodes that hold container data in some situations, it's a fix for containers and consequently should go into 22.2.

code/parse/sexp.cpp
```diff
@@ -2464,8 +2465,8 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 				}
 				Assert(ship_node >= 0);
 
-				// we can't check special-arg ships
-				if (Sexp_nodes[ship_node].flags & SNF_SPECIAL_ARG_IN_NODE)
+				// we can't check nodes whose value can change
+				if (can_node_value_change(node))
```
jg18, line 2469: I know I should check both here and in similar cases that the node is specifically one of a special argument, string variable, or string container data before skipping the check. I'll add that before I mark the PR as ready for review.

jg18:
Passes manual test.

I'm able to open in FRED the mission @MjnMixael posted on Issue #4226 without error.

Goober5000 (approved):
This looks good. 👍  One optional comment.

PR #4462, opened 2022-07-01
Fix gcc12 warnings

the-maddin:
Fixes the only two gcc12 warnings by being more explicit

### DIFF
--- a/code/fireball/fireballs.cpp
+++ b/code/fireball/fireballs.cpp
@@ -407,10 +407,15 @@ void fireball_parse_tbl()
 	{
 		if (fi.lod_count > 1)
 		{
+			Assertion(fi.lod_count < MAX_FIREBALL_LOD, "Fireball LOD (%d) greater than MAX_FIREBALL_LOD %d.", fi.lod_count, MAX_FIREBALL_LOD);
+
 			auto lod0 = fi.lod[0].filename;
 
+			if (strlen(lod0) > MAX_FIREBALL_BASENAME_LEN)
+				Error(__FILE__, __LINE__, "A fireball base file name may not be longer than %d characters due to the LOD naming scheme. (is %s)", MAX_FIREBALL_BASENAME_LEN, lod0);
+
 			for (int j = 1; j < fi.lod_count; ++j)
-				sprintf(fi.lod[j].filename, "%s_%d", lod0, j);
+				sprintf(fi.lod[j].filename, "%.*s_%d", MAX_FIREBALL_BASENAME_LEN, lod0, j % MAX_FIREBALL_LOD);
 		}
 	}
 
--- a/code/fireball/fireballs.cpp
+++ b/code/fireball/fireballs.cpp
@@ -412,7 +412,11 @@ void fireball_parse_tbl()
 			auto lod0 = fi.lod[0].filename;
 
 			if (strlen(lod0) > MAX_FIREBALL_BASENAME_LEN)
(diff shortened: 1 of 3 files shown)


code/fireball/fireballs.cpp
```diff
@@ -407,10 +407,15 @@ void fireball_parse_tbl()
 	{
 		if (fi.lod_count > 1)
 		{
+			Assertion(fi.lod_count < MAX_FIREBALL_LOD, "Fireball LOD (%d) greater than MAX_FIREBALL_LOD %d.", fi.lod_count, MAX_FIREBALL_LOD);
+
 			auto lod0 = fi.lod[0].filename;
 
+			if (strlen(lod0) > MAX_FIREBALL_BASENAME_LEN)
+				Error(__FILE__, __LINE__, "A fireball base file name may not be longer than %d characters due to the LOD naming scheme. (is %s)", MAX_FIREBALL_BASENAME_LEN, lod0);
```
Goober5000, line 415: This should be `Error(LOCATION` to be consistent with the rest of the code.  And in fact, rather than making this an `Error`, change it to a `Warning` and just skip past the invalid LOD.

code/fireball/fireballs.cpp
```diff
@@ -407,10 +407,15 @@ void fireball_parse_tbl()
...
+
 			for (int j = 1; j < fi.lod_count; ++j)
-				sprintf(fi.lod[j].filename, "%s_%d", lod0, j);
+				sprintf(fi.lod[j].filename, "%.*s_%d", MAX_FIREBALL_BASENAME_LEN, lod0, j % MAX_FIREBALL_LOD);
```
Goober5000, line 418: Instead of making a new #define which is only used here, change this to `MAX_FILENAME_LEN - 3`.  This makes explicit what is actually happening.

And remove `j % MAX_FIREBALL_LOD` since this obfuscates the assignment, and is unnecessary due to the newly added Assertion.

Goober5000 (changes requested):

code/fireball/fireballs.cpp
the-maddin: Will do. I'd like to point out, however, that in these cases there can be no single invalid LOD, but all LODs except LOD0 will be skipped, due to having too long file names.

code/fireball/fireballs.cpp
the-maddin: `j % MAX_FIREBALL_LOD` is the only possible and single part of this commit necessary to make gcc12 understand that the preceding format string is indeed safe in this context. Adding comment to clarify.
The intention of the #define was to make the LOD naming scheme obvious outside of the implementation, but I can oblige with a magic number.

code/fireball/fireballs.cpp
Goober5000: Good point.  Even so, I think a Warning is still appropriate, because even though the situation is not desirable, it is still recoverable.

code/fireball/fireballs.cpp
Goober5000: That's odd, but ok.

Magic numbers should be assessed on a case-by-case basis.  In this case, the need for the number is dependent on the format string, and the entire context is all on one line.  So I don't think a separate #define, let alone spreading out the context over multiple files, is the right thing to do.

Goober5000: Or, hmm, the context is in one scoped block, rather than one line.  If you wanted to put a `constexpr` above line 414, that would work.

Goober5000 (approved):
Cool, looks good.

PR #4463, opened 2022-07-02
Clamp values to their correct ranges when loading and saving pilot and campaign files

### LINKED ISSUE #3321: Graphical glitch/assert when opening options menu as a specific pilot
When I try to open the options menu with a certain pilot, in debug builds I get an assert, and in release builds I get a graphical glitch.
My other pilots aren't affected. I was able to access the menu without a glitch or crash using this pilot earlier.
I did this all with the March 24th 2021 nightly. I still get the graphical glitch and assert with the borked pilot and only the borked pilot in 21.0.0. I haven't yet been able to break another pilot file in this way.
[Here are the pilot's files.](https://github.com/scp-fs2open/fs2open.github.com/files/6198781/BorkedPilot.zip)
[Here's a log.](https://fsnebula.org/log/605b4c79b3f5418863d7bf59)

Kiloku:
Should fix #3321. Tested with the provided save files and the defective value was clamped as expected. 

Also tried forcing a few wrong values by manually editing a JSON pilot file and playing it in the RC, which saved the wrong values to the campaign file, and then loaded the same pilot/campaign in my dev build, resulting in the values being corrected.

### DIFF
--- a/code/pilotfile/csg.cpp
+++ b/code/pilotfile/csg.cpp
@@ -1209,23 +1209,35 @@ void pilotfile::csg_read_settings()
 void pilotfile::csg_write_settings()
 {
 	startSection(Section::Settings);
+	clamped_range_warnings.clear();
 
 	// sound/voice/music
+	clamp_value_with_warn(&Master_sound_volume, 0.f, 1.f, "Effects Volume");
 	cfwrite_float(Master_sound_volume, cfp);
+	clamp_value_with_warn(&Master_event_music_volume, 0.f, 1.f, "Music Volume");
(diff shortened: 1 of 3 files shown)


Goober5000 (approved):
LGTM

JohnAFernandez (commented):
This really is good work. I think if you have the time, changing the variable names to the UI settings names in the messageswould make more sense, since users have to make changes to their settings.

Kiloku:
Sure! I can do that later tonight or tomorrow morning

JohnAFernandez:
I think we're good! 👍🏼

PR #4464, opened 2022-07-02
add x500 and x1000 movement speeds

Goober5000:
As requested by @Baezon.


JohnAFernandez (approved):
I didn't realize we added scan codes manually.  What a mess.

PR #4469, opened 2022-07-06
Voice acting manager enhancements

Goober5000:
Four new functions relating to synchronizing personas...
1) Copy personas from messages to ships
2) Copy personas from ships to messages
3) Clear personas from ships that don't send messages
4) Set head ANIs using personas in messages.tbl

### DIFF
--- a/fred2/voiceactingmanager.cpp
+++ b/fred2/voiceactingmanager.cpp
@@ -291,7 +291,7 @@ CString VoiceActingManager::generate_filename(CString section, int number, int d
 				sender[j] = '_';
 		}
 
-		// flatten muliple underscores
+		// flatten multiple underscores
 		j = 1;
 		while( sender[j] != '\0' ) {
 			if ( sender[j-1] == '_' && sender[j] == '_' ) {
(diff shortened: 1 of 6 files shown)


fred2/voiceactingmanager.cpp
```diff
@@ -943,3 +951,176 @@ void VoiceActingManager::OnBnClickedIncludeSender()
...
+		get_valid_sender(sender_buf, NAME_LENGTH, message, &sender_shipnum);
+
+		// if it's a ship, copy the persona
+		if (sender_shipnum >= 0)
+		{
+			auto sender_shipp = &Ships[sender_shipnum];
+			auto persona_to_copy = (messages_to_ships) ? message->persona_index : sender_shipp->persona_index;
+
+			// don't copy None, and only copy wingman personas
+			if (persona_to_copy >= 0 && (Personas[persona_to_copy].flags & PERSONA_FLAG_WINGMAN))
```
wookieejedi, line 978: I take it the reason only wingmen personas are copied here (and deleted for later on) is b/c there are many other checks throughout message code that limit things to just wingmen personas?

Goober5000: Well, that may be related. My reason is that there is only one Terran Command persona, so mission designers could conceivably use it for any Terran head animation, or all of them, in one mission.  The persona isn't standardized in the way that the wingmen are.

wookieejedi: Ah, gotcha. There's also the possibility of mods that have more than one command personas for a given species, but I take it you have thought of that already.

Goober5000: Yep.  If mods created multiple Command personas and used them for different missions, and if mods were consistent with the use of head animations, then it would be appropriate to include them.  But occasionally mods will use a Command persona with a different head animation, or assign the Command persona to a ship.  That would result in undesirable side effects if the voice manager dialog processed those personas.

wookieejedi: Makes sense, thanks for the extra explanation.

wookieejedi (approved):

Issue #4405, opened 2022-06-14
FRED2 hotkeys in event editor allow impossible (and crash-causing) sexp combinations


MoerasGrizzly:
FRED2 hotkeys in event editor allow impossible (and crash-causing) sexp combinations.

Steps to reproduce:
Copy a sexp.
Right click another sexp
When paste (overwrite) is not possible, press the hotkey for paste (overwrite).
when paste (add child) is not possible, press the hotkey for paste (add child).

Whilst I'm here anyway, the "delete" hotkey never works in the event editor.

jg18:
There are related issues with being able to use these keyboard shortcuts when you shouldn't be.

For example, if you try Ctrl+V in the Events Editor on a node when nothing is on the clipboard, FRED crashes with an assertion failure.

Not sure yet what's going on with the keyboard shortcut for Delete Item not working.

jg18:
Looks like this'll be a really hard problem to fix, due to the hackish nature of the FRED codebase (see FRED's About dialog).

The code that determines which of Delete Item/Cut/Copy/Paste/Add-Paste should be enable is tightly interwoven with the code that builds the right-click menu.

I looked into pulling the former apart from the latter and ultimately gave up, deciding that it'd make a bad situation much worse.

While the reported issue is definitely legit, the conclusion may have to be that FRED users use these keyboard shortcuts "at their own risk" and it's assumed they know what they are doing.

If FREDders use the context menu only, then it should be guaranteed that the menu options are enabled only when they're applicable.

It's probably best to not even add keyboard support for "Delete Item". Since there'll be no safeguards that it'd be usable only when it should be, and since the action is extremely difficult to reverse. we don't want to make it extra easy for people to inadvertently delete stuff.

The best solution there is probably to remove the keyboard shortcut for Delete from the UI, so it's clear that deleting by keyboard shortcut isn't supported.

jg18:
Unfortunately, I think we've hit the limit on what we can fix during this RC cycle.

Further fixes will require more sweeping (and thus riskier) changes and consequently are better done sometime after a stable release is out.

Even though we can't guarantee any ETA for a fix, we appreciate the report, especially given repro instructions.

z64555:
Punting to 22.4 upon jg18's recommendation.

TRBlount:
@jg18 Should we keep this on the board for 23.0 or shift it to 23.2?

Goober5000:
Bumping to 23.2.

[...]

PR #4470, opened 2022-07-06
Provide a limited partial fix for FRED keyboard shortcuts

jg18:
Makes these changes in FRED's Events Editor:
1. Ensures the keyboard shortcuts for Paste/Add-Paste won't trigger a crash if the clipboard is empty.
2. Remove "Delete" from UI as a keyboard shortcut for "Delete Item". It never worked, and as explained in the related issue, it shouldn't be made to.

Intended as a very limited partial fix for Issue #4405 until someone can do a proper rewrite of `sexp_tree::right_clicked()`.

### DIFF
--- a/fred2/fred.rc
+++ b/fred2/fred.rc
@@ -445,7 +445,7 @@ IDR_MENU_EDIT_SEXP_TREE MENU
 BEGIN
     POPUP "Edit sexp tree"
     BEGIN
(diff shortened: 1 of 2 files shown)


jg18:
I suppose I should look into making similar changes to qtFRED, if applicable. check.

JohnAFernandez (approved):
Simple

Goober5000 (approved):
Indeed

PR #4473, opened 2022-07-07
address issues causing network connectivity failures on macOS

notimaginative:
Fixes network connectivity issues on macOS due to a mismatch of struct sizes, which macOS appears to be rather particular about. Successfully tested on macOS, Linux, and Windows.


Goober5000 (approved):

JohnAFernandez (approved):
And once again, Taylor is our hero

PR #4474, opened 2022-07-07
Removed assertion that prevents stationary homing weapons

EatThePath:
Legalize standing still


JohnAFernandez (changes requested):
We still need to ensure this code is working correctly.  What if the speed is negative somehow?

EatThePath:
@JohnAFernandez I recall this came up in your MP testing initally, have you learned anything more about this case or have any opinion on if this is worth pursuing further?

PR #4475, opened 2022-07-07
fix erroneous position/angle modification

Goober5000:
Due to rounding and conversion approximations, it was possible for FRED to think the position and/or angle was modified when no such modification had occurred.  This could result in erroneous updates.  In the case of angles this could result in a group of ships being assigned the wrong orientation.

Follow-up to #2614.

### DIFF
--- a/fred2/orienteditor.cpp
+++ b/fred2/orienteditor.cpp
@@ -186,7 +186,7 @@ BOOL orient_editor::OnInitDialog()
 /**
  * Checks whether the position value is close enough to the input string value that we can assume the input has not changed.
  */
-bool orient_editor::is_close(float val, const CString &input_str)
+bool orient_editor::is_close(float val, const CString &input_str) const
 {
 	val = perform_input_rounding(val);
 	float input_val = convert(input_str);
@@ -198,7 +198,7 @@ bool orient_editor::is_close(float val, const CString &input_str)
 /**
  * Checks whether the angle value is close enough to the input string value that we can assume the input has not changed.
  */
-bool orient_editor::is_angle_close(float rad, const CString &input_str)
+bool orient_editor::is_angle_close(float rad, const CString &input_str) const
 {
 	float deg = perform_input_rounding(to_degrees(rad));
 	float input_deg = normalize_degrees(convert(input_str));
@@ -391,7 +391,7 @@ float orient_editor::normalize_degrees(float deg)
 /**
  * Extract a float from the CString, being mindful of any comma separators.
  */
-float orient_editor::convert(const CString &str)
+float orient_editor::convert(const CString &str) const
 {
 	char buf[256];
 	size_t i, j, len;
@@ -411,7 +411,7 @@ float orient_editor::convert(const CString &str)
  * This accounts for not only decimal rounding to the precision of the input box, but also floating point rounding due to inexact fractions such as 1/10.
  * See also GitHub PR #4475.
  */
-float orient_editor::perform_input_rounding(float val)
+float orient_editor::perform_input_rounding(float val) const
 {
 	CString str;
 	str.Format(INPUT_FORMAT, val);
--- a/fred2/orienteditor.cpp
+++ b/fred2/orienteditor.cpp
@@ -182,9 +182,27 @@ BOOL orient_editor::OnInitDialog()
 	return TRUE;
 }
 
-bool orient_editor::close(float val, const CString &str)
+bool orient_editor::close(float val, const CString &input_str)
 {
-	float diff = val - convert(str);
+	CString str;
+	str.Format("%.01f", val);
+	val = convert(str);
+
+	float input_val = convert(input_str);
+
+	float diff = val - input_val;
+	return diff > -input_threshold && diff < input_threshold;
+}
+
+bool orient_editor::angle_close(float rad, const CString &input_str)
+{
+	CString str;
+	str.Format("%.01f", to_degrees(rad));
+	float deg = convert(str);
+
+	float input_deg = normalize_degrees(convert(input_str));
+
+	float diff = deg - input_deg;
 	return diff > -input_threshold && diff < input_threshold;
 }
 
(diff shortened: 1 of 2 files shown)


fred2/orienteditor.cpp
```diff
@@ -182,9 +182,27 @@ BOOL orient_editor::OnInitDialog()
 	return TRUE;
 }
 
-bool orient_editor::close(float val, const CString &str)
+bool orient_editor::close(float val, const CString &input_str)
 {
-	float diff = val - convert(str);
+	CString str;
+	str.Format("%.01f", val);
```
jg18, line 188: A comment explaining why we're doing this would be nice IMO, maybe even citing the related GH Issue.

fred2/orienteditor.cpp
```diff
@@ -182,9 +182,27 @@ BOOL orient_editor::OnInitDialog()
...
+	val = convert(str);
+
+	float input_val = convert(input_str);
+
+	float diff = val - input_val;
+	return diff > -input_threshold && diff < input_threshold;
```
jg18, line 194: I was going to suggest changing to
```
return abs(diff) < input_threshold;
```
which IMO is easier to read, but looks like the code already uses the comparison used here.

fred2/orienteditor.cpp
```diff
@@ -182,9 +182,27 @@ BOOL orient_editor::OnInitDialog()
...
 
-bool orient_editor::close(float val, const CString &str)
+bool orient_editor::close(float val, const CString &input_str)
 {
-	float diff = val - convert(str);
+	CString str;
+	str.Format("%.01f", val);
+	val = convert(str);
+
+	float input_val = convert(input_str);
```
jg18, line 191: Is it a problem that we don't know the precision of the value encoded in `input_str`, given that we're enforcing a precision on `val` via `str`? Or do we know?

fred2/orienteditor.cpp
```diff
@@ -182,9 +182,27 @@ BOOL orient_editor::OnInitDialog()
...
+	return diff > -input_threshold && diff < input_threshold;
+}
+
+bool orient_editor::angle_close(float rad, const CString &input_str)
+{
+	CString str;
+	str.Format("%.01f", to_degrees(rad));
+	float deg = convert(str);
+
+	float input_deg = normalize_degrees(convert(input_str));
```
jg18, line 203: Why the different calls to `to_degrees()` vs. `normalize_degrees()`?

fred2/orienteditor.cpp
```diff
@@ -182,9 +182,27 @@ BOOL orient_editor::OnInitDialog()
...
+	float input_val = convert(input_str);
+
+	float diff = val - input_val;
+	return diff > -input_threshold && diff < input_threshold;
+}
+
+bool orient_editor::angle_close(float rad, const CString &input_str)
+{
+	CString str;
+	str.Format("%.01f", to_degrees(rad));
```
jg18, line 200: Same comments as above that an explanation here would be nice and also if it's a concern that `input_rad` doesn't seem to get formatted this way.

fred2/orienteditor.cpp
jg18: Also, since we use this call below, perhaps a function like
```
CString normalize_float(float value) const;
```
(or equivalent if MFC doesn't like handling `CString` objects that way) but maybe that's overkill.

jg18 (changes requested):
Would appreciate an extra comment explaining the logic, see comments for details, but looks good overall. A few other questions/thoughts.

jg18:
Also, to check, have you tested this change?

fred2/orienteditor.cpp
Goober5000: Ok.

fred2/orienteditor.cpp
Goober5000: Eh, might as well.

fred2/orienteditor.cpp
Goober5000: The problem is that we do know the precision, but due to a combination of format rounding and floating point rounding, the values can drift apart.  Hopefully the new commit clarifies this.

fred2/orienteditor.cpp
Goober5000: One is converting units and then normalizing; the other is just normalizing.

fred2/orienteditor.cpp
Goober5000: See upcoming commit.

Goober5000:
Yup, I tested it.  I noticed this bug when I was editing a mission with a ship whose angle was set to 32.1, and FRED kept thinking I modified it when I didn't it.  During the course of the PR I tested both position and angle input.

Anyway, see the latest commit.

fred2/orienteditor.cpp
```diff
@@ -21,7 +21,8 @@
 static char THIS_FILE[] = __FILE__;
 #endif
 
-const float input_threshold = 0.01f;		// smallest increment of input box
+constexpr auto INPUT_THRESHOLD = 0.01f;		// smallest increment of input box
+constexpr auto INPUT_FORMAT = "%.01f";
```
jg18, line 25: Based on [this page](https://medium.com/@samathy_barratt/constantly-confusing-c-const-and-constexpr-pointer-behaviour-7eddb041e15a), `constexpr char *` is equivalent to `char * const` and not `const char *`.

I think either `const auto` or `const auto const` would be better.

fred2/orienteditor.h
```diff
@@ -70,7 +71,10 @@ class orient_editor : public CDialog
 
 private:
 	float convert(const CString &str);
-	bool close(float val, const CString &str);
+	float perform_input_rounding(float val);
+
+	bool is_close(float val, const CString &input_str);
+	bool is_angle_close(float rad, const CString &input_str);
```
jg18, line 77: Optional: maybe make these functions `const`, although as you know, `const` correctness can trigger a ripple effect, so maybe not worth it.

jg18 (changes requested):
Thanks, I'm a fan. :+1: Just one last thing, plus one optional thing.

fred2/orienteditor.cpp
Goober5000: That page covers `constexpr char *`, but not `constexpr auto`.  String literals, written out in the code, always have the type of `const char *`.  So I'm pretty sure that `auto` would interpret them correctly.

fred2/orienteditor.h
Goober5000: Ok.

fred2/orienteditor.cpp
jg18: Looks like Visual Studio's tool tip on similar code agrees with you. So `constexpr auto` it is.

jg18 (approved):
Looks good. As usual, thanks for working with me through the comments. 🙂

PR #4476, opened 2022-07-08
use proper bounds check when bashing hull in red-alert

### LINKED ISSUE #4472: Red Alert hull and ammo values are carrying over incorrectly
This issue was identified in the Procyon Insurgency mod, where there is a string of four red alert missions. 

When the player finishes each mission, they'll start the next one with the lowest of `+Initial Hull:` and their hull at the end of the previous mission, however the expected behavior is to start with the highest of these values. 
@Goober5000 identified that the issue was caused by a combination of the [593be42e66](https://github.com/scp-fs2open/fs2open.github.com/commit/593be42e66) and [a04e83ab01](https://github.com/scp-fs2open/fs2open.github.com/commit/a04e83ab01) commits.

Goober5000:
This range check is to ensure that the ship's hull is in a valid range, so it should therefore check that the hull is less than the max strength, not less than the current strength.  This was inadvertently changed in 593be42e66 and a04e83ab01, but retail behavior is to set the exact hull from the previous mission.

Tested for both regular ships and delayed-arrival ships.  Fixes #4472.

### DIFF
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -885,8 +885,14 @@ void red_alert_bash_wingman_status()
 							mprintf(("Invalid ship class specified in red alert data for ship %s. Using mission defaults.\n", shipp->ship_name));
 					}
 
+					float max_hull;
+					if (shipp->special_hitpoints)
+						max_hull = shipp->ship_max_hull_strength;
+					else
+						max_hull = Ship_info[shipp->ship_info_index].max_hull_strength;
+
 					// restore hull (but not shields)
-					if (ras->hull >= 0.0f && ras->hull <= ship_objp->hull_strength)
+					if (ras->hull >= 0.0f && ras->hull <= max_hull)
 						ship_objp->hull_strength = ras->hull;
 					else
 						mprintf(("Invalid health in red alert data for ship %s. Using mission defaults.\n", shipp->ship_name));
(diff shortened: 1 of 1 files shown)


code/missionui/redalert.cpp
```diff
@@ -885,8 +885,14 @@ void red_alert_bash_wingman_status()
 							mprintf(("Invalid ship class specified in red alert data for ship %s. Using mission defaults.\n", shipp->ship_name));
 					}
 
+					float max_hull;
+					if (shipp->special_hitpoints)
+						max_hull = shipp->ship_max_hull_strength;
+					else
+						max_hull = Ship_info[shipp->ship_info_index].max_hull_strength;
+
```
jg18, line 893: Would an `Assertion()` that `max_hull` is something sane be useful, say `> 0.0f`? Ditto for similar-looking code below.

code/missionui/redalert.cpp
```diff
@@ -979,9 +985,15 @@ void red_alert_bash_wingman_status()
 						}
 					}
 
+					float max_hull;
+					if (pobjp->special_hitpoints)
+						max_hull = pobjp->ship_max_hull_strength;
+					else
+						max_hull = Ship_info[pobjp->ship_class].max_hull_strength;
```
jg18, line 992: Since this code appears twice, perhaps a `get_max_hull()` function would be useful? Or are `pobjp` here and `shipp` above different types?

code/missionui/redalert.cpp
```diff
@@ -979,9 +985,15 @@ void red_alert_bash_wingman_status()
...
+
 					// restore hull (but not shields)
-					if (ras->hull >= 0.0f && ras->hull <= (pobjp->initial_hull * pobjp->ship_max_hull_strength / 100.0f))
```
jg18, line 983: I wonder what was up with this complicated formula.

jg18 (approved):
I'm not super familiar with this code, but this looks reasonable to me. A few optional thoughts.

code/missionui/redalert.cpp
Goober5000: I don't think this is necessary.  It already happens before this in the code flow.

code/missionui/redalert.cpp
Goober5000: They are different types.

code/missionui/redalert.cpp
Goober5000: For parse objects, the initial hull is stored as a number between 0 and 100.  For ships that are present, the hull is stored as the actual number of hitpoints.  This converts from the first to the second.

PR #4479, opened 2022-07-08
Allow turrets to use their firepoints for target-leading calculations

### LINKED ISSUE #4288: FR: Flag to account for turret arm length when leading targets
I have anti-fighter turrets with firepoints over 400 meters away from their centers, it makes a difference.

Kiloku:
Most turrets can lead just fine without this, but in use cases where the firepoint is far from the center of the turret's "body" (an extra long barrel, for example) this can cause the turret to not aim correctly.  

This PR adds a ship flag that can be added to turret subsystems to enable it for them directly, as well as a game settings option to enable it for all turrets. 

Resolves #4288

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -108,6 +108,7 @@ bool Always_warn_player_about_unbound_keys;
 leadIndicatorBehavior Lead_indicator_behavior;
 shadow_disable_overrides Shadow_disable_overrides {false, false, false, false};
 float Thruster_easing;
+bool Always_use_distant_firepoints;
 
 void mod_table_set_version_flags();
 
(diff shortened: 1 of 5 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -92,6 +92,7 @@ bool Show_subtitle_uses_pixels;
 int Show_subtitle_screen_base_res[2];
 int Show_subtitle_screen_adjusted_res[2];
 bool Always_warn_player_about_unbound_keys;
+bool Always_use_distant_firepoints;
```
BMagnu, line 95: You need an additional entry resetting this value to false in ``mod_table_reset()``, since otherwise it'd be totally unset when the table option isn't specified. This _should_ be false for booleans, but it's not proper style to rely on that.

BMagnu (changes requested):
LGTM, mostly, apart from one comment.
As far as I can tell, stuffing the actual firing point over just the turret point for these functions and globals set within these functions is pretty much always correct, but since it can affect AI balance putting it behind a flag is the right thing to do.

EatThePath:
I'll add I've tested this some in gameplay and it does seem to fix what I needed it to fix

BMagnu (approved):

Issue #2157, opened 2019-11-25
ship-maneuver changes cause ludicrous speed


MjnMixael:
BtA's mission Rebels and Revolutionaries displays this problem.

Oct 24 build, the Seeker ships fly at a normal speed (about 20 m/s). This is done with a ship-maneuver sexp to tell the ship to fly forward at 100% velocity while rotating slightly.

After Oct 26 changes submitted by @Goober5000 the Seeker ships fly off into the sunset at +160 m/s.

Goober5000:
Screenshot...
https://cdn.discordapp.com/attachments/223511363807346691/648545575381237781/screen0037.png

Goober5000:
I tried the most recent build from source, and that worked perfectly fine.  Next I tried the 20191112 build, and that worked too.  Finally I tried the 20191026 build, and that also worked.  In all three builds, the ship went 20 m/s as expected.  I confirmed that I was running the proper builds by checking the build ID string in the main hall.

Goober5000:
Is this still a problem?

MjnMixael:
Unsure. It has been on my list to look in to again and I just haven't.

z64555:
Is the rotation on any one axis? Bank, perhaps?

JohnAFernandez:
We don't have reproduction and MJN has too much on his plate in RL to really look at this.  I don't think we can easily fix this right now in the next few weeks.  

But if anyone *is* looking into it, I would suggest looking at the fact that the object was docked to *30* other objects and not just looking at the sexp.  I'm not sure if @Baezon's changes to the moment of inertia for docked objects may have inadvertently fixed this as well, because notice that the sexp was slightly turning the object.  

But like I said, I just don't think we can fix this without being able to reproduce it, and it may be a while before MJN or someone else can give us a good handle on how to do that.

Baezon:
I'm with Cyborg here, without a reliable means to produce it there's not much we can do. I don't feel confident enough to close it, but I don't think it's immediately urgent enough to warrant a high priority for the stable.

z64555:
Alright, bumping it down to Low for now

JohnAFernandez:
BTA team has not seem to encountered it recently.   Closing until there is a new report.

MjnMixael:
This showed back up again for a player. [See here.](https://discord.com/channels/223511295431933953/223521053517938688/921161090044428338)

[...]

PR #4483, opened 2022-07-10
Fix some bugs with maneuvers

Baezon:
Two things: split `ai_control_info_check` into `ai_control_info_flags_upkeep`, which keeps track of the flags and sets or resets them accordingly and `ai_control_info_apply` which then uses those flags and sets the directional values. `ai_control_info_flags_upkeep` is then made unconditional, even if the AI mode dictates it should not act on maneuver overrides, keep track of the flags and ensure they set and reset properly. Additionally, `player-not-use-ai` completely resets maneuvers, preventing any accidental 'leakage' of its flags. Also makes clear that `PF_NO_DAMP` is only used for maneuvers. 

This is possibly, but not certainly, related to #2157

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -14763,20 +14763,16 @@ static void ai_control_info_apply(ai_info *aip)
 	if (aip->ai_override_flags.none_set())
 		return;
 
-	bool no_lat = false;
-	bool no_rot = false;
+	bool lateral_maneuver = false;
+	bool rotational_maneuver = false;
 
-	if (!aip->ai_override_flags[AI::Maneuver_Override_Flags::Lateral_never_expire] && timestamp_elapsed(aip->ai_override_lat_timestamp))
-		no_lat = true;
+	if (aip->ai_override_flags[AI::Maneuver_Override_Flags::Lateral_never_expire] || !timestamp_elapsed(aip->ai_override_lat_timestamp))
+		lateral_maneuver = true;
 
-	if (!aip->ai_override_flags[AI::Maneuver_Override_Flags::Rotational_never_expire] && timestamp_elapsed(aip->ai_override_rot_timestamp))
-		no_rot = true;
+	if (aip->ai_override_flags[AI::Maneuver_Override_Flags::Rotational_never_expire] || timestamp_elapsed(aip->ai_override_rot_timestamp))
+		rotational_maneuver = true;
 
-	if (no_rot && no_lat) {
-		aip->ai_override_flags.reset();
-	}
-	else
-	{
+	if (lateral_maneuver || rotational_maneuver) {
 		if (aip->ai_override_flags[AI::Maneuver_Override_Flags::Full_rot])
 		{
 			AI_ci.pitch = aip->ai_override_ci.pitch;
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -14710,7 +14710,8 @@ void ai_frame(int objnum)
 	}
 }
 
-static void ai_control_info_check(ai_info *aip, ship_info *sip, physics_info *pi)
+// set and removes any maneuver flags for ai maneuver overrides
+static void ai_control_info_flags_upkeep(ai_info* aip, ship_info* sip, physics_info* pi)
 {
 	if (aip->ai_override_flags.none_set())
 		return;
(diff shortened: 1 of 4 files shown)


code/ai/aicode.cpp
```diff
@@ -14736,6 +14737,41 @@ static void ai_control_info_check(ai_info *aip, ship_info *sip, physics_info *pi
...
+	}
+	else
+	{
+		if (aip->ai_override_flags[AI::Maneuver_Override_Flags::Dont_bank_when_turning] || sip->flags[Ship::Info_Flags::Dont_bank_when_turning])
+			AI_ci.control_flags |= CIF_DONT_BANK_WHEN_TURNING;
+		if (aip->ai_override_flags[AI::Maneuver_Override_Flags::Dont_clamp_max_velocity] || sip->flags[Ship::Info_Flags::Dont_clamp_max_velocity])
+			AI_ci.control_flags |= CIF_DONT_CLAMP_MAX_VELOCITY;
+		if (aip->ai_override_flags[AI::Maneuver_Override_Flags::Instantaneous_acceleration] || sip->flags[Ship::Info_Flags::Instantaneous_acceleration])
+			AI_ci.control_flags |= CIF_INSTANTANEOUS_ACCELERATION;
+	}
```
Goober5000, line 14751: If I'm reading this correctly, it looks like these ship_info flags will no longer be checked if there is no maneuver in progress.  This would break some existing functionality; for example ground vehicles shouldn't bank when turning even if they are following waypoints.

The AI control flags are intended to be set if there are maneuver flags *or* if the ship class entry has flags.

code/physics/physics.h
```diff
@@ -32,7 +32,7 @@
 #define PF_GLIDING				(1 << 14)
 #define PF_FORCE_GLIDE			(1 << 15)
 #define PF_NEWTONIAN_DAMP		(1 << 16)	// SUSHI: Whether or not to use newtonian dampening
-#define PF_NO_DAMP				(1 << 17)	// Goober5000 - don't damp velocity changes in physics; used for instantaneous acceleration
```
Goober5000, line 35: Per above comment, the `PF_NO_DAMP` flag ought to be usable for more than just maneuvers, just like the `PF_NEWTONIAN_DAMP` flag is.

Goober5000 (commented):
Asking for elaboration on a few parts.

code/ai/aicode.cpp
Baezon: This might have been buggy before then, as this code was only run for maneuvers even before this. I will double-check this behavior.

code/physics/physics.h
Baezon: I would prefer a separate flag for any new functionality like this. Otherwise, `player-not-use-ai` would be wiping an unrelated flag. It is only because it is currently only used by maneuvers that it can be safely reset by `player-not-use-ai`.

code/ai/aicode.cpp
Baezon: I can confirm that `Instantaneous_acceleration` doesn't work on ships which aren't players or using maneuvers. This makes sense, since AI don't use CI, at least rotationally. For that, it would need to be in `ai_turn_towards_vector`, which `Dont_bank_when_turning` does correctly and still works. Laterally, however, there's no reason that it shouldn't work, and so should definitely be before the initial early return. (`Dont_clamp_max_velocity` also doesn't work, but is a bit of a weird one, AI never ask for more lateral speed than they have, but should this flag mean they should accelerate indefinitely as they are able???)

I would be in favor of these changes to make these flags work more generally for AI, but maybe not so late in the 'RC game' in this PR.

Goober5000: Mmh.  Well, that's fair, especially as the mission for which I need these flags isn't released yet and probably won't be released before 22.4 is out either.  And so far I have only tested these flags on player-controlled ships.  I'll open a new issue.

code/physics/physics.h
Goober5000: What, you mean like `PF_MANEUVER_NO_DAMP` and `PF_AI_NO_DAMP`?  That seems awkward, since physics should be independent of AI.

Is there no way that `sexp_player_use_ai` can conditionally remove that flag?

Baezon: Not easily, there's no way to know how the flag was applied, either due to an active maneuver or some other method. If you, say, wanted a ship to fundamentally never have damping via this flag, you could simply add a conditional that would preclude it from removing it if it had that ship flag. But having it as a separate flag would mean you wouldn't have to remember those places and accordingly update them.

Goober5000: Mmh.  That would be hacky, but probably the most efficient solution.

Goober5000 (approved):
Approved, though I'll be back to bug you about #4484 in the next release cycle

Baezon:
As mentioned, I am *not sure at all* this resolves #2157. And it should yet be closed after this is merged.

JohnAFernandez:
No one is sure.  But we are addressing what we can.

code/ai/aicode.cpp
```diff
@@ -14736,6 +14737,41 @@ static void ai_control_info_check(ai_info *aip, ship_info *sip, physics_info *pi
...
+	bool no_rot = false;
+
+	if (!aip->ai_override_flags[AI::Maneuver_Override_Flags::Lateral_never_expire] && timestamp_elapsed(aip->ai_override_lat_timestamp))
+		no_lat = true;
+
+	if (!aip->ai_override_flags[AI::Maneuver_Override_Flags::Rotational_never_expire] && timestamp_elapsed(aip->ai_override_rot_timestamp))
+		no_rot = true;
+
 	if (no_rot && no_lat) {
 		aip->ai_override_flags.reset();
```
JohnAFernandez, line 14776: Is this now redundant since all ships go through what seems to be the same check in `ai_control_info_flags_upkeep`?

Baezon: Ah, yes, that is part of upkeeping flags so there's no need for it here.

PR #4486, opened 2022-07-12
Fix type checking and FRED help for containers in special argument lists

jg18:
Enforce the restriction that using `validate-argument` and `invalidate-argument` with container arguments isn't supported.

However, `validate-all-arguments` and `invalidate-all-arguments` apply to container arguments as well.

These constraints are listed in the [containers forum thread](https://www.hard-light.net/forums/index.php?topic=98209.msg1914261#msg1914261) (see Feature 6), which is currently the only documentation for containers.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -11029,7 +11029,9 @@ void sexp_change_argument_validity(int n, bool invalidate)
 		// first we must check if the arg_handler marks a selection. At the moment random-of is the only one that does this
 		arg_n = CDR(arg_handler);
 		while (invalidate && (arg_n != -1)) {
-			if (Sexp_nodes[arg_n].flags & SNF_ARGUMENT_SELECT) {
+			// (in)validating container arguments isn't supported
+			if ((Sexp_nodes[arg_n].subtype != SEXP_ATOM_CONTAINER_NAME) &&
+				Sexp_nodes[arg_n].flags & SNF_ARGUMENT_SELECT) {
 				// now check if the selected argument matches the one we want to invalidate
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -11029,7 +11029,9 @@ void sexp_change_argument_validity(int n, bool invalidate)
 		// first we must check if the arg_handler marks a selection. At the moment random-of is the only one that does this
 		arg_n = CDR(arg_handler);
 		while (invalidate && (arg_n != -1)) {
-			if (Sexp_nodes[arg_n].flags & SNF_ARGUMENT_SELECT) {
+			// (in)validating container arguments isn't supported
+			if ((Sexp_nodes[arg_n].subtype != SEXP_ATOM_CONTAINER_NAME) &&
```
Goober5000, line 11033: `SEXP_ATOM_CONTAINER_NAME` isn't supported, but `SEXP_ATOM_CONTAINER_DATA` is?

jg18: No, it isn't supported, either.

Support for using Replace Container Data in argument list was removed from FRED in PR #4392 but I forgot then to add a check in `check_sexp_syntax()`. The new commit fixes that.

Also, it occurred to me that since the argument type for `(in)validate-argument` is `OPF_ANYTHING`, which excludes containers (see `check_sexp_syntax()`), an `Assertion()` makes more sense than an `if`-check, so that's now fixed, too.

Goober5000: Ok.

Goober5000 (approved):

PR #4489, opened 2022-07-13
Simplify logic here

JohnAFernandez:
Break when you've found a match, instead of just mindlessly comparing the rest.

This is technically just an optimization and safety change, but if the inner and outer loops had been switched, it would have set the wrong flag if two copies of the same flag were listed in the mission file.


JohnAFernandez:
I'm going to make sure this gets in before 24.2

wookieejedi (approved):
Nice and concise

JohnAFernandez:
I was able to run a test of the changed snippet in an online compiler today, and based on that and reading the PR again, I feel confident that this is good to merge.

PR #4490, opened 2022-07-13
Allow multilock between traitors

JohnAFernandez:
Use the proper attacking function here, as traitors need to be able to attack each other in dogfight mode.


Goober5000 (changes requested):
Shouldn't this be `!attacks`?

JohnAFernandez:
Ah, thanks.  I forgot it was an early out.  I can't really test it at the moment, which is why it's in draft.

Goober5000 (approved):
Looks good, though a test before merge does make sense.

JohnAFernandez:
Closes #4493

JohnAFernandez:
Finally tested and confirmed fixed.

PR #4491, opened 2022-07-13
some fixes for playing movies via script

Goober5000:
The movie player code was missing some important functionality, in particular:
1) There was only one timing method to get real-time elapsed time, which was not suitable for pausing or time compression
2) There was an improper cast for the FPS property, causing a script error
3) The `getSeconds` method lost precision during casting

This patches those deficiencies.

### DIFF
--- a/code/scripting/api/libs/time_lib.cpp
+++ b/code/scripting/api/libs/time_lib.cpp
@@ -10,12 +10,12 @@ namespace api {
 
 ADE_LIB(l_Time, "Time", "time", "Real-Time library");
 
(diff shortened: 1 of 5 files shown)


code/scripting/api/libs/time_lib.cpp
```diff
@@ -15,5 +15,10 @@ ADE_FUNC(getCurrentTime, l_Time, nullptr, "Gets the current real-time timestamp.
 	return ade_set_args(L, "o", l_Timestamp.Set(timer_get_nanoseconds()));
```
JohnAFernandez, line 15: I think the documentation should be more explicit for this.  Something like "Gets the current real-time timestamp, not taking into account mission pause or UI use."

JohnAFernandez (approved):

PR #4498, opened 2022-07-14
Bugfix for not present ships in `clear-goals`

Baezon:
Absent ships or wings should simply be skipped instead of stopping.


JohnAFernandez (approved):
Kablam! Asteroth does it again!

JohnAFernandez:
Huh, unstated assumption in the code is that ship names should never be the exact same as a wing name.

PR #4500, opened 2022-07-18
Include more valid object types in this check

JohnAFernandez:
The previous version of the fix only included ships because it was believed to be the object that a ship used to be guarding.  Now it's clear that this is an object that the ship is currently targeting *after* guarding something.

So remove the UNREACHABLE and manually check the valid object types with an assert.  Then only continue with the function if it is the right target type.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -10015,12 +10015,26 @@ void guard_object_was_hit(object *guard_objp, object *hitter_objp)
 			aip->submode = SM_ATTACK;
 			aip->submode_start_time = Missiontime;
 			aip->active_goal = AI_ACTIVE_GOAL_DYNAMIC;
-		} else if (Objects[aip->target_objnum].type == OBJ_SHIP) {
-			int	num_attacking_cur, num_attacking_new;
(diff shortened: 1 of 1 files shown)


code/ai/aicode.cpp
```diff
@@ -10015,12 +10015,26 @@ void guard_object_was_hit(object *guard_objp, object *hitter_objp)
...
 			aip->active_goal = AI_ACTIVE_GOAL_DYNAMIC;
-		} else if (Objects[aip->target_objnum].type == OBJ_SHIP) {
-			int	num_attacking_cur, num_attacking_new;
+		} else {
+			// This section used to be a lot simpler, but some "invalid" object types were probably getting into num_ships_attacking
+			// below, so we have to manually return for some object types and make sure to assert for other object types
+			// that are symptomatic of a larger problem with the engine.
+
+			// This just checks to see that the target was a valid type. If it is not one of these, something really
+			// could be drasticly wrong in the engine.
```
Baezon, line 10024: "drastically" -Goober, probably

Baezon (approved):

PR #4501, opened 2022-07-19
Fix 4276

### LINKED ISSUE #4276: "Free Look View" conflicts with thrust controls
When "free look view" is used while "reverse thrust" is held down, the throttle acts as if "reverse thrust" is not held down.

For example, hold "reverse thrust" to stop for rearming. While rearming and holding "reverse thrust", use "free look view" to look around, but the throttle will act as if "reverse thrust" is not held down (may go back to auto speed).

z64555:
Move processing of thrust keys to be within the same block as the throttle keys

* Remove do_thrust_keys() function
* Clean up documentation of Viewer_mode flags
Fixes #4276

### DIFF
--- a/code/globalincs/systemvars.h
+++ b/code/globalincs/systemvars.h
@@ -31,23 +31,23 @@
 #define	GM_CAMPAIGN_MODE				(1 << 10)			// are we currently in a campaign.
 #define GM_LAB							(1 << 11)			// We are currently in the F3 lab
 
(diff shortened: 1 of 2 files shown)


Goober5000 (approved):

daftmugi:
@z64555 Thank you for the fix! It's great to be able to more easily look around during a rearm.

PR #4502, opened 2022-07-19
Preliminary Changelog for 22.2.0

Kiloku:
This is an *incomplete* changelog for 22.2.0, made by reading each commit up to RC5, reading their PRs, reading linked issues, and writing down my understanding of the effective changes. 

I added a new "feature" to the changelog itself: Collapsible subsections, by topic.

I skipped fixes for issues that were added during the 22.2.0 development cycle, as well as CI/build related commits (with the exception of the Mac related ones).  

This is in draft form, since I still need to write up changes from the post RC5 commits, and also because there are changes yet to be committed. 

Reminders for before merging:
- [ ] Add the link to the announcement thread in HLP once it exists.
- [x] Correct the date to the day the Release is actually posted.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -19,7 +19,7 @@ Headlining features will always be on top.
 > Consequently, containers allow for handling data in much more sophisticated ways than variables. 
 > Containers come in two types:
 > * List container - an ordered sequence of values, sort of a cross between Python lists and deques
-> * Map container - pairs of keys associated with data, like Lua tables or Python dictionaries
+> * Map container - pairs of keys associated with data, like Lua tables or Python dictionaries  
 > Further details in the [SEXP Containers announcement thread](https://www.hard-light.net/forums/index.php?topic=98209.0)
 </details>
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,29 +4,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [22.2.0] - 2022-07-XX ([Thread](https://www.hard-light.net/forums/index.php?topic=XX))
-
-```
-Changelog note: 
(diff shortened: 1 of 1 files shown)


Kiloku:
Since the github diff UI won't allow reading the pretty version of the changelog, I recommend checking it out here: https://github.com/Kiloku/fs2open.github.com/blob/markdown-testing/Changelog.md

Goober5000:
Outstanding work. 👍  I read through it and I have a couple comments:

"Add some checks before converting a message to Command" is in the FRED section.  This is more accurately categorized as a gameplay bugfix.  It patches the behavior of messages sent from built-in personas.

The following items were rolled back:
> Briefing Editor: Prevent the ship class from being changed if the icon image has a special closeup (like a jump node)
> Briefing Editor: If the icon image is changed to a special closeup, automatically choose a ship class that doesn't have a custom icon

The briefing editor no longer restricts or sets the ship class based on the ship type, or vice versa.

JohnAFernandez (changes requested):

Changelog.md
```diff
@@ -366,6 +383,10 @@ Headlining features will always be on top.
 - Fix a bunch of graphics functions to respect GR_STUB, preventing problems in standalone server
 - Clean up line endings in several files (replaced CRLF with LF basically everywhere)
 - Fixed a timestamp issue which helps lay the groundwork to add In-Game Joining to multiplayer
+- Prevent pilot and campaign files from having invalid values, by clamping them within the expected ranges when saving and loading. 
+  - Also warn player if such an invalid value was detected, telling them which setting had the wrong value and what it was reset to.
+- Try to avoid invalid types in `obj_team()`
```
JohnAFernandez: Now that I see this again, probably better to say something like, "Try to avoid assertions by preventing invalid types from being handled by `obj_team()`"

TRBlount (approved):

PR #4503, opened 2022-07-19
Ensure that SEXP shudown is before script shutdown

BMagnu:
We must do sexp_shutdown() before we clear the scripting state. Otherwise FSO will try to destroy the LuaFunction references still held by the SEXP system, which will cause a crash as the lua state (and thus its function registry) is a nullptr / not allocated anymore by that point.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -6683,15 +6683,14 @@ void game_shutdown(void)
 		gr_clear();
 		gr_flip();
 	}
-
+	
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
This looks good, but I would put it before `scpui::shutdown_scripting()`.  That call, the `Script_system.Clear()` call, and the `scpui::shutdown()` call look like they should be grouped together.

Goober5000 (approved):

PR #4506, opened 2022-07-21
only attenuate a persistent sound by engine strength if it's an engine sound

Goober5000:
Previously the persistent object sound management code would attenuate all script-assigned sounds by engine strength, and would not play them at all if the engine was off.  This fixes the code to only apply that logic if the sound is in fact and engine sound.

There is also a small amount of refactoring.

### DIFF
--- a/code/object/objectsnd.cpp
+++ b/code/object/objectsnd.cpp
@@ -461,10 +461,10 @@ void obj_snd_do_frame()
 	}
 
 	closest_dist = 1000000.0f;
-	closest_objp = NULL;
+	closest_objp = nullptr;
 
-	object *observer_obj = NULL;
-	if (Viewer_obj != NULL) {
+	object *observer_obj = nullptr;
+	if (Viewer_obj != nullptr) {
 		observer_obj = Viewer_obj;
 	} else if (Viewer_mode & VM_OTHER_SHIP && Player_ai->target_objnum != -1) {
 		// apparently Viewer_obj is still null when Viewing Other Ship externally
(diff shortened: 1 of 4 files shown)


code/object/objectsnd.cpp
```diff
@@ -612,7 +618,7 @@ void obj_snd_do_frame()
 						break;
 
 					default:
-						Int3();	// get Alan
+						UNREACHABLE("Unhandled object type for persistent sound; get Alan");
```
JohnAFernandez, line 621: You can leave the get Alan, of course, but make sure a person only loosely connected to us will know to report it to FSO.  Something along the lines of "get Alan -- I mean, an SCP coder!"  

Also, if we have some weird object type proliferating here, we should know what type it is.

JohnAFernandez (changes requested):

code/object/objectsnd.cpp
Goober5000: That kinda takes the fun out of it, but changed.

code/object/objectsnd.cpp
```diff
@@ -535,6 +536,8 @@ void obj_snd_do_frame()
 					speed_vol_multiplier = sip->min_engine_vol + ((1.0f - sip->min_engine_vol) * percent_max);
 				}
 			}
+
+			// check conditions for subsystem sounds
 			if (osp->ss != NULL)
```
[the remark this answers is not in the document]
Goober5000: Done.

JohnAFernandez (approved):

Issue #4504, opened 2022-07-20
RC7 crashes across the board on BPC.


MoerasGrizzly:
In several War in Heaven missions (that I've found thus far), the RC7 build hard crashes to desktop. These crashes do not occur on the RC6 build.

On Delanda Est (BP2-15) and Sunglare (BP2-16) and the RC7 fast debug build, the following errors are thrown before the hard crash:
```
Warning: sexp-script-eval failed to evaluate string "proBoxValue()"; check your syntax
File: sexp.cpp
Line: 24031
```

In Pawns of a Board of Bone (BP2-14), the following error is thrown:
```
Assert: "aip->target_objnum == OBJ_SHIP || aip->target_objnum == OBJ_WEAPON || aip->target_objnum == OBJ_DEBRIS || aip->target_objnum == OBJ_ASTEROID || aip->target_objnum == OBJ_WAYPOINT"
aicode.cpp
Line: 10027
This function just discovered that Gemini 3 has an invalid target object type of 1. This is bad. Please report!

ntdll.dll! ZwWaitForSingleObject + 20 bytes
KERNELBASE.dll! WaitForSingleObjectEx + 142 bytes
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
fs2_open_22_2_0_RC7_x64_AVX-FASTDBG.exe! <no symbol>
[...]

PR #4507, opened 2022-07-22
prevent a crash on game exit

Goober5000:
When `mission_campaign_clear()` is called, all persistent SEXP nodes (those used by the campaign) are marked as non-persistent.  When `sexp_shutdown()` was moved in #4503, this caused `Sexp_nodes` to be nullptr when the campaign was cleared, which caused a crash.  This adds checks to prevent the crash.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -6684,19 +6684,6 @@ void game_shutdown(void)
 		gr_flip();
 	}
 	
-	// Free SEXP resources
-	sexp_shutdown();
-	
-	// Free the scripting resources of the new UI first
-	scpui::shutdown_scripting();
-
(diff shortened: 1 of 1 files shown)


Goober5000:
That's a good question.  It looks like after `player_tips_close()`, the more fundamental stuff starts getting shut down, beginning with `control_config_common_close()`.  Since both SEXPs and scripting are pretty fundamental, I could move them to the beginning of that section.

z64555:
Might this affect #4504?

Goober5000:
> Might this affect #4504?

No, I don't expect it would be related.

z64555 (approved):
Hmmm... I don't see anything that may be an issue with just moving the sexp and scripting calls down.

PR #4510, opened 2022-07-24
prevent too many dynamic SEXPs from breaking the SEXP system

Goober5000:
SEXP categories will not support more than 256 SEXPs; this is why the CHANGE2 category had to be created when CHANGE got full.  Unfortunately, dynamic SEXPs will count toward this limit, so mysterious errors will occur for the ones whose category indexes roll over.  The problem is not easily fixable, but at least the engine can detect the situation, warn about it, and refrain from adding any problematic SEXPs.

Since too many *total* SEXPs would cause a similar problem, this adds a warning for that case too.

### DIFF
--- a/code/parse/sexp/sexp_lookup.cpp
+++ b/code/parse/sexp/sexp_lookup.cpp
@@ -125,8 +125,10 @@ void parse_sexp_table(const char* filename) {
 				luaSexp->parseTable();
 
 				int op = add_dynamic_sexp(std::move(instance), SEXP_GOAL_OPERATOR);
-				luaSexp->registerAIMode(op);
-				luaSexp->maybeRegisterPlayerOrder(op);
+				if (op >= 0) {
+					luaSexp->registerAIMode(op);
+					luaSexp->maybeRegisterPlayerOrder(op);
+				}
 			}
 			required_string("#End");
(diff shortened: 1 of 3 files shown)


code/parse/sexp/sexp_lookup.cpp
```diff
@@ -177,8 +179,19 @@ int add_dynamic_sexp(std::unique_ptr<DynamicSEXP>&& sexp, int type)
 	new_op.min = sexp->getMinimumArguments();
 	new_op.max = sexp->getMaximumArguments();
 
+	auto free_op_index = get_next_free_operator(sexp->getCategory());
```
z64555, line 182: I'm not entirely comfortable with using an `auto` since its later used in a bitwise OR operation.  Should it resolve to a non-integer type for any reason it would invalidate the `new_op.value` later on.

LGTM otherwise.

z64555 (changes requested):

code/parse/sexp/sexp_lookup.cpp
Goober5000: Good point. Changed.

z64555 (approved):
LGTM

Issue #4509, opened 2022-07-23
Various multiplayer issues


Goober5000:
I ran into these today while playing the first three missions of ST:R co-op.

- [ ] Changing the control presets caused the game to end for everyone, although I didn't see a hard crash.  I had been using the default binding, but when I launched the first mission I realized I needed to bind my joystick rudder to a different axis.  As soon as I did this, the mission ended.
- [ ] It seemed really difficult to inflict damage to a player in a dogfight, compared to the AI in a co-op mission.
- [ ] More than once, with more than one player, the `MESSAGE_STRAY_WARNING` and `MESSAGE_STRAY_WARNING_FINAL` played in quick succession, even though we were in the normal boundaries of the mission.
- [ ] Despite `MESSAGE_STRAY_WARNING_FINAL` playing, which is supposed to be followed after 5 seconds by blowing up the player's ship, no such execution occurred.

z64555:
Assigning to low priority since its multiplayer.  The first item has a workaround (albeit a lame one) of refraining to change presets during a match.

Items 2 and 3 may be a bogus warning that may have a different check than the kill condition.

Item 4 is, well, multi being multi. :shrug:

JohnAFernandez:
For 2 and 3, I just saw a possible source of the bug. Here's the multiplayer check for the message.
`if (MULTI_CONNECTED(Net_players[idx]) && !MULTI_STANDALONE(Net_players[idx]) && !MULTI_OBSERVER(Net_players[idx]) && (Objects[Net_players[idx].m_player->objnum].type != OBJ_GHOST) ) {`

It should check that the object *is* a ship, not that the object is not a ghost.  Because ghost objects are deleted in multi so that we can do respawn cleanly.

That would explain why no one is dying, because the ship would already be dead.


4 is going to be based on ping, currently.  I have to think of a long term solution for that, because there really is no way to handle high-ping, client-client weapons fire without either a direct connection between those two computers or trusting the client's hit detection.  We may have to bite the bullet and just do the client hit detection....

notimaginative:
For item 1 did you change the name of the preset in the popup?

I tried to reproduce the problem thinking it was a timeout issue, but accidentally discovered that game controls are processed while a popup is active. So any inputs (key, mouse, joy) that you pressed while that save preset popup was active would have been processed by the game. Depending on what was typed you may have ended the mission without realizing it.

[...]

PR #4511, opened 2022-07-25
Only check that *ships* have left the mission

JohnAFernandez:
Other types are possible here because multi deletes the ghost object, at least, that is what I understood from the code and comments two years ago, when I first started looking at respawn bugs.  So it is much better to just ensure that the object is a ship, instead of checking against what it shouldn't be before checking the distance.

Related, but not guaranteed to fix part of #4509.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -9115,7 +9115,7 @@ static void ship_check_player_distance()
 		if (MULTIPLAYER_MASTER) {
 			// warn all players
 			for (idx=0; idx<MAX_PLAYERS; idx++) {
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
This seems like it would have no effect.  What object types could a player be, other than a ship, a ghost, or an observer (which is also checked)?

And before you say that more object types exist -- indeed they do, but the mission was playing normally aside from the bug.  If the player happened to be a "weapon", more things would have gone wrong.

JohnAFernandez:
When a ghost is deleted, which does happen in multi, the object type is changed to `OBJ_NONE` though, which is allowed by the former version of the check.  The delete object function only returns at the deletion of a ghost when in single player.  So it really should be `== OBJ_SHIP`.

Goober5000 (approved):
Ah, ok. 👍

PR #4528, opened 2022-08-07
Add default shockwave damage types functionality.

MageKing17:
Shockwaves can have a separate damage type from the weapon that created them, but one common footgun related to this is that they *only* have a damage type if you specify one, rather than inheriting their parent weapon's damage type. I saw it suggested more than once that some sort of automation here would be welcomed, so I tried to make things easier for modders. This includes setting a global default, inheriting from the parent weapon, and/or generating a damage type based on the parent weapon's damage type but with an added suffix. All of these options are also available for dinky shockwaves, with the non-dinky options applying to both by default.

I tried to make applying all the new options simultaneously have meaningful behavior, as well as making the feature as modular-table-friendly as possible by only finalizing shockwave damage types after parsing is over. That being said, if anyone has any ideas to make this functionality even more modder-friendly (IIRC suffixes were @EatThePath's idea), go ahead and comment; if it's not too complex, I should be able to throw it in, and even if it is someone may figure out how to implement it without requiring a full-blown overhaul of the parsing system... and if it *does* require a full-blown overhaul of the parsing system, it may happen anyway, so you never know.


Baezon (approved):
Looks good, although I feel I shoulds raise the question of whether `Shockwaves_inherit_parent_damage_type` should be default-on for mods targeting newer verisons. The fact that they don't is fairly unintuitive...

MageKing17:
I'm generally in favour of defaulting intuitive behavior when targeting the latest version, but I didn't want to do it unilaterally.

Goober5000:
I'd be in favor of it.

EatThePath:
@MageKing17 Do you want to make the change to defaults in this PR or should we leave that for later and merge this one?

MageKing17:
> Do you want to make the change to defaults in this PR or should we leave that for later and merge this one?

I have no objection to doing it in this PR, since nobody seems to mind this being default-on for mods targeting new versions of FSO. Based on the example of d8987b3de08e6db35dfd65d1a3b2c2e4534150d6, it should be based on whether the mod supports 22.4, as the next stable release; any objections?

wookieejedi:
Sounds good to me.

EatThePath:
Sounds reasonable to me.

Goober5000:
Sounds good, and the new commit looks good too

PR #4531, opened 2022-08-11
Allow Mods to specific default multi update speeds

### LINKED ISSUE #4497: Change default Multiplayer Object Update rate to "LAN" when creating a new pilot
Currently, the default is "High", but since we always recommend to all players that they switch to "LAN" before they play multiplayer, it seems reasonable to changed the default. 

There is precedent for changing default settings, for example, new pilots start out with the Details settings at maximum.

wookieejedi:
Previosuly, the default object update speed setting for new players was `High`. Many mods, such as Fate of the Galaxy, do multiplayer tests where we have to ensure players are using the highest object update speed, which is LAN. This extra step is cumbersome and prone to players' forgetfulness.

This PR allows mods to set the default multiplayer object update speed for new pilots to mitigate this issue. Tested and works as expected.

EDIT: Closes #4497.

### DIFF
--- a/code/network/multi_options.cpp
+++ b/code/network/multi_options.cpp
@@ -349,7 +349,7 @@ void multi_options_set_local_defaults(multi_local_options *options)
 	if ( Psnet_connection == NETWORK_CONNECTION_DIALUP ) {
 		options->obj_update_level = OBJ_UPDATE_LOW;
 	} else {
-		options->obj_update_level = OBJ_UPDATE_HIGH;
+		options->obj_update_level = Default_multi_object_update_level;
 	}
 }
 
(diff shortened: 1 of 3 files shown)


Kiloku:
Related: #4497 

IMO, this is even better than making it the default globally

code/network/multi_options.cpp
```diff
@@ -349,7 +349,7 @@ void multi_options_set_local_defaults(multi_local_options *options)
 	if ( Psnet_connection == NETWORK_CONNECTION_DIALUP ) {
 		options->obj_update_level = OBJ_UPDATE_LOW;
```
JohnAFernandez, line 350: As much as it seems silly, it's good that we are keeping the top part of this if statement the way it is, in case someone really does have a slow connection.

JohnAFernandez (approved):

code/network/multi_options.cpp
Kiloku: As rare as it might be, there definitely still are people on slow connections. Even if not exactly dial-up, I know of places where entire households will use a single phone's 2G or 3G internet connection routed to other devices via hotspot.

PR #4532, opened 2022-08-12
Automatically calculate length of flag name lists.

MageKing17:
`Ai_flag_names[]`, `Object_flag_names[]`, and `Ship_flag_names[]` all had an associated `#define` containing their length that had to manually be incremented whenever new flag names were added to the list. This commit instead makes the number of flags in the list be computed by simple `sizeof` division, so now adding a new entry just requires adding the entry, rather than remembering to bump a `#define`. Tested in FRED to make sure populating `OPF_SHIP_FLAG` entries still works.


Goober5000 (approved):
Good change

PR #4443, opened 2022-06-26
more cleanup in preparation for model translation

Goober5000:
1) Clean up stepped rotation, fixing some logic, adding safeguards for invalid values, and allowing rotations to go in the opposite direction
2) Add more specific SEXP error checking for AWACS and rotating subsystems
3) Assorted refactoring and polishing

PR #4535, opened 2022-08-13
Fix False Null MOI warnings

### LINKED ISSUE #4534: False Null MOI warnings
The first 22.3 nightly gives these warnings where 22.2.0 does not. Does not apply to all models. I don't see a pattern to what models trigger it. Triton in FSU is a perfect example. Also happens with the Fenris. I have verified the POFs in question absolutely do have MOI data.

Load up 22.2.0 debug with FSU 4.6.6. Go to the lab and view the Triton. No warning.
Load up 22.3 debug with FSU 4.6.6. Go to the lab and view the Triton. Null MOI warning is triggered.

wookieejedi:
In #4443, `fl_near_zero` was set to use default epsilon values in the `IS_VEC_NULL` check. In this comparison, epsilon is the wrong threshold because it used to have a much smaller threshold. The fix restores the previous definitions of `IS_VEC_NULL` and `IS_VEC_NULL_SQ_SAFE` by exposing the epsilon value as an optional parameter (similar to how `IS_NEAR_ZERO` used to work. This PR also restores `IS_VEC_NULL_SQ_SAFE` where it was replaced by `IS_VEC_NULL.`

Fixes #4534.

### DIFF
--- a/code/math/floating.h
+++ b/code/math/floating.h
@@ -96,7 +96,7 @@ float fl_isqrt( float x )
 } 
 */
 
-// sees if a floating point number is within the minimum tolerance of zero
+// sees if a floating point number is within a certain threshold (by default, epsilon) of zero
 inline bool fl_near_zero(float a, float e = std::numeric_limits<float>::epsilon())
 {
 	return a < e && a > -e;
--- a/code/math/floating.h
+++ b/code/math/floating.h
@@ -97,10 +97,9 @@ float fl_isqrt( float x )
 */
 
 // sees if a floating point number is within the minimum tolerance of zero
-inline bool fl_near_zero(float a)
+inline bool fl_near_zero(float a, float e = std::numeric_limits<float>::epsilon())
 {
-	return a < std::numeric_limits<float>::epsilon()
-		&& a > -std::numeric_limits<float>::epsilon();
+	return a < e && a > -e;
 }
 
 // sees if two floating point numbers are within the minimum tolerance
--- a/code/math/vecmat.h
+++ b/code/math/vecmat.h
@@ -20,14 +20,15 @@
 #define vm_is_vec_nan(v) (fl_is_nan((v)->xyz.x) || fl_is_nan((v)->xyz.y) || fl_is_nan((v)->xyz.z))
 
 //Macros/functions to fill in fields of structures
+//VEC_NULL macros split into two functions in 2009 with commit 75a514b
 
-//macro to check if vector is zero, uses 1e-16 as a minimum tolerance
+//macro to check if vector is close to zero or would be close to zero after squaring
 #define IS_VEC_NULL_SQ_SAFE(v) \
 		(fl_near_zero((v)->xyz.x, (float) 1e-16) && \
 		fl_near_zero((v)->xyz.y, (float) 1e-16) && \
 		fl_near_zero((v)->xyz.z, (float) 1e-16))
 
-// macro to check if vector is zero, uses 1e-36 as a minimum tolerance
+//macro to check if vector is close to zero
 #define IS_VEC_NULL(v) \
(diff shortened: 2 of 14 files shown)


code/math/vecmat.h
```diff
@@ -22,9 +22,15 @@
 //Macros/functions to fill in fields of structures
 
 //macro to check if vector is zero
```
JohnAFernandez, line 24: Can you change this comment to match what `IS_VEC_NULL_SQ_SAFE` is specifically for, and then probably move this comment to be above `IS_VEC_NULL`?

JohnAFernandez (changes requested):

code/math/vecmat.h
wookieejedi: Sure, the original code had it this way but happy to take the opportunity to update comments.

Goober5000: The original macro split happened back in 2009: 75a514b5f9

I would be curious if anyone has a good idea how these thresholds should be determined.  The well-known `epsilon` is for addition and subtraction; I would assume there is some other standard for multiplication and squaring.

JohnAFernandez: My theory is that the `SQ_SAFE` version ensures that squaring a float is safe. But I don't have evidence for that beyond common sense.

Baezon: I think the existing epsilon is fine. This issue is MOI-specific, imo. Due to the way they are calculated and used, they are unfortunately extremely small numbers a lot of the time.

wookieejedi: Makes sense. Are folks good with the current version then?

JohnAFernandez (approved):

code/math/vecmat.h
JohnAFernandez: Yeah, go ahead

code/math/floating.h
```diff
@@ -97,10 +97,9 @@ float fl_isqrt( float x )
 */
 
 // sees if a floating point number is within the minimum tolerance of zero
-inline bool fl_near_zero(float a)
+inline bool fl_near_zero(float a, float e = std::numeric_limits<float>::epsilon())
```
Goober5000, line 100: Since this is adding a new parameter, change the comment to be this:
```
// sees if a floating point number is within a certain threshold (by default, epsilon) of zero
```

code/math/vecmat.h
```diff
@@ -21,10 +21,17 @@
 
 //Macros/functions to fill in fields of structures
 
-//macro to check if vector is zero
-#define IS_VEC_NULL(v) (fl_near_zero((v)->xyz.x) && \
-						fl_near_zero((v)->xyz.y) && \
-						fl_near_zero((v)->xyz.z))
+//macro to check if vector is zero, uses 1e-16 as a minimum tolerance
```
Goober5000, line 24: Change the comment to something like this:
```
// macro to check if vector is close to zero or would be close to zero after squaring
```
And change the next comment to something like this:
```
// macro to check if vector is close to zero
```

Goober5000 (changes requested):

code/math/vecmat.h
Goober5000: Also add the hash of the commit that originally split the macros into two.

code/math/floating.h
wookieejedi: Good idea, just added.

code/math/vecmat.h
wookieejedi: Thanks, just updated.

Goober5000 (approved):

Goober5000 (changes requested):
Actually hang on while I research some floating point things

Goober5000 (approved):

Goober5000:
Approving this for the time being.  Other floating-point changes can go in a follow-up PR.

PR #4536, opened 2022-08-13
Add options to selectively disable shadows

BMagnu:
Some mods can contain models for which having shadows in the techroom / mission select screen is a problem. This allows to bulk-disable rendering of shadows for some of these categories, such as shadows for weapon selection during the mission briefing.


Goober5000 (commented):
Considering how there is very similar logic repeated in several places, is it possible to encapsulate it somehow?

Goober5000 (commented):
This looks much better. 👍  Though I would recommend a high-level comment or two in `shadow_maybe_start_frame` to explain what it's doing.

Goober5000 (approved):
👍

PR #4537, opened 2022-08-14
Prevent subsystem `fireWeapon` scripting function from triggering an assert

Kiloku:
It is possible for a script to call the `fireWeapon` subsystem function passing `wnum` as any integer. 
Down the line, `wnum` ends up passed to `ai_turret::get_turret_weapon_wip`, and could trigger either of those asserts:  
```C++
Assert(weapon_num < MAX_SHIP_WEAPONS);
Assert(weapon_num >= 0);
```

This PR prevents that by giving an error and returning early.


JohnAFernandez (approved):

Goober5000 (changes requested):
Could you make this say `TurretWeaponIndex` to be consistent with the function documentation?

Goober5000 (approved):

Issue #3712, opened 2021-10-09
FR: Add improved contrast options for HUD


Kiloku:
Some HUD elements become hard to see against certain the background of certain missions, or against certain objects, even with the "High Contrast" mode.   
The simplest solution would be to add a stroke or a "drop shadow" in contrasting color to the HUD elements.  
This would probably need to be a command line option. 
The color of the stroke or shadow could either be set in the command line as well, or default to [the most contrasting color to the HUD element's color](https://gamedev.stackexchange.com/questions/38536/given-a-rgb-color-x-how-to-find-the-most-contrasting-color-y). 

Doing it as drop shadow would probably not be difficult from what I can tell of HUD code.
Doing it as a stroke would probably require applying a shader to each HUD element.

z64555:
Drop shadow is probably easiest to do.  Just take the HUD element, darken it, and offset it a few pixels towards a corner, then draw the HUD element as normal on top of it.

Now the problem is... Should this be an engine feature or should it be baked into HUD assets by the modder?  If an engine feature, the commandline option is currently the fastest way to get such a feature in, but if you want it so that the players can adjust it you'll have to use the new SCP menu UI.

Kiloku:
Aren't HUD assets limited in their color palette? Could an edited HUD texture have a functional drop shadow?

MjnMixael:
The HUD currently allows both increased contrast and drop shadow enabled with a keybind. Is this asking for more than that?

Kiloku:
The FR mentions strokes and dynamically changing the colors of the shadows/strokes if needed. My favorite solution would be a stroke around the text and UI elements, but I don't know how feasible this is (I guess we'd have to make a shader that dynamically generates distance fields?). Regardless, I believe that the drop-shadows are good enough for most of the use-cases.

Kiloku:
Ah, the PR for the drops shadows also mentions something else that I think would be very good, but maybe could be a separate Issue:

> Allow HUD elements to be fully opaque (currently even with alpha at 255 and High Contrast mode, they'll be slightly transparent due to how `aabitmaps` work

MjnMixael:
That all seems like overkill to me. I feel like the current options handle the needs but I suppose it won't hurt anything if a developer someday wants to implement all that.

PR #4543, opened 2022-08-15
Add optional Drop Shadows to HUD

Kiloku:
This PR adds a new control (By default Alt+L, but it will be unset for existing pilots) which lets the player enable or disable Drop Shadows on the HUD, affecting Text, the main targeting reticle and the background of the ETS gauges. 

<details><summary><b>Click here to see comparison screenshots</b></summary>

Low contrast, no shadows
locontrast

High contrast, no shadows
hicontrast

Low contrast, with shadows
locontrast_shadowed

High contrast, with shadows
hicontrast_shadowed
</details>

Related: #3712. I'm not sure if we could consider that it closes that FR. IMO there are other possible ways to improve HUD contrast that would be nice to have: 
 * Allow HUD elements to be fully opaque (currently even with alpha at 255, they'll be slightly transparent due to how `aabitmap`s work
 * Use outlines instead of shadows (This would probably require a specific shader) 
 * Allow HUD elements to dynamically change color depending on the colors of what's behind it

### DIFF
--- a/code/controlconfig/controlsconfig.h
+++ b/code/controlconfig/controlsconfig.h
@@ -284,17 +284,17 @@ enum IoActionId : int {
 	CUSTOM_CONTROL_4								= 122,
 	CUSTOM_CONTROL_5								= 123,
 
-	TOGGLE_HUD_SHADOWS								= 124,
 	//!< @n
 	//!< Analog controls
 	//!<-----------------------------
 	//!< All axis actions must be kept together
-	JOY_HEADING_AXIS								= 125,
-	JOY_PITCH_AXIS									= 126,
-	JOY_BANK_AXIS									= 127,
-	JOY_ABS_THROTTLE_AXIS							= 128,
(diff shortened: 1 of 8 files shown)


z64555 (approved):
Changes appear to be straightforward.  I believe you said on Discord that you've tested this to the best of your ability to ensure that it works as intended?  Placing the callback within button_function() near the Toggle HUD Contrast control seems like the best fit for this.

The IoActionID's are currently locked in to their indeces to provide backward compatibility with pre-Controls6 controls config that's held in the pilot file.  All new controls Post-Controls6 do not need to be numbered so they can just be stuffed in just before the CCFG_MAX enum until I can come up with a good way of keeping forward compatibility of pilot file controls.

dariusbei:
This is a great quality of life addition. How easy would it be to make this active by default using a flag in e.g. game_settings.tbl?

Goober5000:
Pretty easy, it looks like.

PR #4548, opened 2022-08-16
Add scripting API and conditions for executing on server only

BMagnu:


code/scripting/api/libs/multi.cpp
```diff
@@ -0,0 +1,15 @@
+#include "multi.h"
+
+#include "network/multi.h"
+
+namespace scripting {
+	namespace api {
+		//**********LIBRARY: Multi
+		ADE_LIB(l_Multi, "Multi", "multi", "Functions for scripting for and in multiplayer environments.");
+
+		ADE_FUNC(isServer, l_Multi, nullptr, "Prints a string", "boolean", "true if the script is running on the server, false if it runs on a client or in singleplayer.")
```
Goober5000, line 10: nitpick: "it runs" should be "it is running" for consistency

code/scripting/scripting.cpp
```diff
@@ -635,6 +649,12 @@ bool ConditionedHook::ConditionsValid(int action, object *objp1, object *objp2,
 					return false;
 				}
 			}
+			
+			case CHC_MULTI_SERVER:
+			{
+				//condition_cached_value is 0 if we execute on clients, 1 on servers
+				return static_cast<bool>(scp.condition_cached_value == 0) != static_cast<bool>(MULTIPLAYER_MASTER);
```
Goober5000, line 656: For clarity I would prefer this:
```
return static_cast<bool>(scp.condition_cached_value == 1) == static_cast<bool>(MULTIPLAYER_MASTER);
```

Goober5000 (changes requested):
Couple suggestions

Goober5000 (approved):

Issue #4546, opened 2022-08-16
FR: Lua access to object scan data


MjnMixael:
Would like to expose Scan Time and Scan Min Distance (if that's a thing, iirc cargo is 150m and bigger ships are a little further?) as part of object data available in Lua, please.

MjnMixael:
Scan Distance looks to be quite a bit more complex with built-in values. I can go without that bit.

PR #4551, opened 2022-08-17
Lua scan time

MjnMixael:
Give Lua access to a ship's scan time fields. Per FR #4546

### DIFF
--- a/code/scripting/api/objs/shipclass.cpp
+++ b/code/scripting/api/objs/shipclass.cpp
@@ -273,6 +273,9 @@ ADE_VIRTVAR(AfterburnerFuelMax, l_Shipclass, "number", "Afterburner fuel capacit
 ADE_VIRTVAR(ScanTime, l_Shipclass, nullptr, "Ship scan time", "number", "Time required to scan, or 0 if handle is invalid. This propery is read-only")
 {
 	int idx;
+	if (!ade_get_args(L, "o", l_Shipclass.Get(&idx)))
+		return ade_set_error(L, "i", 0);
+
 	if (idx < 0 || idx >= ship_info_size())
 		return ade_set_error(L, "i", 0);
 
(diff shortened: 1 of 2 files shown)


code/scripting/api/objs/shipclass.cpp
```diff
@@ -270,6 +270,19 @@ ADE_VIRTVAR(AfterburnerFuelMax, l_Shipclass, "number", "Afterburner fuel capacit
 	return ade_set_args(L, "f", Ship_info[idx].afterburner_fuel_capacity);
 }
 
+ADE_VIRTVAR(ScanTime, l_Shipclass, "number", "Ship scan time", "number", "Time required to scan, or 0 if handle is invalid. This propery is read-only")
```
BMagnu, line 273: Read-only properties should have a nullptr instead of the first type-string (as the first type string indicates the data type the variable can be set to in the docs).
So something like ``ADE_VIRTVAR(ScanTime, l_Shipclass, nullptr, "Ship scan time", "number", "Time required to scan, or 0 if handle is invalid. This propery is read-only")``

BMagnu (changes requested):
I would remove the API function you added to the ship object (``shi.ScanTime``). Since you pull data from the ship class there, it's functionally equivalent to accessing ``ship.Class.ScanTime``, and thus it also makes more sense to be only available through the ship class object and not through the ship object.

code/scripting/api/objs/shipclass.cpp
MjnMixael: Done

BMagnu (approved):

Goober5000 (changes requested):
The virtvar ought to have a check for `ADE_SETTING_VAR` in case a script does try setting the value.  This should invoke `LuaError`.  See `ScanningTimeMultiplier` for an example.

And since the virtvar is readonly, there is no need to have a `time` variable or to fetch it with `ade_get_args`.

BMagnu:
Good catch.

MjnMixael:
Feedback addressed

Goober5000 (changes requested):
It's almost good now, but the `ade_get_args` is still necessary still to get the ship class.  So this slightly modified bit of code needs to be added back:
```
if (!ade_get_args(L, "o", l_Shipclass.Get(&idx)))
	return ade_set_error(L, "i", 0);
```

Goober5000 (approved):
Looks good now.

PR #4552, opened 2022-08-18
enhanced control for the global ship shudder

Goober5000:
Allow the variables controlling ship shudder to be independently set using scripting.  Also add new flags to apply the shudder without decay and apply the shudder regardless of camera.

Also change the shudder timestamp to the new style.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -541,7 +541,9 @@ float Sun_spot = 0.0f;
 big_expl_flash Big_expl_flash = {0.0f, 0.0f, 0};
 
 // game shudder stuff (in ms)
-int Game_shudder_time = -1;
+bool Game_shudder_perpetual = false;
+bool Game_shudder_everywhere = false;
+TIMESTAMP Game_shudder_time = TIMESTAMP::invalid();
 int Game_shudder_total = 0;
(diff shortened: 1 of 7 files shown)


freespace2/freespace.cpp
```diff
@@ -2993,10 +2989,29 @@ void apply_view_shake(matrix *eye_orient)
...
+				Game_shudder_time = TIMESTAMP::invalid();
+			} else {
+				tangles.p += get_shake(Game_shudder_intensity * 0.005f, timestamp_until(Game_shudder_time), Game_shudder_total);
+				tangles.h += get_shake(Game_shudder_intensity * 0.005f, timestamp_until(Game_shudder_time), Game_shudder_total);
+			}
+		}
 	}
 
 	// maybe bail
-	if (tangles.p == 0.0f && tangles.h == 0.0f && tangles.b == 0.0f)
```
wookieejedi, line 2999: I take it changing to the now default fl_near_zero threshold does not visually change things?

Goober5000: Not that I can see.

wookieejedi (approved):
Given tests that visuals of shake are unchanged, the code looks good.

PR #4554, opened 2022-08-18
Add hook just before loadout parsing

### LINKED ISSUE #4361: FR: Add scripting function to recalculate loadout counts
There should be a way for FSO to be forced to recalculate loadout values, so that if a script updates sexp variables, it can be reflected in the loadout screen.  Currently, the variables cannot be updated at the right time to change loadout values.

Alternatively, we could allow direct control over the values.  It all depends on the effort the developer wants to put in.

MjnMixael:
Add On Loadout Parse hook that runs during mission parse but before loadouts are created. Resolves #4361

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -421,6 +421,9 @@ const std::shared_ptr<scripting::Hook> OnDepartureStartedHook = scripting::Hook:
 		{"Ship", "ship", "The ship that has begun the departure process."},
 	});
 
+ const std::shared_ptr<scripting::Hook> OnLoadoutParseHook = scripting::Hook::Factory("On Loadout Parse",
+	"Called during mission load just before parsing the team loadout.",{});
+
 // Goober5000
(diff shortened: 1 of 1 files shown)


code/mission/missionparse.cpp
```diff
@@ -421,6 +421,9 @@ const std::shared_ptr<scripting::Hook> OnDepartureStartedHook = scripting::Hook:
 		{"Ship", "ship", "The ship that has begun the departure process."},
 	});
 
+ const std::shared_ptr<scripting::Hook> OnLoadoutParseHook = scripting::Hook::Factory("On Loadout Parse",
```
Goober5000, line 424: After consultation with @BMagnu, I recommend calling this "On Loadout About to Parse" to make it clear that parsing has not actually happened yet.  Most "On X" hooks are fired after X has happened.

Goober5000 (changes requested):

code/mission/missionparse.cpp
JohnAFernandez: "On Loadout Before Parse"?

Goober5000 (approved):
Looks good now.

Issue #3070, opened 2021-01-02
Scripting FR: Expose Persona


naomimyselfandi:
The scripting API doesn't provide easy access to a ship's persona. Having access to it would provide two benefits: reading a fighter's persona makes it easier to send scripted messages, and being able to write to it allows ships to be given non-wingmate personae (eg. support ships).

Goober5000:
Yeah this would be a good things to have.  Should be simple to implement too.

PR #4555, opened 2022-08-18
Implement ability to get and set a ship's persona index

MjnMixael:
Implements #3070 . Allows getting and setting the ship's persona by index. Tested reading and writing with messages after writing and works as expected. Also testing writing an index that does not exist in messages.tbl. FSO does the same thing with that as it does if it gets an invalid persona from the mission file; it chooses a random persona instead.

### DIFF
--- a/code/scripting/api/objs/ship.cpp
+++ b/code/scripting/api/objs/ship.cpp
@@ -754,7 +754,7 @@ ADE_VIRTVAR(PersonaIndex, l_Ship, "number", "Persona index", "number", "The inde
 
 	ship *shipp = &Ships[objh->objp->instance];
 
-	if(ADE_SETTING_VAR && p_index >= 0)
+	if(ADE_SETTING_VAR && p_index > 0)
 		shipp->persona_index = p_index - 1;
 
 	return ade_set_args(L, "i", shipp->persona_index + 1);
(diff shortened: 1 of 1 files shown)


Goober5000 (changes requested):
This looks good, but two comments:
1) I'd suggest calling it `PersonaIndex`.  If we eventually implement support for Persona as a script-supported type, then we would want to retrieve that with the `Persona` virtvar.
2) Since Lua arrays are 1-based, not 0-based, there should be a -1 on line 758 and a +1 on line 760.  And line 745 should then say "0 if no persona is set".

code/scripting/api/objs/ship.cpp
```diff
@@ -742,6 +742,24 @@ ADE_VIRTVAR(Team, l_Ship, "team", "Ship's team", "team", "Ship team, or invalid
...
+	int p_index = -1;
+	if(!ade_get_args(L, "o|i", l_Ship.GetPtr(&objh), &p_index))
+		return ade_set_error(L, "i", 0);
+
+	if(!objh->IsValid())
+		return ade_set_error(L, "i", 0);
+
+	ship *shipp = &Ships[objh->objp->instance];
+
+	if(ADE_SETTING_VAR && p_index >= 0)
```
Goober5000, line 757: One other comment.  The `p_index >= 0` should now be `p_index > 0` to accommodate the 1-based index.

Goober5000 (approved):

PR #4556, opened 2022-08-18
Fix: Missing includes breaking Linux builds. Including the same thing twice in a row. A completely unnecessary recursive include loop.

PhantomHoover:


Goober5000 (approved):

JohnAFernandez:
Wow.

PR #4560, opened 2022-08-19
floating-point followup

Goober5000:
Floating-point comparisons are tricky beasts.  See this article for a good summary of the issues involved:
https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

This PR improves support somewhat, although there are obviously situations in the code (such as comparisons to 0) that should still be addressed.  This PR makes the following improvements:
1) Tightens the threshold on `IS_VEC_NULL` and `IS_VEC_NULL_SQ_SAFE` from 1e-16 and 1e-36 to FLT_EPSILON
2) Tightens the threshold for parsing `turn_time` in POFs
3) Restores the previous implementation of `fl_equal`
4) Adds a dedicated `IS_MOI_VEC_NULL` for moment of inertia values
5) Improves documentation

Follow-up to #4535 and #4443.

### DIFF
--- a/code/math/floating.h
+++ b/code/math/floating.h
@@ -98,10 +98,13 @@ inline bool fl_near_zero(float a, float e = std::numeric_limits<float>::epsilon(
 	return a < e && a > -e;
 }
 
-// sees if two floating point numbers are within the minimum tolerance
+// sees if two floating point numbers are approximately equal, taking into account the argument magnitudes
+// see commit c62037
+// and see also this article, because fl_equal may need to be rewritten if it is recruited into more demanding situations:
+// https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/
 inline bool fl_equal(float a, float b)
 {
-	return fl_near_zero(a - b);
+	return fl_abs(a - b) <= FLT_EPSILON * MAX(1.0f, MAX(fl_abs(a), fl_abs(b)));
 }
 
 // rounds off a floating point number to a multiple of some number
--- a/code/math/floating.h
+++ b/code/math/floating.h
@@ -104,7 +104,7 @@ inline bool fl_near_zero(float a, float e = std::numeric_limits<float>::epsilon(
 // https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/
 inline bool fl_equal(float a, float b)
 {
-	return fl_abs(a - b) <= FLT_EPSILON * MAX(1.0f, MAX(fl_abs(a), fl_abs(b)));
+	return fl_abs(a - b) <= FLT_EPSILON * std::max({ 1.0f, fl_abs(a), fl_abs(b) });
 }
 
 // rounds off a floating point number to a multiple of some number
--- a/code/math/vecmat.h
+++ b/code/math/vecmat.h
@@ -41,11 +41,11 @@
 #define IS_VEC_NULL(v) IS_VEC_NULL_SQ_SAFE(v)
 
 // macro to check if moment-of-inertia vector is close to zero
-// (uses different thresholds since MOI values are really small)
+// (uses the previous 1e-36 threshold since MOI values are really small)
 #define IS_MOI_VEC_NULL(v) \
-		(fl_near_zero((v)->xyz.x, (float) 1e-20) && \
-		fl_near_zero((v)->xyz.y, (float) 1e-20) && \
-		fl_near_zero((v)->xyz.z, (float) 1e-20))
+		(fl_near_zero((v)->xyz.x, (float) 1e-36) && \
+		fl_near_zero((v)->xyz.y, (float) 1e-36) && \
+		fl_near_zero((v)->xyz.z, (float) 1e-36))
 
 // currently only used to check orientations
 #define IS_MAT_NULL(v) (IS_VEC_NULL(&(v)->vec.fvec) && IS_VEC_NULL(&(v)->vec.uvec) && IS_VEC_NULL(&(v)->vec.rvec))
--- a/code/math/vecmat.h
+++ b/code/math/vecmat.h
@@ -22,18 +22,32 @@
 //Macros/functions to fill in fields of structures
 //VEC_NULL macros split into two functions in 2009 with commit 75a514b
 
-//macro to check if vector is close to zero or would be close to zero after squaring
+// Null vector checks are performed on the following types of vectors:
+// * orientation component vectors
+// * positions
+// * velocities
+// * normals
+// In each of these cases, FLT_EPSILON or 1.192092896e-07F is a reasonable threshold.
+
+// macro to check if vector is close to zero or would be close to zero after squaring
+// (uses FLT_EPSILON; original threshold was 1e-16 which can be tightened up a bit)
 #define IS_VEC_NULL_SQ_SAFE(v) \
-		(fl_near_zero((v)->xyz.x, (float) 1e-16) && \
-		fl_near_zero((v)->xyz.y, (float) 1e-16) && \
-		fl_near_zero((v)->xyz.z, (float) 1e-16))
+		(fl_near_zero((v)->xyz.x) && \
+		fl_near_zero((v)->xyz.y) && \
+		fl_near_zero((v)->xyz.z))
 
-//macro to check if vector is close to zero
-#define IS_VEC_NULL(v) \
-		(fl_near_zero((v)->xyz.x, (float) 1e-36) && \
-		fl_near_zero((v)->xyz.y, (float) 1e-36) && \
-		fl_near_zero((v)->xyz.z, (float) 1e-36))
+// macro to check if vector is close to zero
+// (original threshold was 1e-36 which was too small)
+#define IS_VEC_NULL(v) IS_VEC_NULL_SQ_SAFE(v)
 
+// macro to check if moment-of-inertia vector is close to zero
+// (uses different thresholds since MOI values are really small)
+#define IS_MOI_VEC_NULL(v) \
+		(fl_near_zero((v)->xyz.x, (float) 1e-20) && \
+		fl_near_zero((v)->xyz.y, (float) 1e-20) && \
+		fl_near_zero((v)->xyz.z, (float) 1e-20))
+
+// currently only used to check orientations
 #define IS_MAT_NULL(v) (IS_VEC_NULL(&(v)->vec.fvec) && IS_VEC_NULL(&(v)->vec.uvec) && IS_VEC_NULL(&(v)->vec.rvec))
 
 //macro to set a vector to zero.  we could do this with an in-line assembly
--- a/code/ai/aibig.cpp
+++ b/code/ai/aibig.cpp
@@ -900,7 +900,7 @@ static void ai_big_maybe_fire_weapons(float dist_to_enemy, float dot_to_enemy)
 		
 		//	Chance of hitting ship is based on dot product of firing ship's forward vector with vector to ship
 		//	and also the size of the target relative to distance to target.
-		if (dot_to_enemy > MAX(0.5f, 0.90f + aip->ai_accuracy/10.0f - En_objp->radius/MAX(1.0f,dist_to_enemy))) {
+		if (dot_to_enemy > std::max(0.5f, 0.90f + aip->ai_accuracy / 10.0f - En_objp->radius / std::max(1.0f, dist_to_enemy))) {
 
 			ship *temp_shipp;
 			temp_shipp = &Ships[Pl_objp->instance];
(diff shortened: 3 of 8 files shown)


code/math/floating.h
```diff
@@ -102,10 +102,13 @@ inline bool fl_near_zero(float a, float e = std::numeric_limits<float>::epsilon(
...
 
-// sees if two floating point numbers are within the minimum tolerance
+// sees if two floating point numbers are approximately equal, taking into account the argument magnitudes
+// see commit c62037
+// and see also this article, because fl_equal may need to be rewritten if it is recruited into more demanding situations:
+// https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/
 inline bool fl_equal(float a, float b)
 {
-	return fl_near_zero(a - b);
+	return fl_abs(a - b) <= FLT_EPSILON * MAX(1.0f, MAX(fl_abs(a), fl_abs(b)));
```
BMagnu, line 111: Since MAX is implemented as a macro using a ternary first comparing and then returning it's argument, you're not only doubling the ternary of the inner MAX call, but also the calls to fl_abs. This can be either fixed by storing the arguments in temporary variables, or by using a function (such as std::max) instead of a macro

BMagnu (changes requested):

code/math/vecmat.h
```diff
@@ -22,18 +22,32 @@
...
+		fl_near_zero((v)->xyz.y) && \
+		fl_near_zero((v)->xyz.z))
+
+// macro to check if vector is close to zero
+// (original threshold was 1e-36 which was too small)
+#define IS_VEC_NULL(v) IS_VEC_NULL_SQ_SAFE(v)
+
+// macro to check if moment-of-inertia vector is close to zero
+// (uses different thresholds since MOI values are really small)
+#define IS_MOI_VEC_NULL(v) \
```
wookieejedi, line 45: Why is this `1e-20` now instead of the previous `1e-36`? There's models with MOI values smaller than `1e-20`, and I'm guessing changing this value from anything besides what it was previously will lead to un-wanted/un-needed print statements.

code/math/vecmat.h
```diff
@@ -22,18 +22,32 @@
...
-		(fl_near_zero((v)->xyz.x, (float) 1e-36) && \
-		fl_near_zero((v)->xyz.y, (float) 1e-36) && \
-		fl_near_zero((v)->xyz.z, (float) 1e-36))
-
+		(fl_near_zero((v)->xyz.x) && \
+		fl_near_zero((v)->xyz.y) && \
+		fl_near_zero((v)->xyz.z))
+
+// macro to check if vector is close to zero
+// (original threshold was 1e-36 which was too small)
```
JohnAFernandez, line 40: According to what part of the article or what other research?  I read the article, and I don't see how it points it 1e-36 being too small.

JohnAFernandez (changes requested):

code/math/vecmat.h
Goober5000: I looked at a number of models, including the TSC 2 and the Sathanas, and the smallest magnitude I found was e-19.  Where did you find smaller values?

wookieejedi: Ah, FotG has some that go as small as `e-24`.

code/math/vecmat.h
Goober5000: The "too small" comment isn't based on the article, but on typical usage within FSO.  See the list of four use cases.

I looked at typical magnitudes in coordinates and normals, and in my assessment our floating point values are "effectively zero" long before they ever get to a magnitude of 1e-36.  Even MOI values of 1e-16 to 1e-19 are outliers compared to our usual range of values.

The determination of the threshold is closer to an art than a science.  I suspect whoever came up with that original threshold was thinking about `FLT_MIN`, which is 1e-37.

code/math/vecmat.h
Goober5000: I should also mention that I ran through several missions using a test build based on this PR, and every mission functioned properly without throwing any null vector or null MOI warnings.

code/math/vecmat.h
JohnAFernandez: But the point of this is to make sure that a vector is safe to use, not that it satisfies the range of known fso use cases. It is correct to keep the value as it is, because no one who uses this in the future is going to be thinking, "I am checking to make sure it is in line with known fso use case values."  They are going to be thinking, "I need to make sure bad data is not used here, because other values are going to be based off these." Values down to 1e-36 are still safe to use according to what I've found, and changing low level values because we haven't seen them in our experience is not a good idea.

code/math/vecmat.h
PhantomHoover: Why not just leave it as-is? What are we gaining from introducing warnings on this data?

code/math/floating.h
Goober5000: I found a few other places where this should be done too.

code/math/vecmat.h
Goober5000: The point is to determine what a sensible threshold should be.  If FotG has values down to e-24, then that indicates a better threshold is e-25.

PhantomHoover:
As a general rule, changes to the FSO engine should not reject data that was valid in previous versions of the engine unless there's a very good reason to. We've had *major* problems in recent releases because this principle wasn't respected. This 'cleanup' change is actually subtly changing the behaviour of fundamental numerical operations and I just don't see what the point is in doing that without any concrete bugs or incorrectness to be fixed.

code/math/vecmat.h
Baezon: Based on the way these actual numbers are being used, the main float accuracy concern is that the matrix's determinant still lies within the type's representational bound (since it involves cubes of quantities, this is quite a concern), but since this calculation was changed to use doubles in d92c7a8ea3799adff961d4e3d24f34bdf5f5c5b4, there's no way for this to happen with just floats as input.

Given that, if it can fit in a float, it is theoretically sound. Since this works against forces, the results of using extremely small MOIs tends toward 0, and not something potentially more damaging like infinity.

code/math/vecmat.h
Goober5000: "Whether a vector is safe" is intrinsically tied to "known FSO use cases".  That's one of the takeaways from the article: evaluating floating point values is highly dependent on context.  Even something as natural as using FLT_EPSILON as a default threshold falls down for floating point values greater than 2.0, because the difference between any two floats greater than 2.0 is greater than FLT_EPSILON.

We want to catch errors and bad data.  The way to do this is to determine what the range of values is expected to be, and then set the threshold to be close to, but outside, that range.  If the threshold is too far outside that range, then we risk letting through data that is bad but that still falls within the too-generous threshold.

code/math/vecmat.h
Goober5000: Does this mean that arbitrarily small floating point values can still be valid MOIs?

Or here is another question: do we, in practice, see any MOI matrices where every value is smaller than e-24?

Goober5000:
The intention is never to reject data that was really and truly valid in a previous version.  The objective here is to be deliberate about what valid data *actually is*.  If data should never be smaller than 1e-07, then the threshold should be set at 1e-07, not 1e-36.

Or to use a visual metaphor, if a road is 24 feet wide, and the standard lane width is 12 feet, then it is appropriate to make two 12-feet lanes.  It's not appropriate to keep it one lane because "the road has always had only one lane".  Conversely, if a road is 16 feet wide, then making two 8-feet lanes is not appropriate.  But those are two different situations.

code/math/vecmat.h
Baezon: There is currently going to be some lower bound on extant pofs, excluding those which are all 0, but I don't think it's right to say 'the limit is whatever the current examples are', the decision should be based on the actual way in which these values are _used_. And the way in which they are used I think allows for very very small values. Subnormal values are where the bad things really start, and that starts at e-38. Given it appears the old value was e-36 that still gives 100x of wiggle room.

TLDR: I think we would be better served by taking the 'floor' and adding padding up from there, than taking a crude estimate of what the current lowest is, and subtracting padding down.

Goober5000: Ok, I can go with that.

code/math/vecmat.h
JohnAFernandez: Can you point to any instance where a float being between 1e-7 and 1e-36 caused an actual issue?

Goober5000: That's the problem... if any values slipped through that should have been zero but weren't, and if the magnitudes were  between 1e-7 and 1e-36, we would not have known.  They would have caused logic errors that resulted in weird behavior, but such behavior would not have been flagged.

JohnAFernandez: We haven't documented any unexplained weird behaviors, to my knowledge.  The most significant unexplained issues have been balance changes in some campaigns, but if you are pointing to changes since the value was changed to 1e-36 from 1e-16, then we should be changing the value back to 1e-16 and not going in the other direction. You have yet to explain your reasoning for picking epsilon, because epsilon in the article is not suggested for comparing against zero.

Re-reading the article, the advice that I see is that when comparing against zero, we should use a constant value, which we are.  Floats are still normal up to 1e-36, and the previous value was also constant.  Saying that we are getting logic errors by using 1e-36 is dubious and needs specific examples. We need either a known issue or a specific potential issue that we can logically follow, examine, and potentially critique.  If you aren't sure whether such an issue could exist, then you haven't actually finished your research.  Also, If you need a separate value for model translations, you should specifically say so and explain why and possibly implement a separate macro.

I'm looking at the different places where `IS_VEC_NULL` and `IS_VEC_NULL_SQ_SAFE` are used, and the uses I see are "Is this vector safe enough to do <operation> with?"  

So for instance, thinking abstractly about one use I saw, "Can I do a dot product with this vector?" Doing a dot product requires normalizing the vector.  And if one of the float values of the vector is 1e-35 and the other two are 0.0f, then it would come up with a realistic value when normalized, easily: 1.0f, 0.0f, 0.0f.  Or maybe 1e-35, 1e-40, 1e-40.  Still a realistic value when normalized because the 1e-36 is 4 orders of magnitude greater. You lose a little bit of accuracy if it is something like 1e-35, 1e-36, 1e-36, because there's only so much accuracy that the values can have that close to zero, but that is pretty rare.  Now if you are worried about that edge case, just changing it back to 1e-16, its previous value, then normalizing something like 1e-16 1e-16 would probably not present that same kind of issue.  And that is all that I can think of in this case.  

And the other thing is `IS_VEC_NULL` is used 9 times in ai code (more than 10% of the total use cases), and adjusting these values to more strict than retail, without a good reason could actually lead to subtle ai changes or the weird behavior that you are trying to avoid.

There needs to be a specific reason to change it to float epsilon and not back to its previous value, 1e-16.

Goober5000: In retail, this is what the code looked like:
```
//macro to check if vector is zero
#define IS_VEC_NULL(v) (((v)->x == (float)0.0) && ((v)->y == (float)0.0) && ((v)->z == (float)0.0))
```
The 1e-16 threshold was introduced in commit 06ef99 in 2009 and was, as far as I can tell, arbitrary.

My reasoning is simply that if we are going to check for an arbitrary threshold, then it should be a threshold that is close to the boundary of normal usage.  Epsilon, at 1e-07, appears reasonable here.

The alternative is to choose a threshold based on mathematical analysis, which is preferred, but I wouldn't know how to choose a threshold based on that method.  @Baezon's reasoning from the matrix's determinant is great for the MOI case, but not relevant to the orientation, normal, or vector case.

Goober5000: Mainly I want a good reason to pick whatever strategy we pick.  To be fair, a threshold of epsilon is somewhat arbitrary, but as far as I can tell the 1e-16 threshold is arbitrary too.

Here is what @BMagnu said, which I have permission to quote:
> I think we can argue both ways in a sensible manner:
Pro's to keeping 1e-36:
Keeping it won't suddenly break things for things that *somehow* were below the threshold but worked fine. There is no danger of creating a requirement to update some obscure old mod that might break from this. Additionally, we don't know of any actual example that would immediately fix an error that is slipping by with 1e-36. Most of the math we use should be fine as long as the floats aren't denormal or would cause an INF overflow when we do 1/x. 1e-36 should be sufficient for that.
> 
> Pro's to changing to 1e-7:
There might be things that, at values smaller than 1e-7, are data which can be properly processed and not crash, but are semantically invalid data. This would be then caught and reported. In FSO's coordinate systems, any location or velocity (and maybe even acceleration) closer than 1e-7 is so small that it may as well be a hard 0, semantically (though this may not be the case for other ~~mis~~uses of the vec3d struct)

The latter paragraph is essentially my position.  Furthermore, tightening the threshold would let us know for sure, and the threshold can always be loosened later.

JohnAFernandez: That will teach me (not for the first time) to double check that I've gone far enough back on git blame.

I would prefer that we go with the original threshold after retail's buggy check was changed. `1e-16`. But Lafiel's suggestion on discord that we try out the change is a good one.  So let's go with that.

JohnAFernandez (approved):

PR #4562, opened 2022-08-19
Modernise CFLAGS on Mac/Linux

PhantomHoover:
Per GCC optimisation guide, which applies equally well to clang:
* -funroll-loops should not be used unless directed by profiling. The compiler will already unroll loops when it's a good idea.
* clang and gcc now have -Og optimisation levels designed to enable debugging with good performance.

### DIFF
--- a/cmake/toolchain-apple-clang.cmake
+++ b/cmake/toolchain-apple-clang.cmake
@@ -4,7 +4,7 @@ MESSAGE(STATUS "Doing configuration specific to Apple Clang...")
 # TODO: Actually add Mac specific flags here...
 
 # Suppress specific warning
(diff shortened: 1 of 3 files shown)


EatThePath:
It looks like the failed CI checks are for warnings in code you didn't remotely touch.
would it make sense to rebase on the current master and trigger another round of builds, to see if CI gets it's head out of it's butt, or maybe just force a merge and make sure the next nightly builds linux fine?

EatThePath:
After discord scp discussion it seems very likely that CI is just being confused by the lack of code changes, and I'm going to force the merge with some followup checking to make sure builds still work.

EatThePath (approved):

Goober5000:
Superseded by #5375

PR #4565, opened 2022-08-19
Add $Mod Title and $Mod Version options to game_settings.tbl

### LINKED ISSUE #3417: FR: Mod Version in game_settings.tbl
Pretty minor request, but as someone who juggles multiple mod versions for FSU and BtA, I would find this pretty handy.

In the bottom left corner of the mainhall screen, FSO prints the FSO version. Super helpful. I'd love to be able to do the same thing for the mod version and just pull that data from game_settings.tbl. This could also be printed at the top of the debug log like the FSO version is, though admittedly the log gives more information if the player is using Knossos because the folder structure includes the mod version.

Either...
$Mod Name: Freespace Upgrade
$Mod Version: 4.4.1

Or just

$Mod: Freespace Upgrade 4.4.1

Yes, this would mean that mods that ship without this in game_settings.tbl would risk getting the data pulled from a dependency. I'm not sure there's a solution to that other than encouraging more mods to ship with a proper game_settings.tbl.

Alternatively, maybe there's a way to get this info from Knossos that I didn't think of?

MjnMixael:
Add $Mod Title: and $Mod Version: to game_settings.tbl and display them on the mainhall similar to FSO version. Also add base Lua access (again just like FSO version) to those entries. Resolves #3417

### DIFF
--- a/code/scripting/api/libs/base.cpp
+++ b/code/scripting/api/libs/base.cpp
@@ -582,6 +582,47 @@ ADE_FUNC(getVersionString, l_Base, nullptr,
 	return ade_set_args(L, "s", str.c_str());
 }
 
+ADE_FUNC(getModTitle, l_Base, nullptr,
+         "Returns the title of the current mod as defined in game_settings.tbl. Will return an empty string if not defined.",
+         "string", "The mod title")
+{
+	auto str = Mod_title;
+	return ade_set_args(L, "s", str.c_str());
+}
+
+ADE_FUNC(getModVersion, l_Base, nullptr,
+         "Returns the version of the current mod as defined in game_settings.tbl. Will return an empty string if not defined.",
+         "string, number, number, number", "The mod title; the major, minor, patch version numbers -1 if invalid")
+{
+	auto version = Mod_version;
+
+	int major = -1;
+	int minor = -1;
+	int patch = -1;
+	
+	int i = 0;
+	auto str = Mod_version;
+	while (i <= 3) {
+		size_t pos = str.find_first_of('.');
+		if (pos != SCP_string::npos) {
+			auto ver = str.substr(0, pos);
+			if(!ver.empty() && std::find_if(ver.begin(), ver.end(), [](char c) { return !std::isdigit(c); }) == ver.end()){
+				if (major < 0) {
+					major = std::stoi(str.c_str());
+				} else if (minor < 0) {
+					minor = std::stoi(str.c_str());
+				}
+				str.erase(0, pos + 1);
+			}
+		} else if ((!str.empty() && std::find_if(str.begin(), str.end(), [](char c) { return !std::isdigit(c); }) == str.end()) && (major > -1) && (!minor > -1)) {
+			patch = std::stoi(str.c_str());
+		}
+		i++;
+	}
+
+	return ade_set_args(L, "siii", version.c_str(), major, minor, patch);
+}
+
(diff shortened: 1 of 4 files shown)


code/scripting/api/libs/base.cpp
```diff
@@ -582,6 +582,47 @@ ADE_FUNC(getVersionString, l_Base, nullptr,
...
+		if (pos != SCP_string::npos) {
+			auto ver = str.substr(0, pos);
+			if(!ver.empty() && std::find_if(ver.begin(), ver.end(), [](char c) { return !std::isdigit(c); }) == ver.end()){
+				if (major < 0) {
+					major = std::stoi(str.c_str());
+				} else if (minor < 0) {
+					minor = std::stoi(str.c_str());
+				}
+				str.erase(0, pos + 1);
+			}
```
JohnAFernandez, line 616: So because it's intended to not use the semantic version at the end of the string if it contains something else first, we should do ` else if (major < 0) { break; }`.  Also add a comment that clarifies that this is intended to keep the function returning only one or other type.  Even though it won't really optimize the code that much, it will make the code easier to follow since since we are explicitly avoiding it instead of implicitly.

I think the one thing you can improve on is more comments in general, too.  Especially when we get somewhat complicated code like this.  Not that it's impossible to read, but comments make things a lot easier.

code/scripting/api/libs/base.cpp
```diff
@@ -582,6 +582,47 @@ ADE_FUNC(getVersionString, l_Base, nullptr,
...
+ADE_FUNC(getModTitle, l_Base, nullptr,
+         "Returns the title of the current mod as defined in game_settings.tbl. Will return an empty string if not defined.",
+         "string", "The mod title")
+{
+	auto str = Mod_title;
+	return ade_set_args(L, "s", str.c_str());
+}
+
+ADE_FUNC(getModVersion, l_Base, nullptr,
+         "Returns the version of the current mod as defined in game_settings.tbl. Will return an empty string if not defined.",
```
JohnAFernandez, line 594: You'll definitely want to add the special functionality that you are intending to the documentation here.

JohnAFernandez (changes requested):

code/scripting/api/libs/base.cpp
```diff
@@ -582,6 +582,47 @@ ADE_FUNC(getVersionString, l_Base, nullptr,
...
+			auto ver = str.substr(0, pos);
+			if(!ver.empty() && std::find_if(ver.begin(), ver.end(), [](char c) { return !std::isdigit(c); }) == ver.end()){
+				if (major < 0) {
+					major = std::stoi(str.c_str());
+				} else if (minor < 0) {
+					minor = std::stoi(str.c_str());
+				}
+				str.erase(0, pos + 1);
+			}
+		} else if ((!str.empty() && std::find_if(str.begin(), str.end(), [](char c) { return !std::isdigit(c); }) == str.end()) && (major > -1) && (minor > -1)) {
```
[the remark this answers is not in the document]
MjnMixael: Done. And thanks for the explanation. Helps me learn.

code/scripting/api/libs/base.cpp
MjnMixael: Done

code/scripting/api/libs/base.cpp
MjnMixael: Explained better.

JohnAFernandez (approved):

PR #4567, opened 2022-08-20
fix stepped rotation

Goober5000:
As part of the cleanup in preparation for model translation, there was a bit of code review that was actually incorrect.  The calculation of the current step needs to be rounded down, because the step isn't actually complete until it fully reaches the next number.  Rounding the step causes the submodel to jump around in all sorts of weird ways.

Follow-up to #4443.


BMagnu (approved):
Evidently this slipped past me in the last PR. Rounding down is obviously the correct thing to do.
explicitly flooring before casting to int is kind of doubling up for guaranteed positive numbers, but oh well

Goober5000:
That's true, the `floor` isn't technically needed here.

PR #4569, opened 2022-08-20
convert some timestamps to the new system

Goober5000:
Mostly debris and asteroid timestamps, but also arc timestamps for ships.


JohnAFernandez (approved):
LGTM

PR #4571, opened 2022-08-20
Make cutscenes.tbl extendable with -csn.tbm

### LINKED ISSUE #3693: FR: Modular Cutscenes.tbl
I honestly didn't think I'd ever need this, but here we are. I'd like some basic modular features added to custscenes.tbl. I'd like to be able to add cutscenes using a .tbm.

The reason for this is that I'm creating an optional "bonus features" package for BtA. Think of it like special features on blurays. Missions/models that were cut... and also, as applies here, cutscenes.

MjnMixael:
Allows adding additional cutscenes from a modular table. Resolves #3693

### DIFF
--- a/code/cutscene/cutscenes.cpp
+++ b/code/cutscene/cutscenes.cpp
@@ -66,8 +66,6 @@ void parse_cutscene_table(const char* filename)
 		skip_to_string("#Cutscenes");
 		ignore_white_space();
 
-		bool isFirstCutscene = true;
-
 		while (required_string_either("#End", "$Filename:"))
 		{
 			required_string("$Filename:");
@@ -91,9 +89,8 @@ void parse_cutscene_table(const char* filename)
 
 			cutinfo.flags.reset();
 
-			if (isFirstCutscene)
+			if (Cutscenes.empty())
 			{
-				isFirstCutscene = false;
 				// The original code assumes the first movie is the intro, so make it viewable
 				cutinfo.flags.set(Cutscene::Cutscene_Flags::Viewable);
 			}
--- a/code/cutscene/cutscenes.cpp
+++ b/code/cutscene/cutscenes.cpp
@@ -35,7 +35,6 @@ const char* Cutscene_mask_name[GR_NUM_RESOLUTIONS] = {
 		"2_ViewFootage-m"
 };
 
-int Num_cutscenes;
 int Description_index;
 SCP_vector<cutscene_info> Cutscenes;
 
@@ -51,7 +50,7 @@ void cutscene_close()
 
 static cutscene_info *get_cutscene_pointer(char *cutscene_filename)
 {
-	for (int i = 0; i < Num_cutscenes; i++) {
+	for (int i = 0; i < (int)Cutscenes.size(); i++) {
 		if (!stricmp(cutscene_filename, Cutscenes[i].filename)) {
 			return &Cutscenes[i];
 		}
@@ -93,6 +92,8 @@ void parse_cutscene_table(const char* filename)
 
 		while (required_string_either("#End", "$Filename:"))
 		{
+			bool csn_new = false;
+			
 			required_string("$Filename:");
 			stuff_string(csnt.filename, F_PATHNAME, MAX_FILENAME_LEN);
 
@@ -118,13 +119,14 @@ void parse_cutscene_table(const char* filename)
 			} else {
 				// Don't create cutscene if it has +nocreate and is in a modular table.
 				if (!create_if_not_found && Parsing_modular_table) {
-					if (!skip_to_start_of_string_either("$Fileame:", "#End")) {
+					if (!skip_to_start_of_string_either("$Filename:", "#End")) {
 						error_display(1, "Missing [#End] or [$Filename] after cutscene %s", csnt.filename);
 					}
-					return;
+					continue;
 				}
 				Cutscenes.push_back(csnt);
-				csnp = &Cutscenes[Num_cutscenes++];
+				csn_new = true;
+				csnp = &Cutscenes[Cutscenes.size() -1];
 			}
 
 			//If $Name is not found use the $Filename instead because this field needs data
@@ -132,8 +134,8 @@ void parse_cutscene_table(const char* filename)
 			if (optional_string("$Name:")) {
 				stuff_string(csnp->name, F_NAME, NAME_LENGTH);
 				mprintf(("Adding cutscene %s\n", csnp->name));
-			} else if (!Parsing_modular_table) {
-				strncpy(csnp->name, csnt.filename, 32);
+			} else if (strlen(csnp->name) == 0) {
+				strcpy_s(csnp->name, csnt.filename);
 				mprintf(("Missing $Name: in cutscene.tbl. Using filename %s instead.\n", csnt.filename));
 			}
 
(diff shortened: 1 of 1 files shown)


Goober5000 (changes requested):
Looks reasonable, but there are two side effects that will need to be considered:
1) The `isFirstCutscene` logic sets the first cutscene parsed to be visible by default.  With modular tables, this will now happen to the first cutscene in every file, whereas it should only happen to the first cutscene in cutscenes.tbl.  I suggest getting rid of that variable and changing the check to `if (Cutscenes.empty())`.
2) There are no checks for duplicate cutscene entries, although to be fair there weren't any in the previous version of the code either.  Currently every entry parsed will be added, even if it's a duplicate.

MjnMixael:
Addressed by adding +nocreate support

code/cutscene/cutscenes.cpp
```diff
@@ -35,6 +35,7 @@ const char* Cutscene_mask_name[GR_NUM_RESOLUTIONS] = {
 		"2_ViewFootage-m"
 };
 
+int Num_cutscenes;
```
Goober5000, line 38: Since `Cutscenes` is a vector, a `Num_cutscenes` variable is redundant.  Change the loop comparison to `i < (int)Cutscenes.size()`, and change the pointer assignment to `csnp = &Cutscenes[Cutscenes.size() - 1];` or `csnp = &Cutscenes.back();`

code/cutscene/cutscenes.cpp
```diff
@@ -48,38 +49,101 @@ void cutscene_close()
...
+				create_if_not_found = false;
+			}
+
+			// Does this cutscene exist already?
+			// If so, load this new info into it
+			csnp = get_cutscene_pointer(csnt.filename);
+			if (csnp != NULL) {
+				if (Parsing_modular_table) {
+					nprintf(
+						("Warning", "More than one version of cutscene %s exists; using newer version.\n", csnt.filename));
```
Goober5000, line 112: Should this really be a warning?  In most cases a modular table modifying a previously parsed entry is a normal thing.

code/cutscene/cutscenes.cpp
```diff
@@ -48,38 +49,101 @@ void cutscene_close()
...
+						"Error:  Cutscene %s already exists.  All cutscene names must be unique.",
+						csnt.filename);
+				}
+			} else {
+				// Don't create cutscene if it has +nocreate and is in a modular table.
+				if (!create_if_not_found && Parsing_modular_table) {
+					if (!skip_to_start_of_string_either("$Fileame:", "#End")) {
+						error_display(1, "Missing [#End] or [$Filename] after cutscene %s", csnt.filename);
+					}
+					return;
```
Goober5000, line 124: This will stop parsing the entire file.  Use `continue` instead of `return`, so the code flow will resume parsing at the next iteration of the loop.

code/cutscene/cutscenes.cpp
```diff
@@ -48,38 +49,101 @@ void cutscene_close()
...
+						("Warning", "More than one version of cutscene %s exists; using newer version.\n", csnt.filename));
+				} else {
+					error_display(1,
+						"Error:  Cutscene %s already exists.  All cutscene names must be unique.",
+						csnt.filename);
+				}
+			} else {
+				// Don't create cutscene if it has +nocreate and is in a modular table.
+				if (!create_if_not_found && Parsing_modular_table) {
+					if (!skip_to_start_of_string_either("$Fileame:", "#End")) {
```
Goober5000, line 121: Typo in `$Filename:`

code/cutscene/cutscenes.cpp
```diff
@@ -48,38 +49,101 @@ void cutscene_close()
...
-			drop_white_space(buf);
-			compact_multitext_string(buf);
-			cutinfo.description = vm_strdup(buf);
+			//If $Name is not found use the $Filename instead because this field needs data
+			//but only do this if if we're parsing the tbl and not a tbm
+			if (optional_string("$Name:")) {
+				stuff_string(csnp->name, F_NAME, NAME_LENGTH);
+				mprintf(("Adding cutscene %s\n", csnp->name));
+			} else if (!Parsing_modular_table) {
+				strncpy(csnp->name, csnt.filename, 32);
```
Goober5000, line 136: `strncpy` is a bit messy to use, and this would have to be tweaked in a couple of ways, but fortunately it isn't even needed here.  Replace this line with
```
strcpy_s(csnp->name, csnt.filename);
```

code/cutscene/cutscenes.cpp
```diff
@@ -48,38 +49,101 @@ void cutscene_close()
...
-			stuff_string(buf, F_MULTITEXT, sizeof(buf));
-			drop_white_space(buf);
-			compact_multitext_string(buf);
-			cutinfo.description = vm_strdup(buf);
+			//If $Name is not found use the $Filename instead because this field needs data
+			//but only do this if if we're parsing the tbl and not a tbm
+			if (optional_string("$Name:")) {
+				stuff_string(csnp->name, F_NAME, NAME_LENGTH);
+				mprintf(("Adding cutscene %s\n", csnp->name));
+			} else if (!Parsing_modular_table) {
```
Goober5000, line 135: There is a subtle bug here: if a brand new entry is added in a modular table, and this entry does not have a `$Name:` field, the code would allow a blank name to slip through.  So I suggest changing
```
} else if (!Parsing_modular_table) {
```
to
```
} else if (strlen(csnp->name) == 0) {
```

code/cutscene/cutscenes.cpp
```diff
@@ -88,31 +152,29 @@ void cutscene_init()
 				stuff_int(&junk);
 			}
 
-			cutinfo.flags.reset();
+			csnp->flags.reset();
```
Goober5000, line 155: This will clear the flags for any modular table entry that modifies an existing entry, even if the new modular table entry doesn't modify flags.  To avoid this, you will need to enclose this in a check with a `first_time` variable.  This variable should default to false but be set to true after the line `Cutscenes.push_back(csnt);`

Goober5000 (changes requested):
I was thinking of just a check for a duplicate entry along with a warning; I didn't expect full `+nocreate` support.  This is fine, but it adds quite a bit of complexity, which requires a bit more refinement.

code/cutscene/cutscenes.cpp
MjnMixael: Engine Wash parsing (where most of this was read and adapted from) has a similar warning which is why it also exists here. I don't care either way. Just trying to match the style elsewhere in the engine.

Goober5000: Hmm.  I'd say engine washes are a special case, and the warning isn't needed here.  Looking through the code, engine washes are the only other place this sort of warning is displayed.

code/cutscene/cutscenes.cpp
```diff
@@ -88,31 +154,31 @@ void cutscene_init()
 				stuff_int(&junk);
 			}
 
-			cutinfo.flags.reset();
+			if (csn_new == true) {
```
Goober5000, line 157: Since `csn_new` is a boolean variable, this can simply be `if (csn_new)`.

Goober5000 (changes requested):
Just a couple other things.  One comment here, and see also the comment on the engine washes.

Goober5000 (approved):

PR #4572, opened 2022-08-20
Allow changing Animation Instance speed

BMagnu:
This enables scripting to change the speed multiplier of any animation. If it is not yet running, the speed change will also affect the animation once it is started.

### DIFF
--- a/code/model/modelanimation.cpp
+++ b/code/model/modelanimation.cpp
@@ -258,11 +258,12 @@ namespace animation {
 		animList.parentSet->initializeSubmodelBuffer(pmi, applyBuffer);
 
 		for (const auto& anim : animList.animationList) {
-			switch (anim->m_instances[pmi->id].state) {
+			const auto& instanceData = anim->m_instances[pmi->id];
+			switch (instanceData.state) {
 			case ModelAnimationState::RUNNING_FWD:
 			case ModelAnimationState::RUNNING_RWD:
 			case ModelAnimationState::NEED_RECALC:
-				anim->play(frametime, pmi, applyBuffer);
+				anim->play(frametime * instanceData.speed, pmi, applyBuffer);
 				break;
 			case ModelAnimationState::COMPLETED:
 			case ModelAnimationState::PAUSED:
@@ -787,6 +788,11 @@ namespace animation {
 		for (auto anim : animations)
 			anim->m_instances[pmi->id].instance_flags.set(flag, set);
 	}
+
+	void ModelAnimationSet::AnimationList::setSpeed(float speed) const {
+		for (auto anim : animations)
+			anim->m_instances[pmi->id].speed = speed;
+	};
 
 	ModelAnimationSet::AnimationList& ModelAnimationSet::AnimationList::operator+=(const AnimationList& rhs) {
 		Assertion(pmi == rhs.pmi, "Tried to concatenate two AnimationLists of different model instances!");
--- a/code/model/modelanimation.cpp
+++ b/code/model/modelanimation.cpp
@@ -774,7 +774,7 @@ namespace animation {
 	int ModelAnimationSet::AnimationList::getTime() const {
 		float duration = 0.0f;
 
-		for (auto anim : animations) {
+		for (const auto& anim : animations) {
 			if (anim->m_instances[pmi->id].state == ModelAnimationState::UNTRIGGERED)
 				continue;
 			float localDur = anim->m_instances[pmi->id].duration;
@@ -785,12 +785,12 @@ namespace animation {
 	}
 
 	void ModelAnimationSet::AnimationList::setFlag(Animation_Instance_Flags flag, bool set) const {
-		for (auto anim : animations)
+		for (const auto& anim : animations)
 			anim->m_instances[pmi->id].instance_flags.set(flag, set);
 	}
 
 	void ModelAnimationSet::AnimationList::setSpeed(float speed) const {
-		for (auto anim : animations)
+		for (const auto& anim : animations)
 			anim->m_instances[pmi->id].speed = speed;
 	};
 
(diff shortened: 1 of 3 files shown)


code/scripting/api/objs/ship.cpp
```diff
@@ -1649,6 +1649,34 @@ ADE_FUNC(triggerSubmodelAnimation, l_Ship, "string type, string triggeredBy, [bo
 	return Ship_info[shipp->ship_info_index].animations.parseScripted(model_get_instance(shipp->model_instance_num), animtype, trigger).start(forwards ? animation::ModelAnimationDirection::FWD : animation::ModelAnimationDirection::RWD, forced || instant, instant, pause) ? ADE_RETURN_TRUE : ADE_RETURN_FALSE;
 }
 
+ADE_FUNC(setAnimationSpeed, l_Ship, "string type, string triggeredBy, [number speedMultiplier = 1.0]",
+	"Sets the speed multiplier at which an animation runs. Anything other than 1 will not work in multiplayer. Type is the string name of the animation type, "
```
JohnAFernandez, line 1653: Because mission time is synced now, there is probably a way to transmit an update to the speed and have the client sync the rotation, as long as mission time from `Multi_timing_info._current_time` in `multi_time_manager.cpp` is sent with it.  Any other change to animations can probably be handled the same way.

That's for the future, of course, assuming anyone asks for it.

code/scripting/api/objs/ship.cpp
```diff
@@ -1649,6 +1649,34 @@ ADE_FUNC(triggerSubmodelAnimation, l_Ship, "string type, string triggeredBy, [bo
...
+
+	auto animtype = animation::anim_match_type(type);
+	if (animtype == animation::ModelAnimationTriggerType::None)
+		return ADE_RETURN_FALSE;
+
+	ship* shipp = &Ships[objh->objp->instance];
+
+	Ship_info[shipp->ship_info_index].animations.parseScripted(model_get_instance(shipp->model_instance_num), animtype, trigger).setSpeed(speed);
+
+	return ADE_RETURN_NIL;
```
JohnAFernandez, line 1677: Why NIL here?

code/scripting/api/objs/ship.cpp
BMagnu: Yes. Eventually, I want this and the new instance animation flags to sync.
Trouble with syncinc speed is the use case. I expect this value to be updated every frame in some uses to make the animation speed (e.g. of walker legs) match the ship speed. Hence sending things each frame, especially in larger values, could lead to serious desync. So even if we sync this speed value, we'll also need some mechanism that occasionally helps to resync animations over time. But yeah, that is definitely for the future and will need careful consideration

JohnAFernandez: Dropped packets are the enemy here.... Rare as they are.

code/scripting/api/objs/ship.cpp
BMagnu: You could return true on success, but given that the only failure mode really is a misspelled animation type or a nil ship object in the first place, it's not really something where a program would need to dynamically know whether the operation succeeded

code/scripting/api/objs/ship.cpp
BMagnu: Absolutely

code/scripting/api/objs/ship.cpp
JohnAFernandez: It might end up being useful in some cases to know the difference between a nil ship object and a successful animation adjustment.  But then again, maybe they more often than not have checked that the ship object is not nil already?

BMagnu: Yeah, basically most scripts do check ship:isValid() before doing stuff with it. And should be doing it, really.

JohnAFernandez (approved):

code/model/modelanimation.cpp
```diff
@@ -788,6 +789,11 @@ namespace animation {
 			anim->m_instances[pmi->id].instance_flags.set(flag, set);
 	}
 
+	void ModelAnimationSet::AnimationList::setSpeed(float speed) const {
+		for (auto anim : animations)
+			anim->m_instances[pmi->id].speed = speed;
+	};
```
Goober5000, line 795: Is this actually a `const` function if it mutates its own state?

And it looks like it doesn't actually mutate it, since the loop variable `anim` is a copy, not a reference.

code/scripting/api/objs/ship.cpp
```diff
@@ -1649,6 +1649,34 @@ ADE_FUNC(triggerSubmodelAnimation, l_Ship, "string type, string triggeredBy, [bo
...
+
+	if (!ade_get_args(L, "oss|f", l_Ship.GetPtr(&objh), &type, &trigger, &speed))
+		return ADE_RETURN_NIL;
+
+	if (!objh->IsValid())
+		return ADE_RETURN_NIL;
+
+	auto animtype = animation::anim_match_type(type);
+	if (animtype == animation::ModelAnimationTriggerType::None)
+		return ADE_RETURN_FALSE;
```
Goober5000, line 1671: Is `ADE_RETURN_FALSE` the right thing to do here, given your comments in the other conversation?

Goober5000 (changes requested):

code/model/modelanimation.cpp
BMagnu: it creates a copy of a shared_ptr, of which it modifies the data.
Yes, this could also be a const reference to the shared_ptr (and should be, really). But since it just modifies the ptr's data and not the ptr itself, the List, which just holds the ptr stays constant

code/scripting/api/objs/ship.cpp
BMagnu: True, this also just needs to return nil, as the other function

Goober5000 (approved):

code/model/modelanimation.cpp
Goober5000: Weird but ok.

PR #4574, opened 2022-08-20
Allow tech-room to auto switch to single-player mode

wookieejedi:
Many buttons in the main menu auto switch to either multi-player mode or single-player mode. For example, the `Campaign` door always switches to single-player, while the `Multiplayer` door auto switches to multiplayer mode. This is very useful for mods, especially with new players who are not sure where and how to switch between multi and single player pilots.

This PR adds an option to allow mods to auto switch to single-player for the `Techroom`, which is especially useful for playing simulator missions. The method used here is the same method that the campaign door uses to auto switch to single-player.

Tested and works as expected.

### DIFF
--- a/code/menuui/mainhallmenu.cpp
+++ b/code/menuui/mainhallmenu.cpp
@@ -30,6 +30,7 @@
 #include "menuui/playermenu.h"
 #include "menuui/snazzyui.h"
 #include "mission/missioncampaign.h"
+#include "mod_table/mod_table.h"
 #include "network/multi.h"
 #include "network/multi_voice.h"
 #include "network/multiui.h"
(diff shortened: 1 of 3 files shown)


code/menuui/mainhallmenu.cpp
```diff
@@ -917,6 +918,12 @@ void main_hall_do(float frametime)
 
 				// clicked on the tech room region
 				case TECH_ROOM_REGION:
+					// Check if Techroom should auto-switch to singleplayer mode --wookieejedi
+					if (Techroom_autoswitches_to_singleplayer) {
+						Player->flags &= ~PLAYER_FLAGS_IS_MULTI;
+						Game_mode = GM_NORMAL;
+					}
+
```
JohnAFernandez, line 926: I really think the actual logic for this should be in Freespace.cpp during the game state change.  And we need to keep track of whether or not we were in Multiplayer before, and here's why:

If we were in multi mode where the mainhall has only a ready room region without a dedicated multiplayer region, then switching to the tech room will force it to single player, and the user will have to go to the barracks to switch it back to multiplayer so that the ready room region will take them to the PXO or game list screen.

JohnAFernandez (changes requested):

code/menuui/mainhallmenu.cpp
wookieejedi: Ah yeah that's a good point. I reckon moving all the logic into that same space would be useful? IE taking out the existing little chunks here and moving all the mode switch logic to the one section during the game state change.

wookieejedi: Heh as it turns out, going through all that checking made me realize that I can do all of this via existing scripting, so no need to clutter up state change code with a new PR.

PR #4575, opened 2022-08-20
Prioritize players in object update

JohnAFernandez:
This makes servers update player ships more often, even when they are not front and center in front of the player.  This hopefully will cut down on some of the player jitteriness.  Because client player ships have extra latency in a server-based system, they should have been prioritized when the system was changed over between FS1 and FS2.

Testing has occurred, and it is definitely not making anything worse, although our multiplayer testing base is not as scientific these days.

### DIFF
--- a/code/network/multi_obj.cpp
+++ b/code/network/multi_obj.cpp
@@ -202,67 +202,80 @@ float multi_oo_calc_pos_time_difference(int player_id, int net_sig_idx);
 #define OO_SUBSYS_TIME				1000
 
 // timestamp values for object update times based on client's update level.
-// Cyborg17 - This is the one update number I have adjusted, because it's the player's target.
+// Cyborg - Many of these values have been adjusted over time, but not based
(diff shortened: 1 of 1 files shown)


BMagnu (approved):
Looks safe enough.
The values need indeed to be properly computed to bandwidth at some point, though they seem reasonable for a first guess (although I find it somewhat strange you reduced some update rates. Did you run into actual measurable trouble for some players with weaker internet?)

JohnAFernandez:
That is a great question. The LAN values were originally written under the assumption that they would only be used by people running LAN games.  Now that we have really high bandwidths that the high setting could not completely utilize, most players set this setting to LAN.  So it should be treated more as a "very high" setting instead of being an infinite fount of ship updates.  The ones I reduced are for ships that the player cannot see with the default camera that are at least 500 meters away from the player.  I also increased another value, for the same kind of reason: the update time should not be exactly the same for both High and LAN.

PR #4577, opened 2022-08-21
Enable LuaAI to use more arguments

BMagnu:
Previously, LuaAI could only use _one_ singular target parameter that had to conform to the OSWPT type.
Now, LuaAI can take any number (including repetitive) of arguments, with the restrictions that all non-target-parameter arguments will be nil if called as a player order, and that they cannot be used within target validation for player orders.


Goober5000 (approved):
I don't have any comments on this one, so approved. 👍

PR #4578, opened 2022-08-21
Script function for `turnTowardsOrientation`

wookieejedi:
A simple copy of the same function on `ai_helper`, now usable on plain ships too.

### DIFF
--- a/code/scripting/api/objs/ship.cpp
+++ b/code/scripting/api/objs/ship.cpp
@@ -1008,6 +1008,30 @@ ADE_FUNC(turnTowardsPoint,
 	return ADE_RETURN_NIL;
 }
 
+ADE_FUNC(turnTowardsOrientation,
+	l_Ship,
+	"orientation target, [boolean respectDifficulty = true, vector turnrateModifier /* 100% of tabled values in all rotation axes by default */]",
+	"turns the ship towards the specified orientation during this frame",
+	nullptr,
+	nullptr)
+{
+	object_h ship;
+	matrix_h* target;
+	bool diffTurn = true;
+	vec3d* modifier = nullptr;
+
+	int argnum = ade_get_args(L, "oo|bo", l_Ship.Get(&ship), l_Matrix.GetPtr(&target), &diffTurn, l_Vector.GetPtr(&modifier));
+	if (argnum == 0) {
+		return ADE_RETURN_NIL;
+	}
+
+	matrix* mat = target->GetMatrix();
+	vec3d targetVec = mat->vec.fvec + ship.objp->pos;
+
+	ai_turn_towards_vector(&targetVec, ship.objp, nullptr, nullptr, 0.0f, (diffTurn ? 0 : AITTV_FAST), &mat->vec.rvec, modifier);
+	return ADE_RETURN_NIL;
+}
+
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
If this is a copy, can the two functions be consolidated somehow?  Maybe two script functions could delegate to the same implementation function?

wookieejedi:
Good question. I was following @Baezon's lead with his copy of `turnTowardsPoint`, but would be happy to discuss thoughts about how to go about this.

Goober5000:
Look at the `AxisActionInverted_sub` function, which basically the same function repeated multiple times, with each virtvar corresponding to its own movement axis.

Baezon:
There's very little to consolidate, the functions basically do nothing but grabbing lua values and sending them to the according function.

wookieejedi:
Yeah, the actual ai function call is a line, so adding an additional level of consolidation seems like a bit too much added abstraction/complexity?

code/scripting/api/objs/ship.cpp
```diff
@@ -1008,6 +1008,30 @@ ADE_FUNC(turnTowardsPoint,
...
+	bool diffTurn = true;
+	vec3d* modifier = nullptr;
+
+	int argnum = ade_get_args(L, "oo|bo", l_Ship.Get(&ship), l_Matrix.GetPtr(&target), &diffTurn, l_Vector.GetPtr(&modifier));
+	if (argnum == 0) {
+		return ADE_RETURN_NIL;
+	}
+
+	matrix* mat = target->GetMatrix();
+	vec3d targetVec = mat->vec.fvec + ship.objp->pos;
```
JohnAFernandez, line 1029: I think this would be the part that you could encapsulate.  Something like `vm_find_goal_fvec(vec3d* dest, matrix* mat, vec3d* pos)
{
*dest = mat->vec.fvec + pos;
}`

The point in making this function is to standardize handling, reduce copy and paste errors, and make the method more easily discoverable. But it's a small difference, so the utility in creating it is not as high, imo.  I'm hitting approve.

JohnAFernandez (approved):

code/scripting/api/objs/ship.cpp
wookieejedi: Good point, that makes sense. @Goober5000 does encapsulating in that way sound good to you as well?

Goober5000: Hmm. I was thinking that both functions could delegate to a single helper function, but that `l_Ship` vs `l_AI_Helper` throws a wrench into things.  Plus, `turnTowardsPoint` is already duplicated.

That suggested function would be nice, but it's such a small change it may not be worth it.  As said, the utility is not high.

I'll approve as-is, no changes required.

Goober5000 (approved):

code/scripting/api/objs/ship.cpp
wookieejedi: Sounds good. Yeah, if in the future a PR to encapsulate both is warranted by the addition of more LUA AI and plain ship AI modes then that might be useful, but for now I'll follow the previous lead and keep as is. Thanks for the discussion!

PR #4579, opened 2022-08-21
Implement renderTechModel() and renderTechModel2() for weapons

### LINKED ISSUE #4353: FR: renderTechModel() API for weapons
Exactly like the ships version, but for weapon models. Should default to the $Tech Model: if it exists, otherwise fallback to the base $Model File:

MjnMixael:
Since Tech model files are not required for weapons, this version exits early and returns false if Tech model does not exist. I partially implemented falling back to the regular POF file if it exists and tech model doesn't, but since in weapons the regular POF file do not have closeup_pos or closeup_zoom defined, it wasn't as good. Thought it was better to stick with just the tech model, which does get those properties defined, as that's what the method is called in the first place.

Resolves #4353

### DIFF
--- a/code/scripting/api/objs/weaponclass.cpp
+++ b/code/scripting/api/objs/weaponclass.cpp
@@ -4,6 +4,8 @@
 #include "weaponclass.h"
 #include "model.h"
 #include "weapon/weapon.h"
+#include "graphics/matrix.h"
+#include "vecmath.h"
 
 namespace scripting {
 namespace api {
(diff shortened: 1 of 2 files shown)


Goober5000 (changes requested):
Piggybacking on @Baezon 's comments from Discord, I think some consolidation would be useful here:

1) In shipclass.cpp, partition the code from line 714 (`//Clip`) through line 750 (`gr_reset_clip()`) into a separate function, with a function header like this:
```
void render_tech_model_helper(int modelnum, const matrix *orient, const char *team_name = nullptr)
```
2) Move the `render_info` variable inside this function, along with the call to `render_info.set_team_color` which will be done if the team name is non-null
3) Call this function from all four `renderTechModel` Lua functions

I don't think it's worth deprecating the Lua function names in favor of new ones.  Both `renderTechModel` and `renderTechModel2` have been around for a long time, so it makes sense to keep the names consistent in weaponclass.cpp.

Let me know if any of that isn't clear or needs additional information.

EatThePath:
I think a general consolidation of tech rendering functions even beyond these new ones would probably be good. I intend to add some more configurability to them by the end of the year anyway, @Goober5000, @Baezon, would it be acceptable to have this merged and assign me a cleanup that includes that consolidation?

Goober5000:
Sure, I'd be ok with that.

Goober5000 (approved):

PR #4582, opened 2022-08-22
Model translation

Goober5000:
This implements model translation, including the following:
* submodel support
* animation support
* basic POF continuous translation
* basic POF stepped translation
* SEXPs to control translation
* scripting support to read and write submodel offsets

This is merely the latest step in a long line of PRs leading to this point, starting with Swifty's submodel instance refactor and continuing through my submodel overhaul PR's and Lafiel's animation PRs.

~~This PR is in draft status until #4560 is merged and until we can coordinate a multi version bump.~~


code/network/multi.h
```diff
@@ -71,9 +71,10 @@ class player;
 // version 55 - 8/28/2021 - Adding multi-compatible animations
 // version 56 - 8/28/2021 - Fix animations for 22_0 release
 // version 57 - 6/5/2022 - Upgrade interpolation, fix multiplayer sexp handling, and enable player orders to exceed 16
+// version 58 - ??/??/2022 - Submodel translation
```
BMagnu, line 74: TBD

code/parse/sexp.cpp
```diff
@@ -20389,23 +20415,44 @@ void sexp_rotating_subsys_set_turn_time(int node)
...
+		}
+		if (is_rotation)
+			subsys->submodel_instance_1->turn_accel = PI2 / accel;
+		else
+			subsys->submodel_instance_1->shift_accel = accel;
 	}
 	else
-		subsys->submodel_instance_1->current_turn_rate = PI2 / turn_time;
+	{
+		if (is_rotation)
```
BMagnu, line 20451: Should this not also set the instant_translate_accel flag?

code/scripting/api/objs/modelinstance.cpp
```diff
@@ -148,6 +148,29 @@ ADE_VIRTVAR(Orientation, l_SubmodelInstance, "orientation", "Gets or sets the su
 	return ade_set_args(L, "o", l_Matrix.Set(matrix_h(&smi->canonical_orient)));
 }
 
+ADE_VIRTVAR(Offset, l_SubmodelInstance, "vector", "Gets or sets the translated submodel instance offset", "vector", "Offset, or zero vector if handle is not valid")
```
BMagnu, line 151: Since there is a difference between orientation and offset in there being a non-instance offset, I'd like a clarification here that this is _only_ the offset relative to the already existing submodel offset-to-model-center

code/scripting/api/objs/subsystem.cpp
```diff
@@ -156,6 +156,31 @@ ADE_VIRTVAR(GunOrientation, l_Subsystem, "orientation", "Orientation of turret g
 	return ade_set_args(L, "o", l_Matrix.Set(matrix_h(&smi->canonical_orient)));
 }
 
+ADE_VIRTVAR(Offset, l_Subsystem, "vector", "Translated offset of subobject or turret base", "vector", "Offset, or zero vector if handle is not valid")
```
BMagnu, line 159: same here

BMagnu (changes requested):
Looks mostly good, Some questions and comments

code/parse/sexp.cpp
Goober5000: Good question; I had to think about that one.  I don't think this is necessary... in both the rotation and translation cases, the current rate is set in addition to the desired rate.  This will instantly cause the submodel to move at the new velocity.

Furthermore, the SEXP is a one-time operation; if the flag were to be set, then every future change in velocity would also have instantaneous acceleration.

code/scripting/api/objs/modelinstance.cpp
Goober5000: Ok.

code/scripting/api/objs/subsystem.cpp
Goober5000: Ok.

code/parse/sexp.cpp
BMagnu: Okay, makes sense

BMagnu (approved):
We need to actually remember to amend to comments for the multi bump, but otherwise, looks good to me now.

code/network/multi.h
```diff
@@ -71,9 +71,10 @@ class player;
...
 // STANDALONE_ONLY
 
-#define MULTI_FS_SERVER_VERSION							57
+#define MULTI_FS_SERVER_VERSION							58
```
JohnAFernandez, line 77: We won't need a bump until we change the packing function `multi_pack_unpack_subsystem_list` in multiultil.cpp to actually pack and unpack the new information.

JohnAFernandez (changes requested):

Goober5000:
Ok, I left the comments modified but removed the multi version bump.  I also rebased onto the latest master.  I'll wait until the next nightly before taking this out of draft status.

JohnAFernandez (approved):

PR #4583, opened 2022-08-22
Fix cfread string's handling of null terminated strings

Baezon:
This function stuffs a string from a file into a buffer and requires that the length of the string be strictly less than the size of the buffer it was going in, so that it could forcibly add a 0 after the end of the string and still fit in the buffer. This is a little more conservative than necessary, since it prevents properly null terminated strings of length, say, 32 from fitting in a 32 buffer, because it wants to add another at the end, and even more problematic if the file is trying to 4 byte align and has padded the string up to 32. 
Having a string equal to the size of the buffer is only a problem if it contains no terminator at all, which an assertion has been added for, and in which case it will simply overwrite the last character, otherwise, it is safe simply to add the terminator after the string within the buffer in case it didn't come with one.

Also fixes some mistaken BUFFER_SIZE - 1 arguments to this function, the argument should always match exactly the actual length of the buffer being used.

### DIFF
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -1243,11 +1243,13 @@ void cfread_string_len(char *buf,int n, CFILE *file)
 {
 	int len;
 	len = cfread_int(file);
-	Assertion( (len < n), "len: %i, n: %i", len, n );
-	if (len)
+	Assertion( (len <= n), "String cannot fit in buffer, len: %i, buffer: %i", len, n );
+	if (len) {
 		cfread(buf, len, 1, file);
+		Assertion(len != n || buf[len - 1] == 0, "Unterminated string does not fit in buffer");
+	}
 
-	buf[len] = 0;
+	buf[MIN(len, n - 1)] = 0;
 }
 
 SCP_string cfread_string_len(CFILE *file)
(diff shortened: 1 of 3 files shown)


BMagnu (approved):

Goober5000 (changes requested):
One convention that is used somewhat inconsistently in the codebase, but I've tried to follow/enforce when possible, is `len` for the string length and `size` for the buffer size.  Can we rename this function to `cfread_string_len` with the comments and Assertion message changed accordingly?

Alternatively -- which would be my preference, although it is a more widespread change -- we could keep the name `cfread_string_len` and instead change all of the usages to size-1.

Baezon:
I'm fine with being consistent about `len` and `size` here, but even so, `cfread_string_len` is not an inaccurate name. It's fundamentally `cfread`ing a `string` of length `len` from the file, that's the important bit, not really the size of the buffer it's going to stuff it in.

Goober5000:
In that case, can you rename the `n` parameter to something like `buffer_size`, or add some documentation?

Goober5000 (approved):
That works 👍

wookieejedi (dismissed):
It looks like #4590 might need to go in first to prevent headaches?

PR #4584, opened 2022-08-22
Scripting functions necessary for SCPUI Fiction Viewer

MjnMixael:


wookieejedi (approved):
Looked over this earlier today then again now and it all looks good to me. Nice work in picking this up.

JohnAFernandez (approved):
My uninitiated eyes cannot find anything wrong.  We'll deal with anything we missed as they come up.

PR #4585, opened 2022-08-23
fix overriding of originally-empty sounds

Goober5000:
Follow-up to #3743.  A sound that was originally parsed as empty, but then re-parsed as non-empty, would still have the `GAME_SND_NOT_VALID` flag from the original attempt.  This fixes that by clearing the flag each time the filename is read.

This also slightly moves the `gs->pitch_range` assignment so both functions are consistent.


JohnAFernandez (approved):
Good find.

PR #4586, opened 2022-08-23
Scripting functions required for SCPUI conversion of Loop Brief.

MjnMixael:
Should be everything required. Tested both accepting and denying a loop using these functions sends the player to the correct mission and sets the exit loop mission correctly as well.

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -515,7 +515,7 @@ ADE_FUNC(getLoopBrief,
 ADE_FUNC(setLoopChoice, l_UserInterface_LoopBrief, "boolean", "Accepts mission outcome and then True to go to loop, False to skip", nullptr, nullptr)
 {
 	bool choice = false;
-	ade_get_args(L, "b", &choice);
+	ade_get_args(L, "|b", &choice);
 
 	if (choice) {
 		// select the loop mission
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -512,7 +512,7 @@ ADE_FUNC(getLoopBrief,
 	return ade_set_args(L, "o", l_LoopBriefStage.Set(Campaign.missions[Campaign.current_mission]));
 }
 
-ADE_FUNC(setLoopChoice, l_UserInterface_LoopBrief, "boolean", "Accepts mission outcome and then True to go to loop, False to skip", NULL, NULL)
+ADE_FUNC(setLoopChoice, l_UserInterface_LoopBrief, "boolean", "Accepts mission outcome and then True to go to loop, False to skip", nullptr, nullptr)
 {
 	bool choice = false;
 	ade_get_args(L, "b", &choice);
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -6,6 +6,7 @@
 
 #include "cmdline/cmdline.h"
 #include "gamesnd/eventmusic.h"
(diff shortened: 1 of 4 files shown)


wookieejedi (approved):
Also looks overall good to me, and good to hear that you have successes with testing it.

code/scripting/api/libs/ui.cpp
```diff
@@ -491,5 +492,42 @@ ADE_FUNC(runBriefingStageHook, l_UserInterface_CmdBrief, "number oldStage, numbe
...
+	"loop_brief_stage",
+	"The loop brief data")
+{
+	return ade_set_args(L, "o", l_LoopBriefStage.Set(Campaign.missions[Campaign.current_mission]));
+}
+
+ADE_FUNC(setLoopChoice, l_UserInterface_LoopBrief, "boolean", "Accepts mission outcome and then True to go to loop, False to skip", nullptr, nullptr)
+{
+	bool choice = false;
+	ade_get_args(L, "b", &choice);
```
Goober5000, line 516: If this is intended to be a required parameter, then there should be an `if (!ade_get_args...)` check with an appropriate return value.

Alternatively, this could be considered an optional parameter, defaulting to the value of `choice`.  In this case, an `if()` check is unnecessary, but the parameter string should be "|b" to indicate that the boolean is optional.

code/scripting/api/objs/loop_brief.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+//**********HANDLE: loop_briefing
+ADE_OBJ(l_LoopBriefStage, cmission, "loop_brief_stage", "Loop Brief stage handle");
+
+ADE_VIRTVAR(Text,
+	l_LoopBriefStage,
+	nullptr,
+	"The text file of the stage",
+	"loop_brief_stage",
+	"The text file")
```
Goober5000, line 14: These should be "text", not "text file", methinks.

code/scripting/api/objs/loop_brief.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+	return ade_set_args(L, "s", current->mission_branch_desc);
+}
+
+ADE_VIRTVAR(AniFilename,
+	l_LoopBriefStage,
+	nullptr,
+	"The text file of the stage",
+	"loop_brief_stage",
+	"The text file")
```
Goober5000, line 33: These should not be "text file".

code/scripting/api/objs/loop_brief.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+	return ade_set_args(L, "s", current->mission_branch_brief_anim);
+}
+
+ADE_VIRTVAR(AudioFilename,
+	l_LoopBriefStage,
+	nullptr,
+	"The text file of the stage",
+	"loop_brief_stage",
+	"The text file")
```
Goober5000, line 52: Nor these.

Goober5000 (changes requested):
A few comments.

MjnMixael:
Fixes made

code/scripting/api/objs/loop_brief.cpp
```diff
@@ -11,7 +11,7 @@ ADE_VIRTVAR(Text,
 	nullptr,
 	"The text file of the stage",
```
Goober5000, line 12: This still says "text file".  Everything else looks good though.

MjnMixael: Done

Goober5000 (approved):

PR #4590, opened 2022-08-25
Virtual / Modular POFs

BMagnu:
This PR adds the basic framework to handle table-defined virtual POFs. These allow modular manipulation and combination of models during FSO runtime.
Right now, only the very basic features (copying submodel model data, renaming submodels, moving submodels, editing some POF header data) are supported, basically only allowing some dynamic kitbashing.
Due to the already large size of the PR, the following planned features will be deferred to a later PR:

- Replacing Submodels
- Processing more than just model data when copying / replacing
    - Allow to manipulate turret data
    - Allow to manipulate weapon fire point data
    - Allow to manipulate glowpoint data
    - Allow to manipulate dockbay paths and data
- Allow copying / replacement of said data individually, without needing to copy parent submodels to the target ship
- Enable shield modification and replacement
- Enable debris copying and replacement
- Enable eye point and more POF header data manipulation
- Automatize some processes currently manual
    - Enable the system to automatically rename submodels with conflicting names
    - Enable automatic bounding box / radius recalculation to fit new / changed submodels into the POF
- Optimization
    - Caching. Right now, Virtual POFs will only cache the end product. If a model is required multiple times for generating the final virtual POF, it will be loaded anew each time it is referenced.
- Likely more I've forgotten at the moment.

### DIFF
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -887,31 +887,27 @@ struct model_read_deferred_tasks {
 		int gun_subobj_nr;
 		vec3d turretNorm;
 		int n_slots;
-		std::vector<vec3d> firingpoints;
+		SCP_vector<vec3d> firingpoints;
 	};
 
 	struct texture_idx_replace {
-		std::map<int, int> replacementIds;
+		SCP_map<int, int> replacementIds;
 	};
 
 	//Key: Subsystem Name
-	std::unordered_map<SCP_string, model_subsystem_parse> model_subsystems;
+	SCP_unordered_map<SCP_string, model_subsystem_parse, SCP_string_lcase_hash, SCP_string_lcase_equal_to> model_subsystems;
 	//Key: Subsystem Name
-	std::unordered_map<SCP_string, engine_subsystem_parse> engine_subsystems;
+	SCP_unordered_map<SCP_string, engine_subsystem_parse, SCP_string_lcase_hash, SCP_string_lcase_equal_to> engine_subsystems;
 	//Key: Parent Subobject Nr
-	std::unordered_map<int, weapon_subsystem_parse> weapons_subsystems;
+	SCP_unordered_map<int, weapon_subsystem_parse> weapons_subsystems;
 	//Key: Subobject Nr
-	std::unordered_map<int, texture_idx_replace> texture_replacements;
+	SCP_unordered_map<int, texture_idx_replace> texture_replacements;
 };
 
 class model_parse_depth {
-	std::unordered_map<SCP_string, int> depth{};
+	SCP_unordered_map<SCP_string, int, SCP_string_lcase_hash, SCP_string_lcase_equal_to> depth{};
 public:
-	model_parse_depth() = default;
-	model_parse_depth(const model_parse_depth&) = default;
-	model_parse_depth& operator=(const model_parse_depth&) = default;
-	inline int& operator[](SCP_string name) {
-		SCP_tolower(name);
+	inline int& operator[](const SCP_string& name) {
 		return depth[name];
 	}
 };
@@ -967,7 +963,7 @@ void model_delete_instance(int model_instance_num);
 // Goober5000
 void model_load_texture(polymodel *pm, int i, char *file);
 
-std::set<int> model_get_textures_used(polymodel* pm, int submodel);
+SCP_set<int> model_get_textures_used(polymodel* pm, int submodel);
 
 // Returns a pointer to the polymodel structure for model 'n'
 polymodel *model_get(int model_num);
@@ -1441,7 +1437,7 @@ void model_page_in_textures(int modelnum, int ship_info_index = -1);
 // given a model, unload all of its textures
 void model_page_out_textures(int model_num, bool release = false);
 // given a model, without respect to usage state of the polymodel
-void model_page_out_textures(polymodel* pm, bool release = false, const std::set<int>& skipTextures = {}, const std::set<int>& skipGlowBanks = {});
+void model_page_out_textures(polymodel* pm, bool release = false, const SCP_set<int>& skipTextures = {}, const SCP_set<int>& skipGlowBanks = {});
 
 void model_do_intrinsic_motions(object *objp);
 
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -1176,7 +1176,7 @@ void submodel_get_cross_sectional_avg_pos(int model_num, int submodel_num, float
 void submodel_get_cross_sectional_random_pos(int model_num, int submodel_num, float z_slice_pos, vec3d* pos);
   
 extern int model_find_submodel_index(int modelnum, const char *name);
-extern int model_find_submodel_index(const polymodel& pm, const char* name);
+extern int model_find_submodel_index(const polymodel* pm, const char* name);
 
 // gets the index into the docking_bays array of the specified type of docking point
 // Returns the index.  second functions returns the index of the docking bay with
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -909,7 +909,7 @@ void model_free_all();
 void model_instance_free_all();
 
 // Loads a model from disk and returns the model number it loaded into.
-int model_load(const char *filename, int n_subsystems, model_subsystem *subsystems, int ferror = 1, int duplicate = 0);
+int model_load(const char *filename, int n_subsystems, model_subsystem *subsystems, int ferror = 1, int duplicate = 0, int depth = 0);
 
 int model_create_instance(int objnum, int model_num);
 void model_delete_instance(int model_instance_num);
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -868,6 +868,13 @@ public:
 	vertex_buffer detail_buffers[MAX_MODEL_DETAIL_LEVELS];
 };
 
+struct model_subsystem_parse {
+	int subobj_nr;
+	float rad;
+	vec3d pnt;
+	SCP_string props;
+};
+
 // Iterate over a submodel tree, starting at the given submodel root node, and running the given function for each node.  The function's signature should be:
 //
 // void func(int submodel, int level, bool isLeaf);
@@ -909,7 +916,7 @@ void model_free_all();
 void model_instance_free_all();
 
 // Loads a model from disk and returns the model number it loaded into.
-int model_load(const char *filename, int n_subsystems, model_subsystem *subsystems, int ferror = 1, int duplicate = 0, int depth = 0);
+int model_load(const char *filename, int n_subsystems, model_subsystem *subsystems, int ferror = 1, int duplicate = 0);
 
 int model_create_instance(int objnum, int model_num);
 void model_delete_instance(int model_instance_num);
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -868,11 +868,32 @@ public:
 	vertex_buffer detail_buffers[MAX_MODEL_DETAIL_LEVELS];
 };
 
-struct model_subsystem_parse {
-	int subobj_nr;
-	float rad;
-	vec3d pnt;
-	SCP_string props;
+struct subsystem_parse_list {
+	struct model_subsystem_parse {
+		int subobj_nr;
+		float rad;
+		vec3d pnt;
+		SCP_string props;
+	};
+
+	struct engine_subsystem_parse {
+		int thruster_nr;
+	};
+
+	struct weapon_subsystem_parse {
+		int turret_nr;
+		int gun_subobj_nr;
+		vec3d turretNorm;
+		int n_slots;
+		std::vector<vec3d> firingpoints;
+	};
+
+	//Key: Subsystem Name
+	std::unordered_map<SCP_string, model_subsystem_parse> model_subsystems;
+	//Key: Subsystem Name
+	std::unordered_map<SCP_string, engine_subsystem_parse> engine_subsystems;
+	//Key: Parent Subobject Nr
+	std::unordered_map<int, weapon_subsystem_parse> weapons_subsystems;
 };
 
 // Iterate over a submodel tree, starting at the given submodel root node, and running the given function for each node.  The function's signature should be:
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -902,13 +902,7 @@ struct model_read_deferred_tasks {
 	SCP_unordered_map<int, texture_idx_replace> texture_replacements;
 };
 
-class model_parse_depth {
-	SCP_unordered_map<SCP_string, int, SCP_string_lcase_hash, SCP_string_lcase_equal_to> depth{};
-public:
-	inline int& operator[](const SCP_string& name) {
-		return depth[name];
-	}
-};
+using model_parse_depth = SCP_unordered_map<SCP_string, int, SCP_string_lcase_hash, SCP_string_lcase_equal_to>;
 
 // Iterate over a submodel tree, starting at the given submodel root node, and running the given function for each node.  The function's signature should be:
 //
@@ -1182,6 +1176,7 @@ void submodel_get_cross_sectional_avg_pos(int model_num, int submodel_num, float
 void submodel_get_cross_sectional_random_pos(int model_num, int submodel_num, float z_slice_pos, vec3d* pos);
   
 extern int model_find_submodel_index(int modelnum, const char *name);
+extern int model_find_submodel_index(const polymodel& pm, const char* name);
 
 // gets the index into the docking_bays array of the specified type of docking point
 // Returns the index.  second functions returns the index of the docking bay with
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -23,6 +23,8 @@
 #include "object/object.h"
 #include "ship/ship_flags.h"
 
+#include <set>
+
 class object;
 class ship_info;
 class model_render_params;
@@ -931,6 +933,8 @@ void model_init();
 
 // call to unload a model (works like bm_unload()), "force" SHOULD NEVER BE SET outside of modelread.cpp!!!!
 void model_unload(int modelnum, int force = 0);
+// Directly frees polymodel data, regardless of state and usage in ship classes. Use with caution. Will not clear textures
+void model_free(polymodel* pm);
 
 // Call to free all existing models
 void model_free_all();
@@ -944,6 +948,8 @@ void model_delete_instance(int model_instance_num);
 
 // Goober5000
 void model_load_texture(polymodel *pm, int i, char *file);
+
+std::set<int> model_get_textures_used(polymodel* pm, int submodel);
 
 // Returns a pointer to the polymodel structure for model 'n'
 polymodel *model_get(int model_num);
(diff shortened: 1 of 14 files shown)


fred2/management.cpp
```diff
@@ -411,6 +412,7 @@ bool fred_init(std::unique_ptr<os::GraphicsOperations>&& graphicsOps)
 	animation::ModelAnimationParseHelper::parseTables();
 	obj_init();
 	model_free_all();				// Free all existing models
+	virtual_pof_init();
```
Goober5000, line 415: This also needs to be initialized in qtFRED.

code/model/model.h
```diff
@@ -23,6 +23,8 @@
 #include "object/object.h"
 #include "ship/ship_flags.h"
 
+#include <set>
+
```
Goober5000, line 27: This shouldn't be #included; we have `SCP_` equivalents of most collections.  See comments below.

code/model/model.h
```diff
@@ -868,6 +870,52 @@ class polymodel
...
+	struct engine_subsystem_parse {
+		int thruster_nr;
+	};
+
+	struct weapon_subsystem_parse {
+		int turret_nr;
+		int gun_subobj_nr;
+		vec3d turretNorm;
+		int n_slots;
+		std::vector<vec3d> firingpoints;
```
Goober5000, line 890: Use `SCP_vector` instead.

code/model/model.h
```diff
@@ -868,6 +870,52 @@ class polymodel
...
+	};
+
+	struct texture_idx_replace {
+		std::map<int, int> replacementIds;
+	};
+
+	//Key: Subsystem Name
+	std::unordered_map<SCP_string, model_subsystem_parse> model_subsystems;
```
Goober5000, line 898: Use `SCP_unordered_map` instead - here and on the following lines.

code/model/model.h
```diff
@@ -868,6 +870,52 @@ class polymodel
...
+	struct weapon_subsystem_parse {
+		int turret_nr;
+		int gun_subobj_nr;
+		vec3d turretNorm;
+		int n_slots;
+		std::vector<vec3d> firingpoints;
+	};
+
+	struct texture_idx_replace {
+		std::map<int, int> replacementIds;
```
Goober5000, line 894: Use `SCP_map` instead.

code/model/model.h
```diff
@@ -868,6 +870,52 @@ class polymodel
...
+	std::unordered_map<int, texture_idx_replace> texture_replacements;
+};
+
+class model_parse_depth {
+	std::unordered_map<SCP_string, int> depth{};
+public:
+	model_parse_depth() = default;
+	model_parse_depth(const model_parse_depth&) = default;
+	model_parse_depth& operator=(const model_parse_depth&) = default;
+	inline int& operator[](SCP_string name) {
```
Goober5000, line 913: This operator can be removed.

code/model/model.h
```diff
@@ -868,6 +870,52 @@ class polymodel
...
+	//Key: Subsystem Name
+	std::unordered_map<SCP_string, engine_subsystem_parse> engine_subsystems;
+	//Key: Parent Subobject Nr
+	std::unordered_map<int, weapon_subsystem_parse> weapons_subsystems;
+	//Key: Subobject Nr
+	std::unordered_map<int, texture_idx_replace> texture_replacements;
+};
+
+class model_parse_depth {
+	std::unordered_map<SCP_string, int> depth{};
```
Goober5000, line 908: This can be replaced with
```
SCP_unordered_map<SCP_string, int, SCP_string_lcase_hash, SCP_string_lcase_equal_to>
```

code/model/model.h
```diff
@@ -917,6 +967,8 @@ void model_delete_instance(int model_instance_num);
 // Goober5000
 void model_load_texture(polymodel *pm, int i, char *file);
 
+std::set<int> model_get_textures_used(polymodel* pm, int submodel);
```
Goober5000, line 970: Use `SCP_unordered_set` instead.

code/model/model.h
```diff
@@ -1388,6 +1440,8 @@ void model_page_in_textures(int modelnum, int ship_info_index = -1);
 
 // given a model, unload all of its textures
 void model_page_out_textures(int model_num, bool release = false);
+// given a model, without respect to usage state of the polymodel
+void model_page_out_textures(polymodel* pm, bool release = false, const std::set<int>& skipTextures = {}, const std::set<int>& skipGlowBanks = {});
```
Goober5000, line 1444: Use `SCP_unordered_set` instead.

code/model/modelreplace.h
```diff
@@ -0,0 +1,58 @@
...
+
+class VirtualPOFOperation {
+public:
+	virtual ~VirtualPOFOperation() = default;
+	virtual void process(polymodel*, model_read_deferred_tasks& subsysList, model_parse_depth depth) const = 0;
+};
+
+struct VirtualPOFDefinition {
+	SCP_string name, basePOF;
+	std::vector<std::unique_ptr<VirtualPOFOperation>> operationList;
```
Goober5000, line 22: Use `SCP_vector`

code/model/modelreplace.h
```diff
@@ -0,0 +1,58 @@
...
+};
+
+class VirtualPOFOperationRenameSubobjects : public VirtualPOFOperation {
+	std::unordered_map<SCP_string, SCP_string> replacements;
```
Goober5000, line 26: Use `SCP_unordered_map`

code/model/modelreplace.h
```diff
@@ -0,0 +1,58 @@
...
+public:
+	VirtualPOFOperationRenameSubobjects();
+	void process(polymodel* pm, model_read_deferred_tasks& deferredTasks, model_parse_depth depth) const override;
+};
+
+class VirtualPOFOperationAddSubmodel : public VirtualPOFOperation {
+	SCP_string subobjNameSrc, subobjNameDest;
+	SCP_string appendingPOF;
+	std::unique_ptr<VirtualPOFOperationRenameSubobjects> rename = nullptr;
+	bool copyChildred = true;
```
Goober5000, line 36: `copyChildren`?

Goober5000 (changes requested):
Partial review.  Will look at it more later.

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,409 @@
...
+		if (operation_parser != virtual_pof_operations.end())
+			toFill.operationList.emplace_back(operation_parser->second());
+		else {
+			error_display(1, "Unknown operation type %s in virtual POF %s!", operation_type, name.c_str());
+			virtual_pof_list.pop_back();
+			skip_to_start_of_string_either("$POF:", "#End");
+			return;
+		}
+	} while (optional_string_either("$POF:", "#End") == -1);
+	Mp -= 5;
```
Goober5000, line 93: Is this backing up to the start of `$POF:`?  What happens if it ends the loop on `#End` instead?

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,409 @@
...
+void virtual_pof_init() {
+	if (cf_exists_full("virtual_pofs.tbl", CF_TYPE_TABLES))
+		parse_virtual_pof_table("virtual_pofs.tbl");
+
+	parse_modular_table(NOX("*-pof.tbm"), parse_virtual_pof_table);
+}
+
+// Internal helper functions
+
+#define REPLACE_IF_EQ(data, source, dest) if ((data) == source) (data) = dest;
```
Goober5000, line 125: Considering that this is a macro, and that all the occurrences are used with parameters called `source` and `dest`, you could rewrite the macro to only require the first parameter -- or alternatively add a second macro that calls the first

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,409 @@
...
+	if (optional_string("$Rename Subobjects:")) {
+		rename = make_unique<VirtualPOFOperationRenameSubobjects>();
+	}
+
+	required_string("+Destination Subobject:");
+	stuff_string(subobjNameDest, F_NAME);
+}
+
+void VirtualPOFOperationAddSubmodel::process(polymodel* pm, model_read_deferred_tasks& deferredTasks, model_parse_depth depth) const {
+	polymodel* appendingPM = new polymodel();
```
Goober5000, line 169: Since you've used smart pointers elsewhere, why not make this a `unique_ptr`?

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,409 @@
...
+	model_read_deferred_tasks appendingSubsys;
+	SCP_set<int> keepTextures;
+	read_model_file(appendingPM, appendingPOF.c_str(), 0, appendingSubsys, depth);
+
+	int src_subobj_no = -1;
+	for (int i = 0; i < appendingPM->n_models; i++) {
+		if (!stricmp(appendingPM->submodel[i].name, subobjNameSrc.c_str()))
+			src_subobj_no = i;
```
Goober5000, line 177: Might want to break when the subobj is found.  Or maybe create a separate `find_submodel` function

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	int dest_subobj_no = -1;
+	for (int i = 0; i < pm->n_models; i++) {
+		if (!stricmp(pm->submodel[i].name, subobjNameDest.c_str()))
+			dest_subobj_no = i;
```
Goober5000, line 183: Same here

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,409 @@
...
+			rename->process(appendingPM, appendingSubsys, depth);
+
+		SCP_vector<int> to_copy_submodels;
+		bool has_name_collision = false;
+		if (copyChildren) {
+			model_iterate_submodel_tree(appendingPM, src_subobj_no, [&to_copy_submodels, &has_name_collision, pm, appendingPM](int submodel, int /*level*/, bool /*isLeaf*/) {
+				to_copy_submodels.emplace_back(submodel);
+				for (int i = 0; i < pm->n_models; i++) {
+					if (!stricmp(pm->submodel[i].name, appendingPM->submodel[submodel].name))
+						has_name_collision = true;
```
Goober5000, line 200: Same here

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,409 @@
...
+				}
+				});
+		}
+		else {
+			to_copy_submodels.emplace_back(src_subobj_no);
+			for (int i = 0; i < pm->n_models; i++) {
+				if (!stricmp(pm->submodel[i].name, appendingPM->submodel[src_subobj_no].name))
+					has_name_collision = true;
```
Goober5000, line 208: Same here

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,409 @@
...
+			}
+		}
+
+		if (!has_name_collision) {
+			//Make sure to keep old data
+			bsp_info* oldSubmodels = pm->submodel;
+			int old_n_submodel = pm->n_models;
+
+			//Realloc new submodel array of proper size
+			pm->n_models += (int)to_copy_submodels.size();
```
Goober5000, line 218: I presume submodel post-processing, such as with `look_at_submodel_names` and `dock_parent_submodel_names`, will be done after all submodel copying is done?

It looks like the bit of code a few lines later with `change_submodel_numbers` handles this, except that it doesn't look like `parent_submodel` on the docking bays is handled.

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,409 @@
...
+		setOffset = make_unique<vec3d>();
+		stuff_vec3d(setOffset.get());
+	}
+}
+
+void VirtualPOFOperationChangeData::process(polymodel* pm, model_read_deferred_tasks& deferredTasks, model_parse_depth /*depth*/) const {
+	int subobj_no = -1;
+	for (int i = 0; i < pm->n_models; i++) {
+		if (!stricmp(pm->submodel[i].name, submodel.c_str()))
+			subobj_no = i;
```
Goober5000, line 372: Here too the comment about breaking out of the loop, or about creating a new function to search for a submodel name

Goober5000 (changes requested):

code/model/modelreplace.h
```diff
@@ -0,0 +1,58 @@
+#pragma once
+
+#include "globalincs/vmallocator.h"
+#include "model/model.h"
+
+#include <memory>
+
+bool model_exists(const SCP_string& filename);
+
+bool read_virtual_model_file(polymodel* pm, const SCP_string& filename, model_parse_depth depth, int ferror, model_read_deferred_tasks& deferredTasks);
```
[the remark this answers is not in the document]
BMagnu: Since virtual POF parsing is basically depth-first searching through the dependency tree, every layer needs to keep a copy of the depth for the next branch of the tree. While the depth needs to propagate downwards the dependency tree, it must not propagate to sibling branches or parents.
Consider the following example:

Virtual pof A replaces model A
Virtual pof B uses model A
Virtual pof C uses model B and model A

If you now try to load model C, the following resolve will happen:
Load virtual model C.
Virtual C loads virtual model B.
Virtual B loads virtual model A (depth = 0).
Virtual A loads real model A (depth = 1).
Virtual C loads virtual model A (depth = 0).
Virtual A loads real model A (depth = 1).

These last two steps are the issue. If depth is passed by reference, then the depth for model A during the "C loads virtual A" step will be 2 (since it would have been incremented the first time we loaded the two A models) instead of the correct 0.

code/model/modelreplace.cpp
BMagnu: Because we need to manually clear out it's dynamically allocated pointers using model_free anyways.
We could change model_free to not actually free the polymodel pointer itself and make it a smart pointer, but since it's still required to call model_free anyways, it'd defeat the purpose. To really make that worthwhile, we'd need to go through the model struct and replace all pointers and dynamically allocated memory with smart pointers (which might be a good idea, but that's out of scope for this PR, and is kinda needed all across the codebase).

code/model/modelreplace.cpp
BMagnu: Yes, that is not yet handled.
For now, it doesn't matter, as only submodel data is copied. Since the original submodels keep their numbers, the docking bays on the old model stay valid. And since you cannot yet copy the docking bays of the other model, there is yet no need to do this.
But as the comment says: With Phase 2, _all_ submodel references in the polymodel will need to be processed, and not only those within the submodel structs

code/model/modelreplace.cpp
Goober5000: Ah, that makes sense.

code/model/modelreplace.cpp
Goober5000: Ok, sounds good.

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,396 @@
...
+				newSubmodel.depth += deltaDepth;
+
+				//Store texture replacement indices
+				const auto& textureIDs = model_get_textures_used(appendingPM, to_copy_submodels[i]);
+				for (const auto& textureID : textureIDs) {
+					auto it = textureIDReplace.find(textureID);
+					if (it == textureIDReplace.end()) {
+						int newID = pm->n_textures++;
+						if (pm->n_textures > MAX_MODEL_TEXTURES) {
+							Warning(LOCATION, "Failed to add submodel to virtual POF, combined POF has too many (over %d) textures.", MAX_MODEL_TEXTURES);
```
Goober5000, line 243: This message should include the POF and submodel names.

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,396 @@
...
+			}
+
+			//Actually copy textures
+			for (const auto& usedTexture : textureIDReplace) {
+				pm->maps[usedTexture.second] = appendingPM->maps[usedTexture.first];
+				keepTextures.emplace(usedTexture.first);
+			}
+		}
+		else {
+			Warning(LOCATION, "Failed to add submodel to virtual POF, original POF already has a subsystem with the same name as was supposed to be added.");
```
Goober5000, line 286: Add POF and submodel names

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,396 @@
...
+		}
+	}
+	else {
+		Warning(LOCATION, "Failed to add submodel to virtual POF, specified subobject was not found. Returning original POF");
```
Goober5000, line 290: Add POF and submodel names

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,396 @@
...
+	
+	model_page_out_textures(appendingPM, true, keepTextures);
+	model_free(appendingPM);
+}
+
+VirtualPOFOperationRenameSubobjects::VirtualPOFOperationRenameSubobjects() {
+	while (optional_string("+Replace:")) {
+		SCP_string replace;
+		stuff_string(replace, F_NAME);
+		SCP_tolower(replace);
```
Goober5000, line 302: I don't think `SCP_tolower` is needed anymore

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,396 @@
...
+		auto it = copy_engine_subsys.find(replace.first);
+		if (it != copy_engine_subsys.end())
+			deferredTasks.engine_subsystems.emplace(replace.second, it->second);
+	}
+}
+
+VirtualPOFOperationChangeData::VirtualPOFOperationChangeData() {
+	required_string("+Submodel:");
+	stuff_string(submodel, F_NAME);
+	SCP_tolower(submodel);
```
Goober5000, line 346: Maybe remove `SCP_tolower`

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,396 @@
...
+		setOffset = make_unique<vec3d>();
+		stuff_vec3d(setOffset.get());
+	}
+}
+
+void VirtualPOFOperationChangeData::process(polymodel* pm, model_read_deferred_tasks& deferredTasks, model_parse_depth /*depth*/) const {
+	int subobj_no = model_find_submodel_index(pm, submodel.c_str());
+
+	if (subobj_no == -1) {
+		Warning(LOCATION, "Failed to find submodel to change data of. Returning original POF");
```
Goober5000, line 358: Add POF and submodel names

code/model/modelreplace.cpp
```diff
@@ -0,0 +1,396 @@
...
+			deferredTasks.engine_subsystems.emplace(replace.second, it->second);
+	}
+}
+
+VirtualPOFOperationChangeData::VirtualPOFOperationChangeData() {
+	required_string("+Submodel:");
+	stuff_string(submodel, F_NAME);
+	SCP_tolower(submodel);
+
+	if (optional_string("+Set Offset:")) {
```
Goober5000, line 348: I'm curious: are these offsets specified from the center of the whole model, or relative to the old submodel center, or relative to the new submodel center?

Goober5000 (changes requested):
A few more things.

code/model/modelreplace.cpp
Goober5000: Needs to be updated to use the new function

code/model/modelreplace.cpp
Goober5000: Here too

code/model/modelreplace.cpp
BMagnu: These offsets just hard-rewrite the offset stored in the submodel data, which means that they'll be relative to the new parent submodel center

Goober5000: Ah, ok

Goober5000 (approved):
Approved 👍

PR #4591, opened 2022-08-25
Scripting functions required for SCPUI conversion of all four Tech Room UIs

MjnMixael:
This adds an ScpUi API for access to the many tech room related bits including the simulator room, cutscenes viewer, and credits screen. I think I covered everything. Anything missed will be in a later PR.

Some files, notably readyroom.cpp and credits.cpp needed some slight rearranging because the mission list and credits data are not created/parsed until that UI is actually loaded. The restructure allows for the API to call those methods and then access the data without loading the rest of the retail UI elements. While a larger restructure could have made the files more efficient overall, I felt this was the best way to accomplish what was needed while keeping the impact to the affected files as minimal as possible. That said, I'm open to instruction how this could be improved if necessary.

### DIFF
--- a/code/menuui/credits.cpp
+++ b/code/menuui/credits.cpp
@@ -218,7 +218,7 @@ static float Credits_scroll_rate			= 15.0f;
 static float Credits_artwork_display_time	= 9.0f;
 static float Credits_artwork_fade_time		= 1.0f;
 
-static SCP_vector<SCP_string> Credit_text_parts;
+credits_info Credits_Info;
 
 static bool Credits_parsed;
 
@@ -258,6 +258,20 @@ void credits_start_music()
 	} else {
 		nprintf(("Warning", "Cannot play credits music\n"));
 	}
+}
+
+char* credits_get_music_filename(const char* music)
+{
+	int credits_spooled_music_index = event_music_get_spooled_music_index(music);
+	if (credits_spooled_music_index != -1) {
+		char* credits_wavfile_name = Spooled_music[credits_spooled_music_index].filename;
+		if (credits_wavfile_name != NULL) {
+			credits_load_music(credits_wavfile_name);
+			return credits_wavfile_name;
+		}
+		return nullptr;
+	}
+	return nullptr;
 }
 
 int credits_screen_button_pressed(int n)
@@ -304,8 +318,11 @@ void credits_parse_table(const char* filename)
 		{
 			int temp;
 			stuff_int(&temp);
-			if (temp > 0)
+			if (temp > 0) {
 				Credits_num_images = temp;
+			}
+
+			Credits_Info.num_images = Credits_num_images;
 		}
 		if (optional_string("$Start Image Index:"))
 		{
@@ -320,24 +337,31 @@ void credits_parse_table(const char* filename)
 			{
 				Credits_artwork_index = Credits_num_images - 1;
 			}
+			Credits_Info.start_index = Credits_artwork_index;
 		}
 		if (optional_string("$Text scroll rate:"))
 		{
 			stuff_float(&Credits_scroll_rate);
-			if (Credits_scroll_rate < 0.01f)
+			if (Credits_scroll_rate < 0.01f) {
 				Credits_scroll_rate = 0.01f;
+			}
+			Credits_Info.scroll_rate = Credits_scroll_rate;
 		}
 		if (optional_string("$Artworks display time:"))
 		{
 			stuff_float(&Credits_artwork_display_time);
-			if (Credits_artwork_display_time < 0.01f)
+			if (Credits_artwork_display_time < 0.01f) {
 				Credits_artwork_display_time = 0.01f;
+			}
+			Credits_Info.art_display_time = Credits_artwork_display_time;
 		}
 		if (optional_string("$Artworks fade time:"))
 		{
 			stuff_float(&Credits_artwork_fade_time);
-			if (Credits_artwork_fade_time < 0.01f)
+			if (Credits_artwork_fade_time < 0.01f){
 				Credits_artwork_fade_time = 0.01f;
+			}
+			Credits_Info.art_fade_time = Credits_artwork_fade_time;
 		}
 		if (optional_string("$SCP Credits position:"))
 		{
@@ -374,6 +398,7 @@ void credits_parse_table(const char* filename)
 			if (first_run && !Credits_parsed && line == mod_check)
 			{
 				credits_text.append(unmodified_credits);
+				Credits_Info.credit_parts.push_back(unmodified_credits);
 			}
 
 			first_run = false;
@@ -382,6 +407,7 @@ void credits_parse_table(const char* filename)
 			{
 				// If the line is empty then just append a newline, don't bother with splitting it first
 				credits_text.append("\n");
+				Credits_Info.credit_parts.push_back("\n");
 			}
 			else
 			{
@@ -403,6 +429,8 @@ void credits_parse_table(const char* filename)
 				{
 					credits_text.append(SCP_string(lines[i], charNum[i]));
 					credits_text.append("\n");
+					Credits_Info.credit_parts.push_back(SCP_string(lines[i], charNum[i]));
+					Credits_Info.credit_parts.push_back("\n");
 				}
 			}
 		}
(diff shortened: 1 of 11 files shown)


code/menuui/credits.cpp
```diff
@@ -260,6 +260,20 @@ void credits_start_music()
 	}
 }
 
+char* credits_get_music_filename(const char* music)
```
BMagnu, line 263: this should really return a ``const char*``, since there is really no usecase in which you want to be able to modify the name of the music filename.

code/scripting/api/libs/ui.cpp
```diff
@@ -643,5 +673,177 @@ ADE_FUNC(getFictionMusicName, l_UserInterface_FictionViewer, nullptr,
...
+	l_UserInterface);
+
+ADE_FUNC(buildMissionList,
+	l_UserInterface_TechRoom,
+	nullptr,
+	"Builds the mission list for display. Must be called before the sim_mission handle will have data",
+	"number",
+	"Returns 1 when completed")
+{
+	API_Access = true;
```
BMagnu, line 691: This is bad style. The ``API_Access`` variable is a global variable in your code. However, it's only used to modify the details of the code path of the function. Yes, this is done in some of the very old volition code as well, but it's bad there as well. Mostly cause it makes it a lot harder to follow where which values are set and what they will do.
Instead, you should pass it as a parameter to ``api_sim_room_build_mission_list`` and subfunctions

code/scripting/api/objs/techroom.h
```diff
@@ -0,0 +1,14 @@
...
+#include "menuui/readyroom.h"
+#include "cutscene/cutscenes.h"
+#include "scripting/ade_api.h"
+
+namespace scripting {
+namespace api {
+
+DECLARE_ADE_OBJ(l_TechRoomMission, sim_mission);
+DECLARE_ADE_OBJ(l_TechRoomCutscene, cutscene_info);
+
```
BMagnu, line 12: Similar to my comment in #4602, these should just store an int, the index into their respective arrays.
This is much better, especially if this API should ever be able to be used by _anything_ other than SCPUI

code/menuui/credits.h
```diff
@@ -12,10 +12,28 @@
...
+	int num_images;							// the mission filename
+	int start_index;						// the mission description
+	float scroll_rate;						// the mission designer
+	float art_display_time;					// the mission designer
+	float art_fade_time;					// the mission designer
+	SCP_vector<SCP_string> credit_parts;	// the lines of credits
+	SCP_string credits_complete;			// the lines of credits
+};
+
+extern credits_info Credits_Info;
```
BMagnu, line 26: I have to ask: What do you need this struct for? When are these values ever different from the globals that you use to fill these up with data?

code/scripting/api/objs/techroom.cpp
```diff
@@ -0,0 +1,150 @@
+#include "techroom.h"
+
+namespace scripting {
+namespace api {
+
+//**********HANDLE: tech missions
+ADE_OBJ(l_TechRoomMission, sim_mission, "sim_mission", "Tech Room mission handle");
+
+ADE_VIRTVAR(Name, l_TechRoomMission, nullptr, "The name of the mission", "sim_mission", "The name")
```
BMagnu, line 9: This returns a string, not a sim_mission

code/scripting/api/objs/techroom.cpp
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", current->name);
+}
+
+ADE_VIRTVAR(Filename, l_TechRoomMission, nullptr, "The filename of the mission", "sim_mission", "The filename")
```
BMagnu, line 23: same here

code/scripting/api/objs/techroom.cpp
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", current->filename);
+}
+
+ADE_VIRTVAR(Description, l_TechRoomMission, nullptr, "The mission description", "sim_mission", "The description")
```
BMagnu, line 37: same here

code/scripting/api/objs/techroom.cpp
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", current->mission_desc);
+}
+
+ADE_VIRTVAR(Author, l_TechRoomMission, nullptr, "The mission author", "sim_mission", "The author")
```
BMagnu, line 51: same here

code/scripting/api/objs/techroom.cpp
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", current->author);
+}
+
+ADE_VIRTVAR(Visibility, l_TechRoomMission, nullptr, "If the mission should be visible by default", "sim_mission", "1 if visible, 0 if not visible, returns nil if not a campaign mission")
```
BMagnu, line 65: this returns number not sim_mission

code/scripting/api/objs/techroom.cpp
```diff
@@ -0,0 +1,150 @@
...
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "i", current->visible);
+}
+
+//**********HANDLE: tech cutscenes
+ADE_OBJ(l_TechRoomCutscene, cutscene_info, "custscene_info", "Tech Room cutscene handle");
+
+ADE_VIRTVAR(Name, l_TechRoomCutscene, nullptr, "The name of the cutscene", "custscene_info", "The cutscene name")
```
BMagnu, line 82: this returns string again

code/scripting/api/objs/techroom.cpp
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", current->name);
+}
+
+ADE_VIRTVAR(Filename, l_TechRoomCutscene, nullptr, "The filename of the cutscene", "custscene_info", "The cutscene filename")
```
BMagnu, line 96: this too

code/scripting/api/objs/techroom.cpp
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", current->filename);
+}
+
+ADE_VIRTVAR(Description, l_TechRoomCutscene, nullptr, "The cutscene description", "custscene_info", "The cutscene description")
```
BMagnu, line 110: this too

code/scripting/api/objs/techroom.cpp
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", current->description);
+}
+
+ADE_VIRTVAR(Visibility,
```
BMagnu, line 124: this returns a number.
But since it only returns 0 and 1, should it not just be changed to return a boolean?

BMagnu (changes requested):

code/scripting/api/libs/ui.cpp
MjnMixael: Fixed

code/menuui/credits.h
MjnMixael: Fixed per our discussion

code/scripting/api/objs/techroom.cpp
MjnMixael: Good idea.

code/scripting/api/objs/techroom.h
MjnMixael: Done.

BMagnu (approved):

The-E (approved):

PR #4593, opened 2022-08-26
Fix Mac builds by overriding py7zr version

virusman:
Looks like the base GH Actions image got updated (newer Python?) and it now fails to install the older version of py7zr.


JohnAFernandez (approved):

BMagnu (approved):
This should work as a stopgap for now.

PR #4595, opened 2022-08-26
Some User Interface API cleanup and generalization

MjnMixael:
These are useful in places beyond just Command Brief.

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -126,7 +126,7 @@ ADE_VIRTVAR(DefaultTextColorTag,
 	}
 	
 	SCP_string tagStr;
-	char defaultColor;
+	char defaultColor = default_briefing_color;
 
 	switch (UiScreen) {
 		case 1:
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
I think this looks reasonable, but I'm curious about the reason for the move.  According to the code documentation, "a sublibrary is basically a nested namespace".  But if the API is `ui.runBriefingStageHook` under `l_UserInterface_CmdBrief`, then it should also be `ui.runBriefingStageHook` under `l_UserInterface_Brief`.  Is this correct?

MjnMixael:
The truth is I did not read into that one enough yet to know if it was C.Brief specific or general to all briefs. That's the only reason I did not move it yet.

Goober5000:
Oh, well I can tell you that `On Briefing Stage` is fired for both command briefings and regular briefings.  But what I mean is, what does the move actually do?  Does it affect the script API, or is it simply for code organization, or does it do a third thing?

MjnMixael:
Moving the items affects the scripts in that they'll be called by ui.Briefing.ACTION now instead of ui.CmdBrief.ACTION, so a slight affect there. However yes, it is mostly organizational.

Goober5000 (approved):
Ok, gotcha.  That all sounds good then.

PR #4596, opened 2022-08-26
Allow `test_line_of_sight` to return intersecting object signature

wookieejedi:
The `test_line_of_sight` can currently provide the closest intersecting distance, but it would be very useful, especially for AI scripting, if this function could also return the signature of the intersecting object. This PR adds that functionality.

Tested and works as expected. Happy to discuss strategies for changing implementation, too. The extra, third return value for object signature was added to `getLineOfSightFirstIntersect` function because that function already returns 2 values. Thus, it will remain compatible with older scripts which only save the first 2 returned values.

### DIFF
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1571,7 +1571,7 @@ ADE_FUNC(hasLineOfSight, l_Mission, "vector from, vector to, [table excludedObje
 	return testLineOfSight_internal(L, false);
 }
 
-ADE_FUNC(getLineOfSightFirstIntersect, l_Mission, "vector from, vector to, [table excludedObjects /* expects list of objects, empty by default */, boolean testForShields = false, boolean testForHull = true, number threshold = 10.0]", "Checks whether the to-position is in line of sight from the from-position and returns the distance and object signature to the first interruption of the line of sight, disregarding specific excluded objects and objects with a radius of less then threshold.", "boolean, number", "true and zero and nil if there is line of sight, false and the distance and Signature otherwise.")
+ADE_FUNC(getLineOfSightFirstIntersect, l_Mission, "vector from, vector to, [table excludedObjects /* expects list of objects, empty by default */, boolean testForShields = false, boolean testForHull = true, number threshold = 10.0]", "Checks whether the to-position is in line of sight from the from-position and returns the distance and object signature to the first interruption of the line of sight, disregarding specific excluded objects and objects with a radius of less then threshold.", "boolean, number", "number", "true and zero and -1 if there is line of sight, false and the distance and Signature otherwise.")
 {
 	return testLineOfSight_internal(L, true);
 }
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1571,7 +1571,7 @@ ADE_FUNC(hasLineOfSight, l_Mission, "vector from, vector to, [table excludedObje
 	return testLineOfSight_internal(L, false);
 }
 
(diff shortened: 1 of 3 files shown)


code/scripting/api/libs/mission.cpp
```diff
@@ -1569,7 +1571,7 @@ ADE_FUNC(hasLineOfSight, l_Mission, "vector from, vector to, [table excludedObje
...
 }
 
-ADE_FUNC(getLineOfSightFirstIntersect, l_Mission, "vector from, vector to, [table excludedObjects /* expects list of objects, empty by default */, boolean testForShields = false, boolean testForHull = true, number threshold = 10.0]", "Checks whether the to-position is in line of sight from the from-position and returns the distance to the first interruption of the line of sight, disregarding specific excluded objects and objects with a radius of less then threshold.", "boolean, number", "true and zero if there is line of sight, false and the distance otherwise.")
+ADE_FUNC(getLineOfSightFirstIntersect, l_Mission, "vector from, vector to, [table excludedObjects /* expects list of objects, empty by default */, boolean testForShields = false, boolean testForHull = true, number threshold = 10.0]", "Checks whether the to-position is in line of sight from the from-position and returns the distance and object signature to the first interruption of the line of sight, disregarding specific excluded objects and objects with a radius of less then threshold.", "boolean, number", "true and zero and nil if there is line of sight, false and the distance and Signature otherwise.")
```
BMagnu, line 1574: "boolean, number" needs to be changed to "boolean, number, number" here to reflect on the arguments returned.
Also, I am reasonably sure that you return -1 and not nil for the signature when no collision is observed. Not that that'd be bad, it's just not consistent with the documentation

BMagnu (changes requested):

code/scripting/api/libs/mission.cpp
wookieejedi: Thanks, just updated. And testing to double check, indeed no intersection is -1.

BMagnu (approved):

Baezon:
Why not simply return the object?

Goober5000 (changes requested):
Agreed, returning the actual object would save steps on both sides.

wookieejedi:
@Baezon and @Goober5000 Good idea, I've updated with object returning. Very immense thanks to @BMagnu for the pointer pointer help/code

code/scripting/api/libs/mission.cpp
```diff
@@ -1513,14 +1513,15 @@ ADE_FUNC(hasCommandBriefing,
 	return ade_set_args(L, "b", mission_has_cmd_brief() != 0);
 }
 
-int testLineOfSight_internal(lua_State* L, bool returnDist) {
+int testLineOfSight_internal(lua_State* L, bool returnDist_and_Sig) {
```
Goober5000, line 1516: `returnDist_and_Sig` should be renamed now

code/scripting/api/libs/mission.cpp
```diff
@@ -1569,7 +1570,7 @@ ADE_FUNC(hasLineOfSight, l_Mission, "vector from, vector to, [table excludedObje
...
+ADE_FUNC(getLineOfSightFirstIntersect, l_Mission, "vector from, vector to, [table excludedObjects /* expects list of objects, empty by default */, boolean testForShields = false, boolean testForHull = true, number threshold = 10.0]", "Checks whether the to-position is in line of sight from the from-position and returns the distance and object signature to the first interruption of the line of sight, disregarding specific excluded objects and objects with a radius of less then threshold.", "boolean, number, object", "true and zero and nil if there is line of sight, false and the distance and intersecting object otherwise.")
```
Goober5000, line 1573: "and object signature" should be "and intersecting object"

Goober5000 (changes requested):

code/scripting/api/libs/mission.cpp
wookieejedi: Thanks, just updated on both.

Goober5000 (approved):

PR #4597, opened 2022-08-26
Base method to determine Debug vs Release

MjnMixael:


Goober5000 (changes requested):
This is a great idea.  But we don't even need a variable here.  Just put this in the `inDebug` function:
```
#ifndef NDEBUG
return ADE_RETURN_TRUE;
#else
return ADE_RETURN_FALSE;
#endif
```

MjnMixael:
Just trying to follow existing conventions. Changing conventions on a whim makes learning the codebase an exercise in frustration. Changed.

code/scripting/api/libs/base.cpp
```diff
@@ -522,6 +522,14 @@ ADE_FUNC(inMissionEditor, l_Base, nullptr, "Determine if the current script is r
 	return ade_set_args(L, "b", Fred_running != 0);
 }
 
+ADE_FUNC(inDebug, l_Base, nullptr, "Determinds if FSO is running in Release or Debug", "boolean", "true if debug, false if release") {
```
Baezon, line 525:
```suggestion
ADE_FUNC(inDebug, l_Base, nullptr, "Determines if FSO is running in Release or Debug", "boolean", "true if debug, false if release") {
```

Goober5000 (approved):

Goober5000:
I can see how this would be confusing, but it's not a difference in convention -- it's an example of misinterpreting which convention to use.  This is something that is not always obvious and is usually learned over time, as one gets more familiar with the code.

PR #4598, opened 2022-08-27
Some extra safety checks for fiction viewer scripts

wookieejedi:
Protects `Fiction_viewer_active_stage` if it reads a stage of -1 (it would have thrown a vector access error because the index is out of range). Also, given a lot of other scripting functions have additional built in error checks, I just added those to the fiction viewer functions.


Goober5000 (approved):
Good changes. 👍

PR #4599, opened 2022-08-28
Implemented a command line flag that forces a fixed RNG seed,

EatThePath:
Reseeds the rng every mission load, if not in MP.

~~In draft because I'm having compiler difficulties and haven't been able to test any.~~

Now tested and appears functional. Doesn't result in noticeable repeatability in chaotic cutscenes, the ultimate goal, but that's expected as this is only one small piece of the puzzle.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -476,6 +476,8 @@ bool Cmdline_dump_packet_type = false;
 #ifdef WIN32
 bool Cmdline_alternate_registry_path = false;
 #endif
+uint Cmdline_randseed = 0;
+bool Cmdline_randfixed = false;
 
 // Developer/Testing related
 cmdline_parm start_mission_arg("-start_mission", "Skip mainhall and run this mission", AT_STRING);	// Cmdline_start_mission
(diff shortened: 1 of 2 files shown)


Goober5000 (changes requested):
Can we use a different term than "fixed"?  That's a word with multiple meanings.

You could save a variable by just testing that the seed is greater than zero.  On that note, your current code allows a seed of zero, which is disallowed according to the implementation of `Random::seed`.

EatThePath:
> Can we use a different term than "fixed"? That's a word with multiple meanings.

I assume the issue is potentional confusion with fix as in 'bugfix'? 

Fixed is the only term I can really think of being used regularly for this kind of thing. Do you have a preferred alternative? Everything else that comes to mind for me also has potential for confusion too.

Goober5000:
Yeah, I was thinking about potential confusion with "bugfix".  It's hard to think of an alternative, but consulting a thesaurus:
* latched
* clamped
* locked
* pinned
* uniform

"Latched" has engineering heritage, while "pinned" has web heritage.

Goober5000 (approved):
👍

PR #4602, opened 2022-08-29
SCPUI API elements required for the Briefing

MjnMixael:
This is everything required but the icon map itself, though it does include a placeholder API function for my own sanity and the sanity of the scripts until I can spend time getting it working.

### DIFF
--- a/code/scpui/elements/ScrollingTextElement.cpp
+++ b/code/scpui/elements/ScrollingTextElement.cpp
@@ -56,7 +56,8 @@ void ScrollingTextElement::OnChildRemove(Element* child)
 void ScrollingTextElement::OnUpdate()
 {
 	if (_newAnimationStarted) {
-		gamesnd_play_iface(InterfaceSounds::BRIEF_TEXT_WIPE);
+		//Disabled because this should be controled by the Lua to prevent it playing on frame 0
+		//gamesnd_play_iface(InterfaceSounds::BRIEF_TEXT_WIPE);
 		_newAnimationStarted = false;
 	}
 }
--- a/code/scpui/elements/ScrollingTextElement.cpp
+++ b/code/scpui/elements/ScrollingTextElement.cpp
@@ -52,10 +52,6 @@ void ScrollingTextElement::OnChildRemove(Element* child)
 	_animation_start_time = Rocket::Core::GetSystemInterface()->GetElapsedTime();
 }
 
-void ScrollingTextElement::OnUpdate() {
-	// This method originally existed to play a sound effect which has been removed
-	// in favor of doing it manually. -Mjn
-}
 void ScrollingTextElement::OnBeforeRender()
 {
 	if (_animation_start_time >= 0.0f) {
--- a/code/scpui/elements/ScrollingTextElement.cpp
+++ b/code/scpui/elements/ScrollingTextElement.cpp
@@ -51,7 +51,6 @@ void ScrollingTextElement::OnChildRemove(Element* child)
 	Element::OnChildRemove(child);
 	_animation_start_time = Rocket::Core::GetSystemInterface()->GetElapsedTime();
 }
-
 void ScrollingTextElement::OnBeforeRender()
 {
 	if (_animation_start_time >= 0.0f) {
--- a/code/scpui/elements/ScrollingTextElement.cpp
+++ b/code/scpui/elements/ScrollingTextElement.cpp
@@ -45,21 +45,16 @@ void ScrollingTextElement::OnChildAdd(Element* child)
 {
 	Element::OnChildAdd(child);
 	_animation_start_time = Rocket::Core::GetSystemInterface()->GetElapsedTime();
-	_newAnimationStarted = true;
 }
 void ScrollingTextElement::OnChildRemove(Element* child)
 {
 	Element::OnChildRemove(child);
 	_animation_start_time = Rocket::Core::GetSystemInterface()->GetElapsedTime();
-	_newAnimationStarted = true;
 }
-void ScrollingTextElement::OnUpdate()
-{
-	if (_newAnimationStarted) {
-		//Disabled because this should be controled by the Lua to prevent it playing on frame 0
-		//gamesnd_play_iface(InterfaceSounds::BRIEF_TEXT_WIPE);
-		_newAnimationStarted = false;
-	}
+
+void ScrollingTextElement::OnUpdate() {
+	// This method originally existed to play a sound effect which has been removed
+	// in favor of doing it manually. -Mjn
 }
 void ScrollingTextElement::OnBeforeRender()
 {
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -21,6 +21,7 @@
 #include "mission/missionlog.h"
 #include "mission/missionmessage.h"
 #include "mission/missiontraining.h"
+#include "missionui/missionbrief.h"
 #include "missionui/missioncmdbrief.h"
 #include "missionui/redalert.h"
 #include "nebula/neb.h"
@@ -1307,6 +1308,17 @@ ADE_FUNC(isTraining, l_Mission, nullptr, "Get whether or not the current mission
 	bool b = false;
 
 	if (The_mission.game_type & MISSION_TYPE_TRAINING) {
+		b = true;
+	}
+
+	return ade_set_args(L, "b", b);
+}
+
+ADE_FUNC(isScramble, l_Mission, nullptr, "Get whether or not the current mission being played is a scramble mission", "boolean", "true if in training, false if not")
+{
+	bool b = false;
+
+	if (brief_only_allow_briefing()) {
 		b = true;
 	}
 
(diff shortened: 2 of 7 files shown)


code/scripting/api/objs/briefing.cpp
```diff
@@ -0,0 +1,198 @@
+
+#include "briefing.h"
+
+namespace scripting {
+namespace api {
+
+//**********HANDLE: briefing
+ADE_OBJ(l_BriefStage, brief_stage, "briefing_stage", "Briefing stage handle");
+
+ADE_VIRTVAR(Text, l_BriefStage, nullptr, "The text of the stage", "briefing_stage", "The text")
```
BMagnu, line 10: The fifth argument here is the return type and should be "string" and not "briefing_stage"

code/scripting/api/objs/briefing.cpp
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return ade_set_args(L, "s", stage->text);
+}
+
+ADE_VIRTVAR(AudioFilename,
+	l_BriefStage,
+	nullptr,
+	"The filename of the audio file to play",
+	"briefing_stage",
```
BMagnu, line 28: same here, needs to be "string"

code/scripting/api/objs/briefing.cpp
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return ade_set_args(L, "s", stage->voice);
+}
+
+ADE_VIRTVAR(isVisible,
+		l_BriefStage,
+		nullptr,
+		"The result of the stage formula",
+		"briefing_stage",
```
BMagnu, line 47: boolean, not briefing_stage

code/scripting/api/objs/briefing.h
```diff
@@ -0,0 +1,17 @@
...
+#include "scripting/ade_api.h"
+
+namespace scripting {
+namespace api {
+
+DECLARE_ADE_OBJ(l_BriefStage, brief_stage);
+
+DECLARE_ADE_OBJ(l_Brief, briefing);
+
+DECLARE_ADE_OBJ(l_Goals, mission_goal);
```
BMagnu, line 14: This is an issue.
We really don't want to create lua objects based on complex types.
This creates copies in lua, and will cause multiple issues, such as
1. Being slow
2. Any changes will not be propagated back to FSO
3. The object won't pick up changes made in FSO

To fix this, we need to make some new object types. l_Brief can just use int instead of the briefing class, and store only the briefing index. See for example the team.h/.cpp API classes as to how that's done.
l_BriefStage should IMO take a class that contains a brief_stage pointer. Since neither brief stages nor briefings are in dynamically allocated memory, that should be safe. An example how this needs to look is waypointlist_h in waypoint.h (just without the name business for the briefing stage)
l_Goals can be an int again, like l_Brief.

code/scripting/api/objs/briefing.cpp
```diff
@@ -0,0 +1,198 @@
...
+		return ADE_RETURN_NIL;
+	}
+
+	return ade_set_args(L, "i", brief->num_stages);
+}
+
+//**********HANDLE: mission goals
+ADE_OBJ(l_Goals, mission_goal, "mission_goal", "Mission objective handle");
+
+ADE_VIRTVAR(Name, l_Goals, nullptr, "The name of the goal", "mission_goal", "The goal name")
```
BMagnu, line 104: This needs "string" instead of "mission_goal" again

code/scripting/api/objs/briefing.cpp
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", current->name);
+}
+
+ADE_VIRTVAR(Message, l_Goals, nullptr, "The message of the goal", "mission_goal", "The goal message")
```
BMagnu, line 118: string again

code/scripting/api/objs/briefing.cpp
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", current->message);
+}
+
+ADE_VIRTVAR(Type, l_Goals, nullptr, "The goal type", "mission_goal", "primary, secondary, bonus, or none")
```
BMagnu, line 132: string again

code/scripting/api/objs/briefing.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+	default:
+		type = "none";
+		break;
+	}
+
+	return ade_set_args(L, "s", type);
+}
+
+ADE_VIRTVAR(Team, l_Goals, nullptr, "The goal team", "mission_goal", "The goal team")
```
BMagnu, line 167: Here, I'd return a team object instead of an index, by returning ``return ade_set_args(L, "o", l_Team.Set(idx));``.
Also, this means that the documentation must specify that it returns "team" and not "mission_goal"

code/scripting/api/objs/briefing.cpp
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "i", current->team);
+}
+
+ADE_VIRTVAR(isGoalValid, l_Goals, nullptr, "The goal validity", "mission_goal", "true if valid, false otherwise")
```
BMagnu, line 181: boolean, not mission_goal

BMagnu (changes requested):

code/scripting/api/objs/briefing.h
MjnMixael: Can you clarify? I duplicated exactly what m!m did for Cmd Briefs. These don't ever change during runtime.

code/scripting/api/objs/briefing.cpp
MjnMixael: Another instance where I duplicated what was done in cmd_brief.cpp. However, this one I'm understanding. Looks like cmd_brief.cpp will need a slight cleanup to match. I'll add it to me to-do list for the API.

code/scripting/api/libs/mission.cpp
```diff
@@ -1290,6 +1292,61 @@ ADE_FUNC(isInCampaign, l_Mission, NULL, "Get whether or not the current mission
...
+{
+	bool b = false;
+
+	if (The_mission.flags[Mission::Mission_Flags::No_briefing]) {
+		b = true;
+	}
+
+	return ade_set_args(L, "b", b);
+}
+
```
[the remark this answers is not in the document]
MjnMixael: Can do for all of these except the Training check because that expression returns an int.

code/scpui/elements/ScrollingTextElement.cpp
```diff
@@ -56,7 +56,8 @@ void ScrollingTextElement::OnChildRemove(Element* child)
 void ScrollingTextElement::OnUpdate()
 {
 	if (_newAnimationStarted) {
-		gamesnd_play_iface(InterfaceSounds::BRIEF_TEXT_WIPE);
+		//Disabled because this should be controled by the Lua to prevent it playing on frame 0
+		//gamesnd_play_iface(InterfaceSounds::BRIEF_TEXT_WIPE);
```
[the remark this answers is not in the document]
MjnMixael: Ah I see. This will leave an empty method behind because it seems like librocket expects OnUpdate to exist.

code/scripting/api/objs/briefing.cpp
BMagnu: Yes, it'll likely also need changing in cmd_brief
Must have slipped though

code/scpui/elements/ScrollingTextElement.cpp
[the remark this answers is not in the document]
BMagnu: If you delete the method both from ScrollingTextElement.h and .cpp, it should be fine. But IMO more important is that you delete _newAnimationStarted in ScrollingTextElement.h and all its mentions in the cpp, since they aren't needed anymore

code/scripting/api/objs/briefing.h
BMagnu: Right, and I think that cmd_briefs need to be changed as well.
What happens here, is that every time you get a briefing stage from lua, it will make a _copy_ of the briefing stage. A copy which is unrelated to the one from FSO, afterwards. As long as you are guaranteed that the Lua API will never change anything in that copy, _and_ that FSO wil never modify the briefing stage while you have your copy in lua, that is fine-ish. But if you can't guarantee that, it becomes a problem. This also implies, that you must pay attention that no one ever implements anything which might be able to modify briefings from Lua, which has reasonable use cases IMO.
I will make a PR later today which'll fix this and the documentation strings for cmd_briefs, then you can see what'd need to be changed.

BMagnu: See #4646

MjnMixael: Ok, excellent. That gives me an example to follow. I'll see if I can figure this out.

MjnMixael: Ok, I think I've set this up correctly and my tests show it still works as expected.

BMagnu (approved):
LGTM now

PR #4603, opened 2022-08-29
more timestamp migration

Goober5000:
More conversion of timestamps to the new system.  Follow-up to https://github.com/scp-fs2open/fs2open.github.com/pull/4569.

### DIFF
--- a/code/actions/common.h
+++ b/code/actions/common.h
@@ -18,7 +18,7 @@ struct ProgramLocals {
 	matrix localOrient = vmd_identity_matrix;
 	int hostSubobject = -1;
 
-	int waitTimestamp = 0;
+	TIMESTAMP waitTimestamp = TIMESTAMP::invalid();
 
 	expression::ProgramVariables variables;
 };
--- a/code/actions/types/WaitAction.cpp
+++ b/code/actions/types/WaitAction.cpp
@@ -20,7 +20,7 @@ WaitAction::~WaitAction() = default;
 
 ActionResult WaitAction::execute(actions::ProgramLocals& locals) const
 {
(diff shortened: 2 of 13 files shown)


code/actions/common.h
```diff
@@ -18,7 +18,7 @@ struct ProgramLocals {
 	matrix localOrient = vmd_identity_matrix;
 	int hostSubobject = -1;
 
-	int waitTimestamp = 0;
+	TIMESTAMP waitTimestamp = TIMESTAMP::invalid();
```
JohnAFernandez, line 21: Looks good just one question.  -1 invalid is the legacy equivalent to invalid or is it <= 0? I see that you adjust the places where it's checked so that the logic still works the same. But I think you should add a comment where valid, never, and immediate are defined with the old values so that if someone is writing with the new functions, they can easily relate them to the old functions they are used to.

Goober5000: There is no clean correspondence with the old values.  Both 0 and -1 are used to mean both "invalid" and "never" in different parts of the code.  That's one of the motivations for the new timestamp system in the first place.  The conversion to the new timestamp system is necessarily going to involve examining the old code for context clues.

JohnAFernandez: Ok, but 1 still means immediate, and that explanation would be helpful.  Somewhere in the definitions.

Goober5000: Isn't this self-explanatory?
```
inline bool isImmediate() const { return m_val == 1; }
```

Goober5000: Ok, see new commit.

JohnAFernandez (approved):
That was exactly what I was hoping for. I really do appreciate the extra effort!

PR #4605, opened 2022-08-30
Associate only a single pof version to submodel translation

Baezon:
While a form of 'backwards compatibility' by including some features but not others might seem helpful, having new features 'interleaved' through the versions *greatly* complicates version checking (note how the check went from 5 comparisons to 1), is practically not very useful, and this situation will worsen with every feature added this way. New features should be placed in the newest version, like most other software.


Goober5000 (approved):
Very good points.  Agreed.

PR #4607, opened 2022-08-31
Allow mods to change Emp flash color, intensity, and string

wookieejedi:
EMP is the only text warning that does not use an entry in the strings table (and thus it's the only text warning gauge that mods cannot change and/or translate). This PR adds a string entry for it to allow mods to translate and change the string like all other HUD gauge text warnings.

Also, EMP flash has a hard coded color and strength factor for the flash, whereas generic pain and shield pain have a configurable flash strength factor (shield pain flash is based on shield color, while generic pain is based on damage dealt). This PR also allows modders to change the strength intensity and color following previously established methods.

This PR also fixes a case of "Launch" that fell through the cracks. The function is supplied the translated string, but within the function it does not check for the translation.

Everything is tested and works as expected.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -382,19 +382,19 @@ void parse_mod_table(const char *filename)
 
 		if (optional_string("$Generic Pain Flash Factor:")) {
 			stuff_float(&Generic_pain_flash_factor);
-			if (!fl_equal(Generic_pain_flash_factor, 1.0f))
+			if (Generic_pain_flash_factor > 0.0f)
 				mprintf(("Game Settings Table: Setting generic pain flash factor to %.2f\n", Generic_pain_flash_factor));
 		}
 
 		if (optional_string("$Shield Pain Flash Factor:")) {
 			stuff_float(&Shield_pain_flash_factor);
-			if (!fl_near_zero(Shield_pain_flash_factor, 0.01f))
+			if (Shield_pain_flash_factor > 0.0f)
 				 mprintf(("Game Settings Table: Setting shield pain flash factor to %.2f\n", Shield_pain_flash_factor));
 		}
 
 		if (optional_string("$EMP Pain Flash Factor:")) {
 			stuff_float(&Emp_pain_flash_factor);
-			if (!fl_near_zero(Emp_pain_flash_factor, 0.01f))
+			if (Emp_pain_flash_factor > 0.0f)
 				mprintf(("Game Settings Table: Setting EMP pain flash factor to %.2f\n", Emp_pain_flash_factor));
 		}
 
@@ -402,7 +402,7 @@ void parse_mod_table(const char *filename)
 			int rgb[3];
 			stuff_int_list(rgb, 3);
 			if ((rgb[0] >= 0 && rgb[0] <= 255) && (rgb[1] >= 0 && rgb[1] <= 255) && (rgb[2] >= 0 && rgb[2] <= 255)) {
-				Emp_pain_flash_color = std::make_tuple(static_cast<float>(rgb[0]/255), static_cast<float>(rgb[1]/255), static_cast<float>(rgb[2]/255));
+				Emp_pain_flash_color = std::make_tuple(static_cast<float>(rgb[0])/255, static_cast<float>(rgb[1])/255, static_cast<float>(rgb[2])/255);
 			} else {
 				error_display(0, "$EMP Pain Flash Color is %i, %i, %i. "
 					"One or more of these values is not within the range of 0-255. Assuming default color.", rgb[0], rgb[1], rgb[2]);
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -381,21 +381,30 @@ void parse_mod_table(const char *filename)
 		}
 
 		if (optional_string("$Generic Pain Flash Factor:")) {
-			stuff_float(&Generic_pain_flash_factor);
-			if (Generic_pain_flash_factor > 0.0f)
-				mprintf(("Game Settings Table: Setting generic pain flash factor to %.2f\n", Generic_pain_flash_factor));
+			float temp;
+			stuff_float(&temp);
+			if (temp >= 0.0f) {
+				mprintf(("Game Settings Table: Setting generic pain flash factor to %.2f\n", temp));
+				Generic_pain_flash_factor = temp;
+			}
 		}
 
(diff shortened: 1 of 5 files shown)


code/weapon/emp.cpp
```diff
@@ -407,13 +407,17 @@ void emp_start_local(float intensity, float time)
...
 	// determine how much we have to decrement the effect per second
 	Emp_decr = Emp_intensity / time;
 
 	// play a flash
-	game_flash( 1.0f, 1.0f, 0.5f );
+	game_flash(
+		std::get<0>(Emp_pain_flash_color) * Emp_pain_flash_factor,
+		std::get<1>(Emp_pain_flash_color) * Emp_pain_flash_factor,
+		std::get<2>(Emp_pain_flash_color) * Emp_pain_flash_factor
+	);
```
Goober5000, line 420: I don't think this actually works as coded.  The default value is 255,255,125 multiplied by 1.0, but the function expects values from 0 to 1.  This will get clamped to 1,1,1.

Goober5000 (changes requested):
Couple things.

code/mod_table/mod_table.cpp
```diff
@@ -390,6 +392,23 @@ void parse_mod_table(const char *filename)
 				 mprintf(("Game Settings Table: Setting shield pain flash factor to %.2f\n", Shield_pain_flash_factor));
 		}
 
+		if (optional_string("$EMP Pain Flash Factor:")) {
+			stuff_float(&Emp_pain_flash_factor);
+			if (Emp_pain_flash_factor != 0.0f)
```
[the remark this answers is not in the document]
wookieejedi: Could do that, definitely. I was following what was precedented above, so if this changes should I then change the preceding two similar checks in Generic and Shield pain flash factors?

code/mod_table/mod_table.cpp
```diff
@@ -943,6 +962,8 @@ void mod_table_reset()
 	Beams_use_damage_factors = false;
 	Generic_pain_flash_factor = 1.0f;
 	Shield_pain_flash_factor = 0.0f;
+	Emp_pain_flash_factor = 1.0f;
+	Emp_pain_flash_color = std::make_tuple(static_cast<ubyte>(255), static_cast<ubyte>(255), static_cast<ubyte>(125));
```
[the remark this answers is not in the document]
wookieejedi: Good point, will change

code/mod_table/mod_table.cpp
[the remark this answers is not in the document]
Goober5000: Sure, that would keep them consistent.

code/weapon/emp.cpp
wookieejedi: Good point, changed.

code/mod_table/mod_table.cpp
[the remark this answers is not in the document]
wookieejedi: Ah forgot to respond to this, all have been updated to be consistent using `fl_near_zero`.

code/mod_table/mod_table.cpp
```diff
@@ -382,19 +382,19 @@ void parse_mod_table(const char *filename)
 
 		if (optional_string("$Generic Pain Flash Factor:")) {
 			stuff_float(&Generic_pain_flash_factor);
-			if (Generic_pain_flash_factor != 1.0f)
```
Goober5000, line 383: I just noticed that this isn't being compared to 0, it's being compared to 1.  So it will need `fl_equal` instead.

code/mod_table/mod_table.cpp
```diff
@@ -943,6 +962,8 @@ void mod_table_reset()
 	Beams_use_damage_factors = false;
 	Generic_pain_flash_factor = 1.0f;
 	Shield_pain_flash_factor = 0.0f;
+	Emp_pain_flash_factor = 1.0f;
+	Emp_pain_flash_color = std::make_tuple(static_cast<ubyte>(255), static_cast<ubyte>(255), static_cast<ubyte>(127));
```
Goober5000, line 966: You might want to do a small optimization here and make these all floats from 0 to 1.  Since that is their use case, it makes sense to convert them when they are initialized and parsed, not every time `game_flash` is called.

Goober5000 (changes requested):

code/mod_table/mod_table.cpp
wookieejedi: Good catch, just changed.

code/mod_table/mod_table.cpp
wookieejedi: Good idea, changed to floats instead of ubtyes.

code/mod_table/mod_table.cpp
```diff
@@ -380,16 +382,33 @@ void parse_mod_table(const char *filename)
 
 		if (optional_string("$Generic Pain Flash Factor:")) {
 			stuff_float(&Generic_pain_flash_factor);
-			if (Generic_pain_flash_factor != 1.0f)
+			if (!fl_equal(Generic_pain_flash_factor, 1.0f))
```
Goober5000, line 385: Come to think of it, why is this being compared to 1 anyway?  The shield pain and EMP pain flash factors are only set if the factor is nonzero.

(And, bleh, the parser should make sure all three are positive too.)

code/mod_table/mod_table.cpp
```diff
@@ -380,16 +382,33 @@ void parse_mod_table(const char *filename)
...
+			stuff_float(&Emp_pain_flash_factor);
+			if (!fl_near_zero(Emp_pain_flash_factor, 0.01f))
+				mprintf(("Game Settings Table: Setting EMP pain flash factor to %.2f\n", Emp_pain_flash_factor));
+		}
+
+		if (optional_string("$EMP Pain Flash Color:")) {
+			int rgb[3];
+			stuff_int_list(rgb, 3);
+			if ((rgb[0] >= 0 && rgb[0] <= 255) && (rgb[1] >= 0 && rgb[1] <= 255) && (rgb[2] >= 0 && rgb[2] <= 255)) {
+				Emp_pain_flash_color = std::make_tuple(static_cast<float>(rgb[0]/255), static_cast<float>(rgb[1]/255), static_cast<float>(rgb[2]/255));
```
Goober5000, line 405: The cast to float should be done before the division, otherwise you'll be dividing an int by an int which in this case will give you only 0 or 1.

Goober5000 (approved):

Goober5000:
Looks good now!

PR #4608, opened 2022-09-01
Allow assigning support ship per species

### LINKED ISSUE #3352: FR: Support Ship <-> Species Improvements
Support ships are strange. Any ship will use any same-team support ship currently in the field regardless of species. However, if a species has no support ship defined in ships.tbl, then all ships of that species cannot call for support even from same-team support ships.

My proposed solution is an entry in species_defs, 
`$Species Support Ship: Ship Class`

But there may be better solutions as I'm unsure what assumptions the codebase makes regarding support ships/species and how they interact.

The main issue I'm trying to avoid here is having multiple duplicate ships in ships.tbl that I have to remember to update every time I make adjustments to that ship class. So consider this a modding QoL request.

naomimyselfandi:
For when you want to share a support ship between species. For BTA, but hey, maybe someone else wants to use species for organization like we do.

Closes #3352

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -616,7 +616,7 @@ void parse_mission_info(mission *pm, bool basic = false)
 
 	// for each species, store whether support is available
 	for (int species = 0; species < (int)Species_info.size(); species++) {
-		if (Species_info[species].support_ship_index > 0) {
+		if (Species_info[species].support_ship_index >= 0) {
 			pm->support_ships.support_available_for_species |= (1 << species);
 		}
 	}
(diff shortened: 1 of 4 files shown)


code/mission/missionparse.cpp
```diff
@@ -8131,31 +8126,17 @@ void mission_bring_in_support_ship( object *requester_objp )
...
-		if ( i != -1 )
-			pobj->ship_class = i;
-		else
+	if (pobj->ship_class < 0) {
+		int requester_species = Ship_info[requester_shipp->ship_info_index].species;
+		int ship_class = Species_info[requester_species].support_ship_index;
+		if ( ship_class >= 0 ) {
+			pobj->ship_class = ship_class;
+		} else {
 			Int3();				// BOGUS!!!!  gotta figure something out here
```
Goober5000, line 8138: While you're here, I suggest replacing this `Int3()` with a `Warning()`.

code/mission/missionparse.cpp
```diff
@@ -614,15 +615,9 @@ void parse_mission_info(mission *pm, bool basic = false)
...
-	{
-        for (auto it = Ship_info.cbegin(); it != Ship_info.cend(); ++it)
-		{
-			if ((it->flags[Ship::Info_Flags::Support]) && (it->species == species))
-			{
-				pm->support_ships.support_available_for_species |= (1 << species);
-				break;
-			}
+	for (int species = 0; species < (int)Species_info.size(); species++) {
+		if (Species_info[species].support_ship_index > 0) {
```
Goober5000, line 619: This >0 should be >=0.

Goober5000 (changes requested):
Nicely done!  Two comments.

code/mission/missionparse.cpp
naomimyselfandi: Done.

code/mission/missionparse.cpp
naomimyselfandi: Fixed.

Goober5000 (approved):
Ah, that's even better.

code/mission/missionparse.cpp
Goober5000: Ah, that's even better. 👍

PR #4609, opened 2022-09-01
Respect hashes in ship and wing names in the loadout screen

naomimyselfandi:
Exactly what it says on the tin. For mission design trickery.


Goober5000 (changes requested):
For this block of code, instead of truncating the name, I recommend replacing `sp->ship_name` with `sp->get_display_name()`.  This will not only be consistent with display names used elsewhere, it will also handle the hash truncation.  (If a ship has a truncated name, it will now be assigned as the default display name.)

EDIT: Apparently Github didn't like my code selection.  I selected the second half of the diff, starting with `bool trim = true;`.

naomimyselfandi:
Now using `get_display_name()`.

Goober5000 (approved):
👍

PR #4610, opened 2022-09-01
handle irc motd error to prevent connection hang

notimaginative:
A MOTD missing (422) message is non-fatal and should be handled the same as a MOTD end (376) message to prevent the chat connection process from hanging.


notimaginative:
Also removes extra motd line check (code 372) that did nothing but waste time.

Goober5000 (approved):

PR #4613, opened 2022-09-01
Add interpolate matrices for multi

JohnAFernandez:
This adds interpolation of matrices via axis angle calculations, and a way to calculate last orient from angular velocity.  The physics code is from @Baezon, with just a few edits.  We've tested and everything works as expected.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -2955,25 +2955,20 @@ void vm_interpolate_angles_quick(angles *dest0, angles *src0, angles *src1, floa
 }
 
 // Interpolate between two matrices, using t as a percentage progress between them.
+// Intended values for t are [0.0f, 1.0f], but only values close to zero rejected, 
+// as you could conceivably use these calculations to find a rotation that is more 
+// than 100% of the rotation.
 // derived by Asteroth from our AI code
-void vm_interpolate_matrices(matrix* out_orient, matrix* curr_orient, matrix* goal_orient, float t) 
+void vm_interpolate_matrices(matrix* out_orient, const matrix* curr_orient, const matrix* goal_orient, float t) 
 {
-	// check the two cases where it doesn't make sense to go through the whole function
+	// check the case where it doesn't make sense to go through the whole function
 	if (fl_near_zero(t)) {
 		*out_orient = *curr_orient;
 		return;
 	}
 
-	Assertion(t <= 1.0f, "The function vm_interpolate_matrices was passed a bad percentage of %f.  This is an error that points to a coder mistake.  Please report to the SCP!",t);
-
-	// any value > 1.0f should be a bug.
-	if (t >= 1.0f) {
-		*out_orient = *goal_orient;
-		return;
-	}
-
 	matrix Mtemp1;
- 
+
 	vm_copy_transpose(&Mtemp1, curr_orient);     // Mtemp1 = curr ^-1
  
 	matrix matrix_delta;        // rotation matrix from curr_orient to goal_orient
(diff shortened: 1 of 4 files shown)


code/math/vecmat.cpp
```diff
@@ -2954,6 +2954,54 @@ void vm_interpolate_angles_quick(angles *dest0, angles *src0, angles *src1, floa
...
+	// check the two cases where it doesn't make sense to go through the whole function
+	if (fl_near_zero(t)) {
+		*out_orient = *curr_orient;
+		return;
+	}
+
+	Assertion(t <= 1.0f, "The function vm_interpolate_matrices was passed a bad percentage of %f.  This is an error that points to a coder mistake.  Please report to the SCP!",t);
+
+	// any value > 1.0f should be a bug.
+	if (t >= 1.0f) {
```
Baezon, line 2970: I would handle this in the caller's context, t out of 0..1 is odd, but perfectly mathematically sensical, and there might be contexts where someone might want it.

Baezon (changes requested):

code/math/vecmat.cpp
JohnAFernandez: Okay.  Lucky I already do check it outside the funciton.

Baezon (approved):

code/math/vecmat.cpp
```diff
@@ -2954,6 +2954,46 @@ void vm_interpolate_angles_quick(angles *dest0, angles *src0, angles *src1, floa
 	dest0->b = src0->b + (arc_measures.b * interp_perc);
 }
 
+// Interpolate between two matrices, using t as a percentage progress between them.
+// derived by Asteroth from our AI code
+void vm_interpolate_matrices(matrix* out_orient, matrix* curr_orient, matrix* goal_orient, float t)
```
Goober5000, line 2959: I recommend making the second and third parameters `const matrix*`

code/math/vecmat.cpp
```diff
@@ -2954,6 +2954,46 @@ void vm_interpolate_angles_quick(angles *dest0, angles *src0, angles *src1, floa
 	dest0->b = src0->b + (arc_measures.b * interp_perc);
 }
 
+// Interpolate between two matrices, using t as a percentage progress between them.
+// derived by Asteroth from our AI code
+void vm_interpolate_matrices(matrix* out_orient, matrix* curr_orient, matrix* goal_orient, float t) 
+{
+	// check the two cases where it doesn't make sense to go through the whole function
```
Goober5000, line 2961: This is only one case?  Unless this is also meant to refer to the check on line 2982

code/math/vecmat.cpp
```diff
@@ -2954,6 +2954,46 @@ void vm_interpolate_angles_quick(angles *dest0, angles *src0, angles *src1, floa
 	dest0->b = src0->b + (arc_measures.b * interp_perc);
 }
 
+// Interpolate between two matrices, using t as a percentage progress between them.
```
Goober5000, line 2957: Since percentage generally means 0..100, I would either explicitly say 0..1, or else say "fraction" instead of "percentage"

code/math/vecmat.h
```diff
@@ -588,6 +588,10 @@ void vm_match_bank(vec3d* out_rvec, const vec3d* goal_fvec, const matrix* match_
 // You will get strange results otherwise.
 void vm_interpolate_angles_quick(angles* dest0, angles* src0, angles* src1, float interp_perc);
 
+// Interpolate between two matrices, using t as a percentage between them.
+// derived by Asteroth from our AI code
+void vm_interpolate_matrices(matrix* out_orient, matrix* curr_orient, matrix* goal_orient, float t);
```
Goober5000, line 593: Changes to the header and comments in the cpp should be replicated here

Goober5000 (changes requested):
A few comments for polishing

code/math/vecmat.cpp
```diff
@@ -2954,6 +2954,48 @@ void vm_interpolate_angles_quick(angles *dest0, angles *src0, angles *src1, floa
...
+// Intended values for t are [0.0f, 1.0f], but 1.0f is not checked, as you could conceivably
```
Goober5000, line 2958: "1.0f is not checked" might imply that 1.0 doesn't actually work.  Instead, say that `t` is not restricted to that range.  Incidentally, negative values are also mathematically valid -- for the same reason as values >1.

Goober5000 (changes requested):

Goober5000 (approved):
Ok, I'll take that.

JohnAFernandez:
I did a merge commit so that the matrix interpolation function would remain a separate commit.

PR #4615, opened 2022-09-02
Scpui API debriefing library

MjnMixael:
Required elements for the Debriefing API. Should be everything required and is working via testing. Unfortunately because of how this particular UI works I have to init and then close the whole thing so it's worth checking that everything is done properly.


code/scripting/api/libs/mission.cpp
```diff
@@ -1513,6 +1513,21 @@ ADE_FUNC(hasCommandBriefing,
 	return ade_set_args(L, "b", mission_has_cmd_brief() != 0);
 }
 
+ADE_FUNC(hasDebriefing,
+	l_Mission,
+	nullptr,
+	"Determines if the current mission has a command briefing",
```
Goober5000, line 1519: "command briefing" instead of "debriefing"

code/scripting/api/libs/mission.cpp
```diff
@@ -1513,6 +1513,21 @@ ADE_FUNC(hasCommandBriefing,
...
+	"boolean",
+	"true if command briefing, false otherwise.")
```
Goober5000, line 1521: and here

code/scripting/api/libs/mission.cpp
```diff
@@ -1513,6 +1513,21 @@ ADE_FUNC(hasCommandBriefing,
...
+{
+	bool check = false;
+	if (The_mission.flags[Mission::Mission_Flags::Toggle_debriefing]) {
+		check = true;
```
Goober5000, line 1525: Shouldn't this check default to `true`?  Most missions have debriefings.

code/scripting/api/libs/ui.cpp
```diff
@@ -542,6 +556,225 @@ ADE_FUNC(getCmdBriefing,
...
+
+	if (Subspace_ambient_right_channel.isValid()) {
+		snd_stop(Subspace_ambient_right_channel);
+		Subspace_ambient_right_channel = sound_handle::invalid();
+	}
+	snd_stop(Radar_static_looping);
+	Radar_static_looping = sound_handle::invalid();
+	snd_stop(Target_static_looping);
+	shipfx_stop_engine_wash_sound();
+	Target_static_looping = sound_handle::invalid();
```
Goober5000, line 596: These lines should all be replaced with a call to `game_stop_looped_sounds()`.

code/scripting/api/libs/ui.cpp
```diff
@@ -542,6 +556,225 @@ ADE_FUNC(getCmdBriefing,
...
+	"API for accessing data related to the debriefing UI.<br><b>Warning:</b> This is an internal "
+	"API for the new UI system. This should not be used by other code and may be removed in the future!",
+	l_UserInterface);
+
+ADE_FUNC(initDebriefing,
+	l_UserInterface_Debrief,
+	nullptr,
+	"Builds the debriefing, the stats, sets the next campaign mission, and makes all relevant data accessible",
+	"number",
+	"Returns 1 when completed")
```
Goober5000, line 572: It would be better to return `true` when completed, unless there are other possible return values.  Alternatively, the return value may not be needed at all.

code/scripting/api/libs/ui.cpp
```diff
@@ -542,6 +556,225 @@ ADE_FUNC(getCmdBriefing,
...
+		return ade_set_args(L, "b", false);
+	}
+}
+
+ADE_FUNC(replayMission,
+	l_UserInterface_Debrief,
+	nullptr,
+	"Resets the mission outcome and begins the current mission again if in campaign",
+	"number",
+	"Returns 1 when completed")
```
Goober5000, line 745: Same comment about return value

code/scripting/api/libs/ui.cpp
```diff
@@ -542,6 +556,225 @@ ADE_FUNC(getCmdBriefing,
...
+
+	return ade_set_args(L, "i", 1);
+}
+
+ADE_FUNC(acceptMission,
+	l_UserInterface_Debrief,
+	nullptr,
+	"Accepts the mission outcome, saves the stats, and begins the next mission if in campaign",
+	"number",
+	"Returns 1 when completed")
```
Goober5000, line 764: Same comment about return value

code/scripting/api/objs/debriefing.cpp
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", stage->recommendation_text);
+}
+
+ADE_VIRTVAR(isValid,
```
Goober5000, line 56: `isValid` is used, by convention, to see whether the Lua object itself is a valid object.  This should have another name, perhaps `isVisible`.

Goober5000 (changes requested):
Partial review for now

code/scripting/api/libs/ui.cpp
MjnMixael: This exists in freespace.h which is the reason used all of these here instead. From what I understand, I can't call a method that exists outside of Code.

code/scripting/api/libs/mission.cpp
MjnMixael: Yes. Good catch

code/scripting/api/libs/ui.cpp
Goober5000: No, it's still possible; `game_stop_looped_sounds()` is called from keycontrol.cpp.

There's no inherent restriction against calling functions between projects, but there are occasional gotchas.  The main gotcha for our purposes is that any function in freespace.h must have a "stub function" that is defined when you're not compiling freespace.cpp.  This is why FRED and qtFRED have stub functions in their respective copies of fredstubs.cpp.

MjnMixael: Alright

code/scripting/api/objs/debriefing.h
```diff
@@ -0,0 +1,15 @@
...
+#include "mission/missionbriefcommon.h"
+#include "missionui/missiondebrief.h"
+#include "scripting/ade_api.h"
+
+namespace scripting {
+namespace api {
+
+DECLARE_ADE_OBJ(l_DebriefStage, debrief_stage);
+
+DECLARE_ADE_OBJ(l_Debrief, debriefing);
```
BMagnu, line 12: Similar to the other SCPUI PR's, this is a problem: We don't want to copy the debrief object here

BMagnu (changes requested):

BMagnu (approved):
It's good now from the perspective of my comments regarding the underlying data type of the new API object.
For the other points, I defer to @Goober5000.

code/scripting/api/libs/mission.cpp
```diff
@@ -1506,13 +1506,28 @@ ADE_FUNC(isRedAlertMission,
 ADE_FUNC(hasCommandBriefing,
 	l_Mission,
 	nullptr,
-	"Determines if the current mission has a command briefing",
```
Goober5000, line 1509: This should still say "command briefing"

code/scripting/api/libs/mission.cpp
```diff
@@ -1506,13 +1506,28 @@ ADE_FUNC(isRedAlertMission,
...
 {
 	return ade_set_args(L, "b", mission_has_cmd_brief() != 0);
 }
 
+ADE_FUNC(hasDebriefing,
+	l_Mission,
+	nullptr,
+	"Determines if the current mission has a debriefing",
+	"boolean",
+	"true if command briefing, false otherwise.")
```
Goober5000, line 1521: "debriefing" instead of "command briefing"

code/scripting/api/libs/mission.cpp
```diff
@@ -1506,13 +1506,28 @@ ADE_FUNC(isRedAlertMission,
...
+{
+	bool check = true;
+	if (The_mission.flags[Mission::Mission_Flags::Toggle_debriefing]) {
+		check = false;
+	}
+	
+	return ade_set_args(L, "b", check);
```
Goober5000, line 1528: This can be streamlined to
```
return ade_set_args(L, "b", The_mission.flags[Mission::Mission_Flags::Toggle_debriefing]);
```

code/scripting/api/libs/ui.cpp
```diff
@@ -542,6 +558,208 @@ ADE_FUNC(getCmdBriefing,
...
+ADE_FUNC(getTraitor,
+	l_UserInterface_Debrief,
+	nullptr,
+	"Get the traitor stage",
+	"debriefing_stage",
+	"The traitor stage or nil if the player is not traitor. Will also clear out any earned stats if true")
+{
+	if (Turned_traitor) {
+		Player->flags &= ~PLAYER_FLAGS_PROMOTED;
+		scoring_level_init(&Player->stats);
```
Goober5000, line 686: I would move the clearing of stats to another function.  By convention, `get` and `is` functions are not supposed to have side effects.

code/scripting/api/libs/ui.cpp
```diff
@@ -542,6 +558,208 @@ ADE_FUNC(getCmdBriefing,
...
+	l_UserInterface_Debrief,
+	nullptr,
+	"Gets whether or not the player must replay the mission. Will also clear out any earned stats if true",
+	"boolean",
+	"true if must replay, false otherwise")
+{
+
+	if (Must_replay_mission) {
+		Player->flags &= ~PLAYER_FLAGS_PROMOTED;
+		scoring_level_init(&Player->stats);
```
Goober5000, line 703: Same comment regarding clearing of stats.

code/scripting/api/libs/ui.cpp
```diff
@@ -542,6 +558,208 @@ ADE_FUNC(getCmdBriefing,
...
+		return ade_set_args(L, "b", true);
+	} else {
+		return ade_set_args(L, "b", false);
+	}
+}
+
+ADE_FUNC(replayMission,
+	l_UserInterface_Debrief,
+	nullptr,
+	"Resets the mission outcome and begins the current mission again if in campaign",
```
Goober5000, line 728: Is campaign mode relevant to this function?

code/scripting/api/libs/ui.cpp
```diff
@@ -542,6 +558,208 @@ ADE_FUNC(getCmdBriefing,
...
+
+	API_Access = false;
+
+	return ade_set_args(L, "b", true);
+}
+
+ADE_FUNC(acceptMission,
+	l_UserInterface_Debrief,
+	nullptr,
+	"Accepts the mission outcome, saves the stats, and begins the next mission if in campaign",
```
Goober5000, line 747: Is campaign mode relevant to this function?

code/scripting/api/objs/debriefing.cpp
```diff
@@ -0,0 +1,132 @@
...
+	return ade_set_args(L, "s", stage.getStage()->recommendation_text);
+}
+
+ADE_VIRTVAR(isVisible,
+	l_DebriefStage,
+	nullptr,
+	"The result of the stage formula",
+	"boolean",
+	"true if the stage should be displayed, false otherwise")
+{
```
Goober5000, line 74: See the new PR #4657.  The `isVisible` virtvar should be changed to a `checkVisible` function, for the same reasons as in that PR.  The documentation should also have the same updates as in that PR.

code/scripting/api/objs/player.cpp
```diff
@@ -173,6 +173,28 @@ ADE_VIRTVAR(AutoAdvance,
 	return ade_set_args(L, "b", plr->get()->auto_advance != 0);
 }
 
+ADE_VIRTVAR(SkipPopup,
```
Goober5000, line 176: For clarity this should be called `ShowSkipPopup`

code/scripting/api/objs/debriefing.cpp
```diff
@@ -0,0 +1,132 @@
...
+	debrief_stage_h stage;
+	if (!ade_get_args(L, "o", l_DebriefStage.Get(&stage))) {
+		return ADE_RETURN_NIL;
```
Goober5000, line 77: This can be changed to return false.  Invalid stages certainly won't be shown, and there are other ways elsewhere to check for stage validity.

Goober5000 (changes requested):
A couple new things.  The most important is the `isVisible` change.

code/scripting/api/libs/ui.cpp
MjnMixael: For this one yes. It calls debrief_accept() which has explicit checks for what to do in campaign mode. If not in campaign made then it just send you back to the mainhall.

code/scripting/api/objs/debriefing.cpp
MjnMixael: Note: after the discussion in #4657, I revisited this section. I decided here to keep checkVisible which differs from the briefing API because debriefing has multiple different types of stages (promotion, badge, traitor, standard) and wanted to allow the API the greatest flexibility in how to handle them. The latest commit bypasses evaluating sexps during Init when run through the API, leaving them to be evaluated here as designed.

Goober5000: That all sounds good.

code/scripting/api/objs/debriefing.cpp
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	return ade_set_args(L, "s", stage.getStage()->recommendation_text);
+}
+
+ADE_VIRTVAR(checkVisible,
+	l_DebriefStage,
+	nullptr,
+	"Evaluates the stage formula and returns the result. Could potentially have side effects if the stage formula has a 'perform-actions' or similar operator. "
+	"Note that the standard UI evaluates the formula exactly once per stage on briefing initialization.",
```
Goober5000, line 72: Should say "debriefing initialization" here

code/scripting/api/objs/debriefing.cpp
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "s", stage.getStage()->recommendation_text);
+}
+
+ADE_VIRTVAR(checkVisible,
```
Goober5000, line 68: This should be a function, not a virtvar

code/scripting/api/objs/debriefing.cpp
```diff
@@ -0,0 +1,133 @@
...
+	"true if the stage should be displayed, false otherwise")
+{
+	debrief_stage_h stage;
+	if (!ade_get_args(L, "o", l_DebriefStage.Get(&stage))) {
+		return ADE_RETURN_FALSE;
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
```
Goober5000, line 83: This can be removed after the virtvar is changed to a function

code/scripting/api/libs/ui.cpp
Goober5000: 👍

Goober5000 (commented):
Just a couple more things.

code/scripting/api/libs/ui.cpp
Goober5000: Ok

code/scripting/api/objs/debriefing.cpp
MjnMixael: Right. Done.

Goober5000 (approved):

PR #4617, opened 2022-09-05
Automatic thruster activation

Baezon:
If 'Use flags' are omitted from thrusters, their activation criteria are dynamically calculated on the fly automatically. 
- Minor fix by making sure `norm` is normalized, it was always `norm` * `length` long so behavior is unchanged, norm is just guaranteed to be normalized now. 
- Distance cutoff slightly modified to also take into account length. 
- Also for shits and giggles gliding will now activate relevant thrusters appropriately.


EatThePath (approved):
Code looks good to me, and it appears to test as advertised.

Talon1024:
I remember Wing Commander Saga made thruster plumes disappear whenever the pilot activated glide mode...

PR #4620, opened 2022-09-05
Properly encapsulate special flag parsing

BMagnu:
Previously, special flag parsing was done entirely manual from the raw flag strings afterwards, with the implict guarantee that no special flag is forgotten about.
This now properly encapsulates this, properly handling automatic dispatch of special flags to respective parsing functions (with the common logic taken out into the caller function). In addition, special flags and unknown inputs are now properly separated.
Overall, this makes it a lot easier, simpler and more robust to add more special flags.


code/parse/parselo.h
```diff
@@ -158,8 +158,8 @@ extern void parse_string_flag_list(int *dest, flag_def_list defs[], size_t defs_
 // A templated version of parse_string_flag_list, to go along with the templated flag_def_list_new.
 // If the "is_special" flag is set, or a string was not found in the def list, it will be added to the unparsed_or_special_strings Vector
 // so that you can process it properly later
```
Goober5000, line 160: Can you update the comment to account for the new parsing capability?

Also it would be good to generally explain what the `special_flag_def_list_new` type does and how it can be used.

code/parse/parselo.h
```diff
@@ -158,8 +158,8 @@ extern void parse_string_flag_list(int *dest, flag_def_list defs[], size_t defs_
...
-template<class T, class Flagset>
-void parse_string_flag_list(Flagset& dest, flag_def_list_new<T> defs [], size_t n_defs, SCP_vector<SCP_string>* unparsed_or_special_strings)
+template<class Flagdef, class Flagset>
+void parse_string_flag_list(Flagset& dest, const Flagdef defs[], size_t n_defs, SCP_vector<SCP_string>* unparsed_or_special_strings)
```
Goober5000, line 162: With the change in usage, I assume the last parameter should be `unparsed` rather than `unparsed_or_special`

Goober5000 (changes requested):

code/globalincs/vmallocator.h
```diff
@@ -1,6 +1,7 @@
 #ifndef _VMALLOCATOR_H_INCLUDED_
 #define _VMALLOCATOR_H_INCLUDED_
 
+#include <algorithm>
```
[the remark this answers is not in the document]
BMagnu: Cause the header already used things from algorithm. It just never was an issue, since it was (before) always included somewhere where algorithm was already included prior. Since I added it into pstypes.h, where algorithm was not included, it would error otherwise

code/parse/parselo.h
Goober5000: Ah, ok.

Goober5000 (approved):

PR #4622, opened 2022-09-06
Prevent closed thread from Video Player from re-closing

BMagnu:
The Video Player keeps a thread for decoding. When you tell the player to close, all of its data is deallocated, and the decoding thread is closed. While deallocating the data is secured against deleting it multiple times, closing the thread is not. We should check if the thread is still joinable before we just blindly call join on it to prevent crashes. If it's not joinable, we can pretty much just ignore it at that state


JohnAFernandez (approved):

JohnAFernandez:
LGTM.  Thanks for being awesome!

PR #4623, opened 2022-09-06
Add maneuvering thruster easing

Baezon:
After some double checking to make sure it was only cosmetic, the `phys_info` thrust values are now dedicated to being cosmetic only, this way they can be cosmetically altered via a new 'thruster easing' parameter. Helps make thruster activation less jittery when used by fighters.

Does some miscellaneous clean up putting all the thrust values into a single vector, and make a similarly cosmetic `rotational_thrust` vector, which takes the place of `desired_rotvel` for use in thrusters. 
Remove vestigal field `prev_fvec`.
Fixes several erroneous docs which says the values are 0-1 when they're really -1 - 1.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -9523,7 +9523,7 @@ float dock_orient_and_approach(object *docker_objp, int docker_index, object *do
 
 	aip = &Ai_info[Ships[docker_objp->instance].ai_index];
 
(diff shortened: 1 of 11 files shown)


Goober5000 (approved):
LGTM

EatThePath:
I got a chance to do some testing of this and all seems good

PR #4624, opened 2022-09-06
Small fix for banking using AI functions

Baezon:
Follow-up to #4417. Even if the ship has reached its target position it may still need to bank and so should continue if non-zero.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -2048,8 +2048,8 @@ void vm_angular_move_forward_vec(const vec3d* goal_f, const matrix* orient, cons
 				theta_goal.xyz.y = w_in->xyz.y * d;
 			}
 		}
-		// continue to interpolate, unless we also have no velocity, in which case we have arrived
-		else if (vm_vec_mag_squared(w_in) < SMALL_NUM * SMALL_NUM) {
+		// continue to interpolate, unless we also have no velocity (and dont need to bank), in which case we have arrived
+		else if (vm_vec_mag_squared(w_in) < SMALL_NUM * SMALL_NUM && bank_vel == 0.0f) {
 			*next_orient = *orient;
 			vm_vec_zero(w_out);
 			return;


code/math/vecmat.cpp
```diff
@@ -2048,8 +2048,8 @@ void vm_angular_move_forward_vec(const vec3d* goal_f, const matrix* orient, cons
 				theta_goal.xyz.y = w_in->xyz.y * d;
 			}
 		}
-		// continue to interpolate, unless we also have no velocity, in which case we have arrived
-		else if (vm_vec_mag_squared(w_in) < SMALL_NUM * SMALL_NUM) {
+		// continue to interpolate, unless we also have no velocity (and dont need to bank), in which case we have arrived
+		else if (vm_vec_mag_squared(w_in) < SMALL_NUM * SMALL_NUM && bank_vel == 0.0f) {
```
BMagnu, line 2052: Is this not a case where you should ``fl_near_zero``? Yes, in the "no bank" cases it should pretty much always be a hard 0, but who knows if that value might be programmatically determined in the future, at which point it might not be an exact 0 anymore

Baezon: This definitely seems to me to be a hard "was assigned to 0" or not situation. I want to check for that it was assigned to 0, which is what everything that wants 0 banking does. There should never be a situation where 'no bank' is requested by some sort of float math which only gets 'approximately' 0. If there's one thing we *can* expect from float equality, it's that assigning 0.0f will always get you *exactly* 0.

BMagnu: > There should never be a situation where 'no bank' is requested by some sort of float math
Well, if we're sure of this assumption, then I agree

BMagnu (approved):

code/math/vecmat.cpp
Baezon: I did check, and all situations for which it is clear that 'no bank' is needed does so by using`0.0f`

PR #4625, opened 2022-09-06
fix subtitle clipping

Goober5000:
For subtitles drawn using the `$Show-subtitle base resolution:` feature, text near the side of the screen would get clipped off mysteriously:

screen0261

The clipping corresponded precisely to the edge of a 4:3 boundary:

4_3_box

This led me to discover that `gr_screen.clip_right_unscaled` was still set for a 4:3 box, rather than a 16:10 box as was expected.  Explicitly setting the clip boundary when setting the screen scale fixes the issue:

screen0262

For any graphics coders in the know, is this actually the proper fix?

### DIFF
--- a/code/camera/camera.cpp
+++ b/code/camera/camera.cpp
@@ -831,7 +831,10 @@ void subtitle::do_frame(float frametime)
 
 	auto sizing_mode = do_screen_scaling ? GR_RESIZE_FULL : GR_RESIZE_NONE;
 	if (do_screen_scaling)
+	{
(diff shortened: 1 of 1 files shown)


BMagnu (approved):
Should be simple enough, with the clip resize mode set to full by default.

Goober5000:
> Should be simple enough, with the clip resize mode set to full by default. After all, the subtitle resize is set. After that you set the clip to match the target resize. And with the resize mode of the clip actually using the just-computed resize, it should thus always match the intended subtitle screen target. Which is, IMO, the correct thing to do here

This sounds reasonable, but I'm wondering why the clip didn't match the widescreen proportions in the first place.

PR #4632, opened 2022-09-08
Fix issue with new pilots getting demoted too fast

### LINKED ISSUE #4614: Assert with new pilot, promotion, mission replay
If a fresh pilot plays a mission and is granted a promotion and then chooses to replay the mission, an assert will be triggered. "player rank became negative"

Use this campaign with a brand new pilot. In the first mission kill the manticore and the depart. You will be given a promotion in the Debrief. Do not accept. Click Replay mission. Assert triggers here.
[DebriefAssert.zip](https://github.com/scp-fs2open/fs2open.github.com/files/9474726/DebriefAssert.zip)

The-E:
If a new pilot manages to gain a rank during the first mission and then decides to replay that mission, the check to reset the rank was accidentally demoting the pilot by two ranks instead of one. Inverting the check and just iterating through the ranks table until the fitting rank can be found fixes this.

This fixes #4614


wookieejedi (approved):
Good find and fix

PR #4633, opened 2022-09-09
convert event-related timestamps to new system

Goober5000:
This change covers just the timestamps associated with goals and events, for better review and isolation purposes.

Follow-up to #4569.

### DIFF
--- a/code/mission/missiongoals.cpp
+++ b/code/mission/missiongoals.cpp
@@ -175,7 +175,7 @@ int Num_goals = 0;								// number of goals for this mission
 int Event_index = -1;  // used by sexp code to tell what event it came from
 bool Log_event = false;
 bool Snapshot_all_events = false;
-int Mission_goal_timestamp;
+TIMESTAMP Mission_goal_timestamp;
 
 mission_event Mission_events[MAX_MISSION_EVENTS];
 mission_goal Mission_goals[MAX_GOALS];		// structure for the goals of this mission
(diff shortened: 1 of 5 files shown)


Goober5000:
After extensive testing and fixing several bugs, this is now ready for review.

code/mission/missiongoals.cpp
```diff
@@ -1522,3 +1551,23 @@ int mission_event_find_sexp_tree(int root_node)
...
+
+int mission_event_lookup(const char *name)
+{
+	Assertion(name != nullptr, "Event name cannot be null!");
+
+	for (int i = 0; i < Num_mission_events; ++i)
+		if (!stricmp(Mission_events[i].name, name))
+			return i;
+	return -1;
+}
```
BMagnu, line 1573: What are these added for?
Remnants of a different PR maybe?

BMagnu (changes requested):
Looks mostly good.
As this is fairly deep in the logic, it's gonna need a lot of testing, but it sounds like you did that.
This interval behaviour will need to be well-documented on the respective wiki page as well.
One question though

code/mission/missiongoals.cpp
Goober5000: These are in preparation for #4634, but I can move them to that PR if you think I should.

BMagnu: eh, its fine in that case. I was just wondering

Issue #4184, opened 2022-04-18
Confusing short-circuit behavior when a variable is passed to `has-docked-delay`'s third argument in a repeating event


naomimyselfandi:
`has-docked-delay` and `has-undocked-delay`'s third parameter controls how many times the ships in question need to have (un)docked. When this parameter is satisfied with an SEXP variable (or, probably, with any non-constant SEXP), the expression short-circuits after returning `true`, leading to confusing behavior in repeating events.

Ideally, those operators wouldn't short-circuit when the third argument isn't a constant, but that would be a breaking change. Adding a parameter to decide whether it should short-circuit would be another possibility; however, since the behavior isn't particularly hard to work around, simply documenting it with a suggestion to use `every-time` might be sufficient.

Goober5000:
The short-circuiting behavior isn't due to the third argument being a constant or a variable (because `eval_num` doesn't care and will correctly evaluate it either way), but due to the entire SEXP being based on the mission log.  Any SEXP that uses the mission log - which generally corresponds to any SEXP that uses `-delay` - will short-circuit.

I'm having trouble picturing exactly how your repeating event would work, but I suspect that if you substituted a different mission-log-based SEXP you would find that it has the same issue.

naomimyselfandi:
I mean... I understand _why_ it behaves that way, more or less. That doesn't make it any less surprising (in a "principle of least astonishment" sense), and it wouldn't be any less surprising if we'd encountered it as `( is-destroyed-delay 0 "@PriorityTarget[Aquarius 1]" )` or `( is-cargo-known-delay 0 "@SuspiciousTransport[Lambda 1]" )` or what have you. I'd actually go so far as to say that short-circuiting an operation that can't be short-circuited is a bug. Either way, it's certainly _undesired_ behavior... which is why I suggested some way to opt out. (Maybe a `do-not-short-circuit` operator?)

The repeating event in this case has a shuttle repeatedly undock with one ship, dock with another, and repeat that process a large number of times. Mjn implemented it with two variables that tracked how many times the transport had docked to both other ships, which really looked like it should have worked.

Goober5000:
I'm certainly in favor of the principle of least astonishment, but mission-log SEXPs have *always* short-circuited, regardless of their arguments.  And I'm not clear on what might imply that it *can't* be short-circuited.

I hear you about the undesired behavior, and it's certainly come up in the past -- we have both `is-tagged` and `has-been-tagged-delay` for example.

[...]

PR #4634, opened 2022-09-09
add reset-event and reset-goal

Goober5000:
These new SEXP operators clear all status information from an event or goal, making it as if the event or goal had never been evaluated.

Implements feature discussed in #4184.

~~In draft status pending tests.~~  Now tested.  ~~Depends on #4633.~~


code/parse/sexp.cpp
```diff
@@ -17367,6 +17369,59 @@ int sexp_goal_incomplete(int n)
...
+		auto name = CTEXT(n);
+		int event_num = mission_event_lookup(name);
+		if (event_num >= 0)
+		{
+			auto eventp = &Mission_events[event_num];
+
+			eventp->result = 0;
+			eventp->timestamp = TIMESTAMP::invalid();
+
+			// clear some flags but not all
```
naomimyselfandi, line 17388: I don't know if this comment adds any value. It just says what the code does, and there isn't any complex magic here that benefits from being summarized. The comment would be more useful if it said why it's clearing these flags in particular.

code/parse/sexp.cpp
```diff
@@ -17367,6 +17369,59 @@ int sexp_goal_incomplete(int n)
...
+			eventp->flags &= ~MEF_DIRECTIVE_SPECIAL;
+			eventp->flags &= ~MEF_DIRECTIVE_TEMP_TRUE;
+			eventp->flags &= ~MEF_TIMESTAMP_HAS_INTERVAL;
+
+			eventp->count = 0;
+			eventp->satisfied_time = TIMESTAMP::invalid();
+			eventp->born_on_date = TIMESTAMP::invalid();
+			eventp->previous_result = 0;
+
+			if (eventp->formula >= 0)
```
naomimyselfandi, line 17399: `flush_sexp_tree` already short-circuits on a negative node index.

(More curiosity than anything, but when would an event not have a formula?)

code/parse/sexp.cpp
```diff
@@ -17367,6 +17369,59 @@ int sexp_goal_incomplete(int n)
...
+				flush_sexp_tree(eventp->formula);
+		}
```
naomimyselfandi, line 17401: This should probably warn if the lookup failed.

code/parse/sexp.cpp
```diff
@@ -17367,6 +17369,59 @@ int sexp_goal_incomplete(int n)
...
+	}
+}
+
+/**
+ * Resets a goal, its status, and its nodes so that it is as if the goal had never been evaluated.
+ */
+void sexp_reset_goal(int node)
```
naomimyselfandi, line 17408: Same comments as above about `flush_sexp_tree` and warning on lookup failure.

Goober5000:
Just a note to say thanks for the review.  I'll get to these comments as soon as I can - just been busy.

code/parse/sexp.cpp
Goober5000: Good point

code/parse/sexp.cpp
Goober5000: So it does.  Tweaked.

Hmm... it looks like an event's formula is set to -1 when it is failed, or known-false.  That will need to be modified for this PR to work.

code/parse/sexp.cpp
Goober5000: Good point

Goober5000: Actually... none of the other sexps warn when an event lookup fails.  This includes the retail `is-event-true-delay` etc.  I suppose that shouldn't rule out warning here though.

Goober5000:
Comments addressed.  I'll still need to figure out a way to prevent the event formula from being set to -1.

Goober5000:
Ok, this has been tested and should be ready now.

JohnAFernandez:
I took a look, and I'm ready to approve pending another quick look by @naomimyselfandi

PR #4636, opened 2022-09-09
properly account for string sizes in debrief_choose_voice

### LINKED ISSUE #4635: Crash when traitor debrief
There's a crash when becoming a traitor and ending the mission.

This can be reproduced using `fs2_open` (on Linux) without mods. Crashes with both `22.2` and git `16a8a46`.

An example that reproduces the crash:
1. Start `fs2_open` without mods.
2. Go to the tech room mission simulator and play SM01-01.fs2.
3. Fire on a friendly to become a traitor.
4. Jump out.
5. (Crash `free(): invalid pointer`)

```
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140737177834496) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140737177834496) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140737177834496, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff5842476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff58287f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff58896f6 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff59dbb8c "%s\n")
    at ../sysdeps/posix/libc_fatal.c:155
#6  0x00007ffff58a0d7c in malloc_printerr (str=str@entry=0x7ffff59d9764 "free(): invalid pointer") at ./malloc/malloc.c:5664
#7  0x00007ffff58a2ac4 in _int_free (av=<optimized out>, p=<optimized out>, have_lock=0) at ./malloc/malloc.c:4439
#8  0x00007ffff58a54d3 in __GI___libc_free (mem=<optimized out>) at ./malloc/malloc.c:3391
#9  0x00007ffff5d4be8f in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_assign(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () at /lib/x86_64-linux-gnu/libstdc++.so.6
#10 0x00007ffff5d4c23d in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::operator=(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () at /lib/x86_64-linux-gnu/libstdc++.so.6
#11 0x0000555556823189 in debrief_traitor_init() ()
#12 0x0000555556825771 in debrief_init() ()
#13 0x000055555634fbff in game_enter_state(int, int) ()
#14 0x00005555564b3560 in gameseq_set_state(int, int) ()
#15 0x000055555634d8e5 in game_process_event(int, int) ()
#16 0x00005555564b39ca in gameseq_process_events() ()
#17 0x0000555556351651 in game_main(int, char**) ()
#18 0x0000555556353dd0 in main ()
```

Goober5000:
The null terminator was written past the end of the string buffer, which intruded on the next field in the struct.  Fixing this fixes #4635.  Also adjust the string concatenation a couple lines up because it could also write past the end of the buffer.

### DIFF
--- a/code/missionui/missiondebrief.cpp
+++ b/code/missionui/missiondebrief.cpp
@@ -943,10 +943,10 @@ void debrief_choose_voice(char *voice_dest, size_t buf_size, char *voice_base, i
 	else
 	{
 		strncpy(voice_dest, "9_", buf_size);
-		strncat(voice_dest, voice_base, buf_size);
+		strncat(voice_dest, voice_base, buf_size-2);
 	}
 	// Ensure null terminator
-	voice_dest[buf_size] = '\0';
+	voice_dest[buf_size-1] = '\0';
 }
 
 void debrief_choose_medal_variant(char *buf, int medal_earned, int zero_based_version_index)


Kiloku (commented):
Logic seems right to me. From reading the docs on `strncat`, the null-terminator should be appended automatically, but I see no harm in setting it directly.

Goober5000:
True, but the risk here is the lack of null terminator from the `if (default_to_base)` block.

Kiloku (approved):

PR #4638, opened 2022-09-10
Allow Autotarget to get new target when current target is deathrolling

wookieejedi:
Discussion starts here: https://discord.com/channels/223511295431933953/722747742996267068/1018264927699476571


Kiloku (approved):
Simple enough, and testing on Star Fox EH (which is the mod that sparked this request), achieves the desired behavior

PR #4640, opened 2022-09-11
several minor HUD enhancements

Goober5000:
add `getDirectiveLines` and `setPosition` script functions properly error if trying to set a HUD gauge name
remove the unneeded `Directive_frames_loaded` variable improve accuracy of SEXP help for HUD gauge operators

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -439,13 +439,13 @@ const char* HudGauge::getCustomGaugeText()
 	return custom_text.c_str();
 }
 
-void HudGauge::updateGaugeCoords(int _x, int _y)
+void HudGauge::setGaugeCoords(int _x, int _y)
 {
 	position[0] = _x;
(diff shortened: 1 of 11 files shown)


BMagnu (commented):
Looks good mostly.
Just a question on the documentation change regarding "Works only for custom gauges".
Did it never work to set certain retail gauges? Did that change? Does it still work but due to the uncertain nature of which works and which doesn't it's just removed from the docs?

Goober5000:
On closer inspection, these operators *did* work with retail gauges back when they were first introduced.  So that's a regression.  I'll modify them.

Goober5000:
Ugh, this was broken back in the Antipodes #7 merge, in 2010.  See commit 09fad95070.

I've done some modifications to allow certain sexps to work for either custom or builtin types, as their capabilities allow.

BMagnu (approved):

PR #4641, opened 2022-09-11
Add Angle SEXPs

BMagnu:
This adds SEXPs to simplify angle and direction testing.
One, ``angle-vectors`` simply allows for calculation of an angle between two vectors.
The other, ``angle-facing-object`` returns the angle between the fvec of a ship and the direction from a ship to a target object. This massively simplifies "is ship facing object sufficiently"-checks.


Goober5000 (approved):
Looks good.

PR #4642, opened 2022-09-11
Auto Truncate default ship names

wookieejedi:
Previously, placing a ship that had a long class name might create a default name that was too long and thus trigger a buffer overflow assert.

Such behavior was confusing if someone was just placing things in FRED and not sure at all what happened or how to fix it.

This PR replaces the assert with code to auto truncate the ship class part of the name to ensure that the default name will fit into the designated size.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -10046,16 +10046,26 @@ int ship_create(matrix* orient, vec3d* pos, int ship_type, const char* ship_name
 	// but it breaks multi, so there will just be a warning on debug instead.
 	if ((ship_name == nullptr) || (ship_name_lookup(ship_name) >= 0) /*|| (ship_find_exited_ship_by_name(ship_name) >= 0)*/) {
 		// regular name, regular suffix
+		char default_name[NAME_LENGTH];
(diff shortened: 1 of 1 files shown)


code/ship/ship.cpp
```diff
@@ -10046,16 +10046,26 @@ int ship_create(matrix* orient, vec3d* pos, int ship_type, const char* ship_name
 	// but it breaks multi, so there will just be a warning on debug instead.
 	if ((ship_name == nullptr) || (ship_name_lookup(ship_name) >= 0) /*|| (ship_find_exited_ship_by_name(ship_name) >= 0)*/) {
 		// regular name, regular suffix
+		char default_name[NAME_LENGTH];
```
Goober5000, line 10049: I'd call this `base_name` instead; `default_name` refers to the entire thing, including the suffix.

Goober5000 (changes requested):
One comment, otherwise this looks good.  I double checked the math too.

code/ship/ship.cpp
wookieejedi: Good point, just changed

Goober5000 (approved):

PR #4644, opened 2022-09-12
remove two unnecessary asserts

Goober5000:
It's not necessary here to assert on the correct variable type, as having the wrong type here won't cause the code to break.  More importantly, the error will have been flagged before this point anyway, and this code is used for formatting the SEXP to be displayed in an error message.  It's not helpful to pre-empt an error message with an assert.


wookieejedi (approved):
Looks reasonable to me.

PR #4646, opened 2022-09-12
Fix the cmd_briefing scripting API

BMagnu:
The API previously had incorrect documentation, and was using a copy of the cmd_briefing object for the lua data, instead of a reference to the object. This is troublesome, since that implies that no changes can be propagated to and from FSO and can cause hard-to-find bugs if there is ever any API method that can modify the object


Goober5000 (approved):
Good thinking.

PR #4647, opened 2022-09-12
better adjustment of timestamp delta

Goober5000:
When debugging, don't adjust the timestamp by the exact difference, since that would make the frametime 0 and cause div-0 errors.  Instead pretend that the previous frame was a reasonable duration.  Also, perform the timestamp check separately from the frametime check, since they are not always consistent.


wookieejedi (approved):
Went over this a few times and looks logical.

PR #4652, opened 2022-09-14
Improved shaders and image based lighting

qazwsxal:
This was made in collaboration with @EatThePath and should be merged in after his "specular rubicon" PR #4302 . 

Main contributions in this PR:
1) The lighting model has been changed to be as close to that used by Unreal Engine 4 as possible. The main justification for this is that people are using tools like Substance Painter which use a UE4-like lighting model, and shouldn't be surprised by ships looking brighter/darker in FSO.

2) Improved image based lighting. The calculation of envmap sharpness has been fixed (no more mirror ships with when using slightly rough surfaces) and a low-rez irradiance cubemap (16x16x6) is generated along with the envmap, allowing for environment lighting to contribute to diffuse illumination too.


SamuelCho:
Also a much needed change. In regards to IBL, much was left to be desired after the initial PBR implementation. Especially love the cubemaps diffuse lighting contribution. Again, great work.

The-E (approved):

PR #4653, opened 2022-09-14
Improved beam, blob and warp lighting

qazwsxal:
This was made in collaboration with @EatThePath, and focuses on improving the look of beams, weapons and warp lighting
This should be merged in after #4652 as it relies upon some of the changes made there.

Based off research by epic games, [Real Shading in Unreal Engine 4](https://cdn2.unrealengine.com/Resources/files/2013SiggraphPresentationsNotes-26915738.pdf), the tube and point light lighting calculations have been modified. 

Firstly, the point-source approximation used by beams/tube lights has been changed. Instead of choosing the nearest point of the tube light to the shaded fragment, the eye direction and surface normal are taken into account. This improves the look of the reflections of beams (especially on curved surfaces)

Secondly, all beam and point lights now have a sourceRadius. This slightly modifies the lighting calculation, nudging the position of the point-source-approximation so that the specular reflections appear to have a physical size. This prevents glossy reflections of point sources from diminishing into a single point and gives weapon lighting true "thickness" 

22.2.0

With improved lighting
improved

This also improves the look of subspace portal lighting:
(from joe's warmachine stream)
joe's warmachine stream

Pics from ETP's work:
ETP1
ETP2

### DIFF
--- a/code/def_files/data/effects/deferred-f.sdr
+++ b/code/def_files/data/effects/deferred-f.sdr
@@ -48,6 +48,9 @@ layout (std140) uniform lightData {
 
 uniform float sourceRadius = 5.0; // This should be passed in part of lightData
 
+// Nearest point sphere and tube light calculations taken from 
+// "Real Shading in Unreal Engine 4" by Brian Karis, Epic Games
+// Part of SIGGRAPH 2013 Course: Physically Based Shading in Theory and Practice
 void GetLightInfo(vec3 position, in float alpha, in vec3 reflectDir, out vec3 lightDir, out float attenuation, out float area_normalisation)
 {
 	if (lightType == LT_DIRECTIONAL) {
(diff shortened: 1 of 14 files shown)


SamuelCho:
great work! I always wanted to get some sort of area lights feature working after the PBR/HDR changes but hadn't had the time. thanks for putting some time into this.

code/starfield/starfield.cpp
```diff
@@ -468,16 +472,8 @@ void parse_startbl(const char *filename)
...
-					stuff_float(&sbm.spec_g);
-					stuff_float(&sbm.spec_b);
-				}
-				else {
-					sbm.spec_r = sbm.r;
-					sbm.spec_g = sbm.g;
-					sbm.spec_b = sbm.b;
-				}
+				if (optional_string("$SunSpecularRGB:"))
+					Warning(LOCATION, "Sun %s tried to set SunSpecularRGB. This feature has been deprecated and will be ignored.", sbm.filename);
```
The-E, line 476: I am not sure how often this is used, but I think this can be downgraded to a log print instead of a Warning

The-E (changes requested):

The-E (approved):

PR #4654, opened 2022-09-14
Light power fix

qazwsxal:
@EatThePath has suggested some increases to light brightness necessary because of the improved shaders introduced by #4652 and #4653.


EatThePath:
Of note this adjustment was tested on the default settings of the MVPs and brings brightness levels to as near as exactly matching the pre-#4652 as I can distinguish, which should minimize the amount of shift typical mods see in their look

The-E (approved):

PR #4655, opened 2022-09-14
improve huge_fire_info lookup

Goober5000:
According to EatThePath's frame profiling, this one ship name lookup takes 2% of the frame time.  It can be easily optimized by making use of the ship registry.


JohnAFernandez (approved):
Good to leverage existing data

PR #4656, opened 2022-09-14
Scpui api briefing map

MjnMixael:
Creates an API that allows drawing the currently loaded mission's briefing icon map and ability to manipulate what stage it is on.

Note that because of how retail rendered the cut-to-stage static, rendering of that has been removed from the API to be handled by the scripts so that it can be properly sized dynamically.

Also note that clicking on the icons of the map rendered through the API is not currently possible. That will take further investigation to sort out and allow. There are several bits of code that are commented out, but kept to make that later PR easier.

### DIFF
--- a/code/missionui/missionbrief.cpp
+++ b/code/missionui/missionbrief.cpp
@@ -1948,6 +1948,7 @@ void brief_api_do_frame(float frametime)
 				time,
 				Current_brief_stage);
 
+			//A few items commented out, but keeping a record for closeup icon fixing in a later PR - Mjn
 			// Brief_playing_fade_sound = 0;
 			Last_brief_stage = Current_brief_stage;
 			// brief_reset_icons(Current_brief_stage);
@@ -2051,7 +2052,7 @@ void brief_close()
 void brief_api_close()
 {
 	if (Brief_inited == FALSE) {
-		nprintf(("Warning", "brief_close() returning without doing anything\n"));
+		nprintf(("Warning", "brief_api_close() returning without doing anything\n"));
 		return;
 	}
 
--- a/code/missionui/missionbrief.cpp
+++ b/code/missionui/missionbrief.cpp
@@ -1018,6 +1018,7 @@ void brief_api_init()
 	hud_anim_init(&Fade_anim, bstat_x, bstat_y, Brief_static_name[gr_screen.res]);
 	hud_anim_load(&Fade_anim);
 
+	//This will need to be replaced with an API version in a later PR - Mjn
 	common_select_init();
 
 	// This loads the static bitmap for stage cuts - Mjn
--- a/code/missionui/missionbrief.cpp
+++ b/code/missionui/missionbrief.cpp
@@ -61,11 +61,6 @@ static int Brief_goals_coords[GR_NUM_RESOLUTIONS][4] = {
 	}
 };
 
-// static effect coords for API
-int bstat_x;
-int bstat_y;
-int brief_api;
-
 static int	Current_brief_stage;	// what stage of the briefing we're on
 static int	Last_brief_stage;
 static int	Num_brief_stages;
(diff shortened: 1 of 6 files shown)


code/missionui/missionbrief.cpp
```diff
@@ -976,6 +976,62 @@ void brief_init()
...
+		brief_set_new_stage(&Briefing->stages[0].camera_pos,
+			&Briefing->stages[0].camera_orient,
+			0,
+			Current_brief_stage);
+		brief_reset_icons(Current_brief_stage);
+	}
+
+	Brief_mouse_up_flag = 0;
+	Closeup_font_height = gr_get_font_height();
+	Closeup_icon = NULL;
```
BMagnu, line 1031: this should be ``nullptr`` instead of ``NULL``.
``NULL`` is a remnant of old C-style code, and while it's not too much of a problem, we intend to replace it in code that we touch as to eventually get rid of all NULLs

code/missionui/missionbrief.cpp
```diff
@@ -1787,6 +1870,110 @@ void brief_do_frame(float frametime)
...
+				check_jump_flag = 0;
+			}
+
+			if (check_jump_flag) {
+				if (abs(Current_brief_stage - Last_brief_stage) > 1) {
+					Quick_transition_stage = Current_brief_stage;
+					Current_brief_stage = Last_brief_stage;
+					Assert(Current_brief_stage >= 0);
+					Start_fade_up_anim = 1;
+					goto Transition_done;
```
BMagnu, line 1908: Oof, this is some old and crusty code.
I get that you don't want to rework all of that though, and since it's a mirror of the non-API function, I guess it's fine.

code/scripting/api/libs/ui.cpp
```diff
@@ -2,6 +2,8 @@
 //
 #include "ui.h"
 
+#include "freespace.h"
```
BMagnu, line 5: This is a problem. We _really_ don't want to include freespace.h from _any_ file within the code folder. Do you really need this? I can't find the point in this file that is relying on some freespace.h declared value or function.

code/scripting/api/libs/ui.cpp
```diff
@@ -654,14 +668,102 @@ ADE_FUNC(skipTraining,
...
+// For now any loadout error checking needs to happen in the script to prevent FSO from
+// generating a popup that will not be interactible from Librocket. A later update to this
+// method will introduce return values instead of generating popups and those return values
+// can be used to handle loadout popups on the script side
+ADE_FUNC(commitToMission,
+	l_UserInterface_Brief,
+	nullptr,
+	"Commits to the current mission with current loadout data, and starts the mission. WIP, do not use!",
+	nullptr,
+	"nothing")
```
BMagnu, line 680: just make it a ``nullptr`` here as well instead of nothing, as you did with the other functions as well

code/scripting/api/libs/ui.cpp
```diff
@@ -654,14 +668,102 @@ ADE_FUNC(skipTraining,
...
+	return ADE_RETURN_NIL;
+}
+
+ADE_FUNC(callNextMapStage,
 	l_UserInterface_Brief,
 	nullptr,
-	"Starts the briefing map for the current mission. Doesn't currently do anything.",
+	"Sends the briefing map to the next stage.",
 	nullptr,
 	"nothing")
```
BMagnu, line 727: same here

code/scripting/api/libs/ui.cpp
```diff
@@ -654,14 +668,102 @@ ADE_FUNC(skipTraining,
...
+	brief_do_prev_pressed();
+	return ADE_RETURN_NIL;
+}
+
+ADE_FUNC(callFirstMapStage,
+	l_UserInterface_Brief,
+	nullptr,
+	"Sends the briefing map to the first stage.",
+	nullptr,
+	"nothing")
```
BMagnu, line 751: same here

BMagnu (changes requested):
Some comments

code/scripting/api/libs/ui.cpp
MjnMixael: That is for line 717, specifically the `flRealframetime` value

code/missionui/missionbrief.cpp
MjnMixael: Yes, for now rewriting to avoid goto is outside the scope and the API brief map should work exactly as the normal one does.

code/scripting/api/libs/ui.cpp
BMagnu: I see. and other libs seem to do it as well like this. Meh.
Not a fan of how this is handled, but for this specific value it's ok I guess based on precendence

BMagnu (approved):

PR #4657, opened 2022-09-15
additional polishing of the briefing API

Goober5000:
Follow-up to #4602.  Overall the PR was good, but there are a few extra small changes that should go in. In order of priority:

1) The `exitLoop`, `skipMission`, and `skipTraining` functions should check for campaign mode.
2) The `isVisible` virtvar should be changed to a function, because calling it may have side effects.  In standard usage, the briefing formula is evaluated exactly once per stage on briefing initialization, so scripts should follow this convention.
3) The `hasGoalsSlide` function should be called `hasGoalsStage`, because "stage" is the common terminology in FSO.
4) Paging out mission messages is already done in `game_level_close` and should be removed from not only the scripting method but the method it was based on.
5) For consistency, unused parameters should use the `SCP_UNUSED` macro.
6) The `isTraining` flag check can be slightly streamlined using the double-not trick.

The `isVisible` virtvar => `checkVisible` function change and the `hasGoalsSlide` => `hasGoalsStage` rename will affect @MjnMixael's existing scripts.


code/scripting/api/libs/mission.cpp
```diff
@@ -1387,13 +1387,7 @@ ADE_FUNC(isInCampaignLoop, l_Mission, nullptr, "Get whether or not the current m
...
 ADE_FUNC(isTraining, l_Mission, nullptr, "Get whether or not the current mission being played is a training mission", "boolean", "true if in training, false if not")
 {
-	bool b = false;
-
-	if (The_mission.game_type & MISSION_TYPE_TRAINING) {
-		b = true;
-	}
-
-	return ade_set_args(L, "b", b);
+	return ade_set_args(L, "b", !!(The_mission.game_type & MISSION_TYPE_TRAINING));
```
BMagnu, line 1390: Hmm, would it not be easier to read if we just properly cast to a bool using either a c-style cast or a static cast?

BMagnu (changes requested):

code/scripting/api/libs/mission.cpp
Goober5000: I suppose so.

BMagnu (approved):

MjnMixael:
After reading this, I believe there is more discussion to be had specifically regarding the evaluation of sexps to get visible stages. #4656 implements a way to init the Briefing array and it calls brief_compact_stages(), which is required to make sure the briefing map only renders the unlocked stages. That method also evaluates the SEXPs for each stage. So between getting each stage's data here and setting up the rendering of the map, the sexps will be evaluated twice.

As you pointed out this is not generally a problem except for perform-actions. My thinking is that one case or the other needs to be able to evaluate for valid stages without running perform-actions.

Goober5000:
> EDIT: Alternatively an initBriefing func is added and the eval_sexp methods removed here and in the briefing map API because we can then assume the only stages remaining in the array are valid.

I like this option best.  It seems both the simplest and the least likely to cause problems elsewhere.  It also removes a category of invalid state, which is [good programming practice](https://kevinmahoney.co.uk/articles/my-principles-for-building-software/#make-invalid-states-unrepresentable).

MjnMixael:
Can you do that, then, as a part of this polishing PR and I will adjust the map PR accordingly?

Goober5000:
Yes, I can do that later this evening.

Goober5000:
It looks like very little needs to be changed.  I remove the `checkVisible` function and added an `initBriefing` function which will depend on the upcoming PR.

BMagnu (approved):

PR #4658, opened 2022-09-15
Fix Animation 3D Sound

BMagnu:
Previously, all animation sounds were played as 2D sounds.
Now, if they're properly attached to a parent submodel providing position (and they're on an actual object-model as opposed to stuff like skyboxes or cockpits without a position), the sound will play as a proper 3D sound

### DIFF
--- a/code/model/modelanimation_segments.cpp
+++ b/code/model/modelanimation_segments.cpp
@@ -1,5 +1,7 @@
 #include "model/modelanimation_segments.h"
 
+#include "render/3d.h"
+
 namespace animation {
 
 	ModelAnimationSegment* ModelAnimationSegmentSerial::copy() const {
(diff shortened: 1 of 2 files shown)


wookieejedi:
Tested and works great!

code/model/modelanimation_segments.cpp
```diff
@@ -1213,37 +1217,73 @@ namespace animation {
...
+		if (snd_is_playing(m_instances[pmi->id].currentlyPlaying))
+			snd_stop(m_instances[pmi->id].currentlyPlaying);
 
 		if (m_end.isValid()) 
-			m_instances[pmi_id].currentlyPlaying = snd_play(gamesnd_get_game_sound(m_end));
+			m_instances[pmi->id].currentlyPlaying = playSnd(pmi, m_end, false);
+	}
+
+	sound_handle ModelAnimationSegmentSoundDuring::playSnd(polymodel_instance* pmi, const gamesnd_id& sound, bool loop) {
+		if (m_submodel != nullptr && pmi->objnum > 0) {
```
Goober5000, line 1236: Should this be >= ?

BMagnu: As per object.h:
``int Object_next_signature = 1;	//0 is bogus, start at 1``
Similarly, the scripting API also considers objnum == 0 to be invalid.

Hence, I'd say that checking against > 0 in correct, but don't ask my why it was made that way.

BMagnu: right. Object _signatures_ are always > 0, but objnums can be 0, I think

Goober5000 (approved):

PR #4659, opened 2022-09-15
Fix retail UI crash when campaign missions are selected in simulator

qazwsxal:
Small oversight in the new UI code, wrong function was being called when building the list of campaign missions in the simulator.


The-E (approved):

JohnAFernandez (approved):
Thanks

PR #4662, opened 2022-09-16
Fix smart spawn flag

wookieejedi:
Courtesy of Lafiel, this code fixes special flags such as those related to `spawn`. (Bug arose in #4620).

Tested and fixes the problem and works as expected.


BMagnu (approved):
Yup, was a copy/paste issue of mine when I did that PR.
This is absolutely needed

PR #4663, opened 2022-09-16
lighting followup

Goober5000:
Two small fixes to follow up the lighting PRs

1) Change the mprintf to a Warning, both to be consistent with other deprecation warnings and to notify modders that they might need to accommodate the new lighting system
2) Parse any leftover data so that legacy tables don't cause FSO to choke on parsing

### DIFF
--- a/code/starfield/starfield.cpp
+++ b/code/starfield/starfield.cpp
@@ -472,8 +472,20 @@ void parse_startbl(const char *filename)
 				stuff_float(&sbm.b);
 				stuff_float(&sbm.i);
 
-				if (optional_string("$SunSpecularRGB:"))
-					mprintf(("Sun %s tried to set SunSpecularRGB. This feature has been deprecated and will be ignored.\n", sbm.filename));
+				if (optional_string("$SunSpecularRGB:")) {
+					SCP_string warning;
+					sprintf(warning, "Sun %s tried to set SunSpecularRGB. This feature has been deprecated and will be ignored.", sbm.filename);
+
+					float spec_r, spec_g, spec_b;
+					stuff_float(&spec_r);
+					stuff_float(&spec_g);
+					stuff_float(&spec_b);
+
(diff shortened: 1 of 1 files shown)


EatThePath (commented):
This was originally a warning and was changed based on reviewer feedback from @The-E so I think his input is needed before merging the warning change. The parsing is definitely a fair catch, though.

Just for context, when I looked through a few mods I couldn't find anyone using this feature fwiw. It was hardly a comprehensive survey, but given how niche and quirky this feature is I wouldn't be surprised if it's entirely unused. But that doesn't really change how it should be handled as a message, in principal.

Goober5000:
Well, Scroll used it.  Our background artist is going to investigate whether any changes are needed.  But we wouldn't have known about this particular aspect of the upgrade if the parser hadn't choked on the legacy values.  Ironically, that served the purpose in this case that a Warning would normally serve.

EatThePath:
Ah, I see. Well, if they have trouble with the new setup let me know I suppose.

The-E:
I requested that the Warning be changed to a log print because this is a case where the extra data doesn't do any harm as it's just ignored; this is something that we do in a lot of places already (game settings, model animations). Now, to be fair, we aren't consistent about this, there are many other places where we do warn for issues like this, but in this particular case the change that this is related to is already so huge that anyone who is this particular about the lighting in their mod will already need to revisit their missions anyway.

EatThePath:
In discussion about Scroll's backgrounds it turned out that the specular for the stars was being set, but it was being set to the same as the standard RGB. That's no different from the default case to begin with. So perhaps a warning could be contingient on checking if the specular is being set to anything meaningful, and ignore cases where the specular line was superflous to begin with. That can still be a log print, as it's a chance for table cleanup but has no real impact on the game state

Goober5000:
We are only this particular about the star lighting for one specific background in Scroll.  The fact that the overall lighting change is so huge is a reason to *keep* the warning, because that provides a clue as to what past assumptions are no longer applicable.

I can see the point about distinguishing between significant and insignificant values, so I've pushed some new changes.

EatThePath (approved):
Looks good to me

Issue #4637, opened 2022-09-10
Models disappear in Twisted Infinities surface mission


Goober5000:
This bug affects the TI mission "Sneak and Destroy", MoonSurface.fs2.  It's not part of the main campaign but can be found in the tech room.

For most of the player perspectives, the surface is not rendered - and not only the surface, but the player ship as well.  This can be seen in the attached video:
https://user-images.githubusercontent.com/1878523/189500047-319858fc-80e4-4618-bb7d-918077bd34e4.mp4

The bug appeared sometime between 22.0 and 22.2.

The-E:
Twisted Infinities uses a custom script to distribute and then render a bunch of rocks on the surface of that moon. When that script is disabled, everything renders normally; this indicates that the culprit is somewhere within the `gr.drawModelOOR` method.

The-E:
Looking at the game in renderdoc reveals that the script is actually executed several times per frame. If we put in a manual guard against that (by checking in the script itself if it ran this frame), it works as intended; the true cause of this issue might be a combination of the script running more often than intended and some wonkiness involving `gr.drawModelOOR`.

PR #4664, opened 2022-09-16
use queued rendering instead of immediate rendering for scripts

Goober5000:
Scripts that render models in the On Object Render hook were effectively doing immediate renders in the middle of queued renders.  This caused some data, such as model orientation, to 'leak' between renders.  It might also have something to do with why the bug in #4637 had the particular graphical side effects that it did.

Anyway, changing the immediate renders to queued renders fixes the leaky orientations, and is probably a good upgrade for other reasons as well.

### DIFF
--- a/code/object/object.cpp
+++ b/code/object/object.cpp
@@ -35,6 +35,7 @@
 #include "object/objectshield.h"
 #include "object/objectsnd.h"
 #include "observer/observer.h"
+#include "scripting/api/libs/graphics.h"
 #include "scripting/scripting.h"
 #include "playerman/player.h"
 #include "radar/radar.h"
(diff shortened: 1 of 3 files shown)


BMagnu:
Is it sufficient to "just add them to the queue"?
I'd imagine that the queue would need to be resorted, since we still need to ensure proper order, both in terms of shader switching efficiency, as well as draw order for any semi-transparent objects in the scene

Goober5000:
That is beyond my knowledge.  But just going by the usage elsewhere, I think this is sufficient.  Look at the code where external weapon models are rendered along with the ships they are mounted on: the functions just call `model_render_queue` for the additional models, without any apparent sorting or queue management.

EatThePath (approved):
This looks sensible enough to me. It might lead to a single frame lag, or perhaps more like a half frame, in drawing, but that's a negligible factor in my mind.

PR #4666, opened 2022-09-17
Fix subsystem health and comparison frames

JohnAFernandez:
This should fix health and *some* rotations appearing on the client.  Based on our conversations with Lafiel, it will be much more complicated to completely implement rotations.  So a complete rewrite for that system is (and of this part of the packet) is on the way.  As usual these days, this just needs a quick final test before it can be merged.


wookieejedi:
I'm going to give it a test on my local setup, too.

wookieejedi:
Tests show this works as expected.

code/network/multi_obj.cpp
```diff
@@ -1890,89 +1894,123 @@ int multi_oo_unpack_data(net_player* pl, ubyte* data, int seq_num, int time_delt
...
+						data_idx++;
+					}
 
-				if (flags[i] & OO_SUBSYS_TRANSLATION_y) {
-					subsysp->submodel_instance_1->canonical_prev_offset.xyz.y = subsysp->submodel_instance_1->canonical_offset.xyz.y;
-					subsysp->submodel_instance_1->canonical_offset.xyz.y = subsys_data[data_idx];
-					data_idx++;
-				}
+					if (flags[i] & OO_SUBSYS_ROTATION_1h) {
+						if (animations_valid) {
```
wookieejedi, line 1932: Super minor, but might not need all that extra whitespace to the right.

wookieejedi: image

wookieejedi (approved):
I’ve tested this thoroughly and it provides a very needed fix for players when attacking subsystems. Going over the code multiple times also looks good (minus some nitpicky whitespace), so it has my approval. Ideally anyone else interested will also have some time to look it over.

PR #4667, opened 2022-09-17
Flag for seamless animations with startup period

BMagnu:
This is a rather specific flag for a certain but common type of animation, namely seamlessly looping animations which have an initial startup from a different rest state to their seamless loop.
This flag acts as a loop + reset on startup flag combination which, instead of skipping back to the end, skips back to the part where the seamless loop begins, both in forwards and backwards direction. Once the loop is stopped, the animation will run until the next loop would begin, and then play the startup sequence in reverse.

### DIFF
--- a/code/model/modelanimation.cpp
+++ b/code/model/modelanimation.cpp
@@ -12,12 +12,15 @@ extern float flFrametime;
 
 namespace animation {
 
-	flag_def_list_new<animation::Animation_Flags> Animation_flags[] = {
-		{ "auto reverse",				animation::Animation_Flags::Auto_Reverse,						true, false },
-		{ "reset at completion",		animation::Animation_Flags::Reset_at_completion,		        true, false },
-		{ "loop",						animation::Animation_Flags::Loop,						        true, false },
-		{ "random starting phase",		animation::Animation_Flags::Random_starting_phase,				true, false },
-		{ "pause on reverse",			animation::Animation_Flags::Pause_on_reverse,					true, false },
+	special_flag_def_list_new<animation::Animation_Flags, ModelAnimation&> Animation_flags[] = {
+		{ "auto reverse",				animation::Animation_Flags::Auto_Reverse,						true },
+		{ "reset at completion",		animation::Animation_Flags::Reset_at_completion,		        true },
+		{ "loop",						animation::Animation_Flags::Loop,						        true },
+		{ "random starting phase",		animation::Animation_Flags::Random_starting_phase,				true },
+		{ "pause on reverse",			animation::Animation_Flags::Pause_on_reverse,					true },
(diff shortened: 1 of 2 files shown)


code/model/modelanimation.cpp
```diff
@@ -12,12 +12,20 @@ extern float flFrametime;
...
+		{ "loop",						animation::Animation_Flags::Loop,						        true },
+		{ "random starting phase",		animation::Animation_Flags::Random_starting_phase,				true },
+		{ "pause on reverse",			animation::Animation_Flags::Pause_on_reverse,					true },
+		{ "seamless with startup",		animation::Animation_Flags::Seamless_with_startup,				true, [](const SCP_string& from, ModelAnimation& anim) {
+			anim.m_flagData.loopsFrom = static_cast<float>(std::atof(from.c_str()));
+			anim.m_flags.set(animation::Animation_Flags::Loop);
+			anim.m_flags.set(animation::Animation_Flags::Seamless_with_startup);
+
+			//While this flag implies Reset_at_completion semantically, it's implemented in a conflicting manner. Hence, make sure that seamless will take priority, and force-disable reset at completion if set
+			anim.m_flags.set(animation::Animation_Flags::Reset_at_completion, false);
```
Goober5000, line 27: This works if the "reset at completion" flag comes first, but if it comes later, then this will not prevent both flags from being set.  So some sort of check should happen after all flags are parsed.

code/model/modelanimation.h
```diff
@@ -60,11 +60,13 @@ namespace animation {
 		Loop,					//Will automatically loop the animation once it completes. Is compatible with Reset_at_completion to loop back from the start instead of reversing. Incompatible with Auto_reverse
 		Random_starting_phase,  //When an animation is started from an untriggered state, will randomize its time to any possible time of the animation + possibly on the reverse, if the animation would automatically enter that
 		Pause_on_reverse,		//Will cause any start in RWD direction to behave as a call to pause the animation. Required (and also only really useful) when a looping animation is supposed to be triggered by an internal engine trigger
+		Seamless_with_startup,	//Provides autoamtic handling of animations that loop with an initialization part (effectively looping from a specific time)
```
Goober5000, line 63: "automatic" misspelled

Goober5000:
Two comments, plus a clang warning:
```

code/model/modelanimation.cpp
BMagnu: This works due to the way the parsing stuff does things.
First, all non-special flags are parsed and set. Afterwards, all special or unparsed flags are checked and their functor is run.
So as long as you only modify non-special flags, this kind of flag modification should be safe

Goober5000: Ah, ok.

Goober5000 (approved):

PR #4669, opened 2022-09-17
Allow movies to be alpha blended through the LUA Movie Player API

TRBlount:


Goober5000 (commented):
LGTM, and works in testing, though I'd like a graphics coder to give formal approval.

The-E (approved):

PR #4671, opened 2022-09-18
add -dualscanlines as a HUD config option

Goober5000:
Allow different types of cargo scans, including the -dualscanline behavior, to be specified in the HUD gauges table.  Right now the options are "default" and "dualscanlines".


JohnAFernandez:
Looks like this also deprecates the command line argument.

Goober5000:
It *could* deprecate the command-line argument.  I didn't explicitly do so; I just made it activate the option for all parsed HUDs.  But I can deprecate it if that makes sense.

wookieejedi (approved):
Looks good, plus the added bonus of disco scan lines!

JohnAFernandez (approved):

PR #4673, opened 2022-09-18
a couple of animation fixes

Goober5000:
Done at Lafiel's direction.


BMagnu (approved):
Required fixes for bugs likely caused by #3928 and followup PRs

PR #4676, opened 2022-09-19
Fix VS Warning

wookieejedi:
If this is trivial and not needed feel free to dismiss, but this one-line PR simply fixes a warning that keeps showing up when building with VS.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -10087,9 +10087,9 @@ int ship_create(matrix* orient, vec3d* pos, int ship_type, const char* ship_name
 		strcpy_s(shipp->ship_name, base_name);
 
 		// if generated name will be longer than allowable name, truncate the class section of the name by the overflow
-		int char_overflow = static_cast <int> (strlen(base_name) + strlen(suffix)) - (NAME_LENGTH - 1);
+		int char_overflow = (static_cast<int> (strlen(base_name)) + static_cast<int> (strlen(suffix))) - (NAME_LENGTH - 1);
 		if (char_overflow > 0) {
-			shipp->ship_name[strlen(base_name) - char_overflow] = '\0';
+			shipp->ship_name[static_cast<int> (strlen(base_name)) - char_overflow] = '\0';
 		}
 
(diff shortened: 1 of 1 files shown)


code/ship/ship.cpp
```diff
@@ -10087,7 +10087,7 @@ int ship_create(matrix* orient, vec3d* pos, int ship_type, const char* ship_name
 		strcpy_s(shipp->ship_name, base_name);
 
 		// if generated name will be longer than allowable name, truncate the class section of the name by the overflow
-		int char_overflow = (strlen(base_name) + strlen(suffix)) - (NAME_LENGTH - 1);
+		int char_overflow = static_cast <int> (strlen(base_name) + strlen(suffix)) - (NAME_LENGTH - 1);
```
JohnAFernandez, line 10090: I think it's a fine quick change to make. But static cast the individual strlen calls. m!m mentioned on a review in the past that putting more than one thing in a cast expression doesn't work as you would expect.

JohnAFernandez (dismissed):

code/ship/ship.cpp
wookieejedi: Thanks for the heads up, that is excellent to know. I've updated so each is cast individually.

Goober5000 (changes requested):
Let's make sure this is trhe appropriate fix.  What is the Visual Studio warning?

wookieejedi:
Good question, `'initializing': conversion from 'size_t' to 'int', possible loss of data`

Goober5000 (changes requested):
~~In that case, try removing the existing casts, but casting `char_overflow` to `size_t`.~~

Goober5000:
Try this instead:
```
int char_overflow = static_cast<int>(strlen(base_name) + strlen(suffix)) - (NAME_LENGTH - 1);
if (char_overflow > 0) {
	shipp->ship_name[strlen(base_name) - static_cast<size_t>(char_overflow)] = '\0';
}
```

Goober5000 (approved):

PR #4679, opened 2022-09-19
Allow API to optionally not start the next mission.

MjnMixael:
Changes these two methods to allow the API to specify whether to go straight to the next (or current) mission briefing. This allows implementation of the debriefing ESC popup menu as well as skipping to the next mission. The latter requires closing out the debriefing with one of these two options so that we avoid an assert upon entering the next debriefing.

Also removes the return value because it was meaningless.

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -885,7 +885,7 @@ ADE_FUNC(canSkip,
 
 ADE_FUNC(replayMission,
 	l_UserInterface_Debrief,
-	"[boolean restart = false]",
+	"[boolean restart = true]",
 	"Resets the mission outcome, and optionally restarts the mission at the briefing; "
 	"true to restart the mission, false to stay at current UI. Defaults to true.",
 	nullptr,
@@ -912,7 +912,7 @@ ADE_FUNC(replayMission,
 
 ADE_FUNC(acceptMission,
 	l_UserInterface_Debrief,
-	"[boolean start = false]",
+	"[boolean start = true]",
 	"Accepts the mission outcome, saves the stats, and optionally begins the next mission if in campaign; "
 	"true to start the next mission, false to stay at current UI. Defaults to true.",
 	nullptr,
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/ui.cpp
```diff
@@ -885,40 +885,57 @@ ADE_FUNC(canSkip,
 
 ADE_FUNC(replayMission,
 	l_UserInterface_Debrief,
+	"[boolean restart = false]",
```
Kiloku, line 888: This should say "restart = true" to match the default

MjnMixael: Right. Done.

code/scripting/api/libs/ui.cpp
```diff
@@ -885,40 +885,57 @@ ADE_FUNC(canSkip,
...
 
-	gameseq_post_event(GS_EVENT_START_GAME);
+	if (restart) {
+		gameseq_post_event(GS_EVENT_START_GAME);
+	}
 
 	API_Access = false;
 
-	return ade_set_args(L, "b", true);
+	return ADE_RETURN_NIL;
```
Kiloku, line 910: Is it okay to change the return type? A previous script that used this would possibly break, but I'm not sure if it's really something that needs to be checked (in which case other scripts probably never used it)

code/scripting/api/libs/ui.cpp
```diff
@@ -885,40 +885,57 @@ ADE_FUNC(canSkip,
...
 }
 
 ADE_FUNC(acceptMission,
 	l_UserInterface_Debrief,
+	"[boolean start = false]",
```
Kiloku: Same as before, the actual default value is `true`.

code/scripting/api/libs/ui.cpp
MjnMixael: Unless someone wrote a script in the last day other than me, then yes, it's OK to change.

Kiloku (approved):

Issue #4665, opened 2022-09-16
FR/objective list: Enhanced scripting development experince


EatThePath:
It'd be really nice if there were a means to do scripting development without restarting the game for every change, and the blingblong of error message popups haunts my nightmares. There are a number of things that I think taken all together would enable some tantalizing workflow possibilities, so I'm making one issue right now instead of shotgunning a bunch of small ones into the tracker.

- [ ] lua eval/ REPL on debug console, with automatic session logging 
- [ ] Debug console in detached window, usable with or without game paused.
- [ ] Lua errors and warnings piped to console (optionally?) with no popup or pause.
- [ ] Lua `error()` calls being non-fatal events.
- [ ] API functions to check current hooks and remove existing hooks

If all that were in place I think effective REPL style development would be in reach with a bit of supporting script design.

PR #4681, opened 2022-09-19
-luadev flag and lua log filter

EatThePath:
Adds new default 'lua' filter category for all lua errors and output and -luadev flag that currently only makes lua errors non-fatal.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -512,6 +512,7 @@ cmdline_parm graphics_debug_output_arg("-gr_debug", nullptr, AT_NONE); // Cmdlin
 cmdline_parm log_to_stdout_arg("-stdout_log", nullptr, AT_NONE); // Cmdline_log_to_stdout
 cmdline_parm slow_frames_ok_arg("-slow_frames_ok", nullptr, AT_NONE);	// Cmdline_slow_frames_ok
 cmdline_parm fixed_seed_rand("-seed", nullptr, AT_INT);	// Cmdline_rng_seed,Cmdline_reuse_rng_seed;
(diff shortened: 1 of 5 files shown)


EatThePath:
Implements two points of #4665

code/cmdline/cmdline.cpp
```diff
@@ -512,6 +512,7 @@ cmdline_parm graphics_debug_output_arg("-gr_debug", nullptr, AT_NONE); // Cmdlin
 cmdline_parm log_to_stdout_arg("-stdout_log", nullptr, AT_NONE); // Cmdline_log_to_stdout
 cmdline_parm slow_frames_ok_arg("-slow_frames_ok", nullptr, AT_NONE);	// Cmdline_slow_frames_ok
 cmdline_parm fixed_seed_rand("-seed", nullptr, AT_INT);	// Cmdline_rng_seed,Cmdline_reuse_rng_seed;
+cmdline_parm luadev_arg("-luadev", "nullptr", AT_NONE);	// Cmdline_slow_frames_ok
```
BMagnu, line 515: That comment seems slightly outdated...

BMagnu (changes requested):
Mostly good.
Also, do I correctly understand this, that it will keep printing lua to the debug log without any special logging config?

code/cmdline/cmdline.cpp
EatThePath: Whoops, I'll update that...

EatThePath:
> Mostly good. Also, do I correctly understand this, that it will keep printing lua to the debug log without any special logging config?

Yeah, I tested to make sure that it defaults to lua on if not specified.

BMagnu (approved):

PR #4682, opened 2022-09-20
Mission time

MjnMixael:
Add isInMission and changes getMissionTime to always return the mission time even outside of a mission

### DIFF
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1368,7 +1368,7 @@ ADE_VIRTVAR(ShudderIntensity, l_Mission, "number", "Gets or sets the shudder int
 
 ADE_FUNC(isInMission, l_Mission, nullptr, "get whether or not a mission is currently being played", "boolean", "true if in mission, false otherwise")
 {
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/mission.cpp
```diff
@@ -1369,6 +1366,11 @@ ADE_VIRTVAR(ShudderIntensity, l_Mission, "number", "Gets or sets the shudder int
 	return ade_set_args(L, "f", intensity);
 }
 
+ADE_FUNC(isInMission, l_Mission, nullptr, "get whether or not a mission is currently being played", "boolean", "true if in mission, false otherwise")
+{
+	return ade_set_args(L, "b", (Game_mode & GM_IN_MISSION));
```
Goober5000, line 1371: Since `(Game_mode & GM_IN_MISSION)` is a varags parameter, it should be explicitly made a bool.  Either cast it or add `!= 0`.

Goober5000 (changes requested):

code/scripting/api/libs/mission.cpp
MjnMixael: Done

Goober5000 (approved):

PR #4683, opened 2022-09-20
Allow modifying map container data when the map is an argument

jg18:
Allow a map container's data for existing keys to be modified when the map container is used in the special argument list.

Adding or removing keys is still prohibited in this situation due to concurrent modification.


JohnAFernandez (approved):

JohnAFernandez:
Still looks good, after adding that fix.

Goober5000 (commented):
Belatedly, this looks good.

PR #4684, opened 2022-09-20
add setFlag and getFlag for ships and parse objects

Goober5000:
This mirrors alter-ship-flag and are-ship-flags-set.

~~Currently in draft pending testing.~~  After testing (with valuable assistance from @wookieejedi) and some additional features including arbitrary arguments and a parse-object indexer, this is ready for review.


wookieejedi (approved):
A highly useful scripting exposure PR. I've tested set and get flags for both ships and parsed ships and everything works well. Overall looks good to me.

PR #4685, opened 2022-09-20
Update max priority for scripting AIgoals

wookieejedi:
Previously the scripting function `giveOrder` capped goals at priority 100, but the max is 200. This PR fixes that.


EatThePath (approved):
lgtm

JohnAFernandez (approved):
Agreed.

PR #4690, opened 2022-09-20
Adds a cache for modular POFs

BMagnu:
Previously, all loads for modular pofs were continually repeated. This changes this to cache models loaded from disk.
Since such a cache implies that the base models loaded from disk are immutable and reusable, some access had to be refactored. This consists mostly of macroifying the change_submodule_number (and name) functions, so that they provide methods for both in-place editing (when they need to operate on the current, mutable base model) as well as for editing an rvalue temporary copy of the data from the cache to be put moved into the current virtual pof with minimal code repetition. The templates in the macro may look scary, but all they do is allow any kind of map-like for providing the replacement arguments over just one variety of SCP_map.
This PR is a requirement for the followup-additions of features for modular pofs, as the changed API access has significant impact on further code (hence I also added caching much earlier than anticipated; it'd just get very very hard to add it after the fact if none of the modules takes the immutability of data in the cache into account)

### DIFF
--- a/code/model/modelreplace.cpp
+++ b/code/model/modelreplace.cpp
@@ -387,8 +387,6 @@ void VirtualPOFOperationAddSubmodel::process(polymodel* pm, model_read_deferred_
 				pm->maps[usedTexture.second] = appendingPM->maps[usedTexture.first];
 				appendingPMholder->keepTexture(usedTexture.first);
 			}
-
-			VirtualPOFOperationAddTurret::addTurretToPM(pm, deferredTasks, 0, appendingPM, appendingSubsys);
 		}
 		else {
 			Warning(LOCATION, "Failed to add submodel %s of POF %s to virtual POF %s, original POF already has a subsystem with the same name as was supposed to be added.", subobjNameSrc.c_str(), appendingPOF.c_str(), virtualPof.name.c_str());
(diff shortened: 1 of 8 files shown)


code/model/modelreplace.cpp
```diff
@@ -122,26 +190,77 @@ void virtual_pof_init() {
...
-		for (auto& debris : submodel.live_debris)
-			REPLACE_IF_EQ(debris);
-		REPLACE_IF_EQ(submodel.look_at_submodel);
-		REPLACE_IF_EQ(submodel.my_replacement);
-		REPLACE_IF_EQ(submodel.next_sibling);
-		REPLACE_IF_EQ(submodel.parent);
-	}
-}
-
+#define REPLACE_IF_STRIEQ(data) if (stricmp(data, source.c_str()) == 0) { strncpy(data, dest.c_str(), dest.size()); data[dest.size()] = '\0'; }
```
Goober5000, line 193: why not use `strcpy_s`?

code/model/modelreplace.cpp
```diff
@@ -122,26 +190,77 @@ void virtual_pof_init() {
...
+
+//Generates one function for replacing data in a type, which is a map entry of which the key may be replaced. Takes an rvalue reference, used for making a copy and modifying the tempoary to then assign it somewhere
```
Goober5000, line 195: "temporary" misspelled

code/model/modelreplace.cpp
```diff
@@ -122,26 +190,77 @@ void virtual_pof_init() {
...
+#define CHANGE_HELPER_MAP_KEY(name, intype, argtype) template<typename map_t> static typename std::enable_if<std::is_same<typename map_t::value_type, std::pair<const argtype, argtype>>::value, intype>::type name(intype&& pass, map_t replace){ \
+	const auto it = replace.find(pass.first); \
+	intype input = { (it == replace.end() ? pass.first : it->second), pass.second }; \
+	for (const auto& replacement : replace) { \
+		const argtype& source = replacement.first; \
+		const argtype& dest = replacement.second;
+#define CHANGE_HELPER_MAP_KEY_END } return input; }
+
+//Generates two functions for replacing data in a type. One that takes an rvalue reference, used for making a copy and modifying the tempoary to then assign it somewhere, and one which takes an lvalue reference for modifying in-place
```
Goober5000, line 204: "temporary"

code/model/modelreplace.cpp
BMagnu: Added a proper strncpy_s here

Goober5000 (approved):

PR #4694, opened 2022-09-21
Scpui ship select

MjnMixael:
Gives Lua access to the loadout wing slots. I had considerable trouble with this one due to lack of coding skill. Open to suggestions.

This is the first PR of several required to get ship and weapon select fully functional.

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -1093,7 +1093,7 @@ ADE_INDEXER(l_Loadout_Wings, //
 {
 	int idx;
 	if (!ade_get_args(L, "*i", &idx))
-		ade_set_error(L, "o", l_Loadout_Wing.Set(ss_wing_info_h()));
+		return ade_set_error(L, "o", l_Loadout_Wing.Set(ss_wing_info_h()));
 	idx--; //Convert to Lua's 1 based index system
 	return ade_set_args(L, "o", l_Loadout_Wing.Set(ss_wing_info_h(idx)));
 }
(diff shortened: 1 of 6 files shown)


code/scripting/api/libs/ui.cpp
```diff
@@ -1036,6 +1038,71 @@ ADE_FUNC(getFictionMusicName, l_UserInterface_FictionViewer, nullptr,
...
+ADE_LIB_DERIV(l_Loadout_Wings, "Loadout_Wings", nullptr, nullptr, l_UserInterface_ShipWepSelect);
+ADE_INDEXER(l_Loadout_Wings, //
+	"number Index",
+	"Array of loadout wing data",
+	"loadout_wing",
+	"loadout handle, or invalid handle if index is invalid")
+{
+	int idx;
+	if (!ade_get_args(L, "*i", &idx))
+		return ade_set_error(L, "s", "");
```
BMagnu, line 1096: Unlike what the docs say, you're not returning an invalid handle (which'd be ``ade_set_error(L,"o", l_Loadout_Wing.Set(ss_wing_info_h()));``), but an empty string

BMagnu (changes requested):
One small comment

code/scripting/api/libs/ui.cpp
MjnMixael: Fixed

BMagnu: I'm reasonably sure that you still have to _return_ the ``ade_set_error`` there, not just call it.
But yeah, that'll be doing a proper invalid handle

MjnMixael: Whoops. Didn't intend to replace the return. Fixed

BMagnu (approved):

PR #4698, opened 2022-09-22
Scpui loadout pools

MjnMixael:
API access to the ship and weapon loadout pools. Relies on #4694

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -1185,8 +1185,36 @@ ADE_FUNC(initSelect,
 	return ADE_RETURN_NIL;
 }
 
+ADE_FUNC(getNumShipsAvail,
+	l_UserInterface_ShipWepSelect,
+	"number Index",
+	"Gets the amount of a ship available. Index is index into Ship Classes.",
+	"number",
+	"amaount available")
+{
+	int idx;
+	if (!ade_get_args(L, "*i", &idx))
+		return ADE_RETURN_NIL;
+	idx--; // Convert to Lua's 1 based index system
+	return ade_set_args(L, "i", Ss_pool[idx]);
+}
+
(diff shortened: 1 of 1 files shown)


BMagnu (changes requested):
Hmm, I think it'd be better if you could pass Ship_class / Weapon_class userdata objects here instead of just an int. You should be able to easily get the index from those on the C++ side, and it'd easily allow to get shipclasses by name or from specific objects for example for these methods.

MjnMixael:
That seems like unnecessary complexity for this specific case, though. These values are only valid after running initSelect() for the current mission. Seems like the script should do exactly what FSO does here; iterate over the entire list of weapons to build the icon list once. I don't see much use in trying to do that arbitrarily by name or object.

MjnMixael:
> Hmm, I think it'd be better if you could pass Ship_class / Weapon_class userdata objects here instead of just an int. You should be able to easily get the index from those on the C++ side, and it'd easily allow to get shipclasses by name or from specific objects for example for these methods.

That seems like unnecessary complexity for this specific case, though. These values are only valid after running initSelect() for the current mission. Seems like the script should do exactly what FSO does here; iterate over the entire list of weapons to build the icon list once. I don't see much use in trying to do that arbitrarily by name or object.

BMagnu (approved):
Not too much a fan of the resulting API inconsistency to other places that need ship / weapon classes, but fair enough, regarding this specific use case

MjnMixael:
Bah. Hold off on merging this. FSO writes to the pool directly as weapons and ships are out into the loadout and that is recalled if player_loadout is saved. I'm going to have to redo these as indexers so they can be written to also.

MjnMixael:
Sorry to have you review this only to realize that I needed to change it. It's ready for real this time. Changed it from Func to Indexer and made the pools writable. Tested and working.

BMagnu (approved):

PR #4699, opened 2022-09-22
Fix lighting flickering on specmap-less models

qazwsxal:
Big shader PR #4652 made models always receive image based lighting (bar command line options) - even those without specular maps. However, the associated shader_flags logic wasn't updated. this meant that ships without specular maps would render environment lighting with uninitialized data describing its orientation. This resulted in flickering as the data would differ frame-to-frame. 

This PR moves the logic outside of the "if has specular map" conditional, making sure the data is initialised if environment map rendering is happening


wookieejedi (approved):

JohnAFernandez (approved):
Ah, now it's making sense.  Good detective work!

PR #4701, opened 2022-09-22
Prevent dangling threads from crashing FSO on close

BMagnu:
If we open a thread with async.run, we lose complete control over it.
This means, that when closing the scripting state, we don't actually take the thread with us.
To prevent the C++ side of things to try to push the thread off of the lua state way after the state is already dead, we remember all threads, and once the state gets destroyed, we forcibly push every thread handle off the stack. This has twofold effects: One, the C-Side of the handle gets invalidated, causing the destruction of the Thread object to not try to modify the lua stack, and two, cause the thread to be properly garbage collected on the lua side of things during deallocation.

As a sidenote, there is no built in way to stop threads in lua. They are an object like any other, and will be GC'd once they have no references left. I'm not entirely sure what the interpreter does with the _actual_ system thread it started, and when it'll be closed. But the final lua_close of the main thread should kill it at the latest.

This is very much not a pretty way of doing things, but unless we have other ways of pinging each thread object and invalidating the UniqueLuaReference they hold, we will keep crashing if a rouge thread survives the scripting environment.
The other alternative is to hard-disallow infinite loops in async runs, and then wait for each open thread on close of the script environment.

### DIFF
--- a/code/scripting/lua/LuaTable.cpp
+++ b/code/scripting/lua/LuaTable.cpp
@@ -17,9 +17,6 @@ LuaTable LuaTable::create(lua_State* state) {
 LuaTable::LuaTable() : LuaValue() {
 }
 
-LuaTable::LuaTable(const LuaTable& other) : LuaValue(other) {
-}
-
 LuaTable::~LuaTable() {
 }
 
--- a/code/scripting/lua/LuaTable.h
+++ b/code/scripting/lua/LuaTable.h
@@ -100,12 +100,6 @@ class LuaTable: public LuaValue {
 	LuaTable();
 
 	/**
-     * @brief Copy-constructor
-     * @param other The other table.
-     */
-	LuaTable(const LuaTable& other);
-
-	/**
      * Dereferences the stored reference to the table if it exists.
      */
 	~LuaTable() override;
--- a/code/scripting/lua/LuaThread.cpp
+++ b/code/scripting/lua/LuaThread.cpp
@@ -21,7 +21,7 @@ LuaThread LuaThread::create(lua_State* L, const LuaFunction& func)
 	int stack = lua_gettop(L);
 
 	//Make sure to create the function that clears the LuaThread reference BEFORE creating the userdata value, otherwise the function will be garbage-collected itself before it's called.
-	thread.deleterFunc = LuaFunction::createFromStdFunction(L, [threadRef](lua_State* L, const LuaValueList& params) -> LuaValueList {
+	thread.deleterFunc = LuaFunction::createFromStdFunction(L, [threadRef](lua_State*, const LuaValueList&) -> LuaValueList {
 		if(!threadRef.expired())
 			threadRef.lock()->removeReference();
 		return {};
@@ -53,8 +53,8 @@ LuaThread LuaThread::create(lua_State* L, const LuaFunction& func)
 LuaThread::LuaThread() = default;
 LuaThread::LuaThread(lua_State* luaState, lua_State* thread) : LuaValue(luaState), _thread(thread) {}
 
-LuaThread::LuaThread(LuaThread&& other) noexcept = default;
-LuaThread& LuaThread::operator=(LuaThread&& other) noexcept = default;
+LuaThread::LuaThread(LuaThread&&) = default;
+LuaThread& LuaThread::operator=(LuaThread&&) = default;
 
 LuaThread::~LuaThread() = default;
 
--- a/code/scripting/lua/LuaThread.cpp
+++ b/code/scripting/lua/LuaThread.cpp
@@ -8,14 +8,19 @@
 
 namespace luacpp {
 
-SCP_unordered_set<LuaThread*> LuaThread::threads;
+SCP_unordered_map<lua_State*, SCP_unordered_set<LuaThread*>> LuaThread::threads;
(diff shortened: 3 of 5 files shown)


EatThePath:
Isn't disallowing infinite loops actually the halting problem?

BMagnu:
I meant that as a documentation thingy; Write in the docs of async run that you may not code infinite loops, or you _will_ crash FSO with this.

EatThePath:
> I meant that as a documentation thingy; Write in the docs of async run that you may not code infinite loops, or you _will_ crash FSO with this.

Ah, that makes more sense then.

asarium:
I do not have enough context to know what exactly is wrong here but I saw several usages of the term "system thread" and just wanted to clarify that a "LuaThread" has nothing to do with an OS level thread. They are coroutines that have their own stack but they very much run on the "standard" game thread and do not hold a reference to a OS thread.

BMagnu:
> I do not have enough context to know what exactly is wrong here but I saw several usages of the term "system thread" and just wanted to clarify that a "LuaThread" has nothing to do with an OS level thread. They are coroutines that have their own stack but they very much run on the "standard" game thread and do not hold a reference to a OS thread.

Huh okay. Thought that lua_newthread actually made a seperate thread.
Still, this is a needed and valid solution.
The TL;DR for what is wrong is that a running coroutine has a reference to its own shared pointer and will thus keep alive. So far so good, but that means that the LuaThread object, which is a LuaValue and thus holds a UniqueLuaReference, will also keep alive. Even when the lua_State is already long deallocated. To solve this, we manually keep reference to all threads, and invalidate their UniqueLuaReference once the parent lua_State dies so that when the LuaThread object finally deallocates, it won't try to remove things off of the long invalid lua_State

asarium:
Another option could be to store a lua user data value with a `__gc` function for every thread into the Lua reference table and use that to clean up the needed references. The existing Lua API system should be capable of supporting that.
This solution probably works but introduces a reference to our scripting state system where there was none before which decreases the modularity of the code.

BMagnu:
I tried that with the following code, but it seems to break thread execution (it results in thread resumption seemingly trying to call the created table object instead of the thread function) and I don't fully understand why.
```cpp
auto threadRef = std::weak_ptr<UniqueLuaReference>(thread.getReference());

thread.deleterTable = LuaTable::create(luaThread);
thread.deleterFunc = LuaFunction::createFromStdFunction(luaThread, [threadRef](lua_State* L, const LuaValueList& params) -> LuaValueList {
	if(!threadRef.expired())
		threadRef.lock()->removeReference();
	return {};
	});
thread.deleterTable.pushValue(luaThread);
thread.deleterFunc.pushValue(luaThread);
lua_setfield(luaThread, -2, "__gc");

lua_pop(L, 2);
```
If I am to make this work using internal lua userdata modification, I will need some help with this, given that I don't really know the Lua API well.
On the other hand though: Yes, the existing solution decreases modularity, but really, do we need that here? When will we ever have a situation where we create a Lua thread that's outside of our scripting system...

asarium:
In order to use `__gc` on a thread you need to do the following:

1. Create a lua table
2. Set the `__gc` field of that table to the destructor function
3. Push the table and the thread on the Lua stack
4. Call `lua_setmetatable` to set the pushed table as the metatable of the thread

As far as I can tell, that should work for threads and should be called when the thread is cleaned up when the Lua state is closed since that should clean up all allocated storage.

If that does not work then the current changes are also fine but the current code could be considered hacky so there should be some effort to investigate alternative solutions first.

BMagnu:
This doesn't seem to work.
https://www.lua.org/pil/29.html
"Lua provides finalizers in the form of the __gc metamethod. This metamethod only works for userdata values."
I can try to make a dummy userdata object, maybe that'll work. And if I make it on the child thread, it'll get cleaned when that thread dies, maybe.

asarium:
Hmm, that is unfortunate. I thought that worked with all types.
One of the big issues I had with the object lifetimes in the async code was that nothing would keep a reference around to the coroutine once it has been started so the code always stores a strong pointer to itself in the callbacks to ensure the object stays alive.

However, it might be possible to remove that issue by keeping the reference to the `resolve_context` around via a userdata object as you suggested. Then `run_resolve_context` could be changed to only keep around weak pointers since the userdata would be responsible for keeping it alive while the coroutine still exists.

To fix the issue at hand the current code is sufficient but please continue looking into this since using the garbage collector is the cleaner, and arguably more correct solution for this problem.

asarium (approved):

BMagnu:
Okay. Once I have time, I'll test the custom userdata thing.
In the meantime, we should merge this so that we don't create crashes on shutdown anymore

BMagnu:
Okay, I managed to figure out how to make it work using the Lua GC. 
The dummy userdata object will be cleared when a thread dies, in which case the weak ref of the lambda will be invalid and it'll just quit quietly, or when the parent thread dies and deallocs its references, in which case the C-side Reference of the Thread will be invalidated

BMagnu:
A _really_ pretty way would be to redo it by keeping resolve_context as a real userdata value, but rewriting it to incorporate that is currently beyond the level of time I have to invest in this, I think.

The-E (approved):

PR #4702, opened 2022-09-23
Moved more things to debug filters, and associated cleanup

EatThePath:
This is a number of items that to me practically never have general log value and hide things that do have value. So, if anything shouldn't be hidden away like this, by all means say so

### DIFF
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -1272,7 +1272,7 @@ bool bm_load_and_parse_eff(const char *filename, int dir_type, int *nframes, int
 	}
 	catch (const parse::ParseException& e)
 	{
-		nprintf(("BmpMan","BMPMAN: Unable to parse '%s'!  Error message = %s.\n", filename, e.what()));
+		mprintf(("BMPMAN: Unable to parse '%s'!  Error message = %s.\n", filename, e.what()));
 		unpause_parse();
 		return false;
 	}
@@ -1291,13 +1291,13 @@ bool bm_load_and_parse_eff(const char *filename, int dir_type, int *nframes, int
 	} else if (!stricmp(NOX("pcx"), ext)) {
 		c_type = BM_TYPE_PCX;
 	} else {
-		nprintf(("BmpMan","BMPMAN: Unknown file type in EFF parse!\n"));
+		mprintf(("BMPMAN: Unknown file type in EFF parse!\n"));
 		return false;
 	}
 
 	// did we do anything?
 	if (c_type == BM_TYPE_NONE || frames == 0) {
-		nprintf(("BmpMan","BMPMAN: EFF parse ERROR!\n"));
+		mprintf(("BMPMAN: EFF parse ERROR!\n"));
 		return false;
 	}
 
@@ -1526,7 +1526,7 @@ int bm_load_animation(const char *real_filename, int *nframes, int *fps, int *ke
 	// it's an effect file, any readable image type with eff being txt
 	if (type == BM_TYPE_EFF) {
 		if (!bm_load_and_parse_eff(filename, dir_type, &anim_frames, &anim_fps, &key, &eff_type)) {
-			nprintf(("BmpMan","BMPMAN: Error reading EFF\n"));
+			mprintf(("BMPMAN: Error reading EFF\n"));
 			if (img_cfp != nullptr)
 				cfclose(img_cfp);
 			return -1;
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -1272,7 +1272,7 @@ bool bm_load_and_parse_eff(const char *filename, int dir_type, int *nframes, int
 	}
 	catch (const parse::ParseException& e)
 	{
-		mprintf(("BMPMAN: Unable to parse '%s'!  Error message = %s.\n", filename, e.what()));
+		nprintf(("BmpMan","BMPMAN: Unable to parse '%s'!  Error message = %s.\n", filename, e.what()));
 		unpause_parse();
 		return false;
 	}
@@ -1291,13 +1291,13 @@ bool bm_load_and_parse_eff(const char *filename, int dir_type, int *nframes, int
 	} else if (!stricmp(NOX("pcx"), ext)) {
 		c_type = BM_TYPE_PCX;
 	} else {
-		mprintf(("BMPMAN: Unknown file type in EFF parse!\n"));
+		nprintf(("BmpMan","BMPMAN: Unknown file type in EFF parse!\n"));
 		return false;
 	}
 
 	// did we do anything?
 	if (c_type == BM_TYPE_NONE || frames == 0) {
-		mprintf(("BMPMAN: EFF parse ERROR!\n"));
+		nprintf(("BmpMan","BMPMAN: EFF parse ERROR!\n"));
 		return false;
 	}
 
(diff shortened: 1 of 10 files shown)


wookieejedi:
I use the parsing model and sub model (and which is detail level) along with converting sounds myself, though not sure how common that is.

Goober5000 (changes requested):
Since you're doing associated cleanup, you might want to change your "nullptr" for -luadev.

Goober5000 (changes requested):
Also your `nprintf` commands in `ADE_FUNC(warning` and `ADE_FUNC(error` need newlines.

code/bmpman/bmpman.cpp
```diff
@@ -1272,7 +1272,7 @@ bool bm_load_and_parse_eff(const char *filename, int dir_type, int *nframes, int
 	}
 	catch (const parse::ParseException& e)
 	{
-		mprintf(("BMPMAN: Unable to parse '%s'!  Error message = %s.\n", filename, e.what()));
+		nprintf(("BpMan","BMPMAN: Unable to parse '%s'!  Error message = %s.\n", filename, e.what()));
```
Goober5000, line 1275: This abbreviation should be "BmpMan", not "BpMan", surely.

Though backing up a bit, if these BmpMan messages are legit errors, why hide them?

code/bmpman/bmpman.cpp
```diff
@@ -1526,12 +1526,12 @@ int bm_load_animation(const char *real_filename, int *nframes, int *fps, int *ke
...
 	if (type == BM_TYPE_EFF) {
 		if (!bm_load_and_parse_eff(filename, dir_type, &anim_frames, &anim_fps, &key, &eff_type)) {
-			mprintf(("BMPMAN: Error reading EFF\n"));
+			nprintf(("BpMan","BMPMAN: Error reading EFF\n"));
 			if (img_cfp != nullptr)
 				cfclose(img_cfp);
 			return -1;
 		} else {
-			mprintf(("BMPMAN: Found EFF (%s) with %d frames at %d fps.\n", filename, anim_frames, anim_fps));
+			nprintf(("BMPMAN: Found EFF (%s) with %d frames at %d fps.\n", filename, anim_frames, anim_fps));
```
Goober5000, line 1534: This `nprintf` is missing a category

code/model/modelread.cpp
```diff
@@ -3211,7 +3211,7 @@ int model_load(const  char* filename, int n_subsystems, model_subsystem* subsyst
 
 	TRACE_SCOPE(tracing::LoadModelFile);
 
-	mprintf(( "Loading model '%s' into slot '%i'\n", filename, num ));
+	nprintf(("Model" "Loading model '%s' into slot '%i'\n", filename, num ));
```
Goober5000, line 3214: This `nprintf` is missing a comma, which effectively means it's missing a category

code/model/modelread.cpp
```diff
@@ -3396,7 +3396,7 @@ int model_load(const  char* filename, int n_subsystems, model_subsystem* subsyst
 					dl2--;	// Start from 1 up...
 					if (dl2 >= sm1->num_details ) sm1->num_details = dl2+1;
 					sm1->details[dl2] = j;
-  				    mprintf(( "Submodel '%s' is detail level %d of '%s'\n", sm2->name, dl2 + 1, sm1->name ));
+  				    nprintf(("Model" "Submodel '%s' is detail level %d of '%s'\n", sm2->name, dl2 + 1, sm1->name ));
```
Goober5000, line 3399: This one too

code/model/modelread.cpp
EatThePath: I'm shocked that compiles honestly

EatThePath:
> I use the parsing model and sub model (and which is detail level) along with converting sounds myself, though not sure how common that is.

if it matters, I didn't touch any messages about converting sounds in general, just the retail numbered sounds.

wookieejedi:
> > I use the parsing model and sub model (and which is detail level) along with converting sounds myself, though not sure how common that is.
> 
> if it matters, I didn't touch any messages about converting sounds in general, just the retail numbered sounds.

Ah that makes sense, thanks!

Goober5000 (commented):
Looks mostly good now, but my question remains: Are some of these legit errors?  If they are errors, not just spurious messages, they should be kept.

EatThePath:
> Looks mostly good now, but my question remains: Are some of these legit errors? If they are errors, not just spurious messages, they should be kept.

I agree. I may have been a bit hasty in places, I'll do my best to give it another look through over the weekend and move anything serious back out.

code/bmpman/bmpman.cpp
The-E: I would agree - if something critical happens, like an asset not being loaded, then that should be in the normal log.

EatThePath: Latest pushes hopefully cover both these issues.

Goober5000 (approved):
Looks good now.

code/model/modelread.cpp
```diff
@@ -3403,7 +3403,7 @@ int model_load(const  char* filename, int n_subsystems, model_subsystem* subsyst
 					dl2--;	// Start from 1 up...
 					if (dl2 >= sm1->num_details ) sm1->num_details = dl2+1;
 					sm1->details[dl2] = j;
-  				    mprintf(( "Submodel '%s' is detail level %d of '%s'\n", sm2->name, dl2 + 1, sm1->name ));
+  				    nprintf(("Model", "Submodel '%s' is detail level %d of '%s'\n", sm2->name, dl2 + 1, sm1->name ));
```
BMagnu, line 3406: Not sure how I feel about this one being hidden.
Accidental LOD-ing is _the_ most common cause for submodels misbehaving I have seen, and it being hidden makes it a lot harder to spot on a glance from logs from non-tech-savvy people.

Goober5000: Good point

code/bmpman/bmpman.cpp
```diff
@@ -1600,7 +1600,7 @@ int bm_load_animation(const char *real_filename, int *nframes, int *fps, int *ke
 				throw e;
 			}
 			else {
-				mprintf(("Failed to load apng: %s\n", e.what()));
+				nprintf(("apng","Failed to load apng: %s\n", e.what()));
```
Goober5000, line 1603: This should be displayed

code/graphics/generic.cpp
```diff
@@ -217,7 +217,7 @@ int generic_anim_stream(generic_anim *ga, const bool cache)
 				ga->png.anim = new apng::apng_ani(ga->filename, cache);
 			}
 			catch (const apng::ApngException& e) {
-				mprintf(("Failed to load apng: %s\n", e.what() ));
+				nprintf(("apng","Failed to load apng: %s\n", e.what() ));
```
Goober5000, line 220: This should be displayed

code/model/modelread.cpp
Goober5000, line 3406: This should be displayed

Goober5000 (changes requested):
On second thought, three more messages should be in the log by default, including the one @BMagnu pointed out

code/model/modelread.cpp
EatThePath: Alright. I think the message is awkwardly formatted and confusing in isolation, but that's a seperate issue. I will revert.

code/graphics/generic.cpp
EatThePath: I have a minor vendeta against these apng messages... in my most recent level load in WM I got almost 200 lines of this complaining about *glow maps* that are all valid. This was supposedly fixed after I filed it for #3644, it's the exact smage message, but I can't remember if I verified the fix at the time or if this was just working as intended.

I can restore them I guess, if they're ever actually useful for casual debugging.

Goober5000 (approved):

PR #4705, opened 2022-09-23
Merge Show Ship and Cockpit Renderpasses

BMagnu:
The first followup to #3445.
This removes the two different functions for external ship model and cockpit rendering, and instead fuses them into one function.
Apart from the evidently higher maintainability, this has a few advantages:
1. Due to the shipmodel now being in the deferred cockpit pass, ship model and cockpit can shadow one another
2. The external ship model is now also rendered deferred at no additional cost.
3. The external ship model is rendered, like the cockpit, with a 0,0,0 view matrix which avoids the stuttering issues observed for close models like show ship / cockpit when far from the origin.
4. It allows for finer control over the relation between cockpit and show ship model as well as their render settings

### DIFF
--- a/code/graphics/opengl/gropengldeferred.cpp
+++ b/code/graphics/opengl/gropengldeferred.cpp
@@ -143,6 +143,8 @@ void gr_opengl_deferred_lighting_finish()
 	{
 		GR_DEBUG_SCOPE("Build buffer data");
 
+		auto lp = lighting_profile::current();
+
 		auto header = uniformAligner.getHeader<deferred_global_data>();
 		if (Shadow_quality != ShadowQuality::Disabled) {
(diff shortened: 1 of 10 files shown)


wookieejedi:
Overall great to have this 2nd follow-up PR ready for review, following in the 3 step plan that was decided regarding the plan for deferred cockpit/show ship rendering. 

Heavily tested and this PR works very well. The mod-controls are very useful and the multiplier value for the deferred pass allows mods to achieve visuals deferred distinctions between interior cockpit/show ship and exterior ship that would otherwise not be possible. Also the fact that this fixes the visual shaking of show-ship fixes a critical bug with show-ship (and show-ship is needed to show ship animations and texture changes, so cockpit models alone cannot always be used).

The-E (approved):

PR #4707, opened 2022-09-24
fix off-by-one error in Lua line numbering

Goober5000:
This fixes the newline count used to make the parsed Lua code align with the source file.


wookieejedi (approved):

Baezon:
This has always bothered me, thanks.

PR #4708, opened 2022-09-24
Make sure immobile objects always have 0 velocity

Baezon:
Collisions with immobile objects... are fundamentally broken, obviously as a physics concept it is nonsensical, and if the non-immobile object has enough mass it can just burrow straight through immobile objects. I'm not going to tackle that right now, but we can at least make sure things don't blow up too badly when it happens. Physics impacts will currently update the speed of an immobile object, but it still won't move, giving it 'phantom' speed, which causes 'phantom collisions' and screws up a bunch of stuff. Rather than hunt all that down, I think its best to just always re-affirm every frame that immobile objects have 0 velocity, just in case it comes from weird places.

Position modification is much rarer, so I did go in and explicitly add checks for the few times the physics tries to 'nudge' things outside of regular physics movement.

### DIFF
--- a/code/object/object.cpp
+++ b/code/object/object.cpp
@@ -1530,6 +1530,12 @@ void obj_move_all(float frametime)
 				// physics
 				obj_move_call_physics(objp, frametime);
 			}
+		} else if (objp->flags[Object::Object_Flags::Immobile]) {
+			// make sure velocity is always 0 for immobile things!
+			vm_vec_zero(&objp->phys_info.vel);
+			vm_vec_zero(&objp->phys_info.desired_vel);
+			vm_vec_zero(&objp->phys_info.rotvel);
+			vm_vec_zero(&objp->phys_info.desired_rotvel);
 		}
 
 		// Submodel movement now happens here, right after physics movement.  It's not excluded by the "immobile" flag.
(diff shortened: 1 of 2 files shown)


code/object/object.cpp
```diff
@@ -1530,6 +1530,12 @@ void obj_move_all(float frametime)
 				// physics
 				obj_move_call_physics(objp, frametime);
 			}
+		} else if (objp->flags[Object::Object_Flags::Immobile]) {
+			// make sure velocity is always 0 for immobile things!
+			vm_vec_zero(&objp->phys_info.vel);
+			vm_vec_zero(&objp->phys_info.desired_vel);
+			vm_vec_zero(&objp->phys_info.rotvel);
+			vm_vec_zero(&objp->phys_info.desired_rotvel);
```
Goober5000, line 1538: This will zero out the velocity even when hull strength is <= 0.  I suggest rearranging the conditions so that the `Immobile` flag is only checked once.

Baezon: I definitely misread the condition, but this still doesn't apply to dying immobile objects, because they will pass the first condition and go into the first block.

Goober5000 (approved):

Issue #4113, opened 2022-03-28
FR: Discord Integrations should be optional


z64555:
A discussion brought up by freijon on Discord prompted the case of making our Discord Rich Presence support into something optional.

I'm thinking at the least a commandline option should be added to opt-out (or opt-in) to using Discord RP, and a Cmake option should be added to allow the Discord RP libs to be excluded.

freijon:
Maybe a short background explaination:
Most Linux distros are 'binary', meaning the packages they install are pre-compiled and the package maintainer decides which features to include. Same for the AppImage. Then there are 'source' based distros (they are a minority) where the end-user decides which features to include. They are presented with a list of features to include/exclude at build time. If we disable discord-rpc it means don't build the discord library and remove all calls to that library.
For the 'binary' distros or Mac OS / Windows which have the feature compiled in by default, it would certainly be nice to have a cmd line option to opt-out sharing this info online.

Goober5000:
Implemented in #4711

PR #4711, opened 2022-09-24
created new "load_last" universal mod directory

EatThePath:
Implements #4113. In summary:
Conventional wisdom is that anything put in the fs2/data folder will override mod data but this is actually false, a file there will be shadowed or overridden by anything in mod files. It's possible that loose files in fs2/data override VP files, I haven't verified that for sure, but not all mods are distributed as such.
Clearly a community need/desire for something that overrides mod data exist, or there'd be no reason for that misconception to arise. So, this PR creates that capability.
This is implemented as effectively a 'built in' mod directory, named "load_last". Anything placed in fs2/load_last/data will be loaded after/above all roots in fs2/, though the appdata roots will still take priory.

### DIFF
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -592,6 +592,7 @@ void cf_build_root_list(const char *cdrom_dir)
 	if ( !_getcwd(working_directory, CF_MAX_PATHNAME_LENGTH ) ) {
 		Error(LOCATION, "Can't get current working directory -- %d", errno );
 	}
+
 	if (Cmdline_override_data) {
 
 		//To allow user-override of mod files, this new root takes presecdence over all mod roots.
@@ -613,6 +614,7 @@ void cf_build_root_list(const char *cdrom_dir)
 
 		cf_build_pack_list(last);
 	}
+
 	//now that the _override root is finished(or not), handle all mods
 	cf_add_mod_roots(working_directory, CF_LOCATION_ROOT_GAME);
 
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -593,8 +593,29 @@ void cf_build_root_list(const char *cdrom_dir)
 		Error(LOCATION, "Can't get current working directory -- %d", errno );
 	}
 
+	//To allow user-override of mod files, this new root takes presecdence over all mod roots.
+	cf_root	*last = nullptr;
+	last = cf_create_root();
+
+	last->location_flags |= CF_LOCATION_ROOT_GAME | CF_LOCATION_TYPE_ROOT | CF_LOCATION_TYPE_SECONDARY_MODS;
+
+	last->path = working_directory;
+	if (last->path.back() != DIR_SEPARATOR_CHAR) {
+		last->path += DIR_SEPARATOR_CHAR;
+	}
+	last->path += "load_last";
+	last->path += DIR_SEPARATOR_CHAR; 
+
+	last->roottype = CF_ROOTTYPE_PATH;
+
+	cf_init_root_pathtypes(last);
+
+	cf_build_pack_list(last);
+
+	//now that the load_last root is finished, handle all mods
 	cf_add_mod_roots(working_directory, CF_LOCATION_ROOT_GAME);
 
+	//finally, handle the default root.
 	root = cf_create_root();
 
 	root->location_flags |= CF_LOCATION_ROOT_GAME | CF_LOCATION_TYPE_ROOT;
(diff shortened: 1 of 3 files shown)


EatThePath:
Just to add, I did do testing of all my statement about loading. I had a lighting profile with massive overexposure in my fs2/data/tables and one with normal exposure in my mod, with different file names, and the exposure was normal on launch. After adding this new root I put another overexposed profile file in my fs2/load_last/data/tables and got my eyes properly melted.

EatThePath:
> [  FAILED  ] PilotPlayerFileTest.binaryToJSONConversion
I assume that my changes have fouled up some aspect of the directory handling, but haven't been able to trace it all out. I've requested Taylor review since we've talked about this objective and similar matters before, if there's a better way to acomplish these objectives or if further changes are needed I'm all ears.

notimaginative:
Haven't had a chance to really look at the unit test problem but I'll check it before the night is over.

My main issues with this are that 1) it really needs to be behind a cmdline option, and 2) the directory name needs to be changed to something more accurate.

I'm not sure whether a cmdline option to enable it or disable it is more appropriate, but for even basic troubleshooting purposes this feature needs to be optional. I would personally prefer that this be opt-in, but I'm not going to make a fuss over that. Though if someone is knowledgeable enough to be using this feature then presumably they shouldn't have an issue enabling it with a cmdline option.

And the directory name should be more appropriate to its purpose: overriding mod data. "load_last" is really only descriptive for when you're parsing modular tables/scripts. For all other game assets it's loaded first (apart from appdata). But in both cases it is overriding the main mod folders. My suggestion is `_override`, which would hopefully not clash with any mods, and it should be pretty obvious what's going to happen when files are in there.


> It's possible that loose files in fs2/data override VP files

Correct, but that only applies to the VP files in fs2. That works for mods too, so loose files in modA/data will override VP files in modA. Basically it's just loose files, then VPs, for each root. But overall root order still applies, so loose files in fs2/data won't override anything from modA

notimaginative:
Regarding the unit test failure, the tests run in portable mode, so no user directories. Due to this the main directory ends up being "load_last" instead of the test data directory and it can't access the pilot files due to their special handling.

If you put this feature behind an opt-in cmdline option then the issue will fix itself. Alternatives are to disable this feature in portable mode, or put it behind an opt-out cmdline option and modify the tests to use that option.

EatThePath:
I've no problem with any of those suggestions or requests. I think opt in flag makes sense, and makes it marginally less likely for someone to accidentally inject overrides into a mod when they don't intend to. Should also fix the test issue, as you say. Thanks for the input.

EatThePath:
@notimaginative I think I've addressed everything now

code/cfile/cfilesystem.cpp
```diff
@@ -592,9 +592,31 @@ void cf_build_root_list(const char *cdrom_dir)
 	if ( !_getcwd(working_directory, CF_MAX_PATHNAME_LENGTH ) ) {
 		Error(LOCATION, "Can't get current working directory -- %d", errno );
 	}
+	if (Cmdline_override_data) {
```
notimaginative, line 595: Some spacing before and after this block would be preferred. Less likely to be confused as being part of the `if` statement above here.

code/cfile/cfilesystem.cpp
```diff
@@ -592,9 +592,31 @@ void cf_build_root_list(const char *cdrom_dir)
...
+		}
+		last->path += "_override";
+		last->path += DIR_SEPARATOR_CHAR; 
+
+		last->roottype = CF_ROOTTYPE_PATH;
+
+		cf_init_root_pathtypes(last);
+
+		cf_build_pack_list(last);
+	}
```
notimaginative, line 615: Blank line after this for clarity.

code/cmdline/cmdline.cpp
```diff
@@ -516,6 +516,7 @@ cmdline_parm log_to_stdout_arg("-stdout_log", nullptr, AT_NONE); // Cmdline_log_
 cmdline_parm slow_frames_ok_arg("-slow_frames_ok", nullptr, AT_NONE);	// Cmdline_slow_frames_ok
 cmdline_parm fixed_seed_rand("-seed", nullptr, AT_INT);	// Cmdline_rng_seed,Cmdline_reuse_rng_seed;
 cmdline_parm luadev_arg("-luadev", "nullptr", AT_NONE);	// Cmdline_lua_devmode
+cmdline_parm override_arg("-override", nullptr, AT_NONE);	// Cmdline_override_data
```
notimaginative, line 519: This may be better as `-override_data` instead, or anything more descriptive really, just so that it's perfectly clear what is going to happen when this option is used. Otherwise it's a bit vague and could be for anything.

notimaginative (changes requested):
A couple of small tweaks needed but otherwise it looks good.

notimaginative (approved):

PR #4712, opened 2022-09-24
Lua API read-only access to the actual loadout

MjnMixael:
Provides read-only access to the actual loadout used to setup the mission. A later PR will add the ability to write to these values.

### DIFF
--- a/code/scripting/api/objs/shipwepselect.cpp
+++ b/code/scripting/api/objs/shipwepselect.cpp
@@ -166,7 +166,7 @@ ADE_INDEXER(l_Loadout_Amount,
 		LuaError(L, "This property is read only.");
 	}
 
-	if (idx < 1 || idx > 7) {
+	if (idx < 1 || idx > MAX_SHIP_WEAPONS) {
 		return ADE_RETURN_NIL;
 	};
 	idx--; // Convert to Lua's 1 based index system
@@ -176,7 +176,7 @@ ADE_INDEXER(l_Loadout_Amount,
 
 ADE_FUNC(__len, l_Loadout_Amount, nullptr, "The number of weapon banks in the slot", "number", "The number of banks.")
 {
-	return ade_set_args(L, "i", MAX_SHIP_PRIMARY_BANKS + MAX_SHIP_SECONDARY_BANKS);
+	return ade_set_args(L, "i", MAX_SHIP_WEAPONS);
 }
 
 //**********HANDLE: loadout weapon
@@ -200,7 +200,7 @@ ADE_INDEXER(l_Loadout_Weapon,
 		LuaError(L, "This property is read only.");
 	}
 
-	if (idx < 1 || idx > 7) {
+	if (idx < 1 || idx > MAX_SHIP_WEAPONS) {
 		return ADE_RETURN_NIL;
 	};
 	idx--; // Convert to Lua's 1 based index system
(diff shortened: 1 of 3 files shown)


code/scripting/api/objs/shipwepselect.cpp
```diff
@@ -119,6 +143,137 @@ ADE_VIRTVAR(ShipClassIndex,
...
+		return ADE_RETURN_NIL;
+
+	// Will eventually be writable so that Lua can manipulate the player loadout!
+	// Note may need to do some checking that the bank is valid for the current ship
+	// and that the amount-to-write is the correct size for the weapon/ship cargo size
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	if (idx < 1 || idx > 7) {
```
The-E, line 169:
```suggestion
	if (idx < 1 || idx > MAX_SHIP_WEAPONS) {
```
Leaving a magic number like that is just going to lead to issues when someone increases the number of primary or secondary banks

code/scripting/api/objs/shipwepselect.cpp
```diff
@@ -119,6 +143,137 @@ ADE_VIRTVAR(ShipClassIndex,
...
+		return ADE_RETURN_NIL;
+	};
+	idx--; // Convert to Lua's 1 based index system
+
+	return ade_set_args(L, "i", current.getBank()->wep_count[idx]);
+}
+
+ADE_FUNC(__len, l_Loadout_Amount, nullptr, "The number of weapon banks in the slot", "number", "The number of banks.")
+{
+	return ade_set_args(L, "i", MAX_SHIP_PRIMARY_BANKS + MAX_SHIP_SECONDARY_BANKS);
```
The-E, line 179:
```suggestion
	return ade_set_args(L, "i", MAX_SHIP_WEAPONS);
```
MAX_SHIP_WEAPONS is defined as the sum of PRIMARY_BANKS and SECONDARY_BANKS

code/scripting/api/objs/shipwepselect.cpp
```diff
@@ -119,6 +143,137 @@ ADE_VIRTVAR(ShipClassIndex,
...
+	if (!ade_get_args(L, "oi", l_Loadout_Weapon.Get(&current), &idx))
+		return ADE_RETURN_NIL;
+
+	// Will eventually be writable so that Lua can manipulate the player loadout!
+	// Note may need to do some checking that the bank is valid for the current ship
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	if (idx < 1 || idx > 7) {
```
The-E, line 203:
```suggestion
	if (idx < 1 || idx > MAX_SHIP_WEAPONS) {
```

code/scripting/api/objs/shipwepselect.cpp
```diff
@@ -119,6 +143,137 @@ ADE_VIRTVAR(ShipClassIndex,
...
+		return ADE_RETURN_NIL;
+	};
+	idx--; // Convert to Lua's 1 based index system
+
+	return ade_set_args(L, "i", current.getBank()->wep[idx] + 1);
+}
+
+ADE_FUNC(__len, l_Loadout_Weapon, nullptr, "The number of weapon banks in the slot", "number", "The number of banks.")
+{
+	return ade_set_args(L, "i", MAX_SHIP_PRIMARY_BANKS + MAX_SHIP_SECONDARY_BANKS);
```
The-E, line 213:
```suggestion
	return ade_set_args(L, "i", MAX_SHIP_WEAPONS);
```

The-E (changes requested):

code/scripting/api/objs/shipwepselect.cpp
MjnMixael: Ah, I missed that that one existed or I would have used it. Done.

The-E (approved):

Issue #3935, opened 2022-01-17
FR: A way to add glowpoints to ships without altering the POF


MjnMixael:
Glowpoints.tbl gives the impression that it was intended to eventually include a way to add glowpoints via table and not in the POF. I could really use this option for a feature I'm working on in BtA.

Glowpoints require a position, a normal vector, a radius, a parent subobject, and a bitmap. Optional data includes Displacement Time, On Time, Off Time, LOD, and Type.

Glowpoints are also stored as N numbers of individual points in a bank. Ships can have multiple banks of points.

BMagnu:
Implemented by #4713

PR #4713, opened 2022-09-24
Modular POF Features

BMagnu:
The first big set of features for Virtual / Modular POFs (#4590)

Adds:
1. Copying Turrets
2. Copying Glowpoints
3. Copying Engines
4. Copying Special-Point Subsystems
5. Automatic handling of attached Glowpoints and Turrets when a Submodel Tree is copied

While not the proposed solution, this implements #3935

### DIFF
--- a/code/model/modelreplace.cpp
+++ b/code/model/modelreplace.cpp
@@ -641,7 +641,7 @@ VirtualPOFOperationAddGlowpoint::VirtualPOFOperationAddGlowpoint() {
 	stuff_string(renameSubmodel, F_NAME);
 }
 
-void VirtualPOFOperationAddGlowpoint::process(polymodel* pm, model_read_deferred_tasks& deferredTasks, model_parse_depth depth, const VirtualPOFDefinition& virtualPof) const {
+void VirtualPOFOperationAddGlowpoint::process(polymodel* pm, model_read_deferred_tasks& /*deferredTasks*/, model_parse_depth depth, const VirtualPOFDefinition& virtualPof) const {
 	auto appendingPM = virtual_pof_build_cache(appendingPOF, depth);
 
 	int dest_subobj_no = model_find_submodel_index(pm, renameSubmodel.c_str());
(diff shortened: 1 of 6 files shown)


code/model/modelreplace.cpp
```diff
@@ -407,6 +483,229 @@ void VirtualPOFOperationAddSubmodel::process(polymodel* pm, model_read_deferred_
...
+	required_string("+POF to Add:");
+	stuff_string(appendingPOF, F_FILESPEC);
+
+	required_string("+Source Turret:");
+	stuff_string(baseNameSrc, F_NAME);
+
+	required_string("+Destination Turret:");
+	stuff_string(baseNameDest, F_NAME);
+
+	if (optional_string("+Destination Barrel:")) {
```
Goober5000, line 496: If +Destination Barrel: is specified, will only the barrel be copied?

code/model/modelreplace.cpp
```diff
@@ -189,8 +196,37 @@ void virtual_pof_init() {
 
 // Internal helper functions
 
+template<typename T, typename member_t>
+T object_copy_including_array_member(const T& item, member_t T::* ptm, int T::* size) {
+	T result{ item };
+	result.*ptm = (member_t) vm_malloc(sizeof(typename std::remove_pointer<member_t>::type) * (result.*size));
```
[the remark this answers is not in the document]
BMagnu: Not in this case, since it's for cloning some structures within polymodel, and the respective structures in question have their fields allocated with malloc. More importantly, they're later deallocated with free, so if we insert stuff copied here, we must malloc

code/model/modelreplace.cpp
BMagnu: No.
This is less optional then it seems.
Effectively, if you have a single part turret as the source, this will do nothing.
If you have a multipart turret as the source, both base and barrel need a target submodel specified. If you don't, it will warn and abort later during processing of the replacement step.

code/model/modelreplace.cpp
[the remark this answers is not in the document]
Goober5000: Ok

code/model/modelreplace.cpp
Goober5000: But specifying a source gun turret is not required?

BMagnu: Yes.
Since the model will have stored informations about existing barrels in the source base subsys information.
Of if a specified base has a barrel, the modular pof system knows about it, and it does not need to be specified in addition to it

Goober5000: Ok.

Goober5000 (approved):

PR #4714, opened 2022-09-25
Flag to turn off continuous turn when attacking

Baezon:
When fightercraft are ordered to attack big ships (if they pass some very modest conditions) they will enter this submode, which has them do nothing but turn in place for a second or two. I have no idea what the intention of this was, and it frustrates modders attempts to understand wtf the AI is doing, in addition to looking bizarre. This submode is used in other contexts, which this flag does not touch.


JohnAFernandez (approved):
Simple enough

PR #4717, opened 2022-09-26
Make Wss_slots writable

MjnMixael:
The next step for scpui loadout. This makes the loadout itself writable by Lua. It comes with checks in place to prevent setting weapon banks to invalid weapons for the current ship in the slot as well as preventing overloading the weapon amounts.

### DIFF
--- a/code/scripting/api/objs/shipwepselect.cpp
+++ b/code/scripting/api/objs/shipwepselect.cpp
@@ -1,5 +1,7 @@
 #include "shipwepselect.h"
 #include "ship/ship.h"
+#include "weapon/weapon.h"
+#include "missionui/missionweaponchoice.h"
 
 namespace scripting {
 namespace api {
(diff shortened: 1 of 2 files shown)


code/scripting/api/objs/shipwepselect.cpp
```diff
@@ -218,19 +279,37 @@ ADE_OBJ(l_Loadout_Ship, int, "loadout_ship", "Loadout handle");
...
-		LuaError(L, "This property is read only.");
+		if (shipIndex > -1) {
+			// If we're not trying to empty the slot then make sure it's a valid ship index
+			if (shipIndex > ship_info_size()) {
+				LuaError(L, "The provided ship index is invalid!");
+			} else {
+				Wss_slots[current].ship_class = shipIndex;
+				//Reset all currently loaded weapons here
+				int i;
+				for (i = 0; i < MAX_SHIP_WEAPONS; i++) {
```
BMagnu, line 305: Just combine these and add the ``int`` in front of the ``i = 0`` in the for loop instead of using an extra line to allocate the int

BMagnu (changes requested):
Tiny tidyup thing, looks good otherwise

code/scripting/api/objs/shipwepselect.cpp
BMagnu, line 305:
```suggestion
				for (int i = 0; i < MAX_SHIP_WEAPONS; i++) {
```

MjnMixael: Done

BMagnu (approved):

PR #4719, opened 2022-09-27
Reset loadout to default and handle commit mission error returns

MjnMixael:
Changes `commit_pressed()` from a void to an int and introduces return values based on the built-in errors. Also adds an API_Access parameter to suppress popups if true so that scpui can handle them with it's own dialog prompts.

Also adds resetSelect() which resets Wss_slots, Ss_pool, and Wl_pool to the mission's default loadout.

Tests of both are working as intended to both scpui Commit presses and retail Commit presses.

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -685,13 +685,14 @@ ADE_FUNC(skipTraining,
 ADE_FUNC(commitToMission,
 	l_UserInterface_Brief,
 	nullptr,
-	"Commits to the current mission with current loadout data, and starts the mission. WIP, do not use!",
-	nullptr,
-	nullptr)
+	"Commits to the current mission with current loadout data, and starts the mission. Returns an integer to represent "
+	"built-in errors or 0 if successful. 1 = general error, 2 = player ship has no weapons, 3 = the required weapon was not found "
+	"loaded on a ship, 4 = 2 or more required weapons were not found loaded on a ship, 5 = a gap in a ship's weapon banks was discovered "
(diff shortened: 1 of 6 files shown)


code/scripting/api/libs/ui.cpp
```diff
@@ -683,13 +683,14 @@ ADE_FUNC(skipTraining,
...
+	"built-in errors or 0 if successful. 1 = general error, 2 = player ship has no weapons, 3 = the required weapon was not found "
+	"loaded on a ship, 4 = 2 or more required weapons were not found loaded on a ship, 5 = a gap in a ship's weapon banks was discovered "
+	"and all empty banks must be at the bottom of the list.",
+	"number error",
+	"the error value")
 {
-	SCP_UNUSED(L);
-	commit_pressed();
-	return ADE_RETURN_NIL;
+	return ade_set_args(L, "i", commit_pressed(1));
```
BMagnu, line 693: If you replace the return type with enum, you have two options of handling this here.
Either you can create a full switch case statement to replace the enum values with the number, or you can do a static cast:
``static_cast<int>(commit_pressed(1))`` will convert the enum class into an int, with numbers assigned incrementally starting at 0 in the order that you defined the return codes in the enum class. So just make sure that you keep the order correct when you define the enum class, and you can simply cast here.

BMagnu (changes requested):

code/scripting/api/libs/ui.cpp
MjnMixael: Done and Done.

BMagnu (approved):

PR #4720, opened 2022-09-27
Fix Remove-Order not working for Play Dead Orders

wookieejedi:
Play-dead orders set a special case of the ai mode (`AIM_PLAY_DEAD`), which had special handling to return early in the AI frame function. Remove-order does not change the ai mode though, and thus even when a play-dead order was removed the `AIM_PLAY_DEAD` remained, which would continually trigger an early out in AI frame, which in turn would make turrets not fire on ships, even when the play-dead order was removed. No other orders had this early return, which is why remove order works for them.

This PR accounts for that special handling case. There also seemed to be an extra cleanup line for rearm and repair, so this was added right after that step.

Tested and works as expected.

### DIFF
--- a/code/ai/ai_flags.h
+++ b/code/ai/ai_flags.h
@@ -147,6 +147,7 @@ namespace AI {
 		Ignore_aspect_when_leading,
 		Fix_good_rearm_time_bug,
 		No_continuous_turn_on_attack,
+		Fixed_removing_play_dead_order,
 
 		NUM_VALUES
 	};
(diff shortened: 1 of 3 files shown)


BMagnu (commented):
I wonder how many past mods, and maybe even retail, unknowingly make use of this behaviour...
Depending on that, we might need to gate this behind a game_settings or AIP flag.
Thoughts?

wookieejedi:
Yeah, I'm good with that and will add a flag later today if I don't hear otherwise.

wookieejedi:
Now behind a flag but also under the enabled flags for `22.4`

JohnAFernandez (approved):
LGTM now that it's gated. Targeted to 22.4 is a good way to do it, too.

BMagnu (approved):

PR #4726, opened 2022-09-29
flip decals right-side up

Goober5000:
For some reason, decals appear upside-down unless the uvec is flipped.

This screenshot shows a test mission with the fix applied.
screen0322

### DIFF
--- a/code/decals/decals.cpp
+++ b/code/decals/decals.cpp
@@ -413,9 +413,12 @@ matrix4 getDecalTransform(Decal& decal) {
 	// this will lead to a situation where we would look at the decal texture "from behind" causing the texture to
 	// appear flipped. We fix that here for the graphics transform by inverting the Z scaling.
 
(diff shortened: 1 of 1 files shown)


Goober5000:
Here is a test mod.

[decalTest.zip](https://github.com/scp-fs2open/fs2open.github.com/files/9670312/decalTest.zip)

wookieejedi (approved):
Simple enough, and at this point is a welcome fix

wookieejedi:
Reason for this flip seems to be because of line 722 in `render.cpp` within the function `setupTransforms(`
```
	if (gr_screen.rendering_to_texture != -1) {
		// Flip the Y-axis when rendering to texture
		path->translate(0.f, i2fl(gr_screen.max_h));
		path->scale(1.f, -1.f);
	}
```

Goober5000:
After consultation with m!m, this could be the result of different coordinate systems, e.g. origin at the bottom of the texture vs the top. FSO places Y=0 at the top, while OpenGL (we assume) places it at the bottom.

PR #4732, opened 2022-09-29
Add local coordinate relations to allow restoring old absolute behaviour

BMagnu:
When fixing +Absolute in a previous PR, I was unknowingly removing a bug that has found use in some indev mods.
This adds a +Local mode, which works as the old, broken +Absolute.
+Relative is a movement in relative space. (i.e., 0,0,0 will be no change)
+Local is a movement to absolute coordinates local to this single animation. (i.e., 0,0,0 will be the state at the beginning of the animation)
+Absolute is a movement to absolute coordinates within the FoR of the parent, considering all other active animations. (i,e. 0,0,0 will be the actual absolute 0 coordinate of the parent FoR)

### DIFF
--- a/code/model/modelanimation.cpp
+++ b/code/model/modelanimation.cpp
@@ -74,9 +74,11 @@ namespace animation {
 			}
 
 			/* fall-thru */
-		case ModelAnimationState::NEED_RECALC:
+		case ModelAnimationState::NEED_RECALC: {
+			ModelAnimationSubmodelBuffer currentAnimationDelta;
+			m_set->initializeSubmodelBuffer(pmi, currentAnimationDelta);
 			//Store the submodels current data as the base for this animation and calculate this animations parameters
-			m_animation->recalculate(applyBuffer, pmi);
+			m_animation->recalculate(applyBuffer, currentAnimationDelta, pmi);
 
 			instanceData.duration = m_animation->getDuration(pmi->id);
 			instanceData.state = ModelAnimationState::RUNNING_FWD;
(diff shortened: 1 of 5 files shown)


code/model/modelanimation_segments.cpp
```diff
@@ -1114,18 +1106,14 @@ namespace animation {
...
 			stuff_vec3d(&parse);
 			offset = std::move(parse);
-			isAbsolute = optional_string("+Absolute");
-			bool relative = optional_string("+Relative");
-
-			if (isAbsolute && relative) {
-				error_display(1, "Translation cannot both be absolute and relative!");
-			}
+			int result = optional_string_one_of(3, "+Relative", "+Local", "+Absolute");
+			isAbsolute = static_cast<ModelAnimationCoordinateRelation>(result < 0 ? 0 : result);
```
Goober5000, line 1116: Considering how this appears a few times, I'd suggest a utility method to parse the values.

code/model/modelanimation.cpp
```diff
@@ -1463,7 +1466,7 @@ namespace animation {
 			}
 			else {
 				auto subsys = sip->animations.getSubmodel(sp->subobj_name);
-				auto rot = std::shared_ptr<ModelAnimationSegmentSetOrientation>(new ModelAnimationSegmentSetOrientation(subsys, angle, isRelative));
+				auto rot = std::shared_ptr<ModelAnimationSegmentSetOrientation>(new ModelAnimationSegmentSetOrientation(subsys, angle, isRelative ? ModelAnimationCoordinateRelation::RELATIVE_COORDS : ModelAnimationCoordinateRelation::ABSOLUTE_COORDS));
```
Goober5000, line 1469: I would replace all the `isRelative`, `absolute`, etc. variables with something like `relationType`.

Goober5000 (changes requested):
Some recommended polishing.

Goober5000:
I'd have approved-merged this, except IceCreamMan is ok with getting the build by Saturday, so that will leave enough time for some polishing.

Goober5000 (approved):

PR #4733, opened 2022-09-30
Add AI flag to support ai bay arrivals > 10 seconds

wookieejedi:
Currently, if an AI ship is emerging from a bay path, the AI will stop following that path after only 10 seconds (and some distance away from the starting point). This results in AI breaking off a path early if a bay arrival path is long and/or uses a slow bay arrive multipliers. For bays that require the ship to follow the exact path this results in AI that then run into the parent ship and each other.

This PR adds a flag which disables that 10 second timeout. Also, in my tests having a flag to disable works best, as opposed to setting a max time (which is hard to calculate with varying ship speeds and varying bay path lengths).

PR is tested and works as expected.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -606,6 +606,8 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$remove-goal properly removes play-dead order:", AI::Profile_Flags::Fixed_removing_play_dead_order);
 
+				set_flag(profile, "$bay arrivials not time limited:", AI::Profile_Flags::Disable_bay_emerge_timeout);
+
 
 				// if we've been through once already and are at the same place, force a move
 				if (saved_Mp && (saved_Mp == Mp))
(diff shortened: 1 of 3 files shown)


code/ai/ai_profiles.cpp
```diff
@@ -606,6 +606,8 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$remove-goal properly removes play-dead order:", AI::Profile_Flags::Fixed_removing_play_dead_order);
 
+				set_flag(profile, "$bay arrivials not time limited:", AI::Profile_Flags::Disable_bay_emerge_timeout);
```
Goober5000, line 609: "arrivials"

Goober5000 (changes requested):

code/ai/ai_profiles.cpp
wookieejedi: thanks, just fixed!

Goober5000 (approved):

PR #4735, opened 2022-09-30
When rendering stuff outside of ON OBJECT RENDER hooks, use immediate mode

The-E:


EatThePath (approved):
Looks solid, restores old behavior from before #4664 when not in the object render hook, and resolves a regression stemming from that change.

I think it could benefit from a sprinkle of explanatory comments for future tinkerers, but I wouldn't make that a blocking request

EatThePath (approved):
comments look good

Goober5000 (changes requested):
The same fix should be applied to `drawModelOOR`, immediately below it, since these are twin functions.

Also, if these two script functions are really intended to be used outside On Object Render, and we intend to support that functionality, then the function header comments should be updated.

The-E:
> The same fix should be applied to `drawModelOOR`, immediately below it, since these are twin functions.
> 
> Also, if these two script functions are really intended to be used outside On Object Render, and we intend to support that functionality, then the function header comments should be updated.

Done

Goober5000:
Good, but regarding the documentation - if our intent is that scripters can use this anywhere, then the "Use with extreme care" sentence should be removed from both functions' headers.

The-E:
> Good, but regarding the documentation - if our intent is that scripters can use this anywhere, then the "Use with extreme care" sentence should be removed from both functions' headers.

done

Goober5000 (approved):
Looks good now, although the previous multi-string single-arg description was turned into a multi-arg description

code/scripting/api/libs/graphics.cpp
```diff
@@ -856,8 +856,8 @@ ADE_FUNC(draw3dLine, l_Graphics, "vector origin, vector destination, [boolean tr
 
 // Aardwolf's test code to render a model, supposed to emulate WMC's gr.drawModel function
 ADE_FUNC(drawModel, l_Graphics, "model model, vector position, orientation orientation",
-         "Draws the given model with the specified position and orientation - Use with extreme care, may not work "
-         "properly in all scripting hooks.  Note: this method does NOT use CurrentResizeMode.",
+         "Draws the given model with the specified position and orientation",
+	     "Note: this method does NOT use CurrentResizeMode.",
```
Goober5000, line 860:
```suggestion
         "Draws the given model with the specified position and orientation.  "
	     "Note: this method does NOT use CurrentResizeMode.",
```

PR #4737, opened 2022-09-30
Api to render the 3d selection effect

MjnMixael:
Adds an API to render the 3D select for ships and weapons at arbitrary coordinates and at arbitrary size.

### DIFF
--- a/code/scripting/api/objs/shipclass.cpp
+++ b/code/scripting/api/objs/shipclass.cpp
@@ -1145,6 +1145,75 @@ ADE_FUNC(renderTechModel2, l_Shipclass, "number X1, number Y1, number X2, number
 	return ade_set_args(L, "b", true);
 }
 
+ADE_FUNC(renderSelectModel,
+	l_Shipclass,
+	"shipclass, boolean restart, number x, number y, [number width = 629, number height = 355, selection effect = current mod setting]",
+	"Draws the 3D select ship model with the chosen effect at the specified coordinates. Restart should "
+	"be true on the first frame this is called and false on subsequent frames. Valid selection effects are 1 (fs1) or 2 (fs2).",
+	"boolean",
+	"true if rendered, false if error")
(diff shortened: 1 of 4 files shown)


code/scripting/api/objs/shipclass.cpp
```diff
@@ -1145,6 +1145,81 @@ ADE_FUNC(renderTechModel2, l_Shipclass, "number X1, number Y1, number X2, number
 	return ade_set_args(L, "b", true);
 }
 
+ADE_FUNC(renderSelectModel,
+	l_Shipclass,
+	"shipclass, boolean restart, number x, number y, [number width = 629, number height = 355, number = currentEffectSetting]",
```
BMagnu, line 1150: usually, the member that the function is called on (i.e. the shipclass in this case) is not part of the list of explicit arguments (since you write ``shipclass:renderSelectModel(restart ...)`` in lua). So just remove that first shipclass in this documentation string and you're good.

code/scripting/api/objs/weaponclass.cpp
```diff
@@ -765,6 +766,74 @@ ADE_FUNC(renderTechModel2,
 	return ade_set_args(L, "b", true);
 }
 
+ADE_FUNC(renderSelectModel,
+	l_Weaponclass,
+	"weaponclass, number x, number y, [number width = 629, number height = 355, number = currentEffectSetting]",
```
BMagnu, line 771: Same here

BMagnu (changes requested):
Small documentation issue, otherwise LGTM

code/scripting/api/objs/shipclass.cpp
MjnMixael: Right. I think I originally had these under UI but it felt better here. Anyway. Fixed!

BMagnu (approved):

PR #4744, opened 2022-10-02
add WaypointSpeedCap

Goober5000:
As requested by FotG


wookieejedi (approved):
Tested and works as expected, thanks! Only minor aspect would be to change the case to lower-case for the first first to match other ship variables `WaypointSpeedCap->waypointSpeedCap` (just a thought though).

Goober5000:
But all the ship virtvars already capitalize the first letter - see PrimaryBanks.

Issue #4648, opened 2022-09-13
Tips.tbl doesn't do key remapping


Goober5000:
In `player_tips_init()`, tips are parsed without doing any key remapping.  The string should be processed using `message_translate_tokens()`.

maingotj:
Hi, I'm new to this project could I work on this?

Goober5000:
> Hi, I'm new to this project could I work on this?

Certainly!

kkruiliu:
Can you elaborate on what the problem is?  Why Tips.tbl need to do key mapping?

kkruiliu:
Where the Tips.tbl located?

kkruiliu:
> In `player_tips_init()`, tips are parsed without doing any key remapping. The string should be processed using `message_translate_tokens()`.
Hello! Where the Tips.tbl located?

Goober5000:
> > In `player_tips_init()`, tips are parsed without doing any key remapping. The string should be processed using `message_translate_tokens()`.
> > Hello! Where is the Tips.tbl located?

It can be found in `root_fs2.vp`.  You'll need a VP viewer such as VPView32 to extract it.

kkruiliu:
> > > In `player_tips_init()`, tips are parsed without doing any key remapping. The string should be processed using `message_translate_tokens()`.
> > > Hello! Where is the Tips.tbl located?
> 
> It can be found in `root_fs2.vp`. You'll need a VP viewer such as VPView32 to extract it.

Hello! I am a beginner at the project and I am trying to solve this bug. 
I tried to convert "message_translate_tokens(stuff_and_malloc_string(F_NAME, NULL))" into char* type and did "Player_tips[Num_player_tips++] = mychar*"; but it didn't work, some tips are missing and some are only showing key.  Is there any more information I need to know? What "message_translate_tokens" function exactly do?

kkruiliu:
#4746

kkruiliu:
#4746

z64555:
`message_translate_tokens()` may be found in `.\code\mission\missiontraining.cpp`.  Its documentation is a little thin, so you may want to search the codebase for all usages of `message_translate_tokens()` to see how it is used.

PR #4746, opened 2022-10-02
tip.tbl_fix_attempt

kkruiliu:
I am trying to fix the bug that tip.tbl doesn't do any key remapping. The commented part in "void Deal_with_delimiter() " is trying to find remapped keys by finding the difference between the default config and the current config.


code/menuui/playermenu.cpp
```diff
@@ -1391,7 +1392,17 @@ void player_tips_init()
...
 			}
-			Player_tips[Num_player_tips++] = stuff_and_malloc_string(F_NAME, NULL);
+
+			//SCP_string str =  message_translate_tokens(stuff_and_malloc_string(F_NAME, NULL));  
+			//char* tmp = &str[0];
+			//Player_tips[Num_player_tips++] = tmp;
+		    //std:: vector<char> v(str.begin(), str.end());
+			// char* ca = &v[0]; // pointer to start of vector
+		    //char* chr = const_cast<char*>(str.c_str());
+			// Player_tips[Num_player_tips++] = chr;
```
The-E, line 1402: While it is normal for people experimenting with a solution to write code that is later found to be unnecessary, we generally do not expect that code to be committed to the main code repository. Code that is commented out should not be part of your PR, unless there are specific reasons to include those lines, which should be laid out in a comment.

code/menuui/playermenu.cpp
```diff
@@ -1430,7 +1441,7 @@ void player_tips_popup()
 
 	// randomly pick one
 	tip = (int)frand_range(0.0f, (float)Num_player_tips - 1.0f);
-
+	// int tip2= rand()%Num_player_tips;                   //"randomly" pick one tip, used for de-bug;
```
The-E, line 1444: See above comment.

code/parse/parselo.cpp
```diff
@@ -1391,17 +1394,67 @@ void stuff_string_line(SCP_string &outstr)
...
+				SCP_string rep = "you've remaped this key";        //replace the remaped key into tips; 
+			
+			while ((pos2 = str.find(ans, pos2)) != SCP_string::npos) {
+				str.replace(pos2, ans.length(), rep);
+				pos2 += rep.length();
+			}
+			}
+		}
+} 
+
```
The-E, line 1457: Several notes here:
* our code style generally prefers using lowercase with underscores for function names. This function should be renamed to `deal_with_delimiters` to stay within our guidelines.
* this code has very inconsistent indentation.
* this function is declared, but never actually called. Is it necessary? Then it should be called somewhere. If not, why is it part of this PR?
* this method uses `std::vector` and `std::stack`. While we do not have an `SCP_stack` yet, we do generally use `SCP_` variants for STL containers in order to have the opportunity to more easily swap in our own allocators, should that be necessary. So, the recommendation here would be to introduce an SCP_stack based on the precedent set by SCP_vector and use that. Assuming, of course, that this function actually is necessary.

The-E (changes requested):

code/parse/parselo.cpp
```diff
@@ -1391,17 +1394,67 @@ void stuff_string_line(SCP_string &outstr)
 // the default string length if using the F_NAME case.
 char *stuff_and_malloc_string(int type, const char *terminators)
 {
-	SCP_string tmp_result;
-
+	SCP_string tmp_result;  
 	stuff_string(tmp_result, type, terminators);
-	drop_white_space(tmp_result);
-
+	tmp_result=message_translate_tokens(tmp_result.c_str());
```
JohnAFernandez, line 1399: So, the problem with this solution is that adding `message_translate_tokens` to this function will change the behavior everywhere it's called. It is called in at least 7 other places and that could have unintended effects.  

`mission_branch_brief_sound` is set using this function, for instance, and I believe that may cause the sound not to play in edge cases.  There should be a way to ensure that `message_translate_tokens` is only called when intended.

code/menuui/playermenu.cpp
z64555: I'd like to mention here that Github has the option of making PR Drafts, which show that the contribution is still a work in progress.  Standard PR's are assumed to be work that is completed, functioning on the author's machine, and ready for review.

kkruiliu: Thank you guys!

sonar-project.properties
```diff
@@ -0,0 +1,12 @@
...
+
+# This is the name and version displayed in the SonarCloud UI.
+#sonar.projectName=fs2open.github.com
+#sonar.projectVersion=1.0
+
+# Path is relative to the sonar-project.properties file. Replace "\" by "/" on Windows.
+#sonar.sources=.
+
+# Encoding of the source code. Default is default system encoding
+#sonar.sourceEncoding=UTF-8
```
JohnAFernandez, line 12: What am I looking at here?  Are you trying to add this to your pr?

z64555:
Ok. time out.  You've contaminated this PR by starting work on another project before making that branch.  I suggest you make a feature branch for Issue #4648 on what you have currently.  Then, reset your local master branch to be a clean clone of the `fs2open.github.com` repo.

  If you're using a git GUI client such as TortoiseGIT, a rebase operation where you skip all commits you've made may be the fastest option.  At worst case you'll have to re-clone the repo and force-push to your remote fork.

Once you have a feature branch for #4648, please try making a PR draft again.

z64555:
Please be sure to read and follow all of our guidelines on the [wiki.](https://github.com/scp-fs2open/fs2open.github.com/wiki)

PR #4750, opened 2022-10-03
add getMusicScore and setMusicScore

Goober5000:
Two new Lua API methods, plus some minor refactoring.


wookieejedi (approved):
Scripting is simple enough and looks good to me, I'll give others a chance to see that files beyond just those are changing as part of the name update, too.

code/scripting/api/objs/enums.h
```diff
@@ -119,8 +119,13 @@ const int32_t LE_MOVIE_PRE_GAME              = 106;
 const int32_t LE_MOVIE_PRE_DEBRIEF           = 107;
 const int32_t LE_MOVIE_POST_DEBRIEF          = 108;
 const int32_t LE_MOVIE_END_CAMPAIGN          = 109;
+const int32_t LE_SCORE_BRIEFING              = 110;
+const int32_t LE_SCORE_DEBRIEFING_SUCCESS    = 111;
+const int32_t LE_SCORE_DEBRIEFING_AVERAGE    = 112;
+const int32_t LE_SCORE_DEBRIEFING_FAILURE    = 113;
+const int32_t LE_SCORE_FICTION_VIEWER        = 114;
```
BMagnu, line 126: Since you explicitly check for index _ranges_ and not just single indices, I'd like a comment here that warns about potentially inserting other values here, and keeping the order intact

BMagnu (changes requested):
I'd add a small comment, looks good otherwise

BMagnu (approved):

PR #4751, opened 2022-10-03
Ship/Wing arrival script functions

Goober5000:
Some functions to make ships or wings arrive immediately, regardless of their arrival cues.

Part of `mission_eval_arrivals` was split off to create `mission_maybe_make_wing_arrive`.  Hiding whitespace changes in the diff will make this easier to read.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -353,13 +353,7 @@ matrix Parse_viewer_orient;
 
 int Loading_screen_bm_index=-1;
 
-// definitions for timestamps for eval'ing arrival/departure cues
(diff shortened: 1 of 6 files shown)


wookieejedi (commented):
I take it removing the un-needed timestamps did not lead to anything strange in your testing?

Goober5000:
Right, removing those timestamps had no adverse effect.

wookieejedi:
Sounds good that tests confirmed everything is working with no side effects.

wookieejedi (approved):
Approving, given update that tests confirmed everything is working with no side effects and update is logical and provides very useful feature.

PR #4752, opened 2022-10-03
add arrival and departure info for wings, ships, and parse objects

Goober5000:
Virtual variables corresponding to the same fields as in `set-arrival-info` and `set-departure-info`.  Some refactoring as well.

~~In draft status pending tests and rebase.~~


wookieejedi (approved):
Very useful values to expose, works well with tests, and overall looks good.

PR #4756, opened 2022-10-05
Add Lua API for ship registry

Goober5000:


code/scripting/api/libs/mission.cpp
```diff
@@ -337,6 +338,35 @@ ADE_FUNC(__len, l_Mission_SEXPVariables, NULL, "Current number of SEXP variables
 	return ade_set_args(L, "i", sexp_variable_count());
 }
 
+//****SUBLIBRARY: Mission/ShipRegistry
+ADE_LIB_DERIV(l_Mission_ShipRegistry, "ShipRegistry", nullptr, "The mission's ship registry: all ships parsed, created, or exited that the mission knows about", l_Mission);
```
BMagnu, line 342: Should include that certain things, like ship-vanish, will not make ships sppear in the correct part of the mission log (or at all).

code/scripting/api/libs/mission.cpp
```diff
@@ -337,6 +338,35 @@ ADE_FUNC(__len, l_Mission_SEXPVariables, NULL, "Current number of SEXP variables
...
+
+		//Lua-->FS2
+		idx--;
+	}
+
+	return ade_set_args(L, "o", l_ShipRegistryEntry.Set(idx));
+}
+
+ADE_FUNC(__len, l_Mission_ShipRegistry, nullptr,
+		 "Number of ship registry entries in the mission.  The value returned is generally stable but will change if a ship is created using ship-create.",
```
BMagnu, line 363: Possibly also with certain methods of vanishing ships?

code/scripting/api/objs/ship_registry_entry.cpp
```diff
@@ -0,0 +1,85 @@
...
+ADE_FUNC(getShip, l_ShipRegistryEntry, nullptr, "Return the ship associated with this ship registry entry", "ship", "The ship, or nil if handle is invalid")
+{
+	int idx;
+	if (!ade_get_args(L, "o", l_ShipRegistryEntry.Get(&idx)))
+		return ADE_RETURN_NIL;
+
+	if (idx < 0 || (size_t)idx >= Ship_registry.size())
+		return ADE_RETURN_NIL;
+
+	return ade_set_args(L, "o", l_Ship.Set(object_h(Ship_registry[idx].objp)));
```
BMagnu, line 80: If a ship has a status other than present, this won't return a valid ship handle, right? if so, this should be reflected in the documentation.

BMagnu (changes requested):

code/scripting/api/libs/mission.cpp
Goober5000: I don't think that's necessary to say here, since the ship registry is completely independent of the mission log.  While "vanished" ships don't appear in the mission log, they are cleaned up just like any other exited ship.  The "exited" status in the ship registry covers all these cases.

code/scripting/api/libs/mission.cpp
Goober5000: No, the number of ship registry entries covers all ships that are known to the mission at any time.  Vanished ships are still known because they existed at one point.

code/scripting/api/objs/ship_registry_entry.cpp
Goober5000: Good point.

code/scripting/api/libs/mission.cpp
BMagnu: Ah, I see. I mixed that up then

BMagnu (approved):

PR #4759, opened 2022-10-05
add wing flags

Goober5000:
Add setFlag and getFlag for wings, just like for ships and parse objects.  Also some minor refactoring.  And fix `PARSEABLE_WING_FLAGS` which was a bit too low.

### DIFF
--- a/code/scripting/api/objs/wing.cpp
+++ b/code/scripting/api/objs/wing.cpp
@@ -71,6 +71,70 @@ ADE_FUNC(isValid, l_Wing, NULL, "Detects whether handle is valid", "boolean", "t
 	return ADE_RETURN_TRUE;
 }
 
+ADE_FUNC(setFlag, l_Wing, "boolean set_it, string flag_name", "Sets or clears one or more flags - this function can accept an arbitrary number of flag arguments.  The flag names are currently limited to the arrival and departure parseable flags.", nullptr, "Returns nothing")
+{
+	int wingnum;
+	bool set_it;
+	const char *flag_name;
+
+	if (!ade_get_args(L, "obs", l_Wing.Get(&wingnum), &set_it, &flag_name))
+		return ADE_RETURN_NIL;
+	int skip_args = 2;	// not 3 because there will be one more below
+
+	if (wingnum < 0 || wingnum >= Num_wings)
+		return ADE_RETURN_NIL;
+
+	auto wingp = &Wings[wingnum];
+
+	do {
+		for (size_t i = 0; i < Num_wing_flag_names; i++)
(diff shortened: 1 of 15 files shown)


code/scripting/api/objs/wing.cpp
```diff
@@ -71,6 +71,70 @@ ADE_FUNC(isValid, l_Wing, NULL, "Detects whether handle is valid", "boolean", "t
...
+					return ADE_RETURN_FALSE;
+			}
+		}
+
+	// read the next flag
+	internal::Ade_get_args_skip = ++skip_args;
+	} while (ade_get_args(L, "|s", &flag_name) > 0);
+
+	// if we're still here, all the flags we were looking for were present
+	return ADE_RETURN_TRUE;
```
BMagnu, line 135: This will return true if the specified flag(s) didn't exist at all. This case should be checked and at least warned, better errored.

code/scripting/api/objs/wing.cpp
```diff
@@ -71,6 +71,70 @@ ADE_FUNC(isValid, l_Wing, NULL, "Detects whether handle is valid", "boolean", "t
...
+	auto wingp = &Wings[wingnum];
+
+	do {
+		for (size_t i = 0; i < Num_wing_flag_names; i++)
+		{
+			if (!stricmp(Wing_flag_names[i].flag_name, flag_name))
+			{
+				if (!wingp->flags[Wing_flag_names[i].flag])
+					return ADE_RETURN_FALSE;
+			}
```
BMagnu, line 127: Like in the above loop you should break when you find the flag but its set to true.

code/ship/ship_flags.h
```diff
@@ -271,7 +271,7 @@ namespace Ship {
     // ignore_count, reinforcement, no_arrival_music, no_arrival_message, no_arrival_warp, no_departure_warp,
 	// same_arrival_warp_when_docked, same_departure_warp_when_docked, no_dynamic, and nav_carry_status
     // Should that change, bump this variable and make sure to make the necessary changes to parse_wing (in missionparse)
-#define PARSEABLE_WING_FLAGS 9
+#define PARSEABLE_WING_FLAGS 10
```
BMagnu, line 274: Did you add a flag somewhere? Why was this bumped?

BMagnu (changes requested):

code/ship/ship_flags.h
Goober5000: There are currently 10 parseable flags.  The number was wrong.

BMagnu (approved):

PR #4761, opened 2022-10-05
Add maximum self repair percentages for health and subsystems

wookieejedi:
Adds two new options to help with the automatic self repairs for hull and subsystems: `$Hull Self Repair Maximum:` and `$Subsystem Self Repair Maximum:`. These values allow mods to specify how much a ship's hull or subsystem can self repair to, which is very useful to have some auto repair but not enough to make the hull or subsystem 100%.

Also cleans up the `ship_auto_repair_frame` function with removing un-used define values, improving comments, and making formatting consistent. Also updates the parse sections of repair rate for hull and subsystems to print to the log if values are being clamped.

Probably easiest to review with hiding whitespace changes.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -119,8 +119,6 @@ extern bool splodeing;
 extern float splode_level;
 extern int splodeingtexture;
 
-#define SHIP_REPAIR_SUBSYSTEM_RATE	0.01f
-
 // The minimum required fuel to engage afterburners
 static const float DEFAULT_MIN_AFTERBURNER_FUEL_TO_ENGAGE = 10.0f;
 
@@ -1011,6 +1009,9 @@ void ship_info::clone(const ship_info& other)
 	hull_repair_rate = other.hull_repair_rate;
 	subsys_repair_rate = other.subsys_repair_rate;
 
+	hull_repair_max = other.hull_repair_max;
+	subsys_repair_max = other.subsys_repair_max;
+
 	sup_hull_repair_rate = other.sup_hull_repair_rate;
 	sup_shield_repair_rate = other.sup_shield_repair_rate;
 	sup_subsys_repair_rate = other.sup_subsys_repair_rate;
@@ -1330,6 +1331,9 @@ void ship_info::move(ship_info&& other)
 
 	hull_repair_rate = other.hull_repair_rate;
 	subsys_repair_rate = other.subsys_repair_rate;
+
+	hull_repair_max = other.hull_repair_max;
+	subsys_repair_max = other.subsys_repair_max;
 
 	sup_hull_repair_rate = other.sup_hull_repair_rate;
 	sup_shield_repair_rate = other.sup_shield_repair_rate;
@@ -1753,8 +1757,11 @@ ship_info::ship_info()
 	}
 
 	hull_repair_rate = 0.0f;
-	//-2 represents not set, in which case the default is used for the ship (if it is small)
+	// -2 represents not set, in which case the default is used for the ship (if it is small)
 	subsys_repair_rate = -2.0f;
+
+	hull_repair_max = 1.0f;
+	subsys_repair_max = 1.0f;
 
 	sup_hull_repair_rate = 0.15f;
 	sup_shield_repair_rate = 0.20f;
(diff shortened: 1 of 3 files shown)


code/ship/ship.cpp
```diff
@@ -9072,46 +9127,54 @@ static void ship_auto_repair_frame(int shipnum, float frametime)
...
 
-	if (sp->flags[Ship::Ship_Flags::Dying]) // do not repair if already dead 
+	// do not repair if already dead 
+	if (sp->flags[Ship::Ship_Flags::Dying])
 		return;
 
-	//Repair the hull...or maybe unrepair?
-	if(sip->hull_repair_rate != 0.0f)
+	// Repair the hull...or maybe unrepair?
+	if (!fl_equal(sip->hull_repair_rate, 0.0f))
```
Baezon, line 9135: This should remain as a strict equality, we're not looking for the result of numerical operations which were intended to bring it to 0 but may have slightly missed due to float imprecision, we are looking specifically for *assigned to 0* (by the initialization code), in which case it will be *exactly* 0. All other cases should be acted upon.

code/ship/ship.cpp
```diff
@@ -9072,46 +9127,54 @@ static void ship_auto_repair_frame(int shipnum, float frametime)
...
 	{
-		objp->hull_strength += sp->ship_max_hull_strength * sip->hull_repair_rate * frametime;
+		float max_hull_repair_strength = sp->ship_max_hull_strength * sip->hull_repair_max;
+		objp->hull_strength += max_hull_repair_strength * sip->hull_repair_rate * frametime;
 
-		if (objp->hull_strength > sp->ship_max_hull_strength)
-			objp->hull_strength = sp->ship_max_hull_strength;
+		if (objp->hull_strength > max_hull_repair_strength)
+			objp->hull_strength = max_hull_repair_strength;
```
Baezon, line 9141: This is admittedly unlikely, but this will make the ship unable to regain a hull strength above its hull repair max, regardless of the source. For instance, while the hull repair is active, and a SEXP causes it to repair above the threshold, this will bring it back down again. Instead of capping the `objp->hull_strength` the added health due to hull repair should itself be limited to only adding if below the threshold.

code/ship/ship.cpp
```diff
@@ -9072,46 +9127,54 @@ static void ship_auto_repair_frame(int shipnum, float frametime)
...
-	if(sip->subsys_repair_rate == -2.0f)
-		real_repair_rate = SHIP_REPAIR_SUBSYSTEM_RATE;
+	// determine if repair rate is default or specified
+	if (sip->subsys_repair_rate == -2.0f)
+		real_repair_rate = 0.01f;
 	else
 		real_repair_rate = sip->subsys_repair_rate;
 
+	// do not bother repairing if rate equals 0 --wookieejedi
+	if (fl_equal(real_repair_rate, 0.0f))
```
Baezon, line 9157: As above, should be strict equality instead of `fl_equal`

code/ship/ship.cpp
```diff
@@ -9072,46 +9127,54 @@ static void ship_auto_repair_frame(int shipnum, float frametime)
...
 
-	// only allow for the auto-repair of subsystems on small ships
-	//...NOT. Check if var has been changed from def -C
-	if ( !(sip->is_small_ship()) && sip->subsys_repair_rate == -2.0f)
+	// do not repair if it is a large ship with a default repair rate
+	if (!(sip->is_small_ship()) && sip->subsys_repair_rate == -2.0f)
 		return;
 	
-	if(sip->subsys_repair_rate == -2.0f)
-		real_repair_rate = SHIP_REPAIR_SUBSYSTEM_RATE;
```
Baezon, line 9095: You said it was unused but it appears `SHIP_REPAIR_SUBSYSTEM_RATE` was used right here? Even if it's only used once, keeping magic numbers away from the code is still a good idea.

code/ship/ship.cpp
```diff
@@ -9123,9 +9186,9 @@ static void ship_auto_repair_frame(int shipnum, float frametime)
 
 			// do incremental repair on the subsystem
 			// check for overflow of current_hits
-			ssp->current_hits += ssp->max_hits * real_repair_rate * frametime;
-			if ( ssp->current_hits > ssp->max_hits ) {
-				ssp->current_hits = ssp->max_hits;
+			ssp->current_hits += max_subsys_repair_hits * real_repair_rate * frametime;
+			if (ssp->current_hits > max_subsys_repair_hits) {
+				ssp->current_hits = max_subsys_repair_hits;
```
Baezon, line 9191: As above, prevents subsystems from being repaired to above the threshold from any source, etc.

Baezon (changes requested):

code/ship/ship.cpp
wookieejedi: Good to know, thanks!

code/ship/ship.cpp
wookieejedi: Changes committed to fix this following similar method that the guardian code uses (per Discord discussion).

Baezon (approved):

PR #4764, opened 2022-10-06
3d select settings

MjnMixael:
Moves -ship_choice_3d and -weapon_choice_3d to game_settings.tbl and splits the multiple changes they controlled into separate unique choices; 3d ship select & effect, 3d weapon select & effect, 3d ship icons, 3d weapon icons, 3d overhead & style. Adds an API for scpui to get these settings and adjusts renderOverheadModel to be able to choose between styles to match.

This does fully deprecate -ship_choice_3d and -weapon_choice_3d in favor of players utilizing the new load_last directory added in #4711 . If that's not desired, I can revert that change and set those commandlines to work in tandem with the new game_settings.tbl options.


wookieejedi:
> This does fully deprecate -ship_choice_3d and -weapon_choice_3d in favor of players utilizing the new load_last directory added in #4711 . If that's not desired, I can revert that change and set those commandlines to work in tandem with the new game_settings.tbl options.

I'm guessing that there will be folks who will want to keep using the launcher flags (even though I am not one of them). Happy to go with whatever the consensus is though

MjnMixael:
I'm going to make my case, but I will also go with the consensus.

Commandline is an unintuitive and difficult-to-manage place for user options, especially in Knossos. We've used it because we didn't really have a great place for such options until recently. Now we have the load_last directory which is designed exactly for this kind of thing. A player can write a quick text file and the option will be applied to all mods they play forever. In addition we have a working SCPUI options menu which can change these sorts of options in-game for mods that take advantage of it. (My hope is that, in the future most will.) 

Most of our long time players are used to the commandline, but we're finally getting to a place where we can abandon it for user-facing settings. I'm thinking the 3D select settings, which are not game-breaking in any way (the game will default to files and then fallback to 3D generated if the files don't exist automatically), makes a good first-case to start retraining about that load_last directory or the in-game SCPUI options menu.

And finally, this PR splits up the two commandlines into eight distinct options that can be independently mixed and matched to the mod creator's design. 3D select view with 2D icons, pre-rendered anim for weapons with 3D icons, etc. Whatever is desired. If I were to re-implement the commandlines, they would not have that level of fine-tuning and would remain as two global flags controlling all eight.

wookieejedi:
Thanks for that detailed and thorough reasoning. You've convinced me that the current method you are using is likely the best.

code/mod_table/mod_table.cpp
```diff
@@ -31,6 +31,12 @@ bool Damage_impacted_subsystem_first;
 bool Cutscene_camera_displays_hud;
 bool Alternate_chaining_behavior;
 bool Use_host_orientation_for_set_camera_facing;
+bool Use_3d_ship_select;
+bool Use_3d_ship_icons;
+bool Use_3d_weapon_select;
+bool Use_3d_weapon_icons;
+bool Use_3d_overhead_ship;
+int Default_overhead_ship_style;
```
EatThePath, line 39: Style should really be either an enum, or if it's only ever going to be one of two styles possibly a boolean.

EatThePath (changes requested):
Besides one quibble, looks good to me.

code/mod_table/mod_table.cpp
MjnMixael: It only has two styles right now. If someone adds one in the future, I thought leaving as int would make that simpler.

MjnMixael: Alright, I have set it up as an enum and added some defensive error checking.

EatThePath (approved):

PR #4767, opened 2022-10-07
enhance ba.print and add ba.println

Goober5000:
There was code that could interpret any Lua type as an argument for printing, so I adapted it to be used in `ba.print()`, `ba.warning()`, and `ba.error()`.  Also `ba.println()` seemed like a useful thing to add.

Viewing the diff without whitespace changes should cut down on clutter.

With this PR, the following test script...
```
ba.println(5)

local t = { 1, 2 }
ba.println(t)

ba.println("x")
```
...produces the following output...
```
5.000000
Table w/ no metatable
x
```

### DIFF
--- a/code/scripting/ade.cpp
+++ b/code/scripting/ade.cpp
@@ -804,122 +804,144 @@ ade_indexer::ade_indexer(lua_CFunction func,
 	LibIdx = ade_manager::getInstance()->getEntry(parent.GetIdx()).AddSubentry(ate);
 }
 
-//*************************Lua functions*************************
-//WMC - Spits out the current Lua stack to "stackdump"
-//This includes variable values, but not names
-void ade_stackdump(lua_State *L, char *stackdump)
+SCP_string ade_tostring(lua_State *L, int argnum, bool add_typeinfo)
 {
-	char buf[512];
-	int stacksize = lua_gettop(L);
+	SCP_string buf;
 
 	//Lua temps
 	double d;
 	int b;
-	char *s;
-	for(int argnum = 1; argnum <= stacksize; argnum++)
-	{
-		int type = lua_type(L, argnum);
-		sprintf(buf, "\r\n%d: ", argnum);
-		strcat(stackdump, buf);
-		switch(type)
-		{
-			case LUA_TNIL:
-				strcat(stackdump, "NIL");
-				break;
-			case LUA_TNUMBER:
-				d = lua_tonumber(L, argnum);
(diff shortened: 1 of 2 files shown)


EatThePath:
no up to tackling the code changes at the moment, but the one off the cuff thing I can say is that if I'm printing variables, just knowing something is a table is less helpful than I'd like. As a scripter I'd probably keep using my own version of this idea which I implemented in lua, and which will recursively print any table's keys and values. In fact that's becoming my go-to way to print snapshots of state, since fennel makes it a piece of cake to throw together a table of variables keyed by their names and it keeps me from having to check if things are nil to avoid blowing up concatenations.

Lacking that capability isn't a blocker by any means, but it's a potential improvement to consider

EatThePath (approved):

code/scripting/ade.cpp
```diff
@@ -804,122 +804,144 @@ ade_indexer::ade_indexer(lua_CFunction func,
...
-				}
+		case LUA_TFUNCTION:
+		{
+			buf = "Function";
+			auto upname = lua_getupvalue(L, argnum, ADE_FUNCNAME_UPVALUE_INDEX);
+			if (upname != nullptr)
+			{
+				buf += " ";
+				buf += lua_tostring(L, -1);
+				buf += "()";
```
EatThePath, line 891: Personally I'd prefer this chunk and others like it to be using sprintf or similar rather than concatenating three times, but it's really a matter of taste. Looks good time me overall

code/scripting/api/libs/base.cpp
```diff
@@ -26,38 +26,60 @@
...
+
 	if (Cmdline_lua_devmode) {
-		nprintf(("scripting","WARNING: %s\n", lua_tostring(L, -1)));
-	}
-	else {
-		Warning(LOCATION, "%s", lua_tostring(L, -1));
+		nprintf(("scripting", "WARNING: %s\n", str.c_str()));
+	} else {
+		Warning(LOCATION, "%s", str.c_str());
 	}
```
BMagnu, line 71: If you ifndef NDEBUG the print / println calls, why not ifndef the warning as well, since it'll only do stuff in debug mode anyways? Could save some performance, especially with heavy use of warning calls for debug

Goober5000:
Putting this in draft pending a few minor updates following the above comments.

Goober5000:
Addressed the warning comment, but I'm not really familiar with how to navigate a table on the C/C++ side.

PR #4768, opened 2022-10-09
Correctly set credits default even for scpui access

MjnMixael:
Sets the credits music default earlier so that SCPUI doesn't get nil data if no table explicitly sets the music.

### DIFF
--- a/code/menuui/credits.cpp
+++ b/code/menuui/credits.cpp
@@ -198,7 +198,7 @@ static credits_screen_buttons Buttons[NUM_BUTTONS][GR_NUM_RESOLUTIONS] = {
 //XSTR:ON
 };
 
-char Credits_music_name[NAME_LENGTH];
+char Credits_music_name[NAME_LENGTH] = "Cinema";
 static int	Credits_music_handle = -1;
 static UI_TIMESTAMP	Credits_music_begin_timestamp;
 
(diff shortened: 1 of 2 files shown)


code/menuui/credits.cpp
```diff
@@ -198,16 +198,16 @@ static credits_screen_buttons Buttons[NUM_BUTTONS][GR_NUM_RESOLUTIONS] = {
...
 static UI_TIMESTAMP	Credits_music_begin_timestamp;
 
 static int	Credits_frametime;		// frametime of credits_do_frame() loop in ms
 static int	Credits_last_time;		// timestamp used to calc frametime (in ms)
 static float Credits_counter;
 
-int Credits_num_images;
-int Credits_artwork_index;
+int Credits_num_images = DEFAULT_NUM_IMAGES;
+int Credits_artwork_index = 0;
```
BMagnu, line 210: You set this to 0 instead of -1 as the original code did.
What is the reasoning behind this?

MjnMixael: Good question. I was originally thinking FSO defaulted to 0 if it wasn't set in the table. Further investigation shows it actually sets a random value if < 0. I've appended a commit here to match that functionality in the virtvar.

BMagnu (approved):

Issue #3391, opened 2021-04-19
FR: Alternate table file organization option


EatThePath:
This is a bit out there, but I figured I ought to ask anyway. Like many of my FRs, while I would love it if someone else implements it, this is at least half an open chance for people to tell me if they'd have problems with the proposed changes in case I get motivated enough on the matter to implement it myself.

I, for reasons both aesthetic and practical, prefer to develop my mods as a large number of single-entry tbm files, or in some cases 2-5 closely related entries, rather than one monolithic table for everything. the problem is that throwing several dozen tables of mixed types into data/tables makes a brand new unpleasant mess, and the way that table types are distinguished means it's impractical to even sort these files and get them clustered by type.

This can be managed somewhere by making separate packages for each tbm type, but that's also a mess of it's own kind too. What I would like instead is to be able to make subfolders, data/tables/weapons and so forth, and just have the game assume any .tbm in those folders is of the appropriate type. Then I can stop worrying that I'm going to eventually stub my toe on some package load order headache, and I can get back the four characters in the filename that usually have to go to the table type suffix. In a 32 character string that gets a confining surprisingly fast.

MageKing17:
Seems to be along the same lines as #449, except that was never merged (presumably because APNGs solved the same basic problem in a different way); I don't see there being an equivalent of APNGs for tables, though, and also I'm fairly certain this could be implemented in an even simpler fashion (since the subdirectory could be added at each `parse_modular_table()` call rather than needing to happen dynamically for, potentially, every EFF).

notimaginative:
Currently working on something like this, but I don't think it's exactly what you're asking for. What I'm doing is a cosmetic change simply as an aid for data organization. So you could put weapon tables in `data/tables/weapons` but it doesn't just assume that the tables in there are for weapons, you would still need the `-wep.tbm`.

What I'm working on would allow for things like:
`data/models/terran/fighters/fighter01.pof`
`data/maps/terran/fighters/fighter01/fighter01-*.dds`
`data/effects/weapons/primary/shot01/shot01.eff`
`data/effects/weapons/primary/shot01/shot01*.dds`

It's purely organizational at the cfilesystem level, allowing those files to be found and loaded without requiring any additional code or mod changes. Tables, scripts, sounds, etc. would all work the same way too.

wookieejedi:
Can likely close now that #4771 is merged?

PR #449, opened 2015-11-18
EFF subdirectories

ln-zookeeper:
This allows animation frames of eff animations to be located in a subdirectory. This is useful for de-cluttering the effects directory in particular.

Usage and limitations as per http://www.hard-light.net/forums/index.php?topic=88139.0

Before merging, this needs testing in different mod environments: VP's, multiple mods, replacing anims from retail or other mods, and so on. I've pretty much only tested this in a development setting, as in a single mod as a bare directory, no VP's.

Notes:

I'm not sure about the original rationale behind the changes to the bm_unlock()/bm_unload() calls in bmpman.cpp; I'm sure they were intended as bugfixes, but I don't recall the specifics. I included them now because they've been like this in FotG for a long time without problems, but I'll have to split those into a separate commit or remove entirely before merging.

PR #4771, opened 2022-10-10
add support for subfolders to cfile

notimaginative:
Allow indexing of files in subfolders of known cfile path types as if the files in those subfolders were directly part of the known path. Essentially a cosmetic change allowing mod makers to organize their files in a better way without requiring any direct content changes or special code handling to take advantage of the feature.

### DIFF
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -1848,28 +1848,56 @@ int cfile_get_path_type(const SCP_string& dir)
 {
 	SCP_string buf = dir;
 
-	// Remove trailing slashes; avoid buffer overflow on 1-char strings
-	while (buf.size() > 0 && (buf[buf.size() - 1] == '\\' || buf[buf.size() - 1] == '/')) {
-		buf.resize(buf.size() - 1);
+	// remove leading and trailing slashes
+	if ( !buf.empty() ) {
+		auto start = buf.find_first_not_of("\\/");
+		auto end = buf.find_last_not_of("\\/");
+
+		if ( (start > 0) || (end < buf.length()-1) ) {
+			buf = buf.substr(start, end-start+1);
+		}
 	}
 
-	// Remove leading slashes
-	while (buf.size() > 0 && (buf[0] == '\\' || buf[0] == '/')) {
-		buf = buf.substr(1);
+	if (buf.empty()) {
+		return CF_TYPE_ROOT;
 	}
 
 	// Use official DIR_SEPARATOR_CHAR
-	for (char& c : buf) {
-		if (c == '\\' || c == '/') {
-			c = DIR_SEPARATOR_CHAR;
-		}
+	char bad_sep = '/';
+
+	if (bad_sep == DIR_SEPARATOR_CHAR) {
+		bad_sep = '\\';
 	}
+
+	std::replace(buf.begin(), buf.end(), bad_sep, DIR_SEPARATOR_CHAR);
+
+	// identify path type
+	auto best_match = CF_TYPE_INVALID;
 
 	for (auto& Pathtype : Pathtypes) {
-		if (Pathtype.path != nullptr && buf == Pathtype.path) {
-			return Pathtype.index;
+		if ( !Pathtype.path ) {
+			continue;
+		}
+
+		// skip root, it should have been detected before getting here
+		if (Pathtype.index == CF_TYPE_ROOT) {
+			continue;
+		}
+
+		// don't allow unknown subdirs directly under data
+		if (Pathtype.index == CF_TYPE_DATA) {
+			if (buf == Pathtype.path) {
+				best_match = CF_TYPE_DATA;
+			}
+
+			continue;
+		}
+
+		// for everything else just find closest match, allowing for unknown subdirectories
+		if ( !buf.compare(0, strlen(Pathtype.path), Pathtype.path) ) {
+			best_match = Pathtype.index;
 		}
 	}
 
-	return CF_TYPE_INVALID;
+	return best_match;
 }
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -469,7 +469,7 @@ int cf_delete(const char *filename, int path_type, uint32_t location_flags)
 
 	Assert(CF_TYPE_SPECIFIED(path_type));
 
-	cf_create_default_path_string(longname, path_type, filename, false, location_flags);
+	cf_create_default_path_string(longname, path_type, filename, location_flags);
 
 	return (_unlink(longname.c_str()) != -1);
 }
@@ -619,7 +619,7 @@ void cf_create_directory(int dir_type, uint32_t location_flags)
 	int i;
 
 	for (i=num_dirs-1; i>=0; i-- )	{
-		cf_create_default_path_string(longname, dir_tree[i], nullptr, false, location_flags);
+		cf_create_default_path_string(longname, dir_tree[i], nullptr, location_flags);
 		if (stat(longname.c_str(), &statbuf) != 0) {
 			mprintf(( "CFILE: Creating new directory '%s'\n", longname.c_str() ));
 			mkdir_recursive(longname.c_str());
@@ -644,7 +644,7 @@ void cf_create_directory(int dir_type, uint32_t location_flags)
 //
 
 CFILE* _cfopen(const char* source, int line, const char* file_path, const char* mode, int type, int dir_type,
-               bool localize, uint32_t location_flags)
+               bool /* localize */, uint32_t location_flags)
 {
 	/* Bobboau, what is this doing here? 31 is way too short... - Goober5000
 	if( strlen(file_path) > 31 )
@@ -685,7 +685,7 @@ CFILE* _cfopen(const char* source, int line, const char* file_path, const char* 
 			// Create the directory if necessary
 			cf_create_directory(dir_type, location_flags);
 
-			cf_create_default_path_string(longname, dir_type, file_path, false, location_flags);
+			cf_create_default_path_string(longname, dir_type, file_path, location_flags);
 		}
 		Assert( !(type & CFILE_MEMORY_MAPPED) );
 
@@ -735,10 +735,8 @@ CFILE* _cfopen(const char* source, int line, const char* file_path, const char* 
 	//================================================
 	// Search for file on disk, on cdrom, or in a packfile
 
-	char copy_file_path[MAX_PATH_LEN];  // FIX change in memory from cf_find_file_location
-	strcpy_s(copy_file_path, file_path);
+	auto find_res = cf_find_file_location(file_path, dir_type, location_flags);
 
-	auto find_res = cf_find_file_location( copy_file_path, dir_type, localize, location_flags );
 	if ( find_res.found ) {
 
 		// Fount it, now create a cfile out of it
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -1857,13 +1857,8 @@ int cfile_get_path_type(const SCP_string& dir)
 	}
 
 	// Use official DIR_SEPARATOR_CHAR
-	char bad_sep = '/';
-
-	if (bad_sep == DIR_SEPARATOR_CHAR) {
-		bad_sep = '\\';
-	}
-
-	std::replace(buf.begin(), buf.end(), bad_sep, DIR_SEPARATOR_CHAR);
+	extern void normalize_directory_separators(SCP_string &str);
+	normalize_directory_separators(buf);
 
 	// identify path type
 	auto best_match = CF_TYPE_INVALID;
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -68,7 +68,7 @@ cf_pathtype Pathtypes[CF_MAX_PATH_TYPES]  = {
 	{ CF_TYPE_VOICE_TRAINING,		"data" DIR_SEPARATOR_STR "voice" DIR_SEPARATOR_STR "training",				".wav .ogg",						CF_TYPE_VOICE	},
 	{ CF_TYPE_MUSIC,				"data" DIR_SEPARATOR_STR "music",											".wav .ogg",						CF_TYPE_DATA	},
 	{ CF_TYPE_MOVIES,				"data" DIR_SEPARATOR_STR "movies",											".mve .msb .ogg .mp4 .srt .webm .png",CF_TYPE_DATA	},
-	{ CF_TYPE_INTERFACE,			"data" DIR_SEPARATOR_STR "interface",										".pcx .ani .dds .tga .eff .png .jpg",	CF_TYPE_DATA	},
+	{ CF_TYPE_INTERFACE,			"data" DIR_SEPARATOR_STR "interface",										".pcx .ani .dds .tga .eff .png .jpg .rml .rcss",	CF_TYPE_DATA	},
 	{ CF_TYPE_FONT,					"data" DIR_SEPARATOR_STR "fonts",											".vf .ttf .otf",						CF_TYPE_DATA	},
 	{ CF_TYPE_EFFECTS,				"data" DIR_SEPARATOR_STR "effects",											".ani .eff .pcx .neb .tga .jpg .png .dds .sdr",	CF_TYPE_DATA	},
 	{ CF_TYPE_HUD,					"data" DIR_SEPARATOR_STR "hud",												".pcx .ani .eff .tga .jpg .png .dds",	CF_TYPE_DATA	},
@@ -77,6 +77,7 @@ cf_pathtype Pathtypes[CF_MAX_PATH_TYPES]  = {
 	{ CF_TYPE_SQUAD_IMAGES,			"data" DIR_SEPARATOR_STR "players" DIR_SEPARATOR_STR "squads",				".pcx .png .dds",						CF_TYPE_PLAYERS	},
 	{ CF_TYPE_SINGLE_PLAYERS,		"data" DIR_SEPARATOR_STR "players" DIR_SEPARATOR_STR "single",				".pl2 .cs2 .plr .csg .css .json",	CF_TYPE_PLAYERS	},
 	{ CF_TYPE_MULTI_PLAYERS,		"data" DIR_SEPARATOR_STR "players" DIR_SEPARATOR_STR "multi",				".plr .json",						CF_TYPE_PLAYERS	},
+	{ CF_TYPE_PLAYER_BINDS,			"data" DIR_SEPARATOR_STR "players" DIR_SEPARATOR_STR "presets",				".json",							CF_TYPE_PLAYERS	},
 	{ CF_TYPE_CACHE,				"data" DIR_SEPARATOR_STR "cache",											".clr .tmp .bx",					CF_TYPE_DATA	}, 	//clr=cached color
 	{ CF_TYPE_MULTI_CACHE,			"data" DIR_SEPARATOR_STR "multidata",										".pcx .png .jpg .dds .fs2 .txt",		CF_TYPE_DATA	},
 	{ CF_TYPE_MISSIONS,				"data" DIR_SEPARATOR_STR "missions",										".fs2 .fc2 .ntl .ssv",				CF_TYPE_DATA	},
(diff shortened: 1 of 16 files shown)


notimaginative:
**This has yet to be tested on Windows and may have unintended side effects for FRED2 as well! Leaving as draft until sufficient testing and feedback has been done.**

To take advantage of this feature you just need to put files in folders. Any subfolder of an existing cfile path type is treated as if it was directly in that base path.

For instance:

`data/effects/wepanim.eff`
`data/effects/webanim_....dds`

could instead be organized like:

`data/effects/weapons/wepanim/wepanim.eff`
`data/effects/weapons/wepanim/wepanim_....dds`

or like:

`data/effects/weapons/wepanim.eff`
`data/effects/weapons/wepanim/wepanim_....dds`

and everything would just work. Models, maps, sounds, tables, scripts, ... can all be organized the same way: however you like!

Scripts also now support requiring other scripts directly out of subfolders too. This allows you to have scripts with the same name, organized in different subfolders, and load the specific script that you want.

For instance, you have a `starfield.lua` script that is different between missions. You can organize it like so:

`data/scripts/missions/sm01-01/starfield.lua`
`data/scripts/missions/sm01-02/starfield.lua`

and use `require("missions.sm01-01.starfield")` to get your desired script.

notimaginative:
Known quirks/issues:

- **Double indexing**: cfile paths with known subfolders will have the files in those subfolders indexed twice. Currently this is only affecting `data/interface` as workarounds exist for the other instances
- **Lack of list specificity**: For direct file requests (`cfopen()`) it's possible to get a file directly from a subpath as opposed to any file with that name. `cf_get_file_list()` does not currently support this ability
- **Load/sort ordering of scripts and tables**: Files in subfolders are indexed in place, meaning that the files in those subfolders are ordered in a list by the order of the parent folder. This may not behave as expected in some cases

For the sort order issue, an example would be where you have a `zLoadFirst-sct.tbm` and a folder named `zLoadFirst` containing the files `aaa-sct.tbm` and `zzz-sct.tbm`. Although the files are all sorted in alphabetical order, the folder will be sorted *after* the similarly named script tbm. So when processed into the file list for parsing it will actually be `zzz-sct.tbm` which is loaded first.

Essentially, this file structure as seen in a directory listing:

`data/scripts/last-sct.tbm`
`data/scripts/zLoadFirst/aaa-sct.tbm`
`data/scripts/zLoadFirst/zzz-sct.tbm`
`data/scripts/zLoadFirst-sct.tbm`

will actually be indexed in the following order: `last-sct.tbm`, `zLoadFirst-sct.tbm`, `aaa-sct.tbm`, `zzz-sct.tbm`
and when the file list is returned for parsing it looks like: `zzz-sct.tbm`, `aaa-sct.tbm`, `zLoadFirst-sct.tbm`, `last-sct.tbm`

notimaginative:
Testing and feedback would be appreciated from @Goober5000, @EatThePath, and @wookieejedi

notimaginative:
Obsoletes #449. May provide a sufficient alternative to #3391.

EatThePath:
Hopefully will be able to test soon, but some comments from reading the description:
The quirk of subfolder load order seems potentially confusing, but also useful. Probably good on a whole. 

Just to check my understanding...
As I read it, if I have a mod stack of `-mod a,b`, and files
`a:data\tables\blah-shp.tbm`
`b:data\tables\ships\blah-shp.tbm`
then only one of the `blah-shp.tbm` files will be parsed, just as if both were in the tables dir as before?
But if I have
`a:data\scripts\a\blah.lua`
`b:data\scripts\b\blah.lua`
Then both `blah.lua` files will be loaded and accessible with the right paths?

In that last case, would both `blah.lua` files be able to potentially load the other, if the right path is used?

notimaginative:
For your first example, that's correct. Right now the list functions have no concept of subfolders so as far as that code goes those two files are just the same thing in different roots and the second one is filtered out. You would have the same problem if they were in the same mod too, although the indexing order would come in to play then.

The plan is to modify those list functions to allow something like looking for `ships/*-shp.tbm` (or even just `ships/*.tbm`) and just return the files which are in a `ships` subfolder. The code for that is just a bit messier than I originally thought so I haven't managed to get it working yet.

For your second example, that's also correct. You would be able to directly reference `a/blah.lua` in another script, and reference `b/blah.lua` from `a/blah.lua`. But if you just tried to load `blah.lua` then it would grab `a/blah.lua` because it was indexed first.

So you wouldn't want to use the same file names for texture maps or effects just because they are in subfolders, for example. When opening files like that it only searches for the file name and doesn't take folders into account. So having `data/maps/fighter/hull.png` and `data/maps/capship/hull.png` would be a really bad idea since you couldn't choose which to load and it would just grab the first one found.

EatThePath:
sounds fairly good all around. It might be useful for me to have the lua behavior apply to config files too, but not crucial. Being able to list with a subfolder path might be more important, but the lack of it wouldn't be a merge blocker to my mind.

Some of the motivation behind #3391 is me chafing under the 31 character limit, and if I'm making a lot of individual tables I get the desire to not type those four obligatory characters every time. But those are relatively minor things, and the bigger organizational goals of that FR would be served by this arrangement. I wouldn't object to that issue being closed with this.

I'll try to give this some practical testing. Now that I've refreshed my memory I definitely have some use cases to try.

notimaginative:
Both `cfopen()` and `cf_find_file_location()` support subfolders, so it should work with config files too as far as I know. Within scripts you should be able to just write out the full path to the file. For other code you just use a path relative to the cfile pathtype in the filename. So if the file is `data/config/folder/file.cfg` you set CF_TYPE_CONFIG as the dir_type/pathtype and `folder/file.cfg` as the filename. It will parse the filename, extract the path, and go from there. And it will convert directory separators as well so you don't have to worry about which to use.

If it doesn't work let me know and I'll make sure to get it fixed.

Subfolders in file lists is just a convoluted process but I do intend to get it working before this is merged. For other files like textures, effects, and sounds, that code does technically support subfolders but there isn't really a good way to modify the code to use them. I suppose that including a path with the filename in a table entry or model to load textures/effects might work, but I haven't tried it.

I should also note that in all of this new code I'm removing the 31 character limit as I go. VP files will still have the limit, and I can't fix that without a new archive format, but at the cfile level at least the limit is largely gone. We just need to start converting other references in the code to use SCP_string instead of a fixed length.

EatThePath:
> Both `cfopen()` and `cf_find_file_location()` support subfolders, so it should work with config files too as far as I know. Within scripts you should be able to just write out the full path to the file. For other code you just use a path relative to the cfile pathtype in the filename. So if the file is `data/config/folder/file.cfg` you set CF_TYPE_CONFIG as the dir_type/pathtype and `folder/file.cfg` as the filename. It will parse the filename, extract the path, and go from there. And it will convert directory separators as well so you don't have to worry about which to use.

Handy!

To be specific about my goals. I want to be able to get a list of list all of the config files in `data/config/modulename/` in all roots, so I can read them all. This is to facilitate tbm-like features for script configs.

> I should also note that in all of this new code I'm removing the 31 character limit as I go. VP files will still have the limit, and I can't fix that without a new archive format, but at the cfile level at least the limit is largely gone. We just need to start converting other references in the code to use SCP_string instead of a fixed length.

Good to hear!

notimaginative:
> To be specific about my goals. I want to be able to get a list of list all of the config files in `data/config/modulename/` in all roots, so I can read them all. This is to facilitate tbm-like features for script configs.

Gotcha. That should be doable once I get subfolders working in the list functions (probably next weekend).

notimaginative:
Got it partially working with lists now and I'll finish it up over the next week as time allows. It has three different search modes:

`*-shp.tbm`: default behavior, basically just how it's always worked

`folder/*-shp.tbm`: similar to default behavior but only matching files in `folder` subpath

`*/*-shp.tbm`: matches files with filter in all paths, allowing for duplicate filenames if the path is different. So for your previous `blah-shp.tbm` example, this search would include both of the files

EatThePath:
those search methods seem to cover the bases at a glance, if I think of any issues I'll let you know.

I've reorganized dev warmachine to heavily use this and it works best I can tell. I'm around 10 fps on the machine I currently have access to, so I can't stresstest it much, but it launches without evident issue.

One thought: It seems extra easy now to accidentally duplicate a filename in this system. It would probably be useful to have a warning if a file 'shadows' another file in the same root.

notimaginative:
> One thought: It seems extra easy now to accidentally duplicate a filename in this system. It would probably be useful to have a warning if a file 'shadows' another file in the same root.

Yeah, that's an issue I haven't figured out a good way to deal with yet. It's a primary issue with file lists as a list with both `file.cfg` and `folder/file.cfg` can open the one in `folder` when asking for just `file.cfg` if the folder version was indexed first. There is currently no way to work around that and I'm not even sure if it's possible without breaking something.

I'll be thinking on it this week, so maybe I'll have a solution by the weekend.

EatThePath:
I was thinking more like your example of a `hull.dds` texture. If I have `mymod:data/maps/gtcbig/hull.dds` and `mymod:data/maps/gtfsmall/hull.dds`, I'd want to have a warning on startup. But having one of those plus `theirmod:data/maps/hull.dds` would be fine. That seems like something you could check during indexing, but it might be expensive...

notimaginative:
I'm hesitant to do that since the performance hit could vary quite wildly and we'd surely get complaints. But it is possible and we *could* limit it to just debug builds. Maybe I'll try adding something and we can just experiment with it a bit.

As an alternative, what do you think about a comprehensive check via cmdline option? Sort of like `-pofspew` but with a report detailing whatever cfile stuff might be of interest. Not as helpful as an auto check during indexing at startup, but perhaps more useful long term as a diagnostic tool. That would be outside the scope of this PR though.

notimaginative:
I've gotten most of the kinks worked out on lists and so far all of my tests are getting predictable results. I'm going to sit on it another day or two before pushing just in case additional tweaks are required. It's going to take someone else having a go at it to give things a proper test.

The two versions of `cf_get_file_list()` should work the same. However `cf_get_file_list_preallocated()` will **not** support subfolders due to the limited string length. I don't believe that's going to be an issue though. Plus, reworking the code to only use the newer, truly dynamic version of `cf_get_file_list()` is next on my cfile agenda once this PR clears.

The big problem is shadowing, and I'm just not sure how to deal with it. For testing I have a mod with a VP containing `data/tables/test-sct.tbm` and a loose file of `data/tables/test/test-sct.tbm`. With a standard search of `*-sct.tbm` a single `test-sct.tbm` is returned and the loose file is executed. With a glob search of `*/*-sct.tbm` both files are returned correctly, however the loose file is executed twice since it shadows the one in the VP. Obviously this shadowing is across roots. I've thought of a little hack to deal with that (prepending `./` to the filename), though it would break sorting the list by name.

But as that test shows, warning about shadowed files will need to take all roots into account to really be effective.

EatThePath:
I'll try to get more testing of a reorganized mod in this weekend, the situation keeping me from it is now resolved.

When it comes to warnings, I think it's probably something that this feature can be merged without, we can always work more warnings in later and I think the game is better off with subfolders and no new warnings than it is without subfolders.

notimaginative:
I have worked up a simple shadow detection check which might do the trick. Two issues are that it can't work as a `Warning()` due to some weird SDL windowing issue that I'm not keen to track down at the moment, and depending on the file layout it can trigger a lot of false positives.

It's easy enough to just print it to the debug log to deal with issue one. That will make the log a bit messy though. The windowing issue is that, during this stage of the init process, creating a Warning() dialog appears to prevent the main game window from being created. At least on Linux.

I'm a little concerned with the false positives though. It triggers on *possible* shadowing, but it's the actual usage which determines whether it's a problem. So having this check default to enabled may be a huge nuisance if there are more than a couple of shadowed files.

notimaginative:
Test for shadowed files added with a warning to the debug log. Give it a try and let me know if you think that is good enough or not. But again, the false positives once the subfolder setup is fully in use may just annoy people. I made sure that it's easy enough to disable though.

EatThePath:
I pulled that and gave it a spin and my dev copy of warmachine has 270 shadowing warnings. Looks like about half of them are because BP actually has a subfolder in their voice/personas folder with duplicated files, so those are "real" shadowings, but the rest look to be intentional/desirable shadowings.

EatThePath:
Hit this one by going to the tech room mission simulator. 


ASSERTION: "rc == Num_campaigns" at missioncampaign.cpp:312

Assert: "rc == Num_campaigns"
missioncampaign.cpp
Line: 312

ntdll.dll! NtWaitForSingleObject + 20 bytes
KERNELBASE.dll! WaitForSingleObjectEx + 142 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! dump_stacktrace + 283 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! os::dialogs::AssertMessage + 608 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! mission_campaign_build_list + 408 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! sim_room_init + 3923 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! game_enter_state + 1593 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! gameseq_set_state + 299 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! game_process_event + 121 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! game_main + 952 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! SDL_main + 343 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! main_getcmdline + 245 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! invoke_main + 50 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! __scrt_common_main_seh + 302 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! __scrt_common_main + 14 bytes
fs2_open_22_3_0_x64_AVX2-FASTDBG.exe! WinMainCRTStartup + 14 bytes
KERNEL32.DLL! BaseThreadInitThunk + 20 bytes
ntdll.dll! RtlUserThreadStart + 33 bytes

notimaginative:
Nothing to be down about the BP issue from a code perspective. And I think there are some similar cases with other mods as well. It's just going to have to be dealt with on the content side.

Sim room bug should be fixed now. A misguided attempt to optimize things messed it up.

notimaginative:
I haven't found any other issues in my own testing or further areas of improvement. Setting this as ready for review, and hopefully some additional testing will be done by others as well.

code/cfile/cfilesystem.cpp
```diff
@@ -188,48 +189,96 @@ cf_root *cf_create_root()
...
+	if ( !a.sub_path.empty() && b.sub_path.empty() ) {
+		return stricmp(a.sub_path.c_str(), b.name.c_str()) < 0;
+	}
+
+	if ( !a.sub_path.empty() && !b.sub_path.empty() ) {
+		int rc = stricmp(a.sub_path.c_str(), b.sub_path.c_str());
+
+		if (rc) {
+			return rc < 0;
+		}
```
BMagnu, line 216: This is redundant, no?
``if(rc)`` is equivalent to ``if(rc != 0)``, and since ``rc < 0`` implies ``rc != 0``, you can just return ``rc < 0``, no?

code/cfile/cfile.cpp
```diff
@@ -1848,28 +1842,56 @@ int cfile_get_path_type(const SCP_string& dir)
...
-		if (c == '\\' || c == '/') {
-			c = DIR_SEPARATOR_CHAR;
-		}
+	char bad_sep = '/';
+
+	if (bad_sep == DIR_SEPARATOR_CHAR) {
+		bad_sep = '\\';
 	}
 
+	std::replace(buf.begin(), buf.end(), bad_sep, DIR_SEPARATOR_CHAR);
```
BMagnu, line 1866: Should / Could this not be a call to ``normalize_directory_separators``, given that that's already a function?

code/cfile/cfilesystem.cpp
```diff
@@ -1202,11 +1418,31 @@ CFileLocationExt cf_find_file_location_ext( const char *filename, const int ext_
...
+		filespec.erase(0, seperator+1);
+	}
 
 	// strip any existing extension
-	strncpy(filespec, filename, MAX_FILENAME_LEN-1);
+	// (NOTE: to be fully retail compatible, we need to support multiple periods for something like *_1.5.wav,
+	//        which means that we need to strip a length of >2 only, assuming that all valid ext are at least 2 chars)
+	auto dot = filespec.find_last_of(".");
+
+	if ( (dot != SCP_string::npos) && ((filespec.length() - dot) > 2) ) {
```
BMagnu, line 1441: Why do we even need to care about the extension length at all here?
Should it not be safe to assume that the last dot is the extension, but any preceding dots are always just part of the name?

code/cfile/cfilesystem.cpp
```diff
@@ -1202,11 +1418,31 @@ CFileLocationExt cf_find_file_location_ext( const char *filename, const int ext_
...
+		sub_path += DIR_SEPARATOR_CHAR;
+
+		filespec.erase(0, seperator+1);
+	}
 
 	// strip any existing extension
-	strncpy(filespec, filename, MAX_FILENAME_LEN-1);
+	// (NOTE: to be fully retail compatible, we need to support multiple periods for something like *_1.5.wav,
+	//        which means that we need to strip a length of >2 only, assuming that all valid ext are at least 2 chars)
+	auto dot = filespec.find_last_of(".");
```
BMagnu, line 1439: For consistencies sake with all the other "find last char" methods used in this file, this probably should be an rfind too. And search for a char literal, not a string literal

code/cfile/cfilesystem.cpp
```diff
@@ -1753,42 +2021,36 @@ int cf_get_file_list(int max, char** list, int pathtype, const char* filter, int
...
-						l = f->name_ext.length();
-					}
+				auto len = fullname.length();
 
-					list[num_files] = (char *)vm_malloc(l + 1);
-					strncpy(list[num_files], f->name_ext.c_str(), l);
-					list[num_files][l] = 0;
+				list[num_files] = (char *)vm_malloc(len + 1);
+				strncpy(list[num_files], f->name_ext.c_str(), len);
+				list[num_files][len] = 0;
```
BMagnu, line 2053: Why are you allocating a string of the length of fullname and then copying name_ext into it?
Should this whole block not just be ``list[num_files] = vm_strdup(f->name_ext.c_str());``?

BMagnu (changes requested):
Looks mostly good, especially with all the apparent testing that's already happened.
Some questions and comments

code/cfile/cfilesystem.cpp
notimaginative: This is just a comparison of the paths so it needs to fall through to the filename comparison if the paths are equal.

code/cfile/cfile.cpp
notimaginative: I considered that but ultimately decided against it since I didn't want to make `normalize_directory_separators` a global function for a single use case. I'm still on the fence over it though and am open to making the change if anyone else is really for it.

BMagnu: In that case, how about adding the function as an external function declaration to cfile.cpp? That way it won't have to be globally available and you can avoid the code repetition here.

code/cfile/cfilesystem.cpp
BMagnu: Fair then

code/cfile/cfilesystem.cpp
notimaginative: Yep, that's my fault. I did that in a couple of other places too for some reason but went back and fixed them. Obviously missed one.

code/cfile/cfilesystem.cpp
notimaginative: That was how Volition did it and I didn't even think of changing it while working on this. Good time to get that fixed though!

notimaginative: This actually pointed out a bug too since it should have been using the generated `fullname` rather than `f->name_ext` here. That would have been a pretty annoying bug to track down if it had slipped through.

code/cfile/cfile.cpp
BMagnu: Yeah, it'd make it a globally linked function, but as long as its not in a header (and thus polluting the available functions in whereever it or its includees are included), I recon its not too bad.

EatThePath:
I need to roll back to this and retest the newest version, I've been wrapped up in my own branches a lot.

> Nothing to be done about the BP issue from a code perspective. And I think there are some similar cases with other mods as well. It's just going to have to be dealt with on the content side.

I agree with that for sure. So long as it's a log print and not a full Warning, the shadowing messages seem appropriate. I'm tempted to make a new default-on "load order" filter category, but that can be a followup if it feels appropriate once people get some experince with the messages.

BMagnu (approved):
Looks good now.

wookieejedi:
Per Discord Discussion:
I'll merge tomorrow, and any other potential merges tomorrow can be evaluated on a case-by-case basis

PR #4772, opened 2022-10-10
Toggle cockpit

MjnMixael:
Makes cockpits toggle-able in mission and adds a new HUD gauge option to make a gauge render only when cockpits are active, only when cockpits are not active, or render for both (default). Also adds a Lua method to be able to toggle the cockpits.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -136,6 +136,7 @@ ship	*Player_ship;
 int		*Player_cockpit_textures;
 SCP_vector<cockpit_display> Player_displays;
 bool disableCockpits = false;
+bool cockpitActive = false;
 
 wing	Wings[MAX_WINGS];
 bool	Ships_inited = false;
@@ -7457,8 +7458,12 @@ void ship_render_player_ship(object* objp) {
 			|| !(Viewer_mode & VM_EXTERNAL));
 
 	//Nothing to do
-	if (!(renderCockpitModel || renderShipModel))
+	if (!(renderCockpitModel || renderShipModel)) {
+		cockpitActive = false;
 		return;
+	}
+
+	cockpitActive = true;
 
 	//If we aren't sure whether cockpits and external models can share the same worldspace,
 	//we need to pre-render the external ship hull without shadows / deferred and give the cockpit precedence,
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -7451,7 +7451,7 @@ void ship_render_player_ship(object* objp) {
 
 	const bool hasCockpitModel = sip->cockpit_model_num >= 0;
 
-	const bool renderCockpitModel = Viewer_mode != VM_TOPDOWN && hasCockpitModel;
+	const bool renderCockpitModel = (Viewer_mode != VM_TOPDOWN) && hasCockpitModel && !disableCockpits;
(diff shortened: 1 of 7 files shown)


code/ship/ship.cpp
```diff
@@ -7450,19 +7452,24 @@ void ship_render_player_ship(object* objp) {
...
+	}
+
+	cockpitActive = true;
 
 	//If we aren't sure whether cockpits and external models can share the same worldspace,
 	//we need to pre-render the external ship hull without shadows / deferred and give the cockpit precedence,
 	//unless this ship has no cockpit at all
-	const bool prerenderShipModel = renderShipModel && hasCockpitModel && !Cockpit_shares_coordinate_space;
+	const bool prerenderShipModel =
+		renderShipModel && hasCockpitModel && !Cockpit_shares_coordinate_space && !disableCockpits;
```
BMagnu, line 7472: Whether or not we want this line to change is subject to debate:
Effectively, in some cases we need to disable deferred rendering on the show ship model when we might have a cockpit model. Now, the distinction between ``hasCockpit`` and ``renderCockpitModel`` was made to explicitly make changing this line not required.
The original thought was, that if the cockpit _might_ show and we can't guarantee that we can render the show ship model with deferred, always prevent the show ship from being deferred. While this, in some cases, prevents deferred rendering where there could be deferred rendering, it also prevents the user from a noticable change in how the rest of the ship is rendered when the cockpit is en- or disabled.

TL;DR: If there is any chance that the cockpit en- or disable will be possible by hotkey or in a way that does _not_ require a menu access, and thus any chance in render mode is very visible, this line should not be changed, and ship model prerender should _only_ depend on the _possibiliy_ of a cockpit model being there, and not one _actually_ being there in the moment.

code/ship/ship.cpp
```diff
@@ -7450,19 +7452,24 @@ void ship_render_player_ship(object* objp) {
...
-	const bool renderCockpitModel = Viewer_mode != VM_TOPDOWN && hasCockpitModel;
+	const bool renderCockpitModel = (Viewer_mode != VM_TOPDOWN) && hasCockpitModel && !disableCockpits;
 	const bool renderShipModel = (sip->flags[Ship::Info_Flags::Show_ship_model])
 		&& (!Viewer_mode || (Viewer_mode & VM_PADLOCK_ANY) || (Viewer_mode & VM_OTHER_SHIP) || (Viewer_mode & VM_TRACK)
 			|| !(Viewer_mode & VM_EXTERNAL));
 
 	//Nothing to do
-	if (!(renderCockpitModel || renderShipModel))
+	if (!(renderCockpitModel || renderShipModel)) {
+		cockpitActive = false;
```
BMagnu, line 7462: cockpitActive should not be set if we render cockpit or show ship, but likely to the value of ``renderCockpitModel`` directly, which corresponds to whether a cockpit is drawn or not, as opposed to also depending on the show ship flag.

BMagnu (changes requested):

code/ship/ship.cpp
MjnMixael: I'll digest this more later, but without changing that line, the game would ctd upon toggle.

BMagnu: That is interesting. It shouldn't. This needs to be evaluated further, where and why that CTD occurs.

BMagnu (approved):

PR #4774, opened 2022-10-11
fix updating of ui_timestamps

Goober5000:
Since the debug pause check needs to use UI timestamps, and since UI timestamps must keep track of time regardless of time compression, pausing, or mission status, move the updating and checking to `game_do_state_common`.

Also fix a type conversion warning.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -4312,18 +4312,6 @@ void game_set_frametime(int state)
 		Frametime = MAX_FRAMETIME;
 	}
 
-#ifndef NDEBUG
-	// If the frame took more than 5 seconds, assume we're tracing through a debugger.  If timestamps are running, correct the elapsed time.
-	if (!Cmdline_slow_frames_ok && !timestamp_is_paused() && (Last_frame_ui_timestamp.isValid())) {
-		auto delta_timestamp = ui_timestamp_since(Last_frame_ui_timestamp);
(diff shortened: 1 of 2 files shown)


freespace2/freespace.cpp
```diff
@@ -4354,7 +4342,6 @@ void game_set_frametime(int state)
 
 	Last_time = thistime;
 	Last_frame_timestamp = _timestamp();
-	Last_frame_ui_timestamp = ui_timestamp();
```
notimaginative, line 4357: Does `Last_frame_timestamp` not need to be moved as well? Or will the temporary de-sync when the timer offset happens not cause an issue?

Goober5000: I'm assuming `Last_frame_timestamp` doesn't need to be moved, for the following reasons:
1. Frametime-related values have always been updated in this specific function
2. The only time this function is not called during a frame is when the game is paused, which is a state wherein timestamps don't need to be updated anyway (the internal `Timestamp_paused_at_counter` variable only needs to be set when the pause starts)
3. The only place `Last_frame_timestamp` is used is for checking control key or button presses, and in that situation there is no check for duration, only whether the timestamp has changed

notimaginative: Gotcha, makes sense then.

notimaginative (approved):

PR #4775, opened 2022-10-11
Modular species table

MjnMixael:
Makes species.tbl modular, adds +nocreate support, and allows `intel.tbl` as an alias of `species.tbl`


MjnMixael:
I have tested that..

1) The retail table still parses as expected
2) The many many tests still parse. I tested each one manually. Most of the time the result properly gets more data than it used to before it stops parsing and a warning is now provided (not an error). Because of these differences the github tests still fail, though. I have removed them for now.
3) Modular tables parse and add data
4) Modular tables correctly modify previously existing data and warn if the previous entry can't be found (ie: wasn't created earlier in parsing)

I don't really understand why species.tbl is the only tbl with all this rigorous testing in github.. but if you all still want all those tests just let me know where I can learn how to modify the tests for how the tbl is parsed now.

BMagnu (changes requested):
A tiny change.
Otherwise this is good.

BMagnu (approved):

PR #4778, opened 2022-10-13
enable target info in hud_gauges.tbl

MjnMixael:
Adds a way to enable extra target info from hud_gauges.tbl. Does not deprecate the commandline but the table setting, if parsed, will take precedence


wookieejedi (changes requested):
Overall looks good to me, though you might want to update the comment in `code/cmdline/cmdline.cpp` at line 384.

MjnMixael:
Done

wookieejedi (approved):

PR #4785, opened 2022-10-14
Update `$POF file Techroom:` Usage

wookieejedi:
Currently `$POF file Techroom:` is only used in the weapons loadout screen (despite having techroom in the name). This PR expands the use of the `$POF file Techroom:` to shiploadout select and the actual techroom.

If it is desired, I can put it behind a game settings flag, but IMO it would make the most sense for modders to have this just work as the actual name implies.

Tested and works as expected.

### DIFF
--- a/code/menuui/techmenu.cpp
+++ b/code/menuui/techmenu.cpp
@@ -330,13 +330,7 @@ void techroom_select_new_entry()
 			i++;
 		}
 
-		if (sip->pof_file_tech[0] != '\0') {
-			// This cannot load into sip->subsystems, as this will overwrite the subsystems model_num to the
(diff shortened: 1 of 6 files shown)


BMagnu:
I concur on not making that a flag and just enabling it for the techroom and stuff

Goober5000 (changes requested):
While this looks fine, considering the number of times the same code appears, how about putting it in a function with a header like this:
```
int model_load(ship_info *sip, bool prefer_tech_model)
```

wookieejedi:
> , bool prefer_tech_model

Good idea, update pushed.

Goober5000 (approved):

PR #4786, opened 2022-10-14
Additional Target Monitor features

MjnMixael:
This allows setting a wireframe color override globally for the HUD target monitor. Additionally it allows setting the target monitor style and wireframe color per gauge. Priority is in order of gauge > global > default. Still allows changing of the gauge style with the keyboard shortcut but could be paired with $Lock Wireframe Mode: to give HUDs their own unique target monitor style.

### DIFF
--- a/code/hud/hudtargetbox.cpp
+++ b/code/hud/hudtargetbox.cpp
@@ -100,8 +100,11 @@ int Cargo_scan_coords[GR_NUM_RESOLUTIONS][4] = {
 // first element is time flashing expires
 int Targetbox_flash_timers[NUM_TBOX_FLASH_TIMERS];
 
+int CurrentWire = 0;
 int Targetbox_wire = 0;
 int Targetbox_shader_effect = -1;
+color Targetbox_color;
+bool Targetbox_color_override = false;
 bool Lock_targetbox_mode = false;
 
 // Different target states.  This drives the text display right below the hull integrity on the targetbox.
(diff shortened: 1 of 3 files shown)


code/hud/hudtargetbox.cpp
```diff
@@ -100,8 +100,11 @@ int Cargo_scan_coords[GR_NUM_RESOLUTIONS][4] = {
 // first element is time flashing expires
 int Targetbox_flash_timers[NUM_TBOX_FLASH_TIMERS];
 
+int CurrentWire = 0;
 int Targetbox_wire = 0;
```
BMagnu, line 104: Is this value still needed, or entirely superceded by CurrentWire?

MjnMixael: It's still used to track the global override setting from the table (where CurrentWire is considers both global and local). It's not needed in this particular file, but the rest of the variables are still declared here so I left it with them.

BMagnu: You removed a place where Targetbox_wire  was modified in this file. Doesn't this affect the other uses of the Targetbox_wire variable as well then?

MjnMixael: Previously, Targetbox_wire held the table value of $Wireframe Targetbox: and was checked directly in the switch statements on which kind of render method should be used. These were its only uses.

With this PR, Targetbox_wire only holds the top-level table value. That value is used to initialize the per-gauge value of the same option. So each target monitor gauge starts with the default value and can be individually modified if desired.

The second use of Targetbox_wire (in the switch statement) is replaced by using CurrentWire instead.

BMagnu: Okay.

BMagnu (approved):

PR #4787, opened 2022-10-14
create debris using script functions

Goober5000:
This is a new Lua API function that allows creation of arbitrary debris pieces, whether chunks of a ship's hull or shards of shrapnel.  All of the parameters are optional, and a wide variety of situations are possible ranging from vaporized bits to static chunks to chunks propelled by an explosion.

### DIFF
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1108,19 +1108,18 @@ ADE_FUNC(createDebris,
 	{
 		if (lua_isnumber(L, 2))
 		{
-			ade_get_args(L, "|i", &submodel_num);
+			ade_get_args(L, "|*i", &submodel_num);
 			submodel_num--; // Lua --> C/C++
 		}
 		else
 		{
 			const char *name = nullptr;
-			ade_get_args(L, "|s", &name);
+			ade_get_args(L, "|*s", &name);
 			submodel_num = model_find_submodel_index(model_num, name);
 		}
 	}
 
-	internal::Ade_get_args_skip += 2;
-	ade_get_args(L, "|oooffoof",
+	ade_get_args(L, "|**oooffoof",
 		l_Vector.GetPtr(&pos),
 		l_Matrix.GetPtr(&orient_h),
 		l_Enum.GetPtr(&create_flags),
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1069,7 +1069,7 @@ ADE_FUNC(createDebris,
 
 	// validate some arguments
 
-	if (mh != nullptr && !mh->IsValid() || smh != nullptr && !smh->IsValid() || create_flags != nullptr && !create_flags->IsValid())
+	if ((mh != nullptr && !mh->IsValid()) || (smh != nullptr && !smh->IsValid()) || (create_flags != nullptr && !create_flags->IsValid()))
 		return ade_set_args(L, "o", l_Debris.Set(object_h()));
 
 	if (source_ship != nullptr)
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1111,7 +1111,7 @@ ADE_FUNC(createDebris,
 			ade_get_args(L, "|*i", &submodel_num);
 			submodel_num--; // Lua --> C/C++
 		}
-		else
+		else if (lua_isstring(L, 2))
 		{
 			const char *name = nullptr;
 			ade_get_args(L, "|*s", &name);
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -61,6 +61,7 @@
 #include "scripting/api/objs/weapon.h"
 #include "scripting/api/objs/weaponclass.h"
 #include "scripting/api/objs/wing.h"
+#include "scripting/lua/LuaConvert.h"
 #include "scripting/lua/LuaFunction.h"
 #include "scripting/lua/LuaTable.h"
 #include "scripting/scripting.h"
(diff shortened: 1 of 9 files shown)


code/scripting/api/libs/mission.cpp
```diff
@@ -1031,6 +1032,88 @@ ADE_FUNC(createShip,
 		return ade_set_error(L, "o", l_Ship.Set(object_h()));
 }
 
+ADE_FUNC(createDebris,
+	l_Mission,
+	"[model, submodel, vector position, orientation, enumeration create_flags, number hitpoints, number spark_timeout_seconds, team, shipclass source_class, ship source_ship, vector explosion_center, number explosion_force_multiplier]",
+	"Creates a chunk or shard of debris with the specified parameters.  Vectors are in world coordinates.  Any parameter can be nil or negative to specify defaults.  Create flags can be any combination of DC_IS_HULL, DC_VAPORIZE, DC_IS_PERSISTENT, DC_SET_VELOCITY, or DC_FIRE_HOOK.",
```
BMagnu, line 1038: This API is strange.
What is model, if I want to make a debris piece of a shipclass?
What actually happens if I create Debris of an actual ship instance? Will that die?

I'd expect the API to actually be two functions.
For Hull Debris: (position, orientation, ship_class, submodel / submodel_name, create_flags, hitpoints, spark_timeout_seconds, explosion_center, explosion_force_multiplier)
For non-hull debris: (position, orientation, submodel (since that contains a polymodel pointer), create_flags, hitpoints, spark_timeout_seconds, explosion_center, explosion_force_multiplier)

In addition "to specify defaults" is a problem. What are the defaults? For model? No idea. For position? No idea. For orientation? Probably random or identity (I know from reading the code that its random, but a user will not). Same with most of the other arguments

code/scripting/api/objs/enums.h
```diff
@@ -127,6 +127,10 @@ const int32_t LE_SCORE_FICTION_VIEWER        = 114;
 const int32_t LE_NOT_YET_PRESENT             = 115; // the sequence and offsets of these three #defines should correspond to the ShipStatus enums
 const int32_t LE_PRESENT                     = 116;
 const int32_t LE_EXITED                      = 117;
+const int32_t LE_DC_IS_HULL                  = (1<<0); // the following definitions are bitfields independent of the usual indexing
+const int32_t LE_DC_VAPORIZE                 = (1<<1);
+const int32_t LE_DC_SET_VELOCITY             = (1<<2);
+const int32_t LE_DC_FIRE_HOOK                = (1<<3);
```
BMagnu, line 133: This will cause problems.
For one, these Enums will compare equal to other enums with the same index, which is not what we want.
Secondly, the getName function of these enums will return the first enum whose index is equal to their index, which is also not the correct name.
Thirdly, there is currently no API operator that would allow a user to combine two of these creation flags.

Right now, the global enum class is not suitable for these kind of bitfield enums. It would first need a refactor to be capable of holding multiple enums with the same value, but different indices. 

A _very_ ugly quick fix would be to modify the old enum indices in such a way, that neither 1, 2, 4 or 8 exist apart from these new bitfield values, but that jeopardizes anything which might have looked at the actual value of the Enum, as well as pretty much requiring a quick followup with a refactor to enums.

BMagnu (changes requested):

code/scripting/api/libs/mission.cpp
Goober5000: Yeah it's not the best, but I'm not sure how to meaningfully make it better.

Model is there because the code allows you to specify a model but not a submodel if you're creating generic non-hull debris.  Creating debris of a live ship does not cause the live ship to die.

I could explicitly list the defaults of each parameter, but that might get unwieldy.

code/scripting/api/objs/enums.h
Goober5000: There are API operators that allow one to combine enums... they are EnumAND and EnumOR.  They were originally created for the OS_ definitions that are used for object sounds.  Those enums supposedly already have the equality and getName issues you mention, but in practice this hasn't been a problem, at least with the scripts they have been used in.

Yeah, a refactor to enum flags might be desirable.  I'm not sure offhand how to do that though, except maybe by just duplicating the existing enum code and tweaking it for bitfields.

BMagnu: Oh god, it's horrible. I see.
Well...
Keep it for now then, I have an idea how to improve that and will do so after this PR

code/scripting/api/libs/mission.cpp
BMagnu: Okay, I suggest the following API for less confusion:
``createDebris(ship | shipclass | model | submodel source, submodel_name | nil submodel_source (must be nil if source is submodel, can be nil to spawn all debris submodel), vector position, orientation, enumeration create_flags, number hitpoints, number spark_timeout_seconds, team, vector explosion_center, number explosion_force_multiplier)``

This has the advantage, that the ship/model/submodel arguments are all mutually exlusive, as makes sense logically here, and the user won't need a bunch of nils in normal operations for the parameters of possible sources the user doesn't need.

To differentiate between the userdata types, you will need a new ade function. I suggest to use these in LuaConvert.h/.cpp:
```cpp
template<typename T>
bool ade_odata_is_userdata_type(lua_State* L, int stackposition, const ::scripting::ade_obj<T>& obj_type) {
	return ade_odata_is_userdata_type(L, stackposition, obj_type.LibIdx);
}

bool ade_odata_is_userdata_type(lua_State* L, int stackposition, size_t typeIdx) {
	// WMC - Get metatable
	lua_getmetatable(L, stackposition);
	int mtb_ldx = lua_gettop(L);
	Assert(!lua_isnil(L, -1));

	// Get ID
	lua_pushstring(L, "__adeid");
	lua_rawget(L, mtb_ldx);

	if (lua_tonumber(L, -1) != typeIdx) {
		lua_pushstring(L, "__adederivid");
		lua_rawget(L, mtb_ldx);
		if ((size_t)lua_tointeger(L, -1) != typeIdx) {
			return false;
		}
		lua_pop(L, 1);
	}
	lua_pop(L, 2);
	return true;
}
```

The logical consequence is also, that the following function is replaced in the same file, with the following code:
```cpp
bool ade_odata_helper(lua_State* L, int stackposition, size_t idx) {
	if (!ade_odata_is_userdata_type(L, stackposition, idx)) {
		// Issue the LuaError here since this is the only place where we have all relevant information
		LuaError(L, "Argument %d is the wrong type of userdata; '%s' given, but '%s' expected", stackposition,
			::scripting::internal::getTableEntry((size_t)lua_tointeger(L, -2)).Name,
			::scripting::internal::getTableEntry(idx).GetName());
		return false;
	}
	else {
		return true;
	}
}
```

Then you can call ``ade_odata_is_userdata_type`` like so, to test for example if the first argument is of type ship:
```cpp
//Stack position is 1-indexed for arguments of called functions
bool is_ship = ade_odata_is_userdata_type(L, 1, l_Ship);
```

BMagnu (approved):

Goober5000:
Converting to draft for a investigation into a crash in `ade_odata_is_userdata_type`

PR #4788, opened 2022-10-15
fix makeShipArrive and makeWingArrive for reinforcements

Goober5000:
The script functions should make the ships/wings arrive even if they are reinforcement, so this PR fixes that.  Also, this PR adds some additional verification of the `num_uses` of reinforcements, especially when the player calls for a wing that is already present but has more uses remaining.


JohnAFernandez (approved):
Solid

PR #4791, opened 2022-10-16
Enable client turrets to track targets

JohnAFernandez:
This allows clients to user server data to move turrets according to the timing and target data on the server.  Uses a small packet to do so.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -49,6 +49,7 @@
 #include "model/model.h"
 #include "network/multi.h"
 #include "network/multimsgs.h"
+#include "network/multi_turret_manager.h"
 #include "network/multiutil.h"
 #include "object/deadobjectdock.h"
 #include "object/objcollide.h"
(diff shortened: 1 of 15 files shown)


code/ai/aicode.cpp
```diff
@@ -11719,6 +11726,27 @@ void ai_process_subobjects(int objnum)
...
+					pss->turret_enemy_sig = (pss->turret_enemy_objnum > -1) ? Objects[pss->turret_enemy_objnum].signature : 0;
+
+					// if the remote instance had a nullptr, first will be false. second holds the actual angle.
+					if (network_turret_data.angle1.first && pss->submodel_instance_1 != nullptr){
+						pss->submodel_instance_1->cur_angle = network_turret_data.angle1.second;
+					}
+
+					if (network_turret_data.angle2.first && pss->submodel_instance_2 != nullptr){
+						pss->submodel_instance_2->cur_angle = network_turret_data.angle2.second;
+					}
```
BMagnu, line 11745: Will this not set cur_angle each frame?
Intuitively, this should be set once when the packet arrives, and from then the client tracks the transmitted target itself based on the turret_enemy_objnum, no?
Or am I missing something?

BMagnu (commented):
Looks sensible, but I have one question

code/ai/aicode.cpp
JohnAFernandez: Ah, good question! It will only set it once because, `Multi_Turret_Manager.get_latest_packet()` discards all packets except those that are not yet elapsed. Those that haven't elapsed are added to the `retained` vector that is made the value in the map for that net signature/subsystem pair.

code/network/multi_turret_manager.cpp
```diff
@@ -0,0 +1,76 @@
...
+    if (winning_index < 0 && retained.empty()) {
+        mprintf(("There seems to be a non-fatal mistake in multiplayer code.  Please report to Cyborg or check that multiplayer_turret_manager::find() has correct logic.\n"));
+        return false;
+
+    } else if (winning_index > -1) {
+        // we found an appropriate packet.
+        *packet = it->second[winning_index];
+
+        // restore only the elements that we need to keep around.
+        it->second = retained;
```
BMagnu, line 48: this should really be moved, otherwise you'll be making unecessary copies of all elements
```suggestion
        it->second = std::move(retained);
```

code/network/multi_turret_manager.cpp
```diff
@@ -0,0 +1,76 @@
...
+        return true;
+
+    // here none of the received packets are yet available.
+    } else {
+        // restore only the elements that we need to keep around.
+        it->second = retained;
+        return false;
```
BMagnu, line 55: If this is the case, none of the packets was yet available. This can also be formulated as "all packets will be available in the future".
In that case, why do you need the copy at all? should it->second not still be correct as is then?

BMagnu (changes requested):
I see how it works now.
If that is the case, some small performance changes in your turret manager

BMagnu (approved):

code/network/multi_turret_manager.h
```diff
@@ -0,0 +1,96 @@
...
+    int target_objnum;
+    std::pair<bool, float> angle1;
+    std::pair<bool, float> angle2;
+
+    turret_packet() = default;
+    
+    turret_packet(int time_in, ushort target_netsig, std::pair<bool, float> ang1_in, std::pair<bool, float> ang2_in) 
+    {
+        time = _timestamp(time_in + Multi_Timing_Info.get_mission_start_time());
+        target_objnum = OBJ_INDEX(multi_get_network_object(target_netsig));
```
notimaginative, line 22: `multi_get_network_object()` can return `nullptr`, so you should always check its return value before use. Particularly in this case since it would result in an invalid `target_objnum`. You should get the target object, verify it's valid, then set `target_objnum` accordingly (OBJ_INDEX or -1).

code/network/multimsgs.cpp
```diff
@@ -9114,3 +9115,133 @@ void process_sexp_packet(ubyte *data, header *hinfo)
...
+	// send the time that this happened.
+	int time_out = Multi_Timing_Info.get_current_time();
+	ADD_INT(time_out);
+	ADD_USHORT(parent_sig);
+
+	ubyte last_index = (subsys_indexes.size() < MAX_TURRET_TRACK_CHANGES) ? static_cast<ubyte>(subsys_indexes.size()) : MAX_TURRET_TRACK_CHANGES;
+
+	// this is not usually done, but if the number of subsystems actually ends up being different than limit,
+	// this ubyte in the packet will get overwritten at the end of this function.
+	ADD_DATA(last_index);
```
notimaginative, line 9160: The subsys count should be the first thing in the packet to make it easier to update. Move it first, get rid of the magic number for its position, and just use `HEADER_LENGTH` as the offset when updating.

And to avoid confusion it should really be set to zero by default instead of the max possible count;

code/network/multimsgs.cpp
```diff
@@ -9114,3 +9115,133 @@ void process_sexp_packet(ubyte *data, header *hinfo)
...
+
+	// and here's where the correction is done.  This should be rare.
+	if (count != last_index_record) {
+		// this is a little hacky, but best to work through the serialization macros to rewrite the data.
+		int packet_size_record = packet_size;
+		packet_size = SIZE_LOCATION_IN_PACKET;
+		ADD_DATA(count);
+		packet_size = packet_size_record;
+	}
+
```
notimaginative, line 9199: You should change that count > 0 here before sending and avoid wasting bandwidth on useless packets.

code/network/multimsgs.cpp
```diff
@@ -9114,3 +9115,133 @@ void process_sexp_packet(ubyte *data, header *hinfo)
...
+	Assertion(MULTIPLAYER_MASTER, "A non-server is trying to send a turret tracking packet. This is a coder mistake, please report!");
+
+	SCP_vector<int> overflow_list;
+
+	// somehow trying to send for an an invalid or non-network object
+	if (parent_sig == 0) {
+		// returning an empty list will signal the manager to clear its entries
+		return overflow_list;
+	}
+
```
notimaginative, line 9134: It would probably be good to verify that `subsys_indexes` isn't empty before continuing with this. Though really, it shouldn't even get here if that is the case.

code/network/multi_turret_manager.h
```diff
@@ -0,0 +1,96 @@
...
+            return;
+        }
+
+        _packets_out[net_signature].push_back(subsys_index);
+    }
+
+    void send_queued_packets(){
+        for (auto& queued : _packets_out) {
+            extern SCP_vector<int> send_turret_tracking_packet(ushort parent_sig, SCP_vector<int>& subsys_indexes);
+            queued.second = send_turret_tracking_packet(queued.first, queued.second);
```
notimaginative, line 91: You should verify that `queued.second` actually has data before processing the packet. As `_packets_out` still contains all of the packets which have previously been processed, it's loaded with useless data which is just wasting time and bandwidth.

freespace2/freespace.cpp
```diff
@@ -1052,6 +1053,7 @@ void game_level_init()
 	stars_pre_level_init();
 	neb2_level_init();
 	nebl_level_init();
+	Multi_Turret_Manager.reset(); // Cyborg, this can safely be done after everything else.  At some point, I'll probably consolidate these.
```
notimaginative, line 1056: This appears to be a cleanup function so it should really go in `game_level_close()` instead.

code/network/multimsgs.cpp
```diff
@@ -9114,3 +9115,133 @@ void process_sexp_packet(ubyte *data, header *hinfo)
 
 	sexp_packet_received(received_packet, num_ubytes);
 }
+
+// TODO, calculate how many we can send.
+constexpr ubyte MAX_TURRET_TRACK_CHANGES = 32;
```
notimaginative, line 9120: If my math is right then you should be able to add about 170 turrets per packet. Best to just slave this to the max packet size rather than some magic number.

Inside the sending function set a const variable to `MAX_PACKET_SIZE - 20` (technically the 20 is a magic number too, but it's a small safety guard). Then add turrets until `packet_size` hits this variable or you've added all the turrets.

notimaginative (changes requested):

code/network/multimsgs.cpp
JohnAFernandez: Well, that data would be nonsense because it's attached to a net signature of 0 and nothing should be done with it. So this is a way for us to gracefully tell the manager to knock it off.  But it really shouldn't try to send info if the net signature is 0 in the first place. 

How about adding an assertion that `net_signature != 0` and keeping this as a valid way to fail in the quite unlikely event that the manager tries to send this packet in release?

code/network/multi_turret_manager.h
JohnAFernandez: Yes, you're right that I should check if it's empty so that we don't end up sending an empty packet.

But part of the design of `send_turret_tracking_packet` is to return only the subsystems that there wasn't room for in the packet for the manager to reattempt later.  So it won't contain the packets already sent.  It will still have the ship in the unordered map, but the vector in most cases will be empty after the packet is sent.

code/network/multimsgs.cpp
JohnAFernandez: If we had a lot of packets to write, I would suggest making a global `MAX_SAFE_PACKET_SIZE` define any packet could use.  But, this far along I don't see the point.

I think you're right about comparing it to max packet size, because the pack function for the angles is variable as well.

code/network/multimsgs.cpp
JohnAFernandez: Wait, you meant that I should also check if the list is empty?

code/network/multimsgs.cpp
notimaginative: There are one or two other examples in the code where it's done this way (though I don't remember them off-hand). MAX_PACKET_SIZE is already the largest safe packet size though, the extra 20 bytes is so that you don't have to get super tricky with the size check.

So you can safely do `while (packet_size < MY_MAX_SIZE) {...}` and not have to worry about the exact size of the data being added. Each turret looks to add about 7 bytes so it should end the `while` loop with 13 bytes to spare. If that makes sense.

JohnAFernandez: I think I'm just going to keep it at 20 because in case anyone has to change this packet in the future, it would probably be good to have a good amount of padding.  I don't think it's going to be that often that a ship hits the limit.

notimaginative (approved):

wookieejedi:
Just adding here that given this needs a multi bump, we'll probably wait until #4836 is approved to merge both.

PR #4795, opened 2022-10-17
Bump the MAX_INSTANCE_DEPTH constant to 32

### LINKED ISSUE #4794: instance_depth<MAX_INSTANCE_DEPTH assert when asset with large number of submodels is placed/loaded in FRED with dockpoint viewing enabled.
As of recent builds, an asset we use now causes FRED to encounter an error when it is placed in FRED or a mission with it is loaded in FRED. assert is only in FRED with Show Model Dock Points or Show Model Paths selected. This particular model has a subobject depth of up to 12.

Note: If a mission with the asset is loaded in FRED and the camera is not facing the asset, FRED will not crash until the asset enters the field of view.

A FRED-Debug-Log is attached.

[fred2_open.log](https://github.com/scp-fs2open/fs2open.github.com/files/9795820/fred2_open.log)

UPDATE: A zip containing the asset is now attached.

[Zephfuellines.zip](https://github.com/scp-fs2open/fs2open.github.com/files/9802883/Zephfuellines.zip)

The-E:
This bump should be sufficient to cover complex animation states. This fixes #4794

### DIFF
--- a/code/render/3dsetup.cpp
+++ b/code/render/3dsetup.cpp
@@ -44,7 +44,7 @@ vec3d Object_position;
 matrix	Object_matrix;			// Where the opject is pointing in World coordinates
 
 
-#define MAX_INSTANCE_DEPTH	10
+#define MAX_INSTANCE_DEPTH	32
 
 struct instance_context {
 	matrix m;


JohnAFernandez (approved):
*Shrug of approval*

PR #4796, opened 2022-10-17
Add an update interval for client update packets

JohnAFernandez:
Client update packets are sent from the server to client to update the health and threats on the hud.  Currently, there is a one-size fits all interval of 333 ms.  Very noticeable, especially at high pings (missile damage could show up half a second after impact, for instance). This was discovered while doing code analysis for the ETS refactor.

So if we've set a higher update rate in the multi options menu, allow for shorter interval times.


wookieejedi (approved):
Tested and works as expected.

PR #4797, opened 2022-10-18
Add a bunch of movement stats to ship class script api

Baezon:
Mostly copies of the same fields on `physics_info`, but on the ship class, so you know what the base values were if you're messing with the physics values.


BMagnu (approved):
One day I want to get rid of "Setting ... is not supported" and make procedural / dynamic ship classes a thing...

PR #4798, opened 2022-10-18
Damage, swarm, corkscrew info virtvars

MjnMixael:
As requested by ETP. Adds Lua access to these weapon settings and values. I opted to keep the individual data grouped so adding only 3 virtvars instead of 10. Thought it was cleaner that way, but can easily split it up if preferred.

### DIFF
--- a/code/scripting/api/objs/weaponclass.cpp
+++ b/code/scripting/api/objs/weaponclass.cpp
@@ -334,7 +334,7 @@ ADE_VIRTVAR(FreeFlightTime, l_Weaponclass, "number", "The time the weapon will f
 	return ade_set_args(L, "f", Weapon_info[idx].free_flight_time);
 }
 
-ADE_VIRTVAR(SwarmInfo, l_Weaponclass, nullptr, nullptr, "number, number", "The number of swarm missiles, the swarm wait; nil if not defined.")
+ADE_VIRTVAR(SwarmInfo, l_Weaponclass, nullptr, nullptr, "boolean, number, number", "If the weapon has the swarm flag (true/false), the number of swarm missiles (0 or value), the swarm wait (0 or value).")
 {
 	int idx;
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/weaponclass.cpp
```diff
@@ -317,6 +334,58 @@ ADE_VIRTVAR(FreeFlightTime, l_Weaponclass, "number", "The time the weapon will f
...
+
+ADE_VIRTVAR(CorkscrewInfo, l_Weaponclass, nullptr, nullptr, "boolean, number, number, number, boolean, number", 
+	"If the weapon has the corkscrew flag (true/false), the number of corkscrew missiles fired (0 or value), the radius (0 or value), the fire delay (0 or value), counter rotate settings (true/false), the twist value (0 or value).")
+{
+	int idx;
+	if (!ade_get_args(L, "o", l_Weaponclass.Get(&idx)))
+		return ade_set_error(L, "i", 0);
+
+	if (idx < 0 || idx >= weapon_info_size())
+		return ade_set_error(L, "i", 0);
```
Baezon, line 365: I would return false, on `SwarmInfo` too, that way you get false if it was an invalid handle or it didn't have the flag, makes it slightly simpler from the scripter's point of view. And also, once you make it clear what this returns if the handle is invalid, you don't have to keep repeating it imo ("0 or value").

Baezon (changes requested):

code/scripting/api/objs/weaponclass.cpp
MjnMixael: Good idea. Though I'd think ADE_RETURN_NIL would be better because for some of the values false is valid data?

Baezon: I suppose that is true, yes, perhaps best not to conflate validly false with invalid.

Baezon (approved):

PR #4799, opened 2022-10-18
Add self and naomi to SCP credits

MjnMixael:


wookieejedi (changes requested):
Might also suggest @BMagnu and @EatThePath and @Kiloku

Kiloku:
Just Kiloku is fine. Thanks!

wookieejedi (approved):

PR #4801, opened 2022-10-19
Allow writing to player mission specific stats

### LINKED ISSUE #4686: Writable mission stats
Now that individual mission stats are exposed (as opposed to all-time stats), they should be made writable. With that in place, a checkpoint system can save/load the player's stats back to what they were at the time of saving in a mission which fully rounds out what a built-in checkpoint system would do in a modern game.

I'm not going to tackle this anytime soon, so placing it here in case someone feels like getting to it first.

MjnMixael:
Allow scripts to write to the current player's mission specific stats. Resolves #4686

### DIFF
--- a/code/scripting/api/objs/player.cpp
+++ b/code/scripting/api/objs/player.cpp
@@ -541,12 +541,20 @@ ADE_VIRTVAR(MissionPrimaryShotsFired,
 	"The score value")
 {
 	scoring_stats_h* ssh;
-	if (!ade_get_args(L, "o", l_ScoringStats.GetPtr(&ssh))) {
+	int setv;
+	if (!ade_get_args(L, "o|i", l_ScoringStats.GetPtr(&ssh), &setv)) {
 		return ade_set_error(L, "i", -1);
 	}
 
 	if (!ssh->isValid()) {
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/player.cpp
```diff
@@ -700,6 +764,39 @@ ADE_FUNC(getMissionShipclassKills,
...
+	if (!ade_get_args(L, "ooi", l_ScoringStats.GetPtr(&ssh), l_Shipclass.Get(&ship_idx), &setv)) {
+		return ade_set_error(L, "i", -1);
+	}
+
+	if (!ssh->isValid()) {
+		return ade_set_error(L, "i", -1);
+	}
+
+	if (ship_idx < 0 || ship_idx >= (int)Ship_info.size()) {
+		return ade_set_error(L, "i", -1);
```
BMagnu, line 790: These three should probably return false and not -1, based on your documentation

BMagnu (changes requested):

code/scripting/api/objs/player.cpp
MjnMixael: Good call. Done.

BMagnu (approved):

PR #4804, opened 2022-10-20
cockpitinactive gauges should render in chase view

MjnMixael:
Gauges that are set to render if the cockpit is inactive should also render in chase view, even if the current ship has a cockpit and cockpits are enabled.

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -1178,7 +1178,7 @@ bool HudGauge::canRender()
 	}
 	
 	if (render_for_cockpit_toggle > 0) {
-		if (((Viewer_mode & (VM_CHASE)) != 0) && (render_for_cockpit_toggle == 2)) {
+		if ((Viewer_mode & VM_CHASE) && (render_for_cockpit_toggle == 2)) {
 			return true;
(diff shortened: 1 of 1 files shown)


code/hud/hud.cpp
```diff
@@ -1178,7 +1178,9 @@ bool HudGauge::canRender()
 	}
 	
 	if (render_for_cockpit_toggle > 0) {
-		if (cockpitActive && (render_for_cockpit_toggle == 2)) {
+		if (((Viewer_mode & (VM_CHASE)) != 0) && (render_for_cockpit_toggle == 2)) {
```
JohnAFernandez, line 1181: You can simplify by simply putting `(Viewer_mode & VM_Chase)` instead of comparing to zero and putting VM_Chase in parentheses.

wookieejedi: The earlier part of the code a few lines up uses a `== 0` check for a bit, and in theory that could be simplified too, but is possibly beyond the scope of this PR.

MjnMixael: Done

JohnAFernandez (approved):

PR #4806, opened 2022-10-21
UI Popup Scripting Hook

BMagnu:
This is required to allow SCPUI to hook into and replace the system UI hooks.

### DIFF
--- a/code/scripting/global_hooks.cpp
+++ b/code/scripting/global_hooks.cpp
@@ -77,6 +77,28 @@ const std::shared_ptr<Hook> OnDebrisDeath = Hook::Factory(
 		{"Hitpos", "vector", "The world coordinates of the killing blow.  Could be nil."},
 	});
 
+const std::shared_ptr<OverridableHook> OnDialogInit = OverridableHook::Factory("On Dialog Init",
+	"Invoked when a system dialog initalizes. Override to prevent the system dialog to load dialog-related resources (requires retail files)",
+	{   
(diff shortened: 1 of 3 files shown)


code/scripting/global_hooks.cpp
```diff
@@ -77,6 +77,32 @@ const std::shared_ptr<Hook> OnDebrisDeath = Hook::Factory(
 		{"Hitpos", "vector", "The world coordinates of the killing blow.  Could be nil."},
 	});
 
+const std::shared_ptr<OverridableHook> OnDialogInit = OverridableHook::Factory("On Dialog Init",
+	"Invoked when a system dialog initalizes. Override to prevent the system dialog to load dialog-related resources (requires retail files)",
+	{   
+		{"Choices",
+			"table",
+			"A table containing the different choices for this dialog. Contains subtables, each consisting of "
+			"Positivity (an int, 0 if neutral, 1 if positive, and 2 if negative) and "
```
Goober5000, line 86: Why not -1 for negative?

code/scripting/global_hooks.cpp
```diff
@@ -77,6 +77,32 @@ const std::shared_ptr<Hook> OnDebrisDeath = Hook::Factory(
...
+			"Text (s string, the text of the button)."},
```
Goober5000, line 87: "a string" instead of "s string"

Goober5000 (approved):

Issue #4816, opened 2022-10-23
I broke default filters


EatThePath:
#4700 made all debug filters default to on if a blank file is found. Oops. I could have sworn I tested that but obviously I didn't.
I haven't actually checked behavior with no filter file at all come to think of it, but presumably that's working fine or people would have been vocal about it immediately

PR #4817, opened 2022-10-23
Actually dynamically populate the empty debug filter files

EatThePath:
rather than just enableing everything.

### DIFF
--- a/code/osapi/outwnd.cpp
+++ b/code/osapi/outwnd.cpp
@@ -221,9 +221,15 @@ void outwnd_print(const char *id, const char *tmp)
 		Assert( strlen(id)+1 < NAME_LENGTH );
 		outwnd_filter_struct new_filter;
 
+		new_filter.enabled = false;
+
 		strcpy_s(new_filter.name, id);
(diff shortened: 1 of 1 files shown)


EatThePath:
fixes #4816

code/osapi/outwnd.cpp
```diff
@@ -221,9 +221,15 @@ void outwnd_print(const char *id, const char *tmp)
 		Assert( strlen(id)+1 < NAME_LENGTH );
 		outwnd_filter_struct new_filter;
 
+		new_filter.enabled = false;
+
 		strcpy_s(new_filter.name, id);
```
Goober5000, line 226: I would put these on three consecutive lines, with whitespace afterward, because creating the struct is closely related to initializing it.

Goober5000 (approved):
Looks good.  One non-blocking suggestion.

code/osapi/outwnd.cpp
EatThePath: @Goober5000 sounds good, made a change

PR #4818, opened 2022-10-24
Allow adding custom head anis to the list in FRED

### LINKED ISSUE #3875: Feature Request: Table-Defined Head ANI List to Override or Complement Default List
Basically, I'd like the ability to define via table, either in a new table or as an add-on to the messages.tbl file (and tbm) what head anis I'd like to be available in FRED by default for a specific mod or total conversion.  I've been FREDding with TBP and I've found that there is an extensive list of head anis available, but as none of them are default you have to manually copy and paste the filename into FRED for each mission you want to use them in.  Thankfully, once you've manually added them into a mission they will continue to show up, but a pre-definable list would be useful, as well as the ability to force FRED to ignore the default heads that Freespace expects.

MjnMixael:
Implements the ability to define a custom list of Head Anis in game_settings.tbl and uses that to populate the list in the FRED Events Editor. Fixes #3875

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -45,6 +45,7 @@ int FS2NetD_port;
 int Default_multi_object_update_level;
 float Briefing_window_FOV;
 bool Disable_hc_message_ani;
+SCP_vector<SCP_string> Custom_head_anis;
 bool Red_alert_applies_to_delayed_ships;
 bool Beams_use_damage_factors;
 float Generic_pain_flash_factor;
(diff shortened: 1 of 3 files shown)


wookieejedi:
Not totally sure, but is there also code here that should be added to the QTFRED files?

MjnMixael:
> Not totally sure, but is there also code here that should be added to the QTFRED files?

I have no idea. I'm not setup to build qtFred and I don't see any qtFred related files in VS.

BMagnu (approved):
LGTM.
And due to the architecture of the FRED dialog, it shouldn't be an issue if that list contains duplicates or anything.
As for qtFred: Yes, it'd be nice to add support to it, but IMO it's not a blocker to this PR. But I can defer to the qtfred folks' opinion there if need be.

JohnAFernandez:
The only thing that would need to be added to be added to qtFRED is the logic that was added to the Fred 2 files.  But since you're not set up to build it, we should just create an issue that the feature needs to be added to qtFRED.

qtFRED's file structure is a little more complicated than FRED2, and its event editor is crash happy (I don't think anyone has tried to fix it.)

Goober5000 (commented):
I would say this is more properly suited to messages.tbl, rather than game_settings.tbl.

BMagnu:
While an entirely new section in messages.tbl is plausible, I'd concur in the assessment that the other head-ani-appearance flag in the game settings table as precedent will make it more suitable to be in game settings, even if just for consistency's sake

Goober5000:
Ok, fair enough.

PR #4822, opened 2022-10-25
libRocket Texture Passthrough

BMagnu:
This allows to directly use a texture object for image sources in librocket over having to render on screen over librocket or to have to pass it as a blob, which is especially useful for rendering continually updating textures with librocket.
To make sure we don't run into problems with passing ownership and stuff, I had to redo the lock_count system for lua textures. This is a good thing anyways IMO, since we would have leaked textures which are created in lua, used / locked somewhere else, and then went out of scope in lua as it is now, whereas this new way will properly reference count them.

### DIFF
--- a/code/scripting/api/libs/graphics.cpp
+++ b/code/scripting/api/libs/graphics.cpp
@@ -1519,8 +1519,8 @@ ADE_FUNC(createTexture, l_Graphics, "[number Width=512, number Height=512, enume
 	if(idx < 0)
 		return ade_set_error(L, "o", l_Texture.Set(texture_h()));
 
-	//Since creating a render target does not increase load count, we need to still do so even though we just created the texture
-	return ade_set_args(L, "o", l_Texture.Set(texture_h(idx)));
+	//Since creating a render target does not increase load count, when creating the scripting texture object we need to pass true to the constructor so that the constructor will increase the load count for us.
+	return ade_set_args(L, "o", l_Texture.Set(texture_h(idx, true)));
 }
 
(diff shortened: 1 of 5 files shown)


code/scripting/api/libs/graphics.cpp
```diff
@@ -1519,6 +1519,7 @@ ADE_FUNC(createTexture, l_Graphics, "[number Width=512, number Height=512, enume
 	if(idx < 0)
 		return ade_set_error(L, "o", l_Texture.Set(texture_h()));
 
+	//Since creating a render target does not increase load count, we need to still do so even though we just created the texture
```
Goober5000, line 1522: Can you reword this comment?  "We still need to do so" is confusing, as the connection between the comment and the parameter (which is not specified and therefore uses its default value) requires an inference on top of another inference.

code/scripting/api/libs/ui.cpp
```diff
@@ -240,6 +241,19 @@ ADE_FUNC(playCutscene, l_UserInterface, "string Filename, boolean RestartMusic,
...
 }
 
+ADE_FUNC(linkTexture, l_UserInterface, "texture texture", "Links a texture directly to librocket.", "string", "The url string for librocket, or an empty string if invalid.")
+{
+	texture_h* tex;
+
+	if (!ade_get_args(L, "o", l_Texture.GetPtr(&tex)))
+		return ade_set_error(L, "s", "");
+
+	if(!tex->isValid())
```
Goober5000, line 251: Check that `tex` is nullptr here.  I discovered recently that if an argument is present but has the wrong type, `ade_get_args` will succeed despite the pointer not being assigned.

Goober5000 (changes requested):

Goober5000 (approved):

Issue #4821, opened 2022-10-25
Bad Loadout can cause Int3


BMagnu:
In some cases, an insufficient weapon / ship pool can cause an Int3 on briefing init which CTD's debug.
At the same time, it seems to not really be an issue, since it was able to pass it just fine on release, properly loading into the mission with no error or strange behaviour (albeit a no-briefing mission and using SCPUI).
Not sure what the specific conditions are, but it runs into the int3 on [missionshipchoice.cpp#L2784](https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/missionui/missionshipchoice.cpp#L2784).
I am reasonably sure that that really should be just a warning, or at most a failed to load mission.

MjnMixael:
If memory serves from my deep dive into missionshipchoice then this int3() seems to be an error entirely. The function here is resetting (or setting) which ship is the one being viewed; not selected as in chosen for the wing, but simply being looked at in the select UI. First it tries to select the default ship in the loadout. If that doesn't exist then it tries to select the first ship in wss_slots (the actual loadout). If that doesn't exist then it looks like it's trying to select the first ship in the mission's pool. Only if that doesn't exist should it ever really have a problem because the UI won't have anything to display. But to be able to get into a select UI with no ships in wss slots and no ships in the pool seems highly unlikely using retail UI.

This shouldn't trigger at all in SCPUI, but it looks like it's called in common_select_init() which SCPUI uses unaltered.

The quick fix for SCPUI would be to modify common_select_init() and the SCPUI methods that call it so it can skip over calling this function. But I'm also not convinced an int3() should exist on this line at all because it's also there in line 2793 as a final break. Here we still have a possible out; selecting the first ship of the pool.

PR #4824, opened 2022-10-26
Further limit what built-in UI elements can be setup when using SCPUI

MjnMixael:
Partial fix for #4821. Stops SCPUI from being able to even get into functions that only handle retail UI elements.

### DIFF
--- a/code/missionui/missionshipchoice.h
+++ b/code/missionui/missionshipchoice.h
@@ -86,7 +86,7 @@ void draw_wing_block(int wb_num, int hot_slot, int selected_slot, int class_sele
 void ship_select_init();
 void ship_select_do(float frametime);
 void ship_select_close();
-void ship_select_common_init();
+void ship_select_common_init(bool API_Access);
 void ship_select_common_close();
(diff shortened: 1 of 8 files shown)


code/missionui/missionshipchoice.h
```diff
@@ -86,7 +86,7 @@ void draw_wing_block(int wb_num, int hot_slot, int selected_slot, int class_sele
 void ship_select_init();
 void ship_select_do(float frametime);
 void ship_select_close();
-void ship_select_common_init();
+void ship_select_common_init(bool API_Access);
```
BMagnu, line 89: It's dangerous to re-declare functions with default parameters.
I'd just throw this declaration out, since missionshipchoice.cpp includes missionscreencommon.h anyways

BMagnu (changes requested):

code/missionui/missionshipchoice.h
MjnMixael: Done

BMagnu (approved):

PR #4826, opened 2022-10-26
Optional discord presence

MjnMixael:
Sets up discord presence to allow toggling. Adds a game_settings.tbl flag to enable or disable it (Could be command line, but leaning in to the load_last directory instead). Also adds a constructed built-in option to toggle it during a game instance if the active mod is using SCPUI. The built-in option toggle will override the game_settings.tbl flag.

Partially implements #4113


wookieejedi (approved):
Quite a useful feature, and overall LGTM

PR #4827, opened 2022-10-26
asteroid field fix/refactor

Goober5000:
Some refactoring with asteroid fields:
1. Moves asteroid throwing targets into the Asteroid_field struct, in anticipation of potentially multiple fields in the future
2. Moves some initialization code into `asteroid_level_init`
3. Calls `asteroid_level_init` when a mission is cleared
4. Loads asteroid targets as part of the asteroid field update/init logic.  This fixes a FRED bug reported in Discord where asteroid targets stay in the dialog when a new mission is loaded
5. Remove `Default_asteroid_throwing_behavior` flag as it is equivalent to checking that the target list is empty

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -2145,6 +2145,7 @@ void asteroid_frame()
 		return;
 	}
 
+	// If there are no explicit targets, fall back to default retail targeting behavior
 	if (Asteroid_field.target_names.empty()) {
 		int objnum = set_asteroid_throw_objnum();
 		if (Asteroid_targets.empty() || Asteroid_targets[0].objnum != objnum) {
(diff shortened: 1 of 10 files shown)


code/asteroid/asteroid.cpp
```diff
@@ -2133,7 +2145,7 @@ void asteroid_frame()
 		return;
 	}
 
-	if (Default_asteroid_throwing_behavior) {
+	if (Asteroid_field.target_names.empty()) {
```
Baezon, line 2148: Removing `Default_asteroid_throwing_behavior` is fine, but throw in a little comment here clarifying that this is being done because we're falling back to default behavior.

Baezon (changes requested):

Baezon (approved):

PR #4831, opened 2022-10-27
take non-autoscaling AI classes into account when autoscaling

### LINKED ISSUE #4516: Autoscaling AI classes are affected by the presence of non-autoscaling AI classes in ai.tbl
In FSO exists the lovely flag:
$Autoscale by AI Class Index.
BP in particular uses this for the Fury AI

what nobody knew up until very recently thanks to @EatThePath , this flag still affects autoscaling!
Specifically, the FS2 default AI classes will still count the non-autoscaling classes towards their own autoscaling process.

In effect this means that if you have a setup like this:
```
Coward
None
Lieutenant
Captain
Major
Colonel
General
20 non-autoscaling classes
```
The general class will be much weaker then the lieutenant class in the following setup

```
20 non-autoscaling classes
Coward
None
Lieutenant
Captain
Major
Colonel
General
```

Non-autoscaling classes are not affected by this, but autoscaling classes should not be impacted by non-autoscaling ones either. In a mod that uses both (like say, BP), stuff like AI turret fire rate and the like will be affected every time a new AI Class is added. This is all measures of undesirable, and massively affects mission balance.

Goober5000:
This adds an AI profiles flag that filters autoscaling so that only the index and number of autoscaled classes are considered, not the index and number of all AI classes.

Fixes #4516.

### DIFF
--- a/code/ai/ai.h
+++ b/code/ai/ai.h
@@ -463,6 +463,8 @@ typedef struct ai_info {
 	ai_lua_parameters lua_ai_target;
 } ai_info;
 
+int ai_get_autoscale_index(int absolute_index = -1);
+
 // Goober5000
 typedef struct {
 	vec3d docker_point;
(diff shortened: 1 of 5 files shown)


wookieejedi (approved):
LGTM

MoerasGrizzly:
<3 this should make my life a lot easier

PR #4833, opened 2022-10-28
Dont add whitespace for no_fred ships in the main list

### LINKED ISSUE #3633: FRED List cleanup when using ship flag "No_Fred"
If you set a ship flag No_Fred it does remove the ship from the list in FRED, however it leaves behind a blank space. So if you have a lot of these then it creates a massive blank space in the FRED list. Ideally, instead of just blanking out the space, it would just remove the space entirely. See the attached screenshot where I have 31 ships listed as No_Fred and have a long way to scroll to find Waypoint and Jumpnode.

image

MjnMixael:
Turns out this list already looks up objects by name with the exception of waypoints and jump nodes. So this modifies how we create the list slightly by adding a counter and doing checks against the counter total instead of `ship_info.size()` This allows to be able to not list white space in order to keep indices. This also moves where we set the expected index of waypoints and jump nodes to after the counter is complete. It changes them from `ship_info.size() + N` to counter total + N.

I spent a few hours FREDing on this build and didn't run into any bugs, but it is probably worth giving this PR just a little extra scrutiny to be sure. The FRED code isn't the most readable.

Fixes #3633

### DIFF
--- a/fred2/mainfrm.cpp
+++ b/fred2/mainfrm.cpp
@@ -420,7 +420,7 @@ void color_combo_box::DrawItem(LPDRAWITEMSTRUCT lpDrawItemStruct) {
 
 		COLORREF newTextColor = RGB(0x80, 0x80, 0x80);  // light gray
 		if (!fDisabled) {
-			if (z >= ship_type_combo_box_size)
+			if (z >= (int)ship_type_combo_box_size)
 				newTextColor = RGB(0, 0, 0);
 			else {
 				species_info *sinfo = &Species_info[Ship_info[z].species];
@@ -473,7 +473,7 @@ int color_combo_box::GetCurSelNEW() {
 
 	// see if we have a special item (>= Ship_info.size())
 	cur_sel = GetCurSel();
-	if (cur_sel >= ship_type_combo_box_size) {
+	if (cur_sel >= (int)ship_type_combo_box_size) {
 		return cur_sel;
 	}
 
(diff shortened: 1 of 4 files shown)


fred2/mainfrm.cpp
```diff
@@ -118,13 +119,18 @@ void CMainFrame::init_tools() {
     for (auto it = Ship_info.cbegin(); it != Ship_info.cend(); ++it) {
         // don't add the pirate ship
         if (it->flags[Ship::Info_Flags::No_fred]) {
-            m_new_ship_type_combo_box.AddString("");
+            //m_new_ship_type_combo_box.AddString("");
             continue;
-        }
-
-        m_new_ship_type_combo_box.AddString(it->name);
+		} else {
```
JohnAFernandez, line 124: looks like this else block has an extra indentation.

fred2/mainfrm.cpp
```diff
@@ -77,6 +77,7 @@ static UINT indicators[] =
 
 CMainFrame *Fred_main_wnd;
 color_combo_box m_new_ship_type_combo_box;
+size_t num_ships_in_combo_box = 0;
```
JohnAFernandez, line 80: Should probably try to find a more specific name.... ship_type_combo_box_size ?

JohnAFernandez (changes requested):
Two quick things.  Good work overall!  It really does look fine.

fred2/mainfrm.cpp
MjnMixael: Done and done

JohnAFernandez (approved):

PR #4834, opened 2022-10-28
implement modular nebula.tbl

### LINKED ISSUE #3581: FR: Make Nebula table modular
With #3507 merged, it might be appropriate to add modular table options, at least for the new poof styles, both for just extending tables, as well as for +nocreate options modifying the (new) poofs.

MjnMixael:
Makes nebula.tbl extendable with -neb.tbm. Closes #3581

### DIFF
--- a/code/nebula/neb.cpp
+++ b/code/nebula/neb.cpp
@@ -172,7 +172,7 @@ static poof_info* get_nebula_poof_pointer(char* nebula_name)
 	}
 
 	// Didn't find anything.
-	return NULL;
+	return nullptr;
 }
 
 void parse_nebula_table(const char* filename)
@@ -232,7 +232,7 @@ void parse_nebula_table(const char* filename)
 				// Does this poof exist already?
 				// If so, load this new info into it
 				poofp = get_nebula_poof_pointer(pooft.name);
-				if (poofp != NULL) {
+				if (poofp != nullptr) {
 					if (!Parsing_modular_table) {
 						error_display(1,
 							"Error:  Nebula Poof %s already exists.  All nebula poof names must be unique.",
@@ -263,13 +263,15 @@ void parse_nebula_table(const char* filename)
 				if (poof_new) {
 					required_string("$Bitmap:");
 					stuff_string(name, F_NAME, MAX_FILENAME_LEN);
+					strcpy_s(poofp->bitmap_filename, name);
+					generic_anim_init(&poofp->bitmap, name);
 				} else {
-					optional_string("$Bitmap:");
-					stuff_string(name, F_NAME, MAX_FILENAME_LEN);
+					if (optional_string("$Bitmap:")) {
+						stuff_string(name, F_NAME, MAX_FILENAME_LEN);
+						strcpy_s(poofp->bitmap_filename, name);
+						generic_anim_init(&poofp->bitmap, name);
+					}
 				}
-
-				strcpy_s(poofp->bitmap_filename, name);
-				generic_anim_init(&poofp->bitmap, name);
 
 				if (optional_string("$Scale:"))
 					poofp->scale = ::util::parseUniformRange<float>(0.01f, 100000.0f);
--- a/code/nebula/neb.cpp
+++ b/code/nebula/neb.cpp
@@ -217,7 +217,7 @@ void parse_nebula_table(const char* filename)
 				if (Poof_info.size() < MAX_NEB2_POOFS) {
 					Poof_info.push_back(pooft);
 				} else {
-					mprintf(("Warning: more than 32 poofs are defined! Skipping poof %s\n", pooft.name));
+					Warning(LOCATION, "More than 32 poofs are defined! Skipping poof %s\n", pooft.name);
 				}
 			} else if (optional_string("$Name:")) { // new style
 				stuff_string(pooft.name, F_NAME, NAME_LENGTH);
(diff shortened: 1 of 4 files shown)


code/nebula/neb.cpp
```diff
@@ -191,62 +201,98 @@ void neb2_init()
...
 
-					strcpy_s(new_poof.bitmap_filename, name);
-					generic_anim_init(&new_poof.bitmap, name);
+					if (poof_new) {
+						required_string("$Bitmap:");
+						stuff_string(name, F_NAME, MAX_FILENAME_LEN);
+					} else {
+						optional_string("$Bitmap:");
+						stuff_string(name, F_NAME, MAX_FILENAME_LEN);
+					}
```
Baezon, line 257: These blocks appear to be the same; typo?

BMagnu (changes requested):
Okay, in general, this is a good first step.
But here comes the big but:
Nebula poofs are indexed using bitflags, not anything reasonable. This means, that -neb.tbm's behaviour will be depending on loading order. This is, IMO, not really acceptable.
I suspect, to make this really work well without someone thinking "why did all my poofs just break" is to de-bitfieldify the poofs at least in mission files, similar to how #4282 did with player orders

code/nebula/neb.cpp
```diff
@@ -163,22 +163,32 @@ void neb2_get_eye_orient(matrix *eye_matrix);
 // NEBULA FUNCTIONS
 //
 
-// initialize neb2 stuff at game startup
-void neb2_init()
+static poof_info* get_nebula_pointer(char* nebula_name)
```
Baezon, line 166: Minor, but specify that this function is referring to poofs, specifically. `get_nebula_poof_pointer` or something.

code/nebula/neb.cpp
```diff
@@ -191,62 +201,98 @@ void neb2_init()
 
 		// poofs
 		while (required_string_one_of(3, "#end", "+Poof:", "$Name:")) {
-
+			bool create_if_not_found = true;
 			if (Poof_info.size() < MAX_NEB2_POOFS) {
```
Baezon, line 205: This will need to be moved somewhere else I think. Even if we are at max poofs, these entries could merely be modifying them, which should still be ok. I guess where the `push_backs` are actually, happening just skip them in that case?

Baezon (changes requested):

MjnMixael:
Well there was an attempt... I didn't think to check that. I can probably follow the PR you listed as an example and do the same.

Baezon:
Nebula's referred to by bitflags by *missions*. New style poofs are referred to by name like everything so that shouldn't be a problem. With Mjn's current set up, old style poofs are purely additive, and modifications are based on the name, which for old style poofs is the same as the bitmap, which I think does make them ambiguous, but I think overall this is fine.

code/nebula/neb.cpp
MjnMixael: No. Bitmap is required, so if we're making a new poof, require the string. Otherwise, it's optional.

code/nebula/neb.cpp
MjnMixael: Ah yeah. Good catch.

BMagnu:
> Poofs are referred to by bitflags by _missions_. In the tables new style poofs are referred to by name like everything else so that shouldn't be a problem. With Mjn's current set up, old style poofs are purely additive, and modifications are based on the name, which for old style poofs is the same as the bitmap, which I think does make them ambiguous, but I think overall this is fine.

Yes, but if you have a missionfile that wants to display the poof at bit 5, then what that poof actually is will depend on the loading order, and for that the missions also need to refer by name. Or am I missing something?

MjnMixael:
You're right. By adding modular tables, this PR removes the guarantee of poof load order from a single table file. So to ensure that behavior doesn't change between game instances and what order the table files are parsed, the missions need to refer to poof by name.

code/nebula/neb.cpp
MjnMixael: Done

code/nebula/neb.cpp
MjnMixael: and done

code/nebula/neb.cpp
```diff
@@ -191,62 +201,109 @@ void neb2_init()
...
+				strcpy_s(pooft.name, name);
 
-					strcpy_s(new_poof.name, name);
+				generic_anim_init(&pooft.bitmap, name);
 
-					generic_anim_init(&new_poof.bitmap, name);
+				if (Poof_info.size() < MAX_NEB2_POOFS) {
+					Poof_info.push_back(pooft);
+				} else {
+					mprintf(("Warning: more than 32 poofs are defined! Skipping poof %s\n", pooft.name));
```
Baezon, line 220: This (and the similar check below) should probably be an actual warning

Baezon (changes requested):

code/nebula/neb.cpp
MjnMixael: Done

Baezon (approved):
I'm happy, although someone more well-versed in FRED should look at those parts too.

code/nebula/neb.cpp
```diff
@@ -191,62 +201,109 @@ void neb2_init()
...
+						continue;
+					}
+				}
 
+				if (poof_new) {
 					required_string("$Bitmap:");
 					stuff_string(name, F_NAME, MAX_FILENAME_LEN);
+				} else {
+					optional_string("$Bitmap:");
+					stuff_string(name, F_NAME, MAX_FILENAME_LEN);
```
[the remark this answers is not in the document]
MjnMixael: Are you wanting to prevent someone adding `$Bitmap: ` with no string defined afterwards here? (Essentially trying to clear the bitmap in a tbm?)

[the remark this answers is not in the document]
BMagnu: The opposite.
Right now, you could not specify $Bitmap (if its a nocreate entry), but it would still try to parse the filename after the nonexistant token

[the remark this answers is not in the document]
MjnMixael: Oh. Right. I see what you're saying now. Yeah. Will do.

BMagnu (approved):

PR #4836, opened 2022-10-28
Fix in game joining 

JohnAFernandez:
This has been a team effort with @notimaginative and @wookieejedi, working on this on and off for about 2 years.  Based on recent tests, this should fix all the final bugs.  This also makes in-game joining default on, which is what people expect these days, honestly.

Just waiting on some final tests to ensure everything is working.

### DIFF
--- a/code/globalincs/flagset.h
+++ b/code/globalincs/flagset.h
@@ -150,28 +150,23 @@ class flagset {
 	void set_from_vector(const SCP_vector<ubyte>& vector_in) 
 	{
 		size_t flag = 0; 
-		int bitshift = 0;
 
 		// just iterate through the entries, and set flags as appropriate.
 		for (auto& entry : vector_in) {
 
-			if (entry & (1 << bitshift)) {
-				values.set(flag, true);
-			}
+			for (uint bitshift = 0; bitshift < UBYTE_SIZE; bitshift++){
+				if (entry & (1 << bitshift)) {
+					values.set(flag, true);
+				}
 
-			++flag;
+				++flag;
 
-			// this function is used in multi where the size of flagsets can be different
-			// on server and client.  Small differences in flagset sizes should not affect
-			// gameplay, but we have to make sure not to exceed the size of the flagset.
-			if (flag >= SIZE){
-				return;
-			} 
-
-			++bitshift;
-
-			if (bitshift == UBYTE_SIZE){
-				bitshift = 0;
+				// this function is used in multi where the size of flagsets can be different
+				// on server and client.  Small differences in flagset sizes should not affect
+				// gameplay, but we have to make sure not to exceed the size of the flagset.
+				if (flag >= SIZE){
+					return;
+				} 
 			}
 		}
 	}
--- a/code/globalincs/flagset.h
+++ b/code/globalincs/flagset.h
@@ -6,6 +6,8 @@
 
 #include "osapi/dialogs.h"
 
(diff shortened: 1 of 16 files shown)


wookieejedi:
Tests with coop, dogfight, and TvT all work, which includes testing of joining in-game, leaving, then joining in-game again.

code/globalincs/flagset.h
```diff
@@ -114,6 +116,66 @@ class flagset {
...
+			if (flag >= SIZE){
+				return;
+			} 
+
+			++bitshift;
+
+			if (bitshift == UBYTE_SIZE){
+				bitshift = 0;
+			}
+		}
```
BMagnu, line 176: Would this not also need to iterate over the bits in each incoming byte?
This reads to me as if this would only check bit (n mod 8) of byte n for all bytes.

code/network/multi_ingame.cpp
```diff
@@ -1127,13 +1139,35 @@ void send_ingame_ships_packet(net_player *player)
...
+
+		for (auto& item : ship_flags_out){
+			ADD_DATA(item);
+		} 
+
+		check = shipp->flags.to_ubyte_vector(ship_flags_out);
+
+		Assertion(check <= 255, "Somehow FSO thinks there are too many object flags, %d * 8 to be exact. Please report!", static_cast<int>(check));
+
+		temp_size = static_cast<ubyte>(Objects[so->objnum].flags.to_ubyte_vector(obj_flags_out));
```
BMagnu, line 1164: This looks like a copy-paste error. Should it not look like above:
```suggestion
		check = Objects[so->objnum].flags.to_ubyte_vector(obj_flags_out);

		Assertion(check <= 255, "Somehow FSO thinks there are too many object flags, %d * 8 to be exact. Please report!", static_cast<int>(check));

		temp_size = static_cast<ubyte>(check);
```

BMagnu (changes requested):

code/globalincs/flagset.h
JohnAFernandez: You're right, I need an inner loop.

code/network/multi_ingame.cpp
JohnAFernandez: Yes, luckily this only affected error checking.

code/globalincs/flagset.h
```diff
@@ -114,6 +116,61 @@ class flagset {
...
+	}
+
+	void set_from_vector(const SCP_vector<ubyte>& vector_in) 
+	{
+		size_t flag = 0; 
+
+		// just iterate through the entries, and set flags as appropriate.
+		for (auto& entry : vector_in) {
+
+			for (uint bitshift = 0; bitshift < sizeof(vector_in.front()); bitshift++){
```
BMagnu, line 157: err, ``sizeof(vector_in.front())`` will be 1, no? since it's in bytes, not bit. I'd expect this to need to be ``UBYTE_SIZE ``

BMagnu (changes requested):

code/globalincs/flagset.h
JohnAFernandez: Thanks, I rarely use sizeof

BMagnu (approved):

wookieejedi (commented):
Waiting on a few more changes and tests still.

wookieejedi (changes requested):
Will be approved once test and changes are worked out.

BMagnu (approved):

PR #4839, opened 2022-10-30
add a few scripting dock functions

Goober5000:
Add setDocked, setUndocked, and jettison.  All have been tested.

### DIFF
--- a/code/scripting/api/objs/ship.cpp
+++ b/code/scripting/api/objs/ship.cpp
@@ -24,6 +24,7 @@
 #include "ship/shipfx.h"
 #include "hud/hudets.h"
 #include "object/object.h"
+#include "object/objectdock.h"
 #include "model/model.h"
 #include "ship/ship.h"
 #include "parse/parselo.h"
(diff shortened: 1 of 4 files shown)


code/scripting/api/objs/ship.cpp
```diff
@@ -2337,6 +2338,147 @@ ADE_FUNC(setGlowPointBankActive, l_Ship, "boolean active, [number bank]", "Activ
...
 
+ADE_FUNC(setDocked, l_Ship, "ship dockee_ship, [string | number docker_point, string | number dockee_point]", "Immediately docks this ship with another ship.", nullptr, "Returns nothing")
+{
+	object_h *docker_objh = nullptr, *dockee_objh = nullptr;
+	int docker_point = -1, dockee_point = -1;
+
+	if (!ade_get_args(L, "oo", l_Ship.GetPtr(&docker_objh), l_Ship.GetPtr(&dockee_objh)))
+		return ADE_RETURN_NIL;
+
+	if (docker_objh == nullptr || dockee_objh == nullptr || !docker_objh->IsValid() || !dockee_objh->IsValid())
```
BMagnu, line 2349: This should also check for equality of docker and dockee, since that will most likely result in a very invalid state

code/scripting/api/objs/ship.cpp
```diff
@@ -2337,6 +2338,147 @@ ADE_FUNC(setGlowPointBankActive, l_Ship, "boolean active, [number bank]", "Activ
...
+		return ADE_RETURN_NIL;
+	}
+
+	//Make sure that the specified dockpoints are all free (if not, do nothing)
+	if (dock_find_object_at_dockpoint(docker_objh->objp, docker_point) != nullptr ||
+		dock_find_object_at_dockpoint(dockee_objh->objp, dockee_point) != nullptr)
+	{
+		nprintf(("scripting", "Cannot dock %s to %s because at least one of the specified dockpoints is occupied", docker_shipp->ship_name, dockee_shipp->ship_name));
+		return ADE_RETURN_NIL;
+	}
```
BMagnu, line 2408: Since this can be a somewhat common failure case, I'd suggest making the function return a bool.
If we have an invalid input, just return nil as you do, if the docking fails due to circumstances of the surrounding state, return false, and if it succeeds, return true

code/scripting/api/objs/ship.cpp
```diff
@@ -2337,6 +2338,147 @@ ADE_FUNC(setGlowPointBankActive, l_Ship, "boolean active, [number bank]", "Activ
...
+		while (object_is_docked(docker_objh->objp))
+		{
+			object_jettison_cargo(docker_objh->objp, dock_get_first_docked_object(docker_objh->objp), jettison_speed, true);
+		}
+	}
+
+	return ADE_RETURN_NIL;
+}
+
+ADE_FUNC(setUndocked, l_Ship, "[ship dockee_ship]", "Immediately undocks this ship from another ship or from an arbitrary number of ships - this function can accept an arbitrary number of ship arguments.", nullptr, "Returns nothing")
```
BMagnu, line 2455: This docstring is ambiguous in regards to "this / another ship", especially considering that usually, "this" would need to refer to the object on which the function is called . I suggest both this and jettison to be reworded like "undocks / jettisons dockee from this ship". Also, the docstring needs to explictly mention the special mode for no ships listed, as well as specify varargs, like ``[ship... dockee_ships /* All docked ships by default */]``

code/scripting/api/objs/ship.cpp
```diff
@@ -2337,6 +2338,147 @@ ADE_FUNC(setGlowPointBankActive, l_Ship, "boolean active, [number bank]", "Activ
...
+	if (!ade_get_args(L, "o", l_Ship.GetPtr(&docker_objh)))
+		return ADE_RETURN_NIL;
+
+	if (docker_objh == nullptr || !docker_objh->IsValid())
+		return ADE_RETURN_NIL;
+
+	return jettison_helper(L, docker_objh, 0.0f, 1);
+}
+
+ADE_FUNC(jettison, l_Ship, "number jettison_speed, [ship dockee_ship]", "Jettisons one or more ships at the specified speed - this function can accept an arbitrary number of ship arguments.", nullptr, "Returns nothing")
```
BMagnu, line 2468: See above

code/scripting/api/objs/ship.cpp
```diff
@@ -2337,6 +2338,147 @@ ADE_FUNC(setGlowPointBankActive, l_Ship, "boolean active, [number bank]", "Activ
...
+	{
+		// Goober5000 - as with ai_deathroll_start, we can't simply iterate through the dock list while we're
+		// undocking things.  So just repeatedly jettison the first object.
+		while (object_is_docked(docker_objh->objp))
+		{
+			object_jettison_cargo(docker_objh->objp, dock_get_first_docked_object(docker_objh->objp), jettison_speed, true);
+		}
+	}
+
+	return ADE_RETURN_NIL;
```
BMagnu, line 2452: Could / should this return the number of objects undocked, especially interesting in the "no args specified"-case?

BMagnu (changes requested):

code/scripting/api/objs/ship.cpp
Goober5000: That's a good idea.

code/scripting/api/objs/ship.cpp
Goober5000: Hmm...

BMagnu (approved):

PR #4842, opened 2022-10-31
Clearable Ship/Subsystem Targets via script

Baezon:
Previously, setting `Target` for either would have been ignored entirely.

### DIFF
--- a/code/scripting/api/objs/ship.cpp
+++ b/code/scripting/api/objs/ship.cpp
@@ -781,6 +781,7 @@ ADE_VIRTVAR(Target, l_Ship, "object", "Target of ship. Value may also be a deriv
 			aip->target_objnum = OBJ_INDEX(newh->objp);
 			aip->target_signature = newh->sig;
 			aip->target_time = 0.0f;
+			set_targeted_subsys(aip, nullptr, -1);
 
 			if (aip == Player_ai)
 				hud_shield_hit_reset(newh->objp);
@@ -788,9 +789,8 @@ ADE_VIRTVAR(Target, l_Ship, "object", "Target of ship. Value may also be a deriv
 			aip->target_objnum = -1;
 			aip->target_signature = -1;
 			aip->target_time = 0.0f;
+			set_targeted_subsys(aip, nullptr, -1);
 		}
-
-		set_targeted_subsys(aip, NULL, -1);
 	}
 
 	return ade_set_object_with_breed(L, aip->target_objnum);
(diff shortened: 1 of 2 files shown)


code/scripting/api/objs/ship.cpp
```diff
@@ -775,28 +775,22 @@ ADE_VIRTVAR(Target, l_Ship, "object", "Target of ship. Value may also be a deriv
...
+			aip->target_time = 0.0f;
 
-			set_targeted_subsys(aip, NULL, -1);
+			if (aip == Player_ai)
+				hud_shield_hit_reset(newh->objp);
+		} else if (lua_isnil(L, 2)) {
+			aip->target_objnum = -1;
+			aip->target_signature = -1;
+			aip->target_time = 0.0f;
 		}
```
BMagnu, line 791: If neither of these two is the case (i.e., an invalid target object or not a target object) there should probably be a lua error. Or at least, it should return from the function instead of setting some targeted subsys to null.

code/scripting/api/objs/ship.cpp
```diff
@@ -775,28 +775,22 @@ ADE_VIRTVAR(Target, l_Ship, "object", "Target of ship. Value may also be a deriv
...
+
+		set_targeted_subsys(aip, NULL, -1);
```
BMagnu, line 793: nullptr

code/scripting/api/objs/subsystem.cpp
```diff
@@ -434,12 +434,18 @@ ADE_VIRTVAR(Target, l_Subsystem, "object", "Object targeted by this subsystem. I
...
+			ss->turret_enemy_objnum = OBJ_INDEX(objh->objp);
+			ss->turret_enemy_sig = objh->sig;
+			ss->targeted_subsys = nullptr;
+			ss->scripting_target_override = true;
+		} else if (lua_isnil(L, 2)) {
+			ss->turret_enemy_objnum = -1;
+			ss->turret_enemy_sig = -1;
+			ss->targeted_subsys = nullptr;
+			ss->scripting_target_override = true;
+		}
```
BMagnu, line 448: Should you decide to throw a lua error for the above issue, this would need to as well

BMagnu (changes requested):

code/scripting/api/objs/ship.cpp
Baezon: Ah, yes, `set_targeted_subsys` should be inside the blocks, but I didn't want to introduce errors for a situation where before it would've done nothing. I can if that's preferred, but that was my thinking.

BMagnu: Okay, then probably ficing the set_targeted_subsys but not throwing any errors is probably best

BMagnu (approved):

PR #4843, opened 2022-10-31
Modular tips

MjnMixael:
Vectorize tips to remove the limit of 40 and make the table extendable with -tip.tbm

### DIFF
--- a/code/menuui/playermenu.cpp
+++ b/code/menuui/playermenu.cpp
@@ -1370,16 +1370,12 @@ DCF(bastion, "Temporarily sets the player to be on the Bastion (or any other mai
 	}
 }
 
-#define MAX_PLAYER_TIPS			40
-
-SCP_string Player_tips[MAX_PLAYER_TIPS];
-int Num_player_tips;
+SCP_vector<SCP_string> Player_tips;
 int Player_tips_shown = 0;
 
 // tooltips
 void player_tips_init()
 {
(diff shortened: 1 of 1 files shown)


code/menuui/playermenu.cpp
```diff
@@ -1370,33 +1370,25 @@ DCF(bastion, "Temporarily sets the player to be on the Bastion (or any other mai
...
-			if (Num_player_tips >= MAX_PLAYER_TIPS) {
-				break;
-			}
-
 			SCP_string buf;
 			stuff_string(buf, F_MESSAGE);
-
-			Player_tips[Num_player_tips++] = message_translate_tokens(buf.c_str());
+			
+			Player_tips.push_back(message_translate_tokens(buf.c_str()));
```
JohnAFernandez, line 1391: You shouldn't need to use a c_str() here because the vector already holds SCP_string.

JohnAFernandez (changes requested):

code/menuui/playermenu.cpp
MjnMixael: That's for `message_translate_tokens` which expects a char*

JohnAFernandez: That's what I get for reading too quick.  Nifty that it returns an SCP string.

JohnAFernandez (approved):

PR #4844, opened 2022-11-01
fix timestamp_in_between functions

Goober5000:
Fix the comparison and add documentation to the compare functions.

Thanks to @JohnAFernandez for noticing the bug.


JohnAFernandez (approved):
Fingers crossed we weren't depending on buggy behavior over in multi where `timestamp_in_between` is used.

PR #4845, opened 2022-11-01
Modular iff table

MjnMixael:
With unlimited IFFs, it'd be nice to be able to modify and organize them using modular table features, so this implements that functionality.

### DIFF
--- a/code/iff_defs/iff_defs.cpp
+++ b/code/iff_defs/iff_defs.cpp
@@ -98,15 +98,16 @@ int iff_init_color(int r, int g, int b)
 	return idx;
 }
 
-char traitor_name[NAME_LENGTH];
-SCP_vector<SCP_vector<SCP_string>> attack_names;
+//Initialize this as the retail default - Mjn
+static char traitor_name[NAME_LENGTH] = "Traitor";
+static SCP_vector<SCP_vector<SCP_string>> attack_names;
 
 struct observed_color_t {
 	char iff_name[NAME_LENGTH];
 	int color_index;
 };
 
-SCP_vector<SCP_vector<observed_color_t>> observed_color_table;
+static SCP_vector<SCP_vector<observed_color_t>> observed_color_table;
 
 void resolve_iff_data()
 {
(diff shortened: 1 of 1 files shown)


code/iff_defs/iff_defs.cpp
```diff
@@ -289,30 +399,56 @@ void iff_init()
...
-			// get required IFF info ----------------------------------------------
 
 			// get the iff name
 			required_string("$IFF Name:");
-			stuff_string(iff->iff_name, F_NAME, NAME_LENGTH);
+			stuff_string(ifft.iff_name, F_NAME, NAME_LENGTH);
+			mprintf(("IFFs got to name %s\n", ifft.iff_name));
+			if (optional_string("+nocreate")) {
+				if (!Parsing_modular_table) {
+					Warning(LOCATION, "+nocreate flag used for iff in non-modular table\n");
```
BMagnu, line 409: Since we are in table parse, use ``error_display(0, "Message");`` for warnings over ``Warning``, as this'll log the exact line the warning came from to the user

code/iff_defs/iff_defs.cpp
```diff
@@ -289,30 +399,56 @@ void iff_init()
...
-			stuff_int_list(rgb, 3, RAW_INTEGER_TYPE);
-			iff->color_index = iff_init_color(rgb[0], rgb[1], rgb[2]);
+			if (optional_string("$Colour:")) {
+				stuff_int_list(rgb, 3, RAW_INTEGER_TYPE);
+				iffp->color_index = iff_init_color(rgb[0], rgb[1], rgb[2]);
+			}
+			if (optional_string("$Color:")) {
+				stuff_int_list(rgb, 3, RAW_INTEGER_TYPE);
+				iffp->color_index = iff_init_color(rgb[0], rgb[1], rgb[2]);
+			}
```
BMagnu, line 451: This should probably be an optional string either to avoid code repetition here. Also, what will this do in a modular table where color isn't defined? I think this also needs an else if not nocreate, then error:
```suggestion
			if (optional_string_either("$Colour:", "$Color:") != -1) { 
				stuff_int_list(rgb, 3, RAW_INTEGER_TYPE);
				iffp->color_index = iff_init_color(rgb[0], rgb[1], rgb[2]);
			}
			else if (create_if_not_found) {
				error_display(1, "Error:  IFF %s must be specified a color.", ifft.iff_name);
			}
```

BMagnu (changes requested):

code/iff_defs/iff_defs.cpp
MjnMixael: I think I initialized the color to white to make sure it has data if this option was skipped in a modular table. Will double check when I make the other changes.

MjnMixael: It sets color as white when a new IFF is being parsed at line 436 and then overwrites that color here if it's defined. All other changes made.

BMagnu: I see, should be fine if documented as new default then.

BMagnu (approved):

PR #4846, opened 2022-11-01
Add `isMultipart` and `getModelFlag` functions on subsystems

Baezon:
`getModelFlag` following the  same general API as `get/setFlag` on ships.


wookieejedi (approved):
Definitely useful additions for scripting, looks good.

PR #4849, opened 2022-11-02
increase object limits

MjnMixael:
Increases limits as discussed in #SCP. I'm in no big hurry to get this committed, but submitting the PR to ensure the discussion is finished and a decision is made on the best numbers to use.

### DIFF
--- a/code/globalincs/globals.h
+++ b/code/globalincs/globals.h
@@ -32,8 +32,8 @@
 #define SHORT_CALLSIGN_PIXEL_W	80		// max width of short_callsign[] in pixels
 
 // from ship.h
-#define MAX_SHIPS					400			// max number of ship instances there can be.DTP; bumped from 200 to 400
-#define SHIPS_LIMIT					400			// what MAX_SHIPS will be at release time (for error checking in debug mode); dtp Bumped from 200 to 400
+#define MAX_SHIPS					500		// max number of ship instances there can be.DTP; bumped from 200 to 400, then to 500 in 2022
+#define SHIPS_LIMIT					500		// what MAX_SHIPS will be at release time (for error checking in debug mode); dtp Bumped from 200 to 400, then to 500 in 2022
 
 // from missionparse.h and then redefined to the same value in sexp.h
 #define TOKEN_LENGTH	32
@@ -61,7 +61,7 @@
 #define MAX_COMPLETE_ESCORT_LIST	20
              
 // from weapon.h
-#define MAX_WEAPONS	2000
+#define MAX_WEAPONS	3000		//Increased from 2000 to 3000 in 2022
 
 #define MAX_WEAPON_TYPES				500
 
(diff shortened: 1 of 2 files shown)


wookieejedi:
This increase is very reasonable and would be highly utilized by FotG for one.

wookieejedi:
I also might recommend putting comments after all the bumped sections along with the even perhaps the year?

BMagnu:
IMO, these increases are proper like this. But yeah, the comments for the limits would still need to be edited

wookieejedi (approved):

code/globalincs/globals.h
```diff
@@ -32,7 +32,7 @@
 #define SHORT_CALLSIGN_PIXEL_W	80		// max width of short_callsign[] in pixels
 
 // from ship.h
-#define MAX_SHIPS					400			// max number of ship instances there can be.DTP; bumped from 200 to 400
+#define MAX_SHIPS					1000		// max number of ship instances there can be.DTP; bumped from 200 to 400, then to 1000 in 2022
 #define SHIPS_LIMIT					400			// what MAX_SHIPS will be at release time (for error checking in debug mode); dtp Bumped from 200 to 400
```
EatThePath, line 36: do we need to update here too?

EatThePath (commented):
Also should we consider bumping the ship class limit for Event Horizon while we're at it?

code/globalincs/globals.h
MjnMixael: Good catch. I think you're right.

notimaginative:
Just to point out, in case it wasn't considered, this adds nearly 8 MiB onto the stack. The `MAX_SHIPS` bump alone counts for about 5 MiB of that.

Goober5000:
Yeah, these bumps are pretty substantial.  Can we get away with smaller bumps?  IIRC the number tossed around for `MAX_SHIPS` was 500, not 1000.

MjnMixael:
Alright. Set to 500 for ships now.

code/globalincs/globals.h
```diff
@@ -32,8 +32,8 @@
 #define SHORT_CALLSIGN_PIXEL_W	80		// max width of short_callsign[] in pixels
 
 // from ship.h
-#define MAX_SHIPS					400			// max number of ship instances there can be.DTP; bumped from 200 to 400
-#define SHIPS_LIMIT					400			// what MAX_SHIPS will be at release time (for error checking in debug mode); dtp Bumped from 200 to 400
+#define MAX_SHIPS					500		// max number of ship instances there can be.DTP; bumped from 200 to 400, then to 1000 in 2022
+#define SHIPS_LIMIT					500		// what MAX_SHIPS will be at release time (for error checking in debug mode); dtp Bumped from 200 to 400, then to 1000 in 2022
```
wookieejedi, line 36: Comments needs to be updated to 500 instead of 1000 :)

wookieejedi (changes requested):

code/globalincs/globals.h
MjnMixael: details... details...

Goober5000:
How about 3000 for weapons and 4500 for objects?

wookieejedi:
I might say at least 5000 for objects, for example to account for the higher asteroid increase.

MjnMixael:
Changed to 3000 and 5000 respectively

wookieejedi (approved):

wookieejedi:
Given the new changes and overall good compromise I'm going to merge this in a day or so (pending any objections).

PR #4850, opened 2022-11-02
Minor lighting shader fixes

qazwsxal:
While working on shader culling, @EatThePath noticed some issues with rough dielectric surfaces being too shiny at sharp angles. Through a process of elimination he narrowed it down to the geometry term using the wrong equation for surface roughness. 

While fixing, I also noticed the `area_normalisation` term that stops wide beams from being too bright wasn't actually in use, so that's been applied to deferred lighting.

 To prevent further mistakes, there's some minor cleanup to make it more clear what we're doing. (This also gets rid of a sqrt() call, so hopefully there's some optimisation there)


wookieejedi (approved):
Tested and everything seems to work well for me. Nice cleanup and fix.

PR #4851, opened 2022-11-02
Modular medals

MjnMixael:
Makes medals.tbl extendable with -mdl.tbm. Ended up being a bigger PR than expected because it fixes a number of related issues to make it easier to deal with modular medals including better handling of kill count badge promotion to remove requiring badges be tabled in order of kill count, better handling of finding the Rank medal, adding the ability to specify a mask value for each medal, and entirely removing `Num_medals` in favor of `Medals.size()`.

### DIFF
--- a/code/stats/medals.cpp
+++ b/code/stats/medals.cpp
@@ -209,7 +209,7 @@ static medal_stuff* get_medal_pointer(char* medal_name)
 	return nullptr;
 }
 
-static int get_medal_position(char* medal_name)
+static int get_medal_position(const char* medal_name)
 {
 	for (int i = 0; i < (int)Medals.size(); i++) {
 		if (!stricmp(medal_name, Medals[i].name)) {
--- a/code/stats/medals.cpp
+++ b/code/stats/medals.cpp
@@ -283,7 +283,6 @@ void parse_medals_table(const char* filename)
 			medal_display_info display_t;
 			medal_display_info* display_p;
 			bool create_if_not_found = true;
-			bool new_medal = true;
 
 			required_string("$Name:");
 			stuff_string(medal_t.name, F_NAME, NAME_LENGTH);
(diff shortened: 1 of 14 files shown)


code/stats/medals.cpp
```diff
@@ -199,13 +197,35 @@ const char* medal_stuff::get_display_name() const {
 	}
 }
 
-void parse_medal_tbl()
+static medal_stuff* get_medal_pointer(char* medal_name)
```
BMagnu, line 200: this should take a const char*

code/stats/medals.cpp
```diff
@@ -236,128 +256,158 @@ void parse_medal_tbl()
...
-				temp_display.coords[GR_640].y = Default_medal_coords[GR_640][Num_medals][1];
+			else if (Medals.size() < NUM_MEDALS_FS2) {
+				display_p->coords[GR_640].x = Default_medal_coords[GR_640][Medals.size()][0];
+				display_p->coords[GR_640].y = Default_medal_coords[GR_640][Medals.size()][1];
 			}
 			else {
-				Warning(LOCATION, "No default GR_640 position for medal '%s'!", temp_medal.name);
-				temp_display.coords[GR_640].x = 0;
-				temp_display.coords[GR_640].y = 0;
+				Warning(LOCATION, "No default GR_640 position for medal '%s'!", medal_p->name);
```
BMagnu, line 343: Since this is during table parsing, ``error_display(0, "");`` is preferable to Warning

code/stats/medals.cpp
```diff
@@ -366,44 +416,25 @@ void parse_medal_tbl()
 					if (optional_string("+Persona:")) {
 						stuff_int(&persona);
 						if (persona < 0) {
-							Warning(LOCATION, "Debriefing text for %s is assigned to an invalid persona: %i (must be 0 or greater).\n", temp_medal.name, persona);
+							Warning(LOCATION,
```
BMagnu, line 419: same here

BMagnu (changes requested):

PR #4852, opened 2022-11-02
Allow overriding `gvec` in `fireWeapon` on subsystems

Baezon:
This way, singlepart turrets can be directed to fire in a given direction, instead of purely straight out from the turret normal.


wookieejedi (commented):
Should the function documentation be updated with this option?

wookieejedi (approved):

PR #4854, opened 2022-11-03
Modular rank table

MjnMixael:
Free Ranks from their retail shackles. Vectorizes Ranks and adds bounds checks to places where it's accessed. Removes the limit of 10 and assumptions that ranks will match retail's 10. UI elements that display rank will check if the player rank > than the current max. If so, display the max without changing player rank. Adds modular -rnk.tbm parsing with +nocreate support. 

Disallows Ranks to have equal values and will automatically increment one of them by 1 point if detected and log the change. Will then auto sort ranks after all parsing is done if Ranks is not ordered by point values or Rank equality is detected. Will warn if auto sorting is done and log the changes.

### DIFF
--- a/code/pilotfile/pilotfile.cpp
+++ b/code/pilotfile/pilotfile.cpp
@@ -245,7 +245,7 @@ void pilotfile::update_stats_backout(scoring_struct *stats, bool training)
 
 	if (stats->m_promotion_earned >= 0) {
 		// deal with a multi-rank promotion mission
-		for (i = 0; i < MAX_FREESPACE2_RANK; ++i) {
+		for (i = 0; i < ((int)Ranks.size() - 1); ++i) {
 			if (p_stats->score >= Ranks[i].points) {
 				p_stats->rank = i;
 			}
--- a/code/pilotfile/pilotfile.cpp
+++ b/code/pilotfile/pilotfile.cpp
@@ -245,7 +245,7 @@ void pilotfile::update_stats_backout(scoring_struct *stats, bool training)
 
 	if (stats->m_promotion_earned >= 0) {
 		// deal with a multi-rank promotion mission
-		for (i = 0; i < ((int)Ranks.size() - 1); ++i) {
+		for (i = 0; i < ((int)Ranks.size()); ++i) {
 			if (p_stats->score >= Ranks[i].points) {
 				p_stats->rank = i;
 			}
--- a/code/stats/scoring.cpp
+++ b/code/stats/scoring.cpp
@@ -46,7 +46,7 @@ extern debriefing Traitor_debriefing;
 traitor_stuff Traitor;
 
 // these tables are overwritten with the values from rank.tbl
-rank_stuff Ranks[NUM_RANKS];
+SCP_vector<rank_stuff> Ranks;
 
 // scoring scale factors by skill level
 float Scoring_scale_factors[NUM_SKILL_LEVELS] = {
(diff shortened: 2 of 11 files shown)


code/stats/scoring.cpp
```diff
@@ -57,32 +57,111 @@ float Scoring_scale_factors[NUM_SKILL_LEVELS] = {
...
+			} else {
+				// Don't create rank if it has +nocreate and is in a modular table.
+				if (!create_if_not_found && Parsing_modular_table) {
+					if (!skip_to_start_of_string_either("$Name:", "#end")) {
+						error_display(1, "Missing [#end] or [$Name] after rank %s", rank_t.name);
+					}
+					continue;
+				}
+				Ranks.push_back(rank_t);
+				rank_p = &Ranks[Ranks.size() - 1];
```
BMagnu, line 136: preferredly, this'd be ``Ranks.back()``

code/stats/scoring.cpp
```diff
@@ -130,6 +207,75 @@ void parse_rank_tbl()
...
+				Ranks[j].points++;
+				shouldSort = true;
+				mprintf(("Rank %s and %s are equal! Increasing %s to %i\n",
+					Ranks[i].name,
+					Ranks[j].name,
+					Ranks[j].name,
+					Ranks[j].points));
+			}
+		}
+	}
```
BMagnu, line 227: This is insufficient.
Assume you add ranks with points 2, 1 and 1.
It'll check the 2 with both 1's and be okay.
Then it'll check the first 1 and increase the second 1 to 2, but that'll never be checked against the first 2 again.
I recommend just erroring on two identical scores.

BMagnu (changes requested):

code/stats/scoring.cpp
MjnMixael: Hmm. I thought I checked that case specifically but it's been months since. I'll give it all another look. I've learned a lot since, too.

code/pilotfile/pilotfile.cpp
```diff
@@ -245,7 +245,7 @@ void pilotfile::update_stats_backout(scoring_struct *stats, bool training)
 
 	if (stats->m_promotion_earned >= 0) {
 		// deal with a multi-rank promotion mission
-		for (i = 0; i < MAX_FREESPACE2_RANK; ++i) {
+		for (i = 0; i < ((int)Ranks.size() - 1); ++i) {
```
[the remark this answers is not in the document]
MjnMixael: No, that was a copy/paste error.

code/stats/scoring.cpp
MjnMixael: Yeah, I don't know what I was thinking when I wrote that. Good catch, though. I think this is probably solvable, but yeah let's go path of least resistance. Error applied.

BMagnu (approved):

code/stats/scoring.cpp
```diff
@@ -1362,7 +1505,7 @@ float scoring_get_scale_factor()
 void scoring_bash_rank(player *pl,int rank)
 {	
 	// if this is an invalid rank, do nothing
-	if((rank < RANK_ENSIGN) || (rank > RANK_ADMIRAL)){
```
Goober5000, line 1365: Off-by-one error in the code change here.  The > should now be >=.

code/pilotfile/pilotfile.cpp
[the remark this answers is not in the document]
Goober5000: This actually looks like it was an error in the original code.

Goober5000 (changes requested):

code/stats/scoring.cpp
MjnMixael: fixed

Goober5000 (approved):

PR #4856, opened 2022-11-04
refactor mission initialization

Goober5000:
Gather all the mission initialization code into a common function for clarity and ease of expansion.  Previously, the various parsing routines contained initialization as well as parsing.  As a bonus, quite a few redundant initializations are streamlined.

~~This is in draft status pending testing.~~


wookieejedi (approved):
My testing with both FRED and FSO shows everything is working as expected, though would still likely be valuable to have perhaps another set of eyes for anyone else interested.

PR #4860, opened 2022-11-05
set-asteroid-field and set-debris-field sexps

### LINKED ISSUE #2554: FR: toggle-asteroids sexp
I would find it useful to be able to toggle asteroid fields on/off for in-mission jumps. I know it generally works well enough to just jump your player and ships to an area far away from the asteroid field. However I have seen instances of AI deciding to make it a priority to fly 20k meters back to the asteroid field to shoot at asteroids. Being able to remove it from the mission would prevent that from occurring.

While my specific request is more to be able to turn asteroids off... I can forsee it being useful to also turn them on.

I suspect this sexp request may not be as simple as it seems.. so I'm open to suggestions on how to actually pull it off. The real goal is to stop the AI from even caring about asteroids that are so far away that you can't see them. So maybe there's a better idea on how to do that than toggling the field on/off.

### LINKED ISSUE #2932: Manage asteroid fields dynamically during a mission via SEXPs
**Asteroid Field** parameters are fixed and cannot be changed during a mission. This leads to a number of relevant drawbacks in mission design.

For instance, I may wish to trigger the _Inner Box_ and more aggressive asteroids under specific circumstances, yet the parameter managing that particular setting is fixed for the entire mission and cannot be edited. A similar limit affects asteroid speed, types, size and so on. All parameters are pre-determined in FRED and cannot change during the mission.

Another major drawback is experienced should the mission have stages _other_ than the one involving the asteroid field itself. One such scenario is a mission where we jump to an asteroid field, or jump out from it at some point. While setting various stages across huge distances in FRED somehow helps "hiding" the asteroid field, the problem isn't solved at all. Asteroid fields can be noticed and extra measures have to be implemented in order to make them untargetable (e.g., HUD radar range limits). Also, the game still treats them as a valid feature of the mission even when though they're no longer needed, so the engine may experience unnecessary slowdowns by loading asteroids which no longer serve a purpose in the mission itself.

For this reason, I think one or more SEXPs could be used to manage all these parameters dynamically during a mission, and eventually cause an asteroid field to be created from scratch or disappear when the mission no longer requires it.

MjnMixael:
Adds a sexp to create an asteroid field mid-mission. Also does a few minor cleanup jobs in asteroid.cpp: prevents loading an asteroid pof if it's already loaded, properly sets asteroid pofs as unloaded on mission close, adds a warning if the min/max bounds are incorrect.

Fixes #2554

Fixes #2932


MjnMixael:
Note that I will follow this up with some asteroid.cpp cleanup that reintroduces the maybe_change_name function as part of a new display name feature.

code/parse/sexp.cpp
```diff
@@ -15073,6 +15075,327 @@ void sexp_toggle_asteroid_field(int n)
...
+	Asteroid_field.max_bound = vm_vec_new((float)o_maxx, (float)o_maxy, (float)o_maxz);
+
+	vec3d a_rad;
+	vm_vec_sub(&a_rad, &Asteroid_field.max_bound, &Asteroid_field.min_bound);
+	vm_vec_scale(&a_rad, 0.5f);
+	float b_rad = vm_vec_mag(&a_rad);
+
+	Asteroid_field.bound_rad = MAX(3000.0f, b_rad);
+
+	asteroid_create_all();
```
Baezon, line 15396: Relatively minor, but both these SEXPs are quite 'big'. It might be better to decouple the SEXP handling and the the actual asteroid/debris handling, so the actual deletion/making of the fields are handled in functions in their respective files, which these SEXPs call, and only worry about retrieving the right arguments for it. Also makes it easier to reuse this logic in different contexts in the future.

code/asteroid/asteroid.cpp
```diff
@@ -2040,7 +2089,9 @@ static void asteroid_parse_tbl()
 
 			asteroid_parse_section(&new_asteroid);
 
-			maybe_change_asteroid_name(&new_asteroid);
+			// I really don't think this is needed and actually causes a bug in FRED
+			// making it impossible to select some debris types-Mjn
+			//maybe_change_asteroid_name(&new_asteroid);
```
Baezon, line 2094: As mentioned, this is fine, but be sure to update the hud targetbox code https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/hud/hudtargetbox.cpp#L1246 to remove trailing nubmers, which is what I think what we agreed should be sufficient.

Baezon (changes requested):

code/parse/sexp.cpp
MjnMixael: Good point. Taken care of.

code/asteroid/asteroid.cpp
MjnMixael: I'll punt this to #4870 which re-introduces the whole feature with a distinct asteroid display name.

Baezon (approved):

PR #4861, opened 2022-11-06
More old dead shader override code that isn't hooked to anything anymore

EatThePath:
effectively a followup to #4823, working along similar lines but hitting a few things that were a tad less evident on my last sweep.

After this there are several elements of the model data shader uniform that are unused. Altering shader uniforms is something that can easily go wrong, and I'm not quite done with them yet, so for the moment I've marked all the dead ones' names to make them stick out and intend to do a followup very soon.


JohnAFernandez (approved):
giphy

PR #4862, opened 2022-11-06
improve SEXP syntax checking

Goober5000:
When the top-level operator of a SEXP is not found, flag the error instead of crashing.  This is helpful for missions that contain unsupported scripted SEXPs.


wookieejedi (approved):
Very helpful cleanup IMO

PR #4863, opened 2022-11-06
add numDocked and isDocked

Goober5000:
Two useful script functions.


wookieejedi (approved):
Quite useful, and overall looks good to me.

PR #4864, opened 2022-11-07
Shaders flag slaughter

EatThePath:
Also fixes the no-specular regression, after a fashion, and adds configurability for the default material properties of unset maps, and removes many of the compiler directive flags in shaders for simplicity and stuff.

They're all related changes, honest! 

In draft because none of these are actually done yet. The no-spec works, anyway. Not all the configuration is implemented. There's a squirrely bug remaining in skybox rendering. There's some dirty hacks in the implementation I can probably fix and some I probably can't. I may need to make another bmpman function to modify a texture in place to clean everything up.

But if you feel like looking at my mess, now you can!

### DIFF
--- a/cmake-variants.yaml
+++ b/cmake-variants.yaml
@@ -1,0 +1,16 @@
+buildType:
+  default: release
+  description: R
+  choices:
+    release:
+      short: Release
+      long: Release builds
+      buildType: Release
+    fastDebug:
+      short: FastDebug
+      long: Debug with optimizations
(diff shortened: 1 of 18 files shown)


The-E:
This PR is somewhat outdated following #5055, so there is some major rework needed. Closing this for now.

EatThePath:
agreed on the closing, will reoopen or open a fresh PR if and when I can bring the branch up to date.

PR #4865, opened 2022-11-07
clean up some obsolete fields

Goober5000:
Remove `behavior`, `type`, and `wing`, which are either unused or redundant.


wookieejedi (approved):
Everything worked as expected in my mission tests and overall looks good.

PR #4866, opened 2022-11-07
add wing formation align button

Goober5000:
This button allows wings to be arranged in their formation so that they don't scramble around when they arrive.

~~This is in draft status as it depends on #4865.~~


fred2/wing_editor.h
```diff
@@ -51,6 +51,7 @@ class wing_editor : public CDialog
 	int		m_waves;
 	int		m_threshold;
 	int		m_formation;
+	CString	m_formation_scale;
```
wookieejedi, line 54: This is nitpicky, but it would be nice looking if this was aligned with surrounding entries :)

Goober5000: If tabs are set to 4 spaces, it lines up. :)

wookieejedi (approved):
Looks good now, will be quite useful, too.

PR #4867, opened 2022-11-07
Refactor Lua Enums

BMagnu:
This PR refactors the lua enum system away from arbitrary indices to it being a proper enum. At the same time, it facilitates explicit bitfield values to allow for bitfield enums without the double-usage of indices as was the case previously.
This fixes a number of bugs associated with lua enums:
1. Bitfield enums now no longer test as equal to other unrelated enums.
2. Bitfield enums now display their proper name instead of an arbitrary unrelated one.
3. Combined bitfield enums now properly display which combination of enums they are.

### DIFF
--- a/code/scripting/api/objs/enums.h
+++ b/code/scripting/api/objs/enums.h
@@ -7,147 +7,181 @@
 #include "globalincs/pstypes.h"
 #include "scripting/ade_api.h"
 
+#include <tl/optional.hpp>
+
 namespace scripting {
 namespace api {
 
-const int32_t LE_ALPHABLEND_FILTER           = 14;
-const int32_t LE_ALPHABLEND_NONE             = 27;
-const int32_t LE_CFILE_TYPE_NORMAL           = 20;
-const int32_t LE_CFILE_TYPE_MEMORY_MAPPED    = 21;
-const int32_t LE_MOUSE_LEFT_BUTTON           = 1;
-const int32_t LE_MOUSE_RIGHT_BUTTON          = 2;
-const int32_t LE_MOUSE_MIDDLE_BUTTON         = 3;
-const int32_t LE_ORDER_ATTACK                = 28;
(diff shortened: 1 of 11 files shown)


code/scripting/api/objs/enums.h
```diff
@@ -7,148 +7,182 @@
...
-const int32_t LE_EXITED                      = 117;
-const int32_t LE_DC_IS_HULL                  = (1<<0); // the following definitions are bitfields independent of the usual indexing
-const int32_t LE_DC_VAPORIZE                 = (1<<1);
-const int32_t LE_DC_SET_VELOCITY             = (1<<2);
-const int32_t LE_DC_FIRE_HOOK                = (1<<3);
-
-const int ENUM_NEXT_INDEX = 118; // <<<<<<<<<<<<<<<<<<<<<<
-extern flag_def_list Enumerations[];
-extern size_t Num_enumerations;
+enum lua_enum : int32_t {
```
Goober5000, line 15: should this be `enum class`?

Goober5000 (commented):
Very clever. 👍  Looks mostly good as far as I can see; one question.

code/scripting/api/objs/enums.h
BMagnu: Yes and no.
While scoping it in an ``enum class`` certainly makes sense and would clean up the global namespace, it also comes with some downsides.
Notably, since it wasn't scoped before, scoping it now would mean that every usage of lua enum values in every file needs to be touched and adjusted for the new scope. With every enum being used once to thrice, that's a lot of additional line bloat in this PR.
But it certainly would be possible, if really wanted

Goober5000: Ah, good point.  I'm fine with leaving it as-is.

Goober5000 (approved):

PR #4869, opened 2022-11-07
Fix num comparison for FS2 medals

MjnMixael:


JohnAFernandez (approved):

MjnMixael:
Hold up. This causes a crash launching FSO on retail. I think I see why.

PR #4875, opened 2022-11-09
dynamic events and goals

Goober5000:
Make events and goals dynamic in missions and campaigns.  When parsing, initialize all fields to their defaults.

~~In draft status pending tests.~~  Tested by @wookieejedi and myself; works well so far

### DIFF
--- a/code/gamesnd/eventmusic.cpp
+++ b/code/gamesnd/eventmusic.cpp
@@ -1032,7 +1032,7 @@ int event_music_primary_goals_met()
 
 		// If the mission goals aren't met (or there are no goals), or if victory 2 music has already played, then go
 		// to the next default track
-		if ( !mission_goals_met() || Victory2_music_played || (Num_goals == 0)) {
+		if ( !mission_goals_met() || Victory2_music_played || Mission_goals.empty()) {
 			Patterns[next_pattern].next_pattern = Patterns[next_pattern].default_next_pattern;
 		} else {
 			Victory2_music_played = true;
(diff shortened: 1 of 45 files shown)


notimaginative:
LGTM, but I mainly focused on the multi stuff. I didn't notice issues anywhere else though.

One request... multi still has a limit of goals being less than `UINT8_MAX`, and the static assert was removed, so adding a regular Assertion() after goals are parsed in multi mode would be nice. 255 goals is a lot, but some folks around here take that kind of thing as a challenge.

Goober5000:
> LGTM, but I mainly focused on the multi stuff. I didn't notice issues anywhere else though.
> 
> One request... multi still has a limit of goals being less than `UINT8_MAX`, and the static assert was removed, so adding a regular Assertion() after goals are parsed in multi mode would be nice. 255 goals is a lot, but some folks around here take that kind of thing as a challenge.

There's already a new check when goals are parsed.  Did you catch that, or is that insufficient?

```
void parse_goals(mission *pm)
{
	required_string("#Goals");

	while (required_string_either("#Waypoints", "$Type:")){
		parse_goal(pm);
	}

	if ((pm->game_type & MISSION_TYPE_MULTI) && Mission_goals.size() > UINT8_MAX)
		throw parse::ParseException("Number of goals is too high and breaks multi!");
}
```

notimaginative:
Ah ha, yep, missed that. You covered the bases with everything else I was concerned with so I figured it had to be in there somewhere. I guess I was expecting it to be done a different way and my brain just didn't pick up on it. But that works fine. 👍

PR #4876, opened 2022-11-10
replace keybinds in debriefing

### LINKED ISSUE #3241: FR: Enable $default keys$ in debriefings, recommendations

MjnMixael:
Quick addition to replace keybinds in debriefings at the same time that the code already replaces variables and containers. Fixes #3241

### DIFF
--- a/code/missionui/missiondebrief.cpp
+++ b/code/missionui/missiondebrief.cpp
@@ -25,6 +25,7 @@
 #include "mission/missionbriefcommon.h"
 #include "mission/missioncampaign.h"
 #include "mission/missiongoals.h"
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
Very useful, clear, and simple change.

PR #4878, opened 2022-11-10
ship flag to exempt it from nebula targeting restrictions

### LINKED ISSUE #4195: FR: Exempt a ship from nebula targeting restrictions
(Minor spoilers for _Between the Ashes: The Growing Silence_.)

BTA has a use case for making a ship (specifically, a Pharos buoy) exempt from the targeting distance limit in a nebula. A new SEXP operator or ship flag would be ideal, whichever is easier to implement.

MjnMixael:
Adds a ship flag to exempt a ship from nebular targeting range restrictions and make it always targetable. Fixes #4195

### DIFF
--- a/code/ship/awacs.cpp
+++ b/code/ship/awacs.cpp
@@ -232,6 +232,15 @@ float awacs_get_level(object *target, ship *viewer, int use_awacs)
 			return FULLY_TARGETABLE;
 	}
 
+	// check for an exempt ship. Exempt ships are _always_ visible
+	if (target->type == OBJ_SHIP)
+	{
+		Assert( shipp != NULL );
+		int target_is_exempt = (shipp->flags[Ship::Ship_Flags::No_nebula_targeting_limits]);
+		if (target_is_exempt)
+			return FULLY_TARGETABLE;
+	}
+
 	// check for a tagged ship. TAG'd ships are _always_ visible
 	if (target->type == OBJ_SHIP)
 	{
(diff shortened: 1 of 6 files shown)


MjnMixael:
After some internal BtA discussion we think it's better to make this not restricted to nebular AWACS, but also max-targeting-range as well. A ship with this flag will be always targetable in all circumstances.

code/ship/awacs.cpp
```diff
@@ -221,6 +216,19 @@ float awacs_get_level(object *target, ship *viewer, int use_awacs)
 
 		check_huge_ship = (sip->is_huge_ship());
 	}
+
+	// check for an exempt ship. Exempt ships are _always_ visible
+	if (target->type == OBJ_SHIP) {
+		Assert(shipp != NULL);
```
JohnAFernandez, line 222: Since you're going to add an assert check against NULL here, modernize the function with `nullptr`.  So initialize the variable with `nullptr` intead of `NULL` and replace `NULL` in all the places it is checked.  

Basically, it's not worth it to change all the `NULL`s in the codebase to `nullptr` at once since the benefit is small, but we shouldn't add new `NULL`s  in PRs and should update old code that is in the same function.  Might be a good time to update `ship_info *sip = NULL;`, too.

qtfred/src/mission/missionsave.cpp
```diff
@@ -3294,6 +3294,9 @@ int CFred_mission_save::save_objects()
 			if (shipp->flags[Ship::Ship_Flags::Cannot_perform_scan]) {
 				fout(" \"cannot-perform-scan\"");
 			}
+			if (shipp->flags[Ship::Ship_Flags::No_targeting_limits]) {
+				fout(" \"cannot-perform-scan\"");
```
JohnAFernandez, line 3298: Copy paste mistake?

JohnAFernandez (changes requested):

code/ship/awacs.cpp
MjnMixael: Done

qtfred/src/mission/missionsave.cpp
MjnMixael: yup

JohnAFernandez (approved):

PR #4879, opened 2022-11-10
Fix poof name parsing and fix lab not knowing about the new parsing

MjnMixael:
#4856 broke the new nebula poof parsing by zero'ing the list and then immediately setting all the bits before the parse happened. This resulted in Neb2_poof_flags == -1 before it even starts trying to set bits from the list of poof names. This PR removes that and moves the bit setting to it's own method and then calls that method during mission parsing. Also makes sure the Lab is aware of the new poof parsing option.

### DIFF
--- a/code/nebula/neb.cpp
+++ b/code/nebula/neb.cpp
@@ -332,6 +332,19 @@ void neb2_init()
 	parse_modular_table("*-neb.tbm", parse_nebula_table);
 }
 
+// set the bits for poofs from a list of poof names
+void neb2_set_poof_bits(SCP_vector<SCP_string> list)
+{
+	Neb2_poof_flags = 0; //Make absolutely sure flags are zero'd before we start adding to it-Mjn
+	for (const SCP_string& thisPoof : list) {
+		for (int i = 0; i < (int)Poof_info.size(); i++) {
+			if (Poof_info[i].name == thisPoof) {
+				Neb2_poof_flags |= (1 << i);
+			}
+		}
+	}
+}
+
 bool poof_is_used(size_t idx) {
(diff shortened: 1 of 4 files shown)


code/nebula/neb.cpp
```diff
@@ -332,6 +332,19 @@ void neb2_init()
 	parse_modular_table("*-neb.tbm", parse_nebula_table);
 }
 
+// set the bits for poofs from a list of poof names
+void neb2_set_poof_bits(SCP_vector<SCP_string> list)
```
Goober5000, line 336: This should be a reference (and probably a const reference) since there's no need to copy the entire list:
```suggestion
void neb2_set_poof_bits(const SCP_vector<SCP_string>& list)
```

code/nebula/neb.cpp
```diff
@@ -332,6 +332,19 @@ void neb2_init()
...
+{
+	Neb2_poof_flags = 0; //Make absolutely sure flags are zero'd before we start adding to it-Mjn
+	for (const SCP_string& thisPoof : list) {
+		for (int i = 0; i < (int)Poof_info.size(); i++) {
+			if (Poof_info[i].name == thisPoof) {
```
Goober5000, line 341: I'd recommend a case-insensitive comparison, since most mission parsing checks are case-insensitive.
```suggestion
			if (SCP_string_lcase_equal_to()(Poof_info[i].name, thisPoof)) {
```

Goober5000 (changes requested):
Makes sense.  Two comments.

code/nebula/neb.cpp
MjnMixael: Done

Goober5000 (approved):

PR #4880, opened 2022-11-10
Modular bolts

MjnMixael:
Modular lightning table with better errors (proper warnings and errors instead of asserts) and a little general cleanup to use fewer casts; make the code more readable.

### DIFF
--- a/code/nebula/neblightning.cpp
+++ b/code/nebula/neblightning.cpp
@@ -282,7 +282,7 @@ void parse_lightning_table(const char* filename)
 				}
 			}
 
-			if (bolt_p->texture < 0)
+			if (!Fred_running && (bolt_p->texture < 0))
 				error_display(1, "Bolt %s has no texture defined.", bolt_p->name);
 
 			// glow
--- a/code/nebula/neblightning.cpp
+++ b/code/nebula/neblightning.cpp
@@ -427,7 +427,7 @@ static void verify_storm_bolts()
 		Storm_types[i].num_bolt_types = (Storm_types[i].num_bolt_types - error_count);
 
 		//Semi duplicate error message, but relevant in case all bolts were removed
-		if (Storm_types[i].num_bolt_types = 0)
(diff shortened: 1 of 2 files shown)


code/nebula/neblightning.cpp
```diff
@@ -152,122 +152,245 @@ DCF(lightning_intensity, "Sets lightning intensity between 0.0 and 1.0 (Default
...
-				new_bolt_type.glow = bm_load(name);
+			if (optional_string("+b_glow:")) {
+				stuff_string(name, F_NAME, sizeof(name));
+				if (!Fred_running) {
+					bolt_p->glow = bm_load(name);
+				}
 			}
 
+			if (bolt_p->glow < 0)
+				error_display(1, "Bolt %s has no glow texture defined.", bolt_p->name);
```
Baezon, line 297: No textures makes sense, but is no glow necessarily a problem?

Baezon (changes requested):

code/nebula/neblightning.cpp
MjnMixael: You know.. now that you ask, I'm not sure the glow texture is even used anywhere. At least I can't seem to find it. I added this check here because it was required by the original parsing code.

MjnMixael: OK, I did a test with a build that completely removed `int glow` from the bolt struct and threw away the data if parsed. Things seemed to work without issue or error in the several retail nebula missions I tested. Reading the code it looks like the glow is actually determined by the `+b_add` value. So it would probably make sense to deprecate the glow texture completely, but I'd feel more comfortable doing that if someone could double check my findings.

Baezon: It's fine, no need to go far (at least on this PR), I'd just remove the warning for no glow.

MjnMixael: Fair enough. Done.

Baezon (approved):

PR #4881, opened 2022-11-10
fix swapping of things in FRED

Goober5000:
This properly updates event and goal signatures when events are moved around.  Fixes Mantis bugs 1005, 1961, and 2998.


wookieejedi (approved):
Great to have those 3 mantis bugs finally fixed

PR #4883, opened 2022-11-11
bump MAX_AI_PROFILES

Goober5000:
Previous limit lasted from 2005-11-21 until now (from commit b33caf6f45).  We'll see how long this one lasts.


EatThePath (approved):
Really, having this as a fixed length array seems a little mad at this stage, but I'm not in a hurry to take on another cleanup right now so might as well bump

PR #4884, opened 2022-11-11
add set-order-allowed-for-target sexp

### LINKED ISSUE #4196: FR: Blacklisting problematic orders
Sometimes, FSO will permit an order to be issued that directly contradicts the players' orders, or just doesn't make sense in the fiction of the mission (calling an attack on Arjuna 1 in "Enter the Dragon" comes to mind, or Beta wing in BTA's "Playing Pirate"). An SEXP operator to blacklist these orders would be useful.

Proposed API:
```
( set-order-allowed-for-target
   ( false )
   "Arjuna 1"
   "Destroy My Target"
   "Additional orders if desired"
)
```

Ideally, a wing or team could be passed instead of a ship. For example, a FREDer might disable "Capture My Target" orders for "<any hostile>", then enable them for a few specific ships.

MjnMixael:
Adds feature to check disallow orders based on the target of the order. For example, if a particular hostile should not be destroyed you can prevent the player for issuing an attack order on just that one ship. Fixes #4196

Also adds a short bit of error checking to set-player-orders while I'm in there to prevent a similar assertion if a mission is run with invalid orders in the sexp.


code/parse/sexp.cpp
```diff
@@ -12403,6 +12410,50 @@ void sexp_set_player_orders(int n)
 	}
 }
 
+void sexp_set_order_allowed_target(int n)
```
JohnAFernandez, line 12413: I think you should add the for here.  The meaning is a little too ambiguous here without, imho.

code/parse/sexp.cpp
```diff
@@ -12403,6 +12410,50 @@ void sexp_set_player_orders(int n)
...
+{
+	bool allow_order;
+	std::set<size_t> orders;
+
+	auto ship_entry = eval_ship(n);
+	if (!ship_entry || !ship_entry->shipp) {
```
JohnAFernandez, line 12419: Style: there should be a new line before and after if statements to make them distinct from surrounding code.

code/parse/sexp.cpp
```diff
@@ -12403,6 +12410,50 @@ void sexp_set_player_orders(int n)
...
+	allow_order = is_sexp_true(n);
+	n = CDR(n);
+	do {
+		for( size_t order : default_orders){
+			//Once we exceed the number of valid orders, break and warn
+			if (order >= Player_orders.size()) {
+				Warning(LOCATION, "Invalid order name %s found in sexp!", CTEXT(n));
+				break;
+			}
+			// OPF_AI_ORDER returns hud_name, so we must compare to that instead of parse_name
```
JohnAFernandez, line 12435: For example, a new line before this comment would be good.

code/parse/sexp.cpp
```diff
@@ -12403,6 +12410,50 @@ void sexp_set_player_orders(int n)
...
+			if (!stricmp(CTEXT(n), Player_orders[order].hud_name.c_str())) {
+				orders.insert(order);
+				break;
+			}
```
JohnAFernandez, line 12439: A new line after this curly brace is not necessary, as there is another curly brace right after it, so there is no need to further highlight the if statement.

code/parse/sexp.cpp
```diff
@@ -12403,6 +12410,50 @@ void sexp_set_player_orders(int n)
...
+		
+	// set or unset the orders
+	if (allow_order) {
+		shipp->orders_allowed_against.insert(orders.begin(), orders.end());
+	}
+	else {
+		std::set<size_t> diff;
+		std::set_difference(shipp->orders_allowed_against.begin(), shipp->orders_allowed_against.end(), orders.begin(), orders.end(),
+							std::inserter(diff, diff.end()));
+		shipp->orders_allowed_against = diff;
```
JohnAFernandez, line 12453: You're a pretty thorough tester, but I'm still going to ask if this logic branch was tested.  Not that I think you made any mistakes, but I've never used `set_difference` myself.  I probably would have wrote manual code (not that it's better, I just didn't know it existed).

JohnAFernandez (changes requested):
Main things I see are style related.

code/parse/sexp.cpp
```diff
@@ -12403,6 +12410,50 @@ void sexp_set_player_orders(int n)
...
+
+	auto ship_entry = eval_ship(n);
+	if (!ship_entry || !ship_entry->shipp) {
+		return;
+	}
+	auto shipp = ship_entry->shipp;
+
+	const std::set<size_t>& default_orders = ship_set_default_orders_against();
+	n = CDR(n);
+	allow_order = is_sexp_true(n);
```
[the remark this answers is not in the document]
MjnMixael: Done

code/parse/sexp.cpp
MjnMixael: I'd bet money that I thought about typing the "for" and it just never made it to the keyboard... Fixed.

code/parse/sexp.cpp
MjnMixael: All above feedback should be addressed now.

code/parse/sexp.cpp
MjnMixael: Yeah, I tested many variations of adding one or multiple orders and then removing one or multiple orders. It all seemed to work as expected.

Now, I don't pretend to fully comprehend Sets (seems a lot like a bitfield but for an object array). This was the first time I dealt with one, but most order lists in the game are done with Sets so I matched that. In fact almost this entire method was copied from the one above `sexp_set_player_orders`, including `set_difference`. I'm about as confident in the code as a novice coder like me can be.

code/parse/sexp.cpp
```diff
@@ -12411,6 +12418,52 @@ void sexp_set_player_orders(int n)
...
+	if (!ship_entry || !ship_entry->shipp) {
+		return;
+	}
+
+	auto shipp = ship_entry->shipp;
+
+	const std::set<size_t>& default_orders = ship_set_default_orders_against();
+	n = CDR(n);
+	bool allow_order;
+	allow_order = is_sexp_true(n);
```
JohnAFernandez, line 12435: You can do `bool allow_order = is_sexp_true(n)`

MjnMixael: Done

JohnAFernandez (approved):

PR #4885, opened 2022-11-11
fix campaign loading in FRED

Goober5000:
Load campaigns using the full path provided by the user, not just from the configured mod paths.

Also tweaks initialization based on the campaign having already been cleared in `mission_campaign_clear()`.

### DIFF
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -1713,6 +1713,41 @@ int cf_get_file_list(SCP_vector<SCP_string>& list, int pathtype, const char* _fi
 	}
 
 	SCP_string filespec;
+	SCP_vector<_file_list_t> files;
+
+	// do we have a full path already?
+	if (is_absolute_path(_filter)) {
+		// check for full path and only grab matching files from it if specified
+		auto last_separator = strrchr(_filter, DIR_SEPARATOR_CHAR);
+
+		Assertion(Get_file_list_filter == nullptr, "File list filter not supported with absolute paths!");
+
+		const char *fltr = last_separator + 1;
+		filespec = SCP_string(_filter, strlen(_filter) - strlen(last_separator));
+
+		cf_get_list_of_files(filespec, files, fltr);
+
+		for (auto &file : files) {
+			SCP_string::size_type pos = file.name.find_last_of('.');
+
+			if (pos != SCP_string::npos) {
+				list.push_back(file.name.substr(0, pos));
+			} else {
+				list.push_back(file.name);
+			}
+
+			if (info) {
+				tinfo.write_time = file.m_time;
+				info->push_back(tinfo);
+			}
+		}
+
+		if (sort != CF_SORT_NONE)	{
+			cf_sort_filenames(list, sort, info);
+		}
+
+		return (int)list.size();
+	}
 
 	bool check_duplicates = !list.empty();
 
@@ -1749,7 +1784,6 @@ int cf_get_file_list(SCP_vector<SCP_string>& list, int pathtype, const char* _fi
 	}
 
 	SCP_string fullname;
-	SCP_vector<_file_list_t> files;
 
 	cf_get_list_of_files(filespec, files, filter.c_str());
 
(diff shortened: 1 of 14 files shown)


fred2/campaigntreewnd.cpp
```diff
@@ -133,8 +132,9 @@ void campaign_tree_wnd::OnCpgnFileOpen()
 		if (!strlen(name))
 			return;
 
-		string_copy(Campaign.filename, name, MAX_FILENAME_LEN - 1);
-		Campaign_tree_formp->load_campaign();
+		full_path = dlg.GetPathName();
```
notimaginative, line 135: How long is that pointer valid? I don't really see anything in the documentation, but this might rely on it being valid and unchanged longer than is safe to do so.

Goober5000: Good point.  I changed things up a bit.

fred2/campaignfilelistbox.cpp
```diff
@@ -46,38 +46,57 @@ END_MESSAGE_MAP()
...
+	}
+	else
+	{
+		SCP_vector<file_list_full_info> files;
+		cf_get_list_of_files((LPCTSTR)path, files, "*.fs2");
+
+		for (const auto &file: files)
+			mission_filenames.push_back(file.name);
+	}
+
```
notimaginative, line 75: The changes to `cf_get_file_list` I made mean you can just do something like:

```
SCP_string filter;

if (!path.IsEmpty()) {
    filter = (LPCTSTR)path;
    filter += DIR_SEPARATOR_CHAR;
}

filter += "*.fs2";

cf_get_file_list(mission_filenames, CF_TYPE_MISSIONS, filter.c_str());
```

and it should work in either case with less code. Basically if the `filter` argument is an absolute path then it will split off the actual filter and only return matching files in that exact path. Then you would haven't to make that internal `list_of_files` global.

Goober5000: Gotcha

code/cfile/cfilesystem.cpp
```diff
@@ -1530,6 +1530,41 @@ int cf_get_file_list(SCP_vector<SCP_string>& list, int pathtype, const char* fil
...
+
+	// check for full path and only grab matching files from it if specified
+	auto last_separator = strrchr(filter, DIR_SEPARATOR_CHAR);
+
+	// do we have a full path already?
+	if (last_separator) {
+		Assertion(Get_file_list_filter == nullptr, "File list filter not supported with absolute paths!");
+
+		const char *fltr = last_separator + 1;
+		filespec = SCP_string(filter, strlen(filter) - strlen(last_separator));
```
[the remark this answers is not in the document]
Goober5000: Done

fred2/campaigneditordlg.cpp
```diff
@@ -139,11 +141,17 @@ void campaign_editor::OnLoad()
...
+		full_mission_path.Append(Campaign.missions[Cur_campaign_mission].name);
 
-	if (res.found) {
-		FREDDoc_ptr->SetPathName(res.full_name.c_str());
+		auto res = cf_find_file_location((LPCTSTR)full_mission_path, CF_TYPE_MISSIONS, false);
+		if (res.found) {
+			FREDDoc_ptr->SetPathName(res.full_name.c_str());
+		}
 	}
+
```
notimaginative, line 154: Is this correct? I looks like it will only look for files if the campaign path is set. Wasn't there a fallback here previously for when it's not set? I'm guessing that the addition of `GetPathWithoutFile()` made that obsolete, and the IsEmpty check was meant to be removed?

Goober5000: Right, previously it would load from the regular mod data sources.  However, for whatever reason, retail missions stored in root_fs2.vp would not load properly.  It found them but couldn't read them.  I decided to restrict loading to only those cases where the mission was found on the file system, since that's really the only way FRED is used anyway.

notimaginative: Ah, okay then.

notimaginative (approved):

PR #4891, opened 2022-11-13
add cannot-perform-scan flag

Goober5000:
Prevents the player from performing a cargo or subsystem scan.  Can be toggled using alter-ship-flag.

Also adds some missing logic for Fail_sound_locked_primary, Fail_sound_locked_secondary, and Aspect_immune.


wookieejedi (approved):
Looks good, also good to have the fixes with the existing 3 newer flags.

PR #4898, opened 2022-11-15
hud-force-sensor-static sexp

MjnMixael:
Sexp to toggle sensor status as if sensors were damaged below 30%.


code/parse/sexp.cpp
```diff
@@ -37520,6 +37537,12 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"\tRest:\tHUD Gauge Group name\r\n"
 	},
 
+	{OP_HUD_FORCE_SENSOR_STATIC, "hud-force-sensor-static\r\n"
+		"\tActivates or deactivates hud static as if sensors are damaged."
+		"Takes 1 Arguments...\r\n"
+		"\t1:\tBoolean, whether or not to display this gauge\r\n"
```
Goober5000, line 37543: "display this gauge" isn't really accurate as an argument description for this sexp.  Also "argument" should be singular.

Goober5000 (changes requested):
One comment on the sexp description.  Also, this new variable should be initialized in `hud_init_target_static` in case the player restarts the mission while the sexp is active.

code/ship/subsysdamage.h
```diff
@@ -40,6 +40,7 @@
 #define SENSOR_STR_RADAR_NO_EFFECTS		0.4f	// % strength of sensors at which no negative effects on radar
 #define MIN_SENSOR_STR_TO_RADAR			0.1f	// % strength of sensors at which radar ceases to function
 
+extern bool Target_static_forced;				// For forcing damaged radar effect by sexp - Mjn
```
Baezon, line 43: Minor, but why is it called `Target_static_forced`? This doesn't actually have anything to do with the target, right? Shouldn't it be like `Radar_static_force` or something?

code/parse/sexp.cpp
MjnMixael: Done

Goober5000 (approved):

code/ship/ship.cpp
```diff
@@ -203,6 +203,8 @@ int	Num_ship_subobj_types;
 int	Num_ship_subobjects;
 int	Player_ship_class;	// needs to be player specific, move to player structure	
 
+bool Target_static_forced = false; //forced static effect sexp initialization
```
Goober5000, line 206: One more rename

MjnMixael: done

PR #4900, opened 2022-11-16
hook up cone light profile values

EatThePath:
I really would like this to be in 22.4 but sneaky rakes keep getting in the way of me properly testing it...


wookieejedi (approved):
Tested and works as expected

PR #4901, opened 2022-11-16
Fix a warning

JohnAFernandez:
We shouldn't be checking that the `cci_h` is valid by seeing if it is greater than -1 because by that point it never will be.  Instead, we should be checking that the incoming id is valid and setting to the invalid state if it is not.

### DIFF
--- a/code/scripting/api/objs/control_binding.cpp
+++ b/code/scripting/api/objs/control_binding.cpp
@@ -10,7 +10,7 @@ cci_h::cci_h() { idx = IoActionId::CCFG_MAX; }
 
 cci_h::cci_h(int n_id) { if (n_id < 0 || n_id > static_cast<int>(IoActionId::CCFG_MAX)) idx = IoActionId::CCFG_MAX; else idx = static_cast<IoActionId>(n_id); }
 
-bool cci_h::IsValid() { return idx < IoActionId::CCFG_MAX; }
+bool cci_h::IsValid() { return idx != IoActionId::CCFG_MAX; }
 
 IoActionId cci_h::Get() { return idx; }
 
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/control_binding.cpp
```diff
@@ -6,11 +6,11 @@
...
 namespace api {
 
-cci_h::cci_h() { idx = CCFG_MAX; }
+cci_h::cci_h() { idx = IoActionId::CCFG_MAX; }
 
-cci_h::cci_h(int n_id) { idx = static_cast<IoActionId>(n_id); }
+cci_h::cci_h(int n_id) { if (n_id < 0 || n_id > static_cast<int>(IoActionId::CCFG_MAX)) idx = IoActionId::CCFG_MAX; else idx = static_cast<IoActionId>(n_id); }
 
-bool cci_h::IsValid() { return (idx > -1 && idx < IoActionId::CCFG_MAX); }
+bool cci_h::IsValid() { return idx < IoActionId::CCFG_MAX; }
```
z64555, line 13: Nitpick.  Enumerations should not be treated as integers and instead be treated as unique identifiers or states, this means that the >, >=, <, and <= operators should not be used on enumerations.

This is often overlooked for various reasons, and we're not that uptight about this rule, either. As an example we exploit that enums *can* be treated as int's to check for sub-ranges within the enum (specifically the joystick axes).

I'll Allow It.

z64555 (approved):
LGTM

code/scripting/api/objs/control_binding.cpp
JohnAFernandez: I mean, it's never too late to start doing things the right way.  Shall I do `idx != IoActionId::CCFG_MAX;` instead?

PR #4902, opened 2022-11-16
Fix Flashlights

wookieejedi:
Area_normalisation term in deferred shader was not getting set properly for cone lights, which caused things like ship headlights to not work properly.

Tested and fixes issue as expected. Thanks to Qaz and EatThePath for pointing out how to fix it!


EatThePath (approved):
looks good!

PR #4904, opened 2022-11-16
Resolve SEXP Vs. Edit Ship discrepancy in maximum character limit for callsigns

LightblueMobius:
It turns out that "Edit Ship" in FRED allows the addition of callsigns with a 18-character limit, while the **ship-change-callsign** SEXP boosts that limit to 31. This means that long callsigns can only be added via SEXP, and it's not possible to have them by default.

EDIT: _the Ship Editor limit also seems to be dependent on UI, because MjnMixael just got a 24-character limit in FRED._

Unless resolving the conflict poses certain coding issues, both limits should be the same because the current discrepancy may lead FREDders to confusion.

If it can't be done for any reason I'm not aware of, at least it should be specified that the two limits are different. One possibility is modifying the description of how the SEXP works:

```
ship-change-callsign
     Changes the callsign of a ship.  Takes 2 or more arguments...
     1: The callsign to display or empty to remove
     Rest: The ships to display the new callsign
```

The description may be changed into something like this:

```
ship-change-callsign
     Changes the callsign of a ship. Please note that this SEXPs boosts the character limit of callsigns to 31, while Edit Ship has that limit set to 18 characters. Takes 2 or more arguments...
     1: The callsign to display or empty to remove
     Rest: The ships to display the new callsign
```


wookieejedi:
Always great to see a first time PR. ATM though this PR needs to be cleaned up and based off of current FSO master, so that we can see that changed files clearly without the hundreds of lines of other code changes git is showing :)

wookieejedi:
#4905 has fix

PR #4911, opened 2022-11-19
Setting up Vampirism again... again

Strygon:
I've returned to this project after a year of not working on it. Got most of the changes from last time added, along with warnings for missing flags or bad values. The core principle is still the same:

-Vampirism flag enables the ability to regain health by dealing hull damage to a ship.
-"$Vampiric Healing Factor:" is the multiplier by which the final hull damage dealt is multiplied by. This value now cannot go below 0.

This now DOES work with shockwaves and beams and primaries and secondaries.

### DIFF
--- a/code/ship/shiphit.cpp
+++ b/code/ship/shiphit.cpp
@@ -2459,13 +2459,13 @@ static void ship_do_damage(object *ship_objp, object *other_obj, vec3d *hitpos, 
 				object* parent = &Objects[other_obj->parent];
 
 				if ((parent->type == OBJ_SHIP) && (parent->signature == other_obj->parent_sig)) {
-					ship* shipparent = &Ships[parent->instance];
+					ship* shipp_parent = &Ships[parent->instance];
 
 					if (!parent->flags[Object::Object_Flags::Should_be_dead]) {
 						parent->hull_strength += damage * wip->vamp_regen;
 
-						if (parent->hull_strength > shipparent->ship_max_hull_strength) {
-							parent->hull_strength = shipparent->ship_max_hull_strength;
+						if (parent->hull_strength > shipp_parent->ship_max_hull_strength) {
+							parent->hull_strength = shipp_parent->ship_max_hull_strength;
 						}
 					}
 				}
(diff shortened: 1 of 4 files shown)


code/ship/shiphit.cpp
```diff
@@ -2448,14 +2448,33 @@ static void ship_do_damage(object *ship_objp, object *other_obj, vec3d *hitpos,
 				apply_diff_scale = false;
 			}
 		}
+
+		// if weapon is vampiric, slap healing on shooter instead of target
+		weapon_info_index = shiphit_get_damage_weapon(other_obj);
+		if (weapon_info_index >= 0) {
```
JohnAFernandez, line 2454: So, one style thing.  You'll want to put a new line before an if statement to make it distinct from surrounding code.  Also after if there's code after it, too, but this one is just curly braces after

Baezon (approved):
Approving for the changes I suggested in the original PR.

code/ship/shiphit.cpp
```diff
@@ -2448,14 +2448,38 @@ static void ship_do_damage(object *ship_objp, object *other_obj, vec3d *hitpos,
...
+		weapon_info_index = shiphit_get_damage_weapon(other_obj);
+
+		if (weapon_info_index >= 0) {
+			weapon_info* wip = &Weapon_info[weapon_info_index];
+
+			if ((wip->wi_flags[Weapon::Info_Flags::Vampiric]) && (other_obj->parent > 0)) {
+				object* parent = &Objects[other_obj->parent];
+
+				if ((parent->type == OBJ_SHIP) && (parent->signature == other_obj->parent_sig)) {
+					ship* shipparent = &Ships[parent->instance];
```
JohnAFernandez, line 2462: So two more style things you should be aware of.  There are lots of different cases that we use in FSO for different variable scopes, and some files even differ on that.  For locals, though, we use snake_case, and if you're in a file that uses it for locals, you should follow that convention.  Again some files might do other things.  When you're in those files, just follow what's done there.

The other style thing is that if a variable is a pointer, you should make it part of the variable name.  For instance, you'll see `objp` all over the place in FSO because it means object pointer.  `shipp` means ship pointer.  These pointers are pretty short lived, though, so it's not that big a deal.

I definitely think you should switch the case of shipparent, but I think that's the only thing I'm going to require.

JohnAFernandez (changes requested):

JohnAFernandez (approved):

PR #4913, opened 2022-11-19
Adjust ssssssssssi in one more place

JohnAFernandez:
This fixes both pre-mission and in-mission crashes regarding special_ship_ship_info_index

Also adds an adjustment to sssi when a ship in a wing dies.


wookieejedi (approved):
Thanks for the quick fix, also noting here that BtA and Warmachine tests confirmed these updates fixed things.

PR #4914, opened 2022-11-19
add text-to-speech support for macOS

notimaginative:


notimaginative:
First attempt at something like this so hopefully it works out. Tested on a recent mac, but nothing else. Should work for any mac running version 10.3+.

code/sound/speech.mm
```diff
@@ -0,0 +1,171 @@
...
+	if (Speech_init) {
+		return true;
+	}
+
+	synth = [[NSSpeechSynthesizer alloc] init];
+
+	Speech_init = true;
+
+	return true;
+}
```
JohnAFernandez, line 25: Looks like this function can only return true.  Is the return value necessary?

CMakeLists.txt
```diff
@@ -83,9 +89,11 @@ IF(RESET_INSTALL_PREFIX)
 	ENDIF(NOT $ENV{FS2PATH} STREQUAL "")
 ENDIF(RESET_INSTALL_PREFIX)
 
-IF (WIN32)
-	OPTION(FSO_USE_SPEECH "Use Windows specific text-to-speach libraries" ON)
+IF(WIN32 OR APPLE)
+	OPTION(FSO_USE_SPEECH "Use text-to-speach libraries" ON)
```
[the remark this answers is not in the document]
notimaginative: Linux does have *some* support but I'm not exactly sure to what degree it works. Someone might want to go in and fix it at some point.

In this particular spot it's just setting the default to ON for windows and mac, where support is pretty much guaranteed.

code/sound/speech.mm
notimaginative: It has to match the windows implementation, which returns a lot of bool's for some reason.

JohnAFernandez: As suspected.

JohnAFernandez (approved):

Issue #3849, opened 2021-12-23
FRED crash/assertion failure with rename variable + Cancel from Events Editor


jg18:
Repro instructions:
1. Download [ModifyVariableTest.fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/7770848/ModifyVariableTest.fs2.txt) and remove the `.txt` extension.
2. Open the mission in a FastDebug FRED build and open FRED's Events Editor.
3. Open the Modify Variable dialog and rename `MyVariable` to, say, `MyNewVariable`, then press OK to close the Modify Variable dialog.
4. Press Cancel in the Events Editor and press No when asked whether to keep your changes.
5. Save the mission.

The following assertion failure occurs:

![resulting assertion failure](https://user-images.githubusercontent.com/6183333/147279218-911f0222-34ab-4046-aa74-1acab1017e6f.png).

The cause is that Canceling from the Events Editor results in `Sexp_nodes` not being updated with the new variable name, so `CTEXT()`'s attempt to look up the old variable name in `Sexp_variables` before saving the mission fails.

z64555:
According to @jg18 on discord "[3849 is] a rare edge case that I personally do not feel is worth investing time/energy (a perpetually scarce resource) into right now, but if someone else does, more power to them"

Punting to 22.4.

Goober5000:
Fixed in 4916

PR #4916, opened 2022-11-20
guard against unknown variables

Goober5000:
If a variable can't be found, fail gracefully instead of Asserting.  These changes will allow the mission to be saved and/or loaded, and will also allow the user to open the Event Editor and replace the nonexistent variable.  This fixes the crash/assertion failure in #3849, as well as the case where the variable is edited in the mission file.


wookieejedi (approved):
Looks good, and certainly a useful change that makes things less annoying tedious when fixing names.

PR #4919, opened 2022-11-21
fix full ship list selection in FRED

### LINKED ISSUE #4918: ship list selection in FRED omits entries
#4833 introduced a bug where the last few ships loaded by FRED would not appear in the ship class dropdown list.  The number of ships not appearing is equal to the number of ships omitted from the list via the `no_fred` flag.  In their place, the dropdown displays extra copies of Waypoint or similar.

Fixing this bug while still allowing #4833 to work as designed is proving rather tricky.  ~~I have a fix in progress but it's not ready yet.~~

Goober5000:
1. Use combo box item data to keep track of which ship class is represented by which item.  This allows the dropdown and the ship info list to work together without having 1:1 correspondence.
2. Rename cur_model_index to cur_ship_type_combo_index to be more accurate.
3. Remove GetCurSelNEW and SetCurSelNEW as they are now redundant.

Fixes #4918.


fred2/mainfrm.cpp
```diff
@@ -123,17 +123,21 @@ void CMainFrame::init_tools() {
             continue;
         } else {
             m_new_ship_type_combo_box.AddString(it->name);
+			m_new_ship_type_combo_box.SetItemData((int)ship_type_combo_box_size, std::distance(Ship_info.cbegin(), it));
```
JohnAFernandez, line 126: Looks like your tab key got stuck

Goober5000: Indeed.  Changed.

PR #4921, opened 2022-11-22
Add Option to allow weapons that hit player to place per-weapon sound

wookieejedi:
Weapons that lasers or missiles that hit the player also play a hard coded sound, which is not ideal when trying to give valuable audio feedback to player.

This PR allows modders to stop the behavior and have the weapon class impact sound specified in the weapons table play instead of the hard coded sounds.

Tested and works as expected.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -121,6 +121,7 @@ float Thruster_easing;
 bool Always_use_distant_firepoints;
 bool Discord_presence;
 bool hotkey_always_hide_ships;
+bool Use_weapon_class_sounds_for_hits_to_player;
 
 static auto DiscordOption = options::OptionBuilder<bool>("Other.Discord", "Discord Presence", "Toggle Discord Rich Presence")
 							 .category("Other")
@@ -765,6 +766,10 @@ void parse_mod_table(const char *filename)
 			}
 		}
 
+		if (optional_string("$Use per-weapon impact sounds for hits to player:")) {
+			stuff_boolean(&Use_weapon_class_sounds_for_hits_to_player);
+		}
+
 		optional_string("#FRED SETTINGS");
 
 		if (optional_string("$Disable Hard Coded Message Head Ani Files:")) {
@@ -1219,6 +1224,7 @@ void mod_table_reset()
 	Always_use_distant_firepoints = false;
 	Discord_presence = true;
 	hotkey_always_hide_ships = false;
+	Use_weapon_class_sounds_for_hits_to_player = false;
 }
 
 void mod_table_set_version_flags()
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -6792,7 +6792,7 @@ void weapon_hit_do_sound(object *hit_obj, weapon_info *wip, vec3d *hitpos, bool 
 		// play a shield hit if shields are above 10% max in this quadrant
 		if ( shield_str > 0.1f ) {
 			// Play a shield impact sound effect
-			if ( hit_obj == Player_obj ) {
+			if ( !(Use_weapon_class_sounds_for_hits_to_player) && (hit_obj == Player_obj)) {
 				snd_play_3d( gamesnd_get_game_sound(GameSounds::SHIELD_HIT_YOU), hitpos, &Eye_position );
 				// AL 12-15-97: Add missile impact sound even when shield is hit
 				if ( wip->subtype == WP_MISSILE ) {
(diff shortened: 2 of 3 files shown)


code/weapon/weapons.cpp
```diff
@@ -6792,7 +6792,7 @@ void weapon_hit_do_sound(object *hit_obj, weapon_info *wip, vec3d *hitpos, bool
 		// play a shield hit if shields are above 10% max in this quadrant
 		if ( shield_str > 0.1f ) {
 			// Play a shield impact sound effect
-			if ( hit_obj == Player_obj ) {
+			if ( !(Use_weapon_class_sounds_for_hits_to_player) && (hit_obj == Player_obj)) {
```
JohnAFernandez, line 6795: What will happen if the weapon impact sound is not specified?

wookieejedi: As far as I know that's just one of those fields that needs to be specified in the weapons table

JohnAFernandez: That makes sense.  Wouldn't want a weapon to silently impact.

JohnAFernandez (approved):

JohnAFernandez (approved):
Still looks good but maybe someone else wants to take a look too.

code/mod_table/mod_table.cpp
```diff
@@ -765,6 +766,10 @@ void parse_mod_table(const char *filename)
 			}
 		}
 
+		if (optional_string("$Use per-weapon impact sounds for hits to player:")) {
```
Goober5000, line 769: I'd say "$Use weapon class..." in the string too.  "Per-weapon" could imply it can be different for every shot.

code/weapon/weapons.cpp
Goober5000: Well nothing *needs* to be specified now that we have TBMs...

code/weapon/weapons.cpp
```diff
@@ -6792,28 +6798,32 @@ void weapon_hit_do_sound(object *hit_obj, weapon_info *wip, vec3d *hitpos, bool
...
```
Goober5000, line 6801: This condition should be changed to check that the shield hit sound is valid.  If it isn't, then the behavior should be as if the game_settings flag wasn't active.
```suggestion
			if ( !(Use_weapon_class_sounds_for_hits_to_player && wip->shield_impact_snd.isValid()) && (hit_obj == Player_obj)) {
```

code/weapon/weapons.cpp
```diff
@@ -6792,28 +6798,32 @@ void weapon_hit_do_sound(object *hit_obj, weapon_info *wip, vec3d *hitpos, bool
...
-				snd_play_3d( gamesnd_get_game_sound(GameSounds::SHIELD_HIT), hitpos, &Eye_position );
+				if (wip->shield_impact_snd.isValid()) {
+					snd_play_3d( gamesnd_get_game_sound(wip->shield_impact_snd), hitpos, &Eye_position );
+				} else {
+					snd_play_3d( gamesnd_get_game_sound(GameSounds::SHIELD_HIT), hitpos, &Eye_position );
+				}
 			}
 		} else {
 			// Play a hull impact sound effect
 			switch ( wip->subtype ) {
```
Goober5000, line 6816: Same here.  Unfortunately this section will have to be refactored a bit.  The `weapon_play_impact_sound` function will need an auxiliary function to just determine the sound to play, and then this sound will need to be checked for validity in the same parenthetical as `Use_weapon_class_sounds_for_hits_to_player`, just like above.

Goober5000 (changes requested):
In order for default behavior to be used if no sounds are specified, this will need a few modifications.

code/weapon/weapons.cpp
Goober5000: Never mind this suggestion

code/weapon/weapons.cpp
Goober5000: Never mind this suggestion

Baezon:
FWIW, I think that this game settings flag *should* play the default player hit sounds if there is no impact sound specific to that weapon.

As a separate feature, specific to each weapon they might have their own 'player hit' version, which the modder can make silent if they wish 'a silent impact' to override everything else. Not for this PR, but as justification for the mod-wide flag to prefer some kind of sound.

Goober5000:
Hmm, the question then becomes -- is "no weapon-specific impact sound" the same thing as "invalid weapon-specific impact sound"?  And are either of those the same thing as "valid, but silent, weapon-specific impact sound"?

wookieejedi:
Yeah, I had felt that if a modder wants the weapon to play the impact sound specified, but a weapon does not have one then that they would expect it not to play one. For example, I set sounds to -1 commonly for special weapons for that reason. And If sound is -1 then the `isValid` check will fail. 

Happy to keep discussing though, perhaps I'm missing something.

Baezon:
Yeah, after a second thought, I have come around that idea. If a weapon doesn't normally play an impact sound, I think it would be unusual to play sounds when only hitting the player, when in all other cases it wouldn't. If the modder intentionally wants that (or the reverse), explicit silent sounds, is unideal, but ultimately a fairly trivial cost,

wookieejedi:
Thanks, looks like we are all relatively on the same page now with leaving currently implemented code as is :)

Goober5000 (approved):
Sounds good

Issue #4512, opened 2022-07-26
EFX Reverb/sound environment should not apply  to music or dialogue - and some other things


MoerasGrizzly:
I've been diving into FSO's EFX and Sound Environment settings lately, and there's a lot to enjoy in this otherwise undocumented feature. However, there's several issues that I've run into:

1. The reverb effects apply to music and dialogue. Music should flat out not be affected by any EFX reverb solution for obvious reasons. Dialogue being affected by reverb should not be applied in situations where the dialogue is coming in through a radio (or headset), which in FS2's case is all the dialogue.
2. An exception to this could be if the sound enviroment settings can also be applied in menu screens, but right now they aren't.
3. Currently the FRED2 sound environment selector dropdown box doesn't display properly. I have to manually scroll up or down when selecting the box to see any entries. This is slightly different on my laptop, where it displays one setting.
4. Currently the FRED2 sound environment selector doesn't allow all the [default EFX presets](https://github.com/kcat/openal-soft/blob/master/include/AL/efx-presets.h) to be set. 
5. Currently whatever the default EFX setting is seems to vary from system to system. On my big PC with EFX enabled the game sounds very echo-y by default, whilst when playing on my laptop with EFX enabled, I get no reverb effects unless otherwise specified (which is, imo, the desirable effect).
6. EFX Sound Enviroment settings can currently be set on a per-mission basis, but I think the ability to set them on a per-ship basis could be very useful for some modders. The EFX system can be used to simulate the sound distortion that occurs in cockpits (see the RACING_INCAR_ presets) and such a system would allow something like FOTG or Diaspora to model sound distortion varying based on the size of the ship.

Right now the system is a bit overzealous and not well exposed to users., but I happen to think it has a lot of potential with a few tweaks.

MjnMixael:
Are effects still applied to music?

MoerasGrizzly:
> Are effects still applied to music?

Not as far as I can tell.

Damn, 4 years ago already xD

PR #4923, opened 2022-11-22
avoid playing voice messages with aux effects (EFX)

notimaginative:


notimaginative:
Addresses part of #4512.

This should probably be revisited in the future as part of making EFX more useful, should any brave soul desire to attempt such a thing.

MoerasGrizzly:
> 

I mean I'm already including EFX presets in BP

MoerasGrizzly:
Just tested the fix, it works as intended. Reverb gets applied to effects properly, but not to voices.

Ofc. I can't say anything about knock-on effects at this time.

PR #4925, opened 2022-11-22
pre parse icons to get an accurate count

### LINKED ISSUE #4894: Modular Icon tables isn't modular.
When a modular icon table is used in current nightly as of writing, FSO will complain that all species not defined in said modular icon table are not set. Even if they are in icons.tbl (like Vanilla's).

MjnMixael:
Fixes #4894 by pre-parsing the icons.tbl to get a count of how many times $Name: is found. Then during actual parse if icons.tbl we use that to limit how many species we're defining icons for instead of `unique_icons_species.size()` because some of those species may be defined later by a modular table. There may be a better way to do this, but it was the best idea I could think of this morning.

### DIFF
--- a/code/mission/missionbriefcommon.cpp
+++ b/code/mission/missionbriefcommon.cpp
@@ -345,6 +345,36 @@ void brief_icon_parse_cleanup() {
 	}
 }
 
+int Num_icons_in_table = 0;
+
+// This is explicitely used to count the number of icons listed in the tbl
+// so that we can correctly parse each icon into a species without any off-by-N errors.
+// This allows modular icons.tbls to build upon the retail icons.tbl.
+void brief_pre_parse_icons()
+{
+	try {
+		read_file_text("icons.tbl", CF_TYPE_TABLES);
+		reset_parse();
+
+		required_string("#Start");
+
+		//If we're in the new format, then nothing to do!
+		if (check_for_string("$Species:"))
+			return;
+
+		char junk[MAX_FILENAME_LEN];
+		while (optional_string("$Name:")) {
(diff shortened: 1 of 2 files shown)


code/mission/missionbriefcommon.cpp
```diff
@@ -417,14 +447,36 @@ void brief_parse_icon_tbl(const char* filename)
...
+
+			size_t Num_species_in_table = 0;
+			
+			//Calculate how many species we're going to try to parse icons for using the pre-parsed count
+			if (Num_icons_in_table % MIN_BRIEF_ICONS != 0)
+				Warning(LOCATION,
+					"An incorrect number of icons was found in icons.tbl. There should be %i icons per species listed!",
+					MIN_BRIEF_ICONS);
+
+			Num_species_in_table = Num_icons_in_table / MIN_BRIEF_ICONS;
```
Goober5000, line 459: Suggestion: use a temporary variable here, because at this intermediate step, `Num_species_in_table` does not actually represent the number of species in the table.

code/mission/missionbriefcommon.cpp
```diff
@@ -345,6 +345,36 @@ void brief_icon_parse_cleanup() {
 	}
 }
 
+static int Num_icons_in_table = 0;
```
Goober5000, line 348: Suggestion: make this `size_t` rather than `int`

Goober5000 (approved):
This is a pretty good solution.  Approving with two non-blocking suggestions.

code/mission/missionbriefcommon.cpp
MjnMixael: Alright. Done.

Goober5000 (approved):
👍

Issue #3643, opened 2021-09-06
FR: no_fred for weapons


EatThePath:
I have a lot, a LOT of weapons I will never ever want to assign to something in fred, generally submunittions but also upstream weapons I need to have in the files but only get in the way in fred list. A flag to hide them would be very, very helpful.

Goober5000:
I have a plan for how to do this, thanks to #4919.

Goober5000:
Implemented by 4927

PR #4927, opened 2022-11-23
add no_fred flag for weapons, and patch no_fred for ships

Goober5000:
The new `no_fred` flag prevents weapons from being listed in the dropdown lists in the weapon dialog.  Also this fixes `no_fred` ships to not appear either: previously they were hidden in the main window dropdown but still appeared in the ship editor.

In order to vastly simplify the dropdown list code, the weapon dialog was refactored to use arrays of variables rather than one hardcoded variable per bank.

Implements #3643.


wookieejedi (approved):
Testing and examined everything in detail and it all looks good and works as expected.

PR #4929, opened 2022-11-23
fix the rest of the autoscaled attributes

### LINKED ISSUE #4922: Following up 4516
Returning to this: The fix doesn't work fully. Although things like firerate appear to work correctly, stuff like afterburner usage is still affected by the scaling bug.
[ai.zip](https://github.com/scp-fs2open/fs2open.github.com/files/10050936/ai.zip)
[ai2.zip](https://github.com/scp-fs2open/fs2open.github.com/files/10050938/ai2.zip)

Steps to reproduce:
On FSBlue with a current nightly, try playing SM1-03 with AI1 and AI2, and notice how with the file included in AI2.zip, the AI reaches the Iceni much faster.

edit: on review of the submitted bugfix, it seems like several aspects that are affected by AI autoscaling have not been touched. I'm in particular worried about
$Afterburner Use Factor:
$Shockwave Evade Chances Per Second:
$Get Away Chance:
$Secondary Range Multiplier:

@EatThePath did a review of the AI's autoscaling a while back. I'll attach it here for your convenience:
[ai_autoscale_digging.txt](https://github.com/scp-fs2open/fs2open.github.com/files/10050994/ai_autoscale_digging.txt)

_Originally posted by @MoerasGrizzly in https://github.com/scp-fs2open/fs2open.github.com/issues/4516#issuecomment-1321269704_

Goober5000:
PR #4516 did not cover all of the autoscaled AI class attributes, since a few of them were labeled differently.  This adds checks to grab the correct autoscale index for the remaining cases:

```
$Afterburner Use Factor:
$Shockwave Evade Chances Per Second:
$Get Away Chance:
$Secondary Range Multiplier:
```

Fortunately, Sushi's investigation was pretty thorough.  Between these remaining cases and @EatThePath's autoscaling review, this PR should cover the complete set of autoscaling situations.

Fixes #4922.

### DIFF
--- a/code/ai/ai.h
+++ b/code/ai/ai.h
@@ -461,6 +461,7 @@ typedef struct ai_info {
 	ai_lua_parameters lua_ai_target;
 } ai_info;
 
+int ai_maybe_autoscale(int absolute_index = -1);
 int ai_get_autoscale_index(int absolute_index = -1);
 
 // Goober5000
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -558,6 +558,14 @@ int ai_get_autoscale_index(int absolute_index)
 	return index;
 }
 
+int ai_maybe_autoscale(int absolute_index)
+{
+	if (The_mission.ai_profile->flags[AI::Profile_Flags::Adjusted_AI_class_autoscale])
+		return ai_get_autoscale_index(absolute_index);
+	else
+		return absolute_index;
+}
+
 /**
  * Initialize an AI class's nonrequired values to defaults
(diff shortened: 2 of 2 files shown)


wookieejedi (approved):
Excellent to finally have this properly fixed, and looks good!

MoerasGrizzly:
Formidable! This seems to work pretty well now. 
My test case has been SM1-01, and changing the herc waves to go through all five AI classes. They seem to scale as expected. Comparisons are always tricky with AI ofc, but they're now a lot more eager to do stuff like use their afterburners (where they weren't for the longest time in FSblue until we found out about this bug).

Goober5000:
Glad to hear that!

PR #4932, opened 2022-11-25
make sexp_tree dragging more user-friendly

Goober5000:
FREDders expect events (and goals and campaign links) to land where you drag them.  So when dragging items earlier in the list, insert them before the item they are dropped on.


wookieejedi (approved):
Overall looks good. Seems like a useful QoL update, just wanted to check if you have asked any other FREDers about this?

Goober5000:
Yup, and it was received with enthusiasm 👍

wookieejedi:
Sounds great!

PR #4934, opened 2022-11-25
don't process turrets or subsystems when playing dead

### LINKED ISSUE #4933: Mjolnirs spin while deactivated in Scroll's 'Thor's Hammer'
As seen here, at 3:26: https://www.youtube.com/watch?v=eILRbC1uMFU

Goober5000:
In the submodel reorganization in commit 575c050da44c213c4ef127588880f24101ca480f, subobject movement was moved out of the AI pipeline, which inadvertently removed the play-dead check that prevents subsystem movement.  This broke a few missions such as Scroll's 'Thor's Hammer'.  This PR restores the check.

Fixes #4933.


wookieejedi (approved):
Good find and quick fix

PR #4936, opened 2022-11-26
Animate More Things

BMagnu:
This PR finally adds support for triggered animations on skyboxes and cockpits.
In turn, better support for scripting is added, allowing to acquire and hold handles to triggered animations for more than one frame, making handling of starting / stopping / setting speed for running animations much easier.


code/model/modelanimation.cpp
```diff
@@ -838,7 +839,22 @@ namespace animation {
 		return ret;
 	};
 
+	std::set<SCP_string> ModelAnimationSet::getRegisteredAnimNames() const {
+		std::set<SCP_string> ret;
```
Goober5000, line 843: Does it make sense to use SCP_set rather than std::set (here and elsewhere)?

code/starfield/starfield.cpp
```diff
@@ -2301,9 +2301,7 @@ void stars_set_background_model(const char *model_name, const char *texture_name
 	if (Nmodel_num >= 0) {
 		model_page_in_textures(Nmodel_num);
 
-		if (model_get(Nmodel_num)->flags & PM_FLAG_HAS_INTRINSIC_MOTION) {
-			Nmodel_instance_num = model_create_instance(-1, Nmodel_num);
-		}
+		Nmodel_instance_num = model_create_instance(-1, Nmodel_num);
```
Goober5000, line 2304: Elsewhere I've tried to avoid creating model instances unless they are needed.  Since I assume the majority of skyboxes will not need instances, this seems inefficient.  On the other hand, it's just one model per mission and `polymodel_instance` is pretty lightweight.

qtfred/src/mission/missionsave.cpp
```diff
@@ -2487,6 +2487,50 @@ int CFred_mission_save::save_mission_info()
 			fout_version("\n\n$Skybox Model: %s.pof", out_str);
 		}
 		fso_comment_pop();
+
+		const auto& anim_triggers = The_mission.skybox_model_animations.getRegisteredAnimNames();
+		const auto& moveable_triggers = The_mission.skybox_model_animations.getRegisteredMoveables();
+
+		if (!anim_triggers.empty() || !moveable_triggers.empty()) {
+			fso_comment_push(";;FSO 22.3.0;;");
```
Goober5000, line 2495: I'd change this to 22.4 before merging.

fred2/missionsave.cpp
```diff
@@ -2773,6 +2773,50 @@ int CFred_mission_save::save_mission_info()
 			fout_version("\n\n$Skybox Model: %s.pof", out_str);
 		}
 		fso_comment_pop();
+
+		const auto& anim_triggers = The_mission.skybox_model_animations.getRegisteredAnimNames();
+		const auto& moveable_triggers = The_mission.skybox_model_animations.getRegisteredMoveables();
+
+		if (!anim_triggers.empty() || !moveable_triggers.empty()) {
+			fso_comment_push(";;FSO 22.3.0;;");
```
Goober5000, line 2781: I'd change this to 22.4 before merging.

Goober5000 (commented):
Incomplete review on 14/17 files

code/starfield/starfield.cpp
BMagnu: We'd have to pass the information that a submodel animation might exist to there, which isn't worth the effort for one measly model instance IMO

Goober5000: Fair enough

code/scripting/api/libs/mission.cpp
```diff
@@ -1962,6 +1963,111 @@ ADE_FUNC(getLineOfSightFirstIntersect, l_Mission, "vector from, vector to, [tabl
...
+	const char* trigger = nullptr;
+
+	if (!ade_get_args(L, "sss", &target, &type, &trigger))
+		return ADE_RETURN_NIL;
+
+	polymodel_instance* pmi;
+	animation::ModelAnimationSet* set;
+
+	if (stricmp(target, "cockpit") == 0) {
+		if (Player_ship->cockpit_model_instance < 0)
```
Goober5000, line 1983: I would also add a check for `Player_ship` being null.  Probably unlikely to get hit, but you never know.

code/scripting/api/libs/mission.cpp
```diff
@@ -1962,6 +1963,111 @@ ADE_FUNC(getLineOfSightFirstIntersect, l_Mission, "vector from, vector to, [tabl
...
+	luacpp::LuaTable values;
+
+	if (!ade_get_args(L, "sst", &target, &name, &values))
+		return ADE_RETURN_NIL;
+
+	polymodel_instance* pmi;
+	animation::ModelAnimationSet* set;
+
+	if (stricmp(target, "cockpit") == 0) {
+		if (Player_ship->cockpit_model_instance < 0)
```
Goober5000, line 2030: Also check `Player_ship` here.

code/ship/ship.cpp
```diff
@@ -7709,8 +7719,12 @@ void ship_init_cockpit_displays(ship *shipp)
 	ship_set_hud_cockpit_targets();
 }
 
-void ship_clear_cockpit_displays()
+void ship_clear_cockpit_displays(ship* shipp)
```
Goober5000, line 7722: I'm thinking this should be called `ship_close_cockpit_displays`, but maybe that is out of scope for this PR.

Goober5000 (commented):
Finished review.

Goober5000 (approved):

PR #4945, opened 2022-11-28
modular ai classes table

MjnMixael:
Makes ai.tbl modular with -aic.tbm. The real benefit is modifying previously parsed classes, but adding via tbm is also allowed. In that case a warning is printed to the log that it may affect ai autoscaling since autoscaling works with the ai class index position. However, the effects of that can be limited or removed with other existing table options, so it's really up to the modder on that.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -653,7 +653,7 @@ static ai_class* get_ai_class_pointer(const char* class_name)
 	return nullptr;
 }
 
-static int num_classes_from_tbm = 0;
+static bool class_added_in_tbm = false;
 
 void parse_ai_class()
 {
@@ -695,7 +695,7 @@ void parse_ai_class()
 
 		if (Parsing_modular_table) {
 			mprintf(("New Ai Class '%s' found in modular table\n", aicp->name));
-			num_classes_from_tbm++;
+			class_added_in_tbm = true;
 		}
 
 	}
@@ -938,7 +938,7 @@ void ai_init()
 
 		parse_modular_table("*-aic.tbm", parse_aitbl);
 
-		if (num_classes_from_tbm > 0) {
+		if (class_added_in_tbm) {
 			mprintf(("One or more Ai Classes were added in a modular table. This may affect "
 					 "ai autoscaling. If that's not desired then use a tbl instead of a tbm!\n"));
 		}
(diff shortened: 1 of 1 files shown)


code/ai/aicode.cpp
```diff
@@ -572,8 +572,17 @@ int ai_maybe_autoscale(int absolute_index)
...
+	// Set the lowest values here so the table can be modular
+	for (int i = 0; i < NUM_SKILL_LEVELS; i++) {
+		aicp->ai_accuracy[i] = FLT_MIN;
+		aicp->ai_evasion[i] = FLT_MIN;
+		aicp->ai_courage[i] = FLT_MIN;
+		aicp->ai_patience[i] = FLT_MIN;
+	}
+
+	for (int i = 0; i < NUM_SKILL_LEVELS; i++)
 	{
```
BMagnu, line 586: Why are these two loops, why not merge them into one?

code/ai/aicode.cpp
```diff
@@ -881,15 +927,24 @@ void ai_init()
...
+		Ai_class_names = (char**)vm_malloc(Num_alloced_ai_classes * sizeof(char*));
+
+		parse_aitbl("ai.tbl");
+
+		parse_modular_table("*-aic.tbm", parse_aitbl);
+
+		if (class_added_in_tbm) {
+			mprintf(("One or more Ai Classes were added in a modular table. This may affect "
+					 "ai autoscaling. If that's not desired then use a tbl instead of a tbm!\n"));
 		}
```
BMagnu, line 944: Not sure about this. It seems to me as any autoscaling does not happen during parse, which means the behaviour is unaffected as far as I can tell

BMagnu (changes requested):

code/ai/aicode.cpp
MjnMixael: My understanding is that ai autoscaling is done by the index of the ai classes in the array. When adding an ai class using a tbm, even if someone intends to use it as something between Coward and Captain, that probably won't be the case just because of where it's stored. There are already methods to turn off autoscaling, but there's not really a way to change how it works (IE: Ordering the classes according to difficulty). Hence the warning.

code/ai/aicode.cpp
MjnMixael: Eh, I was probably trying to change as little code as possible. Merged.

code/ai/aicode.cpp
BMagnu: Ah, I see.

BMagnu (approved):

PR #4946, opened 2022-11-28
per weapon auto aim feature

MjnMixael:
Adds the ability to do autoaim on a per weapon basis in addition to per ship or globally from ai profiles.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -11661,9 +11661,6 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
 	int			banks_fired;				// used for multiplayer to help determine whether or not to send packet
 	banks_fired = 0;			// used in multiplayer -- bitfield of banks that were fired
 	bool has_fired = false;		// used to determine whether we should fire the scripting hook
-	bool has_autoaim, has_converging_autoaim, needs_target_pos;	// used to flag weapon/ship as having autoaim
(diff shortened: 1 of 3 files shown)


code/ship/ship.cpp
```diff
@@ -11782,6 +11759,30 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
```
BMagnu, line 11759: Since we now do autoaim per weapon, we should move the variable declaration of ``float autoaim_fov`` from the top of the function to the top of this for loop, to make sure that it doesn't erroneously use it later when it must be defined per-weapon

code/ship/ship.cpp
```diff
@@ -11782,6 +11759,30 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
...
+			}
+			else
+			{
+				target_position = Objects[aip->target_objnum].pos;
+			}
+
+			dist_to_target = vm_vec_dist_quick(&target_position, &obj->pos);
+		}
+
 		if (needs_target_pos) {
```
BMagnu, line 11786: Since this if is directly after an if with the same condition, just fuse these two ifs

BMagnu (changes requested):

BMagnu (approved):

Goober5000 (approved):

Issue #4429, opened 2022-06-20
No method available in LUA to clear mouse button downcount


z64555:
Revealed by issue #4173, the `isMouseButtonDown` LUA method only checks if a mouse button is currently down.  It does not clear the downcounter, which is incremented on every mouse button event, and can lead to controls conflicts/overlays with the LUA scripts.

This also means that there's no method to service a singular mouse button event in LUA, meaning mission designers have to come up with their own solutions to implement a one-shot reaction event instead of continuously firing the event while the mouse button is down.

Old scripts, such as Axem's Lua Promptbox script, got around the issue by turning on/off the mouse to prevent the controls from collecting mouse events.  Now that the mouse is toggled between mouse mode and joystick mode, scripts for new mods will have to account for the possibility that both joy0 controls and mouse controls may be used/bound.

z64555:
Some notes:
 * `mouse.h:mouse_down()` is mainly used outside of the mission playspace, AKA while within some menus that's outside of the normal gameplay.
 * `mouse.h:mouse_down_count()` is used exclusively within normal gameplay.
 * `mouse_down()` only checks if the quered button is down at the time of the check, not if it had gone down at any time before the check
 * `mouse_down_count()` checks if the queried button is down at the time of the check *or* if it had gone down at any time before the check.
 * `mouse_down_count()` clears the downcounter by default.

`mouse_down()` works within the context of the menus UI because execution more or less quickly loops over the checks continously in a single thread and code execution is tightly governed.

It doesn't work that well, if at all, within gameplay because gameplay has an aspect that menus does not: time.

The controls code has a very waterfall like schema which checks all controls to see if they had perhaps been activated since last frame.  The control checks then see what button/key they are bound to, and checks their downcounter.  If a control isn't checked for whatever reason, the downcounter remains untouched until its picked up by the control's handler.

`controlsconfig.h:CCI`'s have a new bool called `locked` which may be used to temporarily disable controls, which may be of some use for LUA scripts.

Since LUA scripts can run their own MenuUI within the gamespace, they're in a weird position where they need to be able to run a menu system and be able to emulate hardcoded control triggers.

Goober5000:
See #4948, although I wrote that before reading the latest comment.  I left it in draft pending discussion.

z64555:
Another observation,

[...]

PR #4948, opened 2022-11-29
add mouseButtonDownCount and flushMouse

Goober5000:
Implements #4429.


Goober5000:
Looking at @z64555 's latest comment on #4429, isMouseButtonDown might be useful after all and we may not want to deprecate it.

z64555:
> Looking at @z64555 's latest comment on https://github.com/scp-fs2open/fs2open.github.com/issues/4429, isMouseButtonDown might be useful after all and we may not want to deprecate it.

yeah, which I've said at least a few times that some consideration has to be put in to consider if isMouseButtonDown would still be useful.  mouseButtonDownCount appears to be able to emulate isMouseButtonDown by passing false to the reset argument, but I can't say for certain if that's enough of a nail for isMouseButtonDown's coffin.

Goober5000:
Ok, this is ready for review.  It looks like the `isMouseButtonDown` function is useful on its own, so I've un-deprecated it.

z64555:
After some thought I've decided that `isMouseButtonDown()` should be kept to detect if a mouse button is currently down and `mouseButtonDownCount()` should be used to detect the number of times a button went down.

There's two primary types of controls in the codebase, one is a continuously firing control (such as weapons) and the other is a one-shot control (such as comms, secondary weapon selection, etc).

Continuous controls need to check the button state, as it should be triggering every frame that the control is down.  One-shots however need to check the falling edges of the button state (when the button goes down), and potentially trigger for each of those edges so that "Press X a certain number of times, downscroll the same number of times" holds true.

Although `mouseButtonDownCount()` may be able to emulate `isMouseButtonDown()` by being told to not reset the counter and thus return true, it won't be able to detect when the button is let up.

z64555 (approved):
Mousewheel directions are not included, but those are a separate problem I hope to addressed Soon(TM).

PR #4949, opened 2022-11-29
clean traitor.tbl parsing

MjnMixael:
Cleans up traitor.tbl parsing by making `$Num Stages` and `$Formula` optional entries. Both entries' values are required to be something specific and in the case of Num Stages would assert if anything else was in the table. So, this sets the values and skips straight to parsing `$Multi Text` so that the retail table still parses properly.

Kept the old parsing as comments in case traitor.tbl is expanded upon sometime in the future.


Goober5000 (approved):
Good idea.

PR #4956, opened 2022-12-01
Fixes CTD with decals on dying ships

wookieejedi:
If a ship was getting shot with weapons that used decals, the decal code would hit an assertation, because the invalid decals were not being all removed.

This PR updates the decal cleanup loop, to ensure that no decals are missed, and all invalid ones are removed. Tests confirm this fixes the issue. Thanks to @BMagnu and @asarium for the clarifications and advice with this problem. 

Also cleaned up some comments and moved another assertation before setting of variables.


Goober5000 (approved):
Excellent detective work 👍

PR #4957, opened 2022-12-01
Expose decal options useful for modders

wookieejedi:
Now that decals are stable, this PR exposes two options that are very useful for modders:
1) Use a random range for the radius of decals (lifetime already uses a random range to excellent effect)
2) Add option to allow decals to use a random rotation when created.

Both are tested and work well.


Goober5000 (changes requested):
Commented elsewhere, but here they are again:
> PI should be PI2
> angle initialization should use `vm_angles_new`

Goober5000 (approved):

PR #4961, opened 2022-12-04
clean up support ship mode in all cases

Goober5000:
Whether a ship is a support ship or its target, and whether it is destroyed or vanished, its repair/rearm mode should be cleaned up in a consistent way.  This fixes a bug in Series Resurrecta where calling `ship-vanish` on a support ship put things into a broken state.


wookieejedi (approved):
Cleanup LGTM

PR #4962, opened 2022-12-04
add some additional weapon checks

Goober5000:
Enhance set-primary-weapon and set-secondary-weapon to check the validity of the weapon bank and weapon type.


wookieejedi (approved):
LGTM

PR #4963, opened 2022-12-04
make message queue dynamic

Goober5000:
Use a vector for storing the message queue so it can be whatever size the mission needs.

As part of this, add `free` and `vm_free` deleters to be usable with `unique_ptr` and enhance `insertion_sort` to be usable with vectors.

Fixes ~~#4959~~ and #4968.

### DIFF
--- a/code/globalincs/memory/memory.h
+++ b/code/globalincs/memory/memory.h
@@ -46,3 +46,24 @@ inline void *vm_realloc(void *ptr, size_t size)
 
 	return ret_ptr;
 }
+
+
+// For use with unique_ptr
+template <typename T>
+struct FreeDeleter
+{
+	void operator()(T* const p) const
+	{
+		free(p);
+	}
+};
+
+// For use with unique_ptr
+template <typename T>
+struct VmFreeDeleter
+{
+	void operator()(T* const p) const
+	{
+		vm_free(p);
+	}
+};
--- a/code/globalincs/memory/memory.h
+++ b/code/globalincs/memory/memory.h
@@ -47,17 +47,6 @@ inline void *vm_realloc(void *ptr, size_t size)
 	return ret_ptr;
 }
 
-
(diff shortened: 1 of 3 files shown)


code/globalincs/memory/memory.h
```diff
@@ -46,3 +46,24 @@ inline void *vm_realloc(void *ptr, size_t size)
...
+
+// For use with unique_ptr
+template <typename T>
+struct VmFreeDeleter
+{
+	void operator()(T* const p) const
+	{
+		vm_free(p);
+	}
+};
```
BMagnu, line 69: I suggest actually adding something like this, since this is a semi-common use case:
```cpp
//Managed unique pointer for data allocated using vm_malloc. Never use with data allocated with new or malloc
template<typename T> using SCP_vm_unique_ptr = std::unique_ptr<T, VmFreeDeleter<T>>;
```

code/mission/missionmessage.cpp
```diff
@@ -138,28 +138,25 @@ int Message_expire;							// timestamp to extend the duration of message bracket
 
 typedef struct message_q {
 	fix	time_added;					// time at which this entry was added
-	int	window_timestamp;			// timestamp which will tell us how long we have to play the message
+	TIMESTAMP window_timestamp;		// timestamp which will tell us how long we have to play the message
 	int	priority;					// priority of the message
 	int	message_num;				// index into the Messages[] array
-	char *special_message;			// Goober5000 - message to play if we've replaced stuff (like variables)
+	std::unique_ptr<char, VmFreeDeleter<char>> special_message;	// Goober5000 - message to play if we've replaced stuff (like variables)
```
BMagnu, line 144: This should then use the new using

BMagnu (changes requested):

BMagnu (approved):

PR #4965, opened 2022-12-05
SEXP operators, categories, and subcategories refactoring

Goober5000:
This re-engineers the SEXP identifier and categorization system to lift the limit on the number of SEXPs that can be added to categories.  This is needed because Scroll overran the limit due to the number of scripted SEXPs added.

~~In draft status pending tests.~~


code/parse/sexp/sexp_lookup.cpp
```diff
@@ -217,6 +187,17 @@ DynamicSEXP* get_dynamic_sexp(int operator_const)
 
 	return iter->second.get();
 }
+int get_category_of_subcategory(int subcategory_id)
+{
+	auto& global = globals();
```
BMagnu, line 192: should be const

code/parse/sexp/sexp_lookup.cpp
```diff
@@ -237,31 +218,29 @@ void dynamic_sexp_shutdown()
...
+	return category_id;
+}
+int add_subcategory(int parent_category, const SCP_string& name)
+{
+	auto& global = globals();
 
-	int subcategory_id = (max_subcategory_value + 1) | parent_category;
+	int subcategory_id = global.next_free_subcategory_id++;
 	op_submenu.push_back({ name, subcategory_id });
+	global.subcategory_to_category[subcategory_id] = parent_category;
```
BMagnu, line 243: The cleaner way to do this is to do an emplace(subcategory_id, parent_category) here.

code/parse/sexp/sexp_lookup.cpp
```diff
@@ -217,6 +187,17 @@ DynamicSEXP* get_dynamic_sexp(int operator_const)
...
 	return iter->second.get();
 }
+int get_category_of_subcategory(int subcategory_id)
+{
+	auto& global = globals();
+
+	auto iter = global.subcategory_to_category.find(subcategory_id);
+	if (iter == global.subcategory_to_category.end()) {
+		return OP_CATEGORY_NONE;
+	}
```
BMagnu, line 197: Right now this is fine, but if someone will use this method in the future, they'll be confused as to why this won't resolve built-in categories, as they're never added to the map.

fred2/sexp_tree.cpp
```diff
@@ -1574,11 +1573,7 @@ void sexp_tree::right_clicked(int mode)
 		for (j=0; j<(int)Operators.size(); j++) {
 			z = 0;
 			if (m_mode == MODE_CAMPAIGN) {
-				if (Operators[j].value & OP_NONCAMPAIGN_FLAG)
-					z = 1;
-
-			} else {
-				if (Operators[j].value & OP_CAMPAIGN_ONLY_FLAG)
+				if (!usable_in_campaign(Operators[j].value))
 					z = 1;
```
BMagnu, line 1577: This seems off. You've disabled noncampaign flags in campaigns, but no longer disable campaign only flags out-of-campaign. Is that intended?

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -6285,14 +6284,8 @@ std::unique_ptr<QMenu> sexp_tree::buildContextMenu(QTreeWidgetItem* h) {
...
 		z = 0;
 		if (_interface->requireCampaignOperators()) {
-			if (Operators[j].value & OP_NONCAMPAIGN_FLAG) {
+			if (!usable_in_campaign(Operators[j].value))
 				z = 1;
-			}
-
-		} else {
-			if (Operators[j].value & OP_CAMPAIGN_ONLY_FLAG) {
-				z = 1;
```
BMagnu, line 6294: Same here

BMagnu (changes requested):

fred2/sexp_tree.cpp
Goober5000: Yes, because currently there *are* no campaign-only sexps.  I figured if we ever need them, we can defer this until that time.

qtfred/src/ui/widgets/sexp_tree.cpp
Goober5000: same response

BMagnu (approved):
Seems fine now. This will need testing.
Also, I am reasonably certain that this constitutes a multi bump, given that we send SEXP operator ID's, so might also need to be considered

Goober5000:
I've tested it pretty well both in-mission and in FRED, but that's a good point about multi.

notimaginative:
Yep, this is going to require a multi bump.

I don't have time to do a multi test atm but I'll run a couple in a few hours to verify all is good there. Apart from needing the multi bump it looks okay to me though, so I don't foresee any problems.

notimaginative:
Only had a chance to do a quick test but everything appeared to work fine.

notimaginative (approved):

PR #4966, opened 2022-12-05
Properly init cockpit_model_instance to -1

BMagnu:
This would otherwise cause a crash when a cockpitless model gets loaded, leaves cockpit_model_instance at 0, and then tries to clear the cockpit instance 0 again once the ship gets cleaned.


wookieejedi (approved):

Goober5000:
Follow-up to #4936

PR #4967, opened 2022-12-05
Mod Table Settings for Arc Widths

wookieejedi:
There is currently a way to set the color of EMP and Damage Arcs in game_settings.tbl, but no way to change the width.

This PR adds options useful to modders to set the width and how it is calculated. Looking over the code and asking around, I developed these exposed options.

The options also allow modders to control the arc widths in multiple ways. For example, if you want to set a constant width for all ships you would set the `+Do Not Use Width Multiplier Over Radius:` to something huge, and if you wanted no arcs ever you would set the `+Width Default:` to 0.

They are tested and work as expected.

Also fixes old copy paste bug from #2053.


Goober5000 (approved):

code/model/modelrender.cpp
```diff
@@ -863,6 +854,17 @@ void model_render_add_lightning( model_draw_list *scene, model_render_params* in
...
+      
+      			// try and scale the size a bit so that it looks equally well on smaller vessels
+			width = Arc_width_default_damage;
+			if (pm->rad < Arc_width_no_multiply_over_radius_damage) {
+				width *= (pm->rad * Arc_width_radius_multiplier_damage);
+
+				if (width < Arc_width_minimum_damage) {
+					width = Arc_width_minimum_damage;
+				}
+			}
```
MjnMixael, line 866: You're going to want to copy this width stuff into the MARC_TYPE_DAMAGED case as well now, too.

wookieejedi: thanks, just pushed

PR #4969, opened 2022-12-05
custom colors for passive lightning

### LINKED ISSUE #3933: FR: additional option to choose color for passive lightning.
Hoping to get another +option for Passive Lightning in ships.tbl that can specify the color for that specific arc.

MjnMixael:
Implements custom colors per arc for passive lightning, defaults to damage arc color. Also cleans up handling of damage and emp arc colors by initiating the color during parse rather than on each frame for each arc. Fixes #3933


code/model/modelrender.cpp
```diff
@@ -847,25 +847,33 @@ void model_render_add_lightning( model_draw_list *scene, model_render_params* in
...
 			if ( Random::flip_coin() )	{
-				gr_init_color(&primary, std::get<0>(Arc_color_damage_p1), std::get<1>(Arc_color_damage_p1), std::get<2>(Arc_color_damage_p1));
+				primary = smi->arc_primary_color_1[i];
 			} else {
-				gr_init_color(&primary, std::get<0>(Arc_color_damage_p2), std::get<1>(Arc_color_damage_p2), std::get<2>(Arc_color_damage_p2));
+				primary = smi->arc_primary_color_2[i];
 			}
 
-			gr_init_color(&primary, std::get<0>(Arc_color_damage_s1), std::get<1>(Arc_color_damage_s1), std::get<2>(Arc_color_damage_s1));
+			secondary = smi->arc_secondary_color[i];
```
Baezon, line 865: Just for posterity, I suppose, it's unideal that this color info needs to be put on the submodel instance, even though it isn't *really* specific to the instance, but by this point things like the `ship_info` have been discarded awhile ago so that kind of data is no longer easily accessible. It's fine for now, but maybe something we should think about more in the future, if we have more 'ship_info-based rendering changes'.

Baezon (approved):

code/model/modelrender.cpp
MjnMixael: Yeah, that's precisely why I did it this way. At the very least passing the objnum down would help getting back to ship_info in cases like this where we might want access to it.

PR #4971, opened 2022-12-06
fix a crash when the Knossos explodes in A Flaming Sword

Goober5000:
The ship being accessed here is the viewer object, not the loop iterator.

Follow-up to #4936.


wookieejedi (approved):
Nice quick and simple fix

PR #4972, opened 2022-12-06
clean up mission_do_departure

Goober5000:
Since `mission_do_departure` can be called for multiple frames in a row, make sure that certain things run only once.

Viewing this PR with whitespace changes ignored will provide a more succinct overview.


wookieejedi (approved):
This will definitely help reduce log clutter and overall looks good.

PR #4975, opened 2022-12-06
enhance sendMessage to allow messages sent from ships

Goober5000:
Allow `mn.sendMessage` to send from a ship source, not only a string source, by allowing the ship object to be specified as the first argument.  Also add `ship:sendMessage` which is a more streamlined version of the same thing.


wookieejedi (approved):
Robust fix and overall looks good and works as expected.

PR #4980, opened 2022-12-07
Exposes wing formation scale and name via scripting

wookieejedi:
Now that wing formation and scale can be changed on the fly via sexp, this PR allows scripters to get and set those values. Particularly useful for LUA AI orders where you want to the current formation to tighter or loosen up or choose a form-up formation (and any combination of those).

Tested and everything works as expected.


Goober5000:
Handled by #4983

PR #4981, opened 2022-12-07
Fix edge case bugs with moveables in combination with special polymodels being animated

BMagnu:
Also forgot to account for interaction with ship templates again.


Goober5000 (approved):

wookieejedi (approved):
Thanks for the quick fix!

PR #4983, opened 2022-12-08
exposes wing formation via scripting

Goober5000:
A rework of #4980, adding a wing formation type and a way to access them via `tb.WingFormations`.

Also fixes some flaws in the fireball class, which the wing formation class is based on.

### DIFF
--- a/code/scripting/api/objs/wingformation.cpp
+++ b/code/scripting/api/objs/wingformation.cpp
@@ -1,0 +1,82 @@
+//
+//
+
+#include "wingformation.h"
+
+#include "ship/ship.h"
+
+namespace scripting {
+namespace api {
+
+// NOTE: the wing formation index runs 1 through Wing_formations.size(), offset by 1 from the internal contents.  Index 0 is Default.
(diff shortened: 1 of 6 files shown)


code/scripting/api/objs/wingformation.cpp
```diff
@@ -0,0 +1,82 @@
...
+	if (formation_id < 0 || formation_id >= (int)Wing_formations.size())
+		return ade_set_error(L, "s", "");
+
+	if (ADE_SETTING_VAR)
+		LuaError(L, "This property is read-only");
+
+	return ade_set_args(L, "s", Wing_formations[formation_id].name);
+}
+
+ADE_FUNC(isValid, l_WingFormation, NULL, "Detects whether handle is valid", "boolean", "true if valid, false if handle is invalid, nil if a syntax/type error occurs")
```
wookieejedi, line 69: nullptr?

wookieejedi (approved):
Looks good now. Also, noting here that this closes #4980

PR #4984, opened 2022-12-08
Add FRED SpaceMouse support

### LINKED ISSUE #3673: FR: QtFRED 3D Mouse Support
This is a long term stretch goal, but a worthy one if you ask me.

I'd love to have 3D Mouse Support in the main QtFRED UI for camera navigation. 3DConnexion has an SDK available at no cost [here.](https://3dconnexion.com/us/software-developer-program/)

I want it bad enough that I'll purchase a 3D Mouse for the developer that decides to take this on when the time is right.

BMagnu:
For now, only some 3Dconnexion devices.
Closes #3673.
Button support is possible too, but put off for now.


code/io/spacemouse.h
```diff
@@ -0,0 +1,62 @@
...
+			@returns Whether or not the button is pressed. Always false for buttons not present on this space mouse.
+			*/
+			bool isButtonPressed(size_t number);
+
+			/*
+			@brief Polls connected HID devices and tests for known and supported space mice
+			@param pollingFrequency The frequency any found space mice should be polled for HID updates, in ms.
+			@returns An optional SpaceMouse object, if found
+			*/
+			static std::unique_ptr<SpaceMouse> searchSpaceMouses(int pollingFrequency = 10);
```
Goober5000, line 59: I wonder whether this should be `searchSpaceMice`...

code/io/spacemouse.h
```diff
@@ -0,0 +1,62 @@
...
+			enum class Protocol { CONNEXION_3D_OLD, CONNEXION_3D_NEW } protocol;
+		};
+
+		class SpaceMouse {
+			const SpaceMouseDefinition& m_definition;
+			const int m_pollingFrequency;
+
+			hid_device* m_deviceHandle;
+			SpaceMouseMovement m_current;
+			SCP_vector<bool> m_keypresses;
```
Goober5000, line 32: Given the weird specification of `std::vector<bool>`, are there any gotchas to using it here?

code/io/spacemouse.cpp
```diff
@@ -0,0 +1,179 @@
...
+			switch (data.type) {
+			case 1:
+				m_current.translation.xyz.x = static_cast<float>(data.data.type1.trans_x) / 150.0f;
+				m_current.translation.xyz.y = static_cast<float>(data.data.type1.trans_y) / -150.0f;
+				m_current.translation.xyz.z = static_cast<float>(data.data.type1.trans_z) / -150.0f;
+				break;
+			case 2:
+				m_current.rotation.p = static_cast<float>(data.data.type2.rot_x) / -350.0f;
+				m_current.rotation.b = static_cast<float>(data.data.type2.rot_y) / 350.0f;
+				m_current.rotation.h = static_cast<float>(data.data.type2.rot_z) / 350.0f;
```
Goober5000, line 59: Where do these divisors come from?

qtfred/src/mission/EditorViewport.cpp
```diff
@@ -318,9 +319,22 @@ void EditorViewport::process_system_keys(int key) {
...
 
+		if (spacemouse != nullptr) {
+			auto spacemouse_movement = spacemouse->getMovement();
+			spacemouse_movement.handleNonlinearities(Fred_spacemouse_nonlinearity);
+			view_controls.pitch += spacemouse_movement.rotation.p;
+			view_controls.vertical += spacemouse_movement.translation.xyz.z;
+			view_controls.heading += spacemouse_movement.rotation.h;
+			view_controls.sideways += spacemouse_movement.translation.xyz.x;
+			view_controls.bank += spacemouse_movement.rotation.b;
+			view_controls.forward += spacemouse_movement.translation.xyz.y;
```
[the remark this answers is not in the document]
BMagnu: Yeah.
While I haven't tested it in QTFred but just normal FRED (I assume that they react identically to the axis mapping), these axis mappings are so that it feels intuitive to me.
The actual x/y/z mappings of the translation vector are chosen in such a way that for the other mode that uses not the CI-struct but actual vec3d / angles, I can just add the vector and not have to change around axes.

code/io/spacemouse.h
BMagnu: changed

code/io/spacemouse.cpp
BMagnu: So, the sign of the divisiors is set so that the reported value is in the right direction for FSO's coordinate system.
The value is kind of arbitrary, really. I started with 350 as that's what other FOSS implementations use (presumably to get the 2^15 down to roughly -100 to 100, I guess?), but then translation was much too slow in FRED, even considering the speed settings. So I just bumped it a bit.
If someone wants to change that, the nonlinearity settings should suffice just fine for that.

code/io/spacemouse.h
BMagnu: No, shouldn't.
Random access is still available, so it doesn't really concern us here, since we never take a pointer to a value.
But it doesn't matter too much anyways, since the buttons aren't yet parsed from the protocol, and are also as of yet never used. I plan to, eventually, but that's a followup PR

code/io/spacemouse.cpp
Goober5000: All right

Goober5000 (approved):

PR #4987, opened 2022-12-09
Feature/dear imgui

The-E:
This PR is still in draft because there's still some features and cleanup missing, but it's close enough to done that getting feedback from others would be good.

The goal of this is to revamp the lab again, this time by incorporating the [Dear Imgui](https://github.com/ocornut/imgui) library to handle all the messy stuff of UI building. Since this library uses a fundamentally different approach to doing UI stuff than what wmcgui used, this required a fundamental restructuring of the UI code, but the result, I feel, is worth it.


code/globalincs/pstypes.h
```diff
@@ -20,6 +20,8 @@
 #include "globalincs/vmallocator.h"
 #include "utils/strings.h"
 
+#include "imgui.h"
```
BMagnu, line 23: Is there a pressing need to have an imgui include this far up in the chain, instead of, say, in the cpp-files of the lab itself?
99% of the code will not need it, but it will basically pollute the namespace with imgui stuff for (almost) every file.

The-E: Yes, unfortunately. Dear Imgui clashes with our custom allocator, requiring a special case to be added later in that same file.

BMagnu: If that is just for the declaration of ``ImDrawListSplitter``, couldn't you just forward declare the class? Since it's just used as a pointer, an incomplete type should suffice here

The-E: Problem is that that doesn't work, at least on MSVC. Trying to forward declare this struct just leads to compiler errors.

BMagnu: I just tested this. There is a way to solve this properly (tested on a local checkout):
Add an enable_if to the template memset method we have and use the forward declare. This prevents the ImGui stuff from seeing this function overload which was for some reason preferred to the more specified function overload for the normal forward declare:
```cpp
template<typename T>
void *memset_if_trivial_else_error(typename std::enable_if<!std::is_same<T, ImDrawListSplitter>::value, T>::type *memset_data, int ch, size_t count)
{
```

The-E: still fails on my end: ```C:\Users\fwolt\Source\fs2open.github.com\code\globalincs\pstypes.h(497): error C2338: static_assert failed: 'memset on non-trivial object'
  C:\Users\fwolt\Source\fs2open.github.com\lib\imgui\imgui.h(2456): note: see reference to function template instantiation 'void *std::memset_if_trivial_else_error<ImDrawListSplitter>(T *,int,size_t)' being compiled
          with
          [
              T=ImDrawListSplitter
          ]```

BMagnu: Just to make sure, you replaced the already templated memset and not the ImGui memset, right?
On MSVC2017, the following compiles when replacing the memset section, and with the imgui include in pstypes.h removed
```cpp
struct ImDrawListSplitter;

// Put into std to be compatible with code that uses std::mem*
namespace std
{
	template<typename T>
	using trivial_check = std::is_trivially_copyable<T>;

	template<typename T>
	void* memset_if_trivial_else_error(typename std::enable_if<!std::is_same<T, ImDrawListSplitter>::value, T>::type* memset_data, int ch, size_t count)
	{
		static_assert(trivial_check<T>::value, "memset on non-trivial object");
		return ptr_memset(memset_data, ch, count);
	}

	// assume memset on a void* is "safe"
	// only used in cutscene/mveplayer.cpp:mve_video_createbuf()
	inline void *memset_if_trivial_else_error(void *memset_data, int ch, size_t count)
	{
		return ptr_memset(memset_data, ch, count);
	}

	// Dear ImGui triggers these as well, so we need to let them pass
	inline void *memset_if_trivial_else_error(ImDrawListSplitter *memset_data, int ch, size_t count)
	{
		return ptr_memset(memset_data, ch, count);
	}
```
I still get some (probably unrelated) warnings from the lab code, but nothing that prevents compilation outright

The-E: ahh, that's what I was missing. Works now.

EatThePath (commented):
Partial review done on my end, I have yet to really tackle `lab_ui.cpp` , and I am just assuming the /lib/ code is taken as-is so didn't read through that any. Everything else looks good so far, will try to take another run at `lab_ui.cpp` soon.

The-E:
> Partial review done on my end, I have yet to really tackle `lab_ui.cpp` , and I am just assuming the /lib/ code is taken as-is so didn't read through that any. Everything else looks good so far, will try to take another run at `lab_ui.cpp` soon.

yeah, everything under lib is just a straight copy of imgui 1.88... which I should update to 1.89.1, come to think of it

code/lab/dialogs/lab_ui.cpp
```diff
@@ -0,0 +1,964 @@
...
+			}
+		}
+	}
+}
+
+void LabUi::build_options_menu()
+{
+	with_Menu("Options")
+	{
+		ImGui::MenuItem("Render options", nullptr, &show_render_options_dialog);
```
EatThePath, line 150: Chunks like this make me think this code might do with a `using namespace Imgui;` at the top. With these undecorrated macros everywhere, being explicit about the namespace for methods feels more confusing than clarifying, personally. Not a make or break issue though.

code/lab/dialogs/lab_ui.cpp
```diff
@@ -0,0 +1,964 @@
...
+		with_Window("Select object and background")
+		{
+
+			with_CollapsingHeader("Displayed Object")
+			{
+				build_ship_list();
+
+				build_weapon_list();
+			}
+
```
EatThePath, line 190: I forgot to say anything while testing but seeing it here reminded me of the simplicity of making these kinds of changes in imGui...
I think having the background and object lists in a single window is a bit of a step back. More than once I've gone through several ships and backgrounds in one session, in lists long enough that the new arrangment would be kind of a hassle. Having them side by side in seperately scrolled lists keeps from having to bounce up and down.
On the other hand I could easily take that on as my own followup PR if you don't want to make changes to the layout at this stage

EatThePath (approved):
Besides a few optional comments it looks good to me

code/lab/dialogs/lab_ui.cpp
The-E: Good idea tho, done.

code/lab/dialogs/lab_ui.cpp
The-E: Also done.

TRBlount (commented):
It looks like the WMCGUI impl was correctly ported to imgui. I'm going to build and test this for a bit to see if I can break anything.

TRBlount (approved):
I was unable to find any issues in game and everything appears to be implemented correctly. LGTM

PR #4995, opened 2022-12-12
fix subcategory comparison

### LINKED ISSUE #4994: FRED no longer allows selection of proper sexps in context menus
The latest builds of FRED are unusable because all sexps (except maybe LuaSEXPs?) are greyed out when they shouldn't be. Guessing goober didn't test his big sexp merges in #4965

Goober5000:
Empty subcategories are now OP_SUBCATEGORY_NONE, not -1.  Fixes #4994.

Also fixes a cast warning in missionparse.cpp.


MageKing17 (approved):
Tested; fixes the issue.

PR #5000, opened 2022-12-13
Allow player turrets to fire despite play dead

### LINKED ISSUE #4998: On BPC "The Blade Itself" (the Karuna Command mission), turrets do not fire on current nightlies.
As ever, mission works as intended in FSO 22.2

Thus far have been able to determine that mission
works fine on 20221123
hard crashes on 20221127
no longer works as intended on 20221128

Baezon:
Follow-up to #4934. Fixes #4998 and restores behavior from before #3713 . A little unfortunate, maybe, but designers can still lock player turrets if they don't want them to do anything.


Goober5000 (changes requested):
The same modification should be done in `ship_move_subsystems`

Goober5000 (approved):

PR #5003, opened 2022-12-14
Generalize linear interpolation

JohnAFernandez:
This creates a (very small) api for linear interpolation between two recorded positions, velocitites, and desired velocities and moves multiplayer over to that system.  Now that it is generalized, other games systems (esp. weapons fire) can now use this code to temporarily set an object's position to what it would be between frames.

In draft pending testing.

### DIFF
--- a/code/object/object.cpp
+++ b/code/object/object.cpp
@@ -2101,7 +2101,7 @@ bool obj_compare(object* left, object* right) {
 	return OBJ_INDEX(left) == OBJ_INDEX(right);
 }
 
-void physics_populate_snapshot(physics_snapshot& snapshot, object* objp)
+void physics_populate_snapshot(physics_snapshot& snapshot, const object* objp)
 {
     Assertion(objp != nullptr, "Bad object (nullptr) passed to physics_overwrite_snapshot, please report to the SCP!");
 
--- a/code/object/object.cpp
+++ b/code/object/object.cpp
@@ -657,8 +657,11 @@ void obj_delete(int objnum)
 		Error( LOCATION, "Unhandled object type %d in obj_delete_all_that_should_be_dead", objp->type );
 	}
 
+	// this avoids include issues from physics state, multi interpolate and object code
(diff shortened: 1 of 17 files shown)


JohnAFernandez:
I was finally able to run a test for this now that I have a better dev machine.  Interpolation works just like in the old system, so I'm marking as ready for review.  This is a stepping stone to getting primary fire rates higher than the frame rate.

JohnAFernandez:
This really needs to get in before 24.2, otherwise higher weapon fire rate than framerate is literally not going to happen.

code/object/object.cpp
```diff
@@ -2095,3 +2100,28 @@ bool obj_compare(object* left, object* right) {
 
 	return OBJ_INDEX(left) == OBJ_INDEX(right);
 }
+
+void physics_populate_snapshot(physics_snapshot& snapshot, object* objp)
```
BMagnu, line 2104: Ideally, this takes a const pointer.

BMagnu (changes requested):
A very tiny constness thing, but this looks good and clean.

code/object/object.cpp
JohnAFernandez, line 2104:
```suggestion
void physics_populate_snapshot(physics_snapshot& snapshot, const object* objp)
```

code/object/object.h
```diff
@@ -387,4 +384,30 @@ void obj_render_queue_all();
...
+
+/**
+ * @brief Populate a physics snapshot directly from the info in an object
+ *
+ * @param[in,out] snapshot Destination physics snapshot
+ * @param[in]     objp The object pointer that we are pulling information from
+ *
+ * @author J Fernandez
+ */
+void physics_populate_snapshot(physics_snapshot& snapshot, object* objp);
```
JohnAFernandez, line 398:
```suggestion
void physics_populate_snapshot(physics_snapshot& snapshot, const object* objp);
```

BMagnu (approved):

PR #5008, opened 2022-12-14
Add missing asteroid debris argument type error checking

MjnMixael:
There is a case where a mod could have the three asteroids but no debris types, so this checks for that and invalidates the set-debris-field sexp in that case. It also checks for invalid debris types in the sexp itself.


fred2/sexp_tree.cpp
```diff
@@ -3498,6 +3498,12 @@ int sexp_tree::query_default_argument_available(int op, int i)
 			}
 			return 0;
 
+		case OPF_ASTEROID_DEBRIS:
+			if ((Asteroid_info.size() - 3) > 0) {
```
Goober5000, line 3502: Constants or defines are preferred over plain numbers, so this `3` should be `NUM_ASTEROID_POFS`

Goober5000 (changes requested):

fred2/sexp_tree.cpp
MjnMixael: Done

Goober5000 (approved):

PR #5009, opened 2022-12-14
Dynamic sexp enum

### LINKED ISSUE #3381: FR: Further expansion on LUA SEXPs types
The end-goal here is once again FREDer QoL. I have a handful of lua sexps that accept custom data. I was thinking just how much easier it would be to have a pre-populated list in FRED of custom data similar to some retail sexps. Examples of where this would be useful...

- List of music files that can be played back
- Ship variants that can be set (I use a script that will automatically set turrets to certain weapons with a single sexp)
- Custom send message sexp could use a short list to create message priority (Low, Normal, High)

My idea is something like this where the function would get the data as a string, but in this way it would pre-populate acceptable strings in FRED...

For longer lists:
```
#LUA SEXPs
$Operator: dynamic-sexp
$Category: Change
$Subcategory: Scripted
etc....
Parameter:
  +Description: String
  +Type: Enum
  +Enum Name: String
#END

#ENUMS
$Enum: String
  +String
  +String
  +String
$Enum: String
  +String
  +String
  +String
Etc.....
#END
```

For shorter lists we could just do it right in the Parameter section:
```
#LUA SEXPs
$Operator: dynamic-sexp
$Category: Change
$Subcategory: Scripted
etc....
Parameter:
  +Description: String
  +Type: Enum
  +Enum: ( "High" "Normal" "Low" )
#END
```

A second semi related request would be the ability to list multiple types in a Parameter. Think SEXPs that can accept a Ship or Wing, or a Ship and a String (#Command for example in a messages based function).

```
$Parameter:
  +Description: The source of the message
  +Type: ship
  +Type: string ; eg. #Command
```

MjnMixael:
Adds the ability to specify a custom enumeration for dynamic lua sexps. Internally the sexp argument type starts at the end of the type enum list and is incremented from there. That is used to find the correct list of enumerations in the `Dynamic_enums` vector. This fixes #3381 .

Currently this is only done inline, meaning if you have the same enum list more than once it will need to be specified for each sexp. There is a path forward from here to specify enums earlier in the sexps.tbl and then reference those for each sexp which will be cleaner. But I felt that would be better as a follow-up enhancement to this feature since this PR has enough spaghetti.

### DIFF
--- a/code/parse/sexp.h
+++ b/code/parse/sexp.h
@@ -28,6 +28,7 @@ struct ship_obj;
 
 // Operator argument formats (data types of an argument)
 enum : int {
+	UNUSED,							// argument types need to start at 1 instead of 0
 	OPF_NONE,						// argument cannot exist at this position if it's this
 	OPF_NULL,						// no value.  Can still be used for type matching, however
 	OPF_BOOL,
@@ -133,7 +134,7 @@ enum : int {
 	OPF_WING_FORMATION,				// Goober5000 - as defined in ships.tbl
 
 	//Must always be at the end of the list
-	First_available_list_id
+	First_available_opf_id
 };
 
 struct dynamic_sexp_enum_list {
--- a/code/parse/sexp.h
+++ b/code/parse/sexp.h
@@ -135,110 +135,12 @@ enum : int {
 	First_available_list_id
 };
 
-// Operator argument formats (data types of an argument)
-/*#define OPF_NONE 1 // argument cannot exist at this position if it's this
-#define	OPF_NULL				2		// no value.  Can still be used for type matching, however
-#define	OPF_BOOL				3
-#define	OPF_NUMBER				4
-#define	OPF_SHIP				5
-#define	OPF_WING				6
-#define	OPF_SUBSYSTEM			7
-#define	OPF_POINT				8		// either a 3d point in space, or a waypoint name
-#define	OPF_IFF					9
-#define	OPF_AI_GOAL				10		// special to match ai goals
-#define	OPF_DOCKER_POINT		11		// docking point on docker ship
-#define	OPF_DOCKEE_POINT		12		// docking point on dockee ship
-#define	OPF_MESSAGE				13		// the name (id) of a message in Messages[] array
-#define	OPF_WHO_FROM			14		// who sent the message -- doesn't necessarily have to be a ship!!!
-#define	OPF_PRIORITY			15		// priority for messages
-#define	OPF_WAYPOINT_PATH		16		// name of a waypoint
-#define	OPF_POSITIVE			17		// positive number or zero
-#define	OPF_MISSION_NAME		18		// name of a mission for various mission related things
-#define	OPF_SHIP_POINT			19		// a waypoint or a ship
-#define	OPF_GOAL_NAME			20		// name of goal (or maybe event?) from a mission
-#define	OPF_SHIP_WING			21		// either a ship or wing name (they don't conflict)
-#define OPF_SHIP_WING_WHOLETEAM			22	// Karajorma - Ship, wing or an entire team's worth of ships
-#define	OPF_SHIP_WING_SHIPONTEAM_POINT	23	// name of a ship, wing, any ship on a team, or a point
-#define OPF_SHIP_WING_POINT		24
-#define OPF_SHIP_WING_POINT_OR_NONE	25	// WMC - Ship, wing, point or none
-#define	OPF_SHIP_TYPE			26		// type of ship (fighter/bomber/etc)
-#define	OPF_KEYPRESS			27		// a default key
-#define	OPF_EVENT_NAME			28		// name of an event
-#define	OPF_AI_ORDER			29		// a squadmsg order player can give to a ship
-#define	OPF_SKILL_LEVEL			30		// current skill level of the game
-#define	OPF_MEDAL_NAME			31		// name of medals
-#define	OPF_WEAPON_NAME			32		// name of a weapon
-#define	OPF_SHIP_CLASS_NAME		33		// name of a ship class
-#define	OPF_CUSTOM_HUD_GAUGE	34		// name of custom HUD gauge
-#define	OPF_HUGE_WEAPON			35		// name of a secondary bomb type weapon
-#define	OPF_SHIP_NOT_PLAYER		36		// a ship, but not a player ship
-#define	OPF_JUMP_NODE_NAME		37		// name of a jump node
-#define	OPF_VARIABLE_NAME		38		// variable name
-#define	OPF_AMBIGUOUS			39		// type used with variable
-#define	OPF_AWACS_SUBSYSTEM		40		// an awacs subsystem
-#define OPF_CARGO				41		// Goober5000 - a cargo string (currently used for set-cargo and is-cargo)
-#define OPF_AI_CLASS			42		// Goober5000 - an AI class
-#define OPF_SUPPORT_SHIP_CLASS	43		// Goober5000 - a support ship class
-#define OPF_ARRIVAL_LOCATION	44		// Goober5000 - a ship arrival location
-#define OPF_ARRIVAL_ANCHOR_ALL	45		// Goober5000 - all of a ship's possible arrival anchors
-#define OPF_DEPARTURE_LOCATION	46		// Goober5000 - a ship departure location
-#define OPF_SHIP_WITH_BAY		47		// Goober5000 - a ship with a fighter bay
-#define OPF_SOUNDTRACK_NAME		48		// Goober5000 - the name of a music soundtrack
-#define OPF_INTEL_NAME			49		// Goober5000 - the name of an intel entry in species.tbl
-#define OPF_STRING				50		// Goober5000 - any old string
-#define OPF_ROTATING_SUBSYSTEM	51		// Goober5000 - a rotating subsystem
-#define OPF_NAV_POINT			52		// Kazan	  - a Nav Point name
-#define OPF_SSM_CLASS			53		// Goober5000 - an SSM class
-#define OPF_FLEXIBLE_ARGUMENT	54		// Goober5000 - special to match for when-argument
-#define OPF_ANYTHING			55		// Goober5000 - anything goes, except containers
-#define OPF_SKYBOX_MODEL_NAME	56		// taylor - changing skybox model
-#define OPF_SHIP_OR_NONE		57		// Goober5000 - an "optional" ship argument
-#define OPF_BACKGROUND_BITMAP	58		// phreak - name of a background bitmap
-#define OPF_SUN_BITMAP			59		// phreak - name of a background bitmap
-#define OPF_NEBULA_STORM_TYPE	60		// phreak - name a nebula storm
-#define OPF_NEBULA_POOF			61		// phreak - name of a nebula poof
-#define OPF_TURRET_TARGET_ORDER	62		// WMC - name of a turret target type (see aiturret.cpp)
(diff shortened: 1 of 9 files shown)


code/parse/sexp/LuaSEXP.cpp
```diff
@@ -518,6 +525,33 @@ void LuaSEXP::parseTable() {
...
+
+			SCP_vector<SCP_string> list_items;
+			while (optional_string("+Enum:")) {
+				SCP_string item;
+				stuff_string(item, F_NAME);
+				list_items.push_back(item);
+			}
+
+			if ((int)list_items.size() > 0) {
+				thisList.list = list_items;
```
BMagnu, line 546: Make it ``thisList.list = std::move(list_items);``
This helps the compiler to optimize, since you tell it "I don't need list_items anymore, you are allowed to destroy it in the process of moving it to thisList.list". This may enable the compiler to avoid copying the list, and for example in this case automatically reusing some space.

fred2/sexp_tree.cpp
```diff
@@ -7348,6 +7356,25 @@ sexp_list_item *sexp_tree::get_container_multidim_modifiers(int con_data_node) c
 	return head.next;
 }
 
+sexp_list_item* sexp_tree::check_for_dynamic_sexp_enum(int opf)
+{
+	sexp_list_item head;
+
+	int item = opf - First_available_list_id;
+
+	if (item <= (int)Dynamic_enums.size()) {
```
BMagnu, line 7365: this must be a true <, not a <=.
accessing element i from a vector with size i will cause a CTD

fred2/sexp_tree.cpp
```diff
@@ -7348,6 +7356,25 @@ sexp_list_item *sexp_tree::get_container_multidim_modifiers(int con_data_node) c
...
+
+		for (int i = 0; i < (int)Dynamic_enums[item].list.size(); i++) {
+			head.add_data(Dynamic_enums[item].list[i].c_str());
+		}
+		return head.next;
+	} else {
+		// else if opf is invalid do this
+		UNREACHABLE("Unhandled SEXP argument type!"); // unknown OPF code
+		return NULL;
```
BMagnu, line 7374: nullptr

fred2/sexp_tree.cpp
```diff
@@ -7348,6 +7356,25 @@ sexp_list_item *sexp_tree::get_container_multidim_modifiers(int con_data_node) c
...
+sexp_list_item* sexp_tree::check_for_dynamic_sexp_enum(int opf)
+{
+	sexp_list_item head;
+
+	int item = opf - First_available_list_id;
+
+	if (item <= (int)Dynamic_enums.size()) {
+
+		for (int i = 0; i < (int)Dynamic_enums[item].list.size(); i++) {
+			head.add_data(Dynamic_enums[item].list[i].c_str());
```
BMagnu, line 7368: Just make this a for each loop:
```suggestion
		for (const SCP_string& enum_item : Dynamic_enums[item].list) {
			head.add_data(enum_item.c_str());
```

fred2/sexp_tree.cpp
```diff
@@ -3499,7 +3499,15 @@ int sexp_tree::query_default_argument_available(int op, int i)
 			return 0;
 
 		default:
-			Int3();
+			if (Dynamic_enums.size() > 0) {
```
BMagnu, line 3502: checking for !Dynamic_enums.empty() would be somewhat cleaner, IMO

code/parse/sexp.h
```diff
@@ -27,109 +27,121 @@ struct ship_obj;
...
+	OPF_ANY_HUD_GAUGE,				// Goober5000 - both custom and builtin
+	OPF_WING_FLAG,					// Goober5000 - The name of a wing flag
+	OPF_ASTEROID_DEBRIS,			// MjnMixael - Debris types as defined in asteroids.tbl
+	OPF_WING_FORMATION,				// Goober5000 - as defined in ships.tbl
+
+	//Must always be at the end of the list
+	First_available_list_id
+};
+
+typedef struct enum_list {
```
BMagnu, line 139: I'd prefer if the type was called "lua_enum_list" or better "dynamic_sexp_enum_list", otherwise people in other parts of the code may get confused if this shows up in autocomplete.
Also, a slightly more modern way of declaring structs is 
```cpp
struct type_name {
    ...
};
```
This is moooostly equivalent*, but allows for forward declaration if it is ever needed (yes, yes, I know there are some technical differences that are very much not equivalent, but for the purposes of this...).

code/parse/sexp/sexp_lookup.cpp
```diff
@@ -28,6 +28,7 @@ struct global_state {
 	int next_free_operator_id = First_available_operator_id;
 	int next_free_category_id = First_available_category_id;
 	int next_free_subcategory_id = First_available_subcategory_id;
+	int next_free_list_id = First_available_list_id;
```
BMagnu, line 31: Similarly here, I'd prefer "enum_list" over "list" here, to help clue in people unfamiliar.

BMagnu (changes requested):
Looks solid.
A few teensy tiny style things, but good to go otherwise

code/parse/sexp.h
MjnMixael: Good idea. I've done that and everything else you suggested in the review.

BMagnu (approved):
Should be good to go

code/parse/sexp.h
```diff
@@ -27,109 +27,121 @@ struct ship_obj;
...
-#define OPF_ANIMATION_NAME 		96		// Lafiel
-#define OPF_CONTAINER_VALUE		97		// jg18 - Container data and map container keys
-#define OPF_DATA_OR_STR_CONTAINER	98	// jg18 - any data, or a container that is accessed via strings
-#define OPF_TRANSLATING_SUBSYSTEM	99	// Goober5000 - a translating subsystem
-#define OPF_ANY_HUD_GAUGE		100		// Goober5000 - both custom and builtin
-#define OPF_WING_FLAG			101		// Goober5000 - The name of a wing flag
-#define OPF_ASTEROID_DEBRIS		102		// MjnMixael - Debris types as defined in asteroids.tbl
-#define OPF_WING_FORMATION		103		// Goober5000 - as defined in ships.tbl
+enum : int {
+	OPF_NONE,						// argument cannot exist at this position if it's this
```
Goober5000, line 31: This should start at 1, like the old `OPF_NONE`.  The function `query_operator_argument_type` uses a return value of 0 to indicate that no `OPF_` values were found.

code/parse/sexp.h
```diff
@@ -27,109 +27,121 @@ struct ship_obj;
...
+	OPF_CONTAINER_VALUE,			// jg18 - Container data and map container keys
+	OPF_DATA_OR_STR_CONTAINER,		// jg18 - any data, or a container that is accessed via strings
+	OPF_TRANSLATING_SUBSYSTEM,		// Goober5000 - a translating subsystem
+	OPF_ANY_HUD_GAUGE,				// Goober5000 - both custom and builtin
+	OPF_WING_FLAG,					// Goober5000 - The name of a wing flag
+	OPF_ASTEROID_DEBRIS,			// MjnMixael - Debris types as defined in asteroids.tbl
+	OPF_WING_FORMATION,				// Goober5000 - as defined in ships.tbl
+
+	//Must always be at the end of the list
+	First_available_list_id
```
Goober5000, line 136: I would recommend calling this `First_available_enum_list_id` to match the other change.  Or perhaps `First_available_opf_id`.

code/parse/sexp/sexp_lookup.cpp
```diff
@@ -198,8 +199,13 @@ DynamicSEXP* get_dynamic_sexp(int operator_const)
 
 	return iter->second.get();
 }
-int get_category_of_subcategory(int subcategory_id)
+int increment_list_id()
```
Goober5000, line 202: This should be called `increment_enum_list_id`.

fred2/sexp_tree.cpp
```diff
@@ -3499,7 +3499,15 @@ int sexp_tree::query_default_argument_available(int op, int i)
...
-			Int3();
+			if (!Dynamic_enums.empty()) {
+				if ((type - First_available_list_id) < (int)Dynamic_enums.size()) {
+					return 1;
+				} else {
+					UNREACHABLE("Unhandled SEXP argument type!");
+				}
+			} else {
+				UNREACHABLE("Unhandled SEXP argument type!");
+			}
```
Goober5000, line 3510: The changes in sexp_tree.cpp and sexp_tree.h should be replicated in the qtfred counterparts of these files.

Goober5000 (changes requested):

code/parse/sexp.h
MjnMixael: Enum is not strictly accurate and opf isn't really understandable to anyone not familiar with the code, but ok whatever.

code/parse/sexp.h
MjnMixael: Good catch

fred2/sexp_tree.cpp
MjnMixael: I have no way to build or test qtfred and the code is slightly different. But based on reading it, I think what I did is accurate.

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -5128,6 +5128,25 @@ sexp_list_item *sexp_tree::get_container_multidim_modifiers(int con_data_node) c
 	return head.next;
 }
 
+sexp_list_item* sexp_tree::check_for_dynamic_sexp_enum(int opf)
+{
+	sexp_list_item head;
+
+	int item = opf - First_available_list_id;
```
Goober5000, line 5135: This should be `First_available_opf_id`

code/parse/sexp.h
```diff
@@ -28,7 +28,7 @@ struct ship_obj;
 
 // Operator argument formats (data types of an argument)
 enum : int {
-	UNUSED,							// argument types need to start at 1 instead of 0
+	OPF_UNUSED,						// argument types need to start at 1 instead of 0
```
Goober5000, line 31: Good catch here, you beat me to it

fred2/sexp_tree.cpp
Goober5000: A best-effort basis should be good enough for now.

code/parse/sexp.h
Goober5000: I'm open to other suggestions.  I like `First_available_opf_id` because opf_ is distinct enough to come up easily in a code search.

fred2/sexp_tree.cpp
```diff
@@ -3499,7 +3499,15 @@ int sexp_tree::query_default_argument_available(int op, int i)
...
-			Int3();
+			if (!Dynamic_enums.empty()) {
+				if ((type - First_available_opf_id) < (int)Dynamic_enums.size()) {
+					return 1;
+				} else {
+					UNREACHABLE("Unhandled SEXP argument type!");
+				}
+			} else {
+				UNREACHABLE("Unhandled SEXP argument type!");
+			}
```
Goober5000, line 3510: The qtfred file has a `query_default_argument_available` function (it has two of them, in fact) so this code should be copied to the default case, around line 1690.

qtfred/src/ui/widgets/sexp_tree.cpp
MjnMixael: done

fred2/sexp_tree.cpp
MjnMixael: and done

Goober5000 (approved):

PR #5012, opened 2022-12-15
fictionviewer localization

### LINKED ISSUE #1836: Translatable Fiction Viewer
I'm not entirely certain how this feature would work, but we need a way for the fiction viewer to work with translations... preferably a way where translations can live together in the same modpack. So we need some way to tell FSO... For English use file 'fictionstage2-eng.txt' and for German use file 'fictionstage3-ger.txt'.

MjnMixael:
Implements true fiction viewer localization. If language setting is not English then it will try to find the fiction viewer file with the language extension applied. If the file specified is `somefile.txt` then it will search for `somefile-gr.txt`, for example, if the language setting is German. If that file is not found then it will default to the base filename `somefile.txt`.

Also updates the scpui virtvar with the same code.

Fixes #1836 for real this time

### DIFF
--- a/code/scripting/api/objs/fictionviewer.cpp
+++ b/code/scripting/api/objs/fictionviewer.cpp
@@ -1,4 +1,5 @@
 #include "fictionviewer.h"
+#include "localization/localize.h"
 
 namespace scripting {
 namespace api {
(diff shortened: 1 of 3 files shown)


BMagnu (approved):

code/scripting/api/objs/fictionviewer.cpp
```diff
@@ -31,7 +32,9 @@ ADE_VIRTVAR(TextFile, l_FictionViewerStage, nullptr, "The text file of the stage
 		LuaError(L, "This property is read only.");
 	}
 
-	return ade_set_args(L, "s", stage.getStage()->story_filename);
+	SCP_string localized_filename = get_localized_fiction_filename(stage.getStage()->story_filename);
+
+	return ade_set_args(L, "s", localized_filename);
```
Goober5000, line 37: I think this line needs a `.c_str()`

Goober5000 (changes requested):

code/scripting/api/objs/fictionviewer.cpp
MjnMixael: Yes it does

Goober5000 (approved):

PR #5022, opened 2022-12-18
enhance handling of motion debris

MjnMixael:
Enhance motion debris handling in a number of ways that, hopefully, allow it to be more easily enhanced for new kinds of effects in the future. Stores the motion debris defines as a struct in a vector and allows parsing N types of motion debris. Cleans up stars.tbl parsing of motion debris and silently deprecates the old style parsing (IE: Just prints a thing to the log that the old style was used.) Adds error checking upon motion debris load. If it's not found, set motion debris to not render. Finally adds a new sexp `set-motion-debris` that allows arbitrarily changing the current motion debris type that's being rendered.

Oh and I fixed a slight bug with asteroid debris type error checking while I was in there to properly not warn when "<none>" was used in the sexp.


code/starfield/starfield.cpp
```diff
@@ -48,7 +47,7 @@ typedef struct {
 	vec3d pos;
 	int vclip;
 	float size;
-} motion_debris;
+} motion_debris_vclips;
```
Baezon, line 50: I would nominate this still be called `motion_debris` and the `SCP_vector<motion_debris>` below be called `motion_debris_info`, to vaguely match our naming distinction between static data and individual instance data in other places.

code/starfield/starfield.cpp
```diff
@@ -242,24 +228,39 @@ void stars_load_debris_vclips(debris_vclip *vclips)
...
+	}
+
+	// if not using custom the load the default type for the mission
+	if (fullneb && debris_name.empty()) { // if we're in nebula mode
+		debris_name = "nebula";
+	} else if (debris_name.empty()) {
+		debris_name = "default";
 	}
+
+	int debris_position = get_motion_debris_by_name(debris_name);
```
Baezon, line 250: It's a bit pedantic but I do prefer the more explicit `index` in these cases instead of the more vague `position`.

code/starfield/starfield.cpp
```diff
@@ -563,34 +569,104 @@ void parse_startbl(const char *filename)
...
-				else {
-					Warning(LOCATION, "Could not load normal motion debris '%s'; maximum of %d exceeded.", name, MAX_DEBRIS_VCLIPS);
+				if (count == MAX_MOTION_DEBRIS_BITMAPS) {
+					Motion_debris.push_back(this_debris);
+				} else {
+					error_display(0, "Not enough bitmaps defined for motion debris '%s'. Skipping!\n", this_debris.name.c_str());
 				}
 			}
 
 			CHECK_END();
```
Baezon, line 602: These check ends will require #end right? Aren't the new style motion debris definitions supposed to go up here too? Unless I'm mistaken, if I only wanted to use the new style definitions I would still need some extra random `#end`s before I could say them.

code/starfield/starfield.cpp
```diff
@@ -563,34 +569,104 @@ void parse_startbl(const char *filename)
...
+				in_check = true;
+
+				stuff_string(name, F_NAME, MAX_NAME_LEN);
+
+				// check that the name is unique
+				int check = get_motion_debris_by_name(name);
+
+				if (check >= 0) {
+					error_display(0, "Motion debris name %s is already used; All names must be unique. Skipping!\n", name);
+					skip_to_start_of_string_either("$Motion Debris Name:", "#end");
```
Baezon, line 645: This is modular so overwriting an existing one should be allowed right? I'm not asking you to implement this if so, as this isn't inherently bad to do, but I wanted to make sure it wasn't a mistake.

Baezon (changes requested):

code/starfield/starfield.cpp
MjnMixael: I did test that and the extra #ends are not required. I'm not exactly sure I follow how those CHECK_END() methods work, but they don't seem to get in the way here.

code/starfield/starfield.cpp
MjnMixael: I have a different struct called `motion_debris` and would prefer this one stay as `motion_debris_vclips` because it's more accurate. This particular struct is for each debris bitmap clip being rendered out of the max 300. `motion_debris` is the struct for what is now `Motion_debris_info`.

code/starfield/starfield.cpp
MjnMixael: Done

code/starfield/starfield.cpp
Baezon: I'm not going to push this too hard, but this struct isn't just the bitmap, it has everything specific an individual piece of... motion debris. It would be like calling `weapon` `weapon_bitmap` and `weapon_info` `weapon` or calling `ship` `ship_model` and `ship_info` `ship`. Not necessarily wrong per se, but there is a precedent for calling each individual thing which receives the graphics `thing`, and naming the more abstract class of them `thing_info`. Although I realize now each of these doesn't actually 'know' what type they are, but that's why I suggested the change.

MjnMixael: `Motion_debris` is now an array of `motion_debris_instance` struct objects. `Motion_debris_info` is now a vector of `motion_debris_type` struct objects.

code/starfield/starfield.cpp
MjnMixael: This now allows replacing an existing motion debris define. The downside for now, without adding full +nocreate support, is that any old data that wasn't replaced will stick around. For now this isn't a problem because you really should be replacing all 4 bitmaps with each define. But if additional data is added to the table parsing here, then it'd probably be wise to implement +nocreate and full error checking at that time.

code/starfield/starfield.cpp
Baezon: I'll take it.

Baezon (approved):

code/starfield/starfield.cpp
MjnMixael: actually, it's late and I'm tired and should have done this the first time around. Now it will correctly require all 4 bitmaps to be defined during new parsing and error otherwise. This way leftover data is not an issue. Replacing one or more bitmaps, if that's a desired feature, can be done later with +nocreate.

code/parse/sexp.cpp
```diff
@@ -3940,7 +3941,7 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, i
 					return SEXP_CHECK_TYPE_MISMATCH;
 				}
 
-				if (get_asteroid_position(CTEXT(node)) < 0) {
+				if (stricmp(CTEXT(node), SEXP_NONE_STRING) && (get_asteroid_position(CTEXT(node)) < 0)) {
```
Goober5000, line 3944: This could be submitted on its own as a separate bugfix PR that would not be subject to the code freeze.

code/starfield/starfield.cpp
```diff
@@ -151,25 +150,10 @@ typedef struct vDist {
...
-
-//XSTR:OFF
-debris_vclip Debris_vclips_normal[MAX_DEBRIS_VCLIPS] = { { -1, -1, "debris01" }, { -1, -1, "debris02" }, { -1, -1, "debris03" }, { -1, -1, "debris04" } };
-debris_vclip Debris_vclips_nebula[MAX_DEBRIS_VCLIPS] = { { -1, -1, "nebdeb01" }, { -1, -1, "nebdeb02" }, { -1, -1, "nebdeb03" }, { -1, -1, "nebdeb04" } };
-debris_vclip *Debris_vclips = Debris_vclips_normal;
-//XSTR:ON
-
-int stars_debris_loaded = 0;	// 0 = not loaded, 1 = normal vclips, 2 = nebula vclips
+SCP_vector<motion_debris_types> Motion_debris_info;
+motion_debris_bitmaps* Motion_debris_ptr;
```
Goober5000, line 156: This should be initialized to `nullptr` here.  There's only one other place in the code it is set to `nullptr`, and there are ways it could be skipped.

code/starfield/starfield.cpp
```diff
@@ -203,36 +184,41 @@ static int Mission_env_map = -1;
 static bool Env_cubemap_drawn = false;
 static bool Irr_cubemap_drawn = false;
 
-void stars_release_debris_vclips(debris_vclip *vclips)
+int get_motion_debris_by_name(SCP_string name) {
+	for (int i = 0; i < (int)Motion_debris_info.size(); i++) {
+		if (!stricmp(Motion_debris_info[i].name.c_str(), name.c_str())) {
```
Goober5000, line 189: This can use `SCP_string_lcase_equal_to` since both sides are SCP_string

code/starfield/starfield.h
```diff
@@ -148,7 +164,9 @@ void stars_preload_background_bitmap(const char *fname);
 
 void stars_set_nebula(bool activate, float range);
 
-void stars_load_debris(int fullneb = 0);
+int get_motion_debris_by_name(SCP_string name);
+
+void stars_load_debris(int fullneb = 0, SCP_string = "");
```
Goober5000, line 169: The parameters here ought to be `const SCP_string &` to avoid unnecessary copying of the string

fred2/sexp_tree.h
```diff
@@ -295,7 +295,8 @@ class sexp_tree : public CTreeCtrl
 	sexp_list_item *get_listing_opf_wing_flags();
 	sexp_list_item *get_listing_opf_team_colors();
 	sexp_list_item *get_listing_opf_nebula_patterns();
-	sexp_list_item* get_listing_opf_asteroid_debris();
+	sexp_list_item *get_listing_opf_asteroid_debris();
+	sexp_list_item *get_listing_opf_motion_debris();
```
Goober5000, line 299: The changes in sexp_tree.h and sexp_tree.cpp should be replicated in their qtfred counterparts.

Goober5000 (changes requested):
A few things.

code/starfield/starfield.cpp
MjnMixael: done

code/starfield/starfield.cpp
MjnMixael: k

code/starfield/starfield.h
MjnMixael: ok

fred2/sexp_tree.h
MjnMixael: done

Goober5000 (approved):

PR #5031, opened 2022-12-19
Fix issue with FSO-only scripts in FRED

### LINKED ISSUE #5026: Enabling scripts in FRED causes parsing errors
The errors that are caused when launching FRED maybe shouldn't be hit because parsing between FSO and FRED should be identical.

Use the attached tables. a_fred-sct.tbm should be the only script that's actually run. The mv_exp-sct-tbm is set to $Application: FS2_Open but an error on parse is hit forcing FRED to close.
[tables.zip](https://github.com/scp-fs2open/fs2open.github.com/files/10256055/tables.zip)

The-E:
when parsing a script table that includes hooks for a different application (i.e. FRED scripts in FSO or vice versa), only print something to the log instead of erroring out

this fixes #5026

### DIFF
--- a/code/scripting/global_hooks.cpp
+++ b/code/scripting/global_hooks.cpp
@@ -368,6 +368,33 @@ const std::shared_ptr<Hook<>> OnCheat = Hook<>::Factory("On Cheat",
 		{ "Cheat", "string", "The cheat code the user typed" },
 	});
 
+const std::shared_ptr<Hook<>> OnMissionAboutToEndHook = Hook<>::Factory("On Mission About To End",
+	"Called when a mission is about to end but has not run any mission-ending logic",
+	{});
+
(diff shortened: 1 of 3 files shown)


BMagnu (changes requested):
That's the wrong place to fix this.
In no circumstance should this if not cause an error.
The bug referenced _should not cause_ this. It means it found a nonexistent condition, or interpreted an action as a condition.
The underlying cause needs to be found. This check is, in theory, a parsing check and should _not_ be influenced by how the conditions evaluate

The-E:
Ahhhh, found the issue. "On Mission End" isn't parsed, because it technically doesn't exist within the FRED context (it's declared in `freespace.cpp`), which causes the parser to error out because it doesn't see an `#End` where it expects one.

BMagnu (approved):

PR #5033, opened 2022-12-20
Enhance Dynamic SEXPs with parent parameters

MjnMixael:
Enhances the dynamic sexp system to allow for "parent" parameters of individual parameters. This allows, for example, `OPF_SUBSYSTEM` to be used as a return type. You can specify a specific parameter index with `+Parent Paremeter Index` or if omitted, the parent parameter is assumed to be the previous parameter.


code/parse/sexp.cpp
```diff
@@ -899,6 +899,45 @@ sexp_ai_goal_link Sexp_ai_goal_links[] = {
...
 
+SCP_vector<dynamic_sexp_parameter_list> Dynamic_parameters;
+
+int get_dynamic_parameter_index(SCP_string op_name, int param)
+{
+	for (int i = 0; i < (int)Dynamic_parameters.size(); i++) {
+		if (!stricmp(Dynamic_parameters[i].operator_name.c_str(), op_name.c_str())) {
+
+			// Make sure there are actually parameters for this operator
+			if ((int)Dynamic_parameters[i].parameter_map.size() > 0) {
```
JohnAFernandez, line 910: No need to cast to int, if it returns a uint or size_t and you are checking if it's grater than zero.  Also, does parameter_map have an empty() method?  I'm unable to check.

code/parse/sexp.cpp
```diff
@@ -899,6 +899,45 @@ sexp_ai_goal_link Sexp_ai_goal_links[] = {
 
 SCP_vector<dynamic_sexp_enum_list> Dynamic_enums;
 
+SCP_vector<dynamic_sexp_parameter_list> Dynamic_parameters;
+
+int get_dynamic_parameter_index(SCP_string op_name, int param)
```
Goober5000, line 904: `op_name` should use `const SCP_string &`

code/parse/sexp.cpp
```diff
@@ -899,6 +899,45 @@ sexp_ai_goal_link Sexp_ai_goal_links[] = {
...
+{
+	for (int i = 0; i < (int)Dynamic_parameters.size(); i++) {
+		if (!stricmp(Dynamic_parameters[i].operator_name.c_str(), op_name.c_str())) {
```
Goober5000, line 907: Comparison should use `SCP_string_lcase_equal_to()()`

code/parse/sexp/LuaSEXP.cpp
```diff
@@ -595,6 +660,51 @@ void LuaSEXP::parseTable() {
 					_name.c_str()));
 			}
 
+		}
+		bool parent_required = false;
+		for (int i = 0; i < (int)parent_parameter_required.size(); i++) {
+			if (strcmp(type.first.c_str(), parent_parameter_required[i].c_str()) == 0) {
```
Goober5000, line 666: A case-sensitive comparison between SCP_strings can simply use ==

code/parse/sexp/LuaSEXP.cpp
```diff
@@ -595,6 +660,51 @@ void LuaSEXP::parseTable() {
...
+			int this_index = -1;
+			
+			if (optional_string("+Parent Parameter Index:")) {
+				stuff_int(&this_index);
+				this_index--;
+			}
+
+			if (this_index >= (param_index + 1)) {
+				error_display(1,
+					"Ship Paramater Index '%i' cannot be greater or equal to %i (the current parameter index)!\n",
```
Goober5000, line 683: `Parameter`

code/parse/sexp/LuaSEXP.cpp
```diff
@@ -595,6 +660,51 @@ void LuaSEXP::parseTable() {
...
+					this_index,
+					param_index + 1);
+			}
+			
+			std::pair<int, int> param_map(param_index, this_index);
+
+			// check if this operator already has an entry for dynamic parameters
+			int dyn_index = -1;
+			for (int i = 0; i < (int)Dynamic_parameters.size(); i++) {
+				if (!stricmp(Dynamic_parameters[i].operator_name.c_str(), _name.c_str())) {
```
Goober5000, line 693: `SCP_string_lcase_equal_to()()`

Goober5000 (changes requested):
A few polishing comments

code/parse/sexp.cpp
MjnMixael: done

Goober5000 (approved):

PR #5037, opened 2022-12-21
patch hud-lock iteration for objects that aren't ships

Goober5000:
Since `hud_lock_acquire_uncaged_target` assumes every item is a ship, it probably intended to iterate over `Ship_obj_list`, not `obj_used_list`.

### DIFF
--- a/code/hud/hudlock.cpp
+++ b/code/hud/hudlock.cpp
@@ -836,14 +836,15 @@ void hud_lock_acquire_uncaged_target(lock_info *current_lock, weapon_info *wip)
 	bool actively_locking = false;
 
 	for ( A = GET_FIRST(&obj_used_list); A !=END_OF_LIST(&obj_used_list); A = GET_NEXT(A) ) {
-		ship* sp = &Ships[A->instance];
+		if (A->flags[Object::Object_Flags::Should_be_dead])
+			continue;
 
 		if (!weapon_multilock_can_lock_on_target(Player_obj, A, wip, &dot))
 			continue;
 
 		bool in_range = weapon_secondary_world_pos_in_range(Player_obj, wip, &A->pos);
 
-		if ( Ship_info[sp->ship_info_index].is_big_or_huge() ) {
+		if ( A->type == OBJ_SHIP && Ship_info[Ships[A->instance].ship_info_index].is_big_or_huge() ) {
 			lock_info temp_lock;
 
 			temp_lock.obj = A;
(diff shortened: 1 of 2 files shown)


Goober5000:
This raises the question: how much has `hud_lock_acquire_uncaged_target` actually been used?  Surely someone would have run across a crash or memory access violation before now.

JohnAFernandez:
It looks like it is part of figuring out mutlilock hud targeting indicators.  I think it would probably be just a CTD on release, so when people hit it, it's probably not reported.

code/hud/hudlock.cpp
```diff
@@ -835,7 +835,10 @@ void hud_lock_acquire_uncaged_target(lock_info *current_lock, weapon_info *wip)
 	int least_num_locks = INT_MAX;
 	bool actively_locking = false;
 
-	for ( A = GET_FIRST(&obj_used_list); A !=END_OF_LIST(&obj_used_list); A = GET_NEXT(A) ) {
+	for ( auto so = GET_FIRST(&Ship_obj_list); so != END_OF_LIST(&Ship_obj_list); so = GET_NEXT(so) ) {
```
JohnAFernandez, line 838: Good fix.  Can we use either so or sp, since we don't need both?

JohnAFernandez (changes requested):

code/hud/hudlock.cpp
Goober5000: We can't, unfortunately.  The `so` variable is of the `ship_object` type, which is distinct from both the `object` and `ship` types.

What I could do is give it a `ship_obj` name to make that a little more clear.

Goober5000: Cancel that... `ship_obj` is the type itself, not `ship_object`.  And `so` is the iterator variable used pretty much everywhere.

Goober5000: Actually, there's a better fix here.

Baezon:
This doesn't crash or cause any directly harmful behavior because the first thing `weapon_multilock_can_lock_on_target` does is return false if `(target_objp->type != OBJ_SHIP && !checkWeapons)` (and it isn't checking weapons).

Goober5000:
Yeah, that's fortunate... although there's still the `ship *sp` assignment that could cause trouble.

I've modified the PR with some alternate safety checks.

JohnAFernandez (approved):

PR #5039, opened 2022-12-21
more robust object iteration

### LINKED ISSUE #5034: Vanished ships are still valid and accessible (for the frame they were vanished)
This is mostly an issue for scripts, as sexps appear to mostly use the registry which is properly updated with vanishing.

But once vanished, that ship handle is still valid, the ship still has hitpoints, is still returned by `mn.getShipList()`, and indeed there does not appear to be *any way* to determine that that ship has been vanished. This is, needless to say, highly unintuitive and confusing. 

Given that vanishing doesn't actually do all that terribly much in the moment but simply applies the `Should_be_dead` flag, this result is understandable after some investigation, but I'm not sure if `isValid()` should check for that flag, if the flag should be exposed to scripters, or the script system itself changed to not allow modification or access to such ships at all (that might seem extreme, since destroying the ship in other ways typically still has the ship be accessible because the ship object is truly still present, but given that 'vanish' is the closest thing to an 'instantly delete' button, I don't think its a terribly unreasonable option).

Goober5000:
When iterating through various object collections, check for the should-be-dead flag to avoid returning objects that are on the verge of being deleted.

Separately, this adds the `isExpiring` function to check for should-be-dead objects.

This PR also makes object iteration consistent across all object types, both to reduce code duplication and to ensure that the length and iteration are done using the same algorithm.  Furthermore, if the number of should-be-dead objects is nonzero, the Lua array count for that frame will be different from the internal array count for that frame.

Fixes #5034.

In draft status pending tests.

### DIFF
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -83,9 +83,46 @@ extern TIMESTAMP Game_shudder_time;
 extern int Game_shudder_total;
 extern float Game_shudder_intensity;
 
+constexpr int COUNT_OBJECTS = -1000;
+
+// Returns the indexth object subclass
+template <typename A>
+int object_subclass_at_index(A& object_subclass_array, size_t array_size, int index)
+{
+	int count = 0;
+
+	for (size_t i = 0; i < array_size; ++i)
+	{
+		int objnum = object_subclass_array[i].objnum;
+		if (objnum < 0 || objnum >= MAX_OBJECTS)
+			continue;
+		if (Objects[objnum].flags[Object::Object_Flags::Should_be_dead])
+			continue;
+
+		++count;
+
+		if (count == index)
+		{
+			return object_subclass_array[i].objnum;
+		}
+	}
+
+	if (index == COUNT_OBJECTS)
+		return count;
+	else
+		return -1;
+}
+
+// Counts the number of indexable objects
+template <typename A>
+int object_subclass_count(A& object_subclass_array, size_t array_size)
+{
+	return object_subclass_at_index(object_subclass_array, array_size, COUNT_OBJECTS);
+}
+
+
 namespace scripting {
 namespace api {
-
 
 //**********LIBRARY: Mission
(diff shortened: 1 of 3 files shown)


code/scripting/api/libs/mission.cpp
```diff
@@ -83,10 +83,40 @@ extern TIMESTAMP Game_shudder_time;
 extern int Game_shudder_total;
 extern float Game_shudder_intensity;
 
+constexpr int COUNT_OBJECTS = -1000;
+
+// Returns the indexth object subclass, or counts the number of indexable objects
+template <typename A>
+int object_subclass_at_index(A object_subclass_array, size_t array_size, int index = COUNT_OBJECTS)
```
Baezon, line 90: It's much too confusing to have two completely different kinds of interactions handled by the same function, even if their functionality would be largely similar, and the fact that you have to explicitly preface when you call this function to count stuff with `this function call counts the X that we can access` shows that its too confusing otherwise at a glance. I would definitely ask for a separate function `count_objects()` or something with a clearer name, that does nothing but call `object_subclass_at_index()` with `COUNT_OBJECTS` and to make its last parameter mandatory.

Goober5000: Ok, fair point

code/scripting/api/libs/mission.cpp
```diff
@@ -83,10 +83,47 @@ extern TIMESTAMP Game_shudder_time;
...
+			continue;
+		if (Objects[objnum].flags[Object::Object_Flags::Should_be_dead])
+			continue;
+
+		if (count == index)
+		{
+			return object_subclass_array[i].objnum;
+		}
+
+		++count;
```
MageKing17, line 107: Won't this mean that, when counting objects, the returned count will always be 1 higher than the actual number of objects?

Goober5000:
Ok, all scripting functions in this PR have been tested and verified to work.

wookieejedi (approved):
Everything looks good and tests work as expected

PR #5040, opened 2022-12-21
add should-be-dead checks for all ship iterations

### LINKED ISSUE #4990: Loading checkpoint on BP2-06 "Darkest Hour" on nightlies crashes the build. Works fine on FSO 22.2
[As found by user DAK on discord.](https://discord.com/channels/223511295431933953/255067124836335616/1051107703726030858)

Assert is "Only ships can have subsystems!"
[fs2log.txt](https://github.com/scp-fs2open/fs2open.github.com/files/10200337/fs2log.txt)

Goober5000:
When iterating through `Ship_obj_list`, check the `Should_be_dead` flag before using the object.  This fixes #4990 and adds defense-in-depth checks to guard against similar errors elsewhere.

I considered moving the `Ship_obj_list` removal from `obj_delete` to `ship_cleanup`, but decided it would be prudent to avoid fundamental changes to lifecycle management.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -435,6 +435,9 @@ void garbage_collect_path_points()
 	//	in pp_xlate, mark all used Path_point records
 	for ( so = GET_FIRST(&Ship_obj_list); so != END_OF_LIST(&Ship_obj_list); so = GET_NEXT(so) ) {
 		A = &Objects[so->objnum];
+		if (A->flags[Object::Object_Flags::Should_be_dead])
+			continue;
+
 		ship	*shipp = &Ships[A->instance];
 		if (shipp->ai_index != -1) {
 			ai_info	*aip = &Ai_info[shipp->ai_index];
@@ -466,6 +469,9 @@ void garbage_collect_path_points()
 	//	Now, using pp_xlate, fixup all aip->path_cur and aip->path_start indices
 	for ( so = GET_FIRST(&Ship_obj_list); so != END_OF_LIST(&Ship_obj_list); so = GET_NEXT(so) ) {
 		A = &Objects[so->objnum];
+		if (A->flags[Object::Object_Flags::Should_be_dead])
+			continue;
+
 		ship	*shipp = &Ships[A->instance];
 		if (shipp->ai_index != -1) {
 			ai_info	*aip = &Ai_info[shipp->ai_index];
@@ -1849,6 +1855,9 @@ int num_enemies_attacking(int objnum)
 
 	for ( so = GET_FIRST(&Ship_obj_list); so != END_OF_LIST(&Ship_obj_list); so = GET_NEXT(so) ) {
 		objp = &Objects[so->objnum];
+		if (objp->flags[Object::Object_Flags::Should_be_dead])
+			continue;
+
 		Assert(objp->instance != -1);
 		sp = &Ships[objp->instance];
 
@@ -1902,6 +1911,9 @@ float get_wing_lowest_max_speed(object *objp)
 
 	for ( so = GET_FIRST(&Ship_obj_list); so != END_OF_LIST(&Ship_obj_list); so = GET_NEXT(so) ) {
 		o = &Objects[so->objnum];
+		if (o->flags[Object::Object_Flags::Should_be_dead])
+			continue;
+
 		ship	*oshipp = &Ships[o->instance];
 		ai_info	*oaip = &Ai_info[oshipp->ai_index];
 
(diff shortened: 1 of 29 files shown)


The-E (approved):

code/ai/aicode.cpp
```diff
@@ -435,6 +435,9 @@ void garbage_collect_path_points()
 	//	in pp_xlate, mark all used Path_point records
 	for ( so = GET_FIRST(&Ship_obj_list); so != END_OF_LIST(&Ship_obj_list); so = GET_NEXT(so) ) {
 		A = &Objects[so->objnum];
+		if (A->flags[Object::Object_Flags::Should_be_dead])
+			continue;
+
```
JohnAFernandez, line 440: I took a quick look at ship cleanup, and these paths won't get cleaned up if a ship was vanished while exiting out of a bay.  I think this would barely happen, but I think this points to the fact that we really should vectorize or maybe refactor path points.  There's no reason to limit the number, and this function probably doesn't need to exist.

code/missionui/redalert.cpp
```diff
@@ -734,6 +734,8 @@ void red_alert_store_wingman_status()
 	// store status for all existing ships
 	for ( so = GET_FIRST(&Ship_obj_list); so != END_OF_LIST(&Ship_obj_list); so = GET_NEXT(so) ) {
 		ship_objp = &Objects[so->objnum];
+		if (ship_objp->flags[Object::Object_Flags::Should_be_dead])
+			continue;
```
JohnAFernandez, line 738: So, this one is a little tricky.  I think we need to check that vanished ships will be added to the Ships_exited container so that we don't accidentally leave out a ship.

Goober5000:
Putting this in draft to avoid inadvertent merge until I address Cyborg's comments.

code/missionui/redalert.cpp
Goober5000: I just looked at this and vanished ships are intentionally not added to Ships_exited.  What are you referring to by "accidentally leave out"?

code/ai/aicode.cpp
Goober5000: Can you elaborate on the connection between ship cleanup and path garbage collection?  The `garbage_collect_path_points` function is only called twice, both when paths are created, as opposed to when ships exit.  It also does not seem to depend in any way on whether a ship is vanished or not.

code/missionui/redalert.cpp
JohnAFernandez: So let's say the vanished ship was in the starting wings.  I'm not sure what the result would be if red alert processing then skipped that ship here because it was previously vanished or has the `Should_be_dead` flag.  Would that ship be missing in hud gauge on the next mission? Or would something else possibly worse happen.

Additionally, if vanished ships are never added to `Ships_exited` then it could be that vanished ships in starting wings are not handled correctly at all.  This is a very specific case.  If vanished ships are not handled correctly, anyway, then this change will not really change behavior except to also exclude ships that were vanished in the same frame that red alert is triggered. 

At some point someone may want to make sure that vanished ships in starting wings are handled correctly in red alert.

JohnAFernandez: But this change seems to reinforce the current behavior, at least, whether it's buggy or works.  So I don't think it's an issue here.  I just really think someone should make sure the two features work together as intended.

code/ai/aicode.cpp
JohnAFernandez: I mean it's just an issue with being able to create enough path points.  If, in the course of creating path points, a ship vanished this frame holds many path points, this check may keep enough of them from being freed for the ships being created.  I mean, it's an extreme edge case and not likely to happen.  I just think someone should make path points dynamic at some point and clean up those points when the ship departs.

Goober5000: I looked over the functions again and the path points should be properly cleaned up any time they are needed, whether a ship is vanished or not.  Vectorizing the list might be a good idea, but that's a separate issue from this PR.  In the meantime, 1000 is high enough that I don't think it's likely to be hit, in any case.

code/missionui/redalert.cpp
Goober5000: If a ship in the starting wings is vanished, it would be treated as if it wasn't included in the red-alert at all; that is, if the same ship appears in the next mission, it would be a "brand new" ship.

code/ai/aicode.cpp
JohnAFernandez: Agreed

JohnAFernandez:
I think this is good to go.

Goober5000:
I'll give this another once-over after Christmas, then merge it.

PR #5041, opened 2022-12-22
consolidation of tech rendering methods

### LINKED ISSUE #4628: Cleanup and general consolidation of tech rendering functions
From #4579 by @EatThePath:

> I think a general consolidation of tech rendering functions even beyond these new ones would probably be good. I intend to add some more configurability to them by the end of the year anyway.

MjnMixael:
Moves most of the renderTechModel and renderTechModel2 stuff into a single method in modelrender.cpp. Fixes #4628

### DIFF
--- a/code/model/modelrender.cpp
+++ b/code/model/modelrender.cpp
@@ -21,6 +21,7 @@
 #include "jumpnode/jumpnode.h"
 #include "lighting/lighting.h"
 #include "math/staticrand.h"
+#include "missionui/missionscreencommon.h"
 #include "mod_table/mod_table.h"
 #include "nebula/neb.h"
 #include "particle/particle.h"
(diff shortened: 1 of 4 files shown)


code/model/modelrender.cpp
```diff
@@ -3080,3 +3081,108 @@ void model_render_set_wireframe_color(color* clr)
...
+	if (!lighting || !model_lighting)
+		render_flags |= MR_NO_LIGHTING;
+
+	render_info.set_flags(render_flags);
+
+	if (is_ship) {
+		model_render_immediate(&render_info, model_num, model_instance, orient, &vmd_zero_vector);
+	} else {
+		model_render_immediate(&render_info, model_num, orient, &vmd_zero_vector);
+	}
```
BMagnu, line 3173: I know that the function didn't originally do this, but before the first render function, you must make sure that either, the following variable is set ``Shadow_override = true;``, or you need to recalculate the shadow map (see ``draw_model_rotating`` is missionscreencommon.cpp). Otherwise the model will just look broken when shadows are enabled.
I'd suggest just setting Shadow_override, and then after rendering you just need to make sure that it gets set back to what it was before

BMagnu (changes requested):
Ideally, I'd like a followup issue / PR, that merges these scripting techRenderFunctions with the _actual_ tech render functions (which should also be consolidated, given that there is techroom, ship select, etc.), but that can be deferred for now, I guess.
One small bug that we should fix while we're on it though:

code/model/modelrender.cpp
MjnMixael: The select and overhead rendering methods are quite a bit more complex. It'll take someone with more skill than I to consolidate those if that's really desired.

Setting shadow override is done.

BMagnu: Yeah, while we will eventually want that, it's not a priority, and this can be merged as-is then

BMagnu (approved):

PR #5042, opened 2022-12-22
A few missed items and fixes to the asteroid debris sexp

MjnMixael:


qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -5017,6 +5027,22 @@ sexp_list_item *sexp_tree::get_listing_opf_sexp_containers(ContainerType con_typ
 	return head.next;
 }
 
+sexp_list_item* sexp_tree::get_listing_opf_asteroid_debris()
```
Goober5000, line 5030: One minor quibble... this function in FRED2 is between `get_listing_opf_nebula_patterns` and `get_listing_opf_game_snds`.  Wherever the function might happen to be, it should be in the same relative location in FRED and qtFRED to make it easier to compare the two codebases.

(Keeping this function where it is and moving the FRED2 function is a potential solution.)

qtfred/src/ui/widgets/sexp_tree.h
```diff
@@ -349,6 +349,7 @@ class sexp_tree: public QTreeWidget {
 	sexp_list_item *get_listing_opf_functional_when_eval_type();
 	sexp_list_item *get_listing_opf_animation_name(int parent_node);
 	static sexp_list_item *get_listing_opf_sexp_containers(ContainerType con_type);
+	sexp_list_item* get_listing_opf_asteroid_debris();
```
Goober5000, line 352: Same comment for the header file.

Goober5000 (changes requested):

JohnAFernandez:
This is a reasonable request.  Having the FRED2 and qtFRED code bases be similar where they are the same makes qtFRED easier to work on.

Goober5000 (approved):

PR #5044, opened 2022-12-22
prevent access to goals outside the vector

MjnMixael:
Add some error checking here to prevent lua scripts from causing a CTD by access a goal outside the valid range.

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -798,7 +798,7 @@ ADE_INDEXER(l_Briefing_Goals,
 	"number Index",
 	"Array of goals",
 	"mission_goal",
-	"goal handle, or nil if index is invalid")
+	"goal handle, or invalid handle if index is invalid")
 {
 	int idx;
 	if (!ade_get_args(L, "*i", &idx))
@@ -808,7 +808,7 @@ ADE_INDEXER(l_Briefing_Goals,
 	idx--;
 
 	if ((idx < 0) || idx >= (int)Mission_goals.size())
-		return ADE_RETURN_NIL;
+		return ade_set_args(L, "o", l_Goals.Set(-1));
 
 	return ade_set_args(L, "o", l_Goals.Set(idx));
 }
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -798,11 +798,17 @@ ADE_INDEXER(l_Briefing_Goals,
 	"number Index",
 	"Array of goals",
 	"mission_goal",
-	"goal handle, or invalid handle if index is invalid")
+	"goal handle, or nil if index is invalid")
 {
 	int idx;
 	if (!ade_get_args(L, "*i", &idx))
 		return ade_set_error(L, "s", "");
+
+	// convert from lua index
+	idx--;
+
+	if ((idx < 0) || idx >= (int)Mission_goals.size())
+		return ADE_RETURN_NIL;
 
 	return ade_set_args(L, "o", l_Goals.Set(idx));
 }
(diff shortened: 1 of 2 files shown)


JohnAFernandez (approved):

code/scripting/api/libs/ui.cpp
```diff
@@ -798,12 +798,18 @@ ADE_INDEXER(l_Briefing_Goals,
...
 {
 	int idx;
 	if (!ade_get_args(L, "*i", &idx))
 		return ade_set_error(L, "s", "");
 
+	// convert from lua index
+	idx--;
+
+	if ((idx < 0) || idx >= (int)Mission_goals.size())
+		return ADE_RETURN_NIL;
```
Goober5000, line 811: Other indexers in the codebase return an invalid handle when the index is out of range, so for consistency's sake this one should do the same thing.  In the Personas indexer, for example, the code is:
```
if (idx < 0 || idx >= Num_personas)
		return ade_set_args(L, "o", l_Persona.Set(-1));
```
The -1 index represents an invalid persona.  Then there is a separate isValid function:
```
ADE_FUNC(isValid, l_Persona, NULL, "Detect if the handle is valid", "boolean", "true if valid, false otherwise")
{
	int idx = -1;

	if (!ade_get_args(L, "o", l_Persona.Get(&idx)))
		return ADE_RETURN_FALSE;

	return ade_set_args(L, "b", idx >= 0 && idx < Num_personas);
}
```

Goober5000 (changes requested):
Doing a spot check in the code, most of the indexers I've found return an invalid handle, not nil.  The only exceptions I see are `l_Ship_Pool` and `l_Weapon_Pool` which return numbers.

code/scripting/api/libs/ui.cpp
MjnMixael: You sure are a needy little one, aren't you. Done.

Goober5000: The `l_Goals` type still needs an `isValid` function.

MjnMixael: that's a whole lot of a code for something SCPUI _does not_ need and does not use.

Baezon: You and many others have cursed the inconsistency in other parts of the codebase, please let's not introduce more. An `isValid` function on lua objects of all types is completely reasonable and appropriate.

Please also bear in mind that features should not *only* relate to whatever use you have for them, and that other people may want to use this in other ways more than just SCPUI. Modders can use the FSO in all sorts of creative ways, and things should be kept as general purpose as possible, especially for something as basic as an `isValid` function.

MjnMixael: I fully expect SCP to go on a search to make sure every single indexer has this function.

Goober5000: Sometimes that's what happens.  The two goals here are a) consistency with the rest of the Lua API, and b) handling any possible scenario.  One of those scenarios is checking for a valid handle.

Goober5000 (approved):

Goober5000:
On a related topic, the various virtvars for the `l_Goals` object do not check whether the `current` variable is in range before accessing the array.  That should be patched at some point, either in this PR or another one.

MjnMixael:
> On a related topic, the various virtvars for the `l_Goals` object do not check whether the `current` variable is in range before accessing the vector. That should be patched at some point, either in this PR or another one.

With this in place, you can't get to the virtvars with an invalid index anyway. If you want to add that code, have at it. MVSC is -> that way.

Goober5000:
> With this in place, you can't get to the virtvars with an invalid index anyway.

You can, if you obtain an invalid `l_Goals` handle.

MjnMixael:
If only it returned nil instead.

JohnAFernandez:
So, not sure to merge this because of Poe's Law.  @MjnMixael are you referring to `code/scripting/api/objs/briefing.cpp` line 168 or are you suggesting that you are not going to implement making sure that `l_Goals` will be nil, but that someone else should?

PR #5047, opened 2022-12-24
Allow for arbitrary briefing icon scaling

MjnMixael:
Adds the ability to set a scaling factor for each briefing icon, allowing to scale them up or down.


fred2/briefingeditordlg.cpp
```diff
@@ -473,6 +475,9 @@ void briefing_editor_dlg::update_data(int update)
 			}
 			strcpy_s(ptr->icons[m_last_icon].closeup_label, buf);
 
+			if (m_icon_scale)
```
BMagnu, line 478: You sure this is correct?
m_icon_scale is set to -1 by default, which evaluates to true (as only 0 is false)

qtfred/ui/BriefingEditorDialog.ui
```diff
@@ -160,6 +160,13 @@
             </property>
            </widget>
           </item>
+		  <item row="0" column="0">
+		   <widget class="QLabel" name="iconScale">
+		    <property name="text">
+		     <string>Scale</string>
+		    </property>
+		   </widget>
+		  </item>
```
BMagnu, line 169: This seems insufficient to me.
Doesn't there need to be some code that actually updates bi->scale for qtFRED, and not just the input field?

BMagnu (changes requested):

fred2/briefingeditordlg.cpp
MjnMixael: It defaults to -1, but any new icon is actually set to 100 (percent). But either way, we shouldn't allow negative values because that's nonsensical and I'm not sure what the icon drawing function will actually do with it. Changed the check to > 0.

qtfred/ui/BriefingEditorDialog.ui
MjnMixael: Yes, this is correct. The briefing editor dialog for qtFRED currently has no code applied to it anywhere, just the window layout is done.

BMagnu: Fair

BMagnu (approved):

code/graphics/render.h
```diff
@@ -26,7 +26,7 @@ void gr_flash_alpha(int r, int g, int b, int a);
  * @param resize_mode The resize mode for translating the coordinated
  * @param mirror @c true to mirror the image
  */
-void gr_aabitmap(int x, int y, int resize_mode = GR_RESIZE_FULL, bool mirror = false);
```
Goober5000, line 29: Hang on... why is there a scale parameter added to `gr_aabitmap`?  In all the other places this is used, the scaling is done first, then the function is called.

Goober5000 (changes requested):

code/graphics/render.h
MjnMixael: Always enjoy when you decide you need the be the final arbiter of approved PRs.

No other usage of this actually scales the bitmap. They scale the screen coordinates which is different. Scale is added here to manipulate the image being rendered itself same as the mirror parameter. You don't mirror and then render. You don't scale and then render.

MjnMixael: In fact scale needs to be passed another level deeper still from here in order to actually render the bitmap correctly.

PR #5052, opened 2022-12-28
apply the HDR colorspace stuff _after_ mixing in the team color glows

The-E:


Baezon:
I can't speak to the changes themselves, but I confirmed this fixed an apparent discontinuity between regular glow maps and team color glow maps handling as discussed on the discord.

EatThePath (approved):
I'd considered changing this area in some of my recent passes but decided to leave well enough alone... since it turns out things weren't well enough after all, this is a fine change in my book.

PR #5053, opened 2022-12-28
add ability to disable grid rendering for each stage

MjnMixael:
Adds a checkbox for each briefing stage to optionally disable rendering the grid for that stage


fred2/fred.rc
```diff
@@ -683,8 +683,8 @@ CAPTION "About FRED2"
 FONT 8, "MS Sans Serif"
 BEGIN
     ICON            IDR_MAINFRAME,IDC_STATIC,10,17,20,20,SS_REALSIZEIMAGE
-    LTEXT           ABOUT_LINE_1,IDC_STATIC,39,8,184,10,SS_NOPREFIX
-    LTEXT           L"Copyright \x00A9 1999 Volition, Inc.  All Rights Reserved",IDC_STATIC,39,17,164,8
```
Goober5000, line 687: These two lines should be reverted

fred2/fred.rc
```diff
@@ -3046,7 +3047,7 @@ END
 
 STRINGTABLE
 BEGIN
-    IDR_MAINFRAME           MAINFRAME_CONTENTS
```
Goober5000, line 3049: This line should be reverted

Goober5000 (changes requested):
Looks good, aside from some auto-formatted .rc definitions that should be rolled back

Goober5000 (approved):

PR #5054, opened 2022-12-28
choose random splash image

MjnMixael:
As requested by @wookieejedi . This allows to specify a list of splash screen bitmaps in game_settings.tbl and then one will be chosen at random to display on game init. If the vector is empty then it defaults to pre_load/2_pre_load as is the current behavior.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -7456,7 +7456,7 @@ void game_title_screen_display()
 		Game_title_logo = bm_load(Game_logo_screen_fname[gr_screen.res]);
 
 		if (!Splash_screens.empty()) {
-			Game_title_bitmap = bm_load(Splash_screens[Random::next(0, (int)Splash_screens.size() - 1)].c_str());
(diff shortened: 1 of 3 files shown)


freespace2/freespace.cpp
```diff
@@ -7454,7 +7454,12 @@ void game_title_screen_display()
 	if(!condhook_override)
 	{
 		Game_title_logo = bm_load(Game_logo_screen_fname[gr_screen.res]);
-		Game_title_bitmap = bm_load(Game_title_screen_fname[gr_screen.res]);
+
+		if (!Splash_screens.empty()) {
+			Game_title_bitmap = bm_load(Splash_screens[Random::next(0, (int)Splash_screens.size() - 1)].c_str());
```
Goober5000, line 7459: The `Random::next` formula already excludes the upper bound, so the -1 is not needed.

Goober5000 (changes requested):

Goober5000 (approved):

PR #5055, opened 2022-12-28
The other great shader flag removal effort

The-E:
Currently, we are spending a lot of frametime swapping shaders. There are literally thousands of possible shader combinations that can occur during gameplay, and the way we differentiate shader feature sets means that at any point, we can run into a situation where a new shader has to be compiled, linked and switched to. This causes noticeable stutters, which are bad.

This attempts to solve this issue by turning some of the flags used by our main model shader into uniforms. 
As of this moment, this effort has resulted in a massive reduction in the number of unique shaders we need to run missions. In current master, Icarus requires 212 shaders to run, some of which need to be compiled at runtime; with the changes here, that number is reduced to 26 - only two of which are involved in actually rendering models. This eliminates the infamous warpin stutter completely.


wookieejedi:
Just reporting here that tests on both dedicated graphics card and Intel Integrated Graphics work as expected.

wookieejedi (approved):
Approving from successful testing and to ideally kick off more analysis.

PR #5057, opened 2022-12-29
Use `textify_scancode_universal()` in more places.

MageKing17:
Controls6 (PR #4133) changed `textify_scancode()` to be more l10n-aware, and added `textify_scancode_universal()` to continue serving the function of providing the values expected by scripts. Except there was one `textify_scancode()` call in the scripting system that wasn't changed, and it also wasn't used in FRED (for providing values for the `OPF_KEYPRESS` argument type, as used with `key-pressed`). This commit fixes both of these oversights, which *should* mean that all instances where a script or mission file needs to reference a keypress now use the same strings as before controls6.


wookieejedi (approved):
Good catch

PR #5058, opened 2022-12-29
add should-be-dead checks for missile and asteroid iterations

Goober5000:
Follow-up to #5050 and #5040, along the same lines.  Also uses `Missile_obj_list` and `Asteroid_obj_list` instead of `obj_used_list` where appropriate.

~~In draft status until #5050 is merged, as it depends on that.~~


wookieejedi (approved):
Solid follow-up to previous similar PRs and overall LGTM

PR #5060, opened 2022-12-29
SCPUI needs to be able run game_close_level

### LINKED ISSUE #4755: Mainhalls stop animating under particular SCPUI condition
Using SCPUI for any mission-loaded UI and then backing out of the mission causes the mainhall animations to stop animating until a mission is played and completed.

Specific steps to reproduce. 

1) Utilize the SCPUI mod found [here](https://www.hard-light.net/forums/index.php?topic=98473.0) and launch on a very recent nightly. It should launch out of the box into retail + scpui
2) Start a mission by clicking on the ready room. Any mission, doesn't matter.
3) Back out of the mission by pressing Esc
4) Mouseover any animated mainhall element. They will not animate, but the sound will be played and clicking still works.

I have tried to track this one down on and off for a month and have been unsuccessful. My guess is it has something to do with a librocket context not getting completely unloaded.

EDIT: This issue is a case-study in how keywords cause SCP to ignore tickets.

MjnMixael:
Fixes #4755 ~~by making sure the timer is not ever paused during a mainhall game state. We can't rely on UI-specific code to restart the timer after mission load if SCPUI is being used, so it should be done whenever the mainhall is entered probably. I didn't do a deep dive into the timer functions, so maybe there's a more appropriate fix elsewhere.~~

-by adding the ability to run `game_level_close()` from the scripting interface.

### DIFF
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1464,7 +1464,7 @@ ADE_FUNC(loadMission, l_Mission, "string missionName", "Loads a mission", "boole
 	return ADE_RETURN_TRUE;
 }
 
-ADE_FUNC(unloadMission, l_Mission, "boolean forceUnload", "Stops the current mission and unloads it. If forceUnload is true "
+ADE_FUNC(unloadMission, l_Mission, "[boolean forceUnload]", "Stops the current mission and unloads it. If forceUnload is true "
 	"then the mission unload logic will run regardless of if a mission is loaded or not. Use with caution.", nullptr, nullptr)
 {
 	bool forceUnload = false;
--- a/code/scripting/api/libs/mission.cpp
+++ b/code/scripting/api/libs/mission.cpp
@@ -1464,11 +1464,13 @@ ADE_FUNC(loadMission, l_Mission, "string missionName", "Loads a mission", "boole
 	return ADE_RETURN_TRUE;
 }
 
-ADE_FUNC(unloadMission, l_Mission, NULL, "Stops the current mission and unloads it", NULL, NULL)
+ADE_FUNC(unloadMission, l_Mission, "boolean forceUnload", "Stops the current mission and unloads it. If forceUnload is true "
+	"then the mission unload logic will run regardless of if a mission is loaded or not. Use with caution.", nullptr, nullptr)
 {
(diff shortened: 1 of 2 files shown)


Goober5000:
I'll take a look at this.  The timer code can be dicey, so I want to look at some of the possible code flow paths.

Goober5000:
#4755 is going to need a more comprehensive fix.  I've posted a comment there.

MjnMixael:
> #4755 is going to need a more comprehensive fix. I've posted a comment there.

Not that comprehensive. Just give SCPUI the ability to call the level close method. Tested and works.

Goober5000:
Hang on... I ran a search just now for `game_level_close()` and there is already a `mn.unloadMission()` function.  I missed this earlier because I searched for `freespace_stop_mission()`.

MjnMixael:
> mn.unloadMission()

That method does not work presumably because SCPUI doesn't pass the `if(Game_mode & GM_IN_MISSION)` test

MjnMixael:
can't wait to see why this fix won't work either

code/scripting/api/libs/mission.cpp
```diff
@@ -1464,11 +1464,13 @@ ADE_FUNC(loadMission, l_Mission, "string missionName", "Loads a mission", "boole
 	return ADE_RETURN_TRUE;
 }
 
-ADE_FUNC(unloadMission, l_Mission, NULL, "Stops the current mission and unloads it", NULL, NULL)
+ADE_FUNC(unloadMission, l_Mission, "boolean forceUnload", "Stops the current mission and unloads it. If forceUnload is true "
```
Goober5000, line 1467:
```suggestion
ADE_FUNC(unloadMission, l_Mission, "[boolean forceUnload]", "Stops the current mission and unloads it. If forceUnload is true "
```

Goober5000 (changes requested):
Looks good.  Only thing is that the documentation should reflect that the parameter is optional.

code/scripting/api/libs/mission.cpp
MjnMixael: Done

Goober5000 (approved):

JohnAFernandez (approved):

PR #5061, opened 2022-12-30
Add `turret-fired-since` SEXP

### LINKED ISSUE #3763: FR: Sexp turret-fired-since
I'm honestly surprised this doesn't exist. It should work just like primary-fired-since or secondary-fired-since. Becomes true if the specified turret has fired in the last N milliseconds.

Baezon:
Closes #3763

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -16751,7 +16751,6 @@ void sexp_set_mission_mood (int node)
 int sexp_turret_fired_delay(int node)
 {
 	ship* shipp;
-	int requested_bank, delay, last_fired = -1;
 	bool is_nan, is_nan_forever;
 
 	auto ship_entry = eval_ship(node);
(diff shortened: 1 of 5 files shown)


code/parse/sexp.cpp
```diff
@@ -286,6 +286,7 @@ SCP_vector<sexp_oper> Operators = {
 	{ "is-primary-selected",			OP_IS_PRIMARY_SELECTED,					2,	2,			SEXP_BOOLEAN_OPERATOR,	},
 	{ "is-secondary-selected",			OP_IS_SECONDARY_SELECTED,				2,	2,			SEXP_BOOLEAN_OPERATOR,	},
 	{ "primary-fired-since",			OP_PRIMARY_FIRED_SINCE,					3,	3,			SEXP_INTEGER_OPERATOR,	},	// Karajorma
+	{ "turret-fired-since",				OP_TURRET_FIRED_SINCE,					3,	3,			SEXP_INTEGER_OPERATOR, },	// Asteroth
```
Goober5000, line 289: I'd put this after `primary-fired-since` and `secondary-fired-since`, vs. between them

Goober5000 (changes requested):
This looks good, although since we are slowly migrating over to the type-safe timestamps, I'd like this new field to be the `TIMESTAMP` type

Goober5000 (approved):
Looks good.

PR #5066, opened 2022-12-31
use correct joystick id type (instance -> device)

### LINKED ISSUE #4793: Joystick ID and Joystick device ID are mixed up.
Need to double check with SDL to see what the definitions are. They seem to be in reverse order of each other in terms of which was connected to the PC first / order of which controller was enumerated.

This bug report however is focused on the two's usage within the joy-sdl ident code.  Some places the ID is used, other places the device ID is used (with the same variable).

notimaginative:
fixes #4793

### DIFF
--- a/code/io/joy-sdl.cpp
+++ b/code/io/joy-sdl.cpp
@@ -207,7 +207,7 @@ json_t* joystick_serializer(Joystick* joystick)
 		return json_null();
 	}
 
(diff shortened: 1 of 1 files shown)


JohnAFernandez (approved):
I'm on the side of something like this getting included on 23.0.  We're not that deep in the release cycle, so we'll have time for feedback on this.

wookieejedi:
Agreed.

PR #5071, opened 2023-01-02
is-ship-emp-active sexp

### LINKED ISSUE #2856: FR: a SEXP to check if the EMP effect is applied to a ship
As it say in the headline: I would like to request a SEXP that allows to check if the EMP effect is applied to a ship, please. 
An option to check for the time remaining in the effect would nice, too.

This is so you can delay mission critical messages during combat that features EMP weapons or the gameplay effects that make use of the EMP weapons, e.g. switching the IFF of sentry gun each time it is hit with an EMP effect.

MjnMixael:
Adds sexp to check if a ship or list of ships are currently experiencing EMP effects. Fixes #2856


code/parse/sexp.cpp
```diff
@@ -16372,6 +16373,48 @@ int sexp_are_wing_flags_set(int node)
...
+
+		// see if the ship has already exited the mission (either through departure or destruction)
+		if (ship_entry->exited_index >= 0) {
+			return SEXP_KNOWN_FALSE;
+
+		}
+		// ship is in mission
+		else if (ship_entry->shipp) {
+			if (ship_entry->shipp->emp_intensity > 0.0f) {
+				num_known++;
```
BMagnu, line 16402: No need to count here.
Really, as soon as you have any ship that is _not_ counting up here, you can immediately return false as there is at least one ship missing

BMagnu (changes requested):

code/parse/sexp.cpp
MjnMixael: Yeah you're right. I copied that from a similar sexp, but yeah, removed.

BMagnu (approved):

Issue #5005, opened 2022-12-14
Nightlies break BP2-08 The Intervention


MoerasGrizzly:
[As reportred by Muninn_Crow on discord.](https://discord.com/channels/223511295431933953/255067124836335616/1052300088669638676)

In War In Heaven -Act 2, there is a test flight with lots of dialogue options. Alas, in 'pressing the issue' of flight morale, the options for the following path stop advancing. (Tested twice)
Dialogue path: 4. Direct > 3. Personal Strengths > 1. That's the spirit > 1. Discuss squadron morale > 2. Persist > 2. Call her out > (1. Demand better performance;  2. Critique specific behavior) Unable to continue.

Just tested the mission Intervention in the stable build. Works perfectly fine. :/

Goober5000:
I've determined that this is somehow related to #4954 or #4633, but that part of the logic has already been exhaustively tested in everything from FSPort to Scroll to the other Blue Planet mission.  Plus, this doesn't happen for all conversation events, just this particular one.  So this is puzzling.

Goober5000:
I have a better idea of what's going on now but I need to do some more testing to be sure.

Goober5000:
Okay, after extensive testing and looking at this from multiple angles, I believe the event chaining behavior is is working as designed.  As described in commit 06fc8c73a5b892513fe06eab80d67f2a45a075de, the event timestamp field was originally a Schroedinger's hybrid of two data types, that nevertheless worked correctly for almost all cases due to a cascade of coincidences.  BP2-08 contains a mission bug that "still worked" due to a bug in the original event timestamp implementation.  When the timestamp implementation was fixed, the mission bug caused the mission to break.

Events that are chained to repeating events may only proceed once the repeating event has completed -- or, if the chain delay is within the rotation interval, once the delay has elapsed.  In other words, when the event fires, a stopwatch starts, but the stopwatch resets every time the event repeats.  If the chain delay is within the stopwatch interval, then it will fire; if the chain delay is outside the interval, then it cannot fire because the stopwatch resets before the interval elapses.  But once the repetition cycle is complete, the stopwatch will no longer reset.

This is the way the design is intended to work anyway; the original implementation's bug caused the stopwatch to not reset after the first interval.

[...]

PR #5073, opened 2023-01-03
restore retail behavior of chaining to repeating events

Goober5000:
To avoid changing behavior of existing mods, simulate the retail behavior unless the bugfix is explicitly activated.  The buggy retail behavior is to evaluate a chained event, not skip it, if the event is repeating, regardless of whether the repeat timestamp has elapsed.

"Fixes" #5005.


wookieejedi (approved):
Good digging

PR #5076, opened 2023-01-04
Warn dogfight weapons

### LINKED ISSUE #3100: FRED Should prevent dogfight missions that don't have dogfight weapons allowed.
Migrated Mantis issue 2107.

Load up this mission in multi and you will find you cannot play it because there are no allowed weapons. Dogfight missions need dogfight weapons.

Notably the crash reported in the original Mantis ticket is no longer an issue. This is now just something that FRED should warn the FREDer about.

[Untitled2.zip](https://github.com/scp-fs2open/fs2open.github.com/files/5790337/Untitled2.zip)

MjnMixael:
Fixes #3100 and Mantis 2107 by doing a quick check for dogfight weapons for each ship in the loadout if the mission is a dogfight mission. If not at least one weapon for each ship type then show a warning.

### DIFF
--- a/fred2/missionsave.cpp
+++ b/fred2/missionsave.cpp
@@ -3955,13 +3955,16 @@ int CFred_mission_save::save_players()
 				fout("\t@%s\t", Sexp_variables[var_idx].variable_name);
 			} else {
 				fout("\t\"%s\"\t", Ship_info[Team_data[i].ship_list[j]].name);
-				// Check the weapons pool for at least one dogfight weapon for this ship type
-				for (int wepCount = 0; wepCount <= Team_data[i].num_weapon_choices; wepCount++) {
-					if (Ship_info[Team_data[i].ship_list[j]].allowed_weapons[Team_data[i].weaponry_pool[wepCount]] & DOGFIGHT_WEAPON) {
-						num_dogfight_weapons++;
-						break;
-					}
 
+				// Check the weapons pool for at least one dogfight weapon for this ship type
+				if (IS_MISSION_MULTI_DOGFIGHT) {
+					for (int wepCount = 0; wepCount <= Team_data[i].num_weapon_choices; wepCount++) {
+						if (Ship_info[Team_data[i].ship_list[j]].allowed_weapons[Team_data[i].weaponry_pool[wepCount]] &
+							DOGFIGHT_WEAPON) {
+							num_dogfight_weapons++;
+							break;
+						}
+					}
 				}
 			}
 
--- a/fred2/missionsave.cpp
+++ b/fred2/missionsave.cpp
@@ -3985,7 +3985,7 @@ int CFred_mission_save::save_players()
 		// make sure we have at least one dogfight weapon for each ship type in a dogfight mission
 		if (IS_MISSION_MULTI_DOGFIGHT && (num_dogfight_weapons != Team_data[i].num_ship_choices)) {
 			MessageBox(nullptr,
-				"Warning: This mission is a dogfight mission but no dogfight weapons are available in the loadout!", "No dogfight weapons",
+				"Warning: This mission is a dogfight mission but no dogfight weapons are available for at least one ship in the loadout!", "No dogfight weapons",
 				MB_OK);
 		}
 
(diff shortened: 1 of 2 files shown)


qtfred/src/mission/missionsave.cpp
```diff
@@ -3757,6 +3758,17 @@ int CFred_mission_save::save_players()
...
+				// Check the weapons pool for at least one dogfight weapon for this ship type
+				if (IS_MISSION_MULTI_DOGFIGHT) {
+					for (int wepCount = 0; wepCount <= Team_data[i].num_weapon_choices; wepCount++) {
+						if (Ship_info[Team_data[i].ship_list[j]].allowed_weapons[Team_data[i].weaponry_pool[wepCount]] &
+							DOGFIGHT_WEAPON) {
+							num_dogfight_weapons++;
+							break;
+						}
+					}
+				}
```
JohnAFernandez, line 3771: Based on how I'm reading it, this new code should be outside the if that it is nested in.  I'm reading that we're only checking for a class if someone specified a concrete count instead of a variable based count.  Shouldn't it be checked regardless?

qtfred/src/mission/missionsave.cpp
```diff
@@ -3773,6 +3785,13 @@ int CFred_mission_save::save_players()
 
 		fout(")");
 
+		// make sure we have at least one dogfight weapon for each ship type in a dogfight mission
+		if (IS_MISSION_MULTI_DOGFIGHT && (num_dogfight_weapons != Team_data[i].num_ship_choices)) {
+			MessageBox(nullptr,
+				"Warning: This mission is a dogfight mission but no dogfight weapons are available for at least one ship in the loadout!", "No dogfight weapons",
+				MB_OK);
+		}
```
JohnAFernandez, line 3793: Suggestion if you have the skill to pull it off: record the ships that don't have dogfight weapons available and then add them to this warning.  An `SCP_string` would make the process fairly easy.

JohnAFernandez (changes requested):

qtfred/src/mission/missionsave.cpp
MjnMixael: The number of ships this could include is technically unlimited, making the error message potentially incredibly long. Otherwise I'd do it.

Instead it's probably better to print each ship to the log and include that instruction in the message itself.

JohnAFernandez: Ah, smart.

qtfred/src/mission/missionsave.cpp
MjnMixael: Yes you're right. Done

qtfred/src/mission/missionsave.cpp
MjnMixael: and done

fred2/missionsave.cpp
```diff
@@ -3966,10 +3969,34 @@ int CFred_mission_save::save_players()
 			} else {
 				fout("%d\n", Team_data[i].ship_count[j]);
 			}
+
+			// Check the weapons pool for at least one dogfight weapon for this ship type
+			if (IS_MISSION_MULTI_DOGFIGHT) {
+				for (int wepCount = 0; wepCount <= Team_data[i].num_weapon_choices; wepCount++) {
```
Goober5000, line 3975: I believe this `<=` should be `<`

qtfred/src/mission/missionsave.cpp
```diff
@@ -3769,10 +3772,34 @@ int CFred_mission_save::save_players()
 			} else {
 				fout("%d\n", Team_data[i].ship_count[j]);
 			}
+
+			// Check the weapons pool for at least one dogfight weapon for this ship type
+			if (IS_MISSION_MULTI_DOGFIGHT) {
+				for (int wepCount = 0; wepCount <= Team_data[i].num_weapon_choices; wepCount++) {
```
Goober5000, line 3778: Same here

Goober5000 (changes requested):

fred2/missionsave.cpp
MjnMixael: done

qtfred/src/mission/missionsave.cpp
MjnMixael: done

Goober5000 (approved):

JohnAFernandez (approved):

PR #5077, opened 2023-01-04
Allow gameplay_profiles.tbl to be an alias of ai_profiles.tbl

### LINKED ISSUE #3456: Proposal: Rename Ai profiles.tbl
The name is misleading and continues to trip up even people who have worked with the engine for years. I propose that a new name should be selected and treated as synonymous by the engine ~~similar to game_settings.tbl and mod.tbl~~. Documentation would be switched to the new name as much as possible. Changing over the internal variable and type names would probably be too much work to be done comprehensively, and too noisy, so I imagine that'd be left alone, but at least for the end users the purpose can be more clear that way.

I would name it mission_settings or perhaps gameplay_profiles or something, and there should be a new tbm suffix too of course.

MjnMixael:
This PR fixes #3456 by allowing `gameplay_profiles.tbl` to be an alias of `ai_profiles.tbl` . There is precedent for this with `intel.tbl` <-> `species.tbl` and this works PR exactly the same way.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -652,12 +652,12 @@ void ai_profiles_init()
 	// init retail entry first
 	parse_ai_profiles_tbl(NULL);
 
-	// Allow gameplay_profiles.tbl to be a alias of ai_profiles.tbl, but only load one or the other.
-	// gameplay_profiles.tbl would be newer so assume intended, but print to the log to be sure - Mjn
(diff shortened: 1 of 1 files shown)


MjnMixael:
> I recommend `mission_profiles.tbl` since this is per-mission. `gameplay_profiles.tbl` sounds very similar to `game_settings.tbl` which is mod-wide, so this could lead to confusion.

done

Goober5000 (approved):

PR #5078, opened 2023-01-04
Disallow overly large input in FRED Orientation Editor

### LINKED ISSUE #4564: FRED: protect ship coordinates from overflow
In FRED2, you can input an arbitrarily large number into an object's coordinates, and exceed what I presume is INT_MAX value for them. I also managed to cause an object's orientation values to become NaN if I point the object at something that is influenced by the above.

MjnMixael:
Fixes #4564 by limiting input to `INT_MAX` whenever we're converting the string to a float. qtFRED limits input while typing, so this fix isn't needed there.

### DIFF
--- a/fred2/orienteditor.cpp
+++ b/fred2/orienteditor.cpp
@@ -411,6 +411,11 @@ float orient_editor::convert(const CString &str) const
 		val = (float)INT_MAX;
 	}
 
+	// disallow input smaller than INT_MIN
+	if (val < (float)INT_MIN) {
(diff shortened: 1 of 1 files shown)


Goober5000 (changes requested):
Good idea.  This should check `INT_MIN` as well.

MjnMixael:
> Good idea. This should check `INT_MIN` as well.

done

Goober5000 (approved):

PR #5079, opened 2023-01-04
Make sexp node comments more visible at a glance

MjnMixael:
An idea came up during a BtA team FREDing session to make node comments in the event editor more visible by including them in the help text box. This PR implements the feature. For root "Event" nodes it will display any comments instead of being blank. For Messages and Sexps any comments will be appended to the message text or sexp help.

qtFRED does not yet have support for node comments, so the base code is added where it should be, but commented out until Goober finishes porting the feature.

### DIFF
--- a/fred2/sexp_tree.cpp
+++ b/fred2/sexp_tree.cpp
@@ -4791,7 +4791,9 @@ void sexp_tree::update_help(HTREEITEM h)
 	SCP_string nodeComment;
 
 	if (thisIndex >= 0) {
-		nodeComment = "Node Comments:\r\n   " + Event_annotations[thisIndex].comment;
+		if (!Event_annotations[thisIndex].comment.empty()) {
+			nodeComment = "Node Comments:\r\n   " + Event_annotations[thisIndex].comment;
+		}
 	} else {
 		nodeComment = "";
 	}
@@ -4802,6 +4804,11 @@ void sexp_tree::update_help(HTREEITEM h)
 			mini_help_box->SetWindowText("");
 		return;
 	}
+
+	// Now that we're done with top level nodes we can add the empty lines because
+	// everything else below is supposed to have help text
+	if (!nodeComment.empty())
+		nodeComment.insert(0, "\r\n\r\n");
 
 	if (SEXPT_TYPE(tree_nodes[i].type) == SEXPT_OPERATOR)
 	{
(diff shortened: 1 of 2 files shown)


fred2/sexp_tree.cpp
```diff
@@ -4904,7 +4912,7 @@ void sexp_tree::update_help(HTREEITEM h)
 			if (query_operator_argument_type(index, c) == OPF_MESSAGE) {
 				for (j=0; j<Num_messages; j++)
 					if (!stricmp(Messages[j].name, tree_nodes[i].text)) {
-						text.Format("Message Text:\r\n%s", Messages[j].message);
+						text.Format("Message Text:\r\n%s\r\n\r\n%s", Messages[j].message, nodeComment.c_str());
```
Goober5000, line 4915: mmmm... these two blank lines will always be appended.  I would find a way to only append them if `nodeComment` exists, perhaps by adding another `%s` just for them.  For messages that are almost long enough to fill the text box, this could cause them to scroll when they would otherwise fit.

fred2/sexp_tree.cpp
```diff
@@ -4915,11 +4923,14 @@ void sexp_tree::update_help(HTREEITEM h)
...
 
 	code = get_operator_const(tree_nodes[i].text);
-	str = help(code);
-	if (!str)
-		str = "No help available";
+	auto str = help(code);
+	if (!str) {
+		text.Format("No help available\r\n\r\n%s", nodeComment.c_str());
+	} else {
+		text.Format("%s\r\n\r\n%s", str, nodeComment.c_str());
```
Goober5000, line 4930: Same comments with the two blank lines.  If `nodeComment` will always be preceded by two blank lines, then they could be added up on line 4794.

fred2/sexp_tree.cpp
```diff
@@ -4915,11 +4923,14 @@ void sexp_tree::update_help(HTREEITEM h)
...
+	}
 
-	help_box->SetWindowText(str);
+	help_box->SetWindowText(text);
```
Goober5000, line 4933: Since the function argument is now a `CString`, it should have a `(LPCSTR)` cast.  This is usually automatic, but may not work with all build configurations.

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -2749,8 +2749,20 @@ void sexp_tree::update_help(QTreeWidgetItem* h) {
 		}
 	}
 
+	// Node comments are not yet implemented in qtFRED, so just adding some base code here
+	// that can be used when the feature is completed - Mjn
+
+	//int thisIndex = event_annotation_lookup(h);
+	SCP_string nodeComment;
+
+	//if (thisIndex >= 0) {
```
Goober5000, line 2758: Comments from FRED's sexp_tree.cpp should also apply to qtFRED's sexp_tree.cpp where applicable.

Goober5000 (changes requested):

qtfred/src/ui/widgets/sexp_tree.cpp
MjnMixael: done

Goober5000 (approved):
Looks good. 👍

PR #5082, opened 2023-01-05
This case was never ported to qtFRED

MjnMixael:


JohnAFernandez (approved):

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -1346,6 +1346,7 @@ int sexp_tree::get_default_value(sexp_list_item* item, char* text_buf, int op, i
 	case OPF_SUBSYSTEM:
 	case OPF_AWACS_SUBSYSTEM:
 	case OPF_ROTATING_SUBSYSTEM:
+	case OPF_TRANSLATING_SUBSYTEM:
```
tomimaki, line 1349: Subsytem is typo or intended?

MjnMixael: Typo. Good catch.

Goober5000 (changes requested):
Great catch.

Looking at `get_listing_opf_subsystem`, there are a few additional edits, related to `OPS_TRANSLATE`, that are needed in order to fully sync the two files.

Goober5000 (changes requested):
Almost done.  The `OPS_TRANSLATE` definition needs to be added to the `// specific types of subsystems we're looking for` list, in the same order as in FRED's sexp_tree.cpp.

MjnMixael:
> Almost done. The `OPS_TRANSLATE` definition needs to be added to the `// specific types of subsystems we're looking for` list, in the same order as in FRED's sexp_tree.cpp.

yeah ok. I'm kinda done cleaning up the qtFRED mess. Feel free to edit this PR as you like. I have neither the time or motivation to keep chasing this one in circles and I was told recently that's all the reason I need not to do stuff.

MjnMixael:
OK fine. Last one because I found another thing missing.

Goober5000 (approved):

PR #5084, opened 2023-01-06
Fix some issues with damage application

Baezon:
Poor Nuke was clearly aware of the risk of accidentally applying player damage difficulty scaling more than once, but alas, that's what's happened.

The code is really pointlessly confusing (which may have contributed to the bug being written), so I've tried to give it a good makeover and comment more. 

The main issue is that shockwaves and area blasts will erroneously always report as having hit the shield, usually not a big deal, it checks the shield is empty, so in essence all of the damage overflows to the hull anyway. However, damage difficulty scaling *is* applied to the damage before being passed to the hull damage calculation, where it gets applied *again*, squaring the scale factor and possibly greatly changing the damage in effect. So, unless the backwards compatible flag is checked, remove difficulty scaling from the 'overflow' damage.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -772,7 +772,8 @@ void ai_profile_t::reset()
 	if (mod_supports_version(3, 6, 10)) {
 		flags.set(AI::Profile_Flags::Reset_last_hit_target_time_for_player_hits);
 	}
-	if (mod_supports_version(3, 6, 14)) {
+	// backwards compatible flag for a bug accidentally introduced during this time
+	if (mod_supports_version(3, 6, 14) && !mod_supports_version(23, 0, 0)) {
 		flags.set(AI::Profile_Flags::Carry_shield_difficulty_scaling_bug);
 	}
 	if (mod_supports_version(21, 4, 0)) {
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -610,6 +610,8 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$use adjusted AI class autoscale:", AI::Profile_Flags::Adjusted_AI_class_autoscale);
 
+				set_flag(profile, "$carry shield difficulty scaling bug:", AI::Profile_Flags::Carry_shield_difficulty_scaling_bug);
+
 
 				// if we've been through once already and are at the same place, force a move
 				if (saved_Mp && (saved_Mp == Mp))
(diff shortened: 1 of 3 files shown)


code/ship/shiphit.cpp
```diff
@@ -2358,103 +2357,64 @@ static void ship_do_damage(object *ship_objp, object *other_obj, vec3d *hitpos,
...
+				shield_damage *= difficulty_scale_factor;
 			}
 
-			if (piercing_pct > 0.0f) {
-				damage += (piercing_pct * pre_shield);
-				subsystem_damage += (piercing_pct * pre_shield_ss);
-			}
-		}
+			float shield_factor = 1.0f;
+			if (weapon_info_index > 0 )
```
JohnAFernandez, line 2385: Greater than zero?

JohnAFernandez (approved):

code/ai/ai_profiles.cpp
```diff
@@ -770,6 +772,9 @@ void ai_profile_t::reset()
 	if (mod_supports_version(3, 6, 10)) {
 		flags.set(AI::Profile_Flags::Reset_last_hit_target_time_for_player_hits);
 	}
+	if (mod_supports_version(3, 6, 14)) {
+		flags.set(AI::Profile_Flags::Carry_shield_difficulty_scaling_bug);
+	}
```
MageKing17, line 777: Shouldn't this not apply to mods targeting 23.0?

Baezon: This bug has existed since 3.6.14 and has been affecting any mods unintentionally balanced with it in mind, so unless it is applied retroactively, it could affect their behavior. This is *really* retroactive application, but it's as close as we can get.

MageKing17: No, I perfectly understand why it should apply to mods targeting older versions; I just don't understand why this conditional is left open so that mods targeting 23.0 default to buggy behavior.

Baezon: Oh! I misunderstood, why does not *not* apply to 23.0. To be honest, I don't know why I hadn't even considered a slightly more complicated check, I'm glad you mentioned it because I was somewhat upset to 'introduce' an opt-out bug to new verisons.

Baezon:
This might be able to go in now, but thinking about it, I'm too worried about breakage of mods which don't use a targeted version, in which case the best we can do is not break during a release cycle.

code/ai/ai_profiles.cpp
```diff
@@ -770,6 +772,10 @@ void ai_profile_t::reset()
 	if (mod_supports_version(3, 6, 10)) {
 		flags.set(AI::Profile_Flags::Reset_last_hit_target_time_for_player_hits);
 	}
+	// backwards compatible flag for a bug accidentally introduced during this time
+	if (mod_supports_version(3, 6, 14) && !mod_supports_version(23, 0, 0)) {
```
Goober5000, line 776: If this PR is not going in until after stable, then this check should be `mod_supports_version(23, 0, 1)`

Goober5000 (changes requested):

code/ai/ai_profiles.cpp
Baezon: Not 23.2 for the next stable?

Goober5000: I would assume mods would want to test using nightly builds, which would be versioned 23.1.  You'd want the expression to work as soon as it is merged, not wait until we enter a new stable build cycle.

(Maybe 23.1 is a better fit than 23.0.1, but either way it shouldn't be 23.2.)

Baezon: Ok, I see. Nightlies seem to immediately increment the minor so I think 23.1 would do it.

Goober5000 (approved):

PR #5085, opened 2023-01-06
prevent ship from managing ETS if it is playing dead

Goober5000:
This was the bug found recently that was thought to be related to #2157.  It turns out that ships will manage ETS even if they are playing dead, which caused a ship to accelerate during a ship-maneuver as it directed more power to engines.  So this PR adds a "$ships playing dead don't manage ETS:" flag to control this behavior.  Tested and works.

This also capitalizes two flags that did not have an initial capital letter.


MoerasGrizzly:
Wouldn't it make more sense for ships playing dead to maximize their shields?

MjnMixael:
> Wouldn't it make more sense for ships playing dead to maximize their shields?

I would expect ships I tell to "play dead" not to do anything at all. As a FRED, I don't see "play dead" as a behavior but as a complete lack of behavior. Maybe I'm the only one, but my intention when I issue that order is for the ships to not at all be controlled by the AI.

Goober5000:
> I would expect ships I tell to "play dead" not to do anything at all. As a FREDer, I don't see "play dead" as a behavior but as a complete lack of behavior. Maybe I'm the only one, but my intention when I issue that order is for the ships to not at all be controlled by the AI.

Same.

There's also `ai-stay-still` which is intended to stay in one place and not move, but does not imply anything about the AI.

JohnAFernandez (approved):

PR #5086, opened 2023-01-06
Possible fix for weapon fired hook conditions

### LINKED ISSUE #5080: ammo packs reload infinitely on RC builds
@BMagnu Bug has not fully been fixed, a new issue has arisen instead: When firing your primary weapons in Her Finest Hour (for instance), all ammo is instantly replenished as if an ammo pack were used. However, the ammo pack is not selected by default at the start of the mission, and the ammo pack itself is not used up.

Weirdly enough, cycling through your primary weapon selections even once prevents the issue from occuring. But as long as you don't touch the period key at all, you get infinite ammo on both your sidhe and grimmlers.

I tested this with 22.2 stable and it does not occur there.

_Originally posted by @MoerasGrizzly in https://github.com/scp-fs2open/fs2open.github.com/issues/4952#issuecomment-1352347291_

MjnMixael:
Moves the `OnWeaponFired` and `OnPrimaryFired` hooks up into the loop that checks which banks have actually fired a weapon which fixes #5080 . The issue is that `weapon_idx` doesn't always refer the weapon that was fired. Instead it refers to the weapon that tried to fire. In the case of 5080, this included the Ammo Pack if the player was set to Linked Primary Mode. The Ammo Pack is set to No Linked Primaries so the code above here exits early and doesn't fire the weapon. But for the scripting hooks, that's too late because `weapon_idx` has already been changed to refer to the Ammo Pack.

With this PR fix we can pass the weapon of the bank that actually fired to the hook. This also means that the hook will fire once for each bank, though. I _think_ that's actually preferred because it'd be more accurate but I'm unsure. It's early and I didn't sleep great so I'm probably missing some other unintended effect that may have.

### DIFF
--- a/code/scripting/global_hooks.cpp
+++ b/code/scripting/global_hooks.cpp
@@ -273,6 +273,7 @@ const std::shared_ptr<Hook<WeaponUsedConditions>> OnPrimaryFired = Hook<WeaponUs
 	{
 		{"User", "ship", "The ship that has fired the weapon."},
 		{"Target", "object", "The current target of this ship."},
+		{"BankIndex", "number", "The index of the bank that has fired."},
 	});
 
 const std::shared_ptr<Hook<WeaponUsedConditions>> OnSecondaryFired = Hook<WeaponUsedConditions>::Factory("On Secondary Fire",
--- a/code/scripting/global_hooks.cpp
+++ b/code/scripting/global_hooks.cpp
@@ -265,6 +265,7 @@ const std::shared_ptr<Hook<WeaponUsedConditions>> OnWeaponFired = Hook<WeaponUse
 	{
 		{"User", "ship", "The ship that has fired the weapon."},
 		{"Target", "object", "The current target of this ship."},
+		{"BankIndex", "number", "The index of the bank that has fired."},
 	});
 
 const std::shared_ptr<Hook<WeaponUsedConditions>> OnPrimaryFired = Hook<WeaponUsedConditions>::Factory("On Primary Fire",
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -12540,19 +12540,19 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
 			target = &Objects[Player_ai->target_objnum]; 
 
 		for (int bank = 0; bank < MAX_SHIP_PRIMARY_BANKS; bank++) {
-			if(banks_fired & (1 << bank))
+			if(banks_fired & (1 << bank)){
 				//Start Animation in Forced mode: Always restart it from its initial position rather than just flip it to FWD motion if it was still moving. This is to make it work best for uses like recoil.
 				sip->animations.getAll(model_get_instance(shipp->model_instance_num), animation::ModelAnimationTriggerType::PrimaryFired, bank).start(animation::ModelAnimationDirection::FWD, true);
-		}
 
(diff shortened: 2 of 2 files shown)


BMagnu:
Mhm, I see.
I wonder if there are other scripts that might be broken through this change...
There'd be an option to make it fire only once by changing the condition to take a vector of weapons that just need to contain the number specified in the condition. But that requires discussion IMO on what this hook actually _should_ do.

MoerasGrizzly:
If in doubt, please consult the 22.2 and earlier behaviour.

MjnMixael:
> If in doubt, please consult the 22.2 and earlier behaviour.

It's not quite that simple. The old behavior was buggy but in different ways. That's why the whole scripting system got an overhaul in the first place.

BMagnu:
22.2 and earlier likely called the hook once, allowing any of the weapons that fired. Which would be different from how this PR approaches the hook.
I do tend to agree that that is the correct thing to do however, since it pretty much constitutes the core difference between OnWeaponFired (a player presses the trigger once) and OnWeaponCreated (which shoots out four projectiles), I guess. But that is open for debate, and other ideas.

MageKing17:
> 22.2 and earlier likely called the hook once, allowing any of the weapons that fired.

They did not; they only matched on whatever bank was currently equipped (which could be first or last, depending on which direction the banks were linked from).

Goober5000:
Having the script hook fire once for each bank that has fired sounds reasonable IMO.  It's simple and straightforward.  It does lead to the awkward situation if two banks with the same weapon fire while linked, but that may not be a problem in practice.

On that note, it would be useful to add `BankIndex` as a hook variable.  That would help scripts disambiguate duplicate calls if needed.

Using a vector as a parameter would add friction both on the C++ side and on the Lua side.  Additionally, it would only be needed in a very specific scenario, and even then, the need is theoretical at this point.  In my judgement simplicity and ease-of-use should tilt the balance in favor of one call per bank.

MjnMixael:
I have added BankIndex as a parameter

Goober5000:
Good - although it should be set to bank+1 due to Lua's indexing.

The same hook vars are used for `OnPrimaryFired`, so the hook factory declaration should be updated too.  And that raises the question - does this PR's solution also need to be applied to `OnSecondaryFired`?

MjnMixael:
You're free to investigate OnSecondaryFired on your own, I'm not doing it in this PR. I suspect the logic is quite different since there's no linked fire mode.

BMagnu (changes requested):
I've reinvestigated the question of "should this hook fire once per chained weapon bank", and I've come to the conclusion that I think it really should not.
Mostly cause it didn't before, and will break scripts already out there, which I've found in at least JAD, maybe some others too.
IMO, the condition for On*Fired should be changed from taking a single weapon index, to an SCP_vector of weapon indices, and trigger once.
For things that should happen "for each fired weapon", and thus require multiple calls for multiple banks, should likely use OnWeaponCreated.

Baezon:
> IMO, the condition for On*Fired should be changed from taking a single weapon index, to an SCP_vector of weapon indices, and trigger once.
This isn't a bad idea, but how would the script API actually return this? A table of weapon handles? And, of course, doing this would also break scripts using this hook much harder than running it more than once sometimes.

Goober5000:
The `BankIndex` number could be changed to a table, and the table could be populated the same way as `PrimaryBanks` and `PrimaryAmmo` for `l_ParseSubsystem`.

PR #5091, opened 2023-01-08
Override to set glowpoint bank to default off

MjnMixael:
Adds a boolean to glowpoint overrides that allows forcing the bank to default as off. Useful for glowpoints that cast light, making sure they are off in the tech room, ship/weapon select and other places.


wookieejedi:
Very good idea, this will be quite useful.

Goober5000 (approved):
LGTM

PR #5094, opened 2023-01-08
Handle homing secondaries in new lead indicator behaviors

### LINKED ISSUE #5088: Missile lead indicator does not show properly on alternate $Lead Indicator Behavior modes.
On the $Lead Indicator Behavior: Average and $Lead Indicator Behavior: Multiple modes, the missile lead indicator displays weirdly.

In a lot of cases it'll only show when your primary weapons are in range, even though you're using, say, a heatseeker missile with a very long range, like the Stiletto.

To test:
Set $Lead Indicator Behavior: Average or Multiple in your game settings.
Fire up The Sixth Wonder
Pick Stiletto missiles for one of your banks.
Switch to said missiles.

Notice how the missile lead indicator does not display, even though the Stiletto has a range of 5.5km

Kiloku:
fixes #5088

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -4080,22 +4080,18 @@ void HudGaugeLeadIndicator::renderLeadCurrentTarget()
 	}
 	else return;
 
-	//do dumbfire lead indicator - color is orange (255,128,0) - bright, (192,96,0) - dim
+	//do dumbfire or non-merged secondary lead indicator - color is orange (255,128,0) - bright, (192,96,0) - dim
 	//phreak changed 9/01/02
 	if((swp->current_secondary_bank>=0) && (swp->secondary_bank_weapons[swp->current_secondary_bank] >= 0)) {
 		int bank=swp->current_secondary_bank;
-		wip=&Weapon_info[swp->secondary_bank_weapons[bank]];
+		weapon_info* swip = &Weapon_info[swp->secondary_bank_weapons[bank]];
 
-		//get out of here if the secondary weapon is a homer or if its out of range
-		if ( wip->is_homing() && !wip->wi_flags[Weapon::Info_Flags::Dont_merge_indicators] )
-			return;
-
-		double max_dist = MIN((wip->lifetime * wip->max_speed), wip->weapon_range);
+		double max_dist = MIN((swip->lifetime * swip->max_speed), swip->weapon_range);
 
 		if (dist_to_target > max_dist)
 			return;
 
-		if (hud_calculate_lead_pos(&Player_obj->pos, &lead_target_pos, &target_pos, targetp, wip, dist_to_target))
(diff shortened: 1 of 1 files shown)


Kiloku:
I suggest @EatThePath test this if you can. It's code related to a feature you requested, so I don't know if you have a more robust scenario to see if I ended up breaking something while fixing this.

code/hud/hudtarget.cpp
```diff
@@ -4080,22 +4080,18 @@ void HudGaugeLeadIndicator::renderLeadCurrentTarget()
...
+		weapon_info* swip = &Weapon_info[swp->secondary_bank_weapons[bank]];
 
-		double max_dist = MIN((wip->lifetime * wip->max_speed), wip->weapon_range);
+		double max_dist = MIN((swip->lifetime * swip->max_speed), swip->weapon_range);
 
 		if (dist_to_target > max_dist)
 			return;
 
-		if (hud_calculate_lead_pos(&Player_obj->pos, &lead_target_pos, &target_pos, targetp, wip, dist_to_target))
+		if (hud_calculate_lead_pos(&Player_obj->pos, &lead_target_pos, &target_pos, targetp, (swip->is_homing() && !swip->wi_flags[Weapon::Info_Flags::Dont_merge_indicators]) ? wip : swip , dist_to_target))
```
JohnAFernandez, line 4094: I think this may cause some unintended behavior.  This may force the lead indicator to display for primary weapons if they are all out of range but `(swip->is_homing() && !swip->wi_flags[Weapon::Info_Flags::Dont_merge_indicators])` is true.  For instance, Grizzly's stiletto, mentioned in the issue, will have this as true, because retail data is not going to set this SCP flag, of course.

So we need to figure out what to do if primaries are out of range.  If primaries are out of range, and a stiletto is in range, the indicator should display using the stiletto's info.  If the primaries are out of range, and a non-homing weapon or a weapon with the `Dont_merge_indicators` flag is set, the indicator should *still* be displayed using that other weapon's info.

So, as best I can tell, instead of using the primary weapon info to fall back on if `if ( wip->is_homing() && !wip->wi_flags[Weapon::Info_Flags::Dont_merge_indicators] )` is true, we should be checking in the primary weapon section if any of the primary weapons caused the indicator to be rendered, and then expanding the `if ( wip->is_homing() && !wip->wi_flags[Weapon::Info_Flags::Dont_merge_indicators] )` with a bool so that it's something like `if ( wip->is_homing() && !wip->wi_flags[Weapon::Info_Flags::Dont_merge_indicators] && primary_rendered )`

Kiloku: My opinion is that the whole lead indicator code is too complicated for what it does and deserves a rewrite. In which case I think the best path forward is to actually not merge this PR, close it, and then do a new one for the rewrite. The bug this PR aims to fix only happens on modded behavior and even then is not game-breaking, so it can wait the next release.

TRBlount:
Is this still in a draft state?

JohnAFernandez:
I'm going to go with Kiloku's plan and close this PR.  It's just going to get pushed back over and over again.  But someone needs to work on refactoring this functions soon.

JohnAFernandez:
I'm going to go with Kiloku's plan and close this PR.  It's just going to get pushed back over and over again.  But someone needs to work on refactoring this functions soon.

PR #5095, opened 2023-01-09
Avoid calling `asteroid_add_target()` in FRED.

### LINKED ISSUE #5092: 23.0.0 RC3 FRED crashing without error when loading mission
Using build FSO-23.0.0-RC3.

May need to check how FRED is saving mission files as unable to open files recently saved in RC3. 
Opening recently saved missions in RC3 crashes FRED without error message or debug.
On reverting back to stable 22.0.0 error messages report missing $AI Behavior: under #Objects, and +Neb2Flags in#Background bitmaps.
Once added back in the mission loads again in stable 22.0

MageKing17:
PR #5058 made `count_incident_asteroids()` use `Asteroid_obj_list` instead of `obj_used_list`; a perfectly sensible change in and of itself. However, PR #4827 had previously made `asteroid_obj_list_init()` only be called if FRED wasn't running (presumably because of the iteration through `Asteroid_objs` to clear flags, which is wasted effort in FRED). The problem is that without `list_init()` being called on `Asteroid_obj_list`, its `prev` and `next` were `nullptr` instead of linking back to itself, which means the beginning of the list and the end of the list were no longer considered the same (the iterator's `begin()` calls `GET_FIRST(head)`, which is effectively `head->next`, while `end()` calls `END_OF_LIST(head)`, which is effectively just `head`). This meant that iterating over `list_range(Asteroid_obj_list)` was returning `nullptr` as the first element of the list, and trying to dereference that would cause a crash. This was still not a problem in and of itself, as long as nothing tried to iterate over `Asteroid_obj_list` while FRED was running, which theoretically shouldn't be necessary...

PR #3704 changed mission parsing so that when parsing a ship with a name that matches one of the `Asteroid_field.target_names`, it would call `asteroid_add_target()` to initialize the relevant data for the asteroid field to fling asteroids at that ship. `asteroid_add_target()` calls `count_incident_asteroids()`, which (since PR #5058) iterates over `Asteroid_obj_list`, causing a crash due to dereferencing a null pointer on any mission that defines `$Asteroid Targets:` (and then actually has such a target be parsed).

While this could be solved by unconditionally initializing `Asteroid_obj_list`, there's really no need to do anything with `Asteroid_obj_list` in FRED, so instead the `asteroid_add_target()` call has been gated behind a `!Fred_running` check, and `Assertion()`s have been added to make sure `Asteroid_obj_list` has been initialized at the beginning of any function that does anything with it (besides initialize it), to catch similar problems in the future.

Fixes #5092.

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -90,9 +90,9 @@ SCP_vector<asteroid_target> Asteroid_targets;
 static int count_incident_asteroids(int target_objnum)
 {
 	object* asteroid_objp;
-	int		count;
+	int		count = 0;
 
-	count = 0;
+	Assertion(GET_FIRST(&Asteroid_obj_list) != nullptr, "count_incident_asteroids() called before Asteroid_obj_list was initialized; this shouldn't happen. Get a coder!");
 
 	for (auto aop: list_range(&Asteroid_obj_list)) {
 		asteroid_objp = &Objects[aop->objnum];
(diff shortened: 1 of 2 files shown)


Goober5000 (approved):
Good catch and good fix. 👍

PR #5098, opened 2023-01-09
fix team color parsing

MjnMixael:
"None" is a valid team color setting so we don't need to warn in the log and then set team color to "none". Also fixes a small bug where FRED would never show "None" as selected in the Initial Status dialog box.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -2934,9 +2934,12 @@ int parse_object(mission *pm, int  /*flag*/, p_object *p_objp)
 		stuff_string(temp, F_NAME, NAME_LENGTH);
 		p_objp->team_color_setting = temp;
 
-		if (Team_Colors.find(p_objp->team_color_setting) == Team_Colors.end()) {
-			mprintf(("Invalid team color specified in mission file for ship %s, resetting to default\n", p_objp->name));
-			p_objp->team_color_setting = sip->default_team_name;
+		//If team color is not default then verify
+		if (stricmp(p_objp->team_color_setting.c_str(), sip->default_team_name.c_str()) != 0) {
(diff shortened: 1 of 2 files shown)


code/mission/missionparse.cpp
```diff
@@ -2934,9 +2934,12 @@ int parse_object(mission *pm, int  /*flag*/, p_object *p_objp)
 		stuff_string(temp, F_NAME, NAME_LENGTH);
 		p_objp->team_color_setting = temp;
 
-		if (Team_Colors.find(p_objp->team_color_setting) == Team_Colors.end()) {
-			mprintf(("Invalid team color specified in mission file for ship %s, resetting to default\n", p_objp->name));
-			p_objp->team_color_setting = sip->default_team_name;
+		//If team color is not default then verify
+		if (stricmp(p_objp->team_color_setting.c_str(), sip->default_team_name.c_str()) != 0) {
```
JohnAFernandez, line 2938: Just a quick thing here.  SCP_strings (std::string) can just use == instead of having to use stricmp.

JohnAFernandez (changes requested):

code/mission/missionparse.cpp
MjnMixael: Done

MjnMixael: Actually hang on. I do want stricmp here because we want case insensitive comparison.

JohnAFernandez: ok

JohnAFernandez (approved):

code/mission/missionparse.cpp
Goober5000: `SCP_string_lcase_equal_to` could be used

PR #5100, opened 2023-01-10
Fix regression in sexp argument saving

### LINKED ISSUE #5099: FRED has become nagari sensitive and is being manipulated into changing sexps.
[bp2-23.zip](https://github.com/scp-fs2open/fs2open.github.com/files/10385186/bp2-23.zip)

War In Heaven's Universal Truth's opening scene has a section where Sathani bombard earth. With my more recent updates to BP, this stopped working. I never could figure out why that was.

This is the relevant SEXP tree. This represents the tree in its working order: 
afbeelding

However, when the mission is saved, this happens: 
afbeelding

Behaviour also occurs in FSO 22.2 stable. Behaviour occurs without warning, regardless if a fast debug build is used or not. 

Notably, when the mission is loaded into FSO without FRED's alterations, it works correctly. It no longer does after it is loaded and saved in fred, even if that is the only action taken.

The-E:
When checking sexp text strings, make sure to account for usages of `<argument>`

fixes #5099

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -4783,7 +4783,9 @@ void stuff_sexp_text_string(SCP_string &dest, int node, int mode)
 			{
 				// if (for whatever reason) we have an empty string or an invalid number, print 0
 				// do not do this trick in error mode, because we want to know what the text actually is
-				if ((mode != SEXP_ERROR_CHECK_MODE) &&
+				// note that we let <argument> strings pass here
+				if ((mode != SEXP_ERROR_CHECK_MODE) && 
+					(!(Sexp_nodes[node].flags & SNF_SPECIAL_ARG_IN_NODE)) &&
 					(*ctext_string == '\0' || !can_construe_as_integer(ctext_string)))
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -4783,7 +4783,9 @@ void stuff_sexp_text_string(SCP_string &dest, int node, int mode)
 			{
 				// if (for whatever reason) we have an empty string or an invalid number, print 0
 				// do not do this trick in error mode, because we want to know what the text actually is
+				// note that we let <argument> strings pass here
 				if ((mode != SEXP_ERROR_CHECK_MODE) &&
+					(stricmp(ctext_string, "<argument>")) &&
```
Goober5000, line 4788: This should check `Sexp_nodes[node].flags & SNF_SPECIAL_ARG_IN_NODE` per the recent reduction of sexp string comparisons.

Goober5000 (changes requested):

Goober5000 (approved):

PR #5101, opened 2023-01-10
fredTunes, Fredify, Fred Music, or whatever. It's a music player.

MjnMixael:
Adds a Music Player to FRED that lists all the loaded music files from data/music and allows playing them while you work on your mission. Includes an autoplay feature. If set, then FRED will just keep playing music down the list until it reaches the end. In addition, you can also ignore music tracks from the music player with a game_settings.tbl list for modders who want ignore upstream music files they don't use. You can also view the music.tbl from the Music Player dialog to remember which files are part of which soundtrack.

NOTE: I don't have a qtFRED dev environment setup, but once I get my mod out I intend to dive in head first on that and will port over this feature then.


fred2/musicplayerdlg.cpp
```diff
@@ -0,0 +1,214 @@
...
+#endif
+
+music_player_dlg::music_player_dlg(CWnd* pParent /*=nullptr*/)
+	: CDialog(music_player_dlg::IDD, pParent)
+{
+
+	m_music_item = "";
+	m_music_id = -1;
+	m_cursor_pos = -1;
+	m_autoplay = FALSE;
```
JohnAFernandez, line 24: Default to autoplay off?  Do you believe this is what most users will want?

fred2/musicplayerdlg.cpp
```diff
@@ -0,0 +1,214 @@
...
+	if (m_music_id < 0) {
+		thisMusic = m_music_item + ".ogg";
+		m_music_id = audiostream_open(thisMusic.c_str(), ASF_EVENTMUSIC);
+	}
+
+	if (m_music_id >= 0) {
+		audiostream_play(m_music_id, 1.0f, 0);
+	} else {
+		thisMusic = m_music_item;
+		mprintf(("FRED failed to open music file %s in the music player\n", thisMusic.c_str()));
```
JohnAFernandez, line 119: Is there a way to show this on the dialog window?  I'm not as good at MFC as I am with qt, but it could be good to have visual feedback, since a warning would probably be overkill.

fred2/musicplayerdlg.cpp
```diff
@@ -0,0 +1,214 @@
...
+		return;
```
JohnAFernandez, line 120: Extra return?  The function is about to terminate anyway.

fred2/musicplayerdlg.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+	if (m_music_id >= 0) {
+		audiostream_close_file(m_music_id, 0);
+		m_music_id = -1;
+	}
+}
+
+bool music_player_dlg::SelectNextTrack()
+{
+	if ((m_cursor_pos >= 0) && (m_cursor_pos < (m_num_music_files - 1))) {
```
JohnAFernandez, line 134: Suggestion: we could wrap this to position 0 if on the last track fairly easily.

fred2/musicplayerdlg.cpp
```diff
@@ -0,0 +1,214 @@
...
+		UpdateSelection();
+		return true;
+	}
+	
+	return false;
+}
+
+bool music_player_dlg::SelectPrevTrack()
+{
+	if ((m_cursor_pos > 0) && (m_cursor_pos <= (m_num_music_files - 1))) {
```
JohnAFernandez, line 146: I almost marked this as not working for the last item on the list because I misread it for a second.  It might be a tiny bit clearer to do `(m_cursor_pos < (m_num_music_files))` since then you're using the same operator in these two similar, consecutive functions, but I'm not insisting on that change.

fred2/musicplayerdlg.cpp
```diff
@@ -0,0 +1,214 @@
...
+		m_cursor_pos--;
+		m_music_list.SetCurSel(m_cursor_pos);
+		UpdateSelection();
+		return true;
```
JohnAFernandez, line 150: Like in the sister function, it should be pretty simple to make it wrap to the last item from the 0th item, but again, just a suggestion.

fred2/res/play.bmp
JohnAFernandez, line 1: Is there anything that's relying on the filename being what it was previously?

fred2/musicplayerdlg.cpp
MjnMixael: It's what I would want. Most of the time I would use this to pick music for missions rather than just use it to listen to music while I FRED.

fred2/musicplayerdlg.cpp
MjnMixael: In most music players, a loop method is it's own button that allows no loop, loop all, or loop 1. I'm not implementing any of that here, so I'd rather keep it simple to match what people expect.

fred2/musicplayerdlg.cpp
MjnMixael: See above.

fred2/res/play.bmp
MjnMixael: Anything that uses the play.bmp will get it by calling the IDB_PLAY define. For completeness I checked the other UIs that use it and they still show it correctly.

fred2/musicplayerdlg.cpp
MjnMixael: Honestly, I'd rather do a warning. The music player window isn't very big and would make a sentence like that hard to read even if I could find a place for it.

fred2/res/play.bmp
JohnAFernandez: Thank you!

fred2/musicplayerdlg.cpp
JohnAFernandez: I think a warning would be good then, since the mission maker will not have to check the debug log to see if the song failed to open.

fred2/fredview.cpp
```diff
@@ -4125,6 +4127,17 @@ void CFREDView::OnCalcRelativeCoords()
 	dlg.DoModal();
 }
 
+void CFREDView::OnMusicPlayer()
+{
+	Assert(Music_player_dialog.GetSafeHwnd());
```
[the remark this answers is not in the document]
MjnMixael: Done

fred2/musicplayerdlg.cpp
MjnMixael: Removed

fred2/musicplayerdlg.cpp
MjnMixael: Ah, yeah. That's dumb. Fixed.

JohnAFernandez:
I think we're good to go.

JohnAFernandez (approved):

Issue #5020, opened 2022-12-17
FR: Allow multiple species assignments to messages.tbl - and also the option to never auto assign a persona


MoerasGrizzly:
wrt messages.tbl, there's some stuff that could make that table a bit less messy for BP in particular.
Specifically: The ability to assign a persona to multiple species, ergo:
Persona5
+Terran
+GTVA-TEI

Simms
+DoNotAssign

BP has:
Factions, rather then species, so some pilot personas make sense for multiple "species".
Several personas that are tied to a specific character. By default, these could get assigned to random pilots (FSBlue players may remember hearing Corey or Taylor during the FS2 campaign). Currently, BP has a workaround for that, but I'd rather just have a neat engine solution.

Goober5000:
Implemented in #5104

PR #5104, opened 2023-01-12
Persona species enhancements

Goober5000:
Implements features from #5020:
1. Personas can now be assigned to multiple species
2. Personas can be excluded from automatic assignment if the game picks a random persona

Also removes the obsolete `PERSONA_FLAG_VASUDAN` and turns the `substitute_missing_messages` field into a proper flag to match the others.

Additionally, this fixes the addition of the -Vas suffix on the wingmen personas in FRED, which had been broken since the species_defs upgrade.


wookieejedi (approved):
Useful changes and LGTM

PR #5106, opened 2023-01-12
Allow negative number in event editor

### LINKED ISSUE #4292: FR: FRED - infinite repeat/trigger count
As we already know, to have an event firing infinitely we need to put a massive number into the repeat count, or as some prefer, edit the mission file in a text editor and place the -1 value in there.

I am requesting a different method to handle this, for the sake of ease of use.

Variant 1: just let repeat count and possibly trigger count accept negative numbers, with any negative number being treated as infinity.
Variant 2: instead of variant 1, place a pair of checkboxes marked as "infinite" for both values, when marked the place for inserting a value being grayed out.
Variant 3: since this is usually used to have an event fire infinitely, often every frame, maybe instead let's have a "repeat every frame" or "evaluate every frame" checkbox (that when active would gray out all of the boxes - for repeat count, trigger count and interval time) and a "repeat infinitely" or "evaluate infinitely" checkbox (that when active would gray out repeat and trigger count boxes, but not interval time box).

MjnMixael:
Based on my reading of the code at missionparse.cpp line 4946, a negative number in Repeat Count or Trigger Count means to do it indefinitely. Issue #4292 says, there's no way to specify infinite count in FRED outside of putting in a very large number. This PR changes those input boxes by removing ES_NUMBER (which only allows strictly digits) and changing the variable from UINT to int. Numbers smaller than -1 are changed to -1 because there's no point to smaller numbers currently. If text is input then a warning box will popup asking for an integer and the value changed to 0.

I did not fully test this and can't until this weekend when I can also look at the qtFRED side of the code for the same thing. Draft for now.

FSO currently replaces any Count higher than MAX_INT with -1 on save anyway. So I'm fairly confident there's no issues with missionparse or missionsave.

Fixes #4292


MoerasGrizzly:
There's actually some ~2015 missions I've seen that had -1 as trigger count. They got converted into 1s sometime later.

MjnMixael:
> There's actually some ~2015 missions I've seen that had -1 as trigger count. They got converted into 1s sometime later.

This is merely a UI change. It does not change how the game handles -1 as a trigger count because that is currently valid. Create a mission in the latest nightly and set the count to the highest number and save. Check the mission and it will be -1. That's expected behavior.

MoerasGrizzly:
> Check the mission and it will be -1. That's expected behavior. In fact BP has a ton of these.

Yes. But if you open the mission again in fred, it'll display a 1. It will then also _save this as a 1 rather then a minus one_

MoerasGrizzly:
To explain this a little bit further: This would not only be a nice convenience feature, it would also solve a bug where FRED messes with events without user intervention.

MjnMixael:
> > Check the mission and it will be -1. That's expected behavior. In fact BP has a ton of these.
> 
> Yes. But if you open the mission again in fred, it'll display a 1. It will then also _save this as a 1 rather then a minus one_ when the mission is saved again.

That's actually not true. Under current behavior a -1 in the mission file is translated to a maximum UINT value (over 4 million).

The-E:
This PR addresses the issue in the correct way; I don't think it needs to be a draft. This is a fix that should go into 23.0

MjnMixael:
> This PR addresses the issue in the correct way; I don't think it needs to be a draft. This is a fix that should go into 23.0

I just need to handle the qtFred side which I can do this evening.

The-E (approved):

PR #5107, opened 2023-01-13
some validity checks to order sending

MjnMixael:
As requested by Wookie, this PR does a few things to verify some order-sending things. 

First it checks that at least one ship in a wing can depart before marking the order as allowed on the Comms Board. 

Second it enhances `ship_get_random_ship_in_wing()` to be able to check if a ship can actually perform an order. If it can't, then don't select the ship. This prevents ships that can't perform a "wing" or "all ships" order from acknowledging the order as if they can.

### DIFF
--- a/code/hud/hudsquadmsg.cpp
+++ b/code/hud/hudsquadmsg.cpp
@@ -231,7 +231,6 @@ void hud_add_issued_order(const char *name, int order);
 void hud_update_last_order(const char *target, int order_source, int special_index);
 bool hud_squadmsg_is_target_order_valid(size_t order, ai_info *aip = nullptr);
 bool hud_squadmsg_ship_valid(ship *shipp, object *objp = nullptr);
-bool hud_squadmsg_ship_order_valid( int shipnum, int order );
 
 // function to set up variables needed when messaging mode is started
 void hud_squadmsg_start()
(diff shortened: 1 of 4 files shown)


wookieejedi:
Tested and works as expected. Orders do not show up for wings if all ships in wings cannot do that order, and similarly no confirmation message is sent if the player tries to issue an order the AI cannot do via `All Fighters`.

Goober5000 (approved):
LGTM too

code/hud/hudsquadmsg.cpp
```diff
@@ -2098,6 +2097,18 @@ void hud_squadmsg_wing_command()
...
+			int active = 0;
+			for (int i = 0; i < wingp->current_count; i++) {
+				if (hud_squadmsg_ship_order_valid(wingp->ship_index[i], (int)order_id)) {
+					active = 1;
+					break;
+				}
+			}
+			MsgItems[Num_menu_items].active = active;
+		}
+
```
Goober5000, line 2111: Why is this check actually restricted to DEPART_ITEM?  Shouldn't wing orders always be unavailable if no ship in the wing can perform them?  And if so, are there any similar checks in other parts of the squadmsg code?

wookieejedi: I'm not sure why it's like that in the code, but indeed I reckon wing orders should be unavailable if no ship in the wing can perform them, which this PR seems to remedy in my tests. Also IIRC there's not really any similar checks for wings or all fighters.

wookieejedi: EDIT: Finally recall that adding this extra check is needed b/c warp ability can be set via flag and the code seemed to check for protecting and attacking orders already.

PR #5108, opened 2023-01-14
Remove the screen signature

### LINKED ISSUE #5064: Graphics initialization during parse?
PR #4969 caught my eye because it performs graphics operations during parsing, which is a code smell.  Furthermore, the game_settings.tbl and *-mod.tbms are parsed before `gr_init` is called, so graphics aren't even initialized at this point.

Looking at the code flow more closely, the only part of `gr_init_color` that depends on the graphics context is `gr_screen.signature`.  It just so happens that the default value of the signature field is 0, and the first `Gr_signature` value is *also* 0, so at least this is consistent.  Is there any risk that out-of-date signature values will cause problems for subsequent signatures?

Looking at other instances of `gr_init_color` throughout the code, there are a few that occur frequently (including one in `batching_add_line`) and a few others that occur only during ship/weapons table parsing, so this sort of split decision isn't limited to the mod table.

Wondering what more savvy graphics coders, such as @Baezon, @notimaginative, @TRBlount, and @EatThePath, have to say about this.

The-E:
As discussed in #5064, there is a piece of legacy code in the graphics initialization code that references a "screen signature". Back in the olden days, this value was used to track whether or not certain data needed to be reinitialized in response to changes in the graphics environment; this is no longer a concern with our current render backend.

Closes #5064

### DIFF
--- a/code/anim/animplay.cpp
+++ b/code/anim/animplay.cpp
@@ -397,8 +397,6 @@ int anim_show_next_frame(anim_instance *instance, float frametime)
 
 		t1 = timer_get_fixed_seconds();
 		for ( i = 0; i < frame_diff; i++ ) {
-			anim_check_for_palette_change(instance);			
-
 			// if we're playing backwards, every frame must be a keyframe and we set the data ptr here
 			if(instance->direction == ANIM_DIRECT_REVERSE){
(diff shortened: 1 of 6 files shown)


Goober5000 (approved):
Nice work, this fits what I had hoped could be done.

PR #5111, opened 2023-01-15
Improve handling of failure to find spawn weapons

Baezon:
`spawn_type` (or `spawn_wep_index` as I've renamed it) is at first an index into the raw parsed weapon names used in the flag, to be filled with the actual weapon info index later during parsing, but if the weapon cannot then be found at that time, that index into `Spawn_names` is kept as is and used as a weapon index resulting in odd scenarios where it will spawn random other weapons if any of the syntax was mispelled.


MageKing17 (approved):
Sensible sanity checks, and no errors jump out at me.

Issue #5112, opened 2023-01-15
Heavy permenant performance degredation from trails


EatThePath:
in a test mod I fire huge numbers of weapons with trails in a short period of time. Before this happens I have a solid 120 fps, afterwards in an otherwise identical scene (all trails faded and weapons gone) I idle in the 17-20fps. A mission restart brings idle up to 20-25ish.

this seems to be tied entirely to trails. If I turn trails off I don't see this, toggling other features such as weapon lighting doesn't have an impact.

Profiling shows that the vast majority of frametime in the slow period is spent in `gropengltnl.cpp:gr_opengl_update_buffer_data()`

Attached is a test mod, launch it with fps display on and run the included mission, and you'll see the behavior in moments.
[trailthrash.zip](https://github.com/scp-fs2open/fs2open.github.com/files/10419375/trailthrash.zip)

The-E:
After taking a quick look, it's pretty clear that we have a ressource management problem here. Most of the permanently lost frametime is spent updating buffer objects that are, for all intents and purposes, empty; they represent trails that have already decayed.

This can be shown by altering `batching.cpp:946` thusly:
```
	if (num_items && draw_queue->buffer_num.isValid()) {
		gr_update_buffer_data(draw_queue->buffer_num, draw_queue->buffer_size, draw_queue->buffer_ptr);
	}
```

This prevents the buffer update from happening when there are no primitives associated with the batch, leading to the framerate recovering back to normal once all the trails have expired.

This obviously requires further investigation and work.

PR #5115, opened 2023-01-15
do not update the gl buffer for a batch if the batch is empty

The-E:
This addresses behaviour described in #5112. Profiling reveals that we are spending a lot of time per frame updating the buffers we use for batched rendering, even if the batches are nominally "empty" and thus wouldn't be rendered. This shows for example in cases where a lot of trails are spawned; even once all those trails have expired, the buffers created for them are still opened and updated, which involves a round trip through the GPU driver and PCIe bus, with all the overhead that implies. 
The change here removes the buffer update in those cases, meaning that while batch rendering itself might still be slow in extreme cases (see #5112 for an example mod and mission), it will no longer incur a _permanent_ performance penalty.

This is part of an ongoing effort to improve the performance of this piece of code, and should be considered a stopgap measure at best.

### DIFF
--- a/code/render/batching.cpp
+++ b/code/render/batching.cpp
@@ -943,7 +943,8 @@ void batching_allocate_and_load_buffer(primitive_batch_buffer *draw_queue)
 		offset += item->n_verts;
 	}
 
-	if (draw_queue->buffer_num.isValid()) {
+	// if there are no items in this batch, we will never render it and thus there is no need to update it in vmem
+	if (num_items && draw_queue->buffer_num.isValid()) {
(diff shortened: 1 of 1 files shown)


EatThePath:
I've been running tests with this change this morning and in the most extreme cases it improves my framerate by a factor of about 6x. My test case is a quite pathological one, but I think this is likely to help with the reports we've had of performance degradation over long play sessions and after events like big ship deaths. 
We'll hopefully be able to iron out the root causes more over time but the benefits are too big to delay for that, and even if we do finish them this seems like it'd still be a useful optimization.

EatThePath (approved):

JohnAFernandez (approved):

Issue #4188, opened 2022-04-20
Chrome will not download from non-HTTPS URL's


z64555:
Chrome, and likely several other browsers, will start to aggressively ignore URL's that are not HTTPS.  In the case of Chrome, it may, at most, open a tab and then immediately close it.

This is an issue for our scp.indiegames host which doesn't use the HTTPS secure protocol, so that will have to be updated in the near future.

Goober5000:
This is being handled, but may not be ready in time for 23.0.

Goober5000:
Server has been migrated and HTTPS support has been added.

PR #5118, opened 2023-01-16
update the .mako file for release posts

Goober5000:
1. Change the "previous release" thread to 22.2
2. Change http to https
3. Link to the github Bug Triage page rather than the Mantis filter


Goober5000:
Bleh, now they are.  I had to revert a few to http.  Fixing #4188 (recently back in progress) will handle indiegames but the same thing will need to be done for fsmods.  One thing at a time.

z64555 (approved):

PR #5121, opened 2023-01-16
Feature/pool/pooled trails

EatThePath:
This branch is a test case for a new generational index based data structure that aims to replace our linked lists. The expected outcomes include safer references between objects, fewer stale pointer or index related issues, better iteration performance especially over long play sessions and easy switching from fixed length to variable length data structures.

This branch is not finished, but it is in a state where others may be able to usefully comment on the idioms in it and point out any missteps or missed opportunities.

The code in pool.h is definitely not done, nor very presentable. Trails.cpp contains the most examples of lower level pointer junk needing to be rewritten, and the rest show what changes have been required in routine external code having to touch trails.

I'm open to both hearing about ways this switch can be made more seamless, and also "while we're at it" ideas on ways we can improve the codebase's long term health by deviating a bit more.

### DIFF
--- a/code/globalincs/pool.h
+++ b/code/globalincs/pool.h
@@ -76,26 +76,27 @@ class SCP_Pool {
 		gEntry<T> n;
 		n.generation = 0;
 		//Assuming trivial constructor behavior here...
+		n.stored = optional<T>(T());
 		storage.push_back(n);
-		pool_index i(storage.size()-1,0);
+		pool_index i((int) storage.size()-1,0);
 		return i;
 	};
 
 	//Internal function to use an existing free slot for a new object
 	pool_index use_free(){
 		auto i = known_empty.back();
-		Assertion(storage[i].value == tl::nullopt ,"Attempted to use_free on a slot that was not free");
+		Assertion(storage[i].stored == tl::nullopt ,"Attempted to use_free on a slot that was not free");
 		//values are created as nullopt when expanding a list
 		//and resest to nullopt here so we need to construct a new object to go in there.
-		storage[i].value = T();
+		storage[i].stored = T();
 		storage[i].generation++;
-		pool_index r(i,storage[i].generation);
+		pool_index r((int) i,(int) storage[i].generation);
 		known_empty.pop_back();
 		return r;
 	}
 	//internal function to find or create a new slot
 	//if size is capped and there's no slot, return nullopt.
-	optional<pool_index> get_new(){
+	optional<pool_index> internal_get_new(){
 		if (known_empty.empty()) {
 			if (capped && storage.size()>=cap) {
 				return tl::nullopt;
@@ -104,7 +105,7 @@ class SCP_Pool {
 		}
 		return use_free();
 	};
-	//Internal function to 
+	//Internal function to fill out the empty list when the storage is known to be clear.
 	void fill_empty_list(){
 		known_empty.clear();
 		if((int) storage.size()>0){
@@ -178,9 +179,7 @@ class SCP_Pool {
 
 
 	optional<pool_index> getNew(){
-		T n = T();
-		auto i = add(n);
-		return i;
+		return internal_get_new();
 	};
 
 	//Store a value in the vector
--- a/code/globalincs/pool.h
+++ b/code/globalincs/pool.h
@@ -1,0 +1,344 @@
+#pragma once
+
+#include <tl/optional.hpp>
+//#include <vector>
+
+using  tl::optional;
+struct pool_index{
+	private:
+	int index;
+	int generation;
+	bool is_null = true;
+	public:
+	pool_index(){ nullify();};
+	pool_index(int i,int g,bool b = false){
+		if(b){ nullify();
+		}
+		else{
+			set(i,g);
+		}
+	};
+	void set(int i,int g){
+		is_null = false;
+		index = i;
+		generation = g;
+	}
+	void nullify()
+	{
+		index = -1;
+		generation = -1;
+		is_null=true;
+	};
+	bool operator==(pool_index &rhs) const {
+		if(is_null && rhs.is_null)
+			return true;
+		if(is_null || rhs.is_null)
+			return false;
+		return (index == rhs.index && generation == rhs.generation);
+	};
+	pool_index& operator=(pool_index const&rhs) {
+		if(rhs.is_null)
+			nullify();
+		else
+			set(rhs.index, rhs.generation);
+		return *this;
+	};
+	int i() const { return index;};
+	int g() const { return generation;};
+	bool null() const { return is_null;};
+};
+
+
+
+template  <typename T>
+struct gEntry {
+	int generation;
+	optional<T> stored;
+
+};
+
+using std::vector;
+template  <typename T>
+//rules:
+//Caps can only ever be set at the current size or smaller
+//shrinking the vector would lead to forgetting the generation counts and then potential slot reuse if allowed to expand later.
+class SCP_Pool {
+	vector<gEntry<T>> storage;
+	vector<size_t> known_empty;
+	bool capped = false;
+	size_t cap = 0;
+
+	/*!
+	//Internal function to expand the storage list with a new object
+	//for use when there are no empty slots and the storage is not capped
+	!*/
+	pool_index add_new(){
+		gEntry<T> n;
+		n.generation = 0;
+		//Assuming trivial constructor behavior here...
+		storage.push_back(n);
+		pool_index i(storage.size()-1,0);
+		return i;
+	};
+
+	//Internal function to use an existing free slot for a new object
+	pool_index use_free(){
+		auto i = known_empty.back();
+		Assertion(storage[i].value == tl::nullopt ,"Attempted to use_free on a slot that was not free");
+		//values are created as nullopt when expanding a list
+		//and resest to nullopt here so we need to construct a new object to go in there.
+		storage[i].value = T();
+		storage[i].generation++;
+		pool_index r(i,storage[i].generation);
+		known_empty.pop_back();
+		return r;
+	}
+	//internal function to find or create a new slot
+	//if size is capped and there's no slot, return nullopt.
+	optional<pool_index> get_new(){
+		if (known_empty.empty()) {
+			if (capped && storage.size()>=cap) {
+				return tl::nullopt;
+			}
+			return add_new();
+		}
+		return use_free();
+	};
+	//Internal function to 
+	void fill_empty_list(){
+		known_empty.clear();
+		if((int) storage.size()>0){
+			for (int i=((int) storage.size())-1; i>=0;i--){
+				known_empty.push_back(i);
+			}
+		}
+	}
+
+	public:
+	SCP_Pool<T>() = default;
+	SCP_Pool<T>(size_t cap){
+		reset(cap);
+	};
+
+	void reset(){
+		capped = false;
+		cap = 0;
+		for(gEntry<T> entry: storage){
+			if (entry.stored != tl::nullopt)
+				entry.stored.reset();
+		}
+		fill_empty_list();
+	}
+
+	void reset(size_t cap_in){
+		capped = true;
+		cap=MAX(cap_in,storage.size());//shrinking a container is illegal.
+
+		for(gEntry<T> entry: storage){
+			if (entry.stored != tl::nullopt)
+				entry.stored.reset();
+		}
+		for (size_t i=storage.size(); i<cap;i++){
+			gEntry<T> e;
+			e.generation = 0;
+			e.stored = tl::nullopt;
+			storage.push_back(e);
+		}
+		fill_empty_list();
+	}
+
+	//Index into the storage... bare reference
+	T& operator[](pool_index i) {
+		T* v = nullptr;
+		if (i.i() >= storage.size()) {
+			return *v;
+		}
+		if (storage[i.i()].generation!= i.g()){
+			return *v;
+		}
+		if(storage[i.i()].stored==tl::nullopt) 
+			return *v;
+//		v = (storage[i.i()].stored).operator->();
+		return storage[i.i()].stored.operator*();
+	};
+
+
+	optional<T*> get_pointer(pool_index i){
+		if (i.i() >= storage.size()) {
+			return tl::nullopt;
+		}
+		if (storage[i.i()].generation!= i.g()){
+			return tl::nullopt;
+		}
+		optional<T> s = storage[i.i()].stored;
+		T* p = s.operator->();
+		return p;
+	};
+
+
+
+	optional<pool_index> getNew(){
+		T n = T();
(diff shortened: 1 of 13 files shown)


EatThePath:
one thing to add: I set up a test mod to as much as I can stress trail creation, iteration and destruction for bench marking purposes. The branch and the baseline are so close together in performance I would call it measurement error, save for the fact that in nearly every case the branch is ever so slightly faster than the baseline. The baseline code isn't really a hot spot even in my most taxing test case so it's not too surprising that it's not a significant difference, but the trend is encouraging for the possibility of performance gains when applied more broadly.

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+		T* operator*(){
+			//vec->get_pointer(full_position);
+			if(vec->check(full_position))
+				return &vec->operator[](full_position);
+			return nullptr;
+		};
+		T& operator&(){
+//			T t = *vec[full_position];
+			return vec->operator[](full_position);
+		};
```
BMagnu, line 357: It is a _bad_ idea to overload operator& and not returning a pointer.
Really, there is no real reason in most cases, and definitely not here, to overload it.
Just let the built in operator return the address of the iterator when used.

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+		using storage_t = vector<gEntry<T>>;
+		using storage_p = vector<gEntry<T>>*;
+
+		iter(SCP_Pool<T> *vec_ptr, storage_p storage_ptr){
+			storage = storage_ptr;
+			vec = vec_ptr;
+		};
+		iter(SCP_Pool<T> *vec_ptr, storage_p storage_ptr, int start_pos){
+			storage = storage_ptr;
+			vec = vec_ptr;
```
BMagnu, line 305: I'd prefer if vec / vec_ptr would be references here.
Also, they should be assigned using member initializer lists.
storage on the other hand, should really not be stored as a variable at all, but accessed through vec. Which, if that is a reference, is not actually a dereference that costs anything.

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+	struct iter{
+		using iterator_category = std::input_iterator_tag;
+		using index = int;
+		using storage_t = vector<gEntry<T>>;
+		using storage_p = vector<gEntry<T>>*;
+
+		iter(SCP_Pool<T> *vec_ptr, storage_p storage_ptr){
+			storage = storage_ptr;
+			vec = vec_ptr;
+		};
```
BMagnu, line 302: This construtor probably should not exist, or at the very least, just be an alias to iter(..., ..., 0), as it allows to create an iterator that points to an invalid element if element 0 is empty

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+	SCP_Pool<T>::iter begin(){
+		return iter(this,&storage,0);
+		//if (known_empty_) {
+		//statements
+		//}
+	};
+	SCP_Pool<T>::iter end(){
+		return iter(this,&storage,(int) storage.size());
+	}
+};
```
BMagnu, line 374: Now that we've had so much fun with this iterator, how about we do it again.
On a serious note, having a const_iterator that does basically the same except returning const references is absolutely needed here.
Both are cbegin and cend useful overloads to enable optimizations, having a ``SCP_Pool<T>::const_iterator begin() const;`` overload is cruical to allow foreaching and iterating the pool even when we just have a const reference to it, which tends to be often.

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+	 assert(r.value() != nullptr);
+	 T *v = r.value();
+	 return v;
+   };
+   bool check(){
+	 return vec->check(ind);
+   }
+   void destroy(){
+	 vec->remove(ind);
+   }
```
BMagnu, line 403: Not sure if I like gRef. It encourages users to keep these over just the pool indices, and if you have a list of gRefs you're just wasting storage by keeping the pointer to the pool multiple times. Also, since it doesn't provide all overloads of the pool access itself and does everything through pointers, it'll be slower and less optimizable

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+};
+
+
+
+template  <typename T>
+struct gEntry {
+	int generation;
+	optional<T> stored;
+
+};
```
BMagnu, line 61: gEntry looks to be purely internal to the pool, so it should probably be an internal subclass of pool. That way, since you also always have the same T as the pool itself, you don't need it to be a template.

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
+#pragma once
+
+#include "globalincs/pstypes.h"
+#include <tl/optional.hpp>
+#include <csignal>
+//#include <vector>
+
+using  tl::optional;
+struct pool_index{
```
BMagnu, line 9: Can we call it SCP_pool_index, just to keep the naming consistent with the pool itself?

BMagnu (changes requested):
Okay, I've had a look at this.
Errm, I may have been a bit strict, but since this is a very cool thing to have for the SCP in wide usage as a generic container, I'd want it to be as polished as possible.

code/globalincs/pool.h
EatThePath: for myself ergonomics overcomes consistency in this case. pool_index is already more than I'd like to be typing for this, honestly. Since it's going to be used all over the damn place in the end I'm more than willing to bend to consensus on this one, if one can be established.

code/globalincs/pool.h
EatThePath: You're right, it should probably even be a private subclass.

code/globalincs/pool.h
EatThePath: gRef is vestigial from something I was trying earlier on and got talked out of by blount. I didn't fully remove it because I'm considering trying to do a proof of concept/final pitch of the idea as an alternative version of this branch, but my drive to do that is waning

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+	~PoolExceptionFull() noexcept override = default;
+};
+
+
+
+
+using std::vector;
+template  <typename T>
+//rules:
+//Caps can only ever be set at the current size or smaller
```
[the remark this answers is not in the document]
EatThePath: right, bit of a brain slip while writing that comment

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+		if (i.null())
+			throw PoolExceptionNullIndex("null index SCP_Pool access attempted");
+		if (i.i() >= storage.size())
+			throw PoolExceptionOutOfRange("out of range SCP_Pool access attempted");
+		if (storage[i.i()].generation!= i.g())
+			throw PoolExceptionStaleIndex("Stale SCP_Pool access attempted");
+		if(storage[i.i()].stored==tl::nullopt)
+			throw PoolExceptionNullValue("Empty SCP_Pool access attempted");
+//		v = (storage[i.i()].stored).operator->();
+		return storage[i.i()].stored.operator*();
```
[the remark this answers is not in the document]
EatThePath: if it works, yeah. Syntax there has been challenging at points, I'll give it a shot

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+		if (i.i() >= storage.size()) {
+			return tl::nullopt;
+		}
+		if (storage[i.i()].generation!= i.g()){
+			return tl::nullopt;
+		}
+		optional<T> s = storage[i.i()].stored;
+		T* p = s.operator->();
+		return p;
+	};
```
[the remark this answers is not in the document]
EatThePath: The one objection that comes to mind immediately is, what if someone wants to make a pool of pointers for some reason? in that case storing nullptr and having an empty/stale slot might be distinct conditions you'd want to handle differently

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+		if(i.null())
+			return false;
+		auto * t = &operator[](i);
+		return (t!=nullptr);
+	}
+
+
+
+	//iterator shit.
+	struct iter{
```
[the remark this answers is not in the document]
EatThePath: this and many of following comments all fall under the scope of `The code in pool.h is definitely not done, nor very presentable`. Trying to find good documentation and examples on properly implementing an iterator was somewhat bewildering, and so this is in 'whatever, make it work' state right now. Still, it's useful to have the pointers on what needs to be done :)

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
+#pragma once
+
+#include "globalincs/pstypes.h"
+#include <tl/optional.hpp>
+#include <csignal>
```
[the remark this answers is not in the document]
EatThePath: once again that pesky inclusion addon and overzealous intellisense messing things up... will clean.

code/globalincs/pool.h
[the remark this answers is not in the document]
BMagnu: >  what if someone wants to make a pool of pointers for some reason
In that case, if you store a T*, then this'll return an optional<T**>. 
If you wanted direct-to-pointer access, you basically need an override for this function that changes how it behaves using enable_if depending on the type of T.
So basically, you'd still get a pointer to the element, and not the pointer-element itself. Hence, a nullptr still can only mean that the element isn't valid, not that a nullptr was stored.

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+		if(rhs.is_null)
+			nullify();
+		else
+			set(rhs.index, rhs.generation);
+		return *this;
+	};
+	int i() const { return index;};
+	int g() const { return generation;};
+	bool null() const { return is_null;};
+	bool has_value() const { return !is_null;};
```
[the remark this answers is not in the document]
The-E: Given that `inline` is, at best, interpreted as a weak suggestion by compilers, this seems unnecessary to me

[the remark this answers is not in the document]
EatThePath: Documentation I found about inline also suggests that these would be treated as marked by it anyway... but I put it on anyway as it might be useful as a hint to future programmers.

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+		}
+		optional<T> s = storage[i.i()].stored;
+		T* p = s.operator->();
+		return p;
+	};
+
+	optional<pool_index> get_new(){
+		T* n = new T();
+		return store(*n);
+	};
```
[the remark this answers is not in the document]
EatThePath: I thought I had tried it without now and it errored, but trying again it works fine. 
Leaving unresolved for now because forwarding is a good idea.

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+		};
+		iter& operator++(){
+			inner_position++;
+			//we want to skip empty slots
+			while(inner_position < storage->size()
+					&& storage->at(inner_position).stored == tl::nullopt 
+				  ){
+				inner_position++;
+				//the while condition isn't short-circuiting so we're getting invalid accesses.
+				if(inner_position==storage->size()) break;
```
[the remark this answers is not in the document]
EatThePath: Well, the comment documents what I thought I was seeing. I suppose I should loop back at some point and reverify it down the line to figure out if it's real or I was crossed up somehow.

code/globalincs/pool.h
EatThePath: deleted for now

code/globalincs/pool.h
EatThePath: I've been trying to figure out where I got in my head to do this and failed so far. It was either some guidance I read somewhere about doing range-based for loop support, or just trying to minimize the change in client code.
To the latter, if I follow your guidance on return types for `*` and `->`, then in trails.cpp when I need to get a trail pointer out of an iterator, the syntax becomes `trail * trailp=iter.operator->();` or alternatively `trial *trailp = &*iter;` works. If I make `operator&()` return a pointer to the current item then `trail * = &iter;` it works just fine, at least. Not sure if that's a safe practice in this case? Certainly makes it hard to pass around pointers to an iterator, I suppose. For the working branch I'm now using the `&*` in the relevant places

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+			   else{
+				 full_position.nullify();
+			   }
+			}
+			else{full_position.nullify();}
+			return *this;
+		};
+		iter& operator++(int i){
+			operator++();
+			return *this;
```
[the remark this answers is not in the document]
EatThePath: manually added this change to my working branch

code/globalincs/pool.h
```diff
@@ -0,0 +1,404 @@
...
+//Caps can only ever be set at the current size or smaller
+//shrinking the vector would lead to forgetting the generation counts and then potential slot reuse if allowed to expand later.
+class SCP_Pool {
+	/**************************************************************************
+	 * Private data members
+	 *
+	 *************************************************************************/
+	vector<gEntry<T>> storage;
+	vector<size_t> known_empty;
+
```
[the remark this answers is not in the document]
BMagnu: Probably needs to be ``!std::is_reference<T>::value``, since the () operator is C++14

code/globalincs/pool.h
```diff
@@ -0,0 +1,374 @@
...
+	tl::optional<pool_index> get_new(){
+		T n = T();
+		return store(n);
+	};
+
+	tl::optional<pool_index> get_new(T&& args){
+		T n = T(std::forward<T>(args));
+		return store(n);
+	};
```
BMagnu, line 245: Close, but now you still expect someone to construct the element for you (and forwarding that), instead of forwarding the args themselves to the constructor.
Effectively, I'd expect something like:
```cpp
template<typename... args_t>
tl::optional<pool_index> get_new(args_t&&... args){
	return store(T(std::forward<args_t>(args)...));
};
```

code/globalincs/pool.h
BMagnu: So, ``&*iter`` would in fact be the preferred way of getting a pointer to a trail.
_However_, as it is a pointer to something internally stored in a vector, it's not allowed to live longer than a single frame anyways, in which case its fine that it looks ugly, cause you really should be using references unless its for interfacing with legacy code that is built on pointers.

BMagnu (commented):
Yup, iterator looks better now, apart from the comments that are yet unresolved

code/globalincs/pool.h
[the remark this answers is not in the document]
EatThePath: Looks like it works

code/globalincs/pool.h
EatThePath: Alright, unless I encounter a compelling reason to want to bring it back I'll remove this overload.

PR #5122, opened 2023-01-17
make collision_group_id accessible to parse objects

Goober5000:
Add collision groups to parse objects so they can be set before arrival via script.


wookieejedi (approved):
Useful feature and overall LGTM

PR #5125, opened 2023-01-17
Toggle auto-assigning personas

MjnMixael:
Adds a global toggle for the persona auto-assignment feature. My mod does not have any personas that aren't tied to specific characters or events so this global mod just blanket turns the feature off. Ships without a persona will never send built-in messages.


BMagnu (approved):

code/mod_table/mod_table.cpp
```diff
@@ -1136,6 +1137,10 @@ void parse_mod_table(const char *filename)
 			stuff_boolean(&Hotkey_always_hide_hidden_ships);
 		}
 
+		if (optional_string("$Allow automatically assigning personas:")) {
+			stuff_boolean(&Auto_assign_personas);
```
Goober5000, line 1141: Almost every boolean flag in game_settings or ai_profiles is something that defaults to false and is optionally set to true.  This should be consistent with that, e.g. `$Do not automatically assign personas:`

Goober5000 (changes requested):

code/mod_table/mod_table.cpp
MjnMixael: When you go adjust `$AI use framerate independent turning:` and the few ai_profiles flags I found to make this a true statement across the board then I'll bother with this. Till then, don't merge it, I guess. I don't have time for petty OCD stuff.

Goober5000: The fact that a standard has not been met in 100% of cases is not a valid reason to abandon the standard.

Goober5000: That's an interesting way of looking at it.  But the original pattern was not "do something that isn't normally done" vs "stop doing something that's normally done", but rather "does this alter the base game behavior".  That produces an easy way to check flags, both in mods and in the code: if it alters the behavior, it's true; if it doesn't alter it, it's false.

Goober5000: Any other coders feel like weighing in?

wookieejedi: Hmm I can see both, if we are favoring modders though having the logic match the flag would be the most useful for modders IMO.

PR #5127, opened 2023-01-18
don't blast no-collide objects

### LINKED ISSUE #5068: Derelict markers are blasted by asteroids
From Discord:
> Campaign: Derelict
> Its the 2nd mission where you gotta deploy the nav beacons
> The markers (the green circles) fly off when an asteroid either touches it or is blasted near it. It took me a few minutes to progress cuz the circle flew away many times.
> Doesnt happen when the ship flies into it tho. Just blasts that makes it fly away

This should be fixable by searching for `// don't blast navbuoys` in the codebase and adding a check for the no-collide object flag at the top of the loop.  There are five such loops in the code.

Later on these area-effect loops should probably be consolidated, but that can be done after 23.0.

Goober5000:
In the same way that navbuoys aren't blasted by explosions, no-collide objects shouldn't be blasted either.

Fixes #5068.

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -1579,8 +1579,8 @@ static void asteroid_do_area_effect(object *asteroid_objp)
 		if (ship_objp->flags[Object::Object_Flags::Should_be_dead])
 			continue;
 
-		// don't blast navbuoys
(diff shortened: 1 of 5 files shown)


wookieejedi (approved):
LGTM

PR #5128, opened 2023-01-18
Remove nebula bitmap limit

MjnMixael:
Removes MAX_NEB2_BITMAPS which was previously just 10; a limit I hit just but modularizing the MediaVPs nebula.tbl which only adds three bitmaps to retail's 8. Instead of simply raising the limit this PR vectorizes the bitmap list. Bitmaps are only ever referenced by name (unlike poofs which use flags) so it makes sense to just make this a quick vector of strings.

Sneaks in one small QoL parsing feature to check for end of file in a modular table so that -neb.tbms which don't specify poofs won't require two #end lines.

### DIFF
--- a/fred2/bgbitmapdlg.cpp
+++ b/fred2/bgbitmapdlg.cpp
@@ -278,9 +278,9 @@ void bg_bitmap_dlg::create()
 		m_skybox_heading = m_skybox_heading + 360;
 
 
-	for(i=0; i<MAX_NEB2_BITMAPS; i++){
-		if(strlen(Neb2_bitmap_filenames[i]) > 0){ //-V805
-			((CComboBox*)GetDlgItem(IDC_NEB2_TEXTURE))->AddString(Neb2_bitmap_filenames[i]);
+	for (i = 0; i < (int)Neb2_bitmap_filenames.size(); i++) {
+		if (Neb2_bitmap_filenames[i].length() > 0) {
+			((CComboBox*)GetDlgItem(IDC_NEB2_TEXTURE))->AddString(Neb2_bitmap_filenames[i].c_str());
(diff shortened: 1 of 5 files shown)


fred2/bgbitmapdlg.cpp
```diff
@@ -278,9 +278,9 @@ void bg_bitmap_dlg::create()
 		m_skybox_heading = m_skybox_heading + 360;
 
 
-	for(i=0; i<MAX_NEB2_BITMAPS; i++){
-		if(strlen(Neb2_bitmap_filenames[i]) > 0){ //-V805
-			((CComboBox*)GetDlgItem(IDC_NEB2_TEXTURE))->AddString(Neb2_bitmap_filenames[i]);
+	for (i = 0; i < (int)Neb2_bitmap_filenames.size(); i++) {
+		if (Neb2_bitmap_filenames[i].length() > 0) {
```
Baezon, line 282: Very minor, but I'm wonder if all these `len > 0` checks should be kept. It was only like that because before, that's what the whole array was initialized with, forcing it to assume "" was an unused slot.  Now, "" as a name could be technically valid, but obviously that's a terrible idea...

Baezon (approved):

fred2/bgbitmapdlg.cpp
MjnMixael: Hmm. While technically valid I think `stuff_string()` will error on an empty string. So yeah, I think it's impossible to have an item in the list with a name of zero length. These can probably go.

MjnMixael: Tested and parsing fails on an empty filename string, so I have removed the length checks.

PR #5130, opened 2023-01-19
Better account for runtime gravity changes

Baezon:
Whether a weapon has `PF_CONST_VEL` or `PF_BALLISTIC` depends on whether there is gravity *at the moment*, not just when it was fired.

### DIFF
--- a/code/object/objcollide.cpp
+++ b/code/object/objcollide.cpp
@@ -1042,3 +1042,29 @@ void obj_sort_and_collide(SCP_vector<int>* Collision_list)
 	}
 	obj_find_overlap_colliders(sort_list_y, sort_list_z, 2, true);
 }
+
+void collide_apply_gravity_flags_weapons() {
+	for (int i = 0; i < MAX_OBJECTS; i++) {
+		object* obj = &Objects[i];
+		if (Objects[i].type != OBJ_WEAPON)
+			continue;
+
+		weapon* wp = &Weapons[Objects[i].instance];
+		weapon_info* wip = &Weapon_info[wp->weapon_info_index];
+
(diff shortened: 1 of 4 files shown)


code/object/objcollide.cpp
```diff
@@ -1042,3 +1042,29 @@ void obj_sort_and_collide(SCP_vector<int>* Collision_list)
 	}
 	obj_find_overlap_colliders(sort_list_y, sort_list_z, 2, true);
 }
+
+void collide_apply_gravity_flags_weapons() {
+	for (int i = 0; i < MAX_OBJECTS; i++) {
```
Goober5000, line 1047: Why not iterate through `obj_used_list` or `Missile_obj_list`?

I suppose if primaries can home now, that rules out the second; but the first should be an option.

Baezon: Ah, true, I would've wanted a `weapon_used_list`, but `obj_used_list` is better for what I ended up going with.

code/object/objcollide.cpp
```diff
@@ -1042,3 +1042,28 @@ void obj_sort_and_collide(SCP_vector<int>* Collision_list)
 	}
 	obj_find_overlap_colliders(sort_list_y, sort_list_z, 2, true);
 }
+
+void collide_apply_gravity_flags_weapons() {
+	for (object* obj = GET_FIRST(&obj_used_list); obj != END_OF_LIST(&obj_used_list); obj = GET_NEXT(obj)) {
+		if (obj->type != OBJ_WEAPON)
```
Goober5000, line 1048: you'll want to skip objects with the should-be-dead flag

(I don't think it matters here, but just for consistency)

Goober5000 (changes requested):

Goober5000 (approved):

Issue #5131, opened 2023-01-19
Asteroid Gravity is sorta half baked


MjnMixael:
There are a few issues that need to be solved and there are a number of ways to go about it.

1) Active asteroid fields don't obey gravity at all. That may be by design.

2) Passive asteroid fields do obey gravity. However, the gravity will pull the asteroids outside the bounding box. Asteroid wrapping only happens for active fields so passive asteroids will just be pulled off into the distance forever and your asteroid field is gone.

3) The current asteroid wrapping functions explicitly do not apply to passive fields. Perhaps a new function is the best method here? One that only applies to passive fields if gravity is active.

Baezon:
1. I actually didn't realize this, it is definitely not by design.
2/3. It sounds like passive asteroid wrapping may be the best approach here, probably as an option, but otherwise it's working as expected.

Asteroid gravity was largely implemented for the sake of completeness more than anything else, because I don't really know what the design should be for asteroids with gravity. 'Stay inside this box' and 'fall' are fundamentally incompatible, even with wrapping, wrapping doesn't happen if the player is looking at them, it's meant to be something hidden from the player.

MjnMixael:
I toyed around with letting passive fields wrap and have a working PR. I expect getting active fields to obey gravity will be more complex because they are functionally similar to homing weapons.

Baezon:
They shouldn't be, active asteroids are aimed, but they don't home tmk. That said, even if it isn't that hard, maybe it's still not such a great idea, since they still kind of have this abstract notion 'other forces at play which would be interfered with by gravity'. Throwing asteroids with ballistic trajectories would be kind of silly.

Goober5000:
@Baezon At this point, can this be closed?

Baezon:
Given the way the issue has been recontextualized, I think this is a reasonable request and can stay open. It shouldn't be too difficult, but it isn't high on my priority list atm.

MjnMixael:
It's been 3 years and this isn't something I personally need anymore, so I'm going to close this.

PR #5132, opened 2023-01-19
Let passive fields wrap if gravity is involved

MjnMixael:
Handles items 2 & 3 in issue #5131 . This PR lets passive asteroids wrap if gravity is involved. Could be made toggle-able if that's desired but my opinion is this is the best way to keep asteroids in the box while also obeying the designers environment rules. Without wrap, the asteroids will eventually get beyond the box and the flyable mission space because there is no code to remove them otherwise.

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -1073,8 +1073,8 @@ void asteroid_delete( object * obj )
  */
 static void asteroid_maybe_reposition(object *objp, asteroid_field *asfieldp)
 {
-	// passive field does not wrap
-	if (asfieldp->field_type == FT_PASSIVE) {
+	// passive field does not wrap if there is no gravity
+	if (IS_VEC_NULL(&The_mission.gravity) && (asfieldp->field_type == FT_PASSIVE)) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


code/asteroid/asteroid.cpp
```diff
@@ -1100,31 +1100,38 @@ static void asteroid_maybe_reposition(object *objp, asteroid_field *asfieldp)
...
 					dot = vm_vec_dot(&Eye_matrix.vec.fvec, &vec_to_asteroid);
 					
-					if ( (dot > 0.7f) && (dist < (asfieldp->bound_rad * 1.3f)) ) {
-						// player would see asteroid pop out other side, so reverse velocity instead of wrapping						
-						objp->pos = old_asteroid_pos;		
-						vm_vec_copy_scale(&objp->phys_info.vel, &old_vel, -1.0f);
-						objp->phys_info.desired_vel = objp->phys_info.vel;
-						Asteroids[objp->instance].target_objnum = -1;
+					//probably don't reverse direction if there's gravity involved
+					if (IS_VEC_NULL(&The_mission.gravity)) {
```
Baezon, line 1116: Would this not also allow the asteroid to leave the box? Since it wouldn't be wrapping or changing velocity.

Baezon (changes requested):

code/asteroid/asteroid.cpp
Baezon: Whoops, misread, undoing of the wrapping is not done if this is skipped. This would result in a possibly visible 'wrap' though.

MjnMixael: From what I understand, that will only happen if it was being targeting or looked at. Both are ways to get asteroids to leave the box even in retail. There is no hard culling of asteroids outside the box in any circumstance so stubborn players can make it happen no matter what.

Baezon (approved):

PR #5133, opened 2023-01-20
Lua method to multiply two orientations

MjnMixael:
What it says on the tin.


Baezon (changes requested):
Err, lua matrices can be multiplied the typical way, `local new_orient = orient1 * orient2`

MjnMixael:
Well that's news to me

PR #5134, opened 2023-01-21
Parse and save asteroids by name

MjnMixael:
Since asteroids are modular now we should parse and save them by name instead of by index.

This also cleans up a duplicate asteroid.cpp function.


code/mission/missionparse.cpp
```diff
@@ -5595,35 +5595,61 @@ void parse_asteroid_fields(mission *pm)
...
+				}
+			}
+
+			// Get asteroids by name
+			for (int j = 0; j < MAX_ACTIVE_DEBRIS_TYPES; j++) {
+				if (optional_string("+Field Debris Type Name:")) {
+					SCP_string ast_name;
+					stuff_string(ast_name, F_NAME);
+					int subtype = get_asteroid_index(ast_name.c_str());
+					// If the returned index is valid but not one of the firs three then it's a debris type instead of asteroid
```
Goober5000, line 5645: `firs` is a typo

Goober5000 (approved):

code/mission/missionparse.cpp
MjnMixael: Typo is fixed. I could go either way on this one's inclusion in 23.0 but it should be at least considered to prevent modular asteroids being loaded in different orders for the few mods that may take advantage of the feature.

Goober5000: I ran it by the team.  We considered it, but the consensus was that we should hold off until the stable release.

MoerasGrizzly:
I take it that this'll break backwards compatability right?

Goober5000:
For missions, yes, due to the format change.  But I have a feature in progress that will warn about that.

Goober5000 (changes requested):
#5152 has been added which will handle the change in version.  Rebase or merge this PR to include the changes from master, and then update `MISSION_VERSION` in missionparse.cpp to be 
```
gameversion::version(23, 1, 0);
```
Then we can merge this tomorrow.

wookieejedi:
Looks like this is good to merge now?

Goober5000:
Yup.

Goober5000 (approved):

PR #5138, opened 2023-01-24
hud-force-emp-effect sexp

### LINKED ISSUE #5103: FR: an "EMP-effect" sexp
Just an EMP-effect that fizzles your HUD with modifiable time and intensity.
Probably best added into the "cutscene" category.

MjnMixael:
Adds a sexp to force the emp effect on the player. Allows for customizable intensity, duration, and hud gauge warning.

Fixes #5103


code/parse/sexp.cpp
```diff
@@ -39013,6 +39065,14 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"\t1:\tBoolean, whether or not to enable sensor static\r\n"
 	},
 
+	{OP_HUD_FORCE_EMP_EFFECT, "hud-force-emp-effect\r\n"
+		"\tActivates or deactivates emp effect for the player."
+		"Takes 2 or more Arguments...\r\n"
+		"\t1:\tNumber, emp intensity (0 to 500)\r\n"
+		"\t2:\tNumber, emp duration in seconds\r\n"
```
Goober5000, line 39072: Millisecond durations would be useful.  It can just be divided by 1000 before being passed to `emp_start_local`.

code/weapon/emp.h
```diff
@@ -87,7 +87,7 @@ void emp_process_ship(ship *shipp);
 
 // start the emp effect for MYSELF (intensity == arbitrary intensity variable, time == time the effect will last)
 // NOTE : time should be in seconds
-void emp_start_local(float intensity, float time);
+void emp_start_local(float intensity, float time, SCP_string text = "");
```
Goober5000, line 90: Use `const SCP_string &` to avoid unnecessary string copies.

code/weapon/emp.cpp
```diff
@@ -412,8 +412,13 @@ void emp_start_local(float intensity, float time)
...
 	}
 
-	// start the emp icon flashing
-	hud_start_text_flash(XSTR("Emp", 1670), 5000);
+	// if text is not set then use the default "Emp" string
+	if (text.length() == 0)
+		text = XSTR("Emp", 1670);
+
+	// start the emp icon flashing if text is not "none"
+	if (text != "none")
```
Goober5000, line 420: This should use a case-insensitive comparison

Goober5000 (changes requested):

code/parse/sexp.cpp
MjnMixael: what would be the point? `emp_start_local` takes seconds so it'd be rounded anyway.

Goober5000: It takes seconds, but the seconds are specified as a `float`.  So passing it 1.5 should be valid.

code/weapon/emp.h
MjnMixael: This cannot be const or I can't copy the Emp string into it if length is 0

MjnMixael: Eh whatever. I don't care that much. I'll just add additional unnecessary code to make you happy.

code/parse/sexp.cpp
```diff
@@ -12299,6 +12300,43 @@ void sexp_hud_force_static(int n)
...
+			if (!stricmp(Messages[i].name, text.c_str())) {
+				text = Messages[i].message;
+				break;
+			}
+		}
+
+		text = message_translate_tokens(text.c_str());
+	}
+
+	emp_start_local(i2fl(intensity), i2fl(duration/1000), text);
```
Goober5000, line 12337: The division by 1000 should be done outside the parentheses, after the cast to float.  Inside the parentheses, it's integer division and will be rounded.
```suggestion
	emp_start_local(i2fl(intensity), i2fl(duration)/1000, text);
```

Goober5000 (changes requested):

code/weapon/emp.h
Goober5000: A cascading if() statement would avoid the need for copying the Emp string.

code/parse/sexp.cpp
MjnMixael: done

Goober5000 (approved):

PR #5139, opened 2023-01-24
has-armor-type sexp

### LINKED ISSUE #2854: FR: a SEXP to check the currently applied armor of an object
Just as is say, I would like to request the inclusion of a SEXP that lets you check which armor class is currently applied to a ship, please.

This is for gameplay effects that use dynamic hull and shield armor changes, e.g. entering and leaving field effects that interfere with shields or active armor (buff or debuff area of effect basically).
Currently you have to work around and use a ship flag as proxy check.

MjnMixael:
Adds the sexp to check the current armor of a ship or ship's subsystem. Fixes #2854

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -279,6 +279,7 @@ SCP_vector<sexp_oper> Operators = {
 	{ "has-been-tagged-delay",			OP_HAS_BEEN_TAGGED_DELAY,				2,	INT_MAX,	SEXP_BOOLEAN_OPERATOR,	},
 	{ "are-ship-flags-set",				OP_ARE_SHIP_FLAGS_SET,					2,	INT_MAX,	SEXP_BOOLEAN_OPERATOR,	},	// Karajorma
 	{ "are-wing-flags-set",				OP_ARE_WING_FLAGS_SET,					2,	INT_MAX,	SEXP_BOOLEAN_OPERATOR, },	// Goober5000
+	{ "has-armor-type",					OP_HAS_ARMOR_TYPE,						3,	3,			SEXP_BOOLEAN_OPERATOR, },	// MjnMixael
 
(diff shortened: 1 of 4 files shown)


Goober5000 (changes requested):

code/parse/sexp.cpp
```diff
@@ -25197,6 +25198,52 @@ int sexp_is_in_mission(int node)
...
+			if (ss == nullptr) {
+				continue;
+			}
+
+			currentArmor = ss->armor_type_idx;
+		}
+	}
+
+	if (currentArmor == armor)
+		return SEXP_TRUE;
```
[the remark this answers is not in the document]
MjnMixael: done

code/parse/sexp.cpp
```diff
@@ -25197,6 +25198,59 @@ int sexp_is_in_mission(int node)
...
+	// get armor from sexp
+	int armor;
+	if (!stricmp(SEXP_NONE_STRING, CTEXT(node))) {
+		armor = -1;
+	} else {
+		armor = armor_type_get_idx(CTEXT(node));
+
+		//warn and return if the armor is not found
+		if (armor < 0) {
+			Warning(LOCATION, "Armor %s not found on ship %s", CTEXT(node), shipp->ship_name);
```
Goober5000, line 25223: The warning message is not quite correct... the armor is not recognized as an armor type.  Yes, it's not found on the ship, but more fundamentally, it's not a valid entry in the armor table.

Goober5000 (changes requested):

code/parse/sexp.cpp
MjnMixael: done

Goober5000 (approved):

Goober5000:
0rph3u5 expressed his appreciation:

> please relay my thanks to everyone involved 👍

PR #5141, opened 2023-01-24
Allow way to prevent scpui loading hires animations

MjnMixael:
By default SCPUI/librocket follows FSO's trend of substituting `2_animname.ext` for `animname.ext` if the game's resolution is larger than a certain amount. SCPUI can correctly scale animations dynamically so this behavior isn't required anymore. 

The substitution can be a problem in the case of naming conflicts with upstream files. A local mod may have a hires cb_default but the game hands over the retail 2_cb_default instead. That is not ideal. However, there may still be situations (retail/mediavps) where it would be preferred to get the hires filename in place of the normal one. So this PR allows a quick toggle to enable or disable the behavior for SCPUI.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -131,6 +131,7 @@ bool Always_use_distant_firepoints;
 bool Discord_presence;
 bool Hotkey_always_hide_hidden_ships;
 bool Use_weapon_class_sounds_for_hits_to_player;
+bool SCPUI_loads_hi_res_animations;
 
 static auto DiscordOption = options::OptionBuilder<bool>("Other.Discord", "Discord Presence", "Toggle Discord Rich Presence")
 							 .category("Other")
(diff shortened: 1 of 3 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -765,6 +766,10 @@ void parse_mod_table(const char *filename)
 
 		}
 
+		if (optional_string("$SCPUI attempts to load hires animations:")) {
```
Goober5000, line 769: It's conceivable mods that don't use SCPUI may want to use this option, so I'd use `$Attempt to load hi-res animations:` or something similar.

MjnMixael: SCPUI refers to all the UI stuff available through librocket and this setting exclusively alters the way librocket libraries load animations for streaming.

Goober5000: I was thinking about all the other uses of `generic_anim_init_and_stream`, not just the one in RocketRenderingInterface.cpp.  And all the other places `2_` is used.

However, point taken, that's unnecessarily generalizing a very specific feature.

Goober5000 (approved):

PR #5145, opened 2023-01-26
Flag to apply hud shudder to chase view

### LINKED ISSUE #3563: Feature Gap: Screen Shake not allowed in 3rd person view
Mods should be able to force screen shake to also occur in third person camera views.

I haven't taken a look at the code, but I anticipate this to be an easy addition.

MjnMixael:
Quick flag to apply hud shudder (afterburner and engine wash) also to chase view. Does not override the command line toggle. Fixes #3563

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -2935,8 +2935,13 @@ void apply_view_shake(matrix *eye_orient)
 	tangles.h = 0.0f;
 	tangles.b = 0.0f;
 
-	// do shakes that only affect the HUD (unless disabled by cmdline in singleplayer)
-	if (Viewer_obj == Player_obj && (!Cmdline_no_screenshake || Game_mode & GM_MULTIPLAYER)) {
+	bool do_hud_shudder = (Viewer_obj == Player_obj);
+
+	if ((Viewer_mode & VM_CHASE) && !Apply_shudder_to_chase_view)
+		do_hud_shudder = true;
+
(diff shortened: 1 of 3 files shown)


freespace2/freespace.cpp
```diff
@@ -2935,8 +2935,13 @@ void apply_view_shake(matrix *eye_orient)
 	tangles.h = 0.0f;
 	tangles.b = 0.0f;
 
-	// do shakes that only affect the HUD (unless disabled by cmdline in singleplayer)
-	if (Viewer_obj == Player_obj && (!Cmdline_no_screenshake || Game_mode & GM_MULTIPLAYER)) {
+	bool do_hud_shudder = (Viewer_obj == Player_obj);
+
+	if ((Viewer_mode & VM_CHASE) && !Apply_shudder_to_chase_view)
```
BMagnu, line 2940: Shouldn't this be ``&& Apply_shudder_to_chase_view`` instead? 
At least by name, the flag suggests that it explicitly enables shudder in chase mode, not the other way around

BMagnu (changes requested):

freespace2/freespace.cpp
MjnMixael: Yes it should. I remember testing this one pretty thoroughly.. I must have been cleaning up some code and that snuck in. Built and tested just now to double check.

BMagnu (approved):

PR #5149, opened 2023-01-27
Refactor Msg_scrollback_lines

### LINKED ISSUE #3662: Scrollback newlines are queried with the wrong font
When messages are added to the scrollback, they get newlines inserted based on their width. This calculation of width happens with the currently selected font (in ``hud_add_msg_to_scrollback``, in hudmessage.cpp), regardless of which font is used to actually draw the respective message.
Observed by @JadedDragoon with AxMessage, since that script changes around fonts. This results in the training messages, rendered via scrollback apparently, to have incorrect linebreaks.

MjnMixael:
Ended up having a couple hours free, so I went at it. Refactor's handling of message scrollback. Changes the linked list to a vector which removes the soft limit of 1000 message lines. Generally just cleans up the code. This PR also moves line splits to init of the scrollback game state instead of when a message is added to scrollback. 

I tested scrollback with 0 messages, with more messages than can be displayed at once (requiring scrolling), and messages with/without needing to be split (once or more than once).

Things to note:
1) Msg_scrollback_vec contains unsplit scrollback messages. This will be later used for SCPUI access.
2) Msg_scrollback_lines contains the actual split lines for display and should only be valid when scrollback game state is active.

This possibly resolves #3662 if that was referring to line splits in the UI rather than in the little blue rendering box seen during a mission because the font can be reset on the game state change and the line splits will be done based on the currently active font. I didn't thoroughly test this because there was no clear example in the ticket to try and reproduce. My local tests showed proper line splits throughout.

### DIFF
--- a/code/hud/hudmessage.cpp
+++ b/code/hud/hudmessage.cpp
@@ -133,13 +133,8 @@ SCP_vector<HUD_message_data> HUD_msg_buffer;
 
 int HUD_msg_inited = FALSE;
 
-// There is a maximum number of lines that will be stored in the message scrollback.  Oldest
-// messages are deleted to make way for newest messages.
-#define MAX_MSG_SCROLLBACK_LINES	1000
-line_node Msg_scrollback_lines[MAX_MSG_SCROLLBACK_LINES];
-
-line_node Msg_scrollback_free_list;
-line_node Msg_scrollback_used_list;
+SCP_vector<line_node> Msg_scrollback_vec;
+static SCP_vector<line_node> Msg_scrollback_lines;
 
 #define	MAX_HUD_FT	1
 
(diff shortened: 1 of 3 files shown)


code/hud/hudmessage.cpp
```diff
@@ -742,34 +652,28 @@ int hud_get_scroll_max_pos()
...
-		if (EMPTY(&Msg_scrollback_used_list) || !HUD_msg_inited) {
+		if (Msg_scrollback_lines.size() <= 1 || !HUD_msg_inited) {
 			max = 0;
-
 		} else {
-			ptr = GET_FIRST(&Msg_scrollback_used_list);
-			while (ptr != END_OF_LIST(&Msg_scrollback_used_list)) {
-				if (ptr->source != HUD_SOURCE_HIDDEN) {
+
+			for (int i = 0; i < (int)Msg_scrollback_lines.size(); i++) {
```
The-E, line 662: The int-casting can be avoided like so:

```suggestion
			for (size_t i = 0; i < Msg_scrollback_lines.size(); i++) {
```

code/hud/hudmessage.cpp
```diff
@@ -923,20 +822,69 @@ void hud_scrollback_button_pressed(int n)
...
+
+			int max_width = Hud_mission_log_list2_coords[gr_screen.res][2];
+			if (width > max_width) {
+				char c_text[HUD_MSG_LENGTH_MAX];
+				strcpy_s(c_text, node_msg.text.c_str());
+
+				char* text = c_text;
+				char* split = nullptr;
+
+				split = split_str_once(text, max_width);
```
The-E, line 850: I think declaration and first assignment can be merged here
```suggestion
				char* split = split_str_once(text, max_width);
```

The-E (changes requested):

code/hud/hudmessage.cpp
MjnMixael: Right. Done.

code/hud/hudmessage.cpp
MjnMixael: and done

The-E (approved):

PR #5154, opened 2023-01-29
Optimizing the trail renderer by avoidng repeated batch_find calls

EatThePath:
Building large trails is an annoyingly costly process in terms of frame times. This makes it a very little bit cheaper. Not a big enough optimization to have very visible effects, but one I think is worth doing.

### DIFF
--- a/code/render/batching.cpp
+++ b/code/render/batching.cpp
@@ -874,8 +874,8 @@ void batching_add_quad(int texture, vertex *verts)
 
 void batching_add_quad(int texture, vertex *verts, primitive_batch *batch)
 {
+	Assertion((texture >= 0), "batching_add_...() attempted for invalid texture");
 	if ( texture < 0 ) {
-		Int3();
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


Baezon:
A different idea of doing the same thing that is more extensible: what if the batching code simply had a global of the last-used texture id and batch, and if the same texture id is used, use the last used batch as well, since that's what the batch is based on right?

EatThePath:
> A different idea of doing the same thing that is a little more extensible: what if the batching code simply had a global of the last-used texture id and batch, and if the same texture id is used, use the last used batch as well, since that's what the batch is based on right?

That would probably work, and it'd probably be faster than the status quo, but it would still be a test and possible branch executed on each segment of every trail. Maybe it'd optimize away, but I wouldn't trust it to be quite as fast.

Still, might be worth doing for general batch streamlining. It'd need to check the batch info as well as the texture info, I think.

code/render/batching.cpp
```diff
@@ -872,6 +872,15 @@ void batching_add_quad(int texture, vertex *verts)
 	batching_add_quad_internal(batch, texture, verts);
 }
 
+void batching_add_quad(int texture, vertex *verts, primitive_batch *batch)
+{
+	if ( texture < 0 ) {
+		Int3();
```
JohnAFernandez, line 878: We have `verify()` and `verification()` that can be used to replace `int3()`.

JohnAFernandez (changes requested):

JohnAFernandez (approved):

PR #5157, opened 2023-01-30
Refactor mission log scrollback

MjnMixael:
Reorganizes how the mission log scrollback is initialized and drawn. Instead of separate linked lists for timestamp and entries, this uses a single vector for each line. The vector entry holds all the required information required for drawing, including text, color, etc. This will make SCPUI access much easier in a follow-up PR.

Tested viewing the scrollback with a small log (no scrolling required), a large log, and with as many different entry types as I could reasonably come up with. Each type of log entry seems to work exactly as it did before.

### DIFF
--- a/code/mission/missionlog.cpp
+++ b/code/mission/missionlog.cpp
@@ -716,25 +716,25 @@ void message_log_init_scrollback(int pw)
 		thisEntry.timestamp = entry->timestamp;
 
 		// keep track of base color for the entry
-		int color;
+		int thisColor;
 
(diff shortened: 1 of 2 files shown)


wookieejedi (approved):
My testing shows this all works well, and overall looks good. Also provides some useful utility in that once it is merged, it will make issues such as #5187 easier to address. Also noting that it is always useful to have another set of eyes on the code, too.

The-E (approved):

PR #5158, opened 2023-01-30
Add virtvars for goal satisfaction and score

MjnMixael:
A couple more goal related virtvars I'll need for SCPUI

### DIFF
--- a/code/scripting/api/objs/briefing.cpp
+++ b/code/scripting/api/objs/briefing.cpp
@@ -224,6 +224,34 @@ ADE_VIRTVAR(Team, l_Goals, nullptr, "The goal team", "team", "The goal team")
 	return ade_set_args(L, "o", l_Team.Set(Mission_goals[current].team));
 }
 
+ADE_VIRTVAR(isGoalSatisfied, l_Goals, nullptr, "The status of the goal", "number", "0 if failed, 1 if complete, 2 if incomplete")
+{
+	int current;
+	if (!ade_get_args(L, "o", l_Goals.Get(&current))) {
+		return ADE_RETURN_NIL;
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/briefing.cpp
```diff
@@ -224,6 +224,34 @@ ADE_VIRTVAR(Team, l_Goals, nullptr, "The goal team", "team", "The goal team")
...
+	int current;
+	if (!ade_get_args(L, "o", l_Goals.Get(&current))) {
+		return ADE_RETURN_NIL;
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "i", l_Team.Set(Mission_goals[current].satisfied));
```
BMagnu, line 238: Where does the l_team come from? You're not assigning an object here, so shouldn't you just pass the int?

code/scripting/api/objs/briefing.cpp
```diff
@@ -224,6 +224,34 @@ ADE_VIRTVAR(Team, l_Goals, nullptr, "The goal team", "team", "The goal team")
...
+	int current;
+	if (!ade_get_args(L, "o", l_Goals.Get(&current))) {
+		return ADE_RETURN_NIL;
+	}
+
+	if (ADE_SETTING_VAR) {
+		LuaError(L, "This property is read only.");
+	}
+
+	return ade_set_args(L, "i", l_Team.Set(Mission_goals[current].score));
```
BMagnu, line 252: here as well

BMagnu (changes requested):

code/scripting/api/objs/briefing.cpp
MjnMixael: Copy/paste oversight.

BMagnu (approved):

Issue #4840, opened 2022-10-31
22.2.0 failing to build Vulkan backend with Vulkan 1.3.224 or newer


brad0:
Afer upgrading to Vulkan 1.3.224 (or newer) we noticed the Vulkan backend fails to build like so..

```
/home/ports/pobj/fs2open-22.2.0/fs2open.github.com-release_22_2_0/code/graphics/vulkan/VulkanRenderer.cpp:794:21: error: no viable overloaded '='
        m_graphicsPipeline = m_device->createGraphicsPipelineUnique(nullptr, pipelineInfo);
        ~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/local/include/vulkan/vulkan.hpp:1199:20: note: candidate function not viable: no known conversion from 'ResultValue<UniqueHandle<vk::Pipeline, vk::DispatchLoaderDynamic>>' to 'const vk::UniqueHandle<vk::Pipeline, vk::DispatchLoaderDynamic>' for 1st argument
    UniqueHandle & operator=( UniqueHandle const & ) = delete;
                   ^
/usr/local/include/vulkan/vulkan.hpp:1201:20: note: candidate function not viable: no known conversion from 'ResultValue<UniqueHandle<vk::Pipeline, vk::DispatchLoaderDynamic>>' to 'vk::UniqueHandle<vk::Pipeline, vk::DispatchLoaderDynamic>' for 1st argument
    UniqueHandle & operator=( UniqueHandle && other ) VULKAN_HPP_NOEXCEPT
                   ^
```

brad0:
ping.

The-E:
Vulkan support is still at an incredibly early stage (as in, _it doesn't work yet_), so this is somewhat expected.

That said, I am working on bringing it forward.

brad0:
> Vulkan support is still at an incredibly early stage (as in, _it doesn't work yet_), so this is somewhat expected.
> 
> That said, I am working on bringing it forward.

Ah, I was not aware of that.

Thanks.

PR #5160, opened 2023-02-01
make the vulkan support buildable again

The-E:
we are currently shipping an old copy of `vulkan-hpp`, which has since been rolled into the official vulkan sdk. This PR removes that library, and refactors our existing code to use the base sdk version instead.

Several notes:
* This change means that the vulkan sdk is retrieved automatically during CI builds
* Vulkan support has been disabled for Win32 (x86) builds in CI, as neither the github action used to install the sdk nor cmake's vulkan finder properly support x86
* Vulkan support also has been disabled for MacOS in CI due to issues with (of all things) the test executable not linking correctly
* Building a vulkan-enabled exe locally _requires_ the vulkan sdk to be installed locally as well
* This doesn't improve functionality. Vulkan support is still deeply WIP.


The-E:
This addresses #4840

The-E:
*sigh*
of course, in my wisdom I forgot to add functionality to cmake to download the vulkan sdk....

freespace2/SDLGraphicsOperations.cpp
```diff
@@ -153,8 +155,10 @@ SDLGraphicsOperations::SDLGraphicsOperations() {
 }
 SDLGraphicsOperations::~SDLGraphicsOperations() {
 	SDL_QuitSubSystem(SDL_INIT_VIDEO);
-	ImGui_ImplSDL2_Shutdown();
-	ImGui_ImplOpenGL3_Shutdown();
+	if (!Cmdline_vulkan) {
+		ImGui_ImplSDL2_Shutdown();
```
BMagnu, line 159: In the manager you run the SDL2 Frame in Vulkan mode, but here you don't do the shutdown in Vulkan mode.
These should likely be symmetric, so either don't do the SDL2 Frame in Vulkan mode, or (more likely AFAICT) do the SDL2 shutdown in vulkan mode as well

BMagnu (changes requested):

BMagnu (approved):

PR #5162, opened 2023-02-01
increase mission author name length limit

MjnMixael:
Change author from a char array to a basic string.

This particular string is never used outside of FRED and SCPUI, both of which handle it internally as CString and SCP_string, we're just truncating it to pass it around internally. The FRED field allows unlimited input and SCPUI can handle long strings in a more modern way (can wrap, auto-overflow, etc). So let's just keep it as a string and remove the limit entirely.

### DIFF
--- a/fred2/missionnotesdlg.cpp
+++ b/fred2/missionnotesdlg.cpp
@@ -314,7 +314,7 @@ void CMissionNotesDlg::OnOK()
 	pad_with_newline(m_mission_notes, NOTES_LENGTH - 1);
 
 	string_copy(The_mission.name, m_mission_title, NAME_LENGTH - 1, 1);
-	The_mission.author = m_designer_name;
+	string_copy(The_mission.author, m_designer_name, true);
 	string_copy(The_mission.loading_screen[GR_640], m_loading_640, NAME_LENGTH - 1, 1);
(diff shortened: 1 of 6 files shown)


fred2/missionnotesdlg.cpp
```diff
@@ -314,7 +314,7 @@ void CMissionNotesDlg::OnOK()
 	pad_with_newline(m_mission_notes, NOTES_LENGTH - 1);
 
 	string_copy(The_mission.name, m_mission_title, NAME_LENGTH - 1, 1);
-	string_copy(The_mission.author, m_designer_name, NAME_LENGTH - 1, 1);
+	The_mission.author = m_designer_name;
```
Goober5000, line 317: The `string_copy` function calls `set_modified` if the old name and new name are different.  This assignment should do the same thing.

There is a `string_copy` for C-style strings and for CStrings, but not for SCP_strings.  I'd suggest making a SCP_string version using the CString version as a basis.

Goober5000 (changes requested):

fred2/missionnotesdlg.cpp
MjnMixael: This is done

Goober5000: So it is.  Cool

Goober5000 (approved):

PR #5163, opened 2023-02-01
Lua update discord

MjnMixael:
Allow Lua to set the Discord presence to custom strings. This is intended for SCPUI where game state changes sometimes skip over the built-in logic and custom game states might warrant updating the presence as well.


The-E (approved):

code/libs/discord/discord.cpp
```diff
@@ -272,5 +272,15 @@ void shutdown()
 	shutdown_discord();
 }
 
+void set_presence_string(SCP_string text)
```
Goober5000, line 275: This ought to be `const SCP_string&`, like `set_presence`

Goober5000 (changes requested):
Small change

code/libs/discord/discord.cpp
MjnMixael: done

Goober5000 (approved):

code/libs/discord/discord.h
```diff
@@ -5,5 +5,7 @@ namespace discord {
 
 void init();
 void shutdown();
+void set_presence_string(SCP_string text);
```
Goober5000: This will need `const SCP_string &` also

Goober5000 (dismissed):

PR #5165, opened 2023-02-01
Fix WeaponUsedConditions

### LINKED ISSUE #5080: ammo packs reload infinitely on RC builds

BMagnu:
Supercedes #5086.
Fixes #5080.

This is the IMO proper way of fixing the bug in question.
With this change, the hook behaves as it is documented.

Previously, if a weapon class was specified, the hook would only fire if the selected weapon matches the weapon class. Now, the hook will fire if any fired weapon will match the weapon class.
At the same time, the hook will, as before, only fire once per gun trigger, keeping scripts working as intended with no change.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1889,7 +1889,7 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 				turret->last_fired_weapon_info_index = turret_weapon_class;
 
 				if (scripting::hooks::OnTurretFired->isActive()) {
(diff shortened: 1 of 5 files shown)


MageKing17 (approved):
Code looks good, and the solution seems ideal.

PR #5169, opened 2023-02-03
Feature for countermeasures to respect cargo size

MjnMixael:
What it says on the tin. Game_settings.tbl flag to turn the feature on, otherwise cargo size is always assumed to be 1.0f. This bit may not be required because cargo size for all weapons/countermeasures is initialized to 1.0f... but there's no guarantee some mod somewhere didn't try something since Cargo Size would have been parsed and just gone unused for CMs.


BMagnu (commented):
Tenative approval.
I'm a bit hesitant though over all the added float divisions by one that are assumed to not change a value if cast from int to float divided by one cast to int.
It's probably fine, but if this comes back to bite some day, we'd be better off just doing the division if its needed, and not if the cargo size is 1

MjnMixael:
> Tenative approval. I'm a bit hesitant though over all the added float divisions by one that are assumed to not change a value if cast from int to float divided by one cast to int. It's probably fine, but if this comes back to bite some day, we'd be better off just doing the division if its needed, and not if the cargo size is 1

That's a fair point. I've added a commit that gates all the float divisions behind the mod flag, otherwise previous behavior is preserved.

BMagnu (approved):

Goober5000 (commented):
What about, instead, always doing the division by cargo size, but using the flag to assign the cargo size (either 1 or the parsed value) at the time it is actually parsed?

BMagnu:
Point is that we don't actually want the division out of fear of floating point accuracies.
Can we guarantee that for all countermeasure sizes and on all processors, a float div by 1 will always yield the same value, even when cast to int afterwards?
If so, sure, otherwise, it'll be a problem

Goober5000:
Bleh, good point.  I would tentatively say yes, because a parsed 1.0f and an assigned 1.0f should both work as expected, and consistently, on IEEE 754 platforms.  But I can't be sure.  @Baezon, any insight?

Baezon:
IEEE specifies that float arithmetic operations (+-/*) should return the exact same closest representable float for the exact same inputs, so it should be ok. In practice I think this only need be as accurate as other ammo handling, which I notice often prefers to do `(int)std::lround(capacity / cargo_size)`

Goober5000:
So changing the implementation works... but the current implementation also works.  I'm ok with sticking with the implementation as coded.  Merging.

PR #5174, opened 2023-02-07
get audiostream duration in float seconds

MjnMixael:
ADE_FUNC to get the duration of an audiostream.

### DIFF
--- a/code/sound/audiostr.cpp
+++ b/code/sound/audiostr.cpp
@@ -1175,7 +1175,7 @@ int audiostream_is_paused(int i)
 	return (int) Audio_streams[i].Is_Paused();
 }
 
-int audiostream_get_duration(int i)
+double audiostream_get_duration(int i)
 {
 	if (i == -1)
 		return 0;
@@ -1185,7 +1185,7 @@ int audiostream_get_duration(int i)
 	if (Audio_streams[i].status == ASF_FREE)
 		return -1;
 	
-	return fl2i(1000.0f * Audio_streams[i].Get_Duration());
+	return Audio_streams[i].Get_Duration();
 }
 
 void audiostream_set_sample_cutoff(int i, uint cutoff)
--- a/code/sound/audiostr.cpp
+++ b/code/sound/audiostr.cpp
@@ -112,7 +112,7 @@ public:
 	void Fade_and_Stop();
 	void	Set_Volume(float vol);
 	float	Get_Volume();
-	int     Get_Duration();
+	double  Get_Duration();
 	void	Init_Data();
 	void	Set_Sample_Cutoff(uint sample_cutoff);
 	void	Set_Default_Volume(float vol) { m_lDefaultVolume = vol; }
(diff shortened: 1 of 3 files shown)


code/sound/audiostr.cpp
```diff
@@ -819,6 +820,11 @@ float AudioStream::Get_Volume()
 	return m_lVolume;
 }
 
+int AudioStream::Get_Duration()
+{
+	return fl2i(1000.0f * m_fileProps.duration);
+}
```
Goober5000, line 826: Since this is an accessor function, it should return the field directly as a `double`, just like `Get_Volume` returns its own field directly.  Then the conversion to milliseconds should be done in `audiostream_get_duration`.

code/sound/audiostr.h
```diff
@@ -66,6 +66,9 @@ void audiostream_set_volume(int i, float volume);
 // see if a particular stream is paused
 int audiostream_is_paused(int i);
 
+// get duration of a stream
```
Goober5000, line 69: The comment should indicate milliseconds

Goober5000 (changes requested):

Goober5000 (approved):

EatThePath:
Do any of our other lua API functions retrieve int milliseconds rather than float seconds? It makes sense with sexps but in Lua it seems like an unneeded source of surprise

MjnMixael:
> Do any of our other lua API functions retrieve int milliseconds rather than float seconds? It makes sense with sexps but in Lua it seems like an unneeded source of surprise

Time returns in the lua API are all over the place. This is correctly documented so not a source of surprise.

EatThePath:
I think it's a source of some surprise, as someone who happens upon the function in a script or sees it in a list of functions, they may make assumptions rather than dig up the documentation. For that reason, and because there are only a handful of returns or parameters that return ms(at least, ones that have documented that fact) I think it would be better for the API to have this return seconds instead.

I'm not going to block merging for it, but I do think it'd be the better choice.

MjnMixael:
> I think it's a source of some surprise, as someone who happens upon the function in a script or sees it in a list of functions, they may make assumptions rather than dig up the documentation. For that reason, and because there are only a handful of returns or parameters that return ms(at least, ones that have documented that fact) I think it would be better for the API to have this return seconds instead.
> 
> I'm not going to block merging for it, but I do think it'd be the better choice.

I think you're making a lot of assumptions yourself. I, for one, am not generally interested in bending over backwards for scripters who don't use the documentation if they end up unsure about something. Scripting blind is an 'at your own risk' activity. But just because you've become the squeaky wheel here, I've converted it to seconds.

MjnMixael:
> I think it's a source of some surprise, as someone who happens upon the function in a script or sees it in a list of functions, they may make assumptions rather than dig up the documentation. For that reason, and because there are only a handful of returns or parameters that return ms(at least, ones that have documented that fact) I think it would be better for the API to have this return seconds instead.
> 
> I'm not going to block merging for it, but I do think it'd be the better choice.

I see no universe where people should be scripting without the documentation and if they are, that's their own issue to solve. But fine. I've made it float seconds. Now stop bugging me about it.

code/sound/audiostr.h
```diff
@@ -66,6 +66,9 @@ void audiostream_set_volume(int i, float volume);
 // see if a particular stream is paused
 int audiostream_is_paused(int i);
 
+// get duration of a stream in milliseconds
```
Goober5000, line 69: This should now be changed to seconds
```suggestion
// get duration of a stream in seconds
```

Goober5000 (changes requested):

Goober5000 (approved):

PR #5176, opened 2023-02-10
Remove some old debug key checks in mainhall

MjnMixael:
Removes some older mainhall shortcuts from debug that have outlived their usefulness and get in the way of newer features like mainhall cheats.


BMagnu (approved):
Yeah, these can probably go.

PR #5178, opened 2023-02-11
Lua method to free all loaded models

MjnMixael:
I realize this one is inherently risky because if someone uses it at the wrong time, fun things could happen. However, as SCPUI becomes more fully featured I have run into issues regarding having too many models loaded and giving the scripts a way to cleanup after themselves seems like a good idea.

For example; if 3D ship and weapon selection is enabled then the current SCPUI scripts will, during the first mission load of an instance of FSO, load and render every player allowed ship and weapon to a texture, saving each to a png blob for use in ship and weapon selection. In practice it works well, but lower end computers ran into out of memory errors.

I instead moved that whole function to On Game Init which solves the out of memory during a mission load for those computers (because FSO frees all models before starting a mission load), but it means that now until the player runs a mission the game has a large number of models and weapons sitting in memory. If they then go to the tech room and start viewing ships they can eventually run into the same issues because SCPUI has no way to clean up.

So with this I can free up all loaded models after the icons are made or when the tech room is loaded/exited, etc.

### DIFF
--- a/code/scripting/api/libs/graphics.cpp
+++ b/code/scripting/api/libs/graphics.cpp
@@ -2169,5 +2169,13 @@ ADE_FUNC(screenToBlob, l_Graphics, nullptr, "Captures the current render target 
 	return ade_set_args(L, "s", gr_blob_screen().c_str());
 }
 
+ADE_FUNC(freeAllModels, l_Graphics, nullptr, "Releases all loaded models and frees the memory. Intended for use in UI situations "
+	"and not within missions. Use at your own risk!", nullptr, nullptr)
+{
+	model_free_all();
+
+	return ADE_RETURN_NIL;
+}
+
 } // namespace api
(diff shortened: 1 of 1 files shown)


Goober5000:
@notimaginative , are there any gotchas to be aware of when freeing models?

notimaginative:
> @notimaginative , are there any gotchas to be aware of when freeing models?

Not really, if used properly. If it's used as a cleanup function, it's intended purpose, then everything should be fine. And that looks to be the intended use here. The only real caveat is to make sure that it's not called after mission parse.

Goober5000 (changes requested):
Can you add that caveat to the function documentation?

MjnMixael:
> Can you add that caveat to the function documentation?

done

Goober5000 (approved):

PR #5182, opened 2023-02-12
Fix thruster sounds and add flag for player thruster sounds

Baezon:
Due to being connected to rendering, player thruster starts and stops wouldn't play in the cockpit, were played at the wrong location in 3d space and loops wouldn't play at all due to lacking `OS_PLAY_ON_PLAYER`. This inconsistent behavior is fixed and splits out sounds into their own function, and additionally a flag is added in order to hear player thrusters at all, similar to turret sounds which I think makes sense.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -20164,13 +20164,10 @@ void ship_render_batch_thrusters(object *obj)
 
 	if ( Rendering_to_shadow_map ) return;
 
-	physics_info *pi = &Objects[shipp->objnum].phys_info;
-	float render_amount;
-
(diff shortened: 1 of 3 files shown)


code/ship/ship.cpp
```diff
@@ -20041,89 +20170,14 @@ void ship_render_batch_thrusters(object *obj)
 	for ( int i = 0; i < sip->num_maneuvering; i++ ) {
```
Goober5000, line 20170: `pi` is now an unused variable in this function

Goober5000 (approved):

PR #5184, opened 2023-02-13
Add option to deprecate hooks

### LINKED ISSUE #3171: FR: Add deprecation option to scripting hooks
As noted by @asarium on #3151, we should make hooks be deprecatable. This will also be interesting for the new controls, as they introduce a new method of binding buttons from script which should make the OnAction, OnKey\* and OnMouse\* hooks obsolete.
As I intend to convert all old hooks to the new style of hooks in #3151 eventually, if this is implemented, deprecation should be compatible with said new style of hooks as well.

> Also a good point about the deprecation. However, that will require some more work to integrate it into the existing deprecation system so that is out of scope for this PR.

_Originally posted by @asarium in https://github.com/scp-fs2open/fs2open.github.com/issues/3151#issuecomment-766967833_

### LINKED ISSUE #4742: Regression: Splash screen conditional and global hooks
I discovered that the Splash screen conditional and global hooks no longer run what's in them, and tracked it down to June 20, 2021 `Moved script initialization to take place after all table loading to facilitate script condition caching` #3522. 

For example, this script will never trigger:
```
#Conditional Hooks

$Application: FS2_Open

$On Splash Screen: [
    ba.warning("zzz \n")
]
+Override: true

#End
```

I'll also add that whether or not Override is set doesn't seem to make a difference for this test.
Debug log shows the `game_title_screen_display()` function is running ofc (`freespace.cpp` line 7452)
```
SCRIPTING: Splash screen overrides checked
Compiling new shader:
    Default material
Shader Variant Features:
SCRIPTING: Splash screen conditional hook has been run
```

BMagnu:
Adds the ability to deprecate scripting hooks and does so with the OnDeath and the OnSplashScreen hooks.
Fixes #4742 and closes #3171.
This PR approaches hook deprecation with two stages:
There is a warn-type deprecate that will warn but run the hook if the target version is below the deprecation version, and error out if the target version is the same or above the deprecation version.
Secondly, there is an error-type deprecate that will always hard-error if a user attempts to use it for situations (like the OnSplashScreen override) where function of the hook / override cannot be feasibly enabled, even in a limited fashion for mods targeting older versions.
In addition, these override-modes can be set differently for the actual hook and its override, if applicable.

### DIFF
--- a/code/mod_table/mod_table.h
+++ b/code/mod_table/mod_table.h
@@ -57,7 +57,6 @@ extern float Generic_pain_flash_factor;
 extern float Shield_pain_flash_factor;
 extern float Emp_pain_flash_factor;
 extern std::tuple<float, float, float> Emp_pain_flash_color;
-extern gameversion::version Targeted_version;
 extern SCP_string Window_title;
 extern SCP_string Mod_title;
 extern SCP_string Mod_version;
@@ -155,3 +154,5 @@ void mod_table_reset();
  * @return @c true if the mod specified support for this or a later version, @c false otherwise
  */
 bool mod_supports_version(int major, int minor, int build);
+
+bool mod_supports_version(const gameversion::version& version);
--- a/code/mod_table/mod_table.h
+++ b/code/mod_table/mod_table.h
@@ -57,6 +57,7 @@ extern float Generic_pain_flash_factor;
 extern float Shield_pain_flash_factor;
 extern float Emp_pain_flash_factor;
 extern std::tuple<float, float, float> Emp_pain_flash_color;
+extern gameversion::version Targeted_version;
 extern SCP_string Window_title;
 extern SCP_string Mod_title;
 extern SCP_string Mod_version;
--- a/code/scripting/api/libs/engine.cpp
+++ b/code/scripting/api/libs/engine.cpp
@@ -68,7 +68,7 @@ ADE_FUNC(addHook,
 			LuaError(L, "Hook '%s' is deprecated since version %s and cannot be used if the mod targets that version or higher!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
 		}
 		else {
-			Warning(LOCATION, "Hook '%s' is deprecated from version %s should be replaced!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+			Warning(LOCATION, "Hook '%s' is deprecated from version %s and should be replaced!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
 			shownWarn = true;
 		}
 		if (override_func.isValid()) {
@@ -79,7 +79,7 @@ ADE_FUNC(addHook,
 				LuaError(L, "Overriding Hook '%s' is deprecated since version %s and cannot be used if the mod targets that version or higher!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
 			}
 			else if (!shownWarn) {
-				Warning(LOCATION, "Overriding Hook '%s' is deprecated from version %s should be replaced!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+				Warning(LOCATION, "Overriding Hook '%s' is deprecated from version %s and should be replaced!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
 			}
 		}
 	}
--- a/code/scripting/api/libs/engine.cpp
+++ b/code/scripting/api/libs/engine.cpp
@@ -64,7 +64,7 @@ ADE_FUNC(addHook,
 		if (deprecation.level_hook == HookDeprecationOptions::DeprecationLevel::LEVEL_ERROR) {
 			LuaError(L, "Hook '%s' is removed since version %s and cannot be used!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
 		}
-		else if (deprecation.deprecatedSince <= Targeted_version) {
+		else if (mod_supports_version(deprecation.deprecatedSince)) {
 			LuaError(L, "Hook '%s' is deprecated since version %s and cannot be used if the mod targets that version or higher!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
 		}
 		else {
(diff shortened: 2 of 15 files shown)


code/mod_table/mod_table.h
```diff
@@ -57,6 +57,7 @@ extern float Generic_pain_flash_factor;
 extern float Shield_pain_flash_factor;
 extern float Emp_pain_flash_factor;
 extern std::tuple<float, float, float> Emp_pain_flash_color;
+extern gameversion::version Targeted_version;
```
Goober5000, line 60: `Targeted_version` is scoped to mod_table.cpp, and the standard use is to call `mod_supports_version`.  I would favor adding an overload of `mod_supports_version` that works on `gameversion::version` rather than exposing the variable.

code/scripting/api/libs/engine.cpp
```diff
@@ -58,6 +58,32 @@ ADE_FUNC(addHook,
 		action.hook.override_function.function = override_func;
 	}
 
+	if (action_hook->getDeprecation()) {
+		const auto& deprecation = *action_hook->getDeprecation();
+		bool shownWarn = false;
+		if (deprecation.level_hook == HookDeprecationOptions::DeprecationLevel::LEVEL_ERROR) {
+			LuaError(L, "Hook '%s' is removed since version %s and cannot be used!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+		}
+		else if (deprecation.deprecatedSince <= Targeted_version) {
```
Goober5000, line 67: An overloaded `mod_supports_version` can be used here

code/scripting/api/libs/engine.cpp
```diff
@@ -58,6 +58,32 @@ ADE_FUNC(addHook,
...
+			LuaError(L, "Hook '%s' is deprecated since version %s and cannot be used if the mod targets that version or higher!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
```
Goober5000, line 68: I would change "cannot" to "should not" here

code/scripting/api/libs/engine.cpp
```diff
@@ -58,6 +58,32 @@ ADE_FUNC(addHook,
...
+		}
+		else {
+			Warning(LOCATION, "Hook '%s' is deprecated from version %s should be replaced!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+			shownWarn = true;
+		}
+		if (override_func.isValid()) {
+			if (deprecation.level_override == HookDeprecationOptions::DeprecationLevel::LEVEL_ERROR) {
+				LuaError(L, "Overriding Hook '%s' is removed since version %s and cannot be used!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+			}
+			else if (deprecation.deprecatedSince <= Targeted_version) {
```
Goober5000, line 78: An overloaded `mod_supports_version` can be used here

code/scripting/api/libs/engine.cpp
```diff
@@ -58,6 +58,32 @@ ADE_FUNC(addHook,
...
+		const auto& deprecation = *action_hook->getDeprecation();
+		bool shownWarn = false;
+		if (deprecation.level_hook == HookDeprecationOptions::DeprecationLevel::LEVEL_ERROR) {
+			LuaError(L, "Hook '%s' is removed since version %s and cannot be used!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+		}
+		else if (deprecation.deprecatedSince <= Targeted_version) {
+			LuaError(L, "Hook '%s' is deprecated since version %s and cannot be used if the mod targets that version or higher!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+		}
+		else {
+			Warning(LOCATION, "Hook '%s' is deprecated from version %s should be replaced!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
```
Goober5000, line 71: "and should"

code/scripting/api/libs/engine.cpp
```diff
@@ -58,6 +58,32 @@ ADE_FUNC(addHook,
...
+			shownWarn = true;
+		}
+		if (override_func.isValid()) {
+			if (deprecation.level_override == HookDeprecationOptions::DeprecationLevel::LEVEL_ERROR) {
+				LuaError(L, "Overriding Hook '%s' is removed since version %s and cannot be used!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+			}
+			else if (deprecation.deprecatedSince <= Targeted_version) {
+				LuaError(L, "Overriding Hook '%s' is deprecated since version %s and cannot be used if the mod targets that version or higher!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
```
Goober5000, line 79: I would change "cannot" to "should not" here

code/scripting/api/libs/engine.cpp
```diff
@@ -58,6 +58,32 @@ ADE_FUNC(addHook,
...
+			}
+			else if (!shownWarn) {
+				Warning(LOCATION, "Overriding Hook '%s' is deprecated from version %s should be replaced!", action_hook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
```
Goober5000, line 82: "and should"

code/scripting/scripting.cpp
```diff
@@ -778,11 +764,39 @@ bool script_action::ConditionsValid(const linb::any& local_condition_data) const
...
+				error_display(1, "Overriding Hook '%s' is removed since version %s and cannot be used!", parentHook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+			}
+			else if (deprecation.deprecatedSince <= Targeted_version) {
+				error_display(1, "Overriding Hook '%s' is deprecated since version %s and cannot be used if the mod targets that version or higher!", parentHook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+			}
+			else if (!shownWarn) {
+				error_display(0, "Overriding Hook '%s' is deprecated from version %s should be replaced!", parentHook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+			}
+		}
+	}
```
Goober5000, line 798: Same comments as in the other function

code/scripting/scripting.cpp
```diff
@@ -855,6 +869,34 @@ bool script_state::ParseCondition(const char *filename)
...
+				if (deprecation.level_override == HookDeprecationOptions::DeprecationLevel::LEVEL_ERROR) {
+					error_display(1, "Overriding Hook '%s' is removed since version %s and cannot be used!", currHook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+				}
+				else if (deprecation.deprecatedSince <= Targeted_version) {
+					error_display(1, "Overriding Hook '%s' is deprecated since version %s and cannot be used if the mod targets that version or higher!", currHook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+				}
+				else if(!shownWarn) {
+					error_display(0, "Overriding Hook '%s' is deprecated from version %s should be replaced!", currHook->getHookName().c_str(), gameversion::format_version(deprecation.deprecatedSince).c_str());
+				}
+			}
```
Goober5000, line 897: Same comments as in the other function

Goober5000 (changes requested):

code/scripting/api/libs/engine.cpp
BMagnu: Hmm, I'm on the fence.
Cause in case it's set to warning mode but the target version is past the deprecation version, it'll error out here. And if it errors out, it kind of can't be used, as the error will stop FSO.

Goober5000: You're right, I had mistaken this line for a line that warned instead of errored.  I withdraw my comment.

Goober5000 (approved):

PR #5186, opened 2023-02-14
On warp complete hook

MjnMixael:
This allows, for instance, scripts to make a ship do the warp out effect but without actually removing it from the mission. Should make imjump scripts and sexps much simpler and cleaner.


Goober5000 (approved):
Good idea, and LGTM, though I would want to double check with @BMagnu on account of the hook conditions

BMagnu (approved):

PR #5190, opened 2023-02-15
Lua method to check if the game has focus

### LINKED ISSUE #5164: It may be useful for SCPUI to know when FSO gets or loses focus
SCPUI scripts continue to run even if FSO has lost focus. In most cases this isn't an issue, but I have one where it does matter.

Seems like this may be simple to implement. Start with `gr_activate()` in 2d.cpp line 1694. Current plan is `ba.hasFocus` that returns true or false based on `gr_activate`.

MjnMixael:
Fixes #5164

### DIFF
--- a/code/osapi/osapi.cpp
+++ b/code/osapi/osapi.cpp
@@ -421,7 +421,7 @@ void os_cleanup()
 // window management -----------------------------------------------------------------
 
 // Returns 1 if app is not the foreground app.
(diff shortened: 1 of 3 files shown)


Goober5000 (approved):
It would be good to change `os_foreground` to return a bool instead of an int.  That isn't a requirement but would be a nice cleanup.

PR #5191, opened 2023-02-15
Unlimited Power

### LINKED ISSUE #3583: Script/SEXP FR: Control over lightning
I've always thought it would be neat if modders could do things with the nebula lightning effect outside of nebula missions.

What I would like is to have some way to spawn a lightning bolt at will, ideally through scripting. It would allow for a number of cool things, from dramatic lightning strikes, to proper lightning guns, to **unlimited power!**.

From what I recall from glancing at some of the relevant code years ago, it *seemed* like exposing the lightning bolt creation function itself wouldn't be overly complicated, but that non-nebula missions would also need to be able to spawn lightning at all (see nebl_process() in neblightning.cpp for instance).

MjnMixael:
Adds a sexp and lua method to create a lightning bolt between two arbitrary points. You can, optionally, not play the default thunder sound effect. Fixes #3583


code/nebula/neblightning.h
```diff
@@ -45,6 +45,7 @@ typedef struct l_bolt {
 	int		delay;				// delay stamp
 	int		strikes_left;		// #of strikes left
 	float	width;
+	bool	playSound;
```
Baezon, line 48: Pedantic, I know, but should be snake case. Ditto for the parameter names of the same, etc.

Baezon (changes requested):

code/nebula/neblightning.h
MjnMixael: Done

Baezon (approved):

Goober5000 (changes requested):
Since a new `OPF_` definition was added, `check_sexp_syntax` needs to be updated to handle it.  Also, a new `SEXP_CHECK_INVALID_BOLT_TYPE` should be added with a corresponding check in `sexp_error_message`.

MjnMixael:
> Since a new `OPF_` definition was added, `check_sexp_syntax` needs to be updated to handle it. Also, a new `SEXP_CHECK_INVALID_BOLT_TYPE` should be added with a corresponding check in `sexp_error_message`.

done

Goober5000 (approved):

PR #5199, opened 2023-02-18
Add `GetVertex()` and `NumVertices()` script functions and clean up random submodel point retrieval a bit

Baezon:
`GetVertex` allows for retrieval of specific vertices or random ones, and with `NumVertices()` a scripter could even do some advanced 'model parsing' type operations if they were feeling spicy, but random retrieval was my main goal.

Additionally, cleans up `submodel_get_two_random_points_better` a bit, there's no real reason for this function to get two points at a time, even if that's a common case, resulting in awkward scenarios where callers need to declare useless temporaries when they only need one. Instead it should return a single vertex, and you can call it more than once if you need more (and for what it's worth, the Shivan Comm Node *does* have a collision tree).

### DIFF
--- a/code/model/modelinterp.cpp
+++ b/code/model/modelinterp.cpp
@@ -1229,42 +1229,25 @@ static int submodel_get_points_internal(int model_num, int submodel_num)
 /**
  * Gets two random points on a model
  */
-void submodel_get_two_random_points(int model_num, int submodel_num, vec3d *v1, vec3d *v2, vec3d *n1, vec3d *n2 )
+vec3d submodel_get_random_point_fallback(int model_num, int submodel_num)
 {
 	int nv = submodel_get_points_internal(model_num, submodel_num);
 
-	// this is not only because of the immediate div-0 error but also because of the less immediate expectation for at least one point (preferably two) to be found
+	// this is not only because of the immediate div-0 error but also because of the less immediate expectation for at least one point to be found
 	if (nv <= 0) {
 		polymodel *pm = model_get(model_num);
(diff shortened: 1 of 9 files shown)


code/model/modelinterp.cpp
```diff
@@ -1273,12 +1256,10 @@ void submodel_get_two_random_points_better(int model_num, int submodel_num, vec3
 			submodel_num = pm->detail[0];
 		}
 
-		// the Shivan Comm Node does not have a collision tree, for one
 		if (pm->submodel[submodel_num].collision_tree_index < 0) {
 			nprintf(("Model", "In submodel_get_two_random_points_better(), model %s does not have a collision tree!  Falling back to submodel_get_two_random_points().\n", pm->filename));
```
The-E, line 1260: a) I have never, ever seen this print
b) the function name referenced here should be updated to reflect the actual function throwing this warning

The-E (changes requested):
One small change, then it should be good

Goober5000:
> (and for what it's worth, the Shivan Comm Node does have a collision tree).

The Shivan Comm Node most certainly *did not* have a collision tree when I tested it.  Did you test both the retail and MVP versions of the model?

Anyway, that's the entire reason for that if() block; the function crashed before I added it.

Baezon:
The MVPs comm node indeed does not... I had assumed it was a volition comment, actually. I didn't think not having a collision tree was even possible, and `submodel_get_random_point_fallback` doesn't even return correct vertex positions. I've updated the PR, but I'm investigating this further...

Baezon:
Opened #5208, to fix the lack of a collision tree sometimes. With that PR, all models can be asserted to have collision trees. (The vertex positions were fine btw, the mvps comm node is just set up a bit differently)

Goober5000:
Ah, nice

The-E (approved):

Issue #1708, opened 2018-04-19
Assert: "tree_index >= 0" after #1630 when model with $nocollide_this_only explodes


tomimaki:
Release build gives me multiple times
![nocollide](https://user-images.githubusercontent.com/18072803/39009055-d45342a8-440a-11e8-8d5d-4960a28103c9.png)
and debug build: ASSERTION: "tree_index >= 0" at modelread.cpp:5232
[fs2_open.log](https://github.com/scp-fs2open/fs2open.github.com/files/1929204/fs2_open.log)

Example is Shivan Commnode from Mediavps 2014. Load mission Into Lion's Den (loop2-2.fs2), pack Trebuchet, destroy Crystal in Commnode and wait until it explode.

Another example is Vishnan_Protector from BP but can't find reliable mission.

ngld:
The debug assertion is caused by the fact that [`submodel_get_two_random_points_better`](https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/model/modelinterp.cpp#L1327) assumes that detail0 always has a collision tree. If the tree hasn't been loaded, the `tree_index` is `-1` which then causes that assertion in `model_get_bsp_collision_tree`.

A nice solution would be to make `submodel_get_two_random_points_better` use a collide subobject if one is available instead of detail0 but I don't know enough about the code to do this myself.

Goober5000:
I don't know about collision subobjects myself, but it's a simple matter to fall back to the old method, so I made a new PR to do just that.

tomimaki:
I can confirm that #1733 fixed issue.

Issue #4089, opened 2022-03-22
Doubled armor on crystal in Shivan Comm Node after 88c1d8a02a9062c394262aca3888e2debb15c22a commit


tomimaki:
So after:
```
88c1d8a02a9062c394262aca3888e2debb15c22a is the first bad commit
commit 88c1d8a02a9062c394262aca3888e2debb15c22a
Author: Goober5000 <ipw47@mailbox.org>
Date:   Sat Dec 12 02:04:02 2020 -0500

    consolidate model instance code

 code/ai/aibig.cpp                     |   3 +-
 code/ai/aicode.cpp                    |   5 +-
 code/ai/aiturret.cpp                  |  13 +-
 code/decals/decals.cpp                |  11 +-
 code/hud/hudtarget.cpp                |   1 -
 code/menuui/techmenu.cpp              |  21 +-
 code/mission/missionparse.cpp         |   2 +-
 code/missionui/redalert.cpp           |   2 +-
 code/model/model.h                    |  53 ++---
 code/model/modelanim.cpp              |  54 ++---
 code/model/modelcollide.cpp           |  13 +-
 code/model/modelread.cpp              | 381 ++++++++++++----------------------
 code/model/modelrender.cpp            |  35 +---
 code/network/multi_obj.cpp            |  24 ++-
 code/network/multimsgs.cpp            |  16 +-
 code/object/collideshipship.cpp       |  10 +-
 code/parse/sexp.cpp                   |  40 ++--
 code/scripting/api/objs/subsystem.cpp |  10 +-
 code/ship/ship.cpp                    |  61 +++---
 code/ship/ship.h                      |   4 +-
 code/ship/shipfx.cpp                  |  31 +--
 code/ship/shiphit.cpp                 |   6 +-
 22 files changed, 352 insertions(+), 444 deletions(-)
```
crystal in Shivan Comm Node looks weird in mission or lab, like it have doubled armor.
Before:
https://user-images.githubusercontent.com/18072803/159471258-0983ea96-02dd-4278-94e3-6599c342d713.mp4
After:
https://user-images.githubusercontent.com/18072803/159473353-dda672ae-c97e-47db-ae4a-f692a463dcc5.mp4
Similar in mission.

Of course it's possible that commit uncovered some bug in model.

z64555:
Ok, it looks like the crystal has a LOD or a copy of a LOD that's static in addition to the crystal that's rotating properly.

[...]

PR #5203, opened 2023-02-18
Remove previous bsp tree skipping optimization

Baezon:
Reverts #1630. The collision tree is a very useful data structure, even if it isn't being used for literal collision its the primary source of of simple vertex position and vertex count look ups. Although FotG may have been fine, it immediately caused issues in models with no_collide detail0s #1708. #1733 was a decent enough band-aid, but we really should just have the tree if we've loaded the model. With this, *all* models should have collision trees, the tree is how polygons are provided in the first place in the pof, so this is a guarantee.

### DIFF
--- a/code/debris/debris.cpp
+++ b/code/debris/debris.cpp
@@ -253,8 +253,8 @@ void debris_process_post(object * obj, float frame_time)
 
 			vec3d v1, v2, v3, v4;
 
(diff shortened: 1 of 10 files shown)


Goober5000 (approved):

Goober5000:
Nice work.  I wonder if this might help with #4089 as well.

wookieejedi:
Sounds good to me as well!

PR #5207, opened 2023-02-19
Fix negative child check inversion.

MageKing17:
In PR #5202, a bunch of `Assert(child >= 0);` statements were replaced with `if (child < 0) return nullptr;`, except the first one kept the assertion's condition (inverting the logic). This fixes this accidental inversion.


wookieejedi (approved):

Goober5000 (approved):
Whoops. Thanks.

PR #5212, opened 2023-02-21
Add collision flag for colliding with all polies

Baezon:
Adds `MC_COLLIDE_ALL`, which will accrue all hits including those against backfacing polies into a vector. Since this makes `mc_info` not trivially memcopyable anymore, they're replaced with typical deep copy assignments.

This flag is not yet used, but will be for a feature in development by @BMagnu

### DIFF
--- a/code/model/model.h
+++ b/code/model/model.h
@@ -1297,7 +1297,7 @@ inline void mc_info_init(mc_info *mc)
 #define MC_SUBMODEL				(1<<6)			// If this is set, only check the submodel specified in mc->submodel_num. Use with MC_CHECK_MODEL
 #define MC_SUBMODEL_INSTANCE	(1<<7)			// Check submodel and its children (of a rotating submodel)
 #define MC_CHECK_INVISIBLE_FACES (1<<8)		// Check the invisible faces.
-#define MC_COLLIDE_ALL (1<<8)				// Returns ALL hits via hit_points_all, including backfacing polies hits
+#define MC_COLLIDE_ALL (1<<9)				// Returns ALL hits via hit_points_all, including backfacing polies hits
 
 
 /*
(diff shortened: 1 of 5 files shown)


code/model/model.h
```diff
@@ -1294,6 +1297,7 @@ inline void mc_info_init(mc_info *mc)
 #define MC_SUBMODEL				(1<<6)			// If this is set, only check the submodel specified in mc->submodel_num. Use with MC_CHECK_MODEL
 #define MC_SUBMODEL_INSTANCE	(1<<7)			// Check submodel and its children (of a rotating submodel)
 #define MC_CHECK_INVISIBLE_FACES (1<<8)		// Check the invisible faces.
+#define MC_COLLIDE_ALL (1<<8)				// Returns ALL hits via hit_points_all, including backfacing polies hits
```
BMagnu, line 1300: Should that not be ``1<<9``?

BMagnu (changes requested):

code/model/model.h
Baezon: 🤦

BMagnu (approved):

The-E (commented):
Two questions on this: 
1. why have both the single hit return and an optional multi hit return? I think it would be cleaner if we always return a vector, with the understanding that it will only have more than one entry if `MC_COLLIDE_ALL` is set?
2. in the single return case, we also record the normal of the polygon being hit, why aren't we doing that for the `MC_COLLIDE_ALL` case?

Baezon:
1. Since collision is decently hot, I wanted to reduce impact where possible. Tmk as long as the vectors aren't added to, there's no allocation made for them, so existing uses where they don't need to vector are largely unaffected. I don't feel particularly strongly about it though, and using the vector would also be ok if desired.
2. Simply because the use case doesn't require it. There's quite a bit of data recorded for the hits which could theoretically be returned in the 'bulk' case, edge hits, hit_bitmap, hit_u/v, I just didn't add anything that wasn't required. If we were to go that route, I would probably feel stronger about replacing the returns entirely with the vectors.

The-E:
alright, sounds good to me.

PR #5213, opened 2023-02-21
Check duplicate enum items

MjnMixael:
Review of #5036 suggested to me that we should be checked for duplicate enum items here, too.

### DIFF
--- a/code/parse/sexp/LuaSEXP.cpp
+++ b/code/parse/sexp/LuaSEXP.cpp
@@ -577,6 +577,14 @@ void LuaSEXP::parseTable() {
 					item.resize(NAME_LENGTH - 1);
 				}
 
+				// Case insensitive check if the item already exists in the list
+				for (int i = 0; i < (int)thisList.list.size(); i++) {
+					if (!stricmp(item.c_str(), thisList.list[i].c_str())) {
+						error_display(0, "Enum item '%s' already exists in list %s. Skipping!\n", item.c_str(), thisList.name.c_str());
+						continue;
+					}
+				}
+
 				thisList.list.push_back(item);
 			}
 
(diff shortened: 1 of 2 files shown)


BMagnu (approved):

code/parse/sexp/LuaSEXP.cpp
```diff
@@ -577,6 +577,14 @@ void LuaSEXP::parseTable() {
 					item.resize(NAME_LENGTH - 1);
 				}
 
+				// Case insensitive check if the item already exists in the list
+				for (int i = 0; i < (int)thisList.list.size(); i++) {
+					if (!stricmp(item.c_str(), thisList.list[i].c_str())) {
+						error_display(0, "Enum item '%s' already exists in list %s. Skipping!\n", item.c_str(), thisList.name.c_str());
+						continue;
```
Goober5000, line 584: This will continue the inner loop, not the outer loop, and thus it won't skip adding the duplicate enum item to the list.

code/parse/sexp/LuaSEXP.cpp
```diff
@@ -577,6 +577,14 @@ void LuaSEXP::parseTable() {
 					item.resize(NAME_LENGTH - 1);
 				}
 
+				// Case insensitive check if the item already exists in the list
+				for (int i = 0; i < (int)thisList.list.size(); i++) {
+					if (!stricmp(item.c_str(), thisList.list[i].c_str())) {
```
Goober5000, line 582: This can use `SCP_string_lcase_less_than()(item, thisList.list[i])`

code/parse/sexp/sexp_lookup.cpp
```diff
@@ -76,6 +76,14 @@ void parse_sexp_table(const char* filename) {
 						item.resize(NAME_LENGTH - 1);
 					}
 
+					// Case insensitive check if the item already exists in the list
```
Goober5000, line 79: Same comments apply here

Goober5000 (changes requested):

Goober5000 (approved):

PR #5219, opened 2023-02-24
Remove myself from code owners

asarium:
I have not been able to keep up with code reviews for quite some time so I should not be a default one.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,13 +1,5 @@
 
-cmake/ @asarium
-
 ci/ @asarium @z64555
-
-code/executor/ @asarium
-
-code/scpui/ @asarium
-
-code/scripting/ @asarium
 
 code/controlconfig/ @z64555
 


cmake/util.cmake
```diff
@@ -70,7 +70,7 @@ MACRO(COPY_FILE_TO_TARGET _target _file)
 	if(UNIX)
 		ADD_CUSTOM_COMMAND(
 			TARGET ${_target} POST_BUILD
-			COMMAND cp -a "${_file}"  "$<TARGET_FILE_DIR:${_target}>/${LIBRAY_DESTINATION}/"
+			COMMAND cp -au "${_file}"  "$<TARGET_FILE_DIR:${_target}>/${LIBRAY_DESTINATION}/"
```
The-E, line 73: this seems like an accidental change? It breaks compilation on MacOS....

The-E (changes requested):

cmake/util.cmake
asarium: There was a build break in a previous iteration where a file was copied twice. I looked up the appropriate option in my `man cp` but of course Mac has to be special with their tools. I changed it to always use the CMake variant.

asarium: Whenever I change something, another thing breaks, I'll just revert these changes...

The-E (approved):

asarium:
Seems like the error is back but since no code changes were made, I will just force the merge.

PR #5223, opened 2023-03-01
Add shudder multipliers

Baezon:
Works very similarly to the recoil multipliers, only affecting camera shudder instead of physical recoil. Both the ship and the weapon can have multipliers allowing them to behave differently than their mass x velocity might suggest.


wookieejedi (approved):
Quite useful and LGTM

JohnAFernandez (approved):

PR #5225, opened 2023-03-02
Fix On Beam Fired hook variable name

Baezon:
Small oversight from #4960. This used to be "User"


BMagnu (approved):
And it's still user when the hook is called... Good catch

PR #5229, opened 2023-03-04
add some XSTR entries

### LINKED ISSUE #5227: More XSTR entries
Please could you replace the following hardcoded texts with XSTR entries?
- controlsconfigcommon.cpp (l.646): strcat_s(text, " or ");
- controlsconfigcommon.cpp (l.656): strcpy_s(text, "None");

I also found an harcoded text in the case of 'transfer-cargo' sexp: the cargo of the first ship becomes "Nothing" (hardcoded) after the transfer. I guess the required modification involves the function sexp_transfer_cargo (sexp.cpp) but I am not sure where exactly.

Goober5000:
Add some entries for "or" and "None", mostly in the controls code.  Note that there is already an XSTR entry for "none", but it does not have the right capitalization.  Also note that "None" is localized in `textify_scancode` but not `textify_scancode_universal`.

This also makes "Nothing" cargo use appropriate localization.

Fixes #5227.

### DIFF
--- a/code/localization/localize.cpp
+++ b/code/localization/localize.cpp
@@ -62,7 +62,7 @@ bool *Lcl_unexpected_tstring_check = nullptr;
 // NOTE: with map storage of XSTR strings, the indexes no longer need to be contiguous,
 // but internal strings should still increment XSTR_SIZE to avoid collisions.
 // retail XSTR_SIZE = 1570
-// #define XSTR_SIZE	1672
+// #define XSTR_SIZE	1675
 
 
 // struct to allow for strings.tbl-determined x offset
--- a/code/controlconfig/controlsconfigcommon.cpp
+++ b/code/controlconfig/controlsconfigcommon.cpp
@@ -643,7 +643,9 @@ const char *translate_key(char *key)
 
 	if (!first.empty() && !second.empty()) {
 		strcpy_s(text, first.textify().c_str());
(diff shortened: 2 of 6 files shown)


code/localization/localize.cpp
```diff
@@ -62,7 +62,7 @@ bool *Lcl_unexpected_tstring_check = nullptr;
 // NOTE: with map storage of XSTR strings, the indexes no longer need to be contiguous,
 // but internal strings should still increment XSTR_SIZE to avoid collisions.
 // retail XSTR_SIZE = 1570
-// #define XSTR_SIZE	1672
+// #define XSTR_SIZE	1675
```
z64555, line 65: This says you added 3 XSTR's, but I only counted 2 additions in your changes.  1672 seems to have been skippd?

z64555 (changes requested):
Double check usage of XSTR indices, LGTM otherwise.

code/localization/localize.cpp
Goober5000: 1672 is the "or" in controlsconfigcommon.cpp

z64555: Oh, there it is.  It didn't get the highlighting I was expecting, because the lines were completely new. 🙄

z64555 (approved):

PR #5235, opened 2023-03-05
use large asteroids speed when chucking asteroids

niffiwan:
Chucked asteroids are always large. Therefore use their speed, not the speed of the 1st parsed/small asteroid. Should not affect retail because it uses the same speed for all three asteroids

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -966,7 +966,8 @@ static void asteroid_aim_at_target(object *objp, object *asteroid_objp, float de
 		vm_vec_add2(&rand_vec, &objp->orient.vec.rvec);
 	vm_vec_normalize(&rand_vec);
 
-	speed = Asteroid_info[ASTEROID_TYPE_LARGE].max_speed * (frand()/2.0f + 0.5f);
+	auto asp = &Asteroids[asteroid_objp->instance];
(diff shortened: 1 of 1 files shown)


code/asteroid/asteroid.cpp
```diff
@@ -966,7 +966,7 @@ static void asteroid_aim_at_target(object *objp, object *asteroid_objp, float de
 		vm_vec_add2(&rand_vec, &objp->orient.vec.rvec);
 	vm_vec_normalize(&rand_vec);
 
-	speed = Asteroid_info[0].max_speed * (frand()/2.0f + 0.5f);
+	speed = Asteroid_info[ASTEROID_TYPE_LARGE].max_speed * (frand()/2.0f + 0.5f);
```
Goober5000, line 969: In the event that non-large asteroids are ever aimed, I suggest this instead:
```suggestion
	auto asp = &Asteroids[asteroid_objp->instance];
	speed = Asteroid_info[asp->asteroid_type].max_speed * (frand()/2.0f + 0.5f);
```

Goober5000 (changes requested):

code/asteroid/asteroid.cpp
niffiwan: good idea - done

Goober5000 (approved):

Issue #3968, opened 2022-01-27
$attach_thrusters: in the POF doesn't seem to work as expected


MjnMixael:
I'm honestly not sure this ever worked, but it's clear what the model creator intended. Attached here is the Albanus freighter with scripted animations (using the new animation system). The POF specifies $attach_thrusters: on both the frontpods and readpods submodels. The thruster glows do not move with the submodels when animated.

Furthermore, when the front and rear pods are setup as actual subsystems then the thruster glows don't show at all.
[albanus.zip](https://github.com/scp-fs2open/fs2open.github.com/files/7953933/albanus.zip)

wookieejedi:
Interesting... With FotG we use attach thrusters on our X-wing engines, which are subsystems and are children of animated wings (using the new animation system). Everything works as expected with that ship, but perhaps it works out for use since the engines themselves are not moving but rather their parent sub-models.

wookieejedi:
Also, it seems the model does not specify an axis for the sub models, so try specifying one. I found out today that it needs to be specified to get glow points and eyepoints to attach.

MjnMixael:
Good thought. I gave that a try, setting the submodels' rotation axis to X which didn't work. I then additionally tried again with a defined rotation axis while also setting them back up to be subsystems. But just as before that caused the thrusters to not be visible at all.

MjnMixael:
Turns out this was an incomplete feature and this flag never worked. Since the flag is now being removed entirely, this can be closed.

Baezon:
Gonna give this a shot since it shouldn't be incomplete, it should just work without any additional input.

Baezon:
For what it's worth they *do* indeed need to be subsystems as you suspected, but I'm happy to report whatever was causing them to disappear in that case seems to have been fixed!

PR #5238, opened 2023-03-06
Remove `Attach_thrusters` submodel flag

Baezon:
Even in zookeeper's original commit this flag was vestigal, and thruster movement on moving engine subsystems was done unconditionally!

https://github.com/scp-fs2open/fs2open.github.com/commit/bcf9fd90bef5c1f5229cc561d9d543ef89e81f4d


MjnMixael:
According to #3968 FOTG uses this flag?

niffiwan:
yes, `Attach_thrusters` is used on a handful of ships. But if it does nothing I'm sure we can remove the flag from the POFs. @wookieejedi ?

wookieejedi:
Good to know, I'll double check to see if removing it is safe for our models.

wookieejedi:
Yep, removing that line is safe as stated.

PR #5240, opened 2023-03-07
Adjust movement sanity checks to account for more situations

Baezon:
This bit of code to override turret rotation axes will erroneously trigger on subobjects that are children turret gun objects, and messing up the turret. So instead defer the checks until after parsing, and based on the actual base/gun pairs for each turret, instead of guessing based on the name. 

It's a little cruder than I'd like but the logic and sequencing is a bit fragile. Moreover, unless I'm mistaken I believe this operation is entirely skipped for virtual pofs with turrets added after the fact, @BMagnu may know more.

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -2856,32 +2856,22 @@ modelread_status read_model_file_no_subsys(polymodel * pm, const char* filename,
 
 	for (i = 0; i < pm->n_models; i++) {
 		if (pm->submodel[i].parent < 0) {
-			// to make sure the can_move flag is properly propagated from parent to child, we must recursively traverse through children
-			SCP_vector<int> submodel_stack;
-			submodel_stack.push_back(i);
+			model_iterate_submodel_tree(pm, i, [&](int submodel, int /*level*/, bool /*isLeaf*/)
+				{
+					auto sm = &pm->submodel[submodel];
 
-			while (!submodel_stack.empty()) {
-				int n = submodel_stack[submodel_stack.size() - 1];
-				submodel_stack.pop_back();
-				auto sm = &pm->submodel[n];
-
-				auto parent_sm = &pm->submodel[sm->parent];
-				bool is_turret = false;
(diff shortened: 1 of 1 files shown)


code/model/modelread.cpp
```diff
@@ -2860,6 +2850,23 @@ modelread_status read_model_file_no_subsys(polymodel * pm, const char* filename,
 		}
 	}
 
+	// ---------- submodel movement sanity checks ----------
+
+	for (i = 0; i < pm->n_models; i++) {
```
Goober5000, line 2855: Can we guarantee that no child subobject will be iterated before any of its parent subobjects?  The `do_movement_sanity_checks` function also sets the `Can_move` flag, and this guarantee is necessary for the flag to propagate down the submodel tree.

Goober5000 (commented):
This looks very good.  One comment.  And I too am curious what @BMagnu has to say about modular POFs.

code/model/modelread.cpp
Baezon: Ah right, that definitely isn't guaranteed tmk. Was going to handle that. Recursively traversing the submodels through childrne is much more complicated, but some thing I'm quite used to by this point in pof tools lol.

Goober5000: Hmm.  I was going to suggest using `model_iterate_submodel_tree`, but at this point in the model processing, the tree isn't set up yet.

Goober5000: Another hmm.  This is in the subobject parsing code:
```
				parent = cfread_int(fp);
				sm->parent = parent;
				auto parent_sm = parent < 0 ? nullptr : &pm->submodel[parent];
```
I'd guess this code would break if there *wasn't* a guarantee that a parent is parsed before any of its children.

Baezon: I'm not convinced it would necessarily be obvious if things failed that way, but yes, the parsing does require that parents be ordered in the file before their children, this does not however mean that their id, their position in the array, is the same order, so that must also be guarded against.

Goober5000 (approved):

PR #5244, opened 2023-03-08
Add fov values to .pof models

Baezon:
Allows parsing $base_fov and $max_fov from turret subobject properties, allowing all 3 values to be defined in the model, as well as overriden by table. Additionally cleans up the overall handling and consistency of these values a bit.

### DIFF
--- a/code/model/modelread.cpp
+++ b/code/model/modelread.cpp
@@ -711,18 +711,42 @@ void set_subsystem_info(int model_num, model_subsystem *subsystemp, char *props,
 	} else if ( strstr(lcdname, "radar") ) {
 		subsystemp->type = SUBSYSTEM_RADAR;
 	} else if ( strstr(lcdname, "turret") ) {
-		float angle;
 
 		subsystemp->type = SUBSYSTEM_TURRET;
-		if (in(p, props, "$fov"))
-			get_user_prop_value(p+4, buf);			// get the value of the fov
-		else
-			strcpy_s(buf,"180");
-		angle = fl_radians(atoi(buf))/2.0f;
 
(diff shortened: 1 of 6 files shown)


code/model/modelread.cpp
```diff
@@ -711,18 +711,42 @@ void set_subsystem_info(int model_num, model_subsystem *subsystemp, char *props,
...
+			CLAMP(value, 0.0f, 360.0f);
+			float angle = fl_radians(value) / 2.0f;
 			subsystemp->turret_fov = cosf(angle);
+		}		
+
+		if (!subsystemp->flags[Model::Subsystem_Flags::Turret_base_fov_overridden]) {
+			if (in(p, props, "$base_fov"))
+				get_user_prop_value(p + 9, buf);			// get the value of the fov
+			else
+				strcpy_s(buf, "360");
```
BMagnu, line 733: Shouldn't this set the value to -1, in order to pass the "no restriction" checks for the fov?

BMagnu (changes requested):

code/model/modelread.cpp
Baezon: It does yeah, after it goes through the math below.

BMagnu: That deserves at least a comment.
In additions, I am not really happy that this depends on float calculations resulting in an exact value, and the maximum possible value of this function to boot.
I would really like a check for == 360 while the thing is still an int.

Baezon: Ok, may as well just set the default values directly then.

BMagnu (approved):

PR #5245, opened 2023-03-08
fixup error checking in regards to sexps using dynamic parameters

MjnMixael:
This fixes some issues with how Lua SEXPs with subsystem or dockpoint parameters will work in sexp.cpp error checking. I got lucky in my earlier tests in that my specific set of test sexps passed the checks by just happening to have the right parameters in the right places to match existing built-in sexps.

My earlier code was a copy/paste from sexp_tree.cpp to here, assuming it worked the same but of course it's just slightly different. Here the code is running down the entire node tree instead of just a specific sexp, so we have to compensate for that. Instead of returning the index to the parent parameter within the single node, I need to get the index of the parent parameter within the entire sexp list. A matter of simple math. But also since in this case sexp node names count in the index, I need to ++ to account for that as well.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -2636,7 +2636,7 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
 						if (get_operator_const(op_node) < First_available_operator_id) {
 							ship_node = CDR(op_node);
 						} else {
-							int r_count = get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum) + 1; //account for the node header
+							int r_count = get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum);
 							
 							if (r_count < 0)
 								error_display(1,
@@ -2645,9 +2645,10 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
 									argnum,
 									Sexp_nodes[op_node].text);
 							
-							ship_node = 0; //initialize it I guess
+							r_count++;     // account for the node header
+							ship_node = op_node; //initialize it I guess
 							while (r_count >= 0) {
-								ship_node = CDR(op_node);
+								ship_node = CDR(ship_node);
 								r_count--;
 							}
 						}
@@ -3367,7 +3368,7 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
 							ship_node = CDR(z);
 						} else if (type == OPF_DOCKER_POINT) {
 							if (get_operator_const(op_node) >= First_available_operator_id) {
-								int r_count = get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum) + 1; //account for the node header
+								int r_count = get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum);
 								
 								if (r_count < 0)
 									error_display(1,
@@ -3376,9 +3377,10 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
 										argnum,
 										Sexp_nodes[op_node].text);
 								
-								ship_node = 0;   // initialize it I guess
+								r_count++;           // account for the node header
+								ship_node = op_node; // initialize it I guess
 								while (r_count >= 0) {
-									ship_node = CDR(op_node);
+									ship_node = CDR(ship_node);
 									r_count--;
 								}
 								break;
@@ -3388,7 +3390,7 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
 						} else if (type == OPF_DOCKEE_POINT) {
 							ship_node = CDDDR(z);
 						} else if (get_operator_const(op_node) >= First_available_operator_id) {
-							int r_count = get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum) + 1; //account for the node header
+							int r_count = get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum);
 							
 							if (r_count < 0)
 								error_display(1,
@@ -3397,9 +3399,10 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
 									argnum,
 									Sexp_nodes[op_node].text);
 							
-							ship_node = 0;   // initialize it I guess
+							r_count++;           // account for the node header
+							ship_node = op_node; // initialize it I guess
 							while (r_count >= 0) {
-								ship_node = CDR(op_node);
+								ship_node = CDR(ship_node);
 								r_count--;
 							}
 							break;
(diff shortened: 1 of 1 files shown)


The-E (approved):

code/parse/sexp.cpp
```diff
@@ -2636,9 +2636,10 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
 						if (get_operator_const(op_node) < First_available_operator_id) {
 							ship_node = CDR(op_node);
 						} else {
-							ship_node = get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum);
+							ship_node = op_node + get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum); //error checking uses the whole tree
+							ship_node++; //dynamic params are saved at a 0 based index, but with the whole tree we need to move up a spot
 
-							if (ship_node < 0)
+							if (ship_node < op_node)
```
Goober5000, line 2642: This won't get the correct node in certain cases.  Sexp node iteration is based on Lisp, and in particular uses Lisp's CAR and CDR operations.  To get the *n*th argument, do `node = CDR(node)` *n* times.  The `(ship_node < 0)` check is correct because n will be negative if there are no subsequent nodes.

code/parse/sexp.cpp
```diff
@@ -3356,15 +3357,30 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
...
-							ship_node = CDR(z);
-						else if (type == OPF_DOCKER_POINT)
+						if (get_operator_const(op_node) == OP_AI_DOCK) { // ai-dock with dockee
 							ship_node = CDR(z);
-						else if (type == OPF_DOCKEE_POINT)
+						} else if (type == OPF_DOCKER_POINT) {
+							if (get_operator_const(op_node) >= First_available_operator_id) {
+
+								ship_node = op_node + get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum); //error checking uses the whole tree
+								ship_node++; //dynamic params are saved at a 0 based index, but with the whole tree we need to move up a spot
```
Goober5000, line 3366: Use CDR here too.

code/parse/sexp.cpp
```diff
@@ -3356,15 +3357,30 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
...
+								ship_node = CDR(z);
+							}
+						} else if (type == OPF_DOCKEE_POINT) {
 							ship_node = CDDDR(z);
-						else if (get_operator_const(op_node) >= First_available_operator_id) {
+						} else if (get_operator_const(op_node) >= First_available_operator_id) {
 
-							ship_node = get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum);
+							ship_node = op_node + get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum);//error checking uses the whole tree
+							ship_node++; //dynamic params are saved at a 0 based index, but with the whole tree we need to move up a spot
```
Goober5000, line 3383: And here

Goober5000 (changes requested):

code/parse/sexp.cpp
MjnMixael: That's actually not true here. I did a lot of testing and stepping through the code this morning to see exactly what was going on. The code I wrote provides the correct index from which to check the ship. And doing it with CDR sets `ship_node` to exactly the same integer as my original code. This is truly just you being anal retentive again. I look forward to the real world example where the code I wrote originally provides the incorrect node position.

Goober5000: It is precisely the "real world" application that this guards against.  I wouldn't be surprised if, most of the time, the argument nodes are found in contiguous order.  But under heavy use the sexp node array could get fragmented, and in that situation you might find a gap in the node sequence, or a node appearing earlier in the Sexp_nodes array than its parent.  This is the nature of resource pools whose members are allocated and freed.  And if you ran across a bug in that real-world situation, it would be incredibly difficult to reproduce.  You yourself acknowledged that you got lucky in your original tests; this is just the same thing with another layer of subtlety.

code/parse/sexp.cpp
```diff
@@ -2636,13 +2636,20 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
 						if (get_operator_const(op_node) < First_available_operator_id) {
 							ship_node = CDR(op_node);
 						} else {
-							ship_node = get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum);
-
-							if (ship_node < 0)
-								error_display(1, "Expected to find a dynamic lua parent parameter for node %i in operator %s but found nothing!",
+							int r_count = get_dynamic_parameter_index(Sexp_nodes[op_node].text, argnum) + 1; //account for the node header
+							
+							if (r_count < 0)
```
Goober5000, line 2641: The +1 in the previous line is going to prevent `(r_count < 0)` from ever being true.  You could alternatively adjust the `while` condition in the subsequent loop, or turn it into a `for` loop.

code/parse/sexp.cpp
```diff
@@ -2636,13 +2636,20 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
...
+									"found nothing!",
 									argnum,
 									Sexp_nodes[op_node].text);
-
+							
+							ship_node = 0; //initialize it I guess
+							while (r_count >= 0) {
+								ship_node = CDR(op_node);
+								r_count--;
+							}
```
Goober5000, line 2652: This won't iterate over the nodes correctly.  CDR moves to the node's first child, so this will just take the first child, then the first child again, etc.  You'll need to do `ship_node = CDR(ship_node)`.  Consequently you can initialize `ship_node` to `op_node`.

code/parse/sexp.cpp
```diff
@@ -3356,22 +3363,45 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
 					} else {
 						z = op_node;
 
-						if (get_operator_const(op_node) == OP_AI_DOCK)	// ai-dock with dockee
-							ship_node = CDR(z);
-						else if (type == OPF_DOCKER_POINT)
+						if (get_operator_const(op_node) == OP_AI_DOCK) { // ai-dock with dockee
```
Goober5000, line 3366: Same comments apply to this instance and the next one.  In fact, since the same block of code is repeated three times, it would be good to put it in a function.  A 3x repetition is a common rule of thumb for this, especially if the repeated code block is more than a couple lines long.

Goober5000 (changes requested):

code/parse/sexp.cpp
MjnMixael: Done

code/parse/sexp.cpp
MjnMixael: Done

code/parse/sexp.cpp
MjnMixael: No

code/parse/sexp.cpp
MjnMixael: done again

code/parse/sexp.cpp
Goober5000: Well, the motivation for this rule of thumb is to make the coder's life easier.  But if you like doing 3x the amount of maintenance work, I won't stand in your way.

Goober5000 (approved):
Looks good now.

PR #5246, opened 2023-03-08
Add flag for star gamma correction

Baezon:
Follow-up for #5222.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -137,6 +137,7 @@ bool SCPUI_loads_hi_res_animations;
 bool Countermeasures_use_capacity;
 bool Play_thruster_sounds_for_player;
 std::array<std::tuple<float, float>, 6> Fred_spacemouse_nonlinearity;
+bool Stars_no_gamma_correction;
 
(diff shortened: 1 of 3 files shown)


PhantomHoover:
By the way, as discussed on Discord, I think we should avoid this approach in favour of just having a single acceptable behaviour for retail starfields, such as just making them uniformly brighter.

The-E:
Yeah, agreed with PH there. The retail behaviour was so obviously wrong that "disable retail stars, use a skybox" became the de facto standard amongst mission creators, the original gamma correction change made them usable again, and this change here just turns the clock back half way.

Baezon:
What exactly do you suggest? The ability to set their color beyond 255 or whatever, so they'd stay bright even when moving?

The-E:
Yes. The basic behaviour of motion-blurred stars retaining their overall brightness should be kept; what's needed here (I think) is properly integrating them into the render pipeline as actually "bright" objects, i.e. objects subject to tonemapping and bloom.

Baezon:
After discussion, I think it's agreed a flag, at least, is not the best way to approach this problem.

PR #5252, opened 2023-03-10
these should be capitalized

MjnMixael:
Small change to capitalize the built-in icon names to match all other icons generally.


JohnAFernandez (approved):
I like changes like these.

PR #5253, opened 2023-03-12
More flexible dynamic messages

naomimyselfandi:
General overhaul of the dynamic messaging system. The Big New Thing(TM) is _message filters,_ which can be used to govern when a message plays in response to specific details about the message's sender or subject (e.g. the target of an order, the attacker in a Help or Check 6 message, etc.) The old Beta/Gamma/Delta/Epsilon Arrived messages have all been folded into these filters, allowing any wing to have arrival messages defined for it.

### DIFF
--- a/code/hud/hudsquadmsg.cpp
+++ b/code/hud/hudsquadmsg.cpp
@@ -657,22 +657,22 @@ void hud_squadmsg_repair_rearm( int toggle_state, object *objp)
 	// message and bail
 	if ( is_support_allowed(tobj) ) {
 		if ( mission_is_repair_scheduled( tobj ) ) {
-			message_send_builtin(MESSAGE_ALREADY_ON_WAY, nullptr, nullptr, 0, 0, multi_player_num, multi_player_team);
+			message_send_builtin(MESSAGE_ALREADY_ON_WAY, nullptr, nullptr, multi_player_num, multi_player_team);
 		} else {
 			robjnum = hud_support_find_closest(OBJ_INDEX(tobj));
 			if ( robjnum != -1 ) {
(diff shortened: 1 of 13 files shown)


code/hud/hudsquadmsg.cpp
```diff
@@ -1239,7 +1240,7 @@ int hud_squadmsg_send_ship_command( int shipnum, int command, int send_message,
 			return 0;
 		
 		case STAY_NEAR_ME_ITEM:
-		case STAY_NEAR_TARGET_ITEM:		// can't attack anything on same team
```
wookieejedi, line 1242: Overall, also looks like there's multiple cases where comments are removed, but tagging all of those would likely be redundant at this early review stage before hearing back.

wookieejedi:
Very useful PR and highly appreciate you taking this update on! I started a first pass review with some questions to kickstart things off. Thanks!

code/hud/hudsquadmsg.cpp
```diff
@@ -1078,25 +1077,26 @@ int hud_squadmsg_send_ship_command( int shipnum, int command, int send_message,
 		Assert ( ainfo->shipnum != -1 );
 		ship_team = Ships[ainfo->shipnum].team;		// team of the ship issuing the message
 
-		switch ( command ) {									// value of k matches the #defines for ship messages
+		switch (command) {
```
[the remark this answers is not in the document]
The-E: I think so. `k` is no longer used in this version of the code, so it is pretty meaningless now

code/hud/hudsquadmsg.cpp
The-E: I don't think this particular comment is relevant either

The-E (approved):

PR #5256, opened 2023-03-14
Expose texture filtering as a built-in option

MjnMixael:
Exposes texture filtering among the graphics options in the Options Manager. I'm not 100% sure if this one would require a game restart, but for now, it does because that's safer.


The-E (approved):

The-E:
Also, for the record, this option does not _technically_ need a restart in order to work, but implementing that means either reloading all currently loaded textures or at least changing texture storage settings for every one of them

PR #5266, opened 2023-03-17
synchronize simulation timestamps across the entire frame

Goober5000:
Prior to the timer upgrade, all simulation timestamps shared identical timer and timestamp values for the duration of the frame, since they were only updated within the `game_set_frametime` function.  Since the timer upgrade depends on the system clock, timestamps evaluated near the end of the frame will be very slightly later than timestamps evaluated near the beginning of the frame.  It is desirable to keep them all in sync.

This PR tweaks the timestamp system to take a snapshot at the beginning of the frame, then use that snapshot for the entire duration of the frame.  This only affects simulation timestamps; real-time timestamps still use the live timer.

### DIFF
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -41,6 +41,9 @@ static bool Timestamp_sudo_paused = false;
 static uint64_t Timestamp_microseconds_at_mission_start = 0;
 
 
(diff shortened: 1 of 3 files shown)


Baezon (approved):

wookieejedi:
Alas `async` and mission time are not acting reliably after this commit.

PR #5267, opened 2023-03-17
fix an edge case in is-event-true-delay and related sexps

Goober5000:
Originally the check for event delay was this:
```
if ( (fix) Mission_events[i].timestamp + delay >= Missiontime )
    return SEXP_FALSE;
```
This returns false even when the mission time is equal to the timestamp.  Therefore the timestamp check needs to use >, not >=.

Thanks to Durandal and ShivanSpS for helping to track this down.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -17471,7 +17471,7 @@ int sexp_event_delay_status( int n, int want_true, bool use_msecs = false)
 			if (!Fixed_chaining_to_repeat && Mission_events[i].flags & MEF_TIMESTAMP_HAS_INTERVAL) {
 				/* do not set rval */;
 			}
-			else if (!timestamp_elapsed(timestamp_delta(Mission_events[i].timestamp, delay))) {
+			else if (!timestamp_elapsed_last_frame(timestamp_delta(Mission_events[i].timestamp, delay))) {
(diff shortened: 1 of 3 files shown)


code/parse/sexp.cpp
```diff
@@ -17471,7 +17471,7 @@ int sexp_event_delay_status( int n, int want_true, bool use_msecs = false)
 			if (!Fixed_chaining_to_repeat && Mission_events[i].flags & MEF_TIMESTAMP_HAS_INTERVAL) {
 				/* do not set rval */;
 			}
-			else if (!timestamp_elapsed(timestamp_delta(Mission_events[i].timestamp, delay))) {
+			else if (!timestamp_elapsed_last_frame(timestamp_delta(Mission_events[i].timestamp, delay))) {
```
Baezon, line 17474: Since this is a bit unusual, might be worth adding a note here about the backward compatibility this is for.

Baezon (changes requested):

code/parse/sexp.cpp
Goober5000: Ok, comment added.  LMK what you think.

Baezon: Sounds good.

Baezon (approved):

PR #5277, opened 2023-03-21
XWI import for FotG

Goober5000:
This adds the ability to import X-Wing missions into FSO using the FotG modpack.  The following milestones have been accomplished:

- Import vazor222's old branch
- Clean up mission initialization
- Import ships
- Import wings
- Import position and orientation
- Import craft AI and status
- Import cargo and special ships
- Import craft markings
- Handle arrival cues
- Handle formations

Still to do:

- Handle mission objectives
- Handle debriefing
- Handle craft orders
- Import space objects
- Handle mission end event (captured/rescued)
- Handle mission complete messages
- Import briefing


Goober5000:
An external tool would require so much infrastructure that you'd practically be duplicating the relevant parts of the code library -- i.e. data structures and mission saving.  You might as well just use the actual code and save a lot of work.  And if you're using the actual code, you would find yourself copying large parts of FRED into a new application just to add an importer.  You might as well just add the importer to FRED and save yourself the overhead of keeping two projects in sync.

code/mission/xwingbrflib.cpp
JohnAFernandez, line 1: I think these new files should go in their own subfolder if left in the mission folder to reduce clutter.  `code/mission/import/xwing` or `code/mission/xwing`

Goober5000: Not a bad idea.  I like `code\mission\import`

Goober5000: Updated

code/mission/import/xwingmissionparse.cpp
```diff
@@ -0,0 +1,743 @@
...
+		case XWMFlightGroupType::fg_Gunboat:
+			return nullptr;
+		case XWMFlightGroupType::fg_Transport:
+			return "DX-9 Stormtrooper Transport";
+		case XWMFlightGroupType::fg_Shuttle:
+			return "Lambda-class T-4a Shuttle";
+		case XWMFlightGroupType::fg_Tug:
+			return nullptr;
+		case XWMFlightGroupType::fg_Container:
+			return nullptr;
```
wookieejedi, line 249: New addition that could be used here are `DV-3 Container` and `DV-3 Cargo Freighter#NoCargo`

code/mission/import/xwingmissionparse.cpp
```diff
@@ -0,0 +1,743 @@
...
+		case XWMFlightGroupType::fg_Calamari_Cruiser:
+			return "Liberty Type Star Cruiser";
+		case XWMFlightGroupType::fg_Nebulon_B_Frigate:
+			return "EF76 Nebulon-B Escort Frigate";
+		case XWMFlightGroupType::fg_Corellian_Corvette:
+			return "CR90 Corvette";
+		case XWMFlightGroupType::fg_Imperial_Star_Destroyer:
+			return "Imperial Star Destroyer";
+		case XWMFlightGroupType::fg_TIE_Advanced:
+			return nullptr;
```
wookieejedi, line 261: Craft that could be used here is `TIE Advanced X1`

code/mission/import/xwingmissionparse.cpp
```diff
@@ -0,0 +1,743 @@
...
+
+	if (mission_event_lookup(event_name) >= 0)
+		return;
+
+	Mission_events.emplace_back();
+	auto event = &Mission_events.back();
+	event->name = event_name;
+
+	if (is_wing(fg))
+		sprintf(sexp_buf, "( when ( true ) ( fotg-wing-attacked-init \"%s\" ) )", fg_name);
```
The-E, line 89: hang on, is this an actual sexp?

I know why this is here, but I am strictly against encoding a dependency on a modpack this way. Why isn't this part of some translation table that could be shipped separately?

code/mission/import/xwingmissionparse.cpp
```diff
@@ -0,0 +1,743 @@
...
+			return "CR90 Corvette";
+		case XWMFlightGroupType::fg_Imperial_Star_Destroyer:
+			return "Imperial Star Destroyer";
+		case XWMFlightGroupType::fg_TIE_Advanced:
+			return nullptr;
+		case XWMFlightGroupType::fg_B_Wing:
+			return "B-wing Starfighter";
+		default:
+			break;
+	}
```
The-E, line 266: No. No. Absolutely not. This is unacceptable. This not only _should_ be something defined in a table somewhere, it _must_ be. 

Yes, yes, yes, I do realize that this is highly purpose-bound code and all that, but again: We should not encode this strong a dependency on anything that isn't FS2 retail.

code/mission/import/xwingmissionparse.cpp
```diff
@@ -0,0 +1,743 @@
...
+		case XWMCraftAI::ai_Rookie:
+			return "Cadet";
+		case XWMCraftAI::ai_Officer:
+			return "Officer";
+		case XWMCraftAI::ai_Veteran:
+			return "Captain";
+		case XWMCraftAI::ai_Ace:
+			return "Commander";
+		case XWMCraftAI::ai_Top_Ace:
+			return "General";
```
The-E, line 380: Again: Make this a table setting.

The-E (changes requested):

code/mission/import/xwingmissionparse.cpp
Goober5000: It is a scripted sexp in the FotG modpack.

I know what you mean, having a dependency on a mod is a code smell, but this is one of the rare times it is justified.  The XWI importer needs to process a lot of data in several very specific ways.  It isn't generalizable.  You can't create a table for that sort of information, not without it requiring a lot of pointlessly over-engineered options like this:
```
$FG Attack Check Event: FG %d Attack Check
+IfWing: ( when ( true ) ( fotg-wing-attacked-init "%s" ) )
+IfShip: ( when ( true ) ( fotg-ship-attacked-init \"%s\" ) )
```
If you know, from the task requirements, that you will only ever have one use case, then it makes sense to handle that use case as directly and simply as possible.  Creating a bunch of arbitrary hoops to jump through is just doing work for no benefit.  Premature complexity should be avoided even more strongly than premature optimization.

The only parts of the importer that could plausibly be specified in a table are the lists: the ship classes, ranks, formations, and whether a ship is a fighter or bomber.  Everything else involves a great deal of moving parts that have to fit together in a specific way and usually in a specific order.  That sort of code belongs in the importer.  It can't be done in a table, and it can't even be done in a script.

TRBlount (changes requested):
This is definitely out of scope for FSO. Even without reading through all the changes, this places too much code relating to a single mod into FSO. Disregarding how technically wrong that is, I worry that merging this could result in claims of favoritism being direct at us. 

What is the driving force behind trying to shoehorn a specific importer that isn't mod agnostic into FSO?

Goober5000:
Well, read my comment above: https://github.com/scp-fs2open/fs2open.github.com/pull/5277#discussion_r1198515691

I don't think this is out of scope for FSO, and nor do I think it is favoritism.  If someone were to design another importer, say, one that imported Wing Commander missions and was dependent on the WCSaga mod, I would have no issue including it, providing the menu option and the code were well-compartmentalized (as is done here for X-Wing).

Specific importers are necessarily going to be dependent on the mods that they are designed for.  I mean, you can't use the FS1 mission importer without either FSPort or another mod that provides the FS1 ship and weapon entries.

Baezon:
Reposting from @PhantomHoover from the discord
> what i’m proposing is that it can be as messy and hardcoded as you like, just don’t merge it. make special purpose builds of it that people can use when they need the feature
it’s basically just using a custom build of FSO as if it’s an external tool
this wouldn’t necessarily be appropriate if, say, fotg is anticipating advertising XWI mission import as a headline feature for novice modders. but if it’s only gonna be used by a couple of people who know what they’re doing anyway they’re not gonna have trouble managing a special build for it

This sounds like the best of both worlds, none of the invested work is wasted, and can be as dialed-in to FotG as necessary, and the SCP needn't worry about maintaining it.

MageKing17:
> Specific importers are necessarily going to be dependent on the mods that they are designed for. I mean, you can't use the FS1 mission importer without either FSPort or another mod that provides the FS1 ship and weapon entries.

This is, IMO, more of an argument for stripping out the FS1 mission importer than it is for merging this.

Goober5000:
> Reposting from @PhantomHoover from the discord
> 
> > what i’m proposing is that it can be as messy and hardcoded as you like, just don’t merge it. make special purpose builds of it that people can use when they need the feature
> > it’s basically just using a custom build of FSO as if it’s an external tool
> > this wouldn’t necessarily be appropriate if, say, fotg is anticipating advertising XWI mission import as a headline feature for novice modders. but if it’s only gonna be used by a couple of people who know what they’re doing anyway they’re not gonna have trouble managing a special build for it
> 
> This sounds like the best of both worlds, none of the invested work is wasted, and can be as dialed-in to FotG as necessary, and the SCP needn't worry about maintaining it.

Yeah, this sounds like the best way to keep all parties happy.

Goober5000:
Closing in favor of moving this PR to the FotG repository: https://github.com/Fate-of-the-Galaxy/fs2open.github.com/pull/2

PR #5279, opened 2023-03-22
Add clip planes to game_settings.tbl

Baezon:
And remove the -clipdist cmd line parameter that used to also control it. Will need some hefty documentation on the wiki to explain what these are for, how they work, how they relate to depth testing and z-fighting, etc...

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -790,6 +790,29 @@ void parse_mod_table(const char *filename)
 			stuff_boolean(&SCPUI_loads_hi_res_animations);
 		}
 
+		if (optional_string("$Far clip plane:")) {
+			stuff_float(&Max_draw_distance);
+
+			if (fl_near_zero(Max_draw_distance)) {
+				Warning(LOCATION, "The $Far clip plane must be above 0. Using default value.\n");
+				Max_draw_distance = Default_max_draw_distance;
+			} 
+		}
+
+		if (optional_string("$Near clip plane:")) {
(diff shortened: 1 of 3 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -790,6 +790,29 @@ void parse_mod_table(const char *filename)
 			stuff_boolean(&SCPUI_loads_hi_res_animations);
 		}
 
+		if (optional_string("$Far clip plane:")) {
```
The-E, line 793: I think naming these parameter "Max draw distance" and "Min draw distance" is clearer and allows us to get away with not having to explain the concept of a clip plane to users

code/mod_table/mod_table.cpp
```diff
@@ -790,6 +790,29 @@ void parse_mod_table(const char *filename)
...
+		}
+
+		if (optional_string("$Near clip plane:")) {
+			stuff_float(&Min_draw_distance);
+
+			if (fl_near_zero(Min_draw_distance)) {
+				Warning(LOCATION, "The $Near clip plane must be above 0. Using default value.\n");
+				Min_draw_distance = Default_min_draw_distance;
+			}
+			if (Min_draw_distance >= Max_draw_distance) {
```
The-E, line 809: This check should be pulled out of the Near clip plane conditional. We need to ensure that min<max regardless of which of these values is set; checking only one is insufficient.

The-E (changes requested):

The-E (approved):

PR #5282, opened 2023-03-23
FIX: Incorrect gamma calculations in shaders.

qazwsxal:
Noticed an issue with gamma correction being overzealously applied to the emissive channel, resulting in envmap shines being too dim, likely introduced by #5055 . 
The emissive channel is a bit weird, as multiples sources are added together during g-buffer creation, this means it's better to convert everything to linear colour space when you're reading from the textures rather than waiting until the end.


The-E (approved):

qazwsxal:
23.0 
screen0058
nightlies:
screen0061
This patch:
screen0065

PR #5285, opened 2023-03-24
Modular messages table

### LINKED ISSUE #4626: FR: Modular messages table
Maybe worth leaving on the shelf for parse items, very low priority but it would be nice to have so here it is. Warmachine's teetering modstack aside I just generally work better with 'single topic' files. One TBM per persona, for instance.

MjnMixael:
Makes messages.tbl extendable with -msg.tbm. Fixes #4626

Also vectorizes Personas to make the modularization a little easier.

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -206,7 +206,7 @@ void persona_parse()
 	
 	bool dup = false;
 	if (message_persona_name_lookup(this_persona.name) >= 0) {
-		Warning(LOCATION, "Duplicate Persona %s found!", this_persona.name);
+		Warning(LOCATION, "Duplicate Persona %s found, ignoring!", this_persona.name);
 		dup = true;
 	}
 
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -127,8 +127,7 @@ int MessageQ_num;			// keeps track of number of entries on the queue.
 #define MESSAGE_SOON_TIMESTAMP			5000		// "soon" messages must play within 5 seconds
 
 // Persona information
-int Num_personas;
-Persona *Personas = NULL;
+SCP_vector<Persona> Personas;
 
 const char *Persona_type_names[MAX_PERSONA_TYPES] =
 {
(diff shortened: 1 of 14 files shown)


code/mission/missionmessage.cpp
```diff
@@ -196,52 +195,31 @@ int comm_between_player_and_ship(int other_shipnum, bool for_death_scream);
...
-	memset(&Personas[Num_personas], 0, sizeof(Persona));
+	Persona this_persona;
 
 	required_string("$Persona:");
-	stuff_string(Personas[Num_personas].name, F_NAME, NAME_LENGTH);
+	stuff_string(this_persona.name, F_NAME, NAME_LENGTH);
+	
+	bool dup = false;
+	if (message_persona_name_lookup(this_persona.name) >= 0) {
+		Warning(LOCATION, "Duplicate Persona %s found!", this_persona.name);
```
JohnAFernandez, line 209: Should probably say that this persona is going to be ignored, too.

code/mission/missionmessage.cpp
```diff
@@ -728,17 +720,35 @@ void messages_init()
...
+		// parse any modular tables
+		parse_modular_table("*-msg.tbm", parse_msgtbl);
+
+		// save the number of builtin message things -- make initing between missions easier
+		Num_builtin_messages = Num_messages;
+		Num_builtin_avis = Num_message_avis;
+		Num_builtin_waves = Num_message_waves;
+
+		table_read = 1;
 	}
```
JohnAFernandez, line 742: Exceptions will no longer be thrown?

JohnAFernandez (changes requested):

code/mission/missionmessage.cpp
MjnMixael: try/catch exists within `parse_msgtbl()`. I didn't think we needed it here also because any parsing errors will happen within that method.

code/mission/missionmessage.cpp
MjnMixael: Good call. Done.

JohnAFernandez (approved):

PR #5286, opened 2023-03-25
Force expand wings instead of toggle for scpui

MjnMixael:
Toggle caused the list to load different each time it was called. This enforces the same behavior every time.


JohnAFernandez (approved):
LGTM

PR #5287, opened 2023-03-25
Revert "synchronize simulation timestamps across the entire frame"

wookieejedi:
EDIT: reverts scp-fs2open/fs2open.github.com#5266

Alas async and mission time are not acting reliably after this commit. Seems like removing the `if (start_frame)` check fixes things, but of course that's not the proper fix since that just restores the old behavior. ~~I've tried tracking things down and updating places where I think things should go, but no luck so far. It looks like Goober is pretty busy with RL ATM too, so I'm going to post this revert PR to fix the bug in the interim.~~

From Goober on Discord:
```This will need a proper investigation but I don't know how much time I'll have in the next few days to look at it. I think 5287 is thinking along the right lines but is not actually the correct fix. I'd say that leaving the missiontime and time compression bugs in the nightly builds would be a lot more disruptive than simply reverting to the pre-21st behavior.  So we should revert 5266 for now, then re-add 5266 with the time compression fix once we figure out what's going on. ```

Did some more examinations, seems the timestamp goes negative at mission start b/c `Timestamp_offset_from_counter` is larger than the value of `get_performance_counter()`.
```
Entering game at time =  12.194

start frame was true for timestamp

 'timestamp_raw' = 121939419 (get_performance_counter())
 
 'Timestamp_offset_from_counter' = 198181376 

 'timestamp_raw' updated = -76241957 (raw - offset_counter) 
```

For comparison, with the older behavior this is what an example of what happens
```
Entering game at time =  19.070

start frame was true for timestamp (always)
  
  'timestamp_raw' = 190700853 (get_performance_counter())

  'Timestamp_offset_from_counter' = 190648661 

  'timestamp_raw updated' = 52192 (raw - offset_counter) 
```

Ultimately  I traced the reason that the `Timestamp_offset_from_counter` gets too high is because in `timestamp_update_time_compression()` the call to `timestamp_get_raw` was using the new default false argument. It looks like it needs to have the new `true` optional argument added instead. EDTI: that alone does not quite fix time compression though. 

```
	// grab the independent variable of the equation before we change anything
	auto timestamp_raw = timestamp_get_raw();

	// we need to move the counter offset to make the raw timestamp zero (so that it can start ticking with a new multiplier)
	Timestamp_offset_from_counter += timestamp_raw;
	Timestamp_paused_at_counter += timestamp_raw;
```

### DIFF
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -41,9 +41,6 @@ static bool Timestamp_sudo_paused = false;
 static uint64_t Timestamp_microseconds_at_mission_start = 0;
 
 
-static uint64_t timestamp_get_raw(bool start_frame = false);
-
-
 static uint64_t get_performance_counter()
 {
 	Assertion(Timer_inited, "This function can only be used when the timer system is initialized!");
@@ -76,12 +73,6 @@ void timer_init()
 
 		atexit(timer_close);
 	}
-}
-
-void timer_start_frame()
-{
-	// take a snapshot of the raw timestamp at the beginning of the frame
-	timestamp_get_raw(true);
 }
 
 // ======================================== getting time ========================================
@@ -138,21 +129,15 @@ std::uint64_t timer_get_nanoseconds()
     return static_cast<uint64_t>(time * Timer_to_nanoseconds);
 }
 
-static uint64_t timestamp_get_raw(bool start_frame)
+static uint64_t timestamp_get_raw()
 {
-	static uint64_t timestamp_raw = 0;
-
-	// The simulation timestamp is only updated at the beginning of the frame
-	// because we want all timestamps within a frame to be identical.
-	if (start_frame)
-	{
-		if (Timestamp_is_paused)
-			timestamp_raw = Timestamp_paused_at_counter;
-		else
-			timestamp_raw = get_performance_counter();
-
-		timestamp_raw -= Timestamp_offset_from_counter;
+	uint64_t timestamp_raw;
+	if (Timestamp_is_paused) {
+		timestamp_raw = Timestamp_paused_at_counter;
+	} else {
+		timestamp_raw = get_performance_counter();
 	}
+	timestamp_raw -= Timestamp_offset_from_counter;
 
 	return timestamp_raw;
 }
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -41,6 +41,9 @@ static bool Timestamp_sudo_paused = false;
 static uint64_t Timestamp_microseconds_at_mission_start = 0;
 
 
+static uint64_t timestamp_get_raw(bool start_frame = false);
+
+
 static uint64_t get_performance_counter()
 {
 	Assertion(Timer_inited, "This function can only be used when the timer system is initialized!");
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by wookieejedi** [comment]
Bit of an update after talking with Durandal: 

So #4633 likely caused event timings to not work for missions Durandal was working on.
#5267 was supposed to fix it, but as it turns out #5266 was needed in addition to actually fix it. 
But now with #5266 merged, the async and mission times are not working in other missions. 

Lafiel also mentioned `I suspect the issue is that we have things relying on frametime that don't happen in freespace.cpp's frame method`, which prompted me to try and put this in places that I thought it should go but to no avail (as posted above). 

So overall a true and proper fix is needed in any case.

**Comment by Baezon** [comment]
Could you be more specific as to how exactly async and missiontime aren't working? As you mentioned the change was made to fix buggy behavior and I'm a little hesitant to revert back to different, but still buggy behavior.

**Comment by wookieejedi** [comment]
Sure! So one example is:
```
$State: GS_STATE_GAME_PLAY
$On State Start: [
async.run(function()
	async.await(mn.waitAsync(0.1))
end)
]
```

The 0.1 delay will not trigger until 5-15 seconds or more into the mission. 
Another example is with a mission where ships a moving at the mission start, with sitting in the briefing for a while then going into the mission it seems that up to a minute or more has passed in mission based on the location of the ships.

**Comment by Baezon** [comment]
For what it's worth, I'm not able to replicate with just that script snippet. This sounds worse now, but I really don't understand this code and I'm not able to do much debugging until after the weekend so I'm quite hesitant to rubberstamp this myself.

**Comment by wookieejedi** [comment]
Thanks taking a look at it. Yeah that makes sense to me, so I'm good with just having this PR open as a way to help facilitate a proper fix.

**Comment by JohnAFernandez** [comment]
As I posted on SCP internal I have a guess:

What I'm reading suggests that `Timestamp_microseconds_at_mission_start` is set incorrectly.  `timestamp_get_microseconds()` *now* gets an updated value once per frame, but that value is not updated unless `timer_start_frame()` has been called.  During gameplay this is done at the start of the frame, as it should be.  But at mission load when `Timestamp_microseconds_at_mission_start` is set, `timer_start_frame()` is not called and the value `timestamp_get_microseconds()` fetches to set `Timestamp_microseconds_at_mission_start` is stale.

So if I'm right, we should change

```void timestamp_start_mission()
{
    Timestamp_microseconds_at_mission_start = timestamp_get_microseconds();
}
```

to this:

```void timestamp_start_mission()
{
        timer_start_frame();
    Timestamp_microseconds_at_mission_start = timestamp_get_microseconds();
}
```

because that would allow `timestamp_get_microseconds()` to retrieve an up to date value and hopefully fix these items:

"The 0.1 delay will not trigger until 5-15 seconds or more into the mission.
Another example is with a mission where ships a moving at the mission start, with sitting in the briefing for a while then going into the mission it seems that up to a minute or more has passed in mission based on the location of the ships."

wookieejedi:
Ah I had the same thought and tried that as a part of my earlier tests, but alas it did not work

wookieejedi:
Updated original comment post with details from additional examinations, so additional eyes would be good to have again.

JohnAFernandez (approved):
Based on what I saw, this fix makes sense.

wookieejedi:
Just noting this here from Discord:

I sent the build to Durandel since he had the missions that the fixes on the 21st were supposed to fix and he said he should be able to test it later today.

wookieejedi:
UPDATE: Durandal tested events that were not working before the #5266, and the seem to work with this fix. Overall since #5266 time compression does not reliably work in regards to events that rely on timers though. This PR does not fix that aspect.

PR #5289, opened 2023-03-25
restored the keyboard hotkey for envmap export.

EatThePath:
There's a gui button for this too, but the hotkey is still listed in the ui text and it's a bit more convenient for doing large batches of these, so I put it back in.


wookieejedi (approved):
LGTM

PR #5290, opened 2023-03-25
Dehardcode Command arrival message chance

naomimyselfandi:
Historically, there's been a hardcoded 25% chance for Command to announce an enemy wing arrival. This is problematic for mods without Command, like BTA.


JohnAFernandez (approved):
LGTM

Issue #5278, opened 2023-03-22
FR: Additional live-debris options.


DefCynodont:
Three hopefully minor requests: (in descending order of priority)

- [x] A table option to allow live-debris spawned by a destroyed subsystem to collide with the ship that the subsystem was a part of. 
(this please)

- [x] A game setting or mission setting to make Asteroids and Ship Debris / Live-Debris affected by weapon mass (IE: pushed by weapons)

- [x] A way to multiply (by percentage) the mass value of live-debris and/or ship debris. In the case of live-debris, it should be by-subsystem.

Baezon:
Apparently, #1 is already a thing1

MjnMixael:
This is complete, now, right?

Baezon:
Yep! Forgot to close it.

PR #5291, opened 2023-03-28
Whackable debris and asteroids

Baezon:
Part of #5278. With the relevant flags, debris and asteroids can be whacked around too.

### DIFF
--- a/code/network/multimsgs.cpp
+++ b/code/network/multimsgs.cpp
@@ -6889,12 +6889,11 @@ void send_asteroid_hit( object *objp, object *other_objp, vec3d *hitpos, float d
 {
 	int packet_size;
 	ubyte data[MAX_PACKET_SIZE], packet_type;
-	vec3d vec = vmd_zero_vector;
 
 	if ( hitpos == nullptr )
-		hitpos = &vec;
+		hitpos = &vmd_zero_vector;
 	if ( force == nullptr )
-		force = &vec;
+		force = &vmd_zero_vector;
 
 	// build up an asteroid hit packet
 	BUILD_HEADER( ASTEROID_INFO );
--- a/code/network/multimsgs.cpp
+++ b/code/network/multimsgs.cpp
@@ -6889,11 +6889,12 @@ void send_asteroid_hit( object *objp, object *other_objp, vec3d *hitpos, float d
 {
 	int packet_size;
 	ubyte data[MAX_PACKET_SIZE], packet_type;
-	vec3d vec;
+	vec3d vec = vmd_zero_vector;
 
-	vm_vec_zero(&vec);
-	if ( hitpos != NULL )
-		vec = *hitpos;
+	if ( hitpos == nullptr )
+		hitpos = &vec;
+	if ( force == nullptr )
+		force = &vec;
 
 	// build up an asteroid hit packet
 	BUILD_HEADER( ASTEROID_INFO );
@@ -6907,7 +6908,7 @@ void send_asteroid_hit( object *objp, object *other_objp, vec3d *hitpos, float d
 	} else {
 		ADD_USHORT( other_objp->net_signature );
 	}
-	ADD_VECTOR( vec );
+	ADD_VECTOR( (*hitpos) );
 	ADD_FLOAT( damage );
 	ADD_VECTOR( (*force) );
 	
--- a/code/network/multimsgs.cpp
+++ b/code/network/multimsgs.cpp
@@ -5503,7 +5503,7 @@ void process_debris_update_packet(ubyte *data, header *hinfo)
 	// blow it up
 	case DEBRIS_UPDATE_NUKE:
 		if(objp != &bogus_object)
-			debris_hit(objp,NULL,&objp->pos,1000000.0f);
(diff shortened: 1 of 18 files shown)


code/weapon/weapons.cpp
```diff
@@ -7054,45 +7054,66 @@ int weapon_area_calc_damage(object *objp, vec3d *pos, float inner_rad, float out
...
+			model_num = Ship_info[Ships[objp->instance].ship_info_index].model_num;
+			break;
+		case OBJ_ASTEROID:
+			model_num = Ship_info[Ships[objp->instance].ship_info_index].model_num;
+			break;
+		case OBJ_DEBRIS:
+			model_num = Ship_info[Ships[objp->instance].ship_info_index].model_num;
+			break;
+		default:
+			UNREACHABLE("unhandled object type");
```
JohnAFernandez, line 7097: I would add to this error to ask the user to reach out to SCP.

JohnAFernandez (changes requested):
This just needs a multi bump, but otherwise, it's fine.

code/network/multimsgs.cpp
```diff
@@ -6909,6 +6909,7 @@ void send_asteroid_hit( object *objp, object *other_objp, vec3d *hitpos, float d
 	}
 	ADD_VECTOR( vec );
 	ADD_FLOAT( damage );
+	ADD_VECTOR( (*force) );
```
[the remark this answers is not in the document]
notimaginative: `force` is also optional so that needs to be guarded against like with `hitpos`. Right now this code is just going to crash half the time.

code/weapon/weapons.cpp
```diff
@@ -7054,45 +7054,66 @@ int weapon_area_calc_damage(object *objp, vec3d *pos, float inner_rad, float out
...
  * @param blast				Force of blast
  * @param make_shockwave	Boolean, whether to create a shockwave or not
  */
-void weapon_area_apply_blast(vec3d * /*force_apply_pos*/, object *ship_objp, vec3d *blast_pos, float blast, int make_shockwave)
+void weapon_area_apply_blast(vec3d * /*force_apply_pos*/, object *objp, vec3d *blast_pos, float blast, bool make_shockwave)
 {
 	vec3d		force, vec_blast_to_ship, vec_ship_to_impact;
 	polymodel		*pm;
 
+	Assertion(objp->type == OBJ_SHIP || objp->type == OBJ_ASTEROID || objp->type == OBJ_DEBRIS, "weapon_area_apply_blast can only be called on ships asteroids or debris");
```
Goober5000, line 7070: commas would be nice in this message

code/weapon/weapons.cpp
```diff
@@ -7054,45 +7054,66 @@ int weapon_area_calc_damage(object *objp, vec3d *pos, float inner_rad, float out
...
+	switch (objp->type) {
+		case OBJ_SHIP: 
+			model_num = Ship_info[Ships[objp->instance].ship_info_index].model_num;
+			break;
+		case OBJ_ASTEROID:
+			model_num = Ship_info[Ships[objp->instance].ship_info_index].model_num;
+			break;
+		case OBJ_DEBRIS:
+			model_num = Ship_info[Ships[objp->instance].ship_info_index].model_num;
+			break;
```
Goober5000, line 7095: Er, the models for asteroids and debris are using the wrong arrays/vectors here.

You could use `object_get_model` if you want.

Baezon: Whoops. Thanks, I looked for a function like that but I couldn't find it

code/network/multimsgs.cpp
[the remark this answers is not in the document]
Baezon: How would I go about a multi bump?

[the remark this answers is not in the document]
Goober5000: Yep, bumping `MULTI_FS_SERVER_VERSION` will do it.

code/network/multimsgs.cpp
```diff
@@ -6889,11 +6889,12 @@ void send_asteroid_hit( object *objp, object *other_objp, vec3d *hitpos, float d
 {
 	int packet_size;
 	ubyte data[MAX_PACKET_SIZE], packet_type;
-	vec3d vec;
+	vec3d vec = vmd_zero_vector;
```
Goober5000, line 6892: It looks like you could avoid using `vec` at all, and just do e.g. `hitpos = &vmd_zero_vector`

Baezon: If you mean form the callers prespective, I had considered this, but I was a little bit worried about future-proofing I guess, and figured it would be safest to simply assume that callers might provide nullptr.

Goober5000: I mean that the `vec3d vec` temporary variable isn't needed here, since if `hitpos` and/or `force` are null, they can be pointed to the zero vector.  But it's a trivial optimization.

Baezon: Oh nvm, I see now.

JohnAFernandez (approved):

PR #5292, opened 2023-03-28
Adjustments to max vel changes from ETS

Baezon:
The ship's `current_max_speed` stores the max speed given the ship's ETS settings and it's what most things will read from, while every frame its own physics `phys_info.max_vel.xyz.z` is updated to that value. This indirection is harmful because it basically prevents any modification of a ship's max velocity through scripting through its physics values. Instead the `current_max_speed` should be removed, and `phys_info.max_vel.xyz.z` should be modified and acted upon directly, they store the exact same value (which was also enforced every frame).

Other changes in hudets.cpp meant to facilitate script access is that max speed should only be adjusted when the engine energy is actually adjusted, prior the speed was set based on engine energy every frame, even on ships without ETS.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -6534,7 +6534,6 @@ void ship::clear()
 	weapon_recharge_index = INTIAL_WEAPON_RECHARGE_INDEX;
 	engine_recharge_index = INTIAL_ENGINE_RECHARGE_INDEX;
 	weapon_energy = 0;
-	current_max_speed = 0.0f;
 	next_manage_ets = timestamp(0);
 
 	flags.reset();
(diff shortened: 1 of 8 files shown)


code/ship/ship.cpp
```diff
@@ -11399,14 +11396,6 @@ void change_ship_type(int n, int ship_type, int by_sexp)
...
 	}
 
-	// avoid cases where either of these are 0
-	if (sp->current_max_speed != 0 && sip_orig->max_speed != 0) {
-		sp->current_max_speed = sip->max_speed * (sp->current_max_speed / sip_orig->max_speed);
-	}
-	else {
-		sp->current_max_speed = sip->max_speed;
-	}
-
```
Baezon, line 11409: Previously the (proportional) max speed would've been maintained, but ETS settings *would not* (in `ets_init_ship` right below). Resulting in inconsistent scenarios where the max speed is higher than the current energy setting should allow.

I've decided I err on the side of resetting the speed entirely, along with energy settings, but I would be amenable to maintaining energy settings (and the max speed with it).

Baezon:
`shipclass` already contains `VelocityMax` for the tabled values, and `physics` contains the current values, if that's what you mean.

wookieejedi:
Oh that's right, I've been away from that script section for too long it seems. Thanks for the follow up!

wookieejedi (approved):

PR #5297, opened 2023-03-31
$Beaminfo implies "beam" flag!

Baezon:
Fix for an annoying case I ran into, beam weapon behaved oddly because I forgot to add the "beam" flag. These flags are dumb. If they specified any $Beaminfo, it's a beam full-stop.


wookieejedi (approved):
Makes sense to me

PR #5300, opened 2023-03-31
Add $CockpitLaunchSnd

Baezon:
Although $LaunchSnd will attempt to play in 2d when appropriate, given that it is expected to be used in both 3d (others shooting) and 2d (you shooting) contexts, in practice modders are forced to make them 3d, and are unable to make use of nice stereo for their weapons. This adds $CockpitLaunchSnd, a separate sound only for the player, so the modder can use an appropriate dedicated 2d sound.

As always, diff is a bit cleaner with 'hide whitespace changes'.


wookieejedi (approved):
Quite useful, and LGTM

Issue #5299, opened 2023-03-31
stack overflow crash in new message code


Goober5000:
The following messages.tbl, used by the currently-in-progress Terminus mod, causes FSO to crash with a stack overflow:
[messages.tbl.txt](https://github.com/scp-fs2open/fs2open.github.com/files/11117967/messages.tbl.txt)

This happens to be a buggy messages.tbl, in the sense that the data is incomplete, but the situation should ideally be handled by a Warning somewhere.

Steps to reproduce:
1. Add the messages.tbl to a base FS2 install
2. Load any FS2 mission with support enabled
3. Call for support

The messages.tbl lists "Terran Command" as the sole Command persona, but all of the Command senders have been changed to "Sengoku Control".  The `get_builtin_message` function is called for the "Rearm warping in" message, with a requirement for an exact persona match.  However, no message match is found because no message is sent by "Terran Command".  This part works properly, but the subsequent part recurses into an endless loop.

There is one identifiable bug here, which may or may not be the root cause: when the `fallback` message is `MESSAGE_NONE`, the code recurses into `get_builtin_message(fallback, persona, sender, subject, false);`, but this causes `Builtin_messages` to check an index of -1.  Eventually `Builtin_messages[-1].fallback` is checked, which evaluates to 0, or Wingman 1.  Then Wingman 1 falls back to Terran Command, and the cycle repeats.

Looks to be caused by #5253, so pinging @naomimyselfandi for insight.

naomimyselfandi:
Well, I can't fault your analysis, and this could really happen with any incomplete persona (off the top of my head, dying with BP's Olefumi around could trigger it). Expect a PR this evening Eastern time.

PR #5304, opened 2023-04-02
Message filter follow up

naomimyselfandi:
Fixes the issue identified in #5299 and another that BTA noticed with support ships and the new auto-assignment configuration.

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -2042,7 +2042,7 @@ bool excludes_current_mood(int message) {
 	return false;
 }
 
-int get_builtin_message(int type, int persona, ship* sender, ship* subject, bool require_exact_persona_match) {
+int get_builtin_message_inner(int type, int persona, ship* sender, ship* subject, bool require_exact_persona_match) {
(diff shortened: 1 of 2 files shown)


code/mission/missionmessage.cpp
```diff
@@ -2121,15 +2126,23 @@ int get_builtin_message(int type, int persona, ship* sender, ship* subject, bool
 	} else {
 		int fallback = Builtin_messages[type].fallback;
 		if (fallback != MESSAGE_NONE) {
-			return get_builtin_message(fallback, persona, sender, subject, true);
-		} else if (require_exact_persona_match && persona_allows_substitution(persona)) {
```
Goober5000, line 2125: This is the only place where `persona_allows_substitution` is referenced.  Should it be added somewhere else?

naomimyselfandi: Yes, thank you for catching that.

Goober5000 (approved):
Ok, looks good AFAICT

wookieejedi:
Agreed

PR #5307, opened 2023-04-03
Explicitely do not split lines when making the message log

MjnMixael:
Originally I had set this to just use the screen width for message lines, assuming no line could be more than that and lines would never be split for SCPUI. In hindsight, that was dumb and of course lower resolutions would run into problems. So this makes line splitting an explicit choice.

### DIFF
--- a/code/mission/missionlog.cpp
+++ b/code/mission/missionlog.cpp
@@ -605,7 +605,7 @@ void message_log_add_seg(log_text_seg* entry, int x, int msg_color, const char* 
 	entry->flags = flags;
 }
 
-void message_log_add_segs(const char* source_string, int msg_color, int flags = 0, SCP_vector<log_text_seg> *entry = nullptr)
+void message_log_add_segs(const char* source_string, int msg_color, int flags = 0, SCP_vector<log_text_seg> *entry = nullptr, bool split_string = false)
 {
 	if (!source_string) {
 		mprintf(("Why are you passing a NULL pointer to message_log_add_segs?\n"));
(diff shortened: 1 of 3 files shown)


JohnAFernandez (approved):
LGTM

code/mission/missionlog.h
```diff
@@ -115,7 +115,7 @@ extern int message_log_color_get_team(int msg_color);
 // get the actual color for a line item
 extern color log_line_get_color(int tag);
 
-void message_log_init_scrollback(int pw);
+void message_log_init_scrollback(int pw, bool split_string = true);
```
Goober5000, line 118: Is it worthwhile making `split_string` a required parameter, rather than an optional one?

code/mission/missionlog.cpp
```diff
@@ -605,7 +605,7 @@ void message_log_add_seg(log_text_seg* entry, int x, int msg_color, const char*
 	entry->flags = flags;
 }
 
-void message_log_add_segs(const char* source_string, int msg_color, int flags = 0, SCP_vector<log_text_seg> *entry = nullptr)
+void message_log_add_segs(const char* source_string, int msg_color, int flags = 0, SCP_vector<log_text_seg> *entry = nullptr, bool split_string = false)
```
Goober5000, line 608: Same question.

MjnMixael: It's used more than a handful of times and only one of those times (lua api) needs this parameter set to false. It makes more sense to leave it as optional, default true.

Goober5000: Ok.

PR #5311, opened 2023-04-03
Add +Use Ship as Template

BMagnu:
Allows to table sort of derived ships.
Behaves pretty much identical to templates (which should be preferred), but allows to use ships that may be present in a parent mod or are already tabled as templateless ships to be used as such.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -2228,9 +2228,29 @@ static void parse_ship(const char *filename, bool replace)
 				sip->clone(Ship_templates[template_id]);
 				strcpy_s(sip->name, fname);
 				new_name = true;
+				create_if_not_found = false;
 			}
 			else {
 				Warning(LOCATION, "Unable to find ship template '%s' requested by ship class '%s', ignoring template request...", template_name, fname);
+			}
+		}
+	}
+	if( optional_string( "+Use Ship as Template:" ) ) {
+		if ( !create_if_not_found ) {
+			Warning(LOCATION, "Both '+nocreate' and '+Use Ship as Template:' were specified for ship class '%s', ignoring '+Use Ship as Template:'\n", fname);
+		}
+		else {
+			char template_name[SHIP_MULTITEXT_LENGTH];
+			stuff_string(template_name, F_NAME, SHIP_MULTITEXT_LENGTH);
+			int template_id = ship_info_lookup_sub(template_name);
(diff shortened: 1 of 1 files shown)


code/ship/ship.cpp
```diff
@@ -2228,12 +2228,32 @@ static void parse_ship(const char *filename, bool replace)
...
+				create_if_not_found = false;
 			}
 			else {
 				Warning(LOCATION, "Unable to find ship template '%s' requested by ship class '%s', ignoring template request...", template_name, fname);
 			}
 		}
 	}
+	if( optional_string( "+Use Ship as Template:" ) ) {
+		if ( !create_if_not_found ) {
+			Warning(LOCATION, "Both '+nocreate' and '+Use Ship as Template:' were specified for ship class '%s', ignoring '+Use Ship as Template:'\n", fname);
```
MageKing17, line 2240: This conditional could also be reached if `+Use Template:` and `+Use Ship as Template:` were used in the same entry, so this warning might not be accurate.

MageKing17 (commented):
There's a reason Variants aren't already in the codebase; namely, what happens if the ship you're using as a "template" is modified by a modular table parsed after this one?

code/ship/ship.cpp
BMagnu: Fair point on the warning.
As for the modification of a ship after a template is defined:
I want to warn about this exact possibility in the documentation on the wiki. It is _explicitly_ possible to modify a ship after cloning it. This is intended as an expert option, for situations where you need to clone from upstream mods, whereas normal users are supposed to use templates anyways.
If _really_ necessary, I could add a Ship-Info flag that warns about modifying a cloned ship later down the line unless explicitly flagged as intended when modifying the ship, but I think that's overkill

MageKing17: > If _really_ necessary, I could add a Ship-Info flag that warns about modifying a cloned ship later down the line unless explicitly flagged as intended when modifying the ship, but I think that's overkill

I agree; that sounds like overkill. As long as the documentation for the feature explicitly points out that any changes to the base ship class in later-parsed entries won't be reflected in ship classes that used them as a template, I think adding this as an option should be fine.

MageKing17 (approved):
Well, I don't like the amount of duplicated code (offends my DRY sensibilities), but not enough to hold up this PR.

Issue #5226, opened 2023-03-03
FR: make per-mission "Player starts in chase view" a toggle instead?


dariusbei:
FRED currently has a "Player starts in chase view" mission flag in the mission specs window, but for mods  with  
"$Player starts in third person/chase view by default: TRUE" 
in game_settings.tbl there's no similar function for player starting in first person/cockpit view.
Would it be possible to have the feature a toggle? Or implement a similar flag for first person?

![fred](https://user-images.githubusercontent.com/47288956/222699641-e0afd6df-a2a4-47aa-b97d-20f98205477a.JPG)

Goober5000:
Yes, this should be as simple as changing the "No Debriefing" to "Toggle Debriefing" flag was.

Goober5000:
Implemented in #5312

PR #5312, opened 2023-04-05
make 'player starts in chase view' a toggle

Goober5000:
Per request of #5226.  This PR also makes the variable names more consistent.


Baezon (changes requested):
A minor addition I'd like to see, if you don't mind me throwing one on, is that ESC, if in chase view, will revert to cockpit view, before then triggering the quit menu. ESC should probably instead revert from the non-default view to the default view in general (So when starting in chase, ESC will cancel cockpit view, then bring up the quit menu).

Goober5000:
That was trickier than I thought, but done.

Baezon (approved):

PR #5314, opened 2023-04-05
"On Frame but actually Wait before you continue"-Executor

BMagnu:
What this does is basically providing an async.run that is able to run _and await_ an async with a built-in On Frame executor, but in a way that blocks _all other engine code_.
This makes it useful for things that need multiple frames, but the engine is in an unstable state where there simply are not multiple frames to be had until other things happen.
Two examples quickly become apparent for when this occurs:
1. Imagine you want to add animated transitions between game states. If you hook into a game state change, you simply never get a frame, so any attempt at starting an async will end up happening simultaneously to the stuff in the next game state. This new function will allow to start and run an async that yields and draws multiple frames _during_ the state change of FSO.
2. Say you have a script that does things that take a long time, like load many models from disk. You may be somewhere like on mission init, where you can't allow the game to proceed before you're done loading, but the loading takes so long, that windows starts flagging FSO as unresponsive, or you want to display a progress bar. This would allow you to add yields and draw calls for a progress bar interleaved into the loading code, without running the risk of FSO proceeding.

There is precedent in the engine for "halting" the state like this with a loop containing os_poll and gr_flip in stuff like how popups are handled, so this is not actually doing any new unknown witchcraft.

### DIFF
--- a/code/scripting/api/libs/async.cpp
+++ b/code/scripting/api/libs/async.cpp
@@ -218,7 +218,7 @@ ADE_FUNC(run,
 
 ADE_FUNC(awaitRunOnFrame,
 	l_Async,
-	"function() => any body",
+	"function() => any body, [boolean allowMultiProcessing = false]",
 	"Runs an asynchronous function in an OnFrameExecutor context and busy-waits for the coroutine to finish. "
 	"Inside this function you can use async.await to suspend the function until a promise resolves. "
 	"This is useful for cases where you need a scripting process to run over multiple frames, even when "
@@ -227,9 +227,10 @@ ADE_FUNC(awaitRunOnFrame,
 	"The result of the function body or nil if the function errored")
 {
 	luacpp::LuaFunction body;
+	bool allowMulti = false;
 	std::shared_ptr<executor::IExecutionContext> context;
 
-	if (!ade_get_args(L, "u", &body)) {
+	if (!ade_get_args(L, "u|b", &body, &allowMulti)) {
 		return ADE_RETURN_NIL;
 	}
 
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/async.cpp
```diff
@@ -214,6 +216,60 @@ ADE_FUNC(run,
...
+		game_set_frametime(-1);
+		game_do_state_common(gameseq_get_state(), 1);	// do stuff common to all states
+
+		gr_restore_screen(screen_id);
+		gr_flip();
+	}
+
+	gr_free_screen(screen_id);
+
+	if(promise.isErrored())
```
z64555, line 262: There is no check for when the promise gets invalid.

code/scripting/api/libs/async.cpp
```diff
@@ -214,6 +216,60 @@ ADE_FUNC(run,
...
+		LuaError(thread);
+		return true;
+	});
+
+	auto promise = runAsyncCoroutine(std::move(coroutine), executor::OnFrameExecutor, std::move(context));
+
+	int screen_id = gr_save_screen();
+
+	//Keep drawing pretend-frames until we have the promise resolved
+	while (promise.isValid() && !(promise.isResolved() || promise.isErrored())){
```
z64555, line 250: This loop doesn't handle multi packets, which may or may not be an issue later on.  So... we'll see if it does later on.

z64555 (changes requested):

code/scripting/api/libs/async.cpp
BMagnu: This was intentional, actually.
Mostly because we want to 100% avoid the game being able to switch states during this, as well as explicitly allowing for something like "wait _now_ to be done loading stuff before you're allowed to proceed", which also _must_ make multi wait.
It's possible to add a parameter to the function, but the full-blocking mode must be an option, at least.
Plus, a few popups do it this way, so it can't be that much of a dealbreaker

z64555: Well, I raise this point because it is probable the function will be (ab)used in a different way in the future by modders, so I want to see if there's anything we can do to prevent a crash or freeze.

z64555 (approved):

Goober5000 (approved):
Reviewed; no additional comments

PR #5317, opened 2023-04-06
Ets max speed fixes

Baezon:
Follow-up for #5292. A little more thorough on when changes to ETS values should change the max ship speed, including equalizing, multi packets and sexps.


wookieejedi (approved):
LGTM

Issue #5283, opened 2023-03-24
FR: Assign species-wide warp sfx/vfx in species_defs.tbl 


Kanangus:
Pretty straight forward, right now warp effects can only be assigned to individual ships instead of in the species defs table.

Goober5000:
Implemented in #5318

PR #5318, opened 2023-04-10
add per-species warp effects

Goober5000:
This allows warp parameters to be specified in species_defs.tbl, rather than requiring them to be specified for each ship in ships.tbl.  Since this introduces a cyclic dependency between species_defs.tbl and sounds.tbl, the parsing of sounds.tbl is broken into two stages.

Implements request from #5283.


wookieejedi:
Makes sense, though to reduce tedium of setting common things for each ship `Ship Templates` are highly useful (perhaps I should highlight those to the requester, Kanangus if they wish to have a faster way to set multiple aspects for many ships with existing methods.

Goober5000:
Well this particular feature is requested by Series Resurrecta.  They have different warp effects for Shivans, and it's just easier to set one species-wide warp effect than for each individual ship.

wookieejedi:
Right, makes sense. Templates are under-utilized IMO, so ideally they can be used in the team wants to easily set effects or values for a range of ships quickly in the future.

TRBlount:
Personally, I would prefer adding a post-processing function to resolve warp parameters in species_def as opposed to completely reordering the parsing for sounds and species_defs. Is there any particular reason you chose to reorder over post-processing?

Goober5000:
> Personally, I would prefer adding a post-processing function to resolve warp parameters in species_def as opposed to completely reordering the parsing for sounds and species_defs. Is there any particular reason you chose to reorder over post-processing?

That would be worse, I think.  The sounds.tbl has the advantage that it can be very easily split into two parts.  The first part can be parsed before species are initialized, and the second part can be parsed afterwards.  It's not ideal, but it's easy to understand and fairly painless.  The ideal would have been to separate the two sections when the table was originally designed, but c'est la vie.

Adding a post-processing function to species_def would be making the exact same design decision as this PR -- that is, dividing a parsing routine into before and after -- but in a much uglier way.  The code would have to skip all the warp fields during the first file read, and then afterwards iterate through all the species all over again, skipping all the fields *except* for the warp fields.  Then it would have to go back and back-fill the necessary parameters.  This would also violate the assumption that data structures are fully-formed and immutable after their parsing is finished: they would be in a half-valid state until the second pass for the warp fields.

In summary, splitting sounds.tbl involves the minimum amount of code and has the minimum impact.

TRBlount:
That's a fair point I suppose. Have the Series Resurrecta been given a test build and rigorously checked the functionality?

Goober5000:
Well, no, they wanted to avoid reworking their mod pack before the feature was merged.  But I ran several tests on a local test mod.  (That's how I discovered the circular dependency; warp sounds were not working before that was fixed.)

TRBlount (approved):
Barring any other objections, I'm going to tentatively merge this. Make sure to complete documentation of these features on the Wiki before 23.2 is released.

PR #5323, opened 2023-04-11
Add arbitrary curves support

### LINKED ISSUE #3045: Particle Effects FR: Ability to make particles accelerate or decelerate over time
This would take the form of an additive-velocity parameter, which could be a positive or negative float. Once this was established, it seems likely (though it's hard to say for certain) that it would be straightforward to add more such parameters, for ‘jerk’ and ‘snap’ control.

This would help tremendously in making certain types of effects, such as spark showers, look more natural. (Unrealistic, of course, but that’s a bit of a lost cause.)

Baezon:
Essentially, allows for modders to define arbitrary functions, using piece-wise components, made from polynomial, circular or previously defined functions. We can then apply those functions to map values in a wide-variety of domains and applications. For simplicity of the initial commit, just two applications are added here, particle +Size over lifetime and particle +Velocity scalar over lifetime.

```
$Name: test1
$KeyFrames:
(0, 1): Polynomial, 2, true
(1, 0): Constant

$Name: test2
$KeyFrames:
(0, 0): Circular
(0.5, 1): EaseInQuad    ; a built-in curve
(1, 0): Constant
```

Closes #3045

### DIFF
--- a/code/math/curve.cpp
+++ b/code/math/curve.cpp
@@ -1,0 +1,241 @@
+/*
+ * Copyright (C) Volition, Inc. 1999.  All rights reserved.
+ *
+ * All source code herein is the property of Volition, Inc. You may not sell 
+ * or otherwise commercially exploit the source or things you created based on the 
+ * source.
+ *
+*/ 
+
+#include "math/curve.h"
+#include "parse/parselo.h"
+
+SCP_vector<Curve> Curves;
+
+int curve_get_by_name(SCP_string in_name) {
+	for (int i = 0; i < (int)Curves.size(); i++) {
+		if (Curves[i].name == in_name)
+			return i;
+	}
+
+	return -1;
+}
+
+void parse_curve_table(const char* filename) {
+	try
+	{
+		read_file_text(filename, CF_TYPE_TABLES);
+		reset_parse();
+
+		optional_string("#Curves");
+
+		while (optional_string("$Name:")) {
+			SCP_string name;
+			stuff_string(name, F_NAME);
+
+			int index = curve_get_by_name(name);
+
+			if (index < 0) {
+				Curve curv = Curve(name);
+				curv.ParseData();
+				Curves.push_back(curv);
+			} else {
+				Curves[index].ParseData();
+			}
+		}
+	}
+	catch (const parse::ParseException& e)
+	{
+		mprintf(("TABLES: Unable to parse '%s'!  Error message = %s.\n", filename, e.what()));
+	}
+}
+
+void fill_default_curves() {
+	for (int i = 2; i < 6; i++) {
+		for (int rev = 0; rev < 2; rev++) {
+			SCP_string name("EaseIn");
+			switch (i) {
+			case 2: name += "Quad"; break;
+			case 3: name += "Cubic"; break;
+			case 4: name += "Quart"; break;
+			case 5: name += "Quint"; break;
+			}
+			if (rev)
+				name += "Rev";
+
+			Curves.emplace_back(name);
+			Curves.back().keyframes.push_back(curve_keyframe{ vec2d { 0.0f, rev ? 1.0f : 0.0f}, CurveInterpFunction::Polynomial, (float)i, 1.0f });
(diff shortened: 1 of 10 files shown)


code/math/curve.cpp
```diff
@@ -0,0 +1,241 @@
...
+ * source.
+ *
+*/ 
+
+#include "math/curve.h"
+#include "parse/parselo.h"
+
+SCP_vector<Curve> Curves;
+
+int curve_get_by_name(SCP_string in_name) {
```
Goober5000, line 15: The parameter should be `const &`

code/math/curve.cpp
```diff
@@ -0,0 +1,241 @@
...
+	for (int i = 0; i < (int)Curves.size(); i++) {
+		if (Curves[i].name == in_name)
```
Goober5000, line 17: Should this be case-insensitive?  If so you can use `SCP_string_lcase_equal_to()()`

code/math/curve.cpp
```diff
@@ -0,0 +1,241 @@
...
+	return -1;
+}
+
+void parse_curve_table(const char* filename) {
+	try
+	{
+		read_file_text(filename, CF_TYPE_TABLES);
+		reset_parse();
+
+		optional_string("#Curves");
```
Goober5000, line 30: Shouldn't this be `required_string`?

code/math/curve.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+			if (index < 0) {
+				Curve curv = Curve(name);
+				curv.ParseData();
+				Curves.push_back(curv);
+			} else {
+				Curves[index].ParseData();
+			}
+		}
+	}
```
Goober5000, line 46: Probably want a `required_string("#End")` here

code/math/curve.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+	if (cf_exists_full("curves.tbl", CF_TYPE_TABLES))
+		parse_curve_table("curves.tbl");
+
+	parse_modular_table(NOX("*-crv.tbm"), parse_curve_table);
+}
+
+Curve::Curve(SCP_string in_name)
+{
+	name = in_name;
```
Goober5000, line 115: The parameter should be `const &`, or alternatively the assignment should use `std::move`

code/math/curve.cpp
Baezon: Wow thats a fkin weird syntax, a struct with an overloaded call operator?? But w/e as long as it does what it do

code/math/curve.cpp
Baezon: Ehh, I'm not a fan of overly strict table syntax, unless there's a good reason to I wasn't planning on bothering.

code/math/curve.cpp
Goober5000: Yeah.  It's a functor, and it's the way to create a custom equality or hash function for a STL collection.

It could probably be made into a regular function on its own, now that I think of it.

code/math/curve.cpp
Goober5000: It's fine for now.  The main reason is if another section were ever added after it.  IIRC one of the existing tables didn't do that and it made things trickier.

Goober5000 (approved):

code/math/curve.h
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (C) Volition, Inc. 1999.  All rights reserved.
+ *
+ * All source code herein is the property of Volition, Inc. You may not sell 
+ * or otherwise commercially exploit the source or things you created based on the 
+ * source.
+ *
+*/
```
JohnAFernandez, line 8: Do we need to have this?  There's no volition source code in this file or curve.cpp? Also, SCP contributions are under the Unlicense.  I don't think I've seen anyone else add this to other new files.

Baezon: Idk I specifically looked at other added files and all the ones I could find had it, which is the only reason I did, but now I can't find any which do so I guess maybe not.

Goober5000: Some brand new files have a variation on it, for example objectdock.cpp or mod_table.cpp.  But that's more of a "monkey see, monkey do" than a rule.

PR #5325, opened 2023-04-12
convert hotkey types to enum

Goober5000:
Here's a suggested followup to #5249.  This converts the hotkey type to an enum for improved type safety.  It also adds the corresponding Lua enums and adjusts the Type virtvar to return them.

### DIFF
--- a/code/scripting/api/objs/missionhotkey.cpp
+++ b/code/scripting/api/objs/missionhotkey.cpp
@@ -59,23 +59,43 @@ ADE_VIRTVAR(Text, l_Hotkey, nullptr, "The text of this hotkey line", "string", "
 ADE_VIRTVAR(Type,
 	l_Hotkey,
 	nullptr,
-	"The type of this hotkey line. 0 for nothing, 1 for heading, 2 for wing, 3 for ship, 4 for ship in a wing",
-	"number",
+	"The type of this hotkey line: HOTKEY_LINE_NONE, HOTKEY_LINE_HEADING, HOTKEY_LINE_WING, HOTKEY_LINE_SHIP, or HOTKEY_LINE_SUBSHIP.",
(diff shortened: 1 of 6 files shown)


code/scripting/api/objs/missionhotkey.cpp
```diff
@@ -52,7 +54,25 @@ ADE_VIRTVAR(Type,
...
+			break;
+		case HotkeyLineType::WING:
+			eh_idx = LE_HOTKEY_LINE_WING;
+			break;
+		case HotkeyLineType::SHIP:
+			eh_idx = LE_HOTKEY_LINE_SHIP;
+			break;
+		case HotkeyLineType::SUBSHIP:
+			eh_idx = LE_HOTKEY_LINE_SUBSHIP;
+			break;
```
MjnMixael, line 72: Switch should have a default case per your own previous reviews just in case more hotkey enums are added in the future for whatever reason.

MjnMixael (changes requested):
These changes look overall workable, but I must say that I'm generally against this for such a simple case. It adds complexity to the Lua for minimum improvement in readability in the C++. I compared this against my script for this UI and I'll have to add code, not remove, to make it work again with these changes. Not to mention the documentation of Lua Enums is incredibly bad.

Goober5000:
> It adds complexity to the Lua for minimum improvement in readability in the C++.

Well the goals here are to improve type safety in the C++ and readability in the Lua, with readability in the C++ being less important.  Having scripting code that reads
```
key.Type == HOTKEY_LINE_HEADING
```
is more meaningful than
```
key.Type == 1
```
and using the enum token names also guards against any possibility of the numbers changing in the future.

The documentation of Lua Enums can be improved, if you can point out areas where clarification or elaboration would help.

Issue #5255, opened 2023-03-14
Handling of absent tables


EatThePath:
Context: from time to time I work on a minimal total conversion template. At some point in 23.0's dev cycle it was in a launchable state, but subsequent changes to the asteroids and ranks table handling have made it unlaunchable because of missing data. possibly more, since I haven't resolved those cases yet.

In some cases this extra validation is probably the best way to go. When practical, I think it'd be preferable to let a TC omit tables controlling systems they don't intend to use. Unfamiliarity with these systems means I don't have any idea which bin these two cases fall in.

MjnMixael:
Leaving ranks and asteroid data uninitialized is a bad idea. That's what was happening before. The data was just undefined and there were no checks in place for that, especially for ranks.

The only solution to allowing you to omit these tables would be to add a built-in table with default data. That gets more complicated with asteroids since you need the actual asteroid model and texture data as well. It'd be easier for you to just create tiny placeholder tables. You can get away with just a single rank, now. But asteroids is a whole different story. Asteroid.cpp is built around having at least 4 asteroids and 1 debris defined.

PR #5329, opened 2023-04-12
Allow 0-Asteroid setups

BMagnu:
Implements part of #5255.
This allows for an asteroid table to be functionally empty, not requiring any asteroids to be set up.
Evidently, this will disallow the usage of asteroid fields, but some TC's may not need any, so allow these to set up without asteroids.

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -2454,6 +2454,10 @@ void asteroid_init()
 	// parse any modular tables
 	parse_modular_table("*-ast.tbm", asteroid_parse_tbl);
 
+	//No asteroids defined. Bail!
+	if (asteroid_list.empty())
+		return;
+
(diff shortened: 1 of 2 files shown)


Baezon:
I'm surprised this is all it takes! What exactly does it say in FRED if you try to add one? What if you use `set-asteroid-field`?

MjnMixael:
I think FRED will just complain that you have to select an asteroid if you try to create a field and click OK. The asteroid lists will be empty which is allowable in MFC.

For the sexps (set-asteroid-field and set-debris-field), it should be checked that the list builder doesn't crash FRED. I can't recall if those get 'none' appended or not.

MjnMixael (changes requested):
I checked. The debris sexp gets 'none' applied so it should be fine and the asteroid sexp doesn't list the 'roids'.

MjnMixael (approved):

PR #5330, opened 2023-04-12
Add beam particles

### LINKED ISSUE #3050: Particle Effects/Beam Weapons FR: Ability to spawn particles in a tube along the length of a beam
I would like to be able to spawn particle effects in a tube, of adjustable radius, along the length of an active beam. This would allow for various new visual effects -- especially vapor that lingers after the beam dies.

Because this specifies a region in which the effect would spawn, it would need some sort of "density" parameter -- the probability that a given point within the tube will spawn the particle effect each frame.

For cone-type effects, the 'incoming' vector should be the direction in which the beam is being fired. (Direction control is necessary for effective use of laser-type particles, which are very relevant for this feature.)

Baezon:
Adds a new `SourceOriginType` for particle sources corresponding to the entire length of a beam. All the special beam handling definitely... *chafes* a little with the framework, but I'm confident that if beams are to be done in the framework this is the best way to do it. Key here is that when asked for the origin's position, it gives random positions along its length, allowing all the existing shape and volume effects to otherwise work the same. 

The random distribution along its length is also biased towards the start of the beam, linearly decaying to 0 at its maximum range. This is done so that the distribution of particles essentially matches the opacity of the beam graphic itself, which also linearly decays to 0 towards its end.

Closes #3050

### DIFF
--- a/code/particle/ParticleSource.h
+++ b/code/particle/ParticleSource.h
@@ -67,7 +67,7 @@ class SourceOrigin {
 	/**
 	 * @brief Gets the current, global position of the origin
 	 * 
-	 * Be aware for beam soruces this will give *random* points along its length
+	 * Be aware for beam sources this will give *random* points along its length
 	 * 
 	 * @param posOut The pointer where the location will be stored
 	 */
--- a/code/particle/ParticleSource.h
+++ b/code/particle/ParticleSource.h
@@ -25,7 +25,7 @@ namespace particle {
 enum class SourceOriginType {
 	NONE, //!< Invalid origin
 	VECTOR, //!< World-space offset
-	BEAM,
+	BEAM, //!< A beam
 	OBJECT, //!< An object
 	PARTICLE //!< A particle
 };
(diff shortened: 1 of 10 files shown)


code/particle/ParticleSource.h
```diff
@@ -65,6 +66,9 @@ class SourceOrigin {
 
 	/**
 	 * @brief Gets the current, global position of the origin
+	 * 
+	 * Be aware for beam soruces this will give *random* points along its length
```
Goober5000, line 70: "sources"

EatThePath:
Would it be much extra work to add particles to warmup and warmdown states? I would have uses for it.

Baezon:
Handling warmup and warmdown correctly is something I've already given some thought to, and definitely something I'd want too, but I decided to leave it off this commit, because I believe it will require some special treatment.

X3N0-Life-Form:
Nice to have beam particles to go with our particle beams.

Goober5000 (approved):

PR #5333, opened 2023-04-14
Flush empty cbrief voice filenames to "None"

Baezon:
Just like briefings and debriefs do, this way the user can just delete the field for the same effect.


Goober5000 (approved):
Looks good.  I'd suggest doing the same thing with `ani_filename` and `"<default>"`.

PR #5335, opened 2023-04-14
Randomize particle rotation

Baezon:
Still leaves room for more in-depth options later, but for now just a global flag. Also removes `particle_index` dunno wtf 'dynamic array usage' is about but it was used before for a janky random 90 degree rotation (which was also broken a few years ago)

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -139,6 +139,7 @@ bool Auto_assign_personas;
 bool Countermeasures_use_capacity;
 bool Play_thruster_sounds_for_player;
 std::array<std::tuple<float, float>, 6> Fred_spacemouse_nonlinearity;
+bool Randomize_particle_rotation;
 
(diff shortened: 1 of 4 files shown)


wookieejedi (approved):
Tests work well and overall LGTM

code/mod_table/mod_table.cpp
```diff
@@ -139,6 +139,7 @@ bool Auto_assign_personas;
 bool Countermeasures_use_capacity;
 bool Play_thruster_sounds_for_player;
 std::array<std::tuple<float, float>, 6> Fred_spacemouse_nonlinearity;
+bool Randomize_particle_rotation;
```
Goober5000, line 142: `Randomize_particle_rotation` should be initialized in `mod_table_reset`

Goober5000 (changes requested):

Goober5000 (approved):

PR #5336, opened 2023-04-14
tweak rank handling in Lua

Goober5000:
Follow-up to #5280:

1. Make the Rank virtvar actually return a rank object, to be consistent with other Lua API methods
2. Add a RankIndex virtvar to replicate the previous functionality
3. Make RankIndex return the expected 1-based index for Lua

This is a breaking change to the API, but since the original API is less than a month old, the impact should be small.

### DIFF
--- a/code/scripting/api/objs/player.cpp
+++ b/code/scripting/api/objs/player.cpp
@@ -11,6 +11,7 @@
 #include "pilotfile/pilotfile.h"
 #include "playerman/player.h"
 #include "stats/medals.h"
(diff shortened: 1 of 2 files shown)


MjnMixael (changes requested):
If you're going to change my API again, can you at least do it right? The rank index should be a virtvar of the rank object.

Goober5000:
Can do.  I've pushed an update.

MjnMixael (approved):

PR #5338, opened 2023-04-15
Add an Ai goal flag to afterburn hard

### LINKED ISSUE #5303: FR: Force afterburn on AI orders
As per discussion with Asteroth, this is a request for two additional arguments to AI orders.

The first arg is true/false bool for whether or not the AI uses afterburners to close distance to a target during an order, such as waypoints or destroy target.

The second is a defined radius around the target in which the AI should not use afterburners while executing the order.

This is a balance critical active mod request.

Baezon:
Allows modders to more or less force AI to afterburn as hard as possible, either by simply adding or short-cutting existing afterburning behavior. Not as a matter of difficulty or anything, but mission choreography. They still should afterburn intelligently though, this only makes them hustle to get near their goal, and they won't burn so hard they can't properly complete their goal, whether it's attacking or defending or whatever, their behavior once they get here is basically unchanged.

~~Additionally, this removes some superfluous checks for having afterburners or whether they are locked, `afterburner_start()` already checks both of these so they just muck up the code.~~ Since it makes different decisions based on the status of the afterburners they are not superfluous >.>

Closes #5303

### DIFF
--- a/code/ai/aibig.cpp
+++ b/code/ai/aibig.cpp
@@ -628,15 +628,16 @@ bool ai_new_maybe_reposition_attack_subsys() {
 	// full speed until you're close, then slow down a bit
 	accelerate_ship(aip, subsys_distance > 200.0f ? 1.0f : subsys_distance / 200.0f);
 
-	if (ai_willing_to_afterburn_hard(aip) && subsys_distance > 1000.0f)
-		ai_afterburn_hard(Pl_objp, aip);
-	else if (subsys_distance > 300.0f) {
-		if (ai_maybe_fire_afterburner(Pl_objp, aip)) {
-			afterburners_start(Pl_objp);
-			aip->afterburner_stop_time = Missiontime + 2 * F1_0;
+	if (!(Ships[Pl_objp->instance].flags[Ship::Ship_Flags::Afterburner_locked])) {
+		if (ai_willing_to_afterburn_hard(aip) && subsys_distance > 1000.0f)
+			ai_afterburn_hard(Pl_objp, aip);
+		else if (subsys_distance > 300.0f) {
+			if (ai_maybe_fire_afterburner(Pl_objp, aip)) {
+				afterburners_start(Pl_objp);
+				aip->afterburner_stop_time = Missiontime + 2 * F1_0;
+			}
 		}
-	}
-	else
+	} else
 		afterburners_stop(Pl_objp);
 
 	return true;
--- a/code/ai/aibig.cpp
+++ b/code/ai/aibig.cpp
@@ -850,7 +850,7 @@ void ai_big_chase_attack(ai_info *aip, ship_info *sip, vec3d *enemy_pos, float d
 			} else {
 				accelerate_ship(aip, accel);
 
-				if (ai_willing_to_afterburn_hard(aip) && accel > 0.99f) {
+				if (ai_willing_to_afterburn_hard(aip) && accel > 0.99f && !(shipp->flags[Ship::Ship_Flags::Afterburner_locked])) {
 					ai_afterburn_hard(Pl_objp, aip);
 				} else if ((aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use]) && !(shipp->flags[Ship::Ship_Flags::Afterburner_locked]) && (accel > 0.95f)) {
 					if (ai_maybe_fire_afterburner(Pl_objp, aip)) {
--- a/code/ai/aibig.cpp
+++ b/code/ai/aibig.cpp
@@ -628,16 +628,17 @@ bool ai_new_maybe_reposition_attack_subsys() {
 	// full speed until you're close, then slow down a bit
 	accelerate_ship(aip, subsys_distance > 200.0f ? 1.0f : subsys_distance / 200.0f);
 
-	if (!(Ships[Pl_objp->instance].flags[Ship::Ship_Flags::Afterburner_locked])) {
-		if (subsys_distance > 300.0f) {
-			if (ai_maybe_fire_afterburner(Pl_objp, aip)) {
-				afterburners_start(Pl_objp);
-				aip->afterburner_stop_time = Missiontime + 2 * F1_0;
-			}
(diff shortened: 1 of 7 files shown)


code/ai/aicode.cpp
```diff
@@ -12529,7 +12586,7 @@ int ai_formation()
 
 	ship_info *sip = &Ship_info[shipp->ship_info_index];
 	bool ab_allowed = false;
-	if ((sip->flags[Ship::Info_Flags::Afterburner]) && !(shipp->flags[Ship::Ship_Flags::Afterburner_locked]) && (aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use])) {
+	if ((aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use]) || ai_willing_to_afterburn_hard(aip)) {
```
BMagnu, line 12589: This removes the check for locked afterburners.
Is this intentional?

code/ai/aicode.cpp
```diff
@@ -10734,8 +10784,11 @@ void ai_guard()
 			compute_desired_rvec(&rvec, &goal_point, &Pl_objp->pos);
 			ai_turn_towards_vector(&goal_point, Pl_objp, nullptr, nullptr, 0.0f, 0, &rvec);
 
-			if ((aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use]) && !(shipp->flags[Ship::Ship_Flags::Afterburner_locked]) && (accel_scale * (0.25f + dist_to_goal_point/700.0f) > 0.8f)) {
-				if (ai_maybe_fire_afterburner(Pl_objp, aip)) {
```
BMagnu, line 10738: here as well

code/ai/aicode.cpp
```diff
@@ -10611,16 +10655,22 @@ void ai_big_guard()
 		ai_turn_towards_vector(&goal_pt, Pl_objp, nullptr, nullptr, 0.0f, 0);
 		accelerate_ship(aip, 1.0f);
 
-		if ((aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use]) && !(shipp->flags[Ship::Ship_Flags::Afterburner_locked]) && (cur_guard_rad > 1.1f * max_guard_dist)) {
+
+		bool free_ab_use = aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use];
+		if ((free_ab_use || ai_willing_to_afterburn_hard(aip)) && (cur_guard_rad > 1.1f * max_guard_dist)) {
```
BMagnu, line 10660: here as well

code/ai/aicode.cpp
```diff
@@ -4685,7 +4715,9 @@ void ai_fly_to_target_position(vec3d* target_pos, bool* pl_done_p=NULL, bool* pl
 			ab_allowed = false;
 	}
 
-	if (!(sip->flags[Ship::Info_Flags::Afterburner]) || (shipp->flags[Ship::Ship_Flags::Afterburner_locked]) || !(aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use])) {
+	if (!(aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || 
+		aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use] || 
+		ai_willing_to_afterburn_hard(aip))) {
```
BMagnu, line 4720: here as well

code/ai/aibig.cpp
```diff
@@ -849,7 +850,9 @@ void ai_big_chase_attack(ai_info *aip, ship_info *sip, vec3d *enemy_pos, float d
 			} else {
 				accelerate_ship(aip, accel);
 
-				if ((aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use]) && !(shipp->flags[Ship::Ship_Flags::Afterburner_locked]) && (accel > 0.95f)) {
+				if (ai_willing_to_afterburn_hard(aip) && accel > 0.99f) {
+					ai_afterburn_hard(Pl_objp, aip);
+				} else if ((aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use]) && !(shipp->flags[Ship::Ship_Flags::Afterburner_locked]) && (accel > 0.95f)) {
```
BMagnu, line 855: But here interestingly not.

code/ai/aicode.cpp
BMagnu: Just read the comment, duh.

code/ai/aibig.cpp
Baezon: Actually... on second thought, these often aren't superfluous...

It makes *different decisions* based on whether the afterburners are locked or not. It's trying and failing to afterburn vs not even attempting to afterburn and doing something else instead... so I will have to amend these regardless.

code/ai/aicode.cpp
```diff
@@ -12529,7 +12594,7 @@ int ai_formation()
 
 	ship_info *sip = &Ship_info[shipp->ship_info_index];
 	bool ab_allowed = false;
-	if ((sip->flags[Ship::Info_Flags::Afterburner]) && !(shipp->flags[Ship::Ship_Flags::Afterburner_locked]) && (aip->ai_flags[AI::AI_Flags::Free_afterburner_use] || aip->ai_profile_flags[AI::Profile_Flags::Free_afterburner_use])) {
```
BMagnu, line 12532: Well, if we want to keep the AB lock checks, it should probably also return here

BMagnu (approved):

Goober5000 (changes requested):
One discrepancy; the rest looks ok

Goober5000 (approved):

PR #5340, opened 2023-04-16
Add debris density

Baezon:
Just like ship density, allows for a ship class specific multiplier for debris mass and MOI, including subsystem debris-specific density.

Address the last point of #5278

### DIFF
--- a/code/debris/debris.cpp
+++ b/code/debris/debris.cpp
@@ -374,6 +374,7 @@ MONITOR(NumHullDebris)
  * @param exp_center	Explosion center in global coordinates
  * @param hull_flag		Whether this debris is a chunk of a ship's hull (as opposed to shrapnel)
  * @param exp_force		Explosion force, used to assign velocity to pieces. 1.0f assigns velocity like before. 2.0f assigns twice as much to non-inherited part of velocity
+ * @param source_subsys	The subsystem this debris came from, if any
  */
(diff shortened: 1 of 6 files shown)


wookieejedi (approved):
Nice and straightforward

code/debris/debris.cpp
```diff
@@ -375,7 +375,7 @@ MONITOR(NumHullDebris)
  * @param hull_flag		Whether this debris is a chunk of a ship's hull (as opposed to shrapnel)
  * @param exp_force		Explosion force, used to assign velocity to pieces. 1.0f assigns velocity like before. 2.0f assigns twice as much to non-inherited part of velocity
  */
-object *debris_create(object *source_obj, int model_num, int submodel_num, vec3d *pos, vec3d *exp_center, bool hull_flag, float exp_force)
+object *debris_create(object *source_obj, int model_num, int submodel_num, vec3d *pos, vec3d *exp_center, bool hull_flag, float exp_force, ship_subsys* source_subsys)
```
JohnAFernandez, line 378: Could you add this new parameter to the doxy documentation above it?

JohnAFernandez (changes requested):
Just one thing

code/debris/debris.cpp
Baezon: whoops, ty

JohnAFernandez (approved):

PR #5341, opened 2023-04-16
Make restriction which ships accept Lua player orders configurable

BMagnu:
This is the pendant to +Target Restriction: which can disable certain Lua AI orders for invalid targets, in that it disables giving Lua AI orders to certain ships.
Since this _should_ be done using objecttypes, this is more for context-sensitive AI orders that need finer control (such as orders that are only valid for the player's wingmates, or orders like formation flying which only make sense when given to a full wing and not just one ship)


wookieejedi (approved):
Tests work successfully and overall looks good to me

PR #5343, opened 2023-04-16
synchronize simulation timestamps across the entire frame

Goober5000:
Prior to the timer upgrade, all simulation timestamps shared identical timer and timestamp values for the duration of the frame, since they were only updated within the `game_set_frametime` function.  Since the timer upgrade depends on the system clock, timestamps evaluated near the end of the frame will be very slightly later than timestamps evaluated near the beginning of the frame.  It is desirable to keep them all in sync.

This PR tweaks the timestamp system to take a snapshot at the beginning of the frame, then use that snapshot for the entire duration of the frame.  This only affects simulation timestamps; real-time timestamps still use the live timer.

This is a redo of PR #5266, with some small additional tweaks.

Update: this now fixes the bug discovered in #5287.

### DIFF
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -634,8 +634,8 @@ void timestamp_update_time_compression()
 	}
 
 	// grab the independent variable of the equation before we change anything
-	// (we need to take a new snapshot to get the accurate value, but this is ok
-	// since time compression is only updated at the start of the frame)
+	// (we need to get the live value to be accurate, which takes a new snapshot,
+	// but this is ok since time compression is only updated at the start of the frame)
 	auto timestamp_raw = timestamp_get_raw(true);
 
 	// we need to move the counter offset to make the raw timestamp zero (so that it can start ticking with a new multiplier)
@@ -649,6 +649,9 @@ void timestamp_update_time_compression()
 	// now we can set the new info
 	Timestamp_current_time_compression = Game_time_compression;
 	Timestamp_time_compression_multiplier = static_cast<float>(Game_time_compression) / F1_0;
+
+	// and now take a new snapshot so that the raw timestamp is correct for this frame
+	timestamp_get_raw(true);
 }
 
 // ======================================== mission-specific stuff ========================================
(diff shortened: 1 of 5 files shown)


Goober5000:
Theoretically this should be susceptible to the same bug that #5287 discussed, but I have not been able to reproduce it.  Neither time compression nor `runAsync` show any problems.

JohnAFernandez:
I highly recommend sending this to other testers who were having issues with the last version of this patch, because I honestly do not see a difference between the last attempt and this one.

wookieejedi:
Fortunately, was able to set Goober up with a reproducible bug case, though in any case additional testing with others such as Durandal (who was testing before) will be needed too as the process progresses.

Goober5000:
> I honestly do not see a difference between the last attempt and this one.

The first commit is exactly the same as before, and this PR adds two more polishing.  This is a starting point; I am using it as a basis for testing.  I'll put it in draft to be clear.

Goober5000:
Ok, I found the cause of the bug and fixed it.  When time compression changes, the elapsed time must be consistent, even when the rate of time changes.  To keep this simple, the y-intercept of the current timeline is set to 0, and the new timeline starts from that point.  The introduction of timestamp snapshots caused inaccuracies because the calculations would be based on not-quite-instant values.  Normally it would be hard to notice this, but FotG had some scripts that made some very dramatic time compression changes in very quick succession.  We have to thank FotG here because, as hard as this was to track down, it would have been much worse without their script.

The solution is to always make sure the timer re-takes the snapshot whenever time compression changes.  This itself would make the snapshot inconsistent for the frame where the time compression change takes place, except for the fortuitous discovery that most time compression changes *also* take place at the start of the frame.  So this works out very nicely.  It was a relatively simple matter to change "most" to "all", so all methods of changing time compression now occur at precisely the same spot.

wookieejedi:
Tested and works for me. The code update makes sense too, as it follows the similar logic to updating the booleans for `timestamp_get_raw` while also adding the dually important fix within `change_time_compression`. I'll send to Durandal for him to test and update with a review/approval if the build works for him, too.

wookieejedi (approved):
Additional tests from Durandal and myself and everything is working as expected. Overall code looks good too, as it follows the similar logic to updating the booleans for `timestamp_get_raw` while also adding the dually important fix within `change_time_compression`

PR #5344, opened 2023-04-17
Fix glowpoints on animated submodels

Baezon:
Since #2983, this does not depend on being a ship, and will allow glow points to follow animated submodels on models which arent ships.


wookieejedi (approved):

Goober5000:
Oh, neat!

PR #5345, opened 2023-04-17
Finalize scpui api

MjnMixael:
The bulk of this PR changes the API descriptions to remove the warnings that it's experimental or testing. As of this commit, I consider the API finalized and ready for general use.

One small addition is that this PR also changes three debrief API methods from using a global to using a parameter to match the rest of the API. Should be no change in functionality.

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -78,7 +78,7 @@ namespace scripting {
 namespace api {
 
 //*************************Global UI stuff*************************
-ADE_LIB(l_UserInterface, "UserInterface", "ui", "Functions for managing the \"scpui\" user interface system.");
+ADE_LIB(l_UserInterface, "UserInterface", "ui", "Functions for managing the \"SCPUI\" user interface system.");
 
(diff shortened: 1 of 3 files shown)


code/scripting/api/libs/ui.cpp
```diff
@@ -469,8 +465,7 @@ ADE_FUNC(acceptPilot, l_UserInterface_Barracks, "player selection", "Accept the
 ADE_LIB_DERIV(l_UserInterface_Options,
 	"OptionsMenu",
 	nullptr,
-	"API for accessing values specific to the options screen.<br><b>Warning:</b> This is an internal "
-	"API for the new UI system. This should not be used by other code and may be removed in the future!",
+	"API for accessing values specific to the options screen.",
```
wookieejedi, line 468: Are `Screen` and `UI` the same in these comments? Some say `options screens` while others say `options UI`. A few screen/UI names are also capitalized? Sorry if this is very minor, just wanted to double check now rather than later.

MjnMixael: Eh, might as well be consistent. Done.

wookieejedi (approved):

PR #5348, opened 2023-04-17
Deprecate old versions in favor of versions that use color objects

MjnMixael:
I'm not sure if this is the best way to do this or even worth the possible annoyance for scripts. However, since the Lua API has full color support I was curious what it would take to convert past methods to get and set using the proper color object rather than raw rgb(sometimes a) values. Turns out there were only a handful of places.

Take it or leave it. Doesn't change underlying functionality, just unifies color handling across the API and deprecates the old methods.

### DIFF
--- a/code/scripting/api/libs/graphics.cpp
+++ b/code/scripting/api/libs/graphics.cpp
@@ -552,19 +552,19 @@ ADE_FUNC(setColor,
 ADE_FUNC(getColor,
 	l_Graphics,
 	"boolean",
-	"Gets the active 2D drawing color. True to return raw rgb, false to return a color object. Defaults to true.",
+	"Gets the active 2D drawing color. False to return raw rgb, true to return a color object. Defaults to false.",
 	"number, number, number, number | color",
 	"rgba color which is currently in use for 2D drawing")
 {
 	if(!Gr_inited)
 		return ADE_RETURN_NIL;
 
-	bool rc = true;
+	bool rc = false;
 	ade_get_args(L, "|b", &rc);
 
 	color cur = gr_screen.current_color;
 
-	if (rc) {
+	if (!rc) {
 		return ade_set_args(L, "iiii", (int)cur.red, (int)cur.green, (int)cur.blue, (int)cur.alpha);
 	} else {
 		return ade_set_args(L, "o", l_Color.Set(cur));
--- a/code/scripting/api/libs/graphics.cpp
+++ b/code/scripting/api/libs/graphics.cpp
@@ -284,13 +284,30 @@ ADE_FUNC(clear, l_Graphics, nullptr, "Calls gr_clear(), which fills the entire s
 	return ADE_RETURN_NIL;
 }
 
-ADE_FUNC(clearScreen, l_Graphics, "[number red, number green, number blue, number alpha]",
-         "Clears the screen to black, or the color specified.", nullptr, nullptr)
+ADE_FUNC(clearScreen, l_Graphics, "[number red, number green, number blue, number alpha] | [color Color]",
+	"Clears the screen to black, or the color specified.",
+	nullptr,
+	nullptr)
 {
-	int r,g,b,a;
-	r=g=b=0;
-	a=255;
-	ade_get_args(L, "|iiii", &r, &g, &b, &a);
+	int r, g, b, a;
+	r = g = b = 0;
+	a = 255;
+
+	if (lua_isnumber(L, 1)) {
+		ade_get_args(L, "|iiii", &r, &g, &b, &a);
+
+	} else {
+		color col;
+		gr_init_alphacolor(&col, 0, 0, 0, 255);
+
+		ade_get_args(L, "|o", l_Color.Get(&col));
+		r = col.red;
+		g = col.green;
+		b = col.blue;
+		a = col.alpha;
+	}
+
+	
 
 	//WMC - Set to valid values
 	if(r != 0 || g != 0 || b != 0 || a!= 255)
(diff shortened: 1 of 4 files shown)


Goober5000 (changes requested):
This is a great idea.  But rather than deprecating the old functions, we should keep them around so that modders have the flexibility of using either approach.

code/scripting/api/libs/hud.cpp
```diff
@@ -178,6 +250,23 @@ ADE_FUNC(getHUDGaugeColorInMission,
...
+	"Red, green, blue, and alpha of the gauge")
+{
+	int idx = getDefaultGaugeIndex(L);
+
+	if ((idx < 0) || (idx >= (int)default_hud_gauges.size()))
+		return ADE_RETURN_NIL;
+
+	color c = default_hud_gauges[idx]->getColor();
+
+	return ade_set_args(L, "iiii", l_Color.Set(c));
```
Goober5000, line 267: This should have a type of "o", not "iiii".

Goober5000 (changes requested):
Also noticed this

BMagnu:
I am undecided on this.
On the one hand we have precedent in deprecating stuff like drawMonochromeBitmap.
On the other hand, these color functions are _much_ more prevalent to the point where basically _every_ script that renders stuff currently in existence would be incompatible with mods targeting 23.2.
Thus, while I like adding options to handle color objects, but I would also lean against deprecating the old functions. Ideally, I'd like the functions to also not be two different functions, but one function that determines whether it has a color object or an RGBA number tuple from the type of its (first) argument. That'd allow the same functionality without cluttering the API

MjnMixael:
> 

I had thought about this. I couldn't come up with a way for a single method to do both without requiring a bunch of nil arguments along the way. For example, say you have `gr.clearScreen()` and you want to pass a color to it, but we also don't want to break existing usage of it. So we add [color] as an optional parameter after rgba. In order to call it with a color you have to send `gr.clearScreen(nil, nil, nil, nil, color)` which also isn't super ideal. I'm open to ideas, though, because I have similar thoughts to you. These are some widely used functions.

Goober5000:
That's an even better idea.  You could check the first color parameter using `lua_isnumber`.  If it's true, get three (or four) numbers; if not, get one object.  See `getDefaultGaugeIndex` for a number/string approach, or `mn.startMission` for a string/enumeration approach.

code/scripting/api/libs/graphics.cpp
```diff
@@ -502,31 +519,56 @@ ADE_FUNC(setCamera, l_Graphics, "[camera Camera]", "Sets current camera, or rese
...
+	gr_set_color_fast(&col);
 
 	return ADE_RETURN_NIL;
 }
 
-ADE_FUNC(getColor, l_Graphics, nullptr, "Gets the active 2D drawing color", "number, number, number, number" , "rgba color which is currently in use for 2D drawing")
+ADE_FUNC(getColor,
+	l_Graphics,
+	"boolean",
+	"Gets the active 2D drawing color. True to return raw rgb, false to return a color object. Defaults to true.",
```
Goober5000, line 555: I'd prefer true to return a color object, and false to return raw rgb, defaulting to false -- with the rationale that setting a flag to true changes the default behavior -- but I won't make that a condition of the review.

code/scripting/api/libs/hud.cpp
```diff
@@ -103,17 +104,33 @@ ADE_FUNC(getHUDConfigShowStatus, l_HUD, "number|string gaugeNameOrIndex", "Gets
...
+         "number|string gaugeNameOrIndex, [number red, number green, number blue, number alpha] | [color Color]",
+	"Modifies color used to draw the gauge in the pilot config",
+	"boolean",
+	"If the operation was successful")
 {
 	int idx = getDefaultGaugeIndex(L);
 	int r = 0;
 	int g = 0;
 	int b = 0;
 	int a = 0;
```
Goober5000, line 116: I know this isn't your change, but I'm assuming `a` should default to 255 in the event that the script supplies numbers but doesn't supply all four.

code/scripting/api/libs/hud.cpp
```diff
@@ -125,33 +142,57 @@ ADE_FUNC(setHUDGaugeColor, l_HUD,
 
 ADE_FUNC(getHUDGaugeColor,
 	l_HUD,
-	"number|string gaugeNameOrIndex",
-	"Color specified in the config to draw the gauge",
+	"number|string gaugeNameOrIndex, [boolean ReturnType]",
+	"Color specified in the config to draw the gauge. True to return raw rgba, false to return color object. Defaults to true.",
```
Goober5000, line 146: Same comment as `gr.getColor` about true vs. false.

code/scripting/api/libs/hud.cpp
```diff
@@ -125,33 +142,57 @@ ADE_FUNC(setHUDGaugeColor, l_HUD,
...
 	"number, number, number, number",
 	"Red, green, blue, and alpha of the gauge")
 {
 	int idx = getDefaultGaugeIndex(L);
 
+	bool rc = true;
+	ade_get_args(L, "|b", &rc);
```
Goober5000, line 153: This function should be tested.  The `getDefaultGaugeIndex(L)` function already gets one argument, and the `ade_get_args` function does not pick up at the next spot in the parameter list, as one might expect.  I believe calling `ade_get_args(L, "*|b", &rc);` instead will fix it.

code/scripting/api/libs/hud.cpp
```diff
@@ -163,19 +204,26 @@ ADE_FUNC(setHUDGaugeColorInMission, l_HUD,
 
 ADE_FUNC(getHUDGaugeColorInMission,
 	l_HUD,
-	"number|string gaugeNameOrIndex",
+	"number|string gaugeNameOrIndex, [boolean ReturnType]",
```
Goober5000, line 207: Same comment as `gr.getColor` about true vs. false.  Also the documentation should indicate what true means and what false means.

code/scripting/api/objs/object.cpp
```diff
@@ -661,27 +662,28 @@ ADE_FUNC(removeSound, l_Object, "soundentry GameSnd, [subsystem Subsys=nil]",
 }
 
 
-ADE_FUNC(getIFFColor, l_Object, "number, number, number", 
-	"Gets the IFF color of the object",
-	"number, number, number", 
+ADE_FUNC(getIFFColor, l_Object, "boolean ReturnType", 
+	"Gets the IFF color of the object. True to return raw rgb, false to return color object. Defaults to true.",
```
Goober5000, line 666: Same comment as `gr.getColor` about true vs. false.

code/scripting/api/objs/object.cpp
```diff
@@ -661,27 +662,28 @@ ADE_FUNC(removeSound, l_Object, "soundentry GameSnd, [subsystem Subsys=nil]",
...
-	color* col = hud_get_iff_color(objp);
-
-	int r = col->red;
-	int g = col->green;
-	int b = col->blue;
+	color* cur = hud_get_iff_color(objp);
 
-	return ade_set_args(L, "iii", r, g, b);
+	if (rc) {
+		return ade_set_args(L, "iii", (int)cur->red, (int)cur->green, (int)cur->blue, (int)cur->alpha);
```
Goober5000, line 683: The number of characters in the format string does not match the number of variables returned.  Instead of truncating it to three you could bump it "officially" to four... this will not break old scripts that only expect three values because the fourth value will be silently dropped.

code/scripting/api/objs/team.cpp
```diff
@@ -41,26 +42,27 @@ ADE_VIRTVAR(Name, l_Team, "string", "Team name", "string", "Team name, or empty
 
 ADE_FUNC(getColor,
 	l_Team,
-	nullptr,
-	"Gets the IFF color of the specified Team",
+	"boolean ReturnType",
+	"Gets the IFF color of the specified Team. True to return raw rgb, false to return color object. Defaults to true.",
```
Goober5000, line 46: Same comment as `gr.getColor` about true vs. false.

code/scripting/api/objs/team.cpp
```diff
@@ -41,26 +42,27 @@ ADE_VIRTVAR(Name, l_Team, "string", "Team name", "string", "Team name, or empty
...
-	color* col = iff_get_color_by_team(idx, 0, 0);
-
-	r = col->red;
-	g = col->green;
-	b = col->blue;
+	color* cur = iff_get_color_by_team(idx, 0, 0);
 
-	return ade_set_args(L, "iii", r, g, b);
+	if (rc) {
+		return ade_set_args(L, "iii", (int)cur->red, (int)cur->green, (int)cur->blue, (int)cur->alpha);
```
Goober5000, line 61: As with `getIFFColor`, the number of characters does not match the number of variables, but this could also officially be bumped to four.

Goober5000 (commented):
Looks overall pretty good.  Some comments.

code/scripting/api/libs/hud.cpp
MjnMixael: Changes tested and working.

code/scripting/api/objs/object.cpp
MjnMixael: Changed

code/scripting/api/objs/team.cpp
MjnMixael: Changed

Goober5000 (approved):
Cool.  Approved.

PR #5349, opened 2023-04-17
Cleanup/batching int3

EatThePath:
A followup to #5154, I didn't want to leave this error handling in an inconsistent state state but also didn't want to mix the changes there with a broader pass.

### DIFF
--- a/code/render/batching.cpp
+++ b/code/render/batching.cpp
@@ -672,8 +672,8 @@ void batching_add_laser_internal(primitive_batch *batch, int texture, vec3d *p0,
 
 void batching_add_bitmap(int texture, vertex *pnt, int orient, float rad, float alpha, float depth)
 {
-	if (texture < 0) {
-		Int3();
+	Assertion((texture >= 0), "batching_add_...() attempted for invalid texture");
+	if ( texture < 0 ) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


code/render/batching.cpp
```diff
@@ -884,6 +894,16 @@ void batching_add_tri(int texture, vertex *verts)
...
 
+void batching_add_tri(int texture, vertex *verts, primitive_batch *batch)
+{
+	Assertion((texture >= 0), "batching_add_...() attempted for invalid texture");
+	if ( texture < 0 ) {
+		return;
+	}
+
+	batching_add_tri_internal(batch, texture, verts);
+}
```
JohnAFernandez, line 905: So ... what does this function do? Extra checking?

code/render/batching.cpp
```diff
@@ -872,10 +873,19 @@ void batching_add_quad(int texture, vertex *verts)
...
 
+void batching_add_quad(int texture, vertex *verts, primitive_batch *batch)
+{
+	Assertion((texture >= 0), "batching_add_...() attempted for invalid texture");
+	if ( texture < 0 ) {
+		return;
+	}
+
+	batching_add_quad_internal(batch, texture, verts);
+}
```
JohnAFernandez, line 884: Same here.

code/render/batching.cpp
EatThePath: Both of these were inherited from #5154's change sets, I didn't realize they'd still show up here. I've rebased on the merge commit on there and these have dropped out of the github change view, since they aren't actually changes.

JohnAFernandez (approved):

PR #5350, opened 2023-04-18
warn when a critical file might be be shadowed

notimaginative:
If any files considered critical (like a table or mission) might be shadowed then give a proper warning. The line in the debug log is also changed to say "A critical file being indexed..." for these instances so that they are easier to track down.

### DIFF
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -775,9 +775,10 @@ int is_ext_in_list( const char *ext_list, const char *ext )
 // Run a basic test for indexed files that may be shadowed
 #ifndef NDEBUG
 	#define ENABLE_SHADOW_CHECK		1
+	SCP_vector<SCP_string> critical_shadowed;
 #endif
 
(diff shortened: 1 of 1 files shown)


notimaginative:
@The-E @BMagnu You guys think this will be enough info to make it useful? I limited it to files from `data/tables` and `data/missions` for simplicity but that can be expanded in the future if needed.

code/cfile/cfilesystem.cpp
```diff
@@ -1143,6 +1150,24 @@ void cf_build_file_list()
 		}
 	}
 
+#ifndef NDEBUG
+	// if some special/critical files might be shadowed then make sure the user knows about it
+	if ( !critical_shadowed.empty() && !running_unittests ) {
+		SCP_string shadowed;
+		const auto count = critical_shadowed.size();
```
JohnAFernandez, line 1157: Only suggestion I see is that you could use this count variable in the loop range check instead of another `critical_shadowed.size()` call, since you already have it set up anyway.

JohnAFernandez (approved):

code/cfile/cfilesystem.cpp
notimaginative: Good point. I changed the loop late during testing to keep the dialog size down and didn't think about the obvious shortcut there.

PR #5351, opened 2023-04-18
FRED Dialog for Volumetrics

BMagnu:
This adds a menu for the new Volumetric Nebulae (#5231) in FRED.
The dialog contains (almost) all available features. One notable exception to this is an interactive editor for the noise functions, but as implementing that basically involves building something to quickly view 3D-Noise (or sliced 2D-Noise) at the minimum, and a blender-node-style editor at maximum, I will do this in a separate follow-up PR. Plus, if you really need it, you can still edit the mission file.
Also, as I have no proper way to do Qt-dev, no QtFRED editor as of yet.


fred2/volumetricsdlg.cpp
```diff
@@ -0,0 +1,391 @@
...
+	if (The_mission.volumetrics) {
+		const volumetric_nebula& volumetrics = *The_mission.volumetrics;
+		m_enabled = true;
+		m_volumetrics_hull = volumetrics.hullPof.c_str();
+		m_position = volumetrics.pos;
+		m_color = { static_cast<int>(std::get<0>(volumetrics.nebulaColor) * 255.0f),
+					static_cast<int>(std::get<1>(volumetrics.nebulaColor) * 255.0f),
+					static_cast<int>(std::get<2>(volumetrics.nebulaColor) * 255.0f) };
+		m_opacity = volumetrics.alphaLim;
+		m_distance = volumetrics.visibility;
```
Goober5000, line 108: Should either "distance" or "visibility" be renamed?

fred2/volumetricsdlg.cpp
```diff
@@ -0,0 +1,391 @@
...
+		m_steps = volumetrics.steps;
+		m_resolution = volumetrics.resolution;
+		m_oversampling = volumetrics.oversampling;
+		m_henyeyGreenstein = volumetrics.henyeyGreensteinCoeff;
+		m_sunFalloffFactor = volumetrics.globalLightDistanceFactor;
```
Goober5000, line 113: Same question about these two

fred2/volumetricsdlg.cpp
```diff
@@ -0,0 +1,391 @@
...
+	m_noiseScaleSub(14.0f),
+	m_noiseIntensity(1.0f),
+	m_noiseResolution(5)
+{}
+
+volumetrics_dlg::~volumetrics_dlg()
+{
+}
+
+BOOL volumetrics_dlg::OnInitDialog()
```
Goober5000, line 70: I think the first line of this function should be a call to `CDialog::OnInitDialog();`

Goober5000 (commented):
This is very well done.  I really like the use of macros; those might have been useful to simplify some other dialogs.

Just a few comments.

fred2/volumetricsdlg.cpp
BMagnu: They should probably both be called visibilityDistance or opacityDistance.

fred2/volumetricsdlg.cpp
BMagnu: This one, just like ``m_opacity = volumetrics.alphaLim;`` has a bit of an issue with "userfacing name" vs "technical name".
``globalLightDistanceFactor`` is a reasonably accurate name for what the variable actually affects within the shaders, namely serving as a factor for the distance travelled by the global light in the nebula.
In practice, while descriptive to a developer, this is very unintuitive to an end user. For an end user "Sun Falloff" seems easier to understand (I hope), because it describes what happens visually.
So I am not quite sure.
It would certainly be possible to rename the variable here, but then we have a variable that doesn't quite fit the name of the label next to its text input, and I am unsure about what I prefer there.

fred2/volumetricsdlg.cpp
BMagnu: But since unifying that involves renaming stuff in the volumetrics API, some labels in the lab, and so on, that should probably be in a separate PR

Goober5000: Fair enough

fred2/volumetricsdlg.cpp
Goober5000: I'm fine with keeping it as-is

Goober5000 (approved):

PR #5353, opened 2023-04-19
Joe fixes

Goober5000:
Two fixes for bugs encountered by RedMageJoe on today's stream, plus a fix for an issue in Burning Heaven.

1. if a ship can't find a dock path when ordered to dock, the game shouldn't crash
2. if the Instructor is known to a mission, training messages should depend on its presence or absence (this fixes training messages after the instructor is destroyed, while also preserving the new functionality to send training messages even if no instructor exists)
3. properly handle wingmen that are set to be destroyed before the mission starts


wookieejedi (approved):
Good finds and teamwork, LGTM

PR #5354, opened 2023-04-20
Organize and map additional keys to librocket

MjnMixael:
Simple PR that slightly reorganizes the keymapping list and maps some of the other keys that were originally left off. Without this, librocket/scpui cannot detect those keypresses.


wookieejedi (approved):
LGTM

PR #5355, opened 2023-04-20
enhance some SEXP warnings

Goober5000:
These messages add the context of the actual SEXPs causing the warnings to be displayed.  Motivated by Admiral Nelson's mod debugging work.


wookieejedi (approved):
Useful and straightforward

PR #5360, opened 2023-04-21
improve spawn collision checking

### LINKED ISSUE #5357: Location of arrrrwalktheplank ship spawn (Volition Bravos) influenced by prop ships.
In missions with very large ship props (e.g. "ships" with rotating pofs for use as moving backgrounds, such as the asteroid background in BTRL) the spawn location for Volition Bravos is set very far away from the player as seen in the picture.
screen0161

Reproduceable in Solaris, missions 1-3 & 5.fs2. Removing the prop background object (Astlayer.pof) makes VB warp in closer. 
In other missions, Volition Bravos appears as expected, near the player.

Goober5000:
Take into account the collision flag and collision groups when spawning a ship.  Should fix #5357.

Also clean up the code a bit, including setting a few `int`s to `bool`s and adding a missing axis case.

### DIFF
--- a/code/network/multi_respawn.cpp
+++ b/code/network/multi_respawn.cpp
@@ -13,6 +13,7 @@
 #include "network/multi_respawn.h"
 #include "network/multi.h"
 #include "object/object.h"
+#include "object/objcollide.h"
 #include "globalincs/linklist.h"
 #include "network/multimsgs.h"
 #include "network/multiutil.h"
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):
Makes sense and LGTM

Goober5000:
Darius also confirmed it fixes the issue

PR #5361, opened 2023-04-21
FRED object editor QOL improvements

### LINKED ISSUE #5356: FR: better handling of "point to" feature for orientation editor in FRED
Current implementation of the "point to" feature requires the developer to needlessly check a box when simply selecting a ship in the Object: field should suffice. Object: field also requires developers to pick from a dropdown menu instead of typing in a name, this makes things extremely clunky in object-dense missions. Alternate solution would be a point-and-click selection method for determining the object the ship(s) should be oriented to.

Goober5000:
1. When the "object" radio button is selected, the "point to" check box will also be selected
2. When an object is selected from the object list, the "object" radio button and "point to" check box will also be selected

Fixes #5356.


wookieejedi (approved):
Useful and straightforward

PR #5366, opened 2023-04-22
some FRED cleanup

Goober5000:
1. Disambiguate dialog IDs for the music player and the volumetric nebula dialogs
2. Fix compilation of qtFRED
3. Small refactor of Personas access to use STL functions where appropriate


wookieejedi (approved):
LGTM

PR #5368, opened 2023-04-23
Fix max Z speed when using ETS and engine damage tracking

### LINKED ISSUE #5367: Bug with waypoint speed following and damaged engines
```
DefCynodont — Yesterday at 10:39 PM
I have just confirmed, in the MVPs, at 48% engine integrity, a Faustus will become "effectively disabled". 
it looks like the moment the engine hitpoints are supposed to trigger the max-speed reduction, it just reduces the max speed to zero
```

[https://discord.com/channels/223511295431933953/223511391531827200/1099525030469181471](url)

Confirmed that #5292 causes this. 

Test mission:
[EngineHealthSpeedTest.zip](https://github.com/scp-fs2open/fs2open.github.com/files/11302827/EngineHealthSpeedTest.zip)

wookieejedi:
Follow up for #5137 and #5292. This engine strength check thus needs to be added to all instances where `max_vel.xyz.z` is updated, so a function was created and added in respective locations. Fixes #5367.

### DIFF
--- a/code/hud/hudets.cpp
+++ b/code/hud/hudets.cpp
@@ -118,7 +118,6 @@ void update_ets(object* objp, float fl_frametime)
 		}
 	}
 
-	// update max speed, moved from 'update_ets' function as part of max speed cleanup -Asteroth and wookieejedi
 	// AL 11-15-97: Rules for engine strength affecting max speed:
 	//						1. if strength >= 0.5 no affect
 	//						2. if strength < 0.5 then max_speed = sqrt(strength)
@@ -130,9 +129,9 @@ void update_ets(object* objp, float fl_frametime)
 
 	// only update max speed if aggregate engine health has changed
 	// which helps minimze amount of overrides to max speed
-	if (eng_current_strength != ship_p->prev_engine_aggregate_hits) {
+	if (eng_current_strength != ship_p->prev_engine_aggregate_strength) {
 		ets_update_max_speed(objp);
-		ship_p->prev_engine_aggregate_hits = eng_current_strength;
+		ship_p->prev_engine_aggregate_strength = eng_current_strength;
 
 		// check if newly updated max speed should be reduced due to engine damage
 		// don't let engine strength affect max speed when playing on lowest skill level
--- a/code/hud/hudets.cpp
+++ b/code/hud/hudets.cpp
@@ -57,6 +57,7 @@ void ets_init_ship(object* obj)
 // -------------------------------------------------------------------------------------------------
 // update_ets() is called once per frame for every OBJ_SHIP in the game.  
 // The amount of energy to send to the weapons and shields is calculated.
+// The max speed is also updated if there is a change in aggregate engine health.
 // The amount of time elapsed from the previous call is passed in as the parameter fl_frametime.
 //
 // parameters:   obj          ==> object that is updating their energy system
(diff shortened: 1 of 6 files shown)


code/hud/hudets.cpp
```diff
@@ -174,6 +159,34 @@ float ets_get_max_speed(object* objp, float engine_energy)
 	}
 }
 
+void update_max_speed(object* ship_objp)
+{
+	// get max possible speed then adjust and set based on engine health
```
JohnAFernandez, line 164: Need some asserts and early returns so that if someone calls this function incorrectly in the future, we won't be left with a mystery ctd.

JohnAFernandez (changes requested):

code/hud/hudets.cpp
wookieejedi: Thanks for kicking off the review. The `ets_get_max_speed` function above uses these checks, so I used that format in this new push.

JohnAFernandez: I mean that's all I saw.  It's not a large pr, and it makes sense.  Standardizing how a value is set should logocally fix the issue and prevent future bugs as well.

JohnAFernandez (approved):

code/ship/shiphit.cpp
```diff
@@ -2409,6 +2410,8 @@ static void ship_do_damage(object *ship_objp, object *other_obj, vec3d *hitpos,
 		// apply damage to subsystems, and get back any remaining damage that needs to go to the hull
 		damage = do_subobj_hit_stuff(ship_objp, other_obj, hitpos, submodel_num, damage, &apply_hull_armor);
 
+		ets_update_max_speed(ship_objp);
+
```
Baezon, line 2414: I would be as specific as possible and check directly for an engine subsystem being damaged to under the threshold inside the function. The case where we *know for certain* the max speed *will* change, not just whenever the ship takes damage.

The idea is to allow scripts to mess with velocity as much as possible, and only override it when we absolutely must.

Baezon (changes requested):

code/hud/hudets.cpp
```diff
@@ -117,6 +118,31 @@ void update_ets(object* objp, float fl_frametime)
...
+	//						2. if strength < 0.5 then max_speed = sqrt(strength)
+	//
+	//					 This will translate to 71% max speed at 50% engines, and 31% max speed at 10% engines
+	//
+
+	float eng_current_strength = ship_get_subsystem_strength(ship_p, SUBSYSTEM_ENGINE);
+
+	// only update max speed if aggregate engine health has changed
+	// which helps minimze amount of overrides to max speed
+	if (eng_current_strength != ship_p->prev_engine_aggregate_hits) {
```
Baezon, line 133: Minor, but since we're going to compare this to `strength` it should probably be named as such instead of `..._hits` since we aren't using direct hp values.

code/hud/hudets.cpp
```diff
@@ -117,6 +118,31 @@ void update_ets(object* objp, float fl_frametime)
 		}
 	}
 
+	// update max speed, moved from 'update_ets' function as part of max speed cleanup -Asteroth and wookieejedi
```
Baezon, line 121:
```suggestion
```

Baezon (changes requested):

code/hud/hudets.cpp
wookieejedi: Good point, just updated

Baezon (approved):

PR #5372, opened 2023-04-24
Ensure pmi for moveable submodel glow points in 3d ship select

### LINKED ISSUE #5371: Viewing MVPS Medusa in Ship/Weapon select causes CTD
Viewing the current Medusa during ship/weapon select in the MediaVPs while 3D Ship/Weapon select is turned on will cause an immediate CTD.

Baezon:
Follow-up for #5344, which exposes a reliance on a valid `pmi`! The 3d ship select doesn't have one, maybe we can get one I dunno, but skip this if there is no `pmi`, the submodel probably isn't moving anyway.

Fixes #5371

### DIFF
--- a/code/model/modelrender.cpp
+++ b/code/model/modelrender.cpp
@@ -1672,11 +1672,14 @@ void model_render_glowpoint_bitmap(int point_num, vec3d *pos, matrix *orient, gl
 		vm_vec_sub(&loc_offset, &gpt->pnt, &submodel_static_offset);
 
 		tempv = loc_offset;
-		if (IS_VEC_NULL(&loc_norm)) {	// zero vectors are allowed for glowpoint norms
-			model_instance_local_to_global_point(&loc_offset, &tempv, pm, pmi, bank->submodel_parent);
-		} else {
-			vec3d tempn = loc_norm;
-			model_instance_local_to_global_point_dir(&loc_offset, &loc_norm, &tempv, &tempn, pm, pmi, bank->submodel_parent);
+		if (pmi) {
+			if (IS_VEC_NULL(&loc_norm)) {	// zero vectors are allowed for glowpoint norms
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):

code/model/modelrender.cpp
```diff
@@ -1672,11 +1672,14 @@ void model_render_glowpoint_bitmap(int point_num, vec3d *pos, matrix *orient, gl
...
 
 		tempv = loc_offset;
-		if (IS_VEC_NULL(&loc_norm)) {	// zero vectors are allowed for glowpoint norms
-			model_instance_local_to_global_point(&loc_offset, &tempv, pm, pmi, bank->submodel_parent);
-		} else {
-			vec3d tempn = loc_norm;
-			model_instance_local_to_global_point_dir(&loc_offset, &loc_norm, &tempv, &tempn, pm, pmi, bank->submodel_parent);
+		if (pmi) {
+			if (IS_VEC_NULL(&loc_norm)) {	// zero vectors are allowed for glowpoint norms
+				model_instance_local_to_global_point(&loc_offset, &tempv, pm, pmi, bank->submodel_parent);
```
Goober5000, line 1677: Without calling a function like this, `loc_offset` would remain local to the submodel.  If there is a hierarchy of submodels, the coordinates would be local to the leaf submodel and would not appear in the proper place overall.

Instead, call `model_local_to_global_point`, here and in the next block, if `pmi` is null.  This is the equivalent function that assumes all submodels remain at their origin position and rotation.

Goober5000 (changes requested):

code/model/modelrender.cpp
Goober5000: Oh... you're right, `model_find_submodel_offset` does everything we need to do here.

Goober5000 (approved):

PR #5373, opened 2023-04-24
Remove model octants

Baezon:
This data structure had the gall to expose a bug I wrote so I will destroy this data structure.

In all seriousness, the octants are a data structure which contains pointers to the polygon centers, arranged into the 8 octants of the model's bounding box. There is a lot of faff to create and manage these things, they are rarely used (possibly more so in the past) and the single 8x optimization layer (?) is weak and not thorough (compared to something like the proper bsp trees). So I have removed it completely, and everything which used it has had equivalent replacements.

1. AI uses octants to choose random polygon centers to shoot at. In my infinite wisdom, in version 23 pofs, which no longer contain polygon center data, I thought to instead create the center vector on the fly... and then fill the pointer with a reference to that local variable. This is the bug, which luckily in practice resulted in 0 vectors for all of them, rather than them pointing god knows where, but still causes the issue of all AI aiming at the center of any version 23 pof model, instead of all over the hull.
Instead the job of filling a list polygon centers is given to the bsp collision tree, which seems a much more apt location. AI now choose from this vector. They used to only choose from the 4 closest octants, rather than from all of them at once, but since they do a ray collision to the chosen point, they will choose points on the side of the ship facing them either way.
2. Slash beams would use the bounding boxes of the octants as well as a big hardcoded table to choose slash points. Choosing from opposing bbox face centers, opposing bbox edge centers, or opposing bbox corners. This is really unncecessary, the overall bounding box and a pinch of math is more then sufficient to do the same thing.
3. Shields without collision trees (old pofs), would use the octants to do a quick out and narrow down to the octant before searching individual tris. Losing this is a shame, but this optimization certainly isn't as useful as it once was, new pofs will have bsp trees which are much better at this job, and isn't worth keeping octants for.


wookieejedi (approved):
Definitely will be good to fix AI firing at more than just large ship centers. Tests work as expected for me.

PR #5374, opened 2023-04-25
Add FRED support for the "required weapon" feature

Goober5000:
The "required weapon" feature has existed since 2012, but FRED support had never been implemented.  This adds it.


wookieejedi (approved):
Will be good to have this accessible and overall LGTM

PR #5375, opened 2023-04-25
Modernise CFLAGS on Mac/Linux

Goober5000:
Per GCC optimisation guide, which applies equally well to clang:
* -funroll-loops should not be used unless directed by profiling. The compiler will already unroll loops when it's a good idea.
* clang and gcc now have -Og optimisation levels designed to enable debugging with good performance.

(second attempt at PR #4562)

### DIFF
--- a/cmake/toolchain-apple-clang.cmake
+++ b/cmake/toolchain-apple-clang.cmake
@@ -4,7 +4,7 @@ MESSAGE(STATUS "Doing configuration specific to Apple Clang...")
 # TODO: Actually add Mac specific flags here...
 
 # Suppress specific warning
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -funroll-loops -fsigned-char -Wno-unknown-pragmas")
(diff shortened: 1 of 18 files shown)


Goober5000:
Well I solved the unused variable warnings, but now there's this interesting warning:
```
/__w/fs2open.github.com/fs2open.github.com/code/jpgutils/jpgutils.cpp: In function 'int jpeg_read_bitmap(const char*, ubyte*, ubyte*, int, int)':
/__w/fs2open.github.com/fs2open.github.com/code/jpgutils/jpgutils.cpp:170:56: error: argument 'image_data' might be clobbered by 'longjmp' or 'vfork' [-Werror=clobbered]
  170 | int jpeg_read_bitmap(const char *real_filename, ubyte *image_data, ubyte * /*palette*/, int dest_size, int cf_type)
      |                                                 ~~~~~~~^~~~~~~~~~
```

Goober5000:
@notimaginative pinging since it looks like you did a lot of work in that function

Goober5000:
Ok.  I'll have to get to this after the weekend but I'll make the change.

Goober5000:
Ok, done.

notimaginative (approved):

PR #5377, opened 2023-04-27
remove some obsolete FRED components

Goober5000:
- remove FolderDlg which hasn't been needed since we dropped support for Visual Studio 6
  - (see version history of MFC, https://mariusbancila.ro/blog/2015/08/12/version-history-of-vc-mfc-and-atl/)
- remove dialog1 which isn't used


wookieejedi (approved):
Useful cleanup

PR #5379, opened 2023-04-27
Some additional splash screen features

MjnMixael:
Adds a few features, tabled in game_settings, for fading in and out the splash screen. Fade in and fade out are separate values. Current splash logo and splash image use the exact same fade times. This PR also includes a toggle to automatically center the splash logo rather than draw it top-left.

It's worth noting this moves `game_title_screen_close()` above the script hooks for On Game Init and On Splash Screen. This is so that anything that runs on Game Init that might draw to the screen won't have the splash screen fade-out rendered directly afterwards. I think the hooks were put here originally so that scripts could run On Splash Screen. I briefly considered a new hook On Splash End, but figured since On Splash Screen is already semi deprecated and already runs after splash drawing this would be the easier change.

~~Submitting as draft because I still need to figure out how to do this based on actual game time rather than just using a fade multiplier which is likely subject to processor speeds. I also still want to look into allowing playing a sound as requested by Darius.~~

Sound is not currently possible because ffmpeg isn't initialized at this point. That might be solvable, but not for this PR.


wookieejedi (approved):
Great quality of life features to help bring the splash screen. Tests work well and overall LGTM

Goober5000 (changes requested):

PR #5385, opened 2023-04-30
Update recent particle optimizations to fix edge case slowdowns

wookieejedi:
#5328 added a very useful culling for particles created by specific particle effects (ie weapons, now beams), though in an effort to simplify the culling process, that PR removed the necessary culling of default smoke on ships. This PR restores the culling of default smoke while also keeping the new culling of particle specific effects. Tested and works as expected, including fixing the slowdowns caused by the aforementioned removal of smoke effect culls. Thanks to @Baezon for all his discussion and help with this.


Baezon (approved):
The old culling, while bespoke and only working on the 'smoke' types, was *very* harsh and it turns out on heavy missions that harshness is required to some degree.

EatThePath:
I can confirm significant performance improvement in high intensity missions in this branch vs recent nightlies

PR #5388, opened 2023-05-01
Fix HDR conversions in main shader

BMagnu:
#5055 changed the way SDR->HDR conversion was handled for textures.
This was incorrect for some cases, causing #5282.
#5282 however did not fully fix the issues but only fix them for handling of envmaps (and in the process likely incorrectly appyling SDR->HDR conversions in the SDR-only non-deferred pipeline, but I have not validated this).
This PR properly fixes the SDR->HDR conversion for all textures simply by reverting to how the old shader deals with the conversion, just with the new style of conditionals in the shader. This fixes some models being way too dark models with high specularity and low-light glowmaps.

### DIFF
--- a/code/def_files/data/effects/main-f.sdr
+++ b/code/def_files/data/effects/main-f.sdr
@@ -236,6 +236,10 @@ void main()
 		}
 		baseColor = texture(sBasemap, vec3(diffuseTexCoord, float(sBasemapIndex)));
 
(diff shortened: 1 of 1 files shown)


Baezon:
I can also confirm this fixes a bunch of issues I noted, but didn't actually understand!

wookieejedi:
Tested with FotG and also looks good.

The-E (approved):

PR #5392, opened 2023-05-04
Collision fix for weapons with lots of inherited velocity

Baezon:
A weapon can exceed its own `max_vel` via inherited velocity, and otherwise screw up this collision extrapolation.


Goober5000 (approved):
Are there any other places this might need to be checked?

Baezon:
Yes, but it appears to be already done so. `weapon_will_never_hit` and `check_inside_radius_for_big_ships` are the main ones that do this sort of extrapolation and `weapon_will_never_hit` already accounts for this.

PR #5394, opened 2023-05-06
Add additional checks for Animation List handling of nonexistent polymodel instances

BMagnu:
Fixes a crash when something tries to actively start an animation on an already no-longer existing ship.


PhantomHoover:
This appears to fix a nasty bug in Solaris testing where a ship dying after its parent will cause the animation code to try to run on an empty animation list, which without this fix immediately trips an assert.

wookieejedi (approved):

PR #5396, opened 2023-05-11
Fix two undocking bugs for small ships

wookieejedi:
Summary:
Fighters and bombers object types can be edited to accept docking or undocking orders. Undocking presented two bugs:

1. Undock never had the path properly reset to -1 when done (dock does). This lead to fighters or bombers never being able to properly attack large ships after undocking due to a conditional if check regarding paths in `ai_big_chase()` (specifically line 1338 in in `aibig.cpp`). This PR properly resets the path when undocking of a fighter is complete. (Recall undocking is completely seperate then bay arrivals or departures). This bug was likely not noticed before now because Fighters and bombers cannot normally undock, only re-arm, that has to be edited in the objecttables, so in retail it was never noticed. Also, re-arm has it's own path and orders, so that case got taken care of elsewhere. In addiition most ships that do undock are cargo ships or ships with turrets, and thus do not have primary weapons, which means the the primary bug never got noticed. Finally, fighters and bombers that are allowed and then given an undock order can still attack other fighters and bombers, this edge case was only when they would try to attack a large ship, so I'm guessing even mods that used them that way didn't really notice.

2. Previously, player or sexp given orders could override an ongoing undock order, which resulted in ships snapping back into place and other not correct movements. Joe recently documented this on his streams here (https://www.youtube.com/watch?v=0L-1FiYmN1g 10:40 mark, thanks to Goober too for the reference).  As it turns out there was already a fix in place to deal with preventing orders when reariming, so the same solution was used to fix the undocking orders.

Both fixes are tested and confirmed. Happy to answer additional questions, and once again thanks to Goober and BMagnu for assistance in tracking this down.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -3338,6 +3338,8 @@ void ai_do_objects_undocked_stuff( object *docker, object *dockee )
 		dockee_aip->support_ship_objnum = -1;
 		docker_aip->support_ship_signature = -1;
 		dockee_aip->support_ship_signature = -1;
+		// clear out points for docker 
+		docker_aip->path_start = -1;
 	}
 
 	// unlink the two objects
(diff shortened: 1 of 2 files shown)


Goober5000:
Excellent detective work and write-up. 👍

Baezon (approved):

Goober5000 (approved):

PR #5397, opened 2023-05-12
add ship:isPlayer()

Goober5000:
Add a Lua function to check whether a ship is a player ship.


wookieejedi (approved):
Definitely a useful and straightforward change, LGTM

PR #5399, opened 2023-05-12
Fix modular pof handling of dock / path parent names

BMagnu:
Well, turns out that the "submodel_parent" field of the paths are misused as subsystem names in certain cases, even if a submodel isn't found.
So allow modular pofs to rename these even if they don't actually match a submodel (unless a submodel was required before), and also allow to explicitly clear the name in case it was a subsystem before.

### DIFF
--- a/code/model/modelreplace.cpp
+++ b/code/model/modelreplace.cpp
@@ -877,6 +877,10 @@ void VirtualPOFOperationAddDockPoint::process(polymodel* pm, model_read_deferred
 			strcpy_s(pm->paths[i + splinefrom].parent_name, targetParentSubsystem->c_str());
 			pm->paths[i + splinefrom].parent_submodel = submodel_index;
 		}
+		else {
+			//Make sure that the path's parent name is cleared in this case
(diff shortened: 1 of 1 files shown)


Baezon (approved):

code/model/modelreplace.cpp
```diff
@@ -938,26 +942,33 @@ void VirtualPOFOperationAddPath::process(polymodel* pm, model_read_deferred_task
...
-
 	int submodel_index = -1;
+	bool clear_parent = false;
 	if (targetParentSubsystem) {
-		submodel_index = model_find_submodel_index(pm, targetParentSubsystem->c_str());
-		if (submodel_index < 0) {
-			Warning(LOCATION, "Submodel %s does not exist on POF %s for virtual POF %s. Returning original POF", targetParentSubsystem->c_str(), pm->filename, virtualPof.name.c_str());
-			return;
-		}
+		if (*targetParentSubsystem == "<none>")
```
Goober5000, line 948: I suggest a case-insensitive match using `SCP_string_lcase_equal_to`.  A modder could conceivably use `<None>`

PR #5400, opened 2023-05-12
MSAA

BMagnu:
This PR adds MSAA to our deferred rendering pipeline.
What this basically means, is that by adding ``-msaa 4`` (or for strong systems ``-msaa 8``) to the command line, you can activate MSAA, which will smooth any glinting or other artifacts caused by subpixel geometry. For an example, see this comparison, where the original code causes flickers and glinting on the rotating ISD, while the MSAA pipeline looks a lot smoother (only makes sense if you watch it in fullscreen on at least a 1080p monitor).

https://github.com/scp-fs2open/fs2open.github.com/assets/6238428/ccd8b01f-f28d-41ba-a820-fdffc469c7fe

On the technical sides, it uses some fancy tricks to allow MSAA to work in a deferred pipeline without incurring the cost of potentially shading all samples, not just all pixels. A paper on this will hopefully follow soon.


wookieejedi:
Very fantastic PR to have. The updated aspect is even more apparent in full-screen in-game for folks who might be curios :)

MjnMixael:
Anti-aliasing is already a selectable option in the Options Manager used by SCPUI. This new setting should be added to that list. The relevant code is in 2d.cpp at line 405.

BMagnu:
Yeah, it should get an options manager option.
However, as it is to be used in addition to traditional AA, it needs its own

The-E:
Haven't run the code yet on my end, but definitely great work.

One suggestion though: since the command line parameter can have only three valid values, the command line parser should check that the value set is one of them.

BMagnu:
> Haven't run the code yet on my end, but definitely great work.
> 
> One suggestion though: since the command line parameter can have only three valid values, the command line parser should check that the value set is one of them.

This validation _is_ done, but it's part of the graphics pipeline.
This is due to two things:
1. it may be dependent on the backend. The vulkan implementation may have different restrictions.
2. It is dependent on hardware, so a glGetintegerfv call is necessary to figure out what the hardware supports.

The-E:
> This validation _is_ done, but it's part of the graphics pipeline. This is due to two things:
Yep, seen the check in the code, it's all good. I was just thinking that since the only allowable values are 0, 4, 8 and 16, we can just check for those directly at the parsing stage; that we need to check with opengl if we can actually do that much msaa is fine, but that doesn't actually stop us from doing a basic sanity check during cmdline parsing.

> 1. it may be dependent on the backend. The vulkan implementation may have different restrictions.
It won't. This is dependant on the hardware or driver; vulkan doesn't unlock more capabilities at this basic a level 
> 2. It is dependent on hardware, so a glGetintegerfv call is necessary to figure out what the hardware supports. This is also only possible once the window has initialized, and not yet during command line parsing
yep, this is true

Just to be clear: I think the validation as currently implemented is usable as-is. It's just that I personally would prefer we reject obviously bogus values outright as early as possible, and since there are only a handful of values that could ever be valid, this is something we can do here.

BMagnu:
I suppose.
But then it is on whoever implements Vulkan (so you, I guess) to potentially adjust the pre-validation of MSAA values in case vulkan can be more permissive.

code/graphics/opengl/gropengldeferred.cpp
```diff
@@ -92,7 +106,57 @@ void gr_opengl_deferred_lighting_end()
 
 	GR_DEBUG_SCOPE("Deferred lighting end");
 
+	if (Cmdline_msaa_enabled > 0) {
+		GL_state.BindFrameBuffer(Scene_framebuffer);
```
The-E, line 110: Just for completeness' sake, adding a debug scope here would be good

```suggestion
		GR_DEBUG_SCOPE("MSAA Pass");
		GL_state.BindFrameBuffer(Scene_framebuffer);
```

BMagnu: good idea

The-E (approved):

code/def_files/data/effects/msaa-f.sdr
```diff
@@ -0,0 +1,198 @@
...
+	float fov;
+};
+
+const float voxelDepth = 2.5f;
+const float voxelDepthFalloff = 2.5f;
+
+#define SORT(a, b) if (a > b) { float t = dists[a]; dists[a] = dists[b]; dists[b] = t; }
+
+//The following sorting networks for median calculation are NOT complete. Calculations that don't affect the
+//median entry are remive, and thus, while the median element is correct, the rest of the array is NOT fully sorted
```
Goober5000, line 28: "remive"?

PR #5404, opened 2023-05-15
Traitor overrides feature

MjnMixael:
Adds a new Traitor Overrides feature as discussed with WookieJedi. This allows much more fine-tuned control of the traitor debrief than persona system allows.

You define a traitor debrief in traitor.tbl under the new `#traitor overrides` section. Each debrief gets a unique Name identifier. You set the text, audio file, and recommendation text for each one. A new sexp is used in FRED to define which traitor override debrief to use.

The traitor override pointer defaults to nullptr. If it remains nullptr then the old persona setup is used. Otherwise the debrief will use the override.

### DIFF
--- a/code/mission/missionparse.h
+++ b/code/mission/missionparse.h
@@ -150,7 +150,7 @@ typedef struct mission {
 	int	command_persona;
 	char command_sender[NAME_LENGTH];
 	int debriefing_persona;
-	traitor_override* traitor_override_t;
+	traitor_override_t* traitor_override;
 
 	// Goober5000
 	char event_music_name[NAME_LENGTH];
--- a/code/mission/missionparse.h
+++ b/code/mission/missionparse.h
@@ -25,6 +25,7 @@
 #include "sound/sound.h"
 #include "mission/mission_flags.h"
 #include "nebula/volumetrics.h"
+#include "stats/scoring.h"
 
 //WMC - This should be here
 #define FS_MISSION_FILE_EXT				NOX(".fs2")
@@ -149,6 +150,7 @@ typedef struct mission {
 	int	command_persona;
 	char command_sender[NAME_LENGTH];
 	int debriefing_persona;
+	traitor_override *trtr_override;
 
 	// Goober5000
 	char event_music_name[NAME_LENGTH];
--- a/code/mission/missionparse.h
+++ b/code/mission/missionparse.h
@@ -150,7 +150,7 @@ typedef struct mission {
 	int	command_persona;
 	char command_sender[NAME_LENGTH];
 	int debriefing_persona;
-	traitor_override *trtr_override;
+	traitor_override* traitor_override_t;
 
 	// Goober5000
 	char event_music_name[NAME_LENGTH];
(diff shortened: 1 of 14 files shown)


code/mission/missionparse.h
```diff
@@ -149,6 +150,7 @@ typedef struct mission {
 	int	command_persona;
 	char command_sender[NAME_LENGTH];
 	int debriefing_persona;
+	traitor_override *trtr_override;
```
Goober5000, line 153: I'd suggest using `traitor_override_t` for the type and `traitor_override` for the field.  This is how it's done in several other places, e.g. `ai_profile_t`, `asteroid_t`, `opengl_shader_t`.  This would avoid the need to abbreviate the field.

code/stats/scoring.cpp
```diff
@@ -57,6 +59,17 @@ float Scoring_scale_factors[NUM_SKILL_LEVELS] = {
 	1.25f					// insane
 };
 
+traitor_override* get_traitor_override_pointer(const SCP_string& name)
+{
+	for (int i = 0; i < (int)Traitor_overrides.size(); i++) {
+		if (name == Traitor_overrides[i].name) {
```
Goober5000, line 65: This should ideally be case-insensitive.  Most other string lookups, including ones set by sexps, are case-insensitive.

code/parse/sexp.cpp
```diff
@@ -15821,6 +15832,18 @@ void sexp_set_debriefing_persona(int node)
 	The_mission.debriefing_persona = persona;
 }
 
+// MjnMixael
+void sexp_set_traitor_override(int node)
+{
+	SCP_string override = CTEXT(node);
+
+	if (!stricmp(override.c_str(), "none")) {
```
Goober5000, line 15840: Since the sexp argument list is populated using the token `SEXP_NONE_STRING`, that should be used here too.

fred2/sexp_tree.cpp
```diff
@@ -3519,6 +3519,12 @@ int sexp_tree::query_default_argument_available(int op, int i)
 			}
 			return 0;
 
+		case OPF_TRAITOR_OVERRIDE:
+			if (Traitor_overrides.size() > 0) {
```
Goober5000, line 3523: Here and in qtfred, `empty()` could be used instead of `size() > 0`.  Same effect either way, just a style thing,

Goober5000 (changes requested):

code/parse/sexp.cpp
MjnMixael: See `sexp_set_skybox_model_preload` or the `OPF_TEAM_COLOR` error check.

fred2/sexp_tree.cpp
MjnMixael: Style/consistency? Who cares? It's a free for all.

```
case OPF_ASTEROID_DEBRIS:
			if ((Asteroid_info.size() - NUM_ASTEROID_POFS) > 0) {
				return 1;
			}
			return 0;

		case OPF_MOTION_DEBRIS:
			if (Motion_debris_info.size() > 0) {
				return 1;
			}
			return 0;

		case OPF_BOLT_TYPE:
			if (Bolt_types.size() > 0) {
				return 1;
			}
			return 0;

		case OPF_TRAITOR_OVERRIDE:
			return Traitor_overrides.empty() ? 0 : 1;
```

code/mission/missionparse.h
Goober5000: The `_t` token should be the type name, as the t stands for type.

code/stats/scoring.cpp
Goober5000: If you are performing a *case-sensitive* check on a SCP_string, using == is perfectly reasonable.  But == won't work for case-insensitive checks.

And, separately, when performing case-insensitive checks, `SCP_string_lcase_equal_to()()` is favored over `stricmp` because it avoids unnecessarily converting both SCP_strings to C-strings.

fred2/sexp_tree.cpp
Goober5000: True enough.

Goober5000 (approved):
👍

PR #5406, opened 2023-05-17
Add player damage inflicted factor

Baezon:
Surprisingly not a thing until now! Adds a difficulty dependent buff to player damage.


Goober5000 (changes requested):
If this field is optional, then if it is never specified in a mod, it would scale all the damage to 0.

Looking at the code, the reason this doesn't happen for `player_damage_scale`, which is treated the same way, is that all the defaults are supplied in `code\def_files\ai_profiles.tbl`.  So you'd need to do the same for the new field.

Baezon:
Ah whoops, yes, forgot amend that file.

Goober5000 (approved):

Issue #5405, opened 2023-05-17
Nintendo Switch controllers cause FSO to crash


Vakarias:
Title largely sums this up. If a Switch controller is connected, FSO will crash with an assert from SDL_JoystickOpen.

```Assert: "_joystick != nullptr"
File: joy-sdl.cpp
Line: 540
Failed to open a joystick, get a coder!
```

z64555:
An inspection by @qazwsxal and Cyborg last night revealed an assert was being triggered within the constructor of `io::joystick::Joystick` defined at `joy-sdl.cpp:544`.

z64555:
@Vakarias Is the crash still occurring?

PR #5407, opened 2023-05-17
Fix 5405

z64555:
As it was recently proven with #5405, joysticks may be enumerated by the OS but may fail when SDL attempts to open them.  The original assert within the io::joystick::Joystick constructor was triggered when a Nintendo Switch controller was plugged in and effectively halted FSO from recovering.

This fix changes that `Assertion` into a throw and `enumerateJoysticks` has been updated to catch the thrown exception and log the error message.  It should allow users to play FSO with having the bugged-out controller disabled/unavailable.

These changes should also hide the GUID of the controller, since get-flags and joy-info need to enumerate the joysticks before they can get the GUID.

### DIFF
--- a/code/io/joy-sdl.cpp
+++ b/code/io/joy-sdl.cpp
@@ -547,8 +547,11 @@ namespace joystick
 	{
 		_joystick = SDL_JoystickOpen(device_id);
 
-		if (_joystick == nullptr)
-			throw std::exception("Failed to open a joystick, get a coder!");
+		if (_joystick == nullptr) {
+			SCP_stringstream msg;
+			msg << "Failed to open joystick with id " << device_id << ", get a coder!";
+			throw std::runtime_error(msg.str());
+		}
 
 		fillValues();
 	}
--- a/code/io/joy-sdl.cpp
+++ b/code/io/joy-sdl.cpp
@@ -329,7 +329,7 @@ void enumerateJoysticks(SCP_vector<JoystickPtr>& outVec)
 
 				outVec.push_back(std::move(ptr));
 			}
-			catch (const std::exception e)
+			catch (const std::exception &e)
 			{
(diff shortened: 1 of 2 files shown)


code/io/joy-sdl.cpp
```diff
@@ -537,7 +547,8 @@ namespace joystick
 	{
 		_joystick = SDL_JoystickOpen(device_id);
 
-		Assertion(_joystick != nullptr, "Failed to open a joystick, get a coder!");
+		if (_joystick != nullptr)
```
JohnAFernandez, line 550: `!= nullptr`

z64555:
oook... I was wondering why it was failing.  Looked again and this time it has meaningful error code.  Last time I looked I thought I saw some automation schenanigans and didn't want to deal with it.

https://github.com/scp-fs2open/fs2open.github.com/actions/runs/5004561006/jobs/8967295012?pr=5407#step:8:831

So, apparently M$ added some overrides to std::exception that's not in the C++ standard.  I'll need to look into something "simple" to allow throwing strings, maybe with std::runtime_error or std::logic_error.

Goober5000:
That may not be an API issue:
```
/__w/fs2open.github.com/fs2open.github.com/code/io/joy-sdl.cpp:332:32: error: catching polymorphic type 'const class std::exception' by value [-Werror=catch-value=]
  332 |    catch (const std::exception e)
```
Change that to `const std::exception &e` and see what happens.

z64555:
I'm not entirely happy with this solution, but I don't currently have enough willpower to fully investigate how we deal with recoverable errors.

Goober5000 (approved):

Goober5000:
It looks good to me, and the latest commit is reasonable and IMO non-hackish.

PR #5411, opened 2023-05-19
add support for loose lz4 compressed files

notimaginative:
Allows for loose lz4 files to be indexed and used properly by cfile. Based on changes from @Shivansps.

### DIFF
--- a/code/cfile/cfilesystem.cpp
+++ b/code/cfile/cfilesystem.cpp
@@ -897,9 +897,9 @@ void cf_search_root_path(int root_index)
 
 		for (auto &file : files) {
 			auto ext_idx = file.name.rfind('.');
-			auto org_name = file.name;
+			auto orig_name = file.name;
 
-			if ( ext_idx != SCP_string::npos && file.name.substr(ext_idx) == ".lz4" ) {
+			if ( ext_idx != SCP_string::npos && SCP_string_lcase_equal_to()(file.name.substr(ext_idx), SCP_string(".lz4")) ) {
 				file.name = file.name.erase(ext_idx, file.name.length());
 				ext_idx = file.name.rfind('.');
 			}
@@ -908,7 +908,7 @@ void cf_search_root_path(int root_index)
 				continue;
 			}
 
-			check_file_shadows(root_index, i, file.name, file.sub_path, org_name.c_str());
+			check_file_shadows(root_index, i, file.name, file.sub_path, orig_name.c_str());
 
 			cf_file *cfile = cf_create_file();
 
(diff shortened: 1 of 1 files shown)


code/cfile/cfilesystem.cpp
```diff
@@ -893,12 +897,18 @@ void cf_search_root_path(int root_index)
 
 		for (auto &file : files) {
 			auto ext_idx = file.name.rfind('.');
+			auto org_name = file.name;
+
+			if ( ext_idx != SCP_string::npos && file.name.substr(ext_idx) == ".lz4" ) {
```
Goober5000, line 902: Should the extension check be case-insensitive?

code/cfile/cfilesystem.cpp
```diff
@@ -893,12 +897,18 @@ void cf_search_root_path(int root_index)
 
 		for (auto &file : files) {
 			auto ext_idx = file.name.rfind('.');
+			auto org_name = file.name;
```
Goober5000, line 900: I'd call this `orig_name` because org often stands for organization.

code/cfile/cfilesystem.cpp
Shivansps: Compressor should only produce the extension in lowercase but i guess it could be modified by external sources.

daftmugi: Shouldn't the extension be `.lz41`, not `.lz4`?
The `.lz4` extension may cause confusion, because it is not standard LZ4 and tools, such as `lz4`, will not be able to read it.

Shivansps: Im not going to argue this, do whatever do you think is right and ill support it.

Goober5000: IIRC the cfile system is designed to handle only three-character extensions.  Back when support for APNG was added, there was quite a bit of difficulty with the `.apng` extension before we ended up just using `.png` for APNGs.

notimaginative: In this case it doesn't have to be three characters since it's basically stripped off and not officially present in the index as far as cfile is concerned. The three character issue is still a thing, it just doesn't apply here.

I do think it should just use `.lz4` though. All of my testing has been done with standard cmdline tools like `lz4` and it's worked fine. No compatibility issues have been encountered. As far as I know it's just the standard `lz4` format since this code only applies to regularly compressed files and not anything fancy for use in/with VP archives.

Goober5000 (approved):

code/cfile/cfilesystem.cpp
daftmugi: Well, I'm confused. I thought the plan was for FSO to only support LZ41 and not LZ4, because LZ4 does not support random access. The wiki page only mentions the `.lz41` extension.
https://wiki.hard-light.net/index.php/Compression_Support

notimaginative: Perhaps I'm confused too then. And my test setup must have been total bonkers since it all appeared to work for me using standard lz4 compressed files. @Shivansps is needed to clarify things I guess.

But if that is the case then I agree that `.lz41` is the proper... well, *only* way to go.

Shivansps: yeah it works in the same way as in the vps. the reason for that is that the lz4 frame format do not support random access for some reason. Files produced by the extandart tool are in a difirent format and they should not work.
.lz41 it is then.

PR #5412, opened 2023-05-19
Use volume particles for poofs

Baezon:
This why they can do 'gentle' fades into geometry, instead of clipping into them.


wookieejedi (approved):
Looks pretty straightforward to me, though would be worth a look by one of the graphics coders too I reckon.

PR #5413, opened 2023-05-19
Mission-defined cutscenes dialog editor

MjnMixael:
Finally implements a dialog editor for mission-defined cutscenes. It is based heavily on the Objectives editor.

One notable change is the MOVIE_* defines are changed to an enum:int similar to how we do sexps now, making it easy to add new ones in the future. Additionally, it changes how Post Briefing cutscenes are handled. It's really really (and I cannot stress this enough; **really**) dumb that those were added because they are identical to Pre Game, but whatever. The handling changes here allow preserving both functionality and mission file syntax.


fred2/missioncutscenesdlg.h
```diff
@@ -0,0 +1,83 @@
...
+ * All source code herein is the property of Volition, Inc. You may not sell 
+ * or otherwise commercially exploit the source or things you created based on the 
+ * source.
+ *
+*/
+
+
+
+#include "Sexp_tree.h"
+#include "mission/missiongoals.h"
```
Goober5000, line 13: I'm guessing this #include isn't needed for cutscenes.

fred2/missioncutscenesdlg.h
```diff
@@ -0,0 +1,83 @@
...
+
+/////////////////////////////////////////////////////////////////////////////
+// CMissionGoalsDlg dialog
```
Goober5000, line 16: This comment should be updated

fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+	Assert(i < (int)m_cutscenes.size());
+	m_cutscenes[i].formula = node;
+	return;
+}
+
+void CMissionCutscenesDlg::move_handler(int node1, int node2, bool insert_before)
+{
+	int index1, index2, s;
+	mission_cutscene g;
```
Goober5000, line 445: nitpick: since in the mission goals editor the `g` stood for `goal`, this should be changed to `c` for `cutscene`

Goober5000 (changes requested):
Overall this looks pretty good, and basing it on the mission goals editor was the right call.  The one major change needed is that the post-briefing cutscene really needs to be a synonym for the pre-game cutscene, not its own separate definition.

So the separate definition should be removed from the enum list as well as freespace.cpp and the two instances of missionsave.cpp.  My hunch is that there shouldn't need to be any additional compatibility code beyond the pre-existing compatibility check in missionparse.cpp -- the reason is that just because `$Post-briefing Cutscene:` is accepted during parse doesn't necessarily mean that it should be written on save.  (Note for example that both `$Message:` and `$MessageNew:` can be read but only `$MessageNew:` is written.)

This means that in the cutscene dialog drop-down, "Post-briefing" would not be a separate selectable option.  The only other consideration is letting the user know that pre-game and post-briefing are the same thing, but your list of cutscene descriptions already accomplishes that.

MjnMixael:
> Overall this looks pretty good, and basing it on the mission goals editor was the right call. The one major change needed is that the post-briefing cutscene really needs to be a synonym for the pre-game cutscene, not its own separate definition.
> 
> So the separate definition should be removed from the enum list as well as freespace.cpp and the two instances of missionsave.cpp. My hunch is that there shouldn't need to be any additional compatibility code beyond the pre-existing compatibility check in missionparse.cpp -- the reason is that just because `$Post-briefing Cutscene:` is accepted during parse doesn't necessarily mean that it should be written on save. (Note for example that both `$Message:` and `$MessageNew:` can be read but only `$MessageNew:` is written.)
> 
> This means that in the cutscene dialog drop-down, "Post-briefing" would not be a separate selectable option. The only other consideration is letting the user know that pre-game and post-briefing are the same thing, but your list of cutscene descriptions already accomplishes that.

Nah. Message and MessageNew are not synonyms. One is an actual upgrade. So that's a poor comparison. You wrote bad code when you added Post-Brief. Now it must be dealt with either by making it functionally irrelevant as you suggest or dealing with it explicitly as I suggest. Both are valid.

fred2/missioncutscenesdlg.cpp
MjnMixael: For the record, I am absolutely not updating this. Consider it me taking a stand regarding goober's OCD holding up fully functioning PRs over things as mundane as single-letter local variable names. Take the fully-functioning feature or don't. Your choice.

Goober5000: So you'll change a single-letter variable to "fu" but not to it actual abbreviation?

MjnMixael: Correct. Because I'm petty.

Goober5000: > Correct. Because I'm petty. And I have used the same variable naming scheme for the qtFRED version I've written. Because I'm very petty.

We are amused.  Very well, I won't make this a condition of my review.  It's more useful for the project to have your pettiness on record so that other coders know what they're dealing with.

code/mission/missionparse.cpp
```diff
@@ -1015,25 +1015,29 @@ void parse_cutscenes(mission *pm)
 
 		while (!optional_string("#end"))
 		{
-			// this list should correspond to the MOVIE_* #defines
-			scene.type = optional_string_one_of(8,
+			// this list should correspond to the MOVIE_* enums
```
Goober5000, line 1018: The code in missionparse.cpp should be reverted (except that changing "#defines" to "enums" is fine).  The previous code handled the situation adequately and the warning is not necessary -- in fact the warning is wrong because the intent is to fully support "$Post-briefing Cutscene" as a valid string.

code/mission/missionparse.h
```diff
@@ -94,13 +94,20 @@ typedef struct support_ship_info {
...
+	MOVIE_PRE_FICTION,
+	MOVIE_PRE_CMD_BRIEF,
+	MOVIE_PRE_BRIEF,
+	MOVIE_PRE_GAME,
+	MOVIE_PRE_DEBRIEF,
+	MOVIE_POST_DEBRIEF,
+	MOVIE_END_CAMPAIGN,
+
+	// Must always be at the end of the list
+	First_available_movie_id
```
Goober5000, line 109: `First_available_movie_id` is misleading because unlike sexp operators, there isn't a way for modders to add additional movie types.  Change this to something like `Num_movie_ids` or `Movie_id_count`.

Goober5000 (changes requested):
Looking at my notes from June, there were two other points:

code/mission/missionparse.cpp
MjnMixael: Actually, my intention is to deprecate it and that's what my code does.

code/mission/missionparse.h
MjnMixael: Really? Ok. I'm just closing the PR. Not worth it.

PR #5421, opened 2023-05-25
Virtual POFs: Modify Subsystems

BMagnu:
Allows modifying subsystem definitions


Goober5000 (approved):
Looks good, and `+Set Radius:` tests work.

PR #5424, opened 2023-05-26
patch qtFRED's handling of hud gauges

### LINKED ISSUE #5422: qtFRED does not correctly get custom hud gauges
suspect whoever added custom hud gauges as their own sexp a while back never bothered to update qtFRED

Goober5000:
There was a missing function call in qtFRED's initialization, which prevented custom HUD gauges from being loaded.  This adds that.  Additionally, this updates the deduplication code in `get_listing_opf_custom_hud_gauge` to match its counterpart in FRED.

Fixes #5422.


wookieejedi (approved):
LGTM

PR #5428, opened 2023-05-27
remove hud_squadmsg_save_keys()

Goober5000:
Remove this function and associated items, since they are commented out and unused.

### DIFF
--- a/code/controlconfig/controlsconfig.cpp
+++ b/code/controlconfig/controlsconfig.cpp
@@ -1359,9 +1359,6 @@ bool control_config_accept(bool API_Access)
 		}
 	}
 	
-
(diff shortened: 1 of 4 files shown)


z64555:
The heck is this hack you're removing, lol.

Also, is `keys_used` used by anything else? If not, it may also be a candidate for removal.

Goober5000:
Well, `keys_used` is used in `hud_squadmsg_read_key()` and `hud_squadmsg_get_key()`.  It does look like it has a purpose.

z64555 (approved):
I.. still don't know what the intent was behind the hack.  But since it was disabled to begin with we probably don't need it.

Issue #5109, opened 2023-01-14
Using a Container List in waypoint order causes int3


MjnMixael:
Supplying the waypoint for an ai-waypoints or ai-waypoints-once sexp by way of List Container -> get_random causes an int3 in aigoals.cpp at line 913.

PR #5429, opened 2023-05-27
Allow using container data with ai-waypoints* SEXPs

jg18:
Using Replace Container Data as the "waypoint path" argument in `ai-waypoints` and `ai-waypoints-once` currently works in FRED but triggers a crash ('Int3()') in game. This change fixes that.

Should fix Issue #5109.


JohnAFernandez (approved):

jg18:
On second thought, there are a number of SEXPs that use a waypoint path as an argument. Checking those as part of this PR would be useful, if not too much work.

jg18:
OK, I'm satisfied that the other SEXPs that use waypoint paths as an argument can't run into this issue, so pending manual test results, I think we're good to merge.

PR #5433, opened 2023-05-30
Sexp tree searching and filtering

Goober5000:
A feature that has been on the wish lists of many FREDders for a long time... the ability to search for, and filter, SEXP operators.

filtering

This adds the following:
1. A pop-up menu showing a list of operators
2. Type-to-search ability to filter the list
3. Approximate string matching, not just substring matching
4. Tooltips displaying SEXP help for each operator
5. Incompatible operators are disabled, cannot be selected, and show tooltips indicating why they are disabled
6. Enter-to-confirm, escape-to-cancel

Thanks to Lafiel for writing the `stringcost` algorithm for comparing string matches.


fred2/OperatorComboBox.cpp
```diff
@@ -0,0 +1,431 @@
...
+			}
+			else
+			{
+				buffer = helptext;	// never mind the MSVC warning; helptext will be non-null and non-empty here
+			}
+			replace_all(buffer, "\t", "    ");
+			m_tooltiptextA = buffer.c_str();
+			m_tooltiptextW = buffer.c_str();
+
+#ifndef _UNICODE
```
BMagnu, line 337: Since the code in both ifndef parts looks identical, could we not just leave it?

fred2/sexp_tree.cpp
```diff
@@ -1944,33 +1914,141 @@ SCP_string sexp_tree::match_closest_operator(const SCP_string &str, int node)
 	arg_num = find_argument_number(z, node);
 	opf = query_operator_argument_type(op, arg_num);	// check argument type at this position
 
-	for (i=0; i<(int)Operators.size(); i++) {
-		opr = query_operator_return_type(i);			// figure out which type this operator returns
+	// find the best operator
+	int best = sexp_match_closest_operator(str, opf);
+	if (best < 0)
+		return str;
```
BMagnu, line 1920: This needs to explicitly warn or something.
Also, not sure if just returning str is safe. It's probably _really_ easy to enter something that causes FRED to crash the next time something is done on the SEXP tree

fred2/sexp_tree.cpp
```diff
@@ -1944,33 +1914,141 @@ SCP_string sexp_tree::match_closest_operator(const SCP_string &str, int node)
...
+		case ID_EDIT_BG_COLOR:
+		case ID_REPLACE_NUMBER:
+		case ID_REPLACE_STRING:
+		case ID_ADD_STRING:
+		case ID_ADD_NUMBER:
+		case ID_EDIT_CUT:
+		case ID_DELETE:
+		case IDC_SEXP_POPUP_LIST:
+			if (id >= FIRST_OP && id < sexp::operator_upper_bound())
+				Error(LOCATION, "A resource definition (%d) overlaps with an operator value!", id);
```
BMagnu, line 2036: Since this is a debug-only error, and a static coding error and not a runtime data error, it should be erroring using Assert or unreachable

BMagnu (changes requested):
Some small comments

fred2/OperatorComboBox.cpp
Goober5000: Hmm, looks like you're right.

fred2/OperatorComboBox.cpp
```diff
@@ -0,0 +1,431 @@
...
+			{
+				int opr_type;
+				map_opf_to_opr(m_opf_type, opr_type);
+
+				buffer = "The operator \"";
+				buffer += op_index >= 0 ? Operators[op_index].text : "<invalid operator>";
+				buffer += "\" cannot be selected because it has an incompatible return type.\r\n\tReturns: ";
+				buffer += opr_type_name(query_operator_return_type(op_const));
+				buffer += "\r\n\tExpected: ";
+				buffer += opr_type_name(opr_type);
```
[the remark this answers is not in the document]
Goober5000: That's a pretty insightful question.  The last statement should be correct: The operator popup box will only be displayed for *operators*, not for any possible argument.  Therefore the only return values under consideration are the ones that have a known return type; and all known return types are listed in `opr_type_name`.

fred2/sexp_tree.cpp
Goober5000: I'll add a warning.  But FRED should be able to handle situations with invalid operators now.  I did a lot of stress testing of the sexp_tree code; see #4916 and #4932 for example.

BMagnu (approved):

PR #5437, opened 2023-05-31
Stop lighting profile warning in specific case

MjnMixael:
If you have a mod that doesn't define any lightning profiles and your missions don't have the `$Lighting Profile:` line in the file, then FSO will try to switch to a default lighting profile that doesn't exist and warn the user that it tried to switch to lighting profile `''`, a zero length profile name, and that it couldn't find that profile. So in that cause, just skip trying to switch lighting profiles instead of sending a warning.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -780,13 +780,16 @@ void parse_mission_info(mission *pm, bool basic = false)
 			WarningEx(LOCATION, "Mission: %s\nUnknown AI profile %s!", pm->name, temp );
 	}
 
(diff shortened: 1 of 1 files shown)


MjnMixael:
I should add that I didn't do an in-depth look at how profiles work. There might be a better way to fix this. But missions/mods that don't use lighting profiles shouldn't be getting a warning for not having a lighting profile for every mission load.

MjnMixael:
Er.. hang on. I actually think this is a deeper issue with how FRED is handling lightning profiles that's allowing a blank profile to be saved.

MjnMixael:
Different PR on the way.

PR #5439, opened 2023-06-01
Fixes for lab backgrounds and curve initialization

Goober5000:
1. Patch the string skipping behavior for backgrounds in the lab.  Not all the strings were accounted for in each section, which would occasionally cause some sections to be incorrectly skipped; and one of the strings was wrong.  Also tidy up the assignment of `ltp_name`.
2. Since `curves_init()` was added to FRED, add it to qtFRED too.  Also sort the list of #include statements in that file.

Followup to #5380 and #5408.  Also fixes a bug mentioned in Discord where background bitmaps were not being loaded in the lab.


wookieejedi (approved):
LGTM

PR #5442, opened 2023-06-04
some refinements for retiming of object collision pairs

Goober5000:
1. Add a function to retime *only* the collision pairs involving a specific object.  This is useful when only one object has moved, since only collisions involving that object are affected.
2. Use this function when objects are moved for autopilot, via sexp, and via script.

Previously collision pairs were not retimed when objects were moved via script, leading to bugs.  This fixes that.


wookieejedi:
Tested and works as expected and fixes bug cases. These were long standing bugs in setting large changes to positions and orientations via script that were finally found with a reproducible test mission, and with this PR these issues are finally resolved and fixed. I would be happy to provide additional detail and test examples to highlight that highlight the previous bugs if anyone requests them, too.

BMagnu (approved):

Baezon (approved):

PR #5444, opened 2023-06-05
refinements of jump node features

Goober5000:
1. Deprecate the old set-jumpnode-name sexp now that there is a set-jumpnode-display-name
2. Make the display name design consistent with ships and weapons; in particular add a `HasDisplayName()` function and automatically generate the display name for hash-truncated names
3. Keep the display name in sync when the name is changed in the FRED dialog
4. Fix the overriding of hash-truncated names with display names on the HUD
5. Fix the incorrect setting of the alphacolor flag when using FRED
6. Fix array out-of-bounds access when no object is selected

Follow-up to #5401.


wookieejedi (approved):
Overall looks good to me

TRBlount (approved):

PR #5450, opened 2023-06-07
Validate chunk size to avoid align crash and add align skip flag

Shivansps:
This fixes the INF_Antaeus.pof crash of Inferno: Alliance due to unknown data found on the model BSP Data chain.

Explanation: 
While parsing the bsp data, the alignment code reads every chunk of the chain one by one, on this particular model on one of the chains at the final 8 bytes there is invalid data that looks like corrupted or trash data, this has an invalid chunk type and a negative chunk size, causing the bsp_pointer to go out of bounds.

Fix:
A check is added to make sure the chunk size value is not negative or has a value that may cause the bsp pointer to go over the max bsp_size.

A troubleshoot flag is also added to disable this process.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -235,6 +235,7 @@ Flag exe_params[] =
 	{ "-prefer_ipv4",		"Prefer IPv4 DNS lookups",					true,	0,									EASY_DEFAULT,					"Troubleshoot", "http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-prefer_ipv4", },
 	{ "-prefer_ipv6",		"Prefer IPv6 DNS lookups",					true,	0,									EASY_DEFAULT,					"Troubleshoot", "http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-prefer_ipv6", },
 	{ "-log_multi_packet",	"Log multi packet types ",					true,	0,									EASY_DEFAULT,					"Troubleshoot", "http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-log_multi_packet",},
+	{ "-no_bsp_align",		"Disable pof BSP data alignment",			true,	0,									EASY_DEFAULT,					"Troubleshoot", "http://www.hard-light.net/wiki/index.php/Command-Line_Reference#-no_bsp_align", },
 #ifdef WIN32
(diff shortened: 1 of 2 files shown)


code/model/modelread.cpp
```diff
@@ -6079,6 +6086,14 @@ uint align_bsp_data(ubyte* bsp_in, ubyte* bsp_out, uint bsp_size)
 			memcpy(&bsp_chunk_size, bsp_in + 4, 4);
 		}
 
+		//Chunk size validation, if fails change it to copy the remaining data in chain
+		auto max_size = end - bsp_in;
+		if (bsp_chunk_size < 0 || bsp_chunk_size > max_size) {
+			
+			mprintf(("Invalid BSP Chunk size detected during BSP data align: Chunk Type: %d, Chunk Size: %d, Max Size: %d\n", bsp_chunk_type, bsp_chunk_size, static_cast<uint>(max_size)));
```
Baezon: This should be a warning, it's a fairly serious problem the modder should know about.

Baezon (changes requested):

code/model/modelread.cpp
Shivansps: changed to a warning.

Baezon (approved):

code/model/modelread.cpp
```diff
@@ -6079,6 +6086,13 @@ uint align_bsp_data(ubyte* bsp_in, ubyte* bsp_out, uint bsp_size)
 			memcpy(&bsp_chunk_size, bsp_in + 4, 4);
 		}
 
+		//Chunk size validation, if fails change it to copy the remaining data in chain
+		auto max_size = end - bsp_in;
+		if (bsp_chunk_size < 0 || bsp_chunk_size > max_size) {
```
Goober5000, line 6091: Methinks neither the change to `int` nor the check for <0 is needed.  A negative number, interpreted as an unsigned value, will always be larger than the largest possible signed value.

code/cmdline/cmdline.cpp
```diff
@@ -2021,6 +2024,10 @@ bool SetCmdlineParams()
 	{
 		Cmdline_lang = lang_arg.str();
 	}
+
+	if (no_bsp_align_arg.found()) {
+		Cmdline_no_bsp_align = 1;
```
Goober5000, line 2029: Since this is a boolean, it should be assigned `true`, not `1`.

Goober5000 (changes requested):

code/model/modelread.cpp
Shivansps: you were right about that, it didnt even crossed my mind at the time. I changed it to use uint now.

Goober5000 (approved):

PR #5457, opened 2023-06-14
fix angle conversions in FRED

### LINKED ISSUE #5390: FRED Background Editor Warnings
I don't know where this bug is being created, but it's annoying and FRED needs to chill.

I have mission files made using the old background coordinate system. Newer FRED will convert those coordinates to a 0-359 setup. Bank is greyed out for suns. In some cases, converted Bank coordinates for suns can have a >359 value which FRED will warn about every time you click on something. But you can't change the value because the field is greyed out.

If bank is unused in the new coordinate system, then it should be zero'd automatically.

Goober5000:
In certain cases, angles would be rounded to 360, where the expected range is 0-359.  The range was guarded properly for skyboxes but not for bitmaps and suns.  This fixes that.  Additionally, since suns do not bank, the angle conversion now sets the bank to 0.

Fixes #5390.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -5471,7 +5471,10 @@ void parse_one_background(background_t *background)
 
 		// correct legacy bitmap angles which used incorrect math in older versions
 		if (!background->flags[Starfield::Background_Flags::Corrected_angles_in_mission_file])
+		{
 			stars_correct_background_sun_angles(&sle.ang);
+			sle.ang.b = 0.0f;	// stars do not bank
+		}
 
 		// scale
 		required_string("+Scale:");
--- a/fred2/bgbitmapdlg.cpp
+++ b/fred2/bgbitmapdlg.cpp
@@ -275,12 +275,9 @@ void bg_bitmap_dlg::create()
 	m_skybox_heading = fl2ir(fl_degrees(skybox_angles.h));
 
 	//make sure angle values are in the 0-359 degree range
-	if (m_skybox_pitch < 0)
-		m_skybox_pitch = m_skybox_pitch + 360;
-	if (m_skybox_bank < 0)
-		m_skybox_bank = m_skybox_bank + 360;
-	if (m_skybox_heading < 0)
-		m_skybox_heading = m_skybox_heading + 360;
+	if (m_skybox_pitch < 0) m_skybox_pitch += 360;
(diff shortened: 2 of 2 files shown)


Baezon (changes requested):
The usage of all these 'add 360' or 'subtract 360' seems suspicious to me, would we not be better served by having a strategically placed mod 360 somewhere? Since that seems to be heart of the issue.

Goober5000:
> The usage of all these 'add 360' or 'subtract 360' seems suspicious to me, would we not be better served by having a strategically placed mod 360 somewhere? Since that seems to be heart of the issue.

Not exactly.  Here's the conversion routine:
```
void stars_correct_background_sun_angles(angles* angs_to_correct)
{
	matrix mat;
	vm_angles_2_matrix(&mat, angs_to_correct);
	vm_transpose(&mat);
	vm_extract_angles_matrix(angs_to_correct, &mat);
	angs_to_correct->p = fmod(angs_to_correct->p + PI2, PI2);
	angs_to_correct->b = fmod(angs_to_correct->b + PI2, PI2);
	angs_to_correct->h = fmod(angs_to_correct->h + PI2, PI2);
}
```

The `fmod` function already ensures that the angle will be between 0 and 359.99999...  The problem is that numbers 359.5 and higher are rounded to 360.

The skybox adjustments are different because those are moving a range of -180-179 to 0-359.

Baezon:
Fair enough, although that does seem like the mod should then be when the float -> int cast happens. Part of it is that a mod 360 is much clearer as 'this value should always be 0-359' rather than the addition or subtraction, which don't necessarily convey that on their own at all, since there's lots of cases like being negative or >720 where it still wouldn't do that. You do have a comment that explains why this is done, but still. Although I'm reminded the plain % mod preserves sign which is dumb so I guess it wouldn't help for negatives...

Goober5000:
My thinking is:
1. As you said, plain mod doesn't work for the skybox angles
2. The bitmap and sun angles should already be in the correct range; the only issue is if we get an angle of 360, and that only happens with rounding
3. Either mod or subtraction can work to wrap 360 back to 0
4. Preferring subtraction allows a grossly-out-of-range angle to be flagged using the existing error checks

Baezon (approved):
Alright then, just wanted to talk it out.

Goober5000:
Certainly, always happy to do so.

PR #5462, opened 2023-06-17
allow getting of mission info even for files with long names

### LINKED ISSUE #5459: FRED allows for saving missions with filenames that are too long
It seems that FRED doesn't check when saving a mission whether the mission file name is too long.

The maximum file name is 31 chars (`MAX_FILENAME_LEN` - 1) *including* the ".fs2" extension.

The result is a crash when opening the Campaign Editor, specifically in `get_mission_info()` in missionparse.cpp, in the call to `strcat_s()`.

IMO FRED should refuse to save a mission if the FREDder specifies a file name that is too long.

Goober5000:
Issue #5459 notes that mission files that are too long will crash the FRED campaign editor, but this is an artificial restriction.  The only need for a string buffer here is to enforce the correct mission extension, so use a SCP_string for this, and add checks for whether string manipulation is in fact necessary.  (In most cases it isn't.)

Fixes #5459.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -6410,19 +6410,25 @@ bool post_process_mission(mission *pm)
 
 int get_mission_info(const char *filename, mission *mission_p, bool basic, bool filename_is_full_path)
 {
-	int filelength;
-	char real_fname_buf[MAX_FILENAME_LEN];
-	const char *real_fname = real_fname_buf;
+	static SCP_string real_fname_buf;
+	const char *real_fname = nullptr;
 	
 	if (filename_is_full_path) {
 		real_fname = filename;
 	} else {
-		strncpy(real_fname_buf, filename, MAX_FILENAME_LEN-1);
-		real_fname_buf[sizeof(real_fname_buf)-1] = '\0';
-	
-		char *p = strrchr(real_fname_buf, '.');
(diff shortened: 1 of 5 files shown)


wookieejedi (approved):

qazwsxal:
FWIW this isn't a completely artificial restriction, you can't pack files with names longer than 32 characters into a VP, so ideally there should be some sort of warning.

Goober5000:
Well at this point the VP is there only place there is a length restriction.  So the warnings ought to be built into VP editors, rather than FSO itself, IMO.

qazwsxal:
It's a *fatal error* on VP creation. There's an opportunity here to save fredders (particularly new ones unfamiliar with FSO's idiosyncrasies) from running into packaging issues later. It doesn't matter which VP viewer creates the file, if the official packaging format can't support filenames greater than 32 chars, then the mission editor should make the user aware of that.

MageKing17:
As long as VP files are the default (and like it or not, they are), and as long as most of the tools used to make VP files don't warn about too-long filenames (and I don't know whether any do), then FRED should absolutely warn when saving a mission that can't be packed into a VP file.

Goober5000:
Ok, fair enough.

wookieejedi (approved):

Issue #5458, opened 2023-06-16
Containers marked "Save on mission close" don't persist between missions


dilmahg:
While developing a two mission campaign, I created a map container with a single key/value combination with persistence option set to "Save on mission close" in the first mission. This data is unavailable in the second mission, even after confirming identical container names and persistence settings. Confirmed that data is unavailable using an "is-container-empty" sexp in mission 2.

jg18:
Thanks for filing this.

Does the issue occur with list containers?

Does the issue occur with containers marked "save on mission close" and/or "eternal"?

jg18:
@dilmahg I'm having trouble reproducing this issue. See attached test campaign.

The messages that are sent in the second mission use the values from the first mission. I included a persistent variable as an experimental control.

[issue_5458.fc2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/11772396/issue_5458.fc2.txt)
[issue_5458_1.fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/11772397/issue_5458_1.fs2.txt)
[issue_5458_2.fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/11772398/issue_5458_2.fs2.txt)

dilmahg:
Hey @jg18, I just had a look at those two and realised I forgot to mention a crucial detail - I was using the sexp "add-to-map" to add the first key/value combination to the map container in the first mission. I created the map container itself using the "Add/Modify Container" dialogue with the data type as "string" and key type as "string" too, but didn't give it any key/value combinations. 

Are you able to reproduce the issue using a sexp to add the key/value combination?

dilmahg:
Okay, I fixed it @jg18! By changing the containers to "save on mission complete", it worked. If you're happy, I can add that little tidbit to the wiki so people learn from my mistakes?

jg18:
Closing this issue as resolved. If someone finds a problem with container persistence in the future, please file a new issue instead of posting here, so that the discussions don't get mixed up. Thanks!

jg18:
Oh, so there really *is* an issue here. "Save On Mission Complete" works as expected, but "Save On Mission Close" doesn't. Reopening with updated title.

jg18:
Fixed with PR #5463.

PR #5463, opened 2023-06-18
Fix issues with campaign-persistent containers

jg18:
Campaign-persistent containers were getting clobbered by red alert containers and also weren't always being cleared alongside campaign-persistent variables.

Fixes Issue #5458.

### DIFF
--- a/code/mission/missioncampaign.cpp
+++ b/code/mission/missioncampaign.cpp
@@ -857,7 +857,9 @@ int mission_campaign_previous_mission()
 	}
 
 	// copy backed up containers over
-	Campaign.persistent_containers = Campaign.red_alert_containers;
+	for (const auto& ra_container : Campaign.red_alert_containers) {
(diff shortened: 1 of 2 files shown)


JohnAFernandez (approved):
Nice to see it was a simple fix.

code/mission/missioncampaign.cpp
```diff
@@ -857,7 +857,9 @@ int mission_campaign_previous_mission()
 	}
 
 	// copy backed up containers over
-	Campaign.persistent_containers = Campaign.red_alert_containers;
+	for (const auto& ra_container : Campaign.red_alert_containers) {
+		Campaign.persistent_containers.emplace_back(ra_container);
+	}
```
jg18, line 862: Using `std::copy()` both here and below would result in shorter code, although this version allows for more easily lining up this code with the equivalent code for persistent variables.

jg18:
@dilmahg verified the fix. Merging.

PR #5464, opened 2023-06-18
add const where appropriate to the modelrender API

Goober5000:
See title.  There were some cascade effects, but the changes are mostly contained to the model code.  The calls to `gr_render_model` and `g3_render_primitives` use `const_cast` because constifying that API would require a whole other PR.


wookieejedi (approved):
Straightforward cleanup and also works in my tests.

PR #5468, opened 2023-06-20
additional cleanup from PR 5462

Goober5000:
Some streamlining and constifying of the readyroom popup list and associated functions.

Followup to #5462.

### DIFF
--- a/code/hud/hudconfig.cpp
+++ b/code/hud/hudconfig.cpp
@@ -1266,7 +1266,7 @@ void hud_config_button_do(int n)
 
 	case HCB_SAVE_HCF:		
 		int exists;
-		char *out;
+		const char *out;
 
 		// get the text in the input control
 		exists = 0;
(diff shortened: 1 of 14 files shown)


code/hud/hudconfig.cpp
```diff
@@ -1708,10 +1708,9 @@ void hud_config_color_load(const char *name)
 {
 	int idx;
 	char str[1024];
-	char *fname;
 	ubyte r, g, b, a;
 
-	fname = cf_add_ext(name, ".hcf");
+	auto fname = cf_add_ext(name, ".hcf");
```
jg18, line 1713: `const auto`? Not required but makes it easier to tell it's `const`.

jg18: IMHO `const char *` is easier to read and not much more text than `auto` but seems like the latter is SCP standard.

Goober5000: At one point our CI would complain about long type names where `auto` would suffice, but that's currently disabled.  I'm fine with putting `const char *` here.

wookieejedi (approved):
Works in my tests and PR is straightforward.

PR #5469, opened 2023-06-21
tweak filename length check for PNGs

### LINKED ISSUE #5387: Spurious "Texture name too long" warnings on ships
If a ship texture name is long but not *too* long like 22 characters I guess, you start getting warnings like this:
```
Warning: Passed filename, 'iimpvarproornglo-reflect.png', is too long to support an extension and frames!!
Maximum length for an ANI/EFF/APNG, minus the extension, is 22 characters.
```
This makes sense, because a texture is first attempted as animated, and then if that fails it falls back to non-animated (a common control flow for this kinda thing), and your non-animated texture works as normal. Except these warnings will constantly bother you because the initial attempt at animated is what will generate the warning.

Goober5000:
This is a bit tricky because both animated and non-animated PNGs use the same extension.  So for PNGs we need to defer the filename length check until we know for sure that it is an animated image.

Fixes #5387.

~~In draft pending testing.~~  Tested by @Baezon


Goober5000:
@Baezon Could you give this a test?  It needs to allow non-animated PNG files through while still performing the length check for animated ones.

Baezon (approved):
Tested and works as expected.

PR #5475, opened 2023-07-01
prevent AI autoscaling for only a single autoscaled class

### LINKED ISSUE #5474: Bug(CTD):  adjusted AI class autoscale with a single autoscaling AI class
A minimal repro has not been constructed, but it appears that if the `use adjusted AI class autoscale` aip feature is on and all but one ai class has `$Autoscale by AI Class Index: NO` then a divide by zero exception can occur in `aicode.cpp:13118 in ai_manage_shield()`

Goober5000:
When only one AI class performs autoscaling, the math formulas which use autoscaling will become meaningless (e.g. in `ai_maybe_launch_cmeasure()`) or will cause divide-by-zero errors (e.g. in `ai_manage_shield()`).  And autoscaling is not very useful outside of the specific ways that Volition used it.  So this adds some warnings and disables single-occurrence autoscaling.

Fixes #5474.  ~~In draft status pending testing.~~

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -954,6 +954,21 @@ void ai_init()
 					 "ai autoscaling. If that's not desired then use a tbl instead of a tbm!\n"));
 		}
 
+		// autoscaling is meaningless if only one AI class has it, and will produce either ineffectual or
+		// divide-by-zero math expressions if $use adjusted AI class autoscale: is enabled
+		int num_autoscaled_classes = 0;
+		int autoscaled_class_index = -1;
+		for (int i = 0; i < Num_ai_classes; ++i) {
+			if (Ai_classes[i].ai_class_autoscale) {
+				num_autoscaled_classes++;
+				autoscaled_class_index = i;
+			}
+		}
+		if (num_autoscaled_classes == 1) {
+			Warning(LOCATION, "Autoscaling is meaningless when only one AI class has autoscaling enabled.  Turning off autoscaling for %s.", Ai_classes[autoscaled_class_index].name);
+			Ai_classes[autoscaled_class_index].ai_class_autoscale = false;
+		}
+
 		atexit(free_ai_stuff);
 
 		ai_inited = 1;


Goober5000:
Seems to work based on Woolie Wool's table posted ion #5474.

code/ai/aicode.cpp
```diff
@@ -954,6 +954,23 @@ void ai_init()
...
+			if (Ai_classes[i].ai_class_autoscale) {
+				num_autoscaled_classes++;
+				autoscaled_class_index = i;
+			}
+		}
+		if (num_autoscaled_classes == 1) {
+			Warning(LOCATION, "Autoscaling is meaningless when only one AI class has autoscaling enabled.  Turning off autoscaling for %s.", Ai_classes[autoscaled_class_index].name);
+			Ai_classes[autoscaled_class_index].ai_class_autoscale = false;
+		} else if (num_autoscaled_classes < 4) {
+			Warning(LOCATION, "This mod has only %d autoscaled AI classes.  Autoscaling is of limited utility with fewer than 4 autoscaled classes.", num_autoscaled_classes);
```
Baezon, line 971: I'm not following the thread on why >1 but <4 autoscaled classes is "of limited utility"?

Goober5000: It's subjective, but I would consider 4 the minimum number of classes for autoscaling to be useful, as you'd have 0, 33%, 67%, and 100% of whatever value is being scaled.  With 3 classes you'd have 0, 50%, and 100%, which is clunky, and with 2 classes it's just 0 and 100%.

Baezon: It is the subjectivity which makes the warning feel a bit weird. Although it is very discontinuous, that does seem to be like otherwise reasonable, expected behavior (to contrast with 1, which who even knows if that should even be interpreted as 0 or 100%, regardless of the math involved in the engine). Unless there's a good specific reason why these values should be discouraged, I'm not sure we should discourage them simply because it isn't retail's 5.

Goober5000: Eh, fair enough.

Baezon (approved):

PR #5476, opened 2023-07-02
Clear C++ Handles to LuaThread-stuff more aggressively

BMagnu:
Followup to #4701.
While that last PR apparently properly cleared the thread handle, it added three more handles: A function, a table, and a userdata.
In _some_ cases, it seems that during teardown these also didn't get properly cleared along with the thread, so this PR also manually detaches the LuaHandle from the C++ side, so that deallocation does not cause a crash on exit.


wookieejedi:
Performed additional tests where crash was previously occurring, and this PR properly fixes it.

JohnAFernandez (approved):

PR #5482, opened 2023-07-04
fix some bugs revealed by ships with long display names

Goober5000:
Several related fixes:
1. Copy the actual ship name, not the display name, to `killer_parent_name` for two reasons:
  a. Since `killer_parent_name` is used to look up the ship, it must be the actual ship name, not the display name.
  b. Copying the display name can cause a buffer overflow because display names can be longer than 31 characters.
2. The display name functionality is moved to the more appropriate location in `player_generate_death_message()`.
3. Remove `player_get_killer_weapon_name()` which only served as a proxy check for whether the weapon index was used or not.
4. Use `ship_registry_get()` rather than `ship_name_lookup()`.
5. Add comments to clarify the use of `killer_parent_name`.

Fixes a bug reported in Discord.

### DIFF
--- a/code/playerman/player.h
+++ b/code/playerman/player.h
@@ -179,7 +179,7 @@ public:
 	int				killer_objtype;							// type of object that killed player
 	int				killer_species;							// Species which killed player
 	int				killer_weapon_index;						// weapon used to kill player (if applicable)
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):
LGTM

PR #5483, opened 2023-07-07
Improved hud gauge loading safety

Baezon:
While *most* HUD gauges will check the validity of their bitmaps before rendering, a random smattering do not. This PR ensures *every* HUD render is gated behind a check for its bitmap and converts any `== -1` to `< 0` (as in debug uninitialized memory is often that giant ass negative int) and prevents any horrible bmpman assertions from rearing their ugly head.

It should be noted *all* invalid/missing bitmaps were already warned about before this point.


wookieejedi (approved):
LGTM

PR #5487, opened 2023-07-11
Add fast versions of script setting position and orientation

Baezon:
#5442 adds invalidation of collision check timings on ships who've had their position or orientation changed via script, which prevents any erroneous collision behavior, but since this involves going through all the collision pairs, if done too much like say, on dozens of ships every frame (not that I would ever do such a thing >.>) the cost can add up.

This adds 'unsafe' setPosition and setOrientation functions without this step, accepting responsibility for collision issues and making clearer the relationship to *how far* the object is moved.

### DIFF
--- a/code/scripting/api/objs/object.cpp
+++ b/code/scripting/api/objs/object.cpp
@@ -133,7 +133,7 @@ ADE_VIRTVAR(Radius, l_Object, "number", "Radius of an object", "number", "Radius
 	return ade_set_args(L, "f", objh->objp->radius);
 }
 
-ADE_VIRTVAR(Position, l_Object, "vector", "Object world position (World vector)", "vector", "World position, or null vector if handle is invalid")
+ADE_VIRTVAR(Position, l_Object, "vector", "Object world position (World vector)", "vector", "World position, or null vector if handle is invalid. Setting this value will recalculate collision culling with regards to this object.")
 {
 	object_h *objh;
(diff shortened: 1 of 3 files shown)


code/scripting/api/objs/object.cpp
```diff
@@ -194,6 +216,23 @@ ADE_VIRTVAR(Orientation, l_Object, "orientation", "Object world orientation (Wor
 	return ade_set_args(L, "o", l_Matrix.Set(matrix_h(&objh->objp->orient)));
 }
 
+ADE_FUNC(SetOrientNoCollideRetime, l_Object, "orientation", "Object world orientation (World orientation)", "vector", "Updates the orientation WITHOUT recalculating collisions. This is faster, but may cause incorrect collisions if moved too far.")
```
Goober5000, line 219: Much as I'm reluctant to make this function name even more verbose, I think `Orientation` should be spelled out, since it's spelled out in the property.

Goober5000 (commented):
I like this idea, and I will probably use it in some scripts of my own.  Although there should be an additional function that just retimes collision pairs.  (Or possibly multiple functions: retime one object, retime all objects, or retime a set of objects.  Adapt as you please.)  This will allow scripts to set all the coordinates they want, and then do a single retiming at the end.

Baezon:
Closed in favor of #5496

PR #5488, opened 2023-07-11
Fix weapon flags +override handling wrt spawn flag

Baezon:
Move `+override` to right *before* flags, instead of after allowing them to properly update spawn weapons.

Since the "Spawn xxx" flag stuffs extra related data while parsing, +override never worked properly with spawn weapons; if the new set of flags *also* had a spawn weapon it would inherit the extra spawn data and inadvertently spawn both the new and old spawn weapons. And since `+override` comes afterward, it comes too late to distinguish new or old spawn weapons, making it difficult to clear the data after the fact. Putting it beforehand allows us to clear out old data first, before parsing the new flags and stuffing the new spawn weapons.

As long as a mod targets the stable after this one (23.4??), the modder will be required to use the new positioning.

EDIT: Using `pause_parse()` and `unpause_parse()` as described below the positioning will not be affected nor any flags introduced. This now simply fixes the issue entirely in-engine

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -623,13 +623,9 @@ int weapon_info_get_index(weapon_info *wip)
 //	Parse the weapon flags.
 void parse_wi_flags(weapon_info *weaponp, flagset<Weapon::Info_Flags> preset_wi_flags)
 {
-	if (!Flags_override_old_position && optional_string("+override")) {
-		weaponp->wi_flags = preset_wi_flags;
-		weaponp->num_spawn_weapons_defined = 0;
-		required_string("$Flags:");
-	} else if (!optional_string("$Flags:"))
-		//Make sure we HAVE flags :p
-		return;
+    //Make sure we HAVE flags :p
+    if (!optional_string("$Flags:"))
+        return;
 
 	// To make sure +override doesn't overwrite previously parsed values we parse the flags into a separate flagset
     SCP_vector<SCP_string> unparsed;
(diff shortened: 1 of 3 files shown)


Goober5000 (commented):
As a preliminary matter, can we do a little bit of parsing gymnastics to keep `+override` where it is and yet still check for it before we do the flags?

One way I can think of is to call `pause_parse()`, skip to the end of the `$Flags` list, check for `+override`, and then call `unpause_parse()` to return to the original position.

Baezon:
Me and Dahblount, we're trying to brainstorm ways to do that earlier lol, although we couldn't figure out how, let me try that.

MageKing17:
> +override never worked properly with spawn weapons

Actually, it worked just fine until #4620 changed when special flags were processed; this is actually a regression.

Baezon:
Not the most elegant solution in the code, but it does work, not introducing any changes to the table is greatly preferable.

code/weapon/weapons.cpp
```diff
@@ -627,15 +627,26 @@ void parse_wi_flags(weapon_info *weaponp, flagset<Weapon::Info_Flags> preset_wi_
     if (!optional_string("$Flags:"))
         return;
 
+	// have to do this slightly out of order in order to handle spawn properly
+	// skip OVER flags, check for +override and reset num_spawn_weapons_defined
+	// otherwise if done afterward, new and old spawn weapons can get mixed up
+	pause_parse();
+	SCP_string flags;
+	stuff_string(flags, F_RAW);
```
MageKing17, line 635: Shouldn't this use `stuff_string_list()` in case "+override" isn't on a separate line?

Goober5000: I think so

Baezon: Done

Goober5000 (approved):

code/weapon/weapons.cpp
```diff
@@ -627,15 +627,26 @@ void parse_wi_flags(weapon_info *weaponp, flagset<Weapon::Info_Flags> preset_wi_
     if (!optional_string("$Flags:"))
         return;
 
+	// have to do this slightly out of order in order to handle spawn properly
```
Goober5000, line 630: Looks like this block has different indentation, which CI is complaining about

Baezon: WOW, this is scuffed. I knew the whitespace got rendered weirdly on github, but man... the whole function is a mess of tab vs space mismatch, who on EARTH does an *8 spaces tab*????

PR #5491, opened 2023-07-12
When playing the demo, don't try to reset if no campaign is active.

### LINKED ISSUE #5484: CTD caused by running the FS2 demo when another campaign is active
As reported by Renegade Paladin on Discord:
> A story in three (really four) parts.  It says that the current campaign cannot be found.  Fair enough.  Then it says that the campaign room isn't available outside the full version but I can restart the campaign.  Okay... I restart the campaign.  It gives the thing that warns that progress will be lost.  Clicking okay crashes to desktop.

Base_Profile_Screenshot_2023 07 06_-_20 37 20 22
Base_Profile_Screenshot_2023 07 06_-_20 37 33 75
Base_Profile_Screenshot_2023 07 06_-_20 37 47 39

MageKing17:
If the "Campaign Room only available in full version" popup is shown, it tries to reset `Active_campaign_index`, except that if a pilot is loaded with a missing campaign file, `Active_campaign_index` is -1. This causes an out-of-bounds array access, so let's present an alternate popup in this case (either one to select whatever the first campaign file found is, or the generic `popup_game_feature_not_in_demo()` if no campaign files are found).

Fixes #5484.

### DIFF
--- a/code/menuui/readyroom.cpp
+++ b/code/menuui/readyroom.cpp
@@ -1910,8 +1910,18 @@ void campaign_room_do_frame(float  /*frametime*/)
 
 	// If we don't have a mask, we don't have enough data to do anything with this screen.
 	if (Campaign_background_bitmap_mask == -1) {
-		{
-			//popup_game_feature_not_in_demo();
+		if ((Active_campaign_index < 0) || (Active_campaign_index >= Num_campaigns)) {
+			// Player is trying to use a regular pilot in the demo.
+			if (Num_campaigns < 1) {
+				// If there are no campaigns loaded, there's really nothing left to do.
+				popup_game_feature_not_in_demo();
+				return;
+			}
+			int select_default = popup(PF_USE_AFFIRMATIVE_ICON|PF_BODY_BIG, 2, "Exit", "Select first campaign", "Campaign Room only available in full version. However, you may select the first campaign found.");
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
Yeah, this looks like the way to solve it.

The one loose end here is what happens if there really are no campaigns.  The popup for "The currently active campaign cannot be found" is a blocking popup to ensure that there is always a valid campaign.  But the `(Num_campaigns < 1)` code path here could allow the player to click ok and end up in the main hall with no valid campaign.

I suppose a case could be made to open a separate issue for that.

MageKing17:
> The one loose end here is what happens if there really are no campaigns. The popup for "The currently active campaign cannot be found" is a blocking popup to ensure that there is always a valid campaign. But the `(Num_campaigns < 1)` code path here could allow the player to click ok and end up in the main hall with no valid campaign.

Actually, this code path *doesn't* allow the player to end up in the main hall with no valid campaign, because it doesn't set `Campaign_room_no_campaigns` (like the non-demo campaign room does when returning to the main hall with no campaigns), so the "The currently active campaign cannot be found" popup should trigger again when returning to the main hall.

Goober5000:
Oh, is that so?  Ok then.

Goober5000 (approved):

PR #5492, opened 2023-07-12
Ignore voice recognition events while game is paused.

MageKing17:
If you pause the game, say something into the microphone, then unpause, it can abruptly perform an action based on voice recognition; this is probably counter to player expectations. This adds a `gameseq_get_state() != GS_STATE_GAME_PAUSED` check to the existing conditional.

### DIFF
--- a/code/sound/voicerec.cpp
+++ b/code/sound/voicerec.cpp
@@ -88,7 +88,7 @@ namespace
 		switch (e.syswm.msg->msg.win.msg)
 		{
 		case WM_RECOEVENT:
-			if (Game_mode & GM_IN_MISSION && Cmdline_voice_recognition && gameseq_get_state() != GS_STATE_GAME_PAUSED)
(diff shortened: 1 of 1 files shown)


Goober5000 (changes requested):
Good catch.  I think this should also check `popup_active()` because pressing Escape mid-mission technically doesn't count as pausing.

Goober5000 (changes requested):
Good catch.  I think this should also check `popup_active()` because pressing Escape mid-mission technically doesn't count as pausing.

MageKing17:
Threw in a `popupdead_is_active()` check as well, because regular controls also stop being processed while that's true.

Goober5000 (approved):
I think `popup_active()` encompasses `popupdead_is_active()`, but even if it's redundant it should be fine.

MageKing17:
> I think `popup_active()` encompasses `popupdead_is_active()`, but even if it's redundant it should be fine.

It does not; `popupdead_is_active()` sidesteps the normal popup code, so `popup_active()` will return false when it returns true.

Goober5000:
OIC.  Ok then.

PR #5493, opened 2023-07-12
Make `lock_fov` settable for aspect/javelin missiles.

MageKing17:
The code has always checked a per-weapon-class value called `lock_fov` when comparing the dot product of a ship to its target to see if it can lock on to it, however the weapon table hasn't ever included a way to set this value; it's always the default value of 0.85. This just adds "+Lock FOV:" entries to aspect/javelin homing missile parsing after "+Seeker Strength:", which will set `lock_fov` as long as the supplied input is between 0 and 360 degrees.

(Note to self: don't forget to update the wiki whenever this gets merged.)

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -1551,6 +1551,20 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 					wip->seeker_strength = 2.0f;
 			}
 
+			if (optional_string("+Lock FOV:"))
+			{
+				float temp;
+				stuff_float(&temp);
+				if (temp > 360 || temp < 0)
+				{
+					error_display(0, "Lock FOV values should be between 0 and 360 degrees; ignoring value of %f for missile \'%s\'.", temp, wip->name);
+				}
+				else
+				{
+					wip->lock_fov = cosf(fl_radians(temp * 0.5f));
+				}
+			}
+
 			if (optional_string("+Target Lead Scaler:"))
 			{
 				stuff_float(&wip->target_lead_scaler);


code/weapon/weapons.cpp
```diff
@@ -1478,6 +1478,21 @@ int parse_weapon(int subtype, bool replace, const char *filename)
...
 			}
 
+			if (optional_string("+Lock FOV:"))
+			{
+				float temp;
+				//actually lock dot, but whatever
+				stuff_float(&temp);
+				if (temp > 1 || temp < -1)
+				{
+					error_display(0, "Lock FOV values with a magnitude higher than 1 make no sense; ignoring value of %d for missile \'%s\'.", temp, wip->name);
```
Goober5000, line 1488: The `%d` should be `%f`

code/weapon/weapons.cpp
```diff
@@ -1551,6 +1566,21 @@ int parse_weapon(int subtype, bool replace, const char *filename)
...
```
Goober5000, line 1576: Here too

Baezon (changes requested):
Excellent idea, I was always surprised to see that `lock_fov` was hardcoded, although I never needed to change it myself before.

This should however do some conversion and expose the value to modders as degrees isntead of a dot product to match similar angle-based options.

Goober5000 (changes requested):
As a user-friendliness enhancement, methinks `lock_fov` should be specified in the table in degrees, then computed on parse.

Baezon:
You owe me a soda

MageKing17:
> This should however do some conversion and expose the value to modders as degrees instead of a dot product to match similar angle-based options.

Can you name an example of such a "similar angle-based option"?

Baezon:
Basically any time angles are needed. Missile view cone, and turret fovs are ones that come to mind that specifically turn them into dot products, but table degrees to internal radians (although not explicitly the case here) is even more common which you can find many examples of just by searching `fl_radians` in table parsing areas.

MageKing17:
I saw `+View Cone:` almost immediately after I posted that; I also realized that I misread the code and `fov` is used for homing weapons looking for a new target; `lock_fov` is solely used by aspect/javelin missiles, so I'll be correcting both of those issues.

Baezon (approved):

Goober5000 (commented):
One more question.  Is cutting the angle in half correct here?

MageKing17:
> One more question. Is cutting the angle in half correct here?

Yes; otherwise, we're only measuring half of the field of view (center to side instead of side to side).

Baezon:
Tmk, various fields waffle on center-to-edge vs edge-to-edge, so it's fine as long as it's clear to the user which it is.

Goober5000 (approved):

MageKing17:
Other FOV fields that get converted to a normalized dot product (like the aforementioned "+View Cone:", but also "$Turret Base FOV:" or "$Turret Barrel FOV:") also divide the input in half, so this is (AFAICT) the most consistent way to do it.

PR #5496, opened 2023-07-17
Improve performance of collision retiming

Baezon:
Follow-up to #5487 

#5442 adds invalidation of collision check timings on ships who've had their position or orientation changed via script, which prevents any erroneous collision behavior, but since this involves going through all the collision pairs, if done too much like say, on dozens of ships every frame (not that I would ever do such a thing >.>) the cost can add up.

Instead of adding extra script bells and whistles to skip this step in 5487, the cost of fixing this invalidation can be greatly improved. Instead of immediately retiming all or even some pairs, any operations will instead set a global flag, then once per frame directly before collision go through all the pairs to retime them, *and* cull any invalid pairs.

Simply going through the pairs is often the bigger cost than the actually re-triggered collisions! ~~For this reason, I have removed the single object version of the collision retiming, since although it makes logical sense, it actually does not actually save much actual performance.~~ (After review below, an object flag is added so that only the specifically invalidated objects are recalculated during the single pass so as to avoid resetting potentially still correct timestamps) So by doing once per frame, we avoid multiple calls in one frame from adding up, and by culling we also make any subsequent triggers hopefully quicker as well. 

After the stable, there is still the possibility of adding a scripted object flag to avoid triggering this global flag for further optimization, but I think this suffices for the stable.

### DIFF
--- a/code/autopilot/autopilot.cpp
+++ b/code/autopilot/autopilot.cpp
@@ -1023,7 +1023,7 @@ void nav_warp(bool prewarp=false)
 
 				// retime collision pairs
 				if (Objects[Ships[i].objnum].flags[Object::Object_Flags::Collides])
-					obj_collide_cache_stale();
+					obj_collide_obj_cache_stale(&Objects[Ships[i].objnum]);
 		}
 	}
 }
--- a/code/autopilot/autopilot.cpp
+++ b/code/autopilot/autopilot.cpp
@@ -1023,7 +1023,7 @@ void nav_warp(bool prewarp=false)
 
 				// retime collision pairs
 				if (Objects[Ships[i].objnum].flags[Object::Object_Flags::Collides])
-					obj_collide_retime_cached_pairs(&Objects[Ships[i].objnum]);
+					obj_collide_cache_stale();
 		}
 	}
 }
--- a/code/object/objcollide.cpp
+++ b/code/object/objcollide.cpp
@@ -25,7 +25,6 @@
 int Num_pairs = 0;
 int Num_pairs_checked = 0;
 
-SCP_vector<object*> Stale_collision_time_objects;
 SCP_vector<int> Collision_sort_list;
 
 class collider_pair
@@ -615,7 +614,6 @@ void obj_collide_retime_cached_pairs()
 		if (pair.signature_a != pair.a->signature || pair.signature_b != pair.b->signature) {
 			it = Collision_cached_pairs.erase(it);
 		} else {			
-			mprintf(("\n%d: %f and %f stale", Framecount, pair.a->radius, pair.b->radius));
 			pair.next_check_time = timestamp(0);
 			it++;
 		}
@@ -624,7 +622,7 @@ void obj_collide_retime_cached_pairs()
 
 void obj_collide_cache_stale()
 {
-	//Collision_cache_stale = true;
+	Collision_cache_stale = true;
 }
 
 //local helper functions only used in objcollide.cpp
@@ -886,14 +884,11 @@ void obj_collide_pair(object *A, object *B)
         collision_info->next_check_time = timestamp(0);
     }
 
-	mprintf(("\n%d: %f and %f checking", Framecount, A->radius, B->radius));
-
     if ( valid ) {
         // if this signature is valid, make the necessary checks to see if we need to collide check
         if ( collision_info->next_check_time == -1 ) {
             return;
         } else if ( !timestamp_elapsed(collision_info->next_check_time) ) {
-			mprintf(("\n%d: %f and %f skipped", Framecount, A->radius, B->radius));
 			return;
         }
     } else {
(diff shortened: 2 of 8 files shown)


code/autopilot/autopilot.cpp
```diff
@@ -1023,7 +1023,7 @@ void nav_warp(bool prewarp=false)
 
 				// retime collision pairs
 				if (Objects[Ships[i].objnum].flags[Object::Object_Flags::Collides])
-					obj_collide_retime_cached_pairs(&Objects[Ships[i].objnum]);
+					obj_collide_cache_stale();
```
Goober5000, line 1026: What about marking a single object stale, using an object flag in the same way as `Docked_already_handled` for example?  Then the code could still limit itself to one iteration through the list per frame, but it would check the stale flag in addition to any other checks.  This stale flag would then be cleared after its timestamp is reset.

Goober5000 (commented):
Interesting approach.  I don't like the idea of a global stale flag, because it has the effect of disabling the collision cache in any mission where *any* object becomes stale a lot.  Benchmarks do speak louder than words though.

A few comments.

code/object/objcollide.cpp
```diff
@@ -607,23 +608,21 @@ void obj_reset_colliders()
 
 void obj_collide_retime_cached_pairs()
 {
-	for ( auto& pair : Collision_cached_pairs ) {
-		pair.second.next_check_time = timestamp(0);
+	auto it = Collision_cached_pairs.begin();
+	while (it != Collision_cached_pairs.end()) {
+		auto &pair = it->second;
+		if (pair.signature_a != pair.a->signature || pair.signature_b != pair.b->signature) {
```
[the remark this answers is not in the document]
Baezon: This is true, they are not directly related, but it's not worth doing on it's own, and we're here anyway, so it might as well be done when they must be iterated over regardless. And they only "should" be removed in the first place, because this very loop is slowed down by them, invalid pairs does not really impinge on normal collision evaluation.

Baezon:
Also added a tracing scope for it to keep an eye on it for the future.

code/object/objcollide.cpp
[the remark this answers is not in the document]
Goober5000: Ok.

code/object/objcollide.cpp
```diff
@@ -607,23 +608,25 @@ void obj_reset_colliders()
 
 void obj_collide_retime_cached_pairs()
```
Goober5000, line 609: This function should be renamed - maybe `obj_collide_retime_stale_pairs()`.

Goober5000 (approved):
Looks good now, but one more thing.

code/object/objcollide.cpp
```diff
@@ -607,23 +608,25 @@ void obj_reset_colliders()
...
-		pair.second.next_check_time = timestamp(0);
+	TRACE_SCOPE(tracing::RetimeCollisionCache);
+
+	auto it = Collision_cached_pairs.begin();
+	while (it != Collision_cached_pairs.end()) {
+		auto &pair = it->second;
+		if (pair.signature_a != pair.a->signature || pair.signature_b != pair.b->signature) {
+			it = Collision_cached_pairs.erase(it);
+		} else {
+			if (pair.a->flags[Object::Object_Flags::Collision_cache_stale] || pair.b->flags[Object::Object_Flags::Collision_cache_stale])
```
Goober5000, line 619: The flag should be cleared on the stale object.

Goober5000 (changes requested):
Actually another thing.

code/object/objcollide.cpp
Baezon: Whoops!

Baezon: Eugh, crap now we're back to the issues with object flags, it's difficult to know while we're iterating when we're done with an object and to clear its flag.

Baezon: It's not the prettiest but may as well bring back the list of stale cache objects.

Goober5000: Oh, right, because the stale object may have more than one collision pair.  That definitely makes it more complicated.

code/object/objcollide.cpp
```diff
@@ -44,7 +44,7 @@ class collider_pair
 	{}
 };
 
-static bool Collision_cache_stale = false;
+static SCP_vector<object*> Collision_cache_stale_objects;
```
Goober5000, line 47: I wonder if this should be a set rather than a vector.  It looks like the vector will still work, but it might store more than one copy of the same pointer if the same object was invalidated more than once in the same frame.

Baezon: Ah, I didn't know C++ had sets! Good point

Goober5000 (approved):

PR #5497, opened 2023-07-17
Add a warning for cases when we couldn't load a trail bitmap

The-E:
Based on [this question by Kitsune on discord](https://discord.com/channels/223511295431933953/223511363807346691/1130507954873958501), I did some digging and it turns out if a trail bitmap is missing, the engine never warns about it; we silently swallow this on release builds and assert unhelpfully on debug and fastdbg. we shouldn't do that.

### DIFF
--- a/code/weapon/trails.cpp
+++ b/code/weapon/trails.cpp
@@ -131,7 +131,6 @@ void trail_render( trail * trailp )
 	trail_info *ti	= &trailp->info;
 
 	if (ti->texture.bitmap_id <= 0) {
-		Warning(LOCATION, "Trail bitmap %s could not be loaded.", ti->texture.filename);
 		return;
 	}
 
--- a/code/weapon/trails.cpp
+++ b/code/weapon/trails.cpp
@@ -129,6 +129,12 @@ void trail_render( trail * trailp )
 		return;
 
 	trail_info *ti	= &trailp->info;
+
+	if (ti->texture.bitmap_id <= 0) {
+		Warning(LOCATION, "Trail bitmap %s could not be loaded.", ti->texture.filename);
+		return;
+	}
+
 	auto batchp = batching_find_batch(ti->texture.bitmap_id, batch_info::FLAT_EMISSIVE);
 
 	if (trailp->single_segment) {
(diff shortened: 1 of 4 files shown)


code/weapon/weapons.cpp
```diff
@@ -1982,6 +1982,10 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		if ( required_string("+Bitmap:") ) {
 			stuff_string(fname, F_NAME, NAME_LENGTH);
 			generic_bitmap_init(&ti->texture, fname);
+
+			if (ti->texture.bitmap_id == -1) {
+				Warning(LOCATION, "Trail bitmap %s could not be loaded. Trail will not be rendered.", fname);
+			}
```
Baezon, line 1988: Uh, is this in the right place? `generic_bitmap_init()` always sets `bitmap_id` to -1, so I assume this warning should be after the load is actually attempted.

The-E: oops, changed

Baezon (changes requested):
Now that we correctly warn for missing bitmaps, the trail rendering should probably explicitly check for a missing bitmap too and early return, so even on debug we don't run head long into an assertion.

Baezon:
Oh, I see. Personally I would prefer `trail_render()` early return, and the batching functions continue to assert. The assertion did its job alerting us the trails weren't being checked, but if we remove the assertion we wouldn't be alerted the same way in the future should some other graphics need `batching_add_quad` and make the same mistake.

code/weapon/trails.cpp
```diff
@@ -129,6 +129,12 @@ void trail_render( trail * trailp )
 		return;
 
 	trail_info *ti	= &trailp->info;
+
+	if (ti->texture.bitmap_id <= 0) {
+		Warning(LOCATION, "Trail bitmap %s could not be loaded.", ti->texture.filename);
```
Baezon, line 134: One last thing, if you'll permit me, the warning in `weapon_load_bitmaps` *should* have already warned about this, so we shouldn't warn again here, especially since in a render function it would be triggering every frame which is always very annoying.

The-E: Good? Honestly, if you clicked through the first warning, getting annoyed at the others is on you.

Baezon: There's no reason to repeatedly bother the user when they've knowingly accepted and ignored the issue. There are a host of legitimate reasons why someone may understand a warning and proceed anyway. They may be attempting to fix a different issue. They may not be the actual modder, but instead someone else trying to verify some other aspect of the mod.

The-E: fair enough

Baezon (approved):

PR #5501, opened 2023-07-19
Explicitly zero out persona bitfields

naomimyselfandi:
Personae have a flag and a species bitfield. They're built up during parsing, but never explicitly zeroed out. This leads to ships occasionally getting an incorrect persona (especially common with support ships, which never have an explicit persona).

Additionally, this commit addresses an issue with the UES Aranyaka (and potentially other support ships belonging to species without a support persona). I'm honestly not sure why they were ever able to speak; however, with this change, they'll never be able to. As such, this commit introduces a "default support persona" to retain the desired behavior.

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -141,7 +141,7 @@ const char *Persona_type_names[MAX_PERSONA_TYPES] =
 //XSTR:ON
 };
 
-int Default_command_persona;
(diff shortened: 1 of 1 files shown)


TRBlount (approved):
Looks good.

JohnAFernandez (approved):

naomimyselfandi:
> Looks good.

Thank you!

PR #5504, opened 2023-07-21
Fix for countermeasures with a single segment trail

Baezon:
Single segment trail optimization relies on the fact that the object has a constant velocity, so when countermeasures go and *change* that velocity after creating the weapon (and telling the trail a *wrong velocity* beforehand), it gets ugly. `CONST_VEL` doesn't mean what it used to these days... but I'm confident this sort of thing is extremely rare.


JohnAFernandez (approved):
Agreed.  One ill hack deserves another.

PR #5506, opened 2023-07-23
Restore (mostly-)retail debrief failure counting behavior.

MageKing17:
Back in 2008 (in commit 2c45dbbb5f77404e4d1833369cd2f40594731a81), some retail code incrementing `Player->failures_this_session` was removed, seemingly by accident. A month later, a bug related to this was fixed (in commit 28a408deee3496b8f6d81768778c3ff1f2397515) in a different way. Since this fix is rendered redundant by the restoration of the original behavior, I've removed it and put back in the old behavior. This is only "mostly-retail" because the retail code didn't check `Game_mode & GM_CAMPAIGN_MODE`.

This restores the old behavior of allowing a player to skip a mission without having to die (and see the death popup, which doesn't happen in some missions that show a debriefing if the player dies). Since the logic is so straightforward (and other "mission failed" checks in the debriefing code use the same or at least very similar logic), it *should* work with any existing mods... but that "should" is so wide, and this change could affect so many things, that I've got it in draft form to start with because more research is needed before fiddling with a foundational system like the debriefing screen.

### DIFF
--- a/code/mission/missioncampaign.cpp
+++ b/code/mission/missioncampaign.cpp
@@ -1114,7 +1114,6 @@ void mission_campaign_mission_over(bool do_next_mission)
 
 	// Goober5000 - player-persistent variables are handled when the mission is
 	// over, not necessarily when the mission is accepted
-	Player->failures_this_session = 0;
 
 	// update campaign.mission stats (used to allow backout inRedAlert)
 	// .. but we don't do this if we are inside of the prev/current loop hack
--- a/code/missionui/missiondebrief.cpp
+++ b/code/missionui/missiondebrief.cpp
@@ -2015,6 +2015,14 @@ void debrief_init(bool API_Access)
 		debrief_init_music();
 	}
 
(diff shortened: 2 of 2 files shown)


JohnAFernandez:
I'm thinking that checking the campaign mode keeps the second check from relying on bad data.

Goober5000 (approved):
This is a good fix and I'm satisfied that it should work as implemented.  The retry count is handled in both debriefing and the death popup so it should work in almost all cases.  Also, since the check is based on the current mission while the player is in campaign mode, it should be robust against any type of campaign.  (Even the looping campaigns in JAD and the mission hub will move to at least one other mission before coming back to the same mission.)

The one situation not handled here is if there is a mission failure where the player is dumped out to the main hall.  But I think that is outside the scope of what the 5x failure popup should try to handle.

Goober5000:
Can this be taken out of Draft?  I think this is a good fix, and I also need it to be merged as prerequisite for a mission-skip mitigation feature I'm working on.

PR #5512, opened 2023-07-25
ai-disable-ship-tactical and ai-disarm-ship-tactical

Goober5000:
Add "tactical" versions of disable and disarm goals which only perform the specific task without implying anything else.  The target ship is not protected and attack goals are not cleared from other ships.

There is also a new AI profiles flag to make this the default hud squad message command instead of the classic disable/disarm.

~~In draft status pending tests.~~  Tested by @wookieejedi


wookieejedi (approved):
Tested and works as expected and code is logical and looks good.

PR #5516, opened 2023-07-26
POF version fix

Goober5000:
Modelread.cpp supports POF versions as low as 1900, but early model versions were not loaded properly because they relied on preprocessor directives rather than version checks.  This changes the POF loading to correctly check version numbers and chunk IDs.

Fixes crashes when attempting to load a FS1 model.


Baezon (approved):

Baezon:
Although, its sort of a fix, its also sort of a feature... So atm, I'm leaning towards after the stable

Goober5000:
Fair enough

PR #5518, opened 2023-07-26
Add feature to always display the damage gauge

### LINKED ISSUE #5472: Feature Request: Permanently Enable Damage Gauge
The current damage gauge only appears once a subsystem or hull integrity drops below a certain percentage. I would like to be able to permanently enable this gauge via some command/function in a -hdg table. The reason is for use in modeled cockpits. This way the information is always rendered to the cockpit canvas and is present and available to the player.

MjnMixael:
Implements #5472 with a simple boolean flag for the damage gauge. If true then all damage values will always be drawn.

A note on hud.cpp line 2024: Changing the value here from 1000.0f to 1.0f prevents the subsystem damage text from flashing at mission start if Always Display is true but still ensures that the text flashes without the feature if the damage is less than 1.0f where 1.0f == 100% hull integrity.

Fixes #5472

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -2021,7 +2021,7 @@ void hud_damage_popup_init()
 	Damage_flash_timer =	1;
 
 	for ( i = 0; i < SUBSYSTEM_MAX; i++ ) {
-		Pl_hud_subsys_info[i].last_str = 1000.0f;
+		Pl_hud_subsys_info[i].last_str = 1.0f;
 		Pl_hud_subsys_info[i].flash_duration_timestamp = 1;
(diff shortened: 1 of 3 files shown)


code/hud/hud.cpp
```diff
@@ -2021,7 +2021,7 @@ void hud_damage_popup_init()
 	Damage_flash_timer =	1;
 
 	for ( i = 0; i < SUBSYSTEM_MAX; i++ ) {
-		Pl_hud_subsys_info[i].last_str = 1000.0f;
```
Baezon, line 2024: Do you know why it was initialized to 1000? Seems like a strange value to initialize 'last strength' to...

Baezon (approved):

PR #5519, opened 2023-07-27
Remove new lines from default generated mission fields

MjnMixael:
Simply removes the newline characters from the default mission Notes and Description. SCPUI can be set to actually display these values but requires additional code to remove the newlines because they mess with the default SCPUI mission grid view. It's minor to deal with, but I'm also thinking it's pretty minor to just remove them by default.


BMagnu (approved):
Fair enough.

PR #5520, opened 2023-07-27
SEXP parsing cleanup and fixes

Goober5000:
This is a cleanup and upgrade of the SEXP parsing mechanism that makes the whole thing a lot more robust.  Error handling has been upgraded, several cases that previously caused a crash will fail gracefully, and some cases that previously caused an error can now be recovered with a straightforward "Attempt to load the mission failed".  The case that caused parsing to fail on Black Horizon mission 7, "Alone in the Void" -- a message name that started with an `@` character -- now works successfully.  The code has also been refactored and cleaned up a bit.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -33364,7 +33364,12 @@ const char *CTEXT(int n)
 
 			if (!(Sexp_nodes[arg_n].flags & SNF_CHECKED_ARG_FOR_VAR))
 			{
-				Sexp_nodes[arg_n].cached_variable_index = check_sexp_node_text_for_sexp_variable(text);
+				// nodes that have an officially formatted variable will store the variable index in the token
+				if (Sexp_nodes[arg_n].type & SEXP_FLAG_VARIABLE)
+					Sexp_nodes[arg_n].cached_variable_index = atoi(text);
+				else
+					Sexp_nodes[arg_n].cached_variable_index = check_sexp_node_text_for_sexp_variable(text);
+
 				Sexp_nodes[arg_n].flags |= SNF_CHECKED_ARG_FOR_VAR;
 			}
 
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -4156,7 +4156,7 @@ bool get_unformatted_sexp_variable_name(char *unformatted, const char *formatted
 		return false;
 
 	strncpy(unformatted, intermediate, n);
-	unformatted[n] = '\0';
+	unformatted[n] = 0;
 	return true;
 }
 
@@ -4298,23 +4298,29 @@ void localize_sexp(int text_node, int id_node)
 
 // Advance to and consume the closing parenthesis of a sexp, in case of a parse error.
 // If a closing parenthesis is not found, this advances to the end of the string.
-void skip_sexp()
+void skip_sexp(bool within_quotes = false)
 {
 	int level = 1;
 
 	while (*Mp != '\0')
 	{
-		if (*Mp == ')')
+		if (*Mp == '\"')
+			within_quotes = !within_quotes;
+
+		if (!within_quotes)
 		{
-			level--;
-			if (level == 0)
+			if (*Mp == ')')
 			{
-				Mp++;
-				return;
+				level--;
+				if (level == 0)
+				{
+					Mp++;
+					return;
+				}
 			}
+			else if (*Mp == '(')
+				level++;
 		}
-		else if (*Mp == '(')
-			level++;
 
 		Mp++;
 	}
@@ -4361,7 +4367,7 @@ int get_sexp()
 			// was closing quote not found?
 			if (*(Mp + 1 + len) != '\"') {
 				error_display(0, "Unexpected end of quoted string embedded in sexp!");
-				skip_sexp();
+				skip_sexp(true);
 				return Locked_sexp_false;
 			}
 			// bump past closing quote
@@ -4401,7 +4407,8 @@ int get_sexp()
 				}
 				// bad format
 				if (*Mp == '(') {
-					error_display(1, "Mismatched parentheses while parsing SEXP!  Current parse position:\n%s", Mp);
+					char buf[512];
+					error_display(1, "Mismatched parentheses while parsing SEXP!  Current parse position:\n%s", three_dot_truncate(buf, Mp, 512));
 					return Locked_sexp_false;
 				}
 				Mp++;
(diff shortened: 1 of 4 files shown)


code/parse/sexp.cpp
```diff
@@ -36649,30 +36729,30 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"arrays and pointers.\r\n\r\nPlease note that only numeric variables are supported.  Any "
 		"attempt to access a string variable will result in a value of SEXP_NAN_FOREVER being returned.\r\n\r\n"
 		"Takes 1 argument...\r\n"
-		"\t1:\tIndex of variable, from 0 to MAX_SEXP_VARIABLES - 1." },
+		"\t1:\tIndex of variable, from 0 to " SCP_TOKEN_TO_STR(MAX_SEXP_VARIABLES) " - 1." },
```
MageKing17, line 36732: I realize this would require adding, say, a `MAX_SEXP_VARIABLE_INDEX` `#define`... but it would make much more sense to just display `249` instead of `250 - 1`.

MageKing17 (changes requested):
Couple of things struck me as iffy, but on the whole this cleanup seems like a good idea.

code/parse/sexp.cpp
```diff
@@ -4261,6 +4296,30 @@ void localize_sexp(int text_node, int id_node)
...
+		if (*Mp == ')')
+		{
+			level--;
+			if (level == 0)
+			{
+				Mp++;
+				return;
+			}
+		}
+		else if (*Mp == '(')
```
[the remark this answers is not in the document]
Goober5000: Open parentheses within quoted strings should be fine.  It's only in the middle of variables, containers, operators, or numbers that they would cause problems.

code/parse/sexp.cpp
```diff
@@ -4298,173 +4356,189 @@ int get_sexp()
...
 
-				strncpy(token, Mp + 1, len);
+				strncpy(token, startp + 1, len);
 				token[len] = 0;
 				node = alloc_sexp(token, SEXP_ATOM, SEXP_ATOM_STRING, -1, -1);
 			}
-
-			// bump past closing \" by 1 char
-			Mp += (len + 2);
-
```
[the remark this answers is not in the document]
Goober5000: Yeah, I went through several iterations before I arrived at this design.  The main issue is that `Mp` needs to point to the following token when `get_sexp()` returns, otherwise you'll have an endless loop.  So I started adding "bumps" before each return, but that started getting ugly because there was a lot of repeated code and yet not all the code bumped the same amount.  And container tokens had to be bumped differently because there was an additional `get_sexp()` in the middle of it.  Rather than introduce a whole lot of fiddly and brittle code, it made more sense to just advance `Mp` as it read the token, while keeping track of where the token started.

code/parse/sexp.cpp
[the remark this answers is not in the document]
Goober5000: Oh wait, I see what you mean.

code/parse/sexp.cpp
```diff
@@ -4298,173 +4356,189 @@ int get_sexp()
...
+			size_t len = 0;
+			while (*Mp != ')' && !is_white_space(*Mp)) {
+				// end of string or end of file
+				if (*Mp == '\0') {
+					error_display(0, "Unexpected end of sexp!");
+					return Locked_sexp_false;
+				}
+				// bad format
+				if (*Mp == '(') {
+					error_display(1, "Mismatched parentheses while parsing SEXP!  Current parse position:\n%s", Mp);
```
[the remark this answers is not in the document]
Goober5000: I didn't realize it dumped the entire un-truncated text to the debug log, oof.  Yeah I'll use a `[...]`.

code/parse/sexp.cpp
Goober5000: Well there are a lot of uses of `SCP_TOKEN_TO_STR`.  It would be necessary to add a special `#define` for each different token.

code/parse/sexp.cpp
```diff
@@ -4130,21 +4130,61 @@ int check_sexp_syntax(int node, int return_type, int recursive, int *bad_node, s
...
+	constexpr size_t var_token_size = 2 * (TOKEN_LENGTH - 1) + 4;		// @variable_token[contents_token]\0
+	char variable_token[var_token_size];
+	char variable_name[NAME_LENGTH];
+
+	// token is too long?
+	if (n_chars >= var_token_size)
+		return -1;
+
+	strncpy(variable_token, str, n_chars);
+	variable_token[n_chars] = 0;
```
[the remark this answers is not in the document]
Goober5000: Ok.

code/parse/sexp.cpp
MageKing17: Yes, but by my count that's exactly two tokens, which isn't a lot of overhead compared to the user not having to do math to understand the help text.

code/parse/sexp.cpp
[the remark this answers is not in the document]
MageKing17: If it's solely for consistency with the other blocks, then fair enough.

code/parse/sexp.cpp
Goober5000: Ok.

Goober5000:
Ok, pushed some changes.

code/parse/sexp.cpp
```diff
@@ -4361,7 +4367,7 @@ int get_sexp()
 			// was closing quote not found?
 			if (*(Mp + 1 + len) != '\"') {
 				error_display(0, "Unexpected end of quoted string embedded in sexp!");
-				skip_sexp();
+				skip_sexp(true);
```
MageKing17, line 4370: Logically-speaking, this should be `skip_sexp(false)`, as the closing quote not being found means that it will never turn off `within_quotes` and thus skip the entire rest of the file. Unless stopping all parsing of the file is intentional? In which case, I feel like calling `skip_sexp()` here at all gives the wrong impression.

MageKing17 (changes requested):

code/parse/sexp.cpp
Goober5000: Skipping the rest of the file or string is indeed the intention.  Since the quote never ends, the parser should go as far as it can.  And since that is essentially skipping the rest of the sexp, `skip_sexp()` works here IMO.  The alternative is to use `Mp += (len + 1);` which admittedly is more efficient but I think is less clear.

MageKing17: I think `skip_sexp()` is less clear because it's not immediately obvious that the entire rest of the file will be skipped.

Goober5000: It's consistent with the pattern for every other error check in this section.  But I can add a comment; perhaps that will help.

MageKing17: That works.

MageKing17 (approved):

Issue #4533, opened 2022-08-12
FR: Lua hooks into hardcoded UI controls


MjnMixael:
Specifically I'd like to be able to capture the ESC key in GS_STATE_SCRIPTING and handle it in the Lua rather than it always going back to the previous gamestate. But I'm sure there could be more potential uses if this feature was broadened and generalized. Here's my discussion with @BMagnu regarding the issue.

> MjnMixael — 06/06/2022
> My use case here is to completely stop FSO from handling ESC in during the scripting game state and handle it in Lua so that I can choose whether to leave the game state or perform something else.
> BMagnu — 06/06/2022
> You'd need to use the action to which ESC is usually bound. But:
> I don't think that is currently possible, because the scripting state does not actually use "the controls" to check whether it has gotten the ESC.
> Rather, it uses a hardcoded, non-rebindable check to the (system) keyboard buffer (similar to, say, pressing F2 in mission for the options) to see whether it should quit.
> The new system is primarily for (rebindable) controls, replacing the on-action hook with more detail.
> Your usecase requires the non-rebindable On Key Pressed hook to be overridable.

BMagnu:
My idea for this, at least for this use case, would be to make a very low-level hook to ``key_inkey()``.
The advantage is that it would be able to catch and override very low-level events, like pressing F2 to go to the options menu.
But that has a few disadvantages:
1. It's highly context sensitive. Some things, like the F2 button, are hardcoded pretty deep and will happen really always except in some few gamestates. Then some other things, like navigating popups, is similarly hardcoded, but only in the popup part, and even then it can depend on the actual popup itself. Navigating this context, and what button does what when is going to be very hard for scripters and would undoubtedly require some knowledge of the C++ code.
2. It's going to be called often. At least once per frame, but possibly more often. Every time the code will query keypresses, effectively. This _could_ in some cases, cause confusion for scripters.

Currently, there are already On Key Pressed and On Key Released. These are, AFAICT, called by the SDL callbacks, and would be even more low level, however due to their nature it is very hard to override them (it'd need a buffer of overridden keypresses that then do not need to be removed on release and stuff), as well as not corresponding to the engine actually _checking_ for keypresses, which forces the scripter to ignore keypresses based on when they occur, and not when they are queried (which makes things hard that interact with holding a key).

Ideas for a better solution, or does this sound suitable?

MjnMixael:
[...]

PR #5522, opened 2023-07-28
Custom lua action for mainhall esc keypress

MjnMixael:
BtA is modernizing the UI experience. PilotSelect is becoming a title screen that includes player selection among other options. As part of this, we're removing the Barracks. Players that want to switch pilots must return to the title screen.

In order to make this seamless I want a way to force the ESC keypress during the mainhall not always be "quit game", but instead return to the game title screen.

Unless #4533 is going to be completed soon, this small table feature allows customizing the last non-customizable aspect of each mainhall with a custom script action, similar the "Script" option for door clicks.

### DIFF
--- a/code/menuui/mainhallmenu.cpp
+++ b/code/menuui/mainhallmenu.cpp
@@ -970,8 +970,20 @@ void main_hall_do(float frametime)
 				case ESC_PRESSED:
 					// if there is a help overlay active, then don't quit the game - just kill the overlay
 					if (!help_overlay_active(Main_hall_overlay_id)) {
-						gamesnd_play_iface(InterfaceSounds::IFACE_MOUSE_CLICK);
-						main_hall_exit_game();
+
+						// if there is no custom script action then exit
+						if (Main_hall->esc_action.empty()) {
+							gamesnd_play_iface(InterfaceSounds::IFACE_MOUSE_CLICK);
+							main_hall_exit_game();
+						} else {
+							const char* lua = Main_hall->esc_action.c_str();
+							bool success = Script_system.EvalString(lua, lua);
+							if (!success)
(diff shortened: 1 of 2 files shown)


BMagnu (commented):
Okay,  I want to put how to deal with this up for discussion:
On the one hand, it's been convention to _not_ have Lua code like this as part of the tables, hence why the new IO hooks are done like SEXPs and not built into the controlsconfigdefault table.
On the other hand, the mainhall is the one thing that already breaks that convention by defining custom doors like this.
I am leaning towards making this an explicit, overridable hook in the global_hooks.cpp, but this _might_ be fine as well given the precedent set by the doors?

MjnMixael:
> Okay, I want to put how to deal with this up for discussion: On the one hand, it's been convention to _not_ have Lua code like this as part of the tables, hence why the new IO hooks are done like SEXPs and not built into the controlsconfigdefault table. On the other hand, the mainhall is the one thing that already breaks that convention by defining custom doors like this. I am leaning towards making this an explicit, overridable hook in the global_hooks.cpp, but this _might_ be fine as well given the precedent set by the doors?

In general I'd agree with you. I certainly wouldn't do this anywhere else but the mainhall precisely because the precedent is already set by the doors.

JohnAFernandez:
We often respect the precedent set by code already in a file.  And given that MJN has more or less effectively been in charge of table code for at least the last year, I think it should be allowed.

JohnAFernandez:
> he does not have sole ownership of the table code 

All I meant is that his work over the past 2 years or so on tables has been the most extensive by far.  I don't think he owns table code, and all the PRs are still under review.

JohnAFernandez (approved):

PR #5523, opened 2023-07-28
Make the escort list show ship names as well as callsigns.

MageKing17:
In #5477, I made the escort list respect hidden names by copying the behavior from `message_queue_process()`, which already checked those flags; however, this resulted in some dramatic behavioral changes for mods using callsigns for escort ships, so making the "replace ship name with callsign" part of the code optional, and disabled by default, should result in minimal changes for existing mods while still fixing the bug of the escort list not hiding ship names properly, and still allowing mods to show callsigns in the escort list if they *do* want that.

Tagging this as "regression" because it's not *strictly* a bug, but affects backwards-compatibility nonetheless. Making this a draft because I don't want to change this again without some proper feedback along the lines of: should it append the callsign instead, like the target window? Should there be more than just two options (like, say, `$Escort callsign behavior: NEVER`)? @Goober5000 made this suggestion on Discord, but it would be significantly more involved to implement:
> How about some game_settings.tbl options allowing the modder to specify formats for the target view, the messages, and the escort list?  The format could use tags like `$name` and `$callsign`, and that way modders could configure displaying ship name, callsign, either, or both; and with whatever parentheses or special characters they want

Further questions about this concept: would each instance of ship names being shown have three options, one for if a callsign is set, one for if the name is hidden, and one for if neither of the above is true? Or would there be some sort of conditional syntax to make things be/not be shown based on some other thing? You have to do *something* more, because otherwise `$name ($callsign)` would result in empty parentheses for ships without callsigns.

### DIFF
--- a/code/hud/hudescort.cpp
+++ b/code/hud/hudescort.cpp
@@ -300,7 +300,6 @@ void HudGaugeEscort::renderIcon(int x, int y, int index)
 	float	shields, integrity;
 	int		screen_integrity, offset, objnum = -1;
 	char	buf[255];
-	const player *pl = nullptr;
 
 	auto eship = get_escort_entry_from_index(index);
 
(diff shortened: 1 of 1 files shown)


TRBlount (approved):

Goober5000 (commented):
The game_settings.tbl options would provide the most flexibility, but as a solution to the current issue, I think using the same behavior as the target box would work.  This would display "Shipname" for ships without a callsign, and "Shipname (Callsign)" for ships with one.  This solution preserves seeing the ship name in all cases, while also satisfying the needs of multiplayer.

TRBlount:
I share the same opinion as Goober. It would be consistent with other feature(s) in the code at a bare minimum.

MageKing17:
I've changed the code to (mostly) match the target box behavior.

Goober5000 (approved):
Looks good.

PR #5524, opened 2023-07-28
aspect-ratio-specific splash screens

Goober5000:
Adds the ability to specify an aspect ratio, or a range of aspect ratios, to be used by splash screens.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -68,7 +68,7 @@ SCP_string Window_title;
 SCP_string Mod_title;
 SCP_string Mod_version;
 bool Unicode_text_mode;
-SCP_vector<SCP_string> Splash_screens;
+SCP_vector<splash_screen> Splash_screens;
 int Splash_fade_in_time;
 int Splash_fade_out_time;
 bool Splash_logo_center;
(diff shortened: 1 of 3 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -228,16 +228,36 @@ void parse_mod_table(const char *filename)
...
+				}
+
+				if (splash.aspect_ratio_exact != 0.0f && (splash.aspect_ratio_min != 0.0f || splash.aspect_ratio_max != 0.0f)) {
+					Warning(LOCATION, "Game Settings Table: An exact aspect ratio and either a min or max aspect ratio were supplied for '%s'.  Only the exact value will be used.", splash.filename.c_str());
+					splash.aspect_ratio_min = 0.0f;
+					splash.aspect_ratio_max = 0.0f;
+				}
+
+				if (splash.aspect_ratio_exact == 0.0f && splash.aspect_ratio_min == 0.0f && splash.aspect_ratio_max == 0.0f) {
+					splash.is_default = true;
```
BMagnu, line 257: Can we have a check to validate that there is at least one default splash screen available?
Not sure we should be relying on bm_load to throw an error if the default graphic isn't supplied with something like a TC that forgot to define a default splash.

freespace2/freespace.cpp
```diff
@@ -7522,8 +7522,48 @@ void game_title_screen_display()
...
+			} else {
+				return !ss.is_default;
+			}
+		}), Splash_screens.end()
+	);
+
+	if (Splash_screens.size() == 1) {
+		Game_title_bitmap = bm_load(Splash_screens[0].filename.c_str());
+	} else if (!Splash_screens.empty()) {
+		Game_title_bitmap = bm_load(Splash_screens[Random::next((int)Splash_screens.size())].filename.c_str());
```
BMagnu, line 7566: Is it sensible to have two lists or a priority sorted list, to basically allow the game to reliably pick an exact match over a range match that may be meant as a backup?

BMagnu (changes requested):

code/mod_table/mod_table.cpp
Goober5000: Splash screens are optional.  "Default" here just means that it will be a candidate if there are matches to the aspect ratio.  If no splash screen is applicable, the game falls back to `Game_title_screen_fname`.  If no file is loaded at all, `game_title_screen_draw` makes allowances for the bitmap being -1.

Goober5000: Unless you mean the warning `"Could not open file 2_Preload!"` that would be shown.

Goober5000: I'm not sure a full-fledged check makes sense.  A file check would need to duplicate some of the code that already exists in `bm_load`.  I could add a log statement though.

BMagnu: Fair enough, I guess it's fine as is then

freespace2/freespace.cpp
Goober5000: The section labeled `// Search splash screens for matches` already does this.  If one or more exact matches exist, it picks from them.  If there are no exact matches but there are range matches, it picks from that list.  If there are no matches of any type, it picks from the defaults.  If there are no matches and no defaults, it fallse back to `Preload` and `2_Preload`.

BMagnu: Ah, I see

BMagnu (approved):

PR #5526, opened 2023-07-29
add screen resolution scaling for hud-set-coords

### LINKED ISSUE #5525: Bug: hud-set-coords lacks resolution scaling
hud-set-coords takes two numbers for the X and Y axis and is resolution-agnostic, meaning the sexp can presently only be set for one resolution.

Solutions here would be to switch to a scalar percentage system from screen center, or my preferred option is to allow fredders to set different offsets as they function presently for different resolutions, eg

Hud-Gauge-Name
Resolution 1
X
Y
Resolution 2
X
Y

Etc

Goober5000:
The hud-set-coords SEXP has always used exact pixels, without regard for the current screen resolution.  This adds a similar scaling mechanism as is used for show-subtitle.

Fixes #5525.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -129,6 +129,7 @@ bool Supernova_hits_at_zero;
 bool Show_subtitle_uses_pixels;
 int Show_subtitle_screen_base_res[2];
 int Show_subtitle_screen_adjusted_res[2];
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):
Very useful fix and LGTM

TRBlount (approved):

PR #5528, opened 2023-07-29
'detonate on expiration' weapon flag

### LINKED ISSUE #5521: Detonate-on-death flag for weapons
Weapons spawning their hit effects when they run out of life rather than just disappearing is pretty. This can be done using $Detonation Range.

However, setting the $Detonation Range to a value equal to the weapon's maximum range (such that it detonates during the same frame in which it dies) will cause the effect to only spawn sometimes, probably because the order in which weapon-killing and effect-spawning occur is nondeterministic. In order to get reliable behavior, one must increase the lifetime of the weapon slightly.

In complex table-stack situations, such as are common with the MediaVPs, this can result in unexpected overrides and bugs.

The best solution to this problem is a simple flag one can add to a weapon which causes it to always detonate when its life runs out.

Goober5000:
For mods which want primary weapons to 'detonate', setting the detonation range is not reliable because, due to rounding imprecision, either the range (which detonates) or the lifetime (which does not) may happen first.  So this adds a flag to always detonate a weapon when its lifetime expires.  Secondary weapons have this flag by default, but primary weapons can add this flag in the $Flags list.  Additionally, if the detonation range and the predicted range due to lifetime are equal, the flag is added automatically.

Resolves #5521.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -243,6 +243,7 @@ special_flag_def_list_new<Weapon::Info_Flags, weapon_info*, flagset<Weapon::Info
 	{ "no evasion",						Weapon::Info_Flags::No_evasion,						    true },
  	{ "don't merge lead indicators",	Weapon::Info_Flags::Dont_merge_indicators,			    true },
 	{ "no_fred",						Weapon::Info_Flags::No_fred,							true },
+	{ "detonate on expiration",			Weapon::Info_Flags::Detonate_on_expiration,				true },
 };
 
 const size_t num_weapon_info_flags = sizeof(Weapon_Info_Flags) / sizeof(special_flag_def_list_new<Weapon::Info_Flags, weapon_info*, flagset<Weapon::Info_Flags>&>);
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):

code/weapon/weapons.cpp
```diff
@@ -5718,27 +5734,28 @@ void weapon_process_post(object * obj, float frame_time)
...
-			if (wip->is_homing()) {
-				Homing_misses++;
-			}
 		} else {
             obj->flags.set(Object::Object_Flags::Should_be_dead);
 		}
+
+		if (wip->is_homing()) {
+			Homing_misses++;
+		}
```
JohnAFernandez, line 5758: I'm not 100% sure this is intentional.  Previously this was triggered if `( wip->subtype & WP_MISSILE ) {` .  So this should either be moved back to somewhere it has the equivalent behavior, or changed to `if ( wip->subtype & WP_MISSILE  && wip->is_homing) {`

JohnAFernandez (changes requested):

code/weapon/weapons.cpp
Goober5000: It's intentional, since primary weapons can home now.  But upon taking a closer look at that variable, its only use is to provide information for an `nprintf` that is now commented out.  So I'll remove it.

JohnAFernandez: All right.  Yeah, any coder that needs that information to balance AI can recreate the statements pretty easily.

JohnAFernandez (approved):

PR #5534, opened 2023-07-31
Add dkjson as a default lua file

MjnMixael:
Adds dkjson, as used by a large number of community scripts, as a default script bundled with FSO. The convenience of always knowing that json functions are available cannot be understated. The main thing I'm unsure of is license compatibility.

Notably, this script creates no globals of any kind. It must be used by first requiring it with `require('dkjson')`.

### DIFF
--- a/code/def_files/data/scripts/dkjson.lua
+++ b/code/def_files/data/scripts/dkjson.lua
@@ -1,0 +1,714 @@
+-- Module options:
+local always_try_using_lpeg = true
+local register_global_module_table = false
+local global_module_name = 'json'
+
+--[==[
+
+David Kolf's JSON module for Lua 5.1/5.2
+
+Version 2.5
+
+
+For the documentation see the corresponding readme.txt or visit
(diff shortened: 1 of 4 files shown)


z64555:
As mentioned on discord, dkjson's licensing verbage would make it appear to be permissive and compatible with FSO's.  However I do question about how well (or ill) defined as to what is the scope "Software."  Is it limited to just that file, or can it be interpreted as files that use it, or files that *it* uses?

The-E:
> As mentioned on discord, dkjson's licensing verbage would make it appear to be permissive and compatible with FSO's. However I do question about how well (or ill) defined as to what is the scope of "the Software." Is it limited to just that file, or can it be interpreted as files that use it, or files that _it_ uses?

In this context, "software" refers to dkjson and _only_ dkjson. Given that dkjson does not pull in dependencies itself, this discussion is somewhat moot - the licensing information in dkjson allows us to freely distribute it, and that's basically it.

The-E (approved):

PR #5535, opened 2023-07-31
Set timer padding sexp

MjnMixael:
BtA has a couple missions with "time skips". In order to help sell the illusion we dreamt up this sexp that simply pads out the HUD timer with N seconds. It's entirely an illusion. Does not affect actual mission time or the mission log in any way.

Also adds a virtvar to get/set the same value.


code/parse/sexp.cpp
```diff
@@ -30421,6 +30440,9 @@ int query_operator_argument_type(int op, int argnum)
 		case OP_TIME_TO_GOAL:
 				return OPF_SHIP;
 
+		case OP_SET_HUD_TIME_PAD:
+			return OPF_POSITIVE;
```
BMagnu, line 30444: Is there a case to be made to allow for negative values of pad?
I assume that if the final time will be negative, it may render weirdly, but since it's purely optical, this probably could be OPF_NUMBER, especially since the scripting API doesn't have the restriction on the sign

BMagnu (approved):

code/parse/sexp.cpp
MjnMixael: Fair point and not requiring someone to use the Lua just to do negative on the off chance they want to would be nice. I'll switch it to OPF_NUMBER.

PR #5536, opened 2023-08-01
Mission custom data

MjnMixael:
Adds custom data to missions. Includes a dialog editor accessible through Mission Specs. Also includes a script function that only runs in FRED to add "default custom data". The intention here is to be able to preload mission files with k/v pairs and a description is then displayed in the dialog editor to explain what a key does and the possible values; making customizing scripted behavior per-mission much more accessible to those who don't know much about Lua scripts.

I put custom data in it's own section in the mission file because I'm working on an additional feature that would save data to the same location; another kind of custom data set available to scripts.


code/mission/missionparse.cpp
```diff
@@ -6002,6 +6004,32 @@ void parse_sexp_containers()
...
+void apply_default_custom_data(mission* pm)
+{
+	for (const auto& def : Default_custom_data) {
+		int count = 0;
+		for (const auto& listed : pm->custom_data) {
+			if (listed.first != def.key) {
+				count++;
+			}
+		}
+		if (count == (int)pm->custom_data.size()) {
```
JohnAFernandez, line 6027: For new code, we should use c++ style casts.  In this case `static_cast<int>(pm->custom_data.size()` should work fine.  I know it sounds picky, but c style casts can accidentally do unintended things, like remove const or use reinterpret cast if we're not careful. Also c++ style casts are easier to search for and therefore audit.

fred2/customdatadlg.h
```diff
@@ -0,0 +1,74 @@
...
+	DECLARE_MESSAGE_MAP()
+
+  private:
+	bool has_containers() const
+	{
+		return !m_custom_data.empty();
+	}
+	int num_containers() const
+	{
+		return (int)m_custom_data.size();
```
JohnAFernandez, line 63: static_cast is fine here, too

fred2/customdatadlg.h
```diff
@@ -0,0 +1,74 @@
+/*
+ * You may not sell or otherwise commercially exploit the source or things you
+ * create based on the source.
+ */
+
+#ifndef _CUSTOMDATA_H
+#define _CUSTOMDATA_H
```
JohnAFernandez, line 7: This is a very technical thing and completely optional, but `#pragma once` should allow you to do what these two lines do (keep the compiler from recompiling this files multiple times, I believe), and then you don't need `#endif // _CUSTOMDATA_H` at the end either.

fred2/customdatadlg.cpp
```diff
@@ -0,0 +1,304 @@
...
+{
+	const int index = m_data_lister.GetCurSel();
+
+	if (index == LB_ERR) {
+		// TODO: clear key and data text  edit boxes?
+		return;
+	}
+
+	Assert(m_data_lister.GetCount() > 0);
+	Assert(index >= 0 && index < (int)m_custom_data.size());
```
JohnAFernandez, line 132: static cast works here

fred2/customdatadlg.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+	const SCP_string& key = m_lister_keys[index];
+
+	update_text_edit_boxes(key, m_custom_data[key]);
+	update_help_text("No help text provided");
+
+	for (int i = 0; i < (int)Default_custom_data.size(); i++) {
```
JohnAFernandez, line 139: static cast works here

fred2/customdatadlg.cpp
```diff
@@ -0,0 +1,304 @@
...
+		if (Default_custom_data[i].key == key) {
+			update_help_text(Default_custom_data[i].description);
+		}
+	}
+
+}
+
+void CustomDataDlg::OnPairAdd()
+{
+	add_pair_entry((int)m_custom_data.size());
```
JohnAFernandez, line 149: Static Cast

fred2/customdatadlg.cpp
```diff
@@ -0,0 +1,304 @@
...
+{
+	if (!edit_boxes_have_valid_data(false)) {
+		return;
+	}
+
+	CEdit* data_edit = (CEdit*)GetDlgItem(IDC_CUSTOM_DATA);
+	CString data_str;
+	data_edit->GetWindowText(data_str);
+
+	CEdit* key_edit = (CEdit*)GetDlgItem(IDC_CUSTOM_KEY);
```
JohnAFernandez, line 162: You can leave this one alone.  The hope being that we could eventually just upgrade to qtFRED and not have to deal with this MFC weirdness anymore.

fred2/customdatadlg.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+	update_data_lister();
+	update_text_edit_boxes("", "");
+}
+
+void CustomDataDlg::OnPairRemove()
+{
+	const int index = m_data_lister.GetCurSel();
+
+	Assert(index >= 0 && index < (int)m_custom_data.size());
```
JohnAFernandez, line 175: static cast

JohnAFernandez (changes requested):
Really just one thing.  This does look like a lot of work was put in and I'm sorry to add more, but with what I've found out about c-style casts, it's just better to stop adding them.

(It only took 3 months to get to review this!)

code/mission/missionparse.cpp
MjnMixael: Converted the ones that needed to stay as Ints because FRED. Changed one or two to size_t. Deleted the others because those functions are actually unused leftovers from copy/paste the containers dialog. Should be good to go. 👍

JohnAFernandez (approved):

PR #5537, opened 2023-08-01
Fix: Prevent loadout related CTD

MjnMixael:
Prevents a vector out of range CTD if a script tries to set the amount of a weapon slot when the currently loaded weapon is invalid.

### DIFF
--- a/code/scripting/api/objs/shipwepselect.cpp
+++ b/code/scripting/api/objs/shipwepselect.cpp
@@ -190,7 +190,7 @@ ADE_INDEXER(l_Loadout_Amount,
 	if (ADE_SETTING_VAR) {
 		ship_info* sip = &Ship_info[current.getBank()->ship_class];
 		int widx = current.getBank()->wep[idx];
-		if (widx >= 0 || widx < (int)Weapon_info.size()) {
+		if (widx >= 0 && widx < (int)Weapon_info.size()) {
 			weapon_info* wip = &Weapon_info[idx];
 			int capacity = 0;
 			// Calculate max capacity of the weapon in the current bank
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/shipwepselect.cpp
```diff
@@ -189,19 +189,24 @@ ADE_INDEXER(l_Loadout_Amount,
 	idx--; // Convert from Lua
 	if (ADE_SETTING_VAR) {
 		ship_info* sip = &Ship_info[current.getBank()->ship_class];
-		weapon_info* wip = &Weapon_info[current.getBank()->wep[idx]];
-		int capacity = 0;
-		//Calculate max capacity of the weapon in the current bank
-		if (wip->subtype == WP_LASER) {
-			capacity = 1;
+		int widx = current.getBank()->wep[idx];
+		if (widx >= 0) {
```
JohnAFernandez, line 193: Might as well check that it's also less than the number of valid weapons.  `weapon_info_size()` iirc

JohnAFernandez (changes requested):
Sorry, meant for that last comment to be request changes.

code/scripting/api/objs/shipwepselect.cpp
MjnMixael: Yeah might as well. Done.

JohnAFernandez: Do I need more coffee or should that be an `&&`?

MjnMixael: It is apparently I that needs more coffee. It should be `&&`.

JohnAFernandez (approved):

PR #5542, opened 2023-08-04
Add death message as a dialog HV

MjnMixael:
Adds the death message that is normally drawn to the screen as an HV of the death popup hook so that SCPUI can #dothingswithit


BMagnu (approved):

MjnMixael:
[Edit DEATH messages to add additional static like BtA1](https://trello.com/c/xcDafiFq/132-edit-death-messages-to-add-additional-static-like-bta1)

PR #5544, opened 2023-08-05
Mission custom strings

MjnMixael:
Here is Custom Strings. Works very similarly to Custom Data except that the strings are more like Mission Messages but without the 512 character limit. Custom Strings have a name, a value, and a text string. Names must be unique. Values are there to associate strings into arbitrary groups. The strings themselves are saved as multitext to allow for newlines and other things that custom data does not allow.

Furthermore this PR includes the first FRED scripting hook. On Mission Specs OK will give a way for Lua to know to check if Custom Data/Strings has been changed. At that point the scripts can use that information to update LuaEnums. The idea here is that LuaEnums can reference arbitrary Custom Data and Strings. Utilizing this scripting hook means that sexp datapoints can be updated with new custom data/strings without restarting FRED or reloading the mission. 

It's basically live CFG/JSON files that many scripts have already been using. But now all manageable within FRED itself.

Relies on #5536 . So that should be reviewed and merged first. Submitting as Draft until then.


fred2/customstringsdlg.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+	OnButtonCancel();
+}
+
+void CustomStringsDlg::OnListerSelectionChange()
+{
+	const int index = m_data_lister.GetCurSel();
+
+	if (index == LB_ERR) {
+		// TODO: clear key and data text  edit boxes?
```
JohnAFernandez, line 132: I am not sure if this should be done or not.

fred2/customstringsdlg.cpp
```diff
@@ -0,0 +1,375 @@
...
+		return;
+	}
+
+	Assert(m_data_lister.GetCount() > 0);
+	Assert(index >= 0 && index < (int)m_custom_strings.size());
+
```
JohnAFernandez, line 138: Cue Assertion and casts.  There's more than just here, I think, but this was the first I found.

fred2/customstringsdlg.cpp
```diff
@@ -0,0 +1,375 @@
...
+}
+
+void CustomStringsDlg::OnEnKillFocusEditString()
+{
+	m_text_edit_focus = false;
+}
+
+bool CustomStringsDlg::query_modified() const
+{
+	//return The_mission.custom_strings != m_custom_strings;
```
JohnAFernandez, line 373: I'm guessing this wasn't working?   Sounds like you would need a nested loop to properly check if anything has been altered.

While most of the code benefits from using a vector, I think this is where using an unordered map (hash map in other languages) would have made things a lot easier here.  Anything where you deal with something having unique keys, an ordered or unordered map will probably have benefits.

JohnAFernandez (changes requested):
Well, it looks like the standard stuff so far.  I think you might have benefited from an unordered map somewhere, but the feature is basically done.

code/mission/missionparse.cpp
```diff
@@ -6105,6 +6105,24 @@ void parse_custom_data(mission* pm)
...
+			stuff_string(cs.value, F_NAME);
+
+			required_string("+String:");
+			stuff_string(cs.text, F_MULTITEXT);
+
+			pm->custom_strings.push_back(cs);
+		}
+
+		required_string("$end_custom_strings");
+	}
```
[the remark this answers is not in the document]
MjnMixael: Mission file versioning is handled in a different section. If custom_strings.size() > 0 then mission save version 24.0 will be required.

fred2/customstringsdlg.cpp
MjnMixael: Ah yes. I obviously intended to do that and forgot.

fred2/customstringsdlg.cpp
MjnMixael: Right. It took me a while to remember this one... FRED doesn't seem to like comparing vectors or even std::equal. So I've implemented a poor-man's version here now instead. It'll do the job.

JohnAFernandez (approved):

wookieejedi:
Will plan to merge this later today unless someone else beats me to it

PR #5551, opened 2023-08-09
remove Int3()'s in ai_mission_goal_achievable

Goober5000:
As reported by Woutersmits, the GTT Nova did not dock with the GTD Hall in the first mission of Black Horizon.  Attempting to run a debug build caused a CTD.  Investigation showed that there was an invalid dockpoint triggering an Int3().  The dockpoint has been fixed, but to aid troubleshooting in the future, the Int3()'s in this function have been changed to more appropriate functions.

### DIFF
--- a/code/ai/aigoals.cpp
+++ b/code/ai/aigoals.cpp
@@ -1749,7 +1749,7 @@ ai_achievability ai_mission_goal_achievable( int objnum, ai_goal *aigp )
 		}
 
 		default:
-			Int3();
+			UNREACHABLE("Unhandled AI goal %d", aigp->ai_mode);
 			status = 0;
 			break;
 	}
@@ -1864,7 +1864,7 @@ ai_achievability ai_mission_goal_achievable( int objnum, ai_goal *aigp )
 		}
 
 		if ( (aigp->dockee.index == -1) || (aigp->docker.index == -1) ) {
-			Int3();			// for now, allender wants to know about these things!!!!
+			Warning(LOCATION, "Cannot determine docking information for %s!", Ships[objp->instance].ship_name);			// for now, allender wants to know about these things!!!!
 			return ai_achievability::NOT_ACHIEVABLE;
 		}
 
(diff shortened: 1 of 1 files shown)


code/ai/aigoals.cpp
```diff
@@ -1995,7 +1995,7 @@ ai_achievability ai_mission_goal_achievable( int objnum, ai_goal *aigp )
 			// short circuit a couple of cases.  Ship not arrived shouldn't happen.  Ship gone means
 			// we mark the goal as not achievable.
 			if ( status == SHIP_STATUS_NOT_ARRIVED ) {
-				Int3();										// get Allender.  this shouldn't happen!!!
+				UNREACHABLE("get Allender.  this shouldn't happen!!!");	// get Allender.  this shouldn't happen!!!
```
JohnAFernandez, line 1998: This message needs more descriptive info.  You can keep get Allender, but someone who doesn't have the skill or desire to look through our source code needs to know the severity of the mistake and what they should do about it.

code/ai/aigoals.cpp
```diff
@@ -1983,7 +1983,7 @@ ai_achievability ai_mission_goal_achievable( int objnum, ai_goal *aigp )
 			else if ( status == SHIP_STATUS_UNKNOWN )
 				return ai_achievability::NOT_KNOWN;
 
-			Int3();		// get allender -- bad logic
+			UNREACHABLE("get allender -- bad logic");		// get allender -- bad logic
```
JohnAFernandez, line 1986: We could dump the status variable here.  "Get allender -- I mean, get the SCP" would direct people better, too.

code/ai/aigoals.cpp
```diff
@@ -1940,7 +1940,7 @@ ai_achievability ai_mission_goal_achievable( int objnum, ai_goal *aigp )
 				aigp->ai_submode = ship_find_subsys( &Ships[sindex], aigp->docker.name );
 				aigp->flags.remove(AI::Goal_Flags::Subsys_needs_fixup);
 			} else {
-				Int3();
+				UNREACHABLE("Target name %s is not an arrived ship!", aigp->target_name);
```
JohnAFernandez, line 1943: I need to look at this when I've had more coffee, but I'm not convinced this should be an UNREACHABLE yet.

JohnAFernandez (changes requested):

code/ai/aigoals.cpp
Goober5000: Good idea

code/ai/aigoals.cpp
Goober5000: I think this situation is ruled out by the rest of the logic.  It would be cleaner to refactor this entire function to use the ship registry, but that's a topic for another PR.

JohnAFernandez (approved):

Issue #5531, opened 2023-07-30
FR: add <all engines>, <all turrets> to set-armor-type SEXP


Durandal-Thoth:
Individual turrets and engines must currently be enumerated when using set-armor-type, unlike similar sexps which allow for blanket solutions as described in the title. This can get messy when dealing with large numbers of ships with differing configurations during arguments.

Goober5000:
Implemented by #5554

PR #5554, opened 2023-08-09
enhance set-armor-type to work with generic subsystems

Goober5000:
What it says on the tin.  Implements #5531.

Also reorders two arguments on the popup menu to be consistent with other sexps.


wookieejedi (approved):
Works as expected and code is clear and straightforward.

PR #5557, opened 2023-08-13
Lua method to launch arbitrary url

MjnMixael:
Exposes the URL launching method from multi_pxo to Lua for SCPUI. It would be handy to be able to link to a mod's forum or other related links directly from the game, much like how modern games do the same.


JohnAFernandez (approved):
LGTM

PR #5558, opened 2023-08-14
some streaming animation fixes

Goober5000:
1. Since other functions in the scripting API remove extensions before loading the files, and since the internal functions expect them to be removed, the streaming anim API should handle that too.
2. Make streaming animations obey the current scripting resize mode.
3. Make streaming animations obey the passed alpha value.


wookieejedi (approved):
Looks good to me at least

PR #5560, opened 2023-08-16
improve launching of FRED help

Goober5000:
1. Fall back to the FREDzone help docs if the local files cannot be found
2. Tweak url_launch error message and add a return value

This consequently removes the ReleaseWarning which wasn't the friendliest or clearest way to tell the user about a missing help file.


wookieejedi (approved):
Tests work and overall looks good

PR #5561, opened 2023-08-16
add "special warp physics" to custom warpin parameters

Goober5000:
Supercaps use a physics flag on warpin that makes them slow down faster than they would under the normal physics code.  This adds that flag as an option for custom warp parameters.

Since flags are not yet parsed when warp parameters are parsed, the code needs to skip ahead and peek at the flags in the same way as the weapon override.


code/ship/ship.cpp
```diff
@@ -2728,7 +2728,7 @@ int parse_warp_params(const WarpParams *inherit_from, WarpDirection direction, c
 		if (accel_exp >= 0.0f)
 			params.accel_exp = accel_exp;
 		else
-			Warning(LOCATION, "%s specified as less than 0 on %s '%s'; value ignored", str, info_type_name, info_name);
+			error_display(0, "%s specified as less than 0 on %s '%s'; value ignored", str, info_type_name, info_name);
```
Baezon, line 2731: Is there a reason to prefer `error_display`? I'll admit that `Warning(..` is much more intuitive to me than `error_display(0, ..`, I can never remember which one 0 or 1 is.

Baezon (approved):

code/ship/ship.cpp
Goober5000: Increasing number = increasing severity, so 0 = warning and 1 = error.

The `error_display` function is specifically intended to be used while parsing.  It displays the file name and line number of the file being parsed; and it also maintains warning and error counts for parsing, distinct from "all" warning and error counts.

Regular `Warning` and `Error` do not display anything about the file being parsed.

PR #5567, opened 2023-08-22
handle extension mismatch in event music

Goober5000:
~~Currently, menu music will trim the extension and look for both OGG and WAV files, but event music will use the exact extension specified.  This behavior dates all the way back to commit f969eb52a2 from 2010.  There doesn't seem to be an obvious reason for this, and it leads to a bug in FSPort where in some cases music will not play during a mission, in situations where the table uses .wav but the music pack provides .ogg.  So, let's make event music extension-agnostic.~~

Currently, menu music will trim the extension and look for both OGG and WAV files, but event music will use the exact extension specified.  This is intended but could lead to unexpected results in mods, so display a warning for the modder.

### DIFF
--- a/code/sound/ffmpeg/FFmpegWaveFile.cpp
+++ b/code/sound/ffmpeg/FFmpegWaveFile.cpp
@@ -211,6 +211,20 @@ bool FFmpegWaveFile::Open(const char* pszFilename, bool keep_ext)
 			auto res = cf_find_file_location(pszFilename, CF_TYPE_ANY);
 
 			if (!res.found) {
+#ifndef NDEBUG
+				char truncated_filename[MAX_FILENAME_LEN];
+				strcpy_s(truncated_filename, filename);
+				auto ext_ch = strrchr(truncated_filename, '.');
+				if (ext_ch != nullptr)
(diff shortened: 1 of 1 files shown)


notimaginative:
Think I figured it out. It's the pattern transitions in music.tbl (num_measures / samples_per_measure) which could differ with different audio formats. So that's why the the filename was kept, to avoid it using an audio file that didn't match the values in the table.

I'm not sure to what degree that's actually a problem, but if I recall correctly, it does sound like a bug that someone reported once upon a time.

notimaginative:
The original commit is https://github.com/scp-fs2open/fs2open.github.com/commit/32361c79eef438af0fbf740076bc6e4da320589b, from 2007. It's 4 different changes altogether, for some reason, but the bulk of it is this particular change.

Goober5000:
Ah, that makes sense.  Hmm.

Goober5000:
Ok, I've changed the PR to display a warning instead, and I'll edit the modpack to specify the appropriate extension when the new music files are used.

notimaginative (approved):
Yeah that looks good to me. Probably the safest way to handle it too.

PR #5568, opened 2023-08-22
fix crash in campaign editor

### LINKED ISSUE #5514: OpenGL crash in FRED campaign editor when adding end-of-campaign link
This is bizarre, but easily reproducible.
1. Launch FRED
2. Launch the campaign editor
3. Add one mission to the newly created campaign
4. Right-click on that mission and select "End of Campaign"
5. FRED crashes and exits

The crash does not occur in Debug but does occur in Fast Debug and Release.  The specific bug is an access violation in `gr_opengl_sync_fence()` in gropenglsync.cpp.

(As a slight complication, it crashes for Galemp in Release, but not for me.  It does crash for me in Fast Debug, but not in Debug.)

Stack trace:
```
 	00007fff2eb012e0()	Unknown
>	fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!gr_opengl_sync_fence() Line 5	C++
 	[Inline Frame] fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!std::_Func_class<void *>::operator()() Line 968	C++
 	[Inline Frame] fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!gr_sync_fence() Line 1263	C++
 	fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!graphics::util::UniformBufferManager::onFrameEnd() Line 96	C++
 	[Inline Frame] fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!uniform_buffer_managers_retire_buffers() Line 2628	C++
 	fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!gr_flip(bool) Line 2589	C++
 	fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!render_frame() Line 1673	C++
 	fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!CFREDApp::OnIdle(long lCount) Line 448	C++
 	fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!CWinThread::Run() Line 621	C++
 	fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!CWinApp::Run() Line 787	C++
 	fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!AfxWinMain(HINSTANCE__ * hInstance, HINSTANCE__ * hPrevInstance, char * lpCmdLine, int nCmdShow) Line 47	C++
 	fred2_open_23_1_0_x64_AVX2-FASTDBG.exe!WinMain(HINSTANCE__ * hInstance, HINSTANCE__ * hPrevInstance, char * lpCmdLine, int nCmdShow) Line 26	C++
 	[External Code]	
```

Goober5000:
The crash in #5514 appears to be caused by some of the graphics code not being in the expected state for the synchronization.  Looking at the code in `render_frame()`, a `gr_flip()` happens after one frame ends and before the next frame begins.  That implies that this `gr_flip()` should not be here, as it is in the middle of a frame.  For additional evidence, `gr_flip()` modifies the "uniform buffer manager" state, and that is part of the stack trace preceding the crash.  Removing this `gr_flip()` fixes the crash.

The strange thing is that this code *doesn't run at all* while the campaign editor is active.  I verified this by adding a MessageBox in place of `gr_flip()`; the MessageBox doesn't display until the campaign editor is exited.  And yet, FRED crashes with `gr_flip()` present and doesn't crash with it removed.  This calls for a new category of bug; I propose "spookybug", from spooky action-at-a-distance.

Fixes #5514.

### DIFF
--- a/fred2/fred.cpp
+++ b/fred2/fred.cpp
@@ -670,8 +670,6 @@ void update_map_window() {
 
 	render_frame();	// "do the rendering!"
 
-	gr_flip();
-
 	show_control_mode();
 	process_pending_messages();
 


BMagnu (approved):
Yeah, there shouldn't be a flip immediately after the last without drawing.
That's certainly not sensible.

PR #5569, opened 2023-08-23
Do not validate loadouts

MjnMixael:
Adds a feature/flag to loadouts to prevent FRED from validating the weapons pool. You may be wondering why I've written this code because validation was added for a very good reason! Well, BtA2 manages the loadout entirely with variables and scripts. FRED's validation treats variables as 0 amount in the pool and automatically appends the "weapons used in wings" to the loadout pool. That's pretty handy 9 times out of 10, but for our very advanced use-case we need FRED to not do that.

So this flag will simply stop FRED from checking the loadout pool and trust that the mission designer is making the correct choices about what's available.


TRBlount (changes requested):
This PR affects the mission save format. Please update `check_for_23_3_data()` to check if any team has the `do_not_validate` flag set.

wookieejedi:
LGTM to me, what about you @TRBlount?

TRBlount (approved):
Looks good.

PR #5572, opened 2023-08-23
fix validity check for soundtracks

Goober5000:
Since event music soundtracks use the exact file extension specified (see #5567), the validity check should also use the exact file extension specified.

Also, move the validity check to after parsing has completed because extensions may be overridden.

### DIFF
--- a/code/gamesnd/eventmusic.cpp
+++ b/code/gamesnd/eventmusic.cpp
@@ -311,14 +311,8 @@ void event_music_init()
 			if (!cf_exists_full(filename, CF_TYPE_MUSIC))
 			{
 #ifndef NDEBUG
-				char truncated_filename[MAX_FILENAME_LEN];
-				strcpy_s(truncated_filename, filename);
-				auto ext_ch = strrchr(truncated_filename, '.');
-				if (ext_ch != nullptr)
-					*ext_ch = '\0';
-
 				// see if the file exists with a different extension
-				auto res = cf_find_file_location_ext(truncated_filename, NUM_AUDIO_EXT, audio_ext_list, CF_TYPE_MUSIC);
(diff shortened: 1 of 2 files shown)


code/gamesnd/eventmusic.cpp
```diff
@@ -289,6 +289,49 @@ void event_music_init()
...
+				char truncated_filename[MAX_FILENAME_LEN];
+				strcpy_s(truncated_filename, filename);
+				auto ext_ch = strrchr(truncated_filename, '.');
+				if (ext_ch != nullptr)
+					*ext_ch = '\0';
+
+				// see if the file exists with a different extension
+				auto res = cf_find_file_location_ext(truncated_filename, NUM_AUDIO_EXT, audio_ext_list, CF_TYPE_MUSIC);
+				if (res.found)
+					Warning(LOCATION, "Soundtrack file %s was not found with its specified extension, but another file %s exists in the modpack.  Please update the extension and adjust audio specifications if necessary.", filename, res.name_ext.c_str());
```
notimaginative, line 323: You can really just use `cf_exists_full_ext()` here to simplify things, and it should strip the extension for you too.

So this should be all you need:
```
if (cf_exists_full_ext(...)) {
    Warning(...);
}
```

notimaginative (changes requested):

code/gamesnd/eventmusic.cpp
Goober5000: Taking care of the extension is nice, but I do want to use `cf_find_file_location_ext` so I can show the file that was actually found.

notimaginative: Ah, good point. I missed that you were doing that.

notimaginative (approved):

PR #5573, opened 2023-08-24
Fix issues with newer Nvidia driver

### LINKED ISSUE #5430: Strange lighting artefact. NVIDIA driver bug?
Illustrated with this video. https://www.youtube.com/watch?v=e5HpH2XuKSI&
Reproduceable with Solaris 2.2.0 on knossos, mission 15.fs2 in the Black Sun campaign.

First noticed when upgraded NVIDIA drivers. Artefacts seems to disappear with the February 2023 GeForce RTX 3070 drivers.
https://www.nvidia.com/en-gb/drivers/results/199863/

BMagnu:
What I _think_ happened here is that the Nvidia compiler saw the calculation of ``lightPosition`` in the vertex shader (which turns out to be a dynamically uniform expression) as optimizable, but somehow messed up how its passed as a parameter to the fragment shader. If this calculation is moved to the fragment shader, which should be performance-equivalent since it's a dynamically uniform expression, the issue goes away.
Fixes #5430

### DIFF
--- a/code/def_files/data/effects/deferred-f.sdr
+++ b/code/def_files/data/effects/deferred-f.sdr
@@ -3,8 +3,6 @@
 
 #include "shadows.sdr" //! #include "shadows.sdr"
 
-in vec3 beamVec;
-in vec3 lightPosition;
 out vec4 fragOut0;
 
 uniform sampler2D ColorBuffer;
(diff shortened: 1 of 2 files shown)


The-E:
This does indeed fix the issue for me. VERY well done!

The-E (approved):

PR #5577, opened 2023-08-25
additional music checks

### LINKED ISSUE #5576: New soundtrack validity checks cause erroneous warnings for some mods
In #5572 validity checks were added without considering the full implications for mods that don't use soundtracks.

Listing at least one soundtrack in music.tbl is currently required by FSO. So it has been common to list the non-existent sound file as silence.wav, like so: `$Name:	silence.wav 0.0 0` for each required $Name: in a $Soundtrack. 

This was enough to get past the parsing error and FSO would never do anything, even if a mission called the soundtrack with these files. Now, since a valid file is required, it can cause further _unclear_ problems (asserts that don't offer much info) in missions that erroneously try to play the psuedo soundtrack because the Num Measures and Samples Per Measure are allowed to pass parsing as 0 when that's nonsensical.

There are a number of ways to fix this.. the best would be to allow parsing 0 soundtracks. At a minimum checking Num Measures and Samples Per Measure during parsing would work to avoid the asserts later.

There is also an assert if there aren't enough $Name: fields for a soundtrack, so that should also be validated. This assert was never triggered when the sound file didn't exist. But now that it's required to exist, further validity checks must be added.

Goober5000:
This adds several fixes to make the event and menu music more robust:
1. The game will now run with zero soundtracks.  All accesses into a potentially empty vector should now have checks.
2. Soundtracks and menu music are now checked for validity before being played (not just when importing a FSM mission).
3. The parsing validity checks and the in-game event music now check for skippable patterns using the same function.
4. Zero measures and zero samples-per-measure now cause the game to skip a pattern.  Negative values now trigger a warning.
5. Omitting some or all of the `$Name:` lines for a soundtrack will not cause an assert.

Fixes #5576.  Follow-up to #5572.

### DIFF
--- a/code/gamesnd/eventmusic.cpp
+++ b/code/gamesnd/eventmusic.cpp
@@ -257,6 +257,19 @@ SCP_vector<menu_music> Spooled_music;
 int Mission_music[NUM_SCORES];
 
 
+bool event_music_pattern_is_skippable(const SOUNDTRACK_PATTERN_INFO &pattern)
+{
+	// check for blank or "none"
+	if (pattern.fname[0] == '\0' || !strnicmp(pattern.fname, NOX("none.wav"), 4))
+		return true;
+
+	// no content means this pattern isn't intended to play
+	if (pattern.num_measures == 0 || pattern.samples_per_measure == 0)
+		return true;
+
+	return false;
+}
+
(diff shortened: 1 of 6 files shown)


wookieejedi (approved):
Good to have this cleaned up and LGTM.

PR #5580, opened 2023-08-31
Change ubuntu-latest to ubuntu-20.04

Goober5000:


notimaginative:
I don't think the change is needed for many of those places. It should only be necessary for the `build_linux` jobs, where it actually compiles the game. So just the one change in nightly/release/test should do the trick. Assuming I understand how the workflow stuff works anyway.

The standalone yaml probably doesn't need it either since it appears to build as part of the docker image and therefore shouldn't need to be compatible with anything but itself.

PR #5581, opened 2023-09-01
Follow-up for briefing icon scaling

TRBlount:
This extends the handling of scaling to all briefing icon types and uses the standard method of saving briefing icon info as the old method did not account for all cases.


Baezon:
Is this fixing any bugs? If not, this should wait for the stable.

TRBlount:
Yes actually, this fixes briefing scaling in FRED and rectifies the behavior of missionsave for icons.

Baezon (approved):

PR #5582, opened 2023-09-01
Update starfield animation handling

TRBlount:
This fixes the formula (Which has been incorrect for quite some time) used to get the next frame of an animation used in a starfield and extends a previous change to apply to all bitmap animation types.

### DIFF
--- a/code/starfield/starfield.cpp
+++ b/code/starfield/starfield.cpp
@@ -108,12 +108,10 @@ typedef struct starfield_bitmap_instance {
 	int div_x, div_y;								// # of x and y divisions
 	angles ang;										// angles from FRED
 	int star_bitmap_index;							// index into starfield_bitmap array
-	int timestamp;
-	int cur_frame;
 	int n_verts;
 	vertex *verts;
 
-	starfield_bitmap_instance() : scale_x(1.0f), scale_y(1.0f), div_x(1), div_y(1), star_bitmap_index(0), timestamp(0), cur_frame(0), n_verts(0), verts(NULL) {
+	starfield_bitmap_instance() : scale_x(1.0f), scale_y(1.0f), div_x(1), div_y(1), star_bitmap_index(0), n_verts(0), verts(NULL) {
 		ang.p = 0.0f;
 		ang.b = 0.0f;
 		ang.h = 0.0f;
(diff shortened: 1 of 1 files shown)


code/starfield/starfield.cpp
```diff
@@ -1584,13 +1545,13 @@ void stars_draw_bitmaps(int show_bitmaps)
...
 			if (Starfield_bitmaps[star_index].fps) {
-				bitmap_id = stars_get_next_frame(Starfield_bitmap_instances[idx], Starfield_bitmaps[star_index], timestamp());
+				bitmap_id = Starfield_bitmaps[star_index].bitmap_id + ((timestamp() * Starfield_bitmaps[star_index].fps / MILLISECONDS_PER_SECOND) % Starfield_bitmaps[star_index].n_frames);
 			} else {
 				bitmap_id = Starfield_bitmaps[star_index].bitmap_id;
 			}
 		} else {
 			if (Starfield_bitmaps[star_index].fps) {
-				bitmap_id = stars_get_next_frame(Starfield_bitmap_instances[idx], Starfield_bitmaps[star_index], timestamp());
+				bitmap_id = Starfield_bitmaps[star_index].bitmap_id + ((timestamp() * Starfield_bitmaps[star_index].fps / MILLISECONDS_PER_SECOND) % Starfield_bitmaps[star_index].n_frames);
```
Baezon, line 1554: An int being divided by an int, this is integer division so if this is 30 fps / 1000 won't it always be 0?

Baezon (changes requested):
You should mention what is the specific error being corrected here?

code/starfield/starfield.cpp
TRBlount: The error was a unit mismatch, it's been corrected now and tested to function correctly. After this change, the output is corrected via:
```
timestamp() * Starfield_bitmaps[star_index].fps / MILLISECONDS_PER_SECOND
=
milliseconds * (frames / seconds) / 1000
=
seconds * 1000 * (frames / seconds) / 1000
= 
frames
```

Baezon: `seconds * 1000 * (frames / seconds) / 1000` this is the part that worries me, since it's unclear what the order of operations is, and it would change the result.
I trust you that it works out, but I would appreciate parentheses `((timestamp() * Starfield_bitmaps[star_index].fps) / MILLISECONDS_PER_SECOND)` to make it more clear that this operation is executing as it should.

TRBlount: This should be addressed now.

Baezon (approved):

PR #5583, opened 2023-09-02
Add VR support through OpenXR

BMagnu:
This adds OpenXR bindings in order to bring HMD-based VR and its associated headtracking and stereoscopic rendering into FSO. It does this by providing a state machine. If a frame wants to be stereoscopic, it needs to announce this to the OpenXR integration, which sets up the necessary frame details and then expects two gr_flips, one for each eye. If it gets a gr_flip outside of this, it'll assume it's a monoscopic 2D-frame, and will create a fake 3D environment with the frame on a flat plane in this space.

Some caveats for VR in FSO though:
Edit: Mostly superceded by https://wiki.hard-light.net/index.php/Virtual_Reality

To enable VR, set -vr in the commandline
On Windows, the WMR and Oculus runtime are _not_ compatible with OpenGL. You must set a different runtime. SteamVR as the OpenXR runtime is known to work. Depending on whether or not you had the Headset on when you started FSO, you may need to press Win-Y to pass through the mouse to FSO.
On Linux, ~~session initialization is untested as of now. If anyone has a linux setup with VR, please reach out to me!~~ Edit: Seems to work fine!
On Mac, VR with OpenGL will not work, full stop.
Launching in windowed mode is encouraged.
Anything rendered by scripts (including AxMessage!) will not properly adjust for the eyes. Only if something is rendered through the functions specially provided for scripted HUD gauges will it work properly. I have an updated AxMessage for that in the works, but all currently distributed versions will not be readable in VR.

Reported Issues and their fixes:
OpenXR toolkit (Oculus) can crash if it's an old version. Make sure your OpenXR toolkit is up to date
FSO breaks at high resolutions on Windows, so use ``-window_res`` to reduce the resolution of the window while keeping the render resolution high.


code/cmdline/cmdline.cpp
```diff
@@ -1843,6 +1847,14 @@ bool SetCmdlineParams()
 		}
 	}
 
+	if (vr.found()) {
+		Cmdline_enable_vr = true;
+		if (fov_arg.found() || fov_cockpit_arg.found()) {
+			Warning(LOCATION, "FoV Command line settings found. VR-Mode will override these with the HMD's native FoV...");
```
The-E, line 1853: I wonder if it makes sense to just ignore the fov arguments, given how much more relevant they are in a VR context. I know nothing about VR, so this is a genuine question: Shouldn't we be using the headset FOV exclusively?

BMagnu: Yes, this is exactly what happens.
In VR mode, once the headset is initialized a fair bit later, the FOV setting will be overridden by what the API reports.
It just warns the user here, so that they don't get confused why the setting doesn't change anything

The-E: argh, sorry, misread the code :D

Goober5000:
This is a lot of code.  Can some of it be spun off into smaller PRs?  "Getting the HUD code in order", for example, could be its own PR.

BMagnu:
> This is a lot of code. Can some of it be spun off into smaller PRs? "Getting the HUD code in order", for example, could be its own PR. So could the version update in `code/scripting/api/libs/base.cpp` and related files.

It's certainly not impossible, though a lot of work.
Also, it'd need some commitments from reviewers in that case, since I don't really want to have to chase after reviews four times instead of once, especially with the low amount of active graphics coders we have.

BMagnu:
It's now at a point where I am reasonably sure it's finished.
There are some more optional things, like the option to reset VR position, set a custom scale for HUD and UI, but that's all stuff that can reasonably go into followup PRs.
As for splitting this into smaller PRs, I'll try the following:
1. Split of the Version struct change
2. I'll try to split off the fixed for the HUD when slewing
3. Depending on how well 2. works, I might also split off the groundwork for asymmetric FoV's.

The-E (approved):

PR #5585, opened 2023-09-04
additional beam-related features for Lua

Goober5000:
1. Add a `vanish()` function for beam objects, to be consistent with other object types
2. Add a `BeamFree` virtvar, analogous to `TurretLocked`
3. Add timestamp variants of TurretLocked and BeamFree to be consistent with sexps
4. Add `BeamLife`, `BeamWarmup`, and `BeamWarmdown` virtvars so scripts can access table properties
5. Add "On Beam Death", "On Beam Warmup", and "On Beam Warmdown" hooks
6. Clarify some existing documentation and patch a couple things

### DIFF
--- a/code/scripting/api/objs/subsystem.cpp
+++ b/code/scripting/api/objs/subsystem.cpp
@@ -588,7 +588,7 @@ ADE_VIRTVAR(TurretLocked, l_Subsystem, "boolean", "Whether the turret is locked.
 	return ade_set_args(L, "b", (sso->ss->weapons.flags[Ship::Weapon_Flags::Turret_Lock]));
 }
 
-ADE_VIRTVAR(TurretLockedWithTimestamp, l_Subsystem, "boolean", "Behaves like TurretLocked, but when the turret is freed, the next-fire timestamp is set to be consistent with SEXP behavior.", "boolean", "True if turret is locked, false otherwise")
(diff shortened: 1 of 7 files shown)


code/scripting/api/objs/subsystem.cpp
```diff
@@ -587,6 +588,60 @@ ADE_VIRTVAR(TurretLocked, l_Subsystem, "boolean", "Whether the turret is locked.
 	return ade_set_args(L, "b", (sso->ss->weapons.flags[Ship::Weapon_Flags::Turret_Lock]));
 }
 
+ADE_VIRTVAR(TurretLockedWithTimestamp, l_Subsystem, "boolean", "Behaves like TurretLocked, but when the turret is freed, the next-fire timestamp is set to be consistent with SEXP behavior.", "boolean", "True if turret is locked, false otherwise")
```
Baezon, line 591: I would be more specific and say that this specifically will add a short random delay before firing (as as to be consistent yadda yadda), a user isn't going to know what "the next-fire timestamp is set" implies. Ditto below.

Although... the next fire time stamp is settable via script *anyway*, so I have to wonder, if the scripter really wants this short delay perhaps it would be best to simply let them do it? Rather than complicate the API with this tiny convenience.

Baezon (changes requested):

code/scripting/api/objs/subsystem.cpp
Goober5000: I think it's worthwhile to have a script operation that behaves exactly like the sexp, even if it's equivalent to two existing script operations.  Plus, if this operation is used a lot, the scripter would otherwise have to keep the extra operation in mind or use `mn.evaluateSEXP`.

I'll update the description; that's a good idea.

Baezon (approved):

PR #5588, opened 2023-09-06
Fix for modular table contrail initialization

Baezon:
If new contrails are defined in a modular tbm, certain fields may not be initialized correctly (namely `a_decay_exponent` and `texture_stretch`, which must be 1), since they are typically only initialized on the ship's first definition, so I've added an explicit trail info initializer. Ship contrails are the only one with this issue, since they are the only circumstance of an unknown, variable amount of trail_infos.


Goober5000 (approved):
This is a good candidate for putting into a SCP_vector, but for the time being, this is a good fix.

PR #5591, opened 2023-09-09
Make invalid model_subsystem model_nums a hard error

Baezon:
I have looked at making this fail more gracefully, but the system is quite entangled. The `ship_info` has already allocated these `model_subsytem`s and they are implicitly trusted by the very subsystem itself, as well as any other ship classes which use the same model, which will *also* lean on those same `model_subsystem`s. Failing to find the model entry fails to set up a lot of data which rapidly escalates into memory errors. Consider this PR a request for help on how to make this gracefully fail as much as an advocacy to make it an error.

I have seen this surprise users at least a dozen times now. It introduces bizarre bugs and behaviors which users have difficulty reconciling with the nature of the warning, and, in the most recent case, *earlier instances of this issue on different ships masked the existence of the issue on another ship* (presumably some shenanigans with `ship_subsys_free_list`) further making it difficult for the user to diagnose the problem. Users cannot be expected to reason about memory errors, and so, we should not let the program proceed.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -7400,7 +7400,10 @@ static int subsys_set(int objnum, int ignore_subsys_info)
 	{
 		model_system = &(sinfo->subsystems[i]);
 		if (model_system->model_num < 0) {
-			Warning (LOCATION, "Invalid subobj_num or model_num in subsystem '%s' on ship type '%s'.\nNot linking into ship!\n\n(This warning means that a subsystem was present in the table entry and not present in the model\nit should probably be removed from the table or added to the model.)\n", model_system->subobj_name, sinfo->name );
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
I'm ok with this, and your explanation sounds reasonable, but I'm surprised the problems are as significant as you describe.

There is a possibility there may be some mods in the wild that will no longer be able to launch due to this change, but this isn't the first time a FSO fix has required a mod update.

Baezon:
It is possible, although I think it is unlikely, that a mod *could* proceed anyway without observable effects. But given I know its a memory bug of some kind, I am also quite sure that similar issues *could also* crop up if they decide to play missions in a different order (by causing ships to load in a different order, and step on each other in different ways), which is also kind of really bad.

Goober5000 (approved):
I'm ok with approving this.

PR #5594, opened 2023-09-10
Save player after changing campaign in scpui

MjnMixael:
SCPUI loads the campaign file from the player file after leaving the Barracks and in a few other cases. Normally this is fine, but if the player recently changed campaigns, the player save file will be out of date. So this one-line change forces saving the player file whenever the campaign is changed in SCPUI.


Baezon:
This would be a bugfix, yeah?

MjnMixael:
> This would be a bugfix, yeah?

Yes I think this qualifies as a bugfix.

Baezon (approved):

TRBlount (approved):

Issue #5503, opened 2023-07-20
FR: disable requirement for functioning engines from "ship-maneuver"-sexp


skdKitsune:
The "ship-maneuver"-sexp currently requires a functioning engine subsystem to function. 

This is inconvenient, especially when trying to have "husks" of "dead ships" moving e.g. from an explosion blast.

Goober5000:
Implemented in #5597.

PR #5597, opened 2023-09-14
add a maneuver-despite-engines flag

Goober5000:
This flag ignores engine status when determining whether and how a ship can maneuver.  Implements #5503.

### DIFF
--- a/code/hud/hudets.cpp
+++ b/code/hud/hudets.cpp
@@ -125,25 +125,25 @@ void update_ets(object* objp, float fl_frametime)
 	//					 This will translate to 71% max speed at 50% engines, and 31% max speed at 10% engines
 	//
 
-	float eng_current_strength;
+	float engine_aggregate_strength;
 	if (ship_p->flags[Ship::Ship_Flags::Maneuver_despite_engines]) {
-		// Since prev_engine_aggregate_strength isn't used anywhere else, it's ok to fake setting the strength to 100% here
-		eng_current_strength = 1.0f;
+		// Pretend our strength is 100% when this flag is active
+		engine_aggregate_strength = 1.0f;
 	} else {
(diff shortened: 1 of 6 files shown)


code/hud/hudets.cpp
```diff
@@ -125,10 +125,16 @@ void update_ets(object* objp, float fl_frametime)
 	//					 This will translate to 71% max speed at 50% engines, and 31% max speed at 10% engines
 	//
 
-	float eng_current_strength = ship_get_subsystem_strength(ship_p, SUBSYSTEM_ENGINE);
+	float eng_current_strength;
+	if (ship_p->flags[Ship::Ship_Flags::Maneuver_despite_engines]) {
+		// Since prev_engine_aggregate_strength isn't used anywhere else, it's ok to fake setting the strength to 100% here
```
BMagnu, line 130: While this may be the case now, I'd rather future proof it.
It shouldn't be too hard to, instead setting eng_current_strength, simply call ets_update_max_speed and skip the ``eng_current_strength != ship_p->prev_engine_aggregate_strength`` condition

BMagnu (changes requested):

code/hud/hudets.cpp
Goober5000: After fiddling with the code for quite a bit, I think it needs to stay the way it is.  The strength/prev_strength check is needed to minimize changes to the ETS system and prevent `ets_update_max_speed` from being called every frame -- see #5368 and related issues/PRs.  The flagged behavior is appropriate when these constraints are taken into account, and the `prev_engine_aggregate_strength` field really is supposed to only be used here.  So I've edited the command and made a few other changes to clarify the situation.

BMagnu: With the comment in ship.h it should be fine

BMagnu (approved):

PR #5598, opened 2023-09-14
Fix a few typos

twolife:
Reported by Lintian when packaging the engine for Debian; it shouldn't hurt

### DIFF
--- a/fred2/campaigntreeview.cpp
+++ b/fred2/campaigntreeview.cpp
@@ -118,7 +118,7 @@ void campaign_tree_view::OnDraw(CDC* pDC)
 	Assert(r);
 
 	pDC->SelectObject(&black_pen);
-	// draw level seperators
+	// draw level separators
 	for (i=1; i<total_levels; i++) {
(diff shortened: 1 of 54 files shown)


fred2/campaigntreeview.cpp
```diff
@@ -118,7 +118,7 @@ void campaign_tree_view::OnDraw(CDC* pDC)
 	Assert(r);
 
 	pDC->SelectObject(&black_pen);
-	// draw level seperators
+	// draw level separator
```
Goober5000, line 121: This should be plural
```suggestion
	// draw level separators
```

Goober5000 (changes requested):
A lot of changes, but they all look good.  Just one small tweak needed.

fred2/campaigntreeview.cpp
twolife: Ooops, good catch :-)

Goober5000 (approved):

Goober5000:
@twolife We're ready to merge this, but one of the files now has conflicts due to the addition of some XSTRs.  Could you updaate the PR?

PR #5599, opened 2023-09-14
Actually reset the weapon pool with SCPUI

MjnMixael:
Fixes a bug where `ui.ShipWepSelect.resetSelect` would slowly deplete the available weapon pool until no weapons are available.

### DIFF
--- a/code/missionui/missionweaponchoice.h
+++ b/code/missionui/missionweaponchoice.h
@@ -12,6 +12,8 @@
 #ifndef __MISSION_WEAPON_CHOICE_H__
 #define __MISSION_WEAPON_CHOICE_H__
 
+#include "mission/missionparse.h"
+
 class p_object;
 struct wss_unit;
 class ship_weapon;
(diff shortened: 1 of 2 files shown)


code/missionui/missionweaponchoice.h
```diff
@@ -12,6 +12,8 @@
 #ifndef __MISSION_WEAPON_CHOICE_H__
 #define __MISSION_WEAPON_CHOICE_H__
 
+#include "mission/missionparse.h"
```
TRBlount, line 15: I don't believe this include is actually needed here and could be replaced with a forward declaration of `struct team_date` since `wl_init_pool` only takes a pointer. This is however just a maintainability recommendation and shouldn't affect anything. Other than that this looks good.

MjnMixael: I tried a forward declaration originally within the header and couldn't get it to build without error. Perhaps I did it wrong.

TRBlount: After some more inspection, it looks like `team_data` is an anonymous struct, hence why it can't be forward declared. If it was changed from:

```
typedef struct {
...
} team_data;
```
to:
```
typedef struct team_data {
...
} team_data;
```

It would work.

MjnMixael: I feel like that's outside the scope of this PR, but I'll add it if you want me to.

TRBlount (approved):

code/missionui/missionweaponchoice.h
TRBlount: It's fine. We can handle it after the Release if need be.

PR #5600, opened 2023-09-15
Add default player weapon flag

### LINKED ISSUE #5552: FR: "Default Unavailable" flag for "player allowed" flagged weapons
All weapons flagged with "player allowed" are presently available by default in the loadout pool in FRED. More often than not this is a hindrance, especially earlier on in campaigns when most tech is unavailable, or when a weapon is used only once or twice in a campaign.

"player allowed" flagged weapons would check for the "default unavailable" flag, and if it is present, not default to available in the loadout pool.

MjnMixael:
Adds the weapon equivalent to "default player ship" flag. This flag defines what weapons are default in the player loadout in FRED/qtFRED. Fixes #5552


JohnAFernandez (approved):
So what I'm reading is that basically keeps us from having to uncheck 1230812312937 weapons on the loadout screen.

MjnMixael:
> So what I'm reading is that basically this keeps us from having to uncheck 1230812312937 weapons on the loadout editor.

Yup. That's correct

Issue #5495, opened 2023-07-15
Lua drawstring wordwrap is expensive


EatThePath:
Not a critical issue, but one worth attending to sooner or later: Profiling shows `gr.drawString` appears to spend a lot of time working on splitting strings for wordwrapping. Two potential options for mitigating this have been suggested, and I think both are worth implementing.

1. Expose a version of `split_str()` to the lua API and provide a non-wrapping way to draw strings, possibly just as optional functionality of gr.drawString. That way scripts can retain the wrapped string and only ever have to find the wrap points once.
2. Add some optimizations to `split_str()`. We need to split on newlines, yes, but after that It should probably check the length of the full line first as an early out. If a line doesn't fit, we can probably make a guess and jump in the middle rather than always iterating from the start. Calling `gr_get_string_size()` on every character is the expensive part and we should avoid that as much as possible.

It'd probably be worthwhile to check if using SCP_String in `split_str()` instead of char[] would work out, because it could certainly make the code cleaner. I can easily script a pathological test case to ensure we're not leaving performance on the table for it.

MageKing17:
> provide a non-wrapping way to draw strings

This is literally just "call `gr.drawString()` without an `x2` argument".

> If a line doesn't fit, we can probably make a guess and jump in the middle rather than always iterating from the start.

Intuitively, it seems like a binary search would be a good choice here, since repeatedly dividing the search space in half *should* be more efficient than iterating character by character, unless the break point is *extremely* close to the beginning of a very long string (which is the kind of worst-case you can't really optimize for, since wrapping that would pretty much be slow no matter what).

EatThePath:
> This is literally just "call gr.drawString() without an x2 argument".

Ah, the reason that didn't sink in the first several times you said it is that I figured if you didn't provide an x2 it'd wrap at the edge of the screen, but I can't find any evidence of that being the case, so my bad.

Agreed on the binary search.

PR #5604, opened 2023-09-16
small speedup to drawString

Goober5000:
Since `gr_string` can accept a length parameter, there's no need to copy each line before it is drawn.

Related to #5495 but not one of the two options mentioned.

### DIFF
--- a/code/scripting/api/libs/graphics.cpp
+++ b/code/scripting/api/libs/graphics.cpp
@@ -1402,17 +1402,8 @@ static int drawString_sub(lua_State *L, bool use_resize_arg)
 		int curr_y = y;
 		for(int i = 0; i < num_lines; i++)
 		{
-			//Contrary to WMC's previous comment, let's make a new string each line
-			int len = linelengths[i];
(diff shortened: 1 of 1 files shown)


code/scripting/api/libs/graphics.cpp
```diff
@@ -1402,17 +1402,8 @@ static int drawString_sub(lua_State *L, bool use_resize_arg)
...
-			char *buf = new char[len+1];
-			strncpy(buf, linestarts[i], len);
-			buf[len] = '\0';
-
 			//Draw the string
-			gr_string(x,curr_y,buf,resize_mode);
-
-			//Free the string we made
-			delete[] buf;
+			gr_string(x,curr_y,linestarts[i],resize_mode,linelengths[i]);
```
wookieejedi, line 1406: Pretty minor, but might be a good chance to add some whitespace separation since functions above and below have it and this is now just the 1 line

Goober5000: Okay

wookieejedi (approved):

PR #5605, opened 2023-09-16
More message filters

naomimyselfandi:
This PR introduces a few enhancements to the dynamic message system:

* Message filters can look at a ship's IFF.
* A dynamic message can be filtered based on the presence of other ships in the mission, and can do so with a range check. Alongside an IFF filter, this can be used to play different versions of a message when the speaker is "in battle" or not.
* The tiebreaker system can be further customized, for cases where two messages have filters but the author wants one to beat the other. (Maybe one of your pilots has a special arrival message for HoL ships, and another for enemy bombers. Currently, when this pilot announces HoL bombers, they'll pick one of those two at random. This feature lets you pick one message or the other.)
* Lastly, as a QoL enhancement, when a message is assigned a mood, it can automatically exclude all other moods. (This could already be done manually, but it gets annoying when you add more moods...)

All of these features are desired for BTA's upcoming release.

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -45,6 +45,7 @@
 bool Allow_generic_backup_messages = false;
 float Command_announces_enemy_arrival_chance = 0.25;
 
+#define DEFAULT_MOOD 0
 SCP_vector<SCP_string> Builtin_moods;
 int Current_mission_mood;
 
@@ -426,6 +427,16 @@ void handle_legacy_backup_message(MissionMessage& msg, SCP_string wing_name) {
 	strcpy(msg.name, backup);
 }
 
+int lookup_mood(SCP_string& name) {
+	for (auto i = Builtin_moods.begin(); i != Builtin_moods.end(); ++i) {
+		if (*i == name) {
+			return (int) std::distance(Builtin_moods.begin(), i);
+		}
+	}
+	Warning(LOCATION, "Message.tbl has an entry for mood type %s, but this mood is not in the #Moods section of the table.", name.c_str());
+	return -1;
+}
+
 // parses an individual message
 void message_parse(MessageFormat format) {
 	MissionMessage msg;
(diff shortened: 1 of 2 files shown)


code/mission/missionmessage.cpp
```diff
@@ -403,6 +427,16 @@ void handle_legacy_backup_message(MissionMessage& msg, SCP_string wing_name) {
 	strcpy(msg.name, backup);
 }
 
+int lookup_mood(SCP_string& name) {
```
BMagnu, line 430: We should be sticking to making things like this name parameter const as often as we can

code/mission/missionmessage.cpp
```diff
@@ -486,45 +520,31 @@ void message_parse(MessageFormat format) {
...
-		if (!found) {
-			// found a mood, but it's not in the list of moods at the start of the table
-			Warning(LOCATION, "Message.tbl has an entry for mood type %s, but this mood is not in the #Moods section of the table.", buf.c_str()); 
-		}
-	}
-	else {
-		msg.mood = 0;
+		SCP_string buf;
+		stuff_string(buf, F_NAME);
+		msg.mood = lookup_mood(buf);
```
BMagnu, line 527: If a mood isn't found, msg.mood will now be set to -1 instead of 0. Which in turn means that erroneous moods for messages caused the messages to be boosted when the mood _was_ actually 0.
While I doubt that that has been taken advantage off in an existing mod, it should probably at least get a comment here acknowledging that change, in case anyone comes looking in the future.

code/mission/missionmessage.cpp
```diff
@@ -2045,10 +2086,30 @@ bool filters_match(MessageFilter& filter, ship* it) {
...
 	}
 }
 
+bool outer_filters_match(MessageFilter& filter, int range, ship* sender) {
+	for (auto i: list_range(&Ship_obj_list)) {
+		auto obj = &Objects[i->objnum];
+		if (obj->flags[Object::Object_Flags::Should_be_dead]) {
+			continue;
+		}
+		if ((range >= 0) && (vm_vec_dist(&obj->pos, &Objects[sender->objnum].pos) > range)) {
```
BMagnu, line 2100: \>= ?
I mean, I guess an uninitialized range is set to -1, but is there ever a case where range == 0 is sensible?

BMagnu (changes requested):
A few small comments and questions, but looking good overall

code/mission/missionmessage.cpp
naomimyselfandi: Since the sender itself is ignored (`if (sender->objnum == i->objnum) { continue; }`), the only way a range of 0 could ever be satisfied is with non-colliding objects in exactly the same position. I don't think it matters in practice, but `> 0`  is probably clearer, so I'm happy to make that change.

code/mission/missionmessage.cpp
naomimyselfandi: This is somewhere where my C++ is a little lacking. `SCP_string const &` means "a reference to an SCP_string which we promise not to change" and potentially allows for compiler optimizations, right? What about declaring the function itself `const`? (To be clear, I'm asking from an educational perspective ^^ )

code/mission/missionmessage.cpp
naomimyselfandi: Would it? Looking at the code on `master`, I'm not sure that `msg.mood` is actually initialized on a missing mood. Isn't that UB?

code/mission/missionmessage.cpp
BMagnu: Yes, basically.
``SCP_string const &`` gives you a reference to a string you _cannot_ change. The compiler will stop you from any things that would change the string, so not only allows it for more optimization, but also is simply a safeguard for you as the developer to not do things you shouldn't.
Declaring the function const in this context is not possible. You could declare the return type const (``const int lookup_mood``), but declaring the _function_ const (``int lookup_mood(...) const {``) is only possible if it's a member function of a class (which it isn't, as a global function), and will basically mean "disallow any modification of member variables", which allows you to call the function on an object of that class which is const itself.

code/mission/missionmessage.cpp
BMagnu: True, turns out I was confusing things, and the default-initialization implicitly causing a zero-initialization is only correct for variables with a static lifetime

BMagnu (dismissed):

code/mission/missionmessage.cpp
```diff
@@ -2045,8 +2086,32 @@ bool filters_match(MessageFilter& filter, ship* it) {
...
+		    && filter_matches(Ship_info[it->ship_info_index].class_type, filter.type_bitfield)
+		    && filter_matches(it->team, filter.team_bitfield);
+	}
+}
+
+bool outer_filters_match(MessageFilter& filter, int range, ship* sender) {
+	for (auto i: list_range(&Ship_obj_list)) {
+		auto obj = &Objects[i->objnum];
+		// Ignore dying ships
+		if (obj->flags[Object::Object_Flags::Should_be_dead]) {
```
TRBlount, line 2098: Bit of a nitpick here, but `Object::Object_Flags::Should_be_dead` is distinct from `Object::Object_Flags::Dying` and the comment seems to imply that only dying ships are skipped. I think the intent here is possibly to skip checks for both Dead and Dying ships, in which case `Object::Object_Flags::Dying` should also be added here and the comment updated to reflect both scenarios.

code/mission/missionmessage.cpp
```diff
@@ -61,6 +63,16 @@ builtin_message Builtin_messages[] = {
   #undef X
 };
 
+#define BUILTIN_BOOST_LEVEL_ONE  1
+#define BUILTIN_BOOST_LEVEL_TWO  2
+#define BUILTIN_MATCHES_TYPE     4
+#define BUILTIN_MATCHES_FILTER   8
+#define BUILTIN_MATCHES_MOOD    16
+#define BUILTIN_MATCHES_SPECIES 32
+#define BUILTIN_MATCHES_PERSONA 64
```
TRBlount, line 72: Aside from the new BOOST_LEVEL entries, the BUILTIN_MATHES_* constants were originally static const int. While I'm fine with them either way, there may be some additional safety by making them `constexpr` instead of a macro, in addition to being easier to work with in debuggers.

code/mission/missionmessage.cpp
```diff
@@ -403,6 +427,16 @@ void handle_legacy_backup_message(MissionMessage& msg, SCP_string wing_name) {
 	strcpy(msg.name, backup);
 }
 
+int lookup_mood(SCP_string const& name) {
+	for (auto i = Builtin_moods.begin(); i != Builtin_moods.end(); ++i) {
+		if (*i == name) {
```
TRBlount, line 432: Should mood lookups be matched case-insensitively? If that is the case, then using `lcase_equal` as they are SCP_strings.

code/mission/missionmessage.cpp
naomimyselfandi: Done.

code/mission/missionmessage.cpp
naomimyselfandi: Done.

TRBlount (approved):

Issue #5601, opened 2023-09-15
Bug: deteriorating performance in FSblue missions


skdKitsune:
As requested my full post from Discord:

"For a few weeks now, I've had severe performance problems in BPC and my own mod, which is heavily influenced by BPC and its graphical and gameplay elements. The problem is that most missions run at a solid 116fps (capped) until something specific happens (tinfoil hat theory is that it has something to do with mission goals, but I'm not sure anymore).

I've been using "loop 1-1 Rebels & Renegades" as my benchmark mission in both my own mod and BPC, since I've got 100% reproducable performance drops there. (I'm only using the in-engine fps overlay, since nvidia and rivatuner overlays eat performance in FS2)
The first happens after I scanned the Hinton; rock-steady 116 until that, then fluctuating down to 110.
The second happens when in skirmish with the Yaaru and vasudan craft; dropping down into the low 100's and somewhat recovering after every enemy craft is dead.
The third drop happens on cue 100%, when the Iceni says "All units, defend the Iceni and the Hinton." via the event "Guard Us" and drops my framerate into the 90's
After that, my framerate deteriorates into the high 60's for the remainder of the mission.
Another thing: once I die, my framerate immediately stabilizes to my locked 116fps, no matter what's still on-screen for the duration of my death screen.

Now, just for fun, I've tried to see how the normal MediaVPs mission plays out (only tested once for now) and the performance was perfectly fine at mostly locked 116fps. The 100% reproducable drops from BPC and my own mod were not present."

skdKitsune:
Relevant videos with frametime that show my issue, recorded with Shadowplay (not sure if I am allowed to post youtube links here):

https://www.youtube.com/watch?v=xtwR9OitxOw /MVPS (no or very slight performance deterioration)

https://www.youtube.com/watch?v=k-XtPNKePU4 /my own mod based on FSblue (BPC has the exact same performance problems)

skdKitsune:
UPDATE: the backlog of axem's message script seems to be the culprit upon further testing on my end.
I am losing 2-15 frames per line of created backlog. 
This framerate drop gets mostly remedied by turning off the HUD, but returns upon turning it on again.

MjnMixael:
Going to close this. Feel free to re-open if the issue occurs again.

PR #5607, opened 2023-09-16
LuaDrawstring optimization

EatThePath:
in draft due to lack of testing and likely need to relocate the code.

Implements the ideas in #5495 and should address #5601, testing pending.

~~The optimizations are implemented in the `gr.drawString()` api function directly rather than applying to `split_str()`. This was done for initially testing the algorithm, as there are around 20 call sites for `split_str()` I wanted to keep the change confined to one targeted area, and I didn't want to think through all the arcane-ass c string contortions in there to find what I could and couldn't cut out.

I think this should definitely go in `split_str()` and see wider testing if the initial testing pans out, though.

I also think that many call sites means it needs to wait for stable before going in, it's just too much potential for chaos at this stage of the RC push.~~

I'm not currently up to trying to apply this to every use case of split_str, and implement every feature it does perfectly, so it exists now as it's own function. Eliminating split_str's current form entirely is something that would be positive for performance, but this covers the one case that has actually been identified as expensive.

There's a fair bit of string copying and potential allocation overhead in this algorithm that might cause some pause, but none of it shows up as remotely worrisome in pathological-case benchmark tests I've done.

### DIFF
--- a/code/parse/parselo.cpp
+++ b/code/parse/parselo.cpp
@@ -15,7 +15,10 @@
 #include <csetjmp>
 
 #include <cctype>
+#include <string>
+#include "globalincs/safe_strings.h"
 #include "globalincs/version.h"
+#include "globalincs/vmallocator.h"
 #include "localization/fhash.h"
 #include "localization/localize.h"
 #include "mission/missionparse.h"
@@ -29,6 +32,8 @@
 #include "utils/encoding.h"
 #include "utils/unicode.h"
 
+#include <stdint.h>
+#include <string.h>
 #include <utf8.h>
 
 using namespace parse;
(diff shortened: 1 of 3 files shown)


EatThePath:
Pushed changes that bring it close to ready for merging. Still some cleanup(extraneous auto-includes, dead comments, etc) and commenting to be done but the core algorithm is fairly reviewable at this stage.

The-E (approved):

code/parse/parselo.cpp
```diff
@@ -3807,6 +3812,108 @@ int split_str(const char *src, int max_pixel_w, SCP_vector<int> &n_chars, SCP_ve
 	return line_num;
 }
 
+// A narrower but much faster alternative to split_str(), takes a string and a max pixel length, returns a vector with
+// one string per line. Does not currently support a max line count or ignoring of characters.
+SCP_vector<SCP_string>
```
Goober5000, line 3817: This should return a collection of offsets and lengths, not a list of strings.  This avoids copying the strings (multiple times, in this particular implementation) or erasing string characters (again, multiple times), therefore improving the algorithm's efficiency.  A vector of `std::pair<size_t, size_t>` would work well.

code/parse/parselo.cpp
```diff
@@ -3807,6 +3812,108 @@ int split_str(const char *src, int max_pixel_w, SCP_vector<int> &n_chars, SCP_ve
...
+{
+	// To avoid any unexpected side effects, we're copying the orignal string.
+	SCP_string new_string = SCP_string(source_string);
+
+	SCP_vector<SCP_string> lines = SCP_vector<SCP_string>();
+
+	while (strip_leading_whitespace && new_string.length() > 0 && is_white_space(new_string[0])) {
+		new_string.erase(0, 1);
+	}
+	if (new_string.length() == 0)
```
Goober5000, line 3828: This can check for `empty()` rather than `length() == 0`.  Ditto elsewhere in this function

Goober5000 (dismissed):
The algorithm looks reasonable, but the efficiency can be improved by modifying how the data structures are handled.

code/parse/parselo.cpp
EatThePath: I think this would be a bad change. In strict terms yes what you describe is the more optimal method, however, the split method was only ever a hot path due to the old method being stupendously wasteful. The overhead these changes would remove accounts for a miniscule amount of frame time, and I have tested unrealistic extreme use cases to confirm that.

Yes, client code of the old method is built around that return type. And frankly, I think that makes the client code harder to read. I think it is an example of excessively, unjustifiably low-level code that makes many parts of the engine harder to work with than needed. I would only want my code to handle strings that way if it was within a type that hid that complexity away.

To put it bluntly, rewriting and retesting this just so that I can go from extreme cases seeing a 5-6ms speedup to, optimistically, a 5.1-6.1ms speedup, is not a use of my time I am interested in.

PR #5608, opened 2023-09-17
move -no_unfocused_pause

Goober5000:
As requested on Discord.  I don't use this option myself, so I welcome any arguments why this should or shouldn't change categories.


Goober5000:
It's from here:
https://discord.com/channels/223511295431933953/372927881048883201/1152747027759910963

I could go either way, but one user missed it, so I thought it was worth thinking about.

JohnAFernandez:
Looking at the other options in the screenshot he provided, I think this is a good move after all.  I would just put it by other window based options.

JohnAFernandez (approved):

PR #5610, opened 2023-09-17
sync setting and getting of collides and no_collide

Goober5000:
The `collides` and `no_collide` flags are equivalent, but opposite.  So when sexps or scripts get or set those flags, allow each to be used as an anti-synonym (not exactly an antonym) for the other.


wookieejedi (approved):
Works in my tests and looks good

PR #5612, opened 2023-09-18
Add nebula-fade-poof SEXP

### LINKED ISSUE #2689: FR: Nebula rendering enhancements & configurability

MjnMixael:
This fixes #2689 by adding a method to do the unfinished task.

`nebula-fade-poof` is different from `nebula-toggle-poof` in a major key way; it does not force a full regeneration of the poofs cloud. The two sexps could probably be merged, but I felt this was a simpler and cleaner approach. Fading poofs in and out really assumes the player continues to move around and offers a more smooth transition between available nebula types.


code/nebula/neb.h
```diff
@@ -73,6 +73,10 @@ typedef struct poof_info {
 	::util::UniformFloatRange rotation;
 	float view_dist;
 	::util::UniformFloatRange alpha;
+	int fade_start;
+	int fade_duration;
+	bool fade_in;
+	float fade_multiplier;
```
Baezon, line 79: The only thing I would say, is to separate these values from the others a bit and comment that they are dynamic values which can vary during a mission, unlike the others which are always completely static,

Baezon (changes requested):
Good work, requiring the camera to move is unideal, but as you say, is much simpler and avoids any 'discontinutities'.

code/nebula/neb.h
MjnMixael: Good idea. Done.

Baezon (approved):

MjnMixael:
Added a few fixes based on testing with Durandal

code/nebula/neb.cpp
```diff
@@ -811,6 +871,17 @@ void upkeep_poofs()
...
+	if (!Neb2_poofs.empty()) {
+		for (size_t i = 0; i < Neb2_poofs.size();) {
+			if (!poof_is_used(Neb2_poofs[i].poof_info_index)) {
+				Neb2_poofs[i] = Neb2_poofs.back();
+				Neb2_poofs.pop_back();
+			} else // if we needed to cull we should not advance because we just moved a new poof into this spot
+				i++;
+		}
+	}
+
```
Baezon, line 884: This should be moved into the above logic ( if too far || !is_used(..) cull it ). There can be an indefinite number of poofs so this is a potential performance concern. I *think* doing it that way is faster, but more importantly I think it's a bit clearer to have a single pass for culling.

PR #5618, opened 2023-09-21
Fred cutscene dialog

MjnMixael:
Adds a dialog editor for managing cutscene sexps. Also deprecates the one alias that was added since the dialog editor explains each option, making it no longer necessary to have such aliases.


fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,466 @@
...
+// Initialization: sets up internal working copy of mission cutscenes and cutscene trees.
+void CMissionCutscenesDlg::load_tree()
+{
+	m_cutscenes_tree.select_sexp_node = select_sexp_node;
+	select_sexp_node = -1;
+
+	m_cutscenes_tree.clear_tree();
+	m_cutscenes.clear();
+	m_sig.clear();
+	for (int i=0; i<(int)The_mission.cutscenes.size(); i++) {
```
JohnAFernandez, line 130: c++ cast

fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,466 @@
...
+}
+
+// create the CTreeCtrl tree from the cutscene tree, filtering based on m_display_cutscene_types
+void CMissionCutscenesDlg::create_tree()
+{
+	m_desc = _T(cutscene_descriptions[m_display_cutscene_types]);
+
+	m_cutscenes_tree.DeleteAllItems();
+	m_cutscenes_tree.reset_handles();
+	for (int i = 0; i < (int)m_cutscenes.size(); i++) {
```
JohnAFernandez, line 152: Cast

fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,466 @@
...
+{
+	UpdateData(TRUE);
+	create_tree();
+}
+
+// New tree item selected.  Because cutscene info is displayed for the selected tree item,
+// we need to update the display when this occurs.
+void CMissionCutscenesDlg::OnSelchangedCutscenesTree(NMHDR* pNMHDR, LRESULT* pResult)
+{
+	NM_TREEVIEW* pNMTreeView = (NM_TREEVIEW*)pNMHDR;
```
JohnAFernandez, line 176: I would try a reinterpret cast here, but I'm not going to require it.  I know this comes with the territory of FRED code.

fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,466 @@
...
+
+	HTREEITEM h2;
+	m_cutscenes_tree.update_help(h);
+	while ((h2 = m_cutscenes_tree.GetParentItem(h)) != 0) {
+		h = h2;
+	}
+
+	int z = (int)m_cutscenes_tree.GetItemData(h);
+	int i;
+	for (i = 0; i < (int)m_cutscenes.size(); i++) {
```
JohnAFernandez, line 190: Two more casts

fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,466 @@
...
+		if (m_cutscenes[i].formula == z) {
+			break;
+		}
+	}
+
+	Assert(i < (int)m_cutscenes.size());
```
JohnAFernandez, line 196: Assertions, especially for original code, are a better choice than Asserts.  It's always good to have a message that encourages reporting of a bug, and we can also pass the value in the message so that we have a least a little better idea what's going on.

fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,466 @@
...
+
+	UpdateData(FALSE);
+	GetDlgItem(IDC_CUTSCENE_TYPE_DROP) -> EnableWindow(TRUE);
+	GetDlgItem(IDC_CUTSCENE_NAME) -> EnableWindow(TRUE);
+
+	UpdateData(FALSE);
+}
+
+// handler for context menu (i.e. a right mouse button click).
+void CMissionCutscenesDlg::OnRclickCutscenesTree(NMHDR* pNMHDR, LRESULT* pResult)
```
JohnAFernandez, line 227: Looks like pNMHDR is an unused parameter

fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,466 @@
...
+			}
+
+			Assert(i < (int)m_cutscenes.size());
+			m_cutscenes.erase(m_cutscenes.begin() + i);
+			m_sig.erase(m_sig.begin() + i);
+
+			return node;
+
+		default:
+			Int3();
```
JohnAFernandez, line 342: Casts, Assertion instead of Assert, and Int3()

You can use UNREACHABLE() instead of Int3(), which allows a message to be displayed to the user, like Assertion does.

fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,466 @@
...
+		return;
+	}
+
+	UpdateData(TRUE);
+	UpdateData(TRUE);  // doesn't seem to update unless we do it twice..
+
+	m_cutscenes[cur_cutscene].type = m_cutscene_type;
+
+	HTREEITEM h = m_cutscenes_tree.GetSelectedItem();
+	Assert(h);
```
JohnAFernandez, line 360: Assertion, also we need an early return if h is NULL/nullptr, I would think.  I can't find a definition for GetParentItem, though.

fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,466 @@
...
+	int i;
+
+	for (i = 0; i < (int)m_cutscenes.size(); i++) {
+		if (m_cutscenes[i].formula == old) {
+			break;
+		}
+	}
+
+	Assert(i < (int)m_cutscenes.size());
+	m_cutscenes[i].formula = node;
```
JohnAFernandez, line 426: Casts and assertion

fred2/missioncutscenesdlg.cpp
```diff
@@ -0,0 +1,466 @@
...
+	}
+	while (index1 > index2 + offset + 1) {
+		m_cutscenes[index1] = m_cutscenes[index1 - 1];
+		m_sig[index1] = m_sig[index1 - 1];
+		index1--;
+	}
+
+	m_cutscenes[index1] = fu;
+	m_sig[index1] = s;
+}
```
JohnAFernandez, line 466: There are a couple more casts and asserts in this function.  Sorry ....

JohnAFernandez (changes requested):

fred2/missioncutscenesdlg.cpp
MjnMixael: Hmm, yes I think you are right. A return instead of an assertion is what this should be.

fred2/missioncutscenesdlg.cpp
MjnMixael: I tried removing it, but it seems like it's something that ON_NOTIFY expects to have. Build failed without it.

fred2/missioncutscenesdlg.cpp
MjnMixael: No problem. I think I got them all. Converted most to size_t, used a few static_casts where needed. Updated asserts to assertions, etc. Let me know if there's anything I missed.

fred2/missioncutscenesdlg.cpp
JohnAFernandez: Ok, no problem

JohnAFernandez (approved):

PR #5622, opened 2023-09-23
SCPUI Options Localization

MjnMixael:
Adds localization support to the SCPUI options builders. Because these are constructed before the localization system is initialized this includes a new method Lafiel and I came up with to set strings to XSTRs but delay their processing until localization has been initialized. I searched the code and the last built-in xstr I could find was 1673. But to be sure I started these at 1700. I know at least BtA uses 1674+ a few for Custom Controls. Since #5210 mapified the built-in strings, it's not essential that they be sequential anyway. Where possible I used existing options xstr IDs from the retail strings.tbl.

Additionally this cleans up all the options builders throughout the files to have consistent whitespace for readability.


MjnMixael:
For completeness sake, and safekeeping, here's a list of all the built-in XSTRs this adds.
[new_xstrs.txt](https://github.com/scp-fs2open/fs2open.github.com/files/12705572/new_xstrs.txt)

code/localization/localize.cpp
```diff
@@ -481,6 +482,35 @@ void parse_tstringstbl(const char *filename)
...
 }
 
+struct xstr_delayed_order {
+	SCP_string& toFill;
+	const char* name;
+	int xstr;
+};
+
+static void lcl_delayed_xstr_internal(tl::optional<xstr_delayed_order> to_init) {
+	static SCP_vector<xstr_delayed_order> delayed_init;
```
BMagnu, line 492: Just as a point of clarification before anyone asks:
``delayed_init`` is very explicitly a static function variable as making it a global variable runs into issues of the order in which static globals are initialized, and it not being guaranteed to be initialized before the (also static global) option variables need to request a delayed XSTR. As a static function variable however, it is always guaranteed to be initialized by the time this function is called.
Yes, it does require this weird calling convention of fill the optional to add a value, keep it empty to fulfill the delayed XSTRs, but since this is an exclusively hidden part of the API, with any user just seeing the ``lcl_delayed_xstr`` function, it should be fine.

code/globalincs/systemvars.cpp
```diff
@@ -245,57 +245,89 @@ detail_levels Detail_defaults[NUM_DEFAULT_DETAIL_LEVELS] = {
...
-                                                                  { 3, "High" },
-                                                                  { 4, "Ultra" }, };
-
-const auto ModelDetailOption __UNUSED =
-	options::OptionBuilder<int>("Graphics.Detail", "Model Detail", "Detail level of models").importance(8).category(
-		"Graphics").values(DetailLevelValues).default_val(MAX_DETAIL_LEVEL).change_listener([](int val, bool) {
-		Detail.detail_distance = val;
-		return true;
-	}).finish();
+const SCP_vector<std::pair<int, std::pair<const char*, int>>> DetailLevelValues = {{ 0, {"Minimum", 1780}},
```
z64555, line 248: This is probably a late observation, but usage of an int-string pair is an odd choice to associate an int value with a string when the position of the strings in its containing vector can suffice.  If this is a limitation of the ::values member and mechanisms using it, then please disregard.

MjnMixael: That is waaaay outside the scope of this PR and something part of the initial design of the OptionsBuilder Values that m!m wrote back in 2019. I'm not going to touch that here.

code/localization/localize.cpp
```diff
@@ -62,7 +63,7 @@ bool *Lcl_unexpected_tstring_check = nullptr;
 // NOTE: with map storage of XSTR strings, the indexes no longer need to be contiguous,
 // but internal strings should still increment XSTR_SIZE to avoid collisions.
 // retail XSTR_SIZE = 1570
-// #define XSTR_SIZE	1675
+// #define XSTR_SIZE	1798
```
TRBlount, line 66: After examining the provided `new_xstrs.txt` and #5627, the new strings should start at 1675 and not 1700. The logic here is that all built-in XSTR IDs should be contiguous. The new value of XSTR_SIZE should be 1774 and the new XSTRs should go from 1675 to 1773.

TRBlount (changes requested):

code/localization/localize.cpp
MjnMixael: You're not concerned about bumping into mods using this IDs for something? Also note that XSTR_SIZE define is just a comment now. I can still update the comment, but the define is technically gone.

TRBlount: The logic here is actually related to collisions, but sort of reversed. The engine shouldn't have to avoid collisions with mods using XSTRs, mods should be avoiding collisions with the engine. The principle is similar to port allocations for TCP/UDP where lower ports (IDs) are reserved for specific uses and applications (mods) should avoid colliding with reserved regions.

TRBlount (approved):

TRBlount:
Not sure why but there is suddenly a merge conflict.

MjnMixael:
> Not sure why but there is suddenly a merge conflict.

Resolved

code/graphics/2d.cpp
```diff
@@ -162,17 +165,18 @@ static bool mode_change_func(os::ViewportState state, bool initial)
...
-								   .importance(98)
-								   .default_val(os::ViewportState::Fullscreen)
-								   .change_listener(mode_change_func)
-								   .finish();
+static auto WindowModeOption __UNUSED = options::OptionBuilder<os::ViewportState>("Graphics.WindowMode",
+                     std::pair<const char*, int>{"Window Mode", 1772},
+                     std::pair<const char*, int>{"Controls how the game window is created", 1773})
+                     .category("Graphics")
+                     .level(options::ExpertLevel::Beginner)
+                     .values({{os::ViewportState::Fullscreen, {"Fullscreen", 1674}},
```
TRBlount, line 173: Sorry, it looks like I missed this earlier, but isn't ID 1674 already reserved for "Nothing"?

TRBlount (approved):
Ok, now we're good.

PR #5623, opened 2023-09-23
fix macOS build packaging issues

notimaginative:
- handle double archiving of macOS builds so they'll work properly with Knossos
 - fix issue where only a single configuration was included in the package (Release or FastDebug, not both)

### DIFF
--- a/.github/workflows/build-nightly.yaml
+++ b/.github/workflows/build-nightly.yaml
@@ -288,12 +288,12 @@ jobs:
         run: $GITHUB_WORKSPACE/ci/linux/run_tests.sh
       - name: Package build result
         working-directory: ./build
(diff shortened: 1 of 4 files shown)


notimaginative:
If there is a better way to do this please let me know. Be nice to see this fixed before the next stable release.

z64555:
Are there any issues that this PR addresses? Would be good to link/mention them here so they get closed properly.

Were you able to test the build chain on your fork?

notimaginative:
No open issues on this as far as I'm aware. I've mentioned the problem several times on discord but there really just aren't enough macOS users around to complain about it.

And this has NOT been tested. The resulting build would have to be pushed to nebula to know for sure if it's all fixed. Any new issue introduced here should be limited to macOS builds, but they're almost useless without these fixes anyway.

PR #5629, opened 2023-09-26
add support for predicting arrival warp distance

Goober5000:
This adds the ability to calculate the arrival distance of a ship using the default warp effect.  There is a new option (in both FRED and qtFRED) to display a gray wireframe outline of the ship projected forward to its final location.

Credit to @Baezon for figuring out the math involved.

### DIFF
--- a/code/ship/shipfx.cpp
+++ b/code/ship/shipfx.cpp
@@ -3730,6 +3730,29 @@ int WE_Default::getWarpOrientation(matrix* output)
     return 1;
 }
 
+float shipfx_calculate_arrival_warp_distance(object *objp)
+{
+	Assert(objp != nullptr && objp->type == OBJ_SHIP);
+
+	// c.f. WE_Default::warpStart()
+	float half_length, warping_dist;
+	if (object_is_docked(objp))
+	{
+		half_length = dock_calc_max_semilatus_rectum_parallel_to_axis(objp, Z_AXIS);
+		warping_dist = 2.0f * half_length;
+	}
+	else
+	{
+		warping_dist = ship_class_get_length(&Ship_info[Ships[objp->instance].ship_info_index]);
+		half_length = 0.5f * warping_dist;
+	}
+	float warping_time = shipfx_calculate_warp_time(objp, WarpDirection::WARP_IN, half_length, warping_dist);
+	float warping_speed = warping_dist / warping_time;
+
+	// the total distance is a full length from its current position, plus the time it takes to slow down from its warping speed
+	// TODO
+}
+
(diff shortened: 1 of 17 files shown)


code/ship/shipfx.cpp
```diff
@@ -3726,6 +3726,38 @@ int WE_Default::getWarpOrientation(matrix* output)
     return 1;
 }
 
+float shipfx_calculate_arrival_warp_distance(object *objp)
+{
+	Assert(objp != nullptr && objp->type == OBJ_SHIP);
```
JohnAFernandez, line 3731: Please expand to an assertion

code/ship/shipfx.cpp
```diff
@@ -3726,6 +3726,38 @@ int WE_Default::getWarpOrientation(matrix* output)
...
+	}
+	float warping_time = shipfx_calculate_warp_time(objp, WarpDirection::WARP_IN, half_length, warping_dist);
+	float warping_speed = warping_dist / warping_time;
+
+	// the total distance is a full length from its current position, plus the distance it takes to slow down from its warping speed
+	float decel_time_const = objp->phys_info.forward_decel_time_const;
+	if (Ship_info[Ships[objp->instance].ship_info_index].flags[Ship::Info_Flags::Supercap]) {
+		// super cap style warpins are annoying, one time constant while above their max speed, a different one while slowing down from there
+		float supercap_slowdown_dist = (warping_speed - (objp->phys_info.max_vel.xyz.z + 5.0f))  * SUPERCAP_WARP_T_CONST;
+		float regular_slowdown_dist = (objp->phys_info.max_vel.xyz.z + 5.0f) * decel_time_const;
```
JohnAFernandez, line 3753: Is there any constant we can reference here for the 5.0f, or is it just a fudge factor?

JohnAFernandez (changes requested):
Looks pretty good.  It definitely ended up simpler than what I expected anyway.  Just two quick things.

code/ship/shipfx.cpp
Baezon: Not right now there is not, it is written the same way in physics.

JohnAFernandez: Ok, it would be great to change that to a constant, if for no other reason than 5.0f appears 403 times in the codebase, so at another point in the future it would be hard to track down all the related places.

Baezon: And the place it is used in physics.cpp is none of them, it uses `5`, instead ;)

JohnAFernandez (approved):

PR #5630, opened 2023-09-26
Revert "fix macOS build packaging issues"

notimaginative:
Although this fix works, it had the unintended consequence of causing nebula to reject the nightly builds due to a limit on file name length.


Goober5000 (approved):

notimaginative:
An alternate solution was made so this reversion PR is no longer needed.

PR #5631, opened 2023-09-26
Add waypoint as allowed lua sexp parameter type

MjnMixael:
Adds waypoints `OPF_POINT` as an allowed LuaSEXP parameter type.

### DIFF
--- a/code/parse/sexp/LuaAISEXP.cpp
+++ b/code/parse/sexp/LuaAISEXP.cpp
@@ -9,7 +9,7 @@
 using namespace luacpp;
 
 namespace sexp {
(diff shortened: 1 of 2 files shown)


JohnAFernandez (commented):
Deceptively simple.  The related code is already set up to handle a waypoint?

MjnMixael:
> Deceptively simple. The related code is already set up to handle a waypoint?

`eval_object_ship_wing_point_team()` and related code at LuaSEXP.ccp line 238 should do the whole shebang.

JohnAFernandez (approved):

naomimyselfandi:
> Deceptively simple. The related code is already set up to handle a waypoint?

In addition to Mike's response above, we discovered this issue when we tried to table a Lua AI operator as accepting a `waypoint`, got an error, and used `ship+waypoint` as a workaround. That works great, aside from being able to pass a ship (which doesn't make sense for this operator).

PR #5634, opened 2023-09-27
Fix for improved subsys attack pathing

Baezon:
The target world pos should use `model_local_to_global_point` since it could be at the end of whoever knows how many subobjects parented together.


Goober5000 (approved):
Looks good, although it might be better to call the variable `target_pm` or something so that `pm` is not used to refer to two different models

PR #5636, opened 2023-09-28
Pause and Resume voices

MjnMixael:
Not sure if this ever worked in retail, but it should in 2023. This adds sound methods to pause and resume sound handles which are then used in mission message to pause and resume voices. I've added the appropriate pause/resume methods to all areas expected (essentially all the same places where weapons and audiostreams (music) are paused as well).

### DIFF
--- a/code/sound/sound.cpp
+++ b/code/sound/sound.cpp
@@ -966,6 +966,48 @@ void snd_stop(sound_handle sig)
 }
 
 /**
+ * Pauses a sound
+ *
+ * @param sig handle to sound, what is returned from snd_play()
+ */
+void snd_pause(sound_handle sig)
+{
+	int channel;
+
+	if (!ds_initialized)
+		return;
+	if (!sig.isValid())
+		return;
+
+	channel = ds_get_channel(sig);
+	if (channel == -1)
+		return;
+
+	ds_pause_channel(channel);
+}
+
+/**
+ * Resumes a sound
+ *
+ * @param sig handle to sound, what is returned from snd_play()
+ */
+void snd_resume(sound_handle sig)
+{
+	int channel;
+
+	if (!ds_initialized)
+		return;
+	if (!sig.isValid())
+		return;
+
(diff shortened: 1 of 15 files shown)


code/sound/sound.cpp
```diff
@@ -1145,6 +1187,38 @@ int snd_is_playing(sound_handle sig)
...
+// snd_is_paused()
+//
+// Determine if a sound is paused
+//
+// returns:			1				=> sound is currently paused
+//						0				=> sound is not paused
+//
+// parameters:		sig	=> signature of sound, what is returned from snd_play()
+//
+int snd_is_paused(sound_handle sig)
```
JohnAFernandez, line 1200: So, since you're writing new functions, these should really just return bools.  Not that there's really any chance of this, but it's safer to return only either true or false and not accidentally something that the calling code is not equipped to handle.

code/sound/sound.cpp
```diff
@@ -1145,6 +1187,38 @@ int snd_is_playing(sound_handle sig)
...
+{
+	int channel, is_paused;
```
JohnAFernandez, line 1202: I know other reviewers say to "match the formatting of the surrounding code", but imho, variables should be declared as close to their assignment as possible, even if the surrounding code is not doing that.  I won't insist on this change, but it would fewer lines.

JohnAFernandez (changes requested):
The actual feature seems to be straightforward enough.  Just a couple notes.

code/sound/sound.cpp
MjnMixael: I can do that, it's really simple. It will be the only functions in the entire sound code that work that way, though.

MjnMixael: Done

code/sound/sound.cpp
MjnMixael: and Done

JohnAFernandez (approved):
Thanks!

PR #5638, opened 2023-09-28
Player general orders

MjnMixael:
Here is player "general" orders. This new feature leverages LuaAI's foundation to be able to create new orders on the Comm Board that the player can select. These orders do not apply to a ship, however. They simply run arbitrary Lua script with an optional `target` parameter. `Ai_helper.Ship` will always be nil here.

You can define an order and set it to an arbitrary category that will show on the top-level Comm Board menu. Orders are then grouped by category.

General orders can be enabled or disabled. They are disabled by default on mission load. Disabled means they will not be visible at all on the Comm Board.

General orders can be valid or invalid. They are invalid by default on mission load. Invalid orders will be visible but greyed out on the Comm Board.

Mission file parsing and saving of default enabled/validated orders is included with this PR. A UI interface is planned and will look very similar to the Allowed Orders dialog in the Ship Editor. However I have a several PRs already that upgrade the Mission Specs dialog (where this will go) that I want to wait for a merge to avoid merge conflicts.

This does also come with sexps to enable/disable and validate/invalidate orders on the fly during a mission. That will suffice for now until I can get the UI dialog created.


code/ai/ailua.cpp
```diff
@@ -48,6 +48,149 @@ const player_order_lua* ai_lua_find_player_order(int sexp_op) {
...
+
+	return count;
+}
+
+SCP_vector<SCP_string> ai_lua_get_general_order_categories(bool enabled_only)
+{
+	SCP_vector<SCP_string> list;
+
+	for (auto order : Lua_player_orders) {
+		SCP_string cat = order.second.category;
```
BMagnu, line 68: Similarly, this should be a const SCP_string&, since you never modify it.

code/ai/ailua.cpp
```diff
@@ -48,6 +48,149 @@ const player_order_lua* ai_lua_find_player_order(int sexp_op) {
...
+		}
+	}
+
+	return list;
+}
+
+SCP_vector<SCP_string> ai_lua_get_enabled_general_orders_by_category(SCP_string cat) {
+	SCP_vector<SCP_string> list;
+
+	for (auto order : Lua_player_orders) {
```
BMagnu, line 103: Same here

code/ai/ailua.cpp
```diff
@@ -48,6 +48,149 @@ const player_order_lua* ai_lua_find_player_order(int sexp_op) {
...
+			if (cat == list[i]) {
+				break;
+			}
+		}
+	}
+
+	return list;
+}
+
+SCP_vector<SCP_string> ai_lua_get_enabled_general_orders_by_category(SCP_string cat) {
```
BMagnu, line 100: As this is also is just a read-only thing, it should be a const SCP_string& as well

code/ai/ailua.cpp
```diff
@@ -48,6 +48,149 @@ const player_order_lua* ai_lua_find_player_order(int sexp_op) {
...
+	SCP_vector<SCP_string> list;
+
+	for (auto order : Lua_player_orders) {
+		if (order.second.cur_valid) {
+			list.push_back(order.second.parseText);
+		}
+	}
+
+	return list;
+}
```
BMagnu, line 147: Frankly, these four should all be combined into one function, that takes three parameters:
bool onlyEnabled
bool onlyValid
const SCP_string& category (if empty, any category)

It'd cut down on a lot of code repetition and functions in the namespace

code/ai/ailua.cpp
```diff
@@ -48,6 +48,149 @@ const player_order_lua* ai_lua_find_player_order(int sexp_op) {
...
+
+int ai_lua_find_general_order_id(SCP_string name) {
+	for (auto order : Lua_player_orders) {
```
BMagnu, line 150: Both of these should be const&'s

code/ai/ailua.cpp
```diff
@@ -67,8 +214,13 @@ void run_ai_lua_action(const luacpp::LuaFunction& action, const ai_mode_lua& lua
 	auto retVals = action.call(Script_system.GetLuaSession(), luaParameters);
 
 	if (!retVals.empty() && retVals[0].getValueType() == luacpp::ValueType::BOOLEAN) {
-		if (retVals[0].getValue<bool>())
-			ai_mission_goal_complete(aip);
+		if (retVals[0].getValue<bool>()) {
+
+			// If we don't have a ship then it's a general order and we have nothing else to do
+			if (aip->shipnum < 0) {
```
BMagnu, line 220: Isn't this the wrong way?
We only need to complete the goal, if the shipnum is larger or equal to 0 and valid?

code/hud/hudsquadmsg.cpp
```diff
@@ -1630,13 +1631,38 @@ void hud_squadmsg_type_select( )
...
+	int lua_order_count = 0;
+
+	// Now get a list of all lua categories to add. Meow.
+	SCP_vector<SCP_string> lua_cat_list = ai_lua_get_general_order_categories();
+
+	// If we have too many categories then we have an oopsie!
+	if ((int)lua_cat_list.size() > (MAX_MENU_ITEMS - NUM_COMM_ORDER_TYPES)) {
+		Warning(LOCATION,
+			"Too many defined Lua General Order Categories! Any categories beyond %i will be ignored.",
+			MAX_MENU_ITEMS - NUM_COMM_ORDER_TYPES);
```
BMagnu, line 1645: Is it at all possible to validate this during parsing and not during display?

code/hud/hudsquadmsg.cpp
```diff
@@ -1667,6 +1693,13 @@ void hud_squadmsg_type_select( )
 	MsgItems[TYPE_REPAIR_REARM_ITEM].active = 1;				// this item will always be available (I think)
 	MsgItems[TYPE_REPAIR_REARM_ABORT_ITEM].active = 0;
 
+	for(auto cat : lua_cat_list){
```
BMagnu, line 1696: should be const auto&

code/hud/hudsquadmsg.cpp
```diff
@@ -2078,6 +2114,60 @@ void hud_squadmsg_ship_command()
...
+
+		//If it's not a general order then do not add it.
+		if (!lua_porder->generalOrder) {
+			continue;
+		}
+
+		//If it's not part of the selected category then do not add it.
+		if (lua_porder->category != Lua_sqd_msg_cat) {
+			continue;
+		}
```
BMagnu, line 2139: This is player_order number of string comparisons each frame.
Not pretty, but it _might_ be sufficient for now.

fred2/missionsave.cpp
```diff
@@ -4148,6 +4149,38 @@ int CFred_mission_save::save_players()
...
+
+	if (Mission_save_format != FSO_FORMAT_RETAIL && (e_list.size() > 0)) {
+		if (optional_string_fred("+General Orders Enabled:", "#Players"))
+			parse_comments();
+		else
+			fout("\n+General Orders Enabled:");
+
+		fout(" (");
+
+		for (SCP_string order : e_list) {
```
BMagnu, line 4162: const&

fred2/missionsave.cpp
```diff
@@ -4148,6 +4149,38 @@ int CFred_mission_save::save_players()
...
+
+	if (Mission_save_format != FSO_FORMAT_RETAIL && (v_list.size() > 0)) {
+		if (optional_string_fred("+General Orders Valid:", "#Players"))
+			parse_comments();
+		else
+			fout("\n+General Orders Valid:");
+
+		fout(" (");
+
+		for (SCP_string order : v_list) {
```
BMagnu, line 4178: const&

qtfred/src/mission/missionsave.cpp
```diff
@@ -4142,6 +4143,38 @@ int CFred_mission_save::save_players()
...
+
+	if (save_format != MissionFormat::RETAIL && (e_list.size() > 0)) {
+		if (optional_string_fred("+General Orders Enabled:", "#Players"))
+			parse_comments();
+		else
+			fout("\n+General Orders Enabled:");
+
+		fout(" (");
+
+		for (SCP_string order : e_list) {
```
BMagnu, line 4156: const&

qtfred/src/mission/missionsave.cpp
```diff
@@ -4142,6 +4143,38 @@ int CFred_mission_save::save_players()
...
+
+	if (save_format != MissionFormat::RETAIL && (v_list.size() > 0)) {
+		if (optional_string_fred("+General Orders Valid:", "#Players"))
+			parse_comments();
+		else
+			fout("\n+General Orders Valid:");
+
+		fout(" (");
+
+		for (SCP_string order : v_list) {
```
BMagnu, line 4172: const&

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -5137,6 +5145,19 @@ sexp_list_item* sexp_tree::get_listing_opf_traitor_overrides()
 	return head.next;
 }
 
+sexp_list_item* sexp_tree::get_listing_opf_lua_general_orders()
+{
+	sexp_list_item head;
+
+	SCP_vector<SCP_string> orders = ai_lua_get_all_general_orders();
+
+	for (auto val : orders) {
```
BMagnu, line 5154: const&

fred2/sexp_tree.cpp
```diff
@@ -7536,6 +7544,19 @@ sexp_list_item* sexp_tree::get_listing_opf_traitor_overrides()
 	return head.next;
 }
 
+sexp_list_item* sexp_tree::get_listing_opf_lua_general_orders()
+{
+	sexp_list_item head;
+
+	SCP_vector<SCP_string> orders = ai_lua_get_all_general_orders();
+
+	for (auto val : orders) {
```
BMagnu, line 7553: const&

BMagnu (changes requested):
Needs more const references!
Otherwise some small issues, but generally good to go.
Nice to see LuaAI being expanded in this direction!

code/ai/ailua.cpp
MjnMixael: yes, I believe you are correct

code/hud/hudsquadmsg.cpp
MjnMixael: Just have to move the relevant defines in to the header, I suppose.

code/hud/hudsquadmsg.cpp
MjnMixael: Hmm. Definitely worth some thought and maybe a follow-up.

BMagnu (approved):

wookieejedi:
Will plan to merge later today

PR #5641, opened 2023-09-30
tweak directory copying for frameworks on macOS

notimaginative:
CMake's copy_directory functions follow symlinks recursively as opposed to preserving them. For mac frameworks, which make use of symlinks for versioning, this causes the files to be duplicated multiple times in the copied framework. Using rsync instead, just for frameworks, will properly duplicate the framework while still avoiding the "file exists" error from cp.


notimaginative:
I'm not particularly happy having to do this, but I haven't found a better option. The alternative is allowing bundled frameworks to be 2.5x their normal size for no reason.

Goober5000 (approved):

PR #5642, opened 2023-09-30
Fix null dereference when a subsystem is hit for damage by a null object

BMagnu:
I'm not exactly 100% sure what causes this specific situation, but looking at the function, it seems that getting a null object as the other hit object is allowed.
This fixes the damage carry check, and since not carrying the damage will happen with all non-shockwaves, it'll also happen with null objects. The check in L 958 is safe btw, since weapon_info_index is guaranteed to be -1 for a null other_obj (in fact, checks further up in the function already rely on that)

### DIFF
--- a/code/ship/shiphit.cpp
+++ b/code/ship/shiphit.cpp
@@ -698,7 +698,7 @@ float do_subobj_hit_stuff(object *ship_objp, object *other_obj, vec3d *hitpos, i
 	}
 
 	//	Shockwave damage is applied like weapon damage.  It gets consumed.
-	if (other_obj_is_shockwave)	// Goober5000 check for NULL
+	if (other_obj_is_shockwave)
 	{
 		//	MK, 9/2/99.  Shockwaves do zero subsystem damage on small ships.
 		// Goober5000 - added back in via flag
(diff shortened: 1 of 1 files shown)


Baezon (approved):
Fwiw, this is explicitly the case for the `explosion-effect` sexp, not sure if any others though.

Goober5000 (changes requested):
There is another potential null dereference a few lines further down:
```
// Nuke: this will finally factor it in to damage_to_apply and i wont need to factor it in anywhere after this
damage_to_apply = Armor_types[subsystem->armor_type_idx].GetDamage(damage_to_apply, dmg_type_idx, ss_dif_scale, other_obj->type == OBJ_BEAM);
```

Optional suggestion: consider making a few `other_object_is_X` boolean variables, à la `ship_do_damage`.  This would make object checks a little cleaner throughout `do_subobj_hit_stuff`.

code/ship/shiphit.cpp
```diff
@@ -953,9 +957,9 @@ float do_subobj_hit_stuff(object *ship_objp, object *other_obj, vec3d *hitpos, i
 
 			// if this subsystem doesn't carry damage then subtract it off of our total return
 			if (subsystem->system_info->flags[Model::Subsystem_Flags::Carry_no_damage]) {
-				if ((other_obj == nullptr) || (other_obj->type != OBJ_SHOCKWAVE) || (!(subsystem->system_info->flags[Model::Subsystem_Flags::Carry_shockwave]))) {
+				if (other_obj_is_shockwave || (!(subsystem->system_info->flags[Model::Subsystem_Flags::Carry_shockwave]))) {
```
Goober5000, line 960: This should be `!other_obj_is_shockwave`

Optional: remove the outer pair of parentheses from the second half of the condition to make it more readable

code/ship/shiphit.cpp
```diff
@@ -694,7 +698,7 @@ float do_subobj_hit_stuff(object *ship_objp, object *other_obj, vec3d *hitpos, i
 	}
 
 	//	Shockwave damage is applied like weapon damage.  It gets consumed.
-	if ((other_obj != NULL) && (other_obj->type == OBJ_SHOCKWAVE))	// Goober5000 check for NULL
+	if (other_obj_is_shockwave)	// Goober5000 check for NULL
```
Goober5000, line 701: You could remove the `// Goober5000 check for NULL` part of this line since the null check is now done earlier.

Goober5000 (changes requested):

PR #5645, opened 2023-09-30
Lua Child Enums for Sexp types

MjnMixael:
This creates a new sexp type "child enum" that requires a Lua Enum to be the parent. This allows for creating truly arbitrary groups of data within sexp parameters. For example, a data-saving sexp could have as it's first parameter a list of categories of data to save. The child enum would then display valid values based on the parent list.


qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -5177,6 +5182,43 @@ sexp_list_item* sexp_tree::get_listing_opf_asteroid_debris()
...
 
+sexp_list_item* sexp_tree::get_listing_opf_lua_enum(int parent_node, int arg_index)
+{
+	int this_index = get_dynamic_parameter_index(tree_nodes[parent_node].text, arg_index);
+
+	// first child node
+	int child = tree_nodes[parent_node].child;
+	if (child < 0)
+		return nullptr;
+
```
JohnAFernandez, line 5193: Just a note on code like this. If we were in a gameplay function, it would be best to move the `get_dynamic_parameter_index` function call below the early return check.  Here it basically doesn't matter, unless FRED or qtFRED end up being wildly different than initially designed in the future.

If you were to move it, it would make the could a bit clearer (assignment closer to variable use is generally better for readability), but I am far from requiring that change.

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -5177,6 +5182,43 @@ sexp_list_item* sexp_tree::get_listing_opf_asteroid_debris()
...
+			"Expected to find an enum parent parameter for node %i in operator %s but found nothing!",
+			arg_index,
+			tree_nodes[parent_node].text);
+	}
+
+	sexp_list_item head;
+
+	int item = get_dynamic_enum_position(tree_nodes[child].text);
+
+	if (item > 0 && item < (int)Dynamic_enums.size()) {
```
JohnAFernandez, line 5209: Quick cast change.

qtfred/src/ui/widgets/sexp_tree.cpp
```diff
@@ -5177,6 +5182,43 @@ sexp_list_item* sexp_tree::get_listing_opf_asteroid_debris()
...
+
+	if (this_index >= 0) {
+		for (int count = 0; count < this_index; count++) {
+			child = tree_nodes[child].next;
+		}
+	} else {
+		error_display(1,
+			"Expected to find an enum parent parameter for node %i in operator %s but found nothing!",
+			arg_index,
+			tree_nodes[parent_node].text);
```
JohnAFernandez, line 5202: Just double checking that this doesn't need an early return?

MjnMixael: Honestly, it's been so long since I wrote this code that it's hard to remember for sure. But it does look like it. Added.

JohnAFernandez (approved):

PR #5646, opened 2023-09-30
tweak some model warning messages

Goober5000:
Per a thread in the TBP forum, elaborate on these model warnings a bit.

Two other small code tweaks as well.


JohnAFernandez (approved):
LGTM, it should be noted the two "code tweaks" are just whitespace changes for clarity.

PR #5649, opened 2023-10-01
fix ShipStatus states to correctly account for exploding ships

Goober5000:
Adds a `DEATH_ROLL` status to specifically cover the state after ships have been destroyed but before they exit.  I've also audited all code that interacts with the ship registry to use the correct state with this new state taken into consideration.

This fixes access to exploding ships that was broken when the ship registry framework was introduced.  In particular, the `when destroyed X beam-lock-all X` QoL feature now works again.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -386,7 +386,7 @@ int is_preferred_weapon(int weapon_num, object *firer_objp, object *target_objp)
 			continue;
 
 		auto ship_entry = &Ship_registry[hfi->ship_registry_index];
-		if (ship_entry->status != ShipStatus::PRESENT)
+		if (!ship_entry->objp)
 			continue;
 
 		int signature = ship_entry->objp->signature;
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -6585,6 +6585,7 @@ void eval_object_ship_wing_point_team(object_ship_wing_point_team *oswpt, int no
 				break;
 
 			case ShipStatus::PRESENT:
+			case ShipStatus::DEATH_ROLL:
 				oswpt->type = OSWPT_TYPE_SHIP;
 				break;
 
@@ -6853,7 +6854,7 @@ int sexp_is_destroyed(int n, fix *latest_time)
 			if (ship_entry->status == ShipStatus::NOT_YET_PRESENT)
 				return SEXP_CANT_EVAL;
 
-			if (ship_entry->status == ShipStatus::EXITED)
+			if (ship_entry->status == ShipStatus::DEATH_ROLL || ship_entry->status == ShipStatus::EXITED)
 			{
 				// check the mission log
 				if (mission_log_get_time(LOG_SHIP_DESTROYED, ship_entry->name, nullptr, &time) || mission_log_get_time(LOG_SELF_DESTRUCTED, ship_entry->name, nullptr, &time))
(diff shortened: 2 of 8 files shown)


code/ai/aicode.cpp
```diff
@@ -386,7 +386,7 @@ int is_preferred_weapon(int weapon_num, object *firer_objp, object *target_objp)
 			continue;
 
 		auto ship_entry = &Ship_registry[hfi->ship_registry_index];
-		if (ship_entry->status != ShipStatus::PRESENT)
+		if (!ship_entry->objp)
```
TRBlount, line 389: This change is inconsistent with a similar change in missiontraining.cpp. It should be changed to a direct comparison to `nullptr`.

code/parse/sexp.cpp
```diff
@@ -12335,7 +12346,7 @@ void sexp_add_goal(int n)
 	auto ship_entry = eval_ship(n);
 	if (ship_entry)
 	{
-		if (ship_entry->status != ShipStatus::PRESENT)
+		if (!ship_entry->shipp)
```
TRBlount, line 12349: ditto

code/parse/sexp.cpp
```diff
@@ -12362,7 +12373,7 @@ void sexp_remove_goal(int n)
...
```
TRBlount, line 12376: ditto

code/parse/sexp.cpp
```diff
@@ -12394,7 +12405,7 @@ void sexp_clear_goals(int n)
 		auto ship_entry = eval_ship(n);
 		if (ship_entry)
 		{
-			if (ship_entry->status != ShipStatus::PRESENT)
+			if (!ship_entry->shipp)
```
TRBlount, line 12408: ditto

code/parse/sexp.cpp
```diff
@@ -16225,7 +16236,7 @@ void sexp_deal_with_ship_flag(int node, bool process_subsequent_nodes, Object::O
 			continue;
 
 		// if ship is in-mission
-		if (ship_entry->status == ShipStatus::PRESENT)
+		if (ship_entry->shipp)
```
TRBlount, line 16239: ditto

code/parse/sexp.cpp
```diff
@@ -9412,7 +9421,7 @@ int sexp_last_order_time(int n)
 
 		auto ship_entry = eval_ship(n);
 		if (ship_entry) {
-			if (ship_entry->status == ShipStatus::PRESENT) {
+			if (ship_entry->shipp) {
```
TRBlount, line 9424: ditto

TRBlount (changes requested):

code/ai/aicode.cpp
Goober5000: Forgive me but this sounds like an AI-generated comment.  The change in missiontraining.cpp is the exception here, and if you want this PR to use consistent checks, *that* should be what is changed.  The only reason I didn't use this pattern in missiontraining.cpp is because that already had a comparison to nullptr, but I think it makes more sense to change missiontraining.cpp if any change is needed.

Casting a pointer to a bool is valid, so this is a matter of style.  As for why to prefer this style in this PR, there are quite a few comparisons throughout it, so using the more concise style greatly enhances readability.  They are consistent with pre-existing boolean checks, and they avoid cluttering the condition with mentions of `nullptr` over and over.

TRBlount: Fair enough I suppose.

TRBlount (approved):

PR #5654, opened 2023-10-03
Fix near z planes for the cockpit rendering

BMagnu:
This gets rid of dark, unshaded spheres in the cockpit close to the camera by doing two things:
1. Unify all near-z distances used in cockpit rendering to one common value, and make that a proper variable instead of hardcoding it everywhere.
2. Properly pass that to the deferred lighting shader instead of having a magic constant for closest allowed shading distance in there.

### DIFF
--- a/code/graphics/matrix.cpp
+++ b/code/graphics/matrix.cpp
@@ -16,6 +16,8 @@ matrix4 gr_texture_matrix;
 
 matrix4 gr_env_texture_matrix;
 
+float gr_near_plane;
+
 bool gr_htl_projection_matrix_set = false;
 
 static int modelview_matrix_depth = 1;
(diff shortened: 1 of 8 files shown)


code/graphics/matrix.cpp
```diff
@@ -16,6 +16,8 @@ matrix4 gr_texture_matrix;
 
 matrix4 gr_env_texture_matrix;
 
+float gr_near_plane;
```
Goober5000, line 19: Does this need a default initializer?

The-E (approved):

code/graphics/matrix.cpp
Goober5000: Perhaps not -- according to Lafiel, you cannot render before you set the projection matrix, so it should never be possible to read from it.

PR #5655, opened 2023-10-03
Make version parsing available

BMagnu:
Extracts the version parsing code from the Lua API.
Prerequisite for #5583

### DIFF
--- a/code/globalincs/version.cpp
+++ b/code/globalincs/version.cpp
@@ -75,31 +75,45 @@ version::version(int major_in, int minor_in, int build_in, int revision_in) :
 version::version(const SCP_string& semver, int missing) : major(missing), minor(missing), build(missing), revision(missing) {
 	int i = 0;
 	auto str = semver;
+	int has_fields = 0;
 	while (i <= 4) {
 		size_t pos = str.find_first_of('.');
+		SCP_string ver;
 		if (pos != SCP_string::npos) {
-			auto ver = str.substr(0, pos);
-			if (!ver.empty() && std::find_if(ver.begin(), ver.end(), [](char c) { return !std::isdigit(c, SCP_default_locale); }) == ver.end()) {
-				if (major < 0) {
-					major = std::stoi(str.c_str());
-				}
-				else if (minor < 0) {
-					minor = std::stoi(str.c_str());
-				}
-				else if (build < 0) {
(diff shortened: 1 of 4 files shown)


Goober5000 (commented):
This is a good idea in principle, but there are now two version parsing functions -- this and `parse_version_inline`.  Can they be consolidated?

Goober5000 (commented):
Huh, clever.

Is it possible for a version to be successfully parsed, yet have one of its numbers be -1?  It looks like there might need to be a check at the bottom of `version::version` to guard against this.

BMagnu:
As is now, negative versions are allowed, but will prevent a revision to be parsed if one exists.
Previously, negative versions were also allowed, but would also allow a revision to be parsed.
I'll change it to allow for revision with negative versions

code/globalincs/version.cpp
```diff
@@ -75,31 +75,45 @@ version::version(int major_in, int minor_in, int build_in, int revision_in) :
...
-				else if (build < 0) {
-					build = std::stoi(str.c_str());
-				}
-				str.erase(0, pos + 1);
-			}
-			else if (major < 0) {
-				break; //Break out of the loop if the first string is not a digit.
-			}
+			ver = str.substr(0, pos);
+			str.erase(0, pos + 1);
```
Goober5000, line 84: With the new copy statement a few lines above, I don't think this statement here is correct

code/globalincs/version.cpp
```diff
@@ -75,31 +75,45 @@ version::version(int major_in, int minor_in, int build_in, int revision_in) :
...
+			case 1:
+				minor = std::stoi(ver.c_str());
+				++has_fields;
+				break;
+			case 2:
+				build = std::stoi(ver.c_str());
+				++has_fields;
+				break;
+			case 3:
+				revision = std::stoi(str.c_str());
```
Goober5000, line 106: Does this `str` also need to be `ver`?

BMagnu: done

code/globalincs/version.cpp
Goober5000: Retracted - I misinterpreted what was going on

Goober5000 (approved):

PR #5656, opened 2023-10-03
Fix HUD slewing

BMagnu:
This PR fixes wildly inconsistent HUD slewing behaviour in FSO.
Some gauges just didn't slew at all, some slewed incorrectly, while some were allowed to slew even if they must not slew ever.
This also removes the effect of HUD_offset_x/y, as that seems to be partly redundant with HUD_nose_x/y, and causes HUD gauges to never be anywhere near correct as soon as its nonzero. However, as there may have been plans for it, it's just commented out instead of removed outright.
Prerequisite for #5583.

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -1033,8 +1033,8 @@ void HudGauge::renderCircle(int x, int y, int diameter, bool filled)
 
 void HudGauge::setClip(int x, int y, int w, int h)
 {
-	int hx = fl2i(HUD_offset_x);
-	int hy = fl2i(HUD_offset_y);
+	int hx;
+	int hy;
 
 	if ( gr_screen.rendering_to_texture != -1 ) {
 		gr_set_screen_scale(canvas_w, canvas_h, -1, -1, target_w, target_h, target_w, target_h, true);
@@ -1051,6 +1051,8 @@ void HudGauge::setClip(int x, int y, int w, int h)
 
 		gr_set_clip(hx, hy, w, h);
 	} else {
+		hx = fl2i(HUD_offset_x);
+		hy = fl2i(HUD_offset_y);
 		if (reticle_follow) {
 			hx += HUD_nose_x;
 			hy += HUD_nose_y;
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -91,10 +91,6 @@ color HUD_color_debug;										// grey debug text shown on HUD
 
 static sound_handle Player_engine_snd_loop = sound_handle::invalid();
 
-// HUD render frame offsets
-float HUD_offset_x = 0.0f;
-float HUD_offset_y = 0.0f;
-
 // the offset of the player's view vector and the ship forward vector in pixels (Swifty)
 int HUD_nose_x;
 int HUD_nose_y;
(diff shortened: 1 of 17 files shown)


Goober5000:
In the original public source release, HUD_offset_x/y were used to "jitter stuff", primarily the shield icon, directives, reticle, and target box.  HUD_nose_x/y did not exist.  So my guess is that HUD_nose_x/y was intended to supersede HUD_offset_x/y but did so incorrectly or incompletely.

Here is the function that calculated the offsets:
```C++
// set the offset values for this render frame
void HUD_set_offsets(object *viewer_obj, bool wiggedy_wack)
{
	if ( (viewer_obj == Player_obj) && wiggedy_wack ){		
		vector tmp;
		vertex pt;
		ubyte flags;		

		HUD_offset_x = 0.0f;
		HUD_offset_y = 0.0f;

		vm_vec_scale_add( &tmp, &Viewer_obj->pos, &Viewer_obj->orient.fvec, 100.0f );
		
		flags = g3_rotate_vertex(&pt,&tmp);

		if (flags == 0) {

			g3_project_vertex(&pt);

			if (!(pt.flags & PF_OVERFLOW))	{
				HUD_offset_x -= 0.45f * (i2fl(gr_screen.clip_width)*0.5f - pt.sx);
				HUD_offset_y -= 0.45f * (i2fl(gr_screen.clip_height)*0.5f - pt.sy);
			}
		}

		if ( HUD_offset_x > 100.0f )	{
			HUD_offset_x = 100.0f;
		} else if ( HUD_offset_x < -100.0f )	{
			HUD_offset_x += 100.0f;
		}

		if ( HUD_offset_y > 100.0f )	{
			HUD_offset_y = 100.0f;
		} else if ( HUD_offset_y < -100.0f )	{
			HUD_offset_y += 100.0f;
		}

	} else {
		HUD_offset_x = 0.0f;
		HUD_offset_y = 0.0f;
	}
}
```

Goober5000 (commented):
Code LGTM but want to resolve above conversation.

BMagnu:
My guess is that the offset here is attempting to compensate for the eyepoint moving off-center, but doing so arbitrarily multiplied with a magic constant of 0.45, and clamping an offset to the bottom right, while just adding a weird offset when shifting to the top left.
In addition, only few gauges even used it (as opposed to nose, which is built into the general rendering functions and only rarely needs to be considered manually), and those who did didn't differentiate between when being RTT'd and when not, resulting in utterly broken RTT.
I assume that nose can just supercede it, since at its core, it does the same: Calculate the visual offset of the eyepoint, and compensate the hud position so that it is positioned just as if the HUD's center is mounted on a loooong pike extruding from the ship's center forward.
And since that uses the actual view matrix, it accounts for all manners of actual screenshake, so I'd say we're good.

Goober5000:
Gotcha.  Ok, then it sounds like HUD_offset_x/y should be completely taken out, not just commented.

code/hud/hud.cpp
```diff
@@ -3744,7 +3744,7 @@ int hud_objective_notify_active()
  * @param wiggedy_wack
  * @param eye_orient 
  */
-void HUD_set_offsets(object *viewer_obj, int wiggedy_wack, matrix *eye_orient)
+void HUD_set_offsets()
 {
 	if ( Viewer_mode & ( VM_TOPDOWN | VM_CHASE ) ) {
```
Goober5000, line 3749: It looks like the logic from the old function call
```
int dont_offset = ((Game_mode & GM_MULTIPLAYER) && (Net_player->flags & NETINFO_FLAG_OBSERVER));
```
should be rolled into this condition.

Goober5000 (changes requested):
Actually one more thing...

code/hud/hud.cpp
BMagnu: That condition was never actually applied to nose, only to offset.
And just because you're an observer, doesn't mean that your hud can now suddenly move massively different IMO

Goober5000: Ok, I suppose not.

PR #5659, opened 2023-10-03
allow modification of warp audible range via objecttypes.tbl

Goober5000:
The retail warp range multiplier can yield ridiculous results, like an Elysium warp being audible over 9000 meters away.  So allow the multiplier to be set via ship type.

### DIFF
--- a/code/fireball/fireballs.cpp
+++ b/code/fireball/fireballs.cpp
@@ -61,35 +61,33 @@ static auto WarpOption __UNUSED = options::OptionBuilder<bool>("Graphics.3dWarp"
  */
 void fireball_play_warphole_open_sound(int ship_class, fireball *fb)
 {
-	gamesnd_id sound_index;
-	float		range_multiplier = 1.0f;
-	object	*fireball_objp;	
-		
 	Assert((fb != NULL) && (fb->objnum >= 0));
 	if((fb == NULL) || (fb->objnum < 0)){
 		return;
 	}
-	fireball_objp = &Objects[fb->objnum];
+	auto fireball_objp = &Objects[fb->objnum];
+	auto sound_index = gamesnd_id(GameSounds::WARP_IN);
 
-	if(fb->warp_open_sound_index.isValid()) {
+	if (fb->warp_open_sound_index.isValid()) {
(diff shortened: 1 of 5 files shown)


Baezon (commented):
For whatever reason, `fireball_play_warphole_close_sound` completely lacks this logic and multiplier... perhaps one should be added for closing too?

Goober5000:
Huh, yeah, good catch.

Goober5000:
Modified and fixed a few things, plus handled the multiplier.

code/fireball/fireballs.cpp
```diff
@@ -61,55 +61,56 @@ static auto WarpOption __UNUSED = options::OptionBuilder<bool>("Graphics.3dWarp"
...
+	if (fb->warp_close_sound_index.isValid()) {
 		sound_index = fb->warp_close_sound_index;
 	} else if ( fb->flags & FBF_WARP_CAPITAL_SIZE ) {
 		sound_index = gamesnd_id(GameSounds::CAPITAL_WARP_OUT);
 	} else {
 		return;
 	}
 
-	snd_play_3d(gamesnd_get_game_sound(sound_index), &fireball_objp->pos, &Eye_position, fireball_objp->radius); // play warp sound effect
+	snd_play_3d(gamesnd_get_game_sound(sound_index), &fireball_objp->pos, &Eye_position, fireball_objp->radius, NULL, 0, 1.0F, SND_PRIORITY_SINGLE_INSTANCE, NULL, fb->warp_sound_range_multiplier); // play warp sound effect
```
Baezon, line 113: Well, note that before it would've had 1, we don't want to suddenly 6x multiply the closing sounds for capitals so we might need a separate value for closing as opposed to opening.

Goober5000: That wouldn't happen.  The multiplier was only applied to big ships, excluding capital ships.  And only capital ships play the warp closing sound.  Therefore, the default multiplier here is still 1.

Baezon: Ah, I see

Baezon (approved):

PR #5660, opened 2023-10-03
Adds custom data for intel entries

MjnMixael:
Works exactly like the custom data in ships, weapons, and cutscenes tables.


wookieejedi (approved):
Quite straightforward and the uses are very nice as well.

wookieejedi:
Will plan to merge this tomorrow.

PR #5662, opened 2023-10-04
Intro about to play hook

MjnMixael:
I needed to reliably run a script for SCPUI after all game init methods have been completed but before the game actually moves on to the Intro and other game states. On Splash Screen worked but is deprecated, so this adds a new hook just before the intro that allows me to do what I need.

### DIFF
--- a/freespace2/freespace.cpp
+++ b/freespace2/freespace.cpp
@@ -6653,6 +6653,10 @@ int game_main(int argc, char *argv[])
 		return 0;
 	}
 
+	if (scripting::hooks::OnIntroAboutToPlay->isActive()) {
+		scripting::hooks::OnIntroAboutToPlay->run();
+	}
+
 	if (!Is_standalone) {
 		movie::play("intro.mve");
 	}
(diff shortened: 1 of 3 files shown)


freespace2/freespace.cpp
```diff
@@ -6653,6 +6653,11 @@ int game_main(int argc, char *argv[])
 		return 0;
 	}
 
+	Script_system.RunInitFunctions();
```
TRBlount, line 6656: What's the purpose of this being duplicated from `game_init`?

MjnMixael: Oh rip lol. I'm in a meeting and did not fully read this. That's a copy/paste error. Removed.

freespace2/freespace.cpp
```diff
@@ -1983,7 +1983,6 @@ void game_init()
 
 	game_title_screen_close();
 
-	Script_system.RunInitFunctions();
```
TRBlount, line 1986: I think you removed the wrong one.

TRBlount (changes requested):

freespace2/freespace.cpp
MjnMixael: I'm gonna come back to this after my meeting when I can focus

MjnMixael: Ok, I've done it correctly this time.

BMagnu (approved):

freespace2/freespace.cpp
TRBlount: Completely forgot to finish this review. Looks good now.

TRBlount (approved):

PR #5663, opened 2023-10-05
remove incorrect setting of ship status

### LINKED ISSUE #5661: RC8 and later: Ai wingmen refuse any order at all times
Steps to reproduce:

1. Play a mission where you can order your wingmen
2. target a wingman - note how it has the "no orders" indicator - regardless of whatever order they were given in FRED.
3. order your wingman to guard you
4. notice how very briefly, "guard alpha 1" flashes before the craft resets to "no orders".

Goober5000:
There is a bit of code in missionparse that sets the parse object because in-game joining needs it.  This is unrelated to ship status, so the status shouldn't be changed.  This fixes the bug and adds a comment to clarify the situation, as well as some warnings to help track down incorrect joining states.

Fixes #5661.  Follow-up to #5658 (where the regression was introduced) and #4836 (which fixed in-game joining).

Also adds a ship registry entry for support ships, since in testing that issue was (correctly) flagged by the warnings.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -4269,7 +4269,7 @@ int parse_wing_create_ships( wing *wingp, int num_to_create, bool force_create, 
 			}
 
 			// subsequent waves of ships will not be in the ship registry, so add them
-			if (!ship_registry_get(p_objp->name))
+			if (!ship_registry_exists(p_objp->name))
 			{
 				ship_registry_entry entry(p_objp->name);
 				entry.status = ShipStatus::NOT_YET_PRESENT;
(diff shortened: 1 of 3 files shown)


code/mission/missionparse.cpp
```diff
@@ -2514,11 +2514,21 @@ int parse_create_object_sub(p_object *p_objp, bool standalone_ship)
...
+
+		if (entry->status == ShipStatus::INVALID) {
+			Warning(LOCATION, "Potential in-game join bug: ship registry status for %s is INVALID", shipp->ship_name);
+		}
+		if (entry->p_objp == nullptr) {
+			Warning(LOCATION, "Potential in-game join bug: ship registry parse object for %s is nullptr", shipp->ship_name);
+		} else if (entry->p_objp != p_objp) {
+			Warning(LOCATION, "Potential in-game join bug: ship registry parse object for %s is different from its expected value", shipp->ship_name);
+		}
+
```
notimaginative, line 2531: Would these do better under a multi check, or perhaps tweak the message a bit? The message text seems specific to multi but since you hit it in single player too I wonder if it might be confusing if hit in other non-multi circumstances.

Goober5000: Ideally it shouldn't be hit in single player from now on, but yeah that's a good point.  I'll edit them.

notimaginative (approved):

PR #5670, opened 2023-10-06
fix and upgrade ai-fly-to-ship

Goober5000:
The ai-fly-to-ship goal was added early in FSO's development but never worked right.  Until recently it could crash; and even aside from the crash the AI would try to reach the target's exact position, causing repeated collisions.  This folds ai-fly-to-ship into ai-stay-near-ship, since the behaviors are conceptually the same except for fly-to being satisfied as soon as the target is in range.

This also adds a new "escort" mode usable by both behaviors.  The default Volition behavior is to swoop toward the target in a long, graceful arc; the new mode generates a goal point and flies directly to that point.  These can be toggled via an optional sexp parameter.

There is also some minor cleanup and a new vec3d operator for convenience.

Needed for Scroll.  Depends on #5667; in draft until that is merged.


code/ai/aigoals.cpp
```diff
@@ -1022,17 +1022,30 @@ void ai_add_goal_sub_sexp( int sexp, int type, ai_info *aip, ai_goal *aigp, char
...
 		aigp->priority = atoi( CTEXT(CDDR(node)) );
+
+		// distance from ship
 		if ( CDDDR(node) < 0 )
 			aigp->float_data = 300.0f;
 		else
 			aigp->float_data = i2fl(eval_num(CDDDR(node), is_nan, is_nan_forever));
-		aigp->ai_mode = AI_GOAL_STAY_NEAR_SHIP;
+
+		// (the next argument is whether to afterburn)
```
Baezon, line 1039: Put a brief additional comment here saying that it's handled further below, instead of right here (Not that I'm actually sure handling the flag in one fell swoop at the end of the function is a good way to do it...).

Baezon (changes requested):
There's some original code which was dealing with `Use_ap_cinematics` and `AutoPilotEngaged`, I'll admit to not understanding it well myself but does it still work as intended wrt this goal?

Goober5000:
> There's some original code which was dealing with `Use_ap_cinematics` and `AutoPilotEngaged`, I'll admit to not understanding it well myself but does it still work as intended wrt this goal?

Well, it never worked in the first place -- the fly-to-ship behavior was never used by the autopilot code.  So, the answer is no? 😀

Baezon (approved):

PR #5671, opened 2023-10-07
add "Calculate subsystem hitpoints after parsing" option

Goober5000:
Subsystem hitpoints are calculated as a percentage of the ship's hull strength as they are parsed.  If a modular table modifies the ship's hull strength, the subsystems are not modified at the same time.  This can lead to unreasonably large or small subsystem hitpoint values, as recently seen in the 1.3.0 version of Cardinal Spear.

This PR adds an option to defer hitpoint calculation until after parsing so that the subsystem is a percentage of the correct hull strength.  Since this may affect the balance of existing mods, it is off by default unless the build targets 24.0 or later.

This also removes the long-unused `hull_percentage_of_hits` check and adds several log statements for significant options in game_settings.tbl.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -5002,13 +5002,6 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 		subsystems[i].reset();
 	}
 	
-	float	hull_percentage_of_hits = 100.0f;
-	//If the ship already has subsystem entries (ie this is a modular table)
-	//make sure hull_percentage_of_hits is set properly
-	for(auto i=0; i < sip->n_subsystems; i++) {
-		hull_percentage_of_hits -= sip->subsystems[i].max_subsys_strength / sip->max_hull_strength;
-	}
-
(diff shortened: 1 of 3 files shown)


code/ship/ship.cpp
```diff
@@ -5390,13 +5386,6 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 		}
 	}	
 
-	// must be > 0//no it doesn't :P -Bobboau
-	// yes it does! - Goober5000
-	// (we don't want a div-0 error)
-	if (hull_percentage_of_hits <= 0.0f )
-	{
-		//Warning(LOCATION, "The subsystems defined for the %s can take more (or the same) combined damage than the ship itself. Adjust the tables so that the percentages add up to less than 100", sip->name);
-	}
```
Baezon, line 5399: O.o Bit of a bizarre, completely unrelated warning given what it's actually checking and the comments.

Baezon (approved):

PR #5676, opened 2023-10-12
Add command line option for window resolution

BMagnu:
This allows FSO to render the actual game at a different resolution from the window, by routing all rendering through an additional framebuffer (only if the option is actually used, for performance reasons).

I.e. -res 1920x1080 -window_res 1000x1000 will render the game and UI at 1080p, but the window will be 1000 by 1000 pixels large, with the image stretched to fit.

There are mainly two uses for this:
1. It allows a modder to test HUD / Mainhall / UI-setup for high resolutions or strange aspect ratios, even if the modder has no screen that could actually support that resolution.
2. In VR, users will regularly run FSO at resolutions higher than their displays, as the VR image is sent to the HMD before the downscaling. Without this, they'd also need a screen to support this to allow for mouse input / rendering over the entire game area. Thus, required for #5583.

### DIFF
--- a/code/cmdline/cmdline.cpp
+++ b/code/cmdline/cmdline.cpp
@@ -1734,7 +1734,7 @@ bool SetCmdlineParams()
 			sprintf(override, "{\"width\":%d,\"height\":%d}", width, height);
 			options::OptionsManager::instance()->setOverride("Graphics.Resolution", override);
 		} else {
-			mprintf(("Failed to parse -res parameter \"%s\". Must be in format \"<width>x<height>\".\n", Cmdline_res));
+			Warning(LOCATION, "Failed to parse -res parameter \"%s\". Must be in format \"<width>x<height>\".\n", Cmdline_res);
 		}
 	}
 	if(window_res_arg.found()){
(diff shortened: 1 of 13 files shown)


wookieejedi:
Update: this looks like it's the last of the PreReq PRs that the larger #5583 needs. Any graphics coders up for possibly giving this one a review?

code/cmdline/cmdline.cpp
```diff
@@ -1735,6 +1737,16 @@ bool SetCmdlineParams()
...
 		}
 	}
+	if(window_res_arg.found()){
+		int width = 0;
+		int height = 0;
+
+		if ( sscanf(window_res_arg.str(), "%dx%d", &width, &height) == 2 ) {
+			Cmdline_window_res.emplace(width, height);
+		} else {
+			mprintf(("Failed to parse -res parameter \"%s\". Must be in format \"<width>x<height>\".\n", Cmdline_res));
```
JohnAFernandez, line 1747: Might be eligible for a warning, since this happens right at program start and the user might not look to think in the debug log for what they did wrong.

BMagnu: Done!

The-E (approved):

PR #5677, opened 2023-10-12
rewrite primary weapon cycling

### LINKED ISSUE #5529: Bug: Primary weapon linking edge cases
[Testcase_2.zip](https://github.com/scp-fs2open/fs2open.github.com/files/12209380/Testcase_2.zip)

This is kinda hard to explain so here goes:
If you have a ship with one or two ammo weapons.
And one of those weapons can not be selected along with the others.
If that last weapon can no longer be selected, the player can also not link their two remaining primary weapons.

As an example, I've included a "test case" mission. It's designed to be run in Blue Planet Complete (the version that is on knossos).
To reproduce:
1) Shoot your primary weapons for a bit.
2) Use the supplied ammo pack
3) Swap your primary weapons a bunch of times and notice that they'll no longer link reliably.

Bug is present both in RC and last stable.

Goober5000:
Remove a whole bunch of ad-hoc code piled on top of the retail weapon cycling behavior, then replace it with a robust algorithm that takes all relevant attributes into account.

Fixes #5529.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -13621,7 +13621,7 @@ done_secondary:
 }
 
 // Goober5000
-static int primary_out_of_ammo(ship_weapon *swp, int bank)
+static bool primary_out_of_ammo(const ship_weapon *swp, int bank)
 {
 	// true if both ballistic and ammo <= 0,
 	// false if not ballistic or if ballistic and ammo > 0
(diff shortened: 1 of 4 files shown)


code/ship/ship.cpp
```diff
@@ -13630,66 +13630,38 @@ static int primary_out_of_ammo(ship_weapon *swp, int bank)
...
 
-	shipp = &Ships[objp->instance];
-	sip = &Ship_info[shipp->ship_info_index];
-	swp = &shipp->weapons;
+	auto shipp = &Ships[objp->instance];
+	auto sip = &Ship_info[shipp->ship_info_index];
+	auto swp = &shipp->weapons;
 
-	Assert(direction == CYCLE_PRIMARY_NEXT || direction == CYCLE_PRIMARY_PREV);
+	Assert((swp->current_primary_bank >= 0) && (swp->current_primary_bank < swp->num_primary_banks));
```
Baezon, line 13660: iirc `current_primary_bank` will be `-1` for ships without primary weapons. Will this function be avoided if they press the cycle primary key in that case?

Baezon (changes requested):

code/ship/ship.cpp
Goober5000: Oh, good catch.  I moved this Assert from its original location; I'll move it back.

Baezon (approved):

PR #5678, opened 2023-10-13
Allow to Capture Mouse in FSO window

BMagnu:
Also, ensure that FSO is on top when it starts.
Required for #5583.


code/cmdline/cmdline.cpp
```diff
@@ -414,6 +416,7 @@ int Cmdline_ship_choice_3d = 0;
 int Cmdline_weapon_choice_3d = 0;
 int Cmdline_autopilot_interruptable = 1;
 int Cmdline_stretch_menu = 0;
+bool Cmdline_capture_mosue = false;
```
Goober5000, line 419: mosue -> mouse

Goober5000 (commented):
I can't speak to the SDL stuff, but the rest looks fine.

code/cmdline/cmdline.cpp
BMagnu: Done

Goober5000 (approved):

PR #5679, opened 2023-10-15
Fix two small memory leaks

JohnAFernandez:
Fixes coverity 1093679 and 1093680 by freeing if an early return is reached.


Goober5000 (approved):
Good catches.

MjnMixael:
[Use scripts to add the Greenbark messages from BtA1 to the message scrollback log](https://trello.com/c/vqps9R6I/166-use-scripts-to-add-the-greenbark-messages-from-bta1-to-the-message-scrollback-log)

PR #5680, opened 2023-10-15
Change underlying type to avoid endianness problem

JohnAFernandez:
Before this change if the wrong endianness was used, the c style cast that was used here would lead to the argument being passed to getsockopt to lack the actual information it needed.

Fixes coverity 1093482

### DIFF
--- a/code/network/chat_api.cpp
+++ b/code/network/chat_api.cpp
@@ -203,8 +203,8 @@ int ConnectToChatServer(char *serveraddr, char *nickname, char *trackerid)
 			{
 				// make sure that we don't have any connect() errors (since it's non-blocking)
 				int err_val = 0;
-				size_t err_val_size = sizeof(err_val);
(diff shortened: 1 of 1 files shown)


notimaginative (changes requested):
This wouldn't be an endianness issue as it doesn't deal with any data external to the host system, so I'm not sure what the warning message is about.

But, the type of `err_val_size` really should have been `socklen_t`, which would also have avoided the c-style casting. For the sake of portability that's probably the better fix for this. Assuming coverity doesn't complain about that too for some reason.

JohnAFernandez:
Well, we may never know if it will complain because our Coverity checks are no longer active.  In any case, thanks for taking a look.

notimaginative (approved):
LGTM.

Although I'm a bit surprised the CI let that c-style cast on `err_val` go without a warning.

PR #5684, opened 2023-10-16
don't kill/self-destruct ship if it's already dying

Goober5000:
If a ship is in the process of dying, it shouldn't be killed or self-destructed again.  This fixes a bug noticed by RedMageJoe during the Derelict playthrough.

(This situation is distinct from a ship being hit while in the middle of its death roll.  That shortens the death roll but does not try to kill the ship again.)


JohnAFernandez (approved):
LGTM

What actually happened?

Baezon (approved):
It creates lots of big explosions and can delay actually finishing exploding. This is something I actually noticed in Blighted but never got around to fixing it.

Goober5000:
Yup, what he said.  In the Derelict case, a repeating event that subtracted hull points from a ship would cause the self-destruct to trigger over and over once the hull got below 0.

PR #5685, opened 2023-10-16
Coverity null pointer checking issues

JohnAFernandez:
Another five coverity issues regarding dereference of NULL and nullptr pointers.


code/camera/camera.cpp
```diff
@@ -163,7 +163,7 @@ void camera::set_object_host(object *objp, int n_object_host_submodel)
 	set_custom_orientation_function(NULL);
 	if(n_object_host_submodel > 0)
 	{
-		if(objp->type == OBJ_SHIP)
+		if(objp != nullptr && objp->type == OBJ_SHIP)
```
Goober5000, line 166: At the top of this function, these lines can be removed:
```
if(objp == NULL)
		object_host = object_h();
```
Constructing `object_h` with a null argument is equivalent to constructing it with no arguments (although this wasn't the case at the time the coverity results were generated).  Also, the `object_host` variable is immediately overwritten anyway.

Goober5000 (changes requested):

code/camera/camera.cpp
JohnAFernandez: Usually I don't like adding these secondary changes, but it really is immediately overwritten. Done

Goober5000 (approved):

PR #5686, opened 2023-10-16
Lua add scrollback

MjnMixael:
Lua method to add a message to the mission message scrollback without sending it as an actual message in the mission.


MjnMixael:
[Use scripts to add the Greenbark messages from BtA1 to the message scrollback log](https://trello.com/c/vqps9R6I/166-use-scripts-to-add-the-greenbark-messages-from-bta1-to-the-message-scrollback-log)

code/scripting/api/libs/mission.cpp
```diff
@@ -988,6 +988,74 @@ ADE_FUNC(sendPlainMessage,
...
+			return ADE_RETURN_FALSE;
+		} else {
+			if (source < HUD_SOURCE_TEAM_OFFSET) {
+				LuaError(L, "Source index must be greater than the %i built-in sources.", HUD_SOURCE_TEAM_OFFSET);
+				return ADE_RETURN_FALSE;
+			}
+		}
+	} else {
+		if (!ade_get_args(L, "s|i", &message, l_Enum.GetPtr(&esp))) {
+			return ADE_RETURN_FALSE;
```
TRBlount, line 1015: The function signature of `addMessageToScrollback` must be changed.  Allowing a number is going to cause scripts to break if the hud sources are ever changed.  Instead the argument should be either an enumeration or an IFF type.  The `ade_odata_is_userdata_type` function can be used to check what the type is.

TRBlount (changes requested):

code/scripting/api/libs/mission.cpp
MjnMixael: Good idea. Didn't know that `is_userdata_type` function existed. Done.

PR #5688, opened 2023-10-18
Force MSAA in-game toggle to require restart

MjnMixael:
When switching MSAA from off to on it can cause the game rendering to go completely black.
image

So this PR enforces that by requiring a restart for the in-game version of the option.


wookieejedi (approved):
good catch

PR #5690, opened 2023-10-18
Fix glare shining through cockpits in free-look

### LINKED ISSUE #3098: Sun flares aren't properly blocked by cockpit models
Basically sun flares come right though the cockpit even if the cockpit is solid. It is most noticeable if the cockpit has subobjects which don't seem to block flares at all. If condensed into 1 detail0 object it will block them most of the time but they still leak through at times especially when close to the edge of the model or when using freelook.

This is especially noticeable around the e7 in Diaspora. Play the 2nd to last mission in the campaign, choose the e7, leave the launch tube and look at the sun. See attached image.
screen0122

This bug is migrated from Mantis.

wookieejedi:
Fixes #3098. The `shipfx_eye_in_shadow` function did a collision check for cockpits, bit did not previously take into account of the viewpoint had a non-forward orientation (IE the player was using free-look). This PR sets the collide check to take that into account with a 1 line fix. Tested with multiple FotG cockpits and fixes the issue.

### DIFF
--- a/code/ship/shipfx.cpp
+++ b/code/ship/shipfx.cpp
@@ -866,7 +866,7 @@ bool shipfx_eye_in_shadow( vec3d *eye_pos, object * src_obj, int sun_n )
 				mc.model_instance_num = -1;
 				mc.model_num = sip->cockpit_model_num;
 				mc.submodel_num = -1;
-				mc.orient = &Eye_matrix;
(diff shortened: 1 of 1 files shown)


BMagnu (approved):

JohnAFernandez:
Force pushed to check our automation to see if work around still works.

PR #5693, opened 2023-10-19
add workaround for broken windows workflow

notimaginative:
Temporary fix for broken windows workflow. Hopefully a more permanent solution will be found soon and we can then remove this.

Reference: https://github.com/actions/runner-images/issues/8598

### DIFF
--- a/.github/workflows/build-nightly.yaml
+++ b/.github/workflows/build-nightly.yaml
@@ -144,6 +144,12 @@ jobs:
           vulkan-query-version: 1.3.204.0
           vulkan-components: Vulkan-Headers, Vulkan-Loader
           vulkan-use-cache: true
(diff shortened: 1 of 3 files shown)


notimaginative:
Setting this as a draft until there is a consensus as to whether to use this temporary fix or to wait for a proper solution from github. Also this isn't directly tested but a copy&paste for the recommended workaround from linked github issue.

JohnAFernandez:
It would be nice to have this, so I can continue work on my github actions script.

Goober5000 (approved):
Well it fixed the automatic builds.  I say let's use it.  We can revert it easily enough if we need to.

JohnAFernandez:
Confirmed: this does fix the issue in my script, so it should probably be merged.

PR #5694, opened 2023-10-19
Upgrade Lua Sound Handling methods

MjnMixael:
This PR upgrades the LUA Sound Handle methods to add the newer pause/resume methods. It also allows for specifying using either the Effects volume multiplier or the Voice volume multiplier where appropriate. 

(Note that music handling is not added here because music should still be done with AudioStream as well as the fact that there's no sound_handle methods that take the master music volume into account. They generally assume either voice or effects.)

Relies on PR #5636 and set as draft until that is merged.


JohnAFernandez (approved):
LGTM

PR #5696, opened 2023-10-20
Add deferred lighting as an in-game option

MjnMixael:


code/graphics/opengl/gropengl.cpp
```diff
@@ -78,6 +78,19 @@ static std::unique_ptr<os::OpenGLContext> GL_context = nullptr;
 static std::unique_ptr<os::GraphicsOperations> graphic_operations = nullptr;
 static os::Viewport* current_viewport = nullptr;
 
+// used by In-Game Options menu
+static bool DeferredLightingEnabled = true;
+
+static auto DeferredLightingOption = options::OptionBuilder<bool>("Graphics.DeferredLighting",
+                     std::pair<const char*, int>{"Deferred Lighting", 1783},
+                     std::pair<const char*, int>{"Enables or disables deferred lighting", 1784})
```
TRBlount, line 86: These should be IDs 1782 and 1783 respectively. Just a reminder that the last used ID should always be `XSTR_SIZE-1`. If the valid range of IDs is 0-1783, then XSTR_SIZE should be 1784. Treat XSTR_SIZE as "the next available/unused ID".

TRBlount (changes requested):

code/graphics/opengl/gropengl.cpp
MjnMixael: Bleh. It's confusing to do it that way for a glorified comment, but alright.

MjnMixael: Done

TRBlount (approved):

PR #5698, opened 2023-10-20
Waypoints upgrade

Goober5000:
As requested by @Durandal-Thoth, this feature allows waypoint paths to be traveled in reverse (finally putting the retail `WPT_BACKTRACK` flag to use) and also started in mid-path, i.e. at an index other than the first or last point.

This PR incorporates quite a bit of refactoring of waypoint.cpp and consequential cascading const-correctness; that could be split into a separate PR if desired.


code/network/multi_obj.cpp
```diff
@@ -1493,7 +1493,7 @@ int multi_oo_pack_data(net_player *pl, object *objp, ushort oo_flags, ubyte *dat
 		// either send out the waypoint they are trying to get to *or* their current target
 		if (umode == AIM_WAYPOINTS) {
 			// if it's already started pointing to a waypoint, grab its net_signature and send that instead
-			if ((aip->wp_list != nullptr) && (aip->wp_list->get_waypoints().size() > aip->wp_index)) {
+			if ((aip->wp_list != nullptr) && (aip->wp_index >= 0 && aip->wp_index < (int)aip->wp_list->get_waypoints().size())) {
```
JohnAFernandez, line 1496: We gotta start using c++ style casts for new code.  C style casts will literally do things you don't want them to do.

code/object/waypoint.cpp
```diff
@@ -178,14 +184,13 @@ waypoint_list *find_matching_waypoint_list(const char *name)
 waypoint *find_matching_waypoint(const char *name)
 {
 	Assert(name != NULL);
-	SCP_list<waypoint_list>::iterator ii;
 
-	for (ii = Waypoint_lists.begin(); ii != Waypoint_lists.end(); ++ii)
+	for (auto &ii: Waypoint_lists)
```
JohnAFernandez, line 188: Can this loop be const, too?

code/object/waypoint.cpp
```diff
@@ -308,47 +304,33 @@ waypoint *find_waypoint_at_index(waypoint_list *list, int index)
...
-	for (ii = Waypoint_lists.begin(); ii != Waypoint_lists.end(); ++ii)
-	{
-		if (&(*ii) == wp_list)
-			return index;
-		index++;
-	}
+	ptrdiff_t index = wp_list - Waypoint_lists.data();
+
+	if (index >= 0 && index < (ptrdiff_t)Waypoint_lists.size())
+		return (int)index;
```
JohnAFernandez, line 320: C++ style casts here

code/scripting/api/objs/order.cpp
```diff
@@ -303,7 +305,7 @@ ADE_VIRTVAR(Target, l_Order, "object", "Target of the order. Value may also be a
 		case AI_GOAL_WAYPOINTS_ONCE:
 			// check if waypoint order is the current goal (ohp->odx == 0) and if it is valid
 			if ( (ohp->odx == 0) && (aip->wp_index != INVALID_WAYPOINT_POSITION) &&
-				(aip->wp_index < aip->wp_list->get_waypoints().size()) ) {
+				(aip->wp_index >= 0 && aip->wp_index < (int)aip->wp_list->get_waypoints().size()) ) {
```
JohnAFernandez, line 308: c++ cast

JohnAFernandez (changes requested):

code/network/multi_obj.cpp
Goober5000: All right

code/object/waypoint.cpp
Goober5000: Can do

Goober5000: Actually it can't, because the return pointer is not const (and the return pointer is used in non-const ways).

JohnAFernandez (approved):

PR #5699, opened 2023-10-21
handle parse objects that are out of order

Goober5000:
Occasionally, FRED can save a mission with a wing's ships out of order; this can happen if ships are deleted and FRED finds a free index in the middle of the ship list.  This can cause benign but confusing mismatches in the ship registry, the hotkey list, and other places -- and it's also possible some mismatches may not be so benign, e.g. in multiplayer.  Anyway, it's better for the ships to appear in the order the wing expects them, so this adds a routine to swap them around if necessary.

Due to a change in indentation, I recommend viewing the diff while hiding whitespace changes.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -4710,50 +4710,77 @@ void parse_wing(mission *pm)
 		free_sexp2(wing_goals);  // free up sexp nodes for reuse, since they aren't needed anymore.
 	}
 
+	// make a temporary map since we want to keep this separate from the ship registry
+	SCP_unordered_map<SCP_string, int> parse_object_indexes;
+	i = 0;
+	for (const auto& pobj : Parse_objects)
+		parse_object_indexes.emplace(pobj.name, i++);
+
+	// Goober5000 - to avoid confusing mismatches in the ship registry and hotkey list, and possibly other places,
+	// make sure the order of parse objects matches their order in the wing
+	auto prev_iter = parse_object_indexes.end();
+	for (i = 0; i < wingp->wave_count; i++) {
+		auto this_iter = parse_object_indexes.find(ship_names[i]);
+		if (this_iter == parse_object_indexes.end()) {
+			Error(LOCATION, "Cannot load mission -- for wing %s, ship %s is not present in #Objects section.\n", wingp->name, ship_names[i]);
+			break;
+		}
+
+		if (i > 0) {
+			// compare the parse object indexes of the previous ship and this ship
+			if (this_iter->second < prev_iter->second) {
+				// swap the parse objects
+				std::swap(Parse_objects[this_iter->second], Parse_objects[prev_iter->second]);
+
+				// swap the indexes in our temporary map
+				std::swap(this_iter->second, prev_iter->second);
+
+				// start over from the beginning of the wing
+				i = -1;
+				this_iter = parse_object_indexes.end();
+			}
+		}
+
+		prev_iter = this_iter;
+	}
+
 	// set the wing number for all ships in the wing
 	for (i = 0; i < wingp->wave_count; i++ ) {
 		char *ship_name = ship_names[i];
-		int assigned = 0;
 
 		// Goober5000 - since the ship/wing creation stuff is reordered to accommodate multiple docking,
 		// everything is still only in the parse array at this point (in both FRED and FS2)
 
 		// find the parse object and assign it the wing number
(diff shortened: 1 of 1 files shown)


JohnAFernandez:
@notimaginative Do you think multi relies on the old order?  Maybe for in-game joining?  Do you think we need a multi bump here?

notimaginative:
Not sure really. I planned to do some testing on this but haven't gotten the chance so far.

The in-game join crashes were just sort of random so it's hard to say whether they're actually related to this issue or not. But when those crashes did happen it was just a constant barrage of them all of a sudden over multiple quick sessions. The Terran Gauntlet was one mission I tested a lot with so we just need to see if it has an order problem.

Either way I'll try and do at least some testing this afternoon with and without this change to see what happens. Maybe trigger some of the new warnings too and hope the helps pin something down so we can stop guessing.

Goober5000:
This wouldn't require a multi bump.  It's just rearranging parse objects on the fly as they are being loaded from the mission.  It would have the same effect as going into the mission in a text editor and swapping the objects

Goober5000:
The Terran Gauntlet does indeed have a wing with fighters out of order: Pisces 5 comes before Pisces 4.

JohnAFernandez:
It would matter if the parse object order was relied on to be the same.  You never know what obscure thing that multi relies on since it is such a tacked on system.  That's why I asked Taylor.  He's knows the packets which transmit loadouts and in-game joining better, which are the most likely to rely on the parse object code to be the same.

In fact, I may have just thought of an area that does.

notimaginative:
I'm getting a variety of completely unrelated crashes (like `vsnprintf()` no longer taking `nullptr` for size check). Possibly Sonoma upgrade related since this is the first real play testing I've done since the upgrade. But I have a free day tomorrow so I'll try and get it sorted out.

Goober5000:
> It would matter if the parse object order was relied on to be the same. You never know what obscure thing that multi relies on since it is such a tacked on system. That's why I asked Taylor. He's knows the packets which transmit loadouts and in-game joining better, which are the most likely to rely on the parse object code to be the same.

Since the swapping happens during parsing, and the ordering is well-defined, then every multi participant would see a consistent order.  And don't worry, I also messaged Taylor about this PR as it relates to multiplayer.

> Edit: Thankfully, after checking, it looks like net signature is assigned as objects are parsed and the intra-wing swap happens after, so they retain the correct net signatures.

This raises a good point: does the order of the net signatures matter?  With this PR, if you sorted by net signature the order would not match the order of `Parse_objects`.  It would be possible to fix this if we need to.

notimaginative:
I haven't gotten to test this set of changes yet, but I do see where it could cause an issue with net signatures. However that might be primarily due to issues with the current code which we've simply not noticed previously.

Net signatures are assigned in order to the parse object in `parse_create()`, and this signature is later copied to the respective `Object` in `parse_create_object_sub()`. When an object is created in multi the `Object` signature is updated to be what the server says, but the signature for the parse object is never changed from what it was at mission load. So the net signatures can end up different causing issues with stuff like `mission_parse_get_parse_object()` where the server signature is one thing and it tries to match with a signature that was created by the client during mission load. This can lead to an object not being found, or an entirely different object being returned instead.

I think this is the main issue that in-game is running in to. And is almost certainly lead to a previous "fix" that I now believe was incorrect.

Cyborg will need to check me on this, but it might cause (or be causing?) an issue with rollback too. Since the object is loaded into rollback with the parse order signature during `parse_create_object_sub()` and then again with a possibly different net signature after the `Object` signature is changed to match the server.

And I haven't even gotten to really looking at signature handling for wings or other non-ship object types yet.

So I don't think this PR would create a new problem, but would likely exacerbate an existing one. At the least I believe a multi bump will be required to fix the underlying issue.

This is probably going to need a lot more testing to really sort out.

JohnAFernandez:
I didn't even think about rollback.  I can check that today, but I am hoping I relied on net signatures well enough that they wouldn't be affected, if net signatures are properly synced.

You know, as much as Volition tried to avoid sending ship names to assign net signatures, that might be the better long term option, as that would gaurantee that all ships would be synced up. (The mission files dont even need to have ships in the same order in that case)

Goober5000:
If the parse objects are swapped, it's easy enough to swap back the net signatures so that they are in their originally assigned order.

notimaginative (approved):
LGTM

PR #5700, opened 2023-10-21
change "special warp physics" to "supercap warp physics"

Goober5000:
Per discussion with @Baezon, this is a more precise term.


Baezon (approved):

JohnAFernandez (approved):
LGTM

I think the name is clearer as well.

PR #5701, opened 2023-10-21
add const where appropriate to the jump node API

Goober5000:
See title.  A followup of sorts to #5464.

### DIFF
--- a/code/jumpnode/jumpnode.cpp
+++ b/code/jumpnode/jumpnode.cpp
@@ -177,9 +177,9 @@ const object *CJumpNode::GetSCPObject() const
 /**
  * @return Color of jump node when rendered
  */
-const color *CJumpNode::GetColor() const
+const color &CJumpNode::GetColor() const
 {
-	return &m_display_color;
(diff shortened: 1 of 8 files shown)


code/jumpnode/jumpnode.cpp
```diff
@@ -177,23 +177,23 @@ object *CJumpNode::GetSCPObject()
 /**
  * @return Color of jump node when rendered
  */
-color CJumpNode::GetColor()
+const color *CJumpNode::GetColor() const
```
JohnAFernandez, line 180: Why is this a pointer now?  Why not a const reference?

Goober5000: Mostly just to be consistent with `GetSCPObject()` and `GetPosition()`.

JohnAFernandez: I guess I'm just not convinced that this is a good reason to use a pointer over a reference here.

JohnAFernandez (approved):
Much appreciated

PR #5702, opened 2023-10-22
Fix Redefined Macro Warning

JohnAFernandez:
This fixes a small warning on mac compilers that OS_USED was previously defined in the SDK.

### DIFF
--- a/code/object/objectsnd.h
+++ b/code/object/objectsnd.h
@@ -14,7 +14,7 @@
 
 #include "gamesnd/gamesnd.h"
 
-#define	OS_USED					(1<<0)
+#define	OS_IN_USE				(1<<0)		// changed from OS_USED since it shadowed another macOS macro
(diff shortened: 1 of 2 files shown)


code/object/objectsnd.h
```diff
@@ -14,7 +14,7 @@
 
 #include "gamesnd/gamesnd.h"
 
-#define	OS_USED					(1<<0)
+#define	OS_IN_USE				(1<<0)		// changed from OS_USED since it shadowed another macos macro
```
Goober5000, line 17: "macos" could be misread as a typo for "macros".  Could you capitalize the term or reword the comment?

Goober5000 (changes requested):

code/object/objectsnd.h
JohnAFernandez: All set

Goober5000 (approved):

PR #5703, opened 2023-10-22
Add CMakePresets.json to gitignore

JohnAFernandez:
This would be pretty helpful to me, as for some reason on my Mac, vscode does not automatically know how to build FSO and I have to create a preset to build it.  But every time I do, it gets added to the changes I have to uncheck when making commits.  LMK if this change is undesired for any reason as I could live with the inconvenience, if necessary.


Goober5000 (approved):
I think this is fine 👍

PR #5704, opened 2023-10-22
Convert sprintf to snprintf part 1

JohnAFernandez:
This partially addresses a very common warning from clang when compiling FSO that sprintf is deprecated and needs to be converted to snprintf.  Now, I don't think the security issues are a huge deal for us, but on the compilation output, it will print out multiple lines for each occurrence, and we use sprintf upwards of 800 times.

This pr converts about a third of them.  Note that the SCP_string overload of sprintf is not affected.

In draft until I can do some basic smoke testing.  The changes were not complicated, but they are extensive.


JohnAFernandez:
I did not see any odd text behavior when I tested this patch.

PR #5706, opened 2023-10-22
fix change-soundtrack after <No Music>

### LINKED ISSUE #5705: change-soundtrack sexp doesn't work when no music is defined in mission specs.
Issue does not occur in FSO 23.0. Does occur in 23.2 stable.
Steps to reproduce:
Open mission bp2-05 with FSO 23.2
notice a distinct lack of music.

When a soundtrack is enabled in music specifications, the sexp works as normal. Even when a completely different soundtrack is selected (I selected "Joshua"), the mission starts out with the correct music.

The bug seems to occur whenever change-soundtrack is called. If a mission starts silent or starts with a "play-sound-from-file" and then later calls a soundtrack, it also won't work.

Goober5000:
The fix in #5578 assumed that a previously playing soundtrack would properly set the timestamp, but if no soundtrack was previously played at all, this wouldn't happen.  So if we deliberately don't play a soundtrack, be sure to set the timestamp as if `event_music_do_frame()` had been called.

Fixes #5705.

### DIFF
--- a/code/gamesnd/eventmusic.cpp
+++ b/code/gamesnd/eventmusic.cpp
@@ -606,7 +606,10 @@ void event_music_level_start(int force_soundtrack)
 
 	// Check that soundtrack index is in range, except that < 0 means no soundtrack
 	if (Current_soundtrack_num < 0)
+	{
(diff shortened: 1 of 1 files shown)


JohnAFernandez (approved):
LGTM

PR #5707, opened 2023-10-23
move g3_project_vertex in HudGaugeReticleTriangle::renderTriangle

Goober5000:
In commit 09fad95070, the call to `g3_project_vertex` was moved outside the if() block.  In every other case where the vertex's `codes` field is compared with 0 (to see whether the vertex is on-screen or not) the call to `g3_project_vertex` appears after the check.  I therefore presume that the move was a mistake.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -2985,7 +2985,6 @@ void HudGaugeReticleTriangle::renderTriangle(vec3d *hostile_pos, int aspect_flag
 	cur_dist = vm_vec_dist_quick(&Player_obj->pos, hostile_pos);
 
 	g3_rotate_vertex(&hostile_vertex, hostile_pos);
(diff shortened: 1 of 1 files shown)


Baezon (changes requested):
Typically, when getting the screen space position of a vertex, if it would be outside the view frustum, it's pointless to bother converting it to actual screen coordinates (which is what `g3_project_vertex` does) and doing anything with it, which is why `codes` is checked first. But `hostile_vertex` is not a typical case, regardless of whether the missile is on screen or not *the triangle must be drawn pointing to its theoretical screen position*, even if that point is off screen, especially so, even.

Without filling the `screen` variable, lines 3032 and 3034 would be acting on uninitialized data, as they do not care whether the point is on screen or not.

Goober5000:
Gotcha.  Ok, then this was a red herring.

PR #5708, opened 2023-10-23
Remove Unused Functions

JohnAFernandez:
These functions have not been called from what I can tell since the source code was released.  But they did happen to trigger a clang warning, so it's time to get rid of them.


Goober5000 (approved):
This is a fascinating glimpse at the hidden history of FreeSpace development.  It's like code archaeology.

PR #5713, opened 2023-10-23
Add Game Option to refresh selected weapon in loadout

wookieejedi:
This PR adds an option to always ensure the viewed weapon is one that can actually be on the selected ship when first clicking on the weapon loadout.

If you go to the loadout in the briefing and view the weapons on your ship, then go and change your ship class and go back to the weapon loadout screen, you will see that the weapon shown in the window on the right side is from your very first viewing before you changed your ship class. Displaying a weapon not allowed on a ship by default can be confusing to players, especially on mods like FotG where most ships have unique weapons.

The player can still of course click on a non-allowed weapon to view it, but with this PR the player will be initially shown a selected weapon that is allowed on the ship when opening the loadout editor.

Tested and works as expected.

### DIFF
--- a/code/missionui/missionweaponchoice.cpp
+++ b/code/missionui/missionweaponchoice.cpp
@@ -1515,13 +1515,13 @@ void wl_maybe_reset_selected_slot()
 
 /**
  * If Selected_wl_class is -1, choose the first weapon available from the pool for an animation
- *  - on second thought, choose the first weapon that is oin the ship, then go to the pools
+ *  - on second thought, choose the first weapon that is on the ship, then go to the pools
  */
 void wl_maybe_reset_selected_weapon_class()
 {
 	int i;
 
-	if ( Selected_wl_class >= 0 ) 
+	if ( (Selected_wl_class >= 0) && !(Always_reset_selected_wep_on_loadout_open) ) 
 		return;
 
 	Assert( Wss_slots != NULL );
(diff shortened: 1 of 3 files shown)


Goober5000 (changes requested):
This seems like a kludge.  If there's an actual bug, we should find the root cause and fix it, not add an option to do a certain behavior sometime and another behavior other times.

Do we have a github issue for the bug in question?  And do we have a consistently reproducible test case?

wookieejedi:
Ah so I wouldn't really call it a bug. All that is happening is that the game remembers what weapon you had previously clicked on (it does not try to apply it to the ship or anything). 

To view what is happening, just open the weapon loadout, click on a weapon, switch back to the ship loadout and change ships. Then go back to the weapon loadout. You'll see the weapon loadout will continue to show the weapon you had selected the first time. Again, not a bug, just for mods with lots of custom ship weapon combinations it would be very useful, especially for new players, if the weapon loadout view forgot it's old selection and instead showed the current weapons on the selected ship by default.

Goober5000 (commented):
Oh, I see what you mean,  Ok, that makes sense.

In that case, why not move the option to immediately before `$Weapons inherit parent collision group:`?  That would group it with the other ship and weapon selection options.

(Long term, we should probably upgrade mod_table to handle parsing its options in any order, the same as ai_profiles does.)

wookieejedi:
Good idea, just updated the position.

Goober5000 (approved):
Looks good!

PR #5715, opened 2023-10-23
Bugfix HUD target direction triangle

Baezon:
Follow-up to #5656. In that PR, in determining the direction to point for the HUD target triangle, the calculation uses the `screen` position of the projected point, but `g3_project_vertex` early returns and leaves `screen` uninitialized if the point is behind the camera. However, VR requires the screen position and not the old `world` position (the point transformed into the camera space).

This PR still sets the `PF_OVERFLOW` flag, but projects the point regardless (and flips it because everything would be backwards). This way it can still give a meaningful result, while alerting any callers it's behind the camera. All callers properly check for this flag in all relevant behavior, or have been updated to do so to ensure behavior is unchanged.

Tagging @BMagnu to hopefully ensure this still works as expected in VR.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -3029,6 +3029,7 @@ void HudGaugeReticleTriangle::renderTriangle(vec3d *hostile_pos, int aspect_flag
 		}
 	}
 
+	// even if the screen position overflowed, it will still be pointing in the correct direction
 	unsize( &hostile_vertex.screen.xyw.x, &hostile_vertex.screen.xyw.y );
 
 	ang = atan2_safe(-(hostile_vertex.screen.xyw.y - tablePosY), hostile_vertex.screen.xyw.x - tablePosX);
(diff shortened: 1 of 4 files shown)


code/hud/hudtarget.cpp
```diff
@@ -6521,11 +6522,9 @@ void HudGaugeOffscreen::calculatePosition(vertex* target_point, vec3d *tpos, vec
 	codes_or = (ubyte)(target_point->codes | eye_vertex->codes);
 	clip_line(&target_point,&eye_vertex,codes_or,0);
 
-	if (!(target_point->flags&PF_PROJECTED))
-		g3_project_vertex(target_point);
+	g3_project_vertex(target_point);
 
-	if (!(eye_vertex->flags&PF_PROJECTED))
-		g3_project_vertex(eye_vertex);
+	g3_project_vertex(eye_vertex);
```
Baezon, line 6527: Also, a bit tenuously related, but the preceding `clip_line()` function will *change* the `world` position, so the point should be re-projected into `screen` no matter what! I'm fairly sure the transformation preserves its screen direction, but still.

code/render/3dmath.cpp
```diff
@@ -225,20 +225,25 @@ int g3_project_vertex(vertex *p)
 
 	if ( p->flags & PF_PROJECTED )
 		return p->flags;
+	
+	w = 1.0f / p->world.xyz.z;
```
BMagnu, line 229: Might this need a z == 0 safeguard?

BMagnu (commented):
One question, otherwise LGTM.
As a sidenote: This isn't purely a VR thing, but will cause issues in all cases where the HUD slews, so its required for any form of head tracking

code/render/3dmath.cpp
Baezon: I guess? It didn't before obviously, and it *is* float math, for which we shouldn't expect 0 to naturally crop up. Should we just put in some arbitrarily big number in that case?

BMagnu: IMO, all three components of xyw should probably be zero in that case

Baezon: Although, returning 0,0 would make our subsequent `atan2()` call *also* divide by 0...

BMagnu: I mean, it's using atan2_safe, so _should_ be fine

Baezon: Oh. True.

BMagnu (approved):

PR #5717, opened 2023-10-24
New division by zero coverity issues

JohnAFernandez:
This pr adds checks for the easily fixable divide by zero coverity issues that were uncovered by the most recent scan.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -13044,9 +13044,20 @@ void ai_manage_shield(object *objp, ai_info *aip)
 		//	Scale between 1x and 3x based on ai_class (SUSHI: only if autoscale is on)
 		if (aip->ai_class_autoscale)
 		{
-			if (The_mission.ai_profile->flags[AI::Profile_Flags::Adjusted_AI_class_autoscale])
-				delay = delay + delay * (float)(3 * (ai_get_autoscale_index(Num_ai_classes) - ai_get_autoscale_index(aip->ai_class) - 1) / (ai_get_autoscale_index(Num_ai_classes) - 1));
-			else
+			int autoscale_index = ai_get_autoscale_index(Num_ai_classes);
+
+			// 	Cyborg: Make sure that autoscale index is greater than one before dividing, coverity 1523548
+			if (The_mission.ai_profile->flags[AI::Profile_Flags::Adjusted_AI_class_autoscale] && autoscale_index > 1)
+				delay = delay + delay * (float)(3 * (autoscale_index - ai_get_autoscale_index(aip->ai_class) - 1) / (autoscale_index - 1));
+			else {
+				static bool autoscale_checked = false;
+
+				// warn if there's only one autoscale ai class
+				if (autoscale_index == 1 && !autoscale_checked) { 
+					mprintf(("Warning! Having only 1 autoscale ai class will disable autoscale features that rely on division."));
+					autoscale_checked = true;
+				}
+
 				delay = delay + delay * (float)(3 * (Num_ai_classes - aip->ai_class - 1) / (Num_ai_classes - 1));
 		}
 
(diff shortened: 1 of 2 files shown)


code/ai/aicode.cpp
```diff
@@ -9379,9 +9379,13 @@ void ai_chase()
 										firing_range = MIN((swip->max_speed * swip->lifetime * 0.75f), swip->weapon_range);
 									else
 									{
-										float secondary_range_mult = (aip->ai_secondary_range_mult == FLT_MIN)
-											? (float)(Game_skill_level + 1 + (3 * ai_maybe_autoscale(aip->ai_class)/(ai_maybe_autoscale(Num_ai_classes) - 1)))/NUM_SKILL_LEVELS
-											: aip->ai_secondary_range_mult;
+										float secondary_range_mult;
+										
+										// Cyborg: need to check for division by zero, coverity 1523419
+										if (aip->ai_secondary_range_mult == FLT_MIN && (ai_maybe_autoscale(Num_ai_classes) > 1))
```
Goober5000, line 9385: This will introduce a floating point underflow (multiplying values by `FLT_MIN`) if the multiplier is `FLT_MIN` and `ai_maybe_autoscale` returns 1.

code/ai/aicode.cpp
```diff
@@ -13044,9 +13048,20 @@ void ai_manage_shield(object *objp, ai_info *aip)
 		//	Scale between 1x and 3x based on ai_class (SUSHI: only if autoscale is on)
 		if (aip->ai_class_autoscale)
 		{
-			if (The_mission.ai_profile->flags[AI::Profile_Flags::Adjusted_AI_class_autoscale])
-				delay = delay + delay * (float)(3 * (ai_get_autoscale_index(Num_ai_classes) - ai_get_autoscale_index(aip->ai_class) - 1) / (ai_get_autoscale_index(Num_ai_classes) - 1));
-			else
+			int autoscale_index = ai_get_autoscale_index(Num_ai_classes);
```
Goober5000, line 13051: The `autoscale_index` variable should be renamed `autoscale_num_classes` for clarity.  Note that both the class index and the number of classes are passed to `ai_get_autoscale_index`.

code/ai/aicode.cpp
```diff
@@ -13044,9 +13048,20 @@ void ai_manage_shield(object *objp, ai_info *aip)
...
+			else {
+				static bool autoscale_checked = false;
+
+				// warn if there's only one autoscale ai class
+				if (autoscale_index == 1 && !autoscale_checked) { 
+					mprintf(("Warning! Having only 1 autoscale ai class will disable autoscale features that rely on division."));
+					autoscale_checked = true;
+				}
+
 				delay = delay + delay * (float)(3 * (Num_ai_classes - aip->ai_class - 1) / (Num_ai_classes - 1));
```
Goober5000, line 13065: If `Num_ai_classes` is 1, there will still be a div-0 here.  If that's the case, the best thing to do is to not adjust the delay.

code/ai/aicode.cpp
```diff
@@ -13417,8 +13432,13 @@ int ai_acquire_emerge_path(object *pl_objp, int parent_objnum, int allowed_path_
 			}
 		}
 
-		// cycle through the allowed paths
-		bay_path = allowed_bay_paths[Ai_last_arrive_path % num_allowed_paths];
+		// Cyborg: fall back to standard behavior if no allowed paths to avoid divide by zero -- Coverity 1523287
```
Goober5000, line 13435: Within this if() block, the fact that `allowed_path_mask` is not 0 means there should be guaranteed to be at least one allowed path.  It would be good to add an `UNREACHABLE` for the 0-case.  Is `UNREACHABLE` enough to silence Coverity or is the if() needed too?

Goober5000 (changes requested):

code/ai/aicode.cpp
JohnAFernandez: It probably won't silence coverity, but I can always dismiss the issue after adding the `UNREACHABLE`

code/ai/aicode.cpp
JohnAFernandez: Had a discussion on discord and we decided on `secondary_range_mult = (float)(Game_skill_level + 2)/NUM_SKILL_LEVELS;` as a default, since that is what it comes out to for captain in retail.

code/ai/aicode.cpp
JohnAFernandez: Actually, with an if, it should avoid the issue.

Goober5000 (approved):

PR #5719, opened 2023-10-26
Fix for particle culling

Baezon:
If a particle has an `attached_objnum` then `pos` is *relative* to that object, so keep in mind when calculating distance from camera for culling.


JohnAFernandez (approved):
LGTM
Crazy what one tiny mistake will do.

PR #5726, opened 2023-10-28
check for out-of-range persona

Goober5000:
For missions which use the Persona Index method, make sure the mod doesn't load a persona index that is out of range for the personas it knows about.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -3445,8 +3445,13 @@ int parse_object(mission *pm, int  /*flag*/, p_object *p_objp)
 
 	// parse the persona index if present
 	// For backwards compatbility only
-	if (optional_string("+Persona Index:"))
+	if (optional_string("+Persona Index:")) {
 		stuff_int(&p_objp->persona_index);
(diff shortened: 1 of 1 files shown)


code/mission/missionparse.cpp
```diff
@@ -3445,8 +3445,13 @@ int parse_object(mission *pm, int  /*flag*/, p_object *p_objp)
 
 	// parse the persona index if present
 	// For backwards compatbility only
-	if (optional_string("+Persona Index:"))
+	if (optional_string("+Persona Index:")) {
 		stuff_int(&p_objp->persona_index);
+		if (p_objp->persona_index < -1 || p_objp->persona_index >= (int)Personas.size()) {
+			Warning(LOCATION, "Persona index for %s is out of range (%d)!  Setting to -1.", p_objp->name, p_objp->persona_index);
```
JohnAFernandez, line 3451: This phrasing implies that the range is index. ` Persona index %d for %s` is clearer.

JohnAFernandez (changes requested):

code/mission/missionparse.cpp
Goober5000: Updated

JohnAFernandez (approved):

PR #5729, opened 2023-10-28
Change the scope of this pointer

JohnAFernandez:
This keeps us from using a pointer to a variable that went out of scope. Coverity  1523195


notimaginative:
Would it not be easier to just change the scope of `world_goal_pos` here instead?

JohnAFernandez:
I guess.  I think I always feel nervous relying too much on pointers, but that would be the smaller change.

notimaginative (approved):

PR #5732, opened 2023-10-30
Fix for ets changes when engines are <15%

Baezon:
The ship's maximum velocity is updated when the ship's engine strength changes, but `ship_get_subsystem_strength` always reports 15% when below 15%, causing the code to erroneously believe the engine strength isn't changing (when it is probably auto-repairing). Repairing under this threshold wouldn't cause the max velocity to change, but it is supposed to prevent ETS changes from ignoring the whole engine damage bit, so use the actual engine strength to determine when to apply 'damaged' engine velocities.

### DIFF
--- a/code/hud/hudets.cpp
+++ b/code/hud/hudets.cpp
@@ -125,30 +125,32 @@ void update_ets(object* objp, float fl_frametime)
 	//					 This will translate to 71% max speed at 50% engines, and 31% max speed at 10% engines
 	//
 
-	float engine_aggregate_strength;
+	float effective_engine_strength;
+	float actual_engine_strength;
 	if (ship_p->flags[Ship::Ship_Flags::Maneuver_despite_engines]) {
 		// Pretend our strength is 100% when this flag is active
-		engine_aggregate_strength = 1.0f;
+		effective_engine_strength = 1.0f;
+		actual_engine_strength = 1.0f;
 	} else {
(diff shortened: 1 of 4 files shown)


code/hud/hudets.cpp
```diff
@@ -133,11 +133,16 @@ void update_ets(object* objp, float fl_frametime)
 		engine_aggregate_strength = ship_get_subsystem_strength(ship_p, SUBSYSTEM_ENGINE);
 	}
 
+	// very annoying, but ship_get_subsystem_strength will typically cap at no lower than 15% strength reported
+	// causing the condition below to possibly erroneously believe the engine strength isn't changing while being
+	// repaired below that threshold. use the ACTUAL strength for this check
+	float actual_engine_strength = ship_get_subsystem_strength(ship_p, SUBSYSTEM_ENGINE, false, true);
```
Goober5000, line 139: This should be inside the `else` block a few lines up - it should follow the rules of the `Maneuver_despite_engines` flag.  If the flag is active, it should be 1.0f, just like its counterpart.

Optional: call the variable `actual_engine_aggregate_strength` and rename the other variable, perhaps `effective_engine_aggregate_strength`

Goober5000 (changes requested):

code/hud/hudets.cpp
Baezon: >.> these names are too long
We don't need 'aggregate', that can be assumed.

Goober5000 (approved):

PR #5739, opened 2023-10-30
how about an if/then/else that actually works?

MjnMixael:
Do this with less stupid


MoerasGrizzly:
and here I thought I was doing something wrong this whole time...

JohnAFernandez (commented):
Ok, so the way I'm reading this, it won't work as intended if an order is both onlyValid and onlyEnabled at the same time.  Is that possible?

JohnAFernandez (approved):
That does make sense, and the calls to this function show that.  If you feel inclined, I would add a comment explaining that only one or the other is true, but otherwise

LGTM

PR #5742, opened 2023-10-31
Get rid of remaining int3s in AI code

JohnAFernandez:
Replace many int3's with UNREACHABLE or ASSERTION as necessary.


JohnAFernandez:
Note to self: John, do not merge this until you have fixed the comments.  Thanks!

Issue #5725, opened 2023-10-27
BUG: 'Timestamp_Since' called with invallid Timestamp


DefCynodont:
I have a mission where I repeatedly get 'Assert: "stamp.isValid() && !stamp.isNever()"' We have ruled out any suspect scripts already, and it no SEXPs are firing when this happens, nor does it always happen every time, but when it does it's always in the same place in the mission.

A Log is attached.
[fs2_open.log](https://github.com/scp-fs2open/fs2open.github.com/files/13193782/fs2_open.log)

Any help would be appreciated.

Goober5000:
Unfortunately the FastDebug log doesn't have a meaningful stack trace.  Are you able to generate a log from a regular Debug build?

MjnMixael:
Relevant stack trace

Assert: "stamp.isValid() && !stamp.isNever()"
timer.cpp
Line: 354
timestamp_since was called with an invalid timestamp!

ntdll.dll! NtWaitForSingleObject + 20 bytes
KERNELBASE.dll! WaitForSingleObjectEx + 142 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! dump_stacktrace + 283 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! os::dialogs::AssertMessage + 839 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! timestamp_since + 74 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! training_check_objectives + 164 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! game_simulation_frame + 1704 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! game_frame + 2187 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! game_do_state + 1635 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! game_main + 1059 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! SDL_main + 343 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! main_getcmdline + 245 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! invoke_main + 50 bytes
fs2_open_23_3_0_x64_AVX2-FASTDBG.exe! __scrt_common_main_seh + 302 bytes
[...]

PR #5748, opened 2023-11-01
Event currency tweaks

Goober5000:
In the course of investigating #5725, I made a few minor code enhancements:
1. convert event status to enum
2. split off the event currency functionality from `Sexp_useful_number`, for clarity
3. move where `Sexp_useful_number` (now `Assume_event_is_current`) is set, to increase its visibility and group it with more appropriate code
4. fix `sexp_event_delay_status` to properly set `Assume_event_is_current` in all return paths

### DIFF
--- a/code/mission/missiongoals.cpp
+++ b/code/mission/missiongoals.cpp
@@ -774,18 +774,18 @@ void mission_goal_status_change( int goal_num, int new_status)
 }
 
 // return value:
-//   EVENT_UNBORN    = event has yet to be available (not yet evaluatable)
-//   EVENT_CURRENT   = current (evaluatable), but not yet true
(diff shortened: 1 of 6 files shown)


code/mission/missiongoals.cpp
```diff
@@ -964,7 +964,7 @@ void mission_process_event( int event )
 			Mission_events[event].count = Directive_count;
 		}
 
-		if (Sexp_useful_number){
+		if (Assume_event_is_current){
 			Mission_events[event].flags |= MEF_CURRENT;
```
JohnAFernandez, line 968: Is the check here even needed?  I am guessing it was originally written this way because Assume_event_is_current could be marked as false in the intervening function calls.

JohnAFernandez (approved):

PR #5750, opened 2023-11-01
Provide a way to account for timer padding in the mission log

MjnMixael:
Simply saves the current timer padding when a log entry is created. Provides a way for SCPUI to get the padded time instead of the regular time.

This takes the padded timer feature a little further than I originally intended but it ended up being confusing to have the mission clock say 25+ minutes while messages and events were still happening at -7 minutes. So this provides SCPUI a way to keep the illusion alive if a mod wants to.

I thought about making it default behavior and adding it to the retail UI as well, but that seemed a step too far. Open to discussion on that, though. Regular time is still useful for development and testing purposes because that's the real game clock.

### DIFF
--- a/code/hud/hudmessage.cpp
+++ b/code/hud/hudmessage.cpp
@@ -645,7 +645,7 @@ void hud_add_msg_to_scrollback(const char *text, int source, int t)
 	}
 
 	// create the new node for the vector
-	line_node newLine = {t, source, 0, 1, w, ""};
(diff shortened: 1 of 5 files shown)


Baezon (approved):
I personally, would have it be default and in the retail UI, although I don't think it needs to be necessarily done in this PR. The intention of the feature is quite clear and the usefulness of it only applying 'partially' is surely marginal. It is already opt-in to using this feature, I don't think the user should need to opt into more stuff to make the feature actually work entirely, in what will almost certainly be the most common case.

PR #5751, opened 2023-11-01
upgrade timestamps referenced in freespace.cpp

Goober5000:
Convert additional timestamps to use the new timestamp system.  This covers all timestamps referred to in freespace.cpp.  For purposes of organization, the fade-in/fade-out timestamps have been moved to light.h|cpp and the training timestamps have been moved to missiontraining.h|cpp.

There are also several related code enhancements:
1. `FadeType` has been converted from `enum` to `enum class`
2. Initialization of the various training variables has been added to `training_mission_init` (previously they had never been reset between missions)
3. Validity checks have been added to the relevant locations for these timestamps

### DIFF
--- a/code/globalincs/systemvars.cpp
+++ b/code/globalincs/systemvars.cpp
@@ -12,7 +12,6 @@
 #include "debugconsole/console.h"
 #include "globalincs/pstypes.h"
 #include "graphics/2d.h"
-#include "io/timer.h"
 #include "nebula/neb.h"
 #include "options/Option.h"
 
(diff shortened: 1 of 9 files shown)


JohnAFernandez (commented):
Ok. The code looks good.  To keep things simple I think we should keep the c-casts as they are.  Did you run a test on training missions?

Goober5000:
I had tested missions with capship explosions and cutscene fades, but good point -- I just finished testing targeting and training.  All work.

JohnAFernandez (approved):
Thanks!

PR #5752, opened 2023-11-01
Mod setting to disable FSO's internal loadout restoration

MjnMixael:
SCPUI rolls it's own, much more robust, loadout saving and restoration system. It turns out there are rare cases where the built-in loadout system can conflict with it, so this exposes a toggle to disable the internal one entirely.

Note: this does not disable direct loadout restore which is still used when a player restarts a mission immediately.


wookieejedi (approved):
Very straightforward and works in my tests

JohnAFernandez (approved):

Issue #5365, opened 2023-04-22
FR: Possible applications for curves


Kestrellius:
@Baezon observed that we should collect desirable uses for the new curve system in an FR or something, so here we are.

- Particle length over lifetime (for particles that have that, obviously)

- Laser attributes over lifetime
    - Width
    - Length
    - Alpha

- Beam attributes over lifetime
    - Width
    - Alpha

- Projectile attributes over lifetime
    - Speed
    - Homing turn rate
    - Damage

- Random ranges in particles.tbl

Baezon:
These are mostly good ideas, and I plan on hitting all those I can, but I do need to admit that doing random distributions as curves, while fairly intuitive from a users standpoint, is mathematically almost intractable. This would require integrating (not so bad) and inverting (much worse) the various interpolation functions, which is possible for quadratics (degree 2), possible at great expense for degree 3 and 4, and provably impossible for degree 5 and up, let alone non-integer degrees.

MoerasGrizzly:
- scale gravity coeffecient with speed, so that lift can be simulated in an arcade-ish fashion
- weapon velocity degrading over time as if to simulate athmospheric drag
- scale weapon mass over time (yes, seriously. a few BP folks want the morning star to only push at close range and I'd love to be able to do that without multi-weapon hackery)

PR #5753, opened 2023-11-01
Moar curves

Baezon:
For #5365, adds laser length radius and alpha over time, damage over time, beam alpha over time, homing turn rate over time and shockwave radius over time.

Also removes vestigal `max_lifetime` from weapons.

### DIFF
--- a/code/weapon/shockwave.cpp
+++ b/code/weapon/shockwave.cpp
@@ -12,6 +12,7 @@
 #include "gamesnd/gamesnd.h"
 #include "globalincs/linklist.h"
 #include "io/timer.h"
+#include "math/curve.h"
 #include "model/modelrender.h"
 #include "nebula/neb.h"
 #include "object/object.h"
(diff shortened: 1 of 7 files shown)


code/weapon/shockwave.cpp
```diff
@@ -346,8 +357,7 @@ void shockwave_move(object *shockwave_objp, float frametime)
 			asteroid_hit(objp, nullptr, nullptr, damage, nullptr);
 			break;
 		case OBJ_WEAPON:
-			wip = &Weapon_info[Weapons[objp->instance].weapon_info_index];
-			if (wip->armor_type_idx >= 0)
+			if (wip && wip->armor_type_idx >= 0)
```
JohnAFernandez, line 360: Good catch here

JohnAFernandez (approved):
A lot of good stuff in here.  I honestly didn't spot anything wrong

PR #5757, opened 2023-11-02
Fix Coverity buffer not null terminated issues

JohnAFernandez:
This should fix the outstanding buffer not null terminated issues on coverity.  If there is a better option, especially for the two places that add assertions, please let me know, as c-strings are not my strongest suit.

### DIFF
--- a/code/missionui/missiondebrief.cpp
+++ b/code/missionui/missiondebrief.cpp
@@ -952,10 +952,11 @@ void debrief_choose_voice(char *voice_dest, size_t buf_size, char *voice_base, i
 void debrief_choose_medal_variant(char *buf, int medal_earned, int zero_based_version_index)
 {
 	Assert(buf != NULL && medal_earned >= 0 && zero_based_version_index >= 0);
+	Assertion(strlen(Resolution_prefixes[gr_screen.res]) + strlen(Medals[medal_earned].debrief_bitmap) < MAX_FILENAME_LEN, "The filename for %s is over the filename length limit of 31 characters.  On release builds, this filename will be shortened and will not work.", buf);
 
 	// start with the regular file name, adapted for resolution
-	sprintf(buf, NOX("%s%s"), Resolution_prefixes[gr_screen.res], Medals[medal_earned].debrief_bitmap);
-
+	snprintf(buf, MAX_FILENAME_LEN, NOX("%s%s"), Resolution_prefixes[gr_screen.res], Medals[medal_earned].debrief_bitmap);
+	
 	// if the medal has multiple versions, we may want to choose a specific bitmap
 	char *p = strstr(buf, "##");
 	if (p != NULL)
@@ -975,6 +976,7 @@ void debrief_choose_medal_variant(char *buf, int medal_earned, int zero_based_ve
 
 		sprintf(number, NOX("%.2d"), version);
 		Assert(strlen(number) == 2);
+		// coverity 1522943: Covscan thinks p is not null terminated, but buf, the string that it points to *is* null terminated.
 		strncpy(p, number, 2);
 	}
 }
--- a/code/model/modelanimation.cpp
+++ b/code/model/modelanimation.cpp
@@ -1182,7 +1182,8 @@ namespace animation {
 
 	SCP_string anim_name_from_subsys(model_subsystem* ss) {
 		char namelower[MAX_NAME_LEN];
-		strncpy(namelower, ss->subobj_name, MAX_NAME_LEN);
+		strncpy(namelower, ss->subobj_name, MAX_NAME_LEN - 1);
+		namelower[MAX_NAME_LEN - 1] = '\0';
 		strlwr(namelower);
 		return namelower;
 	}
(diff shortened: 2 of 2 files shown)


code/missionui/missiondebrief.cpp
```diff
@@ -975,7 +975,8 @@ void debrief_choose_medal_variant(char *buf, int medal_earned, int zero_based_ve
 
 		sprintf(number, NOX("%.2d"), version);
 		Assert(strlen(number) == 2);
-		strncpy(p, number, 2);
+		// need to explicitly copy the null terminator, coverity 1522943
+		strncpy(p, number, 3);
```
Goober5000, line 979: This is a false positive.  The intent of this code is to replace exactly two characters (the "##") in the middle of the file name (`buf`).  If the null terminator is copied, the rest of the file name will be truncated.

code/model/modelanimation.cpp
```diff
@@ -1182,6 +1182,7 @@ namespace animation {
 
 	SCP_string anim_name_from_subsys(model_subsystem* ss) {
 		char namelower[MAX_NAME_LEN];
+		Assertion(strlen(ss->subobj_name) < MAX_NAME_LEN, "ss->subobj_name is not null terminated in anim_name_from_subsys: %s", ss->subobj_name);
 		strncpy(namelower, ss->subobj_name, MAX_NAME_LEN);
```
Goober5000, line 1186: This can be replaced with 
```
strncpy(namelower, ss->subobj_name, MAX_NAME_LEN - 1);
namelower[MAX_NAME_LEN - 1] = '\0';
```

code/model/modelanimation.cpp
```diff
@@ -1535,6 +1536,8 @@ namespace animation {
 			auto anim = std::shared_ptr<ModelAnimation>(new ModelAnimation(true));
 
 			char namelower[MAX_NAME_LEN];
+			// Double Check that this string is null terminated since strncpy does not explicitly null terminate Cov 1523269
+			Assertion(strlen(sp->subobj_name) < MAX_NAME_LEN, "sp->subobj_name is not properly null terminated in parseLegacyAnimationTable: %s", sp->subobj_name);
 			strncpy(namelower, sp->subobj_name, MAX_NAME_LEN);
```
Goober5000, line 1541: Similarly here

Goober5000 (changes requested):

code/missionui/missiondebrief.cpp
JohnAFernandez: I had assumed they were at the end. Got it.

JohnAFernandez: Wait is that guaranteed?

Goober5000: Yes.  The `buf` variable is already null-terminated because of the `sprintf`.  If `buf` contains "##", those two characters are replaced with two new characters.  If it doesn't, no replacement occurs.

Goober5000: If you want to improve the buffer safety of `buf`, you could check that the string lengths of `Resolution_prefixes[gr_screen.res]` plus `Medals[medal_earned].debrief_bitmap` are less than or equal to `MAX_FILENAME_LEN - 1` (the `1` is for the null terminator).

code/missionui/missiondebrief.cpp
```diff
@@ -952,10 +952,11 @@ void debrief_choose_voice(char *voice_dest, size_t buf_size, char *voice_base, i
 void debrief_choose_medal_variant(char *buf, int medal_earned, int zero_based_version_index)
 {
 	Assert(buf != NULL && medal_earned >= 0 && zero_based_version_index >= 0);
+	Assertion(strlen(Resolution_prefixes[gr_screen.res]) + strlen(Medals[medal_earned].debrief_bitmap) < MAX_FILENAME_LEN, "The filename for %s is over the filename length limit of 31 characters.  On release builds, this filename will be shortened and will not work.", buf);
```
Goober5000, line 955: Use `%d` instead of writing the number of characters explicitly.  Also the filename is almost certainly truncated on both release _and_ debug.

code/missionui/missiondebrief.cpp
```diff
@@ -952,10 +952,11 @@ void debrief_choose_voice(char *voice_dest, size_t buf_size, char *voice_base, i
...
 
 	// start with the regular file name, adapted for resolution
-	sprintf(buf, NOX("%s%s"), Resolution_prefixes[gr_screen.res], Medals[medal_earned].debrief_bitmap);
-
+	snprintf(buf, MAX_FILENAME_LEN, NOX("%s%s"), Resolution_prefixes[gr_screen.res], Medals[medal_earned].debrief_bitmap);
```
Goober5000, line 958: I think it should be `MAX_FILENAME_LEN - 1`, judging from the other uses of snprintf.  `MAX_FILENAME_LEN` is the buffer size, not the number of non-null characters.  (There are several naming inaccuracies like this throughout the code.)

Goober5000 (changes requested):
A couple additional comments.

Another thing that would be useful is to check all the calling functions and change the buffer declarations of `char[80]` to `char[MAX_FILENAME_LEN]`.  (The buffer size of `MAX_FILENAME_LEN` is eventually enforced after the filename maxes its way to `bm_load`.)

code/missionui/missiondebrief.cpp
JohnAFernandez: It trips an assert on debug, so you're not going to be able to continue, and the user already knows this.  I am encouraging the modder not to ignore this, because it will not do what they want it to do on release.

Goober5000: Mmm, ok

code/missionui/missiondebrief.cpp
JohnAFernandez: Ok, I checked documentation and wrote a test program.  It looks like the size argument should be the length of the destination string and should not adjust for the null terminator.  snprintf will place a null terminator at the end, even if the source string is longer than the destination string.  The only problematic thing that can be done is to specify a length longer than the destination string.  Then the destination string will be null terminated at the start.


Screen Shot 2023-11-12 at 10 52 44 PM

Goober5000: Well it does look like `snprintf` treats its *n* argument differently than `strncpy` does...
> snprintf() copies at most n – 1 characters into the output buffer and then appends a terminating null character.

and
> Writes the results to a character string buffer. At most bufsz - 1 characters are written. The resulting character string will be terminated with a null character.

This means that `MAX_FILENAME_LEN` is, in fact, the correct number to use, and therefore quite a few instances of `snprintf` in the codebase shortchange their strings by 1 char.

PR #5760, opened 2023-11-03
Fix #5759 Sticky Support Ship

### LINKED ISSUE #5759: Sticky Support Ship
Seeing [this](https://github.com/scp-fs2open/fs2open.github.com/issues/3853) or a similar bug can be consistently reproduced on final mission of Frontline 2335 FSO Version : 23.3.0-20231102

After calling support ship support ship docks and does support fine but then the Reload complete message plays twice, your ships location seems to jump and the support ship wont let go.

Telling the support ship to jumps out jumps you out and ends the mission.
Abort rearm does nothing.

Debug log : https://fsnebula.org/log/6544cf0c2250e58fbcded79e
Video https://youtu.be/Wz0GY4affYA (Bug at 0.45)

Not happening on 23.3.0-20231030, So most likely caused by #5611

TheForce172:
Fix #5759 by making sure the code is only run once on undock and add extra check to make sure that the ending code is only run if doing a force repair.

### DIFF
--- a/code/ai/aicode.cpp
+++ b/code/ai/aicode.cpp
@@ -12809,18 +12809,19 @@ void ai_do_repair_frame(object *objp, ai_info *aip, float frametime)
 
 				//	See if fully repaired.  If so, cause process to stop.
 				if ( repaired ) {
-
 					// We can repair ships without being docked, so only do this if we are docked
 					if (repair_aip->submode == AIS_DOCK_4) {
 						repair_aip->submode = AIS_UNDOCK_0;
-					} else { // Undocking ends the repair, but since we don't need to do that in this case, end it here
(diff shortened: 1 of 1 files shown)


JohnAFernandez (commented):
This patch seems to make sense.  Did you make sure that it works in cases other than the buggy one you were fixing?

TheForce172:
Yep, Checked Player Rearm, AI Rearm and Force Rearm. All Work

JohnAFernandez (approved):
Nice work then!

MjnMixael:
Oh nice. Good solution. LGTM, too.

MjnMixael (approved):

PR #5761, opened 2023-11-03
These std moves were suggested by Coverity as optimizations

JohnAFernandez:
I'm not even listing the issue numbers as there were more than 5.  Also, if we switch to c++14 in the future, there would more that we could do in Lambda expressions.


Goober5000 (approved):

TRBlount:
I would practice caution here. Have these changes been tested? I did some digging and while it looks safe at a glance, I would not count on any compilers to catch if the moved variables get reused anywhere.

JohnAFernandez:
Just going to leave it in draft until I can get a smoke test done.

JohnAFernandez:
Smoke test done.  Everything that I can run seems to work after the change.

JohnAFernandez:
Smoke test complete, with nothing to report.  Merging.

PR #5762, opened 2023-11-03
Command fixes and tweaks

### LINKED ISSUE #5749: Translated Command Sender is not sending messages
Heyya,

in my effort to make a small update for my translation library after 18 months (time flies), i found out, that Command does not send messages in every mission in every build after 22.2.0 RC4 in every non-english translation.

After doing some research, I found out that #4440 seems to introduced a problem that $Command Sender: with a XSTR is to be recognized by the game as an invalid message source in foreign languages and thus will not play. As the debug build from 2022-06-26  is giving me following error that confirms that.

> WARNING: "Invalid message source 'Kommandozentrale'.  Unable to send message." at sexp.cpp:13356

In the mission files i am using a $Command Sender: XSTR("Command", 2046), entry, that ensures that all kind of command messages are properly send by 'Kommandozentrale' in the german translation. But the same error appears in the French translation with 'Commandement' aswell, of course.

Goober5000:
In a few places where the game checks for the command sender, it should check for #Command as well, as long as the mission flag to override #Command is present.

Additionally, while the mission's command sender was originally intended to be a ship, it has been used to specify an alternative to #Command as well.  But this caused problems when the "override #Command" mission flag was active, as it transformed a hashed source into a non-hashed source.  So this commit adds a check to treat the source as hashed, if necessary, if a replacement has occurred.

Fixes #5749.

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -1844,7 +1844,10 @@ void message_queue_message( int message_num, int priority, int timing, const cha
 		// of the wave and head
 		// ADDENDUM -- Since the special hack is specifically for mission-unique messages, don't
 		// convert built-in messages to Command
-		if ( builtin_type < 0 && !stricmp(who_from, The_mission.command_sender) ) {
+		if ( builtin_type < 0
+			&& (!stricmp(who_from, The_mission.command_sender)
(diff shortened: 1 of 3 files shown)


JohnAFernandez (approved):
LGTM

PR #5763, opened 2023-11-04
Coverity Unused values Issues

JohnAFernandez:
Some coverity issues in the "Unused Values" category.  In draft because I need to see this patch's effect on targeting large ship subsystems.


code/hud/hudlock.cpp
EatThePath, line 1: I have a hard time intuiting what if anything these changes might have or what might have not been working before hand, but I can say that I tested some multilock, multi-turret-targetting gameplay and all seems to work as before

JohnAFernandez: Thanks for checking!

JohnAFernandez:
Game seems unaffected (as expected) from tests.

Goober5000 (dismissed):

code/hud/hudlock.cpp
```diff
@@ -719,7 +719,7 @@ void hud_do_lock_indicator(float frametime)
 	}
 }
 
-void hud_lock_acquire_current_target(object *target_objp, ship_subsys *target_subsys)
+void hud_lock_acquire_current_target(object *target_objp, ship_subsys **target_subsys)
```
Goober5000, line 722: This will now modify the source target subsys, which is obviously the *intent* of the function but didn't happen before now.  Have the repercussions been checked?

PR #5765, opened 2023-11-05
optimization for checking whether subsystems are blown

Goober5000:
In many calls to `ship_get_subsystem_strength`, the caller doesn't care about the actual number, just whether the subsystem type is destroyed.  So this adds an optimization to check specifically for that.  In the case of `SUBSYSTEM_ENGINE`, this avoids an iteration over every subsystem on the ship which, for a ship with many turrets, can be significant.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -14856,7 +14856,8 @@ float ship_get_subsystem_strength(const ship *shipp, int type, bool skip_dying_c
 	strength = shipp->subsys_info[type].aggregate_current_hits / shipp->subsys_info[type].aggregate_max_hits;
 	Assert( strength != 0.0f );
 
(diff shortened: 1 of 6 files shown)


code/ship/ship.cpp
```diff
@@ -14809,12 +14809,32 @@ int ship_find_subsys(ship *sp, const char *ss_name)
 	return -1;
 }
 
+// an optimization of the below function that skips the subsystem iteration if we don't care about the exact number
+bool ship_subsystems_blown(const ship* shipp, int type, bool skip_dying_check)
+{
+	Assert ( (type >= 0) && (type < SUBSYSTEM_MAX) );
```
JohnAFernandez, line 14815: This PR really is simple.  Just change this to an assertion, and I think we're good to go.

JohnAFernandez (changes requested):

JohnAFernandez (approved):

PR #5766, opened 2023-11-05
Expose more submodel instance and submodel attributes and functions to Lua

BMagnu:


code/scripting/api/objs/modelinstance.cpp
```diff
@@ -256,6 +280,31 @@ ADE_FUNC(findWorldPointAndOrientation, l_SubmodelInstance, "vector, orientation"
...
+{
+	vec3d pnt;
+	matrix orient;
+
+	submodelinstance_h* smih;
+	vec3d local_pnt;
+	matrix_h* local_orient;
+	bool fromWorld = true;
+
+	if (!ade_get_args(L, "ooo|b", l_SubmodelInstance.GetPtr(&smih), l_Vector.Get(&local_pnt), l_Matrix.GetPtr(&local_orient), &fromWorld))
```
Goober5000, line 293: If this function is going from the model or global frame-of-reference to local, then the args here should be named `global_`, not `local_`.  Or maybe the `local_` and non-`local_` named variables just need to be swapped.

code/model/modelread.cpp
```diff
@@ -4768,6 +4768,51 @@ void model_instance_global_to_local_dir(vec3d* out_dir, const vec3d* in_dir, con
...
+	constexpr int preallocatedStackDepth = 5;
+	std::tuple<const matrix*, const vec3d*, const vec3d*> preallocatedStack[preallocatedStackDepth];
+
+	auto submodelStack = pm->submodel[submodel_num].depth <= preallocatedStackDepth ? preallocatedStack : new std::tuple<const matrix*, const vec3d*, const vec3d*>[pm->submodel[submodel_num].depth];
+	int stackCounter = 0;
+
+	int mn = submodel_num;
+
+	//Go up the chain of parents to build a stack of transformations from parent -> child
+	while ((mn >= 0) && (pm->submodel[mn].parent >= 0)) {
```
Goober5000, line 4783: Should this chain have the `use_last_frame` check to select orient vs. prev_orient?

BMagnu: Probably doesn't need it.
It's only included in the functions that explicitly need it for collision. All others don't, including the ones for the other direction.

code/scripting/api/objs/modelinstance.cpp
BMagnu: Fixed the naming

Goober5000 (approved):

PR #5767, opened 2023-11-05
Add a scripting-in-mission state

BMagnu:
If you're wondering how this differs from the existing scripting state, there are three main points:
1. It _does not_ end the mission and clear its data, which allows this to be used during a mission, similar to the pause state.
2. It is not set as the current state, but rather push on top of state stack, allowing to return to the gameplay state.
3. It does not process key events, which means it does not automatically catch stuff like F2 for options, or returns to mainhall on escape.

That makes this overall useful to have pause-like states for scripting and/or SCPUI


wookieejedi:
Works as expected and is highly useful.

JohnAFernandez (approved):
It seems fine to me.  It's more of a game state PR than a scripting PR.  And I think you've changed all the applicable game state places.

PR #5771, opened 2023-11-06
protect model_collide input values

Goober5000:
Add `const` to the `mc_info` struct so its values aren't inadvertently overwritten.  Tweak some of the collision code to prevent overwrites.

Also remove the `flak_muzzle_flash` declaration since that function no longer exists.


JohnAFernandez (approved):
LGTM

PR #5772, opened 2023-11-06
protect eval_enemy_obj_struct input values

Goober5000:
Add `const` to the `eval_enemy_obj_struct` struct so its values aren't inadvertently overwritten.  Handle a bit of const-correctness propagation.  Also add default member initialization to the struct.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -84,16 +84,16 @@ typedef struct eval_enemy_obj_struct {
 	int			current_enemy = -1;
 
 
-	float		nearest_attacker_dist = 0.0f;		// nearest ship	
+	float		nearest_attacker_dist = 99999.0f;		// nearest ship
 	int			nearest_attacker_objnum = -1;
 
-	float		nearest_homing_bomb_dist = 0.0f;	// nearest homing bomb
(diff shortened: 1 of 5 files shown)


code/ai/aiturret.cpp
```diff
@@ -72,29 +72,29 @@ const char* Turret_valid_types[NUM_TURRET_TYPES] = {
...
-	int			current_enemy;
+	const vec3d	*tpos = nullptr;
+	const vec3d	*tvec = nullptr;
+	const ship_subsys *turret_subsys = nullptr;
+	int			current_enemy = -1;
 
 
-	float			nearest_attacker_dist;			// nearest ship	
-	int			nearest_attacker_objnum;
+	float		nearest_attacker_dist = 0.0f;		// nearest ship
```
JohnAFernandez, line 87: These three distances should match the declaration in the one place it is initialized.  Especially since 0.0f would imply that the nearest objects are imminently attacking.

JohnAFernandez (changes requested):

code/ai/aiturret.cpp
Goober5000: Fair point; done

JohnAFernandez (approved):

PR #5773, opened 2023-11-06
convert some instances of frand_range to Random::next

Goober5000:
Some of the `frand_range` calls were used for integer calculations, so we can use `Random::next` now.

This also reduces the number of C-style casts in the codebase.

### DIFF
--- a/code/ship/shipfx.cpp
+++ b/code/ship/shipfx.cpp
@@ -2474,7 +2474,7 @@ void shipfx_do_lightning_frame( ship * /*shipp*/ )
 	shipp->lightning_stamp = timestamp(stamp);
 
 	// ah, now we can create some lightning bolts
-	count = Random::next(0, count);
+	count = Random::next(count+1);
 	while(count > 0){
 		// get 2 points on the hull of the ship
 		v1 = submodel_get_random_point(shipp->modelnum, 0);
@@ -2497,7 +2497,7 @@ void shipfx_do_lightning_frame( ship * /*shipp*/ )
 		binfo.num_strikes = 3;
 		binfo.noise = 0.045f;
 		binfo.life = 375;
-		binfo.delay = Random::next(0, 1600);
+		binfo.delay = Random::next(1600+1);
 		nebl_bolt(&binfo);
 		count--;
 
@@ -2523,7 +2523,7 @@ void shipfx_do_lightning_frame( ship * /*shipp*/ )
 		binfo.num_strikes = 3;
 		binfo.noise = 0.045f;
 		binfo.life = 375;
-		binfo.delay = Random::next(0, 1600);
+		binfo.delay = Random::next(1600+1);
 		nebl_bolt(&binfo);
 		count--;
 	}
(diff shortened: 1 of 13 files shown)


code/ship/shipfx.cpp
```diff
@@ -2474,7 +2474,7 @@ void shipfx_do_lightning_frame( ship * /*shipp*/ )
 	shipp->lightning_stamp = timestamp(stamp);
 
 	// ah, now we can create some lightning bolts
-	count = (int)frand_range(0.0f, (float)count);
+	count = Random::next(0, count);
```
Baezon, line 2477: Looks like some of these can be just `Random::next(max)` now? Trivial, but nice to be consistent

code/ship/shipfx.cpp
```diff
@@ -2497,7 +2497,7 @@ void shipfx_do_lightning_frame( ship * /*shipp*/ )
 		binfo.num_strikes = 3;
 		binfo.noise = 0.045f;
 		binfo.life = 375;
-		binfo.delay = (int)frand_range(0.0f, 1600.0f);
+		binfo.delay = Random::next(0, 1600);
```
Baezon, line 2500: ditto

code/ship/shipfx.cpp
```diff
@@ -2523,7 +2523,7 @@ void shipfx_do_lightning_frame( ship * /*shipp*/ )
...
```
Baezon, line 2526: ditto

code/ship/shipfx.cpp
```diff
@@ -2613,7 +2613,7 @@ void shipfx_do_shockwave_stuff(ship *shipp, shockwave_create_info *sci)
 		sci2.rot_angles.b = frand_range(0.0f, 1.99f*PI);
 		sci2.rot_angles.h = frand_range(0.0f, 1.99f*PI);
 
-		shockwave_create(shipp->objnum, &shockwave_pos, &sci2, SW_SHIP_DEATH, (int)frand_range(0.0f, 350.0f));
+		shockwave_create(shipp->objnum, &shockwave_pos, &sci2, SW_SHIP_DEATH, Random::next(0, 350));
```
Baezon, line 2616: ditto

code/weapon/beam.cpp
```diff
@@ -1523,7 +1523,7 @@ void beam_generate_muzzle_particles(beam *b)
 	b->Beam_muzzle_stamp = timestamp(hack_time);
 
 	// randomly generate 10 to 20 particles
-	particle_count = (int)frand_range(0.0f, (float)wip->b_info.beam_particle_count);
+	particle_count = Random::next(0, wip->b_info.beam_particle_count);
```
Baezon, line 1526: ditto

Baezon (changes requested):

code/ship/shipfx.cpp
Goober5000: Some, but not all.  `Random::next(low, high)` will include both the low and the high value, while `Random::next(modulus)` will include 0 but exclude modulus.  Here, the full count is a valid possibility so the (0, N) call should be used.

Baezon: Ah, I see. Perhaps `Random::next(max+1)`? It's not a big deal, but it's not super obvious just by looking at it.

Also, on second thought, I believe float casting rounds down, so a frand_range will basically never include its top value anyway.

Goober5000: Ok, how's this?

Baezon: Now, I'm starting to have second thoughts lol so I'm just going to approve it before I bikeshed it further and it really isn't a big deal

Baezon (approved):

PR #5776, opened 2023-11-07
FRED import fixes

Goober5000:
Some patches for mission imports:
1. Check that the path isn't empty (the user didn't select "This PC") before continuing
2. Count successful imports
3. Avoid treating an import as a newly saved single mission unless the import was successful


wookieejedi (approved):

PR #5777, opened 2023-11-07
Only untarget jumpnode if we're in a mission

### LINKED ISSUE #5768: Bug: Hard Crash on loading mission with
Screen Shot 2023-11-05 at 6 23 22 PM

In short, mission load chokes on loading the "Hidden By Default" subspace node option.  Not sure if it's a regression because I don't know for sure if anyone has actually used this option successfully.

MjnMixael:
Fixes #5768


wookieejedi (approved):

JohnAFernandez (approved):
Thanks for finding this!  I am still getting my debugger set up.

PR #5778, opened 2023-11-07
Don't ignore user-set -O* flags

freijon:
Currently, the optimization flag is hard-coded to -O2 (for release) and -Og (for debug). It would be better to leave the choice of optimization to the user. See downstream bug: https://bugs.gentoo.org/887671

This PR implements changes that detect if a user-set -O* flag exists. If so, use them. Otherwise, use the previously hard-coded values.

### DIFF
--- a/cmake/toolchain-gcc.cmake
+++ b/cmake/toolchain-gcc.cmake
@@ -113,19 +113,41 @@ if(SUPPORTS_STRINGOP_TRUNCATION)
 	set(COMPILER_FLAGS "${COMPILER_FLAGS} -Wno-stringop-overflow")
 endif()
 
-set(COMPILER_FLAGS_RELEASE "-O2 -Wno-unused-variable -Wno-unused-but-set-variable -Wno-array-bounds -Wno-empty-body -Wno-clobbered  -Wno-unused-parameter")
+# Check if there is a user-set optimisation flag
+string(REGEX MATCH "-O[a-zA-Z|0-9]+" CXX_OPT_FLAG ${CXX_BASE_FLAGS})
+string(REGEX MATCH "-O[a-zA-Z|0-9]+" C_OPT_FLAG ${C_BASE_FLAGS})
 
-set(COMPILER_FLAGS_DEBUG "-Og -g -Wshadow")
+# If no user-set opt flag, set -O2 and -Og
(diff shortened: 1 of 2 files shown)


Goober5000:
CI checks show this:
```
/__w/fs2open.github.com/fs2open.github.com/code/utils/HeapAllocator.cpp: In member function 'void util::HeapAllocator::checkRangesMerged()':
/__w/fs2open.github.com/fs2open.github.com/code/utils/HeapAllocator.cpp:181:9: error: variable 'lastEnd' set but not used [-Werror=unused-but-set-variable]
  size_t lastEnd = 0;
         ^
```
I suppose that's due to the new flag settings.  I'll make a separate PR for that and then you can rebase this PR when that's merged.

cmake/toolchain-gcc.cmake
```diff
@@ -113,19 +113,41 @@ if(SUPPORTS_STRINGOP_TRUNCATION)
...
+	set(C_OPT_FLAG_RELEASE "-O2")
+	set(C_OPT_FLAG_DEBUG "-Og")
+else()
+	set(C_OPT_FLAG_RELEASE "${C_OPT_FLAG}")
+	set(C_OPT_FLAG_DEBUG "${C_OPT_FLAG}")
+endif()
+
+set(CXX_FLAGS_RELEASE "${CXX_OPT_FLAG_RELEASE} -Wno-unused-variable -Wno-unused-parameter")
+set(C_FLAGS_RELEASE "${C_OPT_FLAG_RELEASE} -Wno-unused-variable -Wno-unused-parameter")
+
```
JohnAFernandez, line 138: I only see one thing.  Looks like a copy-paste mistake.  `-Wno-unused-but-set-variable -Wno-array-bounds -Wno-empty-body -Wno-clobbered` are missing in the release flags in this second file.

JohnAFernandez (changes requested):

JohnAFernandez:
Thanks for the contribution, btw.

cmake/toolchain-gcc.cmake
freijon: Good catch, thanks!

JohnAFernandez (approved):

Issue #5780, opened 2023-11-07
Solaris 11.4 (2018 build) GCC 7.3.0 warnings in current master


chief1983:
After looking into whether I can still compile on Solaris with latest master, as I had a previous Solaris warnings ticket open until recently, turns out after patching a few issues, I can compile, which now results in all these new warnings.

I can't use the latest Solaris 11.4 currently, which comes with GCC 11, as it also includes a copy of the library fribidi, which is also installed by the only (outdated) ffmpeg build currently available in an IPS repository, and the two would conflict.  May try to set up a boot environment that has an updated 11.4 and build/install ffmpeg from scratch following a guide I found, but the IPS packaging step wasn't working so far.  Guessing I can just install it like I did SDL2 though, instead of messing with IPS.  All that said, I don't think anything other than the GCC version itself would affect these warnings.

```[ 50%] Building CXX object code/CMakeFiles/code.dir/cmeasure/cmeasure.cpp.o
/usr/share/games/fs2source/fs2_open/code/cfile/cfilesystem.cpp: In function ‘void cf_build_file_list()’:
/usr/share/games/fs2source/fs2_open/code/cfile/cfilesystem.cpp:1175:71: warning: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 4 has type ‘unsigned int’ [-Wformat=]
        "%lu file(s) detected, including:\n%s", count, shadowed.c_str());
                                                                       ^
```

```[ 54%] Building CXX object code/CMakeFiles/code.dir/graphics/line_draw_list.cpp.o
/usr/share/games/fs2source/fs2_open/code/graphics/grbatch.cpp: In member function ‘void geometry_batcher::allocate(int, int)’:
/usr/share/games/fs2source/fs2_open/code/graphics/grbatch.cpp:64:52: warning: declaration of ‘quad’ shadows a global declaration [-Wshadow]
 void geometry_batcher::allocate(int quad, int n_tri)
                                                    ^
In file included from /usr/include/string.h:15:0,
                 from /usr/gcc/7/include/c++/7.3.0/cstring:42,
                 from /usr/share/games/fs2source/fs2_open/code/prefix_header.h:8,
                 from /usr/share/games/fs2source/fs2_open/build/code/cotire/code_CXX_prefix.cxx:4,
                 from /usr/share/games/fs2source/fs2_open/build/code/cotire/code_CXX_prefix.hxx:4:
[...]

PR #5782, opened 2023-11-07
a few build fixes

Goober5000:
1. use `SIZE_T_ARG` instead of `"%lu"`; should fix format warning in #5780
2. since `HeapAllocator::checkRangesMerged()` has no effect unless running in debug, `#ifdef` out the function; should fix CI warning in #5778

### DIFF
--- a/code/utils/HeapAllocator.cpp
+++ b/code/utils/HeapAllocator.cpp
@@ -177,6 +177,7 @@ bool HeapAllocator::check_connected_range(const MemoryRange& left, const MemoryR
 	return left.offset + left.size == right.offset;
 }
 void HeapAllocator::checkRangesMerged() {
+#ifndef NDEBUG
 	bool first = true;
 	size_t lastEnd = 0;
 	for (auto& range : _freeRanges) {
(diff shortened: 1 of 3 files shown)


code/utils/HeapAllocator.cpp
```diff
@@ -177,6 +177,7 @@ bool HeapAllocator::check_connected_range(const MemoryRange& left, const MemoryR
 	return left.offset + left.size == right.offset;
 }
 void HeapAllocator::checkRangesMerged() {
+#ifndef NDEBUG
```
JohnAFernandez, line 180:
```suggestion
// This is a debug only function because its values are only used in debug and linters will get tripped up otherwise.
#ifndef NDEBUG
```

JohnAFernandez: I think it'd be good to mark this with a comment so that we don't have to look through commit history to know what's going on here.  You can write your own version if you want, of course.

JohnAFernandez (changes requested):

code/utils/HeapAllocator.cpp
Goober5000: Good idea.  After a bit of consideration I think this version works. 👍

JohnAFernandez (approved):

PR #5785, opened 2023-11-08
fix invalid pointer access to waypoint lists

Goober5000:
Previously (since 2012!) a waypoint would store its parent list as a pointer.  However, this pointer can become invalidated if the waypoint list vector is resized.  To prevent this, store the parent list as an index.

This also renames the waypoint's `objnum` field to `m_objnum` to match the naming of the other fields.


JohnAFernandez (approved):
LGTM

PR #5787, opened 2023-11-09
refactor argument functions

Goober5000:
Move some special-arg helper functions to sexp.h, as requested by jg18.  While we're at it, also refactor the names and add documentation.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -1100,12 +1100,6 @@ SCP_vector<std::pair<const char*, int>> Sexp_replacement_arguments;
 int Sexp_current_argument_nesting_level;
 
 
-// Goober5000
-bool is_blank_argument_op(int op_const);
-bool is_blank_of_op(int op_const);
-bool is_for_blank_op(int op_const); // jg18
-int get_handler_for_x_of_operator(int node);
-
 //Karajorma
 int get_generic_subsys(const char *subsy_name);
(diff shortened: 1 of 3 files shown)


code/parse/sexp.cpp
```diff
@@ -11713,35 +11707,40 @@ void sexp_change_argument_validity(int n, bool invalidate)
...
-int get_handler_for_x_of_operator(int n)
+/**
+ * Given any SEXP node, return the node of the argument provider operator used in the enclosing when-argument-type SEXP, or -1 if not found.
+ */
+int find_argument_provider(int node)
 {
 	int conditional, arg_handler;
 
-	if (n < 0) {
+	if (node < 0) {
```
jg18, line 11717: Nice readability change. 👍

code/parse/sexp.h
```diff
@@ -1423,6 +1423,7 @@ extern bool sexp_recoverable_error(int num);
 extern const char *sexp_error_message(int num);
 extern int count_free_sexp_nodes();
 
+
```
jg18, line 1426: Optional: maybe remove if not needed.

jg18 (approved):
LGTM. Nice work on renaming and adding documentation. 👍

code/parse/sexp.h
Goober5000: The blank line is intentional - to set off the prototypes in different groups

PR #5788, opened 2023-11-09
rcs thruster refactor

Goober5000:
Here's some low-hanging fruit for reducing the memory footprint of `ship_info`.  This turns the maneuvering thrusters array into a vector, refactors the names a bit, fixes thruster activation when `-nosound` is active, reduces the number of calls to `ship_get_thruster_status`, adds some const-correctness, and converts the timestamps to the new system.

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -229,12 +229,12 @@ flag_def_list Armor_flags[] = {
 const int Num_armor_flags = sizeof(Armor_flags)/sizeof(flag_def_list);
 
 
-flag_def_list_new<Thruster_Flags> Man_types[] = {
+flag_def_list_new<Thruster_Flags> RCS_types[] = {
     { "Bank right", Thruster_Flags::Bank_right, true, false },
     { "Bank left",	Thruster_Flags::Bank_left,	true, false },
     { "Pitch up",	Thruster_Flags::Pitch_up,	true, false },
     { "Pitch down", Thruster_Flags::Pitch_down, true, false },
-    { "Yaw right", Thruster_Flags::Yaw_right, true, false },
+    { "Yaw right",  Thruster_Flags::Yaw_right,  true, false },
     { "Yaw left",	Thruster_Flags::Yaw_left,	true, false },
     { "Slide right",Thruster_Flags::Slide_right,true, false },
     { "Slide left", Thruster_Flags::Slide_left, true, false },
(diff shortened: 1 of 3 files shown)


code/ship/ship.cpp
```diff
@@ -4707,31 +4696,28 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 		}
 	}
 
-	man_thruster *mtp = NULL;
-	man_thruster manwich;
 	while(optional_string("$Thruster:"))
 	{
+		rcs_thruster_info *mtp = nullptr;
+		rcs_thruster_info manwich;
```
Baezon, line 4702: These should probably be renamed, as funny as `manwich` was, it's a bit of a non-sequitur now. `rtp` and `rcs_info` or something

code/ship/ship.cpp
```diff
@@ -9983,65 +9965,70 @@ float ship_get_thruster_status(object* obj, man_thruster* mtp) {
...
 
-		float activated = ship_get_thruster_status(obj, mtp);
+	// make sure there are enough thruster instances
+	while (shipp->rcs_activity.size() < sip->rcs_thrusters.size())
+		shipp->rcs_activity.emplace_back(TIMESTAMP::invalid(), -1, 0.0f);
 
-		vec3d thruster_pos;
-		vm_vec_unrotate(&thruster_pos, &mtp->pos, &obj->orient);
-		vm_vec_add2(&thruster_pos, &obj->pos);
+	for (int i = 0; i < static_cast<int>(sip->rcs_thrusters.size()); i++)
```
Baezon, line 9983: This should be a `size_t i` no? Wouldn't need the cast then and I think the other usages would still work as well.

code/ship/ship.cpp
```diff
@@ -20328,10 +20315,16 @@ void ship_render_batch_thrusters(object *obj)
 
 	if ( Rendering_to_shadow_map ) return;
 
-	for ( int i = 0; i < sip->num_maneuvering; i++ ) {
-		man_thruster *mtp = &sip->maneuvering[i];
+	for (int i = 0; i < static_cast<int>(shipp->rcs_activity.size()); i++)
```
Baezon, line 20318: ditto

Baezon (changes requested):

code/ship/ship.cpp
JohnAFernandez: And nothing is stopping us from putting a comment, too `rcs_trhuster_info rcs_info;  // The variable formerly known as manwich`

Goober5000: Fair enough

code/ship/ship.cpp
Goober5000: Yeah good point

code/ship/ship.cpp
Goober5000: done

Baezon (approved):

PR #5792, opened 2023-11-11
Add ability to remove an enum item from a list

MjnMixael:
What it says on the tin. Takes the item name, finds it in the list, and yeets it into the void.

Also fix up some casts while I'm here.

### DIFF
--- a/code/scripting/api/objs/LuaEnum.cpp
+++ b/code/scripting/api/objs/LuaEnum.cpp
@@ -92,16 +92,15 @@ ADE_FUNC(removeEnumItem,
 	size_t i;
 
 	for (i = 0; i < lua_enum.getEnum()->list.size(); i++) {
-		if (stricmp(item_name, lua_enum.getEnum()->list[i].c_str())) {
-			return ADE_RETURN_FALSE;
-		} else {
-			break;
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/LuaEnum.cpp
```diff
@@ -77,5 +77,32 @@ ADE_FUNC(addEnumItem,
...
+	const char* item_name;
+	if (!ade_get_args(L, "os", l_LuaEnum.Get(&lua_enum), &item_name))
+		return ADE_RETURN_FALSE;
+
+	size_t i;
+
+	for (i = 0; i < lua_enum.getEnum()->list.size(); i++) {
+		if (stricmp(item_name, lua_enum.getEnum()->list[i].c_str())) {
+			return ADE_RETURN_FALSE;
+		} else {
```
JohnAFernandez, line 97: I have to be reading this wrong.  If any item on the list is different than the provided name, abort the function and return false?

MjnMixael: See, this is why I shouldn't test using only the first item in the list. Good catch. Fixed.

JohnAFernandez (approved):
Yes, this works better.

PR #5794, opened 2023-11-11
FRED missionsave updates

Goober5000:
1. Use the `optional_string_fred`/`fout` pattern so the FRED comment parser can follow along with the mission file
2. Since the #Custom Data section is optional and requires a version bump, don't output the section header if the section is empty


wookieejedi (approved):
Straightforward and works for my tests

PR #5798, opened 2023-11-13
Add child enum suffix option

MjnMixael:
This feature adds, hopefully, the final level of arbitrary do-whatever-you-want-ness to lua sexps. The existing `child_enum` parameter type takes whatever the parent parameter's text is and looks for a Lua Enum list with that name. It was originally intended to be able to have the parent be a Lua Enum list of Lua Enum lists. But I realized it can be whatever you want, currently. The parent could a `ship` type and then the game would look for a Lua Enum list with the name of the ship. That's handy, but it's a one-off. You can only have one list with that name.

With this upgrade a `child_enum` can optionally specify a list name suffix. So if the parent parameter is a ship, the child_enum can specify to look for a Lua Enum list that's `shipname + suffix`.

This also downgrades an Error to a log print. If the code can't find the specified enum list we don't have to hard stop. We can just return `<none>` and move on.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -916,6 +916,8 @@ SCP_vector<dynamic_sexp_enum_list> Dynamic_enums;
 
 SCP_vector<dynamic_sexp_parameter_list> Dynamic_parameters;
 
+SCP_vector<dynamic_sexp_child_enum_suffixes> Dynamic_enum_suffixes;
+
 int get_dynamic_parameter_index(const SCP_string &op_name, int param)
 {
 	for (int i = 0; i < (int)Dynamic_parameters.size(); i++) {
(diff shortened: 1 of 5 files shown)


code/parse/sexp.cpp
```diff
@@ -953,6 +955,21 @@ int get_dynamic_parameter_index(const SCP_string &op_name, int param)
 	return -1;
 }
 
+// Gets the custom suffix to append when using child enums. If we have a match
+// for operator name and parameter index, then return the suffix. Otherwise return
+// an empty string
+SCP_string get_child_enum_suffix(const SCP_string& op_name, int param_index)
+{
+	for (int i = 0; i < (int)Dynamic_enum_suffixes.size(); i++) {
```
JohnAFernandez, line 963: Cast

JohnAFernandez (changes requested):
Looks like just that one thing.  At least as far as I can tell.  Sexp code is not my specialty, but on the other hand, it's less than 50 lines.

code/parse/sexp.cpp
MjnMixael: Changed to size_t

JohnAFernandez (approved):

wookieejedi:
Will plan to merge later today

PR #5799, opened 2023-11-13
Add safety to crusty old physics code (coverity issue)

JohnAFernandez:
These few lines of code were what many coverity issues were pointing at.  After looking at the surrounding functions and figuring out what these variables are for, I've determined that nothing is guaranteeing that Tai and Taf are greater than 0.  The only reason why we probably haven't seen issues related to this is that use of the Ai and Af are *currently* gated behind Tai and Taf being > 0 (depending on the variable).  So, that if we ever wanted to expand the avd_movement class, we might run into those values being garbage.  So, we should should ensure that these Ai and Af are set to zero (no acceleration) when no acceleration time is passed to these functions.

In draft until I can run a test.  I do this a lot, but I'm a lot closer to being able to run these kinds of tests than I have before.  

If I did something incorrectly, it would show up in camera movement, especially camera movement performed by sexps.

### DIFF
--- a/code/physics/physics.cpp
+++ b/code/physics/physics.cpp
@@ -1315,8 +1315,24 @@ void avd_movement::setAVD(float final_position, float total_movement_time, float
 	TSi = timestamp();
 	
 	Vm = (Pf-Pi-0.5f*(Vi*Tai)-0.5f*(Vf*Taf)) / (Tf - 0.5f*Tai - 0.5f*Taf);
-	Ai = (Vm-Vi)/Tai;
-	Af = (Vf-Vm)/Taf;
+
+	// Cyborg - Many coverity dividing by zero issues were caused by this piece of low level code. Most places that called this function were also not checking inputs. 
+	// The two if blocks in this function do not guarantee that they will not be float zero (e.g. tai = 2.0 taf = 0.0 Tf = 1.0)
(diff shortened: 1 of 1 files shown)


code/physics/physics.cpp
```diff
@@ -1315,8 +1315,24 @@ void avd_movement::setAVD(float final_position, float total_movement_time, float
 	TSi = timestamp();
 	
 	Vm = (Pf-Pi-0.5f*(Vi*Tai)-0.5f*(Vf*Taf)) / (Tf - 0.5f*Tai - 0.5f*Taf);
-	Ai = (Vm-Vi)/Tai;
-	Af = (Vf-Vm)/Taf;
+
+	// Cyborg - Many coverity dividing by zero issues were caused by this piece of low level code. Most places that called this function were also not checking inputs. 
+	// The two if blocks in this function do not guarantee that they will not be float zero (e.g. tai = 2.0 taf = 0.0 Tf = 1.0)
+	// The reason why this code use to work even when there were divisions by zero is that the NAN results were gated behind Taf > 0.0 and Tai > 0.0
```
Goober5000, line 1321: use to -> used to

Goober5000 (approved):
Great detective work, and good explanation.  One optional typo fix, otherwise it looks good.

JohnAFernandez:
I was finally able to run this today.  Changes seem to be working.

PR #5801, opened 2023-11-13
Allow Modders to Set Rotational Scaling Multiplier for Asteroids/Debris

wookieejedi:
Previously, there was no way to limit the velocity at which asteroids/debris randomly rotated. There was a way to set maximum speed, thus this PR follows that convention and adds a way to cap the rotational velocity scaling. This PR allows modders to set caps per type, including setting the value to 0 for no rotation.

A single scaling value was used b/c it was the most straightforward value to se via table and it was the employed the least code changes. Asteroids pick a random normalized vector for their rotation when spawning, and then that value is scaled by a random multiplier.

### DIFF
--- a/code/asteroid/asteroid.cpp
+++ b/code/asteroid/asteroid.cpp
@@ -401,12 +401,24 @@ object *asteroid_create(asteroid_field *asfieldp, int asteroid_type, int asteroi
 	vec3d rotvel;
 	if ( Game_mode & GM_NORMAL ) {
 		vm_vec_rand_vec_quick(&rotvel);
-		vm_vec_scale(&rotvel, frand()/4.0f + 0.1f);
+		float rot_vel_multiplier = frand()/4.0f + 0.1f;
+		// cap rotational velocity scaling if max is set --wookieejedi
+		if (asip->max_rotational_vel_multiplier >= 0.0f) {
+			vm_vec_scale(&rotvel, MIN(asip->max_rotational_vel_multiplier, rot_vel_multiplier));
+		} else {
+			vm_vec_scale(&rotvel, rot_vel_multiplier);
+		}
 		objp->phys_info.rotvel = rotvel;
 		vm_vec_rand_vec_quick(&objp->phys_info.vel);
 	} else {
 		static_randvec( rand_base++, &rotvel );
-		vm_vec_scale(&rotvel, static_randf(rand_base++)/4.0f + 0.1f);
+		float rot_vel_multiplier = static_randf(rand_base++)/4.0f + 0.1f;
+		// cap rotational velocity scaling if max is set --wookieejedi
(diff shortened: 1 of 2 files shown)


Goober5000 (approved):
LGTM

Baezon (commented):
I'm curious if a cap here is the right way to go, rather than simply an unconditional multiplier? With just a cap, you can't for instance have asteroids that spin really fast, or if the cap is really low, almost all will be capped and so their speeds will be correctly slow, but almost all the same value, without randomness. 

I don't know the use case, so if this is indeed desired then all's well, this PR needn't even implement that, but I thought I'd mention it to confirm

wookieejedi:
> I'm curious if a cap here is the right way to go, rather than simply an unconditional multiplier? With just a cap, you can't for instance have asteroids that spin really fast, or if the cap is really low, almost all will be capped and so their speeds will be correctly slow, but almost all the same value, without randomness.
> 
> I don't know the use case, so if this is indeed desired then all's well, this PR needn't even implement that, but I thought I'd mention it to confirm

A good question indeed. Overall, the thinking was a straightforward way to slow down the rotations while still keeping some randomness. I agree that limits having really fast spinners, though an unconditional multiplier would need a min and a max value I reckon, to allow the randomness to still be incorporated. I'll leave this open for a day or so to see if that's a feature anyone else really might need, then if so perhaps a more complex refactor could be incorporated.

Baezon:
> though an unconditional multiplier would need a min and a max value I reckon, to allow the randomness to still be incorporated.
Since it would just be a multiplier, all the original randomness would remain intact, simply scaled down(or up). I do think that if the goal is "the rotation behavior is fine, just too fast/slow" that an unconditional multiplier sounds like what you're going for. Scaling *the magnitude of the randomness* is separate, and could be done if desired, but the relative magnitude of the randomness will always be preserved by simply scaling it.

wookieejedi:
> > though an unconditional multiplier would need a min and a max value I reckon, to allow the randomness to still be incorporated.
> 
> Since it would just be a multiplier, all the original randomness would remain intact, simply scaled down(or up). I do think that if the goal is "the rotation behavior is fine, just too fast/slow" that an unconditional multiplier sounds like what you're going for. Scaling _the magnitude of the randomness_ is separate, and could be done if desired, but the relative magnitude of the randomness will always be preserved by simply scaling it.

Ah I see what you mean, so it would end up being something like this:
```
vm_vec_scale(&rotvel, (rot_vel_multiplier * frand()/4.0f + 0.1f))
```

Baezon:
Right, where `rot_vel_multiplier` is your new parameter, which would default to 1.

wookieejedi:
Sounds good, I'll change to that method.

wookieejedi:
Has the added benefit of making the code cleaner, too. Thanks!

Baezon (approved):

PR #5802, opened 2023-11-13
Add way to offset directive help text only

MjnMixael:
It's bothered my design sensibilities for a long time that Directives and Directive Help Text are both aligned left. There have even been instances where someone reported they didn't understand a directive in BtA that turned out to be helper text for the directive above.

So to help with clarity this PR adds a way to offset only help text lines to help make things a little more obvious.

### DIFF
--- a/code/hud/hudparse.cpp
+++ b/code/hud/hudparse.cpp
@@ -3708,7 +3708,7 @@ void load_gauge_directives(gauge_settings* settings)
 	char fname_middle[MAX_FILENAME_LEN] = "directives2";
 	char fname_bottom[MAX_FILENAME_LEN] = "directives3";
 	int bottom_bg_offset = 0;
-	int key_line_offset = 0;
+	int key_line_x_offset = 0;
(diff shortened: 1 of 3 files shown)


code/hud/hudparse.cpp
```diff
@@ -3756,6 +3757,9 @@ void load_gauge_directives(gauge_settings* settings)
 	if ( optional_string("Max Line Width:") ) {
 		stuff_int(&max_line_width);
 	}
+	if (optional_string("Key Line Offset:")) {
```
wookieejedi, line 3760: What do you think about calling this Key Line X Offset?

wookieejedi:
This is a very useful enhancement!

code/hud/hudparse.cpp
MjnMixael: Done

wookieejedi (approved):

JohnAFernandez (approved):

PR #5806, opened 2023-11-15
Fix for changing the number of banks on a ship

Baezon:
The number of capacities is initialized to the previous number of banks, and expected to match the number of banks (not talking to the model yet, so just still based on the number of table entries), but the actual listing of banks can change immediately afterward, erroneously triggering the 'incongruous num banks and num capacities' warning if they dont explicitly set the number of capacities to the new number. Initialize *after* any bank number changes may have happened.


wookieejedi (approved):
Good find

JohnAFernandez (approved):

Goober5000 (approved):
~~Seconded~~ Thirded

PR #5807, opened 2023-11-16
Revert removal of allowRelative in `applyToParticleInfo`

Baezon:
Follow-up to #5755. Before, `allowRelative` was always off, and disabled carrying over many possibly valuable bits of information from a particle's source downstream to the particle effect, why should we withhold it? Well what it means by "allow relative" is that by setting an `attached_objnum` we are implying that the particle's velocity and position are *relative* to that object, and it turns out that all weapon effects, inflight effect, homed flight effect, etc use `SourceOriginType::OBJECT` which enables that behavior if we unconditionally 'allow relative'.

None of the particle info math is built to take this into account, so for now, it needs to remain always off. There is other useful data here, so this does not disable as much as it once did, but the key 'relative motion' aspect is `attached_objnum` and `m_offset`. Otherwise, it simply takes global position like the others.

### DIFF
--- a/code/particle/ParticleSource.cpp
+++ b/code/particle/ParticleSource.cpp
@@ -84,15 +84,21 @@ void SourceOrigin::getHostOrientation(matrix* matOut) const {
 	}
 }
 
-void SourceOrigin::applyToParticleInfo(particle_info& info) const {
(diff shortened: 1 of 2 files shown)


wookieejedi (approved):
LGTM

PR #5808, opened 2023-11-16
rename Vaporize Percent Chance

Goober5000:
This option was misleadingly named when originally introduced, so give it a more sensible name.  The old name is still accepted for backwards compatibility.


wookieejedi:
I'm trying to recall, was vaporization chance used in the table b/c that's what it's called in FRED?

Goober5000:
The "vaporize" flag sets both instant explosion and tiny shards, but Swifty's option only controls the former.  Not sure why he picked that name.

wookieejedi:
Oh interesting

wookieejedi (approved):
Works in my tests and code looks good

PR #5809, opened 2023-11-17
Move no campaign warning to on Ready Room click

MjnMixael:
This PR moves the valid campaign check to when the player clicks on Ready Room rather than at Mainhall Init. This offers a few advantages. First it allows getting into the game and not getting immediately greeted with a warning in some cases. Players can set options, view the tech room, etc.. all without having a valid campaign selected. Second, developers can do development on a mod without having a campaign file created yet. This still allows them to play single missions and everything. Mainhalls already default to using mainhall 0 if the campaign is invalid. Essentially this PR serves to lower the friction of the initial in-game experience.

I couldn't find any obvious problems in my testing. As best I can tell, all the appropriate places where this matters also have their own checks in place. For example, switching to Campaigns in the Simulator Room has its own warning popup and then simply stays in the single missions list.

In the place of the original check, this PR adds a log print.

### DIFF
--- a/code/menuui/mainhallmenu.cpp
+++ b/code/menuui/mainhallmenu.cpp
@@ -851,6 +851,8 @@ void main_hall_do(float frametime)
 			break;
 	}
 
+	extern bool Campaign_room_no_campaigns;
+
 	// do any processing based upon what happened to the snazzy menu
 	switch (snazzy_action) {
 		case SNAZZY_OVER:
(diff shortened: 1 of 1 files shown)


Baezon (approved):

JohnAFernandez (approved):

JohnAFernandez:
Honestly, I am highly supportive of this change.  It is a constant pain point for me whenever I am hit by this warning when entering the mainhall.  And if it's not causing any problems, then I can't think of any reason *not* to make this change.

PR #5810, opened 2023-11-17
Remove is-docked check for transfer-cargo sexp

MjnMixael:
Transfer-cargo is a convenience sexp to copy the cargo string from one ship to another and set the source ship cargo string to "nothing". I find it odd that this is limited to only ships that are docked as a mod could have ships that teleport cargo or other sci-fi tropes. Or a mission could need to transfer cargo between two props without docking them. Really, there's lots of possibilities and reasons to do this with ships that aren't docked together. If a FREDer does want to be sure to only transfer cargo between two docked ships, there is the is-docked sexp!

Notably the sexp description does not say it must be between two docked ships, so this PR simply removes that check and let's FREDers run wild with imagination. I can't think of a way this would affect previous uses of this sexp because it would have triggered a warning as-is, but I may be missing something. If anyone thinks it's important, this could be done behind an optional true/false parameter instead.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -15245,12 +15245,6 @@ void sexp_transfer_cargo(int n)
 	if (!ship2 || !ship2->shipp)
 		return;
 
-	// we must be sure that these two objects are indeed docked
-	if (!dock_check_find_direct_docked_object(ship1->objp, ship2->objp)) {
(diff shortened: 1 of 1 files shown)


Baezon (approved):
Obviously Volition had a specific idea of how this would be used, but as you say there's no point in enforcing constraints without good reason.

PR #5811, opened 2023-11-17
implement on-mission-skip SEXP

Goober5000:
Adds a SEXP that specifically checks for whether a player has skipped a mission.  This can be used to set tech room entries, allowed ships/weapons, and other campaign data that would otherwise break if a player skipped a mission.

This only applies to the the failed-five-times popup.  For this to work with the `skipmemymissionyo` cheat, it would be necessary to load every intervening mission and check the events.  That being said, `skipmemymissionyo` already automatically allows all ships and weapons, so there is much less, if any, need for this SEXP in that case.

This has been tested with a simple two-mission campaign and works for allowed ships/weapons, tech room, and scripts.  ~~It is in draft status pending the resolution of #5506 and #4634~~.

### DIFF
--- a/code/mission/missioncampaign.cpp
+++ b/code/mission/missioncampaign.cpp
@@ -1642,18 +1642,18 @@ void mission_campaign_end_close()
 
 /**
  * Skip to the next mission in the campaign
- * this also posts the state change by default.  pass 0 to override that
+ * this also posts the state change
  */
-void mission_campaign_skip_to_next(int start_game)
+void mission_campaign_skip_to_next()
 {
+	// mark mission as skipped
+	Campaign.missions[Campaign.current_mission].flags |= CMISSION_FLAG_SKIPPED;
+
 	// mark all goals/events complete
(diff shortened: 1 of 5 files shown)


Goober5000:
With the merges of 5506 and 4634, this now works for the debriefing popup as well as the on-death popup; so it is now ready for review.

JohnAFernandez (approved):
This does seem like it's fine.  I'm just anticipating some possible side effects on events that are not on mission skip.  It should be checked that events not marked with on-skip-mission are not affected.

Goober5000:
> This does seem like it's fine. I'm just anticipating some possible side effects on events that are not on mission skip. It should be checked that events not marked with on-skip-mission are not affected.

Yes, it is checked.  The function `mission_goal_mark_events_complete()` explicitly runs only the SEXPs where `on-mission-skip` is the top-level operator.  Nothing about the event itself is modified, and nothing happens for any other situation.

Issue #5378, opened 2023-04-27
A FR for general ship trail enhancements


LaserWitch:
Desired new table options
 * Boolean controlling if trail is a contrail (present in nebula only) or not (present in all missions)
 * Boolean controlling if trail threshold measures speed in absolute terms or relative to a direction of travel.
 * Vector normal controlling what direction the speed will be measured in, intended to permit different contrails to appear when flying or drifting in different directions
 * Float controlling minimum speed trail appears at.

Contrail speed threshold is currently set on the mission level. I propose treating that as a multiplier, based on the default threshold compared to the mission threshold, and then applying that to contrail type trails only. Non-contrails are desired for engine trails or light streaks and so should not be affected by nebula related settings.

Replaces my previous request of #5274

Kestrellius:
A possibility that occurs to me: use a curve for a trail alpha multiplier over speed (relevant type of speed, whatever that might be given settings), so that the trail can gradually fade in.

Issue #5274, opened 2023-03-20
A Request for Direction-Blind Nebula Contrails


LaserWitch:
Please allow options for contrails to be created when ships are traveling directions other than forwards

LaserWitch:
closed in favor of #5378

PR #5815, opened 2023-11-18
Implement direction-blind contrails option.

MageKing17:
One of the features requested in #5378 (specifically the one requested in #5274), this adds a new "$Contrails use absolute speed:" option to game_settings.tbl, which makes the `ct_below_limit()` function check `objp->phys_info.speed` instead of `objp->phys_info.fspeed`. This allows contrails to appear when moving in a direction other than forwards.


wookieejedi (approved):
Useful update and very straightforward

JohnAFernandez (approved):

Issue #5494, opened 2023-07-13
FRED: Team Loadout Editor Issues/Bugs


LuytenKy:
The first primary issue is that the FRED Team Loadout Editor has UI glitches that occur due to poorly formatted UI elements, primarily the checkboxes. Additionally, FRED randomly reassigns boxes that have already been unchecked.

The checkboxes are placed differently in the original version of FreeSpace, In such a way that they are spread apart from each other to avoid overlapping.

This is how it's displayed in the original.
![OG Loadout](https://github.com/scp-fs2open/fs2open.github.com/assets/135379429/517d6fd7-b0cb-4a51-b2da-179d539c61bb)

And this is how it's displayed in FSO, unchecking the boxes causes overlapping.
![FSO TeamLoadout](https://github.com/scp-fs2open/fs2open.github.com/assets/135379429/a46a4e45-8620-454c-93f4-7f5deffefafa)

A possible fix for this issue is to rearrange the checkboxes to avoid overlapping.

All versions of FreeSpace including the original have a bug in which the checkboxes are randomly rechecked after they have already been unchecked.

Note that the number of reassigned boxes may differ.

To recreate it, create a new FRED mission, go to the Team Loadout editor, uncheck all boxes, click Ok, reopen the Team Loadout editor, and you will see that some boxes will be reassigned again.

Goober5000:
Hmm.  What FSO build are you using, and what operating system are you running on?  The checkboxes do not overlap for me running the latest nightly on Windows 11.

Goober5000:
I've confirmed the random weapon selection bug though.

Goober5000:
This is proving rather tricky to track down.  There are a lot of interlinked dependencies and spaghetti logic.  I'm going to focus on other tasks for the time being.

LuytenKy:
> Hmm. What FSO build are you using, and what operating system are you running on? The checkboxes do not overlap for me running the latest nightly on Windows 11.

I'm using FSO 23.0.0, With Windows 10.

Goober5000:
Well I may have figured this out while working on a different bugfix.

Goober5000:
Fixed via #5817 and #5819

PR #5817, opened 2023-11-18
fix loadout dialog updating in FRED

Goober5000:
We need to call `UpdateData()` in the `OnOK()` function and in several other places so that the member variables are populated with the most recent input from the UI.  This prevents certain input (such as Player Entry Delay changes) from being lost by being overwritten with stale stored values.

This fixes a bug reported by Kitsune where the Player Entry Delay box did not work properly.  It also fixes the random checkbox issue from #5494.

Many thanks to Kitsune for assistance with testing.


JohnAFernandez (approved):
LGTM

PR #5818, opened 2023-11-18
Remove redundant batching_render_all calls

EatThePath:
We're currently calling batching_render_all 2-3 times a frame(one appears to be either optimized out or optimized in such a way profiling can't see it), and I can't see any reasons why that is needed or useful. Buffer submission calls are a serious bottleneck in my data, and while my testing has been somewhat loose, improvement from consolidating is clear. On a RTX 4080, in a particle-heavy situation with uncapped framerates, I've seen lows improve from around 80fps to around 140fps, around 5ms frametime improvement.

I've tried to find any code paths where this would result in batches not being rendered, or effects being rendered differently and not found anything yet.


Baezon (approved):
Let's see what happens.

JohnAFernandez:
Time to hit the button, I think

PR #5825, opened 2023-11-20
Fix for lasers with head on bitmaps

Baezon:
Follow-up to #5753. Gotta recalculate the ubyte alpha here from the float 0-1 so don't forget the 255.0


wookieejedi (approved):

JohnAFernandez (approved):

JohnAFernandez:
I'm going to merge so it's in the next nightly, despite it not being that long since it was posted.

PR #5826, opened 2023-11-21
Scpui language selector

MjnMixael:
Adds the ability to select the game's language from the in-game options menu with SCPUI. Right now this requires the game to be restarted because many xstr strings are initialized on game_init and not when they are actually displayed/needed. This could be changed in the future.

This also adds a feature to strings.tbl to assign an XSTR to the language name itself.

This also adds a feature to game_settings.tbl to prevent FSO from initializing the three additional built-in languages. This allows mods who don't support those languages to remove them from the SCPUI language selector.

### DIFF
--- a/code/localization/localize.cpp
+++ b/code/localization/localize.cpp
@@ -19,6 +19,7 @@
 #include "parse/parselo.h"
 #include "playerman/player.h"
 #include "mod_table/mod_table.h"
+#include "options/Option.h"
 
 #include <tl/optional.hpp>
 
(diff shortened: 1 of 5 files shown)


code/localization/localize.cpp
```diff
@@ -97,17 +98,57 @@ SCP_unordered_map<int, char*> Lcl_ext_str_explicit_default;
...
+	return vals;
+}
+
+static SCP_string language_display(int value)
+{
+	return XSTR(Lcl_languages[value].lang_name, Lcl_languages[value].xstr);
+}
+
+// Used to persist the language selection
+int Language_choice_index = -1;
```
JohnAFernandez, line 117: So it looks like you're using this to have the language choice persist.  Is there any danger of this index pointing to the wrong language if one mod is closed and another mod with different languages is opened?

code/localization/localize.cpp
```diff
@@ -140,41 +177,43 @@ void lcl_init(int lang_init)
 
 	// if we only have one language at this point, we need to setup the builtin languages as we might be dealing with an old style strings.tbl
 	// which doesn't support anything beyond the builtin languages. Note, we start at i = 1 because we added the first language above.
-	if ((int)Lcl_languages.size() == 1) {
-		for (i=1; i<NUM_BUILTIN_LANGUAGES; i++) {
+	if ((No_built_in_languages == false) && (static_cast<int>(Lcl_languages.size()) == 1)) {
```
JohnAFernandez, line 180: You can do `!No_built_in_languages`.  I think most would find that clearer.

code/localization/localize.cpp
```diff
@@ -140,41 +177,43 @@ void lcl_init(int lang_init)
...
 
 			// look it up
-			for (idx = 0; idx < (int)Lcl_languages.size(); idx++) {
-				if (!stricmp(Lcl_languages[idx].lang_name, lang_string)) {
-					lang = idx;
-					break;
-				}
+			for (int idx = 0; idx < static_cast<int>(Lcl_languages.size()); idx++) {
+				lang = lcl_find_lang_index_by_name(lang_string);
 			}
```
TRBlount, line 217: Why is there no way for a successful lang to escape the loop?

code/localization/localize.h
```diff
@@ -37,6 +37,7 @@
 // language info table
 typedef struct lang_info {
 	char lang_name[LCL_LANG_NAME_LEN + 1];				// literal name of the language
+	int xstr;											// the xstr id for the language name itself. Used for the built-in langauge selector
```
TRBlount, line 40: More of a nitpick. Language is misspelled as "langauge" in the later part of this string.

code/localization/localize.cpp
MjnMixael: Line 194 addresses that issue.

code/localization/localize.h
MjnMixael: @JohnAFernandez promised I got all those typos! 😛

code/localization/localize.cpp
MjnMixael: I'm realizing you probably meant saving the index into the array rather than the language name could lead to language silliness between mods. I'll have to think about that. I was originally saving the string name but it wasn't playing nice with the options manager.

MjnMixael: So this sent me down one hell of a rabbit hole but I ended up learning a lot about the `OptionsBuilder` parameters in the process. Now this saves the language selection as json data `{"name": LANGNAME, "ext", LANGEXT}` and uses both to try and match the language exactly. If it doesn't match both strings then the language selector will go for the command line and then finally the old ini setting if everything else fails.

JohnAFernandez: Good job!

JohnAFernandez (approved):

PR #5829, opened 2023-11-22
replace pointers with indexes in ship registry and oswpt

Goober5000:
This refactors the ship registry and the OSWPT struct to store data as indexes rather than pointers.  This is a prerequisite for making these collections dynamic in the future.  See also #5800.

~~In draft pending tests.~~ Tested and self-reviewed; ready for peer review.


EatThePath (approved):
With the number of things touched I don't think I could ever do enough testing to make sure everything is working right, but I think this is a bullet that needs biting sooner than later. I've done my best to scan through the changes and it all seems sensible and consistent to me.

Goober5000:
I tested several missions in several mods, and I tried a bunch of situations likely to cause problems if problems were to show up.  Everything worked as intended.

Thanks for the review!

PR #5831, opened 2023-11-23
add Lua function for movie duration

Goober5000:
This retrieves the duration from the movie stream.  Thanks to @asarium for pointing me in the right direction.

### DIFF
--- a/code/scripting/api/objs/movie_player.cpp
+++ b/code/scripting/api/objs/movie_player.cpp
@@ -53,10 +53,10 @@ ADE_VIRTVAR(FPS, l_MoviePlayer, "number", "Determines the frames per second of t
 {
 	movie_player_h* ph = nullptr;
 	if (!ade_get_args(L, "o", l_MoviePlayer.GetPtr(&ph)))
-		return ade_set_args(L, "i", -1);
+		return ade_set_args(L, "f", -1);
 
 	if (ph == nullptr || !ph->isValid()) {
-		return ade_set_args(L, "i", -1);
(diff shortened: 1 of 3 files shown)


code/scripting/api/objs/movie_player.cpp
```diff
@@ -66,6 +66,24 @@ ADE_VIRTVAR(FPS, l_MoviePlayer, "number", "Determines the frames per second of t
...
 
+ADE_VIRTVAR(Duration, l_MoviePlayer, "number", "Determines the duration in seconds of this movie <b>Read-only</b>", "number",
+            "The duration of the movie or -1 if handle is invalid")
+{
+	movie_player_h* ph = nullptr;
+	if (!ade_get_args(L, "o", l_MoviePlayer.GetPtr(&ph)))
+		return ade_set_args(L, "f", -1);
+
+	if (ph == nullptr || !ph->isValid()) {
+		return ade_set_args(L, "f", -1);
```
JohnAFernandez, line 77: I couldn't find any other places where we return -1 as a float in response to an invalid value.  Would an lua script which checks that the float return value == -1 work?

Goober5000: Well I was following the precedent of `FPS`, which does return -1.  That doesn't mean I'd always do it that way, but I figured I should be consistent within the same class.

A Lua script which checks == -1 should indeed work, but the recommended function is `isValid`, which can be found about 100 lines farther down in movie_player.cpp.

JohnAFernandez (approved):

PR #5833, opened 2023-11-26
Update PR build image and compilers

asarium:
This advances the versions for Clang to 16 and GCC to 13 along with fixing any warnings these new versions are detecting now.

The new image versions also include Vulkan so the additional workflow step for installing that is not necessary anymore.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,20 +47,6 @@ endif()
 
 PROJECT(FS2_Open LANGUAGES ${FSO_LANGUAGES})
 
-find_program(CCACHE_PROGRAM ccache)
-# Currently only GCC support ccache properly without generating additional warnings
-# Enable ccache on Mac even for clang
-if(CCACHE_PROGRAM AND ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR APPLE))
-	if(NOT CMAKE_C_COMPILER_LAUNCHER)
-		message(STATUS "Using ccache for C compilation: ${CCACHE_PROGRAM}")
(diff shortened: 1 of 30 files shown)


CMakeLists.txt
```diff
@@ -47,20 +47,6 @@ endif()
 
 PROJECT(FS2_Open LANGUAGES ${FSO_LANGUAGES})
 
-find_program(CCACHE_PROGRAM ccache)
-# Currently only GCC support ccache properly without generating additional warnings
-# Enable ccache on Mac even for clang
-if(CCACHE_PROGRAM AND ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR APPLE))
```
asarium, line 53: Previously, this took away the choice from the user to use or not use ccache and caused issues for me locally so I just removed this since it serves no purpose.

code/globalincs/pstypes.h
```diff
@@ -288,7 +288,7 @@ constexpr bool LoggingEnabled = false;
 // Disabling this functionality is dangerous, crazy values can run rampant unchecked, and the longer it's disabled
 // the more likely you are to have problems getting it working again.
 #if defined(NDEBUG)
-#	define Assert(expr) do { ASSUME(expr); } while (false)
+#	define Assert(expr) do { ASSUME(expr); (void)sizeof(expr); } while (false)
```
asarium, line 291: These changes will mark the values used in the expression as "used" and prevent warnings in release builds. This should have no effects on the generated code since everything is discarded but some usages of the assertion macros will now no longer work and required some adjustments.

Goober5000 (approved):

PR #5835, opened 2023-11-27
Verify this is in bounds and add vector_inbound macro

MjnMixael:
This language check should have verified the index is in bounds for the vector, so this PR fixes that. 

It also adds a method and an overload to do the inbounds check a little more clearly where needed in the future.

### DIFF
--- a/code/globalincs/vmallocator.h
+++ b/code/globalincs/vmallocator.h
@@ -26,6 +26,16 @@ bool SCP_vector_contains(SCP_vector<T>& vector, T item) {
 }
 
 template <typename T>
+bool SCP_vector_inbounds(SCP_vector<T>& vector, int idx) {
+	return ((idx >= 0) && (idx < static_cast<int>(vector.size())));
+}
+
+template <typename T>
+bool SCP_vector_inbounds(SCP_vector<T>& vector, size_t idx) {
+	return ((idx >= 0) && (idx < vector.size()));
+}
+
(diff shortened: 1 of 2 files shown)


Baezon (approved):

wookieejedi (approved):

code/globalincs/vmallocator.h
```diff
@@ -25,6 +25,16 @@ bool SCP_vector_contains(SCP_vector<T>& vector, T item) {
...
 }
 
+template <typename T>
+bool SCP_vector_inbounds(SCP_vector<T>& vector, int idx) {
+	return ((idx >= 0) && (idx < static_cast<int>(vector.size())));
+}
+
+template <typename T>
+bool SCP_vector_inbounds(SCP_vector<T>& vector, size_t idx) {
+	return ((idx >= 0) && (idx < vector.size()));
```
JohnAFernandez, line 35: size_t is unsigned so idx >= 0 will always be true.  Some compilers will even complain about comparisons like that.

JohnAFernandez (changes requested):
Quick thing

code/globalincs/vmallocator.h
JohnAFernandez, line 35:
```suggestion
	return idx < vector.size();
```

For convenience.  Feel free to ignore if you have already made changes or want to alter the formatting.

PR #5838, opened 2023-11-27
constify some turret code

Goober5000:
Apply `const` to a whole bunch of functions.  Also change a bunch of ints to bools.

Depends on #5837, ~~so in draft until that is merged~~.  This branch is just a refactor with no change in functionality.


wookieejedi (approved):
Overall looks good and also works as expected in my tests. Happy to have anyone else look this over if they wanted to.

PR #5839, opened 2023-11-28
make radar timestamps type-safe

Goober5000:
Also convert true/false variables to boolean.


JohnAFernandez (approved):
Ok, I couldn't see anything wrong.

PR #5840, opened 2023-11-28
Add cockpit 'sway'

Baezon:
Just adds a portion of the ship's acceleration to the camera, probably quite dependent on the size of the cockpit model so it's a multiplier the modder can fiddle with to their content along with a scripted toggle. Also normalizes naming of `cockpitActive` and `disableCockpits`

### DIFF
--- a/code/hud/hud.cpp
+++ b/code/hud/hud.cpp
@@ -1178,9 +1178,9 @@ bool HudGauge::canRender()
 	if (render_for_cockpit_toggle > 0) {
 		if ((Viewer_mode & VM_CHASE) && (render_for_cockpit_toggle == 2)) {
 			return true;
-		} else if (cockpitActive && (render_for_cockpit_toggle == 2)) {
+		} else if (Cockpit_active && (render_for_cockpit_toggle == 2)) {
 			return false;
-		} else if (!cockpitActive && (render_for_cockpit_toggle == 1)) {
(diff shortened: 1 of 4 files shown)


wookieejedi (approved):
Works as expected and code is very straightforward

JohnAFernandez (approved):
Don't want to use a define or constexpr for the 5000 factor?

Baezon:
Eh, I mean I thought about it, it's ultimately a completely arbitrary number I just didn't want users to have to put in really small numbers, I put it in the ballpark of 1 for them, but if it's a big deal I'd just remove it and punt the responsibility to modders.

JohnAFernandez:
What I mean is that its better to have it in a define or constexpr so that if we did want to change or trace through code in the future, we would only have to deal with one definition instead of the three different ones.  And if we did want to expose it to modders in the future, then that change would be easier, too.

Baezon:
Yes, I understand what you mean. I am saying that it is already exposed as `+Cockpit Sway Multiplier:`. It would work perfectly fine if the 5000 were completely removed, since the multiplier could be adjusted to account for it. It's not a meaningful value, it is completely arbitrary, which is why I don't want to add a constant for it.

wookieejedi:
Now that all comments incorporated, I will plan to merge this later today if no one else beats me to it.

PR #5844, opened 2023-11-29
FRED on mission load hook

MjnMixael:
Quick FRED hook to run when a mission has been loaded. This will allow scripts to update any Lua Enums based on the parsed mission's data.


JohnAFernandez (approved):
LGTM

PR #5845, opened 2023-11-29
Scpui multi pxo api

MjnMixael:
This is the complete API for SCPUI Multi PXO and Multi PXO Help game states. Relies on #5830, so draft until that is merged.

### DIFF
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -2530,8 +2530,10 @@ ADE_FUNC(getChat, l_UserInterface_MultiPXO, nullptr, "Gets the entire chat as a 
 
 	LuaTable chats = LuaTable::create(L);
 
-	for (auto i = 0; i < Multi_pxo_chat.size(); i++) {
-		chats.addValue(i + 1, Multi_pxo_chat[i].text);
+	for (size_t i = 0; i < Multi_pxo_chat.size(); i++) {
+		SCP_list<chat_line>::iterator line = Multi_pxo_chat.begin();
+		std::advance(line, i);
+		chats.addValue(i + 1, line->text);
 	}
 
 	return ade_set_args(L, "t", chats);
@@ -2556,7 +2558,7 @@ ADE_FUNC(getPlayers, l_UserInterface_MultiPXO, nullptr, "Gets the entire player 
 
 	LuaTable chats = LuaTable::create(L);
 
-	for (auto i = 0; i < Multi_pxo_players.size(); i++) {
+	for (size_t i = 0; i < Multi_pxo_players.size(); i++) {
 		chats.addValue(i + 1, Multi_pxo_players[i]);
 	}
 
--- a/code/scripting/api/libs/ui.cpp
+++ b/code/scripting/api/libs/ui.cpp
@@ -2548,7 +2548,7 @@ ADE_FUNC(getChat,
 
 		if (spacePos != nullptr) {
 			// Calculate the length of the callsign
-			int len = spacePos - line.text;
+			int len = static_cast<int>(spacePos - line.text);
 
 			// Copy the callsign
 			std::strncpy(callsign, line.text, len);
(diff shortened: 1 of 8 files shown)


notimaginative:
This needs a rebase, doesn't it? Looks like all the vectorization stuff is here again.

MjnMixael:
> This needs a rebase, doesn't it? Looks like all the vectorization stuff is here again.

Ah yup. Done.

code/scripting/api/libs/ui.cpp
```diff
@@ -2479,5 +2481,309 @@ ADE_FUNC(closePause, l_UserInterface_PauseScreen, nullptr, "Makes sure everythin
...
+	sprintf(cmd, "/WHOIS %s", plr_name);
+
+	multi_pxo_chat_send(cmd.c_str());
+
+	// We have to process the pxo api in order for the response to come back
+	while (stricmp(User_req_channel, "") == 0) {
+		if (Multi_pxo_connected) {
+			multi_pxo_api_process();
+		}
+	}
```
notimaginative, line 2634: Check `Multi_pxo_connected` first thing in the function since nothing here works otherwise.

And unless this is all running in a separate thread the `while` loop here is going to block everything until it fills in `User_req_channel`. If there is an error or the user isn't on a channel or the user is/goes offline then that will never happen. But ideally you should just use `GetChannelByUser()` calls here instead of inventing something new.

Also you need to make a call to `os_sleep(10)` between calls to `multi_pxo_api_process()` to give the system time to handle network stuff.

code/scripting/api/libs/ui.cpp
```diff
@@ -2479,5 +2481,309 @@ ADE_FUNC(closePause, l_UserInterface_PauseScreen, nullptr, "Makes sure everythin
...
+
+	SCP_string response;
+	SCP_string channel;
+
+	if ((ptr_s)User_req_channel == -1) {
```
notimaginative, line 2639: `User_req_channel` is never -1. That is a return value from `GetChannelByUser()` in case of error.

notimaginative (changes requested):

code/scripting/api/libs/ui.cpp
```diff
@@ -2479,5 +2481,309 @@ ADE_FUNC(closePause, l_UserInterface_PauseScreen, nullptr, "Makes sure everythin
...
+{
+	SCP_UNUSED(L);
+
+	using namespace luacpp;
+
+	LuaTable chats = LuaTable::create(L);
+
+	for (size_t i = 0; i < Multi_pxo_chat.size(); i++) {
+		SCP_list<chat_line>::iterator line = Multi_pxo_chat.begin();
+		std::advance(line, i);
```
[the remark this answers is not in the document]
JohnAFernandez: I think I mentioned it on the last review.

code/scripting/api/libs/ui.cpp
MjnMixael: I have a vague memory of trying `GetChannelByUser()` but ran into some issue relating to how the UI through SCPUI runs things differently. But I dunno. It's been so long since I wrote all this that I really don't remember anything concrete anymore and I don't currently have a ton of motivation to dive back in and figure it all out again. I've made the easy change you requested. I guess the rest will have to wait until I feel up to getting neck deep in the multi code again.

code/scripting/api/libs/ui.cpp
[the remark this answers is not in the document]
MjnMixael: Done

code/scripting/api/libs/ui.cpp
MjnMixael: Alright, I dove back into this and remembered I set it up the way I did originally so that the Lua side was forced to wait for a valid response. I still think that's the better way here given how this is probably going to be used in the UI but I changed it to use GetChannelByUser() and the Lua script will have to be smart enough to handle looping or threading as necessary.

code/scripting/api/libs/ui.cpp
MjnMixael: No longer relevant.

code/scripting/api/libs/ui.cpp
notimaginative: I meant to use GetChannelByUser(), but otherwise still do it the same way. That way you use the existing code for the request and error handling. So something like...
```
char *channel = nullptr;

GetChannelByUser(plr_name);

while (channel == nullptr) {
    os_sleep(10);
    multi_pxo_api_process();
    channel = GetChannelByUser(nullptr);
}
```
Then have your previous checks below that but using `channel` instead of `User_req_channel`.

Basically the first call to GetChannelByUser() initializes it and sends the request (return value doesn't matter on first call). Then you call it repeatedly with nullptr and check the return value until it's non-null. The return value will then either be "-1" if an error occurred, or the channel name. 

With your recent change it actually wouldn't work at all. Sorry for the confusion there.

notimaginative (approved):

PR #5847, opened 2023-11-29
Remove the cd checks that always returned true anyway

MjnMixael:
Per discussion in #scp, this removes the checks for a cd-rom. The checks always returned a val of 1 anyway.

I left the bits in place for the network packets because I wasn't sure what should be done there.


JohnAFernandez (approved):
Looks good.

PR #5848, opened 2023-11-29
Fix medal parsing issue

### LINKED ISSUE #5846: 23.2.1 Every medal in debriefing screen is the SOC unit crest
Basically title. Medal case is working fine, afaict.

MjnMixael:
Fixes #5846 by making sure we only load the retail FS2 values if we're parsing a tbl and not a tbm. This also moves the data verification to a post process method that is ran after all tbls and tbms are parsed.


JohnAFernandez (approved):
LGTM

PR #5850, opened 2023-12-01
Convert multi team chat to use a struct/list setup

MjnMixael:
A cleanup/refresh of storing and displaying team chat lines using a struct/SCP_list instead of an array of char arrays where various message parameters were stored and parsed out of the string itself.

### DIFF
--- a/code/missionui/chatbox.cpp
+++ b/code/missionui/chatbox.cpp
@@ -816,10 +816,18 @@ void chatbox_render_chat_lines()
 	int ly = Chatbox_begin_y;
 	int line_height = gr_get_font_height() + 1;
 
-	for (int i = Brief_start_display_index; i < static_cast<int>(Brief_chat.size()); i++) {	
+	if (Brief_chat.empty()) {
+		return;
+	}
+
+	Assertion(Brief_start_display_index >= 0 && Brief_start_display_index < static_cast<int>(Brief_chat.size()),
+		"Chat display index is invalid, please report!");
+
+	auto chat = Brief_chat.begin();
+	std::advance(chat, Brief_start_display_index);
+
+	while (chat != Brief_chat.end() && count < Chatbox_max_lines) {	
 		if (count < Chatbox_max_lines) {
-			SCP_list<brief_chat>::iterator chat = Brief_chat.begin();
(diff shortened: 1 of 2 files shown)


JohnAFernandez:
Going to push back to 24.2, but I'll also review this soon.

code/missionui/chatbox.cpp
```diff
@@ -888,66 +812,73 @@ void chatbox_add_line(const char *msg, int pid, int add_id)
...
-						gr_bitmap(Chatbox_icon_x,ly-2,GR_RESIZE_MENU);
+	int count = 0;	
+	int ly = Chatbox_begin_y;
+	int line_height = gr_get_font_height() + 1;
+
+	for (int i = Brief_start_display_index; i < static_cast<int>(Brief_chat.size()); i++) {	
+		if (count < Chatbox_max_lines) {
+			SCP_list<brief_chat>::iterator chat = Brief_chat.begin();
+			std::advance(chat, i);
+
```
notimaginative, line 823: This could really be replaced with something like:
```
auto chat = Brief_chat.begin();
std::advance(chat, Brief_start_display_index);

while (chat != Brief_chat.end() && count < Chatbox_max_lines) {
```
and then `++chat` when you increment the `count` value. You might also want to verify that `Brief_start_display_index` is a sane value just before this and exit early if it's not (same if `Brief_chat` is empty).

That should be a little easier to read and would definitely be faster.

notimaginative (changes requested):

code/missionui/chatbox.cpp
MjnMixael: Sure. 

As I'm reading it, I'm thinking `Brief_start_display_index` should probably be in an Assertion. It should never be a bad value and if we get here and it is then there's a deeper underlying issue.

notimaginative: Looks good, but you can remove the `if (count < Chatbox_max_lines)` too since it's covered by the same statement in the `while()`.

MjnMixael: Right.. duh. Done.

notimaginative (approved):

PR #5853, opened 2023-12-01
add mac arm64 builds to CI workflow

notimaginative:


notimaginative:
Depends on https://github.com/scp-fs2open/fs2open.github.com/pull/5852 going through before it can actually work.

Goober5000 (approved):

PR #5855, opened 2023-12-03
change two global variables to be scoped

Goober5000:
These variables are only ever used in the scope of `os_get_legacy_user_dir()`, so they can be scoped rather than global.  This should fix a Coverity issue on the `DCF(low_mem` debug console function.


JohnAFernandez (approved):
This may end up closing dozens of coverity issues, as there are a lot of repeats.

Goober5000:
Let's find out.

PR #5862, opened 2023-12-05
allow nonexistence of arrival overlays

Goober5000:
Ever since retail, the code has assumed that a soundtrack would always have arrival music.  This removes that requirement.  If arrival music is missing, it will just be skipped.

Tested by @skdKitsune.


JohnAFernandez (approved):
LGTM

PR #5863, opened 2023-12-05
Make beams respect 'check hull' and 'can damage shooter' flags

Baezon:
Yes, this will change behavior for mods using these options. But I hopefully should not need to mention how incredibly unintuitive and annoying it would be to have flags that are very explicit on what they do, only to just magically not work in certain situations, and magically require you to know to enable a completely different flag in a completely different file, otherwise.

EDIT: To clarify, with the new 'check hull' handling this will only early interrupt a beam that started unobscured and became obscured during its firing. 'check hull' as it is will prevent a beam from starting to fire if it is obscured.

### DIFF
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -532,6 +532,8 @@ int beam_fire(beam_fire_info *fire_info)
     flagset<Object::Object_Flags> default_flags;
 	if (!wip->wi_flags[Weapon::Info_Flags::No_collide])
 		default_flags.set(Object::Object_Flags::Collides);
+	if (wip->wi_flags[Weapon::Info_Flags::Can_damage_shooter])
+		default_flags.set(Object::Object_Flags::Collides_with_parent);
 
 	// create the associated object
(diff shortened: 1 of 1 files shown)


Goober5000 (commented):
I'm a little wary of this.  It's a good change, but `"check hull"` is a pretty old flag and there are a lot of mods that use it.  Do we know how much of a balance impact this may have?

Baezon:
Not really, no. I suspect it's low, but I don't know.

I'd want to get lots opinions on this, from modders especially, since I really really don't want to make this flag.

JohnAFernandez:
I'm fine with the change, because even if old mods use this flag, I'm not anticipating it making a big difference.

Baezon:
To clarify, since this wasn't immediately obvious to me, with the new 'check hull' handling this will only early interrupt a beam that started unobscured and became obscured during its firing. 'check hull' as it is will prevent a beam from starting to fire if it is obscured.

JohnAFernandez:
I had that behavior in mind when I made my comment.

JohnAFernandez (approved):
It seems like most people think that this change makes sense.  I think overall, the impact of this will be very low, since the beam would hit the parent ship anyway, if not turned off, and it's not likely to start hitting the target again afterward, at least for long.  

IMHO, balance is not greatly affected by low probability events.

Goober5000 (approved):
Agreed

PR #5865, opened 2023-12-06
Scpui multi options

MjnMixael:
This PR wires up the relevant multi options and grants access to reading/writing the login info and IP Address table where applicable. Note that reading the password is disallowed by design.


wookieejedi:
Thanks for your continued work on this. This PR in particular is quite useful since it will allow mods using multi and SCPUI to have a unified UI for all settings (as opposed to previously having to use the old UI setting to set import multi values).

code/scripting/api/libs/options.cpp
```diff
@@ -59,5 +61,92 @@ ADE_FUNC(discardChanges, l_Options, nullptr, "Discard any changes made to the op
 	return ADE_RETURN_TRUE;
 }
 
+ADE_VIRTVAR(MultiLogin, l_Options, "string", "The multiplayer PXO login name", "string", "The login name")
```
JohnAFernandez, line 64: login instead of login name?

JohnAFernandez (approved):
I think we're good.  I spotted one thing in the documentation, but I'll leave that up to you.

PR #5866, opened 2023-12-06
add more autosaving in FRED

Goober5000:
FRED has an autosave feature but it wasn't invoked in too many places.  This adds autosaving whenever a major dialog is closed, unless it was closed by hitting Cancel.  NB: Autosave is not yet implemented in qtFRED.

Also, `addvariabledlg` incorrectly referred to `OnOk` instead of `OnOK` and `CMessageEditorDlg` incorrectly referred to `OnOK` instead of `OnOk`, so this fixes those and distinguishes `OnOk` by renaming it `OnButtonOk` to be consistent with the event editor and other dialogs.

### DIFF
--- a/fred2/addvariabledlg.cpp
+++ b/fred2/addvariabledlg.cpp
@@ -138,7 +138,7 @@ void CAddVariableDlg::OnOK() {
 		//		}
 		//
 		//		m_sexp_var_index = sexp_add_variable(temp_value, temp_name, type);
-		//		this get done for free CDialog::OnOk() UpdateData(TRUE);
+		//		this get done for free CDialog::OnOK() UpdateData(TRUE);
 		m_create = true;
 
 		CDialog::OnOK();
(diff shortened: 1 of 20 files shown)


fred2/addvariabledlg.cpp
```diff
@@ -138,7 +138,7 @@ void CAddVariableDlg::OnOK() {
 		//		}
 		//
 		//		m_sexp_var_index = sexp_add_variable(temp_value, temp_name, type);
-		//		this get done for free CDialog::OnOk() UpdateData(TRUE);
+		//		this get done for free CDialog::OnOK() UpdateData(TRUE);
```
wookieejedi, line 141: Just something I noticed, but this has OK capitalized, while the actual function use is `Ok`

Goober5000: That's the distinction I mentioned in the PR description.  The `CDialog` function is `OnOK`, with the O and K capitalized.  This is the default MFC function that is invoked in several situations, such as when the user presses Enter.  The FRED code occasionally used `OnOk` and `OnButtonOk` so I've renamed all of them to `OnButtonOk`.

wookieejedi: Oh right, that all makes sense now, thanks!

PR #5870, opened 2023-12-09
add a new check to the Voice Acting Manager

Goober5000:
This adds a new button to check if messages sent by `"<any wingman>"` have at least one ship with that persona.


wookieejedi (approved):
Handy check feature and looks good to me

PR #5880, opened 2023-12-11
Fix renderOverheadModel's reliance on wss_slots

MjnMixael:
The method to render the 3d overhead model for weapon select has been intertwined with `wss_slots` and that was always kind of hacky. That forced SCPUI to do further hacks with its loadout handling to get it to work correctly. Well this PR does away with all the hacks and passes an array to the render method instead of the `wss_slots` index. The Lua call has been updated to accept either the slot index or a table.

I'm *probably* the only one using this lua method with SCPUI right now so there could be an argument to making the lua method only accept the table and not the slot for backwards compatibility. I could go either way.

Further explanation: `wss_slots` was only used to check what weapon was loaded into each weapon slot and if its index matched the currently selected weapon then draw the lines to the bank. So this PR sends a small array with the weapon indexes in its place. This means that the lua renderOverheadModel can be used even if `wss_slots` is empty.


JohnAFernandez (approved):
LGTM

PR #5884, opened 2023-12-13
str_wrap_to_width followup

Goober5000:
This refactors the `str_wrap_to_width` function into one that returns string offsets and lengths rather than newly constructed strings.  The algorithm is unchanged (except for one minor bugfix), but the performance is improved because the function no longer copies and re-copies the strings.  This also takes advantage of the optional parameters in `gr_string` and `gr_get_string_size` that provide a string length.

This also fixes an edge case in string wrapping where a word that fits exactly on the border no longer rapidly flips between one line and the next.

As a secondary cleanup, by using `size_t` in the right places, it is no longer necessary to cast a whole bunch of `size_t` parameters to `int`.  This allows a bunch of casts to be removed at the cost of adding one cast of `int` to `size_t`.

Tested with standard string wrapping, newline wrapping, and long-strings-within-newlines wrapping.  Also tested with several different string wrapping sources.

Followup to #5607.

### DIFF
--- a/code/camera/camera.cpp
+++ b/code/camera/camera.cpp
@@ -754,7 +754,7 @@ subtitle::subtitle(int in_x_pos, int in_y_pos, const char* in_text, const char* 
 		//Get text size
 		for(int i = 0; i < num_text_lines; i++)
 		{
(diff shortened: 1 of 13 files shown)


JohnAFernandez (approved):
LGTM

PR #5886, opened 2023-12-14
make is-in-mission work for wings

Goober5000:
Since its creation, the `is-in-mission` sexp operator has only checked for ships.  But mission designers might assume this also works for wings, leading to a situation where the sexp always returns false because no ship can have the same name as a wing.  Fortunately wing functionality can be added to this sexp without affecting ship functionality.


JohnAFernandez (approved):
LGTM

PR #5887, opened 2023-12-14
check for potential issues in FRED missions

Goober5000:
Add a routine to the sexp syntax checker that also checks for potential issues.  These are not definitive errors, but situations that might trip up mission designers.  Currently there is one issue checked: the destruction of a wing that has multiple waves and arrives from a docking bay.


wookieejedi (approved):
Useful check and looks good to me

PR #5888, opened 2023-12-14
fix crash in QtFRED

Goober5000:
Not sure how this option became false, but it should be true in order for the checkable menu option to be set up properly.


JohnAFernandez (approved):
Thank you!

PR #5890, opened 2023-12-15
Make internal table file match updated parsing preference

MjnMixael:
The name of this table flag was changed for parsing, so we need to update the internal tables to match to avoid log warnings that mods can't fix


Baezon (approved):

JohnAFernandez (commented):
Thanks for fixing this

JohnAFernandez (approved):

PR #5891, opened 2023-12-15
Self Credits update

MjnMixael:
Small thing, but I've started using my real name for my interwebs contributions at HLP, including BtA & FSU, and elsewhere. I'd like to switch to using my real name for SCP as well.


JohnAFernandez (approved):
This seems fone to me.

PR #5892, opened 2023-12-19
Fix Coverity WorkFlow

JohnAFernandez:
This fixes the weekly coverity workflow by ny using container images and switching to Checkout V1.

Also removes Vulkan SDK install, since it seems to be taken care of by the new containers or is disabled.


Goober5000 (approved):

JohnAFernandez:
For further explanation on switching to version 1, version 4 does not seem to play well with our containers, so I matched the nightly script's use of v1.

PR #5893, opened 2023-12-20
Fix coverity time t issues

JohnAFernandez:
The underlying type for time_t is a long, but for some reason, all the uses convert to int.  This really isn't much of an issue, but it's nice to have more room.  The multiplayer server code specifically could use more room, since it tracks how long the server is on, and in the case of a standalone, that could be a long time.  There's still a couple places where a conversion to int is used, but in more essential parts of the code, like Pilot code.

Everything seemed to still work fine in my test.

I changed `reliable_connect_time = 0;` to `reliable_connect_time = -1;` since that is how it is initialized elsewhere.  I did remove one use which has not been used for a while and is less accurate than our current time tracking.


Goober5000 (approved):
LGTM.  The only concern I had was about `reliable_connect_time`, but you've addressed it.

PR #5895, opened 2023-12-21
Retail ingame options

MjnMixael:
Adds a new In-Game Options game state accessible from the Options game state if In-Game Options are enabled. This new state uses ImGui to build out controls for all options that have been added to OptionsManager allowing for in-game control of things that were previously only available via table, commandline, launcher flag, or SCPUI. Things brings parity of options to all games played in the FSO engine and should help simplify the user-configuration pipeline going forward.

This PR touches a lot of files because I had to localize OptionsManager's category string, but otherwise the core updates are in the Options\ data folder files.

Options Windows and Options Controls are created dynamically based on OptionsManager and thus any future configuration option or category that is created will automatically show up here as long as the option is one of the supported types (Binary, Multiselect, Float Range, Int Range).

Note that this changes the `Using_in_game_options` global to default true else this UI's configurations would be ignored.


code/options/Ingame_Options.cpp
```diff
@@ -0,0 +1,103 @@
...
+					}
+
+					getOptConfigurator()->binary_options.push_back(thisPair);
+				}
+			}
+		// If it's a range option then setup the necessary float or int variables for imgui
+		} else if (thisOpt->getType() == options::OptionType::Range) {
+
+			// OptionsManager stores the value as a serialized string, so we need to return it to a float
+			auto f_val = std::stof(val.serialized.c_str());
```
JohnAFernandez, line 74: I'm not up to date on the options builder pipeline, but I believe here it's loading info from raw loaded data? If the contents of the options builder have not been validated already, we should add a try/catch block since `stof` can throw an exception.

code/options/dialogs/ingame_options_ui.cpp
```diff
@@ -0,0 +1,175 @@
...
+			
+			// Selectors
+			if (thisOpt->getType() == options::OptionType::Selection) {
+				auto values = thisOpt->getValidValues();
+				
+				if ((values.size() == 2) && !(thisOpt->getFlags()[options::OptionFlags::ForceMultiValueSelection])) {
+
+					// On/Off options
+
+					int idx = getOptConfigurator()->get_binary_option_index(thisOpt->getTitle());
```
JohnAFernandez, line 39: Nitpick - Extra new line here

JohnAFernandez (changes requested):
Quick question about one of the function calls here, and a nitpick.

code/options/Ingame_Options.cpp
MjnMixael: That's a good idea. I think I did it correctly. Let me know.

JohnAFernandez: It seems fine.

JohnAFernandez (approved):

PR #5896, opened 2023-12-22
Coverity: Create a copy of the object

JohnAFernandez:
Instead of returning the unique pointer, which would make it a shared unique pointer (bad).  This method is only used to get the value stored in the config and the reference will be destroyed pretty quickly after being called.  Sure, it is more efficient to use the pointer to get the info, but I don't think it's worth it to rewrite the methods here, and we don't need the efficiency since this is used to update the UI.

And it's not feasible to convert to shared pointer, since there are so many references to unique pointers in this code.  It's just better to just make a quick copy here.

I have to turn this through a test, still.

### DIFF
--- a/code/options/OptionsManager.cpp
+++ b/code/options/OptionsManager.cpp
@@ -40,15 +40,14 @@ std::unique_ptr<json_t> OptionsManager::getValueFromConfig(const SCP_string& key
 {
 	auto override_iter = _config_overrides.find(key);
 	if (override_iter != _config_overrides.end()) {
-		// We return a reference to an existing object so we need to increment the reference count
-		json_incref(override_iter->second.get());
-		return std::unique_ptr<json_t>(override_iter->second.get());
+		json_t temp = *override_iter->second.get();
+		return std::unique_ptr<json_t>(&temp);
 	}
 
 	auto changed_iter = _changed_values.find(key);
 	if (changed_iter != _changed_values.end()) {
-		json_incref(changed_iter->second.get());
-		return std::unique_ptr<json_t>(changed_iter->second.get());
+		json_t temp = *changed_iter->second.get();
+		return std::unique_ptr<json_t>(&temp);
 	}
 
 	auto parts = parse_key(key);


JohnAFernandez:
I'm still doing research, trying to figure out if this is complete.  Because the `json_t` type seems to want to track the number of references to it, and I'm not sure why yet.  (The previous method of incrementing the reference number is not really correct, because it never gets decremented again.)

code/options/OptionsManager.cpp
```diff
@@ -40,15 +40,14 @@ std::unique_ptr<json_t> OptionsManager::getValueFromConfig(const SCP_string& key
 {
 	auto override_iter = _config_overrides.find(key);
 	if (override_iter != _config_overrides.end()) {
-		// We return a reference to an existing object so we need to increment the reference count
-		json_incref(override_iter->second.get());
-		return std::unique_ptr<json_t>(override_iter->second.get());
+		json_t temp = *override_iter->second.get();
```
asarium, line 43: This creates a copy of data that we do not have a full definition of at this point (for example, see the `json_object_t` type to see what the actual structure of this looks like when we are dealing with an object.

In short, this is not how you copy an instance of `json_t`.

code/options/OptionsManager.cpp
```diff
@@ -40,15 +40,14 @@ std::unique_ptr<json_t> OptionsManager::getValueFromConfig(const SCP_string& key
...
+		return std::unique_ptr<json_t>(&temp);
```
asarium, line 44: This returns a pointer to stack allocated storage but managed by a unique pointer. This will lead to a dangling reference as soon as this function returns.

asarium (changes requested):
For coverity issues, could you add the description of the issue you are trying to fix. The code looks initially correct to me so knowing what coverity is seeing as an issue here would help.

My guess is that coverity fails to "see" the default deleter that is defined here: https://github.com/scp-fs2open/fs2open.github.com/blob/master/code/libs/jansson.h#L18

This ensures that the jannson reference count is decremented when the "unique pointer" is no longer in use.

PR #5898, opened 2023-12-22
make error messages more accurate

Goober5000:
There are several SEXP argument types that themselves encompass multiple argument types: for example, ships and wings.  This could be misleading when there is an error message displayed.  So, this updates the error codes to explicitly cover all of the possibilities.


code/parse/sexp.cpp
```diff
@@ -33959,7 +33964,22 @@ const char *sexp_error_message(int num)
...
+			return "Invalid ship/wing/ship on team/waypoint name";
+
+		case SEXP_CHECK_INVALID_SHIP_WING_POINT:
+			return "Invalid ship/wing/waypoint name";
+
+		case SEXP_CHECK_INVALID_ORDER_RECIPIENT:
+			return "Invalid order recipient";
+
+		case SEXP_CHECK_INVALID_SHIP_WING_WHOLETEAM:
+			return "Invalid ship/wing/team name";
```
JohnAFernandez, line 33982: Looks like this is a good change, but these message are a little obtuse. (Yes, I know the format matches the old ones, but is that a good thing?)

Would it perhaps be clearer to say, at least for some of these messages something like, "Invalid name that does not match a ship, wing or team"?

Honestly without some clarification, I would not be able to tell what an "Invalid ship/wing/ship" is supposed to be.

JohnAFernandez (changes requested):

code/parse/sexp.cpp
Goober5000: I could change this to "Invalid ship, wing, or team name", with similar changes to the other messages that contain forward slashes.  Would that work?

JohnAFernandez: I think so

Goober5000: Ok, done

JohnAFernandez (approved):

PR #5903, opened 2023-12-28
Improve description of `type` and `instance`

Baezon:
Although they may not seem so offensive, I distinctly remember finding these descriptions phenomenally unhelpful and confusing when I was first starting because it didn't actually use FSO example types.


Goober5000 (approved):
Although I will miss the Descent references, this is a good change

PR #5907, opened 2023-12-31
Fix unbalanced stack when retrieving Lua main thread

### LINKED ISSUE #5904: Nested LUA coroutines cause crashes on MACs
I found that nesting LUA coroutines causes crashes on MAC. Overall single coroutines work just fine on Mac, but if I make a coroutine and then call a function that runs another coroutine the crash occurs (again only on MAC). The error is thrown from `LuaFunction.cpp:155` and states `Error: Reference does not refer to a function!` (an image of this error and using the provided simple repo case is attached below).

image

Me and taylor and chief were trying to figure out how to fix it, but alas we came up short, and taylor suggested asking someone with more experience with FSO LUA internals. 

I've attached a simple repro case, which is just a single small script that will trigger when any mission starts. I've also attached an image of the code for convenience. 
[corotest-sct.zip](https://github.com/scp-fs2open/fs2open.github.com/files/13795540/corotest-sct.zip)

image

Wondering if either @BMagnu or @asarium might have any insight on how to fix this, especially since it is preventing MAC users from using certain scripts and playing FotG.

asarium:
After getting the main state from the Lua registry, this function had two Lua thread values on its Lua stack but we only removed one which lead to a value being left on the stack.

Why did this not blow up before? Because undefined behavior.

The only place where this was called was when creating a Lua reference value where luaL_ref is called in the same expression (see UniqueLuaReference::create). The evaluation order of the parameters for the make_shared call is not defined by the standard and for some reason all major compilers (except clang) choose to execute the luaL_ref call before the getMainThread call.

For those compilers, everything is fine. We have some extra values on the Lua stack but that gets cleaned up after the API function finishes.

However, clang first evaluates the getMainThread which leaves an extra Lua thread on the Lua stack. The luaL_ref call then references this thread value instead of the intended value. In the concrete case, this meant that LuaFunction saw that the reference points to a thread and rightfully threw an exception.

This fixes #5904.

### DIFF
--- a/code/scripting/lua/LuaUtil.cpp
+++ b/code/scripting/lua/LuaUtil.cpp
@@ -55,7 +55,9 @@ lua_State* getMainThread(lua_State* L)
 	lua_rawget(L, LUA_REGISTRYINDEX);
 
 	auto state = lua_tothread(L, -1);
-	lua_pop(L, 1);
+
+	// Need to pop both the thread from L and the main thread to return the stack back to neutral
+	lua_pop(L, 2);
 
 	return state;
 }


wookieejedi (approved):
Very much appreciate your effort and time to dig into this one, finally removes that previously incredibly difficult crash. Very handy the fix is just one line change, too.

PR #5910, opened 2024-01-02
campaign error message improvement

Goober5000:
This improves campaign load error handling in several ways.  The code now makes a better distinction between load failures that occurred because the file was missing, vs ones that occurred because of some actual error such as a parsing failure or savefile incompatibility.  The error when the player clicks on the campaign room has been updated to reflect this.

Also, the fallback behavior has been changed.  If the player's current campaign cannot be found, the code now automatically falls back to the mod's default campaign or the first campaign in the campaign list.  To avoid problems, this only happens when the game is started for the first time or when the player chooses a new pilot; in all other cases there is no fallback from the current campaign, whether or not it can be found.  (This fallback behavior was inadvertently already implemented for mods which 'ignored' the player's current campaign.)

The main hall code also correctly checks for a campaign load error rather than assuming the current campaign has valid data.

Finally, some exceptions have been changed to Assertions and the `Campaign_file_missing` variable has been changed to a boolean.

### DIFF
--- a/code/mission/missioncampaign.h
+++ b/code/mission/missioncampaign.h
@@ -161,8 +161,8 @@ extern SCP_vector<SCP_string> Ignored_campaigns;
 
 extern char Default_campaign_file_name[MAX_FILENAME_LEN - 4];
 
-// if the campaign file is missing this will get set for us to check against
-extern int Campaign_file_missing;
+extern bool Campaign_file_missing;	// if the campaign file is missing this will get set for us to check against
+extern int Campaign_load_failure;
 
 /*
  * initialise Player_loadout with default values
--- a/code/mission/missioncampaign.h
+++ b/code/mission/missioncampaign.h
@@ -30,8 +30,6 @@ struct sexp_variable;
 #define CAMPAIGN_ERROR_SAVEFILE			-4
 #define CAMPAIGN_ERROR_IGNORED			-5
 
-inline bool campaign_reportable_error(int val) { return val != 0 && val != CAMPAIGN_ERROR_MISSING && val != CAMPAIGN_ERROR_IGNORED; }
-
 // types of campaigns -- these defines match the string literals listed below which
 // are found in the campaign files.  I don't think that we need campaigns for furball
 // missions.
--- a/code/mission/missioncampaign.h
+++ b/code/mission/missioncampaign.h
@@ -30,6 +30,8 @@ struct sexp_variable;
 #define CAMPAIGN_ERROR_SAVEFILE			-4
 #define CAMPAIGN_ERROR_IGNORED			-5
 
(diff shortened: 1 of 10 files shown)


code/mission/missioncampaign.h
```diff
@@ -30,6 +30,8 @@ struct sexp_variable;
 #define CAMPAIGN_ERROR_SAVEFILE			-4
 #define CAMPAIGN_ERROR_IGNORED			-5
 
+inline bool campaign_reportable_error(int val) { return val != 0 && val != CAMPAIGN_ERROR_MISSING && val != CAMPAIGN_ERROR_IGNORED; }
```
notimaginative, line 33: Unless I've missed it, this doesn't appear to be used outside of `missioncampaign.cpp`. If that's the case it would be best to just move it there so that it's not global.

notimaginative (changes requested):

code/parse/parselo.cpp
```diff
@@ -2104,10 +2104,9 @@ int parse_get_line(char *lineout, int max_line_len, const char *start, int max_s
 // Goober5000 - added ability to read somewhere other than Parse_text
 void read_file_text(const char *filename, int mode, char *processed_text, char *raw_text)
 {
-	// copy the filename
-	if (!filename)
-		throw parse::ParseException("Invalid filename");
```
Baezon, line 2109: Is there a reason we shouldn't still throw a parse exception if no filename was given?

Baezon (changes requested):

code/mission/missioncampaign.h
Goober5000: True, it's not used outside the .cpp file, but since the #defines are in the header, I thought this function should be too.  It represents an essential property of the different error types.

code/parse/parselo.cpp
Goober5000: There shouldn't ever be a case where these functions are passed a NULL filename.  Empty filename perhaps, but a NULL filename is a coding error.

Baezon: I agree, but that doesn't change the situation, I'm not advocating the removal of the assertion. If the filename is null, release builds *definitely* should not attempt to do anything with it and instead throw a parse exception to whatever the outer context is.

Goober5000: I'm skeptical that the exception would ever be thrown, but ok.

Baezon: Me too, but I've grown wary of assertions of a thing which should always be true... but then continue anyway if it fails, guaranteeing a crash when a safer early out could have been done instead.

Baezon (approved):

code/mission/missioncampaign.h
JohnAFernandez: And when you're running on a craptop like me, compilation time matters a lot.

Goober5000: Ok, done.

notimaginative (approved):

PR #5914, opened 2024-01-03
Print option settings to the log

### LINKED ISSUE #5900: FR: INI/settings reporting in debug log.
The debug log's utility in diagnosing problems can sometimes come from seeing the player's command line flags. As we move more and more things into ingame settings/launcher built .ini files, it might make sense to have the game dump those settings to the log on startup so we don't lose that capability.

MjnMixael:
Adds code to print player options to the log at game start and when the options game state is closed. Fixes #5900.

Has the additional benefit of not only printing most of the ini file data, but also any retail options as well. Once #5895 is merged there is a small bit of additional code here that can be uncommented to only print in-game options if they are active while also still always printing built-in retail options.

This also fixes an incorrect xstr setting for the mouse input option title in OptionsManager.


JohnAFernandez (approved):
Really good change, tbh

PR #5916, opened 2024-01-03
upgrade the OPF_POSITIVE number hack

Goober5000:
This is a new method of ensuring that OPF_POSITIVE numbers are positive.  It works for all numeric arguments, including variables and literals, and caches the status of the node to avoid lookups for every negative number on every frame.  Tested on both operator and non-operator cases.

Follow-up to #5915.  ~~In draft until that is merged.~~


wookieejedi (approved):
Looks good and works in my tests

PR #5917, opened 2024-01-03
Ingame options respect mods

### LINKED ISSUE #5697: In-game Options Discussion
Chatting with @MjnMixael about the paradigm for in-game options and wanted to post this here for additional discussion.

Regarding where things are saved for settings, there are some variations:

Game settings that were always in retail (sound level, nebula detail, etc) are saved in the player file, which is great since different mods and pilots can keep settings. Unfortunately, the settings like resolution and all the graphic in-game options that relate to command lines are not saved in the player file but rather the singular common `fs2_open.ini`. 

With in-game options you can open one mod, toggle a graphic option like `soft_particles` on, then whatever mod you run after that will have `soft_particles` enabled (unless that mod has `-disable soft_particles` in its command line). 

Though given that FSO needs to read those graphic options before starting that's why they are not in the player file. Still, it definitely made it a bit puzzling at first why some settings stuck across various mods and others did not.

From Mjn
`I think resolving this will require a kind of fundamental shift to how FreespaceOpen is treated. For decades we've considered "FreespaceOpen" the game and the mods as the game content. But this requires that we consider mods as their own unique games and thus we probably need to look at modifying how data is stored in the .ini again or provide some other new file to check for in-game options as overrides to everything else if they exist.`

Posting this as an issue here to promote additional discussion in the community.

MjnMixael:
This PR fixes #5697 by changing how in-game settings are saved and loaded. The key thing Wookie and I wanted here was to have these settings saved in a way that's specific to each mod but agnostic of mod versions (as far as Knossos/Nebula goes). The challenge there is that different mod versions are entirely unique mods to FSO even though Knossos knows better.

The solution this PR offers is a new ini file saved to `%appdata/HardLightProductions\FreeSpaceOpen\data`. This file is titled based on the `Cmdline_mod` parameter (`retail_fs2_settings.ini` if it's nullptr) by trimming a mod folder's version data off of it. For example, `MPVS-4.7.2` becomes `MVPS_settings.ini` and will work for all version of the MediaVPs.

There's almost certainly room for improvement here but this needs to be solved sooner rather than later and a PR will get the discussion really going.

### DIFF
--- a/code/localization/localize.cpp
+++ b/code/localization/localize.cpp
@@ -64,7 +64,7 @@ bool *Lcl_unexpected_tstring_check = nullptr;
 // NOTE: with map storage of XSTR strings, the indexes no longer need to be contiguous,
 // but internal strings should still increment XSTR_SIZE to avoid collisions.
 // retail XSTR_SIZE = 1570
-// #define XSTR_SIZE	1854 // This is the next available ID
+// #define XSTR_SIZE	1855 // This is the next available ID
 
 
 // struct to allow for strings.tbl-determined x offset
--- a/code/menuui/playermenu.cpp
+++ b/code/menuui/playermenu.cpp
@@ -408,7 +408,7 @@ void player_select_do()
 	}
 
 	if (!Ingame_options_save_found && Using_in_game_options) {
-		popup(PF_BODY_BIG | PF_USE_AFFIRMATIVE_ICON, 1, POPUP_OK, XSTR("In-game Options are enabled but a save file could not be found. You may need to update your settings in the Options menu.", 1816));
+		popup(PF_BODY_BIG | PF_USE_AFFIRMATIVE_ICON, 1, POPUP_OK, XSTR("In-game Options are enabled but a save file could not be found. You may need to update your settings in the Options menu.", 1842));
 	}
 		
 	// set the input box at the "virtual" line 0 to be active so the player can enter a callsign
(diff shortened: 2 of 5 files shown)


JohnAFernandez (approved):
AFAICT, this is really well done.  It is really hard to visualize string manipulation (at least for me, so I think I would like to know what kind of test cases you guys tried before I approve, in case I (or someone else) can think of some others that we can try.

JohnAFernandez:
And then I hit the approve button.

JohnAFernandez (changes requested):
As requested, just whatever types of test cases you guys tried, just in case.

JohnAFernandez:
I managed to get this result just by playing around with some semi-nonsense semver strings:

Screenshot 2024-02-10 172250

What do you think, as intended?

MjnMixael:
> Sorry for the slow reply, I'm always missing updates on github.
> 
> I managed to get this result just by playing around with some semi-nonsense semver strings:
> 
> Screenshot 2024-02-10 172250
> 
> What do you think, as intended?
> 
> FWIW, I think your test cases will handle the vast majority of cases, and something like this will probably never be chosen.

I tend to agree and even in the case of the -semver-string.semver it fellback gracefully to fs2_open.ini which is good. That's what I wanted; anything weird happens? Use fs2_open.ini.

JohnAFernandez (approved):

MjnMixael:
Please do not merge this until after #5895 is merged. There's many options-related PRs upcoming and they all touch localization...

MjnMixael:
This needs an xstr update before merge.

MjnMixael:
XSTR has been handled. This is good to merge.

PR #5920, opened 2024-01-05
Properly check if we're running in a windowed mode

MjnMixael:
There were a few cases that didn't account for the newer in-game option when checking for Windowed mode. This PR resolves that which fixes the issue where the in-game option wasn't accounted for at start up or when creating dialog prompts.

### DIFF
--- a/code/graphics/2d.cpp
+++ b/code/graphics/2d.cpp
@@ -1754,17 +1754,6 @@ bool gr_init(std::unique_ptr<os::GraphicsOperations>&& graphicsOps, int d_mode, 
 	return true;
 }
 
-void gr_force_windowed()
-{
-	if ( !Gr_inited ) {
-		return;
-	}
-
-	if ( Os_debugger_running ) {
-		os_sleep(1000);
-	}
-}
-
 int gr_activated = 0;
 void gr_activate(int active)
 {
--- a/code/graphics/2d.cpp
+++ b/code/graphics/2d.cpp
@@ -1779,7 +1779,7 @@ void gr_activate(int active)
 	}
 
 	if (active) {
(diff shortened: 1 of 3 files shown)


JohnAFernandez (approved):

TRBlount:
After being able to fully examine this PR and #5895, I have some issues I'd like to bring up. Currently, the way your system appears to work means that the game options can only be set from command-line or only from in-game. This could cause many players some grief as they'll have their command-line already configured and then suddenly they aren't working when they try to set options in-game instead. 

I think this is an issue that needs to be addressed.

MjnMixael:
> After being able to fully examine this PR and #5895, I have some issues I'd like to bring up. Currently, the way your system appears to work means that the game options can only be set from command-line or only from in-game. This could cause many players some grief as they'll have their command-line already configured and then suddenly they aren't working when they try to set options in-game instead.
> 
> I think this is an issue that needs to be addressed.

If you have ideas then I'm all for it. In-game options, as designed, is meant to be a replacement for all the commandline stuff. That's why there's a mod.tbl toggle to turn it on or off. If you'd like to redo all of that, then feel free. That's outside the scope of this PR, though. This PR address an inconsistency that's even more confusing.. that in this specific case Windowed Mode does not fall into either category, but a awkward, opaque hybrid.

Baezon:
This sounds like more of a #5895 thing, but fwiw I am totally in favor of defaulting to using in game options once all users can be guaranteed to have access to it (through the new imgui screen). The consternation of some players with commandline settings who didn't know they needed another cmdline option to override in game options is absolutely worth the benefit.

JohnAFernandez:
I stand by my original approval as well

JohnAFernandez:
Based on further discussion and consensus on discord, this is being merged.

PR #5924, opened 2024-01-06
Cockpit fov option

MjnMixael:
Adds Cockpit FOV as an in-game option along with a global to toggle using it, per discussion with Lafiel. Also sets a define for the default FOV of 0.75 that is used in multiple places. There are lot of options PRs pending with XSTR changes so this will just be set as draft until those are merged and I can update this one's XSTRs, but otherwise the code is complete.

### DIFF
--- a/code/camera/camera.cpp
+++ b/code/camera/camera.cpp
@@ -14,8 +14,9 @@
 #include "ship/ship.h"        //compute_slew_matrix
 
 //*************************IMPORTANT GLOBALS*************************
-fov_t VIEWER_ZOOM_DEFAULT = 0.75f;			//	Default viewer zoom, 0.625 as per multi-lateral agreement on 3/24/97
-fov_t COCKPIT_ZOOM_DEFAULT = 0.75f;
+#define DEFAULT_FOV 0.75f;
+fov_t VIEWER_ZOOM_DEFAULT = DEFAULT_FOV;
+fov_t COCKPIT_ZOOM_DEFAULT = DEFAULT_FOV;
 float Sexp_fov = 0.0f;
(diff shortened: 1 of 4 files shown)


code/camera/camera.cpp
```diff
@@ -71,11 +74,48 @@ auto FovOption = options::OptionBuilder<float>("Graphics.FOV",
 					      return true;
 					 })
 					 .display(fov_display)
-					 .default_val(0.75f)
+					 .default_val(fov_default)
 					 .level(options::ExpertLevel::Advanced)
```
JohnAFernandez, line 78: Why not use `DEFAULT_FOV` here?

code/camera/camera.cpp
```diff
@@ -54,6 +54,9 @@ APPLY_TO_FOV_T(*, mul)
 APPLY_TO_FOV_T(+, add)
 APPLY_TO_FOV_T(-, sub)
 
+// Used to set the default value for in-game options
+static float fov_default = DEFAULT_FOV;
```
JohnAFernandez, line 58: If we keep this one, it should be also be `const` or `constexpr`.

JohnAFernandez (changes requested):

code/camera/camera.cpp
MjnMixael: Options Builder wouldn't accept the define as a parameter causing builds to fail. I'm not sure why exactly.

code/camera/camera.cpp
MjnMixael: Done

JohnAFernandez (approved):

PR #5926, opened 2024-01-07
LUA method to check if a movie is currently playing

MjnMixael:
A quick LUA method to return true/false if a movie is currently playing. This allows me to fix a bug in SCPUI where cutscenes are not visible in some cases, even though they are actually being played.

While this borders on a new feature, this is something that SCPUI & BtA need and it's exceedingly low impact. If possible, I'm hoping that this can make it into 24.0 as BtA intends to release on that build.

### DIFF
--- a/code/cutscene/cutscenes.cpp
+++ b/code/cutscene/cutscenes.cpp
@@ -25,6 +25,8 @@
 
 extern int Cmdline_nomovies;
 
+bool Movie_active = false;
+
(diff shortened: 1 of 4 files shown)


JohnAFernandez (approved):
Looks good, and I support this getting added before stable.

JohnAFernandez:
There were no objections. Merging.

PR #5928, opened 2024-01-08
Allow lua to lock/unlock cutscenes

MjnMixael:
Allows scripts to arbitrarily lock/unlock cutscenes which will allow for mods to lock/unlock cutscenes at times outside of playing campaign missions if they desire.


JohnAFernandez (approved):
LGTM

PR #5929, opened 2024-01-08
Cleanup ui cutscene api

MjnMixael:
Cleanup & formalize cutscene Lua API. Fixes spelling errors and strange method names. Also fixes Lua index off-by-one error that prevented getting cutscene at position 0 because Lua starts at 1. Allows getting a cutscene by name or index now and adds isValid() to match other Lua Indexers. Also prevents crash if accessing an out of bounds cutscene index.

Note that the spelling error, while printed to the scripting.html, is not actually used by scripts, so these changes should be backwards compatible.

### DIFF
--- a/code/cutscene/cutscenes.cpp
+++ b/code/cutscene/cutscenes.cpp
@@ -60,6 +60,18 @@ static cutscene_info *get_cutscene_pointer(char *cutscene_filename)
 	return NULL;
 }
 
+int get_cutscene_index_by_name(const char* name)
+{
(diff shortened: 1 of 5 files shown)


JohnAFernandez (changes requested):
Just fix your casts and you should be good.

JohnAFernandez (approved):

JohnAFernandez:
Based on discussion on discord, this contains a true fix and should be merged so that it can be included in stable.

PR #5930, opened 2024-01-08
Fix voice volume typo

MjnMixael:


wookieejedi (approved):

JohnAFernandez (changes requested):
Hey, this involves pilot file code, and we need someone who knows pilot file to take a look.  A test for forwards compatibility would work too. (Can previous versions of fs2 load a pilot file saved with this code amd vice versa, I suspect not)

MjnMixael:
> Hey, this involves pilot file code, and we need someone who knows pilot file to take a look. A test for forwards compatibility would work too. (Can previous versions of fs2 load a pilot file saved with this code amd vice versa, I suspect not)

Well that teaches me to make assumptions that FSO handles a json file like a normal application...

So it looks like the only way to really do this is by bumping the PLR_VERSION to 5, which will trigger a popup to players that their pilot file will be converted to version 5, and including whatever is needed to do conversion to the new file version.

That's a whole lotta no thanks from me over a typo I just happened to notice. I'd rather just leave the typo. So I'm gonna just close this one. Someone else can come up with a real fix in the future if they want to.

JohnAFernandez:
I think that's why no one else has fixed this typo previously.

PR #5931, opened 2024-01-09
Clarify which SEXPS work on ships not arrived

Dark-Visor:
SEXP helps should let the FREDder know whether a given ship argument needs to be already in-mission for the SEXPression to operate. Update the helps to include (ship needs to be in-mission) or (ship does not need to be in-mission) or other relevant helpful information.


Goober5000 (dismissed):
Nice work getting this all done. 👍

PR #5935, opened 2024-01-11
connect ship and parsed ship in Lua API

Goober5000:
Add `getShip()` to parsed ships and `getParsedShip()` to ships to allow the two APIs to connect to one another.


wookieejedi (approved):
Very handy, and looks good to me

PR #5936, opened 2024-01-11
Expose parse skip_to_string

MjnMixael:


code/scripting/api/libs/parse.cpp
```diff
@@ -81,6 +81,29 @@ ADE_FUNC(displayMessage, l_Parsing, "string message, boolean error = false",
...
+
+	const char* str;
+	if (!ade_get_args(L, "s", &str)) {
+		return ADE_RETURN_FALSE;
+	}
+
+	try {
+		skip_to_string(str);
+	} catch (const parse::ParseException& e) {
+		mprintf(("PARSE: Error while parsing: %s\n", e.what()));
```
JohnAFernandez, line 100: It might be good to mark this a parse error while lua is attempting it, but otherwise, approved.

JohnAFernandez (approved):

PR #5937, opened 2024-01-12
some parse upgrades

Goober5000:
Turn `Parse_names` into a vector and add SCP_string overloads of ship registry functions.  Refactor calling code to use registry as appropriate.  Remove an unused overload of `mission_parse_get_parse_object`.


wookieejedi (approved):
Looks good to me and things work as expected, though wouldn't hurt to have another set of eyes on this, too.

PR #5943, opened 2024-01-16
fix time-to-goal edge cases

Goober5000:
This fixes a few items that were revealed in the course of Dark Visor's SEXP documentation PR.  The EXITED status should return NAN_FOREVER, not just NAN.  The CANT_EVAL value is never returned from the sexp on account of ship status, due to the preceding check, and is not an appropriate return value anyway; so that is removed.  Finally, for a ship that can't move, the NAN return value is better than CANT_EVAL, since it represents the numeric situation more accurately and can also be checked by `is-nan`.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -19672,12 +19672,9 @@ int sexp_time_to_goal(int n)
 {
 	auto ship_entry = eval_ship(n);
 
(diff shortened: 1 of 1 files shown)


JohnAFernandez (approved):
I think this makes sense, but just because of the off-chance that a mod is relying on the way it currently is, I'm going to mark it as waiting for stable.

Issue #5941, opened 2024-01-15
FR: Override the $Randomize particle rotation: flag on a per-effect basis


MoerasGrizzly:
The $Randomize particle rotation flag is awesome, but there's some effects (like Diaspora's exp_nuke and kestrellius's flare effects) that expect particles to not be rotated. Please add a flag to particle_effects where we can set this value for individual tabled effects, overriding whatever default is set by the game_settings flag.

Baezon:
He's almost certainly referring to the [these particle effects](https://wiki.hard-light.net/index.php/Particle_Effects), the ones handled in the `particle` folder.

Goober5000:
Ok, I think I see what this is doing.  The `BeamPiercingEffect` and `ParticleEmitterEffect` types do not have parsing implemented, but these do not appear on the wiki page.  All the ones that do, go through the `ParticleProperties::parse` function.

Baezon:
Yes, those are special cases. I don't think we really need to worry about having this toggle on *every possible particle* of which there are a great many, I think having it be on the main particle effects defined in the typical way using a -part.tbm would be sufficient for now, since that is much simpler to modify and is the great majority of the particle effects people care about.

Goober5000:
Implemented by #5944

PR #5944, opened 2024-01-17
allow finer control of particle rotation randomization

Goober5000:
This adds a new field to particle effects so that particle rotation can be configured on a per-type basis, not just on a blanket basis.  Implements #5941.  Tested by @MoerasGrizzly.


code/particle/util/ParticleProperties.cpp
```diff
@@ -62,6 +62,21 @@ void ParticleProperties::parse(bool nocreate) {
...
+
+	if (optional_string("+Rotation:")) {
+		SCP_string buf;
+		stuff_string(buf, F_NAME);
+		if (!stricmp(buf.c_str(), "DEFAULT")) {
+			m_rotation_type = RotationType::DEFAULT;
+		} else if (!stricmp(buf.c_str(), "RANDOM")) {
+			m_rotation_type = RotationType::RANDOM;
+		} else if (!stricmp(buf.c_str(), "FIXED")) {
+			m_rotation_type = RotationType::FIXED;
```
Baezon, line 74: I'm wondering if this might be called "screen aligned" or something instead, since "fixed" might be a bit vague. One might expect that to mean fixed with regard to some global reference frame instead of the viewer's point of view!

Baezon (changes requested):
I was thinking of a boolean to invert whatever `Randomize_particle_rotation` is but I like this 3 state enum a lot better.

code/particle/util/ParticleProperties.cpp
Goober5000: Agreed, and I'm certainly open to other terminology.  @MoerasGrizzly does `SCREEN_ALIGNED` work?

Kestrellius:
This looks really promising!
@EatThePath should probably be looped in on this; she was working on the same issue previously.

code/particle/util/ParticleProperties.cpp
MoerasGrizzly: screen_aligned definitely works.

Goober5000: Ok, modified

code/particle/util/ParticleProperties.cpp
```diff
@@ -62,6 +62,21 @@ void ParticleProperties::parse(bool nocreate) {
...
 	}
+
+	if (optional_string("+Rotation:")) {
+		SCP_string buf;
+		stuff_string(buf, F_NAME);
+		if (!stricmp(buf.c_str(), "DEFAULT")) {
+			m_rotation_type = RotationType::DEFAULT;
+		} else if (!stricmp(buf.c_str(), "RANDOM")) {
+			m_rotation_type = RotationType::RANDOM;
+		} else if (!stricmp(buf.c_str(), "SCREEN_ALIGNED")) {
```
Baezon, line 73: A little nitpicky, but i prefer omitting underscores from user facing strings where possible. "SCREEN ALIGNED"

Goober5000: Okay.  How about all three, underscore, dash, and space?

Baezon: Works for me.

Baezon (approved):

EatThePath:
> This looks really promising! @EatThePath should probably be looped in on this; she was working on the same issue previously.

My ambition got me hung up on two things:
1. wanting to implement something like omnibeam rotation pattern, or perhaps like soundsets, that would let stylizished narrow rotation options
2. Wanting to flip images too, as asymmetric particles that are always seen 'facing' one direction could look unnatural in their own way.

Neither really needs to be, or in the first case really should be, present in this PR, IMO.

PR #5946, opened 2024-01-18
fix splash crash and add supplemental script hook

Goober5000:
This PR fixes a crash in the initialization sequence and makes a few defense-in-depth improvements to address related risks.

PR #5379 introduced a bug that everyone, including me, missed the first time around:
> It's worth noting this moves `game_title_screen_close()` above the script hooks for On Game Init and On Splash Screen. This is so that anything that runs on Game Init that might draw to the screen won't have the splash screen fade-out rendered directly afterwards.

Now, script initialization occurs through several functions called in sequence.  During this sequence, the script system is not completely initialized, and therefore `os_poll()` cannot be called lest it invoke other scripting hooks.  Unfortunately, `game_title_screen_close()` calls `os_poll()`, so it needs to be moved back to its original location.  To address the needs of mods that want to draw to the screen without having the splash screen fade-out rendered on top of it, this PR also provides a new On Splash End hook, which was the alternate solution originally proposed in #5379.

This also moves `Viewer_mode = 0` and `mod_table_post_process()` prior to script initialization so that scripts can depend on completely configured data, and moves sexp.html creation out of the middle of script initialization and over to a more appropriate location in `game_main` where other similar command-line functions are handled.

### DIFF
--- a/code/scripting/global_hooks.cpp
+++ b/code/scripting/global_hooks.cpp
@@ -12,6 +12,10 @@ const std::shared_ptr<Hook<>> OnGameInit = Hook<>::Factory("On Game Init",
 	tl::nullopt,
 	CHA_GAMEINIT);
 
+const std::shared_ptr<Hook<>> OnSplashEnd = Hook<>::Factory("On Splash End",
+	"Executed just after the splash screen fades out.",
+	{});
+
 const std::shared_ptr<Hook<>> OnIntroAboutToPlay = Hook<>::Factory("On Intro About To Play",
 	"Executed just before the intro movie is played.",
 	{});
--- a/code/scripting/global_hooks.h
+++ b/code/scripting/global_hooks.h
@@ -6,6 +6,7 @@ namespace scripting {
 namespace hooks {
 
 extern const std::shared_ptr<Hook<>>									OnGameInit;
+extern const std::shared_ptr<Hook<>>									OnSplashEnd;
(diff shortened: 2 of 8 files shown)


BMagnu (commented):
Mhm, this seems to be the sensible way of doing it, as I'm personally not in favor of somehow reusing the OnSplashScreen for this.
I do wonder whether or not we should eventually put the entire ``runInit`` and ``OnGameInit`` stuff into its own function, and until that is run, block all other hooks to avoid such issues in the future, but I guess that can wait.

Goober5000:
Uh, possibly.  The first thing that comes to mind is some sort of `Scripting_inited` global variable to match existing ones like `Ships_inited`.  This would be set to true immediately before `OnGameInit` is run, and the `run()` function would immediately return if this variable was not yet true.

This sort of thing could be added to this PR or deferred to a future PR.

BMagnu:
Yeah.
I think that may be a good idea.
Probably could / should be added to this one, removing the need for a post-splash hook, as the init hook could be left where it is.
Especially since we already have the quite-close OnIntroAboutToPlay hook as well

Goober5000:
I think the post-splash hook should be kept.  Even though we have OnIntroAboutToPlay, I assume the motivation for #5379 was to have some post-splash effect that played as early in the game sequence as possible.  Normally OnGameInit would be ideal for this, but the splash fade-out needs to happen after that hook has run.

BMagnu (approved):
Okay, that seems fine now

PR #5947, opened 2024-01-19
Fix: Keep Chase view from crashing in multi

### LINKED ISSUE #5875: Game crashes when external view button is pressed in observer mode
Still present on 23.2.1

JohnAFernandez:
This keeps observers from accidentally crashing the game by using the proper get eyepoint function when using chase view.  Observers will simply use their natural eyepoint instead of trying to be chase their non-existent ship.  Chase view will still be available on targeted ships.

Fixes #5875 

Tested and works as expected.

Btw, this is one of those "hide the whitespace" PR's.


Goober5000 (approved):
Short, sweet, and simple. 👍

PR #5948, opened 2024-01-19
Fix MSAA for some graphics drivers

BMagnu:
Some OpenGL drivers, apparently mostly Apple's, don't just ignore redeclarations of built in variables (as Linux & Windows drivers for AMD and Nvidia seem to), but actively error out.
So, remove the redeclaration, and MSAA will work again on macs and possibly other drivers that might have had this behaviour.


wookieejedi (approved):
Tests on Nvidia and Mac show fix works.

PR #5949, opened 2024-01-20
Allow ba.xstr to lookup from strings.tbl

### LINKED ISSUE #5945: FR: Need ba.xstr but for internal strings
SCPUI often just uses the internal string ID for identical error messages and dialog prompts but has no way to get the translated string from strings.tbl rather than tstrings.tbl. A quick lua method to expose that should be pretty easy.

MjnMixael:
Fixes #5945 by allowing ba.XSTR to optionally check internal strings instead of tstrings.tbl. Defaults to tstrings.tbl. Will be useful for SCPUI.


JohnAFernandez (approved):
LGTM

PR #5954, opened 2024-01-24
Fix discord option needing restart

MjnMixael:
Discord option never needs a restart, so this should always return true.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -162,16 +162,14 @@ static auto DiscordOption __UNUSED = options::OptionBuilder<bool>("Game.Discord"
                                if (!val) {
                                     Discord_presence = false;
                                     libs::discord::shutdown();
-                                    return true;
                                }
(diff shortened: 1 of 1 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -162,16 +162,14 @@ static auto DiscordOption __UNUSED = options::OptionBuilder<bool>("Game.Discord"
...
                           } else {
                                if (val) {
                                     Discord_presence = true;
                                     libs::discord::init();
-                                    return true;
                                }
                           }
-                          return false;
+                          return true;
                      })
```
JohnAFernandez, line 173: If it should always return true, should it return anything at all?

MjnMixael: Yes. .change_listener requires a book's return to know if an option successfully persisted or not. That tells it whether or not the option requires a restart.

JohnAFernandez (approved):

PR #5955, opened 2024-01-25
Pause the UI properly based on user settings

MjnMixael:
Similar to windowed mode, this was another case of an in-game option that needed a few more checks added to make sure we respect the correct user setting.

I removed the commandline check in osapi opting to keep all the pause globals in freespace.cpp that osapi doesn't have access to. So rather than more includes we can just return early from the pause/unpause functions.


JohnAFernandez:
Still waiting for any other possible feedback, but so far this fix looks like it'll be merged for 24.0

JohnAFernandez (approved):

PR #5960, opened 2024-01-28
Increase JOY_NUM_BUTTONS to 128 and add required localizations

derrickturk:
I want to put this out as less of a "pull request" and more of a strawman or question from a drive-by contributor - is this worth doing? Are there non-obvious downsides (performance? backward compatibility with old pilot profiles [maybe the movement of the hat switch identifiers has bad implications]? etc.).

Bumping the joystick button limit to match what I understand to be the maximum allowed by DirectInput (plus the now excellent multiple-controller support) would be a big help for those of us with "fancy" HOTAS setups. Many mid- and high-end controllers support acting as multiple "virtual controllers", but for at least some this doesn't work with FS2Open due to each "virtual controller" having the same identifiers.

### DIFF
--- a/code/controlconfig/controlsconfigcommon.cpp
+++ b/code/controlconfig/controlsconfigcommon.cpp
@@ -442,7 +442,23 @@ const char* Joy_button_text_german_u[] = {
 	"Knopf 13",		"Knopf 14",		"Knopf 15",		"Knopf 16",		"Knopf 17",		"Knopf 18",
 	"Knopf 19",		"Knopf 20",		"Knopf 21",		"Knopf 22",		"Knopf 23",		"Knopf 24",
 	"Knopf 25",		"Knopf 26",		"Knopf 27",		"Knopf 28",		"Knopf 29",		"Knopf 30",
-	"Knopf 31",		"Knopf 32",		"Hut Hinten",	"Hut Vorne",	"Hut Links",	"Hut Rechts"
+	"Knopf 31",	"Knopf 32",	"Knopf 33",	"Knopf 34",	"Knopf 35",	"Knopf 36",
+	"Knopf 37",	"Knopf 38",	"Knopf 39",	"Knopf 40",	"Knopf 41",	"Knopf 42",
+	"Knopf 43",	"Knopf 44",	"Knopf 45",	"Knopf 46",	"Knopf 47",	"Knopf 48",
+	"Knopf 49",	"Knopf 50",	"Knopf 51",	"Knopf 52",	"Knopf 53",	"Knopf 54",
+	"Knopf 55",	"Knopf 56",	"Knopf 57",	"Knopf 58",	"Knopf 59",	"Knopf 60",
+	"Knopf 61",	"Knopf 62",	"Knopf 63",	"Knopf 64",	"Knopf 65",	"Knopf 66",
+	"Knopf 67",	"Knopf 68",	"Knopf 69",	"Knopf 70",	"Knopf 71",	"Knopf 72",
+	"Knopf 73",	"Knopf 74",	"Knopf 75",	"Knopf 76",	"Knopf 77",	"Knopf 78",
+	"Knopf 79",	"Knopf 80",	"Knopf 81",	"Knopf 82",	"Knopf 83",	"Knopf 84",
+	"Knopf 85",	"Knopf 86",	"Knopf 87",	"Knopf 88",	"Knopf 89",	"Knopf 90",
+	"Knopf 91",	"Knopf 92",	"Knopf 93",	"Knopf 94",	"Knopf 95",	"Knopf 96",
+	"Knopf 97",	"Knopf 98",	"Knopf 99",	"Knopf 100",	"Knopf 101",	"Knopf 102",
+	"Knopf 103",	"Knopf 104",	"Knopf 105",	"Knopf 106",	"Knopf 107",	"Knopf 108",
+	"Knopf 109",	"Knopf 110",	"Knopf 111",	"Knopf 112",	"Knopf 113",	"Knopf 114",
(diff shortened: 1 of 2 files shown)


Baezon:
Have you tested whether this is sufficient to actually use buttons past 32? I assume you have such a joystick?

derrickturk:
Yes, it's worked fine with my setup (joystick + a separate throttle with ~100-odd buttons) on "my machine" (Windows 10, build from above patch on MSVC 2022, MediaVPs, Knossos.NET) in limited playtesting. I don't have the capability at the moment to test much beyond that.

JohnAFernandez:
So it sounds like it fixes issues for joysticks with near the max number of buttons, so I think we just need someone to test the converse: that joysticks with fewer buttons (and mice and keyboards) still work with these changes.

z64555 (approved):
I had wanted to use the button name translation procedural instead of manually defining a name for each one as done here, but I suppose this is still acceptable for now.

I checked SDL's wiki to see if they had a recommended button limit but I didn't come across such a thing.  It holds its numbers as a uint16_t and there's the possibility that somewhere in FSO they use the button array as a char or uint8_t.  In either case the value 128 is well within the limits of the data types so I don't forsee any conflicts.

For completeness sake, we should pause merging until somebody with a more baseline joystick or controller is able to confirm that the changes don't break anything, as @JohnAFernandez suggested.

derrickturk:
@z64555 what did you have in mind for a procedural translation? I like the idea but didn't feel comfortable either going from static arrays to a lookup function or introducing some kind of metaprogramming (templates or macros) to generate the static arrays. That second option is the way I would lean but I am a total rando to this codebase!

JohnAFernandez:
Hey, I'm just converting this to draft until we figure out if we want to make the string generation dynamic or not, so that no one merges it by accident.  I'm good with it being static.

derrickturk:
I summoned something from the dreaming void. It appears to be valid C++20: https://gist.github.com/derrickturk/51c68fb15cbf82211dc4c812832487b3

I think there may be an even shakier, standards-wise, way to "count" using the preprocessor, but it's beyond me. At most I'd probably use an "X-macro" to simplify the generation of arrays entries, like so: https://gist.github.com/derrickturk/2e26b18de7f075b0b069bae63ec24df8. At least that keeps the "counting" simple and in one place, although it introduces the requirement to make sure that the count constants get updated concurrently with the definition list. I think keeping it as is would be just fine too and possibly the least confusing option for most programmers.

derrickturk:
I guess I should elaborate that the above techniques are both "static" in the sense of operating at or before compile-time and generating honest-to-god constant arrays. There would be a whole other set of approaches for generating these at runtime - probably simpler to program but at the expense of a tiny bit of performance and complexity (when do we initialize them, etc etc).

JohnAFernandez:
I don't think we'd be able to upgrade to c++ 20 just for this.  We do have some c++ 17 features added via small libraries, but we're targeted to c++11 for now.

derrickturk:
IMO it's not worth it for the metaprogramming here - I've blown up several "C++20" compliant compilers with similar code (the core idea comes from a named-tuple implementation) so it's a bit "bleeding edge".

My personal comfort zone would be using an X-macro to keep the arrays global, static, and constant, but I don't see that pattern elsewhere in the codebase and I'd be reluctant to throw it in here as a one-off unless it's well-known to other developers.

Probably the best answers, again IMO, are to keep things as is, or perhaps introduce a singleton like `io::mouse::CursorManager` which generates the tables dynamically once and gets initialized "early" with a static `init()` function.

wookieejedi:
We had someone recently ask about increasing the number of buttons, so just wanted to check in on this PR. What's the remaining blockers to move it out of draft and into review readiness? :)

derrickturk:
IIRC @z64555 built a proper implementation of procedurally-generated button names in a previous commit. I read some of the discussion around that and I believe the blocker on increasing `JOY_NUM_BUTTONS` was an interaction with the control presets code. I have not had a chance to investigate that further yet, unfortunately.

wookieejedi:
Chatting with z in Discord, at this point all that is needed for the control limit bump is "essentially its a one-line change thats blocked by needing to test if it would break anything"

wookieejedi:
@derrickturk I've gone ahead and made just the bump PR separately in #6479; out of curiosity, would you be able to test that PR?

derrickturk:
> @derrickturk I've gone ahead and made just the bump PR separately in #6479; out of curiosity, would you be able to test that PR?

I was able to build your branch and try it out last night. I didn't test exhaustively but I was able to bind buttons in the 50s and 60s without a problem and use them in-flight. I was also able to save and load control presets. So this change seems OK as far as I can tell!

The only issue I noticed is not related to your changes, but showed up when binding a hat switch:
image

I bound the shield quadrant controls to the corresponding hat directions (e.g. forward should be hat up...), but the labels in the binding UI show up incorrectly (my "up" shows as "right", etc.). I've verified with Windows joy.cpl that the directions reported by my stick's hat correspond to the physical hat directions.

wookieejedi:
Thanks for testing! Hmm and that hat direction bug only occurs on this branch and not a current nightly, correct?

derrickturk:
> Thanks for testing! Hmm and that hat direction bug only occurs on this branch and not a current nightly, correct?

The bug is present in the nightly builds as of nightly 20241222. I suspect it's nothing to do with your button limit change but rather something to do with the changes to procedural generation of button labels. If I had to guess, I'd closely check the decision around here to re-order the label arrays: https://github.com/scp-fs2open/fs2open.github.com/pull/6371/commits/0f14b61dfb96124a13e2a538d91398ad4d39b2d7.

wookieejedi:
Good to know! Thanks we will be sure to dig into that!

wookieejedi:
@derrickturk I've fixed/merged the hat labeling bug and also merged the button limit bump. Thanks again for all your work and testing. I will also now close this PR :)

PR #5962, opened 2024-01-28
fix warp time durations

Goober5000:
Properly set the warp durations when ships arrive and depart, so that the warp animation can open and close smoothly without any discontinuities.

~~In draft pending additional tests.~~  Tested by myself and Gamma39er.

### DIFF
--- a/code/ship/shipfx.cpp
+++ b/code/ship/shipfx.cpp
@@ -612,7 +612,7 @@ static void shipfx_actually_warpout(ship *shipp, object *objp)
 	}
 }
 
-void WE_Default::compute_warpout_stuff(float *warp_time, vec3d *warp_pos)
+void WE_Default::compute_warpout_stuff(float *ship_move_time, vec3d *warp_pos)
 {
 	Assertion(isValid(), "Warp effect must be valid when compute_warpout_stuff() is called!");
 
@@ -678,11 +678,8 @@ void WE_Default::compute_warpout_stuff(float *warp_time, vec3d *warp_pos)
 		ship_move_dist = warp_dist - half_length;
 	}
 
-	// Calculate how long to fly through the effect.  Not to get to the effect, just through it.
-	*warp_time = warping_time;
-
-	// Acount for time to get to warp effect, before we actually go through it.
-	*warp_time += ship_move_dist / warping_speed;
+	// Calculate time to get to warp effect, before we actually go through it.
+	*ship_move_time = ship_move_dist / warping_speed;
 
 	if (m_portal_objnum < 0)
 	{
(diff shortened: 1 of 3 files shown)


Gamma39er:
Reviewing the code showed that the rest of the code assumed a warp open and close duration of 2.0 seconds. The durations were essentially hardcoded to 2.35 seconds despite the rest of the calculations for the timing of the warp. In fact if the total warp time is under 4 seconds, `fireball_create` clamps the value to 4.0 seconds.

The discontinuity occurs because `total_time` would be less than the sum of `warp_open_duration` and `warp_close_duration`. In `fireball_wormhole_intensity` the warp must complete it's opening animation before moving on to the close animation. By the time the open animation is done, we are already closer to the end of the total_time, placing the animation in the middle of the closing animation, causing the discontinuity. Ships at around 21 meters, like the Apollo, barely saw any discontinuity. However, all 16 meter ships saw a bigger discontinuity because the `total_time` of the warp is dependent on the length of the ship.

Part of testing was to determine which was more important to maintain for visual correctness, the 2.35 second open and close durations or the total time calculation. After observing a `total_time` calculation that included a 2.35 second open and close duration and comparing that with a 2 second open and close duration, we determined that the difference between 2.35 seconds and 2.0 seconds was minimal and that the `total_time` calculation using 2.0 second durations was more important. The longer timing with 2.35 seconds made jumping out at the end of missions feel strange as it took slightly too long.

code/ship/shipfx.cpp
```diff
@@ -3491,14 +3488,15 @@ int WE_Default::warpStart()
...
-		// for ship to go thru, and 'SHIPFX_WARP_DELAY' (1.5 secs) to go away.
-		effect_time = warping_time + SHIPFX_WARP_DELAY + SHIPFX_WARP_DELAY;
+		// for arrivals, opening and closing times are both SHIPFX_WARP_DELAY
+		open_time = close_time = SHIPFX_WARP_DELAY;
 	}
 	else
 	{
-		compute_warpout_stuff(&effect_time, &pos);
-		effect_time += SHIPFX_WARP_DELAY;
+		// for departures, the opening time is the time it takes for the ship to actually approach the warp effect (not de jure, but de facto)
```
JohnAFernandez, line 3496: "(not de jure, but de facto)" does not make sense.  You're saying "not rightfully so, but as it is because it is what it is".  But what is the rightful way to do it in the first place?

Goober5000: What I meant was "this is probably not what Volition originally intended, but is how it ended up working".  In other words, I assume they intended for the opening and closing times to be `SHIPFX_WARP_DELAY` for all four possibilities (arrival open, arrival close, departure open, departure close).  But perhaps due to unintended side effects from implementation, the departure open time will always vary from ship to ship, depending on how long it takes that ship to move up to its departure point.

Goober5000: I re-worded the comment.  See what you think.

JohnAFernandez: That is much clearer, thanks

JohnAFernandez (approved):

PR #5964, opened 2024-01-29
Add opf custom string as a lua sexp return type

MjnMixael:
Exactly that. Adds Custom String as a return type that lua sexps can require. The return is a table value containing the three elements of the custom string exactly as if it was accessed by mn.CustomStrings.


JohnAFernandez (approved):
TGIS

PR #5965, opened 2024-01-29
switch to new linux build image

notimaginative:
Upgrades to Ubuntu 20.04 to fix Node v20 problem.
Bumps compiler to gcc-9.

NOTE: This breaks support in official builds for Linux distros older than Ubuntu 20.04


JohnAFernandez:
I'm going to go ahead and merge because we need to start running new CI checks on the other PRs.

PR #5967, opened 2024-01-31
bitstring poofs

Goober5000:
Change `Neb2_poof_flags` to use a bit string of unlimited length to allow for an arbitrary number of nebula poofs.

~~In draft pending testing.~~ Tested by Ice Cream Man.

### DIFF
--- a/code/lab/renderer/lab_renderer.cpp
+++ b/code/lab/renderer/lab_renderer.cpp
@@ -1,14 +1,15 @@
-#include "lab/renderer/lab_renderer.h"
+#include "freespace.h"
 #include "globalincs/vmallocator.h"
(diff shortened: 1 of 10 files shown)


JohnAFernandez (changes requested):
Just a couple things.  `get_int` and `set_int` are horribly generic names.  Since they are supposed to be used with bit arrays, why not something like `bit_array_get_int` and `bit_array_set_int`?

Also, some basic comments at the head of the new functions would be good.

Goober5000:
Good points; done.

JohnAFernandez (approved):

JohnAFernandez:
Looks like a `set_int` call escaped.

PR #5971, opened 2024-02-03
Fix "missing" command brief data

JohnAFernandez:
This fixes command briefing data apparently disappearing.  The ui was not being properly initialized because model_changed was being called before the ui was initialized, leading to default values.

This also fixes a few other minor issues.

Fix is tested and works as expected.

### DIFF
--- a/qtfred/src/mission/dialogs/CommandBriefingDialogModel.cpp
+++ b/qtfred/src/mission/dialogs/CommandBriefingDialogModel.cpp
@@ -28,16 +28,16 @@ void CommandBriefingDialogModel::initializeData()
 		strcpy_s(_wipCommandBrief.background[1], "<default>");
 	}
 
-	if (_wipCommandBrief.num_stages <= 0) {
-		_currentStage = 0;
-	} else {
-		_currentStage = 1;
-	}
+	_currentStage = 0;
 
 	int i;
 
 	for (i = 0; i < _wipCommandBrief.num_stages; i++) {
 		_wipCommandBrief.stage[i] = Cur_cmd_brief->stage[i];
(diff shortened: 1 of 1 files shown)


qtfred/src/mission/dialogs/CommandBriefingDialogModel.cpp
```diff
@@ -28,16 +28,16 @@ void CommandBriefingDialogModel::initializeData()
...
+	_currentStage = 0;
 
 	int i;
 
 	for (i = 0; i < _wipCommandBrief.num_stages; i++) {
 		_wipCommandBrief.stage[i] = Cur_cmd_brief->stage[i];
+		strcpy_s(_wipCommandBrief.stage[i].ani_filename, Cur_cmd_brief->stage[i].ani_filename);
+		_wipCommandBrief.stage[i].text = Cur_cmd_brief->stage[i].text;
+		_wipCommandBrief.stage[i].wave = Cur_cmd_brief->stage[i].wave;
+		strcpy_s(_wipCommandBrief.stage[i].ani_filename, Cur_cmd_brief->stage[i].wave_filename);
```
Goober5000, line 40: Looks like a copy-paste error... should the first argument be `wave_filename`?

JohnAFernandez, line 40:
```suggestion
		strcpy_s(_wipCommandBrief.stage[i].wave_filename, Cur_cmd_brief->stage[i].wave_filename);
```

JohnAFernandez: Thanks for catching that

Issue #5968, opened 2024-01-31
messages set to priority 'high' are not sent by command when the original sender is destroyed.


MoerasGrizzly:
The send-message sexp mentions that:
> High priority are not interrupted by anything, and will be sent by Command if the sender is destroyed.

As of FSO 23.2.1 and 24.0, this is no longer the case. I do remember that this worked as described in earlier versions.

MoerasGrizzly:
I do also vaguely recall that the <any wingman> setting would flat out not send a message if a wingman with the matching persona could not be found, even when the message priority is set to 'high'. I'd prefer it if that particular option was a little bit more robust too.

Goober5000:
The messaging system is definitely supposed to fall back to Command if a High priority message needs to be sent with no wingman available.  This is used in FS1 and in the Port.  I don't recall offhand if the behavior is supposed to be different for `<any wingman>` vs a specific wingman.

Goober5000:
Well I'm getting inconsistent results testing send-message on retail FS1 and retail FS2.  More investigation is needed.

Goober5000:
> As of FSO 23.2.1 and 24.0, this is no longer the case. I do remember that this worked as described in earlier versions.

@MoerasGrizzly When was the last version in which this happened as described?

Goober5000:
Ok it turns out that `send-message` still works correctly.  The idiosyncratic detail here is that the High messages are indeed sent by Command, but *the personas are still checked*.  If the High priority message has a persona, this persona is converted to Command if and only if a) it's an in-mission message, and b) the file has a `1_`, `2_`, etc. prefix that can be converted to an existing file with a prefix `TC_`.  On the other hand, if the High priority message has *no* persona assigned, the message will be sent regardless.

Incidentally, there is no difference in handling between `<any wingman>` and a specific wingman.

See also #4150, #4078, #4357, and their linked issues.

MoerasGrizzly:
> > As of FSO 23.2.1 and 24.0, this is no longer the case. I do remember that this worked as described in earlier versions.
> 
> @MoerasGrizzly When was the last version in which this happened as described?

22.0 or 22.2

Goober5000:
Well here is the mission I used for testing.  You will need to also copy TC_arrival.wav into the `voice/special` folder.
[0send-message-test.fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/14152589/0send-message-test.fs2.txt)

[...]

PR #5972, opened 2024-02-03
update sexp help for send-message

Goober5000:
Clarify message priorities, especially the conditions that allow High messages to be sent by Command.  Motivated by #5968.


JohnAFernandez (approved):
Long since due

PR #5974, opened 2024-02-04
use C++ casts for floating.h and floating.cpp

Goober5000:
Change the casts used in the convenience #defines in floating.h.  Since these #defines are used pretty widely, this should reduce the count of C-style casts in the codebase by quite a bit.

Also change the casts in floating.cpp while we're at it, and remove some obsolete or unused code.

### DIFF
--- a/code/math/floating.cpp
+++ b/code/math/floating.cpp
@@ -22,13 +22,13 @@
  */
 float fl_roundoff(float x, int multiple)
 {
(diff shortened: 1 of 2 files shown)


JohnAFernandez:
I don't think this is linked to any coverity issues.  But it does help with stability.

Goober5000:
Doesn't coverity flag the use of C-style casts rather than C++-style casts?

PR #5975, opened 2024-02-05
fix error handling in cfile_push_chdir()

notimaginative:
Old Volition bug where the original directory is pushed onto a stack before the chdir() call. If that call fails then the stack is left with a used entry that won't be removed until restart and can cause the stack to fill up. Now it's added only if the chdir() call succeeds.

Also take the opportunity to make Cfile_stack dynamic.

### DIFF
--- a/code/cfile/cfile.cpp
+++ b/code/cfile/cfile.cpp
@@ -91,12 +91,10 @@ cf_pathtype Pathtypes[CF_MAX_PATH_TYPES]  = {
 };
 // clang-format on
 
-#define CFILE_STACK_MAX	8
 
 int cfile_inited = 0;
-static int Cfile_stack_pos = 0;
 
-static char Cfile_stack[CFILE_STACK_MAX][CFILE_ROOT_DIRECTORY_LEN];
+static std::vector<std::string> Cfile_stack;
 
 std::array<CFILE, MAX_CFILE_BLOCKS> Cfile_block_list;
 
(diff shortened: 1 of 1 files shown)


code/cfile/cfile.cpp
```diff
@@ -363,13 +360,16 @@ int cfile_chdir(const char *dir)
...
-
-	if ( !Cfile_stack_pos )
+	if (Cfile_stack.empty())
 		return -1;
 
-	Cfile_stack_pos--;
-	return cfile_chdir(Cfile_stack[Cfile_stack_pos]);
+	int rc = cfile_chdir(Cfile_stack.back().c_str());
+
+	Assertion(rc == 0, "Failed to chdir() to previous directory (%s)!", Cfile_stack.back().c_str());
```
JohnAFernandez, line 368: Might be helpful to know the value of rc here, too.

JohnAFernandez (approved):

code/cfile/cfile.cpp
notimaginative: I thought about that, but it's actually far less useful than the directory name itself. And 99% of the time this is just going back to the working directory. So if we can't do that then we're pretty much fubar.

notimaginative: I should add that the rc value is not the result of the OS chdir() command itself, which is why it isn't particular useful beyond whether it worked or not.

PR #5977, opened 2024-02-05
Fix faulty / missing optimization for AMD shader compilation

BMagnu:
Turns out that #5055 had an edge case in which performance for AMD cards with shadows enabled dropped off a cliff.

This has several reasons and fixes, some of them to be addressed in a followup PR after this.
But most of all, this is due to the AMD shader compiler, apparently, not correctly optimizing away the huge swath of unneeded lighting code after the return in the ``#ifdef FLAG_SHADOW_MAP``.
As the code is always dead in this configuration, it's no difficulty to just ifdef it out in that case, approximately tripling FPS on AMD cards with shadows enabled. With this, performance is still significantly lower than before #5055, but no longer downright unplayable.

### DIFF
--- a/code/def_files/data/effects/main-f.sdr
+++ b/code/def_files/data/effects/main-f.sdr
@@ -118,10 +118,13 @@ uniform sampler2DArray sMiscmap;
 uniform sampler2DArray shadow_map;
 
 out vec4 fragOut0;
+
+#ifndef FLAG_SHADOW_MAP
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
Tests work on my end on Nvidia, and Shivan has tested on AMD, so approving

The-E (approved):

PR #5978, opened 2024-02-05
Add runtime flag to split up the main render shader

BMagnu:
The change in #5055 caused performance regressions for AMD cards.

One of the significant bottlenecks is addressed by #5977.
Other bottlenecks include: The large number of named uniforms set, the large amounts of untaken branches, etc.

To address this issue without either undoing #5055 or introducing double shaders, this PR replaces the branches for flag checks with preprocessor-defined switches. Depending on the mode, these switches can either be normal runtime if's (which _should_ provide pretty much exactly the same shader code and thus compiled binary as is in master right now), or compiletime ifdefs that will split the shader. This improves performance, but due to the more shaders required and dynamically compiled, reintroduces the stutters on ship jump in.

To do all this without double shaders, it adds some shader preprocessor directives to FSO:
- ``#conditional_include`` which allows to dynamically add shaders depending on whether or not FSO reports any given graphics capability as available.
- ``#predefine`` which acts as a directive to define a very simple replacement macro that is guaranteed to be executed before the GLSL preprocessor, and thus allows the generation of GLSL preprocessor directives.
- ``#prereplace`` which acts as the directive to inject code defined with ``#predefine``

In addition to just change how branches are implemented, it's also used to reduce the amount of data passed between stages if possible, as well as reduce the number of uniform samplers existing and updated if possible.

While not as big as an improvement as #5977, it's still somewhat of a performance improvement. And as it does not require double shaders, causes changes in the shader if the flag is disabled, and adds capabilities that may be interesting for future work, I still believe it is worth to be included in the long run.

Interestingly, this flag also fixes the incompatibilities of our main shader with old / obscure GPU's, as it successfully allowed running FSO on the RK3588+Panfork without needing manually patched drivers, as reported by @Shivansps.


JohnAFernandez:
As this is targeted for 24.2, and we are trying to limit changes to 24.0, I'm adding the waiting for stable label.

wookieejedi:
Will plan to merge this later today

PR #5979, opened 2024-02-05
Unhardcode renderSelectModel zoom value

MjnMixael:
I was fighting a UI bug where the weapons were rendering way closer to the camera than they should have been in SCPUI. Once I started digging further I found these hardcoded zoom multipliers. I'm sure I copied these from the retail UI code but that no longer makes sense here and it's poor design.

So this exposes the zoom multiplier with a default value of what it is currently. Additionally this fixes a small error in the scripting.html documentation.

While this one walks the line between feature and fix, and BtA can release without it, we'd love to be able to have this fix in 24.0.0 since it's pretty well contained to the scripting API.

### DIFF
--- a/code/scripting/api/objs/shipclass.cpp
+++ b/code/scripting/api/objs/shipclass.cpp
@@ -1267,10 +1267,10 @@ ADE_FUNC(renderTechModel2, l_Shipclass, "number X1, number Y1, number X2, number
 
 ADE_FUNC(renderSelectModel,
 	l_Shipclass,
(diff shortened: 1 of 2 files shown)


JohnAFernandez (approved):

JohnAFernandez:
Well, I'll bring it up in internal.  My description of this situation is "a feature enhancement that fixes undesired behavior."

JohnAFernandez:
No one spoke against it, and Goober is also for it.

PR #5983, opened 2024-02-08
fix FRED's "new object type" popup

Goober5000:
1. Fix the item ID indexing on the popup submenus and make sure they correspond with the ship dropdown menu
2. Remove the "player start" popup menu option as it has very limited utility and was already removed in qtFRED
3. Reduce a bunch of casts by using more appropriate popup menu types
4. Actually place an object in the mission when one is chosen from the popup menu

Also:
1. Remove the `Default_player_model` variable as its use is redundant
2. Remove unused arguments of `create_player()`


wookieejedi (approved):
Works as expected and looks good to me, of course anyone else who wants to put another set of eyes on this is welcome, too

JohnAFernandez (approved):
Also LGTM

PR #5987, opened 2024-02-12
make several miscellaneous timestamps type-safe

Goober5000:
Trail timestamps and a few top-level timestamps.  Also a few `static` and `constexpr` keywords for good measure.


JohnAFernandez (approved):
Good limited changes.

PR #5992, opened 2024-02-16
Change special IFF color `#define`s into `extern int`s.

MageKing17:
Before this commit, `IFF_COLOR_SELECTION`, `IFF_COLOR_MESSAGE`, and `IFF_COLOR_TAGGED` were hardcoded indices into the `Iff_colors` vector. This could cause problems if any of these three colors were set to the same RGB values (since `iff_init_color()` doesn't create a new entry if the color already exists in the vector), and if a modular table tried to change any of these colors nothing would happen (since it would just create a new entry in `Iff_colors` without doing anything to the hardcoded indices).

This commit instead changes the `#define`s into `extern int`s, and assigns the return value of `iff_init_color()` to those global variables upon parsing the special color values, meaning that any combination of the three can be set to the same color without causing issues, and modular tables can actually modify them.

Potential clutter from unused values in the `Iff_colors` vector is basically unchanged (technically slightly worse if a modular table overrides any of these special colors with a color that doesn't already exist in the vector), but should maybe be looked at in the future.

### DIFF
--- a/code/iff_defs/iff_defs.cpp
+++ b/code/iff_defs/iff_defs.cpp
@@ -24,6 +24,10 @@ int radar_iff_color[5][2][4];
 int iff_bright_delta;
 int *iff_color_brightness = &iff_bright_delta;
 
+int IFF_COLOR_SELECTION = 0;
+int IFF_COLOR_MESSAGE   = 1;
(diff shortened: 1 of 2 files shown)


Goober5000 (approved):
Good change.

PR #5995, opened 2024-02-20
cover all stages of ship warping in the Lua API

Goober5000:
This improves the ability of Lua to check for ship arrival status:

1. Deprecates the `isWarpingIn` function, which can be misleading, and replaces it with the equivalent `isWarpingStage1`
2. Adds an `isWarpingStage2` function to fill a gap in functionality
3. Adds an `isArrivingWarp` function (complementing the existing `isDepartingWarp` function) to cover both stage 1 and stage 2


wookieejedi (approved):
Very useful to have these options and looks good to me

PR #5997, opened 2024-02-21
allow targeting subsystems on already targeted ship under reticule

ferralverrall:
This is a streamlining of functionality, that makes it so one doesn't need to hit a second key to select the subsystem under the reticule.  Instead once you have the capital ship selected, hit Target Under Reticule and it will target a subsystem under the reticule if the feature is enabled and there isn't a new target under the reticule.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -419,7 +419,7 @@ void parse_mod_table(const char *filename)
 			stuff_boolean(&Dont_automatically_select_turret_when_targeting_ship);
 		}
 
-		if (optional_string("$Auto select subsystem under reticle when targeting ship:")) {
+		if (optional_string("$Auto select subsystem under reticle when targeting same ship:")) {
 			stuff_boolean(&Automatically_select_subsystem_under_reticle_when_targeting_same_ship);
 		}
 
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -48,6 +48,7 @@ bool Enable_external_default_scripts;
 int Default_detail_level;
 bool Full_color_head_anis;
 bool Dont_automatically_select_turret_when_targeting_ship;
+bool Automatically_select_subsystem_underreticle_when_targeting_ship;
 bool Always_reset_selected_wep_on_loadout_open;
 bool Weapons_inherit_parent_collision_group;
 bool Flight_controls_follow_eyepoint_orientation;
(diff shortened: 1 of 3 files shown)


code/mod_table/mod_table.cpp
```diff
@@ -48,6 +48,7 @@ bool Enable_external_default_scripts;
 int Default_detail_level;
 bool Full_color_head_anis;
 bool Dont_automatically_select_turret_when_targeting_ship;
+bool Automatically_select_subsystem_underreticle_when_targeting_ship;
```
Baezon, line 51: Very pedantic, but let's put an underscore between `under` and `reticle` for the consistency.

Baezon (changes requested):
This value is initialized to false, and nothing can set it to true! You probably want to parse a flag in game_settings.tbl, via `parse_mod_table()`.

code/mod_table/mod_table.cpp
```diff
@@ -1443,6 +1444,7 @@ void mod_table_reset()
 	Default_detail_level = 3; // "very high" seems a reasonable default in 2012 -zookeeper
 	Full_color_head_anis = false;
 	Dont_automatically_select_turret_when_targeting_ship = false;
+	Automatically_select_subsystem_underreticle_when_targeting_ship = false;
```
BMagnu, line 1447: Well, you introduce this bool to toggle your feature, but to actually make use of it, you need to actually parse a value from the table and populate the bool.
It should probably be somewhere towards the end of the #Other settings group of settings.

MjnMixael: #5145 should provide a simple roadmap to working with mod_table.cpp to add the necessary boolean table flag.

ferralverrall: I was thinking HUD settings?  Similar to the Dont_automatically_select_turret_when_targeting_ship, which has some code in the same area.

BMagnu (approved):

code/mod_table/mod_table.cpp
```diff
@@ -418,6 +419,10 @@ void parse_mod_table(const char *filename)
 			stuff_boolean(&Dont_automatically_select_turret_when_targeting_ship);
 		}
 
+		if (optional_string("$Auto select subsystem under reticle when targeting ship:")) {
+			stuff_boolean(&Automatically_select_subsystem_under_reticle_when_targeting_ship);
+		}
```
Goober5000, line 424: Pedantically speaking, if the subsystem is only selected the second time the ship is targeted, this isn't quite accurate.  You could say "targeting same ship" (and `targeting_same_ship`) as a possible alternative.

ferralverrall: yeah, I have been struggling to name it something that makes sense and not super long.  I've definitely failed at both.

Goober5000: Don't be discouraged; it's a tricky problem to begin with.  These sorts of custom situations are highly specific and often idiosyncratic; witness the "$Don't automatically select a turret when targeting a ship:" right next to it.

And your existing name was 95% of the way there; this is just a suggestion to bring it all the way home.

ferralverrall: good catch

Goober5000 (approved):

Baezon (approved):

PR #6000, opened 2024-02-22
Scanning is chaos, let's make it sane

### LINKED ISSUE #5922: FR: subsystem scanning that doesn't say "cargo"
Currently, when a subsystem has a specific cargo assigned to it, on scan it's hardcoded to print "cargo: CARGONAME".

Request an option to not print "cargo".

MjnMixael:
Are you tired of remembering whether you need Scannable to get cargo names to appear? Or maybe you're sick of Toggle Subsystem Scanning having different effects based on the ship type? Did you once want to have a mission where a transport wasn't scannable at all? Then have I got the PR for you!

This PR introduces a game_settings flag to unify scanning behavior. When enabled, all ships, no matter the ship class, will react the same way to all the scanning ship flags.

1) They are not scannable by default
2) Settings Scannable makes them.. uh.. scannable and cargo will be revealed by default
3) A new flag available with alter-ship-flag, called no-scanned-cargo, changes the behavior to scanned/unscanned instead of revealing the cargo
4) Toggle Subsystem Scanning must be enabled to swap scanning mode into subsystem scanning mode and works for all ship types

ADDITIONALLY: This PR fixes #5922 by allowing a mission file to define $Cargo Title: (or +Cargo Title: for subsystems). When this is set instead of Cargo: ITEM or Passengers: ITEM, the string will concatenate TITLE: ITEM. Note that I didn't not add FRED UI support because I'm within a couple months of diving into qtFRED in earnest, so I thought I'd save the effort. FRED will correctly read and save the Title values, though.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -1587,16 +1587,12 @@ int hud_target_ship_can_be_scanned(ship *shipp)
 	if (shipp->flags[Ship::Ship_Flags::Cargo_revealed])
 		return 0;
 
-	// allow ships with scannable flag set
-	if (shipp->flags[Ship::Ship_Flags::Scannable])
+	// The old behavior treats some ship types as always scannable. The new behavior only checks the ship's Scannable flag.
+	if (shipp->flags[Ship::Ship_Flags::Scannable]) {
 		return 1;
-
-	if (!Use_new_scanning_behavior) {
-		// ignore ships that don't carry cargo
-		if ((sip->class_type < 0) || !(Ship_types[sip->class_type].flags[Ship::Type_Info_Flags::Scannable]))
-			return 0;
-	} else {
-		// In new behavior if we are here then the ship does not have the scannable flag so it's not scannable!
+	} else if (Use_new_scanning_behavior) {
+		return 0;
+	} else if ((sip->class_type < 0) || !(Ship_types[sip->class_type].flags[Ship::Type_Info_Flags::Scannable])) {
 		return 0;
 	}
 
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -1591,9 +1591,14 @@ int hud_target_ship_can_be_scanned(ship *shipp)
 	if (shipp->flags[Ship::Ship_Flags::Scannable])
 		return 1;
 
-	// ignore ships that don't carry cargo
-	if ((sip->class_type < 0) || !(Ship_types[sip->class_type].flags[Ship::Type_Info_Flags::Scannable]))
+	if (!Use_new_scanning_behavior) {
+		// ignore ships that don't carry cargo
+		if ((sip->class_type < 0) || !(Ship_types[sip->class_type].flags[Ship::Type_Info_Flags::Scannable]))
+			return 0;
+	} else {
+		// In new behavior if we are here then the ship does not have the scannable flag so it's not scannable!
 		return 0;
+	}
 
 	return 1;
 }
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -449,11 +449,10 @@ void parse_mod_table(const char *filename)
 				stuff_boolean(&HUD_shadows);
 			}
 
-		if (optional_string("$Unify scanning behavior:")) {
-			stuff_boolean(&Use_new_scanning_behavior);
-		}
+			if (optional_string("$Unify scanning behavior:")) {
+				stuff_boolean(&Use_new_scanning_behavior);
+			}
 
-		optional_string("#SEXP SETTINGS");
 			if (optional_string("$Don't show callsigns in the escort list:")) {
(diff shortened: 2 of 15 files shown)


Baezon:
As an aside, I think there's a good argument to be made this be default for mods targeting 24 and up.

code/hud/hudtarget.cpp
```diff
@@ -1591,9 +1591,14 @@ int hud_target_ship_can_be_scanned(ship *shipp)
 	if (shipp->flags[Ship::Ship_Flags::Scannable])
 		return 1;
 
-	// ignore ships that don't carry cargo
-	if ((sip->class_type < 0) || !(Ship_types[sip->class_type].flags[Ship::Type_Info_Flags::Scannable]))
+	if (!Use_new_scanning_behavior) {
```
naomimyselfandi, line 1594: Inverting the condition here would let this be unified with the previous check, making it easier to read.

```cpp
// The old behavior treats some ship types as always scannable. The new behavior only checks the ship's Scannable flag.
if (shipp->flags[Ship::Ship_Flags::Scannable])
    return 1;
} else if (Use_new_scanning_behavior) {
    return 0;
} else if ((sip->class_type < 0) || !(Ship_types[sip->class_type].flags[Ship::Type_Info_Flags::Scannable])) {
    return 0;
}
```

code/playerman/playercontrol.cpp
```diff
@@ -1608,24 +1606,34 @@ bool player_inspect_cargo(float frametime, char *outstr)
...
 		if ( dot < CARGO_MIN_DOT_TO_REVEAL ) {
-			if (reveal_cargo)
-				strcpy(outstr,XSTR( "cargo: <unknown>", 86));
-			else
-				strcpy(outstr,XSTR( "not scanned", 87));
-			hud_targetbox_end_flash(TBOX_FLASH_CARGO);
-			Player->cargo_inspect_time = 0;
-			return true;
+			if (reveal_cargo) {
+				if (cargo_sp->cargo_title[0] != '\0') {
```
naomimyselfandi, line 1610: #5922 asked for the ability to hide the `cargo:` prefix entirely. Could we support that as well? I hate to suggest overloading the hash sign yet again, but it is something of a convention.

MjnMixael: Ah, good thought. I'll come up with something for that.

MjnMixael: Title starting with '#' will now make the Title entirely invisible.

code/hud/hudtarget.cpp
MjnMixael: Good call. Done.

code/ship/ship.cpp
```diff
@@ -657,6 +658,7 @@ ship_flag_description Ship_flag_descriptions[] = {
 	{ Ship_Flags::No_builtin_messages,			"Ship will not send any persona messages."},
 	{ Ship_Flags::Scramble_messages,			"All messages sent from this ship will appear scrambled, as if the ship had been hit by an EMP."},
 	{ Ship_Flags::Maneuver_despite_engines,		"Ship can maneuver even if its engines are disabled or disrupted" },
+	{ Ship_Flags::No_scanned_cargo,             "Ship cargo will never be revealed and will instead only show scanned or not scanned."},
```
BMagnu, line 661: AFAICT, this flag only works in the new scanning mode. This description should ideally note this.

BMagnu (changes requested):
Looks good overall. 
One small documentation thing.
In addition, it seems like you addressed the comments naomi made, but I don't see repsective changes in the code for these. Is there maybe a commit still missing and waiting to be pushed? I'd love to see the changes requested by naomi in the PR as well.

code/ship/ship.cpp
MjnMixael: Done. The PR is now also showing the updates Naomi requested.

MjnMixael: Er.. maybe not. I'm rather confused on my commit history here and it's been long enough I can't recall exactly what Naomi and I landed on for those requests. I wouldn't have said "done" if I didn't do anything.

code/mod_table/mod_table.cpp
```diff
@@ -448,6 +449,11 @@ void parse_mod_table(const char *filename)
 				stuff_boolean(&HUD_shadows);
 			}
 
+		if (optional_string("$Unify scanning behavior:")) {
+			stuff_boolean(&Use_new_scanning_behavior);
+		}
+
+		optional_string("#SEXP SETTINGS");
```
BMagnu, line 456: Uh, this seems like it got here by accident

BMagnu (changes requested):
Hm, the naomi-requested changes still seem like they haven't properly propagated to this PR.

code/mod_table/mod_table.cpp
MjnMixael: Ok I just kinda redid things manually. I think everything's addressed now including this conflict resolution error

BMagnu (approved):

PR #6001, opened 2024-02-23
add Team virtvar to parse objects

Goober5000:
Previously, Team was an attribute available on ships, but not parsed ships.


wookieejedi (approved):
Nice and straightforward

PR #6005, opened 2024-02-26
add extra ways to modify background elements in missions

Goober5000:
This adds Lua API methods for getting suns and bitaps by index, as well as modifying their scale and position.  This removes the need to completely add and remove a background element that only needs to be modified.


wookieejedi (approved):
Tested and all new features work and everything else works as expected. Would be useful to get a second set of eyes on this, too

BMagnu (changes requested):
Some small niceties

BMagnu (approved):

PR #6006, opened 2024-02-26
allow invalid arguments to be nil

Goober5000:
A design choice very early in FSO's scripting development established that invalid results return invalid objects.  By using a game_settings.tbl flag, this allows invalid results to return nil instead.

~~This also standardizes on `isValid()` (versus capitalized versions) and makes validity member functions `const`.~~  See #6009 

Thanks to Lafiel for help with the template stuff.

### DIFF
--- a/code/mod_table/mod_table.cpp
+++ b/code/mod_table/mod_table.cpp
@@ -150,6 +150,7 @@ bool Randomize_particle_rotation;
 bool Calculate_subsystem_hitpoints_after_parsing;
 bool Disable_internal_loadout_restoration_system;
 bool Contrails_use_absolute_speed;
+bool Lua_API_returns_nil_instead_of_invalid_object;
 
 static auto DiscordOption __UNUSED = options::OptionBuilder<bool>("Game.Discord",
                      std::pair<const char*, int>{"Discord Presence", 1754},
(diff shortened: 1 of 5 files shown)


code/scripting/api/objs/subsystem.cpp
```diff
@@ -26,12 +26,12 @@ ship_subsys_h::ship_subsys_h() : object_h() {
 ship_subsys_h::ship_subsys_h(object* objp_in, ship_subsys* sub) : object_h(objp_in) {
 	ss = sub;
 }
-bool ship_subsys_h::isSubsystemValid() const { return object_h::IsValid() && objp->type == OBJ_SHIP && ss != nullptr; }
+bool ship_subsys_h::isSubsystemValid() const { return object_h::isValid() && objp->type == OBJ_SHIP && ss != nullptr; }
```
BMagnu, line 29: So, I wonder whether it's possible to make this an override of isValid instead of a different function to allow this change to also be usable for subsystems.
Then again, if a significant portion of the codebase relies on being able to distinguish whether a subsystem is invalid but its parent object is not, that might be hard.

BMagnu (commented):
Looks generally good.
One question on submodel handles.
Otherwise, since I had significant involvement in designing that template, someone else should probably at least give it a peek.

code/scripting/api/objs/subsystem.cpp
Goober5000: Why would it be an override of `object_h`'s `isValid`?  The `ship_subsys_h` handle isn't a subclass of `object_h`.

And yes, subsystems can be valid if the parent ship isn't valid.  An example would be looking up a nonexistent subsystem on a valid ship.

BMagnu: Unfortunately, ship_subsys_h _is_ a subclass of object_h, for whatever godforsaken reason.
ship_subsys isn't of object, but their lua handles are related classes

Goober5000: Yeah, you're right, the subsystem handle should also use `isValid`.  I'll make the change and roll it into the earlier refactor commit.

Goober5000:
Moving to draft pending the merge of #6009.

BMagnu (approved):

PR #6007, opened 2024-02-26
Fix broken collision pair caching

BMagnu:
Turns out, when we last increase the maximum object count, we inadvertently broke collisions.
Collision pair caching used a 12-bit bitshift, so any objects with an objnum >= 4096 (which are quite common in large missions with our new cap) could overwrite collision data from older collision pairs.
For stuff like weapons or other things with short lifetimes, this isn't really noticable, but it can lead to something like a "will-never-collide" of a ship-weapon pair to overwrite the next collision check of a ship-ship pair, causing that ship-ship pair to never collide again.
This PR does two things:
1. Increase the bitshift to 16 bits, so that we actually use all available space in the collision pair key and no longer run into issues with 5000 objects.
2. Add a static_assert that will error out if the MAX_OBJECTS define is increased past the limit given by the 16 bits from the collision pair cache.

Since this is quite the severe regression, I suggest a point release for 24.0 from this.

### DIFF
--- a/code/globalincs/globals.h
+++ b/code/globalincs/globals.h
@@ -72,6 +72,7 @@
 #define MAX_POLYGON_MODELS  300
 
 // object.h
+//If this value exceeds 2^16-1, this will break collision pair caching as is. Proceed with caution.
 #define MAX_OBJECTS			5000	//Increased from 3500 to 5000 in 2022	
 
 // from weapon.h (and beam.h)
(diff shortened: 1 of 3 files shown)


The-E (changes requested):
Maybe add a comment to MAX_OBJECTS about this as well?

wookieejedi:
Tested and fixes issues and things work as expected. Thanks for figuring this out!

The-E (approved):

PR #6009, opened 2024-02-26
isValid refactor

Goober5000:
Refactor the scripting code in several ways to prepare for #6006:

1. Standardize on `isValid()` versus other capitalization
2. Make validity member functions `const`.
3. Prevent `object_h` from being subclassed; change the other handles that use it to do so via member variables
4. Change `isSubsystemValid` to `isValid`

### DIFF
--- a/code/scripting/api/objs/subsystem.cpp
+++ b/code/scripting/api/objs/subsystem.cpp
@@ -26,12 +26,12 @@ ship_subsys_h::ship_subsys_h() : object_h() {
 ship_subsys_h::ship_subsys_h(object* objp_in, ship_subsys* sub) : object_h(objp_in) {
 	ss = sub;
 }
-bool ship_subsys_h::isSubsystemValid() const { return object_h::IsValid() && objp->type == OBJ_SHIP && ss != nullptr; }
+bool ship_subsys_h::isSubsystemValid() const { return object_h::isValid() && objp->type == OBJ_SHIP && ss != nullptr; }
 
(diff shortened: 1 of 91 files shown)


code/scripting/api/objs/subsystem.cpp
```diff
@@ -20,19 +20,16 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum,
...
+ship_subsys_h::ship_subsys_h(object* objp_in, ship_subsys* sub) : objh(objp_in), ss(sub) {}
+
+bool ship_subsys_h::isValid() const { return objh.isValid() && objh.objp->type == OBJ_SHIP && ss != nullptr; }
 
 void ship_subsys_h::serialize(lua_State* /*L*/, const scripting::ade_table_entry& /*tableEntry*/, const luacpp::LuaValue& value, ubyte* data, int& packet_size) {
 	ship_subsys_h subsys;
 	value.getValue(l_Subsystem.Get(&subsys));
-	const ushort& netsig = subsys.IsValid() ? subsys.objp->net_signature : 0;
-	const int& subsys_index = subsys.isSubsystemValid() ? ship_get_subsys_index(subsys.ss) : -1;
+	const ushort& netsig = subsys.isValid() ? subsys.objh.objp->net_signature : 0;
```
BMagnu, line 31: This probably needs to check for subsys.objh.isValid(), since it can (and maybe should) still send an invalid subsys of a valid obj.

BMagnu (changes requested):

code/scripting/api/objs/subsystem.cpp
Goober5000: Point.  Changed.

BMagnu (approved):

PR #6016, opened 2024-02-28
fix help descriptions for string-greater-than and string-less-than

Goober5000:


wookieejedi (approved):
Nice simple and tiny documentation change

PR #6020, opened 2024-02-29
Fix retail options not saving

MjnMixael:
#5895 turned `Using_in_game_options` on by default. This caused FSO to load all options from the new Options Manager values rather than the old registry values. Since retail obviously doesn't use SCPUI, we need to make sure that we then properly save retail options values to Options Manager also.


wookieejedi (approved):
Works for me, but would be good to have another set of eyes. Definitely an important PR to get in sooner rather than later since game options don't work as expected without merging it.

PR #6021, opened 2024-03-01
Unlimited Asteroid/Debris Field Types & Subtypes

MjnMixael:
This PR does two main things. First, it allows debris fields to have unlimited types of debris active. Second it entirely refactors how asteroid subtypes are handled so that asteroid subtypes are no longer limited to retail's "Brown", "Blue", and "Orange" color types. The new parsing allows defining an arbitrary number of asteroid subtypes and an asteroid field can have unlimited subtypes active as well.

A follow-up PR will enhance the current asteroid field sexps to allow enabling more than three debris types or three asteroid subtypes.

The FRED Asteroid dialog editor has been updated to allow selecting multiple debris types or asteroid subtypes. QtFRED has been updated to not CTD, but the editor was written to be very specific to retail's asteroid data structure... so in order to keep this already large PR manageable I have opted to handle the QtFRED update in a follow-up PR.


wookieejedi:
This will be very useful for many folks, and we are also excited to use this functionality in FotG, too.

code/asteroid/asteroid.cpp
```diff
@@ -796,35 +806,13 @@ void asteroid_create_debris_field(int num_asteroids, int asteroid_speed, int deb
...
-			debris3 = -1;
-			count++;
-			continue;
-		}
-	}
+	Asteroid_field.field_debris_type = debris_types;
+
+	int count = static_cast<int>(Asteroid_field.field_debris_type.size());
 
 	Asteroid_field.num_used_field_debris_types = count;
```
BMagnu, line 817: Since we're already refactoring here, we should just remove `num_used_field_debris_types` entirely and use `Asteroid_field.field_debris_type.size()` in it's place everywhere it's used. Less opportunity for these variables to go out of sync and break stuff...

code/asteroid/asteroid.cpp
```diff
@@ -262,6 +262,36 @@ static void inner_bound_pos_fixup(asteroid_field *asfieldp, vec3d *pos)
...
+	for (size_t j = 0; j < Asteroid_field.field_asteroid_type.size(); j++) {
+		if (counter == 0) {
+			subtype = Asteroid_field.field_asteroid_type[j];
+			break;
+		} else
+			counter--;
+	}
+
+	return subtype;
+}
```
BMagnu, line 282: Since `num_used_field_debris_types` is equal to `field_asteroid_type.size()`, this effectively just boils down to a much much simpler function with no loop:
```suggestion
const SCP_string& pick_random_asteroid_type()
{
	return Asteroid_field.field_asteroid_type[Random::next(Asteroid_field.field_asteroid_type.size())];
}
```

code/asteroid/asteroid.cpp
```diff
@@ -625,22 +661,10 @@ void asteroid_create_all()
...
-			asteroid_load(ASTEROID_TYPE_SMALL, 1);
-			asteroid_load(ASTEROID_TYPE_MEDIUM, 1);
-			asteroid_load(ASTEROID_TYPE_LARGE, 1);
-		}
-
-		if (Asteroid_field.field_asteroid_type[2]) {
-			asteroid_load(ASTEROID_TYPE_SMALL, 2);
-			asteroid_load(ASTEROID_TYPE_MEDIUM, 2);
-			asteroid_load(ASTEROID_TYPE_LARGE, 2);
+		for (auto subtype : Asteroid_field.field_asteroid_type) {
```
BMagnu, line 664: This should be a const auto& subtype to avoid a string copy here

code/asteroid/asteroid.cpp
```diff
@@ -796,35 +806,13 @@ void asteroid_create_debris_field(int num_asteroids, int asteroid_speed, int deb
...
-			continue;
-		}
-		if (debris3 >= 0) {
-			Asteroid_field.field_debris_type[i] = debris3;
-			debris3 = -1;
-			count++;
-			continue;
-		}
-	}
+	Asteroid_field.field_debris_type = debris_types;
```
BMagnu, line 813: Here, you copy debris_types twice. Once when you call the function, and then again here.
To avoid this, just replace this assignment here with a move assignment:
```suggestion
	Asteroid_field.field_debris_type = std::move(debris_types);
```

code/asteroid/asteroid.cpp
```diff
@@ -2233,15 +2207,71 @@ static void asteroid_parse_section()
...
+			stuff_string(thisType.type_name, F_NAME);
+
+			required_string("+POF file:");
+			stuff_string(thisType.pof_filename, F_NAME, MAX_FILENAME_LEN);
+
+			thisType.model_number = -1;
+			thisType.model_pointer = nullptr;
+
+			bool exists = false;
+			for (auto type : asteroid_p->subtypes) {
```
BMagnu, line 2260: This auto should be const auto& as well

code/asteroid/asteroid.cpp
```diff
@@ -2439,6 +2469,31 @@ int get_asteroid_index(const char* asteroid_name)
 	return -1;
 }
 
+// For FRED. Gets a list of unique asteroid subtype names
+SCP_vector<SCP_string> get_list_valid_asteroid_subtypes()
+{
+	SCP_vector<SCP_string> list;
+
+	for (auto this_asteroid : Asteroid_info) {
```
BMagnu, line 2477: This one as well

code/asteroid/asteroid.cpp
```diff
@@ -2439,6 +2469,31 @@ int get_asteroid_index(const char* asteroid_name)
...
+		if (this_asteroid.type != ASTEROID_TYPE_DEBRIS) {
+			for (auto subtype : this_asteroid.subtypes) {
```
BMagnu, line 2479: This one as well

code/asteroid/asteroid.cpp
```diff
@@ -2439,6 +2469,31 @@ int get_asteroid_index(const char* asteroid_name)
...
+				bool exists = false;
+				for (auto entry : list) {
```
BMagnu, line 2481: This one as well

code/asteroid/asteroid.cpp
```diff
@@ -2511,14 +2573,17 @@ static void verify_asteroid_list()
...
-		if (Asteroid_info[i].pof_files[0][0] == '\0')
+		/*if (Asteroid_info[i].pof_files[0][0] == '\0')
 			Error(LOCATION, "Missing [POF File 1] for %s", Asteroid_info[i].name);
 
 		if (Asteroid_info[i].type != ASTEROID_TYPE_DEBRIS) {
 			if (Asteroid_info[i].pof_files[1][0] == '\0')
 				Error(LOCATION, "Missing [POF File 2] for %s", Asteroid_info[i].name);
 			if (Asteroid_info[i].pof_files[2][0] == '\0')
 				Error(LOCATION, "Missing [POF File 3] for %s", Asteroid_info[i].name);
+		}*/
```
BMagnu, line 2584: Just get rid of the old code here

code/asteroid/asteroid.cpp
```diff
@@ -2740,19 +2810,27 @@ void asteroid_page_in()
 					}
 				} else {
 					// ASTEROID FIELDS - use subtype k, if active
-					if (!Asteroid_field.field_asteroid_type[k]) {
+					bool active = false;
+					for (auto active_type : Asteroid_field.field_asteroid_type) {
```
BMagnu, line 2814: const auto& again

code/asteroid/asteroid.h
```diff
@@ -54,13 +53,20 @@ typedef struct asteroid_split_info {
 	int		max;				//maximum asteroids to spawn
 } asteroid_split_info;
 
+// Data structure for storing asteroid subtype info. POFs, model pointer, model num, etc.
+typedef struct asteroid_subtype_info {
+	char        pof_filename[MAX_FILENAME_LEN];
+	polymodel*  model_pointer;
```
BMagnu, line 59: I'm a bit wary on saving model pointers, especially when we also store the index which should be used to retrieve them when accessing the model, but since the old code did so too, I guess it's fine for now...

code/asteroid/asteroid.cpp
```diff
@@ -732,25 +746,21 @@ void asteroid_create_asteroid_field(int num_asteroids, int field_type, int aster
...
 		count++;
 	}
 	if (blue) {
-		Asteroid_field.field_asteroid_type[1] = true;
+		Asteroid_field.field_asteroid_type.push_back("Blue");
 		count++;
 	}
 	if (orange) {
-		Asteroid_field.field_asteroid_type[2] = true;
+		Asteroid_field.field_asteroid_type.push_back("Orange");
```
BMagnu, line 763: You refer to these retail names a few times, here, in missionparse, etc.
At that point, these should probably become defines or something.

code/mission/missionparse.cpp
```diff
@@ -5845,68 +5845,83 @@ void parse_asteroid_fields(mission *pm)
...
+						ast_name = colors[k];
+						break;
 					}
 				}
+
+				auto list = get_list_valid_asteroid_subtypes();
+
+				//validate the asteroid subtype name
+				bool valid = false;
+				for (auto entry : list) {
```
BMagnu, line 5912: const auto&

qtfred/src/mission/dialogs/AsteroidEditorDialogModel.cpp
```diff
@@ -96,14 +96,52 @@ bool AsteroidEditorDialogModel::getEnhancedEnabled()
...
+
+	SCP_string name = "Brown";
+	if (type == _AST_BLUE) {
+		name = "Blue";
+	} else if (type == _AST_ORANGE) {
+		name = "Orange";
+	}
+
+	bool in_list = false;
+	for (auto asteroid : _field_asteroid_type) {
```
BMagnu, line 109: const auto&

BMagnu (changes requested):
A few stylistic points.

code/asteroid/asteroid.cpp
MjnMixael: That's a good idea. The sexps are going to need quite a bit of rewrite to handle this new code, so I'll punt on this for when I redo those in a follow-up.

code/asteroid/asteroid.cpp
MjnMixael: Took care of this. There is one spot it was used for "backwards compatibility" that seems less than ideal. You'll see my comments in the code. I'd like your opinion.

code/asteroid/asteroid.h
MjnMixael: I'm with you. I don't have a design in my head to undo that and I can't fully invest time into refactoring that right now either. Good thing to do in a follow-up, though.

code/asteroid/asteroid.cpp
BMagnu: I'd agree on less then ideal, but it's fine for now. Probably indeed hardcoded for retail or something.
If it turns out that this isn't needed anymore at some point, we can still remove it later.

code/asteroid/asteroid.cpp
BMagnu: I see, fair enough.

BMagnu (approved):
Seems good.

PR #5302, opened 2023-03-31
Scpui Hud Config API

MjnMixael:
API for recreating the HUD configuration UI. This one's a little messier than I'd like but a lot of this HUD stuff is hardcoded here. I think there's significant opportunity for improvement and customization, but I'll save that for a follow up PR.

This also vectorizes the .hcf file list and removes the limit it had.

EDIT: I should add that I tested the API pretty thoroughly. I also tested the retail HUD Configuration UI in both 1024 and 480 just to be sure they still work as expected.

PR #6022, opened 2024-03-02
Fix CTD with HUD Preset Switching

wookieejedi:
After #5302 it was discoved that pressing the `Next` button in the HUD presets would cause a CTD.

Thanks for @BMagnu for figuring this out
`If HC_current_file is HC_preset_filenames.size() - 1, then it'll oob on a vector access.`

Testing this PR fixes that crash.

### DIFF
--- a/code/hud/hudconfig.cpp
+++ b/code/hud/hudconfig.cpp
@@ -1323,10 +1323,8 @@ void hud_config_button_do(int n)
 			break;
 		}
 
-		if (HC_current_file >= (int)HC_preset_filenames.size()) {
+		if ((++HC_current_file) >= (int)HC_preset_filenames.size()) {
 			HC_current_file = 0;
-		} else {
-			HC_current_file++;
 		}
 
 		// load em up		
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):

code/hud/hudconfig.cpp
```diff
@@ -1323,10 +1323,8 @@ void hud_config_button_do(int n)
 			break;
 		}
 
-		if (HC_current_file >= (int)HC_preset_filenames.size()) {
+		if ((++HC_current_file) >= (int)HC_preset_filenames.size()) {
```
MjnMixael, line 1326: This should use a static_cast<int>. Or you could also just SCP_vector_inbounds for a more complete check.

MjnMixael (changes requested):

code/hud/hudconfig.cpp
BMagnu: > Or you could also just SCP_vector_inbounds for a more complete check
Possible, but when using this it's important to make the increment of the index before the check

wookieejedi: Sounds good, updated with static_cast for now

MjnMixael (approved):

PR #6023, opened 2024-03-02
add weapon preloader

Goober5000:
`ship-create` has had a preloader for many years, but `weapon-create` and `beam-create` did not.  This remedies that.


wookieejedi (approved):

PR #6025, opened 2024-03-05
add support for decompression of DDS textures

### LINKED ISSUE #4920: Hercules, Zeus, other ships render as black with newer MediaVPs on Mac
Just documenting this as a known issue in case anyone else notices it and wonders what's going on.

The OpenGL implementation on macOS does not have the `GL_ARB_texture_compression_bptc` extension and therefore lacks support for BC7 textures. As there is no fallback if BC7 isn't supported, any textures using that format are unable to render.

Metal *does* support BPTC however so it's an issue specific to the OpenGL implementation on macOS and not a general issue with the platform. Adding support for Metal (or Vulkan) is the only good way to address this problem. Mods looking to have good macOS support should avoid using BC7 compressed textures until a such time as a code solution has been implemented.

notimaginative:
Adds bcdec lib to convert compressed DDS textures to RGBA based on what the system supports. BC7 if BPTC extension not supported, DXT? if S3TC extension not supported.

Should support cubemaps and volume textures but those were not tested.

Addtional changes:
 - remove indexed DDS support since it had bugs but didn't work anyway (R8 should be used instead)
 - fix issues with volume textures (nothing to test with)
 - remove old compression support flags as they are now obsolete

Fixes #4920

### DIFF
--- a/code/bmpman/bmpman.cpp
+++ b/code/bmpman/bmpman.cpp
@@ -1027,10 +1027,6 @@ int bm_is_compressed(int num) {
 	auto entry = bm_get_entry(num);
 	BM_TYPE type = BM_TYPE_NONE;
 
-	//duh
-	if (!Use_compressed_textures)
-		return 0;
-
 	type = entry->comp_type;
 
 	switch (type) {
--- a/code/ddsutils/bcdec.h
+++ b/code/ddsutils/bcdec.h
@@ -1,0 +1,1329 @@
+﻿/* bcdec.h - v0.96
+   provides functions to decompress blocks of BC compressed images
+   written by Sergii "iOrange" Kudlai in 2022
+
+   This library does not allocate memory and is trying to use as less stack as possible
+
+   The library was never optimized specifically for speed but for the overall size
+   it has zero external dependencies and is not using any runtime functions
+
+   Supported BC formats:
+   BC1 (also known as DXT1) + it's "binary alpha" variant BC1A (DXT1A)
+   BC2 (also known as DXT3)
+   BC3 (also known as DXT5)
+   BC4 (also known as ATI1N)
+   BC5 (also known as ATI2N)
+   BC6H (HDR format)
+   BC7
+
+   BC1/BC2/BC3/BC7 are expected to decompress into 4*4 RGBA blocks 8bit per component (32bit pixel)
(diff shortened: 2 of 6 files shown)


notimaginative:
As a memory optimization this does resize the de-compressed textures using mimaps. After testing a max size of 1024x1024 was chosen. Without this the game can easily become unusable due to the size of textures in use by some mods.

As an example, the Triton from the MVPS. Uncompressed at full size the complete set of textures with mipmaps requires ~1.5 GB of VRAM. After the "resize" it's down to a more manageable 24 MB.

wookieejedi:
Thank you for all your time with creating this PR, it will be fantastic to finally have BC7 textures finally not just be black on MACs

chief1983:
This worked great on my Mac with the test mission I had received previously that used a BC7 planet texture.  I did notice a couple of issues in the logs that were probably not related to this change, but wanted to document here and figured you could look into later if they sound worth any investigation, or decide they aren't a problem:
```
./fs2_open_24_1_0_arm64-DEBUG.app/Contents/MacOS/fs2_open_24_1_0_arm64-DEBUG -mod fotg -window
2024-03-25 16:17:35.582 fs2_open_24_1_0_arm64-DEBUG[53262:14339740] WARNING: Secure coding is automatically enabled for restorable state! However, not on all supported macOS versions of this application. Opt-in to secure coding explicitly by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState:.
UNSUPPORTED (log once): POSSIBLE ISSUE: unit 0 GLD_TEXTURE_INDEX_2D_ARRAY is unloadable and bound to sampler type (Float) - using zero texture because texture unloadable
```

notimaginative:
The first warning appears to be new as I only noticed it showing up recently. Are you on 14.4+ by any chance? It was after the update that I noticed it and I don't recall seeing it on earlier versions, but I could be wrong. Either way it's on my todo list.

That second warning (the GLD_TEXTURE_INDEX_2D_ARRAY one) has been there a while and I'm really not sure where it's coming from. I've tried debugging it several times but haven't figured it out. Nothing is obviously broken as far as I can tell so I've just decided to ignore it until proven otherwise.

chief1983:
I'm apparently on 14.3.1, didn't know 14.4 was out yet.  2021 Macbook M1 Pro.

wookieejedi (approved):
Given both taylor and chief (the main Mac devs) have tested, I'm going to go ahead and approve this to help spur on movement.

PR #6027, opened 2024-03-05
limit debug log spam from file shadows

notimaginative:
Always log the possible shadow if:
 1) it's a critical file
 2) it's in the same root

Otherwise log it behind a debug filter ("CFileSystem").


notimaginative:
This worked for me in a limited test but please give it a try @EatThePath and confirm it's catching all that it should.

BMagnu (approved):

Issue #6026, opened 2024-03-05
Index out of bounds when attempting to modify a container key


Goober5000:
In this test mission...
[container numbers.fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/14491465/container.numbers.fs2.txt)
...there is a container that maps strings to numbers.  Right-clicking on the `<any string>` sexp node in this event causes FRED to crash.
![crash event](https://github.com/scp-fs2open/fs2open.github.com/assets/1878523/a9ac2c45-d5e3-47a3-b68d-bbe0dd6d2e4b)

The "numbers" container contains a single entry, mapping the string "key" to the number 17.  The exception occurs in sexp_tree.cpp line 919, in an attempt to access the -1 index of the `Operators` vector, because the `op` variable is somehow equal to -1.

PR #6030, opened 2024-03-08
Fix crash when right-clicking on a container modifier

jg18:
Fix regression from PR #5793.

Fixes issue #6026.


JohnAFernandez (approved):

jg18:
I'd like to think on this a little bit, to avoid unintended consequences, but I think the change is the desired behavior;.

Goober5000 (approved):
Looks good and fixes the crash in my tests.

Issue #6029, opened 2024-03-07
Containers do not return the proper numbers in eval_num


Goober5000:
Test mission:
[TestEscort.fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/14518847/TestEscort.fs2.txt)

There is a string-to-number map containing ships and their escort priorities.  The mission attempts to add each using `add-remove-escort`.  However, the `eval_num` function does not correctly evaluate the container, causing the value to be interpreted as an unknown string (OP_NOT_AN_OP) that evaluates to 0.

jg18:
Fixed with PR #6031.

PR #6031, opened 2024-03-08
Handle container data nodes correctly with eval_num()

jg18:
Fixes Issue #6029.


Goober5000 (approved):
Nice and simple, and I confirmed the fix works.

PR #6034, opened 2024-03-09
Add scripted model 'lightning' arcs

Baezon:
- Adds `AddElectricArc` script function, to add those lightning arcs to a ship arbitrarily
- Removes `MAX_SHIP_ARCS` in lieu of simply `MAX_ARC_EFFECTS`. No reason to limit ships to 5 when the model limit is 8. Additionally `MAX_SHIP_DAMAGE_ARCS` limits the number of specifically damage/emp arcs at once so any limits as far as visuals are concerned are preserved
- Ship and scripted arcs can now have an explicit width
- Also fixes some bonehead mistakes I made with the docs on `NumVertices` and `GetVertex`


code/ship/shipfx.cpp
```diff
@@ -2176,6 +2176,20 @@ void shipfx_do_lightning_arcs_frame( ship *shipp )
 
 						shipp->arc_type[j] = MARC_TYPE_SHIP;
 
+						if (arc_info->width > 0.0) {
+							shipp->arc_width[j] = arc_info->width;
+						} else {
+							// same as MARC_TYPE_DAMAGED
```
Goober5000, line 2182: What does `// same as MARC_TYPE_DAMAGED` actually mean here?  The comment deleted from modelrender.cpp, 
`// try and scale the size a bit so that it looks equally well on smaller vessels`, was reasonably descriptive, but I'm not sure how to interpret the new comment.

code/ship/shipfx.cpp
```diff
@@ -2176,6 +2176,20 @@ void shipfx_do_lightning_arcs_frame( ship *shipp )
 
 						shipp->arc_type[j] = MARC_TYPE_SHIP;
 
+						if (arc_info->width > 0.0) {
```
Goober5000, line 2179: This should be `0.0f`

code/ship/ship.cpp
```diff
@@ -5038,6 +5038,13 @@ static void parse_ship_values(ship_info* sip, const bool is_template, const bool
 		stuff_float(&new_info.duration);
 		required_string("+Frequency:");
 		stuff_float(&new_info.frequency);
+
+		if (optional_string("+Width:")) {
+			stuff_float(&new_info.width);
+		} else {
+			new_info.width = 0.0;
```
Goober5000, line 5045: This should also be `0.0f`

code/scripting/api/objs/ship.cpp
```diff
@@ -2740,6 +2740,49 @@ ADE_FUNC(jettison, l_Ship, "number jettison_speed, [ship... dockee_ships /* All
...
+
+	if (!objh->isValid())
+		return ADE_RETURN_FALSE;
+
+	auto shipp = &Ships[objh->objp->instance];
+
+	// spawn the arc in the first unused slot
+	for (int i = 0; i < MAX_ARC_EFFECTS; i++) {
+		if (!shipp->arc_timestamp[i].isValid()) {
+			shipp->arc_timestamp[i] = _timestamp((int)(duration * MILLISECONDS_PER_SECOND));
```
Goober5000, line 2765: Use a C++-style cast or the `fl2i` (or `fl2ir`) macro here.

Goober5000 (commented):
A few comments

code/ship/shipfx.cpp
Baezon: It uses the same width calculation `MARC_TYPE_DAMAGED` arcs use in `model_render_add_lightning`, I'll comment to that effect.

Goober5000 (approved):

PR #6035, opened 2024-03-11
enhancements for submodel collisions

Goober5000:
In the first commit:
* Expose NoCollide and NoCollideThisOnly as virtual variables for submodels
* Allow submodels to be optionally specified in object:checkRayCollision
* Be sure to calculate ship-to-ship hit position before the override collision hook is called, in case the hook needs this value.  (The other collision hooks already calculate the hit position before the override hook.)

And in the second:
* Add the submodel that was hit to the On Ship Collision scripting hook.  (This will only be non-nil if the ship was the "heavy object".)
* Add a `getModel()` function to the scripted submodel object.
* Fix some incorrect documentation in `OnShipCollision`.

These will enhance the movement script in Star Fox: Event Horizon.

### DIFF
--- a/code/scripting/api/objs/model.cpp
+++ b/code/scripting/api/objs/model.cpp
@@ -412,6 +412,19 @@ ADE_FUNC(GetVertex, l_Submodel, nullptr, "Gets the specified vertex, or a random
 	return ade_set_args(L, "o", l_Vector.Set(vert));
 }
 
+ADE_FUNC(getModel, l_Submodel, nullptr, "Gets the model that this submodel belongs to", "model", "A model, or an invalid model if the handle is not valid")
+{
+	submodel_h *smh = nullptr;
+
+	if (!ade_get_args(L, "o", l_Submodel.GetPtr(&smh)))
+		return ade_set_error(L, "o", l_Model.Set(model_h()));
+
+	if (!smh->isValid())
+		return ade_set_error(L, "o", l_Model.Set(model_h()));
+
+	return ade_set_args(L, "o", l_Model.Set(model_h(smh->GetModelID())));
+}
+
 ADE_FUNC(getFirstChild, l_Submodel, nullptr, "Gets the first child submodel of this submodel", "submodel", "A submodel, or nil if there is no child, or an invalid submodel if the handle is not valid")
 {
 	submodel_h *smh = nullptr;
--- a/code/scripting/api/objs/model.cpp
+++ b/code/scripting/api/objs/model.cpp
@@ -368,7 +368,6 @@ ADE_VIRTVAR(Radius, l_Submodel, nullptr, "Gets the submodel's radius", "number",
 	return ade_set_args(L, "f", smh->GetSubmodel()->rad);
 }
 
-
 ADE_FUNC(NumVertices, l_Submodel, nullptr, "Returns the number of vertices in the submodel's mesh", "submodel", "The number of vertices, or 0 if the submodel was invalid")
 {
 	submodel_h* smh = nullptr;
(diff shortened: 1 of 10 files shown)


code/scripting/api/objs/model.cpp
```diff
@@ -473,6 +485,38 @@ ADE_FUNC(isValid, l_Submodel, nullptr, "True if valid, false or nil if not", "bo
...
+	submodel_h* smh = nullptr;
+
+	if (!ade_get_args(L, "o", l_Submodel.GetPtr(&smh)))
+		return ade_set_error(L, "b", false);
+
+	if (!smh->isValid())
+		return ade_set_error(L, "b", false);
+
+	if (ADE_SETTING_VAR)
+		smh->GetSubmodel()->flags.set(Model::Submodel_flags::No_collisions);
```
Baezon, line 499: I am really not a fan of allowing scripts to set values in static data like the models themselves. Is there a reason this couldn't be a flag/function on a `submodel_instance` instead? This would also make it more useful in general, since you could toggle specific ship submodels on and off, rather than all ships using that model.

code/scripting/global_hooks.cpp
```diff
@@ -108,17 +108,16 @@ const std::shared_ptr<Hook<ShipSourceConditions>> OnDebrisCreated = Hook<ShipSou
...
-			"The ship object with which the \"other\" object collided with. Provided for consistency with other "
-			"collision hooks."},
-		{"Ship", "ship", "Same as \"Object\""},
+	"with each ship as the \"Self\" parameter.",
+	{{"Self", "object", "The object the ship collided with."},
+		{"Object", "ship",
+			"The ship that collided with \"Self\". Provided for consistency with other collision hooks."},
+		{"Ship", "ship", "For ship-on-ship collisions, the same as \"Self\". For ship-on-object collisions, the same as \"Object\"."},
 		{"Hitpos", "vector", "The world position where the collision was detected"},
+		{"ShipSubmodel", "submodel", "The submodel of \"Ship\" involved in the collision, if \"Ship\" was the heavy object"},
```
Baezon, line 117: Pedantic, but "was the heavier object" might be a bit more clear to someone not familiar with the internal significance of the 'heavy' and 'light' object.

Baezon (changes requested):

code/scripting/global_hooks.cpp
Goober5000: Ok, changed.

code/scripting/api/objs/model.cpp
Goober5000: Whoops.  Now it handles setting the flag.

On the one hand, I ended up using a different strategy in the script I was writing, so toggling the flag is no longer needed for my particular use case.  On the other hand, this might still be useful for another script.

This can't be a flag on `submodel_instance` because of the way the collision code works.  Not every model has a model instance (and submodel instances), although ships do.  More importantly, if I understand the collision code correctly, submodel instances are only processed if the submodel is actually moving.

The reason for allowing the flag to be set is to give scripts more control over collision detection.  The idea is to let the script toggle the flag, do a collision check, and then put the flag back the way it was.  For Event Horizon, this would allow flying ships to collide with the ground, while preventing land-based ships from colliding with it.  (I ended up using the +Override mechanism instead.)

Baezon: It is reassuring to know there are 'safe' uses, and that *you* intend to use it safely, but the flag is not required to be used that way, and so remains something of a footgun. A footgun is not called that way because *there merely exist* safe uses of it, but conversely because *there merely exist* unsafe ways to use it.

Calling it a footgun or unsafe might be slightly hyperbolic, but I'm still not entirely convinced the utility outweighs the hassle of getting it to work right in what the user might otherwise see as a simple use case, given how much of FSO internals with regards to models it implicitly exposes. Escpecially considering, again, it would apply to *all* uses of the model, *even entirely different ship classes*.

I don't want to embroil you in a fight over this forever, so it might be worth getting other reviewers to chime in, If others disagree, then it is what it is.

Goober5000: Well Lafiel already said separately that he broadly agrees with your comments.  I'll remove the ability to set the flag.  It's no longer necessary for my use case, and I would hope that other use cases could similarly find an alternate way to accomplish what they need.

PR #6036, opened 2024-03-11
remove Docker setup for standalone servers

notimaginative:
This isn't working properly and just fails every time it tries to build. As no one appears to actually use or want this it's been deemed not worth fixing.


notimaginative:
For any future dev that might wish to revisit this... the issue is believed to be due to changes made to Alpine Linux (specifically, libudev). Updating the FSO build setup to support those changes may fix the issue. Alternatively, switching the base image to Ubuntu or Debian instead may be all that is required to get it working once more.

However neither option was considered to be worth the effort.

Goober5000 (approved):

PR #6039, opened 2024-03-13
add percent-ships-scanned

Goober5000:
Kind of surprised this SEXP operator hasn't been added before now.  It complements the existing percent-X operators in the Objectives category.

Also, this fixes and clarifies the help documentation for the various percent-X operators.  Additionally, the percent operators now correctly return KNOWN_FALSE when the false state is known.


wookieejedi (approved):
Straightforward and looks good to me, also will be quite useful

PR #6040, opened 2024-03-14
Default post processing to on because so many things require it

MjnMixael:


wookieejedi (approved):

JohnAFernandez (approved):

JohnAFernandez:
Good thing someone ran into this issue relatively quickly after it was implemented

PR #6041, opened 2024-03-15
improve set-support-ship description

Goober5000:
Per Durandal, a few attributes of this SEXP can be documented a little better.


wookieejedi (approved):
Three lines of description updates look good, but PR title might want to be fixed (ship instead of script :) )

Goober5000:
Whoops :)

PR #6045, opened 2024-03-16
Allow player-specific options even when using In-Game Options

MjnMixael:
This PR moves the methods to set the serialized values for options into OptionsManager for more global use. Then it removes the check for `Using_in_game_options` from csg.cpp and plr.cpp and instead always loads the player or campaign options and then runs those OptionsManager methods to make sure the serialized values are kept in sync for the UI.

I know @notimaginative was concerned about multi options, but the only one I could definitively find that's saved to the player file was the Local Broadcast setting. Let me know if I missed one, though, and I can add it here.

### DIFF
--- a/code/menuui/optionsmenu.cpp
+++ b/code/menuui/optionsmenu.cpp
@@ -709,6 +709,9 @@ void options_change_gamma(float delta)
 
 	os_config_write_string( NULL, NOX("GammaD3D"), tmp_gamma_string );
 
+	// The Gamma option sets its display value differently to the serialized value itself
+	// so we'll leave this here instead of trying to make a global method that works just
+	// for this one specific case
 	if (Using_in_game_options) {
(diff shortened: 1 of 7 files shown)


notimaginative:
I haven't looked this over yet so some of this may be covered already...

`obj_update_level` is pilot specific since you may want it different between LAN and online play.

PXO is controlled by the `protocol` option. It will be `NET_TCP` for strictly LAN & Direct TCP/IP play, or `NET_VMT` to use PXO instead.

MjnMixael:
Failed on the player file test is interesting. I'm pretty certain

> I haven't looked this over yet so some of this may be covered already...
> 
> `obj_update_level` is pilot specific since you may want it different between LAN and online play.
> 
> PXO is controlled by the `protocol` option. It will be `NET_TCP` for strictly LAN & Direct TCP/IP play, or `NET_VMT` to use PXO instead.

I don't think any of those three are currently covered by OptionsManager, so we should be good then.

BMagnu (approved):
This should be fine as far as I can tell.

PR #6047, opened 2024-03-16
Add AI Flag to `align to target when guarding stationary ship`

wookieejedi:
When a ship is ordered to guard another ship that is stationary and not incredibly large, the AI will fly to a position then turn and face away from the targeted ship. This behavior is not ideal in various scenarios, including both blockade style formations and overall visuals.

The PR adds a flag that tells the AI to align to the target ship it is guarding once the AI has reached it's guarding position in this situation, which allows all the guarding ships to have the same orientation and look much better and work better for directional defenses.

Tested and works as expected. Also changed some `NULL` to `nullptr` in this function for cleanup.


Goober5000 (approved):
Huh, as simple as that?

PR #5894, opened 2023-12-21
Coverity: Change many auto calls to const auto& to avoid unnecessary copies.

JohnAFernandez:
Technically these are all optimizations.  Some of them might even have the tiniest impact.

There are also a couple of very small refactors that also avoid unnecessary copies instead of usting auto &.

PR #6060, opened 2024-03-23
fix crash caused by using a moved variable

Goober5000:
In #5894, in the `add_dynamic_sexp` function in sexp_lookup.cpp, a call to `sexp->getHelpText()` was moved to the bottom of the function.  Normally this would be fine, except that this occasion caused the `sexp` variable to be accessed after a `std::move`.  The crash can be avoided by moving the last access of the `sexp` variable just a little bit earlier.


JohnAFernandez (approved):

JohnAFernandez:
Just noting this is not usable in the point release as it fixes code added after 24.0 stable.

Issue #5959, opened 2024-01-27
FR: full sexp 'boolean' return from lua


EatThePath:
I have on occasion wanted to have a lua sexp return the values like SEXP_KNOWN_FALSE to mesh a little more closely with the event evaluation system. An enum for this or new userdata type seems like the cleanest way to do this.

EatThePath:
After a little more poking at the idea that prompted this, it would likely need to include a new value $Return Type: in the sexp table, that would let the operators using it be used in the root position of an event.

Goober5000:
Sounds about right.  The return type could use an enum class with values corresponding to the `SEXP_` boolean values, and then these could be converted in the `LuaSEXP::getSexpReturnValue` function.

naomimyselfandi:
`lua-save-file-exists` would benefit greatly from this: currently, using it in the obvious way, i.e. `( when ( lua-save-file-exists "whatever" ) ( do-something-related-to-save-files ) )`, causes disk IO every frame if the file doesn't exist.

Goober5000:
I recently added a caching mechanism to alleviate this problem: https://github.com/FSO-Scripters/fso-scripts/pull/59

But this is not to take away from your comment, which is an excellent use case for this FR.

Goober5000:
Turns out that this can be done in a way that's fully compatible with the current API.  PR shortly.

Goober5000:
Implemented in #6063

PR #6063, opened 2024-03-24
allow boolean and numeric scripted SEXPs to return special values

Goober5000:
This adds the special SEXP return values (SEXP_KNOWN_TRUE, SEXP_NAN, and so forth) as enums, and adds code to intercept these enums if they are returned where the code would normally expect a boolean or number.  If an enum is found, it is converted to its special return value equivalent.

With this change, a script can end a function with `return SEXP_KNOWN_TRUE` or similar.  No other API changes are needed; in particular, no special `$Return Type:`s are needed.

Implements #5959.


wookieejedi (approved):
Straightforward, useful, and make sense

PR #6068, opened 2024-03-25
Multi Start Game UI API

MjnMixael:


JohnAFernandez (approved):
Straightforward once you turn off whitespace changes.

code/scripting/api/libs/ui.cpp
```diff
@@ -2805,5 +2807,110 @@ ADE_FUNC(getHelpText, l_UserInterface_MultiPXO, nullptr, "Gets the help text lin
...
+ADE_FUNC(runNetwork, l_UserInterface_MultiStartGame, nullptr, "Runs the network required commands to update the status text", nullptr, nullptr)
+{
+	SCP_UNUSED(L);
+
+	multi_start_game_do(true);
+
+	return ADE_RETURN_NIL;
+}
+
+ADE_FUNC(setName, l_UserInterface_MultiStartGame, "string Password", "Sets the game's name", "boolean", "True if successful, false otherwise")
```
notimaginative, line 2850: "Password"? Guessing it should be "Name" or similar.

notimaginative (changes requested):

code/scripting/api/libs/ui.cpp
MjnMixael: Good catch. Fixed.

notimaginative (approved):

PR #6069, opened 2024-03-26
Add bad-rearm-time sexp

Baezon:
What it says on the tin. AI won't rearm at all while its a bad time.


Goober5000 (approved):
Looks good, and is a reasonable extrapolation from the current logic.

PR #6071, opened 2024-03-26
adjust ship registry checks to accommodate support ships

Goober5000:
This does two things:

1. Fix a bug in `mission_bring_in_support_ship` that assigns an incorrect parse object index to the support ship's ship registry entry.  Since `Arriving_support_ship` is not an item of `Parse_objects`, it should not have an index.
2. Adjust the in-game join checks at the bottom of `parse_create_object_sub` to more scrupulously account for index mismatches and bounds errors, with a specific exception for the arriving support ship.


wookieejedi (approved):

PR #6072, opened 2024-03-28
Add 'alignment' for poofs

Baezon:
For now just vertically, its fairly specific for the moment, but if you use an animated graphic with a rain or snow texture, poofs becomes suitable for basic weather effects. The door is left open for other kinds of alignment behavior as well.

### DIFF
--- a/code/nebula/neb.cpp
+++ b/code/nebula/neb.cpp
@@ -840,13 +840,13 @@ void new_poof(size_t poof_info_idx, vec3d* pos) {
 	new_poof.flash = 0;
 	new_poof.radius = pinfo->scale.next();
 	new_poof.pt = *pos;
+	new_poof.rot_speed = fl_radians(pinfo->rotation.next());
 	new_poof.alpha = pinfo->alpha.next();
 	new_poof.anim_time = frand_range(0.0f, pinfo->bitmap.total_time);
 	if (pinfo->alignment != vmd_zero_vector)
 		vm_vec_rand_vec(&new_poof.up_vec);
 	else
 		new_poof.up_vec = vmd_y_vector;
-	new_poof.rot_speed = fl_radians(pinfo->rotation.next());
 
 	Neb2_poofs.push_back(new_poof);
 }
--- a/code/nebula/neb.cpp
+++ b/code/nebula/neb.cpp
@@ -288,6 +288,16 @@ void parse_nebula_table(const char* filename)
 					poofp->density = 1 / (poofp->density * poofp->density * poofp->density);
 				}
 
+				if (optional_string("$Alignment:")) {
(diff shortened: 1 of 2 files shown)


code/nebula/neb.cpp
```diff
@@ -288,6 +288,16 @@ void parse_nebula_table(const char* filename)
 					poofp->density = 1 / (poofp->density * poofp->density * poofp->density);
 				}
 
+				if (optional_string("$Alignment:")) {
+					SCP_string type;
+					stuff_string(type, F_NAME);
+
+					if (!stricmp(type.c_str(), " VERTICAL"))
```
Goober5000, line 295: The `stuff_string` function should skip over any whitespace so the leading space before VERTICAL wouldn't be detected.

code/nebula/neb.cpp
```diff
@@ -833,7 +843,10 @@ void new_poof(size_t poof_info_idx, vec3d* pos) {
 	new_poof.rot_speed = fl_radians(pinfo->rotation.next());
 	new_poof.alpha = pinfo->alpha.next();
 	new_poof.anim_time = frand_range(0.0f, pinfo->bitmap.total_time);
-	vm_vec_rand_vec(&new_poof.up_vec);
+	if (pinfo->alignment != vmd_zero_vector)
+		vm_vec_rand_vec(&new_poof.up_vec);
+	else
+		new_poof.up_vec = vmd_y_vector;
```
Goober5000, line 849: Should this be `new_poof.up_vec = pinfo->alignment;`?

code/nebula/neb.cpp
```diff
@@ -974,10 +987,16 @@ void neb2_render_poofs()
...
+			if (pinfo->alignment != vmd_zero_vector)
+				scalar = 0.0f;
 
 			vm_vec_scale_add(&view_pos, &eye_pos, &eye_orient.vec.fvec, scalar);
 
 			view_pos -= pf.pt;
+
+			if (pinfo->alignment != vmd_zero_vector)
+				vm_project_point_onto_plane(&view_pos, &view_pos, &pinfo->alignment, &vmd_zero_vector);
+
```
Goober5000, line 999: Is there a distinction between `pinfo->alignment` and `pf.up_vec` here?

Baezon: Yes, the `up_vec` depends on the specific poof in question, which can rotate over time on its own, specified by $Rotation. I don't really think it would look very good, but I tried to make sure rotation could at least still work in some form even if alignment is being used.

Goober5000: Fair enough.  Just wanted to make sure they were both used the way they should be.

Goober5000 (approved):

PR #6073, opened 2024-03-30
Multi host settings API

MjnMixael:
Setting as draft for now because I still need to wire up and test several of these methods.. but if someone more familiar with the multi code wanted to give it a preliminary look, that'd be helpful.


MjnMixael:
Everything's tested and seems to work. Ready for review.

notimaginative:
I should have probably done the reviews in the correct order, but I didn't. So I reviewed #6077 first and it just has most the same stuff here. Same issues too. So would you prefer that I redo those comments here, or can you sort that out?

MjnMixael:
> I should have probably done the reviews in the correct order, but I didn't. So I reviewed #6077 first and it just has most the same stuff here. Same issues too. So would you prefer that I redo those comments here, or can you sort that out?

No worries. I can sort that out.

MjnMixael:
#6077 is identical to this PR but with some additional API code for Multi Sync so all of those requested changes have been applied here as well.

MjnMixael:
#6077 includes all of this PR and then some so I'm just gonna close this one

PR #6078, opened 2024-04-02
Multi general api

MjnMixael:
This PR reorganizes the API slightly, putting them roughly in top->bottom order as you'd progress through the UIs. It also moves Netgame, NetPlayer, Status Text, and Info Text up into the newer MultiGeneral category since they apply to more places than just MultiHostSetup or MultiJoinGame. While doing so it implements some additional validation for if those are accessed outside of proper Multi game states.

Finally this PR also implements `MultiGeneral.getChat()` and `.sendChat()`.

Be aware that for reviewing purposes, ui.cpp looks like a large diff but is actually nearly unchanged with the exception of the added code for MultiGeneral's getChat(), sendChat(), quitGame() and setPlayerState(). The rest of the changes that github is showing is just re-organized but unchanged code.

### DIFF
--- a/code/missionui/chatbox.cpp
+++ b/code/missionui/chatbox.cpp
@@ -403,7 +403,7 @@ size_t chatbox_get_msg_target_length(char *msg)
 }
 
 // send a msg to the game chat
-void chatbox_send_msg(char msg[150])
+void chatbox_send_msg(char* msg)
 {
 	// check if this message is supposed to have a recipient
 	int target = chatbox_get_msg_target_type(msg);
@@ -421,7 +421,7 @@ void chatbox_send_msg(char msg[150])
 		send_game_chat_packet(Net_player, msg, target);
 	} else {
 		// copy the name of the player the message is being sent to
-		char temp[150];
+		char temp[CHATBOX_MAX_LEN];
 		strncpy(temp, msg + 1, target_length - 2);
 		temp[target_length - 2] = '\0';
 		send_game_chat_packet(Net_player, msg, target, nullptr, temp);
@@ -433,11 +433,11 @@ void chatbox_send_msg(char msg[150])
 // automatically split up any input text, send it, and leave the remainder 
 void chatbox_autosplit_line()
 {
-	char msg[150];
+	char msg[CHATBOX_MAX_LEN];
 	
 	// if the chat line is getting too long, fire off the message, putting the last
 	// word on the next input line.
-	memset(msg,0,150);
+	memset(msg, 0, CHATBOX_MAX_LEN);
 	Chat_inputbox.get_text(msg);
 	const char* remainder = ""; 
 
@@ -463,7 +463,7 @@ void chatbox_autosplit_line()
 
 		if (target != MULTI_MSG_ALL) {
 			// we need to add the target the message is going to before we add the rest of the string
-			char temp[150];
+			char temp[CHATBOX_MAX_LEN];
 			strncpy(temp, msg, target_length);
  			temp[target_length] = ' '; 
  			temp[target_length+1] = '\0'; 
--- a/code/missionui/chatbox.cpp
+++ b/code/missionui/chatbox.cpp
@@ -407,7 +407,6 @@ void chatbox_send_msg(char* msg)
 {
 	// check if this message is supposed to have a recipient
 	int target = chatbox_get_msg_target_type(msg);
-	size_t target_length = chatbox_get_msg_target_length(msg);
 
 	// tack on the null terminator in the boundary case
 	size_t x = strlen(msg);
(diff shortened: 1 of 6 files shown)


MjnMixael:
EDIT: I added a couple more general methods I need here. quitGame and setPlayerState

code/missionui/chatbox.cpp
```diff
@@ -402,28 +402,49 @@ size_t chatbox_get_msg_target_length(char *msg)
...
+		send_game_chat_packet(Net_player, msg, target, nullptr, temp);
+	}
+
+	chatbox_add_line(msg, MY_NET_PLAYER_NUM);
+}
+
 // automatically split up any input text, send it, and leave the remainder 
 void chatbox_autosplit_line()
 {
 	char msg[150];
```
notimaginative, line 436: Going ahead and changing this to `char msg[CHATBOX_MAX_LEN]` would be a nice thing to do while your here. Looks like there is only the two places left to change the magic number.

code/missionui/chatbox.cpp
```diff
@@ -433,22 +454,16 @@ void chatbox_autosplit_line()
...
+		// check if this message is supposed to have a recipient
+		int target = chatbox_get_msg_target_type(msg);
+		size_t target_length = chatbox_get_msg_target_length(msg);
 
 		if (target != MULTI_MSG_ALL) {
 			// we need to add the target the message is going to before we add the rest of the string
+			char temp[150];
 			strncpy(temp, msg, target_length);
  			temp[target_length] = ' '; 
  			temp[target_length+1] = '\0';
```
notimaginative, line 469: As mentioned in the earlier change request, the size of `temp` should be de-magic-numbered here, as well as moving `target_length`.

code/missionui/chatbox.h
```diff
@@ -59,6 +59,9 @@ int chatbox_scroll_down();
 // clear the contents of the chatbox
 void chatbox_clear();
 
+// send a message to the chatbox
+void chatbox_send_msg(char msg[150]);
```
notimaginative, line 63: As mentioned earlier, this should be a ptr to `msg`.

code/scripting/api/libs/ui.cpp
```diff
@@ -3200,135 +3283,255 @@ ADE_FUNC(setGameType,
...
-ADE_LIB_DERIV(l_Active_Games, "ActiveGames", nullptr, nullptr, l_UserInterface_MultiJoinGame);
-ADE_INDEXER(l_Active_Games,
-	"number Index",
-	"Array of active games",
-	"active_game",
-	"active game handle, or invalid handle if index is invalid")
+ADE_FUNC(runNetwork,
+	l_UserInterface_MultiSync,
+	nullptr,
+	"Runs the network required commands to update the lists and run the chat",
```
notimaginative, line 3502: This only covers multi sync, correct? So not list updates and chat.

notimaginative (changes requested):

code/scripting/api/objs/multi_objects.cpp
```diff
@@ -113,6 +118,15 @@ netgame_info* net_game_h::getNetgame() const
 	return &Netgame;
 }
 
+bool net_game_h::isValid() const
+{
+	if (!(Game_mode & GM_MULTIPLAYER) && (Netgame.name[0] != '\0')) {
```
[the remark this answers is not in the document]
MjnMixael: I could have sworn I tested that.. but you're right. That is an error.

code/missionui/chatbox.h
MjnMixael: char* and CHATBOX_MAX_LEN all handled, I think.

code/scripting/api/libs/ui.cpp
MjnMixael: Yup. Good catch.

code/missionui/chatbox.cpp
```diff
@@ -402,28 +402,49 @@ size_t chatbox_get_msg_target_length(char *msg)
...
+	// if I'm the server, then broadcast the packet
+	chatbox_recall_add(msg);
+	if (target != MULTI_MSG_EXPR) {
+		send_game_chat_packet(Net_player, msg, target);
+	} else {
+		// copy the name of the player the message is being sent to
+		char temp[150];
+		strncpy(temp, msg + 1, target_length - 2);
+		temp[target_length - 2] = '\0';
+		send_game_chat_packet(Net_player, msg, target, nullptr, temp);
```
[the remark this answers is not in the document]
MjnMixael: OK, target_length has been moved, too.

notimaginative (approved):

PR #6080, opened 2024-04-04
more electrical arc API features

Goober5000:
Follow-up to #6034.  This slightly changes the function signature of `AddElectricArc` but it should be fine since it hasn't appeared in a stable release yet.


Baezon:
I'm a little unsure about this unless we're going to go all the way and actually expose information about the other arc types. It's a little weird to be able to modify arcs you didn't create, while also not really being able to poll any information about them, even if I know you intend to only modify ones the scripter created.

Baezon (approved):
Alright, it's not a big deal I guess, but I do dislike the feeling some change or addition has been made with the very specific author's use case in mind and none others.

PR #6086, opened 2024-04-05
Allow for more field of fire cooldown shenanigans

Baezon:
This allows a scripter to override the 'fof cooldown' i.e. where in between its regular fof and its max fof it is at the moment, this way they can make it be based on whatever, instead of just spread and reset rate. As part of this +Max Fof no longer requires spread rate, since it's harmless to provided a max fof but no engine builtin way for it to reach that max.

### DIFF
--- a/code/scripting/api/objs/ship_bank.cpp
+++ b/code/scripting/api/objs/ship_bank.cpp
@@ -395,12 +395,6 @@ ADE_VIRTVAR(FOFCooldown, l_WeaponBank, "number", "The FOF cooldown value. A valu
 	if(!bh->isValid())
 		return ade_set_error(L, "f", -1.f);
 
-	if (ADE_SETTING_VAR) {
-		CLAMP(newValue, 0.0f, 1.0f);
-		bh->typeh.sw->primary_bank_fof_cooldown[bh->bank] = newValue;
-		return ade_set_args(L, "f", bh->typeh.sw->primary_bank_fof_cooldown[bh->bank]);
-	}
-
 	switch(bh->typeh.type)
 	{
 		case SWH_PRIMARY:
@@ -409,6 +403,13 @@ ADE_VIRTVAR(FOFCooldown, l_WeaponBank, "number", "The FOF cooldown value. A valu
 			float reset_amount = (timestamp_until(bh->typeh.sw->last_primary_fire_stamp[bh->bank]) / 1000.0f) * wif->fof_reset_rate;
 			auto val = bh->typeh.sw->primary_bank_fof_cooldown[bh->bank] + reset_amount;
 			CLAMP(val, 0.0f, 1.0f);
+
+			if (ADE_SETTING_VAR) {
+				CLAMP(newValue, 0.0f, 1.0f);
+				bh->typeh.sw->primary_bank_fof_cooldown[bh->bank] = newValue;
+				val = newValue;
+			}
+
 			return ade_set_args(L, "f", val);
 		}
 		case SWH_SECONDARY:
--- a/code/scripting/api/objs/ship_bank.cpp
+++ b/code/scripting/api/objs/ship_bank.cpp
@@ -399,8 +399,6 @@ ADE_VIRTVAR(FOFCooldown, l_WeaponBank, "number", "The FOF cooldown value. A valu
 	{
 		case SWH_PRIMARY:
 		{
-			auto val = bh->typeh.sw->primary_bank_fof_cooldown[bh->bank];
-
 			if (ADE_SETTING_VAR) {
 				CLAMP(newValue, 0.0f, 1.0f);
 				bh->typeh.sw->primary_bank_fof_cooldown[bh->bank] = newValue;
--- a/code/scripting/api/objs/ship_bank.cpp
+++ b/code/scripting/api/objs/ship_bank.cpp
@@ -399,18 +399,14 @@ ADE_VIRTVAR(FOFCooldown, l_WeaponBank, "number", "The FOF cooldown value. A valu
 	{
 		case SWH_PRIMARY:
 		{
-			auto wif = &Weapon_info[bh->typeh.sw->primary_bank_weapons[bh->bank]];
-			float reset_amount = (timestamp_until(bh->typeh.sw->last_primary_fire_stamp[bh->bank]) / 1000.0f) * wif->fof_reset_rate;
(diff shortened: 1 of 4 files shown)


JohnAFernandez (approved):
LGTM

code/scripting/api/objs/ship_bank.cpp
```diff
@@ -389,15 +389,16 @@ ADE_VIRTVAR(FOFCooldown, l_WeaponBank, "number", "The FOF cooldown value. A valu
...
+	if(!ade_get_args(L, "o|f", l_WeaponBank.GetPtr(&bh), &newValue))
 		return ade_set_error(L, "f", -1.f);
 
 	if(!bh->isValid())
 		return ade_set_error(L, "f", -1.f);
 
 	if (ADE_SETTING_VAR) {
-		LuaError(L, "This function does not support setting values yet!");
-		return ade_set_error(L, "f", -1.f);
+		CLAMP(newValue, 0.0f, 1.0f);
```
Goober5000, line 399: This code should raise an error for secondary and tertiary blocks, I should think.

In fact you might be able to move the ADE_SETTING_VAR code to the top of the SWH_PRIMARY block just below.

code/scripting/api/objs/weaponclass.cpp
```diff
@@ -712,6 +712,36 @@ ADE_VIRTVAR(BurstDelay, l_Weaponclass, "number", "The time in seconds between sh
 	return ade_set_args(L, "f", Weapon_info[idx].burst_delay);
 }
 
+ADE_VIRTVAR(FieldOfFire, l_Weaponclass, "number", "The angular spread for shots of this weapon.", "number", "Fof in degrees, or 0 if handle is invalid")
```
Goober5000, line 715: Would be good to specify that these two virtvars are in radians

Goober5000 (changes requested):

code/scripting/api/objs/weaponclass.cpp
Baezon: They do both specify their units are degrees, if that's what you mean.

Goober5000: *surprised pikachu face*

You're right, they do.  Not only that, the angle is treated as degrees in the call to `vm_vec_random_cone`.  Ok then.

Although this does bring to light an unrelated bug.  I'll submit a PR momentarily.

code/scripting/api/objs/ship_bank.cpp
```diff
@@ -409,6 +403,13 @@ ADE_VIRTVAR(FOFCooldown, l_WeaponBank, "number", "The FOF cooldown value. A valu
 			float reset_amount = (timestamp_until(bh->typeh.sw->last_primary_fire_stamp[bh->bank]) / 1000.0f) * wif->fof_reset_rate;
 			auto val = bh->typeh.sw->primary_bank_fof_cooldown[bh->bank] + reset_amount;
 			CLAMP(val, 0.0f, 1.0f);
+
+			if (ADE_SETTING_VAR) {
+				CLAMP(newValue, 0.0f, 1.0f);
+				bh->typeh.sw->primary_bank_fof_cooldown[bh->bank] = newValue;
```
Goober5000, line 409: This should go at the top of the block because that way the code will properly handle the `reset_amount`.  Assigning `val = newValue` is then not needed.

Baezon: I'm not sure I follow, if it's being set then the reset_amount doesn't matter, it's going to be overriden anyway.

Goober5000: Oh, so the value that is being set should be the grand total, without adding `reset_amount`?  Ok, that makes sense.

Although `reset_amount` will be added the very next time the value is retrieved.  Should the function account for that, perhaps by setting `newValue - reset_amount`?

Goober5000: Okay

Baezon:
Refactored a bit to handle fof cooldown per frame instead of 'on demand' and tested and works now with constant reading of the value but 'intermittent' setting of the value (which did not work properly before).

code/ship/ship.cpp
```diff
@@ -10448,6 +10448,19 @@ void ship_process_post(object * obj, float frametime)
...
 
+	// update fof cooldowns
+	for (int i = 0; i < shipp->weapons.num_primary_banks; i++) {
+		if (shipp->weapons.primary_bank_weapons[i] < 0)
+			continue;
+
+		weapon_info* wip = &Weapon_info[shipp->weapons.primary_bank_weapons[i]];
+		
+		if (wip->fof_reset_rate > 0.0f) {
+			shipp->weapons.primary_bank_fof_cooldown[i] -= wip->fof_reset_rate * flFrametime;
```
Goober5000, line 10459: Minor quibble... since `frametime` is passed as an argument to this function, it should be used in preference to `flFrametime`

code/weapon/weapons.cpp
```diff
@@ -3449,23 +3449,29 @@ int parse_weapon(int subtype, bool replace, const char *filename)
...
-					wip->max_fof_spread = 0.0f;
-				}
+			if (wip->max_fof_spread <= 0.0f) {
+				Warning(LOCATION, "WARNING: +Max FOF must be at least as big as $FOF for '%s'! Defaulting to match $FOF, no spread will occur!", wip->name);
+				wip->max_fof_spread = 0.0f;
 			}
 		}
+
+		if (max_required && wip->max_fof_spread <= 0.0f) {
+			Error(LOCATION, "ERROR: +Max FOF for '%s' must be used if +FOF Spread Rate: is used!", wip->name);
```
Goober5000, line 3473: Instead of making this an `Error`, what about making it a `Warning` and setting the values back to factory defaults?

Also, use `error_display` (0 for warning, 1 for error) so that the message displays the parsed file and parsed line number.

Goober5000 (commented):
Some additional things

code/weapon/weapons.cpp
Baezon: The only reason it's an error is because it was strictly required by the previous parsing, for posterity.

Goober5000 (approved):
👍

code/weapon/weapons.cpp
Goober5000: Fair... the comment could be directed at the previous parsing code as well.

PR #6089, opened 2024-04-08
fix issues with get_text() and set_text() for INPUTBOX

notimaginative:
Fix multiple buffer out-of-bounds issues with INPUTBOX text. Change get_text() to use safer calling methods and use strncpy() in set_text() to avoid issue with incoming text being longer than what is allocated for the control.


MjnMixael (approved):
Looks good and worked in tests.

PR #6090, opened 2024-04-08
Misc Methods for SCPUI Multi

MjnMixael:
This PR does two core things for SCPUI Multi. 

First it properly commits to the mission using the multi methods. In doing this I changed the ui.Commit method to return an enum instead of an int and I removed the boolean gates for the popups because there's a proper API for popups now. Technically this is a breaking change but I'm under the impression that SCP prefers, if the userbase for lua methods is small, to fix the method as cleanly as possible. I'd be shocked if anyone was working on a Lua UI overhaul other than me. But if a full deprecation or alternate method is preferred then I can do that instead.

Second it adds the ability to get and set Netgame.Locked. Fairly straightforward.


notimaginative:
Looks okay to me but my brain is fried at the moment so I could have missed something. Might be good to have @wookieejedi or @JohnAFernandez give it a look when time allows.

If no one else has approved it by Monday night I'll give it another look.

notimaginative (approved):

PR #6092, opened 2024-04-10
Check if inputbox is valid

MjnMixael:
The recent fixes to InputBox handling exposed a different issue with SCPUI and logging into PXO. During the login process the method `multi_pxo_process_common()` is run, but since it's run by FSO and not the LUA API it correctly has `api_access` set to false. Thus is also runs `multi_pxo_chat_process()` which expects the chat InputBox to have non null data.

However since SCPUI has already overridden the ui creation at this point the InputBox does not get created. I thought through a number of ways to fix this, but in the end I decided a simple method to test if an InputBox is valid was the simplest.

### DIFF
--- a/code/network/multi_pxo.cpp
+++ b/code/network/multi_pxo.cpp
@@ -3202,6 +3202,11 @@ void multi_pxo_chat_process()
 	char msg[512];
 	int msg_pixel_width;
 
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):
LGTM

notimaginative (approved):

Issue #6076, opened 2024-04-01
Hide key-reset SEXP from FRED


jlberman:
There is no purpose to key-reset since there's key-reset-multiple.  I get that legacy campaigns still might use it, but it should be hidden from new content.

Goober5000:
The `key-reset` operator very specifically and intentionally works once regardless of the number of repeat iterations.  This is because the training missions have a certain set of operators (messages and Instructor ship goals) that repeat until the player has pressed a certain key, but the key must not be reset for every repeat iteration.

The question is whether modern campaigns would ever use such a technique.  I tend to doubt it, but it's an open question.

We could simply hide the `key-reset` operator as you suggest and force all FREDders to either a) manually enter that operator into the event, or b) implement the behavior using a different technique.

jlberman:
I like option A.  If you want to use key-reset, you're doing it very deliberately.

Goober5000:
Implemented in #6094

PR #6094, opened 2024-04-11
hide key-reset

Goober5000:
Hides the `key-reset` operator in FRED and elaborates on the help descriptions for both it and `key-reset-multiple`.  Implements #6076.


wookieejedi (approved):
Good doc changes, and hiding reset makes sense to me, but happy to hear any other opinions.

PR #6096, opened 2024-04-11
Team Loadout API for SCPUI

MjnMixael:
Adds everything that should be needed for multiplayer loadout selection. Adds the ship/weapon packet sending and disallows directly setting wss slots if in multiplayer mode.

Also fixes a few small issues and adds stopMainhallMusic because in multi it needs to be stopped manually.

### DIFF
--- a/code/network/multiteamselect.cpp
+++ b/code/network/multiteamselect.cpp
@@ -348,10 +348,6 @@ int Multi_ts_hotspot_index = -1;
 #define TS_SWAP_PLAYER_PLAYER								4
 #define TS_MOVE_PLAYER										5
 
-// packet codes
-#define TS_CODE_LOCK_TEAM									0						// the specified team's slots are locked
-#define TS_CODE_PLAYER_UPDATE								1						// a player slot update for the specified team
-
 // team data
 #define MULTI_TS_FLAG_NONE									-2						// slot is _always_ empty
 #define MULTI_TS_FLAG_EMPTY								-1						// flag is temporarily empty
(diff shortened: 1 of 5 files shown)


notimaginative (changes requested):

code/network/multiteamselect.cpp
```diff
@@ -1242,12 +1242,31 @@ void multi_ts_blit_wings()
 	}	
 }
 
+bool multi_ts_is_player_in_slot(int slot_idx)
+{
+	Assertion(slot_idx > 0, "Call to check if player is allowed in a negative number slot!");
```
notimaginative, line 1247: Should be `>=` there.

And don't forget the `Net_player != nullptr` check too.

code/network/multiteamselect.cpp
```diff
@@ -1242,12 +1242,31 @@ void multi_ts_blit_wings()
...
+	Assertion(slot_idx < MAX_WSS_SLOTS, "Call to check if player is allowed in slot greater than Max Slots!");
+
+	return (Multi_ts_team[Net_player->p_info.team].multi_ts_player[slot_idx] != nullptr);
+}
+
+bool multi_ts_is_player_allowed_in_slot(int slot_idx)
+{
+	Assertion(slot_idx > 0, "Call to check if player is allowed in a negative number slot!");
```
notimaginative, line 1255: Should be `>=` there.

And don't forget the `Net_player != nullptr` check too.

notimaginative (changes requested):

code/network/multiteamselect.cpp
MjnMixael: Done.

notimaginative (approved):

PR #6098, opened 2024-04-11
Improve Mainhall Button Responsiveness

wookieejedi:
Previously, there was a 175 millisecond delay when hovering over a mainhall animation before it triggered. This PR makes that value settable via the mainhall table.

Tested and works as expected. Now custom mainhalls have much more of a spring int their step.

Thanks Lafiel and MJN!

### DIFF
--- a/code/menuui/mainhallmenu.cpp
+++ b/code/menuui/mainhallmenu.cpp
@@ -2224,6 +2224,9 @@ void parse_main_hall_table(const char* filename)
 
 		if (optional_string("$Num Resolutions:")) {
 			stuff_int(&num_resolutions);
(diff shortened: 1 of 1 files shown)


MjnMixael (approved):

Goober5000 (changes requested):
Can you move the `if (num_resolutions < 1)` check back above line 2229, or even better, inside the "$Num Resolutions:" block?  Since the parsing of the value and its bounds check are closely related, it would be good to keep them together.

wookieejedi:
> if (num_resolutions < 1)

Good point, just updated.

Goober5000 (approved):

PR #6099, opened 2024-04-11
Ingame prejoin API

MjnMixael:
Should be all that's needed to finalize the ingame prejoin ui


code/network/multi_ingame.cpp
```diff
@@ -88,6 +88,8 @@
 LOCAL	int	Ingame_ships_deleted = 0;
 //LOCAL	int	Ingame_ships_to_delete[MAX_SHIPS];
 
+SCP_vector<int> Ingame_ship_choices;
```
notimaginative, line 91: This is going to be a pretty small vector but it's still good to clean up after once it's no longer needed. A `.clear()` call followed by a `.shrink_to_fit()` somewhere that's safe to call from both UI's once the selection screen closes.

I can't tell if `multi_ingame_select_close()` gets called when using the scripted interface, but if so that would be the place to put it.

notimaginative (changes requested):

code/network/multi_ingame.cpp
MjnMixael: `multi_ingame_select_close()` isn't called because it only handles retail UI stuff. But I've added a clear and shrink_to_fit to it and the API close method.

PR #6103, opened 2024-04-16
add 'In back of ship' arrival location

Goober5000:
Allows ships and wings to arrive behind a ship, to complement 'In front of ship'.  In fact, the arrival code is nearly identical, with just a small tweak to distinguish ahead from behind.

Since the new arrival location represents a change in mission format, this also adds code to detect for its usage and set the mission format to 24.1.  Most missions can continue to be saved as version 23.1.

Tested and works as expected.

### DIFF
--- a/code/mission/missionparse.cpp
+++ b/code/mission/missionparse.cpp
@@ -242,7 +242,7 @@ const char *Icon_names[MIN_BRIEF_ICONS] = {
 
 // definitions for arrival locations for ships/wings
 const char *Arrival_location_names[MAX_ARRIVAL_NAMES] = {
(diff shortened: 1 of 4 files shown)


Baezon (commented):
It's not a blocker, but since you're in the area, and it seems like an obvious extension, I wonder if its possible to get top/bottom/left/right too? Would need a little bit different algorithm, I would use `vm_rot_point_around_line` around the relevant axes.

Goober5000:
Yeah, that should be pretty doable.

Goober5000:
Done

Baezon (approved):

PR #6105, opened 2024-04-18
replace pointers with indexes in object_h

Goober5000:
This refactors the object_h struct to store objects as indexes rather than pointers.  This is a prerequisite for making the collection of objects dynamic in the future.  See also #6100.

Follow-up to #5800 and #5829.

~~In draft pending tests.~~  Tested and ready for review.


wookieejedi (approved):
Tests work for me and overall looks good. Hapy to have another set of eyes on it, too

PR #6107, opened 2024-04-18
add API for glow point banks

Goober5000:
Previously, a model API existed for glow points mounted on thrusters, but not glow points mounted on glow point banks.  This adds glow point bank support.  Also, this refactors and cleans up the implementation of the model API a bit.

Depends on #6100; ~~in draft until that is merged.~~

### DIFF
--- a/code/scripting/api/objs/model.cpp
+++ b/code/scripting/api/objs/model.cpp
@@ -74,80 +74,87 @@ ADE_VIRTVAR(Submodels, l_Model, nullptr, "Model submodels", "submodels", "Model 
 		return ade_set_error(L, "o", l_ModelSubmodels.Set(model_h()));
 
 	if (ADE_SETTING_VAR)
-		LuaError(L, "Attempt to use Incomplete Feature: Modelsubmodels copy");
+		LuaError(L, "Assigning submodels is not supported");
 
 	return ade_set_args(L, "o", l_ModelSubmodels.Set(model_h(pm)));
 }
 
 ADE_VIRTVAR(Textures, l_Model, nullptr, "Model textures", "textures", "Model textures, or an invalid textures handle if the model handle is invalid")
 {
-	model_h *mdl = NULL;
-	model_h *oth = NULL;
-	if(!ade_get_args(L, "o|o", l_Model.GetPtr(&mdl), l_ModelTextures.GetPtr(&oth)))
+	model_h *mdl = nullptr;
+	if (!ade_get_args(L, "o", l_Model.GetPtr(&mdl)))
 		return ade_set_error(L, "o", l_ModelTextures.Set(model_h()));
 
 	polymodel *pm = mdl->Get();
-	if(pm == NULL)
+	if (!pm)
 		return ade_set_error(L, "o", l_ModelTextures.Set(model_h()));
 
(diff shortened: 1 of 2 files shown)


code/scripting/api/objs/model.cpp
```diff
@@ -608,8 +614,7 @@ ADE_INDEXER(l_ModelTextures, "texture", "number Index/string TextureName", "text
 		return ade_set_error(L, "o", l_Texture.Set(texture_h()));
 
 	polymodel *pm = mth->Get();
-
-	if (!mth->isValid() || s == NULL || pm == NULL)
+	if (s == NULL || pm == NULL)
```
BMagnu, line 617: if we're touching this, we should change NULL to nullptr.

code/scripting/api/objs/model.cpp
```diff
@@ -737,52 +716,37 @@ ADE_OBJ(l_ModelThrusters, model_h, "thrusters", "The thrusters of a model");
...
 ADE_INDEXER(l_ModelThrusters, "number Index", "Array of all thrusterbanks on this thruster", "thrusterbank", "Handle to the thrusterbank or invalid handle if index is invalid")
 {
-	model_h *trh = NULL;
-	const char* s    = nullptr;
-	thrusterbank_h newThr;
+	model_h *mdl = nullptr;
+	int idx = -1;
 
-	if (!ade_get_args(L, "os|o", l_ModelThrusters.GetPtr(&trh), &s, l_Thrusterbank.Get(&newThr)))
+	if (!ade_get_args(L, "oi", l_ModelThrusters.GetPtr(&mdl), &idx))
```
BMagnu, line 735: Will this potentially break things?

code/scripting/api/objs/model.cpp
```diff
@@ -829,38 +799,27 @@ ADE_FUNC(__len, l_Thrusterbank, NULL, "Number of thrusters on this thrusterbank"
...
-	const char* s       = nullptr;
-	glowpoint_h *glh = NULL;
-
-	if (!ade_get_args(L, "os|o", l_Thrusterbank.GetPtr(&tbh), &s, l_Glowpoint.GetPtr(&glh)))
-		return ade_set_error(L, "o", l_Glowpoint.Set(glowpoint_h()));
+	thrusterbank_h *tbh = nullptr;
+	int idx = -1;
 
-	if (!tbh->isValid() || s==NULL)
+	if (!ade_get_args(L, "oi", l_Thrusterbank.GetPtr(&tbh), &idx))
```
BMagnu, line 805: same here, does this need to keep compat?

BMagnu (changes requested):

code/scripting/api/objs/model.cpp
Goober5000: No, for all practical purposes this has always behaved as a numeric indexer.  The old code did not index by string (e.g. as a name); what it did was load the argument as a string and then use `atoi` to interpret it as a number.  The new code just loads the argument as a number.

code/scripting/api/objs/model.cpp
Goober5000: Same response - this was a numeric indexer before, and it's still a numeric indexer.

BMagnu (approved):

PR #6110, opened 2024-04-19
Account for scripting when dealing with velocity

Baezon:
This adds a flag `PF_SCRIPTED_VELOCITY` whenever velocity for an object is altered via script, which suppresses typical engine calculations for updating velocity for that frame.

If you try to modify velocity via script, it mostly doesn't work. The flight model, in most cases, does not really care too much about previously existing velocity, it's calculated from `prev_ramp_vel` and a host of other things related to the controls/engine of the ship. Depending on when you make the change, certain other things, beyond physical effects, may notice and take into account the change you made, like the HUD reporting speed or something, but either way when it comes time for physics it will be largely ignored.

This makes it very annoying and fiddly to work with, and in general I tend to simply act directly on position for this reason. This PR is intended to address that.

THIS MAY CHANGE SCRIPT BEHAVIOR AS A RESULT. Changes that would not have been reflected before *will* be reflected now. It think that this is not very common, but it is an important thing to do, and ideally *without* any new flags. I would however, be open to using a flag instead if others disagree.

Also: always keep speed and fspeed up to date when changes to velocity are made, it's bad if those disagree, even temporarily.


Goober5000:
> Also: always keep speed and fspeed up to date when changes to velocity are made, it's bad if those disagree, even temporarily.

If that's the case, should setting speed set fspeed and vice versa?

code/physics/physics.cpp
```diff
@@ -378,9 +378,12 @@ void physics_sim_vel(vec3d * position, physics_info * pi, matrix *orient, vec3d*
...
 	*position += grav_disp;
 
-	// update world velocity
-	vm_vec_unrotate(&pi->vel, &local_v_out, orient);
-	pi->vel += grav_vel;
+	// update world velocity	
+	if (!(pi->flags & PF_SCRIPTED_VELOCITY)) {
+		vm_vec_unrotate(&pi->vel, &local_v_out, orient);
+		pi->vel += grav_vel;
+		pi->flags &= ~PF_SCRIPTED_VELOCITY;
```
Goober5000, line 385: Why would this unset the flag immediately after checking that it's unset?

Baezon: Whoops!

Goober5000 (approved):

Baezon:
Actually, something came up something else came up.

Baezon:
In the initial version of this PR, the physics would 'respect' changes to velocity, by not touching it for that frame, but it would still apply the engine's calculated *displacement*. This is still very unintuitive, I think it's reasonable for people to think of velocity as being the change in position over time, so if the displacement is still being applied, again, the velocity just seems like its lying.

I have changed it so that 'scripted velocity' applies to displacement as well. If you change velocity via script, *then that actually dictates the change in position for that frame*. Now it works more like position does, the engine touches it too, but if you override it, it's respected.

Because this even more so could really significantly change behavior where this might've been used before, I bit the bullet and made it a flag.

code/mod_table/mod_table.cpp
```diff
@@ -1372,6 +1373,10 @@ void parse_mod_table(const char *filename)
 				stuff_boolean(&Contrails_use_absolute_speed);
 			}
 
+			if (optional_string("$Scripted velocity:")) {
```
Goober5000, line 1376: This should be a little more verbose - `$Scripted velocity:` by itself is unclear.  Maybe `$Scripted velocity overrides physical velocity:`?

code/scripting/api/objs/physics_info.cpp
```diff
@@ -284,7 +284,7 @@ ADE_VIRTVAR(SlideDecelerationTime, l_Physics, "number", "Time to decelerate from
 	return ade_set_args(L, "f", pih->pi->slide_decel_time_const);
 }
 
-ADE_VIRTVAR(Velocity, l_Physics, "vector", "Object world velocity (World vector)", "vector", "Object velocity, or null vector if handle is invalid")
+ADE_VIRTVAR(Velocity, l_Physics, "vector", "Object world velocity (World vector). Setting this value will have minimal effect unless the $Scriped velocity game settings flag is used.", "vector", "Object velocity, or null vector if handle is invalid")
```
Goober5000, line 287: Typo in `$Scripted`

code/physics/physics.cpp
```diff
@@ -371,16 +371,22 @@ void physics_sim_vel(vec3d * position, physics_info * pi, matrix *orient, vec3d*
...
-	vec3d world_disp;
-	vm_vec_unrotate (&world_disp, &local_disp, orient);
-	*position += world_disp;
-	*position += grav_disp;
-
-	// update world velocity
-	vm_vec_unrotate(&pi->vel, &local_v_out, orient);
-	pi->vel += grav_vel;
+	
+	if (!Scripted_velocity || !(pi->flags & PF_SCRIPTED_VELOCITY)) {
```
Goober5000, line 375: Should this `||` be `&&`?  Or alternatively, remove the `Scripted_velocity` check from here and add it right before `pih->pi->flags |= PF_SCRIPTED_VELOCITY;` in the Lua API.

Goober5000 (changes requested):

code/mod_table/mod_table.cpp
Baezon: Even when I have the words to spare, it's often still too complicated to try and explain exactly how it works, so I'm loathe to introduce more words that I feel like will only confuse further.

Goober5000 (approved):

Issue #5878, opened 2023-12-10
make mission log dynamic


Goober5000:
As described in #5877, very complex missions can overflow the mission log, leading to mysterious problems.  The short-term fix is to simply bump the limit.  The more appropriate long-term fix is to make the mission log dynamic and to remove the culling functions.

Goober5000:
Implemented in #6111

PR #6111, opened 2024-04-23
make mission log dynamic

Goober5000:
Change the mission log from a static array to a dynamic vector.  This removes any chance of the mission log running into a limit (as regularly happened with the last mission of Shepherds before the limit bump) and also prevents strange mission bugs caused by culling obsolete entries.

This also unifies the API a bit by renaming the public `message_log_` function prefixes to `mission_log_`, removing the redundant `Num_log_lines` variable, and removing `Log_scrollback_vec` from the header.

Implements #5878.

### DIFF
--- a/code/mission/missionlog.cpp
+++ b/code/mission/missionlog.cpp
@@ -230,14 +230,6 @@ void mission_log_add_entry(LogType type, const char *pname, const char *sname, i
 	}
 
 #ifndef NDEBUG
-	if ( (Log_entries.size() % 10) == 0 ) {
-		static size_t largest_size_save = Log_entries.size();
-		if ( (Log_entries.size() > 350) && (Log_entries.size() > largest_size_save) ){
-			largest_size_save = Log_entries.size();
-			nprintf(("missionlog", "new highwater point reached for mission log (" SIZE_T_ARG " entries).\n", largest_size_save));
-		}
-	}
-
 	float mission_time = f2fl(Missiontime);
 	int minutes = (int)(mission_time / 60);
 	int seconds = (int)mission_time % 60;
@@ -352,7 +344,7 @@ int mission_log_get_time_indexed( LogType type, const char *pname, const char *s
 				}
 
 				if ( (!stricmp(entry->pname, pname) && !stricmp(entry->sname, sname)) || (!stricmp(entry->pname, sname) && !stricmp(entry->sname, pname)) ) {
-					found = 1;
+					found = true;
 				}
 			} else {
 				// for non dock/undock goals, then the names are important!
@@ -368,11 +360,11 @@ int mission_log_get_time_indexed( LogType type, const char *pname, const char *s
 				// if we are looking for a subsystem entry, the subsystem names must be compared
 				if ((type == LOG_SHIP_SUBSYS_DESTROYED || type == LOG_CAP_SUBSYS_CARGO_REVEALED)) {
 					if ( (sname == NULL) || !subsystem_stricmp(sname, entry->sname) ) {
-						found = 1;
+						found = true;
 					}
 				} else {
 					if ( (sname == NULL) || !stricmp(sname, entry->sname) ) {
-						found = 1;
+						found = true;
 					}
 				}
 			}
--- a/code/mission/missionlog.cpp
+++ b/code/mission/missionlog.cpp
@@ -54,30 +54,30 @@ void mission_log_add_entry(LogType type, const char *pname, const char *sname, i
 	}
 
 	Log_entries.emplace_back();
-	auto entry = &Log_entries.back();
+	auto &entry = Log_entries.back();
 
-	entry->type = type;
+	entry.type = type;
 	if ( pname ) {
 		Assert (strlen(pname) < NAME_LENGTH);
-		strncpy_s(entry->pname, pname, NAME_LENGTH - 1);
+		strncpy_s(entry.pname, pname, NAME_LENGTH - 1);
 	} else
-		strcpy_s( entry->pname, EMPTY_LOG_NAME );
+		strcpy_s( entry.pname, EMPTY_LOG_NAME );
 
 	if ( sname ) {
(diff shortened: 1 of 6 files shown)


code/mission/missionlog.cpp
```diff
@@ -359,15 +226,15 @@ void mission_log_add_entry(LogType type, const char *pname, const char *sname, i
...
-	if ( !(last_entry % 10) ) {
-		if ( (last_entry > LOG_HALFWAY_REPORT_NUM) && (last_entry > last_entry_save) ){
-			nprintf(("missionlog", "new highwater point reached for mission log (%d entries).\n", last_entry));
+	if ( (Log_entries.size() % 10) == 0 ) {
+		static size_t largest_size_save = Log_entries.size();
+		if ( (Log_entries.size() > 350) && (Log_entries.size() > largest_size_save) ){
+			largest_size_save = Log_entries.size();
+			nprintf(("missionlog", "new highwater point reached for mission log (" SIZE_T_ARG " entries).\n", largest_size_save));
 		}
 	}
```
BMagnu, line 239: This seems kinda superflous now.
I bet this was only there to monitor that a safe size for the missionlog was picked.
I'd honestly advocate to just get rid of this wholesale.
If not, I'd at least want a slight tidyup in moving the minimum count to print about this into the static variable, and not in the if clause like this:
```suggestion
	if ( (Log_entries.size() % 10) == 0 ) {
		static size_t largest_size_save = 350;
		if ( Log_entries.size() > largest_size_save ){
			largest_size_save = Log_entries.size();
			nprintf(("missionlog", "new highwater point reached for mission log (" SIZE_T_ARG " entries).\n", largest_size_save));
		}
	}
```

code/mission/missionlog.cpp
```diff
@@ -479,14 +335,11 @@ void mission_log_add_entry_multi( LogType type, const char *pname, const char *s
...
-	log_entry *entry;
 	Assertion(count > 0, "The count parameter is %d; it should be greater than 0!", count);
 
-	entry = &log_entries[0];
-
-	for (i = 0; i < last_entry; i++, entry++) {
-		found = 0;
+	for (const auto &ii: Log_entries) {
+		auto entry = &ii;
+		bool found = false;
```
BMagnu, line 342: Changing found to a bool, I think assignments to found should also assign bool literals instead of ints as it's currently the case.

code/mission/missionlog.cpp
```diff
@@ -46,139 +34,27 @@
...
-	// for a new entry
-	mission_log_obsolete_entries(type, pname);
-
-	entry = &log_entries[last_entry];
-
-	if ( last_entry == MAX_LOG_ENTRIES ){
-		return;
-	}
+	Log_entries.emplace_back();
+	auto entry = &Log_entries.back();
```
BMagnu, line 57: I'd love for this (and other auto-assigned types) to at least be an auto* instead for readability

code/mission/missionlog.cpp
```diff
@@ -438,23 +305,12 @@ void mission_log_add_entry(LogType type, const char *pname, const char *sname, i
...
-
-	entry = &log_entries[last_entry];
-
-	if ( last_entry == MAX_LOG_ENTRIES ){
-		return;
-	}
-
-	last_entry++;
+	Log_entries.emplace_back();
+	auto entry = &Log_entries.back();
```
BMagnu, line 313: same here

code/mission/missionlog.cpp
```diff
@@ -479,14 +335,11 @@ void mission_log_add_entry_multi( LogType type, const char *pname, const char *s
...
-	int i, found;
-	log_entry *entry;
 	Assertion(count > 0, "The count parameter is %d; it should be greater than 0!", count);
 
-	entry = &log_entries[0];
-
-	for (i = 0; i < last_entry; i++, entry++) {
-		found = 0;
+	for (const auto &ii: Log_entries) {
+		auto entry = &ii;
```
BMagnu, line 341: same here

code/mission/missionlog.cpp
```diff
@@ -555,13 +406,10 @@ int mission_log_get_time( LogType type, const char *pname, const char *sname, fi
...
-	int i;
-	log_entry *entry;
-	int count = 0;  
-
-	entry = &log_entries[0];
+	int count = 0;
 
-	for (i = 0; i < last_entry; i++, entry++) {
+	for (const auto& ii : Log_entries) {
+		auto entry = &ii;
```
BMagnu, line 412: same here

code/mission/missionlog.cpp
```diff
@@ -661,28 +509,25 @@ void message_log_add_segs(const char* source_string, int msg_color, int flags =
...
-	mission_log_cull_obsolete_entries();  // compact array so we don't have gaps
 
 	Log_scrollback_vec.clear();
-	Num_log_lines = 0;
 
-	log_entry* entry;
-	for (int i=0; i<last_entry; i++) {
-		entry = &log_entries[i];
+	for (const auto& ii : Log_entries) {
+		auto entry = &ii;
```
BMagnu, line 530: same here

BMagnu (changes requested):
A few small things, otherwise this looks good

code/mission/missionlog.cpp
Goober5000: Yeah, fair point, I'll just delete it.

code/mission/missionlog.cpp
Goober5000: Yeah.  Originally I wanted to minimize the size of the diff but I'll put this in a new commit.

code/mission/missionlog.cpp
Goober5000: Interesting... I'm of the opposite opinion, that using `auto*` is less readable than simply using `auto`.  Unlike `const auto` or `auto &`, `auto *` doesn't change the meaning of the declaration here.

BMagnu (approved):

PR #6112, opened 2024-04-23
fix ballistic primary banks exploding

Goober5000:
When game sounds were changed to allow named entries, it became possible for named entries to unexpectedly replace numbered entries.  This was noticed most obviously in the ballistic rearm sound effect, which started using explosions rather than the standard rearm sound effects.  This occurred because MV_Tiny and MV_Small were added into indexes 200 and 201.

To restore the original behavior, the code now prevents sound effects added via TBM from being placed into reserved indexes.  The reserved indexes are instead assigned placeholders, which can subsequently be updated or left alone as the modder prefers.

This also makes the code a bit more robust including patching some sound validity checks in `gamesnd_game_sound_try_load` and `snd_load`.


wookieejedi (approved):
Works in my tests and looks good to me

PR #6113, opened 2024-04-25
don't short-circuit eval_sexp on any nodes in a when-argument tree

Goober5000:
This is a follow-up to #2942.  That PR solved most of the short-circuiting issues with nested conditionals under when-argument, but it was still possible for the code to incorrectly short-circuit on a nested conditional that had `<argument>` in an action operator but not a condition operator.  The cleanest way to fix this is to not short-circuit on *any* nodes that are part of a when-argument tree.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -1079,6 +1079,7 @@ void sexp_copy_variable_from_index(int node);
 void sexp_copy_variable_between_indexes(int node);
 
 int verify_vector(const char *text);
+bool is_descendant_of_when_argument_op(int node);
 
 
 #define ARG_ITEM_F_DUP	(1<<0)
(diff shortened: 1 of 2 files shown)


Goober5000:
The attached mission demonstrates the problematic behavior.  The "Target And Approach" event sends messages repeatedly whenever the player has targeted one of the transports and is within 200 meters of it.  But if Transport 1 is destroyed or departed, the event will no longer work for Transport 2.  The nested "when" does not contain `<argument>` in the condition and so the code does not prevent short-circuiting.  When this PR is applied, the event works correctly in all situations.

[departure-check.fs2.txt](https://github.com/scp-fs2open/fs2open.github.com/files/15104028/departure-check.fs2.txt)

wookieejedi (approved):
Good catch, and tests seem to work well on my end, too

PR #6114, opened 2024-04-26
make fireball types dynamic

Goober5000:
As requested by Event Horizon, this removes the limit on fireball types.  The fixed array has been replaced with a vector and the relevant code updated accordingly.  There is also a new version of `stuff_int_list()` to parse ints into a vector.

This also adds bounds checking for explosion animations that are parsed for ship classes, ship types, and asteroids, since previously there was no guarantee that the indexes would be in range.

Tested by myself and Ice Cream Man.


wookieejedi (approved):
Good update, and tests work. Also LGTM

Issue #6064, opened 2024-03-24
FR: Settings for movie subtitles


Achille44:
In some cases, the movie subtitles (.srt files) are difficult to read on screen due to the colors/brightness of the video.
It would be nice to have a way to customize how the subtitles are displayed (at least the possibility to change the color of the text and the coordinates where the text appears).
Maybe these settings could be added to game_settings.tbl (same as "$Movie subtitle font")?

Goober5000:
Dark Visor will be taking a look at this.

Goober5000:
Implemented in #6116

PR #6116, opened 2024-04-28
Update Settings for movie subtitles #6064

Dark-Visor:
Update mod_table and movie files to include settings for the color and transparency of the subtitles.

Implements #6064

### DIFF
--- a/code/cutscene/movie.cpp
+++ b/code/cutscene/movie.cpp
@@ -45,6 +45,7 @@ struct PlaybackState {
 	vec2d posBottomRight;
 
 	int subtitle_font = -1;
+	color subtitle_color;
 };
 
 void processEvents()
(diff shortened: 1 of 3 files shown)


Goober5000 (commented):
Will check it out again after tomorrow's changes

MjnMixael:
If this is meant to address visibility issues with specific cutscenes, shouldn't this be a setting in cutscenes.tbl instead of a mod-wide parameter?

Dark-Visor:
> If this is meant to address visibility issues with specific cutscenes, shouldn't this be a setting in cutscenes.tbl instead of a mod-wide parameter?

Thanks MjnMixael, the mod-wide approach seemed the best one for now based on that $Movie subtitle font is already in game_settings.tbl and it made most sense to have them together. Maybe at a later stage both font and color could be added to cutscenes.tbl , potentially along with font position settings as well.

Goober5000 (approved):
This looks good now.  I'll get a test build to @Achille44.

Goober5000 (changes requested):
Build sent.  There is one more usability change we should make; I'll wait for the test results in case there are others.

Goober5000 (approved):
Nice work. 👍

PR #6119, opened 2024-05-01
Add waypoints-no-formation flag

Baezon:
Ships in a wing automatically getting into formation with their wing leader when they share the same waypoint order is nice, but can surprise and frustrate fredders in certain circumstances, #6056 and $fix ai path order bug being good examples of this. They were genuine bugs which deserved fixing anyway, but sometimes its more hassle to the fredder than its worth, so this lets them turn it on and off at will.

### DIFF
--- a/code/ai/ai_flags.h
+++ b/code/ai/ai_flags.h
@@ -32,6 +32,8 @@ namespace AI {
 		Unload_primaries,			//	Fire primaries as fast as possible!
 		Trying_unsuccessfully_to_warp,	// Trying to warp, but can't warp at the moment
 		Free_afterburner_use,		// Use afterburners while following waypoints or flying towards objects
+		Waypoints_no_formation,		// wont get into formation when running waypoints as part of a wing
+
 
 		NUM_VALUES
 	};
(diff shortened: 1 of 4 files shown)


Goober5000 (changes requested):
There isn't really a way to set wing flags unless you are using scripting or you specify the flag in the mission file.  (And in the latter case, you'd need to add FRED support.)  I would recommend taking that out.

This PR also has a ship flag which looks redundant to the AI flag.  It also isn't checked in the `ai_formation()` condition.

Goober5000:
> Wing flags can be changed via alter-wing-flag, that and alter-ship-flag were the main ways I expected this to be used.

Ah, fair enough.  I missed the `alter-wing-flag` sexp.

> Also, where is this redundant flag?

Since both the `alter-ship-flag` sexp and the `ship:setFlag` scripting function check both AI flags and ship flags, there isn't a need to have both flags.

Baezon:
Can you point to a specific line? I added a wing flag and an AI flag, there isn't actually a ship flag.

Goober5000:
Bleh, you're right.  My brain linked ship_flags.h with a ship flag, even though it was a wing flag.

Goober5000 (approved):
Everything looks good 👍

Issue #6118, opened 2024-05-01
FR: after a crash, load most recent autosave


jlberman:
When Fred crashes, I would like a prompt asking whether I want to load the most recent autosave on the next launch.

Goober5000:
Implemented in #6120

PR #6120, opened 2024-05-02
Fred autorecover

Goober5000:
FRED will periodically autosave the mission to a series of Backup.NNN files in %APPDATA%, but this is not widely known, and in any case the user typically must restore the file himself in the event of a crash.  This PR adds code to the File -> Open routine that will automatically check all of the autosaved files whenever the user opens the mission.  If an autosaved file has the same creation date but a newer modification time than the desired file, FRED will prompt the user to open the autosave instead.

This PR adds equivalent code to qtFRED as well, but the autosaving of Backup.NNN files has not been implemented in qtFRED yet.

Implements #6118.

### DIFF
--- a/code/cfile/cfile.h
+++ b/code/cfile/cfile.h
@@ -377,6 +377,7 @@ struct CFileLocation {
 	SCP_string full_name;
 	size_t size          = 0;
 	size_t offset        = 0;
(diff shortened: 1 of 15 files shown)


wookieejedi (approved):

PR #6121, opened 2024-05-02
Fix DRADIS blips

Goober5000:
Several improvements and fixes to the DRADIS radar, one of which (tracking blip expiration) dates back to the original implementation.

1. Convert `sweep_duration` from seconds to milliseconds to work correctly with the timestamp system.  Also clarify use of `game_get_overall_frametime()`, especially with `ba.getFrametimeOverall()`; and add `ba.getSecondsOverall()` to do what `ba.getFrametimeOverall()` allegedly did.
2. Fix a bug where blips expired immediately every frame; they now track their expiration by object
3. Store object numbers rather than object pointers
4. Clamp both the high bound and the low bound of alpha values
5. Change sweep_percent field to sweep_angle to be more accurate
6. Remove some obsolete commented code

Follow-up to #5839.  Fixes blips immediately fading out in Solaris, BTRL, and Diaspora.

### DIFF
--- a/code/radar/radardradis.cpp
+++ b/code/radar/radardradis.cpp
@@ -35,7 +35,7 @@
 
 HudGaugeRadarDradis::HudGaugeRadarDradis():
 HudGaugeRadar(HUD_OBJECT_RADAR_BSG, 255, 255, 255), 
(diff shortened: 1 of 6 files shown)


wookieejedi (approved):
Good fix that will be useful for multiple mods, and overall LGTM

PR #6122, opened 2024-05-02
Accept "color" and "colour" in game_settings.tbl

Dark-Visor:
Allow optional strings in game_settings.tbl to recognise both spellings - color/colour.


Goober5000 (approved):
Looks good 👍

PR #6125, opened 2024-05-07
make initialization consistent among FSO, FRED, and qtFRED

Goober5000:
Make the initialization sequence consistent so that FSO, FRED, and qtFRED call their initialization functions in the same order.  Most of the changes involve bringing FRED and qtFRED into sync with FSO.  This fixes fireball prerequisites for the ship types table, applies the script initialization fixes from #5946 to FRED, and generally removes the possibility of any surprises due to FRED differing from FSO.

Also remove the empty `brief_init_colors()` function.


wookieejedi (approved):
Works in my tests and overall looks reasonable, would be happy to have a second pair of eyes on this, too

Issue #6075, opened 2024-03-31
Toggle Vanilla Time Compression - Speedrun Feature


CRGorman:
As part of the rules for speedrunning both TGW and 2, modifications cannot provide a unique advantage to the player. As such, the ability to push time compression past 4x is not permitted. I would like to suggest a configuration flag that locks time compression to only the classic values of 1x-4x.

Goober5000:
Implemented in #6128

PR #6128, opened 2024-05-09
allow restriction to vanilla time compression range

Goober5000:
This provides a command-line argument to restrict the time compression range to the classic 1x - 4x.  Implements #6075.

I'm not entirely satisfied with `-orig_speedx_range` but it's the best I came up with that fit into the 19-character limit.  I'd be open to other suggestions.


JohnAFernandez:
-clamp_time_compression-?

CRGorman:
> `-retail_time_compress`?

I think I like this one best.

Goober5000:
> `-clamp_time_compress`? `-retail_time_compress`?

Well, both of those are more than 19 characters (including the dash).

`-orig_time_compress` or `-fs2_time_compress` would fit, but they don't include the word "range".  My feeling was that the option needs to include "range", or "restrict", or "limit", but perhaps one could argue that the option description provides enough information.

Baezon:
Perhaps this is a separate concern, but if this is for the purposes of speedrunning, should this not in some way be displayed graphically? So that someone else watching gameplay can verify it.

Goober5000:
I suppose the option could be called `-speedrun_constraints` and it could add some sort of HUD overlay...

CRGorman:
I think that would be clear since the rules say the clock must always be visible. But if we want to go there, we could tie it to other speed run enhancements I have in mind.

Baezon:
Perhaps we'll leave this PR as is then, and do a separate 'speedrun settings' flag, which would imply this flag/also show something graphical for it.

You should definitely post an issue for whatever else enhancements you had in mind.

Baezon (approved):

PR #6130, opened 2024-05-10
Created per-weapon field for laser min pixel size.

Kestrellius:
There is now a field in weapons.tbl to set the min pixel size for the specific laser weapon being defined. If this field is not set, the laser uses the universal min pixel size defined in game_settings.tbl.


code/weapon/weapons.cpp
```diff
@@ -1261,6 +1261,12 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 	if (optional_string("@Laser Glow Tail Scale:")) {
 		stuff_float(&wip->laser_glow_tail_scale);
 	}
+  
+	float mpsbuffer;
+	if (parse_optional_float_into("@Min Pixel Size:", &mpsbuffer))
+		wip->min_pixel_size = mpsbuffer;
+	else
+		wip->min_pixel_size = Min_pixel_size_laser;
```
BMagnu, line 1269: This should not be handled using the else case, but rather be set as default during the reset method of the weapon_info struct.

BMagnu: And if you do so, you also don't need the mpsbuffer variable. You can either do it using the if-optionalstring-then-stuff method, or parse-optional-float-into directly

BMagnu (changes requested):

code/weapon/weapons.cpp
Kestrellius: Made the requested changes.

BMagnu (approved):

PR #6133, opened 2024-05-11
Fix option manager importance

Baezon:
Really, fixing category comparison, allowing importance to be used. When options are displayed, they are sorted first by their category, and then by the supplied 'importance' value. But the category comparison was done directly, using direct pointer values, in effect never comparing equal and never using importance.


Goober5000 (approved):
Good catch. 👍

PR #6134, opened 2024-05-11
Allow removing tabled flags

### LINKED ISSUE #5905: FR: A "remove flags" TBM option for ships and weapons.
For mods that built off Vanilla FS2 and the MVPs, ships and weapon flags can be a bit of a pain: Sometimes you want to remove just the one, like "in tech database". For ships, this requires a bit of copy-paste work. For weapons, this requires a *significant* amount of copy and paste work. @naomimyselfandi offered to provide a "remove flag" table option, and I'd really like that to be a thing in the feature.

naomimyselfandi:
Adds a `$Remove Flags:` option for ships, weapons, and subsystems. Closes #5905.

### DIFF
--- a/code/globalincs/flagset.h
+++ b/code/globalincs/flagset.h
@@ -37,6 +37,11 @@ class flagset {
 		return result;
 	}
 
+	flagset<T>& operator&=(const flagset<T>& other) {
+		this->values &= other.values;
+		return *this;
+	}
+
 	flagset<T> operator+(const T flag) const {
 		flagset<T> result = *this;
 		result.set(flag);
@@ -67,7 +72,6 @@ class flagset {
 
 	flagset<T>& operator|=(const flagset<T>& other) {
 		this->values |= other.values;
-
 		return *this;
 	}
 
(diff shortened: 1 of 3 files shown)


EatThePath:
It seems like it might make sense to have remove flags above add flags. The only case that comes to mind immediately is weapon spawn flags, but even just for peace of mind on a modder's part it feels more right to remove existing flags and then add new ones to me, and a single table is unlikely to ever need to remove flags it just added, so besides potential weird parsing corner cases there shouldn't be any end user need to have remove second.

Not a change I'll demand, but I'd be interested in your thoughts on it.

BMagnu (changes requested):
Two things:
1. Unless you have a good reason to prefer the opposite, I'd agree with EatThePath here: Removal of flags should be above addition of flags in the table, thus allowing to remove flags set prior, while preventing accidental removal of flags that were just set.
2. There is a large number of formatting changes to parts unrelated to this PR. It would be good to clean up and remove those, otherwise it really mucks up the git commit history, making it hard to track down bugs later.

naomimyselfandi:
Thinking about it further, there are enough edge cases that I'm not sure this handles properly that I'd like to hold off on it.

PR #6135, opened 2024-05-11
Do not unload debrief bitmaps in API mode

MjnMixael:
Fixes an issue where SCPUI Debrief would accidentally unload whatever's in the first 2 bitmap slots, usually fonts

I couldn't decide whether to init `Background_bitmaps` to -1, set it to -1 if loading is skipped, or just skip unloading. All would work. I opted for skipping unloading to match how the API does everything else.


JohnAFernandez (approved):
I think this is a good fix for the issue in question.

TRBlount (approved):

PR #6138, opened 2024-05-11
Support custom message types

### LINKED ISSUE #6132: Additional $Name entries in Messages.tbl
Currently Messages.tbl supports the following list of $Name entries: https://wiki.hard-light.net/index.php/Messages.tbl#.24Name:_2

It is actually possible to add additional entries of any name (for example $Oh No) and call them via send-messages in an Event. Having these additional entries causes Warnings on mission load however, and while for the most part they seem to work they occasionally will not load the correct Persona.

naomimyselfandi:
This PR allows custom message types to be defined. These can be used as acknowledgements for LuaAI orders, or with the new `send-builtin-message` operator.

Closes #6132.


code/mission/missionmessage.cpp
```diff
@@ -783,6 +783,56 @@ void parse_msgtbl(const char* filename)
...
+		return MESSAGE_PRIORITY_HIGH;
+	} else if (optional_string("Low")) {
+		return MESSAGE_PRIORITY_LOW;
+	} else {
+		required_string("Normal");
+		return MESSAGE_PRIORITY_NORMAL;
+	}
+}
+
+void parse_custom_message_table(const char* filename) {
```
naomimyselfandi, line 808: Why, you may ask, is this its own table, rather than a section within `messages.tbl`? Well, `messages.tbl` isn't actually parsed until a mission is loaded; conversely, LuaAI, a consumer of these message types, is loaded during startup. Loading `messages.tbl` earlier ended up having a cascading effect on the load order due to message filters. I know it's possible to load only part of a table, but I couldn't get it working the way I wanted it to. I'd certainly be willing to go that route if someone could point me in the right direction.

PR #6139, opened 2024-05-11
Adds muzzle effect functionality.

Kestrellius:
In weapons.tbl, there is now an option to, instead of defining a retail-style muzzle flash, set a tabled particle effect to activate when the weapon is fired.

The effect's source velocity is set to the parent object's velocity (usually the firing ship), so using +Parent velocity factor will allow the muzzle effect to stay with a moving ship rather than quickly falling behind. It would be useful to also have the option of parenting the particle effect to the ship, but if that functionality is ever added to particle effects, it should as far as I can tell be simple to slot it into the implementation I have here.

Addressing the possible argument that this should be done via the events system instead: I agree that the events system will be useful for certain types of very complex muzzle effects, once it has triggers for that properly set up. However, most muzzle flashes, even complex ones, don't require anything more than one particle effect as implemented here, and dealing with the event system (making a table entry for the event, in addition to the table entry for the particle effect) is extra complexity that will be dead weight most of the time. I think that both options should exist so that modders can use whichever is more appropriate.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -2111,9 +2111,6 @@ void turret_swarm_fire_from_turret(turret_swarm_info *tsi)
 
 		// muzzle flash?
 		if (Weapon_info[tsi->weapon_class].muzzle_effect.isValid()) {
-			//vec3d gun_world_pos;
-			//vm_vec_unrotate(&gun_world_pos, turret_pos, &Objects[OBJ_INDEX(objp)].orient);
-			//vm_vec_add2(&gun_world_pos, &Objects[OBJ_INDEX(objp)].pos);
 			//spawn particle effect
 			auto particleSource = particle::ParticleManager::get()->createSource(Weapon_info[tsi->weapon_class].muzzle_effect);
 			particleSource.moveTo(&turret_pos);
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1988,7 +1988,19 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 					}
 
 					// do mflash if the weapon has it
-					if (wip->muzzle_flash >= 0) {
+					if (wip->muzzle_effect.isValid()) {
+						//copy to convert from const vecs to vecs
+						//should be removed when changes are made to particle functions
+						vec3d firing_pos2 = *firing_pos;
+						vec3d firing_vec2 = *firing_vec;
+						//spawn particle effect
+						auto particleSource = particle::ParticleManager::get()->createSource(wip->muzzle_effect);
+						particleSource.moveTo(&firing_pos2);
+						particleSource.setOrientationFromVec(&firing_vec2);
+						particleSource.setVelocity(&Objects[parent_ship->objnum].phys_info.vel);
+						particleSource.finish();
+					}
+					else if (wip->muzzle_flash >= 0) {
 						mflash_create(firing_pos, firing_vec, &Objects[parent_ship->objnum].phys_info, wip->muzzle_flash);
 					}
 
@@ -2098,7 +2110,18 @@ void turret_swarm_fire_from_turret(turret_swarm_info *tsi)
 		}
 
 		// muzzle flash?
-		if (Weapon_info[tsi->weapon_class].muzzle_flash >= 0) {
+		if (Weapon_info[tsi->weapon_class].muzzle_effect.isValid()) {
+			//vec3d gun_world_pos;
+			//vm_vec_unrotate(&gun_world_pos, turret_pos, &Objects[OBJ_INDEX(objp)].orient);
+			//vm_vec_add2(&gun_world_pos, &Objects[OBJ_INDEX(objp)].pos);
+			//spawn particle effect
+			auto particleSource = particle::ParticleManager::get()->createSource(Weapon_info[tsi->weapon_class].muzzle_effect);
+			particleSource.moveTo(&turret_pos);
+			particleSource.setOrientationFromVec(&turret_fvec);
+			particleSource.setVelocity(&Objects[tsi->parent_objnum].phys_info.vel);
+			particleSource.finish();
+		}
+		else if (Weapon_info[tsi->weapon_class].muzzle_flash >= 0) {
 			mflash_create(&turret_pos, &turret_fvec, &Objects[tsi->parent_objnum].phys_info, Weapon_info[tsi->weapon_class].muzzle_flash);
 		}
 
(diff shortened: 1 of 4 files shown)


code/weapon/weapons.cpp
```diff
@@ -6791,6 +6795,16 @@ int weapon_create( const vec3d *pos, const matrix *porient, int weapon_type, int
...
+	if (wip->muzzle_effect.isValid()) {
+		auto particleSource = particle::ParticleManager::get()->createSource(wip->muzzle_effect);
+		particleSource.moveTo(&wp->start_pos);
+		particleSource.setOrientationMatrix(orient);
+		if(parent_objp != NULL){
+			particleSource.setVelocity(&parent_objp->phys_info.vel);
+		}
+		particleSource.finish();
+	}
+
```
Baezon, line 6807: This should be alongside typical muzzle flash handling in `shipfx_flash_create`. There are a number of concerns to be dealt with which that function does, like not doing them while in cockpit view, and also imagine shotgun-type weapons, we almost always don't want duplicate effects triggered for every pellet.

Baezon (changes requested):

code/weapon/weapons.cpp
Kestrellius: Ah, okay. I had noticed that the traditional muzzle flashes were done there and was mostly chalking it up to typical Volition weirdness; it seemed like a very strange decision to me. But that makes sense.

Kestrellius: How's this? Not sure if it's the best control flow.

Baezon: It's a bit uh... hard to read... you'll want to be careful about any automatic formatters, in addition to it being difficult to see the changes you actually made, it'll also make mincemeat of the blame.

Kestrellius: I see.

Baezon: Other than that, based on what I can tell the meaningful changes are, it looks good to me, basically how I would've done it. You might pull out
```
vec3d real_dir;
vm_vec_rotate(&real_dir, gun_dir, &objp->orient);
```
since it appears to be the same for both of them.

Kestrellius: All right. I wasn't sure what real_dir was for, and based on what it looked like, I was thinking it maybe rotated the direction vector according to the orientation of the weapon (which might be relevant in cases where the weapon is not being fired straight forward, e.g. with FoF).

Baezon: In both cases it's for determining the orientation of the effect (though they don't use a full orientation, they only need the forward direction) since neither the particle or muzzle effect are going to maintain any reference to the weapon itself.

Goober5000 (changes requested):
Please ~~revert~~ remove the commit that reformatted most of shipfx.cpp.  There are two very important reasons for this.  First, it cripples the ability to compare the file against its version history, which is necessary for tracking down bugs and feature histories.  Second, it adds a significant amount of "noise" to the PR, making it almost impossible to follow the meaningful changes.

code/weapon/weapons.cpp
Kestrellius: Okay, the real_dir rotation was in fact...rotating the direction upward by a few degrees for some reason? I don't really understand why, but I removed it and now it works properly.

Kestrellius:
All right, working on that.

code/weapon/weapons.cpp
Baezon: Oh, I didn't realize it was causing issues for you. Yeah I realize now it's turning the world direction into a local direction, since muzzle flashes are rendered local to their parent ship. But that's not true for particle effects which will mostly use world for everything.

Kestrellius:
Do the commits I just pushed remove the problem? If not, I'll branch as described.

code/ship/shipfx.cpp
```diff
@@ -1051,12 +1051,28 @@ void shipfx_flash_create(object *objp, int model_num, vec3d *gun_pos, vec3d *gun
...
-	if ((Weapon_info[weapon_info_index].muzzle_flash >= 0) && !(Weapon_info[weapon_info_index].wi_flags[Weapon::Info_Flags::Flak]) &&
+	bool player_show_ship_model =
+		objp == Player_obj && Ship_info[Ships[objp->instance].ship_info_index].flags[Ship::Info_Flags::Show_ship_model];
+	if (!(Weapon_info[weapon_info_index].wi_flags[Weapon::Info_Flags::Flak]) &&
 		(objp != Player_obj || Render_player_mflash || (!in_cockpit_view || player_show_ship_model))) {
-		vec3d real_dir;
-		vm_vec_rotate(&real_dir, gun_dir,&objp->orient);	
-		mflash_create(gun_pos, &real_dir, &objp->phys_info, Weapon_info[weapon_info_index].muzzle_flash, objp);		
+			if (Weapon_info[weapon_info_index].muzzle_effect.isValid()) {
+				//if there's a muzzle effect entry, we use that
```
Baezon, line 1059: (I don't think this comment applies here)

Kestrellius: Hmm, okay. I guess maybe it would make sense to put it before the if clause?

Goober5000:
That looks a lot better.  It doesn't completely solve the problem yet because there is still a reformatting in the git history.  But if you combine the most recent three commits into one commit, it will clean everything up.  To do this, do an interactive rebase (`git rebase -i master`) and change the most recent two commits ("Reverts previous commit" and "Clean version of shipfx move") to `fixup`.  This will combine them with the immediate prior commit.

This can be a tricky process, so if it isn't clear, feel free to say so and I'll elaborate on any part.

code/ship/shipfx.cpp
Baezon: Oh whoops, my bad, I misunderstood, I thought this was an old comment pertaining to the old muzzle flash entry handling, you've got it right.

Baezon:
Honestly, much easier to just squash for the merge, imo.

Goober5000:
Theoretically, but that runs the risk of the PR being accidentally merged without squashing.

Kestrellius:
All right, that should hopefully solve the reformatting issue. Now I need to move a couple comments around, and make sure to cover some more instances of muzzle flashes that Baezon noticed.

Goober5000:
Much better, thank you. 👍

code/ai/aiturret.cpp
```diff
@@ -2098,7 +2110,18 @@ void turret_swarm_fire_from_turret(turret_swarm_info *tsi)
 		}
 
 		// muzzle flash?
-		if (Weapon_info[tsi->weapon_class].muzzle_flash >= 0) {
+		if (Weapon_info[tsi->weapon_class].muzzle_effect.isValid()) {
+			//vec3d gun_world_pos;
+			//vm_vec_unrotate(&gun_world_pos, turret_pos, &Objects[OBJ_INDEX(objp)].orient);
+			//vm_vec_add2(&gun_world_pos, &Objects[OBJ_INDEX(objp)].pos);
```
Baezon, line 2116: If you're sure this isn't necessary should probably remove these.  It does appear to be already in world coordinates so probably not necessary, but if you're unsure you should test it. I'm otherwise not a big fan of adding commented-out code.

Kestrellius: Oh! Whoops, yeah, that's not supposed to be there. Thank you for catching it.

Baezon (approved):

Goober5000 (approved):

PR #6140, opened 2024-05-12
several small fixes, mostly for FRED/qtFRED

Goober5000:
1. Add the "Override #Command in event messages" checkbox to qtFRED
2. Add the "Preload Subspace Tunnel" checkbox to the proper button group in qtFRED
3. Reorder lines in fred.rc so the declarations match the order they appear in the dialog
4. Use the correct name when reporting fireball index errors for ship classes


JohnAFernandez (approved):
Thank you

PR #6142, opened 2024-05-12
Allow defining custom message types

### LINKED ISSUE #6132: Additional $Name entries in Messages.tbl

naomimyselfandi:
Closes #6132.


code/parse/sexp.cpp
```diff
@@ -14742,6 +14756,73 @@ void sexp_send_message(int n)
...
+	return nullptr;
+}
+
+void sexp_send_builtin_message(int n) {
+	// Argument 1: Message type
+	auto type = get_builtin_message_type(CTEXT(n));
+	n = CDR(n);
+	// Argument 2: Subject (or <none>)
+	auto subject_name = CTEXT(n);
+	auto need_subject = stricmp(subject_name, SEXP_NONE_STRING);
```
BMagnu, line 14785: We should explicitly make a comparison here or cast to bool, otherwise this'll get int as type, which may behave unintuitively for anyone working on this code.
```suggestion
	auto need_subject = stricmp(subject_name, SEXP_NONE_STRING) != 0;
```

code/parse/sexp.cpp
```diff
@@ -14742,6 +14756,73 @@ void sexp_send_message(int n)
...
+			message_send_builtin(type, nullptr, subject);
+			return;
+		} else {
+			auto sender = get_builtin_message_sender(sender_name);
+			if (sender != nullptr) {
+				message_send_builtin(type, sender, subject);
+				return;
+			}
+		}
+	}
```
BMagnu, line 14823: Dunno about the use case, so this can be ignored if you think it doesn't matter, but:
Does it maybe make sense to return whether a message was actually sent? If all given senders are invalid or silenced, it could be important to tell the fredder.

BMagnu (changes requested):
Small style thing and one question.

code/parse/sexp.cpp
naomimyselfandi: I thought about this, but I don't think it's a good idea. We'd need to make a second operator that differs only in return type: just making it return a boolean would force the FREDer to wrap it in a `when` in the common case, and adding an optional argument for a variable that holds the result isn't possible because it's variadic. I feel like the "right" way to handle this is to come up with some sort of "has side-effects" flag that lets a non-void operator be used as if it were void; I'm sure there are other operators where that could be convenient.

BMagnu: Fair enough, that makes sense!

PR #6144, opened 2024-05-12
properly distinguish between adjusting voice and effects volume

Goober5000:
When messages are distorted due to EMP effects, the voice volume should be controlled by the voice setting, not the effects setting.  Fixes a bug noticed by @skdKitsune, a portion of which dates back to the original public source release.

Also, when volume is adjusted in `snd_set_volume`, make sure the correct voice volumes are used.  Fixes a copy-paste bug from #5694.

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -40,7 +40,6 @@
 #include "sound/fsspeech.h"
 #include "species_defs/species_defs.h"
 #include "utils/Random.h"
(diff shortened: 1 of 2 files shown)


Goober5000:
Testing has shown there are some additional factors at work here...

Goober5000:
Okay, all of the volume mismatches here have been found and fixed.

wookieejedi (approved):

PR #6145, opened 2024-05-12
Expose Collision Rotation Factor to Players

wookieejedi:
Previously, there were hardcoded checks in place to make players have 0 rotational effects when colliding with another ship. This PR adds a new value, `+Player Multiplier for Rotation Factor:` that allows modders to have players experience rotational effects from collisions. Also performed some cleanup of the light and heavy collisional rotational calculations by removing older if-blocks following useful recommendation from @Baezon.

Tested and works as expected.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -633,7 +633,7 @@ void parse_ai_profiles_tbl(const char *filename)
 
 				set_flag(profile, "$align to target when guarding stationary ship:", AI::Profile_Flags::Align_to_target_when_guarding_still);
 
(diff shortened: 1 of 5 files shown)


Baezon (commented):
Hmm, I was thinking of a global value, rather than per ship. Does your use case have it be different on different ships?

wookieejedi:
Ah yeah a global value is a better idea, will also help keep the current physics strut leaner.

wookieejedi:
Updated to global value in AI profiles as requested, thanks!

Baezon (approved):

Goober5000 (approved):
👍

PR #6146, opened 2024-05-13
Laser bolt offset

Kestrellius:
Adds an option in weapons.tbl to offset the rendered position of a laser bolt relative to the position of the weapon object, according to a vector.

The behavior of offsets along the z axis (the direction the laser is pointing) is intuitive; however, the orientation of the x and y axes are determined by the rotational orientation of the weapon -- which by default seems to match the orientation of the firing ship, but which would change if the weapon rotates for some reason. I plan to clearly explain this in wiki documentation, and recommend that modders typically only use z offset.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -8795,11 +8795,13 @@ void weapon_render(object* obj, model_draw_list *scene)
 			if (laser_length < 0.0001f)
 				return;
 
-			const float scaled_laser_length_pos_offset = laser_length*wip->laser_pos_offset.xyz.x;
+			vec3d rotated_offset;
+
+			vm_vec_unrotate(&rotated_offset, &wip->laser_pos_offset, &obj->orient);
 
 			vec3d headp, tailp;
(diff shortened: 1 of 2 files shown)


Baezon (approved):
You love to see 'em, simple, but open-ended.

code/weapon/weapons.cpp
```diff
@@ -8791,6 +8795,14 @@ void weapon_render(object* obj, model_draw_list *scene)
 			if (laser_length < 0.0001f)
 				return;
 
+			vec3d rotated_offset;
+
+			vm_vec_unrotate(&rotated_offset, &wip->laser_pos_offset, &obj->orient);
+
+			vec3d headp, tailp;
+				vm_vec_scale_add(&tailp, &obj->pos, &rotated_offset, laser_length);
```
Goober5000, line 8803: Is this line correct?  Or should it be:
```C++
vm_vec_add(&tailp, &obj->pos, &rotated_offset);
```

Goober5000 (changes requested):
Can you document somewhere (as you clarified to me) that laser position defaults to the tail of the bolt, not the head?

Also, one code question.

Goober5000 (approved):
Sounds good 👍

PR #6147, opened 2024-05-13
Changes particle effect functions to take consts.

Kestrellius:
This makes it so that the functions for setting up a particle effect, which don't need non-const data, take consts. This way they'll be easier to use when the data you have is a const.


Goober5000 (changes requested):
The declarations need to be changed in the .cpp file as well.

Also, now that #6139 has been merged, update your master branch, then rebase this PR onto master so it's based on the latest changes.  Then you can change this section from that PR:
```C++
//copy to convert from const vecs to vecs
//should be removed when changes are made to particle functions
```

Kestrellius:
Okay, will do.

Goober5000 (approved):
👍

PR #6148, opened 2024-05-13
refine process for capping speed due to physics whack/shock/collide

Goober5000:
The three functions `physics_apply_whack`, `physics_apply_shock`, and `physics_collide_whack` limited speed after the physics had been applied, but they limited absolute speed, not the speed due to the physics.  This modifies the process to limit the change in speed.  It also adds an AI profiles flag to disable the limit if desired.

This also tweaks the three functions (by editing comments and moving lines) to be more consistent, since all three are doing very similar calculations near the end.

### DIFF
--- a/code/ai/ai_flags.h
+++ b/code/ai/ai_flags.h
@@ -163,6 +163,7 @@ namespace AI {
 		Hudsquadmsg_tactical_disarm_disable,
 		Align_to_target_when_guarding_still,
 		Debris_respects_big_damage,
+		Dont_limit_change_in_speed_due_to_physics_whack,
 
 		NUM_VALUES
 	};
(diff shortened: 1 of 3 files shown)


Baezon:
What's the thought process behind the change? I can't understand a flag toggling it off, but why are we applying only to the delta now? Did this situation occur for you or another modder?

Goober5000:
This situation occurs in a Scroll mission under development.  A ship moving at more than 500 m/s is incorrectly buffeted around due to whacks and also has its speed reduced under 500 m/s.  Hence the flag.

As for the change from absolute speed to delta is because the more I started thinking about the original rationale, the more a delta made sense.  Volition didn't want a situation where ships got yeeted off at ludicrous speeds.  In normal FS gameplay, an absolute value of 500 is almost the same as a delta value of 500.  But for mods that use higher speeds, collisions shouldn't suddenly behave incorrectly just because vehicles are playing at 100s of m/s rather than 10s of m/s.

Baezon (approved):
Alright, works for me.

Goober5000:
Tests are working but marking this is draft to check a few more things.

PR #6149, opened 2024-05-15
Make headon transition behavior consistent between laser and glow bitmaps

Kestrellius:
Previously, the transition state between main and headon bitmaps was calculated independently for a laser's laser bitmap and its glow bitmap. Sometimes, this caused the glow bitmap to change earlier than the laser bitmap, or vice versa. This is undesirable behavior; the state of the glow bitmap should always exactly match the state of the laser bitmap in this regard.

I altered the logic so that glow headon rendering uses the alpha multiplier already calculated for the main bitmap, ensuring that they will always be the same. As a bonus, this approach should be some amount more performant, since we only have to do the alpha-determination math one time instead of twice.

Note: if there is no laser bitmap rendered, the headon transition math for it will never be run, and the alpha multiplier will always stay at its default, 1.0. In other words, if you have no valid laser bitmap, but you _do_ have a glow bitmap, it will never transition to its headon bitmap no matter what. I could add special-case logic to deal with this, but I do not expect it to be relevant very often, and if it does come up it's trivial to work around.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -8740,16 +8740,6 @@ float weapon_render_headon_bitmap(object* wep_objp, vec3d* headp, vec3d* tailp, 
 	return side_alpha;
 }
 
-// renders another glow bitmap on top of the regular bitmap based on the angle of the camera to the front of the laser
-// uses the alpha data already gathered by weapon_render_headon_bitmap, because the alpha of the glow should always match the alpha of the laser
-void weapon_render_headon_glow_bitmap(float side_alpha, vec3d* headp, vec3d* tailp, int bitmap, float width1, float width2, int r, int g, int b){
-	float head_alpha = 1.0 - side_alpha;
-
-	r = (int)(r * head_alpha);   g = (int)(g * head_alpha);   b = (int)(b * head_alpha);
-
-	batching_add_laser(bitmap, headp, width1, tailp, width2, r, g, b);
-}
-
 void weapon_render(object* obj, model_draw_list *scene)
 {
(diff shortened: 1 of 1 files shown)


code/weapon/weapons.cpp
```diff
@@ -8740,6 +8740,16 @@ float weapon_render_headon_bitmap(object* wep_objp, vec3d* headp, vec3d* tailp,
...
 
+// renders another glow bitmap on top of the regular bitmap based on the angle of the camera to the front of the laser
+// uses the alpha data already gathered by weapon_render_headon_bitmap, because the alpha of the glow should always match the alpha of the laser
+void weapon_render_headon_glow_bitmap(float side_alpha, vec3d* headp, vec3d* tailp, int bitmap, float width1, float width2, int r, int g, int b){
+	float head_alpha = 1.0 - side_alpha;
+
+	r = (int)(r * head_alpha);   g = (int)(g * head_alpha);   b = (int)(b * head_alpha);
+
+	batching_add_laser(bitmap, headp, width1, tailp, width2, r, g, b);
+}
```
Baezon, line 8751: This feels a bit trivial to have a function for. Yeah, it saves a bit of space for the caller, but I'm not sure it's offset by the introduction of this indirection.

Baezon (changes requested):
In addition to the below, can I request you change the `static_cast<int>` and `static_cast<float>` in this area to `fl2i(..` and `i2fl(..` respectively? I wasn't aware of them at the time, and they are a real eye sore.

Baezon (approved):

Issue #3818, opened 2021-12-18
Add/Modify Container dialog for qtFRED


jg18:
Add to qtFRED the "Add/Modify Container" dialog that was added to FRED in PR #3463.

No idea when I/someone will get to this, but creating a ticket to keep track of it.

jg18:
@JohnAFernandez since you're working on a unified qtFRED dialog for variables and containers, would you like to assign this to yourself?

MjnMixael:
Considering this completed as the sexp tree stuff will be covered in #6982

PR #6150, opened 2024-05-16
qtFRED variables editor

JohnAFernandez:
Draft because there are a couple more issues to iron out with character limits of list and map items and applying the changes to the mission, but it's close enough to being done that people can start taking a look.  The basic functionality is pretty well tested, but I think we'll need some real-world testing once it's in.

Note: This allows duplicate keys until the dialog is accepted, at which point it checks for duplicate keys, warns the user and prevents saving, listing the exact items that are problematic.  This allows users to make edits without qtFRED erasing user inputs if they happen to be duplicates.  There's some new features as well, like value-key switching.  Variable and container copying, etc.

### DIFF
--- a/qtfred/src/mission/dialogs/VariableDialogModel.h
+++ b/qtfred/src/mission/dialogs/VariableDialogModel.h
@@ -71,7 +71,7 @@ public:
 	SCP_string changeVariableName(int index, SCP_string newName);
 	SCP_string copyVariable(int index);
 	// returns whether it succeeded
-	bool removeVariable(int index);
+	bool removeVariable(int index, bool toDelete);
 
 	// Container Section
 
@@ -97,15 +97,15 @@ public:
 	SCP_string addContainer(SCP_string nameIn);
 	SCP_string copyContainer(int index);
 	SCP_string changeContainerName(int index, SCP_string newName);
-	bool removeContainer(int index);
+	bool removeContainer(int index, bool toDelete);
 
 	SCP_string addListItem(int index);
 	SCP_string copyListItem(int containerIndex, int index);
-	bool removeListItem(int containerindex, int index);
+	bool removeListItem(int containerindex, int index, bool toDelete);
 
 	std::pair<SCP_string, SCP_string> addMapItem(int index);
 	std::pair<SCP_string, SCP_string> copyMapItem(int index, int itemIndex);
-	bool removeMapItem(int index, int rowIndex);
+	bool removeMapItem(int index, int rowIndex, bool toDelete);
 
 	void shiftListItemUp(int containerIndex, int itemIndex);
 	void shiftListItemDown(int containerIndex, int itemIndex);
@@ -134,6 +134,8 @@ private:
 	SCP_vector<containerInfo> _containerItems;
 	int _listTextMode = 0;
 	int _mapTextMode = 0;
+
+	static int _deleteWarningCount = 0;
 
 	static SCP_string clampIntegerString(SCP_string source);
 
--- a/qtfred/src/mission/dialogs/VariableDialogModel.h
+++ b/qtfred/src/mission/dialogs/VariableDialogModel.h
@@ -75,6 +75,7 @@ public:
 	// returns whether it succeeded
 	bool removeVariable(int index, bool toDelete);
 	bool safeToAlterVariable(int index);
+	bool safeToAlterVariable(const VariableInfo& variableItem);
 
 	// Container Section
 
@@ -127,6 +128,7 @@ public:
 	void swapKeyAndValues(int index);
 	
 	bool safeToAlterContainer(int index);
+	bool safeToAlterContainer(const containerInfo& containerItem);
 
 	const SCP_vector<std::array<SCP_string, 3>> getVariableValues();
 	const SCP_vector<std::array<SCP_string, 3>> getContainerNames();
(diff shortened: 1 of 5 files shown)


JohnAFernandez:
While I still have to a last run of tests on the last two features I added, this is now basically done and ready for review.

JohnAFernandez:
Yesum

qtfred/src/ui/dialogs/VariableDialog.cpp
```diff
@@ -0,0 +1,1873 @@
...
+	this->setFocus();
+	ui->setupUi(this);
+	resize(QDialog::sizeHint()); // The best I can tell without some research, when a dialog doesn't use an underlying grid or layout, it needs to be resized this way before anything will show up 
+
+	// Major Changes, like Applying the model, rejecting changes and updating the UI.
+	// Here we need to check that there are no issues with variable names or container names, or with maps having duplicate keys.
+	connect(ui->OkCancelButtons, &QDialogButtonBox::accepted, this, &VariableDialog::checkValidModel);
+	// Reject if the user wants to.
+	connect(ui->OkCancelButtons, &QDialogButtonBox::rejected, this, &VariableDialog::reject);
+	connect(this, &QDialog::accepted, _model.get(), &VariableDialogModel::apply);
```
MjnMixael, line 26: If we change these to Qt slots then it might be good to follow the accepted/rejected signal setup used in the new asteroid editor for consistency. Up to you whether to do this now or leave it to a follow up.

qtfred/src/ui/dialogs/VariableDialog.cpp
```diff
@@ -0,0 +1,1873 @@
...
+
+	connect(ui->swapKeysAndValuesButton,
+		&QPushButton::clicked,
+		this,
+		&VariableDialog::onSwapKeysAndValuesButtonPressed);
+
+	connect(ui->selectFormatCombobox,
+		QOverload<int>::of(&QComboBox::currentIndexChanged),
+		this,
+		&VariableDialog::onSelectFormatComboboxSelectionChanged);
```
MjnMixael, line 211: One thing I think we should do in general is change all of these connect() lines to Qt slots. The pattern is usually something like on_uiElementName_SIGNAL(). So for example, the above could be changed to a method called on_swapKeysAndValuesButton_clicked(). The Qt docs can help figure out what the signals are and what their signatures are.

That said, I'll leave it up to you as to whether you want to do that refactor now as part of this PR or leave it to a follow-up. I'm good with either option.

qtfred/src/ui/dialogs/VariableDialog.cpp
```diff
@@ -0,0 +1,1873 @@
...
+
+	// this will update the list/map items.
+	updateContainerOptions(safeToAlterContainer);
+
+	_applyingModel = false;
+};
+
+void VariableDialog::updateVariableOptions(bool safeToAlter)
+{
+	int row = getCurrentVariableRow();
```
MjnMixael, line 1357: If converting to slots, you might need to add Qt's signal blocker here to prevent recursive loops when the code sets a ui elements state.

qtfred/src/ui/dialogs/VariableDialog.cpp
```diff
@@ -0,0 +1,1873 @@
...
+		ui->setVariableAsEternalcheckbox->setEnabled(true);
+		ui->setVariableAsEternalcheckbox->setChecked(_model->getVariableEternalFlag(row));
+	}
+
+	ui->networkVariableCheckbox->setChecked(_model->getVariableNetworkStatus(row));
+}
+
+void VariableDialog::updateContainerOptions(bool safeToAlter)
+{
+	int row = getCurrentContainerRow();
```
MjnMixael, line 1444: Same here regarding the signal blocker.

qtfred/src/ui/dialogs/VariableDialog.cpp
```diff
@@ -0,0 +1,1873 @@
...
+			ui->setContainerAsEternalCheckbox->setEnabled(true);
+			ui->setContainerAsEternalCheckbox->setChecked(_model->getContainerEternalFlag(row));
+		}
+
+	}
+}
+
+void VariableDialog::updateContainerDataOptions(bool list, bool safeToAlter)
+{
+	int row = getCurrentContainerRow();
```
MjnMixael, line 1576: And probably here, too.

qtfred/src/mission/dialogs/VariableDialogModel.h
```diff
@@ -0,0 +1,266 @@
...
+				return &_containerItems[containerIndex].numberValues[itemIndex];	
+			}
+		}
+		
+		return nullptr;
+	}
+
+
+	// many of the controls in this editor can lead to drastic actions, so this will be very useful.
+	bool confirmAction(SCP_string question, SCP_string informativeText)
```
MjnMixael: This is a UI action inside the model which isn't totally ideal but I'm not sure how I'd design it differently right now.

MjnMixael (approved):
Overall this looks good. I've made some suggestions based on how I'd like to design QtFRED going forward but this PR is a monster and it's over a year old. I'd lean towards getting the functionality in and refactoring in a follow-up but I'll leave that decision up to you.

MjnMixael:
Note this fixes #3818

PR #6152, opened 2024-05-17
add createRandomOrientation

Goober5000:
Considering that we have `ba.createRandomVector()`, we should have `ba.createRandomOrientation()` too.  This is based on an algorithm devised by @Baezon.

### DIFF
--- a/code/scripting/api/libs/base.cpp
+++ b/code/scripting/api/libs/base.cpp
@@ -214,6 +214,21 @@ ADE_FUNC(createRandomVector, l_Base, nullptr, "Creates a fairly random normalize
 	return ade_set_args(L, "o", l_Vector.Set(v3));
 }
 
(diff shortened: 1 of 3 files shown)


code/scripting/api/libs/base.cpp
```diff
@@ -214,6 +214,21 @@ ADE_FUNC(createRandomVector, l_Base, nullptr, "Creates a fairly random normalize
 	return ade_set_args(L, "o", l_Vector.Set(v3));
 }
 
+ADE_FUNC(createRandomOrientation, l_Base, nullptr, "Creates a fairly random orientation object.", "orientation", "Orientation object")
```
Baezon, line 217: I don't think we have to hedge our bets here with "fairly", I am quite sure this is uniformly random in the space of all possible matrices.

Goober5000: Okay

Baezon (approved):

PR #6156, opened 2024-05-18
Better rand vec for createRandomorientation

Baezon:
Follow-up to #6152. Forgot that `vm_vec_rand_vec` still uses the 'cube' method instead of the one in `vm_vec_random_in_sphere` which actually uses spherically uniform points. I thought about updating createRandomVector this way, and then updating `vm_vec_rand_vec` this way, and well maybe that's better for a separate PR.


Goober5000 (approved):
Yeah, that's a good idea.

PR #6157, opened 2024-05-18
Fix beam particle velocity

Kestrellius:
Previously, particle effects spawned from beams had their source velocity set once at the initiation of the effect, which meant that particles that continuously spawned from a slashing beam would inherit their velocity from the beam's *starting* state, such that the direction would be wrong as soon as the beam moved. I've fixed that by making beam-sourced particles get their velocity from the beam's current direction vector (it might actually be the last frame's; I'm not sure. Either way, it's close enough.), multiplied by its weapon speed.

I am somewhat concerned about performance here -- I don't have much frame of reference for how much it costs to look up these values. In particular, multiplying the vel by the beam's max_speed is only there to make things slightly more intuitive for the modder; without it, one can get identical behavior by just changing the particle effect's +Parent Velocity Factor. So if retrieving the max_speed from weapon_info has a non-negligible cost, I'd be inclined to just remove it.

Another note: a very small commit that removes unnecessary const keywords from two bools, which I was planning to include in a different pull request, seems to have ended up in this branch by mistake. This feels sort of sloppy, and if there's desire for me to disentangle these two commits and separate out the PRs more elegantly, I can, but as far as I can tell including it here won't actually hurt much of anything, so for now I'll PR as-is.

### DIFF
--- a/code/particle/ParticleSource.cpp
+++ b/code/particle/ParticleSource.cpp
@@ -128,6 +128,13 @@ vec3d SourceOrigin::getVelocity() const {
 			return m_origin.m_object.objp->phys_info.vel;
 		case SourceOriginType::PARTICLE:
 			return m_origin.m_particle.lock()->velocity;
+		case SourceOriginType::BEAM: {
+			beam* bm = &Beams[m_origin.m_object.objp->instance];
+			vec3d vel;
+			vm_vec_normalized_dir(&vel, &bm->last_shot, &bm->last_start);
+			vm_vec_scale(&vel, Weapon_info[bm->weapon_info_index].max_speed);
+			return vel;
+		}
 		default:
 			return m_velocity;
 	}
--- a/code/particle/ParticleSourceWrapper.h
+++ b/code/particle/ParticleSourceWrapper.h
@@ -59,9 +59,9 @@ namespace particle
 
 		void setOrientationFromNormalizedVec(const vec3d* normalizedDir, bool relative = false);
 
-		void setOrientationFromVec(const vec3d* dir, const bool relative = false);
+		void setOrientationFromVec(const vec3d* dir, bool relative = false);
 
-		void setOrientationMatrix(const matrix* mtx, const bool relative = false);
+		void setOrientationMatrix(const matrix* mtx, bool relative = false);
 
 		void setOrientationNormal(const vec3d* normal);
 


Baezon (commented):
Yeah, it's not like... great, performance-wise, but correctness, which this definitely sounds like is more important. It could be much worse. Most particle time is spent 'moving' and rendering them, this getVelocity is done once per particle when it is created, in general far less than per particle, per frame.

Actually, is this order of `last_start` and `last_shot`, right? It's the other way around in `beam_set_state`, but maybe that one is wrong.

Kestrellius:
This way makes the particles flow forward along the beam given a positive +Parent Velocity Factor, which seems the more intuitive behavior to me. When I copy-pasted the math, before I switched them around, the particles would flow backward, toward the muzzle.

It's not a huge deal, since again you can just make the factor negative in the effect entry, but this seems better to me. Plausibly beam_set_state should be modified? I don't know what all it does.

Kestrellius:
Actually, I'm surprised this is done per particle -- it seems like it _should_ be possible to do it per effect. Since this is getting the _source_ velocity, which AFAICT should always be the same for every particle spawned in one frame by one effect.

Baezon:
Perhaps, but the origin doesn't do any per frame updating. As you have written it, anyway, it is calculated when it is asked for, which is when they spawn.

And yes, it sounds like the direction should be reversed in `beam_set_state`.

Goober5000:
> without it, one can get identical behavior by just changing the particle effect's +Parent Velocity Factor

Can `+Parent Velocity Factor` be used to override your new calculation?  It looks like this new calculation is used for all particles regardless of that.  I can picture someone wanting to use both fast-moving and slow-moving particles for the same beam.


> Another note: a very small commit that removes unnecessary const keywords from two bools, which I was planning to include in a different pull request, seems to have ended up in this branch by mistake.

It's not ideal, but this is such a small change with minimal impact that it's fine.

(If the const/bool change is in *both* branches, and the other branch is merged first, this change will be moot anyway.)

Kestrellius:
> Can +Parent Velocity Factor be used to override your new calculation?

Yes! So, what's being changed in this PR is the _source_ velocity, which is independent of the particles' base velocity (which might be in a sphere or cone distribution, or whatever).

The only way the source velocity affects the particles at all is if +Parent Velocity Factor is set to something non-zero -- in which case the source velocity, multiplied by the factor, is added to whatever other velocity the particles have.

(Also, +Parent Velocity Factor can be set to a random range.)

> It's not ideal, but this is such a small change with minimal impact that it's fine.

Yeah, that's kind of what I figured.

Kestrellius:
> Can +Parent Velocity Factor be used to override your new calculation?

Actually... Well, okay. It _is_ true that it's no longer possible to replicate the previous behavior in the case of slash beams. But the previous behavior was ~a bug, and there is not much reason to want it.

Baezon:
The `const` doesn't make any sense in that context, its a bool which is passed by value anyway.

JohnAFernandez:
I mean, most times we don't go through that much effort, but if the value of that parameter isn't designed to change, that best practice still applies.  It's a pattern that Rust matches by making all variables const by default.

Kestrellius:
While I was working on a different feature, those functions not taking consts came up. Goober recommended that I change them to take consts, so I did, but I naively changed all the parameters to be consts, including bools. Goober (or someone? It's been a little while) indicated that adding const to a bool didn't do anything and I should remove those consts for cleanness reasons, so I did.

This debate seems to hinge on whether "const bool" is actually meaningfully different from "bool", and I don't know enough about the topic to have a position.

JohnAFernandez:
They don't do anything *in these functions* but removing them doesn't honestly make them more clean, and adding the additional change just makes the commit history less easy to read.  The difference is not in the variable, but in that the compiler will check if we accidentally change that value if it's const.  If these were long functions, it would be worth keeping as is, but because the functions are so short, it's not worth any more mental energy to think about.

PR #6161, opened 2024-05-21
Allow replacement of skybox textures

BMagnu:
Add a SEXP for this purpose and expose the skybox model & instance to scripting.


code/scripting/api/libs/mission.cpp
```diff
@@ -2177,6 +2178,44 @@ ADE_VIRTVAR(SkyboxAlpha, l_Mission, "number", "Sets or returns the current skybo
...
+
+		if (Nmodel_num >= 0) {
+			model_page_in_textures(Nmodel_num);
+
+			Nmodel_instance_num = model_create_instance(model_objnum_special::OBJNUM_NONE, Nmodel_num);
+			The_mission.skybox_model_animations.initializeMoveables(model_get_instance(Nmodel_instance_num));
+		}
+
+		// Since we have a new skybox we need to rerender the environment map
+		stars_invalidate_environment_map();
```
Goober5000, line 2208: What about calling `stars_set_background_model` instead?  The `texture_name` parameter would be nullptr and the flags and alpha would be the current flags and alpha.

code/parse/sexp.cpp
```diff
@@ -615,6 +615,7 @@ SCP_vector<sexp_oper> Operators = {
 	{ "get-collision-group",			OP_GET_COLGROUP_ID,						1,	1,			SEXP_ACTION_OPERATOR,	},
 	{ "change-team-color",				OP_CHANGE_TEAM_COLOR,					3,	INT_MAX,	SEXP_ACTION_OPERATOR,	},	// The E
 	{ "replace-texture",				OP_REPLACE_TEXTURE,						3,  INT_MAX,	SEXP_ACTION_OPERATOR,   },  // Lafiel
+	{ "replace-texture-skybox",				OP_REPLACE_TEXTURE_SKYBOX,						2, 2,	SEXP_ACTION_OPERATOR,   },  // Lafiel
```
Goober5000, line 618: My preference is `replace-skybox-texture` but I'll defer to whatever you and wookieejedi like.

code/model/modelrender.cpp
```diff
@@ -3072,6 +3072,40 @@ void model_render_set_wireframe_color(const color* clr)
 	Wireframe_color = *clr;
 }
 
+void modelinstance_replace_active_texture(polymodel_instance* pmi, const char* old_name, const char* new_name)
```
Goober5000, line 3075: This should bail (and probably warn) if `pmi` is nullptr.  Ships always have a model instance, but other object types are not guaranteed to.

code/parse/sexp.cpp
```diff
@@ -20775,14 +20777,20 @@ void ship_copy_damage(ship *target_shipp, ship *source_shipp)
...
-void sexp_replace_texture(int n)
+void sexp_replace_texture(int n, bool skybox)
 {
 	auto old_name = CTEXT(n);
 	n = CDR(n);
 
 	auto new_name = CTEXT(n);
 	n = CDR(n);
 
+	if (skybox) {
```
Goober5000, line 20788: This should check that `Nmodel_instance_num` is >= 0 (it could be -1 if the mission does not have a skybox).

Goober5000 (changes requested):

code/scripting/api/libs/mission.cpp
BMagnu: That'd need some rework then.
texture_name nullptr would unload whatever was in Nbitmap before. Does that have unintended side-effects?
The function would also need to be changed to accept already loaded models over a model file path, as switching models has to work from preloaded files with scripting.

code/model/modelrender.cpp
BMagnu: I'd have imagined that that's the responsibility of the caller, as this function is never directly exposed. But an Assert here is probably correct then

code/parse/sexp.cpp
wookieejedi: Ah yeah, following the format of the other skybox sexps (``"set-skybox-orientation" `"set-skybox-alpha"`) I see how this makes sense.

BMagnu: okay

code/scripting/api/libs/mission.cpp
Goober5000: Nbitmap is only meaningful in the context of the current model, so yes, unloading it when the model is changed is appropriate.  (In the original public source release, it is used as the spherec.pof texture file when the full nebula is shown.)

Changing the function to accept a model_num rather than a model_name looks like it could be done without much trouble, assuming a wrapper function that accepts model_name is also added.  The motivation here is to reduce code duplication in the script function.

wookieejedi:
Tested in multiple use cases and everything works as expected.

Goober5000 (approved):

PR #6162, opened 2024-05-22
Move Ship Editor Files And Coverity Clean-up

TheForce172:
Does what is says on the tin. Moves qtFRED's ship editor files into there own subfolder because they were getting a bit unwieldy in the mixed folder.

Also includes a clean-up of the files to use size_t and C++ style casts.

Required for Weapons Editor Pull Request


JohnAFernandez (commented):
It's looking good to me so far, but it looks like there's some issues with QVariant, which are getting caught by CI.

JohnAFernandez (approved):

PR #6165, opened 2024-05-22
Fix for 'decloaking' effect

### LINKED ISSUE #6164: 'Decloaking' effect does not work if 'Cloak' effect allowed to complete
The binary toggling of rendering or not when the cloak or decloak finishes still works as normal, but the transitionary wobbly visual effect for decloaking does not start at all, if a cloaking effect on the same ship has already finished, instead the ship stays invisible until the effect ends, where it becomes visible again. Tested using a fenris.

As far as I can tell, the proper values are being passed as parameters as part of the `model_material`

Baezon:
Follow-up to #4757. Fixes #6164. If the effect timestamp is valid and not yet elapsed, then the cloak flag should be off. Only a fully completed, `disables_rendering` effect gets the 'cloak' flag.


wookieejedi (approved):

Goober5000 (approved):
Good catch.

PR #6166, opened 2024-05-23
Makes random life factor of children configurable.

Kestrellius:
While working on another feature that involves weapon spawning, I discovered that the lifetime of spawned weapons is automatically increased or decreased by a hardcoded random factor bounded at 0.2 in either direction, and began screaming.

@Baezon pointed out that the proper way to approach this would be to make it configurable, so I did that. I think it's sort of a tossup whether it's better to have it be a parameter on the parent weapon or on the child weapon, but I went with child. I also made it so that the parameter can be arbitrarily large, but the resulting random multiplier for lifetime is bounded at 0.01.

Also, I'm 100% open to suggestions for the parameter name, because this one sucks.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -2616,9 +2616,9 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		}
 	}
 
-	if (optional_string("$As Child Life Rand Factor:"))
+	if (optional_string("$Lifetime Variation Factor When Child:"))
 	{
-		stuff_float(&wip->as_child_life_rand_factor);
+		stuff_float(&wip->lifetime_variation_factor_when_child);
 	}
 
 	if (wip->wi_flags[Weapon::Info_Flags::Local_ssm] && optional_string("$Local SSM:"))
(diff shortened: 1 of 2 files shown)


code/weapon/weapons.cpp
```diff
@@ -2616,6 +2616,11 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		}
 	}
 
+	if (optional_string("$As Child Life Rand Factor:"))
```
Baezon, line 2619: This is a really ugly and confusing name... but I'm not sure I have anything better. Certainly, if this applied to children of *this* weapon the naming would be much easier, but it was probably the right call to put it on the children themselves.

I would at least, spell out words like "random" and "lifetime" (also to make its like to $lifetime clearer), I think we can afford the verbosity here.

"Lifetime Random Factor When Spawned" maybe?

EatThePath: > "Lifetime Random Factor When Spawned" maybe?
I'd suggest "variation factor" instead of random factor but otherwise, that seems as good as it'll get.

Kestrellius: Yep, that sounds viable.

I might also... Well, hmm. Is it better to have this clustered with the spawning options like it is now, or next to $lifetime?

Baezon: Here, with the spawn stuff, imo.

Kestrellius: Went with "Lifetime Variation Factor When Child". "Spawned" seems too ambiguous to me -- I think it's not obvious that a weapon getting created normally doesn't count as "spawning".

Baezon (approved):

PR #6169, opened 2024-05-27
Weapons Editor Dialog and Table Viewer

TheForce172:
After several months of work.(And a year's break in the middle) the revised weapons dialog is done!

image

The Weapons list on the the left is functionally the same as the dropdowns in the old editor except they no longer filter out big weapons for small ships as their were places in retail that contradicted this rule anyway (the Alstor) and confused the old editor.

The tree view on the right list Turrets/Pilot and their banks. that contain the current type of weapon. It features multiselect that allows only one type of item to be selected at a time. So either Turrets/Pilot or banks not both.
Features setting the weapon by drag and drop or multi-select and a button press. This should hopefully be faster than having to scroll down through the list for every weapon.

Also supports setting the ai class via multi-select and combo box.

And finally it includes a weapon TBL viewer that functions exactly like the ships table viewer.

### DIFF
--- a/qtfred/src/mission/dialogs/ShipEditor/ShipWeaponsDialogModel.cpp
+++ b/qtfred/src/mission/dialogs/ShipEditor/ShipWeaponsDialogModel.cpp
@@ -1,0 +1,379 @@
+#include "ShipWeaponsDialogModel.h"
+namespace fso {
+namespace fred {
+Banks::Banks(const SCP_string& name, int aiIndex, int ship, int multiedit, ship_subsys* subsys)
+	: name(name), subsys(subsys), ship(ship), m_isMultiEdit(multiedit), initalAI(aiIndex)
+{
+	aiClass = aiIndex;
+}
+void Banks::add(Bank* bank)
+{
+	banks.push_back(bank);
+}
+Bank* Banks::getByBankId(const int id)
+{
+	for (auto bank : banks) {
+		if (id == bank->getWeaponId())
+			return bank;
+	}
+	return nullptr;
+}
+SCP_string Banks::getName() const
+{
+	return name;
+}
+int Banks::getShip() const
+{
+	return ship;
+}
+ship_subsys* Banks::getSubsys() const
+{
+	return subsys;
+}
+bool Banks::empty() const
+{
+	return banks.empty();
+}
+const SCP_vector<Bank*> Banks::getBanks() const
+{
+	return banks;
+}
+int Banks::getAiClass() const
+{
+	if (name == "Pilot") {
+		return Ships[ship].weapons.ai_class;
+	} else {
+		return subsys->weapons.ai_class;
+	}
+}
+void Banks::setAiClass(int newClass)
+{
+	if (m_isMultiEdit) {
+		object* ptr;
+		int inst;
+		ptr = GET_FIRST(&obj_used_list);
+		while (ptr != END_OF_LIST(&obj_used_list)) {
+			if (((ptr->type == OBJ_SHIP) || (ptr->type == OBJ_START)) && (ptr->flags[Object::Object_Flags::Marked])) {
+				inst = ptr->instance;
+				if (name == "Pilot") {
+					Ships[inst].ai_index = newClass;
+				} else {
+					subsys->weapons.ai_class = newClass;
+				}
+			}
+		}
+	} else {
+		if (name == "Pilot") {
+			Ships[ship].weapons.ai_class = newClass;
+		} else {
+			subsys->weapons.ai_class = newClass;
+		}
+	}
+}
+int Banks::getInitalAI()
+{
+	return initalAI;
+}
+Bank::Bank(const int weaponId, const int bankId, const int ammoMax, const int ammo, Banks* parent)
+{
+	this->weaponId = weaponId;
+	this->bankId = bankId;
+	this->ammo = ammo;
+	this->ammoMax = ammoMax;
+	this->parent = parent;
+}
+int Bank::getWeaponId() const
+{
+	return weaponId;
+}
+int Bank::getAmmo() const
+{
+	return ammo;
+}
+int Bank::getBankId() const
+{
+	return bankId;
+}
+int Bank::getMaxAmmo() const
+{
+	return ammoMax;
+}
+void Bank::setWeapon(const int id)
+{
+	weaponId = id;
+	if (Weapon_info[id].subtype == WP_LASER || Weapon_info[id].subtype == WP_BEAM) {
+		if (parent->getName() == "Pilot") {
+			ammoMax = get_max_ammo_count_for_primary_bank(parent->getShip(), bankId, id);
+		} else {
+			ammoMax = get_max_ammo_count_for_primary_turret_bank(&parent->getSubsys()->weapons, bankId, id);
+		}
+	} else {
+		if (parent->getName() == "Pilot") {
+			ammoMax = get_max_ammo_count_for_bank(parent->getShip(), bankId, id);
+		} else {
+			ammoMax = get_max_ammo_count_for_turret_bank(&parent->getSubsys()->weapons, bankId, id);
+		}
+	}
+}
+void Bank::setAmmo(const int newAmmo)
+{
+	this->ammo = newAmmo;
+}
+namespace dialogs {
+ShipWeaponsDialogModel::ShipWeaponsDialogModel(QObject* parent, EditorViewport* viewport, bool isMultiEdit)
+	: AbstractDialogModel(parent, viewport)
+{
+	initializeData(isMultiEdit);
+}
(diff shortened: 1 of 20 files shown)


MjnMixael:
Double checking that this is ready for review?

TheForce172:
Yep

qtfred/src/mission/dialogs/ShipEditor/ShipWeaponsDialogModel.cpp
```diff
@@ -0,0 +1,379 @@
...
+		return subsys->weapons.ai_class;
+	}
+}
+void Banks::setAiClass(int newClass)
+{
+	if (m_isMultiEdit) {
+		object* ptr;
+		int inst;
+		ptr = GET_FIRST(&obj_used_list);
+		while (ptr != END_OF_LIST(&obj_used_list)) {
```
MjnMixael, line 55: My knowledge of the FSO macros for linked lists is a bit fuzzy but this looks like an infinite loop as there's no GET_NEXT(). What moves the ptr forward here?

qtfred/src/mission/dialogs/ShipEditor/ShipWeaponsDialogModel.cpp
```diff
@@ -0,0 +1,379 @@
...
+	} else {
+		return subsys->weapons.ai_class;
+	}
+}
+void Banks::setAiClass(int newClass)
+{
+	if (m_isMultiEdit) {
+		object* ptr;
+		int inst;
+		ptr = GET_FIRST(&obj_used_list);
```
MjnMixael, line 54: This could be simplified.
object* ptr = GET_FIRST(&obj_used_list);

and `inst` looks like it can be moved entirely inside the while loop.

MjnMixael: This could be simplified.
object* ptr = GET_FIRST(&obj_used_list);

and inst looks like it can be moved entirely inside the while loop.

qtfred/src/mission/dialogs/ShipEditor/ShipWeaponsDialogModel.cpp
MjnMixael, line 52: My knowledge of the FSO macros for linked lists is a bit fuzzy but this looks like an infinite loop as there's no GET_NEXT(). What moves the ptr forward here?

qtfred/src/mission/dialogs/ShipEditor/ShipWeaponsDialogModel.cpp
```diff
@@ -0,0 +1,379 @@
...
+	bool first = true;
+	object* ptr;
+
+	m_ship = _editor->cur_ship;
+	if (m_ship == -1)
+		m_ship = Objects[_editor->currentObject].instance;
+
+	if (m_isMultiEdit) {
+		ptr = GET_FIRST(&obj_used_list);
+		while (ptr != END_OF_LIST(&obj_used_list)) {
```
MjnMixael, line 139: Same possible infinite loop here. What iterates the ptr?

qtfred/src/mission/dialogs/ShipEditor/ShipWeaponsDialogModel.cpp
```diff
@@ -0,0 +1,379 @@
...
+	initializeData(isMultiEdit);
+}
+void ShipWeaponsDialogModel::initializeData(bool isMultiEdit)
+{
+	m_isMultiEdit = isMultiEdit;
+	PrimaryBanks.clear();
+	SecondaryBanks.clear();
+	int inst;
+	bool first = true;
+	object* ptr;
```
MjnMixael: Some of these can be declared in a more confined scope below.

qtfred/src/mission/dialogs/ShipEditor/ShipWeaponsDialogModel.cpp
```diff
@@ -0,0 +1,379 @@
...
+		}
+	}
+}
+bool ShipWeaponsDialogModel::apply()
+{
+	if (m_isMultiEdit) {
+		object* ptr;
+		int inst;
+		ptr = GET_FIRST(&obj_used_list);
+		while (ptr != END_OF_LIST(&obj_used_list)) {
```
MjnMixael, line 339: Another possible infinite loop and vars that can be simplified or contained in a tighter scope.

qtfred/src/mission/dialogs/ShipEditor/ShipWeaponsDialogModel.h
```diff
@@ -0,0 +1,89 @@
...
+
+  private:
+	void saveShip(int inst);
+	void initPrimary(const int inst, bool first);
+
+	void initSecondary(int inst, bool first);
+	void initializeData(bool multi);
+	int m_isMultiEdit;
+	int m_ship;
+	int big = 1;
```
MjnMixael: Might as well change this too a bool for clarity if I'm reading it's usages right?

qtfred/src/ui/dialogs/ShipEditor/ShipWeaponsDialog.cpp
```diff
@@ -0,0 +1,181 @@
...
+ShipWeaponsDialog::ShipWeaponsDialog(QDialog* parent, EditorViewport* viewport, bool isMultiEdit)
+	: QDialog(parent), ui(new Ui::ShipWeaponsDialog()), _model(new ShipWeaponsDialogModel(this, viewport, isMultiEdit)),
+	  _viewport(viewport)
+{
+	ui->setupUi(this);
+
+	//Connect the mode change buttons
+	connect(ui->radioPrimary, &QRadioButton::toggled, this, [this](bool param) { modeChanged(param, 0); });
+	connect(ui->radioSecondary, &QRadioButton::toggled, this, [this](bool param) { modeChanged(param, 1); });
+	connect(ui->radioTertiary, &QRadioButton::toggled, this, [this](bool param) { modeChanged(param, 2); });
```
MjnMixael: One thing I think we should do in general is change all of these connect() lines to Qt slots. The pattern is usually something like `on_uiElementName_SIGNAL()`. So for these it'd be `on_radioPrimary_toggled(bool checked)`. The Qt docs can help figure out what the signals are and what their signatures are.

That said, I'll leave it up to you as to whether you want to do that refactor now as part of this PR or leave it to a follow-up. I'm good with either option.

qtfred/src/ui/dialogs/ShipEditor/ShipWeaponsDialog.cpp
```diff
@@ -0,0 +1,181 @@
...
+	} else {
+		ui->TBLButton->setEnabled(false);
+	}
+}
+
+void ShipWeaponsDialog::aiClassChanged(const int index) {
+	m_currentAI = ui->AICombo->itemData(index).toInt();
+}
+
+void ShipWeaponsDialog::on_AIButton_clicked() {
```
MjnMixael: It'd be good to follow a naming pattern throughout QtFRED if we can. This button should probably be aiButton. Minor nitpick, really, but the consistency does help overall later on.

qtfred/src/ui/dialogs/ShipEditor/ShipWeaponsDialog.h
```diff
@@ -0,0 +1,64 @@
...
+	explicit ShipWeaponsDialog(QDialog* parent, EditorViewport* viewport, bool isMultiEdit);
+	~ShipWeaponsDialog() override;
+
+  protected:
+	void closeEvent(QCloseEvent*) override;
+
+  private slots:
+	void on_AIButton_clicked();
+	void on_setAllButton_clicked();
+	void on_TBLButton_clicked();
```
MjnMixael: Same with this button name.

qtfred/ui/ShipWeaponsDialog.ui
```diff
@@ -0,0 +1,223 @@
...
+ </customwidgets>
+ <resources>
+  <include location="../resources/resources.qrc"/>
+ </resources>
+ <connections>
+  <connection>
+   <sender>buttonClose</sender>
+   <signal>clicked()</signal>
+   <receiver>fso::fred::dialogs::ShipWeaponsDialog</receiver>
+   <slot>accept()</slot>
```
MjnMixael: These connections in Qt are super easy to miss when debugging. What do you think of adopting the model I used in the asteroid editor rewrite instead? Remove this connection and add to the dialog header an `on_yourOkAndCancelButtonsName_accepted()` (or _rejected()) and those call the relevant accept/reject/apply/whatever methods as needed?

Again, I'll leave the decision to do this now or later up to you.

MjnMixael (changes requested):
Overall well done. Saw some minor nitpicks, a few possible infinite loops, and suggested some changes that we can do in a follow-up later on if you prefer.

PR #6170, opened 2024-05-27
Arrival/Departure Paths Dialog

TheForce172:
Not much to say on this one. It's pretty much just a copy paste of the FRED one.

### DIFF
--- a/qtfred/src/ui/dialogs/ShipEditor/ShipPathsDialog.cpp
+++ b/qtfred/src/ui/dialogs/ShipEditor/ShipPathsDialog.cpp
@@ -17,7 +17,7 @@ ShipPathsDialog::ShipPathsDialog(QWidget* parent,
 	ui->setupUi(this);
 	connect(ui->pathList, &QListWidget::itemChanged, this, &ShipPathsDialog::changed);
 	connect(this, &QDialog::accepted, _model.get(), &ShipPathsDialogModel::apply);
-	connect(this, &QDialog::rejected, _model.get(), &ShipPathsDialogModel::reject);
+	connect(ui->buttonBox, &QDialogButtonBox::rejected, this, &ShipPathsDialog::rejectHandler);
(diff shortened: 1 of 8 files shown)


qtfred/src/ui/dialogs/ShipEditor/ShipPathsDialog.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+		if (button == DialogButton::Cancel) {
+			event->ignore();
+			return;
+		}
+
+		if (button == DialogButton::Yes) {
+			accept();
+			return;
+		}
```
JohnAFernandez, line 47: Maybe we could do a pair of OK/Cancel buttons in the dialog and automatically accept if they hit OK?  As little friction as possible is better, imho.

TheForce172: It does see line 18-19 connect to the ok/cancel buttons on the dialog. This code runs only when the user presses the OS close button

JohnAFernandez: Ah, you're right.  Sorry for missing that.  We can still add a confirmation to the cancel button by using this connect statement instead of the default
`connect(ui->OkCancelButtons, &QDialogButtonBox::rejected, this, &yourDialog::reject);`

One of the goals will be to add confirmations to the cancel buttons that don't have them when we get to the polishing stage, anyway.

TheForce172: Credit where credits due, I didn't come up with it. I copied it from whoever did the mission specs dialog.

JohnAFernandez (approved):

JohnAFernandez:
Well done!  Looks cleaner than mine, tbh

PR #6173, opened 2024-05-30
Exposes the New Flight Cursor Mode to Scripting and Fixed Aiming Setting

wookieejedi:
Thanks to #6136, FSO supports switching between two methods of flying. This PR exposes those settings and values to scripters. Tested and works as expected.

Also, fixed the setting where the aiming would always aim at the cursor, even if the table setting was off.


Goober5000 (approved):

Baezon (approved):
Approved with regard to the fix in ship.cpp, `aims_at_flight_cursor` is meant to basically force the flight cursor behavior (but not the other way around!)

code/scripting/api/libs/controls.cpp
```diff
@@ -314,6 +315,79 @@ ADE_FUNC(AxisInverted, l_Mouse, "number cid, number axis, boolean inverted", "Ge
...
+			Player_flight_mode = FlightMode::FlightCursor;
+		else
+			Player_flight_mode = FlightMode::ShipLocked;
+	}
+
+	if (Player_flight_mode == FlightMode::FlightCursor)
+		return ADE_RETURN_TRUE;
+	else
+		return ADE_RETURN_FALSE;
+}
```
Baezon, line 334: Ah, one more thing, I do plan on having other options for this soon. So we might not want to lock this into a boolean. I don't really know how the lua enums work though... maybe a string name input?

Baezon (changes requested):

code/scripting/api/libs/controls.cpp
wookieejedi: Good idea, just updated.

Baezon (approved):

PR #6174, opened 2024-05-30
Allow overriding low-level keypresses

### LINKED ISSUE #4533: FR: Lua hooks into hardcoded UI controls

BMagnu:
Closes #4533.

This PR makes the On Key Pressed hook overridable.
Overriding a keypress will make the press basically non-existent to FSO. The only thing that the keypress will have an effect on apart from the override itself is that an On Key Released hook will still fire once the respective key is released.

This allows scripts as requested in the linked issue, such as the following to catch all escape's while in the scripting state:
```
$State: GS_STATE_SCRIPTING
$Raw KeyPress: Esc
$On Key Pressed: [
]
+Override: [
   return true
]
```

In order to achieve this, a bit of code-cleanup was needed. Key-Down checks now entirely run through function calls as opposed to reading the raw key array. The underlying data structures were cleaned, with unused ones removed and tidied. All data was moved within the keyboard struct. Key hold down timings had to be rewritten, which incidentally also fixes a bunch of potential edge case bugs I saw in the old code. Finally, the key queue is implemented as an actual queue and not as a poor-man's ring buffer.

In addition, the condition logic for key presses and raw key presses now caches, making checks by condition much faster than before, and thus also much faster than comparing in lua itself.

### DIFF
--- a/code/controlconfig/controlsconfig.cpp
+++ b/code/controlconfig/controlsconfig.cpp
@@ -2091,7 +2091,7 @@ int control_config_bind_key_on_frame(int ctrl, selItem item, bool API_Access)
 		}
 
 		if ((ctrl == BANK_WHEN_PRESSED || ctrl == GLIDE_WHEN_PRESSED) && (Last_key >= 0) && (k <= 0) &&
-			!keyd_pressed[Last_key]) {
+			!key_is_pressed(Last_key)) {
 			// If the selected cc_item is BANK_WHEN_PRESSED or GLIDE_WHEN_PRESSED, and
 			// If the polled key is a modifier, and
 			// k was consumed, and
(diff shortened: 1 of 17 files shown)


Goober5000 (approved):
Looks good 👍

PR #6178, opened 2024-06-02
Expose Impact Damage Type to Scripting

wookieejedi:
There's already a sexp that does this, but having the script version for ships is quite useful, too. For example, scripters can set impact damage types for just the player at various difficulties.

Tested and works as expected.

### DIFF
--- a/code/scripting/api/objs/ship.cpp
+++ b/code/scripting/api/objs/ship.cpp
@@ -245,6 +245,43 @@ ADE_VIRTVAR(ShieldArmorClass, l_Ship, "string", "Current Armor class of the ship
 	return ade_set_args(L, "s", name);
 }
 
+ADE_VIRTVAR(ImpactDamageClass, l_Ship, "string", "Current Impact Damage class", "string", "Impact Damage class name, or empty string if none is set")
+{
+	object_h *objh;
+	const char* s    = nullptr;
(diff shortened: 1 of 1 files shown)


code/scripting/api/objs/ship.cpp
```diff
@@ -245,6 +245,43 @@ ADE_VIRTVAR(ShieldArmorClass, l_Ship, "string", "Current Armor class of the ship
...
+	if (ADE_SETTING_VAR && s != nullptr) {
+		for (size_t i = 0; i < Damage_types.size(); i++) {
+			if (!stricmp(Damage_types[i].name, s)) {
+				damage_index = (int)i;
+				break;
+			}
+		}
+		if (damage_index > 0) {
+			shipp->collision_damage_type_idx = damage_index;
+		}
```
Baezon, line 272: The new damage type should be set to `damage_index` unconditionally, including setting it to -1 if no match was found for consistency with the set armor type ones.

Baezon (changes requested):

code/scripting/api/objs/ship.cpp
wookieejedi: Good point, just updated, thanks!

Baezon (approved):

Goober5000 (approved):

PR #6179, opened 2024-06-03
Increase Lua safety

BMagnu:
Increases overall safety by preventing lua from executing malicious code.
We actually do this kind of securing for the os library already, but the io library has similar issues.
We don't want to disable all io though, so apply some fairly restrictive filtering as to what can be opened.

Tested to work with scripts that use the io lib

### DIFF
--- a/code/scripting/lua.cpp
+++ b/code/scripting/lua.cpp
@@ -9,7 +9,7 @@ extern "C" {
 #include "scripting/lua/lua_ext.h"
 
 #ifdef WIN32
-#include <fileapi.h>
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Looks good as far as I can tell, though interfacing with the Lua backend always involves a great deal of [heavy wizardry](http://www.catb.org/jargon/html/H/heavy-wizardry.html).

BMagnu:
Confirmed to work with Zero Serenity's speedrun script, so it should be good to be merged now

PR #6180, opened 2024-06-04
utility functions for counting and searching

Goober5000:
Inspired by #6178, this moves the existing `count_items_with_name()` functions into utility.h and adds additional functions for finding specific items with a given `.name` or, more generally, a given field.  The `Damage_types` search has been converted to illustrate how these functions can be used.

### DIFF
--- a/code/globalincs/utility.h
+++ b/code/globalincs/utility.h
@@ -147,4 +147,145 @@ typename T::size_type stringcost(const T& op, const T& input, typename T::size_t
     return cost;
 }
 
+template <typename T>
+int count_items_with_name(const char* name, const T* item_array, int num_items)
+{
+	Assert(name != nullptr && item_array != nullptr);
+
+	int count = 0;
+	for (int i = 0; i < num_items; ++i)
+		if (!stricmp(name, item_array[i].name))
+			++count;
+
+	return count;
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):
Quite handy and looks good to me

code/globalincs/utility.h
```diff
@@ -147,4 +147,145 @@ typename T::size_type stringcost(const T& op, const T& input, typename T::size_t
     return cost;
 }
 
+template <typename T>
+int count_items_with_name(const char* name, const T* item_array, int num_items)
+{
+	Assert(name != nullptr && item_array != nullptr);
```
JohnAFernandez, line 153: This one should be changed to an Assertion so that we can know which of these would be nullptr when triggered.

JohnAFernandez (changes requested):

code/globalincs/utility.h
Goober5000: Okay

Goober5000: Perhaps... but then I thought of another option.  See latest commit, which just returns 0 or -1 as appropriate.

JohnAFernandez (approved):

PR #6181, opened 2024-06-04
Make the $Shots parameter work for beams.

Kestrellius:
This makes it so that the $Shots parameter, normally used for multishot weapons like shotguns (not to be confused with +shots, a beam info parameter used for antifighter beams), works correctly for beam weapons.

Multishot beams now work properly if they're fired normally from turrets, fired using the fire-beam SEXP, or fired as fighter primaries.

For Type 5 beams, the +burst rotation pattern will advance for each shot fired simultaneously, in addition to advancing between shots in a burst, allowing a modder to have beams which fire several shots in a pattern. This was previously possible using beams spawned as submunitions or from subspace, but now it works for beams fired normally. (This was the main reason I wanted this feature.)

I also modified the use of the burst rotation pattern so that if given a burst index larger than the length of the pattern, it wraps around to the beginning. Previously, a larger burst index would access memory beyond the array, and either crash the game or fire the beam at random angles.

All of this should be completely backwards-compatible, with the sole exception of cases where mods had beams with completely superfluous multishot specified. (Or cases where a beam was accessing random memory beyond the burst rot pattern, I guess?)

Since it involves loops added to preexisting code, some of the implementation is a little messy, especially the fighter primary version. I'm open to thoughts on how to make the code cleaner.

### DIFF
--- a/code/ai/aiturret.cpp
+++ b/code/ai/aiturret.cpp
@@ -1781,45 +1781,49 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum, 
 			if (!(turret->weapons.flags[Ship::Weapon_Flags::Beam_Free])) {
 				return false;
 			}
-			beam_fire_info fire_info;
 
-			// stuff beam firing info
-			memset(&fire_info, 0, sizeof(beam_fire_info));
-			fire_info.accuracy = 1.0f;
-			fire_info.beam_info_index = turret_weapon_class;
-			fire_info.beam_info_override = nullptr;
-			fire_info.shooter = &Objects[parent_objnum];
-			fire_info.target = turret_enemy_objp;
-			if (wip->wi_flags[Weapon::Info_Flags::Antisubsysbeam])
-				fire_info.target_subsys = turret->targeted_subsys;
-			else
-				fire_info.target_subsys = nullptr;
-			fire_info.turret = turret;
-			fire_info.burst_seed = old_burst_seed;
-			fire_info.fire_method = BFM_TURRET_FIRED;
-			fire_info.per_burst_rotation = swp->per_burst_rot;
-			fire_info.burst_index = old_burst_counter;
+			int current_burst_index = old_burst_counter * wip->shots;
+			for (int i = 0; i < wip->shots; i++) {
+				beam_fire_info fire_info;
 
-			// fire a beam weapon
-			weapon_objnum = beam_fire(&fire_info);
+				// stuff beam firing info
+				memset(&fire_info, 0, sizeof(beam_fire_info));
+				fire_info.accuracy = 1.0f;
+				fire_info.beam_info_index = turret_weapon_class;
+				fire_info.beam_info_override = nullptr;
+				fire_info.shooter = &Objects[parent_objnum];
+				fire_info.target = turret_enemy_objp;
+				if (wip->wi_flags[Weapon::Info_Flags::Antisubsysbeam])
+					fire_info.target_subsys = turret->targeted_subsys;
+				else
+					fire_info.target_subsys = nullptr;
+				fire_info.turret = turret;
+				fire_info.burst_seed = old_burst_seed;
+				fire_info.fire_method = BFM_TURRET_FIRED;
+				fire_info.per_burst_rotation = swp->per_burst_rot;
+				fire_info.burst_index = current_burst_index;
 
-			if (weapon_objnum != -1) {
-				objp = &Objects[weapon_objnum];
+				// fire a beam weapon
+				weapon_objnum = beam_fire(&fire_info);
 
-				parent_ship->last_fired_turret = turret;
-				turret->last_fired_weapon_info_index = turret_weapon_class;
-				turret->turret_last_fired = _timestamp();
+				if (weapon_objnum != -1) {
+					objp = &Objects[weapon_objnum];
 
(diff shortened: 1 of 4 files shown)


Baezon (approved):

code/ai/aiturret.cpp
```diff
@@ -1781,45 +1781,49 @@ bool turret_fire_weapon(int weapon_num, ship_subsys *turret, int parent_objnum,
...
+				fire_info.target = turret_enemy_objp;
+				if (wip->wi_flags[Weapon::Info_Flags::Antisubsysbeam])
+					fire_info.target_subsys = turret->targeted_subsys;
+				else
+					fire_info.target_subsys = nullptr;
+				fire_info.turret = turret;
+				fire_info.burst_seed = old_burst_seed;
+				fire_info.fire_method = BFM_TURRET_FIRED;
+				fire_info.per_burst_rotation = swp->per_burst_rot;
+				fire_info.burst_index = current_burst_index;
```
Goober5000, line 1804: This could be simplified to something like
```
fire_info.burst_index = old_burst_counter * wip->shots + i;
```
Then you wouldn't need the separate `current_burst_index` variable.  This helps with readability because my initial impression was that `current_burst_index` was unchanged for every iteration of the loop.

code/parse/sexp.cpp
```diff
@@ -21375,16 +21374,20 @@ void sexp_beam_fire(int node, bool at_coords)
...
-	} else {
-		// it would appear the turret doesn't have any beam weapons
-		Warning(LOCATION, "Couldn't fire turret on ship %s; subsystem %s has no beam weapons", CTEXT(node), CTEXT(CDR(node)));
-	}
+		// fire the beam
+		if (fire_info.beam_info_index != -1) {
+			fire_info.fire_method = BFM_TURRET_FORCE_FIRED;
+			int current_burst_index = 0;
+			for ( int i = 0; i < Weapon_info[fire_info.beam_info_index].shots; i++ ) {
+				fire_info.burst_index = current_burst_index;
```
Goober5000, line 21382: Here `current_burst_index` will always be the same as `i`, so you can do 
```
fire_info.burst_index =i;
```
and again you wouldn't need a separate `current_burst_index` variable.

code/ship/ship.cpp
```diff
@@ -12701,16 +12669,49 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
...
+
+						if (aip->target_objnum >= 0) {
+							fbfire_info.target = &Objects[aip->target_objnum];
+						} else {
+							fbfire_info.target = NULL;
+						}
+						fbfire_info.turret = &shipp->fighter_beam_turret_data;
+						fbfire_info.bfi_flags = BFIF_IS_FIGHTER_BEAM;
+						fbfire_info.bank = bank_to_fire;
+						fbfire_info.burst_index = current_burst_index;
```
Goober5000, line 12701: Same comment from aiturret.cpp applies here.

code/ship/ship.cpp
```diff
@@ -12636,23 +12636,21 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
...
-						points = num_slots;
-					}else{
-						points = winfo_p->b_info.beam_shots;
-					}
+					numtimes = winfo_p->shots;
+					points = MIN(winfo_p->b_info.beam_shots, num_slots);
 				} else if (winfo_p->wi_flags[Weapon::Info_Flags::Cycle]) {
-					points = 1;
+					numtimes = 1;
+					points = MIN(num_slots, winfo_p->shots);
```
Goober5000, line 12647: Is this the right thing to do here?

code/ai/aiturret.cpp
Kestrellius: Hmm. My inclination was to set it up outside the loop, so that we only have to do the multiplication once, but plausibly readability is more important here than that small bit of performance. I'll make the change if you think it's for the best.

code/ship/ship.cpp
Kestrellius: According to the current way "cycle" works, yes. With "cycle" enabled, $Shots doesn't change the number of shots generated by an individual firing event; instead, it controls the number of firepoints from which the weapon fires. So you can have a bank with 4 firepoints, and a weapon with a $Shots of 3 and the "cycle" flag, and what'll happen is that when the weapon fires, it'll fire from 3 firepoints, then cycle to the next set of three on the next shot.

I think that this is maybe not a great way to handle the behavior and we should possibly think about changing it, but that's beyond the scope of this feature.

code/ai/aiturret.cpp
JohnAFernandez: The optimization you're trying might also be done by the compiler.

Kestrellius:
I made the changes @Goober5000 requested. Still not certain that avoiding the current_burst_index variable is worth the performance cost of having to do the multiplication every loop.

code/ai/aiturret.cpp
Kestrellius: Makes sense!

Baezon:
I can assure you the performance cost is going to be completely negligible. you would have to try several orders of magnitude harder to have any noticeable performance impact. Readability is 100% the only concern here.

Kestrellius:
All right!

Goober5000:
Yup, what he said.

Goober5000 (approved):

PR #6182, opened 2024-06-04
Assorted beam fixes

Baezon:
A few fixes here

- Fighters use a fake turret for fighter beams including filling in `turret_fov`, possibly based on the weapon's field of fire. `turret_fov` is *only* used by the beam code in order to prematurely terminate a beam that has gone out of the turret's fov. This should *never* apply to fighter beams, because it makes no sense.
The fake turret should have an fov of -1 (for 360 coverage) and the code that checks against the fov shouldn't bother with checking fighter beams in any case.
- `beam_aim` early returns for non-targeting beams without coordinates or a target. This would apply to omni beams fired without a target, which is perfectly fine! Nor should `beam_get_binfo` early return in similar circumstances.
- Relatedly, the `Force_beam_turret_fov` is really important! Without it, moving subojects will be disregarded, and other fovs like elevation and base fov will be disregarded. It's a no brainer fix, and should target new versions.
- Exceeding -PI2 for "Per Burst Rotation" is not a concern, all negative values are interpreted as random.

### DIFF
--- a/code/ai/ai_profiles.cpp
+++ b/code/ai/ai_profiles.cpp
@@ -836,5 +836,6 @@ void ai_profile_t::reset()
 	}
 	if (mod_supports_version(24, 2, 0)) {
 		flags.set(AI::Profile_Flags::Debris_respects_big_damage);
+		flags.set(AI::Profile_Flags::Force_beam_turret_fov);
 	}
 }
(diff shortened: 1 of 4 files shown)


Goober5000 (approved):
Ok, makes sense and code all looks good.

JohnAFernandez (approved):
Look good to me too.

Issue #6117, opened 2024-04-30
FR: Multi-line custom data values


EatThePath:
The "Custom data" table data now available in many tables appears to be restricted to a single line per value. It'd be useful to me if there were a syntax for multi-line values. I'm not sure currently what the best approach for this would be, and it's not a blocker for anything I'm doing, so I'm throwing it in the pile in case anyone else is inspired to handle it.

MjnMixael:
The very design of the Custom Data methods precludes multi line just by the nature of how it was designed within FSO's parsing. That's why I developed Custom Strings for missions. It'd probably be easier to bring Custom Strings over to other places where Custom Data is used rather than forcing multiline to work here.

EatThePath:
It'd be nice from a user perspective to have both under the banner but I'm not particularly fussed by it if it has to be a separate section with a different name, whatever works so long as it's readable by scripts in a broadly similar manner.

MjnMixael:
The problem is FSO's parsing, specifically that tokens end at the end of the line unless you're in the special multiline parser that reads until it finds `$end_multi_text`. To add that to existing Custom Data risks breaking the parser because, iirc, it's a custom parser that looks for "+Val:" then reads until whitespace. That becomes the key. Everything after that to the end of the line becomes the value.

To add multiline here would require that the parsing change from

> $Custom Data:
> 	+Val: Headlights 6 7
> 	+Val: allowedCMs AN/ALE-650, DIRCM BriteBurst
> 	+Val: faction GTA
> $end_custom_data

to 

> $Custom Data:
> 	+Val: Headlights 6 7
> 	$end_multi_text
> 	+Val: allowedCMs AN/ALE-650, DIRCM BriteBurst
> 	$end_multi_text
> 	+Val: faction GTA
> 	$end_multi_text
> $end_custom_data

Which I think will break existing iterations. I don't think there's a foolproof way to read until end-of-line OR $end_multi_text.

Bringing Custom Strings over is a bit more future proof because you get a key, a value, and a multiline string which makes it ideal especially for XSTRs. Take this example from BtA2. The script gets all three strings and is able to quickly and efficiently translate the string with the value as the XSTR ID.

> $begin_custom_strings
> 
> $Name: LokiString
> +Value: 11372
> +String: Reserved for recon use
> $end_multi_text
> 
> $Name: ZeusString
> +Value: 11373
> +String: Reserved for assault use
> $end_multi_text
> 
> $end_custom_strings

[...]

PR #6184, opened 2024-06-10
Custom Strings for ships and weapons

MjnMixael:
Implements a request from #6117

### DIFF
--- a/code/scripting/api/objs/shipclass.cpp
+++ b/code/scripting/api/objs/shipclass.cpp
@@ -1173,6 +1173,66 @@ ADE_FUNC(hasCustomData, l_Shipclass, nullptr, "Detects whether the ship class ha
 	return ade_set_args(L, "b", result);
 }
 
+ADE_VIRTVAR(CustomStrings,
+	l_Shipclass,
+	nullptr,
+	"Gets the indexed custom string table for this ship. Each item in the table is a table with the following values: "
+	"Name - the name of the custom string, Value - the value associated with the custom string, String - the custom "
+	"string itself.",
+	"table",
+	"The ship's custom data table")
+{
+	int idx;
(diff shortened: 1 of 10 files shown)


code/scripting/api/objs/shipclass.cpp
```diff
@@ -1173,6 +1173,66 @@ ADE_FUNC(hasCustomData, l_Shipclass, nullptr, "Detects whether the ship class ha
...
+
+	return ade_set_args(L, "t", &table);
+}
+
+ADE_FUNC(hasCustomStrings,
+	l_Shipclass,
+	nullptr,
+	"Detects whether the ship has any custom strings",
+	"boolean",
+	"true if the ship's custom_strings is not empty, false otherwise")
```
JohnAFernandez, line 1221: Suggestion "true if the ship has custom strings, false otherwise" ?

code/scripting/api/objs/weaponclass.cpp
```diff
@@ -621,6 +621,66 @@ ADE_FUNC(hasCustomData, l_Weaponclass, nullptr, "Detects whether the weapon clas
...
+
+	return ade_set_args(L, "t", &table);
+}
+
+ADE_FUNC(hasCustomStrings,
+	l_Weaponclass,
+	nullptr,
+	"Detects whether the weapon has any custom strings",
+	"boolean",
+	"true if the weapon's custom_strings is not empty, false otherwise")
```
JohnAFernandez, line 669: Same here, I guess

JohnAFernandez (approved):

code/scripting/api/objs/shipclass.cpp
MjnMixael: The language here is the same as the other place where custom strings is used (mission) so I'm opting not to change it for consistency.

PR #6185, opened 2024-06-10
Assign hud in ships.tbl

MjnMixael:
Adds a method to assign HUD gauge configs in ships.tbl for each ship rather than using the $Ships() setting in hud_gauges.tbl. This has the distinct advantage of being more modular and a little more clear to how we generally table things in FSO.

Also fixes a sexp error checking bug that would flag any custom gauge not assigned to the player ship as possibly invalid despite the fact that the player ship could change during loadout. Here we should just validate that the gauge exists at all, not that it's assigned to a specific ship.

### DIFF
--- a/code/hud/hudparse.cpp
+++ b/code/hud/hudparse.cpp
@@ -394,7 +394,7 @@ void parse_hud_gauges_tbl(const char *filename)
 				}
 				break;
 			case 1:
-				mprintf(("$Ships in hud_gauges.tbl and -hdg.tbms is deprecated. Use \"$Name:\" and define Gauge Config Settings in ships.tbl per-ship instead.\n"));
+				mprintf(("$Ships in hud_gauges.tbl and -hdg.tbms can be replaced. Gauge Config Settings can now be defined per-ship in ships.tbl. \n"));
 
 				int shiparray[256];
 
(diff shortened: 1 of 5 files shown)


code/hud/hudparse.cpp
```diff
@@ -391,6 +394,8 @@ void parse_hud_gauges_tbl(const char *filename)
 				}
 				break;
 			case 1:
+				mprintf(("$Ships in hud_gauges.tbl and -hdg.tbms is deprecated. Use \"$Name:\" and define Gauge Config Settings in ships.tbl per-ship instead.\n"));
```
JohnAFernandez, line 397: Is this truly getting deprecated?  This sounds more like a QOL upgrade.

MjnMixael: Not really but when $ships replaced $ship for convenience a similar warning was added so I just followed the pattern. 🤷‍♂️

JohnAFernandez: Hmmmm ... I wish I had noticed that and said something. I don't want modders to think that the old method is disallowed if it's not....  

Maybe?

`$Ships in hud_gauges.tbl and -hdg.tbms can be replaced. Gauge Config Settings can now be defined per-ship in ships.tbl. \n`

I'm having trouble wording "disfavored" but not "disallowed" without sounding like I'm from 1850.

MjnMixael: Good enough for government work. Changed.

JohnAFernandez (approved):

PR #6186, opened 2024-06-11
Fix edge case in MSAA shader

BMagnu:
In rare cases before a background with a position value of 0,0,0 (i.e., model-less, emissive background), the MSAA shader could produce unwanted glints.
This was mostly apparent when using cockpits with deferred lighting and MSAA, as the entire background scene was rendered as a model-less emissive.

Previously, the shader would just hard-set the texel weight to 1 when the median distance was 0 to actually render-in background at dist 0, but evidently, it cannot set the weight to 1 for texels that don't conform to that median. By making sure that this only applies to texels whose individual distance is 0, similar to the median distance, this problem is avoided.

Also fix that the alpha channel of the normal buffer, used for gloss, should obviously not be part of the normalization process here.


wookieejedi (approved):
Thanks for tracking this down!

PR #6192, opened 2024-06-16
add support for PXO's UDP hole punch feature

notimaginative:
Should allow for most games, even those behind CG-NAT, to work without having to set up port forwarding. This set of changes is only needed for game servers to work with the new feature. No special handling is required for clients, so any build from the last few years should work for them.

### DIFF
--- a/code/network/gtrack.cpp
+++ b/code/network/gtrack.cpp
@@ -63,6 +63,7 @@ static unsigned int FirstGameOverPacket;
 static int SendingGameOver;
 //End New 7-9-98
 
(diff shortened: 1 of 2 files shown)


notimaginative:
This is known to not work with certain NAT types. It may also cause issues with clients being unable to see game servers on the same LAN (if using PXO). In such cases, if possible, standard port forwarding should still be done.

Also, if the `-prefer_ipv6` command line option is used by a client, then servers unable to use IPv6 will ignore the hole punch request.

JohnAFernandez (approved):
Pretty simple.  I guess most of the hard work was on the PXO side.

PR #6196, opened 2024-06-20
Fix Modern-style Sounds Using Number Indices

wookieejedi:
PR #6112 protected retail sound indices from being overwritten with the intent of when modders used string names for sounds,. Though that PR ended up being a bit overprotective because it made it so that using the modern sound format but still using numbered names did not work. In other words, for mods that use the modern sound format but still use numbered names, FSO sees the new format but then sees the numbers are within a protected range and offsets them, which means the sound index values referenced in ships and weapons table now are not correct and thus do not play.

This PR fixes that edge-case. Tested and works as expected. Also, if folks would rather just rever the entire protective offset aspect I am fine with that and we can instead just make a warning to the modder that their sound name will overwrite a protected range (though the changes of this PR to check if the name is the index should still be included, since an overwrite warning is not needed if the index and name are already the same).

### DIFF
--- a/code/gamesnd/gamesnd.cpp
+++ b/code/gamesnd/gamesnd.cpp
@@ -1260,23 +1260,25 @@ void parse_sound_table(const char* filename)
 					if (gamesnd_parse_line(&tempSound, "$Name:", &Snds))
 					{
 						int tempIndex = static_cast<int>(Snds.size());
+						bool name_not_match_index = (atoi(tempSound.name.c_str()) != tempIndex) && !tempSound.sound_entries.empty() && (tempSound.sound_entries[0].filename[0] != '\0');
 
 						if (tempSound.flags & GAME_SND_RETAIL_STYLE)
 						{
 							// retail sounds must have names that match their indexes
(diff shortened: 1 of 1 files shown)


JohnAFernandez (approved):
Lgtm

Goober5000 (changes requested):
I want to look at this closely and run some tests, so let's put this on hold for a bit.

JohnAFernandez:
So, this is actually a pretty simple change.  The logic is changed in only two places, the rest is whitespace or optimization.  Wookie has already tested the changes as well.  I just don't think there's much risk in merging it.  If you want to investigate if any further changes need to be made, couldn't you do that after this is merged?

Goober5000:
It's not a question of whether there will be any fixes needed after this PR, but whether this PR is in fact the correct fix.  It may pass wookieejedi's tests but it also needs to pass the Inferno tests that were the reason for the original PR.

My schedule had been pretty hectic the last few days but it's lighter today, so I expect to have an answer soon.

Goober5000:
Root cause identified.  Unfortunately, this PR only hides the problem; it does not solve it.  PR #6212 has the proper fix.

PR #6198, opened 2024-06-21
Fix ingame options init bugs

### LINKED ISSUE #6193: Persistent popup window from missing in-game options config
After #5917, launching the game with in-game options enabled, but without a config currently existing, will trigger a popup warning you to update your settings. This popup cannot be escaped and the game must be forcibly killed in order to exit.

Specifically the `Using_in_game_options` variable is true while the `Ingame_options_save_found` is false. As there is no variable set to indicate that the popup has been seen, it will be constantly triggered on every frame which prevents anything else from happening.

### LINKED ISSUE #6194: Crash bug when in-game options config is missing
After #5917, on a fresh install, the game will crash if no in-game options config exists.

This is due to the in-game options config files being under the `data` directory of the config path. However at the point that the config files are loaded this directory may not exist. As such the `fopen()` call in `osregistry.cpp:949` will fail since the full path to the file does not exist. And since there isn't a null check on the return value, the subsequent call to `fclose()` will trigger a crash.

MjnMixael:
Fixes #6193 by adding a bool check for if the popup has been displayed and fixes #6194 by adding a null check. If the file couldn't be created then default to the old fs2_open.ini instead.

### DIFF
--- a/code/menuui/playermenu.cpp
+++ b/code/menuui/playermenu.cpp
@@ -394,6 +394,7 @@ void player_select_init()
 
 // no need to reset this to false because we only ever see player_select once per game run
 static bool Startup_warning_dialog_displayed = false;
+static bool Save_file_warning_displayed = false;
 
 void player_select_do()
 {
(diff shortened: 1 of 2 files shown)


JohnAFernandez (approved):
Thanks!

PR #6199, opened 2024-06-21
Prevent SCPUI flicker on Win11

### LINKED ISSUE #6197: W11 and NVidia Cards cause screen flickering with SCPUI in Fullscreen
Very specific bug that still has a wide enough reach to be noticed by multiple users. As noted in title, screen flickering occurs when using SCPUI in Fullscreen on Windows 11 with Nvidia cards.

BMagnu:
This is kind of an ugly hack by forcing a backbuffer copy. The performance cost should be more or less negligible.
No idea _why_ it is a problem for us to directly render to the backbuffer, but it seems to be one.
As far as we're aware, the problem is limited to Win11 and possibly nvidia cards.
Similarly, _any_ sort of buffer between our rendering and the final thing that gets sent to the screen seems to fix the issue as well. This includes a buffer by the window compositor when FSO is windowed, a buffer for overlays by renderdoc and similar tools, our own buffering, etc.
Fixes #6197.

### DIFF
--- a/code/graphics/2d.cpp
+++ b/code/graphics/2d.cpp
@@ -1601,6 +1601,17 @@ bool gr_init(std::unique_ptr<os::GraphicsOperations>&& graphicsOps, int d_mode, 
 		}
 	}
 
+	if (!gr_is_viewport_window() && !Cmdline_window_res.has_value()) {
+		// For whatever reason, it seems that a combination of Win 11 and presumably NVidia GPU's causes weird artifacts.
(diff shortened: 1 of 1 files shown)


wookieejedi:
Fixes things in my tests.

wookieejedi:
Given this is a graphics related point, would be good to have @The-E or @TRBlount input, too.

The-E (approved):

PR #6206, opened 2024-06-25
Don't kill paused messages

### LINKED ISSUE #6168: as.pauseVoiceMessages() doesn't always unpause messages that were playing
Basically that. Pausing works. Unpausing sometimes doesn't work. I'm guessing the files reached some sort of timer and were released or something.

MjnMixael:
Fixes #6168 by adding an explicit check for paused messages. If they are paused then we shouldn't kill them.

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -1449,6 +1449,11 @@ void message_queue_process()
 			if ((Playing_messages[i].wave.isValid()) && (snd_time_remaining(Playing_messages[i].wave) > 250))
 				wave_done = 0;
 
+			// Don't kill paused messages
+			if ((Playing_messages[i].wave.isValid()) && snd_is_paused(Playing_messages[i].wave)) {
+				wave_done = 0;
+			}
+
 			// Goober5000
 			if (fsspeech_playing())
 				wave_done = 0;


code/mission/missionmessage.cpp
```diff
@@ -1449,6 +1449,11 @@ void message_queue_process()
 			if ((Playing_messages[i].wave.isValid()) && (snd_time_remaining(Playing_messages[i].wave) > 250))
 				wave_done = 0;
 
+			// Don't kill paused messages
+			if ((Playing_messages[i].wave.isValid()) && snd_is_paused(Playing_messages[i].wave)) {
```
wookieejedi, line 1453: Nice fix, sorry to be nitpicky but if you feel so inclined this could be an or check with the above if, though I reckon you kept them separate for readability.

MjnMixael: I did that originally but yeah I thought this was easier to read and there was already precedent for multiple statements that set `wave_done = 0`

wookieejedi: I'm good with it as is, if someone else really wants to have it be an OR then they can let us know.

JohnAFernandez (approved):
I'm good with it separate because there's already another similar check lower.  Compiler will probably combine them.

wookieejedi (approved):

PR #6207, opened 2024-06-25
Expose method to kill all current voice messages

MjnMixael:
Add a way for Lua to stop all currently playing voice messages. This is handy for checkpoints, for example; being able to stop messages when a checkpoint is selected.

Also removes an Assert in favor of a return. I don't see a reason an Assert is needed in the kill function when instead it can just do nothing if there's nothing to do.

### DIFF
--- a/code/mission/missionmessage.cpp
+++ b/code/mission/missionmessage.cpp
@@ -959,7 +959,9 @@ void message_kill_all( int kill_all )
 {
 	int i;
 
-	Assert( Num_messages_playing );
+	if (Num_messages_playing <= 0) {
+		return;
+	}
 
 	// kill sounds for all voices currently playing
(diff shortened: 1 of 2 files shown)


JohnAFernandez (approved):
Agreed.  I am going to make an issue for the remaining calls for message_kill_all, because they were paranoid and created duplicate checks to Num_messages_playing

MjnMixael:
> Agreed. I am going to make an issue for the remaining calls for message_kill_all, because they were paranoid and created duplicate checks to Num_messages_playing

Good idea. Happy to tackle that once this is merged, too.

JohnAFernandez:
I think your time is more valuable than that, and we can leave it as a possible first PR for someone.

wookieejedi:
Will plan to merge this later today unless someone else beats me to it.

PR #6212, opened 2024-06-27
properly handle game sounds with numeric IDs

Goober5000:
Patches the gamesnd system in several ways to properly fix the bug identified in #6196.  Due to the way the new sound system was written, sounds with numeric IDs were not necessarily added in their indexed location, even though the code assumed (even prior to #6112) that this would be the case.  The code now uses the proper index for both old *and* new style sounds.  This is done by ensuring, in `gamesnd_parse_entry()`, that the vector contains an entry for the specified numeric ID before doing a search for that ID.

Due to the slightly modified population of the gamesnd vectors, the `Snds_iface_handle` vector is now populated after parsing the various sounds.tbl files rather than during parsing.

Separately, this makes consistent the handling of the `sound_entries` vector which could potentially be empty but also treated an empty state as an assertion failure.

### DIFF
--- a/code/gamesnd/gamesnd.cpp
+++ b/code/gamesnd/gamesnd.cpp
@@ -770,8 +770,8 @@ void parse_gamesnd_soundset(game_snd* gs, bool no_create) {
 	// If this gets called then we just saw "+Entry:" so we can begin processing the first entry immediately
 
 	do {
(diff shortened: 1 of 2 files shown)


wookieejedi (approved):
Looks good and works in my tests

PR #6213, opened 2024-06-27
Cleanup message_kill_all

### LINKED ISSUE #6208: Cleanup: message_kill_all doesn't need to check Num_messages_playing
Agreed.  I am going to make an issue for the remaining calls for message_kill_all, because they were paranoid and created duplicate checks to Num_messages_playing

_Originally posted by @JohnAFernandez in https://github.com/scp-fs2open/fs2open.github.com/pull/6207#pullrequestreview-2139811574_

Basically, with 6207, we no longer need to obsessively check Num_messages_playing.  And previously, there was even some paranoid/duplicate checking.  This should be cleaned up for readability.

MjnMixael:
I originally wanted to do this in #6207 but decided to keep that PR trim.. so now I'm back to finish the job and satisfy my OCD a little bit.

This fixes #6208 by generally cleaning up the code surrounding message_kill_all; Change it to bool from int parameter, remove most checks for `Num_messages_playing` before calling `message_kill_all` now that it can just return instead of assert. There are a few places where I left in the additional checks to `Num_messages_playing` because they serve as nice optimization points to skip a whole bunch of other stuff that wouldn't be needed if no messages are playing; all of them part of `message_queue_process` in missionmessage.cpp.

### DIFF
--- a/code/io/keycontrol.cpp
+++ b/code/io/keycontrol.cpp
@@ -798,11 +798,8 @@ void process_debug_keys(int k)
 		// play the next mission message
 		case KEY_DEBUGGED + KEY_V:		
 			extern int Message_debug_index;
-			extern int Num_messages_playing;
 			// stop any other messages
-			if(Num_messages_playing){
-				message_kill_all(1);
-			}
+			message_kill_all(true);
 
 			// next message
 			if(Message_debug_index >= Num_messages - 1){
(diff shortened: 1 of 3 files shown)


wookieejedi:
Will plan to merge today or tomorrow morning unless someone else beats me to it.

PR #6214, opened 2024-06-27
Deprecate some commandlines and make some options not require restart

### LINKED ISSUE #6205: Commandline Cleanup
There are multiple options now that have been moved to In-Game Options menu which is now also available in retail without SCPUI. In general it is expected the only use of `-no_ingame_options` is for debugging purposes only.

With that in mind those options are ripe for being deprecated in the commandline entirely, especially ones that do not require a game restart. This list is not exhaustive and some of these may make sense to keep in commandline for troubleshooting or development purposes.

- [x] -bloom_intensity
- [x] -nomotiondebris
- [x] -nolightshafts
- [x] -warp_flash
- [x] -no_screenshake
- [x] -enable_3d_warp

MjnMixael:
24.2 introduces the in-game options menu officially and as long as we are retraining our player base we might as well rip the band-aid clean off.

This PR deprecates many of the commandlines in favor of using the in-game options menu. This only affects options that do not require a restart. Notably this PR also fixes a couple options that claimed to need a restart but actually don't.

Warp Flash and Motion Debris worked right away without requiring a restart. 3D Shockwave required a slight alteration to make sure the next mission doesn't use the previously loaded values and instead loads the new ones.

Fixes #6205


JohnAFernandez:
Code looks good afaict.  Going to hold off so that more eyes can get a look at it, just in case.

wookieejedi (approved):
Looks good to me as well

JohnAFernandez (approved):
Yeah, I think it's been long enough.

wookieejedi:
Yup, will plan to merge later today unless someone else beats me to it.

PR #6216, opened 2024-06-28
Clarify some SEXP operator documentation

naomimyselfandi:
Exactly what it says on the tin:

* Clarify some ambiguous documentation
* Warn about non-obvious edge cases, especially around `<argument>`s
* Remove a couple "generic he"s

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -37153,7 +37153,8 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 
 	{ OP_IS_SUBSYSTEM_DESTROYED_DELAY, "Is subsystem destroyed delay (Boolean operator)\r\n"
 		"\tBecomes true <delay> seconds after the specified subsystem of the specified "
-		"ship is destroyed.\r\n\r\n"
+		"ship is destroyed. Subsequent repairs to that subsystem have no effect on this "
+		"operator.\r\n\r\n"
 		"Returns a boolean value.  Takes 3 arguments...\r\n"
 		"\t1:\tName of ship the subsystem we are checking is on.\r\n"
 		"\t2:\tThe name of the subsystem we are checking status of.\r\n"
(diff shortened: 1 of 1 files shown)


code/parse/sexp.cpp
```diff
@@ -37153,24 +37153,28 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 
 	{ OP_IS_SUBSYSTEM_DESTROYED_DELAY, "Is subsystem destroyed delay (Boolean operator)\r\n"
 		"\tBecomes true <delay> seconds after the specified subsystem of the specified "
-		"ship is destroyed.\r\n\r\n"
+		"ship is destroyed. Subsequent repairs to that subsystem have no effect on this "
+		"operator.\r\n\r\n"
```
Baezon, line 37157: Perhaps specifying "is destroyed for the first time", would be even clearer. Ditto for the similar ones below.

code/parse/sexp.cpp
```diff
@@ -38736,9 +38735,7 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"\t2:\tA short string which will appear in the warning, or a message if the string matches a message name.\r\n"},
 
 	{ OP_GRANT_PROMOTION, "Grant promotion (Action operator)\r\n"
-		"\tIn a single player game, this function grants a player an automatic promotion to the "
-		"next rank which the player can obtain.  If he is already at the highest rank, this "
-		"operator has no effect.  It takes no arguments." },
+		"\tIn a single player game, this function grants a player an automatic promotion to the next rank which the player can obtain. If they are already at the highest rank, this operator has no effect. It takes no arguments." },
```
Baezon, line 38738: Even if it isn't necessarily consistent among the descriptions, keep the line breaks this description had. Ditto with set-subspace-drive below.

Baezon (changes requested):

code/parse/sexp.cpp
naomimyselfandi: Done. May I ask what the pattern is supposed to be for when I write operators in the future?

Baezon: The idea is that on smaller screens, it can be cumbersome to read the source code if a get line gets too long, I forget what the actual character number is, I know visual studio has a feature where till draw a little line on screen so someone knows not to write a line longer than that.

But overall, its a very minor concern, as mentioned, FSO isn't exactly consistent about it, and screens are pretty big these days, but its still a little easier to read.

Baezon (approved):

PR #6218, opened 2024-06-29
Adds conditional impact functionality.

Kestrellius:
This introduces a feature that allows weapons to have specialized impact particle effects that play only under particular conditions. Each conditional impact has an armor type for which it plays, and may also have minimum and maximum thresholds for health and hit angle. It can also be set as a dinky impact, to only play if the weapon isn't armed. On hit, all conditional impacts whose requirements are met will be played. Iff no conditional impacts are played, the normal impact effect or explosion will be played.

See demonstration:

[https://www.youtube.com/watch?v=WobH_JkeMAo](url)

Notes:

I use a short-circuiting logical AND to check whether the hit object exists and is a ship. If the hit object doesn't exist -- which is to say, it's a nullptr -- trying to access it (by checking whether it's a ship) causes a crash. Given that I have the check for whether it's null first, and the AND is short-circuiting, AFAICT this should never happen, but I wanted to check whether others considered this structure adequately safe. If it's not, I can just use nested if-clauses.

Possible future aspirations:

Contrary to what I previously thought, the logic for piercing impacts, which depends on armor type, does not take subsystem armor types into account -- only main hull and shield armor. I followed that pattern for this feature, but in the future it seems like it would be desirable to fix that.

Also, I might like to allow conditional impacts for things like weapons, debris, or asteroids. Probably the way to do this would be to allow those objects to have armor types.

### DIFF
--- a/code/weapon/weapon.h
+++ b/code/weapon/weapon.h
@@ -305,10 +305,10 @@ enum class HomingAcquisitionType {
 
 struct ConditionalImpact {
 	particle::ParticleEffectHandle effect;
-	float min_health_threshold_0_to_1;
-	float max_health_threshold_0_to_1;
-	float min_angle_threshold_degrees;
-	float max_angle_threshold_degrees;
+	float min_health_threshold; //factor, 0-1
+	float max_health_threshold; //factor, 0-1
+	float min_angle_threshold; //in degrees
+	float max_angle_threshold; //in degrees
 	bool dinky;
 };
 
--- a/code/weapon/weapon.h
+++ b/code/weapon/weapon.h
@@ -305,10 +305,10 @@ enum class HomingAcquisitionType {
 
 struct ConditionalImpact {
 	particle::ParticleEffectHandle effect;
-	float min_health_threshold;
-	float max_health_threshold;
-	float min_angle_threshold;
-	float max_angle_threshold;
+	float min_health_threshold_0_to_1;
+	float max_health_threshold_0_to_1;
+	float min_angle_threshold_degrees;
+	float max_angle_threshold_degrees;
 	bool dinky;
 };
 
--- a/code/weapon/weapon.h
+++ b/code/weapon/weapon.h
@@ -303,6 +303,15 @@ enum class HomingAcquisitionType {
 	RANDOM,
 };
 
+struct ConditionalImpact {
+	particle::ParticleEffectHandle effect;
+	float min_health_threshold;
+	float max_health_threshold;
+	float min_angle_threshold;
+	float max_angle_threshold;
+	bool dinky;
+};
+
 struct weapon_info
 {
 	char	name[NAME_LENGTH];				// name of this weapon
@@ -485,6 +494,8 @@ struct weapon_info
 
 	particle::ParticleEffectHandle piercing_impact_effect;
 	particle::ParticleEffectHandle piercing_impact_secondary_effect;
+
+	SCP_map<int, SCP_vector<ConditionalImpact>> conditional_impacts;
 
 	particle::ParticleEffectHandle muzzle_effect;
 
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -2356,20 +2356,20 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 	while (optional_string("$Conditional Impact Effect:")) {
 		int armor_index;
 		ConditionalImpact ci;
-		ci.min_health_threshold = 0.0;
-		ci.max_health_threshold = 1.0;
-		ci.min_angle_threshold = 0.0;
-		ci.max_angle_threshold = 180.0;
+		ci.min_health_threshold_0_to_1 = 0.0;
+		ci.max_health_threshold_0_to_1 = 1.0;
+		ci.min_angle_threshold_degrees = 0.0;
+		ci.max_angle_threshold_degrees = 180.0;
 		ci.dinky = false;
 		required_string("+Armor Type:");
 			stuff_string(fname, F_NAME, NAME_LENGTH);
 		armor_index = armor_type_get_idx(fname);
 		required_string("+Effect Name:");
 			ci.effect = particle::util::parseEffect(wip->name);
-		parse_optional_float_into("+Min Health Threshold:", &ci.min_health_threshold);
-		parse_optional_float_into("+Max Health Threshold:", &ci.max_health_threshold);
-		parse_optional_float_into("+Min Angle Threshold:", &ci.min_angle_threshold);
-		parse_optional_float_into("+Max Angle Threshold:", &ci.max_angle_threshold);
+		parse_optional_float_into("+Min Health Threshold:", &ci.min_health_threshold_0_to_1);
+		parse_optional_float_into("+Max Health Threshold:", &ci.max_health_threshold_0_to_1);
+		parse_optional_float_into("+Min Angle Threshold:", &ci.min_angle_threshold_degrees);
+		parse_optional_float_into("+Max Angle Threshold:", &ci.max_angle_threshold_degrees);
 		parse_optional_bool_into("+Dinky:", &ci.dinky);
 		SCP_vector<ConditionalImpact> ci_vec;
 		if (wip->conditional_impacts.count(armor_index) == 1) {
@@ -7532,7 +7532,7 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
 	vec3d reverse_incoming = (vmd_zero_vector - weapon_obj->orient.vec.fvec);
 
 	if (hitnormal) {
-		hit_angle = vm_vec_delta_ang(hitnormal, &reverse_incoming, NULL);
+		hit_angle = vm_vec_delta_ang(hitnormal, &reverse_incoming, nullptr);
 	}
 
 	if (other_obj != nullptr && other_obj->type == OBJ_SHIP) {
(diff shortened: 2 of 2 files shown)


code/weapon/weapons.cpp
```diff
@@ -7509,7 +7584,7 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
 		}
 
 		particleSource.finish();
-	} else if (wip->dinky_impact_weapon_expl_effect.isValid() && !armed_weapon) {
+	} else if (!valid_conditional_impact && wip->dinky_impact_weapon_expl_effect.isValid() && !armed_weapon && !valid_conditional_impact) {
```
JohnAFernandez, line 7587: Looks like `!valid_conditional_impact` is here twice

code/weapon/weapons.cpp
```diff
@@ -7497,7 +7524,55 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
 		weapon_hit_do_sound(other_obj, wip, hitpos, armed_weapon, quadrant);
 	}
 
-	if (wip->impact_weapon_expl_effect.isValid() && armed_weapon) {
+
+ 	bool valid_conditional_impact = false;
+	int relevant_armor_idx = -1;
+	float relevant_fraction = 1.0f;
+	float hit_angle = 0.0f;
+	vec3d reverse_incoming = (vmd_zero_vector - weapon_obj->orient.vec.fvec);
```
Baezon, line 7532: The unary operator (-) works on vectors! So you can write `vec3d reverse_incoming = -weapon_obj->orient.vec.fvec` which should be a little clearer.

code/weapon/weapons.cpp
```diff
@@ -7497,7 +7524,55 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
...
+
+	if (hitnormal) {
+		hit_angle = vm_vec_delta_ang(hitnormal, &reverse_incoming, NULL);
```
Baezon, line 7535: Always `nullptr` when you're adding it yourself (or touching code that used a `NULL`), or CI will complain at you.

code/weapon/weapons.cpp
```diff
@@ -7497,7 +7524,55 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
...
+	}
+
+	if (other_obj != nullptr && other_obj->type == OBJ_SHIP) {
```
Baezon, line 7538: To clarify, this is totally safe, used all over the codebase and very typical way to safely access the relevant fields.

code/weapon/weapons.cpp
```diff
@@ -7497,7 +7524,55 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
...
+		shipp = &Ships[other_obj->instance];
+		if (quadrant == -1) {
+			relevant_armor_idx = shipp->armor_type_idx;
+			relevant_fraction = other_obj->hull_strength / shipp->ship_max_hull_strength;
+		} else {
+			relevant_armor_idx = shipp->shield_armor_type_idx;
+			relevant_fraction = ship_quadrant_shield_strength(other_obj, quadrant);
+		}
+		if (wip->conditional_impacts.count(relevant_armor_idx) == 1) {
+			SCP_vector<ConditionalImpact> cis = wip->conditional_impacts[relevant_armor_idx];
```
JohnAFernandez, line 7548: You could make this a `const &auto` reference, which avoids copying the vector over, and makes it a little less busy.  `const &SCP_vecor<ConditionalImpact>` is acceptable as well.

Kestrellius:
Okay, never mind, it looks like all the armor logic is in ship.cpp? So trying to apply it to anything else would be a huge pain. Maybe a "default" armor type that's stored as -1 or something, which is used any time the weapon hits something that doesn't have an armor type.

code/weapon/weapons.cpp
JohnAFernandez, line 7538: This it the short circuiting you're talking about?  It looks fine to me.  It's just a general safety thing to check that the object is not null before dereferencing, tbh.  But if it's already guaranteed to be a non-null pointer, then you don't need to check.

JohnAFernandez (changes requested):

code/weapon/weapons.cpp
Kestrellius: Good catch! Thank you.

code/weapon/weapons.cpp
JohnAFernandez: Edit, I'm tired and I think it's `const auto&` and `const &SCP_vector<CondintionalImpact>

code/weapon/weapons.cpp
```diff
@@ -2353,6 +2353,33 @@ int parse_weapon(int subtype, bool replace, const char *filename)
...
+		armor_index = armor_type_get_idx(fname);
+		required_string("+Effect Name:");
+			ci.effect = particle::util::parseEffect(wip->name);
+		parse_optional_float_into("+Min Health Threshold:", &ci.min_health_threshold);
+		parse_optional_float_into("+Max Health Threshold:", &ci.max_health_threshold);
+		parse_optional_float_into("+Min Angle Threshold:", &ci.min_angle_threshold);
+		parse_optional_float_into("+Max Angle Threshold:", &ci.max_angle_threshold);
+		parse_optional_bool_into("+Dinky:", &ci.dinky);
+		SCP_vector<ConditionalImpact> ci_vec;
+		if (wip->conditional_impacts.count(armor_index) == 1) {
```
Baezon, line 2375: I was going to balk at `count() == 1` being used to determine if a key exists but apparently that's the typical way of doing this!

code/weapon/weapon.h
```diff
@@ -303,6 +303,15 @@ enum class HomingAcquisitionType {
 	RANDOM,
 };
 
+struct ConditionalImpact {
+	particle::ParticleEffectHandle effect;
+	float min_health_threshold;
+	float max_health_threshold;
+	float min_angle_threshold;
+	float max_angle_threshold;
```
Baezon, line 311: Minor, but these could benefit from units, 0-1 factor, and degrees, etc.

Baezon (changes requested):

Baezon:
> Okay, never mind, it looks like all the armor logic is in ship.cpp? So trying to apply it to anything else would be a huge pain. Maybe a "default" armor type that's stored as -1 or something, which is used any time the weapon hits something that doesn't have an armor type.

The _code_ handling armor types is in ship.cpp but that doesn't mean it only applies to ships. Weapons, for instance, also have armor types.

code/weapon/weapons.cpp
Kestrellius: As far as I can tell it's the _only_ way of doing this. I tried other options!

code/weapon/weapons.cpp
Kestrellius: According to VS, I can't.

`no operator "-" matches these operands`
`unary '-': 'vec3d' does not define this operator or a conversion to a type acceptable to the predefined operator`
`built-in operator '-' cannot be applied to an operand of type 'vec3d'`

code/weapon/weapon.h
Kestrellius: Good idea.

code/weapon/weapons.cpp
Baezon: My apologies, too much time working with pof tools where this is the case. Either way the convention is `vm_vec_negate()`
```
vec3d reverse_incoming = weapon_obj->orient.vec.fvec;
vm_vec_negate(&reverse_incoming);
```
It is a little more verbose this way, and this is verging on pedantic, but is more consistent with the rest of the codebase and more explicit.

code/weapon/weapon.h
```diff
@@ -303,6 +303,15 @@ enum class HomingAcquisitionType {
 	RANDOM,
 };
 
+struct ConditionalImpact {
+	particle::ParticleEffectHandle effect;
+	float min_health_threshold_0_to_1;
+	float max_health_threshold_0_to_1;
+	float min_angle_threshold_degrees;
+	float max_angle_threshold_degrees;
```
Baezon, line 311: Ah, to be more specific, I think it should be added as comments, you can leave the variable names as is, and indeed it is conventional to leave units out of the variable names (save for exceptional circumstances).

Baezon:
I can't add another comment for some reason, but if you'll indulge me to add one more note, it occurs to me that this should probably be called simply "$Conditional Impact", leaving off "effect", since I think the logical next step would be unique sound effects, which could be slotted in alongside, and by renaming it now we'll save ourselves the frustration later.

code/weapon/weapons.cpp
```diff
@@ -7497,7 +7524,55 @@ void weapon_hit( object * weapon_obj, object * other_obj, vec3d * hitpos, int qu
...
+		if (quadrant == -1) {
+			relevant_armor_idx = shipp->armor_type_idx;
+			relevant_fraction = other_obj->hull_strength / shipp->ship_max_hull_strength;
+		} else {
+			relevant_armor_idx = shipp->shield_armor_type_idx;
+			relevant_fraction = ship_quadrant_shield_strength(other_obj, quadrant);
+		}
+		if (wip->conditional_impacts.count(relevant_armor_idx) == 1) {
+			const auto* cis = &wip->conditional_impacts[relevant_armor_idx];
+			for (auto &ci : *cis) {
```
JohnAFernandez, line 7549: So I didn't mean a pointer.  Not that this code is particularly complicated, but when you type `const type*` only the pointer and not the contents are const.

But actually, if we're not going to use cis except as a bound for ci it might be best to do...

`for (const auto& ci : wip->conditional_impacts[relevant_armor_idx])

Kestrellius:
> I can't add another comment for some reason, but if you'll indulge me to add one more note, it occurs to me that this should probably be called simply "$Conditional Impact", leaving off "effect", since I think the logical next step would be unique sound effects, which could be slotted in alongside, and by renaming it now we'll save ourselves the frustration later.

Very good point.

Baezon (approved):

JohnAFernandez (approved):
Good work!  Make sure to run it through one more test as we made a lot of changes.

Kestrellius:
Yep, test seems to behave as expected.

PR #6219, opened 2024-07-01
enhance lock-perspective to also lock slew mode

Goober5000:
Adds an optional argument to the `lock-perspective` sexp to allow locking of the 'free look' or 'padlock' view.  This prevents the player from looking around which might not be desired in certain forced perspective situations.

### DIFF
--- a/code/io/keycontrol.cpp
+++ b/code/io/keycontrol.cpp
@@ -218,6 +218,7 @@ static struct Cheat cheatsTable[] = {
 
 int Tool_enabled = 0;
 bool Perspective_locked=false;
+bool Slew_locked=false;
 
 extern int AI_watch_object;
 extern int Countermeasures_enabled;
--- a/code/io/keycontrol.cpp
+++ b/code/io/keycontrol.cpp
@@ -217,8 +217,6 @@ static struct Cheat cheatsTable[] = {
 
 
 int Tool_enabled = 0;
-bool Perspective_locked=false;
-bool Slew_locked=false;
 
 extern int AI_watch_object;
 extern int Countermeasures_enabled;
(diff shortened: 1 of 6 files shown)


code/io/keycontrol.cpp
```diff
@@ -218,6 +218,7 @@ static struct Cheat cheatsTable[] = {
 
 int Tool_enabled = 0;
 bool Perspective_locked=false;
+bool Slew_locked=false;
```
BMagnu, line 221: Why is this in keycontrol and not playercontrol?

code/playerman/playercontrol.cpp
```diff
@@ -397,35 +397,37 @@ void do_view_slew()
 	view_modify(&chase_slew_angles, &Viewer_slew_angles_delta, PI_2, PI2/3);
 
 	// Check Track target
-	if (Viewer_mode & VM_TRACK) {
+	if (Viewer_mode & VM_TRACK && !Slew_locked) {
```
BMagnu, line 400: The fact that slew-locked prevents tracking should also be documented somewhere.

BMagnu (changes requested):

code/playerman/playercontrol.cpp
Goober5000: `VM_TRACK` is a variant of padlock mode, so I figured it would be implicit.  But I can make it implicit.

code/io/keycontrol.cpp
Goober5000: Simply because `Perspective_locked` is also here.  If there's a better place to put them, then I'd like to move both of them together.

code/playerman/playercontrol.cpp
BMagnu: That may not be evident to some modders, so it might be good to be explicit here

BMagnu (approved):

Issue #6210, opened 2024-06-25
FR: Streamline definition of turret AI in FRED


Durandal-Thoth:
With current architecture, turret AI defaults to ship table default, meaning if you wish to change a ship's AI in FRED then this must be done manually through a series of dropdown menus with each ship as well. This can get quite tedious on ships with large numbers of turrets.

Personally suggest a checkbox in loadout screen for turret AI to inherent parent ship AI.

Baezon:
Well, not that this necessarily obviates your request, but you'll probably want to know the FSO *does not* use the turret's AI class for anything. All turret behavior modified by AI class uses only the ship's overall AI class.

Durandal-Thoth:
May as well expand this in a request for aip defining different stats per turret AI level then I guess.

Goober5000:
It's best if each GitHub issue only covers a single item.

PR #6221, opened 2024-07-03
create a `process_ship_subsystems` routine for subsystem operations

Goober5000:
Many SEXP operators that work on subsystems use very similar code.  This new function allows the operator implementations to be streamlined and makes future modifications easier.  For example, the `change-ai-class` operator now works with names such as `<all turrets>` as well as specific subsystems.

Inspired by #6210 although not a direct implementation of it.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -5542,6 +5542,80 @@ player *get_player_from_ship_node(int node, bool test_respawns = false, int *net
 	return get_player_from_ship_entry(ship_entry, test_respawns, netplayer_index);
 }
 
+/*
+ * Iterates through a subsystem list and applies the given function to each subsystem, handling the "hull" subsystem as well as (optionally) generic subsystems.  The function's prototype should be:
+ * 
+ * void f(bool is_hull, bool is_sim_hull, bool is_shields, ship_subsys *ss);
+ * 
+ * If any of the booleans are true, ss will be nullptr.
+ */
+template <typename Function>
+void process_ship_subsystems(const ship_registry_entry *ship_entry, bool allow_generic_subsystems, int node, bool iterate_nodes, const char *sexp_operator_name, Function f)
+{
+	Assertion(sexp_operator_name, "sexp_operator_name must not be null!");
+	if (node < 0)
+		return;
+
+	if (!ship_entry || !ship_entry->has_shipp())
+		return;
+	auto shipp = ship_entry->shipp();
+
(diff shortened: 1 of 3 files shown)


code/parse/sexp.cpp
```diff
@@ -5541,6 +5541,80 @@ player *get_player_from_ship_node(int node, bool test_respawns = false, int *net
...
 
+/*
+ * Iterates through a subsystem list and applies the given function to each subsystem, handling the "hull" subsystem as well as (optionally) generic subsystems.  The function's prototype should be:
+ * 
+ * void f(bool is_hull, bool is_sim_hull, bool is_shields, ship_subsys *ss);
+ * 
+ * If any of the booleans are true, ss will be nullptr.
+ */
+template <typename Function>
+void process_ship_subsystems(const ship_registry_entry *ship_entry, bool allow_generic_subsystems, int node, bool iterate_nodes, const char *sexp_operator_name, Function f)
```
BMagnu, line 5552: Given the function signature is constant, I'd prefer f to be of type `std::function<void(bool, bool, bool, ship_subsys*)>`.
On second thought, the first three bool parameters appear to be mutually exclusive, so I'd even more prefer for this to be something like an
``enum class SubsystemType : uint8_t {SUBSYSTEM, HULL, SIM_HULL, SHIELDS}``

code/parse/sexp.cpp
```diff
@@ -15722,60 +15679,32 @@ void sexp_destroy_subsys_instantly(int n)
...
-				}
-			}
-		}
-		// normal subsystems
-		else
+	// process the subsystems, which could be hull, sim hull, or a generic type like <all engines>
+	process_ship_subsystems(ship_entry, true, n, true, "destroy-subsys-instantly", [&](bool, bool, bool, ship_subsys *ss)
+	{
+		// this sexp is only for subsystems
+		if (ss)
```
BMagnu, line 15688: This is kind of a weird way of checking that the first three bools are false, and nonintuitive in a sense.

code/parse/sexp.cpp
BMagnu: Also, for later:
Given the frequency of the `any_subsystem` check, how about checking that within `process_ship_subsystems` and returning that as a bool? That way you can get around the need to juggle this bool around the lambdas in the invocation.

BMagnu (changes requested):

BMagnu (approved):

Issue #6010, opened 2024-02-27
FSO sometimes takes an extremely long time to parse mv_effects-part.tbm


MjnMixael:
I think everyone noticed with the latest mediavps that the splash screen can sometimes seem to stall or take a long time. I noticed that it's always during parsing of mv_effects-part.tbm. The file is only 3200 lines and nothing immediately stands out as an error in it to me.

BMagnu:
Given that we've done rudimentary improvements, and are not sure why this takes much longer on specific machines, we're punting further improvements to 25.0 to get more time for analysis.

BMagnu:
As the particle system has been entirely reworked for 25.0 (#6465), is this slowdown still the case on current master?

PR #6223, opened 2024-07-03
address 6010

Goober5000:
I took a close look at the code involved in parsing particle tables.  There is no smoking gun, but there are several ways to polish the code which I have implemented here.  The most likely culprits for the slowdown, IMO, are the repeated conversions between SCP_string and C-string, especially in `ParticleManager::getEffectByName()`; followed by the omission of `std::move` in several places.

Addresses #6010.


wookieejedi (approved):
LGTM

PR #6224, opened 2024-07-03
fix coverity task

Goober5000:
Per @notimaginative, change the `compiler` variable to uppercase to match its usage in the rest of the file.  This should allow the coverity task to start running again.


JohnAFernandez (approved):
Thank you!

PR #6229, opened 2024-07-04
Fix `short read` error with red alert csg file reading

wookieejedi:
In current master, the debug log commonly shows `CSG => Warning: (0x0011) Short read, information may have been lost!`.

Thanks to @notimaginative for pointing the following issue out: The function `pilotfile::csg_read_redalert()` returns if `ship_list_size` is zero, but it still needs to check the `wing_list_size` a bit below that. This has likely not caused data loss issues ATM, because it's reasonable to assume that if `ship_list_size` is zero then so is `wing_list_size`. But it's still a bug and should be fixed before it does cause some real data loss.

This PR moves the relevant read sections to within each respective if-block. That way the function is capable of checking both `ship_list_size` and `wing_list_size` on every load (even if `ship_list_size` is zero).

### DIFF
--- a/code/pilotfile/csg.cpp
+++ b/code/pilotfile/csg.cpp
@@ -812,84 +812,82 @@ void pilotfile::csg_read_redalert()
 
 	ship_list_size = cfread_int(cfp);
 
-	if (ship_list_size <= 0) {
-		return;
-	}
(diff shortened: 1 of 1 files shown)


Goober5000 (approved):
Good catch 👍

PR #6232, opened 2024-07-05
Several coverity fixes

Goober5000:
Fix, or mark as false positive, several issues flagged by Coverity.
* using `std::move` instead of copying, and change some parameters from value to reference
* fix or mark potential copy-paste mistakes
* various checking of parameter and return values
* miscellaneous fixes, and change some 'intentional' markers to 'false positive' markers


JohnAFernandez (approved):
LGTM

PR #6233, opened 2024-07-05
Fix Insignia Checkbox in the Lab

### LINKED ISSUE #6062: Show Insignia checkbox in the F3 Lab does nothing
The insignia checkbox neither disables or enables insignia. While I'm at it, it'd be nice to be able to select particular insignia in order to test-view them like team colors

JohnAFernandez:
Add a ship flag and an MR flag to turn rendering on and off.  Also standardize the type for MR flags to be uint64_t so that the new flag will fit.

Fixes #6062 

Note: (1<<32) and above causes UB, so I just put the value and put the bitshift in the comments.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -7351,7 +7351,7 @@ void HudGaugeHardpoints::render(float  /*frametime*/)
 	vertex draw_point;
 	vec3d subobj_pos;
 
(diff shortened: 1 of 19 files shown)


Goober5000 (approved):
Eventually this should probably move to `flagset<>`, but this is a reasonable intermediate step.  There are also a few `MR_EMPTY_SLOT*` flag definitions that can be filled in with new flags.

PR #6235, opened 2024-07-06
Allow accepting a pilot without changing gamestate

MjnMixael:
There are some actions in SCPUI barracks menu that can cause Player to get set to nullptr and there's currently no way to accept a player without also returning to the mainhall. This should never have been a combined action but m!m didn't have that kind of foresight in 2018.. so here we are.

An optional parameter that allows skipping returning to the mainhall as part of this function which is only used by the API. This will allow me to fix a bug in SCPUI where Player gets left as nullptr.

I should clarify that the only code that runs `barracks_accept_pilot` is the Lua API. The retail UI runs `barracks_pilot_accepted` instead.

### DIFF
--- a/code/menuui/barracks.cpp
+++ b/code/menuui/barracks.cpp
@@ -694,7 +694,7 @@ int barracks_pilot_accepted()
 	return 0;
 }
 
-void barracks_accept_pilot(player* plr) {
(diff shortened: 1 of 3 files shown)


BMagnu (approved):

wookieejedi:
Will plan to merge this later today unless someone else beats me to it.

PR #6237, opened 2024-07-07
use proper stats block for display in barracks

notimaginative:
There are 3 different stats blocks: two in the pilot file for single and multi, and one in the campaign savefile. The stats block in the savefile has historically been used in the barracks if possible since it's sanitized for the current mod data. Unfortunately this means that the stats will change if the campaign changes or is restarted.

However it is possible to easily produce a sanitized scoring struct based on the stats blocks from the pilot file. This allows us to use true all-time stats for display in the barracks while still accounting for mod differences.

### DIFF
--- a/code/menuui/barracks.cpp
+++ b/code/menuui/barracks.cpp
@@ -568,14 +568,12 @@ int barracks_new_pilot_selected()
 		return -1;
 	}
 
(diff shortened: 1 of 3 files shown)


notimaginative:
Still needs additional testing as I had a very limited data set to test with.

Also note that pilot and squad images are not covered by this change so it's possible for them to either not render or use a different image than when is expected with different mods. This should be non-fatal and really just a display glitch.

PR #6239, opened 2024-07-07
Respect tabled delay for arrival and backup messages

### LINKED ISSUE #5427: Add Timestamp Check to Friendly Wing Arrival Messages
Currently there is not timestamp check for sending Friendly Arrival Messages for Wings (though there is for Hostile Arrival Wings). So, if two or more wings arrive at the exact same time then multiple overlapping arrival message will be sent.

EDIT:
@naomimyselfandi has kindly stated she is interested in creating a fix for this issue, which is very appreciated!

naomimyselfandi:
Closes #5427.

### DIFF
--- a/code/mission/missionmessage.h
+++ b/code/mission/missionmessage.h
@@ -115,7 +115,7 @@ X(SUPPORT_KILLED,      "Support Killed",       100, -1,  0,     HIGH, SOON, NONE
                                                                                                  \
 /* Global status */                                                                              \
 X(ALL_ALONE,           "All Alone",             50, -1,  0,     HIGH, ANYTIME, NONE),            \
-X(ARRIVE_ENEMY,        "Arrive Enemy",         100, -1,  0,     LOW, SOON, NONE),                \
+X(ARRIVE_ENEMY,        "Arrive Enemy",         100, -1,  30000, LOW, SOON, NONE),                \
 X(OOPS,                "Oops 1",               100, -1,  0,     HIGH, ANYTIME, NONE),            \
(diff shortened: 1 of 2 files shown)


JohnAFernandez (approved):
LGTM.  I was going to ask about how this affects other mods, but these are builtin messages, and modders can always playtest the missions on the next stable.

naomimyselfandi:
> LGTM. I was going to ask about how this affects other mods, but these are builtin messages, and modders can always playtest the missions on the next stable.

Even so, I think that's a fair question, and it's one I did think about. The practical answer is that it has very little impact. For mods that have not tried to table a custom `+Minimum Delay:` for `Arrival` or `Backup` messages, the only impact here is that friendly wings arriving at the same time won't talk over each other. I legitimately can't imagine a scenario where them doing so would be desired, especially since it wouldn't work reliably; if the mission is trying to achieve a special effect, perhaps to convey a disorganized retreat or misjump, it would use `send-message`, which isn't impacted by this at all.

For mods that have tried to table a custom delay for one of those types, their custom delay will suddenly start being respected. The only ways I can imagine this being harmful is if the modder tabled a custom delay, noticed that it didn't seem to work, tabled a very large or very small delay to test that, and then didn't remove it; or they tabled a delay that "felt right" but was actually too large or too small and _didn't_ notice that it wasn't being respected. Given how niche of a feature customizing these delays is in the first place, both cases seem very unlikely.

wookieejedi (approved):
Looks good and works in my tests.

wookieejedi:
Will plan to merge this later today or tomorrow unless someone else beats me to it

PR #6241, opened 2024-07-08
optimize collision code a bit

Goober5000:
This was motivated by the recent Coverity report.  Coverity noticed that a few of the copy assignments could be replaced by move assignments, but on further investigation one copy each in `ship_weapon_check_collision()` and `beam_collide_ship()` can be entirely removed, another either/or copy in each can be replaced with an either/or pointer assignment, and a few copies can be done only when needed.  Another vec3d copy can be replaced with a vec3d pointer since there is no longer a need to keep a local copy.

### DIFF
--- a/code/object/collideshipweapon.cpp
+++ b/code/object/collideshipweapon.cpp
@@ -66,7 +66,7 @@ static void update_danger_weapon(object *pship_obj, object *weapon_obj)
  * Deal with weapon-ship hit stuff.  
  * Separated from check_collision routine below because of multiplayer reasons.
  */
-static void ship_weapon_do_hit_stuff(object *pship_obj, object *weapon_obj, vec3d *world_hitpos, vec3d *hitpos, int quadrant_num, int submodel_num, vec3d /*not a pointer intentionaly*/ hit_dir)
+static void ship_weapon_do_hit_stuff(object *pship_obj, object *weapon_obj, vec3d *world_hitpos, vec3d *hitpos, int quadrant_num, int submodel_num, const vec3d *hit_dir)
 {
 	weapon	*wp = &Weapons[weapon_obj->instance];
 	weapon_info *wip = &Weapon_info[wp->weapon_info_index];
@@ -77,7 +77,7 @@ static void ship_weapon_do_hit_stuff(object *pship_obj, object *weapon_obj, vec3
 	vec3d force;
 
 	vec3d worldNormal;
-	model_instance_local_to_global_dir(&worldNormal, &hit_dir, pm, pmi, submodel_num, &pship_obj->orient);
+	model_instance_local_to_global_dir(&worldNormal, hit_dir, pm, pmi, submodel_num, &pship_obj->orient);
 
 	// Apply hit & damage & stuff to weapon
 	weapon_hit(weapon_obj, pship_obj,  world_hitpos, quadrant_num, &worldNormal);
@@ -143,7 +143,7 @@ static void ship_weapon_do_hit_stuff(object *pship_obj, object *weapon_obj, vec3
 		model_instance_global_to_local_dir(&weapon_up, &weapon_obj->orient.vec.uvec, pm, pmi, submodel_num, &pship_obj->orient);
 
 		matrix decal_orient;
-		vm_vector_2_matrix_norm(&decal_orient, &hit_dir, &weapon_up); // hit_dir is already normalized so we can use the more efficient function
+		vm_vector_2_matrix_norm(&decal_orient, hit_dir, &weapon_up); // hit_dir is already normalized so we can use the more efficient function
 
 		// randomize angle of decal if specified --wookeejedi
 		if (wip->impact_decal.random_rotation) {
@@ -161,7 +161,7 @@ extern int Framecount;
 
 static int ship_weapon_check_collision(object *ship_objp, object *weapon_objp, float time_limit = 0.0f, int *next_hit = nullptr)
 {
-	mc_info mc, mc_shield, mc_hull;
+	mc_info mc_hull, mc_shield, *mc;
 	ship	*shipp;
 	ship_info *sip;
 	weapon	*wp;
@@ -223,17 +223,15 @@ static int ship_weapon_check_collision(object *ship_objp, object *weapon_objp, f
 
 	// Goober5000 - I tried to make collision code much saner... here begin the (major) changes
 
-	// set up collision structs
-	mc.model_instance_num = shipp->model_instance_num;
-	mc.model_num = sip->model_num;
-	mc.submodel_num = -1;
-	mc.orient = &ship_objp->orient;
-	mc.pos = &ship_objp->pos;
-	mc.p0 = &weapon_start_pos;
-	mc.p1 = &weapon_end_pos;
-	mc.lod = sip->collision_lod;
-	mc_shield = mc;
-	mc_hull = mc;
+	// set up collision struct
+	mc_hull.model_instance_num = shipp->model_instance_num;
+	mc_hull.model_num = sip->model_num;
+	mc_hull.submodel_num = -1;
+	mc_hull.orient = &ship_objp->orient;
+	mc_hull.pos = &ship_objp->pos;
+	mc_hull.p0 = &weapon_start_pos;
+	mc_hull.p1 = &weapon_end_pos;
+	mc_hull.lod = sip->collision_lod;
 
 	// (btw, these are leftover comments from below...)
 	//
@@ -260,6 +258,9 @@ static int ship_weapon_check_collision(object *ship_objp, object *weapon_objp, f
 
 	// check shields for impact
 	if (!(ship_objp->flags[Object::Object_Flags::No_shields])) {
+		// set up collision struct
+		mc_shield = mc_hull;
+
 		if (sip->flags[Ship::Info_Flags::Auto_spread_shields]) {
 			// The weapon is not allowed to impact the shield before it reaches this point
 			vec3d shield_ignored_until = weapon_objp->last_pos;
@@ -451,20 +452,22 @@ static int ship_weapon_check_collision(object *ship_objp, object *weapon_objp, f
 	// see which impact we use
 	if (shield_collision)
 	{
-		mc = mc_shield;
+		mc = &mc_shield;
 		Assert(quadrant_num >= 0);
 		valid_hit_occurred = 1;
 	}
 	else if (hull_collision)
 	{
-		mc = mc_hull;
+		mc = &mc_hull;
 		valid_hit_occurred = 1;
 	}
+	else
+		mc = nullptr;
 
 	// deal with predictive collisions.  Find their actual hit time and see if they occured in current frame
 	if (next_hit && valid_hit_occurred) {
 		// find hit time
-		*next_hit = (int) (1000.0f * (mc.hit_dist*(flFrametime + time_limit) - flFrametime) );
+		*next_hit = (int) (1000.0f * (mc->hit_dist*(flFrametime + time_limit) - flFrametime) );
 		if (*next_hit > 0)
 			// if hit occurs outside of this frame, do not do damage 
 			return 1;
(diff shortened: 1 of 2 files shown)


code/object/collideshipweapon.cpp
```diff
@@ -66,7 +66,7 @@ static void update_danger_weapon(object *pship_obj, object *weapon_obj)
  * Deal with weapon-ship hit stuff.  
  * Separated from check_collision routine below because of multiplayer reasons.
  */
-static void ship_weapon_do_hit_stuff(object *pship_obj, object *weapon_obj, vec3d *world_hitpos, vec3d *hitpos, int quadrant_num, int submodel_num, vec3d /*not a pointer intentionaly*/ hit_dir)
+static void ship_weapon_do_hit_stuff(object *pship_obj, object *weapon_obj, vec3d *world_hitpos, vec3d *hitpos, int quadrant_num, int submodel_num, vec3d *hit_dir)
```
BMagnu, line 69: If this can safely be pointered, then it should definitely also be consted to indicate that the vector isn't modified.

BMagnu (changes requested):

code/object/collideshipweapon.cpp
Goober5000: Agreed, but I figured I'd defer that to the upcoming PR dedicated to `const`.  I can still add `const` here if you'd rather I do this particular one now instead of later.

BMagnu: Mhm, it's fine if said const PR is gonna come soon.
Otherwise, having a non-const pointer is kinda ugly

Goober5000: Ok, done

BMagnu (approved):

qazwsxal:
Is there any profiling results to see what sort of difference this makes?

Goober5000:
I haven't run any.

qazwsxal:
Ok, so just to be clear, this hasn't been tested for performance regressions either? I've had pretty unintuitive results when it comes to collision code performance in the past.

code/object/collideshipweapon.cpp
```diff
@@ -66,7 +66,7 @@ static void update_danger_weapon(object *pship_obj, object *weapon_obj)
  * Deal with weapon-ship hit stuff.  
  * Separated from check_collision routine below because of multiplayer reasons.
  */
-static void ship_weapon_do_hit_stuff(object *pship_obj, object *weapon_obj, vec3d *world_hitpos, vec3d *hitpos, int quadrant_num, int submodel_num, vec3d /*not a pointer intentionaly*/ hit_dir)
+static void ship_weapon_do_hit_stuff(object *pship_obj, object *weapon_obj, vec3d *world_hitpos, vec3d *hitpos, int quadrant_num, int submodel_num, const vec3d *hit_dir)
```
MageKing17, line 69: If it used to not be a pointer intentionally, there needs to be *some* kind of comment explaining why changing it won't cause problems.

MageKing17 (changes requested):
One change needs further clarification; additionally, this *needs* performance testing to ensure it actually *is* an optimization, or at least not a regression.

EatThePath:
> One change needs further clarification; additionally, this _needs_ performance testing to ensure it actually _is_ an optimization, or at least not a regression.

Since I was pinged, I'll chime in to concur. Even changes to this area that aren't motivated by optimization really should have some benchmark comparisons to ensure there isn't a regression, it's just above one of our hottest loops and as @qazwsxal points out it's often unpredictable what will affect it, and how.

That said my gut is that this is will be an effectively null impact either way. This is near enough the hot loop to warrant caution but the real heat is inside model_collide, if memory serves. A few small copies more or less outside of it seems likely negligible.

I'm not really up for doing the testing myself, at the moment. If someone wants tips on how to do it, I can probably help with that, but I don't have free time I'd want to allocate to this in the immediate future.

code/object/collideshipweapon.cpp
Goober5000: That was what I alluded to here:
> Another vec3d copy can be replaced with a vec3d pointer since there is no longer a need to keep a local copy.

To elaborate, at one point in time `ship_weapon_do_hit_stuff()` modified `hit_dir`, so the parameter was copied in order to avoid modifying the vec3d from the calling function.  Later on the code that modified `hit_dir` was removed, but the parameter was not changed.

The `hit_dir` parameter, as well as the code that modified it, was added in commit a3422b4d01.  Here is the relevant part:
```C++
@@ -236,7 +240,7 @@ void update_danger_weapon(object *ship_obj, object *weapon_obj)

 // function to actually deal with weapon-ship hit stuff.  separated from check_collision routine below
 // because of multiplayer reasons.
-void ship_weapon_do_hit_stuff(object *ship_obj, object *weapon_obj, vector *world_hitpos, vector *hitpos, int quadrant_num, int submodel_num)
+void ship_weapon_do_hit_stuff(object *ship_obj, object *weapon_obj, vector *world_hitpos, vector *hitpos, int quadrant_num, int submodel_num, vector /*not a pointer intentionaly*/ hit_dir)
 {
        weapon  *wp = &Weapons[weapon_obj->instance];
        weapon_info     *wip = &Weapon_info[wp->weapon_info_index];
@@ -287,13 +291,19 @@ void ship_weapon_do_hit_stuff(object *ship_obj, object *weapon_obj, vector *worl
        }

        if(quadrant_num == -1){
-               /*weapon_info   *wip = &Weapon_info[Weapons[weapon_obj->instance].weapon_info_index];
+               weapon_info     *wip = &Weapon_info[Weapons[weapon_obj->instance].weapon_info_index];
                decal_point dec;
                dec.orient = weapon_obj->orient;
+               vector hit_fvec;
+               vm_vec_negate(&hit_dir);
+               vm_vec_avg(&hit_fvec, &hit_dir, &weapon_obj->orient.vec.fvec);
+               vm_vec_normalize(&hit_fvec);
+               dec.orient.vec.fvec = hit_fvec;
+               vm_fix_matrix(&dec.orient);
                dec.pnt.xyz = hitpos->xyz;
                dec.radius = wip->decal_rad;
                if((dec.radius > 0) && (wip->decal_texture > -1))
-               decal_create(ship_obj, &dec, submodel_num, wip->decal_texture, wip->decal_backface_texture );*/
+               decal_create(ship_obj, &dec, submodel_num, wip->decal_texture, wip->decal_backface_texture, wip->decal_glow_texture, wip->decal_burn_texture, wip->decal_burn_time);
        }
```

Goober5000:
I haven't tested for performance regressions but I *have* tested this in mission.  I have no objection to performance testing but this is a pretty straightforward reduction in the number of copies.  It doesn't change any algorithms and it doesn't involve adding or removing loops.

I presume the impact of this PR will be small but positive.  There are no drastic changes, but since these functions are run every time collisions are checked between a ship and a weapon/beam, reducing the number of copies for each check has gotta be worth something.

I don't really have the cycles for performance testing myself. @qazwsxal since you brought it up are you willing to do it?

qazwsxal:
I haven't got bandwidth either, [here's my python scripts for profiling though](https://gist.github.com/qazwsxal/75bdd712628d97c6ec682b3f57d6d034), `run_icarus.py` does 5 runs of BP's icarus mission at a low resolution to avoid rendering overhead then dumps results to a numpy blob. Then `plot_files.py` generates a couple of graphs showing individual frametimes and averages, then compares the two on a third graph.

BMagnu:
A quick profiling run shows that in the collision-heavy segments, performance is effectively identical.
I'm less sure about the non-collision-heavy segements (like the first two cuts of icarus), as I see significant variance between runs in thse 300+FPS sections.

code/object/collideshipweapon.cpp
MageKing17: By "some kind of comment" I meant a *code* comment; if for no other reason than so that a future change won't accidentally reintroduce the reason it was copied in the first place.

Goober5000: Why though?  The fact that the pointer is `const` will guard against the source vector being modified, which was the reason for the copy.  The code that actually *modified* the copy no longer exists.

Adding a comment reminds me of trying to prove a negative.  "If you ever add code to modify this vector, you'll need to make a local copy."  You could just as reasonably make that comment for *any* `const vec3d *` in the codebase.

MageKing17: Not every `const vec3d *` in the codebase used to be a `vec3d` with a comment saying "not a pointer intentionaly".

qazwsxal:
> A quick profiling run shows that in the collision-heavy segments, performance is effectively identical. I'm less sure about the non-collision-heavy segements (like the first two cuts of icarus), as I see significant variance between runs in thse 300+FPS sections.

From an optimisation perspective, my general feeling with the 300+FPS sections is to practically ignore them. The main thrust of optimisation is to make the same missions playable on less powerful machines, and to make more intensive missions possible. 300FPS is as playable as 500FPS or 200FPS for almost everyone out there, so I don't think it's worth caring about.

That said, if performance is effectively identical, then what's the point of this PR? If it's making the code more readable, then sure, but it appears to be bogged down in a discussion about whether to add a comment or not.

Goober5000:
There are several goals here, in no particular order:
1. Address the Coverity issues
2. Improve performance to the extent possible in these functions
3. Apply polishing where appropriate

It's already had one approval, and not really bogged down, as the remaining review point is about something rather trivial.  Hopefully we can get that resolved soon.

code/object/collideshipweapon.cpp
Goober5000: There wasn't a 200 IQ master plan behind that comment, just a very simple reason: the variable was copied because it was modified locally.  The previous code has been superseded; the current code no longer modifies it; and this PR enforces const-ness.

The original comment, and commit, was made by Bobboau.  He had, shall we say, a rather creative way of implementing many of his features.  There were many times when he pioneered some feature that was rewritten later on, to a greater or lesser extent (e.g. decals, fighter beams, model animations).  In none of those cases did we need to add comments explaining an API that was no longer in use.

If you won't budge, I'll add the comment, but I don't think it's either necessary or clarifying.

MageKing17: > There wasn't a 200 IQ master plan behind that comment, just a very simple reason: the variable was copied because it was modified locally. The previous code has been superseded; the current code no longer modifies it; and this PR enforces const-ness.

The reason for the comment is irrelevant; the point is that anyone looking at the diff would see a comment being removed and the thing the comment was trying to prevent from happening actually happened. The commit message didn't explicitly explain, and only "alluded to" explaining it. Any confusion caused by this could be avoided by a very simple clarifying comment being added at the same time.

Goober5000: That rationale makes a lot more sense.  In that case, I propose amending the commit message.  Would that be acceptable?

MageKing17: Is there some reason not to comment the code, where the rationale would be immediately visible upon seeing the diff, rather than putting it in the commit message, where there would be an extra step to comprehension?

JohnAFernandez: Unless I'm mis-reading this discussion, I think you should take this suggestion to add a comment as a way to ease others into understanding why you made the change you made and why it is not a mistake.  Just because the code can be read by itself and understood, doesn't mean that adding a comment is valueless.  

You add value to code with comments by scaffolding other's understanding of it. You can potentially save someone a lot of investigative time whever something really subtle is wrong and every last line has to be analyzed.  We get plenty of bug reports that can't be reproduced, and you can't always rely on the debugger.

Goober5000: > Is there some reason not to comment the code, where the rationale would be immediately visible upon seeing the diff, rather than putting it in the commit message, where there would be an extra step to comprehension?

A person viewing the diff would also be viewing the commit message, and therefore any confusion about the *change* in code can be clarified in the commit message.  But a person viewing the current state of the code would see only the current, correct version.

@JohnAFernandez I agree 100% with you in general, but this particular situation is an exception, for the reasons stated earlier.

Goober5000: Well I have amended the previous commit to clarify the commit message, and added a new commit to comment unnecessarily about the `hit_dir` parameter.  I hope this is satisfactory.

MageKing17:
Thank you for adding the comment; now, where are your performance tests?

Goober5000:
> Thank you for adding the comment; now, where are your performance tests?

@BMagnu ran a profiling run a few posts up which should cover that.

MageKing17:
> @BMagnu ran a profiling run a few posts up which should cover that.

No it shouldn't; no data was provided. I say again, where are your performance tests to prove that your optimization isn't a regression?

Goober5000:
I'm happy to take him at his word, but based on your comment I've asked him to upload his data.  But he's busy for the short term so it may take several days.

BMagnu:
Okay, finally had time to rerun the benchmarks (icarus, 10 runs per option) properly with a lot of repetitions and influence of the nvidia driver minimized.
image
The first plot / the blue line in the third plot shows the pre-PR performance. The second plot / the orange line in the third plot shows the post-PR performance. There seems to be little difference overall, if not a tiiiiiiny speedup post-PR.

MageKing17 (approved):

PR #6243, opened 2024-07-08
Per weapon auto convergence

### LINKED ISSUE #4947: Feature Request: Per-Weapon convergences.
Heya

Seeing @MjnMixael implement per-weapon auto-aim had me thinking: What about per-weapon convergences? Would love to be able to, for example, have each weapon convergence at its maximum range.

MjnMixael:
This adds per-weapon convergence features, notably allowing per-weapon autoaim + convergence. I moved the autoaim flags to Object:: from Ship:: so they could be reused. Seemed the most sensible place.. but could go elsewhere if there's a better spot.

While technically a new feature, this allows BtA to fix an issue with one of our weapons.. so really hoping to get this in for 24.2.

Fixes #4947

### DIFF
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -12940,9 +12940,6 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
 							*/
 							const bool std_convergence_flagged = (sip->aiming_flags[Object::Aiming_Flags::Std_convergence]) || (winfo_p->aiming_flags[Object::Aiming_Flags::Std_convergence]);
 
-							//Must have the auto convergence flag and a valid target
-							const bool do_auto_convergence = (aip->target_objnum != -1) && auto_convergence_flagged;
-
 							if (has_autoaim && in_automatic_aim_fov) {
 								vec3d firing_vec;
 
@@ -12955,13 +12952,13 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
 								}
 
 								vm_vector_2_matrix(&firing_orient, &firing_vec, NULL, NULL);
-							} else if (std_convergence_flagged || do_auto_convergence) {
+							} else if (std_convergence_flagged || (auto_convergence_flagged && (aip->target_objnum != -1))) {
 								// std & auto convergence
 								vec3d target_vec, firing_vec, convergence_offset;
 								
 								// make sure vector is of the set length
 								vm_vec_copy_normalize(&target_vec, &firing_orient.vec.fvec);
-								if (do_auto_convergence) {
+								if (auto_convergence_flagged && (aip->target_objnum != -1)) {
 									// auto convergence
 									vm_vec_scale(&target_vec, dist_to_aim);
 								} else {
--- a/code/ship/ship.cpp
+++ b/code/ship/ship.cpp
@@ -12690,7 +12690,7 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
 				dist_to_aim = vm_vec_mag_quick(&plr_to_target_vec);
 
 				// Use the hightest minimum convergence distance
-				float min_convergence_distance = std::max(sip->minimum_convergence_distance, winfo_p->minimum_convergence_distance);
+				const float min_convergence_distance = std::max(sip->minimum_convergence_distance, winfo_p->minimum_convergence_distance);
 
 				// minimum convergence distance
 				if (min_convergence_distance > dist_to_aim) {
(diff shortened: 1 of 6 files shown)


code/ship/ship.cpp
```diff
@@ -12458,9 +12458,9 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
...
-		has_autoaim = ((has_converging_autoaim || (sip->aiming_flags[Ship::Aiming_Flags::Autoaim]) || (winfo_p->autoaim_fov > 0.0f) || (sip->bank_autoaim_fov[bank_to_fire] > 0.0f)) && aip->target_objnum != -1);
-		needs_target_pos = ((has_autoaim || (sip->aiming_flags[Ship::Aiming_Flags::Auto_convergence])) && aip->target_objnum != -1);
+		has_converging_autoaim = ((sip->aiming_flags[Object::Aiming_Flags::Autoaim_convergence] || winfo_p->aiming_flags[Object::Aiming_Flags::Autoaim_convergence] || (The_mission.ai_profile->player_autoaim_fov[Game_skill_level] > 0.0f && !( Game_mode & GM_MULTIPLAYER ))) && aip->target_objnum != -1);
+		has_autoaim = ((has_converging_autoaim || (sip->aiming_flags[Object::Aiming_Flags::Autoaim]) || winfo_p->aiming_flags[Object::Aiming_Flags::Autoaim] || (sip->bank_autoaim_fov[bank_to_fire] > 0.0f)) && aip->target_objnum != -1);
+		needs_target_pos = ((has_autoaim || has_converging_autoaim) && aip->target_objnum != -1); // This change needs some scrutiny but it seems right?
```
wookieejedi, line 12463: Good question. In current master this part of the check is `(sip->aiming_flags[Ship::Aiming_Flags::Auto_convergence])` and in the code here it is replaced with `has_converging_autoaim`.

`(sip->aiming_flags[Ship::Aiming_Flags::Auto_convergence])` is convergence with no autoaim on, and thus not the same as `has_converging_autoaim`. I would say using `(sip->aiming_flags[Object::Aiming_Flags::Auto_convergence]) || winfo_p->aiming_flags[Object::Aiming_Flags::Auto_convergence])` would fit as a more proper replacement. (Unless I'm missing something). What do you think?

MjnMixael: Ah yes. Good catch. I conflated auto_convergence with autoaim_convergence!

wookieejedi: Happy to help!

MjnMixael: I'm going to break down these bools for readability, I think. There's just a lot of checks here. The complier should optimize these out anyway.

MjnMixael: Ok, give that a look. I think that's a lot more readable

code/ship/ship.cpp
```diff
@@ -12458,9 +12463,17 @@ int ship_fire_primary(object * obj, int force, bool rollback_shot)
...
+		
+		// This logic is getting a little unweildy.. so let's break it down slightly for readability.
+		// First check our ship/weapon flags
+		const bool autoaim_flagged = ((sip->aiming_flags[Object::Aiming_Flags::Autoaim]) || (winfo_p->aiming_flags[Object::Aiming_Flags::Autoaim]));
+		const bool autoaim_convergence_flagged = ((sip->aiming_flags[Object::Aiming_Flags::Autoaim_convergence]) || (winfo_p->aiming_flags[Object::Aiming_Flags::Autoaim_convergence]));
+		const bool auto_convergence_flagged = ((sip->aiming_flags[Object::Aiming_Flags::Auto_convergence]) || (winfo_p->aiming_flags[Object::Aiming_Flags::Auto_convergence]));
+
+		// Now add in other checks like difficulty and game mode as appropriate. Also make sure we have a target and an FOV if necessary.
+		has_converging_autoaim = ((autoaim_convergence_flagged || (The_mission.ai_profile->player_autoaim_fov[Game_skill_level] > 0.0f && !( Game_mode & GM_MULTIPLAYER ))) && aip->target_objnum != -1);
+		has_autoaim = ((autoaim_flagged || has_converging_autoaim || (sip->bank_autoaim_fov[bank_to_fire] > 0.0f)) && aip->target_objnum != -1);
```
wookieejedi, line 12475: Looks good, just want to double check on one more:
Original: 
`has_autoaim = ((has_converging_autoaim || (sip->aiming_flags[Ship::Aiming_Flags::Autoaim]) || (winfo_p->autoaim_fov > 0.0f) || (sip->bank_autoaim_fov[bank_to_fire] > 0.0f)) && aip->target_objnum != -1);`

New:
`has_autoaim = ((autoaim_flagged || has_converging_autoaim || (sip->bank_autoaim_fov[bank_to_fire] > 0.0f)) && aip->target_objnum != -1);`

Looks like the new check is missing `(winfo_p->autoaim_fov > 0.0f)`?


Other very nitpicky thing, should `an FOV` by `a FOV` in the comments?

MjnMixael: Yes that one's correct. Previously we checked that the weapon autoaim_fov > 0 to determine if we had autoaim on the weapon. Now that the weapon has the aiming flags we can check the flag directly. The autoaim_fov is used later to determine what fov to actually aim with.

MjnMixael: "an FOV" when spoken aloud is how I'd say it... typed it's probably "a FOV". Either way probably not worth a whole commit just to change that.

wookieejedi: Sounds all good to me then!

code/ship/ship.cpp
wookieejedi: Thanks, that's definitely a good change.

wookieejedi (approved):
Previous comments addressed and my tests show things working as expected.

BMagnu (approved):

Issue #2875, opened 2020-11-08
Building on Mac with prebuilt ffmpeg libraries fails with Xcode 12.1


jg18:
When I run `xcodebuild -configuration Release` I get this output at the end:

```
CodeSign build/bin/Release/fs2_open_20_1_0_x64.app
    cd /path/to/fs2open
    export CODESIGN_ALLOCATE=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/codesign_allocate
    
Signing Identity:     "-"

    /usr/bin/codesign --force --sign - --entitlements /path/to/fs2open/build/freespace2/FS2_Open.build/Release/Freespace2.build/fs2_open_20_1_0_x64.app.xcent --timestamp=none /path/to/fs2open/build/bin/Release/fs2_open_20_1_0_x64.app
/path/to/fs2open/build/bin/Release/fs2_open_20_1_0_x64.app: code object is not signed at all
In subcomponent: /path/to/fs2open/build/bin/Release/fs2_open_20_1_0_x64.app/Contents/Frameworks/libswresample.2.1.100.dylib
Command /usr/bin/codesign failed with exit code 1

** BUILD FAILED **


The following build commands failed:
	CodeSign build/bin/Release/fs2_open_20_1_0_x64.app
(1 failure)
```

As a workaround, when I run the `codesign` command on all of the ffmpeg dylib files, such as with
```
for i in `ls -1 *.dylib`; do echo $i; /usr/bin/codesign --force --sign - --entitlements /path/to/fs2open/build/freespace2/FS2_Open.build/FastDebug/Freespace2.build/fs2_open_20_1_0_x64-FASTDBG.app.xcent  $i; done
```
then it works. It looks like some .dylibs are already signed and some aren't. It might be good to have them all signed by the same cert.

Using macOS 10.15.4 with Xcode 12.1 (Build version 12A7403).

PR #6246, opened 2024-07-09
update to newest prebuilt libs

notimaginative:
Supports newer SDL2 from the prebuilt package on win32/win64
Adds support for Windows ARM64
Adds option to use prebuilt SDL2/OpenAL/Freetype libs on Linux


notimaginative:
Fixes #2875

Goober5000 (approved):

PR #6248, opened 2024-07-11
Bug fix for aspect invulnerability timing... again

Baezon:
Follow-up to #5595. As flagged by coverity, the previous PR took two conditions, one using `max_lifetime - lifeleft` and the other using `lifetime - lifeleft` and merged them into a single case `Missiontime - time_created`, which as mentioned there *is* more robust but in practice mostly matches the behavior of `max_lifetime - lifeleft`. What is the difference between `max_lifetime` and `lifetime`? It was removed in #5753 but it was very simple and not used for much
```
	if (wip->is_locked_homing()) {
		// locked homing missiles have a much longer lifespan than the AI think they do
		wip->max_lifetime = wip->lifetime * LOCKED_HOMING_EXTENDED_LIFE_FACTOR; // this is 1.2f
	}
```

Essentially, original handling was that bomb invulnerability time was extended by 20% of the weapon's lifetime. So in the case the fix flag is not being used, that time should be added back in.

### DIFF
--- a/code/object/collideweaponweapon.cpp
+++ b/code/object/collideweaponweapon.cpp
@@ -64,12 +64,17 @@ int collide_weapon_weapon( obj_pair * pair )
 		if (!(wipA->wi_flags[Weapon::Info_Flags::No_radius_doubling])) {
 			A_radius *= 2;		// Makes bombs easier to hit
 		}
+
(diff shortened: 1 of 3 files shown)


Goober5000 (approved):
Looks good 👍

PR #6249, opened 2024-07-11
Interpolate particle spawning positions for moving spawner objects

Kestrellius:
Previously, particle effects from -part.tbm could be spawned more than once per frame, but this would only result in the particles clumping up at the spawner's position each frame. Smoothly spawning many particles along a fast-moving object's path was impossible. This PR makes it so that particles spawned between frames are placed partway along the spawner's trajectory in accordance with spawn timing, with millisecond precision.

Limitations:
- For now, this applies only to the "object" source type, so it won't affect particles spawned from beams or as trails from other particles. I intend to extend it to beams in the future, but that will be slightly more complicated so I'm leaving it off for now. I'm not sure it's even worth doing for the particle source type -- particles don't store last-frame position data, so I'd need to add that, which might be a significant performance burden.
- The interpolation is linear. For objects on a curved trajectory, the interpolation on each frame will be a straight line rather than following the curve (though this isn't very noticeable even for tightly-curving objects), and it doesn't take acceleration into account. Both of these things could probably be fixed with some effort, but I'm not convinced it's necessary.
- Millisecond precision means that there's a limit on how close together spawns can occur, relative to the speed of the spawner object. For a spawner traveling at 1000 m/s, the spawns will be no closer than 1 meter. I have ideas about how to increase the precision, and if a use case arises, I will probably implement them.

Other details:
- I fixed a preexisting issue that would cause the game to hang when +Effects per second was set to a value higher than 1000. Because the delay between spawns was calculated using milliseconds, a per-second value above 1000 would cause the delay to be rounded to 0, and the game would attempt to spawn infinitely many particles. I clamped the delay to no less than 1 millisecond, and added a warning for values above 1000.
- I made +Effects per second a random range, which effectively allows the delay between spawns to be randomized. My one concern about this is how it's presented to the modder -- the phrasing might lead people to think the randomization only happens once per second or something, rather than every spawn.

### DIFF
--- a/code/particle/ParticleSource.h
+++ b/code/particle/ParticleSource.h
@@ -269,6 +269,8 @@ class SourceTiming {
 	 */
 	float getLifeTimeProgress() const;
 
+	int getNextCreationTime();
+
 	/**
 	 * @brief Determine if the timestamp for the next particle creation has expired
 	 * @return @c true if the timestamp has expired, false otherwise
--- a/code/particle/ParticleSource.h
+++ b/code/particle/ParticleSource.h
@@ -67,9 +67,14 @@ class SourceOrigin {
 	/**
 	 * @brief Gets the current, global position of the origin
 	 * 
+	 * For object sources, can interpolate between object's current position and its position last frame
+	 * 
 	 * Be aware for beam sources this will give *random* points along its length
 	 * 
 	 * @param posOut The pointer where the location will be stored
+	 * 
+	 * @param interp For objects, the point's position between the current-frame
+	 * and last-frame positions of the object, with 0.0 being current-frame and 1.0 being last-frame
 	 */
 	void getGlobalPosition(vec3d* posOut, float interp = 1.0f) const;
 
--- a/code/particle/ParticleSource.h
+++ b/code/particle/ParticleSource.h
@@ -274,7 +274,7 @@ class SourceTiming {
 	 */
 	float getLifeTimeProgress() const;
 
-	int getNextCreationTime();
+	int getNextCreationTime() const;
 
 	/**
 	 * @brief Determine if the timestamp for the next particle creation has expired
--- a/code/particle/ParticleSource.h
+++ b/code/particle/ParticleSource.h
@@ -71,7 +71,7 @@ class SourceOrigin {
 	 * 
 	 * @param posOut The pointer where the location will be stored
 	 */
-	void getGlobalPosition(vec3d* posOut) const;
+	void getGlobalPosition(vec3d* posOut, float interp = 1.0f) const;
 
 	void getHostOrientation(matrix* matOut) const;
 
@@ -96,7 +96,7 @@ class SourceOrigin {
 	 *
 	 * @param info The particle_info this should be applied to
 	 */
-	void applyToParticleInfo(particle_info& info, bool allowRelative = false) const;
+	void applyToParticleInfo(particle_info& info, bool allowRelative = false, float interp = 1.0f) const;
 
 	/**
 	 * @brief Gets the velocity of the origin host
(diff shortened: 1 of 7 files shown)


code/particle/util/EffectTiming.h
```diff
@@ -66,7 +66,7 @@ class EffectTiming {
 	 * TimingState and then pass it to this function.
 	 * @return @c true if a new effect should be created, @c false. This might return @c true multiple times per frame.
```
Baezon, line 67: Make sure to update this documentation with the new kind of return

code/particle/ParticleSource.h
```diff
@@ -71,7 +71,7 @@ class SourceOrigin {
 	 * 
 	 * @param posOut The pointer where the location will be stored
```
Baezon, line 72: Don't forget to document the changes here too

Baezon (changes requested):
Something that occurred to me looking at the code, particles don't have a `last_pos` but they *do* have a velocity...

It isn't necessarily going to be representative 100% of the time, but negative one frame's worth of velocity will in practice pretty much the particles last position in almost all cases and might be good enough for this.

code/particle/ParticleSource.h
Kestrellius: All right, taken care of!

Kestrellius:
> Something that occurred to me looking at the code, particles don't have a `last_pos` but they _do_ have a velocity...
> 
> It isn't necessarily going to be representative 100% of the time, but negative one frame's worth of velocity will in practice pretty much the particles last position in almost all cases and might be good enough for this.

Yeah, maybe. I could experiment with it. I think it'd be exactly correct in all cases except where velocity curves are concerned. If those are in play...it should be theoretically possible to calculate exactly how far the particle has traveled based on the curve? Not sure how difficult, or how expensive, it'd be, though.

Baezon:
I would 100% not bother. It would require integrating the curve function, which is not always mathematically possible depending on the exact curve. The curve will update the velocity every frame, I still think that is good enough in 99% of cases, but again not necessarily representative. I'll leave to you whether you want to follow up on that, but it is extremely simple and *mostly* correct.

Baezon (approved):

Kestrellius:
I had a thought.

The way I have it set up when spawning from objects, the spawn points are spread between the object's current position and its position last frame. With a particle, we don't have the last frame data; we can use the velocity to approximate, but that will fail sometimes if there are rapid shifts in velocity.

However, what the velocity _does_ constitute, as far as I can tell, is reliably accurate data about the particle's position _next_ frame. So what if instead of spreading the spawn points _backward_, we spread them _forward_? Then it would always be accurate.

I don't love the inconsistency between spawner types, but it might be worth it?

Kestrellius:
> However, what the velocity _does_ constitute, as far as I can tell, is reliably accurate data about the particle's position _next_ frame. So what if instead of spreading the spawn points _backward_, we spread them _forward_? Then it would always be accurate.

...Oh, wait. We don't know how long the next frame is going to be. So this won't work.

Never mind.

Baezon:
`flFrametime` *is* explicitly how long the next frame, the one being evaluated, is being estimated to take, and what all the simulation is assuming and acting on. That doesn't *at all* guarantee it is correct, but it will be in 99.9% of cases.

I don't think interpolating forward or backward makes that much difference, it will occasionally be wrong (in cases where the particle was just spawned or is about to terminate), and also occasionally be wrong due to acceleration, but again, it works in 99.9% of cases. It *really* is not worth the headache agonizing over the 0.01% of cases where it isn't perfect. 

Perfect is the enemy of good enough.

Kestrellius:
Pushed support for interpolation when spawning particles from particles. It should be completely accurate.

PR #6250, opened 2024-07-12
only run the Coverity workflow on the SCP repository

Goober5000:
This prevents Coverity from running on any forked repository, by default; both because the workflow will fail, and also because the workflow's results would be identical.  (Any forked repository can, of course, change this within its own scope.)


JohnAFernandez:
Thank you, I've been meaning to do this.

JohnAFernandez (approved):

PR #6253, opened 2024-07-14
Ensure guards ignore protected attackers

Baezon:
Since guards going after a ship which is threatening their charge is an implicit attack order, it should be subject to the protected flag like all other similar situations, otherwise they'll chase but not fire (protected also stops firing) at the ship. 

This is definitely a bug, and an uncommon one, which is almost certainly not present in retail, but better to err on the side of caution with AI so I made it a flag targeting new versions.

### DIFF
--- a/code/ai/ai_flags.h
+++ b/code/ai/ai_flags.h
@@ -164,6 +164,7 @@ namespace AI {
 		Align_to_target_when_guarding_still,
 		Debris_respects_big_damage,
 		Dont_limit_change_in_speed_due_to_physics_whack,
+		Guards_ignore_protected_attackers,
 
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):
Nice catch

Goober5000 (approved):
I just checked the retail code, and the guard function does not prevent the AI from chasing a protected ship, so a flag is the right call.

Baezon:
I did not mean to imply the bug was not present in retail, but rather that a situation where it would've made a difference was not present in retail, but I have not specifically tested for it.

Goober5000:
Ok, fair enough.

PR #6254, opened 2024-07-15
Conditional Impacts followup

Kestrellius:
This is a follow up to [#6218](https://github.com/scp-fs2open/fs2open.github.com/pull/6218), and makes a few changes.

First, I switched the ordering of a couple of the table lines. Previously, the armor type came first, then the effect name, then the other conditions. However, the armor type is a condition, whereas the effect is what happens when the conditions are met, and I thought that each of those things should be clustered. So now, the effect name goes at the end.

This is modder-facing, but the feature was merged like a week ago and I would be very surprised if anyone had used it yet, so I expect it'll be fine.

Second, I added support for conditional impacts when hitting weapons (such as bombs). This got a little thornier than I anticipated, and required making a significant change in collideweaponweapon.cpp -- weapon-weapon impacts now call weapon_hit(), where previously they just set the weapon's lifetime to a very low value and let lifetime expiration take care of it.

This is sort of a big alteration. I'm not sure I've correctly understood all the implications, and I won't be surprised if the PR is rejected for this reason. (Now that I think of it, I should have put these two changes in separate branches. Oh well -- they're simple; if I need to I can split them out manually without much trouble.) There is one change in behavior that I can think of, though it's purely cosmetic: before, hitting a weapon wouldn't play the impact effect (unless 'detonate on expiration' was set), whereas now it will. Possibly there's some mod somewhere whose art style was importantly dependent on this?

### DIFF
--- a/code/object/collideweaponweapon.cpp
+++ b/code/object/collideweaponweapon.cpp
@@ -125,12 +125,10 @@ int collide_weapon_weapon( obj_pair * pair )
 			if (wipA->weapon_hitpoints > 0) {
 				if (wipB->weapon_hitpoints > 0) {		//	Two bombs collide, detonate both.
 					if ((wipA->wi_flags[Weapon::Info_Flags::Bomb]) && (wipB->wi_flags[Weapon::Info_Flags::Bomb])) {
-						wpA->lifeleft = 0.001f;
-						wpA->weapon_flags.set(Weapon::Weapon_Flags::Begun_detonation);
 						wpA->weapon_flags.set(Weapon::Weapon_Flags::Destroyed_by_weapon);
-						wpB->lifeleft = 0.001f;
-						wpB->weapon_flags.set(Weapon::Weapon_Flags::Begun_detonation);
+						weapon_hit(A, B, &A->pos, -1, nullptr);
 						wpB->weapon_flags.set(Weapon::Weapon_Flags::Destroyed_by_weapon);
+						weapon_hit(B, A, &B->pos, -1, nullptr);
 					} else {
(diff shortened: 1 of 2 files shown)


Baezon (changes requested):
The main difference I can see is that there would be hit sounds for lasers hitting bombs. Leaving aside whether this should have been allowed as a separate issue... should be simple enough to check if the other object is a weapon in `weapon_hit_do_sound`, and not play sounds in that case.

Kestrellius:
> The main difference I can see is that there would be hit sounds for lasers hitting bombs. Leaving aside whether this should have been allowed as a separate issue... should be simple enough to check if the other object is a weapon in `weapon_hit_do_sound`, and not play sounds in that case.

Okay. So, as best as I can tell, it should already work like this. There's a switch in weapon_hit_do_sound, conditioned on the weapon being a non-missile, and it checks for OBJ_SHIP and OBJ_ASTEROID, and then the default just returns. Without playing a sound, as far as I can determine. So if we have a non-missile hitting a weapon, it should end up in that default and not play a sound.

Baezon (approved):
Ah, yes, you are right. It should be ok then otherwise, as far as I can tell. The other collision handling pair functions already use `weapon_hit`.

wookieejedi:
This one appears ready to merge? Just checking specifically b/c it is the past 24.2 marked PR

Kestrellius:
Should be ready to merge, yeah!

wookieejedi:
Sounds good. Will merge today or tomorrow if no one else beats me to it.

PR #6255, opened 2024-07-16
change atan2_safe to atan2

### LINKED ISSUE #3730: Displayed Bank Angle has a Strange Range
(First reported by Strygon on Discord)

[Image 1](https://cdn.discordapp.com/attachments/223511391531827200/899087002157330462/unknown.png)
[Image 2](https://cdn.discordapp.com/attachments/223511391531827200/899087025741901854/unknown.png)

As shown, the Bank angle displayed has a strange range of -90 degrees to 270.  After a quick poking about I found a strange usage of `atanf()` inside of `atan2_safe()` within vecmat.cpp.  Discussion on Discord with DahBlount led me to suspect this is a workaround to C++98's buggy implementation of atan2 is causing the range error.

Modern versions of C++ should have an `atan2()` that works correctly when the inputs are at 0 or near epsilon of the floating point specification, meaning Volitions workaround shouldn't be needed anymore.

Note: the angles are correct, but not intuitively displayed for humans.

Goober5000:
According to DahBlount:
> Most atan2 implementations for a long time had problems when the inputs were close to machine epsilon or when x was less than or equal to 0

And according to z64555:
> Modern versions of C++ should have an atan2() that works correctly when the inputs are at 0 or near epsilon of the floating point specification, meaning Volition's workaround shouldn't be needed anymore.

So this removes `atan2_safe` and changes all instances to `atan2`.  Consequently, the angle range in FRED now shows as [-180, 180] as expected.

Fixes #3730.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -3046,7 +3046,7 @@ void HudGaugeReticleTriangle::renderTriangle(vec3d *hostile_pos, int aspect_flag
 	// even if the screen position overflowed, it will still be pointing in the correct direction
 	unsize( &hostile_vertex.screen.xyw.x, &hostile_vertex.screen.xyw.y );
 
-	ang = atan2_safe(-(hostile_vertex.screen.xyw.y - tablePosY), hostile_vertex.screen.xyw.x - tablePosX);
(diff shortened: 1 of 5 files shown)


Baezon (approved):
(It should be noted that `atan2` has the exact same returns `atan2_safe` has for the same 'safety' special cases so this should not result in any behavior changes)

PR #6258, opened 2024-07-16
fix `NextFireTimestamp`

### LINKED ISSUE #6256: `NextFireTimestamp` does not return mission time
`NextFireTimestamp` on subsystems is documented as
```
 number NextFireTimestamp = number
    The next time the turret may attempt to fire
    Value: Mission time (seconds) or -1 on error
```
But the return is instead the amount of real time passed as seconds since the start of the program. This obviously quite difficult to work with. I don't know if it's always been this way, or if any other timestamp script values are affected...

Goober5000:
Changes `NextFireTimestamp` to work as advertised.  Fixes #6256.

EDIT: Also handles special timestamp values in `timestamp_until()` and `timestamp_since()`.

### DIFF
--- a/code/scripting/api/objs/subsystem.cpp
+++ b/code/scripting/api/objs/subsystem.cpp
@@ -656,7 +656,7 @@ ADE_VIRTVAR(NextFireTimestamp, l_Subsystem, "number", "The next time the turret 
 		float delta = newVal - currentTime;
 		if (delta < 0.0f)
 		{
-			Warning(LOCATION, "NextFireTimestamp: Specified value is in the past; setting to the current time");
+			mprintf(("NextFireTimestamp: Specified value is in the past; setting to the current time\n"));
 			delta = 0.0f;
 		}
 
--- a/code/scripting/api/objs/subsystem.cpp
+++ b/code/scripting/api/objs/subsystem.cpp
@@ -649,12 +649,21 @@ ADE_VIRTVAR(NextFireTimestamp, l_Subsystem, "number", "The next time the turret 
 	if (!sso->isValid())
 		return ade_set_error(L, "f", -1.0f);
 
-	if(ADE_SETTING_VAR)
+	float currentTime = f2fl(Missiontime);
+
+	if (ADE_SETTING_VAR)
 	{
-		sso->ss->turret_next_fire_stamp = (int)(newVal * 1000);
+		float delta = newVal - currentTime;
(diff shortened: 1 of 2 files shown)


code/scripting/api/objs/subsystem.cpp
```diff
@@ -649,12 +649,21 @@ ADE_VIRTVAR(NextFireTimestamp, l_Subsystem, "number", "The next time the turret
...
+
+	if (ADE_SETTING_VAR)
 	{
-		sso->ss->turret_next_fire_stamp = (int)(newVal * 1000);
+		float delta = newVal - currentTime;
+		if (delta < 0.0f)
+		{
+			Warning(LOCATION, "NextFireTimestamp: Specified value is in the past; setting to the current time");
+			delta = 0.0f;
+		}
```
Baezon, line 661: What's wrong with setting a time in the past?

Goober5000: Well, the earliest the turret can fire is *now*, not several seconds ago.

Baezon: Sure, it may not obviously affect behavior, but it's a valid input for the value. A scripter might be interested in how long the turret has been able to fire, or want to that to be a particular value. It's a timestamp, and there's nothing inherently wrong with a timestamp in the past, right?

Goober5000: It's certainly *possible*, but I don't understand why you would want to *set* the timestamp to some point in the past.  It seems to me that a negative RHS value is likely to be a bug or mistake.

Baezon: Just because we can't think of the use case doesn't mean I think we should immediately warn for it. Sure it may be a bug, but if it is a valid input, I don't think we should be complaining so loudly. In my opinion, scripting should really only warn on things which are completely invalid, and have *no possible* valid usage.

Goober5000: Actually, I can't change it, because the game doesn't actually allow you to create a timestamp in the past.  All timestamps with a negative delta are treated as invalid.

Baezon (changes requested):
After a lot of back and forth, I think we got it figured out. It's not a deal breaker, but I still feel the negative delta case should probably be handled silently, even some innocuous float math could accidentally make it negative.

Baezon (approved):

PR #6260, opened 2024-07-20
some coverity fixes

Goober5000:
Check the return value of `eval_wing()`, and tweak the checks for `closeup_pos` in `draw_model_rotating()`.  This addresses a few Coverity defects.

EDIT: Also address two false positives and refine the usage of `sexp_get_variable_index()`.


code/io/timer.h
```diff
@@ -63,6 +63,7 @@ class TIMESTAMP : public util::ID<timestamp_tag, int, -1>
 
 	inline bool isNever() const { return m_val == 0; }
 	inline bool isImmediate() const { return m_val == 1; }
+	inline bool isFinite() const { return isValid() && !isNever(); }
 };
```
JohnAFernandez, line 67: Honestly, these timestamp changes should be separate.

Goober5000: Ok

Goober5000: I've split them off into PR #6278

wookieejedi (approved):
Overall looks good to me but happy to have another set of eyes if anyone wants to.

PR #6262, opened 2024-07-21
Fix: Always check for nullptrs and make red-alert respect cur_waves from last mission

JohnAFernandez:
Fixes part of the cause of a crash reported on discord.  This is now very unlikely to happen, but everywhere else in this function checks that `p_objp` is not null before dereferencing.  Also, while working on this, goober and I saw that the current wave was sometimes exceeding the wave count of the second mission, so fix that as well.

Tested on the data that originally crashed FSO, and it no longer does.  Ships also properly show up as Gamma 5, Gamma 6, etc

### DIFF
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -938,7 +938,9 @@ void red_alert_bash_ship_status()
 				wingp->current_wave = rws->latest_wave;
 				
 				// make sure that num_waves can accomodate having extra wings.
+				// this would happen if the mission we are entering has too few waves.
 				if (wingp->num_waves < wingp->current_wave){
+					mprintf(("Red alert warning: make sure that the current mission has at least as many waves as exited last mission. Bashing the number of waves of wing %s to %i!\n", wingp->name, wingp->current_wave));					
 					wingp->num_waves = wingp->current_wave;
 				}
 
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -939,7 +939,8 @@ void red_alert_bash_ship_status()
 				
 				// make sure that num_waves can accommodate the new value of current_wave.
 				// this would happen if the mission we are entering has too few waves.
-				if (wingp->num_waves < wingp->current_wave){
+				if (wingp->num_waves < wingp->current_wave)
+				{
 					mprintf(("Red alert warning: wing %s in the current mission has fewer waves than in the last mission.  Bashing the number of waves from %i to %i.\n", wingp->name, wingp->num_waves, wingp->current_wave));
 					wingp->num_waves = wingp->current_wave;
 				}
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -936,6 +936,11 @@ void red_alert_bash_ship_status()
 
 				// assign the wave number to the wing
 				wingp->current_wave = rws->latest_wave;
+				
+				// make sure that num_waves can accomodate having extra wings.
+				if (wingp->num_waves < wingp->current_wave){
+					wingp->num_waves = wingp->current_wave;
+				}
 
 				if (rws->latest_wave > 1)
 				{
--- a/code/missionui/redalert.cpp
+++ b/code/missionui/redalert.cpp
@@ -937,10 +937,10 @@ void red_alert_bash_ship_status()
 				// assign the wave number to the wing
 				wingp->current_wave = rws->latest_wave;
 				
-				// make sure that num_waves can accomodate having extra wings.
+				// make sure that num_waves can accommodate the new value of current_wave.
 				// this would happen if the mission we are entering has too few waves.
 				if (wingp->num_waves < wingp->current_wave){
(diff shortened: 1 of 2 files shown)


code/ship/ship.cpp
```diff
@@ -11902,7 +11902,12 @@ void change_ship_type(int n, int ship_type, int by_sexp)
 
 	// zookeeper - If we're switching in the loadout screen, make sure we retain initial velocity set in FRED
 	if (!(Game_mode & GM_IN_MISSION) && !(Fred_running)) {
-		Objects[sp->objnum].phys_info.speed = (float) p_objp->initial_velocity * sip->max_speed / 100.0f;
+		if (p_objp != nullptr){
```
Goober5000, line 11905: The nullptr check should be part of the enclosing if().  If there is no parse object, there is no need to change the velocity at all, and it should not be set to 0.  And if the velocity isn't changed, the rest of the if() block is redundant.

code/missionui/redalert.cpp
```diff
@@ -936,6 +936,13 @@ void red_alert_bash_ship_status()
 
 				// assign the wave number to the wing
 				wingp->current_wave = rws->latest_wave;
+				
+				// make sure that num_waves can accomodate having extra wings.
+				// this would happen if the mission we are entering has too few waves.
+				if (wingp->num_waves < wingp->current_wave){
+					mprintf(("Red alert warning: make sure that the current mission has at least as many waves as exited last mission. Bashing the number of waves of wing %s to %i!\n", wingp->name, wingp->current_wave));
```
Goober5000, line 943: I intended the "make sure..." comment to go above the if(); it is a description of what the code is doing, not a warning to the user.  Anyway your comment is fine too; but "extra wings" is misleading and should be reworded.  Between that and tweaking the `mprintf`, I suggest:

```suggestion
				// make sure that num_waves can accommodate the new wave number.
				// this would happen if the mission we are entering has too few waves.
				if (wingp->num_waves < wingp->current_wave){
					mprintf(("Red alert warning: wing %s in the current mission has fewer waves than last mission. Bashing the number of waves from %i to %i.\n", wingp->name, wingp->num_waves, wingp->current_wave));
```

Goober5000 (changes requested):

code/ship/ship.cpp
JohnAFernandez: It's conceivable that your assumption could be invalidated.  Ship lab code could call this in the future, or some other block of code.  So it's better for this to be accessible for odd cases where we wouldn't have a parse object.

code/missionui/redalert.cpp
JohnAFernandez: Uh... I'm not sure which if you are referencing.  It is currently over an if.

Sure I misspelled accommodate, but extra wings is not misleading because they are extra wings from the last mission that are not present in the current mission.  You saying "the new wave number" is redundant.  It's like saying, "Make sure the box can accommodate the box".

It's also better to be more direct in the mprintf because the modder is less likely to just overlook the statement since debug logs have so much information.  However it is helpful to add the previous number of waves, so I can do that tomorrow.

code/ship/ship.cpp
Goober5000: The initial velocity is a property of the parse object.  That's not an assumption, that's a fact of the code.  If there is no parse object, it is meaningless to talk about the initial velocity.  Therefore, if there is no parse object, all of the initial velocity-related code can be skipped.

Suppose a ship is created via ship-create.  In that situation it has no parse object.  So if change-ship-class is called on it, the speed should not be set to 0; the speed should be left alone.

code/missionui/redalert.cpp
Goober5000: > Uh... I'm not sure which if you are referencing. It is currently over an if.

In other words, the text that is currently inside the if() (and inside the mprintf) is what I intended to be part of the comment above the if().

> Sure I misspelled accommodate, but extra wings is not misleading because they are extra wings from the last mission that are not present in the current mission. You saying "the new wave number" is redundant. It's like saying, "Make sure the box can accommodate the box".

"Extra wings" could mean something like Delta, Epsilon, etc. in addition to Alpha, Beta, and Gamma.  The term *wings* refers to the named collection of ships.  The term *waves* refers to how many "copies" of the collection there can be.  That's why I said "extra wings" is misleading; the term "wings" is being used in a nonstandard way.

By "the new wave number" I'm referring to the new value of `current_wave`, but I'm open to other phrasing if there's something that would be clearer.  Maybe literally "the new value of current_wave".

> It's also better to be more direct in the mprintf because the modder is less likely to just overlook the statement since debug logs have so much information.

The problem though is that the comment I suggested was intended to describe code, and be viewed by coders.  By putting it in the log, it is viewable by modders, and thus it can easily be interpreted as instructions to modders, when that's not the intention.

Goober5000:
Since Cyborg is temporarily inactive, after consultation with Lafiel, I pushed an update.

Goober5000 (approved):

Issue #6203, opened 2024-06-25
Feature request: disable friendly fire damage globally


MoerasGrizzly:
ai_profiles and FRED flag that disables friendly fire globally, per-mission or even temporarily through a SEXP could be rather useful.

It's not entirely simple from a design perspective. In my mind, a "no friendly fire damage" flag should:
Block friendlies from shooting eachother
still allow players to damage themselves with their own missiles (with a seperate flag to disable that instead)
still allow damage inflicted by one's own bombs being shot down by hostile forces (whilst preventing any side from shooting down their own bombs alltogether)

The game already has a `$Max Beam Friendly Fire Damage:` flag, so if that could be extended to include other weapons that'd be great.

PR #6265, opened 2024-07-23
Friendly fire control flags and SEXP

Kestrellius:
Adds two AI profile/mission profile flags: a cap on weapon same-team damage, and a cap on weapon self damage. These complement an existing flag that is a cap on beam same-team damage. (Beam self damage happening at all requires such narrow conditions that I don't think it's worth bothering with.) Also adds a SEXP to set the three flags to any value.

Mostly resolves [#6203](https://github.com/scp-fs2open/fs2open.github.com/issues/6203).

Modifying FRED is apparently a gigantic pain, so that bit is going to be someone else's problem.


code/ai/ai_profiles.cpp
```diff
@@ -761,8 +769,10 @@ void ai_profile_t::reset()
 		player_damage_inflicted_scale[i] = 0;
 
         subsys_damage_scale[i] = 0;
-        beam_friendly_damage_cap[i] = 0;
-        turn_time_scale[i] = 0;
+        beam_friendly_damage_cap[i] = -1.f;
+		weapon_friendly_damage_cap[i] = -1.f;
+        weapon_self_damage_cap[i] = -1.f;
+		turn_time_scale[i] = 0;
```
Goober5000, line 775: The indentation looks a bit inconsistent here

code/ai/ai_profiles.h
```diff
@@ -9,10 +9,12 @@
 #ifndef _AI_PROFILES_H_
 #define _AI_PROFILES_H_
 
+#include<array>
 #include "globalincs/globals.h"
 #include "globalincs/pstypes.h"
 #include "globalincs/systemvars.h"
 #include "ai/ai_flags.h"
+#include<tl/optional.hpp>
```
Goober5000, line 17: These two includes are no longer needed

code/parse/sexp.cpp
```diff
@@ -13320,6 +13321,37 @@ void sexp_player_use_ai(int flag)
 	}
 }
 
+void sexp_set_friendly_damage_caps(int n) {
+	bool is_nan;
+	bool is_nan_forever;
+	ai_profile_t& aip = *The_mission.ai_profile;
+
+	float beam_friendly_cap = i2fl(eval_num(n, is_nan, is_nan_forever));
```
Goober5000, line 13329: It's a good idea to check for `is_nan` or `is_nan_forever` being set to true.  Same with the other two caps.

code/parse/sexp.cpp
```diff
@@ -13320,6 +13321,37 @@ void sexp_player_use_ai(int flag)
...
+	for (auto &bfc : aip.beam_friendly_damage_cap) {
+		bfc = beam_friendly_cap;
+	}
```
Goober5000, line 13332: This will set damage caps for *all* difficulty levels at once.  Consider some other options:
1. Set the caps for only the current difficulty level
2. Add an argument to set the caps for only the specified difficulty level
3. Use multiple arguments and allow the FREDder to specify multiple caps for the different difficulty levels
4. Add other SEXP operators to allow setting caps independently
5. Some combination of the above

I think if I were doing this, I'd make three SEXP operators, each of which took a difficulty argument and a number argument.  Or you could make one of the arguments the *type* of cap, which would let you use one operator to handle all the different types

code/parse/sexp.cpp
```diff
@@ -32015,6 +32054,8 @@ int query_operator_argument_type(int op, int argnum)
 			else
 				return OPF_VARIABLE_NAME;
 
+		case OP_SET_FRIENDLY_DAMAGE_CAPS:
+			return OPF_POSITIVE;
```
Goober5000, line 32058: This should be `OPF_NUMBER` otherwise FREDders would not be able to pass -1 to turn off the cap

code/parse/sexp.cpp
```diff
@@ -40542,6 +40585,14 @@ SCP_vector<sexp_help_struct> Sexp_help = {
 		"\tCauses the player's ship to not be controlled by the FreeSpace AI.  Takes 0 arguments.\r\n"
 	},
 
+	// Kestrellius
+	{ OP_SET_FRIENDLY_DAMAGE_CAPS, "set-friendly-damage-caps\r\n"
+		"\tSets limits on damage weapons and beams can do to friendly targets. Takes 1 to 3 arguments.\r\nArguments left blank will leave the values unmodified.\r\n"
```
Goober5000, line 40590: The description should mention that a negative number lifts the cap

code/object/collideshipweapon.cpp
```diff
@@ -104,6 +104,17 @@ static void ship_weapon_do_hit_stuff(object *pship_obj, object *weapon_obj, vec3
 	if (wip->damage_curve_idx >= 0)
 		damage *= Curves[wip->damage_curve_idx].GetValue(f2fl(Missiontime - wp->creation_time) / wip->lifetime);
 
+	// if this is friendly fire, we check for the friendly fire cap values
+	if (wp->team == shipp->team) {
+		if (&Objects[weapon_obj->parent] == pship_obj && The_mission.ai_profile->weapon_self_damage_cap[Game_skill_level] >= 0.f) {
```
Goober5000, line 109: Does this correctly handle multiple-level spawns?  E.g. if a weapon spawns a weapon that spawns another weapon?  The immediate parent would be the parent weapon, not the parent ship.

code/weapon/shockwave.cpp
```diff
@@ -347,21 +348,54 @@ void shockwave_move(object *shockwave_objp, float frametime)
...
+				!(objp->flags[Object::Object_Flags::Invulnerable])) {
 				weapon_do_electronics_effect(objp, &sw->pos, sw->weapon_info_index);
 			}
-			ship_apply_global_damage(objp, shockwave_objp, &sw->pos, damage, sw->damage_type_idx );
+
+			ship* shipp = &Ships[objp->instance];
+
+			// we make sure that this shockwave was spawned by a weapon that ultimately came from a ship, and then check
+			// to see if it's friendly fire
+			if (shockwave_objp->parent >= 0
```
Goober5000, line 362: Same question about multiple weapon spawns

Goober5000 (changes requested):
Several comments.  Also, with the default value now being -1, the use of `beam_friendly_damage_cap` in beam.cpp should be tweaked.

code/object/collideshipweapon.cpp
Kestrellius: A submunition's parent is the grandparent ship, not the spawner weapon. (When submunitions are spawned, the submunition's parent gets set to the spawner's parent, so it's consistent through any number of generations.) I think there's some specific behavior that depends on this, though I don't remember exactly what it was.

code/weapon/shockwave.cpp
Kestrellius: Same thing applies here.

Kestrellius:
Good catches, thank you!

code/parse/sexp.cpp
Kestrellius: I arranged for it to apply only to the current difficulty. Honestly I'd prefer not to make changes that alter the argument configuration if I can get away with it, just because it's fiddly and I'm still not very familiar with the SEXP code. Separately, I'm a little skeptical that applying different damage caps for different difficulties on the fly in-mission is ever likely to be relevant.

Still, applying it only to the current difficulty feels more elegant to my intuitions than applying it to all, so here we are.

Goober5000: Current difficulty works 👍

Goober5000 (approved):
Great work on this, including the revisions. 👍

MoerasGrizzly:
Thank you

PR #6268, opened 2024-07-24
Remove some overzealous prints

Baezon:
- Normal-based particle effects being spawned in a situation with no normal is a totally expected situation with a totally reasonable fall back. For instance, a weapon impact effect which spawns a particle as though it were 'bouncing' off its target would generate this warning everytime it detonated without hitting anything.
- Null vectors being normalized was definitely more a concern the developers had while developing, now, scripters can call this function directly, so it can certainly not be assumed and once again, has reasonable (and obvious, to a possibly concerned scripter) fallback

Most importantly of all, both of these situations can happen numerous times, and can REALLY clutter the log. I'm open to `nprintf` instead, but I don't think they're necessary.

### DIFF
--- a/code/math/vecmat.cpp
+++ b/code/math/vecmat.cpp
@@ -470,9 +470,6 @@ float vm_vec_copy_normalize(vec3d *dest, const vec3d *src)
 
 	//	Mainly here to trap attempts to normalize a null vector.
 	if (m <= 0.0f) {
-		mprintf(("Null vec3d in vec3d normalize.\n"
-				 "Trace out of vecmat.cpp and find offending code.\n"));
-
 		dest->xyz.x = 1.0f;
 		dest->xyz.y = 0.0f;
 		dest->xyz.z = 0.0f;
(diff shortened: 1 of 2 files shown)


code/math/vecmat.cpp
```diff
@@ -470,9 +470,6 @@ float vm_vec_copy_normalize(vec3d *dest, const vec3d *src)
 
 	//	Mainly here to trap attempts to normalize a null vector.
 	if (m <= 0.0f) {
-		mprintf(("Null vec3d in vec3d normalize.\n"
```
JohnAFernandez, line 473: This one has helped when we have to debug object updates in multiplayer, so an nprintf for that one would be nice.

Baezon: There's no "Math" category, is "Network" OK?

JohnAFernandez: Yes, thanks

wookieejedi:
Is this one planned to get resolved before the RC phase starts?

Baezon:
Whoops, yeah, kinda forgot.

Goober5000 (approved):
👍

PR #6269, opened 2024-07-25
Beam slash curves

Kestrellius:
Sets up curve parameters to control the slash position and rotational position of type 5 beams over their lifetimes.

### DIFF
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -1079,7 +1079,7 @@ void beam_type_omni_move(beam* b)
 
 	float rotation_amount;
 	if (Weapon_info[b->weapon_info_index].b_info.t5info.rot_curve_idx >= 0) {
-		rotation_amount = Curves[Weapon_info[b->weapon_info_index].b_info.t5info.rot_curve_idx].GetValue(BEAM_T(b));
+		rotation_amount = Curves[Weapon_info[b->weapon_info_index].b_info.t5info.rot_curve_idx].GetValue(BEAM_T(b)) * PI2;
 	} else {
 		rotation_amount = (b->life_total - b->life_left) * b->type5_rot_speed;
 	}
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -1079,7 +1079,7 @@ void beam_type_omni_move(beam* b)
 
 	float rotation_amount;
 	if (Weapon_info[b->weapon_info_index].b_info.t5info.rot_curve_idx >= 0) {
-		rotation_amount = Curves[Weapon_info[b->weapon_info_index].b_info.t5info.rot_curve_idx].GetValue(BEAM_T(b)) * b->type5_rot_speed;
+		rotation_amount = Curves[Weapon_info[b->weapon_info_index].b_info.t5info.rot_curve_idx].GetValue(BEAM_T(b));
 	} else {
 		rotation_amount = (b->life_total - b->life_left) * b->type5_rot_speed;
 	}
--- a/code/weapon/beam.cpp
+++ b/code/weapon/beam.cpp
@@ -1077,18 +1077,31 @@ void beam_type_omni_move(beam* b)
 	vec3d actual_dir;
 	bool no_sweep = vm_vec_dot(&b->binfo.dir_a, &b->binfo.dir_b) > 0.9999f;
 
+	float rotation_amount = (b->life_total - b->life_left) * b->type5_rot_speed;
+
+	if (Weapon_info[b->weapon_info_index].b_info.t5info.rot_curve_idx >= 0) {
+		rotation_amount = Curves[Weapon_info[b->weapon_info_index].b_info.t5info.rot_curve_idx].GetValue(BEAM_T(b)) * b->type5_rot_speed;
+	}
+
 	if (b->rotates) {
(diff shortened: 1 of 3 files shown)


code/weapon/weapons.cpp
```diff
@@ -3078,6 +3078,14 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 				stuff_vec3d(&t5info->end_pos_rand);
 			}
 
+			if (optional_string("+Slash completion over beam lifetime curve:")) {
+				SCP_string curve_name;
+				stuff_string(curve_name, F_NAME);
+				t5info->slash_pos_curve_idx = curve_get_by_name(curve_name);
+				if (t5info->slash_pos_curve_idx < 0)
+					Warning(LOCATION, "Unrecognized slash position curve '%s' for weapon %s", curve_name.c_str(), wip->name);
+			}
```
Baezon, line 3087: Should probably have another warning if `no_translate` is true, i.e. its not actually going to slash.

code/weapon/weapons.cpp
```diff
@@ -3078,6 +3078,14 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 				stuff_vec3d(&t5info->end_pos_rand);
 			}
 
+			if (optional_string("+Slash completion over beam lifetime curve:")) {
```
Baezon, line 3081: It's not a huge deal, but this name is a little odd to me, and I'm wondering if there's a better one. How do you feel about 'Slash interpolation..."? It's more 'jargony' but feels more specific, slash "completion" feels like it isn't clear what precisely it refers to.

code/weapon/beam.cpp
```diff
@@ -1077,18 +1077,32 @@ void beam_type_omni_move(beam* b)
 	vec3d actual_dir;
 	bool no_sweep = vm_vec_dot(&b->binfo.dir_a, &b->binfo.dir_b) > 0.9999f;
 
+	float rotation_amount;
+	if (Weapon_info[b->weapon_info_index].b_info.t5info.rot_curve_idx >= 0) {
+		rotation_amount = Curves[Weapon_info[b->weapon_info_index].b_info.t5info.rot_curve_idx].GetValue(BEAM_T(b)) * b->type5_rot_speed;
+	} else {
+		rotation_amount = (b->life_total - b->life_left) * b->type5_rot_speed;
+	}
```
Baezon, line 1085: I'm wondering if instead perhaps it is better to ignore continuous rotation/rotation speed entirely, simply have the curve define the position entirely. Either you specify a rotation speed per second, or provide a curve to determine its angular position over time.

Baezon (changes requested):

code/weapon/weapons.cpp
Kestrellius: How about "position"? I think "interpolation" relies on a certain way of thinking about the beam's behavior and the underlying code, and I worry that a lot of people just won't have any idea what it means in this situation.

code/weapon/beam.cpp
Kestrellius: I'm inclined against, mostly because having a separate parameter that's a multiplier makes it easier to reuse curves. In general, I think it's a good policy to keep the implied domain (or at least the reference frame) of curves within 0-1 unless there's a particular reason to do otherwise.

Having to make a new curve every time you want a different rotation amount, even though you want the same pattern otherwise, would get fiddly.

Baezon: That's fair, but I feel it's really going to complicate what `$continuous rotation` does. It's either an angular speed per second, or a curve output multiplier. The documentation for both the rotation and the curve become more cumbersome, and its more mental load for the modder. Keeping things logistically simpler for the modder is good too, but it is at least, very simple for the modder to understand to create a new curve for each new beam (in the worst situation).

code/weapon/weapons.cpp
Baezon: I'm ok with "position"

code/weapon/beam.cpp
Kestrellius: Hmm. That's true.

What about multiplying by $continuous rotation * lifetime? That way, the total rotation is consistent between constant rotation and a curve that starts at 0 and ends at 1.

(If I had the option, I would maybe have instead made the curve a rotation speed multiplier, but of course that's not feasible with how it's set up.)

Baezon: I'd be okay with it, but I do want to ask, is it important that constant rotation and a curve be consistent with each other in this way (especially if there are benefits to not doing so)? My impression was that modders, if allowed, would be using and thinking about one or the other and not really both.

Kestrellius: Yeah, that makes sense. My concern is pretty much purely "I'd like there to be a multiplier for the curve so that you don't have to embed whatever specific rotation amount you want in the curve itself."

So, you know what, for the moment let's just put in the raw curve, and possibly in the future a more elegant way of handling that issue will present itself.

Baezon: I *would* multiply by 2pi at least, so 1 is a full rotation. Aint nobody trying to type in tau by hand.

Kestrellius: Actually... Hold on. There _is_ a more robust way to handle this, and I need to think about whether it makes sense here. If it does, implementing this feature in the current way will produce (minor) waste.

Kestrellius: All right, we're just doing it the simple way for now.

Baezon (approved):

PR #6274, opened 2024-07-27
Fix trail overlap

Kestrellius:
This makes a change to the way trails are rendered in order to remedy an issue where trail segments would visibly "overlap" with themselves when twisted. It does this by increasing the number of tris in each segment from 2 to 4. This appears to have a significant performance impact in very trail-heavy scenarios, so the change only takes effect when particle detail is set to max. (I would be entirely willing to switch it to a different detail setting, like model, if that's deemed more appropriate.)

More profiling would be useful in future to figure out exactly where the performance drop is coming from, but for now, under typical circumstances, it should be fine.

### DIFF
--- a/code/render/batching.cpp
+++ b/code/render/batching.cpp
@@ -1017,6 +1017,7 @@ void batching_render_batch_item(primitive_batch_item* item,
 	} else {
 		batched_bitmap_material material_def;
 
+		material_def.set_cull_mode(item->batch_item_info.mat_type != batch_info::FLAT_EMISSIVE_WITH_BACKFACES);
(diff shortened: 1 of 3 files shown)


BMagnu (commented):
LGTM (except for the thing already flagged by the CI), but I've had too much of a hand in this already to be comfortable giving the only approval for a merge

Baezon (changes requested):
I'm very nervous about introducing significant performance burdens on situations which otherwise would not have needed to worry about this issue at all, and don't benefit from it much.

If this symptom is what I think it is, it really should only be noticeable by the player in close proximity, in which case, since this geometry is being produced on the fly, why not only introduce these extra tris when near to the current eye position? That way scenes with lots of distant long and straight trails would be basically unaffected.

PR #6275, opened 2024-07-28
Fix a sound crash

JohnAFernandez:
Reported and fix tested by FotG team:

Missing this logic and assignment in the sound code would cause a crash in specific circumstances.  When a sound file was used more than once in the Sounds table, the sound entry was not being populated with the index to the `Sounds` array.  Later on, when we attempted to access the `Sounds` array to play the sound, the unpopulated index entry was found, triggering an assert.

AFAICT, this buggy behavior was exposed when other sound code was fixed and started accessing the correct field which again, is that same index into the `Sounds` array.


wookieejedi (approved):
Fixes crash and overall works as expected, thanks!

PR #6278, opened 2024-07-29
some timestamp refinements

Goober5000:
1. Make sure the legacy `timestamp_until()` and `timestamp_since()` functions check for any type of invalid timestamp
2. Add a useful `isFinite()` function
3. A timestamp delta on an immediate timestamp is more properly the actual delta, rather than another immediate timestamp

### DIFF
--- a/code/io/timer.cpp
+++ b/code/io/timer.cpp
@@ -180,7 +180,7 @@ UI_TIMESTAMP ui_timestamp() {
 
 TIMESTAMP timestamp_delta(TIMESTAMP stamp, int delta_ms)
 {
-	if (!stamp.isValid() || stamp.isNever())
+	if (!stamp.isFinite())
 		return stamp;
 
 	if (stamp.isImmediate())
@@ -191,7 +191,7 @@ TIMESTAMP timestamp_delta(TIMESTAMP stamp, int delta_ms)
 
 UI_TIMESTAMP ui_timestamp_delta(UI_TIMESTAMP stamp, int delta_ms)
 {
-	if (!stamp.isValid() || stamp.isNever())
+	if (!stamp.isFinite())
 		return stamp;
 
 	if (stamp.isImmediate())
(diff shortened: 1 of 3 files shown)


wookieejedi (approved):
Tests work as expected for me and is straightforward PR. Good to see `isFinite()` being in both TIMESTAMP and `UI_TIMESTAMP`, too.

code/io/timer.cpp
```diff
@@ -180,17 +180,23 @@ UI_TIMESTAMP ui_timestamp() {
 
 TIMESTAMP timestamp_delta(TIMESTAMP stamp, int delta_ms)
 {
-	if (!stamp.isValid() || stamp.isImmediate() || stamp.isNever())
+	if (!stamp.isValid() || stamp.isNever())
```
MageKing17, line 183: This could be replaced with the new function:
```suggestion
	if (!stamp.isFinite())
```

code/io/timer.cpp
```diff
@@ -180,17 +180,23 @@ UI_TIMESTAMP ui_timestamp() {
...
+	if (stamp.isImmediate())
+		return TIMESTAMP(timestamp_ms() + delta_ms);
+
 	return TIMESTAMP(stamp.value() + delta_ms);
 }
 
 UI_TIMESTAMP ui_timestamp_delta(UI_TIMESTAMP stamp, int delta_ms)
 {
-	if (!stamp.isValid() || stamp.isImmediate() || stamp.isNever())
+	if (!stamp.isValid() || stamp.isNever())
```
MageKing17, line 194: Same here.
```suggestion
	if (!stamp.isFinite())
```

MageKing17 (commented):
The new `isFinite()` function could immediately be put to use simplifying conditionals.

code/io/timer.cpp
Goober5000: Also done

PR #6283, opened 2024-08-07
add is-species SEXP operator

Goober5000:
As noted on Discord, there was previously no `is-species` operator.  Now there is.

As a bonus, this also fixes a bug where the pre-existing `is-iff` operator did not work with wings that had not yet arrived.

Tested and handles all possibilities.


wookieejedi (approved):
Useful, and looks good to me

PR #6286, opened 2024-08-08
improve sounds.tbl index ordering

Goober5000:
Although sound indexing is now done by name, mods which index by number (sc. a number that is treated as a name) expect the indexes to be in numeric order.  When multiple levels of -snd.tbm files add mixed retail-style and new-style sounds, this assumption can be violated due to named sounds being placed into indexes that later files expect to be available.

This PR refines the parsing so that sounds which use a numeric index can pre-empt existing named sounds that have been assigned to that index.  Since named sounds can go anywhere, they are just moved to the end of the list.  This allows the ranges 192-199 and 202-249 to be available for numeric indexes as expected under the old style.  MediaVPs named sounds which were previously parsed into those indexes are bumped to the next available index.

Additionally, this PR doubles the range of numbers that are treated as indexes, to more realistically accommodate additional numbered sounds.  (The number/index distinction is rather arbitrary and is primarily used to avoid creating large sections of unused sounds if a mod were to specify a very large number, e.g. 10001, for a sound name.)

Finally, this changes parse warnings to use `error_display()` so that the warning displays the parsed file and line number.

### DIFF
--- a/code/gamesnd/gamesnd.cpp
+++ b/code/gamesnd/gamesnd.cpp
@@ -222,6 +222,8 @@ static EnhancedSoundData Default_sound_priorities[NUM_RETAIL_GAMEPLAY_SOUNDS] =
 
 static const EnhancedSoundData default_enhanced_sound_data(SND_ENHANCED_PRIORITY_MEDIUM_HIGH, 1);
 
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):
Works in my tests and overall looks good.

wookieejedi:
Actually hang on

PR #6288, opened 2024-08-10
Fix renderLeadCurrentTarget

### LINKED ISSUE #5088: Missile lead indicator does not show properly on alternate $Lead Indicator Behavior modes.

JohnAFernandez:
This fixes an edge case in the Average and Multiple modes, where secondary lead indicator rendering was skipped by continue statements when the primary weapons were out of range.   I also added some comments that help people actually understand what this function does as it is very confusing.

Also this remove the return on the else.  I have looked over this function for a long time.  It doesn't need the return because that return will be reached when there are no valid primary weapons, but everything after it only has to do with secondary weapon indicator rendering.

I did test it, but I could use some independent verification that this didn't break any indicator behavior, although it really shouldn't.

Fixes #5088

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -4039,6 +4039,7 @@ void HudGaugeLeadIndicator::renderLeadCurrentTarget()
 	//overwritten if using non-default leadIndicatorBehaviors, this is desired.
 	bank_to_fire = hud_get_best_primary_bank(&prange);
 	wip = &Weapon_info[swp->primary_bank_weapons[bank_to_fire]];
+	bool no_primary_indicator = false;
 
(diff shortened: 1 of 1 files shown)


TRBlount (approved):

wookieejedi:
Will plan to merge this soon unless someone else beats me to it

PR #6292, opened 2024-08-12
safety check for lua_tostring()

### LINKED ISSUE #6290: Missing `.Action` while setting up a LuaSEXP yields a CTD
This script caused a CTD during game start. It wasn't immediately clear that I'd forgotten `.Action` in the assignments.

```lua
#Conditional Hooks

$Application: FS2_Open

$On Game Init: [
  (function()

  local function transform(text)
    local message = mn.Messages[text]
    if message and message:isValid() then
      text = message.Message
    end
    text = ba.replaceTokens(text)
    text = ba.replaceVariables(text)
    return text
  end

  mn.LuaSEXPs['set-directive-text'] = function(event, text)
    event.DirectiveText = transform(text)
  end

  mn.LuaSEXPs['set-directive-keypress-text'] = function(event, text)
    event.DirectiveKeypressText = transform(text)
  end

  end)()
]

#End
```

Goober5000:
The return value of `lua_tostring()` is NULL if (and only if) the string conversion failed.  Many places in the code did not check for this situation, which led to the crash in #6290.

Fixes #6290.

### DIFF
--- a/code/globalincs/utility.h
+++ b/code/globalincs/utility.h
@@ -309,4 +309,12 @@ int find_item_with_name(const VECTOR_T& item_vector, const SCP_string& str)
 	return -1;
 }
 
+template <typename NULLISH_T>
+NULLISH_T coalesce(NULLISH_T possibly_null, NULLISH_T value_if_null)
+{
+	Assertion(value_if_null != nullptr, "value_if_null can never be null itself!");
+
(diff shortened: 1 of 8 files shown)


wookieejedi (approved):
Looks good to me

JohnAFernandez (approved):

PR #6293, opened 2024-08-13
Remove some int3s from offscreen indicator drawing

Baezon:
Under some very odd scenarios the math for doing an offscreen indicator might not get a valid position, but there's 0 reason to `int3()` about it.

Users *have* run into it, and I have not investigated it, but Mjn has to some extent: "This is something with 4k resolution. I suspect an actual engine bug. Something about BtA is causing xpos in the HudGaugeOffscreen::calculatePosition to be larger than 0.. at least in my reproductions.. and that is not expected."

But if this is a matter of an errant frame or two, I'm not sure trying to keep this around is worth the hassle. In both cases, the function will simply abort and not render the indicator.

### DIFF
--- a/code/hud/hudtarget.cpp
+++ b/code/hud/hudtarget.cpp
@@ -6541,7 +6541,7 @@ void HudGaugeOffscreen::calculatePosition(vertex* target_point, vec3d *tpos, vec
 	g3_project_vertex(eye_vertex);
 
 	if (eye_vertex->flags&PF_OVERFLOW) {
-		Int3();			//	This is unlikely to happen, but can if a clip goes through the player's eye.
+		//	This is unlikely to happen, but can if a clip goes through the player's eye.
 		Player_ai->target_objnum = -1;
 		if (!in_frame)
 			g3_end_frame();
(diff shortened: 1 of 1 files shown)


code/hud/hudtarget.cpp
```diff
@@ -6541,7 +6541,7 @@ void HudGaugeOffscreen::calculatePosition(vertex* target_point, vec3d *tpos, vec
 	g3_project_vertex(eye_vertex);
 
 	if (eye_vertex->flags&PF_OVERFLOW) {
-		Int3();			//	This is unlikely to happen, but can if a clip goes through the player's eye.
+		//	This is unlikely to happen, but can if a clip goes through the player's eye.
 		Player_ai->target_objnum = -1;
```
JohnAFernandez, line 6545: I'm not sure this should stay either.  Removing the player's target here doesn't make sense, unless I'm reading this wrong.

Baezon: yeah good point

JohnAFernandez (approved):

PR #6295, opened 2024-08-13
Upgraded Asteroid Field SEXPs

MjnMixael:
Follow-up to #6021 which enhanced asteroid/debris fields to allow more than three types to be active at once. This PR replaces the older set-asteroid-field and set-debris-field SEXPs with the newer config-asteroid-field and config-debris-field sexps that allow selecting more than three types to be active.

One core difference here is that the sexp for Active Asteroid Fields no longer sets the field targets within the same SEXP. A new config-field-targets SEXP has been added that allows adding/removing field targets arbitrarily which actually ends up being kind of a nice side-effect upgrade to the whole system.

The old SEXPs are deprecated and hidden, but still function as they used to according to my tests.

### DIFF
--- a/code/parse/sexp.cpp
+++ b/code/parse/sexp.cpp
@@ -16381,7 +16381,7 @@ void sexp_toggle_asteroid_field(int n)
 	Asteroids_enabled = is_sexp_true(n);
 }
 
-void sexp_set_asteroid_field(int n)
+void sexp_set_asteroid_field(int n, bool new_sexp)
 {
 	bool is_nan, is_nan_forever;
 
@@ -16405,25 +16405,27 @@ void sexp_set_asteroid_field(int n)
 
 	SCP_vector<SCP_string> asteroid_types;
 
-	if (n >= 0) {
-		if (is_sexp_true(n)) {
-			asteroid_types.push_back("Brown");
+	if (!new_sexp) {
+		if (n >= 0) {
+			if (is_sexp_true(n)) {
+				asteroid_types.push_back("Brown");
+			}
+			n = CDR(n);
 		}
-		n = CDR(n);
-	}
 
-	if (n >= 0) {
-		if (is_sexp_true(n)) {
-			asteroid_types.push_back("Blue");
+		if (n >= 0) {
+			if (is_sexp_true(n)) {
+				asteroid_types.push_back("Blue");
+			}
+			n = CDR(n);
 		}
-		n = CDR(n);
-	}
 
-	if (n >= 0) {
-		if (is_sexp_true(n)) {
-			asteroid_types.push_back("Orange");
+		if (n >= 0) {
+			if (is_sexp_true(n)) {
+				asteroid_types.push_back("Orange");
+			}
+			n = CDR(n);
 		}
-		n = CDR(n);
 	}
 
 	int o_minx = -1000, o_miny = -1000, o_minz = -1000;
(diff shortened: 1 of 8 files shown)


code/parse/sexp.cpp
```diff
@@ -16458,13 +16487,30 @@ void sexp_set_asteroid_field(int n)
...
 
 	SCP_vector<SCP_string> targets;
-	if (n >= 0) {
-		for (; n >= 0; true ? n = CDR(n) : n = -1) {
-			auto ship_entry = eval_ship(n);
-			if (!ship_entry)
-				continue;
+	if (!new_sexp) {
+		if (n >= 0) {
+			for (; n >= 0; true ? n = CDR(n) : n = -1) {
```
BMagnu, line 16492: ``true ? n = CDR(n) : n = -1`` seems weird.
That just boils down to ``n = CDR(n)``. Either it should probably be simplified to that, or something may be amiss with the conditional here.

code/parse/sexp.cpp
```diff
@@ -16458,13 +16487,30 @@ void sexp_set_asteroid_field(int n)
...
+					continue;
 
-			targets.push_back(ship_entry->name);
+				targets.push_back(ship_entry->name);
+			}
+		}
+	} else {
+		if (n >= 0) {
+			auto list = get_list_valid_asteroid_subtypes();
+			for (; n >= 0; true ? n = CDR(n) : n = -1) {
```
BMagnu, line 16503: Same here

code/parse/sexp.cpp
```diff
@@ -16514,22 +16561,31 @@ void sexp_set_debris_field(int n)
...
-	int debris1 = -1;
-	if (n >= 0) {
-		debris1 = get_asteroid_index(CTEXT(n));
-		n = CDR(n);
-	}
+	SCP_vector<int> debris_types;
+	if (!new_sexp) {
+		if (n >= 0) {
+			int debris1 = get_asteroid_index(CTEXT(n));
+			if (debris1 > 0) {
```
BMagnu, line 16568: Other places using this function check for ``>= 0`` instead, should this be here as well?

code/parse/sexp.cpp
```diff
@@ -16514,22 +16561,31 @@ void sexp_set_debris_field(int n)
...
+		}
 
-	int debris2 = -1;
-	if (n >= 0) {
-		debris2 = get_asteroid_index(CTEXT(n));
-		n = CDR(n);
-	}
+		if (n >= 0) {
+			int debris2 = get_asteroid_index(CTEXT(n));
+			if (debris2 > 0) {
```
BMagnu, line 16576: Same here?

code/parse/sexp.cpp
```diff
@@ -16514,22 +16561,31 @@ void sexp_set_debris_field(int n)
...
+			n = CDR(n);
+		}
 
-	int debris3 = -1;
-	if (n >= 0) {
-		debris3 = get_asteroid_index(CTEXT(n));
-		n = CDR(n);
+		if (n >= 0) {
+			int debris3 = get_asteroid_index(CTEXT(n));
+			if (debris3 > 0) {
```
BMagnu, line 16584: Same here?

BMagnu (changes requested):
A few questions and possible tidyups

code/parse/sexp.cpp
MjnMixael: I'm not sure what I was smoking when I put that there. I'm sure I copied it from the old asteroid sexp but given I wrote that too, I'm still not sure. Bet I copied something similar from somewhere else, modified it for testing, and promptly forgot about it.

Either way, yes you're right. Fixed.

BMagnu (approved):

code/parse/sexp.cpp
MjnMixael: Ah yes, good catch. Fixed.

Issue #6287, opened 2024-08-09
Performance regression tied to options manager


EatThePath:
An exploratory profiling of master found approximately 30% of runtime in the options manager code, checking the deferred rendering option. I have not done a hard A/B overall performance test for this yet but until proven otherwise I think this warrants a critical status.

This has a few different ways to be attacked. I think we should do both of the first two at least. None of them should be that hard, but I don't know if or when I'll tackle them personally, so anyone else can be my guest to address this in whole or in part.

1. In my test case the options manager is not finding any saved setting and falling back to default. In `OptionsManager::getValueFromConfig()` a missing value is communicated by raising an exception which I believe Option.getValue() catches, and json deserializer follows this pattern as well. This, as a TODO comment in the code points out, is not a good idea as part of normal code flow. `tl::optional<T>` is ideal for this use, and already present in the codebase(it's all over the animations code).
2. We call `light_deferred_enabled()` in several places, at different depths of the drawing code, and it checks the option manager. To insulate drawing from any options manager overhead, we should probably do that check at the start of frames and save the value to a simple variable that `light_deferred_enabled()` returns. Then `light_deferred_enabled()` will probably even get optimized into oblivion by the compiler, if I were to guess.
3. In at least one case, `model_draw_list()`, we call `light_deferred_enabled()` directly, and also call other checks that call it again themselves, in the same line. Not doing that would help, but it would also be a negligible issue if the previous two items are already handled.

EatThePath:
I've got an implementation of approach 1 that seems to work, and needs a bit of a cleanup before submission still. The result is good, but there remains a measurable overhead from constant attempts to look up a config value by name string. Approach 2. will definitely also be needed to get out of this fully, and if I do it I'll approach it as a separate PR as the two change sets won't really overlap at all.

PR #6296, opened 2024-08-14
Switch options manager from signaling empty keys with exceptions to use optional<> 

EatThePath:
This cleans up an old TODO and addresses(partially) #6287. Exceptions are no longer routine for unset options, but other exception sources are possible in getting options so the handling is still present.

I have run the game with these changes and it still appears to function properly in general and respect my options and defaults, but if there's any in-depth stress testing of the options manager that should be undertaken to verify it I am unsure what it would be.

### DIFF
--- a/code/options/Option.cpp
+++ b/code/options/Option.cpp
@@ -123,7 +123,7 @@ OptionBase::OptionBase(SCP_string config_key, SCP_string title, SCP_string descr
 }
 
 //Return the option value from the config
-std::unique_ptr<json_t> OptionBase::getConfigValue() const { return _parent->getValueFromConfig(_config_key); }
+tl::optional<std::unique_ptr<json_t>> OptionBase::getConfigValue() const { return _parent->getValueFromConfig(_config_key); }
 
 //Return the option expert_level value
 ExpertLevel OptionBase::getExpertLevel() const { return _expert_level; }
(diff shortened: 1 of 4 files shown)


MjnMixael:
I'm not super familiar with the tl:: library so I'm going to refrain from an actual review. However, if the code does what it looks like it does then this looks like a fine way to address the issue. The only other tests I might ask for (if you haven't already done them) are:

- New pilot with $enable in-game options: TRUE
- New pilot with $enable in-game options: FALSE (this one should be quick and should bypass Options Manager entirely)
- New mod with $enable in-game options: TRUE
- New mod with with $enable in-game options: FALSE (this one should be quick and should bypass Options Manager entirely)

By "new" I mean with no previous data existing for the pilot/mod in appdata. Specifically checking for pilot files and mod.ini files. Note that mods get their own .inis now and fallback to the old default .ini if it can't be found. So a test where the default .ini is missing might also be warranted.

EatThePath:
I haven't tested all those cases, I'll see what I can do, thanks.
The only SCP person I'm sure has some working knowledge of tl::optional is @BMagnu.

BMagnu (commented):
Do we still need the try catch around the getConfigValue with this implementation?

EatThePath:
> Do we still need the try catch around the getConfigValue with this implementation?

getValueFromConfig, which getConfigValue calls, still has a possible throw path just below my changes. It seemed sufficiently separate that I wanted to leave it alone, so keeping the catch seems prudent.

BMagnu (approved):

PR #6299, opened 2024-08-15
Fix sorting of older but active directives

### LINKED ISSUE #6294: Bug: completed directives becoming incomplete again displaces extent incomplete directives in list
With the way the directives list currently functions, new directives appear at the bottom of the list and are pushed upwards as new ones are added. If directives make it to the top of the list without being completed, completed directives beneath them are pushed above the incomplete directives and off of the screen. This is a little complicated, so example below. O is an incomplete directive, X is a completed one.

O
O
X
X
X
X
O
O

If a third O were to appear at the bottom, the X closest to the top would be moved above the uppermost O, off of the list.

In certain cases however, this fails to work correctly, and an O is pushed off of the list instead.

Specifically in the case of a directive which was completed much earlier becoming valid as an active directive again. The most textbook example of this would be a wing which was marked destroyed much earlier in a mission returning.

In this case, I _think_ FSO pulls this directive from above rather than drawing it at the bottom as though it were a new directive, and in doing so, appears to displace currently incomplete directives. This makes a good amount of sense if the directives list is not full, but if it is full, becomes a real problem.

My proposal is to add a game flag to draw "reactivated" directives at the bottom of the list as though they're new, displacing completed ones the way that a new directive does. This will alter behavior even in retail missions, hence the need for a flag to maintain legacy behavior.

I am not entirely sure if this is an accurate summation of how directives function. It is a best guess. There is absolutely a bug happening, but I may not be completely correct about the mechanics of how it is occurring. I will say that I've reached out to several SCP folks, and none of them have been able to give me a completely concrete answer on what intended behavior is, and as such I think this one should probably be marked discussion as well.

BMagnu:
FSO should, based on code comments and common sense, display the five must current directives. But, if there are directives older then that which are still active, then these should be preferred over the oldest directives that are already completed and still shown.

Unfortunately, the logic to do that has been bugged since retail, but since you need an active directive that's sixth oldest or older to notice the issue, it probably slipped past so far.

The issues with the prior logic were:
- Usage of an incorrect loop iterator variable, resulting in replacement of the n-th shown directive in order to display the n-th older-but-active directive, regardless of the state of shown directives.
- Since the directive that is shown instead of the replaced directive is older than all other directives, it should be shown on the top of the list. For this, the directives above the slot that is freed are shifted down. There was an off-by-one error, causing the directive above the slot to be overwritten, while the slot itself (which is "free" and should be used by the directives still shown) was not overwritten.
- The directive that is replaces is not deleted, it's just switched with the one that is to-be-shown. The prior code didn't use the directive to be replaced for this however, but the last directive in the list.
- Volition never considered the case, that there is no room available. If the logic had worked as described in comments, then a case where more than ``Max_directives`` were active at once, the logic would have written out of bounds of the directives display array

fixes #6294

### DIFF
--- a/code/mission/missiontraining.cpp
+++ b/code/mission/missiontraining.cpp
@@ -507,10 +507,9 @@ void sort_training_objectives()
 		}
 	}
 
-
-	int slot_idx, unkn_vis, last_directive;
 	// go through list and bump as needed
 	for (i=0; i<num_offset_events; i++) {
+		int slot_idx, unkn_vis, slot_directive;
 
 		// find most recent directive that would not be shown
 		for (unkn_vis=offset-1; unkn_vis>=0; unkn_vis--) {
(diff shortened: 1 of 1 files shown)


wookieejedi (approved):

PR #6300, opened 2024-08-15
Accessibility IFF colors

MjnMixael:
Adds a feature BtA wanted that I completely forgot about until today. This enables an new in-game option to enable Accessibility colors for IFFs. The colors are basically overrides of the normal colors and are defined in the iff_defs.tbl in the same way as regular colors. If Accessibility Colors are enabled then those are the colors returned whenever an IFF color is requested from anywhere.

This requires mods to opt-in and fully support the feature. As such, there is a global toggle `$Accessibility Supported:` in the iff_defs.tbl. This is false by default and if it is false the Accessibility option will not even show in the in-game options menu. It must be explicitly enabled and colors defined.

### DIFF
--- a/code/iff_defs/iff_defs.cpp
+++ b/code/iff_defs/iff_defs.cpp
@@ -175,13 +175,14 @@ void resolve_iff_data()
 			int target_iff = iff_lookup(observed_color.iff_name);
 
 			// valid?
-			if (target_iff >= 0)
+			if (target_iff >= 0) {
 				iff->observed_color_map[target_iff] = observed_color.color_index;
-			else
+			} else{
 				Warning(LOCATION,
 					"Observed color IFF %s not found for IFF %s in iff_defs.tbl!\n",
 					observed_color.iff_name,
 					iff->iff_name);
+			}
 		}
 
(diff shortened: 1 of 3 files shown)


Baezon (changes requested):
As far as I can tell this is just a single alternate set of colors right? It's not a huge deal, but as I understand it, accessibility colors often some in sets, with different sets of colors suited for the various kinds of color vision deficiency. I'm wondering if it might be better to lay some of the groundwork, at least table-wise, to make sure that isn't too awkward to do later or something.

MjnMixael:
> As far as I can tell this is just a single alternate set of colors right? It's not a huge deal, but as I understand it, accessibility colors often some in sets, with different sets of colors suited for the various kinds of color vision deficiency. I'm wondering if it might be better to lay some of the groundwork, at least table-wise, to make sure that isn't too awkward to do later or something.

This PR suits BtA's specific needs to be able to set two high contrast colors that works work for pretty much all color deficiencies. If someone else wants to take this further then that's for another PR.

Baezon (approved):
Alright, I'll defer to you on it, just thinking about the future.

PR #6301, opened 2024-08-15
Add bindings and shaders for bent normals used in ambient occlusion

devthenecro:
As discussed in HL forums https://www.hard-light.net/forums/index.php?topic=99354.0

Example of production use: https://dev.epicgames.com/documentation/en-us/unreal-engine/bent-normal-maps?application_version=4.27

This feature is used to improve accuracy of ambient occlusion by using pre-generated normal maps that contain vectors pointing in least occluded direction.
Reflective light is handled by calculating dot product between the reflected vector and the normal, and then mixing this with ambient occlusion map.
Diffuse ambient lighting is simply shifted to this new vector.
Primary contribution of this feature is however direct lighting, as this allows AO to work with direct lighting without creating nasty artifact or looking like a decal.

Example
https://imgsli.com/Mjg4MTUy


qazwsxal:
Just had a little look at the shaders, my understanding of bent normals was that they were meant to be used for indirect lighting only as an approximation to GI, but it looks like you've applied them to the deferred/direct lights too? (deferred-f.sdr line 210)

devthenecro:
Bent normals do hold occlusion directionality, while the AO map holds the scalar value of the occlusion. Together it can approximate a cone of visibility, so it should be possible to check this to create decent approximation of the direct light occlusion.

See for example https://github.com/GameTechDev/XeGTAO/blob/a5b1686c7ea37788eeb3576b5be47f7c03db532c/Source/Rendering/Shaders/Lighting/vaPointLight.hlsli#L33

What I still need to do is to pass the entire vector and not just the value at view angle, so that the occlusion can be relatively accurate for diffuse too.

PR #6302, opened 2024-08-15
Fixes CTD with new Friendly Fire Feature

wookieejedi:
On current master I was hitting a CTD when firing a weapon with a shockwave at close range (ie player is within the shockwave zone). This code is part of the new friendly fire feature from #6265. The minidump showed `other_obj` was `null`, so this PR adds some safety checks.

Tested and this fixes the crashes.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -7456,8 +7456,8 @@ void weapon_do_area_effect(object *wobjp, shockwave_create_info *sci, vec3d *pos
 
 			// if this is friendly fire, we check for the friendly fire cap values
 			if (wp->team == shipp->team) {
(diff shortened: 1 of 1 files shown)


Goober5000 (changes requested):
I believe this is actually the wrong fix, and that almost all of the `other_obj` references should instead be `objp`.  The checks here are dealing primarily with the *objp* receiving the area effect, not the *other_obj* that was actually impacted.  In fact, for clarity, `other_obj` should be renamed something like `impacted_obj`.

wookieejedi:
Just for convenience I copied the part of 6265 that was new in this file:
Changed Sections


Also, renaming `other_obj` to `impacted_obj` would of course touch a lot of this entire function, but if that's the agreed upon course of action I can change it.

Goober5000:
Most of the current `other_obj` references should be `objp`, so after that is fixed, renaming `other_obj` will not touch much.

Goober5000 (approved):
Quite a tricky bug to solve, but this looks good now. 👍

Issue #4294, opened 2022-05-21
FR: FRED - lock object's position/heading flags


MitoPL:
In FRED, there is a "Does not move" checkbox in Ship editor, Misc category. In the past this did not prevent ships from turning, just from moving from their coordinates. It seems like this characteristic was used by some FREDers to create ships that would turn to engage enemies but not move from their position, in tandem with ai-chase and unlocked turrets.

One such example is in Aftermath Reboot, mission Act6-g, "Keeping our gains". There are two cruisers acting as artillery, with the player being able to order them to attack currently targeted capital ship with keypresses, and the mission just runs a clear-goals and then add-goal "ai-chase" to make said ships turn to a designated target while still not moving from their position. It's a surprisingly simple solution, and during my attempts at fixing said mission to work properly with modern builds, I learned that there isn't really an easy and supported way to do that - beyond custom ship movement SEXPs or dynamically calculating set-object-facing transition times for the proccess to look believeable.

My suggestion would be to introduce two flags in place of the current one - let's call them "Lock position" and "Lock heading" - the first one alone would be a quick and easy fix for the situation I faced, using the two in tandem could easily replace the current "Does not move" flag, and I think using the latter only, beyond some cosmetic uses, could possibly simplify building simple convoys - single waypoint for the entire convoy, order all vessels to follow a waypoint and then lock the heading for all of them. Although I'm not particularily sure how the AI would behave when nearing said waypoint.

wookieejedi:
Just a quick note: `Although I'm not particularly sure how the AI would behave when nearing said waypoint.` AI wants to have it's center point pass through the waypoint, so they will all converge on it and not stay in a nice formation.

MitoPL:
Thanks for the insight. So quick explanation:

I'm suggesting that said convoy ships could not change their orientation in said use case, and just fly forward, parallel to each other. I'm more worried about what would happen if the AI was getting rather close to the waypoint in question - with the waypoint being, for example, on a 45deg angle from a ship's heading. Would the ship just trudge on at same speed, or would it slow down in because of its forward speed component lowering, as if it attempted to turn? What if said ship overshot a waypoint for some reason, would it just stop moving forward? (I would assume the FREDer in this case would use a departure cue of x distance from object, but let's assume an oversight somewhere).

[...]

PR #6304, opened 2024-08-17
implement don't-change-position and don't-change-orientation

Goober5000:
Per #4294, this adds two new flag that, together, are equivalent to the legacy "immobile" flag but individually provide more fine-grained control.  Since the physics for position and orientation have a bit of overlap, the new flag functionality simply reverts whatever position or orientation changes were done by the physics sim.

In the FRED and qtFRED ship flags dialogs, the "immobile" flag (aka "Does Not Move") is replaced by the two new flags, but in FSO the situation is more complicated.  The old and new flags coexist and have overlapping functionality, so the mission designer must be careful to set and unset the appropriate flag (that is, it is possible for both "immobile" and "don't-change-position" to be set at the same time, in which case the ship cannot change position without both flags being unset).  The `set-immobile` and `set-mobile` legacy SEXPs ensure that the new flags are removed so as not to conflict with the legacy flag, but the `alter-ship-flag` and `are-ship-flags-set` SEXPs and the `getFlag` and `setFlag` script functions use exactly what the FREDder/scripter specifies and nothing else.

(I attempted to completely replace "immobile" with the two newer flags, but it proved intractable to do this while maintaining reverse compatibility: retrofitting single-flag functions to apply dual flags for one particular situation made the functions harder to understand as well as introducing many edge cases and opportunities for bugs or "gotchas".)

One a related but tangential note, this PR also moves some non-physics code out of the physics function where it both makes more sense and also won't be unintentionally bypassed by using these flags.

Implements #4294.


code/object/object.cpp
```diff
@@ -1593,15 +1560,27 @@ void obj_move_all(float frametime)
 				// physics
 				obj_move_call_physics(objp, frametime);
 			}
-		} else {
-			// make sure velocity is always 0 for immobile things!
+		}
+
+		// if the object isn't supposed to move, roll back the movement
```
Baezon, line 1565: I would be slightly more specific here, these *should* already be the case. But are here as just as a last-line-of-defense to ensure collision or anything else in the future doesn't accidentally let velocity or position changes 'leak' out. We don't want to accidentally imply that movement, despite these flags, is expected and therefore needs to be rolled back.

Baezon (changes requested):

code/object/object.cpp
Goober5000: It's not really a last line of defense; in many cases it's a first line of defense.  Without these lines, a ship flying waypoints would continue to move, and a ship blasted by a shockwave would change position.  (There might be other cases too.)  So yes, movement is expected here.

Baezon: A ship wouldn't be able to turn because of the skip in `ai_turn_towards_vector` and wouldn't be able to move under it's own power because of the skip immediately before this section, although a shockwave would still go through. This section is quite new, and so for a long immobile went without this and still sort-of worked. There *are* cases for which this is necessary, but I would consider them unusual and undesirable, I don't want other coders seeing this and thinking that because this exists, immobility doesn't need to be handled elsewhere, because it'll just get rolled back in the end.

Goober5000: The section isn't new; compare the previous version of object.cpp.  Previously, `obj_move_call_physics` was skipped and velocity (both linear and rotational) was zeroed.  The new code is similar, but treats linear and rotational velocity individually, and only skips physics if *both* flags are in effect.  I considered modifying `obj_move_call_physics` to check the flags individually, but the code just got too tangled.

To the original point, I can enhance the code comment.  What about something like this:
```
// If the object isn't supposed to move, roll back any movement that occurred.  Most of the movement should already have been skipped, but this ensures complete immobility.
```

Baezon: I'd be fine with that.

Goober5000: Done.

Baezon (approved):

PR #6306, opened 2024-08-19
Adds warning if cond-impact armor type not found.

Kestrellius:
Previously the modder wouldn't be warned if a conditional impact couldn't find the specified armor type and set it to -1. This fixes that.

### DIFF
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -2383,6 +2383,9 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		required_string("+Armor Type:");
 			stuff_string(fname, F_NAME, NAME_LENGTH);
 		armor_index = armor_type_get_idx(fname);
+		if (armor_index < 0) {
+			Warning(LOCATION, "Armor type '%s' not found for conditional impact in weapon %s!", fname, wip->name);
+		}
 		parse_optional_float_into("+Min Health Threshold:", &ci.min_health_threshold);
 		parse_optional_float_into("+Max Health Threshold:", &ci.max_health_threshold);
 		parse_optional_float_into("+Min Angle Threshold:", &ci.min_angle_threshold);
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -2380,12 +2380,18 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		ci.min_angle_threshold = 0.f;
 		ci.max_angle_threshold = 180.f;
 		ci.dinky = false;
+
+		bool invalid_armor = false;
 		required_string("+Armor Type:");
 			stuff_string(fname, F_NAME, NAME_LENGTH);
 		if (!stricmp(fname, "NO ARMOR")) {
 			armor_index = -1;
 		} else {
 			armor_index = armor_type_get_idx(fname);
+			if (armor_index < 0) {
+				Warning(LOCATION, "Armor type '%s' not found for conditional impact in weapon %s!", fname, wip->name);
+				invalid_armor = true;
+			}
 		};
 		parse_optional_float_into("+Min Health Threshold:", &ci.min_health_threshold);
 		parse_optional_float_into("+Max Health Threshold:", &ci.max_health_threshold);
@@ -2400,10 +2406,8 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 			ci_vec.insert(ci_vec.end(), existing_cis.begin(), existing_cis.end());
 		}
 		ci_vec.push_back(ci);
-		if (armor_index >= 0) {
+		if (!invalid_armor) {
 			wip->conditional_impacts[armor_index] = ci_vec;
-		} else {
-			Warning(LOCATION, "Armor type '%s' not found for conditional impact in weapon %s!", fname, wip->name);
 		}
 	}
 
--- a/code/weapon/weapons.cpp
+++ b/code/weapon/weapons.cpp
@@ -2375,21 +2375,17 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 	while (optional_string("$Conditional Impact:")) {
 		int armor_index;
 		ConditionalImpact ci;
-		ci.min_health_threshold = 0.0;
-		ci.max_health_threshold = 1.0;
-		ci.min_angle_threshold = 0.0;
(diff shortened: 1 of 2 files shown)


Goober5000 (changes requested):
A few things:
1. You've mentioned that "no armor" is a valid condition, so there should be a way to explicitly indicate "no armor" without triggering the warning.
2. If the armor type is *intended* to be valid but isn't found, the entry should not be added to `wip->conditional_impacts`.
3. While we're doing a follow-up to conditional impacts, there is a `wpB->lifeleft = 0.001f;` in collideweaponweapon.cpp that was missed in #6254.  This should be removed to be consistent with the other removals.

Kestrellius:
All right, made those changes. Now, hitting a target with no defined armor type will trigger the "no armor" condition, if the weapon has one. It's still the case that if no conditions are met -- the target has armor but it isn't listed in the conditions, or the weapon is detonating without hitting anything, or any other reason -- the weapon will proceed to activate the $Impact Effect, if present.

I did observe in testing a slightly confusing behavior where shooting a ship during its death roll results in no conditions being met. I'd seen this previously, and chalked it up to the dying ship being considered armorless debris -- but this time, hitting an armorless object like debris resulted in "Conditional Effect A", whereas hitting the death-rolling ship resulted in "Impact Effect B". In other words, the death-rolling ship behaved like it had armor, but _different_ armor than the ship normally had. Not sure what's up with that.

Or, I suppose, the other, probably more realistic, possibility is that an impact against a death-rolling ship isn't considered to be an impact with an object. Weird, but more plausible.

code/weapon/weapons.cpp
```diff
@@ -2382,10 +2382,15 @@ int parse_weapon(int subtype, bool replace, const char *filename)
 		ci.dinky = false;
 		required_string("+Armor Type:");
 			stuff_string(fname, F_NAME, NAME_LENGTH);
-		armor_index = armor_type_get_idx(fname);
-		if (armor_index < 0) {
-			Warning(LOCATION, "Armor type '%s' not found for conditional impact in weapon %s!", fname, wip->name);
-		}
+		if (*fname == *"NO ARMOR") {
```
Goober5000, line 2385: C-style strings require a comparison function, such as `stricmp`, which stands for case-insensitive string compare.  To test for equality, check `stricmp(string1, string2) == 0`, or more succinctly, `!stricmp(string1, string2)`.

More information here.  This covers `strcmp` which is the case-sensitive version, but `stricmp` works the same way:
https://en.cppreference.com/w/cpp/string/byte/strcmp

code/weapon/weapons.cpp
```diff
@@ -2382,10 +2382,15 @@ int parse_weapon(int subtype, bool replace, const char *filename)
...
+			armor_index = -1;
+		} else {
+			armor_index = armor_type_get_idx(fname);
+			if (armor_index < 0) {
+				Warning(LOCATION, "Armor type '%s' not found for conditional impact in weapon %s!", fname, wip->name);
+				continue;
```
Goober5000, line 2391: The `continue` will skip the rest of the parsing.  That will cause future conditional impacts (or even future lines in the weapon entry) to throw parse errors.  Instead, the rest of the parsing should continue as usual, but the entry should not be added to `wip->conditional_impacts`.

Goober5000 (changes requested):
Two review comments in-line with the code.

Also,
> I did observe in testing a slightly confusing behavior where shooting a ship during its death roll results in no conditions being met. I'd seen this previously, and chalked it up to the dying ship being considered armorless debris -- but this time, hitting an armorless object like debris resulted in "Conditional Effect A", whereas hitting the death-rolling ship resulted in "Impact Effect B". In other words, the death-rolling ship behaved like it had armor, but _different_ armor than the ship normally had. Not sure what's up with that.

Well that's weird.  Sounds like it will be necessary to trace through the code to find out what's going on.  I can walk you through that if you want.

Kestrellius:
> Two review comments in-line with the code.
> 
> Also,
> 
> > I did observe in testing a slightly confusing behavior where shooting a ship during its death roll results in no conditions being met. I'd seen this previously, and chalked it up to the dying ship being considered armorless debris -- but this time, hitting an armorless object like debris resulted in "Conditional Effect A", whereas hitting the death-rolling ship resulted in "Impact Effect B". In other words, the death-rolling ship behaved like it had armor, but _different_ armor than the ship normally had. Not sure what's up with that.
> 
> Well that's weird. Sounds like it will be necessary to trace through the code to find out what's going on. I can walk you through that if you want.

Okay, I figured it out.

So conditional impacts have, optionally, a minimum and maximum health threshold, so that you can make an effect only play if the target is above or below a certain health percentage. This is checked against the target's current hitpoints divided by its maximum health. If not defined, the min and max default to 0 and 1, to cover the entire range and be valid for all possible health quantities.

...except that apparently, during a ship's death role, _its hitpoints can go negative_. Which means that the health fraction is negative, which means that it's below the minimum threshold, so the conditional effect for the ship's armor type isn't valid. Switching the default minimum to float `lowest()` fixes it. I guess I'll also change the default max to float `max`, just in case there's some bizarre edge case where an object can have extra health?

Not sure if, uh, anything should be done about objects having negative health. Seems like the sort of thing that should maybe not happen, but apparently it's worked fine all these years?

code/weapon/weapons.cpp
Goober5000: The `continue` behaves exactly as you linked.  It proceeds to the next iteration, as you said, and it also skips the rest of the parsing, as I said.  But this leaves the entry fields *unparsed*.  That means they will stick around to interfere with whatever is parsed next.

Also, this new change isn't correct.  It won't add the legitimate no-armor case, and it will also warn on both the legitimate and illegitimate no-armor cases.  You were closer the previous time.  What you need to do is check the `armor_index` when first parsed, and warn when first parsed, but then use a flag to skip adding the entry to `wip->conditional_impacts` if it is an illegitimate no-armor case.

Goober5000:
Ah, good sleuthing.  Yes, object health can go negative when an object is destroyed.  This is not only allowed, but expected.  Shooting an object after it "dies" causes the death roll explosion to happen more quickly.

code/weapon/weapons.cpp
Kestrellius: Oh, okay, I see.

Goober5000 (approved):

PR #6311, opened 2024-08-22
fix waypoint marking

Goober5000:
When the Show Waypoints option is toggled, the marking was not updated as it was for the other options (Show Ships, Show Starts, Show IFFs).  Fix this by calling `correct_marking()`.


Kestrellius (approved):
This looks probably correct to the best of my understanding!

wookieejedi (approved):

PR #6314, opened 2024-08-23
Generic random distributions, including curve-based

Kestrellius:
This PR 1. adds support for probability distributions defined by curves and 2. allows all tabled instances of uniform random ranges to use uniform, normal, or curve-based ranges (without breaking backwards-compatibility, of course).

@Baezon will probably want to see this.

### DIFF
--- a/code/math/curve.h
+++ b/code/math/curve.h
@@ -37,6 +37,9 @@ public :
 	//Get
 	float GetValue(float x_val) const;
 
+	// Get
+	float GetValueIntegrated(float x_val) const;
+
 	//Set
 	void ParseData();
 };
--- a/code/math/curve.h
+++ b/code/math/curve.h
@@ -44,6 +44,6 @@ public :
 extern SCP_vector<Curve> Curves;
 
 extern int curve_get_by_name(const SCP_string& in_name);
-extern int pdf_to_cdf(Curve curve);
+extern int pdf_to_cdf(int curve);
 extern void curves_init();
 
--- a/code/math/curve.h
+++ b/code/math/curve.h
@@ -1,3 +1,5 @@
+#pragma once
+
 /*
  * Copyright (C) Volition, Inc. 1999.  All rights reserved.
  *
--- a/code/math/curve.h
+++ b/code/math/curve.h
@@ -44,5 +44,6 @@ public :
 extern SCP_vector<Curve> Curves;
 
 extern int curve_get_by_name(const SCP_string& in_name);
+extern int pdf_to_cdf(Curve curve);
 extern void curves_init();
 
--- a/code/utils/RandomRange.h
+++ b/code/utils/RandomRange.h
@@ -533,9 +533,10 @@ class ParsedRandomRange {
 	
 
   public:
-	ParsedRandomRange(variant random_range)
+	  template<typename T>
+	ParsedRandomRange(T&& random_range)
 	{
-		m_random_range = random_range;
+		m_random_range = std::forward<T>(random_range);
 	}
 	ParsedRandomRange() {
 		m_random_range = UniformFloatRange();
--- a/code/utils/RandomRange.h
+++ b/code/utils/RandomRange.h
@@ -165,7 +165,7 @@ class BoundedNormalDistribution {
 	inline BoundedNormalDistribution() : BoundedNormalDistribution(param_type{std::normal_distribution<float>::param_type{0.5f, 1.f}, 0.f, 1.f}) {}
 	inline BoundedNormalDistribution(param_type curve) : m_param(curve) {}
 	inline void reset() {}
-	inline param_type param()
+	inline param_type param() const
 	{
 		return m_param;
 	}
@@ -359,7 +359,7 @@ class CurveNumberDistribution {
 	inline CurveNumberDistribution() : CurveNumberDistribution(param_type{-1, NAN, NAN}) {}
 	inline CurveNumberDistribution(param_type curve) : m_param(curve) {}
 	inline void reset() {}
-	inline param_type param()
+	inline param_type param() const
 	{
 		return m_param;
 	}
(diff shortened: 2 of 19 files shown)


code/utils/RandomRange.h
```diff
@@ -261,4 +346,185 @@ UniformRange<Value> parseUniformRange(Value min = std::numeric_limits<Value>::mi
...
 }
+
+using CurveFloatRange = RandomRange<float, CurveNumberDistribution, std::minstd_rand>;
+
+inline CurveFloatRange parseCurveFloatRange(float min = std::numeric_limits<float>::lowest()/2.1f, float max = std::numeric_limits<float>::max()/2.1f) {
+	CurveNumberDistribution::param_type curve_params;
+
+	optional_string("(");
+	stuff_float(&curve_params.min);
+	stuff_float(&curve_params.max);
```
Baezon, line 447: The user is just... typing in the curve's min and max? Those should be able to be calculated pretty easily from the curve itself (probably ahead of time and stored on the curve object) rather than allowing for possible incongruity between the actual curve and what the user types in here.

code/utils/RandomRange.h
```diff
@@ -261,4 +346,185 @@ UniformRange<Value> parseUniformRange(Value min = std::numeric_limits<Value>::mi
...
+				return (current_pos - curve_min) / (curve_max - curve_min) * (param.max - param.min) + param.min;
+			}
+			if (current_value > rand) {
+				upper_bound = current_pos;
+			} else {
+				lower_bound = current_pos;
+			}
+		}
+		// remap the values to the distribution's range
+		return (((lower_bound + upper_bound) / 2.f) - curve_min) / (curve_max - curve_min) * (param.max - param.min) + param.min;
```
Baezon, line 397: There's a possible divide by 0 here if the curve's y values are 0.

code/math/curve.h
```diff
@@ -37,12 +39,16 @@ public :
...
+	float GetValueIntegrated(float x_val) const;
+
 	//Set
 	void ParseData();
 };
 
 extern SCP_vector<Curve> Curves;
 
 extern int curve_get_by_name(const SCP_string& in_name);
+extern int pdf_to_cdf(int curve);
```
Baezon, line 52: Statically converting the curves into their cdf versions ahead of time, which you decided against, isn't a bad idea. Why not?

Baezon (changes requested):
Using an iterative approach for integration is a good idea, and should be pretty robust going forward, but using a curve in this way comes with some thorny caveats...

Since the engine doesn't know how you plan to use a curve ahead of time, y = 0 or negative y values are otherwise valid inputs, but are nonsense when converted into a cdf. Aside from the division by 0, it shouldn't be unsafe, though, I think the negative Y portions will simply never be chosen in the distribution.

There's also some possibly significant performance implications, imagine doing this iteration.... for multiple values on a particle... every frame... for each one of possibly thousands of particles.

You get what you pay for, simply being expensive isn't an inherent sin, but these concerns (including the negative Y shenanigans) are a nasty sort of 'gotcha' for modders since they involve very particular inputs when used in very particular ways, and it's awkward to document them. This isn't fatal, but really deserves some work to ensure these edge cases don't sneak up on people.

code/utils/RandomRange.h
Kestrellius: These are for remapping the range, after the curve provides its value. (And I'm pretty sure it works correctly, because I tested it and it behaved like I expected.)

I'm absolutely not budging on this one. It's _definitely_ necessary. The median use case here is that you have a handful of common distribution curves, each of which is used several times. If there wasn't a remapping where you're specifying the range, you'd have to make separate curves for all those dozens of cases, go and edit a separate file every time you wanted to tweak your particle velocity or whatever, _come up with names_ for all of them, try to keep track of which ones are for what... It's _so_ much more usable this way.

code/utils/RandomRange.h
Kestrellius: I think I'll insert some checks to just reject any curves that go below 0 being used for distributions during parsing. Same with curves with zero height.

code/math/curve.h
Kestrellius: Okay. So, I mostly followed @BMagnu's lead on this PR, because he actually understands the math. But the argument against static conversion, as I remember it, went something like this:

In order to get an inverted CDF from a PDF, we have to integrate and then invert. Integration we can do, but inversion we basically can't. The binary search approach..._somehow_ lets us skip the inversion step.

I think there was also something about elegance concerns? Though of course that's more subjective.

Kestrellius:
Re: performance --

In testing, I was spawning particles in batches of 1000 as weapon impacts, sometimes with three or four curve distributions on the effect. FSO never seemed to struggle with this, at least not until there were so many particles active at once that it started to chug just from that. I don't recall ever seeing hitches on spawning.

code/math/curve.h
Kestrellius: Okay, commentary from BMagnu that I will just copypaste here:

> TL;DR for that:
> Curves are currently not capable of representing an arbitrary integral of the input. Notably the Circular keyframe type requires stuff that curves just can't represent when integrated. And interpolating using linear functions seems just plain ugly if you can get an exact integral as well. Plus, it's really not that expensive. Especially for curves with 2 keyframes, it's actually identical to evaluating a normal curve, and beyond that, it's not that much more expensive that it's actually any sort of issue.

code/utils/RandomRange.h
Baezon: Could they perhaps be optional? I've mentioned before, it's not great to have the same thing defined twice in two different places, there's definite utility here as you've pointed out, but it's easy to imagine a modder who doesn't need to use it and isn't thinking about it being confused by being asked to provide these numbers which may well be the exact same as they already defined in their curve.

Using the curve as is would be the simple, default approach, remapping an existing curve would be the 'advanced' slightly more input approach.

Kestrellius: Hmm. Okay, yeah, not a bad idea. I can look into that.

code/math/curve.h
Baezon: Ah yes, that's right, I forgot about the need for inversion.

Issue #6313, opened 2024-08-22
Fix mysterious CTD in BWO dump


Goober5000:
As has been known for a while, the BWO dump will crash after entering a mission through the tech room while running on SSE2 builds.  Thanks to testing from Rajjia, the crash has been narrowed down to something related to fonts.tbl.  I have a hunch what might be causing this, so I expect to have a fix soon.

Goober5000:
Hunch unsuccessful.  Omega 2, give it a shot.

Goober5000:
Removing this from 24.2 unless another breakthrough happens.

Goober5000:
Smoking gun found.

PR #6315, opened 2024-08-23
fix memory management in FontManager

Goober5000:
This refactors FontManager to improve it in several ways:
1. Change all the dynamic memory management to use `new`, rather than mixing `new` and `vm_malloc`
2. Change the current font to use an index rather than a pointer
3. Take advantage of more C++11 features to avoid redundant lookups
4. Use case-insensitive name comparison
5. Avoid overflowing the `fnt->filename` field
6. Use `Warning` rather than `Error` in a non-critical error situation

Motivated by #6313 but does not fix the crash.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -5,6 +5,7 @@ modernize-*,\
 -modernize-make-unique,\
 -modernize-raw-string-literal,\
 -modernize-use-default-member-init,\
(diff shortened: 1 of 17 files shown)


wookieejedi (approved):
Works in my tests and overall looks good

